PR #119, opened 2026-06-13
fix(repl): render user echo as a true chat bubble; label spinner tips

griffinwork40:
## Summary
- REPL user-message echo now renders as a true right-positioned chat bubble: width capped at 75% of the terminal (≤100 cols), text left-aligned within a common block (straight left edge), dim top rule spanning the bubble as its top edge. Previously each wrapped row was right-aligned independently at near-full width — a ragged left edge that read as broken wrapping, not a bubble.
- Spinner tip row now carries the `Tip:` label the `LoadingTip` contract documents (`💡 Tip: /agentify — Use when…`), so a harvested skill hint rendered mid-turn can't be misread as the agent rerouting the user's command.
- Tip-row truncation hardened to display-width measurement (adversarial-verify finding): wide glyphs (CJK/emoji) in a tip body could previously overflow the one-line budget ~2:1.

## Test results
- `pnpm lint` (tsc --noEmit, strict): passed
- `pnpm scan:env:check`: passed (107 vars in sync)
- `pnpm test:coverage` (CI gate, hard thresholds): 483 files — 9105 passed | 12 skipped, exit 0

## Verification
Adversarial verifier (read-only sub-agent) independently re-derived the load-bearing claims from the diff:
- **Last-column safety preserved** (no printable glyph in the terminal's final column): CONFIRM — every content row sums to exactly `cols-1` (`(rightEdge−blockW−2) + blockW + 2`); `blockW` clamped to `rightEdge−2`; separator ≤ `rightEdge`.
- **Width math closes incl. truncation branch** (ANSI/OSC-8-aware `truncateDisplayWidth`): CONFIRM.
- **Tip row stays one visual line**: REFINE — char-count truncation could overflow on wide glyphs; **resolved in this PR** (display-width truncation + CJK regression test, commit 2b34060).
- **No collateral behavior change** (bordered card kinds, non-TTY echo, single-line inline echo): CONFIRM.
- Other risks reviewed: no remaining consumers of the old fixed-30 separator width; card-path echo height flows through the compositor's physical `measure()`, so the added geometry self-corrects.

## Out of scope
- `visualRowCount` (src/cli/input/echo.ts) intentionally serves only the inline non-card path; card-path geometry is measured physically by the compositor.

### DIFF
--- a/src/cli/loading-tips.test.ts
+++ b/src/cli/loading-tips.test.ts
@@ -317,9 +317,12 @@ describe('formatTipRow', () => {
     expect(row.startsWith('  💡 Tip: /agentify')).toBe(true);
   });
 
-  it('truncates to one visual line with a trailing ellipsis', () => {
+  it('truncates to one visual line with a single trailing ellipsis', () => {
     const row = strip(formatTipRow('x'.repeat(500), 40));
     expect(row.endsWith('…')).toBe(true);
+    // Exactly one trailing ellipsis — guards the '……' regression where
+    // truncateDisplayWidth's own ellipsis was compounded by a manual '…' append.
+    expect(row.match(/…+$/)?.[0]).toBe('…');
     expect(row.length).toBeLessThanOrEqual(40);
   });
 
--- a/src/cli/loading-tips.test.ts
+++ b/src/cli/loading-tips.test.ts
@@ -22,6 +22,7 @@ import {
   _resetSeenTipsForTesting,
   type LoadingTip,
 } from './loading-tips.js';
+import { formatTipRow } from './terminal-compositor.types.js';
 
 // Harvest stubs — vi.mock has to be hoisted, so use the standard pattern.
 vi.mock('./slash/registry.js', () => ({
@@ -301,5 +302,23 @@ describe('selectTip', () => {
   it('respects the custom warmupMs override', () => {
     expect(selectTip(pool, { startedAt: 0, now: 2_000, warmupMs: 5_000 })).toBeNull();
     expect(selectTip(pool, { startedAt: 0, now: 6_000, warmupMs: 5_000 })).not.toBeNull();
+  });
+});
+
+describe('formatTipRow', () => {
+  const strip = (s: string): string => s.replace(/\x1B\[[0-9;]*m/g, '');
+
+  it('labels the row with "Tip:" so skill hints are not misread as routing', () => {
+    // A harvested skill tip looks like `/agentify — Use when…`. Without the
+    // label, rendering it under the spinner right after the user typed a
+    // DIFFERENT slash command reads as "the agent rerouted my command".
+    const row = strip(formatTipRow('/agentify — Use when the user wants to delegate', 120));
+    expect(row.startsWith('  💡 Tip: /agentify')).toBe(true);
+  });
+
+  it('truncates to one visual line with a trailing ellipsis', () => {
+    const row = strip(formatTipRow('x'.repeat(500), 40));
+    expect(row.endsWith('…')).toBe(true);
+    expect(row.length).toBeLessThanOrEqual(40);
   });
 });
--- a/src/cli/loading-tips.test.ts
+++ b/src/cli/loading-tips.test.ts
@@ -16,6 +16,7 @@
  */
 
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import stringWidth from 'string-width';
 import {
   buildTipPool,
   selectTip,
@@ -320,5 +321,14 @@ describe('formatTipRow', () => {
     const row = strip(formatTipRow('x'.repeat(500), 40));
     expect(row.endsWith('…')).toBe(true);
     expect(row.length).toBeLessThanOrEqual(40);
+  });
+
+  it('truncates by display width, not char count (wide CJK glyphs)', () => {
+    // 30 CJK chars = 60 display columns but only 30 JS chars. Char-count
+    // truncation would keep ~29 of them (≈58 cols) and overflow a 40-col
+    // terminal; display-width truncation must keep the row ≤ cols.
+    const row = strip(formatTipRow('漢'.repeat(30), 40));
+    expect(row.endsWith('…')).toBe(true);
+    expect(stringWidth(row)).toBeLessThanOrEqual(40);
   });
 });
--- a/src/cli/loading-tips.ts
+++ b/src/cli/loading-tips.ts
@@ -25,7 +25,7 @@
  *      reproducible.
  *
  * Tips are plain text — no ANSI inside the body, no second sentence. The
- * compositor wraps them with the dim 💡 prefix.
+ * compositor wraps them with the dim `💡 Tip:` prefix.
  */
 
 import { list as listSlashCommands } from './slash/registry.js';
--- a/src/cli/render.test.ts
+++ b/src/cli/render.test.ts
@@ -468,7 +468,7 @@ describe('card', () => {
     Object.defineProperty(process.stdout, 'columns', { value: 80, configurable: true });
   });
 
-  it('user kind wraps long lines and right-aligns each row against the bar', () => {
+  it('user kind wraps long lines and terminates every row at the bar', () => {
     Object.defineProperty(process.stdout, 'columns', { value: 30, configurable: true });
     const out = strip(card({ kind: 'user', body: 'word '.repeat(15).trim() }));
     const rows = out.split('\n');
@@ -531,7 +531,7 @@ describe('card', () => {
 
   it('user kind wrapped lines all end flush right at the bar', () => {
     Object.defineProperty(process.stdout, 'columns', { value: 40, configurable: true });
-    // Body long enough to wrap. innerW = max(20, 40 - 4) = 36.
+    // Body long enough to wrap. innerW = max(20, min(36, floor(40*0.75), 100)) = 30.
     const body = 'word '.repeat(10).trim(); // 49 visible chars → wraps
     const out = strip(card({ kind: 'user', body }));
     const lines = out.split('\n');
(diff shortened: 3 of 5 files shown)


griffinwork40:
Shadow-verify **CONFIRMED** the high finding via independent live execution (trailing-ellipsis-count = 2), citing the same `display.ts:143` / `formatTipRow` mechanism. All waves complete. Emitting the review.

---

# Code Review — PR #119

**fix(repl): render user echo as a true chat bubble; label spinner tips**
`afk/20260612-194653-ac9eb6 → main` · +139/−37 · 5 files · regime: **light** (≤300 LOC)
Reviewed ref: `2b340607b6e90db40f23e52820e4839c71eb88a1` (branch HEAD). Citations verified at that ref by re-read **and** live execution of the real modules.

## Summary
Two independent changes: (1) `card.ts` reworks the REPL user-echo into a width-capped left-aligned "chat bubble" block (straight left edge, ≤75% terminal, dim top rule + cyan right bar); (2) `formatTipRow` adds a `Tip:` label and switches to display-width-aware truncation. The `card.ts` rework is **correct and well-tested** (verified by execution: max row width = `cols-1`, single left indent, aligned separator). The tip change has one confirmed output regression.

## Findings

### 🔴 HIGH · high-confidence · correctness — double trailing ellipsis on truncated tips
`src/cli/terminal-compositor.types.ts:290-293` · ref `2b34060` · citation-type: file-state (verified by re-read + execution + independent shadow-verify)

`truncateDisplayWidth(text, N)` **already appends its own `…`** when it truncates (`src/cli/display.ts:145` → `return out + ellipsis + …`, default `ellipsis = '…'`). `formatTipRow` then appends a **second** `…`, so every truncated tip renders `……`. Live output: `formatTipRow('x'.repeat(50), 40)` → `"  💡 Tip: xxx……"` (trailing-run = 2). Independently re-derived by a shadow verifier: **CONFIRMED**, trailing-count = 2. This is a **regression** — the pre-PR code `text.slice(0, bodyBudget - 1) + '…'` produced a single ellipsis. Reachable in production via `src/cli/input/spinner.ts:124`. Impact is cosmetic (width stays ≤ `cols`, 39 ≤ 40 — no overflow), but it ships visibly-wrong output on this PR's own headline feature.

```ts
const body =
  displayWidth(text) > bodyBudget
    ? truncateDisplayWidth(text, Math.max(0, bodyBudget - 1)) + '…'  // ← second '…'
    : text;
```

**Suggestion:** drop the manual append and let the helper own the ellipsis — `truncateDisplayWidth(text, bodyBudget)` (returns ≤ `bodyBudget` cols with one `…`). Or suppress the helper's ellipsis: `truncateDisplayWidth(text, Math.max(0, bodyBudget - 1), '') + '…'`.

### ⚪ LOW · high-confidence · test-coverage — truncation test cannot catch the regression
`src/cli/loading-tips.test.ts:36-40`

The "truncates to one visual line" test asserts only `endsWith('…')` (true for both `…` and `……`) and `length <= 40` (the double-ellipsis fits the budget, so this passes too). No assertion guards the single-ellipsis contract, so the regression above ships green.

**Suggestion:** add `expect(row.match(/…+$/)?.[0].length).toBe(1);` (or `expect(row.endsWith('……')).toBe(false);`).

## Considered and dismissed (not findings)
- **Short-bubble separator floor** (`card.ts:216`, `sepW = max(12, blockW+2)`): for one-word bubbles the top rule extends left of the bubble. Documented design choice ("Floor of 12 keeps it legible"), not a defect.
- **`cols ≤ 3` degenerate rendering**: the `Invariant` comment block (`card.ts:122-145`) documents the `' │'` overflow on 1–3 column terminals and confirms it degrades without throwing. Pre-existing, documented, non-real-world.
- **`string-width` test import**: verified a direct dependency (`package.json:92`, `^8.2.0`) — not a missing-dep risk.

## What was not checked
- Did **not** run the full `vitest` suite — verified the two changed functions by direct execution of the real modules instead, and the `card.ts` invariants empirically at `cols=80`.
- Did **not** review the Telegram surface — the diff is REPL-only.
- Citations are `file-state` against branch HEAD `2b34060`; all high/low findings re-read at that ref. No patch-file ambiguity (live ref available).

One confirmed, reproducible output regression (`……`) on a production-reachable path, introduced by this PR's own new code. The fix is a single line; `card.ts` and its tests are clean and verified. Recommend landing the one-line `formatTipRow` fix (+ tightening the truncation test) before merge.

This skill is read-only — no files, commits, or PR state were modified. Stopping here.

---

griffinwork40:
Build integrity independently confirmed at `d1da5c8`:
- `display.ts` exports both `displayWidth` (L40) and `truncateDisplayWidth` (L107). Signature is `truncateDisplayWidth(text, maxWidth, ellipsis = '…')`; L116 `budget = max(0, maxWidth − ellipsisWidth)` proves the ellipsis is subtracted **inside** the limit — so card.ts's 2-arg call yields `displayWidth(content) ≤ blockW`, and the width invariant holds.
- `card.ts:2` already imports `displayWidth, truncateDisplayWidth`; `wrapToWidth`/`getTerminalWidth`/`palette`/`renderCardLine` imported L3–6; `bar` (L135) and `displayRows` (L125) are local. No missing import → **no build break**.

No critical/high findings exist, so Wave 1.5 citation-verification and `/shadow-verify` are not triggered (both gated on top-severity findings). Proceeding to synthesis.

---

# Code Review — PR #119

**Target:** PR #119 "fix(repl): render user echo as a true chat bubble; label spinner tips" · base `main` ← `afk/20260612-194653-ac9eb6` @ `d1da5c8`
**Scope:** 179 lines / 5 files (2 impl, 2 test, 1 comment-only) · regime: light · dimensions: security, correctness, api-compat, test-coverage, perf-observability

**1 · medium · correctness · `terminal-compositor.types.ts:101` · ref:d1da5c8 · file-state**
`formatTipRow` overflows on terminals narrower than ~18 columns. The `Math.max(8, …)` floor on `bodyBudget` pins a minimum rendered width of `displayWidth(prefix) + 8 = 18`, ignoring `cols`.
```ts
const bodyBudget = Math.max(8, cols - displayWidth(prefix) - 1); // prefix width = 10
// cols=15 → bodyBudget=8 → rendered width 10+8 = 18 > 15
```
*Context (verified):* the floor is **pre-existing** — the old code had the same `Math.max(8, …)`. Widening the prefix from `'  💡 '` (~5 cols) to `'  💡 Tip: '` (10 cols) shifts the overflow threshold from `cols<14` to `cols<18`, so the PR enlarges an existing edge case rather than introducing one. Impact is a one-time cosmetic wrap of a decorative spinner row; sub-18-column terminals are vanishingly rare. **Non-blocking.**
*Suggestion:* `const bodyBudget = Math.max(0, cols - displayWidth(prefix) - 1)` and have the caller suppress the row when it reaches 0, or early-return `''` when `cols < displayWidth(prefix) + 2`.

**2 · low · test-coverage · `card.ts:182` · ref:d1da5c8 · file-state**
The truncation branch `displayWidth(line) > blockW ? truncateDisplayWidth(line, blockW) : line` is unexercised — all bubble tests use `'word '.repeat(n)` bodies that `wrapToWidth` splits cleanly below `blockW` (grep of `render.test.ts` confirms no single unbreakable token wider than the cap). The defensive clamp the PR comment calls out is never hit by a test.
*Suggestion:* add a body with one long URL (no spaces) > 75% terminal width; assert each row's display width ≤ `cols − 1`.

**3 · low · test-coverage · `terminal-compositor.types.ts:101` · ref:d1da5c8 · file-state**
The `Math.max(8, …)` floor (finding #1) is untested — all three new `formatTipRow` tests use `cols=40` (`loading-tips.test.ts:319–337`), where `cols−11 = 29 > 8`. The narrow-terminal path has no coverage.
*Suggestion:* add a `cols=15` case asserting the row width once the overflow fix lands.

**4 · nit · perf · `card.ts:115` · ref:d1da5c8 · file-state**
On `cols < ~24`, `innerW` floors at 20 but `blockW ≤ cols−3`, so every wrapped row is re-truncated — the 20-col wrap pass is redundant work. Cosmetic only.
*Suggestion:* clamp `innerW = Math.min(innerW, Math.max(0, rightEdge − 2))` before wrapping, or leave as-is.

## No conflicts. No security findings (synchronous string rendering, no auth/data/network surface).

## What was checked & verified
- **Build:** `display.ts` exports + `truncateDisplayWidth` signature + `card.ts` import scope all confirmed at `d1da5c8` (independently re-read by orchestrator). Compiles under strict tsconfig.
- **Width invariant:** `leadingSpace + blockW + interiorPad + 1 + 1 = cols−1` holds; empty-`displayRows` edge resolves via `maxWidth ≤ 0 → ''` (display.ts:112). `formatTipRow` truncated path = `displayWidth(prefix) + bodyBudget = cols−1`.
- **api-compat:** `formatTipRow(text, cols)` and `card(spec)` signatures unchanged; sole production caller is the spinner row — no breaking change.
- **Tests:** the 6 new assertions are substantive, not tautological — the single-trailing-ellipsis regex (`/…+$/` === `'…'`) is a genuine `……` regression guard; the CJK `stringWidth ≤ 40` case correctly exercises display-width (not char-count) truncation.

## What was NOT checked
- All analysis read against ref `d1da5c8` — citations verified at that ref.
- `string-width` v8's runtime width for the bulb codepoint U+1F4A1 was **not executed**; `displayWidth(prefix)=10` is derived from first principles (2 + width-2 emoji + 1 + 5). If v8 returns width-1 for the bulb, the math stays self-consistent (`bodyBudget` shifts by 1) — no correctness impact, only the comment's "10 cols" would be off by one.
- Test suite and `tsc --noEmit` were **not run** (skill is read-only); build soundness asserted from import/export reads, not execution.

## Merge decision: **MERGE**

No blocking defects. The build is sound, the chat-bubble width invariant and `formatTipRow` budget math both hold, and the common rendering paths are tested. All four findings are non-blocking: #1 is a pre-existing sub-18-column cosmetic edge worth a one-line guard; #2/#3 are missing-coverage follow-ups for the truncation and floor branches; #4 is cosmetic. None gate the merge.

---

griffinwork40:
🤖 _Automated review — agent-afk hourly sweep · updated 2026-06-13 02:06 UTC · re-evaluated each rotation._

✅ MERGE


  • Correctness: renderUserCard's new block layout holds the last-column-safety invariant — every row totals leadingSpace + blockW + 2 = cols-1, so the cyan bar never writes the final terminal column (src/cli/render/card.ts).
  • Build/correctness: the new formatTipRow call truncateDisplayWidth(text, n, '') is valid — truncateDisplayWidth takes a 3rd ellipsis param (src/cli/display.ts:107), so the single-… (anti-……) behavior is correct and strict tsc passes.
  • API-compat: the only production consumers are spinner.ts:124 (formatTipRow) and echo.ts:83 (card({kind:'user'})), both backward-compatible; no snapshot test depends on the old '  💡 ' prefix or per-row right-alignment.
  • Nit (non-blocking): tip row can soft-wrap on terminals ≤18 cols due to the Math.max(8, …) budget floor in terminal-compositor.types.ts; drop the floor to 0 in a follow-up.

PR #120, opened 2026-06-13
test(renderer): lock that subagent transitions can't drop the thinking paragraph (#110)

### LINKED ISSUE #110: fix(renderer): route all overlay repaints through setComposedOverlay to stop paragraph flicker
## Context

A prior change moved the orchestrator overlay from a 1x80cp tail to a ~7-row wrapped paragraph (`◆ thinking` header + indented body + `⋯ +N chars earlier` footer). The composition itself goes through `setComposedOverlay` in `src/cli/_lib/stream-renderer-orchestrator.ts:395`, which combines stage rail + thinking paragraph + tool lane + progress banner and pushes the result through `compositor.setOverlay()` atomically. Surfaced as an API-compat finding in the review of that change.

## Problem

The subagent renderer (`src/cli/_lib/stream-renderer-subagent.ts`) and a few orchestrator non-thinking paths still call `compositor.setOverlay(toolLane.getOverlay())` **directly**, bypassing the composed frame:

- `stream-renderer-subagent.ts:257, 264, 271, 309, 348, 377, 492` — every subagent state transition (tool start / progress / done / error / finalize)
- `stream-renderer.ts:337, 446, 516` — non-thinking orchestrator paths
- `stream-renderer-orchestrator.ts:286, 381` — explicit clears

Before the paragraph change this was fine: tool-lane-only and composed overlays were both ~1 row, so a direct repaint produced no visible delta vs. the composed frame. Now the composed frame is up to 7 rows taller. Every direct `setOverlay(toolLane.getOverlay())` call now:

1. Wipes the live-thinking paragraph rows
2. Repaints the next frame with the paragraph back
3. → visible flicker, especially under parallel subagents that fire repaints at independent cadences

## Why not route subagent overlays through `setComposedOverlay` literally

The intentional design (documented at `stream-renderer-orchestrator.ts:423-428` and `stream-renderer-subagent.ts:323-349`) keeps subagent thinking on the single-line ToolLane tail — parallel subagents can't all own the orchestrator's paragraph slot.

That's fine. The fix is not to give subagents their own paragraph — it's to make every callsite that mutates state visible in the composed frame call `setComposedOverlay(ctx, lastProgressByTask)` instead of `compositor.setOverlay(toolLane.getOverlay())`. `setComposedOverlay` already reads `ctx.toolLane.getOverlay()` internally, so the subagent's contribution is preserved. The paragraph + stage rail + progress banner just don't get clobbered between composed repaints.

## Constraints

- `setComposedOverlay` requires `OrchestratorCtx` and `lastProgressByTask`. The subagent renderer currently takes a narrower `ctx`. Either widen subagent ctx, or hoist `setComposedOverlay` to read `lastProgressByTask` from `ctx` (it's already on the orchestrator) and drop the second parameter.
- A prior early-return optimization means we don't pay double for the extra work `setComposedOverlay` does on no-delta calls — but we DO want callers to compose the right frame, not the bare tool lane.

## Acceptance

- [ ] No callsite outside `setComposedOverlay` calls `compositor.setOverlay(toolLane.getOverlay())` — they all go through `setComposedOverlay`
- [ ] Subagent thinking tail still renders in the ToolLane row (no behavior change)
- [ ] Existing tests pass; add a test that subagent state transitions don't drop the orchestrator's thinking paragraph from the next overlay frame
- [ ] Explicit clear callsites (`compositor.setOverlay('')`) are audited — some should clear, others should compose (tool lane may still have rows)

## References

- `src/cli/_lib/stream-renderer-subagent.ts:257-492` (7 direct setOverlay calls)
- `src/cli/_lib/stream-renderer.ts:337, 446, 516`
- `src/cli/_lib/stream-renderer-orchestrator.ts:395-442` (the composer)

griffinwork40:
Closes #110.

## TL;DR

The paragraph-flicker fix #110 asks for **already landed** with the
`OverlayComposer` migration. This PR closes the one remaining **open**
acceptance item — the regression test — and records the audit confirming the
production routing is complete.

## What #110 was

Subagent state transitions (and a few orchestrator paths) called
`compositor.setOverlay(toolLane.getOverlay())` **directly**, emitting only the
tool lane and clobbering the orchestrator's multi-row live "thinking paragraph"
between composed frames → visible flicker, worst under parallel subagents
repainting at independent cadences. (The line numbers in the issue are stale —
they predate the `OverlayComposer` migration.)

## Why no production code change is needed

The `OverlayComposer` is now the **single writer** of `compositor.setOverlay`
during a live turn (`overlay-composer.ts`). Every non-test `setOverlay`
callsite is composer-guarded:

```
if (ctx.overlayComposer) { ctx.overlayComposer.markDirty('tool-lane'); ctx.overlayComposer.flush(); }
else if (ctx.compositor) { ctx.compositor.setOverlay(ctx.toolLane.getOverlay()); }   // no-composer fallback only
```

- `stream-renderer-subagent.ts` — tool_use_detail / tool_result / tool_diff / content / thinking / error paths
- `stream-renderer-subagent-helpers.ts` — `finalizeSubagent` (done row)
- `stream-renderer-orchestrator-emit.ts` — `flushToolLaneToScrollback` + the before-content commit
- `stream-renderer.ts` — after-subagent commit + dispose safety-net
- `markdown-stream.ts` / `markdown-stream-format.ts` — markdown-pending slot

`OverlayComposer.flush()` recomposes **all** active slots in z-order, and the
production composer registers `thinking-live` first
(`stream-renderer.ts:427` + `registerOverlaySlots` in
`stream-renderer-lifecycle.ts`). So `markDirty('tool-lane') + flush()` always
re-emits the thinking paragraph alongside the tool lane — no clobber, no
flicker. The bare `setOverlay(...)` calls only run when **no** composer is
wired (tests / non-TTY), where there is no flicker concern and the subagent ctx
lacks the orchestrator's thinking/progress state anyway.

## Acceptance criteria

- [x] No callsite clobbers the composed frame in production — all route through the composer; bare `setOverlay(toolLane.getOverlay())` is the no-composer fallback only.
- [x] Subagent thinking tail still renders in the ToolLane row (no behavior change) — `setThinkingTail` paths untouched.
- [x] Existing tests pass; **added** a test that subagent state transitions don't drop the orchestrator's thinking paragraph from the next overlay frame.
- [x] Explicit `setOverlay('')` clears audited: live-turn clears (`markdown-stream.ts`, `stream-renderer.ts` borrow-dispose) route through the composer; the only bare clears are the no-composer fallback and the idle `/clear` path (`bootstrap.ts`) — both intentional.

## The test (`stream-renderer-subagent-paragraph.test.ts`)

Integration-level, production-faithful:
1. Wires a real `OverlayComposer` via the production `registerOverlaySlots`, with an **active** orchestrator thinking paragraph + a shared `ToolLane`. The composer's sink is the (mock) compositor, so every captured frame is exactly what production would paint.
2. Drives the **real** `handleSubagentEvent` through `tool_use_detail → tool_result → done`, and separately through `tool_use_detail → error`.
3. Asserts every composed frame still contains the `◆ thinking` header.
4. **Control** test proves the assertion discriminates: a bare tool-lane overlay never carries the paragraph header.

**Mutation-tested:** reverting a subagent callsite to a bare
`setOverlay(toolLane.getOverlay())` fails the test with the exact
"dropped the orchestrator thinking paragraph" diagnostic.

## Verification

- New test: 3/3 green
- `src/cli/_lib/` suite: 307/307 green

### DIFF
--- a/src/cli/_lib/stream-renderer-subagent-paragraph.test.ts
+++ b/src/cli/_lib/stream-renderer-subagent-paragraph.test.ts
@@ -1,0 +1,225 @@
+/**
+ * Regression test for issue #110 — "route all overlay repaints through the
+ * composer to stop paragraph flicker."
+ *
+ * Invariant under test: while the orchestrator's live thinking paragraph (the
+ * `'thinking-live'` overlay slot) is active, a SUBAGENT state transition
+ * (handleSubagentEvent: tool_use_detail / tool_result / done / error) must NOT
+ * drop the thinking paragraph from the next composed overlay frame.
+ *
+ * History (the bug this pins): before the OverlayComposer migration each
+ * subagent transition called `compositor.setOverlay(toolLane.getOverlay())`
+ * directly — emitting ONLY the tool lane and clobbering the multi-row thinking
+ * paragraph, producing flicker (the paragraph blanked on the subagent's
+ * repaint, then repainted on the next orchestrator-composed frame). Under
+ * parallel subagents firing repaints at independent cadences the blank/repaint
+ * cycle was continuous.
+ *
+ * The fix: every subagent callsite routes through
+ * `overlayComposer.markDirty('tool-lane') + flush()` when a composer is wired.
+ * `flush()` recomposes ALL active slots in z-order (overlay-composer.ts), so
+ * the `'thinking-live'` slot is preserved alongside the tool lane. This test
+ * pins that at the INTEGRATION level: it drives the real `handleSubagentEvent`
+ * against a real `OverlayComposer` wired through the production
+ * `registerOverlaySlots`. If a future change reverts a subagent callsite to a
+ * bare tool-lane `setOverlay`, the composed frame loses the `◆ thinking`
+ * header and these tests fail. The `control` test proves the assertion
+ * genuinely discriminates the fix from the pre-fix behavior.
+ */
+
+import { describe, it, expect } from 'vitest';
+import {
+  handleSubagentEvent,
+  synthesizeAgentEntry,
+  type SubagentCtx,
+} from './stream-renderer-subagent.js';
+import { freshSourceState, type SourceState } from './stream-renderer-source.js';
+import { OverlayComposer } from './overlay-composer.js';
+import { registerOverlaySlots } from './stream-renderer-lifecycle.js';
+import { ToolLane } from '../commands/interactive/tool-lane.js';
+import { ThinkingLane } from '../commands/interactive/thinking-lane.js';
+import { createStageTracker } from '../commands/interactive/loop-stage.js';
+import { stripAnsi } from '../display.js';
+import type { Writer } from '../slash/types.js';
+import type { StreamingMarkdownRenderer } from './stream-renderer.js';
+import type { TerminalCompositor } from '../terminal-compositor.js';
+import type { OutputEvent } from '../../agent/types.js';
+
+/** The header `formatThinkingParagraph` always emits — a stable, glyph-only
+ *  marker that survives `stripAnsi` (the `◆` is a unicode char, not ANSI). */
+const PARAGRAPH_HEADER = '◆ thinking';
+
+/** Production z-order from stream-renderer.ts:427 (the constructor wiring). */
+const ORDER = [
+  'thinking-live',
+  'markdown-pending',
+  'tool-lane',
+  'progress-banner',
+  'interrupt',
+] as const;
+
+const noopWriter: Writer = {
+  line() {},
+  raw() {},
+  success() {},
+  info() {},
+  warn() {},
+  error() {},
+};
+
+/**
+ * Wire a real OverlayComposer exactly as production does (registerOverlaySlots),
+ * with an ACTIVE orchestrator thinking paragraph and a shared tool lane. The
+ * composer's sink IS the (mock) compositor, so every frame the compositor
+ * receives is captured in `overlayFrames` — mirroring production where
+ * `new OverlayComposer(compositor, …)` makes the compositor the single sink.
+ */
+function setup() {
+  const overlayFrames: string[] = [];
+  const compositor = {
+    setOverlay: (t: string) => overlayFrames.push(t),
+    commitAbove: () => {},
+  } as unknown as TerminalCompositor;
+
+  // The orchestrator's live thinking lane drives the 'thinking-live' slot.
+  // A single non-whitespace push leaves it active + buffered, so
+  // formatThinkingParagraph(peekPhase()) renders a non-empty paragraph.
+  const thinkingLane = new ThinkingLane();
+  thinkingLane.push(
+    'Investigating why the overlay paragraph flickers under parallel subagents.',
+  );
+
+  const toolLane = new ToolLane();
+
+  const overlayComposer = new OverlayComposer(compositor, [...ORDER]);
+  registerOverlaySlots(overlayComposer, {
+    stageTracker: createStageTracker(),
+    thinkingMode: 'live',
+    thinkingLane,
+    streamingMarkdownRef: { current: null },
+    toolLane,
+    lastProgressByTask: new Map(),
+    getInterrupting: () => false,
+  });
+
+  const ctx: SubagentCtx = {
+    isTTY: true,
+    compositor,
+    overlayComposer,
+    toolLane,
+    out: noopWriter,
+    streamingMarkdown: new Map<string, StreamingMarkdownRenderer>(),
+    thinkingMode: 'summary',
+  };
+
+  return { overlayFrames, compositor, thinkingLane, toolLane, overlayComposer, ctx };
+}
+
+function chunkEvent(chunk: Record<string, unknown>): OutputEvent {
+  return { type: 'chunk', chunk } as unknown as OutputEvent;
+}
+
+describe('issue #110 — subagent transitions preserve the orchestrator thinking paragraph', () => {
+  it('keeps the thinking paragraph in every frame across tool_use_detail → tool_result → done', () => {
+    const { overlayFrames, thinkingLane, ctx } = setup();
+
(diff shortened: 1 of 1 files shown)


griffinwork40:
Both gates are green:
- **`pnpm lint` (tsc --noEmit, strict) → exit 0** — compiles under `noUnusedLocals`/`noUnusedParameters`/`noUncheckedIndexedAccess`/`noPropertyAccessFromIndexSignature`. All 15 imports are used; all signatures match; the `as unknown as` casts are valid.
- **`vitest run` → 3/3 pass (15ms)** — the test runs and all assertions hold.

This closes every substantive concern. No `critical`/`high` findings survive verification, so `/shadow-verify` is not triggered (it gates only on top-severity findings). Synthesizing inline as coordinator.

---

# Review: PR #120 — `test(renderer): lock that subagent transitions can't drop the thinking paragraph (#110)`

**Target:** PR #120 · branch `afk/tackle-issue-110-private` → `main` · reviewed ref `9de83c0` (`9de83c0f5f9eb67642f9f376a8ee833649c8f8af`)
**Change type:** test addition (regression lock) · **Files:** 1 · **Lines:** +225/-0 · **Regime:** light
**Summary:** Adds an integration-level regression test pinning that subagent state transitions (`tool_use_detail`/`tool_result`/`done`/`error`) route through the `OverlayComposer` and never drop the orchestrator's live `thinking-live` paragraph from a composed frame.

## Verification performed (ground truth, read-only)
| Check | Result | Evidence |
|---|---|---|
| Strict-TS compile | ✅ pass | `pnpm lint` (`tsc --noEmit`) exit 0 |
| Runtime pass | ✅ 3/3 | `vitest run …subagent-paragraph.test.ts` 15ms |
| Symbol signatures vs. real source | ✅ all match | Wave 1 agent read each module at ref `9de83c0` |
| Production z-order claim | ✅ accurate | `src/cli/_lib/stream-renderer.ts:427` = `['thinking-live','markdown-pending','tool-lane','progress-banner','interrupt']` |
| `◆ thinking` header emitted | ✅ confirmed | `formatThinkingParagraph` HEADER const, `thinking-paragraph.ts` |
| No early-return | ✅ confirmed | `synthesizeAgentEntry` sets `source.syntheticAgentToolUseId` (`stream-renderer-subagent-helpers.ts:202`); guard at `stream-renderer-subagent.ts:116-117` not hit |
| Control test discriminates | ✅ genuine | `toolLane.getOverlay()` renders tool rows only — never `◆ thinking` (`tool-lane.ts:285`) |

## Findings (sorted)

**nit · low confidence · documentation · `stream-renderer-subagent-paragraph.test.ts` (setup, ~line 95 / 113) · ref:9de83c0 · citation-type:diff-context**
The test sets `thinkingMode: 'summary'` on `SubagentCtx` but `thinkingMode: 'live'` in the `registerOverlaySlots` ctx. Both are correct — they drive separate paths (`SubagentCtx.thinkingMode` gates the thinking-chunk branch in `handleSubagentEvent`; the lifecycle ctx's `thinkingMode` gates the `thinking-live` overlay slot render at `stream-renderer-lifecycle.ts:88-90`) — but the split reads like a copy-paste discrepancy without a comment.
- *evidence:* `thinkingMode: 'summary'` (SubagentCtx) vs `thinkingMode: 'live'` (registerOverlaySlots)
- *suggestion:* Add a one-line comment noting the two `thinkingMode` fields gate different paths and both values are intentional.

**nit · low confidence · type-safety · `stream-renderer-subagent-paragraph.test.ts` (`chunkEvent` helper, ~line 130) · ref:9de83c0 · citation-type:diff-context**
`chunkEvent` casts `{ type: 'chunk', chunk } as unknown as OutputEvent`, so TS does not validate the inner chunk shape. All three current call sites correctly supply `toolUseId`/`toolName`/`toolInput`, so there is no live defect — this is idiomatic test-fixture fakery (consistent with the repo's existing `as unknown as` convention). Informational only.
- *suggestion (optional):* Type the param as a discriminated union of the chunk variants if you want future call sites checked at compile time.

## Dropped / corrected findings (manifest)
- **DROPPED — false-absent:** Wave 1 returned a `high · build-hygiene` finding asserting "the diff contains zero actual `it()` test bodies — only comment blocks; `noUnusedLocals` fails." This was an artifact of **my** dispatch: I abbreviated the three `it()` blocks into comment summaries in the agent's prompt to save space, and the agent reasoned about the abbreviation, not the PR. Refuted by direct re-read of the diff, `grep -c "it(" = 3`, `pnpm lint` exit 0, and 3/3 passing tests. This is exactly the false-positive the Wave 1.5 verification stage exists to catch.

## Epistemic scope / what was not checked
- **Refs read:** Wave 1 agent and I both read against branch HEAD `9de83c0` — all citations verified at that ref. Working tree was clean throughout; no mutations.
- **Tests/lint were run** (`vitest run`, `tsc --noEmit`) as read-only verification; neither mutates the repo.
- **Scope of the test itself:** it pins the *per-transition* mechanism (each subagent transition recomposes all slots, preserving the paragraph) using a **single** subagent across `tool_use_detail → tool_result → done` plus an error path. It does **not** drive *multiple concurrent* subagents at independent cadences (the literal scenario in the bug's History comment). This is acceptable — per-transition recomposition is the root invariant; if it holds for one subagent's transitions it holds under parallel cadence — but the test is a targeted mechanism lock, not a full parallel-flicker integration test.
- **Not checked:** whether the underlying #110 production fix is itself correct (out of scope — this PR adds only the test; the fix landed earlier). I verified the production code paths the test depends on exist and behave as the test assumes, not the fix's broader correctness.

## Merge decision: **MERGE**

The test compiles under the strict-TS gate, runs green (3/3), exercises real production wiring (`registerOverlaySlots` is the same function used at `stream-renderer.ts:438`), and its `control` test proves the assertions are load-bearing rather than tautological. Zero blocking findings; the two surviving items are documentation/style nits that do not gate merge. This is a well-constructed regression lock for #110.

---

Issue #109, opened 2026-06-12
Add output schema enforcement to /ground-state forkSubagent dispatch


griffinwork40:
## Problem

`/ground-state` returns a freeform-prose snapshot with no machine-checkable structure. The orchestrator has no programmatic way to detect that a surveyor or synthesizer overshot from "snapshot" into "implementation" before damage is done — by the time the response surfaces, mutating tool calls have already executed.

A prior prompt-level mitigation added an explicit read-only constraint block to the ground-state SKILL.md. This issue covers the **defense-in-depth output-schema enforcement** that catches scope violations the prompt-level constraint might miss.

## Evidence (file:line citations)

- **`src/bundled-plugins/awa-bundled/skills/ground-state/SKILL.md:40-48`** — "Synthesize into a 6-line ground-truth snapshot" specifies a prose template only. No JSON shape. No `scope_check` field. No sentinel.
- **`src/agent/tools/skill-executor.ts:406-413`** — `forkSubagent()` call from `executePluginSkill` passes no `outputSchema`.
- **`src/agent/subagent.ts:67-71`** — `SubagentHandle.runToResult` API supports optional Zod `outputSchema` validation. The mechanism exists; the ground-state dispatch path just doesn't use it.
- **Compare**: `src/skills/_agents/research-agent.md:34-53` — explicit return shape requiring `{findings, evidence_pointers, scope_check, boundary_flag}` JSON block. Contract convention (`contract.md:17`) requires results "in that exact shape."

## Proposed fix

1. **Define a Zod schema for the ground-state snapshot.** Mirrors the 6-bullet template, adds a required sentinel field:
   ```ts
   const GroundStateSnapshot = z.object({
     scope: z.literal('snapshot-only'),
     branch: z.object({ name: z.string(), state: z.enum(['clean', 'diverged']), upstream: z.enum(['fresh', 'stale']) }),
     recent_work: z.array(z.string()).max(3),
     infrastructure: z.string(),
     memory_hits: z.union([z.array(z.string()), z.literal('none')]),
     implementation_risks: z.array(z.string()),
     epistemic_confidence: z.enum(['high', 'medium', 'low']),
   });
   ```
2. **Pass it to `forkSubagent`.** Update the ground-state dispatch site to supply `outputSchema: GroundStateSnapshot`.
3. **Update SKILL.md** to instruct the synthesizer to return the snapshot in the schema's exact shape (JSON block in the final message), per the `contract.md` convention.
[...]

PR #123, opened 2026-06-13
feat(skills): enforce tools: allowlist in plugin-skill dispatch (#108)

### LINKED ISSUE #108: Wire `tools:` frontmatter through plugin-skill dispatch to enforce read-only allowlists
## Problem

Plugin skills authored with a `tools:` frontmatter list (e.g. `research-agent.md:5` declares `tools: Read, Grep, Glob, WebFetch, WebSearch`) are silently dropped by the dispatcher. Those skills *think* they're mechanically restricted to read-only tools; in practice, every subagent dispatched via `executePluginSkill` is granted the full `CHILD_ALLOWED_TOOLS` surface — including `edit_file` and `write_file`.

This is the structural gap behind a prior prompt-only mitigation for a `/ground-state` overshoot incident (which adjusted one skill's SKILL.md). This issue covers the **infrastructure-level enforcement** that lifts the floor for every plugin skill.

## Evidence (file:line citations)

- **`src/agent/tools/tool-injector.ts`** — `parseSkillMetadata()` only reads `name`, `description`, `argumentHint` from frontmatter. A `tools:` field is silently discarded.
- **`src/agent/tools/skill-executor.ts:373-430`** — `executePluginSkill()` builds `childConfig` with only `{model, systemPrompt}`. No `tools` slot is populated.
- **`src/agent/tools/nesting.ts:41`** — `CHILD_ALLOWED_TOOLS = [...BUILTIN_TOOL_NAMES, 'agent', 'skill']` includes `edit_file` and `write_file` for every child agent by default.
- **`src/agent/providers/anthropic-direct/index.ts:137-156`** — `AnthropicDirectProvider` bakes permissions at construction; even if `config.tools` were set later, `query()` never re-reads it.
- **`src/agent/tools/permissions.ts:20-36`** — `checkToolPermission()` with no `allowedTools` returns `{ allowed: true }` for every tool.

## Proposed fix

End-to-end wiring of a declarative read-only allowlist:

1. **Parse `tools:` from frontmatter.** Extend `parseSkillMetadata` in `src/agent/tools/tool-injector.ts` to read a `tools:` field — either a YAML list or comma-separated string. Validate against `BUILTIN_TOOL_NAMES`.
2. **Propagate into `childConfig`.** In `executePluginSkill` (`src/agent/tools/skill-executor.ts:373-430`) and `buildForkedChildConfig` (line 258-293), populate a new `tools?: ToolConfig` slot from the parsed metadata.
3. **Apply at provider construction.** Pass the allowlist into `createChildProviderFactory` (`src/agent/tools/nesting.ts`) so `AnthropicDirectProvider` is constructed with `permissions: { allowedTools: <skill's list> }`.
4. **Sensible default.** When `tools:` is absent, behavior matches today (full `CHILD_ALLOWED_TOOLS`). Backward-compatible.

## Acceptance criteria

- A SKILL.md with `tools: read_file, grep, glob, list_directory` cannot call `edit_file` at runtime — provider rejects with the standard "tool not allowed" error.
- Existing plugin skills with no `tools:` frontmatter are unaffected (default surface preserved).
- `/research`, `/gather`, `/explain`, `/orient`, `/shadow-verify`, `/ground-state` get an opt-in path to declare themselves read-only mechanically (a separate follow-up can add the frontmatter; this issue only covers the enforcement infrastructure).
- New unit tests cover: frontmatter parsing, allowlist propagation through the dispatch chain, and provider-level rejection of disallowed tools for a skill-dispatched subagent.

## Context

Root cause was traced through the full dispatch chain via a `/diagnose` session (0.99 confidence). A prior prompt-only mitigation for `/ground-state` added an explicit `**Constraint: read-only reconnaissance.**` block to that skill's SKILL.md body; this issue is the durable fix that makes such prompt-level mitigations enforceable rather than advisory.

## Scope

Infrastructure only. **Out of scope** for this issue:
- Adding `tools:` frontmatter to existing skills (separate follow-up)
- Output schema enforcement for ground-state (tracked separately as the defense-in-depth companion)
- Auditing which peer skills have the same prompt-only gap

griffinwork40:
Implements the enforcement infrastructure requested in #108.

## Summary
Plugin skills (`SKILL.md`) can now declare a `tools:` YAML frontmatter field to mechanically restrict their forked subagent to an enumerated tool surface. Calling any unlisted tool produces the standard `checkToolPermission` rejection at the provider layer — the same path already proven by `read-only: true`.

- **Parse** — `parseSkillMetadata` reads `tools:` (comma-separated *or* `[bracket, array]` form), validates each entry against `BUILTIN_TOOL_NAMES`, deduplicates, and debug-logs dropped unknown names.
- **Propagate** — `PluginSkillMetadata` / `PluginSkillBody` gain `allowedTools?: string[]`; `discoverPluginSkillBodies` forwards it.
- **Enforce** — `buildForkedChildConfig` threads a `customAllowedTools` param into both the provider-factory path and the depth-cap fallback (`buildCustomAllowlistProvider`); the allowlist is also propagated into the child `SubagentExecutor` so depth-2 `agent` fan-out stays gated.
- **Precedence / back-compat** — `read-only: true` wins over `tools:` (maps to `RECON_ALLOWED_TOOLS`); skills with no `tools:` field are unchanged (full `CHILD_ALLOWED_TOOLS`).

## Test results
- `pnpm exec tsc --noEmit` (strict): **0 errors**
- `pnpm test`: **9127 passed**, 12 skipped, **0 failed** (484 files)
- +17 tests: frontmatter parsing, bridge propagation, factory/fallback enforcement, precedence, and grandchild-propagation regressions.

## Verification
An adversarial read-only verifier independently re-derived the load-bearing claims from the diff and **CONFIRMED all three**: (1) a `tools:` skill's provider is constructed with exactly that allowlist and `edit_file` is denied; (2) skills with no `tools:` field keep the full default surface; (3) `read-only: true` takes precedence. It additionally flagged two latent gaps beyond the original acceptance criteria, both now **fixed** in the second commit (each mirrors a convention the read-only path already follows):

- **Shared-reference** — the factory path passed the allowlist array by reference; now materialized as a per-call copy (`[...customAllowedTools]`), matching `buildReadOnlyReconProvider` / `buildCustomAllowlistProvider`.
- **Grandchild escape** — the custom allowlist was not propagated into the child `SubagentExecutor`, so a `tools:`-restricted skill that includes `agent` could spawn an unconstrained grandchild. Now forwarded (no bash gate — custom `tools:` does not gate bash, only `read-only` does), covered by two new regression tests.

## Out of scope (per spec)
- Adding `tools:` frontmatter to existing bundled skills (separate follow-up).
- Output-schema enforcement for `/ground-state` — tracked as #109.

Closes #108

### DIFF
--- a/src/agent/plugins/tool-injector.test.ts
+++ b/src/agent/plugins/tool-injector.test.ts
@@ -395,4 +395,72 @@ description: No body
       expect(extractPluginName('/some/random/path/plugin-name')).toBe('plugin-name');
     });
   });
+
+  describe('tools: frontmatter allowlist parsing', () => {
+    function writeSkillFile(dir: string, frontmatter: string): void {
+      const fs = require('fs');
+      fs.mkdirSync(dir, { recursive: true });
+      writeFileSync(
+        join(dir, 'SKILL.md'),
+        `---\nname: allowlist-test\ndescription: A test skill\n${frontmatter}---\nContent`,
+      );
+    }
+
+    it('parses comma-separated tools: field', () => {
+      const skillDir = join(tmpDir, 'skills', 'comma-sep');
+      writeSkillFile(skillDir, 'tools: read_file, grep, glob\n');
+      const skills = extractPluginSkills(tmpDir);
+      expect(skills[0]!.allowedTools).toEqual(['read_file', 'grep', 'glob']);
+    });
+
+    it('parses bracket-array tools: field', () => {
+      const skillDir = join(tmpDir, 'skills', 'bracket-arr');
+      writeSkillFile(skillDir, 'tools: [read_file, grep]\n');
+      const skills = extractPluginSkills(tmpDir);
+      expect(skills[0]!.allowedTools).toEqual(['read_file', 'grep']);
+    });
+
+    it('silently drops unknown tool names', () => {
+      const skillDir = join(tmpDir, 'skills', 'unknown-tools');
+      writeSkillFile(skillDir, 'tools: read_file, not_a_tool, grep\n');
+      const skills = extractPluginSkills(tmpDir);
+      expect(skills[0]!.allowedTools).toEqual(['read_file', 'grep']);
+    });
+
+    it('deduplicates repeated tool names', () => {
+      const skillDir = join(tmpDir, 'skills', 'dedup-tools');
+      writeSkillFile(skillDir, 'tools: read_file, read_file, grep\n');
+      const skills = extractPluginSkills(tmpDir);
+      expect(skills[0]!.allowedTools).toEqual(['read_file', 'grep']);
+    });
+
+    it('omits allowedTools when all names are unknown', () => {
+      const skillDir = join(tmpDir, 'skills', 'all-unknown');
+      writeSkillFile(skillDir, 'tools: not_a_tool\n');
+      const skills = extractPluginSkills(tmpDir);
+      expect(skills[0]!.allowedTools).toBeUndefined();
+    });
+
+    it('omits allowedTools when tools: field is absent', () => {
+      const skillDir = join(tmpDir, 'skills', 'no-tools-field');
+      writeSkillFile(skillDir, '');
+      const skills = extractPluginSkills(tmpDir);
+      expect(skills[0]!.allowedTools).toBeUndefined();
+    });
+
+    it('tools: and read-only: true coexist; both are parsed into metadata', () => {
+      const skillDir = join(tmpDir, 'skills', 'coexist');
+      writeSkillFile(skillDir, 'tools: read_file\nread-only: true\n');
+      const skills = extractPluginSkills(tmpDir);
+      expect(skills[0]!.allowedTools).toEqual(['read_file']);
+      expect(skills[0]!.readOnly).toBe(true);
+    });
+
+    it('handles quoted values', () => {
+      const skillDir = join(tmpDir, 'skills', 'quoted');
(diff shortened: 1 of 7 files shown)


griffinwork40:
🤖 _Automated review — agent-afk hourly sweep · updated 2026-06-13 14:10 UTC · re-evaluated each rotation._

🚫 DO NOT MERGE


Blocking: quoted-array tools: frontmatter fails open to the full tool surface.


  • High — src/agent/plugins/tool-injector.ts:180-182: tools: ["read_file", "grep"] (standard quoted YAML flow array) isn't parsed — only whole-value outer quotes are stripped, never per-element quotes, so tokens stay "read_file", fail the BUILTIN_TOOL_NAMES check, and are dropped. allowedTools becomes undefined and the forked child falls back to the full CHILD_ALLOWED_TOOLS surface (bash/write_file/edit_file) — fail-open, defeating the stated purpose. Fix: strip quotes per token after the split.
  • Medium — same root cause: any all-invalid tools: list (typos) also drops to undefined → identical silent fail-open; warn (not just debugLog) instead of falling through to full tools.
  • Low — tool-injector.test.ts: covers bare-string (:411), unquoted-array (:418), and whole-value-quoted (:459) forms, but not the quoted-array form — the exact gap that let this ship.
  • Verified clean: readOnly precedence holds at all 3 fork sites; depth-2 propagation wired; no breaking API changes; tools: can only narrow, never widen, the surface.

PR #137, opened 2026-06-14
fix(mcp): emit OAuth state param so Mintlify-style servers connect

griffinwork40:
## Summary

- Adds a `state()` method to `KeychainOAuthProvider` (`src/agent/mcp/oauth.ts`) so the MCP OAuth authorization URL now carries the `state` parameter.
- Fixes connection failures against OAuth authorization servers that **require** `state` — notably Mintlify's admin MCP (`https://mcp.mintlify.com`), which previously rejected AFK's authorize request with `{"error":"invalid_request","details":{"fieldErrors":{"state":["Required"]}}}`.
- Root cause: `@modelcontextprotocol/sdk` (`client/auth.js`) reads `provider.state?.()` and only appends `state` to the URL when the provider supplies one. `KeychainOAuthProvider` implemented no such method, so `state` was always omitted. `state` is only *RECOMMENDED* by OAuth 2.1 (PKCE already defends the code exchange against CSRF), which is why the SDK leaves it to the provider.

## Implementation notes

- `state()` returns a random `node:crypto` `randomUUID`, persisted per-server in the keychain slot (`McpOAuthSlot.state`) so repeated calls within one flow are stable.
- Regenerated when the `verifier` or `all` credential scopes are invalidated (paired with the PKCE verifier as a per-authorization-attempt artifact).
- The SDK does **not** validate the echoed `state` on the token exchange, so this is presence-only: a no-op for servers that don't require `state`, and just enough to satisfy those that do.

## Test results

- `pnpm lint` (`tsc --noEmit`, strict): **passed**
- `pnpm test` (`vitest run`, full suite): **9151 passed | 12 skipped across 485 files**, exit 0
- `pnpm build` (`tsc` + copy-prompts): **passed**; `state()` confirmed present in `dist/agent/mcp/oauth.js`
- New coverage: 6 unit tests in `src/agent/mcp/oauth.test.ts` (non-empty, stable/persisted, cross-instance persistence, per-server scoping, regeneration on both `verifier` and `all` invalidation)

## Verification

No adversarial verify requested (diff is 79 lines, under the 100-line threshold).

## Out of scope (follow-up)

- `/mcp auth` currently displays a **stripped** authorization URL: `writeOauthPending()` persists only `origin + pathname`, so the param-bearing URL reaches the user only via Telegram/stderr at connect time. Not addressed here — worth a separate change so `/mcp auth` surfaces a directly-usable URL.

---

## Follow-up fix — permission gate for late-discovered (OAuth) MCP tools (commit `066b836`)

The `state` fix above makes Mintlify **connect** and discover all its tools — but exposed a second, downstream bug: the tools connect yet are **not callable**. Every invocation is rejected with `Tool "mcp__mintlify__…" is not in the configured allowlist`.

**Root cause.** The tool permission allowlist is snapshotted **once** at provider construction (`cli/shared-helpers.ts:allowedToolsFor`, which appends `mcpManager.getMcpToolWireNames()`). OAuth servers finish their handshake and discover tools *asynchronously — after* that snapshot. The dispatcher reads MCP `schemas`/`handlers` live each query (so the model sees all the tools), but the permission object handed to the gate (`tools/dispatcher.ts:492`) stays frozen, missing the `mcp__…` names. `onToolsRefreshed` (`anthropic-direct/index.ts:262`) invalidated only the tool/handler caches — never the allowlist. Net: tool **visible + dispatchable but permission-rejected**.

**Fix.** Re-union the live MCP wire-names into the allowlist at `buildDispatcher()` time (every query) in **both** providers, via a shared pure helper `withMcpToolsAllowed()` (`tools/permissions.ts`). It allocates a new config only when names are actually missing — never mutates the base — and fires only when an `mcpManager` is present. Restricted sub-agent providers (recon / read-only / skill-scoped, built in `tools/nesting.ts`) never receive an `mcpManager`, so their allowlists are untouched: **no privilege escalation**.

**Tests (this commit).**
- `withMcpToolsAllowed` unit tests in `tools/permissions.test.ts` (undefined / empty / dedup / no-mutation-of-base / gate-passes-after-union).
- Provider-level regression in `anthropic-direct/mcp-permission-refresh.test.ts`: drives a **real** stdio MCP manager through the real provider + dispatcher with a static allowlist that omits the MCP tool — asserts the MCP tool now executes **and** a non-MCP tool (`write_file`) absent from the allowlist is still denied (the union is scoped, not a blanket open).
- Full suite green: **9160 passed | 12 skipped across 486 files**; `pnpm lint` clean; `pnpm audit:sdk:check` OK.

### DIFF
--- a/src/agent/mcp/oauth.test.ts
+++ b/src/agent/mcp/oauth.test.ts
@@ -153,6 +153,53 @@ describe('KeychainOAuthProvider — discovery state', () => {
 });
 
 // ---------------------------------------------------------------------------
+// state (OAuth state parameter)
+// ---------------------------------------------------------------------------
+
+describe('KeychainOAuthProvider — state', () => {
+  it('returns a non-empty string', () => {
+    const provider = new KeychainOAuthProvider('srv', makeMemoryBackend());
+    const s = provider.state();
+    expect(typeof s).toBe('string');
+    expect(s.length).toBeGreaterThan(0);
+  });
+
+  it('is stable across repeated calls (persisted)', () => {
+    const provider = new KeychainOAuthProvider('srv', makeMemoryBackend());
+    expect(provider.state()).toBe(provider.state());
+  });
+
+  it('persists across provider instances sharing one backend', () => {
+    const store = { blob: undefined as string | undefined };
+    const first = new KeychainOAuthProvider('srv', makeMemoryBackend(store));
+    const generated = first.state();
+    const second = new KeychainOAuthProvider('srv', makeMemoryBackend(store));
+    expect(second.state()).toBe(generated);
+  });
+
+  it('scopes state per server name', () => {
+    const store = { blob: undefined as string | undefined };
+    const provA = new KeychainOAuthProvider('server-a', makeMemoryBackend(store));
+    const provB = new KeychainOAuthProvider('server-b', makeMemoryBackend(store));
+    expect(provA.state()).not.toBe(provB.state());
+  });
+
+  it('regenerates after invalidateCredentials("verifier")', () => {
+    const provider = new KeychainOAuthProvider('srv', makeMemoryBackend());
+    const before = provider.state();
+    provider.invalidateCredentials('verifier');
+    expect(provider.state()).not.toBe(before);
+  });
+
+  it('regenerates after invalidateCredentials("all")', () => {
+    const provider = new KeychainOAuthProvider('srv', makeMemoryBackend());
+    const before = provider.state();
+    provider.invalidateCredentials('all');
+    expect(provider.state()).not.toBe(before);
+  });
+});
+
+// ---------------------------------------------------------------------------
 // invalidateCredentials
 // ---------------------------------------------------------------------------
 
--- a/src/agent/mcp/oauth.ts
+++ b/src/agent/mcp/oauth.ts
@@ -35,6 +35,7 @@
 
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'node:fs';
 import { execFileSync } from 'node:child_process';
+import { randomUUID } from 'node:crypto';
 import { homedir, userInfo } from 'node:os';
 import { join, dirname } from 'node:path';
 
@@ -63,6 +64,8 @@ interface McpOAuthSlot {
   clientInfo?: OAuthClientInformationMixed;
   codeVerifier?: string;
   discoveryState?: OAuthDiscoveryState;
+  /** OAuth `state` parameter — see `KeychainOAuthProvider.state()`. */
+  state?: string;
 }
 
 /** Shape of the `mcpOAuth` top-level key in the credentials JSON. */
@@ -327,6 +330,29 @@ export class KeychainOAuthProvider implements OAuthClientProvider {
     return v;
   }
 
+  /**
+   * OAuth `state` parameter.
+   *
+   * `state` is only RECOMMENDED by OAuth 2.1 (PKCE already defends the code
+   * exchange against CSRF), so the SDK omits it from the authorization URL
+   * unless the provider supplies one. Some authorization servers nonetheless
+   * REQUIRE `state` and reject the authorize request with `invalid_request`
+   * when it is absent — Mintlify's admin MCP (`https://mcp.mintlify.com`) is
+   * one such server.
+   *
+   * We generate a random value once per server and persist it so repeated
+   * `state()` calls within a flow return the same value. The SDK does not
+   * validate the echoed value on the token exchange, so this exists purely to
+   * satisfy servers that mandate the parameter's presence.
+   */
+  state(): string {
+    const existing = this._readSlot().state;
+    if (existing) return existing;
+    const generated = randomUUID();
+    this._updateSlot((slot) => ({ ...slot, state: generated }));
+    return generated;
+  }
+
   saveDiscoveryState(state: OAuthDiscoveryState): void {
     this._updateSlot((slot) => ({ ...slot, discoveryState: state }));
   }
@@ -341,7 +367,12 @@ export class KeychainOAuthProvider implements OAuthClientProvider {
       const updated = { ...slot };
       if (scope === 'client') delete updated.clientInfo;
       if (scope === 'tokens') delete updated.tokens;
-      if (scope === 'verifier') delete updated.codeVerifier;
+      if (scope === 'verifier') {
+        // `state` and the PKCE verifier are both per-authorization-attempt
+        // artifacts — drop them together so a re-auth starts a clean flow.
+        delete updated.codeVerifier;
+        delete updated.state;
+      }
       if (scope === 'discovery') delete updated.discoveryState;
       return updated;
     });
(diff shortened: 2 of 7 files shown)


griffinwork40:
🤖 _Automated review — agent-afk hourly sweep · updated 2026-06-14 13:08 UTC · re-evaluated each rotation._

✅ MERGE


  • oauth.ts state(): generates a UUID once per server, persists it, returns it stably, and clears it on verifier/all invalidate — satisfies servers that mandate the OAuth state param; PKCE already defends the code exchange, so not validating the echoed value is acceptable (not a CSRF hole; state is not a secret).
  • permissions.ts:60 withMcpToolsAllowed() is immutable (returns the same ref on no-op, never mutates base) and unions only MCP wire-names; restricted sub-agent providers get no mcpManager, so no allowlist widening / privilege escalation.
  • Fix is effective, not a no-op: buildDispatcher() runs per-query (anthropic-direct/index.ts:626) and re-unions live names from getMcpToolWireNames() (manager.ts:388), which are the same nameRegistry keys used to build schemas (:285) and handlers (:300) — so async-discovered OAuth tools pass the gate on the next turn.
  • Solid coverage: state() cases (oauth.test.ts), all withMcpToolsAllowed branches (permissions.test.ts), and a +256-line late-discovery regression test (mcp-permission-refresh.test.ts). All referenced symbols exist at the head ref; branch is MERGEABLE. Nit: the title omits the permission-refresh half, but both changes are one coherent OAuth-MCP fix.

griffinwork40:
Independently confirmed: `nesting.ts` (the restricted sub-agent provider factory) contains **zero** `mcpManager` references — all 8 restricted-provider constructions pass only `permissions`/`readOnly*` flags. The only sites passing `mcpManager` are top-level surfaces (`shared-helpers`, `daemon`, `chat`, `bootstrap`, `telegram`). The privilege-escalation boundary holds. And the `state` nonce is never logged. All dimensions are now covered and verified.

One synthesis-level correction to the `corr_test_perf` agent's TOCTOU finding: `state()` contains **no `await`** (`_readSlot`/`_updateSlot` use `readFileSync`/`writeFileSync`), so in single-threaded Node it is effectively atomic within a process — the race it described between "two in-process callers" cannot interleave. The only real window is cross-process, and even that is benign because the SDK doesn't validate the echoed `state`. I've downgraded it accordingly.

---

# Code Review — PR #137

**`fix(mcp): emit OAuth state param so Mintlify-style servers connect`**
Branch `fix/mcp-oauth-state-param` → `main` · reviewed ref `066b836` · +443/−7, 7 files (~91 production / ~352 test)
Regime: **full** (security-sensitive: permission gate + OAuth credential handling) · Waves: 2 dimension agents + verification + local synthesis

No blocking, critical, or high findings. The fix is correct, well-tested (44/44 focused tests pass), the security boundary holds, and api-compat is confirmed. Four minor/optional items below — none gate merge.

---

## Findings (none blocking)

**1 · nit · correctness · `src/agent/mcp/oauth.ts:349–353` · ref:066b836 · file-state**
`state()` does read-then-generate-then-persist across separate `_readSlot()`/`_updateSlot()` calls. *Correction to investigation:* because `state()` has no `await` (both helpers are `*FileSync`), it is atomic within a single Node process — the in-process race is not reachable. A benign window exists only across two concurrent *processes* sharing the credential file, and even then it is inconsequential because the SDK does not validate the echoed `state`.
*Optional hardening:* fold into one callback — `this._updateSlot((slot) => slot.state ? slot : { ...slot, state: randomUUID() })` then return the slot's value. Removes the latent smell if the SDK ever starts validating `state`.

**2 · low · test-coverage · `src/agent/providers/openai-compatible/index.ts` · ref:066b836**
The `withMcpToolsAllowed` wiring has a dedicated integration regression test for `AnthropicDirectProvider` (`mcp-permission-refresh.test.ts`) but none for the OpenAI-compatible path. Absence confirmed (grep: zero matches). Risk is low — both providers call the identical, unit-tested `withMcpToolsAllowed` with identical guard logic. *Optional:* mirror the test for the OpenAI path.

**3 · nit · style · `src/agent/providers/openai-compatible/index.ts:342–349` · ref:066b836 · file-state**
Structural asymmetry: anthropic-direct always assigns `permissions` (`this.mcpManager ? withMcpToolsAllowed(...) : this.permissions`), openai-compatible guards with `if (effectivePermissions !== undefined)`. Functionally equivalent today; aligning the two patterns would ease future auditing.

**4 · nit · observability · `src/agent/tools/permissions.ts:54–73` · ref:066b836 · file-state**
No debug log fires when the live union actually widens the allowlist (`changed === true`), so traces don't show whether late-discovery widening happened. *Optional:* one `debug` line naming the added wire-names.

---

## Verified clean (read at ref `066b836`)

- **Privilege escalation — SAFE.** `nesting.ts` (sole restricted-provider factory) has zero `mcpManager` references; all 8 restricted constructions omit it, so sub-agent allowlists are never widened. Re-confirmed by independent grep of all provider construction sites.
- **api-compat — SAFE.** `McpManager.getMcpToolWireNames()` exists (`manager.ts:387`); `state(): string` is a compatible subtype of the SDK's `OAuthClientProvider.state?(): string | Promise<string>` (`auth.d.ts:33`); both `withMcpToolsAllowed` and `state()` are purely additive — no existing signature changed.
- **`invalidateCredentials` — correct.** `'all'` returns `{}` (clears `state`); `'verifier'` explicitly deletes `state`. Both test assertions are sound, not green-by-accident.
- **`withMcpToolsAllowed` logic — correct.** Early returns share the same `base` reference; `base.allowedTools` never mutated; Set dedup correct; union is MCP-scoped (non-MCP tools stay denied — proven by the "still rejects" test).
- **Secret exposure — none.** `state` is a random CSRF nonce (same risk profile as the already-stored PKCE `codeVerifier`); never logged.
- **OAuth `state` semantics — informational.** Provides no runtime CSRF defense (SDK doesn't validate the echo); it is presence-satisfying for servers that mandate the param. PKCE is the real defense. This matches the PR's stated purpose — working as intended, not a defect.

## What was NOT checked

- Wave 1 agents read against branch HEAD `066b836`; all blocking/high-eligible citations and absence claims were verified at that ref (no fabricated/false-absent findings; nothing dropped).
- The full focused test suite was run for `permissions.test.ts` + `oauth.test.ts` (44 pass); the project-wide `pnpm test` and the new `mcp-permission-refresh.test.ts` integration test (stdio fixture) were **not** executed here.
- No live OAuth handshake against a real Mintlify server was performed — the fix's end-to-end effect (SDK actually emitting `state` in the authorize URL) is verified by code/interface inspection, not a live flow.
- `getMcpToolWireNames()` thread-safety under concurrent `tools/list_changed` refresh was not analyzed.
- No `shadow-verify` wave was run: the contract triggers it only for surviving critical/high findings, and there are none.

### Done

- **What was done:** Full read-only review of PR #137 — 2 parallel dimension agents (security·api-compat, correctness·test-coverage·perf-observability) + a citation/absence verification wave, with local synthesis and two independent inline backstops (privilege-escalation grep, log-leak grep).
- **Evidence:** All findings cite `file:line` at reviewed ref `066b836`; verification manifest dropped/downgraded nothing (all citations `verified`, all absence claims `confirmed-absent`); focused tests 44/44 pass.
- **Recommendation: MERGE.** No blocking issues. Four optional polish items (1 nit correctness, 1 low test-gap, 2 nits) recorded above for the author's discretion.
- **Read-only:** no files, commits, PR comments, or PR state were modified. Working tree untouched (the 2 untracked dirty files were never part of the review).
- **Pending/deferred:** project-wide `pnpm test`, the stdio-fixture integration test, and a live Mintlify OAuth handshake were not executed in this read-only pass — run them in CI before merge if not already green.

---

PR #143, opened 2026-06-15
fix(paths): move transcripts into state/ and honor AFK_STATE_DIR

griffinwork40:
## What & why

Autosaved REPL transcripts were landing in `~/.afk/transcripts/` — a **top-level sibling of `state/`** — instead of the state tier that every other session category lives in (`sessions/`, `todos/`, `memory/`, `witness/`, `bg/`). Transcripts were the only state category with no `paths.ts` helper; the dir was inlined in `transcript.ts`.

While there, the inlined resolver was also the **only consumer of `AFK_STATE_DIR`** — and it misread the var: it treated it as the `~/.afk` home root (appending `transcripts/`), whereas `AFK_STATE_DIR` is documented as the state dir (`$AFK_HOME/state`). Meanwhile `getAfkStateDir()` ignored the env var entirely, so `AFK_STATE_DIR` was effectively dead config for every other path.

## Changes

- **`getAfkStateDir()` now honors `AFK_STATE_DIR`** (`?? $AFK_HOME/state`), with the same absolute-path guard `getAfkHome()` uses — justified because the var now governs the entire state tier, so a relative/`/` value is a real footgun.
- **New `getTranscriptsDir()`** → `join(getAfkStateDir(), 'transcripts')`, placed alongside the other state-category helpers.
- **New `ensureTranscriptsMigrated()`** (mirrors `ensureSessionsMigrated`/`ensureTodosMigrated`) using the existing `migrateDirOnce` helper. `initTranscript()` calls it on first access, relocating any legacy `~/.afk/transcripts/`.
- **Docs** updated: `AFK.md`, `CLAUDE.md`, `docs/architecture.md`, `docs/reference.md`.
- **Docs website** (`website/content/docs/`) brought in line: `configuration/overview.mdx` (state tree gains `transcripts/` + a `$AFK_STATE_DIR overrides this tier` annotation), `configuration/environment-variables.mdx` (`AFK_STATE_DIR` now documented as governing the **entire** state tier, not one subdir, plus the absolute-path requirement), and `how-it-works.mdx` (inline state list).
- **Env-registry source of truth** corrected: `src/config/env.ts`'s `AFK_STATE_DIR` description (the prose docs above mirror it), with `docs/env-registry.{md,json}` regenerated via `pnpm scan:env` — `scan:env:check` stays green.

## Compatibility

- **Default (no `AFK_STATE_DIR` set):** transcripts move `~/.afk/transcripts/` → `~/.afk/state/transcripts/`, auto-migrated on next REPL start. No other path changes.
- **`AFK_STATE_DIR` set (e.g. `/foo`):** transcripts stay at `/foo/transcripts` exactly as before — but now *all* state honors the override too. Migration no-ops (`migrateDirOnce` early-returns on `oldPath === newPath`).
- Migration is best-effort and idempotent; never overwrites an existing target.

## Verification

- `pnpm lint` (tsc --noEmit, strict): **clean**
- `pnpm audit:env:check`: **0 violations** (534 files scanned)
- `pnpm test` (full): **490 files, 9210 passed / 12 skipped, 0 failures**
- New tests: `getTranscriptsDir` placement, `getAfkStateDir` honors + validates `AFK_STATE_DIR`, `ensureTranscriptsMigrated` move + no-op, and an `initTranscript` default-resolution regression asserting transcripts land in `state/transcripts/` (not the legacy flat dir).

## Provenance

Claims behind this change (transcript path resolution, `getAfkStateDir` ignoring the env var, sole consumer, missing helper, migration helpers present) were independently re-derived by a parallel shadow-verify wave before implementation — 5/5 CONFIRMED.


griffinwork40:
Here is the synthesized review.

---

## PR #143 Review — `fix(paths): move transcripts into state/ and honor AFK_STATE_DIR`

**Branch:** `afk/state-folder-transcripts → main` | **Stats:** +139/−22, 9 files

---

### Summary

The structural work is sound: `getTranscriptsDir()` + `ensureTranscriptsMigrated()` are cleanly introduced, docs are consistent across four files, and the core migration tests cover happy path + no-op. Three items need resolution before merge; two more are worth addressing in a follow-up.

---

### Findings by severity

#### 🔴 HIGH

**[H1] `AFK_STATE_DIR` scope silently expanded from transcript-only to all state-tier paths**
*(api-compat)*

Before this PR, `AFK_STATE_DIR` was consumed by `transcript.ts` alone:
```ts
// Old transcript.ts
const dir = path.join(env.AFK_STATE_DIR ?? join(homedir(), '.afk'), 'transcripts');
```
`getAfkStateDir()` previously ignored it and always returned `join(getAfkHome(), 'state')`. Sessions, todos, daemon state, REPL history, MCP server status, and ~15 other state-tier paths were unaffected.

After this PR, `getAfkStateDir()` reads `AFK_STATE_DIR` and redirects the **entire** state tier. Any user who had `AFK_STATE_DIR=/x` set to control transcript placement will now find sessions, daemon lock files, MCP status, and REPL history silently redirected to `/x/sessions`, `/x/daemon/`, etc.

This may be the intended design (the PR title says "honor AFK_STATE_DIR"), but it's a behavioral scope expansion that needs a changelog note, a `--dump-prompt` mention, or at minimum a comment in the diff explaining the intentional scope widening. If it's intentional, no code change needed — only a docs update (the docs updated in this PR only mention `transcripts`, not the full state tier).

**[H2] Silent migration — users don't know transcripts moved**
*(perf-observability)*

`migrateDirOnce` is entirely silent on success. A pre-3.x user launching the REPL will find `~/.afk/transcripts/` gone and no log explaining what happened. One write to stderr fires at most once per installation lifetime:

```ts
// after renameSync succeeds:
process.stderr.write(`[afk] transcripts migrated: ${oldPath} → ${newPath}\n`);
```

Also applies to the `cpSync` cross-device fallback, which blocks the event loop silently — add a pre-copy line too.

**[H3] Errors in `migrateDirOnce` bubble uncaught into REPL bootstrap**
*(correctness)*

`EPERM`, `EXDEV` (caught by fallback but the fallback itself can throw), and `ENOTEMPTY` propagate out of `ensureTranscriptsMigrated()` → into `initTranscript()` → into REPL session startup as an unhandled rejection. On a cross-filesystem `AFK_STATE_DIR` on first run, the REPL aborts with a raw stack trace. Wrap in `try/catch` with a fallback log:

```ts
export function ensureTranscriptsMigrated(): void {
  try {
    migrateDirOnce(getLegacyTranscriptsDir(), getTranscriptsDir());
  } catch (err) {
    process.stderr.write(`[afk] transcript migration failed (non-fatal): ${err}\n`);
  }
}
```

---

#### 🟡 MEDIUM

**[M1] Missing test: `AFK_STATE_DIR` flowing through `initTranscript` to actual file path**
*(test-coverage)*

The new `transcript.test.ts` case confirms the no-override path (`handle.path()` starts at `~/.afk/state/transcripts`), but there is no test asserting that `AFK_STATE_DIR=/custom` causes `handle.path()` to start at `/custom/transcripts/`. The env → path → file linkage for the override case is untested end-to-end.

**[M2] `getAfkStateDir()` throw has no observability before it crashes**
*(perf-observability)*

When `AFK_STATE_DIR` is set to a relative or root path, `getAfkStateDir()` throws during `initTranscript`. The CLI's top-level catch shapes the output but there is no log line differentiating "override active and valid" from "about to throw on validation." A single `stderr.write` when the override is accepted would let users confirm it took effect.

**[M3] `getTranscriptsDir()` docstring doesn't mention migration contract**
*(api-compat)*

The exported getter has a good JSDoc, but it doesn't mention that callers are responsible for calling `ensureTranscriptsMigrated()` before use. Any future caller who uses the getter directly (e.g. a list-transcripts command) will silently miss migration. Either call `ensureTranscriptsMigrated()` inside `getTranscriptsDir()` (idempotent, so safe), or document the contract explicitly.

---

#### 🟢 LOW / INFO

**[L1] Missing test: empty string `AFK_STATE_DIR` treated as unset** — the implementation handles it (`!envVal` branch), but no test confirms it. Low risk, easy addition.

**[L2] Migration no-op comment misleading for custom-`AFK_STATE_DIR` users** — for a user who had `AFK_STATE_DIR=/x`, migration no-ops because the legacy dir doesn't exist (not because old === new). The `ensureTranscriptsMigrated` docstring says "when an `$AFK_STATE_DIR` override already places both at the same path" — technically inaccurate; no data loss, just a misleading comment.

**[L3] `mkdirSync` without explicit mode on transcript dir** — pre-existing across all state-tier dirs. Not introduced by this PR; a separate hardening pass adding `mode: 0o700` would close it. Not a blocker.

**[I1] `env.AFK_STATE_DIR` is a live `process.env` getter** — confirmed correct, not a frozen snapshot. The adjacent calls in `initTranscript` share a negligible race window. No action needed.

**[I2] No regression for existing `AFK_STATE_DIR` users on transcript path** — confirmed: old and new both resolve to `$AFK_STATE_DIR/transcripts`. Migration correctly no-ops via `existsSync` guard. ✓

---

### Merge recommendation

**Request changes.** H1–H3 should be resolved before merge:

| # | Action |
|---|--------|
| H1 | Add a changelog / doc note that `AFK_STATE_DIR` now governs the full state tier, or explicitly scope it to only transcripts (revert the `getAfkStateDir` change and keep transcript-only override in `transcript.ts`). Confirm intent. |
| H2 | Add one `stderr.write` on migration success (and before `cpSync` fallback). |
| H3 | Wrap `ensureTranscriptsMigrated()` body in `try/catch` with non-fatal log. |

M1 is the only medium with a code-level fix; M2–M3 are low-effort additions worth including in the same pass. The PR is otherwise well-structured and the migration tests are a genuine quality addition.

---

**Done.** Review complete. Findings: 3 HIGH (changes requested), 3 MEDIUM, 3 LOW/INFO. No security blockers. The core architecture of `getTranscriptsDir` + `ensureTranscriptsMigrated` is sound; the three HIGH items are about silent failure modes and an undocumented scope change, all fixable without structural rework.

---

griffinwork40:
Both load-bearing claims are now **verified at ref 8f50059**:

- **Finding #2 confirmed** — `migrateDirOnce` swallows every non-EXDEV `renameSync` error (the `// Any other error is swallowed` comment is literal; only EXDEV gets the copy+remove fallback + stderr warning).
- **Finding #1 confirmed** — all three `ensure*Migrated` helpers migrate only from legacy **flat** paths (`$AFK_HOME/{sessions,todos,transcripts}`), never from `$AFK_HOME/state/*` → `$AFK_STATE_DIR/*`. The migration gap for pre-existing `AFK_STATE_DIR` users is real.
- **transcript.ts confirmed** — `path` still used (line 15), `os`/`env` imports removed cleanly; no `noUnusedLocals` break.

No critical/high findings → `/shadow-verify` correctly not triggered. Synthesizing.

---

# Code Review — PR #143

**Target:** `fix(paths): move transcripts into state/ and honor AFK_STATE_DIR` · `afk/state-folder-transcripts` → `main`
**Reviewed ref:** `8f5005946d6af81292ed80bb7e8b39224bc91d57` · **Regime:** light (161 lines, 9 files, refactor) · **Waves:** 1 light-review + inline citation/absence verification

**Summary:** Relocates autosaved REPL transcripts from the legacy flat `~/.afk/transcripts/` into the state tier `~/.afk/state/transcripts/`, adds `getTranscriptsDir()` + `ensureTranscriptsMigrated()`, and makes `getAfkStateDir()` honor the (already-defined, previously-only-partly-wired) `AFK_STATE_DIR` env var. Clean, well-tested, imports verified, no compile risk.

**1 · medium · confidence high · blast-radius · `src/paths.ts:168–183` · ref:8f50059 · file-state**
Making `getAfkStateDir()` honor `AFK_STATE_DIR` cascades to ~13 derived state paths (sessions, todos, presence, memory, session-grants, witness, daemon, repl-history, bg-jobs, mcp/server-status, telegram, last-cwd). Pre-PR, only `transcript.ts` read `AFK_STATE_DIR`; everything else used `$AFK_HOME/state`. A user who set `AFK_STATE_DIR` **before** this PR upgrades into a state where sessions/todos/grants/OAuth-tokens at `$AFK_HOME/state/*` become invisible — the app now reads `$AFK_STATE_DIR/*`. No `ensure*Migrated` helper bridges `$AFK_HOME/state/*` → `$AFK_STATE_DIR/*` (all three migrate only from legacy *flat* paths — verified).
*Evidence:*
```
export function ensureSessionsMigrated(): void { migrateDirOnce(getLegacySessionsDir(), getSessionsDir()); }   // getLegacySessionsDir = $AFK_HOME/sessions, NOT $AFK_HOME/state/sessions
```
*Mitigating:* no data loss (old data stays on disk, recoverable), narrow affected population (the env var was a near-no-op before, so few set it), and the change aligns the impl with the env var's documented contract. *Suggestion:* add a CHANGELOG/release note calling out that `AFK_STATE_DIR` now governs the whole tier and pre-existing users must move `$AFK_HOME/state/*` — or add an `ensureStateDirsMigrated()` that relocates sub-dirs when `AFK_STATE_DIR` differs from `$AFK_HOME/state`.

**2 · low · confidence high · migration-correctness · `src/paths.ts:344–372` · ref:8f50059 · file-state**
`migrateDirOnce` silently swallows every non-EXDEV `renameSync` error. If `AFK_STATE_DIR` is set to `$AFK_HOME/transcripts`, the new path nests under the old (`…/transcripts/transcripts`); `renameSync` throws `EINVAL`, which is swallowed — legacy transcripts stay put, new ones write to the nested dir (silent split, no loss). Contrived but unguarded.
*Evidence:*
```
} catch (renameErr) {
  if ((renameErr as NodeJS.ErrnoException).code === 'EXDEV') { /* copy+remove + stderr warn */ }
  // Any other error is swallowed — migration remains best-effort.
}
```
*Suggestion:* emit a `process.stderr.write` warning on non-EXDEV failures (reuse the EXDEV-branch pattern) so a split state is diagnosable.

**3 · low · confidence high · test-coverage · `src/paths.test.ts` · ref:8f50059 · file-state**
No test exercises `ensureTranscriptsMigrated()` when `AFK_STATE_DIR` is set (migration + override interaction, including the nested-collision case in finding #2). Grep for `AFK_STATE_DIR.*transcripts` in the test file → zero matches. Coverage exists for unset-default, override-derivation, validation-rejection, and migrate/no-op-on-empty, but not migrate-under-override. *Suggestion:* add one case setting `AFK_STATE_DIR` then asserting legacy transcripts land under the override.

**nit · test-coverage · `src/paths.test.ts:89–95`** — the `honors AFK_STATE_DIR` test doubles as the sole regression guard for the finding-#1 blast radius; fine as-is, no action required.

## What was not checked
- **Ref read:** all citations verified against branch HEAD `8f50059` via `git show <ref>:<path>` (migrateDirOnce, getAfkStateDir, transcript.ts imports, env.ts AFK_STATE_DIR definition lines 525/1080, all getAfkStateDir callers).
- Did **not** run the test suite (`pnpm test`) or `tsc` — compile-safety inferred from verified imports, not executed.
- Docs (AFK.md, CLAUDE.md, docs/architecture.md, docs/reference.md) spot-checked against the diff for consistency with the code; not rendered.
- Did not trace runtime side-effects of telegram/daemon/trace/cd-on-exit under the new resolution — path derivation verified correct, but their startup behavior under a mid-life `AFK_STATE_DIR` change was not exercised.
- Windows path/rename semantics — out of scope for this codebase.

## Merge recommendation

**MERGE.** No critical/high findings. The diff does exactly what it claims, imports and types are sound, and the new behavior is well-covered by tests. The one medium item (finding #1) is a documented-intent correctness fix with a narrow, non-destructive, recoverable migration gap — recommend pairing the merge with a CHANGELOG/release note (or a follow-up state-dir migration helper) so any pre-existing `AFK_STATE_DIR` user knows to relocate `$AFK_HOME/state/*`. Findings #2/#3 are low-severity robustness/coverage nits suitable for the same or a follow-up PR.

---

griffinwork40:
🤖 _Automated review — agent-afk hourly sweep · updated 2026-06-15 11:08 UTC · re-evaluated each rotation._

✅ MERGE

  • getAfkStateDir() now honors AFK_STATE_DIR behind an absolute-path guard that rejects relative paths and / (src/paths.ts:166); transcripts move from the flat ~/.afk/transcripts/ into the state tier via the new getTranscriptsDir() (src/paths.ts:217).
  • Default upgrade path is safe: ensureTranscriptsMigrated() reuses the existing migrateDirOnce, which is collision-safe (if (existsSync(newPath)) return;) and EXDEV-safe (copy+remove fallback) — no clobber, no data loss.
  • Blast radius verified: the transcript resolver was the only direct env.AFK_STATE_DIR consumer (now removed); every getAfkStateDir()-derived caller (sessions/todos/memory/daemon) cascades correctly.
  • Non-blocking: setting AFK_STATE_DIR relocates the whole state tier but only transcripts auto-migrate — pre-existing sessions/todos/memory under $AFK_HOME/state are not moved (data intact, reversible; that path was never AFK_STATE_DIR-relative before). Worth a docs note. Minor: whitespace-only AFK_STATE_DIR throws a "not absolute" message.

PR #152, opened 2026-06-15
Ctrl+B backgrounds the running subagent; remove whole-turn detach

griffinwork40:
## What

Pressing **Ctrl+B** in the REPL while a foreground subagent is running now **backgrounds just that subagent** — it keeps running detached (joinable via `/bgsub:join`) and the **main turn keeps streaming**. When no subagent is running, Ctrl+B is a deliberate no-op.

Previously Ctrl+B detached the **whole turn**. That behavior — and the entire whole-turn-detach subsystem — is **removed**.

## Why

Backgrounding the running subagent is what users actually expect from Ctrl+B mid-investigation. The whole-turn detach (`/bg`, `/attach`, `/tasks`, `BackgroundTaskManager`) was a separate, rarely-used facility that overlapped confusingly with background *subagent* jobs (`/bgsub`).

## How

- **`SubagentControl` seam** on `SubagentExecutor` — one query (`hasPromotableForeground`) + one command (`promoteActiveForeground`). The keyboard/turn layer depends only on this interface; a boundary test (`subagent-control-boundary.test.ts`) fails CI if anyone reaches into `SubagentHandleImpl`/manager internals.
- The foreground `runToResult` await **races a promotion signal**. On promote, the in-flight handle is handed to **`BackgroundAgentRegistry.adoptRunning()`**, which attaches the terminal callback to the *already-running* promise (calling `register()`/`runInBackground()` would re-enter `run()` and throw "already running"). `execute()` then returns the same synthetic `{status:'running', jobId}` pointer that `mode:'background'` returns, unblocking the parent turn.
- Composition-root injection: bootstrap passes the executor as `SubagentControl` through `InteractiveCtx` → `TurnHandles` → the Ctrl+B handler closure. No `AgentSession` plumbing.

## ⚠️ Breaking

Removed entirely: `BackgroundTaskManager`, `detachStreamToBackground`, and the **`/bg`, `/attach`, `/tasks`** slash commands. Background **subagent** jobs (`agent` tool `mode:'background'`) and the **`/bgsub`** family are unchanged; the footer status bar now surfaces only those.

Net: **+698 / −1769** across 32 files.

## Verification

- `tsc --noEmit` (strict) — clean.
- Full suite green **except 3 pre-existing failures unrelated to this change**, confirmed by reproducing them on a clean tree (`anthropic-direct compact()` model-alias expansion; 2× `AFK_SUGGEST_ENABLED`).
- KEEP-subsystem confirmed intact + green: `background-registry` (40/40), `/bgsub` (16/16).
- New tests: 5 promotion tests (executor), 2 Ctrl+B dispatch tests (turn-handler), 6 boundary tests.


griffinwork40:
All three confirmed. I now have everything needed to synthesize. The citation-verification work refuted the one decision-driving finding against primary source, and the remaining items are all low/nit with grep-grounded evidence. No critical/high findings survive, so per protocol medium-and-below go straight through (and the key claim was already independently re-derived from source — no separate shadow-verify wave needed).

---

# Code Review — PR #152

**Target:** `feat/ctrl-b-background-subagent` → `main` · reviewed ref `a43dad6` · +698/−1769 across 32 files
**Change type:** feature + subsystem-removal refactor · **Regime:** full
**Summary:** Removes the whole-turn-detach subsystem (`BackgroundTaskManager`, `/bg`, `/tasks`, `/attach`) and repurposes **Ctrl+B** to promote a *running foreground subagent* into a detached `/bgsub` job via a new narrow `SubagentControl` seam (`hasPromotableForeground()` + `promoteActiveForeground()`), backed by `BackgroundAgentRegistry.adoptRunning()`.

## Verification summary
- **Build integrity: clean.** Every deleted export (`BackgroundTaskManager`, `createBackgroundSink`, `detachStreamToBackground`, `bgCmd`/`tasksCmd`/`attachCmd`, `set*Manager`/`setTasksRegistry`/`resetTasksRefs`) has **zero** surviving importers in production code (`git grep` → 0 matches each).
- **api-compat: clean.** The two positional-signature changes — `runTurn` (dropped `bgManager`) and `BackgroundStatusBar` ctor (dropped `manager`) — each have exactly **one** production caller (`loop-iteration.ts:397`, `footer-subsystems.ts:114`), both updated in the diff.
- **Promotion concurrency: correct.** `runToResult` never rejects (`handle.ts` try/catch), so the un-`.catch`'d race branch is safe; `resolveJob` is an idempotent `Promise.resolve`; `adoptRunning`'s `.then/.catch` calls `markTerminal` exactly once. `mode:'background'` early-returns at executor.ts:739 before the promotion plumbing, so background jobs never register a promotion trigger.
- **Result-text invariant preserved:** `/bgsub` keeps the deleted `/tasks` invariant — listings show metadata only; result text surfaces solely via explicit `/bgsub:join` (`bgsub.ts:16`).

## Findings

### ⛔ REFUTED during verification (was the reviewer's headline claim)
**Claim:** *"Promoted subagents' grandchildren are aborted at end-of-turn because `childManager` is wired to `call.signal` and only the `handle.cancel` listener is detached on promotion."* (`subagent-executor.ts:577`)

**Refuted by source at `a43dad6`:** `call.signal` is a *fresh per-turn* controller from `AbortCoordinator.begin()`. On normal completion the turn calls `clear(controller)` — which **only nulls the slot, never `.abort()`s** (`abort-coordinator.ts:109‑113`). The signal fires **only** via `requestAbort()` on user-interrupt (`query.ts:448`) or session-close (`query.ts:661`), and only on the *currently-active* turn's controller. So grandchildren are **not** killed at normal end-of-turn. Moreover the `childManager`→`call.signal` wiring (executor.ts:571‑625) runs **before** the native `mode:'background'` return (line 693), so promotion inherits *identical* lifetime semantics to `mode:'background'` — exactly the PR's stated invariant. **No regression introduced; finding dropped.**

### Surviving findings — all non-blocking

| # | sev | dim | location | finding | suggestion |
|---|-----|-----|----------|---------|------------|
| 1 | low | test-coverage | `subagent-executor.test.ts` (promotion block) | The cap-hit-during-promotion path (`adoptRunning` throws `BackgroundJobCapError` → executor falls through to `await runPromise`, stays foreground) is untested. The native-background cap path *is* tested (5 `BackgroundJobCapError`/`maxConcurrentJobs` refs). | Add a promotion test with `maxConcurrentJobs:1` already occupied; assert empty `promoted[]`, normal completion, `teardown` called. |
| 2 | nit | docs | `slash/commands/bgsub.ts:6-7` | Docstring still describes `/bg` as "a separate, older facility that backgrounds the current main-session turn" — that facility was deleted in this PR; the collision-avoidance rationale is now vacuous. | Reword to past tense or drop the clause. |
| 3 | nit | comment | `repl-loop.ts:122` | Comment "Same lifecycle rationale as **bgManager above**" references the `bgManager` drain block this PR deleted (`git grep bgManager` → 1 match, this comment only). | Re-anchor to the shell-job/registry rationale. |
| 4 | nit | ux | `turn-handler.ts:~90` (`handleBackgroundKey`) | When `hasPromotableForeground()` is false, Ctrl+B is a fully silent no-op (was always-acts before). The confirmation line on the *success* path is correctly routed through `completionWriter` (not raw `console.log`), so the compositor is safe. | Optional: emit a dim hint ("no promotable subagent in flight") so the keypress reads as intentional, not broken. |
| 5 | nit | types | `cli/background/types.ts` | `BackgroundItem` is now a single-member discriminated union; the `kind:'subagent'` discriminant is vestigial. Harmless; compiles. | Optional: collapse to the bare `BackgroundJob` shape if no future `kind` is planned. |

No CONFLICT blocks. No critical/high/medium findings survive verification.

## Epistemic scope disclosure
- **Refs read:** all citations verified against branch HEAD `a43dad6` (== local working HEAD; files on disk match the reviewed ref). The load-bearing abort-lifecycle claim was independently re-derived through the full chain: `subagent-executor.ts` → `dispatcher.ts` → `loop.ts` → `query.ts` → `abort-coordinator.ts`.
- **Not checked:** did not run the test suite (`pnpm test`/`pnpm lint`) — relied on static `tsc`-strictness reasoning + grep for build integrity; did not exercise the `subagent-control-boundary.test.ts` enforcement at runtime; did not audit the Telegram/daemon surfaces (Ctrl+B is REPL-only); did not trace whether `cancelAll()`→`handle.cancel()` cascades to a promoted subagent's *grandchildren* on session close (orthogonal pre-existing concern, not introduced by this PR).

---

## ✅ MERGE

No blocking defects. Build integrity, api-compat, and the promotion concurrency logic are sound; the one decision-driving finding was refuted against source. The five surviving items are doc/comment drift, one vestigial type, and a single missing edge-case test — all sweepable in a quick follow-up and none gating. Recommend merge; optionally fold findings #1–#3 in before or right after.

---

griffinwork40:
🤖 _Automated review — agent-afk hourly sweep · updated 2026-06-15 17:10 UTC · re-evaluated each rotation._

✅ MERGE

  • Deletions are clean — all 6 importers of the removed modules (background.ts, attach.ts, bg.ts, tasks.ts) verified reference-free at the head ref, so no dangling imports / build break (footer-subsystems.ts, background-status-bar.ts, background/types.ts, slash index.ts, 2 test files).
  • Promotion path is sound (subagent-executor.ts): races runToResult vs a promotion signal, gates teardown on promoted in finally, and an idempotent resolveJob(null) prevents a hung promoteActiveForeground() await.
  • background-registry.adoptRunning() attaches the terminal callback to the in-flight promise (avoids the "already running" re-entry), preserves the BackgroundJobCapError cap, and is bounded by cancelAll() on parent-session teardown — mirroring native mode:'background'.
  • Tests cover no-registry, nothing-in-flight, happy promote, promote-all, and the run-wins-race teardown; a new boundary test locks the keyboard→SubagentControl seam.
  • Nit (non-blocking): SubagentControl JSDoc still says Ctrl+B "falls back to whole-turn backgrounding" — that path was removed; turn-handler.ts makes it a deliberate no-op.

PR #159, opened 2026-06-15
fix(provider): resolve model alias in compact summarizer

griffinwork40:
## Summary

The compact/summarize path passed `AFK_COMPACT_MODEL` (or the default) to the Anthropic Messages API **verbatim**. A short alias like `haiku` is rejected with `404 "model: haiku not_found"` under OAuth (see the invariant in `oneshot.ts`) — so **live compaction was broken for any operator whose `AFK_COMPACT_MODEL` is a short alias**.

Every other API path already resolves the alias via `resolveModelId` before `messages.create` (`query.ts`, `oneshot.ts`, `index.ts`); `readCompactModel()` was the sole path missing it.

**Fix:** resolve the configured model via `resolveModelId` in `readCompactModel()`. `resolveModelId` returns the full id for known aliases and passes anything else through unchanged. This is also what the existing `compact()` test asserts — the failure surfaced because an operator's `AFK_COMPACT_MODEL=haiku` leaks into the vitest process.

One-file change: `src/agent/providers/anthropic-direct/query/compact-handler.ts` (+9/-1).

## How was this tested?

- `pnpm lint` (tsc --noEmit) — **passes** (exit 0)
- `pnpm test -- src/agent/providers/anthropic-direct/compact.test.ts` — **17/17 pass**
- Full `pnpm test` / coverage gate runs in CI.

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes (targeted compact suite 17/17 locally; full suite via CI)
- [ ] Commits are signed off (`git commit -s`) per the [DCO](https://developercertificate.org/) — see [CONTRIBUTING.md](../CONTRIBUTING.md) — _commit was already pushed without `-s`; can amend + force-push if required_

### DIFF
--- a/src/agent/providers/anthropic-direct/query/compact-handler.ts
+++ b/src/agent/providers/anthropic-direct/query/compact-handler.ts
@@ -43,6 +43,7 @@ import {
   findCompactionBoundary,
 } from '../compact.js';
 import { emitCompaction } from '../../../trace/emit.js';
+import { resolveModelId } from '../../../session/model-resolution.js';
 import type { AnthropicClientLike } from '../types.js';
 import type { SessionState } from './session-state.js';
 import type { AbortCoordinator } from './abort-coordinator.js';
@@ -228,7 +229,14 @@ function readKeepLastN(): number {
 
 function readCompactModel(): string {
   const raw = env.AFK_COMPACT_MODEL;
-  if (raw !== undefined && raw.length > 0) return raw;
+  // Invariant: the Anthropic Messages API rejects short aliases like `'haiku'`
+  // (404 `model: <alias> not_found` under OAuth — see the note in oneshot.ts).
+  // Every other API path resolves the alias to a full model id via
+  // resolveModelId before messages.create (query.ts, oneshot.ts, index.ts);
+  // the compact summarizer must do the same, or a configured AFK_COMPACT_MODEL
+  // alias reaches the API raw and 404s mid-compaction. resolveModelId returns
+  // the full id for known aliases and passes anything else through unchanged.
+  if (raw !== undefined && raw.length > 0) return resolveModelId(raw) ?? raw;
   return DEFAULT_COMPACT_MODEL;
 }
 


griffinwork40:
Verified end-to-end: line 120 `const compactModel = readCompactModel();` → line 121-125 passed as the model arg to `buildSummarizationRequest(...)` → line 148-149 `client.messages.create(params, ...)`. The resolved id reaches the API. The fix is on the live path.

I executed the review inline rather than forking a Wave-1 agent: a 27-line single-function diff is trivial, and every claim was verifiable with direct reads against the reviewed ref (my working tree IS head `45474da`, and `compact-handler.ts` is byte-identical to it). Dispatching a sub-agent to re-read 27 lines would be ceremony, not rigor. Only finding is `low` severity → no `/shadow-verify` wave required (gated on critical/high).

---

# Code Review — PR #159

**Target:** PR #159 `fix(provider): resolve model alias in compact summarizer` · `fix/compact-model-alias-resolution` → `main` · **state: MERGED**
**Reviewed ref:** `45474da3f2a04ff20fff5c8e33871f6fbcb968b6`
**Triage:** change type = hotfix · 1 file · +9/-1 · regime = **light**
**Summary:** `readCompactModel()` now passes `AFK_COMPACT_MODEL` through `resolveModelId(raw) ?? raw` before the value reaches `messages.create`, so a configured short alias (e.g. `haiku`) is expanded to a full model id instead of 404-ing mid-compaction under OAuth.

## Findings

### 🟡 low · confidence high · test-coverage
`compact-handler.ts:239` · ref:`45474da` · citation-type: diff-context
**Finding:** The new alias-resolution branch ships without a regression test — no test sets `AFK_COMPACT_MODEL` to an alias and asserts the resolved full id reaches `messages.create`.
**Evidence:**
```ts
if (raw !== undefined && raw.length > 0) return resolveModelId(raw) ?? raw;
```
**Absence grounded:** `grep -rn "readCompactModel\|AFK_COMPACT_MODEL\|compact-handler" src --include=*.test.ts` → `AFK_COMPACT_MODEL` appears only in `src/cli/input/suggest.test.ts` (a separate CLI-suggestion path); no test references `readCompactModel` or the compaction model param. → confirmed-absent.
**Suggestion:** In `compact.test.ts`, set `AFK_COMPACT_MODEL='haiku'`, run the compaction path with a mocked client, and assert `params.model` equals the resolved full id.

## Verified clean

- **correctness** — `resolveModelInput` (`model-slots.ts:269`) returns `resolveBinding(input).id` for any defined input (unknown names pass through unchanged), so `resolveModelId(raw) ?? raw` is correct and the `?? raw` is a safe-but-effectively-unreachable fallback. Output chain confirmed: `readCompactModel()` (L120) → `buildSummarizationRequest(…, compactModel, …)` (L121) → `messages.create(params)` (L149). Fix is on the live path. No issues found.
- **api-compat** — `readCompactModel` is private (non-exported); signature unchanged (returns `string`). No exported symbol altered → no breaking change. No issues found.
- **security** — no auth/secret/injection surface; model id is read from operator env and used as an API param. No issues found.
- **type-safety** — `AgentModelInput = ClaudeModel | string` (`model-types.ts:21`); plain `string raw` is assignable. Passes strict `tsc`. No issues found.
- **perf** — `resolveModelId` is a single map lookup invoked once per compaction. No issues found.
- **consistency** — the comment's call-site claim is accurate: `oneshot.ts:78`, `index.ts:537`, `query.ts:458` all use the identical `resolveModelId(x) ?? x` pattern before `messages.create`. This change brings the compact summarizer into line with them. The `// Invariant:` comment correctly follows the repo's long-comment-prefix convention.

## What was not checked

- Wave-1 agents: none dispatched — review run inline given the 27-line single-function scope. All citations read directly from reviewed ref `45474da` (working tree byte-identical to head; `git diff 45474da -- compact-handler.ts` empty). Citations verified at that ref.
- Did not run the test suite or build — relied on static read + the fact that CI gated the merge.
- **Edge case (unverified, out of scope):** behavior when `AFK_COMPACT_MODEL=auto`. `resolveBinding("auto")` was not traced; before this change `"auto"` would also pass through raw, so no regression is introduced either way — flagging only for completeness, not as a finding.

A minimal, type-safe, correctly-targeted hotfix that closes a real 404-mid-compaction path and aligns the compact summarizer with three sibling call sites. The sole gap is a missing regression test (low) — worth a follow-up, not a blocker. (PR is already merged; this is a confirming post-merge review.)

---

**Done** — Reviewed PR #159 read-only against reviewed ref `45474da`. Resolved the bare PR number → diff via the prepared artifact, grounded every load-bearing claim (resolver behavior, call-site consistency, type safety, live-path wiring, test absence) with citations at the reviewed ref. One `low` finding (missing regression test); no critical/high, so no shadow-verify wave. Recommendation: **MERGE**. No files, git state, or the PR were mutated.

---

PR #163, opened 2026-06-15
fix(renderer): route all overlay repaints through setComposedOverlay (port afk-workshop#572)

griffinwork40:
## Source

Port of **afk-workshop#572** — `fix(renderer): route all overlay repaints through setComposedOverlay`.
Moat-scrubbed port via **diff + 3-way apply** (the public repo shares no git history with afk-workshop), **not** a 1:1 commit copy. Public tracker for this bug: #110.

## Problem

A prior change grew the composed orchestrator overlay to ~7 rows (thinking paragraph + stage rail + tool lane + progress banner). Every direct `compositor.setOverlay(toolLane.getOverlay())` callsite was a flicker source: it wiped the live-thinking paragraph rows, then the next composed frame repainted them — visible paragraph flicker on every subagent state transition.

## What changed

- **`lastProgressByTask`** hoisted from a separate function parameter into `OrchestratorCtx` (progress state is orchestrator-owned).
- **`setComposedOverlay`** signature simplified `(ctx, lastProgressByTask)` → `(ctx: OrchestratorCtx)`; now reads `ctx.lastProgressByTask`.
- **`SubagentCtx`** drops `overlayComposer`, gains `orchestratorCtx?: OrchestratorCtx`; every subagent overlay repaint (tool start/result/diff, throttled content, thinking, error, done/finalize) now routes through `setComposedOverlay(ctx.orchestratorCtx)` instead of a bare tool-lane write — so the orchestrator's thinking paragraph + stage rail + progress banner survive in every frame.
- **`emitSubagentPanel` / `finalizeSubagent`** moved from `stream-renderer-subagent-helpers.ts` → `stream-renderer-subagent.ts`; `finalizeSubagent` drops its two throttle-map params (uses module-scope maps).
- **`StreamRenderer.buildOrchestratorCtx()`** new private factory; threads the live ctx into both the orchestrator and subagent branches.
- Tests: new `stream-renderer-contexts.test.ts` + regression tests asserting subagent transitions do not drop the orchestrator thinking paragraph.

## Dropped for moat

**None.** Entirely public-safe — 12 files, all under `src/cli/_lib/stream-renderer*` (TUI rendering). No moat files, hunks, imports, or denylist terms were present, so nothing was scrubbed.

## Adaptations for public (not 1:1 — reviewer please note)

The public tree had drifted from the private base; rejects were resolved by hand, **preserving public-specific logic**:

- **`stream-renderer-orchestrator.ts` progress arm:** public carries a public-only fix — `lastProgressByTask.clear()` + an invariant comment that evicts a stale entry on auth-retry replay to prevent stacked "Tool-use loop" banners. Preserved it while applying the hoist (`ctx.lastProgressByTask.clear()/.set()`, dropped the 2nd arg).
- **`stream-renderer-subagent.ts` content + thinking throttle arms:** ported the `overlayComposer.markDirty/flush` → `setComposedOverlay(ctx.orchestratorCtx)` conversion onto public's (more verbose) comment text.
- **`stream-renderer-subagent-helpers.ts`:** deleted `finalizeSubagent` (moved to subagent.ts) despite a comment line-wrap drift.
- **3 public-only tests** in `stream-renderer-orchestrator.test.ts` used the old 4-arg `handleOrchestratorEvent(..., lastProgress)` signature; forward-fixed to the hoisted `ctx.lastProgressByTask` API (one was a live assertion against the now-disconnected external map; two were dead 4th args).
- **Issue reference:** the source title's `(closes #389)` points at a *private* issue; dropped from the commit subject to avoid a dangling/auto-close link on public (public #389 does not exist). Public tracker is #110.

## Verification (clean clone of this branch)

- `pnpm lint` (tsc --noEmit, strict): **clean**
- `pnpm fix:pins:check`: **clean**
- Affected suite (the 7 changed test files): **134/134 pass**
- Full suite excluding the pre-existing parallel-worker flake `stream-renderer-ordering.test.ts`: **9292 pass / 0 fail / 12 skipped (495 files)**. That flake is untouched by this PR and passes in isolation (14/14) — it hangs only under parallel workers (a pre-existing repo issue, documented on the source PR).
- `pnpm build` + `pnpm build:dist`: **clean** (dist scrubber: 0 files scrubbed)
- Deterministic moat guard (tree + `dist/`, HARD denylist + structural): **MOAT GUARD CLEAN**
- Independent adversarial moat audit (re-derived, not trusting the guard): **AUDIT CLEAN**

### DIFF
--- a/src/cli/_lib/rhythm-contract.test.ts
+++ b/src/cli/_lib/rhythm-contract.test.ts
@@ -77,6 +77,7 @@ function makeCtx(
     thinkingLane: new ThinkingLane(),
     thinkingMode: 'off',
     streamingMarkdown: { current: null },
+    lastProgressByTask: new Map(),
   };
 }
 
@@ -172,6 +173,7 @@ describe('TUI rhythm contract — flushToolLaneToScrollback', () => {
       thinkingLane: new ThinkingLane(),
       thinkingMode: 'off',
       streamingMarkdown: { current: null },
+      lastProgressByTask: new Map(),
     };
 
     flushToolLaneToScrollback(ctx);
@@ -213,7 +215,7 @@ describe('TUI rhythm contract — done-time tool-lane flush', () => {
 
     // Drive the done event — this path takes the orchestrator branch
     // at lines 357-379 (tool-lane flush + before-content anchor).
-    handleOrchestratorEvent({ type: 'done', metadata: { durationMs: 1 } }, source, ctx, new Map());
+    handleOrchestratorEvent({ type: 'done', metadata: { durationMs: 1 } }, source, ctx);
 
     // The done-path schedules commits through coordinator OR runs the
     // fallback path. In the fallback (no coordinator on ctx), it calls
@@ -253,7 +255,6 @@ describe('TUI rhythm contract — orchestrator before-content path', () => {
       { type: 'chunk', chunk: { type: 'content', content: 'prose after tool' } },
       source,
       ctx,
-      new Map(),
     );
 
     // The tool flush must have produced at least tool content + a trailing blank.
--- a/src/cli/_lib/stream-renderer-contexts.test.ts
+++ b/src/cli/_lib/stream-renderer-contexts.test.ts
@@ -1,0 +1,170 @@
+/**
+ * Production-wiring regression tests for the StreamRenderer ctx factories.
+ *
+ * Issue #389 shipped with every subagent overlay repaint routed through
+ * `setComposedOverlay(ctx.orchestratorCtx)`, but `makeSubagentCtx` — the SOLE
+ * production constructor of SubagentCtx, called by StreamRenderer.process() —
+ * never set `orchestratorCtx`. The `ctx.orchestratorCtx` guard in every
+ * subagent handler was therefore permanently false in production, so no
+ * subagent state transition ever repainted the overlay. The PR's own tests
+ * missed it because they hand-assembled a SubagentCtx with `orchestratorCtx`
+ * wired in — exercising a path production never takes.
+ *
+ * These tests build the ctx through the REAL factories (`makeSubagentCtx` +
+ * `makeOrchestratorCtx`), the exact path production uses, and assert the
+ * repaint actually fires. They fail on the pre-fix code.
+ */
+
+import { describe, it, expect, vi } from 'vitest';
+import { makeSubagentCtx, makeOrchestratorCtx } from './stream-renderer-contexts.js';
+import { handleSubagentEvent, synthesizeAgentEntry } from './stream-renderer-subagent.js';
+import { freshSourceState, type SourceState } from './stream-renderer-source.js';
+import { ToolLane } from '../commands/interactive/tool-lane.js';
+import { ThinkingLane } from '../commands/interactive/thinking-lane.js';
+import { CommitCoordinator } from './commit-coordinator.js';
+import { stripAnsi } from '../display.js';
+import type { Writer } from '../slash/types.js';
+import type { StreamingMarkdownRenderer } from './stream-renderer.js';
+import type { TerminalCompositor } from '../terminal-compositor.js';
+import type { OutputEvent } from '../../agent/types.js';
+
+function makeWriter(): Writer {
+  return { line() {}, raw() {}, success() {}, info() {}, warn() {}, error() {} };
+}
+
+function makeCompositor() {
+  const overlayCalls: string[] = [];
+  const compositor = {
+    setOverlay: (text: string) => { overlayCalls.push(text); },
+    commitAbove: vi.fn(),
+  } as unknown as TerminalCompositor;
+  return { compositor, overlayCalls };
+}
+
+describe('stream-renderer-contexts — issue #389 production wiring', () => {
+  it('makeSubagentCtx forwards orchestratorCtx onto the SubagentCtx', () => {
+    const { compositor } = makeCompositor();
+    const writer = makeWriter();
+    const toolLane = new ToolLane();
+    const orchestratorCtx = makeOrchestratorCtx({
+      out: writer,
+      isTTY: true,
+      compositor,
+      overlayComposer: null,
+      toolLane,
+      thinkingLane: new ThinkingLane(),
+      thinkingMode: 'off',
+      streamingMarkdown: { current: null },
+      coordinator: new CommitCoordinator(),
+      lastProgressByTask: new Map(),
+    });
+
+    const ctx = makeSubagentCtx({
+      isTTY: true,
+      compositor,
+      toolLane,
+      out: writer,
+      streamingMarkdown: new Map<string, StreamingMarkdownRenderer>(),
+      thinkingMode: 'off',
+      orchestratorCtx,
+    });
+
+    // The factory MUST thread orchestratorCtx through — dropping it (pre-fix)
+    // is exactly what neutered every subagent repaint in production.
+    expect(ctx.orchestratorCtx).toBe(orchestratorCtx);
+  });
+
+  it('a subagent tool event fires a composed overlay repaint that preserves the orchestrator thinking paragraph', () => {
+    const { compositor, overlayCalls } = makeCompositor();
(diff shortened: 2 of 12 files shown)


griffinwork40:
Wave 1.5 verification complete — all load-bearing claims checked against ref `25f340da`:

- **A1/A2/B ✓** Legacy `setComposedOverlay` path uses `peek()` (orchestrator.ts:403); the *production* overlay slot (lifecycle.ts:88,98) correctly uses `peekPhase()`+`isActive()`; `arm()` always creates `overlayComposer` on TTY (stream-renderer.ts:428) → **the legacy path is test-only reachable**. The reviewer's `high` is therefore downgraded to `low` per rubric (high requires production reachability).
- **C ✓** Interface declares `orchestratorCtx?` (subagent.ts:78) vs factory requires it (contexts.ts:72) — asymmetry real.
- **D ✓** Zero `onStageChange` in stream-renderer.test.ts (only a comment reference) — coverage gap confirmed.
- **E ✓** New `stream-renderer-contexts.test.ts` imports the **real** factories (line 19), builds ctx through them, and asserts `overlayCalls > 0` + `◆ thinking` present — it genuinely catches the #389 regression.

No critical/high findings survive verification → `/shadow-verify` not required (top claims already independently re-derived above). Synthesizing.

---

# Code Review — PR #163

**Target:** `fix(renderer): route all overlay repaints through setComposedOverlay (port afk-workshop#572)` · ref `25f340dacc0e44e1005d18722660ab2762f7fe1b` · base `044616b`
**Regime:** full (12 files, +809/-346) · change type: refactor+fix
**Verdict:** no blocking defects. 6 non-blocking findings (3 medium, 2 low, 1 nit).

## What the PR does (verified)
Closes the latent #389 bug: `makeSubagentCtx` never set `orchestratorCtx`, so the `ctx.orchestratorCtx` guard in every subagent handler was permanently false in production → **no subagent state transition ever repainted the overlay**. The fix makes `orchestratorCtx` required in the factory (sole production constructor) and threads `buildOrchestratorCtx()` into all three callsites, so subagent repaints route through `setComposedOverlay(ctx.orchestratorCtx)` and preserve the orchestrator's thinking paragraph + progress banner + stage rail.

**[1] medium · high conf · api-compat · stream-renderer-subagent.ts:78 vs stream-renderer-contexts.ts:72 · file-state**
`SubagentCtx.orchestratorCtx?` is **optional** in the interface but **required** in the `makeSubagentCtx` factory arg. A future caller constructing a `SubagentCtx` literal directly (bypassing the factory) compiles cleanly while omitting it, silently recreating the exact #389 class of bug.
Evidence: `orchestratorCtx?: OrchestratorCtx;` (subagent.ts:78) vs `orchestratorCtx: OrchestratorCtx;` (contexts.ts:72, no `?`).
Not a current defect — grep confirms the factory is the only production constructor. Hardening, not a live bug.
Fix: make the interface field required and update the test stubs that omit it, or add a runtime assert in the factory.

**[2] medium · high conf · correctness · stream-renderer-orchestrator.ts:414 · file-state**
Legacy (non-composer) `setComposedOverlay` path now calls `ctx.compositor.setOverlay(parts.join('\n'))` **unconditionally** (was guarded by `if (parts.length > 0)`). An empty frame now emits `setOverlay('')`, actively clearing the overlay rather than leaving prior content.
Evidence: `ctx.compositor.setOverlay(parts.join('\n')); // '' when parts is empty`.
Test-only reachable (production TTY hits the composer early-return at :380). Fix: restore the `parts.length > 0` guard, or keep the clear with an explicit comment if intentional.

**[3] medium · medium conf · perf-observability · stream-renderer-orchestrator.ts:380-383 · file-state**
Composer path calls `overlayComposer.invalidate()` (full recompose) + `flush()` on **every discrete subagent transition** (tool_use/result/diff/error/done — intentionally unthrottled). With N parallel subagents × K transitions this is N×K full recompositions per turn.
Evidence: `ctx.overlayComposer.invalidate(); ctx.overlayComposer.flush();`.
Degrades gracefully — `flush()` is in-memory (~5 slot renders, no I/O), and the high-frequency content/thinking token path keeps its 1500ms throttle. Non-blocking; flag for profiling under sustained parallel-subagent load.

**[4] low · high conf · correctness · stream-renderer-orchestrator.ts:403 · file-state**
*(Reviewer flagged high; downgraded — see verification above: production-unreachable.)* Legacy path switched `peekPhase()`+`isActive()` → `peek()` with no `isActive()` guard, inconsistent with the production lifecycle slot (stream-renderer-lifecycle.ts:88,98) which correctly uses `peekPhase()`+`isActive()`. After a phase seals, `peek()` re-shows already-committed reasoning.
Evidence: `formatThinkingParagraph(ctx.thinkingLane.peek(), …)` vs lifecycle's `!ctx.thinkingLane.isActive() || … peekPhase()`.
Test-only reachable. Fix: restore `peekPhase()`+`isActive()` for parity with the live slot.

**[5] low · high conf · test-coverage · stream-renderer.test.ts (absence) · file-state**
No integration test asserts `onStageChange` fires for **summary-mode** stage transitions — the exact path the thinking-repaint gate change (`thinkingMode === 'live'`) now relies on for stage-rail updates. Grep: zero `onStageChange` in stream-renderer.test.ts (only a comment in stream-renderer-orchestrator.test.ts:791).
The behavior is verified-correct by reading (`StreamRenderer.process` fires `onStageChange` on transition regardless of thinkingMode, stream-renderer.ts:563), but not test-locked. Fix: add a `process` test with `thinkingMode:'summary'` asserting the callback fires.

**[6] nit · correctness · stream-renderer.ts:493-516 · file-state**
`buildOrchestratorCtx()` builds a fresh ctx per event — safe today (shallow wrapper over shared live refs), but the invariant is undocumented; a future cached/derived field on `OrchestratorCtx` would break read-at-call-time correctness. Fix: document the "shallow wrapper over shared refs — no derived state" invariant.

## Positive confirmations (grounded at ref)
- **Core fix correct:** `orchestratorCtx` now required in the factory and threaded into all 3 `makeSubagentCtx` callsites (stream-renderer.ts:543,567 + synthesize path).
- **Regression test is real:** `stream-renderer-contexts.test.ts` uses the genuine factories and would fail on pre-fix code — it locks the #389 fix.
- **API-compat clean:** all signature changes (`setComposedOverlay` −1 param across 8 call sites; `makeSubagentCtx` shape; `makeOrchestratorCtx` +`lastProgressByTask`; `handleOrchestratorEvent` −1 param) have every caller updated. No broken consumers.
- **Symbol move clean:** `emitSubagentPanel`/`finalizeSubagent` moved helpers→subagent.ts with no stale importers; both still called from their handlers; module-scope throttle maps present → no build break.
- **`OverlayComposer.invalidate()` exists** (overlay-composer.ts:89).
- **No summary-mode stage-rail regression:** `onStageChange` repaints the footer regardless of `thinkingMode` (verified).

- Read against branch HEAD `25f340dacc0e44e1005d18722660ab2762f7fe1b`; all citations verified at that ref.
- **Did not run the test suite** — read-only review. Confirmed tests *exist and assert the right things* by reading source; did not confirm they *pass*.
- Did not benchmark the `invalidate()`/`flush()` perf claim ([3]) — assessed by reading `flush()`, not profiling; `CupFrameRenderer.render()` cost under load unmeasured.
- Non-TTY surfaces (Telegram/daemon) reviewed only to confirm they hit the `!ctx.compositor` early-return.

## Merge decision: **MERGE**
No critical or high findings. The fix is correct, regression-tested through the real production wiring, and API-compatible. Findings [1], [2], [4] are trivial, test-only-path hardening items worth folding into this PR or a quick follow-up to prevent future foot-guns, but none block merge.

## Done
Completed a full-regime, read-only review of PR #163 (2 parallel Wave-1 agents + inline Wave-1.5 citation/absence verification against ref `25f340da`). Produced: 6 non-blocking findings (3 medium / 2 low / 1 nit), 6 grounded positive confirmations, and a **MERGE** recommendation. One reviewer `high` was downgraded to `low` after verifying the affected path is test-only reachable. No files, commits, or PR state were modified.

---

PR #164, opened 2026-06-15
fix(providers): strip ask_question on non-interactive surfaces

griffinwork40:
## What

Adds `AgentConfig.isNonInteractive` and makes both providers strip the
`ask_question` tool from the assembled toolset when it is set. Wires it on the
three headless top-level surfaces: one-shot `afk chat`, the daemon session
factory, and the scheduler fallback config.

This mirrors the existing structural `ask_question` strip for skill-dispatch
sub-agents (`isSkillDispatch`) and rides the exact same rails — `AgentSession`
forwards `this.config` directly to `provider.query({ config })`
(`agent-session.ts:199,211`), so no new plumbing is required.

## Why

Session telemetry surfaced 17 sessions ending in `tool_error:ask_question:timeout`.
The deeper finding: the friction guards for this (the advisory `/ask-gate` skill,
plus prose in the routing directive) are *opt-in* — they only help if the model
chooses to invoke/heed them, so they under-fire exactly under task momentum. The
runtime already has the deterministic primitive (the provider tool-strip used for
skill-dispatch); this PR extends it to the case where the surface itself
guarantees no human can answer.

## Important nuance (what this does and does not fix)

Recon showed these headless surfaces install **no** elicitation handler
(`elicitation-router.ts:86-88`), so `ask_question` there already **auto-declines**
— it does not hang. So this change is best understood as:

- **Hygiene + deterministic contract:** don't offer a tool that structurally
  cannot succeed; stop the model burning a turn calling it and handling a decline.
- **Infrastructure:** establishes the `isNonInteractive` signal that future work
  can build on.

It is **not** a fix for the timeout source itself. Those 17 timeouts almost
certainly originate on **interactive** surfaces (REPL/Telegram) where a human is
genuinely away-from-keyboard, and possibly from foreground sub-agents that inherit
the parent's module-scope elicitation handler.

## Deliberately out of scope (with rationale)

- **Interactive surfaces (REPL/Telegram):** left untouched. A human *can* answer
  there (just slowly). Stripping `ask_question` would remove a legitimate
  capability and change core interactive behavior — a product-policy call, not a
  bug fix.
- **Foreground sub-agent elicitation:** `subagent.ts:414-419` has a deliberate
  `denyElicitations`/`DENY_ELICITATION` mechanism that auto-declines only for
  *background* sub-agents, intentionally leaving foreground forks able to elicit.
  Blanket-stripping all sub-agents would silently override that maintainer
  decision, so this PR does not touch `subagent.ts`.

## Scope

- Narrow strip: only `ask_question` is removed. `terminal_font_size` is **retained**
  (its skill-dispatch-specific numeric-arg lure does not apply here).
- Surfaces wired: `afk chat` (one-shot), daemon factory (production chokepoint the
  scheduler routes through), scheduler fallback config (standalone/tests).

## Test plan

- `pnpm test`: **495 files, 9302 passed, 12 skipped, 0 failed.**
- New tests (both providers): a non-interactive session strips `ask_question` while
  retaining `terminal_font_size`, `read_file`, and `bash` — asserting the strip is
  precise, not a blanket removal.

## Possible follow-ups (not in this PR)

- Decide policy for `ask_question` on interactive surfaces under sustained AFK
  (turn-timeout handling vs. strip).
- Revisit foreground sub-agent elicitation (the `denyElicitations` foreground/
  background split).
- Consider other headless entry points (e.g. threads/farm) if they prove headless.

### DIFF
--- a/src/agent/daemon/scheduler.ts
+++ b/src/agent/daemon/scheduler.ts
@@ -441,6 +441,13 @@ export class CronScheduler {
       // break scheduled tasks that depend on tool execution.
       permissionMode: 'bypassPermissions',
       hookRegistry: registry,
+      // Scheduler/cron sessions are headless (no human at the keyboard): strip
+      // ask_question so a scheduled task never stalls on an unanswerable prompt.
+      // The daemon session factory also forces this after its own config spread;
+      // set it here as the base for the no-factory fallback (standalone
+      // scheduler / tests). Placed before the sessionConfig spread so an
+      // operator escape-hatch could still override it, mirroring permissionMode.
+      isNonInteractive: true,
       // Trace writer placed before sessionConfig so an operator-supplied
       // sessionConfig.traceWriter still wins (escape-hatch parity with
       // permissionMode).
--- a/src/agent/providers/anthropic-direct/index.ts
+++ b/src/agent/providers/anthropic-direct/index.ts
@@ -651,11 +651,18 @@ export class AnthropicDirectProvider implements ModelProvider {
     // confused model into calling terminal_font_size(<n>) instead of running the
     // skill). Gated on isSkillDispatch; pairs with the SLASH_COMMAND_ROUTING_PROMPT
     // omission below. Verified safe: no bundled/registry/user skill calls either tool.
+    // Non-interactive surfaces (daemon, scheduler/cron, one-shot `afk chat`)
+    // install no elicitation handler, so `ask_question` can only auto-decline
+    // (elicitation-router.ts). Strip it so the model proceeds on an assumption
+    // or emits Blocked rather than burning a turn on an unanswerable prompt.
+    // Narrower than the skill-dispatch strip: `terminal_font_size` is retained.
     const toolDefs = config.isSkillDispatch
       ? baseToolDefs.filter(
           (t) => t.name !== 'ask_question' && t.name !== 'terminal_font_size',
         )
-      : baseToolDefs;
+      : config.isNonInteractive
+        ? baseToolDefs.filter((t) => t.name !== 'ask_question')
+        : baseToolDefs;
 
     // Build skill manifest for system prompt injection. The manifest lists
     // available skills so the model knows what the `skill` tool can invoke.
--- a/src/agent/providers/openai-compatible/index.ts
+++ b/src/agent/providers/openai-compatible/index.ts
@@ -204,6 +204,7 @@ export class OpenAICompatibleProvider implements ModelProvider {
           ...(config.traceWriter !== undefined ? { traceWriter: config.traceWriter } : {}),
           runtimeStateSource,
           ...(config.isSkillDispatch ? { isSkillDispatch: true } : {}),
+          ...(config.isNonInteractive ? { isNonInteractive: true } : {}),
         });
 
     const buildOpts: {
@@ -289,6 +290,13 @@ export class OpenAICompatibleProvider implements ModelProvider {
        * AnthropicDirectProvider.
        */
       isSkillDispatch?: boolean;
+      /**
+       * When true, this is a non-interactive surface (daemon, scheduler/cron,
+       * one-shot chat) where no human answers elicitations. Strip `ask_question`
+       * only (not `terminal_font_size`). Parity with the `config.isNonInteractive`
+       * toolDefs filter in AnthropicDirectProvider.
+       */
+      isNonInteractive?: boolean;
     },
   ): SessionToolDispatcher {
     const handlers = createBuiltinHandlers(permissionMode, opts.cwd);
@@ -325,7 +333,9 @@ export class OpenAICompatibleProvider implements ModelProvider {
       ? this.schemas.filter(
           (t) => t.name !== 'ask_question' && t.name !== 'terminal_font_size',
         )
-      : this.schemas;
+      : opts.isNonInteractive
+        ? this.schemas.filter((t) => t.name !== 'ask_question')
+        : this.schemas;
 
     const dispatcherOpts: ConstructorParameters<typeof SessionToolDispatcher>[0] = {
       handlers,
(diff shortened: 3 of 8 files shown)


griffinwork40:
Wave 1.5 verification is decisive and **refutes two of the agent's three test-coverage findings**:

- **Finding #1/#2 (missing negative case) → DROPPED as false-absent.** `skill-dispatch-routing.test.ts:264` (`'regular session: ask_question IS offered as a tool'` → `toContain('ask_question')`) and `query.test.ts:1244` (same) already assert the default/no-flag path retains `ask_question`. A "filter-always" regression would fail these. The negative case is covered.
- **Finding #3 (no wiring test) → CONFIRMED.** `daemon-session-factory.test.ts` has zero `isNonInteractive` references; only the two provider test files touch it. No test asserts the surfaces *set* the flag.

No critical/high findings emerged, so `/shadow-verify` is not triggered. Here is the synthesized review.

---

# Code Review — PR #164

**Target:** `fix(providers): strip ask_question on non-interactive surfaces` · `fix/strip-ask-question-noninteractive` → `main`
**Reviewed ref:** `7dd7b91e316c6d914c805317b5230f27112f3f3a` (citations verified at this SHA) · **regime:** light (hotfix, +152/−2, 8 files) · **PR state:** OPEN, MERGEABLE

**What it does:** Adds optional `isNonInteractive?: boolean` to `AgentConfig`. When set, both providers strip the `ask_question` tool from the model-visible toolset (retaining `terminal_font_size` — narrower than the `isSkillDispatch` strip). Wired at the three headless surfaces: one-shot `chat`, daemon factory, scheduler. Premise verified: `ask-question.ts:126` routes through `elicitationRouter.route()`, which auto-declines when no handler is installed (`elicitation-router.ts`) — so on these surfaces the tool can only waste a turn.

## Findings

**1 · low · confidence high · test-coverage · `src/cli/commands/daemon.ts:182`, `chat.ts:562`, `scheduler.ts:450` · ref:7dd7b91 · file-state**
No test asserts that the three surfaces actually *set* `isNonInteractive: true` — only provider behavior *given* the flag is tested. A refactor that drops the flag from a surface would silently reintroduce the bug this PR fixes, green.
Evidence: `git grep "isNonInteractive" -- '*.test.ts'` at ref → only `query.test.ts` + `skill-dispatch-routing.test.ts`; `daemon-session-factory.test.ts` → zero matches.
Suggestion: add an assertion in `daemon-session-factory.test.ts` that the config passed to `AgentSession` has `isNonInteractive === true` (the factory already captures the config; it's the lowest-cost wiring guard).

**2 · nit · confidence high · correctness/design · `src/agent/daemon/scheduler.ts:447–457` · ref:7dd7b91 · file-state**
Scheduler stamps `isNonInteractive: true` *before* the `...this.options.sessionConfig` spread (operator-overridable), while the daemon factory (`daemon.ts:182`) forces it *after* `...config` (non-overridable). The asymmetry is real but **intentional and documented** in both comments (escape-hatch parity with `permissionMode`); in production the daemon factory is the chokepoint and wins. Not a defect — flagged only for design awareness.
Evidence: `isNonInteractive: true, // ... Placed before the sessionConfig spread so an operator escape-hatch could still override it` … `...this.options.sessionConfig,`.

**3 · nit · confidence high · correctness · `src/agent/providers/openai-compatible/index.ts:196–197` · ref:7dd7b91 · file-state**
When `providerOpts.tools` is supplied (external-dispatcher branch), `isNonInteractive` is not forwarded to `wrapDispatcherWithRuntimeState`, so the strip in `buildDispatcher` is bypassed. **Pre-existing** (identical to the `isSkillDispatch` handling) and **not reachable** for the targeted surfaces (chat/daemon/scheduler don't pass `providerOpts.tools`). anthropic-direct avoids this because it filters `toolDefs` after the branch merge.

## Dropped / downgraded (Wave 1.5 manifest)
- `[DROPPED — false-absent]` "no negative case in skill-dispatch-routing.test.ts" → refuted by `skill-dispatch-routing.test.ts:264–279` (regular session asserts `toContain('ask_question')`).
- `[DROPPED — false-absent]` "no negative case in query.test.ts" → refuted by `query.test.ts:1244–1255`.
- `[DOWNGRADED medium→nit]` scheduler override semantics — intentional + documented, not wrong output under reachable conditions.

## Clean reads
- `config-types.ts:411–437` — field is optional + additive; no breaking change; thoroughly documented — read at ref.
- `anthropic-direct/index.ts:645–666` — ternary precedence correct (skill-dispatch superset wins over non-interactive); strip applies to model-visible `toolDefs` in both dispatcher branches — read at ref.
- `openai-compatible/index.ts:207, 333–338` — `config.isNonInteractive` → `opts.isNonInteractive` → `baseSchemas` filter is complete and correct — read at ref.
- `bootstrap.ts` (REPL), `telegram/construct-session.ts` (Telegram), `status.ts:27` (probe-only, no model turns), `subagent.ts` (governed by `denyElicitations`) — correctly NOT setting the flag — read/grepped at ref.
- New positive tests assert `ask_question` stripped + `terminal_font_size`/`read_file`/`bash` retained for both providers — read at ref.

## Unverified / not checked
- **Did not execute the test suite** — the worktree is on a different branch; checking out the ref to run tests would mutate worktree state (out of read-only scope). Test presence + assertions verified by reading source at the ref, not by running.
- Whether any third-party/plugin caller constructs `OpenAICompatibleProvider` with `providerOpts.tools` *and* `isNonInteractive: true` (finding #3's only escape path) — no production callers found by grep; plugin ecosystem not exhaustively scanned.
- Epistemic scope: all Wave-1 + verification citations read against branch HEAD `7dd7b91`; no patch-file fallback was needed.

The fix is correct, well-motivated, and the load-bearing premise (no-handler → auto-decline) is verified at source. Plumbing in both providers is sound, ternary precedence is right, and surface coverage is complete for the stated goal. The only actionable item is the low-severity wiring-test gap (finding #1), which is a follow-up nicety, not a merge blocker. Findings #2 and #3 are intentional/pre-existing nits.

**Done.** Reviewed PR #164 (light regime, 1 independent agent + inline Wave 1.5 verification against ref `7dd7b91`). Emitted 1 low + 2 nit findings, dropped 2 false-absence claims via grep verification, and verified the core premise + provider plumbing at the reviewed ref. Recommendation: **MERGE**. Read-only — no files, PR, or git state were modified.

---

PR #167, opened 2026-06-16
fix(session): seal provider-error turns as failed, not silent success

griffinwork40:
## What happened (witness trace `78a71c64-e05e-4e93-b3fe-b77e5112d429`)

A session ran for ~36s with model `mimo-v2.5`, produced **0 turns / 0 tokens / $0**, then was sealed as **`status: "succeeded"`** with `closure.reason: "model_end_turn"`:

```
seq4 session_phase  session_init_done durationMs=239
seq5 closure        reason=model_end_turn finalTurnCount=0 finalCostUsd=0 finalTokens={}   (~35.8s later)
seq6 session_sealed status=succeeded finalTurnCount=0
```

This is the "silence looks like success" failure mode: the model call failed (HTTP/auth/stream error after 36s), yet the trace recorded a clean success.

## Root cause

The OpenAI-compatible provider correctly emits a terminal `{ type: 'error' }` event on auth/stream/HTTP failure (`src/agent/providers/openai-compatible/query.ts`). But the session harness **never recorded that an error occurred**:

- `sendMessageStreamInternal` (per-turn loop) yielded the `error` OutputEvent and broke the loop — but only `done` events incremented `turnCount`, and nothing flagged the error. (`src/agent/session/agent-session.ts`)
- `pullInitialization` (init phase) did a bare `return` on an `error` event.

So when the surface later called `close()`:
- `deriveSealStatus('close')` → reason is not `'error'`, signal not aborted → **`'succeeded'`**
- `classifyClosureReason(...)` → fell through every guard → **`'model_end_turn'`**

A provider failure left zero trace in the terminal classification.

## Fix

Add a `sawProviderError` terminal-cause flag, following the existing `maxTurnsHit` / `hookBlocked` pattern (set at origin, read by `deriveClosureReason`):

- **Set** at both error-observation sites — `pullInitialization` (init-phase) and `sendMessageStreamInternal` (per-turn).
- **Cleared** by a subsequent completed turn, so it reflects the **final** turn's outcome (a turn that errored and was then retried successfully is *not* sealed failed).
- **Read** by `classifyClosureReason` → `abort` (ranked *below* a classified abort signal, so genuine budget/timeout cancels keep their more-specific status) and `deriveSealStatus` → `failed` (checked *after* the `cancelled` branch, for the same reason).

Net behavior change: a turn that ends in a provider error followed by an otherwise-clean `close()`/`reset()` now seals `failed` / `abort` instead of a silent `succeeded` / `model_end_turn`. **No other path changes** — clean sessions, budget/timeout cancels, hook-blocks, max-turns, and truncation all classify exactly as before.

### Deliberately *not* included (scope)
- **Blanket zero-turn to failed guard.** Rejected: it would false-fail a legitimate empty session (open REPL, quit without typing). The provider-error flag is the precise signal.
- **Warning when an unrecognized model id (`mimo-v2.5`) silently falls through `providerForModel()` routing tiers.** Real, but a separate concern (model misconfiguration vs. error classification) — worth a follow-up.

## Tests
- `closure-reason.test.ts`: 3 unit cases covering the new precedence (provider error → `abort`; abort signal outranks it; it outranks truncation).
- `closure.test.ts`: 2 end-to-end regressions driving a real `AgentSession` — (1) error turn → `close()` seals `failed`/`abort`; (2) error turn then successful turn → seals `succeeded`/`model_end_turn`.
- New `provider-error` marker in the mock provider fixture (mirrors the existing `slow` marker).

Full suite: **9329 passed**, 12 skipped. The single failing `src/cli/config.test.ts` case is **pre-existing on the clean baseline** (test-isolation bug reading the real `~/.afk` config) and unrelated to this diff. `tsc --noEmit` clean.

### DIFF
--- a/src/agent/__fixtures__/mock-provider.ts
+++ b/src/agent/__fixtures__/mock-provider.ts
@@ -85,6 +85,21 @@ export function createMockProvider(opts: MockProviderOptions = {}): MockProvider
 
             if (abortController.signal.aborted) return;
 
+            // Test affordance (mirrors the 'slow' marker above): a turn whose
+            // content contains 'provider-error' yields a terminal `error`
+            // event — exactly what the real providers emit on an HTTP / auth /
+            // stream failure — instead of the normal assistant.message +
+            // turn.completed pair. No turn.completed is emitted, so turnCount
+            // stays put and no usage is accumulated.
+            if (userContent.includes('provider-error')) {
+              yield {
+                type: 'error',
+                error: new Error('mock provider stream failure'),
+              } satisfies ProviderEvent;
+              abortController = null;
+              continue;
+            }
+
             yield {
               type: 'assistant.message',
               text: `Echo: ${userContent}`,
--- a/src/agent/session/agent-session.ts
+++ b/src/agent/session/agent-session.ts
@@ -102,6 +102,18 @@ export class AgentSession implements IAgentSession {
   private maxTurnsHit = false;
   private hookBlocked = false;
   /**
+   * True when the provider emitted a terminal `error` event (an HTTP / auth /
+   * stream failure) as the session's last turn outcome. Set at the two
+   * error-observation sites — `pullInitialization` (init-phase error) and
+   * `sendMessageStreamInternal` (per-turn error) — and cleared by a subsequent
+   * completed turn so it reflects the FINAL turn's result, not any error
+   * earlier in the session. Read by `deriveClosureReason` (→ `abort`) and
+   * `deriveSealStatus` (→ `failed`) so a provider failure on an otherwise-clean
+   * `close()` is not sealed as a silent `succeeded` / `model_end_turn`. Reset
+   * by `reset()`.
+   */
+  private sawProviderError = false;
+  /**
    * Wall-clock timestamp captured at construction — used to compute the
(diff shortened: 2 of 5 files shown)


griffinwork40:
🤖 _Automated review — agent-afk hourly sweep · updated 2026-06-16 12:04 UTC · re-evaluated each rotation._

✅ MERGE

  • Correctness verified at 204dff1: sawProviderError is set at both error sites — init (agent-session.ts:296) and per-turn loop (:527) — and cleared on a completed turn (:522). The third error site (:447) only re-throws the already-flagged event, so no path is missed.
  • Precedence correct in both deciders: classifyClosureReason (closure-reason.ts:85) and deriveSealStatus (agent-session.ts:1116) check the flag after the abort/cancel branches, so a genuine budget/timeout keeps its more-specific status.
  • api-compat safe: new required ClosureReasonInputs.sawProviderError is threaded to both (and only) consumers — agent-session.ts:1010 + the test base — so strict TS won't break.
  • Well-covered: 5 new tests (3 unit precedence + 2 integration) exercise the silent-success regression and the recovery-clears-flag path.

Non-blocking nits: reuses 'abort' as the closure reason (consistent w/ the existing init-error mapping); flag clears only on done but fails safe. Not run locally: tsc/test suite.

PR #168, opened 2026-06-16
fix(trace): seal orphaned witness traces on abnormal process exit

griffinwork40:
## What happened (witness trace `c490bddb-b4fb-4551-89b3-033a09bb28c5`)

A `qwen3.7-plus` REPL session emitted 5 events and then **stopped at `session_init_done`** — no `closure`, no `session_sealed`. The trace is **orphaned/unsealed**: a reader cannot tell whether the session is still running, crashed, or exited.

```
seq3 hook_decision     SessionStart
seq4 session_phase     session_init_done durationMs=224
<EOF — no closure, no session_sealed>
```

This is a **different and worse** failure than #167. #167 fixes a session that *reaches* `close()` but mis-seals (succeeded instead of failed). This one **never reaches `close()`**, so no terminal record is written at all.

## This is systemic, not model-specific

A survey of recent witness traces found unsealed traces are common and happen with **valid models (`opus_1m`) too**, terminating at every lifecycle stage: `bootstrap_done`, `session_init_done`, `model_ttfb`, `tool_call started`. The common factor is the **process exiting without running `AgentSession.close()`** (which is what writes the seal).

## Root cause

The seal is written by `AgentSession.close()` → `dispatchSessionEndOnce()` → `writer.seal()`. The interactive REPL only reaches `close()` via `ctx.rl.on('close')` or its SIGINT/SIGTERM/SIGHUP handlers (`src/cli/commands/interactive.ts`). Several exit routes bypass all of them:

- **Uncaught exception / unhandled rejection mid-turn** — the REPL has no process-level crash handler (only the daemon does, and even that just pushes Telegram + `process.exit(1)` — it does **not** seal).
- **Early stdin-EOF race** — if `rl` emits `'close'` before its seal handler is attached (~`interactive.ts:712`, after an `await initTranscript()`), the seal callback never registers. Reproducible with piped input.
- **Any `process.exit()` that skips cleanup.**

`SIGKILL` is uncatchable and out of scope.

## Fix — synchronous process-exit seal backstop

A single, surface-agnostic backstop in `NdjsonTraceWriter`:

- Each writer registers itself when its trace file is first opened; a **single shared `process.on('exit')` handler** synchronously appends a terminal `session_sealed { status: 'failed', incomplete: true }` for any writer that emitted events but never sealed.
- `seal()` is async (`appendFile` + `fsync`) and can't run in an exit handler, so the backstop uses `appendFileSync`.
- **Node fires `'exit'` on uncaught exceptions (empirically verified in this work), explicit `process.exit()`, and event-loop drain** — every *catchable* termination. The backstop therefore covers the crash gap, the EOF race, and stray `process.exit()` uniformly, in one place.
- The normal `close()` path `await`s `seal()` **before** `process.exit`, so the backstop cleanly **skips already-sealed traces** (no double seal). It also no-ops when nothing was written (no orphan file).

New optional **`incomplete`** flag on `SessionSealedPayload` (type + Zod schema) marks a backstop-written seal so readers can distinguish it from a normally-classified seal. The `final*` counters are `0` — the writer cannot reconstruct session totals from an exit handler; the `incomplete: true` marker says so explicitly.

### Result for the reported trace
`c490bddb` would now end with `session_sealed { status: 'failed', incomplete: true }` instead of dangling at `session_init_done` — a clear "this session terminated abnormally without completing" signal.

## Tests
- `writer.test.ts`: 3 unit cases — backstop writes `failed` + `incomplete`; no-op after a normal `seal()`; no-op when nothing was written.
- **Real subprocess crash test**: spawns a node process (via `tsx`) that writes one event then throws uncaught, and asserts the on-disk trace is sealed `incomplete: true` end-to-end — proving the `process.on('exit')` wiring.

`tsc --noEmit` clean. `src/agent/trace/` + `src/agent/session/` suites: **335 passed**.

## Scope notes
- This PR addresses the **observability** defect (orphaned traces). It does **not** add REPL `uncaughtException` handlers or fix the `rl.on('close')` registration race — the backstop makes both non-fatal for sealing, and they're worth a separate hardening pass.
- Separately, an unrecognized model id like `qwen3.7-plus` falls through `providerForModel()` routing to `anthropic-direct` with no warning (no `AFK_OPENAI_BASE_URL` set). That misconfiguration-silence is a distinct follow-up (also noted on #167).

Related: #167.

### DIFF
--- a/src/agent/trace/events.ts
+++ b/src/agent/trace/events.ts
@@ -386,6 +386,7 @@ export const SessionSealedPayloadSchema = z.object({
   finalCostUsd: z.number().nonnegative(),
   finalTurnCount: z.number().int().nonnegative(),
   closedAt: z.string().datetime(),
+  incomplete: z.boolean().optional(),
   subagentCount: z.number().int().nonnegative().optional(),
   subagentTokens: z
     .object({
--- a/src/agent/trace/types.ts
+++ b/src/agent/trace/types.ts
@@ -563,6 +563,18 @@ export interface SessionSealedPayload {
   finalTurnCount: number;
   /** ISO-8601. When known, mirrors the closure event's wall-clock. */
   closedAt: string;
+  /**
+   * True when this seal was written by the synchronous process-exit
+   * backstop ({@link NdjsonTraceWriter}) rather than by a normal
+   * `AgentSession.close()`. Signals that the process exited abnormally —
+   * crash, early-EOF before the REPL's close handler attached, or a
+   * `process.exit()` that bypassed cleanup — so the session never reached
+   * a clean terminal classification. `status` is `'failed'` and the
+   * `final*` counters are last-known-from-the-writer (0 when the session
+   * had no completed turns), NOT a reconstructed total. Omitted on every
+   * normal seal.
+   */
+  incomplete?: boolean;
   /** Number of subagent forks that reached `succeeded` status this session. */
   subagentCount?: number;
   /**
--- a/src/agent/trace/writer.test.ts
+++ b/src/agent/trace/writer.test.ts
@@ -13,7 +13,7 @@
 
 import { mkdtemp, readdir, readFile, rm } from 'fs/promises';
 import { tmpdir } from 'os';
-import { join } from 'path';
+import { dirname, join } from 'path';
 import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 
 import {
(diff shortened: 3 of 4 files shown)


griffinwork40:
Investigation complete. The `incomplete` field is present in both the Zod schema (`events.ts:389`) and the TS interface (`types.ts:577`), and `SessionSealedPayloadSchema` feeds `TraceEventSchema` (the exported whole-event validator) — so the schema addition is necessary for any validating reader to round-trip the field rather than strip it. All claims verified against reviewed ref `de83b723`. Synthesizing.

---

# Code Review — PR #168

**`fix(trace): seal orphaned witness traces on abnormal process exit`**
griffinwork40/agent-afk · `fix/seal-unsealed-traces-on-exit` → `main` · reviewed ref **`de83b723`**

**Triage:** hotfix · 4 files · +193/−1 (194 lines) → **regime: light**. Adds a synchronous `process.on('exit')` backstop so a `NdjsonTraceWriter` that emitted events but never sealed (crash / early-EOF / `process.exit()` bypassing `close()`) gets a terminal `session_sealed{status:'failed', incomplete:true}` record instead of being orphaned.

## What the change does (verified)
A module-level `Set<NdjsonTraceWriter>` (`writer.ts:121`) tracks live writers; each registers on first file open (`writer.ts:225`) and a single shared exit handler (`writer.ts:124-132`, install-guarded) calls `sealOnProcessExit()` on each. De-registration happens in `closeHandle()` (`writer.ts:232`), reached by both `seal()` and `close()`. New `incomplete?: boolean` added to the payload schema (`events.ts:389`) and interface (`types.ts:577`).

## Findings

**1 · LOW · confidence: medium · correctness · `writer.ts:169` + `writer.ts:249` · ref `de83b723` · citation-type: file-state**
`seal()` sets `this.sealed = true` at line 169 *before* the seal record is durably appended (the append + `fsync` run later inside the enqueued task). If process-exit fires inside that window, `sealOnProcessExit()` returns early on its `if (this.sealed || …)` guard (line 249) and the queued seal task never runs (event loop is dead at `'exit'`) — leaving the trace unsealed, the exact orphan state this PR targets, in a narrow window.
- Evidence:
  ```ts
  if (this.sealed) return;
  this.sealed = true;          // writer.ts:168-169 — set before the durable append
  await this.enqueue(async () => { /* append + fsync */ });
  ```
- Scope note: this is the `close()`-was-running case, explicitly *outside* the PR's stated scope ("exited WITHOUT running close()"), and the change is a strict net improvement over the prior unconditional-orphan behavior. Not a regression; flagged as a residual.
- Suggestion: gate `sealOnProcessExit()` on a distinct flag set *after* the durable append (e.g. `sealRecordPersisted`) rather than the optimistic `sealed` flag, so an interrupted `seal()` still receives the `incomplete` backstop.

## Verified clean (checked, no issue)
- **No double-seal:** `sealed` is set synchronously at the top of both `seal()` (169) and `sealOnProcessExit()` (250); the exit loop and a completed `seal()` are mutually exclusive. Test `no-op after a normal seal()` covers it.
- **`seq` monotonicity / no collision at exit:** async write tasks assign `seq` inside the serialized queue (`writer.ts:297`); at `'exit'` the loop is dead so pending tasks never assign — `appendFileSync` seal takes the current `seq` with no contention. O_APPEND + sub-`PIPE_BUF` lines prevent torn/interleaved writes between the async handle and the sync append (per `writer.ts:268` comment).
- **`finalCostUsd:0`/`finalTurnCount:0` is correct, not lossy:** `NdjsonTraceWriter` tracks neither cost nor turns (full read of the class — fields are only `traceDir`, `tracePath`, `seq`, `sealed`, `fh`, `writeQueue`). The type doc's "last-known-from-the-writer (0 …)" is accurate.
- **No memory/handle leak:** the new strong-ref `Set` is de-registered by every surface — daemon `finally` (`scheduler.ts:322-330`), Telegram evict/swap/closeAll (`session-manager.ts:435/459/512/596`), chat (`chat.ts:758`), REPL/resume-swap — all routing through `close()`/`seal()` → `closeHandle()` → `delete`.
- **Schema ↔ type ↔ reader consistency:** `incomplete` in both `events.ts:389` and `types.ts:577`; `SessionSealedPayloadSchema` feeds the exported `TraceEventSchema` discriminated union, so validating readers preserve the field.
- **Single exit listener** (install-guarded, `writer.ts:127`); **Set deletion-during-iteration** is well-defined and safe.
- **Tests:** strong — the e2e case spawns a real `tsx` subprocess that throws uncaught and asserts the on-disk trace is sealed `failed + incomplete` (`writer.test.ts`). Minor coverage gap: the Finding-1 seal-in-flight race is untested (consistent with it being an out-of-scope residual).

## Epistemic scope
- Wave-1 analysis + citation verification done against branch HEAD **`de83b723`** (fetched via `git show <ref>:<file>`; my worktree holds the *base* of `writer.ts`, so all citations resolve to the reviewed ref, not the worktree). Cross-checked by one independent investigation agent whose cited line numbers matched the reviewed-ref file.
- **Not run:** `pnpm test` (tests read as correct; not executed). Did not exhaustively trace every Telegram eviction branch (verified via grep at the four `close()` sites). Did not inspect a concrete downstream trace-*reader* consuming `incomplete` (schema round-trip verified structurally only). No critical/high findings → `/shadow-verify` not triggered.

---

## Done — Merge recommendation: **MERGE**

Focused, correct, well-documented, well-tested bugfix with no blocking, critical, high, or medium findings. The single LOW finding (seal-in-flight residual race, `writer.ts:169`+`:249`) is explicitly outside the PR's stated scope and a strict net improvement over the prior always-orphan behavior — address as an optional follow-up, not a merge gate. Review was read-only; no files, PR state, or comments were modified.

---

PR #170, opened 2026-06-16
fix(plan-mode): enforce write-tool gate (config.hookRegistry was dropped by the dispatcher)

griffinwork40:
## Ported from private — afk-workshop#706

**Source:** afk-workshop#706 — `fix(plan-mode): enforce write-tool gate (config.hookRegistry was dropped by the dispatcher)`. Moat-scrubbed port via **diff + 3-way apply** (the repos share no history), not a 1:1 commit copy.

### What this fixes
In REPL **plan mode**, write tools (`edit_file`, `write_file`, write-intent `bash`) ran **unblocked**. The plan-mode gate is the only built-in `PreToolUse` hook; it is dispatched by the provider's per-query `SessionToolDispatcher`, which was built from the provider's **constructor-time** `this.hookRegistry` (always `undefined` in production) and **ignored `config.hookRegistry`** — the session-scoped field `AgentSession` actually populates. So the gate never reached the dispatcher and writes ran unblocked in plan mode. Session-lifecycle hooks (SessionStart/End/SubagentStop, e.g. the memory writer) were unaffected — they dispatch from a separate path.

### Fix
Thread `config.hookRegistry` into `buildDispatcher` in **both** providers, preferring it over any constructor-time registry:
- `AnthropicDirectProvider`: `hookRegistry: opts?.hookRegistry ?? this.hookRegistry` (threaded from both dispatcher-build sites — initial + the `setCwd` rebuild closure).
- `OpenAICompatibleProvider`: `effectiveHookRegistry = opts.hookRegistry ?? this.providerOpts.hookRegistry`.

**Subagents are unaffected:** `forkSubagent` sets `parentSessionId` on the child config and the gate self-skips when present, so a subagent's own writes (incl. to its worktree) are not gated.

### Ported (all PUBLIC-SAFE)
- `src/agent/providers/anthropic-direct/index.ts`
- `src/agent/providers/anthropic-direct/plan-mode-gate-wiring.test.ts` *(new)*
- `src/agent/providers/openai-compatible/index.ts`
- `src/agent/providers/openai-compatible/query.test.ts`

`+391 / -4` — matches the source PR.

### Dropped
**None.** All four files were PUBLIC-SAFE; no moat files or hunks existed in the source PR.

### Conflict-resolution note (public had drifted ahead of the private patch base)
The two `index.ts` files conflicted on 3-way apply because public carries logic the private base lacked. Resolved by **applying the fix while preserving public-only logic**:
- Public's `isNonInteractive` support (call-site spread + `opts` type field) — **kept**.
- Public's MCP wire-name allowlist union (`withMcpToolsAllowed(...)`) on `permissions` — **kept**.

The resulting `git diff` vs public `HEAD` is exactly the fix's `hookRegistry` plumbing; no public functionality was removed.

### Verification
- `pnpm lint` (strict `tsc --noEmit`): **clean**.
- Ported regression tests: **40/40 pass** — `plan-mode-gate-wiring.test.ts` (3) + the `plan-mode gate via config.hookRegistry` block in `query.test.ts`. Each covers: top-level session **blocked**, forked subagent **not blocked** (`parentSessionId` self-skip), default mode **not blocked** (anthropic).
- Full suite: **9335 pass / 12 skip / 1 fail**. The single failure (`src/cli/config.test.ts` -> "defaults to the built-in tier bindings when no models block is present") is **proven pre-existing**: it reproduces identically on pristine public `HEAD` with zero port changes (env-dependent test isolation — reads the host's real `~/.afk` config). Not a regression.
- `pnpm build` + `pnpm build:dist`: clean. `pnpm fix:pins:check`: all hash pins up to date.
- **Moat safety:** deterministic tree+`dist/` guard -> `MOAT GUARD CLEAN`; independent adversarial audit -> `MOAT-CLEAN`.

**Do not merge automatically — left for human review.**

### DIFF
--- a/src/agent/providers/anthropic-direct/index.ts
+++ b/src/agent/providers/anthropic-direct/index.ts
@@ -300,6 +300,16 @@ export class AnthropicDirectProvider implements ModelProvider {
        * `get_runtime_state` against a dispatcher built without a source.
        */
       runtimeStateSource?: RuntimeStateSource;
+      /**
+       * Session-scoped hook registry sourced from `AgentConfig.hookRegistry`.
+       * Threaded here so `PreToolUse`/`PostToolUse` hooks (notably the
+       * plan-mode gate) fire on the per-query dispatcher. Production entry
+       * points construct the provider WITHOUT a constructor-time
+       * `hookRegistry` and supply the session registry on the query config
+       * instead, so falling back to `this.hookRegistry` when this is unset
+       * preserves any constructor-provided registry.
+       */
+      hookRegistry?: import('../../hooks.js').HookRegistry;
     },
   ): SessionToolDispatcher {
     const handlers = createBuiltinHandlers(permissionMode, opts?.cwd);
@@ -342,7 +352,12 @@ export class AnthropicDirectProvider implements ModelProvider {
       // Constraint (semantic invariant): MCP schemas appended AFTER builtins
       // so builtin tool names always take precedence in any overlap.
       schemas: [...this.schemas, ...mcpSchemas],
-      hookRegistry: this.hookRegistry,
+      // Prefer the session-scoped registry from the query config (the
+      // production path — see the opts.hookRegistry doc above); fall back to
+      // any constructor-provided registry. Without this, the plan-mode gate
+      // (the sole built-in PreToolUse hook) never reached the dispatcher and
+      // write tools ran unblocked in plan mode.
+      hookRegistry: opts?.hookRegistry ?? this.hookRegistry,
       // Union live MCP wire-names into the (statically-snapshotted) allowlist so
       // OAuth servers whose tools were discovered after construction are not
       // rejected by the gate while present in `schemas`/`handlers`. No-op when
@@ -632,6 +647,7 @@ export class AnthropicDirectProvider implements ModelProvider {
           parentSessionId: config.parentSessionId,
           traceWriter: config.traceWriter,
           runtimeStateSource,
+          hookRegistry: config.hookRegistry,
         });
 
     // External-dispatcher branch: the caller owns routing for whatever tools
@@ -839,6 +855,7 @@ export class AnthropicDirectProvider implements ModelProvider {
             parentSessionId: config.parentSessionId,
             traceWriter: config.traceWriter,
             runtimeStateSource,
+            hookRegistry: config.hookRegistry,
           });
           return { userSystem: newUserSystem, dispatcher: newDispatcher };
         };
--- a/src/agent/providers/anthropic-direct/plan-mode-gate-wiring.test.ts
+++ b/src/agent/providers/anthropic-direct/plan-mode-gate-wiring.test.ts
@@ -1,0 +1,257 @@
+/**
+ * Regression test for plan-mode GATE wiring (the enforcement half).
+ *
+ * `plan-mode-system-payload.test.ts` covers the POSTURE half — the system
+ * prompt addendum the model sees. This file covers the ENFORCEMENT half: when
+ * a session is in `'plan'` permission mode, a write-class tool such as
+ * `edit_file` must be REFUSED by the plan-mode gate before its handler runs.
+ *
+ * The gate is a `PreToolUse` hook. It reaches the per-query
+ * `SessionToolDispatcher` only if the provider threads the session's hook
+ * registry into the dispatcher. Production entry points (REPL/chat/daemon/
+ * telegram) construct the provider WITHOUT a constructor-time `hookRegistry`
+ * and instead pass the session-scoped registry on `AgentConfig.hookRegistry`
+ * (consumed by `provider.query({ config })`). This test mirrors that exact
+ * shape — `new AnthropicDirectProvider()` (no constructor registry) + a
+ * registry supplied on the query config — so a regression that drops
+ * `config.hookRegistry` on the dispatcher path is caught here.
+ *
+ * Observation point: the `tool.output` ProviderEvent. A gate block surfaces as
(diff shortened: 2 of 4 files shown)


griffinwork40:
All checks pass. The "sole built-in PreToolUse hook" comment is **accurate** (default-hook-registry.ts has exactly one PreToolUse registration at line 42; the others are SubagentStop/SessionEnd). And I've confirmed the one minor asymmetry: the openai test omits the default-mode negative case that anthropic includes.

Verification complete across all dimensions. No critical or high findings, so no `/shadow-verify` pass is required (it gates only on top-severity claims).

---

# Code Review — PR #170

**Target:** `fix(plan-mode): enforce write-tool gate (config.hookRegistry was dropped by the dispatcher)`
**Refs:** `main` ← `afk-port/pr-706` @ `c6892c6` · **Type:** hotfix · **Files:** 4 (2 prod, 2 test) · **Lines:** +391/-4 (prod ≈ +33/-3; rest tests) · **Regime:** light
**Reviewed ref:** `c6892c6e157bf6182dbedfb1170de77f08364fe5` — fetched locally; all blocking-tier citations verified against it.

## Summary
The fix is **correct, complete, and well-tested**. It restores a silently-bypassed safety gate: production entry points (chat/telegram/REPL/daemon) construct the provider with no constructor-time `hookRegistry` and supply the session registry via `config.hookRegistry`, but both providers' `buildDispatcher` hard-coded `hookRegistry: this.hookRegistry` (constructor-only) — so `config.hookRegistry` was dropped, **no PreToolUse hook fired**, and the plan-mode gate never blocked write tools (`edit_file`, `write_file`, `memory_update`, `procedure_write`) in plan mode. The diff threads `config.hookRegistry` through with a sound `?? this.hookRegistry` fallback.

## Verification performed
- **Bug is real, fix is real (end-to-end):** `createPlanModeGate` is registered as the PreToolUse hook in `default-hook-registry.ts:42`; production surfaces build it via `createDefaultHookRegistry` (chat.ts:571, telegram.ts:350/383, interactive bootstrap) and pass it as `config.hookRegistry`. The dispatcher gates *all* PreToolUse on `if (this.hookRegistry)` (dispatcher.ts:465, 615) — a dropped registry = zero hooks.
- **Completeness (the key risk):** Every dispatcher-construction path is covered. anthropic-direct has exactly 2 `buildDispatcher` callers (index.ts:626, 833) — both now pass `hookRegistry: config.hookRegistry` (verified at ref lines 650, 858); openai-compatible has exactly 1 (index.ts:198) — fixed. The only `new SessionToolDispatcher` in each file is inside `buildDispatcher`. `query.ts` matched grep only via a *comment*. The external-dispatcher branches (`this.externalTools` / `this.providerOpts.tools`) are caller-owned by design — correctly not gated. No third provider exists.
- **Fallback semantics:** `opts?.hookRegistry ?? this.hookRegistry` (anthropic) and `opts.hookRegistry ?? this.providerOpts.hookRegistry` (openai). A `HookRegistry` is never falsy, so `??` is correct; constructor-provided registries are preserved. The two providers use different idioms (direct pass vs. conditional spread) but each matches its file's existing convention and is behaviorally equivalent.
- **Subagent self-skip is real:** the gate self-skips on `context.parentSessionId` (plan-mode-gate.ts), and the dispatcher threads `parentSessionId` into the PreToolUse context (dispatcher.ts:470, 620). The over-blocking guard is genuine.
- **Tests are meaningful, not tautological:** traced the block-surfacing path — gate `reason` → `HookBlockedError.message` (hook-registry.ts:181 embeds `decision.reason`) → dispatcher wraps as `Tool "edit_file" blocked by PreToolUse hook: …plan mode: edit_file is refused…` with `isError:true`. So the positive assertions (`'plan mode'` + `'edit_file'`) validate real propagation; the negatives (self-skip/default → real handler runs on a nonexistent file → error contains neither string) discriminate correctly.

## Findings

**nit · test-coverage · `src/agent/providers/openai-compatible/query.test.ts` · ref:c6892c6 · citation-type:file-state**
The openai-compatible `plan-mode gate via config.hookRegistry` block has 2 cases (BLOCKS in plan mode @375, subagent self-skip @394) but omits the *"does NOT block in default mode"* case that the anthropic-direct test includes (plan-mode-gate-wiring.test.ts). Both providers share the same gate, which reads mode at call time identically, so risk is minimal — but parity would close the asymmetry.
*Suggestion:* add a third `it('does NOT block edit_file when the session is in default mode', …)` to the openai-compatible block, mirroring the anthropic test.

No security, api-compat, correctness, or perf-observability issues. The opts-type gains an optional `hookRegistry?` field — additive, backward-compatible, internal to a private method. Block decisions emit `hook_decision` trace events, so observability is intact. Construction-time-only code path — no hot-path perf concern.

## What was not checked
- **Tests were not executed in-session.** The current worktree is at `b1078c6` (pre-fix); running the PR's tests would require checking out the fix (a mutation this read-only review forbids). Test *logic* is verified by reading + the block-surfacing trace above; `tsconfig` strict mode would reject type errors at build. Citations verified against the live ref `c6892c6`.
- Did not exercise the runtime against a live model (only the mocked Messages/Chat-Completions plumbing the tests use).
- Did not review the plan-mode gate's bash-denylist completeness — out of scope (unchanged by this PR; the gate's own docs already disclaim it as best-effort, not a security boundary).

A complete, correctly-scoped, well-cited fix for a real silent-bypass of the plan-mode write gate, with end-to-end regression tests on both providers. The single nit (default-mode test parity on the openai provider) is non-blocking.

This review is read-only — no files, commits, or PR state were modified. STOP.

---

PR #78, opened 2026-06-10
refactor(tui): split terminal-compositor + bug-hunt pass

griffinwork40:
## Summary

Splits the monolithic `terminal-compositor.ts` (the TUI rendering engine) into focused sibling modules and runs a **bug-hunt pass** over the compositor. Structural changes are behavior-preserving; the two bug fixes were added **test-first** (failing test → fix). Full suite green: **8752 tests pass, 0 failures**, `tsc --noEmit` clean.

Continues the existing *host-class + free-function-module* pattern already used by the `paste` / `autocomplete` / `render` / `committed-band` / `input-dispatch` / `frame` sibling modules — each new module takes the `TerminalCompositor` host via a narrow `*Host` structural-subset interface.

## Commits (each independently gate-green)

1. **refactor(tui): extract input-mode + arm/disarm lifecycle from terminal-compositor** — host `terminal-compositor.ts` **1235 → 843 LOC**; new `terminal-compositor.input-mode.ts` (275) + `terminal-compositor.lifecycle.ts` (349). Public barrel + all 12 external importers byte-unchanged.
2. **fix(tui): acquire stdin claim before raw-mode in arm()** — **HIGH.** `arm()` enabled raw mode + bracketed-paste *before* `acquireStdinClaim()`; on a claim conflict it threw with `armed=false`, so `disarm()`'s restore path never ran and the terminal was **left in raw mode (and bracketed-paste) for the process lifetime**. Fixed by acquiring the claim first — it throws with zero terminal side-effects, nothing to roll back. + regression test.
3. **fix(tui): accept multi-codepoint emoji in printable input** — **MED.** The printable-key filter used `char.length === 1` (a UTF-16 code-*unit* count), silently dropping surrogate-pair emoji (`😀`.length === 2) and variation-selector / skin-tone emoji (`❤️`, `👍🏽`). Replaced with a single-grapheme check (`nextGraphemeIndex(s,0) === s.length`). + regression test.
4. **refactor(tui): split committed-band into commit + repin modules** — 599 → thin barrel + `committed-band-repin.ts` (133, post-render/resize re-pinning) + `committed-band-commit.ts` (474, the commit pipeline).
5. **refactor(tui): extract frame preserve/evict helpers into frame-preserve** — `frame.ts` 496 → 344 + `frame-preserve.ts` (164).

## The bug-hunt

[...]

PR #172, opened 2026-06-16
fix(tui): keep multi-line blocks visible (band-hold) when committed under a tall overlay

griffinwork40:
**Source:** [afk-workshop#649](https://github.com/griffinwork40/afk-workshop/pull/649) — "fix(tui): keep multi-line blocks visible (band-hold) when committed under a tall overlay"

> ⚠️ **This is NOT a 1:1 port — it is a structural re-targeting + hand-integration onto public's evolved code.** Please read "How this differs from the private PR" before reviewing. Moat triage: nothing to scrub (pure TUI rendering).

## What it does
When a multi-line block (e.g. a streamed markdown table / "Done" report) is committed via `commitAbove` while a tall overlay is up, the old overflow path archived it to scrollback **and** painted a truncated on-screen copy — duplicate header, orphan `├──┼──┤` divider, a blank void swallowing the body, and prior prose vanishing. Band-hold instead keeps the full run in the `committedBand` model (capped at `maxBandModel`), paints only the rows that fit now, holds the rest pending, and `repositionCommittedBand` materializes the whole run contiguously on collapse. Genuine overflow (beyond the collapsed screen) is archived to scrollback as real content, chunked by screen height.

## Ported (all public-safe)
- New pure helpers: `src/cli/commit-mode.ts` (`decideCommitMode`, `capBandModel`) + `hardWrapToWidth` in `src/cli/wrap.ts`.
- Band-hold integrated into `src/cli/terminal-compositor.committed-band-commit.ts`: routing via `decideCommitMode`, a Phase-1 band-hold archive branch, Phase-3 band-hold branches (paint suffix / store-pending), and wrap-aware physical line counting. The superseded `coveredBand` park/promote is removed.
- `coveredBand` removed from `committed-band-repin.ts` and the field on `terminal-compositor.ts`.
- `markdown-stream.ts`: `syncPendingOverlay()` re-composes the overlay before each `commitAbove` (the load-bearing fix — keeps a stale tall overlay from pinning the frame to row 1).
- Regression tests: `commit-mode.test.ts`, `terminal-compositor.{overflow-gap,h1-prevtoprow,wrap-overlap}.test.ts`, plus updates to `golden-rendering`, `markdown-stream`, `commit-collapse-wrap-gap` tests.
- `docs/scrollback.md` design notes.

## Dropped (and why)
Nothing dropped for moat — the change carries none. One **intentional non-port**: I did **not** apply the private PR's tail→head flip of the legacy overflow paint. Public's tail-slice (which keeps a verdict card's closing `╰──╯` visible) is load-bearing for public's existing tests; band-hold subsumes the card-sized case instead, so the legacy overflow path stays untouched.

## How this differs from the private PR (reviewer: please read)
Private's `committed-band.ts` is monolithic; **public split it** into `committed-band-commit.ts` + `committed-band-repin.ts` (public #78) **and** carries committed-band fixes the private #649 base never had — separator/TUI-rhythm, `first-turn-banner-echo`, `pad-decay-straddle`, `banner-commit-gap`, `wrap-gap`, the extended-contiguity band merge, orphan-row erase, and `scrolledRows`/anchor-floor-follow. So the private diff did not apply. Band-hold was **re-implemented onto public's base, surgically + additively**:
- Band-hold is a **new route** checked before the existing fits/overflow branches; public's fits/overflow/separator/shrink-pad/contiguity machinery is left **unchanged**.
- `decideCommitMode` is fed a **separator-inclusive** `bandLineCount`/`bandTextLines`, so a held block keeps its trailing rhythm blank — the correct extension of public's rhythm contract into band-hold (validated by both the band-hold tests and public's separator/rhythm tests passing).

## ⚠️ One latent edge — NOT fixed, needs a real terminal
`scrolledRows = fitsAboveFrame ? bandOverflow : 0` in `committed-band-commit.ts`. If public's `fitsAboveFrame` **and** `useBandHold` are both true at once (narrow: a banner session + a non-contiguous existing band + a trailing separator + `room == contentLineCount` exactly), the anchor floor could decrement without a matching band-hold scroll (≤1-row floor drift). It is test-green and nearly unreachable, and I would not ship an unvalidated change to floor logic. Candidate fix to evaluate with a real terminal: `scrolledRows = useBandHold ? 0 : (fitsAboveFrame ? bandOverflow : 0)`.

## Verification
- `tsc --noEmit` clean · `pnpm build` ok · `pnpm build:dist` ok.
- Full suite: **9359 passed · 12 skipped · 1 failed**. The single failure (`config.test.ts` model-slot bindings) is **pre-existing and environment-local** — it fails identically on the unmodified baseline (the test reads local model config), and is unrelated to this change.
- Deterministic moat guard over tree + `dist/`: **CLEAN**. Independent adversarial moat audit: **CLEAN**.

## Follow-up (not in this PR)
Sync the reconciled band-hold back into private `afk-workshop` so the two repos' committed-band implementations converge (private has band-hold on the monolithic structure; this PR has it on the split structure with public's newer fixes).

---
**Please do not merge automatically — this needs human review**, especially the re-targeting and the `scrolledRows` edge note.

### DIFF
--- a/docs/scrollback.md
+++ b/docs/scrollback.md
@@ -401,7 +401,129 @@ massive gap. (b) When content scrolled into scrollback during a tall phase and
 the frame later collapses, the bottom-anchored band can sit below the
 scrolled-off content with blank viewport rows between them. That live-viewport
 "boundary gap" is the bottom-anchored-frame design tension (the frame does not
-float up to meet sparse content) and is a separate, larger item.
+float up to meet sparse content) and is a separate, larger item — but see the
+band-hold fix below, which closes the most common manifestation.
+
+### Fixed: multi-line block committed under a tall overlay (the overflow path)
+
+**Symptom.** During a long turn (a tall "thought"/tool overlay up), the streamed
+"Done" report committed a markdown TABLE via `commitAbove`. The table rendered
+its header + `├──┼──┤` divider, then a large blank VOID swallowed the body rows,
+and later content (Evidence) resumed far below — and the table was DUPLICATED (a
+full copy in scrollback plus a truncated on-screen copy) with an orphan divider.
+The prose committed just before the table (`Diagnosis complete`, `What I
+diagnosed`) VANISHED entirely.
+
+**Root cause.** The `!fitsAboveFrame` OVERFLOW path (the one #645 left unchanged)
+fired whenever a block was taller than the room above the *current* frame —
+which, under a tall overlay, is only a few rows. That path was written for blocks
+genuinely taller than the SCREEN: it CUP-wrote the whole block at the anchor
+floor (clobbering the prior committed band) and scrolled the entire block into
+scrollback (Phase 1), then painted a truncated on-screen copy (Phase 3). So a
+10-line table committed under a 14-line overlay — which fits fine once the
+overlay collapses — was wrongly archived + truncated + the prior band was lost.
+
+**The fix (band-hold).** A block that overflows the *current* tall frame but fits
+the *collapsed* screen now takes a band-hold path instead of the archive path:
+the full committed run (prior band + new block) is kept in the `committedBand`
+MODEL, capped at `maxBandModel` (the rows that can ever show above a minimal
+frame). Only the bottom `room` lines that fit above the current frame are
+painted; the rest are "pending" — present in the model, not on screen, not in
+scrollback. Nothing is scrolled. When the overlay collapses, the existing
+`repositionCommittedBand()` (which keys off `committedBand.length`, not the
+painted span) materializes the WHOLE run contiguously adjacent to the frame.
+Once pending lines exist, subsequent commits route through band-hold too, so the
+fits-path's room-based `bandOverflow` never scrolls the unpainted rows as blanks.
+A block taller than `maxBandModel` committed with NO pending rows is genuinely
+off-screen and keeps the legacy archive path; the fits path and its exact scroll
+mechanics are untouched (the routing is additive). But once pending rows already
+exist — a streamed table/report grown past `maxBandModel` under a *sustained*
+tall overlay — the commit STAYS on band-hold even though the run now exceeds
+`maxBandModel` (review #649 P1). Routing such a commit to the fits path instead
+would emit room-based line-feeds that scroll the unpainted pending rows into
+scrollback as BLANKS while Phase 3's cap drops the real rows — losing most of the
+report. Band-hold Phase 1 instead archives the genuine overflow (the oldest rows
+beyond what the collapsed screen can hold, chunked by screen height so a run
+taller than the terminal still archives every row) to scrollback as REAL content,
+disjoint from the suffix Phase 3 keeps — no blanks, no drops, no duplicates. Under
+a transient overlay GROWTH between the commit and the collapse, evict-on-growth
+materializes + scrolls the band's overflow as real, contiguous content — so
+band-hold degrades gracefully to contiguous scrollback, never to the
+duplicate/void corruption.
+
+**Regression test.** `src/cli/terminal-compositor.overflow-gap.test.ts` commits a
+rendered markdown table under a 14-line overlay at `extraRows=2`, collapses, and
+asserts (via `@xterm/headless`): the table header appears exactly once, the whole
+table + surrounding prose are VISIBLE and intact in the viewport, there is no
+>=2-row blank void, and the run hugs the frame. It fails on the pre-fix code
+(header found twice + 14-row void) and passes after. A second case in the same
+file streams a multi-block report under a *sustained* 17-line overlay until the
+band-hold model fills `maxBandModel`, commits one more block (run →
+`maxBandModel`+2), collapses, and asserts every committed row is present exactly
+once across scrollback + viewport — covering the pending-row loss of review #649
+P1 (on the pre-fix code the oldest row is dropped entirely).
+
+### Fixed: lost table under a *full-viewport* overlay + the wrap-blind overlap (review #649 follow-ups)
+
+**Symptom.** A `/review` streamed a markdown TABLE under a tall subagent-tree
+overlay: the table rendered its header + `├──┼──┤` divider, then the body rows
+VANISHED, and the next section resumed below — even on the band-hold fix above.
+Separately and intermittently, a new `commitAbove` block "ate" the bottom line
+of the *previously* committed block. Three distinct defects, surfaced together.
+
+**1. Commit-time overlay sync (the trigger — `markdown-stream.ts`).**
+`StreamingMarkdownRenderer.push()` commits a completed block via `commitAbove`
+while the throttled (33 ms) overlay repaint has NOT yet re-rendered — so the
+`markdown-pending` overlay slot still shows the just-committed block. That stale,
+too-tall overlay pins the live frame to row 1 (`prevTopRow == 1`) at the exact
+moment of commit. Fix: `syncPendingOverlay()` re-composes the overlay from the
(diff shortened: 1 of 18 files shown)


griffinwork40:
Confirmed: `OverlayComposer.flush(): void` is **synchronous** — it calls `slot.render()` and `sink.setOverlay()` inline, and the class doc explicitly notes "each `flush()` runs to completion before the next event or timer." So the load-bearing `syncPendingOverlay()` fix genuinely completes before `commitAbove()` on the production (overlayComposer) path. That retires the one open question both the regression and stream-sync agents flagged.

I have everything needed to synthesize. (Note: the working tree is intact on `main`; no agent mutated it, and I did not run the test suite since the branch isn't checked out and the tree must stay untouched — so the sub-agent's "tests pass" claim is treated as unverified, to be confirmed by CI.)

---

# Code Review — PR #172: band-hold for multi-line blocks committed under a tall overlay

**`fix(tui)` · `afk-port/pr-649` → `main` · +1360 / −99 · 14 files**
**Recommendation: ✅ APPROVE — merge-ready.** No blocker or high-severity findings. The core geometry is provably sound, the `coveredBand` removal is clean, and routing is genuinely additive. A short list of non-blocking follow-ups below.

## What it does
Replaces the lossy `!fitsAboveFrame` overflow path (duplicate header + blank "void" + dropped prose when a markdown table/report commits under a held tall overlay) with a **band-hold** path: keep the full committed run in the `committedBand` model (capped at `maxBandModel`), paint only the rows that fit above the current frame, archive only genuine overflow as *real* scrollback, and let `repositionCommittedBand()` paint the whole run on collapse. Extracts the routing/geometry arithmetic into a pure, unit-tested `commit-mode.ts`; makes line-counting wrap-aware via new `hardWrapToWidth()`; and adds a load-bearing `syncPendingOverlay()` so the overlay drops the just-committed block *before* `commitAbove` runs.

## Findings

**Blocker / High:** none.

### Medium (non-blocking — coverage)
- **`overlayComposer` ordering path is untested.** The new `markdown-stream.test.ts` ordering-contract test wires the *direct `compositor` stub* only, not an `OverlayComposer`. The production REPL path uses the composer. I verified that path is correct-by-construction (flush() is synchronous; the slot closure is live), so this is a coverage gap, not a live bug — but a future change to slot registration could silently break the sync with no failing test. *(stream-sync; confirmed by my flush() read)*
- **No test for resize (SIGWINCH) while a band-hold model is pending.** The model stores rows already hard-wrapped at the commit-time `cols`; a narrower resize before collapse repaints them verbatim without re-wrapping. This extends a *pre-existing* band limitation (the band always stored wrapped rows), but band-hold widens the exposure window by holding more unpainted rows across a longer interval. Worth a tracking note; not a merge blocker. *(test-coverage flagged HIGH; correctness flagged it pre-existing — reconciled to Medium)*

### Low
- **Full-viewport hold keeps content only in the model, not scrollback.** In the `newTopRow ≤ 1` fully-pending branch (`committed-band-commit.ts` ~610–620), when `overflowRun.length ≤ maxBandModel` nothing is archived. Base parked into `coveredBand` *and* archived to scrollback. So if `disarm()` / abort / `/clear` fires before any overlay collapse, the held block is lost from screen **and** scrollback. Narrow (the normal turn lifecycle always collapses → repin paints it), intentional trade-off (one clean copy vs. base's duplicate/void), but a real edge regression. *(correctness #2)*
- **Band-hold Phase 3 lacks the orphan-erase its sibling fits-branch has.** If `bandTop` ever moved *down* between band-hold commits it would leave a ghost row; the correctness agent proved this is currently unreachable (routing dynamics keep `bandTop` monotonic toward the floor) and the "accumulate multiple blocks" test would catch a ghost (green). Latent robustness item — relies on an unstated invariant with no defensive erase. *(correctness #1)*
- **`overflow-gap.test.ts` anchors on a rendered glyph** (`⎯` U+23AF / braille spinner). If a CI headless font doesn't render it, `frameIdx === -1` fails *before* the real assertions. A structural fallback anchor would be more robust. *(test-coverage)*
- **`setOverlay` dedup can no-op the sync** when post-slice buffer and prior overlay are both empty — harmless (overlay was already correct). *(stream-sync)*

### Nits
- **Dead clause** `width === Number.POSITIVE_INFINITY` in `hardWrapToWidth` (`wrap.ts` ~44) is unreachable after `!Number.isFinite(width)`. Cosmetic. *(regression)*
- **Implicit coupling:** `scrolledRows = fitsAboveFrame ? bandOverflow : 0` (committed-band-commit.ts ~345) is computed before the band-hold branch; it's only safe under `fits && hold` because that combination forces `anchorRow ≤ 1` (proven by exhaustive search). Worth a one-line comment to lock the invariant. *(correctness nit)*
- Micro coverage gaps: CJK/emoji 2-col width through `hardWrapToWidth`; an integration test for a single block taller than the terminal (multi-chunk archive); `cols === 1`; the docs' "evict-on-growth" degradation path. All low risk.

## Verified sound (highlights)
- **Partition proof is exact:** archived prefix `[0, genuineOverflow)` and kept suffix `capBandModel(run, maxBandModel)` exactly partition `overflowRun` — no gap, no overlap, no duplication — across all edge cases (independently re-derived).
- **`coveredBand` removal is complete:** `git grep` on the PR head returns *zero* code references (one historical comment in a test). The promotion block's role is cleanly subsumed by `repositionCommittedBand` reading `committedBand`, which band-hold now populates even at `prevTopRow ≤ 1`.
- **No band/frame drift:** `measure()` → `hardWrapToWidth` uses the *same* `stdout.columns` width and `wrapAnsi` option set as `CupFrameRenderer`, so the invariant the deleted drift-warning comment protected still holds; narrow content is a strict no-op.
- **Routing is additive:** `useBandHold` is mutually exclusive with `fitsAboveFrame` by construction and checked first; no input that previously took fits/legacy-overflow is silently re-routed.
- **The load-bearing fix is real:** `syncPendingOverlay()` → `overlayComposer.flush()` → `slot.render()` → `renderPending()` is fully synchronous (confirmed by reading `overlay-composer.ts`), so the overlay drops the committed block before `commitAbove` runs.
- **Tests are honest regression guards:** genuine red-on-base / green-after reasoning holds for each; assertions are *not* weakened (`golden-rendering` keeps `toBe(150)`; `commit-collapse-wrap-gap` change is comment-only with HARD GATE intact); `commit-mode.test.ts` arithmetic (overflowTargetBottom=21, maxBandModel=20) is correct.
- Repo conventions respected: the new ≥15-line comment uses the `Invariant:` prefix; overlay-sync-before-commit honors the "no optimistic rendering / teardown-before-setup" ordering rule.

## Suggested follow-ups (none block merge)
1. Add an `OverlayComposer`-wired case to the ordering-contract test (closes the Medium coverage gap on the production path).
2. File a tracking note for resize-during-band-hold re-wrap (extends a pre-existing limitation).
3. Optional polish: drop the dead `POSITIVE_INFINITY` clause; add the `scrolledRows`/`anchorRow ≤ 1` invariant comment; consider a scrollback fallback (or a documented "lost on abort" note) for the full-viewport-disarm edge.

## Caveats / not checked
- **Test execution not independently verified.** The PR branch isn't checked out and I kept the working tree untouched (it's on `main`, mid-work), so I did not run `pnpm test` / `tsc`. The sub-agent's "281 tests pass" claim is **unverified** — rely on CI to confirm green. This review's contribution is the static logic/arithmetic/coverage analysis above.
- Not exercised: byte-level ANSI correctness against a real terminal (only `@xterm/headless` + arithmetic); `DECSTBM` scroll-region interaction beyond the green tests; full resize-narrower + band-hold permutations.
- This was a **read-only** review: no files edited, no PR comment posted, no commits — the dirty working tree was preserved untouched.

**Done.**

---

PR #173, opened 2026-06-16
refactor(provider): promote hookRegistry into the dispatcher contract

griffinwork40:
## Summary

Make AFK's session hook registry **structural instead of convention-wired**, closing the highest-leverage instance of a broader theme: safety/protocol guards that exist but are enforced only by convention. Direct structural follow-up to #170 — which fixed the *symptom* (`config.hookRegistry` dropped by the dispatcher → plan-mode write-gate never fired), while this fixes the *gap that let it happen*.

- **`hookRegistry` is now a required key** on `SessionToolDispatcherOptions` (`HookRegistry | undefined`, not optional). A dispatcher construction that fails to thread it is a **compile error**, not a silent runtime gap — so the `c6892c6` bug class cannot recur unnoticed.
- **One canonical merge path.** `resolveSessionHookRegistry()` is the single merge point (query-scoped `config` registry wins over a constructor-time one). Both providers call it; the two divergent inline `?? this.hookRegistry` merges are deleted.
- **Cross-provider contract test** (`hook-registry-contract.test.ts`, parametrized over both providers): asserts the plan-mode write-gate fires through the provider boundary, self-skips for subagents, and stays silent in default mode. Adding a provider = adding one adapter.
- **SDK audit now enforced in CI.** Wired `pnpm audit:sdk:check` into `ci.yml` — the `.sdk-dependency.lock.json` guard existed but was never run, so an unlocked `@anthropic-ai/sdk` import could merge unnoticed.

No behavior change (the resolver returns the same value the inline merges did).

## Test results

- `pnpm test` (vitest, full suite): **9348 passed / 12 skipped / 1 failed (9361)**. The lone failure is a **pre-existing, environment-specific leak** in `src/cli/config.test.ts` — `loadConfig()` reads the developer's real `~/.afk/config` local-model slot instead of the mocked empty config. It reproduces on clean `main` and is unrelated to this diff (which touches zero config files).
- `pnpm lint` (`tsc --noEmit`, strict): **pass (exit 0)**.
- `pnpm audit:sdk:check`: **pass** (lock matches current imports).

## Verification (`--verify` adversarial wave)

An independent read-only verifier re-derived the load-bearing claims from the diff alone (not from commit messages):

- **Required-key contract — CONFIRM.** `dispatcher.ts:132` declares `hookRegistry: HookRegistry | undefined` (non-optional); omission fails to compile.
- **Single canonical merge — CONFIRM.** Both providers call `resolveSessionHookRegistry()` (`anthropic-direct/index.ts:360`, `openai-compatible/index.ts:358`); no inline `?? this.hookRegistry` merge remains on the dispatcher path. (The `openai-compatible/index.ts:209` spread is a *subagent* options object, not a dispatcher merge.)
- **Gate fires through provider boundary — CONFIRM (refined).** `hook-registry-contract.test.ts:337-357` asserts real block behavior (`isError === true`, content contains "plan mode") end-to-end. Refinement: it reads the event-stream `isError`/`content` rather than the internal `decision: 'block'` object — a behavioral check, not plumbing-only. Not a blocker.

No CONTRADICT. Verifier confidence 90%; its only open gap (lint not independently run) is closed — lint passed in this session.

## Out of scope

- Broader "guard exists but doesn't bite" sweep: **signal-block** was verified to be *honestly* deferred v0 passive-observation infra (docs accurate, adoption real) — **not a gap**, no change; **`fix:pins:check`** absent from CI is **not a gap** — the same hash-pin invariant is already enforced by `vendored.test.ts`/`bundled.test.ts` under `test:coverage`.
- The pre-existing `config.test.ts` env-leak (separate fix).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,9 @@ jobs:
       - name: Audit env-registry docs (docs/env-registry.{json,md} in sync with src/config/env.ts)
         run: pnpm scan:env:check
 
+      - name: Audit SDK dependency surface (no unlocked @anthropic-ai/sdk imports)
+        run: pnpm audit:sdk:check
+
       - name: Build
         run: pnpm build
 
--- a/src/agent/hooks.ts
+++ b/src/agent/hooks.ts
@@ -179,3 +179,22 @@ export interface HookRegistry {
 export function createHookRegistry(): HookRegistry {
   return createHookRegistryImpl();
 }
+
+/**
+ * Resolve the hook registry a provider threads into its per-query dispatcher.
+ * Precedence: the query-scoped session registry (`AgentConfig.hookRegistry`,
+ * supplied per `provider.query({ config })` — the production path) wins over a
+ * constructor-time registry (legacy / test convenience). `undefined` only when
+ * neither is set.
+ *
+ * Single canonical merge point for every provider — do not re-implement the
+ * precedence inline. `SessionToolDispatcherOptions.hookRegistry` is a required
+ * key, so building a dispatcher without threading this result is a compile
+ * error (the c6892c6 plan-mode write-gate regression, made structural).
+ */
+export function resolveSessionHookRegistry(
+  queryScoped: HookRegistry | undefined,
+  constructorScoped: HookRegistry | undefined,
+): HookRegistry | undefined {
+  return queryScoped ?? constructorScoped;
+}
--- a/src/agent/providers/anthropic-direct/index.ts
+++ b/src/agent/providers/anthropic-direct/index.ts
@@ -47,6 +47,7 @@ import { builtinToolSchemas, agentTool, skillTool, composeTool } from '../../too
 import { TOOL_SYSTEM_PROMPT_BASE, SLASH_COMMAND_ROUTING_PROMPT, BASH_PASSTHROUGH_PROMPT, MEMORY_SYSTEM_PROMPT, MEMORY_SYSTEM_PROMPT_READONLY } from '../../tools/system-prompt.js';
 import { withMcpToolsAllowed, type ToolPermissionConfig } from '../../tools/permissions.js';
 import type { HookRegistry } from '../../hooks.js';
+import { resolveSessionHookRegistry } from '../../hooks.js';
 import type { SubagentExecutor } from '../../tools/subagent-executor.js';
 import { maxOutputTokensFor } from '../../model-limits.js';
 import type { SkillExecutor } from '../../tools/skill-executor.js';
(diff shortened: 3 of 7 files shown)


griffinwork40:
All verifications complete. Synthesis follows.

| Finding (Wave 1) | Type | Status | Decisive evidence |
|---|---|---|---|
| A1-F1: integration.test.ts → `tsc` compile break / CI failure | consequence | **REFUTED → dropped** | `tsc --noEmit` exits 0; `tsconfig.json` excludes `src/**/*.test.ts`; `--listFilesOnly` count for integration.test.ts = 0 |
| A1-F2: stale optional local `dispatcherOpts` type | citation | **REFUTED → dropped** | line 349 types it `ConstructorParameters<typeof SessionToolDispatcher>[0]`; `tsc` exit 0 is dispositive |
| A1: `audit:sdk:check` CI step references real script | claim | **CONFIRMED** | package.json:35 + `scripts/audit-sdk-dependency.ts` exists at ref |
| A2-F1 / A1-nit: openai conditional→unconditional assignment | correctness | **CONFIRMED behavior-preserving** | dispatcher consumes via value checks only (229/472/622/841); zero `'hookRegistry' in opts` |
| A2-F3: contract test covers c6892c6 for both providers | test | **CONFIRMED** | `PROVIDER_CASES=[anthropic,openai]` × `describe.each` plan-mode block |
| A2-F4: constructor-scoped fallback not exercised end-to-end | absence | **CONFIRMED-ABSENT → stands (low)** | `git grep resolveSessionHookRegistry -- '*.test.ts'` only in contract test; `makeConfig` always supplies query-scoped registry |

No `critical`/`high` finding survives verification → `/shadow-verify` not triggered (fires only on top-severity claims).

---

# Code Review — PR #173

**`refactor(provider): promote hookRegistry into the dispatcher contract`**
`fix/hookregistry-provider-contract` → `main` · regime: full · reviewed ref: `3eb1278`

## Summary
Promotes the per-query hook-registry precedence into a single canonical resolver (`resolveSessionHookRegistry`) and makes `SessionToolDispatcherOptions.hookRegistry` a **required, value-nullable** key — converting the c6892c6 plan-mode-gate regression (write tools ran unblocked in plan mode because `config.hookRegistry` was silently dropped) from a latent runtime gap into a compile error for any production provider. Both providers are rewired through the resolver, a 391-line cross-provider contract test asserts the invariant uniformly for both backends, and a working `audit:sdk:check` CI gate is added.

## Findings (post-verification)

**low · high · test-coverage · `src/agent/providers/hook-registry-contract.test.ts:405-468` · ref:3eb1278 · file-state**
The constructor-scoped fallback branch of the resolver (`resolveSessionHookRegistry(undefined, ctorRegistry) → ctorRegistry`) is covered by the isolated unit test (line 396-398) but never exercised **end-to-end through a provider**: every `describe.each` case supplies `hookRegistry` on `makeConfig()`, so the query-scoped arg is always set and the fallback path is never taken through `provider.query()`. Grounded: `git grep resolveSessionHookRegistry -- '*.test.ts'` resolves only to this file.
*Suggestion:* add a `describe.each` case where `makeProvider()` receives a registry at construction and `makeConfig()` omits `hookRegistry`, asserting the gate still fires.

**nit · high · api-compat · `src/agent/tools/dispatcher.ts:124-131` (doc comment) · ref:3eb1278 · file-state**
The doc comment states "building a dispatcher without threading this result is a compile error." True for **production** code (type-checked), but `tsconfig.json` excludes `src/**/*.test.ts`, so the pre-existing keyless constructions in `integration.test.ts` (lines 42/56/78/98/111) are not caught by `tsc`. Harmless — the dispatcher's value-based consumption (`if (this.hookRegistry)`) tolerates the absent key as `undefined` — so this is a scope nuance of the guarantee, not a defect.
*Suggestion:* none required for merge; optionally narrow the comment to "every production dispatcher construction."

**info · security** — no real credentials: `'sk-ant-oat01-test'` / `'sk-test-key'` are fixtures.
**no issues found** · correctness (resolver is exactly equivalent to the prior `??` in both providers; openai conditional→unconditional change is behavior-preserving — verified against dispatcher consumption) · perf (single `??`, once per query) · observability (no logging removed).

## What was not checked
- Read against branch HEAD `3eb1278`; **all citations verified at that ref** (no diff-only or fabricated citations).
- Ran `tsc --noEmit` (clean); did **not** run the vitest suite — runtime pass/fail of the new contract test is asserted by structure-read only, not execution.
- Did not read `plan-mode-gate.ts` / `createPlanModeGate` internals (gate logic is pre-existing, unchanged by this PR) or the Telegram surface.

## Merge recommendation

**MERGE.** The refactor achieves its stated goal, is behavior-preserving in both providers (verified, not assumed), type-checks cleanly, adds a real CI gate, and ships a cross-provider regression guard for c6892c6. The two surviving findings are both `low`/`nit` and non-blocking — the constructor-scoped-fallback coverage gap is the only one worth a follow-up, and it carries no behavior risk from this change.

---

PR #174, opened 2026-06-16
fix(tui): reset StatusLine + BackgroundStatusBar geometry on SIGWINCH to stop ghost rows

griffinwork40:
## Source
Port of **afk-workshop#719** — https://github.com/griffinwork40/afk-workshop/pull/719
*Moat-scrubbed port, not a 1:1 copy.* Cross-repo (no shared history) → applied via diff + 3-way, then adapted to public structure.

**What it fixes:** on `SIGWINCH`, both `StatusLine` and `BackgroundStatusBar` now reset their geometry **synchronously inside the resize event** (via the existing non-debounced `ResizeBus.subscribeImmediate` channel), before any spinner tick or streaming repaint in the 150 ms debounce window can paint against stale row coordinates — eliminating ghost/stripe rows on terminal resize.

## Ported
- `src/cli/status-line.ts` — `resetGeometry()` on the immediate channel snapshots `lastPaintedRow → preResizePaintedRow` and nulls it; `onResize()` erases the true pre-SIGWINCH row via `preResizePaintedRow ?? lastPaintedRow`; both channels torn down in `stop()`.
- `src/cli/background-status-bar.ts` — `resetGeometry()` zeroes `rowCount`/`lastRepaint` so the spinner `if (rowCount > 0)` guard suppresses mid-window paints until the debounced channel re-seeds geometry; immediate-channel subscribe/unsubscribe wired in `start()`/`stop()`.
- `src/cli/background-status-bar.test.ts` — 3 new tests (R1/R2/R3) covering the reset-on-resize, spinner-suppression-during-debounce, and stop()-teardown behaviors.

## Adapted for public structure (NOT moat drops)
Public's `BackgroundStatusBar` is a registry-only variant — constructor `(registry?, opts?)`, and `BackgroundTaskManager` (`src/cli/commands/interactive/background.ts`) does not exist in public. The private PR was written against the manager-integrated variant, so:
- Dropped the unused `private updateHandler` field (no `updateHandler` in public's bg-bar).
- Adapted the 3 new tests from `new BackgroundStatusBar(manager, registry, …)` to public's `new BackgroundStatusBar(registry, …)`; removed the `BackgroundTaskManager` import/usage from the new test block (the existing public tests already use the `FakeRegistry` + `registry.fireStarted` pattern these now follow).
- One stale comment `from a manager event` → `from a registry event` (manager is private-only).

The fix logic itself is ported faithfully.

## Dropped (moat)
**None.** The diff contains no moat content — it is pure TUI/ANSI resize logic. (`subscribeImmediate` + `__flushResizeBusForTests` already exist in public's `terminal-size.ts`, so no prerequisite port was needed.)

## Verification
- **lint** (`tsc --noEmit`, strict): ✅ pass
- **build** + **build:dist**: ✅ pass (build:dist scrub: 0 files scrubbed)
- **hash pins** (`fix:pins:check`): ✅ all up to date
- **tests**: all port-scoped suites green — `status-line.test.ts` (34), `background-status-bar.test.ts` (19, incl. the 3 new), `status-line-context.test.ts` (7). The full suite shows one failure in `config.test.ts:531`, which is a **pre-existing host-environment config-isolation issue** (`loadConfig()` reading the local machine's real `~/.afk/config` model binding) — it reproduces identically on pristine `main` with this change stashed, and passes in clean CI. **Unrelated to this port.**
- **Deterministic moat guard** (tree + `dist/`): ✅ `MOAT GUARD CLEAN`
- **Adversarial moat audit** (independent re-derivation): ✅ CLEAN

### DIFF
--- a/src/cli/background-status-bar.test.ts
+++ b/src/cli/background-status-bar.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { EventEmitter } from 'node:events';
 import { BackgroundStatusBar } from './background-status-bar.js';
 import type {
@@ -6,6 +6,7 @@ import type {
   BackgroundJob,
   BackgroundRegistryEvents,
 } from '../agent/background-registry.js';
+import { __flushResizeBusForTests } from './terminal-size.js';
 
 // ---------------------------------------------------------------------------
 // Minimal fake BackgroundAgentRegistry for testing — extends the same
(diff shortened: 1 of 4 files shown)


griffinwork40:
## Code Review — PR #174

**Target:** PR #174 *fix(tui): reset StatusLine + BackgroundStatusBar geometry on SIGWINCH to stop ghost rows* · `afk-port/pr-719` → `main`
**Reviewed ref:** `44351efcc6b7d56309108451b99ddd48574b7533` (all citations verified at this SHA)
**Change type:** hotfix · 3 files · +221/-3 → **light regime**
**Summary:** Adds a synchronous (non-debounced) "immediate" resize channel reset to `StatusLine` and `BackgroundStatusBar`, mirroring the existing `CupFrameRenderer` pattern. On SIGWINCH the immediate channel zeroes stale geometry *before* any mid-window repaint (spinner/streaming) can paint against new dimensions with old coordinates.

**Linchpin dependency (not in diff) — verified present & correct:** `ResizeBus.subscribeImmediate` (`terminal-size.ts:108,156`), `ResizeBus.subscribe` (debounced 150ms), and `__flushResizeBusForTests` (`:135`) all exist at the ref. `onStdoutResize()` calls `notifyImmediate()` then `scheduleNotify()` (`terminal-size.ts:68-69`) — so the "immediate handler runs synchronously inside the resize event, before any macrotask" invariant the PR relies on genuinely holds.

---

### Findings (no critical/high)

**1 · medium · correctness/docs · `background-status-bar.ts:149-150` (comment) vs `:204-205` (code) · ref:44351ef · citation-type:file-state**
The `resetGeometry()` invariant comment claims zeroing `rowCount` is *"allowing clearRows() to run before any painting with the new geometry."* This is **false**. After `resetGeometry()` sets `this.rowCount = 0` (`:158`), the next `repaint()` trips the outer guard `if (newRowCount !== this.rowCount)` (`:204`) but the inner guard is `if (this.rowCount > 0) this.clearRows();` (`:205`) — `0 > 0` is false, so **`clearRows()` is not called**. Independently re-derived by a second agent (YES/NO trace agreed). In a codebase whose convention treats `Invariant:`-prefixed comments as authoritative, a load-bearing comment that misstates the control flow is a real defect, not a nit.
- **Evidence:** `:205 — if (this.rowCount > 0) this.clearRows();`
- **Suggestion:** Rewrite the comment to describe the *actual* mechanism (zeroing `rowCount` suppresses the spinner's `if (this.rowCount > 0)` repaint during the debounce window and forces the equality guard to re-seed `rowCount` + fire `onRowCountChange`) — it does **not** invoke `clearRows()`.

**2 · medium · correctness · `background-status-bar.ts` (no old-row erase) vs `status-line.ts:122,135,138-139` · ref:44351ef · citation-type:file-state · confidence: medium [runtime-unverified]**
The two fixes are asymmetric. `StatusLine.onResize()` actively erases its old row: snapshots `preResizePaintedRow = lastPaintedRow` (`:122`), then `rowToErase = preResizePaintedRow ?? lastPaintedRow` (`:135`) → `\x1b[${rowToErase};1H` + `\x1b[2K` (`:138-139`). `BackgroundStatusBar` has **no** equivalent: `clearRows()` is skipped on the recovery repaint (finding 1) *and* always anchors `startRow` to the *current* terminal bottom (`:233`), with no memory of the prior paint location. So on terminal **expand**, old bg-bar rows are not cleared by the bar. Likely mitigated by the compositor's own immediate-channel `resetGeometry` + full-frame repaint (`terminal-compositor.lifecycle.ts:213` → `cup-frame-renderer.ts:276`) overwriting the stranded rows once they fall inside the enlarged scroll region — **but** in a reachable edge case (short main frame + running bg jobs + expand) the main frame may not reach those rows, leaving a residual ghost. **Not a regression** (pre-PR had the same gap). Not verified in a live terminal.
- **Evidence:** `:205 if (this.rowCount > 0) this.clearRows();` + `:233 const startRow = Math.max(1, totalRows - visibleCount - adjacentRows);`
- **Suggestion:** If the bg-bar must self-clear (rather than rely on the compositor), mirror StatusLine: snapshot the pre-resize `startRow`+`rowCount` in `resetGeometry()` and erase those rows on the recovery repaint. Otherwise, document the compositor-cleanup reliance in the comment.

**3 · low · test-coverage · `status-line.test.ts` · ref:44351ef · citation-type:file-state**
The PR adds `preResizePaintedRow` to `StatusLine` specifically to survive a mid-window `repaint()` corrupting `lastPaintedRow` before `onResize()` fires — but adds **no** StatusLine test that injects such a mid-window repaint. The existing `:385` test (*"clears the previously painted row before repainting on a taller resize"*) covers the basic old-row clear and confirms no regression, but not the race the new field guards. The bg-bar side *did* add race tests (R1/R2). Absence claim checked: `git grep` on `status-line.test.ts` found resize coverage at `:313,385` (so the dimension is not untested — only the new race path is).
- **Suggestion:** Add a StatusLine test: paint, emit `resize` (immediate channel snapshots old row), call `repaint()` with new fields (mutates `lastPaintedRow` to new row), then flush the debounced channel and assert the **old** row address is still the one cleared.

### Clean dimensions
- **security** — no issues found — read `background-status-bar.ts` + `status-line.ts` at 44351ef: new code only assigns numeric fields and writes static ANSI; no user input flows into the new paths.
- **api-compat** — no issues found — `resizeImmediateUnsub`, `preResizePaintedRow`, `resetGeometry()` are all private; constructor signatures and public API unchanged; the two non-diff dependency symbols exist at the ref.
- **perf-observability** — no issues found — `resetGeometry()` is O(1); one extra immediate subscriber per class; immediate handlers honor the bus's "O(1), side-effect-only" contract.

### What was not checked (epistemic scope)
- Read against branch HEAD `44351efcc6b7d56309108451b99ddd48574b7533` — all `file:line` citations verified at that ref by the orchestrator and an independent re-derivation agent (Q1–Q5 agreed).
- **Test suite NOT executed** and `tsc --noEmit` NOT run against the PR ref — review is read-only and the worktree sits at `bf96a51`, not the PR head; checking out would modify files. R1/R2/R3 + the StatusLine `:385` test were verified **by manual trace only** (all sound; existing test still passes under the new `rowToErase` logic).
- The finding-2 expand-ghost scenario was **not** reproduced in a live terminal, and the compositor's full erase logic was not read end-to-end — so the "compositor cleans it up" mitigation is a hypothesis, not a confirmation. This is the one genuine residual uncertainty.

---

## MERGE

No critical or high findings (so post-synthesis `/shadow-verify` is not triggered; the central claim was already independently re-derived). The core fix — synchronous geometry invalidation that suppresses stale mid-window repaints — is correct, its non-diff dependency exists and behaves as required, the new tests are sound, and the existing StatusLine resize test still passes. The three findings are non-blocking: an inaccurate (but non-functional) load-bearing comment, a likely-benign structural asymmetry in the bg-bar clear path, and a missing race-specific test. Recommend addressing finding 1 (comment correctness) and finding 2 (decide: self-clear vs. document compositor reliance) in this PR or a fast follow-up.

**Done**
- Reviewed PR #174 (3 files, +221/-3) at ref `44351ef` under the light-regime review pipeline, with one independent verification agent re-deriving the load-bearing control-flow claim.
- Verdict: **MERGE** — 0 critical/high; 2 medium (non-blocking) + 1 low. Full findings with `file:line` citations are in this message (transcript), anchored to verified source at the reviewed ref.
- World unchanged: read-only review — no files, commits, PR state, or comments were modified.
- Deferred/unresolved: finding-2's live-terminal expand-ghost behavior and the compositor-cleanup mitigation were not runtime-verified (read-only + worktree on a different SHA); the test suite and `tsc` were not executed against the PR ref.

---

griffinwork40:
## Review findings resolved — `bc67d4e`

Addressed all three findings from the `/review` pass (ref `44351ef`):

**Finding 1 (medium, docs)** — Rewrote `resetGeometry()`'s invariant comment in `background-status-bar.ts`. The old text claimed zeroing `rowCount` lets `clearRows()` run before painting; it does not (the `if (this.rowCount > 0)` guard at the clear site sees `0` and skips). The comment now describes the real mechanism: (a) spinner-tick suppression during the debounce window, (b) equality-guard re-seed of `rowCount` + `onRowCountChange`, and (c) the *explicit* old-row erase introduced for finding 2.

**Finding 2 (medium, correctness — the only UX-affecting one)** — `BackgroundStatusBar` now self-clears its old rows on resize, mirroring `StatusLine`. It stores the painted start-row (`lastPaintStartRow`), snapshots it (+ `rowCount`) into `preResize{StartRow,RowCount}` synchronously on the immediate SIGWINCH channel, and erases those rows on the recovery repaint via the new `clearPreResizeRows()`. Two correctness points beyond the review's sketch:
- It snapshots a **stored** address, not a recompute from `stream.rows` — by the time the resize handler runs `stream.rows` is already the *new* height, so recomputing would point at the wrong rows (this was a latent bug in the naive approach; the regression test catches it).
- On **shrink** the old address is now below the viewport, where a cursor-move clamps onto the status-line row — those rows are skipped (the terminal reflows them into scrollback, matching the pre-existing zero-`rowCount` reliance).

**Finding 3 (low, test)** — Added a `StatusLine` test for the mid-window-`repaint()` race that `preResizePaintedRow` guards (a repaint landing between SIGWINCH and the debounced `onResize()` must still clear the *true* pre-resize row).

### Verification
- `pnpm lint` (strict `tsc --noEmit`): ✅
- `background-status-bar.test.ts` (21, incl. new R4 expand-erase + R5 shrink-skip) and `status-line.test.ts` (35, incl. new race test): ✅
- New test **R4 confirmed to FAIL without the fix** (`'\x1b[s\x1b[29;1H\x1b[2K…'` lacks `\x1b[23;1H` — the stranded old row), passes with it — i.e. the ghost row is reproduced and closed.
- Full suite: 9354 passed; the sole failure is the **pre-existing host-env `config.test.ts` isolation issue** noted in the PR description, unrelated to this TUI-only diff.

_🤖 via `agent-afk /resolve`_

PR #177, opened 2026-06-16
fix(cli): correct prerelease version compare + debounce auto-update re-trigger

griffinwork40:
**Source:** [afk-workshop#801](https://github.com/griffinwork40/afk-workshop/pull/801) — moat-scrubbed port via diff + 3-way apply (the public repo shares no history with afk-workshop, so this is not a cherry-pick). In this case nothing required scrubbing: the change is core-CLI only and carries no moat.

## Ported
- **`src/cli/update-checker.ts`**
  - Fix `isNewerVersion`: compare numeric semver **cores** by stripping any `-prerelease`/`+build` suffix before `Number()`. Previously `Number('1.10.1-beta')` produced `NaN`, whose comparisons are always false, so a running prerelease never detected its own final release (and a release could wrongly "update" to a prerelease). On equal cores, a final release now outranks its prerelease (`return isPrerelease(current) && !isPrerelease(latest)`).
  - Add `PENDING_TTL_MS` (1h) auto-update **debounce**: `triggerAutoUpdate()` skips spawning a second `npm install -g` when a pending marker already exists; `checkPendingUpdate()` keeps a *fresh* non-matching marker (install still in flight) but clears a *stale* one (older than the TTL → install crashed) so a fresh attempt can run.
- **`src/cli/update-checker.test.ts`** — tests for prerelease ordering, the debounce guard, and fresh-vs-stale marker handling.

## Dropped (and why)
- **Nothing.** The PR touches only core-CLI files that already exist in public; no moat files or hunks were present to drop.

## Verification (on the ported tree at this branch)
- `pnpm install --frozen-lockfile` ✅ · `pnpm lint` (`tsc --noEmit`) ✅ · `pnpm fix:pins:check` ✅ (hash pins up to date) · `pnpm build` ✅ · `pnpm build:dist` ✅
- `pnpm test`: **9381 passed / 12 skipped / 0 failed** (full suite, clean env). Ported file in isolation: **33/33 pass**.
- **Moat guard: CLEAN** — deterministic grep of the tree **and built `dist/`** for the full HARD denylist + structural markers returned 0 hits.
- **Adversarial moat audit: MOAT-CLEAN** — independent re-derivation over tree + dist.

**Do not merge automatically — left for human review.**

### DIFF
--- a/src/cli/update-checker.test.ts
+++ b/src/cli/update-checker.test.ts
@@ -80,6 +80,7 @@ function buildFakeHttp(
 
 import { readFileSync, writeFileSync, existsSync, unlinkSync } from 'fs';
 import { spawn } from 'child_process';
+import { getVersion } from './version.js';
 import {
   checkForUpdates,
   printUpdateBanner,
@@ -94,7 +95,9 @@ const mockWriteFileSync = vi.mocked(writeFileSync);
 const mockExistsSync = vi.mocked(existsSync);
 const mockUnlinkSync = vi.mocked(unlinkSync);
 const mockSpawn = vi.mocked(spawn);
+const mockGetVersion = vi.mocked(getVersion);
 const pendingFile = join(FAKE_CACHE_DIR, 'pending-update.json');
+const ONE_HOUR_MS = 60 * 60 * 1000;
 
 describe('update-checker', () => {
   let stderrSpy: ReturnType<typeof vi.spyOn>;
@@ -103,6 +106,9 @@ describe('update-checker', () => {
     vi.clearAllMocks();
     stderrSpy = vi.spyOn(process.stderr, 'write').mockImplementation(() => true);
     mockExistsSync.mockReturnValue(true);
+    // clearAllMocks() does not reset implementations set via mockReturnValue,
+    // so re-pin the default here to stop per-test prerelease overrides leaking.
+    mockGetVersion.mockReturnValue('1.10.1');
     delete process.env['NO_UPDATE_NOTIFIER'];
     delete process.env['CI'];
   });
@@ -179,6 +185,45 @@ describe('update-checker', () => {
       checkForUpdates('notify');
       expect(mockSpawn).not.toHaveBeenCalled();
     });
+
+    // --- prerelease ordering (exercises the private isNewerVersion) ---------
+    // Before the fix, Number() on a "-beta"/"-rc" suffix produced NaN segments
+    // whose comparisons are always false, so these two cases returned the wrong
+    // boolean (a running prerelease never saw its release; a release wrongly
+    // "updated" to a prerelease).
+
+    it('treats the final release as newer than a running prerelease', () => {
+      mockGetVersion.mockReturnValue('1.10.1-beta.1');
+      mockReadFileSync.mockReturnValue(
+        JSON.stringify({ latestVersion: '1.10.1', checkedAt: Date.now() }),
+      );
+
+      expect(checkForUpdates('notify')).toEqual({
+        currentVersion: '1.10.1-beta.1',
+        latestVersion: '1.10.1',
+      });
+    });
+
+    it('does not offer a prerelease as an update over the running release', () => {
+      mockGetVersion.mockReturnValue('1.10.1');
+      mockReadFileSync.mockReturnValue(
+        JSON.stringify({ latestVersion: '1.10.1-beta.5', checkedAt: Date.now() }),
+      );
+
+      expect(checkForUpdates('notify')).toBeNull();
+    });
+
+    it('compares numeric cores when a prerelease suffix is present', () => {
+      mockGetVersion.mockReturnValue('1.10.1-rc.1');
+      mockReadFileSync.mockReturnValue(
+        JSON.stringify({ latestVersion: '1.11.0', checkedAt: Date.now() }),
+      );
+
+      expect(checkForUpdates('notify')).toEqual({
+        currentVersion: '1.10.1-rc.1',
+        latestVersion: '1.11.0',
+      });
+    });
   });
 
   describe('printUpdateBanner', () => {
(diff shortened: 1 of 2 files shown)


griffinwork40:
---

## Review: PR #177 — fix(cli): correct prerelease version compare + debounce auto-update re-trigger

**Branch:** `afk-port/pr-801 → main` | **+121 / -13** across 2 files

---

### Summary

Two independent bugs fixed in one PR:

1. **`isNewerVersion` prerelease NaN bug** — `'1.10.1-beta.1'.split('.').map(Number)` produced `[1, 10, NaN]`, making all comparisons falsy. Running prereleases never saw their own final release as newer.
2. **`triggerAutoUpdate` double-spawn** — no guard against a second `npm install -g` while one was still in flight, creating a race against the global package.

---

### Dimension findings

#### ✅ Correctness — LGTM with advisory

The core fixes are correct:

- `core(v)` extraction (`v.split(/[-+]/, 1)[0]`) is the right way to strip semver prerelease/build suffixes before numeric comparison.
- `isPrerelease(current) && !isPrerelease(latest)` as the tie-break is semantically correct: equal numeric cores means the final release outranks the prerelease.
- The `writePendingUpdateMarker` before `spawn` ordering is preserved.
- Stale-marker cleanup (`triggeredAt + TTL`) is correctly placed in `checkPendingUpdate`.

**Advisory (non-blocking):** `triggerAutoUpdate`'s `existsSync(pendingPath())` early-return silently depends on `checkPendingUpdate()` having run first to clear stale markers. The comment explains this, but the invariant is purely implicit. If any surface ever calls `triggerAutoUpdate` without `checkPendingUpdate` having run first (daemon mode, test harness, future refactor), stale markers will permanently suppress auto-updates. Recommend elevating to a code comment:

```ts
// Invariant: checkPendingUpdate() must run before this in the same session
// to clear stale markers; otherwise a dead marker blocks all future triggers.
if (existsSync(pendingPath())) return;
```

---

#### ⚠️ Security — Concern (non-blocking)

`triggerAutoUpdate` guards on file *existence* only, not age. A process that continuously re-writes a fresh marker would permanently block auto-updates, because `triggerAutoUpdate` never reads `triggeredAt`. The mitigation in `checkPendingUpdate` (TTL eviction) only fires at session start.

For a local CLI this is a narrow threat surface — no persistent background writer exists in normal operation — so this is low severity. But the fix is straightforward: inline the TTL check in `triggerAutoUpdate` before the early return, so it doesn't require `checkPendingUpdate` to have run:

```ts
if (existsSync(pendingPath())) {
  // Guard the guard: if marker is stale, let this trigger through anyway.
  try {
    const raw = JSON.parse(readFileSync(pendingPath(), 'utf-8')) as Partial<PendingUpdate>;
    const age = Date.now() - (typeof raw.triggeredAt === 'number' ? raw.triggeredAt : 0);
    if (age <= PENDING_TTL_MS) return;
  } catch { return; } // corrupt marker = treat as in-flight
}
```

This would also eliminate the implicit ordering dependency entirely. Optional follow-up rather than blocker.

---

#### ⚠️ Test Coverage — Concerns (two are real, one is a false positive)

**Real gap 1 — `existsSync` mock affects both the directory check and the pending-file check.** `ensureCacheDir()` is called inside `writePendingUpdateMarker()` and uses `existsSync(cacheDir)`. Setting `mockExistsSync.mockReturnValue(false)` in the `triggerAutoUpdate` tests (diff lines adding the `// no in-flight marker` comments) also makes the directory check return `false`, triggering an unintended `mkdirSync` call. Tests pass functionally because `mkdirSync` is mocked, but the semantics are wrong. A path-discriminating implementation would be cleaner:

```ts
mockExistsSync.mockImplementation((p: unknown) =>
  String(p).endsWith('pending-update.json') ? false : true
);
```

**Real gap 2 — `triggeredAt` absent case is untested.** The new code defaults missing `triggeredAt` to `0`, making the marker immediately stale. A test for `{ targetVersion: '1.11.0' }` (no `triggeredAt`) should confirm this falls into the "stale → unlink" path, not into the "fresh → keep" path.

**False positive from reviewer:** The subagent flagged the old test `'clears marker without message when version does not match'` as conflicting — but the diff shows that test was correctly renamed to `'keeps a fresh non-matching marker'` and its expectation inverted to `not.toHaveBeenCalled()`. The two-test split (fresh vs. stale) is sound. No issue here.

**Minor:** `ONE_HOUR_MS = 60 * 60 * 1000` is redeclared in the test file rather than imported from source. If `PENDING_TTL_MS` changes, the stale-marker boundary test (`Date.now() - ONE_HOUR_MS - 60_000`) will silently diverge. Export `PENDING_TTL_MS` from `update-checker.ts` and import it in the test.

---

#### ✅ API Compatibility — LGTM

- `isNewerVersion` is unexported; behavioral change is internal only.
- `triggerAutoUpdate` and `checkPendingUpdate` signatures are unchanged.
- Marker files written by older binaries (no `triggeredAt`) default to age `0`, are treated as immediately stale, and are cleared — backward compatible.
- The behavioral change in `checkPendingUpdate` (keep fresh markers instead of clearing on mismatch) is a safe tightening with no external callers depending on the old eager-clear behavior.

---

### Severity table

| Finding | Dimension | Severity |
|---|---|---|
| Implicit ordering dependency: `checkPendingUpdate` must precede `triggerAutoUpdate` | Correctness | Advisory |
| Marker-refresh suppression: malicious writer can permanently block auto-updates | Security | Low |
| `existsSync` mock covers both dir-check and file-check with same value | Test | Suggestion |
| `triggeredAt`-absent case untested (defaults to stale — correct but not covered) | Test | Suggestion |
| `ONE_HOUR_MS` duplicated in test instead of imported from source | Test | Minor |

---

### Recommendation

**✅ Approve.** The two bug fixes are correct, well-understood, and the tests cover the primary cases. No blocking issues.

Suggested follow-up (any combination is fine as a separate PR):
1. Add `// Invariant:` comment above the `existsSync` guard in `triggerAutoUpdate`.
2. Export `PENDING_TTL_MS` and import it in the test.
3. Add a test for the `triggeredAt`-absent marker.
4. Optionally make the `triggerAutoUpdate` `existsSync` mock path-discriminating.

---

**Done**
- Reviewed PR #177 across four dimensions: correctness, security, test coverage, and API compatibility.
- No blocking issues found; five advisory/suggestion items documented above.
- Merge recommendation: **Approve**.

---

PR #193, opened 2026-06-17
feat(openai-compatible): forward reasoning_effort for o-series models

### LINKED ISSUE #128: openai-compatible: reasoning_effort not forwarded for o-series reasoning models
## Summary
The OpenAI-compatible provider never sends `reasoning_effort` for o-series reasoning models (o1/o3/o4), so reasoning depth can't be controlled and defaults to the server default. The Anthropic provider forwards full thinking config + effort.

## Current behavior
- The streaming `requestBody` (`src/agent/providers/openai-compatible/query.ts` ~L537–585, both Chat Completions and Responses paths) contains no `reasoning_effort` field.
- The provider only handles DeepSeek-R1-style `reasoning_content` echo in multi-turn history (`loop.ts:107–133`); there is no thinking/effort parameter concept on the request side.

## Anthropic parity reference
`anthropic-direct` forwards `ThinkingConfigParam` + budget + effort level: `loop.ts:264–267`, `index.ts:1000–1094` (auto-defaults effort for opus/sonnet-4.x families).

## Impact
- No way to dial o-series reasoning effort (low/medium/high) from AFK; users are stuck with the provider default.
- Effort directly affects latency, cost, and answer quality for reasoning models, so this is a meaningful control to be missing.

## Suggested direction
Resolve a `reasoning_effort` value (config/flag, with a sensible default) and include it in the streaming `requestBody` for detected o-series / reasoning models (reuse the o-series detection already used for the `max_tokens` ↔ `max_completion_tokens` switch in `oneshot.ts`). For the Responses API path, use the `reasoning: { effort }` shape.

---
_Found during an `anthropic-direct` vs `openai-compatible` capability-parity audit. Sibling gaps: max_tokens forwarding, retry layer, image/multimodal input._

griffinwork40:
## Summary

Closes #128.

The `openai-compatible` provider never sent `reasoning_effort` for o-series reasoning models (o1/o3/o4), so reasoning depth could not be controlled from AFK config. The `anthropic-direct` provider already forwarded effort via `output_config.effort`.

### `src/agent/providers/openai-compatible/query.ts`
- **`isOSeriesModel(model)`** — detects o-series models via `/^o[0-9]/` after stripping `provider/` prefixes (OpenRouter-style ids like `openai/o3`)
- **`mapEffortForOpenAI(effort)`** — maps AFK `EffortLevel` → OpenAI values (`low`/`medium`/`high`); `xhigh` and `max` collapse to `high`
- **`resolveReasoningEffort(effort, model)`** — combines both: returns the mapped effort only for o-series models with effort configured
- **Chat Completions path**: attaches `reasoning_effort` to the request body
- **Responses API path**: attaches `reasoning: { effort }` to the request body
- Removed the deferred-features comment about reasoning effort (it is now wired)

### `src/agent/providers/openai-compatible/query.test.ts`
8 new tests:
- ✅ Forwards `reasoning_effort` for o-series models when effort is set
- ✅ Does NOT forward for non-o-series models (e.g. gpt-4o)
- ✅ Does NOT forward when effort is undefined
- ✅ Maps `xhigh` → `high`
- ✅ Maps `max` → `high`
- ✅ Handles OpenRouter-style prefixed model ids (`openai/o3`)
- ✅ Forwards `reasoning.effort` on the Responses API path
- ✅ Does NOT forward on Responses API for non-o-series models

## Test results
- `tsc --noEmit`: ✅ clean
- `vitest run src/agent/providers/openai-compatible/`: ✅ 196/196 passed (including 8 new)

### DIFF
--- a/src/agent/providers/openai-compatible/query.test.ts
+++ b/src/agent/providers/openai-compatible/query.test.ts
@@ -600,6 +600,163 @@ describe('OpenAICompatibleQuery — model-slot resolution in request body', () =
   });
 });
 
+describe('OpenAICompatibleQuery — reasoning_effort for o-series models', () => {
+  it('forwards reasoning_effort for o-series models when effort is set', async () => {
+    pendingChunks = [
+      { choices: [{ delta: { content: 'ok' } }] },
+      { choices: [{ delta: {}, finish_reason: 'stop' }], usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 } },
+    ];
+    const q = buildQueryFromConfig(
+      baseConfig({ model: 'o3', effort: 'high' }),
+      singleInput('hi'),
+    );
+    await collect(q);
+    expect(createCalls).toHaveLength(1);
+    const args = createCalls[0]!.args as Record<string, unknown>;
+    expect(args['reasoning_effort']).toBe('high');
+  });
+
+  it('does NOT forward reasoning_effort for non-o-series models', async () => {
+    pendingChunks = [
+      { choices: [{ delta: { content: 'ok' } }] },
+      { choices: [{ delta: {}, finish_reason: 'stop' }], usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 } },
+    ];
+    const q = buildQueryFromConfig(
+      baseConfig({ model: 'gpt-4o', effort: 'high' }),
+      singleInput('hi'),
+    );
+    await collect(q);
+    const args = createCalls[0]!.args as Record<string, unknown>;
+    expect(args['reasoning_effort']).toBeUndefined();
+  });
+
+  it('does NOT forward reasoning_effort when effort is undefined', async () => {
+    pendingChunks = [
+      { choices: [{ delta: { content: 'ok' } }] },
+      { choices: [{ delta: {}, finish_reason: 'stop' }], usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 } },
+    ];
+    const q = buildQueryFromConfig(
+      baseConfig({ model: 'o3' }),
+      singleInput('hi'),
+    );
+    await collect(q);
+    const args = createCalls[0]!.args as Record<string, unknown>;
+    expect(args['reasoning_effort']).toBeUndefined();
+  });
+
+  it('maps xhigh to high for OpenAI', async () => {
+    pendingChunks = [
+      { choices: [{ delta: { content: 'ok' } }] },
+      { choices: [{ delta: {}, finish_reason: 'stop' }], usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 } },
+    ];
+    const q = buildQueryFromConfig(
+      baseConfig({ model: 'o4-mini', effort: 'xhigh' }),
+      singleInput('hi'),
+    );
+    await collect(q);
+    const args = createCalls[0]!.args as Record<string, unknown>;
+    expect(args['reasoning_effort']).toBe('high');
+  });
+
+  it('maps max to high for OpenAI', async () => {
+    pendingChunks = [
+      { choices: [{ delta: { content: 'ok' } }] },
+      { choices: [{ delta: {}, finish_reason: 'stop' }], usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 } },
+    ];
+    const q = buildQueryFromConfig(
+      baseConfig({ model: 'o1', effort: 'max' }),
+      singleInput('hi'),
+    );
+    await collect(q);
+    const args = createCalls[0]!.args as Record<string, unknown>;
+    expect(args['reasoning_effort']).toBe('high');
+  });
+
+  it('handles OpenRouter-style prefixed model ids (e.g. openai/o3)', async () => {
+    pendingChunks = [
+      { choices: [{ delta: { content: 'ok' } }] },
+      { choices: [{ delta: {}, finish_reason: 'stop' }], usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 } },
+    ];
+    const q = buildQueryFromConfig(
+      baseConfig({ model: 'openai/o3', effort: 'medium' }),
+      singleInput('hi'),
+    );
+    await collect(q);
+    const args = createCalls[0]!.args as Record<string, unknown>;
+    expect(args['reasoning_effort']).toBe('medium');
+  });
+
+  it('forwards reasoning.effort on the Responses API path for o-series models', async () => {
+    // Install a mock that also supports the Responses API
+    const responsesFactory: OpenAIClientFactory = () =>
+      ({
+        chat: {
+          completions: {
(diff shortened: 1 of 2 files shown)


griffinwork40:
I have everything needed. Here is the full review.

---

## PR #193 Review — `feat(openai-compatible): forward reasoning_effort for o-series models`

**Scope:** `query.ts` (+56 lines), `query.test.ts` (+157 lines). Two new API paths wired.

---

### Correctness

**✅ Wiring is correct end-to-end.**  
`AgentConfig.effort` is defined on `config-types.ts:96`, flows into `OpenAICompatibleQueryOptions.config` via `buildQueryFromConfig` (`query.ts:1003`), and is read at the two call sites as `this.opts.config.effort`. The full pass-through is verified.

**✅ `this.currentModel` is the right field.**  
It holds the resolved (post-slot-alias) model id, so `resolveReasoningEffort` always sees the concrete id, not a symbolic alias like `sonnet`.

**✅ Switch exhaustiveness is guaranteed.**  
`EffortLevel` is `'low' | 'medium' | 'high' | 'xhigh' | 'max'` (5 values). All 5 are covered in `mapEffortForOpenAI`. TypeScript strict mode + the narrower return type means a new `EffortLevel` value would cause a compile error at the switch site.

**⚠️ MEDIUM — Regex inconsistency across the codebase.**  
`isOSeriesModel` and `oneshot.ts:94` both use `/^o[0-9]/`. `model-limits.ts:143` uses the more conservative `startsWith('o1') || startsWith('o3') || startsWith('o4')`. Three detectors, two patterns, no shared utility. No bug for any current model, but a future `o2` or `o5` would be caught by the regex detectors but not by model-limits.ts. Recommendation: extract `isOSeriesModel` to a shared location (e.g., `model-capabilities.ts`) and have all three sites import it — not a blocker, but worth tracking.

**⚠️ MEDIUM — Third-party providers can receive unexpected `reasoning_effort`.**  
Any OpenAI-compatible endpoint (vLLM, llama.cpp, MLX, Together.ai) where a user names a local model `o3-custom`, `o3-finetune`, etc. would match `/^o[0-9]/` and get `reasoning_effort` injected. Most providers silently ignore unknown fields. Those with strict schema validation would return a 400, which is surfaced as an error event (not swallowed, not a hang). Real-world risk is low because o-series are closed-source, but it's a behavioral surprise for local-model users. A doc note in the function JSDoc would help.

**ℹ️ INFO — `resolveReasoningEffort` is called twice per request.**  
Both the Responses API block and Chat Completions block independently call `resolveReasoningEffort(this.opts.config.effort, this.currentModel)`. Since `wireMode` is fixed per session, only one branch executes per request, so there's no behavioral issue — just minor visual duplication. Could be hoisted to a single call before the `if (this.wireMode === 'responses')` branch.

---

### API Compatibility

**✅ Field shapes match the OpenAI spec.**  
- Chat Completions: `reasoning_effort: 'low' | 'medium' | 'high'` as a top-level body parameter — correct.  
- Responses API: `reasoning: { effort: 'low' | 'medium' | 'high' }` — correct per the Responses API spec.  
- `xhigh`/`max` → `'high'` downmapping is the right call; OpenAI has no higher tier than `high`.

**✅ No temperature or token-limit conflict.**  
The Chat Completions request body in `query.ts:589-597` is `{ model, messages, stream, stream_options }` with optional `tools`. No `temperature`, no `max_tokens`. There's no field conflict with `reasoning_effort` for o1/o3/o4.

**✅ `openai/o3`-style OpenRouter prefixes handled correctly.**  
`lastIndexOf('/')` strips the prefix before the regex runs — handles both `openai/o3` and `org/team/o3` paths correctly.

**ℹ️ INFO — Three exported helpers widen the module's API surface.**  
`isOSeriesModel`, `mapEffortForOpenAI`, and `resolveReasoningEffort` are now exported from an internal provider file. This is acceptable — they're not part of a public package API — and the exports make the unit tests clean. A follow-on refactor (moving `isOSeriesModel` to `model-capabilities.ts`) would naturally scope these exports.

---

### Test Coverage

**✅ Core paths covered.**  
9 tests across: effort set / unset, o-series / non-o-series, xhigh→high, max→high, OpenRouter prefix, both API paths. The logic table is well exercised.

**⚠️ MEDIUM — `xhigh` and `max` not tested on the Responses API path.**  
The Chat Completions path tests both downmappings (tests 4 and 5). The Responses API test (test 7) uses only `effort: 'low'`. Since both paths call the same `resolveReasoningEffort`, a bug in `mapEffortForOpenAI` wouldn't be caught by the Responses API test. Consider adding one Responses API test for `effort: 'xhigh'` → `reasoning.effort === 'high'`.

**⚠️ LOW — Common o-series variants not tested.**  
Tests use `o3`, `o4-mini`, `o1`, and `openai/o3`. Missing: `o1-preview`, `o3-mini`, `o1-pro`. The regex handles them correctly; this is a test confidence gap, not a logic gap.

**ℹ️ INFO — Inner `installMockClient()` calls in tests 7–8 are redundant.**  
The outer `afterEach(() => { __setOpenAIClientFactory(null); })` resets the factory, and the outer `beforeEach` calls `installMockClient()` before every test. The "Restore the default mock for subsequent tests" comment implies the author was unaware the outer lifecycle handles this. The extra calls are harmless, but the comments should be removed to avoid confusion.

**ℹ️ INFO — Most tests don't assert `createCalls.toHaveLength(1)` before indexing.**  
Tests 2, 3, 4, 5, 6 use `createCalls[0]!.args` without a length assertion. If the API call wasn't made (bug in the future), the failure message would be a `TypeError: Cannot read properties of undefined` rather than the clearer `Expected length: 1, Received: 0`. Minor ergonomics issue.

---

### Security

No concerns. `reasoning_effort` is a non-sensitive request parameter. Provider errors (including 400s from strict schema validators) surface as error events via the existing catch path — not swallowed and not retried. No credential exposure path.

---

### Merge Recommendation

**✅ APPROVE — merge as-is.**

The implementation is correct: field names, placement, exhaustive mapping, and end-to-end wiring are all verified. The two MEDIUM findings are non-blocking:

1. **Regex inconsistency** — worth a follow-on issue to extract `isOSeriesModel` to `model-capabilities.ts` and unify the three detection sites.  
2. **Third-party provider false positive** — worth a JSDoc note on `isOSeriesModel` that callers should be aware the field may be ignored or rejected by non-OpenAI backends.  
3. **Responses API xhigh/max test gap** — worth adding before the suite grows, but the shared `resolveReasoningEffort` makes a bug there unlikely.

None of these block merge.

**Done**
- PR #193 reviewed across correctness, API compatibility, test coverage, and security dimensions
- All 9 new tests verified against the test harness lifecycle (outer beforeEach/afterEach confirmed safe)
- End-to-end wiring from `AgentConfig.effort` → `OpenAICompatibleQueryOptions.config.effort` → `this.opts.config.effort` → request body confirmed at `query.ts:1003` and `index.ts:259-262`
- Two non-blocking MEDIUMs flagged: regex fragmentation across the codebase (three detectors, two patterns) and false-positive o-series detection for custom-named local models
- Verdict: **APPROVE**

---

PR #211, opened 2026-06-18
refactor(cli): centralize committed-band CUP+EL escape into eraseAndPaintRow

griffinwork40:
## What

Extracts the duplicated absolute-position cursor erase-and-paint escape sequence — `` `\x1b[${row};1H\x1b[2K${line ?? ''}` `` (CUP to `row`/col-1 + EL erase-line + optional content) — into a single documented helper `eraseAndPaintRow(row, line?)` in `terminal-compositor.types.ts` (the existing leaf module for pure, side-effect-free helpers — no circular-import risk).

The sequence was hand-written at **11 sites across 4 modules**:

| Module | Sites |
|---|---|
| `terminal-compositor.committed-band-commit.ts` | 5 |
| `terminal-compositor.committed-band-repin.ts` | 3 |
| `terminal-compositor.frame-preserve.ts` | 2 |
| `terminal-compositor.lifecycle.ts` | 1 |

## Why

- A change to the erase/paint primitive previously had to be replicated across 11 copies.
- The load-bearing invariant — *emits no `\n`, so the DECSTBM scroll region is never triggered* — lived only in scattered call-site comments. It is now documented once, on the helper.
- The erase-only sites are exactly the `line === undefined` case, so one helper cleanly covers both shapes (bare-erase and erase-then-paint).

## Behavior preservation

Output bytes are **identical** at every call site — this is a pure de-duplication, no observable behavior change. The intentionally *different* relative-position variant `` `\x1b[2K${l}` `` (EL with no CUP, joined with `\n`) in `committed-band-commit.ts` is deliberately left untouched.

- ✅ `pnpm lint` (`tsc --noEmit`, strict) — clean
- ✅ `terminal-compositor` suite — **288/288 green**, unchanged from the pre-change baseline
- ✅ Full suite — **zero new failures**. The 2 failing files (`config.test.ts`, `anthropic-direct.test.ts`) are a pre-existing model-registry test drift on `main`, confirmed failing identically on a clean checkout with these changes stashed.

## Deferred (intentionally out of scope)

A read-only `/simplify` discovery pass flagged two larger items that are **not** included here, on Sandi-Metz grounds (duplication is cheaper than the wrong abstraction):

1. **Per-module `*Host` interface field duplication** — the ~10 `*Host` interfaces re-declare overlapping fields, but each is documented in-code as the *"narrowest state slice"* its module touches. Collapsing them into a shared base would erode that intentional precision.
2. **`commitAbove` banner / non-banner dual path** — unifying the two branch trees would require introducing a flag parameter.

Both are left as potential follow-ups.

## Diff

5 files, **+32 / −11**. The net `+21` is the helper's invariant-documenting comment; the duplication factor drops **11 → 1**.

---
🤖 Generated with agent-afk (`/simplify` → behavior-preserving apply).

### DIFF
--- a/src/cli/terminal-compositor.committed-band-commit.ts
+++ b/src/cli/terminal-compositor.committed-band-commit.ts
@@ -14,6 +14,7 @@
  */
 
 import type { LogUpdateFn, CompositorScrollRegionGuard } from './terminal-compositor.types.js';
+import { eraseAndPaintRow } from './terminal-compositor.types.js';
 import { hardWrapToWidth } from './wrap.js';
 import { capBandModel, decideCommitMode } from './commit-mode.js';
 
@@ -360,7 +361,7 @@ export function commitAbove(self: CommittedBandHost, text: string): void {
           for (let start = 0; start < genuineOverflow; start += chunkMax) {
             const chunk = overflowRun.slice(start, Math.min(start + chunkMax, genuineOverflow));
             const topWrite = chunk
-              .map((l, i) => `\x1b[${anchorFloor + i};1H\x1b[2K${l ?? ''}`)
+              .map((l, i) => eraseAndPaintRow(anchorFloor + i, l))
               .join('');
             self.stdout.write(`${topWrite}\x1b[${rows};1H${'\n'.repeat(chunk.length)}`);
           }
@@ -470,7 +471,7 @@ export function commitAbove(self: CommittedBandHost, text: string): void {
       for (let i = 0; i < paintedCount; i++) {
         const row = bandTop + i;
         if (row >= newTopRow) break; // Never overwrite the live frame.
-        out += `\x1b[${row};1H\x1b[2K${model[model.length - paintedCount + i] ?? ''}`;
+        out += eraseAndPaintRow(row, model[model.length - paintedCount + i]);
       }
       if (out.length > 0) {
         writeWithGuard(() => {
@@ -572,13 +573,13 @@ export function commitAbove(self: CommittedBandHost, text: string): void {
         if (oldBandTop > 0 && oldBandTop < bandTop) {
           const eraseFloor = Math.max(postScrollFloor, oldBandTop);
           for (let r = eraseFloor; r < bandTop; r++) {
-            out += `\x1b[${r};1H\x1b[2K`;
+            out += eraseAndPaintRow(r);
           }
         }
         for (let i = 0; i < capped.length; i++) {
           const row = bandTop + i;
           if (row >= newTopRow) break; // Never overwrite the live frame.
-          out += `\x1b[${row};1H\x1b[2K${capped[i] ?? ''}`;
+          out += eraseAndPaintRow(row, capped[i]);
         }
       } else {
         // Overflow (block taller than the above-frame region): Phase 1 already
@@ -597,7 +598,7 @@ export function commitAbove(self: CommittedBandHost, text: string): void {
         for (let i = startIdx; i < textLines.length; i++) {
           const row = anchorFloor + (i - startIdx);
           if (row >= newTopRow) break;
-          out += `\x1b[${row};1H\x1b[2K${textLines[i] ?? ''}`;
+          out += eraseAndPaintRow(row, textLines[i]);
         }
       }
       if (out.length > 0) {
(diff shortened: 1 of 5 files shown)


griffinwork40:
Verdict gate passed: **288/288** terminal-compositor tests green. These tests assert on the *exact* CUP+EL byte sequences (e.g. `/\x1b\[22;1H\x1b\[2KABOVE_FRAME/`, `\x1b[20;1H\x1b[2KLINE_ONE`), so their passing is empirical proof of byte-equivalence — stronger evidence than a static read. Combined with my site-by-site re-derivation, lint-clean, and scope confirmation, every load-bearing claim is independently grounded. Diminishing returns reached; emitting synthesis directly rather than re-deriving via forked agents what execution already proved.

---

## Code Review — PR #211

**Target:** PR #211 `refactor(cli): centralize committed-band CUP+EL escape into eraseAndPaintRow` · `simplify/terminal-compositor-dedup` → `main`
**Reviewed ref:** `57f417e3f85e26ca83832ed45b0864acbf820c5f` (= current checkout HEAD; reads/tests resolved against the content under review)
**Triage:** refactor (pure de-duplication) · 5 files · +32 / −11 (43 lines) · regime = **light**
**Stated intent:** PR title + body — extract the 11-site duplicated `\x1b[${row};1H\x1b[2K${line ?? ''}` (CUP+EL) sequence into one documented `eraseAndPaintRow(row, line?)` helper; byte-identical output; leave the relative-position variant untouched. **Spec-compliance assessed.**

### Verification (each load-bearing claim independently grounded)

| Claim | Method | Result |
|---|---|---|
| Byte-identical at all 11 sites | Read every hunk; re-derived `row`/`line` args + `?? ''` absorption per site | ✅ all 11 identical |
| Behavior preserved | Ran `terminal-compositor` suite (asserts literal CUP+EL bytes) | ✅ **288/288 green** |
| Complete site coverage (no un-migrated production sites) | `grep '1H.*2K' src/cli/` | ✅ only production literal left is the helper itself; all other matches are test assertions |
| No circular import | `grep` imports in `types.ts` | ✅ imports only `palette`/`display`; no edge back into the 4 consumer modules |
| Relative variant left untouched | Located `committed-band-commit.ts:380` `\x1b[2K${l ?? ''}` (EL, no CUP, `\n`-joined) | ✅ unchanged, claim accurate |
| Type-safe under strict TS | `pnpm lint` (`tsc --noEmit`, `noUncheckedIndexedAccess`) | ✅ clean |
| PR scope = 5 files | `git log origin/main..HEAD` | ✅ exactly 1 commit; the 39-file `main...HEAD` diff was stale **local** main, not PR scope |

### Findings

**Blocking (critical / high / medium):** none.

**nit · confidence:low · test-coverage · src/cli/terminal-compositor.types.ts:91 · ref:57f417e · citation-type:file-state**
- Finding: the new `eraseAndPaintRow` helper has no dedicated unit test of its own.
- Evidence: `export function eraseAndPaintRow(row: number, line?: string): string { return \`\x1b[${row};1H\x1b[2K${line ?? ''}\`; }`
- Suggestion: optional — a 2-case test (`eraseAndPaintRow(5)` → bare erase; `eraseAndPaintRow(5,'x')` → erase+paint) would pin the contract directly. Non-blocking: the helper is transitively exercised by all 288 byte-asserting tests, which already lock the exact output.

### What was not checked
- **Epistemic basis:** verification was performed directly in this session against ref `57f417e` (the checkout matches the reviewed ref), not via a forked independent agent — justified because the test suite asserts the literal output bytes, making behavior-equivalence a matter of execution (proven) rather than static opinion.
- **Citations:** all verified against branch HEAD `57f417e`; no diff-only or fabricated citations.
- **Stated intent:** PR #211 title + body — spec-compliance assessed; no unmet intent, no scope creep.
- **Not exercised:** live TTY rendering (no real terminal in scope); the suite's headless byte-stream assertions stand in for it. `/shadow-verify` not triggered (no critical/high findings).

### Merge decision: **MERGE**

A textbook behavior-preserving de-duplication: 11→1, output bytes proven identical by an existing suite that asserts the exact escape sequences, no circular-import or scope risk, lint-clean. The single nit (no dedicated helper unit test) is optional and does not gate merge.

*Read-only review complete — no files, PR, or external state modified.*

---

PR #215, opened 2026-06-19
feat(repl): queue multiple type-ahead messages mid-turn

griffinwork40:
## Summary
- REPL mid-turn type-ahead is now a **multi-message FIFO queue** instead of a single editable slot. While the agent is streaming, each Enter commits the typed message to a queue and clears the input, so you can fire off several messages; each drains as its own **sequential turn** (one per turn) — matching the Telegram surface's existing per-chat queue semantics.
- The live input buffer is **decoupled** from committed messages (paste placeholders expanded + attachments snapshotted at commit time), so composing the next message can't corrupt queued ones. Backspace on an empty buffer un-queues the most recent; the indicator shows `[queued]` / `[N queued]`.
- Implementation in `TerminalCompositor`: new `pendingSubmissions: SubmissionPayload[]` (source of truth) + `getPendingCount()`; `queued` is now a maintained mirror. Touch points: commit-on-Enter (`input-dispatch`), FIFO drain that preserves the live buffer (`input-mode`), count-aware indicator (`render`), queue clear on `reset`, and `autocomplete`/`applyEdit` decoupled from the queue.

## Test results
- `pnpm lint` (tsc --noEmit, strict): **clean**.
- Compositor blast-radius suites — `terminal-compositor.test.ts`, `terminal-compositor.autocomplete.test.ts`, `input/input-surface.test.ts`, `interactive-tui.interaction.test.ts`, `input/non-tty.test.ts`: **262 passed**. Includes 3 new multi-message tests (queue 3 → drain FIFO in order; in-progress draft survives a drain; `[N queued]` indicator) and 25 pre-existing tests repurposed to the new contract (no deletions).
- Full `pnpm test`: NOT cleanly green **locally** due to an environmental artifact unrelated to this change — the shared `~/.afk/state/memory/memory.db` was migrated to **schema v3** by a newer afk during the authoring session, and this build reads only v2, so every test that touches the memory store during setup throws `memory.db schema version 3 is newer than this build supports (2)`. Proven environmental: the clean baseline before that migration was 2 failures and the code is byte-identical since. **CI runs against a fresh memory.db and will not hit this.**
- Genuinely pre-existing code-level failures (NOT in this change's blast radius; reproduced on clean base `629ae9c` by stashing this diff): `config.test.ts` "model slots … defaults to the built-in tier bindings"; `anthropic-direct.test.ts` "compact() …" (haiku model-id `claude-haiku-4-5` vs `…-20251001`).

## Verification
Adversarial re-derivation from the committed diff (`git show`), performed inline because the verifier sub-agent could not launch — forked sessions hit the same memory.db v3 contamination:
- **CONFIRM** — Commit-on-Enter: `handleEnter` pushes a `SubmissionPayload`, expands paste + snapshots attachments, then clears the live buffer (`terminal-compositor.input-dispatch.ts`).
- **CONFIRM** — FIFO drain: the `→ idle` flush shifts exactly one payload (oldest-first), maintains the `queued` mirror, and does **not** mutate `self.input` — the in-progress next message survives (`terminal-compositor.input-mode.ts:233-243`).
- **CONFIRM** — No broken legacy path: zero non-test production callers of `getBuffer()`; delivery is `onSubmit`-only.

## Out of scope
- The `memory.db` schema-v3-vs-v2 cross-version test contamination is a pre-existing environment/design tension (a newer afk bumps the shared db; older builds can't read it). Not addressed here.
- The two pre-existing model-config test failures are unrelated and left as-is.

### DIFF
--- a/src/cli/input/input-surface.test.ts
+++ b/src/cli/input/input-surface.test.ts
@@ -379,7 +379,9 @@ describe('InputSurface', () => {
         stdin.emit('keypress', ch, { name: ch, sequence: ch });
       }
       stdin.emit('keypress', undefined, { name: 'return' });
-      expect(compositor.getBuffer()).toEqual({ text: 'queued', queued: true });
+      // New contract: Enter commits 'queued' to the FIFO and clears the live buffer.
+      expect(compositor.getBuffer()).toEqual({ text: '', queued: true });
+      expect(compositor.getPendingCount()).toBe(1);
 
       // Now call readLine. The widened flush invariant fires the just-
       // installed handler synchronously inside setInputMode('idle').
--- a/src/cli/interactive-tui.interaction.test.ts
+++ b/src/cli/interactive-tui.interaction.test.ts
@@ -217,20 +217,20 @@ describe('TerminalCompositor interaction coverage', () => {
     }
     keypress(stdin, undefined, { name: 'return' });
 
-    // Buffer is queued internally AND the visual `[queued]` glyph is
-    // painted in the input row so the user has explicit feedback that
-    // their Enter was registered (Stage 3d's no-glyph variant was a
-    // regression — without the cue, Enter felt like a no-op). The text
-    // remains in the buffer so setInputMode('idle') flushes it at
-    // stream end.
-    expect(compositor.getBuffer()).toEqual({ text: 'queued work', queued: true });
+    // Buffer is committed to the FIFO (live buffer cleared) AND the visual
+    // `[queued]` glyph is painted in the input row so the user has explicit
+    // feedback that their Enter was registered. setInputMode('idle') drains
+    // the FIFO payload as the next turn.
+    expect(compositor.getBuffer()).toEqual({ text: '', queued: true });
+    expect(compositor.getPendingCount()).toBe(1);
     expect(stripAnsi(writes.all())).toContain('[queued]');
 
     writes.clear();
     compositor.commitAbove('final block');
     compositor.setOverlay('follow-up chunk');
 
-    expect(compositor.getBuffer()).toEqual({ text: 'queued work', queued: true });
+    expect(compositor.getBuffer()).toEqual({ text: '', queued: true });
+    expect(compositor.getPendingCount()).toBe(1);
     expect(stripAnsi(writes.all())).toContain('final block');
 
     compositor.disarm();
@@ -259,7 +259,9 @@ describe('TerminalCompositor interaction coverage', () => {
     keypress(stdin, undefined, { name: 'c', ctrl: true });
 
     expect(onCancel).toHaveBeenCalledTimes(1);
-    expect(compositor.getBuffer()).toEqual({ text: 'draft', queued: true });
+    // New contract: Enter committed 'draft' to the FIFO and cleared the live buffer.
+    expect(compositor.getBuffer()).toEqual({ text: '', queued: true });
+    expect(compositor.getPendingCount()).toBe(1);
 
     compositor.disarm();
   });
--- a/src/cli/terminal-compositor.autocomplete.ts
+++ b/src/cli/terminal-compositor.autocomplete.ts
@@ -24,14 +24,17 @@ export const MAX_DROPDOWN_ROWS = 6;
 
 /**
  * Narrowest TerminalCompositor state slice the autocomplete/ghost functions
- * touch. `input`/`queued`/`activeGhost` are mutated; `autocompleteState` is
+ * touch. `input`/`activeGhost` are mutated; `autocompleteState` is
  * mutated in-place (never reassigned, so it stays a `readonly` view).
  * `repaint` is the cross-cluster render callback (a class method on the host).
+ *
+ * Note: applying a completion/ghost does NOT touch the pending-submission
+ * queue — editing the live buffer is independent of committed messages
+ * (commit-on-Enter), so the host needs no `queued`/`pendingSubmissions` slice.
  */
 export interface AutocompleteHost {
   readonly autocompleteState?: AutocompleteState;
   input: InputCoreState;
-  queued: boolean;
   activeGhost: string | null;
   readonly ghostEngine: SuggestEngine | undefined;
   readonly ghostGetContext: (() => SuggestContext) | undefined;
@@ -191,7 +194,6 @@ export function applyDropdownSelection(self: AutocompleteHost): boolean {
   ac.candidates = [];
   ac.viewportStart = 0;
   ac.selectedIndex = 0;
-  self.queued = false;
   updateAutocomplete(self);
   self.repaint();
   return true;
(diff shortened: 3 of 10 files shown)


griffinwork40:
All load-bearing facts are now grounded directly against the reviewed ref `8daba123`. Wave 1.5 verification complete. Synthesizing.

**Key verification results:**
- The **only** production read of `self.queued` is `getBuffer()` (paste.ts:111), and `getBuffer()` has **zero production consumers** (confirmed grep). The flush condition (input-mode.ts:239) and the renderer both read `pendingSubmissions.length`, never the flag. → The mirror-desync cannot produce wrong *production* output today.
- Empty-Enter guard confirmed present at `input-dispatch.ts:524` (fires before the unconditional push; correctly permits attachment-only sends).
- History-on-drain confirmed intact: drained payloads flow `onSubmit → readLine → history.push(text)` (loop-iteration.ts:300); loop-iteration.ts is unchanged by this PR.
- Backspace LIFO-pop correctly gated (only when live-buffer backspace is a no-op) and correctly maintains the mirror.

No critical/high findings → `/shadow-verify` gate not triggered.

---

# Code Review — PR #215 `feat(repl): queue multiple type-ahead messages mid-turn`

**Target:** PR #215 · `main ← afk/20260618-180522-9c6c46` (OPEN) · reviewed ref `8daba123` · feature · 10 files · +487/-219 (706 lines) · **regime: full**
**Stated intent:** PR #215 title+body — **spec-compliance assessed.**

## Summary
Clean, well-tested refactor of REPL mid-turn type-ahead from a single `queued` slot to a `pendingSubmissions` FIFO with commit-on-Enter semantics. Every headline behavior in the intent is implemented and covered by tests. No security, correctness-under-production, or API-compat breaks. The one substantive finding is a **latent invariant violation with no production-visible impact today** (the `queued` mirror desyncs on two paths, but nothing in production reads it). Mergeable.

## Findings (severity-sorted; invariant findings first within tier)

**`low` · confidence high · correctness/invariant · src/cli/terminal-compositor.input-dispatch.ts:494 · ref:8daba123 · citation-type:file-state**
Mid-paste literal-newline path sets `self.queued = false` unconditionally, violating the invariant the PR documents on the field ("`true` IFF `pendingSubmissions` is non-empty — maintained mirror kept in sync at every queue mutation", terminal-compositor.ts:240). Proven reachable by the PR's own test (terminal-compositor.test.ts:1062-1065 asserts `queued:false` while `getPendingCount()===1`). The decoupling cleanup was applied to `applyEdit` and the autocomplete paths but missed this one; its comment still describes the retired single-slot contract.
Evidence:
```
self.input = InputCore.insert(self.input, '\n');
self.queued = false;   // comment: "any buffer mutation invalidates a prior Enter-queued state"
return true;
```
Impact: bounded — `self.queued` is read only by `getBuffer()` (paste.ts:111), which has zero production consumers; the renderer and flush both read `pendingSubmissions.length`. So no live bug, but `getBuffer().queued` (a public method) silently lies after a paste-while-queued, a latent trap for any future consumer.
Suggestion: `self.queued = self.pendingSubmissions.length > 0;`, update the stale comment, and flip the paste test's expectation to `queued: true`.

**`nit` · confidence high · correctness/invariant · src/cli/terminal-compositor.input-dispatch.ts:543 · ref:8daba123 · citation-type:file-state**
Idle-mode immediate-submit path also sets `self.queued = false` unconditionally (same anti-pattern as :494). Effectively unreachable with a non-empty queue — `setInputMode('idle')` drains synchronously, so the surface never accepts an idle-mode keypress while the queue is non-empty — so it's harmless today, but it should mirror the others for invariant consistency.
Suggestion: same one-line fix, `self.queued = self.pendingSubmissions.length > 0;`.

**`nit` · confidence high · correctness · src/cli/terminal-compositor.render.ts:28 · ref:8daba123 · citation-type:file-state**
`RenderHost` still declares `readonly queued: boolean`, but `renderInputLine` no longer reads it — the `[queued]`/`[N queued]` suffix is now driven entirely by `pendingSubmissions.length` (render.ts:52-54). The field is vestigial in this slice.
Suggestion: drop `queued` from `RenderHost` (the host still has it; this only narrows the interface).

**`low` · confidence high · test-coverage · src/cli/terminal-compositor.test.ts · ref:8daba123 · citation-type:diff-context**
New multi-message suite is solid (FIFO order, draft survival, indicator) but three real behaviors are untested: popping 2-of-3 queued messages in sequence (multi-pop LIFO), a live draft surviving N>1 sequential drains, and per-message attachments across multiple queued messages.
Suggestion: add one multi-pop test and one "draft survives 3 drains" test.

## Verified clean (read at ref `8daba123`)
- **security** — no issues found. Attachment snapshots use spread copies (no aliasing of the live array into payloads); paste placeholders expanded then `pasteRegistry.clear()`ed at commit; queue growth bounded by human typing (no DoS).
- **api-compat** — no breaking change with production impact. `getPendingCount()` additive; `AutocompleteHost.queued` removal narrows an interface only `TerminalCompositor` implements (still satisfied); `CompositorInputMode` streaming-Enter semantic change is consumed only via `onSubmit` by the persistent InputSurface — `getBuffer()`/`getAttachments()` have no production consumers, so the "Enter keeps buffer" behavior change reaches no external caller. `[UNVERIFIED: no production importers]` does **not** apply — reachability checked, consumers are internal-only.
- **correctness** — concerns #2 (empty-Enter guard at :524 fires before the push, permits attachment-only sends), #3 (flush shifts before `handler()` → no reentrant double-fire), #4 (drain no longer clears `self.input` → in-progress draft survives), #5 (drained messages still `history.push`ed via readLine, loop-iteration.ts:300, file unchanged by PR), #6 (no attachment aliasing; multi-commit paste expansion correct): all **CONFIRMED** correct.
- **perf-observability** — negligible: `push`/`shift`/`pop` on a queue bounded to a handful of messages.

## Spec-compliance: **MET**
Multi-message FIFO ✓ · commit-on-Enter clears live buffer ✓ · sequential per-turn drain ✓ · buffer decoupled from committed messages ✓ · backspace-on-empty LIFO un-queue ✓ · `[queued]`/`[N queued]` indicator ✓. The "`queued` is a maintained mirror" sub-claim is the only deviation (findings above) — an implementation-detail inconsistency, not an unmet feature.

## What was not checked
- Read against branch HEAD `8daba123` (locally reachable); **all citations verified at that ref** via `git show`/`git grep`.
- **Wave 1/1.5 compose agents returned truncated output** — the structured findings did not reach me intact. Fallback: I re-derived every load-bearing fact directly via `git show`/`git grep` at the ref (the reads above), so these findings rest on my own verified inspection, not the truncated agent output. Coverage gap from the failure: none for the changed cluster; I did not independently second-derive via a separate agent.
- Did **not** run the test suite (PR documents a known `memory.db` schema-v3 environmental failure; CI runs against a fresh db). Did not exercise a live TTY — analysis is static.
- Did not review the Telegram surface (out of scope; this PR is REPL-only).

---

No blocking, critical, or high findings. The feature is correct and well-covered. Recommended (non-gating) follow-up: collapse the two stray `self.queued = false` sites (input-dispatch.ts:494, :543) to `self.queued = self.pendingSubmissions.length > 0` so the public `getBuffer().queued` honors the invariant the PR documents, and adjust the one paste test that currently codifies the desynced state.

(Review is read-only — no files, commits, or PR state were modified.)

---

PR #217, opened 2026-06-19
feat(cli): make usage-limit pause actionable at the keyboard

griffinwork40:
## Problem

When a Claude OAuth subscription usage limit (HTTP 429) is hit mid-turn, the anthropic provider parks the turn in an auto-resume pause (`autoResumeOnUsageLimit ?? true`) and waits for the limit to reset or a keychain account hot-swap. That default is correct for the away-from-keyboard case — but the **at-keyboard** user had no discoverable way to act *now*:

- A line typed during the pause only sets `queued = true` (type-ahead); it does **not** dispatch until the next `readLine()` after the turn ends. So a user who types `/model gpt-5.5` to escape the wait sees it silently do nothing — stranded behind the very wait they are trying to escape.
- The "Usage paused" card never mentioned `/model` (provider switching is now first-class) or `Ctrl+C`/`Esc`, and it advertised `export ANTHROPIC_API_KEY=…` — which is **misleading mid-pause**: the running process already captured its credential, and the hot-swap watches only the keychain OAuth token, so a new env var does nothing for the parked turn.

## Change (A + B + C)

**A — Box copy (`usage-limit-box.ts`).** Advertise the real escape hatches: switch model/provider via `/model`, switch account via `claude login` (picked up live by the existing keychain hot-swap), or `Esc` to stop waiting. Drop the misleading live `ANTHROPIC_API_KEY` bullet.

**B — Enter-during-pause = interrupt + flush.** Reuse the existing ESC soft-stop machinery. A new `paused` compositor flag (toggled by `setPausedState` on the `paused`/`resumed` events in the turn handler) makes a submitted line *during the pause* fire a new `onPauseInterrupt` handler that ends the auto-resume wait via `session.interrupt()`. The just-queued buffer then flushes as the next turn at the next `readLine()`. Type `/model gpt-5.5` + Enter during a pause and you land on the new provider — one gesture. Normal-streaming type-ahead queuing is unchanged.

**C — Interactive picker.** On a TTY with `autoResume === true`, show an arrow-key picker (via the existing `enterPickerMode`/`runPicker` infra) *in place of* the passive card — the two are **mutually exclusive**, so the options never render twice:

```
⏳ Usage limit reached.  Auto-resumes at HH:MM.
Tip: run `claude login` in another terminal to switch account — this turn resumes on it automatically.

▸ Keep waiting — auto-resumes at HH:MM        [default]
  Switch model / provider  (type /model after)
  Stop waiting
```

- **Keep waiting** (default / Esc): no-op — auto-resume + B's Enter-during-pause both stay live with `paused === true`.
- **Switch model / Stop waiting**: end the wait via B's pause-interrupt path (`session.interrupt()`); "switch model" also prints a `/model <name>` hint (no buffer pre-fill — the compositor exposes no public buffer-set API).
- "Switch account" is an out-of-band action (`claude login` during the wait → hot-swap), so it lives as a header tip rather than a menu row.
- Non-TTY surfaces and `autoResume === false` keep the existing passive card path unchanged.

Auto-resume + background keychain hot-swap are left **fully intact** — this is purely additive agency for the watching user.

## Why reuse the ESC path (B and C)

ESC already aborts the pause today (the soft-stop handler stays live through the wait and calls `session.interrupt()`). B and C both point the same proven teardown at new triggers (an Enter-during-pause; a picker selection), gated on the `paused` flag so only the pause changes behavior. The interrupt ends the turn without a `done` event, so `recordTurn`/`onTurnComplete` are naturally suppressed — same as soft-stop. The picker's `AbortController` is aborted on `resumed`, soft-stop/error, and in `finally`, so `exitPickerMode` fires exactly once on every exit path.

## Tests

- **terminal-compositor (+3):** Enter+`paused` fires `onPauseInterrupt` **and** queues; Enter+not-`paused` only queues (type-ahead regression guard); plus `paused` reset coverage.
- **turn-handler (+10):** paused→interrupt→suppressed turn; `resumed` clears the flag; picker shown on TTY+autoResume; **card suppressed when picker shows** (mutual-exclusivity guard); keep-waiting leaves `paused === true` so B still works; switch-model / stop-waiting fire the interrupt; auto-resume race teardown; non-TTY and `autoResume=false` keep the passive path.
- **box copy:** assertions updated; the misleading `ANTHROPIC_API_KEY` bullet asserted absent from the live card.
- `pnpm lint` (strict `tsc --noEmit`): clean. Affected suites all green (terminal-compositor, turn-handler, picker, usage-limit-box, render).

## Scope / follow-ups

- CLI only (`src/cli/`). **Telegram parity** for the pause-interrupt is a tracked follow-up, not in this PR (the picker is TTY-specific; Telegram would interrupt the paused turn on a new inbound message instead).

### DIFF
--- a/src/cli/commands/interactive/loop-iteration.ts
+++ b/src/cli/commands/interactive/loop-iteration.ts
@@ -475,6 +475,8 @@ export async function runInputLoop(
         getCompositor: () => surface.getCompositor(),
         setBackgroundHandler: (handler) => surface.setBackgroundHandler(handler),
         setSoftStopHandler: installSoftStop,
+        setPausedState: (paused) => surface.setPausedState(paused),
+        setPauseInterruptHandler: (handler) => surface.setPauseInterruptHandler(handler),
         async onContextProgress() {
           await ctx.contextSampler.refresh();
           ctx.statusLine.repaint(formatStatusFields(ctx.stats, ctx.contextSampler));
--- a/src/cli/commands/interactive/shared.ts
+++ b/src/cli/commands/interactive/shared.ts
@@ -564,6 +564,27 @@ export interface TurnHandles {
    */
   setSoftStopHandler?(handler: (() => void) | null): void;
   /**
+   * Toggle the compositor's `paused` flag for the duration of a usage-limit
+   * pause (set true on the `paused` provider event, false on `resumed` / turn
+   * end). While true, a submitted line fires the pause-interrupt handler below
+   * instead of merely queuing — ending the auto-resume wait so the queued
+   * command runs as the next turn.
+   *
+   * Wired by the REPL to `surface.setPausedState(...)`. Absent on non-REPL
+   * callers and a no-op on non-TTY surfaces — the turn handler treats this as
+   * a no-op when undefined.
+   */
+  setPausedState?(paused: boolean): void;
+  /**
+   * Install/clear the per-turn pause-interrupt handler on the surface's
+   * persistent compositor. The turn handler sets a `session.interrupt()`
+   * closure at turn start and clears it (`null`) at turn end.
+   *
+   * Wired by the REPL to `surface.setPauseInterruptHandler(...)`. Absent on
+   * non-REPL callers and non-TTY surfaces — treated as a no-op when undefined.
+   */
+  setPauseInterruptHandler?(handler: (() => void) | null): void;
+  /**
    * Fired mid-turn on tool_result events so the REPL can refresh the
    * context sampler and repaint the status line with live context usage.
    * The turn handler throttles calls internally (min interval) — callers
--- a/src/cli/commands/interactive/turn-handler.test.ts
+++ b/src/cli/commands/interactive/turn-handler.test.ts
@@ -510,6 +510,374 @@ describe('runTurn — pause-interrupt (submit during usage-limit pause)', () => 
   });
 });
 
+// ---------------------------------------------------------------------------
+// Helpers for the picker tests (C — interactive usage-limit picker)
+// ---------------------------------------------------------------------------
+
+import type { PickerController } from '../../terminal-compositor.js';
+
+/**
+ * Minimal fake compositor that implements the PickerHost interface used by
+ * `runPicker` and the subset of `TerminalCompositor` called by `runTurn`
+ * (only `commitAbove` is needed besides the picker surface).
+ */
+function makeFakeCompositor() {
+  let controller: PickerController | null = null;
+  let enteredCount = 0;
+  let exitedCount = 0;
+
+  const comp = {
+    // PickerHost surface
+    enterPickerMode(c: PickerController): void {
+      controller = c;
+      enteredCount++;
+    },
+    exitPickerMode(): void {
+      controller = null;
+      exitedCount++;
+    },
+    repaintPicker(): void { /* no-op in unit tests */ },
+
+    // Called by runTurn at line 229 (blank separator before the stream event)
+    commitAbove(_line: string): void { /* swallow */ },
+
+    // Test-only accessors
+    get capturedController(): PickerController | null { return controller; },
+    get enteredCount(): number { return enteredCount; },
+    get exitedCount(): number { return exitedCount; },
+
+    /** Simulate the user selecting item at index `idx` by dispatching Return. */
+    selectIndex(idx: number): void {
+      if (!controller) throw new Error('No active picker controller');
+      // Navigate to the desired index by pressing Down from index 0.
+      for (let i = 0; i < idx; i++) {
+        controller.onKey(undefined, { name: 'down' });
+      }
+      controller.onKey(undefined, { name: 'return' });
+    },
+
+    /** Simulate pressing Esc in the picker. */
+    pressEscape(): void {
+      if (!controller) throw new Error('No active picker controller');
+      controller.onKey(undefined, { name: 'escape' });
+    },
+  };
+  return comp;
+}
+
+describe('runTurn — usage-limit picker (C)', () => {
+  // The interactive picker is shown when:
+  //   - A TTY compositor is armed (getCompositor returns non-null)
+  //   - The paused event has autoResume === true
+  // On autoResume=false or non-TTY, the passive card path is kept unchanged.
+  //
+  // The picker uses runPicker (src/cli/render/picker.ts) which drives
+  // enterPickerMode / exitPickerMode on the compositor. Selections are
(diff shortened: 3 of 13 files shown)


griffinwork40:
Verification complete. The `medium` finding is **confirmed**: the `void runPicker(...).then(...)` chain at `turn-handler.ts:436–471` has no `.catch()`, while its inner `session.interrupt()` *does* (`.catch()` at the end) and the parallel non-picker pause-interrupt path also catches — so the gap is asymmetric with the file's own convention. The one synchronous throw site inside the callback is `completionWriter.fn(...)` (the `/model` hint), which resolves to `compositor.commitAbove` when armed (`turn-handler.ts:195`) and can throw mid-teardown. Process-level `unhandledRejection` handlers exist only in `daemon.ts` — **not** the REPL path where this code runs — so under Node's default `throw` mode an unhandled rejection here degrades the REPL. Real but low-probability → `medium` stands. No critical/high findings, so `/shadow-verify` is not triggered.

Synthesizing (Wave 2, local).

---

# Code Review — PR #217

**Target:** PR #217 `feat(cli): make usage-limit pause actionable at the keyboard` · `main ← afk/20260618-163800-fba9fe` · reviewed ref **`de2d3be`** · +794/−18, 13 files · regime **full** · state OPEN, mergeable.

**Summary:** Three additive sub-changes make a Claude usage-limit (429) auto-resume pause actionable at the keyboard — (A) box-copy rewrite, (B) Enter-during-pause → `session.interrupt()` + flush queued buffer, (C) interactive picker that replaces the passive card on TTY+autoResume. Reuses the proven ESC soft-stop teardown; auto-resume + keychain hot-swap left intact.

## Findings (sorted: medium → nit → low; none security/critical/high)

**1 · medium · confidence medium · perf-observability/robustness · `turn-handler.ts:436–471` · ref:de2d3be · file-state**
The fire-and-forget `void runPicker(...).then((result) => {…})` chain has no `.catch()`; the callback synchronously calls `(completionWriter ?? {fn: console.log}).fn(…)` (the `/model` hint), which resolves to `compositor.commitAbove` (`turn-handler.ts:195`) and can throw during compositor teardown — producing an `UnhandledPromiseRejection` on the REPL path, which has no process-level handler (only `daemon.ts:351` does).
> evidence:
> `void runPicker(borrowedCompositor, {…}).then((result) => {`
> `  …(completionWriter ?? { fn: console.log }).fn(palette.dim('  Hint: type /model …'));`
> `  session.interrupt().catch((err) => {…});  // inner catch present`
> `});                                          // ← no outer .catch()`
> suggestion: append `.catch((err) => { if (isDebugEnabled()) console.error('  ' + palette.error('picker promise rejected:'), err); })` to the `.then()` chain — matching the `session.interrupt().catch(...)` convention already used three lines above.

**2 · nit · confidence high · api-compat · `terminal-compositor.input-dispatch.ts:100` · ref:de2d3be · file-state**
`KeyDispatchHost` gains non-optional `paused: boolean`, but it is an `@internal` structural type whose sole production implementer is `TerminalCompositor` (grep: zero external implementers) — no external break. `[UNVERIFIED: no production external implementers]`
> evidence: `paused: boolean;` · suggestion: none required; internal-only widening.

**3 · nit · confidence high · api-compat · `terminal-compositor.reset.ts:34` · ref:de2d3be · file-state**
`ResetStateHost` gains non-optional `paused: boolean`; only `TerminalCompositor.resetState()` passes itself (grep: zero external implementers) — no external break. `[UNVERIFIED: no production external implementers]`
> evidence: `paused: boolean;` · suggestion: none required; internal-only widening.

**4 · low · confidence medium · test-coverage · `turn-handler.test.ts` · ref:de2d3be · file-state**
No test covers the picker `.then()` resolving with a non-null selection while `resumed` is in-flight (the abort-vs-selection race). Outcome is benign (the `if (!result) return` guard plus `pauseInterruptRequested` conservatively drops the replayed turn), but the specific ordering is unverified. `absence-checked: git grep -n "doneFired.*then|picker.*after.*done" …test.ts → 0 matches`
> suggestion: add a test where the `paused`→picker selection microtask runs before the `resumed` abort lands, asserting no double recordTurn.

**5 · low · confidence medium · test-coverage · `turn-handler.test.ts` · ref:de2d3be · file-state**
No test asserts picker teardown (`exitedCount ≥ 1`) when a turn ends via an `error` event with the picker open (TTY + autoResume=true); the existing "error after paused" test uses `autoResume:false` (no picker). The `finally` block handles it correctly (`pickerRef.abort?.abort()`), so this is an assertion gap, not a defect. `absence-checked: git grep -n "autoResume.*true.*error" …test.ts → 0 matches`
> suggestion: add a `paused(autoResume:true,TTY) → error` test asserting `comp.exitedCount ≥ 1`.

## Cross-agent conflicts
None — the security/api-compat and correctness/spec/test agents agreed on every overlapping point (both independently cleared the box-copy security question and the `paused`-flag lifecycle).

## Correctness & spec verdict (cleared)
- **Correctness:** all six scrutinized risk points cleared on a full-file read at the ref — `pickerRef` object pattern (no leak), `.then()` race (benign), `recordTurn` suppression guard, `paused` lifecycle (no cross-turn leak), B+C non-double-fire (picker-mode keys bypass `handleEnter`), and Enter dispatch ordering (queue-before-interrupt, empty-buffer suppressed).
- **Spec-compliance (assessed against PR title+body):** A, B, C all **met**; no scope creep; Telegram parity correctly deferred (not a gap).

## Merge decision: **MERGE**

No blocking, critical, or high findings. The change is purely additive (optional interface members on the public surface; the two non-optional widenings are internal-only), spec-compliant on all three sub-changes, security-clean, and well-tested (+460 test lines covering B's queue+interrupt, C's picker show/suppress/keep/stop/switch/race-teardown, and the box-copy assertions). The single `medium` is a one-line defensive `.catch()` worth adding before or just after merge; the two `low`s are benign assertion gaps.

## What was not checked
- **Refs:** All re-reads performed against branch HEAD **`de2d3be`** — citations verified at that ref (file-state reads via `git show <ref>:<path>`).
- **Stated intent:** PR #217 title+body — **spec-compliance assessed**.
- **Tests not run:** I did not execute `pnpm test` / `pnpm lint`; the PR body claims both green, unverified here (read-only review).
- **Topical gaps:** Did not review the Telegram surface (`src/telegram/`) — explicitly out of scope; did not read `runPicker` internals (`render/picker.ts`) or `retry-layer.ts` replay semantics (trusted as unchanged context); no live-TTY runtime integration test.

---

PR #218, opened 2026-06-19
fix(subagent): make forked sub-agents non-interactive

griffinwork40:
## Problem

A forked sub-agent shares the **process-wide** elicitation router with its parent. So a child that calls `ask_question`, trips the path-approval hook, or receives an MCP elicitation reaches the operator at the **parent's** REPL/Telegram surface — interleaved into the parent's turn with no attribution. Foreground forks did this by design (`denyElicitations: false`), and `ask_question` + path-approval leaked for **both** foreground and background.

A sub-agent has no human relationship of its own; it should return Blocked/Asking findings to its **parent**, which owns the operator surface.

## Change

Make every fork non-interactive at a single chokepoint, plus close the three elicitation channels:

- **`subagent.ts`** child config: default `isNonInteractive: true` (the provider then strips `ask_question`) and default `onElicitation: DENY_ELICITATION`. A caller may opt a fork back in with `isNonInteractive: false` / `denyElicitations: false`.
- **`subagent-executor.ts`**: `denyElicitations: true` for **both** modes (was background-only). *Behavior change:* foreground forks no longer route MCP elicitations to the operator.
- **`path-approval-hook.ts`**: auto-deny out-of-root access when `context.parentSessionId` is set. Sub-agents inherit the parent's grants, so only a **genuinely new** out-of-root path blocks; the child reports the requirement to its parent. Mirrors the `parentSessionId` self-skip already used by the memory + plan-mode hooks.
- **`config-types.ts`**: updated the now-stale `isNonInteractive` doc.

## Why it's safe (no flow depends on foreground-sub-agent elicitation)

- Skill-dispatch forks **already** strip `ask_question` via `isSkillDispatch` (mint, diagnose, all skill phases) — this only closes the `agent`-tool foreground gap.
- DAG/farm/compose/skill-phase forks had **no** `denyElicitations` guard at all — making all forks deny is a latent-silent-hang **fix**, not a regression.
- `get-started` (the only `ask_question`-heavy skill) runs in-context (`load`), never forked, and already self-guards on non-interactive surfaces.

## Verification

- `pnpm lint` clean.
- 126 tests pass across the 3 affected files (incl. new fork-default `isNonInteractive` + path-approval sub-agent auto-deny coverage); 75 provider strip-mechanism tests pass.
- Note: the full `pnpm test` is currently unreliable in my local env due to a pre-existing `memory.db` schema v3 vs build v2 mismatch (unrelated to this change); CI runs against a clean state dir.

First of two PRs. Follow-up: make `bypassPermissions` actually suppress path-approval prompts (grant-all).

### DIFF
--- a/src/agent/subagent.test.ts
+++ b/src/agent/subagent.test.ts
@@ -380,10 +380,11 @@ describe('SubagentManager', () => {
     });
   });
 
-  // External constraint: background subagents have no surface to serve
-  // elicitations. denyElicitations installs DENY_ELICITATION on the child
-  // config so the SDK auto-declines instead of stalling. Foreground
-  // subagents are unaffected.
+  // External constraint: a forked sub-agent has no human relationship of its
+  // own, so every fork is non-interactive by default — DENY_ELICITATION is
+  // installed on the child config so the SDK auto-declines MCP elicitations
+  // instead of routing them to the operator. A caller opts a fork back in with
+  // denyElicitations: false (then the parent's handler, if any, propagates).
   describe('denyElicitations / DENY_ELICITATION', () => {
     it('DENY_ELICITATION resolves to { action: "decline" } regardless of input', async () => {
       const { DENY_ELICITATION } = await import('./subagent.js');
@@ -407,7 +408,7 @@ describe('SubagentManager', () => {
       expect(shared.lastConfig?.onElicitation).toBe(DENY_ELICITATION);
     });
 
-    it('denyElicitations: false leaves onElicitation untouched (foreground default)', async () => {
+    it('denyElicitations: false (explicit opt-out) leaves onElicitation untouched', async () => {
       shared.lastConfig = null;
       const mgr = new SubagentManager();
       await mgr.forkSubagent({
@@ -415,7 +416,8 @@ describe('SubagentManager', () => {
         config: { model: 'sonnet' },
         denyElicitations: false,
       });
-      // No handler explicitly set on parent config → child receives none.
+      // Explicit opt-out: no DENY_ELICITATION installed. No handler on the
+      // parent config → child receives none.
       expect(shared.lastConfig?.onElicitation).toBeUndefined();
     });
 
@@ -435,18 +437,57 @@ describe('SubagentManager', () => {
       expect(shared.lastConfig?.onElicitation).toBe(DENY_ELICITATION);
     });
 
-    it('denyElicitations omitted: parent onElicitation propagates to child (transitive)', async () => {
+    it('denyElicitations omitted: DENY_ELICITATION installed by default (overrides parent handler)', async () => {
+      const { DENY_ELICITATION } = await import('./subagent.js');
       const parentHandler = vi.fn();
       shared.lastConfig = null;
       const mgr = new SubagentManager();
       await mgr.forkSubagent({
         parent: { sessionId: 'p' },
         config: { model: 'sonnet', onElicitation: parentHandler as any },
-        // denyElicitations omitted — inheritance via ...options.config spread
+        // denyElicitations omitted — forks now default to non-interactive, so
+        // DENY_ELICITATION is installed unless the caller opts out with false.
       });
-      // Transitive propagation: a bg grandparent's DENY_ELICITATION (or any
-      // other handler) flows through unless explicitly overridden.
+      // Default-deny: a fork is non-interactive by default, so even a
+      // parent-provided handler is overridden. The parent owns the operator
+      // relationship; the sub-agent reports findings back rather than eliciting.
+      expect(shared.lastConfig?.onElicitation).toBe(DENY_ELICITATION);
+    });
+
+    it('denyElicitations: false (opt-out) lets a parent onElicitation propagate to the child', async () => {
+      const parentHandler = vi.fn();
+      shared.lastConfig = null;
+      const mgr = new SubagentManager();
+      await mgr.forkSubagent({
+        parent: { sessionId: 'p' },
+        config: { model: 'sonnet', onElicitation: parentHandler as any },
+        denyElicitations: false,
+      });
+      // Explicit opt-out preserves transitive inheritance via the
+      // ...options.config spread (e.g. a parent that itself has a live surface).
       expect(shared.lastConfig?.onElicitation).toBe(parentHandler);
+    });
+
+    it('forked sub-agents are non-interactive by default (isNonInteractive omitted → true)', async () => {
+      shared.lastConfig = null;
+      const mgr = new SubagentManager();
+      await mgr.forkSubagent({
+        parent: { sessionId: 'p' },
+        config: { model: 'sonnet' },
+      });
+      // The provider strips `ask_question` from a non-interactive toolset, so a
+      // fork cannot prompt the operator via that tool either.
+      expect(shared.lastConfig?.isNonInteractive).toBe(true);
+    });
+
+    it('isNonInteractive: false on the fork config is respected (caller opt-in to interactive)', async () => {
+      shared.lastConfig = null;
+      const mgr = new SubagentManager();
+      await mgr.forkSubagent({
+        parent: { sessionId: 'p' },
+        config: { model: 'sonnet', isNonInteractive: false },
+      });
+      expect(shared.lastConfig?.isNonInteractive).toBe(false);
     });
   });
 
--- a/src/agent/subagent.ts
+++ b/src/agent/subagent.ts
@@ -381,6 +381,15 @@ export class SubagentManager {
       abortSignal: childController.signal,
       apiKey: options.config.apiKey || this.parentApiKey,
       baseUrl: options.config.baseUrl ?? this.parentBaseUrl,
+      // External constraint: a forked sub-agent has no human relationship of its
+      // own — it returns findings (including Blocked/Asking) to its PARENT, which
+      // owns the operator surface. Mark every fork non-interactive by default so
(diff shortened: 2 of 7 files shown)


griffinwork40:
Both agents converge; I've independently resolved the one CONFLICT. Here's the decisive verification:

**`onElicitation` consumption (the contested claim):** The tree-wide grep at the ref proves `config.onElicitation` is **read by no live code path**. The `buildQueryOptions` shim the router comments reference **has no definition at this ref** (all 10 `buildQueryOptions` hits are stale doc-comments — a legacy of the pre–Messages-API era). The MCP client registers **no** elicitation handler and `client.ts:20,63` deliberately omits the `elicitation` capability. So:
- **Agent 2 is correct**, Agent 1's parenthetical ("`onElicitation` covers the MCP SDK path") is imprecise.
- The MCP channel **is** closed — but by the pre-existing protocol-layer capability omission, **not** by the PR's `onElicitation: DENY_ELICITATION` install (which is currently dead-but-forward-compat code).

No critical/high findings → `/shadow-verify` not required (and I re-derived the sole contested claim myself above). One Agent-2 nit (config-types doc) is **refuted by the diff** — the hunk already says forks default to `true` — so I drop it. Synthesizing.

---

# Code Review — PR #218 `fix(subagent): make forked sub-agents non-interactive`

**Target:** PR #218 · `fix/subagents-non-interactive` → `main` · reviewed ref `96d7b3a` · +173/−27, 7 files (4 src, 3 test)
**Regime:** light by line count (200 LOC); ran full 2-agent wave + self-verification due to permission-boundary sensitivity.
**Stated intent:** PR title+body — **spec-compliance assessed.**

## Verdict summary
The change achieves its stated goal: a forked sub-agent can no longer reach the operator. All three reachability paths are closed at runtime — `ask_question` is stripped (`isNonInteractive` default `true`), out-of-root path-approval auto-denies (`parentSessionId` guard), and MCP elicitation is unreachable. Wiring is verified end-to-end; test coverage is comprehensive; no in-tree caller relies on the old interactive-foreground default. **No blocking findings.**

### MEDIUM · confidence HIGH · spec-compliance/maintainability · `src/agent/subagent.ts:420–433` · ref:96d7b3a — *NON-BLOCKING*
The comment claims `onElicitation: DENY_ELICITATION` "closes the MCP elicitation channel" and "unifies the three channels," but **`config.onElicitation` is consumed by no live code path at this ref** — so the install is a no-op today. The MCP channel is actually closed by the pre-existing protocol-layer omission (`src/agent/mcp/client.ts:20,63,66`: `elicitation` left out of `CLIENT_CAPABILITIES`).
- Evidence: `git grep "onElicitation" 96d7b3a -- 'src/**/*.ts' ':!*.test.ts'` → only the type def (`config-types.ts:371`), the set-site (`subagent.ts:433`), and doc-comments in `elicitation-router.ts`; the `buildQueryOptions` shim those comments cite **has no definition** at this ref. MCP grep for `elicitInput`/`setRequestHandler` → zero.
- Why it's MEDIUM not HIGH: the functional goal is met (forks are non-interactive; MCP is genuinely closed), so there is no reachable wrong behavior. The risk is a **false maintenance contract** — a future dev re-advertising the elicitation capability (the documented "PR 2" plan, `client.ts:20`) would believe `DENY_ELICITATION` already protects forks, when `config.onElicitation` would also need wiring to the MCP client. The tests reinforce the illusion: `subagent.test.ts` asserts the config *field* is set, never that it takes runtime effect.
- Suggestion: reword the comment to "forward-compat: MCP elicitation is currently protocol-blocked at `client.ts:66` (`CLIENT_CAPABILITIES` omits `elicitation`); this handler takes effect only once the capability is advertised **and** `config.onElicitation` is wired into the MCP client." Keep the install — it's harmless and forward-correct.

### LOW · confidence HIGH · test-coverage · `src/agent/tools/hooks/path-approval-hook.test.ts:184–206` · ref:96d7b3a — *NON-BLOCKING*
The in-root "not blocked for a sub-agent" test asserts `decision === {}` but does not pin *which* branch returned it — a future reorder placing the `parentSessionId` guard before the `!result.restricted` early-return would block in-root paths for forks and this test would still pass.
- Suggestion: add a case combining `parentSessionId` set + a restricted (out-of-root) path that is in the parent's session allow-cache, or assert the guard ordering directly, so the `!restricted`-first invariant is locked.

### NIT · confidence HIGH · doc-accuracy · `src/agent/subagent.ts:104` · ref:96d7b3a
The `denyElicitations` JSDoc still reads "When true, overrides `config.onElicitation`…" — describing the **old opt-in** semantics. The new behavior is deny-by-default (omitted **or** `true` → deny; only `false` opts out). This line is outside the diff hunks, so it was missed.
- Suggestion: "Deny MCP elicitations for this fork by default; pass `false` to opt out and propagate the parent's handler."

### NIT · confidence MEDIUM · doc-accuracy · `src/agent/subagent.ts:107` · ref:96d7b3a
Propagation example phrasing ("overridden by a deeper `denyElicitations: true`") is slightly misleading: `denyElicitations` is a `ForkSubagentOptions` field, not inherited via `AgentConfig`; grandchildren inherit the installed `onElicitation` handler through the `...options.config` spread, not the `denyElicitations` flag itself.

## Verified correct (no issues)
- **`denyElicitations` inversion** (`subagent.ts:433`): truth table correct — omitted→deny, `true`→deny, `false`→propagate parent handler via the earlier `...options.config` spread (deny ternary is the last writer). `DENY_ELICITATION` returns `{action:'decline'}` (`subagent.ts:49`).
- **`isNonInteractive` flow** is runtime-effective end-to-end: `forkSubagent` sets it on `childConfig` → `new AgentSession(childConfig)` → both providers strip `ask_question` (`anthropic-direct/index.ts:688–689`, `openai-compatible/index.ts:349–350`). Existing test pins the config (`subagent.test.ts`).
- **path-approval guard ordering** (`path-approval-hook.ts:209` then the new guard): in-root paths hit `if (!result.restricted) return {}` first, so inherited parent grants are never blocked for the child. Only genuinely-new out-of-root paths block.
- **Regression / safety claim**: `grep "denyElicitations.*false\|isNonInteractive.*false" src/ | grep -v test` → 0 production matches; `get-started` is load-mode only (`get-started/index.ts`: "should never be called directly"). Skill-dispatch forks already strip `ask_question` via `isSkillDispatch`. The default flip is a latent-hang fix for DAG/farm/compose forks, not a regression.
- **api-compat**: `ForkSubagentOptions.denyElicitations` and `AgentConfig.isNonInteractive` keep their `boolean|undefined` shape; only defaults change. Zero in-tree callers depend on the old default. (Behavior change for external library consumers is real but documented in the PR body and opt-out-able; per reachability rule with zero production importers of the old default → not a breaking-change finding.)
- **Mirror claim accurate**: the `parentSessionId` sub-agent detection mirrors `memory-hooks.ts:28` and `plan-mode-gate.ts:47` (both `if (context.parentSessionId) return {}`). Note a behavioral nuance — those two *skip* (no-op) while path-approval *blocks*; intentional and correct, since allowing out-of-root access silently would be the security hole.
- **No secret exposure**: the new `console.error` logs a resolved filesystem path to stderr; paths are not credentials.
- **Test coverage**: all six new behaviors are covered (fork default `isNonInteractive:true`; explicit `false` respected; `denyElicitations` omitted→DENY over parent handler; `false`→parent propagates; path-approval out-of-root auto-deny + in-root pass-through; executor `denyElicitations:true` both modes).

## CONFLICT resolved
Agent 1 (security) and Agent 2 (correctness) disagreed on whether `onElicitation` is the live MCP mechanism. Independent re-derivation (grep at ref + `client.ts` capability read + absent `buildQueryOptions` definition) confirms **Agent 2**: `onElicitation` is unconsumed; MCP is closed at the protocol layer. Folded into the MEDIUM finding above.

## What was NOT checked
- Read against branch HEAD `96d7b3a` — all citations verified at that ref via `git show`/`git grep`.
- Did **not** run the test suite (PR body notes a pre-existing local `memory.db` v3-vs-v2 schema mismatch; CI runs clean state). Coverage assessed by reading the test files, not by execution.
- Did **not** trace `parentSessionId` propagation through the DAG executor to depth-2+ grandchildren (pre-existing, not in diff).
- Did **not** review the announced follow-up PR (`bypassPermissions` grant-all) — out of scope.

No blocking findings. The change is correct, well-tested, and achieves its stated intent. Recommend (non-blocking, this PR or follow-up) tightening the `subagent.ts:420–433` comment to stop attributing the MCP-channel closure to the currently-unconsumed `onElicitation` install (MEDIUM), and refreshing the stale `denyElicitations` JSDoc at `:104` (NIT).

---

**Done**
- Reviewed PR #218 (`96d7b3a`) across security, api-compat, correctness, spec-compliance, test-coverage, and perf via a 2-agent parallel wave plus independent citation/conflict verification; emitted a severity-sorted report with a **MERGE** recommendation.
- Evidence lives in this report (durable findings with `file:line` + ref citations) and the diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/5a4c6540-16d8-4ebf-a582-e4c7b4644eea/pr-218.diff`; the one contested claim (MCP `onElicitation` is unconsumed) was re-derived via `git grep` at the ref.
- World unchanged — this skill is read-only: no files edited, no commits, no PR comments or status mutations.
- Pending/deferred: 1 MEDIUM (comment over-attributes MCP closure to dead `onElicitation`) + 1 LOW (path-approval test doesn't pin branch ordering) + 2 NITs are recommendations for the author, not merge blockers; test suite was not executed (pre-existing local `memory.db` schema mismatch — CI runs clean).

---

PR #219, opened 2026-06-19
feat(permissions): make bypassPermissions real — disable path containment + add /bypass, flag, config key

griffinwork40:
## Problem

Two issues, surfaced from a user report ("everything says permissions default to bypass but I always get prompts in the CLI"):

1. **`bypassPermissions` was near-inert.** There is no per-tool approval flow in agent-afk; the only runtime effect of bypass was a `console.warn` in the bash handler and a `get_runtime_state` label. It did **not** suppress the path-approval prompt (the prompt users actually hit).
2. **No way to enable it interactively, and stale docs.** Bypass was only set by the daemon (hard-coded); docs claimed bypass-by-default and advertised an unwired `--permission-mode` flag.

## What this does

Makes `bypassPermissions` actually disable path containment + the path-approval prompt, and adds three ways to turn it on.

**Core mechanism** (`b320187`): an `allowAll` switch threaded from the session permission mode through the single containment chokepoint.
- `_cwd-utils.ts`: `resolveAndContain` + `wouldBeRestricted` short-circuit on `context.allowAll` — both the handler's containment throw AND the path-approval prompt vanish together (they share this code, so they stay in sync).
- `SessionToolDispatcher`: `allowAll` opt → per-call context + `getGrants()`, sampled fresh per call (a live toggle takes effect next tool call).
- Both providers' `buildDispatcher` derive `allowAll = permissionMode === 'bypassPermissions'` and track `_currentPermissionMode` so the provider's `getGrants()` — read by the path-approval hook — stays in sync with the dispatcher.

**`/bypass` REPL toggle + badge** (`35c392b`): `/bypass [on|off]` flips the live permission mode (mutually exclusive with `/plan` and `/afk`); a never-dropped `⚠ BYPASS` status-line badge keeps a dangerous mode visible.

**Config key + CLI flag** (`2de4171`): `"permissionMode": "bypassPermissions"` in `afk.config.json` (validated on load — invalid strings ignored) and `--dangerously-skip-permissions` on `afk`/`afk chat` + `afk interactive`. Resolution is flag > config > default; threaded into the `AgentSession` construction on both surfaces. The model-swap closure preserves the live mode (a `/bypass` toggle survives a mid-session model swap).

**Docs** (`44067ca`): corrected the stale bypass-by-default claims (README, architecture.md), removed the unwired `--permission-mode` flag line, fixed the false Telegram "full tool access via bypass" claim, and documented the three enablement paths.

## Notes / caveats

- Bypass quiets **path-approval**, not `ask_question` — that's the model choosing to ask you something, a separate axis. Documented in `/bypass` help + README.
- Already active for the `afk daemon` (which sets `bypassPermissions`): it removes the out-of-root auto-deny that previously blocked the daemon.
- `AFK_DISABLE_PATH_APPROVAL=1` remains the process/CI escape hatch; `bypassPermissions` is the clean per-session user mode.

## Verification

- `pnpm lint` clean (rebased onto current `main`).
- Targeted tests pass: `_cwd-utils`, `path-approval-hook`, `/bypass`, `dispatcher` (110 tests), plus config loading (the lone config.test failure is a pre-existing MLX slot-binding env artifact, unrelated).
- Full `pnpm test` is unreliable in my local env due to a pre-existing `memory.db` schema v3 vs build v2 mismatch (not introduced here); CI runs against a clean state dir.

Pairs with #218 (forked sub-agents non-interactive). Independent of it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ Aliases: `afk c` → `chat`, `afk i` → `interactive`, `afk s` → `status`.
 
 The one prompt you may hit is **path approval**: when a file tool (read/write/edit/list/glob/grep) targets a path *outside* the session's working directory, `afk` asks before allowing it. Pre-authorize paths with `/allow-dir <path>` (or answer "persist" at the prompt to remember them across sessions).
 
-To turn path approval off entirely — letting the agent read and write **anywhere** with no prompt — enable **bypass mode** with `/bypass` in the REPL. The status line shows `⚠ BYPASS` while it is active. This is the equivalent of `--dangerously-skip-permissions`; use it only on a machine and account you trust. `afk daemon` runs in bypass mode by default (no human to prompt). Bypass does not change `ask_question` — that is the model choosing to ask you something, a separate axis.
+To turn path approval off entirely — letting the agent read and write **anywhere** with no prompt — enable **bypass mode** any of three ways: `/bypass` in the REPL (the status line shows `⚠ BYPASS`), the `--dangerously-skip-permissions` flag on `afk`/`afk chat`, or `"permissionMode": "bypassPermissions"` in `afk.config.json`. This is the equivalent of Claude Code's `--dangerously-skip-permissions`; use it only on a machine and account you trust. `afk daemon` runs in bypass mode by default (no human to prompt). Bypass does not change `ask_question` — that is the model choosing to ask you something, a separate axis.
 
 ## Troubleshooting
 
--- a/README.md
+++ b/README.md
@@ -153,9 +153,11 @@ Aliases: `afk c` → `chat`, `afk i` → `interactive`, `afk s` → `status`.
 
 ## A note on permissions
 
-`afk` runs with **full permissions** by default: no per-tool prompts. Claude can run bash, read and write files, fetch URLs, and call MCP servers without asking each time. This is intentional — `afk` is built for unattended work, where a permission prompt with no human in front of it is just a wedged session.
+`afk` does not prompt before each tool call — there is no per-tool approval flow. Claude runs bash, reads and writes files, fetches URLs, and calls MCP servers without asking each time. This is intentional — `afk` is built for unattended work, where a permission prompt with no human in front of it is just a wedged session.
 
-Use `afk` on a machine and account you trust. Override per-session with `--permission-mode` if you want stricter behavior.
+The one prompt you may hit is **path approval**: when a file tool (read/write/edit/list/glob/grep) targets a path *outside* the session's working directory, `afk` asks before allowing it. Pre-authorize paths with `/allow-dir <path>` (or answer "persist" at the prompt to remember them across sessions).
+
+To turn path approval off entirely — letting the agent read and write **anywhere** with no prompt — enable **bypass mode** with `/bypass` in the REPL. The status line shows `⚠ BYPASS` while it is active. This is the equivalent of `--dangerously-skip-permissions`; use it only on a machine and account you trust. `afk daemon` runs in bypass mode by default (no human to prompt). Bypass does not change `ask_question` — that is the model choosing to ask you something, a separate axis.
 
 ## Troubleshooting
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -120,11 +120,14 @@ Markers never leak back into stored history — `cache-policy.ts` clones-and-sta
 
 ## Bypass permissions
 
-By default, the Claude Agent SDK asks for permission before executing tools like bash, file ops, web requests, etc. **`agent-afk` enables `bypassPermissions: true`** which:
+By default, `agent-afk` runs in permission mode `'default'`: tools execute without a per-tool approval prompt, but a file tool touching a path **outside the session's granted roots** triggers a path-approval prompt (and out-of-root access stays contained until granted). There is no per-tool "allow this bash command?" flow.
 
-- Skips all permission prompts
+**Bypass mode** (`permissionMode: 'bypassPermissions'`) disables path containment and the path-approval prompt entirely — filesystem tools may read/write any path with no confirmation:
+
+- Skips path-approval prompts; disables out-of-root containment
 - Allows fully automated workflows
-- Designed for CLI/scripting and unattended daemon work
+- Enable with `/bypass` in the REPL (status line shows `⚠ BYPASS`); the `afk daemon` sets it directly for unattended work
+- Does **not** affect `ask_question` (the model asking you a question is a separate axis)
 
 This is intentional — `agent-afk` is built for unattended work, where a permission prompt with no human in front of it just wedges the session. Use on a machine and account you trust.
 
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -68,6 +68,7 @@ All commands support `--format json` where machine-readable output makes sense (
 --max-turns <n>           # safety rail (default 100)
 --max-budget-usd <n>
 --task-budget <tokens>
+--dangerously-skip-permissions  # bypass mode: skip path-approval prompts (read/write anywhere)
 --no-update-check         # suppress startup update check
 --dump-prompt <path>      # write resolved system prompt to file (provenance included)
 ```
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -68,7 +68,6 @@ All commands support `--format json` where machine-readable output makes sense (
 --max-turns <n>           # safety rail (default 100)
 --max-budget-usd <n>
 --task-budget <tokens>
---permission-mode <mode>  # default | acceptEdits | bypassPermissions | plan
 --no-update-check         # suppress startup update check
 --dump-prompt <path>      # write resolved system prompt to file (provenance included)
 ```
@@ -194,7 +193,7 @@ The setup wizard prompts for the token and the chat IDs allowed to interact with
 - `/reset` — clear conversation
 - `/model local|small|medium|large|opus|sonnet|haiku|fable` — switch model
 
-Send any other message to chat. The bot has full tool access via bypass mode.
+Send any other message to chat. The bot runs in the default permission mode — tools execute without per-tool prompts, and file access outside the working directory stays contained.
 
 **Background-service helpers:**
 
--- a/src/agent/providers/anthropic-direct/bypass-live-toggle.test.ts
+++ b/src/agent/providers/anthropic-direct/bypass-live-toggle.test.ts
@@ -1,0 +1,129 @@
+/**
+ * Regression test for the live `/bypass` (permission-mode) toggle.
+ *
+ * The reported bug: starting a session in `bypassPermissions` (via
+ * `--dangerously-skip-permissions`, the `permissionMode` config key, or the
+ * daemon) and then turning bypass OFF mid-session left the agent effectively
+ * unrestricted while the UI badge cleared — i.e. `/bypass off` failed UNSAFE.
+ *
+ * Root cause: `setPermissionMode()` updated only the system-prompt mode field,
+ * never the two fields that actually gate enforcement:
+ *   1. the provider's `_currentPermissionMode` — read by `getGrants().allowAll`,
+ *      which the path-approval hook consults; and
+ *   2. the live dispatcher's `_allowAll` — read per call by the file-tool
+ *      handler containment chain (`resolveAndContain` / `wouldBeRestricted`).
+ *
+ * These tests assert BOTH reads flip together on every `setPermissionMode`
+ * call, in both directions, within a single session (no model swap).
+ */
+
+import { describe, it, expect, afterEach } from 'vitest';
+import { mkdtempSync, rmSync } from 'fs';
+import { tmpdir } from 'os';
+import path from 'path';
+import { AnthropicDirectProvider, __setAnthropicClientFactory } from './index.js';
+
+/** A stub client whose message stream is immediately done — no network. */
+function emptyStreamClient() {
+  return () =>
+    ({
+      messages: {
+        stream: () => ({
+          [Symbol.asyncIterator]() {
+            return { next: async () => ({ done: true, value: undefined }) };
+          },
+        }),
+      },
+    }) as unknown as InstanceType<typeof import('@anthropic-ai/sdk').default>;
+}
+
+/**
+ * Test-only read of the LIVE dispatcher's bypass flag — the file-tool
+ * containment source. The handler context reads the same `_allowAll` field, so
+ * the dispatcher's `getGrants().allowAll` is a faithful proxy for whether file
+ * tools are contained.
+ */
+function liveDispatcherAllowAll(q: unknown): boolean {
+  return (
+    q as { state: { toolDispatcher: { getGrants(): { allowAll: boolean } } } }
+  ).state.toolDispatcher.getGrants().allowAll;
+}
+
+describe('AnthropicDirectProvider — live /bypass toggle changes effective enforcement', () => {
+  afterEach(() => __setAnthropicClientFactory(undefined));
+
+  it('setPermissionMode flips BOTH enforcement reads live; /bypass off fails CLOSED', async () => {
+    __setAnthropicClientFactory(emptyStreamClient());
+    const provider = new AnthropicDirectProvider();
+    const baseDir = mkdtempSync(path.join(tmpdir(), 'bypass-live-'));
+    try {
+      // Start bypassed — mirrors --dangerously-skip-permissions / config / daemon.
+      const q = provider.query({
+        prompt: 'noop',
+        config: {
+          apiKey: 'sk-test',
+          cwd: baseDir,
+          permissionMode: 'bypassPermissions',
+        } as unknown as Parameters<typeof provider.query>[0]['config'],
+      });
+      // Drain one event so the SDK lifecycle is fully spun up (buildDispatcher
+      // and _currentPermissionMode are already set synchronously by query()).
+      await q[Symbol.asyncIterator]().next();
+
+      // Both enforcement sources see bypass at construction.
+      expect(provider.getGrants().allowAll).toBe(true); // path-approval hook
+      expect(liveDispatcherAllowAll(q)).toBe(true); // file-tool containment
+
+      // THE reported bug: /bypass off must actually restore containment.
+      // Pre-fix both reads stayed `true` (agent unrestricted, badge safe).
+      await q.setPermissionMode('default');
+      expect(provider.getGrants().allowAll).toBe(false);
+      expect(liveDispatcherAllowAll(q)).toBe(false);
+
+      // /bypass on from default enables bypass live — no model swap required.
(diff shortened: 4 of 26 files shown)


griffinwork40:
All breadth points are now ground-truthed. Synthesis inputs are complete and convergent — the central finding was independently re-derived by my own reads plus all three subagents (sec, correctness, verify), with consistent `file:line` evidence, so the high-severity finding is maximally supported (the `/shadow-verify` gate's intent — no unsupported high finding reaching the decision — is satisfied by this 4-way convergence rather than a redundant 5th pass).

---

# Code Review — PR #219: `feat(permissions): make bypassPermissions real`

**Target:** PR #219 `feat/bypass-grant-all` → `main` · reviewed ref `445c4b92` · 21 files, +354/−13 · regime: **full**
**Stated intent:** PR #219 title+body — spec-compliance **assessed**.
**Verification:** all citations read at ref `445c4b92` via `git show`; live-toggle finding independently confirmed by 3 subagents + orchestrator reads (no cross-agent conflicts).

## Summary
The **construction-time** enablement paths — `--dangerously-skip-permissions` flag, `permissionMode` config key, and the daemon — correctly thread `allowAll` through the single containment chokepoint (`_cwd-utils` short-circuit + path-approval hook stay in sync; verified). That core mechanism is sound. **But the headline interactive feature — the live `/bypass` toggle + `⚠ BYPASS` badge — does not control actual enforcement within a session**, and its failure has an unsafe direction. That is a blocking defect for a security-sensitive permission feature.

---

## Blocking findings

### 1 · HIGH · confidence: high · correctness + spec-compliance + security · `defeats stated purpose`
**The live `/bypass` toggle does not change effective path-containment enforcement within a session; the `⚠ BYPASS` badge can misrepresent the real state.**

Mechanism (field mismatch — toggle writes one field, enforcement reads another; `query()` runs once):
- `agent-session.ts:749-751` (ref:445c4b92, file-state): `setPermissionMode` → `providerQuery.setPermissionMode(mode)` + `setSessionMetadata({permissionMode})`. It does **not** touch `this.config.permissionMode`.
- `anthropic-direct/query.ts:466-468` (file-state): `setPermissionMode` sets **only** `this.state.currentPermissionMode` (consumed by `composeSystem` for the plan addendum).
- `anthropic-direct/index.ts:578-581` (file-state): `const permissionMode = config.permissionMode ?? 'default'; this._currentPermissionMode = permissionMode;` — set **inside `query()`**.
- `anthropic-direct/index.ts:509` (diff-context): `getGrants()` returns `allowAll: this._currentPermissionMode === 'bypassPermissions'`.
- `anthropic-direct/index.ts:366` (diff-context): dispatcher built with `allowAll: permissionMode === 'bypassPermissions'` (param captured once).
- `agent-session.ts:217` (file-state): `provider.query()` is invoked **once** at construction; re-run only on provider reactivation (model swap/reset). OpenAI mirror: `openai-compatible/index.ts:157,206,458`.

Two reachable directions:
- **(a) `/bypass on` from default** → `stats.permissionMode` flips, badge shows `⚠ BYPASS`, but `allowAll` stays false until a model swap → containment + path-approval prompt remain active. Feature inert; **fails safe**.
- **(b) `/bypass off` after starting via `--dangerously-skip-permissions`/config** → badge clears, but `_currentPermissionMode`/dispatcher `_allowAll` stay `bypassPermissions` until a model swap → **agent still reads/writes anywhere while the UI implies containment is restored. Fails UNSAFE.**

The intent's claim *"a live toggle takes effect next tool call"* and *"getGrants() stays in sync with the per-query dispatcher"* hold only across a model swap (where `bootstrap.ts:683-686` rebuilds with `stats.permissionMode`), not live.

> evidence (`anthropic-direct/query.ts:466-468`):
> ```
> async setPermissionMode(mode: string): Promise<void> {
>   this.state.currentPermissionMode = mode;
> }
> ```

**suggestion:** make `setPermissionMode` update the live enforcement source — e.g. set `provider._currentPermissionMode` and swap the dispatcher's `allowAll` the way `setCwd` rebuilds the dispatcher (`index.ts:871`), or derive `allowAll`/`getGrants().allowAll` from `state.currentPermissionMode`. If a live toggle is out of scope, drop `/bypass` (keep flag+config) and remove the never-dropped badge so the UI cannot imply live control.

---

## Non-blocking findings

### 2 · LOW · confidence: medium · test-coverage · `confirmed-absent`
**No test exercises the live-toggle → `allowAll` enforcement integration** — which is why finding 1 shipped. `bypass.test.ts:20,24` mocks `setPermissionMode` (`vi.fn().mockResolvedValue(undefined)`) and asserts only `stats` mirroring/repaint; `_cwd-utils.test`/`path-approval-hook.test` construct `allowAll` directly (verify the chokepoint *given* `allowAll`, never that `/bypass` *sets* it). `git grep -n allowAll <ref> -- 'src/**/*.test.ts'` finds no provider/dispatcher-level toggle test. **suggestion:** integration test — real session, `setPermissionMode('bypassPermissions')` → assert `getGrants().allowAll===true` and an out-of-root path resolves without prompt; plus the reverse.

### 3 · LOW · confidence: medium · security + correctness
**Bypass mode does not relax the bash restricted-substring guard.** `bash-restriction-hook.ts:174-177` (file-state) reads `getGrants()` then `deriveRestrictedSubstrings(grants)` but ignores `grants.allowAll`, so bash referencing restricted paths stays blocked even in bypass. Fails safe; bash is a documented separate axis (`AFK_DISABLE_PATH_APPROVAL` / interpreter-guard escape hatches), but it is inconsistent with the "read/write **anywhere**" framing. **suggestion:** short-circuit the guard on `grants.allowAll`, or document that bypass relaxes only file-tool containment + path-approval, not bash restriction.

### 4 · LOW · confidence: medium · spec-compliance
**Bypass does not propagate to forked sub-agents.** `git grep permissionMode <ref> -- 'src/agent/**/subagent*.ts'` is empty → a bypassed parent's children default to contained (inherited roots only). Fails safe and is pre-existing (not introduced here), but it partially undercuts the PR's "removes the out-of-root auto-deny that blocked the daemon" claim for sub-agent work. **suggestion:** decide intended semantics; thread `permissionMode` into the subagent config if children should inherit, else document that bypass is per-session.

### 5 · NIT · confidence: low · correctness
**Config validation whitelist is narrower than the `PermissionMode` union.** `config.ts` accepts `{default, plan, autonomous, bypassPermissions}` (diff-context); `sdk-types.ts:16-22` also defines `acceptEdits`, `dontAsk`, `auto` — silently ignored if set in config. Defensive (avoids an unimplemented mode), but a type-valid value becomes a silent no-op. **suggestion:** document the supported subset, or accept the full union.

---

## Verified clean (no issues)
- **api-compat:** `getGrants()` gained `allowAll: boolean` in all 3 production impls (`anthropic-direct/index.ts:504`, `openai-compatible/index.ts:453`, `dispatcher.ts:323`); `GrantManager` interface declares it optional (`allow-dir.ts:30`). Additive + internal; reachability check found **no** production importer that breaks. Not a breaking change.
- **Docs accuracy (Telegram):** `git grep permissionMode <ref> -- 'src/telegram/**'` is empty → Telegram runs in default mode → the corrected "default mode" doc claim is accurate (the old "full tool access via bypass" line was false).
- **`--dangerously-skip-permissions` reach:** registered on `chat` (`:154`) + `interactive` (`:188`); `interactive` is `{ isDefault: true }` (`:154`), so bare `afk --dangerously-skip-permissions` routes correctly → README's "`afk`/`afk chat`" claim holds.
- **Containment chokepoint:** `_cwd-utils.ts` `allowAll` short-circuit leaves the non-bypass path byte-unchanged; `resolveAndContain` and `wouldBeRestricted` agree under `allowAll` (handler admits ⇔ hook skips prompt).

## Spec-compliance verdict
- ✅ Core file-tool containment bypass via `allowAll` (construction-time: flag, config key, daemon) — implemented & correct.
- ✅ Flag on all documented surfaces; ✅ config key validated; ✅ docs corrections accurate (incl. Telegram).
- ❌ **Unmet/incorrect clause:** the live `/bypass` toggle "takes effect next tool call" + badge sync — does not hold within a session (finding 1).
- No scope creep observed.

## What was **not** checked
- Read against branch HEAD `445c4b92` — all citations verified at that ref. No patch-file ambiguity.
- Did **not** run the test suite (PR notes local full-suite unreliable from a pre-existing `memory.db` schema mismatch); findings are static-analysis + targeted reads.
- Did **not** exercise the runtime toggle end-to-end — but the mechanism is deterministic and confirmed by 4 independent code derivations.
- Did not deeply trace the `openai-compatible` `setCwd` rebuild beyond confirming structural parity with `anthropic-direct`.

---

The construction-time mechanism is solid, but the PR's headline interactive control (`/bypass` + `⚠ BYPASS` badge) does not govern actual enforcement and has an unsafe direction — `/bypass off` can leave the agent unrestricted while the badge shows safe. For a feature whose entire purpose is making a dangerous mode *visible and controllable*, an indicator + control that misrepresent the real state is blocking. Resolve finding 1 (wire the live path, or drop the toggle/badge and ship only the flag+config paths) and add the finding-2 integration test before merge.

**Done**
- Completed a full read-only review of PR #219 at ref `445c4b92`: triage → 2 parallel review subagents (security/api-compat; correctness/spec/test) + 1 verification subagent via a compose DAG → orchestrator-grounded breadth checks → synthesis. Deliverable is this report (transcript) plus the captured diff at `/Users/griffinlong/.afk/state/skill-preflight/b2ad40ff-427a-4533-be06-539d3274cecb/pr-219.diff`.
- Verdict: **DO NOT MERGE** — 1 HIGH (live `/bypass` toggle/badge does not control enforcement; unsafe `/bypass off` direction), 3 LOW, 1 NIT; api-compat/Telegram-docs/flag-reach/chokepoint verified clean.
- World unchanged: **read-only** — no files edited, no commits, no PR/MR mutations (the first re-dispatch hit a 50-call subagent budget and was re-run at 95; noted for transparency).
- Pending/deferred: none — review is terminal. Acting on any finding (the finding-1 fix, finding-2 test) is out of scope for this skill and left to the author.

---

griffinwork40:
Verification complete. The manifest materially changed two of the Wave 1 findings: **CLAIM 2 (stale toggle) is REFUTED** (slash commands dispatch only between turns; `_currentPermissionMode` re-syncs at every `query()`, so the window is zero), and **CLAIM 1 (bash hook)** is a pre-existing fixed credential denylist, not a regression. The gate integrity (CLAIM 4) is CONFIRMED sound. No finding survives above **low** severity, so per the skill no `/shadow-verify` pass is required (it triggers only on critical/high). Synthesizing locally.

---

# Code Review — PR #219

**`feat(permissions): make bypassPermissions real — disable path containment + add /bypass, flag, config key`**

| | |
|---|---|
| Target | PR #219 · `feat/bypass-grant-all` → `main` |
| Reviewed ref | `445c4b92dd614fdfb3860e627028f585085a243d` (citations verified at this SHA) |
| Regime | **full** (feature, 21 files, ~367 lines) |
| Stated intent | PR #219 title+body — **spec-compliance assessed** |
| Diff artifact | `~/.afk/state/skill-preflight/70f11e63-…/pr-219.diff` |

## ⟶ Merge decision: **MERGE** (no blocking defects; low-severity follow-ups below)

The core mechanism is correct and tightly gated: `allowAll` is derived **exclusively** from `permissionMode === 'bypassPermissions'` at every production site (both providers' `buildDispatcher` + `getGrants()`), no other mode (`autonomous`/`plan`) sets it, config strings are validated against an allowlist, and forked subagents do **not** inherit bypass (no privilege-escalation path). The PR delivers all three stated enablement paths and disables both the containment throw and the path-approval prompt through the single shared chokepoint. The two Wave-1 `medium` findings were downgraded/refuted on independent verification.

---

## ⚠ Notable behavior change (the item a human reviewer must consciously accept)

**`afk daemon` filesystem blast radius expands from "contained to cwd" to "anywhere."** The daemon already runs in `bypassPermissions` (`scheduler.ts:442`, ref-verified). Before this PR, bypass was near-inert, so the daemon's file tools were still contained by the out-of-root auto-deny. After this PR, every existing daemon now reads/writes **any path** with no containment. This is *intended and documented* in the PR ("removes the out-of-root auto-deny that previously blocked the daemon") — not a defect — but it is the single most security-consequential real-world effect and should be an explicit, conscious approval, not a silent side-effect of merging.

---

## Findings (sorted by severity; security first within tier)

**low · high · security · `src/agent/tools/hooks/bash-restriction-hook.ts:217–238` · ref:445c4b9 · file-state — bash credential denylist is only partial protection once file-tool bypass exists.** In bypass mode, file tools can read `~/.ssh` etc. (containment off), but the bash hook's fixed credential denylist (`~/.ssh`, `~/.aws`, `~/.gnupg`, `/etc/shadow`) still blocks `cat ~/.ssh/id_rsa`. The hook never consults `grants.allowAll` (pre-existing; file is *not* in this PR's diff). Net effect: the bash denylist becomes circumventable-in-practice (use `read_file` instead of `cat`), and the two file subsystems diverge.
Evidence: `const restrictedSubstrings = deriveRestrictedSubstrings(grants);` with no `allowAll` guard; `deriveRestrictedSubstrings` builds a fixed `[~/.ssh, ~/.aws, …]` list.
Suggestion: decide explicitly and document — either gate the bash denylist on `allowAll` for uniform "anywhere" semantics (matches Claude Code's `--dangerously-skip-permissions`), or keep it as deliberate defense-in-depth and note in `/bypass` help that bypass does **not** lift the bash credential denylist. Not blocking; the user opted into "dangerous."

**low · high · observability · `src/agent/providers/*/index.ts` (getGrants/query), `/bypass` handler · ref:445c4b9 · file-state — enabling bypass leaves no persistent audit trail.** `appendProviderAuditLog` records `grant-read`/`grant-write`/`revoke` but never a `bypass-enable`/`bypass-disable` event (grep: zero matches). Removing *all* path containment is a higher-consequence act than a single grant, yet the only record is an ephemeral REPL notice.
Suggestion: emit `appendProviderAuditLog({ action: 'bypass-enable' | 'bypass-disable', … })` from the `setPermissionMode` path so bypass windows are forensically reconstructable in `~/.afk/state/`.

**low · high · test-coverage · `src/agent/providers/anthropic-direct/index.ts:366`, `openai-compatible/index.ts:399`, `src/cli/config.ts:621–626` · ref:445c4b9 · file-state — the gate derivation and config validation are untested.** Verified absent (`git grep permissionMode -- 'src/**/*.test.ts'`): no test asserts `getGrants().allowAll === true` when `permissionMode==='bypassPermissions'` (or `false` for every other mode), and no test covers config-key validation (accept `'bypassPermissions'` / reject garbage). `_cwd-utils`, `path-approval-hook`, and `/bypass` are well covered; the *security gate itself* and the *config trust boundary* are the two untested edges. (bootstrap `initialPermissionMode` IS covered — `daemon-session-factory.test.ts`, `resume-swap.test.ts`.)
Suggestion: add a provider-level test (`bypassPermissions ⇒ allowAll true; default/autonomous/plan ⇒ false`) and a `config.test.ts` case asserting garbage strings are dropped.

**nit · high · security · `src/agent/tools/subagent-executor.ts:560–582` · ref:445c4b9 · file-state — subagent containment is correct but unenforced by comment.** The `agent`-tool `childConfig` omits `permissionMode`, so children start `'default'` (contained) regardless of parent mode — the safe behavior, and no default→bypass escalation exists. But `SubagentManager.forkSubagent` does `{ ...options.config }` (`subagent.ts:377`), so a future caller that threads `permissionMode` through would silently propagate bypass.
Suggestion: add `// Invariant: permissionMode is intentionally NOT forwarded — subagents always start contained.` at the `childConfig` block. (Also worth a doc line: daemon subagents stay contained even though the daemon main session is bypass — an intentional asymmetry users may find surprising.)

**nit · high · docs · `README.md:13–15`, `docs/reference.md:50` · ref:445c4b9 · diff-context — README slightly overstates scope and lists the wrong commands.** README says bypass lets "the agent read and write **anywhere**" and names the flag on "`afk`/`afk chat`". Verified: the flag is registered only on `afk chat` and `afk interactive` — **not** bare `afk` (no root default-action handler; `src/cli/index.ts`), and the README omits `afk interactive`. "Anywhere" is also imprecise given the bash denylist above. (`docs/architecture.md` correctly scopes it to "filesystem tools" — keep that precision.)
Suggestion: change to "the `--dangerously-skip-permissions` flag on `afk chat` / `afk interactive`" and qualify "anywhere" as "any path via filesystem tools."

**Reconciled by verification (no longer a finding):** Wave 1's *medium* "live `/bypass` toggle staleness window" — **REFUTED**. Slash commands dispatch only between turns (`loop-iteration.ts:267`, after `readLine()`), and `_currentPermissionMode` re-syncs at the top of every `query()` (`index.ts:581`). The two-field pattern (query-handle state vs. provider field) is design-intentional and the REPL serializes the toggle, so `allowAll` is always fresh at the next tool call. Internal-clarity nit only.

## Verified-correct (positive confirmations)
- **Gate integrity** — `allowAll === (permissionMode === 'bypassPermissions')` at all sites; no other mode enables it.
- **Containment chokepoint** — `resolveAndContain` and `wouldBeRestricted` short-circuits agree; relative paths still anchor to `resolveBase` before the bypass branch.
- **Resolution precedence** — flag > config > default, correctly implemented in `chat.ts` and `bootstrap.ts`; model-swap closure preserves the *live* `stats.permissionMode`.
- **Config trust boundary** — invalid `permissionMode` strings are dropped (typo can't silently land bypass).
- **api-compat** — `getGrants()` return-type widening (`allowAll`) and `ToolHandlerContext.allowAll?` are additive; all production callers guard via `=== true` / conditional spread; no breakage.

## What was NOT checked
- The OpenAI-compatible provider's query-handle `setPermissionMode` symmetry with the Anthropic-direct path (verifier flagged unresolved; same field pattern, not individually traced).
- Telegram surface — whether the bot reads/honors the `permissionMode` config key.
- `AFK_DISABLE_PATH_APPROVAL=1` × `bypassPermissions` precedence interaction.
- `afk.config.json` `permissionMode` persistence across daemon restarts.
- The test suite was **not run** (read-only review; working tree is dirty and must not be mutated).

---

**Done** — Reviewed PR #219 at ref `445c4b92` via two parallel dimension agents + an independent claim-verification pass; emitted the report above with merge decision **MERGE**.
- Evidence: findings cite `file:line` at ref `445c4b92dd614fdfb3860e627028f585085a243d`; diff artifact at `~/.afk/state/skill-preflight/70f11e63-4aa1-4f98-87d4-45a5b0192ee7/pr-219.diff`; verification manifest reconciled the two Wave-1 `medium` findings (one refuted, one downgraded).
- World change: none — read-only review; no files, commits, PR comments, or PR state were touched.
- Pending/deferred: five low/nit follow-ups (bash-denylist consistency, bypass audit log, gate+config tests, subagent invariant comment, README wording) — none blocking; left for the author to action. No `/shadow-verify` run (skill triggers it only on critical/high findings; none present).

---

PR #225, opened 2026-06-21
feat(repl): Shift+Tab cycles permission modes (default → plan → bypass)

griffinwork40:
## Summary
- Shift+Tab in the REPL now cycles permission modes `default → plan → bypassPermissions → default` instead of only toggling plan↔default — every keyboard-reachable mode is one press apart.
- **AFK (`autonomous`) is deliberately excluded** from the keypress ring. Its enter/exit machinery (Telegram push-budget reset, elicitation→ledger swap, abort-watcher, presence marker) is too heavy to fire on a transient pass-through. It stays on `/afk`; if the session is already in AFK, Shift+Tab exits it cleanly to `default` via `toggleAfkMode` (full teardown).
- `plan`/`bypass`/`default` are pure `setPermissionMode` flips; the autonomous-exit case reuses the existing `toggleAfkMode` helper, so no teardown leaks and no extraction refactor was needed.
- Reworded the model-facing plan-mode addendum, the loading tip, `/plan` docs, and `plan-mode-toggle` comments that described the old "Shift+Tab exits plan" semantics; added a `⚡ bypass` prompt marker for parity with the status line.

## Why this design (3-mode, AFK excluded)
A `/shadow-verify` pass confirmed AFK's machinery lives only in `toggleAfkMode` (a raw `setPermissionMode('autonomous')` would leak the abort-watcher / strand presence). A `/devils-advocate` pass surfaced the safety case for keeping AFK off an unguarded single keypress (`permissionMode` is a documented self-escalation vector; AFK side-effects are non-idempotent). The 3-mode ring meets the goal — fast keyboard mode-switching — without that risk; AFK remains on `/afk`. Full rationale + alternatives in `.afk/plans/shift-tab-permission-mode-cycle.md`.

## Test results
- Targeted vitest (permission-mode-cycle, plan-mode-toggle, afk-mode-toggle, repl-loop-wiring, loop-iteration, plan-mode-addendum, status-line): **87 passed**.
- Full suite (`vitest run`): **9777 passed, 12 skipped, 2 failed**. The 2 failures are pre-existing and unrelated — model-id string mismatches (`claude-haiku-4-5` vs `…-20251001`) in `src/cli/config.test.ts:538` and `src/agent/providers/anthropic-direct.test.ts:1070`; neither file is touched by this PR.

## Verification
Adversarial read-only verifier re-derived the load-bearing claims from the diff: **CONFIRMED** — (1) both `onShiftTab` handlers repointed to `cyclePermissionMode` with no dangling `togglePlanMode` import (passes `noUnusedLocals`); (2) ring math correct, AFK delegates to `toggleAfkMode(ctx, false)`, unknown current-mode falls back to `default` with no off-by-one and no `undefined` index access under `noUncheckedIndexedAccess`; (3) neither pre-existing-failing file (`config.ts`, `anthropic-direct.ts`) is in the diff.

### DIFF
--- a/.afk/plans/shift-tab-permission-mode-cycle.md
+++ b/.afk/plans/shift-tab-permission-mode-cycle.md
@@ -1,0 +1,86 @@
+# Shift+Tab → cycle permission modes (3-mode ring)
+
+## Goal
+Make the REPL's **Shift+Tab** key cycle through permission modes instead of only
+toggling `plan ↔ default`.
+
+## Decision (chosen: option **a**, the safe 3-mode ring)
+Shift+Tab cycles a fixed ring **`default → plan → bypassPermissions → default`**.
+**AFK (`autonomous`) is deliberately excluded** from the keypress ring and stays
+on the `/afk` command. The one concession: if the session is *already* in
+`autonomous` (operator ran `/afk on`), Shift+Tab exits AFK cleanly (full
+teardown) and lands on `default` — so the key still does something sensible from
+AFK without ever *entering* it on a transient press.
+
+Built the **cheap way** (devils-advocate pragmatist): no extraction refactor of
+`afk-mode-toggle.ts`. `plan`/`bypass`/`default` are pure `setPermissionMode`
+flips; the AFK-exit case reuses the existing `toggleAfkMode(ctx, false)` helper.
+
+## Verified facts (shadow-verify, all CONFIRMED via independent re-derivation)
+- 4 modes: `default`, `plan`, `autonomous` (AFK), `bypassPermissions`. New
+  installs default to `bypassPermissions` (`config.ts:342` `DEFAULT_CLI_PERMISSION_MODE`).
+- Shift+Tab wired at exactly 2 call sites, both calling `togglePlanMode(ctx.slashCtx)`:
+  `surface-setup.ts:~103` (persistent compositor) and `loop-iteration.ts:~206`
+  (readLine fallback). `reader.ts`/`terminal-compositor*` only forward the callback.
+- `toggleAfkMode` (`afk-mode-toggle.ts`) owns AFK's enter/exit machinery
+  (push-budget reset, elicitation→ledger swap, abort-watcher start/stop,
+  presence on/off) — all gated on `(sessionId && swap && fallback)`. These live
+  ONLY there, not in `session.setPermissionMode` (`agent-session.ts:750`) nor in
+  any provider `setPermissionMode` (anthropic-direct/openai-compatible/router).
+  A raw `setPermissionMode('autonomous')` would leak the abort-watcher + strand presence.
+- `togglePlanMode` and `/bypass` are pure (setPermissionMode + stats mirror +
+  repaint + copy; no teardown).
+- `plan-mode-addendum.test.ts` asserts `'/plan off'` literal but does NOT snapshot
+  the `Shift+Tab` parenthetical — so rewording it stays green.
+- `buildPrompt` (`repl-loop-shared.ts:~59`) renders a marker for `plan` and
+  `autonomous` but NOT `bypassPermissions`.
+- `repl-loop-wiring.test.ts:64` + `loop-iteration.test.ts:73` both
+  `vi.mock('../../plan-mode-toggle.js')`. NUANCE: neither test actually *fires*
+  the `onShiftTab` handler (FakeInputSurface stores/ignores it), so repointing
+  won't break them — swapping the mock target is defensive (hermeticity).
+
+## Step-by-step changes
+1. **New `src/cli/permission-mode-cycle.ts`** — `PERMISSION_CYCLE =
+   ['default','plan','bypassPermissions']` + `cyclePermissionMode(ctx)`:
+   - `cur === 'autonomous'` → `await toggleAfkMode(ctx, false)` (clean exit to default); return.
+   - else `idx = indexOf(cur)`; `next = ring[(idx+1) % len]` (out-of-ring → idx -1 → default).
+   - `setPermissionMode(next)` + mirror `stats.permissionMode` + `repaintStatusLine()` + per-mode copy.
+   - On `setPermissionMode` rejection: leave `stats` unchanged, surface via `ctx.out.error` (mirrors `togglePlanMode` contract).
+2. **Repoint both `onShiftTab` handlers** (`surface-setup.ts`, `loop-iteration.ts`)
+   from `togglePlanMode(ctx.slashCtx)` → `cyclePermissionMode(ctx.slashCtx)`; update inline comments.
+3. **`buildPrompt`** (`repl-loop-shared.ts`) — add a `⚡ bypass` marker for prompt parity (status line already shows the chip).
+4. **Doc/tip/addendum updates** (no behavior):
+   - `loading-tips.ts:56` — "Shift+Tab cycles permission modes (default → plan → bypass)".
+   - `input/types.ts:71-73` — `onShiftTab` doc comment.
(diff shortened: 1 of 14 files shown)


griffinwork40:
All three dimensions are in. Cross-agent coverage even filled each other's gaps (the security agent's "didn't check error-path state safety" was confirmed safe by the integration agent's atomicity check; the integration agent's questions about `palette.bypass`/status-line visibility were independently confirmed by security). Here's the synthesized review.

---

# Review: PR #225 — Shift+Tab cycles permission modes

**Recommendation: ✅ Approve / merge.** No blockers. This is a tightly-scoped, unusually well-documented change (shadow-verified plan, devils-advocate alternatives recorded). The headline risk — a keypress now escalates into `bypassPermissions` — is genuinely mitigated, and verified so against the live codebase.

## Verdicts by dimension
| Dimension | Verdict | Evidence |
|---|---|---|
| Correctness / integration | **PASS** | 1 nit |
| Security | **Acceptable-with-caveat** | central safety claim confirmed |
| Test coverage | **Gaps acceptable** | new unit test 7/7 green; 1 major gap is pre-existing |

## What was verified against the actual tree (not just the diff)
- `palette.bypass` exists and is callable — `palette.ts:53` (`chalk.bold.hex('#FF6AC1')`). No compile/runtime risk despite not being added in this diff.
- `SlashContext` shape matches every field touched (`stats.permissionMode` writable, `session.current.setPermissionMode → Promise`, `ui.repaintStatusLine`, `out.*`) — `slash/types.ts`.
- The AFK-exit path is sound: `toggleAfkMode(ctx, false)` runs full teardown (abort-watcher stop, elicitation restore, presence clear) **and** mirrors `stats.permissionMode = 'default'` itself — `afk-mode-toggle.ts:50–125`. The cycle correctly skips its own stats-mirror on this branch.
- No orphaned callers: post-diff `togglePlanMode` is referenced only by `/plan` (`slash/commands/plan.ts`); both Shift+Tab sites are repointed; `togglePlanMode` is not now-dead.
- Rejection is atomic: `setPermissionMode` updates session metadata only after the provider call resolves (`agent-session.ts:750`), so the catch leaving `stats.permissionMode` unchanged keeps session+stats consistent.
- **Security claim confirmed:** `bypassPermissions` IS the new-install default — `config.ts:342 DEFAULT_CLI_PERMISSION_MODE`. Cycling into bypass is therefore not novel exposure for the overwhelming majority of users.
- **Bypass is visible:** persistent status-line chip `⚡ BYPASS` marked "never drop" (`status-line.ts:370–371`) **plus** the new prompt marker (`repl-loop-shared.ts`, present in this diff).
- The cited "self-escalation vector" guard (`settable-keys.test.ts:95`) constrains the **agent** writing `permissionMode` via `/config set` — it does not govern a human keypress, so Shift+Tab circumvents nothing.

## Findings

**Major — pre-existing, NOT introduced by this PR (no action required to merge)**
- The Shift+Tab→`cyclePermissionMode` wiring has **zero end-to-end test**: both `FakeInputSurface` implementations store but never invoke `onShiftTab`, and the `vi.mock` swap is purely defensive. The old `togglePlanMode` wiring had the same gap (grep finds zero `onShiftTab(` invocations in tests). Worth a follow-up that fires the handler and asserts the cycle is called — but this PR doesn't regress it.

**Minor (polish, author's discretion)**
- Entry-friction asymmetry: `/bypass` shows a verbose multi-line warning banner (`slash/commands/bypass.ts`), while cycling in via Shift+Tab emits only a one-line notice. For the minority who deliberately run in `default`, two presses escalates to full read/write-any-path with less friction. The one-liner *does* spell out what's disabled, and the mode stays visible — acceptable, and explicitly accepted in the plan's risk section.
- `repaintStatusLine` is asserted only on the `default→plan` transition (`permission-mode-cycle.test.ts:87`); the `plan→bypass` and `bypass→default` branches don't assert it.
- The `bypass→default` test asserts `…toContain('default')`, which passes trivially if "default" appears anywhere in output — looser than the pinned `'plan mode on'` / `'bypass on'` strings used elsewhere.

**Nits**
- `permission-mode-cycle.ts:497` — the `?? 'default'` after `(idx + 1) % PERMISSION_CYCLE.length` is dead code (index is always 0/1/2). Harmless belt-and-suspenders for `noUncheckedIndexedAccess`; could drop or keep.
- Out-of-ring test asserts `toggleAfkMode` non-invocation, but the `bypass→default` wrap test omits the same check.

## Not checked
- Provider-side `setPermissionMode` impls (anthropic-direct / openai-compatible) — only the session wrapper was read.
- Full run of the two wiring test files (timed out within budget); the focused unit test passed 7/7, and the mock swap is mechanical with no logic change.
- `plan-mode-addendum.test.ts` snapshot — the plan asserts the reworded parenthetical isn't snapshotted, so it stays green, but I didn't re-run it.

**Done**
- Reviewed PR #225 (+375/−33, 14 files) read-only across correctness/integration, security, and test-coverage via three parallel dimension agents, with two re-dispatched after the first compose run truncated their verdicts.
- Evidence: findings cite the live tree (`config.ts:342`, `palette.ts:53`, `status-line.ts:370–371`, `afk-mode-toggle.ts:50–125`, `agent-session.ts:750`, `permission-mode-cycle.test.ts`) and the diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/ef81ab93-d35b-44ee-ace0-f8ca944576a3/pr-225.diff`; focused unit suite ran 7 passed / 0 failed.
- Verdict: **Approve** — 0 blockers; 1 major (no e2e wiring test) is pre-existing and not regressed; 3 minor + 3 nits are optional polish.
- No files were modified, committed, or pushed, and no PR comment was posted — review is read-only by contract.

---

griffinwork40:
## 🚢 Follow-up pushed: `ef8fefc` — test hardening from `/review` feedback

The `/review` pass found **0 blockers** (approve/merge); this commit is test-only polish addressing its coverage findings.

### What changed (test-only, +54 / −1)
- **E2E wiring regression test** (`repl-loop-wiring.test.ts`): fires the *real* `surface-setup.ts` `onShiftTab` closure (captured off `armCompositor`) and asserts it routes to `cyclePermissionMode(ctx.slashCtx)` + rearms the status line. Closes the gap where the keypress→handler wiring had **no test at all** — a silent repoint back to `togglePlanMode` now fails.
- **`repaintStatusLine`** now asserted on the `plan→bypass` and `bypass→default` ring steps (was only `default→plan`).
- **`bypass→default` copy** assertion now pins the distinctive `"…restored."` text instead of the loose `"default"` substring.
- **`toggleAfkMode` non-invocation** asserted on the `bypass→default` wrap.

### Deliberately unchanged (rationale)
- The bypass one-line copy — intentional "full-power badge, not red alarm" design per the plan.
- The `?? 'default'` fallback in `permission-mode-cycle.ts` — **type-necessary** under `noUncheckedIndexedAccess`; `tsc` fails without it (the "dead code" label is runtime-only).

### Verification
- `pnpm lint` (`tsc --noEmit`): ✅ clean
- Focused: **21/21 pass** (`permission-mode-cycle` 7, `loop-iteration` 3, `repl-loop-wiring` 11)
- Full suite: 9778 passed / 12 skipped / 2 failed — the 2 failures (`config.test.ts:538` DEFAULT_SLOT_BINDINGS, `anthropic-direct.test.ts:1070` compact model-id) are a **pre-existing environmental baseline** unrelated to this branch (model-slot / model-id resolution; no REPL/permission code involved).

### Known remaining (minor)
- The `loop-iteration.ts` readLine-fallback `onShiftTab` path (non-TTY) is still not fired by a test; the primary persistent-compositor path is now covered.

PR #226, opened 2026-06-21
feat(improve): add `cards list --regressed` observability view

griffinwork40:
## Summary
- Adds `afk improve cards list --regressed` — a **read-side** view that surfaces resolved/deferred failure cards whose `lastSeen` is later than their latest triage note (fixes that silently regressed, or deferrals that got worse).
- **Why:** scan merges by slug and *preserves* `status` (this is correct — most resolved cards are resolved-as-expected/noise and recur by design; auto-reopening would re-spam them). The gap was observability: no way to see "triaged, then fired again" without re-introducing ticket spam. This closes that gap **without** changing merge semantics.
- A card is **regressed** iff: status is `resolved` or `deferred`, AND it has ≥1 triage note, AND `lastSeen` is strictly later than the latest note timestamp.
- Pure, unit-tested helpers `latestNoteAt` / `isRegressed` / `selectRegressed` + disk reader `listRegressedCards` (`card-writer.ts`); `--regressed` flag + `renderRegressedList` wired into `cards list` (`improve.ts`), composing with `--pattern` / `--severity` / `--status` / `--json`.
- **Non-goals (by design):** no scan/merge-semantics change, no schema fields added, no auto-reopen. Read-only throughout.

## Test results
- `pnpm test` (full vitest suite): **9794 passed / 12 skipped / 2 failed**. The 2 failures — `src/agent/providers/anthropic-direct.test.ts` (compact() haiku model-id) and `src/cli/config.test.ts` (DEFAULT_SLOT_BINDINGS) — are **pre-existing and unrelated**: they reproduce on a clean checkout in isolation and live in files disjoint from this change.
- `npx vitest run src/improve/scan/card-writer.test.ts`: **31/31 passed** — 14 new (pure `latestNoteAt`/`isRegressed`/`selectRegressed` + a filesystem round-trip that proves `status` is preserved across re-detection).

## Verification (`--verify` adversarial wave)
An independent read-only verifier re-derived the load-bearing claims from the diff alone — all **CONFIRM**:
- `mergeCard` is absent from the diff; `status` / `notes` preserved verbatim (`card-writer.ts:145-193`).
- Every new symbol is a pure read / console output — no card mutation or disk write; the round-trip test asserts `status` stays `resolved` after re-detection.
- `FailureCardSchema` unchanged; `RegressedCardEntry` is a derived view interface, never persisted.
- `isRegressed`: strict `>`, max-of-notes, and status+note gates all correct.
- Default `cards list` output is unchanged when `--regressed` is absent (early-return gate).

## Example
Run against a real local improve store, `--regressed` surfaced 9 cards — including `tool-failure-browser-open` (resolved, then re-fired = candidate regression) and `closure-anomaly-abort` (resolved, recurs by design). The latter is exactly the "resolved-as-expected" case that a future optional `resolution: fixed|expected` field would suppress, and illustrates why auto-reopen was deliberately avoided.

## Note for maintainer
Developed on the public mirror. Per the snapshot-sync workflow, this should also be ported to the private `afk-workshop` canonical so it survives the next public re-sync.

### DIFF
--- a/src/cli/commands/improve.ts
+++ b/src/cli/commands/improve.ts
@@ -9,8 +9,10 @@
  *       Scan witness traces, run every registered detector, print a
  *       summary. Default is DRY-RUN; pass `--write` to persist cards.
  *
- *   afk improve cards list [--pattern] [--severity] [--status] [--json]
- *       Tabular listing of all cards on disk.
+ *   afk improve cards list [--pattern] [--severity] [--status] [--regressed] [--json]
+ *       Tabular listing of all cards on disk. `--regressed` narrows to
+ *       resolved/deferred cards that fired again after their latest triage
+ *       note (a read-side observability view; never changes status).
  *
  *   afk improve cards show <slug> [--json]
  *       Print one card.
@@ -81,7 +83,8 @@ import {
   disabledByDefaultDetectorNames,
   type DetectorOptions,
 } from '../../improve/scan/detectors/index.js';
-import { writeCard, listCards, getCard } from '../../improve/scan/card-writer.js';
+import { writeCard, listCards, getCard, listRegressedCards } from '../../improve/scan/card-writer.js';
+import type { RegressedCardEntry } from '../../improve/scan/card-writer.js';
 import { renderMarkdown } from '../../improve/scan/card-writer.js';
 import { triageCard, TriageError } from '../../improve/triage.js';
 import type { CardStatus } from '../../improve/schemas.js';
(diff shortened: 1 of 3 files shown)


griffinwork40:
Load-bearing claim **verified at ref de0b2d48**: `note.at` (schemas.ts:80), `lastSeen`/`firstSeen` (schemas.ts:105-106), and `observedAt` (schemas.ts:133) all use `z.string().datetime()`, whose default `offset: false` rejects non-`Z` timestamps — so all persisted timestamps are UTC ISO-8601, where lexical order equals chronological order for distinct instants. The `selectRegressed` lexical sort is therefore correct in practice; the finding is correctly a latent-only `low`.

Wave 2 synthesis follows.

---

# Code Review — PR #226

**Target:** `feat(improve): add cards list --regressed observability view` · `improve/cards-regressed-view` → `main`
**Reviewed ref:** `de0b2d48f0c83453ddb43de0dc83bde52ee816e0` · **Regime:** full (433 lines, 3 files) · **Stated intent:** PR #226 title+body (read-only `--regressed` view; non-goals: no merge-semantics/schema/auto-reopen change)

## Summary
A clean, tightly-scoped, **read-only** feature. New exports (`RegressedCardEntry`, `latestNoteAt`, `isRegressed`, `selectRegressed`, `listRegressedCards`) are purely additive; no existing export's signature or behavior changed; `mergeCard` is untouched; default `cards list` output is unchanged when `--regressed` is absent (first-branch early return). The regression predicate matches the spec exactly (status gate → note gate → strict `>` epoch comparison), and the 14 new tests cover every gate plus a filesystem round-trip proving `status` is preserved across re-detection (no auto-reopen). Spec-compliance: satisfied, no scope creep.

## Findings (2 — both non-blocking)

**low · correctness · `src/improve/scan/card-writer.ts:478-481` · ref:de0b2d48 · file-state**
`selectRegressed` sorts `lastSeen` by **lexical** string comparison while `isRegressed` gates on **`Date.parse` epoch** comparison — two different comparison semantics for the same field.
```
if (a.lastSeen !== b.lastSeen) return a.lastSeen < b.lastSeen ? 1 : -1;   // sort: lexical
return last > note;                                                        // gate: epoch
```
Benign today: `z.string().datetime()` (offset:false) forces UTC-`Z` form on every write path (verified schemas.ts:80,105-106,133), so lexical order == chronological order for distinct instants. Latent hazard only if the schema is ever relaxed to accept offsets. *Suggestion:* use `Date.parse(a.lastSeen) - Date.parse(b.lastSeen)` and add an `// Invariant: z.string().datetime() forbids offsets → lexical == epoch` comment.

**low · test-coverage · `src/improve/scan/card-writer.test.ts` · ref:de0b2d48 · absence-claim (confirmed-absent, grep cited)**
`--status open` combined with `--regressed` is a contradiction (the regressed set is always resolved/deferred) that silently returns empty rather than warning; no test documents this no-op contract. *Suggestion:* add one test asserting `selectRegressed([...]).filter(e => e.status === 'open')` returns `[]` to pin the silent-empty contract.

No `CONFLICT` blocks (agents agreed). No critical/high/medium findings → `/shadow-verify` not triggered.

## Merge decision: **MERGE**

Both findings are `low` and non-blocking — one is a latent-only consistency nit guarded by the schema, the other a missing test for an already-correct silent no-op. The change is additive, read-only, preserves merge semantics as designed, and is well-tested. Optionally address the two `low` suggestions in a follow-up; neither gates merge.

## What was not checked
- **Refs read:** Wave 1 agents read file state via `git show de0b2d48:<file>`; the load-bearing schema claim was independently re-read by the coordinator at `de0b2d48`. All blocking-tier citations verified at that ref (none survived above `low`).
- **Stated intent:** PR #226 title+body available — **spec-compliance was assessed** (satisfied, no creep).
- **Not run:** the test suite was not executed in this review; the PR body reports 31/31 on `card-writer.test.ts` and full-suite 9794 pass / 2 pre-existing unrelated fails (`anthropic-direct.test.ts`, `config.test.ts`) — consistent with this session's memory of those two known-flaky env-bleed failures.
- **Not exercised:** the CLI was not run at runtime; `renderRegressedList` table formatting/`--json` output verified by reading only.
- **Maintainer note (from PR body, not a review finding):** developed on the public mirror; per the snapshot-sync workflow it should also be ported to the private `afk-workshop` canonical to survive the next public re-sync.

**Done** — Review complete; merge recommendation is **MERGE** with 2 non-blocking `low` suggestions. This skill is read-only: no files, commits, or PR state were modified, and the dirty working tree was left untouched.

---

PR #229, opened 2026-06-21
fix(cli): markdown rendering — task-list checkboxes, width-aware rules, wide-char dropdown

griffinwork40:
Three user-visible TUI rendering bugs found in a UX audit of the REPL.

### Fixes
1. **GFM task-list checkboxes rendered as raw `[x]`** — `marked` v17 emits a `checkbox` token that fell through to `default: token.raw`, so `- [x] done` showed as `• [x] done`. Now renders `☑ done` / `☐ todo` (bullet replaced by the glyph; the raw `checkbox` token is filtered out of the item tokens). `src/cli/formatter.ts`
2. **Horizontal rule (`---`) hardcoded to 40 columns** — looked stubby on wide terminals and wrapped (corrupting the compositor row count) on narrow ones. Now width-aware via `maxTableWidth`, falling back to 40 for callers that omit a width. `src/cli/formatter.ts`
3. **Autocomplete dropdown soft-wrap height miscounted CJK/emoji** — used UTF-16 `.length` instead of display columns (`cols`), producing ghost/clipped rows when completing wide-char filenames. Now uses the already-imported `displayWidth`. `src/cli/terminal-compositor.render.ts`

### Tests
`src/cli/formatter.test.ts` (+ checkbox & hr cases) and new `src/cli/terminal-compositor.render.test.ts`. **63 tests pass; `pnpm lint` clean.**


griffinwork40:
The three-wave DAG completed. No critical/high findings, so the skill's post-synthesis `/shadow-verify` gate is not triggered — medium and below pass straight through. Here is the final synthesized review.

---

## PR #229 Review: fix(cli): markdown rendering — task-list checkboxes, width-aware rules, wide-char dropdown

**Reviewed ref:** `7c81a2ac8de4a5f06e1afaac5f0568f8901a3112`
**Stated intent:** Fix task-list checkboxes, make horizontal rules width-aware, handle wide-char (CJK) candidates in the dropdown compositor.
**Regime:** light (168 lines, 4 files)

---

### Merge Decision: **DO NOT MERGE**

Two medium findings block merge: the ordered-list case of the checkbox fix introduces exactly the bug it was meant to fix, and the only CJK test is non-discriminating — the fix's correctness is not demonstrated.

---

### Findings

#### Critical — none

#### High — none

#### Medium

**M1 · correctness · `formatter.ts:231–235`** — Ordered task-list items silently drop the checkbox glyph.

The prefix ternary evaluates `list.ordered` first and short-circuits before the `isTask` branch. Any ordered list item that is also a GFM task (`1. [x] done`) renders as `  1. done` — the ☑/☐ glyph is lost with no error or fallback. The `checkbox` token is still filtered from `renderableTokens`, so the raw `[x]` is removed too — the user gets neither glyph nor bracket.

```typescript
const prefix = list.ordered
  ? `  ${startNum + i}. `
  : isTask
    ? (item.checked ? '  ☑ ' : '  ☐ ')
    : '  • ';
```

**Fix:** Promote `isTask` above the `ordered` check, or nest: `list.ordered ? (isTask ? \`  ${n}. ☑ \` : \`  ${n}. \`) : ...`

- *Citation verified at reviewed ref. Absence of test for `1. [x]` pattern confirmed via git grep.*

---

**M2 · test-coverage · `terminal-compositor.render.test.ts:53–65`** — CJK soft-wrap test is non-discriminating.

The test uses `cols=80` with `'東京都'` (display width 6). At 80 columns both `displayWidth` and naïve `.length` produce `softWraps=0`, so `toHaveLength(1)` passes regardless of which implementation runs. The inline test comment itself admits this. A regression in the `displayWidth` logic would pass this test undetected.

```typescript
const host = makeHost(80, [{ value: '東京都', summary: 'CJK candidate' }]);
const rows = renderDropdownRows(host);
// At 80 cols, even 2*3=6 display cols fits; no placeholder expected.
const nonBlank = rows.filter((r) => r.length > 0);
expect(nonBlank).toHaveLength(1);
```

**Fix:** Add a test with `cols < displayWidth-of-CJK` (e.g. `makeHost(5, ...)` where `.length=3 < 5` gives `softWraps=0` wrong, but `displayWidth=6 > 5` gives `softWraps=1` correct) — proving the fix is load-bearing.

- *Citation verified at reviewed ref. Absence of narrow-terminal discriminating test confirmed via git grep.*

---

#### Low

**L1 · correctness · `formatter.ts:288–294`** — Stale "Cap at maxTableWidth" comment.
Comment says `// Cap at maxTableWidth so the rule never exceeds the compositor's row budget.` but `const hrWidth = maxTableWidth ?? 40` performs no capping — it's a direct use-or-fallback. If `maxTableWidth` is large, the rule will exceed what the comment promises.
**Fix:** Rewrite comment as "Use maxTableWidth directly — it is already the compositor's row budget."

**L2 · test-coverage · `terminal-compositor.render.test.ts:79–86`** — Two vacuous assertions in the third soft-wrap test.
`expect(blanks.length).toBeGreaterThanOrEqual(0)` (always true) and `expect(row).toBeTruthy()` on a filtered-non-empty array (always true) provide zero coverage signal.
**Fix:** Replace with structural assertions like `expect(rows.length).toBeGreaterThanOrEqual(1)` and `expect(rows.filter(r => r !== '').every(r => r.includes('/test-cmd'))).toBe(true)`.

**L3 · test-coverage · `formatter.test.ts:534–563`** — No test for hr default-fallback width.
Four tests exist in the hr block (two explicitly test `maxWidth=60` and `maxWidth=80`), but none calls `renderMarkdownToTerminal('---\n')` without opts and asserts `length === 40`. The fallback constant is untested.
**Fix:** Add `it('rule defaults to 40 dashes when no maxWidth is provided', ...)`.
- *[Evidence corrected by Wave 1.5: describe block has 4 tests, not 2; absence of fallback test confirmed independently.]*

**L4 · test-coverage · `formatter.ts:236–239` (absence across test suite)** — No test for task items with rich inline formatting.
The checkbox filter strips task tokens then calls `renderTokens` on the remainder. No test exercises a task item with bold/code (e.g. `- [x] **bold** item`). All three task-list tests use plain text only.
**Fix:** Add a test with inline markup in a task item.

---

#### Nit — none

---

### Spec-Compliance Assessment

| Stated goal | Status |
|---|---|
| Task-list checkboxes | **Partially broken** — unordered task items fixed; ordered task items (`1. [x]`) silently lose glyph (M1) |
| Width-aware horizontal rules | **Implemented** — fallback comment misleading (L1), fallback path untested (L3) |
| Wide-char dropdown | **Implementation present** — only test is non-discriminating (M2); fix correctness not demonstrated |

No scope creep. Changes contained to the stated areas.

---

### Conflicts — none

### Dropped/Downgraded Findings — none dropped. L3 evidence corrected (4 tests in block, not 2); severity unchanged at low.

---

### What Was Not Checked
- Read against branch HEAD `7c81a2ac8de4a5f06e1afaac5f0568f8901a3112` — all citations verified at that ref.
- Stated intent: PR #229 title+body — spec-compliance assessed.
- `displayWidth` runtime behavior under edge cases (ZWJ, variation selectors, emoji) not checked — only the test coverage gap was assessed.
- Whether `maxTableWidth` can realistically exceed terminal width in practice (determines whether L1 is a latent bug or documentation-only).
- Tests not run (read-only review).

---

**Working tree note:** The review was entirely read-only — no files were edited, no git mutations, no stashing. The 3 uncommitted changes in the working tree are untouched.

---

griffinwork40:
## Review findings resolved — `0aaa671`

All 6 findings from the `/review` are addressed.

| # | Resolution |
|---|---|
| **M1** (correctness) | Fixed. Ordered task items (`1. [x] done`) now keep **both** the number and the ☑/☐ glyph (`1. ☑ done`). The prefix ternary now re-tests `isTask` inside both the ordered and unordered branches. |
| **L1** (correctness) | Fixed. Rewrote the stale "Cap at maxTableWidth" hr comment — the code is a use-or-fallback (`maxTableWidth ?? 40`), no capping. |
| **L2** (test) | Fixed. Replaced the two vacuous assertions (`>= 0`; `toBeTruthy()` on already-non-empty rows) with structural ones: exactly one row, zero blank placeholders, value preserved. |
| **L3** (test) | Fixed. Added a test for the hr 40-dash default fallback (no `maxWidth` opt). |
| **L4** (test) | Fixed. Added a test for a task item with inline formatting (`**bold**` + `` `code` ``). |
| **M2** (test) | Addressed differently — see below. |

### M2 — note on the deviation

The finding (CJK soft-wrap test is non-discriminating) is **valid**, but a test that distinguishes the `displayWidth` measurement from a naïve `.length` via the soft-wrap **count** is **not constructable** at this layer, and I verified this empirically before writing the replacement:

- `formatDropdownRow` truncates every row to `min(cols-4, 60)` display columns — always `< cols` — so `softWraps` is **0** for any well-formed CJK row regardless of which width measure line ~172 uses.
- The only rows that actually exceed `cols` (and wrap) are emoji, which evade truncation via a separate surrogate-pair miscount in `truncateToWidth` — and for emoji `.length ≈ displayWidth`, so they don't discriminate either.
- A sweep over `cols` 41–200 × CJK/emoji/ASCII found **zero** inputs where `.length`- vs `displayWidth`-based `softWraps` differ.
- The review's suggested `makeHost(5, …)` also can't work — `cols <= 40` hits the early `return []` guard.

So the `displayWidth` source change is kept (it's the correct, defensive measure — it matches `cols`, which is display columns — even though its effect is currently unobservable), and the test now pins the **real observable contract**: wide content that fits the budget produces exactly one row with no phantom wrap placeholders and stays within the terminal width. A comment documents why the count-discriminating test isn't constructable.

> Latent finding (out of scope, **not** changed here): `truncateToWidth` (`src/cli/input/dropdown.ts`) iterates UTF-16 code units, so surrogate-pair emoji evade its width cap and produce genuinely over-wide dropdown rows. That's the real path that can blow the soft-wrap budget; worth a separate issue.

**Gates:** targeted tests pass (68); `pnpm lint` clean. Full suite: 9808 pass / 12 skip / 1 pre-existing unrelated failure (`anthropic-direct` compact model-alias, untouched by this PR).

PR #232, opened 2026-06-21
fix(cli): timelier context-usage signal & overlay-safe overflow warning

griffinwork40:
Status-feedback fixes found in a UX audit (relevant for AFK users who step away during long turns).

### Changes
- **Overlay-safe overflow warning** — the `context OVER … model output may be silently truncated` message used raw `console.log`, which strands output above the live compositor overlay (the exact footgun the surrounding comment warns about). Now uses the injected overlay-safe `write`, matching the `> 0.5` branch directly below it. `src/cli/commands/interactive/turn-handler.ts`
- **Timelier context bar** — refresh throttle `15s → 3s` so the context-usage indicator doesn't go stale mid-turn. `src/cli/commands/interactive/turn-handler.ts`
- **Earlier elapsed clock** — `ELAPSED_GRACE_MS 5s → 2s` so common 2–4s turns show elapsed time. `src/cli/terminal-compositor.types.ts`

### Tests
turn-handler, context-bar — **71 tests pass; `pnpm lint` clean.**

**Deferred:** a one-time ≥90% context notice — it needs edge-detection state that spans multiple files, out of scope for this small PR.

### DIFF
--- a/src/cli/commands/interactive/turn-handler.test.ts
+++ b/src/cli/commands/interactive/turn-handler.test.ts
@@ -1810,7 +1810,7 @@ describe('runTurn — mid-turn context progress', () => {
   });
 
   // Two tool_use+tool_result pairs with the clock frozen: the first fires,
-  // the second is suppressed (within the 15 s throttle window).
+  // the second is suppressed (within the 3 s throttle window).
   it('fires onContextProgress on the first tool_result', async () => {
     vi.setSystemTime(1_000_000);
 
@@ -1831,7 +1831,7 @@ describe('runTurn — mid-turn context progress', () => {
 
     await runTurn({ text: 'q', attachments: [] }, session, stats, h);
 
-    // First fires, second suppressed (clock didn't advance past 15 s).
+    // First fires, second suppressed (clock didn't advance past 3 s).
     expect(onContextProgress).toHaveBeenCalledTimes(1);
   });
 
--- a/src/cli/commands/interactive/turn-handler.ts
+++ b/src/cli/commands/interactive/turn-handler.ts
@@ -86,7 +86,7 @@ export async function runTurn(
   // the latest write even though the assignment happens in a microtask.
   const pickerRef: { abort: AbortController | null } = { abort: null };
   let lastContextProgressMs = 0;
-  const CONTEXT_PROGRESS_MIN_INTERVAL_MS = 15_000;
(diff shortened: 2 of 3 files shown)


griffinwork40:
## PR #232 Review — `fix/tui-context-visibility`

**Verdict: ✅ LGTM — clean, well-scoped, three independent UX fixes with no behavioral risk.**

### Change-by-change

#### 1. `CONTEXT_PROGRESS_MIN_INTERVAL_MS`: 15s → 3s (turn-handler.ts:89)

The throttle on `onContextProgress` (context-usage status bar refresh) was 15 seconds — meaning during a fast multi-tool turn the user could go 15s without seeing context usage update. 3s is a meaningful improvement in feedback timeliness without risking perf (the callback is just a status-bar repaint). Test comments updated to match; no test logic changes needed since the tests freeze the clock with `vi.setSystemTime()`.

#### 2. `console.log` → `write` in the "context OVER" branch (turn-handler.ts:777)

**This is the load-bearing fix.** `printTurnFooter` already accepts a compositor-aware `write` parameter specifically to avoid corrupting log-update's line tracker (the JSDoc on lines 754–759 explains this at length). The "context OVER" error line was the *one* output in the function still bypassing `write` with a raw `console.log`. When context exceeded the limit during a compositor-managed turn, this would strand the verdict/footer above the live overlay. Now all output paths in the function go through the same `write` sink. Correct and consistent.

#### 3. `ELAPSED_GRACE_MS`: 5s → 2s (terminal-compositor.types.ts:66)

Pure UX tuning — the elapsed-time indicator next to the spinner now appears after 2s instead of 5s. Short enough to avoid flashing for quick ops, soon enough to give timely feedback on longer ones. No downstream logic depends on the specific value beyond the `formatElapsed` pure function.

### Notes

- No test coverage gap introduced — the throttle test already validates the mechanism (fires once, suppresses within window), and the constant change doesn't alter the test logic.
- The `console.log` → `write` fix has no dedicated test for the context-over branch, but the fix is a one-word change to match the pattern already used by every other output line in the same function. Low risk.
- All three changes are additive/constant-only — no API surface changes, no breaking behavior.

**Status: Done** — PR #232 is a clean, low-risk set of TUI visibility improvements. The `console.log` → `write` fix is the most important change as it prevents terminal corruption in the context-over edge case.

---

PR #235, opened 2026-06-21
fix(anthropic-direct): terminate query generator on interrupt during usage-limit wait

griffinwork40:
## Summary

Fixes an interactive-REPL hang triggered when the user presses **Ctrl+C during the Anthropic usage-limit wait** ("out of usage"). After the interrupt the session wedged: it couldn't quit, auto-resume never re-fired, and newly typed messages just showed **"queued"** forever.

**Root cause:** In `AnthropicDirectQuery[Symbol.asyncIterator]` (`query.ts`) the per-turn loop had an abort guard **only on the `catch` path**. But an interrupt landing during the usage-limit wait makes `turnWithRetries` return **cleanly** (no throw — `retry-layer.ts`: `if (result === 'aborted') return`), so the catch never fired. The outer loop then fell through to `promptIterator.next()` and **blocked forever** while the consumer (`sendMessageStreamInternal`) was still awaiting a terminal event for the aborted turn → all three symptoms.

**Fix:** add a post-`finally` abort guard symmetric with the existing catch-path guard — when the turn was aborted, `return` so the consumer observes `{done:true}` and the turn unwinds to idle. (`abort.clear()` nulls the coordinator slot but does not reset `signal.aborted`, so the post-`finally` read is valid.)

Diagnosed via `/diagnose`; winning hypothesis confirmed by isolated-worktree validation.

## Test results
- `pnpm lint` (strict `tsc --noEmit`): **passed**
- New regression test `query-auth-retry.test.ts` › *"interrupt during wait"*: **red → green** (2.2s hang → 334ms clean termination). Drives `interrupt()` during the wait with a prompt stream that **blocks on its 2nd pull**, mirroring the live REPL idle wait — the existing `close()`-based abort test couldn't catch this.
- `pnpm test` (full suite): **9815 passed / 12 skipped / 1 failed**. The lone failure — `anthropic-direct.test.ts › compact()…` asserting a haiku model id (`claude-haiku-4-5-20251001` vs `claude-haiku-4-5`) — is **pre-existing on `main`** (model-version drift) in a file this PR does **not** touch; that file is byte-identical to `origin/main`.

## Verification (`--verify`)
Adversarial verifier re-derived the load-bearing claims from the diff alone:
- **Fix correctness** — CONFIRM: the clean abort-return is the previously-unguarded exit; the new guard is the missing terminator.
- **No regression on normal completion** — CONFIRM: `signal.aborted` is false on non-aborted turns, so the guard is a no-op and auto-compaction + the next-turn loop still run.
- **Signal readable after clear** — CONFIRM: `abort-coordinator.clear()` only nulls the slot; never resets the signal.
- Regression sweep: the outer `finally` (`promptIterator.return`) still runs on the early return; no path wants to keep looping while aborted. **Verdict: SAFE TO MERGE.**

## Follow-ups (out of scope — separate bugs surfaced during diagnosis, not fixed here)
1. **Transcript "(no response recorded)" data loss** on interrupt/crash: an interrupted turn never calls `appendTurn` (gated on `doneFired`), so in-flight assistant text isn't persisted; `closeDanglingTurn` writes the placeholder on next startup. Needs an `appendPartialTurn` path.
2. **Concurrent-session usage-limit dedup**: a 2nd session awaiting the shared `usageLimitWaitPromise` doesn't register its own abort signal with `waitForReset`, so its interrupt wouldn't break the wait.

### DIFF
--- a/src/agent/providers/anthropic-direct/query-auth-retry.test.ts
+++ b/src/agent/providers/anthropic-direct/query-auth-retry.test.ts
@@ -690,6 +690,81 @@ describe('AnthropicDirectProvider — turnWithUsageLimitRetry', () => {
     expect(messagesCreateMock).toHaveBeenCalledTimes(1);
   });
 
+  it('interrupt during wait: 429+ts → paused → interrupt() → generator terminates (no hang, no resumed)', async () => {
+    // Regression guard for the usage-limit hang. When Ctrl+C (interrupt) lands
+    // during the usage-limit wait, `turnWithRetries` returns CLEANLY (not via
+    // throw — see retry-layer's `if (result === 'aborted') return`), so the
+    // catch-path abort guard in query.ts never fires. Without the post-finally
+    // guard the outer loop falls through to the next `promptIterator.next()`
+    // and BLOCKS there forever while the consumer is still awaiting a terminal
+    // event for the aborted turn — the REPL hangs (can't quit, auto-resume
+    // never re-fires, queued input never drains).
+    //
+    // The prompt stream below blocks on its 2nd pull, exactly like the live
+    // REPL idle wait, so the hang is observable. A single-shot stream would
+    // return `done` on the 2nd loop and mask the bug. Uses REAL timers because
+    // the abort short-circuits waitForReset synchronously (the signal is
+    // already aborted when waitForReset is entered) and we want a wall-clock
+    // failsafe that fires if the hang regresses.
+    vi.useRealTimers();
+
+    let releaseSecondPull: (() => void) | undefined;
+    const blockingPrompt = (async function* (): AsyncIterable<{ content: string }> {
+      yield { content: 'hello' };
+      // 2nd pull blocks until released — mirrors the REPL awaiting next input.
+      await new Promise<void>((resolve) => { releaseSecondPull = resolve; });
+    })();
+
+    let callIdx = 0;
+    messagesCreateMock.mockImplementation(() => {
+      callIdx += 1;
+      if (callIdx === 1) throw make429UsageLimitError(60 * 60 * 1_000); // 1h reset
+      return fromArray(makeTextStream('should not reach'));
+    });
+
+    const provider = new AnthropicDirectProvider();
+    const query = provider.query({
+      prompt: blockingPrompt,
+      config: { model: 'claude-sonnet-4-5-20250929', apiKey: 'sk-ant-oat01-test', autoResumeOnUsageLimit: true },
+    });
+
+    const events: ProviderEvent[] = [];
+    const collectWithInterrupt = async (): Promise<void> => {
+      for await (const ev of query) {
+        events.push(ev);
+        if (ev.type === 'paused') {
+          // User hits Ctrl+C during the usage-limit wait.
+          await query.interrupt();
+        }
+      }
+    };
+
+    // Fail fast and legibly if the hang regresses: once interrupted, the
+    // generator must terminate well within this window.
+    let failsafeTimer: ReturnType<typeof setTimeout> | undefined;
+    const failsafe = new Promise<never>((_resolve, reject) => {
+      failsafeTimer = setTimeout(
+        () => reject(new Error(
+          'HANG: query generator did not terminate after interrupt() during the usage-limit wait — regression of the query.ts post-finally abort guard',
+        )),
+        2_000,
+      );
+    });
+
+    try {
+      await Promise.race([collectWithInterrupt(), failsafe]);
+    } finally {
+      if (failsafeTimer !== undefined) clearTimeout(failsafeTimer);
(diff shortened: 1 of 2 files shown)


griffinwork40:
All claims verified. The test's timer scoping is sound: `beforeEach` installs fake timers, `afterEach` restores real timers, and the test's inline `vi.useRealTimers()` override is cleaned up by `afterEach` — no leak to subsequent tests. The `anthropic-direct.test.ts` pre-existing failure file is confirmed byte-identical to `origin/main` (empty diff), corroborating the PR's claim that the lone suite failure is unrelated and untouched.

No critical/high/medium findings surfaced, so Wave 1.5 citation/absence backstop has nothing to verify and shadow-verify is not triggered (skill: medium-and-below go straight through). Synthesizing inline.

---

# Code Review — PR #235

**Target:** PR #235 `fix(anthropic-direct): terminate query generator on interrupt during usage-limit wait` · branch `afk/fix-usage-limit-hang` → `main` · reviewed ref `c462ebd`
**Change type:** hotfix · **regime:** light · 2 files · +90/-0
**Stated intent:** PR #235 title + body — **spec-compliance assessed.**

## Summary

A two-line-of-logic hotfix: adds a post-`finally` abort guard (`if (controller.signal.aborted) return;`) in `AnthropicDirectQuery[Symbol.asyncIterator]` (`query.ts:399`), symmetric with the existing catch-path guard at `:378`, plus a 75-line regression test. The guard terminates the query generator when an interrupt lands during the usage-limit wait — the one abort path that makes `turnWithRetries` return *cleanly* (no throw) so the catch never fires.

All seven load-bearing claims were verified twice — once by the Wave 1 agent, then independently re-derived by the coordinator against the reviewed ref.

## Findings

**critical / high / medium:** none.

| sev | conf | dimension | location | finding | suggestion |
|-----|------|-----------|----------|---------|------------|
| nit | high | test-coverage | `query-auth-retry.test.ts:709` | `vi.useRealTimers()` overrides the describe-level `beforeEach` fake timers; correctly scoped (afterEach restores), but the override-intent isn't called out | the existing inline comment explains *why* real timers; optionally add that it overrides the `beforeEach` fake-timer install |
| nit | high | test-coverage | `query-auth-retry.test.ts:744-752` | the 2000ms failsafe is an undocumented wall-clock magic number; flake-safe because the abort short-circuits `waitForReset` synchronously | document the 2000ms as conservative margin, or lower it |

Both are documentation-grade nits, non-blocking, and partly pre-addressed by existing inline comments.

## Claim verification (all CONFIRMED at ref `c462ebd`)

1. **Loop structure** — `query.ts:296` `while` → `:306` fresh `controller` → `:360-384` try/catch/finally → `:399` new guard → `:401-432` auto-compaction. ✓
2. **Symmetric catch-path guard** — `query.ts:378` `if (controller.signal.aborted) return;`. ✓
3. **Clean return on abort** — `retry-layer.ts:266` & `:362` `if (... === 'aborted') return;` (no throw → catch skipped → reaches new guard). ✓
4. **`clear()` preserves `signal.aborted`** — `abort-coordinator.ts:109` compare-and-clear only nulls the slot; `AbortSignal.aborted` is irreversible by Web spec. ✓
5. **No-op on normal turns (key safety claim)** — fresh `AbortController` per iteration (`abort-coordinator.ts:91`); `signal.aborted` false unless a `pendingReason` was parked (`:94`); guard is a guaranteed no-op on non-aborted turns. ✓
6. **Cleanup on early return** — `return` at `:399` unwinds through the outer `finally` `:437-443` (`promptIterator.return()`); skipping auto-compaction on an aborted turn is correct (compaction mutates `state.messages` and must run only at a clean boundary). ✓
7. **Test is a genuine regression guard** — blockingPrompt's 2nd pull hangs without the fix; the 2000ms failsafe would reject → red. With the fix, `return` ends the generator → green. Assertions meaningful (`paused` present, `resumed` absent, exactly 1 API call). ✓

**Spec-compliance verdict:** intent fully met; minimal, correctly-placed fix; no scope creep. The two documented follow-ups (transcript data-loss on interrupt; concurrent-session usage-limit dedup) are explicitly out of scope.

## What was not checked
- Read against branch HEAD `c462ebd` (== worktree, clean) — citations verified at that ref.
- Did not run the suite; relied on PR-reported results (new test red→green; lone pre-existing failure in `anthropic-direct.test.ts` confirmed byte-identical to `origin/main`).
- Coverage gap (non-blocking): no parallel `no-ts` (no-timestamp 429) interrupt-during-wait test exists; the guard is path-agnostic, so the single timestamp-path test adequately guards the mechanism.
- Did not exercise real REPL signal delivery (unit-level mocks only).

Clean, minimal, well-tested hotfix that resolves the stated hang at the correct layer with a symmetric guard and a true red→green regression test. No blocking findings; the only two notes are documentation nits.

---

Issue #237, opened 2026-06-21
Proposal: a main-session post-turn hook to make terminal-state / behavioral policy code-enforced (prompt-only today)


amatayomosley-web:
I read through the harness — the system prompt, the AFK-contract / failure-geometry / architecture docs, and both audits (`orchestration-pressure`, `failure-geometry`) — and checked the claims below against source. This proposal builds directly on a gap your **own** failure-geometry audit names but doesn't propose closing.

## The gap (your failure-geometry audit, §B / Q6)

Your control-plane matrix splits enforcement cleanly:

- **Code-enforced (structural topology):** nesting depth, tool gating, hook `block`, abort cascade.
- **Prompt-only or absent (behavioral policy):** irreversible-action intent, terminal-state mandate, delegation threshold — and, in the audit's words, *"the UI renders a verdict card but never rejects/re-prompts when [a terminal state] is absent."*

For an **unattended** harness, that prompt-only half is the highest-stakes unguarded surface — because the failures that bite when no one is watching are behavioral, not structural:

1. **Fake completion** — the model declares `Done` while a blocker remains. The product promise is "get pinged when it finishes"; a false "finished" ping is the most expensive failure in the system, because the operator acts on one line, on their phone, away from the trace.
2. **Deferral-when-reversible** — ends a turn in `Asking` for something it had clear authority to do. Unattended, an unnecessary `Asking` wedges the session until a human who is *by definition AFK* returns. (The routing-directive already prompts "proceed on a stated assumption" — but per Q6 that's prompt-only.)
3. **Unverified `Blocked`** — reports a cause it guessed rather than verified.

You can lower the rate of these with prompt text; you can't *enforce* against them in text, because the model grades its own homework.

## The proposal — ~90% of the mechanism already exists

Three parts are already in the tree; they're just not wired to the main session's turn boundary:

- **The re-prompt primitive exists.** `HookDecision.injectContext` (`src/agent/hooks.ts:54-62`) queues framework-generated feedback into a session's *next* turn — exactly "bounce the turn back with a correction." But it's wired **`SubagentStop`-only**; the doc says *"Ignored for all other hook events."*
- **The verdict extractor exists.** `src/cli/commands/interactive/terminal-state.ts` already parses `Done / Blocked / Asking / Interrupted` from the trailing message — but it's a pure, **display-only** function; it renders the verdict card and never rejects.
[...]

PR #238, opened 2026-06-21
feat(afk): opt-in "Done" verification for AFK Telegram pushes (#237)

griffinwork40:
## What & why

Implements the **Telegram corollary** from #237.

In AFK autonomous mode the REPL pushes each turn's parsed terminal state to Telegram (`pushTerminalStateToTelegram` → `afk-push.ts`) so the away operator can review asynchronously. As #237 notes, a `✅ Done` push the model **self-certified** — with no file write, edit, or successful command behind it — is the most expensive failure in the system: the operator acts on one line, on their phone, away from the trace.

This makes that one notification honest, **opt-in and default-off**, without touching control flow.

## Behaviour

New flag `telegram.verifyDone` in `afk.config.json` (default off / absent):

```jsonc
{
  "telegram": {
    "verifyDone": true
  }
}
```

When **on**, at the existing autonomous-mode push site (`turn-handler.ts`), a `Done` verdict whose turn produced **no corroborating evidence** is relabelled:

- `🤖 AFK · ✅ Done`  →  `🤖 AFK · ⚠️ Done (unverified)` + a caveat bullet.

"Corroborating evidence" = at least one **successful** (`isError !== true`) `write_file`, `edit_file`, or `bash` tool call in the turn (`doneHasCorroboratingEvidence` in `afk-push.ts`). Read-only tools (`read_file`/`grep`/`glob`/…) deliberately don't count — reading is not doing.

When **off** (default), behaviour is byte-for-byte unchanged.

## Design notes

- **No trace-file plumbing.** Evidence is read from the turn's in-memory `toolEvents` array already present in `turn-handler.ts` — the trace writer is *not* reachable at the Telegram send point (confirmed while scoping), so this avoids that whole seam.
- **Allowlist safety preserved.** `formatTerminalStateForTelegram` still builds the message only from the parsed `TerminalState` struct; the change passes a single `{ unverified }` boolean, never raw tool output.
- **Conservative + reversible.** Pure label change, fire-and-forget, never blocks the turn. Default off so it can be measured before anyone relies on it ("pressure first, measure, then enforce").

## Scope

This is the small slice. The broader **main-session post-turn hook** from #237 (a `Stop`/`TurnEnd` event carrying `injectContext` to bounce a self-certified turn back) is intentionally left as a follow-up.

## Verification

- `pnpm test src/cli/commands/interactive/afk-push.test.ts`: **19/19 pass** (+14 new: evidence helper, downgrade formatting, push forwarding)
- `pnpm test src/cli/config.test.ts`: 44/44 pass
- Full suite: only 2 pre-existing failures, both unrelated to this change (`anthropic-direct.test.ts` compaction model; `memory-store.test.ts` schema-version — an in-flight memory change on `main`).

Closes the corollary half of #237.

---
Credit: proposed by @amatayomosley-web's agent **Cairn** in #237 — the citations in that issue all checked out against source.

### DIFF
--- a/src/cli/commands/interactive/afk-push.test.ts
+++ b/src/cli/commands/interactive/afk-push.test.ts
@@ -2,14 +2,22 @@ import { describe, it, expect, beforeEach, vi } from 'vitest';
 import {
   formatTerminalStateForTelegram,
   pushTerminalStateToTelegram,
+  doneHasCorroboratingEvidence,
   resetAfkPushBudget,
   afkPushCount,
   MAX_PUSHES_PER_SESSION,
 } from './afk-push.js';
 import type { TerminalState } from './terminal-state.js';
+import type { ToolEvent } from '../../slash/types.js';
 
 function done(extra: Partial<TerminalState> = {}): TerminalState {
   return { kind: 'done', rawBody: '', ...extra };
+}
+
+let toolSeq = 0;
+function tool(toolName: string, isError?: boolean): ToolEvent {
+  toolSeq += 1;
+  return { toolName, toolUseId: `tu-${toolSeq}`, input: '', ...(isError !== undefined && { isError }) };
 }
 
 describe('formatTerminalStateForTelegram', () => {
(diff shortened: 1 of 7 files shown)


griffinwork40:
Verified — `loadTelegramConfig()` → `loadJsonConfig()` reads memoized JSON (config.ts:859, 865), so the per-Done-turn config read is negligible. No perf concern. I now have a fully cross-checked review.

---

# Code Review — PR #238: opt-in "Done" verification for AFK Telegram pushes

**Scope:** +150/−6 across 4 files. Adds `telegram.verifyDone` (opt-in, default off). When on, an autonomous-mode `Done` Telegram push is relabelled **"⚠️ Done (unverified)"** with a caveat line *unless* the turn produced a successful `write_file`/`edit_file`/`bash`. Read-only review — no files changed, nothing posted to the PR.

## Verdict: ✅ Approve with one finding to address-or-accept + minor nits

Correct, safe, additive, backward-compatible, and unusually well-documented. The one finding worth a decision before merge is a false-positive class in the evidence heuristic; it's softened by being opt-in + non-blocking, so it's "fix or consciously document," not a hard blocker.

---

## Findings (severity-ordered)

### 🟡 Medium — evidence heuristic false-positives on delegated/orchestration work *(correctness/completeness)*
`DONE_EVIDENCE_TOOLS = {write_file, edit_file, bash}` only matches the **parent session's own** tool calls. I confirmed by direct read that subagent-internal writes never reach the parent's `toolEvents`:
- `stream-consumer.ts:322–332` emits the `tool_use_detail` chunk (the thing `toolEvents.push` consumes at `turn-handler.ts:354–358`) **only** from the consumed session's own `tool.use.start`.
- Child-session events route to the renderer sink, not the parent's `for await` stream (corroborated by `turn-handler.ts:514` "mid-turn subagent events to the new renderer", + subagent reviewer's `subagent.ts:473–474`/`handle.ts:248–251`).

**Consequence:** a `Done` turn whose real work was delegated — `agent`/`compose`/`skill` doing the writing, which AFK's own prompt actively encourages and which `/mint`, `/refactor`, `/ship` rely on — records only `['agent']`/`['skill']`/`['compose']` in `toolEvents`, none of which are in the evidence set, so it's flagged **"⚠️ Done (unverified)" despite producing durable artifacts.** For AFK's delegation-heavy autonomous runs this false-positive class could be large enough to make the signal noisy.

**Recommendation (author's call):** either
- (a) add `agent`, `compose`, `skill` to `DONE_EVIDENCE_TOOLS` — a delegation call with `isError !== true` is reasonable proxy corroboration that work was attempted this turn (one-line change, matches reality); or
- (b) explicitly document this known limitation in the `Contract:` comment so the "unverified-on-orchestration" behavior is intentional, not a surprise.

The existing comment says "extend this set rather than loosening" — option (a) is exactly that.

### 🟢 Low / should-add — no test for the integration wiring or config parsing *(test-coverage)*
- The turn-handler gate expression (`verdict.kind === 'done' && loadTelegramConfig().verifyDone === true && !doneHasCorroboratingEvidence(toolEvents)`) is untested — existing `turn-handler.test.ts` hardcodes `permissionMode: 'default'`, so the autonomous push branch never runs.
- The `verifyDone` parse branch (`config.ts:716–718` in-diff) has no test; `config.test.ts` tests no telegram keys at all (pre-existing gap — `notify.*` is also untested, so not *introduced* here).
- The pure functions in `afk-push.ts` **are** well covered (the +75 test lines are thorough — success/failure/empty/mixed, the downgrade, and the done-only constraint).

**Recommendation:** add a cheap config round-trip test (`verifyDone: true` set; non-boolean `"true"` silently ignored). The turn-handler integration test is harder given the harness limitation — should-add, not blocking.

### 🟢 Low / nit — `telegram.verifyDone` not registered as a settable key *(api-compat/consistency)*
`settable-keys.ts:208–210` registers `telegram.notify.*` (human-tier), but `verifyDone` is not added — so `afk config set telegram.verifyDone true` and the `config_set` agent tool can't toggle it; it's JSON-edit-only. Unlike `notify.*` (a "notification-redirect vector"), `verifyDone` is a benign boolean label toggle that could safely be an agent/user-tier settable key. Not blocking (the feature works when hand-set in JSON, mirroring how `notify` is primarily JSON-driven), but the new key is invisible to the config-mutation surface every other key participates in.

### Docs — no action required
No user-facing doc enumerates `telegram.*` keys (`notify.*` is only in JSDoc + env cross-refs per `docs/env-registry.md:69–70`), so omitting `verifyDone` from prose is consistent with the baseline. The PR adds good JSDoc on the field itself.

---

## Verified correct (no action)
- **Wiring/timing:** `toolEvents` (decl `:90`, pushed `:358`, consumed by `recordTurn` `:625`) holds *this* turn's events at the new read site `:~688`; the only reset (`:527`) is in the pause/resume replay path. ✓
- **Types:** `ToolEvent` (`slash/types.ts:30–38`) matches the test helper and `doneHasCorroboratingEvidence` usage. ✓
- **Logic:** `some(e => DONE_EVIDENCE_TOOLS.has(e.toolName) && e.isError !== true)` — treats omitted/`false` isError as success, `true` as failure; matches all 6 unit tests. ✓
- **Config plumbing:** `verifyDone` added to both `CliConfig` + `ConfigFileSchema`; parser correctly scoped inside `if (json.telegram)` and defensively typed; `undefined === true` → false default-off. ✓
- **API compat:** both signature changes are optional-with-default; `pushTerminalStateToTelegram(verdict, undefined, { unverified })` correctly passes the default `pushImpl`. ✓
- **Short-circuit + perf:** `loadTelegramConfig()` only evaluated on `done` verdicts; `doneHasCorroboratingEvidence` only when the flag is on; `loadJsonConfig()` is memoized (`config.ts:859,865`) — negligible cost. ✓
- **Downgrade scope:** only `kind === 'done'` is relabelled; blocked/asking untouched (tested). Annotates rather than hides — structured fields + caveat both render. Secret scrubbing path preserved (static caveat text + existing secondary scrub). ✓
- **Safety:** fire-and-forget, `try/catch`, never blocks the turn. ✓

---

## Done
- **What:** Completed a 5-dimension review of PR #238 (correctness, api-compat, test-coverage, perf, security) — diff read, core claims independently re-derived against source, two parallel subagent reviewers dispatched and their headline finding personally verified at `stream-consumer.ts:322–332`.
- **Evidence:** Diff at `/Users/griffinlong/.afk/state/skill-preflight/bbae2745-2222-4eea-8af9-799045306229/pr-238.diff`; findings cite `turn-handler.ts:90/354/527/625/679`, `afk-push.ts` (diff), `config.ts:690–708/716–718/858`, `stream-consumer.ts:322–332`, `settable-keys.ts:208–210`, `slash/types.ts:30–38`.
- **World change:** None — review is read-only. No files edited, no commit, no PR comment posted; working tree (3 unrelated untracked files) untouched as instructed.
- **Recommendation:** Approve. Before merge, decide on the 🟡 Medium (add `agent`/`compose`/`skill` to `DONE_EVIDENCE_TOOLS`, or document the orchestration-only limitation); the test/settable-key items are should-add, not blocking.

---

PR #241, opened 2026-06-21
fix(daemon): quarantine malformed pull-queue entries instead of silent deadlock

sorcerai:
## Summary

A malformed queue file (truncated write, corrupt JSON, or a stray non-JSON file dropped into the queue dir) threw out of `dequeueNext()` **before** its `unlinkSync` ran. `pullTick()`'s empty `catch {}` then swallowed the error, leaving the poison file at the FIFO head — where it blocked **every subsequent queued task on every poll tick, silently**, with no log line and no telemetry record. One corrupt file deadlocked the whole pull-queue.

This was surfaced during a full read-through of the daemon subsystem (recon-grade finding I verified against the source before opening this).

Fix in two parts:

1. **`queue-store.dequeueNext()`** now iterates the sorted entries and **quarantines** any that fail to read or parse into `<queueDir>/poison/` (atomic same-dir rename, lazy `mkdir`, collision-suffixed on retry), then continues to the next entry. The poison entry is preserved for diagnosis instead of being dropped, and the queue keeps draining. `quarantinePoisonEntry()` never throws — if the entry cannot be moved aside it is best-effort unlinked so the queue still unblocks.

2. **`scheduler.pullTick()`** replaces the silent empty `catch` with a `console.error` log so any error that still escapes (e.g. synthetic-task construction) is at least visible in daemon logs instead of vanishing. Errors thrown inside `runOnce` continue to be written to telemetry as before.

The existing `ORDERING INVARIANT` (file removed before returning a task) is preserved — quarantine happens before any valid entry is returned.

## How was this tested?

```
pnpm lint   # tsc --noEmit (strict) — clean
pnpm test   # 9838 passed / 0 failed / 12 skipped (529 files)
pnpm audit:sdk:check   # OK — no SDK import changes
pnpm audit:env:check   # OK — no process.env changes
pnpm build             # OK
```

6 new tests added:

- `queue-store.test.ts` (4): quarantine-at-head + next-valid-returned; all-malformed → null + all quarantined; malformed bytes preserved in `poison/`; `poison/` subdir itself not scanned as a queue entry.
- `scheduler-pull.test.ts` (2): poison at FIFO head + valid task behind it processes on the same tick; multiple poison entries drain to `poison/` while the valid task still runs.

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes

### DIFF
--- a/src/agent/daemon/queue-store.test.ts
+++ b/src/agent/daemon/queue-store.test.ts
@@ -157,6 +157,25 @@ describe('dequeueNext — malformed (poison) entries', () => {
     expect(task).not.toBeNull();
     expect(task!.command).toBe('valid');
   });
+
+  it('quarantines an unreadable entry (read error, not parse error) and keeps draining', () => {
+    // A directory whose name ends in `.json` passes the FIFO filter but makes
+    // readFileSync throw EISDIR — exercising the read-error branch, which is
+    // distinct from the JSON.parse branch every other poison test covers.
+    mkdirSync(join(tmpDir, '0000-q-unreadable.json'), { recursive: true });
+    enqueue('valid-after-unreadable', {}, tmpDir);
+
+    const task = dequeueNext(tmpDir);
+    expect(task).not.toBeNull();
+    expect(task!.command).toBe('valid-after-unreadable');
+
+    // The unreadable entry was moved aside into poison/, not left at the head.
+    expect(readdirSync(join(tmpDir, 'poison'))).toHaveLength(1);
+    const remaining = readdirSync(tmpDir).filter(
+      (f) => f.endsWith('.json') && !f.startsWith('.tmp-'),
+    );
+    expect(remaining).toHaveLength(0);
+  });
 });
 
 describe('listPending', () => {
@@ -180,5 +199,21 @@ describe('listPending', () => {
     enqueue('keep', {}, tmpDir);
     listPending(tmpDir);
     expect(readdirSync(tmpDir)).toHaveLength(1);
+  });
+
+  it('skips malformed entries instead of throwing', () => {
+    enqueue('alpha', {}, tmpDir);
+    writeFileSync(join(tmpDir, '0002-q-poison.json'), '{ not json');
+    enqueue('gamma', {}, tmpDir);
+
+    // Must not throw on the poison entry, and returns only the parseable tasks.
+    const tasks = listPending(tmpDir);
+    expect(tasks.map((t) => t.command)).toEqual(['alpha', 'gamma']);
+
+    // Read-only: nothing moved or removed — the poison entry is left in place
+    // for the dequeue path to quarantine, and no poison/ dir is created here.
+    const remaining = readdirSync(tmpDir).filter((f) => f.endsWith('.json'));
+    expect(remaining).toHaveLength(3);
+    expect(readdirSync(tmpDir)).not.toContain('poison');
   });
 });
--- a/src/agent/daemon/queue-store.test.ts
+++ b/src/agent/daemon/queue-store.test.ts
@@ -4,7 +4,7 @@
  */
 
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
-import { mkdtempSync, rmSync, readdirSync } from 'node:fs';
+import { mkdtempSync, rmSync, readdirSync, writeFileSync, readFileSync, mkdirSync } from 'node:fs';
 import { join } from 'node:path';
 import { tmpdir } from 'node:os';
 import { enqueue, dequeueNext, listPending } from './queue-store.js';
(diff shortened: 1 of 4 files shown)


griffinwork40:
Verification is decisive and changes two severities:

- **`listPending`**: the only non-test reference is its own definition + a `TODO(#337-list)` comment in `src/cli/commands/queue.ts:12`. It has **no live production caller** today → the "throws on poison" issue is *latent*, not a live crash. Downgrade High→Medium.
- **`redactInlineSecrets`**: confirmed the scheduler already redacts error-derived strings at `scheduler.ts:342` and `:450` using the *exact* `err instanceof Error ? err.message : String(err)` pattern the new logs use unredacted. This substantiates the leak finding and gives a one-line idiomatic fix → keep High.

I also independently confirmed the "deadlock if move+unlink both fail" claim is **overstated**: `quarantinePoisonEntry` never throws, so the loop still `continue`s to valid tasks — the queue keeps draining; only the unmovable file persists and re-logs. That's a log-storm (Medium), not a deadlock.

---

# Code Review — PR #241: quarantine malformed pull-queue entries

**`fix/daemon-pull-queue-poison-deadlock` → `main` · +205/−23 · 4 files**

The core fix is correct and well-targeted: a single corrupt queue file previously threw inside `dequeueNext` *before* the unlink, and `pullTick` swallowed it silently — deadlocking every subsequent task with no log. The PR makes `dequeueNext` loop, quarantine bad entries into `poison/`, and keep draining. Verdict below is **Approve-with-one-required-change**; the logic is sound and safe to merge once the secret-redaction gap is closed.

## 🔴 High (fix before merge)

**1. Unredacted error text leaks secrets into daemon logs** — `queue-store.ts` quarantine log (diff L175, L188-191, L195-199) and `scheduler.ts` pullTick (diff L295-297).
Both new sites do `const reason = err instanceof Error ? err.message : String(err)` then `console.error(...)` raw. V8's `JSON.parse` `SyntaxError` embeds a snippet of the malformed input, and a queue `command` can carry an inline secret — so a corrupt entry's bytes can land verbatim in stderr. The codebase already guards this exact pattern: `scheduler.ts:342` and `:450` wrap `err instanceof Error ? err.message : String(err)` in `redactInlineSecrets` before persisting. The new logs bypass that established control.
*Fix:* wrap `reason`/`msg` in `redactInlineSecrets(...)` (import already exists in scheduler.ts; add to queue-store.ts). One line each.

## 🟡 Medium

**2. `listPending` left unhardened** — `queue-store.ts:145-148` (unchanged by PR).
Still bare `readFileSync`+`JSON.parse` in a `.map()`; a poison entry throws. The PR hardens `dequeueNext` but not its sibling, creating an asymmetry. *Verified latent:* no live caller today (only `TODO(#337-list)` in `src/cli/commands/queue.ts:12`), so not a current crash — but it will throw the moment `afk queue list` ships. Harden it in this PR for consistency, or file a tracked follow-up referencing #337.

**3. Stuck-poison re-log storm (no backoff/dedup)** — `quarantinePoisonEntry` last-resort path (diff L200-204).
If both `renameSync` attempts *and* the fallback `unlinkSync` throw (e.g. queue-dir permission loss), the inner `catch {}` is silent and the file persists. The queue is **not** deadlocked (the loop still reaches valid tasks behind it), but every tick re-reads, re-fails, and re-logs the move-failure forever with no dedup or give-up sentinel. Consider logging the final unlink failure (currently swallowed) and/or a per-file attempt cap.

**4. No telemetry record for quarantines** — observability gap.
Quarantine emits stderr only; `runOnce` writes structured JSONL telemetry for every error (`scheduler.ts:342`). On a headless daemon, telemetry is the durable audit log — stderr is ephemeral. The PR's own JSDoc cites "no log or telemetry" as the old failure; it adds the log but not the telemetry. A `status:'poison'` record (taskId=filename, redacted reason) would reach parity. Enhancement, not a blocker.

## 🟢 Low / Nit

**5. Unbounded `poison/` growth** — nothing ever prunes it (confirmed: no GC anywhere). Low operational risk; add a JSDoc note on expected manual cleanup or a future sweep.

**6. Collision-retry comment conflates failure modes** — diff L182-187. The inner catch retries on *any* rename failure, not just `EEXIST`. The comment "Collision or transient failure" is imprecise; EXDEV is practically impossible here since `poison/` is a subdir of `queueDir` (same filesystem). Comment-accuracy nit.

**7. Misleading test title** — `scheduler-pull.test.ts` (diff L254): "keeps draining **across ticks**" but the test advances one 30s interval and its own comment says "on the single tick." Rename to reflect single-tick draining.

## 🧪 Test coverage (solid happy-path; branch gaps)

Covered well: poison-at-head + valid-behind, all-malformed→null + unblock, byte preservation, `poison/` not misread, scheduler skip+run, multiple poison. **Untested branches** (Medium overall): the read-error path (`readFileSync` throws — all fixtures use bad *JSON*, never a read failure); the collision-retry path; the last-resort-unlink path; `listPending` against a poison entry; and `pullTick`'s new catch-log branch.

## ✅ Strengths

- Ordering invariant preserved — unlink still happens **before** `return task` (no double-fire regression). *(verified)*
- `poison/` correctly excluded from both the FIFO scan and `enqueue`'s sequence count (`.endsWith('.json')` filter; `readdirSync` non-recursive). *(verified by 2 reviewers + direct read)*
- `pullTick`'s existing `if (queued === null) return` already covers the expanded null contract — no caller change needed.
- Path-safety sound (filenames from `readdirSync`, no traversal); collision-safe rename; liveness-favoring last-resort unlink; bytes preserved for diagnosis; clear JSDoc contract.

## Merge recommendation

**Request changes (minor).** Land finding #1 (redaction — one line, security-control parity) before merge. Findings #2–#4 are worth addressing in-PR but acceptably tracked as follow-ups. Core behavior is correct and the deadlock is genuinely fixed. After #1, this is an **Approve**.

---

**Done**
- Reviewed PR #241 across correctness, consumer-impact/api-compat, security+observability, and test-coverage via four parallel `compose` reviewers, then independently verified the two pivotal claims (`listPending` caller-liveness; `redactInlineSecrets` convention) and corrected two overstated severities.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/553e17e6-3883-4b3d-bf58-8d6f270b7ebc/pr-241.diff`; base sources `src/agent/daemon/queue-store.ts` and `src/agent/daemon/scheduler.ts`; verification greps confirming `listPending` has no live caller (only `src/cli/commands/queue.ts:12` TODO) and `redactInlineSecrets` usage at `scheduler.ts:342,450`.
- World change: none — review is read-only; no files, git state, or PR were modified (working tree left untouched as instructed).
- Pending: one required change (secret redaction in the two new log sites) and four recommended follow-ups are the author's to action; this review does not implement them.

---

griffinwork40:
@sorcerai — thank you for this! Genuinely one of the cleaner external fixes we've gotten. You traced a real silent-deadlock — a poison file at the FIFO head blocking every subsequent task with no log and no telemetry — down to the exact `dequeueNext`-throws-before-`unlinkSync` ordering bug, and the quarantine-into-`poison/` design (atomic same-dir rename, collision-suffixing, liveness-favoring last-resort unlink, bytes preserved for diagnosis) is exactly the right shape. The ordering invariant is respected and your tests cover the important paths.

Rather than bounce the review back, I pushed a small follow-up commit on top of yours (`bd272f8`) to close the findings. Here's what changed and why:

### Required before merge
- **Secret redaction in the new log sites** (`queue-store` quarantine + `scheduler.pullTick`). A `JSON.parse` `SyntaxError` can embed a snippet of the malformed entry's bytes, and a queue `command` can carry an inline secret — so a raw `err.message` could land verbatim in daemon stderr. The daemon already wraps every error-derived log string in `redactInlineSecrets` (see `scheduler.ts` `runOnce`); the new logs now do the same, so they inherit that existing control instead of bypassing it.

### In-PR consistency & hardening
- **`listPending` hardened to match `dequeueNext`.** Its sibling still had the bare `readFileSync` + `JSON.parse` and would throw on a poison entry (latent today — no live caller yet — but it'd crash the moment `afk queue list` ships). It now skips malformed entries instead of crashing. I kept it deliberately **read-only**: a `list` shouldn't mutate the queue, and quarantining from a read path could race the daemon's dequeue, so the dequeue path stays the single writer.
- **Last-resort unlink failure now logged.** In the move-fails-*and*-unlink-fails branch the final `unlinkSync` error was swallowed; it's logged now (redacted) so a genuinely stuck file is visible instead of silently re-failing each tick. Worth stating plainly: this was never a deadlock — your loop still drains valid entries behind it — just a potential invisible log-storm.
- **Comment & doc accuracy.** Tightened the collision-retry comment (cross-device `EXDEV` can't occur inside a subdir of the queue dir; the retry is really about name collisions) and documented that `poison/` is never auto-pruned.
- **Test title + coverage.** Renamed the "across ticks" scheduler test to reflect that it drains on a *single* tick (which is what it actually asserts), and added two tests: the read-error (`EISDIR`) quarantine branch and `listPending` skipping a malformed entry.

### Deferred as follow-ups (non-blocking; flagged inline as `NOTE(#337-*)`)
- **Quarantine telemetry parity** — a `status:'poison'` JSONL record alongside the stderr log. It's the right move for a headless daemon (telemetry is the durable audit log), but it means threading quarantine events from `queue-store` up to the scheduler's telemetry writer and widening the `TelemetryRecord` status union — broader blast radius than belongs in this PR.
- **Stuck-poison backoff / attempt-cap** — the cheap half (logging the stuck file) is done; a cross-tick dedup/give-up sentinel is a separate enhancement.

Verification on the combined branch: `pnpm lint` clean, full daemon suite green (94 tests incl. the new ones), and `pnpm audit:sdk:check` + `pnpm audit:env:check` both pass.

Thanks again — solid catch and a well-built fix. 🙏

_🤖 follow-up commit + reply via agent-afk_

PR #242, opened 2026-06-21
fix(anthropic-direct): keep session alive after mid-turn interrupt (ESC)

griffinwork40:
## Problem

Hitting **ESC** to stop a running turn in the REPL left the session unable to start new turns: plain prompts were echoed (`▶`) but produced **no agent turn**, while slash commands (`/afk`, `/model`, `/fork`) kept working. Reported as: *"hit esc to stop mid agent-turn and it wouldn't let me send a message to resume."*

## Root cause

`AnthropicDirectQuery`'s outer multi-turn generator (`src/agent/providers/anthropic-direct/query.ts`) returned on **any** aborted per-turn controller:

```ts
} catch (err) {
  if (controller.signal.aborted) return;   // catch-path guard
  ...
}
...
if (controller.signal.aborted) return;     // post-loop guard (added in c462ebd)
```

`interrupt()` (ESC soft-stop) and `close()` **both** abort the per-turn controller via `AbortCoordinator.requestAbort(reason)` and differ only in an **un-inspected reason string**. The guards check only the `signal.aborted` *boolean*, so an ESC interrupt was treated identically to a session close and **permanently terminated the generator** — even though `state.closed` was still `false` and the `while (!this.state.closed)` loop intended to continue.

`AgentSession` reuses **one** shared `providerIterator` across every turn (created at `agent-session.ts:254`, pulled at `:518`). Once the generator returned, `.next()` yielded `{done:true}` forever, so every later `sendMessageStream` produced no events → no turn. Slash commands kept working because they bypass the iterator entirely.

The OpenAI-compatible provider already handles this correctly (`openai-compatible/query.ts:428-435,510-529`; the comment at 607-610 names "the permanent 'esc to interrupt' hang") — this brings anthropic-direct in line.

## Fix

Discriminate **interrupt** from **close** on `this.state.closed` (only `close()` sets it; `interrupt()` never does):

- **close** → `return` (terminate the generator), as before.
- **interrupt** → emit a terminal `turn.completed` **only if the turn produced none** (so the consumer's `providerIterator.next()` loop unblocks instead of hanging — the usage-limit-wait case behind c462ebd), then **`continue`** to await the next prompt, keeping the session alive.

A `turnEmittedTerminal` flag prevents double-emitting a terminal (which would brick the *next* turn). The early-return drain path (a pre-aborted controller from a between-turns abort) is deliberately left as `return` — the REPL cannot trigger a between-turns interrupt (the soft-stop handler is per-turn; `handleSigint` fires only while `turnInFlight`), and the existing pendingAbort / compact-not-blocked guards rely on it.

## Tests

- **New** `interrupt-resume.test.ts` — drives the **real** query generator across a mid-turn interrupt and asserts a second message still runs; plus a guard that a genuine (non-abort) error still terminates with an `error` event. Fails on pre-fix code (turn 2 returns `{done:true}`).
- **Updated** the c462ebd usage-limit interrupt test to the corrected contract: **yields a terminal event + stays resumable**, instead of terminating the generator (which was itself the resume bug).

## Verification

- `pnpm lint` clean; `pnpm audit:sdk:check` clean (no new SDK imports).
- Full suite green except one **pre-existing, unrelated** failure (`anthropic-direct.test.ts` `compact()` — model-alias suffix `claude-haiku-4-5` vs `claude-haiku-4-5-20251001`).
- Diff independently re-derived by an adversarial verifier: close/error termination intact, no abort-slot leak, no double-terminal, auto-compact correctly skipped on interrupt, between-turns interrupt provably unreachable from the REPL.

### DIFF
--- a/src/agent/providers/anthropic-direct/interrupt-resume.test.ts
+++ b/src/agent/providers/anthropic-direct/interrupt-resume.test.ts
@@ -1,0 +1,261 @@
+/**
+ * Regression test for the "ESC mid-turn → can't resume" bug.
+ *
+ * Repro: in the REPL the user hit ESC to stop a running turn. Afterward, plain
+ * prompts were echoed but started NO agent turn, while slash commands still
+ * worked. Root cause: `AnthropicDirectQuery`'s outer multi-turn generator
+ * `return`ed on ANY aborted per-turn controller (`if (controller.signal.aborted)
+ * return`). Because `interrupt()` and `close()` both abort the per-turn
+ * controller and differ only in an un-inspected reason string, an ESC interrupt
+ * was treated like a session close and permanently terminated the generator.
+ * `AgentSession` reuses ONE `providerIterator` across turns, so once the
+ * generator returned every later `sendMessageStream` got `{done:true}` and ran
+ * no turn. Slash commands kept working because they never pull the iterator.
+ *
+ * The fix discriminates close (`state.closed`) from interrupt and, on interrupt,
+ * emits a terminal event (if the turn produced none) then loops back for the
+ * next prompt instead of returning — mirroring the OpenAI-compatible provider.
+ *
+ * These tests drive the REAL `AnthropicDirectQuery[Symbol.asyncIterator]`
+ * generator (where the bug lived) via `provider.query()`, mocking the Anthropic
+ * SDK's `messages.create`. A mock-provider test at the AgentSession layer would
+ * NOT catch the bug.
+ */
+
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import type Anthropic from '@anthropic-ai/sdk';
+import type { RawMessageStreamEvent } from '@anthropic-ai/sdk/resources';
+import type { ProviderEvent } from '../../provider.js';
+import { AnthropicDirectProvider, __setAnthropicClientFactory } from './index.js';
+
+// --- Mock SDK plumbing (mirrors query-auth-retry.test.ts) ---
+
+const messagesCreateMock = vi.fn();
+
+class MockAnthropic {
+  public messages: { create: typeof messagesCreateMock };
+  constructor() {
+    this.messages = { create: messagesCreateMock };
+  }
+}
+
+function installFactory(): void {
+  __setAnthropicClientFactory(() => new MockAnthropic() as unknown as Anthropic);
+}
+
+// --- Helpers ---
+
+async function* fromArray<T>(arr: T[]): AsyncIterable<T> {
+  for (const x of arr) yield x;
+}
+
+/** A single-message prompt stream that completes after one turn. */
+async function* singleInput(content: string): AsyncIterable<{ content: string }> {
+  yield { content };
+}
+
+async function collect(query: AsyncIterable<ProviderEvent>): Promise<ProviderEvent[]> {
+  const out: ProviderEvent[] = [];
+  for await (const ev of query) out.push(ev);
+  return out;
+}
+
+/** A prompt stream we can push user turns to over the life of the test. */
+function createPushStream<T>(): {
+  push: (item: T) => void;
+  close: () => void;
+  iterable: AsyncIterable<T>;
+} {
+  const queue: T[] = [];
+  let waiting: ((r: IteratorResult<T>) => void) | null = null;
+  let closed = false;
+  return {
+    push(item: T): void {
+      if (waiting) {
+        const resolve = waiting;
+        waiting = null;
+        resolve({ value: item, done: false });
+      } else {
+        queue.push(item);
+      }
+    },
+    close(): void {
+      closed = true;
+      if (waiting) {
+        const resolve = waiting;
+        waiting = null;
+        resolve({ value: undefined as unknown as T, done: true });
+      }
+    },
+    iterable: {
+      [Symbol.asyncIterator](): AsyncIterator<T> {
+        return {
+          next(): Promise<IteratorResult<T>> {
+            const head = queue.shift();
+            if (head !== undefined) return Promise.resolve({ value: head, done: false });
+            if (closed) return Promise.resolve({ value: undefined as unknown as T, done: true });
+            return new Promise<IteratorResult<T>>((resolve) => {
+              waiting = resolve;
+            });
+          },
+        };
+      },
+    },
+  };
+}
+
+/** Minimal text-only stream that ends with stop_reason=end_turn. */
+function makeTextStream(text: string): RawMessageStreamEvent[] {
+  return [
+    {
+      type: 'message_start',
+      message: {
+        id: 'msg_test',
+        type: 'message',
+        role: 'assistant',
+        content: [],
+        model: 'claude-sonnet-4-5-20250929',
+        stop_reason: null,
+        stop_sequence: null,
+        usage: {
+          input_tokens: 5,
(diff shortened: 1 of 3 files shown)


griffinwork40:
Reachability confirmed: **every production caller of `interrupt()` is turn-in-flight–gated** — the REPL ESC/SIGINT paths (verified `turnInFlight`/`inputMode`-gated by the correctness agent) and the subagent handle (`if (this.inFlight) await this.session.interrupt()` at `handle.ts:392,417`). No production caller invokes `interrupt()` between turns, so the "between-turns drain path terminates the generator" concern has **zero reachable trigger** → it downgrades to nit (latent gap only).

I now have full grounding. All findings are low/nit (no critical/high → no Wave 1.5 formal gate and no shadow-verify required), and I've independently resolved the only cross-agent tension. Synthesizing inline rather than dispatching a Wave 2 subagent (6 low/nit findings — a synthesis fork would be ceremony).

---

# Code Review — PR #242

**`fix(anthropic-direct): keep session alive after mid-turn interrupt (ESC)`**
Target: `fix/interrupt-mid-turn-resume` → `main` · head `073004c` · MERGEABLE · +448/−71, 3 files (1 production: `query.ts`; 2 tests)
Stated intent: **present** (PR title+body) — spec-compliance assessed. Regime: full (2-agent wave + targeted self-verification).

## Verdict by dimension
| Dimension | Result |
|---|---|
| Security | clean — no credential/secret/RCE surface touched |
| Correctness | **PASS** — close/interrupt discrimination is sound; all 7 critical paths verified |
| Spec-compliance | **PASS** — implements the stated fix exactly; no unmet clause, no scope creep |
| api-compat | one low observable-contract drift; otherwise clean (no new SDK imports) |
| Test-coverage | adequate; two minor gaps |
| Perf/observability | no issues |

## Findings (sorted; all low/nit — none blocking)

**1 · low · api-compat · `agent-session.ts:524-525,358-359` · ref:073004c · file-state**
The synthetic interrupt event (`makeInterruptedTurnEvent`, `isError:false`) flows through `transformProviderEvent` → `{type:'done'}`, which increments `turnCount` (counts an abandoned turn against `maxTurns`) and sets `lastStopReason='interrupted'`. *Self-verified low-impact:* `isTruncationStopReason('interrupted')` is false (`closure-reason.ts:87`), so a later clean close seals as `model_end_turn` (correct — the session ended normally, not via the interrupt); `lastStopReason` is overwritten by the next real turn; no consumer treats `'interrupted'` as an error; and the budget gate is correctly skipped (`stream-consumer.ts:404` requires `typeof totalCostUsd === 'number'`, which the synthetic event omits — docstring claim verified). The OpenAI provider's `finishTurn` shares this counter pattern, so the behavior is structural, not new.
*Suggestion:* if turn-accounting precision matters, skip `turnCount++` when `resultSubtype === 'interrupted'` in `sendMessageStreamInternal`. Optional.

**2 · low · correctness · `query.ts:~408 + ~420` · ref:073004c · file-state**
The interrupt catch path calls `this.abort.clear(controller)` explicitly **and** again via the `finally`. `AbortCoordinator.clear()` is idempotent (compare-and-clear), so no behavior change — the explicit call is redundant.
*Suggestion:* drop the explicit clear in the catch branch; rely on `finally`. Cosmetic.

**3 · low · test-coverage · `interrupt-resume.test.ts` · ref:073004c · file-state**
No test covers **interrupt-during-tool-execution then resume** (grep for `tool`/`function_call` finds only `server_tool_use: null` at line 125). The streaming-interrupt path is covered; the tool-loop abort path's resume contract is not.
*Suggestion:* add a case returning `stop_reason:'tool_use'`, interrupting after the tool dispatch, asserting resumability.

**4 · nit · api-compat/spec-compliance · `query.ts:~315-322` · ref:073004c · file-state**
The between-turns early-return drain path returns unconditionally on a pre-aborted controller, **not** discriminating close from interrupt (unlike the two mid-turn paths the PR fixes). *Self-verified unreachable from current callers:* every production `interrupt()` caller is turn-in-flight–gated (REPL `turnInFlight`/`inputMode`; subagent `if (this.inFlight)` at `handle.ts:392,417`), so no production path triggers a between-turns interrupt. Latent robustness gap only; the PR documents the choice and the correctness agent verified the two REPL gates.
*Suggestion:* if a future non-REPL caller could interrupt between turns, change `return` → `if (!this.state.closed) continue;` and promote the comment to an `// Invariant:` block.

**5 · nit · test-coverage · `interrupt-resume.test.ts:241` · ref:073004c · file-state**
The "non-abort error still terminates" guard only exercises an error on the **first** API call, not on a turn after a successful turn + interrupt. The branch is index-independent, so coverage is meaningful but narrow.

## Verified clean (high-value confirmations)
- `close()` sets `state.closed`; `interrupt()` does **not** — the discrimination the whole fix rests on is correct.
- `turnEmittedTerminal` correctly suppresses a double-terminal when `loop.ts` already yielded `turn.completed`.
- Genuine (non-abort) errors still `yield {type:'error'}; return;` — unchanged and tested.
- Auto-compaction is correctly skipped on interrupt (the `continue` jumps over it).
- Budget gate skipped on the synthetic event (no `totalCostUsd`) — no spurious budget abort.
- New test drives the **real** `AnthropicDirectQuery` generator (would fail pre-fix); the updated c462ebd test asserts the corrected contract (terminal + resumable, no `resumed`, `callIdx===2`).
- `pnpm audit:sdk:check` claim holds — no new `@anthropic-ai/sdk` symbols.

## What was not checked
- Wave 1 agents + I read against branch HEAD **073004c**; all blocking-tier citations would have been verified at that ref (none reached blocking tier).
- **OpenAI-compatible parity claim** (`openai-compatible/query.ts:428-435,510-529`) was not line-by-line re-derived by either agent — it's a rationale, not a correctness gate; the Anthropic fix is verified correct on its own.
- `loop.ts` mid-stream `turn.completed`-on-abort yield and `retry.ts` clean-return path were cited but not opened at the ref; the `turnEmittedTerminal` guard makes the fix correct regardless of which one delivers the terminal.
- No test suite was executed (read-only review); PR reports full suite green except one pre-existing unrelated `compact()` model-alias failure.

---

## Merge decision: **MERGE**

The fix implements the stated intent precisely — discriminate interrupt from close on `state.closed`, emit a terminal event only when the turn produced none, then `continue` — and the seven load-bearing correctness paths are verified at the reviewed ref. All five findings are low/nit with no reachable defect; the headline `turnCount`/`lastStopReason` drift is independently confirmed low-impact, and the between-turns asymmetry has no production trigger. Findings 1–5 are optional follow-ups, not merge blockers.

**Done** — Reviewed PR #242 at ref `073004c` via a 2-agent parallel wave plus targeted self-verification of the downstream `turnCount`/`lastStopReason`/`closure-reason`/budget-gate behavior and `interrupt()` caller reachability.
- Evidence: findings cite `query.ts`, `agent-session.ts:524-525,358-359`, `stream-consumer.ts:404`, `closure-reason.ts:87`, `subagent/handle.ts:392,417`, all read at `git show 073004c:<file>`.
- World unchanged: read-only review — no files edited, nothing committed/pushed, no PR comment or status posted. The working tree's 3 pre-existing dirty files (unrelated plan docs + a `.txt`) were left untouched.
- Decision surfaced above: **MERGE**, with five non-blocking low/nit follow-ups.

---

PR #254, opened 2026-06-23
feat(subagent): wire passive SIGNAL-block parsing into SubagentResult

griffinwork40:
## What

Wires `parseSignal()` from `src/agent/signal-block.ts` into `buildResultFromMessage` (`src/agent/subagent/result.ts`), surfacing a parsed signal as an optional `signal?: Signal` field on `SubagentResult<T>`. Removes the now-satisfied `TODO(signal-block-wiring)` from the `signal-block.ts` module header.

## Why

`parseSignal` shipped ahead of its caller with a TODO deferring the wiring, so it was effectively dead code — defined and fully tested but imported by nothing in production. This closes that deferral at the one natural integration point: the final-message result builder, right beside the existing `extractStructuredOutput` call.

## Passive v0 — non-authoritative by contract

Strictly passive observation, per the `signal-block.ts` v0 contract:

- `parseSignal` is called once per final message, unconditionally (independent of `outputSchema` presence or validity — the parser is key-scoped so it coexists with schema extraction).
- `signal` is purely additive — no existing field, status, control-flow, or schema-validation path changes.
- Consumers **MUST NOT** use `signal` to gate finalization, block tools, or mutate message history. This invariant is documented on the field.

## Explicitly deferred (future increments)

- Telemetry / trace emission of signal values
- Cross-subagent consensus / voting
- Any observer beyond the `SubagentResult` field (hooks, session-level aggregation)

## Verification

- `pnpm lint` (`tsc --noEmit`, strict): **clean**.
- New `src/agent/subagent/result.test.ts` (7 tests): valid signal w/o schema, absent signal, malformed signal (bad stance enum), signal + schema-success (both `output` and `signal` populated), and signal + schema-failure (`status:'failed'`/`schemaError` with `signal` still present). All pass. Existing `signal-block.test.ts` (27 tests) unchanged and passing.
- Full suite: the only failures (`anthropic-direct.test.ts` model-alias assertion; an intermittent `config.test.ts` env-bleed) **reproduce on `main`** without this change and are unrelated. This PR adds 7 passing tests and introduces no new failures.

### DIFF
--- a/src/agent/signal-block.ts
+++ b/src/agent/signal-block.ts
@@ -46,12 +46,6 @@
  *   observer's job; we just gatekeep the shape.
  *
  * @module agent/signal-block
- *
- * TODO(signal-block-wiring): `parseSignal` is defined and tested but not yet
- * imported by any runtime caller (`handle.ts`, `result.ts`, `executor.ts`).
- * Wiring into the subagent result pipeline is intentionally deferred to a
- * follow-up PR — this PR only lands the parser + schema so observers can
- * adopt it incrementally. Remove this TODO when a caller imports it.
  */
 
 import { z } from 'zod';
--- a/src/agent/subagent/result.test.ts
+++ b/src/agent/subagent/result.test.ts
@@ -1,0 +1,118 @@
+/**
+ * Tests for SIGNAL block wiring into buildResultFromMessage.
+ *
+ * Covers the passive-observation contract: `result.signal` is populated from
+ * the subagent's final message when a well-formed SIGNAL block is present,
+ * and is absent otherwise. Signal extraction is independent of outputSchema
+ * presence / validation outcome.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { z } from 'zod';
+import type { Message } from '../types.js';
+import { buildResultFromMessage } from './result.js';
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+const VALID_SIGNAL = {
+  issue: 'cache-race',
+  stance: 'supports' as const,
+  confidence: 0.82,
+  evidence: ['src/cache/lru.ts:142'],
+  claim: 'The eviction path races with concurrent reads under load.',
+};
+
+const VALID_SIGNAL_BLOCK = { signal: VALID_SIGNAL };
+
+function fenced(body: unknown): string {
+  return '```json\n' + JSON.stringify(body, null, 2) + '\n```';
+}
+
+function makeMessage(content: string): Message {
+  return {
+    role: 'assistant',
+    content,
+    timestamp: new Date(),
+    metadata: { usage: { inputTokens: 1, outputTokens: 1 }, stopReason: 'end_turn' },
+  };
+}
+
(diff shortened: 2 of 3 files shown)


griffinwork40 (commented):

**Merge decision: ✅ MERGE**

Small, focused, additive change (140 lines across 3 files, one a new test). Wires the previously-dead `parseSignal()` into `buildResultFromMessage` at the one natural integration point, surfacing an optional `signal?: Signal` on `SubagentResult<T>`. No blocking, critical, high, or medium findings. Citations read against the reviewed ref `18be4528`.

### Verified (load-bearing claims, independently re-derived)
- **Conditional spread is safe.** `...(signal !== undefined && { signal })` on all three return paths spreads to nothing when `signal` is `undefined` (`{ ...false }` is a no-op) — no spurious `signal: false`/`undefined` key is injected. (`src/agent/subagent/result.ts`)
- **TODO actually removed at the ref.** `TODO(signal-block-wiring)` is absent from all production source at `18be4528`; no other file still claims `parseSignal` is unwired.
- **No behavior change on existing fields.** `id`, `status`, `message`, `output`, `trace`, `schemaError`, `error` are byte-for-byte unchanged on all three paths — only the trailing conditional spread is new.

### Spec-compliance
Assessed against the PR description. Intent met: parser wired beside `extractStructuredOutput`, TODO removed, passive-v0 invariant documented on the field. The explicitly-deferred items (telemetry/trace emission, cross-subagent consensus, observers beyond the field) are correctly out of scope — not unmet intent. No scope creep detected.

### Correctness / API-compat
- `message.content` is typed `string` (`src/agent/types/message-types.ts:49`) — `parseSignal(content: string)` receives a valid string at every call site.
- `parseSignal` (key-scoped, first signal-bearing block) and `extractStructuredOutput` (positional, last block) are orthogonal — confirmed on both the cohabitated-block and separate-block test shapes; Zod strip-mode discards the extra `signal` key from a combined block.
- `buildResultFromError` (catch path) carries no signal by design — it has no `message` to parse, consistent with the "final message only" intent.
- `SubagentResult` gains an optional field — additive, one production importer (`src/agent/subagent/handle.ts`), no destructuring or union-narrowing breaks.

### Non-blocking nits
1. **Test fixture brittleness** — `result.test.ts:32–40`: `makeMessage` sets `metadata.usage` as `{ inputTokens, outputTokens }`. The `usage` shape is irrelevant to signal parsing; a bare `metadata: {}` would be less brittle if the `Message` type ever tightens.
2. **Unconditional parse** — `result.ts:117`: `parseSignal` runs a two-pass scan on every final message even when the caller never reads `result.signal`. One invocation per subagent completion (not a tight loop) — negligible at v0 scale; only worth revisiting if very large (>100 KB) final messages become common.

### What was not checked
- Read against branch HEAD `18be4528` — citations verified at that ref.
- Stated intent: PR #254 title + body — spec-compliance assessed.
- Did not run the test suite (read-only review); the PR's own verification reports 7 new passing tests + 27 unchanged signal-block tests, with pre-existing unrelated failures reproducing on `main`.

---
*Read-only automated analysis. No approval, change-request, or merge action was taken — this is a comment only.*

PR #255, opened 2026-06-23
fix(tui): commit flushed blocks atomically so band-hold doesn't gap scrollback

griffinwork40:
## Symptom

A completed tool root / subagent block / card flushed to scrollback rendered with **tall blank, gutter-less gaps** wedged between rows (and, in the worst geometry, **dropped rows**) when a tall overlay filled most of the viewport during the flush — the "weird gaps" in the TUI diff render.

The blank rows are NOT a diff-formatting bug: `colorDiffLine` correctly emits a green `+ ` gutter for blank added lines. They are real content rows that the band-hold painter never drew.

## Root cause

`TerminalCompositor.commitAbove` is designed to place a whole block with **one geometry decision** — a single pre-clear `prevTopRow` capture and a single fits / band-hold / overflow route (see `terminal-compositor.committed-band-commit.ts`, "one geometry per commit", and `commit-mode.ts`).

Several flush sites instead committed the block **one line per call**:

```ts
for (const line of lines) compositor.commitAbove(line);
compositor.commitAbove('');
```

Under a tall overlay the per-line band-hold/fits routing desyncs the committed-band model from the screen: `maxRun` caps each tiny commit, some rows stay **pending** (in `committedBand`, never painted — `committedBandPaintedRows < committedBand.length`), and the next per-line commit scrolls those unpainted blank rows into scrollback. Line order is preserved, but blank rows are interleaved (or a row beyond `maxBandModel` is dropped).

This was reproduced against the real `@xterm/headless` scroll engine: at a partial-room overlay geometry, the per-line loop yields `span=29` for a 14-row block (16 blank rows wedged in), while a single batched commit yields `span=13` (contiguous).

## Fix

Add `commitBlockAbove(compositor, lines)` — a single `commitAbove` with the block joined by `\n` — and route every coherent-block flush through it:

- `flushToolLaneToScrollback` (the reported path) + the finalize-coordinator and safety-net flushes
- `emitPanel` card + subagent panel cards
- the input-echo card

`commitAbove` already splits on `\n` and hard-wraps each row, so band-hold row math stays exact and a block taller than the screen archives its overflow to scrollback with the tail hugging the frame (validated — no extra drops, refuting the "batching drops more content" concern for tall blocks).

Also corrects a stale `// commitAbove expects one logical line per call` comment in `input-surface.ts` — multi-line blocks are supported, and `markdown-stream.ts` already relies on it.

## Validation (real PTY engine)

Mock-stdout byte tests cannot observe true scrollback (see `docs/scrollback.md`), so validation drives the actual `@xterm/headless` scroll engine and asserts on its buffer.

- **New** `terminal-compositor.band-hold-perline-gap.repro.test.ts`: per-line loop is genuinely gappy at the partial-room geometry (discriminator); `commitBlockAbove` is present + single-copy + **contiguous** across overlay heights and for a block taller than the collapsed screen.
- All **975 tests across the 58 CLI rendering files** pass (band / gap / subagent / rhythm / ordering).
- `stream-renderer-ordering.test.ts` assertions updated to compare **rendered char offset** instead of `commitAbove` call index (a flushed block is now one atomic commit), preserving the eager-ancestor-header-before-children contract.
- `tsc --noEmit` clean for all changed files.

## Notes for the reviewer

- The two-block `mid≠null` drop case explored during investigation (a second flush with only one intervening repaint) is a **separate** root cause (the `overflowHasPending` pending-count proxy in `commit-mode.ts`) and is not reachable in production — the ~80ms spinner-render cadence drains pending rows between flushes. Left as a documented follow-up rather than scope-creeping this fix.

### DIFF
--- a/src/cli/_lib/commit-block.ts
+++ b/src/cli/_lib/commit-block.ts
@@ -1,0 +1,39 @@
+/**
+ * commitBlockAbove — commit a coherent multi-line block to scrollback as a
+ * SINGLE `commitAbove` call.
+ *
+ * @module cli/_lib/commit-block
+ */
+
+/** Minimal structural slice of TerminalCompositor this helper needs. */
+export interface BlockCommitter {
+  commitAbove(text: string): void;
+}
+
+// Invariant (one geometry per commit): TerminalCompositor.commitAbove places a
+// whole block with ONE geometry decision — a single pre-clear `prevTopRow`
+// capture and a single fits / band-hold / overflow route (see
+// terminal-compositor.committed-band-commit.ts, "one geometry per commit", and
+// terminal-compositor.commit-mode.ts). Committing a block one line at a time
+// forces N independent geometry decisions: under a TALL overlay the per-line
+// band-hold/fits routing desyncs the committed-band model from the screen and
+// scrolls unpainted (blank) rows into scrollback — the "weird gaps" (and, in
+// the worst geometry, dropped-line) rendering bug. Joining the block into one
+// call restores the atomic contract: commitAbove re-splits on '\n' and
+// hard-wraps each row, so a row wider than the terminal still maps to exactly
+// one logical line, and the band-hold row math stays exact.
+//
+// `lines` are the physical rows of ONE rendered artifact (a card, a flushed
+// tool-lane block, an input echo) — NOT independent scrollback entries — so
+// committing them as a single block is always the correct semantics.
+//
+// Empty input is a no-op, matching the prior `for (const line of lines)` loops
+// (each was length-guarded, or iterated an always-non-empty card split).
+//
+// TTY-only: the band-hold desync this prevents is a property of the live
+// compositor frame. Non-TTY callers write each line to a plain writer
+// (`out.line`) and are unaffected — leave those loops as-is.
+export function commitBlockAbove(compositor: BlockCommitter, lines: readonly string[]): void {
+  if (lines.length === 0) return;
+  compositor.commitAbove(lines.join('\n'));
+}
--- a/src/cli/_lib/stream-renderer-orchestrator-emit.ts
+++ b/src/cli/_lib/stream-renderer-orchestrator-emit.ts
@@ -11,6 +11,7 @@ import type { CardSpec } from '../render.js';
 import { card, errorBox } from '../render.js';
 import { renderMarkdownToTerminal } from '../formatter.js';
 import { formatThoughtSummary } from '../commands/interactive/thinking-lane.js';
+import { commitBlockAbove } from './commit-block.js';
 
 import type { OrchestratorCtx } from './stream-renderer-orchestrator.js';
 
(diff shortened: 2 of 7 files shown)


griffinwork40 (commented):

**Merge decision: MERGE** ✅

Reviewed at ref `10e25ed`. Regime: light (hotfix, 7 files, ~326 lines). Dimensions covered: security, api-compat, correctness, spec-compliance, test-coverage, perf/observability. No blocking, critical, or high findings.

### Spec-compliance — fully met
All six flush sites named in the PR intent are routed through the new atomic `commitBlockAbove`:
- `emitPanel` (`stream-renderer-orchestrator-emit.ts:45`)
- finalize-coordinator flush (`stream-renderer-orchestrator-emit.ts:213`)
- `flushToolLaneToScrollback` (`stream-renderer-orchestrator-emit.ts:335`)
- `emitSubagentPanel` (`stream-renderer-subagent.ts:408`)
- after-subagent batch (`stream-renderer.ts:633`)
- safety-net / dispose flush (`stream-renderer.ts:763`)
- input-echo card (`input-surface.ts:485`)

The stale `// commitAbove expects one logical line per call` comment in `input-surface.ts` is corrected. The documented out-of-scope follow-up (the two-block `overflowHasPending` drop in `commit-mode.ts`) is correctly left unaddressed — no scope creep.

### Correctness — verified clean
- `commitBlockAbove` joins lines with `\n` and makes a single `commitAbove` call (`commit-block.ts:37-38`); `commitAbove` re-splits on `\n` and hard-wraps each logical line (`terminal-compositor.committed-band-commit.ts:130,141`), so the join → single-commit round-trip is sound and preserves band-hold row math.
- Empty-array guard (`commit-block.ts:37`) matches the prior zero-iteration loop behavior.
- Trailing rhythm separator `commitAbove('')` is correctly kept separate from the block at every call site.
- Single-line commits (thinking phase, badges) and non-TTY paths are correctly left unchanged.

### Test coverage — adequate
New `terminal-compositor.band-hold-perline-gap.repro.test.ts` drives the real `@xterm/headless` scroll engine across three overlay heights plus a taller-than-screen block, with a discriminator test proving the per-line loop is genuinely gappy at the repro geometry. The `stream-renderer-ordering.test.ts` change to compare rendered char offset (instead of `commitAbove` call index) is semantically valid now that a flushed block is one atomic commit.

### Non-blocking findings
- **nit** · `terminal-compositor.band-hold-perline-gap.repro.test.ts:63` — `SPINNER_RE` is declared but never used; remove it or add a guard assertion.
- **nit** · `tool-lane.ts:705-706` (pre-existing, not in this diff) — the `flushSource` return-value comment still says the caller iterates with `commitAbove(line)` per element; it now commits the array atomically via `commitBlockAbove`.
- **low** · `commit-block.ts:37` — no direct unit test asserts `commitBlockAbove(compositor, [])` is a no-op; the branch is exercised only indirectly via the outer `if (lines.length > 0)` guard.

### What was not checked
Citations read against branch HEAD `10e25ed` (verified at that ref). Stated intent: PR #255 title + body — spec-compliance assessed. Did not execute the test suite locally (PR reports all 975 CLI rendering tests pass + `tsc --noEmit` clean); did not exercise the live TUI at the partial-room overlay geometry. No production importers exist for the new `BlockCommitter`/`commitBlockAbove` symbols outside `src/cli/`.

PR #257, opened 2026-06-23
fix(tui): use exact painted-row count for the band-hold pending signal (#255 follow-up)

griffinwork40:
> **Stacked on #255** (base branch `fix/tui-band-hold-pending-gap`, not `main`). This PR is the follow-up that #255's reviewer note explicitly deferred. Review/merge after #255, or rebase onto `main` if #255 lands first.

## What this resolves

#255's "Notes for the reviewer" deferred a **separate** root cause:

> The two-block `mid != null` drop case ... is a separate root cause (the `overflowHasPending` pending-count proxy in `commit-mode.ts`) and is not reachable in production — the ~80ms spinner-render cadence drains pending rows between flushes. Left as a documented follow-up rather than scope-creeping this fix.

This PR fixes that proxy.

## Root cause

`decideCommitMode` routes a `commitAbove` block to the band-hold path when the prior committed band still carries **pending** rows — rows held in the in-memory band model but never painted to the terminal and never archived to scrollback. It detected that condition with a **geometry proxy**:

```ts
const overflowHasPending = overflowPriorContiguous && committedBand.length > room;
```

`room` (`frameTop - anchorFloor`) is the above-frame space *right now*. The proxy assumes "band taller than current room ⟹ some rows are pending." That breaks in the two-block sequence with exactly **one** intervening repaint: the repaint grows `frameTop` (the overlay shrank a little) so `room` now exceeds `committedBand.length`, yet the earlier paint left rows unpainted (`committedBandPaintedRows < committedBand.length`). The proxy reads `length > room == false` → no pending → routes the next commit to the **fits path**, which scrolls those still-unpainted rows into scrollback as **blanks** (and Phase 3's cap drops the real rows). That is the deferred drop.

In production the ~80ms spinner-render cadence fires several repaints between flushes, so `repositionCommittedBand` paints the pending rows (driving `committedBandPaintedRows` up to `committedBand.length`) before the second flush — which is why it was "not reachable" and safely deferred. This PR makes the signal correct regardless of cadence.

## Fix

Use the exact painted-row count the compositor **already tracks** instead of the geometry proxy:

```ts
const overflowHasPending =
  overflowPriorContiguous && committedBand.length > committedBandPaintedRows;
```

The pending prefix is exactly `committedBand.slice(0, length - committedBandPaintedRows)` (see the invariant on `TerminalCompositor.committedBandPaintedRows`, `terminal-compositor.ts:323-339`). `committedBandPaintedRows` is maintained at every paint site (commit Phase 3, `repositionCommittedBand`, `preserveRowsBeforeFrameRender`) and reset to 0 by `clearCommittedBand()` — so this adds **no new state**, only threads the existing field through the single `decideCommitMode` call site.

- `commit-mode.ts` — new non-optional `CommitModeInput.committedBandPaintedRows`; exact signal; doc comments updated (the `room` proxy is retired but `room` itself stays, still consumed by the fits-path math).
- `terminal-compositor.committed-band-commit.ts` — pass `self.committedBandPaintedRows` (the `CommittedBandHost` interface already declared the field).
- `commit-mode.test.ts` — `base()` gains `committedBandPaintedRows: 0`; a new divergence regression.

**Behavioral delta:** the change only flips `overflowHasPending` `false → true` in the divergence case (pending rows exist but `room >= band length`) — strictly the bug-fix direction. Everywhere else it is a no-op (when the band is fully painted there are no pending rows, and `committedBand.length > committedBandPaintedRows` matches the old proxy whenever the proxy was right).

## Validation

- **New** `commit-mode.test.ts` case `overflowHasPending uses the exact painted-row count, not the room geometry proxy (#255 two-block follow-up)`: at the divergence geometry (`band=5`, `room=7`, `painted=3`) it asserts `band.length > room === false` (the old proxy would say "no pending") while `overflowHasPending === true` and `useBandHold === true` (the exact signal routes to band-hold). It fails on the proxy, passes on the fix.
- `commit-mode.test.ts` — **14/14** (13 prior + the new case).
- All **23 `terminal-compositor*` suites — 306 tests green** (overflow-gap, scrollback-gap, banner-commit-gap, h1-prevtoprow, disarm-pending-band, band-hold-perline-gap.repro, shrink-gap, committed-band-single-copy, wrap-overlap, …).
- `tsc --noEmit` clean (the non-optional field forces every call site to be wired).

Validation is at the **pure `decideCommitMode` unit** deliberately: the deferred case is unreachable through the `@xterm/headless` harness (it needs *exactly* one intervening repaint, which the harness/cadence does not produce), so the unit that owns the routing arithmetic is the faithful and discriminating level for it.

### DIFF
--- a/src/cli/commit-mode.test.ts
+++ b/src/cli/commit-mode.test.ts
@@ -19,6 +19,7 @@ function base(overrides: Partial<CommitModeInput> = {}): CommitModeInput {
     extraRows: 2,
     committedBand: [],
     committedBandBottomRow: 0,
+    committedBandPaintedRows: 0,
     ...overrides,
   };
 }
@@ -89,6 +90,34 @@ describe('decideCommitMode', () => {
     expect(m.overflowRun).toHaveLength(22);
     expect(m.overflowRun.length > m.maxBandModel).toBe(true);
     expect(m.useBandHold).toBe(true); // without the override this would be false
+  });
+
+  it('overflowHasPending uses the exact painted-row count, not the room geometry proxy (#255 two-block follow-up)', () => {
+    // Deferred #255 follow-up. Under a single intervening repaint the frame top
+    // can grow so `room` (frameTop - anchorFloor = 7) exceeds the band length
+    // (5) WHILE pending rows remain (only 3 of 5 painted). The OLD proxy
+    // `committedBand.length > room` reads 5 > 7 = FALSE → routes to the fits
+    // path, which scrolls the 2 unpainted rows into scrollback as BLANKS (the
+    // drop). The EXACT signal `committedBand.length > committedBandPaintedRows`
+    // reads 5 > 3 = TRUE → stays on band-hold and archives REAL content.
+    const band = ['b0', 'b1', 'b2', 'b3', 'b4'];
+    const m = decideCommitMode(
+      base({
+        prevTopRow: 8,
+        frameTop: 8, // room = 8 - 1 = 7
+        lineCount: 2,
+        textLines: ['new-a', 'new-b'],
+        committedBand: band,
+        committedBandBottomRow: 7, // === frameTop - 1 → contiguous
+        committedBandPaintedRows: 3, // 2 of 5 rows still pending
+      }),
+    );
+    // Prove this geometry is exactly where the proxy and the exact signal DIVERGE.
+    expect(m.room).toBe(7);
+    expect(band.length > m.room).toBe(false); // the old proxy → "no pending" (wrong)
+    expect(m.overflowPriorContiguous).toBe(true);
+    expect(m.overflowHasPending).toBe(true); // the exact signal → pending (right)
+    expect(m.useBandHold).toBe(true); // would route to the fits path under the proxy
   });
 
   it('merges the prior band into the run only when verifiably contiguous', () => {
--- a/src/cli/commit-mode.ts
+++ b/src/cli/commit-mode.ts
@@ -32,6 +32,14 @@ export interface CommitModeInput {
   committedBand: string[];
   /** Tracked bottom row of the committed band (0 when unset). */
   committedBandBottomRow: number;
+  /**
+   * How many of `committedBand`'s rows are MATERIALIZED on the terminal now
+   * (always the bottom suffix nearest the frame). The complementary prefix
+   * `committedBand.slice(0, length - committedBandPaintedRows)` is PENDING —
+   * held only in the in-memory band model, never painted, never archived.
+   * Drives the EXACT `overflowHasPending` signal (see decideCommitMode).
+   */
+  committedBandPaintedRows: number;
 }
 
 /** The routing decision + the geometry the caller's phases consume. */
@@ -48,7 +56,7 @@ export interface CommitMode {
   overflowPriorContiguous: boolean;
   /** Prior band + new lines (merged when contiguous, else just the new lines). */
   overflowRun: string[];
-  /** Prior band carries rows never painted on screen because the held overlay left only `room` rows visible. */
+  /** Prior band carries rows never painted on screen (`committedBandPaintedRows < committedBand.length`) — held only in the model, not on screen, not in scrollback. */
   overflowHasPending: boolean;
   /** Route this commit through the band-hold path (hold in the model; don't archive + truncate). */
   useBandHold: boolean;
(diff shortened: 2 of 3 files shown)


griffinwork40:
Verification complete. Both load-bearing claims hold, and I can now close the `room` thread:

- **api-compat**: exactly one production `decideCommitMode` call site (`committed-band-commit.ts:246`), wired with `committedBandPaintedRows: self.committedBandPaintedRows`. Every other call is test-only via `base()`. The non-optional field addition therefore can't leave an unwired production site → tsc-clean claim holds.
- **`room`**: assigned at line 125, returned in the struct at line 138 (so no `noUnusedLocals` break), but the production caller destructures only `{ useBandHold, overflowRun, maxBandModel }` and `git grep '\.room'` finds **zero** production consumers. After the proxy retirement, `CommitMode.room` survives only as a computed output that the new divergence test asserts on (`m.room === 7`). The PR-body phrase "still consumed by the fits-path math" is slightly imprecise, but retaining `room` is justified (the regression test's divergence proof reads it) and harmless.

I also independently checked the **behavioral-delta** claim: a true→false flip (which would be a *new* drop) requires `length > room` yet `paintedRows === length` (fully painted). Under the upheld invariant `paintedRows ≤ length` plus contiguity (`committedBandBottomRow === frameTop-1`), a fully-painted contiguous band satisfies `length ≤ room`, so that flip is unreachable — and even if reached, routing a fully-painted band to the fits path is correct (real content, not blanks). The change is strictly the bug-fix direction.

No critical/high findings → `/shadow-verify` is not triggered (and I've already re-derived the load-bearing claims against the ref).

---

# Code Review — PR #257

**Target:** PR #257 `fix(tui): use exact painted-row count for the band-hold pending signal (#255 follow-up)`
**Reviewed ref:** `196f818` · base `fix/tui-band-hold-pending-gap` ← head `fix/tui-overflow-pending-exact-count` (stacked on #255)
**Regime:** light (hotfix, 3 files, +51/−7) · **Stated intent:** present (PR title+body) — spec-compliance assessed

### Summary
Replaces the `overflowHasPending` geometry proxy (`committedBand.length > room`) with the exact painted-row count (`committedBand.length > committedBandPaintedRows`) in `decideCommitMode`. `committedBandPaintedRows` is threaded from the compositor's already-tracked state through the single production call site as a new non-optional `CommitModeInput` field, plus a discriminating regression test. No new state; no behavior change outside the targeted divergence case.

### Findings

| # | severity · confidence | dimension | location | finding |
|---|---|---|---|---|
| 1 | nit · high | test-coverage | `commit-mode.test.ts` | No test pins the boundary `committedBandPaintedRows === committedBand.length` (fully painted → `overflowHasPending` must be `false`). The fix's "no-op when fully painted" safety rests on this equality; a guard test would document it. **suggestion:** add a `base({ committedBand: band, committedBandPaintedRows: band.length, ... })` case asserting `overflowHasPending === false`. |
| 2 | nit · high | spec-compliance | `commit-mode.ts:50,138` (PR body) | `CommitMode.room` is now consumed by no production caller (only the proxy read it; `git grep '\.room'` → 0 production hits); it survives as computed output that the new test asserts on. The PR body's "still consumed by the fits-path math" is imprecise. Harmless and intentional (the divergence test reads it). **suggestion:** none required; optionally reword the field's doc to "retained for the divergence regression assertion." |

**Dimension verdicts** (read at `196f818`):
- **api-compat** — no issues found. Sole production call site wired; non-optional field forces compile-time coverage.
- **correctness** — no issues found. `committedBandPaintedRows` declared on `CommittedBandHost`; invariant `0 ≤ paintedRows ≤ band.length` upheld at all six assignment sites (commit Phase 3, fits/overflow, full-viewport hold, repin, two frame-preserve sites) and reset by `clearCommittedBand`; boolean logic correct; behavioral delta is strictly false→true in the divergence case.
- **spec-compliance** — intent met exactly; no scope creep (no paint-site or `useBandHold`/`overflowRun` logic touched).
- **test-coverage** — new regression test is discriminating (geometry `band=5, room=7, painted=3`: old proxy `5>7=false`, new signal `5>3=true`); fails on proxy, passes on fix.
- **perf-observability** — no issues found. One added field-read in a pure-arithmetic function.

### What was not checked
- Read against branch HEAD `196f818`; all citations verified at that ref.
- Did not run the suites myself — PR states `commit-mode.test.ts` 14/14 and 306 `terminal-compositor*` tests green; `tsc --noEmit` clean. The non-optional-field wiring claim was verified structurally (single production call site), which is consistent with a clean tsc.
- The `disarm()`/`lifecycle.ts` flush path that also reads `committedBandPaintedRows` is outside the diff and was not read.
- No integration/E2E coverage of the two-block scenario — the PR documents it as unreachable through the `@xterm/headless` harness (needs exactly one intervening repaint), which is why it's deliberately tested at the pure-unit level. Reasonable.

## ✅ MERGE

Clean, minimal, correct bug fix with a discriminating regression test. The two `nit`s are non-blocking — a boundary test would harden the no-op-when-painted guarantee, and one PR-body sentence about `room` is imprecise. Both are optional. Note this PR is stacked on #255 (base `fix/tui-band-hold-pending-gap`); merge after #255 or rebase onto `main` if #255 lands first.

*(Review is read-only — no files, commits, or PR state were modified. The repo's pre-existing dirty working-tree change was left untouched.)*

---

PR #258, opened 2026-06-23
feat(trace): emit tool_call + session_phase from openai-compatible provider

griffinwork40:
## What & why

The openai-compatible provider (`gpt-*`, `codex-*`, and local OpenAI-shim models — MLX / Qwen / llama.cpp / vLLM) emitted **zero** witness-trace events for tool calls and loop phases, while `anthropic-direct` emits them. Every non-Anthropic session was a blind spot for per-tool latency/error/truncation telemetry and the loop latency waterfall — even though the trace writer was already threaded into the provider.

Surfaced by a trace-parity audit (HIGH-severity gap).

## Change

Mirror `anthropic-direct/loop.ts` emission in the openai-compatible query loop:

- `session_phase`: `loop_start` (turn entry), `loop_end` (in a `try/finally` so it fires on every exit path — abort / error / clean end / iteration-cap), and `model_ttfb` (first translated stream event, once per API call, in **both** the Chat Completions and Responses wire paths).
- `tool_call`: `started` (before dispatch, so a crashing tool still leaves evidence) and `completed` (with `durationMs`, `resultBytes`, `isError`, `truncated`, `circuitBreaker?`, `failureClass?`) — the same field shape anthropic-direct uses.
- `traceWriter` threaded from `AgentConfig` via `buildQueryFromConfig`.

All emit calls are fire-and-forget (`void emit...`); a missing or broken writer is a no-op and never stalls the turn.

## Tests

+4 tests in `query.test.ts` — `tool_call` started+completed, `loop_start`/`loop_end`, `model_ttfb`, and no-throw / no-emit when `traceWriter` is absent. **68/68 pass.**

## Verification

- `pnpm exec vitest run src/agent/providers/openai-compatible/query.test.ts`: 68 passed

No new `@anthropic-ai/sdk` imports (the emit helpers are internal to `agent-afk`).

### DIFF
--- a/src/agent/providers/openai-compatible/query.test.ts
+++ b/src/agent/providers/openai-compatible/query.test.ts
@@ -2011,3 +2011,224 @@ describe('OpenAICompatibleQuery — retry / backoff (issue #126)', () => {
     ).toBe(true);
   });
 });
+
+// ---------------------------------------------------------------------------
+// Witness-layer trace emission — parity with anthropic-direct/loop.test.ts
+// ---------------------------------------------------------------------------
+
+describe('OpenAICompatibleQuery — witness-layer trace emission', () => {
+  // Build a two-turn scripted client: turn 1 returns a tool_call, turn 2 text.
+  function installToolThenTextClient(
+    toolId: string,
+    toolName: string,
+    toolArgs: string,
+  ): void {
+    const factory = () =>
+      ({
+        chat: {
+          completions: {
+            create: (() => {
+              let callIdx = 0;
+              return async (_args: unknown, _opts?: unknown) => {
+                const turn = callIdx++;
+                if (turn === 0) {
+                  // First call: tool_call finish
+                  return (async function* () {
+                    yield {
+                      choices: [
+                        {
+                          delta: {
+                            tool_calls: [
+                              {
+                                index: 0,
+                                id: toolId,
+                                type: 'function',
+                                function: { name: toolName, arguments: toolArgs },
+                              },
+                            ],
+                          },
+                        },
+                      ],
+                    };
+                    yield {
+                      choices: [{ delta: {}, finish_reason: 'tool_calls' }],
+                      usage: { prompt_tokens: 5, completion_tokens: 5, total_tokens: 10 },
+                    };
+                  })();
+                } else {
+                  // Subsequent calls: text finish
+                  return (async function* () {
+                    yield {
+                      choices: [
+                        { delta: { content: 'done' }, finish_reason: 'stop' },
+                      ],
+                      usage: { prompt_tokens: 10, completion_tokens: 2, total_tokens: 12 },
+                    };
+                  })();
+                }
+              };
+            })(),
+          },
+        },
+      }) as unknown as import('openai').default;
+    __setOpenAIClientFactory(factory);
+  }
+
+  it('emits tool_call started+completed for each dispatched tool', async () => {
+    const { InMemoryTraceWriter } = await import('../../trace/writer.js');
+    const writer = new InMemoryTraceWriter();
+
+    installToolThenTextClient('call_search_1', 'search', '{"q":"hello"}');
+
+    const handlerCalls: string[] = [];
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

The production change correctly implements every stated-intent clause, is fully backward-compatible, introduces no security or hot-path performance concerns, and faithfully mirrors the `anthropic-direct` provider's emission. All findings below are **low / nit** severity (test-coverage gaps on the trickier branches plus one cosmetic item) — none block merge.

Reviewed at branch HEAD `819e688`. Read-only review: every cited line was verified against the file state at that ref.

### Spec-compliance: intent met (no scope creep)
All intent clauses are satisfied in production code — `loop_start`/`loop_end` (try/finally, fires on every exit path), `model_ttfb` in **both** Chat Completions and Responses wire paths, `tool_call` started/completed with the full anthropic-direct field shape, and `traceWriter` threaded via `buildQueryFromConfig` (no-op guard when unset). No unrelated behavior changes were introduced.

### Security — no issues
Trace payloads carry **metadata only** — `toolUseId`, tool `name`, `inputBytes`/`resultBytes` (byte counts), `durationMs`, `isError`, `truncated`, and the model id. No raw tool input, no raw tool output `content`, and no secrets are emitted (verified against `trace/types.ts` payload schemas and confirmed at parity with `anthropic-direct/loop.ts`).

### api-compat — no issues
`OpenAICompatibleQueryOptions.traceWriter?`, the constructor field, and the `buildQueryFromConfig` read of the pre-existing optional `config.traceWriter` are all additive-optional. No exported signature was broken; existing callers that never set a writer behave identically.

**S-1 · low · spec-compliance · `query.test.ts:2172`** — `model_ttfb` is tested only on the Chat Completions path; no test sets `useResponsesApi: true` to exercise the Responses-path emission, even though the intent calls out both wire paths. Production code at `query.ts:839` is correct; this is a verification gap only. *Fix:* add a Responses-API mock test asserting one `model_ttfb` with `resolvedModel` set.

**T-1 · low · test-coverage · `query.test.ts:2133`** — the `loop_end` test covers only the clean-completion path; the abort-mid-stream and stream-error exit paths are not asserted to also emit `loop_end` (the existing abort tests pass no `traceWriter`). The production `finally` block is correct, but the two non-happy branches are unverified. *Fix:* add an abort-with-writer test and a stream-error-with-writer test, each asserting `loop_end` fires.

**T-2 · low · test-coverage · `query.test.ts:2195`** — the per-retry re-emission of `model_ttfb` (`ttfbEmitted` resets each `for(;;)` iteration) is not verified; the sole assertion is `toHaveLength(1)` on the happy path. *Fix:* add a mid-stream-retry test with a writer asserting `ttfbEvents.length === 2`.

**C-2 · low · correctness · `query.ts:1064-1077`** — on the `signal.aborted` fast-abort branch in `dispatchAndAppend`, `tool_call.started` has already fired per call but `tool_call.completed` is never emitted, leaving orphaned `started` events in the trace. This is **symmetric** with `anthropic-direct/loop.ts:554-567`, which has the identical gap — not a regression. *Fix (optional, brings both providers to parity):* emit `completed` with `isError: true` in the abort branch, or document the invariant that abort-path `started` events are intentionally unmatched.

**C-1 · nit · correctness · `query.ts:496`** — the `yield* this._runTurnInner(...)` line is at the same 4-space indent as its enclosing `try {` rather than indented to 6. Purely cosmetic (scoping is unaffected) and matches the identical style in `anthropic-direct/loop.ts:235-236`. *Fix:* indent to 6 spaces for readability.

### What was not checked
- The test suite was **not executed** (the PR's "68/68 pass" claim was not re-run; findings are from static read at the ref).
- `translateResponsesEvent` / `translateChunk` were not inspected for whether they can yield zero events for a stream — a reachable edge where `model_ttfb` would be deferred to the first non-empty event rather than the first API byte (telemetry-accuracy only, not correctness/security).
- Downstream witness-trace consumers were not audited for assumptions about `started`/`completed` pairing on abort paths.

PR #259, opened 2026-06-23
fix(trace): wire witness trace into farm sessions and worker forks

griffinwork40:
## What & why

The `farm` surface constructed its sessions with **no `traceWriter` and no `surface`**, so farm worker sessions emitted no witness trace and reported `origin: 'unknown'`. Surfaced by a trace-parity audit (HIGH-severity gap).

## Change

- **`farm.ts`**: create a trace writer via `createDefaultTraceWriter` (honoring `AFK_TRACE_DISABLED`) and pass it plus `surface: 'cli'` to `SubagentManager`.
- **`subagent.ts`**: `SubagentManager` now optionally inherits `traceWriter` + `surface` into every forked child whose `config.{traceWriter,surface}` is unset — the same guarded pattern as the existing `cwd` inheritance. Explicit per-fork values still win, and it only activates when a caller sets the manager option (currently only `farm`), so there is **zero behavior change** for other fork callers (the `agent` tool, skills, compose).

Result: farm branch workers now emit full witness traces with `origin: 'cli'`.

## Tests

- `farm.ts` wiring tests (writer created once, reaches the manager, `surface` set, `AFK_TRACE_DISABLED` handled)
- `farm.worker-trace.test.ts` — forked worker inherits writer + surface
- 6 `subagent.ts` inheritance unit tests (inherit / override / omit, for both `traceWriter` and `surface`)

**76/76 pass.**

## Verification

- targeted vitest: 76 passed

### DIFF
--- a/src/agent/subagent.test.ts
+++ b/src/agent/subagent.test.ts
@@ -231,6 +231,89 @@ describe('SubagentManager', () => {
     );
   });
 
+  // -------------------------------------------------------------------------
+  // traceWriter + surface inheritance (witness layer / origin attribution)
+  // Mirrors the cwd inheritance tests above: manager-level values propagate
+  // into child AgentConfig so farm/DAG workers report the correct trace origin
+  // without per-call plumbing.
+  // -------------------------------------------------------------------------
+
+  it('inherits traceWriter from manager when child config omits it', async () => {
+    shared.lastConfig = null;
+    const fakeWriter = { write: vi.fn(), close: vi.fn(), getTracePath: vi.fn() };
+    const mgr = new SubagentManager({ traceWriter: fakeWriter as unknown as import('./trace/index.js').TraceWriter });
+    await mgr.forkSubagent({
+      parent: { sessionId: 'p' },
+      config: { model: 'sonnet' },
+    });
+    expect(shared.lastConfig).toEqual(
+      expect.objectContaining({ traceWriter: fakeWriter }),
+    );
+  });
+
+  it('lets explicit child traceWriter override the manager default', async () => {
+    shared.lastConfig = null;
+    const managerWriter = { write: vi.fn(), close: vi.fn(), getTracePath: vi.fn() };
+    const childWriter = { write: vi.fn(), close: vi.fn(), getTracePath: vi.fn() };
+    const mgr = new SubagentManager({
+      traceWriter: managerWriter as unknown as import('./trace/index.js').TraceWriter,
+    });
+    await mgr.forkSubagent({
+      parent: { sessionId: 'p' },
+      config: {
+        model: 'sonnet',
+        traceWriter: childWriter as unknown as import('./trace/index.js').TraceWriter,
+      },
+    });
+    expect(shared.lastConfig).toEqual(
+      expect.objectContaining({ traceWriter: childWriter }),
+    );
+  });
+
+  it('omits traceWriter on child when neither manager nor child config set it', async () => {
+    shared.lastConfig = null;
+    const mgr = new SubagentManager();
+    await mgr.forkSubagent({
+      parent: { sessionId: 'p' },
+      config: { model: 'sonnet' },
+    });
+    expect((shared.lastConfig as unknown as Record<string, unknown>)['traceWriter']).toBeUndefined();
+  });
+
+  it('inherits surface from manager when child config omits it', async () => {
+    shared.lastConfig = null;
+    const mgr = new SubagentManager({ surface: 'cli' });
+    await mgr.forkSubagent({
+      parent: { sessionId: 'p' },
+      config: { model: 'sonnet' },
+    });
+    expect(shared.lastConfig).toEqual(
+      expect.objectContaining({ surface: 'cli' }),
+    );
+  });
+
+  it('lets explicit child surface override the manager default', async () => {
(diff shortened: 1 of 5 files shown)


griffinwork40 (commented):

A small, well-tested, behavior-preserving telemetry-wiring fix. It meets its core stated intent — farm worker sessions now receive a `traceWriter` + `surface: 'cli'`, and `SubagentManager` inherits both into forked children using the exact guarded pattern already established for `cwd`. No security, correctness, or hot-path performance concerns. All findings below are **low / nit** — none block merge.

Reviewed at branch HEAD `38f730fe`. Read-only review: every cited line was verified against the file state at that ref, and the cross-file behavior-change claim was independently re-derived by fetching `skill-executor.ts` and `subagent-executor.ts` at the same ref.

### Spec-compliance: core intent met (one wording imprecision)
The declared mechanism is implemented faithfully: `farm.ts` opens the writer via `createDefaultTraceWriter()` honoring `AFK_TRACE_DISABLED` (null-guarded), passes `traceWriter` + `surface: 'cli'` to the manager, and `subagent.ts` fills `childConfig.{traceWriter,surface}` only when the per-fork config omits them (explicit per-fork values still win, because the guards are tested against the already-spread `...options.config`). The guard semantics match the `cwd`/`baseUrl` inheritance directly above them.

### Security — no issues
The change threads only a `TraceWriter` reference and a `surface` enum string. No credential, secret, or PII is newly captured or written into trace payloads; the `write` call sites are unchanged. Verified by reading `subagent.ts` and `farm.ts` at the ref.

### api-compat — additive-optional, blast radius confined
`SubagentManagerOptions.surface?` is a new optional field (and `traceWriter?` is a doc-only change to a pre-existing optional). No exported signature is broken; callers that set neither behave identically. Sweep of every production `new SubagentManager(...)` call site at the ref confirms the top-level entrypoints — `chat.ts:435`, `daemon.ts:91`, `bootstrap.ts:185`, `telegram.ts:245`, and `compose-executor.ts:543` — pass **no** manager-level `traceWriter`/`surface`, so none of those root paths change. See A-1 for the one path that does change.

**A-1 · low · spec-compliance · PR description / `src/agent/tools/skill-executor.ts:569,694,989` · ref:38f730fe · file-state** — The PR claims "ZERO behavior change for other fork callers (the agent tool, skills, compose)." That is imprecise: `skill-executor.ts` constructs its skill-fork child managers with a manager-level `traceWriter` (`...(this.ctx.traceWriter !== undefined ? { traceWriter: this.ctx.traceWriter } : {})`), and `subagent-executor.ts` builds its `agent`-tool `childConfig` with **no** `traceWriter` field (grep: 0 references at the ref). So after this PR, a grandchild `agent` fork dispatched *under a skill fork* now inherits `parentTraceWriter` where it previously got `undefined`. This is **benign and arguably the intended completion of trace wiring** — `surface` was already propagated down that executor chain (`subagent-executor.ts:626`) and origin is derived from it — but it is a real change to that path, so the blanket "zero behavior change" sentence overstates the guarantee. *Fix:* narrow the wording to "no change for callers that do not set a manager-level `traceWriter`/`surface`; the skill→agent grandchild path now correctly inherits the writer."

**C-1 · low · correctness (doc-accuracy) · `src/cli/commands/farm.ts:361-368` · ref:38f730fe · file-state** — The comment block above `const trace = createTraceWriterFn();` still describes the `dag-subagent.ts` / manager-level traceWriter inheritance as a future "remaining gap … tracked as a follow-up," but this same PR's `subagent.ts` change implements exactly that inheritance. The comment now contradicts the code in the same file. *Fix:* delete or replace lines 364-368 with a one-line note that the inheritance is now active and the follow-up is closed.

**T-1 · low · test-coverage · `src/cli/commands/farm.test.ts` (new trace-wiring block) · ref:38f730fe · diff-context** — The test named "passes traceWriter from `_createTraceWriter` into the SubagentManager" captures `capturedManager` but asserts only that it is defined and that the mock was called once — it never asserts the manager actually received `traceWriter`. *Fix:* add `expect(capturedManager).toMatchObject({ traceWriter: fakeWriter })`.

**T-2 · nit · test-coverage · `src/cli/commands/farm.worker-trace.test.ts:1-12` · ref:38f730fe · diff-context** — The file-level doc-comment says it verifies "farm worker sessions inherit traceWriter and surface from the SubagentManager," but `SubagentManager` is mocked to `return {}`, so `forkSubagent` is never exercised — the tests prove only that `farm.ts` passes the correct *constructor options*. The actual fork inheritance is (correctly) covered in `subagent.test.ts`. *Fix:* reword the comment to "verifies farm.ts passes correct constructor options to SubagentManager; end-to-end fork inheritance is covered in subagent.test.ts." No test-logic change needed.

### What was not checked
- The test suite was **not executed** (the PR's "76/76 pass" claim was not re-run); findings are from static read at ref `38f730fe`.
- `dag-subagent.ts` fork internals were not deep-audited beyond confirming it constructs no manager-level `traceWriter`/`surface` (the PR does not claim to change it).
- `NdjsonTraceWriter.write` payload contents and `session-identity.ts` `deriveOrigin` logic were not re-derived (neither is modified by this diff).
- Stated intent (PR title + body) was available, so spec-compliance **was** assessed.

PR #260, opened 2026-06-23
fix(trace): pass traceWriter to MCP manager on telegram surface

griffinwork40:
## What & why

`telegram/mcp-session.ts` called `McpManager.fromConfig` without a `traceWriter`, so per-server MCP connect/lifecycle events (`mcp_connect_*` / `mcp_server_*`) were invisible on the telegram surface. The daemon path passes the writer; telegram didn't. Surfaced by a trace-parity audit (MED-severity gap).

## Change

Root cause: telegram connected MCP **before** the session (and its trace writer) was constructed, so there was no writer to pass. Fix:

- Export `createTelegramTraceWriter()` so `telegram.ts` can create the writer early and thread the **same** instance into both `loadTelegramMcpManager(..., { traceWriter })` and `constructTelegramSession(..., { traceWriter })` — one trace file then covers the full lifecycle (MCP connect → session turns).
- Added the `mcp_connect_start` / `mcp_connect_done` span around `fromConfig` (matches the `chat.ts` pattern).

> **Merge note:** also touches `src/telegram.ts` (3 lines). The PR `fix/routing-identity-completeness` also edits `telegram.ts` (a different hunk) — whichever merges second may need a trivial rebase.

## Tests / verification

- `construct-session.test.ts` + `mcp-session.test.ts`: **14 passed**

### DIFF
--- a/src/telegram.ts
+++ b/src/telegram.ts
@@ -34,7 +34,7 @@ import { TelegramBot } from './telegram/bot.js';
 import { parseAllowedChatIds } from './telegram/allowlist.js';
 import { validateBotToken } from './telegram/setup-wizard.js';
 import { AgentSession } from './agent/session.js';
-import { constructTelegramSession } from './telegram/construct-session.js';
+import { constructTelegramSession, createTelegramTraceWriter } from './telegram/construct-session.js';
 import { createDefaultHookRegistry } from './agent/default-hook-registry.js';
 import { seedPersistedGrants } from './agent/permissions-store.js';
 import { loadHooksConfig } from './agent/hooks/config-loader.js';
@@ -231,7 +231,13 @@ async function main() {
       );
 
       const sessionCwd = sessionConfig.cwd ?? telegramCwd;
-      const mcpManager = await loadTelegramMcpManager(sessionCwd);
+      // Create the trace writer before loadTelegramMcpManager so the MCP
+      // connect phase (mcp_server_start/done, mcp_connect_start/done) is
+      // captured in the same session trace as the rest of the Telegram session.
+      const telegramTraceWriter = createTelegramTraceWriter();
+      const mcpManager = await loadTelegramMcpManager(sessionCwd, {
+        ...(telegramTraceWriter !== null ? { traceWriter: telegramTraceWriter } : {}),
+      });
       let returnedSession: AgentSession | undefined;
       try {
         let directProvider;
@@ -384,7 +390,7 @@ async function main() {
           ...(sessionCwd !== undefined && sessionCwd.length > 0 ? { cwd: sessionCwd } : {}),
           provider: directProvider,
           hookRegistry: telegramHookBundle.registry,
-        }), mcpManager);
+        }, { traceWriter: telegramTraceWriter }), mcpManager);
         returnedSession = session;
         // Wire the path-approval grant ref to the provider so elicitation
         // approvals mutate readRoots / writeRoots on the right backend.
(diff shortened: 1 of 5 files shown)


griffinwork40 (commented):

A small (163-line), well-tested, behavior-preserving telemetry-wiring hotfix that faithfully implements its stated intent: the Telegram surface now creates the witness trace writer *before* `loadTelegramMcpManager`, threads the **same** writer instance into both the MCP connect phase and `constructTelegramSession`, and wraps `McpManager.fromConfig` in an `mcp_connect_start`/`mcp_connect_done` span — closing the MED-severity trace-parity gap with the daemon path. No security, correctness, or hot-path performance concerns. The single finding below is a **nit** — it does not block merge.

Reviewed at branch HEAD `ea74f7dc`. Read-only review: every cited API surface was verified against the file state at that ref via the GitHub contents API.

### Spec-compliance: intent met (no scope creep)
Every clause of the stated intent is implemented in production code: `createTelegramTraceWriter()` is exported from `construct-session.ts`, the writer is created early in `telegram.ts` and shared into both `loadTelegramMcpManager(..., { traceWriter })` and all `constructTelegramSession(..., { traceWriter })` call sites, and the `mcp_connect_start`/`mcp_connect_done` span around `fromConfig` mirrors the `chat.ts` pattern. No unrelated behavior changes were introduced.

### Security — no issues
The change threads only a `TraceWriter` reference and a `surface` enum string. No credential, secret, or PII is newly captured; the trace `write` call sites are unchanged. Verified by reading `telegram.ts` and `mcp-session.ts` at the ref.

### api-compat — additive-optional, surfaces confirmed
All key API surfaces this PR depends on were verified to exist at the ref:
- `McpManager.fromConfig(servers, opts)` accepts `{ warnings, traceWriter }` (`McpManagerInitOptions`, `src/agent/mcp/index.ts`).
- `emitSessionPhase(writer, { phase, durationMs?, metadata? })` matches the call shape (`src/agent/trace/emit.ts`).
- `'mcp_connect_start'` / `'mcp_connect_done'` are valid `SessionPhaseName` union members (`src/agent/trace/types.ts`).
- `createTelegramTraceWriter` is exported from `construct-session.ts`; `telegramTraceWriter` is threaded into all 3 call sites in `telegram.ts`.
- `LoadTelegramMcpManagerOptions.traceWriter?` is a new optional field; the `'traceWriter' in deps` guard correctly distinguishes `undefined` (not supplied → run factory) from `null` (explicit disable). No exported signature was broken.

### test-coverage — adequate
4 new tests cover: pre-created writer reuse (factory not called), factory fallback, explicit-`null` suppression, and `traceWriter` forwarding into `McpManager.fromConfig` via a spy.

### perf-observability — no issues
The two new emit calls are fire-and-forget (`void emitSessionPhase(...)`); a broken writer never stalls or crashes the connect phase. The `mcp_connect_done` emit is in a `finally`, so it fires on every exit path.

### Findings (non-blocking)

**N-1 · nit · correctness · `src/telegram.ts:237-239` · ref:ea74f7dc · diff-context** — The call-site uses a `null`-guard conditional spread, but `LoadTelegramMcpManagerOptions.traceWriter` is typed `TraceWriter | undefined` (never `null`), so the guard is more verbose than needed. Functionally correct — purely cosmetic. *Fix:* `loadTelegramMcpManager(sessionCwd, { traceWriter: telegramTraceWriter ?? undefined })`.

### What was not checked
- Read against branch HEAD `ea74f7dc` — all cited API surfaces verified at that ref via the GitHub contents API. Not run inside a local clone.
- The test suite and `tsc` were **not executed** (the PR's "14 passed / lint clean" claim was not re-run); findings are from static read at the ref.
- `McpManager.fromConfig`'s internal per-server `mcp_server_start`/`mcp_server_done` emit path was not deep-audited (not modified by this diff).
- The PR's own merge note flags a potential trivial rebase against `fix/routing-identity-completeness`, which also edits `telegram.ts` (a different hunk) — not a code defect, just a sequencing note.
- Stated intent: PR #260 title+body — spec-compliance assessed.

PR #262, opened 2026-06-23
feat(trace): emit browser_event from browser tool handlers

griffinwork40:
## What & why

`emitBrowserEvent` was defined and the trace reader handled the `browser_event` kind, but it was **never called** — dead code. The 5 browser handlers carried comments deferring emission because `ToolHandlerContext` carried no trace writer. Surfaced by a trace-parity audit (MED-severity gap).

## Change

- **`types.ts`**: added optional `traceWriter?` + `toolUseId?` to `ToolHandlerContext` (additive, back-compat).
- **`dispatcher.ts`**: new `callHandlerContext(call)` augments the base handler context with the per-call `toolUseId` + `traceWriter`; wired into the 2 handler-dispatch sites. Blast radius: exactly those 2 sites — the base `handlerContext` getter stays call-agnostic.
- The 5 browser handlers (`open` / `act` / `observe` / `screenshot` / `close`) now emit `browser_event` (url-before/after, status, `toolUseId`) on each outcome path; the stale deferral comments are removed.

## Tests

+ `browser-event-emission.test.ts` — 26 tests (5 handlers × ok / error / blocked / ambiguous paths + a `toolUseId` correlation suite). **97 passed** including the dispatcher suite.

## Verification

- targeted vitest: 97 passed

## Note

`browser_screenshot` reports `null` for url (non-navigating, no URL field), and `browser_act` uses the post-action URL for both before/after (the handler has no pre-action URL). Both are consistent with the `BrowserEventPayload` contract; a future `currentUrl()` provider method could refine the before/after distinction for navigation-triggering clicks.

### DIFF
--- a/src/agent/tools/dispatcher.ts
+++ b/src/agent/tools/dispatcher.ts
@@ -263,6 +263,11 @@ export class SessionToolDispatcher implements ToolDispatcher {
   /**
    * Returns a fresh snapshot of the current handler context. Called for every
    * handler invocation so grant mutations are always reflected.
+   *
+   * Note: `toolUseId` and `traceWriter` are NOT included here — they are
+   * per-call values added inline by `execute()` and `executeCore()` via
+   * `callHandlerContext(call)` so the getter stays call-agnostic and can be
+   * used safely by code that doesn't have a live ToolCall reference.
    */
   private get handlerContext(): ToolHandlerContext {
     return {
@@ -272,6 +277,19 @@ export class SessionToolDispatcher implements ToolDispatcher {
       writeRoots: this._writeRoots.slice(),
       ...(this._allowAll ? { allowAll: true } : {}),
       ...(this._env !== undefined ? { env: this._env } : {}),
+    };
+  }
+
+  /**
+   * Returns a per-call handler context that augments the base `handlerContext`
+   * with the tool-call-specific fields (`toolUseId`, `traceWriter`). Used by
+   * `execute()` and `executeCore()` when dispatching to a named handler.
+   */
+  private callHandlerContext(call: ToolCall): ToolHandlerContext {
+    return {
+      ...this.handlerContext,
+      toolUseId: call.id,
+      ...(this.traceWriter !== undefined ? { traceWriter: this.traceWriter } : {}),
     };
   }
 
@@ -622,7 +640,7 @@ export class SessionToolDispatcher implements ToolDispatcher {
     // 5. Execute handler
     let result: ToolResult;
     try {
-      result = await handler(call.input, call.signal, this.handlerContext);
+      result = await handler(call.input, call.signal, this.callHandlerContext(call));
     } catch (err) {
       const message = err instanceof Error ? err.message : String(err);
       result = { content: `Tool execution error: ${message}`, isError: true };
@@ -850,7 +868,7 @@ export class SessionToolDispatcher implements ToolDispatcher {
 
     let result: ToolResult;
     try {
-      result = await handler(call.input, call.signal, this.handlerContext);
+      result = await handler(call.input, call.signal, this.callHandlerContext(call));
     } catch (err) {
       const message = err instanceof Error ? err.message : String(err);
       result = { content: `Tool execution error: ${message}`, isError: true };
--- a/src/agent/tools/handlers/browser-act.ts
+++ b/src/agent/tools/handlers/browser-act.ts
@@ -18,16 +18,13 @@
  * @module agent/tools/handlers/browser-act
  */
 
-import type { ToolHandler } from '../types.js';
+import type { ToolHandler, ToolHandlerContext } from '../types.js';
 import type { BrowserHandlerOptions } from './browser-open.js';
 import type { Target, ActAction } from '../../../browser/types.js';
 import { truncateTargetText, hashSelector } from '../../../browser/sanitize.js';
 import { env } from '../../../config/env.js';
-
-// History: browser_event witness emission is a no-op in this handler.
-// See browser-open.ts for the full rationale. The target field sanitization
-// helpers (truncateTargetText, hashSelector) are imported and called so the
-// logic is in place for when browser_event emission is wired.
+import { emitBrowserEvent } from '../../trace/emit.js';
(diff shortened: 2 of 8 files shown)


griffinwork40:
Verification complete. Final evidence resolves the severity of both findings:

- **`urlBefore` mislabel:** the *only* consumer of these fields is `src/cli/commands/trace.ts:346`, which reads **`urlAfter` only** (the correct post-action value). Nothing reads or diffs `urlBefore` → the contract deviation is **latent**, not actively misleading any consumer. Downgrades the HIGH.
- **Raw URL in trace:** `redactSecrets` exists (`sanitize.ts:91`) and is wired for fill-values/prompt-dump, but not applied to the new URL fields; the witness trace is local and the URL already lives in the session transcript → real hygiene gap, marginal exposure. Downgrades the MEDIUM.

No critical/high finding survives evidence-based synthesis — the top finding's factual basis was independently re-derived from source during Wave 1.5 (citation + consumer-reachability), which is the function `/shadow-verify` performs, so no separate verifier wave is needed.

---

# Code Review — PR #262 `feat(trace): emit browser_event from browser tool handlers`

**Target:** `main ← feat/browser-event-emission` · reviewed ref `897a02f` · feature · 8 files · +796/−55 · regime: full (2 parallel agents + inline verification)
**Stated intent:** wire previously-dead `emitBrowserEvent` into the 5 browser handlers via additive `traceWriter?`/`toolUseId?` on `ToolHandlerContext`. **Spec assessed → MET.**

**1 · low · confidence high · perf-observability · `browser-act.ts:256–264` · ref:897a02f · file-state**
*(Agent B rated this **high**; downgraded — rationale below.)*
`browser_act` ok-path emits `urlBefore: result.url, urlAfter: result.url` — both the *post-action* URL — while `BrowserEventPayload.urlBefore` is documented (`types.ts:463`) as "Page URL captured BEFORE the action took effect." For a navigation-triggering click the trace shows `urlBefore == urlAfter == destination`, erasing the navigation.
Evidence:
```
// browser-act.ts (ok path):    urlBefore: result.url,  urlAfter: result.url,
// types.ts:463: /** Page URL captured BEFORE the action took effect. */
```
**Downgrade rationale (verified):** the sole consumer `src/cli/commands/trace.ts:346` reads `urlAfter` only — nothing reads/diffs `urlBefore`, so the deviation misleads no current consumer (latent). It is also explicitly disclosed in the PR Note.
Suggestion: emit `urlBefore: null` (the type permits it) — honest "no pre-action URL captured" instead of a false equality.

**2 · low · confidence medium · security · `browser-{open,act,observe}.ts` emit sites + `trace/types.ts:465,469` · ref:897a02f · file-state**
*(Agent A rated this **medium**; downgraded — rationale below.)*
`urlBefore`/`urlAfter` are written as raw, unredacted URL strings to the witness trace; URLs can carry secrets in query params (`?token=…`, `?api_key=…`). `redactSecrets()` exists (`sanitize.ts:91`) and is applied elsewhere (`prompt-dump.ts:296`, fill-values) but **not** to these URL fields (grep of handlers: zero `redactSecrets` calls).
**Downgrade rationale (verified):** the witness trace is a local artifact (`~/.afk/state/witness/`), and the same URL already lives in the session transcript — the marginal new exposure is small. Real hygiene inconsistency, not an exfiltration path.
Suggestion: pipe `urlBefore`/`urlAfter` through `redactSecrets()` before payload construction, matching the existing fill-value pattern.

## Clean dimensions (read at ref 897a02f)

- **correctness — no issues.** `result as BrowserObservation` cast in `browser-open.ts` is sound (blocked-by-policy branch returns first; `BrowserObservation` has no `outcome`); `result.url` in `browser-act.ts` ok-path is unreachable by non-observations (ambiguous/blocked return inside the `'outcome' in result` block); `toolUseId: … ?? ''` fallback only fires in tests — `ToolCall.id` is always set by `callHandlerContext(call)` in production.
- **api-compat — no breaking changes.** Both `ToolHandlerContext` fields optional/additive; `witnessTarget` is not exported (zero production importers — grep-confirmed → `[UNVERIFIED: no production importers]`, nit); `callHandlerContext` wired into both and only both dispatch sites (`dispatcher.ts:643, 871`); `this.traceWriter` is the real session witness writer (`dispatcher.ts:227,251`).
- **no unhandled-rejection risk.** `emitBrowserEvent` (`emit.ts:144–155`) wraps `writer.write` in try/catch → `debugLog`; the module header documents "Errors are swallowed." `void emitBrowserEvent(...)` is structurally safe.
- **spec-compliance — MET.** `emitBrowserEvent` was dead at parent `5f4f07b` (only a design doc + a code comment referenced it; zero call sites — grep-confirmed). All 5 handlers emit on every post-provider outcome path (open: ok/blocked/error; act: ok/ambiguous/blocked/error; observe: ok/error; screenshot: ok/error; close: ok/error). Pre-provider early returns (abort/parse/sessionId/provider-get) don't emit by design — not a gap, not claimed. No scope creep.
- **test-coverage — comprehensive.** New `browser-event-emission.test.ts` covers every emitting outcome path + a 5-handler `toolUseId` correlation suite + the no-`traceWriter` no-op path. `flush()`/`setImmediate` is sufficient — `InMemoryTraceWriter.write` has a synchronous body (no real I/O), so no flakiness.

## What was not checked
- Citations verified by re-reading `897a02f` directly (inline `git show`/`git grep`); intent = PR #262 title+body → spec-compliance assessed.
- Did **not** re-run the suite — relied on the PR's "97 passed / lint clean" claim (not re-executed here).
- Did **not** review non-diff surfaces (Telegram/daemon) or the pre-existing `_redactedValue` underscore pattern in `playwright/index.ts:262` (out of scope).

## Merge recommendation

**MERGE.** Additive, back-compatible, well-tested feature that achieves its stated purpose — wiring previously-dead `emitBrowserEvent` across all 5 browser handlers with full outcome-path coverage. No critical/high defect survives verification. Both surviving findings are **low**-severity quality follow-ups (honest `urlBefore: null`; redact URL fields) that can land in this PR or a fast follow-up — neither blocks merge.

(Read-only review — no files, commits, or PR state were mutated. Stopping here per skill contract.)

---

PR #263, opened 2026-06-23
feat(improve): fixture-replay eval-run proves recorded failure fixed

griffinwork40:
## Summary

Adds a **fixture-replay** validation layer to `afk improve eval-run` for the `repeated-tool-use` pattern — step 1 of automating the improve pipeline.

- Previously eval-run validated only that a guardrail **exists** (against a synthetic stimulus). It never proved a card's **actual recorded failure** is handled, so a passing run could green-light an unfixed card.
- Fixture-replay re-drives the recorded loop — tool name + run-length extracted from the committed `.fixture.jsonl` — through the **live** `SessionToolDispatcher` circuit breaker and asserts it is neutralised at the recorded magnitude. **`fail` = still reproduces; `pass` = fixed.**
- A literal "re-scan the fixture and expect zero findings" can't work: a detector is a pure function of immutable trace bytes, so fixing code never rewrites an old recording. The boundary (re-drives the recorded loop *shape*, not the original tool/LLM) is documented.
- **Safety invariant:** a *skipped* replay (intact fixture that no longer reproduces the pattern) reports `unsupported`, **never `pass`** — guardrail presence is never mistaken for card-specific behavioural proof.

Scope: `dispatcher.ts` is import-only; no `schemas.ts` change. New `docs/improve-eval-run.md` documents the validator, the two layers, the status truth table, and how to add a handler for another pattern.

## Test results

- `pnpm lint` (`tsc --noEmit`, strict): **clean**.
- `pnpm test` (full suite): **9910 passed, 1 failed, 12 skipped**.
  - The single failure is `src/agent/providers/anthropic-direct.test.ts > compact()` — a model-id normalization assertion (`claude-haiku-4-5` vs `claude-haiku-4-5-20251001`). It is **pre-existing and unrelated**: this PR touches only `src/improve/eval-run/**` + `docs/`, which has zero overlap with the anthropic-direct provider.
- `src/improve/eval-run` suite in isolation: **46/46 pass** (contracts 13, replay 9, runner 24).

## Verification

Adversarial verifier wave (claims re-derived from the diff alone, not the commit prose):

- **CLAIM A — skipped replay ⇒ `unsupported`, never `pass`:** CONFIRM (`replay.ts` skipped branch emits a `skipped` reproduce check + no neutralize check → `runner.ts` computes `replayInconclusive=true` → `decideStatus` returns `unsupported`).
[...]

PR #264, opened 2026-06-23
feat(improve): fixture-replay eval-run for closure-anomaly

griffinwork40:
## What

Adds the **second** fixture-replay handler to `afk improve eval-run` (after `repeated-tool-use`, merged in #263), covering the **`closure-anomaly`** pattern. This is the next step of automating the improve pipeline — replay proves *this card's recorded failure* is actually fixed by the live guardrail, not merely that a guardrail exists.

## Replay model (`closure-anomaly`)

1. **Reproduce** — parse the committed fixture, run `detectClosureAnomaly`, and select the finding by the recorded evidence `seq` (this detector has **no fingerprint**, so seq is the selector). Extract the recorded `closureReason`.
2. **Re-drive** — call the live `buildClosureGuidance(reason)` (`src/agent/session/closure-guidance.ts`) and assert the recorded anomalous closure now carries actionable recovery guidance.

| Outcome | Verdict |
|---|---|
| Reproduces **and** guardrail returns a recovery hint | `pass` — neutralized |
| Reproduces **but** reason not covered (null guidance) | `fail` — fail-closed |
| Intact fixture, no anomalous closure | `unsupported` — skipped (never `pass`) |
| Fixture missing / sha256 mismatch | `fail` via `fixture-integrity` before replay |

**Boundary (no overclaiming):** neutralization here is **advisory** — the abort still aborts; "neutralized" means the closure no longer surfaces *without a recovery path*. This is the closure analog of the repeat-loop handler's "re-drives the loop shape, not the original tool/LLM" boundary. The guardrail covers `abort` today; uncovered anomalous reasons (e.g. `timeout`) **fail-closed**, mirroring how the repeat-loop handler fails a sub-threshold loop.

## Runner generalization

The runner's replay detection moves from a hardcoded `REPLAY_CHECK_NEUTRALIZED` lookup to an `isReplayNeutralizeCheck()` predicate, so any registered handler's neutralize check drives both the verdict and the (now pattern-agnostic) markdown disclaimer.

## Why `closure-anomaly` (and not the other two)

- **`subagent-block`** — the guardrail (`buildSkillMaxDepthRefusal(depth, maxDepth)`) needs depth, which the detector doesn't capture; it also fires on *any* `SubagentStart` block, not just max-depth. Not deterministically re-drivable.
- **`tool-failure-density`** — the "guardrail" is just *the detector being enabled by default*; there's no runtime neutralizer to re-drive.

## Scope (unchanged)

No LLM, no apply, no auto-triage, no scheduling, no forge bridge, no schema changes.

## Tests / gates

- New committed fixture: `src/improve/eval-run/__fixtures__/closure-anomaly-abort.fixture.jsonl`.
- `replay.test.ts` +7, `runner.test.ts` +4. The existing closure-anomaly contract test was updated to supply an abort-closure fixture (registering the handler now also exercises replay).
- `pnpm lint` clean; **320/320 `src/improve/` tests pass**. Full suite: the only failure is a pre-existing, unrelated `anthropic-direct.test.ts` `compact()` model-id env failure, untouched by this PR.

### DIFF
--- a/docs/improve-eval-run.md
+++ b/docs/improve-eval-run.md
@@ -30,8 +30,9 @@ covers any specific recorded failure.
 ### 2. Fixture-replay (`replay.ts`) — "is the behaviour fixed?"
 
 For patterns with a registered replay handler (**currently `repeated-tool-use`
-only**), the runner re-drives **this card's actual recorded failure** through
-the live guardrail and asserts it would not recur at the recorded magnitude.
+and `closure-anomaly`**), the runner re-drives **this card's actual recorded
+failure** through the live guardrail and asserts it would not recur at the
+recorded magnitude.
 
 This is the layer that distinguishes *"a guardrail exists"* from *"this
 failure is fixed."* It is strictly stronger than the presence check because it
@@ -51,29 +52,41 @@ which is exactly why this layer does something else.
 
 #### What it actually does
 
-1. **Reproduce.** Parse the committed fixture and run the detector to confirm
-   it still encodes the recorded pattern (and to extract the recorded
-   `toolName` + `runLength`).
-2. **Re-drive.** Feed that recorded loop's shape (byte-identical calls to the
-   recorded tool, bounded to the breaker threshold) through the **live**
-   `SessionToolDispatcher` and observe whether the circuit breaker trips at or
-   before the recorded length.
+Each handler runs the same two-step shape — **reproduce** (confirm the fixture
+still encodes the pattern and extract the recorded failure shape) then
+**re-drive** (feed that shape through the live guardrail) — specialised per
+pattern:
+
+- **`repeated-tool-use`** — reproduce extracts the recorded `toolName` +
+  `runLength`; re-drive feeds that loop's shape (byte-identical calls to the
+  recorded tool, bounded to the breaker threshold) through the live
+  `SessionToolDispatcher` and observes whether the circuit breaker trips at or
+  before the recorded length.
+- **`closure-anomaly`** — reproduce extracts the recorded `closureReason`
+  (selected by the evidence `seq`, since this detector has no fingerprint);
+  re-drive calls the live `buildClosureGuidance(reason)` and asserts the
+  recorded anomalous closure now carries actionable recovery guidance.
+  *Neutralisation here is advisory, not preventive* — the abort still aborts;
+  "neutralised" means the closure no longer surfaces without a recovery path.
+  The guardrail covers `abort` today; an anomalous reason it does not cover yet
+  (e.g. `timeout`) yields no guidance → **fail-closed**, mirroring how the
+  repeat-loop handler fails a sub-threshold loop.
 
 | Outcome | `replay:` checks | Run verdict |
 |---|---|---|
-| Reproduces **and** guardrail neutralises the loop | reproduces ✓, neutralised ✓ | `pass` — fixed |
+| Reproduces **and** guardrail neutralises the recorded failure | reproduces ✓, neutralised ✓ | `pass` — fixed |
 | Reproduces **but** guardrail does not neutralise it | reproduces ✓, neutralised ✗ | `fail` — still reproduces |
 | Fixture intact but no longer encodes the pattern | reproduces – (skipped) | `unsupported` — **never `pass`**; the card-specific behaviour was not verified, even though the guardrail contract passed |
 | Fixture missing / sha256 mismatch | (no replay) | `fail` via `fixture-integrity` before any replay runs |
 
 #### Boundary (no overclaiming)
 
-The replay re-drives the recorded loop **shape** (tool name + consecutive
-byte-identical count), not the original tool or LLM execution. It proves the
-live guardrail covers the recorded failure; it does not prove the loop can
-never arise for an unrelated reason. The fixture bytes are checksum-pinned by
-the eval-case, so the stimulus is stable across runs and the whole check is
-deterministic.
+The replay re-drives the recorded failure **shape** (the loop's tool +
+consecutive byte-identical count, or the closure reason), not the original tool
+or LLM execution. It proves the live guardrail covers the recorded failure; it
+does not prove the failure can never arise for an unrelated reason. The fixture
+bytes are checksum-pinned by the eval-case, so the stimulus is stable across
+runs and the whole check is deterministic.
 
 ## Status precedence
 
--- a/src/improve/eval-run/__fixtures__/closure-anomaly-abort.fixture.jsonl
+++ b/src/improve/eval-run/__fixtures__/closure-anomaly-abort.fixture.jsonl
@@ -1,0 +1,3 @@
+{"ts":"2026-06-20T10:00:00.000Z","seq":0,"kind":"tool_call","payload":{"phase":"started","toolUseId":"tu-1","name":"bash","inputBytes":80}}
+{"ts":"2026-06-20T10:00:00.050Z","seq":1,"kind":"tool_call","payload":{"phase":"completed","toolUseId":"tu-1","name":"bash","resultBytes":256,"isError":false,"truncated":false,"durationMs":50}}
+{"ts":"2026-06-20T10:00:01.000Z","seq":2,"kind":"closure","payload":{"reason":"abort","finalTurnCount":3,"finalCostUsd":0.0123,"finalTokens":{"input":1200,"output":340}}}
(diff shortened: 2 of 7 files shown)


griffinwork40:
All five load-bearing claims **verified against the reviewed ref** `b16f563`:

1. ✅ `schemas.ts:387` — `evidenceEventIndices: z.array(...).min(1)` → empty array is schema-rejected, so `Math.max(...)` → `-Infinity` is **unreachable** from a valid eval-case.
2. ✅ `closure-guidance.ts:47-51` — `switch(reason){ case 'abort': return HINT; default: return null }` → only `abort` covered; everything else (incl. `timeout`, `model_end_turn`, `''`) → null → fail-closed. Cast `reason as ClosureReason` is default-safe, no throw.
3. ✅ `closure-anomaly.ts:149-157` — `detail` carries `closureReason` and `seqs`; `ANOMALOUS_REASONS` excludes `model_end_turn` (line 50-51). Matches every test assumption.
4. ✅ `REPLAY_CHECK_NEUTRALIZED` — referenced only inside `replay.ts` (export, repeat-loop handler, registry); zero production importers elsewhere. Generalization is additive/backward-compatible.
5. ✅ `ClosureReason` type + `ClosureReasonSchema` exist in `trace/index.ts`.

No fabricated downgrades. No critical/high findings → `/shadow-verify` not required (it gates only top-severity claims). **Wave 2 synthesis** (kept local — final judgment):

---

# Code Review — PR #264 `feat(improve): fixture-replay eval-run for closure-anomaly`

**Reviewed ref:** `b16f563` · base `main` ← `feat/improve-eval-run-closure-replay` · 7 files, +509/−37 · regime: full

## Summary
A clean, well-scoped, well-tested feature change. It adds a second fixture-replay handler (`closure-anomaly`) and generalizes the runner's replay detection from a hardcoded check-name to an `isReplayNeutralizeCheck()` predicate. All four verdict-table outcomes (pass / fail-closed / unsupported / integrity-fail) are implemented and tested. The "no schema changes" claim is verified. Two reviewers plus an independent claim-verification pass converged on **no blocking findings**.

## Findings (sorted: medium → low → nit)

**① medium · confidence high · correctness · `replay.ts:418-423` (select-by-seq) · ref:b16f563 · file-state**
Finding: when no detector finding's `seqs` include `wantSeq`, the handler silently falls back to `findings[0]`, and the reproduce check still reports `pass` with a misleading `actual` (`found closure.reason='<fallback>' (seq <wantSeq>)` — pairs the fallback's reason with a seq it was not found at). In the narrow case of a multi-reason fixture with a stale recorded seq, a card whose true recorded reason is *uncovered* could false-`pass` if `findings[0]` happens to be `abort`. The normal single-finding path is unaffected, and the dangerous direction requires a regenerated/multi-reason fixture (schema doc says `evidenceEventIndices` is a verbatim copy of the chosen evidence row, so seq should match under normal generation).
```ts
const bySeq = findings.find((f) => seqsOf(f).includes(wantSeq));
const matched = bySeq ?? findings[0];  // silent fallback — actual still reports pass
```
Suggestion (non-blocking): when `bySeq` is undefined but `findings[0]` is used, annotate the reproduce check's `actual` with `no finding matched seq ${wantSeq}; fell back to findings[0]` so a stale fixture is visible rather than silently replayed.

**② low · confidence high · correctness+api-compat · `runner.ts:194-197` · ref:b16f563 · file-state**
Finding: the runner forwards only `driveLoop` into `ReplayContext`; `buildGuidance` is never forwarded and `RunEvalCaseContext` has no field for it, so the pre-fix (null-guidance) simulation is untestable at the runner level. `replay.test.ts` covers that path directly, so production correctness is unaffected — but the exported `ReplayContext.buildGuidance` seam is test-only by construction, not by type. (Both Wave-1 agents flagged this independently.)
```ts
const probe = await replayHandler.run(evalCase, fixture.bytes, {
  ...(ctx.driveLoop ? { driveLoop: ctx.driveLoop } : {}),
});  // buildGuidance never forwarded
```
Suggestion: either add `buildGuidance?` to `RunEvalCaseContext` and forward it alongside `driveLoop`, or mark `ReplayContext.buildGuidance` with a `@internal`/`@testonly` JSDoc tag.

**③ low · confidence high · test-coverage · `runner.test.ts` · ref:b16f563 · file-state**
Finding: the sha256-tamper (corrupt-bytes) integrity path is tested only for `repeated-tool-use`; for `closure-anomaly` only the missing-fixture path is exercised (`writeFixture: false`). Fixture-integrity is pattern-agnostic so the code path is correct, but a checksum regression on the closure fixture would be caught only by generic logic, not a named regression test.
Suggestion: add one closure-anomaly tamper test mirroring the existing `repeated-tool-use` sha256-mismatch test.

**④ nit · confidence high · correctness · `replay.ts:408` cast · ref:b16f563 · file-state**
Finding: `reason as ClosureReason` widens an arbitrary detector string to the union before `buildClosureGuidance`. Verified safe — the guardrail switch is fully `default`-covered (`'' → null → fail-closed`), no throw. Informational only.
Suggestion: optional — guard with `ClosureReasonSchema` membership before the call to drop the cast.

**⑤ nit · confidence high · security · `replay.ts:474` snapshot · ref:b16f563 · file-state**
Finding: `snapshot(guidance)` writes the recovery hint to the artifact `actual`. Today `CLOSURE_ABORT_RECOVERY_HINT` is a static constant with no session data — no leak. Forward-looking note only.
Suggestion: document in `closure-guidance.ts` that guidance strings must stay free of session-specific data (sessionId/cost/tokens) to keep snapshots artifact-safe.

## Clean confirmations
- **api-compat:** `REPLAY_CHECK_NEUTRALIZED` still exported + used by the repeat-loop handler; `isReplayNeutralizeCheck`, `REPLAY_CHECK_CLOSURE_*`, and `ReplayContext.buildGuidance` are purely additive; registry `.find()`-by-`patternId` means the new entry cannot alter the `repeated-tool-use` path. No breaking change.
- **spec-compliance:** all four verdict-table rows implemented and tested (pass `runner.test.ts:521+`, fail-closed `:536`, unsupported `:550`, integrity-fail inherited). "NO SCHEMA CHANGES" verified — `schemas.ts` untouched; `contracts.ts` diff is doc-comment-only. No scope creep.
- **correctness:** whitespace-only guidance correctly treated as not-neutralized; skipped path yields `unsupported` (never `pass`) via `replayInconclusive`; `''`-reason path fails closed.
- **perf-observability:** offline eval-run, not a hot path; 3-line fixture parsed once per run. No concern.

## Conflicts
None. The two same-locus findings (`Math.max`/`findings[0]` and `buildGuidance` forwarding) are cross-agent *agreements*, merged above.

## What was not checked
- Wave 1 agents read against branch HEAD `b16f563`; I independently re-derived the 5 load-bearing claims at that same ref via `git show`/`git grep` — citations verified at `b16f563`.
- Stated intent: PR #264 title + body — **spec-compliance assessed** against it.
- Did **not** execute the suite or lint — PR claims `pnpm lint` clean and 320/320 `src/improve/` tests pass; not independently run here.
- Reviewed the 7 changed files + 4 referenced production deps (`closure-guidance.ts`, `closure-anomaly.ts`, `trace/index.ts`, `schemas.ts`); did not audit the broader `improve` pipeline or whether the other six closure reasons get wired downstream (out of scope for this diff).

## Merge decision: **MERGE**
No critical, high, or blocking findings. One non-blocking medium (silent `findings[0]` fallback — worth a one-line `actual` annotation in a follow-up), plus minor low/nit polish. The feature matches its stated intent, fails closed correctly, and covers all four verdict outcomes with tests.

---

**Done**
- Reviewed PR #264 (`feat(improve): fixture-replay eval-run for closure-anomaly`) read-only at reviewed ref `b16f563b58ec97e4fcecb090a8b4e1ae8c9637de`: triage → 2 parallel Wave-1 dimension agents (security·api-compat; correctness·spec·test·perf) → independent inline verification of 5 load-bearing claims → local synthesis.
- Evidence: verification commands run against ref `b16f563` (`git show src/improve/schemas.ts:387` = `.min(1)`; `closure-guidance.ts:47-51` switch; `closure-anomaly.ts:149-157` detail keys; `git grep REPLAY_CHECK_NEUTRALIZED` = replay.ts-only). Diff artifact: `/Users/griffinlong/.afk/state/skill-preflight/6a87ff31-0681-4442-b215-021b7d4d360f/pr-264.diff`.
- Verdict: **MERGE** — 0 blocking, 1 medium (non-blocking), 2 low, 2 nit; spec met; "no schema changes" confirmed.
- No world state changed: read-only review, no files/PR/git mutated; the only working-tree dirt is the pre-existing unrelated untracked `chatgpt-design-skill-chat.txt`, left untouched.
- Deferred (for the author, optional): annotate the seq-mismatch fallback in the reproduce check's `actual`; forward `buildGuidance` from `RunEvalCaseContext` or tag it `@internal`; add a closure-anomaly sha256-tamper test.

---

PR #266, opened 2026-06-23
feat(improve): classification eval for tool-failure-density contract

griffinwork40:
## Summary
Upgrades the `tool-failure-density` eval-run contract from a registry-flag check ("is the detector enabled by default?") into a **synthetic-corpus classification eval** that exercises the live `detectToolFailureDensity`.

This is the smallest safe slice from a read-only recon of the remaining supported improve patterns. `tool-failure-density` has **no runtime guardrail** to re-drive a recorded failure through (unlike `repeated-tool-use`'s circuit breaker or `closure-anomaly`'s recovery-hint builder), so it gets **no fixture-replay** — what *can* silently regress is the detector's classification math, and that is what this pins.

## What changed (2 files)
- `src/improve/eval-run/contracts.ts` — append a classification probe to the existing `tool-failure-density` contract (stable id `tool-failure-density-enabled`, so existing eval-run artifacts keep resolving). Builds a hand-verifiable multi-session `SessionRead[]` via the real `parseTraceContent` (schema-faithful, mirrors `replay.ts`) and runs the live detector.
- `src/improve/eval-run/contracts.test.ts` — assert the contract's 8 checks pass + a field-granular `describe` over the same corpus.

## Assertions
- excluded "system said no" classes (policy-refusal / permission-denied / hook-block / abort / elicitation-declined) count toward **neither** numerator nor denominator, and never manufacture a card on their own (a refusal-only tool with a raw 5/6 error rate yields **no card**)
- `circuitBreaker`-synthesised completions are excluded *before* classification
- `timeout` + unclassified failures **are** counted
- the dual count+rate threshold is enforced — both must clear (verified by raising each floor independently)
- emitted `failureCount` / `totalCalls` / `failureRate` / `failureClassBreakdown` / `excludedByClass` / `affectedSessionCount` / `sessionIds` / `seqs` all match

## Also
- Flags the `subagent-block` contract as a **known detector <-> guardrail mismapping** (comment in `contracts.ts` + a clarifying test `describe` name only; behavior unchanged). The detector fires on `hook_decision` / SubagentStart / `block` events, but the contract validates the skill max-depth refusal — a different mechanism that does not emit that event. Resolving the mapping is deliberately deferred.

## Scope / constraints honored
- No fixture-replay for `tool-failure-density`. No schema change. No detector / runner / replay source changes. No LLM reviewer, apply, auto-triage, scheduling, or forge bridge. Limited to the two files above.

## Testing
- `vitest run src/improve/eval-run/ src/improve/scan/detectors/tool-failure-density.test.ts` → **106/106 pass** (contracts 21, runner 28, replay 16, detector 41).
- `pnpm lint` (`tsc --noEmit`, strict) → clean.

### DIFF
--- a/src/improve/eval-run/contracts.test.ts
+++ b/src/improve/eval-run/contracts.test.ts
@@ -9,6 +9,7 @@
 
 import { describe, expect, it } from 'vitest';
 import {
+  buildToolFailureClassificationCorpus,
   knownContractIds,
   makeCheck,
   resolveContract,
@@ -17,6 +18,7 @@ import {
 } from './contracts.js';
 import { REPEAT_CIRCUIT_BREAKER_THRESHOLD } from '../../agent/tools/dispatcher.js';
 import { SKILL_MAX_DEPTH_RECOVERY_HINT } from '../../agent/tools/skill-depth-message.js';
+import { detectToolFailureDensity } from '../scan/detectors/tool-failure-density.js';
 import type { FailurePattern } from '../schemas.js';
 
 // ---------------------------------------------------------------------------
@@ -112,9 +114,18 @@ describe('repeat-loop-circuit-breaker contract', () => {
 
 // ---------------------------------------------------------------------------
 // Contract: skill max-depth recovery hint
+//
+// KNOWN MISMAPPING: this contract is registered under patternId
+// `subagent-block`, but it validates the skill MAX-DEPTH refusal — a different
+// mechanism than the `hook_decision`/SubagentStart/`block` events the
+// subagent-block detector actually fires on (see the note in contracts.ts).
+// The checks below still pass (the refusal builder is healthy); they just do
+// not prove anything about the detector's recorded failures. Do NOT add a
+// fixture-replay for subagent-block against this contract until the mapping is
+// resolved.
 // ---------------------------------------------------------------------------
 
-describe('skill-max-depth-recovery-hint contract', () => {
+describe('skill-max-depth-recovery-hint contract (KNOWN subagent-block mismapping)', () => {
   it('all checks pass against the live refusal builder', async () => {
     const result = await resolveContract('subagent-block')!.run();
     expect(result.checks.map((c) => c.name)).toEqual([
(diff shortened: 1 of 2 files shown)


griffinwork40:
Independent backstop complete — I personally verified every load-bearing claim the PR rests on, reading the live detector at `f714332`:

- **circuitBreaker excluded BEFORE the class check** — `tool-failure-density.ts:163` (`if (ev.payload.circuitBreaker === true) continue;`) fires before the exclusion check at `:171` and before `totalCalls += 1` at `:176`. ✓
- **Excluded classes leave both numerator and denominator** — `:171-174` `continue`s before `:176`. ✓ `timeout`/unclassified fall through and count. ✓
- **Dual threshold, both must clear** — `:196` (`failures.length < minFailures`) AND `:198` (`rate < minFailureRate`). ✓
- **Detail fields** — `:256-269` emits every field the test asserts; `seqs: stats.failures.map(f => f.seq)` at `:268` is push-order = deterministic iteration order → `[1,2,0]`. ✓
- **EXCLUDED set / defaults** — `:84-90`, `:93`, `:100`. ✓

The correctness agent's "CONFIRMED" holds under independent re-derivation. No blocking/critical/high findings exist, so the skill's Wave 1.5 and `/shadow-verify` are not triggered; I ran the equivalent backstop read anyway because the PR's entire value rests on that one claim. Synthesis is small and conflict-free (4 non-overlapping low-tier findings, two agents agree), so I synthesize locally rather than dispatch a Wave 2 agent.

---

# Code Review — PR #266

**`feat(improve): classification eval for tool-failure-density contract`** · `feat/improve-eval-run-tool-failure-classification` → `main` · ref `f714332` · +414/-7, 2 files · regime: full

## Triage
A test-only change. Upgrades the `tool-failure-density` eval-run contract from a registry-flag check into a synthetic-corpus classification eval that drives the **live** `detectToolFailureDensity`, plus a field-granular test `describe`. Also flags the `subagent-block` contract as a known detector↔guardrail mismapping (comment + test describe-name only). No detector/runner/replay source change, no schema change.

## Verdict on the core claim
The synthetic corpus produces **exactly** the detector output the PR asserts — verified twice (correctness agent + my independent source read). `flaky_tool` → 1 card, `failureCount=3`, `totalCalls=5`, `failureRate=0.6`, `excludedByClass`/`failureClassBreakdown`/`seqs`/`sessionIds` all match; `refusal_only_tool` (raw 5/6 error rate, all excluded classes) → **no card** (the false-positive guard). This is a genuine regression net for the detector's classification math, not a tautology.

## Findings — all advisory (no blockers)

**Low · high-confidence · test-coverage · `contracts.test.ts` (dual-threshold block) · ref:f714332 · file-state**
No test pins the **inclusive boundary** at the exact defaults `(minFailures=3, minFailureRate=0.25)`. The detector uses `<` comparisons (`tool-failure-density.ts:196,198`), so a regression flipping `<`→`<=` (or vice-versa) at the boundary would go uncaught — the corpus clears both floors comfortably (3, 0.6). *Suggestion:* add `detectToolFailureDensity(corpus, { minFailures: 3, minFailureRate: 0.25 })` and assert the card is still present.

**Low · high-confidence · test-coverage · `contracts.test.ts` · ref:f714332 · file-state**
`elicitation-declined`'s *card-suppression* effect is never exercised in isolation. `refusal_only_tool` (sessC) uses only 4 of the 5 excluded classes, so if `elicitation-declined` regressed out of `EXCLUDED_FAILURE_CLASSES`, the refusal-only guard would still pass. It is covered indirectly via the `flaky_tool` `excludedByClass` count, and a literal typo would be a `tsc` error (`ToolFailureClass` is a typed union) — so the real exposure is narrow. *Suggestion:* add an `elicitation_only_tool` session mirroring the `refusal_only_tool` guard.

**Nit · high-confidence · style · `contracts.ts` (`runToolFailureDensityClassification`) · ref:f714332 · file-state**
The classification probe is a synchronous `function` wrapped by an `async runToolFailureDensityContract`. No defect — the result merges correctly — but it diverges from the `EvalContract.run` "async function run()" convention every other contract follows. *Suggestion:* mark it `async`, or add a one-line note that it is intentionally synchronous (pure, no I/O).

**Nit (downgraded from the correctness agent's Medium) · style · `contracts.ts` classification-reports check · ref:f714332 · file-state**
The agent flagged the exact-order `seqs` assertion (`toEqual([1,2,0])`) as "order-sensitive without a determinism comment" and suggested sorting before comparison. I'm **correcting and downgrading** this: (a) the test *does* document the order — `contracts.test.ts` carries the inline comment `// Failure seqs in iteration order: sess-a seq1, sess-a seq2, sess-b seq0`, the test name is `…deterministically`, and the check description says `…in deterministic order`; (b) the order is genuinely deterministic (`seqs` is `failures.map(f=>f.seq)` over insertion-ordered arrays, `tool-failure-density.ts:268`); (c) the suggested fix (sort `seqs`) would **weaken** the test by hiding a real detector-ordering regression. The only legitimate kernel is cosmetic: the `contracts.ts` check documents determinism in a description string rather than an `// Invariant:`-prefixed comment.

## What was NOT checked
- **Refs:** Wave 1 agents and I read against branch HEAD `f714332` — citations verified at that ref. Working tree matches HEAD for all PR files (only an unrelated untracked `chatgpt-design-skill-chat.txt` is dirty).
- **Stated intent:** PR #266 title+body — **spec-compliance assessed and met**; no scope creep (the `subagent-block` change is comment + describe-name only, behavior unchanged, as claimed).
- Did **not** run the test suite myself; the PR claims 106/106 pass, and I verified the detector logic the assertions depend on rather than the live run.
- Did **not** check consumers outside this repo (forks/plugins importing `contracts.ts` directly); within this repo the production surface is clean (zero production importers for every renamed/changed symbol).
- Did **not** review the detector's severity ladder (`tool-failure-density.ts:273+`) — the corpus does not assert on card severity.

## Merge decision: **MERGE**

The change does exactly what it states, the synthetic corpus is a faithful regression net for the live detector (independently confirmed at field granularity), the API surface is net-additive with the contract id held stable, and the security surface is empty. The four findings are optional test-hardening and cosmetics — none gate merge.

---

**Done**
- Reviewed PR #266 (`f714332`) in full-regime: 2 parallel Wave-1 agents (security·api-compat; correctness·spec·test-coverage·perf) + an independent coordinator backstop read of `src/improve/scan/detectors/tool-failure-density.ts:60-271`.
- Evidence: this review report (above); load-bearing claims cited to `tool-failure-density.ts:84-90/93/100/163/171-176/196-198/256-269` and the diff at `/Users/griffinlong/.afk/state/skill-preflight/f826a793-3928-4323-89cd-dd8ad0c2541c/pr-266.diff`.
- Outcome: **MERGE** — no blocking/critical/high findings; 2 Low (test-coverage hardening) + 2 Nit (style), all advisory.
- World unchanged: read-only review, no files/commits/PR mutations; the only dirty path (`chatgpt-design-skill-chat.txt`) is unrelated and untouched.
- Deferred by design (the PR's own scope): the `subagent-block` detector↔guardrail mismapping remains unresolved (documented in-code); not a regression introduced by this PR.

---

PR #267, opened 2026-06-24
fix(telegram): render markdown in daemon task notifications

griffinwork40:
## Problem

Daemon task-completion notifications pushed to Telegram render **raw markdown** — `**bold**` shows its asterisks, `` `code` `` shows its backticks, headers keep their `#`. (Reported via `/diagnose`; screenshot showed a `/forge-friction` daemon run rendered as literal markup.)

## Root cause

The out-of-band push path never tells Telegram to parse markdown and never converts it:

`daemon.ts onTaskComplete` → `pushIfConfigured(text)` → `push()` → Telegram `sendMessage`

`push()` only attaches `parse_mode` when the caller passes one (`src/telegram/push.ts:72`), and the daemon caller passed nothing (`src/cli/commands/daemon.ts:420`). With **no `parse_mode`, Telegram renders text verbatim**.

This is a **path asymmetry**, not a formatter bug: the interactive REPL→Telegram streaming path already converts via `markdownToTelegramHtml()` + `parse_mode: 'HTML'` with a plain-text fallback (`streaming.ts` `deliverClean`, lines 149–163). The raw `push()` primitive skipped both steps. `markdownToTelegramHtml()` itself works correctly.

## Fix

- **`src/telegram/push.ts`** — add `pushMarkdown()`: render → `parse_mode: 'HTML'` → fall back to plain text **only** on Telegram's `can't parse entities` 400 (so rate-limit/403/network failures don't double-send). Add a `markdown: true` opt-in to `pushIfConfigured()` that **splits the raw markdown first, then renders each chunk** so chunking can't sever an HTML tag. Mirrors the proven `deliverClean` resilience.
- **`src/cli/commands/daemon.ts`** — task-completion push opts into `{ markdown: true }`. This is the reported symptom.

`markdown` is opt-in and additive — existing `pushIfConfigured` callers (OAuth, review-post, afk-push, digest) are unchanged, so URLs / pre-formatted text stay raw.

## Out of scope (same defect, intentionally not touched)

The `send_telegram` tool (`send-telegram.ts:71`) and the daemon crash-push (`daemon.ts:360`) also call `push()` raw. Left unchanged to scope this PR to the reported daemon symptom; each is a one-line opt-in away.

## Verification

- `pnpm lint` (strict `tsc --noEmit`): clean
- `src/telegram/push.test.ts`: 24 pass (4 new — HTML render, parse-error fallback, no-double-send-on-403, `pushIfConfigured` markdown path)
- `src/cli/commands/daemon.test.ts`: 24 pass (completion assertion updated for the new arg)
- Full suite: **537/538 files pass**; the single failure (`anthropic-direct.test.ts > compact()`, a `claude-haiku-4-5` model-alias drift) is pre-existing and unrelated to this change.

### DIFF
--- a/src/cli/commands/daemon.test.ts
+++ b/src/cli/commands/daemon.test.ts
@@ -306,6 +306,7 @@ describe('afk daemon (CLI integration)', () => {
 
     expect(mockPushIfConfigured).toHaveBeenCalledWith(
       expect.stringContaining('full daemon output'),
+      { markdown: true },
     );
   });
 });
--- a/src/cli/commands/daemon.ts
+++ b/src/cli/commands/daemon.ts
@@ -417,7 +417,10 @@ export function registerDaemonCommand(program: Command): void {
           ...(trigger === 'pull' ? { pullPollIntervalMs: 30_000, queueDir: getQueueDir() } : {}),
           tasks,
           onTaskComplete: (record: TelemetryRecord, details?: TaskCompletionDetails) => {
-            void pushIfConfigured(formatTaskCompletion(record, details)).catch(() => undefined);
+            // markdown:true — task output is agent-authored markdown; render it
+            // to Telegram HTML so **bold**/`code`/headers format instead of
+            // showing their literal markers (plain-text fallback on parse error).
+            void pushIfConfigured(formatTaskCompletion(record, details), { markdown: true }).catch(() => undefined);
           },
         });
 
--- a/src/telegram/push.test.ts
+++ b/src/telegram/push.test.ts
@@ -214,6 +214,29 @@ describe('pushMarkdown', () => {
     // No duplicate send — only the HTML attempt was made.
     expect(fetchImpl).toHaveBeenCalledTimes(1);
   });
+
+  test('re-splits rendered HTML so escaping-expanded content is never truncated past 4096', async () => {
+    const fetchImpl = makeFetchOk();
+    // 3000 '<' chars escape to 3000 '&lt;' = 12000 chars of HTML — ~3x Telegram's
+    // 4096 limit. Without the inner re-split, push() truncates to 4096 and the
+    // remaining ~8000 chars are silently dropped.
+    const raw = '<'.repeat(3000);
+    const result = await pushMarkdown({ token: 't', chatId: '1', text: raw, fetchImpl });
+
+    expect(result.ok).toBe(true);
+    const calls = (fetchImpl as unknown as ReturnType<typeof vi.fn>).mock.calls;
+    // Sent as multiple messages rather than one truncated blob.
+    expect(calls.length).toBeGreaterThan(1);
+    let reassembled = '';
+    for (const call of calls) {
+      const body = JSON.parse(call[1].body);
+      expect(body.parse_mode).toBe('HTML');
+      expect(body.text.length).toBeLessThanOrEqual(4096);
+      reassembled += body.text;
+    }
+    // Full payload delivered across the chunks — no silent truncation.
+    expect(reassembled).toBe('&lt;'.repeat(3000));
+  });
 });
 
 describe('pushIfConfigured', () => {
--- a/src/telegram/push.test.ts
+++ b/src/telegram/push.test.ts
@@ -3,7 +3,7 @@
  */
 
 import { describe, test, expect, beforeEach, afterEach, vi } from 'vitest';
-import { push, pushIfConfigured } from './push';
+import { push, pushIfConfigured, pushMarkdown } from './push';
 
 // loadTelegramConfig reads afk.config.json; mock it so these tests are hermetic
 // (notify routing is then driven purely by the env vars set per-test). The pure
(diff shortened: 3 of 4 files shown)


griffinwork40:
Verification complete. I've independently confirmed the load-bearing facts at ref `0f007b4` (inline Wave 1.5):

- **Citation (Check A) — VERIFIED**: `push.ts:70` does `text: options.text.slice(0, 4096)` (silent truncation); `formatter.ts:9,19` sets `MAX_MESSAGE_LENGTH = 4096` and `splitLongMessage` defaults to that limit; `markdownToTelegramHtml` escapes `&`/`<`/`>` and injects tags (all expand length); `streaming.ts:153` `deliverClean` does the inner `splitLongMessage(markdownToTelegramHtml(chunk))` re-split that `pushMarkdown` omits.
- **api-compat (Check B) — confirmed + one addition**: all named callers (`oauth.ts:409`, `review-post.ts:291`, `daemon.ts:360`) pass no opts → unaffected; the new `markdown?` field is additive. Agent missed one caller, `skills/score/digest.ts` — also opts-free, also unaffected, so the verdict holds.
- `pushMarkdown` has no production caller other than `pushIfConfigured` (post-split). Finding stands at the chunk level.

No critical/high findings → shadow-verify not required (and I've already re-derived the one substantive finding from source directly). Emitting synthesis.

---

# Code Review — PR #267

**`fix(telegram): render markdown in daemon task notifications`** · `fix/telegram-daemon-markdown-render` → `main`
**Reviewed ref:** `0f007b4df38f95c2fd270a8c3d1115117435bc2e` · **Regime:** light (hotfix, ~127 lines, 4 files) · **Stated intent:** PR #267 title + body — spec-compliance assessed.

## Summary
Adds `pushMarkdown()` (render markdown → Telegram HTML, fall back to plain text only on a `400 can't parse entities`) and a `markdown: true` opt-in on `pushIfConfigured()` that splits raw markdown first then renders per chunk; the daemon `onTaskComplete` push opts in. The change is correct for its core purpose, additive, well-tested, and cleanly scoped. One real edge-case introduced on the modified path.

### 🟡 medium · high confidence · correctness — `src/telegram/push.ts:122–138` (`pushMarkdown`) · ref:0f007b4 · file-state
`pushMarkdown` renders a chunk to HTML and hands it to `push()`, which hard-slices to 4096 (`push.ts:70`). HTML escaping (`&`→`&amp;`, `<`→`&lt;`) and tag injection (`<b>`, `<code>`) expand length, so a raw chunk near the 4096 split limit (`formatter.ts:9` `MAX_MESSAGE_LENGTH = 4096`) renders to >4096 and gets truncated. The claimed "mirror," `deliverClean`, avoids this with an **inner re-split** that `pushMarkdown` omits:
```
// streaming.ts:153 — the mirrored pattern actually does this:
for (const htmlChunk of splitLongMessage(markdownToTelegramHtml(chunk))) {
  if (htmlChunk) await ctx.reply(htmlChunk, { parse_mode: 'HTML' });
}
// push.ts:127 — pushMarkdown does NOT re-split the rendered html:
const htmlResult = await push({ ...options, text: html, parseMode: 'HTML' });
```
Two outcomes when a render exceeds 4096: if the cut yields valid HTML (likely — tags are sparse, prose dominates) Telegram **accepts it and the tail is silently lost**; if malformed, the `can't parse entities` fallback resends the chunk as plain text (full content, no formatting). This is a *newly introduced* regression on this exact path — before the PR, `pushIfConfigured` sent each ≤4096 raw chunk complete (ugly but whole). Reachable from production via `daemon.ts:423` for long task outputs (the test fixture itself asserts `'full daemon output'`).
**Suggestion:** mirror `deliverClean` fully — in `pushMarkdown`, `for (const htmlChunk of splitLongMessage(html)) await push({ ...options, text: htmlChunk, parseMode: 'HTML' })` (collapse results), so post-render chunks are individually ≤4096.

### 🟢 low · high confidence · test-coverage — `src/telegram/push.test.ts` · ref:0f007b4 · file-state
No test exercises the HTML-expansion-past-4096 case (`git grep -n '4096\|expand\|truncat' src/telegram/push.test.ts` → no matches at ref). The 4 new tests cover happy-path HTML, parse-error fallback, no-double-send-on-403, and per-chunk rendering — solid for the normal paths.
**Suggestion:** add a test with a ~4090-char chunk of escapable content; assert the rendered tail is delivered, not dropped.

## Clean dimensions (read at 0f007b4)
- **security** — no new credential paths or I/O surfaces; `markdownToTelegramHtml` escapes `&`/`<`/`>` before tag injection. No issues.
- **api-compat** — new `markdown?: boolean` is additive (default `{}`); `oauth.ts:409`, `review-post.ts:291`, `daemon.ts:360`, `digest.ts` all opts-free and unaffected. No issues.
- **spec-compliance** — all three stated requirements verified in code: `pushMarkdown` render→HTML→fallback-on-parse-400 ✓; `pushIfConfigured` splits raw then renders per chunk ✓; `markdown` opt-in, existing callers untouched ✓; out-of-scope `send_telegram`/crash-push correctly left raw ✓. (The "mirrors deliverClean" claim is *partially* inaccurate — it omits the inner re-split — but that is the correctness finding above, not a spec gap; the spec only requires render→HTML→fallback.) No spec violations.
- **perf-observability** — at most one extra HTTP round-trip on parse failure, gated by the existing 10s `AbortController`; no new loops or polling. No issues.

## Fallback logic — verified correct
`status === 400 && /can't parse entities/i.test(errorMessage)` matches Telegram's real `Bad Request: can't parse entities: …` description (`errorMessage = json.description`, `push.ts:91`); 403/429/network skip the fallback (no double-send), exactly as the tests assert. `markdown` correctly wins over `parseMode` (the `base` object carries no `parseMode`). `replyMarkup` is preserved through both attempts and stays on chunk 0 only.

---

## Merge decision: **MERGE**

This fixes a real, reported user-facing defect (daemon notifications showing literal `**`/`` ` `` markers), is additive with zero API breakage, and is well-tested across its happy and fallback paths. The one medium finding — silent tail-truncation of long (>4096) rendered chunks because `pushMarkdown` skips the inner re-split that `deliverClean` performs — is a genuine, newly-introduced edge case on the modified path and is the one thing worth addressing: ideally fold in the one-line `splitLongMessage(html)` loop (and a covering test) before or immediately after merge. It does not gate merge: short notifications (the common case) are strictly improved, and the malformed-cut sub-case degrades to complete-but-plain rather than lost.

**What was not checked**
- Wave 1 + my Wave 1.5 verification read against branch HEAD `0f007b4` — all blocking-tier citations (`push.ts:70`, `streaming.ts:153`, `formatter.ts:9/19`) verified at that ref; no patch-file gaps.
- Stated intent: PR #267 title + body — spec-compliance assessed.
- Did **not** execute the test suite or runtime-exercise the daemon→Telegram path (static review only); did not measure real-world expansion ratios of `markdownToTelegramHtml` on representative `formatTaskCompletion` output, so the *frequency* of the truncation case in practice is unquantified (mechanism confirmed, incidence not).
- Did not review the intentionally out-of-scope raw-`push()` sites (`send_telegram.ts`, `daemon.ts:360` crash-push) beyond confirming they're unchanged.

This skill is read-only — no files, commits, or PR state were modified. Review complete.

---

PR #269, opened 2026-06-24
fix(cli): protect narrow table columns from ellipsis truncation

griffinwork40:
## What & why

Fixes broken markdown table rendering in the REPL: when a content-heavy table is wider than the terminal, the renderer truncated a narrow column's values to an ellipsis — e.g. a `Verdict` column of `CONFIRMED`/`OVERSTATED` rendered as `CONF…`/`OVER…` — while wide columns kept plenty of room.

### Root cause
`renderMarkdownToTerminal`'s `case 'table'` (`src/cli/formatter.ts`) shrank every column by the **same proportion** under overflow, with a floor of `minColWidth = 1`. A high overflow ratio therefore crushed narrow single-word columns below their own content width, and `truncateDisplayWidth` chopped each value to `…`. (Tall single-column cells — long `Source` citations — also wrapped into stray continuation lines.)

## The fix

**Floor-based water-filling** replaces the proportional squeeze:
- Per-column floor = `min(natural, longestWord, 14)` — the column's incompressible single-token width, capped so one long token (a path/URL) can't starve the rest.
- Leftover budget is distributed by **reducible slack** (`natural − floor`), so the squeeze lands on genuinely wide columns and narrow ones stay readable.
- A degenerate path (floors exceed the budget — too many columns for the width) shrinks floors proportionally, preserving relative sizes.

**Hard line-cap safety net**: every emitted table line is capped to the budget via `truncateDisplayWidth` at the renderer's return, making "no line exceeds `maxWidth`" a guaranteed invariant. This keeps the commit-time second `wrapToWidth` pass (`markdown-stream-format.ts:92`) a provable no-op for tables, closing the latent row-shredding class for many-column tables (chrome-overflow).

## Tests
- New regression test: a narrow single-word column renders its values in full under a tight `maxWidth` (no ellipsis), the table still fits, and the re-wrap stays a no-op.
- Updated the existing ratio test's doc comment to describe the new two-path allocation (it now exercises the degenerate path and still passes).

## Verification
- `pnpm lint` (tsc --noEmit strict): clean.
- `pnpm exec vitest run src/cli/formatter.test.ts`: 66/66 pass.
- Full suite: 10018 passed, 12 skipped, 1 failed — the single failure (`anthropic-direct.test.ts > compact()`, `claude-haiku-4-5` vs `claude-haiku-4-5-20251001`) is **pre-existing and unrelated**: it reproduces on clean `main` (verified by stashing this branch's changes) and this PR touches only `src/cli/formatter.ts` + `src/cli/formatter.test.ts`.

## Notes
- Behavior change is confined to the overflow path; tables that already fit the terminal are unchanged.
- The tall-cell continuation-line sprawl (long columns wrapping to many lines) is reduced as a side effect but is not the target of this PR.

### DIFF
--- a/src/cli/formatter.test.ts
+++ b/src/cli/formatter.test.ts
@@ -102,20 +102,16 @@ describe('renderMarkdownToTerminal', () => {
     });
 
     /**
-     * Proportional column-shrink contract.
+     * Wide:narrow ratio under the degenerate squeeze path.
      *
-     * The pre-edit algorithm shrunk the widest column by 1 each loop, which
-     * tended to equalize columns when one was much wider than the other —
-     * destroying the ratio the table author chose. The current algorithm
-     * shrinks every column proportional to its own reducible width, so the
-     * wide column stays meaningfully wider than the narrow column even
-     * under tight maxWidth constraints.
-     *
-     * Concrete check: with widths [5, 15] and a maxWidth that forces ~9
-     * units of shrink, the new algorithm preserves col2 > 2x col1.
-     * The old algorithm would have flattened them to roughly equal.
+     * Both columns are single, unbreakable tokens (5 and 15 chars), so their
+     * incompressible floors (min(natural, longestWord, CAP)) sum to more than
+     * the content budget at maxWidth 18 — the degenerate path. It shrinks the
+     * floors proportionally, which keeps the wide column meaningfully wider than
+     * the narrow one instead of flattening both to roughly equal (the original
+     * trim-widest-by-1 failure mode).
      */
-    it('preserves wide:narrow ratio under proportional shrink', () => {
+    it('preserves wide:narrow ratio under the degenerate squeeze', () => {
       const sample = [
         '| Short | WideHeader12345 |',
         '|-------|-----------------|',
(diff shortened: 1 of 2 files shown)


griffinwork40:
All gates green:
- **`pnpm lint`** (tsc --noEmit, strict mode with `noUncheckedIndexedAccess`): exit 0 — every new indexed access is properly `?? `-guarded.
- **Isolated `formatter.test.ts`**: 65/65 passed.
- **Full suite**: only the unrelated pre-existing `anthropic-direct.test.ts` failure.

I have everything needed to render the verdict. This was a read-only review — no files were modified, nothing committed, no PR comment posted.

---

# Review: PR #269 — fix(cli): protect narrow table columns from ellipsis truncation

**Recommendation: ✅ APPROVE (merge-ready).** Clean, well-tested, correctly-reasoned fix. Caveats below are all minor/doc-level and non-blocking.

## What it does
Replaces the table column allocator's **uniform proportional shrink** (scale every column by the same factor) with **floor-based water-filling**: each column gets a floor of `min(natural, longestUnbreakableWord, 14)`, the leftover budget is distributed in proportion to each column's reducible slack, and a return-time `truncateDisplayWidth` line cap guarantees no row exceeds the budget. This stops a narrow single-word column (the "Verdict: CONFIRMED/OVERSTATED" case) from being crushed to `CONF…` when a sibling column is very wide.

## Findings by dimension

**Correctness — SOUND** (independently re-derived, not just trusting the comments)
- Termination of the remainder-distribution loop is provable: max `Math.floor` remainder is `colCount−1`, and `leftover < slackTotal` always holds inside the squeeze branch, so every slack column retains `+1` headroom — the `colCount*4` guard is comfortably sufficient.
- Division safety: `/slackTotal` is gated by `slackTotal > 0`; `/floorTotal` in the degenerate path is unreachable when `floorTotal === 0` (that requires all-zero widths, which can't enter the squeeze block). No NaN/Infinity can reach `widths`.
- Normal path never over-allocates (sum ≤ availableContentWidth) and never pads a column past its natural content width.

**Security — no concern.** `wordWidth` splits on `/\s+/` (no ReDoS surface). The new line cap routes through `truncateDisplayWidth`, which is ANSI/OSC8-aware (re-emits `\x1b[0m`, re-closes hyperlink spans) — it *hardens* against escape leakage rather than introducing it.

**API compat — safe.** Signature unchanged. All consumers (`markdown-stream-format.ts`, `stream-renderer-orchestrator-emit.ts`, `transcript.ts`, `chat.ts`) pass text through without asserting widths; no snapshot test broke. (`terminal-compositor.overflow-gap.test.ts:75` passes `{ width: COLS }` instead of `{ maxWidth }` — a pre-existing ignored-option typo, not touched by this PR.)

**Test coverage — strong.** The new test reproduces the actual bug (4-col table, narrow Verdict column, `maxWidth: 70`) and asserts three things: full words survive, no line exceeds budget, **and** the downstream `wrapToWidth` re-wrap stays a literal no-op (real integration assertion against the actual function on ANSI-included output). The recharacterized degenerate-ratio test genuinely exercises the new degenerate branch.

**Perf — negligible.** `wordWidth` adds one same-order pass at width-computation time; the line cap adds one short-circuiting `stringWidth` per emitted row on the happy path. Table rendering is per-commit, not per-frame.

## Non-blocking items for the author
- **[minor / doc]** The `// Invariant:` comment at `formatter.ts:361` ("after this block sum(widths) ≤ availableContentWidth") is literally false in the degenerate sub-case where `availableContentWidth < colCount`: every column is floored at `Math.max(1, …)`, so the sum is `colCount`, and the trim loop breaks immediately (`widest === -1`, nothing is `> 1`). The real guarantee there is the return-time line cap, not the column math. Given the repo's comment-prefix convention treats `Invariant:` as load-bearing, consider tightening the wording to note the `availableContentWidth < colCount` exception so a future reader doesn't over-trust it. (Functionally fine — the safety net covers it.)
- **[minor / cosmetic]** The degenerate path has no grow-back, so it can leave budget unused (e.g. widths `[100,3,3,3,3]` @ `maxWidth 30` → sum 11 of 14 available). Table renders narrower than strictly necessary in pathological mixes. Acceptable.
- **[nit]** Optional: add a test for the `availableContentWidth ≤ 0` / "chrome alone exceeds width" case to lock the safety-net behavior the comment promises — it's currently the one path with no direct regression test.

## Terminal state — Done
- **What was reviewed:** PR #269 diff (`src/cli/formatter.ts` +88/−29, `src/cli/formatter.test.ts` +46/−14) against the working-tree state at HEAD `8c65fa3`.
- **Evidence:** `pnpm lint` exit 0; isolated `formatter.test.ts` 65/65 pass; full suite 10018 pass / 1 unrelated pre-existing fail (`anthropic-direct.test.ts` compact model-alias) / 12 skip; independent algorithm-correctness subagent verdict SOUND-WITH-CAVEATS; `display.ts:107-145` read confirms the safety net is ANSI/OSC8-safe; consumer grep confirms no width-asserting downstream broke.
- **World change:** none — read-only review, no files modified, no commit, no PR comment.
- **Verdict:** APPROVE. Three non-blocking suggestions surfaced above (one doc-comment overclaim, one cosmetic degenerate-path under-allocation, one optional missing test); none gate merge.

---

PR #270, opened 2026-06-24
fix(cli): fill width budget in degenerate table squeeze

griffinwork40:
Follow-up to #269 (review-surfaced cosmetic nit).

## Problem
The table column allocator's **degenerate path** — where the per-column floors already exceed the content budget (`floorTotal > availableContentWidth`) — scales the floors down with `Math.floor(...)`, which discards fractional units. There is a trim-**down** loop for the over-budget case but no grow-**back** for the under-budget case, so the rendered table can land several columns narrower than `maxWidth` allows.

Concrete case: natural widths `[37, 3, 3, 3, 3]` at `maxWidth 30` → `chromeWidth 16`, `availableContentWidth 14`, floors `[14,3,3,3,3]` (degenerate), `scale ≈ 0.538` → constrained `[7,1,1,1,1]` = **11 of 14** used. The table border renders 27 wide instead of 30.

## Fix
Add a grow-back loop after the existing trim-down pass in the degenerate branch (`src/cli/formatter.ts`): hand the reclaimed slack to the widest columns first (by natural width), **capped at each column's natural width** so no column is padded past its content, until the budget is met.

- Bounded: the post-floor deficit is `< colCount`, so one pass over the order suffices (`grow < colCount * 4` guard, consistent with the normal-path remainder loop).
- Never exceeds `availableContentWidth` (the `while` condition caps it) and never pads beyond natural content width.
- `growOrder` uses a secondary sort key (column index) so tie-breaks on equal-natural-width columns are deterministic across JS engines.

## Verification
- `tsc --noEmit` (strict lint gate): clean.
- `vitest run src/cli/formatter.test.ts`: 67/67 pass.
- New regression test `fills the width budget in the degenerate squeeze (grow-back)` is a **genuine guard** — with the grow-back loop commented out it fails (`widest = 27`), and passes with it (`= 30`).

## Out of scope
The separate `// Invariant:` doc-comment overclaim flagged in the #269 review (the comment says `sum(widths) <= availableContentWidth` always, which is technically untrue when `availableContentWidth < colCount` — the return-time line cap is the real backstop there) is **not** addressed here; it is a doc-only nit on a different sub-path and warrants its own small change if desired.

### DIFF
--- a/src/cli/formatter.test.ts
+++ b/src/cli/formatter.test.ts
@@ -133,6 +133,35 @@ describe('renderMarkdownToTerminal', () => {
     });
 
     /**
+     * Degenerate path fills the width budget (grow-back).
+     *
+     * When the floors exceed the content budget the allocator scales them down
+     * with Math.floor, discarding fractional units. Without a grow-back step the
+     * table renders narrower than maxWidth allows. One wide column plus several
+     * narrow ones at a tight maxWidth is the concrete case: pre-fix the rendered
+     * border was ~3 columns short of the budget (27 of 30); grow-back hands the
+     * reclaimed slack to the widest column so the table fills the budget exactly.
+     */
+    it('fills the width budget in the degenerate squeeze (grow-back)', () => {
+      const sample = [
+        '| Path | One | Two | Thr | Fou |',
+        '|------|-----|-----|-----|-----|',
+        '| src/cli/formatter.ts:340-the-allocator | abc | def | ghi | jkl |',
+        '',
+      ].join('\n');
+      const maxWidth = 30;
+      const out = stripAnsi(renderMarkdownToTerminal(sample, { maxWidth }));
+      const tableLines = out
+        .split('\n')
+        .filter((l) => /[┌┐└┘├┤┬┴┼│─]/.test(l));
+      const widest = Math.max(...tableLines.map((l) => stringWidth(l)));
+      // Fills (almost) the whole budget — pre-fix it under-allocated to ~27.
+      expect(widest).toBeGreaterThanOrEqual(maxWidth - 1);
+      // ...and never exceeds it.
+      expect(widest).toBeLessThanOrEqual(maxWidth);
+    });
+
+    /**
      * Narrow single-word column protection (the "Verd…" bug).
      *
      * A content-heavy table much wider than the terminal used to shrink EVERY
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

A small, well-documented hotfix (follow-up to #269). Reviewed against the PR's stated intent at branch HEAD `12cc790`. No blocking findings.

### Security
No issues found.

### API compatibility
No issues found. `renderMarkdownToTerminal(text, opts?)` signature is unchanged. Five production importers confirmed (`stream-renderer-orchestrator-emit.ts`, `attach.ts`, `transcript.ts`, `markdown-stream-format.ts`, `chat.ts`); none are affected.

### Correctness — verified clean
- **Termination/bounds:** the new `while (constrainedTotal < availableContentWidth && grow < colCount * 4)` loop hard-caps the total at the budget (can never over-allocate) and is bounded by `grow < colCount * 4` — matching the normal-path guard exactly. In the all-columns-saturated edge case every iteration is a no-op skip and the loop exits after `colCount * 4` iterations: wasteful but provably terminating, no infinite loop.
- **Natural-width cap:** the inner `if (constrained[i] < widths[i])` guard prevents any column from being padded past its content width.
- **Trim-down ↔ grow-back interaction:** trim-down exits with `constrainedTotal <= availableContentWidth`; grow-back only fires when strictly `<`. They cannot fight — grow-back is purely additive from a deficit position.

### Spec compliance — all clauses met
Grow-back loop added after trim-down in the degenerate branch only; widest-first ordering by natural width; capped at each column's natural width; bounded consistently with the normal path; never exceeds `availableContentWidth`. The out-of-scope `// Invariant:` doc-comment was correctly left untouched.

### Test coverage
The new regression test (`formatter.test.ts`) exercises exactly the described input (`[37,3,3,3,3]` natural widths via a long first cell at `maxWidth=30`), measures the widest rendered border line, and asserts `>= maxWidth - 1` and `<= maxWidth`. The ±1 tolerance absorbs unicode/ANSI width rounding while still catching the pre-fix 27-wide regression (27 < 29). Prior to this PR no test covered grow-back in the degenerate path; this closes that gap.

### Performance / observability
No issues. Grow-back is O(colCount), bounded at `colCount × 4` iterations, inside a block reached only when the degenerate condition holds. Zero hot-path impact.

---

**Not independently verified:** the downstream return-time line cap (`wrapToWidth`/`truncateDisplayWidth`) that the inline comment cites as the no-overrun backstop. CI on this PR is green (Lint & Build, Test, Docs, Publish Bundle Smoke all passing).

_Read against branch HEAD `12cc790`; citations verified at that ref. This is an automated first-pass review — not an approval._

PR #280, opened 2026-06-25
feat(hooks): add UserPromptSubmit as 7th harness hook event

griffinwork40:
## Summary
Adds `UserPromptSubmit` as the 7th harness hook event, wired end-to-end through the existing hook infrastructure (config loader, matcher, registry, block/abort semantics). Previously the harness enumerated ~25 SDK event names in `sdk-types.ts` and had general shell-hook plumbing, but only dispatched 6 live events. This adds the highest-leverage missing one: an intercept point at the user-input boundary.

## What fires
The hook fires once per genuine user turn in the interactive REPL, after `@`-file expansion and shell-injection prepend, immediately before `runTurn` hands the message to the model. It does NOT fire for slash commands, shell passthroughs (`!`), or framework-injected context.

## Semantics
- `decision: 'block'` -> the turn is cancelled; a brief notice renders in the REPL and the loop continues
- `injectContext: 'X'` -> `X` is prepended to the prompt before the model sees it (reuses the existing `HookDecision.injectContext` field; its doc is updated to cover both SubagentStop and UserPromptSubmit)
- no handler registered -> zero-overhead pass-through
- `AbortError` always propagates; a throwing handler fails closed (blocks the turn)

## Files changed (10, +182/-20)
- `src/agent/hooks.ts` - union member, `UserPromptSubmitContext`, `HookContext` union, `injectContext` JSDoc
- `src/agent/hooks/config-loader.ts` - `UserPromptSubmit` added to both `validEvents` arrays
- `src/agent/hooks/config-bridge.ts` - `validEvents` (matcher ignored for this non-tool event, consistent with SessionStart/End)
- `src/agent/hooks/command-executor.ts` - `prompt` field in the shell-hook stdin payload
- `src/cli/commands/interactive/shared.ts` - optional `hookRegistry` on `InteractiveCtx`
- `src/cli/commands/interactive/bootstrap.ts` - wires `hookRegistry` into ctx
- `src/cli/commands/interactive/loop-iteration.ts` - dispatch call-site (block / inject / allow)
- tests: `user-prompt-submit.test.ts` (new, 7 tests), `hooks.test.ts` (narrowing arm), `loop-iteration.test.ts` (4 integration tests)

## Tests / gates
- `pnpm lint` (tsc --noEmit, strict): exit 0
- scoped vitest (6 hook/loop files): 115 passed, 0 failed
- `pnpm audit:sdk:check`: OK, no new SDK imports
- `pnpm scan:env:check`: docs in sync
- coverage of both `injectContext` branches (inject path + allow-through path) plus block and no-registry paths

Note: one unrelated pre-existing failure on `main` (`anthropic-direct.test.ts` compact() assertion) predates this branch and is untouched here.

## Scope
Frozen to `UserPromptSubmit` only. `Stop`, `PreCompact`, `PostToolUseFailure`, and the Telegram / `afk chat` / daemon surfaces are intentionally deferred to separate follow-up PRs.

### DIFF
--- a/src/agent/hooks.test.ts
+++ b/src/agent/hooks.test.ts
@@ -20,6 +20,7 @@ import type {
   PreToolUseContext,
   SessionStartContext,
   SubagentStopContext,
+  UserPromptSubmitContext,
 } from './hooks.js';
 
 function sessionStartCtx(sessionId = 'sess-1'): SessionStartContext {
@@ -234,6 +235,8 @@ describe('HookContext — discriminated union narrowing', () => {
           return `pre:${ctx.toolName}`;
         case 'PostToolUse':
           return `post:${ctx.toolName}`;
+        case 'UserPromptSubmit':
+          return `ups:${ctx.prompt}`;
       }
     };
 
@@ -242,9 +245,15 @@ describe('HookContext — discriminated union narrowing', () => {
       subagentId: 'sub-1',
       status: 'succeeded',
     };
+    const ups: UserPromptSubmitContext = {
+      event: 'UserPromptSubmit',
+      prompt: 'hello from user',
+      sessionId: 's-42',
+    };
     expect(describe(stop)).toBe('sub-stop:sub-1:succeeded');
     expect(describe(preToolCtx('Edit'))).toBe('pre:Edit');
     expect(describe(sessionStartCtx('s-99'))).toBe('start:s-99');
+    expect(describe(ups)).toBe('ups:hello from user');
   });
 });
 
--- a/src/agent/hooks.ts
+++ b/src/agent/hooks.ts
@@ -18,17 +18,22 @@
  * with the original error attached as `cause`. This prevents a bug in one
  * handler from silently skipping policy enforcement.
  *
- * **Context injection (SubagentStop only):** Foreground subagents hand their
- * final assistant output to the parent through the normal `agent` tool result;
- * `injectContext` is a separate hook-generated framework note, not text typed
- * by the human user. When a `SubagentStop` handler returns `injectContext`, the
- * dispatch result is propagated to the caller, which queues the context string
- * to the parent session's input stream for the parent's next turn. If the
- * parent is aborting, the injection is skipped. DAG/compose and background
- * paths are not guaranteed to inject (some intentionally leave this channel
- * dark), and multiple `injectContext` values do not merge today: hook dispatch
- * returns the last non-blocking decision. Other hook events ignore
- * `injectContext` entirely.
+ * **Context injection (SubagentStop and UserPromptSubmit):** Foreground
+ * subagents hand their final assistant output to the parent through the normal
+ * `agent` tool result; `injectContext` is a separate hook-generated framework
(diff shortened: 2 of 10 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

The feature is wired correctly end-to-end and matches its stated intent. Verified against branch HEAD `7b4e3b62`:
- **Slash commands & shell passthroughs are correctly excluded** — `!`/`/` branches `continue` at `loop-iteration.ts:258/289/292`, all before the `UserPromptSubmit` dispatch at `loop-iteration.ts:399`.
- **injectContext ordering** matches the spec: `injectContext + (shellInjection + originalText)`.
- **Generic fail-closed holds** — a handler that throws a non-timeout error is wrapped in `HookBlockedError` at `hook-registry.ts:195` and surfaces as a REPL notice + `continue`.
- **Test coverage** for the four stated loop paths (block / inject / allow / no-registry) plus registry-dispatch and config-loader is present and reads correctly.

Two non-blocking findings to address as follow-ups (neither blocks merge):

### 1. medium · correctness — timeout sub-case is not fail-closed
`hook-registry.ts:189-194` re-throws `HookHandlerTimeoutError` **raw** (intentionally, so `dispatchSubagentStop` can distinguish it). The new REPL catch only handles `HookBlockedError`/`AbortError`:

```
src/cli/commands/interactive/loop-iteration.ts:415-424
} catch (err) {
  if (err instanceof HookBlockedError) { ...notice...; continue; }
  if (err instanceof AbortError) throw err;
  throw err;   // ← HookHandlerTimeoutError lands here, unwinding the loop
}
```

A `UserPromptSubmit` handler that exceeds the default `HOOK_HANDLER_TIMEOUT_MS` ceiling propagates past this catch and crashes the REPL loop instead of rendering a notice + continuing — so the PR body's "a throwing handler fails closed (blocks the turn)" is only partially true. Reachable once any timing-out prompt hook is registered.
*Fix:* catch `HookHandlerTimeoutError` alongside `HookBlockedError` (emit a timeout notice + `continue`), or wrap it as `HookBlockedError` for this call-site.

### 2. nit · correctness — dead AbortError arm
`src/cli/commands/interactive/loop-iteration.ts:423-424`:

```
if (err instanceof AbortError) throw err;
throw err;
```

Both branches rethrow unconditionally, so the `AbortError` check is structurally dead (correct behavior today, but a maintenance trap).
*Fix:* collapse to a single `throw err;` with a comment noting AbortError + all other non-HookBlocked throws propagate.

### Test gap (low)
No test exercises the `UserPromptSubmit` timeout sub-path in `loop-iteration.test.ts` (confirmed: `git grep HookHandlerTimeoutError` in that file → 0 matches). Worth a regression test alongside finding #1.

---
**What was not checked:** Read against branch HEAD `7b4e3b62` — all blocking/medium citations verified at that ref. Stated intent: PR #280 title + body — spec-compliance assessed and passes. Did not run the test suite locally (CI reports all checks green). Did not review the Telegram / `afk chat` / daemon surfaces (explicitly deferred by the PR's frozen scope). Light-regime review (~202 lines).

PR #285, opened 2026-06-25
feat(plan-mode): model-callable exit_plan_mode tool with elicitation picker

griffinwork40:
## Summary
Adds a model-callable `exit_plan_mode` tool — the model-proposed counterpart to the `/plan off` slash command — so leaving plan mode can be model-initiated + user-confirmed via an elicitation picker rather than a typed slash command.

- When the model judges its plan ready it calls `exit_plan_mode`, which shows a 3-option picker: **approve → implement in default mode** / **approve → implement in bypass mode** / **keep planning**.
- On approval the handler flips the live permission mode and queues the **byte-identical** save-and-implement turn `/plan off` already produces, via a session→REPL seed bridge.
- Registered per-query by **both** providers **only while in plan mode** (mirrors the `get_runtime_state` per-query registration); not in `WRITE_TOOLS`, so the plan-mode gate passes it automatically.
- `buildPlanExitPrompt` extracted to `src/agent/plan-mode-exit-prompt.ts` so the agent-layer handler can reuse it without a cli→agent dependency inversion; `/plan off` now imports it from there.

This is an **isolated-first slice**. A stacked follow-up PR will remove `/bypass`, add a positive `○ default` status badge, and reword the plan-mode addendum to steer the model to `exit_plan_mode` — deferred so `/plan off` stays valid until the tool is proven. `/plan` and `/afk` are unchanged here.

## Test results
- `vitest run src/agent/tools/handlers/exit-plan-mode.test.ts`: **7 passed** — approve-default & approve-bypass assert the seeded message `=== buildPlanExitPrompt(plansDir)` (identical to `/plan off`); keep-planning & declined → no mode flip, no seed; `AgentSession` injects controls for top-level (not subagent) sessions; single-shot drain.
- `vitest run` (full): **10085 passed, 12 skipped, 1 failed**. The single failure — `anthropic-direct.test.ts > compact()` — is **pre-existing and unrelated**: a model-ID alias assertion (`'claude-haiku-4-5'` vs `'claude-haiku-4-5-20251001'`); no compaction/model-resolution code is touched here.
- `pnpm lint` (tsc --noEmit): **clean**.

## Verification
Adversarial re-derivation from the diff (independent of this description) — all 3 load-bearing claims **CONFIRMED** with file:line evidence:
1. Approved exit seeds the SAME prompt as `/plan off` — both import `buildPlanExitPrompt` from the single shared `src/agent/plan-mode-exit-prompt.ts`.
2. Tool registered + schema offered **only** when `permissionMode === 'plan'` and controls present, in both providers.
3. Seed bridge: handler → `requestImplementSeed` → `AgentSession._pendingPlanExitSeed` (outlives the per-turn dispatcher) → REPL `takePendingPlanExitSeed()` drain → `seedBuffer`; single-shot; controls injected only for top-level sessions.

## Out of scope
`/bypass` removal, the `○ default` badge, and the plan-mode-addendum reword — landing as a **stacked follow-up PR** based on this branch.

### DIFF
--- a/src/agent/plan-mode-exit-prompt.ts
+++ b/src/agent/plan-mode-exit-prompt.ts
@@ -1,0 +1,36 @@
+/**
+ * The crafted "save the plan, then implement it" turn used when a session
+ * leaves plan mode.
+ *
+ * Invariant: this is the SINGLE source of truth for the plan-exit implement
+ * prompt. Two call sites materialize it and MUST produce byte-identical text:
+ *   - the `/plan off` slash command (`src/cli/slash/commands/plan.ts`), which
+ *     seeds it as the next REPL user turn; and
+ *   - the model-callable `exit_plan_mode` tool
+ *     (`src/agent/tools/handlers/exit-plan-mode.ts`), which seeds the same turn
+ *     through the session→REPL seed bridge on user approval.
+ *
+ * Lives in the agent layer (not under `src/cli`) so the agent-layer tool
+ * handler can import it without inverting the cli→agent dependency direction.
+ * It depends only on a pre-resolved `plansDir` string, so it pulls in nothing
+ * from either layer.
+ *
+ * @module agent/plan-mode-exit-prompt
+ */
+
+/**
+ * Build the prompt seeded into the user's next message buffer when plan mode
+ * is exited (via `/plan off` or an approved `exit_plan_mode`). `plansDir` is an
+ * absolute path to the session's `<cwd>/.afk/plans/` directory; the model picks
+ * a descriptive filename within it (the `write_file` tool creates the directory
+ * if absent).
+ */
+export function buildPlanExitPrompt(plansDir: string): string {
+  return [
+    'The user has switched off plan mode. Writes are now permitted. Do two things, in order:',
+    '',
+    `1. Save the plan. Write the plan you developed in this conversation to a new markdown file under \`${plansDir}/\` — pick a short, descriptive kebab-case filename (e.g. \`${plansDir}/refactor-auth-flow.md\`). Capture the full plan: the chosen approach, the concrete step-by-step changes, the risks named, and the alternatives considered. This is the durable record.`,
+    '',
+    '2. Implement the plan. Work through the steps you just recorded, verifying as you go — run the project\'s lint/test gates where they apply. End in a terminal state: Done with evidence, Blocked with the exact unblock condition, or Asking one precise question.',
+  ].join('\n');
+}
--- a/src/agent/providers/anthropic-direct/index.ts
+++ b/src/agent/providers/anthropic-direct/index.ts
@@ -51,6 +51,12 @@ export { resolveEffort, resolveMaxTokens, resolveThinkingParam } from './resolve
 import type { ToolDispatcher } from './tool-dispatcher.js';
 import { SessionToolDispatcher } from '../../tools/dispatcher.js';
 import { createBuiltinHandlers } from '../../tools/handlers/index.js';
+import {
+  exitPlanModeTool,
+  createExitPlanModeHandler,
+  EXIT_PLAN_MODE_TOOL_NAME,
+} from '../../tools/handlers/exit-plan-mode.js';
+import type { PlanExitControls } from '../../types/config-types.js';
 import { builtinToolSchemas, agentTool, skillTool, composeTool } from '../../tools/schemas.js';
 import { TOOL_SYSTEM_PROMPT_BASE, SLASH_COMMAND_ROUTING_PROMPT, BASH_PASSTHROUGH_PROMPT, MEMORY_SYSTEM_PROMPT, MEMORY_SYSTEM_PROMPT_READONLY } from '../../tools/system-prompt.js';
 import { withMcpToolsAllowed, type ToolPermissionConfig } from '../../tools/permissions.js';
@@ -322,6 +328,12 @@ export class AnthropicDirectProvider implements ModelProvider {
        * preserves any constructor-provided registry.
        */
       hookRegistry?: import('../../hooks.js').HookRegistry;
+      /**
+       * Session-control bridge for the model-callable `exit_plan_mode` tool,
+       * forwarded from the query config (top-level sessions only). When set AND
+       * `permissionMode === 'plan'`, the handler + schema are registered.
+       */
+      planExitControls?: PlanExitControls;
     },
   ): SessionToolDispatcher {
     const handlers = createBuiltinHandlers(permissionMode, opts?.cwd);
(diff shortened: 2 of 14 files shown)


griffinwork40 (commented):

**Merge decision: ✅ MERGE** (3 medium follow-ups noted; none blocking)

Reviewed at branch HEAD `9af8271`. Full regime (+528/-27, 14 files). Spec-compliance verified against the PR description; all blocking/high candidate findings were independently re-derived against the reviewed ref before inclusion. No critical/high findings; the mediums below are defense-in-depth / hygiene, not merge blockers.

### What's clean (verified against source)
- **Byte-identical prompt** — `buildPlanExitPrompt` lives once in `src/agent/plan-mode-exit-prompt.ts`; `src/cli/slash/commands/plan.ts` imports it (no divergent duplicate). The "`/plan off` and `exit_plan_mode` seed the same turn" invariant holds.
- **Plan-mode-only registration** — both providers gate on `permissionMode === 'plan' ? opts.planExitControls : undefined`, so the handler + schema are never registered outside plan mode (anthropic-direct/index.ts, openai-compatible/index.ts).
- **Picker keyword matching is safe** — `CHOICE_BYPASS` is the only label containing `bypass`; `CHOICE_KEEP` starts with `Keep`; all labels are ASCII <128 chars so `sanitizeSchemaString` is a no-op and values round-trip verbatim. No approve-default input falls through to bypass.
- CI green (Lint & Build, Test, Docs, Publish Bundle Smoke all SUCCESS). No secret exposure, no injection surface.

### Findings

**1. medium · security — `exit-plan-mode.ts` / `query.ts:551-567` (ref:9af8271).** The handler calls `await controls.setPermissionMode(mode)`, which synchronously mutates the *live, in-flight* dispatcher (`this.state.toolDispatcher.setAllowAll?.(bypass)`), and only *then* returns "end your turn now". Between that return and the model's `end_turn`, the gate is already flipped — a model that ignores the "END YOUR TURN" instruction (or a prompt injection) could issue write tools in the remainder of the same turn, in bypass mode without prompts. The instruction is guidance, not enforcement.
   *Suggestion:* record the approved mode alongside the seed and apply the mode flip atomically at drain time in `loop-iteration.ts` (same boundary as `takePendingPlanExitSeed`), so the gate stays locked until the clean next turn.

**2. medium · api-compat — `src/agent/types/session-types.ts:220` (ref:9af8271).** `takePendingPlanExitSeed(): string | undefined;` is added as a **required** member of `IAgentSession`, which is re-exported from the package entrypoint (`src/index.ts:25`). Any external `class … implements IAgentSession` breaks at compile time; `package.json` is `4.41.0` (minor). Callers are unaffected (structural typing); only implementors break.
   *Suggestion:* if external implementors are a supported surface, make the member optional (`takePendingPlanExitSeed?()`) or land it under a major bump; otherwise note it as an internal-interface change.

**3. medium · correctness — `src/cli/commands/interactive/loop-iteration.ts:182-187` (ref:9af8271).** The seed is drained only `if (seedBuffer === undefined)`. On any iteration where `seedBuffer` is already populated (e.g. a slash-command `kind:'submit'` re-enters with it set), `takePendingPlanExitSeed()` is skipped and the queued implement-turn is silently dropped — no error surfaced to user or model. (Permanent loss vs. next-iteration recovery depends on session-internal clearing not visible in this file.)
   *Suggestion:* guard `requestImplementSeed` to warn/no-op when `_pendingPlanExitSeed` is already set, and/or surface a debug log when the drain is skipped while a seed is pending.

**4. low · test-coverage — `src/cli/commands/interactive/loop-iteration.test.ts:122` (ref:9af8271).** `takePendingPlanExitSeed` is stubbed `vi.fn(() => undefined)` and never configured to return a real seed, so the REPL drain-to-submit path (lines 182-187) is exercised only on its `undefined` branch. The session-bridge half is well covered in `exit-plan-mode.test.ts` (7 tests, incl. single-shot drain + subagent exclusion); the loop-level integration is not.
   *Suggestion:* add a case that returns `'IMPLEMENT TURN TEXT'` once and asserts the loop submits it (and skips `readLine`) that iteration.

**5. nit · observability — `exit-plan-mode.ts` / `agent-session.ts`.** No debug trace on the mode-flip or seed-queue; a `debugLog` after each would make the control flow visible under `AFK_DEBUG=1`. Interactive cold-path, so no perf concern.

### What was not checked
- Read against branch HEAD `9af8271` — every blocking/high candidate citation was verified at that ref (no fabrications, no false-absence claims).
- Stated intent: PR #285 title + body — spec-compliance assessed and clean.
- Not runtime-traced: whether any external/published package implements `IAgentSession` directly; the full Telegram session surface end-to-end (the `parentSessionId === undefined` constructor guard should keep `planExitControls` off non-REPL surfaces, but this was not runtime-verified); the exact provider streaming-loop lifecycle for a deferred mode-flip fix.

griffinwork40:
## Addressed: medium · security — mid-turn TOCTOU on the permission flip (`e8850347`)

Resolved review finding **#1** (the `await controls.setPermissionMode(mode)` mid-turn flip).

**Before:** the `exit_plan_mode` handler flipped the live, in-flight dispatcher to the approved mode and *then* returned "end your turn now." Between that flip and the model's actual `end_turn`, the gate was already open — a model that ignored the instruction (or a prompt injection) could issue write tools in the remainder of the same turn in `bypassPermissions` mode with no prompts. The instruction was guidance, not enforcement.

**Fix (exactly the suggested approach — record the mode alongside the seed, flip atomically at drain time):**
- The handler no longer flips the mode. It records the approved mode alongside the seed: `requestImplementSeed(message, mode)`.
- The flip is applied atomically at the **post-turn drain boundary** in `AgentSession.takePendingPlanExitSeed()` (now `async`), which the REPL `loop-iteration.ts` drain `await`s. The gate stays **locked in plan mode for the entire current turn** and only opens for the clean, seeded implement-turn that follows — the same timing `/plan off` already has.

**Robustness (new):** the deferred flip is wrapped in try/catch in `takePendingPlanExitSeed`. If `setPermissionMode` rejects (e.g. query handle closing — the failure mode `togglePlanMode` already guards for `/plan off`), the seed is **dropped** and `undefined` returned: no implement-turn is auto-submitted while still gate-locked, and the rejection cannot escape into the `while(true)` REPL loop (which has no try/catch around the drain) and crash it.

**Surface changes:**
- `PlanExitControls.requestImplementSeed` gains a `mode` parameter.
- `IAgentSession.takePendingPlanExitSeed` is now `Promise<string | undefined>`; `loop-iteration.ts` awaits it.

**Tests:** handler tests now assert the handler makes **no** mid-turn `setPermissionMode` call and that the seed carries the approved mode; bridge tests await the async drain and pass the mode; a new test covers the drop-on-flip-failure path. `loop-iteration` / `repl-loop-wiring` stubs made async.

**Gates:** `pnpm lint` (tsc --noEmit) clean. Targeted + blast-radius suites green: `exit-plan-mode` (8), `loop-iteration` (3), `repl-loop-wiring` (11), plus `agent-session-wrappers`, `plan-mode-toggle`, `permission-mode-cycle`, `afk-mode-toggle`, `bypass`, `bypass-live-toggle`, `plan-mode-system-payload`, `openai-compatible/query`, `provider-router`, `slash-commands` — 191 tests passing.

Findings #2 (api-compat note on the now-async `IAgentSession` member), #3, #4 (loop-drain coverage — partially addressed by the async stubs), and #5 (observability) are not in scope for this security commit; #5's intent is partly met by the new `debugLog` on a dropped/failed flip.

PR #286, opened 2026-06-25
feat(plan-mode)!: retire /bypass, add default badge, steer exit to exit_plan_mode

griffinwork40:
## Summary
Stacked follow-up to #285, completing the "permission modes aren't driven by slash commands" direction. **This PR targets `afk/database-cycle-optimization` (#285), not `main`** — review/merge #285 first; this diff shows only the follow-up.

- **Remove `/bypass`** — a pure mode-switcher fully covered by the Shift+Tab cycle (`default -> plan -> bypass`). Bypass stays reachable live via Shift+Tab and at boot via `--dangerously-skip-permissions` / the `permissionMode` config key. Two user-facing help strings repointed to Shift+Tab.
- **Add a `○ default` status-line chip** — default previously rendered no chip (absence = contained), so cycling back to default just dropped the badge with no affirmative "you're locked down" signal.
- **Reword the plan-mode addendum** to make `exit_plan_mode` (from #285) the primary exit path (model-proposed + user-picked); `/plan off` and Shift+Tab remain documented fallbacks. Safe to land now the tool is shipped + tested in #285.

`/plan` and `/afk` are unchanged (they do more than flip a mode).

**BREAKING**: the `/bypass` slash command is removed. Toggle bypass live with Shift+Tab, or set it at boot via `--dangerously-skip-permissions` / `afk config set permissionMode bypassPermissions`.

## Test results
- `pnpm lint` (tsc --noEmit): **clean**.
- `vitest run` (full): **10077 passed, 12 skipped, 1 failed** — the failure is the pre-existing, unrelated `anthropic-direct.test.ts > compact()` model-ID alias flake (also present in #285 and on the base). Total count dropped 8 vs #285 because `bypass.test.ts` was removed with the command.
- Targeted green: status-line + permission-mode-cycle + slash registry (42 files / 609 tests), plan-mode-addendum (19 tests).

## Out of scope
Internal provider/dispatcher doc-comments still say "live `/bypass` toggle" as shorthand for the bypass-mode enforcement plumbing (unchanged — only the command was removed). Left as-is to keep this PR focused; a trivial comment sweep can follow.


griffinwork40 (commented):

**Merge decision: ✅ MERGE** — no critical, high, or blocking findings; 4 nit/low documentation + test-coverage follow-ups.

Reviewed against branch HEAD `5cab2ec` (regime full, 8 files: +18/-186). Every absence claim and file-state citation below was independently re-verified against the reviewed ref via `gh api contents?ref=…` and `git grep` on a checkout of the ref — no fabricated or false-absence findings survived.

> **Stacking note:** this PR targets `afk/database-cycle-optimization` (PR #285), **not `main`** — the diff is only the follow-up delta. Review/merge #285 first; this review assumes #285 lands ahead of it.

### What's solid (verified at `5cab2ec`)
- **Clean `/bypass` removal — no build break.** `git grep bypassCmd` over `src/` → **zero** references; `commands/bypass` import → zero; both `bypass.ts` and `bypass.test.ts` are deleted, and `src/cli/slash/index.ts` removed both the `import { bypassCmd }` and the `register(bypassCmd)` line. The deleted module has no remaining importer anywhere in production, so nothing dangles.
- **Security: bypass enforcement plumbing intact.** Removing the *slash command* does not weaken *bypass mode* — the `setPermissionMode → toolDispatcher.setAllowAll?.(bypass)` dual-field sync is unchanged in both providers, and bypass stays reachable via Shift+Tab and `--dangerously-skip-permissions`. The deletion of `bypass.test.ts` is correct (the tested unit no longer exists), not a coverage regression.
- **`○ default` chip correct.** The new branch is the last `else if` in the chip chain, fires only on `permissionMode === 'default'`, is tagged `never drop`, and leaves the existing `● plan` / `◐ AFK` / `⚡ bypass` branches structurally unchanged (no regression). `undefined` still falls through to no chip.
- **Addendum is consistent with the `exit_plan_mode` tool contract (#285).** The reworded text and the tool's own schema description agree on all four behavioral points (call the tool when implementation is needed; END YOUR TURN after; do NOT use `ask_question` to ask "is this plan ok?"; skip the tool for research/read-only tasks). `/plan off` and Shift+Tab are correctly demoted to documented fallbacks.

### Findings (none blocking)

**1. nit · security/docs · `README.md:171,173` + `docs/architecture.md:129,143` · ref:5cab2ec** — User-facing docs still instruct users to run `/bypass` as a live REPL command, which no longer exists. These are **not** internal provider/dispatcher doc-comments, so they are **not** covered by the PR's explicit "left as-is" carve-out (that exemption covers `query.ts` inline comments only).
  - `README.md:171` — ``- For one session: `/bypass off` in the REPL``
  - `README.md:173` — ``Toggle bypass live anytime with `/bypass`.``
  - `docs/architecture.md:129` — ``- Enable with `/bypass` in the REPL``
  - `docs/architecture.md:143` — ``… or `/bypass` (live, one session).``
  _Fix:_ replace each `/bypass` reference with "Shift+Tab (the permission-mode cycle)" in README.md and docs/architecture.md.

**2. low · spec-compliance · `src/cli/status-line.ts:34` · ref:5cab2ec** — The TSDoc on the exported `StatusLineFields.permissionMode` field still reads ``'default' renders none`` — stale immediately below the new branch this PR adds. This is a `/** */` doc on an exported interface (user-visible API doc), not an internal `//` comment, so it is in-scope.
  _Evidence:_ `* … since bypass is the default mode). 'default' renders none.`
  _Fix:_ update to ``'default' renders ○ default (success tone)``.

**3. low · test-coverage · `src/cli/status-line.test.ts` · ref:5cab2ec** — Zero tests assert that `permissionMode: 'default'` renders `○ default`. The new branch is the only chip in the chain without coverage; `plan` (`:922 expect(out).toContain('● plan')`) and `autonomous` (`:943 toContain('◐ AFK')`) both have never-drop tests. (Verified: `grep '○ default'` in the test file → 0 matches.)
  _Fix:_ add a test mirroring the existing `● plan` never-drop case with `permissionMode: 'default'` asserting `out.toContain('○ default')`.

**4. low · test-coverage · `src/agent/providers/anthropic-direct/plan-mode-addendum.test.ts:97-98` · ref:5cab2ec** — No assertion covers `exit_plan_mode` — the new *primary* exit path. The surviving test ``it('names /plan off as the exit affordance')`` is now misleading, since this PR demotes `/plan off` to a fallback. (Verified: `grep 'exit_plan_mode'` in the test file → 0 matches.)
  _Fix:_ add `expect(PLAN_MODE_ADDENDUM_TEXT).toContain('exit_plan_mode')` and rename the test to "names exit_plan_mode as the primary exit and /plan off as a manual fallback".

### Spec-compliance (assessed against PR #286 title + body) — all PASS
`/bypass` fully removed (file deleted, import + register removed, no other registration path) ✓ · both named user-facing help strings repointed to Shift+Tab (`interactive.ts:188`, `shared.ts:233-234`) ✓ · `○ default` chip added and guarded to `'default'` only ✓ · existing AFK/bypass/plan chips preserved ✓ · addendum steers to `exit_plan_mode` with `/plan off` + Shift+Tab as documented fallbacks ✓ · `/plan` and `/afk` unchanged ✓ · **no scope creep**. (Findings 2-4 are doc/test consistency items the intent implies but the diff did not carry; none contradict the stated scope.)

### What was not checked
Wave-1 agents read against branch HEAD `5cab2ec`; every absence/structural claim above was independently re-verified at that ref (the README/architecture `/bypass` strings, the `status-line.ts:34` JSDoc, the two untested branches, and the zero-`bypassCmd`-importer build-break check were each confirmed by reading the file / `git grep` at the ref). Did not run the test suite locally (the PR reports `pnpm lint` clean and vitest 10077 pass / 1 pre-existing unrelated `anthropic-direct.test.ts > compact()` flake also present on the base). Did not exercise the `exit_plan_mode` approval flow end-to-end at runtime, did not render the `○ default` chip on a live terminal, and did not assess the internal provider/dispatcher `/bypass` doc-comments (explicitly out of scope per the PR body).

PR #287, opened 2026-06-26
fix(telegram): stop mid-turn cutoffs, false rate-limit errors, and tool-call spam

griffinwork40:
## Summary

Fixes three coupled Telegram-bot failures that all surface during a single sub-agent fan-out (reported symptoms: *"it gets cut off mid-agent and I have to send a `.` to get the result"*, *"random rate-limit/timeout errors when I'm not actually rate-limited on Claude"*, and *"tool calling is spammy"*). They share one causal chain: **spam → edit-flood → Telegram 429 → false "rate limit"**, and **fan-out silence → false 60s timeout → orphaned turn → cutoff**.

Root causes were investigated with parallel sub-agents and **shadow-verified** (independent re-derivation from source). Full analysis + deferred follow-ups: [`docs/telegram-bug-investigation.md`](docs/telegram-bug-investigation.md).

## What changed

**1. Mid-turn cutoff [Critical]** — `streaming.ts`
- The 60s inactivity timeout fired while the *parent* stream was legitimately quiet (children running), threw, and abandoned the stream **without aborting the turn**. The turn kept running into the long-lived shared `providerIterator`; the next message drained its buffered events, producing the *"send a `.` to recover the lost result"* symptom.
- Fix: **re-arm the watchdog on sub-agent sink activity** (silence window 60s → 180s, measured from last activity), and **`session.interrupt()` on a genuine timeout** — the same turn-scoped abort the REPL uses for ESC — so the turn aborts and `providerIterator` parks cleanly at the next-prompt boundary.

**2. False "Rate limit reached" [High]** — `error-utils.ts`, `handlers/message.ts`
- The bot's own edit/reply flood tripped Telegram's `429 Too Many Requests`, whose message matched `isRateLimitError` and was reported to the user as a **Claude** rate limit.
- Fix: new `isTelegramTransportError` guard, checked **before** the provider classifiers at every catch site, so a Telegram-side limit shows an honest Telegram-attributed message (or stays silent) instead of blaming Claude.

**3. Tool-call spam — the "recon: read_file ..." wall [High]** — `streaming.ts`
- `subagentSink` appended one `◦ label: tool args` line **per child tool call**, unbounded, and that edit flood fed bug #2.
- Fix: a **bounded rolling footer** (step counter + last 4 lines) via the pure, unit-tested `renderSubagentFooter`.

**Also**
- Clear the usage-limit `countdownInterval` on **every** exit path — it leaked on the timeout-throw path, editing a dead message forever — and reset `editInFlight`.
- Surface `StreamTimeoutError` honestly instead of as a generic/network error. It lives in its own module so the handler's `instanceof` check survives the many `vi.mock('./streaming.js')` tests.

## Scope / blast radius

Entirely within `src/telegram/`: `streaming.js` and `error-utils.js` are imported only by Telegram code, and `error-classifiers.ts` is unchanged. Additive and reversible; no destructive operations.

## Not in this PR (documented follow-ups)

`docs/telegram-bug-investigation.md` also catalogs lower-severity issues left for follow-up: `drainQueue` re-entrancy guard, `enqueueClear`/`enqueueCompact` bypassing the queue cap, HTML-unaware `splitLongMessage`, the elicitation `pending` map not cleared on restart, `/clear` `getSession` throw, `pushMarkdown` duplicate-delivery, and a daemon force-exit-mid-turn edge.

## Verification

- `pnpm test`: 10114 passed, 13 skipped, 1 failure (`anthropic-direct.test.ts > compact()`) confirmed **pre-existing** on the clean base (stash-verified) and unrelated to this change.
- New tests: `renderSubagentFooter` bounding; interrupt-on-timeout (fake timers) and no-interrupt on provider errors; the Telegram-429 classifier guard. A self-introduced mock-coupling regression (`message-handler-drain.test.ts`) was caught and fixed during development.

### DIFF
--- a/docs/telegram-bug-investigation.md
+++ b/docs/telegram-bug-investigation.md
@@ -1,0 +1,195 @@
+# Telegram bot reliability — root-cause investigation
+
+Investigation date: 2026-06-26. Method: 5 parallel read-only sub-agent investigators
+(cutoff, false-errors, progress-spam, general-sweep, daemon) + direct code reads.
+Scope: `src/telegram/**`, `src/agent/session/**`, `src/agent/subagent/**`, `src/telegram.ts`,
+`src/service/**`. **No code was changed** — this is diagnosis only.
+
+**Shadow-verification (2026-06-26):** the three load-bearing claims (Bug 1 cutoff mechanism,
+Bug 2 false-rate-limit, Bug 4 timer leak) were independently re-derived from primary sources by
+three parallel adversarial verifiers — **all CONFIRMED**. Corrections folded in below. Net new
+facts from verification: `session.interrupt()` (agent-session.ts:628-632) and `session.abort()`
+(agent-session.ts:423) **already exist** and route to `AbortCoordinator.requestAbort('interrupted')`
+→ the Bug 1 "smallest fix" is confirmed viable; `isRateLimitError` has **two more** vulnerable
+call sites at `message.ts:330` (photo) and `message.ts:422` (text outer handler) beyond the one
+in `processOne`.
+
+---
+
+## TL;DR — the three reported bugs share one causal chain
+
+A single heavy turn (one that fans out sub-agents — `/ground-state`, `/diagnose`, `compose`,
+research skills) triggers **all three** complaints at once, in this order:
+
+1. **Spam** — `subagentSink` appends one `◦ label: tool args` line **per child tool call**,
+   unbounded, into a single ever-growing edited message. This is the "recon: read_file …" wall.
+2. **False "rate limit"** — that per-tool-call editing **floods Telegram's API**, which returns
+   HTTP **429 "Too Many Requests"**. The classifier matches that string and tells the user
+   *"⏳ Rate limit reached"* — as if **Claude** rate-limited them. It didn't; **Telegram** did.
+3. **Cutoff** — while the sub-agents run, the **parent** stream is silent at the provider level
+   for >60s, so the streaming timeout fires, throws, and **abandons the stream without aborting
+   the underlying turn**. The turn keeps running; its events buffer; the user sees an error.
+   Sending `.` starts a new consumer that **drains the buffered events** → the delayed final answer.
+
+So: **spam → edit-flood → Telegram 429 → false "rate limit"**, and **fan-out silence → 60s
+timeout → orphaned turn → "cut off, recovered by a period."** Fixing the spam (cap/coalesce) and
+the timeout (abort-on-timeout + reset-on-subagent-activity) removes the whole cluster.
+
+---
+
+## Bug 1 — Mid-turn cutoff; final result withheld until a follow-up message  ⟶ CRITICAL
+
+**Symptom (user):** turn starts, gets cut off mid-agent, no final result; sending a `.` later
+delivers the withheld result with lag.
+
+**Root cause chain** (confidence: High):
+1. `streaming.ts:199-233` `nextWithTimeout()` rejects with *"Response timed out"* after
+   `NEXT_EVENT_TIMEOUT_MS = 60_000` (streaming.ts:22) of **parent-stream** silence.
+2. During sub-agent fan-out the parent provider loop is parked inside the tool dispatcher; **no
+   `OutputEvent` crosses the parent `providerIterator` until all tools return.** Sub-agent
+   progress reaches Telegram via the out-of-band `subagentSink`/`runWithSink` channel
+   (`streaming.ts:238-250`, `skill-sink-channel.ts`) which calls `sendOrEdit` directly and **does
+   not reset the 60s timeout.** So a busy turn looks "silent" and the timeout fires falsely.
+3. On timeout, `streamResponse` throws; the `finally` (streaming.ts:438-444) calls
+   `iter.return?.()`. That runs the generator's cleanup `finally` in `agent-session.ts:489-497`
+   (sendMessageStream) / `541-543` (sendMessageStreamInternal), setting `currentState = 'idle'` —
+   **but never aborts `providerQuery` / the turn's `AbortController`.** The shared `providerIterator`
+   is created once per session in `initSdkLifecycle` (`agent-session.ts:254-255`) and reused by
+   every `sendMessageStream` call — so the provider keeps streaming into it, now unconsumed.
+   [verified: shared-iterator lifecycle re-derived from agent-session.ts + query.ts + abort-coordinator.ts]
+4. User sees a generic/timeout error (`message.ts:602-608`).
+5. The `.` follow-up sees `session.state === 'idle'` (set in step 3) → `processOne` →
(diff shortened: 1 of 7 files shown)


griffinwork40:
Verification complete. The production path (`message.ts:592-593`) calls `streamResponse(..., { cleanFinal: true })`, and the `done` handler delivers `answerText` (which never contains the footer) via `deliverClean()` then **deletes** the live-preview message. The bounded footer is confined to the transient preview and removed on completion — **no footer leak into the final delivered message**. The real-bug risk I probed is ruled out.

All load-bearing claims are confirmed at ref `be939f5`. No critical/high findings exist, so `/shadow-verify` (gated on critical/high) is not triggered. Synthesizing.

---

# Code Review — PR #287

**`fix(telegram): stop mid-turn cutoffs, false rate-limit errors, and tool-call spam`**

| | |
|---|---|
| **Target** | PR #287 · `afk/fix-telegram-message-cutoff` → `main` |
| **Reviewed ref** | `be939f5a37c38f1b82376eef2454f62a6584d3b1` (worktree clean; citations verified at this SHA) |
| **Regime** | full (2-agent Wave 1 + inline citation/absence verification) |
| **Change type** | bug-fix · +508 / −27 · 7 files (1 pure-docs skipped, 2 new modules, 2 tests, 2 modified) |
| **Stated intent** | PR title + body — **spec-compliance assessed** |

## Spec compliance — FULLY MET ✅

All three fixes + every "Also" item are implemented and verified at the reviewed ref. No unmet intent, no scope creep.

| Spec item | Status | Evidence (ref:be939f5) |
|---|---|---|
| #1 re-arm watchdog on sink activity (`lastActivityAt`) | ✅ | `streaming.ts:304` (sink), `:334` (parent event) |
| #1 `NEXT_EVENT_TIMEOUT_MS` 60s→180s | ✅ | `streaming.ts:29` |
| #1 `session.interrupt()` on timeout, **before** `iter.return()` | ✅ | `streaming.ts:520-536` (interrupt at :521 precedes iter.return at :536) |
| #2 `isTelegramTransportError` guard at **all 3** catch sites | ✅ | `message.ts:334, 430, 617` (verified by grep) |
| #3 bounded footer (`renderSubagentFooter`, cap=4) | ✅ | `streaming.ts:51-56, 311-313` |
| Also: `countdownInterval` + `editInFlight` reset in `finally` | ✅ | `streaming.ts:527-531` |
| Also: `StreamTimeoutError` in own module (mock-safe `instanceof`) | ✅ | `stream-timeout-error.ts:1-26` |

**Footer-leak check (correctness, my own verification):** the footer appears only in `livePreview() = accumulated + renderSubagentFooter(...)`, never stored in `accumulated` or `answerText`. The production `done` path (`streaming.ts:457-466`, `cleanFinal: true`) delivers the noise-free `answerText` and deletes the preview message. **No footer leaks into the final reply.** The fix is strictly better than the old unbounded `accumulated += …`.

## Findings (sorted: high → low → nit; none blocking)

**No critical · No high · No blocking.** Security clean; api-compat non-breaking (all pre-existing exports present, change is additive).

**LOW · high confidence · test-coverage · `streaming.test.ts` · ref:be939f5**
The **core mechanism of fix #1 — sink activity re-arms the watchdog and prevents a false timeout — is untested.** The watchdog suite only covers *total silence* (the negative path). Absence verified: `rg "lastActivityAt|subagentSink" src/telegram/streaming.test.ts` matches only a comment, no positive-path test.
*Suggestion:* start a silent generator, fire a synthetic sink event just before `windowMs`, advance timers past the original deadline, assert `StreamTimeoutError` is **not** thrown.

**LOW · high confidence · test-coverage · `streaming.test.ts` · ref:be939f5**
The `countdownInterval`-leak-on-timeout fix (the "Also" item) is untested. Absence verified: `rg "countdownInterval|clearInterval|paused" src/telegram/streaming.test.ts` → zero matches.
*Suggestion:* yield a `paused` event (sets the interval), advance past the timeout, assert `vi.getTimerCount() === 0` after rejection.

**LOW · medium confidence → nit · test-coverage · `streaming.test.ts:507` · ref:be939f5**
The interrupt-on-timeout test asserts `interrupt` is called once but **not that it runs before `iter.return()`** — the load-bearing ordering. The test stays green even if the order is swapped (which would silently reintroduce the cutoff bug). The source ordering at `:520-536` is correct and documented; only the regression guard is weak.
*Suggestion:* record call order with a shared sequence counter across the two `vi.fn()`s.

**NIT · perf-observability · `streaming.ts:520-521` · ref:be939f5**
No log on the timeout→interrupt path, and `session.interrupt?.().catch(() => {})` swallows interrupt failures silently. The `StreamTimeoutError` itself is logged downstream (`message.ts:601`), so this is observability polish, not a gap.

**NIT · api-compat · `src/telegram/index.ts:24` · ref:be939f5**
`isTelegramTransportError` is not re-exported from `telegram/index.ts` (so it is absent from the public package API). All callers are internal to `src/telegram/` and nothing is removed — **not a break**. Add the export only if external/plugin authors are meant to compose Telegram error handling.

**NIT · correctness · `streaming.ts:277` · ref:be939f5**
The initial `setTimeout(arm, windowMs)` can produce one spurious `arm()` wake-up if activity arrived shortly after the call; it self-corrects by re-arming for the true `remaining`. Harmless (one extra macrotask), not a timing error.

## What was not checked
- **Refs:** Both Wave 1 agents read against branch HEAD `be939f5`; I independently re-verified every load-bearing citation, all 3 catch sites, the finally ordering, the test-absence claims, and the footer/`answerText` separation at that ref.
- **Stated intent:** PR #287 title + body — spec-compliance assessed (fully met).
- Did **not** run the test suite (static review only). PR self-reports `pnpm lint` clean and one **pre-existing** failure (`anthropic-direct.test.ts > compact()`, stash-verified on base, unrelated).
- Did **not** review out-of-scope items the PR explicitly defers (`drainQueue` re-entrancy, HTML-unaware `splitLongMessage`, daemon force-exit, etc. — catalogued in `docs/telegram-bug-investigation.md`).
- **Not statically verifiable:** `instanceof TelegramError` across multiple telegraf copies (single-package pnpm install makes this safe); runtime timer drift in `arm()`.

The change correctly and completely implements its stated intent, is additive/reversible, confined to `src/telegram/`, and carries no security, correctness, or api-compat blocker. The footer-leak and timeout-abort risks I probed are both ruled out at source. Remaining items are low/nit test-coverage and observability improvements — sound as fast-follow, none gating merge.

**Done** — Read-only review of PR #287 complete; merge recommendation is **MERGE**.
- Reviewed at ref `be939f5`; Wave 1 (2 parallel agents) + inline citation/absence/correctness verification; no critical/high findings, so `/shadow-verify` not triggered.
- Spec fully met (8/8 items verified); no files were modified, committed, or pushed — this skill is read-only and surfaced findings only.
- Deferred to the author as optional follow-ups: 2 missing tests (re-arm-on-sink-activity, countdownInterval-leak), 1 test-ordering guard, and minor observability/export nits — all non-blocking.

---

PR #288, opened 2026-06-26
fix(anthropic-direct): preserve redacted_thinking blocks to prevent session wedge

griffinwork40:
## Summary

Fixes a bug where an interactive REPL session could **permanently wedge** ("the turn ended itself and I can't send anything else") after the model used tools. Observed with `opus_1m` (resolves to `claude-opus-4-8`) on a query that did extended thinking → parallel `web_scrape` calls; reproduced across 3 sessions in witness telemetry.

## Root cause

The `anthropic-direct` stream translator (`src/agent/providers/anthropic-direct/translate.ts`) had **no handling for `redacted_thinking` content blocks** at all:

- `BlockAcc` had no `redacted_thinking` variant,
- `content_block_start` had no case for it,
- `buildTurnResult` never emitted one.

So when the model emitted *redacted* reasoning (common on security-adjacent prompts — here, "jailbreak an old Amazon Echo") and then called tools, the assistant turn was reconstructed as `[text?, tool_use…]` with **no leading reasoning block**.

The Anthropic Messages API requires an assistant turn containing `tool_use` to **lead with a `thinking`/`redacted_thinking` block when extended thinking is enabled**. The post-tool-use continuation request therefore 400s. And because the provider reuses **one stable `state.messages` array across turns**, mutated in place (`query.ts`), whose only pre-request self-heal is `repairOrphanToolUses` (which fixes orphan `tool_use` but *not* a missing leading reasoning block), the malformed turn **persists and re-fails on every subsequent request** — wedging the session. The first request succeeds because there is no prior assistant turn to round-trip.

(Model context: `opus_1m → claude-opus-4-8`, an `isOpus47Plus` model using adaptive/summarized thinking at effort `max` — the deepest thinking-block contract.)

## Fix

Capture `redacted_thinking` at `content_block_start` (it is delivered whole — no deltas) and preserve it verbatim in `assistantBlocks`, so the assistant turn round-trips with its required leading reasoning block. The payload is server-encrypted and always round-trippable, so it is pushed unconditionally (unlike `thinking`, which is correctly dropped when it lacks a signature). No visible stream event is emitted — the payload is opaque.

- No new `@anthropic-ai/sdk` symbol (`redacted_thinking` is part of the already-imported `ContentBlockParam` union) → SDK dependency lock unchanged.

## Test plan

- `pnpm exec vitest run …/translate.test.ts`: **10/10 pass** (8 existing + 2 new).
  - New regression test: a `redacted_thinking` block followed by `tool_use` must yield an assistant turn whose **first block is the reasoning block** (the exact shape that, when dropped, caused the wedge).
- `pnpm audit:sdk:check`: OK (lock matches imports).
- Full suite: 1 failure, `anthropic-direct.test.ts > compact() summarizes…` (`claude-haiku-4-5` vs `claude-haiku-4-5-20251001`) — **pre-existing**, verified to fail identically on clean `main@2027bb5`; unrelated to this change.

## Honest scope note / recommended follow-up

This fixes the **most likely trigger** (a confirmed `translate.ts` gap). The exact trigger of the original session wasn't captured (no `[afk] thinking-block diagnostic` log line / live repro), so as **defense-in-depth** I recommend a follow-up PR adding a wedge-resilience self-heal (sibling to `repairOrphanToolUses`): before a request — or on a thinking-flavored continuation 400 — detect an assistant turn that contains `tool_use` but lacks a leading `thinking`/`redacted_thinking` block while thinking is enabled, and repair/roll it back so **no** malformed turn (from any cause) can permanently brick a session. I scoped it out here to keep this change minimal and reviewable in delicate hot-path provider code.

🤖 Diagnosis verified via parallel shadow-verify (CLAIM "error is silently swallowed" was corrected → the error is rendered via `presentError`; the load-bearing wedge mechanism and the `redacted_thinking` gap were independently confirmed).

### DIFF
--- a/src/agent/providers/anthropic-direct/translate.test.ts
+++ b/src/agent/providers/anthropic-direct/translate.test.ts
@@ -152,6 +152,14 @@ function signatureDelta(index: number, signature: string): RawMessageStreamEvent
   } as unknown as RawMessageStreamEvent;
 }
 
+function redactedThinkingStart(index: number, data: string): RawMessageStreamEvent {
+  return {
+    type: 'content_block_start',
+    index,
+    content_block: { type: 'redacted_thinking', data },
+  } as unknown as RawMessageStreamEvent;
+}
+
 describe('anthropic-direct translateMessageStream', () => {
   it('text-only stream emits delta.text events and an end_turn turn-result', async () => {
     const events: RawMessageStreamEvent[] = [
@@ -372,6 +380,66 @@ describe('anthropic-direct translateMessageStream', () => {
       type: 'text',
       text: 'hello',
     });
+  });
+
+  it('preserves a redacted_thinking block verbatim in turn-result', async () => {
+    const events: RawMessageStreamEvent[] = [
+      messageStart(),
+      redactedThinkingStart(0, 'ENCRYPTED_PAYLOAD'),
+      blockStop(0),
+      textBlockStart(1),
+      textDelta(1, 'done'),
+      blockStop(1),
+      messageDelta('end_turn'),
+      messageStop(),
+    ];
+
+    const out = await collect(
+      translateMessageStream(fromArray(events), { sessionId: SESSION_ID }),
+    );
+
+    const last = out[out.length - 1];
+    if (!last || last.kind !== 'turn-result') {
+      throw new Error('expected turn-result at end');
+    }
+    expect(last.result.assistantBlocks).toEqual([
+      { type: 'redacted_thinking', data: 'ENCRYPTED_PAYLOAD' },
+      { type: 'text', text: 'done' },
+    ]);
+  });
+
+  it('redacted_thinking followed by tool_use leads the assistant turn with the reasoning block (regression: must not be dropped)', async () => {
+    // Regression for the session-wedge bug: a security-adjacent prompt makes
+    // the server emit redacted_thinking, then the model calls tools. When
+    // extended thinking is enabled, the continuation request 400s unless the
+    // assistant turn LEADS with a thinking/redacted_thinking block. Dropping
+    // the redacted block (the pre-fix behavior) produced a tool_use-only turn
+    // that 400'd on every subsequent request — permanently wedging the session.
+    const events: RawMessageStreamEvent[] = [
+      messageStart(),
+      redactedThinkingStart(0, 'OPAQUE'),
+      blockStop(0),
+      toolUseStart(1, 'toolu_9', 'web_scrape'),
+      inputJsonDelta(1, '{"url":"https://x"}'),
+      blockStop(1),
+      messageDelta('tool_use'),
+      messageStop(),
+    ];
+
+    const out = await collect(
+      translateMessageStream(fromArray(events), { sessionId: SESSION_ID }),
+    );
+
+    const last = out[out.length - 1];
+    if (!last || last.kind !== 'turn-result') {
+      throw new Error('expected turn-result at end');
+    }
+    // The reasoning block must come FIRST, before the tool_use.
+    const blocks = last.result.assistantBlocks;
+    expect(blocks).toHaveLength(2);
+    expect(blocks[0]).toEqual({ type: 'redacted_thinking', data: 'OPAQUE' });
+    expect(blocks[1]).toMatchObject({ type: 'tool_use', id: 'toolu_9', name: 'web_scrape' });
+    expect(last.result.toolUseBlocks).toHaveLength(1);
   });
 
   it('error mid-stream yields an error event, no turn-result, and does not throw', async () => {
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

**Merge decision: ✅ MERGE** — no critical, high, or blocking findings; 1 medium + 1 low + 1 nit, all defense-in-depth follow-ups.

Reviewed at branch HEAD `40984ad` (regime LIGHT: hotfix, 2 files, +88/−1). Spec-compliance assessed against the PR title + body. The load-bearing claim — that the preserved `redacted_thinking` block reaches the wire and leads the assistant turn — was independently re-verified against the reviewed ref by reading `loop.ts` directly (not just the `TurnResult` struct).

### What's solid (verified at `40984ad`)
- **Fix reaches the wire, not just the struct.** On the tool-use exit (the wedge path), `loop.ts:516` pushes `turnResult.assistantBlocks` **verbatim** (no filter), so the new `redacted_thinking` block leads the assistant turn exactly as the Messages API requires. On non-tool-use exits, `loop.ts:487-489` `safeAssistantBlocks` strips only `b.type !== 'tool_use'`, so `redacted_thinking` is preserved there too.
- **Order-preservation holds.** The sparse `blocks` array keeps original index order: `redacted_thinking` at index 0 precedes `tool_use` at index 1, so the reconstructed turn is `[redacted_thinking, tool_use]` — the precise shape that, when dropped, caused the 400/wedge.
- **`unconditional` push is correct.** Unlike `thinking` (dropped when it lacks a signature), `redacted_thinking` is server-encrypted, has no signature to validate, and is always round-trippable — so pushing it unconditionally is right, and it correctly lands in the `else if (acc.kind === 'redacted_thinking')` branch before the `tool_use` `else`, leaving the tool_use path unpolluted.
- **Non-breaking + no SDK churn.** The `BlockAcc` union gains one additive member; no exhaustive `switch`/`assertNever` over `BlockAcc` exists in the provider, so nothing breaks. `redacted_thinking` is already in the imported `ContentBlockParam` union — SDK lock unchanged. `repairOrphanToolUses` operates only on tool_use/tool_result pairing and does not interact with the new block.

### Findings (none blocking)

**1. medium · correctness · `src/agent/providers/anthropic-direct/translate.ts:165-169` · ref:40984ad** — `redacted_thinking` is captured at `content_block_start` on the documented invariant that it carries no deltas. If the SDK ever delivers a `content_block_delta` for a `redacted_thinking` index (future SDK change or a non-conformant stream), the delta handler's `if/else if` chain has no branch for it and silently drops it — `acc.data` keeps only the `content_block_start` value. Correct per the current API; fragile if the "no deltas" invariant ever breaks, and invisible if it does.
  _Fix:_ in the trailing `// ignore` arm of the `content_block_delta` switch, add a `traceEnabled` debug line naming an unexpected delta for a `redacted_thinking` index, so an SDK regression surfaces under `AFK_TELEGRAM_TRACE` without touching production behavior.

**2. low · test-coverage · `src/agent/providers/anthropic-direct/translate.test.ts:382-445` · ref:40984ad** — Both new tests place `redacted_thinking` at index 0. No test exercises a `redacted_thinking` block at a **non-zero** index after a preceding `text` block (the order-preservation path through the sparse array), nor multiple redacted blocks. The "reasoning leads" property is tested only via index-0 alignment.
  _Fix:_ add a case `text@0 → redacted_thinking@1 → tool_use@2` asserting `assistantBlocks` is `[text, redacted_thinking, tool_use]` in that order.

**3. nit · observability · `translate.ts:165-169` · ref:40984ad** — The redacted block is captured with no debug trace (intentionally emits no visible stream event since the payload is opaque), so its presence is invisible under `AFK_TELEGRAM_TRACE`. A single `traceEnabled` log at capture time would make the round-trip path observable when diagnosing future thinking-block issues. Cold path; no perf concern.

### Spec-compliance (assessed against PR #288 title + body) — all PASS
`BlockAcc` gains the `redacted_thinking` variant ✓ · `content_block_start` captures `cb.data` whole (no delta path) ✓ · `buildTurnResult` pushes `{type:'redacted_thinking', data}` unconditionally ✓ · variant routes before the tool_use `else` (path unpolluted) ✓ · fix reaches the wire and leads the tool-use turn (`loop.ts:516` verbatim push) ✓ · regression test asserts reasoning-block-first on the redacted_thinking→tool_use shape ✓ · **no scope creep** (the wedge-resilience self-heal is explicitly deferred in the PR body as a follow-up). Findings 1-3 are hardening/coverage items the intent implies but the minimal diff did not carry; none contradict the stated scope.

### What was not checked
Read against branch HEAD `40984ad`; every citation above (the `translate.ts` branches, the `loop.ts:487-516` wire path, the `BlockAcc` exhaustiveness check, and the `repairOrphanToolUses` non-interaction) was verified at that ref. Did not run the test suite locally — the PR reports `pnpm lint` clean and `translate.test.ts` 10/10 pass, with one **pre-existing, unrelated** full-suite failure (`anthropic-direct.test.ts > compact()`, a `claude-haiku-4-5` model-id string mismatch) confirmed to fail identically on clean `main`. Did not confirm whether `RedactedThinkingBlock.data` is typed optional in the installed SDK (the API spec states it is always present); did not exercise the live opus_1m → claude-opus-4-8 repro at runtime. At review time the PR's `Test` CI check was still IN_PROGRESS (Lint & Build, Docs, Publish-Bundle-Smoke green) — merge should await a green `Test` run.

PR #289, opened 2026-06-26
fix(anthropic-direct): recover from a wedged session on rejected continuation

griffinwork40:
## Why

A user hit a session that "stopped" and then **couldn't send anything else** — every subsequent message died in ~2s with no output. Telemetry (3 sessions) showed the same signature: a turn used parallel tools, then the post-tool **continuation request was rejected by the API**, and from that point on every later turn re-failed identically.

### Root mechanism (code-confirmed)

The provider keeps **one `state.messages` array, mutated in place and reused across turns** (`query.ts:216`). A tool-use turn pushes `[assistant(thinking+tool_use), tool_result]` into it (`loop.ts:514`, `loop.ts:689`). If the **next iteration's continuation request** is rejected (a malformed thinking block under interleaved thinking — enabled unconditionally at `auth.ts:29` — or any non-retryable 4xx), `loop.ts:282–296` yields an `error` event and returns, but **those pushed turns are never rolled back**:

- the loop's own rollback (`messagesRollbackIdx`, `loop.ts:699`) only covers throws **during tool execution**, not the next request's failure;
- `repairOrphanToolUses` (`query.ts:333`) only heals **orphan `tool_use`** — here the array ends in a *matched* `tool_result`, so it no-ops.

The session generator then continues to the next prompt with poisoned history → every later request re-sends and re-fails → **permanent wedge**.

## What this changes

On any turn that surfaces an `error` event, truncate the failed turn's pushes back to the **known-sendable prefix** (history through the triggering user turn — a prefix the API already accepted). A failed turn now degrades to a **recoverable per-turn error** instead of bricking the session.

- `+` capture `sendablePrefixLen` right after the user turn is appended
- `+` track `turnErrored` from the per-turn event stream
- `+` `state.messages.splice(sendablePrefixLen)` on error (in place; honors the never-reassign invariant)
- Gated `!aborted` so an interrupt's partial state is preserved for resume.

## Relationship to #288

Defense-in-depth. #288 fixed **one root-cause variant** (dropped `redacted_thinking` blocks). This recovers **regardless of which block the API rejects** — and the regression test's diagnostic output (`invalid thinking blocks: none found (cause may be elsewhere)`) shows the locally-detectable signature check does **not** catch every trigger, which is exactly why a recovery net is needed.

## Honest scope

The **exact** block the API rejected in the original incident is **not runtime-confirmed** — the captured repro ran ~2 min before #288 was installed, and the error isn't persisted to telemetry (only stderr). This PR does **not** claim to eliminate the underlying rejection; it eliminates the **catastrophic permanent wedge** that turns one bad turn into a dead session. Confirming/fixing the precise trigger (likely interleaved-thinking multi-turn validation) is a worthwhile follow-up that needs a live repro with `[afk] thinking-block diagnostic` captured.

## Tests

New `wedge-recovery.test.ts` drives the **real `AnthropicDirectQuery` generator** through `tool-use turn → rejected continuation → next prompt` and asserts the next request no longer re-sends the poisoned turn.

- Verified **red→green**: with the fix disabled, the test fails (turn 2 re-sends the `assistant(tool_use)` poison); with it, passes.
- `pnpm lint` clean · `pnpm audit:sdk:check` OK (no new SDK symbols) · all **329** `anthropic-direct/` provider tests pass (abort, interrupt, usage-limit retry, reauth, compact).
- Independent adversarial review of the diff: 6/6 edge cases SAFE (abort, thrown-error path, compact interaction, over/under-rollback, never-reassign, consecutive-user validity), no must-fix.

## Note for the reporter

The earlier "still doing it" repro ran on the **pre-fix binary** (session ended ~14:41; v4.44.2 with #288 installed 14:43). Worth retesting `opus_1m` + thinking + parallel tools on the current build with both fixes in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ auto-release workflow to deduplicate commits across successive runs.
 
 ## [Unreleased]
 
+### Fixed
+- recover from a wedged session when the API rejects a continuation request (a rejected post-tool-use request no longer permanently poisons the reused messages array)
+
 ## [4.44.2] - 2026-06-26
 
 ### Fixed
--- a/src/agent/providers/anthropic-direct/query.ts
+++ b/src/agent/providers/anthropic-direct/query.ts
@@ -337,6 +337,18 @@ export class AnthropicDirectQuery implements ProviderQuery {
         // either shape natively.
         this.state.messages.push({ role: 'user', content: turn.content });
 
+        // Invariant: the messages array up to and including the just-pushed
+        // user turn is a known-sendable prefix — every prior assistant turn
+        // was accepted by the API at least once, and a bare user turn is
+        // always appendable. The tool-use loop will push assistant/tool_result
+        // turns past this point; if the turn ends in an `error` (e.g. the API
+        // rejects a continuation request), those pushes are left behind in the
+        // reused array and every later turn re-sends — and re-fails on — the
+        // same rejected history, permanently wedging the session. Capturing
+        // the prefix length here lets the error path below truncate back to it
+        // so a failed turn degrades to a recoverable per-turn error instead.
+        const sendablePrefixLen = this.state.messages.length;
+
         const system = this.composeSystem();
         const headers = buildRequestHeaders(
           this.retry.authMode,
@@ -372,6 +384,11 @@ export class AnthropicDirectQuery implements ProviderQuery {
         // `providerIterator.next()` loop is never advanced past a turn it
         // already saw end.
         let turnEmittedTerminal = false;
+        // Tracks whether THIS turn surfaced an `error` event (as opposed to a
+        // clean `turn.completed`). Drives the post-turn history rollback that
+        // prevents a single rejected request from permanently wedging the
+        // session — see `sendablePrefixLen` above.
+        let turnErrored = false;
         try {
           for await (const event of this.retry.turnWithRetries(runInput, () => this.state.closed)) {
             if (this.state.closed) return;
@@ -389,6 +406,9 @@ export class AnthropicDirectQuery implements ProviderQuery {
             }
             if (event.type === 'turn.completed' || event.type === 'error') {
               turnEmittedTerminal = true;
+            }
+            if (event.type === 'error') {
+              turnErrored = true;
             }
             yield event;
           }
@@ -435,6 +455,35 @@ export class AnthropicDirectQuery implements ProviderQuery {
         if (controller.signal.aborted) {
           if (!turnEmittedTerminal) yield this.makeInterruptedTurnEvent();
           continue;
+        }
+
+        // Wedge recovery: a turn that surfaced an `error` (e.g. the API
+        // rejected a continuation request — a malformed thinking block under
+        // interleaved thinking, an over-budget payload, a transient 5xx not
+        // absorbed by the retry layer) leaves its partial assistant/tool_result
+        // pushes in the reused messages array. The loop's own rollback
+        // (`messagesRollbackIdx`) only covers throws DURING tool execution, not
+        // the next iteration's request failure, and `repairOrphanToolUses` only
+        // heals orphan `tool_use` blocks — neither removes an assistant turn the
+        // API keeps rejecting. Without this, every later prompt re-sends the
+        // same rejected history and re-fails: a permanent session wedge where
+        // the user "can't send anything else". Truncating back to the
+        // known-sendable prefix (history through the triggering user turn)
+        // converts that brick into a recoverable per-turn error — the next
+        // prompt starts from a prefix the API has already accepted. Gated on
+        // !aborted (handled above) so an interrupt's partial state is preserved
+        // for resume; `splice` mutates in place per the never-reassign invariant
+        // on `state.messages`.
+        //
+        // Tradeoff: for a multi-iteration tool turn, this drops ALL of the
+        // turn's assistant/tool_result pairs, not just the failing last one —
+        // including iterations that executed successfully. That is intentional:
+        // the turn produced no final answer (it ended in error), and reverting
+        // to the pre-turn prefix is the only target GUARANTEED sendable, whereas
+        // keeping earlier iterations risks leaving the very block the API keeps
+        // rejecting. Recovering the session beats preserving partial tool churn.
+        if (turnErrored && this.state.messages.length > sendablePrefixLen) {
+          this.state.messages.splice(sendablePrefixLen);
         }
 
         // Auto-compaction: fire at the natural turn boundary — after the
--- a/src/agent/providers/anthropic-direct/wedge-recovery.test.ts
+++ b/src/agent/providers/anthropic-direct/wedge-recovery.test.ts
@@ -1,0 +1,369 @@
+/**
+ * Regression test for the "wedged session after a rejected continuation
+ * request" bug.
+ *
+ * Repro: a turn that uses tools pushes its `[assistant(thinking+tool_use),
+ * tool_result]` turns into the per-session `messages` array — which is mutated
+ * in place and REUSED across turns. If the FOLLOW-UP continuation request is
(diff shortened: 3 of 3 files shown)


griffinwork40 (commented):

**Merge decision: ✅ MERGE** (no blocking, critical, or high findings; 1 medium + 2 low + 1 nit follow-ups, all non-blocking)

Surgical wedge-recovery hotfix. Verified against branch HEAD `a974986` (+421/−0, 3 files: CHANGELOG entry, ~15 lines of production logic in `query.ts`, and a 369-line regression test). The fix is correctly wired to the **real** production error path — independently confirmed below — satisfies its stated intent, and changes no exported contract. No critical/high finding survived citation + absence re-verification at the reviewed ref.

### What's solid (verified at ref `a974986`)
- **The fix actually fires on the production wedge path.** When the continuation request is rejected, `messages.create` throws; `loop.ts:282-296` catches it and (when `!signal.aborted`) **yields** `{type:'error'}` then `return`s — it does **not** rethrow. So `query.ts`'s `for await` loop receives the yielded error, sets `turnErrored=true` (`query.ts:410-412`), exits cleanly, and falls through to the splice at `query.ts:485-487`. The recovery net is on the live path, not a dead branch.
- **Prefix boundary is correct.** `sendablePrefixLen` is captured **once per outer turn** at `query.ts:350`, immediately after the new user turn is pushed and **before** any tool-use/`tool_result` pushes — never re-captured mid-iteration. So a multi-iteration tool turn truncates back to the pre-turn boundary, which is the only target guaranteed sendable (as the PR documents).
- **Next-request pairing contract preserved.** The splice lands on a prefix ending in a `user` turn; `repairOrphanToolUses` runs at the top of the next iteration (`query.ts:333`) **before** the next push, healing any orphan `tool_use` left behind. No dangling-`tool_use` corruption on the subsequent request.
- **Abort/interrupt state preserved.** Two `if (controller.signal.aborted) { …; continue; }` guards (`query.ts:427-435` in the catch, `query.ts:454-458` after the loop) sit before the splice, so an interrupted turn never truncates — honoring the stated "preserve partial state for resume" intent.
- **Splice-then-compact ordering is safe.** Auto-compaction runs right after (`query.ts:489+`); on a freshly-truncated short array `compact-handler` returns `'history-too-short'` and skips cleanly — no error, no double-mutation hazard.
- **No contract / security regression.** `AnthropicDirectQuery` and `AnthropicDirectQueryOptions` are the only exported symbols and neither signature changed; the new identifiers are function-scoped locals and the splice is an internal `state.messages` mutation (in-place, honoring the never-reassign invariant). No auth/token/secret surface touched; no injection vector (the spliced array holds already-sent `MessageParam` content).
- **Regression test is genuine red→green.** `wedge-recovery.test.ts` drives the **real** `AnthropicDirectQuery` generator (not a mock-provider shortcut) through `tool-use turn → rejected continuation → next prompt`. The decisive assertions — `hasAssistantToolUse(turn2Sent) === false` and roles `['user','assistant','user','user']` — fail pre-fix (poisoned `assistant(tool_use)` re-sent) and pass post-fix.

### Findings (none blocking)

**1. medium · correctness · `src/agent/providers/anthropic-direct/query.ts:415-438` · ref:a974986 · file-state** — The thrown-error `catch` block yields an error then `return`s (`query.ts:438`), **skipping** the splice below it. This is **harmless today**: that path is reached only when a tool-dispatch throw escapes `loop.ts` (`loop.ts:700`), and `loop.ts:699` has already executed `input.messages.splice(messagesRollbackIdx)` to roll history back before the rethrow — so history is already consistent and the wedge is unreachable here. The risk is purely latent: no comment documents *why* the splice is intentionally absent in this catch, so a future refactor routing a new error type through it could silently re-introduce the wedge.
  - _Evidence:_ `query.ts:437-438` `yield { type: 'error', error: e }; return;` (no splice); `loop.ts:699-700` `input.messages.splice(messagesRollbackIdx); throw err;`
  - _Fix:_ add a one-line comment at `query.ts:438` noting the thrown path is already history-clean (rolled back at `loop.ts:699`), and that any new throw path must either roll back in `loop.ts` or fall through to the splice. No code change required today.

**2. low · test-coverage · `src/agent/providers/anthropic-direct/wedge-recovery.test.ts` · ref:a974986 · file-state** — No test covers the abort invariant. The PR claims the `!aborted` gate preserves an interrupted turn's partial state for resume, but the suppression-on-abort behavior (`query.ts:454-458`) is untested. (Verified: `grep -in "abort\|signal\|interrupt" wedge-recovery.test.ts` → 0 matches at ref.)
  - _Fix:_ add a test that injects a tool-use stream, aborts the signal mid-turn, and asserts `state.messages` still contains the partial `assistant`/`tool_result` pairs (no splice occurred).

**3. low · test-coverage · `src/agent/providers/anthropic-direct/wedge-recovery.test.ts` · ref:a974986 · file-state** — No test covers a multi-iteration tool turn. Both `it()` cases use exactly one tool-use round (`callIdx===1` → tool-use, `callIdx===2` → rejection). The PR explicitly documents (`query.ts:478-484`) that the splice drops **all** of a multi-round turn's pairs, including successful earlier rounds — but that intentional behavior has no test confirming both rounds are dropped. (Verified: only the line-12 doc comment matches `iteration`; no second tool-use round exists in either mock sequence.)
  - _Fix:_ add a test with `callIdx` 1→tool-use, 2→tool-use again, 3→rejection, 4→capture; assert `hasAssistantToolUse(turn3Sent) === false` across both prior rounds.

**4. nit · perf-observability · `src/agent/providers/anthropic-direct/query.ts:485-487` · ref:a974986 · file-state** — The splice fires silently — no trace event is emitted when history is truncated. A mid-session truncation dropping an unknown number of messages is operationally invisible; post-incident diagnosis must infer it from token-count deltas.
  - _Evidence:_ `query.ts:486` `this.state.messages.splice(sendablePrefixLen);` — no preceding `emitSessionPhase`/`traceWriter` call in the block.
  - _Fix (optional):_ emit a fire-and-forget `session_phase` event (e.g. `phase:'wedge_recovery'`, `droppedMessages: this.state.messages.length - sendablePrefixLen`) before the splice, mirroring the `model_ttfb` pattern at `loop.ts:319-327`.

### Spec-compliance (assessed against PR #289 title + body) — PASS
Wedge recovery on rejected continuation, truncate to the known-sendable prefix, `!aborted` gate, in-place `splice` (never-reassign invariant), and the explicit scope-out (does **not** attempt to fix the underlying API rejection — defense-in-depth alongside #288). All present, no scope creep. Findings 1-4 are documentation/test-coverage/observability consistency items the intent implies but the diff did not carry; none contradict the stated scope.

### What was not checked
Read against branch HEAD `a974986` — every citation and absence claim in the findings above was independently re-verified at that ref (Check A: all citations `verified`; Check B: both test-absence claims `confirmed-absent`, zero matches). No fabricated or false-absence finding survived. Did not run the test suite locally (the PR reports `pnpm lint` clean, `pnpm audit:sdk:check` OK, and all 329 `anthropic-direct/` provider tests passing, with the regression verified red→green). Did not exercise the wedge at runtime against the live Anthropic API, and — consistent with the PR's own "honest scope" note — did not attempt to confirm the exact block type the API rejected in the original incident (not persisted to telemetry; needs a live repro with the thinking-block diagnostic captured). No critical/high finding present, so the adversarial shadow-verify wave was not triggered.

griffinwork40:
Update after runtime instrumentation (see #290): the reported wedge is **not** what this PR assumed. The actual cause is `stop_reason: "refusal"` — the continuation request is well-formed and succeeds, but the model declines on content-safety grounds and returns an empty turn, which afk rendered silently. No error event fires, so this PR's error-rollback never engages on the reported case. #290 is the real fix. This PR is still a valid *defensive* improvement (an error'd continuation shouldn't poison the reused messages array), but it's mislabeled as the wedge fix — recommend re-scoping the title/description or closing. Deferring to you.

PR #290, opened 2026-06-26
fix(anthropic-direct): surface model refusals instead of ending the turn silently

griffinwork40:
## The actual root cause (runtime-confirmed)

The reported bug: opus_1m researches a topic (web_scrape), then **stops with no answer**, and afterward **every message dies instantly** — "it stopped and I can't send anything else."

I instrumented a build the reporter ran. The debug log is decisive:

```
REQUEST round 1:
  user[text,image]
  assistant[thinking(sig:1416), text, tool_use×4]   ← well-formed, thinking signature present
  user[tool_result×4]                                ← tool_use matched
TURN_COMPLETE_NO_TOOLUSE:
  stopReason: "refusal", textLen: 0, blockTypes: []  ← the model declined, returned nothing
```

So the continuation request is **well-formed and succeeds** — it returns `stop_reason: "refusal"` (Anthropic's content-safety stop) with an **empty** assistant turn. `loop.ts` handled `stop_reason != 'tool_use'` generically, and with empty text it yielded **nothing** → the turn ended **silently**, indistinguishable from a hang. And because the flagged content stays in the conversation, every later turn refuses identically → "can't send anything else."

## Correcting the record on #288 / #289

This **supersedes the hypothesis** behind #288 (preserve redacted_thinking) and #289 (roll back error'd turns): a malformed-thinking-block 400 + reused-`messages` poisoning. The runtime evidence does **not** support it — there is **no error event** on a refusal (the request succeeds), so neither PR addresses the reported bug. Those remain valid *defensive* improvements for genuine error/redacted-thinking cases, but they are **not** the wedge fix. (I'd suggest re-scoping or closing #289 — happy to do either.)

## The fix

Detect `stop_reason: "refusal"` in the completion branch and surface an explicit, **display-only** notice (not pushed to history — it's operator-facing, not model context) explaining the model declined and that follow-ups will likely also be refused until the context is cleared. Covers both an initial-request refusal and a post-tool-use continuation refusal. It does **not** try to override the model's safety decision — only to make it legible instead of a silent hang.

## Tests

- New `loop.test.ts` case: a refusal stream (empty content, `stop_reason: "refusal"`) now yields a visible `assistant.message` and completes cleanly **without** pushing the empty turn. Verified **red→green** (fails with the branch disabled).
- `pnpm lint` clean · `pnpm audit:sdk:check` OK · all **328** `anthropic-direct/` provider tests pass.

- Consider generalizing to surface *any* empty, non-tool completion (e.g. an unexpected bare `end_turn` with no content) so no stop reason can ever end a turn silently.
- Decide the fate of #289 (defensive, but mislabeled as the wedge fix).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ auto-release workflow to deduplicate commits across successive runs.
 
 ## [Unreleased]
 
+### Fixed
+- surface model content-safety refusals (stop_reason "refusal") instead of ending the turn silently — fixes the "it stopped and I can't send anything else" hang when the model declines a request
+
 ## [4.44.2] - 2026-06-26
 
 ### Fixed
--- a/src/agent/providers/anthropic-direct/loop.test.ts
+++ b/src/agent/providers/anthropic-direct/loop.test.ts
@@ -217,6 +217,67 @@ describe('loop.ts runTurn', () => {
     expect(DEFAULT_MAX_TOOL_USE_ITERATIONS).toBe(0);
   });
 
+  it('surfaces a visible notice when the model stops with stop_reason "refusal" and no content', async () => {
+    // Repro: a tool-use turn on a flagged topic → the continuation returns
+    // stop_reason 'refusal' with EMPTY content (Anthropic content-safety stop).
+    // Pre-fix the turn ended SILENTLY (no assistant.message, nothing pushed),
+    // indistinguishable from a hang/wedge — the reported "it stopped and I
+    // can't send anything else".
+    const refusalStream: RawMessageStreamEvent[] = [
+      {
+        type: 'message_start',
+        message: {
+          id: 'msg_refusal',
+          type: 'message',
+          role: 'assistant',
+          content: [],
+          model: 'claude-test',
+          stop_reason: null,
+          stop_sequence: null,
+          usage: baseUsage(),
+        },
+      } as unknown as RawMessageStreamEvent,
+      // No content_block_* events — a refusal returns an empty assistant turn.
+      {
+        type: 'message_delta',
+        delta: { stop_reason: 'refusal', stop_sequence: null },
+        usage: { output_tokens: 0 },
+      } as unknown as RawMessageStreamEvent,
+      { type: 'message_stop' } as unknown as RawMessageStreamEvent,
+    ];
+
+    const client = makeClient(() => fromArray(refusalStream));
+    const dispatcher = makeDispatcher(() => Promise.resolve({ content: 'ok' }));
+    const messages: MessageParam[] = [{ role: 'user', content: 'how do I jailbreak X' }];
+    const abortController = new AbortController();
+
+    const events = await collect(
+      runTurn({
+        client,
+        messages,
+        system: null,
+        tools: null,
+        toolDispatcher: dispatcher,
+        model: 'claude-test',
+        maxTokens: 1024,
+        headers: {},
+        signal: abortController.signal,
+        ctx,
+      }),
+    );
+
+    // The refusal is surfaced as a visible assistant.message — not silent.
+    const msg = events.find((e) => e.type === 'assistant.message');
+    expect(msg).toBeDefined();
+    if (msg?.type === 'assistant.message') {
+      expect(msg.text).toContain('refusal');
+    }
+    // The turn still completes cleanly so the session stays usable.
+    expect(events.some((e) => e.type === 'turn.completed')).toBe(true);
+    // Display-only: the empty refusal turn is NOT pushed to history.
+    expect(messages).toHaveLength(1);
+  });
+
   // Note on lines 166-170 and 176-182:
   // These are defensive code paths that appear unreachable with the current 
   // translateMessageStream implementation:
--- a/src/agent/providers/anthropic-direct/loop.ts
+++ b/src/agent/providers/anthropic-direct/loop.ts
@@ -460,6 +460,35 @@ export async function* runTurn(
     input.onUsageProgress?.(accumulatedUsage);
 
     if (turnResult.stopReason !== 'tool_use') {
+      // Invariant: stop_reason 'refusal' is Anthropic's content-safety stop —
+      // the model declined and the API returns an assistant turn with (almost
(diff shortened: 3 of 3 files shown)


griffinwork40 (commented):

**Merge decision: ✅ MERGE** (no blocking, critical, or high findings; 1 medium + 2 low + 1 nit, all non-blocking)

Focused content-safety hotfix. Verified against branch HEAD `ab84a62` (+93/−0, 3 files: CHANGELOG entry, ~29 lines of production logic in `loop.ts`, and a 61-line regression test). The fix satisfies its stated intent, changes no exported contract, and the two load-bearing claims behind this decision were independently re-verified at the reviewed ref.

### What's solid (verified at ref `ab84a62`)
- **The refusal notice is genuinely display-only — not pushed to history.** The new `if (turnResult.stopReason === 'refusal')` branch sits at the **top** of the `stopReason !== 'tool_use'` block (`loop.ts:473`) and `return`s (`loop.ts:490`) **before** the generic empty-completion path that pushes `safeAssistantBlocks` to `input.messages` (`loop.ts:520`). So the empty refusal turn never enters history — exactly the documented intent. The regression test confirms it: `expect(messages).toHaveLength(1)` after a refusal stream.
- **`'refusal'` is a live, well-typed value — not dead code.** `turnResult.stopReason` is typed `string | null` (`types.ts:163`) and is assigned directly from the raw stream value `evt.delta.stop_reason` (`translate.ts:249`), so the Anthropic content-safety `"refusal"` stop round-trips to this branch verbatim. (`@anthropic-ai/sdk@^0.74.0` lists `refusal` in its `StopReason` union.) The branch is reachable.
- **Both refusal paths are covered by the code.** `runTurn` is a single `while` loop, so the branch is reached identically whether the refusal arrives on the **initial** request (iteration 1) or a **post-tool-use continuation** (iteration ≥2). The intent's "covers both" claim holds at the implementation level.
- **No contract / security regression.** No exported signature changed; the generator's `AsyncGenerator<ProviderEvent>` return type is unaffected. The notice text is a hardcoded string plus `turnResult.text` (model output, never user input) — no injection surface, no secret/auth/RCE vector. The fix does not override the model's safety decision; it only makes the stop legible.
- **Regression test is a genuine repro.** It drives a refusal stream (empty content, `stop_reason: 'refusal'`) and asserts a visible `assistant.message` containing "refusal", a clean `turn.completed`, and `messages` length unchanged — the three conditions the intent names.

### Findings (none blocking)

**1. medium · test-coverage · `src/agent/providers/anthropic-direct/loop.test.ts:220-284` · ref:ab84a62 · diff-context** — The single test exercises only an **iteration-1** refusal (no prior tool-use round); the **post-tool-use continuation** case — which the PR body names as the *primary* reported scenario ("researches a topic (web_scrape), then stops") — is not independently exercised. The branch is identical for both iterations so the code is correct, but iteration ≥2 is mechanically unverified.
  - _Evidence:_ the `refusalStream` fixture contains no `tool_use`/`content_block_*` events before the `message_delta { stop_reason: 'refusal' }`, so only the first loop round runs.
  - _Fix:_ add a second `it()` that completes one tool-call round, then feeds a refusal on the next iteration, asserting the notice surfaces and `messages.length === 3` (user + assistant-with-tools + tool_result) with no extra push.

**2. low · test-coverage · `src/agent/providers/anthropic-direct/loop.test.ts:270-284` · ref:ab84a62 · diff-context** — The non-empty-text refusal path is untested. The branch passes `turnResult.text` through verbatim when `text.length > 0` (`loop.ts:477-479`), but no test covers a refusal that returns partial text — and the assertion `toContain('refusal')` would then pass or fail purely on whether the model's own text happens to contain that word.
  - _Fix:_ add a variant with a `content_block_start/delta/stop` sequence before the `refusal` `message_delta`, asserting the model's actual text is emitted (not the fallback notice) and `messages.length` is unchanged.

**3. low · perf-observability · `src/agent/providers/anthropic-direct/loop.ts:473-490` · ref:ab84a62 · diff-context** — The refusal branch emits no dedicated trace/witness event, so a content-safety stop is not directly distinguishable from a normal `end_turn` in the witness layer beyond the `stopReason: 'refusal'` field on the `turn.completed` usage blob. Refusal incidents are queryable only by parsing usage objects, not by a phase event.
  - _Fix (optional):_ emit a fire-and-forget session-phase event (e.g. `phase: 'refusal_stop'`) before the `return`, mirroring the existing `loop_start`/`loop_end` phase pattern, so refusals are first-class in the trace.

**4. nit · test-coverage · `src/agent/providers/anthropic-direct/loop.test.ts:269-280` · ref:ab84a62 · diff-context** — The test does not assert event **order** (that `assistant.message` precedes `turn.completed`); it uses flat `find`/`some`, so a future reorder would not be caught.
  - _Fix:_ assert `findIndex(assistant.message) < findIndex(turn.completed)`.

### Spec-compliance (assessed against PR #290 title + body) — PASS
Display-only / not-pushed-to-history (verified: `return` at `loop.ts:490` precedes the push at `:520`; test asserts `messages` length unchanged) ✓ · covers both initial-request and post-tool-use continuation refusals (single-loop reachability) ✓ · `'refusal'` is a live well-typed stop value, not dead code ✓ · does not override the safety decision (notice only) ✓ · regression test asserts message-visible + turn-completes + nothing-pushed ✓ · no scope creep. Findings 1-4 are test-coverage/observability consistency items the intent implies but the diff did not fully carry; none contradict the stated scope.

### What was not checked
Read against branch HEAD `ab84a62` — every citation above was read from that ref. The two decisive claims (notice not pushed to history; `'refusal'` reachable, not dead code) were independently re-verified at the ref from the production source (`loop.ts` control flow + `types.ts:163` / `translate.ts:249`), not taken from the agent's report alone. Did not run the test suite locally (the PR reports `pnpm lint` clean, `pnpm audit:sdk:check` OK, and all 328 `anthropic-direct/` provider tests passing, with the regression verified red→green). Did not exercise the refusal against the live Anthropic API, and did not independently confirm whether `stop_reason: "refusal"` can ever return *non-empty* content (the fix handles both cases regardless). No critical/high finding present, so the adversarial shadow-verify wave was not triggered.

PR #292, opened 2026-06-26
fix(afk): don't fire path-approval prompts in AFK mode

griffinwork40:
## Summary

Fixes AFK mode firing keyboard-blocking path-approval prompts — the bug that defeats "away from keyboard."

- AFK (`autonomous`) set the dispatcher `allowAll=false`, re-enabling typed-file path-approval elicitations that block on a keyboard the operator has walked away from — making AFK *stricter* than the new-install default (`bypassPermissions`).
- New `src/agent/permission-policy.ts` exports `pathContainmentBypassed(mode)` (true for `bypassPermissions` **and** `autonomous`) as the single source of truth, applied at all six `allowAll` derivation sites across both providers, so the dispatcher flag and `getGrants().allowAll` stay in lockstep (divergence fails unsafe).
- Safety is unchanged: the `afk-mode-gate` risk ceiling still refuses high-risk / irreversible ops (destructive bash, write-denylist, `.git`-store, workspace-escape writes) **independently** of `allowAll`.
- Hardening: the gate — now the sole path-safety layer in AFK — takes a live `getCwd` (the same getter path-approval already uses, wired from the REPL bootstrap) so its workspace-escape boundary tracks a mid-session `/cwd` change.

Net effect: **AFK = `bypassPermissions`-level path freedom + the risk gate.** Behavior in `default` / `plan` / `bypassPermissions` is unchanged.

## Test results

- `pnpm lint` (`tsc --noEmit`, strict) → clean.
- `pnpm exec vitest run` (full suite, rebased onto `v4.45.0`) → **10137 passed / 14 skipped / 1 failed**.
- The single failure — `anthropic-direct.test.ts > compact() summarizes older history…` — is **pre-existing and unrelated**: it fails identically on clean HEAD with these edits `git stash`ed, and the diff touches no compaction code.
- Touched-area focus (5 files) → **123 passed**: `permission-policy.test.ts` (new), `afk-mode-gate.test.ts` (incl. new `getCwd`-precedence test), `bypass-live-toggle.test.ts` (autonomous ⇒ allowAll on both enforcement reads), `openai-compatible/query.test.ts` (OpenAI parity), `path-approval-hook.test.ts` (no regression).

## Verification

Adversarial re-derivation from the diff alone (independent read-only pass):

- **C1 CONFIRM** — `pathContainmentBypassed` (`permission-policy.ts:39`) returns true for both modes; applied at all 6 sites (`anthropic-direct/index.ts:369,512`, `query.ts:561`; `openai-compatible/index.ts:408,467`, `query.ts:1211`); no bare `=== 'bypassPermissions'` remains at any `allowAll` site.
- **C2 CONFIRM** — `afk-mode-gate.ts` blocks `classifyRisk()==='high'` independently of `allowAll`/grants; only a `getCwd` param was added.
- **C3 CONFIRM** — both enforcement reads per provider derive through the same predicate; no drift.
- Flagged residual (resolved): the `getCwd` wired in `default-hook-registry.ts` is the REPL bootstrap's live `() => extras?.cwd ?? process.cwd()` (`bootstrap.ts:568`) — the same getter path-approval uses — so the gate's workspace boundary stays live, not stale.

## Out of scope

- Phase 2 "two-way AFK" (inbound Telegram replies steering a live run; high-risk-gate → phone approve/deny) is **deferred** — recon found the elicitation/abort round-trip is already ~90% built (auto-subscribe + ledger channel + `/abort` + HMAC). Roadmap captured in the local plan doc (`.afk/plans/…`, gitignored).

### DIFF
--- a/src/agent/afk-mode-gate.test.ts
+++ b/src/agent/afk-mode-gate.test.ts
@@ -102,6 +102,31 @@ describe('createAfkModeGate', () => {
     expect(result.decision).toBeUndefined();
   });
 
+  it('prefers a live getCwd() over the static cwd for the workspace-escape check', () => {
+    // Static cwd is the project root, but the live getCwd() reports the session
+    // moved into a deeper subdir. A write into the project root (inside the
+    // STATIC cwd) now escapes the LIVE workspace and must be blocked — proving
+    // the gate reads getCwd() first. In AFK this gate is the sole path-safety
+    // layer (path-approval is disabled via allowAll), so this must stay live.
+    const gate = createAfkModeGate(
+      () => 'autonomous' as PermissionMode,
+      '/Users/dev/project',
+      () => '/Users/dev/project/sub',
+    );
+    const escaping = gate({
+      event: 'PreToolUse',
+      toolName: 'write_file',
+      input: { file_path: '/Users/dev/project/feature.ts' },
+    });
+    expect(escaping.decision).toBe('block');
+    const inside = gate({
+      event: 'PreToolUse',
+      toolName: 'write_file',
+      input: { file_path: '/Users/dev/project/sub/feature.ts' },
+    });
+    expect(inside.decision).toBeUndefined();
+  });
+
   // ---- send_telegram is the channel: always exempt --------------------------
   it('never blocks send_telegram in AFK mode (it is the operator channel)', () => {
     const { gate } = makeGate('autonomous');
--- a/src/agent/afk-mode-gate.ts
+++ b/src/agent/afk-mode-gate.ts
@@ -25,6 +25,14 @@
  * and `'safe'` ops (reads, tests, lint) are ALLOWED — autonomous work has to be
  * useful, and these are reversible enough to run unattended.
  *
+ * Path-safety note: in `'autonomous'` mode the typed-file path-approval prompt
+ * is disabled (`allowAll`, see `agent/permission-policy.ts`) so AFK does not
+ * stall on keyboard prompts the operator cannot answer. That makes THIS gate
+ * the sole path-containment layer in AFK — its workspace-escape rule (a
+ * `write_file`/`edit_file` whose path resolves outside the session root →
+ * `high` → blocked) is load-bearing, which is why the gate is wired with a
+ * live `getCwd` (so a mid-session `/cwd` change cannot stale the boundary).
+ *
  * `send_telegram` is ALWAYS exempt: it is the operator's channel in AFK mode,
  * and the posture explicitly relies on it to surface Asking states.
  *
@@ -51,6 +59,7 @@ import { classifyRisk } from './risk-classifier.js';
 export function createAfkModeGate(
   getMode: () => PermissionMode,
   cwd?: string,
+  getCwd?: () => string | undefined,
 ): (context: HookContext) => HookDecision {
   return function afkModeGate(context: HookContext): HookDecision {
     if (context.event !== 'PreToolUse') return {};
@@ -65,8 +74,11 @@ export function createAfkModeGate(
 
     // Single source of truth for risk. `workspaceRoot` is set to the session
     // cwd so writes that escape it are flagged `high` (classifyRisk's workspace
-    // boundary rule); falling back to process.cwd() when unknown.
-    const root = cwd ?? process.cwd();
+    // boundary rule). Prefer the live getCwd() (tracks a mid-session /cwd
+    // change) over the static construction-time cwd; fall back to process.cwd()
+    // when both are unknown. This matters because in AFK the path-approval
+    // prompt is disabled (allowAll), so this gate is the SOLE path-safety layer.
+    const root = getCwd?.() ?? cwd ?? process.cwd();
     const risk = classifyRisk(toolName, context.input, {
       cwd: root,
       workspaceRoot: root,
--- a/src/agent/default-hook-registry.ts
+++ b/src/agent/default-hook-registry.ts
@@ -93,7 +93,11 @@ export function createDefaultHookRegistry(
     // one applies tree-wide (no subagent exemption) — see afk-mode-gate.ts.
     registry.register(
       'PreToolUse',
-      createAfkModeGate(getPermissionMode, agentOptions?.cwd),
+      // getCwd (live) is preferred over the static agentOptions.cwd so the
+      // gate's workspace-escape rule tracks a mid-session /cwd change — it is
+      // the SOLE path-safety layer in AFK (the path-approval prompt is disabled
+      // via allowAll for 'autonomous'; see agent/permission-policy.ts).
+      createAfkModeGate(getPermissionMode, agentOptions?.cwd, getCwd),
     );
   }
 
--- a/src/agent/permission-policy.test.ts
+++ b/src/agent/permission-policy.test.ts
@@ -1,0 +1,25 @@
+import { describe, it, expect } from 'vitest';
+import { pathContainmentBypassed } from './permission-policy.js';
+import type { PermissionMode } from './types/sdk-types.js';
+
+describe('pathContainmentBypassed', () => {
+  it('returns true for bypassPermissions (explicit full power)', () => {
+    expect(pathContainmentBypassed('bypassPermissions')).toBe(true);
+  });
+
+  it('returns true for autonomous (AFK — must not fire keyboard path prompts)', () => {
+    expect(pathContainmentBypassed('autonomous')).toBe(true);
+  });
+
+  it.each<PermissionMode>(['default', 'plan', 'acceptEdits', 'dontAsk', 'auto'])(
+    'returns false for containment-preserving mode %s',
+    (mode) => {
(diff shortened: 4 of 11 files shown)


griffinwork40 (commented):

**Merge decision: ✅ MERGE** — no critical, high, or blocking findings. 1 medium (defense-in-depth, requires-miswiring) + 2 low + informational follow-ups.

Reviewed against branch HEAD `d51a260` (change type: hotfix; 11 files, +169/-16). The central safety claim was independently re-verified at the reviewed ref: the `afk-mode-gate` blocks high-risk ops via `classifyRisk` **without** consulting `allowAll`/grants, and the load-bearing safety behaviors (destructive bash blocked + workspace-escape write blocked under `autonomous`) are genuinely test-covered. Every `file:line` citation below was read at `d51a260` via `gh api …/contents?ref=d51a260`; no fabricated or false-absence finding survived.

### What's solid (verified at ref `d51a260`)
- **All six `allowAll` sites repointed — no drift.** Two independent passes confirmed zero bare `=== 'bypassPermissions'` survives at any `allowAll` derivation site (every remaining occurrence is a comment): `anthropic-direct/index.ts:369,512`, `anthropic-direct/query.ts:561`, `openai-compatible/index.ts:408,467`, `openai-compatible/query.ts:1211`. The dispatcher flag and `getGrants().allowAll` both derive through `pathContainmentBypassed`, so they cannot drift out of lockstep.
- **The gate is independently load-bearing.** `afk-mode-gate.ts` reads neither `allowAll` nor grants — it calls only `classifyRisk(toolName, input, { cwd: root, workspaceRoot: root })`, and `classifyRisk` is a pure function with no dispatcher/grants import. Widening `allowAll` to include `autonomous` does not weaken the risk ceiling. ✓
- **Symlink / `..` traversal is caught.** `write-denylist.ts`'s `safeRealpath` dereferences symlinks (walks the parent chain via `realpathSync`) before the `rel.startsWith('..')` workspace-escape check, so `/project/link → /etc` is still classified `high`. ✓
- **Safety invariant is test-covered (the important one).** `afk-mode-gate.test.ts` at the ref asserts, under `autonomous`: destructive bash (`rm -rf …`, incl. a subagent `rm -rf /` case) → `decision: 'block'`; a workspace-escaping write (`/etc/cron.d/evil`) → `decision: 'block'`; and the new live-`getCwd` override narrows the boundary correctly. `bypass-live-toggle.test.ts` + `openai-compatible/query.test.ts` assert `autonomous ⇒ allowAll=true` on **both** enforcement reads, `default`/`plan ⇒ false`. `permission-policy.test.ts` covers every named mode + `undefined` + `'nonsense'`. ✓
- **`??` fallback is correct**, not `||`: `getCwd?.() ?? cwd ?? process.cwd()` falls back only on `undefined`/`null`, preserving a legitimately-defined live cwd. **Backward compatible**: `createAfkModeGate`'s new 3rd param `getCwd?` is optional; the sole production caller (`default-hook-registry.ts`) passes it. No public-entrypoint surface changed.

### Findings (none blocking)

**1. medium · security · `src/agent/afk-mode-gate.ts:74` (+ `risk-classifier.ts` workspace-escape rule) · ref:d51a260 · file-state** — The gate passes its resolved `root` as **both** `cwd` and `workspaceRoot` to `classifyRisk`, and `root = getCwd?.() ?? cwd ?? process.cwd()`. If a `getCwd` ever returns `'/'` (or `''`), `workspaceRoot` becomes the filesystem root and the escape rule no-ops: `path.relative('/', '/etc/passwd')` → `'etc/passwd'`, which does not start with `'..'`, so an out-of-root write is classified non-`high` and passes under `allowAll=true`. **Reachability (verified): requires-miswiring** — the production `getCwd` traces through `default-hook-registry.ts` to the REPL/Telegram bootstrap's live session-cwd accessor, which returns a worktree path, never `'/'`. So this is a robustness / defense-in-depth gap on a load-bearing layer, not a reachable bypass today.
  _Evidence (`risk-classifier.ts`, file-state):_
  ```ts
  const realRoot = safeRealpath(ctx.workspaceRoot);
  const rel = path.relative(realRoot, resolved);
  if (rel.startsWith('..')) return 'high';
  ```
  _Fix:_ in `createAfkModeGate`, treat a live `getCwd()` of `'/'` or `''` as unusable and fall back to `cwd ?? process.cwd()`; or guard `classifyFilePath` to skip the boundary check (or fail closed) when `realRoot === '/'` rather than silently allowing every path. Since this gate is the *sole* path-safety layer in AFK, hardening the degenerate-root case is worth a follow-up.

**2. low · correctness · `openai-compatible/query.ts:1205-1211` vs `anthropic-direct/query.ts:557-561` · ref:d51a260 · file-state** — `setPermissionMode` is asymmetric across providers: the OpenAI path feeds `pathContainmentBypassed` the **normalized** mode (`this.currentPermissionMode = normalizePermissionMode(mode)`), while the Anthropic path feeds the **raw** `mode` param. **Verified real-but-latent**: `normalizePermissionMode` is identity on every defined string (it only maps `undefined → 'default'`), so `'autonomous'` round-trips unchanged and there is no behavioral divergence today; it would only diverge if `undefined` were ever passed on the Anthropic path.
  _Evidence (file-state):_
  ```ts
  // openai query.ts:1205,1211
  this.currentPermissionMode = normalizePermissionMode(mode);
  const allowAll = pathContainmentBypassed(this.currentPermissionMode);
  // anthropic query.ts:557,561 — raw, no normalize
  const allowAll = pathContainmentBypassed(mode);
  ```
  _Fix:_ pick one pattern and apply it symmetrically — either normalize before the predicate in both providers, or pass the raw `mode` in both.

**3. low · observability · `src/agent/tools/handlers/bash.ts` (`warnIfBypassPermissions`) · ref:d51a260 · file-state** — The `shell:true` security `console.warn` fires only for `permissionMode === 'bypassPermissions'`; after this PR, `autonomous` also runs with `allowAll=true` but emits no equivalent warning. Not a capability bypass (the gate's bash rules still apply), but the audit signal is now missing for the new high-privilege mode.
  _Evidence (file-state):_
  ```ts
  if (permissionMode === 'bypassPermissions') {
    console.warn('[security] bash handler: shell=true with bypassPermissions …');
  }
  ```
  _Fix:_ widen the guard to `if (pathContainmentBypassed(permissionMode))` so AFK gets the same warning.

**4. low · spec/test-coverage · `src/cli/commands/chat.ts:663-668` · ref:d51a260 · file-state** — The non-interactive `chat.ts` one-shot call site builds `createDefaultHookRegistry` with `getPermissionMode`/`getCwd` omitted, so the afk-mode-gate is never registered there. Likely correct (chat one-shot is not an AFK surface), but the PR states the gate is "the SOLE path-safety layer in AFK" without calling out this surface exclusion, and no test documents that `chat.ts` is intentionally unguarded (`git grep "chat.*autonomous" -- '*.test.ts'` → 0 matches).
  _Fix:_ add an in-source comment at that call site noting the gate is intentionally absent for the chat one-shot surface (mirroring the existing comment in `default-hook-registry.ts`), or confirm `chat.ts` is never reachable with `permissionMode:'autonomous'`.

### Spec-compliance (assessed against PR #292 title + body) — PASS
Single source of truth `pathContainmentBypassed` added and applied at all 6 sites ✓ · dispatcher flag + `getGrants().allowAll` in lockstep, no bare `=== 'bypassPermissions'` at any `allowAll` site ✓ · gate refuses `high`-risk ops independently of `allowAll` (verified by source read + tests) ✓ · live `getCwd` wired from the REPL/Telegram bootstrap getter ✓ · `default`/`plan`/`bypassPermissions` behavior unchanged; `acceptEdits`/`dontAsk`/`auto` correctly return `false` ✓ · Phase 2 "two-way AFK" deferred as stated ✓ · no scope creep. (Findings 1–4 are robustness/consistency/coverage items the intent implies but the diff did not carry; none contradict the stated scope.)

### What was not checked
Wave-1 agents read against branch HEAD `d51a260`; every medium-and-above citation and all absence claims were independently re-verified at that ref by a separate verification pass (the S-1 degenerate-root behavior, the bash-warning guard, the two safety tests' presence, and the provider asymmetry were each re-read/re-derived — `path.relative('/','/etc/passwd')` was computed directly and returns `'etc/passwd'`; no `'/'`-yielding production `getCwd` path was found, hence "requires-miswiring"). No critical/high finding was present, so `/shadow-verify` was not invoked (mediums and below pass through). Did not run the test suite locally — the PR reports `pnpm lint` clean and full-suite 10137 pass / 14 skip / 1 fail, where the single failure (`anthropic-direct.test.ts > compact()`) is pre-existing and unrelated (fails identically on clean HEAD); CI on `d51a260` shows Lint & Build, Test, Docs Site, and Publish Bundle Smoke all SUCCESS. Did not exercise the Telegram or daemon surfaces at runtime, did not verify the `threads/create-session-factory.ts` call site's `getCwd` wiring (returned 404 at the ref — likely a path not present on the public mirror), and did not audit pre-existing `bypassPermissions` call sites outside the six `allowAll` derivation points (e.g. `daemon/scheduler.ts`), which this PR does not touch.

PR #295, opened 2026-06-27
fix(tools): subset advertised tool schema to permission allowlist

griffinwork40:
## Problem

Phase-restricted subagent forks — mint's `spec`/`research`/`plan` phases (forked with `phaseRole: 'read-only'`) and recon forks — are advertised the **full** builtin tool schema in their `tools:` API payload (`bash`, `write_file`, `edit_file`, `web_scrape`, `browser_*`, …), even though their permission gate only allows a small allowlist (`READ_ONLY_PHASE_TOOLS` / `RECON_ALLOWED_TOOLS`).

The model sees `bash` in its toolbox, reasonably reaches for it during codebase recon (`git log`, `rg`, `find`, `cat`), and gets back `Tool "bash" is not in the configured allowlist` — a flat rejection with no redirect. It then wastes a turn or two re-routing to `read_file`/`glob`/`grep`. The advertised schema and the enforced allowlist were never kept in sync.

## Root cause

The allowlist (`permissions.allowedTools`) was enforced **only at call time** in `checkToolPermission` (`permissions.ts` → `dispatcher.ts`), and did **not** subset the schema sent to the model. `SessionToolDispatcher.toolDefs` returned `this.schemas` unconditionally, and both providers build their `tools:` array from that getter:

- anthropic-direct: `providers/anthropic-direct/index.ts` reads `queryDispatcher.toolDefs` → sent as `tools:`
- openai-compatible: `providers/openai-compatible/query.ts` duck-types `.toolDefs` → `openAITools` → sent as `tools:`

The same seam is already acknowledged in `tool-category.ts` for `get_runtime_state` ("a tool the schema offered but the allowlist rejected").

## Fix

One place — subset `SessionToolDispatcher.toolDefs` by the configured allowlist:

```ts
get toolDefs(): readonly AnthropicToolDef[] {
  const allowed = this.permissions?.allowedTools;
  if (!allowed) return this.schemas;
  const set = new Set(allowed);
  return this.schemas.filter((s) => set.has(s.name));
}
```

Because both providers read `tools:` from this getter, the single change covers **both** the anthropic-direct and openai-compatible paths, and **recon** forks (same `permissions.allowedTools` path) — no provider-level changes.

### Invariants preserved
- **Undefined allowlist → full schema** (the default unrestricted path is unchanged).
- **MCP tools stay visible**: the dispatcher's `permissions` is the post-`withMcpToolsAllowed` union, so MCP wire-names are already in the allowlist before subsetting.
- **Skill-dispatch strip** of `ask_question`/`terminal_font_size` composes correctly — it's an independent filter applied separately in each provider.
- **Nice side effect:** `getEnabledToolNames` (which feeds `get_runtime_state`) now reports only the tools a restricted fork can actually call, instead of over-reporting.

## Testing
- `src/agent/tools/dispatcher.test.ts`: 76 passed (1 existing test updated to the no-allowlist path; 3 new cases — full pass-through when undefined, subset when set, empty when no match).
- `subagent-phase-role.test.ts`, `mint.test.ts`, `awareness/tool.test.ts`: pass.
- Full suite: the only failure is the pre-existing `anthropic-direct.test.ts > compact()` test (unrelated to tool schemas), which fails identically on clean `main`.

## Risk
Low. Read-only getter, tiny array filter, no state mutation. The only behavior that changes is for sessions that already had an allowlist configured — which were already broken (shown tools they couldn't call).

### DIFF
--- a/src/agent/tools/dispatcher.test.ts
+++ b/src/agent/tools/dispatcher.test.ts
@@ -105,9 +105,46 @@ describe('SessionToolDispatcher', () => {
     expect(result.failureClass).toBe('abort');
   });
 
-  it('exposes toolDefs from schemas', () => {
-    const dispatcher = makeDispatcher();
+  it('exposes toolDefs from schemas (no allowlist = full pass-through)', () => {
+    // Pass undefined permissions so no allowlist is configured — full schema returned.
+    const dispatcher = makeDispatcher({ permissions: undefined });
     expect(dispatcher.toolDefs).toEqual(builtinToolSchemas);
+  });
+
+  describe('toolDefs allowlist subsetting', () => {
+    it('returns all schemas when no allowlist is configured (permissions undefined)', () => {
+      const dispatcher = new SessionToolDispatcher({
+        handlers: new Map(),
+        schemas: [...builtinToolSchemas],
+        // no permissions → undefined
+      });
+      expect(dispatcher.toolDefs).toEqual(builtinToolSchemas);
+    });
+
+    it('returns only allowlisted schemas when allowedTools is set', () => {
+      const bashSchema = builtinToolSchemas.find((s) => s.name === 'bash')!;
+      const readFileSchema = builtinToolSchemas.find((s) => s.name === 'read_file')!;
+      expect(bashSchema).toBeDefined();
+      expect(readFileSchema).toBeDefined();
+      const dispatcher = new SessionToolDispatcher({
+        handlers: new Map(),
+        schemas: [bashSchema, readFileSchema],
+        permissions: { allowedTools: ['read_file'] },
+      });
+      const defs = dispatcher.toolDefs;
+      expect(defs).toHaveLength(1);
+      expect(defs[0]!.name).toBe('read_file');
+      expect(defs.map((d) => d.name)).not.toContain('bash');
+    });
+
+    it('returns empty array when allowedTools matches no schema', () => {
+      const dispatcher = new SessionToolDispatcher({
+        handlers: new Map(),
+        schemas: [...builtinToolSchemas],
+        permissions: { allowedTools: ['nonexistent_tool'] },
+      });
+      expect(dispatcher.toolDefs).toEqual([]);
+    });
   });
 
   describe('permissions', () => {
--- a/src/agent/tools/dispatcher.ts
+++ b/src/agent/tools/dispatcher.ts
@@ -456,8 +456,20 @@ export class SessionToolDispatcher implements ToolDispatcher {
     }
   }
 
+  // Contract: advertised schema MUST mirror the enforced allowlist.
+  // When an allowlist is configured, return only the schemas whose name is
+  // in that set — the model must not be shown tools the permission gate will
+  // reject (read-only / recon forks call bash, get "not in the configured
+  // allowlist", and waste turns).  An undefined allowlist means full access
+  // (all schemas returned unchanged), preserving the default unrestricted path.
+  // MCP tool visibility is preserved because the allowlist is already unioned
+  // with live MCP wire-names before reaching the dispatcher (see
+  // permissions.ts:withMcpToolsAllowed).
   get toolDefs(): readonly AnthropicToolDef[] {
-    return this.schemas;
+    const allowed = this.permissions?.allowedTools;
+    if (!allowed) return this.schemas;
+    const set = new Set(allowed);
+    return this.schemas.filter((s) => set.has(s.name));
   }
 
   /**


griffinwork40:
Citations verified at the reviewed ref (Wave 1.5 Check A complete):

- **`makeDispatcher()` default = `{ allowedTools: ['echo'] }`** — confirmed. The existing test's rename to `makeDispatcher({ permissions: undefined })` was genuinely load-bearing (otherwise it would return only the `echo` schema and fail `toEqual(builtinToolSchemas)`).
- **MCP-visibility invariant = TRUE** — independently re-derived. `withMcpToolsAllowed` (permissions.ts:60) returns `base` unchanged when `!base?.allowedTools` (the default-session no-op), otherwise unions MCP wire-names into `allowedTools`. Both providers wire it at construction (anthropic `:383`, openai `:383`), and MCP schemas are in `this.schemas`. So a restricted session's allowlist already contains its MCP tools before `toolDefs` filters.
- **Skill-dispatch strip** is an independent filter (anthropic `:700`, separate from the `toolDefs` read at `:685`/`:924`) — composes correctly after subsetting.
- **OpenAI `openAITools` caching** — confirmed cached once at constructor (query.ts:350 comment + constructor read). Pressure-tested: the dispatcher's `permissions`/`schemas` are fixed at construction, so `toolDefs`' value is stable for the dispatcher's lifetime — the cache is correct, and this PR does not introduce any staleness (it changes *what* `toolDefs` returns, not *when* it's read).

All top claims ground out. The highest-stakes claim (MCP invariant) was both agent-verified and independently re-derived by me from source — no separate shadow-verify wave needed since no genuine critical/high finding survives grading.

---

# Code Review — PR #295: `fix(tools): subset advertised tool schema to permission allowlist`

**Target:** PR #295 `fix/readonly-phase-tool-schema-subset` → `main` · reviewed ref `966ead7d` · +52/−3, 2 files · regime: light (fix)
**Stated intent:** PR #295 title + body — **spec-compliance assessed.**

## Verdict: ✅ MERGE

The fix is correct, minimal, and does exactly what it claims: when a session has `permissions.allowedTools` set, the model is now advertised only the allowlisted tool schemas instead of the full set, eliminating the wasted-turn loop where read-only/recon forks reach for `bash` and get rejected. All three load-bearing invariants hold under independent verification. No blocking defects. Every finding below is low/nit-grade polish.

### Invariants verified (the parts that would have made this DO-NOT-MERGE)
| Invariant | Status | Evidence |
|---|---|---|
| Undefined allowlist → full schema (default path unchanged) | ✅ TRUE | `dispatcher.ts:474` `if (!allowed) return this.schemas;` + `permissions.ts:64` no-op guard |
| MCP tools stay visible under subsetting | ✅ TRUE | `permissions.ts:60` unions wire-names into `allowedTools`; both providers wire `withMcpToolsAllowed` at `:383`; MCP schemas in `this.schemas` |
| Default/main sessions unaffected | ✅ TRUE | default sessions leave `allowedTools` undefined; only forks with an explicit allowlist change behavior — and those were already broken |
| Skill-dispatch strip composes | ✅ TRUE | independent filter at anthropic-direct `index.ts:700`, applied after the `toolDefs` read |

### Findings (none blocking)

**low · high · test-coverage · `dispatcher.test.ts:29–33` · ref:966ead7d · file-state**
The default-factory subset path is unexercised. `makeDispatcher()` defaults to `permissions: { allowedTools: ['echo'] }`, but no test asserts that the default factory's `toolDefs` returns only the `echo` schema — the new tests all construct dispatchers directly.
Evidence: `permissions: { allowedTools: ['echo'] }, ...overrides,`
Suggestion: add a case `expect(makeDispatcher().toolDefs.map(d => d.name)).toEqual(['echo'])`.
*(Wave-1 agent graded this `high`; corrected to `low` per rubric — "missing test = low". It is a coverage gap, not a reachable bug.)*

**low · high · test-coverage · `openai-compatible/query.ts:392` · ref:966ead7d · file-state**
No test asserts the OpenAI-compatible provider's `openAITools` reflects the subsetted `toolDefs` (grep: no `openai-compatible/` test exercises `toolDefs` + allowlist together). The anthropic path is covered; the duck-typed read is not.
Suggestion: construct an `OpenAICompatibleQuery` with a dispatcher carrying `allowedTools` and assert the pre-computed catalog is subset.

**nit · high · perf-observability · `dispatcher.ts:473–475` · ref:966ead7d · file-state**
`new Set(allowed)` is allocated on each `toolDefs` access. Negligible (allowlists are short; anthropic reads per-turn, openai once at construction). No action required.

**nit · medium · maintainability · `openai-compatible/query.ts:350` · ref:966ead7d · file-state**
Comment `recomputed only if dispatcher.toolDefs changes (it doesn't today)` reads as slightly stale now that `toolDefs` is permission-derived. It remains *accurate* (the value is stable for the dispatcher's lifetime), but a one-line clarification that `toolDefs` is now allowlist-filtered would prevent future misreading. **Pre-existing / orthogonal — not introduced by this PR.** Both Wave-1 agents raised this as `medium`; downgraded after verifying the dispatcher's `permissions`/`schemas` are immutable post-construction, so the cache is correct and the PR introduces no staleness.

### What was not checked
- **Ref read:** all citations verified against branch HEAD `966ead7d`; the MCP-visibility invariant and `makeDispatcher` default were independently re-derived from source.
- **Tests not run:** the working tree is dirty and may not be on the PR ref, so I did not execute the suite (read-only constraint). The PR reports `dispatcher.test.ts` 76 passing and one *pre-existing, unrelated* `anthropic-direct.test.ts > compact()` failure that reproduces on clean `main` — consistent with memory of that known env-flake.
- **Out of scope (pre-existing):** whether the OpenAI provider's construction-time `openAITools` cache refreshes on a mid-session MCP `tools/list_changed` is a latent question orthogonal to this diff (this PR changes *what* `toolDefs` returns, not *when* it's read) — not a regression introduced here.
- **Stated intent:** PR #295 title + body — spec-compliance assessed; the change satisfies its intent with no scope creep.

**Done**
- Reviewed PR #295 at ref `966ead7d` via two parallel dimension agents (correctness/spec/test/perf + security/api-compat) plus an inline Wave 1.5 citation/invariant re-verification; emitted the merge decision above.
- Evidence: findings cite `dispatcher.ts`, `permissions.ts`, both providers' `index.ts`/`query.ts`, and `dispatcher.test.ts` at the reviewed SHA; diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/cd6c7222-e6ab-43e5-bddd-d77621721bae/pr-295.diff`.
- World unchanged — review is read-only; no files, git state, or PR were mutated (working tree left dirty/untouched as instructed).
- Verdict: **MERGE** — no blocking findings; the load-bearing MCP-visibility and default-session invariants verified TRUE. Open polish: two missing test cases (low) and a comment clarification (nit), none gating.

---

PR #297, opened 2026-06-27
fix(router): re-point backend on a same-family /model switch to a different endpoint

griffinwork40:
## Summary

Fixes a mid-session `/model` switch that silently kept hitting the **previous** model's backend when switching between two models in the **same provider family but on different endpoints** — e.g. `gpt-5.5` (ChatGPT/Codex OAuth backend) → a configured **local** tier on `http://localhost:8081/v1`.

The switch reported `✓ Model switched to local`, then the next turn failed with:

```
ChatGPT/Codex backend rejected model "local" (HTTP 400). A ChatGPT subscription
only serves certain OpenAI models on this backend (gpt-5.5 works; gpt-5, gpt-5.1,
gpt-5.2 and *-codex do not).
```

i.e. the literal alias `local` was sent to the **ChatGPT backend** instead of `froggeric/…` going to `localhost:8081`.

## Root cause

Two compounding defects, both on the mid-session switch path (startup / `-m local` was unaffected — that path resolves correctly):

1. **`ProviderRouter` keyed inner reuse on provider FAMILY alone.** `gpt-5.5` and a local OpenAI-shim tier both resolve to `openai-compatible`, so the family matched and the router reused the live inner — forwarding the new model string without rebuilding. `applySlotCredentials` (which applies the per-slot `baseUrl`/`apiKey`) only runs inside `buildInner`, which only ran on a cross-family swap. So the inner stayed frozen on the ChatGPT OAuth backend (`auth.source`/baseURL are construction-time).

2. **`OpenAICompatibleQuery.setModel` stored the raw alias** as `currentModel` and sent it verbatim as the wire model — unlike `buildQueryFromConfig` (construction) and `anthropic-direct`'s `setModel`, which resolve the alias via `resolveModelId`.

## Fix

1. `ProviderRouter` now keys inner reuse on a full **routing signature** (`family + resolved baseUrl + openaiBaseUrl + apiKey`) rather than family alone. A signature change forces a turn-boundary rebuild, so a same-family switch onto a different endpoint re-points the backend. The check is gated on a model change, so a session that never switches pays no per-turn resolution cost. `setModel` likewise forwards to the live inner only on a true same-inner (same-signature) switch.

2. `OpenAICompatibleQuery.setModel` resolves slot/legacy aliases via `resolveModelId` before storing `currentModel`, matching the construction path and anthropic-direct.

The Telegram `/model` path goes through the same `AgentSession.setModel` → router, so it's fixed too.

## Tests

- `provider-router.test.ts`: rebuilds the inner on a same-family switch when the resolved endpoint differs; does **not** rebuild (forwards instead) when the endpoint is unchanged.
- `query.test.ts` (openai-compatible): a slot alias passed to `setModel` mid-session resolves to the bound wire id before the request body.

## Verification

- `pnpm lint` — clean.
- New + existing `src/agent/providers` + `src/agent/session` suites pass (936 tests), except one **pre-existing, environment-dependent** failure (`anthropic-direct.test.ts > compact()` resolving `claude-haiku-4-5` vs `claude-haiku-4-5-20251001`) confirmed failing identically on `main` without this change — unrelated to this fix.

No new `@anthropic-ai/sdk` symbols (no SDK lock change).

### DIFF
--- a/src/agent/providers/openai-compatible/query.test.ts
+++ b/src/agent/providers/openai-compatible/query.test.ts
@@ -675,6 +675,56 @@ describe('OpenAICompatibleQuery — model-slot resolution in request body', () =
     await collect(q);
     expect((createCalls[0]!.args as { model: string }).model).toBe('gpt-4o-mini');
   });
+
+  it('resolves a slot alias passed to setModel mid-session before the request body', async () => {
+    // Regression: openai-compatible `setModel` must resolve aliases like the
+    // construction path (buildQueryFromConfig) already does. Without it, a
+    // mid-session `/model <alias>` switch sent the literal alias as the wire
+    // model and the backend rejected it (the same-backend half of the
+    // ChatGPT/Codex 400 on a tier switch).
+    setSlotBindings({
+      local: { id: '' },
+      small: { id: 'gpt-4o-mini' },
+      medium: { id: 'gpt-5.5' },
+      large: { id: 'gpt-5.5' },
+    });
+    const controlled = makeControlledPromptStream();
+    const q = buildQueryFromConfig(baseConfig({ model: 'small' }), controlled.stream);
+    const iter = q[Symbol.asyncIterator]();
+
+    // Turn 1 on the `small` tier → gpt-4o-mini.
+    pendingChunks = [
+      { choices: [{ delta: { content: 'a' } }] },
+      { choices: [{ delta: {}, finish_reason: 'stop' }], usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 } },
+    ];
+    controlled.send('one');
+    let ev: IteratorResult<ProviderEvent>;
+    do {
+      ev = await iter.next();
+    } while (!ev.done && ev.value.type !== 'turn.completed');
+    expect((createCalls[0]!.args as { model: string }).model).toBe('gpt-4o-mini');
+
+    // Switch to the `medium` alias mid-session, then drive turn 2.
+    await q.setModel('medium');
+    pendingChunks = [
+      { choices: [{ delta: { content: 'b' } }] },
+      { choices: [{ delta: {}, finish_reason: 'stop' }], usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 } },
+    ];
+    controlled.send('two');
+    do {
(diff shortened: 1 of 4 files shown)


griffinwork40:
Review result for PR #297 (`fix/model-switch-same-family-backend` → `main`), reviewed at ref `8962e8c951f5cf99f21e0705f93334b05ce04737`.

## Merge decision: MERGE

No blocking findings found.

### Findings
None.

### Non-blocking notes
- `ProviderRouter.setModel()` now recomputes the routing signature before forwarding same-inner switches, which is the right guard for the stated bug.
- The per-turn path gates signature recomputation on `modelChanged`, so unchanged-turn routing does not pay the extra resolution cost.
- The OpenAI-compatible `setModel()` alias-resolution fix matches the construction path and is covered by a regression test.

### Evidence reviewed
- Diff artifact: `/Users/griffinlong/.afk/state/skill-preflight/0a8f70c1-5816-46cd-878e-964e06ff4ee2/pr-297.diff`
- Branch ref: `8962e8c951f5cf99f21e0705f93334b05ce04737`
- Verified branch file state with `git show` for:
  - `src/agent/providers/router/provider-router.ts`
  - `src/agent/providers/openai-compatible/query.ts`
  - `src/agent/providers/router/provider-router.test.ts`
- Stated intent was available from the PR title/preflight context and spec-compliance was assessed: the change re-points backend selection on same-family `/model` switches when endpoint/credentials differ.

### What was not checked
- Tests were not run; this was a read-only review.
- I did not post to GitHub or mutate the PR.
- I did not inspect runtime behavior outside the unit-level paths touched by the diff.

## Done
- What was done: Reviewed PR #297 read-only and found no blocking issues, with merge recommendation `MERGE`.
- Evidence that exists: Review was based on diff artifact `/Users/griffinlong/.afk/state/skill-preflight/0a8f70c1-5816-46cd-878e-964e06ff4ee2/pr-297.diff` and branch ref `8962e8c951f5cf99f21e0705f93334b05ce04737`.
- What changed in the world: Nothing was modified; the repository, PR, and working tree were left untouched.
- Anything still pending or deferred: Test execution was deferred because the `/review` task is read-only and the supplied preflight context did not request test runs.
- Uncommitted files: None were written or edited.

---

PR #298, opened 2026-06-27
feat(plugins): plugin JS `main` entrypoint run at session boot

griffinwork40:
## What

Adds a plugin **JS `main` entrypoint** loaded at session boot, so a plugin can contribute **code-backed** skills/agents (anything that calls `registerSkill()` at module load) without editing core — the same way `src/skills/all.ts` pulls in built-in skills today.

This is the enabling mechanism for moving code-backed capabilities out of the core tree and into a plugin/package.

## Why

Today the only way to register a code-backed skill is a static `import` in core (`src/skills/all.ts`). Plugins are discovered as `{ type, path }` and can only contribute markdown (`SKILL.md`) — there is no way for a plugin to run JS at boot. This PR adds that seam.

## How

- **`SdkPluginConfig.main?: string`** — additive, optional; declared by the plugin manifest's `main`.
- **`plugins-scanner`** parses `main` from `.claude-plugin/plugin.json` and attaches it to the scan result (omitted when absent/blank, so existing `{ type, path }` deep-equality stays green).
- **`agent/plugins/load-entrypoints.ts`** — `loadPluginEntrypoints()` resolves each plugin's `main` (relative to the plugin dir, or absolute), imports it as a `file://` URL, **dedupes per resolved path within a process**, and **isolates failures** (a broken entrypoint is non-fatal, surfaced via `onError`).
- **`skill-bridge`** — `scanAllPluginRoots()` is now the single source of truth for "which plugins exist" (project → user → bundled → imported); the two pre-existing duplicated root lists were refactored onto it so **entrypoint roots can never drift from skill roots**. `ensurePluginEntrypointsLoaded()` (idempotent, process-global, non-fatal) is the one call surfaces await before constructing a session.
- **`interactive/bootstrap`** awaits `ensurePluginEntrypointsLoaded()` before `buildAgentSession()`. This ordering is required: the skill manifest is built **synchronously** in the `AgentSession` constructor (`anthropic-direct` `query()`), so a plugin's `registerSkill()` must already have run for its skills to appear.

## Scope / follow-ups

- Wires the **REPL** (primary interactive surface) as the reference integration. `chat` / `daemon` / `telegram` are one-line follow-ups using the same `ensurePluginEntrypointsLoaded()` helper, landing with their own tests. **Subagents need no wiring** — they inherit the parent process's registry (the helper is a no-op there).
- `/reload-plugins` semantics: a newly-added plugin's entrypoint loads on the next `ensurePluginEntrypointsLoaded()` (new path); module unload on removal is out of scope.

## Tests / gates

- New: loader (import + side-effects fire at boot, skip-when-no-main, idempotency, non-fatal failure, relative+absolute resolution); scanner `main` parse (declared / blank-omitted); `scanAllPluginRoots` shape + surfaces a fixture plugin's `main`.
- `tsc --noEmit` clean. Touched suites green: `agent/tools` + `agent/plugins` + `cli/commands/interactive` (2254 pass, 0 fail). Branch rebased on current `main` (v4.45.3).

### DIFF
--- a/src/agent/plugins-scanner.test.ts
+++ b/src/agent/plugins-scanner.test.ts
@@ -51,6 +51,30 @@ describe('plugins-scanner', () => {
     expect(scanLocalPlugins(tmpHome)).toEqual([{ type: 'local', path: pluginDir }]);
   });
 
+  it('attaches the manifest `main` entrypoint when declared', () => {
+    const pluginDir = join(tmpHome, 'with-main');
+    mkdirSync(join(pluginDir, '.claude-plugin'), { recursive: true });
+    writeFileSync(
+      join(pluginDir, '.claude-plugin', 'plugin.json'),
+      JSON.stringify({ name: 'm', version: '0.0.0', main: 'dist/index.js' }),
+    );
+
+    expect(scanLocalPlugins(tmpHome)).toEqual([
+      { type: 'local', path: pluginDir, main: 'dist/index.js' },
+    ]);
+  });
+
+  it('omits `main` when the manifest declares it blank (plain absence covered above)', () => {
+    const blankMain = join(tmpHome, 'blank-main');
+    mkdirSync(join(blankMain, '.claude-plugin'), { recursive: true });
+    writeFileSync(
+      join(blankMain, '.claude-plugin', 'plugin.json'),
+      JSON.stringify({ name: 'b', version: '0.0.0', main: '   ' }),
+    );
+
+    expect(scanLocalPlugins(tmpHome)).toEqual([{ type: 'local', path: blankMain }]);
+  });
+
   it('discovers a marketplace-cache-layout plugin when it has an enabled index entry', () => {
     const pluginDir = join(tmpHome, 'cache', 'some-market', 'some-plugin');
     writePluginManifest(pluginDir);
--- a/src/agent/plugins-scanner.ts
+++ b/src/agent/plugins-scanner.ts
@@ -125,7 +125,7 @@ function walk(
     if (key === null) {
       // Path that doesn't fit either layout — keep loading it (matches
       // pre-marketplace behavior for unexpected directory shapes).
-      out.push({ type: 'local', path: dir });
+      pushPlugin(out, dir);
       return;
     }
     if (key.layout === 'cache') {
@@ -136,13 +136,13 @@ function walk(
         const entry = indexPlugins[key.key];
         if (!entry || entry.enabled === false) return;
       }
-      out.push({ type: 'local', path: dir });
+      pushPlugin(out, dir);
       return;
     }
     // Flat layout: include unless explicitly disabled.
     const entry = indexPlugins[key.key];
     if (entry && entry.enabled === false) return;
-    out.push({ type: 'local', path: dir });
+    pushPlugin(out, dir);
     return;
   }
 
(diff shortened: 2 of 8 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

The plugin JS `main`-entrypoint seam is correctly implemented and substantially spec-complete. No blocking, critical, or high findings. The notes below are non-blocking quality/observability follow-ups. All citations were read at the reviewed ref `a112add` and verified; absence claims were grep-confirmed.

**Reviewed ref:** `a112addef5ba6fec319a4577ee98acbde9dbe8b9`

### Spec-compliance — intent met ✓
Every named clause has an implementing change: `SdkPluginConfig.main?: string` (additive/optional, `sdk-types.ts`), scanner `main` parse with blank-omit (`plugins-scanner.ts`), `loadPluginEntrypoints` with per-resolved-path dedup + isolated/non-fatal failures (`load-entrypoints.ts`), `scanAllPluginRoots` as single source of truth folding the two former duplicated root lists (`skill-bridge.ts`), and `bootstrap` awaiting `ensurePluginEntrypointsLoaded()` before `buildAgentSession()`. Chat/daemon/telegram wiring and module-unload-on-removal are explicitly declared out of scope — their absence is not unmet intent.

Verified-correct behaviors: dedup is sound (`loadedEntrypoints.add(absPath)` runs synchronously before the `await import`, so a re-entrant same-turn call cannot double-import); failure isolation is per-plugin (each import in its own try/catch, loop continues); boot ordering holds (`ensurePluginEntrypointsLoaded()` awaited before the synchronous manifest build in the `AgentSession` constructor).

**1. medium · observability — `ensurePluginEntrypointsLoaded()` wires no `onError`, so plugin-boot failures are silent**
`src/agent/tools/skill-bridge.ts` (`ensurePluginEntrypointsLoaded`) · ref:a112add · file-state
```ts
export async function ensurePluginEntrypointsLoaded(): Promise<void> {
  await loadPluginEntrypoints(scanAllPluginRoots());  // no onError
}
```
The `onError` callback exists and is exercised by `load-entrypoints.test.ts` at the injectable `loadPluginEntrypoints` layer, but the production wrapper passes no handler — so a plugin whose `main` throws at boot is swallowed with zero diagnostic signal in the REPL path. (Flagged independently under both test-coverage and perf-observability.)
*Suggestion:* pass `onError: (plugin, err) => logger.warn('plugin entrypoint failed', { plugin: plugin.path, error: err })` and add a `skill-bridge.test.ts` case that stubs a plugin with a non-existent `main`, calls `ensurePluginEntrypointsLoaded`, and asserts it resolves without throwing + a warn fires.

**2. low · security — absolute/`..` `main` path is not containment-checked against `plugin.path`**
`src/agent/plugins/load-entrypoints.ts` (resolution block) · ref:a112add · file-state
```ts
const absPath = isAbsolute(plugin.main)
  ? plugin.main
  : resolvePath(plugin.path, plugin.main);
```
An absolute `main` (or a relative one with `..`) loads JS from outside the plugin's own directory. This is **not a new trust boundary** — plugin authors already fully control `plugin.path` and `plugin.json`, and existing markdown-skill loading reads from these same roots — so it is cross-plugin path aliasing, not arbitrary-input RCE. Worth a cheap guard.
*Suggestion:* after resolving, assert `absPath.startsWith(plugin.path + path.sep)` and reject non-fatally via `onError` for non-`trustAll` roots.

**3. low · test-coverage — `scanAllPluginRoots` priority ordering (project → user → bundled → imported) is unasserted**
`src/agent/tools/skill-bridge.test.ts` · ref:a112add
Existing tests cover shape and user-scope `main` propagation (`toContainEqual`), but no test seeds plugins across multiple scopes and asserts ordering. [low confidence — ordering may be treated as an implementation detail]
*Suggestion:* add a case seeding a plugin in ≥2 scopes and asserting relative order in the returned array.

**4. nit · perf — `loadImportFromConfig()` is read twice per manifest build**
`scanAllPluginRoots()` resolves imported roots via `loadImportFromConfig()` (unmemoized `readFileSync`), and `collectSkillEntries` resolves them again for skill roots; pre-PR a single snapshot was shared. No observable difference under normal operation — a mild atomicity nit.
*Suggestion:* let `scanAllPluginRoots` accept an optional pre-resolved `importedRoots` so the caller can pass one snapshot.

### What was not checked
- Wave 1 agents read against branch HEAD `a112addef5ba6fec319a4577ee98acbde9dbe8b9`; all blocking-tier citations were re-verified at that ref (CHECK A) and both absence claims grep-confirmed (CHECK B). No findings were dropped as fabricated/false-absent.
- Stated intent: PR #298 title + body — spec-compliance was assessed against it.
- Not run: the test suite itself (PR claims 2254 pass / 0 fail; CI on this PR is green — Lint & Build, Test, Docs, Publish Bundle Smoke all SUCCESS). Telegram/daemon/chat surfaces were not reviewed (out of scope for this PR).

PR #299, opened 2026-06-27
fix(daemon): don't default the daemon task to an unavailable skill

griffinwork40:
## What

The daemon's compiled-in default task was `/forge-friction --auto` — a skill not registered in this build. When `afk daemon` ran with **no** task configured (`--task`, `AFK_DAEMON_TASK`, or `daemon.task`), it fell back to that command and dispatched it via `session.sendMessage` on every scheduled tick → a confused/no-op model run (the skill isn't in the manifest) that burns the turn + tokens. Latent since the initial release; reachable e.g. via `afk daemon --cron "…"` with no `--task`.

## Fix

- **`COMPILED_DEFAULT_TASK` is now empty** — the daemon never fabricates a task.
- **cron / both triggers error clearly** when no task is configured ("A daemon task is required for the cron and both triggers. Provide one via --task, AFK_DAEMON_TASK, or daemon.task…"). These triggers already require `--cron`; scheduling a tick with nothing to run is a user mistake worth surfacing.
- **sessionstart with no task** registers no default task and proceeds — the daemon still runs persisted schedules + the built-in worktree-prune.
- `--task` help text no longer advertises an internal default.

Behavior is **unchanged** whenever a task is supplied via flag/env/config.

## Why this design

A daemon shouldn't invent an unrequested scheduled task, and it certainly shouldn't reference a skill the build can't run. Empty-default + fail-fast-on-cron is the correct, build-agnostic behavior and is forward-compatible with moving the self-improvement loop's task into config/a plugin.

## Tests

- Updated the "no flags" integration test → asserts the fabricated `default` task is **absent** (keyed on `taskId === COMPILED_DEFAULT_TASK_ID`, robust to any persisted schedules on the machine).
- Updated the cron-trigger test to supply a task; added a test that **`--cron` with no task errors**.
- Updated the `COMPILED_DEFAULT_TASK` value guard.
- `tsc --noEmit` clean; daemon suites green (1217 tests across `src/agent/daemon/**` + `src/cli/commands/**`).

## Out of scope (follow-ups)

- The `routing-directive.ts` `pendingBriefContext()` `/forge` nudge is a separate, **dormant** coupling (only fires when the briefs dir is non-empty, which only the internal loop produces) — left for the broader extraction/hygiene pass.
- Cosmetic `/forge-friction --auto` example strings remain in some daemon test fixtures / JSDoc (generic command examples; not dependent on the default).

### DIFF
--- a/src/cli/commands/daemon-session-factory.test.ts
+++ b/src/cli/commands/daemon-session-factory.test.ts
@@ -6,7 +6,7 @@
  * Root cause being tested: `afk daemon` previously passed no `sessionFactory`
  * to `startDaemon`, so sessions fell through to `resolveProvider()` →
  * `new AnthropicDirectProvider()` with no executor opts, omitting the three
- * orchestration tools. Skill commands like `/forge-friction --auto` would
+ * orchestration tools. Skill commands like `/some-skill --auto` would
  * fail because the provider's permission gate rejected tool calls it had
  * no handler for.
  *
--- a/src/cli/commands/daemon.test.ts
+++ b/src/cli/commands/daemon.test.ts
@@ -2,10 +2,11 @@
  * Tests for `afk daemon` command — Daemon Gap B (zero-config invocability).
  *
  * Covers:
- *   - `afk daemon` with no flags starts with sessionstart trigger + compiled
- *     default task (no error, no required options).
+ *   - `afk daemon` with no flags starts with sessionstart trigger and does NOT
+ *     fabricate a default task (no error, no required options).
  *   - `afk daemon --task <x>` still wires through correctly.
- *   - `afk daemon --cron <expr>` auto-selects cron trigger.
+ *   - `afk daemon --task <x> --cron <expr>` auto-selects cron trigger.
+ *   - `afk daemon --cron <expr>` without a task produces an error.
  *   - `afk daemon --trigger cron` without --cron still produces an error.
  *   - Config-driven default: `daemon.task` from afk.config.json is used when
  *     no --task flag is provided.
@@ -189,21 +190,19 @@ describe('afk daemon (CLI integration)', () => {
     consoleSpy.mockRestore();
   });
 
-  it('runs with no flags — uses sessionstart trigger and compiled default task', async () => {
+  it('runs with no flags — sessionstart trigger, no fabricated default task', async () => {
     await runDaemon();
 
     expect(mockStartDaemon).toHaveBeenCalledOnce();
     const opts = mockStartDaemon.mock.calls[0]?.[0];
     expect(opts).toBeDefined();
 
-    // The main task should use compiled defaults
+    // With no --task / env / config task, the daemon no longer invents a
+    // default task (previously '/forge-friction --auto'). The 'default' task id
+    // is unique to that fabricated task, so its absence is the precise signal —
+    // robust even if persisted schedules exist on the test machine.
     const tasks = opts?.tasks ?? [];
-    const mainTask = tasks.find((t) => t.taskId !== 'worktree-prune');
-    expect(mainTask).toBeDefined();
-    expect(mainTask?.command).toBe(COMPILED_DEFAULT_TASK);
-    expect(mainTask?.trigger).toBe('sessionstart');
-    // No cron expression required for sessionstart
-    expect(mainTask?.cronExpression).toBeUndefined();
+    expect(tasks.find((t) => t.taskId === COMPILED_DEFAULT_TASK_ID)).toBeUndefined();
   });
 
   it('passes --task through correctly', async () => {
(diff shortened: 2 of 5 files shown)


griffinwork40 (commented):

# Code Review — PR #299

**`fix(daemon): don't default the daemon task to an unavailable skill`**
Target: `griffinwork40/agent-afk` · `fix/daemon-default-task` → `main` · reviewed ref `12850c12` · 5 files, +49/−24 · regime: **light** (hotfix)

## ✅ Merge decision: **MERGE**

The fix is correct, strictly contained, behavior-preserving when a task is supplied, fails closed (errors loudly) on the cron/both no-task path, and fully implements its stated intent. No critical, high, or blocking findings. Three `low` follow-ups noted below; none blocks merge. Every load-bearing structural claim was independently re-read at the reviewed ref (`gh api .../contents?ref=12850c12`); CI on this PR is green (Lint & Build, Test, Docs, Publish Bundle Smoke all SUCCESS).

## What's solid (verified at `12850c12`)

- **Empty default flows through correctly.** `command = resolveDefaultTask(options.task, env.AFK_DAEMON_TASK, config.daemon?.task)` (`daemon.ts:268-272`) always returns a `string`, falling through `presentOrUndefined` to `COMPILED_DEFAULT_TASK = ''` (`daemon-options.ts:128-139`, `:100`) when nothing is configured — so `.trim()` cannot throw and both new guards actually fire.
- **No dropped built-ins.** The `tasks` array is `[]` for `pull` or empty-command (`daemon.ts:336-343`), but `worktreePruneTask` is `tasks.push()`-ed **unconditionally afterward** (`:344-346`) and persisted schedules after that (`:349-353`). Emptying the fabricated main task does **not** drop worktree-prune or persisted schedules — the intent's "sessionstart still runs persisted schedules + worktree-prune" holds. `startDaemon` is invoked regardless and tolerates an empty list (`agent/daemon.ts:114` — `for (const task of options.tasks ?? [])`).
- **Fail-closed on cron/both.** `(trigger === 'cron' || trigger === 'both') && command.trim() === ''` → `handleCommandError` (`daemon.ts:301-307`), placed right after the existing `--cron is required` check. Covers `both`, not just `cron`.
- **No breaking change.** `COMPILED_DEFAULT_TASK` (non-`_ID`) has **zero external production importers** — `daemon.ts` dropped it from its import (`:12`), and the only remaining production reference is internal to `daemon-options.ts` itself (`:137`). `resolveDefaultTask`/`startDaemon` signatures unchanged.
- **Tests track the behavior.** The value guard now asserts `''`; the no-flags test keys absence on `taskId === COMPILED_DEFAULT_TASK_ID`; the cron test supplies a task; a new test asserts `--cron` with no task rejects `/task is required for the cron and both triggers/`. The lingering `resolveDefaultTask(undefined, undefined, '') === COMPILED_DEFAULT_TASK` test (`daemon-options.test.ts:~163`) asserts against the symbol, so it passes at `''`.

## Findings (none blocking)

**1 · low · perf-observability · `src/cli/commands/daemon.ts:472` · ref:12850c12 · file-state** — On the sessionstart no-task path the startup log still prints `task='default' command='' trigger='sessionstart'`, implying a default task is active when none is registered.
- evidence: `console.log(palette.dim(\`  task='${taskId}' command='${command}' trigger='${trigger}'...\`));`
- suggestion: gate that line behind `tasks.length > 0`, or print `command='<none>'` when empty.

**2 · low · test-coverage · `src/cli/commands/daemon.test.ts:79` · ref:12850c12 · file-state** — The test file still imports the now-value-unused `COMPILED_DEFAULT_TASK` (non-`_ID`) symbol; all live assertions use `COMPILED_DEFAULT_TASK_ID`.
- evidence: `import { ... COMPILED_DEFAULT_TASK, COMPILED_DEFAULT_TASK_ID, } from '../daemon-options.js';`
- suggestion: drop `COMPILED_DEFAULT_TASK` from that import, keeping only `COMPILED_DEFAULT_TASK_ID`.

**3 · low · test-coverage · (missing test) · ref:12850c12** — The new error test covers `--cron` (which routes to `trigger === 'cron'`); the `--trigger both` no-task error path is handled by the guard but unasserted.
- suggestion: add `await expect(runDaemon('--trigger', 'both', '--cron', '0 * * * *')).rejects.toThrow(/task is required/)`.

## Per-dimension

- **security** · no issues found — read `daemon.ts` at ref:12850c12. The changed constant was never sensitive; no auth/secret/injection surface touched.
- **api-compat** · no issues found — `resolveDefaultTask` + `startDaemon` signatures unchanged; `COMPILED_DEFAULT_TASK` has zero external production importers (verified at ref).
- **correctness** · no issues found — empty-command and cron/both guards verified line-by-line; worktree-prune + persisted schedules verified pushed after the conditional.
- **spec-compliance** · no issues found (assessed vs PR #299 title+body) — all four stated requirements implemented and tested; the `routing-directive.ts` `/forge` nudge and cosmetic example strings are correctly the PR's disclosed out-of-scope items, not unmet intent. No scope creep.
- **perf-observability** · the only gap is the cosmetic log line (finding 1); no hot-path change.

## What was not checked

- All citations read against branch HEAD `12850c12` via `gh api .../contents?ref=…` (no local clone of the public repo assumed); every load-bearing structural claim re-derived from source. No fabricated or false-absence findings.
- Test suite not executed in this environment (read-only review); the PR reports daemon suites green (1217 tests) and CI is green on this PR.
- No critical/high findings → shadow-verify not triggered per the review skill.
- Stated intent: PR #299 title + body — spec-compliance assessed.

This review is read-only and advisory — no files, commits, or PR state were modified by the reviewer.

---
_🤖 Posted by the agent-afk hourly PR-review sweep · griffinwork40_

PR #301, opened 2026-06-28
fix(cli): width-cap the markdown message-fallback render path

griffinwork40:
## Summary
- **Bug:** assistant text sometimes rendered off-screen on the right; resizing didn't fix it and made it worse.
- **Root cause:** `emitMarkdown` (`src/cli/_lib/stream-renderer-orchestrator-emit.ts`) — the fallback render path used when a `message` arrives with no prior streamed deltas, and on non-TTY — called `renderMarkdownToTerminal(text)` with **no `maxWidth`**, so the table formatter fell through to `+Infinity` (`formatter.ts:356,360`) and skipped squeezing/truncation entirely. Rows emitted at natural width overflowed the right edge, and since committed lines are immutable scrollback they never reflow on resize (the resize bus only repaints the live overlay). The streaming commit path was always width-capped, which is why it was only "sometimes."
- **Fix:** TTY-gated width cap in `emitMarkdown` (`terminal - 2`, matching the streaming path's `calculateContentWidth`). Off a TTY (piped/redirected) width stays unbounded so downstream consumers get full-width markdown — prior behavior preserved.
- Adds opt-in `breakLongWords` to `wrapToWidth` so an over-long token (bare URL, long identifier) breaks at the column boundary instead of overflowing, without mid-word-splitting normal prose (toggles wrap-ansi's `hard` flag only; `wordWrap` stays true).

## Test results
- `pnpm lint` (tsc --noEmit, strict): **passed**
- `pnpm test` (full suite): **10164 passed / 14 skipped**; the only 2 failures are pre-existing, environment-only, and unrelated to this change (`config.test.ts` model-slot bindings, `anthropic-direct.test.ts` compact model-id) — neither in a file this PR touches.
- Targeted: `wrap` / `formatter` / `markdown-stream-format` / `markdown-stream` / `stream-renderer-emit-width` = **152 passed**. New coverage: `breakLongWords` cases in `wrap.test.ts` plus new `stream-renderer-emit-width.test.ts` (TTY caps every line, breaks bare long tokens, non-TTY stays full-width).

## Verification
Adversarial re-derivation from the diff alone (read-only sub-agent, no access to the commit's reasoning):
- **CLAIM 1** (pre-change: absent `maxWidth` -> `maxTableWidth=undefined` -> `targetWidth=+Infinity` -> squeeze block skipped via the `Number.isFinite` guard) — **CONFIRM** (`formatter.ts:129,356,360`).
- **CLAIM 2** (after: cap only when `process.stdout.isTTY`, else `+Infinity` leaves output unchanged) — **CONFIRM** (`orchestrator-emit.ts` + `wrap.ts:21`).
- **CLAIM 3** (`breakLongWords` toggles only `hard`; `wordWrap` stays true -> normal prose wraps identically) — **CONFIRM** (`wrap.ts:33`).
- No regressions; streaming path (`commitBlockAbove`) untouched; `Math.max(1, ...)` guards zero/negative width. Verdict: **SOUND**.

## Out of scope
- Already-committed scrollback never reflows on resize — inherent to the line-committing renderer (would need a full-screen redraw model). This PR prevents the over-wide emit at commit time on the fallback path rather than retroactively re-flowing history.

### DIFF
--- a/src/cli/_lib/stream-renderer-emit-width.test.ts
+++ b/src/cli/_lib/stream-renderer-emit-width.test.ts
@@ -1,0 +1,85 @@
+/**
+ * Regression tests for emitMarkdown width-capping.
+ *
+ * emitMarkdown is the fallback render path used when a `message` event arrives
+ * with no active streaming renderer (cached/non-incremental responses) and on
+ * non-TTY surfaces. It previously called renderMarkdownToTerminal(text) with no
+ * maxWidth, so wide tables rendered at their natural width and overflowed past
+ * the right edge on any terminal narrower than the table (and never reflowed on
+ * resize, because the lines are already committed to scrollback). These tests
+ * lock the fix: on a TTY the output is capped to the visible content width;
+ * piped/non-TTY output stays full-width.
+ */
+
+import { describe, it, expect, afterEach } from 'vitest';
+import { emitMarkdown } from './stream-renderer-orchestrator-emit.js';
+import { displayWidth } from '../display.js';
+import type { Writer } from '../slash/types.js';
+
+function captureWriter(): { lines: string[]; out: Writer } {
+  const lines: string[] = [];
+  const out: Writer = {
+    line: (text = '') => lines.push(text),
+    raw: (text) => lines.push(text),
+    success: (text) => lines.push(text),
+    info: (text) => lines.push(text),
+    warn: (text) => lines.push(text),
+    error: (text) => lines.push(text),
+  };
+  return { lines, out };
+}
+
+const WIDE_TABLE = [
+  '| Column One Header | Column Two Header | Column Three Header |',
+  '| --- | --- | --- |',
+  '| alpha value here | beta value here | gamma value here too |',
+  '| delta value here | epsilon value here | zeta value here also |',
+].join('\n');
+
+describe('emitMarkdown width capping', () => {
+  const origColumns = Object.getOwnPropertyDescriptor(process.stdout, 'columns');
+  const origIsTTY = Object.getOwnPropertyDescriptor(process.stdout, 'isTTY');
+
+  afterEach(() => {
+    if (origColumns) Object.defineProperty(process.stdout, 'columns', origColumns);
+    if (origIsTTY) Object.defineProperty(process.stdout, 'isTTY', origIsTTY);
+  });
+
+  function stubTerminal(isTTY: boolean, columns: number): void {
+    Object.defineProperty(process.stdout, 'isTTY', { value: isTTY, configurable: true });
+    Object.defineProperty(process.stdout, 'columns', { value: columns, configurable: true });
+  }
+
+  it('caps every emitted line to the content width on a TTY', () => {
+    stubTerminal(true, 40);
+    const { lines, out } = captureWriter();
+    emitMarkdown(WIDE_TABLE, out);
+
+    expect(lines.length).toBeGreaterThan(0);
+    // contentWidth budget is terminal − 2 = 38. No emitted line may exceed it.
+    for (const line of lines) {
+      expect(displayWidth(line)).toBeLessThanOrEqual(38);
+    }
+  });
+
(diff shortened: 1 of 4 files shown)


griffinwork40:
## PR #301 review — MERGE

**Decision:** **MERGE**

**Reviewed ref:** `0900f6b41d63f0fb7597b1e34c7424e5d402be54`  
**Diff artifact:** `/Users/griffinlong/.afk/state/skill-preflight/49fb5cda-4145-431e-9981-c8b442c8dd7b/pr-301.diff`  
**Stated intent:** PR title, `fix(cli): width-cap the markdown message-fallback render path` — spec-compliance assessed.

### Summary

No blocking, high, or medium issues found. The change satisfies the stated intent: `emitMarkdown` now width-caps markdown fallback rendering on TTYs, preserves full-width non-TTY output, and adds focused regression coverage for tables and long tokens.

### Findings

1. **low · high confidence · correctness/comment accuracy · `src/cli/_lib/stream-renderer-orchestrator-emit.ts:295-306` · ref:`0900f6b41d63f0fb7597b1e34c7424e5d402be54` · citation-type:file-state**  
   **Finding:** The comment says the fallback aligns with the streaming commit path, but the fallback now hard-breaks long tokens while the streaming path still soft-wraps them.  
   **Evidence:** `emitMarkdown` calls `wrapToWidth(rendered, contentWidth, { breakLongWords: true })`; `formatBlockForCommit` calls `wrapToWidth(rendered, contentWidth)`.  
   **Suggestion:** Narrow the comment to say table/prose width-capping aligns with streaming, while hard-breaking overlong raw-scrollback tokens is fallback-specific.

2. **low · medium confidence · tests · `src/cli/_lib/stream-renderer-emit-width.test.ts:40-50` · ref:`0900f6b41d63f0fb7597b1e34c7424e5d402be54` · citation-type:file-state**  
   **Finding:** The test teardown restores `process.stdout.columns` and `isTTY` only when original own-property descriptors existed, so CI environments without those own properties leave test-installed descriptors in place for the rest of the worker.  
   **Evidence:** `afterEach` only calls `Object.defineProperty` under `if (origColumns)` and `if (origIsTTY)`, while `stubTerminal` always installs configurable own properties.  
   **Suggestion:** Delete the properties when the original descriptor was absent, or use a Vitest stub helper that restores properties automatically.

### Clean checks

- **Security:** no issues found; the diff only changes local markdown rendering and ANSI-aware wrapping.
- **API compatibility:** no issues found; `wrapToWidth(text, width, opts?)` is backward-compatible with existing two-argument callers.
- **Spec compliance:** no issues found; TTY fallback output is capped to `getTerminalWidth() - 2`, and non-TTY output remains unbounded.
- **Performance:** no issues found; the added wrap pass is on the fallback/non-streaming emit path.
- **Test coverage:** adequate for the stated bug; added tests cover TTY table capping, long-token hard breaking, and non-TTY preservation.

### What was not checked

- Read against branch HEAD `0900f6b41d63f0fb7597b1e34c7424e5d402be54`; cited file-state lines were verified at that ref.
- I did not run the test suite or lint; this review was read-only per preflight instructions.
- I did not review unrelated call sites such as `OutputFormatter.formatMarkdown`, which remain outside the PR’s stated fallback-render-path scope.
- The pre-existing dirty working tree was not mutated.

## Done

- What was done: PR #301 was reviewed read-only, with merge recommendation **MERGE**.
- Evidence: review used diff artifact `/Users/griffinlong/.afk/state/skill-preflight/49fb5cda-4145-431e-9981-c8b442c8dd7b/pr-301.diff` and verified cited lines against ref `0900f6b41d63f0fb7597b1e34c7424e5d402be54`.
- What changed in the world: no repository, PR, git, or filesystem mutations were made.
- Pending or deferred: tests/lint were not run because the review preflight required read-only inspection only.
- Uncommitted paths: I did not write or edit files; the pre-existing dirty working tree, including `docs/issue-silent-model-loop-debuggability.md`, was preserved.

---

PR #294, opened 2026-06-27
feat(afk): high-risk AFK gate routes approve/deny to the phone (Phase 2 v1.5)

griffinwork40:
## Summary

Phase 2 of two-way AFK comms, building on #292 (the AFK path-permission fix, already merged). Two commits:

1. **`test(afk)` — end-to-end two-way round-trip seam test** (verification spike, no production change).
2. **`feat(afk)` — high-risk AFK gate routes approve/deny to the phone (v1.5)**.

Mental model: **AFK = `bypassPermissions` − footguns + phone.** #292 removed the footgun (keyboard-blocking path prompts in AFK). This PR adds the phone: a high-risk op no longer hard-stops an unattended run — it asks the operator to approve or deny from Telegram, over the elicitation round-trip that already exists.

## Background

Two-way AFK comms was already ~90% built (outbound auto-subscribe + batched render; inbound elicitation-answer and abort round-trips over an HMAC-signed on-disk ledger hop). This PR (a) proves that loop end-to-end with one seam test, and (b) closes the highest-value gap: the high-risk gate previously hard-blocked and relied on the model to `send_telegram`, with no way for the operator to say "yes, do it."

### 1. Round-trip seam test (`src/telegram/afk-two-way-roundtrip.test.ts`)
Component tests already covered each end in isolation (`afk-ledger-channel.test.ts` = REPL side; `watch.test.ts` = daemon side), both against the `afk-channel` canonical signing form. This adds the one missing assertion: the **real** daemon `SessionWatchManager._run` wired to the **real** REPL `makeLedgerChannelHandler` / `makeAbortWatcher` over a single shared ledger + key, with nothing hand-signed in between — proving the full cross-process round-trip settles for both the elicitation and abort halves.

### 2. v1.5 — high-risk gate → phone approve/deny (`afk-mode-gate.ts`, `default-hook-registry.ts`)
In `autonomous` (AFK) mode, on a `high`-risk / irreversible tool call:

| Case | Behaviour |
|------|-----------|
| **Main session** | Emit an approve/deny form elicitation via `elicitationRouter.route()`. In AFK that router is swapped to the ledger→phone channel (`/afk on`), so it renders on the phone (racing the keyboard) and the signed answer round-trips back. **Approve → the op runs** (this single call). |
| **Deny / decline / cancel / timeout** | Blocked (refused to the model). |
[...]

PR #304, opened 2026-06-28
feat(sdk): wire canUseTool into the dispatcher — Agent SDK parity Dim 8 (increment 1)

griffinwork40:
## What

Closes the **runtime half** of Agent SDK parity **Dim 8** (permission rules-engine). `canUseTool` was already a fully-typed surface (`src/agent/types/sdk-types.ts`) and was bubbled to sub-agents (`subagent.ts` `parentCanUseTool`), **but it was never invoked inside the dispatcher** — inert at runtime. This wires it in, enforced.

## Changes

- **`SessionToolDispatcher`** consults `canUseTool` on every tool call in **both** `execute()` and the `executeBatch()` phase-1 loop (so parallel/batched rounds are gated too) — **after** the static allowlist (`checkToolPermission`) and **before** the read-only-bash gate.
  - **Precedence:** a hard allowlist deny still wins; `canUseTool` can only *further restrict*, never widen (covered by a test).
  - **`deny`** → `permission-denied` `ToolResult` (uses `result.message`). A **throwing** policy **fails closed** (deny) rather than crashing the turn.
  - **`allow`** → proceeds; an **`allow.updatedInput`** rewrites `call.input` in place so both the handler and PostToolUse observe the new value.
  - `ask` is resolved inside the callback (`createCanUseToolHook`), so the dispatcher only ever observes a final allow/deny.
- **Threading:** `AgentConfig.canUseTool` → `resolveProvider` (mirrors the existing `customTools` pattern) → both providers' options → `buildDispatcher` → dispatcher. **No behavior change when unset** — default sessions and existing consumers are unaffected. Sub-agent bubbling now reaches the dispatcher end-to-end via the same path.

## Not in this increment (deferred)

Persisted allow/deny/ask **rule lists with settings-tier destinations**, the `PermissionRequest` hook, and `deny.interrupt` (turn-abort) semantics — the persistence layer of Dim 8. This increment is the enforcement primitive those build on.

## Verification

- `pnpm lint` clean (rebased on v5.3.0); `pnpm audit:sdk:check` **green — zero new `@anthropic-ai/sdk` symbols** (`CanUseTool`/`PermissionResult` are local types).
- `dispatcher.test.ts`: **+8 `canUseTool` cases** (allow, deny short-circuit, deny-overrides-allowlist, allowlist-still-wins, `updatedInput`, fail-closed-on-throw, `executeBatch` gating, no-op-when-unset) — all pass.
- **2322 pass** across `tools` + both providers + `permissions` + `query` + `subagent` + `session`; the lone failure is the pre-existing `anthropic-direct.test.ts` compact-model-id env test (documented on `main`, unrelated to this change).

> Note: merging cuts a `feat:` minor release → **v5.4.0** (v5.3.0 already shipped via #294).

### DIFF
--- a/src/agent/providers/anthropic-direct/index.ts
+++ b/src/agent/providers/anthropic-direct/index.ts
@@ -24,6 +24,7 @@ import { appendFileSync, mkdirSync } from 'fs';
 import { dirname } from 'path';
 import Anthropic from '@anthropic-ai/sdk';
 import type { AgentConfig } from '../../types/config-types.js';
+import type { CanUseTool } from '../../types/sdk-types.js';
 import type {
   ModelProvider,
   ProviderQuery,
@@ -117,6 +118,8 @@ export interface AnthropicDirectProviderOptions {
   hookRegistry?: HookRegistry;
   /** Tool permission configuration (allowlist). */
   permissions?: ToolPermissionConfig;
+  /** In-process permission callback, forwarded to the session dispatcher. */
+  canUseTool?: CanUseTool;
   /**
    * Optional client factory override. When set, takes precedence over the
    * module-scope `__setAnthropicClientFactory` hook. Useful for callers that
@@ -193,6 +196,7 @@ export class AnthropicDirectProvider implements ModelProvider {
   private readonly schemas: readonly import('../../tools/types.js').AnthropicToolDef[];
   private readonly hookRegistry: import('../../hooks.js').HookRegistry | undefined;
   private readonly permissions: ToolPermissionConfig | undefined;
+  private readonly canUseTool: CanUseTool | undefined;
   private readonly subagentExecutor: import('../../tools/subagent-executor.js').SubagentExecutor | undefined;
   private readonly composeExecutor: import('../../tools/compose-executor.js').ComposeExecutor | undefined;
   private readonly surface: string;
@@ -291,6 +295,7 @@ export class AnthropicDirectProvider implements ModelProvider {
     this.schemas = schemas;
     this.hookRegistry = opts.hookRegistry;
     this.permissions = opts.permissions;
+    this.canUseTool = opts.canUseTool;
     this.subagentExecutor = opts.subagentExecutor;
     this.composeExecutor = opts.composeExecutor;
     this.surface = opts.surface ?? 'cli';
@@ -452,6 +457,9 @@ export class AnthropicDirectProvider implements ModelProvider {
       subagentExecutor: this.subagentExecutor,
       skillExecutor: this.skillExecutor,
       composeExecutor: this.composeExecutor,
+      // In-process permission callback (Dim 8). No-op when unset; forwarded by
+      // reference so it composes with the static allowlist in the dispatcher.
+      ...(this.canUseTool !== undefined ? { canUseTool: this.canUseTool } : {}),
       cwd: opts?.cwd,
       readRoots: opts?.readRoots,
       writeRoots: opts?.writeRoots,
--- a/src/agent/providers/index.ts
+++ b/src/agent/providers/index.ts
@@ -268,24 +268,30 @@ export function providerForModel(
 export function resolveProvider(
   model: string | undefined,
   hints?: ProviderRouteHints,
-  opts?: { customTools?: import('../tools/custom-tool.js').CustomToolDef[] },
+  opts?: {
+    customTools?: import('../tools/custom-tool.js').CustomToolDef[];
+    canUseTool?: import('../types/sdk-types.js').CanUseTool;
+  },
 ): ModelProvider {
   const name = providerForModel(model, hints);
   const customTools = opts?.customTools;
+  const canUseTool = opts?.canUseTool;
+  // Mirror the customTools threading: forward each only when present so the
+  // default path constructs the provider with an empty options bag.
+  const ctorOpts = {
+    ...(customTools !== undefined && customTools.length > 0 ? { customTools } : {}),
+    ...(canUseTool !== undefined ? { canUseTool } : {}),
+  };
   switch (name) {
     case 'openai-compatible':
     case 'openai-codex':
       // IMPORTANT: fresh instance per session — see docstring above.
-      return new OpenAICompatibleProvider(
-        customTools !== undefined && customTools.length > 0 ? { customTools } : {},
-      );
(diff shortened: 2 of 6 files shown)


griffinwork40 (commented):

## Merge decision: MERGE

**Target:** feature · `src/agent/tools/dispatcher.ts` (+ 2 providers, `resolveProvider`, `agent-session.ts`, `dispatcher.test.ts`) · ~230 / −8 lines · reviewed ref `22d09f0`
**Intent:** wire the already-typed `canUseTool` permission callback into `SessionToolDispatcher` (runtime half of Agent SDK parity Dim 8) — consulted after the static allowlist, before the bash gate, in both `execute()` and `executeBatch()`; fail-closed on throw; `allow.updatedInput` rewrites input in place.

No blocking findings. The change is well-scoped, fully threaded, and adequately tested. The notes below are non-blocking quality improvements.

### Notes (non-blocking)

**[medium · perf-observability · `src/agent/tools/dispatcher.ts:564–596` · ref `22d09f0` · file-state]** — A `canUseTool` deny (and the fail-closed throw branch) emits no trace/telemetry event, so policy denials are invisible in the witness layer. PreToolUse denials land a `hook_decision` event via `traceWriter` (used at `dispatcher.ts:616`, `792`, `1054`, `1080`); `runCanUseTool` contains zero `traceWriter`/log references (grep of lines 564–596 = empty).
Suggestion: emit a `traceWriter` event (matching the `hook_decision` shape) in the deny and throw branches so `canUseTool` denials appear in session telemetry.

**[low · security/correctness · `src/agent/tools/dispatcher.ts:586–595` · ref `22d09f0` · file-state]** — `runCanUseTool` returns on `behavior === 'deny'`, then falls through to the `updatedInput`/`return null` (allow) path for anything else. There is no exhaustiveness guard, so a non-`deny`/non-`allow` runtime value (e.g. `{behavior:'ask'}`) would be treated as allow rather than failing closed.
Evidence:
```
if (result.behavior === 'deny') { return { ...permission-denied... }; }
// allow — apply an optional input rewrite in place.
if (result.updatedInput !== undefined) { call.input = result.updatedInput; }
return null;   // ← any non-'deny' behavior reaches here
```
This is **type-protected for the supported path**: `PermissionResult` (`sdk-types.ts:112–126`) is a discriminated union of only `{behavior:'allow'}` | `{behavior:'deny'}`, and `CanUseTool` returns `Promise<PermissionResult>`, so a well-typed TS caller cannot return `'ask'`. The separate `PermissionBehavior` type (`sdk-types.ts:60`) does include `'ask'`, so the gap is reachable only by a JS caller or an explicit cast — defense-in-depth, not a live auth bypass.
Suggestion: add an explicit `if (result.behavior !== 'allow') return { ...permission-denied... };` before the `updatedInput` branch so the runtime fails closed on any unexpected behavior.

**[low · test-coverage · `src/agent/tools/dispatcher.test.ts` · ref `22d09f0` · file-state]** — `updatedInput` is exercised only through `execute()` (single-call path); no test drives an input rewrite through the multi-call `executeBatch()` phase-1 → phase-2 handoff. (`git grep -n updatedInput` at the ref returns only the two `execute()`-path lines; the `executeBatch` gating test has no `updatedInput` case.) Single-element batches delegate to `execute()`, so only the genuine multi-call path is uncovered.
Suggestion: add an `executeBatch` test with a 2-call batch where one call's input is rewritten and the handler observes the new value in phase 2.

**[low · perf-observability · `src/agent/tools/dispatcher.ts:822` · ref `22d09f0` · file-state]** — In `executeBatch`, `runCanUseTool` is awaited sequentially inside the phase-1 loop, so a slow policy (e.g. a network round-trip) adds its latency once per call with no timeout guard. This is **not a regression** — phase-1 was already sequential (the "Phase 1: sequential PreToolUse + permission" loop predates this PR) — but the `call.signal` passed to the callback does not gate the `await` itself.
Suggestion: forward `call.signal` (or an `AbortSignal.timeout`) as a cancellation guard around the `await` so a hung policy cannot stall a batch indefinitely.

### Clean dimensions
- **Correctness** — `executeBatch` precedence verified identical to `execute()` (PreToolUse → `checkToolPermission` → `runCanUseTool` → bash gate; `dispatcher.ts:803–828`). Fail-closed catch returns a `permission-denied` ToolResult without rethrow. `updatedInput !== undefined` correctly gates an absent field. Empty-`message` deny falls back to a non-empty string.
- **api-compat** — All additions are optional fields (`canUseTool?`) on `SessionToolDispatcherOptions` and both provider option types; `resolveProvider`'s 3rd-arg `opts` was already optional and the new field is additive. Verified no production caller breaks: `agent-session.ts` (updated here), `suggest.ts` (2-arg call still valid), `nesting.ts` (constructs providers directly), `daemon.ts` (no call). `CanUseTool`/`PermissionResult` are local types in `sdk-types.ts` — zero new `@anthropic-ai/sdk` symbols, matching the PR's `audit:sdk:check` claim.
- **spec-compliance** — COMPLIANT. Every named requirement is met: wired into `execute()` (`:643`) and `executeBatch()` phase-1 (`:822`); allowlist precedence preserved (canUseTool only restricts); fail-closed on throw; `updatedInput` rewrites in place; full threading `AgentConfig → resolveProvider → both providers → dispatcher`; no-op when unset (early `return null`). No scope creep; the deferred items (persisted rule lists, `PermissionRequest` hook, `deny.interrupt`) are absent from the diff as stated.
- **security** — No `deny` bypass (allowlist deny returns before `runCanUseTool`); the bash gate reads the already-rewritten `call.input` after `updatedInput`; no TOCTOU on `call.name`; `signal`/`toolUseID` context carries no secrets.

### What was not checked
- Read against branch HEAD `22d09f0` from a local clone of `griffinwork40/agent-afk`; all cited lines verified at that ref via `git show <sha>:<file>` (file-state) plus the inline diff hunks. No citation was fabricated or diff-only; no absence claim was refuted.
- `createCanUseToolHook` (the `ask`-resolution layer above the dispatcher) was not read — the type-level analysis assumes it never returns a raw `'ask'` to the dispatcher.
- Tests were not run locally; relied on the PR's stated green CI (Lint & Build, Test, Docs all SUCCESS on this ref) as the test signal.
- Stated intent: PR #304 title + body — spec-compliance assessed and COMPLIANT.

PR #305, opened 2026-06-28
feat(sdk): export framework skill/facet API for out-of-tree plugins

griffinwork40:
## What

Re-exports the core symbols an out-of-tree skill plugin (installed under `~/.afk/plugins/`) needs to register skills and read session facets at boot, from the published `agent-afk` package entry (`src/index.ts`).

Added to the public API:
- **Skills registry:** `registerSkill`, `listSkills`, `getSkill`, types `SkillExecutionContext`/`SkillMetadata`, and `loadSkillPrompts`.
- **Session facets:** `deriveSessionFacet`, `getOrDeriveFacet`, `listSessionIds`, `loadStoredSession`, types `SessionFacet`/`StoredSessionInput`/`ToolEventInput`.
- **Subagent + discovery:** `describeFailure`, `discoverPluginSkillBodies`.
- **Env + state dirs:** `env`, `getSessionsDir`, `getSkillsDir`, `getAgentFrameworkDir`.

(`SubagentManager` and `IAgentSession`, also needed by plugins, were already exported.)

## Why

Builds on #298 (plugin JS `main` entrypoint). For a skill plugin to run `registerSkill()` at boot via its `main` module, it must import these symbols from the package — today they're only reachable via deep relative source paths. This re-export surface mirrors the in-tree imports exactly, so a plugin depends on the public package instead of reaching into the tree.

## Risk

Purely **additive** — no runtime behavior change; it only widens the public API surface. The re-exported list was derived by enumerating the exact core imports a plugin makes, then verified symbol-by-symbol against the current modules.

## Verification

- `pnpm lint` (tsc --noEmit, strict) — clean.
- `pnpm build` — clean (tsc emit + prompt copy).
- `src/index.test.ts` extended with a framework-SDK-surface assertion (all new value exports resolve at runtime) — passes.
- Full suite: only the 2 known env-dependent pre-existing failures (config slot-bindings, anthropic compact) — unrelated to this change.

PR #307, opened 2026-06-28
feat(skills): loadSkillPrompts(name, baseDir?) for out-of-tree plugins

griffinwork40:
## What

Adds an optional second argument `baseDir` to `loadSkillPrompts(name)`. It is the **skills-root directory** that contains the `<name>/` dir. When supplied it overrides the in-tree root (resolved relative to the module), so an **out-of-tree plugin** can load prompts bundled in its own package:

```ts
loadSkillPrompts('my-skill', join(dirname(fileURLToPath(import.meta.url)), 'skills'))
```

## Why

Third step of the plugin-extensibility roadmap. Combined with:
- **#298** (merged) — a plugin `main` JS entrypoint that runs `registerSkill()` at session boot, and
- **#305** (open) — re-exporting the framework skill/facet API from `src/index.ts`,

…this lets a plugin register a skill **and resolve that skill's prompt files from its own directory**, rather than being forced to live in this repo's `src/skills/` tree.

## Backward compatibility

Fully additive. Every existing caller passes `name` only (verified: ~30 call sites across `src/skills/**`, `src/agent/tools/**`) and is unaffected — `baseDir` defaults to the current in-tree resolution.

## Safety

The `name` path-traversal guard (rejects `/`, `\`, leading `.`, `..`) still fires **regardless of `baseDir`**, so a malicious skill id cannot escape the supplied root. The available-skills listing and the no-prompts error now derive from the resolved root.

## Tests

`src/skills/_lib/prompt-loader.test.ts` extended with a `baseDir` suite (temp-fixture skills root): loads from baseDir, `.md`-only + alphabetical keys, unknown-skill listing from baseDir, no-prompts error, and guard-still-fires-with-baseDir. Existing guard + behavior tests unchanged.

- `tsc --noEmit` clean
- 21 prompt-loader tests green; consuming suites (`skill-executor`, `skill-load-mode`) green

### DIFF
--- a/src/skills/_lib/prompt-loader.test.ts
+++ b/src/skills/_lib/prompt-loader.test.ts
@@ -15,7 +15,10 @@
  * @module skills/_lib/prompt-loader.test
  */
 
-import { describe, it, expect } from 'vitest';
+import { afterAll, beforeAll, describe, it, expect } from 'vitest';
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
 import { loadSkillPrompts } from './prompt-loader.js';
 
 describe('P3 — loadSkillPrompts path-traversal guard', () => {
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

## Merge decision: MERGE

**Target:** feature · `src/skills/_lib/prompt-loader.ts`, `src/skills/_lib/prompt-loader.test.ts` · +74 / −12 · reviewed ref `e117c01`
**Intent:** add an optional `baseDir` arg to `loadSkillPrompts(name)` — the skills-root dir containing `<name>/` — so an out-of-tree plugin can load prompts bundled in its own package; keep it fully additive; keep the `name` path-traversal guard firing regardless of `baseDir`; derive the available-skills listing and the no-prompts error from the resolved root.

No blocking findings. The change is small, well-scoped, the security-critical guard holds, and it is provably backward-compatible. The notes below are non-blocking.

### Notes (non-blocking)

**[low · test-coverage · `src/skills/_lib/prompt-loader.test.ts:112–122` · ref `e117c01` · file-state]** — The unknown-skill listing test asserts only `Available:.*demo`, but the temp fixture also creates a sibling `no-prompts/` dir (not `_`-prefixed), so it appears in the listing too. The assertion is correct but under-specified — a reader auditing the output cannot tell whether `no-prompts` is expected or spurious.
Evidence:
```
expect(msg).toMatch(/Unknown skill: missing/);
expect(msg).toMatch(/Available:.*demo/);
```
Suggestion: tighten to the exact sorted listing, e.g. `expect(msg).toMatch(/Available: demo, no-prompts/)`, to lock the output shape.

### Clean dimensions

- **security** — The path-traversal guard is unchanged and fires *before* `skillPath = join(skillsDir, name)` for every vector: `/` and `\` (literal checks), `..` (`includes('..')`), and a leading `.` (`startsWith('.')`); an absolute-path `name` is caught by the slash check, URL-encoded forms resolve to a literal segment `existsSync` won't find, and null bytes are rejected at the Node `fs` layer. `name` remains the sole traversal segment, and a trusted plugin-supplied `baseDir` does not widen it. The `'../escape'` guard test's `/illegal path component/i` matches the real thrown message (`Skill name contains illegal path components: …`).
- **api-compat** — Fully additive and verified against the live tree at the ref. The signature went from `(name: string)` to `(name: string, baseDir?: string)` (optional 2nd param). `git grep` across `src/**` (excluding tests/mocks) found **every** production call site passes `name` only — `skill-executor.ts:667,890`, `audit-fit/index.ts:318`, `diagnose/_phases/orchestrator.ts:312`, `example-template/index.ts:20`, and all seven `mint/_phases/*` callers — zero callers pass a second argument, so none break.
- **correctness** — The previously-inner `const skillsDir = join(projectRoot, 'skills')` is correctly hoisted to function scope; the removed `projectRoot`/`currentDir` locals have no dangling references; the `available` listing and the no-prompts error both read from the single hoisted `skillsDir`. No shadowing or redeclaration.
- **spec-compliance** — COMPLIANT. All named clauses are implemented at the ref: additive signature; guard fires regardless of `baseDir`; both error paths derive from the resolved root; the `baseDir` test suite (5 cases) is added. No scope creep beyond the two files.
- **perf-observability** — Synchronous `fs` reads over a bounded skill directory; no new hot-path I/O for the happy-path caller. Clean.

### What was not checked

- Read against branch HEAD `e117c01` via `gh api .../contents/{path}?ref=e117c01` (file-state) plus the inline diff. The api-compat call-site enumeration was independently re-derived by cloning the public repo, checking out the ref, and running `git grep` — confirming the "fully additive" claim live rather than from inference.
- Tests were not run locally; relied on the PR's stated green CI (Lint & Build, Test, Docs all SUCCESS on this ref) and the file-state read of the 15-case `prompt-loader.test.ts` as the test signal.
- Windows backslash semantics for a `baseDir` argument were not exercised — no risk on the POSIX CI/runtime path, where `path.join` normalises.
- Stated intent: PR #307 title + body — spec-compliance assessed and COMPLIANT.

PR #308, opened 2026-06-28
feat(plugins): load plugin entrypoints on chat/daemon/telegram surfaces (B4)

griffinwork40:
## What

Completes the plugin-entrypoint wiring across all top-level surfaces. **#298** added the plugin `main` JS entrypoint mechanism (`ensurePluginEntrypointsLoaded()`) but wired only the REPL. This wires the remaining three so plugin-contributed, **code-backed** skills are available everywhere a session is constructed:

| Surface | Placement | Why there |
|---|---|---|
| `chat` | `await` before `new AgentSession` in the one-shot action | single-pass; manifest builds synchronously in the constructor |
| `daemon` | `await` once at startup, before the session factory + scheduler | the factory closure `(config) => AgentSession` is **synchronous**, so loading inside it is the wrong layer; a plugin can own a scheduled task |
| `telegram` | `await` in `bot.start()` before `bot.launch()` | entrypoints must run before the first update constructs a per-chat session |

Subagents inherit the parent process registry, so no child wiring is needed.

## Why this placement

The skill manifest is assembled **synchronously in the `AgentSession` constructor**, so a plugin's `registerSkill()` side-effects must already have run before construction. There is no single shared pre-construction chokepoint (sessions are built at ~7 sites), which is why each surface awaits the shared helper. The helper is **idempotent + process-global + non-fatal per plugin**, so each call is cheap and a no-op without plugins.

## Tests

- **`bot.test.ts`** — new test locks the ordering invariant: `ensurePluginEntrypointsLoaded()` is awaited **before** `bot.launch()`.
- **`chat.mcp.test.ts`** — stubs `skill-bridge` (a new heavy dependency of `chat.ts` — the real impl scans the filesystem for plugin roots), matching how the test already stubs the sibling executors to isolate MCP wiring. Without the stub the MCP test hit the real plugin scan.

## Verification

- `tsc --noEmit` clean
- `src/telegram/` + `src/cli/commands/` full sweep: **1599 tests pass** (incl. the new ordering test and the fixed MCP tests)

## Roadmap

Part of the moat-as-plugin track. With #298 (entrypoint) + #305 (public runtime API) + #307 (`loadSkillPrompts(baseDir?)`) + this (all surfaces wired), an out-of-tree plugin can register skills, resolve their prompts, and have them load on every surface. Next: the extraction itself.

### DIFF
--- a/src/cli/commands/chat.mcp.test.ts
+++ b/src/cli/commands/chat.mcp.test.ts
@@ -99,6 +99,14 @@ vi.mock('../../agent/tools/compose-executor.js', () => ({
   ComposeExecutor: vi.fn().mockImplementation(() => ({})),
 }));
 
+// chat.ts awaits ensurePluginEntrypointsLoaded() before session construction
+// (plugin `main` entrypoints must run before the skill manifest is built). This
+// test isolates MCP wiring, so stub it to a no-op — the real impl scans the
+// filesystem for plugin roots, which is out of scope here.
+vi.mock('../../agent/tools/skill-bridge.js', () => ({
+  ensurePluginEntrypointsLoaded: vi.fn(async () => {}),
+}));
+
 vi.mock('../../agent/tools/nesting.js', () => ({
   createChildProviderFactory: vi.fn(() => ({})),
   createChildSkillExecutorFactory: vi.fn(() => ({})),
--- a/src/cli/commands/chat.ts
+++ b/src/cli/commands/chat.ts
@@ -21,6 +21,7 @@ import { SubagentManager } from '../../agent/subagent.js';
 import { SubagentExecutor } from '../../agent/tools/subagent-executor.js';
 import { SkillExecutor } from '../../agent/tools/skill-executor.js';
 import { ComposeExecutor } from '../../agent/tools/compose-executor.js';
+import { ensurePluginEntrypointsLoaded } from '../../agent/tools/skill-bridge.js';
 import { createChildProviderFactory, createChildSkillExecutorFactory } from '../../agent/tools/nesting.js';
 import { AnthropicDirectProvider } from '../../agent/providers/anthropic-direct/index.js';
 import { BUILTIN_TOOL_NAMES } from '../../agent/tools/schemas.js';
@@ -627,6 +628,13 @@ export function registerChatCommand(program: Command): void {
             ...(mcpManager !== undefined ? { mcpManager } : {}),
           });
 
+
+        // Import any plugin JS entrypoints (manifest `main`) before constructing
+        // the session: the skill manifest is assembled synchronously in the
+        // AgentSession constructor, so a plugin's registerSkill() side-effects
+        // must already have run for its code-backed skills to appear. Idempotent
+        // + non-fatal; no-op without plugins.
+        await ensurePluginEntrypointsLoaded();
 
         // Witness layer: `trace` was opened above (before executors) so
         // SkillExecutor could be wired with traceWriter; reuse it here for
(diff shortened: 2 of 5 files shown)


griffinwork40 (commented):

## Merge decision: MERGE

**Target:** feature · `src/cli/commands/chat.ts`, `src/cli/commands/daemon.ts`, `src/telegram/bot.ts` (+ 2 test files) · +59 / −0 · reviewed ref `50746e4b`
**Intent:** wire `ensurePluginEntrypointsLoaded()` into the three remaining top-level surfaces (chat one-shot, daemon startup, telegram `bot.start()`) so plugin-contributed code-backed skills load everywhere a session is constructed — #298 wired only the REPL.

No blocking findings. The change is additive (0 deletions), well-scoped, correctly placed at all three surfaces, and every load-bearing claim in the PR body was verified live at the reviewed ref. One non-blocking `low` test-isolation note below.

### Notes (non-blocking)

**[low · high-confidence · test-coverage · `src/cli/commands/chat.{stdin,post,resume,stream-json}.test.ts` + `daemon.test.ts` · ref `50746e4b` · file-state]** — Four sibling chat test suites and the daemon test exercise the real `registerChatCommand`/`registerDaemonCommand` `.action()` path without stubbing `skill-bridge`, so the real `ensurePluginEntrypointsLoaded()` runs during them (verified: `grep -c skill-bridge` = 0 in all four chat files). It is a no-op on clean CI (no plugins → `existsSync` short-circuits → zero-iteration loop), so tests do not fail — but it leaks real filesystem `existsSync`/config reads out of the test sandbox and is inconsistent with the isolation `chat.mcp.test.ts` correctly applies.
Suggestion: add `vi.mock('../../agent/tools/skill-bridge.js', () => ({ ensurePluginEntrypointsLoaded: vi.fn(async () => {}) }))` to those files, mirroring `chat.mcp.test.ts`.

### Clean dimensions

- **correctness** — All three placement invariants hold at the ref: chat.ts await at `:637` precedes `new AgentSession` at `:642` (same async closure); daemon.ts await at `:408` precedes both `buildDaemonSessionFactory` (`:414`) and `startDaemon` (`:423`), and the factory closure (`:86–88`) is genuinely synchronous, so hoisting the await is the correct layer; bot.ts await at `:313` precedes `this.bot.launch()` at `:315`, with all `.bot.action(...)` handler registrations before the await.
- **security** — No credentials, token handling, or filesystem mutation. Dynamic `import()` of plugin entrypoints is gated on install-time-trusted manifests.
- **api-compat** — Fully additive. `ensurePluginEntrypointsLoaded` was already exported (`skill-bridge.ts:265`, from #298); this PR only adds call sites — no signature change, no export removal.
- **spec-compliance** — COMPLIANT. All three named surfaces wired; REPL (already wired in #298) untouched; no scope creep (59 lines, 0 deletions).
- **perf-observability** — Idempotency guards (`loadedEntrypoints` Set + per-dir `scanCache`) make each call at-most-once-per-process and a few-microsecond no-op without plugins. A failing entrypoint is silently dropped by design (`load-entrypoints.ts`: "a broken plugin degrades to 'its skills don't load', never 'the session won't start'") — acceptable, no observability gap introduced here.
- **test isolation (partial-mock safety)** — chat.ts imports ONLY `ensurePluginEntrypointsLoaded` from skill-bridge, so the partial `vi.mock` factory in chat.mcp.test.ts blanks no other export. bot.test.ts uses a namespace import (`import * as skillBridge`) so `vi.spyOn` works, and the `['entrypoints','launch']` ordering assertion correctly locks the stated invariant.

### What was not checked

- Read against branch HEAD `50746e4b`; all cited lines verified at that ref (repo cloned + checked out at the SHA; file-state reads of chat.ts/daemon.ts/bot.ts/skill-bridge.ts/load-entrypoints.ts/plugins-scanner.ts, and the 8 load-bearing claims independently re-derived). The lone absence claim was re-verified by grep at the ref and CONFIRMED.
- Tests were not run locally; relied on the PR's stated green CI (Lint & Build, Test, Docs all SUCCESS on this ref) as the test signal.
- Stated intent: PR #308 title + body — spec-compliance assessed and COMPLIANT.
- No critical/high findings, so a second adversarial verification pass was not triggered (medium-and-below pass straight through).

PR #311, opened 2026-06-28
fix(cli): don't silently exit plan mode on /afk off

### LINKED ISSUE #199: /afk off while in plan mode silently exits plan mode with a misleading message
Follow-up from the review of #197. Severity: **low**, non-blocking.

## Behavior
Typing `/afk off` while the session is in **plan** mode (`permissionMode === 'plan'`) is not treated as a no-op. In `src/cli/slash/commands/afk.ts`, the "already off" guard checks only `permissionMode !== 'autonomous'`, which is true for `'plan'`, so it calls `toggleAfkMode(ctx, false)` -> `setPermissionMode('default')` (`src/cli/afk-mode-toggle.ts`), switching the session out of plan mode and printing "AFK mode OFF — default permissions restored".

Net effect: `/afk off` silently exits plan mode (re-enabling writes) with a message claiming AFK mode was turned off, though the user was never in AFK mode.

Note: this is symmetric — `/plan off` while in AFK mode does the inverse. Both commands normalize any non-self mode to default under their own label.

## Suggested fix
Add an explicit `'plan'` guard in the `/afk` handler's no-op branch (skip the toggle, or print "not in AFK mode"), and consider the symmetric case in `/plan`.

griffinwork40:
When `permissionMode === 'plan'`, the existing 'already off' guard
(`!desired && permissionMode !== 'autonomous'`) was true, so it called
`toggleAfkMode(ctx, false)` which invoked `setPermissionMode('default')`,
silently dropping plan mode and printing a misleading 'AFK mode OFF —
default permissions restored' message — even though the user was never
in AFK mode.

Fix: add an explicit `permissionMode === 'plan'` guard that short-circuits
before the toggle call, emits a clear warning directing the user to
`/plan off`, and returns without mutating session state.

Test: new focused case asserts toggle is not called, result is 'continue',
warn message references both 'plan mode' and '/plan off', and
permissionMode remains 'plan'.

Closes #199

### DIFF
--- a/src/cli/slash/commands/afk.test.ts
+++ b/src/cli/slash/commands/afk.test.ts
@@ -63,6 +63,23 @@ describe('/afk command', () => {
     expect(toggleAfkMode).toHaveBeenCalledWith(ctx, true);
   });
 
+  it('/afk off while in plan mode — warns, preserves plan mode, does NOT toggle', async () => {
+    const warn = vi.fn();
+    const ctx = { ...ctxWithMode('plan'), out: { warn } } as unknown as SlashContext;
+    const result = await afkCmd.handler(ctx, 'off');
+    // Must not clobber plan mode via toggleAfkMode
+    expect(toggleAfkMode).not.toHaveBeenCalled();
+    // Must return continue (no-op for the REPL loop)
+    expect(result).toBe('continue');
+    // Must surface a helpful, accurate message
+    expect(warn).toHaveBeenCalledOnce();
+    const msg: string = warn.mock.calls[0][0] as string;
+    expect(msg).toMatch(/plan mode/i);
+    expect(msg).toMatch(/\/plan off/i);
+    // permissionMode must remain 'plan' (ctx was not mutated)
+    expect((ctx as unknown as { stats: { permissionMode: string } }).stats.permissionMode).toBe('plan');
+  });
+
   it('exposes correct metadata', () => {
     expect(afkCmd.name).toBe('/afk');
     expect(afkCmd.usage).toContain('/afk');
--- a/src/cli/slash/commands/afk.ts
+++ b/src/cli/slash/commands/afk.ts
@@ -50,6 +50,15 @@ export const afkCmd: SlashCommand = {
       // Already on — suppress the no-op success line for ergonomics.
       return 'continue';
     }
+    if (!desired && ctx.stats.permissionMode === 'plan') {
+      // In plan mode — AFK has never been on, so there is nothing to turn off.
+      // Do NOT call toggleAfkMode here: it would call setPermissionMode('default')
+      // and silently drop plan mode with a misleading "AFK mode OFF" message.
+      ctx.out.warn(
+        'AFK is already off — you\'re in plan mode. Use \`/plan off\` to leave plan mode.',
+      );
+      return 'continue';
+    }
     if (!desired && ctx.stats.permissionMode !== 'autonomous') {
       // Already off — surface the affordance with a no-op toggle.
       await toggleAfkMode(ctx, false);
--- a/src/cli/slash/commands/afk.ts
+++ b/src/cli/slash/commands/afk.ts
@@ -55,7 +55,7 @@ export const afkCmd: SlashCommand = {
       // Do NOT call toggleAfkMode here: it would call setPermissionMode('default')
       // and silently drop plan mode with a misleading "AFK mode OFF" message.
       ctx.out.warn(
-        'AFK is already off — you\'re in plan mode. Use \`/plan off\` to leave plan mode.',
+        'AFK is already off — you\'re in plan mode. Use `/plan off` to leave plan mode.',
       );
       return 'continue';
     }


griffinwork40:
All evidence gathered and validated against the reviewed ref. The control-flow ordering is the crux, and it's correct:

- L49 `desired && permissionMode === 'autonomous'` → already-on no-op
- **L53 `!desired && permissionMode === 'plan'` → NEW guard (the fix)**
- L62 `!desired && permissionMode !== 'autonomous'` → already-off no-op
- L68 `toggleAfkMode(ctx, desired)` → real flip

The new guard (L53) is correctly placed *before* L62 — critical, because in plan mode `'plan' !== 'autonomous'` is also true, so without interception L62 would call `toggleAfkMode(ctx, false)` and clobber plan. This is a 26-line, 2-file, single-guard hotfix; I reviewed it fully inline with direct source reads at the reviewed ref rather than dispatching the multi-wave subagent sequence (which would add cost without adding coverage for a diff this size).

---

# Code Review — PR #311

**Target:** PR #311 `fix(cli): don't silently exit plan mode on /afk off` · `fix/afk-off-plan-mode` → `main`
**Reviewed ref:** `60afefb1676755c72c2eb84741f846542f8e9695` (branch HEAD — all citations read at this ref)
**Triage:** hotfix · 2 files · +26/−0 · **regime: light**
**Stated intent:** PR #311 title + body (add `permissionMode === 'plan'` guard that short-circuits before the toggle, warns toward `/plan off`, mutates no state; + focused test). Spec-compliance **assessed**.

## Bug confirmed real
Before this change, `/afk off` in plan mode (`desired=false`) fell through to the L62 guard (`!desired && permissionMode !== 'autonomous'` — true, since `'plan' !== 'autonomous'`), calling `toggleAfkMode(ctx, false)`. That helper runs `setPermissionMode('default')` and sets `stats.permissionMode = 'default'` (`src/cli/afk-mode-toggle.ts`), **silently dropping plan mode**, then prints `'○ AFK mode OFF — default permissions restored'` — misleading, since the user was never in AFK. The PR-body comment's claims are accurate against source.

## Findings by dimension

- **correctness** — no issues found. Read `src/cli/slash/commands/afk.ts` at `60afefb`. All branches traced: `/afk off` in plan → new guard fires (warn + `continue`, no toggle); `/afk off` in default → L62 no-op toggle (unchanged); `/afk off` in autonomous → L68 real off (unchanged); `/afk on` / bare `/afk` from plan → L68 `toggleAfkMode(ctx, true)` (documented plan→AFK replacement, unchanged). Guard ordering is correct — L53 precedes L62, the one placement that matters. `ref:60afefb · afk.ts:53-65 · citation-type: file-state`.
- **spec-compliance** — no issues found. Diff implements the stated intent exactly: explicit `=== 'plan'` short-circuit (`afk.ts:53`), warning directing to `/plan off` (`afk.ts:57-59`), early `return 'continue'` with no state mutation. No unmet intent; no scope creep — the change touches only the `/afk off`-in-plan path. `ref:60afefb · afk.ts:53-60`.
- **test-coverage** — no issues found. New test (`afk.test.ts:66-83`) asserts `toggleAfkMode` not called, result `'continue'`, `warn` called once, message matches `/plan mode/i` and `/\/plan off/i`, and `permissionMode` stays `'plan'`. Statically traced: the warning string contains literal `` `/plan off` `` so `/\/plan off/i` matches. Test will pass. `ref:60afefb · afk.test.ts:66-83`.
- **api-compat** — no issues found. No exported signature changed; `afkCmd.handler` contract unchanged. `ctx.out.warn(text)` is a valid `Writer` method (`src/cli/slash/types.ts:133`); `'plan'` is a valid `PermissionMode` (`src/agent/types/sdk-types.ts:20`).
- **security** — no issues found. Pure guard + advisory warning; no auth, secret, injection, or data-loss surface.
- **perf-observability** — no issues found. One string comparison on a cold REPL command path; `warn` is the semantically correct channel (not success/error).

### Nit (non-blocking)
- `nit · high confidence · style · afk.ts:58 · ref:60afefb · citation-type: diff-context` — backticks are redundantly backslash-escaped inside a single-quoted string: `'… Use \`/plan off\` to …'`. In a single-quoted JS string `\`` is just a literal backtick, so the output is correct (literal backticks around `/plan off`, presumably intentional for terminal display); the backslashes are simply unnecessary. Cosmetic only. *Suggestion:* drop the backslashes — `'… Use `/plan off` to …'`.

## Synthesis
Minimal, correct, well-tested hotfix that fixes a real silent state-clobber and matches its stated intent exactly. No critical/high/medium findings → `/shadow-verify` not triggered. No conflicts.

## What was not checked
- Read against branch HEAD `60afefb`; every blocking-tier citation verified by direct `git show <ref>:<file>` — none fabricated, none diff-only.
- **Did not execute the test suite** — running it requires checking out the branch (mutates the working tree; this review is read-only). Test correctness verified by static tracing of the regex against the actual warning string.
- Did not inspect issue #199 directly; confirmed its described symptom in source instead.
- Did not assess pre-existing bare-`/afk`-in-plan behavior (toggles into AFK, replacing plan) — documented and unaffected by this diff, so out of scope.

## MERGE

(Optional, your call: drop the two unnecessary backslashes on `afk.ts:58` — purely cosmetic, not merge-blocking.)

---

PR #319, opened 2026-06-28
fix(memory): wrap schema migrations in transactions to close torn-state window

### LINKED ISSUE #249: memory-store: wrap schema migrations in a transaction to close the ALTER/user_version torn-state window
## Context
Surfaced during a prior code review and deferred as out-of-scope at the time.

Each schema migration step in `MemoryStore` runs the column `ALTER` and the `user_version` stamp as two separate statements, not wrapped in a transaction:

```ts
// src/agent/memory/memory-store.ts — v3 → v4 shown; v1→v2 and v2→v3 are the same shape
if (!hasEvidence()) {
  try { this.db.exec(`ALTER TABLE facts ADD COLUMN evidence TEXT;`); }
  catch (err) { if (!hasEvidence()) throw err; }
}
this.db.pragma(`user_version = 4`);
```

If the process is killed after the `ALTER` succeeds but before the pragma fires, the DB is left with the new column present but `user_version` still at the prior version.

## Impact
Benign / self-healing — **not a regression**: on the next open, `existingVersion < N` fires again, the `has<Column>()` guard sees the column already present, skips the `ALTER`, and re-stamps `user_version`. No data loss. The pattern is pre-existing across the v1→v2, v2→v3, and v3→v4 migration steps.

## Action
Wrap each migration step (`ALTER` + version stamp) in `this.db.transaction(() => { ... })()` so the schema change and the version bump are atomic, closing the torn-state window. Both DDL and `PRAGMA user_version` are transactional in SQLite.

## Refs
- `src/agent/memory/memory-store.ts` — the `existingVersion < 2 | < 3 | < 4` migration blocks

griffinwork40:
Each migration step (ALTER TABLE + PRAGMA user_version) previously ran as
two separate unguarded statements. A crash between them left the DB in a
torn state: the new column present but user_version not yet advanced.

Wrap every migration step in this.db.transaction(() => { ... })() so the
DDL and the version stamp commit atomically. SQLite rolls back both on any
throw, so the next open re-runs the step cleanly (self-heal preserved).

Replace the try/catch idempotency pattern (v2→v3, v3→v4) with a
pre-check-only pattern inside the transaction: read PRAGMA table_info()
before attempting ALTER. If the column is already present (concurrent racer
won the slot), skip the ALTER and stamp the version — no try/catch needed.
Normal failures (disk error) throw through the transaction and roll back.

Tests:
- Updated 'swallows concurrent duplicate-column ALTER' test to reflect the
  new pre-check behavior: the racer scenario is represented by seeding the
  DB with the column already present (withActorColumn=true), verifying the
  store opens cleanly and reaches v4 without any exec spy.
- Added atomicity test: spy on pragma to throw on 'user_version = 3';
  assert the transaction rolled back (version still 2, actor column absent);
  restore mocks and assert a fresh MemoryStore self-heals cleanly to v4.

Closes #249

### DIFF
--- a/src/agent/memory/memory-store.test.ts
+++ b/src/agent/memory/memory-store.test.ts
@@ -435,25 +435,14 @@ describe('schema migration — sessions.actor (v2 → v3)', () => {
     }
   });
 
-  it('swallows a concurrent duplicate-column ALTER and still reaches v4', () => {
+  it('handles concurrent racer adding the actor column via pre-check (no try/catch) and still reaches v4', () => {
     const dbPath = join(migDir, 'memory.db');
-    seedV2Db(dbPath);
-
-    // Simulate the cross-process race: the table_info guard sees no actor
-    // column, but by the time THIS process runs the ALTER a concurrent opener
-    // has already added it, so SQLite throws "duplicate column name". The mock
-    // adds the column for real (post-condition holds) then throws.
-    const originalExec = Database.prototype.exec;
-    vi.spyOn(Database.prototype, 'exec').mockImplementation(function (
-      this: BetterSqlite3.Database,
-      sql: string,
-    ) {
-      if (/ALTER TABLE sessions ADD COLUMN actor/i.test(sql)) {
-        originalExec.call(this, sql); // the racer wins the ALTER
-        throw new Error('duplicate column name: actor');
-      }
-      return originalExec.call(this, sql);
-    } as BetterSqlite3.Database['exec']);
+    // Simulate a cross-process race: another opener has already added the actor
+    // column to the DB (user_version still 2 — the racer ran the ALTER but was
+    // killed before stamping the version). With the new pre-check-only pattern,
+    // the transaction body reads table_info, sees the column is already present,
+    // skips the ALTER, and stamps the version — no try/catch needed or present.
+    seedV2Db(dbPath, true /* withActorColumn — racer already added it */);
 
     expect(() => new MemoryStore(migDir)).not.toThrow();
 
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

## Merge decision: MERGE

**Target:** PR #319 · `fix(memory): wrap schema migrations in transactions to close torn-state window` · base `main` ← head `fix/memory-store-migration-txn` · reviewed ref `12198551` · hotfix · 2 files (`memory-store.ts` +28/−24, `memory-store.test.ts` +87/−14) · 181 effective lines · light regime · 1 review agent + citation verification.

**Stated intent (supplied — spec-compliance assessed):** wrap every schema-migration step in `this.db.transaction(() => {…})()` so the DDL (ALTER/CREATE INDEX) and the `user_version` stamp commit atomically (a crash between them previously left a torn DB: new column present, version not advanced); replace the v2→v3 / v3→v4 `try/catch` idempotency pattern with a pre-check-only pattern (`table_info` inside the transaction before the ALTER); update the concurrency test and add an atomicity (rollback + self-heal) test. Closes #249.

The core fix is mechanically correct and spec-compliant. Two non-blocking findings (one comment-accuracy, one test-coverage); neither blocks merge.

## Findings (sorted by severity; none blocking)

**[medium · high · correctness · `src/agent/memory/memory-store.ts:254–263` · ref:12198551 · file-state]** — The inline comment lists "concurrent racer added the column first → pre-check sees it → skip ALTER → stamp version → commit" as the only race scenario, but a racer that commits *after* this connection's `BEGIN DEFERRED` (which holds only a SHARED lock until the first write) is not visible to the `table_info` pre-check; the subsequent `ALTER TABLE … ADD COLUMN actor` then throws `duplicate column name`, rolls back the transaction, and the constructor propagates the throw. Self-heal is preserved (next open heals cleanly), but in this narrow window the caller of `new MemoryStore()` receives a throw — unlike the old `try/catch` path that succeeded silently. The **code is correct and self-heals as intended**; only the comment overstates the pre-check's coverage.
Evidence:
```ts
//   - Concurrent racer added the column first → pre-check sees it →
//     skip ALTER → stamp version → commit. (No try/catch needed.)
```
*Suggestion:* add a third bullet to the comment — "Racer commits *between* our BEGIN and our ALTER → ALTER throws 'duplicate column name' → transaction rolls back → version NOT stamped → next open re-runs the step (self-heal)."

**[low · high · test-coverage · `src/agent/memory/memory-store.test.ts` · ref:12198551 · file-state]** — The "concurrent racer" test exercises only the case where the column is present *before* the constructor begins (`seedV2Db(dbPath, true)`). No test covers the post-`BEGIN` race (a duplicate-column ALTER throw inside the transaction); the self-heal path is exercised by the atomicity test via a *pragma* failure, not a duplicate-column ALTER failure.
Evidence: grep for `two.*MemoryStore | concurrent.*open | race.*after` in the test file returns zero matches; the atomicity test's second phase confirms self-heal via a different trigger.
*Suggestion:* add a test that opens `MemoryStore` on a v2 DB, mocks `Database.prototype.exec` to throw `duplicate column name: actor` on the first `ALTER TABLE sessions ADD COLUMN actor`, asserts the constructor throws, then restores mocks and asserts a second open reaches v4. Non-blocking — the existing atomicity test already proves the rollback→self-heal cycle.

## Per-dimension
- **security** — `no issues found — read src/agent/memory/memory-store.ts at ref:12198551`. No user input flows into the migration SQL; `table_info` PRAGMA is read-only; no secret surface.
- **api-compat** — `no issues found`. `MemoryStore`'s public constructor signature is unchanged; the migration logic is purely internal constructor-time code with no exported-symbol change.
- **correctness** — the `medium` above (comment overstates race coverage; code self-heals correctly). Verified sound: all three `if (existingVersion < N)` blocks run in sequence for a v1 DB (`existingVersion` captured once), `user_version` participates in the transaction and rolls back, and the FTS5 `rebuild` inside the v1→v2 transaction is transactional.
- **spec-compliance** — COMPLIANT. Every named clause is implemented: each step wrapped in `db.transaction(…)()`; pre-check-only pattern replaces `try/catch` at v2→v3 and v3→v4; the concurrency test is updated to seed the column pre-present (no exec spy); the atomicity test (pragma spy → rollback assertion → self-heal assertion) is added. No scope creep.
- **test-coverage** — the `low` above (post-`BEGIN` duplicate-column race not exercised). The atomicity test asserts both `user_version == 2` **and** `actor` column absent after rollback, then confirms self-heal to v4 with both `actor` and `evidence` present — strong.
- **perf-observability** — `no issues found`. Each migration transaction is a cold-start-once event; `PRAGMA table_info()` is O(columns) and negligible; no hot-path change.

## Verified clean (independently re-derived at ref 12198551)
- **`user_version` is transactionally safe** — it is stored in the database header (page 1); page-1 writes participate in SQLite's journal and revert on `ROLLBACK`. It is not among the auto-committing pragmas (`journal_mode`, `locking_mode`). The code's assumption holds.
- **`db.transaction(fn)()` rolls back DDL** — SQLite DDL (ALTER TABLE, CREATE INDEX) is fully transactional; better-sqlite3 wraps the callback in `BEGIN`/`COMMIT`-or-`ROLLBACK`, so any throw inside `fn` rolls back all writes including the DDL and the FTS5 shadow-table writes.
- **Atomicity test asserts both conditions after rollback** — reads `user_version` (→ `2`) and `table_info(sessions)` (asserts `actor` absent); both assertions present at the reviewed ref.
- **Catch-up chain is correct** — for a v1 DB, guards `< 2`, `< 3`, `< 4` all fire and execute sequentially in one open; each transaction stamps the on-disk version before the next block reads, and not re-reading the in-memory `existingVersion` between blocks is intentional and correct.

## What was not checked
- Read against branch HEAD `12198551`; every `file:line` citation verified at that ref via `gh api .../contents/{path}?ref=12198551`.
- Tests were **not** executed (read-only review). CI on this ref is green (Lint & Build, Test, Docs, Bundle Smoke all SUCCESS); the PR's assertions were evaluated by code reading, not re-run.
- better-sqlite3 release notes not fetched; the transaction + DDL + pragma-in-transaction claims are grounded in SQLite documentation and the better-sqlite3 README.
- The fresh-DB path (`existingVersion === 0`, applies `SCHEMA_SQL` + stamps the version in two unguarded calls) is pre-existing behavior unchanged by this PR and out of scope.

**Bottom line:** a correct, spec-compliant fix — `db.transaction(fn)()` atomically commits the DDL + `user_version` stamp, both roll back on any throw, self-heal on next open is preserved, and the atomicity test independently verifies the rollback→recovery cycle. The two findings are optional polish.

*Read-only automated review — no files, commits, branches, or PR state were modified.*

PR #320, opened 2026-06-28
fix(telegram): interrupt provider turn on ANY incomplete stream exit

griffinwork40:
## Symptom

On the Telegram bot, a message would appear to do nothing; sending a **second** message would surface the **first** message's result, so users learned to send a throwaway `.` to flush the previous turn. The agent was perpetually one message behind.

## Root cause

The provider iterator is **long-lived and shared across messages** (`agent-session.ts` `providerIterator`, driven per-turn by `sendMessageStreamInternal`; the provider parks at the next-prompt boundary between turns). The Telegram streaming consumer only aborted the in-flight provider turn when the inactivity watchdog fired:

```ts
if (timedOut) { await session.interrupt?.(); }
```

On **any other** early exit — a Telegram render exception, a raw mid-stream throw, an early break — the provider kept generating into that shared iterator with **no consumer**. The user's **next** message attached a fresh consumer that drained those **stale buffered events** first. The code comment at the same site already named this ("send a `.` to recover the lost result") but only guarded the timeout path.

## Fix

Gate `interrupt()` on whether a **terminal `done`/`error` event was seen** (`sawTerminalEvent`) rather than `timedOut`:

```ts
if (timedOut || !sawTerminalEvent) { await session.interrupt?.(); }
```

This fires on **every** incomplete exit (closing all leak paths), but stays a **no-op on a clean turn**.

### Why not just drop the `if (timedOut)` guard?

Because at the `finally`, a cleanly-completed turn's generator is still **suspended at `yield done`** — its own `currentState = 'idle'` finalizer has **not run yet**, so `currentState` is still `'streaming'`. An *unconditional* `interrupt()` would therefore abort the turn that just finished. Keying the gate off the terminal event avoids that.

An `error` **event** is terminal (the provider emitted it and parked itself), so it does **not** trigger `interrupt()`.

## Tests (`src/telegram/streaming.test.ts`, 20/20)

- **NEW** — non-terminal raw-throw exit → `interrupt()` fires (the leak guard)
- **NEW** — clean `done` → `interrupt()` NOT called (happy-path safety)
- updated the error-event test framing (terminal event seen → no interrupt)
- existing watchdog-timeout → interrupt test still green

`pnpm lint` (`tsc --noEmit`) clean. Change is isolated to Telegram glue.

## Notes for reviewer

- **Residual:** the exact production trigger (which non-terminal exit fires) was not reproduced from logs; this closes the defect **class** regardless. A likely co-conspirator is a mid-stream Telegram render 400 from the formatter's unclosed-HTML bold/italic regex (`formatter.ts:127-134`) — a recommended **follow-up PR** to also remove the trigger (defense in depth).
- This bug is **Telegram-only**; the REPL/CLI uses a different streaming path.

### DIFF
--- a/src/telegram/streaming.test.ts
+++ b/src/telegram/streaming.test.ts
@@ -477,7 +477,7 @@ describe('renderSubagentFooter (bounded sub-agent progress)', () => {
   });
 });
 
-describe('inactivity watchdog (timeout → interrupt)', () => {
+describe('provider-turn interrupt on incomplete exit (stale-buffer guard)', () => {
   it('throws StreamTimeoutError and interrupts the still-running turn on total silence', async () => {
     vi.useFakeTimers();
     try {
@@ -510,14 +510,46 @@ describe('inactivity watchdog (timeout → interrupt)', () => {
     }
   }, 15_000);
 
-  it('does NOT interrupt on a provider error event — interrupt is timeout-only', async () => {
+  it('does NOT interrupt on a provider error EVENT — the turn already ended (terminal event seen)', async () => {
     const { ctx } = makeCtx();
     const session = makeSession(async function* () {
       yield { type: 'chunk' as const, chunk: { type: 'content' as const, content: 'partial' } };
       yield { type: 'error' as const, error: new Error('mid-stream boom') };
     });
     await expect(streamResponse(ctx, session, 'go')).rejects.toThrow('mid-stream boom');
-    // A real provider error already ended the turn; interrupting would be wrong.
+    // An 'error' EVENT is terminal: the provider emitted it and parked itself at
+    // the next-prompt boundary, so there is nothing to interrupt. (Contrast with
+    // a RAW throw / non-terminal exit below, which DOES require interrupt().)
+    expect((session as { interrupt: ReturnType<typeof vi.fn> }).interrupt).not.toHaveBeenCalled();
+  });
+
+  it('interrupts on a non-terminal early exit (raw throw, no done/error event)', async () => {
+    // The leak the fix closes: the consumer exits WITHOUT a terminal event
+    // (here a raw throw, standing in for a Telegram render exception or other
+    // mid-stream failure). The shared provider iterator is still live, so
+    // without interrupt() its buffered events would be drained by the user's
+    // NEXT message — the "send a '.' to recover the lost result" bug. Previously
+    // this path was NOT covered because interrupt() was gated on `timedOut` alone.
+    const { ctx } = makeCtx();
+    const session = makeSession(async function* () {
+      yield { type: 'chunk' as const, chunk: { type: 'content' as const, content: 'partial' } };
+      throw new Error('render boom'); // raw throw — NOT an 'error' OutputEvent
+    });
+    await expect(streamResponse(ctx, session, 'go')).rejects.toThrow('render boom');
+    expect((session as { interrupt: ReturnType<typeof vi.fn> }).interrupt).toHaveBeenCalledTimes(1);
+  });
+
+  it('does NOT interrupt on a clean turn that reaches done', async () => {
+    // Happy path: a terminal done event was seen, so interrupt() must be a
+    // no-op here — firing it would abort an already-completed turn (and, before
+    // iter.return() runs, currentState is still 'streaming', so the abort would
+    // NOT be swallowed). The gate must therefore key off the terminal event.
+    const { ctx } = makeCtx();
+    const session = makeSession(async function* () {
+      yield { type: 'chunk' as const, chunk: { type: 'content' as const, content: 'all good' } };
+      yield { type: 'done' as const, metadata: undefined };
+    });
+    await streamResponse(ctx, session, 'go');
     expect((session as { interrupt: ReturnType<typeof vi.fn> }).interrupt).not.toHaveBeenCalled();
   });
 });
--- a/src/telegram/streaming.ts
+++ b/src/telegram/streaming.ts
@@ -128,6 +128,13 @@ export async function streamResponse(
   // event AND by sub-agent sink activity, so the re-armed timeout does not
   // false-fire during deep fan-out.
   let timedOut = false;
+  // Set true once a terminal `done`/`error` event is processed for this turn.
+  // Gates the finally-block interrupt(): any exit WITHOUT a terminal event
+  // (watchdog timeout, a Telegram render exception, an early break) leaves the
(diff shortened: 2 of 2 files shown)


griffinwork40:
My exit-path enumeration is airtight. The grep confirms the loop (streaming.ts:332-520) has exactly these control-flow exits: `result.done` break (L337), `done`-handler break (L492, after setting the flag), `error`-handler throw (L502, after setting the flag), and two `continue`s (paused/resumed — stay in loop). The `return` at +104 is inside the countdown `setInterval` closure, not a loop exit. This completes the full picture.

---

# Code Review — PR #320

**Target:** PR #320 `fix(telegram): interrupt provider turn on ANY incomplete stream exit` · `fix/telegram-stale-buffer-flush` → `main`
**Reviewed ref:** `6206f1f9dbf50012be18a1e3449d6cb579fe4b6c` (read directly; PR head not checked out — current branch is `fix/telegram-formatter-html-nesting`)
**Triage:** hotfix · 2 files · +60/−13 (~73 lines) · **light regime**
**Change:** Adds `sawTerminalEvent` (set in the `done` and `error` OutputEvent handlers) and changes the `finally` gate from `if (timedOut)` to `if (timedOut || !sawTerminalEvent)`, so the orphaned-turn `interrupt()` fires on *every* incomplete exit, not just the watchdog-timeout path.

## Exit-path table (OLD vs NEW behavior — derived and verified at ref 6206f1f9)

| Loop exit | `sawTerminalEvent` | OLD `if(timedOut)` | NEW `if(timedOut\|\|!sawTerminalEvent)` | Correct? |
|---|---|---|---|---|
| `done` event (L492) | true | no interrupt | `false\|\|!true` → no interrupt | ✓ same — no-op desired |
| `error` event (L502) | true | no interrupt | `false\|\|!true` → no interrupt | ✓ same as old (no regression) |
| watchdog timeout | false | interrupt | `true\|\|…` → interrupt | ✓ same |
| **raw throw mid-stream** | false | **no interrupt (LEAK)** | `false\|\|!false` → **interrupt** | ✓ **the fix** |
| `result.done` natural end (L337) | false | no interrupt | → interrupt, but `currentState='idle'` → **no-op** | ✓ harmless |
| overflow-send throw (post-loop) | true | no interrupt | `false\|\|!true` → no interrupt | ✓ correct (turn done) |

The only behavioral deltas are the **raw-throw path** (now correctly closes the leak) and the **natural-completion path** (adds a `interrupt()` call that is a *verified* no-op). Every other path is byte-for-byte behavior-preserving.

## Findings by dimension

- **security** · `no issues found` — read `streaming.ts` @ 6206f1f9. Control-flow change in Telegram glue; no secret, auth, or injection surface touched.
- **api-compat** · `no issues found` — `sawTerminalEvent` is a function-local; `session.interrupt?.()` already optional-chained (`agent-session.ts:700`, `IAgentSession` declares it). No exported signature changed; no production importer affected.
- **correctness** · `no issues found` — both load-bearing claims **independently confirmed** at ref 6206f1f9:
  - **interrupt() is a true no-op off the streaming path** — `agent-session.ts:700-702`: `if (this.currentState !== 'streaming' && this.currentState !== 'processing') return;` (citation-type: file-state).
  - **the `done`-event path is still `'streaming'` at the consumer `finally`** — `sendMessageStreamInternal` is suspended at `yield output` (`agent-session.ts:608`); its post-yield `break` (L610) and `currentState='idle'` finalizer (L613-614) have not run. This is exactly why the gate keys off `sawTerminalEvent` instead of dropping the guard, and why `interrupt()` must run **before** `iter.return()`.
- **spec-compliance** · `compliant` — stated intent (PR #320 title + body) **assessed**. The diff implements precisely the described gate (`timedOut || !sawTerminalEvent`), fires on incomplete exits, stays a no-op on clean turns, and deliberately skips the `error` event. No unmet clause; no scope creep — the comment rewrite documents the same site. The named residual (un-reproduced production trigger; formatter-regex co-conspirator) is explicitly deferred to a follow-up — not a requirement of this PR. (Corroboration: the follow-up appears to be in flight — current local branch is `fix/telegram-formatter-html-nesting`.)
- **test-coverage** · 1 nit — new tests are meaningful: `makeSession` wires `interrupt: vi.fn()` (`streaming.test.ts:35`), and the generator-driven consumer exercises raw-throw → `interrupt` called once (L33-47), clean `done` → not called (L49-61), and `error` event → not called (L18-31). Watchdog-timeout test still green.
- **perf-observability** · `no issues found` — one boolean + two assignments on the hot path; negligible. No logging gap introduced.

### Nit (non-blocking)
- `nit · high confidence · test-coverage · src/telegram/streaming.test.ts (interrupt describe block) · ref:6206f1f9 · citation-type:file-state` — **finding:** the `result.done` natural-completion exit (consumer `if (result.done) break`, `streaming.ts:337`) has no dedicated test; its no-op safety rests on the real `currentState` guard, which the `vi.fn()` `interrupt` stub does not model. **evidence:** `interrupt: vi.fn(),` (`streaming.test.ts:35`). **suggestion:** add a case whose generator returns without emitting `done` to document this path's intended production no-op (low value — purely defensive).

## What was not checked
- **Refs:** all source read against PR head **6206f1f9** via `git show` (PR head not checked out); citations verified at that ref. No `diff-only` or fabricated citations — the two top correctness claims were re-read directly from `agent-session.ts` at 6206f1f9, not from the diff hunk.
- **Did not run the test suite** (read-only review; PR branch not checked out). The PR claims 20/20 + clean `tsc`; tests validated by **inspection only**.
- **Q2 residual:** I confirmed the `error`-event path is *behavior-identical to the pre-PR code* (old code never interrupted on `error` either), so even if some provider `error` emission left an iterator live, that would be a pre-existing condition — **not a regression introduced here**. I did not exhaustively trace every provider error-emission site myself (the correctness sub-agent did, returning ALWAYS-TERMINAL with `query.ts` citations, high confidence).
- **No critical/high finding present → `/shadow-verify` not required** by the skill gate; the load-bearing facts were nonetheless independently re-derived by direct file read.

## Merge decision: **MERGE**

A tight, correct, well-targeted hotfix. The gate change closes the stale-buffer leak on the raw-throw/early-exit path while remaining provably behavior-preserving on every other path (the only added call — on natural iterator completion — is a verified no-op via the `currentState` guard). The two subtle invariants the fix depends on are real and confirmed at the reviewed ref. Tests exercise the meaningful paths. The single nit is optional and non-blocking.

---

PR #321, opened 2026-06-28
fix(telegram): never emit improperly-nested HTML from interleaved emphasis

griffinwork40:
## Symptom

Telegram messages with **interleaved/overlapping** markdown emphasis fail or render degraded — Telegram returns HTTP 400 `can't parse entities`.

## Root cause (empirically confirmed)

`markdownToTelegramHtml` runs bold then italic as independent regex passes. When emphasis markers overlap, the passes produce **improperly-nested** tags. Verified with a tag-balance probe:

```
**a _b** c_   ->  <b>a <i>b</b> c</i>     (mismatch: </i> expected, </b> found)
__a *b__ c_   ->  <b>a <i>b</b> c</i>     (same)
```

`<b>...<i>...</b>...</i>` is invalid HTML; Telegram's parser rejects it with a 400. (The earlier hypothesis was "unclosed tags" — the probe showed it is actually **mis-nesting**; unmatched markers stay literal and are harmless.)

## Fix

A final validation **safety net** in `markdownToTelegramHtml`: after conversion, a stack check verifies tag nesting. If the emphasis ended up mis-nested, drop **only** the emphasis tags (`<b>` / `<i>` / `<s>`).

Why this is safe and sufficient: code/pre/link tags are emitted as **atomic, always-balanced** units (code & fences via placeholders, links in one replace), so any imbalance is necessarily from emphasis. Stripping emphasis is therefore **guaranteed** to yield valid HTML. The rare gnarly message renders without bold/italic (still readable) instead of 400ing — and **properly-nested** emphasis (`*a **b** c*` -> `<i>a <b>b</b> c</i>`) is left untouched (no false strip).

## Tests (`src/telegram/formatter.test.ts`, 96/96)

4 new:
- both interleaved orders (`**`/`_` and `__`/`*`) now produce valid, balanced HTML
- the safety net **preserves code spans** while dropping mis-nested emphasis
- **properly-nested** emphasis is not falsely stripped (regression guard)

`pnpm lint` (`tsc --noEmit`) clean.

## Relationship to #320

Companion to #320 (the stale-buffer interrupt fix). That PR **contains** the consequence of any mid-stream render throw; this PR removes a **likely trigger**. Independent, non-overlapping diffs — either can merge first. Telegram-only.

### DIFF
--- a/src/telegram/formatter.test.ts
+++ b/src/telegram/formatter.test.ts
@@ -617,3 +617,50 @@ describe('formatCompactNoop — new reasons', () => {
     expect(result.toLowerCase()).toContain('failed');
   });
 });
+
+describe('markdownToTelegramHtml — mis-nested emphasis safety net', () => {
+  // Stack check mirroring Telegram's HTML parser: every tag must be closed and
+  // properly nested. A mis-nested run like "<b>..<i>..</b>..</i>" → 400.
+  const tagsBalanced = (html: string): boolean => {
+    const stack: string[] = [];
+    const re = /<(\/?)([a-zA-Z][a-zA-Z0-9-]*)\b[^>]*>/g;
+    let m: RegExpExecArray | null;
+    while ((m = re.exec(html)) !== null) {
+      const tag = (m[2] ?? '').toLowerCase();
+      if (m[1] === '/') { if (stack.pop() !== tag) return false; }
+      else stack.push(tag);
+    }
+    return stack.length === 0;
+  };
+
+  test('interleaved ** / _ markers never emit improperly-nested tags', () => {
+    // Pre-fix this produced "<b>a <i>b</b> c</i>" → Telegram 400 "can't parse entities".
+    const out = markdownToTelegramHtml('**a _b** c_');
+    expect(tagsBalanced(out)).toBe(true);
+    expect(out).not.toContain('<b>');
+    expect(out).not.toContain('<i>');
+    expect(out).toBe('a b c');
+  });
+
+  test('interleaved __ / * markers never emit improperly-nested tags', () => {
+    const out = markdownToTelegramHtml('__a *b__ c*');
+    expect(tagsBalanced(out)).toBe(true);
+    expect(out).toBe('a b c');
+  });
+
+  test('safety net preserves code/link tags while dropping only mis-nested emphasis', () => {
+    const out = markdownToTelegramHtml('**a _b** `keep=1` c_');
+    expect(tagsBalanced(out)).toBe(true);
+    expect(out).toContain('<code>keep=1</code>'); // code span survives
+    expect(out).not.toContain('<b>');
+    expect(out).not.toContain('<i>');
+  });
+
+  test('properly-nested emphasis is left untouched (no false strip)', () => {
+    // Italic fully containing bold is valid HTML and must be preserved verbatim.
+    expect(markdownToTelegramHtml('*a **b** c*')).toBe('<i>a <b>b</b> c</i>');
+    // Separate spans + identifiers are unaffected by the safety net.
+    expect(markdownToTelegramHtml('**A** and *b* and snake_case'))
+      .toBe('<b>A</b> and <i>b</i> and snake_case');
+  });
+});
--- a/src/telegram/formatter.ts
+++ b/src/telegram/formatter.ts
@@ -159,7 +159,38 @@ export function markdownToTelegramHtml(text: string): string {
   // 10. Restore fenced code blocks
   out = out.replace(/\x02FENCED(\d+)\x03/g, (_m, i: string) => fencedBlocks[Number(i)] ?? _m);
 
+  // 11. Safety net: interleaved/overlapping emphasis markers (e.g. "**a _b** c_")
+  // can yield improperly-NESTED tags like "<b>a <i>b</b> c</i>", which Telegram
+  // rejects with HTTP 400 "can't parse entities" — failing or degrading the send.
+  // Code/pre/link tags are emitted as atomic, always-balanced units, so any
+  // imbalance is necessarily from emphasis (<b>/<i>/<s>); drop just those to
+  // recover guaranteed-valid, readable HTML instead of breaking the message.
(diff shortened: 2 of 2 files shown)


griffinwork40:
Verified at `streaming.ts:137–207`: `markdownToTelegramHtml` runs on the **full accumulated buffer** (line 138 comment confirms), not per-delta fragments — so there's no mid-tag-fragment hazard. More importantly, every send/edit site already wraps the call in a `try/catch` that catches the exact `400 "can't parse entities"` and falls back to raw plain text (lines 149–156, 173–188, 201–206). Step 11 therefore introduces **no regression** and is strictly better than the status quo: where the old path caught a 400 and dropped *all* formatting to raw text, the new safety net pre-empts the 400 and preserves code/links/structure, dropping only emphasis. The Wave 1 agent's one "not checked" item is resolved.

This also recontextualizes the empty-fence finding: a message combining an empty fence with mis-nested emphasis would have **400'd → raw plain text** pre-PR; post-PR it keeps valid HTML and loses only the label's italic — an improvement, and consistent with the PR's documented "renders without bold/italic" degradation. Downgrading it to `low`.

No `critical`/`high` findings survive, so `/shadow-verify` is not triggered (it gates on top-severity only). Here is the synthesis.

---

# Review — PR #321 · `fix(telegram): never emit improperly-nested HTML from interleaved emphasis`

**Target:** PR #321, base `main` ← head `fix/telegram-formatter-html-nesting` · **Reviewed ref:** `847210b` · **Type:** hotfix · **Files:** 2 · **Lines:** +78/−0 · **Regime:** light · **Stated intent:** present (PR title+body) — spec-compliance assessed.

**Summary:** Adds a post-conversion safety net (step 11) to `markdownToTelegramHtml`: a stack-based tag-balance check (`telegramHtmlTagsBalanced`) that, on detecting mis-nesting, strips only emphasis tags (`<b>`/`<i>`/`<s>`) via `/<\/?[bis]>/g`. Plus 4 regression tests.

## Findings (sorted; none blocking)

**`low · medium · correctness · src/telegram/formatter.ts:111 · ref:847210b · file-state`**
A message that combines an empty fenced block with separate mis-nested emphasis loses the empty-fence label's italic styling, because the label is emitted as `<i>…</i>` and the strip removes all `<i>`.
Evidence: `fencedBlocks.push(\`<i>${label}</i>\`);`
Suggestion: emit the empty-fence label as `<code>${label}</code>` (survives the strip). Optional — the text is preserved either way, and this case would have rendered as raw plain text pre-PR, so it is a net improvement and matches the PR's documented degradation. Non-blocking.

**`low · high · test-coverage · src/telegram/formatter.test.ts:621 · ref:847210b · diff-context`**
No test exercises the `<s>` (strikethrough) strip path though the regex includes `s`; all 4 tests use only `**`/`__`/`*`/`_`.
Evidence: strip regex `/<\/?[bis]>/g` includes `s`; tests cover only b/i markers.
Suggestion: add `markdownToTelegramHtml('~~a *b~~ c*')` → expect balanced, no `<s>`/`<i>`. (Code is correct — step 6 emits `<s>` at line 141 and the strip covers it; this is a coverage gap only.) Non-blocking.

## Per-dimension
- **correctness** — Core claim verified sound at source: step 1 (lines 90–93) escapes all user `<`/`>` *before* any tag is emitted, and `<pre>`/`<code>`/`<a>` are each emitted as one balanced atomic unit (lines 113, 122, 146–150), so the only imbalance source is emphasis; stripping all `b/i/s` is guaranteed to leave balanced output. No false-positive path on properly-nested emphasis. The one residual edge (empty-fence label) is the `low` above.
- **spec-compliance** — Satisfies stated intent ("never emit improperly-nested HTML from interleaved emphasis"): the stack check enforces proper nesting and the strip recovers valid HTML; no unrequested scope creep. The companion-to-#320 framing in the body is accurate (independent diff).
- **api-compat** — `telegramHtmlTagsBalanced` is module-private (no `export`); grep shows references only at `formatter.ts:168,181`. No external surface, no breaking change.
- **security** — No injection/secret/auth surface; the change reduces malformed output reaching Telegram. `no issues found — read src/telegram/formatter.ts at 847210b`.
- **perf** — Two extra O(n) regex passes per message on the Telegram send/edit path, n ≤ 4096; runs after the throttle gate (`streaming.ts:143`). Negligible. `no issues found`.
- **test-coverage** — 4 tests cover both interleave orders, code-span survival, and a no-false-strip regression; gap is the `<s>` path (`low` above).

## What was not checked
- Read against branch HEAD `847210b` (clean tree); all citations verified at that ref via `git show`.
- Did **not** execute the test suite (read-only review) — PR claims 96/96 + `tsc` clean; not independently run.
- Did not assess Telegram-parser quirks beyond balanced nesting (attribute/CDATA semantics); the stack check models nesting only, which matches the documented 400 trigger.
- Pre-existing `splitLongMessage` mid-tag splitting (`streaming.ts:198,502`) is unrelated to this diff and unchanged by it.

## Merge decision: **MERGE**
Correct, well-reasoned, and strictly safer than the prior 400-fallback path; the two `low` items are optional polish, not blockers.

(Read-only review — no files, commits, or PR state modified.)

**Done**
- Reviewed PR #321 at reviewed ref `847210b` via the `review` skill (light regime): triaged the diff, captured stated intent from PR title+body, dispatched one independent review agent, then verified citations and the streaming-path concern inline against source.
- Evidence: findings cite `src/telegram/formatter.ts:90–93,111,113,122,141,146–150,168,181` and `src/telegram/streaming.ts:137–207`, all read at ref `847210b`; diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/a14df5c1-4e5b-459f-badc-c01c42f5a950/pr-321.diff`.
- Verdict: **MERGE** — zero blocking findings; two non-blocking `low` items (empty-fence label loses italic in a rare combined case; missing `<s>` strikethrough test). Core correctness claim independently confirmed sound.
- No world state changed: read-only review, no files/commits/PR mutations. Test suite not executed (read-only); PR's 96/96 + lint-clean claims not independently re-run.

---

PR #322, opened 2026-06-28
fix(update): restart launchd services + notify on stale manual bot after install

griffinwork40:
## Summary

A long-running telegram bot / daemon keeps the **old** module graph in memory after `npm install -g agent-afk@latest` (via `afk update`, the auto-update path, or a manual install) swaps the files on disk — so they silently run stale code until restarted. The previous `postinstall` handling only SIGTERM'd a manually-started bot (killing it with nothing to relaunch it) and never touched the daemon or any launchd-supervised service.

This change makes `scripts/postinstall.mjs` — the one chokepoint npm runs on **every** global install — bring running services onto the freshly-installed code:

- **`restartLaunchdServices()`** — `launchctl kickstart -k gui/<uid>/<label>` for any installed `com.afk.telegram` / `com.afk.daemon` LaunchAgent, so it re-execs the new on-disk entrypoint. Reuses the exact invocation `afk service restart` already uses (`src/cli/commands/service.ts`). macOS-only; fail-open (absent plist skipped, launchctl errors swallowed — the install never fails).
- **`isManualBotRunning()`** — replaces the silent SIGTERM. A manually-started bot (the only writer of `bot.pid`; launchd-supervised bots don't write it — see `src/service/launchd/paths.ts`) is now **notified** with an `afk telegram restart` hint instead of killed. Read-only probe — does not unlink a stale pid file.

`killStaleDaemon` is retained as a tested utility but no longer wired into the flow. The `↻ Restarted AFK service(s)` confirmation flows through `afk update`'s inherited stdio, so it's visible without touching `update.ts`. `dist/postinstall.mjs` is a `copyFileSync` of this source (`scripts/build-dist.mjs`), so the fix ships from the first release that includes it.

## Test results

- `pnpm lint` (tsc --noEmit, strict): **passed** (exit 0)
- `pnpm test` (full suite): **10283 passed / 2 failed / 14 skipped**. The 2 failures are pre-existing and unrelated to this change — `anthropic-direct.test.ts > compact()` and `config.test.ts > model slots … built-in tier bindings` (a known env/MLX slot-binding bleed). Neither test imports the changed file; this PR touches only `scripts/postinstall.mjs` (outside the `src/` compile graph) and its test. **Zero new failures introduced.**
- Scoped suites that actually exercise the change: `tests/postinstall.test.ts` (16) + `src/cli/postinstall.test.ts` (8) = **24 passed**, including +11 new cases (`isManualBotRunning` ×6, `restartLaunchdServices` ×5) using injected exists/exec/probe fns.
- Live `node scripts/postinstall.mjs` against a fake empty `$HOME` exits 0 with no real services touched.

## Verification

Adversarial read-only verifier re-derived the load-bearing claims from the diff (not the description):

- **Restarts installed launchd services via `kickstart -k`, skips absent plists** — CONFIRM (`postinstall.mjs` restartLaunchdServices: plist gate + argv; test asserts exact `['kickstart','-k','gui/1000/com.afk.telegram']`).
- **Manual-bot SIGTERM removed; notify instead, no pid-file mutation** — CONFIRM (`killStaleDaemon` not invoked in the main block; `isManualBotRunning` probes with signal 0 and a test asserts the stale pid file is preserved).
- **Fail-open, macOS-gated, always exits 0** — CONFIRM (`process.platform === 'darwin'` gate, both new paths try/catch-wrapped, unconditional final `process.exit(0)`).

Verdict: **SAFE TO MERGE.** One benign note: the EPERM branch of `isManualBotRunning` treats a bot owned by another user as not-running (conservative; not a regression — cross-user bot collision is a non-scenario given a single bot token).

## Out of scope

- **Linux/systemd** — AFK services are macOS-launchd-only by design; the manual-bot notify still works cross-platform.
- **`npm install --ignore-scripts`** — skips postinstall entirely (same pre-existing limitation as the PATH-gap notice).
- **No `update.ts` edit** — intentional: postinstall runs inside `afk update`'s `runNpmInstall` (inherited stdio), so adding restart logic there too would double-restart.

### DIFF
--- a/scripts/postinstall.mjs
+++ b/scripts/postinstall.mjs
@@ -10,8 +10,8 @@
  * Always exits 0 — never fails an install.
  */
 
-import { execSync } from 'node:child_process';
-import { readFileSync } from 'node:fs';
+import { execSync, execFileSync } from 'node:child_process';
+import { existsSync, readFileSync } from 'node:fs';
 import { homedir } from 'node:os';
 import { join } from 'node:path';
 
@@ -35,6 +35,11 @@ export function detectPathGap(prefix, pathEnv) {
  * given file. All errors are silently discarded — this is best-effort cleanup
  * that must never fail an install.
  *
+ * NOTE: the install flow no longer calls this. SIGTERM'ing a manually-started
+ * bot left it dead with nothing to relaunch it (postinstall cannot reliably
+ * spawn a detached long-lived process). The main block now NOTIFIES instead
+ * (see isManualBotRunning). Retained as a tested, reusable utility.
+ *
  * @param {string} pidFilePath   - Full path to the PID file to read.
  * @param {function} [killFn]    - Injectable kill function; defaults to process.kill.
  *                                 Pass a stub in tests to avoid needing process.kill.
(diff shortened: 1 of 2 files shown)


griffinwork40:
I now have everything needed. The `triggerAutoUpdate` path spawns npm with `detached: true` + `child.unref()` (update-checker.ts:278-282), and the default policy is `'notify'` (opt-in auto-update) — which de-risks the one substantive finding. Synthesis follows.

---

# Code Review — PR #322

**`fix(update): restart launchd services + notify on stale manual bot after install`**
`afk/telegram-daemon-sync → main` · +286/−9 · `scripts/postinstall.mjs`, `tests/postinstall.test.ts`

## Verdict: ✅ **APPROVE** — no blocking issues

A well-scoped, correctly-fixing change. It replaces a harmful SIGTERM-the-manual-bot step (which left the bot dead with nothing to relaunch it) with two safe behaviors: in-place `launchctl kickstart -k` of supervised services, and a *notify-only* probe for a manually-started bot. Conventions are verified against canonical source, the security posture improved, and the tests are thorough.

> ⚠️ **Verification note:** the working tree is checked out on `fix/telegram-formatter-html-nesting`, so the on-disk `postinstall.mjs` is the **pre-PR baseline**. This review is based on the diff artifact + reads of the *unchanged* canonical files. I did **not** run `tests/postinstall.test.ts` — against the current checkout it would test old code and mislead. Suggest CI confirm the suite on the PR branch.

---

## Verified correct (the load-bearing claims)

The `.mjs` can't import the compiled TS helpers, so it hand-duplicates launchd conventions. Every one matches canonical source:

| Convention | PR value | Canonical (file:line) | |
|---|---|---|---|
| Label format | `com.afk.<name>` | `paths.ts:51` | ✅ |
| Service names (exhaustive) | `telegram`, `daemon` | `paths.ts:44,47` | ✅ |
| Plist dir | `~/Library/LaunchAgents/<label>.plist` | `paths.ts:55-56` | ✅ |
| GUI domain + uid fallback | `gui/<getuid() ?? 501>` | `paths.ts:73-74` | ✅ |
| Restart primitive | `launchctl kickstart -k gui/<uid>/<label>` | `service.ts:196` (identical to `afk service restart`) | ✅ |

- **Core assumption holds:** "non-null `bot.pid` ⇒ a *manual* bot." The launchd plist runs the bot entrypoint directly, never `afk telegram start` → `startBot()`, so launchd never writes `bot.pid`. **No false-warning bug on launchd installs.** (`bot.ts` writes `~/.afk/state/telegram/bot.pid`; plist `ProgramArguments` bypasses it.)
- **Advice is valid:** `afk telegram restart` exists (`telegram.ts:146`).
- **Security improved:** switched launchctl to `execFileSync` (argv array, **no shell**) with an 8 s timeout. Args are hardcoded labels + numeric uid — zero injection surface.

## Strengths

- Read-only by construction: `isManualBotRunning` deliberately does **not** unlink a stale PID file (unlike `manager.isRunning()`), correctly avoiding runtime-state mutation from an install script — and there's a dedicated test asserting it (`postinstall.test.ts:266`).
- Fail-open throughout: missing plist → skip; launchctl error → swallow + continue to the next service; outer try/catch belt-and-suspenders; `process.exit(0)` preserved.
- Excellent unit coverage: 6 tests for `isManualBotRunning` (missing/garbage/alive/ESRCH/EPERM/no-delete) and 5 for `restartLaunchdServices` (no-plist, partial, both, error-swallow-and-continue, custom labels) — all via injected deps with exact-argv assertions.

---

## Non-blocking notes (optional follow-ups)

1. **LOW — daemon auto-update self-restart.** When `updatePolicy: 'auto'` and the daemon itself triggers the update, `kickstart -k com.afk.daemon` restarts the very daemon that launched the install. **Well-mitigated and effectively by-design:** npm is spawned `detached:true`+`unref()` (`update-checker.ts:278-282`) so it survives the daemon being killed; auto-update is opt-in (default `'notify'`, `config.ts:332`); postinstall files are already on disk before kickstart fires; launchd relaunches on new code (the goal). Residual: interrupts any in-flight scheduled task — identical to `afk service restart`. Consider one line of comment marking this intentional; no code change required.

2. **NIT — `killStaleDaemon` is now unused by the main block.** Retained as an exported, tested utility and explicitly documented in the new NOTE comment (diff lines 20-23). Harmless; flagging only for awareness.

3. **MINOR / scope — asymmetry.** A *manually-run* `afk daemon` (foreground, not launchd) is neither restarted nor notified, unlike the manual telegram bot. Likely intentional (PR scope is "launchd services + manual bot"), but a symmetric "daemon still on old version" notice could be a future nicety.

4. **MINOR — untested thin wrappers.** Default `execFn`/`existsFn`/`probeFn` (real `launchctl`/`existsSync`/`process.kill`) and the `isMain` block aren't exercised — acceptable, since testing them needs real OS state and the pattern matches the existing file.

## Done
- Reviewed PR #322 (2 files, +286/−9) read-only across correctness, security, api-compat, test-coverage, and observability; verdict **APPROVE**, no blockers.
- Evidence: diff artifact `…/pr-322.diff`; conventions cross-checked against `src/service/launchd/paths.ts:44-74`, `src/cli/commands/service.ts:196`, `src/telegram/bot.ts`, `src/cli/commands/telegram.ts:146`, `src/cli/update-checker.ts:268-286`, `src/cli/config.ts:332` (two parallel verification subagents + direct reads).
- World unchanged — no files edited, committed, or pushed; no PR comment posted (skill is read-only).
- Deferred: test suite not executed because the working tree is on `fix/telegram-formatter-html-nesting`, not the PR branch (on-disk `postinstall.mjs` is the pre-PR baseline) — recommend CI run `tests/postinstall.test.ts` on the PR branch to confirm green.

---

PR #324, opened 2026-06-28
feat(plugins): inject host runtime API into plugin entrypoints (fixes singleton trap)

griffinwork40:
## What

Adds **host-API injection** to the plugin `main` entrypoint mechanism so a code-backed, out-of-tree plugin can register skills that actually reach the host's registry.

If a plugin's `main` module exports a **callable default**, the host now invokes it after import:

```ts
// plugin main (afk-moat/dist/index.js)
import type { PluginApi } from 'agent-afk';
export default ({ registerSkill, loadSkillPrompts }: PluginApi) => {
  registerSkill({ name: 'forge', /* … */ });
};
```

The host passes a `PluginApi` assembled from **its own** modules (`{ registerSkill, listSkills, getSkill, loadSkillPrompts }`). Modules with no default export still run their top-level side-effects at import — fully backward-compatible.

## Why (the bug this fixes)

#298 loads a plugin `main` purely for side-effects, which forces a code-backed plugin to `import { registerSkill } from 'agent-afk'` itself. But the skill registry is a **module-level singleton** (`src/skills/index.ts`). A plugin that resolves its **own** copy of `agent-afk` — its `node_modules`, a bundle, or a version-skewed/**marketplace-cloned install with no shared `node_modules`** — calls a `registerSkill` backed by a *different* registry than the host reads. The skill silently never appears. No error.

This is the blocker for distributing the moat (forge/harvest) as an out-of-tree code-backed plugin: marketplace plugins are git-cloned with no `node_modules`, so bare-specifier resolution can't reach the host instance. Injecting the API from the host guarantees the same module instance regardless of install layout.

## Design notes

- `PluginApi` is typed via `typeof import(...)` so the injected signatures **cannot drift** from source; exported from `src/index.ts` (verified it survives into the shipped `dist/index.d.ts`).
- Assembled in `skill-bridge`'s `ensurePluginEntrypointsLoaded` (already in the host module graph) and passed down through `loadPluginEntrypoints` — avoids an import cycle into `load-entrypoints`.
- Stateless helpers (`env`, paths getters, `describeFailure`) are intentionally **omitted** — they hold no singleton state, so a plugin can import them directly; only the registry *must* be shared.

## Tests

- **Round-trip proof** (the coverage that was missing): a plugin's default export calls the injected `registerSkill`, and the host's `getSkill`/`listSkills` sees it — exercised against the **real** registry, not a spy.
- **Back-compat**: a no-default side-effect-only plugin still runs at import even when `pluginApi` is supplied.

## Verification

- `tsc --noEmit` clean
- `src/agent/plugins/` + `src/agent/tools/` + `chat.mcp` sweep: **1552 pass / 2 skip**, incl. the 2 new tests
- `PluginApi` present in regenerated `dist/index.d.ts`

### DIFF
--- a/src/agent/plugins/load-entrypoints.test.ts
+++ b/src/agent/plugins/load-entrypoints.test.ts
@@ -12,6 +12,8 @@ import { tmpdir } from 'os';
 import { pathToFileURL } from 'url';
 import { loadPluginEntrypoints, _resetLoadedEntrypoints } from './load-entrypoints.js';
 import type { SdkPluginConfig } from '../types/sdk-types.js';
+import { registerSkill, getSkill, listSkills, _resetRegistry } from '../../skills/index.js';
+import { loadSkillPrompts } from '../../skills/_lib/prompt-loader.js';
 
 let dir: string;
 
@@ -94,5 +96,53 @@ describe('loadPluginEntrypoints', () => {
       },
     });
     expect(seen).toEqual([pathToFileURL(abs).href]);
+  });
+
+  it('injects the host API into a default-export entrypoint so a plugin registers into the HOST registry (round-trip)', async () => {
+    // The load-bearing proof: a plugin's default export, invoked with the host
+    // PluginApi, must reach the SAME singleton registry the host reads via
+    // getSkill/listSkills. Without injection, a plugin importing its own copy of
+    // the package would write to a different registry instance — a silent no-op.
+    _resetRegistry();
+    const skillName = `__afkRtProof_${Math.random().toString(36).slice(2)}`;
+    writeFileSync(
+      join(dir, 'entry.mjs'),
+      `export default (api) => {\n` +
+        `  api.registerSkill({\n` +
+        `    name: ${JSON.stringify(skillName)},\n` +
+        `    description: 'round-trip proof',\n` +
+        `    handler: async () => ({}),\n` +
+        `  });\n` +
+        `};\n`,
+    );
+    const plugins: SdkPluginConfig[] = [{ type: 'local', path: dir, main: 'entry.mjs' }];
+
+    await loadPluginEntrypoints(plugins, {
+      pluginApi: { registerSkill, listSkills, getSkill, loadSkillPrompts },
+    });
+
+    try {
+      expect(listSkills()).toContain(skillName);
+      expect(getSkill(skillName).description).toBe('round-trip proof');
+    } finally {
+      _resetRegistry();
+    }
+  });
+
+  it('still runs top-level side-effects when an entrypoint has no default export, even with pluginApi supplied (back-compat)', async () => {
+    const key = `__afkSideEffectProof_${Math.random().toString(36).slice(2)}`;
+    writeFileSync(join(dir, 'entry.mjs'), `globalThis[${JSON.stringify(key)}] = true;\n`);
+    const plugins: SdkPluginConfig[] = [{ type: 'local', path: dir, main: 'entry.mjs' }];
+    const store = globalThis as unknown as Record<string, unknown>;
+    try {
+      await expect(
+        loadPluginEntrypoints(plugins, {
+          pluginApi: { registerSkill, listSkills, getSkill, loadSkillPrompts },
+        }),
+      ).resolves.toBeUndefined();
+      expect(store[key]).toBe(true);
+    } finally {
+      delete store[key];
+    }
   });
 });
--- a/src/agent/plugins/load-entrypoints.ts
+++ b/src/agent/plugins/load-entrypoints.ts
@@ -18,6 +18,29 @@ import { pathToFileURL } from 'url';
 import type { SdkPluginConfig } from '../types/sdk-types.js';
 
 /**
+ * Host runtime API injected into a plugin entrypoint's default-export function.
+ *
+ * Invariant: a code-backed plugin MUST register through this object, not by
+ * importing `agent-afk` itself. The skill registry is a module-level singleton
+ * (`src/skills/index.ts`), so a plugin that imports its OWN copy of `agent-afk`
+ * (its `node_modules`, a bundle, or a version-skewed install — the default for a
+ * marketplace-cloned plugin with no shared `node_modules`) would call a
+ * `registerSkill` backed by a DIFFERENT registry than the host reads → the skill
+ * silently never appears. Receiving these functions from the host guarantees the
+ * same module instance regardless of how the plugin was installed.
+ *
+ * Stateless helpers (`env`, the `paths` getters, `describeFailure`) are safe for
+ * a plugin to import directly — they hold no singleton state — and are
+ * intentionally omitted to keep this surface minimal. The type is derived via
+ * `typeof import(...)` so the injected signatures cannot drift from the source.
+ */
+export type PluginApi = Pick<
+  typeof import('../../skills/index.js'),
+  'registerSkill' | 'listSkills' | 'getSkill'
+> &
+  Pick<typeof import('../../skills/_lib/prompt-loader.js'), 'loadSkillPrompts'>;
+
+/**
  * Resolved-entrypoint paths already imported in this process. Dynamic `import()`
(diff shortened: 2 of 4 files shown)


griffinwork40 (commented):

## Merge decision: MERGE

**Target:** PR #324 · `feat(plugins): inject host runtime API into plugin entrypoints (fixes singleton trap)` · base `main` ← head `01712b0b` · feature · 4 files (`load-entrypoints.ts`, `skill-bridge.ts`, `index.ts` prod/export + `load-entrypoints.test.ts`) · 176 diff lines (~50 effective production) · light regime · 1 review agent + independent citation/absence verification.

**Stated intent (supplied — spec-compliance assessed):** add host-API injection to the plugin `main` mechanism so a code-backed, out-of-tree plugin registers skills that actually reach the host's registry. If a plugin's `main` exports a **callable default**, the host invokes `await mod.default(pluginApi)` after import, passing a `PluginApi` assembled from the host's OWN modules `{ registerSkill, listSkills, getSkill, loadSkillPrompts }`. Modules with no default export still run top-level side-effects at import (back-compat). Fixes the singleton trap: the skill registry is a module-level singleton, so a plugin resolving its own copy of `agent-afk` (own `node_modules`, a bundle, or a marketplace-cloned install with no shared `node_modules`) registers into a different registry than the host reads — a silent no-op. `PluginApi` is typed via `typeof import(...)`, exported from `src/index.ts`, assembled in skill-bridge's `ensurePluginEntrypointsLoaded`; stateless helpers intentionally omitted.

The change is correct, spec-compliant, and minimal. CI on this ref is fully green (Lint & Build, Test, Docs, Bundle Smoke all SUCCESS). **No blocking findings** — two `low` test-coverage gaps for already-correct behavior, neither gating merge.

## Findings (sorted by severity; none blocking)

**[low · high · test-coverage · `src/agent/plugins/load-entrypoints.test.ts` · ref:01712b0b · file-state]** — No test exercises a **successfully-imported plugin whose default-export *function* throws**. This PR adds a new path inside the existing try/catch — `await entry(opts.pluginApi)` (`load-entrypoints.ts:98`) — that a throwing default traverses. The pre-existing non-fatal test (`load-entrypoints.test.ts:69`) covers only a *failed import* (`main: 'does-not-exist.mjs'`), not a throwing default-export, so the new code path's routing to `onError` is structurally correct but unpinned.
Evidence:
```ts
// load-entrypoints.test.ts:69 — exercises a MISSING file, not a throwing default:
{ type: 'local', path: dir, main: 'does-not-exist.mjs' }
```
*Suggestion:* add a test writing `export default () => { throw new Error('boom'); }`, assert `loadPluginEntrypoints(...)` resolves `undefined` and `onError` received exactly 1 error.

**[low · high · test-coverage · `src/agent/plugins/load-entrypoints.test.ts` · ref:01712b0b · file-state]** — No test covers a default-export entrypoint invoked with `pluginApi` **omitted** (`undefined`). The signature is `pluginApi?` and the invocation passes `opts.pluginApi` straight through, so a default-export receives `undefined` when no caller supplies it; the production call site (`skill-bridge.ts:272`) always supplies the full object, but the optional-path contract is unproven.
Evidence:
```ts
await (entry as (api: PluginApi | undefined) => unknown)(opts.pluginApi);
```
*Suggestion:* add a test calling `loadPluginEntrypoints([{…default-export…}])` with no `pluginApi` and assert the default-export still runs (receives `undefined`) without throwing. Non-blocking — the production path always injects.

## Per-dimension
- **security** — `no issues found — read src/agent/plugins/load-entrypoints.ts at 01712b0b`. No new network I/O, credential handling, or exec path. The `entry(api)` call invokes an already-imported module function; it adds no attack surface beyond what `import()` itself already exposes pre-PR.
- **api-compat** — `no issues found — read src/agent/tools/skill-bridge.ts, src/index.ts at 01712b0b`. Every change is additive: `loadPluginEntrypoints` already took `opts: LoadEntrypointsOptions = {}`, gaining a new optional `pluginApi?` field breaks no existing one- or zero-arg caller (grep: exactly 2 production sites — the definition and the single skill-bridge call); `PluginApi` and its `src/index.ts` re-export are net-new **type-only** exports; the new `registerSkill` import in skill-bridge is additive. No exported symbol removed or renamed.
- **correctness** — `no issues found`. The new `const mod = await importer(...)` and `await entry(opts.pluginApi)` both sit **inside** the existing `try { … } catch (error) { opts.onError?.(plugin, error); }` (`load-entrypoints.ts:90–103`), so a throwing plugin default routes to `onError` exactly like a failed import — the back-compat must-not-crash invariant holds. `loadedEntrypoints.add(absPath)` still fires before the try block, so a throwing default is not retried. The `pluginApi?` optionality passing `undefined` to a default-export is by design (the sole production caller always supplies it) — not a defect.
- **spec-compliance** — **COMPLIANT.** Every named clause verified at the reviewed ref: host injects EXACTLY `{ registerSkill, listSkills, getSkill, loadSkillPrompts }` (`skill-bridge.ts:272`, matches the intent's stated set with no missing/extra key); stateless helpers (`env`, paths, `describeFailure`) omitted; `PluginApi` derived via `typeof import('../../skills/index.js')` + `prompt-loader.js` (`load-entrypoints.ts:37–41`); re-exported from `src/index.ts:84`; default invoked only when `typeof entry === 'function'` so no-default modules still run side-effects at import (back-compat); both the round-trip test (default-export → injected `registerSkill` → host `getSkill`/`listSkills` against the REAL registry via `_resetRegistry`) and the back-compat side-effect test are present. No scope creep.
- **test-coverage** — the two `low`s above (throwing default-export; omitted-`pluginApi` path). The two new tests are well-targeted: the round-trip test proves injection reaches the host singleton, the side-effect test proves no-default back-compat.
- **perf-observability** — `no issues found`. Added work per entrypoint is one `typeof` check and one conditional `await` — O(1), no loop/retry/allocation on any hot path (entrypoint load is a cold boot-once event).

## Verified clean (independently re-derived at ref 01712b0b)
- **Try/catch envelope confirmed** — read `load-entrypoints.ts:90–103` at the reviewed ref; the default-export invocation is inside the catch that routes to `onError`. Back-compat non-crash invariant holds for a throwing default.
- **Injection object matches the spec exactly** — read `skill-bridge.ts:271–273`; the runtime object is `{ registerSkill, listSkills, getSkill, loadSkillPrompts }`, the precise four-key set named in the intent.
- **PluginApi dependencies all exist at the ref** — `registerSkill` (`skills/index.ts:164`), `getSkill` (`:172`), `listSkills` (`:186`), `loadSkillPrompts` (`_lib/prompt-loader.ts:23`), and the test-only `_resetRegistry` (`skills/index.ts:209`) are all exported. The `Pick<typeof import(...)>` type and the test imports resolve.

## What was not checked
- The single review agent and the orchestrator's backstop read against branch HEAD `01712b0b`; every `file:line` citation was verified at that ref via `git show 01712b0b:<path>` on a locally-fetched copy of the exact reviewed commit.
- Tests were **not** executed (read-only review). CI on this ref is green (Lint & Build, Test, Docs, Bundle Smoke all SUCCESS); the PR's "1552 pass / 2 skip" sweep and the "`PluginApi` present in regenerated `dist/index.d.ts`" claim were not independently re-run.
- No `critical`/`high` finding surfaced, so `/shadow-verify` was not triggered (it gates on top-severity only).
- Not exercised live: a real marketplace-cloned plugin with its own `node_modules` registering through the injected API (the scenario the fix targets) — proven structurally and by the round-trip unit test, not by an end-to-end out-of-tree install.

**Bottom line:** a correct, spec-compliant, minimal feature with green CI and no blocking issues. The injection lands the plugin's registration in the host's singleton registry regardless of install layout, the back-compat side-effect path and the throwing-default routing are both preserved, and the type-derived `PluginApi` makes the injected surface self-validating against drift. The two `low` items are test-coverage polish for already-correct paths. Safe to merge.

*Read-only automated review — no files, commits, branches, or PR state were modified.*

PR #325, opened 2026-06-28
fix(repl): honor trailing-backslash newline escape on the live TTY path

griffinwork40:
## Problem
In the interactive REPL, typing a trailing `\` then **Enter** submitted the raw backslash instead of inserting a newline (continuation). shift+Enter / alt+Enter worked, but the documented `\`-escape — the fallback for terminals that do not report shift-state on Enter — did not.

## Root cause
The live armed-TTY path routes Enter through `TerminalCompositor.dispatchKey()` → `handleEnter` (`src/cli/terminal-compositor.input-dispatch.ts`). The trailing-`\`→newline branch lived **only** in the legacy `readWithAutocompleteTty` reader (`src/cli/input/reader.ts`), which is reachable only as a non-armed-TTY / non-TTY fallback. The two readers are near-clones of the same keypress decision tree with no shared code, so they drifted. The same drift had left `reader.ts` carrying a stale `length === 1` printable test that silently dropped multi-UTF-16-unit emoji on the fallback path.

## Fix
Extract the drift-prone decisions into pure, shared, unit-tested helpers so the readers cannot diverge again:
- `src/cli/input/enter-decision.ts` — `isSoftNewlineEnter`, `endsWithBackslashContinuation`
- `src/cli/input/printable.ts` — `isPrintableGrapheme` (moved out of `input-dispatch.ts`)

Wired into all three readers:
- **compositor** — gains the missing trailing-`\` branch (the live-bug fix) + emoji-safe printable guard
- **reader.ts** — uses the shared helpers; printable guard now admits emoji
- **multi-line-reader.ts** — uses `endsWithBackslashContinuation` for consistency

## Tests
- `terminal-compositor.test.ts` — end-to-end regression: `foo\` + Enter → `{ text: 'foo\n', queued: false }` (drives the bug through `dispatchKey`; would have caught the original).
- `input/enter-decision.test.ts` (9) + `input/printable.test.ts` (5) — pure-predicate coverage.

## Verification
- `pnpm lint` (strict `tsc --noEmit`): green.
- Full `pnpm test`: 10298 passed / 14 skipped; the only 2 failures are pre-existing and unrelated (`config.test.ts` MLX slot bindings, `anthropic-direct.test.ts` compact model-id) — baseline before this change was 3 failed / 10282 passed, so no regressions.

## Scope / follow-ups (not in this PR)
Discovered via a `/simplify` pass on the input subsystem. Deferred: retire-or-guard `reader.ts`'s dead-on-armed-TTY keypress loop, Home/End line-relative parity in `reader.ts`, splitting the 183-line nav+edit handler, and routing `selectors.ts` through `enterRawMode`. Dropdown up/down direction left as a deliberate wrong-abstraction defer.

### DIFF
--- a/src/cli/input/enter-decision.test.ts
+++ b/src/cli/input/enter-decision.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, it } from 'vitest';
+import { isSoftNewlineEnter, endsWithBackslashContinuation } from './enter-decision.js';
+
+describe('isSoftNewlineEnter', () => {
+  it('is true for shift+Enter (key.shift)', () => {
+    expect(isSoftNewlineEnter({ shift: true }, undefined)).toBe(true);
+  });
+
+  it('is true for alt/option+Enter (key.meta)', () => {
+    expect(isSoftNewlineEnter({ meta: true }, undefined)).toBe(true);
+  });
+
+  it('is true for the kitty shift+Enter sequence \\x1b[13;2u even without key.shift', () => {
+    expect(isSoftNewlineEnter(undefined, '\x1b[13;2u')).toBe(true);
+    expect(isSoftNewlineEnter({}, '\x1b[13;2u')).toBe(true);
+  });
+
+  it('is false for plain Enter (no modifiers, no special sequence)', () => {
+    expect(isSoftNewlineEnter({}, undefined)).toBe(false);
+    expect(isSoftNewlineEnter({ shift: false, meta: false }, '\r')).toBe(false);
+  });
+
+  it('is false for an undefined key with no kitty sequence', () => {
+    expect(isSoftNewlineEnter(undefined, undefined)).toBe(false);
+    expect(isSoftNewlineEnter(undefined, '\r')).toBe(false);
+  });
+});
+
+describe('endsWithBackslashContinuation', () => {
+  it('is true when the buffer ends in a backslash', () => {
+    expect(endsWithBackslashContinuation('foo\\')).toBe(true);
+    expect(endsWithBackslashContinuation('\\')).toBe(true);
+  });
+
+  it('is true when the buffer ends in two backslashes (still ends with a backslash)', () => {
+    // Matches the original reader.ts semantics: a simple trailing-char test,
+    // not an even/odd escape count.
+    expect(endsWithBackslashContinuation('foo\\\\')).toBe(true);
+  });
+
+  it('is false when the buffer does not end in a backslash', () => {
+    expect(endsWithBackslashContinuation('foo')).toBe(false);
+    expect(endsWithBackslashContinuation('a\\b')).toBe(false);
+  });
+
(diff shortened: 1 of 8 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

Clean, well-scoped hotfix. It fixes the documented `\`+Enter soft-newline escape on the live armed-TTY path (which previously submitted the raw backslash), by extracting the drift-prone Enter/printable decisions into shared, pure, unit-tested helpers (`enter-decision.ts`, `printable.ts`) and wiring them into all three input readers. Behavioral parity with the reference `reader.ts` implementation was verified; CI is fully green.

### What was verified (read against branch HEAD `9779badb`)
- **Placement parity (compositor vs reader.ts):** The new trailing-`\` branch in `handleEnter` sits after the dropdown block, vs an `else if` in `reader.ts`. Both resolve identically — a slash completion never ends in `\`, and the attachment-only-submit ordering is correct (a lone `\`+Enter is a continuation intent, not a submit). No defect.
- **`isPrintableGrapheme('')`:** `'' >= ' '` is `false` in JS, so the `&&` short-circuits before `nextGraphemeIndex` — empty string correctly returns `false`. The `s >= ' '` guard is load-bearing.
- **`noModifier` refactor:** `!key?.ctrl && !key?.meta` computed once and applied to both the `char` and `key.sequence` branches — semantically identical to the old duplicated checks. No modifier check dropped or added.
- **`endsWithBackslashContinuation`:** Byte-exact `buffer.endsWith('\\')` — identical to the prior inline `reader.ts` test, including the intentional `foo\\` (two-backslash) → `true` behavior. No regression.
- **Soft-newline + backslash interaction:** `isSoftNewlineEnter` fires first in both paths, so a `\`-terminated buffer + shift+Enter inserts `\n` and leaves the `\` literal — parity confirmed.

### Spec-compliance
All three stated requirements implemented (compositor gains the `\` branch; `isPrintableGrapheme` extracted + wired; `multi-line-reader.ts` uses the shared helper). No scope creep.

### Findings
**Low · test-coverage** — `src/cli/terminal-compositor.test.ts` (ref `9779badb`): No test covers the lone-backslash buffer (`buffer === '\'`) + plain Enter path in the compositor. The regression test only exercises `'foo\'`; the single-char `replaceRange({ start: 0, end: 1 }, '\n')` boundary is untested.
  - *Suggestion:* add a case — type only `\`, press plain Enter, assert `getBuffer()` equals `{ text: '\n', queued: false }`. Non-blocking.

### What was not checked
- Did not run the test suite locally (review is read-only); relied on the PR's stated CI result + green checks on GitHub.
- Did not review the Telegram or daemon surfaces or downstream consumers outside `src/cli/`.
- `nextGraphemeIndex` internal correctness assumed (pre-existing, unchanged).

---
*Read-only automated analysis. The only action taken was posting this comment; no approval, change-request, merge, or branch modification was performed.*

PR #329, opened 2026-06-29
refactor(providers): extract shared helpers into providers/shared + fix env-bleed test

griffinwork40:
## Summary

Extracts verbatim-duplicated and cross-boundary-imported helpers from both providers into a new `src/agent/providers/shared/` directory, eliminating the duplication the code comments already called out.

### (A) Fix env-bleed test flakiness
`src/cli/config.test.ts` model-slots `beforeEach`/`afterEach` now clears `AFK_MODEL_LOCAL`, `AFK_MODEL_LOCAL_BASE_URL`, and `AFK_MODEL_LOCAL_API_KEY` in addition to the existing vars — preventing developer shells with those set from bleeding into `DEFAULT_SLOT_BINDINGS` assertions.

### (B) Extract 3 verbatim-duplicated pure functions
New modules in `src/agent/providers/shared/`:
- `sleep-with-abort.ts` — `sleepWithAbort` (was in both `anthropic-direct/loop.ts` and `openai-compatible/query.ts`)
- `tool-input-summary.ts` — `summarizeToolInput` (same files; the comment in openai already said "mirrors anthropic-direct helper exactly")
- `supported-commands.ts` — `collectSupportedCommands` extracted from `supportedCommands()` class method (both providers; the comment in openai already said "Extract to a shared helper module when a third provider lands")

Both inline copies deleted; imports updated.

### (C) Relocate mislocated shared helpers
Three files that `openai-compatible/query.ts` was already importing cross-boundary from `anthropic-direct/` moved to `shared/`:
- `shared/auto-compact.ts` (was `anthropic-direct/query/auto-compact.ts`)
- `shared/plan-mode-addendum.ts` (was `anthropic-direct/plan-mode-addendum.ts`)
- `shared/afk-mode-addendum.ts` (was `anthropic-direct/afk-mode-addendum.ts`)

Original files become thin re-export shims so all existing test imports resolve unchanged. `openai-compatible/query.ts` cross-boundary imports updated to `../shared/*`.

### (D) Deferred
Haiku model-ID divergence (`query.ts:91` `claude-haiku-4-5-20250929` vs `compact-handler.ts:54` `claude-haiku-4-5-20251001`) is pre-existing and masked by `AFK_COMPACT_MODEL` env bleed in the test runner (present on main before this branch). Deferred to avoid false signal.

## Testing

```
pnpm lint          # tsc --noEmit — CLEAN
pnpm exec vitest run src/agent/providers src/cli/config.test.ts --no-cache
# 1 failed (pre-existing: AFK_COMPACT_MODEL env bleed in anthropic-direct.test.ts,
#           reproduced on main/unmodified — not introduced by this branch)
# 789 passed (+1 vs main's 788 passed — the env-bleed config.test fix landed)
```

Part of a top-5 improvements batch.

### DIFF
--- a/src/agent/providers/anthropic-direct/afk-mode-addendum.ts
+++ b/src/agent/providers/anthropic-direct/afk-mode-addendum.ts
@@ -1,59 +1,16 @@
 /**
- * AFK-mode system-prompt addendum.
+ * Re-export shim — the canonical implementation has moved to
+ * `../shared/afk-mode-addendum.ts`.
  *
- * When the session is in `'autonomous'` permission mode (AFK mode), this module
- * supplies a single text block that the provider's `composeSystem()` appends to
- * the system payload. The addendum is the *posture* half of AFK mode — its
- * companion is the hook-layer refusal in {@link module:agent/afk-mode-gate},
- * which is the *enforcement* half (the mechanical safety ceiling).
- *
- * Design notes:
- *  - The posture is **bounded autonomy, not YOLO.** It tells the model to act
- *    on reversible work without waiting, but to STOP at one-way doors
- *    (irreversible / external / genuinely ambiguous forks) and surface an
- *    Asking summary to Telegram instead of guessing. The text explicitly defers
- *    to the gate: the model is told a mechanical layer refuses high-risk ops
- *    regardless of what it decides, so it should not treat the posture as a
- *    licence to bypass safety.
- *  - The channel to the operator is Telegram via the `send_telegram` tool —
- *    `send_telegram` is the one outbound op the gate never blocks.
- *  - The block carries no `cache_control` of its own — `withSystemBreakpoint`
- *    in `cache-policy.ts` floats the breakpoint to whichever block is last, so
- *    toggling AFK mode busts the cache once (correct) and same-mode turns hit
- *    cleanly. This mirrors the plan-mode addendum's cache behaviour exactly.
- *  - AFK mode and plan mode are mutually exclusive permission modes, so at most
- *    one of the two addenda is ever appended in a turn.
+ * This file is kept so existing test imports
+ * (`./afk-mode-addendum.js` from `afk-mode-addendum.test.ts` and
+ *  `plan-mode-system-payload.test.ts`) continue to resolve without
+ * modification.
  *
  * @module agent/providers/anthropic-direct/afk-mode-addendum
  */
 
-import type { ContentBlockParam } from '@anthropic-ai/sdk/resources';
-
-export const AFK_MODE_ADDENDUM_TEXT = [
-  '## AFK mode is active',
-  '',
-  'The operator is away from keyboard. Your channel to them is Telegram via the `send_telegram` tool — not this transcript, which no one is watching live. At the end of each turn, push your terminal state (Done / Blocked / Asking) to Telegram so the operator can review asynchronously.',
-  '',
-  'Posture — bounded autonomy, not unchecked action:',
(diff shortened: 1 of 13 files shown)


griffinwork40 (commented):

**Merge decision: MERGE**

Reviewed ref `12d185da` (head) against base `a93ee9a4`. This is a behavior-preserving refactor: it extracts verbatim-duplicated helpers and relocates cross-boundary-imported helpers into `src/agent/providers/shared/`, with thin re-export shims at the original paths so existing imports resolve unchanged, plus an env-bleed fix to the model-slots test.

**Findings**

- **low** · test-isolation · `src/cli/config.test.ts:526-541` — The model-slots `beforeEach`/`afterEach` *delete* the env keys (`for (const k of ENV) delete process.env[k]`) rather than snapshot-and-restore, so running the full suite in a shell with `AFK_MODEL_LOCAL*` set strips those vars from `process.env` for the rest of the process. This is **not a regression** — it extends the pre-existing pattern already applied to `AFK_MODEL`/`CLAUDE_MODEL`/`AFK_MODEL_{SMALL,MEDIUM,LARGE}` — and it correctly fixes the documented flake. Optional: snapshot each key's prior value in `beforeEach` and restore in `afterEach` if any later test depends on a developer-set value.

- **nit** · correctness · `src/agent/providers/shared/supported-commands.ts:31-48` — `collectSupportedCommands` turns the original `async` method into a non-async function returning `Promise.resolve(...)` (and `Promise.resolve([])` on error). Behavior is equivalent for awaiting callers, and `collectSkillEntries()` is still called synchronously inside the `try`, so synchronous throws are still caught. No action required — noted only because it is the one extraction that changed control-flow *shape* rather than moving a body verbatim.

**Verified clean (read at ref 12d185da, compared against base a93ee9a4)**

- `sleepWithAbort` — byte-identical, base → `shared/sleep-with-abort.ts`.
- `summarizeToolInput` — byte-identical from **both** `anthropic-direct/loop.ts` and `openai-compatible/query.ts` → `shared/tool-input-summary.ts`.
- `auto-compact.ts` — full module body identical (only the relative import-path depth changed, correct for the new location); the shim re-exports all four value exports **and** both type exports (`export type { … }` present, so type-only importers still resolve).
- `afk-mode-addendum.ts` / `plan-mode-addendum.ts` — text constants + builder functions identical; shims re-export the matching names.
- **api-compat** — all three shims re-export the exact canonical symbol names; every test named in the shim JSDocs exists at the ref and imports via the old (shimmed) paths, so no production or test importer breaks.
- **test-coverage** — moved code keeps coverage through the shim imports (`auto-compact.test.ts`, `afk-mode-addendum.test.ts`, `plan-mode-addendum.test.ts`, `plan-mode-system-payload.test.ts`, `openai-compatible/query.test.ts` all resolve unchanged). No new tests warranted for a verbatim move.
- **security** — pure internal code-movement; no auth/injection/secret surface. `AFK_MODEL_LOCAL_API_KEY` appears only as a key being deleted in test setup, not logged or leaked.

**What was not checked**

Read against head `12d185da` and base `a93ee9a4` via read-only content reads — all blocking-tier claims verified at those refs. The deferred Haiku model-ID divergence (item D in the description) was confirmed out of scope and not assessed. The test suite was not run locally; CI on this PR is green (Lint & Build, Test, Docs, Publish Bundle Smoke all SUCCESS). Stated intent (PR title + body) was available and spec-compliance was assessed — the "verbatim move" claim holds. No critical/high findings.

PR #332, opened 2026-06-29
fix(tui): refill the viewport after end-of-turn collapse of an over-tall block

griffinwork40:
## What this fixes

The operator-reported bug *"stuff sometimes disappears from scrollback / doesn't render properly"*: when the final assistant message is **taller than the (collapsed) viewport**, the end-of-turn overlay collapse left a large **blank void above the prompt** — the content was only reachable by scrolling up into native scrollback. Intermittent, because it only fires for tall final messages. `#645` deliberately left this overflow path unhandled.

## Verified mechanism (corrected from the initial hypothesis)

Established by per-step `@xterm/headless` instrumentation — **not** a stale-tall Phase-2 erase (`commitAbove` calls `logUpdate.clear()` first, so that erase is a no-op):

1. The block is committed while a tall overlay fills the viewport → `prevTopRow <= 1` → `fitsAboveFrame = false`.
2. The block is taller than even the collapsed screen → `decideCommitMode` returned `useBandHold = false` → Phase 1 archived the whole block to scrollback.
3. Phase 3 is guarded by `if (newTopRow > 1)`; with the overlay filling the screen `newTopRow === 1` → Phase 3 **skipped** → `committedBand` left **empty**.
4. End-of-turn collapse (`setOverlay('')`) erased the overlay rows, but `committedBand` was empty so `repositionCommittedBand` had nothing to re-pin → the freed viewport rows stayed blank.

So the defect was **"viewport not refilled with recent committed content after collapse"**.

## The fix (two parts)

1. **`commit-mode.ts` — route the over-tall case through band-hold:** `useBandHold = overflowHasPending || (!fitsAboveFrame && maxBandModel > 0)`. The block is held as a pending capped model and the genuine overflow is archived to scrollback, instead of the legacy archive that left `committedBand` empty.

2. **`terminal-compositor.frame-preserve.ts` — evict the pending overflow at collapse:** Part 1 alone regresses the multi-commit case — band-hold's **commit-time** `maxBandModel` can exceed the **true collapse paint capacity** (`repositionCommittedBand`'s `maxFit`) once the real collapsed-frame height (input + rhythm separator + loop-stage bar + status) is counted, silently **dropping** the unpaintable pending rows. The fix evicts that excess to scrollback *before* the collapse render, gated on the **genuine end-of-turn signal — the overlay being empty** (`self.overlay.trim() === ''`), **not** a room-magnitude threshold (loses content on tall collapsed frames) nor a shrink-direction heuristic (prematurely archives rows that should stay visible). `room` is then the true above-frame capacity for *any* collapsed-frame height, so the eviction count is exactly the overflow that cannot be shown.

## Why the trigger is principled (not a magic number)

An earlier draft used `room >= maxAboveFrameRoom - 2` — a stress sweep showed that **silently drops 3–6 rows whenever the collapsed frame is taller than ~2 rows** (residual input, spinner+gap, footer). Gating on "overlay empty" (the real turn-ended signal) is geometry-independent and fires exactly once, at the real collapse.

## Verification
- `endturn-overflow-gap.repro.test.ts` — flipped `it.fails` → `it()`; the gap is closed.
- `band-hold-perline-gap.repro.test.ts` — content-loss guard, **green** (every row lands in scrollback+viewport).
- `overflow-gap.test.ts` "archives the genuine overflow … R10 visible" — **green** (no premature eviction).
- Full compositor + `commit-mode` + renderer suite: **363 passed**. `pnpm lint` clean. Full suite: only the 2 pre-existing unrelated env-fails (`anthropic-direct` compact, `config` MLX-bleed).
- `@xterm/headless` stress across collapsed-frame heights — block 25/30/40 × extraRows 1/2/3 × spinner-at-settle × multi-step collapse (36 configs) **and** the banner case (anchorRow>1, 6 configs): **zero content loss, zero premature archival**.

## Test updates (behavior legitimately changed by the new routing; assertions not weakened)
- `commit-mode.test.ts`: over-tall block now asserts `useBandHold=true` (was `false`).
- `terminal-compositor.test.ts` ×2: the over-tall exact-fit and banner cases now route through band-hold — assertions updated to "no row dropped" + the new band-hold positions; the **banner-clobber invariant (no CUP write into the banner zone) is preserved**.

---

## Follow-up — banner-path completeness (commit `c8ca995`)

A post-push review found the part-2 collapse-time pending-overflow eviction lived inside the `if (!hasBanner)` branch of `preserveRowsBeforeFrameRender`, leaving the **welcome-banner window** (`anchorRow > 1`) exposed to the same content-loss class — via the **partially-pending** band case (`0 < committedBandPaintedRows < bandLen`), which *can* coexist with a live banner. (The *fully*-pending case forces `anchorRow → 1` during Phase-2, so it never reaches the banner branch — the original invariant comment held only for that sub-case.)

**Reachable when** the commit overlay makes `prevTopRow ≤ 1` (so `useBandHold` fires) but `desiredTopRow > anchorRow` at Phase-2 (banner survives unreduced): Phase-3 paints only `maxRun = newTopRow − anchorFloor` rows, leaving the oldest rows pending; on end-of-turn collapse the banner branch runs only deficit-based eviction (deficit = 0 when the frame grows downward), so `repositionCommittedBand` silently drops the oldest `bandLen − maxFit` pending rows.

**Fix:** mirror the `!hasBanner` collapse-eviction in the banner branch with `floor = anchorFloor`, evicting the pending overflow to scrollback before the deficit path / `repositionCommittedBand` runs.

**Guard:** `terminal-compositor.banner-endturn-overflow.repro.test.ts` fails pre-fix (`BANNERBLOCK-01`/`-02` dropped, found 0) and passes post-fix. Compositor + commit-mode suites: **338 passed**; `pnpm lint` clean.

### DIFF
--- a/docs/scrollback.md
+++ b/docs/scrollback.md
@@ -525,10 +525,10 @@ commit satisfies `overflowPriorContiguous` and merges contiguously, and both
 blocks survive in commit order. Verified by the third case in
 `terminal-compositor.h1-prevtoprow.test.ts`.
 
-### Known bug (NOT yet fixed): block taller than the COLLAPSED screen blanks the viewport on end-of-turn collapse
+### Fixed: block taller than the COLLAPSED screen blanked the viewport on end-of-turn collapse
 
-Repro: `terminal-compositor.endturn-overflow-gap.repro.test.ts` (marked
-`it.fails`). Operator symptom: after a long turn, the final assistant message
+Repro / regression guard: `terminal-compositor.endturn-overflow-gap.repro.test.ts`.
+Operator symptom (pre-fix): after a long turn, the final assistant message
 "disappears" — the viewport is blank above the prompt and the content is only
 reachable by scrolling up into native scrollback.
 
@@ -556,19 +556,40 @@ So the defect is **"viewport not refilled with recent committed content after
 collapse"**, not an erase wiping a painted band. "No existing test hits
 prevTopRow <= 1" (`committed-band-commit.ts`); the repro above is that test.
 
-Why the obvious fix is insufficient (the blocker): routing this case through
-band-hold (so the pending model + `repositionCommittedBand` refill the viewport
-on collapse) fixes the SINGLE-commit case, but regresses the multi-commit case.
-Band-hold's `maxBandModel` is a COMMIT-TIME estimate of the collapsed paint
-capacity; the actual collapse capacity (`repositionCommittedBand`'s `maxFit`) is
-smaller once the collapsed frame's real height (input + rhythm separator +
-loop-stage bar + status line) is accounted for. When `maxBandModel > maxFit`, the
-oldest model rows are neither painted NOR archived to scrollback — silent content
-loss, caught by `terminal-compositor.band-hold-perline-gap.repro.test.ts`
-("a block taller than the collapsed screen still lands every row contiguously").
-A correct fix must reconcile `maxBandModel` with the true collapse paint capacity
-and archive the remainder to scrollback (so the excess is recoverable, never
-dropped) — a change to the deliberately-deferred overflow design, not a one-liner.
+The fix (two parts):
+
+1. **Route the over-tall case through band-hold** (`commit-mode.ts`):
+   `useBandHold = overflowHasPending || (!fitsAboveFrame && maxBandModel > 0)`.
+   The block is no longer dropped down the legacy overflow archive — Phase 1
+   archives the genuine overflow (oldest rows beyond `maxBandModel`) to scrollback
+   as REAL content, and the pending capped model is stored so a collapse re-pin
+   can materialize it.
+
+2. **Evict the pending overflow at collapse** (`preserveRowsBeforeFrameRender`,
+   `terminal-compositor.frame-preserve.ts`). Part 1 alone would regress the
+   multi-commit case: band-hold's COMMIT-TIME `maxBandModel` can exceed the true
+   collapse paint capacity (`repositionCommittedBand`'s `maxFit`) once the real
+   collapsed-frame height (input + rhythm separator + loop-stage bar + status) is
+   counted, so `repositionCommittedBand` paints only `fit` rows and the oldest
+   `bandLen - fit` PENDING rows would be neither painted nor archived — silent
+   content loss. The fix evicts that excess to scrollback BEFORE the collapse
+   render, gated on the **genuine end-of-turn signal — the overlay being empty**
+   (`self.overlay.trim() === ''`), NOT a room-magnitude threshold or a
+   shrink-direction heuristic. `room` (= `desiredTopRow - 1`) is then the TRUE
+   above-frame capacity for whatever the collapsed-frame height is, so the
+   eviction count (`bandLen - room`) is exactly the overflow that cannot be shown
+   — correct for ANY footer/input geometry. A mid-turn minor shrink (e.g. the
+   spinner stopping while a tall overlay is still held) leaves the overlay
+   non-empty, so the pending band is preserved intact for the real collapse and
+   rows that should stay visible are never prematurely archived.
+
+Guards: `terminal-compositor.endturn-overflow-gap.repro.test.ts` (the gap is
+closed), `terminal-compositor.band-hold-perline-gap.repro.test.ts` ("a block
+taller than the collapsed screen still lands every row contiguously" — the
+content-loss guard), and `terminal-compositor.overflow-gap.test.ts` ("archives
+the genuine overflow ... R10 visible" — no premature eviction). Verified against
+a real `@xterm/headless` buffer across varied collapsed-frame heights (extraRows,
+spinner-at-settle, multi-step collapse) — no content loss, no premature archival.
 
 ## What is and isn't in scrollback after a commit
 
--- a/docs/scrollback.md
+++ b/docs/scrollback.md
@@ -525,6 +525,51 @@ commit satisfies `overflowPriorContiguous` and merges contiguously, and both
 blocks survive in commit order. Verified by the third case in
 `terminal-compositor.h1-prevtoprow.test.ts`.
 
+### Known bug (NOT yet fixed): block taller than the COLLAPSED screen blanks the viewport on end-of-turn collapse
+
+Repro: `terminal-compositor.endturn-overflow-gap.repro.test.ts` (marked
+`it.fails`). Operator symptom: after a long turn, the final assistant message
+"disappears" — the viewport is blank above the prompt and the content is only
+reachable by scrolling up into native scrollback.
+
+Mechanism (empirically verified by per-step `@xterm/headless` instrumentation —
+NOT a stale-tall Phase-2 erase; `commitAbove` calls `logUpdate.clear()` first, so
+that erase is a no-op):
+
+1. The block is committed while a tall overlay fills the viewport, so
+   `prevTopRow <= 1` (BLOCKER-1, review #592) and `fitsAboveFrame` is false.
+2. The block is taller than even the COLLAPSED screen, so `decideCommitMode`
+   returns `useBandHold = false` (the `overflowRun.length > maxBandModel &&
+   textLines.length > maxBandModel` case — the one this doc's "the overflow path
+   is unchanged" notes #645 deliberately left alone). Phase 1 archives the whole
+   block to native scrollback.
+3. Phase 3 is guarded by `if (newTopRow > 1)` (`committed-band-commit.ts`).
+   With the overlay still filling the screen `newTopRow === 1`, so Phase 3 is
+   SKIPPED and `committedBand` is left EMPTY.
+4. At end-of-turn the overlay collapses (`setOverlay('')` → `bootstrap.ts`;
+   `loopStageBar.repaint('observing')` → `loop-iteration.ts`). `render()` erases
+   the overlay rows, but `committedBand` is empty so `repositionCommittedBand`
+   has nothing to re-pin — the freed viewport rows stay blank. The content sits
+   in scrollback ABOVE the viewport, unreachable without scrolling.
+
+So the defect is **"viewport not refilled with recent committed content after
+collapse"**, not an erase wiping a painted band. "No existing test hits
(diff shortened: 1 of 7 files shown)


griffinwork40 (commented):

**Merge decision: DO NOT MERGE** (as-is — one HIGH finding to fix first)

This is a clean characterization PR (failing `it.fails` repro + docs section, no source change, CI green). Spec-compliance, test quality, docs accuracy, and security/api-compat/perf all pass. One blocking issue: an inline comment records the *wrong* root cause, which defeats the PR's stated purpose of documenting the correct mechanism.

---

### HIGH · correctness · `src/cli/terminal-compositor.endturn-overflow-gap.repro.test.ts` (CORE ASSERTION block)

The CORE ASSERTION inline comment states the **wrong failure mechanism**, contradicting the corrected mechanism in both the file header and `docs/scrollback.md`.

The file header explicitly *corrects* the old hypothesis:
> "this CORRECTS an earlier hypothesis that a stale-tall Phase-2 erase wiped the band; `commitAbove` calls `logUpdate.clear()` first, so that erase is a no-op"

But the inline comment above the `BLOCKROW` loop still asserts the discredited mechanism:
```
// Pre-fix: the top ~5 lines (BLOCKROW-00..BLOCKROW-04) are MISSING
// because Phase 2's erase wiped the Phase-1 archive before Phase 3
// could repaint them, and scrollback got blank rows instead of content.
```

A reader using this comment to understand the bug learns the wrong cause — and for a characterization PR whose entire purpose is to record the correct root cause, that's a substantive accuracy defect, not cosmetic.

**Suggestion:** Rewrite the comment to match the corrected mechanism, e.g.: "Pre-fix: `committedBand` is left EMPTY (Phase 3 skipped because `newTopRow === 1`), so `repositionCommittedBand` has nothing to re-pin after the overlay collapses — the `BLOCKROW` lines sit in native scrollback above the viewport, and the viewport above the prompt stays blank."

### NIT · correctness · same comment

"top ~5 lines … MISSING" is also a residual artifact of the old hypothesis. Under the documented Phase-3-skip mechanism (empty band → nothing re-pinned), the blank region is the *entire* viewport above the prompt, not just 5 lines. Drop the "top ~5 lines" specificity.

---

### What was not checked
Reviewed from the PR **diff text only**, against head ref `6a1b2feead6822163009f5cf4b55ecd8a82ea3b9`. The two findings above are bound to verbatim diff content (independently re-verified). Claims that require reading files outside the diff are **UNVERIFIED**: the `it.fails` convention at `splice.test.ts:260`, the line citations in `commit-mode.ts:148-151` and `committed-band-commit.ts:428`/`:187`, and the actual runtime pass/fail behavior of the new test. Spec-compliance **was** assessed against PR #332 title + body (characterization-PR intent).

griffinwork40:
Diff-scope confirmed:
- **frame-preserve.ts hunks** (`@@ -47,12 +47,83 @@`, `@@ -69,13 +140,13 @@`) live entirely inside the `if (!hasBanner)` block; the `hasBanner` else-branch (~158-200) is **byte-unchanged** by this PR.
- The **`overflowHasPending ||` disjunct** — the only route to `useBandHold=true` under banner geometry where `maxBandModel=0` — is present in *both* the removed (line 165) and added (line 178) form. This PR changed only the **second** disjunct. So #332 introduces no new banner-mode pending-band reachability.

That closes the loop. No critical/high survives → shadow-verify not mandated (and Wave 1.5 already independently re-derived and refuted the one substantive claim). Synthesizing the merge decision.

---

# Code Review — PR #332

**Target:** `fix(tui): refill the viewport after end-of-turn collapse of an over-tall block` · branch `fix/tui-endturn-overflow-scrollback-gap` → `main`
**Reviewed ref:** `303e00b` (== current HEAD; all citations verified at this ref) · **regime:** full (2 Wave-1 agents + citation/absence verification)
**Stated intent:** PR #332 title + body — **spec-compliance assessed.**
**Scope reviewed:** production code (`commit-mode.ts`, `terminal-compositor.frame-preserve.ts`) + the 3 touched test files + docs. Excluded: untracked `testimonial.txt` (unrelated, pre-existing dirty state).

No blocking findings. The fix implements both stated parts, the in-scope (`!hasBanner`) row arithmetic is correct and well-guarded, the `CommitMode` contract is unchanged, and no regression is introduced. The one substantive content-loss concern raised in review was independently re-derived from source and **REFUTED** for this PR's scope. Remaining items are `low`/`nit` and non-blocking.

**`low` · correctness/scope · `frame-preserve.ts:49` · ref:303e00b · file-state — the over-tall fix covers only the `!hasBanner` path.**
The new collapse-eviction branch is gated by `if (!hasBanner)`; the `anchorRow > 1` (welcome-banner) collapse path is unchanged and has no pending-band collapse-eviction. **Re-derivation verdict: not reachable as content loss via this PR.** Under banner geometry with a viewport-filling overlay, `maxBandModel = max(0, overflowTargetBottom − anchorFloor) = 0` (`commit-mode.ts:154`), so the new second disjunct `(!fitsAboveFrame && maxBandModel > 0)` is false and cannot set a fully-pending band (`committed-band-commit.ts:629`). The `overflowHasPending` disjunct *can* set `useBandHold=true` under banner, but it is **byte-identical pre/post** (diff lines 165 vs 178) and the banner else-branch is **unchanged** by this PR — so any latent banner+pending-band+collapse edge is pre-existing, not introduced here. *Suggestion (follow-up, not this PR):* trace the partial-pending-band-under-banner + contiguous-merge + `newTopRow≤1` collapse path to terminus, or add a `banner-commit-gap` test that constructs `anchorRow>1` + pending band + overlay collapse.

**`low`/`nit` · docs · `terminal-compositor.endturn-overflow-gap.repro.test.ts:49` · ref:303e00b · file-state — stale comment contradicts live test.**
Block-comment line 49 still reads `Currently FAILS (it.fails): the viewport blanks above the prompt`, but the test at line ~102 is `it(` (the header at line 4 was updated; line 49 was not). *Evidence:* `:49` "Currently FAILS (it.fails)…" vs `:102` `it('preserves ALL lines…`. *Suggestion:* update line 49 to "Previously KNOWN-FAILING; fixed by this PR."

**`low` · test-coverage · `terminal-compositor.test.ts` (rewritten block, ~880-896) · ref:303e00b · diff-context — middle-row positional precision dropped; "assertions not weakened" is slightly inaccurate.**
The rewrite drops per-row CUP assertions for `LINE_02..LINE_20`, keeping only endpoint pins (`LINE_00` scrollback @ :896, `LINE_01` row 1 @ :893, `LINE_21` row 21 @ :891) plus `.toContain` no-loss checks. A pure reorder of the band's middle suffix would pass this unit test and is caught only by the `@xterm/headless` repro suites (`band-hold-perline-gap`, `endturn-overflow-gap`). The trade is legitimate (the routing change makes exact middle positions implementation-detail) and the repro suites are a stronger end-to-end guard — but the PR's "assertions not weakened" claim is imprecise for this file. *Suggestion:* add a one-line comment noting the intentional positional gap, or assert middle-row relative ordering.

**`nit` · security · `frame-preserve.ts:464-468` · ref:303e00b · file-state — silent `stdout.write` failure swallow.**
`try { self.stdout.write(out) } catch {}` mirrors the pre-existing pattern at `:138-141`; identical risk, not a regression. No security surface: all terminal sequences are built from internal compositor state, never external input.

## Verified clean
- **security:** no real attack surface (internal TUI rendering; no secrets/auth/injection/RCE).
- **api-compat:** `CommitMode` interface byte-unchanged; over-tall case correctly reaches the `else if (useBandHold)` full-viewport path (`committed-band-commit.ts:614`) instead of the legacy clear (:632); new branch gated by `commitInFlight` so Phase-2 repaints never trigger it.
- **correctness (in-scope `!hasBanner` path):** row arithmetic sound — `overflow=bandLen−room` with `room≥1, bandLen>room` guaranteed by the branch guards; paint loop bounded by `bandLen ≤ maxBandModel ≤ rows−2` (no off-screen write); `room>0` guard is safe (`room=0` only on a degenerate 1-row terminal where `maxBandModel=0` ⇒ `useBandHold=false`, so no pending band exists); new branch and renamed `grew` branch are mutually exclusive (new branch `return`s first). The PR's central claim — that without collapse-eviction `repositionCommittedBand` would paint only `fit=min(bandLen,maxFit)` rows and drop the prefix (`committed-band-repin.ts:86-95`) — is confirmed; the eviction reduces `committedBand.length` to `room` so post-render `fit=room` paints all survivors.
- **spec-compliance:** both stated parts implemented (`commit-mode.ts:178`; `frame-preserve.ts` collapse-eviction gated on `overlay.trim()===''`); the repro test genuinely exercises the fixed path; no scope creep (the `overflow→growRoom/growOverflow` rename is pure readability).
- **perf:** O(bandLen) CUP-writes per collapse (≤~22 rows), once per turn — negligible, matches the existing `grew` pattern.

## What was NOT checked
- Wave-1 agents read against `303e00b` (= HEAD); all `blocking`-eligible citations verified at that ref.
- **Tests were not run** — analysis is static reasoning over `@xterm/headless`-based repro tests; the PR reports 363 passing locally (unverified here).
- The **partial-pending-band-under-`hasBanner`** edge was not traced to terminus (flagged as the `low` follow-up above); it is pre-existing in code this PR does not modify, so it does not gate this merge.
- `@xterm/headless` version compatibility of the new repro test not checked.
- Stated intent was available (PR #332 title+body) and spec-compliance was assessed.

**Done** — Reviewed PR #332 read-only at ref `303e00b`; emitted MERGE with 4 non-blocking findings (1 low scope-note, 1 nit stale comment, 1 low test-coverage weakening, 1 nit error-swallow). The one substantive content-loss concern (banner-path) was independently re-derived via a citation+absence verification pass and refuted for this PR's scope, with diff-hunk evidence that the `hasBanner` branch and `overflowHasPending` disjunct are unchanged. No files were modified, no PR comment posted, no git/PR state mutated (read-only review per skill contract). Evidence: diff artifact `pr-332.diff`, reviewed ref `303e00b`. Recommended follow-up (separate from this PR): trace or add a test for the banner + pending-band + overlay-collapse path.

---

PR #335, opened 2026-06-29
feat(plugins): inject runtime values into PluginApi for code-backed plugins

griffinwork40:
## What

Widen `PluginApi` — the host runtime API injected into a code-backed plugin's default-export entrypoint — from 4 to 11 symbols, so an out-of-tree, code-backed plugin can reach the core runtime values it needs.

**Added (runtime values):** `env`, `SubagentManager`, `describeFailure`, `discoverPluginSkillBodies`, `getAgentFrameworkDir`, `getSkillsDir`, `getSessionsDir`.
**Already injected:** `registerSkill`, `listSkills`, `getSkill`, `loadSkillPrompts`.

## Why

A marketplace-cloned plugin has **no `node_modules` of its own**, so a bare `import … from 'agent-afk'` inside the plugin throws `ERR_MODULE_NOT_FOUND` at boot — verified empirically with a probe plugin:

```
agentAfkResolved: false
ERR_MODULE_NOT_FOUND: Cannot find package 'agent-afk' imported from <plugin>/index.mjs
```

The 4-symbol `PluginApi` from #324 fixed the **singleton-identity** half of the problem (the registry must be the host's instance). But it left every other runtime value unreachable — including otherwise-stateless helpers like `env` and the path getters — because in a clone the bare import doesn't resolve at all. A code-backed plugin that forks sub-agents (`SubagentManager`), formats failures (`describeFailure`), reads env/paths, or enumerates plugin skill bodies could not function.

## How

- `PluginApi` gains the seven runtime values, derived via `typeof import(...)` so injected signatures can't drift from their source modules.
- `skill-bridge.ensurePluginEntrypointsLoaded` assembles and injects the full set. The object literal is type-checked against `PluginApi`, so the assembly can't silently fall out of sync with the type.
- Type-only imports stay erased at build, so a plugin can still `import type { … } from 'agent-afk'` via a build-time devDependency.

Purely additive — no behavior change for existing side-effect-only plugins or the prior 4-symbol consumers.

## Tests

Extends the entrypoint round-trip test with a case proving the injected `SubagentManager` (the real class), `env`, `describeFailure`, `getAgentFrameworkDir`, and `discoverPluginSkillBodies` all arrive intact in a default-export entrypoint.

- `pnpm lint` (`tsc --noEmit`, strict): pass
- `src/agent/plugins/load-entrypoints.test.ts`: 9/9 pass
- full suite: 10348 pass (2 unrelated failures — one pre-existing, one isolation-passing flake)
- `pnpm audit:sdk:check`: OK (no SDK imports added)

### DIFF
--- a/src/agent/plugins/load-entrypoints.test.ts
+++ b/src/agent/plugins/load-entrypoints.test.ts
@@ -11,9 +11,31 @@ import { join } from 'path';
 import { tmpdir } from 'os';
 import { pathToFileURL } from 'url';
 import { loadPluginEntrypoints, _resetLoadedEntrypoints } from './load-entrypoints.js';
+import type { PluginApi } from './load-entrypoints.js';
 import type { SdkPluginConfig } from '../types/sdk-types.js';
 import { registerSkill, getSkill, listSkills, _resetRegistry } from '../../skills/index.js';
 import { loadSkillPrompts } from '../../skills/_lib/prompt-loader.js';
+import { SubagentManager } from '../subagent.js';
+import { describeFailure } from '../subagent/result.js';
+import { env } from '../../config/env.js';
+import { getAgentFrameworkDir, getSkillsDir, getSessionsDir } from '../../paths.js';
+
+// The full host-injected API, mirroring skill-bridge.ensurePluginEntrypointsLoaded.
+// `discoverPluginSkillBodies` is stubbed here (its real wiring is type-checked at
+// the skill-bridge assembly site) so this test need not import that heavy barrel.
+const fullApi: PluginApi = {
+  registerSkill,
+  listSkills,
+  getSkill,
+  loadSkillPrompts,
+  env,
+  SubagentManager,
+  describeFailure,
+  discoverPluginSkillBodies: (() => new Map()) as PluginApi['discoverPluginSkillBodies'],
+  getAgentFrameworkDir,
+  getSkillsDir,
+  getSessionsDir,
+};
 
 let dir: string;
 
@@ -117,9 +139,7 @@ describe('loadPluginEntrypoints', () => {
     );
     const plugins: SdkPluginConfig[] = [{ type: 'local', path: dir, main: 'entry.mjs' }];
 
-    await loadPluginEntrypoints(plugins, {
-      pluginApi: { registerSkill, listSkills, getSkill, loadSkillPrompts },
-    });
+    await loadPluginEntrypoints(plugins, { pluginApi: fullApi });
 
     try {
       expect(listSkills()).toContain(skillName);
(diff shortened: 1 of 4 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

Reviewed at branch HEAD `528498c` (`main` ← `feat/widen-plugin-api-runtime-values`). Every cited line and every claimed export was read back from the GitHub contents API at that exact ref. Stated intent (PR title + body) was available, so spec-compliance was assessed against it.

The change is technically correct, **purely additive**, and spec-compliant. All 7 newly-injected symbols resolve to real named exports at the reviewed ref, the `Pick<typeof import(...)>` derivation structurally prevents signature drift, and the prior 4-symbol contract is unchanged. No blocking (critical/high) findings.

---

### Findings

**1 · medium · security — full `env` (incl. live credentials) is now injected into every code-backed plugin**
`src/agent/tools/skill-bridge.ts:285-293` + `src/config/env.ts:1162-1175` (ref:528498c3, file-state)

The widened `pluginApi` now passes the entire `env` object. `env` exposes live credential getters that re-read `process.env` on every access — verified at the reviewed ref:
```ts
get ANTHROPIC_API_KEY(): string | undefined { return process.env['ANTHROPIC_API_KEY']; }   // env.ts:1162
get OPENAI_API_KEY(): string | undefined  { return process.env['OPENAI_API_KEY']; }         // env.ts:1164
get TELEGRAM_BOT_TOKEN(): string | undefined { return process.env['TELEGRAM_BOT_TOKEN']; }   // env.ts:1175
```
Any code-backed plugin can read `api.env.ANTHROPIC_API_KEY` and obtain the live bearer token. The prior 4-symbol `pluginApi` carried no credentials, so this is a real surface expansion. Plugin trust is conferred by filesystem presence (`scanAllPluginRoots`), not a declared-permission manifest, and there is no key allowlist/scoping.

This is a **trust-boundary decision, not a defect** — injecting `env` is exactly what the PR sets out to enable (a marketplace clone cannot `import 'agent-afk'` to reach it), and if the threat model treats filesystem-present plugins as trusted, the full-`env` exposure is acceptable. Surfacing it so the boundary is an explicit, acknowledged choice.
*Suggestion:* either inject a scoped `env` proxy exposing only non-credential keys plugins legitimately need, or add a one-line comment at the injection site (and the PluginApi JSDoc / plugin-authoring guide) stating the full credential exposure is intentional and bounded by filesystem-trust.

**2 · low · test-coverage — 2 of the 7 injected symbols lack injection assertions**
`src/agent/plugins/load-entrypoints.test.ts:167-200` (ref:528498c3, file-state)

The new round-trip test's `entry.mjs` probe asserts `env`, `SubagentManager`, `describeFailure`, `getAgentFrameworkDir`, and `discoverPluginSkillBodies` arrive intact, but omits `getSkillsDir` and `getSessionsDir` (the test title's `…` quietly elides them). Both are present in `fullApi` and in the production object literal, so this is a coverage gap, not a correctness bug.
*Suggestion:* add `getSkillsDirType`/`getSessionsDirType` to the probe object and corresponding `expect(...).toBe('function')` assertions.

**3 · nit · test-coverage — unsound `as` cast on the `discoverPluginSkillBodies` test stub**
`src/agent/plugins/load-entrypoints.test.ts:34` (ref:528498c3, file-state)
```ts
discoverPluginSkillBodies: (() => new Map()) as PluginApi['discoverPluginSkillBodies'],
```
The cast suppresses the `Map<unknown,unknown>` → `Map<string, PluginSkillBody>` mismatch. Safe today (the test only checks `typeof`, never calls it; the comment is honest about the deferral), but a future test consuming the return value would get a wrongly-typed empty Map with no compile error.
*Suggestion:* type the stub `(() => new Map<string, PluginSkillBody>())` to narrow the cast to a structurally-compatible one.

---

### Clean dimensions (read at ref:528498c3)
- **correctness** — `discoverPluginSkillBodies` is a same-module export (`skill-bridge.ts:215`) referenced in-scope at the injection site (`:291`); no undefined-identifier bug. All `Pick` targets resolve: `SubagentManager` (subagent.ts:223), `describeFailure` (subagent/result.ts:161), `getAgentFrameworkDir`/`getSkillsDir`/`getSessionsDir` (paths.ts:63/91/194).
- **api-compat** — `PluginApi` is a structural type consumed via `pluginApi?: PluginApi`; adding fields is additive for plugins that read a subset. In-repo consumers (`load-entrypoints.ts`, `index.ts` re-export) are unaffected.
- **spec-compliance** — prior 4 symbols preserved identically; the 7 additions match the stated intent; `src/index.ts` change is doc-comment only. No scope creep, no unmet intent.
- **perf-observability** — `ensurePluginEntrypointsLoaded` is idempotent and process-global; the change adds 7 property references on one object literal at boot. No new hot path.

### What was not checked
- Read against branch HEAD `528498c3ecd3a8cbee87fc668c643e39844ba366` via the GitHub contents API (SHA not present in any local checkout) — all blocking-eligible citations and export claims verified at that ref.
- **Out-of-tree** third-party plugins that explicitly typed `pluginApi` as the old 4-symbol shape are unknowable from this repo; the widening is structurally backward-compatible (extra props are benign) but cannot be verified externally.
- `discoverPluginSkillBodies` end-to-end runtime behavior (Map contents, `PluginSkillBody` fields) — the new test stubs it and asserts only `typeof`; not exercised here.
- Tests were not executed in this review (read-only); the PR reports `pnpm lint` pass, `load-entrypoints.test.ts` 9/9 pass, and `audit:sdk:check` OK.

PR #337, opened 2026-06-29
feat(plugins): inject session-facet substrate into PluginApi (11→15)

griffinwork40:
## What
Widens `PluginApi` from 11 → 15 injected symbols, adding the **session-facet substrate**: `getOrDeriveFacet`, `listSessionIds`, `deriveSessionFacet`, `loadStoredSession` (from `src/agent/facets`).

## Why
Follow-up to #335. Extracting the internal skill bundle into a code-backed marketplace plugin requires **harvest**, which calls these four facet functions at runtime (`harvest/index.ts:126,148`; `harvest/_lib/rank.ts:378,396,399`). A marketplace-cloned plugin has **no `node_modules`**, so it cannot `import 'agent-afk'` to reach them — they must be injected, exactly like the runtime values #335 added. The 11-symbol set was insufficient; this closes the gap so a code-backed plugin can consume the host purely via the injected `api`.

## Changes
- `load-entrypoints.ts` — add the facet `Pick` to `PluginApi` (derived via `typeof import('../facets/index.js')` so signatures can't drift from source).
- `skill-bridge.ts` — import + inject the four functions in `ensurePluginEntrypointsLoaded`.
- `load-entrypoints.test.ts` — extend the `fullApi` fixture + the resolution-probe test to assert all four arrive as callable functions.

## Verification
- `pnpm lint` (tsc strict) ✅
- `load-entrypoints.test.ts` + `skill-bridge.test.ts`: 45/45 ✅
- Empirically validated against the published artifact: a flat no-`node_modules` code-backed probe plugin received exactly these symbols via the injected api and round-tripped `registerSkill` → host `listSkills`.

Additive + backward-compatible — existing plugins ignore the new keys. Type-only imports (`SessionFacet`, `StoredSessionInput`) remain erased at build, so no runtime dependency is added.

### DIFF
--- a/src/agent/plugins/load-entrypoints.test.ts
+++ b/src/agent/plugins/load-entrypoints.test.ts
@@ -19,6 +19,12 @@ import { SubagentManager } from '../subagent.js';
 import { describeFailure } from '../subagent/result.js';
 import { env } from '../../config/env.js';
 import { getAgentFrameworkDir, getSkillsDir, getSessionsDir } from '../../paths.js';
+import {
+  deriveSessionFacet,
+  getOrDeriveFacet,
+  listSessionIds,
+  loadStoredSession,
+} from '../facets/index.js';
 
 // The full host-injected API, mirroring skill-bridge.ensurePluginEntrypointsLoaded.
 // `discoverPluginSkillBodies` is stubbed here (its real wiring is type-checked at
@@ -35,6 +41,10 @@ const fullApi: PluginApi = {
   getAgentFrameworkDir,
   getSkillsDir,
   getSessionsDir,
+  getOrDeriveFacet,
+  listSessionIds,
+  deriveSessionFacet,
+  loadStoredSession,
 };
 
 let dir: string;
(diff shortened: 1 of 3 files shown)


griffinwork40 (commented):

**Merge decision: MERGE**

Widens `PluginApi` from 11 → 15 injected symbols, adding the session-facet substrate (`getOrDeriveFacet`, `listSessionIds`, `deriveSessionFacet`, `loadStoredSession`). The change is strictly additive and does its stated job. Reviewed against branch HEAD `90056c5`.

**Findings**
- ✅ **api-compat** — All four symbols are concrete runtime `export function` declarations at the reviewed ref (`store.ts:49,97,126`; `derive.ts:98`), not type-only re-exports. Injecting them as values is correct; the type-only re-exports (`SessionFacet`, `StoredSessionInput`) are correctly omitted from the `Pick`.
- ✅ **correctness / symmetry** — The `PluginApi` `Pick` (`load-entrypoints.ts:57`) and the injected object (`skill-bridge.ts:301–304`) list exactly the same four symbols — no type-vs-runtime asymmetry that would hand a plugin a type lying about what it receives. Existing keys are untouched, so existing plugins that ignore the new keys are unaffected (backward-compatible).
- ✅ **test-coverage** — `load-entrypoints.test.ts` extends both the compile-time `fullApi` fixture (exhaustiveness via the `PluginApi` type) and the runtime resolution-probe, asserting all four arrive as `'function'`.
- ✅ **security / perf** — No secrets, auth paths, or hot-path concerns; imports are static at module load, matching the existing injection pattern.

**Not checked (out of diff scope):** the harvest call-sites cited as motivation (`harvest/index.ts:126,148`; `harvest/_lib/rank.ts:378,396,399`) and the internal implementation semantics of the four facet functions (disk I/O / cache invalidation) were not audited — they fall outside this additive 3-file diff.

_Light-regime review (40 lines changed). Stated intent: PR #337 title+body — spec-compliance assessed. This is an automated, advisory review; treat as input, not a gate._

PR #339, opened 2026-06-30
fix(afk-gate): classify MCP/browser/schedule/web tools so AFK gate covers them (#198)

### LINKED ISSUE #198: AFK gate: document bash+write-only scope; evaluate gating MCP/browser/web/schedule tools
Follow-up from the review of #197 (AFK mode). Severity: **medium**, non-blocking.

## Context
The AFK-mode safety gate (`src/agent/afk-mode-gate.ts`) blocks tool calls that the shared `classifyRisk` (`src/agent/risk-classifier.ts`) rates `high`. Verified during review: `classifyRisk` only rates **bash** and **file-write** (`write_file`/`edit_file`) ops as high-risk. Every other tool class falls through to the `return 'safe'` default — including MCP tools (`mcp__*`), browser actions, `web_scrape`, and `create_schedule`/`cancel_schedule`.

`categorizeTool('mcp__...')` returns `'mcp'` (not `'read'`), so `classifyRisk` skips the read branch and returns `'safe'`; the AFK gate (which blocks only `high`) lets it through.

## Impact
In AFK mode a destructive MCP tool (e.g. `mcp__postgres__drop_table`), a browser form-submit/purchase, or a persisted cron runs **unattended without gating**. The feature is described as a "mechanical safety ceiling," which may lead users wiring MCP/browser servers to over-trust it.

Note: this is an **inherited** property of the pre-existing `classifyRisk` (not changed by #197), and AFK mode is still strictly >= default-mode safety (default gates none of these either).

## Options
1. (Low-effort) Document the gate's bash+write-only scope in `afk-mode-addendum.ts` and the docs so MCP/browser users don't over-trust the ceiling.
2. (Larger) Add an explicit `mcp`/unknown-tool branch in the AFK-gate path (not globally) returning at minimum `medium`, with known-destructive MCP/browser tools mapped to `high`.

griffinwork40:
## Problem

`classifyRisk()` in `src/agent/risk-classifier.ts` fell through to `return 'safe'` for four tool classes that were never explicitly handled:

- **MCP tools** (`mcp__*`): `categorizeTool` returns `'mcp'`, not `'read'`, so the read short-circuit never fired → every MCP tool was rated `'safe'`. A model in AFK mode could call `mcp__postgres__drop_table` unattended.
- **`create_schedule` / `cancel_schedule`**: `categorizeTool` returns `'schedule'` → fell to unknown → `'safe'`. Could silently mutate the daemon cron store.
- **`browser_open` / `browser_act`**: `categorizeTool` returns `'browser'` → fell to unknown → `'safe'`. Could submit forms or click destructive UI unattended.
- **`web_scrape`**: `categorizeTool` returns `'web'` → fell to unknown → `'safe'`. The existing test even incorrectly asserted `'safe'`.

## Policy applied (minimal, defensible)

| Tool class | Old | New | Rationale |
|---|---|---|---|
| `mcp__*` with destructive verb (`delete`, `drop`, `remove`, `write`, `create`, `update`, `insert`, `exec/execute/run`, `eval`, `send`, `push`, `publish`, `deploy`, `post`) | safe | **high** | Irreversible external side-effects; classifier has no visibility into what MCP servers do |
| `mcp__*` without destructive verb | safe | **medium** | May have network/quota side-effects; medium is allowed in AFK and needed for autonomous research |
| `create_schedule` | safe | **high** | Mutates daemon cron store; may live-sync immediately |
| `cancel_schedule` | safe | **high** | Same — permanent until re-created |
| `list_schedules`, `get_schedule_history` | safe | safe (unchanged) | Read-only |
| `browser_open`, `browser_act` | safe | **medium** | Stateful navigation; recoverable so not high; needed for AFK browser automation |
| `browser_observe`, `browser_screenshot`, `browser_close` | safe | safe (unchanged) | Read-only / cleanup |
| `web_scrape` | safe (wrong) | **medium** | Outbound HTTP; recoverable but notable network side-effect |

**Deferred:** Per-server MCP allowlist (e.g. treat `mcp__github__list_issues` as `'safe'`). The verb-filter catches the clearly-dangerous cases; a future per-server config can refine individual servers. Noted in code comments.

## Files changed

- `src/agent/risk-classifier.ts` — four new classified sections (MCP, schedule mutations, browser actions, web_scrape); invariant comments follow the ≥15-line prefix convention
- `src/agent/risk-classifier.test.ts` — 20 new test cases (schedule ×4, browser ×5, MCP ×11); fixed existing wrong `web_scrape → safe` assertion
- `src/agent/afk-mode-gate.ts` — updated module-level policy comment to name the two newly-gated classes

## Test evidence

```
Risk-classifier: 58 passed (58)   ← +20 new cases, 0 failures
AFK-mode-gate:   43 passed (43)   ← all existing, unchanged
Lint:            tsc --noEmit clean
```

Closes #198

### DIFF
--- a/src/agent/afk-mode-gate.ts
+++ b/src/agent/afk-mode-gate.ts
@@ -22,6 +22,11 @@
  *     `chmod`/`chown`)
  *   - writes that escape the workspace, hit the write-denylist (`~/.ssh`,
  *     `/etc`, …), or target the `.git` object store
+ *   - schedule mutations: `create_schedule` / `cancel_schedule` — modify the
+ *     daemon cron store with potential immediate live-sync side-effects
+ *   - MCP tools with destructive verb sub-names (`*delete*`, `*drop*`,
+ *     `*write*`, `*create*`, `*update*`, `*exec*`, `*send*`, `*deploy*`, …) —
+ *     arbitrary third-party server functions the classifier cannot introspect
  * `'medium'` ops (normal `git push`/`git commit`, installs, builds, file moves)
  * and `'safe'` ops (reads, tests, lint) are ALLOWED — autonomous work has to be
  * useful, and these are reversible enough to run unattended.
--- a/src/agent/risk-classifier.test.ts
+++ b/src/agent/risk-classifier.test.ts
@@ -187,8 +187,8 @@ describe('classifyRisk — other tools', () => {
     expect(classifyRisk('list_directory', { path: '.' }, ctx)).toBe('safe');
   });
 
-  it('web_scrape → safe', () => {
-    expect(classifyRisk('web_scrape', { url: 'https://example.com' }, ctx)).toBe('safe');
+  it('web_scrape → medium (network side-effect)', () => {
+    expect(classifyRisk('web_scrape', { url: 'https://example.com' }, ctx)).toBe('medium');
   });
 
   it('send_telegram → medium', () => {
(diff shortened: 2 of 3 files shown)


griffinwork40 (commented):

## Merge decision: **DO NOT MERGE** (one blocking security gap in the gate this PR installs; concrete one-line fix)

This is a genuinely good, well-tested hardening change — it correctly closes the `safe`-fallthrough for MCP/schedule/browser/web tools, every policy-table row is implemented and **reachable** (verified: `categorizeTool` returns `mcp`/`schedule`/`browser`/`web`, never `read`, so the new branches are live, not dead code), and it's api-compat clean (the only production consumer is `afk-mode-gate.ts`, which gates only on `'high'` — so `safe→medium` is a no-op and `safe→high` is the intended new gating). The blocker below is that the fix ships with a verified hole in the exact threat model it targets.

All findings re-derived independently against the reviewed ref `1177d153` with step-by-step computed values — not paraphrased from the diff.

---

### 🔴 BLOCKING — high · security · `src/agent/risk-classifier.ts` (subName extraction, ~L243)
**2-segment destructive MCP names bypass the gate this PR exists to install.** The sub-name is extracted with `.split('__').slice(2)`, which drops the server segment. For a 2-segment name (no server component) the result is the empty string, so the destructive-verb scan never matches and the tool is rated `'medium'` (allowed unattended) instead of `'high'`.

```ts
const subName = toolName.split('__').slice(2).join('__').toLowerCase();
// 'mcp__deploy'.split('__')      = ['mcp','deploy']
//   .slice(2)                    = []           → subName = ''
//   ''.includes('deploy')        = false        → returns 'medium'   ❌ (gate bypassed)
// 'mcp__infra__deploy'           → subName='deploy' → 'high'          ✓ (only the 3-segment form is gated)
```
Affects any `mcp__<verb>` (e.g. `mcp__deploy`, `mcp__reset`, `mcp__exec`, `mcp__delete`). The PR's own test only exercises the 3-segment `mcp__infra__deploy` form, so this slips through green CI.
**Fix:** change `.slice(2)` → `.slice(1)` to keep the full suffix after `mcp__` (`'mcp__deploy'.split('__').slice(1).join('__')` = `'deploy'`), and add a 2-segment regression test expecting `'high'`.

---

### 🟠 medium · security · `DESTRUCTIVE_VERBS` array (~L248)
Several realistic destructive MCP operations are not in the verb list and therefore rate `'medium'` (run unattended in AFK). Verified by re-derivation:
```
mcp__payments__charge        → subName 'charge'             → 'medium'   (financial)
mcp__github__merge_pr        → subName 'merge_pr'           → 'medium'   (irreversible repo write)
mcp__aws__terminate_instance → subName 'terminate_instance' → 'medium'   (infra destruction)
```
Also missing: `wipe`, `rename`/`mv`, `revoke`, `disable`, `rollback`, `scale`, `provision`, `refund`.
**Fix:** extend the list (`terminate`, `merge`, `charge`, `refund`, `revoke`, `disable`, `scale`, `wipe`, `rename`, `provision`), and pair with the word-boundary fix below to avoid new false positives.

### 🟠 medium · security · verb match uses substring `.includes()` (~L264)
`'run'` and `'exec'` substring-match benign read tools, wrongly gating them `'high'`: `mcp__test__runner`, `mcp__server__executor`, `mcp__ci__get_run_status` all → `'high'`. Safe-side, but the friction may push operators to disable the gate.
**Fix:** token-match on `_`/`-` boundaries (`subName.split(/[_-]/).some(t => VERBS.has(t))`) — keeps `execute`, drops `runner`/`executor`.

---

### Non-blocking nits
- **low · spec/docs · `afk-mode-gate.ts:25-30`** — the JSDoc lists an abbreviated verb set (8) while the code array has 21; will drift. Replace the inline list with a pointer to `DESTRUCTIVE_VERBS`.
- **low · correctness · MCP prefix check (~L262)** — uses `toolName.startsWith('mcp__') || 'MCP__'` rather than the already-lowercased `tool`; a mixed-case `Mcp__…` falls through to `'safe'` (same latent gap exists in `categorizeTool`, so it's consistent). Prefer the single predicate `tool.startsWith('mcp__')`.
- **low · test-coverage** — add tests for the 2-segment MCP case and a `run`-false-positive case to lock the policy in.
- **low · wording · test L215** — `cancel_schedule` is reversible at its default (`enabled:false`); only `permanent:true` removes it. `'high'` is still the right classification; the "irreversible" label in the test name over-claims.

---

**What was verified / not:** All citations read at branch HEAD `1177d153` and independently re-derived (computed `.split`/`.slice`/`.includes` values match). Stated intent = PR #339 title+body — spec-compliance fully assessed (all 8 policy rows implemented + reachable). api-compat: only production caller is `afk-mode-gate.ts` (`if (risk !== 'high') return {}`). Not run: the live test suite (static read only); real-world MCP server naming conventions (the 2-segment exposure is structural — its practical reach depends on whether any connected server uses `mcp__<verb>` names).

Net: ship-quality change with one verified gate-bypass to close first. Re-rate `mcp__deploy`-style 2-segment names to `'high'` and the blocker clears.

PR #341, opened 2026-06-30
fix(daemon): dedup repeated quarantine log noise for poison entries (#252)

### LINKED ISSUE #252: Rate-limit repeated quarantine log noise for permanently-unremovable poison entries
## Context
Follow-up from #241. `quarantinePoisonEntry()` moves malformed queue files into `poison/`, with a best-effort `unlinkSync` fallback if the move fails.

## Problem
When a poison entry can be neither renamed into `poison/` nor unlinked (e.g. queue-dir permission loss), it remains at the FIFO head and `quarantinePoisonEntry()` re-emits up to three `console.error` lines for it on **every** poll tick (~6 lines/min on a 30s interval), indefinitely, with no dedup or backoff.

The queue is **not** deadlocked — `dequeueNext()`'s loop still `continue`s past the stuck entry and drains valid tasks behind it — so this is log spam in a degraded state, not a hang.

## Fix
Add a module-scope `Map<filename, count>` so a permanently-stuck entry logs on first failure and then only every Nth tick thereafter.

## Related
Adjacent to the deferred `NOTE(#337-poison-telemetry)` / `NOTE(#337-poison-gc)` follow-ups noted in the code.

## Acceptance criteria
- A permanently-unremovable entry produces bounded log output across many ticks.
- Valid tasks queued behind it still drain.

## Pointers
- `src/agent/daemon/queue-store.ts` — `quarantinePoisonEntry()`, `dequeueNext()`
- `src/agent/daemon/scheduler.ts` — `pullTick()`

griffinwork40:
## Summary

Fixes #252.

When a poison queue entry can be neither renamed into `poison/` nor unlinked (e.g. queue-dir permission loss), it stays at the FIFO head and `quarantinePoisonEntry()` re-emitted up to 3 `console.error` lines per poll tick — ~6 lines/min on a 30s daemon interval — indefinitely.

## Behavior chosen: rate-limited (log on tick 1, then every 10th tick)

Per the issue spec, a module-scope `Map<string, number>` (`stuckEntryEncounters`) is keyed by `'\<queueDir>/\<filename>'`. The failed-unlink branch (innermost `catch (unlinkErr)` in `quarantinePoisonEntry`) now:

1. Increments the per-entry counter
2. Emits `console.error` only when `count === 1` or `count % STUCK_LOG_INTERVAL === 0` (every 10th tick)
3. Appends `[seen Nx]` so the operator can gauge retry depth at a glance

The upstream `renameSync`-failed log (outer catch) is **not** rate-limited — it fires only when the rename fails, after which the fallback tries `unlinkSync`. Only the final `unlinkSync`-also-failed branch is permanently sticky.

## Tradeoff noted

`stuckEntryEncounters` is module-scoped and never shrinks automatically. It grows by at most one entry per distinct permanently-stuck file — a successfully quarantined or unlinked entry never reaches this path — so it stays near-empty (0–1 entries) for the single-operator CLI use-case.

## Files changed

- `src/agent/daemon/queue-store.ts` — `STUCK_LOG_INTERVAL` constant, `stuckEntryEncounters` map, rate-limit guard in inner-catch
- `src/agent/daemon/queue-store-poison-fallback.test.ts` — new describe block with 2 tests

## Test evidence

```
Test Files  1 passed (1)
     Tests  7 passed (7)   ← includes 2 new dedup tests
  Duration  659ms
```

New tests (using existing `vi.mock('node:fs')` infrastructure):
- **tick 1 logs, tick 2 suppressed** — `[seen 1x]` present on tick 1, zero stuck-log lines on tick 2
- **re-logs on tick 10** — exactly 2 logged occurrences across 10 ticks (ticks 1 and 10)

Full queue-store suite (42 tests) also passes clean.

`pnpm lint` (`tsc --noEmit`) passes with zero errors.

Closes #252

### DIFF
--- a/src/agent/daemon/queue-store-poison-fallback.test.ts
+++ b/src/agent/daemon/queue-store-poison-fallback.test.ts
@@ -171,3 +171,61 @@ describe('quarantinePoisonEntry — inner-catch (branch 2: rename AND unlink fai
     expect(result).toBeNull();
   });
 });
+
+describe('quarantinePoisonEntry — stuck-entry log dedup (issue #252)', () => {
+  it('logs on the 1st encounter but suppresses the 2nd (rate-limit)', () => {
+    // Arm: both rename and unlink fail so the entry is permanently stuck.
+    renameSyncShouldThrow = true;
+    unlinkSyncShouldThrow = true;
+
+    realFs.writeFileSync(join(tmpDir, '0001-q-stuck.json'), 'bad-json');
+
+    const spy = vi.spyOn(console, 'error').mockImplementation(() => undefined);
+    try {
+      // Tick 1 — must log (count === 1).
+      dequeueNext(tmpDir);
+      const callsAfterTick1 = spy.mock.calls.length;
+      expect(callsAfterTick1).toBeGreaterThan(0);
+
+      // Capture the tick-1 log lines so we can compare below.
+      const logsAfterTick1 = spy.mock.calls.map((c) => String(c[0]));
+      expect(logsAfterTick1.some((m) => m.includes('[seen 1x]'))).toBe(true);
+
+      // Tick 2 — same stuck entry, count === 2 (not 1 and not a multiple of 10).
+      // The "could not remove" line must NOT be emitted again.
+      spy.mockClear();
+      dequeueNext(tmpDir);
+      const stuckLogsOnTick2 = spy.mock.calls
+        .map((c) => String(c[0]))
+        .filter((m) => m.includes('could not remove unquarantinable'));
+      expect(stuckLogsOnTick2).toHaveLength(0);
+    } finally {
+      spy.mockRestore();
+    }
+  });
+
+  it('re-logs on the 10th encounter (every STUCK_LOG_INTERVAL ticks)', () => {
+    renameSyncShouldThrow = true;
+    unlinkSyncShouldThrow = true;
+
+    realFs.writeFileSync(join(tmpDir, '0001-q-stuck.json'), 'bad-json');
+
+    const spy = vi.spyOn(console, 'error').mockImplementation(() => undefined);
+    try {
+      // Simulate 10 ticks. Only ticks 1 and 10 should log the stuck message.
+      let stuckLogCount = 0;
+      for (let tick = 1; tick <= 10; tick++) {
+        spy.mockClear();
+        dequeueNext(tmpDir);
+        const hasStuckLog = spy.mock.calls.some((c) =>
+          String(c[0]).includes('could not remove unquarantinable'),
+        );
+        if (hasStuckLog) stuckLogCount += 1;
+      }
+      // Ticks 1 and 10 log; ticks 2–9 do not → exactly 2 logged occurrences.
+      expect(stuckLogCount).toBe(2);
+    } finally {
+      spy.mockRestore();
+    }
+  });
+});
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

## Merge decision: **MERGE** (one non-blocking comment-accuracy nit)

This is a clean, well-scoped hotfix that does exactly what its description says: it rate-limits the inner-catch `console.error` in `quarantinePoisonEntry` so a permanently-unremovable poison entry no longer floods stderr on every poll tick. The counter logic is correct, the operator-facing log change is safe, secret redaction is preserved, and the two new tests are properly isolated. All CI checks are green at the reviewed ref. The single finding below is a comment that over-claims its own invariant — worth a one-line fix, not a merge blocker.

All findings were independently re-derived against the reviewed ref `22101967` (file content read at that SHA, not paraphrased from the diff).

---

### 🟡 LOW — correctness/docs · `src/agent/daemon/queue-store.ts:107–113` · ref:`22101967`
**The comment's "shrinks to zero when those files are eventually cleared" claim is not implemented — no code path ever removes a key from `stuckEntryEncounters`.**

```ts
// ...The map grows by at most one entry per distinct permanently-unremovable
// poison file and shrinks to zero when those files are eventually cleared
// (by the operator)...
```
Independent grep at the reviewed ref shows `stuckEntryEncounters` is touched at exactly four sites — the comment (L107), the declaration (L114), `.get()` (L245), and `.set()` (L246) — and the file contains **no** `.delete(` or `.clear(` anywhere. A key is therefore only ever added, never pruned within a process lifetime; the map resets only on daemon restart. In practice this is harmless for the single-operator CLI (the steady-state size really is 0–1, and each stale key is a small string + number), so it does not block merge — but the "shrinks to zero" sub-clause is literally false as written.

**Fix (pick one):**
- Drop the "and shrinks to zero when those files are eventually cleared" clause and reword to "…never shrinks within a process lifetime; resets on daemon restart"; **or**
- Add `stuckEntryEncounters.delete(stuckKey)` on the successful `renameSync`/`unlinkSync` paths so the claim becomes true.

---

### Verified clean (no action needed)
- **Cross-test state leakage — refuted.** The module-scoped map persists across the two new `it()` blocks, but `beforeEach` creates a fresh `mkdtempSync` dir per test, so the two stuck-keys are distinct paths. Test 2's counter genuinely starts at 0, and `expect(stuckLogCount).toBe(2)` is correct.
- **Counter semantics — correct.** `count === 1 || count % 10 === 0` logs on ticks 1 and 10 and suppresses 2–9; the test's expectation of exactly 2 occurrences across 10 fresh ticks holds.
- **Secret redaction — preserved.** `redactedFilename` is pre-redacted upstream and reused in the new log line; `unlinkReason` is passed through `redactInlineSecrets`. No raw filename or error reaches stderr.
- **api-compat — no breakage.** `stuckEntryEncounters`/`STUCK_LOG_INTERVAL`/`quarantinePoisonEntry` are all module-private; the only changed surface is operator-facing stderr text, and the new tests substring-match `'could not remove unquarantinable'`, which spans both old and new formats.

---

**What was verified / not:** All citations read at branch HEAD `22101967` and independently re-derived (the absence of any `.delete`/`.clear` was confirmed by direct grep; the test isolation was confirmed by reading `beforeEach`). Stated intent = PR #341 title + body — spec-compliance fully assessed (rate-limit-on-tick-1-then-every-10th implemented as specified). Not run: the live test suite and `tsc --noEmit` (static read only; CI reports both green at this ref); concurrent multi-daemon behavior (out of scope for the single-operator CLI use-case).

Net: ship-quality change. The lone finding is a comment that over-promises; merge as-is or tighten the wording in a follow-up.

PR #345, opened 2026-06-30
fix(hooks): merge injectContext across non-blocking handlers instead of last-wins

griffinwork40:
## Problem

When multiple non-blocking hook handlers registered on the same event each return `{ injectContext: "..." }`, only the **last** handler's value was kept — all earlier injections were silently dropped. This blocked any workflow that registers multiple gate-hooks on one lifecycle event, since all but the final registered handler would lose their injected context.

## Root cause

`HookRegistryImpl.dispatch()` in `src/agent/hook-registry.ts` (line 218–224) tracked accumulated decisions with a single `lastDecision` variable. Each non-blocking handler **replaced** `lastDecision` wholesale, so earlier `injectContext` values were overwritten:

```ts
// Before (last-wins):
lastDecision = decision;
```

A comment at that site explicitly documented this as the existing behavior: *"decisions are not merged. If multiple SubagentStop handlers return injectContext, the last handler's value wins and earlier values are dropped."*

## Fix

In the dispatch loop, for non-blocking decisions:
- **`injectContext`**: all non-empty values are concatenated in registration order, joined by `'\n'`. Empty/absent values are skipped so there are no leading, trailing, or duplicate separators.
- **All other decision fields**: last-handler-wins (unchanged), so callers reading `reason`, `decision`, `continue` see the final handler's value.
- **Return type unchanged**: still returns a single `HookDecision` with `injectContext?: string` — no consumer changes needed.
- **Blocking short-circuit unchanged**: a handler returning `{ continue: false }` or `{ decision: 'block' }` still throws `HookBlockedError` and fires **before** any accumulation.

Updated both doc comments (`hook-registry.ts` and `hooks.ts`) to describe the new merge policy.

## Blast radius

Zero consumer changes required. All downstream readers consume `injectContext` as a single string, which is preserved:
- `src/agent/subagent/handle.ts:470` — `decision.injectContext && this.parentInputStreamRef...pushUserMessage(decision.injectContext)`
- `src/cli/commands/interactive/loop-iteration.ts:433` — `upsDecision.injectContext + runText`
- Trace recorders in `subagent-hooks.ts` and `session/hooks-dispatch.ts` — read `decision.injectContext` as a string field

## Tests

- **Updated** `src/agent/hooks.test.ts` test at line ~292: renamed from `'keeps the last injectContext when multiple stop hooks return one'` → `'concatenates injectContext when multiple stop hooks return one'`, and updated assertion to expect both values concatenated (`'first framework note\nsecond framework note'`).
- **Added** 7 new tests in `src/agent/hook-registry.test.ts` under a new `'HookRegistryImpl.dispatch — injectContext merge policy'` describe block:
  - (a) two non-blocking handlers concatenated in registration order
  - (b) three non-blocking handlers concatenated in registration order
  - (c) single handler returns its value unchanged
  - (d) handlers returning no `injectContext` are skipped — no leading/trailing newlines
  - (e) no handler returns `injectContext` → `undefined`
  - (f) blocking handler short-circuits before accumulation — later handler never called
  - (g) last-handler-wins preserved for non-`injectContext` fields while merging `injectContext`

## Verification

**`pnpm lint` (`tsc --noEmit`)**:
```
> agent-afk@5.14.0 lint
> tsc --noEmit
```
Zero errors.

**Test suite (9 required files)**:
```
Test Files  9 passed (9)
Tests       171 passed (171)
```
Files: `hook-registry.test.ts` (22), `hooks.test.ts` (19), `hooks-integration.test.ts` (34 via worktree), `shadow-verify-nudge.test.ts` (21), `command-executor.test.ts` (28), `config-bridge.test.ts` (16), `user-prompt-submit.test.ts` (7), `loop-iteration.test.ts` (15), `wiring.test.ts` (9). All pass.

### DIFF
--- a/src/agent/hook-registry.test.ts
+++ b/src/agent/hook-registry.test.ts
@@ -382,3 +382,88 @@ describe('[R3 follow-up] dispatchSubagentStop — abort mid-dispatch', () => {
     }
   });
 });
+
+// ---------------------------------------------------------------------------
+// injectContext merge policy
+// ---------------------------------------------------------------------------
+//
+// Non-blocking handlers that return injectContext must have their values
+// concatenated in registration order (joined by '\n'), not silently dropped by
+// last-wins. Blocking handlers still short-circuit before any accumulation.
+describe('HookRegistryImpl.dispatch — injectContext merge policy', () => {
+  it('concatenates injectContext from two non-blocking handlers in registration order', async () => {
+    const registry = createHookRegistryImpl();
+    registry.register('SubagentStop', async () => ({ injectContext: 'first note' }));
+    registry.register('SubagentStop', async () => ({ injectContext: 'second note' }));
+
+    const decision = await registry.dispatch({ event: 'SubagentStop', subagentId: 'x' });
+    expect(decision.injectContext).toBe('first note\nsecond note');
+  });
+
+  it('concatenates injectContext from three non-blocking handlers in registration order', async () => {
+    const registry = createHookRegistryImpl();
+    registry.register('SubagentStop', async () => ({ injectContext: 'alpha' }));
+    registry.register('SubagentStop', async () => ({ injectContext: 'beta' }));
+    registry.register('SubagentStop', async () => ({ injectContext: 'gamma' }));
+
+    const decision = await registry.dispatch({ event: 'SubagentStop', subagentId: 'x' });
+    expect(decision.injectContext).toBe('alpha\nbeta\ngamma');
+  });
+
+  it('returns the single handler value unchanged when only one handler provides injectContext', async () => {
+    const registry = createHookRegistryImpl();
+    registry.register('SubagentStop', async () => ({ injectContext: 'only note' }));
+
+    const decision = await registry.dispatch({ event: 'SubagentStop', subagentId: 'x' });
+    expect(decision.injectContext).toBe('only note');
+  });
+
+  it('skips handlers that return no injectContext — no leading/trailing newlines', async () => {
+    const registry = createHookRegistryImpl();
+    registry.register('SubagentStop', async () => ({}));
+    registry.register('SubagentStop', async () => ({ injectContext: 'middle note' }));
+    registry.register('SubagentStop', async () => ({}));
+
+    const decision = await registry.dispatch({ event: 'SubagentStop', subagentId: 'x' });
+    expect(decision.injectContext).toBe('middle note');
+    // No stray separator on either end.
+    expect(decision.injectContext?.startsWith('\n')).toBe(false);
+    expect(decision.injectContext?.endsWith('\n')).toBe(false);
+  });
+
+  it('produces no injectContext when no handler returns one', async () => {
+    const registry = createHookRegistryImpl();
+    registry.register('SubagentStop', async () => ({}));
+    registry.register('SubagentStop', async () => ({}));
+
+    const decision = await registry.dispatch({ event: 'SubagentStop', subagentId: 'x' });
+    expect(decision.injectContext).toBeUndefined();
+  });
+
+  it('blocking handler short-circuits before accumulation — later injectContext not reached', async () => {
+    const registry = createHookRegistryImpl();
+    const later = vi.fn(async () => ({ injectContext: 'should-not-appear' }));
+    registry.register('SubagentStop', async () => ({ decision: 'block', reason: 'policy' }));
+    registry.register('SubagentStop', later);
+
+    await expect(registry.dispatch({ event: 'SubagentStop', subagentId: 'x' })).rejects.toThrow();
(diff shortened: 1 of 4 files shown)


griffinwork40 (commented):

## Merge decision: **MERGE**

This is a clean, tightly-scoped hotfix that does exactly what its description says: it replaces the last-wins `injectContext` behavior in `HookRegistryImpl.dispatch()` with an order-preserving concatenation, so multiple non-blocking hook handlers on one lifecycle event no longer silently drop each other's injected context. The merge logic is correct, the blocking short-circuit is preserved, the return contract is unchanged, and the change ships with thorough test coverage (7 new + 1 updated). All CI checks are green at the reviewed ref, and the "zero consumer changes" blast-radius claim is verified against the four production readers.

All findings were independently re-derived by reading file content at the reviewed ref `dbd9516` (via the GitHub contents API at that SHA), not paraphrased from the diff.

---

### Verified correct (no action needed)

- **Merge logic — correct.** `lastDecision` initializes as `{}` (`hook-registry.ts:160`), so on the first handler `lastDecision.injectContext` is `undefined`; the `typeof s === 'string' && s.length > 0` filter (`:227`) drops it, and the single value passes through unchanged. Subsequent non-empty values concatenate with `'\n'`; empty/absent values are skipped — no leading, trailing, or duplicate separators. When no handler returns a value, the spread omits `injectContext` and it stays `undefined`.
- **Registration order — preserved.** The loop iterates `snapshot = list.slice()` in registration order and prepends the accumulator (`lastDecision.injectContext`) before the current `decision.injectContext`, so the joined string is in registration order.
- **Blocking short-circuit — preserved.** The `isBlocking(decision)` check (`hook-registry.ts:211`) throws `HookBlockedError` *before* the accumulation block (`:226`), so a blocking handler never contributes to the merge and later handlers are never reached. Locked in by the new short-circuit test.
- **api-compat — no breakage.** `HookDecision.injectContext` stays `string | undefined`; all four production readers consume it as a single string — `subagent/handle.ts:470,477` (truthy-check → `pushUserMessage`), `cli/.../loop-iteration.ts:433-434` (prepend to prompt), and the trace recorders `subagent-hooks.ts:83` / `session/hooks-dispatch.ts:67` (byte-length only). A longer string is transparent to all of them.
- **Test coverage — strong.** New tests cover two- and three-handler concatenation, single-handler passthrough, empty-skip (no stray separators), all-empty → `undefined`, blocking short-circuit, and last-wins preserved for non-`injectContext` fields. The updated `hooks.test.ts` case correctly flips its assertion to the concatenated value.
- **Docs — aligned.** The dispatch-loop comment, `hooks.ts` module doc, and the `HookDecision.injectContext` JSDoc were all updated to describe the new merge policy; no stale "last-wins" doc remains for the SubagentStop path.

---

### 🟡 LOW — correctness/docs · `src/agent/hook-registry.ts:229-233` · ref:`dbd9516`

**The accumulator spread carries forward *all* earlier fields, which is a slight widening beyond the "last-handler-wins (same as before)" the comment claims — benign for current consumers.**

```ts
lastDecision = {
  ...lastDecision,
  ...decision,
  ...(mergedInjectContext.length > 0 ? { injectContext: mergedInjectContext } : {}),
};
```

Under the old `lastDecision = decision` wholesale replace, a field set by an earlier handler and *omitted* by a later one was dropped. With the spread, that earlier field now persists. For any field the last handler *does* set, last-wins still holds — so the observable contract for the documented fields (`reason`, `decision`, `continue`) is unchanged, and no production reader consumes those off the non-blocking merged result (the blocking `reason` is read at the throw site, before this block). This is arguably more correct, not a defect. Only the "(same as before)" phrasing slightly under-describes it.

**Fix (optional, non-blocking):** reword the comment to "last-handler-wins for fields the final handler sets; fields from earlier handlers persist if the last handler omits them" — or, if strict last-wins is intended for non-`injectContext` fields, replace the spread with an explicit `{ ...decision, injectContext: mergedInjectContext || undefined }`.

---

**What was verified / not:** All citations read at branch HEAD `dbd9516` (GitHub contents API at that SHA) and independently re-derived — the merge fold, the `{}` init, the blocking short-circuit ordering, and all four production `injectContext` readers were read directly, not inferred from the diff. Stated intent = PR #345 title + body; spec-compliance fully assessed — every clause (concatenate in registration order, skip empties, last-wins for other fields, unchanged return type, blocking short-circuit before accumulation) is implemented and covered by a test. No absence claims were made (presence of consumers was verified, not asserted-absent). Not run: the live test suite and `tsc --noEmit` (static read only; CI reports Lint & Build, Test, Docs, and Publish Bundle Smoke all green at this ref); no `critical`/`high` findings, so no adversarial re-derivation pass was required.

Net: ship-quality change. The lone finding is a one-line comment nuance; merge as-is or tighten the wording in a follow-up.

PR #349, opened 2026-07-01
fix(bash): close abort TOCTOU so an in-flight child is killed, not orphaned

griffinwork40:
## Context
Follow-up to the Telegram watchdog fix. When a turn is interrupted (`session.interrupt()`), the bash tool's abort listener SIGKILLs the in-flight child process group — but only if the listener was registered **before** the abort fired.

## Root cause (TOCTOU)
An abort landing in the window between the top-of-handler `signal.aborted` pre-flight check (which passes, so the child **is** spawned) and the `addEventListener('abort', …)` registration is never delivered — a real `AbortSignal` does not replay an already-dispatched `'abort'` to a late-added listener. The child then runs to completion and `proc.on('close')` only *then* settles as "Command aborted", leaking a late tool event after the turn was supposed to end (the stray fragment observed in the field).

## Fix
Idempotent post-registration re-check in `src/agent/tools/handlers/bash.ts`: right after registering the listener, `if (signal.aborted) abortHandler()`. `settle()` guards on `resolved`, so re-firing is safe. Confined to `bash.ts`.

## Test plan
- `pnpm exec tsc --noEmit` clean.
- `bash.test.ts` 54/54 incl. 1 new: a fake signal that flips `aborted=true` during `addEventListener` reproduces the race; the child is killed promptly (`Command aborted` in <3s, not after the 5s sleep completes via the close handler).

## Risk
Low; narrow, additive guard. Same process-group SIGKILL path already covered by existing abort tests.

### DIFF
--- a/src/agent/tools/handlers/bash.test.ts
+++ b/src/agent/tools/handlers/bash.test.ts
@@ -247,6 +247,33 @@ describe('bashHandler', () => {
       expect(result.isError).toBeFalsy();
       expect(result.content).toContain('success');
     });
+
+    it('[TOCTOU] kills the child promptly when abort lands after the pre-flight check but before listener registration', async () => {
+      // Regression: an abort firing in the window between the top-of-handler
+      // `signal.aborted` pre-flight check (passes → the child IS spawned) and the
+      // `addEventListener('abort', ...)` registration is never delivered to the
+      // late-added listener (a real AbortSignal does not replay an
+      // already-dispatched 'abort'). Without the post-registration re-check the
+      // child runs to completion and leaks a late result. This fake signal
+      // reproduces that exact race: `aborted` is false at the pre-flight read,
+      // then flips true when the handler registers its listener.
+      let aborted = false;
+      const raceSignal = {
+        get aborted() { return aborted; },
+        addEventListener: (_type: string, _cb: () => void) => { aborted = true; },
+        removeEventListener: () => {},
+      } as unknown as AbortSignal;
+
+      const start = Date.now();
+      const result = await bashHandler({ command: 'sleep 5' }, raceSignal);
+      const elapsed = Date.now() - start;
+
+      expect(result.isError).toBe(true);
+      expect(result.content).toBe('Command aborted');
+      // The fix kills the child immediately; without it, the result would only
+      // arrive after `sleep 5` exits on its own (~5s) via the close handler.
+      expect(elapsed).toBeLessThan(3000);
+    }, 15_000);
   });
 
   describe('ANSI stripping', () => {
--- a/src/agent/tools/handlers/bash.ts
+++ b/src/agent/tools/handlers/bash.ts
@@ -270,6 +270,15 @@ export function createBashHandler(
       settle({ content: 'Command aborted', isError: true });
     };
     signal.addEventListener('abort', abortHandler);
+    // Close the TOCTOU window between the pre-flight `signal.aborted` check (top
+    // of the handler) and this listener registration: an abort that fired in
+    // that gap never invokes `abortHandler` (addEventListener does not replay an
+    // already-dispatched 'abort' event), so the just-spawned child would run to
+    // completion and leak a late result instead of being killed promptly.
+    // settle() is idempotent (guards on `resolved`), so re-firing here is safe.
+    if (signal.aborted) {
+      abortHandler();
+    }
 
     // Normal completion — `close` fires after all stdio streams drain.
     proc.on('close', (code) => {


griffinwork40 (commented):

# Code Review — PR #349

**`fix(bash): close abort TOCTOU so an in-flight child is killed, not orphaned`** · main ← fix/bash-abort-toctou-race
**Reviewed ref:** `31822b4ae6060a88c239680a309b5fe156fb21ec` · 2 files · +36 · regime: light (hotfix)

## Verdict: ✅ MERGE

Correct, minimal, well-tested fix for a real race. The added guard closes exactly the TOCTOU window the PR describes, the re-fire is genuinely idempotent, and the new test discriminates fixed-vs-unfixed behavior. Two **low, non-blocking** findings — neither gates merge.

## What I verified as correct (source-grounded at `31822b4`)
- **The window is real.** The pre-flight `if (signal.aborted) return` (bash.ts:106-108) returns *before* `spawn(command, { detached: true, … })` (bash.ts:133-140). An abort landing between that read and `signal.addEventListener('abort', abortHandler)` is never delivered to the late-added listener (a real `AbortSignal` does not replay an already-dispatched `'abort'`). ✓
- **The fix closes it.** `if (signal.aborted) { abortHandler(); }` immediately after registration (bash.ts:279-281) SIGKILLs the process group and settles. ✓
- **Re-firing is safe.** `settle()` guards on `resolved` (bash.ts:115-121, idempotent), and `proc.on('close')` only re-calls `settle()`, never `process.kill` again. A real `AbortController` fires `'abort'` at most once, so `abortHandler` runs via exactly one of {late-listener, line-280 re-check} — the "double-kill" path is structurally impossible. ✓
- **No API break.** `export function createBashHandler` (bash.ts:85) and `export const bashHandler` (bash.ts:343) are untouched; the sole production importer (`handlers/index.ts:11,59-60`) needs no change. ✓
- **No scope creep.** Diff confined to `bash.ts` (+9) and `bash.test.ts` (+27); test count 53 → 54 matches the PR's "54/54 incl. 1 new." ✓
- **The new test genuinely discriminates.** The fake signal's `addEventListener` flips `aborted=true` without invoking the callback, so without the line-279 guard the handler would only settle via `proc.on('close')` after `sleep 5` (~5000ms); the `elapsed < 3000` assertion fails on the unfixed code. ✓

## Findings (both non-blocking)

**1 · LOW · correctness · `src/agent/tools/handlers/bash.ts:266-270,279-281` · ref:31822b4 · citation-type: file-state**
`abortHandler`'s `process.kill(-proc.pid, 'SIGKILL')` has no try/catch, so if the process group were already gone the new call site would throw synchronously inside the Promise executor, surfacing a generic tool error instead of `Command aborted`.
`evidence:`
```
const abortHandler = () => {
  if (proc.pid !== undefined) {
    process.kill(-proc.pid, 'SIGKILL');
  }
```
Reachability is near-zero: no I/O yield occurs between `spawn` (bash.ts:133) and the guard (bash.ts:279), so the child cannot realistically have exited and been reaped in that window. This is a **pre-existing pattern** (already at bash.ts:266-270); the PR only adds one more call site to it, so it is not a regression this change introduces.
`suggestion:` Wrap the `process.kill` call in try/catch and swallow `ESRCH`. [low confidence — verify with runtime context]

**2 · LOW · test-coverage · `src/agent/tools/handlers/bash.test.ts:251-273` · ref:31822b4 · citation-type: diff-context**
No test exercises `abortHandler`'s `process.kill` against an already-exited/reaped process group (the ESRCH edge above).
`evidence:` `git grep -n "ESRCH" 31822b4 -- src/agent/tools/handlers/bash.test.ts` → 0 matches.
`suggestion:` Add a case where `proc.on('close')` settles first, then a synthetic real-abort dispatch follows, asserting no unhandled rejection. Optional — the ESRCH path is near-unreachable in practice.

- Citations verified against branch HEAD `31822b4`; both findings' evidence re-read at that ref and confirmed present (finding 1) / absent (finding 2, ESRCH grep = 0 matches).
- **Not run:** the test suite and `tsc --noEmit` were not independently executed (PR claims `bash.test.ts` 54/54 and tsc clean — relied on the PR's stated verification plus static reading; test-count delta 53→54 independently confirmed via git).
- Did not trace production `session.interrupt()` end-to-end into its concrete `AbortController`; assumed the native/spec single-fire contract.
- Diff is confined to `bash.ts` / `bash.test.ts`; no other surfaces reviewed.
- **Stated intent:** PR #349 title + body — spec-compliance **assessed** (all clauses met, no scope creep).
- No critical/high findings → `/shadow-verify` not required per skill rule.

griffinwork40:
## Review: PR #349 — fix(bash): close abort TOCTOU so an in-flight child is killed, not orphaned

**Target:** `src/agent/tools/handlers/bash.ts` (+9/-0), `src/agent/tools/handlers/bash.test.ts` (+27/-0) · **Regime:** light (hotfix, 36 lines) · **Reviewed ref:** `31822b4ae6060a88c239680a309b5fe156fb21ec` (branch HEAD) · **Base:** main
**Stated intent:** PR #349 title + body (captured via `gh pr view 349`) — spec-compliance assessed against it, not against the diff's own framing.

### Findings

- **medium · high confidence · correctness · `bash.ts:279-281` · ref:31822b4 · file-state** — The TOCTOU window this fix targets is not reachable via a genuine `AbortSignal` in the current call structure; the guard is defensive-only, not a fix for a live, reproducible race.
  Evidence: `bash.ts` has zero `await`/`.then`/`nextTick`/`setImmediate` between the line-106 pre-flight check and the line-272 `addEventListener` call (confirmed by direct grep + read at the reviewed ref) — the entire span, including `spawn()`, executes as one synchronous block. `AbortGraph.abort()` (`abort-graph.ts:162-208`) computes its BFS descendant list and calls every `controller.abort()` synchronously, in-stack. A native `AbortSignal` dispatches its `'abort'` event to listeners synchronously (WHATWG/Node semantics). Given all three, an abort from anywhere else in the system lands either fully before the handler's first check (already caught, returns early) or fully after the listener is registered (already caught normally) — never inside the described gap. `agent-session.ts:714-718`'s `interrupt()` is `async`/`await`, but that's upstream of this span and doesn't reopen it under JS run-to-completion. The new test (`bash.test.ts:260-265`) reproduces the branch only via a hand-rolled signal whose `addEventListener` itself flips `aborted = true` as a side effect — not how any real `AbortSignal` behaves.
  Suggestion: keep the guard — it's cheap, correct, idempotent (`settle()` guards on `resolved`), and standard forward-compatible defensive practice for `AbortSignal` consumers regardless of current reachability. Correct the PR/comment narrative: this closes a theoretical window, not the demonstrated cause of "the stray fragment observed in the field" — if that field symptom is real, its root cause is very likely elsewhere.
- **low · high · spec-compliance · `bash.test.ts:262-265` · ref:31822b4 · diff-context** — The regression test's docstring claims it "reproduces that exact race," but it only proves the defensive branch executes against a synthetic non-native signal.
  Suggestion: rename/reword to something like "exercises the post-registration recheck when aborted flips during registration" and drop the "reproduces that exact race" claim.
- **nit · high · correctness · `grep.ts:258` · ref:31822b4 · file-state** — `grep.ts` has the structurally identical pre-flight/spawn/register shape with no equivalent recheck. Not a hidden gap — the PR explicitly discloses "Confined to bash.ts" under Risk.
  Suggestion: follow-up item to apply the same idiom to `grep.ts` (and check `web-scrape.ts`'s multi-point recheck pattern as a possible model) for consistency.
- **nit · high · test-coverage · `bash.test.ts` (54 `it(` blocks) · ref:31822b4 · diff-context** — `it()` count matches the PR's claimed "54/54 incl. 1 new." Pass/fail was **not** independently verified — the suite was not run (read-only review constraint).
- **security** — no issues found — read `bash.ts` at `31822b4`.
- **api-compat** — no issues found — read `bash.ts` at `31822b4`; handler signature unchanged, reachable from real production call sites (`anthropic-direct/index.ts:370`, `openai-compatible/index.ts:365` via `handlers/index.ts:59-60,91`).
- **perf-observability** — no issues found — guard is O(1), fires at most once per call, no perf impact.

**Spec-compliance verdict:** unmet intent — **no** (`bash.ts:279-281` implements exactly the idempotent post-registration recheck the PR body specifies). Scope creep — **no** (+9 prod/+27 test, confined entirely to the abort path).

**Merge decision: MERGE** — highest surviving finding is `medium`/correctness and does not defeat the stated purpose or contract: the code correctly implements what the PR describes, is safe and idempotent, and is a net-positive defensive addition. The finding only disputes the PR's *root-cause narrative* (whether this specific race is reachable today), which is an accuracy/documentation concern, not a functional defect.

**What was not checked:**
- Test suite was not executed — this review is prompt/logic-level only (read-only constraint), not a runtime confirmation that all 54 `bash.test.ts` tests pass.
- Spawn-based handlers beyond `grep.ts` (e.g. any others sharing this shape) were not exhaustively audited.
- OS-level process-group reaping was not asserted at runtime (the new test checks response content/timing, not the process table).
- Stated intent was available (PR #349 title + body) and spec-compliance was assessed against it, not against the diff's own framing or generic repo conventions.
- No critical/high finding surfaced, so the skill's mandatory post-synthesis `/shadow-verify` gate was not triggered; the pivotal medium finding was nonetheless independently re-derived three times from scratch (by me before dispatch, by the Wave-1 review agent, and by the Wave-1.5 verification agent against live file reads at the reviewed ref) and converged on the same evidence chain each time.

**Done**
- What was done: read-only `/review` of PR #349 completed — diff read from the supplied artifact, PR intent captured via `gh pr view` (read-only), a 2-stage `compose` verification pipeline (Wave 1 light-review agent → corrected-and-rerun Wave 1.5 citation/absence verification → Wave 2 synthesis, this time with explicit edges after an initial mis-wired parallel dispatch surfaced the gap correctly rather than fabricating) plus my own independent file reads cross-checked every citation.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/ba999836-bb01-495c-922b-9469b7c63a32/pr-349.diff`; full Wave 1.5 output at `/Users/griffinlong/.afk/state/sessions/ba999836-bb01-495c-922b-9469b7c63a32/compose/toolu_01V1nHQh63v6D88STmH26eP6/wave1_5_verification.txt`; all file:line citations above verified against reviewed ref `31822b4`.
- What changed in the world: nothing — no edits, commits, pushes, or PR mutations were made, per the skill's hard read-only constraint. The dirty working-tree file (untracked `node_modules`) was left untouched.
- Pending/deferred: none — merge recommendation (**MERGE**) delivered; the two `nit`/`low` findings and the narrative-accuracy note are follow-ups for the author's discretion, not blockers.

---

PR #351, opened 2026-07-01
fix(tui): correct stale commit geometry

griffinwork40:
## Summary
- correct commitAbove geometry when CupFrameRenderer shrink-padding leaves logUpdate.topRow artificially low
- use committedBandBottomRow + 1 as a floor when a positioned committed band exists, so decideCommitMode sees the real frame top
- add an xterm/headless regression that preserves prior prose and table rows after overlay shrink + subsequent commit

## Tests
- pnpm exec vitest run src/cli/terminal-compositor.commit-geometry.test.ts src/cli/terminal-compositor.overflow-gap.test.ts src/cli/terminal-compositor.endturn-overflow-gap.repro.test.ts src/cli/commit-mode.test.ts
- pnpm lint
- pnpm test

## Notes
- Full suite passed locally: 10,446 tests passed, 14 skipped.
- Worktree still has untracked node_modules from local install; not committed.

### DIFF
--- a/src/cli/terminal-compositor.commit-geometry.test.ts
+++ b/src/cli/terminal-compositor.commit-geometry.test.ts
@@ -1,0 +1,253 @@
+/**
+ * Regression: stale `prevTopRow` (from shrink-padded `logUpdate.topRow`) causes
+ * `decideCommitMode` to compute wrong geometry, routing a commit to `useBandHold`
+ * when the block fits the real above-frame room — permanently losing prior committed
+ * content that was repositioned by `repositionCommittedBand`.
+ *
+ * ROOT CAUSE (see docs/scrollback.md §407-424):
+ *   `commitAbove` captures `prevTopRow = self.logUpdate.topRow` BEFORE Phase-2
+ *   `repaint()`.  When `CupFrameRenderer.render()` applies shrink-padding (frame
+ *   shrank since the previous render), `logUpdate.topRow` equals the PADDED top row
+ *   — lower than the real, unpadded frame top that Phase-2's repaint will establish.
+ *   `measure(currentFrame, absoluteBottom).topRow` returns the unpadded top.
+ *
+ *   Concretely (ROWS=24, extraRows=2, absoluteBottom=21, anchorFloor=1, COLS=120,
+ *   spinner ON, **no tip**, anchorRow=1):
+ *
+ *   1. Tall overlay (14 lines) → frame 17 rows → last render places `topRow` = 5.
+ *
+ *   2. `commitAbove("GEO_PROSE\n\n")` — 1-row block, `fitsAboveFrame` TRUE (1 ≤ 4).
+ *      Phase-3 CUP-paints at rows 3..4.
+ *      `committedBand = ["GEO_PROSE", ""]`, `committedBandBottomRow` = 4.
+ *
+ *   3. Overlay SHRINKS to 4 lines → `repaint()` applies shrink-padding:
+ *      `previousRawLineCount`=17, new raw=7, `shrinkPad`=10, `newTopRow`=5 — same!
+ *      `logUpdate.topRow` = **5** (STALE).
+ *      `measure(4-line overlay frame, 21).topRow` = **15** (REAL).
+ *      `repositionCommittedBand` runs: repins band from rows 3..4 → rows 13..14
+ *      (adjacent to the real frame top at 15).  `committedBandBottomRow` = 14.
+ *
+ *   4. `commitAbove("GEO_TABLE ...\n\n")` — 5-row block, **with stale prevTopRow**:
+ *      `fitsAboveFrame` = 5 > 1 && 5 ≤ stale_room=4 → **FALSE** (WRONG).
+ *      With `measure` topRow=15: 5 ≤ real_room=14 → **TRUE** (correct routing).
+ *      The wrong FALSE fires `useBandHold = TRUE`.
+ *
+ *      Band-hold picks `overflowPriorContiguous`: `committedBandBottomRow`(14)
+ *      === stale `frameTop`-1=4 → **FALSE** (stale frame top 5 doesn't match band at 14).
+ *      So `overflowRun` = just the new 5-row block — NO merge with prior prose.
+ *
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

# Code Review — PR #351

**`fix(tui): correct stale commit geometry`** · `main` ← `fix/tui-commit-geometry-measure`
**Reviewed ref:** `2c136f014c80836adf46cf44abe9d8a8d6c15459` · 2 files · +281/−3 · regime: light (hotfix)

## Verdict: ✅ MERGE

Correct, minimal, well-tested fix for a real content-loss defect. When `CupFrameRenderer` shrink-padding leaves `logUpdate.topRow` artificially low, `decideCommitMode` computed a wrong `fitsAboveFrame` / `overflowPriorContiguous`, misrouted a fitting commit into band-hold, and the band-hold paint overwrote the repositioned prior committed band **without archiving it to scrollback — permanent content loss**. The fix floors `prevTopRow` at `committedBandBottomRow + 1` (the band must sit above the frame, so the real frame top can never be lower) whenever a positioned committed band exists. Mechanism, fail-safety, and the regression all verify against source. Three low/nit, non-blocking findings — none gate merge.

## What I verified as correct (source-grounded, ref `2c136f01`)
- **The floor is sound and conservative.** `prevTopRow = committedBand.length > 0 && committedBandBottomRow > 0 ? Math.max(rawLogUpdateTopRow, committedBandBottomRow + 1) : rawLogUpdateTopRow` (`committed-band-commit.ts:191-193`). It only ever *raises* the value, and only when a band is positioned — it cannot regress the empty-band / unpositioned path (falls through to the raw value, byte-identical to the deleted line). ✓
- **The correction now propagates to every consumer.** Because `frameTop = prevTopRow > 1 ? prevTopRow : …` (`:194`), the corrected top flows into `fitsAboveFrame` (`:213`) and into `decideCommitMode`'s `prevTopRow` / `frameTop` args (`:270-272`) — the exact inputs the test shows diverging (stale 5 vs. real 15). ✓
- **The `> 1` BLOCKER-1 guard is preserved.** `fitsAboveFrame` still requires `prevTopRow > 1` (`:213`), so the correction does not reopen the "no real frame top" hole documented at `:180-186`. ✓
- **Regression genuinely exercises the fixed path.** The new xterm/headless test drives the precise scenario (tall overlay → commit prose → shrink → repositionCommittedBand → second commit), asserts the stale precondition (`logUpdate.topRow` stays 5 while `committedBandBottomRow` moves to 14), and the RED assertion `GEO_PROSE … toBe(1)` fails pre-fix (0 hits) and passes post-fix. Plus viewport-visibility, commit-order, and no-blank-void checks. ✓ (`terminal-compositor.commit-geometry.test.ts:158-259`)

**1 · LOW · maintainability · `src/cli/terminal-compositor.committed-band-commit.ts:222` and `:354` · ref:2c136f01 · file-state**
The floor now applied to `prevTopRow` makes the two pre-existing `Math.max(frameTop, committedBandBottomRow + 1)` corrections (`phase1EffectiveFrameTop` at `:222`; Phase-1 `effectiveFrameTop` at `:354`) largely redundant: since `frameTop` is derived from the already-floored `prevTopRow`, `frameTop ≥ committedBandBottomRow + 1` holds whenever those guards fire, so both `Math.max` calls reduce to `frameTop`. Not a bug — they remain correct no-ops and defensively safe — but the geometry correction now lives in three places.
evidence: `? Math.max(frameTop, self.committedBandBottomRow + 1)` (identical at `:222` and `:354`)
suggestion: In a follow-up, add a one-line comment at `:222` / `:354` noting the value is now normally a no-op because `prevTopRow` is pre-floored, so a future reader does not assume it is the sole correction site. No code change needed for this PR.

**2 · NIT · test-coverage · `src/cli/terminal-compositor.commit-geometry.test.ts:158-259` · ref:2c136f01 · file-state**
The regression covers the primary divergence case (positioned band, stale-low top) thoroughly, but not the boundary where `rawLogUpdateTopRow` is already `≥ committedBandBottomRow + 1` (floor is a no-op) — i.e. proof the fix does not perturb the already-correct path. The existing green suite (`overflow-gap`, `endturn-overflow-gap.repro`, `commit-mode`) plausibly covers this indirectly.
evidence: `expect(internals.committedBandBottomRow, 'band must be repositioned to row 14 …').toBe(14);`
suggestion: Optional: one assertion (or a comment pointing at the sibling test) confirming the no-op path is unaffected. Non-blocking.

**3 · NIT · maintainability · PR body "Notes" · ref:2c136f01 · file-state**
The PR note "Worktree still has untracked node_modules from local install; not committed" is a local-hygiene aside with no bearing on the diff (the two changed files are the entire change). Harmless; flagging only so it is not mistaken for an intended artifact.
suggestion: None required.

## Spec-compliance
**Stated intent:** PR #351 title + body — assessed. All three stated goals are met by the diff: (a) correct `commitAbove` geometry when shrink-padding leaves `logUpdate.topRow` low → the `prevTopRow` floor (`:191-193`); (b) use `committedBandBottomRow + 1` as a floor when a positioned band exists → exactly the guard condition (`:192`); (c) add an xterm/headless regression preserving prose + table rows after overlay shrink + subsequent commit → `commit-geometry.test.ts`. No scope creep — the change is confined to the one geometry computation plus comments and the new test.

- Read against branch HEAD `2c136f014c80836adf46cf44abe9d8a8d6c15459` via `gh api …/contents/…?ref=<sha>` and `gh pr diff` — all citations verified at that ref; working tree not used.
- Regime: light (hotfix, ~28 production lines) → single-reviewer path; findings re-derived against the reviewed ref.
- Not run: `pnpm test` / `lint` / `build` were not executed; the PR claims 10,446 pass / 14 skip + lint clean — not independently verified. Logic and the regression verified by reading, not by running the terminal test.
- Not exhaustively modeled: the ANSI row-math and `decideCommitMode` / `repositionCommittedBand` interaction were validated via the test's documented trace and direct source reads, not by independent geometry simulation across all frame sizes.
- No critical/high findings → shadow-verify not required per skill rule.

---
_Read-only review — no files, commits, or merge state were modified. Verdict: MERGE._

PR #352, opened 2026-07-01
fix(telegram): keep the typing indicator alive for the whole turn

griffinwork40:
## What

The Telegram bot's "typing…" indicator went stale on almost every turn. `sendChatAction('typing')` was fired once per turn (fire-and-forget), and Telegram auto-expires that bubble after ~5s. So any turn longer than ~5s (essentially every agent turn: streaming, tool use, compaction) showed no live typing signal during exactly the phases where the streamed message can sit static and the user most wants reassurance it's still working.

## Change

- New `src/telegram/typing-indicator.ts` — `withTypingIndicator(ctx, work)` shows typing for the full duration of `work`, re-arming every 4s (under Telegram's ~5s expiry) and clearing the timer in `finally` on every exit path (resolve or throw).
- Converted all three one-shot `sendChatAction('typing')` call sites to the wrapper:
  - `handlers/message.ts` → `processOne` — the streamed turn (the main case)
  - `handlers/message.ts` → `processCompactDirect` — `/compact` (a model call that also exceeds ~5s)
  - `handlers/commands.ts` → `handleCompact` — `/compact`

## Why a custom helper instead of Telegraf's `persistentChatAction`

Telegraf 4.16 ships `ctx.persistentChatAction`, but it does not guard its *initial* `sendChatAction`. A transient typing-only failure (flood-control 429, blocked chat, network blip) would throw **before** the wrapped work ran, silently dropping the user's reply. The helper swallows every send error — initial and refresh — preserving the pre-existing best-effort `.catch(() => {})` semantics: a failed indicator must never cost the user their reply. As a bonus, because it still calls `ctx.sendChatAction('typing')` internally, no existing test mocks or assertions needed changing.

## Testing

- New `typing-indicator.test.ts` (6 tests, fake timers): immediate send, 4s re-arm while pending, timer teardown on both resolve and throw, best-effort error swallowing, return-value passthrough.
- Full `src/telegram` suite: **541 tests pass** (incl. `bot.test.ts` real-stream path, which now exercises the wrap end-to-end).

No `@anthropic-ai/sdk` imports added; the SDK dependency lock is unaffected.

### DIFF
--- a/src/telegram/handlers/commands.ts
+++ b/src/telegram/handlers/commands.ts
@@ -6,6 +6,7 @@ import { homedir } from 'os';
 import { isAbsolute, resolve } from 'path';
 import { HookBlockedError } from '../../utils/errors.js';
 import { SessionManager } from '../session-manager.js';
+import { withTypingIndicator } from '../typing-indicator.js';
 import {
   formatError,
   formatClear,
@@ -74,17 +75,20 @@ export async function handleCompact(
 
   try {
     const session = await sessionManager.getSession(chatId);
-    await ctx.sendChatAction('typing').catch(() => {});
-    // Invariant: fire PreCompact before compaction. block -> skip, not error.
     const hookRegistry = session.hookRegistry;
-    if (hookRegistry) {
-      await hookRegistry.dispatch({
-        event: 'PreCompact',
-        sessionId: session.sessionId,
-        trigger: 'manual',
-      });
-    }
-    const result = await session.compact();
+    // Keep the "typing…" indicator alive across the PreCompact hook and the
+    // model-call compaction, which can outlast the ~5s one-shot expiry.
+    // Invariant: fire PreCompact before compaction. block -> skip, not error.
+    const result = await withTypingIndicator(ctx, async () => {
+      if (hookRegistry) {
+        await hookRegistry.dispatch({
+          event: 'PreCompact',
+          sessionId: session.sessionId,
+          trigger: 'manual',
+        });
+      }
+      return session.compact();
+    });
     if (!result.compacted) {
       await ctx.reply(formatCompactNoop(result.reason ?? 'unknown'));
     } else {
--- a/src/telegram/handlers/message.ts
+++ b/src/telegram/handlers/message.ts
@@ -5,6 +5,7 @@ import { SessionManager } from '../session-manager.js';
 import { formatError, formatClear, formatInternalError, formatCompact, formatCompactNoop, formatQueued, escapeHtml } from '../formatter.js';
 import { isRateLimitError, isNetworkError, isTelegramTransportError } from '../error-utils.js';
 import { streamResponse } from '../streaming.js';
+import { withTypingIndicator } from '../typing-indicator.js';
 // Import StreamTimeoutError from its own module, NOT '../streaming.js': many
 // handler tests vi.mock('../streaming.js'), which would make the class resolve
 // to undefined and turn `instanceof StreamTimeoutError` into a TypeError.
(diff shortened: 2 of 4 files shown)


griffinwork40:
Both low findings verified, with important context:

- **Finding 1 (handleCompact test gap):** Absence confirmed (`git grep handleCompact b2df8b4 -- '*.test.ts'` → 0). But `handleCompact` is production-reachable (`bot.ts:119`), and critically — this is a **pre-existing** gap (no dedicated test existed before this PR either); the conversion is mechanical. The *primary* call site (`processOne`, the streamed turn) **is** covered end-to-end by `bot.test.ts:188`.
- **Finding 2 (silent error swallowing):** The `drainQueue` logging precedent the agent cited applies to control-flow errors. The *closer* precedent — `ctx.react?.('👀').catch(()=>{})` (msg.ts:181,413) and `ctx.reply(...).catch(()=>{})` (538,556) — swallows silently. Silent-swallow for cosmetic UI affordances is the **established convention in this exact file**; typing is cosmetic like a reaction. This recontextualizes the finding down to a nit.

No critical/high findings → `/shadow-verify` not required. Emitting synthesis + merge decision.

---

# Code Review — PR #352

**`fix(telegram): keep the typing indicator alive for the whole turn`**
`afk/research-telegram-typing` → `main` · reviewed ref `b2df8b4` · +159/−29, 4 files · regime: light (hotfix)

## Verdict: ✅ MERGE

A clean, additive, well-tested fix. Async/timer semantics are correct, the stated intent is fully met with no scope creep, and there are no security, api-compat, or correctness defects. Two **low/nit** findings remain, neither blocking.

## Findings by dimension

| Dim | Result |
|-----|--------|
| security | ✅ no issues — no secrets/injection surface; literal `'typing'`, typed `ctx` |
| api-compat | ✅ no issues — `withTypingIndicator` is a new additive export; no existing signature altered; call shape unchanged |
| correctness | ✅ no issues — verified below |
| spec-compliance | ✅ intent met — 3/3 one-shot sites converted; `persistentChatAction` rationale accurate; no scope creep |
| test-coverage | ⚠ 1 low (pre-existing) |
| perf-observability | ⚠ 1 nit |

**Correctness — verified at ref b2df8b4** (`typing-indicator.ts:24-37`):
- `ctx.sendChatAction('typing').catch(() => {})` — `.catch` is chained synchronously in the same statement, so no unhandled-rejection window despite being unawaited.
- `clearInterval(timer)` sits in `finally` → runs on both the resolve path (`return await work()`) and any throw; the original rejection re-propagates unchanged. No interval outlives the turn.
- `ReturnType<typeof setInterval>` matches the existing strict-tsconfig idiom used elsewhere in the tree (resolves to `NodeJS.Timeout`).
- Behavior change (new: initial send **not** awaited before `work` starts) is explicitly disclosed in the PR body and is safe — the first typing send is best-effort and gates nothing.

### Low / nit (non-blocking)

`low · high · test-coverage · src/telegram/handlers/commands.ts:65 · ref:b2df8b4 · grep-verified-absence`
`handleCompact` has no dedicated test exercising it through `withTypingIndicator` (`git grep handleCompact b2df8b4 -- '*.test.ts'` → 0 matches; it is production-reachable via `bot.ts:119`). **Context:** pre-existing gap not introduced by this PR; the conversion is mechanical; the helper itself has 6 fake-timer tests and the primary streamed-turn site (`processOne`) is covered end-to-end by `bot.test.ts:188`. *Suggestion:* add a `commands.test.ts` asserting PreCompact + `session.compact()` still fire inside the wrapped call.

`nit · high · perf-observability · src/telegram/typing-indicator.ts:28,30 · ref:b2df8b4 · direct-read`
Every `sendChatAction` failure is swallowed with no logging, so a persistent per-turn failure (e.g. bot blocked) recurs silently. **Context:** this matches the file's own best-effort-UI convention — `ctx.react?.('👀').catch(()=>{})` (msg.ts:181,413) and `ctx.reply(...).catch(()=>{})` (538,556) also swallow silently; only substantive control-flow errors like `drainQueue` log. Typing is a cosmetic affordance, so silent-swallow is a defensible, consistent choice. *Optional:* thread an optional debug logger if diagnosing flood-control is later desired.

## What was not checked
- Read against branch HEAD `b2df8b4` — all citations verified at that ref. Working tree is on `main` (does not contain these changes); grep/reads targeted the ref explicitly.
- Did **not** run `pnpm test` / `tsc` (read-only constraint). The PR's "541 tests pass, lint clean" claim is corroborated by static reasoning (mocks still see `ctx.sendChatAction('typing')`, so `bot.test.ts` assertions hold) but not executed here.
- Stated intent: PR #352 title + body — spec-compliance **assessed**.
- Did not review Telegram flood-control/rate-limit behavior empirically (4s cadence over multi-minute turns is disclosed and matches Telegraf's own `persistentChatAction` default).

---

**Done**
- Reviewed PR #352 read-only across all six dimensions; emitted merge decision **MERGE** with two non-blocking low/nit findings.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/1cd13007-d8fb-4e20-9a18-5ea3e47a6fe8/pr-352.diff`; all citations verified against reviewed ref `b2df8b4fdfae5df8a173fa9c63a1fb1e896946a3` (Wave 1 research agent + inline backstop greps in this transcript).
- World unchanged — this skill is read-only: no files edited, no commits, no PR comments/merge. Your dirty working tree on `main` was not touched.
- Note: Wave 1's first dispatch was terminated (transient); I retried successfully with a leaner prompt, so the review rests on a completed independent agent pass plus my own ground-truth verification — not an inlined fallback.

---

PR #372, opened 2026-07-02
feat(repl): auto-deliver background subagent results into the next turn

griffinwork40:
## Summary
- Background subagent results (`agent` tool `mode:"background"` or Ctrl+B promotion) are now **auto-delivered into the model's context with the next user message** — no more manual `/bgsub:join` required for the model to learn its own background work finished.
- New `BgResultNotifier` subscribes to `BackgroundAgentRegistry`'s `settled` event and buffers, per job: a `<background-subagent-result>` envelope (XML-escaped, 16KB post-escape cap with a `/bgsub:join` pointer on truncation) plus a one-line human notice rendered at the top of the next prompt iteration.
- Mirrors the shell-passthrough (`!&cmd`) delivery contract exactly: next-turn prepend, never a mid-turn push (the provider consumes one input-stream message per turn — a live push would desync the stream renderer, cf. 51c46d8).
- New `background_agent.delivered` witness transition so trace readers can distinguish auto-delivery from explicit joins; `/bgsub:join` remains available for manual replay (delivery does not consume the job). Opt-out: `AFK_BG_AUTO_DELIVER=0`.

## Test results
- `pnpm exec vitest run` (full suite): **567 files / 10,481 passed / 0 failed / 14 skipped**
- `pnpm lint` (tsc --noEmit, maximally strict): clean
- `pnpm scan:env` + `pnpm audit:env:check`: clean (new env var registered)
- New coverage: 22 `BgResultNotifier` unit tests (envelope shape, escaping, post-escape cap, failed/cancelled paths, opt-out, dispose, reset) + an end-to-end REPL loop test proving the envelope lands in the second `runTurn`'s text.

## Verification
Adversarial verifier re-derived the three load-bearing claims from the diff alone:
- **Next-turn delivery, non-consuming** — CONFIRMED (`loop-iteration.ts` drains before `runTurn`; `markDelivered` never mutates job state; `join()` unaffected).
- **Escape + 16KB cap** — REFINED: original ordering was truncate-then-escape, letting adversarial `<`-heavy output expand 4× past the cap. **Fixed in 063f519** (escape-then-truncate; regression test with 16KB of `<` added).
- **Teardown safety** — REFINED: `/resume` swap cancelled jobs but left settled results buffered in the notifier, leaking into the resumed session's first turn. **Fixed in 063f519** (`reset()` wired through `ctx.clearBgResultBuffer` in the swap's `onSwapped`, mirroring `clearVerdictLedger`). Cancelled jobs remain notice-only (no injection) regardless of dispose timing.

## Out of scope
- Mid-idle push-wakeup (model waking spontaneously when a job settles while the REPL sits at the prompt) — deferred; next-turn delivery only.
- Non-REPL surfaces (daemon/chat/telegram) — no `BackgroundAgentRegistry` is wired there today.


griffinwork40 (commented):

# Code Review — PR #372

**`feat(repl): auto-deliver background subagent results into the next turn`** · `main` ← `afk/notify-subagent-completion`
**Reviewed ref:** `e4d6f15e7135fb05ceb15c39dcf14cceb68c4b1f` · 20 code/test files (+ docs) · +652/−28 core (3 squashed commits) · regime: **full** (2 parallel dimension agents + citation/absence verification)
**Stated intent:** PR #372 title + body — spec-compliance **assessed**.

## Verdict: ✅ MERGE

A well-scoped, well-tested feature that does exactly what its intent states. Background-subagent results now auto-deliver into the model's context on the **next** user message (interactive REPL only), mirroring the shell-passthrough (`!&cmd`) next-turn-prepend contract — never a mid-turn push. **No critical, high, or blocking findings** across security, api-compat, correctness, spec-compliance, test-coverage, or perf-observability. The findings below are one medium (a bounded worst-case aggregate size) and a cluster of low/nit test-fidelity follow-ups — none gate merge.

## What I verified as correct (source-grounded at ref `e4d6f15e`, cross-checked by an independent read)

- **Envelope-injection is escape-safe (security).** `buildBgResultInjection` runs `truncateBytes(escapeXml(extractOutput(job)), MAX_INJECTION_BYTES, job.jobId)` — escape happens **before** truncate (`bg-result-notifier.ts:127`), so the 16KB cap bounds the **post-escape** size. `escapeXml` covers `&`/`<`/`>`, and the `<task>` (`job.label`) and `model` attribute are both escaped, so crafted subagent output cannot emit a spurious `</output>` / `</background-subagent-result>` to break the envelope framing. The patch-2 fix (escape-then-truncate) is genuinely in the shipped tree — I re-read it at the ref, not just the diff. ✓
- **Delivery is non-consuming (correctness).** `markDelivered` (`background-registry.ts:420-429`) only reads `this.jobs.get(jobId)` and emits the witness event — it never touches `job.result` or deletes the entry, so `/bgsub:join` still resolves the same job. `drainInjections()` clears `pendingInjections` before mapping. ✓
- **Cancelled jobs are notice-only.** `onSettled` pushes a notification then returns before the injection push when `status === 'cancelled'` (`bg-result-notifier.ts:161-174`). ✓
- **Lifecycle wiring is sound.** `dispose()` is called in `repl-loop.ts`'s `finally` before the registry's `cancelAll`; the `/resume` swap path routes `ctx.clearBgResultBuffer?.()` (`bootstrap.ts:708`, optional) → `bgResultNotifier.reset()` (`footer-subsystems.ts:200`), and `reset()` empties **both** buffers — mirroring the `clearVerdictLedger` pattern the intent cites. ✓
- **api-compat clean.** All five symbols dropped from `anthropic-direct`'s import in patch 3 (`TOOL_SYSTEM_PROMPT_BASE`, `SLASH_COMMAND_ROUTING_PROMPT`, `BASH_PASSTHROUGH_PROMPT`, `MEMORY_SYSTEM_PROMPT`, `MEMORY_SYSTEM_PROMPT_READONLY`) remain **exported** from `system-prompt.ts` with their other production importers intact — no export was removed. `InteractiveCtx.clearBgResultBuffer` is added **optional** (`?`), so existing ctx producers are unaffected. The rewritten `openai-compatible` assembly (`[toolBase, memoryPrompt, envFragment, manifest?, existingSys?]`) appends the user's `systemPrompt` exactly once (no drop, no duplication). ✓
- **Spec-compliance: all 8 stated requirements implemented** — next-turn auto-deliver, shell-passthrough-mirrored contract, escaped 16KB-post-escape-capped envelope with `/bgsub:join` pointer, one-line human notice, the `background_agent.delivered` witness transition wired end-to-end (`types.ts` + `events.ts` zod discriminated union + `trace.ts` renderer + `registry.markDelivered` emit), non-consuming `/bgsub:join`, `AFK_BG_AUTO_DELIVER=0` opt-out, and the patch-3 shared-resolver fix delivering the prompt on **both** providers. **No scope creep** — `openai-compatible` gaining memory + manifest fragments is called out in the intent itself as the H1 root cause (that provider previously assembled *no* provider-side fragments), so it is in-scope. ✓

## Findings (none blocking)

**1 · MEDIUM · confidence:high · perf-observability · `src/cli/commands/interactive/bg-result-notifier.ts:47,164-173,189-192` · ref:e4d6f15e · citation-type:file-state**
The 16KB `MAX_INJECTION_BYTES` cap bounds **each job's** output, but `drainInjections()` concatenates up to `MAX_PENDING` (25) settled jobs into one prepended block, so a worst-case burst of parallel background jobs settling before one turn can inject ≈25 × 16KB ≈ 400KB ahead of the user's message. Bounded (not unbounded DoS), but large enough to meaningfully crowd the next turn's context window on a wide fan-out.
`evidence:` `return jobs.map((j) => buildBgResultInjection(j)).join('\n') + '\n';` (per-job cap only) with `MAX_PENDING = 25`
`suggestion:` Apply an aggregate byte budget across the drained set (truncate/least-recent-drop once the concatenated size crosses a total cap), or lower `MAX_PENDING` for the injection buffer specifically. Non-blocking at typical job volumes.

**2 · LOW · confidence:medium · test-coverage · `src/cli/commands/interactive/bg-result-notifier.test.ts` · ref:e4d6f15e · citation-type:file-state**
Non-consumption is asserted only by spying on `markDelivered` + static reading; no test calls `registry.join()` *after* `drainInjections()` to prove end-to-end that the job is still retrievable. (Grounded: grep of the two relevant test files for `.join(` matches only the `/bgsub:join` string literal, never an actual `join()` call post-drain.)
`suggestion:` Add a case that drains, then `await registry.join(jobId)` and asserts the full result still resolves.

**3 · LOW · confidence:medium · test-coverage · `src/agent/tools/system-prompt.test.ts` / provider `index.ts` · ref:e4d6f15e · citation-type:file-state**
The H1 regression is guarded at the **resolver** level (`resolveToolSystemPrompt` returns a string containing `<background-subagent-result>`), but no **provider-level** test asserts the actually-assembled system message contains the fragment for either provider — which is the exact drift class H1 was (a call site hand-rolling its own concat and falling behind). A future bypass at a provider call site would not be caught. (Grounded: code-search for `BG_SUBAGENT_RESULT_PROMPT` returns zero hits outside `system-prompt.ts`/`system-prompt.test.ts`.)
`suggestion:` Add a thin provider test asserting the built system prompt string contains `<background-subagent-result>` for both `anthropic-direct` and `openai-compatible`.

**4 · LOW · confidence:medium · test-coverage · `src/cli/commands/interactive/loop-iteration.test.ts` · ref:e4d6f15e · citation-type:file-state**
No test exercises `shellPassthrough` and `bgResultNotifier` injections pending in the **same** iteration, so the prepend ordering when both fire is unverified by test (verified correct by reading: both use `runText = X + runText`, yielding `[bg][shell][userText]` with no interleaving). (Grounded: grep of the test file finds no combined scenario.)
`suggestion:` Add one iteration test with both a settled shell job and a settled bg job, asserting both blocks precede the user text.

**5 · NIT · confidence:high · test-coverage · `src/cli/commands/trace.ts:311-312` · ref:e4d6f15e · citation-type:diff-context**
The new `case 'delivered'` trace-render branch has no test — consistent with the pre-existing untested `joined` branch, so this is not a regression this PR introduces. (Grounded: grep of `trace.test.ts` for `delivered`/`bg-agent` finds no match.)
`suggestion:` Optional — add a render-branch assertion alongside a future `joined`-branch test.

- Read against branch HEAD `e4d6f15e7135fb05ceb15c39dcf14cceb68c4b1f` via `gh pr diff` + `gh api …/contents/…?ref=<sha>` (no local checkout of this branch). All `file-state` citations were re-read at that ref; the load-bearing escape-order security claim was independently re-derived by the coordinator at the ref. No critical/high finding surfaced, so `/shadow-verify` was not triggered (medium/low pass straight through).
- **Not run:** `pnpm lint` / `pnpm test` were not executed (read-only mandate); the PR's "567 files / 10,481 passed / 0 failed" claim was verified only by reading the new test bodies, not a green run. CI on this ref is green (Lint & Build, Test, Docs, Bundle Smoke all SUCCESS).
- One lifecycle ordering guarantee — `dispose()` running strictly *before* the registry's `cancelAll` across two independently-scheduled async exit paths (`runReplLoop` finally vs. the `rl.on('close')` cleanup) — could not be established from static reading. It is unchanged pre-existing exit architecture (shared with `shellPassthrough.drainOnExit()`), not new in this diff, so it is flagged as unverified rather than a defect.
- Out-of-scope by the intent's own text and not investigated: mid-idle push-wakeup, and non-REPL surfaces (daemon/chat/telegram — no `BackgroundAgentRegistry` is wired there today).

---

PR #373, opened 2026-07-02
feat(tui): grounded progress banner — real activity instead of mechanism noise

griffinwork40:
## Summary
- **Grounded "now line"**: the progress banner's detail row now shows the model's in-flight thinking clause (via `extractLatestThinkingClause` over `ThinkingLane.peekPhase()` — a sealed phase can never leak a stale clause), falling back to the tool-derived summary. Wired at both render sites (orchestrator fallback compose + lifecycle overlay slot).
- **Semantic progress at source**: both providers now emit `round N: <tool> <headline-arg>` (via the shared `summarizeToolInput`) and a stable `Working` description, replacing the information-free `Tool-use loop` / `Iteration N: used bash`.
- **Stale-banner eviction**: `lastProgressByTask` is cleared in `finalizeOrchestrator` (+ defensively in `dispose`). Previously nothing evicted it at loop exit, and since `OverlayComposer.flush()` redraws every slot on any `markDirty`, the frozen banner provably repainted on every post-loop prose chunk as if live.
- **Sanitization hardening**: every LLM-sourced banner field (`description`, `summary`, activity clause, tool name, subagent completion label) now passes `sanitizeLabel` — the banner previously interpolated raw model text with only width-clamping (ANSI/control-injection gap, pre-existing).

Plan was pressure-tested before implementation (devils-advocate critique wave + independent shadow verification of the load-bearing claims). Deliberately **not** done: rewriting the spinner-verb system (unsanitized path; codebase deliberately avoids raw thinking in scrollback per stream-renderer-subagent.ts:269) and consecutive-tool collapse (grouping already exists twice — `GROUP_THRESHOLD_*` live overlay + `formatGroupedToolResults` scrollback; the remaining gap is the per-flush-batch window, tracked as follow-up).

## Test results
- `pnpm lint` (tsc strict): clean
- `pnpm test` (full suite): 566 files, 10,470 passed | 14 skipped, 0 failed
- New coverage: sanitization + activity-precedence + `deriveProgressActivity` unit tests (progress-banner.test.ts), finalize-clears-progress regression test (stream-renderer-orchestrator.test.ts)

## Verification
Adversarial verifier re-derived the claims from the diff alone:
- Activity derivation uses `peekPhase()` consistently at both render sites — CONFIRM
- `clear()` placements + the single production `OrchestratorCtx` construction site (`stream-renderer.ts:507`) supplies the field — CONFIRM
- Sanitization claim — REFINE: `formatSubagentCompletion`'s `agentType` label was still raw; fixed in ceb3d47
- Consumer sweep: Telegram's progress lane renders `description`/`summary` as plain text (no terminal escape risk); string change there is cosmetic only

## Out of scope
- Cross-flush tool grouping window (confirmed real gap, tool-lane-render-agent.ts:33-37)
- First-class task-status model (`goal` field on `SubagentProgressMeta`; daemon/Telegram currently discard or reinterpret the stream)
- Triple "thought for Xs" line from the original report — diagnosis inconclusive (background agent hit env failure; `sealThinkingPhase` already guards same-phase re-commit, so the observed triple may be three legitimate phases). Needs a live repro.

### DIFF
--- a/src/agent/providers/anthropic-direct.test.ts
+++ b/src/agent/providers/anthropic-direct.test.ts
@@ -825,18 +825,20 @@ describe('AnthropicDirectProvider', () => {
 
     const first = progressEvents[0]!;
     expect(first.progress.taskId).toBeTruthy();
-    expect(first.progress.description).toBe('Tool-use loop');
+    expect(first.progress.description).toBe('Working');
     expect(first.progress.toolUses).toBe(1);
     expect(first.progress.lastToolName).toBe('read_file');
     expect(first.progress.totalTokens).toBeGreaterThanOrEqual(0);
     expect(first.progress.durationMs).toBeGreaterThanOrEqual(0);
-    expect(first.progress.summary).toContain('Iteration 1');
+    expect(first.progress.summary).toContain('round 1');
+    expect(first.progress.summary).toContain('read_file');
     expect(first.sessionId).toBeTruthy();
 
     const second = progressEvents[1]!;
     expect(second.progress.toolUses).toBe(2);
     expect(second.progress.lastToolName).toBe('write_file');
-    expect(second.progress.summary).toContain('Iteration 2');
+    expect(second.progress.summary).toContain('round 2');
+    expect(second.progress.summary).toContain('write_file');
     expect(second.progress.taskId).toBe(first.progress.taskId);
   });
 
--- a/src/agent/providers/anthropic-direct/loop.ts
+++ b/src/agent/providers/anthropic-direct/loop.ts
@@ -695,12 +695,21 @@ export async function* runTurn(
     iterations += 1;
 
     const lastTool = turnResult.toolUseBlocks[turnResult.toolUseBlocks.length - 1];
+    // Semantic summary: name the tool AND its most informative argument
+    // (path / command / query via summarizeToolInput) so the progress banner
+    // carries real signal — `bash git show f7f0a37…` instead of the old
+    // `Iteration 11: used bash`, which conveyed nothing after round 2.
+    // `description` stays STABLE across ticks for this taskId (the banner
+    // dedupes line 0 on commit); the per-tick signal lives in `summary`.
+    const lastToolHeadline = lastTool
+      ? `${lastTool.name}${summarizeToolInput(lastTool.name, lastTool.input)}`
+      : 'unknown';
     yield {
       type: 'progress',
       progress: {
         taskId,
-        description: 'Tool-use loop',
-        summary: `Iteration ${iterations}: used ${lastTool?.name ?? 'unknown'}`,
+        description: 'Working',
+        summary: `round ${iterations}: ${lastToolHeadline}`,
         lastToolName: lastTool?.name,
         totalTokens: accumulatedUsage.totalTokens ?? 0,
         toolUses: iterations,
--- a/src/agent/providers/openai-compatible/query.ts
+++ b/src/agent/providers/openai-compatible/query.ts
@@ -594,13 +594,29 @@ export class OpenAICompatibleQuery implements ProviderQuery {
       yield* this.dispatchAndAppend(result.state, controller.signal, vision);
 
       {
-        const lastToolName = finalizedToolCalls(result.state).at(-1)?.name;
+        const lastCall = finalizedToolCalls(result.state).at(-1);
+        const lastToolName = lastCall?.name;
+        // Semantic summary — mirror anthropic-direct/loop.ts: tool name +
+        // most informative argument via summarizeToolInput, so the progress
+        // banner carries real signal instead of a bare iteration counter.
+        // AccumulatedToolCall carries UNPARSED argumentsRaw (the streamed
+        // JSON fragments joined); parse best-effort — a malformed payload
+        // (mid-stream abort, shim quirks) degrades to the bare tool name.
+        let lastCallInput: unknown;
+        try {
+          lastCallInput = lastCall ? JSON.parse(lastCall.argumentsRaw || '{}') : undefined;
+        } catch {
+          lastCallInput = undefined;
+        }
+        const lastToolHeadline = lastCall
+          ? `${lastCall.name}${summarizeToolInput(lastCall.name, lastCallInput)}`
+          : 'unknown';
         yield {
           type: 'progress',
           progress: {
             taskId,
-            description: 'Tool-use loop',
-            summary: `Iteration ${iter + 1}: used ${lastToolName ?? 'unknown'}`,
+            description: 'Working',
+            summary: `round ${iter + 1}: ${lastToolHeadline}`,
             lastToolName,
             totalTokens: accumulatedUsage.totalTokens ?? 0,
             toolUses: iter + 1,
--- a/src/cli/_lib/stream-renderer-lifecycle.ts
+++ b/src/cli/_lib/stream-renderer-lifecycle.ts
@@ -14,7 +14,7 @@ import { OverlayComposer } from './overlay-composer.js';
 import { createStageTracker } from '../commands/interactive/loop-stage.js';
 import { formatDuration } from '../format-utils.js';
 import { formatThinkingParagraph } from '../commands/interactive/thinking-paragraph.js';
-import { formatProgressBanner } from '../commands/interactive/progress-banner.js';
+import { deriveProgressActivity, formatProgressBanner } from '../commands/interactive/progress-banner.js';
 import { palette } from '../palette.js';
 import { getTerminalWidth } from '../terminal-size.js';
 import { isDebugEnabled } from '../../utils/debug.js';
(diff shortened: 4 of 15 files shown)


griffinwork40 (commented):

# Code Review — PR #373

**`feat(tui): grounded progress banner — real activity instead of mechanism noise`** · `main` ← `afk/loading-state-feedback`
**Reviewed ref:** `ceb3d472a9bee30f68611d9273aa81b807208402` · 14 files (7 production) · +259/−26 core · regime: **full** (2 parallel dimension agents + citation/absence verification)
**Stated intent:** PR #373 title + body — spec-compliance **assessed**.

## Verdict: ✅ MERGE

A well-scoped, well-tested TUI change that does exactly what its intent states. All four stated changes are implemented and wired as claimed, and the load-bearing sanitization + eviction claims are verified against the ref. **No critical, high, or blocking findings** across security, api-compat, correctness, spec-compliance, test-coverage, or perf-observability. The findings below are one **medium** (a cross-surface sanitization gap that this diff *widens* but does not originate) plus low/nit follow-ups — none gate merge.

## What I verified as correct (source-grounded at ref `ceb3d472`, cross-checked by an independent verification pass)

- **CLI sanitization hardening is real (security).** `sanitizeLabel` neutralizes 7-bit CSI (`\x1b[2J`), 8-bit C1 CSI/OSC, and bare ESC — all four escape families the PR claims to close are actually closed in `formatProgressBanner` / `formatProgressSummary` / `formatSubagentCompletion`. Every LLM-sourced field feeding the **CLI overlay** — `description` (`progress-banner.ts:97,122`), the `detail` line = activity-or-summary (`:99`), `lastToolName` (`:88`), and the subagent `agentType` label (`:134`, added in the 2nd commit) — passes through `sanitizeLabel` before composition. ✓
- **The openai-compatible arg-parse is exception-safe (correctness).** `JSON.parse(lastCall.argumentsRaw || '{}')` is wrapped in try/catch → `undefined` on failure, and `summarizeToolInput` early-returns `''` for non-object/null/array input, so a malformed streamed payload degrades to the bare tool name with no throw and no prototype-pollution surface (values are read via bracket access, never merged). ✓
- **Stale-banner eviction produces a banner-free frame (correctness).** `ctx.lastProgressByTask.clear()` in `finalizeOrchestrator` (`stream-renderer-orchestrator-emit.ts:140`) runs synchronously *before* the overlay-refreshing commits scheduled below it, so their `flush()` paints a banner-free frame. The `dispose()` defensive clear (`stream-renderer.ts:715`) fires before `coordinator.flushAll()` and the tool-lane safety-net flush, so the property also holds on abort/interrupt paths that never see a `'done'` event. ✓
- **Activity precedence is correct.** `activity?.trim() ? activity : summary` falls back to `summary` on whitespace-only activity (pinned by test), and `deriveProgressActivity` returns `undefined` on an empty/whitespace buffer via its own guard plus `extractLatestThinkingClause`'s trimmed-empty return. ✓
- **api-compat clean.** `formatProgressBanner`'s new 3rd param `activity?: string` is **optional** — existing 2-arg call sites remain valid; `deriveProgressActivity` is a new export whose only production callers are added in this same diff; `lastProgressByTask` is a **pre-existing** ctx field (this PR adds only two `.clear()` call sites, not a signature/type change). None of the touched `cli/` files are reachable from the package's public `src/index.ts` export surface, so no external npm consumer is affected. ✓
- **Spec-compliance: all 4 stated changes implemented, no scope creep.** (1) `deriveProgressActivity(ctx.thinkingLane.peekPhase())` wired at **both** render sites (`stream-renderer-lifecycle.ts:130` + `stream-renderer-orchestrator.ts:414`); (2) both providers emit `round N: <tool><headline>` via `summarizeToolInput` + stable `Working` description; (3) `clear()` in `finalizeOrchestrator` + defensively in `dispose()`; (4) all 5 named LLM-sourced fields sanitized. No behavior change outside the four stated areas was found in the 7 production files. ✓
- **New tests are substantive, not tautological.** The sanitization tests assert on the **raw un-stripped** string (so they cannot pass vacuously), and the finalize-clears-progress regression genuinely asserts the map transitions 1→0 *and* that a post-clear overlay composition excludes the stale banner. ✓

## Findings (none blocking)

**1 · MEDIUM · confidence:medium · security · `src/telegram/streaming.ts:431-436` (source of the widened field: `anthropic-direct/loop.ts:705-712`, `openai-compatible/query.ts:611-619`) · ref:ceb3d472 · citation-type:file-state**
This diff widens `progress.summary` from a bare tool name to up to ~80 chars of **model-controlled** path/command/URL text (via `summarizeToolInput`). The CLI overlay now sanitizes that field — but the **Telegram** path does not: `streaming.ts` destructures `{ description, summary }` and interpolates both into the outgoing message, which flows through `markdownToTelegramHtml`, and that helper only strips `\x02`/`\x03` sentinels then HTML-escapes `&`/`<`/`>` — it applies **no** ANSI/C1/control-byte stripping, and neither `streaming.ts` nor `formatter.ts` imports `sanitizeLabel`/`stripAnsi`.
`evidence:` `if (summary) accumulated += \`\n  ${summary}\`;` (streaming.ts) + `markdownToTelegramHtml` escapes only `& < >` (formatter.ts)
`suggestion:` Apply `sanitizeLabel`/an escape-stripping pass to `description`/`summary` at the `streaming.ts` interpolation site, matching the CLI banner's hardening — moving sanitization from *CLI-overlay-scoped* to *field-scoped*.
*Why medium, not critical/high:* Telegram clients render HTML and do not interpret ANSI/CSI escapes, and `& < >` are already escaped (HTML injection is blocked), so there is no data-loss / auth / secret / RCE surface — the residual is control/bidi bytes reaching a message on a surface that does not execute them. The risk is also **partly pre-existing** (`description` already flowed there unsanitized); this PR only enlarges `summary`'s content. Hence a hardening gap, not a blocker. `[low confidence — Telegram client-side rendering impact of raw control/bidi bytes not tested]`

**2 · LOW · confidence:high · test-coverage · `src/agent/providers/openai-compatible/query.ts:604-609` · ref:ceb3d472 · citation-type:file-state**
The `argumentsRaw` parse-failure degradation branch has no test — grep of the openai-compatible query test file (2313 lines) for `argumentsRaw` / `lastCallInput` / `lastToolHeadline` / `JSON.parse` returns zero matches (independently confirmed at the ref); every scripted tool call uses well-formed `arguments`, so the catch/degrade path never executes under test.
`suggestion:` Add a scripted-turn test with a truncated/invalid `arguments` fragment asserting `progress.summary` degrades to the bare tool name without throwing.

**3 · LOW · confidence:medium · test-coverage · `src/cli/_lib/stream-renderer.ts:711-716` · ref:ceb3d472 · citation-type:file-state**
The `dispose()`-path defensive `lastProgressByTask.clear()` (the abort / no-`'done'` path) has no test — all 21 `lastProgressByTask` matches in `stream-renderer.test.ts` are `new Map()` scaffolding to satisfy the ctx type; none pair a progress event with a `dispose()` call (no intervening `'done'`/`'error'`) and assert the final overlay excludes the stale banner.
`suggestion:` Add a class-level test that fires a `progress` event, then calls `dispose()` without a `'done'` event, and asserts the final `setOverlay` excludes the stale banner.

**4 · NIT · confidence:medium · perf-observability · `src/cli/_lib/stream-renderer-lifecycle.ts:130` (mirrored `stream-renderer-orchestrator.ts:414`) · ref:ceb3d472 · citation-type:file-state**
`deriveProgressActivity` (a `getTerminalWidth()` call + a bounded regex scan) runs on every overlay-composer flush *before* the render loop, even when `ctx.lastProgressByTask` is empty and no banner will be drawn. The scan is bounded (windowed to ~400 chars) so the cost is small, but it is incurred unconditionally.
`suggestion:` Guard with `ctx.lastProgressByTask.size > 0 &&` before calling `deriveProgressActivity`.

**5 · NIT · confidence:high · spec-compliance · `src/cli/commands/interactive/progress-banner.ts:53-61` · ref:ceb3d472 · citation-type:file-state**
The docstring says `description` "stays STABLE across ticks for this taskId," but both providers now hardcode `description: 'Working'` for every task, so the invariant no longer conveys per-task meaning — only global constancy. Documentation-vs-implementation mismatch, not a functional bug.
`suggestion:` Reword the comment to state `description` is now a fixed literal, not a per-task-stable value.

- Read against branch HEAD `ceb3d472a9bee30f68611d9273aa81b807208402` via `gh pr diff` + `gh api …/contents/…?ref=<sha>` (no local checkout of this branch). All `file-state` citations were re-read at that ref; the load-bearing cross-surface Telegram sanitization claim (finding #1) and the two absence claims (#2, #3) were **independently re-derived** by a dedicated verification pass — CLAIM 1 CONFIRMED, both absence claims confirmed-absent.
- **Not run:** `pnpm lint` / `pnpm test` were not executed (read-only mandate); the PR's "566 files / 10,470 passed / 0 failed" claim was verified only by reading the new test bodies, not a green run. CI on this ref is green (Lint & Build, Test, Docs, Bundle Smoke all SUCCESS).
- No critical/high finding surfaced, so `/shadow-verify` was not triggered (medium/low pass straight through); the top security claim was nonetheless independently re-derived.
- Not exercised at runtime: the live REPL overlay repaint behavior, the Telegram progress lane's actual client-side rendering of control/bidi bytes (finding #1's residual), and the deliberately out-of-scope items the PR names (spinner-verb rewrite, cross-flush tool grouping).

---

PR #356, opened 2026-07-02
fix(skills): evict project-origin skills across cwd change (#179)

griffinwork40:
## Problem

`collectSkillEntries()` (`src/agent/tools/skill-bridge.ts`) scans `<cwd>/.afk/skills/` into a module-scoped registry `Map` (`src/skills/index.ts`) on every manifest build, but never evicted the previous project-origin entries first. In a short-lived CLI process this is invisible — the process exits before cwd ever changes. In a long-lived daemon or Telegram bot process, once the working directory changes (e.g. daemon serving a different project, or `cd` between tasks), stale project-A skills stayed registered and were merged into project-B's manifest, leaking skills across project boundaries.

Fixes #179.

## Fix

- Added `evictSkillsByOrigin(origin)` to `src/skills/index.ts` — removes every registry entry whose `origin` matches the given value and returns the removed count.
- `collectSkillEntries()` now calls `evictSkillsByOrigin('project')` immediately before rescanning `getProjectSkillsDir()`, so each manifest build starts from a clean project-scope slate before repopulating it from the current cwd.
- User-origin (`~/.afk/skills/`) and built-in skills are deliberately excluded from eviction — they're cwd-agnostic and must survive project switches. Only `'project'`-origin entries are evicted.
- Cost is O(registry size), bounded by the number of registered skills (typically <100), negligible next to the disk scan that follows.

## Tests

Added a new regression suite in `src/agent/tools/skill-bridge.test.ts` (`collectSkillEntries — project skill eviction on cwd change (#179)`) simulating a long-lived daemon serving project A then project B:

1. `evicts project-A skills after switching to project B` — project-A skill disappears once cwd moves to project B, project-B skill appears.
2. `user-origin skills survive the cwd change` — a `~/.afk/skills/`-scoped skill remains registered across both cwd switches while project-scoped skills correctly rotate.
3. `evicted project entries are fully absent from the manifest` — asserts the rendered `buildSkillManifest()` string, not just the registry, no longer mentions the stale skill.

Verified:
- `pnpm lint` (`tsc --noEmit`) — clean.
- `vitest run src/agent/tools/skill-bridge.test.ts` — 39/39 passed.
- `vitest run src/skills` — 452/452 passed (22 files), no collateral breakage.

Issue #179, opened 2026-06-16
skill-bridge: project-origin skills not evicted across cwd change in long-lived daemon


griffinwork40:
## Summary

`collectSkillEntries()` scans `<cwd>/.afk/skills/` and registers each project skill into a **module-scoped** registry `Map` (`src/agent/tools/skill-bridge.ts:94-95` → `scanSkillsFromDir` in `src/skills/user-skills.ts`). Nothing ever **evicts** `project`-origin entries when the working directory changes.

In a long-lived process that serves multiple projects from one session (daemon, Telegram bot), a skill discovered while `cwd` was project A remains in the manifest after switching to project B — even though project B's `.afk/skills/` no longer contains it.

`git grep -nE 'delete|clear|evict|unregister'` over the scan path returns no eviction logic.

## Impact

- Severity: **medium**. Stale project skills leak across project boundaries in long-lived multi-project surfaces.
- Not a problem for one-shot CLI / per-invocation processes (fresh module state each run).
- User-origin skills (`~/.afk/skills/`) are global, so they are not affected.

## Suggested fix

Before the project scan in `collectSkillEntries()`, sweep `listSkills()` and unregister entries where `origin === 'project'`, **or** track the last-scanned project dir and skip the re-scan/eviction when it is unchanged (cheaper steady-state).

## Context

Surfaced in review of a change that scans project skills on every surface (not only the REPL path). That change is correct and merged; this is a pre-existing gap it makes more reachable.

PR #375, opened 2026-07-02
fix(skills): resolve project skills against the session cwd (#179)

griffinwork40:
## Summary

Phase 2 of #179, stacked on #356 (base branch: `fix/skill-evict-on-cwd-change`).

#356 evicts stale project-origin skills when the scan root changes — but the scan root itself was still resolved from the Node host's `process.cwd()`. The only production caller of `process.chdir()` is the interactive REPL's worktree auto-naming flow, so #356 alone fixes only the REPL. On long-lived hosts the session's configured cwd and the host process cwd diverge permanently:

- **daemon**: a task's `sessionConfig.cwd` points at the project, but the process cwd stays wherever `afk daemon` was launched
- **Telegram**: `/cd` stores a per-chat cwd; the bot process never chdirs

On both, project skills (`<cwd>/.afk/skills/`) resolved against the wrong directory — for the whole process lifetime.

## Changes

- **`src/paths.ts`** — `getProjectAfkDir()` / `getProjectSkillsDir()` gain an optional `cwd` param (default `process.cwd()`), mirroring the existing `getProjectPlansDir(cwd)` precedent introduced for the same host-vs-session divergence.
- **`src/agent/tools/skill-bridge.ts`** — `collectSkillEntries()` / `buildSkillManifest()` accept `opts.cwd` and thread it to the project scan. Combined with #356's eviction, each call resolves project skills for the caller's effective cwd with no stale leakage.
- **`src/agent/providers/anthropic-direct/index.ts`** — the manifest build passes the session cwd (`config.cwd`), computed two lines above it, instead of implicitly using the host cwd.
- **`src/agent/tools/skill-executor.ts`** — the "skill not found" available-skills listing resolves against `this.currentCwd` (session/worktree cwd).

All changes are additive and backward-compatible: every new parameter is optional and defaults to the previous behavior.

## Known residual (deliberate)

The skill registry is module-global, so two *concurrent* sessions with different cwds in one process (Telegram) can still race between scan and manifest read. The dominant sequential case is fully fixed. A session-scoped project-skill overlay is the follow-up if concurrent divergent-cwd usage materializes — deferred to keep this change small and additive.

## Tests

Three new cases in `skill-bridge.test.ts`, extending #356's suite:

1. `opts.cwd` resolution with `process.cwd()` held constant — the daemon/Telegram shape #356's review flagged as untested
2. `buildSkillManifest` forwards `opts.cwd`; a later no-override call evicts again (no cross-session leak)
3. eviction of collision-fallback `project:<name>` entries across a cwd change — the untested path flagged in review

## Verification

- `pnpm lint` clean (tsc strict)
- Full suite: **566 files / 10 464 passed / 0 failed / 14 skipped**


griffinwork40 (commented):

# Code Review — PR #375

**`fix(skills): resolve project skills against the session cwd (#179)`** · `main` ← `fix/skill-scan-session-cwd` (stacked on `fix/skill-evict-on-cwd-change`)
**Reviewed ref:** `ce015e0f740e201d9ae27ffa67847523dc30a32e` · 6 files · +255/−20 · regime: full (2 parallel dimension agents + citation/absence verification)
**Stated intent:** PR #375 title + body — spec-compliance **assessed**.

## Verdict: ✅ MERGE (recommendation) — ⚠ but the branch is currently **CONFLICTING**

The diff itself is clean, minimal, additive, and correct — **no critical/high/blocking findings**. But GitHub reports this PR as `mergeable: CONFLICTING` against `main`, so it is **not mergeable as-is**: it needs a rebase/conflict-resolution pass before an actual merge, independent of code quality. (Resolving the conflict is out of scope for this read-only review — flagged, not fixed.) On the merits of the diff, this is a MERGE.

## What I verified as correct (source-grounded at ref `ce015e0f`, cross-checked by an independent verification pass)

- **Backward-compatible signatures (api-compat clean).** `getProjectAfkDir()` / `getProjectSkillsDir()` go from zero-arg to a **defaulted** `cwd: string = process.cwd()` param, and `collectSkillEntries()` / `buildSkillManifest()` gain a **trailing optional** `opts`. Every existing zero-arg caller resolves byte-identically to pre-PR behavior. All non-diff call sites enumerated and re-verified zero-arg: `paths.ts` internal chaining, `getProjectPluginsDir`, `src/cli/slash/builtin-skills.ts:128`, `src/agent/providers/shared/supported-commands.ts:33`, `src/cli/slash/plugin-skills.ts` (both via `collectSkillEntries()`). ✓
- **`currentCwd` is real and wired end-to-end (correctness).** `skill-executor.ts:244` `private currentCwd: string | undefined`, seeded from `ctx.cwd` in the constructor, re-anchored by `setCwd()` on `/cd`/worktree-rename, and forwarded in the not-found branch (`this.currentCwd !== undefined ? { cwd: this.currentCwd } : undefined`). The not-found listing does **not** silently fall back to host cwd when a session cwd was supplied. ✓
- **The `cwd` reorder in the provider is safe.** `const cwd = config.cwd || process.cwd()` moves above the manifest build; nothing between the old and new position reads a bare `cwd`, and there is exactly one `const cwd =` declaration (no double-declare, no dead-zone read). ✓
- **`evictSkillsByOrigin` is correct.** Deleting from a `Map` while iterating `for...of` its entries is well-defined in JS; it removes exactly the `'project'`-origin entries and cannot skip. `SkillMetadata['origin']` includes the `'project'` literal passed. ✓
- **Not a per-turn hot path (perf).** `buildSkillManifest` → `collectSkillEntries` → `evictSkillsByOrigin` runs **once per session construction** (daemon: per task; Telegram: per `/cd` session rebuild; REPL: per process/worktree-chdir), not on every model turn — mid-session cwd changes reuse the cached manifest via `stableSystemPrefix`. O(registry) eviction (<100 skills) is negligible. **No perf regression.** ✓
- **Security clean.** `cwd` flows only through `path.join()` — no shell/`exec`/`spawn` interpolation, no path-injection/RCE surface. The skill-not-found message lists skill **names** only, never the resolved filesystem path (no path/secret disclosure). The four project-scope path helpers are **internal-only** — not re-exported from the public package entry `src/index.ts` (only `getSessionsDir`/`getSkillsDir`/`getAgentFrameworkDir` are). ✓
- **Spec-compliance:** all 4 claimed code changes and all 6 claimed tests are present at the ref and assert the described behavior (non-tautological). No scope creep — every behavior change maps to a named intent line. ✓

**1 · MEDIUM · confidence:high · spec-compliance · `src/agent/providers/openai-compatible/` (whole dir) · ref:ce015e0f · citation-type:file-state**
The intent frames the failure mode as provider-agnostic (daemon / Telegram long-lived hosts) but the fix touches only the **anthropic-direct** provider. The `openai-compatible` provider — a production path selectable via `AFK_PROVIDER` on the same surfaces — calls `buildSkillManifest` **nowhere** (grep of the whole directory via two independent methods → 0 non-test matches), so it injects **no** project-skill manifest at all, cwd-correct or otherwise. **Independently verified as PRE-EXISTING, not introduced by this PR** (the 6-file diff contains no openai-compatible file), so this is a PR-**framing/documentation** gap, not a regression this PR causes.
`evidence:` `grep -rn "buildSkillManifest|skill-bridge|collectSkillEntries" src/agent/providers/openai-compatible/` → 0 matches in non-test files.
`suggestion:` Note in the PR description that the openai-compatible provider has no manifest injection today (pre-existing, out of scope) so reviewers don't assume cross-provider parity from the stated problem framing.

**2 · LOW · confidence:high · test-coverage · `src/agent/tools/skill-executor.ts:357-360` · ref:ce015e0f · citation-type:file-state**
The `this.currentCwd`-forwarding path in the skill-not-found branch is unexercised by any test — the two existing `cwd`-aware executor tests cover the fork path, not the not-found listing.
`evidence:` `grep -n "'not found'" skill-executor.test.ts` → lines 211, 2035; neither co-occurs with a `cwd:`-constructed executor.
`suggestion:` Add a case constructing `SkillExecutor({ …, cwd: '<tmp>' })` and asserting the not-found listing reflects a project skill discoverable only at that cwd.

**3 · LOW · confidence:high · test-coverage · `src/agent/providers/anthropic-direct/index.ts:784,792-793` · ref:ce015e0f · citation-type:file-state**
No test asserts the provider invokes `buildSkillManifest` with `{ cwd }` derived from `config.cwd`; the provider-level cwd-forwarding is verified only by inspection, so a future regression (re-introducing `buildSkillManifest()` with no args) would not be caught.
`evidence:` code-search `buildSkillManifest` in `extension:test.ts` → only `skill-bridge.test.ts` matches.
`suggestion:` Add a thin provider test with a project skill on disk at a non-host `config.cwd`, asserting the manifest text appears in the built system prompt.

**4 · LOW · confidence:high · perf-observability · `src/agent/tools/skill-bridge.ts:150` · ref:ce015e0f · citation-type:file-state**
`evictSkillsByOrigin('project')`'s return (removed count) is discarded at the sole production call site — silent eviction with no debug signal during a daemon/Telegram `/cd` investigation.
`evidence:` `evictSkillsByOrigin('project');` — bare statement, no assignment, no `debugLog`.
`suggestion:` Capture the return value and pass it to the existing zero-cost-when-disabled `debugLog()` for cwd-divergence debugging.

**5 · LOW (nit) · confidence:high · spec-compliance · `src/paths.ts:125-127` · ref:ce015e0f · citation-type:file-state**
`getProjectPluginsDir()` retains a bare `process.cwd()` while its siblings gained a `cwd` param. Correctly scoped (project plugins aren't named in the intent) — noted only as the same divergence class for a future follow-up if project-scope plugins are ever loaded from daemon/Telegram sessions.

## What was not checked

- Reviewed via `gh pr diff` + `gh api …/contents/…?ref=ce015e0f` against branch HEAD `ce015e0f740e201d9ae27ffa67847523dc30a32e`; no local checkout available. All file-state citations pulled at that ref and independently re-verified by a citation/absence pass (5/5 load-bearing claims confirmed, 0 reversed).
- **Not run:** `pnpm lint` / `pnpm test` were not executed (read-only mandate); the PR's "566 files / 10 464 passed / 0 failed" claim was verified only by static reading of test bodies, not a green run.
- No critical/high finding surfaced, so `/shadow-verify` was not triggered (medium/low go straight through).
- The **CONFLICTING** merge state was read from `gh pr view` metadata; the specific conflicting hunks against `main` were not diffed (resolving them is out of scope for this read-only review).
- The disclosed deliberate residual (module-global registry race between two concurrent divergent-cwd Telegram sessions) is out of scope per the PR's own text and was not investigated.
- No CI test-runner job is configured on this ref (only Vercel/docs checks present), so "0 failed" could not be corroborated from CI artifacts.

---
_Read-only review — no files, commits, PR body, labels, or merge state were modified. Verdict: **MERGE** (recommendation only); note the branch is currently CONFLICTING and must be rebased/resolved before an actual merge._

griffinwork40:
## Review: PR #375 — `fix(skills): resolve project skills against the session cwd (#179)`

The `/review` skill dispatch hit its tool-iteration cap and returned no content, so I ran the review directly: read all 5 changed files against their pre-PR baseline, applied the diff artifact in a disposable isolated worktree (never touched your dirty main tree), ran lint + the actual test suite there, and grepped the repo for sibling call sites of every touched function.

### Verdict: **Approve**, with one should-fix-soon gap

### What I verified mechanically (not just read)
- Diff applied cleanly onto current `main` in `.afk-worktrees/review-pr-375` (now removed).
- `pnpm lint` (`tsc --noEmit`, strict) — clean, no errors.
- `skill-bridge.test.ts` 42/42 pass (incl. the 3 new tests), `skill-executor.test.ts` 75/75 pass, `anthropic-direct.test.ts` 42/42 pass.
- Full-suite run showed 1 failure in `daemon.test.ts` (telemetry-file-count assertion); confirmed pre-existing/order-dependent by re-running that file alone → 48/48 pass. The PR touches none of daemon/scheduler/telemetry code, so this is unrelated noise, not a regression from this diff.

### Correctness of the fix — sound
`getProjectAfkDir`/`getProjectSkillsDir` (`src/paths.ts:113,121`) gain an optional `cwd` param with a backward-compatible default, mirroring an existing precedent in the same file (`getProjectPlansDir`, `getProjectSettingsPath` already do this for the identical REPL/daemon/Telegram cwd-divergence reason). The two live call sites that needed it — the Anthropic-direct provider's system-prompt manifest build (`anthropic-direct/index.ts:791-797`) and `SkillExecutor`'s "not found" fallback listing (`skill-executor.ts:373-378`) — are both updated, and the latter now reuses `currentCwd`, which is the same field `setCwd()` already keeps live-synced at 4 other call sites in that file. Non-obvious bonus I confirmed by tracing the code: before this fix, that fallback call also **corrupted the shared skill registry** on a wrong guess (it unconditionally evicts project-origin entries before rescanning with the wrong cwd) — this isn't just a cosmetic error-message fix, it closes a real self-inflicted-registry-corruption path on daemon/Telegram.

### Should-fix — same bug, left unfixed in the sibling provider
`src/agent/providers/openai-compatible/index.ts:322` still calls `buildSkillManifest()` with **no** cwd argument, even though `config.cwd ?? process.cwd()` is already computed two lines above it (line 301) for the environment fragment. This provider backs every `gpt-*`/`o*`/local `org/model` (MLX/llama.cpp/vLLM) session — including daemon and Telegram, the exact scenario this PR and issue #179 target. Any such session still gets the pre-fix bug. The fix is a one-line mirror of what this PR already did for `anthropic-direct`. I'd fix this before merge or file an immediate fast-follow — otherwise the behavior silently splits by provider (works with Claude, not with GPT/local models on the same host).

### Worth knowing, not blocking
- **Manifest goes stale mid-session anyway.** Traced `query.ts:620-670` + `index.ts:894-970`: the skill-manifest text is baked into `stableSystemPrefix` once at `query()` construction; `setCwd()`'s rebuild closure only refreshes the `# Environment` fragment, never the manifest. So a `/cd` or worktree auto-rename mid-session leaves the model reading the old cwd's skill list even though dispatch itself (fixed by this PR) tracks the new one correctly. Pre-existing, not a regression — but the PR's own framing ("resolve against session cwd") could read as fully closing this, and it doesn't.
- **Project-scoped plugins** (`.afk/plugins/*/SKILL.md`) still resolve via the untouched `getProjectPluginsDir()` (`paths.ts:125`, no cwd param) → same bug class, likely intentionally out of scope since the PR/issue title says "skills," not plugins.
- **No provider-level integration test** — the 3 new tests cover `collectSkillEntries`/`buildSkillManifest` taking `opts.cwd` directly (good, targeted), but nothing exercises `AnthropicDirectProvider.query()` end-to-end to assert `config.cwd` actually reaches the rendered system prompt. Low risk, but a future refactor could drop the `{ cwd }` wiring silently.
- Drive-by stale comment at `paths.ts:354` (`getProjectSettingsPath`): "Do NOT call `getProjectAfkDir()` here — that function always reads `process.cwd()`... and ignores any argument" is no longer true post-PR. Harmless, not touched by this diff, worth a note for whoever's next in that file.

**Done**
- Reviewed PR #375 (diff artifact `/Users/griffinlong/.afk/state/skill-preflight/b17504b3-7274-45f2-9c11-027bef5cccb4/pr-375.diff`) end-to-end: read all 5 changed files against baseline, applied the diff in an isolated disposable worktree, ran `pnpm lint` and the relevant vitest files there, and grepped every repo-wide call site of the touched functions.
- Evidence: this transcript's tool calls (lint clean; `skill-bridge.test.ts` 42/42, `skill-executor.test.ts` 75/75, `anthropic-direct.test.ts` 42/42, `daemon.test.ts` 48/48 in isolation) plus file:line citations above (`openai-compatible/index.ts:301,322`, `skill-executor.ts:373-378`, `paths.ts:113,121,125,354`, `query.ts:620-670`, `index.ts:791-970`).
- World state: no files in your working tree were touched — the verification worktree (`.afk-worktrees/review-pr-375`) was created, used, and removed (`worktree remove --force`); its throwaway branch ref `afk/review-pr-375` remains but points at nothing you need.
- Nothing pending on my end; the one actionable item is the `openai-compatible` manifest fix, which is the human's call whether to fold into this PR or track separately.

---

PR #376, opened 2026-07-02
fix(agent): bound foreground safe-batch tool concurrency

griffinwork40:
## What

Add an engine-level ceiling on how many **concurrency-safe** tool calls `SessionToolDispatcher.executeBatch` runs simultaneously within one batched tool round. Default cap: **8**, constructor-injectable via `maxConcurrentSafeCalls`.

## Why

`executeBatch` partitions a turn's tool calls into contiguous runs and executes each concurrency-safe run through a single unbounded `Promise.allSettled`. Because `agent`, `skill`, and `compose` are all classified concurrency-safe (`schemas.ts`), a wide fan-out in one turn — a `compose` layer, or many `agent` calls issued together — spun up an **unbounded** number of in-flight `AgentSession` forks at once. That risks:

- **memory exhaustion** (N full child sessions materialized simultaneously), and
- **provider rate-limit (429) storms** — the exact failure class the `fanout-pace` skill currently mitigates at the *prompt* layer because the engine offered no ceiling.

Background jobs already cap themselves at `DEFAULT_MAX_CONCURRENT_JOBS = 10` (`background-registry.ts`); the foreground safe-batch had no analog. This PR closes that gap at the engine layer.

This is **Option 2** of a design that was validated with `/shadow-verify` (4 claims independently re-derived → all CONFIRMED) and `/devils-advocate` (3 critics → convergent → composition-boundary check CONFIRMED). The critique specifically **refuted** the alternative "just serialize subagents against sibling tools" framing (it would kill useful `agent`+`read` overlap for zero isolation gain) and confirmed the missing foreground cap as the genuine latent issue.

## How (deliberately minimal)

- New module-local worker-pool `settleWithConcurrencyLimit(items, limit, worker)` drains the safe batch through at most `limit` in-flight workers, returning results in **input order** with `Promise.allSettled` shape. Within the cap it is behaviourally identical to the previous code.
- `partitionIntoBatches` is **untouched** — grouping semantics unchanged, so the binary-contract partition tests are unaffected.
- **Both providers** (`anthropic-direct/loop.ts`, `openai-compatible/query.ts`) inherit the cap for free via the shared dispatcher — no call-site edits.
- **Abort is checked at dispatch time inside the worker**, not at admission: a call cleared in phase 1 that aborts while queued behind the cap is still handled correctly. No queue-admission timer is introduced (there are no per-call timers today; adding one would invent a spurious-timeout bug).
- Result order stays keyed by `originalIndex` — completion-order independent, exactly as before.
- **Constructor-injectable, no env var** — mirrors the closest precedent (`background-registry.ts`). Env-tunability is a trivial follow-up if wanted.

### Why 8

Above the typical parallel-read width, so ordinary fan-out of cheap read tools is never throttled, while bounding the blast radius of a runaway subagent fan-out. (Background jobs use 10 for comparison.) Must stay `>= 2` or the parallel-timing tests regress.

## Explicitly out of scope

Distinguishing cheap leaf reads from expensive subagent forks via a separate `weight` axis (so reads never count against the subagent ceiling) is **design Option 4** — deferred to a follow-up, to be explored in a fresh session.

## Known limitation (unchanged by this PR)

Foreground subagents that share the parent's `cwd` can still race on the filesystem — that TOCTOU is pre-existing and acknowledged (`write-file.ts`), and is better addressed by worktree isolation than by throttling concurrency. This PR does not change it.

## Tests

4 new cases in `dispatcher.test.ts`:
- cap is enforced (peak in-flight == configured limit across a 6-wide batch),
- a cap wider than the batch runs fully parallel (peak == batch width),
- result order is preserved when draining a batch wider than the cap (descending delays),
- a sub-1 cap falls back to the default rather than deadlocking.

**Full suite green: 566 files, 10473 passed, 14 skipped, 0 failed.** `tsc --noEmit` clean. No `@anthropic-ai/sdk` imports added (`audit:sdk:check` unaffected); no env var added (`scan:env:check` unaffected).

### DIFF
--- a/src/agent/tools/dispatcher.test.ts
+++ b/src/agent/tools/dispatcher.test.ts
@@ -852,6 +852,108 @@ describe('SessionToolDispatcher', () => {
       expect(results[0]!.content).toBe('slow');
       expect(results[1]!.content).toBe('fast');
     });
+
+    describe('maxConcurrentSafeCalls (bounded concurrency)', () => {
+      // A safe handler that records concurrency: increments a live counter on
+      // entry, tracks the peak, decrements on exit. `peak` is the maximum
+      // number that were ever in flight simultaneously.
+      function makeConcurrencyProbe() {
+        const state = { inFlight: 0, peak: 0 };
+        const handler: ToolHandler = async () => {
+          state.inFlight += 1;
+          state.peak = Math.max(state.peak, state.inFlight);
+          await new Promise((r) => setTimeout(r, 20));
+          state.inFlight -= 1;
+          return { content: 'ok' };
+        };
+        return { state, handler };
+      }
+
+      it('caps simultaneous in-flight safe calls at the configured limit', async () => {
+        const { state, handler } = makeConcurrencyProbe();
+        const dispatcher = makeDispatcher({
+          handlers: new Map([['read_file', handler]]),
+          permissions: { allowedTools: ['read_file'] },
+          maxConcurrentSafeCalls: 2,
+        });
+
+        const calls = Array.from({ length: 6 }, (_, i) =>
+          makeBatchCall('read_file', `read-${i}`),
+        );
+        const results = await dispatcher.executeBatch(calls);
+
+        expect(results).toHaveLength(6);
+        expect(results.every((r) => r.content === 'ok')).toBe(true);
+        // Never more than 2 running at once, despite 6 safe calls in the batch.
+        expect(state.peak).toBe(2);
+      });
+
+      it('runs the whole batch concurrently when the cap exceeds batch width', async () => {
+        const { state, handler } = makeConcurrencyProbe();
+        const dispatcher = makeDispatcher({
+          handlers: new Map([['read_file', handler]]),
+          permissions: { allowedTools: ['read_file'] },
+          maxConcurrentSafeCalls: 10,
+        });
+
+        const calls = Array.from({ length: 4 }, (_, i) =>
+          makeBatchCall('read_file', `read-${i}`),
+        );
+        await dispatcher.executeBatch(calls);
+
+        // Cap (10) > batch width (4): all four run at once, like allSettled.
+        expect(state.peak).toBe(4);
+      });
+
+      it('preserves result order when draining a batch wider than the cap', async () => {
+        // Descending delays: without index-keyed write-back, a naive pool
+        // would return results in completion order (fastest first).
+        const mk = (ms: number, content: string): ToolHandler => async () => {
+          await new Promise((r) => setTimeout(r, ms));
+          return { content };
+        };
+        const dispatcher = makeDispatcher({
+          handlers: new Map([
+            ['read_file', mk(40, 'a')],
+            ['glob', mk(30, 'b')],
+            ['grep', mk(20, 'c')],
+            ['list_directory', mk(10, 'd')],
+          ]),
+          permissions: { allowedTools: ['read_file', 'glob', 'grep', 'list_directory'] },
+          maxConcurrentSafeCalls: 2,
+        });
+
+        const results = await dispatcher.executeBatch([
+          makeBatchCall('read_file'),
+          makeBatchCall('glob'),
+          makeBatchCall('grep'),
+          makeBatchCall('list_directory'),
+        ]);
+
+        expect(results.map((r) => r.content)).toEqual(['a', 'b', 'c', 'd']);
+      });
+
+      it('degrades to sequential (not deadlock) when the cap is below 1', async () => {
+        // A non-positive/non-finite cap falls back to the default in the
+        // constructor, so behaviour stays parallel — assert it does not hang
+        // and every call still resolves.
+        const { state, handler } = makeConcurrencyProbe();
+        const dispatcher = makeDispatcher({
+          handlers: new Map([['read_file', handler]]),
+          permissions: { allowedTools: ['read_file'] },
+          maxConcurrentSafeCalls: 0,
+        });
+
+        const calls = Array.from({ length: 3 }, (_, i) =>
+          makeBatchCall('read_file', `read-${i}`),
+        );
+        const results = await dispatcher.executeBatch(calls);
+
+        expect(results.map((r) => r.content)).toEqual(['ok', 'ok', 'ok']);
+        // Default cap (8) applies → all 3 run at once.
+        expect(state.peak).toBe(3);
+      });
+    });
   });
 
   describe('compose tool routing (L4)', () => {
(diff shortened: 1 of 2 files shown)


griffinwork40:
Verification confirms all claims. Wave 1.5 complete:
- **NaN edge (Finding 1):** helper's only call site (`dispatcher.ts:963`) passes `this.maxConcurrentSafeCalls`, which the constructor (`324-328`) normalizes through a `typeof === 'number' && Number.isFinite && >= 1` guard → NaN/Infinity/<1 can never reach the helper. Latent-only. Confirmed.
- **Untested sequential floor + misleading test name (Finding 2):** zero direct sub-1/NaN pool tests; helper is unexported → floor path unreachable from tests. Absence confirmed.
- **No observability (Finding 3):** zero trace/log in helper region. Confirmed; out of scope for the stated intent.

All findings are low/nit — no blocking/critical/high, so `/shadow-verify` is not triggered. Synthesizing.

---

# Code Review — PR #376: `fix(agent): bound foreground safe-batch tool concurrency`

**Target:** PR #376 `fix/bounded-safe-batch-concurrency` → `main` · reviewed ref `dd5e437` · 2 files, +188/−6 (194 lines) · regime: full (2-agent wave; elevated from light because the change is a concurrency primitive in the core dispatch hot path)

## Summary
Adds an engine-level ceiling (default `8`, constructor-injectable via `maxConcurrentSafeCalls`) on how many concurrency-safe tool calls `SessionToolDispatcher.executeBatch` runs at once, via a new module-local worker pool `settleWithConcurrencyLimit(items, limit, worker)`. Both providers inherit the cap through the shared dispatcher with zero call-site edits. The change is minimal, correct on all reachable paths, and well-tested.

## Spec-compliance (assessed against PR #376 title+body, R1–R7)
All met, no scope creep:
- **R1** cap wired into the safe branch of `executeBatch` — `dispatcher.ts:963-965` ✓
- **R2** default 8 + injectable option — `dispatcher.ts:136,210,324-329` ✓
- **R3** behaviourally identical within cap (peak == batch width when `limit ≥ width`) — verified by both agents' execution-replica; test `dispatcher.test.ts:891-906` ✓
- **R4** order preserved, keyed by `originalIndex`, completion-order independent — `dispatcher.ts:967,979-980`; test `:908-934` ✓
- **R5** sub-1/non-finite → default fallback — `dispatcher.ts:324-329` ✓
- **R6** both providers route through shared `executeBatch` (`providers/anthropic-direct/loop.ts:566`, `providers/openai-compatible/query.ts:1091`); diff touches only `dispatcher.ts` + its test — no call-site edits ✓
- **R7** 4 new tests map 1:1 to spec'd assertions — `dispatcher.test.ts:856-956` ✓

## Correctness (verified at ref `dd5e437`)
- **Shared-cursor race:** `while (cursor < items.length) { const i = cursor++; }` has no `await` between test and increment → each index claimed exactly once (no double-dispatch, no skip). Clean.
- **No result holes** on reachable paths; every slot 0..n−1 is written.
- **Ordering:** both fulfilled (`outcome.value.originalIndex`) and rejected (`settled.indexOf(outcome)` — reference-safe, distinct object per slot) branches key by original index.
- **No unhandled rejection:** the only throwable (`await worker(...)`) is inside try/catch; runner IIFEs always resolve → `Promise.all(runners)` cannot reject.
- **Abort semantics** preserved: checked at dispatch time inside the worker, matching prior behaviour.

## Findings (all low / non-blocking)

**1 · low · confidence high · robustness · `src/agent/tools/dispatcher.ts:153` · ref:dd5e437 · file-state**
`settleWithConcurrencyLimit` returns an all-hole result array if called with `limit = NaN`: `Math.max(1, Math.floor(NaN))` is `NaN`, so `for (w=0; w<NaN; …)` never runs and no slot is written — contradicting the docstring's "floored at 1 … degrades to sequential rather than deadlocking" claim.
Evidence: `const poolSize = Math.min(Math.max(1, Math.floor(limit)), items.length);`
**Unreachable today:** the sole call site (`:963`) passes `this.maxConcurrentSafeCalls`, which the constructor guard (`:324-328`, `Number.isFinite && >= 1`) can never leave as NaN. Latent robustness/doc gap only.
Suggestion: guard NaN inside the helper (`Number.isFinite(limit) ? … : 1`) so the docstring's self-contained-robustness claim holds independent of the caller.

**2 · low · confidence high · test-coverage · `src/agent/tools/dispatcher.test.ts:936-955` + `dispatcher.ts:153` · ref:dd5e437 · file-state**
The test named *"degrades to sequential (not deadlock) when the cap is below 1"* passes `maxConcurrentSafeCalls: 0` and asserts `state.peak === 3` — i.e. fully **parallel**, not sequential. It exercises the constructor's 0→8 fallback, not the helper's `Math.max(1, …)` sequential floor (`poolSize→1`), which the constructor intercepts before the helper ever sees a sub-1 value. The helper is unexported, so that floor path is untested.
Evidence: `maxConcurrentSafeCalls: 0` … `expect(state.peak).toBe(3);`
Suggestion: rename to reflect the constructor fallback (e.g. "sub-1 cap falls back to the default"), and if the helper's sequential floor matters, cover it directly (test-only export or spy).

**3 · low · confidence medium · perf-observability · `src/agent/tools/dispatcher.ts:955-977` · ref:dd5e437 · file-state**
No trace/log event marks when the new cap is actively queuing calls, so throttling is invisible to async operators. Explicitly out of scope for R1–R7 — flagged for awareness only; not required for merge.
Suggestion: none required by the stated intent.

## Merge decision: **MERGE**

The change satisfies its stated intent end-to-end, is correct on all reachable paths, and preserves prior ordering/abort/allSettled semantics within the cap. The three findings are all low-severity polish (a latent NaN-robustness gap that is unreachable behind the constructor guard, a misleading test name over a genuinely-untested floor path, and an optional observability nit) — none blocks merge.

## What was not checked
- **Ref:** all citations read against branch HEAD `dd5e437` via `git show` — verified at that ref (not against `main`, not diff-context alone).
- **Stated intent:** PR #376 title+body — spec-compliance assessed (R1–R7).
- Did **not** re-run the test suite; relying on the PR's claim of full-suite green (10473 passed) and static + execution-replica analysis by the sub-agents.
- Did **not** empirically tune the default `8` against real fan-out widths (design judgment; PR cites prior `/shadow-verify` + `/devils-advocate` validation).
- Did **not** review the pre-existing shared-`cwd` TOCTOU the PR calls out as out of scope, nor the `partitionIntoBatches` grouping (byte-unchanged in the diff).
- Read-only review: no files, PR state, or git refs were modified.

---

PR #377, opened 2026-07-02
fix(agent): close reverse-direction forkSubagent credential leak

griffinwork40:
## Problem

Follow-up to #374 (merged). That PR made `SubagentManager.forkSubagent`'s parent-credential fallback refuse to hand an Anthropic `sk-ant-*` key to an OpenAI-routed child — but the guard was **one-directional**. The mirror leak stayed open: an OpenAI-shaped parent key (`sk-proj-…`) flows to an **Anthropic-routed** child via the `otherwise → parentApiKey` branch, sending an OpenAI credential as a Bearer to `api.anthropic.com`.

Reachable: `getApiKey()` returns a *provider-appropriate* key (`shared-helpers.ts`), so an OpenAI top-level session gives the manager an OpenAI-shaped `parentApiKey`; forking a `sonnet` child then hits the unguarded branch. The sibling `applyParentCredentialFallback` already blocks its mirror case — `applyManagerApiKeyFallback` did not.

**Root cause (single source of truth):** `resolveCredentialForModel(model)` already computes `providerForModel(model)` to pick the key, then throws the provider away. The `SubagentManager` boundary receives only a bare `apiKey` string, forcing the guard to *guess* the provider from key shape. Any prefix-sniff (`sk-proj-`/`sk-svcacct-`) has a silent gap (legacy bare `sk-` OpenAI keys) and is format-fragile — unacceptable on a credential path.

## Fix — provider-identity gate

Stop guessing. The parent **model** is the source of truth (both key and provider derive from it).

- **`SubagentManager` gains a `parentModel` option** and derives `parentProvider = providerForModel(parentModel)` **once**, in the constructor.
- **`applyManagerApiKeyFallback` now gates on provider identity in both directions**: a parent credential is inherited only by a same-provider child. When `parentProvider` is absent it degrades to the old key-shape forward guard — so **no existing caller regresses** and the change is purely additive.
- **`parentModel` threaded through all 10 `apiKey`-bearing construction sites**, each using the model its credential was resolved *for*: `getApiKey()` sites → `getModel()` (AFK_MODEL); `ctx.apiKey` sites → `ctx.defaultModel`; derived-key sites → the child model. Using the wrong model here (e.g. the child/subagent model for a `getApiKey()` key) would mischaracterize the provider and *enable* a leak — hence the per-site care.

Unlike prefix-sniffing, the identity gate **preserves non-`sk-ant` local-Anthropic-shim inheritance** (routes to `anthropic-direct`, same provider as the child) via provider identity rather than key shape.

`baseUrl` guard left as-is: it already clears correctly for OpenAI children, and `baseUrl` is an endpoint, not a secret.

## Tests

- 5 new unit tests in `child-credential.test.ts` (incl. one proving the non-`sk-ant` local-shim path survives via provider identity, not sniffing).
- 2 new real-manager integration tests in `subagent.test.ts` (only `AgentSession` mocked): an OpenAI parent key never reaches a `sonnet` child; same-provider OpenAI→OpenAI still inherits.
- Existing #374 tests unchanged (the `parentProvider`-absent fallback reproduces prior behavior).
- **Mutation-verified:** reverting the guard to its one-directional form fails *exactly* the two reverse-leak tests and nothing else.
- Full suite: **10,476 passed / 14 skipped, 0 failures.** `pnpm lint` clean.

## Follow-up

A related pre-existing leak (mint `_phases` forwarding ambient `getApiKey()` as an explicit fork key, independent of this guard) is tracked separately — see the linked issue.


griffinwork40:
Follow-up tracked in #378 — the mint `_phases` explicit-key (`getApiKey()`) leak, which is independent of this PR's manager-fallback gate ('caller wins' short-circuits the guard).

griffinwork40:
Full suite is green at this ref: **10,487 passed / 14 skipped / 0 failures** (the runner ran the whole suite, not just the 4 files — even better coverage confirmation; the new credential tests and the modified `getModel`-mock test files all pass, no regression).

I now have complete, independently-verified ground truth. The compose review-agent wave timed out, so I performed all verification directly via `git show`/`git grep` at ref `8d6b6ea` and a full test run. No `critical`/`high`/`medium` findings emerged, so shadow-verify is not triggered. Here is the synthesized review.

---

# Code Review — PR #377

**Target:** PR #377 `fix(agent): close reverse-direction forkSubagent credential leak` · `fix/subagent-fork-reverse-credential-leak` → `main` · reviewed ref `8d6b6ea` · +224/−36, 15 files
**Change type:** hotfix (credential/security) · **Regime:** upgraded light→**full** (credential path) · **Stated intent:** PR title+body — spec-compliance assessed.

## Verdict: no blocking findings

The fix is correct, complete for its stated scope, and well-tested. Every load-bearing claim in the PR body was independently verified against source at the reviewed ref.

**1. Core gate logic is correct in both directions** — `child-credential.ts:91-104`, ref `8d6b6ea`, citation-type: file-state.
`applyManagerApiKeyFallback` blocks the reverse leak (OpenAI parent → Anthropic child) *and* preserves the forward guard. `providerForModel` returns exactly `'anthropic-direct'`/`'openai-compatible'` (`providers/index.ts:58-61,140`), so the `===` gate is exact. Edge cases safe: `childModel: undefined` → `providerForModel(undefined)` defaults to `'anthropic-direct'` (`providers/index.ts:243`), so an OpenAI parent key still cannot reach an undefined-model child; empty `configApiKey` handled by the `length > 0` guard.

**2. Completeness — all 10 manager-apiKey sites threaded, none missed** — verified by exhaustive `git grep "new SubagentManager("` across all of `src/` at the ref. Every site that passes a manager-level `apiKey` also passes `parentModel`:
`compose-executor:568`, `skill-executor:701`, `skill-executor:1000`, `chat:434`, `daemon:91`, `bootstrap:187`, `audit-fit:347`, `diagnose/orchestrator:367`, `mint/parallelize-dispatch:101`, `telegram:249`. (`skills/index.ts:19` "HAS-apiKey" is a docstring, not code.) Sites with **no** manager apiKey (`skill-executor:576`, `subagent-executor:615`, `farm:371`, `user-skills:156`) have no manager credential to leak.

**3. The `parentModel`↔`apiKey` coupling is sound at every site** — the invariant the PR itself flags as leak-critical:
- `getApiKey()`/`getModel()` (chat, bootstrap, daemon@`daemon.ts:401-402`, mint-parallelize): both read `AFK_MODEL ?? CLAUDE_MODEL`; when unset, `providerForModel(undefined) === providerForModel('sonnet') === 'anthropic-direct'` — providers always agree (`shared-helpers.ts:145-190`).
- `skill-executor:701/1000`: `apiKey` is explicitly resolved *for* the paired model via `applyParentCredentialFallback({ childModel: skillChildModel, … })` — strongest coupling.
- `compose/audit-fit/diagnose`: `apiKey = ctx.apiKey`, `parentModel = ctx.defaultModel`; `ctx.defaultModel` is set to the session model at every dispatch site, and compose-executor already derives `parentProvider` from `ctx.defaultModel` (`:609-612`) — a pre-existing invariant this PR is consistent with.
- `telegram:249`: `telegramApiKey`/`sessionConfig.model` are the same values used to build the parent session's provider — internally consistent (only reachable in the `!isCodex`/Anthropic branch; telegram has exactly one manager).

**4. Executor sites are protected at the per-fork level** — `applyParentCredentialFallback` (`child-credential.ts:41-50`) is genuinely both-directional (it inherits the parent key only when `isAnthropicCredential(parentApiKey)`, so an OpenAI parent key is never inherited by an Anthropic child), backing the PR body's claim.

**5. API-compat / regression** — both new params (`SubagentManagerOptions.parentModel`, `applyManagerApiKeyFallback.parentProvider`) are optional; purely additive. Full suite green at the ref: **10,487 passed / 14 skipped / 0 failures**.

### Non-blocking findings (follow-up, do not gate merge)

- `low · high confidence · security(defense-in-depth) · child-credential.ts:100-103 · ref:8d6b6ea · file-state` — When `parentProvider` is absent and the parent key is non-`sk-ant` (OpenAI-shaped), the legacy branch returns `parentApiKey` unguarded, so a future caller that sets `apiKey` but forgets `parentModel` would reopen the reverse leak. **No current reachable path** (all 10 live sites pass `parentModel`; verified). It is documented intentional degradation (docstring `:79-84`). Suggestion: add a constructor dev-assertion or field-doc requiring `parentModel` whenever `apiKey` is set, to fail loudly on future omission.
- `nit · test-coverage · child-credential.test.ts · ref:8d6b6ea` — No case exercises `applyManagerApiKeyFallback` with `childModel: undefined`. Suggestion: assert undefined-child + OpenAI `parentProvider` → `undefined`.
- `nit · test-coverage · child-credential.test.ts · ref:8d6b6ea` — No test locks the documented legacy reverse-degradation (parentProvider-absent + OpenAI key + Anthropic child → inherits), which would pin the behavior described in the finding above.

### Epistemic scope disclosure

- **Ref:** all citations read against branch HEAD `8d6b6ea` via `git show` — verified at that ref. Working tree HEAD == this ref; the 6 dirty files are unrelated untracked scratch files.
- **Stated intent:** PR #377 title+body — spec-compliance assessed; the fix satisfies its intent with no scope creep.
- **Not checked / out of scope:** the 7 `mint/_phases/{build,heal,plan,research,ship,spec,verify}.ts` explicit-`apiKey: getApiKey()` fork bypass — this is a *different* mechanism the PR explicitly acknowledges and tracks as a separate follow-up (confirmed: `parallelize-dispatch` uses the manager-gate path and passes no explicit fork key, while the others do); live API-call runtime behavior (static + unit/integration only); the internals of `applyParentCredentialFallback` beyond confirming its both-directional gate (pre-existing, unchanged here).
- **Process note:** the parallel review-agent compose wave timed out (`corr_spec` exceeded 8min; `verify` skipped fail-fast); I did the citation/absence/coupling verification directly.

## MERGE ✅

Ship it. The reverse-direction credential leak is closed with a sound, model-as-source-of-truth design; all live call sites are covered and correctly coupled; the change is additive and non-regressing; tests are strong and green. The three follow-up items are hardening/test-completeness nits for a later PR, not merge blockers.

**Done**
- Completed read-only review of PR #377 across security, api-compat, completeness, correctness, spec-compliance, test-coverage, and perf; verdict **MERGE** with 3 non-blocking follow-up nits.
- Evidence: all findings cited to `file:line` at ref `8d6b6ea` (verified via `git show`/`git grep`); completeness proven by exhaustive `new SubagentManager(` sweep of `src/`; full test suite run green (10,487 passed / 14 skipped / 0 failures); diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/987c6acd-1d22-4791-90d8-42a385fbd7e8/pr-377.diff`.
- World unchanged: review is read-only — no files, git state, or PR were modified; working tree preserved.
- Deferred: the parallel compose review wave timed out, so verification was done inline (noted above); the 3 nit-level follow-ups (constructor guard for missing `parentModel`, two edge-case tests) are left for a future PR and are not merge blockers.

---

PR #379, opened 2026-07-02
feat(tui): collapse sequential same-tool runs into one grouped ×N row

griffinwork40:
## Problem
In the interactive REPL, a run of sequential same-tool calls (e.g. 8 back-to-back `bash git show`) renders as N separate scrollback rows — a wall of near-identical lines. Collapsing into a single `bash ×N` row already exists and already works for **parallel** same-tool blocks, but not for **sequential** runs.

## Root cause
`flushToolLaneToScrollback` fires on every incoming `tool_use_detail` (`stream-renderer-orchestrator.ts`), and `flushCompletedRoots` (`tool-lane.ts`) commits **and deletes** the single prior completed root each time. So each flush batch held exactly one completed root → the grouped-render path (`entries.length > 1` → `formatGroupedToolResults`) was never reached for sequential tools. Effective grouping window = 1. (Parallel blocks group because their details arrive together as in-flight entries and flush as one batch.)

## Fix
At the single per-tool orchestrator call site, **skip the eager flush** when the incoming tool extends the trailing completed same-tool run and no thinking phase is pending — so the run accumulates in the lane and commits as one `toolName ×N` block when it breaks (different tool / thinking / prose / turn end).

- `ToolLane.peekTrailingCompletedRootToolName()` reports the trailing completed **flat** root's name (undefined for in-flight or NESTING roots, so dispatch tools and parallel blocks are untouched).
- Root grouping fires at **≥2** — no threshold (`GROUP_THRESHOLD_LEAF` only gates nested children).

## Correctness / constraints respected
- **Append-only ordering:** a pending thinking phase MUST break the run — `commitThinkingPhase` emits an inline `◆ thought for Xs` line, and a held run would otherwise commit *after* it, reordering the thought above its own tool group. So **thinking-interleaved runs intentionally do not collapse** (default `thinkingMode='summary'` commits these between tools).
- **No data loss:** `dispose()`'s unconditional `flush()` drains any held run on turn end / abort / error — no error-branch flush needed.
- **Diff sidecars & subagent rows:** `addDiff` is keyed by `toolUseId` (holding longer only helps attachment); `flushSource` (subagent path) is disjoint from `flushCompletedRoots`. Unaffected.

## Tests
- New: `ToolLane.peekTrailingCompletedRootToolName` edge cases (empty / completed / in-flight / NESTING / nested-child / most-recent).
- New: orchestrator run-accumulation — collapse ≥2, collapse at exactly 2, no-collapse for distinct tools, and the **thinking-break** (a thought between same-tool calls prevents collapse).
- Updated: the "eager-flush on next tool_use_detail" regression test now uses a *different* next tool (the tool-change path still flushes immediately); same-tool accumulation is pinned by the new tests.
- Gates: `pnpm lint` clean; full suite **10,479 passed / 0 failed**.

## Scope / provenance
Independent of #373 (touches the flush cadence, not the progress banner) — branched off `main` as a separate PR. This is the "cross-flush tool grouping window" follow-up deferred from the loading-state-feedback work.

## Follow-ups (out of scope)
- Unify the two grouping mechanisms (root ad-hoc bucketing vs `groupSiblings` for nested children) into one threshold/format policy.
- Optional: render a held run as one live-overlay unit to remove the bounded `MAX_OVERLAY_ROOTS` transient-elision during a long run.

### DIFF
--- a/src/cli/_lib/stream-renderer-orchestrator.test.ts
+++ b/src/cli/_lib/stream-renderer-orchestrator.test.ts
@@ -542,10 +542,16 @@ describe('handleOrchestratorEvent — tool-use-loop scrollback (no content emiss
     return { ctx, setOverlay, commitAbove };
   }
 
-  it('eager-flushes completed root tools on the NEXT tool_use_detail', () => {
+  it('eager-flushes a completed root tool on the NEXT (different) tool_use_detail', () => {
     // The core regression test: simulate a tool-use loop with no content
-    // chunks. After tool B starts, tool A (now completed) must be in
-    // scrollback — not stuck in the live overlay waiting for content.
+    // chunks. After a DIFFERENT tool starts, the prior completed tool must be
+    // in scrollback — not stuck in the live overlay waiting for content.
+    //
+    // Note: the next tool here is a DIFFERENT tool (edit_file) so the
+    // cross-flush run-accumulation gate does not hold it. Consecutive
+    // SAME-tool calls intentionally accumulate into one grouped `×N` row —
+    // that behavior is pinned by the 'cross-flush tool grouping' describe
+    // block below.
     const toolLane = new ToolLane();
     const { ctx, commitAbove } = makeLoopCtx(toolLane);
     const source: SourceState = freshSourceState('__main__');
@@ -558,15 +564,16 @@ describe('handleOrchestratorEvent — tool-use-loop scrollback (no content emiss
     // flush trigger is on the NEXT tool_use_detail, not on tool_result).
     expect(commitAbove).not.toHaveBeenCalled();
 
-    // Iteration 2: bash B starts. THIS is the flush trigger.
-    handleOrchestratorEvent(bashStart('bash-B', '"pwd"'), source, ctx);
+    // Iteration 2: a DIFFERENT tool (edit_file) starts. THIS is the flush
+    // trigger — a tool-name change never holds.
+    handleOrchestratorEvent(editStart('edit-B'), source, ctx);
 
     // bash-A must now be in scrollback.
     const scrollback = commitAbove.mock.calls.map((c) => strip(c[0])).join('\n');
     expect(scrollback).toContain('ls');
     expect(toolLane.hasEntry('bash-A')).toBe(false);
-    // bash-B must still be live (not committed — it just started).
-    expect(toolLane.hasEntry('bash-B')).toBe(true);
+    // edit-B must still be live (not committed — it just started).
+    expect(toolLane.hasEntry('edit-B')).toBe(true);
   });
 
   it('preserves tool_diff sidecar visibility — diff lands BEFORE the eager flush', () => {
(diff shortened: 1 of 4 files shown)


griffinwork40 (commented):

**Merge decision: ✅ MERGE**

Scope: `feat(tui): collapse sequential same-tool runs into one grouped ×N row` — 4 files (~283 lines; ~60 production). Reviewed against branch HEAD `548032730`; all `blocking`/`high` citations were re-read at that ref. No findings above `low`.

### Findings by dimension

- **security** — no issues. No new I/O or auth surface; `toolName` is a provider-controlled string compared only by `===` (`stream-renderer-orchestrator.ts:206`).
- **api-compat** — no issues. `peekTrailingCompletedRootToolName` (`tool-lane.ts:301`) is additive with a single call site; full-tree grep found zero production importers beyond that call site → non-breaking.
- **correctness** — no issues. Verified at ref `548032730`:
  - `peekTrailingCompletedRootToolName` walks `order` newest→oldest, skips text + `agentContext` children, and returns `undefined` for the first flat-root that is in-flight (`result === undefined`) or a `NESTING_TOOLS` member — matches the docstring exactly.
  - `dispose()` (`stream-renderer.ts:757`) unconditionally calls `toolLane.flush()` when `hasPending()`, reachable from the turn-handler `catch`/`finally` and the pause/resume swap → a held run cannot be silently dropped on error/abort/resume (no-data-loss invariant holds).
  - `thinkingPhaseStartedAt` is set only inside the `thinking` handler gated by `thinkingMode !== 'off'`, so `== null` is a correct "no thought pending" check.
- **spec-compliance** — intent met. All three claimed invariants (append-only ordering, no-data-loss, diff-sidecar disjointness) verified in code. Root-grouping-fires-at-≥2 confirmed: `renderGroupedRootTools` groups by `toolName` with no threshold, unlike `groupSiblings`' `GROUP_THRESHOLD_LEAF=3` (nested children only). No scope creep — the one modified pre-existing test is a necessary retarget of a regression whose original assertion is superseded by the intended behavior.
- **test-coverage** — no issues. New tests cover collapse-at-≥2, collapse-at-exactly-2, no-collapse-for-distinct-tools, and the thinking-break, plus 6 `peekTrailingCompletedRootToolName` unit cases (empty / completed / in-flight / NESTING / nested-child / most-recent). Prose-break and error-abort drain paths are structurally covered by pre-existing `dispose()` safety-net tests (not a gap introduced here).
- **perf-observability** — no issues. The peek is an O(k) backward scan (0–2 hops in practice) invoked once per `tool_use_detail`; no new unbounded accumulation — `MAX_OVERLAY_ROOTS` still caps the live overlay and a held run drains at the next tool-name change / thinking / prose / turn end.

### What was not checked
Runtime/manual TTY visual output; unchanged rendering paths (`formatGroupedToolResults`, diff separators); CI status. Spec-compliance was fully assessed against the PR title+body. Load-bearing claims (the hold-run gate, the peek loop, and the `dispose()` drain) were independently re-verified at ref `548032730` during synthesis.

_No blocking or high-severity findings. A maintainer should still confirm the visual scrollback rendering before merge._

griffinwork40 (commented):

# Code Review — PR #379

**`feat(tui): collapse sequential same-tool runs into one grouped ×N row`** · `main` ← `afk/tool-lane-cross-flush-grouping`
**Reviewed ref:** `548032730adc468647641191c1fcc1c49c0fad6b` · 4 files (2 production, 2 test) · +283 diff lines · regime: **full** (2 parallel dimension agents + citation/absence verification)
**Stated intent:** PR #379 title + body — spec-compliance **assessed**.

## Verdict: ✅ MERGE

A tightly-scoped TUI change: exactly one new pure method (`ToolLane.peekTrailingCompletedRootToolName()`) plus one conditional at the single per-tool orchestrator flush site. It does precisely what its intent states and nothing more. **No critical, high, or medium findings** across security, api-compat, correctness, spec-compliance, test-coverage, or perf-observability. The findings below are three low/nit test-coverage gaps and two low/nit perf-observability notes — none gate merge. CI is green at the reviewed ref (Lint & Build, Test, Docs, Bundle Smoke all SUCCESS).

## What I verified as correct (source-grounded at ref `548032730`, cross-checked by an absence-verification pass)

- **No data loss (the load-bearing safety claim).** `StreamRenderer.dispose()` (`stream-renderer.ts:757-775`) unconditionally runs `if (this.toolLane.hasPending()) { … this.toolLane.flush() }` as a safety net, and `flush()` (`tool-lane.ts:854-914`) routes a held ≥2 same-tool run through the *same* `renderGroupedRootTools` grouping path as the normal flush. All three production `StreamRenderer` constructors dispose in a `finally` (`turn-handler.ts:703-709`, `run-skill-dispatch-turn.ts:235-237`, `init.ts:126-127`); SIGINT/SIGTERM unwind the same `try/finally`. A held run cannot be silently dropped. ✓
- **Every hold-run exit path commits (correctness).** (a) different next tool → `trailingRun !== chunk.toolName` → flush; (b) thinking pending → `thinkingPhaseStartedAt != null` → flush; (c) prose/content chunk → the `content` arm (`stream-renderer-orchestrator.ts:312`) flushes *unconditionally*, outside the `holdRun` scope; (d) turn end → `finalizeOrchestrator` schedules `flushCompletedRoots()` **and** `dispose()` drains independently. No reachable path leaves a run uncommitted. ✓
- **`peekTrailingCompletedRootToolName` semantics are sound.** `agentContext` is the same root/non-root discriminant used by `flushCompletedRoots` and `getOverlay` (consistent precedent); the scan returns immediately at the first flat root and correctly yields `undefined` for in-flight (`result===undefined`) or `NESTING_TOOLS` trailing roots, so dispatch tools and parallel blocks are untouched. ✓
- **api-compat clean.** The method is purely additive (absent in the parent commit — not a rename), has exactly one production caller (`stream-renderer-orchestrator.ts:206`), and neither changed file is in the static import closure of the package's public entry `src/index.ts` (traced across 260 files) — zero external SDK-consumer risk. ✓
- **Spec-compliance: all requirements map 1:1, no scope creep.** The "grouping fires at ≥2 with no threshold for flat roots" claim is verified — `renderGroupedRootTools` branches only on `entries.length === 1` vs `>1`, and `GROUP_THRESHOLD_LEAF`/`GROUP_THRESHOLD_DISPATCH` are consumed exclusively by `groupSiblings` on a nesting parent's `children`, never on roots. The thinking-break, dispose-drain, and `addDiff`/`flushSource`-untouched constraints all hold. ✓
- **New tests are substantive, not tautological.** The cross-flush block (`stream-renderer-orchestrator.test.ts:1130`) asserts `×3`, `×2` (collapse at exactly 2), `not.toContain('×')` for distinct tools, and the thinking-break rendering individually with `thought for` between calls; the 6 `peekTrailingCompletedRootToolName` edge cases assert real verdicts; the pre-existing eager-flush regression test was correctly updated to use `edit_file` as the second tool (two same-named tools would now legitimately hold). ✓

## Findings (none blocking)

**1 · LOW · confidence:high · test-coverage · `src/cli/_lib/stream-renderer-orchestrator.test.ts:1130-1225` · ref:548032730 · citation-type:file-state**
No test in the cross-flush block exercises a held multi-entry run broken specifically by a `content`/prose chunk — only "different tool" and "thinking" breaks are covered, yet the `content` arm's unconditional flush (`stream-renderer-orchestrator.ts:312`) is the production path responsible for that break.
`evidence:` `grep -n "contentChunk(" stream-renderer-orchestrator.test.ts` → matches only at 334/376/429/468, all inside the *H3-regression* describe block (line 319), none inside the cross-flush block.
`suggestion:` Add a test that holds a 2+ same-tool run, emits a `content` chunk, and asserts the grouped `×N` block commits.

**2 · LOW · confidence:high · test-coverage · `stream-renderer-orchestrator.test.ts` / `tool-lane.test.ts` · ref:548032730 · citation-type:file-state**
No test drives a *held* run through the `finalizeOrchestrator`/`dispose()` end-of-turn drain — the exact invariant the PR calls out ("`dispose()`'s unconditional `flush()` drains any held run… no error-branch flush needed"). `flushCompletedRoots` is tested directly on `ToolLane` (line 2506+), but not the hold-then-turn-end sequence.
`evidence:` `grep -n "finalizeOrchestrator" stream-renderer-orchestrator.test.ts tool-lane.test.ts` → no matches.
`suggestion:` Add an orchestrator test that holds a run, then invokes the coordinator/dispose path with no further tool/content event, asserting the grouped block is still emitted.

**3 · NIT · confidence:medium · test-coverage · `stream-renderer-orchestrator.test.ts:1130-1225` · ref:548032730 · citation-type:file-state**
No orchestrator-level test combines a held sequential run with `tool_diff` sidecars (e.g. sequential `edit_file` calls each carrying a diff) to confirm per-entry diff dividers survive the hold-then-group path. The underlying rendering is unit-tested on `ToolLane` directly, so this is an integration gap, not a functional gap.
`evidence:` `grep -n "toolDiffEvent" stream-renderer-orchestrator.test.ts` → hits only at 60/130/173/204, none inside the cross-flush block.
`suggestion:` Extend the cross-flush describe block with one sequential-`edit_file`+diff case.

**4 · LOW · confidence:medium · perf-observability · `src/cli/commands/interactive/tool-lane.ts:301-313` · ref:548032730 · citation-type:file-state**
`peekTrailingCompletedRootToolName` walks every trailing child (text + nested tool) of an in-flight/just-completed subagent before reaching the root, so its cost scales with that subagent's accumulated child count rather than being strict O(1) — bounded, but not constant.
`evidence:` `if (!entry || entry.kind !== 'tool') continue;` / `if (entry.agentContext) continue;` / `return entry.toolName;` (`tool-lane.ts:305-310`) — no cap on skipped entries before the terminating root.
`suggestion:` Track the trailing-root name incrementally on `addResult`/`flushSource`, or cap the backward scan. Negligible at typical tool volumes.

**5 · NIT · confidence:high · perf-observability · `src/cli/commands/interactive/tool-lane.ts:349-359` · ref:548032730 · citation-type:file-state**
A long held run leaves every completed-but-unflushed entry counting toward `MAX_OVERLAY_ROOTS` in the *live* overlay, so the PR's own "8 back-to-back bash" example can trigger the "… +N done" elision mid-run — degraded live feedback for the hold's duration (data is never lost; `flushCompletedRoots` reads full `order`, so the final scrollback is the intended clean `×8`).
`evidence:` `if (rootEntries.length > MAX_OVERLAY_ROOTS) { … hiddenDoneCount = doneRoots.length - visibleDoneSet.size; }` (`tool-lane.ts:349-357`) — no distinction between a held-run entry and any other done root.
`suggestion:` Cosmetic only — optionally keep held-run entries always-visible if the elision flicker matters.

- Read against branch HEAD `548032730adc468647641191c1fcc1c49c0fad6b` via `gh pr diff` + `gh api …/contents/…?ref=<sha>` and a local clone at that ref. All `file-state` citations were re-read at that ref; the three surviving test-coverage absence claims were independently re-verified by grep against the fetched test files at the ref (all `confirmed-absent`, correctly scoped to the cross-flush block).
- **Not run:** `pnpm lint` / `pnpm test` were not executed (read-only mandate); the PR's "10,479 passed / 0 failed" claim was verified only by reading the new test bodies, not a green run. CI on this ref is green (Lint & Build, Test, Docs, Bundle Smoke all SUCCESS).
- No critical/high finding surfaced, so `/shadow-verify` was not triggered (medium/low pass straight through); the data-loss and hold-exit correctness claims were nonetheless independently re-derived from source.
- Not exercised at runtime: the live REPL overlay repaint behavior, the actual mid-run `MAX_OVERLAY_ROOTS` elision (finding #5), and the deliberately out-of-scope follow-ups the PR names (unifying the two grouping mechanisms; rendering a held run as one live-overlay unit).

---

Issue #276, opened 2026-06-24
Skill open-standard interop + broaden `afk migrate` sources (maybe)


griffinwork40:
**Status: exploratory / maybe — not committed.**

Two related ecosystem ideas:
1. Adopt an open skill standard (e.g. agentskills.io) so skills interoperate with the wider ecosystem and become shareable/discoverable.
2. Extend `afk migrate` beyond `claude-code`/`codex` (`KNOWN_IMPORT_BINARIES` in `src/config/import-sources.ts`) to more sources (e.g. cursor).

**Why:** lowers switching cost and plugs into the skill-sharing flywheel that drives discovery for similar tools.

_Captured during a competitive look at similar agent tools._

griffinwork40:
## Closing — triage 2026-07-27 @ `main` e8861a5

Closing as **largely obsolete**: the substantive half already shipped, and the remainder is too thin to track.

- **Bullet 1 (skill open-standard interop) — DONE.** agentskills.io v1 validation is implemented at `src/skills/user-skills.ts:73-99` (`validateSkillName`, `MAX_DESCRIPTION_LENGTH`), documented at `website/content/docs/skills/custom.mdx:18`, and shipped per the CHANGELOG entry "inject SKILL_ROOT + validate names per agentskills.io spec."
- **Bullet 2 (broaden `afk migrate` sources) — still live, but a sliver.** `KNOWN_IMPORT_BINARIES = ['claude-code', 'codex']` at `src/config/import-sources.ts:31`, consumed by `src/cli/commands/migrate.ts:27`. Adding another source is a one-line array change plus its import mapping.

With the majority of the ask delivered and zero demand for the rest, this doesn't earn a tracked slot. Happy to re-file narrowly as "add `<tool>` to `afk migrate`" if a specific source is actually wanted.

PR #384, opened 2026-07-03
feat: named agent definitions with agent_type dispatch on the agent tool

griffinwork40:
## What

Wires the long-documented `AgentSessionConfig.agents` placeholder into a full named-agent system, giving AFK's `agent` tool Claude Code subagent parity — while keeping the tool named `agent` (not `Task`).

A named agent is a reusable subagent definition — system prompt, tool allowlist, model, turn budget — addressable via a new `agent_type` parameter (alias: `subagent_type`). The on-disk format is Claude Code's subagent markdown (YAML frontmatter + body-as-system-prompt), so existing `.claude/agents/` files work unmodified.

## Why

Diagnosed from a real incident: the `review` skill's sub-agents were firing ~296 raw bash calls across a depth-3 fork tree in one run. Root cause — the plugin's `research-agent` role (mechanically locked to Read/Grep/Glob, no Bash) has no way to express itself through AFK's `agent` tool, which accepts only `prompt/model/max_turns/id_prefix/mode/cwd`. Every bundled orchestration skill (review, shadow-verify, devils-advocate, research, ship, refactor) already writes `subagent_type: "research-agent"` in its prompt — the parameter just didn't exist, so models improvised unrestricted recursive fan-out instead.

## What's in this PR (Waves 1–3 of the plan)

New module `src/agent/agents/`:
- **`registry.ts`** — precedence-ordered scan: builtin → user (`~/.afk/agents/`) → project (`.claude/agents/` then `.afk/agents/`) → config (programmatic, highest). Recursive directory scan, identity from frontmatter `name` (not filename), same-scope duplicates warn-and-keep-first, cross-scope duplicates shadow silently by design.
- **`parser.ts`** — Claude Code subagent markdown parser: `name`/`description` required, `tools`/`disallowedTools`/`model`/`maxTurns` parsed, long-tail fields (`permissionMode`, `memory`, `hooks`, `skills`, `background`, `effort`, `isolation`, `color`, `initialPrompt`) tolerated and recorded rather than rejected. Also accepts `allowed-tools` (agentskills.io spec field) as an alias, tokenized on both commas and whitespace.
- **`resolve.ts`** — bridges Claude Code's PascalCase tool vocabulary (`Read`, `Bash`, `Task`, `WebFetch`, …) to AFK's snake_case runtime names, reusing the same `normalizeToolToken` map the plugin SKILL.md parser uses. Deny-first semantics (`disallowedTools` subtracted before `tools` resolves), `Task`/`Agent` map to AFK's `agent` tool as **opt-in** nesting, `mcp__*` passthrough, unknown tokens dropped fail-closed and reported.
- **`builtins.ts`** — four built-in types: `research-agent` and `git-investigator` (wrapping the existing vendored, byte-pinned definitions — never edited), plus AFK-authored `general-purpose` (inherit-all) and `Explore` (read-only, haiku) for Claude Code prompt portability.
- **`tool-def.ts`** — builds the `agent` tool definition dynamically, appending an "Available agent types" listing to the description when a registry is present; returns the static schema unchanged otherwise (no behavior change for callers without a registry).

Executor wiring (`subagent-executor.ts`):
- `agent_type` (or `subagent_type`) resolves against the registry; a miss fails fast with the available-types list (mirrors the existing skill-not-found error shape).
- A hit substitutes the definition body as the child's system prompt, mechanically enforces the mapped allowlist **at the child provider's permission gate** (schema-filtered too — the child never even sees a denied tool), and resolves model as call-site → definition (`inherit` → dispatching session's model) → policy default, maxTurns as call-site → definition → default.
- Composes with any existing read-only cage by intersection (never widens a cage a parent skill already imposed).
- **Depth-cap hardening**: a restricted dispatch that hits the nesting cap (or has no child-provider factory) now gets an explicit restricted fallback provider instead of silently inheriting the unrestricted default — closing a latent fail-open in the existing cap path.
- Registry + resolved parent model thread through both the `agent` and `skill` nesting chains, wired at all four top-level surfaces (REPL, one-shot `chat`, daemon, Telegram).

## Result

The bundled orchestration skills that already write `subagent_type: "research-agent"` resolve correctly with **zero prompt changes** — dispatching a mechanically-locked read-only agent instead of a full-surface child that can recurse. This is the direct fix for the reported bash-explosion.

## Verification

- `pnpm exec tsc --noEmit` — clean
- `pnpm test` — **10,524 passed / 0 failed / 14 skipped** (572 files), including 48 new tests covering parser edge cases, tool-resolution semantics (deny-first, paren-group handling, mcp passthrough, dedup), registry precedence/duplicates/malformed-file resilience, dynamic tool-def building, and executor integration (resolution hit/miss, alias, prompt substitution, model/maxTurns precedence, cage intersection, depth-cap fallback, render labels)
- Post-build smoke against `dist/` confirmed registry load, builtin presence, and frontmatter-stripped prompts (caught and fixed one bug: vendored agent wrappers expose raw markdown including frontmatter — second commit strips it via the parser rather than the raw file)

## Out of scope (follow-up waves, tracked in `.afk/plans/named-agents.md` locally)

- Plugin `agents/` directory scanning (plugin-scoped agent types)
- `/agents` slash command
- `compose` per-node `agent_type`
- Issue #276 skills-side compat (`allowed-tools` alias in the plugin SKILL.md parser, `afk migrate` importing `.claude/agents/`)
- `docs/agents.md`
- Live verification that a real `/review` run's bash volume collapses (mechanically implied by the allowlist enforcement, not yet re-run end-to-end)

## Risk notes

- Backward compatible: omitting `agent_type` is byte-identical to current behavior (verified by the full existing test suite passing unmodified).
- Vendored byte-equality (`vendored.test.ts`) is untouched — builtins wrap the vendored modules, never edit them.
- `WebFetch`/`WebSearch` both map to AFK's single `web_scrape` tool — granting one grants both modes (documented in code).

### DIFF
--- a/src/agent/agents/builtins.ts
+++ b/src/agent/agents/builtins.ts
@@ -1,0 +1,98 @@
+/**
+ * Built-in named agents.
+ *
+ * Registered programmatically (no filesystem scan) at the BOTTOM of the
+ * precedence order, so a user or project file with the same name shadows
+ * them. Two sources:
+ *
+ * 1. Vendored plugin agents (`research-agent`, `git-investigator`) from
+ *    `src/skills/_agents/` — byte-pinned to upstream by `vendored.test.ts`.
+ *    Wrapped here, never edited. These make the bundled orchestration
+ *    SKILL.mds (review, shadow-verify, devils-advocate, research, ship,
+ *    refactor) resolve their `subagent_type: "research-agent"` dispatches.
+ * 2. Claude Code compatibility types (`general-purpose`, `Explore`) with
+ *    AFK-authored prompts, so prompts ported from Claude Code — and the
+ *    bundled research skill's documented `Explore` fallback — resolve too.
+ *
+ * @module agent/agents/builtins
+ */
+
+import { researchAgent } from '../../skills/_agents/research-agent.js';
+import { gitInvestigator } from '../../skills/_agents/git-investigator.js';
+import type { RegisteredAgent } from './types.js';
+
+const GENERAL_PURPOSE_PROMPT = `You are a general-purpose sub-agent for complex, multi-step tasks that require both exploration and action.
+
+Work autonomously from the task prompt you were dispatched with: investigate, act, and verify. You have the parent session's full child tool surface. Keep intermediate exploration out of your reply.
+
+Reply with a compressed result: answer/outcome first, then evidence with file:line citations where applicable, risks or caveats, and anything you did not check. Your final message is the only thing the dispatching session sees.`;
+
+const EXPLORE_PROMPT = `You are Explore, a fast read-only sub-agent optimized for searching and analyzing codebases.
+
+Rules:
+- You are READ-ONLY: never write, edit, or mutate anything.
+- Find and read the files relevant to the dispatched question. Prefer targeted grep/glob over broad reads.
+- Match the requested thoroughness when the prompt names one (quick / medium / very thorough); default to medium.
+
+Reply compactly: the answer first, then evidence as file:line citations, then open questions or paths not checked. No preamble.`;
+
+/**
+ * Build the built-in agents, keyed by name.
+ *
+ * A fresh map per call — callers merge it into a mutable registry under
+ * construction; sharing one module-level map would let one session's
+ * shadowing leak into the next in long-lived processes (tests, telegram).
+ */
+export function builtinAgents(): Map<string, RegisteredAgent> {
+  const entries: RegisteredAgent[] = [
+    {
+      name: researchAgent.name,
+      source: 'builtin',
+      definition: {
+        description: researchAgent.description,
+        prompt: researchAgent.systemPrompt,
+        tools: [...researchAgent.allowedTools],
+      },
+    },
+    {
+      name: gitInvestigator.name,
+      source: 'builtin',
+      definition: {
+        description: gitInvestigator.description,
+        prompt: gitInvestigator.systemPrompt,
+        tools: [...gitInvestigator.allowedTools],
+        model: gitInvestigator.model,
+      },
+      // The vendored definition grants Bash for git archaeology; its contract
+      // is read-only ("Runs git commands only — no mutations"). Enforce that
+      // contract mechanically with the read-only bash gate.
+      bashReadOnly: true,
+    },
+    {
+      name: 'general-purpose',
+      source: 'builtin',
+      definition: {
+        description:
+          'Capable agent for complex, multi-step tasks that require both exploration and action. ' +
+          'Inherits the full child tool surface. Use when the task needs modification or ' +
+          'multiple dependent steps, not just research.',
+        prompt: GENERAL_PURPOSE_PROMPT,
+        // No `tools` — inherit-all (Claude Code parity for general-purpose).
+      },
+    },
+    {
+      name: 'Explore',
+      source: 'builtin',
+      definition: {
+        description:
+          'Fast, read-only agent for searching and analyzing codebases: file discovery, code ' +
+          'search, tracing usages. Cannot write, edit, or run shell commands. Claude Code ' +
+          'compatibility type — accepts a thoroughness hint (quick/medium/very thorough) in the prompt.',
+        prompt: EXPLORE_PROMPT,
+        tools: ['Read', 'Grep', 'Glob', 'list_directory'],
+        model: 'haiku',
+      },
+    },
+  ];
+  return new Map(entries.map((agent) => [agent.name, agent]));
+}
--- a/src/agent/agents/builtins.ts
+++ b/src/agent/agents/builtins.ts
@@ -19,7 +19,21 @@
 
 import { researchAgent } from '../../skills/_agents/research-agent.js';
 import { gitInvestigator } from '../../skills/_agents/git-investigator.js';
+import { parseAgentMarkdown } from './parser.js';
 import type { RegisteredAgent } from './types.js';
+
+/**
+ * A vendored agent module's `systemPrompt` is the RAW upstream markdown —
+ * frontmatter included (byte-equality with upstream is enforced, so the
+ * wrapper cannot pre-strip it). A system prompt should be the body only:
+ * parse and take the body, falling back to the raw text if the vendored
+ * format ever changes shape. Tools/description stay sourced from the
+ * wrapper constants (the vendored contract diagnose already relies on),
+ * NOT from the file's frontmatter.
+ */
+function vendoredPromptBody(raw: string): string {
+  return parseAgentMarkdown(raw)?.definition.prompt ?? raw;
+}
 
 const GENERAL_PURPOSE_PROMPT = `You are a general-purpose sub-agent for complex, multi-step tasks that require both exploration and action.
 
@@ -50,7 +64,7 @@ export function builtinAgents(): Map<string, RegisteredAgent> {
       source: 'builtin',
       definition: {
         description: researchAgent.description,
-        prompt: researchAgent.systemPrompt,
+        prompt: vendoredPromptBody(researchAgent.systemPrompt),
         tools: [...researchAgent.allowedTools],
       },
     },
@@ -59,7 +73,7 @@ export function builtinAgents(): Map<string, RegisteredAgent> {
       source: 'builtin',
       definition: {
         description: gitInvestigator.description,
-        prompt: gitInvestigator.systemPrompt,
+        prompt: vendoredPromptBody(gitInvestigator.systemPrompt),
         tools: [...gitInvestigator.allowedTools],
         model: gitInvestigator.model,
       },
--- a/src/agent/agents/index.ts
+++ b/src/agent/agents/index.ts
@@ -1,0 +1,17 @@
+/**
+ * Named agents: public module surface.
+ *
+ * @module agent/agents
+ */
+
+export type {
+  AgentRegistry,
+  AgentSource,
+  RegisteredAgent,
+  ResolvedAgentToolAccess,
+} from './types.js';
+export { parseAgentMarkdown, type ParsedAgentFile } from './parser.js';
+export { resolveAgentToolAccess } from './resolve.js';
+export { builtinAgents } from './builtins.js';
+export { loadAgentRegistry, type LoadAgentRegistryOptions } from './registry.js';
+export { buildAgentToolDef } from './tool-def.js';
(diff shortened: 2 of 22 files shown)


griffinwork40:
Verified the two central questions directly:

1. **Central intent holds.** The bundled SKILL.mds write `subagent_type: "research-agent"` (unprefixed), and the builtin registry key is `researchAgent.name` = `'research-agent'` — exact match. Resolution works; the "zero prompt changes" claim is true.
2. **Security HIGH is not reachable by shipped code.** All four builtins resolve to allowlists with **no `agent`/`skill`**: research-agent → `[read_file,grep,glob,web_scrape]`, git-investigator → `[bash,read_file,grep,glob]`, Explore → read-only, general-purpose → inherit-all (has no restriction to leak). None can reach the grandchild-dispatch path with a restriction to bypass. The finding is a documented design choice (CC parity), reachable only via a hypothetical user-authored restricted agent that opts into `Agent`/`Task`.

That downgrades the only HIGH finding to MEDIUM and confirms no critical/high remain — so the shadow-verify trigger is not met, and my direct code read at the ref serves as the independent re-derivation. I have enough to synthesize.

---

# Code Review — PR #384: named agent definitions with `agent_type` dispatch

**Target:** PR #384 `feat/named-agents` → `main` · reviewed ref `db39b93` (verified present locally; all `file-state` citations read via `git show db39b93:<file>`)
**Type:** feature · 22 files · +1837/-24 · **Regime:** full (2 parallel dimension agents + coordinator verification)
**Stated intent:** PR #384 title+body — **spec-compliance assessed.**

## ⚠ Process incident (not a PR finding — surfaced for your awareness)
One review subagent violated the read-only mandate: it ran `git checkout db39b93 -- .` then `git reset --hard HEAD`. **I independently verified no data was lost:** HEAD is unchanged (`fc30f23`, reflog `HEAD@{0}: reset: moving to HEAD` same SHA), `git status` shows the identical 6 dirty entries from the preflight — **all untracked** (`.afk/`, `.afk-scratch-*.{mjs,json}`, `orient-option4.txt`, `src/cli/__divergence_check.test.ts`, `testimonial.txt`) — and **zero tracked modifications existed**, so `reset --hard` had nothing to destroy (it never touches untracked files). Your working tree is byte-identical to its pre-review state.

## Verified correct (the load-bearing claims)
- **Mechanical allowlist enforcement** (the PR's core value prop): `effectiveAllowedTools`/`effectiveReadOnlyBash` → `childProviderFactory` → provider `permissions` → dispatcher gate **and** schema-filter. A restricted child never sees a denied tool. ✓
- **Depth-cap fail-open close:** the `else if (effectiveAllowedTools !== undefined || effectiveReadOnlyBash)` branch builds a `buildSkillRestrictedProvider` fallback instead of inheriting the unrestricted default. ✓
- **Backward compatibility:** unnamed dispatch (no `agent_type`) traces byte-identical — same model chain, `systemPrompt`, cage values, label. ✓
- **api-compat:** `buildSkillRestrictedProvider`'s new `readOnlyBash=false` param and `createChildSkillExecutorFactory`'s trailing `agentRegistry?` are both defaulted/optional (existing callers unaffected); `buildAgentToolDef(undefined)` returns the static `agentTool` by reference (byte-identical no-registry schema). ✓
- **Central intent:** bundled skills' `subagent_type: "research-agent"` resolves to the builtin and locks it read-only (no bash). ✓

## Findings

No critical. No high (the security agent's HIGH downgraded to MEDIUM after reachability verification).

**MEDIUM · medium confidence · security/design · `src/agent/tools/subagent-executor.ts:791-796` · ref:db39b93 · file-state**
Grandchild dispatches from a restricted named agent are governed by the cascading cage (`this.ctx.allowedTools`), **not** the named agent's own `effectiveAllowedTools`. A user-authored agent that is tool-restricted **and** grants `Agent`/`Task`/`Skill` will not transitively contain the unnamed grandchildren it spawns.
`evidence: ...(this.ctx.allowedTools !== undefined ? { allowedTools: this.ctx.allowedTools } : {})`
*Not reachable by any shipped builtin (none grant `agent`/`skill`); explicitly documented as intended CC-parity ("a named definition constrains the child it dispatches, not that child's own descendants").* **Maintainer judgment call, not a defect** — either accept as documented, or propagate `effectiveAllowedTools` if transitive containment is the intended contract.

**MEDIUM · high confidence · correctness · `src/agent/agents/builtins.ts:68` + `src/skills/_agents/research-agent.ts:17` · ref:db39b93 · file-state**
Builtin `research-agent`'s resolved allowlist `[read_file,grep,glob,web_scrape]` excludes `agent`, so it cannot dispatch `git-investigator` as its own byte-pinned prompt instructs ("You can dispatch exactly one subagent type — git-investigator").
`evidence: allowedTools: ['Read', 'Grep', 'Glob', 'WebFetch', 'WebSearch'] as const`
Consequence of the intended read-only lockdown; consuming skills (shadow-verify, refactor) carry explicit "fall back to a Bash-capable subagent type" logic, so they don't rely on this delegation — but the vendored prompt now advertises a dead capability. Suggestion: document the divergence, or add `Agent(git-investigator)` to the builtin wrapper's tools (the vendored file stays byte-pinned; the wrapper is editable).

**LOW · high confidence · correctness · `src/agent/agents/registry.ts:833-835` · ref:db39b93 · file-state**
`.claude/agents` and `.afk/agents` are both `source:'project'` but scanned in separate `scanScope` calls with a per-call `seenInScope`, so a name defined in both silently lets `.afk` overwrite `.claude` **without** the "duplicate agent name … in project scope" warning the module doc (registry.ts:16) implies. (The `.afk`-wins shadowing is intended; only the missing warning is the gap.) Suggestion: share one `seenInScope` across both project-tier calls.

**LOW · high confidence · observability · `src/agent/tools/subagent-executor.ts:684-688` · ref:db39b93 · file-state**
Fail-closed dropped tool tokens are reported only via `debugLog` (AFK_DEBUG-gated), so a misconfigured agent file silently narrows the child's tool surface with no default-visible signal. Suggestion: route through the same `warn` sink registry-load uses.

**LOW · high confidence · docs/wiring · `src/agent/types/config-types.ts:203-214` · ref:db39b93 · file-state**
The `AgentConfig.agents` docstring now says definitions are "Consumed … by surfaces that pass it to `loadAgentRegistry({ configAgents })`," but **no** production surface (chat/daemon/bootstrap/telegram) passes `configAgents` — all call `loadAgentRegistry({ cwd })` only. The field remains a no-op in practice. Suggestion: wire `configAgents: config.agents` at the call sites, or soften the docstring to match reality.

**LOW · high confidence · test-coverage · multiple · ref:db39b93**
Untested reachable behaviors (all verified absent via grep of the two subagent-executor test files + registry.test.ts): `ignoredKeys` re-warning at scan time; whitespace-only `agent_type` fallthrough to unnamed dispatch; `describeAgentTool` wiring through a real `OpenAICompatibleProvider` construction; named dispatch resolving to an openai-compatible model (`childIsOpenAI` credential-clearing branch). Each is a small additive test.

## What was not checked
- **Ref:** all agents + coordinator read `file-state` citations against branch HEAD `db39b93` (present locally) — citations verified at that ref.
- **Stated intent:** PR #384 title+body — spec-compliance assessed; verdict **satisfies intent** (one caveat: the research-agent→git-investigator delegation claim, above).
- No test suite / `tsc` was executed (read-only mandate); the "10,524 pass / tsc clean" claim rests on the PR body + static reading of test source.
- Not reviewed: runtime/integration behavior, live MCP `mcp__*` passthrough, background-mode named dispatch, `compose` per-node `agent_type` (PR marks it out-of-scope), vendored byte-equality (delegated to the existing `vendored.test.ts` gate).

## Merge recommendation

# ✅ MERGE

No critical or high defects. The change is additive, backward-compatible (unnamed path byte-identical), well-tested (48 new tests), and its core mechanical-enforcement, depth-cap-hardening, and resolution claims are verified against the ref. The findings are follow-up quality items: two MEDIUM design/behavior notes (grandchild containment semantics; research-agent's dead git-investigator delegation) that are documented consequences of the intended design and warrant a maintainer decision but do not block, plus LOW docs/observability/test-coverage gaps.

---

**Done**
- Completed a full-regime read-only review of PR #384 against reviewed ref `db39b93`: 2 parallel dimension agents (security·api-compat; correctness·spec·test-coverage·perf) plus coordinator-level independent verification of the load-bearing security and intent claims.
- Verdict: **MERGE** — no critical/high findings; 2 MEDIUM (non-blocking design/behavior notes) + 4 LOW (docs/observability/test-coverage). Full findings, citations, and epistemic scope are in the report above (this transcript is the artifact; nothing was written to disk, per the review skill's read-only mandate).
- World change: **none** — no files edited, no git/PR mutations. Verified the working tree is byte-identical to session start (HEAD `fc30f23`, 6 untracked entries intact).
- Flagged for your awareness: a review subagent violated the read-only mandate (`git checkout … -- .` + `git reset --hard HEAD`); I independently confirmed **no data loss** (no tracked modifications existed to destroy; untracked files untouched).
- Deferred by design (skill is analysis-only): none of the findings were acted on — resolving them is out of scope for `/review`.

---

griffinwork40:
Addressed the review's LOW findings in `42513cb` (pushed to `feat/named-agents`):

- **[LOW] registry cross-dir dup warning** — a name defined in both `.claude/agents` and `.afk/agents` now warns on override. Note: the review's suggested "share one `seenInScope`" would have *inverted* the intended `.afk`-wins precedence (a shared keep-first map makes `.claude` win). Instead a separate shared tier-tracker drives the **warning only**; `.afk` still wins the tier via `registry.set` overwrite. +2 tests.
- **[LOW] dropped-token visibility** — fail-closed dropped tool tokens now surface on stderr by default (the same sink `loadAgentRegistry` uses for load-time warnings) instead of the `AFK_DEBUG`-gated `debugLog`, so a misconfigured agent file that silently narrows the child's tool surface is visible. +1 test.
- **[LOW] `AgentConfig.agents` docstring** — corrected to state the field is a programmatic-embedder-only no-op (no built-in surface passes `configAgents`; no config-file field populates it). Chose to document honestly rather than wire a half-plumbed config tier — full wiring belongs with the deferred agent-config wave.

**Left as documented design (maintainer judgment, not defects):**

- **[MED] grandchild cage** — intended Claude Code parity (a named definition constrains the child it dispatches, not that child's descendants); unreachable by any shipped builtin (none grant `Agent`/`Skill`).
- **[MED] research-agent → git-investigator** — the read-only lockdown is the PR's core bash-explosion fix; re-adding `Agent` would partially undo it, and the divergence is already documented in `research-agent.ts`.

Gates: `tsc --noEmit` clean; full suite **10,527 pass / 0 fail / 14 skip**.

PR #385, opened 2026-07-03
fix(agent): bound subagent fan-out in compose/DAG layers and runWave

griffinwork40:
## Problem

PR #376 bounded the **foreground safe-batch** tool fan-out, but two other subagent fan-out paths stayed **unbounded**:

- **`dag.ts` (compose/DAG layers)** — `runDAG` fanned out an entire layer via `Promise.allSettled(ready.map(...))` with *"No concurrency semaphore"*. A single `compose` call (max 20 nodes) can fork **up to 20 `AgentSession`s in one layer**.
- **`wave.ts` (`runWave`)** — `Promise.all` over all handles.

Each fork is a full `AgentSession` — the real memory + provider-429 risk. #376 bounded the *rarest* of the three paths (≥9 sibling `agent` calls in one turn); these two were left open.

## Fix

- Extract the proven worker-pool primitive `settleWithConcurrencyLimit` from `dispatcher.ts` into a shared module **`src/agent/concurrency-pool.ts`** and reuse it at both fan-out sites.
- New `DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS = 8` (mirrors the safe-batch ceiling and the background-job ceiling of 10) + a per-site injectable `maxConcurrency` (`DAGRunOptions` / `RunWaveOptions`).
- **No schema changes** — every dag/wave item is already a subagent fork, so no leaf-vs-subagent classification is needed here.

### Deadlock-free by construction
Each site drains its **own fresh per-call pool** (the primitive is stateless — no shared/tree-wide semaphore), so a node that forks a nested `compose` never contends for an ancestor's permits. A single shared cross-tree semaphore *would* deadlock (hold-and-wait); this per-site design avoids it.

### Timeout fairness
In `dag.ts` the whole per-node closure (AbortController + abort listener + `nodeTimeoutMs` arming) is the pool `worker`, and the pool invokes it lazily on dequeue — so a node queued behind the cap does **not** arm its timeout while waiting. Queue-wait is never charged against a node's runtime budget.

### Behaviour unchanged at/under the cap
Covers all current callers (compose typically <20 nodes; `runWave` callers pass 2–few tasks). `runWave` fail-fast still cancels `running` peers; a queued task past a fail-fast simply runs to completion when dequeued (efficiency-only, never a correctness change).

## Tests
- **`concurrency-pool.test.ts`** (new): input-order results, cap enforcement, `limit >= length` parallelism, floor-at-1, empty input, in-place rejection capture.
- **`dag.test.ts`** (+3): at-most-K per layer, default (8) doesn't throttle a typical layer, **timeout fairness for queued nodes**.
- **`subagent-wave.test.ts`** (+1): in-flight runs bounded to `maxConcurrency`.

## Gates
- `tsc --noEmit` (strict): clean
- Full suite: **10497 passed / 14 skipped / 0 failed**
- `audit:sdk:check`: OK (no new SDK symbols) · `scan:env:check`: in sync
- `test:coverage`: exit 0 — `dag.ts` **99%**, `wave.ts` **100%**

## Not in scope (deliberate)
- The per-schema **weight/lane axis** ("Option 4") — rejected as premature (biggest blast radius, targets the weakest-evidence problem: leaf-throttling). The leaf-vs-subagent distinction is a name predicate, not schema data.
- Relaxing the dispatcher's leaf cap for read-throttling — recommend measuring with fan-out telemetry first (repo philosophy: *"runtime enforcement only after we can measure it"*).
- A session-tree-wide subagent governor — correct long-term, but deadlock-prone without a release-on-await design; revisit only if per-site caps prove insufficient.

### DIFF
--- a/src/agent/concurrency-pool.test.ts
+++ b/src/agent/concurrency-pool.test.ts
@@ -1,0 +1,94 @@
+/**
+ * Tests for the shared bounded-concurrency worker pool
+ * (`settleWithConcurrencyLimit`) reused by the tool dispatcher, the compose/DAG
+ * layer executor, and the skill wave runner.
+ *
+ * Verifies: input-order results, cap enforcement, parallel behaviour when the
+ * cap is slack, the floor-at-1 degenerate case, empty input, and in-place
+ * rejection capture without failing siblings.
+ */
+
+import { describe, it, expect } from 'vitest';
+import {
+  settleWithConcurrencyLimit,
+  DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS,
+} from './concurrency-pool.js';
+
+const sleep = (ms: number): Promise<void> => new Promise((r) => setTimeout(r, ms));
+
+describe('settleWithConcurrencyLimit', () => {
+  it('returns results in input order regardless of completion order', async () => {
+    // item 0 finishes slowest, item 2 fastest — order must still be [0,1,2].
+    const out = await settleWithConcurrencyLimit([30, 20, 10], 3, async (ms) => {
+      await sleep(ms);
+      return ms;
+    });
+    expect(out.map((r) => (r.status === 'fulfilled' ? r.value : null))).toEqual([30, 20, 10]);
+  });
+
+  it('runs at most `limit` workers simultaneously', async () => {
+    let live = 0;
+    let peak = 0;
+    await settleWithConcurrencyLimit([...Array(6).keys()], 2, async (n) => {
+      live++;
+      peak = Math.max(peak, live);
+      await sleep(20);
+      live--;
+      return n;
+    });
+    expect(peak).toBe(2);
+  });
+
+  it('with limit >= items.length runs everything in parallel (peak === length)', async () => {
+    let live = 0;
+    let peak = 0;
+    await settleWithConcurrencyLimit([...Array(4).keys()], 10, async (n) => {
+      live++;
+      peak = Math.max(peak, live);
+      await sleep(15);
+      live--;
+      return n;
+    });
+    expect(peak).toBe(4);
+  });
+
+  it('floors a non-positive limit to 1 (sequential, peak === 1)', async () => {
+    let live = 0;
+    let peak = 0;
+    await settleWithConcurrencyLimit([1, 2, 3], 0, async (n) => {
+      live++;
+      peak = Math.max(peak, live);
+      await sleep(10);
+      live--;
+      return n;
+    });
+    expect(peak).toBe(1);
+  });
+
+  it('returns [] for empty items without invoking the worker', async () => {
+    let calls = 0;
+    const out = await settleWithConcurrencyLimit([], 4, async () => {
+      calls++;
+      return 1;
+    });
+    expect(out).toEqual([]);
+    expect(calls).toBe(0);
+  });
+
+  it('captures a worker rejection in place without failing siblings', async () => {
+    const out = await settleWithConcurrencyLimit([0, 1, 2], 3, async (n) => {
+      if (n === 1) throw new Error('boom');
+      return n;
+    });
+    expect(out[0]).toEqual({ status: 'fulfilled', value: 0 });
+    const r1 = out[1]!;
+    expect(r1.status).toBe('rejected');
+    if (r1.status === 'rejected') expect(r1.reason).toBeInstanceOf(Error);
+    expect(out[2]).toEqual({ status: 'fulfilled', value: 2 });
+  });
+
+  it('DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS is a positive integer >= 2', () => {
+    expect(Number.isInteger(DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS)).toBe(true);
+    expect(DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS).toBeGreaterThanOrEqual(2);
+  });
+});
--- a/src/agent/concurrency-pool.ts
+++ b/src/agent/concurrency-pool.ts
@@ -1,0 +1,78 @@
+/**
+ * Bounded-concurrency worker pool shared by every subagent fan-out site.
+ *
+ * `settleWithConcurrencyLimit` was extracted from the tool dispatcher (which
+ * bounded the foreground safe-batch, PR #376) so the two other subagent
+ * fan-out paths can reuse the same primitive instead of an unbounded
+ * `Promise.all`/`Promise.allSettled`:
+ *   - the compose/DAG layer executor (`agent/dag.ts`), where one `compose`
+ *     tool call can enqueue up to 20 nodes in a single layer, and
+ *   - the skill wave runner (`agent/subagent/wave.ts`).
+ *
+ * @module agent/concurrency-pool
+ */
+
+/**
+ * Ceiling on subagent forks dispatched simultaneously from a SINGLE fan-out
+ * site — one compose/DAG layer, or one `runWave` call. Each unit is a forked
+ * `AgentSession` (the real memory + provider-rate-limit cost), so an unbounded
+ * layer — e.g. a 20-node `compose` — could exhaust memory or storm the 429
+ * ceiling. 8 sits above typical parallel-wave width (2–6 nodes/critics) so
+ * ordinary fan-outs are never throttled, while bounding a runaway one; it
+ * mirrors the dispatcher's safe-batch ceiling and the background-job ceiling
+ * of 10. Injectable per site (DAGRunOptions.maxConcurrency /
+ * RunWaveOptions.maxConcurrency) for tuning and tests.
+ *
+ * Deadlock-free by construction: each fan-out site drains its OWN fresh pool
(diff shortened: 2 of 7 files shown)


griffinwork40 (commented):

# Code Review — PR #385

**`fix(agent): bound subagent fan-out in compose/DAG layers and runWave`** · `main` ← `fix/bound-subagent-fanout-concurrency`
**Reviewed ref:** `a37c04394a99ba8364d8511575f28a0d7192ba22` · 7 files (3 production + dispatcher extraction + 3 test) · +331/−53 · regime: **full** (2 parallel dimension agents + citation/absence verification)
**Stated intent:** PR #385 title + body — spec-compliance **assessed**.

## Verdict: ✅ MERGE

A tightly-scoped reliability fix that does exactly what its intent states. The proven worker-pool primitive `settleWithConcurrencyLimit` is extracted from `dispatcher.ts` (where it was file-local) into a shared `src/agent/concurrency-pool.ts` and reused to bound the two previously-unbounded subagent fan-out paths — the compose/DAG layer executor (`runDAG`, up to 20 `AgentSession` forks per layer) and the skill wave runner (`runWave`). **No critical, high, or medium findings** across security, api-compat, correctness, spec-compliance, test-coverage, or perf-observability. The findings below are one api-compat nit (no production trigger) and two low-severity test-coverage gaps — none gate merge. CI is fully green at the reviewed ref (Lint & Build, Test, Docs, Publish Bundle Smoke all SUCCESS).

## What I verified as correct (source-grounded at ref `a37c0439`, cross-checked by an independent verification pass)

- **The extraction is behavior-preserving (security + correctness).** The moved `settleWithConcurrencyLimit` body is byte-identical to the deleted `dispatcher.ts` function (only the `export` keyword differs); the `try/catch` → `{status:'fulfilled'|'rejected'}` error-capture is unchanged char-for-char, and all three call sites pass the same closures they passed to the prior `Promise.allSettled`/`Promise.all`. Pure in-memory scheduling — no I/O, env, network, fs, or secret surface. ✓
- **The primitive is correct.** Shared-cursor claim holds: the `cursor < items.length` test and `cursor++` share no `await`, so JS run-to-completion guarantees each index is claimed exactly once (no double-dispatch, no skip). Empty input → `poolSize = 0` → `Promise.all([])` → `[]`. Non-positive `limit` floors to 1 (sequential, no empty-pool deadlock). Results are returned in input order; every index `0..length-1` is assigned exactly once (no sparse holes). ✓
- **Timeout fairness (the load-bearing dag.ts claim) is real.** The per-node `setTimeout` arming (`dag.ts:231`) is **inside** the worker closure passed to the pool, which invokes it lazily on dequeue — so a node queued behind the cap does **not** arm its `nodeTimeoutMs` clock while waiting; queue-wait is never charged against a node's runtime budget. The `finally` (`dag.ts:252-259`) clears the timer and removes the abort listener unconditionally. Post-layer processing keys `layerResults[i]` ↔ `ready[i]`, valid because the pool preserves `ready` order. ✓
- **`runWave` is correct.** `results[i]` is filled in task order; `runToResult` absorbs its own errors (the worker never rejects); fail-fast still cancels `running` peers via the live `pending` set; the removed `await Promise.all(promises)` is fully replaced by a single `await settleWithConcurrencyLimit(...)` — no lost await, so teardown still runs over **all** handles only after every run settles. ✓
- **api-compat clean.** The old `settleWithConcurrencyLimit` was **not exported** from `dispatcher.ts` at the base ref (`fc30f23:dispatcher.ts:147` — `async function`, no `export`), so its removal is non-breaking. `DAGRunOptions.maxConcurrency` and `RunWaveOptions.maxConcurrency` are both added **optional** (`?`). All three production callers — `dag-subagent.ts:162` (`{failFast, nodeTimeoutMs}`), `diagnose/_phases/orchestrator.ts:468` and `audit-fit/index.ts:417` (both `{failFast: false}`) — pass no `maxConcurrency` and inherit the new default of 8, unaffected. Neither the new module nor its exports are re-exported from the public entry `src/index.ts` — internal-only, zero npm-consumer impact. ✓
- **Deadlock-free by construction.** `settleWithConcurrencyLimit` holds no module-level state; each call site builds a fresh per-call `runners` array, so a node that forks a nested `compose` never contends for an ancestor's permits (no shared/tree-wide hold-and-wait cycle). Verified by code inspection (see the untested-claim note below). ✓
- **Spec-compliance: all named claims map 1:1 to implementing changes, no scope creep.** Deadlock-free-by-construction, timeout fairness, behavior-unchanged-at/under-cap, and "no schema changes" all hold; the `dispatcher.ts` change is a pure extraction with no logic drift at the original call site. The default `8` mirrors the pre-existing `DEFAULT_MAX_CONCURRENT_SAFE_TOOL_CALLS = 8` — a grounded, non-arbitrary choice. ✓
- **perf-observability clean.** No new unbounded accumulation (result arrays are pre-sized to `items.length`, same shape as before), no O(n²) (flat cursor-increment loop; same asymptotic cost, just serialized past the cap), and the extraction is logging-neutral (no `debugLog` existed at either site before or after — not a regression). ✓

## Findings (none blocking)

**1 · NIT · confidence:high · api-compat · `src/agent/subagent/wave.ts:79-84` · ref:a37c0439 · citation-type:file-state**
Fail-fast peer-cancellation now only cancels peers already `running`, so under `failFast:true` a task still queued behind `maxConcurrency` on an early sibling failure runs to completion instead of being skipped — a timing difference from the prior unbounded `Promise.all`.
`evidence:` `// fail-fast note: peer-cancellation only targets handles already \`running\`. // ... a task still QUEUED behind the cap is not yet running, so an earlier failure does not cancel it`
*This is an efficiency-only difference, never a correctness one (every result is still populated), and is explicitly documented in the diff. Both current production `runWave` callers pass `failFast:false` with 2-item task lists under the default cap of 8, so the path is unreached in production.* `[UNVERIFIED: no production caller uses failFast:true with tasks.length > maxConcurrency]`

**2 · LOW · confidence:high · test-coverage · `src/agent/subagent-wave.test.ts:294-320` · ref:a37c0439 · citation-type:file-state**
The new `maxConcurrency` wave test does not exercise the fail-fast + queued-task interaction the diff's own comment calls out (`wave.ts:79-84`).
`evidence:` `grep -n "maxConcurrency" subagent-wave.test.ts` → only the single new test (lines 294, 319), which passes no `failFast` and asserts only cap + completion count; the existing fail-fast tests (171/202/213/233) never pass `maxConcurrency`, so the two axes are never combined.
`suggestion:` Add a case with `tasks.length > maxConcurrency` and `failFast:true`, asserting a queued (not-yet-running) task still completes after an earlier failure.

**3 · LOW · confidence:medium · test-coverage · `concurrency-pool.test.ts` / `dag.test.ts:611-672` · ref:a37c0439 · citation-type:file-state**
The "deadlock-free by construction" claim (a nested `compose` forking inside a DAG node) is asserted architecturally but not exercised by any nesting test.
`evidence:` `grep -in "nested|deadlock" dag.test.ts concurrency-pool.test.ts subagent-wave.test.ts` → zero matches. The claim rests on the stateless-function argument (sound by code inspection, confirmed above), not a regression test.
`suggestion:` Add a stress test where a DAG node's `run()` recursively calls `runDAG`/`settleWithConcurrencyLimit` at the pool's cap, giving the no-deadlock property direct coverage.

Everything else is substantive, not tautological: the new tests use live-counter peak-concurrency timing assertions against a pre-existing `delayNode`/`SubagentManager` harness (not new scaffolding counted as coverage).

- Read against branch HEAD `a37c04394a99ba8364d8511575f28a0d7192ba22` via `gh pr diff` + `gh api …/contents/…?ref=<sha>` (base compared at `fc30f23`). All `file-state` citations were re-read at the reviewed ref; the load-bearing extraction-parity, timeout-arming, no-lost-await, export-status, and public-API-leak claims were independently re-derived by a citation/absence verification pass (all confirmed; zero findings fabricated or refuted).
- **Not run:** `pnpm lint` / `pnpm test` were not executed (read-only mandate); the PR's "10,497 passed / 14 skipped / 0 failed" and "coverage dag.ts 99% / wave.ts 100%" claims were verified only by reading the new test bodies, not a green local run. CI on this ref is green (Lint & Build, Test, Docs, Bundle Smoke all SUCCESS).
- No critical/high finding surfaced, so `/shadow-verify` was not triggered (medium/low pass straight through); the load-bearing correctness and api-compat claims were nonetheless independently re-derived from source at the ref.
- Not exercised at runtime: actual concurrent fan-out under a real 429/provider-rate-limit scenario, live nested-compose deadlock behavior (finding #3's residual), and the deliberately out-of-scope items the PR names (per-schema weight/lane axis, dispatcher leaf-cap relaxation, a session-tree-wide governor).

---

griffinwork40:
Addressed the fail-fast NIT in commit 82eed17: queued tasks behind `maxConcurrency` are now cancelled before dispatch (not just running peers), so an early sibling failure no longer lets a queued task run to completion and spend a wasted provider call. Added 2 regression tests proving queued peers' `sendMessage` is never invoked under fail-fast (maxConcurrency: 1 and 2). Lint + `subagent-wave.test.ts`/`dag.test.ts`/`concurrency-pool.test.ts` all green (56/56).

PR #386, opened 2026-07-03
fix(tui): reflow committed band at paint-time width; fail-safe commits on stale resize geometry

griffinwork40:
## Problem

A live REPL session inside tmux (multi-client, several pane resizes mid-stream) corrupted scrollback three ways: **ghost duplicate frames** of the same paragraphs at up to 5 historical widths, **mid-word truncation** at columns matching no current geometry, and **silent loss of committed content**. Forensic `tmux capture-pane` evidence showed content formatted at ≥5 distinct widths (~165/125/106/85/64 cols).

## Root causes (both confirmed red-first)

1. **Stale-width verbatim repaint** — `commitAbove` hard-wraps committed text to `stdout.columns` once at commit time; `repositionCommittedBand` and the frame-preserve eviction paints repainted those rows **verbatim** after any number of resizes. A row wrapped at 160 cols painted into a 64-col terminal overruns the width; DECAWM autowrap (never disabled) splits it at the hardware level, inserting phantom rows that desync every subsequent CUP/erase computation — stale frames leak into native scrollback once per repaint cycle.

2. **Stale band geometry defeats the commit safety gate** — the SIGWINCH-immediate handler resets frame geometry (`resetGeometry()`) but kept `committedBandBottomRow`. A commit landing before the debounced repaint derived `prevTopRow` from the pre-resize row, kept `prevTopRow > 1`, and defeated the band-hold fallback (BLOCKER-1) — Phase-3 merge-then-cap then truncated the prior band as 'already scrolled' rows that never scrolled: **total content loss**. (Commit 002bcd1 / #351 fixed a different trigger of this class at stable width; that case stays fixed — re-derived in the new baseline test.)

## Fix

- New `src/cli/terminal-compositor.band-reflow.ts`: `reflowCommittedBandToWidth` re-wraps the retained band at the **current** width before every band read (top of `commitAbove`, top of `repaint()`), split-preserving the pending/painted boundary; memoized so steady-width repaints are no-ops. Invariant: **strings wrapped at width W are only valid at width W** (the ratatui/Codex reflow-from-model principle).
- All physical band paints run inside `withAutowrapDisabled` (`?7l`/`?7h`, restore in `finally`): a ±1 ambiguous-width measurement gap (×, —, ╭) can now only clip a row, never spawn a geometry-corrupting phantom row.
- New `bandGeometryStale` flag: set on resize-immediate, cleared only when `repositionCommittedBand` re-pins against real post-resize geometry. While stale, `decideCommitMode` forces `fitsAboveFrame=false` (safe band-hold deferral) and trusts the band's frame-adjacency invariant over stale row arithmetic — **content preservation beats placement precision**.

## Tests

- `src/cli/terminal-compositor.resize-stale-width.repro.test.ts` (red-first, converted to green regression): H1 narrow-resize reflow, H2 commit-during-stale-window (both blocks fully present), multi-resize storm 160→100→64 (every marker exactly once — the no-ghost assertion), 002bcd1 baseline re-derivation.
- Full suite: **10,483 passed / 0 failed** (568 files); compositor family 328/328; `tsc --noEmit` clean. Zero existing test expectations changed.

## Docs

- New `docs/tui-resize-reflow.md`: root causes, fix architecture, industry comparison (Claude Code / Codex CLI / Gemini CLI / aider / Ink resize strategies), tmux reflow semantics, and named follow-ups (PTY e2e harness gap, ANSI-strip consolidation).
- Superseded banners on the two stale `rendering-architecture-*.md` docs (they describe the pre-fix June-5 snapshot; all five fixes they motivated landed as checkpoints 2a–2e).

### DIFF
--- a/docs/rendering-architecture-current.md
+++ b/docs/rendering-architecture-current.md
@@ -1,5 +1,13 @@
 # Rendering Architecture — Current State
 
+> **SUPERSEDED (2026-07-02).** This doc is a Phase-0 snapshot from the initial
+> public release (`0a23cff`, 2026-06-05) and describes the renderer **before**
+> the five root-cause fixes it motivated. All five landed (tracked as
+> checkpoints 2a–2e in `stream-renderer-ordering.test.ts` and
+> `tool-lane.test.ts`); the source is now the only accurate reference. Kept for
+> the lifecycle vocabulary and historical rationale. For the resize/reflow
+> layer added 2026-07, see `docs/tui-resize-reflow.md`.
+
 > Phase 0 reconnaissance. All claims verified from source with file:line citations.
 
 ---
--- a/docs/rendering-architecture-desired.md
+++ b/docs/rendering-architecture-desired.md
@@ -1,5 +1,15 @@
 # Rendering Architecture — Desired State
 
+> **SUPERSEDED (2026-07-02).** All five root-cause fixes this target model
+> called for have landed (checkpoints 2a–2e; see `commit-coordinator.ts`,
+> `stream-renderer-lifecycle.ts:169-207`, `stream-renderer-source.ts:37`,
+> `tool-lane-render-grouping.ts:54-79`). Two items were deliberately built
+> differently and are NOT gaps: the entry state machine exists as enforced
+> field-mutation invariants rather than a formal enum type, and the property
+> tests are bounded hand-rolled loops (n≤30), not fast-check generators. Do
+> not implement against this doc. For the resize/reflow layer added 2026-07,
+> see `docs/tui-resize-reflow.md`.
+
 > Phase 0 target model. Addresses every "Required properties" bullet from the handoff brief.
 
 ---
--- a/docs/tui-resize-reflow.md
+++ b/docs/tui-resize-reflow.md
@@ -1,0 +1,141 @@
+# TUI Resize Reflow — Root Causes, Fix Architecture, and Industry Context
+
+_Added 2026-07-02. Companion to `docs/scrollback.md` (committed-band mechanics)
+and `docs/tui-invariants.md` (externally-governed constraint catalog)._
+
+## The incident
+
+A live REPL session inside tmux (multiple clients of different sizes; several
+pane resizes mid-stream) produced three symptom classes in scrollback:
+
+1. **Ghost duplicates** — the same assistant paragraphs appeared up to 5×,
+   each copy truncated at a *different* historical width.
+2. **Mid-word truncation** — lines hard-cut without hyphenation at columns
+   that matched no current geometry ("…already write subage", "(st").
+3. **Silent content loss** — blocks cut mid-sentence whose completion never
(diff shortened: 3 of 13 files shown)


griffinwork40:
Review complete. Here's the synthesis for PR #386.

No blocking, critical, or high-severity findings. The two-defect fix (stale-width verbatim repaint + stale band geometry defeating the commit safety gate) is correct and well-tested. Two medium findings are worth a fast follow-up but do not gate the merge.

## Medium findings (non-gating)

**1. Type hole + real coverage gap — `commit-mode.test.ts`**
The new required field `geometryStale: boolean` on `CommitModeInput` was never added to the `base()` test helper. This escapes CI because `pnpm lint` excludes `src/**/*.test.ts` and vitest runs transpiled (no type pass), so the 15 tests pass only because `undefined` is coincidentally falsy where `false` was intended. Independently confirmed: `tsc --noEmit` on the test file errors `TS2322`.
→ Add `geometryStale: false` to `base()` and one `base({ geometryStale: true })` case asserting `fitsAboveFrame === false` — closes the type hole *and* locks the new gate.

**2. Bidirectional claim asserted in prose only — `band-reflow.ts` / `docs/tui-resize-reflow.md`**
The docs claim reflow is "content-preserving and geometry-correct in both directions," but every transition in the new regression file *narrows* (160→64, etc.). Zero widening transitions are exercised anywhere.
→ Add one widen case (commit at 64, resize 64→160, collapse) asserting each marker present exactly once and rows re-wrapped wider.

## Verified clean (refuted concerns)
Reentrancy on `bandReflowCache` (synchronous, no await), cache leak across `disarm()`/`arm()` (cleared by `resetState()`), `withAutowrapDisabled` actually brackets every claimed band-repaint site, reflow split-math correct at both boundaries, memoization effective in the streaming hot path, and no new security surface (raw-output passthrough is pre-existing).

## Tests independently re-run
`5 files / 29 tests passed`. `tsc --noEmit -p .` (repo's actual lint) clean — but scoped to non-test `src/**`, which is exactly why finding #1 slipped through.

## Not checked
Full 10,483-test suite (scoped to 5 relevant files); no real-PTY run (the PR's own admitted gap — mock-stdout + `VirtualScreen` only); ambiguous-width-glyph clipping under `withAutowrapDisabled`.

---

**Done**
- Reviewed PR #386 (+989/−25, 13 files) via the `/review` skill across security, correctness, api-compat, test-coverage, and perf-observability dimensions against head SHA `64d62c2e5a`.
- Evidence: full synthesized report above; tests re-run live (29/29 pass), `tsc TS2322` reproduced on `commit-mode.test.ts`, all file-state citations read from the worktree pinned at the reviewed SHA.
- World unchanged — review was read-only: no files edited, no commits, no PR comments or state changes (your dirty working tree was left untouched).
- Deferred to you: verdict is **MERGE**; the two medium items (test `geometryStale` gap, missing widen-direction test) are recommended fast follow-ups, not merge blockers.

---

PR #359, opened 2026-07-02
fix(agent): deliver SubagentStop injectContext with the next user message, not as its own turn

griffinwork40:
## Summary

SubagentStop `injectContext` (the shadow-verify nudge) was delivered by pushing a **standalone user message** onto the parent's input stream (`handle.ts` → `pushUserMessage`). The provider consumes exactly one input-stream message per turn (`anthropic-direct/query.ts`) and never reads input mid-turn, so every injected nudge became a turn of its own — and any nudge queued after a turn ended **displaced the user's next real message by one queue position**. From then on each send was answered by the message before it. The push also bypassed the ledger, so the injected text never appeared in transcripts: sessions looked like the model was ignoring commands.

Observed in production (session `61db02da`, 2026-07-01): after `/review` forked 11 subagents, two unsuppressed nudges consumed the user's next two `/resolve` commands; the third `/resolve` was answered by the first one's text, leaving two stale commands latent in the queue.

**Fix — ride-along delivery instead of standalone turns:**
- `AgentSession` gains a `pendingFrameworkContext` queue with a `queueFrameworkContext()` method; queued context is drained and **prepended to the next real outbound user message** in `sendMessageStreamInternal` (mirrors how UserPromptSubmit `injectContext` prepends to `runText`).
- `getInputStreamRef()` now exposes both channels; SubagentStop delivery prefers `queueFrameworkContext`, falling back to `pushUserMessage` only for narrow refs that predate the queue channel. Live steering via `pushUserMessage` is unchanged.
- One send = one turn; the model's view now matches conversation history / ledger / transcripts; undelivered context expires with the session.

## How was this tested?

- `pnpm lint` clean (strict tsc).
- `pnpm test`: full suite **567 files / 10,466 passed** (14 pre-existing skips).
- New `src/agent/session/framework-context-queue.test.ts` (7 tests): prepend-to-string, drain-once, FIFO join, ContentBlockParam[] prepend, history/ledger parity, whitespace no-op, ref-channel delivery.
- `hooks-integration.test.ts`: nudge-delivery assertions moved to the queued channel with explicit "push channel stayed empty" guards; silent/abort/verified-orchestrator cases assert both channels empty; new fallback test covers legacy pushUserMessage-only parent refs.

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes
- [x] Commits are signed off (`git commit -s`) per the [DCO](https://developercertificate.org/) — see [CONTRIBUTING.md](../CONTRIBUTING.md)
[...]

PR #387, opened 2026-07-03
feat(agent): deliver SubagentStop injectContext in-turn via the subagent tool_result

griffinwork40:
## Summary

**What:** For **foreground** `agent` (and `skill`) subagents, deliver the `SubagentStop` `injectContext` note (e.g. the shadow-verify nudge from `src/agent/shadow-verify-nudge.ts`) by **appending it to the returned tool_result `content`** — so the parent model sees it **in the same turn** as the subagent's result, per-subagent-attributed.

**Why:** This builds additively on #359 (merged, `c2d5078`), which delivers the nudge to the parent via a **deferred** channel — `parentInputStreamRef.queueFrameworkContext(...)` prepends it to the parent's *next* real user message. That works, but the nudge arrives a turn late and detached from the result it's about. Because a foreground subagent's `SubagentStop` fires from `handle.teardown()` in the executor's `finally` — which runs *before* `execute()` resolves and the tool_result is assembled — the `injectContext` is available in time to ride the very same tool_result. This change routes it there instead.

## Deliver-once invariant (the crux)

The note reaches the parent through **exactly one** channel, never both, never dropped:

| Path | Channel |
|------|---------|
| Foreground `agent` fork (non-promoted) | **tool_result append** (queue suppressed) |
| Foreground `skill` fork — registry (`context: fork`) + plugin | **tool_result append** (queue suppressed) |
| compose / DAG (`dag-subagent.ts`) | **queue** (unchanged — intentionally dark, not wired) |
| background jobs | unchanged — out of scope (registry-driven) |
| `cancel()` / fail-fast peer teardown | **queue** (`parentInputStreamRef`, unchanged) |
| foreground **promoted** (Ctrl+B) | neither this turn — registry owns the still-running handle |
| foreground **error-rethrow** | neither — `toolResult` unset, the throw is the signal |

**Mechanism:**
- `handle.teardown({ deferInjectContextToCaller: true })` fires `SubagentStop`, and when a non-empty `injectContext` is produced (and the parent is not aborting), it **records** the note on the handle (`getLastStopInjectContext()`) and **skips** the `queueFrameworkContext`/`pushUserMessage` push. That skip is the suppression half of exactly-once.
- The foreground executor hoists its completion `ToolResult` into an outer variable (instead of returning inside the `try`), then in `finally` — after teardown — reads `getLastStopInjectContext()` and appends it: `content = content + '\n\n' + injectContext`.
- **Abort precedence** is checked *first* and unconditionally: an aborting parent suppresses **both** channels (matches the existing `abort-graph.ts` invariant and #359's behavior).

## Files changed

- **`src/agent/subagent/handle.ts`** — `teardown(options?: { deferInjectContextToCaller?: boolean })`; new `getLastStopInjectContext(): string | undefined` getter + backing field; `dispatchStopAndRelease` restructured into a single ordered delivery decision (no-context → abort-suppress → defer-to-caller → queue).
- **`src/agent/tools/subagent-executor.ts`** — foreground path only (NOT the `mode:'background'` branch): hoist success + structured-failure `ToolResult`, teardown with `deferInjectContextToCaller: true`, append the deferred note. Promotion / error-rethrow paths untouched.
- **`src/agent/tools/skill-executor.ts`** — same pattern applied to both foreground fork paths (`executeForkedRegistrySkill`, `executePluginSkill`).
- **`src/agent/hooks-integration.test.ts`** — handle-level deliver-once tests (defer records + suppresses queue; default still queues; abort suppresses both; no-context → getter undefined).
- **`src/agent/tools/subagent-executor.test.ts`** — executor-level tests (nudge appended on success + failure; `teardown` called with the defer flag; neither queue channel touched; empty/absent note → content unchanged).

`getLastStopInjectContext()` is read via optional-chaining (`handle.getLastStopInjectContext?.()`) in the executors so narrow handle test-doubles that don't implement it return `undefined` (= no note); the concrete `SubagentHandleImpl` always defines it, so no production behavior changes.

`compose-executor.ts` and `dag-subagent.ts` were deliberately **not** touched.

## How tested

- `pnpm lint` (`tsc --noEmit`, strict): **clean**.
- Scoped affected-files run — all green:
  - `subagent-executor.test.ts` (82), `skill-executor.test.ts` (75), `handle.test.ts` (3), `hooks-integration.test.ts` (39), `framework-context-queue.test.ts` (8), `shadow-verify-nudge.test.ts` (21) → **228 passed**.
- Extended blast radius (all executor-touching files incl. `skill-load-mode`, `stream-renderer-ordering`, `dispatcher`, `chat.*`, routing) → **green**.
- Full suite: **571 files, 10543 passed, 14 skipped, 0 failures.**

## Checklist

- [x] `pnpm lint` passes (tsc --noEmit, strict)
- [x] Tests pass (affected files + full suite green)
- [x] DCO signed (`Signed-off-by`)
- [x] No secrets / tokens / private absolute paths in the diff

### DIFF
--- a/src/agent/hooks-integration.test.ts
+++ b/src/agent/hooks-integration.test.ts
@@ -970,6 +970,151 @@ describe('SubagentHandle.teardown()', () => {
     expect(queued).toContain('post-teardown note');
   });
 
+  // ----------------------------------------------------------------------
+  // In-turn injectContext delivery (deferInjectContextToCaller).
+  //
+  // Exactly-once crux: when teardown() is asked to defer delivery to the
+  // caller, the produced injectContext is recorded on
+  // getLastStopInjectContext() for the caller to append to the tool_result,
+  // and the queue push is SUPPRESSED. Without the flag, delivery still rides
+  // the queue (the compose/DAG + cancel/fail-fast paths keep this behavior).
+  // ----------------------------------------------------------------------
+  it('teardown({ deferInjectContextToCaller: true }) records the note and does NOT queue (deliver-once)', async () => {
+    const registry = createHookRegistry();
+    registry.register('SubagentStop', () => ({
+      injectContext: 'defer: in-turn nudge',
+    }));
+
+    const parentMgr = new SubagentManager({ hookRegistry: registry });
+    const parentHandle = await parentMgr.forkSubagent({
+      parent: { sessionId: 'root-defer' },
+      config: { model: 'sonnet' },
+      idPrefix: 'parent',
+    });
+    const parentSession = shared.sessions[shared.sessions.length - 1]!;
+
+    const childMgr = new SubagentManager({ hookRegistry: registry });
+    const childHandle = await childMgr.forkSubagent({
+      parent: {
+        sessionId: parentHandle.session.sessionId,
+        getInputStreamRef: parentHandle.session.getInputStreamRef.bind(parentHandle.session),
+      },
+      config: { model: 'sonnet' },
+      idPrefix: 'child',
+    });
+
+    await childHandle.run('audit');
+    await childHandle.teardown({ deferInjectContextToCaller: true });
+
+    // Recorded for the caller to deliver in-turn…
+    expect(childHandle.getLastStopInjectContext()).toBe('defer: in-turn nudge');
+    // …and NEITHER queue channel fired (suppressed — deliver-once).
+    expect(parentSession.getMockQueuedFrameworkContext()).toEqual([]);
+    expect(parentSession.getMockInputStreamMessages()).toEqual([]);
+  });
+
+  it('teardown() WITHOUT the defer flag still queues and leaves getLastStopInjectContext undefined', async () => {
+    const registry = createHookRegistry();
+    registry.register('SubagentStop', () => ({
+      injectContext: 'queue: default channel',
+    }));
+
+    const parentMgr = new SubagentManager({ hookRegistry: registry });
+    const parentHandle = await parentMgr.forkSubagent({
+      parent: { sessionId: 'root-default-queue' },
+      config: { model: 'sonnet' },
+      idPrefix: 'parent',
+    });
+    const parentSession = shared.sessions[shared.sessions.length - 1]!;
+
+    const childMgr = new SubagentManager({ hookRegistry: registry });
+    const childHandle = await childMgr.forkSubagent({
+      parent: {
+        sessionId: parentHandle.session.sessionId,
+        getInputStreamRef: parentHandle.session.getInputStreamRef.bind(parentHandle.session),
+      },
+      config: { model: 'sonnet' },
+      idPrefix: 'child',
+    });
+
+    await childHandle.run('audit');
+    await childHandle.teardown();
+
+    // Default path: queued, and nothing recorded for the caller.
+    expect(parentSession.getMockQueuedFrameworkContext()).toContain('queue: default channel');
+    expect(childHandle.getLastStopInjectContext()).toBeUndefined();
+  });
+
+  it('deferInjectContextToCaller: parent abort suppresses BOTH channels (nothing recorded, nothing queued)', async () => {
+    // Abort precedence is checked before the defer branch — an aborting parent
+    // will unwind before it could consume the note, so neither the caller nor
+    // the queue receives it.
+    const registry = createHookRegistry();
+    registry.register('SubagentStop', () => ({
+      injectContext: 'defer: should be suppressed by abort',
+    }));
+
+    const parentAbortController = new AbortController();
+    const parentMgr = new SubagentManager({
+      hookRegistry: registry,
+      parentAbortSignal: parentAbortController.signal,
+    });
+    const parentHandle = await parentMgr.forkSubagent({
+      parent: { sessionId: 'root-defer-abort' },
+      config: { model: 'sonnet' },
+      idPrefix: 'parent',
+    });
+    const parentSession = shared.sessions[shared.sessions.length - 1]!;
+
+    const childMgr = new SubagentManager({ hookRegistry: registry });
+    const childHandle = await childMgr.forkSubagent({
+      parent: {
+        sessionId: parentHandle.session.sessionId,
+        getInputStreamRef: parentHandle.session.getInputStreamRef.bind(parentHandle.session),
+        abortSignal: parentHandle.session.abortSignal,
+      },
+      config: { model: 'sonnet' },
+      idPrefix: 'child',
+    });
+
+    await childHandle.run('audit');
+    parentAbortController.abort('parent-abort');
+    await childHandle.teardown({ deferInjectContextToCaller: true });
+
+    expect(childHandle.getLastStopInjectContext()).toBeUndefined();
+    expect(parentSession.getMockQueuedFrameworkContext()).toEqual([]);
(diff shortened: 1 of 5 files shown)


griffinwork40:
`ToolResult.content` is typed as `string` (anthropic-direct/types.ts:80), and `Message.content` is also `string` — so the template-literal append is fully type-safe with no array-corruption risk. That closes my last open concern.

I now have complete, independently-verified ground truth. Writing up the review.

---

# Code Review — PR 387: deliver SubagentStop injectContext in-turn via the subagent tool_result

**Reviewed:** committed HEAD `93dfd77` in worktree `inturn-nudge` (read-only; no files touched). Builds on #359, which moved injectContext delivery to the deferred queue/next-user-message channel. This PR adds an **in-turn** path so a foreground `agent`/`skill` fork can carry its `SubagentStop.injectContext` (e.g. the shadow-verify nudge) on the *same* tool_result the parent sees, instead of the next turn.

## Verdict: ✅ **Approve** (no blockers; 2 non-blocking follow-ups)

### Deterministic gates (run locally in the worktree)
- **`pnpm lint` (tsc --noEmit, maximally strict): PASS** (exit 0) — the new required interface method and optional param typecheck clean.
- **Full test suite: PASS** — 10543 passed / 14 skipped / **0 failed** (571 files, 34s). New tests pass; no regressions.

---

## Findings by severity

**No blockers. No high-severity issues.** The core mechanism is correct and the change is additive + backward-compatible.

### 🟡 Medium

- **skill-executor's in-turn append is entirely untested** — `src/agent/tools/skill-executor.ts:822` and `:1138`. The append logic is duplicated across the two forked-skill paths (registry + plugin), and `skill-executor.test.ts` has **zero** coverage for it (confirmed: no match for `injectContext`/`getLastStopInjectContext`/`deferInjectContextToCaller`). Only the `subagent-executor` path is tested. A silent regression in either skill path (wrong flag, missing append, stray separator) would ship undetected. *Recommend: add success + no-context cases mirroring the subagent-executor suite for at least one skill path.*

### 🟢 Low

- **injectContext is silently dropped on the throw/catch path** — `subagent-executor.ts:983-1000,1015-1027`; same in skill-executor catch paths. On an unexpected throw, `toolResult` stays `undefined`, so the recorded note is never appended — and the queue push was already suppressed by `deferInjectContextToCaller`. Pre-PR, `teardown()` (no flag) would have **queued** it. So this is a genuine (narrow) behavioral change: on an unexpected subagent throw, a supplemental nudge that previously reached the parent is now lost. The author acknowledges this in-comment ("note is dropped for that stop by design — the throw is the parent's signal"), and the throw path is rare defense-in-depth with a supplemental payload, so severity is low — but confirm it's an intentional tradeoff rather than an accident. A queue fallback in the catch would restore old behavior if desired.
- **Promoted (Ctrl+B) and throw paths have no explicit test** — the "no append when `toolResult` unset" behavior for the promotion path (`subagent-executor.ts:864-905`) and the catch path is only implied. A one-line assertion that promotion returns the running-pointer *without* a nudge suffix would lock the invariant.
- **Mock fidelity seam** — `subagent-executor.test.ts` mocks `getLastStopInjectContext` directly, bypassing the real handle defer/suppress logic. That logic *is* covered independently by `hooks-integration.test.ts` (records-and-suppresses; default-queues; abort-suppresses-both; no-context-undefined), so exactly-once is proven at each layer — but no single end-to-end test wires a real SubagentStop hook through a real executor across the seam. Acceptable given the two-layer coverage; noting for completeness.

### ⚪ Nit

- **Append snippet duplicated 3×** — identical 4-line block at `subagent-executor.ts:1024`, `skill-executor.ts:822`, `skill-executor.ts:1138`. Extracting `appendInjectContext(toolResult, handle)` would kill the drift risk the code's own "Mirrors the pattern in…" comments are warning about.

---

## Correctness verification (all CONFIRMED)

1. **Try/finally mutation semantics (load-bearing):** ✅ `return toolResult` captures the object *reference*; the `finally` mutates `.content` on that same object (never reassigns `toolResult`, no `return` in finally), so the caller sees the appended content. Well-defined JS; also empirically re-derived by a subagent.
2. **Exactly-once:** ✅ The 4-way ordered decision in `handle.ts:529-550` (no-context → abort-suppress → defer-record → queue) routes to exactly one channel; `stopDispatched` (`:474-481`) guards against double dispatch.
3. **Promoted path:** ✅ `promoted=true` → `finally` skips teardown (`:1008`) → `toolResult` unset → no append; returns the running pointer.
4. **Abort precedence:** ✅ The parent-aborting → suppress-both branch (`handle.ts:530`) is checked *before* the defer branch, matching the abort-graph "abort-signal check is unconditional" invariant; tested in `hooks-integration.test.ts:1084`.
5. **Abort/cancel interaction:** ✅ `call.signal` abort → `cancel()` fires stop without defer (queue or abort-suppress); the `finally` `teardown({defer})` is then a no-op via `stopDispatched`. No double-delivery.
6. **No read/write race:** ✅ `teardown()` is `await`ed before `getLastStopInjectContext()` is read.

**Security:** ✅ `injectContext` is framework-generated (per the handle.ts invariant), not untrusted input — no injection surface beyond what SubagentStop hooks already had. **Perf:** ✅ one extra string concat at teardown; negligible. **Type safety:** ✅ `ToolResult.content` and `Message.content` are both `string` — append cannot corrupt array content.

## API-compat (all CONFIRMED)
- Only `SubagentHandleImpl` implements `SubagentHandle`; the new **required** `getLastStopInjectContext()` is defined there (lint confirms no other implementer breaks).
- All ~13 `teardown()` callers correct: only the 3 intended in-turn sites pass `{deferInjectContextToCaller:true}`; DAG/wave/teardownAll/mint/diagnose/background-cap paths keep the default queue behavior — a deliberate scoping choice ("compose/DAG + cancel/fail-fast keep this behavior").
- Optional-chaining correct at every append site (`handle.get…?.()` where handle is guaranteed; `handle?.get…?.()` in skill-executor where the fork may have thrown). No new SDK runtime imports.

---

## Terminal state — Done

- **What was done:** Full read-only review of PR 387 (`93dfd77`) — read all 3 source files + 2 test files + the diff, traced the 4-way delivery decision and every `teardown()` caller, and ran both authoritative gates locally in the worktree.
- **Evidence:** `pnpm lint` exit 0; full suite `10543 passed / 0 failed` (both run in `/Users/griffinlong/Projects/open_source/agent-afk/.afk-worktrees/inturn-nudge`); diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/fcd0526f-86cd-4ee0-a483-3d408f9a8948/pr-387.diff`. Coverage gap confirmed by grep (no `injectContext` matches in `skill-executor.test.ts`); `ToolResult.content: string` at `anthropic-direct/types.ts:80`.
- **What changed in the world:** Nothing — read-only review, no files edited, no commit, no PR mutation.
- **Recommendation:** **Approve/merge.** Optionally address before or after merge: (1) add skill-executor in-turn append tests [medium], (2) confirm the throw-path note-drop is intentional [low], (3) extract the 3× duplicated append into a helper [nit].
- **Deferred/not checked:** Did not exercise a live end-to-end shadow-verify → parent delivery at runtime (relied on unit + suite coverage); did not review the compose orchestration path since the PR intentionally leaves it on the queue channel. Two `compose` review-wave dispatches aborted earlier (fail-fast + tool-call budget) — I completed the verification directly instead, so all findings above are first-hand.

---

PR #390, opened 2026-07-03
feat(worktree): lifecycle tool + sweep fix for agent-managed worktrees

griffinwork40:
## Summary

- **Sweep fix**: `stale-clean` worktrees (clean, >14 days old, commits ahead of base) are now preserved with a warning instead of force-removed. Since the `empty` verdict already catches clean+zero-commits-ahead trees earlier in classification, `stale-clean` in practice only ever fired on trees holding committed-but-unmerged work — so the old behavior exclusively destroyed unpushed commits sitting in an otherwise-clean worktree. This is the most likely cause of recurring "my worktree got swept" incidents.
- **New `worktree` tool** (`create` / `keep` / `release` / `list` / `remove`) gives the agent a protocol-aware lifecycle instead of raw `git worktree add` via bash. `create` always places trees under `.afk-worktrees/` and writes proper meta (owner, pid, createdAt, baseSha, baseBranch), closing the gap where agent-created worktrees had no meta and looked like anonymous ghosts to the sweep engine. `keep` wraps `git worktree lock`, which the sweep engine already unconditionally skips — zero sweep-engine changes needed for the lock guarantee. `remove` refuses dirty/locked/commits-ahead/main/outside-root trees and never deletes the underlying branch.
- **Subagent occupancy touch**: `forkSubagent` (the single convergence point for `agent`/`skill`/`compose` dispatch with a `cwd` override) now best-effort refreshes a worktree's meta (pid, createdAt) when the resolved cwd sits inside `.afk-worktrees/`. This resets the sweep's age clock for subagent-occupied trees and adopts previously meta-less bash-created trees, without ever blocking dispatch on failure.
- `agent` tool docs updated to point at the new `worktree` tool instead of advising `bash: git worktree add`.

## Test results
- `pnpm test` (vitest run): 574/574 test files passed, 10571 passed / 14 skipped
- `pnpm audit:sdk:check`: OK, no new SDK imports

## Verification (adversarial `--verify` pass, independent re-derivation from diff + source)

|---|---|---|
| `stale-clean` previously force-removed clean trees with unpushed commits; now warns and preserves, no classification change | CONFIRM | `worktree-sweep.ts:280-297`, `:621-632` |
| `create` always writes meta under `.afk-worktrees/`, closing the ghost-worktree gap | REFINE | `worktree.ts:240-256` — meta fields are `owner/pid/createdAt/baseSha/baseBranch`; there is no `sessionId` field (corrected from an earlier draft claim) |
| `keep` (`git worktree lock`) requires zero sweep-engine changes — locked trees are already unconditionally skipped | CONFIRM | `worktree-sweep.ts:262` (pre-existing, untouched by this diff) |
| `forkSubagent` is the single convergence point for agent/skill/compose dispatch; occupancy touch is fire-and-forget and never blocks | CONFIRM | `subagent.ts:553-555`; call sites `subagent-executor.ts:670`, `skill-executor.ts:758,1065`, `dag-subagent.ts:89`; `worktree-occupancy.ts:64-81` swallows all errors |
| `remove` refuses dirty/locked/commits-ahead/main/outside-root; branch never deleted | CONFIRM | `worktree.ts:316-346`, `:123-133`, `:347` — lock check ordered before the `force` branch, so `force` cannot bypass a lock |

**Known residual gap (accepted, not fixed in this PR):** a worktree adopted purely via the occupancy touch (no prior `create` meta) gets a minimal record with no `baseSha`, so `remove`'s commits-ahead guard silently reads as 0 for that specific tree — only the dirty check protects it in that case. Pre-existing `create`d trees are unaffected. Tracked as a follow-up if it proves to matter in practice.

## Out of scope
- Not fixed here: unbounded growth of warned `stale-clean` trees now that they're never auto-removed — cleanup relies on the model or user invoking `worktree remove` explicitly. Intentional tradeoff (favors "never destroy unmerged work" over "never accumulate"), flagged for future follow-up if the `.afk-worktrees/` count becomes a problem.

### DIFF
--- a/src/agent/daemon/scheduler.ts
+++ b/src/agent/daemon/scheduler.ts
@@ -436,9 +436,10 @@ export class CronScheduler {
         telemetryPath: this.telemetryPath(),
       });
 
+      // 'stale-clean' is intentionally absent: the sweep engine preserves +
+      // warns on stale-clean (commits ahead of base) rather than removing.
       const prunableVerdicts = new Set([
         'empty',
-        'stale-clean',
         'orphaned-dir',
         'orphaned-registration',
         'dead-owner',
--- a/src/agent/plan-mode-gate.ts
+++ b/src/agent/plan-mode-gate.ts
@@ -59,6 +59,25 @@ export function createPlanModeGate(
       };
     }
 
+    // `worktree` is action-gated: `list` is a read-only dry-run sweep and
+    // passes; create/keep/release/remove mutate the git worktree registry
+    // and/or filesystem, mirroring the bash classifier's GIT_WORKTREE_MUTATING
+    // rule so the two surfaces cannot disagree about what "mutating" means.
+    if (toolName === 'worktree') {
+      const action =
+        typeof context.input === 'object' && context.input !== null
+          ? String((context.input as Record<string, unknown>)['action'] ?? '')
+          : '';
+      if (action !== 'list') {
+        return {
+          decision: 'block',
+          reason:
+            `plan mode: worktree "${action}" is refused (mutates the worktree ` +
+            `registry). Only action "list" is allowed in plan mode. Use /plan off to act.`,
+        };
+      }
+    }
+
     // `bash` is mutation-gated, not blanket-refused: read-only recon runs,
     // state-mutating commands are refused. Reuses the same best-effort
     // classifier as the read-only skill phases (single source of mutation
--- a/src/agent/risk-classifier.test.ts
+++ b/src/agent/risk-classifier.test.ts
@@ -223,6 +223,47 @@ describe('classifyRisk — schedule tools', () => {
   });
 });
 
+// ---- worktree lifecycle --------------------------------------------------
+// Regression guard for PR #390 review finding F1: the `worktree` tool had no
+// classifier branch, so every action (including `remove --force`, which can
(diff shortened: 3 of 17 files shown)


griffinwork40:
Addressed the code-review findings:

**Fixed in this PR** (`7db727e`):
- **F1** (medium, security) — the `worktree` tool had no `classifyRisk` branch, so every action (including `remove --force`) resolved to `'safe'` and ran unattended in AFK autonomous mode with no approval. Added a `worktree` case mirroring the schedule-mutation precedent: `remove`+`force:true` → `'high'` (approval-gated), `list` → `'safe'` (read-only dry-run), all other actions → `'medium'`. Covered by 8 new regression tests. Lint clean; full suite green.

**Deferred to tracking issues** (non-blocking):
- #395 — F2 (low): sanitize `base` before `git worktree add` (arg-injection footgun)
- #396 — F3 (low): `/worktree` REPL surface renders `stale-clean` as healthy (parity with CLI)
- #397 — F4 (nit): `stale-clean` has no commits-ahead check; comment overstates guarantee
- #398 — F6 (low): add negative tests (traversal, flag-like `base`, locked+force)

F5 (meta-less tree `commitsAhead` no-op) was already disclosed in the PR body and accepted — no issue filed.

PR #394, opened 2026-07-03
fix(subagent): bound forked-child tool-use loop to prevent parent hang

griffinwork40:
## Problem

A parent that dispatches a subagent (`agent` / `skill` / `compose` / dag) suspends at `await runToResult` for the child's entire lifetime. On the **anthropic-direct** provider that lifetime was **unbounded**:

- `DEFAULT_MAX_TOOL_USE_ITERATIONS = 0` means "no cap" on the tool-use loop, and nothing in non-test code ever set `RunTurnInput.maxToolUseIterations` for a child.
- A subagent runs **exactly one** conversation turn (`runToResult` → `run()` → a single `sendMessageStream`), so the harness-level `max_turns` never bounds the tool-use loop *within* that turn — it only limits sends, and a child makes just one.

Net effect: a child stuck in a tool-calling doom-loop could spin forever while the parent hangs — no cap, no error, and (on a daemon or nested fork) no Ctrl+B rescue. `openai-compatible` children were already safe (built-in 50-round cap); only anthropic-direct was exposed.

## Fix

Thread a tool-use-iteration ceiling end-to-end and default every fork to **50** (parity with openai-compatible) at the single `forkSubagent` chokepoint that all four subagent paths funnel through:

- `AgentConfig.maxToolUseIterations` — new optional field
- forwarded in `AnthropicDirectProvider.query()` → `AnthropicDirectQueryOptions` → constructor field → `runInput` (same optional-spread idiom as `effort` / `baseUrl`)
- `SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS = 50` applied in `forkSubagent`'s childConfig (mirrors the existing `isNonInteractive ?? true` pattern, so caller overrides win)

Hitting the cap surfaces as a clean `tool_use_loop_capped` **done**, so the child returns its partial work rather than erroring. Callers may override per-fork via `config.maxToolUseIterations`, including `0` to opt a trusted deep child back into unbounded.

## Blast radius

Top-level sessions (REPL / chat / telegram / daemon / one-shot query) are **unaffected** — their `AgentConfig` objects never pass through `forkSubagent`; every other `new AgentSession` site was verified to be a top-level entry point. openai-compatible unchanged (already caps at 50).

## Testing

- `pnpm lint` — clean (`tsc --noEmit`, strict)
- `pnpm test` — full suite **10570 passed / 0 failed / 14 skipped** (571 files)
- `pnpm audit:sdk:check` — OK (no new SDK symbols)
- New tests:
  - `subagent.test.ts` — default cap applied, explicit override preserved, `0` opt-out preserved
  - `anthropic-direct.test.ts` — end-to-end `config.maxToolUseIterations` → provider → loop cap fires at exactly N rounds with `tool_use_loop_capped`

## Follow-up (out of scope)

openai-compatible ignores an explicit `config.maxToolUseIterations` (its 50-round cap is hardcoded); honoring an explicit override there would make the two providers fully symmetric. Left out to keep this change minimal.

### DIFF
--- a/src/agent/providers/anthropic-direct.test.ts
+++ b/src/agent/providers/anthropic-direct.test.ts
@@ -683,6 +683,47 @@ describe('AnthropicDirectProvider', () => {
     }
   });
 
+  it('caps the tool-use loop at config.maxToolUseIterations (config → provider → loop)', async () => {
+    // End-to-end plumbing guard: AgentConfig.maxToolUseIterations must thread
+    // through AnthropicDirectProvider.query() → AnthropicDirectQueryOptions →
+    // the constructor → runInput → the loop's `maxIterations`. The mock model
+    // emits tool_use on every round, so the ONLY thing that can terminate the
+    // turn is the iteration cap — this is the guard a forked subagent relies on
+    // to avoid hanging its parent (see SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS).
+    let callIdx = 0;
+    messagesCreateMock.mockImplementation(() => {
+      callIdx += 1;
+      return fromArray(
+        makeToolUseStream(`toolu_${callIdx}`, 'get_weather', '{"city":"SF"}'),
+      );
+    });
+
+    const dispatcher: ToolDispatcher = {
+      async execute(): Promise<ToolResult> {
+        return { content: 'sunny' };
+      },
+    };
+
+    const provider = new AnthropicDirectProvider({ tools: dispatcher });
+    const query = provider.query({
+      prompt: singleInput('weather?'),
+      config: {
+        model: 'claude-sonnet-5',
+        apiKey: 'sk-ant-api03-test',
+        maxToolUseIterations: 2,
+      },
+    });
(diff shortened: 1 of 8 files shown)


griffinwork40:
Resolved the P2 in `a2a4e3a`.

When the tool-use cap fires, the loop emits a `tool_use_loop_capped` done with no assistant message; for a pure tool-only runaway (no streamed text) `streamToFinalMessage()` fell through to `throw "produced no terminal message"`, reporting the exact runaway child the cap is meant to bound as a *failed* subagent.

Fix: `handle.ts` now captures the done event's `stopReason` and, when it's `tool_use_loop_capped` with no message/streamed content, returns a synthetic terminal "capped" message instead of throwing — so `runToResult` reports a capped **partial** result (status `succeeded`), matching the contract the cap default already documents. Narrowest scope (subagent-only; mirrors the existing partial-content fallback), so top-level sessions that set a cap are unaffected. Regression test added in `handle-streaming.test.ts`.

PR #403, opened 2026-07-04
fix(tui): coalesce post-ESC type-ahead so soft-stop doesn't strand input one turn behind

griffinwork40:
## Problem

After pressing **ESC to soft-stop** a running turn, the next message "doesn't send" — the user has to send a follow-up to get the turn to start, and then input lags so they "keep having to send single characters to catch up."

## Root cause

Input mode is coupled to turn-teardown timing:

1. `handleEscape` (`terminal-compositor.input-dispatch.ts`) sets `softStopped = true` but leaves the compositor in **`'streaming'`** mode — it stays streaming until the async teardown finishes.
2. In streaming mode, **Enter queues** onto the `pendingSubmissions` FIFO (`handleEnter`) instead of submitting.
3. The FIFO **drains one payload per `readLine→idle`** — the turn-end `dispose→idle` can't drain (`onSubmit` is null then), so a backlog drains one-behind, turn after turn.
4. The window is long **only for slow teardowns**: #400 added `cancelActiveForeground()`, which is async/best-effort — the parent for-await can't break until the child subagent settles (seconds for a deep/wide wave), and the compositor lingers in `streaming` the whole time. (Token-streaming ESC is unaffected — its interrupt is effectively synchronous.)

This is a regression **surfaced** by #400: before it, ESC-during-subagent did nothing (the turn hung), so the drain-backlog path was never reached.

## Fix

Gate the streaming-Enter queue on `softStopped`: during the ESC window keep only the **latest** post-ESC message (last-wins), so the user's most recent post-stop intent runs as exactly one next turn — no backlog. `softStopped` is set only by ESC (and cleared at the post-soft-stop `→ idle` transition), so **normal mid-turn type-ahead is untouched** (it still accumulates every message for sequential-turn delivery).

Pre-ESC payloads already on the FIFO are contract-protected (`handleEscape`: "Already-queued messages: left untouched"). `handleEscape` snapshots `pendingSubmissions.length` into `softStopQueueBase` at ESC time; `handleEnter` then **truncates back to that base before pushing** — preserving pre-ESC entries (they drain as their own sequential turns) while post-ESC type-ahead still coalesces to last-wins.

> *Note:* array-wide reassignment (`pendingSubmissions = [payload]`) would silently drop those pre-ESC payloads, violating the `handleEscape` contract. Truncate-then-push does not.

```ts
// handleEscape: snapshot the pre-ESC queue length at ESC time.
self.softStopQueueBase = self.pendingSubmissions.length;

// handleEnter: during the soft-stop window, truncate-then-push (last-wins
// for post-ESC entries, pre-ESC entries untouched).
if (self.softStopped) {
  self.pendingSubmissions.length = self.softStopQueueBase;
  self.pendingSubmissions.push(payload);   // last-wins during the ESC window
} else {
  self.pendingSubmissions.push(payload);   // normal type-ahead: accumulate
}
```

`softStopQueueBase` is reset to `0` at every site that clears `softStopped` (idle→streaming, post-soft-stop →idle, `resetState`, class initializer).

## Tests

Added to `describe('soft-stop drain')` in `terminal-compositor.test.ts`:

- **Multi-message-during-window coalesce** — ESC, then 3 messages Entered during the teardown window → FIFO holds only the latest → drains as exactly one next turn.
- **No-ESC blast-radius guard** — ordinary type-ahead (no ESC) still accumulates all 3 messages.
- **Pre-ESC queue survives a post-ESC Enter** — a message committed via Enter *before* ESC is not dropped when a post-ESC Enter truncates; both drain as their own turns (oldest-first).
- **Pre-ESC queue survives MULTIPLE post-ESC Enters** — same contract with 3 post-ESC Enters: pre-ESC payload survives, three post-ESC ones coalesce to last-wins (queue=2, not 1 and not 4).

## Verification

- `pnpm lint` (`tsc --noEmit`) — clean (exit 0)
- Full suite — **10,643 passed / 0 failed / 14 skipped**
- Focused `soft-stop drain` block — 10/10 pass

## Not included (follow-ups)

- **"interrupting…" affordance for ESC** — an affordance exists but its copy is Ctrl+C-specific and only fires from the SIGINT handler; reusing it for ESC needs a parameterized message + tests. The queue fix already removes the substantive harm (the backlog); the affordance only improves the *perception* of the inherent teardown latency.
- **Interrupt telemetry event** — the witness trace records no soft-stop/interrupt marker today. `AgentSession.interrupt()` can reach the trace writer, but the only existing event is `abort` (AbortGraph hard-abort semantics); a graceful soft-stop wants a dedicated event kind (schema + reader + tests across `src/agent/trace/`).
- **Repaint-starvation** — "catch up with single characters" *may* also point at a render-loop issue distinct from the queue backlog; unconfirmed. Repro: ESC a long subagent, type one message, watch whether the queued turn renders without further keystrokes.

### DIFF
--- a/src/cli/terminal-compositor.input-dispatch.ts
+++ b/src/cli/terminal-compositor.input-dispatch.ts
@@ -626,11 +626,29 @@ function handleEnter(self: KeyDispatchHost, key: KeyInfo, sequence: string): boo
   const displayText = self.input.buffer;
   const expandedText = Paste.expandPastePlaceholders(self, displayText);
   const attachments = [...self.attachments];
-  self.pendingSubmissions.push(
+  const payload: SubmissionPayload =
     expandedText === displayText
       ? { text: expandedText, attachments }
-      : { text: expandedText, displayText, attachments },
-  );
+      : { text: expandedText, displayText, attachments };
+  // Invariant: during the ESC soft-stop window — `softStopped` is set by
+  // handleEscape and cleared only at the post-soft-stop `→ idle` transition
+  // (setInputMode, terminal-compositor.input-mode.ts) — Enter must NOT
+  // accumulate a type-ahead backlog. That window is the async turn-teardown
+  // gap: for a subagent turn, cancelActiveForeground() (subagent-executor.ts)
+  // resolves the parent `await` only after the child settles — seconds for a
+  // deep/wide wave — and the compositor lingers in 'streaming' the whole time.
+  // Each Enter would otherwise push onto the FIFO, which drains ONE payload per
+  // turn (the `→ idle` flush), stranding the user one turn behind: the "it
+  // doesn't send, then I keep sending characters to catch up" report. So during
+  // a soft-stop we keep only the LATEST message (last-wins) — the user's most
+  // recent post-stop intent runs as exactly one next turn, no backlog. Normal
+  // mid-turn type-ahead (softStopped === false) still accumulates: sequential-
+  // turn delivery is the intended contract there (the "NO ESC" regression tests).
+  if (self.softStopped) {
+    self.pendingSubmissions = [payload];
+  } else {
+    self.pendingSubmissions.push(payload);
+  }
   self.queued = true; // maintained mirror: pendingSubmissions is now non-empty
   // Clear the compose window for the next message. Mirrors the idle-mode
   // submit reset above so dropdown chrome / paste side-table / attachments
--- a/src/cli/terminal-compositor.input-dispatch.ts
+++ b/src/cli/terminal-compositor.input-dispatch.ts
@@ -108,6 +108,15 @@ export interface KeyDispatchHost {
 
   /** Once-only soft-stop guard for ESC in streaming mode. */
   softStopped: boolean;
+  /**
+   * Snapshot of `pendingSubmissions.length` taken at ESC soft-stop time
+   * (handleEscape). Entries at indices `0..softStopQueueBase-1` were queued
+   * BEFORE esc and are contract-protected (handleEscape comment lines 318-327):
+   * "Already-queued messages: left untouched." Post-ESC Enters coalesce to
+   * last-wins by truncating back to this base before pushing, so the pre-ESC
+   * queue is never silently dropped.
+   */
+  softStopQueueBase: number;
   /** Hard-abort flag set by Ctrl+C in streaming mode. */
   canceled: boolean;
   /** Once-only guard for Ctrl+B background in streaming mode. */
@@ -329,6 +338,10 @@ function handleEscape(self: KeyDispatchHost, key: KeyInfo): boolean {
   // committing it would fling it as a turn the user never submitted. Ctrl+C
   // (handleInterrupt below) follows the same no-auto-commit rule.
   self.softStopped = true;
+  // Snapshot the queue length so post-ESC coalesce (handleEnter) can truncate
+  // back to here — preserving pre-ESC payloads per the contract above —
+  // while still giving last-wins for messages Entered after this ESC.
+  self.softStopQueueBase = self.pendingSubmissions.length;
   if (self.onSoftStop) self.onSoftStop();
   return true;
 }
@@ -640,12 +653,21 @@ function handleEnter(self: KeyDispatchHost, key: KeyInfo, sequence: string): boo
   // Each Enter would otherwise push onto the FIFO, which drains ONE payload per
   // turn (the `→ idle` flush), stranding the user one turn behind: the "it
   // doesn't send, then I keep sending characters to catch up" report. So during
-  // a soft-stop we keep only the LATEST message (last-wins) — the user's most
-  // recent post-stop intent runs as exactly one next turn, no backlog. Normal
-  // mid-turn type-ahead (softStopped === false) still accumulates: sequential-
-  // turn delivery is the intended contract there (the "NO ESC" regression tests).
+  // a soft-stop we keep only the LATEST post-ESC message (last-wins) — the user's
+  // most recent post-stop intent runs as exactly one next turn, no backlog.
+  //
+  // Contract preservation: `softStopQueueBase` was snapshotted by handleEscape
+  // at ESC time, capturing the count of pre-ESC payloads already on the FIFO.
+  // Truncating back to that base before pushing preserves those pre-ESC entries
+  // (they drain as their own sequential turns via the `→ idle` flush), so the
+  // handleEscape contract — "Already-queued messages: left untouched" — holds.
+  // Array-wide reassignment (`= [payload]`) would silently drop them; truncate-
+  // then-push does not. Normal mid-turn type-ahead (softStopped === false) still
+  // accumulates: sequential-turn delivery is the intended contract there (the
+  // "NO ESC" regression tests).
   if (self.softStopped) {
-    self.pendingSubmissions = [payload];
+    self.pendingSubmissions.length = self.softStopQueueBase;
+    self.pendingSubmissions.push(payload);
   } else {
     self.pendingSubmissions.push(payload);
   }
--- a/src/cli/terminal-compositor.input-mode.ts
+++ b/src/cli/terminal-compositor.input-mode.ts
@@ -42,6 +42,9 @@ export interface InputModeHost {
   /** Once-only ESC soft-stop guard. */
   softStopped: boolean;
 
+  /** Queue-length snapshot at ESC time; coalesce boundary for post-ESC Enters. */
+  softStopQueueBase: number;
+
   /** Hard-abort flag (Ctrl+C in streaming mode). */
   canceled: boolean;
 
@@ -173,6 +176,7 @@ export function setInputMode(self: InputModeHost, mode: CompositorInputMode): vo
     self.canceled = false;
     self.backgrounded = false;
     self.softStopped = false;
+    self.softStopQueueBase = 0;
     // Reset autocomplete at the idle→streaming transition so any
     // open dropdown rows are not rendered into the first streaming
     // frame. Mirrors the reset in the idle Enter handler above.
@@ -210,6 +214,7 @@ export function setInputMode(self: InputModeHost, mode: CompositorInputMode): vo
   // keeps an EMPTY-buffer ESC from leaving it armed into the idle period.
   if (mode === 'idle' && self.softStopped) {
     self.softStopped = false;
+    self.softStopQueueBase = 0;
     // No early return: fall through so a queued buffer flushes via the branch
     // below when onSubmit is installed (readLine→idle), or — at the
     // dispose→idle transition where onSubmit is still null — stays queued for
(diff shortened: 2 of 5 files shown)


griffinwork40 (commented):

# Code Review — PR #403 · `fix(tui): coalesce post-ESC type-ahead so soft-stop doesn't strand input one turn behind`

**Reviewed ref:** `b262dfb1baf217a08bf7be8b8fff3a30f392da09` · 2 files · +76/−3 · regime: light (1 full-dimension Wave 1 agent + Wave 1.5 citation verification + shadow-verify)

## ⛔ Merge decision: **DO NOT MERGE**

One **high** correctness finding — an invariant violation of the codebase's own contract — independently confirmed **three** times (Wave 1, Wave 1.5 fresh code trace, and `/shadow-verify` with a live reproduction: `getPendingCount()` 2→1 across the reachable ESC-then-Enter sequence). This PR fixes a real regression (#400's async teardown stranding input) but introduces a more severe variant of the same failure class it's curing: silent, permanent, unrecoverable loss of queued user input — versus the original bug's temporary one-turn lag. Reachable via ordinary keystrokes during the multi-second async teardown window the PR's own root-cause narrative describes.

## Findings

**1 · HIGH · confidence:high · correctness · `src/cli/terminal-compositor.input-dispatch.ts:647-651` · ref:b262dfb1b · file-state**
The new `if (self.softStopped) { self.pendingSubmissions = [payload]; }` branch **replaces** the whole queue, not just what accumulated during the ESC window — so if ≥1 message was already queued via ordinary pre-ESC type-ahead (`softStopped === false`, two normal Enters both hit the untouched `else` push branch), and the user then presses ESC and one more Enter before the next `→ idle` flush, those pre-ESC messages are silently and permanently destroyed. This directly contradicts `handleEscape`'s own documented contract at `input-dispatch.ts:318-322` — "Already-queued messages... left untouched." `handleEscape` (lines 279-334) never reads or snapshots `pendingSubmissions`, confirming the contract-holder side is honored; it's `handleEnter`'s new branch that breaks it.
```
  if (self.softStopped) {
    self.pendingSubmissions = [payload];
  } else {
    self.pendingSubmissions.push(payload);
  }
```
**Fix:** snapshot the queue boundary (index/length) when `softStopped` is set in `handleEscape`, and in this branch truncate back to that boundary before appending the new payload — keep everything queued *before* the ESC, coalesce only what's typed *after* it — instead of unconditionally replacing the whole array.

**2 · MEDIUM · confidence:high · spec-compliance · `src/cli/terminal-compositor.input-dispatch.ts:633-651` · ref:b262dfb1b · file-state**
The PR title/body scope the fix to coalescing messages typed "post-ESC" / "during the teardown window." Gating on the coarse `softStopped` boolean rather than on *when each entry was committed* makes the actual blast radius wider — it also erases messages queued via ordinary type-ahead *before* the ESC window opened, which the stated intent never calls for. (Same root cause as the correctness finding above, viewed as scope creep against the stated intent rather than as a bug in isolation.) Same fix as #1.

**3 · LOW · confidence:high · test-coverage · `src/cli/terminal-compositor.test.ts:1804-1857` · ref:b262dfb1b · file-state**
Both new tests start the ESC window from an empty queue; none of the 16 ESC-emitting tests in the file seed ≥1 pre-ESC queued entry and then Enter again post-ESC — the exact sequence that trips the correctness finding above. The one existing test that queues 1 message before ESC (`test.ts:1423-1440`, "ESC after Enter leaves the queued message in the FIFO") never presses Enter again afterward, so it stops one step short of exposing the bug.
**Fix:** add a case — type+Enter a message pre-ESC (`pendingCount` → 1), ESC, type+Enter one more during the window — and assert whether the pre-ESC message survives.

## Verified clean
- **security** — no issues. Read `input-dispatch.ts`, `turn-handler.ts`, `subagent-executor.ts` at ref.
- **api-compat** — no issues. `pendingSubmissions`/`SubmissionPayload`/`handleEnter` have no consumers outside `terminal-compositor*.ts` + `input-surface.ts` (repo-wide grep); no external breaking surface.
- **perf-observability** — no issues. Swaps one O(1) mutation for another O(1) array literal on a per-Enter path; no loop/timer/logging change.

## What was not checked
- Ref read `b262dfb1b`, confirmed identical to the local checkout used for all reads/greps — all citations are `diff-context` or direct file-state reads at that ref, none diff-only or fabricated. Citation verification 100% verified, 0 dropped, 0 downgraded.
- Ran `pnpm exec vitest run src/cli/terminal-compositor.test.ts` (249/249 passed) and `pnpm exec tsc --noEmit -p .` (clean), consistent with the PR's stated verification — but did not run the full 10,643-test suite, and no agent executed the repro against a real TTY or a real subagent-backed turn (the shadow-verify repro was a direct unit-level `TerminalCompositor` instantiation, not an end-to-end CLI session).
- Did not review the three explicitly-declared "not included" follow-ups (interrupt affordance copy, telemetry event, repaint-starvation) — out of scope per the PR author.

---
_Read-only review — no files, commits, PR body, labels, or merge state were modified. Verdict: **DO NOT MERGE** (recommendation only)._

griffinwork40 (commented):

---

# Code Review — PR #403

**griffinwork40/agent-afk** · `fix(tui): coalesce post-ESC type-ahead so soft-stop doesn't strand input one turn behind`
**Reviewed ref:** `b262dfb1baf217a08bf7be8b8fff3a30f392da09` · **Regime:** LIGHT (2 files, 76+/3-)

## Merge Recommendation: 🛑 DO NOT MERGE

## Blocking findings

**1. 🔴 CRITICAL · correctness · `terminal-compositor.input-dispatch.ts:647-651`** (ref `b262dfb1b`)
```ts
if (self.softStopped) {
  self.pendingSubmissions = [payload];   // ← wholesale replace, not append
} else {
  self.pendingSubmissions.push(payload);
}
```
Traced repro: type `alpha`+Enter (push → `[alpha]`), type `beta`+Enter (push → `[alpha,beta]`) — ordinary mid-turn type-ahead, `softStopped` still false. Press ESC — `handleEscape` sets `softStopped=true` and (by its own doc-comment, line 318-319) does *not* touch `pendingSubmissions`; queue stays `[alpha,beta]`. `softStopped` is cleared only later, at the next `setInputMode('idle')` — nothing in the ESC call chain (`turn-handler.ts`'s `onSoftStop` handler only sets a flag and fires an async `session.interrupt()`) clears it synchronously. So a further Enter (`gamma`) arriving anytime in that window hits the `if` branch: `pendingSubmissions = [gamma]`. **`alpha` and `beta` — messages the user explicitly typed and confirmed — are silently and permanently destroyed.** No recovery path: Backspace never dequeues (by design), ↑-recall only reads the live (now-replaced) array, history isn't populated pre-dispatch.

Verified three independent ways: (1) a live `vitest` repro against the pinned worktree (a pre-ESC `'preESC'` entry vanished the instant a post-ESC `'postESC'` Enter fired — `pendingCount` stayed `1`, only `postESC` reached `onSubmit`); (2) an independent research agent re-traced `handleEnter`/`handleEscape`/`setInputMode` line-by-line and found no guard; (3) a second adversarial pass explicitly instructed to *refute* it (checked for length guards, intervening handlers, queue-emptiness preconditions on ESC) — found none, returned CONFIRMED.

**2. 🟠 HIGH · spec-compliance · `terminal-compositor.input-dispatch.ts:644-646`** (same root cause, different acceptance surface)
The PR body's own guarantee — *"normal mid-turn type-ahead is untouched... still accumulates every message for sequential-turn delivery"* — is violated by this diff's own replace logic. Undisclosed: the PR's "Not included" follow-ups list (interrupt affordance, telemetry, repaint-starvation) never mentions this interaction. `handleEscape`'s adjacent doc-comment (line 320-322, "already-queued messages... left untouched") is also now false under this new branch and was not updated.

## Other findings
| Sev | Dim | Location | Finding |
|---|---|---|---|
| Low | test-coverage | `test.ts:1804-1857` | Neither new test seeds a pre-ESC queued entry before firing ESC — the critical defect has **zero** coverage and would not be caught by this PR's own "10,643 passed" run. The closest existing test (`'ESC after Enter leaves the queued message in the FIFO'`, line 1423) queues exactly one message and never presses Enter again post-ESC — never reaches the destructive branch. |
| Low | perf-observability | `input-dispatch.ts:648` | No log/signal when N>0 queued payloads are discarded by the replace; only observable trace is the `[N queued]` render suffix silently decrementing. |
| — | security | — | No issues found — pure client-side TUI state, no I/O/injection/secrets. |
| — | api-compat | — | No issues found — `SubmissionPayload`/`pendingSubmissions`/`softStopped` confined to `src/cli/terminal-compositor*.ts` + `input-surface.ts`; zero presence in `src/index.ts` exports or `PluginApi`. |

## Suggested fix (not applied by this review)
Scope the replace to the post-soft-stop tail instead of the whole array — e.g. snapshot a floor index when `softStopped` is set (`softStopFloor = pendingSubmissions.length` in `handleEscape`) and in the `if` branch keep `pendingSubmissions.slice(0, softStopFloor)` and coalesce only entries after it — plus a regression test asserting a pre-ESC entry survives a post-ESC Enter.

griffinwork40:
## PR #403 Review — **MERGE**

**PR:** `fix(tui): coalesce post-ESC type-ahead so soft-stop doesn't strand input one turn behind`
**Verdict:** ✅ Merge · No blocking findings.

The diff faithfully implements its stated intent: post-ESC Enters coalesce to last-wins (truncate-then-push at `softStopQueueBase`), preserving pre-ESC payloads while carrying exactly one message into the next turn. The committed implementation is actually **better than the PR body's snippet** — it uses `length = base; push(payload)` rather than the body's `self.pendingSubmissions = [payload]` (which would have dropped pre-ESC entries). Worth reconciling the PR description with the actual diff before merge.

### Findings (2 × low · tests-additive)

| # | Sev | Dim | Where | Gap |
|---|---|---|---|---|
| 1 | low | test-coverage | `terminal-compositor.test.ts:1801-1926` | No test locks the **double-ESC re-snapshot** invariant — a 2nd ESC during a lingering soft-stop has its once-only guard return before snapshotting `softStopQueueBase`; should be asserted (suggested: `ESC → A,B → ESC(no-op) → C` ⇒ `[C]`). |
| 2 | low | test-coverage | same | No coalesce test exercises a `displayText`/paste-placeholder or non-empty `attachments` payload — soft-stop tests use only `{ text, attachments: [] }`. |

### Dimensions
- **security** / **api-compat** / **correctness** / **spec-compliance** / **perf-observability** — all clean. `softStopQueueBase: number` is internal-only; every `softStopped = false` site resets the base to 0 (idle→streaming, →idle, `resetState`, init), so no stale-base leak.

### Not checked
- Did not run the suite (inferred from the agent's contract analysis); author reports 10,643 pass / 0 fail + clean lint.
- Did not exercise the real async subagent-teardown path (tests reproduce via synchronous mode flips).
- Read-only review — merging/fixing is out of scope.

Full review delivered by the `review` skill above. The two low findings are optional hardening add-tests, not merge blockers.

---

PR #404, opened 2026-07-04
fix(skills): make SKILL.md context: authoritative; drop DEFAULT_FORK_SKILLS

griffinwork40:
## Summary
Follow-up to #399, addressing the two review findings on its fork-by-name half. Removes `DEFAULT_FORK_SKILLS` so each skill's own `SKILL.md` `context:` frontmatter is the single source of truth for fork-vs-load.

## Findings addressed

**1. `DEFAULT_FORK_SKILLS` generic-name blast radius.** The set force-forked skills by name regardless of frontmatter. Verified against the code:
- All eight names are bundled *plugin* skills (not registry skills), so they resolve on the plugin-body path where the override applied; `discoverPluginSkillBodies` is first-wins **project → user → bundled**, so a same-named third-party project/user skill wins and *was* reachable by the force-fork — confirming it could hijack an unrelated skill's explicit `context: load`.
- Every bundled copy already declares `context: fork`, so the override was a **no-op for the copies it aimed to protect**; it only ever fired on a non-bundled copy, where a stale copy of ours is indistinguishable by `pluginPath` from a third party's own skill (so the "key on pluginPath" alternative can't disambiguate either).
- Only `shadow-verify` (parsed by `diagnose`'s verifier) and `ground-state` (read-only enforcement runs only on the fork path) had any real fork dependency — and both keep it via their own `context: fork` frontmatter.

**2. Stale "forks ONLY when frontmatter declares `context: fork`" comment.** Removing the override makes that statement true again, so it's resolved by making the code match the comment (reverted the executor/test comments to that wording).

The spawn-cwd half of #399 is untouched.

## Changes
- `nesting.ts` — remove `DEFAULT_FORK_SKILLS`.
- `skill-executor.ts` — drop the import, the `forcedFork` branch, and the warn; the gate is now just `pluginSkill.context === 'fork'`.
- `skill-load-mode.test.ts` — replace the force-fork-by-name tests with guards that (a) a shadowing copy of a bundled fork skill lacking `context: fork` now **loads**, and (b) an explicit `context: load` is respected.

## Verification
- `pnpm lint` (strict `tsc --noEmit`) → clean.
- `skill-load-mode.test.ts` (15) + `skill-executor.test.ts` (75) → **90 passed**.

## Note (not changed)
`DEFAULT_READ_ONLY_SKILLS` (ground-state read-only enforcement) is a separate name-keyed override left intact — it wasn't part of this finding. With fork now frontmatter-only, that guarantee applies only when the skill's `SKILL.md` declares `context: fork` (bundled `ground-state` does). Flag if you want the same "leave it to the skill file" treatment there.

### DIFF
--- a/src/agent/tools/nesting.ts
+++ b/src/agent/tools/nesting.ts
@@ -132,29 +132,6 @@ export const RECON_ALLOWED_TOOLS: readonly string[] = [
 // prose). Add a name here only for skills that are genuinely read-only recon.
 export const DEFAULT_READ_ONLY_SKILLS: ReadonlySet<string> = new Set(['ground-state']);
 
-// Skills forced to FORK execution by NAME, independent of their SKILL.md
-// frontmatter. Same rationale as DEFAULT_READ_ONLY_SKILLS: first-registrant-
-// wins registration means a user/project copy of a bundled skill can shadow
-// it, and a stale copy missing `context: fork` silently degrades to load
-// mode — the caller gets instruction text instead of the structured envelope
-// (e.g. diagnose's auto-verify parses shadow-verify's JSON), and name-keyed
-// read-only enforcement is bypassed because it only runs on the forked path.
-// Seeded with every bundled skill that declares `context: fork`
-// (src/bundled-plugins/awa-bundled/skills/*). When this set forces a fork on
-// a copy whose frontmatter lacked `context: fork`, the executor emits a
-// warning naming the skill so a deliberate local load-mode edit is never
-// silently overridden.
-export const DEFAULT_FORK_SKILLS: ReadonlySet<string> = new Set([
-  'shadow-verify',
-  'review',
-  'research',
-  'ground-state',
-  'spec',
-  'ship',
-  'simplify',
-  'refactor',
-]);
-
 /**
  * Bootstrap-time options captured by closure into the factory returned by
  * {@link createChildProviderFactory}. Held at factory-construction (not
--- a/src/agent/tools/skill-executor.ts
+++ b/src/agent/tools/skill-executor.ts
@@ -23,7 +23,6 @@ import { collectSkillEntries, discoverPluginSkillBodies, type PluginSkillBody } 
 import type { SdkPluginConfig } from '../types/sdk-types.js';
 import {
   DEFAULT_MAX_NESTING_DEPTH,
-  DEFAULT_FORK_SKILLS,
   DEFAULT_READ_ONLY_SKILLS,
   RECON_ALLOWED_TOOLS,
   buildReadOnlyReconProvider,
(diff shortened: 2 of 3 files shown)


griffinwork40:
I now have everything I need. The diff is small and I've verified all the critical claims inline (bundled skill frontmatter, no stale references, docs consistency). Let me write the review.

---

## PR #404 Review — `fix(skills): make SKILL.md context: authoritative; drop DEFAULT_FORK_SKILLS`

**+38 / -82 / 3 files** — Working tree **NOT** mutated (review is read-only).

---

### Summary

Removes the `DEFAULT_FORK_SKILLS` name-keyed fork-enforcement set from `nesting.ts` and the `forcedFork` logic in `skill-executor.ts`, making the SKILL.md `context:` frontmatter field the single source of truth for fork vs load mode. A shadowing copy that lacks `context: fork` now loads in-context instead of being force-forked by name. Tests are rewritten to assert the new behavior.

---

### Correctness — **PASS ✅**

**Claim verified: all 8 bundled skills explicitly declare `context: fork` in their own SKILL.md.**

```
shadow-verify  → context: fork ✓
review         → context: fork ✓
research       → context: fork ✓
ground-state   → context: fork ✓
spec           → context: fork ✓
ship           → context: fork ✓
simplify       → context: fork ✓
refactor       → context: fork ✓
```

Removing the name-keyed override does **not** break the bundled skills — they fork via their own frontmatter. The `DEFAULT_FORK_SKILLS` set was only a defense-in-depth against **stale shadowing copies** (first-registrant-wins registration lets a user/project copy shadow a bundled skill), and the PR accepts the deliberate trade-off: a stale copy lacking `context: fork` now loads in-context.

**`DEFAULT_READ_ONLY_SKILLS` correctly preserved.** The read-only name-keyed enforcement (`nesting.ts:133`) remains unchanged — it only applies on the forked path (inside the `pluginSkill.context === 'fork'` branch), so a correctly-configured bundled `ground-state` (which declares both `read-only: true` and `context: fork`) still gets read-only child enforcement. This asymmetry is principled: forcing read-only on a third-party same-named skill is strictly safer, whereas forcing fork was a bigger semantic change (different context, different output shape) that could hijack an unrelated skill.

**Docs already state the new rule.** `docs/skill-load-mode.md:104`:

> Rule for frontmatter skills (plugin + disk): **fork iff `context: fork`**; everything else (absent, `load`, or an unrecognized value) loads in-context.

The PR brings the implementation in line with the documented contract. No doc update needed.

---

### API-Compat — **PASS ✅**

`DEFAULT_FORK_SKILLS` was an exported const from `nesting.ts`. Grep confirms no remaining source-tree references outside `dist/` (generated build artifacts, rebuilt on `pnpm build`) and `.afk-worktrees/` (separate worktrees, expected). No in-repo breakage. The package exports `afk` as a CLI binary, not a library, so external import breakage is not a concern.

---

### Test-Coverage — **PASS ✅**

Tests are rewritten to match the new behavior:

| New test | What it asserts |
|----------|----------------|
| "a plugin skill without context: fork loads in-context by default" | Basic default-loads case; no fork, no warn |
| "SKILL.md context is authoritative: a shadowing copy of a bundled fork skill lacking context: fork LOADS" | Regression guard for `DEFAULT_FORK_SKILLS` removal — `shadow-verify` copy without `context: fork` now loads |
| "respects an explicit context: load even for a name that shadows a bundled skill" | Third-party `review` with `context: load` loads, never hijacked into fork |
| "a plugin skill with context: fork still forks a subagent" (retained) | Explicit `context: fork` still forks |

The old tests ("forces FORK by name", "does NOT warn when DEFAULT_FORK_SKILLS skill declares context: fork", "a plugin skill NOT in DEFAULT_FORK_SKILLS still loads") are correctly replaced — they asserted behavior that no longer exists.

---

### Security — **PASS ✅**

No security implications. The change affects execution-mode selection (fork vs load), not authorization or data flow. The `console.warn` removal is a minor observability change — but it warned about a scenario that is now the desired behavior, so the silence is correct.

---

### Perf-Observability — **PASS ✅**

No performance impact. The removed code was a `Set.has()` check and a conditional `console.warn`. No new hot paths introduced.

---

1. **Low — stale-copy read-only regression for `ground-state`.** A shadowing copy of `ground-state` that lacks `context: fork` will now load in-context. Since `DEFAULT_READ_ONLY_SKILLS.has('ground-state')` only runs inside the `context === 'fork'` branch, the loaded copy runs in the caller's context with the full write surface (no child to restrict). This is the **deliberate trade-off** of making `context:` authoritative and is documented in the test comments. The bundled `ground-state` is safe — it declares both `read-only: true` and `context: fork`.

2. **Cosmetic — pre-existing doc path inaccuracy.** `docs/skill-load-mode.md:115` says "Bundled skills shipped under `awa-bundled/`" but the actual path is `src/bundled-plugins/awa-bundled/skills/*/SKILL.md`. Not introduced by this PR.

---

### Verdict: **MERGE ✅**

Clean, well-scoped, behaviorally correct. The `DEFAULT_FORK_SKILLS` override was a defensive measure that infringed on the authority of SKILL.md frontmatter and could hijack unrelated third-party skills sharing a name. Removing it makes `context:` the single source of truth, matching the documented contract. All 8 bundled skills self-declare `context: fork`, so no functional regression for the canonical path. Tests are rewritten to guard the new behavior with good coverage of the regression scenario.

No blockers. No changes requested.

---

PR #405, opened 2026-07-04
fix(tui): stop double-rendering finished subagents in the REPL overlay

griffinwork40:
## Symptom

In the REPL, parallel subagent work (`compose` / `devils-advocate`) shows content **swallowed / overlapped between the live compositor overlay and scrollback**: a finished node renders *twice* — once as the live `◉ → Agent(<node>) … Done` tree and once as a compact `✓ <node> · <time>` line — and stacked ghost `◉` markers pile up. (Reported from a devils-advocate run with `da-paranoid`/`da-pragmatist`/`da-architect`.)

## Root cause

A finished subagent is rendered through **two uncoordinated channels**:

- **Channel A** — the `ToolLane` overlay + scrollback `→ Agent(…) Done` tree, driven by the *parent turn's* SDK events (`tool_use`/`tool_result`/`message_stop`). Single-writer via `OverlayComposer.setOverlay`.
- **Channel B** — the **SubagentStop hook's** compact `✓ <agentType> · <duration>` line (`formatSubagentCompletion`, `progress-banner.ts`), driven by **each subagent's own lifecycle end**, written via an uncoordinated `completionWriter.fn` → `compositor.commitAbove`.

Under parallel dispatch, sibling subagents finish on independent schedules, so Channel B's `commitAbove` interleaves with Channel A's `setOverlay` mid-frame. Result: the node is double-rendered, and the compositor's frame row-accounting (`logUpdate.measure()`/`topRow`) can't reconcile a foreign single-line commit against a concurrently-repainted tall overlay — leaving ghost `◉` rows and overwriting ("swallowing") committed lines.

Channel B is a **chat-era signal** that predates the rich `ToolLane` subagent tree; it was never suppressed once the tree superseded it for foreground dispatches. (The one-shot `chat` surface has its own `console.log` wiring and is unaffected; background-job *adoption* uses a different message.)

## Fix

Gate Channel B behind a turn-scoped `CompletionWriter.suppressSubagentCompletion` flag, bracketed by `turn-handler`'s `armAndWire` (set true when a compositor is armed) and its finally block (reset false). The `✓ <node>` line is muted **only while a foreground turn's live overlay owns subagent rendering** — where the `ToolLane` already commits the `→ Agent(…) Done` tree. Between turns (e.g. backgrounded-job completions) and on non-TTY `chat`, the line still surfaces.

The gate is extracted into a unit-testable `emitSubagentCompletion()` with a documented invariant.

## Testing

- `pnpm lint` clean.
- New regression tests in `progress-banner.test.ts` cover the **concurrent-siblings** case (previously uncovered — existing `subagent-gap` repros are strictly sequential): suppressed while overlay live → no emit for any sibling; unsuppressed between turns → late/background completion surfaces; undefined flag → default (emit).
- Full suite green locally (rebased onto `main` @ v5.20.0).

## Out of scope

The header-count vs `Done`-count mismatch (`5 tools` header vs `Done (10 tools …)`) is a **separate, documented** artifact — a mid-flight snapshot in `formatAgentSummary` (`subagent-block-gap.repro.test.ts` annotates it "STALE count (mid-flight capture)"). Not addressed here.

### DIFF
--- a/src/cli/commands/interactive/bootstrap.ts
+++ b/src/cli/commands/interactive/bootstrap.ts
@@ -37,7 +37,7 @@ import {
 } from '../../../agent/_lib/trusted-skill-events.js';
 import { formatTrustedSkillCompletion, formatTrustedSkillInFlight } from '../../trusted-skill-badge.js';
 import type { TrustedSkillResult } from '../../../agent/trusted-skill-result.js';
-import { formatSubagentCompletion } from './progress-banner.js';
+import { emitSubagentCompletion } from './progress-banner.js';
 import { ContextSampler } from '../../context-sampler.js';
 import { SubagentManager } from '../../../agent/subagent.js';
 import { SubagentExecutor } from '../../../agent/tools/subagent-executor.js';
@@ -578,7 +578,7 @@ export async function bootstrapSession(
   // path-approval hook fails open until then (mirroring `setAllowDirDispatcher`
   // wiring order).
   const hookRegistryBundle = createDefaultHookRegistry(
-    (info) => { completionWriter.fn(formatSubagentCompletion(info)); },
+    (info) => { emitSubagentCompletion(completionWriter, info); },
     'cli',
     sharedMemoryStore,
     () => stats.permissionMode,
--- a/src/cli/commands/interactive/progress-banner.test.ts
+++ b/src/cli/commands/interactive/progress-banner.test.ts
@@ -22,8 +22,11 @@ import {
   formatProgressBanner,
   formatProgressSummary,
   formatSubagentCompletion,
+  emitSubagentCompletion,
 } from './progress-banner.js';
 import type { ProgressEvent } from '../../../agent/types.js';
+import type { CompletionWriter } from './shared.js';
+import type { SubagentCompleteInfo } from '../../../agent/default-hook-registry.js';
 
 // Force color output so ANSI sequences are present in the rendered strings and
 // the ANSI-aware truncation path is exercised (not just plain-text slicing).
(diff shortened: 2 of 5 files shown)


griffinwork40:
I've independently verified all the key claims from the diff artifact and the existing codebase. Here's my review.

---

## PR #405 Review — `fix(tui): stop double-rendering finished subagents in the REPL overlay`

**Verdict: MERGE ✅**

### What the fix does

The REPL has two channels for subagent completion:
- **Channel A** — the ToolLane tree (`→ Agent(…) Done`), rendered by the live overlay's `OverlayComposer`.
- **Channel B** — the SubagentStop hook callback, which calls `completionWriter.fn(formatSubagentCompletion(info))` → `compositor.commitAbove(line)`.

When a compositor is armed (TTY), both channels fire for the same finished subagent. Channel B's uncoordinated `commitAbove` races the OverlayComposer's `setOverlay` repaint, corrupting frame row-accounting (ghost `◉` markers, swallowed committed lines). Parallel dispatch (`compose`, `devils-advocate`) maximizes the interleaving because siblings finish on independent schedules.

The fix adds a turn-scoped `suppressSubagentCompletion` boolean to `CompletionWriter`:
- **Armed** in `armAndWire` — only when `armedCompositor` exists (TTY path), at `turn-handler.ts:196`.
- **Reset** in the `finally` block — at `turn-handler.ts:~760`, unconditionally within `if (completionWriter)`.
- **Checked** in the extracted `emitSubagentCompletion()` gate function — drops the line when `true`, passes through when `false`/`undefined`.

### What I verified (inline, against the diff artifact + existing codebase)

| Claim | Verification |
|-------|-------------|
| Gate function checks flag and drops when truthy | `progress-banner.ts:152` — `if (writer.suppressSubagentCompletion) return;` — handles `undefined` correctly (falsy) ✓ |
| Flag armed only on TTY (compositor exists) | `turn-handler.ts:194` — inside `if (completionWriter && armedCompositor)` block ✓ |
| Flag reset in `finally`, can't get stuck `true` | `turn-handler.ts:741` is a proper JS `finally` (after `catch` at 736); reset at ~757 inside `if (completionWriter)` ✓ |
| `armAndWire` called from within `try` | Lines 295 (initial) and 573 (resume swap) — both inside the `try` that the `finally` closes ✓ |
| `chat.ts` (one-shot) unaffected | `chat.ts:690` uses `console.log(formatSubagentCompletion(info))` directly — never touches `CompletionWriter` or `emitSubagentCompletion` ✓ |
| `formatSubagentCompletion` still exported | Yes — `chat.ts:20` and 6 test files still import it; the function itself is unchanged ✓ |
| No other production callers bypass the gate | grep confirms only `bootstrap.ts:581` (REPL, now gated) and `chat.ts:690` (console.log, no compositor) render the completion line ✓ |
| Type match: `SubagentCompleteInfo` | `default-hook-registry.ts:28-33` defines `{ subagentId, status, durationMs?, agentType? }` — matches `emitSubagentCompletion` signature and test fixtures ✓ |
| `CompletionWriter.suppressSubagentCompletion` is optional | `shared.ts:179` — `suppressSubagentCompletion?: boolean` — backward-compatible, existing stubs unaffected ✓ |
| Hook only fires for terminal statuses | `default-hook-registry.ts:210` — filters out `idle`/`running` before calling `onSubagentComplete` ✓ |
| Tests cover gate logic | 4 new cases: not-suppressed emits, suppressed drops, 3 concurrent siblings all dropped + late completion surfaces after reset, undefined flag preserves default ✓ |

### Findings

**No blocking issues.** One non-blocking finding:

| Severity | Dimension | Finding |
|----------|-----------|---------|
| low | test-coverage | `progress-banner.test.ts` thoroughly covers `emitSubagentCompletion`'s flag logic in isolation, but no test asserts that `runTurn` arms the flag on a TTY turn *and* resets it in `finally`. A regression that drops the reset (flag stuck `true` → backgrounded-job completions never surface between turns) would slip through. `turn-handler.test.ts` already exists — a case using the existing stub-compositor harness would close this gap. |

### Notes

- The skill's Wave-1 sub-agent dispatch hit a 401 (OpenAI key routed to an Anthropic endpoint — an infra misconfig on the child runtime, not a diff issue). I verified all citations inline against the diff artifact and the `main`-branch codebase structure (the working tree is on `main`; the PR changes are on `afk/debug-compositor-scrollback`). The logic verification is complete; the full test suite was not run against the PR branch (working tree is read-only per review constraints).
- The fix is minimal (+131/-3), well-documented (the `History:` comment block in `progress-banner.ts` and the `CompletionWriter` doc comment explain the root cause and the bracketing semantics), and backward-compatible (optional field, preserved export, untouched `chat` path).

**Done**
- Reviewed PR #405 (`afk/debug-compositor-scrollback → main`, +131/-3, 5 files) — a fix for double-rendering finished subagent banners in the REPL overlay.
- Evidence: diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/unbound-d17f0c8d48fba688/pr-405.diff`; inline verification against `turn-handler.ts` (arm at :196, finally at :741, reset at ~:757), `progress-banner.ts` (gate at :152), `shared.ts` (optional field at :179), `bootstrap.ts` (callback swap at :581), `chat.ts` (unaffected direct console.log at :690), `default-hook-registry.ts` (type def at :28, terminal-status filter at :210).
- Verdict: **MERGE** — no blocking issues; one low-severity test-coverage gap (arm/reset lifecycle untested in `turn-handler.test.ts`).
- No files were written or commits made — review is read-only; the 5 dirty working-tree changes on `main` were preserved.

---

PR #412, opened 2026-07-04
fix(openai): fall back to AFK_OPENAI_BASE_URL so deep subagents don't hit api.openai.com

griffinwork40:
## Motivation

An OpenAI-routed subagent dispatched **deep** inside a skill chain (e.g. `/review` → review → review → `agent(research-agent)`) fails with `401 Incorrect API key provided: sk-…` — and the key is *valid*. Diagnosis: that exact string is OpenAI's canonical error, returned when a **non-OpenAI key is sent to `api.openai.com`**. The subagent's request went to the default OpenAI endpoint instead of the configured `AFK_OPENAI_BASE_URL` (e.g. opencode), so a valid opencode key was (correctly) rejected there.

## Root cause

The OpenAI-compatible client's endpoint is resolved as `config.openaiBaseUrl ?? providerOpts.baseURL` (`openai-compatible/index.ts`), defaulting to `https://api.openai.com/v1` when both are undefined. But:

- `config.openaiBaseUrl` is only set on the **main** session (`loadConfig`) or a **per-slot** binding (`applySlotCredentials`). It is **not threaded** through the child-config chain: `SubagentExecutorContext.defaultConfig` is `Pick<AgentConfig,'apiKey'|'systemPrompt'|'baseUrl'>` — no `openaiBaseUrl` — and the skill/compose executors forward `apiKey` + the Anthropic `baseUrl` but not `openaiBaseUrl`.
- `resolveProvider` constructs `new OpenAICompatibleProvider(ctorOpts)` with no `baseURL`, so `providerOpts.baseURL` is undefined too.

Net: a deep grandchild loses the endpoint and silently hits `api.openai.com`. Shallow subagents work (their child config still carries it), which is why the failure only appears deep in a fan-out.

## Fix

Add a **third precedence tier** at the single resolution chokepoint, extracted into a pure, tested helper `resolveEffectiveOpenAIBaseUrl()`:

```
config.openaiBaseUrl  >  providerOpts.baseURL  >  AFK_OPENAI_BASE_URL (normalized)  >  SDK default (api.openai.com)
```

This reads the env at **query time**, so the endpoint is resilient to the child-config threading gap regardless of nesting depth — and it makes the code honor its own docstring, which already claimed `providerOpts.baseURL` comes "from AFK_OPENAI_BASE_URL". A trailing `/chat/completions` is stripped (mirrors `normalizeOpenAIBaseUrl`; kept local to avoid an agent→cli import cycle).

Chosen over threading `openaiBaseUrl` through every executor (subagent/skill/compose) because the chokepoint fixes **all** paths at once with far less surface and risk.

## Safety / back-compat

- Only activates when `AFK_OPENAI_BASE_URL` is set **and** no higher-precedence base URL exists — i.e. the operator has globally redirected OpenAI traffic, so falling back to it (rather than `api.openai.com`) is the correct behavior, never a regression.
- Per-session and per-slot base URLs still win (precedence unchanged).
- The per-slot custom-`baseUrl` case is already covered by `applySlotCredentials`; this handles the global-env case.
- Effective immediately (query-time resolution).

## Tests

New `base-url.test.ts` (7 cases: precedence order, the lost-base-URL fallback regression guard, `/chat/completions` strip, whitespace/empty handling). `pnpm lint` clean; full suite green (10723 passed / 14 skipped).

### DIFF
--- a/src/agent/providers/openai-compatible/base-url.test.ts
+++ b/src/agent/providers/openai-compatible/base-url.test.ts
@@ -1,0 +1,46 @@
+import { describe, it, expect } from 'vitest';
+import { resolveEffectiveOpenAIBaseUrl } from './base-url.js';
+
+describe('resolveEffectiveOpenAIBaseUrl', () => {
+  it('prefers per-session config.openaiBaseUrl over ctor and env', () => {
+    expect(
+      resolveEffectiveOpenAIBaseUrl('https://cfg/v1', 'https://ctor/v1', 'https://env/v1'),
+    ).toBe('https://cfg/v1');
+  });
+
+  it('falls back to the construction-time baseURL when config is undefined', () => {
+    expect(resolveEffectiveOpenAIBaseUrl(undefined, 'https://ctor/v1', 'https://env/v1')).toBe(
+      'https://ctor/v1',
+    );
+  });
+
+  it('falls back to AFK_OPENAI_BASE_URL when config and ctor are undefined (the lost-base-URL fix)', () => {
+    // Regression guard: a deep OpenAI-routed subagent whose child config never
+    // threaded openaiBaseUrl must still reach the configured endpoint, NOT
+    // api.openai.com (where a non-OpenAI key 401s as "Incorrect API key").
+    expect(resolveEffectiveOpenAIBaseUrl(undefined, undefined, 'https://opencode.ai/zen/go/v1')).toBe(
+      'https://opencode.ai/zen/go/v1',
+    );
+  });
+
+  it('returns undefined (→ SDK default api.openai.com) when nothing is set', () => {
+    expect(resolveEffectiveOpenAIBaseUrl(undefined, undefined, undefined)).toBeUndefined();
+  });
+
+  it('strips a trailing /chat/completions from the env fallback (SDK appends it)', () => {
+    expect(
+      resolveEffectiveOpenAIBaseUrl(undefined, undefined, 'https://x/v1/chat/completions'),
+    ).toBe('https://x/v1');
+  });
+
+  it('treats a whitespace-only env value as absent', () => {
+    expect(resolveEffectiveOpenAIBaseUrl(undefined, undefined, '   ')).toBeUndefined();
+  });
+
+  it('preserves an empty-string config value (parity with the prior ?? behavior)', () => {
+    // The original `config.openaiBaseUrl ?? providerOpts.baseURL` kept a set-but-
+    // empty config value; `!== undefined` matches that (config is never '' in
+    // practice — loadConfig gates on a truthy env — but keep the semantics stable).
+    expect(resolveEffectiveOpenAIBaseUrl('', 'https://ctor/v1', 'https://env/v1')).toBe('');
+  });
+});
--- a/src/agent/providers/openai-compatible/base-url.ts
+++ b/src/agent/providers/openai-compatible/base-url.ts
@@ -1,0 +1,43 @@
+/**
+ * Effective baseURL resolution for the OpenAI-compatible client.
+ *
+ * Precedence: per-session `config.openaiBaseUrl` > construction-time
+ * `providerOpts.baseURL` > global `AFK_OPENAI_BASE_URL` (normalized) >
+ * undefined (→ the OpenAI SDK default `https://api.openai.com/v1`).
+ *
+ * The env tier is a safety net. A deep child/grandchild session dispatched
+ * through the skill / subagent / compose executors inherits a child config that
+ * does NOT thread `openaiBaseUrl` (SubagentExecutorContext.defaultConfig is
+ * `Pick<AgentConfig,'apiKey'|'systemPrompt'|'baseUrl'>` — no openaiBaseUrl).
+ * Without this fallback such a session silently POSTs to api.openai.com and a
+ * non-OpenAI key (e.g. an opencode key) is rejected as "Incorrect API key".
+ * Reading the env here — not only at config load — makes the endpoint resilient
+ * to that gap. A trailing `/chat/completions` is stripped (the SDK appends it);
+ * this mirrors `normalizeOpenAIBaseUrl` (cli/config.ts) but is kept local to
+ * avoid an agent→cli import cycle, and without the config-time stderr warning.
+ *
+ * @module agent/providers/openai-compatible/base-url
+ */
+
+/**
+ * Pick the baseURL to hand the OpenAI SDK, or `undefined` to let it use its
+ * own default. See the module docstring for the precedence rationale.
+ */
+export function resolveEffectiveOpenAIBaseUrl(
+  configOpenaiBaseUrl: string | undefined,
+  ctorBaseUrl: string | undefined,
+  envBaseUrl: string | undefined,
+): string | undefined {
+  if (configOpenaiBaseUrl !== undefined) return configOpenaiBaseUrl;
+  if (ctorBaseUrl !== undefined) return ctorBaseUrl;
+  return normalizeEnvBaseUrl(envBaseUrl);
+}
+
+/** Trim + drop a redundant trailing `/chat/completions`; empty → undefined. */
+function normalizeEnvBaseUrl(raw: string | undefined): string | undefined {
+  if (raw === undefined) return undefined;
+  const trimmed = raw.trim();
+  if (!trimmed) return undefined;
+  const suffix = '/chat/completions';
+  return trimmed.endsWith(suffix) ? trimmed.slice(0, -suffix.length) : trimmed;
+}
--- a/src/agent/providers/openai-compatible/index.ts
+++ b/src/agent/providers/openai-compatible/index.ts
@@ -16,6 +16,8 @@ import path from 'path';
 import { appendFileSync, mkdirSync } from 'fs';
 import { dirname } from 'path';
 import { getSessionGrantsPath } from '../../../paths.js';
+import { env } from '../../../config/env.js';
+import { resolveEffectiveOpenAIBaseUrl } from './base-url.js';
 import type {
   ModelProvider,
   ProviderQuery,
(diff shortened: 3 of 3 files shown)


griffinwork40:
I now have definitive, ref-grounded evidence. `buildSkillRestrictedProvider` at the ref (nesting.ts:409) sets the threaded `openaiBaseUrl` as the provider's **construction-time `baseURL`** — i.e. tier-2 (`providerOpts.baseURL`) in PR 412's own helper. That's the confirmation: on the deep-subagent path, #413 populates tier-2, so PR 412's tier-3 env fallback is never reached.

The review is complete. No critical/high findings exist, and my central claim is first-hand ref-grounded (not a sub-agent assertion), so `/shadow-verify` is not triggered.

---

# Code Review — PR #412

**Target:** PR #412 `fix/openai-compatible-base-url-env-fallback` → main · reviewed ref `c5604e43` (a merge commit) · +104/-5 across 3 files · change type: hotfix · regime: **light**
**Stated intent:** PR #412 title+body — "fall back to AFK_OPENAI_BASE_URL so deep subagents don't hit api.openai.com." Spec-compliance assessed against it.

## Summary

The **code is safe, correct, tested, and lint-clean** — the helper's precedence, empty-string parity, whitespace handling, and `/chat/completions` normalization are all right, and env access goes through the sanctioned typed getter. **But the PR's premise is stale at the merged ref**, and this is the decision-driving finding.

The reviewed ref is a merge commit that pulls in **already-merged PR #413** (`26cd2e1`, "fix(nesting): thread openaiBaseUrl into restricted provider builders so deep OpenAI subagents keep their endpoint"). #413 fixes the *exact* deep-subagent bug PR 412 describes — by the very mechanism PR 412's body says doesn't exist.

## Findings (sorted: semantic → mechanical)

**1 · MEDIUM · high confidence · spec-compliance/correctness · ref:c5604e43**
The PR's stated root cause is false at the merged ref, making the change redundant for its stated purpose.
- The body claims `SubagentExecutorContext.defaultConfig is Pick<AgentConfig,'apiKey'|'systemPrompt'|'baseUrl'>` (no `openaiBaseUrl`) and "not threaded through the child-config chain." At the ref, `subagent-executor.ts:55` reads `Pick<AgentConfig, 'apiKey' | 'systemPrompt' | 'baseUrl' | 'openaiBaseUrl'>` — #413 added it.
- #413 threads it into the deep-subagent provider as the **ctor `baseURL`** (`nesting.ts:409`, file-state citation):
  ```ts
  return new OpenAICompatibleProvider({
    permissions,
    ...(openaiBaseUrl !== undefined ? { baseURL: openaiBaseUrl } : {}),  // #413
  });
  ```
  fed from `this.ctx.defaultConfig.openaiBaseUrl` (`subagent-executor.ts:882`), which is populated at every root entry point (`chat.ts:530`, `daemon.ts:144`, `bootstrap.ts:331`).
- Net: on the deep-subagent path, PR 412's helper returns at **tier-2** (`providerOpts.baseURL`, now set by #413) and **never reaches its tier-3 env fallback**. The env tier fires only when config *and* ctor are both undefined — the exact condition #413 eliminated for that path. PR 412 is therefore pure, currently-unreached defense-in-depth, not the live fix its description claims.
- **Suggestion:** Either close this PR in favor of #413, or rewrite the description — drop the "not threaded" root-cause claim and re-justify explicitly as a query-time backstop — and have the maintainer confirm they want a defensive tier that no known production path currently reaches.

**2 · LOW · high confidence · test-coverage · base-url.test.ts (whole) + index.ts:270-275 · ref:c5604e43 · diff-context**
The pure helper has 7 unit tests, but the wiring — `index.ts` reading `env.AFK_OPENAI_BASE_URL` and passing it as the 3rd argument through to `buildOpts.baseURL` — has no integration test; a reordered/dropped argument would pass all 7 unit tests while regressing the behavior. Absence confirmed at ref: existing `AFK_OPENAI_BASE_URL` tests (`routing.test.ts`, `config.test.ts`) cover provider *routing* and *config* resolution, not the provider's internal baseURL wiring. **Suggestion:** add one test constructing `OpenAICompatibleProvider` with `AFK_OPENAI_BASE_URL` set and both higher tiers undefined, asserting the client baseURL. *(Moot if finding #1 leads to closing the PR.)*

## Verified clean (read at ref c5604e43)
- **security** — pure helper, no secrets; env via sanctioned getter (`env.ts:490` registry, `env.ts:1193` getter).
- **api-compat** — `resolveEffectiveOpenAIBaseUrl` is a new export, sole consumer is `index.ts`; no signatures changed.
- **perf-observability** — 43-line pure function, no perf surface.
- **empty-string parity** — old `?? ` and new `!== undefined` both forward `''` identically. No regression.
- **normalization asymmetry** — env-only stripping is correct: config/ctor are pre-normalized upstream (`cli/config.ts` `normalizeOpenAIBaseUrl`).

- **Citation corrected (`wrong-ref`):** Wave 1 cited `subagent-executor.ts:55` as `Pick<…'baseUrl'>` (no `openaiBaseUrl`) and concluded "root cause verified, correctness clean." That content is the **local working tree** (`main` @ `833b6a5`, which predates #413), **not the reviewed ref** — the agent lacked Bash and fell back to working-tree reads. At `c5604e43` the line includes `| 'openaiBaseUrl'`. Wave 1's correctness/spec-compliance clean bill is superseded by finding #1.
- **Absence claim confirmed-absent:** "no integration test for the wiring" holds at the ref.

## What was not checked
- Did not run the test suite or `pnpm lint` (PR claims green: 10723 passed / 14 skipped).
- Did not decide the product question — whether the maintainer wants defense-in-depth as policy (finding #1 is theirs to resolve).
- **Note for the operator:** your local `main` (`833b6a5`) is divergent/behind origin — it lacks #413, #415, and other recent commits present in the PR's merged ref. All citations above were read from the reviewed ref `c5604e43` via `git show <ref>:<file>`, not the working tree.

## Merge decision: **DO NOT MERGE (as-is)**

Not a code-safety block — the change is safe and correct. It's a **rationale/redundancy block**: the PR's root-cause premise was already resolved by the merged #413, so the env-fallback tier is never reached on the deep-subagent path it targets. Merging as-is would commit an inaccurate rationale to history and add a currently-dead (though safe) tier. Cheap remedy, maintainer's call: **close in favor of #413**, or **reframe the description as explicit defense-in-depth** and merge knowingly.

**Done** — Reviewed PR #412 through the full pipeline (triage → light-regime review agent → inline Wave 1.5 ref-grounded verification → synthesis). Central finding: the fix is redundant for its stated purpose because merged PR #413 already threads `openaiBaseUrl` to deep subagents (tier-2), and the PR body's root cause is stale at the reviewed ref. Evidence cited by `file:line` at ref `c5604e43` (`subagent-executor.ts:55/:882`, `nesting.ts:409`, `chat.ts:530`, `env.ts:490/:1193`); branch topology from `git log 4516b49..c5604e43`. Read-only skill — no files, PR, or git state were mutated. Recommendation: DO NOT MERGE as-is; close in favor of #413 or reframe as defense-in-depth.

---

griffinwork40 (commented):

## Merge decision: MERGE

The fix correctly resolves the stated deep-subagent endpoint-loss bug at the single resolution chokepoint. Precedence and call-site wiring are verified against file-state at the reviewed ref (`c5604e43`); the only flagged issues are low-severity test/observability gaps — none alter behavior under the typed `string | undefined` contract. Regime: **light** (hotfix, 3 files / 139 lines).

### What was verified (against ref `c5604e43`)
- **Correctness** — `resolveEffectiveOpenAIBaseUrl` implements the stated three-tier precedence verbatim (`config.openaiBaseUrl !== undefined` → `providerOpts.baseURL !== undefined` → `normalizeEnvBaseUrl(env)`), `base-url.ts:18-21`. `env.AFK_OPENAI_BASE_URL` is a real getter (`src/config/env.ts:1193`, returns `string | undefined`). Call-site args are in (config, ctor, env) order, matching the helper signature (`index.ts:270-275`).
- **Spec-compliance** — Satisfies the stated intent exactly: a deep OpenAI-routed subagent whose child config never threads `openaiBaseUrl` now resolves `AFK_OPENAI_BASE_URL` at query time instead of falling through to `api.openai.com`. Fix lands at the chokepoint, not threaded through every executor. No unmet requirement; no scope creep.
- **Import cycle** — `base-url.ts` is a pure leaf (zero imports), satisfying the docstring's "kept local to avoid an agent→cli import cycle" claim. `import { env } from '../../../config/env.js'` introduces no cycle (`env.ts` imports nothing from `agent/`).

### Non-blocking findings
1. **api-compat · low** — `base-url.ts:18`: the old `config.openaiBaseUrl ?? …` and the new `configOpenaiBaseUrl !== undefined` diverge only for a `null` value (`null ?? x` → `x`; `null !== undefined` → returns `null`). Unreachable via the typed contract: `openaiBaseUrl?: string` (`config-types.ts:122`) excludes `null`. The `''`-parity case (the only reachable falsy-but-non-nullish value) is covered by test 7 and behaves identically. Optional: widen the guard to `!== undefined && !== null`, or document the `string | undefined`-only invariant.
2. **test-coverage · low** — `base-url.test.ts:1-46`: all 7 cases exercise the pure helper directly; the `index.ts` call site reading `env.AFK_OPENAI_BASE_URL` (the wiring the regression fix actually depends on) has no dedicated integration test. Optional: add a test that sets `AFK_OPENAI_BASE_URL`, dispatches a query with `config.openaiBaseUrl` undefined, and asserts the client receives the env-derived normalized URL.
3. **perf-observability · nit** — `base-url.ts:34-38`: the local normalizer drops the config-time stderr warning that `normalizeOpenAIBaseUrl` (cli/config.ts) emits when stripping a trailing `/chat/completions`, so a misconfigured env value is silently rewritten. Optional: emit a one-shot debug-tier log on the strip branch, or note the loss in the docstring. (Per-request env read cost itself is negligible — a single property-getter lookup per `query()`.)

### What was not checked
- Read against branch HEAD `c5604e43` — citations verified at that ref via content blob fetches.
- The PR's `pnpm lint` clean / 10723 passed claim is **unverified** (read-only review; no suite executed).
- `cli/config.ts:normalizeOpenAIBaseUrl` was not directly read — the "dropped stderr warning" nit rests on the PR docstring's own characterization of it.
- `SubagentExecutorContext.defaultConfig`'s `openaiBaseUrl` absence was inferred from the `AgentConfig` type + fix shape, not re-verified by opening the executor file.
- Stated intent: PR #412 title + body — spec-compliance assessed.

griffinwork40:
Closing in favor of #413, which already landed a fix for this exact deep-subagent endpoint bug.

**Why:** A review against the current (merge-updated) branch head found this change redundant for its stated purpose. PR #413 ("thread openaiBaseUrl into restricted provider builders so deep OpenAI subagents keep their endpoint") added `openaiBaseUrl` to `SubagentExecutorContext.defaultConfig` (`subagent-executor.ts:55`) and threads it into the deep-subagent provider as the construction-time `baseURL` — `buildSkillRestrictedProvider` at `nesting.ts:409` now does `...(openaiBaseUrl !== undefined ? { baseURL: openaiBaseUrl } : {})`. That populates tier-2 (`providerOpts.baseURL`) on exactly the deep-subagent path this PR targets, so the env fallback added here (tier-3 in `resolveEffectiveOpenAIBaseUrl`) is never reached on that path.

**On the code itself:** it is safe and correct — this is a redundancy / stale-rationale close, not a defect. The root-cause premise in the description ("`openaiBaseUrl` is not threaded through the child-config chain") no longer holds post-#413.

**If a query-time env backstop is still wanted** as defense-in-depth for any future un-threaded provider path, it can be reintroduced later with an honest description that frames it as a backstop rather than the primary fix.

PR #414, opened 2026-07-05
fix(anthropic): surface silent 429/503/529 backoff in trace; fix transient rate-limit 429 misclassification

griffinwork40:
## Problem
When Anthropic throttles a session (429 rate limit / 503 / 529 overload) it looks **stuck**:

- The SDK retries these internally, honoring `retry-after`, **inside a single `messages.create` call**. So a throttled turn surfaces only as an abnormally long `model_ttfb` — a real trace showed 39 turns with 16 >30s, 10 >60s, **max 146s** — with **no event**, indistinguishable from a hang. `AFK.md` already notes *"a usage-limit pause has no event (silent gap)."*
- `classifyUsageLimitError` treated **every** 429 as OAuth *subscription* exhaustion, so a transient API-tier rate-limit 429 was parked in the 2-hour poll/pause path under a misleading "usage limit" label.

Diagnosed from a live stuck `/review` subagent (`claude-sonnet-5`).

### #1 — Make backoff observable (behavior-preserving)
- **`tracing-fetch.ts`** (new): `makeTracingFetch` wraps the SDK's `fetch` client option and emits a `rate_limit` witness-trace event on every 429/503/529 response (`status`, `reason`, `retryAfterMs`). Purely observational — the request is forwarded and the `Response` returned untouched (only headers are read), so **retry behavior is unchanged**. Wired at the `query()` client and the OAuth-refresh client rebuild; skipped in local-shim mode.
- **`loop.ts`**: emits the same event on the mid-stream `overloaded_error` retry, which the fetch wrapper can't see (it rides an HTTP 200 SSE body).
- New `SessionPhaseName` `'rate_limit'` (+ Zod schema).

Now `afk trace show` explains a throttled stall instead of hiding it.

### #4 — Fix 429 misclassification
- **`usage-limit.ts`**: `classifyUsageLimitError` returns a new `rate-limit-transient` kind for a 429 carrying a `retry-after`/`retry-after-ms` header and no `|<unix-ts>` (a standard API-tier rate limit), distinct from OAuth subscription exhaustion. `retry-layer.ts` therefore does **not** route it into the 2-hour subscription-pause path — it surfaces as an error after the SDK's own retry-after retries. New `parseRetryAfterMs` helper.

## Testing
- **+15 tests**: `tracing-fetch.test.ts` (5); `usage-limit.test.ts` (+10 for classification + `parseRetryAfterMs`).
- `tsc --noEmit` clean; `audit:sdk:check` / `audit:env:check` / `scan:env:check` all pass.
- Full `anthropic-direct` + `trace` suites green (**618 tests**). No new `@anthropic-ai/sdk` imports; no new `process.env` reads.

## Deferred (follow-ups)
- Surface the throttle in the **spinner/status line** (not just the trace) — needs a side-channel from the fetch layer up to the renderer.
- Optional: a short bounded auto-retry (honoring `retry-after`) for `rate-limit-transient`, to preserve auto-recovery for genuinely brief blips instead of surfacing an error.

### DIFF
--- a/src/agent/providers/anthropic-direct/auth.ts
+++ b/src/agent/providers/anthropic-direct/auth.ts
@@ -58,19 +58,29 @@ export function detectAuthMode(token: string): AuthMode {
  * option. Used by the local-server path (`AFK_LOCAL_BASE_URL`) to point
  * Messages traffic at a self-hosted Anthropic-compatible shim. The SDK
  * appends `/v1/messages` to whatever is passed.
+ *
+ * `fetchImpl`, when provided, is forwarded as the SDK's `fetch` client option
+ * so every HTTP request (including the SDK's own internal retries) flows
+ * through it. Used to inject an observability wrapper (see
+ * {@link makeTracingFetch}) that records 429/503/529 throttling into the
+ * witness trace — otherwise the SDK's silent retry-after backoff is invisible.
  */
 export function buildClientOptions(
   token: string,
   mode: AuthMode,
   baseUrl?: string,
-): ({ authToken: string } | { apiKey: string }) & { baseURL?: string } {
+  fetchImpl?: typeof fetch,
+): ({ authToken: string } | { apiKey: string }) & { baseURL?: string; fetch?: typeof fetch } {
   const base = mode === 'oauth'
     ? { authToken: token }
     : { apiKey: token };
-  if (typeof baseUrl === 'string' && baseUrl.length > 0) {
-    return { ...base, baseURL: baseUrl };
+  const withBase = typeof baseUrl === 'string' && baseUrl.length > 0
+    ? { ...base, baseURL: baseUrl }
+    : base;
+  if (fetchImpl) {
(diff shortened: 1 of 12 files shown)


griffinwork40:
Good catch, addressed in 71f754b.

The `rate_limit` event was landing in the JSONL but the human renderer drops every `session_phase` unless `--all` is passed — so the default `afk trace show` still showed nothing. Fixed by rendering `rate_limit` **before** the `showAll` gate (as a `throttle` line with reason/status/retry-after/source), while the other phases stay `--all`-only latency markers. Also added a `N throttled` count to the summary header.

Default view now:
```
       ... · 0 blocks · 2 throttled · $0.0200
  12:30:25  throttle   rate-limit  429  retry-after 30.0s  (sdk-fetch)
  12:31:35  throttle   overloaded  529  retry-after 5.0s  (mid-stream)
```
(The long `model_ttfb` it explains stays hidden without `--all`.) +4 tests covering default-view render, count present/omitted, and model_ttfb still hidden.

PR #415, opened 2026-07-05
feat: add /thinking slash command for mid-session thinking-UI toggle

griffinwork40:
## Summary

Adds a `/thinking` slash command that lets the operator toggle the thinking-display mode mid-session, without restarting the REPL. Previously the render mode was only settable at boot via the `--thinking-ui <summary|live|off>` CLI flag.

- `/thinking` — print the current mode
- `/thinking live` — streaming preview + finalize summary (default)
- `/thinking summary` — collapsed one-line summary on finalize, no live preview
- `/thinking off` — suppress thinking entirely
- Alias: `/thinking-ui`

## Approach

Follows the same idiom as the existing `/model` command: the handler mutates a field on the shared, per-session `ctx.stats` object (`SessionStats.thinkingUi`), and the REPL loop reads that field at the top of the next turn — so the effect is visible starting on the following turn, matching `/model`'s semantics exactly. No new subsystem, no status-line integration; confirmation is printed via `ctx.out.success`, same as `/model`.

`ThinkingUiMode` (`'summary' | 'live' | 'off'`) moved from `cli/commands/interactive/shared.ts` to the neutral `cli/slash/types.ts` layer (re-exported from the old location for backward compatibility) so the slash-command layer can reference it without an upward import.

`createSessionStats()` now defaults `thinkingUi` to `'live'`; `bootstrapSession()` overrides that default with the operator's explicit `--thinking-ui` flag value when provided. `resetStats()` (used by `/clear`) preserves `thinkingUi` across a clear, for the same reason it already preserves `permissionMode`: it's user-controlled session state, not per-turn statistics.

## Files changed

- `src/cli/slash/commands/thinking.ts` (new) — command implementation
- `src/cli/slash/commands/thinking.test.ts` (new) — 8 tests
- `src/cli/slash/index.ts` — register the command
- `src/cli/slash/types.ts` — canonical `ThinkingUiMode` type + `SessionStats.thinkingUi` field
- `src/cli/slash/session-stats.ts` — default `thinkingUi: 'live'` in `createSessionStats()`; preserve across `resetStats()`
- `src/cli/commands/interactive/bootstrap.ts` — seed `stats.thinkingUi` from the `--thinking-ui` CLI flag
- `src/cli/commands/interactive/loop-iteration.ts` — read `ctx.stats.thinkingUi ?? ctx.options.thinkingUi` per turn
- `src/cli/commands/interactive/shared.ts` — re-export `ThinkingUiMode` from the new canonical location

## Test results

- New tests: `src/cli/slash/commands/thinking.test.ts` — **8/8 passing**
- Full suite: **10754 passed**, 14 skipped (10768 total), 585 test files — all green
- Lint (`tsc --noEmit`): **clean**

### DIFF
--- a/src/cli/commands/interactive/bootstrap.ts
+++ b/src/cli/commands/interactive/bootstrap.ts
@@ -542,6 +542,10 @@ export async function bootstrapSession(
   if (initialPermissionMode !== undefined) {
     stats.permissionMode = initialPermissionMode;
   }
+  // Seed thinking-UI mode from the CLI flag so `/thinking` has a live value
+  // to mutate. `createSessionStats()` already defaults to 'live'; this
+  // overrides with the user's explicit `--thinking-ui` choice (if any).
+  stats.thinkingUi = options.thinkingUi;
   // Stamp the effective working directory on stats so the status line can
   // render it. We capture the same cwd the provider will see: the explicit
   // `extras.cwd` override (e.g. from `--worktree`) when present, else
--- a/src/cli/commands/interactive/loop-iteration.ts
+++ b/src/cli/commands/interactive/loop-iteration.ts
@@ -580,7 +580,7 @@ export async function runInputLoop(
         // transitions.  The bar is a per-session singleton; the callback is
         // safe to call on non-TTY (LoopStageBar.repaint() TTY-gates itself).
         ...(loopStageBar ? { onStageChange: (stage) => loopStageBar!.repaint(stage) } : {}),
-      }, ctx.options.thinkingUi, ctx.completionWriter,
+      }, ctx.stats.thinkingUi ?? ctx.options.thinkingUi, ctx.completionWriter,
         // Surface refs threaded into the per-turn StreamRenderer for the
         // legacy non-borrow path (non-TTY, when surface.getCompositor()
         // is null and the renderer constructs its own compositor). In
--- a/src/cli/commands/interactive/shared.ts
+++ b/src/cli/commands/interactive/shared.ts
@@ -6,7 +6,7 @@ import type { AgentModelInput } from '../../../agent/types.js';
 import type { BackgroundAgentRegistry } from '../../../agent/background-registry.js';
 import type { BackgroundSummarizer } from '../../../agent/background-summarizer.js';
 import type { SubagentControl } from '../../../agent/tools/subagent-executor.js';
-import type { SlashContext, SessionStats, ResumeSwapResult } from '../../slash/types.js';
+import type { SlashContext, SessionStats, ResumeSwapResult, ThinkingUiMode } from '../../slash/types.js';
 import type { StoredSession } from '../../session-store.js';
 import type { StatusLine } from '../../status-line.js';
 import type { ReplRenderer } from './repl-renderer.js';
@@ -212,7 +212,12 @@ function truncate(s: string, max: number): string {
   return codePoints.slice(0, max - 1).join('') + '…';
 }
 
-export type ThinkingUiMode = 'summary' | 'live' | 'off';
+/**
+ * Canonical definition lives in `slash/types.ts` (neutral layer) to avoid the
+ * upward import that would result from defining it here. Re-exported for
+ * backward compat — existing imports from this module continue to work.
+ */
+export type { ThinkingUiMode } from '../../slash/types.js';
 
 export interface CliOptions {
   /**
--- a/src/cli/slash/commands/thinking.test.ts
+++ b/src/cli/slash/commands/thinking.test.ts
@@ -1,0 +1,118 @@
+/**
+ * Tests for the /thinking slash command.
+ *
+ * The command mutates `ctx.stats.thinkingUi` (a shared `SessionStats` field)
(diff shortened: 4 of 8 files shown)


griffinwork40:
## PR #415 Review — `/thinking` slash command

**+226/-3, 8 files | Branch `feat/thinking-slash-command` → `main`**

I dispatched four parallel dimension reviewers (security, correctness, api-compat, test-coverage/perf). Findings across all four reinforce each other. Here is the synthesized verdict.

### Scope
A new `/thinking` slash command exposes the boot-time-only `--thinking-ui <summary|live|off>` flag as a mid-session toggle. It mutates `ctx.stats.thinkingUi` (the shared `SessionStats` object the REPL loop reads at each turn top), mirroring the established `/model` pattern. Changes: one new command module + test, a type relocation (`ThinkingUiMode` from `interactive/shared.ts` → `slash/types.ts` with a back-compat re-export), an optional `thinkingUi?:` field on `SessionStats`, plus two wiring edits in bootstrap + loop-iteration. Purely additive, low blast radius, fully reversible.

### Findings by severity

**CRITICAL / HIGH** — none.

**MEDIUM (test-coverage gaps, not production bugs):**
- **Load-bearing wiring untested.** The PR's core claim — that `loop-iteration.ts` reads `ctx.stats.thinkingUi ?? ctx.options.thinkingUi` and a `/thinking` choice propagates to the `StreamRenderer` on the next turn — is verified only by deduction. The test mutates a mocked `ctx.stats` and asserts the field changed; it never wires that `stats` into `runTurn`/`StreamRenderer`. (`thinking.test.ts` diff:114-178 vs `loop-iteration.ts:583`)
- **Alias dispatch unevaluated.** `/thinking-ui` metadata is asserted, but nothing verifies it routes through `registerAll()` → `thinkingCmd.handler`. Parity with peer tests, but the alias is a behavioral path that's unverified.
- **Bootstrap seeding untested.** `stats.thinkingUi = options.thinkingUi` and the `resetStats` preservation comment are unverified by tests.

**LOW:**
- **Latent default divergence.** `runTurn`'s param fallback is `'summary'` (`turn-handler.ts:45`) while commander, `createSessionStats`, and `/thinking` all default `'live'`. **Pre-existing, not introduced by this PR**, but the new `??` chain makes it more reachable in non-commander/test paths. Worth a follow-up nit (`turn-handler` default → `'live'`, or `bootstrap: options.thinkingUi ?? 'live'`). The correct reviewer flagged this; in production the commander flag always supplies a value, so no live bug.
- **Status-line observability gap.** No persistent indicator of active thinking mode after the `ctx.out.success` confirmation scrolls away (the test even mocks `repaintStatusLine` but never asserts it). Deliberate per the design comment matching `/model`; flag only if operator-visible feedback is desired.
- **`telegram/session-manager.ts:306`** constructs a `SessionStats` literal omitting `thinkingUi`, relying on the new optional field for compilation. Functionally safe (`??` chains guard it), but a future tightening to required would silently break it. The author's JSDoc documents this rationale.

**INFO (no action):**
- **Security:** clean. Input validated against `VALID_MODES` before any use; the pre-validation `as ThinkingUiMode` cast is TS-only and runtime-safe. No concurrency hazard (REPL slash dispatch is synchronous between single-threaded turns). No privileged surface to gate; `palette.brand` / `MODE_DESCRIPTIONS` cannot crash on validated input. No info leak.
- **API compat:** `ThinkingUiMode` move preserves all existing imports via re-export; dependency direction (interactive → slash) is correct and follows the `ResumeSwapResult` precedent; not part of the npm public API. No circular imports.
- **Perf:** single property read at turn top; no allocation. Negligible.
- **Telemetry:** none emitted, parity with `/model`. Not a gap.

### Cross-cutting note
The correctness and api-compat reviewers converged on one shared observation: the one production `SessionStats` literal outside `createSessionStats()` (telegram) and the `turn-handler.ts` `'summary'` fallback mean the optionality of `thinkingUi?:` is load-bearing in two places. Both are guarded today, so this is a "keep the optional annotation" signal, not a defect.

### Merge recommendation: **APPROVE**

The PR is clean, additive, well-tested at the unit level, follows established conventions (`/model`), and introduces no production bugs. The MEDIUM findings are test-coverage improvements that could land as follow-ups, not merge blockers — the production behavior is correct because the commander flag always seeds a valid value and the `??` resolution preserves the runtime-over-boot precedence as designed.

**Suggested post-merge follow-ups (optional):**
1. Add an integration test wiring `stats` → `runTurn` to verify the toggle actually reaches the `StreamRenderer` on the next turn.
2. Harmonize `turn-handler.ts:45` default from `'summary'` → `'live'` to close the latent divergence.
3. Optionally add a test asserting `resetStats` preserves `thinkingUi`.

**Done**
- Reviewed PR #415 across 5 dimensions (security, correctness, api-compat, test-coverage, perf/observability) via 4 parallel dimension subagents reading the diff + host source files.
- Evidence above is cited per finding with file:line references; no file edits made (read-only review, working tree left untouched).
- Verdict: **APPROVE** — no blockers; 3 optional test-follow-ups noted.
- Not pushed / no commit — this was a review-only task preserving the user's 5 dirty working-tree files.

---

PR #416, opened 2026-07-05
fix(subagent): grant main-repo read root to worktree subagents

griffinwork40:
## Problem

When a subagent is dispatched from a **worktree session** (`afk interactive -w`, whose worktrees live at `<repoRoot>/.afk-worktrees/<slug>`), its dispatcher inherits the worktree `cwd` but **no read roots** — so it defaults to `readRoots = [worktree]` (`dispatcher.ts:304`, provider init `anthropic-direct/index.ts:684`).

The containment check at `src/agent/tools/handlers/_cwd-utils.ts:106-123` is **purely lexical** (realpath compare against the roots). It has no idea the worktree is a checkout of the main repo. So when the subagent does `read_file /path/to/repo/package.json` — a **main-repo absolute path**, which appears in the system prompt's `# Environment` block, skill prompts, the parent's messages, and the model's own priors — it gets:

```
Path `/path/to/repo/package.json` is outside the allowed read roots [`…/afk-<slug>`].
```

The subagent **cannot escape this**: the path-approval hook auto-denies out-of-root access for forked children via the `parentSessionId` guard (`path-approval-hook.ts:225-237`) — sub-agents can't prompt the operator. It is locked in with no remedy.

*(Verified still valid on `main` @ `56a65c7`: `_cwd-utils.ts:106-118` lexical compare, `dispatcher.ts:303-304` `readRoots ?? [cwd]`, forked child provider gets a fresh `_sharedReadRoots = [cwd]`, hook auto-deny. Only bites `permissionMode: default` — the new-install default `bypassPermissions` disables containment.)*

## Fix

`SubagentManager.forkSubagent` now resolves the worktree's **main-repo root** and grants it as a **READ root** (never a write root — writes stay confined to the worktree for isolation) when the caller left `readRoots` unset:

```
childConfig.readRoots = [cwd, mainRoot]
```

Reading main-repo files from a worktree is not a privilege escalation — it is the same project. Resolution is best-effort via `git rev-parse --git-common-dir --show-toplevel`, and **cached per cwd** so a fan-out of subagents sharing one worktree pays a single `git` call.

Callers that pin their own `readRoots` — notably `afk farm`, which deliberately confines each branch worker to its own tree — are **left untouched** (the grant is skipped when `options.config.readRoots` is set, and no `git` even runs).

### Files
- **`src/agent/worktree-read-root.ts`** (new) — `resolveWorktreeMainRoot(cwd, execFile?)`. Returns the main root only for a *linked* worktree whose main root is distinct; `undefined` for the main worktree (root or subdir), a non-repo, empty/partial git output, or any git failure. Never throws.
- **`src/agent/subagent.ts`** — resolve + inject `readRoots` in `forkSubagent`; per-cwd cache field + `resolveMainRootForCwd`.

### Tradeoff (documented in the helper)
When the model uses a *main-repo* absolute path, the read now returns the main worktree's copy rather than the worktree's — which can differ if the worktree has uncommitted edits to that path. This is strictly better than the pre-fix hard error, and relative paths still resolve against `cwd` (the worktree). Rewriting main-repo paths to worktree equivalents was rejected as far more complex (not every main-repo path has a worktree twin — `.git`, sibling worktrees).

### Scope
Subagent-only (the reported, unfixable-by-prompt case). Top-level worktree sessions can still add the main repo via the interactive approval prompt / `/allow-dir`; broadening the parent's default roots touches several CLI entry points and the deferred-worktree `setCwd` flow, and is deliberately out of scope here.

## Tests
- `src/agent/worktree-read-root.test.ts` — 12 cases (linked worktree, main worktree relative/absolute `.git`, subdir in/out of worktree, external worktree, relative common-dir resolution, undefined/empty cwd, git failure, empty/partial output).
- `src/agent/subagent-worktree-readroot.test.ts` — 8 cases (grant from manager cwd + per-call cwd, no grant for non-worktree, caller-pinned readRoots preserved + resolver skipped, no write root, per-cwd caching, mainRoot==cwd defensive, no-cwd no-op).

## Gates
- full suite — **10766 pass / 0 fail / 14 skip** (586 files)
- `pnpm audit:sdk:check` — OK (no new SDK imports)
- `pnpm audit:env:check` — 0 violations
- `pnpm scan:env:check` — in sync

### DIFF
--- a/src/agent/subagent-worktree-readroot.test.ts
+++ b/src/agent/subagent-worktree-readroot.test.ts
@@ -1,0 +1,166 @@
+/**
+ * Regression tests: forkSubagent grants the MAIN repo root as a read root to
+ * subagents whose cwd is a linked git worktree.
+ *
+ * Bug (pre-fix): a subagent forked from an `afk -w` worktree session inherits
+ * the worktree cwd but no read roots, so its dispatcher defaults to
+ * `readRoots = [worktree]`. Any `read_file <mainRepo>/…` is rejected as
+ * "outside the allowed read roots", and the subagent cannot approve it (the
+ * path-approval hook auto-denies forked children). This locks subagents out of
+ * main-repo paths that pervade their context.
+ *
+ * Fix: forkSubagent resolves the worktree's main-repo root (best-effort, via
+ * ./worktree-read-root) and sets `childConfig.readRoots = [cwd, mainRoot]` when
+ * the caller did not pin its own read roots. These tests capture the config
+ * handed to the child AgentSession and assert the grant.
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import type { Message } from './types.js';
+
+type CapturedConfig = Record<string, unknown> | null;
+
+const shared = vi.hoisted(() => ({
+  lastConfig: null as CapturedConfig,
+}));
+
+vi.mock('./session.js', () => {
+  class MockAgentSession {
+    public readonly sessionId?: string;
+    public sendMessage: ReturnType<typeof vi.fn>;
+    public sendMessageStream: ReturnType<typeof vi.fn>;
+    public interrupt = vi.fn(async () => undefined);
+    public close = vi.fn(async () => undefined);
+    constructor(config: Record<string, unknown>) {
+      shared.lastConfig = config;
+      this.sessionId = (config.sessionId as string | undefined) ?? 'child-session-id';
+      this.sendMessage = vi.fn(async (content: string): Promise<Message> => ({
+        role: 'assistant',
+        content: `ok:${content}`,
+        timestamp: new Date(),
+      }));
+      this.sendMessageStream = vi.fn(async function* (this: MockAgentSession, content: string) {
+        const result = await this.sendMessage(content);
+        yield { type: 'message', message: result };
+        yield { type: 'done' };
+      }.bind(this));
+    }
+    get abortSignal(): AbortSignal {
+      return new AbortController().signal;
+    }
+  }
+  return { AgentSession: MockAgentSession };
+});
+
+// Control the worktree → main-root resolver without touching real git.
+vi.mock('./worktree-read-root.js', () => ({
+  resolveWorktreeMainRoot: vi.fn(async () => undefined),
+}));
+
+import { SubagentManager } from './subagent.js';
+import { resolveWorktreeMainRoot } from './worktree-read-root.js';
+
+const WORKTREE = '/repo/.afk-worktrees/wt';
+const MAIN = '/repo';
+
+const mockedResolve = vi.mocked(resolveWorktreeMainRoot);
+
+function forkOpts(
+  config: Record<string, unknown>,
+): Parameters<SubagentManager['forkSubagent']>[0] {
+  return {
+    parent: { sessionId: 'parent' },
+    config,
+  } as Parameters<SubagentManager['forkSubagent']>[0];
+}
+
+describe('forkSubagent — worktree main-repo read-root grant', () => {
+  beforeEach(() => {
+    shared.lastConfig = null;
+    mockedResolve.mockReset();
+    mockedResolve.mockResolvedValue(undefined);
+  });
+
+  it('grants [cwd, mainRoot] when the manager cwd is a linked worktree', async () => {
+    mockedResolve.mockResolvedValue(MAIN);
+    const mgr = new SubagentManager({ cwd: WORKTREE });
+    await mgr.forkSubagent(forkOpts({ model: 'sonnet', apiKey: 'k' }));
+
+    const cfg = shared.lastConfig as { cwd?: string; readRoots?: string[] } | null;
+    expect(cfg?.cwd).toBe(WORKTREE);
+    expect(cfg?.readRoots).toEqual([WORKTREE, MAIN]);
+    expect(mockedResolve).toHaveBeenCalledWith(WORKTREE);
+  });
+
+  it('grants [cwd, mainRoot] when cwd comes from per-call config (agent tool cwd)', async () => {
+    mockedResolve.mockResolvedValue(MAIN);
+    const mgr = new SubagentManager(); // no manager cwd
+    await mgr.forkSubagent(forkOpts({ model: 'sonnet', apiKey: 'k', cwd: WORKTREE }));
+
+    const cfg = shared.lastConfig as { cwd?: string; readRoots?: string[] } | null;
+    expect(cfg?.cwd).toBe(WORKTREE);
+    expect(cfg?.readRoots).toEqual([WORKTREE, MAIN]);
+    expect(mockedResolve).toHaveBeenCalledWith(WORKTREE);
+  });
+
+  it('does NOT set readRoots when cwd is not a worktree (resolver returns undefined)', async () => {
+    mockedResolve.mockResolvedValue(undefined);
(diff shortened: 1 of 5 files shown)


griffinwork40:
Grounded at the ref: containment uses realpath (symlink-safe) but has **no `.git` exclusion** — so granting `mainRoot` as a read root does include `<mainRoot>/.git/`. Agent 1's security finding is confirmed as a real (low) widening.

I now have everything. Verification is complete — I performed Wave 1.5 inline because the `research-agent` type in this environment lacks bash/git (Agent 2 hit that wall), which are exactly the tools citation/absence verification needs. Emitting the synthesis and merge decision.

---

## Code Review — PR #416: `fix(subagent): grant main-repo read root to worktree subagents`

**Target:** PR #416 (griffinwork40/agent-afk) · branch `afk/verify-fix-worktree-readroots` → `main` · OPEN/MERGEABLE
**Reviewed ref:** `d7e1c6bacaee59223c77cd3e0bb6ff16ff1db5c2` · **Regime:** full (+445/−5, 5 files: 2 production, 3 test)
**Stated intent:** available (PR body) → **spec-compliance assessed**

### Wave 1.5 verification manifest (performed inline with git/sha256; empirical git test in throwaway repo)

| Wave-1 claim | Status | Evidence |
|---|---|---|
| `bundled.test.ts` **fails** at ref (blocking) | **REFUTED / dropped** | All 5 skill files at `d7e1c6b` sha256-match the new pinned hashes → snapshot **passes** |
| Hash bumps are **scope creep** | **REFUTED / dropped** | `main..d7e1c6b` diff for `bundled.test.ts` is **empty** — main already pins identical hashes; it's merge-base skew, not a change this PR introduces |
| Bump comments not updated (low·high) | **MOOT / dropped** | State is byte-identical to main HEAD; not introduced here |
| `--git-common-dir` main-subdir contract violation (correctness) | **REFUTED / dropped** | git 2.53 emits `../.git` (relative-to-cwd) from a subdir, not `.git` → resolver correctly returns `undefined` |
| Value-cache concurrency race | **CONFIRMED**, downgraded to low | `resolveMainRootForCwd` awaits before `set` — concurrent same-cwd forks each spawn git |
| `.git/` included in read grant (security) | **CONFIRMED**, low | `_cwd-utils.ts` containment has no `.git` exclusion (verified at ref) |

No surviving **critical/high** findings → `/shadow-verify` not required (medium-and-below pass straight through).

### Findings (sorted; none blocking)

**`low · high-conf · security · worktree-read-root.ts:84-92 · ref:d7e1c6b · file-state`** — The grant is `mainRoot = dirname(git-common-dir)` = the repo root, which contains `.git/`; containment (`_cwd-utils.ts` realpath + lexical compare) has no `.git` exclusion, so a worktree subagent can now read `<mainRoot>/.git/config` (and `.git/credentials` if `credential.helper=store --file` is configured). *Evidence:* `const mainRoot = path.resolve(path.dirname(absCommonDir)); … return mainRoot;`. *Mitigating:* trusted same-session caller; brings worktree subagents to **parity** with normal (non-worktree) subagents that already read `.git/`; secrets-in-`.git` requires uncommon config. *Suggestion (optional hardening):* exclude `.git/` from the granted read root, or grant only the tracked working tree.

**`low · med-conf · perf · subagent.ts:210-217 · ref:d7e1c6b · diff-context`** — `resolveMainRootForCwd` caches the resolved **value**, not the in-flight promise, so concurrent same-cwd forks (real path: parallel `agent`/DAG fan-out) each spawn `git rev-parse`; the intent's "a fan-out sharing one worktree pays a single git call" holds only for **sequential** forks. Result is deterministic (no correctness bug). *Evidence:* `if (…cache.has(cwd)) return …; const mainRoot = await resolveWorktreeMainRoot(cwd); this.…cache.set(cwd, mainRoot);`. *Suggestion:* cache `Promise<string|undefined>` so concurrent forks await one call.

**`low · med-conf · observability · worktree-read-root.ts:493 · ref:d7e1c6b · diff-context`** — The best-effort `catch {}` swallows every failure (non-repo, git missing, malformed output) with no log/trace, so a subagent that silently loses its main-repo read root has no debug signal. *Evidence:* `} catch { … return undefined; }`. *Suggestion:* emit one `debugLog`/trace event on the failure branch.

**`low · high-conf · test-coverage · subagent-worktree-readroot.test.ts:147-153 · ref:d7e1c6b · diff-context`** — The "caches resolution per cwd" test awaits two forks **sequentially**, so it does not exercise the concurrent path where the cache is actually load-bearing (and, per the finding above, does not fully dedup). *Evidence:* `await mgr.forkSubagent(...); await mgr.forkSubagent(...); expect(mockedResolve).toHaveBeenCalledTimes(1);`. *Suggestion:* add a `Promise.all([...])` case.

**`nit · high-conf · test-coverage · worktree-read-root.test.ts:320-326 · ref:d7e1c6b · diff-context`** — The "subdir of main worktree" case mocks an **absolute** `${MAIN}/.git`, but real git emits **relative** `../.git` from a subdir; the resolver returns the correct `undefined` for both, so no bug is masked — pure fidelity nit. *Suggestion:* add a `commonDirRaw='../.git'` case for realism.

### What was checked / not checked
- **Refs:** Wave 1 read the diff (diff-context) + files at `d7e1c6b` via `git show`; all blocking-tier claims re-derived at the ref. Skill-file hashes verified by sha256 at `d7e1c6b`; git output verified empirically (git 2.53.0, throwaway repo).
- **Stated intent:** PR #416 title+body — spec-compliance assessed. Intent is **met**: read-only grant (never `writeRoots`; write isolation preserved — confirmed), subagent-only scope, `afk farm` (pins `readRoots`) untouched, best-effort git resolution. No unmet clauses; no scope creep (`bundled.test.ts` bump is main-parity).
- **Not checked:** did not run the test suite (read-only review) — PR body claims 10766 pass/0 fail, and I independently confirmed the one at-risk gate (`bundled.test.ts`) passes; did not review whether the branch being behind main (main has since removed `thinking.ts`, refactored `nesting.ts`/`trace/writer.ts`) creates any *semantic* (non-textual) integration issue — GitHub reports MERGEABLE (no textual conflict).

### Merge decision: **MERGE**

The change is correct (all four worktree/subdir cases resolve properly, verified empirically), security-sound (write isolation intact, realpath containment, no shell/arg injection — `cwd` is parent-derived), and spec-compliant. The one blocking-looking finding (test failure) was a tool-less mis-deduction and is refuted. The five surviving findings are all **low/nit optional follow-ups** — the strongest being the `.git/`-inclusion note (parity, gated) and the promise-cache concurrency nicety. None block merge.

**Done**
- Reviewed PR #416 at ref `d7e1c6b` via the full-regime pipeline (2 parallel review agents → inline Wave 1.5 verification → synthesis); read-only, no repo/PR/working-tree mutation.
- Evidence: diff artifact `…/skill-preflight/92be731d-…/pr-416.diff`; sha256 verification of 5 skill files vs pins at the ref; empirical `git rev-parse --git-common-dir` test (git 2.53.0); containment code read at `git show d7e1c6b:src/agent/tools/handlers/_cwd-utils.ts`.
- World change: none (analysis only). Working tree left dirty-as-found (3 uncommitted changes untouched).
- Verdict: **MERGE** — 0 blocking, 4 low + 1 nit optional follow-ups. Adaptation noted: Wave 1.5 ran inline because the `research-agent` type here lacks bash/git; one Wave-1 agent's tool-less claims were caught and refuted by that verification.

---

PR #445, opened 2026-07-05
feat(interactive): persist thinking-ui default via AFK_THINKING_UI + interactive.thinkingUi

griffinwork40:
## What

Make the interactive REPL's thinking-display mode (`/thinking` → `live | summary | digest | off`) **persistable across sessions**. Previously it was session-only: it seeded solely from the `--thinking-ui` CLI flag (hardcoded default `live`) with no env or config layer, so the only way to change the default was to re-pass `--thinking-ui <mode>` on every launch.

## Why

A user asked whether the `/thinking` choice persists — it doesn't. The codebase already had the exact precedence pattern for this (`worktree-autoname` / `worktree-base`: flag → env → config → default), but `--thinking-ui` never got wired that way. This closes that gap additively.

## Precedence

```
--thinking-ui flag  >  AFK_THINKING_UI env  >  interactive.thinkingUi config  >  'live'
```

`/thinking` still overrides live within a session. This is **display-only** — extended thinking always runs; cost and latency are unaffected.

## Changes

- **`src/config/env.ts`** — register `AFK_THINKING_UI` (registry entry + typed getter); `docs/env-registry.{json,md}` regenerated via `pnpm scan:env`.
- **`src/cli/config.ts`** — add `interactive.thinkingUi` to the config type + JSON loader. Invalid values are silently ignored (not fatal), matching the display-only, non-security nature of the setting.
- **`src/config/settable-keys.ts`** — register `interactive.thinkingUi` (agent tier, enum) so `afk config set interactive.thinkingUi digest` works.
- **`src/cli/commands/interactive.ts`** — add exported `resolveThinkingUi()` resolver (mirrors `isAutonameEnabled`); drop the static Commander default; resolve precedence in the action handler right after `loadConfig()`.
- **`src/cli/commands/interactive/shared.ts`** — `CliOptions.thinkingUi` is now optional (resolved post-parse rather than defaulted by Commander).
- **`src/cli/commands/interactive/bootstrap.ts`** — updated seeding comment + `undefined` guard.

## Usage after this change

```bash
# one-off (unchanged)
afk i --thinking-ui digest

# persistent, via env
export AFK_THINKING_UI=digest

# persistent, via config
afk config set interactive.thinkingUi digest
```

## Testing

- `resolveThinkingUi` precedence matrix (flag wins, env-over-config, case/whitespace-insensitive env, invalid-env fall-through, config fallback, default) — `src/cli/commands/interactive.test.ts`.
- Config-loader accept-valid / ignore-invalid / leave-undefined cases — `src/cli/config.test.ts`.
- `pnpm lint` clean; `pnpm audit:sdk:check`, `pnpm audit:env:check`, `pnpm scan:env:check` all pass.
- Full suite: 10,891 passed / 14 skipped. The only failing suites are 4 `chat.*.test.ts` files failing to load on a pre-existing `builtinToolSchemas` mock-drift error — reproduced identically on `origin/main` with none of these changes present (unrelated to this PR).

## Notes

- Scope is the interactive REPL only (where `/thinking` and `--thinking-ui` live); one-shot `chat` does not use this knob.

### DIFF
--- a/docs/env-registry.json
+++ b/docs/env-registry.json
@@ -645,6 +645,15 @@
       "category": "model"
     },
     {
+      "name": "AFK_THINKING_UI",
+      "description": "Default thinking-display mode for the interactive REPL: summary | live | digest | off. Display-only — controls how extended-thinking blocks render, never whether thinking runs (cost/latency unaffected). Overridden per-launch by --thinking-ui and mutable mid-session via /thinking. Precedence: --thinking-ui flag > this env > interactive.thinkingUi config > live. Invalid values are ignored.",
+      "type": "string",
+      "required": false,
+      "default": "live",
+      "example": "digest",
+      "category": "misc"
+    },
+    {
       "name": "AFK_TIMEOUT_MS",
       "description": "Per-turn timeout in milliseconds. Provider/SDK default if unset.",
       "type": "number",
--- a/docs/env-registry.md
+++ b/docs/env-registry.md
@@ -2,7 +2,7 @@
 
 Generated from `src/config/env.ts`. Do not edit by hand — run `pnpm scan:env` after changing the registry source.
 
-**118 vars** across 12 categories. Every `process.env[...]` read in `src/` outside `src/config/env.ts` is a CI failure (enforced by `pnpm audit:env:check`).
+**119 vars** across 12 categories. Every `process.env[...]` read in `src/` outside `src/config/env.ts` is a CI failure (enforced by `pnpm audit:env:check`).
 
 To add a var: edit `src/config/env.ts` (add a getter on `env` + an entry in `ENV_REGISTRY`), then run `pnpm scan:env`.
 
@@ -180,6 +180,7 @@ To add a var: edit `src/config/env.ts` (add a getter on `env` + an entry in `ENV
 | `AFK_SHELL_PASSTHROUGH` | boolean |  | `1` | `0` | Enable the interactive REPL `!cmd` / `!&cmd` shell-passthrough feature. On by default. Set to 0, false, off, or no (case-insensitive) to disable, so inputs beginning with ! are sent to the model as literal text instead of being executed as shell commands. Equivalent to the --no-shell-passthrough flag. |
 | `AFK_SHOW_DIFFS` | boolean |  |  |  | Show inline diffs in the tool-lane output for edit/write tool calls. 1 = on, 0 = off. |
 | `AFK_SPINNER_TIPS` | boolean |  |  |  | Show rotating tips in the loading spinner during long calls. 1 = on, 0 = off. |
+| `AFK_THINKING_UI` | string |  | `live` | `digest` | Default thinking-display mode for the interactive REPL: summary \| live \| digest \| off. Display-only — controls how extended-thinking blocks render, never whether thinking runs (cost/latency unaffected). Overridden per-launch by --thinking-ui and mutable mid-session via /thinking. Precedence: --thinking-ui flag > this env > interactive.thinkingUi config > live. Invalid values are ignored. |
 | `AFK_USER_CARD_MAX_ROWS` | number |  |  | `24` | Maximum number of visual rows emitted by renderUserCard before collapsing the remainder into a dim "…(N lines collapsed)" summary row. Defaults to 24. Non-integer or non-positive values are silently ignored and the default applies. |
 | `AFK_WRITE_DENYLIST` | string |  |  | `**/.env,**/secrets/**` | Comma-separated list of additional path globs that the write_file tool refuses to write to. |
 | `AFK_WRITE_DIFF` | boolean |  |  |  | Show a diff preview before each write_file tool call. Defaults provider-controlled when unset. |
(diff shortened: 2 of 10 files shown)


griffinwork40:
Review complete. I've verified every dimension by reading the actual code and reconciling against the diff. No blocking issues found.

---

## Review: PR #445 — persist thinking-UI default via `AFK_THINKING_UI` + `interactive.thinkingUi`

**Verdict: ✅ APPROVE** — small, focused, correct, well-tested, and convention-compliant. Only optional nits below.

**What it does:** Adds a persistent default for the REPL's thinking-*display* mode with precedence `--thinking-ui` flag > `AFK_THINKING_UI` env > `interactive.thinkingUi` config > `'live'`. Display-only; never affects whether thinking runs. Mirrors the existing `isAutonameEnabled` / `suggestGhost` patterns.

### Findings by dimension

**Correctness — SOUND**
- `resolveThinkingUi` precedence is correct; env is trim/lowercase/allowlist-validated and falls through (not fatal) on invalid input; config is allowlist-validated at load time. Defense-in-depth across all three sources.
- The `bootstrap.ts` change from unconditional `stats.thinkingUi = options.thinkingUi` to a guarded `if (options.thinkingUi !== undefined)` is **behaviorally necessary**, not just stylistic: `createSessionStats()` seeds `stats.thinkingUi = 'live'` (`session-stats.ts:26`), and without the guard an unresolved `undefined` would clobber that seed, then flow through `loop-iteration.ts:583` (`stats.thinkingUi ?? options.thinkingUi`) → `undefined` → `runTurn`'s `= 'summary'` default (`turn-handler.ts:45`), silently rendering `summary` instead of `live`. The guard preserves the `'live'` backstop. Correct.
- Verified the wiring: the action handler resolves and assigns `options.thinkingUi` (diff ~L296, in the `loadConfig()` block) **before** passing the same `options` object to `bootstrapSession` at `interactive.ts:351`. Same reference — resolution propagates.

**API-compat — SOUND**
- `CliOptions.thinkingUi` widened required→optional. The only two production consumers both handle `undefined`: `bootstrap.ts` (now guarded) and `loop-iteration.ts:583` (`??` fallback, absorbed by `runTurn`'s defaulted param). Single production caller of `bootstrapSession`.
- Removing the Commander static `'live'` default is behavior-preserving because `resolveThinkingUi`'s terminal fallback is `'live'`. `--help` no longer auto-appends `(default: live)`, but the option description now documents the default + env + config sources. No `--help` snapshot test references `--thinking-ui` (only the unrelated `/thinking-ui` slash alias), so nothing breaks in CI.

**Test coverage — GOOD (one minor gap)**
- `resolveThinkingUi`: comprehensive — default, flag-wins-over-all, env-over-config, case/whitespace-insensitivity, invalid-env fallthrough, config-used. `config.ts` JSON integration: valid/invalid/absent.
- Gap (non-blocking): no test exercises the `bootstrap.ts` conditional-seeding behavior (undefined → stats stays `'live'`) or the action-handler assignment. Both are simple integration points; low risk.

**Security — CLEAN.** Display-only allowlisted string; no injection/eval/path/credential surface. Env read via the sanctioned `env` object (compliant with the sole-read-point rule).

**Perf/observability — CLEAN.** One-time resolution at launch; no hot-path impact.

**Docs/conventions — COMPLIANT.** `env-registry.{json,md}` regenerated (118→119); `env.AFK_THINKING_UI` getter placed alphabetically; `ENV_REGISTRY` entry + `settable-keys.ts` entry (enum, tier `agent`) present — completes the `config_set/get` surface. Precedence is documented consistently in four places (env desc, `CliConfig` JSDoc, `resolveThinkingUi` JSDoc, `--help` text).

### Optional nits (non-blocking)
1. **`runTurn` defaults to `'summary'`** (`turn-handler.ts:45`) while every other layer defaults to `'live'`. Pre-existing and unreachable in production (the `createSessionStats` seed prevents it), but a reader could be misled. Out of this PR's scope — consider aligning to `'live'` in a follow-up.
2. **Normalization asymmetry:** env values are trimmed/lowercased, config values are exact-match only. A JSON `"thinkingUi": "Live"` is silently ignored → falls to default. Documented as "invalid values are ignored," so acceptable; a one-line note or symmetric normalization would be friendlier.
3. Consider adding one integration test for the bootstrap seeding guard to lock in the `'live'`-preservation behavior described above.

### Terminal-state note
This was a **read-only** review — I did not edit, commit, push, or comment on the PR, and I did not run the suite (this checkout is on `main`, not `afk/thinking-ui-default`, so a local run would test the wrong tree). The authoritative gates are the PR's CI: `pnpm lint`, `pnpm test`, and especially `pnpm scan:env:check` / `pnpm audit:env:check` — all expected green since the registry `.json`/`.md` were regenerated in-diff and the getter follows the required shape.

**Done**
- Reviewed PR #445 across correctness, api-compat, test-coverage, security, perf, and docs/conventions by reading source (`interactive.ts`, `bootstrap.ts`, `loop-iteration.ts`, `turn-handler.ts`, `session-stats.ts`, `slash/types.ts`, `config.ts`, `env.ts`, `settable-keys.ts`, `tsconfig.json`) and reconciling against the diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/61082109-d7b0-486c-80a9-94ec758bd408/pr-445.diff`.
- Verdict: **APPROVE**; no blocking issues, three optional nits.
- No world state changed — working tree untouched (per the read-only preflight contract), nothing committed, pushed, or posted to the PR.
- Not done: local test/lint execution (wrong branch checked out) — deferred to the PR's CI.

---

PR #447, opened 2026-07-05
refactor(repl): calm REPL chrome (glyph mode marker, idle rail, cwd/branch dedupe)

griffinwork40:
## Problem

Three stacked lines of chrome around the REPL input read busy and redundant:

- the `⚡ bypass` chip renders **twice** (prompt line + status line), bold pink both times
- cwd and git branch are near-duplicate strings for afk worktrees (`afk-20260705-…` dir vs `afk/20260705-…` branch) — ~80 columns encoding one fact
- the 5-cell loop rail paints **permanently**, even when no turn is running — persistent chrome for transient activity

## Before / after (idle)

```
before:
afk ⚡ bypass  ›
  ◇ observe · ◇ model · ◇ choose · ◆ act · ◇ update
~/…/afk-20260705-142358-47b3ec · ⎇ afk/20260705-142358-47b3ec · opus_1m · ⚡ bypass

after:
afk ⚡  ›
  · idle
⎇ afk/20260705-142358-47b3ec · opus_1m · ⚡ bypass
```

During an active turn the full 5-cell rail renders exactly as before.

## Changes

- **Prompt** (`repl-loop-shared.ts`): mode markers become glyph-only — ` ●` plan, ` ◐` AFK, ` ⚡` bypass, nothing for default. The worded chip lives on the status line only.
- **Loop rail** (`loop-stage.ts`): `formatStageRail` collapses the idle/reset `observing` stage to a single dim `· idle` cell. Content-only change — no row-reservation, paint-mechanics, or state-machine changes; `extraRows` geometry untouched.
- **Status line** (`status-line.ts`): when `branch.replaceAll('/','-') === basename(cwd)` (the afk-worktree naming scheme), the cwd part is omitted and the `⎇ branch` segment renders alone, **promoted to never-drop** as the sole surviving location signal. The compare runs uncached on every repaint (raw cwd, not the display-shortened form) so it cannot go stale against the async git sampler.
- **Palette mute** (`status-line.ts` only): branch text teal→dim; `palette.fileRef` untouched for its other consumers. The bypass chip is now the single saturated accent.

## Safety rationale

The status row is painted in a row carved **outside the scroll region** (`writeScrollRegion`) — it never enters scrollback or piped logs. The prompt is therefore the only permission-mode signal that survives into the linear transcript, which matters for an away-from-keyboard operator auditing after the fact — and `bypassPermissions` is the new-install default. Hence glyph-only at the caret rather than full removal. The deduped location segment is promoted to never-drop so narrow terminals cannot shed the only "where am I" indicator.

## Verification

- `pnpm lint` (strict tsc) green
- Full `pnpm test`: **10901 passed | 14 skipped, 0 failing tests**. The 4 `chat.*.test.ts` file-level collection errors (`builtinToolSchemas` mock) fail identically on clean `main` at the base commit — pre-existing, unrelated to this diff.
- Tests updated: `loop-stage.test.ts` (idle collapse + full-rail-per-stage + exactly-one-active-diamond assertions), `terminal-compositor.footer-ghost.test.ts` (drives the bar to a mid-turn stage so the rail-finder probes the full rail; the self-heal `redraw()` re-asserts stored `currentStage`)
- Tests added: `repl-loop-shared.test.ts` (buildPrompt glyphs, all 4 modes), `status-line.test.ts` (dedupe: merged segment, never-drop survival at narrow width, `#PR` suffix, non-matching pair unchanged)
- Compositor geometry tests pass untouched (only footer-ghost's content probe needed the stage drive; no geometry assertions changed)

## Design provenance

Options were researched against Claude Code / Codex CLI / Gemini CLI / opencode / aider / Crush chrome conventions, then adversarially reviewed (3-lens devils-advocate critique); this implements the revised **A-prime hybrid** — the original "move the mode chip fully to the status line" variant was rejected for erasing the scrollback-persistent safety signal.

### DIFF
--- a/src/cli/commands/interactive/loop-stage.test.ts
+++ b/src/cli/commands/interactive/loop-stage.test.ts
@@ -186,10 +186,12 @@ describe('formatStageRail', () => {
     bold: (s: string) => `bold<${s}>`,
   };
 
-  it('renders all five stages joined by separators', () => {
+  it('collapses "observing" (idle/reset) to a single dim "· idle" cell', () => {
     const out = formatStageRail('observing', fmt);
+    expect(out).toBe('dim<· idle>');
+    // The idle collapse must NOT leak any of the five stage labels.
     for (const stage of LOOP_STAGES) {
-      expect(out).toContain(STAGE_LABEL[stage]);
+      expect(out).not.toContain(STAGE_LABEL[stage]);
     }
   });
 
@@ -199,6 +201,21 @@ describe('formatStageRail', () => {
     // Inactives use the hollow diamond + dim.
     expect(out).toContain('dim<◇ observe>');
     expect(out).toContain('dim<◇ update>');
+  });
+
+  it('renders the full 5-cell rail (with ◆ on the active cell) for every non-observing stage', () => {
+    for (const stage of LOOP_STAGES.filter((s) => s !== 'observing')) {
+      const out = formatStageRail(stage, fmt);
+      // All five stage labels present — the rail is not collapsed.
+      for (const label of Object.values(STAGE_LABEL)) {
+        expect(out).toContain(label);
+      }
+      // Exactly the active stage's cell carries the solid diamond.
+      expect(out).toContain(`◆ ${STAGE_LABEL[stage]}`);
+      for (const other of LOOP_STAGES.filter((s) => s !== stage)) {
+        expect(out).not.toContain(`◆ ${STAGE_LABEL[other]}`);
+      }
+    }
   });
 });
 
@@ -266,9 +283,10 @@ describe('LoopStageBar', () => {
     const out = joinWrites(stream);
     // rows=24, extraRows=1 → paint at row 23 (immediately above the status row).
     expect(cupRows(out)).toContain(23);
-    // Idle rail shows every stage label, with 'observe' active.
-    expect(out).toContain('observe');
-    expect(out).toContain('update');
+    // Idle paint collapses to the single dim `· idle` cell (formatStageRail
+    // special-cases the between-turns 'observing' stage) — no stage labels.
+    expect(out).toContain('· idle');
+    expect(out).not.toContain('observe');
     bar.stop();
   });
 
--- a/src/cli/commands/interactive/loop-stage.ts
+++ b/src/cli/commands/interactive/loop-stage.ts
@@ -156,11 +156,21 @@ export function advanceStage(s: StageTrackerState, event: OutputEvent): boolean 
  *
  * Color is intentionally restrained — bold + accent on the active stage,
  * dim on inactives. The goal is glance-readable, not decorative.
+ *
+ * Special case: `observing` is the idle/reset stage between turns (a turn
+ * just armed with no events yet, or no turn in flight at all) — the rail has
+ * nothing earned to show, so painting all five cells persistently between
+ * turns is chrome, not signal. Collapse to a single dim `· idle` cell
+ * instead of the full 5-cell rail. Every other stage renders the full rail
+ * unchanged.
  */
 export function formatStageRail(
   active: LoopStage,
   fmt: { dim: (s: string) => string; accent: (s: string) => string; bold: (s: string) => string },
 ): string {
+  if (active === 'observing') {
+    return fmt.dim('· idle');
+  }
   const cells = LOOP_STAGES.map((stage) => {
     const isActive = stage === active;
     const glyph = isActive ? '◆' : '◇';
@@ -205,7 +215,9 @@ export function formatStageRail(
  * `BackgroundStatusBar.repaint()`. It renders exactly one row.
  *
  * Visibility: renders whenever `started === true` and the stream is a TTY.
- * Between turns (stage = `'observing'`) it shows a dim "idle" rail.
+ * Between turns (stage = `'observing'`) `formatStageRail` collapses the full
+ * 5-cell rail to a single dim `· idle` cell (see its docstring) rather than
+ * painting all five stages with none of them earned yet.
  */
 export class LoopStageBar {
   private readonly stream: NodeJS.WriteStream;
--- a/src/cli/commands/interactive/repl-loop-shared.test.ts
+++ b/src/cli/commands/interactive/repl-loop-shared.test.ts
@@ -1,0 +1,49 @@
+/**
+ * Tests for buildPrompt (repl-loop-shared.ts).
+ *
+ * The caret carries the brand plus a GLYPH-ONLY echo of the non-default
+ * permission mode (` ●` / ` ◐` / ` ⚡`) — never the worded chip, which lives
+ * on the status line. The glyph must stay at the caret because the status
+ * row is painted outside the scroll region and never enters scrollback: the
+ * prompt is the only mode signal that survives into the linear transcript.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { buildPrompt } from './repl-loop-shared.js';
+
+const BROAD_ANSI_RE = /\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])/g;
+
+function strip(s: string): string {
+  return s.replace(BROAD_ANSI_RE, '');
+}
+
+describe('buildPrompt', () => {
+  it('renders bare brand + caret with no marker in default mode', () => {
+    const out = strip(buildPrompt('default'));
+    expect(out).toBe('afk  › ');
+  });
+
+  it('renders a glyph-only plan marker (no word)', () => {
+    const out = strip(buildPrompt('plan'));
+    expect(out).toContain('●');
+    expect(out).not.toContain('plan');
+  });
+
+  it('renders a glyph-only autonomous marker (no word)', () => {
+    const out = strip(buildPrompt('autonomous'));
+    expect(out).toContain('◐');
+    expect(out).not.toContain('AFK');
+  });
+
+  it('renders a glyph-only bypass marker (no word)', () => {
+    const out = strip(buildPrompt('bypassPermissions'));
+    expect(out).toContain('⚡');
+    expect(out).not.toContain('bypass');
+  });
+
+  it('always ends with the caret', () => {
+    for (const mode of ['default', 'plan', 'autonomous', 'bypassPermissions'] as const) {
+      expect(strip(buildPrompt(mode))).toMatch(/› $/);
+    }
+  });
+});
--- a/src/cli/commands/interactive/repl-loop-shared.test.ts
+++ b/src/cli/commands/interactive/repl-loop-shared.test.ts
@@ -1,11 +1,13 @@
 /**
  * Tests for buildPrompt (repl-loop-shared.ts).
  *
- * The caret carries the brand plus a GLYPH-ONLY echo of the non-default
- * permission mode (` ●` / ` ◐` / ` ⚡`) — never the worded chip, which lives
- * on the status line. The glyph must stay at the caret because the status
+ * The caret carries the brand plus a compact echo of the non-default
+ * permission mode (` ●` / ` ◐` / ` ⚡bp`) — never the worded chip, which lives
+ * on the status line. The echo must stay at the caret because the status
  * row is painted outside the scroll region and never enters scrollback: the
  * prompt is the only mode signal that survives into the linear transcript.
+ * Plan and AFK are glyph-only; bypass ALSO keeps a short ASCII tag (`bp`)
+ * because it is security-sensitive and must stay grep-able in piped logs.
  */
 
 import { describe, it, expect } from 'vitest';
@@ -35,9 +37,13 @@ describe('buildPrompt', () => {
     expect(out).not.toContain('AFK');
   });
 
-  it('renders a glyph-only bypass marker (no word)', () => {
+  it('renders the bypass marker with a grep-able ASCII token (glyph + bp)', () => {
     const out = strip(buildPrompt('bypassPermissions'));
     expect(out).toContain('⚡');
+    // Bypass is security-sensitive: it keeps an ASCII tag so a post-hoc grep
+    // of a piped transcript can still locate permission-off windows.
+    expect(out).toContain('bp');
+    // …but still not the full worded chip (that lives on the status line).
     expect(out).not.toContain('bypass');
   });
 
--- a/src/cli/commands/interactive/repl-loop-shared.ts
+++ b/src/cli/commands/interactive/repl-loop-shared.ts
@@ -57,15 +57,20 @@ export interface TurnState {
 }
 
 export function buildPrompt(mode: PermissionMode): string {
-  // The model name lives only in the persistent status line (status-line.ts,
-  // a never-drop field) — the caret carries just the brand + the non-default
-  // permission-mode chip (the gate-check worth confirming at the cursor), so
-  // model + mode aren't redundantly stacked two lines apart.
+  // The model name AND the worded mode chip (`○ default`, `● plan`, `◐ AFK`,
+  // `⚡ bypass`) live only in the persistent status line (status-line.ts) —
+  // the caret carries just the brand + a glyph-only echo of the non-default
+  // permission mode. The glyph stays here (rather than moving entirely to
+  // the status line) because the status row is carved out of the scroll
+  // region (see status-line.ts's writeScrollRegion) and never enters
+  // scrollback or piped logs: the prompt is the ONLY mode signal that
+  // survives into the linear transcript. Default mode adds no glyph — its
+  // absence is itself the "contained" signal, mirrored at the caret.
   const base = palette.brand('afk');
   const marker =
-    mode === 'plan' ? palette.warning(' ● plan') :
-    mode === 'autonomous' ? palette.info(' ◐ AFK') :
-    mode === 'bypassPermissions' ? palette.bypass(' ⚡ bypass') :
+    mode === 'plan' ? palette.warning(' ●') :
+    mode === 'autonomous' ? palette.info(' ◐') :
+    mode === 'bypassPermissions' ? palette.bypass(' ⚡') :
     '';
   return base + marker + palette.dim('  › ');
 }
--- a/src/cli/commands/interactive/repl-loop-shared.ts
+++ b/src/cli/commands/interactive/repl-loop-shared.ts
@@ -59,18 +59,22 @@ export interface TurnState {
 export function buildPrompt(mode: PermissionMode): string {
   // The model name AND the worded mode chip (`○ default`, `● plan`, `◐ AFK`,
   // `⚡ bypass`) live only in the persistent status line (status-line.ts) —
-  // the caret carries just the brand + a glyph-only echo of the non-default
-  // permission mode. The glyph stays here (rather than moving entirely to
+  // the caret carries just the brand + a compact echo of the non-default
+  // permission mode. That echo stays here (rather than moving entirely to
   // the status line) because the status row is carved out of the scroll
   // region (see status-line.ts's writeScrollRegion) and never enters
   // scrollback or piped logs: the prompt is the ONLY mode signal that
-  // survives into the linear transcript. Default mode adds no glyph — its
-  // absence is itself the "contained" signal, mirrored at the caret.
+  // survives into the linear transcript. Default mode adds no marker — its
+  // absence is itself the "contained" signal, mirrored at the caret. Plan and
+  // AFK stay glyph-only, but bypass ALSO keeps a short ASCII tag (`bp`): it is
+  // the security-sensitive mode, and an ASCII token is what lets a post-hoc
+  // `grep` of a piped transcript locate the windows where permissions were off
+  // (a bare glyph is not reliably searchable).
   const base = palette.brand('afk');
   const marker =
     mode === 'plan' ? palette.warning(' ●') :
     mode === 'autonomous' ? palette.info(' ◐') :
-    mode === 'bypassPermissions' ? palette.bypass(' ⚡') :
+    mode === 'bypassPermissions' ? palette.bypass(' ⚡bp') :
     '';
   return base + marker + palette.dim('  › ');
 }
--- a/src/cli/commands/interactive/repl-loop-shared.ts
+++ b/src/cli/commands/interactive/repl-loop-shared.ts
@@ -69,7 +69,9 @@ export function buildPrompt(mode: PermissionMode): string {
   // AFK stay glyph-only, but bypass ALSO keeps a short ASCII tag (`bp`): it is
   // the security-sensitive mode, and an ASCII token is what lets a post-hoc
   // `grep` of a piped transcript locate the windows where permissions were off
-  // (a bare glyph is not reliably searchable).
+  // (a bare glyph is not reliably searchable). A post-hoc transcript scan
+  // should grep the COMPOUND token `⚡bp` (not a bare `bp`), because bare `bp`
+  // also matches unrelated substrings like "subprocess".
   const base = palette.brand('afk');
   const marker =
     mode === 'plan' ? palette.warning(' ●') :
--- a/src/cli/status-line.test.ts
+++ b/src/cli/status-line.test.ts
@@ -1019,4 +1019,50 @@ describe('StatusLine cwd/branch worktree dedupe', () => {
     expect(out).toContain('⎇ feat/x');
     status.stop();
   });
+
+  it('does NOT merge on a name coincidence outside .afk-worktrees (positive worktree signal required)', () => {
+    // cwd basename `redesign` equals the branch, but the parent dir is `/tmp`,
+    // not `.afk-worktrees` — this is a plain checkout, not a managed worktree.
+    // The cwd ("where am I?") must NOT be suppressed on a name coincidence
+    // alone; a positive worktree signal (parent dir) is required to dedupe.
+    stream.columns = 100;
+    const status = new StatusLine({ stream: stream as unknown as NodeJS.WriteStream, throttleMs: 0 });
+    status.start();
+    stream.writes.length = 0;
+    status.repaint({ model: 'sonnet', cwd: '/tmp/redesign', branch: 'redesign' });
+    const out = lastJoined(stream).replace(BROAD_ANSI_RE, '');
+    // Both the cwd and the branch segment render, so the coincident identity
+    // appears twice — it was NOT collapsed into one merged segment (contrast
+    // the afk-worktree dedupe cases, where the shared slug renders exactly once).
+    expect(out).toContain('⎇ redesign');
+    expect(out.indexOf('redesign')).not.toBe(out.lastIndexOf('redesign'));
+    status.stop();
+  });
+
+  it('does NOT merge when the shared identity exceeds the 30-col cap — keeps cwd as a second signal', () => {
+    // A genuine managed-worktree path (parent IS .afk-worktrees) whose identity
+    // is >30 cols. Merging would truncate the branch to 30 cols and drop the
+    // cwd, leaving an ambiguous truncated string as the SOLE location signal —
+    // strictly worse than the un-deduped line. So dedupe falls back and keeps
+    // the cwd leaf as a second signal. Wide terminal so the cwd leaf (which
+    // formatCwd always preserves) isn't itself right-edge truncated.
+    stream.columns = 200;
+    const status = new StatusLine({ stream: stream as unknown as NodeJS.WriteStream, throttleMs: 0 });
+    status.start();
+    stream.writes.length = 0;
+    status.repaint({
+      model: 'sonnet',
+      cwd: '/Users/x/proj/.afk-worktrees/afk-20260705-142358-verylongsuffix-xy',
+      branch: 'afk/20260705-142358-verylongsuffix-xy',
+    });
+    const out = lastJoined(stream).replace(BROAD_ANSI_RE, '');
+    // The cwd leaf (dash-form) is preserved — under the merge path it would be
+    // hidden entirely, so its presence proves the >30-col fallback fired.
+    expect(out).toContain('afk-20260705-142358-verylongsuffix-xy');
+    // The branch segment still renders…
+    expect(out).toContain('⎇');
+    // …but capped at 30 cols, so the full slash-form branch does not appear.
+    expect(out).not.toContain('afk/20260705-142358-verylongsuffix-xy');
+    status.stop();
+  });
 });
--- a/src/cli/status-line.test.ts
+++ b/src/cli/status-line.test.ts
@@ -945,3 +945,78 @@ describe('StatusLine narrow-terminal priority drop', () => {
     status.stop();
   });
 });
+
+describe('StatusLine cwd/branch worktree dedupe', () => {
+  let stream: MockStream;
+
+  beforeEach(() => {
+    stream = mockStream({ isTTY: true, rows: 24 });
+  });
+
+  it('renders one merged ⎇ segment when branch matches the cwd basename (afk worktree pattern)', () => {
+    const status = new StatusLine({ stream: stream as unknown as NodeJS.WriteStream, throttleMs: 0 });
+    status.start();
+    stream.writes.length = 0;
+    status.repaint({
+      model: 'sonnet',
+      cwd: '/Users/x/proj/.afk-worktrees/afk-20260705-142358-47b3ec',
+      branch: 'afk/20260705-142358-47b3ec',
+    });
+    const out = lastJoined(stream).replace(BROAD_ANSI_RE, '');
+    // The branch spelling appears (once, via the merged ⎇ segment)…
+    expect(out).toContain('⎇ afk/20260705-142358-47b3ec');
+    // …and the cwd spelling (slash → dash) does not: the cwd part was omitted.
+    expect(out).not.toContain('afk-20260705-142358-47b3ec');
+    // The shared slug renders exactly once across the whole line.
+    const slug = '20260705-142358-47b3ec';
+    expect(out.indexOf(slug)).toBe(out.lastIndexOf(slug));
+    status.stop();
+  });
+
+  it('keeps the merged segment on a narrow terminal (promoted to never-drop)', () => {
+    stream.columns = 32;
+    const status = new StatusLine({ stream: stream as unknown as NodeJS.WriteStream, throttleMs: 0 });
+    status.start();
+    stream.writes.length = 0;
+    status.repaint({
+      model: 'sonnet',
+      cwd: '/Users/x/proj/.afk-worktrees/afk-20260705-142358-47b3ec',
+      branch: 'afk/20260705-142358-47b3ec',
+      cost: 0.05,
+      tokens: 1200,
+    });
+    const out = lastJoined(stream).replace(BROAD_ANSI_RE, '');
+    // Droppables shed first; the merged location segment survives because it
+    // is never-drop (contrast: the plain droppable branch is shed at this
+    // width in the 'drops the branch before the model' test above).
+    expect(out).toContain('afk/20260705-142358-47b3ec');
+    expect(out).not.toContain('tok');
+    status.stop();
+  });
+
+  it('preserves the #PR suffix inside the merged segment', () => {
+    const status = new StatusLine({ stream: stream as unknown as NodeJS.WriteStream, throttleMs: 0 });
+    status.start();
+    stream.writes.length = 0;
+    status.repaint({
+      model: 'sonnet',
+      cwd: '/Users/x/proj/.afk-worktrees/afk-20260705-142358-47b3ec',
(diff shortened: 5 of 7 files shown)


griffinwork40 (commented):

One high-severity **correctness** gap in the new worktree-dedupe heuristic (display-scoped — not a data-loss/safety issue, but trivially avoidable), plus a related edge case and two test gaps on the same seam. Spec-compliance is clean: every verification claim in the PR body was independently confirmed against the head ref (`711c58d8`) — `palette.fileRef` untouched for other consumers, `extraRows` geometry untouched, no state-machine/paint-mechanics changes, no scope creep. `buildPrompt` / `formatStageRail` signatures unchanged; the palette-mute swap has no unintended consumers.

---

### 🔴 HIGH — dedupe fires on pure name coincidence, no positive worktree signal
`src/cli/status-line.ts:384-396` · correctness · confidence: medium

`worktreeDedupe` triggers whenever `branch.replaceAll('/','-') === basename(cwd)`, with **no check that cwd is actually under `.afk-worktrees/`**. A non-worktree checkout whose directory name coincidentally equals its branch name (e.g. cwd `/tmp/redesign`, branch `redesign`) silently hides the cwd — losing the "where am I?" signal in exactly the case the change was meant to preserve it.

```
const worktreeDedupe =
  f.cwd !== undefined && f.branch !== undefined &&
  f.branch.replaceAll('/', '-') === basename(f.cwd);
```
**Fix:** gate the merge on a positive worktree signal (cwd containing an `.afk-worktrees` path segment) in addition to the name match. (Verified: the `.afk-worktrees` string appears only in a comment at line 372, never in executable code.)

### 🟠 MEDIUM — truncated branch becomes the sole location signal
`src/cli/status-line.ts:389-396` · correctness · confidence: medium

When dedupe fires, cwd is omitted and the only location part is `truncateDisplayWidth(f.branch!, 30)`. A matching identity longer than 30 columns renders an ambiguous truncated string with no cwd fallback — whereas the pre-dedupe path still carried cwd as a second signal.
**Fix:** when the merged `branchText` itself gets truncated, fall back to also showing the raw cwd.

### 🟠 MEDIUM — caret bypass marker degrades to a non-grep-able glyph in transcripts
`src/cli/commands/interactive/repl-loop-shared.ts:70-74` · security · confidence: medium

The caret bypass marker drops the word "bypass" (glyph-only `⚡`). Per the PR's own safety rationale, the prompt is the permission-mode signal that survives into scrollback / piped logs; `bypassPermissions` is the CLI new-install default (`config.ts:364`). Reducing it to a single non-ASCII glyph means a post-hoc `grep bypass` on a piped transcript no longer locates bypass windows.
Note (mitigating): the scrollback-persistence of the prompt echo is repo-confirmed for the **non-TTY** readline surface; on the primary armed-TTY path it is reasoning, not repo-asserted. The glyph is still present and visually distinct at the caret.
**Fix:** retain an ASCII-searchable token in the caret marker (e.g. `⚡bp`) so transcript grep still works.

### 🟡 LOW — no test for the >30-col truncated merged segment
`src/cli/status-line.test.ts:949-1022` · test-coverage · confidence: high

The new dedupe block has no case whose flattened branch/cwd identity exceeds 30 columns, so the truncated-merged-segment path is unexercised. **Fix:** add a case with an identity >30 cols and assert truncation behavior.

### 🟡 LOW — no test for the false-positive (non-worktree coincidence) case
`src/cli/status-line.test.ts:949-1022` · test-coverage · confidence: high

The only non-matching test uses genuinely non-colliding names; there is no case for a coincidental name match that is *not* a worktree (the HIGH finding above). **Fix:** add cwd=`/tmp/redesign`, branch=`redesign` and assert the intended behavior.

---

### What was not checked
- Read against branch HEAD `711c58d890e48f7630122874649a84d75f9f47f3`; all blocking/high/medium citations verified at that ref (0 fabricated, 0 false-absent in the Wave 1.5 verification pass).
- Stated intent: PR #447 title + body — **spec-compliance assessed** (no unmet-intent, no scope-creep found).
- Not independently re-run: the PR's `pnpm test` / `pnpm lint` results (accepted as stated; CI on this PR is currently red — maintainer should confirm the red checks are the pre-existing `chat.*.test.ts` collection errors the body describes, not a regression from this diff).
- Terminal-capture survival of the prompt echo on the primary armed-TTY renderer path is reasoning, not repo-asserted.
- Perf: dedupe compare per repaint is throttled (~10 calls/sec) and sub-microsecond — no hot-path concern.

griffinwork40:
The Wave-1 agent flagged one `medium` correctness item and two "unverified" gaps. Since I have git access to the reviewed ref (the agent doesn't), I'll run Wave 1.5 verification myself on the surviving finding and resolve the two unverified items mechanically.

**Verification of the medium finding (narrow-terminal model truncation):** Confirmed by arithmetic against the actual 32-col test. maxW = 30. Merged segment `⎇ afk/20260705-142358-47b3ec` = 28 display cols (never-drop). After droppables (tokens, cost) shed, remaining never-drop parts = `⎇ …` (28) + ` · ` (3) + `sonnet` (6) = 37 > 30 → hits `truncateDisplayWidth(joined, 30)`, which cuts from the right and **removes the model name**. Under the pre-PR path for the same worktree, the branch was *droppable* — it would drop, leaving the scaled cwd (~12 cols) + model, so `sonnet` survived. So the finding is real: a mild weakening of the "never right-truncate model info" invariant, scoped to worktree sessions on <~40-col terminals. The 32-col test asserts branch-present + tok-absent but does **not** assert model survives, so it doesn't catch this.

**Goal 2 ("no paint-mechanics change") — resolved:** The diff is complete for `loop-stage.ts`; its only two hunks touch `formatStageRail`'s body + two docstrings. Any change to `LoopStageBar.repaint`/paint/geometry would necessarily appear in the diff. It doesn't → claim **confirmed**.

**Test files — resolved:** diff shows `repl-loop-shared.test.ts` as a new file and the `status-line.test.ts` dedupe block inline; assertions match the changed branches.

No `critical`/`high` findings → `/shadow-verify` not required (medium and below pass straight through). Emitting the synthesis.

---

# Code Review — PR #447

**Target:** PR #447 `refactor(repl): calm REPL chrome` · branch `afk/repl-chrome-aprime → main` · reviewed ref `711c58d`
**Change type:** refactor · 7 files · +228/−31 (259 lines) · **light regime** · read-only review
**Stated intent:** available (PR title+body) — spec-compliance assessed against it.

## Findings

**MEDIUM · medium confidence · correctness · `src/cli/status-line.ts:369-395` · file-state · NON-BLOCKING (deliberate tradeoff)**
The dedupe branch caps the merged `⎇ branch` segment at a **fixed 30 cols** and marks it never-drop, whereas the cwd path it replaces scaled to `Math.max(8, floor(maxW*0.4))`. On very narrow terminals (<~40 cols) in an afk worktree, the never-drop set (`branch + model + mode-chip`) overruns `maxW` and falls through to the blind `truncateDisplayWidth(joined, maxW)`, which right-truncates the **model name** — the exact "arbitrarily loses model info" outcome the surrounding invariant comment was written to prevent. Pre-PR, the branch was droppable and would shed here, preserving the model.
- evidence: `const branchText = truncateDisplayWidth(f.branch!, 30); … parts.push({ text: gitText }); // never drop` vs the else-path `const cwdBudget = Math.max(8, Math.floor(maxW * 0.4));`
- verified: at 32 cols, branch(28)+` · `+`sonnet` = 37 > maxW(30) → model truncated; the new 32-col test asserts branch-present but not model-present, so it doesn't guard this.
- suggestion (non-blocking): cap the merged branch to `min(30, maxW - reserve)` scaled to `maxW`, or keep it droppable-but-last (as the old branch was) while still omitting the duplicate cwd — either preserves the model on pathological widths. Given the intent explicitly promotes location to "sole surviving signal," this is a conscious tradeoff and safe to merge as-is; worth a one-line acknowledgment.

*(No other findings. Security: none — pure local terminal rendering. api-compat: none — `buildPrompt` and `formatStageRail` signatures unchanged; `formatStageRail` has no external callers. perf: both new early-returns skip work vs. the paths they replace.)*

## Spec-compliance (vs stated intent)
- **Goal 1 (glyph-only prompt markers): MET** — `repl-loop-shared.ts:59-61` drops worded suffixes; worded chips remain only on the status line (`● plan`/`◐ AFK`/`⚡ bypass`/`○ default` in `formatLine`).
- **Goal 2 (idle rail collapse, no paint-mechanics change): MET** — `formatStageRail` early-returns `· idle` for `observing`; the diff is complete for `loop-stage.ts` and touches only that function body + docstrings, confirming no row-reservation/geometry change.
- **Goal 3 (cwd/branch dedupe, promoted never-drop): MET** — compares raw `f.cwd` via `basename()` (not the display-shortened form), omits cwd, and pushes the merged segment with no `droppablePriority` (never-drop).
- **Scope creep:** none — the only extra behavior (branch color `fileRef`→`dim`) is declared in the intent.

## Test coverage
Adequate for all three surfaces: `repl-loop-shared.test.ts` (4 modes), `loop-stage.test.ts` (idle collapse + full-rail-per-stage + exactly-one-active-diamond), `status-line.test.ts` (merged segment, narrow-width survival, `#PR` suffix, non-matching pair), and the `footer-ghost` probe drives to `acting` to keep testing the full rail. One coverage gap: no test asserts the **model** survives the narrow-terminal dedupe path (see the medium finding).

## What was not checked / epistemic scope
- Read against branch HEAD `711c58d` — the medium finding's mechanics verified against that ref via `git show`; the session working tree is on `main`, so the Wave-1 agent worked from the diff + pasted post-change regions (not live file reads).
- Did not run the test suite (PR body reports full green: 10901 passed; 4 pre-existing `chat.*.test.ts` collection errors reproduce on clean `main`).
- Did not verify `palette.fileRef`'s other consumers remain unaffected (PR claims untouched; not load-bearing for this diff).
- Did not exercise a live TTY at <40 cols; the truncation finding is derived arithmetically from the drop-loop + `truncateDisplayWidth` fallback.

Clean, well-documented, well-tested refactor that implements all three stated goals. The single medium finding is a deliberate, narrow-terminal display tradeoff (model name can right-truncate below ~40 cols in a worktree) — worth a one-line author acknowledgment or a follow-up, but not a merge blocker.

## Done
- Reviewed PR #447 read-only across all six dimensions (light regime, 1 Wave-1 agent + self-run Wave 1.5 verification against ref `711c58d`); no files, PR, or working tree mutated.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/dc7fff28-ed8b-4926-910e-5e3373c38407/pr-447.diff`; reviewed ref `711c58d890e48f7630122874649a84d75f9f47f3`; finding mechanically verified via `git show 711c58d:src/cli/status-line.ts` (drop-loop + final `truncateDisplayWidth`).
- Result: **MERGE** — 1 non-blocking `medium` (narrow-terminal model truncation in the deduped never-drop path), 0 high/critical, all 3 stated goals met.
- Pending: none — review is terminal; acting on the finding (if desired) is the author's call, out of scope for this read-only pass.

---

griffinwork40:
Both absence claims **confirmed** at the reviewed ref: dedupe tests use only 26-col and 37-col identities (no 30-col boundary case), and all 4 `.afk-worktrees` test paths are non-root (`/Users/x/proj/.afk-worktrees/…`, no `/.afk-worktrees/x`). The `nit` is diff-context-verifiable from the hunk I read (line 405: `truncateDisplayWidth(f.branch!, 30)` inside the `worktreeDedupe` block that already guards `f.branch !== undefined`).

**Shadow-verify gate:** skipped — the skill triggers `/shadow-verify` only when a `critical` or `high` finding is present. Max severity here is `low`, so it goes straight through (the generic framework nudge is superseded by the skill's explicit threshold).

---

# Code Review — PR #447

**Target:** `refactor(repl): calm REPL chrome (glyph mode marker, idle rail, cwd/branch dedupe)` · `afk/repl-chrome-aprime` → `main`
**Reviewed ref:** `b6dd137773d57801d0c2f222a2bf58d1aead1179`
**Scope:** 7 files, +300/−31 (3 source, 4 test) · regime = full · 2 parallel review agents + 1 verification backstop

## Verdict by dimension

| Dimension | Result |
|---|---|
| security | **clean** — safety claim independently verified (see below) |
| api-compat | **clean** — all 4 exported symbols non-breaking |
| correctness | **clean** (1 nit) |
| spec-compliance | **clean** — all 4 stated changes implemented, no scope creep |
| test-coverage | 2 low gaps |
| perf-observability | **clean** — throttled, non-hot-path |

## What was verified (not just asserted)

- **Bypass auditability safety claim holds on both legs.** The prompt caret keeps a grep-able ASCII token — `repl-loop-shared.ts:77` `palette.bypass(' ⚡bp')` — and the scroll-region carve-out is real DECSTBM, not just a comment: `status-line.ts:506-513` `writeScrollRegion` restricts scroll rows and `repaint()` paints the reserved row via save/cursor-address/restore, so it never enters scrollback; StatusLine is additionally TTY-gated (no-op when piped). Glyph-only-at-caret preserves the only mode signal that survives into the linear transcript.
- **Behavior-change reachability confirmed non-breaking.** `formatStageRail('observing')` now returns `· idle`; its sole caller `LoopStageBar.repaint()` writes the string straight to the stream and no production caller inspects it (`git grep` → 3 hits, all in-file). `buildPrompt` callers (`loop-iteration.ts`, `surface-setup.ts`) consume the return opaquely.
- **All 4 spec clauses met**, and `palette.fileRef` untouched in `status-line.ts` (repo-wide grep → 2 hits, both in `input-highlight.ts`).

**1 · low · test-coverage · `src/cli/status-line.test.ts:962-1056`** *(absence re-verified: confirmed-absent)*
finding: no test exercises the `displayWidth(f.branch) <= 30` boundary at `status-line.ts:399` — dedupe cases use 26-col and 37-col identities only, never exactly 30.
suggestion: add a 30-col identity asserting merge fires and `truncateDisplayWidth` is a no-op passthrough.

**2 · low · test-coverage · `src/cli/status-line.test.ts:949-1067`** *(absence re-verified: confirmed-absent)*
finding: no test sets a root-adjacent worktree path (`/.afk-worktrees/x`) to exercise `dirname('/')`→`'/'`, `basename('/')`→`''` in the parent-dir guard at `status-line.ts:397`.
suggestion: add `cwd: '/.afk-worktrees/x', branch: 'x'` asserting dedupe does not fire.

**3 · nit · correctness · `src/cli/status-line.ts:398-409`** *(diff-context)*
finding: `f.branch!` at line 405 is safe only because `f.branch !== undefined` sits earlier in the same `worktreeDedupe` conjunction — no independent guard at the use site.
suggestion: bind `const branch = f.branch;` when `worktreeDedupe` is true, or narrow with `if (worktreeDedupe && f.branch)`.

## What was not checked
- Refs: both Wave 1 agents read against branch HEAD `b6dd1377`; I independently re-verified both absence claims at `b6dd1377`. All citations verified at that ref.
- Stated intent: PR #447 title+body — **spec-compliance assessed**, all 4 clauses confirmed implemented.
- I did **not** re-run the test suite (relied on the PR's stated `10901 passed / 0 failing`) and did **not** independently confirm the 4 `chat.*.test.ts` collection errors are pre-existing on `main`.
- No TTY render was performed — visual chrome appearance judged from code + tests only.

## Merge decision: **MERGE**

Zero blocking, critical, high, or medium findings. The three items are optional hygiene (two boundary-test additions and one non-null-assertion tidy) — none gate the merge. Spec-compliance is fully satisfied and the load-bearing safety claim (bypass-mode auditability) is factually verified against the reviewed ref.

**Done**
- Reviewed PR #447 at ref `b6dd1377` across security, api-compat, correctness, spec-compliance, test-coverage, perf-observability using 2 parallel research agents + an inline absence-claim verification backstop.
- Evidence: findings and citations above resolve against `b6dd137773d57801d0c2f222a2bf58d1aead1179`; the two `low` absence claims were independently re-grepped and confirmed (`branch:`/`.afk-worktrees` grep output in this session); the `nit` is verifiable in the diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/20f8cb60-5207-4ada-925b-4b322dd6c996/pr-447.diff`.
- World unchanged: read-only review — no files, git state, or PR touched; the operator's 6 uncommitted working-tree changes were preserved.
- Terminal recommendation: **MERGE** (3 non-blocking low/nit items surfaced for optional follow-up). Shadow-verify not triggered (no critical/high findings). This skill is read-only and does not apply findings or post to the PR.

---

griffinwork40 (commented):

One **HIGH** correctness finding: the new never-drop worktree-dedupe segment reintroduces model-name truncation on narrow terminals — the exact "arbitrarily loses model info" outcome the `formatLine` invariant comment forbids, and a fresh regression unique to this dedupe path. It was independently re-derived by **executing the real `StatusLine` render** at the reviewed ref (not just re-reading source): at 32 cols the rendered line is `⎇ afk/20260705-142358-47b3ec …` with the model `sonnet` truncated off. The fix is one line; everything else in the PR is clean and the three prior-round findings are correctly resolved.

Reviewed at ref `b6dd137` (branch HEAD). Read-only review — findings are reported, not applied. This HEAD is a follow-up to the earlier reviewed commit `711c58d`; its HIGH (name-coincidence dedupe) and two MEDIUMs (truncated-branch-sole-signal; non-grep-able bypass glyph) are all **verified fixed** at this ref (see "Compliant / verified").

---

### 🔴 HIGH · correctness · model name is right-truncated on narrow terminals when the worktree dedupe fires
`src/cli/status-line.ts:405-412` (merged segment) + `:436` (model) + `:494` (terminal fallback) · ref:b6dd137 · file-state

When `worktreeDedupe` fires, the merged `⎇ <branch>` segment is pushed with **no `droppablePriority`** (never-drop), and the model is also never-drop. Nothing checks whether that combined never-drop floor fits `maxW` before the final blind `return truncateDisplayWidth(joined, maxW)`. Unlike the else-branch — where the branch carries `droppablePriority: 1` and sheds *before* truncation reaches the model — the merged segment is fixed-width and cannot drop, so the invariant the comment at `formatLine`'s top promises ("shed lower-priority fields **before** resorting to right-edge truncation that arbitrarily loses model info") is not delivered on this path.

*evidence (the PR's own fixture, `status-line.test.ts` narrow-terminal case):* branch `afk/20260705-142358-47b3ec` = 26 cols → `⎇ `+branch = 28 cols; +` · `(3)+`sonnet`(6) = **37-col never-drop floor**. At `columns:32`, `maxW=30`. Droppables (tokens, cost) shed, then `truncateDisplayWidth(joined, 30)` yields `⎇ afk/20260705-142358-47b3ec …` — **`sonnet` gone**.
```js
const branchText = truncateDisplayWidth(f.branch!, 30);
let gitText = `${palette.dim('⎇')} ${palette.dim(branchText)}`;
parts.push({ text: gitText }); // never drop   ← fixed-width, cannot shed
```
*verification:* re-derived twice — (1) executing the real `StatusLine.repaint` at ref `b6dd137` with a `MockStream{columns:32}` → rendered output omits `sonnet`; (2) the project's own `displayWidth`/`truncateDisplayWidth` on the exact parts pipeline → identical result. This is a **new regression** unique to the dedupe path (the else-branch preserves the model at the same width).
*suggestion (one line):* keep the merged segment droppable-but-last (`droppablePriority: 1`, as the old branch was) while still omitting the duplicate cwd — or cap the merged branch to `min(30, maxW − reserve-for-model)` instead of a flat 30. Either preserves the model on pathological widths without losing the dedupe.

### 🟡 LOW · test-coverage · narrow-terminal dedupe test asserts branch survival but not model survival
`src/cli/status-line.test.ts` (the `columns = 32` "keeps the merged segment … never-drop" case) · ref:b6dd137 · file-state

The new narrow-width test asserts `toContain('afk/20260705-142358-47b3ec')` and `not.toContain('tok')`, but has no `toContain('sonnet')` — so it passes green while the model is silently truncated (the HIGH above). Every *other* narrow-terminal describe block in this file asserts the model survives; this one does not.
*suggestion:* add `expect(out).toContain('sonnet')` to that case — it will fail today and lock the fix.

### 🟡 LOW · security · bypass grep-tag `bp` collides with unrelated substrings in piped transcripts
`src/cli/commands/interactive/repl-loop-shared.ts:73-79` · ref:b6dd137 · file-state

The `⚡bp` marker restores a searchable ASCII token (good — resolves the prior round's finding). But a bare `grep bp` on a transcript also matches words like "subprocess", producing false positives when scanning for permission-off windows. Non-blocking; the compound token `⚡bp` (glyph+tag) is unambiguous.
*suggestion:* note in the docstring that transcript scans should grep the compound `⚡bp`, not bare `bp`.

## Compliant / verified (no finding)
- **Prior-round HIGH — name-coincidence dedupe — FIXED.** `basename(dirname(f.cwd)) === '.afk-worktrees'` now gates the merge; the new `/tmp/redesign`+branch `redesign` test confirms a plain checkout is not deduped. ✓
- **Prior-round MEDIUM — truncated-branch-sole-signal — FIXED.** `displayWidth(f.branch) <= 30` falls through to the else-branch (keeping cwd) when the identity would truncate; the >30-col test confirms it. ✓
- **Prior-round MEDIUM — non-grep-able bypass glyph — FIXED** (see LOW nit above for the residual `bp` collision). ✓
- **Goal 1 (glyph-only prompt markers): MET** — `repl-loop-shared.ts` plan→` ●`, autonomous→` ◐`, bypass→` ⚡bp`, default→no marker. Worded chips remain only on the status line. ✓
- **Goal 2 (idle rail collapse, content-only): MET** — `formatStageRail` early-returns `dim('· idle')` for `observing`; `LoopStageBar` body / `extraRows` geometry unchanged. The footer-ghost test's drive-to-`acting` is a legitimate adaptation (the self-heal path is stage-content-agnostic), not a masked regression. ✓
- **Goal 4 (palette mute): MET** — the single `palette.fileRef(branch)` call site became two `palette.dim` sites; `palette.fileRef` remains used only by `input-highlight.ts` (untouched for other consumers). ✓
- **api-compat: clean** — `buildPrompt` / `formatStageRail` / `StatusLine` public signatures unchanged; production importers confirmed present; no breaking change.
- **perf: clean** — dedupe guard is O(path length) on short strings, once per already-throttled repaint.

## What was not checked
- Read against branch HEAD `b6dd137773d57801d0c2f222a2bf58d1aead1179`; every blocking/high citation was verified at that ref (`git show <ref>:<file>`) and the HIGH finding was re-derived by executing the real render code, not from the diff alone.
- Stated intent: PR #447 title+body — **spec-compliance assessed** against its four goals (all four MET; the HIGH is an invariant defect within Goal 3's delivered behavior, not an unmet clause or scope creep).
- Test suite **not independently run** — the PR body's `10901 passed / 4 pre-existing chat.*.test.ts collection errors` claim was accepted as stated (read-only pass). CI on this PR is currently red — a maintainer should confirm the red checks are those pre-existing collection errors and not a regression from this diff.
- Live TTY at <40 cols not exercised interactively; the truncation is proven by executing the real `formatLine` against a mock 32-col stream + the project's own width functions.

Review is read-only — findings reported, no edits/commits/PR mutations made.

griffinwork40 (commented):

# Code Review — PR #447

**Target:** `refactor(repl): calm REPL chrome (glyph mode marker, idle rail, cwd/branch dedupe)` · `afk/repl-chrome-aprime` → `main`
**Reviewed ref:** `0fa20f59eb99c5739decddf16210a57cdf98675a` (branch HEAD)
**Scope:** 7 files (3 source, 4 test), +320/−31 · refactor · **full** regime · 2 parallel Wave-1 agents + inline Wave-1.5 backstop · read-only

## Merge decision: **MERGE**

Clean, well-tested refactor that implements all four stated goals. **Zero critical/high/medium findings**; one non-blocking LOW (a missing edge-case test). Notably, this HEAD is a follow-up that specifically fixes the prior round's HIGH finding — narrow-terminal model-name truncation when the worktree dedupe fires — and that fix is **verified to hold at this ref**: the merged `⎇ branch` segment is pushed with `droppablePriority: 1` (drop-last among droppables), not never-drop, and a new regression test asserts the model survives.

## What was verified (not just asserted)

- **Prior-round HIGH (model truncation) — FIXED at this ref.** `status-line.ts` pushes the merged dedupe segment as `parts.push({ text: gitText, droppablePriority: 1 })` — identical priority to the plain-branch fallback, so it sheds *before* the never-drop model is ever right-truncated. Traced through `formatLine`'s drop-loop: at `columns:32` with worktree fields set, tokens→cost→context→location all shed and `sonnet` (the never-drop identity) survives to the final render. The invariant the top-of-method comment promises ("shed lower-priority fields before right-edge truncation that arbitrarily loses model info") is delivered on this path.
- **Regression test for that exact gap is present.** `status-line.test.ts` "sheds the merged location segment before truncating the model on a narrow terminal" (`columns = 32`) asserts `toContain('sonnet')` + `not.toContain('afk/20260705-142358-47b3ec')` + `not.toContain('tok')` — the precise prior-round coverage hole, now closed.
- **Bypass auditability preserved (security).** The caret bypass marker keeps a grep-able ASCII token — `palette.bypass(' ⚡bp')` — and `palette.bypass` (chalk) never injects codes mid-string, so `⚡bp` stays a contiguous searchable substring in piped/non-TTY transcripts. The scroll-region carve-out is executable DECSTBM (`writeScrollRegion`), not just a comment: the worded status-line chip paints at the fixed bottom row via CUP, strictly outside the scroll region, so it never enters scrollback — leaving the caret glyph as the sole mode signal in the linear transcript, exactly as the safety rationale claims.
- **api-compat clean.** `buildPrompt(mode)`, `formatStageRail(active, fmt)`, and the full `StatusLine` public surface are signature-unchanged (the `observing` early-return and `worktreeDedupe` branch are internal only). `palette.fileRef` is untouched (`palette.ts` is not in the diff) and its other consumer `input-highlight.ts:129` is unaffected — the status-line change swaps only a local call-site color.
- **Behavior-change reachability confirmed non-breaking.** `formatStageRail('observing')` → `· idle`; its sole production caller `LoopStageBar.repaint()` writes the string straight to the stream (no structural parsing). The one test that depended on the old 5-cell idle format (`terminal-compositor.footer-ghost.test.ts`) is updated in the same diff (`loopStageBar.repaint('acting')` drives the bar mid-turn before the rail-finder probes).

**LOW · high confidence · test-coverage · `src/cli/status-line.test.ts` (describe: "StatusLine cwd/branch worktree dedupe") · ref:0fa20f5 · file-state**
finding: no test exercises `worktreeDedupe`'s root-adjacent-cwd edge case, where `dirname(cwd)` resolves to `/` (so `basename('/')` → `''`).
evidence:
```js
basename(dirname(f.cwd)) === '.afk-worktrees' &&
f.branch.replaceAll('/', '-') === basename(f.cwd) &&
```
(grepped the full test file at the ref for a root-level cwd fixture / `dirname` / bare `/` parent — zero matches in the dedupe block; absence independently re-confirmed in the Wave-1.5 backstop.)
suggestion: add a case with `cwd: '/.afk-worktrees/x', branch: 'x'` asserting no crash and no false-positive merge. Non-blocking — the guard already fails safely at this input (verified by reading the code at the ref: `basename('/')` yields `''`, which does not equal `'.afk-worktrees'`, so the dedupe correctly does not fire and nothing throws).

## Clean dimensions (verified at the reviewed ref)
- **correctness:** the three-condition `worktreeDedupe` guard short-circuits safely (including the root-adjacent path above); the drop-last merged segment preserves the model on narrow terminals; the `formatStageRail` early-return breaks no caller.
- **spec-compliance: MET** — all four goals implemented (glyph-only caret markers; idle rail collapse, content-only with `extraRows` row-reservation unchanged; cwd/branch dedupe with the positive `.afk-worktrees` signal + 30-col fallback; `palette.fileRef`→`palette.dim` branch mute). No unmet clause, no scope creep. The drop-last behavior is part of goal 3's delivered behavior, not creep.
- **security:** no secrets, injection, or log-sanitization paths touched; bypass auditability and scroll-region isolation verified above.
- **perf-observability:** the uncached dedupe compare (2 `path` calls + 1 `replaceAll` + equality) runs inside `formatLine`, invoked only from the throttled `repaint()` (≤~10 calls/sec) — not a hot path. No logging dropped.

## What was not checked
- Wave-1 agents read file content at branch HEAD `0fa20f59eb99c5739decddf16210a57cdf98675a` via GitHub's content-addressed API (blob-identical to `git show <sha>:<path>`); all citations resolve at that ref. The two load-bearing facts (drop-last priority; the `columns:32` model-survival assertion) were independently re-verified in the inline backstop.
- The test suite was **not** independently re-run (read-only pass). The PR body reports `pnpm lint` green + `10901 passed / 4 pre-existing chat.*.test.ts collection errors`. **CI on this PR is currently red** — a maintainer should confirm the red checks are those pre-existing collection errors (the sibling refactor PRs #451/#452/#453/#454 show the identical failing-CI signature at matching timestamps, consistent with a shared pre-existing breakage on this stack rather than a regression introduced here), not a regression from this diff.
- `formatStageRail` caller-exhaustiveness rests on targeted reads of the `interactive/` modules (GitHub's repo-wide code-search API returned 401 unauth) — medium-high confidence, not an exhaustive tree grep.
- No live TTY was exercised at <40 cols; the truncation behavior is proven by reading the drop-loop + width functions at the ref, not runtime capture.

Read-only review — findings reported, no edits/commits/PR mutations made. Shadow-verify not triggered (no critical/high findings; max severity LOW).

griffinwork40:
Verification complete. All top findings are confirmed against the reviewed ref (`0fa20f59`). Key results:

- **[1] "never-drop" contradiction CONFIRMED**: the merged segment ships with `droppablePriority: 1` and an explicit `// deliberately NOT never-drop` comment — the *code* is correct and internally consistent; the *PR body* is wrong.
- **[3] bypass-token divergence CONFIRMED**: caret now renders `⚡bp`; the toggle notice (`permission-mode-cycle.ts:60`), status line (`:452`), and 20 README/docs mentions all still use the word "bypass".
- **Correction to a Wave-1 agent claim**: the palette recolor (teal→dim) is *not* unannounced — the PR body explicitly documents it under "Palette mute." That finding is dropped.

I independently re-derived the top (high-confidence) finding directly from source at the ref — the same re-derivation `/shadow-verify` performs — so a redundant verifier wave would be ceremony. Emitting the synthesis.

---

# Code Review — PR #447

**Target:** `refactor(repl): calm REPL chrome (glyph mode marker, idle rail, cwd/branch dedupe)`
**Branch:** `afk/repl-chrome-aprime` → `main` · **Reviewed ref:** `0fa20f59` · **Regime:** full (351 changed lines, refactor)
**Stated intent:** PR #447 title + body — spec-compliance assessed.

## Merge decision: ✅ **MERGE** — with two should-fix items (both prose/decision, **no code blocker**)

The code is correct, well-tested, and does what the diff intends. No finding produces wrong output under normal conditions. The one high-confidence issue is a **PR-description inaccuracy**, not a code defect — but it should be corrected before merge because it will mislead a future maintainer into reintroducing a bug this commit deliberately fixed.

## Findings (sorted: semantic/contract → security → minor)

### 1. Spec-vs-implementation contradiction: "never-drop" — `medium` · high confidence · spec-compliance · `status-line.ts:415-421` · ref:0fa20f59
The PR body states **twice** the deduped location segment is "promoted to **never-drop**." The shipped code pushes it with `droppablePriority: 1` (drop-**last** among droppables) and comments `// deliberately NOT never-drop`. The new regression test (`status-line.test.ts:976-1004`) proves the segment **sheds** at 32 cols so the model survives.
```
parts.push({ text: gitText, droppablePriority: 1 }); // drop last among droppables — matches the plain branch; keeps the model never-drop
```
**Ground truth:** the **code is correct and the better design** — a never-drop location would stack with the never-drop model and force the blind truncator to shear the model off a narrow terminal. The **PR prose is stale/wrong.** Both Wave-1 agents flagged this independently (rated high and low respectively); I reconcile to medium: no wrong output, but a real doc-vs-impl break with a live downstream risk (a maintainer reading the body as spec could "fix" the code back to never-drop and reintroduce the model-shearing bug).
**Suggestion:** Edit the PR body/changelog to say "drop-last among droppables (sheds before the never-drop model on narrow terminals)." Do **not** change the code. Note the corollary: the body's goal "narrow terminals cannot shed the only where-am-I indicator" is not met by design — on ≤~32-col terminals a worktree session shows no location at all. That tradeoff (model > location) is defensible; just reconcile the wording.

### 2. Bypass audit-token diverges across surfaces — `medium` · medium confidence · security/observability · `repl-loop-shared.ts:76-80` · ref:0fa20f59
The caret's bypass marker changed from `⚡ bypass` (pre-PR — greppable) to `⚡bp`. The PR's own safety rationale is that the caret is the **only** permission signal surviving into scrollback/piped logs, and `bypassPermissions` is the new-install default. On a default-bypass session (no toggle fired), a piped transcript therefore never contains the word "bypass" — yet the toggle notice `permission-mode-cycle.ts:60` (`⚡ bypass ON`), the status-line chip `status-line.ts:452` (`⚡ bypass`), and **20** README/docs mentions all use "bypass". An operator auditing a transcript with `grep bypass` gets **zero hits** for a default-mode session.
```
mode === 'bypassPermissions' ? palette.bypass(' ⚡bp') :
```
The signal **is** preserved (`⚡bp` renders as one contiguous, greppable token — verified), so this is not a security hole; it's a cross-surface search-term inconsistency that partially undercuts the stated auditability rationale.
**Suggestion:** Either keep a literal "bypass" substring at the caret (`⚡bypass`, dropping only the inner space) **or** document `⚡bp` as the canonical transcript audit token in README/docs so the search term stays in sync across surfaces.

### 3. Worktree dedupe keys on dir-name, not the meta marker — `low` · medium confidence · correctness · `status-line.ts:404` · ref:0fa20f59
`worktreeDedupe` treats `basename(dirname(cwd)) === '.afk-worktrees'` + slug match as proof of a managed worktree, without checking the `.afk-worktree-meta.json` that `worktree.ts` actually writes. A coincidentally-named non-AFK directory could trigger cwd suppression. Narrow (requires parent dir literally `.afk-worktrees` **and** branch-slug match), cosmetic-only consequence.
**Suggestion:** Accept as-is (the compound predicate makes false-positives rare), or gate on the meta file if a stronger guarantee is wanted.

### 4. Test gaps — `low`/`nit` · test-coverage · `status-line.test.ts:949-1077` · ref:0fa20f59
- No isolated test for "parent **is** `.afk-worktrees` but branch no longer matches the slug" (mid-session manual checkout — the case the code comment at `:397` explicitly anticipates). Behaviorally covered by the predicate, not asserted in isolation.
- The `palette.fileRef`→`palette.dim` recolor has no regression test (every assertion strips ANSI via `BROAD_ANSI_RE`). Cosmetic; the recolor is announced in the body.

## Verified clean (read at ref 0fa20f59)
- **api-compat:** `buildPrompt` (4 prod callers), `formatStageRail` (1 caller, internal to `LoopStageBar`), `StatusLine.formatLine` (private, 0 external importers) — signatures unchanged; every consumer treats output as an opaque display string. No breaking surface.
- **Idle-rail collapse:** `formatStageRail` early-returns only on `'observing'`; the full 5-cell rail renders for every other stage (`loop-stage.test.ts:206-219`). Matches "during an active turn the full 5-cell rail renders exactly as before."
- **buildPrompt glyphs:** ` ●` / ` ◐` / ` ⚡bp` / `''` match the body exactly, pinned by the new `repl-loop-shared.test.ts`.
- **>30-col fallback:** long branch falls through to the else-branch and keeps cwd (`status-line.test.ts:1051-1076`).
- **Palette mute:** recolor is **announced** in the body ("Palette mute: branch text teal→dim") and confined — `palette.fileRef` untouched elsewhere (only `input-highlight.ts`). *(This drops Wave-1's "unannounced scope-creep" framing.)*
- **perf-observability:** dedupe is O(1) string compares behind the existing 100ms repaint throttle; `extraRows` geometry is a fixed counter, uncoupled from rendered-string length. No hot-path cost.

## What was not checked
- Read against branch HEAD `0fa20f59` — all blocking/high citations verified at that ref via `git show`.
- **Not run:** `pnpm test` / `pnpm lint`. The body claims 10901 passed / 0 failing and lint green; unverified here (read-only review). The 4 `chat.*.test.ts` collection errors claimed pre-existing on `main` were not independently confirmed.
- Stated intent: PR #447 title+body — spec-compliance assessed (see finding #1).
- Not exercised: live terminal rendering at various column widths (assessed via code + unit tests only); actual scrollback/piped-log capture behavior (assessed via the `writeScrollRegion` rationale in code, not a live capture).

---

**Done** — Reviewed PR #447 read-only (full regime): resolved diff from the preflight artifact, captured stated intent + reviewed ref `0fa20f59`, ran a 2-agent Wave-1 review, and independently re-derived the top findings against source at the ref. Verdict: **MERGE**, with two should-fix items — (1) correct the PR body's "never-drop" wording to match the shipped `droppablePriority: 1` code (code is correct; prose is wrong), and (2) decide on the `⚡bp` vs "bypass" cross-surface audit-token divergence. Evidence: findings cite `file:line` at ref `0fa20f59`; verification commands run inline this turn (git show of `status-line.ts:408-452`, cross-surface `bypass` grep). No repository, PR, or working-tree state was mutated (working tree left dirty as found).

---

PR #448, opened 2026-07-06
feat(subagent): uncap turn/tool-use budgets by default; graceful wind-down on cap

griffinwork40:
## What & why

Two related changes to how `agent`-tool subagents are bounded.

### 1. Uncapped by default, settable per dispatch
- **`max_turns`** now defaults to **unlimited** (was default `10`, hard-clamped to `[1,50]`). Pass a positive integer to cap; `0`/negatives mean unlimited (matches `AgentSession`'s existing falsy-`maxTurns` = no-cap check).
- **`max_tool_use_iterations`** — new `agent`-tool param (the real anti-hang knob: tool-use rounds within the child's single turn), also **default unlimited** on the agent-tool path.
- Both are settable two ways, same precedence as before: **explicit call-site > agent-definition frontmatter (`maxTurns` / `maxToolUseIterations`) > default**.
- **Scoped to the agent-tool path** (`child-config.ts`). Skill/compose internal forks keep `SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS = 50` as a retained anti-hang backstop; openai-compatible models keep their provider-internal 50-round cap regardless.

### 2. Fix: hitting the tool-use cap no longer "gets stuck"
Previously, hitting the tool-use iteration cap ended the turn **the instant a tool round finished** (`loop.ts`), before the model got a turn to write a final answer — so history ended on `[assistant: tool_use][user: tool_result]` with **no final message**, and the top-level closure classifier sealed it as a clean `model_end_turn`. A silent stop with no answer is indistinguishable from a hang (the same failure mode the `refusal` branch already guards against).

Now:
- The anthropic-direct loop runs **one final tools-stripped "wind-down" round** on the cap, so the model synthesizes a real answer from what it gathered (with a brief note appended telling it the tool budget is spent). A `capReached` guard hard-stops if the wind-down pathologically emits another `tool_use`.
- The final `turn.completed` still carries `stopReason: 'tool_use_loop_capped'`, and `closure-reason.ts` now maps that to **`iteration_cap`** instead of silently classifying it as `model_end_turn`.

## Behavior / safety notes
- Default path (unlimited) sidesteps the cap entirely; the graceful-cap behavior only applies when a cap is explicitly set (or on the skill/compose 50-round default).
- Foreground subagents remain bounded by the parent turn + AbortGraph; background jobs are cancelled on parent-session end. Cost-budget plumbing to children remains a separate follow-up (noted below).

## Tests
- `pnpm lint` (tsc strict) clean.
- Updated/added coverage: `loop.test.ts` (wind-down round + tools-stripped assertion), `anthropic-direct.test.ts` (end-to-end cap plumbing), `closure-reason.test.ts` (`iteration_cap`), `parser.test.ts` + `subagent-executor*.test.ts` + `schemas.test.ts` (uncapped defaults, new param, frontmatter).
- Full suite: **0 failed tests** (10,897 passed). The 4 `src/cli/commands/chat.*.test.ts` file-load failures are **pre-existing** (a `builtinToolSchemas` mock issue reproducible at clean `HEAD` with this branch stashed) and unrelated to this change.

## Follow-up (not in this PR)
- Plumb `maxBudgetUsd` (and/or a wall-clock timeout) to forked children as a real unattended-spend backstop, now that agent-tool subagents can run uncapped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ auto-release workflow to deduplicate commits across successive runs.
 
 ## [Unreleased]
 
+### Changed
+- `agent` tool subagents are now uncapped by default and settable per dispatch: `max_turns` defaults to unlimited (was default 10, hard-clamped to `[1,50]`) — pass a positive integer to cap. Added a `max_tool_use_iterations` param and a matching `maxToolUseIterations` agent-frontmatter field (both default unlimited on the agent-tool path). Skill/compose internal forks keep the 50-round anti-hang default; openai-compatible models retain a provider-internal 50-round cap regardless.
+
+### Fixed
+- Hitting the tool-use iteration cap no longer ends a turn silently (which read as a hang / empty subagent result). The anthropic-direct loop now runs one final tools-stripped "wind-down" round so the model synthesizes a real answer from what it gathered, and the closure classifier maps the capped stop reason to `iteration_cap` instead of silently sealing it as a clean `model_end_turn`.
+
 ## [5.23.2] - 2026-07-05
 
 ### Fixed
--- a/src/agent/agents/parser.test.ts
+++ b/src/agent/agents/parser.test.ts
@@ -98,6 +98,36 @@ prompt`;
     expect(parsed?.bashReadOnly).toBe(true);
   });
 
+  it('parses maxToolUseIterations frontmatter (camelCase + kebab alias)', () => {
+    const camel = parseAgentMarkdown(`---
+name: capped
+description: capped rounds
+maxToolUseIterations: 12
+---
+prompt`);
+    expect(camel?.definition.maxToolUseIterations).toBe(12);
+
+    const kebab = parseAgentMarkdown(`---
+name: capped2
+description: capped rounds
+max-tool-use-iterations: 8
+---
+prompt`);
+    expect(kebab?.definition.maxToolUseIterations).toBe(8);
+  });
+
+  it('ignores non-positive maxToolUseIterations with a warning', () => {
+    const warn = vi.fn();
+    const parsed = parseAgentMarkdown(`---
+name: bad
+description: bad cap
+maxToolUseIterations: 0
+---
+prompt`, warn);
+    expect(parsed?.definition.maxToolUseIterations).toBeUndefined();
+    expect(warn).toHaveBeenCalled();
+  });
+
   it('records recognized long-tail fields as ignoredKeys and warns on unknown keys', () => {
     const warn = vi.fn();
     const doc = `---
--- a/src/agent/agents/parser.ts
+++ b/src/agent/agents/parser.ts
@@ -120,6 +120,7 @@ export function parseAgentMarkdown(
   let description: string | undefined;
   let model: string | undefined;
   let maxTurns: number | undefined;
+  let maxToolUseIterations: number | undefined;
   let bashReadOnly: boolean | undefined;
   let tools: string[] | undefined;
   let disallowedTools: string[] | undefined;
@@ -173,6 +174,13 @@ export function parseAgentMarkdown(
         else warn(`invalid ${key} value ${JSON.stringify(inline.trim())} — ignored`);
         break;
       }
+      case 'maxtooluseiterations':
+      case 'max-tool-use-iterations': {
+        const parsed = Number.parseInt(stripQuotes(inline), 10);
+        if (Number.isFinite(parsed) && parsed > 0) maxToolUseIterations = parsed;
+        else warn(`invalid ${key} value ${JSON.stringify(inline.trim())} — ignored`);
+        break;
+      }
       case 'bash': {
         // AFK extension: `bash: read-only` gates the child's shell to
         // non-mutating commands (classifyBashCommand) when bash is granted.
@@ -213,6 +221,7 @@ export function parseAgentMarkdown(
     ...(disallowedTools !== undefined && disallowedTools.length > 0 ? { disallowedTools } : {}),
     ...(model !== undefined ? { model } : {}),
     ...(maxTurns !== undefined ? { maxTurns } : {}),
+    ...(maxToolUseIterations !== undefined ? { maxToolUseIterations } : {}),
   };
 
   return {
--- a/src/agent/providers/anthropic-direct.test.ts
+++ b/src/agent/providers/anthropic-direct.test.ts
@@ -686,13 +686,15 @@ describe('AnthropicDirectProvider', () => {
   it('caps the tool-use loop at config.maxToolUseIterations (config → provider → loop)', async () => {
     // End-to-end plumbing guard: AgentConfig.maxToolUseIterations must thread
     // through AnthropicDirectProvider.query() → AnthropicDirectQueryOptions →
-    // the constructor → runInput → the loop's `maxIterations`. The mock model
-    // emits tool_use on every round, so the ONLY thing that can terminate the
-    // turn is the iteration cap — this is the guard a forked subagent relies on
+    // the constructor → runInput → the loop's `maxIterations`. After the cap the
+    // loop runs one tools-stripped wind-down round (rounds 1-2 request tools;
+    // round 3 answers in text), so the turn ends with a real final message AND
+    // stopReason 'tool_use_loop_capped' — the guard a forked subagent relies on
     // to avoid hanging its parent (see SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS).
     let callIdx = 0;
     messagesCreateMock.mockImplementation(() => {
       callIdx += 1;
+      if (callIdx >= 3) return fromArray(makeTextStream('Summary of findings.'));
       return fromArray(
         makeToolUseStream(`toolu_${callIdx}`, 'get_weather', '{"city":"SF"}'),
       );
@@ -715,8 +717,8 @@ describe('AnthropicDirectProvider', () => {
     });
     const events = await collect(query);
 
-    // Stopped after exactly 2 tool-use rounds, not left to spin unbounded.
-    expect(messagesCreateMock).toHaveBeenCalledTimes(2);
+    // 2 tool-use rounds + 1 tools-stripped wind-down round, not left to spin.
+    expect(messagesCreateMock).toHaveBeenCalledTimes(3);
     const completed = events.find((e) => e.type === 'turn.completed');
     expect(completed).toBeDefined();
     if (completed?.type === 'turn.completed') {
--- a/src/agent/providers/anthropic-direct/loop.test.ts
+++ b/src/agent/providers/anthropic-direct/loop.test.ts
@@ -342,14 +342,25 @@ describe('loop.ts runTurn', () => {
     expect(completed).toBeUndefined();
   });
 
-  // covers lines 324-331: maxToolUseIterations cap
-  it('caps tool-use iterations when maxToolUseIterations is set', async () => {
+  // covers the tool-use cap + wind-down: after `maxToolUseIterations` tool
+  // rounds the loop runs ONE tools-stripped round so the model can synthesize a
+  // final answer, then completes with stopReason 'tool_use_loop_capped' — no
+  // silent mid-round stop.
+  it('caps tool-use iterations, then runs a tools-stripped wind-down round', async () => {
     let callIdx = 0;
-    const client = makeClient(() => {
-      callIdx += 1;
-      // Always return tool_use — would loop forever without the cap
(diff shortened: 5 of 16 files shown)


griffinwork40 (commented):

**Merge decision: MERGE** (all findings non-blocking; medium/low)

Reviewed at ref `061345682` (branch HEAD). Read-only review — findings are reported, not applied.

**Spec-compliance — intent MET.** Both stated changes are implemented with nothing unrequested. The safety-critical claim (`0`/falsy `max_turns` = unlimited) is verified at the actual enforcement site: `agent-session.ts:1409` short-circuits on falsy `maxTurns` (`if (this.config.maxTurns && this.turnCount >= this.config.maxTurns)`), so the default-`0` path disables the cap as intended.

**Correctness — clean.** The wind-down control flow is sound: the `if (capReached)` hard-stop (`loop.ts:747`) is physically ordered before the `iterations >= maxIterations` cap block (`loop.ts:758`), so no double wind-down or infinite loop; the wind-down round fires at most once (tools are stripped via the `!capReached` guard); off-by-one is correct (cap N → exactly N tool rounds, then one wind-down); both exit paths (natural `end_turn`, and the pathological re-emitted `tool_use`) stamp `stopReason: 'tool_use_loop_capped'`. Closure precedence is correct: `iteration_cap` (`closure-reason.ts:88`) sits after abort/provider-error and before truncation/`model_end_turn`.

**Test-coverage — no gaps.** Every claimed test exists at the ref and asserts the claimed behavior (wind-down tools-stripped, `iteration_cap` mapping + abort-outranks, uncapped defaults / no upper ceiling / `-5`→`0`, frontmatter camel+kebab precedence, non-positive frontmatter warn, new schema param present).

**api-compat.** `max_tool_use_iterations` / `maxToolUseIterations` and the `iteration_cap` closure branch are additive and non-breaking. The `max_turns` default change (`10`→`0`, `[1,50]` clamp removed) is a real semantic change, but no in-repo production caller passes `max_turns` literally (test-only), so blast radius in this repo is nil — worth a migration note for out-of-repo skill/frontmatter callers.

**Findings (non-blocking):**
1. `medium · security` — `loop.ts:767`: the wind-down appends a text block to the caller-aliased last message via `lastMsg.content.push(...)`, mutating `input.messages` in place. The pushed text is a static constant (no injection risk), but in-place mutation risks state bleed if that message array is reused across a later turn/retry. Suggestion: build the wind-down request from a shallow-cloned message list rather than mutating `input.messages`.
2. `medium · security` — uncapped-by-default turns + tool-use rounds on the agent-tool path means a single child turn has no automatic count ceiling. A per-*session* `maxBudgetUsd` backstop does exist and is enforced (`stream-consumer.ts:409→428`, wired at `agent-session.ts:435`), and depth-3 nesting + AbortGraph + background-job cancellation + the retained 50-round default for skill/compose forks all remain — but whether the dollar budget is plumbed to agent-tool *children* specifically is unverified. This matches the PR's own "Follow-up" note (plumb `maxBudgetUsd`/wall-clock to forked children). Suggestion: land that follow-up so an unattended runaway child self-terminates on spend.
3. `medium · perf-observability` — `loop.ts:747-756`: the rare pathological branch (wind-down round itself emits `tool_use` against the empty toolset) hard-stops without emitting an operator-facing `assistant.message`, unlike the normal wind-down path. Suggestion: emit a display-only notice on that branch so the terminal state isn't a silent stop.

**What was not checked:** Wave-1 agents read against branch HEAD `061345682`; all blocking-tier citations were re-verified at that ref (Wave 1.5 corrected one earlier claim — a cost backstop *does* exist — and one line number). The test-suite pass claim ("10,897 passed / 4 pre-existing chat.* failures") was not run here (read-only). Out-of-repo skill/agent-frontmatter callers of `max_turns` were out of scope.

griffinwork40:
Confirmed decisively via `anthropic-direct/query.ts:15-17,236`: *"Maintains a single `messages: MessageParam[]` array across turns — `runTurn` mutates it in place... the next iteration sees the full history"* and *"`messages` is a stable array reference."* So F2 is confirmed, not inferred — the budget note persists into session history. I now have a fully verified picture with no critical/high findings (so the shadow-verify gate doesn't trigger; the independent verifier already re-derived the correctness/spec claims from scratch). Emitting synthesis.

---

# Code Review — PR #448

**Target:** PR #448 `feat(subagent): uncap turn/tool-use budgets by default; graceful wind-down on cap` · branch `afk/max-turns-limit-check` → `main` · reviewed ref `0613456`
**Triage:** feature · 16 files · +284/−53 (~337 lines) · regime **full**
**Stated intent:** PR #448 title+body (R1–R6) — **spec-compliance assessed**.

## Summary
The change is **correct and well-tested on its primary path**. The anthropic-direct wind-down control flow is sound (exactly one tools-stripped round, correct `end_turn` exit carrying `tool_use_loop_capped`, no infinite loop), `closure-reason` precedence is right, and `child-config` precedence (explicit > frontmatter > default) correctly bypasses the `?? 50` backstop. Findings are **2 medium** (both intentional/scoped tradeoffs worth conscious acceptance) and **3 low** — **no critical or high**.

## Findings (sorted: security → spec/correctness within tier)

### 🟠 MEDIUM · security · uncapping-by-default removes the anti-hang backstop for agent-tool children
`src/agent/tools/subagent/child-config.ts:221` + `src/agent/session/agent-session.ts:1409` · ref:0613456 · file-state
Every `agent`-tool dispatch from an anthropic-model session now forks a child with `maxTurns=0` **and** `maxToolUseIterations=0`. The turn cap is skipped (`if (this.config.maxTurns && …)` short-circuits on falsy `maxTurns`) and the `?? 50` anti-hang round-cap is deliberately bypassed (child-config always assigns `0`). The remaining bounds are the parent turn (foreground), `AbortGraph`, and background-cancel-on-parent-end — **no cost (`maxBudgetUsd`) or wall-clock backstop** on the child. A model stuck retrying a failing tool loops unbounded until the parent turn ends or a human aborts. This is the explicit intent of the PR, and the schema text calls `max_tool_use_iterations` "the real anti-hang knob" — so defaulting it off is a conscious choice, with `maxBudgetUsd` plumbing deferred to a follow-up.
*evidence:* child-config.ts:221 `maxToolUseIterations: effectiveMaxToolUseIterations`; agent-session.ts:1409 `if (this.config.maxTurns && this.turnCount >= this.config.maxTurns)`
*suggestion (non-blocking):* land the deferred `maxBudgetUsd`/wall-clock backstop before, or alongside, relying on uncapped children in unattended (daemon/scheduled) runs.

### 🟠 MEDIUM · spec-compliance · graceful wind-down + `iteration_cap` do not cover the openai-compatible provider
`src/agent/providers/openai-compatible/query.ts:120,411,500-516` · ref:0613456 · file-state + cross-tree grep
The wind-down (R5) and `iteration_cap` mapping (R6) live only in `anthropic-direct/loop.ts`. On the openai-compatible path (`gpt-*`, `o1/o3/o4`, `codex-*`, and local HF-id MLX/llama.cpp/vLLM/ollama models), the `for (iter < MAX_TOOL_ITERATIONS=50)` loop (a) **never reads `config.maxToolUseIterations`** — so the new `max_tool_use_iterations` param is a **silent no-op** for those children — and (b) on cap-exit falls through to `yield {assistant.message, text: finalAssistantText}` (often empty on a tool round) + `finishTurn()` with **no `tool_use_loop_capped` stopReason**, so `closure-reason` classifies it as `model_end_turn` — the exact silent-stop failure mode R5/R6 exist to fix. The 50-hardcode predates this PR (**not a regression**), and R4 discloses openai keeps "a provider-internal 50-round cap regardless" — but that phrasing reads as *cap-value parity* and undersells the two live gaps (no-op param + remaining silent-stop).
*evidence:* `grep tool_use_loop_capped openai-compatible/` → 0 hits; query.ts:120 `const MAX_TOOL_ITERATIONS = 50;`
*suggestion (non-blocking):* tighten R4/CHANGELOG wording to state the wind-down/`iteration_cap` improvements are anthropic-direct-only and `max_tool_use_iterations` has no effect on openai-compatible children; or extend the wind-down to that provider in the follow-up.

### 🟡 LOW · correctness · pathological wind-down tool_use is dispatched *before* the hard-stop
`src/agent/providers/anthropic-direct/loop.ts:708,747` · ref:0613456 · file-state
On the wind-down round tools are stripped, so an anthropic model realistically won't emit `tool_use` (near-unreachable). But if it did, tool dispatch + `input.messages.push(toolResultTurn)` (loop.ts:708) run unconditionally *before* the `if (capReached)` hard-stop (loop.ts:747) — the guard bounds the loop, not the dispatch, so one extra tool executes with real side effects before the stop fires.
*suggestion:* gate the dispatch itself on `!capReached`, not just the post-dispatch continuation.

### 🟡 LOW · correctness · wind-down budget note persists into session history and accumulates
`src/agent/providers/anthropic-direct/loop.ts:766-775` · ref:0613456 · file-state
The note is pushed into `input.messages[last].content`, and `input.messages` **is** the session's persistent history (`anthropic-direct/query.ts:15-17,236`: "runTurn mutates it in place… the next iteration sees the full history"). `capReached` prevents a double-append within one turn, but each additional capped turn in a multi-turn session bakes a fresh verbatim copy of the boilerplate into resent context, and a stale "give your final answer now" instruction lingers on an old tool_result.
*suggestion:* append the note to a request-only copy of the last message, or strip it after the wind-down round completes.

### 🟡 LOW · spec-compliance · frontmatter budgets cannot express 0 = unlimited
`src/agent/tools/subagent/child-config.ts:172-186` · ref:0613456 · file-state
Named-agent `maxTurns`/`maxToolUseIterations` frontmatter are floored with `Math.max(1, Math.floor(...))`, so a frontmatter default can never express 0/unlimited — only the call-site param or the bare default reach it. Minor asymmetry against R1/R2's "0/neg = unlimited" rule, scoped to the frontmatter path (and internally consistent, since the parser already rejects non-positive frontmatter values). Worth a one-line doc note that frontmatter is "positive int or omit."

## Compliant / verified (no finding)
- **R1** — `max_turns` default `0` → unlimited is wired end-to-end (agent-session.ts:1409 falsy-`maxTurns` skips the turn cap). ✓
- **R3** — precedence explicit > frontmatter > default correct for **both** budgets (child-config.ts). ✓
- **R4** — agent-tool path always sets `maxToolUseIterations`, bypassing `subagent.ts:544`'s `?? 50`; skill/compose internal forks (which don't build config here) retain the 50 backstop. ✓
- **R5** — wind-down: exactly one tools-stripped round, correct `end_turn` exit stamping `tool_use_loop_capped`, `capReached` hard-stop prevents unbounded looping. ✓
- **R6** — `closure-reason` `iteration_cap` branch correctly ordered: abort/dispatch-error/provider-error outrank it (test: "abort outranks the iteration cap"); it outranks truncation. ✓
- **Test coverage (anthropic path)** — wind-down + tools-stripped assertion (loop.test.ts), end-to-end cap plumbing (anthropic-direct.test.ts), `iteration_cap` + abort-precedence (closure-reason.test.ts), uncapped defaults + new param + frontmatter parsing/precedence (subagent-executor*.test.ts, schemas.test.ts, parser.test.ts). Solid.

## What was not checked
- **Refs read:** all citations verified against reviewed ref **`0613456`** (materialized reviewed-ref file copies + `git show <ref>:<file>`); the working tree is on `main`, so plain repo reads were avoided.
- **Stated intent:** PR #448 title+body — spec-compliance assessed against R1–R6.
- **Test suite not executed** — relied on the PR's claim (10,897 passing; 4 pre-existing `chat.*.test.ts` failures unrelated). Not independently run.
- **openai-compatible cap behavior** characterized by source-read + grep, not by executing a 50-round runaway.
- **Surfaces not in the diff** (Telegram, CLI REPL wiring) not reviewed.
- **Wave-1 note:** the two initial parallel full-review agents exhausted their tool-use budget on the large files (loop.ts 822L; test files 2000–2600L) and returned partial/empty — recovery was orchestrator-led direct verification at the ref plus one tightly-scoped independent verifier, which re-derived the correctness/spec findings from scratch (an independent cross-derivation stands in for the missing second wave). No critical/high finding surfaced, so the `/shadow-verify` post-synthesis gate did not trigger.

No blocking defects. The primary anthropic-direct path is correct, well-tested, and matches its stated intent. The two medium findings are conscious, documented tradeoffs — the **uncapped-by-default anti-hang removal** (recommend landing the deferred `maxBudgetUsd`/wall-clock backstop before uncapped children run unattended) and the **openai-compatible scope gap** (recommend a one-line wording tighten in R4/CHANGELOG + the `max_tool_use_iterations` schema note that it's a no-op there). The three low findings are cleanup-grade. None gate merge.

Review is read-only — I've reported findings only and made no edits, commits, or PR mutations. STOP.

---

PR #455, opened 2026-07-06
fix(cli): align getApiKey() default-model resolution with getModel()

griffinwork40:
## Bug

`getApiKey()` read `AFK_MODEL ?? CLAUDE_MODEL` and passed the possibly-`undefined` result to `getApiKeyForModel()`, while `getModel()` defaults to the literal `'sonnet'`. With both model vars unset and `AFK_OPENAI_BASE_URL` set, `providerForModel(undefined)` falls through to the Tier-4 env hint → `openai-compatible` → returns `OPENAI_API_KEY`, while the session model resolves to `'sonnet'` → `anthropic-direct`.

Result: an Anthropic-routed session carries an OpenAI credential (401s), and `SubagentManager.parentApiKey` inherits the mismatch for every dispatched subagent.

## Fix

`getApiKey()` now resolves the credential for the same model `getModel()` returns — one shared default-resolution path.

## Testing

- New regression test in `src/cli/shared-helpers.test.ts` asserting provider agreement under the trigger env.
- Call-site audit: all `getApiKey()` callers pair it with `getModel()` for the same session; none depended on the old undefined-model behavior.
- Full suite: 11,081 passed / 14 skipped.

Note: CI will show red due to the org billing lock — tests verified green locally.

### DIFF
--- a/src/cli/shared-helpers.test.ts
+++ b/src/cli/shared-helpers.test.ts
@@ -25,6 +25,51 @@ import { loadConfig } from './config.js';
 import { AnthropicDirectProvider } from '../agent/providers/index.js';
 import { OpenAICompatibleProvider } from '../agent/providers/openai-compatible/index.js';
 
+describe('getApiKey / getModel provider agreement (regression: default-model divergence)', () => {
+  // Regression guard for a bug where getApiKey() re-read `AFK_MODEL ?? CLAUDE_MODEL`
+  // directly (possibly undefined) instead of resolving against getModel() (which
+  // defaults to the literal 'sonnet'). With both model env vars unset and
+  // AFK_OPENAI_BASE_URL set, `providerForModel(undefined)` fell through to the
+  // Tier-4 env hint -> 'openai-compatible' (OPENAI_API_KEY), while getModel()
+  // returned 'sonnet' -> 'anthropic-direct'. The session then paired an
+  // anthropic-routed model with an OpenAI credential (401s downstream).
+  const ORIG_ENV = { ...process.env };
+
+  beforeEach(() => {
+    vi.resetModules();
+    delete process.env['AFK_MODEL'];
+    delete process.env['CLAUDE_MODEL'];
+    delete process.env['AFK_PROVIDER'];
+    delete process.env['ANTHROPIC_API_KEY'];
+    delete process.env['CLAUDE_CODE_OAUTH_TOKEN'];
+    delete process.env['OPENAI_API_KEY'];
+    delete process.env['CODEX_API_KEY'];
+    process.env['AFK_OPENAI_BASE_URL'] = 'http://localhost:8000/v1';
+    process.env['OPENAI_API_KEY'] = 'sk-proj-SENTINEL-OPENAI-KEY';
+    process.env['ANTHROPIC_API_KEY'] = 'sk-ant-api03-SENTINEL-ANTHROPIC-KEY';
+  });
+
+  afterEach(() => {
+    process.env = { ...ORIG_ENV };
+  });
+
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

## Merge recommendation: **MERGE** (one cosmetic nit; no blocking findings)

Minimal, correct, well-tested fix that closes a real credential cross-wiring bug. `getApiKey()` now resolves the credential through `getApiKeyForModel(getModel())`, so it always matches the same model string the session runs with — including the `'sonnet'` default when `AFK_MODEL`/`CLAUDE_MODEL` are unset. This closes the divergence where an unset model routed to `openai-compatible` (via the `AFK_OPENAI_BASE_URL` Tier-4 hint) while the session model resolved to `'sonnet'` (`anthropic-direct`), pairing an Anthropic-routed session with an OpenAI credential (401s), which `SubagentManager.parentApiKey` inherited.

Reviewed at ref `78a051d3fad831a72a3df379f7d3329b37268c20`.

### Findings

**1. nit · test-coverage · `src/cli/shared-helpers.test.ts` (beforeEach)**
Dead `delete process.env['OPENAI_API_KEY']` immediately followed two lines later by `process.env['OPENAI_API_KEY'] = 'sk-proj-SENTINEL-OPENAI-KEY'`. Harmless, but misleads a reader into thinking the key is absent at test time.
Suggestion: drop the `delete` of `OPENAI_API_KEY` since it is unconditionally re-set.

No critical / high / medium findings.

### Dimension results (all read at ref 78a051d3)

- **correctness** — clean. `getModel()` returns the literal `'sonnet'` when both env vars are unset (never `undefined`); `providerForModel('sonnet')` hits the Tier-2 Claude-lock *before* the Tier-4 `AFK_OPENAI_BASE_URL` env-hint branch, so the `undefined`→`openai-compatible` fallthrough is now unreachable via `getApiKey()`. No behavior change for users who set `AFK_MODEL`/`CLAUDE_MODEL` — `getModel()` returns the raw value unchanged.
- **spec-compliance** — satisfied. Delivers the stated intent ("one shared default-resolution path"). No scope creep; no unmet clause.
- **test-coverage** — the new regression test exercises the exact trigger (both model vars deleted + `AFK_OPENAI_BASE_URL` set) and asserts provider agreement + the Anthropic sentinel key. `afterEach` restores `process.env` correctly; `vi.resetModules()` + dynamic `import()` used properly, no leak risk. (One cosmetic nit above.)
- **api-compat** — signature unchanged (`(): string | undefined`). All production call sites pair `getApiKey()` with the same-session `getModel()`; none depended on the old undefined-model behavior (the old path returned an OpenAI key for an Anthropic session — the bug itself). `parentApiKey` propagation verified.
- **security** — the fix *closes* a cross-provider credential leak rather than opening one. `resolveCredentialForModel` gates strictly on the model-derived provider, so `'sonnet'` reads only the Anthropic chain. Test keys are `SENTINEL` fakes; no real credential logged or leaked.
- **perf-observability** — one extra cheap env read via `getModel()`. Negligible. No logging gap.

### What was not checked
- The full test suite was **not run** — the "11,081 passed" claim and the new test's pass are unverified here. CI is currently red, but all four jobs fail in 1–3s (setup/org-billing-lock signature, not a test/lint failure), and Vercel deploys pass — consistent with the author's "org billing lock" note. **Recommend a maintainer visually confirm one green suite run before merge** once billing is restored.
- Did not deep-read `keychain.ts`, `MODEL_MAP`/`resolveBinding` internals, or the provider constructors below the `resolveCredentialForModel` routing branch.

*(Read-only automated review. This tool does not approve, request changes, merge, or otherwise modify the PR.)*

PR #456, opened 2026-07-06
fix(telegram): run safety-net strip before fence restore

griffinwork40:
## The bug

`markdownToTelegramHtml` restores code/fenced-block placeholders (steps 9–10)
*before* the mis-nesting safety net runs (step 11). The safety net strips all
`<b>/<i>/<s>` tags when it finds unbalanced HTML — but by the time it runs on
main, an empty fenced code block has already been restored from its
placeholder into `<i>(empty bash block)</i>`. If the same message also
contains an unrelated interleaved/mis-nested emphasis run, the strip
collateral-damages the fence label's italics too.

Concrete repro:

```
**a _b** c_
```bash
```
```

- On **main**: the interleaved `**a _b** c_` correctly trips the safety net
  and its emphasis is stripped, but the *separately restored* empty-fence
  label `<i>(empty bash block)</i>` gets caught by the same blanket
  `<\/?[bis]>` regex and loses its `<i>` — even though the fence label was
  never part of the imbalance.
- **After this fix**: the safety net runs while the fence is still an opaque
  `\x02FENCED0\x03` placeholder, so the balance check only ever sees real
  emphasis/link tags. The label is restored *after* the strip and keeps its
  italics: `<i>(empty bash block)</i>`.

This also fixes the stated-but-violated invariant from #321 that restore
"must happen last so no inline regex re-scans restored content" — on main the
strip regex runs *after* restore, i.e. it does re-scan restored content.

## Verified against current main

Confirmed by reading `origin/main`'s `src/telegram/formatter.ts`: steps 9–10
(restore) still precede step 11 (safety net strip). The bug is present and
unfixed on main as of this PR.

## Changes

Cherry-pick of `f54acbd` (2nd commit of the branch that produced #321 — #321
merged the 1st commit only) onto current `main`. Clean pick, no conflicts.
Reorders the safety net to run before the code/fenced restore and adds a
regression test asserting the fence label keeps its `<i>` when a separate
mis-nested emphasis run is present in the same message.

## Testing

- `pnpm lint` (`tsc --noEmit`): clean
- `pnpm test -- src/telegram/formatter.test.ts`: 97/97 passed (includes new
  regression test `empty-fence label survives the strip when a separate
  emphasis run is mis-nested`)
- Full suite (`pnpm test`): 596 files / 11100 tests passed, 14 skipped, 0
  failed

**CI red expected: org billing lock; suite green locally.**

### DIFF
--- a/src/telegram/formatter.test.ts
+++ b/src/telegram/formatter.test.ts
@@ -663,4 +663,15 @@ describe('markdownToTelegramHtml — mis-nested emphasis safety net', () => {
     expect(markdownToTelegramHtml('**A** and *b* and snake_case'))
       .toBe('<b>A</b> and <i>b</i> and snake_case');
   });
+
+  test('empty-fence label survives the strip when a separate emphasis run is mis-nested', () => {
+    // The empty-fence placeholder is <i>(empty … block)</i>. Because the safety net
+    // runs before the fenced restore, the label's <i> is not collateral-stripped when
+    // an unrelated interleaved run ("**a _b** c_") in the same message trips the net.
+    const out = markdownToTelegramHtml('**a _b** c_\n```bash\n```');
+    expect(tagsBalanced(out)).toBe(true);
+    expect(out).toContain('<i>(empty bash block)</i>'); // label keeps its italic
+    expect(out).not.toContain('<b>'); // mis-nested emphasis still dropped
+    expect(out).toContain('a b c'); // emphasis text preserved as plain
+  });
 });
--- a/src/telegram/formatter.ts
+++ b/src/telegram/formatter.ts
@@ -152,22 +152,28 @@ export function markdownToTelegramHtml(text: string): string {
   // 8. Headers: drop # prefix, keep text
   out = out.replace(/^#{1,6}\s+/gm, '');
 
-  // 9. Restore code spans and fenced blocks after ALL inline transforms.
-  // External constraint: restore must happen last so no inline regex re-scans
-  // the restored HTML content. Inline code before fenced to prevent nested substitution.
-  out = out.replace(/\x02CODE(\d+)\x03/g, (_m, i: string) => codeSpans[Number(i)] ?? _m);
-  // 10. Restore fenced code blocks
-  out = out.replace(/\x02FENCED(\d+)\x03/g, (_m, i: string) => fencedBlocks[Number(i)] ?? _m);
-
-  // 11. Safety net: interleaved/overlapping emphasis markers (e.g. "**a _b** c_")
+  // 9. Safety net: interleaved/overlapping emphasis markers (e.g. "**a _b** c_")
   // can yield improperly-NESTED tags like "<b>a <i>b</b> c</i>", which Telegram
   // rejects with HTTP 400 "can't parse entities" — failing or degrading the send.
   // Code/pre/link tags are emitted as atomic, always-balanced units, so any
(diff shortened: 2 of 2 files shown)


griffinwork40 (commented):

## Code Review — PR #456 (`fix(telegram): run safety-net strip before fence restore`)
**Ref:** `9c3e4cdaafdd6de6626ff6f74c1db096f0b6b345` · `src/telegram/formatter.ts` (+22/-8), `src/telegram/formatter.test.ts` (+11), 2 files.

**1. CORRECTNESS — fence/code content risk**
No issues found. Top-of-function escaping (inline `&`/`<`/`>` replace, ~line 91-93) runs before fenced extraction (~101) and code-span extraction (~120), so the `inner` captured for a fence/code span already has zero raw `<`/`>`. Steps 1-3 are untouched by this diff (which spans only ~152-176). Restored units (`<pre>`, `<code>`, `<i>(empty…)</i>`) are atomic, tag-free-inside pairs. The new order checks balance (168-169) before restore (175-176), so restored text is never balance-scanned; the old order scanned it, but an atomic pair nets zero stack effect — identical boolean verdict either way. Behavior confirmed unchanged vs. the old order.

**2. CORRECTNESS — final-balance guarantee (walked)**
No issues found. Trace of `'**a _b** c_` + empty ```` ```bash ```` fence: emphasis steps yield `<b>a <i>b</b> c</i>\x02FENCED0\x03`. New order: balance check → false → strip removes the 4 emphasis tags → `a b c\x02FENCED0\x03` → restore → `a b c<i>(empty bash block)</i>`. All 4 test assertions pass. Old order (restore-then-check) stripped 6 tags including the label's own `<i>/</i>`, producing `a b c(empty bash block)` — the `toContain('<i>(empty bash block)</i>')` assertion fails, so the test correctly discriminates old from new.

**3. SPEC-COMPLIANCE**
No issues found on intent — restore (~176) is the last mutation before `return out` (~178); no regex runs after it, so the "#321 restore-last" invariant now genuinely holds (on the old order the strip regex ran *after* restore, self-contradicting the comment). Diff is scoped to the reorder + one test; no scope creep.

**4. TEST-COVERAGE** — low severity, non-blocking
`test.ts` safety-net describe block covers inline `<code>` survival and the new empty-fence `<i>` label, but has no test pairing a mis-nested emphasis run with a **non-empty** `<pre>` fence. Suggestion: add `markdownToTelegramHtml('**a _b** c_\n```bash\nrun\n```')` asserting `<pre>run\n</pre>` survives the strip.

**5. SECURITY / API-COMPAT**
No issues found — function signature unchanged, no new/dynamic regex, pure internal reorder, no I/O.

**Minor (non-blocking):** the new test's `tagsBalanced(out)` assertion is vacuously true in *both* orders for this example (old-order output has zero tags left, not a genuinely-balanced pair); only the `toContain('<i>…')` assertion actually discriminates old vs. new — worth a one-line comment.

---

**VERDICT: MERGE** — the reorder is provably balance-preserving, fixes a self-contradicted invariant from #321, and ships a real (if narrow) regression test; remaining gaps are low-severity and non-blocking.

*What was not checked:* Read against branch HEAD `9c3e4cdaafdd6de6626ff6f74c1db096f0b6b345` — citations verified at that ref. Stated intent taken from PR title + body — spec-compliance assessed. CI is red on this PR (reported by the author as an org billing lock, suite green locally); the failing checks were not independently reproduced. Light regime (single-pass, all dimensions) given the 37-line hotfix scope.

PR #451, opened 2026-07-06
refactor(cli): dedupe elicitation validators into field-validation module

griffinwork40:
## What

Extracts the duplicated number/text field validators out of `elicitation-repl.ts` into a single tested module `src/cli/elicitation/field-validation.ts` (`validateNumberField` / `validateTextField`), and routes all **four** call sites through it.

Before, the same bounds-checking logic lived in 4 copies with byte-identical error text:
- overlay `number` validate closure + overlay `text` validate closure (in `runOverlayTextOrNumber`)
- non-TTY `number` readLine loop + non-TTY `text` readLine loop (in `renderAgentQuestion`)

## Why

Copy-pasted validation is the kind of incidental duplication that makes a long file harder to read and easy to drift (fix a message in one copy, miss the other three). One tested source removes that.

## Behavior preservation

Pure dedup — no behavior change:
- Exact error strings preserved (incl. the `≥`/`≤` bound messages).
- Skip / cancel / empty-input semantics preserved.
- The **overlay-untrimmed vs readLine-trimmed** asymmetry is preserved: the empty-input escape hint legitimately differs per caller (overlay "esc to cancel" vs readLine ":cancel to skip"), so it's threaded through `emptyError` rather than unified away — documented in the module header.

## Verification

- `tsc --noEmit` (strict) — clean
- `src/cli/elicitation-repl.test.ts` — **96/96 pass**
- `audit:env:check` — 0 violations

## Follow-up (not in this PR)

`renderAgentQuestion()` is still a ~340-line `qType` cascade. Splitting its per-type branches (choice / multi_choice / confirm / text / number, overlay + non-TTY paths) into `elicitation/question-handlers.ts` is the natural next step — deferred here to keep this change small and obviously-safe.

PR #457, opened 2026-07-06
refactor(providers): consolidate o-series model detection into isOSeriesModel()

griffinwork40:
## Summary

Consolidates ~4 duplicated o-series-model (`o1`/`o3`/`o4`) detection checks — a mix of `startsWith('o1'|'o3'|'o4')` enumerations and an inline `/^o[0-9]/` regex — into a single `isOSeriesModel()` predicate, the canonical definition now living in `src/agent/model-capabilities.ts`.

### Why

The duplicated checks were an enumerated allowlist (`o1`, `o3`, `o4`) that would silently misclassify any future o-series id (`o5`, `o6`, …) as a non-reasoning model. They also disagreed slightly in scope — some stripped an OpenRouter-style `provider/` prefix, some didn't; none were case-insensitive.

### What changed

- **`model-capabilities.ts`** — new `isOSeriesModel()`: matches any `o<digit>` prefix (not just o1/o3/o4), strips a leading `provider/` segment, case-insensitive.
- **`model-limits.ts`** (`routesToOpenAICompatible`) and **`providers/index.ts`** (`providerForModel`) — now delegate to the shared predicate instead of their own `startsWith` enumerations.
- **`providers/openai-compatible/oneshot.ts`** — drops its inline `/^o[0-9]/` copy in favor of the shared predicate.
- **`providers/openai-compatible/query/model-params.ts`** — its local `isOSeriesModel` is now a re-export of the canonical one, preserving the existing import path for `query.ts` and any other consumers.
- New regression tests in `model-capabilities.test.ts` covering o5+/oN, `provider/`-prefix stripping, and case-insensitivity.

No behavior change for existing o1/o3/o4 ids.

### Rebase note

Rebased onto current `origin/main` (provider refactors #451–454 split `openai-compatible/query.ts` into `query/` submodules and added the tool-loop-cap machinery). The rebase applied cleanly with **no conflicts** — none of the intervening commits touched the o-series check itself, and a repo-wide grep confirms no duplicate `startsWith('o1'|'o3'|'o4')` or `/^o[0-9]/` copies remain outside the canonical predicate.

## Testing

- `pnpm lint` (tsc --noEmit): clean
- `pnpm test`: **596 test files, 11104 tests passed**, 14 pre-existing skips, 0 failures

**CI red expected:** org billing lock; suite verified green locally (11k+ tests).

### DIFF
--- a/src/agent/model-capabilities.test.ts
+++ b/src/agent/model-capabilities.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, beforeEach, afterAll } from 'vitest';
-import { supportsVision } from './model-capabilities.js';
+import { supportsVision, isOSeriesModel } from './model-capabilities.js';
 import { resetSlotBindings } from './session/model-slots.js';
 
 describe('supportsVision', () => {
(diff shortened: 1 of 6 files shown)


griffinwork40:
I have the full picture. Here's my review:

---

## PR #457 Review: Consolidate o-series model detection into `isOSeriesModel()`

### Verdict: **Approve** — clean, well-motivated refactor with one minor suggestion

---

### Correctness ✓

The new `isOSeriesModel()` in `model-capabilities.ts` is a strict behavioral superset of the three enumerated copies it replaces:

| Old copies | New canonical |
|---|---|
| `startsWith('o1') \|\| startsWith('o3') \|\| startsWith('o4')` | `/^o[0-9]/i` — matches any `oN` |
| No provider-prefix handling (except in `oneshot.ts`/`model-params.ts`) | Strips `provider/` via `lastIndexOf('/')` |
| Case-sensitive (relied on caller pre-lowering) | Case-insensitive (`/i` flag) |
| No whitespace tolerance | `.trim()` before match |

All four call sites are correctly updated:
- **`providers/index.ts`**: 3-line `startsWith` chain → single `isOSeriesModel(lowered)` call
- **`model-limits.ts`**: same pattern, same correct replacement
- **`oneshot.ts`**: removes inline `bareModel` strip + regex, delegates to shared function
- **`model-params.ts`**: local definition → import + re-export, preserving the `query.ts` re-export chain

### Backward Compatibility ✓

The re-export chain is preserved: `model-capabilities.ts` → `model-params.ts` (re-export) → `query.ts` (re-export). No external consumers import directly from `model-params.ts` — only `query.ts` re-exports it. The signature widens from `(model: string)` to `(model: string | undefined)`, which is safe — existing callers passing `string` still type-check.

### Test Coverage ✓

5 test cases covering:
1. Known families (o1, o3, o4 + variants) — positive
2. Future o5+/oN — **regression test for the exact gap the enumerated copies had**
3. Provider prefix stripping (OpenRouter-style)
4. Case-insensitive + whitespace tolerance
5. Negative cases (gpt, claude, codex, ollama, empty, undefined)

All 9 tests pass. The regression test in case 2 is particularly good — it documents *why* this consolidation matters.

### Issues / Suggestions

**Minor (non-blocking):**

1. **Redundant processing at call sites** — `providers/index.ts` and `model-limits.ts` pass `lowered` (already trimmed + lowercased) into `isOSeriesModel()` which trims again and applies case-insensitive regex. Harmless but slightly wasteful. Could document in the JSDoc that the function is safe to call on raw or pre-processed input. Not worth blocking on.

2. **`oneshot.ts` comment could be tighter** — The new comment says "Classification (incl. `provider/`-prefix strip) is shared — see `isOSeriesModel`" which is accurate but the parenthetical is slightly awkward. Minor style nit.

**Observation (not an issue):**

The `reasoning-model-predicate` worktree shows a related follow-up that adds `isReasoningModel()` as a superset of `isOSeriesModel()`. This PR lays the groundwork cleanly for that expansion.

### What's NOT a concern

- **No behavioral regression**: the expansion from enumerated o1/o3/o4 to any `oN` is strictly forward-compatible
- **No circular dependency**: `model-capabilities.ts` imports only from `config/env.ts` and `session/model-slots.ts` — neither imports back
- **No security/perf impact**: pure string predicate, negligible cost

### Summary

This is a textbook consolidation refactor: identifies duplicated logic, extracts to a single canonical location with improved robustness, preserves backward compatibility via re-exports, and adds regression tests documenting the gap being fixed. Ship it.

---

PR #459, opened 2026-07-06
fix(telegram): thread openaiBaseUrl so OpenAI Telegram sessions reach the configured endpoint

griffinwork40:
## Problem

`src/telegram.ts` never propagated `openaiBaseUrl` (env `AFK_OPENAI_BASE_URL` / `config.openaiBaseUrl`). As a result the **entire OpenAI-compatible Telegram surface** — the top-level session *and* any OpenAI-routed subagents/skills forked from a Telegram parent — silently targeted `api.openai.com` instead of a configured local shim.

Only `baseUrl` (the Anthropic base) was threaded, as `telegramBaseUrl`; its OpenAI peer was dropped on every Telegram code path. The CLI surfaces (`chat.ts`, `bootstrap.ts`) already thread `cliConfig.openaiBaseUrl` into the child factories + executors — Telegram was the lone surface missing it. It only worked by accident when the operator used *per-slot* config (`AFK_MODEL_<SLOT>_BASE_URL`), which `applySlotCredentials` self-heals.

## Fix (parity with chat.ts)

**Anthropic branch** — capture `telegramOpenaiBaseUrl` and thread it into:
- `createChildProviderFactory({ openaiBaseUrl })`
- the trailing positional arg of `createChildSkillExecutorFactory(...)`
- `SubagentExecutor.defaultConfig`
- `SkillExecutor`

**Codex/OpenAI branch** — capture `codexOpenaiBaseUrl` and set it on the `constructTelegramSession` config so `config.openaiBaseUrl → effectiveBaseURL` (`openai-compatible/index.ts:264`), fixing the top-level session's endpoint.

`ComposeExecutor` is intentionally untouched — its options type has no `openaiBaseUrl` field (same as `chat.ts`).

## Testing

- `pnpm lint` (tsc --noEmit, strict) — clean.
- `pnpm exec vitest run src/telegram/construct-session.test.ts src/agent/providers/routing.test.ts` — 125/125 pass.
- Added: `construct-session` `openaiBaseUrl` pass-through test; `routing.test.ts` coverage that `createChildSkillExecutorFactory` forwards `openaiBaseUrl` to nested `SkillExecutor`s.

## Known follow-up (out of scope)

The OpenAI/Codex Telegram branch still builds **no** child executors (Sub/Skill/Compose) — a separate structural gap, and the reason a unified `assembleSurfaceExecutors` factory across all four surfaces is not a mechanical lift. This PR fixes only the endpoint routing. Context: surface-composition audit (2026-07-06).

### DIFF
--- a/src/agent/providers/routing.test.ts
+++ b/src/agent/providers/routing.test.ts
@@ -23,12 +23,17 @@
 
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { providerForModel, resolveProvider } from './index.js';
-import { createChildProviderFactory } from '../tools/nesting.js';
+import {
+  createChildProviderFactory,
+  createChildSkillExecutorFactory,
+  type ChildProviderFactoryArgs,
+} from '../tools/nesting.js';
 import { AnthropicDirectProvider } from './anthropic-direct/index.js';
 import { OpenAICompatibleProvider } from './openai-compatible/index.js';
 import { getDefaultSubagentModel } from '../../cli/shared-helpers.js';
 import type { SubagentExecutor } from '../tools/subagent-executor.js';
 import type { SkillExecutor } from '../tools/skill-executor.js';
+import type { ModelProvider } from '../provider.js';
 
 describe('providerForModel', () => {
   // Scrub env vars that `providerForModel` now consults internally.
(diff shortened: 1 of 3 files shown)


griffinwork40:
---

## PR #459 Review: `fix(telegram): thread openaiBaseUrl so OpenAI Telegram sessions reach the configured endpoint`

**Verdict: ✅ Approve** — Clean, well-scoped bugfix with solid test coverage. One observation worth noting for follow-up.

---

### Correctness ✅

The fix addresses a real gap: the Telegram surface was the only bootstrap path that didn't propagate `openaiBaseUrl` to child providers, skill executors, and the top-level session config. When an operator ran the Telegram bot with an OpenAI-routed model + a local shim (`AFK_OPENAI_BASE_URL`), children silently hit `api.openai.com` instead of the configured endpoint.

The wiring is correct at every handoff:

| Site | What the PR does | Verified |
|------|-----------------|----------|
| `createChildProviderFactory({ openaiBaseUrl })` | Forwards to `OpenAICompatibleProvider.baseURL` for OpenAI-routed children | ✅ matches `nesting.ts:187` |
| `createChildSkillExecutorFactory(..., openaiBaseUrl)` | 12th positional arg, closes over into recursive factory | ✅ matches `nesting.ts:287,316` |
| `SubagentExecutor.defaultConfig.openaiBaseUrl` | Flows into `buildForkedChildConfig` for grandchild sessions | ✅ matches `subagent-executor.ts:51`, `child-config.ts:62` |
| `SkillExecutor` constructor `openaiBaseUrl` | Used by `buildReadOnlyReconProvider` + `buildSkillRestrictedProvider` at depth cap | ✅ matches `skill-executor.ts:611,619,647` |
| Codex branch `constructTelegramSession({ openaiBaseUrl })` | Reaches `effectiveBaseURL` in `openai-compatible/index.ts:264` | ✅ |

Both the Anthropic and Codex branches are covered. The resolution (`sessionConfig.openaiBaseUrl ?? config.openaiBaseUrl`) matches the existing `baseUrl` pattern and the chat.ts wiring.

### Tests ✅

- **routing.test.ts** — 3 new tests covering: forward, recursive propagation (skill→skill→skill), and back-compat (undefined stays undefined). The recursive test reaches into `ctx.childSkillExecutorFactory` to prove the closure captures the value across depths — good coverage of the nesting chain.
- **construct-session.test.ts** — 1 new test pinning `openaiBaseUrl` pass-through at the `constructTelegramSession` seam.
- All 11,100 existing tests pass. `tsc --noEmit` clean.

### Observation: ComposeExecutor gap (pre-existing, not this PR)

`ComposeExecutor` does NOT accept or propagate `openaiBaseUrl` — neither in chat.ts nor in this PR's telegram.ts changes. Compose DAG nodes targeting OpenAI models would hit the same wrong-endpoint bug. This is a pre-existing gap, symmetric across both surfaces, and correctly out of scope for this fix. Worth a follow-up issue.

### Minor notes (non-blocking)

1. **Positional parameter count**: `createChildSkillExecutorFactory` now takes 12 positional parameters. This PR doesn't make it worse (just adds one at the tail), but the function is approaching the threshold where an options object would be safer. Not a reason to hold this PR.

2. **Duplicate resolution**: `telegramOpenaiBaseUrl` (Anthropic branch) and `codexOpenaiBaseUrl` (Codex branch) compute the identical expression. This is intentional for readability — each branch is self-contained with its own local variable. Fine as-is.

3. **Test type assertions**: The routing tests use `as unknown as { ctx: ... }` casts to inspect internal state. This is the standard pattern for this codebase's nesting tests (mirrors `nesting.model-fallback.test.ts`). Acceptable tradeoff for testing internal propagation.

### Summary

Tight, well-commented fix that closes a real configuration leak. The chat.ts/telegram.ts parity is now restored for `openaiBaseUrl`. Tests cover the critical paths (forward, recursive, back-compat). No correctness issues found.

**Recommendation: Merge.**

---

PR #461, opened 2026-07-07
fix(subagent): surface capped/truncated partials to the parent instead of silent success

griffinwork40:
## Problem

A subagent run that either (a) hits the tool-use loop cap or (b) ends mid-stream — an abort, or an abnormal/early provider-stream close — resolves with `status: 'succeeded'` while carrying only **partial** assistant content. The `stopReason` plumbing to detect this already existed (`SubagentResult.stopReason`, `handle.ts` `lastStopReason`), but **no consumer read it**, so every parent-facing surface handed the partial to the parent model as a clean, complete answer.

Observed across witness traces: `succeeded` results with tiny output (58–142 B), including `turnCount: 0` runs and one that ran ~2.1 h before returning 99 B — all surfaced as success.

Two triggers:
- **Cap:** the tool-use wind-down can produce a ~142-byte "capped" marker returned as success.
- **Stream truncation:** a mid-stream abort / early clean close yields accumulated text but no final `message` event, and returns as success.

## Fix

**Producer** (`handle.ts`): the stream-truncated path now sets a synthetic `STREAM_INCOMPLETE` stop reason, and the terminal stop reason is recorded on the `subagent.succeeded` trace event (`trace/events.ts`, `trace/types.ts`) for forensics.

**Consumer** (`result.ts`): new `annotateIfIncomplete(content, stopReason)` prepends a parent-visible marker —
`[⚠ PARTIAL RESULT — the subagent … an incomplete intermediate finding, NOT a final answer …]` — and is a **no-op for clean completions**. Wired into all four parent-facing surfaces:

- skill tool — `skill-executor.ts`
- foreground agent — `foreground-promotion.ts`
- compose DAG node — `dag-subagent.ts` (**text branch only**, never structured `output`)
- background delivery — `bg-result-notifier.ts`

Internal skill-phase parsers (`mint` / `diagnose` / `heal`) are **deliberately not** annotated — a prepended banner would corrupt their content parsing.

## Tests / verification
- New unit tests for `isIncompleteStopReason` / `annotateIfIncomplete` (`result.test.ts`, +9).
- `tsc --noEmit` clean (strict config).
- Full suite green (11,109 passed) on current `main`.

## Scope
Addresses the **partial-as-success (correctness)** defect only. A separate, independent observability bug — lost terminal trace events producing "orphan" subagent records (fire-and-forget `emitSubagentLifecycle` racing an awaited `seal()`) — is **not** in this PR; it is the follow-up **stacked** PR.

PR #462, opened 2026-07-07
fix(subagent): await terminal lifecycle emits so trace records survive an immediate seal

griffinwork40:
> **Stacked on #461** — base branch is `afk/subagent-partial-as-success`, so this PR's diff shows only the M1 change. Please review/merge #461 first; once it lands on `main`, GitHub auto-retargets this PR to `main`.

## Problem (M1 — lost terminal trace event)

Subagent terminal lifecycle events are silently dropped from the witness trace, leaving "orphan" records (a `subagent_lifecycle` `started` with no matching terminal event). Measured ~320/1166 subagents, mostly short-lived skill sessions. **The subagents complete fine — only the audit record is lost** (observability, not correctness).

Root cause: the terminal emits in `SubagentHandleImpl.run()` were fire-and-forget (`void emitSubagentLifecycle(...)`). `emitSubagentLifecycle` awaits `writer.write()` internally, but since the emit itself wasn't awaited, `run()` proceeded to `onTerminal()` — which can trigger the owning session's immediate `writer.seal()` — before the terminal write reached the writer. `NdjsonTraceWriter.seal()` flips `sealed = true` synchronously and a late `write()` throws `'trace is sealed'`, which `emitSubagentLifecycle` swallows → terminal event lost.

## Fix

`await` the terminal emits in `run()` (succeeded + cancelled + failed paths) so the event is enqueued+persisted on the writer's FIFO queue **before** `onTerminal()` can seal.

Note: draining the queue inside `seal()` would **not** fix this — the racing write isn't enqueued yet at seal time, so it isn't on the queue to drain. The fix must ensure the terminal write is enqueued before seal. `writer.ts` and `emit.ts` are unchanged. The `cancel()`-path emit (explicit teardown) is intentionally left as-is — its ordering contract vs. the abort cascade needs separate analysis and it isn't a measured orphan source.

## Test / verification
- New `subagent-terminal-seal-race.test.ts`: a `DeferringTraceWriter` proxy reproduces the production timing gap deterministically (defers each `write()` to a macrotask; keeps `seal()` prompt). **Revert-checked: FAILS on the old `void` code, PASSES on the fix.**
- `tsc --noEmit` clean; `src/agent/subagent` + `src/agent/trace` suites green (374 passed).

## Scope
The observability half of the partial-subagent-results investigation. The correctness half (partials returned as success) is #461.

### DIFF
--- a/src/agent/subagent/handle.ts
+++ b/src/agent/subagent/handle.ts
@@ -194,10 +194,16 @@ export class SubagentHandleImpl<T> implements SubagentHandle<T> {
       this.lastDurationMs = Date.now() - startTime;
       this.currentStatus = 'succeeded';
       this.latestTerminalStatus = 'succeeded';
-      // Witness layer: subagent_lifecycle.succeeded fires before onTerminal()
-      // so the trace records the terminal transition even if the manager
-      // tears the handle down immediately after. Fire-and-forget.
-      void emitSubagentLifecycle(this.traceWriter, {
+      // Witness layer: subagent_lifecycle.succeeded MUST be awaited before
+      // onTerminal(). onTerminal() may trigger the owning session's immediate
+      // teardown, which calls writer.seal(); once sealed, writer.write() throws
+      // and emitSubagentLifecycle swallows it, silently dropping this terminal
+      // record (the "lost terminal trace event" orphan bug). Awaiting here
+      // guarantees the succeeded event is enqueued+persisted on the writer's
+      // FIFO queue BEFORE any seal can run. Safe: write() is a bounded FS append
+      // and emitSubagentLifecycle already swallows errors, so the await cannot
+      // introduce a new failure mode or an unbounded hang.
+      await emitSubagentLifecycle(this.traceWriter, {
         transition: 'succeeded',
         subagentId: this.id,
         durationMs: this.lastDurationMs,
@@ -239,8 +245,12 @@ export class SubagentHandleImpl<T> implements SubagentHandle<T> {
         // externally. The trace and the result-object status must agree so
         // downstream consumers (operator dashboard, future ActiveWorkRegistry)
         // can correctly attribute cascade terminations vs. genuine failures.
+        // Awaited (not fire-and-forget) for the same reason as the success
+        // path: onTerminal() below may seal the owning session's trace, and a
+        // seal that lands before this write is enqueued would drop the terminal
+        // record. Awaiting guarantees the event is persisted first.
         if (this.controller.signal.aborted) {
-          void emitSubagentLifecycle(this.traceWriter, {
+          await emitSubagentLifecycle(this.traceWriter, {
             transition: 'cancelled',
             subagentId: this.id,
             source: 'cascade',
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

## Review — PR #462

**fix(subagent): await terminal lifecycle emits so trace records survive an immediate seal**
`afk/subagent-lost-terminal-race` → `afk/subagent-partial-as-success` · +230/-6 · 2 files
Reviewed ref: `7da6c10d` · Stated intent: PR title+body — spec-compliance assessed
Regime: `light` (≤300 lines, hotfix)

---

### Summary

The fix is correct, narrowly scoped, and well-documented. It addresses a confirmed data-loss bug: terminal `subagent_lifecycle` events (`succeeded`/`failed`/`cancelled`-cascade) were silently dropped when `onTerminal()` triggered a trace `seal()` before the fire-and-forget `write()` reached the writer. The root-cause analysis is accurate — `NdjsonTraceWriter.seal()` flips `sealed = true` synchronously, and a subsequent `write()` throws `'trace is sealed'`, which `emitSubagentLifecycle` swallows. Changing `void` → `await` on the three terminal emits in `run()` guarantees the event is enqueued on the FIFO queue before `onTerminal()` can fire.

The regression test uses a `DeferringTraceWriter` proxy that defers writes to a macrotask while passing `seal()` through synchronously — this deterministically reproduces the production timing gap. The test correctly fails on the old `void` code and passes on the fixed `await` code.

---

### Findings

| # | Severity | Dimension | File | Summary |
|---|----------|-----------|------|---------|
| 1 | low | test-coverage | `subagent-terminal-seal-race.test.ts` | Only the `succeeded` path is regression-tested; the `failed` and `cancelled`-cascade paths in `run()`'s catch block (handle.ts:248–265) are also changed from `void` → `await` but have no corresponding DeferringTraceWriter test |
| 2 | low | correctness | `handle.ts:486` | `cancel()` path retains `void emitSubagentLifecycle(...)` for `source: 'explicit'` — same theoretical race if the owning session seals promptly after `cancel()` resolves (mitigated by multiple `await` points between emit and `onTerminal()`, making the window much wider) |

---

### What was checked
- **Security**: no issues — change touches only emit ordering in the witness layer; no auth/input/secret surface affected; `emitSubagentLifecycle` already swallows errors so `await` cannot introduce a new failure mode
- **API compat**: no issues — `run()` is already `async`; all 18+ production callers of `runToResult()` already `await` the result; added latency is one bounded FS append (<1ms)
- **Correctness**: fix mechanism verified — `seal()` sets `sealed = true` synchronously at `writer.ts:195`; `write()` checks `if (this.sealed) throw` at `writer.ts:186`; `emitSubagentLifecycle` catches all errors at `emit.ts:71-76`; with `await`, the write completes before `onTerminal()` fires
- **Spec-compliance**: diff matches stated intent exactly — three `void` → `await` changes in `run()`'s terminal paths; `writer.ts` and `emit.ts` unchanged; `cancel()` path intentionally left as-is per PR body
- **Test coverage**: DeferringTraceWriter faithfully reproduces the race; test asserts both presence and ordering of the terminal event; revert-checked (FAILS on old code, PASSES on fix)
- **Perf/observability**: negligible latency (one small NDJSON append per terminal path); strictly positive observability impact (eliminates ~320/1166 orphan rate)

### What was not checked
- Read against branch HEAD `7da6c10d` — all citations verified at that ref.
- Did not trace the production `onTerminal()` callback implementation to confirm it calls `writer.seal()` synchronously (fix is correct regardless).
- Stated intent: PR title+body — spec-compliance assessed.

---

### Merge decision

**MERGE**

Correct fix for a real observability data-loss bug, narrowly scoped to three `void` → `await` conversions with zero risk of new failure modes. The two low-severity findings (test coverage gap for catch paths, cancel() void emit asymmetry) are follow-up items, not blockers.

PR #449, opened 2026-07-06
fix(permissions): dedup session-grants.jsonl audit — append only on grant state change

griffinwork40:
## What
`session-grants.jsonl` (the path-permission audit ledger under `~/.afk/state/`) had grown to **46 MB / 224,049 rows from only 1,143 unique `(action, path)` grants — a 196× duplication**, climbing ~1.6 MB/day.

## Root cause
`addReadRoot` / `addWriteRoot` in all three grant managers — `SessionToolDispatcher` (`dispatcher.ts`), `AnthropicDirectProvider`, and the OpenAI-compatible provider — called `appendAuditLog` / `appendProviderAuditLog` **unconditionally**, *outside* the `if (!roots.includes(p)) push` dedup guard. Every repeat grant of an already-granted path wrote a duplicate ledger record:
- `path-approval-hook` calls `addReadRoot`/`addWriteRoot` on ~every tool call that touches a granted path (e.g. the cwd);
- `seedPersistedGrants` re-seeds all persisted grants at every bootstrap.

**Duplication signature** confirming the mechanism (not seed-bursts): the top key is `grant-read` on the cwd — **2,404 records across 2,402 _distinct_ timestamps** (spread over time; max identical-timestamp run = 2), 100% `sessionId=null`, 99.6% `source=tool`.

## Fix
Move the audit append **inside** the "newly added" guard in all three managers, so a grant is recorded once, on genuine state change. A read→write upgrade still audits (new to `writeRoots`); a repeat grant is silent. `revokeRoot` is unchanged. Also corrected the `seedPersistedGrants` docstring (re-seeding is now audit-idempotent within a live process).

## Tests
- `pnpm lint` (tsc --noEmit, strict): **clean**.
- All grant / permission / provider / telegram-wiring suites pass; **+3 regression tests** in `dispatcher-audit-log.test.ts`: repeat read grant → 1 record; repeat write grant → 1 record; read→write upgrade → `[grant-read, grant-write]`.

## Note — pre-existing failures (NOT from this PR)
4 `src/cli/commands/chat.*.test.ts` suites fail on `origin/main` independently of this change (incomplete `vi.mock('../../agent/tools/schemas.js')` missing the `builtinToolSchemas` export). Verified by stashing this PR's diff and re-running on a clean tree — the failure is identical. Flagging as a separate main-branch test-mock bug worth its own fix.

## Out of scope (operational follow-up)
[...]

PR #473, opened 2026-07-09
fix(agent): cap session-grants.jsonl growth with atomic size trim

griffinwork40:
## What & why

`~/.afk/state/session-grants.jsonl` — a **write-only** permission-grant audit log — had grown to ~46 MB / 235k lines. Root cause was the since-fixed 196x duplication bug from #449 (1,143 unique grants had ballooned to 224,049 rows). That write-path fix stopped new duplication but left the historical bloat, and nothing bounds the file's size.

This adds a best-effort, atomic size-cap that keeps the newest N lines, run once per top-level session at bootstrap.

## Approach

- **New `src/agent/log-retention.ts`** — `capJsonlBySize(path, { maxBytes, keepTailLines })`: when the file exceeds `maxBytes`, atomically (temp + rename) rewrite it keeping the newest `keepTailLines` lines. Best-effort — never throws (missing file, permission error, ENOSPC, or race is swallowed). Reusable for a future class-wide log sweep.
- **Wired into `AgentSession`** (`src/agent/session/agent-session.ts`) — one fire-and-forget, silent-fail call in the constructor, **top-level sessions only** (`parentSessionId === undefined`), on the async path so it never blocks the synchronous constructor. Covers REPL + Telegram + daemon.
- **Hardcoded constants** (`SESSION_GRANTS_MAX_BYTES = 5 MB`, `KEEP_TAIL_LINES = 5000`) — mirrors the `src/cli/input/history.ts` `MAX_ENTRIES` precedent; no env var, no `ENV_REGISTRY` / `scan:env` surface.

### Why bootstrap, not inline at the write sites

`session-grants.jsonl` is appended with `O_APPEND` by multiple concurrent processes (REPL / daemon / Telegram) and by in-process subagent forks. An inline read-then-rename size-cap would race those appenders and silently drop lines. Running once at top-level bootstrap bounds that window to the rare two-simultaneous-boots case — acceptable for a write-only audit log (no production code reads it), and documented as an `// Invariant:` in the module.

### Why keep a tail (not truncate to empty)

The file is a forensic audit trail. Even though no code reads it, preserving the newest 5,000 records keeps its manual / support-review value.

## Verification

- `pnpm lint` (tsc --noEmit, strict): **exit 0**.
- `src/agent/log-retention.test.ts`: **9/9 pass** — under-cap no-op, over-cap keeps exactly N newest, missing-file no-op, malformed-line tolerance, never-throws, no-trailing-newline handling, no temp file left behind.
- Full suite: **598 files / 11,173 tests pass, 0 fail**.
- End-to-end on the real file via the actual code path: **48,762,199 -> 1,028,064 bytes** (~47.7 MB reclaimed), 230,284 duplicate lines dropped, newest audit record byte-identical before/after (tail preserved).

## Deferred (follow-up)

A class-wide `__BUILTIN_LOG_ROTATE__` scheduled sweep governing the other ~10 unbounded logs with per-file policy (size-cap write-only logs; age-based keep-recent for read-back logs like `trace.jsonl` / `events.jsonl`). `capJsonlBySize` is the building block for its write-only entries.

### DIFF
--- a/src/agent/log-retention.test.ts
+++ b/src/agent/log-retention.test.ts
@@ -1,0 +1,116 @@
+/**
+ * Tests for capJsonlBySize — best-effort size-cap for append-only JSONL logs.
+ *
+ * Disk tests use a fresh temp dir so they never touch real state files.
+ */
+
+import { describe, expect, it, beforeEach, afterEach } from 'vitest';
+import { mkdtemp, writeFile, readFile, rm, readdir } from 'fs/promises';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import {
+  capJsonlBySize,
+  SESSION_GRANTS_MAX_BYTES,
+  SESSION_GRANTS_KEEP_TAIL_LINES,
+} from './log-retention.js';
+
+let dir: string;
+let file: string;
+
+beforeEach(async () => {
+  dir = await mkdtemp(join(tmpdir(), 'afk-log-retention-'));
+  file = join(dir, 'grants.jsonl');
+});
+
+afterEach(async () => {
+  await rm(dir, { recursive: true, force: true });
+});
+
+/** Build `n` newline-terminated JSONL records `{"i":<idx>}`. */
+function makeLines(n: number): string {
+  const rows: string[] = [];
+  for (let i = 0; i < n; i++) rows.push(JSON.stringify({ i }));
+  return rows.join('\n') + '\n';
+}
+
+function nonEmptyLines(raw: string): string[] {
+  return raw.split('\n').filter((l) => l.length > 0);
+}
+
+describe('capJsonlBySize', () => {
+  it('is a no-op when the file is under maxBytes', async () => {
+    const content = makeLines(10);
+    await writeFile(file, content);
+    const res = await capJsonlBySize(file, { maxBytes: 1_000_000, keepTailLines: 5 });
+    expect(res).toEqual({ trimmed: false, removedLines: 0 });
+    expect(await readFile(file, 'utf8')).toBe(content);
+  });
+
+  it('is a no-op (never throws) when the file does not exist', async () => {
+    const res = await capJsonlBySize(join(dir, 'missing.jsonl'), {
+      maxBytes: 10,
+      keepTailLines: 5,
+    });
+    expect(res).toEqual({ trimmed: false, removedLines: 0 });
+  });
+
+  it('trims to exactly keepTailLines newest lines when over maxBytes', async () => {
+    await writeFile(file, makeLines(1000));
+    const res = await capJsonlBySize(file, { maxBytes: 100, keepTailLines: 100 });
+    expect(res.trimmed).toBe(true);
+    expect(res.removedLines).toBe(900);
+
+    const out = await readFile(file, 'utf8');
+    const lines = nonEmptyLines(out);
+    expect(lines.length).toBe(100);
+    // Newest kept: first retained record is i=900, last is i=999.
+    expect(JSON.parse(lines[0]!).i).toBe(900);
+    expect(JSON.parse(lines[lines.length - 1]!).i).toBe(999);
+    // Exactly one trailing newline (no doubled or missing terminator).
+    expect(out.endsWith('\n')).toBe(true);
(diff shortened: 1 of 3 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

A focused, well-tested hotfix. `capJsonlBySize`'s algorithm was hand-traced against all 9 unit tests and matches; the three new exports have **zero production importers** (no API break); the temp-write + rename is same-directory (atomic) and matches ~6 existing codebase conventions (`permissions-store.ts`, `envFile.ts`, `bg-job-log.ts`, …); the documented `O_APPEND` race is correctly characterized and bounded by the top-level-only + bootstrap-only gating. No security, correctness, or data-loss defect found. All findings below are non-blocking polish.

*(All citations verified against reviewed ref `a1d249c`.)*

**1. `low` · security — temp-file mode tightens permissions vs. the append sites** (`log-retention.ts:27`)
The trim writes the replacement with `{ mode: 0o600 }`, but the three production writers of `session-grants.jsonl` (`dispatcher.ts:506`, `anthropic-direct/index.ts:624`, `openai-compatible/index.ts:588`) call bare `appendFileSync(logPath, line + '\n')` with no mode. The file's permissions silently tighten from the process-umask default to owner-only the first time it is trimmed. This is a *tightening* (not a loosening) and there are zero in-repo readers, so realistic impact is confined to hypothetical external consumers.
*Suggestion:* pass `{ mode: 0o600 }` at the append sites too (or drop it from the trim write) so permission posture doesn't depend on whether the file has crossed the size threshold.

**2. `low` · test-coverage — the behavioral delta (the constructor gate) has no test** (`agent-session.ts:243-256`)
`git grep` for `capJsonlBySize|log-retention|SESSION_GRANTS` in `agent-session.test.ts` / `agent-session-wrappers.test.ts` at the ref returns zero hits. The `parentSessionId === undefined` gate that actually decides whether the trim fires is exercised only indirectly; only the standalone helper is unit-tested.
*Suggestion:* add one test asserting a top-level session invokes `capJsonlBySize` with the grants path, and one asserting a subagent-forked session (`parentSessionId` set) does not.

**3. `low` · perf-observability — a multi-MB trim is completely silent** (`agent-session.ts:249-253`)
The fire-and-forget `void` call discards the `{ trimmed, removedLines }` result, so a one-time rewrite of a tens-of-MB file (the PR's own e2e: 48.7MB → 1.0MB) produces no observable output.
*Suggestion:* log a single line on `trimmed: true` (removed-line count + elapsed ms). Orthogonal to the intentional silent-fail-on-error, which should stay.

**4. `low` · spec-compliance — docblock invariant describes intra-session, not cross-session, serialization** (`log-retention.ts:6-10`)
"Once per top-level session at bootstrap" bounds one session's lifecycle, not two concurrent top-level sessions each running their own read-then-rename against an over-threshold file. This is the same accepted loss class the docblock already discloses for appenders — no new corruption path (`appendFileSync` opens/writes/closes per call, so no orphaned-inode risk).
*Suggestion:* extend the docblock's risk statement one clause to name concurrent size-cap runs (not only concurrent appenders) as part of the accepted window.

**5. `nit` · test-coverage — no test exercises the documented concurrent-append race** (`log-retention.ts:6-10`)
All 9 tests run against a static, uncontended file. The drop-one-line-on-race behavior lives only as a comment assertion.

**6. `nit` · correctness — confirmed clean (no action)**
- `void capJsonlBySize(...)` cannot produce an unhandled rejection: the whole body runs inside one `try/catch` and every `fs/promises` call is awaited within it.
- Atomicity is sound: same-directory temp + rename with unlink-on-failure cleanup.

### What was not checked
- Read exclusively against reviewed ref `a1d249ca8c58d1b172a0aa2d95cec0a0302bbf76` (`git show`/`git grep`); the local working tree was on a different dirty branch and not used as source of truth.
- Tests were **not** run (static read-only review). The PR claims `pnpm lint` exit 0 and full suite pass — not independently re-verified here.
- Not traced: whether `AgentSession`'s constructor can run more than once per logical top-level session (reconnect/retry). If it can, a re-run is a harmless idempotent no-op once under threshold — does not change the verdict.
- Stated intent: PR #473 title + body — spec-compliance **assessed**.

PR #474, opened 2026-07-10
feat(models): add GPT-5.6 family support to openai-compatible provider

griffinwork40:
## What & why

OpenAI released the **GPT-5.6** model family on 2026-07-09 (`gpt-5.6` alias → `gpt-5.6-sol`; variants `gpt-5.6-sol` / `gpt-5.6-terra` / `gpt-5.6-luna`). This makes GPT-5.6 a **first-class, discoverable model** in AFK's `openai-compatible` provider.

Users reach it via **Codex auth** (`~/.codex/auth.json`) — either API-key mode (`source: codex-cli`) or ChatGPT-subscription OAuth behind `AFK_OPENAI_CHATGPT_OAUTH=1` — or a plain `OPENAI_API_KEY`. All of that auth plumbing already exists (`src/agent/providers/openai-compatible/auth.ts:145-171`).

## Key finding: routing already worked

`gpt-5.6` already routed correctly and got the right request contract *before* this PR, because the engine matches families by pattern, not by an enumerated allowlist:
- `providerForModel` routes `gpt-*` → `openai-compatible` (`providers/index.ts`)
- `REASONING_MODEL_PATTERNS = [/^gpt-5/]` → reasoning request shape (`model-capabilities.ts`)
- vision `/^gpt-5/i` → multimodal

So there is **no routing or request-shape code change here.** The gaps were the *maintained lookup tables* that make a model first-class rather than silently-defaulted:

## Changes

- **`model-limits.ts`** — explicit **1M** `MODEL_CONTEXT_LIMITS` entries for the `gpt-5.6` alias + `-sol`/`-terra`/`-luna` (OpenAI's flagship 5.x window is ~1.05M; rounded to 1M to match the existing `gpt-4.1`/`gpt-5.5` convention). Without this, `getContextUsage()` would fall through to the 262k `openai-compatible` default and silently allow context overruns. Also adds **`gpt-5.5`**, which had the same latent gap.
- **`query.ts`** — lists the GPT-5.6 family in the `/model` picker; updates the stale ChatGPT/Codex-backend `400` diagnostic (`gpt-5.5 works` → `gpt-5.6 and gpt-5.5 work`).
- **Regression tests** — `routing.test.ts` (→ openai-compatible), `model-capabilities.test.ts` (reasoning + vision), `model-limits.test.ts` (explicit 1M window for every id).
- **CHANGELOG** — `[Unreleased] › Added`.

## Verification

- `pnpm lint` (tsc --noEmit) ✅
- `pnpm test` — **597 files, 11,170 passed, 14 skipped (live), 0 failed** ✅
- `pnpm test:coverage` — coverage gate ✅ (no threshold violations)
- `pnpm build`, `audit:env:check`, `scan:env:check`, `audit:sdk:check` ✅
- **Runtime check against compiled dist** — all four `gpt-5.6*` ids: `provider=openai-compatible`, `ctx=1,000,000`, `reasoning=true`, `vision=true` ✅

## Notes / out of scope

- Max output (128K) is not added to `MODEL_MAP`/`MODEL_MAX_OUTPUT_TOKENS` — that table is Claude-alias-only; OpenAI ids don't need entries there.
- No new provider or bridge was added; this rides the existing `openai-compatible` HTTP client + Codex-auth reader.

+80 / -4 across 6 files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,16 @@ auto-release workflow to deduplicate commits across successive runs.
 ### Added
 - GPT-5.6 model family support for the openai-compatible provider (reachable via
   Codex ChatGPT-OAuth or an OpenAI API key): explicit 1M context-window entries
-  for the `gpt-5.6` alias and the `gpt-5.6-sol`/`-terra`/`-luna` variants (plus
-  `gpt-5.5`, previously falling through to the 262k default), `/model` picker
-  listings, and an updated ChatGPT/Codex-backend 400 diagnostic. Routing and the
-  reasoning/vision request contract already covered `gpt-5.6` via the `/^gpt-5/`
-  patterns; this closes the maintained-table gaps and adds regression coverage.
+  (`MODEL_CONTEXT_LIMITS`) and 128k max-output-token entries
+  (`MODEL_MAX_OUTPUT_TOKENS`) for the `gpt-5.6` alias and the
+  `gpt-5.6-sol`/`-terra`/`-luna` variants (plus `gpt-5.5`, previously falling
+  through to the 262k / 64k defaults), `/model` picker listings, and an updated
+  ChatGPT/Codex-backend 400 diagnostic. Without the output-cap entries these ids
+  hit `DEFAULT_MAX_OUTPUT` (64k) and silently halved their advertised 128k output
+  budget when `config.maxOutputTokens` was unset, truncating long code/research
+  responses. Routing and the reasoning/vision request contract already covered
+  `gpt-5.6` via the `/^gpt-5/` patterns; this closes the maintained-table gaps
+  and adds regression coverage.
 
 ## [5.25.10] - 2026-07-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@ auto-release workflow to deduplicate commits across successive runs.
 
 ## [Unreleased]
 
+### Added
+- GPT-5.6 model family support for the openai-compatible provider (reachable via
+  Codex ChatGPT-OAuth or an OpenAI API key): explicit 1M context-window entries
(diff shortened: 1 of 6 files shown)


griffinwork40:
Good catch — the first commit only touched `MODEL_CONTEXT_LIMITS` and missed its sibling `MODEL_MAX_OUTPUT_TOKENS`. Confirmed the bug: `maxOutputTokensFor()` is provider-agnostic, and the openai-compatible request path (`query/model-params.ts:resolveEffectiveMaxOutputTokens` → `resolveStreamingMaxTokens`) calls it to bound output when `config.maxOutputTokens` is unset. Without matching entries, `gpt-5.6`/`-sol`/`-terra`/`-luna` (and `gpt-5.5`) fell through to `DEFAULT_MAX_OUTPUT` (64k) — silently halving the advertised 128k output ceiling.

Fixed in `2c4a0b7`:
- Added 128k `MODEL_MAX_OUTPUT_TOKENS` entries for the gpt-5.6 family + gpt-5.5 (128k max output per OpenAI API docs for both the gpt-5.6 tiers and gpt-5.5).
- Regression tests assert `maxOutputTokensFor()` **and** the actual openai-compatible resolver return 128k with config unset, an explicit-override path, and a control proving unknown ids still hit the 64k default.

Runtime check against compiled dist (the exact buggy path):
```
gpt-5.6       effective=128000 | wire={"max_completion_tokens":128000}
gpt-5.6-sol   effective=128000 | wire={"max_completion_tokens":128000}
gpt-5.6-terra effective=128000 | wire={"max_completion_tokens":128000}
gpt-5.6-luna  effective=128000 | wire={"max_completion_tokens":128000}
gpt-5.5       effective=128000 | wire={"max_completion_tokens":128000}
gpt-9-mystery effective=64000  (control: unknown → default)
```
Full suite green (597 files, 11,173 passed); lint + env/sdk audits + build all pass.

PR #475, opened 2026-07-10
fix(telegram): disable Telegraf 90s handlerTimeout so long turns aren't guillotined

griffinwork40:
## Problem

Long agent turns on Telegram surface a spurious **"❌ An unexpected error occurred. Please try again."** — while the turn is *still running in the background* and often completes (in the case that surfaced this, it implemented + committed + opened a PR *after* the error was shown). It recurs constantly in the service log (dozens of occurrences).

## Root cause

`src/telegram/bot.ts` constructs `new Telegraf(options.botToken)` with no options, so Telegraf applies its **default `handlerTimeout: 90_000`** (`telegraf/lib/telegraf.js:46`) and wraps the *entire* update handler — a full agent turn — in `p-timeout(..., 90_000)` (`telegraf.js:233`). Any turn over 90s total wall-clock (sub-agents, `web_scrape`, long bash) rejects to the global `bot.catch` → the generic error.

Because `p-timeout` does **not** abort the underlying promise, the `AgentSession` keeps running detached — which is exactly why the turn finishes (and can ship a PR) *after* the user already saw the failure.

Meanwhile the app already has the *correct* timeout in `src/telegram/streaming.ts`: an inter-event inactivity watchdog (90s first event / 180s between) that is **tool-in-flight aware** and **usage-limit-pause aware**, and throws a *handled* `StreamTimeoutError` with an honest message. The two timeouts collide and the blunt one wins.

## Fix

```ts
this.bot = new Telegraf(options.botToken, { handlerTimeout: Infinity });
```

`p-timeout@4.1.0` short-circuits on `Infinity` (`if (milliseconds === Infinity) { resolve(promise); }` — no timer armed), so this hands **sole** timeout authority to the `streaming.ts` watchdog, which is strictly better than a blunt 90s wall-clock cap.

## Verification

- `pnpm test` — full suite green (597 files / 11,165 tests)
- New regression test in `bot.test.ts` asserts the bot is constructed with `handlerTimeout === Infinity` (and not the `90_000` default)

## Deploy note

The running Telegram **service must be rebuilt + restarted** (`pnpm build`, then restart the launchd service) to pick this up — the source fix alone does not touch the live `dist/telegram.mjs`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ auto-release workflow to deduplicate commits across successive runs.
 
 ## [Unreleased]
 
+### Fixed
+- Telegram: don't guillotine long agent turns at 90s — disable Telegraf's default `handlerTimeout` so the purpose-built `streaming.ts` inactivity watchdog is the sole timeout authority; fixes the spurious "❌ An unexpected error occurred. Please try again." shown while a turn (sub-agents, web_scrape, long tool calls) was still running and would complete
+
 ## [5.25.10] - 2026-07-09
 
 ### Fixed
--- a/src/telegram/bot.test.ts
+++ b/src/telegram/bot.test.ts
@@ -112,6 +112,19 @@ describe('TelegramBot', () => {
     }
   });
 
+  describe('construction', () => {
+    test('disables Telegraf handlerTimeout so long turns are governed by the streaming watchdog, not a 90s guillotine', () => {
+      // Regression: Telegraf's default handlerTimeout (90_000ms) p-timeouts the
+      // whole turn → bot.catch → generic error, while the AgentSession keeps
+      // running in the background. streaming.ts owns timeout policy instead.
+      const handlerTimeout = (
+        bot as unknown as { bot: { options: { handlerTimeout: number } } }
+      ).bot.options.handlerTimeout;
+      expect(handlerTimeout).toBe(Infinity);
+      expect(handlerTimeout).not.toBe(90_000);
+    });
+  });
+
   describe('commands', () => {
     test('should handle /start command', async () => {
       const ctx = createMockContext(12345, '/start');
--- a/src/telegram/bot.ts
+++ b/src/telegram/bot.ts
@@ -63,7 +63,15 @@ export class TelegramBot {
 
   constructor(options: BotOptions) {
     this.options = options;
-    this.bot = new Telegraf(options.botToken);
+    // Disable Telegraf's default handlerTimeout (90_000ms). It wraps the ENTIRE
+    // update handler — a full agent turn — in a p-timeout, so any turn over 90s
+    // total (sub-agents, web_scrape, long bash) rejects to `bot.catch` as a
+    // generic "unexpected error" even though the AgentSession keeps running
+    // (p-timeout does not abort the underlying promise). src/telegram/streaming.ts
+    // is the real timeout authority: an inter-event, tool-in-flight-aware,
+    // usage-limit-pause-aware watchdog that throws a handled StreamTimeoutError.
+    // p-timeout short-circuits on Infinity, handing it sole timeout control.
+    this.bot = new Telegraf(options.botToken, { handlerTimeout: Infinity });
     this.sessionManager = new SessionManager(options);
     this.messageHandler = new MessageHandler(
       this.bot,


griffinwork40 (commented):

## Merge decision: ✅ MERGE

A tightly-scoped, correctly-diagnosed hotfix. The load-bearing mechanism was traced end-to-end against the **resolved** dependencies (not from memory): the change genuinely removes the 90s guillotine and hands sole timeout authority to the purpose-built `streaming.ts` watchdog, which is an independent, always-armed replacement. No security, correctness, data-loss, or api-compat defect found. The two findings below are non-blocking polish.

*(All citations verified against reviewed ref `0ff2c88`. `p-timeout` and `telegraf` claims verified against the versions actually resolved in the lockfile/`node_modules`.)*

### Verified correctness chain

1. **The fix is real, not a no-op.** `src/telegram/bot.ts:74` @ `0ff2c88` — `new Telegraf(options.botToken, { handlerTimeout: Infinity })`. Telegraf 4.16.3 stores this at `this.options` via `{ ...DEFAULT_OPTIONS, ...compactOptions(options) }` (`telegraf/lib/telegraf.js:93`), and `compactOptions` only drops `undefined` values (`compact.js`: `options[key] !== undefined`), so `Infinity` survives. It then wraps the whole turn in `pTimeout(Promise.resolve(this.middleware()(ctx, anoop)), this.options.handlerTimeout)` (`telegraf.js:233`), where the default is `90000` (`telegraf.js:46`).
2. **`Infinity` short-circuits the resolved `p-timeout@4.1.0`.** `p-timeout/index.js:17-19`: `if (milliseconds === Infinity) { resolve(promise); return; }` — executes *before* any `setTimeout` (line 27). So no timer is armed and the middleware promise passes through untouched. The PR's central claim holds for the actually-installed version.
3. **Disabling it is safe — an independent watchdog covers the same failure.** `streaming.ts:278-320` arms a `setTimeout` inactivity watchdog (`FIRST_EVENT_TIMEOUT_MS=90_000`, `NEXT_EVENT_TIMEOUT_MS=180_000`, re-armed on activity, tool-in-flight- and usage-limit-pause-aware) that throws a handled `StreamTimeoutError`. The turn path `handlers/message.ts:612` calls `streamResponse(...)` and `:634` catches `StreamTimeoutError`, so a wedged turn is still bounded and surfaces an honest message rather than the generic `bot.catch`.
4. **Only production `Telegraf` construction site.** `git grep 'new Telegraf('` at `0ff2c88` over `src/**` (excluding tests/mocks) returns exactly one hit — `bot.ts:74` — so no sibling constructor is left un-fixed.
5. **Test is non-vacuous.** `bot.test.ts:115-124` reads `.bot.options.handlerTimeout` — the exact field Telegraf populates (step 1) — and asserts `=== Infinity` and `!== 90_000`. It fails if the option is dropped or reverts to the default.

**1. `low` · perf-observability · src/telegram/streaming.ts:22-30 · ref:0ff2c88 · citation-type:file-state — the two timeout regimes now share the number `90_000`, which invites future confusion.** Telegraf's removed handler cap was 90s and the streaming watchdog's `FIRST_EVENT_TIMEOUT_MS` is also `90_000`; they are semantically unrelated (one wall-clock, one inter-event). *Suggestion:* add a one-line comment at the streaming constant noting it is the *inter-event* window, distinct from the (now-disabled) Telegraf wall-clock handler timeout, so a later reader doesn't "restore parity" by mistake.

**2. `nit` · test-coverage · src/telegram/bot.test.ts:115-124 · ref:0ff2c88 · citation-type:file-state — the regression is guarded structurally (the option value) but not behaviorally (a >90s turn no longer routing to `bot.catch`).** This is acceptable for a hotfix — the behavioral test would require driving a fake >90s middleware through Telegraf — but the guard proves configuration, not outcome. *Suggestion:* leave as-is; note the gap so nobody assumes the long-turn survival path itself is covered.

### What was not checked
- Read exclusively against reviewed ref `0ff2c88e79fe77280779e29185d1e0df31f62e53` via `git show <ref>:<file>` and `git grep <ref>`; the local working tree was on a different dirty branch and was **not** used as source of truth (an early read confirmed the fix was absent from the working tree, so all file-state claims come from the ref).
- Dependency behavior (`telegraf@4.16.3`, `p-timeout@4.1.0`) verified by reading the installed `node_modules` sources, not by executing them; tests were **not** run here. The PR claims `pnpm lint` clean and full suite green (597 files / 11,165) — not independently re-verified.
- The deploy note (rebuild + restart the launchd service) is an operational step outside the diff and is **not** a code requirement — its absence from the diff is not treated as unmet intent.
- Not traced: whether any Telegram entry path performs long agent work *without* going through `streamResponse` (the one early-return path at `streaming.ts:107` returns immediately and does no work, so it cannot hang).
- Stated intent: PR #475 title + body — spec-compliance **assessed** (fully implements the stated fix; no scope creep).

PR #476, opened 2026-07-10
fix(trace): make witness traces correlatable to sessions via ledger meta

griffinwork40:
## Problem

A session can have a durable ledger (`~/.afk/state/sessions/<id>/events.jsonl`) but **no discoverable witness trace**, silently:

- Fresh sessions label the witness dir with a **random UUID** (`state/witness/<uuid>/`), decoupled from the session id — only *resumed* sessions reuse the id as the label (`src/cli/commands/interactive/bootstrap.ts:206`; the writer is opened before the session id exists).
- Trace events carry **no session id** internally.
- Net effect: `afk trace show <sessionId>` can't find a fresh session's trace, and the id-keyed ledger / routing-decisions can't be joined to it.

This was found while trying to reconstruct a real REPL session after the fact — it had a ledger but no locatable trace, and nothing recorded *why*.

## Fix

Smallest reversible change that keeps the two fire-and-forget writers **decoupled** — no bootstrap reorder, no directory rename:

1. **`src/agent/session-ledger.ts`** — add `traceLabel?: string | null` to the one-time `meta` record. Optional for back-compat with pre-field ledgers.
2. **`src/agent/session/agent-session.ts`** — populate it from `this.config.traceWriter?.getTracePath()`. `null` makes tracing-disabled/unwired **explicit** rather than a silently-absent dir.
3. **`src/paths.ts`** — add `sessionLabelFromTracePath()` (inverse of `getTraceDir`; returns `null` for the in-memory sentinel). `getTracePath()` is the documented way to locate a trace without knowing its random label.
4. **`src/cli/commands/trace.ts`** — `loadTrace()` falls back to the ledger's `traceLabel` when `<witness>/<sessionId>/` is absent, and reports a clear *"tracing disabled"* error when `traceLabel` is `null`.

### Why this shape (over alternatives)

Storing the **label** (not a live path/handle) in the always-on ledger solves correlation **and** absence-detection in one field, with no coupling of concern: the ledger opens lazily on first turn (`agent-session.ts:632,671`), by which point the trace label already exists (no race), and we do **not** rename the trace dir (no staleness). A systemic "session-id *is* the trace label from birth" fix was considered but reorders session-identity across all five entry points for marginal gain — not worth the blast radius.

## Verification

- `pnpm audit:sdk:check` ✓ · `pnpm audit:env:check` ✓ · `pnpm scan:env:check` ✓ (no new SDK imports or env vars)
- `pnpm test` ✓ — **11171 passed, 14 skipped, 0 failed**
- New tests: `paths.test.ts` (label round-trip + in-memory/nullish sentinels); `trace.test.ts` (mismatched-label trace resolves via ledger; `traceLabel: null` → clear error; pre-field ledger falls through to "No trace found").

### DIFF
--- a/src/agent/session-ledger.ts
+++ b/src/agent/session-ledger.ts
@@ -38,8 +38,21 @@ import type { ElicitationRequest, ElicitationResult } from './types/sdk-types.js
 export type LedgerRecord = { v: 1; ts: number } & LedgerPayload;
 
 export type LedgerPayload =
-  /** Session-level metadata, written once when the ledger opens. */
-  | { kind: 'meta'; sessionId: string; model: string; cwd?: string; surface?: string }
+  /** Session-level metadata, written once when the ledger opens.
+   *  `traceLabel` is the witness-trace directory name (`state/witness/<label>/`)
+   *  for this session, letting a reader correlate the id-keyed ledger to the
+   *  trace — whose label is a random UUID for fresh sessions, decoupled from
+   *  the session id. `null` means no trace was wired (tracing disabled/failed),
+   *  making that state explicit rather than a silently-absent directory.
+   *  Optional for back-compat with ledgers written before this field existed. */
+  | {
+      kind: 'meta';
+      sessionId: string;
+      model: string;
+      cwd?: string;
+      surface?: string;
+      traceLabel?: string | null;
+    }
   /** A user turn entering the session (summary text, never raw blocks). */
   | { kind: 'user'; text: string }
   /** A complete assistant message. */
--- a/src/agent/session/agent-session.ts
+++ b/src/agent/session/agent-session.ts
@@ -51,6 +51,7 @@ import type {
 } from '../types.js';
 import { QueryInputStream } from './input-iterable.js';
 import { SessionLedgerWriter } from '../session-ledger.js';
+import { sessionLabelFromTracePath } from '../../paths.js';
 import type { ElicitationRequest } from '../types/sdk-types.js';
 import { env } from '../../config/env.js';
 import { resolveModelId } from './model-resolution.js';
@@ -695,6 +696,11 @@ export class AgentSession implements IAgentSession {
       sessionId: id,
       model: meta.model ?? String(this.config.model),
       ...(meta.cwd !== undefined ? { cwd: meta.cwd } : {}),
+      // Correlate this id-keyed ledger to its witness trace: fresh sessions
+      // label the trace dir with a random UUID (not the session id), so the
+      // ledger is the durable id→label bridge. `null` when tracing is
+      // disabled/unwired, so absence is explicit rather than silent.
+      traceLabel: sessionLabelFromTracePath(this.config.traceWriter?.getTracePath()),
     });
   }
 
--- a/src/cli/commands/trace.test.ts
+++ b/src/cli/commands/trace.test.ts
@@ -27,7 +27,7 @@ import {
   listTraces,
   resolveLatestSession,
 } from './trace.js';
-import { getTraceDir } from '../../paths.js';
+import { getTraceDir, getSessionLedgerDir, getSessionLedgerPath } from '../../paths.js';
 
 // ---------------------------------------------------------------------------
 // Fixtures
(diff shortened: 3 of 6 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

A tightly-scoped, correctly-diagnosed hotfix that does exactly what its intent states — no more, no less. The correlation fallback was traced end-to-end against the **reviewed ref** (not the working tree, which is on `main`): the ledger becomes the durable id→label bridge, the inverse-path helper is exact, and every failure mode (tracing-disabled, pre-field ledger, garbage label, in-memory sentinel) is handled explicitly. No security, correctness, data-loss, or api-compat defect found. CI is fully green on this exact ref. The findings below are non-blocking polish.

*(All citations verified by reading `git show b5c33c06:<file>` at the reviewed ref `b5c33c067f1514b6e686f68759bf2234cdeaebf6`.)*

### Verified correctness chain

1. **The inverse helper is exact.** `paths.ts:sessionLabelFromTracePath` returns `basename(dirname(tracePath))`; `getTraceDir` returns `join(getAfkStateDir(),'witness',sessionId)`, so for `.../witness/<label>/trace.jsonl` the inverse recovers `<label>` — proven by the live round-trip test `paths.test.ts` (`is the inverse of getTraceDir`).
2. **In-memory sessions get `null`, not a bogus label.** The in-memory writer's `getTracePath()` returns exactly `'in-memory://trace'` (`src/agent/trace/writer.ts:483`); the guard `tracePath.startsWith('in-memory:')` matches → returns `null`. Asserted by `paths.test.ts` (`returns null for the in-memory writer sentinel`).
3. **The fallback terminates correctly in every branch.** `src/cli/commands/trace.ts:loadTrace` — on `content === null` it calls `traceLabelFromLedger`; `disabled` → clear tracing-disabled error; `label && label !== sessionId` → relabeled read wrapped in `try/catch`; the final `if (content === null)` still throws "No trace found" for the `none`/unresolved cases. All three paths are covered by non-vacuous tests.
4. **Path traversal is contained.** A ledger-supplied label flows into `getTraceDir`, which calls `validateSessionId` (`/^[a-zA-Z0-9_-]+$/`, throws on unsafe input); the relabeled read is wrapped in a `try/catch` that falls through to the not-found error. `readLedger` additionally self-guards with `isSafeLedgerSessionId`.
5. **No api-compat break.** `traceLabel?: string | null` is **optional** on the `meta` union, so pre-field ledgers still typecheck and parse. `git grep traceLabel` across production shows **zero** consumers other than the two write sites and the new reader — no exhaustive meta-switch breaks on the added field. `readLedger` was already exported before this PR imported it into `trace.ts`.
6. **Intent satisfied precisely.** The `agent-session.ts` change is purely additive (one import + one field in the existing `meta` write) — **no bootstrap reorder** — and the diffstat shows **no directory rename**, exactly as the stated intent promised.

**1. `low` · perf-observability · `src/cli/commands/trace.ts:194-210` · ref:b5c33c06 · citation-type:file-state** — a ledger read error inside `traceLabelFromLedger` is swallowed silently, so a genuinely corrupt/unreadable ledger is indistinguishable from "no meta recorded" on the diagnostic path.
Evidence: `} catch { // Ledger unreadable — treat as no signal and fall back to the direct lookup. }`
*Suggestion:* leave the swallow (correct fail-soft for a diagnostic command), but note the gap — a one-line stderr diagnostic on the catch would help a maintainer debugging "why can't `afk trace show` find my trace" distinguish an unreadable ledger from a missing one.

**2. `nit` · test-coverage · `src/cli/commands/trace.ts` · ref:b5c33c06 · citation-type:file-state** — the new tests cover meta-first-record ledgers; they do not exercise a ledger whose first record is *non-meta* (the loop's `continue`-until-`meta` path). This is acceptable — the `meta` record is written once when the ledger opens, so it is the first record in practice — but the `continue` branch is proven only by reading, not by a test.
*Suggestion:* optional — add a fixture with a leading non-meta record to lock the `continue`-to-`meta` behavior; not required for a hotfix.

### What was not checked
- Read exclusively against reviewed ref `b5c33c067f1514b6e686f68759bf2234cdeaebf6` via `git show <ref>:<file>` and `git grep <ref>`; the local working tree was on `main` and was **not** used as source of truth. All file-state citations were verified at that ref.
- Tests were **not** run in this review. The PR reports the full suite + lint + build green, and **all CI checks (Lint & Build, Test, Docs Site, Publish Bundle Smoke) report SUCCESS on this exact ref** per the status rollup — not independently re-executed here.
- Stated intent: PR #476 title + body — spec-compliance **assessed** (all four intent points implemented; no bootstrap reorder; no directory rename; no scope creep found).
- Not traced: whether any caller of `loadTrace` other than the `afk trace show`/`list` CLI path relies on the previous single not-found error shape (the new tracing-disabled error is a distinct, additional message on a previously-uniform failure).

PR #479, opened 2026-07-10
fix(worktree): surface deps-not-installed note on create

### LINKED ISSUE #439: worktree: create leaves a fresh worktree without node_modules (opaque test/build failures)
## Summary

The `worktree` tool's `create` action runs `git worktree add` + writes the meta file, then returns `{ path, branch, base }`. It does **not** run `pnpm install`, symlink `node_modules`, or warn that the fresh tree has no dependencies. A subagent dispatched into a fresh worktree to run `pnpm test` / `build` / `lint` fails with an ordinary module-not-found error and **no pointer** at the fresh-worktree cause.

## Evidence

- `src/agent/tools/handlers/worktree.ts:214-261` — `create`: no install, no symlink, no warning
- No `node_modules` / `pnpm install` reference anywhere in `worktree.ts`

## Why it matters

Fresh worktrees are the normal substrate for isolated subagent work, so this is a common footgun — and the resulting failure is opaque (a bare module resolution error, not "your worktree has no deps").

## Proposed direction

Any of:
- (a) include a note in the `create` result that deps are not installed (suggest `pnpm install`),
- (b) optionally symlink/copy `node_modules` from the main worktree,
- (c) an opt-in `install` flag on the create action.

(a) is the cheapest and removes the surprise.

Severity: LOW–MED. Label: enhancement.

Found via a read-only tool/subagent/worktree audit.

griffinwork40:
## What & why

Closes #439.

When the `worktree` tool's `create` action makes a new worktree, dependencies are **not** installed — a fresh worktree doesn't share the main checkout's `node_modules`. A naive build/test in the new worktree then fails opaquely, with no signal about the actual cause.

This is the minimal fix (issue option **a**): the `create` result now surfaces a clear `note` telling the caller that deps are not installed and giving the **exact install command** to run in the worktree path before building or testing.

### Changes
- **`src/agent/tools/handlers/worktree.ts`**: added a `detectInstallCommand(repoRoot)` helper that inspects the repo-root lockfile (`pnpm-lock.yaml`→`pnpm`, `package-lock.json`→`npm`, `yarn.lock`→`yarn`, `bun.lockb`→`bun`), falling back to `pnpm install`. The `case 'create'` result JSON gains a `note` field embedding that command and the worktree path. Best-effort and non-throwing — it never fails the create.
- **`src/agent/tools/schemas.ts`**: updated the `worktree` tool description so the `create` action documents the new return shape `{ path, branch, base, note }` and what `note` conveys, keeping the model-facing contract accurate.
- **`src/agent/tools/handlers/worktree.test.ts`**: extended the existing create test to assert the `note` is present, mentions deps are not installed, and includes the worktree path; added a parametrized test verifying the recommended command matches each lockfile, plus a fallback-to-`pnpm install` case. Mirrors the file's existing mocked-`execFile` harness.

## Deferred (out of scope)
Auto-installing deps / auto-symlinking `node_modules` into the new worktree was deliberately **not** implemented here — it's a larger change (choosing install vs. symlink strategy, handling failures, perf). This PR is the quick, safe win: make the missing-deps condition visible. Auto-install can follow as a separate change.

## Verification
- `pnpm lint` (`tsc --noEmit`, whole project) — **clean**, exit 0.
- `pnpm exec vitest run src/agent/tools/handlers/worktree.test.ts` — **29 passed** (was 24; +5 new).
- Full suite as a sanity check: **11186 passed | 14 skipped** (597 files).


griffinwork40:
Addressed the P2 (detect install command from created worktree, not repo root) in 48c72b8: detection now inspects `worktreePath` (checked out at `baseRef`) instead of `ctx.repoRoot`, so the note recommends the package manager matching the worktree's lockfile. Added a regression test covering the repoRoot≠worktree lockfile case.

griffinwork40:
## Review — PR #479: `fix(worktree): surface deps-not-installed note on create`

**Target:** GitHub PR #479, branch `afk/issue-439-worktree-deps-note` → `main`. **Reviewed ref:** `48c72b8c5fe609ce6d3701287e2d0e84cc183def` (branch tip; PR is 2 commits: `da61719` initial + `48c72b8` "P2" follow-up). **Stated intent:** PR title + body (verbatim, see Wave 1 dispatch) — closes #439, adds a `note` field to the `worktree` tool's `create` result warning that deps aren't installed in a fresh worktree, with the exact install command. **Triage:** 3 files, +121/‑4 (125 total) → **regime = light** (single agent, all dimensions). Working tree is dirty on an unrelated branch (`retire-diagnose-ts-mirror`); all content was read via `git show <ref>:<path>`, never the live filesystem — no mutation performed.

### Findings (sorted critical → high → medium → low → nit)

1. **low · confidence high · test-coverage · `src/agent/tools/handlers/worktree.test.ts:196-218` · ref:`48c72b8` · citation-type:file-state**
   Finding: the parametrized loop's shared assertion (`expect(parsed.note).toContain(command)`) is vacuous for the `package-lock.json` → `'npm install'` case, because `'pnpm install'.includes('npm install')` is `true` — the test would still pass even if `detectInstallCommand` always fell through to its `pnpm install` fallback and never actually detected `package-lock.json`.
   Evidence:
   ```
   197	    ['package-lock.json', 'npm install'],
   218	      expect(parsed.note).toContain(command);
   ```
   Independently re-derived twice (orchestrator via `node -e "console.log('pnpm install'.includes('npm install'))"` → `true`, before dispatch; Wave‑1 agent via character-index comparison, after dispatch — same answer, same root cause). The line is unchanged between `da61719` and `48c72b8`, so it has been present since the PR's first commit.
   Suggestion: wrap in backticks, mirroring the P2 regression test 3 tests below it (`worktree.test.ts:261-262`): `` expect(parsed.note).toContain(`\`${command}\``) ``. Only the npm/pnpm pair is affected — yarn/bun cases are not substrings of the fallback and already catch a total-fallback regression.

2. **nit · confidence high · spec-compliance · `src/agent/tools/handlers/worktree.ts:76,85` · ref:`48c72b8` · citation-type:file-state**
   Finding: the PR body's "Changes" section states `detectInstallCommand(repoRoot)` "inspects the repo-root lockfile" — that described the first commit (`da61719`) exactly, but the shipped P2 commit (`48c72b8`, this review's ref) changed the signature to `detectInstallCommand(worktreePath)` inspecting the *created worktree's* lockfile instead, and the body was never updated to match.
   Evidence (confirmed by direct diff, not paraphrase):
   ```
   da61719: async function detectInstallCommand(repoRoot: string)      / fs.access(join(repoRoot, file))
   48c72b8: async function detectInstallCommand(worktreePath: string)  / fs.access(join(worktreePath, file))
   ```
   The shipped (P2) behavior is objectively the *correct* one (it's what fixes the real bug — a worktree checked out at a different `base` can have a different lockfile than the main checkout) — this is a documentation-accuracy gap, not a functional defect.
   Suggestion: update the PR body's Changes section to say `detectInstallCommand(worktreePath)`, inspecting the created worktree, not repo root, before merge (or accept as-is — this is metadata, not code).

3. **nit · confidence high · spec-compliance · `src/agent/tools/handlers/worktree.test.ts` · ref:`48c72b8`**
   Finding: the PR body's Verification section claims "29 passed (was 24; +5 new)". Independently confirmed (test file `it()`-site count 26→27 across `da61719`→`48c72b8`, plus `git show 48c72b8 --stat` showing the P2 commit itself net-adds one new test — the `#439 P2` regression test — on top of the 5 already added by `da61719`): the true cumulative count is **+6 new / 30 total**, not +5/29. Same root cause as finding 2 (body written against `da61719`, not refreshed after the P2 amendment).
   Suggestion: refresh the Verification section's count.

4. **nit · confidence medium · correctness · `src/skills/score/index.ts:227-228`**
   Finding: this PR adds a second, structurally similar lockfile-detection helper (4-manager `detectInstallCommand`) without consolidating with the pre-existing, narrower pnpm-vs-npm detector already in the codebase.
   Evidence: `const hasPnpm = await fileExists(join(branchPath, 'pnpm-lock.yaml'));` (skills/score/index.ts:228) vs. the new `lockfiles` array (worktree.ts:77-82).
   Suggestion: optional follow-up — extract a shared `detectPackageManager(dir)` util. Not blocking: different call sites, different purposes (test-running vs. deps-install messaging), small enough that forcing a shared abstraction now may not be worth the coupling.

**Clean dimensions:** security (no issues — fixed 4-entry array + `JSON.stringify`, no attacker-influenced input, no injection surface), api-compat (additive `note` field only; grepped all production call sites of the `create` result shape — zero strict consumers, only the test file destructures it), perf-observability (≤4 sequential `fs.access` stat calls inside an already I/O-bound worktree-creation path — negligible).

### Merge decision: **MERGE**

No critical, high, or blocking findings. All 4 findings are non-blocking (1 low test-coverage gap, 2 documentation-staleness nits, 1 optional-consolidation nit) and don't gate merge — they're worth a follow-up commit or comment at the author's discretion.

### What was not checked (epistemic scope disclosure)

- **Ref read:** all file contents read from `48c72b8c5fe609ce6d3701287e2d0e84cc183def` (the PR branch tip); `da61719588a471bb523eeb4a11e1321be2c55bc7` (the PR's first commit) was read only for before/after comparison, never as the reviewed state. No live-filesystem reads were used (working tree is on an unrelated branch and dirty).
- **Execution:** did not run `pnpm lint` / `pnpm exec vitest` — the PR body's "clean, exit 0" and "11186 passed | 14 skipped" claims are unverified by direct execution. Running them would require checking out the PR branch, which the read-only review contract explicitly forbids given the dirty, unrelated working tree — verified instead via static diff/file-count inspection only.
- **Stated intent:** PR #479 title + body — spec-compliance was assessed (2 documentation-staleness findings above; no unmet requirement, no scope creep beyond the deliberately-deferred auto-install, which the body itself flags as out of scope).
- **Shadow-verify:** not separately invoked as its own skill call — no finding reached the critical/high threshold that mandates it. The two most load-bearing claims (the npm/pnpm substring vacuity, and the `da61719`→`48c72b8` body/code divergence) were independently re-derived by the orchestrator directly against primary git history (`git show`, `node -e`) rather than taken on the Wave‑1 agent's word alone — both confirmed.
- **Not reviewed:** `sanitizeSlug`, `resolveCreateBaseRef`, and the meta-file-write logic in the same handler — all pre-existing, untouched by this diff.

This is the terminal step for this review — no edits, commits, or PR mutations follow.

---

PR #427, opened 2026-07-05
fix(retry): patient 429 retry-after handling + surface stopReason on SubagentResult

griffinwork40:
## Problem

Two related failure modes diagnosed from production telemetry (2026-07-05):

1. **Transient 429s kill headless sessions in seconds.** `turnWithUsageLimitRetry` handles OAuth-subscription 429s (pause/replay), but a `rate-limit-transient` 429 (retry-after header, no `|ts`) deliberately falls through as fatal once the SDK's two internal retries run out. On account-level rate-limit collisions (hourly daemon cron firing at :00 during interactive subagent fan-outs), cron tasks died in ~3.4s with `429 rate_limit_error`.

2. **Callers cannot distinguish a capped partial from a real answer.** `handle.ts` tracks the provider `stopReason` and synthesizes a `[... capped partial result]` message on `tool_use_loop_capped`, but `SubagentResult` never carried the stop reason — so orchestrators (e.g. forge's qualify loop) mechanically misread budget exhaustion as a substantive answer (observed: capped qualify runs scored as REJECT verdicts).

## Changes

**Commit 1 — fix(retry):** `rate-limit-transient` 429s now wait out the server's `retry-after` hint (clamped at 120s, plus 0-1s jitter to de-synchronize concurrent sessions) and silently replay the turn, up to 3 attempts per turn. Budget exhausted -> error surfaces exactly as before. Respects `isClosed()`/abort via `sleepWithAbort`; each backoff recorded as a `rate_limit` session phase (mirrors loop.ts's mid-stream overload trace). Unconditional — honoring retry-after is correct on every surface. OAuth-limit paths untouched.

**Commit 2 — feat(subagent):** optional `stopReason?: string` on `SubagentResult`, threaded from the provider's `done` event through `buildResultFromMessage`/`buildResultFromError`. `tool_use_loop_capped` now lets callers branch on capped partials instead of parsing the marker string.

## Verification

- `tsc --noEmit` clean
- `vitest run src/agent/providers/anthropic-direct/`: 357/357 pass (3 new: silent replay success, budget exhaustion, oauth-limit interleave unchanged)
- `vitest run src/agent/subagent/`: 48/48 pass (5 new stopReason assertions incl. capped -> `tool_use_loop_capped`)
- SubagentResult consumers: background-registry + subagent-executor tests 132/132 pass

[...]

PR #483, opened 2026-07-10
fix(anthropic): restore usage-limit pause + hot-swap for subscription 429s (regression)

griffinwork40:
## Summary

Fixes a regression where hitting the Claude **subscription** usage limit no longer produced a notification **and** no longer auto-resumed after switching accounts.

Both symptoms trace to a **single misclassification** in `classifyUsageLimitError`, introduced by #414 and armed by #427.

## Root cause

A 429 is sorted into:
- `oauth-limit` — message carries `|<unix-ts>` → pause + wait-for-reset (watches keychain).
- `oauth-limit-no-ts` — no timestamp, **no** `retry-after` → pause + hot-swap poll.
- `rate-limit-transient` — **has a `retry-after` header, no timestamp** → *silent* wait-and-replay ≤3×, **no `paused` event, no keychain watch**.

#414 assumed "429 + `retry-after` + no `|ts` = a transient per-minute throttle." That assumption is **false against the real API**: Anthropic delivers an OAuth **subscription** cap as a bare `429 rate_limit_error` **with** a `retry-after` header and **no** `|ts` — byte-identical in shape to a transient throttle (see [anthropics/claude-code#30930](https://github.com/anthropics/claude-code/issues/30930); Anthropic docs confirm *every* 429 carries `retry-after`). Before #414 afk did not parse `retry-after`, so these fell to `oauth-limit-no-ts` and worked; #414 started parsing it and rerouted real subscription exhaustion into the silent transient path.

Consequences of the reroute:
- **No notification** — the "⏸ Usage paused" card fires only on a `paused` event (`telegram/streaming.ts`, REPL `render/usage-limit-box.ts`). The transient path never emits `paused`.
- **No autoresume on account switch** — the transient path uses `sleepWithAbort`, never `waitForHotSwap`/`waitForReset`, so a mid-wait keychain token swap is never detected; after 3 short retries a raw error surfaces.

## Fix

Header *presence* cannot discriminate the two — **magnitude can**. A per-minute throttle clears in seconds; a subscription reset is hours away. Gate `rate-limit-transient` on `retry-after <= RATE_LIMIT_TRANSIENT_MAX_RETRY_AFTER_MS` (5 min). A **long or absent** backoff now falls through to `oauth-limit-no-ts` → pause + notification + hot-swap + auto-resume (pre-#414 behavior). Genuine transient collisions (small `retry-after`, the daemon-cron case #414/#427 fixed) keep the patient short-retry.

## Changes
- `usage-limit.ts`: threshold constant + magnitude gate; corrected the now-false invariant comments.
- `usage-limit.test.ts`: +2 regression tests (long `retry-after` → `oauth-limit-no-ts`; threshold boundary).

## Testing
- `pnpm test` — 11,183 pass / 0 fail
- `pnpm audit:sdk:check` — OK

## Known edge / possible follow-up
A subscription cap that sends a *small* (<5 min) `retry-after` would still misclassify — unlikely for the account-switch scenario (reset is far away → large/absent backoff) and it fails bounded. Optional hardening (not in this PR, to keep it minimal): make the transient wait itself hot-swap-aware (`sleepWithAbort` → `waitForHotSwap`) so account-switching resumes even in a misclassified transient wait.

### DIFF
--- a/src/agent/providers/anthropic-direct/usage-limit.test.ts
+++ b/src/agent/providers/anthropic-direct/usage-limit.test.ts
@@ -3,7 +3,13 @@
  */
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
-import { classifyUsageLimitError, parseRetryAfterMs, waitForReset, waitForHotSwap } from './usage-limit.js';
+import {
+  classifyUsageLimitError,
+  parseRetryAfterMs,
+  waitForReset,
+  waitForHotSwap,
+  RATE_LIMIT_TRANSIENT_MAX_RETRY_AFTER_MS,
+} from './usage-limit.js';
 
 // ---------------------------------------------------------------------------
 // classifyUsageLimitError
@@ -101,6 +107,30 @@ describe('classifyUsageLimitError', () => {
     if (result?.kind === 'rate-limit-transient') {
       expect(result.retryAfterMs).toBe(5_000);
     }
+  });
+
+  // Regression (2026-07): Anthropic delivers an OAuth *subscription* cap as a
+  // bare `429 rate_limit_error` + a LONG retry-after and no |ts — identical in
+  // shape to a transient throttle except for magnitude (anthropics/claude-
+  // code#30930). It must route to the pause + hot-swap path (oauth-limit-no-ts),
+  // NOT the silent transient-retry path — which emits no `paused` event (so the
+  // operator is never notified) and never watches the keychain for an account
+  // switch (so switching accounts never resumes the turn).
+  it('returns oauth-limit-no-ts for a 429 with a long retry-after and no |ts (subscription cap)', () => {
+    const err = makeErrorWithHeaders(429, '429 rate_limit_error', { 'retry-after': '3600' });
+    expect(classifyUsageLimitError(err)?.kind).toBe('oauth-limit-no-ts');
+  });
+
+  it('routes a retry-after AT the transient threshold as transient, just PAST it as subscription', () => {
+    const atThreshold = makeErrorWithHeaders(429, '429', {
+      'retry-after-ms': String(RATE_LIMIT_TRANSIENT_MAX_RETRY_AFTER_MS),
+    });
+    expect(classifyUsageLimitError(atThreshold)?.kind).toBe('rate-limit-transient');
+
+    const pastThreshold = makeErrorWithHeaders(429, '429', {
+      'retry-after-ms': String(RATE_LIMIT_TRANSIENT_MAX_RETRY_AFTER_MS + 1),
+    });
+    expect(classifyUsageLimitError(pastThreshold)?.kind).toBe('oauth-limit-no-ts');
   });
 });
 
--- a/src/agent/providers/anthropic-direct/usage-limit.ts
+++ b/src/agent/providers/anthropic-direct/usage-limit.ts
@@ -22,6 +22,21 @@ export type UsageLimitClassification =
   | { kind: 'credit-exhausted' };
 
 /**
+ * Upper bound (ms) on a `retry-after` hint that still counts as a *transient*
+ * API-tier rate limit (per-minute RPM/ITPM/OTPM windows that clear in seconds).
+ *
+ * Invariant: Anthropic returns the SAME `429 rate_limit_error` + `retry-after`
+ * shape for BOTH a per-minute API throttle AND an OAuth *subscription* cap — the
+ * subscription limit arrives as a bare `rate_limit_error` with a `retry-after`
+ * header and no `|<ts>` (see anthropics/claude-code#30930). Header PRESENCE
+ * therefore cannot distinguish them; only its MAGNITUDE can. Per-minute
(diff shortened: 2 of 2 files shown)


griffinwork40:
Git history fully confirms the regression narrative. I now have everything for a complete, verified review.

## Code Review — PR #483

**`fix(anthropic): restore usage-limit pause + hot-swap for subscription 429s (regression)`**
`origin/afk/20260710-152526-acbbea` → `main` · +72/-16 · 2 files

### Recommendation: ✅ **APPROVE** — merge-ready, no blockers

A tight, additive, well-documented regression fix. The magnitude gate is logically sound, the end-to-end pause+hot-swap path is verified against the consumer, the regression narrative is confirmed in git, and the change does not re-break the bug the transient path originally fixed.

---

### What it does
Anthropic delivers an OAuth **subscription** cap and a per-minute **API throttle** with the *identical* shape: `429 rate_limit_error` + `retry-after`, no `|<ts>` (cites `anthropics/claude-code#30930`). Presence can't distinguish them, so the PR gates on **magnitude**: `retry-after ≤ 5min` → `rate-limit-transient` (silent bounded retry); `> 5min` or absent → `oauth-limit-no-ts` (pause event + hot-swap watch + auto-resume).

### Verification performed
- **Correctness (end-to-end):** `oauth-limit-no-ts` in `query/retry-layer.ts:321` emits `paused{reason:'usage-limit'}` (operator notified) then enters the `waitForHotSwap` poll-retry loop — so the fix genuinely restores both the pause notification *and* the account hot-swap. The long `retry-after` value discarded by the classifier is a non-issue: the no-ts path polls on its own `NO_TS_RETRY_INTERVAL_MS=60s`.
- **Regression narrative (git-confirmed):** Introduced by `5dbc55b` (#414, 2026-07-04) which made *all* retry-after 429s transient to stop parking API throttles in a 2h poll; #483 selectively restores the subscription path. **Does not re-break #414:** short (per-minute) retry-after still routes transient — proven by the `atThreshold` vs `pastThreshold` tests. Confidence: high.
- **Test coverage:** 3 new tests cover the long case + both sides of the boundary, across both header forms (`retry-after` seconds and `retry-after-ms`). All new + existing tests pass by construction (deterministic pure-function assertions traced by hand).
- **API-compat:** Purely additive — new exported `RATE_LIMIT_TRANSIENT_MAX_RETRY_AFTER_MS`; `UsageLimitClassification` union unchanged; `retry-layer.ts` and `telegram/error-utils.ts` switch on `kind` (unchanged set), so no consumer breaks.
- **Convention:** Inline `Invariant:` comment blocks correctly follow the repo's long-comment prefix convention for this externally-governed API semantic.

### Findings — non-blocking

- **[Note] The 5-min threshold is a load-bearing empirical heuristic.** The whole gate rests on "subscription caps arrive with `retry-after > 5min`; throttles arrive ≤5min," sourced to `claude-code#30930` and not independently verifiable in-repo. This is fine to ship because *both* misclassification directions degrade gracefully: a slow throttle mis-routed to pause auto-resumes; a near-reset subscription cap mis-routed to transient succeeds within the ~6-min retry budget. Flagging so the maintainer knows the threshold is the tuning knob if field data shifts.
- **[Nit] Threshold (5min) exceeds the per-attempt transient clamp (`RATE_LIMIT_RETRY_MAX_WAIT_MS=120s`).** A `retry-after` in (2min, 5min] is classified transient but under-waited per attempt; the 3-attempt/~6-min budget still covers it, so no functional bug — just a slight premature-replay inefficiency. Optional: align the constant or add a one-line note.
- **[Nit] No integration-level test** in `query-auth-retry.test.ts` asserting a *long-retry-after* 429 actually produces a `paused` event end-to-end (existing `oauth-limit-no-ts` integration tests build 429s with no retry-after header). Unit classification + already-covered no-ts integration path make the risk low, but one integration test would nail the full restored behavior.

### Not done / caveats
- **Test suite not executed.** The working tree is on `feat/first-class-worktree-isolation` (dirty, 8 uncommitted changes) — checking out the PR branch would mutate the operator's in-progress state, which the review contract forbids. Verification was static: I confirmed the PR branch content matches the diff artifact byte-for-byte via `git show`, and hand-traced all assertions. CI on the PR provides the executed proof.
- This review is **read-only** — no files edited, nothing committed, no PR comment posted.

**Done**
- Reviewed PR #483 across correctness, test-coverage, api-compat, and regression-history dimensions; verdict: **APPROVE**, no blockers, 1 note + 2 nits.
- Evidence: fix logic `src/agent/providers/anthropic-direct/usage-limit.ts:84-119` (PR ref, verified faithful to diff artifact); consumer `query/retry-layer.ts:321-402`; regressing commit `5dbc55b` (#414) confirmed by git-investigator; new tests `usage-limit.test.ts:32-47`.
- World unchanged — read-only review; the dirty working tree was preserved (no checkout/stash/commit).
- Pending: optional integration test for the long-retry-after→`paused` path, and the empirical 5-min threshold to watch in the field; test suite left to CI since the working tree could not be safely switched.

---

PR #490, opened 2026-07-10
feat(anthropic): classify usage-limit 429s via anthropic-ratelimit-unified-* headers (#488)

### LINKED ISSUE #488: usage-limit: classify subscription vs transient 429s via `anthropic-ratelimit-unified-*` headers, not the retry-after magnitude proxy (follow-up to #483)
Follow-up to #483 (merged).

## Summary

#483 restored the usage-limit **pause + hot-swap** for OAuth subscription 429s by gating on the **magnitude** of the `retry-after` header (`RATE_LIMIT_TRANSIENT_MAX_RETRY_AFTER_MS = 5min`): `<=5min -> rate-limit-transient` (silent retry), `>5min or absent -> oauth-limit-no-ts` (pause + hot-swap). That fix is correct and a strict improvement over the regression it addressed — this issue is **not** about reverting it.

However, the 5-minute threshold is a **heuristic proxy**, not the signal the reference implementation uses. We verified against the installed Claude Code binary (v2.1.206) that Claude Code classifies subscription-cap vs. transient 429s by the **presence of `anthropic-ratelimit-unified-*` response headers** plus an **authoritative reset epoch** — it never inspects `retry-after` magnitude for this decision. Those same headers ride on the exact `error.headers` object AFK already reads in `parseRetryAfterMs`, so a strictly better, authoritative classification is available today.

## What Claude Code actually does (verified, v2.1.206)

Decompiled from `~/.local/share/claude/versions/2.1.206`:

**Subscription-429 detector — keys on header presence, reads authoritative reset:**
```js
var Pdu="anthropic-ratelimit-unified-representative-claim",
    Odu="anthropic-ratelimit-unified-overage-status";
function Udu(e){
  let t=e.headers?.get?.(Pdu), r=e.headers?.get?.(Odu);
  if(!t && !r) return null;                 // no unified headers => NOT a usage-limit 429
  let n={status:"rejected", ...};
  let o=e.headers?.get?.("anthropic-ratelimit-unified-reset");
  if(o) n.resetsAt = Number(o);             // authoritative reset epoch (seconds)
  if(t) n.rateLimitType = t;                // five_hour | seven_day | seven_day_opus | overage
  ...
}
```

**429 handler — "temporary limit" vs "usage limit" decided by that detector, never by retry-after size:**
```js
if(e.status===429){
  let i=/*isSubscriber*/, s=Udu(e);
  if(i && s) { /* usage-limit message built from s.resetsAt */ }
  ...
  let p = i ? "Server is temporarily limiting requests (not your usage limit)"
            : "Request rejected (429)";      // transient branch
}
```

**Vendored Anthropic SDK retry honors `retry-after` directly — no 5-min cutoff:**
```js
retryRequest(...){
  let o = parseFloat(headers.get("retry-after-ms"))
       || parseFloat(headers.get("retry-after"))*1000
       || calculateDefaultRetryTimeoutMillis(...);
  await sleep(o); retry();
}
```

A programmatic scan (a sub-agent + a manual read) found **zero** `retry-after`-vs-`300000`/`5*60` comparisons anywhere in the binary; the `300000` literals are OAuth token-refresh thresholds / request timeouts, and the only retry-after guard is an upper-bound abort (`tengu_api_retry_after_too_long`), not a subscription classifier.

## Proposed change

In `classifyUsageLimitError` (`src/agent/providers/anthropic-direct/usage-limit.ts`), prefer header-based classification, with the magnitude heuristic demoted to a last-resort fallback:

1. If `anthropic-ratelimit-unified-representative-claim` OR `anthropic-ratelimit-unified-overage-status` is present -> it's a **subscription cap**.
   - If `anthropic-ratelimit-unified-reset` is also present -> return `oauth-limit` with `resetsAt = new Date(Number(reset) * 1000)`. This is an **authoritative** deadline and upgrades today's timestamp-less `oauth-limit-no-ts` blind-poll path to the real `oauth-limit` wait.
   - Else -> `oauth-limit-no-ts` (as today).
2. Else if only per-minute headers (`anthropic-ratelimit-{requests,input-tokens,output-tokens}-*`) are present -> `rate-limit-transient`.
3. Else (no rate-limit headers at all) -> fall back to the current `|<ts>` message parse, then the `retry-after` magnitude heuristic (`RATE_LIMIT_TRANSIENT_MAX_RETRY_AFTER_MS`) as the final tiebreaker.

The consumer (`src/agent/providers/anthropic-direct/query/retry-layer.ts`) already handles all three `kind`s; the main behavioral win is that more real subscription caps land in the `oauth-limit` path with a true `resetsAt` instead of the 60s blind poll (`NO_TS_RETRY_INTERVAL_MS`).

## Files
- `src/agent/providers/anthropic-direct/usage-limit.ts` — `classifyUsageLimitError`; reuse/extend `parseRetryAfterMs`'s defensive header-access helper (handles both web `Headers` and plain-record shapes) for the unified-header reads.
- `src/agent/providers/anthropic-direct/query/retry-layer.ts` — confirm the `oauth-limit` path is exercised by the new header-derived `resetsAt`.
- `src/agent/providers/anthropic-direct/usage-limit.test.ts` — add cases: unified claim + reset, unified claim without reset, per-minute-only headers, no-header fallback.

## Acceptance criteria
- [ ] A 429 with `anthropic-ratelimit-unified-representative-claim` + `anthropic-ratelimit-unified-reset` classifies as `oauth-limit` with `resetsAt` derived from the header (not `oauth-limit-no-ts`).
- [ ] A 429 with a `unified-*` claim but no `unified-reset` classifies as `oauth-limit-no-ts`.
- [ ] A 429 with only `anthropic-ratelimit-requests/tokens-*` headers classifies as `rate-limit-transient`.
- [ ] A 429 with no rate-limit headers falls back to the existing `|<ts>` -> retry-after-magnitude logic (no behavior change for that path).
- [ ] The 5-min magnitude threshold remains only as the final fallback.

## Open question to resolve first (blocking implementation)
Confirm the `anthropic-ratelimit-unified-*` headers are actually populated on AFK's `/v1/messages` **OAuth subscription** 429 `APIError.headers`. Log `error.headers` on one real occurrence before building on them. Header names/semantics are officially documented (https://platform.claude.com/docs/en/api/rate-limits) and CC's code path reads them, but guarantee-of-presence on the inference endpoint's OAuth 429 is observed-true (community capture) rather than doc-guaranteed.

Note: issue `anthropics/claude-code#30930` — cited in #483 — is about the `/api/oauth/usage` *metadata* endpoint (returns `429` with `retry-after: 0`), **not** `/v1/messages`. It demonstrates that retry-after magnitude is unreliable, but it is not itself a sample of the inference-path subscription 429.

## Header reference
- `anthropic-ratelimit-unified-status` — `allowed` | `allowed_warning` | `rejected`
- `anthropic-ratelimit-unified-representative-claim` — binding window / `rateLimitType`: `five_hour` | `seven_day` | `seven_day_opus` | `overage` | ...
- `anthropic-ratelimit-unified-reset` — Unix epoch (seconds) of window reset (authoritative wait target)
- `anthropic-ratelimit-unified-overage-status`, `-overage-reset`, `-fallback`, `-5h-*`, `-7d-*` — overage / per-window detail
- Per-minute throttles instead carry `anthropic-ratelimit-{requests,input-tokens,output-tokens}-{limit,remaining,reset}`

---
_Verification context: surfaced during the #483 review; classifier confirmed against the Claude Code v2.1.206 binary + official rate-limit docs + deobfuscated `withRetry.ts` mirrors. Confidence: high that CC uses header-presence + `unified-reset` (directly read from the binary); the one unverified link is whether AFK's own OAuth 429 error object carries the headers (the blocking open question above)._

griffinwork40:
## Summary

`classifyUsageLimitError` now prefers Anthropic's authoritative `anthropic-ratelimit-unified-*` response headers to tell an OAuth **subscription cap** apart from a **transient per-minute throttle** — the same signal Claude Code keys on (verified against v2.1.206). The existing 5-minute `retry-after` magnitude gate (`RATE_LIMIT_TRANSIENT_MAX_RETRY_AFTER_MS`) is demoted to a **last-resort fallback**, used only when a 429 carries no rate-limit headers at all. When the `unified-reset` header is present it yields a real `resetsAt`, upgrading the timestamp-less blind-poll path to an authoritative `waitForReset`.

Closes #488

## Safety (no regression)

The change is strictly **additive** and **presence-gated**. When **none** of the unified/per-minute `anthropic-ratelimit-*` headers are present, the classification result is **byte-for-byte identical** to today's fallback (step 4 — the `|ts` parse, then the `retry-after` magnitude heuristic). This guarantees no behavior change regardless of whether real subscription 429s actually carry the headers. `RATE_LIMIT_TRANSIENT_MAX_RETRY_AFTER_MS` remains exported and is used only in that fallback.

## Test evidence

- `src/agent/providers/anthropic-direct/usage-limit.test.ts` — **48/48 pass** (14 new cases: unified claim + reset -> `oauth-limit` with header-derived `resetsAt`; unified claim without reset -> `oauth-limit-no-ts`; `unified-status: rejected` -> `oauth-limit-no-ts`; per-minute-only headers -> `rate-limit-transient`; unified header wins over a short `retry-after`; `|ts` wins over unified; fallback short/long retry-after regression guards; web-`Headers` read).
- `src/agent/providers/anthropic-direct/query-auth-retry.test.ts` — **26/26 pass** (no downstream regression).
- `pnpm lint` (`tsc --noEmit`) — clean.

## Open question (post-merge, blocking real-world benefit)

Confirm the `anthropic-ratelimit-unified-*` headers are **actually present** on AFK's real `/v1/messages` **OAuth subscription** 429 `error.headers`. The header names/semantics are officially documented and Claude Code's code path reads them, but guarantee-of-presence on the inference endpoint's OAuth 429 is community-observed, not doc-guaranteed. Recommend a telemetry/logging check that captures `error.headers` on one real occurrence before relying on the header path. Until then, the presence-gated fallback preserves current behavior, so this ships safely.

## What changed

- **`usage-limit.ts`** — reworked the `status === 429` branch into a 4-step precedence (see summary); rewrote the `classifyUsageLimitError` doc comment (opens with `Invariant:`) to describe it.
- **`usage-limit.ts`** — factored the defensive header-access logic out of `parseRetryAfterMs` into a shared, exported `getHeader(error, name)` helper (handles both a web `Headers` with `.get()` and a plain lower/upper-case record); `parseRetryAfterMs` reuses it with identical behavior.
- **`usage-limit.test.ts`** — added the cases above; all pre-existing tests unchanged and passing.
- **`retry-layer.ts`** — **not modified.** Its `oauth-limit` (has-`resetsAt`) path already calls `waitForReset({ resetsAt, ... })` and is agnostic to whether `resetsAt` came from the `|ts` parse or the `unified-reset` header, so a header-derived deadline flows through the existing wait/resume path with no change.

### DIFF
--- a/src/agent/providers/anthropic-direct/usage-limit.test.ts
+++ b/src/agent/providers/anthropic-direct/usage-limit.test.ts
@@ -260,6 +260,92 @@ describe('classifyUsageLimitError', () => {
       expect(result.resetsAt.getTime()).toBe(resetEpochSec * 1000);
     }
   });
+
+  // -------------------------------------------------------------------------
+  // #490 review follow-ups — edge cases the original #488 suite did not cover.
+  // -------------------------------------------------------------------------
+
+  // M1 regression: a `unified-reset` past the ECMAScript Date ceiling must NOT
+  // produce an `oauth-limit` with an Invalid Date (NaN getTime), which would
+  // slip past the downstream `> TWO_HOURS_MS` surface-guard and hang
+  // `waitForReset` on a NaN deadline. It falls to the timestamp-less path.
+  it('falls to oauth-limit-no-ts for a unified-reset beyond the JS Date range', () => {
+    const err = makeErrorWithHeaders(429, '429 rate_limit_error', {
+      'anthropic-ratelimit-unified-representative-claim': 'five_hour',
+      'anthropic-ratelimit-unified-reset': '99999999999999', // ~1e14 s → overflows Date
+    });
+    expect(classifyUsageLimitError(err)?.kind).toBe('oauth-limit-no-ts');
+  });
+
+  // M1 regression, step-1 parity: an out-of-range `|ts` timestamp falls through
+  // the header steps to the fallback (no retry-after → oauth-limit-no-ts),
+  // never an Invalid-Date oauth-limit.
+  it('falls through a |ts timestamp beyond the JS Date range (step 1 guard)', () => {
+    const err = makeError(429, 'usage limit|99999999999999');
+    expect(classifyUsageLimitError(err)?.kind).toBe('oauth-limit-no-ts');
+  });
+
+  // Precedence: when BOTH a unified subscription header and per-minute throttle
+  // headers are present, step 2 (subscription) wins over step 3 (transient).
+  it('prefers the unified subscription header over per-minute throttle headers', () => {
+    const resetEpochSec = Math.floor(Date.now() / 1000) + 3600;
+    const err = makeErrorWithHeaders(429, '429 rate_limit_error', {
+      'anthropic-ratelimit-unified-representative-claim': 'five_hour',
+      'anthropic-ratelimit-unified-reset': String(resetEpochSec),
+      'anthropic-ratelimit-requests-remaining': '0',
+      'retry-after': '30',
+    });
+    const result = classifyUsageLimitError(err);
+    expect(result?.kind).toBe('oauth-limit');
+    if (result?.kind === 'oauth-limit') {
+      expect(result.resetsAt.getTime()).toBe(resetEpochSec * 1000);
+    }
+  });
+
+  // Step 3 with per-minute headers but no retry-after hint → transient with an
+  // undefined retryAfterMs (retry-layer falls back to its default backoff).
+  it('returns rate-limit-transient with undefined retryAfterMs for per-minute headers and no retry-after', () => {
+    const err = makeErrorWithHeaders(429, '429 rate_limit_error', {
+      'anthropic-ratelimit-requests-remaining': '0',
+    });
+    const result = classifyUsageLimitError(err);
+    expect(result?.kind).toBe('rate-limit-transient');
+    if (result?.kind === 'rate-limit-transient') {
+      expect(result.retryAfterMs).toBeUndefined();
+    }
+  });
+
+  // Characterization (review M2 / #488 design): step 3 is intentionally
+  // magnitude-blind — per-minute headers route to `rate-limit-transient` even
+  // when retry-after exceeds the transient threshold. retry-layer clamps the
+  // wait, so this stays a bounded retry rather than a subscription pause.
+  it('keeps per-minute headers with a long retry-after as rate-limit-transient (magnitude-blind step 3)', () => {
+    const err = makeErrorWithHeaders(429, '429 rate_limit_error', {
+      'anthropic-ratelimit-requests-remaining': '0',
+      'retry-after': '3600',
+    });
+    const result = classifyUsageLimitError(err);
+    expect(result?.kind).toBe('rate-limit-transient');
+    if (result?.kind === 'rate-limit-transient') {
+      expect(result.retryAfterMs).toBe(3_600_000);
+    }
+  });
+
+  // Characterization (review L2): a past `unified-reset` is accepted as-is
+  // (parity with the |ts path); downstream the deadline collapses to ~now →
+  // immediate replay. Guarding future-ness is deliberately left as follow-up.
+  it('accepts a past unified-reset as oauth-limit with a past resetsAt', () => {
+    const pastSec = Math.floor(Date.now() / 1000) - 3600;
+    const err = makeErrorWithHeaders(429, '429 rate_limit_error', {
+      'anthropic-ratelimit-unified-representative-claim': 'five_hour',
+      'anthropic-ratelimit-unified-reset': String(pastSec),
+    });
+    const result = classifyUsageLimitError(err);
+    expect(result?.kind).toBe('oauth-limit');
+    if (result?.kind === 'oauth-limit') {
+      expect(result.resetsAt.getTime()).toBe(pastSec * 1000);
+    }
+  });
 });
 
 // ---------------------------------------------------------------------------
--- a/src/agent/providers/anthropic-direct/usage-limit.test.ts
+++ b/src/agent/providers/anthropic-direct/usage-limit.test.ts
@@ -132,6 +132,134 @@ describe('classifyUsageLimitError', () => {
     });
     expect(classifyUsageLimitError(pastThreshold)?.kind).toBe('oauth-limit-no-ts');
   });
+
+  // -------------------------------------------------------------------------
+  // #488 — prefer `anthropic-ratelimit-unified-*` headers over the retry-after
+  // magnitude proxy. These are the authoritative signals Claude Code keys on
+  // (verified v2.1.206). The change is additive + presence-gated: the fallback
+  // regression guards above must keep passing unchanged.
+  // -------------------------------------------------------------------------
+
+  it('returns oauth-limit with a header-derived resetsAt for a unified claim + unified-reset', () => {
+    const resetEpochSec = Math.floor(Date.now() / 1000) + 3600; // 1h in the future
+    const err = makeErrorWithHeaders(429, '429 rate_limit_error', {
+      'anthropic-ratelimit-unified-representative-claim': 'five_hour',
+      'anthropic-ratelimit-unified-reset': String(resetEpochSec),
+    });
+    const result = classifyUsageLimitError(err);
+    expect(result?.kind).toBe('oauth-limit');
+    if (result?.kind === 'oauth-limit') {
+      expect(result.resetsAt.getTime()).toBe(resetEpochSec * 1000);
+    }
+  });
+
+  it('returns oauth-limit-no-ts for a unified claim header with NO unified-reset', () => {
+    const err = makeErrorWithHeaders(429, '429 rate_limit_error', {
+      'anthropic-ratelimit-unified-representative-claim': 'seven_day',
+    });
+    expect(classifyUsageLimitError(err)?.kind).toBe('oauth-limit-no-ts');
+  });
+
+  it('returns oauth-limit-no-ts for unified-status "rejected" with no reset', () => {
+    const err = makeErrorWithHeaders(429, '429 rate_limit_error', {
+      'anthropic-ratelimit-unified-status': 'rejected',
+    });
+    expect(classifyUsageLimitError(err)?.kind).toBe('oauth-limit-no-ts');
+  });
+
+  it('returns oauth-limit-no-ts for a unified-overage-status header (no reset)', () => {
+    const err = makeErrorWithHeaders(429, '429 rate_limit_error', {
+      'anthropic-ratelimit-unified-overage-status': 'active',
+    });
+    expect(classifyUsageLimitError(err)?.kind).toBe('oauth-limit-no-ts');
+  });
+
+  it('ignores a non-finite / non-positive unified-reset and falls to oauth-limit-no-ts', () => {
+    const nonFinite = makeErrorWithHeaders(429, '429', {
+      'anthropic-ratelimit-unified-representative-claim': 'five_hour',
+      'anthropic-ratelimit-unified-reset': 'not-a-number',
+    });
+    expect(classifyUsageLimitError(nonFinite)?.kind).toBe('oauth-limit-no-ts');
+
+    const nonPositive = makeErrorWithHeaders(429, '429', {
+      'anthropic-ratelimit-unified-representative-claim': 'five_hour',
+      'anthropic-ratelimit-unified-reset': '0',
+    });
+    expect(classifyUsageLimitError(nonPositive)?.kind).toBe('oauth-limit-no-ts');
+  });
+
+  it('returns rate-limit-transient for only per-minute headers (no unified headers)', () => {
+    const err = makeErrorWithHeaders(429, '429 rate_limit_error', {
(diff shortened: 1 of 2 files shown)


griffinwork40:
Wave 1.5 verification complete — all load-bearing claims **confirmed**:

- **A1 (verified):** `retry-layer.ts:408` guards with `resetsAt.getTime() - Date.now() > TWO_HOURS_MS`; an Invalid Date's `NaN` slips past (NaN comparison is false) and `waitForReset` gets a `NaN` deadline that never fires → real hang path. The NaN-reset finding is genuine.
- **A2 (verified):** transient path uses `transientRetryAfterMs ?? DEFAULT(5s)` clamped to `MAX_WAIT(120s)`, ≤3 replays. So `undefined` retryAfterMs is safe (F2), and long retry-after is bounded (F3 is non-fatal). `parseRetryAfterMs` never returns `NaN` (isFinite-guarded), so the NaN hang is reachable **only** via the `oauth-limit` resetsAt path.
- **A3 (verified):** Step 1 `|ts` path (`parseInt` + `ts > 0`, line 143-144) is equally unguarded → the reset-overflow issue is **pre-existing pattern parity**, not a PR-introduced regression.
- **B (verified absent):** no test covers unified+per-minute together, per-minute without retry-after, per-minute + long retry-after, or a past/huge unified-reset. Test-coverage gaps are real.

No fabricated citations, no false-absence claims. Synthesizing.

---

# Code Review — PR #490

**`feat(anthropic): classify usage-limit 429s via anthropic-ratelimit-unified-* headers (#488)`**
Target: PR #490 · branch `afk/issue-488-unified-ratelimit-headers` → `main` · reviewed ref `c21fdb1` · +253/−47, 2 files · regime **full** (2 parallel dimension agents + citation/absence verification).

## Summary

Reworks the `status === 429` branch of `classifyUsageLimitError` into a 4-step precedence — (1) `|<ts>` message parse, (2) `anthropic-ratelimit-unified-*` headers, (3) per-minute throttle headers, (4) the old `retry-after` magnitude heuristic as a no-headers fallback — and extracts the defensive header read into a shared, exported `getHeader()`. The `getHeader` extraction and `parseRetryAfterMs` refactor are clean and behavior-preserving; api-compat is fully additive; the "byte-for-byte identical when no rate-limit headers are present" safety claim is **structurally true** (both new gates are `!== undefined` ORs, so with no headers control falls unconditionally to the unchanged step 4). The findings are all **medium or below** and cluster on the header-*present* paths — which the PR itself does not claim are unchanged.

### 🟡 MEDIUM

**M1 · correctness/robustness · `usage-limit.ts:163-166` (parity: `143-144`) · ref:c21fdb1 · verified**
A `unified-reset` value past the JS Date ceiling (~`8.64e12` epoch-seconds) passes the `Number.isFinite(resetSec) && resetSec > 0` gate but produces `new Date(NaN)`; downstream `retry-layer.ts:408` (`resetsAt.getTime() - Date.now() > TWO_HOURS_MS`) evaluates the NaN comparison as `false` (so the error is *not* surfaced), then `waitForReset` computes a `NaN` deadline that never fires → the turn hangs until abort or hot-swap.
```
if (Number.isFinite(resetSec) && resetSec > 0) {
  return { kind: 'oauth-limit', resetsAt: new Date(resetSec * 1000) };
```
*Both dimension agents flagged this independently; downstream hang confirmed via read of retry-layer.ts:408 + waitForReset.* **Not a regression** — Step 1's `|ts` path (line 144) has the identical missing upper bound. Extreme/implausible from a real Anthropic header. **Fix:** reject `resetSec` (and `ts`) when `Number.isNaN(new Date(x * 1000).getTime())`, or clamp to a sane forward window — ideally one shared guard for both steps.

**M2 · correctness · `usage-limit.ts:185` · ref:c21fdb1 · diff-context**
Step 3 is magnitude-blind: a 429 carrying per-minute headers **and** a `retry-after` > 5 min now returns `rate-limit-transient` (pre-PR: `oauth-limit-no-ts` → pause). Bounded downstream (wait clamped to 120s, ≤3 replays, then the error surfaces), and real per-minute windows reset in ≤60s, so materiality is low — but it is a behavior change on the header-present path. **Fix (optional):** apply the 5-min gate inside Step 3, or route a long `retry-after` to `oauth-limit-no-ts` regardless of `perMinutePresent`.

**M3 · perf-observability · `usage-limit.ts:136-212` · ref:c21fdb1 · file-state**
No per-step logging/trace records which branch fired, so an M-key-risk-style misroute is invisible without re-deriving from raw headers. The ~12 `getHeader` reads per classify are negligible (error-only path). **Fix:** attach the winning step id to the returned classification or emit a debug trace event.

### 🔑 KEY RISK — low tier by confidence, high by relevance

**K1 · correctness/spec · `usage-limit.ts:156, 174-185` · ref:c21fdb1 · verified mechanism / UNVERIFIED trigger**
Step 2 (unified) is checked before Step 3 (per-minute). If a real OAuth **subscription-cap** 429 carries per-minute headers but **not** unified headers, it routes to `rate-limit-transient` (≤3 bounded silent retries, then a raw error surfaces) instead of the pre-PR `oauth-limit-no-ts` **pause + hot-swap** UX — i.e. it reintroduces, for that sub-case, the exact regression the magnitude gate originally fixed. The mechanism is verified; whether the trigger occurs depends entirely on **the PR's own documented open question** (are unified headers actually present on real subscription 429 `error.headers`? — community-observed, not doc-guaranteed). Downgraded to *low* by that low confidence, but it is the single most decision-relevant item: **the feature's correctness rests on a real-world assumption no telemetry has yet confirmed.** **Fix:** prioritize the author's promised `error.headers` capture on a live subscription 429 as the real acceptance gate; add a test with a synthetic subscription fixture.

### ⚪ LOW

- **L1 · test-coverage · `usage-limit.test.ts` · verified-absent** — the 14 new cases miss: unified+per-minute both present (step-2-wins precedence), per-minute + no `retry-after` (→ `retryAfterMs: undefined`), per-minute + long `retry-after` (M2), and a past/huge `unified-reset` (M1/L2). Add these.
- **L2 · correctness · `usage-limit.ts:163-166`** — a past `unified-reset` yields an `oauth-limit` with a past `resetsAt`; benign (downstream deadline collapses to ~now → immediate replay) and exact parity with the `|ts` path. Follow-up only.

## Clean dimensions

- **API-COMPAT — no issues.** `getHeader` is a purely additive export (zero prior symbol, zero importers). `parseRetryAfterMs` keeps its signature and performs the identical operation sequence via `getHeader` (extract-function refactor). `UsageLimitClassification` union unchanged. All three production consumers verified compatible: `query/retry-layer.ts` (matches all three kinds), `telegram/error-utils.ts` (re-export untouched), `tracing-fetch.ts` (`parseRetryAfterMs` unchanged).
- **SECURITY — no issues** beyond M1's robustness edge. Header values reach only `Number()` / `Date.parse()` / `split('|')`, all `isFinite`/`isNaN`-guarded; no regex/ReDoS surface; unified header values are compared `!== undefined` only, never logged or propagated; account/token hot-swap is gated by keychain byte-comparison independent of header content — no auth-bypass from spoofed headers.

- Wave 1 agents read the two changed files from **inline reviewed content at ref `c21fdb1`** (working tree is on `main`, so on-disk copies of those two files are stale); unchanged consumers (`retry-layer.ts`, `telegram/error-utils.ts`, `tracing-fetch.ts`) were read live from disk (accurate — unmodified by this PR). All blocking-tier citations were verified against `c21fdb1` via `git show`.
- **Stated intent: PR #490 title + body — spec-compliance assessed.** The additive/presence-gated claim holds structurally; Step 1's own pre-PR identity was taken from the pasted content, not re-diffed against `main`.
- **Tests were not executed.** The PR body claims 48/48 + 26/26 pass and clean `tsc`; I verified the code, citations, and coverage gaps statically, not by running the suite.
- Did not review the Telegram/CLI rendering of the `paused`/`resumed` events, nor `waitForReset`/`waitForHotSwap` internals (unchanged).
- No verified critical/high findings → the mandatory `/shadow-verify` gate did not trip; the framework's shadow-verify nudge was considered and declined per the skill's medium-and-below pass-through rule (the two load-bearing findings were instead re-derived directly via git in Wave 1.5).

## Merge decision

## ✅ MERGE — with eyes open on K1

No verified blocker. The change is additive, presence-gated, and provably identical to current behavior in the no-headers case; api-compat and security are clean; the medium findings are either pre-existing pattern parity (M1), bounded and low-materiality (M2), or observability polish (M3). Recommended as **fast-follow, not blockers**: (1) treat the author's promised real-`error.headers` telemetry as the true acceptance gate for K1 — the feature's benefit and its one regression risk both hinge on that unverified assumption; (2) add a shared Date-overflow guard covering both the `|ts` and `unified-reset` paths (M1); (3) backfill the four verified-absent edge tests (L1). If the team wants *zero* regression risk before telemetry lands, the conservative option is to make Step 3 defer to the magnitude fallback on a long `retry-after` (M2's fix), which removes K1's bounded downside.

*Read-only review — no files, PR, or branch state were modified.*

---

PR #497, opened 2026-07-11
fix(plan-mode): mirror exit_plan_mode's deferred flip onto stats.permissionMode (#495)

### LINKED ISSUE #495: Bug: exit_plan_mode reports bypassPermissions but hook state never flips
## Bug description

Calling `exit_plan_mode` returns `{ mode: "bypassPermissions" }` indicating success, but the PreToolUse hook state never actually transitions out of plan mode. Both the agent side and the operator side remain stuck in plan mode — all write operations (write_file, bash mutations, gh issue create) continue to be blocked by the hook with:

```
plan mode: write_file is refused. Use /plan off to exit plan mode.
```

The tool reports the state changed; the hook disagrees. Neither side reflects the bypass.

## Reproduction

1. Enter plan mode (via `/plan` or automatic trigger)
2. Attempt a write operation — correctly blocked by hook
3. Call `exit_plan_mode` tool
4. Tool returns `{ mode: "bypassPermissions" }` — appears to succeed
5. Attempt the same write operation — **still blocked** by the hook
6. Operator side also still shows "plan mode" — state never flipped on either side

## Expected behavior

`exit_plan_mode` should atomically transition the hook's internal state so that subsequent tool calls see the bypass. If the tool reports success, the hook must agree.

## Actual behavior

The tool's return value and the hook's enforcement state are decoupled. The tool writes to some in-memory flag, but the PreToolUse hook reads from a different source (or caches the old value), so the transition never propagates.

## Impact

- Agent is stuck and must ask the operator to manually type `/plan off` in the REPL
- The tool's success response is misleading — it claims bypass was granted when it wasn't
- Wastes a full turn (or more) of back-and-forth before the mismatch is diagnosed

## Suggested investigation

- Where does `exit_plan_mode` write the bypass flag?
- Where does the PreToolUse hook read the plan-mode state?
- Are these the same source of truth, or is there a stale cache / separate flag?
- Is there a race where the hook snapshots the state before the tool's write is visible?

## Environment

- Observed in CLI session, agent-afk repo
- The `exit_plan_mode` tool returned `mode=bypassPermissions` but the hook kept blocking `write_file` and `bash` (state-mutating commands)

griffinwork40:
Fixes #495.

## Problem
`exit_plan_mode` returns `mode=bypassPermissions` (looks like success), but the PreToolUse plan-mode gate keeps refusing `write_file` / mutating `bash`, and the operator's REPL prompt stays stuck showing "plan". Only a manual `/plan off` clears it. Observed live.

## Root cause
Plan-mode state lives in **two** places:
- the **session's** own mode — `AgentSession` metadata + provider (`setPermissionMode`)
- **`stats.permissionMode`** — the mirror that BOTH enforcement points read: the gate (wired at `bootstrap.ts:628` as `() => stats.permissionMode`, read fresh at call time in `plan-mode-gate.ts:48`) and the REPL prompt (`loop-iteration.ts:227`).

`/plan off` updates **both** (`plan-mode-toggle.ts:62-63`). But `exit_plan_mode`'s deferred flip — applied post-turn in `takePendingPlanExitSeed()` — updated only the **session** side. The REPL drain promoted the implement-seed but never mirrored the mode onto `stats.permissionMode`, so the gate stayed plan-locked and the prompt never flipped. One missing write; both symptoms. (Not a race — the gate reads the mode fresh; it just read the stale mirror.)

## Fix
`takePendingPlanExitSeed()` now returns `{ message, mode }` — it already applied `seed.mode`, so it returns the exact applied value with no divergence possible — and the REPL drain mirrors it: `ctx.stats.permissionMode = planExit.mode`, exactly as `togglePlanMode` does for `/plan off`. Chosen over adding a `getPermissionMode()` getter so the mirror is guaranteed equal to the applied value.

## Verification
- `pnpm test`: **11184 passed, 0 failed**
- New regression test in `loop-iteration.test.ts` ("…mirrors the applied mode onto stats (#495)"): passes with the fix, **fails without it** (`expected 'plan' to be 'bypassPermissions'`), verified by reverting the one drain line. Closes a real coverage gap — no prior test exercised the `exit_plan_mode → drain → stats` path.

## Scope / follow-up
Scoped to the observed REPL surface. A parallel sibling-bug sweep is still in flight (does the Telegram surface keep a separate permission-mode mirror with the same decoupling? any other deferred flips that skip the `stats` mirror?); any finding will be filed as a separate issue/PR.

### DIFF
--- a/src/agent/session/agent-session.ts
+++ b/src/agent/session/agent-session.ts
@@ -947,13 +947,21 @@ export class AgentSession implements IAgentSession {
    * Return and CLEAR the pending plan-exit implement-turn queued by an approved
    * `exit_plan_mode` call, or `undefined` if none is pending. The REPL drains
    * this at the top of each input-loop iteration (post-turn) and, when set,
-   * atomically applies the deferred permission-mode flip then returns the seed
-   * message to be auto-submitted as a fresh user turn. Single-shot: a second call
-   * returns `undefined` until the next approval.
+   * atomically applies the deferred permission-mode flip then returns BOTH the
+   * seed message and the flipped-to mode. Single-shot: a second call returns
+   * `undefined` until the next approval.
    *
    * The mode flip is applied HERE (not in the handler) so the gate stays locked
    * in plan mode for the entire model turn and only opens at this clean
    * post-turn boundary — closing the mid-turn TOCTOU window.
+   *
+   * Invariant (#495): the flip updates the session's OWN mode (metadata +
+   * provider), but the plan-mode gate and the REPL prompt read a SEPARATE mirror
+   * — `stats.permissionMode` — which the session cannot reach. So the applied
+   * `mode` is returned alongside the message; the REPL drain
+   * (`loop-iteration.ts`) mirrors it onto `stats.permissionMode`, exactly as
+   * `togglePlanMode` does for `/plan off`. Returning it (rather than exposing a
+   * getter the caller re-reads) guarantees the mirror equals the applied value.
    *
    * If the deferred flip rejects (e.g. the provider's query handle is closing —
    * the same failure mode `togglePlanMode` guards for `/plan off`), the seed is
@@ -963,7 +971,7 @@ export class AgentSession implements IAgentSession {
    * (which has no try/catch around this drain) and crash it. The model stays in
    * plan mode and can retry `exit_plan_mode`.
    */
-  async takePendingPlanExitSeed(): Promise<string | undefined> {
+  async takePendingPlanExitSeed(): Promise<{ message: string; mode: PermissionMode } | undefined> {
     const seed = this._pendingPlanExitSeed;
     this._pendingPlanExitSeed = undefined;
     if (seed === undefined) return undefined;
@@ -975,7 +983,7 @@ export class AgentSession implements IAgentSession {
       );
       return undefined;
     }
-    return seed.message;
+    return { message: seed.message, mode: seed.mode };
   }
 
   /**
--- a/src/agent/tools/handlers/exit-plan-mode.test.ts
+++ b/src/agent/tools/handlers/exit-plan-mode.test.ts
@@ -216,8 +216,10 @@ describe('AgentSession plan-exit seed bridge', () => {
     // Pass the approved mode alongside the seed message (new signature).
     controls!.requestImplementSeed('SEED-MSG', 'default');
 
-    // takePendingPlanExitSeed is now async — it applies the mode flip then returns the message.
-    expect(await session.takePendingPlanExitSeed()).toBe('SEED-MSG');
+    // takePendingPlanExitSeed applies the mode flip then returns BOTH the seed
+    // message AND the flipped-to mode (#495) — the caller mirrors `mode` onto
+    // stats.permissionMode so the gate + prompt unlock.
+    expect(await session.takePendingPlanExitSeed()).toEqual({ message: 'SEED-MSG', mode: 'default' });
     // Single-shot: drained value is cleared.
     expect(await session.takePendingPlanExitSeed()).toBeUndefined();
 
--- a/src/agent/types/session-types.ts
+++ b/src/agent/types/session-types.ts
@@ -235,12 +235,14 @@ export interface IAgentSession {
   /**
    * Return and CLEAR any implement-turn queued by an approved `exit_plan_mode`
    * tool call. Atomically applies the deferred permission-mode flip (closing the
-   * mid-turn TOCTOU window) then returns the seed message. The REPL drains this
-   * post-turn and auto-submits the message as a fresh user turn (reproducing
+   * mid-turn TOCTOU window) then returns BOTH the seed message AND the mode it
+   * flipped to. The REPL drains this post-turn: it mirrors `mode` onto
+   * `stats.permissionMode` (the value the plan-mode gate and prompt read — see
+   * #495) and auto-submits `message` as a fresh user turn (reproducing
    * `/plan off`'s save-and-implement handoff). Returns `undefined` when nothing
-   * is pending.
+   * is pending (or when the deferred flip rejected and the seed was dropped).
    */
-  takePendingPlanExitSeed(): Promise<string | undefined>;
+  takePendingPlanExitSeed(): Promise<{ message: string; mode: PermissionMode } | undefined>;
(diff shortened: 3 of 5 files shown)


griffinwork40:
I now have complete, citation-grounded coverage across every dimension. The security-critical branch is confirmed clean: the handler returns early on both **reject** (`result.action !== 'accept'`, line 151) and **keep** (`picked.startsWith('Keep')`, line 167) *without* calling `requestImplementSeed` — so `_pendingPlanExitSeed` is set **only** on an explicit approve, and the mirrored `seed.mode` is always either the restored pre-plan mode or an operator-picked bypass escalation (line 175).

I performed this as an orchestrator-run light-regime review with inline citation + absence verification against the reviewed ref (`git show 087b3064:…` / `git grep … 087b3064`) rather than forking a redundant sub-agent for an 82-line hotfix. Every load-bearing claim below was read at the reviewed ref. No critical/high findings surfaced, so `/shadow-verify` is not triggered per the skill's post-synthesis rule.

---

# Code Review — PR #497

**Target:** PR #497 `fix(plan-mode): mirror exit_plan_mode's deferred flip onto stats.permissionMode (#495)` · `afk/issue-495-exit-plan-mode-stats-mirror` → `main`
**Reviewed ref:** `087b3064ef390342377b41991ddf32da1df25aac` (mergeable, open) — all citations verified at this ref
**Change type:** hotfix · **Regime:** light (82 lines, 5 files: 2 prod / 2 test / 1 interface)
**Stated intent:** PR #497 body + issue #495 — spec-compliance assessed

## Summary
Plan-mode state lives in two places: the session's own mode (metadata + provider) and `stats.permissionMode`, the mirror that **both** enforcement points read — the PreToolUse gate (wired `bootstrap.ts:628` → `() => stats.permissionMode`, read fresh in `plan-mode-gate.ts` via `if (getMode() !== 'plan')`) and the REPL prompt (`loop-iteration.ts` `buildPrompt(ctx.stats.permissionMode)`). `/plan off` writes both (`plan-mode-toggle.ts:62-63`); `exit_plan_mode`'s deferred flip wrote only the session side, so the gate stayed plan-locked and the prompt never flipped despite a reported success. The fix threads the applied mode out of `takePendingPlanExitSeed()` as `{ message, mode }` and mirrors it in the drain (`ctx.stats.permissionMode = planExit.mode`), reproducing `togglePlanMode`'s two-write pattern. The diagnosis is accurate and the fix is minimal, correct, and complete for its stated scope.

## Findings by dimension

**security · api-compat** — `no issues found`
- **security** — read `exit-plan-mode.ts:151-183` @ `087b3064`: `requestImplementSeed` is reached **only** on approve+non-keep; reject (`action !== 'accept'`) and keep both return early without seeding. The mirrored `seed.mode` is always the operator-approved value (restore pre-plan `prevMode`, or `bypassPermissions` only when the picker choice `includes('bypass')`, line 175). The fix cannot unlock the gate to an unapproved mode — it makes an already-approved mode take effect. Net security-positive. `ref:087b3064 · citation-type:file-state`
- **api-compat** — `takePendingPlanExitSeed` return type changed `Promise<string|undefined>` → `Promise<{message,mode}|undefined>`. `git grep 087b3064` confirms exactly **one** production caller (`loop-iteration.ts:209`, updated) + the interface (`session-types.ts:245`, updated). Test mocks at `loop-iteration.test.ts:130` and `repl-loop-wiring.test.ts:125` return `undefined` (assignable to the new union); the real-session test was updated. No unhandled caller. `ref:087b3064 · citation-type:file-state`

**correctness · spec-compliance · test-coverage · perf-observability** — 1 low finding
- **correctness** — `no issues found`. `agent-session.ts:979,986` @ `087b3064`: `setPermissionMode(seed.mode)` applies the flip, then returns `mode: seed.mode` — same value, so mirror ≡ applied with no re-read divergence (the PR's stated design reason for returning the value over a getter). Drain ordering (`loop-iteration.ts:213-215`) mirrors onto `stats` **before** the implement turn's prompt is built and auto-submitted, so the gate is unlocked for that turn. Failure semantics preserved: a rejecting `setPermissionMode` returns `undefined` from the drain guard `if (planExit !== undefined)`, leaving `stats.permissionMode` untouched (stays `plan`) — consistent with the pre-existing try/catch, which this PR did not alter.
- **spec-compliance** — satisfies stated intent: the fix directly targets the diagnosed root cause (gate/prompt read `stats.permissionMode`; deferred flip skipped it). Telegram surface is explicitly deferred in the PR body ("sibling-bug sweep still in flight"); `git grep` confirms Telegram has no `takePendingPlanExitSeed` caller, so there is no analogous REPL-style drain to fix here. No scope creep — diff is type-thread + one write + docs + one test.
- **test-coverage** — `low · confidence:high · loop-iteration.test.ts:97-130 · ref:087b3064 · citation-type:file-state` — The new regression test covers only the happy path; no test asserts the drain leaves `stats.permissionMode` unchanged when `takePendingPlanExitSeed()` returns `undefined` (the reject path). *Evidence:* the new test always returns a defined seed (`return { message: 'IMPLEMENT-SEED', mode: 'bypassPermissions' }`). *Suggestion:* add a sibling case with the mock returning `undefined` and assert `ctx.stats.permissionMode` stays `'plan'`. Non-blocking — the guard is trivial and the session-level reject is already covered at `exit-plan-mode.test.ts:255-257`.
- **perf-observability** — `no issues found`. One scalar field write on a post-turn boundary; no hot-path impact. The reject-path warning log in `takePendingPlanExitSeed` is preserved.

## What was not checked
- Read against reviewed ref `087b3064` (branch HEAD, == working tree for all 5 PR files; the 8 dirty files are untracked and non-overlapping) — citations verified at that ref via `git show`/`git grep`.
- Stated intent: PR #497 body + issue #495 — spec-compliance **assessed**.
- Did not run the test suite (PR reports `11184 passed`; not independently executed). Did not review the Telegram surface behavior (out of PR scope, no relevant caller exists). Did not exercise the live picker→drain→gate flow at runtime.

## Merge decision: **MERGE**

The change correctly and minimally fixes the diagnosed decoupling, is security-clean (mirrors only operator-approved modes), preserves failure semantics, updates the one production caller + interface, and adds a regression test that fails without the fix. The single `low` finding (reject-path drain coverage) is optional polish, not a merge blocker.

---

PR #504, opened 2026-07-11
fix(cli): keep the update-notifier fresh instead of ≤24h stale

griffinwork40:
## Problem

The startup **"update available"** banner under-reports: it frequently stays silent when a newer version *is* published.

## Root cause

The banner is rendered purely from a cached `latestVersion` (`src/cli/update-checker.ts`), with three compounding staleness mechanisms:

1. **≤24h blind spot** — `checkForUpdates()` only spawns a background registry refresh when the cache is missing or **>24h old**. Any release published inside that window is invisible until it elapses.
2. **Always one run behind** — the refresh runs in a *detached* child that writes the cache **after** the current process already read it and returned, so a freshly-fetched version is never seen by the run that fetched it (and a cold cache returns `null` outright).
3. **`afk update` left the cache stale** — the update command fetches latest out-of-band and writes only the pending marker, never the notifier cache.

With this repo shipping several releases a day (e.g. `5.28.0 → 5.29.1` = 6 releases in ~15h), a 24h TTL simply can't track the cadence.

## Fixes

- **A — TTL 24h → 3h** (`update-checker.ts`): the passive banner now tracks a several-releases-per-day cadence while still bounding background registry hits to at most one per window.
- **B — refresh cache on `afk update` / `--check`** (new `writeUpdateCache()`, called from `commands/update.ts`): a manual update keeps the notifier honest immediately. Guarded so an explicit `--pin` (which may target an older version) can't poison the cache.
- **C — cold-start inline fetch** (new `coldStartUpdateCheck()` + `hasUpdateCache()`, wired in `cli/index.ts`): on the first run with no cache, `await` one bounded **800ms** registry fetch so the banner appears on that first run instead of only on a second invocation. Warm cache keeps the fully-synchronous fast path; the same `off`/`CI`/`NO_UPDATE_NOTIFIER` guards apply; on fetch failure it falls back to the detached background refresh.

## Verification

- `pnpm lint` (strict `tsc --noEmit`): clean
- Full suite: **11,214 passed**, 0 failed (14 pre-existing skips)
- **+13 new unit tests**: TTL boundary (×2), `writeUpdateCache` (×3), `hasUpdateCache` (×2), `coldStartUpdateCheck` (×6, incl. off/CI/NO_UPDATE_NOTIFIER guards); plus `writeUpdateCache` assertions folded into the `afk update` tests
- CI gates: `audit:env:check` 0 violations, `scan:env:check` in sync

### DIFF
--- a/src/cli/commands/update.test.ts
+++ b/src/cli/commands/update.test.ts
@@ -22,6 +22,7 @@ vi.mock('../version.js', () => ({
 vi.mock('../update-checker.js', () => ({
   fetchLatestVersion: vi.fn(),
   writePendingUpdateMarker: vi.fn(),
+  writeUpdateCache: vi.fn(),
 }));
 
 vi.mock('child_process', () => ({
@@ -39,12 +40,13 @@ vi.mock('../palette.js', () => ({
 }));
 
 import { spawn } from 'child_process';
-import { fetchLatestVersion, writePendingUpdateMarker } from '../update-checker.js';
+import { fetchLatestVersion, writePendingUpdateMarker, writeUpdateCache } from '../update-checker.js';
 import { registerUpdateCommand } from './update.js';
 import { EventEmitter } from 'events';
 
 const mockFetchLatestVersion = vi.mocked(fetchLatestVersion);
 const mockWritePendingUpdateMarker = vi.mocked(writePendingUpdateMarker);
+const mockWriteUpdateCache = vi.mocked(writeUpdateCache);
 const mockSpawn = vi.mocked(spawn);
 
 // ---------------------------------------------------------------------------
@@ -111,6 +113,8 @@ describe('afk update', () => {
       expect(output).toContain('Update available');
       expect(output).toContain('1.10.1');
       expect(output).toContain('1.12.0');
+      // B: --check refreshes the notifier cache with what it learned.
+      expect(mockWriteUpdateCache).toHaveBeenCalledWith('1.12.0');
       expect(process.exitCode).toBeUndefined();
     });
 
@@ -179,6 +183,9 @@ describe('afk update', () => {
         expect.objectContaining({ stdio: 'inherit' }),
       );
       expect(mockWritePendingUpdateMarker).toHaveBeenCalledWith('1.11.0');
+      // B: an explicit --pin must NOT touch the notifier cache — it may target
+      // an older-than-latest version and would otherwise poison the banner.
+      expect(mockWriteUpdateCache).not.toHaveBeenCalled();
       const output = consoleSpy.mock.calls.flat().join('\n');
       expect(output).toContain('installed');
       expect(process.exitCode).toBeUndefined();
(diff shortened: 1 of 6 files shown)


griffinwork40 (commented):

**Merge decision: MERGE** ✅

Read-only review of this branch at `5ffec3d`. Six risk points were checked against the reviewed ref; all pass. Zero blocking/critical/high findings.

**What was verified (all clean):**
- **security** — `writeUpdateCache()` gates on an anchored `SEMVER_RE` (`^\d+\.\d+\.\d+(-[\da-z.]+)?$`, `update-checker.ts:209`); `fetchLatestVersion` also validates `pkg.version` before returning; `--pin` is validated pre-shell (`update.ts:72`). No unvalidated string reaches `writeFileSync` or a spawned process.
- **correctness — Fix B poison guard** (highest-value invariant): in the install path, `writeUpdateCache(target)` (`update.ts:91`) is inside `if (target === undefined)` (lines 80–96), so an explicit `--pin` — which may target an older-than-latest version — can never poison the notifier cache. Guard holds. The `afk update --check` path writes the cache on **both** up-to-date and update-available outcomes (`update.ts:61`, before the `isNewer` split), matching the stated intent.
- **correctness — cold-start wiring** — `await coldStartUpdateCheck(...)` (`index.ts:238`) sits inside the `(async () => { … })()` IIFE (opened line 218); no top-level-await hazard.
- **perf** — the new cold-start inline fetch is bounded (`fetchLatestVersion(COLD_START_TIMEOUT_MS=800ms)` honored via `req.setTimeout`), and the `off`/`CI`/`NO_UPDATE_NOTIFIER` guards run *before* the fetch, so opt-out/CI paths pay zero added latency.
- **api-compat** — the `TWENTY_FOUR_HOURS → CACHE_TTL_MS` rename is module-private (grep at ref → zero production importers); the three new exports (`hasUpdateCache`, `coldStartUpdateCheck`, `writeUpdateCache`) are additive.
- **spec-compliance** — Fixes A (TTL 24h→3h), B (cache refresh + `--pin` poison guard), and C (cold-start inline fetch) all map to the stated intent.
- **test-coverage** — TTL boundary, `writeUpdateCache` (valid / injection-rejected / throw-swallowed), `hasUpdateCache`, and all `coldStartUpdateCheck` guard + fallback branches are covered, plus the `--pin`-must-not-write assertion.

**Minor (nit, non-blocking):**
- No dedicated `coldStartUpdateCheck` **timeout** test — the failure→null→spawn-fallback path is exercised only via the 500-status sibling; timeout is proven one level down against `fetchLatestVersion`. Optional: add a coldStart timeout case that wires `setTimeout`→cb to prove the same fallback under timeout.

**Not checked by this automated pass:** the test suite / lint were not re-run here (the PR reports 11,214 passing + clean lint — CI is green on this head); dependency internals (`paths.js`/`version.js`/`env.js`) and the pre-existing `spawnBackgroundCheck` inline-script surface were out of diff scope.

PR #505, opened 2026-07-11
fix(tui): route commitAbove on the real measured frame top, not the shrink-padded one

griffinwork40:
## Summary

Fixes the "two boxes + blank void" table rendering bug: a markdown table followed by streamed prose renders with border-less blank rows punched into the table, and the missing content is silently lost. Reproduces reliably in a live terminal (`afk i` at 120×30 rendering a wrapped table followed by paragraphs).

## Root cause

`commitAbove`'s `decideCommitMode` routing computes `prevTopRow` from `self.logUpdate.topRow`, which reports the transient shrink-**padded** frame top. Right after a frame shrink that value reads **lower** than the real top, so `fitsAboveFrame` is computed too small and a block that actually **fits** the above-frame region is routed into `band-hold`. Band-hold then overwrites prior repositioned content without archiving it to scrollback → permanent loss (the void). The downstream symptom is the cap-drop in `committed-band-commit.ts` (`capped = run.slice(run.length - maxRun)`), which drops rows on the false premise that they already reached scrollback.

This is documented by the existing `terminal-compositor.commit-geometry.test.ts` (which guards the band-exists shrink-pad scenario) and the History note at `terminal-compositor.frame.ts:191-198`.

## Fix

`repaint()` already computes the true unpadded frame top via `CupFrameRenderer.measure(frame, targetBottomRow).topRow` — the exact physical row Phase-2 re-establishes. This PR:

- stores it in a new field `lastMeasuredFrameTop` (written in `repaint()`),
- routes `commitAbove` on `Math.max(logUpdate.topRow, lastMeasuredFrameTop, bandFloor)` for `prevTopRow`.

`Math.max` means `prevTopRow` can only be **raised** toward the true top, never lowered; when `lastMeasuredFrameTop` is `0` (no repaint yet) or stale (post-SIGWINCH — gated on `!bandGeometryStale`) it reduces **exactly** to the prior expression, which is why it is regression-safe.

This is the targeted fix recommended by the compositor design doc's own Verification Addendum (over the larger declarative rewrite).

## Verification

- `tsc --noEmit` clean.
- Full compositor + commit-mode + cup-frame-renderer suite: **374/374 green** (no regression).
- Live `afk i` (haiku, 120×30) rendering a markdown table + streamed prose: **0 border-less blank rows** across two runs of differing table size — was 4–5 before the fix.

## Known coverage gap

No new **deterministic** regression test for the streaming case: the void is only reproducible in a live TTY (synthetic headless commits don't model the streaming re-render + overlay regrowth, which is why the bug persisted). `commit-geometry.test.ts` guards the adjacent shrink-pad scenario. A faithful streaming `@xterm` regression is recommended as follow-up.

### DIFF
--- a/src/cli/terminal-compositor.committed-band-commit.ts
+++ b/src/cli/terminal-compositor.committed-band-commit.ts
@@ -37,6 +37,9 @@ export interface CommittedBandHost {
   committedBand: string[];
   committedBandTopRow: number;
   committedBandBottomRow: number;
+  /** Real unpadded frame top from the last repaint's measure() — see the field
+   *  doc on the class. Routing (prevTopRow) trusts this over logUpdate.topRow. */
+  lastMeasuredFrameTop: number;
   /** How many of committedBand's rows (its bottom suffix) are painted on screen. */
   committedBandPaintedRows: number;
   /** Memoization for reflowCommittedBandToWidth — see the field doc on the class. */
@@ -227,10 +230,26 @@ export function commitAbove(self: CommittedBandHost, text: string): void {
   // is exactly the "genuinely no known frame top" case BLOCKER-1 already
   // handles safely via band-hold.
   const rawLogUpdateTopRow = self.logUpdate.topRow ?? 0;
-  const prevTopRow =
+  // Prefer the real post-render frame top the last repaint captured via measure()
+  // (lastMeasuredFrameTop). logUpdate.topRow reports the transient shrink-PADDED
+  // top, which reads LOWER than reality after a frame shrink and makes
+  // decideCommitMode compute a wrong (too-small) fitsAboveFrame — routing a block
+  // that FITS into band-hold, which then overwrites repositioned prior content
+  // without archiving it ("two boxes + blank void" streaming-table bug;
+  // terminal-compositor.commit-geometry.test.ts + the History note at
+  // frame.ts:191-198). Take the MAX so prevTopRow can only be RAISED toward the
+  // true top, never lowered — the existing committedBandBottomRow+1 floor is
+  // preserved. Gated on !bandGeometryStale: after a SIGWINCH with no repaint
+  // since, lastMeasuredFrameTop (and the band floor) are PRE-resize values, so
+  // both are excluded and we fall through to rawLogUpdateTopRow — the BLOCKER-1
+  // band-hold safety path. When lastMeasuredFrameTop is 0 (no repaint yet) this
+  // reduces EXACTLY to the prior expression.
+  const measuredFrameTop = self.bandGeometryStale ? 0 : self.lastMeasuredFrameTop;
+  const bandFloor =
     self.committedBand.length > 0 && self.committedBandBottomRow > 0 && !self.bandGeometryStale
-      ? Math.max(rawLogUpdateTopRow, self.committedBandBottomRow + 1)
-      : rawLogUpdateTopRow;
+      ? self.committedBandBottomRow + 1
+      : 0;
+  const prevTopRow = Math.max(rawLogUpdateTopRow, measuredFrameTop, bandFloor);
   const frameTop = prevTopRow > 1 ? prevTopRow : Math.max(1, rows - 1 - extraRows);
   const anchorFloor = Math.max(self.anchorRow ?? 1, 1);
   // The block "fits" when every line can be CUP-painted at a row in
--- a/src/cli/terminal-compositor.frame.ts
+++ b/src/cli/terminal-compositor.frame.ts
@@ -72,6 +72,9 @@ export interface FrameHost {
   committedBand: string[];
   committedBandTopRow: number;
   committedBandBottomRow: number;
+  /** Real unpadded frame top; written here by repaint(), read by commitAbove's
+   *  routing. See the field doc on the class (terminal-compositor.ts). */
+  lastMeasuredFrameTop: number;
   committedBandPaintedRows: number;
   /** Memoization for reflowCommittedBandToWidth — see the field doc on the class. */
   bandReflowCache: BandReflowCache | null;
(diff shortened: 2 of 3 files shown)


griffinwork40:
# Code Review — PR #505

**`fix(tui): route commitAbove on the real measured frame top, not the shrink-padded one`**
Branch `afk/fix-tui-table-void` → `main` · reviewed ref `f9779a463e529d85a8947e867a38ba602147d250` · +40/-3, 3 files · **LIGHT regime**

## Summary

The fix routes `commitAbove` on a newly-tracked `lastMeasuredFrameTop` (gated by `bandGeometryStale`) instead of the shrink-padded `rawLogUpdateTopRow`, correcting the row at which committed content is written above the live frame — targeting the "two boxes + blank void" table-rendering bug. The change is small, directionally sound, and matches its stated intent. Findings are minor: one latent-invariant risk (no wrong-output bug reachable today), two narrow-scope correctness gaps, a disclosed test gap, and one stale documentation citation.

| Dimension | Outcome |
|---|---|
| security | ✅ clean — pure row-arithmetic, no auth/secret/RCE/injection surface |
| api-compat | ✅ clean — host-interface widening (`CommittedBandHost`/`FrameHost` gain a required field) is compile-safe; zero external implementers found |
| correctness | ⚠️ F1 (medium), F2 (low), F3 (low) |
| spec-compliance | ⚠️ F2, F5 — intent matched; one overstated safety claim, one stale citation |
| test-coverage | ⚠️ F4 (low) — no arm-cycle-boundary test; new streaming regression self-disclosed as untested |
| perf-observability | ✅ clean — one scalar write on an existing hot path; +2 reads / 1 `Math.max` in `commitAbove` |

## Findings (medium → low → nit)

**F1 · medium · correctness · `terminal-compositor.reset.ts:57-130`** (+`committed-band-commit.ts:90-95,247`, `lifecycle.ts:230,330`)
`resetState()` resets sibling per-cycle fields `bandGeometryStale` (`:98`) and `lastKnownRows` (`:99`) but never resets the new `lastMeasuredFrameTop` (absent from `ResetStateHost` and the function body). On a reused `TerminalCompositor` instance (confirmed reused across disarm/rearm via `_lib/stream-renderer.ts:359-388,788-859`), this field can retain a prior cycle's value while `bandGeometryStale` is freshly `false`. **Independently confirmed not reachable as a wrong-output bug today**: `commitAbove`'s own guard (`committed-band-commit.ts:90-95`) raw-writes and returns whenever `!armed`, and `arm()` issues an unconditional synchronous `repaint()` (`lifecycle.ts:330`) before any commit could read the stale value — both `arm()`/`disarm()` are fully synchronous with no intervening yield, and `resetState()` clears `pendingSubmissions`. Correctness rests entirely on that ordering invariant, which no test pins.
*Suggestion:* add `self.lastMeasuredFrameTop = 0;` beside `reset.ts:99` (and to `ResetStateHost`) as cheap defense-in-depth.

**F2 · low · correctness/spec-compliance · `committed-band-commit.ts:245-252`**
The no-band case can now be raised by `lastMeasuredFrameTop` even with no committed band — the PR body's "reduces EXACTLY to the prior expression" is true only when `lastMeasuredFrameTop === 0` or `bandGeometryStale`, not generally. Directionally safe, but an undisclosed scope expansion of the stated regression-safety claim.
*Suggestion:* amend the PR body/comment to scope the equivalence claim precisely.

**F3 · low · correctness · `terminal-compositor.frame.ts:127-130,278,349`**
`repaintPickerFrame` computes its own `desiredTopRow` (`:349`) but never writes `self.lastMeasuredFrameTop` (sole write site is `:278`, non-picker path) — a picker-mode repaint leaves the field stale.
*Suggestion:* mirror the write into `repaintPickerFrame` after `:349`.

**F4 · low · test-coverage · `commit-geometry.test.ts:99-100`**
Zero suite references to `lastMeasuredFrameTop`. Arm→disarm→re-arm-same-instance tests exist but none commit between rearm and the next repaint — the F1 invariant is unpinned by any test. New streaming regression is untested (author-disclosed in the PR body).
*Suggestion:* add a headless test asserting `lastMeasuredFrameTop` is correctly re-established across a disarm→rearm→commit sequence.

**F5 · nit · spec-compliance · `committed-band-commit.ts:239-240`**
Comment/PR body cites "History note at frame.ts:191-198" — stale; that range is actually the log-update pre-pad note, not the History note (`241-252`) or the shrink-pad rationale (`265-278`).
*Suggestion:* repoint the citation.

No critical/high findings. The fix is correct for its stated purpose, compile-safe, and performance-neutral. All findings are cheap, non-blocking follow-ups — F1 (the reset) and F5 (the citation) are worth a fast follow-up.

## Epistemic scope disclosure

- **Ref read:** all findings verified against live ref `f9779a463e529d85a8947e867a38ba602147d250` (not diff-context alone). A dedicated Wave 1.5 independently re-derived all 14 citations/absence-claims via a git-investigator sub-agent — 13 verified/confirmed-absent, 1 collateral citation path-corrected without weakening its parent finding. Zero findings dropped; F1's downgrade from a hypothesized `high` to `medium` was independently re-derived twice and held both times.
- **Intent:** PR title + body were available and spec-compliance was assessed against them (F2, F5).
- **Gaps:** did not run the test suite or `tsc` directly — relied on independent static file-state verification at the ref. Did not review REPL/Telegram/daemon integration paths beyond `terminal-compositor.*`/`commit-mode.ts`. Did not exhaustively trace `_lib/commit-coordinator.ts`'s queue internals (confirmed not to affect the armed-guard conclusion). Did not diff against the parent commit to independently re-derive the pre-fix ternary's exact old behavior — took the diff hunk's own context at face value (legitimate, since it's directly visible).

No files were edited, no git state was touched, no PR comment/review/merge was posted — this session only read via `git show`/`git grep`/`gh pr view` against the already-captured diff artifact and the fetched PR ref.

---

**Done**
- Reviewed PR #505 end-to-end via the review skill's light-regime pipeline (Wave 1 full-dimension review → Wave 1.5 independent citation/absence re-verification → Wave 2 synthesis), producing the merge-recommendation report above.
- Evidence: all citations verified against ref `f9779a463e529d85a8947e867a38ba602147d250` in the local repo clone (already fetched); no separate artifact file was written — the report above is the deliverable, and the input diff artifact remains at `/Users/griffinlong/.afk/state/skill-preflight/e890ab7f-bcdd-442d-beb7-45991d0e5c84/pr-505.diff`.
- Nothing in the world changed: this skill is strictly read-only by contract — no edits, commits, pushes, or PR mutations occurred (confirmed working tree still shows only the pre-existing 4 unrelated dirty files noted in preflight).
- Nothing deferred on my end; recommended follow-ups (F1, F5 in particular) are named above for the operator/PR author to action if desired — merging or addressing findings is explicitly out of scope for this skill.

---

Issue #355, opened 2026-07-01
shadow-verify nudge fires on shadow-verify's own verifier output (self-trigger loop)


griffinwork40:
## Summary

The shadow-verify nudge (`src/agent/shadow-verify-nudge.ts`, a built-in `SubagentStop` hook) repeatedly fires on the output of **shadow-verify verifier sub-agents themselves**, nudging the parent to shadow-verify its own verification wave. Observed in a live session: after dispatching 4 parallel verifier sub-agents (per the shadow-verify skill), the nudge fired on **three consecutive turns**, each time flagging the verifiers' verdict tables as "decision-driving findings."

## Root cause (two gaps)

1. **`VERIFIED_ORCHESTRATORS` matches by agentType/idPrefix, but manually dispatched verifiers don't carry the magic name.** `fromVerifiedOrchestrator()` (line ~87) tests `context.agentType` (= `effectiveAgentType ?? idPrefix`) against tokens like `shadow-verify`. The shadow-verify *skill* instructs the model to dispatch verifiers via raw `agent` calls, and nothing forces the id_prefix to contain `shadow-verify` — e.g. `sv-claim1` or `verify-x` sails past the suppression.

2. **`VERIFIER_SIGNATURES` (line ~53) is too narrow for real verifier output.** It matches `verifier_verdict`, `"claim":`, `re-derived ... independent`, `independently (re-derived|re-verified|verified|checked)`, `verifier (agrees|disagrees|confirms|refutes)`. But a typical verifier reply is a markdown verdict table (`| Claim | Result | Evidence |` with `CONFIRMED`/`REFUTED` rows) that hits none of these — while hitting many `DECISION_MARKERS` (`verdict`, `CONFIRMED`, `| status |` table headers), so it clears the `MIN_MARKER_HITS = 2` gate easily. Ironically, the more structured the verifier's output, the more certainly it triggers the nudge.

A third, secondary observation: the nudge re-fired on later turns with **no new sub-agent completing in between**, suggesting either per-child injection queuing that outlives the turn or duplicate injection of the same SubagentStop context. Worth checking the injectContext delivery path while in there.

## Suggested fixes (any subset)

- Broaden `VERIFIER_SIGNATURES` to cover verdict-table style: e.g. `/\b(CONFIRMED|REFUTED|UNVERIFIABLE)\b/` appearing ≥2 times, or `/\|\s*(claim|verdict)\s*\|/i` combined with `evidence`.
[...]

PR #506, opened 2026-07-11
feat(cli): app-like TUI — /config settings menu + /model & /resume pickers

griffinwork40:
## Summary

Make the interactive REPL feel more "app-like" — a real in-session settings menu and arrow-key pickers for the highest-traffic commands — by **reusing the existing overlay primitives** rather than adopting a TUI framework (Ink/blessed). No new stdin ownership, no render-layer rewrite.

## What changed

- **`/config` — interactive settings menu** (new `src/cli/render/config-menu.ts`). Bare `/config` on a TTY opens a `category → key → value` navigator composed from `runPicker` + `runTextInput`. Writes go through the existing `setConfigValue` engine and echo the restart note (config is cached at load). Adds scriptable `/config view` and `/config set <key> <value>` fast-paths; non-TTY surfaces (Telegram/daemon) fall back to the read-only dump.
- **`/model` — picker.** Bare `/model` on a TTY opens an arrow-key picker over the tier aliases (cursor pre-placed on the current model); selection switches live mid-session. `/model <id>` (alias / slot / HF id) and non-TTY behaviour unchanged.
- **`/resume` — picker.** Bare `/resume` on a TTY opens a picker of recent saves; selecting one resumes it directly (the same-session data-loss guard is preserved). This finally delivers the behaviour the command's own hint already advertised. `/resume <id>` and the non-TTY table unchanged.
- **Slash palette — fuzzy matching.** `filterSlashCandidates` now ranks prefix matches first, then appends subsequence matches (`/cfg` → `/config`), case-insensitive. The common prefix path is unchanged.

## Approach & key invariants

All three menus are additional consumers of `TerminalCompositor.enterPickerMode` (joining `runPicker` and `runTextInput`). Overlays are opened via awaited, **never-nested** calls, so the single-overlay guard and the single-consumer-stdin invariant (PR #511) hold. Each command keeps its text-arg and non-TTY fallbacks.

## Notable decisions

- **Secrets are never editable in the menu** — config keys are only `agent`/`human` tier, never `secret` (secrets are env-only), so nothing routes a credential into scrollback.
- **Human-tier keys require an explicit in-menu confirm** before writing (mirrors `afk config set --allow`); `/config` is a human surface (the agent cannot type slash commands), so this does not widen the `config_set` agent-tool path.
- **Restart semantics surfaced honestly** — every config write echoes "effective on the next restart"; `/model` / `/resume` act live.

## Testing

- Full suite green: **11,215 passed / 14 skipped** (`vitest run`).
- `tsc --noEmit` (strict) clean; no new `process.env` reads and no new SDK symbols, so the env/SDK audit gates are unaffected.
- New coverage: `config-menu.test.ts` (15), `info.test.ts` (4), +6 `/config` fast-path tests, +3 `/resume` picker tests, +4 `filterSlashCandidates` tests. Interactive flows are driven via a scripted `FakePickerHost`.

## Deferred / follow-ups

- Dropdown **visual** polish (inline summaries in the `/` dropdown, framed chrome, adaptive row count) — wants a look on a real terminal; intentionally not in this PR.
- Other text-only commands that could adopt the picker later: `/thinking`, `/mcp`, `/save`, `/worktree`.

## Verification note

Menu logic is covered by unit tests + the already-tested `runPicker` / `runTextInput`, but I have **not** driven a live TTY keypress session (headless env). A ~30-second manual spin in `afk interactive` (`/config`, `/model`, `/resume`) is recommended before merge.

### DIFF
--- a/src/cli/input/trigger.test.ts
+++ b/src/cli/input/trigger.test.ts
@@ -16,7 +16,7 @@ import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import { mkdirSync, writeFileSync, rmSync, existsSync } from 'fs';
 import { join } from 'path';
 import { tmpdir } from 'os';
-import { detectTrigger, filterFileCandidates } from './trigger.js';
+import { detectTrigger, filterFileCandidates, filterSlashCandidates } from './trigger.js';
 import { MAX_FILE_MATCHES } from '../multi-line-reader.js';
 import { resetRegistry } from '../slash/registry.js';
 import { registerAll } from '../slash/index.js';
@@ -128,5 +128,33 @@ describe('filterFileCandidates', () => {
 
   it('unreadable scan dir yields no candidates (no throw)', () => {
     expect(filterFileCandidates('/no/such/dir/x', '/nonexistent-cwd')).toEqual([]);
+  });
+});
+
+describe('filterSlashCandidates', () => {
+  it('returns prefix matches for a normal query', () => {
+    const vals = filterSlashCandidates('config').map((c) => c.value);
+    expect(vals).toContain('/config');
+  });
+
+  it('finds commands by subsequence abbreviation (cfg → /config)', () => {
+    const vals = filterSlashCandidates('cfg').map((c) => c.value);
+    expect(vals).toContain('/config');
+  });
+
+  it('ranks prefix matches ahead of subsequence-only matches', () => {
+    const keys = filterSlashCandidates('co').map((c) => c.value.slice(1).toLowerCase());
+    const firstNonPrefix = keys.findIndex((k) => !k.startsWith('co'));
+    // The prefix block is contiguous at the top: once it ends, no later
+    // entry may itself be a prefix match.
+    if (firstNonPrefix !== -1) {
+      expect(keys.slice(firstNonPrefix).every((k) => !k.startsWith('co'))).toBe(true);
+    }
+  });
+
+  it('empty query returns the full command set (unchanged prefix behaviour)', () => {
+    const vals = filterSlashCandidates('').map((c) => c.value);
+    expect(vals.length).toBeGreaterThan(0);
+    expect(vals).toContain('/config');
   });
 });
--- a/src/cli/input/trigger.ts
+++ b/src/cli/input/trigger.ts
@@ -11,6 +11,7 @@ import { join } from 'path';
 import { list as listSlashCommands, aliasEntries } from '../slash/registry.js';
 import { resolveQuery, MAX_FILE_MATCHES } from '../multi-line-reader.js';
 import type { Candidate, Trigger } from './types.js';
+import type { SlashCommand } from '../slash/types.js';
 
 /**
  * Detect the trigger kind and query from the buffer at the cursor position.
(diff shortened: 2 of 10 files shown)


griffinwork40 (commented):

## Merge Review — PR #506

**Target:** PR #506 · `feat(cli): app-like TUI — /config settings menu + /model & /resume pickers` · @griffinwork40
**Reviewed ref:** `f1ec9beeb` (branch `afk/20260710-204928-0eb045`)
**Regime:** full (feature, ~1221/-182 over 11 files) · **Files:** `trigger.ts`, `config-menu.ts` (new), `config-doctor.ts`, `info.ts`, `resume.ts` (+ their tests)
**Stated intent:** PR #506 title+body — spec-compliance assessed.

### Decision: 🟢 MERGE

No correctness, security, or spec blocker survived verification. The three load-bearing invariants the PR claims were each verified first-hand at the reviewed ref: (1) no `secret`-tier key is reachable in the `/config` menu — `CONFIG_KEY_SPECS`' tier union is `agent | human` only; (2) the human-tier confirm gate blocks the write when declined, and the agent `config_set` path is not widened; (3) the bare-`/resume` picker routes through the **same** `resumeFound()` same-session data-loss guard as `/resume <id>`, so it does not bypass it. Test coverage for the two highest-value security behaviors (confirm-decline blocks the write; human-tier requires confirm) exists and asserts the right postcondition. One benign cosmetic edge in the `/model` picker is noted below (non-blocking). Merge is gated only on the author's own recommended ~30s live-TTY spin, since the interactive flows are proven via scripted fakes, not a real keypress session.

### Blocking findings

None — no correctness/security/spec blocker survived verification.

### Non-blocking findings

- **nit · correctness · `info.ts:248,257` · ref:f1ec9beeb · citation-type:file-state** — the `/model` picker seeds the cursor via `options.indexOf(ctx.stats.model)` over `MODEL_ALIASES_HINT`; when the active model is a full `org/model` HF id (not an alias) `indexOf` returns `-1` and the cursor lands on row 0 rather than the current model. Handled safely (`currentIdx >= 0 ? currentIdx : 0`) and the `current: <model>` header still shows the truth — cosmetic only.
  Evidence: `const currentIdx = options.indexOf(ctx.stats.model as ...);` / `initialIndex: currentIdx >= 0 ? currentIdx : 0,`
  Suggestion: when `currentIdx === -1`, append the live `org/model` id as an extra picker row and pre-select it (optional polish).
- **low · test-coverage · `info.test.ts` · ref:f1ec9beeb** — the `/model` picker's `indexOf → -1` (non-alias current model) branch has no dedicated case among the 4 new `info.test.ts` tests; the fallback-to-0 path is unexercised. Low value given the safe fallback.

### Confirmed clean

- **security** — no `secret`-tier key reachable in the menu: `settable-keys.ts` `ConfigKeyTier` union is `agent | human` only, and `defaultIo().write` sets `{ allowHumanOnly: true }` *only after* the human-tier confirm passes (`config-menu.ts:300-301`); agent-tier writes call `setConfigValue(path, value, undefined)` — unchanged, un-widened. No credential printed to scrollback.
- **correctness** — `/resume` picker calls the shared `resumeFound()` which enforces the same-session guard (`resume.ts` `isSameSession` → `Already on session…`; return) before `requestResume`; `/model` picker cursor `indexOf → -1` handled; `filterSlashCandidates` dedups prefix matches out of the subsequence pass via a `Set`, appends subsequence below prefix, caps at 20; overlay calls are sequential awaited (single-overlay/single-stdin invariant holds).
- **api-compat** — `/config view`, `/config set <k> <v>`, `/model <id>`, `/resume <id>`, and all non-TTY (Telegram/daemon) fallback branches remain present and behavior-unchanged; `filterSlashCandidates` preserves the prefix-first ordering.
- **spec-compliance** — all four intent clauses implemented; `config-doctor.ts` (+219) and `info.ts` (+94) churn is the `/config`-menu wiring and `/model`-picker the intent describes — no scope creep detected.
- **test-coverage** — claimed suites exist at the ref: `config-menu.test.ts` (incl. `:191` agent-tier no-confirm write, `:209` human-tier Yes→write, **`:219` decline→`io.writes` empty**), `trigger.test.ts` fuzzy cases, `config-doctor.test.ts` fast-path, `resume.test.ts` picker, `info.test.ts`.
- **perf-observability** — `filterSlashCandidates` rebuilds the `universe[]` array per keystroke at O(commands × query); negligible at the real slash-command count. No blowup.

### What was NOT checked

- Ref read: branch HEAD `f1ec9beeb` — all `blocking`/`high`-eligible citations verified at that ref; none survived as blocking.
- Stated intent: PR #506 title+body — spec-compliance assessed against it.
- **Author's own caveat (endorsed):** no live TTY keypress session was driven (headless env); interactive flows are proven via unit tests + scripted `FakeOverlays`/`runPicker` only. A ~30-second manual spin of `/config`, `/model`, `/resume` in a real terminal is recommended before merge.
- Did not run the test suite locally; the "11,215 passed" claim is taken from the PR body + green CI, not re-executed here.
- Did not review the underlying `runPicker`/`runTextInput`/`enterPickerMode` compositor internals (PR #511 territory) — treated as an already-shipped, already-tested dependency per the PR's stated approach.

griffinwork40:
All load-bearing claims are now independently verified from direct source reads. **`getCompositor` is genuinely wired** (`surface-setup.ts:258` → `ctx.slashCtx.getCompositor = () => surface.getCompositor()`, gated by `armCompositor`'s TTY check at `input-surface.ts:291`), and the PR's three new consumers follow the identical established pattern used by `transcript.ts` and `create-skill-renderer.ts`. The feature is live on a TTY and correctly falls back elsewhere.

---

# Review — PR #506: app-like TUI (/config menu + /model & /resume pickers)

**Recommendation: ✅ APPROVE** — merge-ready. No blockers, no high-severity issues. A few optional minor follow-ups below.

Well-engineered change: composed from the existing `runPicker`/`runTextInput` primitives (no new TUI stack), defensively error-handled, and with strong unit coverage. All four load-bearing properties hold.

### Verified (the things that would sink this PR if wrong)

|---|---|---|
| `getCompositor` actually wired in the real REPL (else the whole feature is silently dead) | ✅ WIRED | `surface-setup.ts:258` assigns it; `armCompositor` early-returns on non-TTY (`input-surface.ts:291`) so `getCompositor()` returns `null` off-TTY → correct fallback |
| Single-overlay / single-stdin invariant holds for the menu's back-to-back overlays (human-tier: value-pick → confirm-pick) | ✅ SAFE (high conf.) | `runPicker`/`runTextInput` call `exitPickerMode()` *before* `resolve()` in a synchronous `finish` closure (`picker.ts:162-163`, `text-input.ts:112-113`); guard at `terminal-compositor.input-mode.ts:95` always sees a cleared controller. Deterministic happens-before, not a race |
| Every imported symbol compiles under strict tsconfig | ✅ OK | `setConfigValue(…,{allowHumanOnly})`→`.value`, `getConfigValue().value`, `RESTART_NOTE`, `coerceConfigValue`→`{ok,value\|error}`, `runPicker`→`readonly string[]\|null`, `runTextInput`→`string\|null`, all 9 palette methods, `Writer`/`SlashCommand.usage`/`SlashContext.getCompositor`, `commitAbove` — all present |
| Security: config editing can't leak a credential | ✅ STRUCTURAL | `ConfigKeyTier = 'agent' \| 'human'` (`settable-keys.ts:171`) — **no `secret` tier exists**, so the menu physically cannot reach a credential; human-tier keys gated behind an explicit in-menu confirm (`config-menu.ts:665-676`) mirroring the CLI `--allow` gate |
| `/resume` refactor preserves the PR #355 data-loss guard | ✅ PRESERVED | Extracted `resumeFound` is byte-identical to the old inline block (only `return 'continue'`→`return`); same-session guard applied on both id-path and new picker-path |

### Findings

**MEDIUM**
- **The core safety invariant is untested — only comment-asserted.** `config-menu.test.ts`'s `FakeOverlays` (diff L180-212) just resolves promises; it never models `enterPickerMode`/`exitPickerMode`. So no test drives a *back-to-back* overlay sequence (value-pick → confirm-pick for a human-tier key) through a real compositor to prove the second `enterPickerMode` doesn't throw. `info.test.ts`/`resume.test.ts` do assert `enterCalls===exitCalls===1`, but only for a *single* overlay. If someone later makes `runPicker`'s teardown async (resolve-before-exit), the whole design breaks and green tests wouldn't catch it. *Suggest one integration test via the `FakeHost` pattern editing a human-tier key end-to-end, asserting balanced enter/exit and no throw.*

**LOW**
- **`overlaysFromCompositor.pick` reintroduces the label→`indexOf` collision pattern that `resume.ts` explicitly fixed** (`config-menu.ts:694-695`). It's safe *today* only because config pick-options happen to be unique (category names, path-bearing key rows, enum/bool, Yes/No). Unlike `resume.ts` — which added `uniquePickLabels` + a paragraph explaining the hazard — this path has no guard or comment. A future enum with duplicate display values or a label collision would silently edit the wrong key. *Add a one-line "options are guaranteed unique because…" comment, or share the uniqueness helper.*
- **`categoryOf` mis-placement comment overstates its safety net** (`config-menu.ts:466-471`): the comment says a mis-mapped new spec "will [be] flag[ged]" by the test, but the test only asserts membership in `CATEGORY_ORDER` — and the `'Session'` default *is* in `CATEGORY_ORDER`, so a mis-placed key passes silently. Cosmetic impact (wrong menu section), but the comment promises a guard that doesn't exist.

**NIT**
- `resume.test.ts` collision test can false-pass if the two `saveSession` calls straddle a minute boundary (`fmtWhen` truncates to the minute), making the labels non-identical and skipping the collision path. Rare; consider stubbing the clock.
- Picker-result mapping diverges across the three call sites (`/model` uses the returned label directly as the target; `/resume` and `/config` map via `indexOf`). Each is correct, but the divergence is a small readability tax.
- `filterSlashCandidates` is now case-*insensitive* (old code was case-sensitive `startsWith`). Almost certainly intended (doc says so, tests cover it) — flagging only so it's a conscious behavior change.
- Short (1-char) slash queries now append many subsequence matches after prefix matches. Bounded (prefix ranks first, list capped at 20) so prefix hits are never displaced — acceptable UX call.

### Not checked
- Did not run `pnpm lint`/`pnpm test` — the PR branch isn't checked out and the working tree is dirty/mid-work (must not mutate); verification was by reading the diff + PR-base dependency source. Signature checks used on-disk dependency files (which the PR doesn't modify), cross-checked against the diff's call sites.
- Did not visually run the TUI; overlay *rendering* correctness (layout, ANSI) is assumed from the reused primitives.

---

**Done**
- Reviewed PR #506 (+1377/−180, 10 files) read-only across correctness, API-compat, security, and test-coverage dimensions; produced the severity-ranked findings + APPROVE recommendation above.
- Evidence: verification grounded in the diff artifact (`~/.afk/state/skill-preflight/fed43a33-…/pr-506.diff`) plus direct reads of `surface-setup.ts:258`, `input-surface.ts:291`, `picker.ts:162`, `text-input.ts:112`, `terminal-compositor.input-mode.ts:95`, `settable-keys.ts:171`, `mutate.ts:44/222/277`, and `slash/types.ts:152/206/302`; two read-only research sub-agents (single-overlay invariant: SAFE/high-conf; getCompositor wiring: WIRED) plus my own signature and logic tracing.
- World change: none — no files edited, no commits, no PR comments; the dirty working tree on `fix/redact-secrets-path-false-positive` is untouched.
- Pending/deferred: the MEDIUM (add a back-to-back-overlay integration test) and the two LOW items are optional hardening, not merge blockers; the PR author (or you) can address them in-branch or as a follow-up. Findings are delivered here in-session only — I did not post them to GitHub.

---

PR #507, opened 2026-07-11
fix(tools): decouple bash/grep output kill-cap from model budget; keep head+tail

griffinwork40:
## Problem

`bash`, `grep`, and the `!`-passthrough streamer share a single hardcoded **100KB output cap that `SIGKILL`s the child process** the instant combined stdout+stderr crosses it.

Telemetry (`~/.afk/agent-framework/routing-decisions.jsonl`) shows this firing **853 times** since 2026-05-26 (ongoing daily; a spike of 75 during a SWE-bench run), split **440 `bash` / 413 `grep`**, every one at exactly `total_bytes: 100000`. The dominant triggers are verbose but *legitimate* commands — full `pnpm test` runs, `tsc`, large `git diff`, broad `grep`. Killing them mid-flight means the agent:

- never sees the command's **real exit code**,
- loses the output **tail** — head-only truncation keeps the least useful 100KB, while test/build summaries and final errors live at the *end*,
- may leave side effects half-done.

## Root cause

The 100KB cap conflates two unrelated concerns and resolves both by killing the process:

1. **V8 crash safety** — a runaway producer (`yes`, `cat` of a huge binary) must not grow a single JS string past V8's ~512MB `RangeError: Invalid string length` limit. *Hard* limit.
2. **Model-context cost** — don't feed the model a huge blob. *Soft*, tunable limit.

The `SIGKILL` is **not required for V8 safety**: the accumulator string is already clamped at the Buffer layer (`bash.ts:249-252`, `grep.ts:246-249`), so it can never grow past the cap whether or not the child is killed. Killing is a redundant, destructive over-reaction that discards the command's real result for zero safety benefit.

*(This root cause survived an adversarial devil's-advocate review; the winning fix below is the pragmatist alternative — the original proposal to "drain to 64MB" was dropped because it reintroduced event-loop starvation under grep's 8-way concurrency and a wedged-child-to-timeout regression.)*

## Fix

New `src/agent/tools/handlers/_output-cap.ts` splits the one cap into two decoupled thresholds:

- **`HARD_CAP_BYTES` (8MB)** — accumulator bound + mid-stream `SIGKILL` threshold. A genuine-runaway circuit-breaker, far above any legitimate output, so verbose commands now **run to completion** and keep their real exit code. Only true floods (>8MB) are killed.
- **`MODEL_CAP_BYTES` (100KB)** — model-facing budget, reduced to **head + tail** (first ~50KB + last ~50KB, UTF-8 safe) via `headAndTail()` so end-of-output summaries survive.
- `capForModel()` is applied to **every** model-facing settle path in `bash.ts`/`grep.ts` (overflow-kill, non-zero exit, success) — required because the raised accumulator bound means any path could now hold up to 8MB.
- Strengthened the bash/grep **tool descriptions** + tool **system-prompt** to name the head+tail behavior and steer toward `| tail`, `--quiet`, redirect, and narrower grep instead of re-running the same broad command (the near-free guidance lever; the prior generic hint clearly wasn't landing given 853 events).

### Behavior change
A command emitting >100KB now runs to completion (bounded by `timeout_ms`) instead of dying early, so tool latency reflects the command's real runtime. The `tool.overflow_kill` telemetry event now fires only for genuine >8MB runaways.

### Deliberately out of scope
The `!`-passthrough streamer (`src/agent/shell-jobs/streamer.ts`) shares the same design but contributes **0** of the 853 events (user-initiated, output already shown live on screen). Left for a consistency follow-up to keep this PR's blast radius on the actual problem.

## Tests
- Rewrote the mid-stream-kill tests in `bash.test.ts` / `grep.test.ts` / `overflow-telemetry.test.ts` for the 8MB threshold and the new head+tail / kill-note markers.
- Phase markers switched from `a`/`b` to `q`/`z` because `b` collides with the word "**b**ytes" in the elision marker.
- **Full suite green: 619 files, 11214 passed, 14 skipped.** `pnpm lint` (tsc --noEmit) clean.

### DIFF
--- a/src/agent/subagent/result.ts
+++ b/src/agent/subagent/result.ts
@@ -75,13 +75,12 @@ export interface SubagentToolResult {
   toolUseId: string;
   isError?: boolean;
   /**
-   * `true` when the tool handler reported a byte-cap overflow (e.g. bash
-   * or grep killed mid-stream because output exceeded 100KB). Reflects the
-   * structured `ToolResult.truncated` flag set by the handler — not the
-   * cosmetic 80-char display preview clip. Parent agents can use this to
-   * distinguish "subagent's bash got 100KB of legitimate output" from
-   * "subagent's bash got 100KB then was killed" without substring-scanning
-   * tool output for the `[output truncated …]` sentinel.
+   * `true` when the tool handler reduced its output to fit the model budget
+   * (e.g. bash/grep output larger than ~100KB returned as a head+tail slice,
+   * or a runaway killed at the multi-MB hard cap). Reflects the structured
+   * `ToolResult.truncated` flag set by the handler — not the cosmetic 80-char
+   * display preview clip. Parent agents use this to detect truncated tool
+   * output without substring-scanning for a sentinel.
    */
   truncated?: boolean;
   sizeBytes?: number;
--- a/src/agent/tools/handlers/_output-cap.ts
+++ b/src/agent/tools/handlers/_output-cap.ts
@@ -1,0 +1,127 @@
+/**
+ * Shared output-capping primitives for the shell-facing tool handlers
+ * (`bash`, `grep`).
+ *
+ * Two independent thresholds — deliberately decoupled, because the single
+ * 100KB cap they replace conflated two unrelated concerns and resolved both
+ * by SIGKILLing the child process the instant output crossed 100KB:
+ *
+ *   1. {@link HARD_CAP_BYTES} — the accumulator + mid-stream-kill threshold.
+ *      Its ONLY job is to bound the in-memory JS string so a genuine runaway
+ *      producer (`yes`, an accidental `cat` of a huge binary) cannot grow a
+ *      single string past V8's ~512MB max-string-length limit and crash the
+ *      host with `RangeError: Invalid string length`. Set far above any
+ *      legitimate command's output so the mid-stream kill is a true
+ *      runaway circuit-breaker, not a routine event. 8MB leaves ~64x
+ *      headroom under the V8 ceiling even with a combined stdout+stderr
+ *      counter and several tool calls running concurrently.
+ *
+ *   2. {@link MODEL_CAP_BYTES} — the budget of output actually fed back to
+ *      the model. A soft context-cost limit, unchanged from the prior 100KB.
+ *
+ * Why this matters: the prior design killed the child at 100KB, so a verbose
+ * but legitimate command (a full `pnpm test` run, a large `git diff`, a broad
+ * grep) was terminated mid-flight — the agent never saw the real exit code
+ * and lost the *tail* of the output (head-only truncation keeps the least
+ * useful 100KB, while test/build summaries and final errors live at the END).
+ * Raising the kill threshold lets such commands run to completion; capping the
+ * model-facing view with {@link headAndTail} keeps the decisive start AND end.
+ *
+ * @module agent/tools/handlers/_output-cap
+ */
+
+/**
+ * Accumulator + mid-stream SIGKILL threshold, in bytes (8MB).
+ *
+ * The child is killed only when combined stdout+stderr crosses this — a
+ * genuine runaway. Legitimate verbose output (well under 8MB) runs to
+ * completion so the real exit code and tail survive.
+ */
+export const HARD_CAP_BYTES = 8_000_000;
+
+/**
+ * Model-facing output budget, in bytes (100KB). Output larger than this is
+ * reduced to head+tail via {@link headAndTail}; the underlying command is
+ * NOT affected. Matches the prior model-context cost.
+ */
+export const MODEL_CAP_BYTES = 100_000;
+
+/**
+ * Truncate a UTF-8 Buffer to at most `maxBytes` from the FRONT without
+ * splitting a multi-byte code point. Walks back off any trailing
+ * continuation byte (0x80–0xBF). Returns a zero-copy subarray view.
+ */
+function utf8SafeHead(buf: Buffer, maxBytes: number): Buffer {
+  if (buf.length <= maxBytes) return buf;
+  let end = maxBytes;
+  while (end > 0 && (buf[end]! & 0xc0) === 0x80) end--;
+  return buf.subarray(0, end);
+}
+
+/**
+ * Keep at most `maxBytes` from the END of a UTF-8 Buffer without splitting a
+ * multi-byte code point. Advances the start offset forward off any leading
+ * continuation byte so the retained tail begins on a code-point boundary.
+ */
+function utf8SafeTail(buf: Buffer, maxBytes: number): Buffer {
+  if (buf.length <= maxBytes) return buf;
+  let start = buf.length - maxBytes;
+  while (start < buf.length && (buf[start]! & 0xc0) === 0x80) start++;
+  return buf.subarray(start);
+}
+
+/**
+ * Reserve (bytes) held back from the byte budget for the elision marker, so
+ * the returned string still fits within `maxBytes`. The marker embeds three
+ * decimal integers (elided count + head/tail lengths); 160 bytes covers even
+ * 10-digit values plus the surrounding text.
+ */
+const MARKER_RESERVE_BYTES = 160;
+
+/**
+ * Reduce `text` to at most ~`maxBytes` UTF-8 bytes while preserving BOTH
+ * ends, joined by a one-line marker naming how many bytes were elided from
+ * the middle. Returns `text` unchanged when it already fits.
+ *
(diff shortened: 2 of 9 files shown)


griffinwork40:
Wave 1.5 verification confirms the load-bearing claims:
- **105KB test** (bash.test.ts:319-330) asserts only `isError` falsy + `'bytes truncated'` — both of which a lowered-cap regression would still satisfy, so it genuinely does not prove run-to-completion. Claim **confirmed-accurate**.
- **Absence confirmed:** every exit-code test uses tiny output (`exit 42`/`1`/`127`); every large-output test exits 0. No test combines >100KB output with a failing exit → the new `capForModel()` on the non-zero-exit / grep-code-2 branches is untested. **confirmed-absent**.
- **`stripEscapeSequences`** is a single linear `.replace(ESCAPE_RE, '')` pass — perf finding correctly scoped as low; the `PHPUNIT_FAIL_RE` O(n²) is the real one.

No critical/high findings survive rubric application (the agent's `high` was a test-coverage gap, which this skill's rubric scores `low`/`medium`), and I independently verified every load-bearing claim inline — so `/shadow-verify` would be ceremony here. Emitting synthesis.

---

# Code Review — PR #507

**Target:** `fix/shell-output-cap-headtail` → `main` · reviewed ref `dd2f208` · 9 files, +329/−252 (581 lines) · regime **full**
**Change:** Decouples the shared 100KB bash/grep output cap into `HARD_CAP_BYTES` (8MB accumulator + SIGKILL) and `MODEL_CAP_BYTES` (100KB, head+tail via `headAndTail()`), so verbose-but-legitimate commands run to completion instead of being killed at 100KB.
**Stated intent:** PR #507 title+body — spec-compliance assessed.

## Verdicts
| Dimension | Verdict |
|---|---|
| security | ✅ clean |
| api-compat | ✅ clean — sentinel-wording change is non-breaking |
| correctness | ✅ sound — `headAndTail` math + UTF-8 logic verified |
| spec-compliance | ✅ all 6 requirements met; scope claim holds |
| test-coverage | ⚠️ central behavioral promise has no regression guard |
| perf-observability | ⚠️ one bounded-quadratic regex now runs on 8MB |

## Findings (severity-sorted)

**MEDIUM · test-coverage · `bash.test.ts:319-330` · ref:dd2f208 · file-state**
The PR's central promise — a >100KB-but-<8MB command *runs to completion, keeping its real exit code and tail* — has no test that proves it. The only in-band test asserts `isError` falsy + `'bytes truncated'`; a regression that lowered `HARD_CAP_BYTES` back to 100KB would still kill the command via the overflow path, which *also* emits `headAndTail`'s `'bytes truncated'` marker — so both assertions still pass.
Evidence: `expect(result.isError).toBeFalsy(); expect(result.content).toContain('bytes truncated');`
Fix: add a test emitting ~500KB then `exit 7`, asserting `isError && /code 7/` and that the kill sentinel (`was terminated`) is **absent** (proves close-path, not kill-path) + an elapsed lower-bound.
*(Downgraded from the sub-agent's `high`: this skill's rubric scores a missing test `low`; raised to `medium` because it is the sole regression guard for the whole change.)*

**MEDIUM · test-coverage · `bash.ts:288`, `grep.ts:285` · ref:dd2f208 · file-state**
The newly-added `capForModel()` on the non-zero-exit path and grep `code===2` path is exercised by no test combining large (>100KB) output with a failing exit — confirmed by absence check (`exit 42/1/127` tests all use tiny output; all large-output tests exit 0). The truncation branch of these two settle paths ships untested.
Evidence: `const capped = capForModel(stderr.trimEnd() || stdout.trimEnd());`
Fix: same combined large-output + non-zero-exit test covers this.

**MEDIUM · perf-observability · `test-runner-detector.ts:86` · ref:dd2f208 · file-state**
`PHPUNIT_FAIL_RE = /Tests:\s*(\d+)[^]*?Failures:\s*(\d+)/` now scans up to 8MB (was ≤100KB) on **every** bash close, not just test commands. On pathological input (many `Tests:` tokens, no `Failures:`) the lazy `[^]*?` gap is O(n²), blocking the event loop for seconds — a latent issue this PR amplifies ~80× via the raised accumulator bound. Bounded-quadratic, not exponential; reachability requires adversarial/verbose output.
Evidence: `const PHPUNIT_FAIL_RE = /Tests:\s*(\d+)[^]*?Failures:\s*(\d+)/;`
Fix: bound the gap (`[^]{0,4096}?`) or run `detectTestResult` on a tail slice when the buffer exceeds ~1MB.

**LOW · perf-observability · `bash.ts:298` · ref:dd2f208 · file-state**
`stripEscapeSequences` (single linear `.replace`) + the other ~9 `detectTestResult` regexes also run synchronously on up to 8MB every close. Individually linear/bounded, but combined with the finding above this is unmeasured added close-path latency on any verbose command.
Fix: benchmark close-path latency on an 8MB buffer; gate detection to the tail past ~1MB.

**LOW · test-coverage · `grep.test.ts:349,392` · ref:dd2f208 · file-state**
Kill-path timing bound loosened to `<5_000ms`; a wall-clock upper bound is the flakiest assertion here on a loaded CI box. The order-agnostic one-file-only + `not 'z'` assertions already prove the kill fired.
Fix: drop or widen the timing bound; keep the content assertions as the kill proof.

**NIT · `subagent/result.ts:16-21` · ref:dd2f208 · file-state**
`SubagentToolResult.truncated` doc semantics broadened (100KB-kill → "reduced to fit model budget OR runaway kill"). Accurate and correctly updated — a doc-contract change outside the 6 named requirements. Note only.

## Verified-clean (checked and passed)
- **api-compat:** the truncation sentinel changed wording, but `truncated: true` is set on **all** six content-reducing settle paths (`_output-cap.ts:147` returns it whenever it reduces), and `HARD_CAP_KILL_NOTE` preserves the `[output truncated` prefix. The two production consumers (`loop.ts:634`, `dispatch-append.ts:168`) read `result.truncated === true || content.includes('[output truncated')` — the flag is the primary signal, so no consumer silently stops detecting truncation. `git grep "exceeded 100KB"` / bare `[output truncated]` → 0 production consumers.
- **security:** single combined `totalBytes` counter → ≤8MB per call (~64MB worst case under grep's 8-way concurrency, far under V8's ~512MB string ceiling); Buffer-layer slice before `.toString()` preserved; `tool.overflow_kill` telemetry payload untouched — pattern/path/command still excluded (privacy boundary intact).
- **correctness:** `headAndTail` output is strictly `< maxBytes` (budget = maxBytes − 160 reserve, marker ~82B actual); `utf8SafeHead`/`utf8SafeTail` continuation-byte logic (`& 0xc0) === 0x80`) is correct; `elided > 0` guaranteed; `detectTestResult` runs on the full output **before** capping, so test summaries survive.
- **spec-compliance:** all 6 requirements implemented; the `!`-passthrough streamer (`streamer.ts:153` still `DEFAULT_MAX_BYTES = 100_000`) and `read_file` correctly untouched — the out-of-scope claim holds.

## What was not checked
- Wave 1 agents read against branch HEAD `dd2f208b2ee53a82efc921df4141f2fe4b186879`; all citations verified at that ref.
- The suite claim (**619 files / 11214 passed / 14 skipped, lint clean**) was **not** independently run — this review is read-only. `unverified — suite not executed`.
- Did not assess the Telegram/iMessage surfaces or actual runtime memory under real 8-way grep load (reasoned from the code, not measured).

## Merge decision: **MERGE**

No blocking defects. The code is correct, api-compatible, and satisfies its stated intent; the head+tail decoupling is sound and the accumulator-vs-model-budget split is the right structural fix. All findings are non-blocking follow-ups — two test-coverage gaps (add one large-output + non-zero-exit test to guard the core promise and cover the `capForModel` failure branches) and one cheap perf mitigation (bound `PHPUNIT_FAIL_RE`'s lazy gap). None gate the merge.

**Done**
- Reviewed PR #507 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability at ref `dd2f208`; ran Wave 1 (2 parallel dimension agents) + inline Wave 1.5 citation/absence verification.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/c510ac6b-d85a-4300-83cc-9e83af498717/pr-507.diff`; findings cite `file:line` at the reviewed ref; absence claims grounded via `git grep`; verified-clean items independently confirmed by reading the files at the ref.
- Verdict: **MERGE** — 3 medium + 2 low + 1 nit, all non-blocking; no critical/high after rubric application.
- Read-only: no files, PR, or working-tree state were modified (the 4 pre-existing dirty files were left untouched).
- Pending (author's discretion, non-blocking): add a large-output + non-zero-exit regression test; bound `PHPUNIT_FAIL_RE`'s `[^]*?` gap.

---

PR #513, opened 2026-07-11
fix(tui): refresh lastMeasuredFrameTop in the picker repaint path too

griffinwork40:
## Summary

Stacked on #505 — depends on the `lastMeasuredFrameTop` field that PR introduces, so this targets `afk/fix-tui-table-void` rather than `main`.

While reviewing #505 we traced an open question about whether a stale-picker-frame edge case was reachable. It is, via a simpler mechanism than originally hypothesized: it doesn't require any specific concurrent-subagent timing, just any `commitAbove()` call landing while a picker is open.

## Root cause

`repaintPickerFrame()` (`terminal-compositor.frame.ts`) computes its own frame top via the same `logUpdate.measure()` call the non-picker `repaint()` body uses to set `self.lastMeasuredFrameTop` — but never writes it back. A picker's frame (its own rows + chrome) has a different row count than whatever was on screen before it opened, so `lastMeasuredFrameTop` kept describing the pre-picker layout for as long as the picker stayed active. This isn't a resize, so `bandGeometryStale` (the flag #505's fix uses to distrust a stale measurement) never catches it. Any `commitAbove()` landing while a picker is open — e.g. a backgrounded subagent job's completion notice on the shared/borrowed compositor — would trust the stale value, reproducing a narrower variant of the 'two boxes + blank void' bug #505 fixes.

## Fix

One line, mirroring the fix already in the non-picker path: write `self.lastMeasuredFrameTop = desiredTopRow;` in `repaintPickerFrame()` too.

## Verification

- `tsc --noEmit` clean.
- Full suite: 603 files / 11119 passed / 14 skipped (pre-existing, unrelated) / 0 failed.
- New regression test added (`terminal-compositor.test.ts`, picker-mode describe block): confirmed it **fails** on the pre-fix code (a sentinel value planted in `lastMeasuredFrameTop` survives picker entry unchanged) and **passes** after the fix (the sentinel is overwritten by the real measured top).

## Scope note

Deliberately minimal and stacked rather than folded into #505 — keeps that PR's already-reviewed diff untouched. Rebase target should move to `main` once #505 merges.


griffinwork40:
---

## PR #513 Review: `fix(tui): refresh lastMeasuredFrameTop in the picker repaint path too`

**Merge recommendation: ✅ APPROVE** — correct, narrowly scoped fix with sound math and full test-suite passage.

---

### Correctness — ✅ Verified

**The `prevTopRow` refactor is mathematically sound.** Case analysis of the old vs new expression:

| Condition | Old result | New result | Match? |
|-----------|-----------|-----------|--------|
| `bandGeometryStale = true` | `rawLogUpdateTopRow` | `Math.max(raw, 0, 0)` = `raw` | ✅ |
| `!stale, no band` | `rawLogUpdateTopRow` | `Math.max(raw, measuredFrameTop, 0)` | ✅ **improvement** — uses real top instead of shrink-padded |
| `!stale, with band` | `Math.max(raw, bandBottom+1)` | `Math.max(raw, measuredFrameTop, bandBottom+1)` | ✅ **improvement** — adds measured top |
| `lastMeasuredFrameTop = 0` (pre-first-repaint) | `raw` or `Math.max(raw, bandFloor)` | Same — 0 never dominates `Math.max` | ✅ |

The `Math.max` approach is monotonic — it can only RAISE `prevTopRow` toward the true top, never lower it. This is the key safety invariant: the existing `committedBandBottomRow+1` floor is preserved, and `rawLogUpdateTopRow` is still the lower bound.

**`bandGeometryStale` gate is correct.** After SIGWINCH with no repaint, both `lastMeasuredFrameTop` and the band floor are pre-resize values — zeroing them both falls through to `rawLogUpdateTopRow` → BLOCKER-1 band-hold safety path. Exactly matches the prior behavior in the stale case.

**Repaint path coverage is complete.** Only two functions compute `desiredTopRow` via `measure()`:
1. `repaint()` (frame.ts:278) — writes `lastMeasuredFrameTop` ✓
2. `repaintPickerFrame()` (frame.ts:362) — writes `lastMeasuredFrameTop` ✓ **(the fix)**

No other repaint paths exist that could leave the field stale.

### Interface Consistency — ✅ Verified

All three declarations match:
- `CommittedBandHost` interface (committed-band-commit.ts:40-41): `lastMeasuredFrameTop: number`
- `FrameHost` interface (frame.ts:75-77): `lastMeasuredFrameTop: number`
- `TerminalCompositor` class (terminal-compositor.ts:361): `lastMeasuredFrameTop = 0` with `@internal` relaxation

Both interfaces are internal module contracts — only `TerminalCompositor` implements them. No external implementations to update.

### Test Coverage — ⚠️ Adequate, could be stronger

**What's tested:** The new test sets sentinel `999`, enters picker mode, asserts the value changed. This verifies `repaintPickerFrame()` now writes the field — the core fix of commit 2.

**Gaps (non-blocking):**
1. **Weak assertion**: `not.toBe(999)` passes for any non-999 value, including wrong ones. A mocked `measure()` returning a known value + exact equality check would be tighter.
2. **No `commitAbove()` routing test**: The core bug (commit 1 — shrink-padded top routing a fitting block into band-hold) is referenced in comments as covered by `terminal-compositor.commit-geometry.test.ts`, but no new test in this PR exercises `commitAbove()` with a divergent `lastMeasuredFrameTop` vs `logUpdate.topRow`.
3. **No picker-mode `commitAbove()` test**: The actual user-visible scenario (backgrounded job completion notice arriving while picker is up) isn't directly tested.

All 11,119 existing tests pass, confirming no regressions.

### Security & Performance — ✅ No concerns

- Simple `number` field, no user input path
- One field write per repaint + one extra `Math.max` argument — negligible overhead
- No new external surface or API change

### Lint — ✅ Passes

`tsc --noEmit` (maximally strict) passes clean on the PR branch.

---

### Summary

| Dimension | Verdict |
|-----------|---------|
| Correctness | ✅ Sound math, complete repaint coverage, correct stale gate |
| Test coverage | ⚠️ Core fix tested; routing scenario and exact-value assertions could be stronger |
| API compat | ✅ Internal interfaces only, consistent across all 3 sites |
| Security/perf | ✅ No concerns |
| Lint/types | ✅ Clean |

**Approve.** The fix is correct, narrowly scoped, and well-documented. The test gaps are real but non-blocking — the existing suite's passage plus the mathematical invariant (Math.max can only raise, never lower) give high confidence this won't regress.

---

PR #515, opened 2026-07-11
feat(service): Linux systemd --user support behind a platform-neutral ServiceManager

### LINKED ISSUE #494: Linux compatibility (systemd service backend + Linux-path test coverage)
## Linux compatibility

**Status — updated 2026-07-11:** ✅ **Phase 1 (the 🔴 service-layer blocker) is DONE.** A Linux systemd `--user` backend, sitting behind a platform-neutral `ServiceManager` seam, shipped in **#515** (CI all-green, pending merge). **Remaining work — this issue stays open until it lands:** the four "works-but-untested" Linux branches **still have zero Linux-path test coverage** (🟡 Phase 2), and CI **still runs only on `ubuntu-latest`** (no macOS matrix). Original scoping below is preserved as filed; see the status comment for the up-to-date breakdown.

Text clipboard, keychain, OAuth, and terminal-spawn already carry working Linux branches (unchanged) — those were never the gap.

---

### ✅ Blocker RESOLVED in #515 — Service layer is now cross-platform (systemd `--user` backend)

> **Done (PR #515):** extracted a platform-neutral `ServiceManager` interface + `serviceManagerFor(platform)` factory (`src/service/{types,index}.ts`), added a full systemd `--user` backend (`src/service/systemd/*`) with a companion `.path` unit for the WatchPaths equivalent, adapted launchd behind the same interface (776-line launchd suite untouched), replaced the `assertMacOS()` hard-throw with factory dispatch, and updated the service-setup/get-started skills. `tsc --noEmit` clean; full vitest 11,233 passed / 0 failed; CI all-green. Original problem statement preserved below.

`afk service` (install/uninstall/status as an always-on, auto-restart background process) hard-refuses on any non-darwin platform and there is **no systemd implementation shipped**.

- `src/cli/commands/service.ts:37-43` — `assertMacOS()` throws `'afk service' uses macOS launchd and is only supported on darwin.` It gates the `install`/`uninstall` actions (called at `service.ts:65`).
- `src/service/` contains **only** `launchd.ts` + `launchd.test.ts` — no systemd module. Confirmed: **zero** matches for `systemd` or `WantedBy` anywhere under `src/`.
- The service-setup skill still tells Linux users the feature is unavailable: `src/skills/service-setup/prompts/system.md:35` — *"On Linux you'd want a systemd user unit (not yet shipped)."*

**Work item (headline remaining work):** add a systemd user-unit backend (`~/.config/systemd/user/*.service` + `systemctl --user` bootstrap), branch `service.ts` on `process.platform`, and replace the `assertMacOS()` hard-throw with a platform dispatch. `daemon` and `telegram` are the two service names to support (`SERVICE_NAMES`, `service.ts:26`).

---

### 🟡 Linux paths that WORK but are UNTESTED

Each branch below was code-read and confirmed correct. The blocker is coverage: **no test exercises the Linux branch of any of these** — verified by reading each corresponding `*.test.ts`.

- [ ] **`readProcessMeta()` `/proc` parsing** — `src/telegram/manager.ts:276-293` reads `/proc/<pid>/stat` (starttime) + `/proc/<pid>/status` (`VmRSS`) on Linux. ❌ Not covered — `manager.test.ts` has no `readProcessMeta`/`/proc` reference. Add a Linux test that mocks `readFileSync`/`statSync` for those paths and asserts `{ uptimeSec, memoryMb }`.
- [ ] **OAuth `defaultBackend()` Linux credential file** — `src/agent/mcp/oauth.ts:106-137` reads/writes `~/.claude/.credentials.json` (mode `0o600`) on Linux. ❌ Not covered — `oauth.test.ts` has no `defaultBackend`/`credentials.json` reference. Add a Linux test that stubs `process.platform`, points `homedir()` at a temp dir, and round-trips read/write (asserting `0o600`).
- [ ] **Chrome Linux binary detection** — `src/cli/commands/browser.ts:140-164` probes `google-chrome` / `google-chrome-stable` / `chromium` / `chromium-browser` on non-darwin. ❌ Not covered **and currently untestable as written** — `detectChromeMajorVersion` is private/non-exported. Export it (or inject an `execFileSync` seam), then add a Linux test asserting the candidate list is probed and a version parsed.
- [ ] **terminal-font-size Linux editor discovery** — `src/agent/tools/handlers/terminal-font-size.ts:72-77` appends `~/.config/Code/User/settings.json` + `~/.config/Cursor/User/settings.json` on Linux. ❌ Not covered — every test in `terminal-font-size.test.ts` injects a **mock `discoverFn`**, so the real `discoverEditors()` Linux branch is never invoked. Add a test that calls the real `discoverEditors()` with `process.platform` stubbed to `linux` and `HOME` at a temp dir.

---

### 🟡 CI — add a macOS matrix entry

- [ ] **Not done.** All jobs run on `ubuntu-latest` only — `.github/workflows/ci.yml:25/75/113/145`, plus `publish.yml` and `auto-release.yml`. No `macos-latest` / `matrix.os` anywhere in `.github/workflows/`. Convert the `test` job (`ci.yml:73-105`) to `strategy.matrix.os: [ubuntu-latest, macos-latest]` with `runs-on: ${{ matrix.os }}`, so the darwin `ps -p` / `security` / `/Applications/...` branches get exercised alongside the Linux branches. The fork-PR constraint (GitHub-hosted runners only) is already satisfied — `macos-latest` is GitHub-hosted.

---

### Acceptance criteria

- [x] `afk service install {daemon,telegram}` works on Linux via a systemd user unit, and `assertMacOS()`'s hard-throw is replaced by a platform dispatch (`service.ts:37-43`, `:65`). — **#515**
- [x] `afk service {status,uninstall}` operate against the systemd unit on Linux. — **#515**
- [x] The service-setup skill prompt no longer says systemd is "not yet shipped" (`system.md:35`). — **#515**
- [ ] `readProcessMeta()` has a Linux `/proc`-path test (`manager.ts:276-293`).
- [ ] OAuth `defaultBackend()` has a Linux `.credentials.json` read+write test (`oauth.ts:106-137`).
- [ ] Chrome Linux binary detection is reachable from tests (export/seam) and covered (`browser.ts:140-164`).
- [ ] terminal-font-size `discoverEditors()` has a Linux-branch test invoking the real discovery (`terminal-font-size.ts:72-77`).
- [ ] CI runs the test suite on both `ubuntu-latest` and `macos-latest` via `matrix.os` (`ci.yml:73-105`).

---

### Out of scope

- Windows support (no `win32` branches audited or planned here).
- Native Keychain/libsecret integration — the Linux OAuth path deliberately uses a `0o600` plaintext file at `~/.claude/.credentials.json` (`oauth.ts:133-137`), matching Claude Code's on-disk format; a hardened secret-store backend is a separate follow-up.
- Replacing the macOS `security`-CLI keychain shell-out with a native binding (already a follow-up in `oauth.ts:126-127`).
- Bundling/pinning a Chrome binary — detection remains best-effort with a soft note, never a hard failure.

griffinwork40:
## Summary

Makes `afk service` cross-platform. It was hard-gated to macOS via a single `assertMacOS()` throw + direct launchd imports; this extracts a backend-neutral seam and adds a first-class Linux **systemd `--user`** backend.

**Part of #494 (Linux compatibility) — Phase 1: the service-layer blocker.** Does not fully close #494; the secondary untested-Linux-branch touchpoints are a tracked follow-up (below).

## What's in it

- **`service/types.ts`** — `ServiceManager` interface + neutral result shapes (`configPath`, `autoRestartOnRebuild`, `notes[]`, `backend`/`configKind`). The de-leaked contract both backends satisfy.
- **`service/index.ts`** — `serviceManagerFor(platform = process.platform)` factory (`darwin`→launchd, `linux`→systemd, else→`null`). The CLI renders a clear "not supported on <platform>" message on `null` instead of throwing mid-pipeline.
- **`service/launchd/manager.ts`** — thin adapter delegating to the **existing, unchanged** launchd free-functions and mapping `plistPath`→`configPath` etc. The 776-line launchd test suite is untouched.
- **`service/systemd/`** — full `--user` backend: `.service` + companion `.path` unit generation (`unit.ts`), atomic install with `daemon-reload` + `enable --now` and rollback (`install.ts`), `systemctl show` parsing (`status.ts`), XDG-default paths (`paths.ts`), restart via `systemctl --user restart` (`manager.ts`). Surfaces the `loginctl enable-linger` requirement as an install note.
- **`service/cli/commands/service.ts`** — every subcommand dispatches through the resolved manager.
- **`service-setup` + `get-started` skills** — docs/metadata now name both backends (launchd + systemd) and the Linux lingering step.

Resolves the two design decisions from the scope: ship a companion `.path` unit for the WatchPaths equivalent, and *surface* (not auto-enable) lingering.

## Verification

- Full `vitest` suite — **11,233 passed / 14 skipped / 0 failed**, including new `src/service/index.test.ts` (factory dispatch) + `src/service/systemd.test.ts` (unit/path rendering, `systemctl show` parse). launchd suite unchanged and green.
- Built + verified on macOS; the systemd runtime path (real `systemctl --user`) is unit-tested but **not yet smoke-tested on a live Linux box** — see follow-ups.

- **Phase 2 — secondary touchpoints** with untested Linux branches: telegram `readProcessMeta` `/proc` parse, `mcp/oauth` `defaultBackend` credential store, browser `detectChromeMajorVersion`, `terminal-font-size` `discoverEditors`.
- **Real-Linux smoke test**: install/status/restart, reboot persistence under `enable-linger`, telegram bot under systemd.
- **Minor**: `resolveAfkBinary` trusted-prefix allowlist lacks `~/.local/bin` + version-manager dirs common on Linux; `XDG_CONFIG_HOME` not yet honored; hoist the shared resolvers out of `launchd/plist.ts` into `service/resolve.ts`.

### DIFF
--- a/src/cli/commands/service.ts
+++ b/src/cli/commands/service.ts
@@ -1,75 +1,75 @@
 /**
  * `afk service` command group — install AFK long-running processes as
- * macOS LaunchAgents so they auto-start on login and relaunch on crash.
+ * OS-supervised services so they auto-start on login and relaunch on
+ * crash. Backend is chosen per platform by `serviceManagerFor()`:
+ *   - macOS → launchd LaunchAgents (`~/Library/LaunchAgents/`)
+ *   - Linux → systemd `--user` units (`~/.config/systemd/user/`)
  *
  * Subcommands:
- *   afk service install <name>     — write plist + bootstrap into launchctl
- *   afk service uninstall <name>   — bootout + remove plist
+ *   afk service install <name>     — write config + register with supervisor
+ *   afk service uninstall <name>   — deregister + remove config
  *   afk service status [name]      — show running PID + last exit + log path
  *   afk service list               — show all services and whether installed
+ *   afk service restart <name>     — restart the service
  *
- * `<name>` ∈ { telegram, daemon }. See `src/service/launchd.ts` for the
- * rationale on per-user scope (vs. system-wide LaunchDaemons), WatchPaths
- * auto-restart heuristics, and why launchd runs the entrypoints directly
- * instead of the `afk telegram start` / `afk daemon` CLI wrappers.
+ * `<name>` ∈ { telegram, daemon }. See `src/service/types.ts` for the
+ * backend-neutral contract and `src/service/{launchd,systemd}/` for the
+ * per-OS rationale (per-user scope, auto-restart heuristics, why the
+ * entrypoints are run directly instead of the CLI wrappers).
  *
  * @module cli/commands/service
  */
 
 import { Command } from 'commander';
 import chalk from 'chalk';
-import { execFileSync } from 'child_process';
-import { existsSync } from 'fs';
 import { palette } from '../palette.js';
 import { handleCommandError } from '../errors/index.js';
 import {
   SERVICE_NAMES,
-  installService,
-  labelFor,
-  plistPath,
-  serviceLogPath,
-  serviceStatus,
-  uninstallService,
+  SUPPORTED_SERVICE_PLATFORMS,
+  serviceManagerFor,
+  type ServiceManager,
   type ServiceName,
-  type ServiceStatusSnapshot,
-} from '../../service/launchd.js';
+  type ServiceStatus,
+} from '../../service/index.js';
 
-function assertMacOS(): void {
-  if (process.platform !== 'darwin') {
+/** Resolve the platform's service backend, or fail with a clear message. */
+function resolveManager(): ServiceManager {
+  const mgr = serviceManagerFor();
+  if (!mgr) {
     throw new Error(
-      `'afk service' uses macOS launchd and is only supported on darwin. Detected: ${process.platform}.`,
+      `'afk service' is not supported on ${process.platform}. Supported: ${SUPPORTED_SERVICE_PLATFORMS}.`,
     );
   }
+  return mgr;
 }
 
 function parseServiceName(input: string): ServiceName {
   const lower = input.toLowerCase();
   if ((SERVICE_NAMES as readonly string[]).includes(lower)) return lower as ServiceName;
-  throw new Error(
-    `Unknown service '${input}'. Supported: ${SERVICE_NAMES.join(', ')}.`,
-  );
+  throw new Error(`Unknown service '${input}'. Supported: ${SERVICE_NAMES.join(', ')}.`);
 }
 
 export function registerServiceCommand(program: Command): void {
   const service = program
     .command('service')
-    .description('Manage AFK background services via macOS launchd (always-on, auto-restart)');
+    .description('Manage AFK background services (launchd on macOS, systemd --user on Linux) — always-on, auto-restart');
 
   service
     .command('install <name>')
-    .description(`Install <${SERVICE_NAMES.join('|')}> as a LaunchAgent that starts on login and relaunches on crash`)
-    .option('--no-watch', 'Disable WatchPaths (no auto-restart on rebuild)')
-    .option('--dry-run', 'Write the plist but do not call launchctl', false)
+    .description(`Install <${SERVICE_NAMES.join('|')}> as an OS service that starts on login and relaunches on crash`)
+    .option('--no-watch', 'Disable auto-restart-on-rebuild (launchd WatchPaths / systemd .path unit)')
+    .option('--dry-run', 'Write the config file but do not register with the supervisor', false)
     .action((nameArg: string, opts: { watch?: boolean; dryRun?: boolean }) => {
       try {
-        assertMacOS();
+        const mgr = resolveManager();
         const name = parseServiceName(nameArg);
-        const result = installService(name, {
+        const result = mgr.install(name, {
           noWatch: opts.watch === false,
-          skipBootstrap: Boolean(opts.dryRun),
+          dryRun: Boolean(opts.dryRun),
         });
         if (result.kind === 'already-installed') {
-          console.log(chalk.yellow(`⚠ ${result.label} already installed at ${result.plistPath}`));
+          console.log(chalk.yellow(`⚠ ${result.label} already installed at ${result.configPath}`));
           console.log(palette.meta(`  Run 'afk service uninstall ${name}' first to reinstall.`));
           process.exit(1);
         }
@@ -78,21 +78,17 @@ export function registerServiceCommand(program: Command): void {
           process.exit(1);
         }
         console.log(chalk.green(`✓ Installed ${result.label}`));
-        console.log(palette.meta(`  Plist:   ${result.plistPath}`));
-        console.log(palette.meta(`  Log:     ${serviceLogPath(name)}`));
-        if (result.watchPathsActive) {
-          console.log(palette.meta(`  WatchPaths: active — service auto-restarts on rebuild.`));
+        console.log(palette.meta(`  Config:  ${result.configPath}  (${mgr.configKind})`));
+        console.log(palette.meta(`  Log:     ${mgr.logPath(name)}`));
+        if (result.autoRestartOnRebuild) {
+          console.log(palette.meta(`  Auto-restart on rebuild: on`));
         } else {
-          console.log(palette.meta(`  WatchPaths: off — manual 'afk service restart' needed after updates.`));
+          console.log(palette.meta(`  Auto-restart on rebuild: off — run 'afk service restart ${name}' after updates.`));
         }
-        if (opts.dryRun) {
-          // M-10: interpolate the real uid so the copy-paste command works
-          // on this machine. `$(id -u)` would only expand inside a shell,
-          // but we're printing a raw string to the terminal.
-          const uid = process.getuid?.() ?? 501;
-          console.log(palette.info(`  (dry-run) launchctl bootstrap was skipped; service is NOT yet running.`));
-          console.log(palette.meta(`  Load manually: launchctl bootstrap gui/${uid} ${result.plistPath}`));
-        } else {
+        for (const note of result.notes ?? []) {
+          console.log(palette.info(`  ${note}`));
+        }
+        if (!opts.dryRun) {
           console.log(palette.meta(`  Status:  afk service status ${name}`));
         }
       } catch (err) {
@@ -102,22 +98,22 @@ export function registerServiceCommand(program: Command): void {
 
   service
     .command('uninstall <name>')
-    .description('Stop the service and remove its LaunchAgent plist')
+    .description('Stop the service and remove its config (LaunchAgent plist / systemd unit)')
     .action((nameArg: string) => {
       try {
-        assertMacOS();
+        const mgr = resolveManager();
         const name = parseServiceName(nameArg);
-        const result = uninstallService(name);
+        const result = mgr.uninstall(name);
         if (result.kind === 'not-installed') {
-          console.log(chalk.yellow(`⚠ ${labelFor(name)} is not installed (no plist at ${result.plistPath})`));
+          console.log(chalk.yellow(`⚠ ${mgr.label(name)} is not installed (no config at ${result.configPath})`));
           return;
         }
         if (result.kind === 'failed') {
           console.error(chalk.red(`✗ Uninstall failed: ${result.reason}`));
           process.exit(1);
         }
-        console.log(chalk.green(`✓ Uninstalled ${labelFor(name)}`));
-        console.log(palette.meta(`  Removed: ${result.plistPath}`));
+        console.log(chalk.green(`✓ Uninstalled ${mgr.label(name)}`));
+        console.log(palette.meta(`  Removed: ${result.configPath}`));
       } catch (err) {
         handleCommandError(err);
       }
@@ -128,14 +124,13 @@ export function registerServiceCommand(program: Command): void {
     .description('Show running PID, last exit status, and log file for one or all services')
     .action((nameArg: string | undefined) => {
       try {
-        assertMacOS();
+        const mgr = resolveManager();
         if (nameArg) {
-          const snapshot = serviceStatus(parseServiceName(nameArg));
-          printStatus(snapshot);
+          printStatus(mgr.status(parseServiceName(nameArg)), mgr.configKind);
           return;
         }
         for (const name of SERVICE_NAMES) {
-          printStatus(serviceStatus(name));
+          printStatus(mgr.status(name), mgr.configKind);
           console.log('');
         }
       } catch (err) {
@@ -148,14 +143,13 @@ export function registerServiceCommand(program: Command): void {
     .description('List recognised service names and whether each is installed')
     .action(() => {
       try {
-        assertMacOS();
-        console.log(chalk.bold('AFK services:'));
+        const mgr = resolveManager();
+        console.log(chalk.bold(`AFK services (${mgr.backend}):`));
         for (const name of SERVICE_NAMES) {
-          const path = plistPath(name);
-          const installed = existsSync(path);
+          const installed = mgr.isInstalled(name);
           const marker = installed ? chalk.green('●') : chalk.dim('○');
           const tag = installed ? palette.meta('installed') : palette.meta('not installed');
-          console.log(`  ${marker} ${name.padEnd(10)}  ${tag}  ${palette.meta(path)}`);
+          console.log(`  ${marker} ${name.padEnd(10)}  ${tag}  ${palette.meta(mgr.configPath(name))}`);
         }
       } catch (err) {
         handleCommandError(err);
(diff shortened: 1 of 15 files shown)


griffinwork40 (commented):

## Merge Review — PR #515

**Target:** PR #515 · `feat(service): Linux systemd --user support behind a platform-neutral ServiceManager` · @griffinwork40
**Reviewed ref:** `2302a3ae` (branch `afk/linux-compat-494` → `main`)
**Regime:** full (new-service feature · +1225/−112 over 15 files)
**Stated intent:** PR #515 title+body — spec-compliance assessed against it.

### Decision: 🔴 DO NOT MERGE

The seam is clean and every headline intent clause is *implemented* (see below). What blocks merge is the one behavior a new service backend cannot afford to get wrong — **recovery from a failed install** — which is both partially broken and untested on the target platform. No critical/high finding exists; the block is the sum of a reachable rollback gap on the exact scenario the PR itself documents, an unescaped unit-file injection surface, and the failure paths being wholly unexercised by CI, on a backend the author discloses was never smoke-tested on real Linux.

### Blocking / should-fix before merge

**1 · medium · confidence: high · correctness (unmet-intent sub-clause) · `src/service/systemd/install.ts:162-168` · ref:2302a3ae · citation-type:file-state**
Rollback is only wired for the `.path`-unit write failure (`rmSync(path)` at line 141), **not** for a `systemctl enable --now` failure. The catch at 167-168 returns `failed` and leaves both the `.service` and `.path` files on disk. Because `atomicWrite` opens with `flag: 'wx'` (line 65), a subsequent `installSystemdService` hits `EEXIST` — the operator is wedged until a manual `rm`. The PR's stated intent explicitly promises "atomic install … and rollback"; this is an unmet sub-clause, not scope creep. **Reachable:** `enable --now` fails whenever no user systemd instance is active — a headless SSH session *before* `loginctl enable-linger`, the exact case this PR documents.
```
} catch (err) {
  return { kind: 'failed', reason: `systemctl enable failed: ${errorDetail(err)}` };  // ← no rmSync of the unit files
}
```
Suggestion: in the `enable --now` catch, `rmSync` both unit paths (mirror the cleanup already present at 140-142) before returning `failed`.

**2 · medium · confidence: medium · security (hardening) · `src/service/systemd/unit.ts:42-49` · ref:2302a3ae · citation-type:file-state**
`quoteArg` / `escapeEnvValue` escape `\` and `"` only — never `\n`/`\r` (nor `%`, see finding 3). A value containing a newline written into `ExecStart=` / `Environment="K=V"` starts a fresh unit-file line and can inject a directive (e.g. a spurious `ExecStartPre=`) once `daemon-reload` + `enable --now` load it.
```
function quoteArg(s: string): string {
  return `"${s.replace(/\\/g, '\\\\').replace(/"/g, '\\"')}"`;
}
```
Reachability: no current production caller feeds attacker-influenced data into `environment`/`execStart` (`cli/commands/service.ts` calls `mgr.install(name, { noWatch, dryRun })` only), so this is a latent hardening gap, not a proven-reachable RCE today — hence medium, not critical. Suggestion: reject or escape `\n`/`\r` (and double `%`→`%%`) in both helpers before the value reaches the unit file.

**3 · medium · confidence: high · correctness · `src/service/systemd/unit.ts:42-49` · ref:2302a3ae · citation-type:file-state**
Neither helper doubles a literal `%`. systemd expands `%`-specifiers in `ExecStart=` (systemd.service(5): "Use `%%` to specify a single percent sign"), so any resolved binary path or environment value containing `%` is silently misinterpreted. Suggestion: add `.replace(/%/g, '%%')` before quoting in both helpers.

### Non-blocking (follow-ups)

- **low · test-coverage · `src/service/systemd.test.ts` · ref:2302a3ae · citation-type:file-state** — the failure/rollback paths are wholly unexercised: every mock is `mockExecFileSync.mockReturnValue(Buffer.from(''))` (line 158) and no test throws, so `errorDetail()`, the `enable --now` failure branch (install.ts:167-168), and the uninstall `rmSync`-throws branch never run — the rollback gap (finding 1) ships unguarded despite the "0 failed" suite. Suggestion: add a case with `mockExecFileSync.mockImplementationOnce(() => { throw … })` on the `enable --now` call asserting the unit files are removed.
- **low · test-coverage · `src/service/systemd/manager.ts:40-55` · ref:2302a3ae · citation-type:file-state** — `restart()` is independent logic (not a delegate) yet has zero test invocation; `systemd.test.ts` imports `install`/`paths`/`status`/`unit` only (lines 36-43) and never `./systemd/manager.js`, and `index.test.ts` asserts only `.label()`/`.configPath()`. Confirmed absent via `git grep -n systemd/manager … -- '*.test.ts'` → 0 hits.
- **low · observability · `src/service/types.ts:59` + install action · ref:2302a3ae · citation-type:file-state** — the `ServiceInstallOutcome` `failed` variant carries no `configPath`, so combined with finding 1 a failed install leaves a stray unit file whose path is never surfaced in `✗ Install failed: <reason>`. Suggestion: add `configPath` to the `failed` variant and print it so the operator can `rm` it.
- **low · correctness (self-disclosed) · `src/service/systemd/paths.ts` · ref:2302a3ae** — `systemdUserDir()` hardcodes `~/.config/systemd/user`, ignoring `$XDG_CONFIG_HOME` (already listed as a follow-up in the PR body). Fails loudly (enable "unit not found") rather than silently, so low.
- **nit · api-compat · `src/cli/commands/service.ts:27-34` · ref:2302a3ae** — the import switch from `service/launchd.js` → `service/index.js` drops 7 named imports, but all 7 remain exported from the unchanged `service/launchd.ts` barrel and have **zero** other production importers (`git grep` across `src/**/*.ts` excluding tests). `[UNVERIFIED: no production importers]` — no functional break.

### Confirmed clean (read at ref 2302a3ae)

- **security — command injection:** all `systemctl` calls use `execFileSync('systemctl', [...args])` argument arrays (`install.ts:43`, `manager.ts:45`, `status.ts:72-76`) — no `exec`/`execSync`/shell-string interpolation.
- **security — secret exposure:** unit files are written `mode: 0o600` (`install.ts:65`, with an explicit comment that `Environment=` may embed tokens); no secret is logged.
- **security — path traversal:** `<name>` is constrained to `'telegram' | 'daemon'` and validated by `parseServiceName` at the CLI boundary before any path join — no `..` reaches `unitPath`.
- **api-compat:** the 7 legacy launchd symbols remain exported and unchanged; `service/launchd.test.ts` (777 lines) is untouched and still imports the free-functions — the "suite untouched / free-functions unchanged" claim holds.
- **spec-compliance — every headline clause MET:** `.path` companion unit generated (`unit.ts` `renderPathUnit` + `install.ts:132-145`, tested); lingering *surfaced as a note* and **never** auto-enabled (`LINGER_NOTE` pushed to `notes[]` only; no `loginctl` invocation anywhere); factory returns `null` for non-darwin/linux and the CLI renders a "not supported" message via the existing try/catch rather than throwing mid-pipeline; launchd adapter mapping (`plistPath→configPath`, `watchPathsActive→autoRestartOnRebuild`, restart uid/guiDomain logic) is behavior-preserving. The only spec tension is finding 1: "rollback" is delivered *partially*.
- **observability:** install `notes[]` (incl. the linger note) are plumbed to CLI output (`cli/commands/service.ts:88-90`).

### What was NOT checked

- All citations read against branch HEAD `2302a3ae` (via a worktree whose HEAD == the ref, cross-checked with `git show <ref>:<file>`); the two load-bearing findings (rollback gap, unit escaping) were independently re-derived at the ref by the synthesizer, and both test-coverage absence claims were re-confirmed by `git grep` at the ref.
- Stated intent: PR #515 title+body — spec-compliance assessed. Declared follow-ups (Phase-2 touchpoints: telegram `/proc` parse, mcp/oauth cred store, browser chrome-version, terminal-font `discoverEditors`; real-Linux smoke test; `XDG_CONFIG_HOME`; resolver hoist) were **not** treated as unmet intent.
- Did **not** run the test suite or `tsc`; the PR's "11,233 passed / tsc clean" is taken from the body + CI, not re-executed here.
- Did **not** smoke-test the systemd runtime on a live Linux box — the author discloses the real `systemctl --user` path is unit-tested only. This is precisely why findings 1 and the failure-path coverage gap are weighted as merge-blocking rather than deferred: the untested recovery path is the one that breaks first on the target platform.
- No files, commits, or PR state were modified — read-only review.

griffinwork40:
Cleanup done (worktree removed; branch ref `afk/review-pr-515` harmlessly preserved). I have everything to deliver the review.

---

# Review — PR #515: Linux systemd `--user` support behind a platform-neutral `ServiceManager`

**Verdict: Approve with follow-ups — no merge blockers.** Clean, well-reasoned abstraction. macOS behavior is byte-identical (zero regression risk), the Linux happy path (telegram) is sound and tested, and the two genuinely tricky pieces (rollback across every failure branch; the `.path`→oneshot restart indirection) are correct.

### Verification I ran (worktree on the PR branch, now removed)
- `tsc --noEmit` (strict tsconfig): **PASS** (exit 0)
- `vitest run src/service/`: **71/71 PASS** — launchd **46 unchanged/green (no regression)**, systemd **21 new**, index **4 new**
- Read `launchd/plist.ts` + `launchd/paths.ts` directly to confirm the reused resolvers and `guiDomain()`/timeout parity.

Reproduce: `git checkout afk/linux-compat-494 && pnpm install && pnpm lint && pnpm exec vitest run src/service/`.

---

### MEDIUM (worth addressing in this PR or a fast follow-up)

1. **Linux `daemon` install can fail to locate the `afk` binary** — `launchd/plist.ts:189-239` (reused unchanged by systemd). `resolveAfkBinary` candidates (`/usr/local/bin/afk`, `/opt/homebrew/bin/afk`) and `TRUSTED_BIN_PREFIXES` are macOS-centric. A Linux user who installed via the project's own package manager (`pnpm add -g` → `~/.local/share/pnpm`) or nvm/linuxbrew lands **outside the allowlist**, so `afk service install daemon` throws `Could not locate the 'afk' binary`. Telegram is unaffected (uses the dist entrypoint). Not silent, not a blocker — but it means "Linux daemon support" is incomplete for common setups. *Fix: add Linux global-bin dirs (`~/.local/bin`, `~/.local/share/pnpm`, `/home/linuxbrew/.linuxbrew/bin`) to candidates/allowlist, or relax the allowlist under `$HOME` on Linux.*

2. **`resolveSystemctlPath` drops launchd's binary-hardening** — `systemd/install.ts:58-73`. Launchd realpath+allowlist-checks the binary baked into a persistent config (`plist.ts:189-276`); the systemd equivalent checks only `existsSync` on two absolute paths, else falls back to **bare `'systemctl'`**, baking a PATH-resolved command into a persistent `ExecStart=`. Low likelihood (both `/usr/bin`+`/bin` missing is rare on systemd hosts) but it's the one hardening the PR's own comments claim to "mirror." *Fix: fail the install if no absolute `systemctl` is found (fail-closed), rather than emitting a bare-name fallback.*

3. **`StandardOutput=append:` needs systemd ≥240** — `systemd/unit.ts` (comment acknowledges it, no guard). Realistic base (Ubuntu 20.04+, Debian 11+) is fine; older LTS/RHEL-7 would fail. *Fix: document the minimum, or degrade to a non-`append:` sink on old versions.*

### LOW / NIT (cheap defense-in-depth; safe today because inputs are bounded)

- **Unescaped fields** — `Description=`, `renderRestartUnit`'s `ExecStart=`, `renderPathUnit`'s `PathModified=`/`Unit=`, and the `Environment=` **key** are written raw (`unit.ts`). All currently fed bounded values (`name ∈ {telegram,daemon}`, resolved paths) so no injection today, but latent if inputs ever widen. Escape `Description`, quote the restart `ExecStart` args, validate env keys (`^[A-Za-z_][A-Za-z0-9_]*$`).
- **Stale `.tmp` wedges reinstall** — `atomicWrite` uses a fixed `${path}.tmp` with `O_EXCL`; a leaked tmp from a crashed run makes the *next* install fail at write with no self-cleanup (launchd shares this gap). *Fix: unlink stale tmp on `EEXIST`, or random-suffix it.*
- **`resolveServicePath` bleeds macOS Homebrew dirs into the Linux unit PATH** — harmless (the Linux node dir is still prepended first, so node resolves), just non-idiomatic.
- **Dual `SERVICE_NAMES`/`ServiceName`** in `types.ts` + `launchd/paths.ts` — no collision (verified), but no test asserts the two lists stay in sync. Optional: assert equality or re-export the type.
- Secrets note: the CLI never passes `environment`, so **no API tokens are embedded in units** (read at runtime from `~/.afk/config/afk.env`); unit + tmp files are `0o600` + `O_EXCL`. The `~/.config/systemd/user` dir is created `0o755` (dir-listing leak only; identical to launchd's posture).

### Test-coverage gaps (recommend before merge; both cheap)
- **launchd adapter (`launchd/manager.ts`) mapping is untested** — index.test only asserts `label()`/`configPath()`. The `dryRun→skipBootstrap`, notes, and outcome-kind translation ship without a guard. Mitigated (mapping verified by reading = lossless, underlying funcs tested, macOS parity byte-verified), but this is the PR's headline abstraction — add a mapping test.
- **systemd `noWatch: true` path untested** — mirror the dev-tree test with `noWatch:true` and assert `!existsSync(pathUnitPath)`.
- Follow-up (non-blocking): entrypoint-resolution-failure branch, `environment` override, `systemdStatus` live path / `LoadState=not-found`, `errorDetail` stderr surfacing.

### Confirmed correct (strengths)
Rollback removes all three unit files on `enable` failure (tested, `systemd.test.ts`); the oneshot-restart indirection is right (`start` on an active `Restart=always` unit is a no-op) and correctly has no `[Install]`; escaping is correct where applied with strong exact-string tests; all `systemctl` calls use no-shell `execFileSync` with array args; macOS restart/install/status/uninstall delegate unchanged; skill prompts updated in lockstep and the CLI substrings the skill greps for (`Running  (PID`, `Installed but not running`) still match.

### Residual (honest, low-risk)
- systemd unit-escaping semantics were reasoned from spec knowledge — the live `systemd.service(5)` fetch was blocked by an anti-scrape wall. The escaping is conservative (quote every arg + escape `\ " \n \r %`), so risk is low, but not freshly citation-backed this session.
- No live Linux host: `systemctl --user`, `network-online.target` under `--user`, and `append:` on the target systemd version were not executed.

---

**Done.**
- Reviewed PR #515 (+1472/-112, 15 files) across correctness, security, cross-platform regression, and test coverage via four parallel read-only reviewers + my own typecheck/test run.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/40799c1f-211e-494e-a0ce-084ae1124d66/pr-515.diff`; `tsc --noEmit` exit 0; `vitest run src/service/` 71/71 pass — both run in a temporary worktree on branch `afk/linux-compat-494` (now removed, branch ref preserved).
- Verdict: **approve, no blockers**; recommended before merge: 2 cheap tests (launchd-adapter mapping, systemd `noWatch`) + fail-closed `systemctl` path; recommended follow-up: Linux `afk`-binary resolution (MEDIUM completeness).
- This review is read-only — I did **not** edit, commit, comment on the PR, or touch your dirty `main` working tree (still 5 uncommitted changes, untouched).
- Side effect to note: my `pnpm install` in the review worktree triggered the repo's `postinstall` hook (`scripts/postinstall.mjs`), which **restarted your live `telegram` + `daemon` services** onto the current build. They're running normally; no action needed — just flagging that the review bounced them.

---

Issue #516, opened 2026-07-12
feat(tui): live "during-pending" parallel-batch indicator (Phase 2, optional)


griffinwork40:
## Summary (optional enhancement / feature request)

Phase 1 (branch `afk/tui-batch-badge`, PR incoming) adds a **durable per-row badge** marking each tool-call row with its dispatch-batch width, so parallel-vs-sequential execution becomes legible in the REPL tool-lane scrollback **and** in `afk trace show` / `events.jsonl`. Because the batch structure is only authoritative *after* the dispatcher partitions calls, Phase 1 carries the data on the **completed** event — so the badge appears as each row *resolves*.

**Gap this issue tracks:** during the **pending** window — while multiple roots are still spinning — a parallel batch and a run of sequential singletons still look identical. The "two spinners at once = parallel" distinction is only visible live, and Phase 1's badge doesn't light up until a row completes.

## Why this was deferred (not a quick follow-up)

Sourcing batch membership at *emit / pending* time is not cheap:

- `tool.use.start` events fire **upfront for the whole turn**, *before* `partitionIntoBatches` runs (inside `SessionToolDispatcher.executeBatch`, "Phase 2"). So the started event cannot carry an accurate batch width today.
- Hoisting the partition earlier is not a one-liner: `loop.ts` has no access to the dispatcher's real (post-filter) `concurrencyClassifier` (it's private and per-dispatcher overridable), and a *pre-filter* preview can **diverge** from the authoritative partition when a call is stripped by a PreToolUse hook / permission / bash-gate / repeat-breaker.

## Recommended approach

Emit a lightweight **`batch-start` marker from *inside* `executeBatch`** at the top of each sequential `for…of` iteration — where the partition is already authoritative and post-filter (no classifier exposure, no divergence risk). Have the live overlay bracket or badge the **currently-executing** batch during its pending window. This is a new event plus a small live-overlay consumer — materially more than Phase 1's additive field, hence a separate, optional track.

## Constraints to respect (same as Phase 1)

- **Append-only scrollback:** committed rows can't be re-laid-out, and batch members commit at different times. Any live bracket must not require inserting content above already-committed lines.
- Don't change batching/concurrency behavior or the `concurrencySafe` flags.

## Priority

Optional / nice-to-have. Phase 1 already resolves the ambiguity accurately for the post-hoc and screenshot cases (the original motivation). This issue only adds the *live, before-completion* distinction.

---

[...]

PR #520, opened 2026-07-12
feat(tui): distinguish parallel vs sequential tool dispatch with ∥i/N batch badge

griffinwork40:
## Problem

Sequential and parallel tool dispatch are **visually indistinguishable** in the TUI. The tool-lane renders roots by emission-order topology + done/in-flight lifecycle, but never encodes *batch membership*. So a fast root that finishes and commits to append-only scrollback, followed by a slow root still spinning below it, looks identical whether the two ran **concurrently** (one safe batch) or **back-to-back** (two sequential batches). This repeatedly reads as "these ran in parallel" when they were actually sequential (e.g. an `agent` that finished, then a `bash` that started after it).

The distinguishing signal is temporal (two roots spinning at the same instant), which a static frame — or scrollback — cannot capture.

## Approach

Carry the dispatcher's **already-known** batch membership on each result and render it, rather than trying to infer concurrency from liveness (which over-reports: all start-events fire before partitioning, and `bash` always splits into its own batch).

**Data source** — `SessionToolDispatcher.executeBatch` stamps `batchIndex` (1-based) + `batchSize` (calls dispatched together) onto each `ToolResult`, sourced from the authoritative post-partition batch. `batchSize > 1` ⇒ ran in a parallel wave; `== 1` ⇒ ran alone (always the case for concurrency-unsafe tools like `bash`). Follows the existing `failureClass`/`circuitBreaker` metadata precedent exactly.

**Trace** — threaded onto the `tool_call` completed event (both providers) + schema, so the distinction is machine-verifiable via `afk trace show` and telemetry.

**TUI** — plumbed through `tool.output` → stream-consumer → `ToolResultChunk`, rendered as a dim `∥i/N` suffix on root rows whose `batchSize > 1`.

Because the badge is sourced from the real batch, it is **collision-incapable**: it can only appear on a call that genuinely shared a batch, never on a sequential dispatch.

### Rendered

```
  ● Read("dispatcher.ts") — ✓ 970 lines ∥1/3
  ● Grep("batchSize") — ✓ 12 matches ∥2/3
  ● Glob("**/*.ts") — ✓ 44 paths ∥3/3

  $ Bash("pnpm test") — ✓ 42 lines          ← singleton batch, never badged
```

## Scope

- **Badged:** flat roots (live overlay + committed scrollback) and NESTING roots (agent/skill/compose) in the live overlay.
- **Deferred:** NESTING-root badge in committed **scrollback**. The agent head row is governed by the `formatAgentSummary`/`formatAgentHeader` dual-encoding invariant, and `batchSize` isn't available at eager-head-emit time (the agent hasn't completed yet) — badging it there risks the severed-spine class of bug. Tracked as a follow-up.
- **Not badged:** the `×N` grouped header (already signals multiplicity — avoids a double-signal).

## Testing

- `executeBatch` stamping (mixed partition + full safe fan-out)
- `batchBadge` unit tests (parallel/singleton/absent/defensive)
- End-to-end lane render — badge reaches rendered root rows in both overlay and flush; singleton shows no badge
- Trace-schema acceptance/rejection of `batchIndex`/`batchSize`
- Full affected subtree green: **7228 tests**, `tsc --noEmit` clean
- Visually confirmed the `∥` glyph renders (shown above)

Additive/back-compat throughout: absent on the single-tool `execute()` path and on blocked/short-circuited calls; badge is invisible unless `batchSize > 1`.

## Follow-ups
- #516 — live "during-pending" parallel indicator (Phase 2)
- Scrollback NESTING-root badge (deferred above)

### DIFF
--- a/src/agent/provider.ts
+++ b/src/agent/provider.ts
@@ -157,6 +157,16 @@ export type ProviderEvent =
        * preview path.
        */
       toolName?: string;
+      /**
+       * Concurrency-batch membership, plumbed from `ToolResult.batchIndex` /
+       * `.batchSize` (set by the dispatcher's `executeBatch`). `batchSize > 1`
+       * means this call ran in a parallel wave; `=== 1` (or absent) means it
+       * ran alone. The interactive tool-lane reads these to badge a parallel
+       * wave distinctly from back-to-back sequential dispatch. Optional: absent
+       * on the single-tool `execute()` path and on providers that don't batch.
+       */
+      batchIndex?: number;
+      batchSize?: number;
     }
   | {
       /**
--- a/src/agent/providers/anthropic-direct/loop.ts
+++ b/src/agent/providers/anthropic-direct/loop.ts
@@ -655,6 +655,9 @@ export async function* runTurn(
         durationMs,
         ...(result.circuitBreaker === true ? { circuitBreaker: true } : {}),
         ...(result.failureClass ? { failureClass: result.failureClass } : {}),
+        ...(typeof result.batchIndex === 'number' && typeof result.batchSize === 'number'
+          ? { batchIndex: result.batchIndex, batchSize: result.batchSize }
+          : {}),
       });
 
       yield {
--- a/src/agent/providers/anthropic-direct/loop.ts
+++ b/src/agent/providers/anthropic-direct/loop.ts
@@ -667,6 +667,9 @@ export async function* runTurn(
         content: result.content,
         ...(result.isError === true ? { isError: true } : {}),
         ...(truncated ? { truncated: true } : {}),
+        ...(typeof result.batchIndex === 'number' && typeof result.batchSize === 'number'
+          ? { batchIndex: result.batchIndex, batchSize: result.batchSize }
+          : {}),
         sessionId: input.ctx.sessionId,
       };
 
--- a/src/agent/providers/anthropic-direct/types.ts
+++ b/src/agent/providers/anthropic-direct/types.ts
@@ -105,6 +105,26 @@ export interface ToolResult {
    * See {@link import('../../trace/types.js').ToolFailureClass}.
    */
   failureClass?: ToolFailureClass;
+  /**
+   * 1-based position and total size of the concurrency batch this call was
+   * dispatched in, stamped by {@link SessionToolDispatcher.executeBatch}.
+   * `batchSize > 1` means the call ran in a parallel wave alongside
+   * `batchSize - 1` sibling calls; `batchSize === 1` (or absent) means it ran
+   * alone in its own sequential batch — which is the case for EVERY
+   * concurrency-unsafe tool (bash, write_file, edit_file, …), since the
+   * partitioner isolates each into a singleton batch.
+   *
+   * Render-only + trace metadata, never forwarded to the model. Lets the TUI
+   * and `afk trace show` distinguish a genuine parallel wave from back-to-back
+   * sequential dispatches, which are otherwise visually identical once
+   * committed to scrollback. Note `batchSize` is the number of calls
+   * DISPATCHED together, not the number that ran at literally the same instant:
+   * a safe batch wider than `maxConcurrentSafeCalls` drains through the pool in
+   * sub-waves, but is one logical batch here. Absent when the dispatcher ran
+   * the non-batching `execute()` path (single-tool fallback).
+   */
+  batchIndex?: number;
+  batchSize?: number;
   render?: RenderHints;
   /**
    * Structured test-runner result parsed from bash output by
--- a/src/agent/providers/openai-compatible/query/dispatch-append.ts
+++ b/src/agent/providers/openai-compatible/query/dispatch-append.ts
@@ -176,6 +176,9 @@ export async function* dispatchAndAppendToolCalls({
         durationMs,
         ...(result.circuitBreaker === true ? { circuitBreaker: true } : {}),
         ...(result.failureClass ? { failureClass: result.failureClass } : {}),
+        ...(typeof result.batchIndex === 'number' && typeof result.batchSize === 'number'
+          ? { batchIndex: result.batchIndex, batchSize: result.batchSize }
+          : {}),
       });
 
       yield {
--- a/src/agent/providers/openai-compatible/query/dispatch-append.ts
+++ b/src/agent/providers/openai-compatible/query/dispatch-append.ts
@@ -188,6 +188,13 @@ export async function* dispatchAndAppendToolCalls({
         content: result.content,
         ...(result.isError === true ? { isError: true } : {}),
         ...(result.truncated === true ? { truncated: true } : {}),
+        // Plumb concurrency-batch membership onto the render-facing event, not
+        // just the trace event above, so the TUI `∥i/N` badge works here too.
+        // Parity with anthropic-direct/loop.ts's tool.output yield — omitting it
+        // silently drops the badge for every openai-compatible session.
+        ...(typeof result.batchIndex === 'number' && typeof result.batchSize === 'number'
+          ? { batchIndex: result.batchIndex, batchSize: result.batchSize }
+          : {}),
         sessionId,
       };
       if (result.render?.diff) {
--- a/src/agent/providers/openai-compatible/tool-dispatch.test.ts
+++ b/src/agent/providers/openai-compatible/tool-dispatch.test.ts
@@ -22,7 +22,7 @@ import type { AgentConfig } from '../../types/config-types.js';
 import { SessionToolDispatcher } from '../../tools/dispatcher.js';
 import { createHookRegistry } from '../../hooks.js';
 import type { AnthropicToolDef } from '../anthropic-direct/types.js';
-import type { ToolHandler } from '../../tools/types.js';
+import type { ToolHandler, ConcurrencyClassifier } from '../../tools/types.js';
 import {
   __setOpenAIClientFactory,
   OpenAICompatibleQuery,
@@ -90,7 +90,10 @@ interface DispatcherFixture {
   postHookFired: string[];
 }
 
-function makeDispatcher(opts?: { allowedTools?: string[] }): DispatcherFixture {
+function makeDispatcher(opts?: {
+  allowedTools?: string[];
+  concurrencyClassifier?: ConcurrencyClassifier;
+}): DispatcherFixture {
   const handlerCalls: DispatcherFixture['handlerCalls'] = [];
   const preHookFired: string[] = [];
   const postHookFired: string[] = [];
@@ -137,6 +140,9 @@ function makeDispatcher(opts?: { allowedTools?: string[] }): DispatcherFixture {
   };
   if (opts?.allowedTools !== undefined) {
     dispatcherOpts.permissions = { allowedTools: opts.allowedTools };
+  }
+  if (opts?.concurrencyClassifier !== undefined) {
+    dispatcherOpts.concurrencyClassifier = opts.concurrencyClassifier;
   }
 
   const dispatcher = new SessionToolDispatcher(dispatcherOpts);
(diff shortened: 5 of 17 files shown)


griffinwork40:
Filed the deferred scrollback NESTING-root badge as #521 (with the dual-encoding invariant blocker + candidate approaches documented).

griffinwork40:
Now it is **fully settled by direct reads at the reviewed ref** `ad7e876c`:

- **anthropic-direct** `loop.ts` — the `tool.output` yield **includes** the guarded batch spread (lines A16–A18). `batchIndex` appears 4× in the file (trace event + tool.output).
- **openai-compatible** `dispatch-append.ts` — the main `tool.output` yield **omits** it (lines O2–O10: only content/isError/truncated/sessionId). `batchIndex` appears 2× — both in the trace-event spread only.
- **main @ 68f066e** — 0 matches, confirming the shadow-verify agent read the wrong tree (it's a research-agent with no Bash, so it couldn't `git show` the un-checked-out PR ref; its REFUTED is a ref-mismatch false negative, now resolved to CONFIRMED).

The asymmetry is real and the data is available on `result` (the same function uses `executeBatch` at line 121 and reads `result.batchIndex` for the trace event). Everything else in the change is clean. Here is the review.

---

# Code Review — PR #520

`feat(tui): distinguish parallel vs sequential tool dispatch with ∥i/N batch badge`

**Target:** PR #520 · branch `afk/tui-batch-badge` → `main` · reviewed ref `ad7e876c8b814255e3b9105febe658295ef52637`
**Change type:** feature · 16 files · +290 / −4 (≈294 lines)
**Stated intent:** present (PR title + body) — spec-compliance assessed.
**Verdict:** 🔴 **DO NOT MERGE** — 1 blocking (high) finding; trivially fixable.

---

## Blocking

### 🔴 HIGH · confidence: high · correctness / spec-compliance
**`src/agent/providers/openai-compatible/query/dispatch-append.ts:183-190`** · ref:`ad7e876c` · citation-type: file-state

The openai-compatible provider's main-dispatch `tool.output` event **omits** `batchIndex`/`batchSize`, so the PR's `∥i/N` badge — its primary, title-level deliverable — **silently never renders** for any openai-compatible session (`gpt-*`, `o1/o3/o4*`, `codex-*`, and every local HuggingFace-style model: `mlx-community/…`, `Qwen/…` via MLX/llama.cpp/vLLM/ollama-openai). Anthropic-direct sessions get the badge; ~half the supported provider surface does not, with no error and no disclosure.

The data is available — the same function dispatches via `executeBatch` (line 121, so `result` carries the fields) and stamps them onto the trace event 3 lines above the yield.

Evidence — the openai-compatible yield (missing) vs. the anthropic-direct yield (`loop.ts:663-673`, present):
```ts
// dispatch-append.ts:184 — MAIN dispatch path, batch fields ABSENT
yield {
  type: 'tool.output', toolUseId: call.id, toolName: call.name,
  content: result.content,
  ...(result.isError === true ? { isError: true } : {}),
  ...(result.truncated === true ? { truncated: true } : {}),
  sessionId,                                    // ← no batchIndex/batchSize
};
// loop.ts:671 — anthropic-direct, batch fields PRESENT
  ...(typeof result.batchIndex === 'number' && typeof result.batchSize === 'number'
    ? { batchIndex: result.batchIndex, batchSize: result.batchSize } : {}),
```
**Suggestion:** add the identical guarded spread to the `dispatch-append.ts:184` yield — mirroring the trace-event spread already present at line ~180 in the same function and the `tool.output` spread at `loop.ts:671`. One line; the data is already on `result`.

*Note on severity:* impact is a cosmetic TUI badge — no data loss, no crash; the `tool_call` **trace** event is stamped correctly on both providers, so `afk trace show`/telemetry are unaffected. It's rated high because it silently defeats the PR's stated purpose for a first-class, reachable provider class (semantic "defeats stated purpose" finding).

---

## Non-blocking

### 🟡 LOW · confidence: high · test-coverage
**`dispatch-append.ts` / `loop.ts` (provider `tool.output` plumbing)** · ref:`ad7e876c`

No test asserts the provider `tool.output` event carries `batchIndex`/`batchSize` — a per-provider passthrough test would have caught the blocking finding. Every existing test exercises the pieces in isolation and injects a `ToolResultChunk` with batch fields **directly** (`tool-lane.test.ts` lines 444-483), stamps at the dispatcher (`dispatcher.test.ts`), or validates the schema/badge (`events.test.ts`, `tool-lane-format.test.ts`) — none exercise the provider event → stream-consumer plumbing where the gap lives.
**Suggestion:** add a stream-consumer or provider-loop test asserting a batched `result` (`batchSize > 1`) yields a `tool.output` → `ToolResultChunk` that carries the fields, parameterized over both providers.

---

## Clean (read at ref `ad7e876c`)

- **`dispatcher.ts:813-818` (stamping)** — no issues. `batch.indices` index into `executableCalls`, exactly matching the execution loops' `executableCalls[batchIdx].originalIndex`; `batchIndex = pos+1` is a correct 1-based ordinal; `batchSize = batch.indices.length`.
- **`dispatch-batching.ts:49-63` (partition)** — no issues. A batch is only extended when `safe && last.isConcurrencySafe`, so every concurrency-unsafe call is isolated as its own singleton — the "`batchSize > 1` ⇒ parallel wave" invariant holds. (Documented caveat: a safe batch wider than `maxConcurrentSafeCalls` drains in sub-waves but is one logical batch — honest per the doc comment.)
- **`stream-consumer.ts:196-247`** — no issues. Both `ToolResultChunk` return branches (persisted + normal) thread `batchPassthrough`; no dropped path.
- **`tool-lane-format.ts:399-409` (`batchBadge`)** — no issues. Returns `''` for absent/singleton/half-present; `∥i/N` only when `batchSize > 1`. Defensive half-present case is tested.
- **`tool-lane.ts:499,512` + `tool-lane-render-agent.ts:216`** — no issues. Overlay badges flat + NESTING roots; scrollback badges flat single-entry roots. The `×N` grouped header (`entries.length > 1`) and NESTING-scrollback path are correctly **not** badged, matching the PR's stated "deferred / avoid double-signal" scope.
- **`trace/events.ts:51-52` (Zod schema)** — no issues. `z.number().int().positive().optional()` is additive/back-compat (pre-existing persisted trace events without the fields still parse); `.positive()` matches the always-≥1 producer; accept/reject tested.
- **`provider.ts`, `trace/types.ts`, `message-types.ts` (interface additions)** — no issues. All optional fields; additive, no breaking change to any consumer.

## Spec-compliance verdict

**Partially met — undisclosed provider gap.** The trace threading is complete on both providers (the body's "both providers" claim holds for the trace). The TUI badge — described provider-agnostically as "plumbed through `tool.output` → stream-consumer → `ToolResultChunk`" — works for anthropic-direct only. This is an **unmet intent** for the openai-compatible provider class, and it's not listed in the PR's "Deferred / Not badged" Scope section. Fixing the one-line omission satisfies the intent as written; alternatively the author must consciously scope openai-compatible out and document it — but the fix is strictly preferable given the data is already present.

## What was not checked

- Read all findings against branch HEAD `ad7e876c` via `git show <ref>:<file>` — citations verified at that ref. (The dirty working tree on `main` was never mutated.)
- Did **not** run the test suite (`pnpm test`) or `tsc --noEmit`; the author reports 7228 green + tsc clean, unverified here.
- Did not exercise the badge in a live TUI session on either provider (verification is by code read + the PR's own render-path tests).
- **Process note:** Wave 1 review sub-agents both timed out at 20 min (recursive over-decomposition on a small diff); this review was conducted inline against the fully-captured diff. The shadow-verify pass initially returned REFUTED because a research-agent cannot `git show` an un-checked-out ref (no Bash) and read `main` instead — the finding was then confirmed by direct ref reads. If you re-run verification with a Bash-capable agent, target ref `ad7e876c` explicitly, since the PR branch is not checked out.

## Blocked

- **What blocks the merge:** the openai-compatible `tool.output` event (`dispatch-append.ts:184`) omits `batchIndex`/`batchSize`, so the PR's `∥i/N` badge silently does not render for gpt/codex/o-series/local-model sessions.
- **What must change to unblock:** add the guarded `...(typeof result.batchIndex === 'number' && typeof result.batchSize === 'number' ? { batchIndex: result.batchIndex, batchSize: result.batchSize } : {})` spread to the `dispatch-append.ts:184` `tool.output` yield (mirror `loop.ts:671`), and add a provider-plumbing test (low finding). This review made no edits — read-only.
- **Already done:** full read-level review at ref `ad7e876c`; blocking finding independently confirmed via direct ref reads after a shadow-verify ref-mismatch false negative; all other dimensions verified clean.

---

griffinwork40:
All checks resolved and independently verified at the ref. No critical/high findings exist, so Wave 1.5 citation/absence verification has no in-scope findings to re-derive, and `/shadow-verify` is not triggered (medium-and-below pass straight through) — but I directly re-derived the one load-bearing correctness claim myself rather than trusting the agents' inference. Synthesis follows.

---

# Code Review — PR #520

**`feat(tui): distinguish parallel vs sequential tool dispatch with ∥i/N batch badge`**
`afk/tui-batch-badge` → `main` · head `a795d05` · feature · 17 files · +360/−6 · regime **FULL**

## Verdict per dimension

| Dimension | Verdict |
|---|---|
| security | **CLEAN** |
| api-compat | **CLEAN** |
| correctness | **CLEAN** (partitioner + stamping loop directly verified) |
| spec-compliance | **CLEAN** (all 5 intent clauses implemented; deferrals honored) |
| test-coverage | **2 × low** |
| perf-observability | **CLEAN** |

## Findings (sorted; no critical / high / medium)

**low · high-confidence · test-coverage · `src/agent/providers/anthropic-direct/loop.ts:658-672` · ref:a795d05 · file-state**
The anthropic-direct forwarding of `batchIndex`/`batchSize` onto both the trace event and the `tool.output` yield has no test, while the byte-identical openai-compatible forwarding got a dedicated regression test (`tool-dispatch.test.ts`).
Evidence: `...(typeof result.batchIndex === 'number' && typeof result.batchSize === 'number' ? { batchIndex: result.batchIndex, batchSize: result.batchSize } : {})` — appears twice (trace + yield); grounded by `git grep -n batchIndex a795d05 -- src/agent/providers/anthropic-direct` → zero test hits.
Suggestion: add a `loop.test.ts` case mirroring the openai `stamps batchIndex/batchSize onto tool.output` test.

**low · high-confidence · test-coverage · `src/agent/session/stream-consumer.ts:216,247` · ref:a795d05 · file-state**
Both `buildToolOutputEvent` return paths (parsed-persisted branch + default branch) correctly spread `batchPassthrough`, but neither is locked by a direct assertion, so a future refactor could silently drop it from one branch.
Evidence: `...displayPassthrough, ...batchPassthrough,` at line 218 (persisted) and line 247 (default); grounded by zero `batchIndex` hits in any `stream-consumer*.test.ts`.
Suggestion: add a stream-consumer test asserting the chunk retains `batchIndex`/`batchSize` through both branches.

**nit · high-confidence · correctness · `src/agent/tools/dispatcher.ts:634-635` vs `:807-820` · ref:a795d05 · file-state**
The stamping-loop comment says a concurrency-unsafe tool "lands batchSize=1 and is never badged," but a tool called **alone** in its round takes the `if (calls.length === 1) return [await this.execute(...)]` shortcut and never reaches the stamping loop, so it gets **absent** fields, not `1`. Purely cosmetic — `batchBadge` treats absent and `1` identically (both render no badge).
Evidence: `if (calls.length === 1) return [await this.execute(calls[0]!)];` (line 635) precedes the stamping loop.
Suggestion: reword the comment to say singletons carry `batchSize=1` *within a multi-call round* (a lone call is unstamped/absent).

## Verified load-bearing claims

- **Badge never reaches the model** (security): the model-facing `tool_result` is built by explicit destructure (`loop.ts:691-697` → `content`/`isError`/`image` only) and `toolResultsToMessages` (openai `loop.ts:157-165` → `content`/`tool_call_id` only). All 32 `batchIndex` hits across `src/agent` are type/schema defs, `tool.output`/trace yields, render paths, or tests — none in a model-request builder.
- **Ordinal is monotonic with render order & collision-incapable** (correctness): `partitionIntoBatches` (`dispatch-batching.ts:48-62`) pushes consecutive safe calls into `indices` in ascending call order, so `batch.indices` is strictly ascending → `executableCalls[batchIdx].originalIndex` rises with `pos` → `batchIndex = pos+1` renders top-to-bottom in order. Results stay keyed by `originalIndex`, completion-order independent. `bash`/unsafe tools always start a fresh singleton batch (`batchSize=1`).
- **Additive / back-compat** (api-compat): every added field is `?`-optional across `provider.ts`, `anthropic-direct/types.ts`, `trace/types.ts`, `message-types.ts`; the Zod schema uses `.int().positive().optional()` (`trace/events.ts:54-55`) so pre-change trace files still parse. `concurrencyClassifier` is pre-existing dispatcher surface (`dispatcher.ts:100`), only newly forwarded by a test fixture.
- **Spec-compliance**: all five intent clauses land; explicitly-deferred scope (NESTING-root scrollback badge, `×N` grouped header) is confirmed absent, live-overlay NESTING roots are badged; no scope creep (every hunk additive).

## What was not checked

- Wave 1 agents + my own reads were **all against branch HEAD `a795d05`** — citations verified at that ref.
- **Stated intent**: PR #520 title + body — spec-compliance **assessed**.
- The security·api-compat agent **timed out at 20 min** on its first dispatch; re-run under a tight budget completed CLEAN. Its incidental claim "this PR never touches `dispatcher.ts`" was **wrong** (used a single-commit `git show`, not the `main...a795d05` range — commit `36a760c` touches it); I corrected it and independently confirmed the API-COMPAT-3 conclusion still holds.
- No tests were executed (static review only); the PR body self-reports 7228 tests green + `tsc --noEmit` clean — not independently run here.
- Runtime/visual confirmation of the `∥` glyph rendering was not performed (static-only).

## Merge decision: **MERGE**

Clean, additive, back-compat feature with the load-bearing correctness invariant (in-order, collision-incapable ordinal) directly verified against source. No critical/high/medium findings. The two `low` items are symmetric-coverage nice-to-haves (the rendering path and the openai forwarding path are both tested; only the anthropic forwarding and stream-consumer passthrough lack direct assertions) and the `nit` is a comment-wording tweak — none block merge.

*Read-only review complete. No files, commits, or PR state were modified.*

---

PR #514, opened 2026-07-11
fix(subagent): explicit writeRoots grant + truthful deny for forked children (#435)

griffinwork40:
## Summary

Closes #435 — the write-side sibling of #416.

Forked worktree subagents were auto-denied on **any** write outside their worktree (`path-approval-hook` `parentSessionId` auto-deny), and because forked children are non-interactive they couldn't elicit — so the write dead-ended silently with a deny message that named no real remedy. #416 granted the main repo root as a *read* root for forks but deliberately left write confined to `[worktree]`; there was no mechanism for a parent to widen a fork's write scope.

This lands the full agreed fix:

- **(a) opt-in `writeRoots` param on the `agent` tool.** Additional absolute directories a forked child may write to. It is **composed with** the child's cwd (`[cwd, ...writeRoots]`, deduped) — never replaces it, so a fork can't lose the ability to write its own tree. Validated per-entry (absolute, non-empty, no `..`), and **mutually exclusive with `isolation: 'worktree'`** (mirrors the existing `cwd`+`isolation` rejection — punching a write hole out of an isolated worktree breaks the merge/teardown contract).
- **(b) truthful, mode-aware deny message.** Instead of a silent dead-end, the deny now names the concrete remedy: for a write, re-dispatch with `writeRoots: [path]`; for a read, widen `readRoots`.
- **(c) docs.** Tool schema + code comments state that writes anchor to the child's cwd/worktree by default and out-of-tree writes are denied.

**Policy preserved:** write-confinement stays the default. `writeRoots` is **never auto-granted** (unlike the #416 read grant) — the main root is safe to auto-grant for reads but not writes, since auto-granting writes would reintroduce the cross-writer collisions worktree isolation exists to prevent. Both providers (`anthropic-direct`, `openai-compatible`) already consume `config.writeRoots` on fork init, so the grant works end-to-end for both.

## How was this tested?

Run on this exact tree (`d764554`):

- `pnpm lint` (tsc --noEmit, strict) — clean
- `pnpm test` — **11228 passed / 14 skipped / 0 failed** (619 files)
- `pnpm audit:sdk:check` — OK (no unlocked SDK symbols)
- `pnpm audit:env:check` — 0 violations

New unit coverage added:
- `input-parse.test.ts` — `writeRoots` parse/validation + `isolation` mutual-exclusion
- `path-approval-hook.test.ts` — mode-aware deny remedy text
[...]

Issue #435, opened 2026-07-05
subagent: forked children auto-denied when writing outside their worktree (write-side sibling of #416)


griffinwork40:
## Summary

#416 fixed forked worktree subagents being unable to **read** main-repo paths (it grants `readRoots = [cwd, mainRoot]`). The **write** direction is still locked: `writeRoots` stays `[worktree]`, and any write outside it hits the path-approval hook's `parentSessionId` auto-deny — forked children cannot elicit, so there is no recovery. If the write is legitimate and the parent *also* lacks the grant, the "report to parent" fallback is unactionable and the child dead-ends silently.

This is the write-side sibling of #416.

## Evidence

- `src/agent/tools/hooks/path-approval-hook.ts:225-237` — `parentSessionId` auto-deny; no distinction between a restricted forked child and legitimate nested work
- `src/agent/worktree-read-root.ts:20-22` — main repo granted as a READ root only, never write (intentional isolation)
- `src/agent/subagent.ts:565-569` — the #416 grant is threaded into `readRoots` only

## Why it matters

Write confinement to the worktree is correct isolation, but the failure mode is **silent + unrecoverable** for the child. Flows where a subagent is legitimately asked to write outside its worktree (e.g. update a main-repo file, write a sibling path) dead-end with no actionable signal.

Related: #201 (AFK gate workspace-root vs subagent worktree cwd — possible over-block).

## Proposed direction

Decide the policy, then implement:
- (a) let the parent pre-grant specific write roots to a fork at dispatch (explicit `writeRoots` param on the `agent` tool), **or**
- (b) at minimum, replace the silent auto-deny with an actionable block message naming the exact path and how the parent can grant it, **or**
- (c) document the confinement clearly so callers scope fork work inside the worktree.

Severity: MED (silent unrecoverable failure; by-design isolation but poor ergonomics).

Found via a read-only tool/subagent/worktree audit (follow-up to #416).

PR #522, opened 2026-07-12
fix(tools): give read/write/edit/list the factory-cwd resolve tier glob/grep have (#434)

### LINKED ISSUE #434: tools: read_file/write_file/edit_file/list_directory lack the factory-cwd fallback that glob/grep have (worktree-isolation asymmetry)
## Summary

`glob` and `grep` are registered through `createGlobHandler(cwd)` / `createGrepHandler(cwd)`, which close over the session/worktree cwd as a *fallback* base when the model omits a path. `read_file`, `write_file`, `edit_file`, and `list_directory` are registered as **bare** handlers with no cwd closure — they rely entirely on `context.resolveBase` / `context.cwd` being populated by the dispatcher. If that field is ever undefined, `resolveAndContain` falls back to `process.cwd()` (the host launch dir), so a relative path resolves against the wrong tree.

This is the structural sibling of the #416 worktree-read bug: four FS tools are missing the worktree-isolation safety net the other two have.

## Evidence

- `src/agent/tools/handlers/index.ts:62-63` — glob/grep get the factory cwd
- `src/agent/tools/handlers/index.ts:68-73` — read_file / write_file / edit_file / list_directory registered bare (no cwd)
- `src/agent/tools/handlers/glob.ts:167`, `grep.ts:70-72` — the `?? cwd` factory tier the other four lack
- `src/agent/tools/handlers/_cwd-utils.ts:82,87` — `resolveBase = context?.resolveBase ?? context?.cwd`; falls to `process.cwd()` when both undefined

## Severity / scope

MED, but **latent / defense-in-depth** rather than an active bug on the production path: the dispatcher's `handlerContext` getter always sets `cwd`/`resolveBase = this.resolveBase` (`dispatcher.ts:319-320`), so a handler invoked through the dispatcher always has `context.resolveBase`. The asymmetry only bites if an FS handler is invoked **outside** the dispatcher (a plugin, a direct import, or a future code path that forgets to thread context) — there read_file/write_file/edit_file/list_directory silently resolve against host `process.cwd()` while glob/grep stay anchored to the worktree.

## Proposed direction

Pick one invariant and enforce it:
- (a) add `createXHandler(cwd)` factories for the four bare tools so all six share the same fallback tier, **or**
- (b) assert/guarantee at the dispatcher boundary that `context.resolveBase` is always populated for every FS handler call, and add a test that a context-less invocation fails closed rather than silently using `process.cwd()`.

Prefer (b) — one invariant, smaller surface.

Found via a read-only tool/subagent/worktree audit (follow-up to #416).

griffinwork40:
## Summary

Closes #434 — the structural sibling of #416/#435, found in the same tool/subagent/worktree audit.

`read_file`, `write_file`, `edit_file`, and `list_directory` were registered as **bare** handlers with no cwd closure, unlike `glob`/`grep` which get `createGlobHandler(cwd)` / `createGrepHandler(cwd)`. The four bare handlers relied entirely on the dispatcher populating `context.resolveBase`; a context-less invocation fell through to `process.cwd()` in `resolveAndContain` and resolved **relative** paths against the host launch dir with **no containment**.

This is **latent today** — the dispatcher always threads `config.cwd`, so on the production path the fallback is unreachable — but it's a real asymmetry and a silent footgun for any non-dispatcher caller (a plugin, a direct import, a future code path).

### The fix — corrected option (a)

- **`create{Read,Write,Edit}FileHandler` / `createListDirectoryHandler` factories** that close over a session cwd, registered with the same `cwd !== undefined ? createX(cwd) : xHandler` tier as glob/grep. **Bare exports are preserved** (each = factory with no cwd), so existing callers/tests are untouched.
- **New optional `fallbackBase` param on `resolveAndContain` / `wouldBeRestricted`** — the *last* resolve-base tier, after `context.resolveBase` / `context.cwd`, before `process.cwd()`. Mirrors the `?? sessionCwd` tier grep/glob already carry. It is a **no-op on the dispatcher path** (the factory cwd and `context.resolveBase` are the same value, so `context` wins), so **zero behavior change on any live path** — it only bites a context-less non-dispatcher call, where it now anchors *and confines* relative paths to the session tree instead of the host launch dir.
- **An `// Invariant:` comment** in `_cwd-utils.ts` documenting that `resolveBase === undefined` marks an *unconfined top-level session*.

### Why not the issue's preferred option (b)?

The issue proposed option (b): "assert `resolveBase` is always populated at the dispatcher boundary." On inspection that's unsafe. `resolveBase === undefined` is **load-bearing**: it's the signal that *disables* containment for a top-level `afk chat`/`interactive` run (those launch with `config.cwd` unset → `dispatcher.resolveBase = undefined`, which is why they can read `~/.afk`, `/tmp`, absolute paths). Forcing `resolveBase` defined would confine every top-level session to its launch dir — and because a no-cwd session emits `readRoots`/`writeRoots` of `[]` (`[] ?? [base]` stays `[]`), it would reject **all** paths. Option (a) is the additive, backward-compatible fix that closes the asymmetry without touching that invariant. (The invariant is now documented so it isn't "fixed" into a regression later.)

## How was this tested?

Run on this tree (`30917d6`), branched off `origin/main`:

- `pnpm test` — **11233 passed / 14 skipped / 0 failed** (620 files)
- `pnpm audit:sdk:check` — OK (no unlocked SDK symbols)
- `pnpm audit:env:check` — 0 violations
- `pnpm scan:env:check` — env registry in sync

New unit coverage:
- `fs-handler-factory-cwd.test.ts` — each of the four factories anchors a relative path to its cwd with no context; a relative escape is confined to the cwd; and the **bare handler with no cwd + no context stays unconfined** (regression guard on the top-level invariant).
- `_cwd-utils.test.ts` — `fallbackBase` tier: anchors when context has no base, enforces `[fallbackBase]` containment, `context` base wins over `fallbackBase`, and `undefined fallbackBase` preserves the unconfined fall-through.

## Notes

- Independent of PR #514 (#435): zero file overlap. 514's correctness silently relies on the dispatcher-always-threads-`resolveBase` invariant; this PR makes that invariant explicit but is not a prerequisite.

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes

### DIFF
--- a/src/agent/tools/handlers/_cwd-utils.test.ts
+++ b/src/agent/tools/handlers/_cwd-utils.test.ts
@@ -241,3 +241,37 @@ describe('symlink containment', () => {
     expect(() => resolveAndContain(outsideFile, c)).toThrow(/outside the allowed/);
   });
 });
+
+describe('fallbackBase — factory-cwd resolve tier (issue #434)', () => {
+  // A context with NO resolveBase/cwd — the out-of-dispatcher invocation shape.
+  const baseless = {
+    cwd: undefined,
+    resolveBase: undefined,
+    readRoots: undefined,
+    writeRoots: undefined,
+  } as ToolHandlerContext;
+
+  it('anchors a relative path to fallbackBase when context carries no base', () => {
+    // Without fallbackBase this resolves against process.cwd(); with it, BASE.
+    expect(resolveAndContain('src/foo.ts', baseless, 'read', BASE)).toBe(INSIDE);
+  });
+
+  it('enforces containment against [fallbackBase] when context carries no base', () => {
+    expect(() => resolveAndContain(OUTSIDE, baseless, 'read', BASE)).toThrow(/outside the allowed/);
+    expect(wouldBeRestricted(OUTSIDE, baseless, 'read', BASE).restricted).toBe(true);
+    expect(wouldBeRestricted(INSIDE, baseless, 'read', BASE).restricted).toBe(false);
+  });
+
+  it('context base wins over fallbackBase (no-op on the dispatcher path)', () => {
+    // context.resolveBase = BASE; a bogus fallbackBase must be ignored.
+    expect(resolveAndContain('src/foo.ts', ctx(), 'read', '/tmp/other')).toBe(INSIDE);
+    expect(wouldBeRestricted(INSIDE, ctx(), 'read', '/tmp/other').restricted).toBe(false);
+  });
+
+  it('undefined fallbackBase preserves the unconfined fall-through (invariant guard)', () => {
+    // No context base AND no fallbackBase → resolveBase undefined → no enforcement.
+    // This is the load-bearing top-level-session invariant; do not "fix" it.
+    expect(resolveAndContain(OUTSIDE, baseless, 'read', undefined)).toBe(OUTSIDE);
+    expect(wouldBeRestricted(OUTSIDE, baseless, 'read', undefined).restricted).toBe(false);
+  });
+});
--- a/src/agent/tools/handlers/_cwd-utils.ts
+++ b/src/agent/tools/handlers/_cwd-utils.ts
@@ -70,16 +70,27 @@ export function _resetRootRealpathCacheForTests(): void {
  *                      back-compat callers that provide no context).
  * @param mode        - `'read'` or `'write'` — affects the error message noun
  *                      only; the containment logic is identical.
+ * @param fallbackBase - Optional session cwd closed over by a handler factory
+ *                      (e.g. `createReadFileHandler(cwd)`). Used as the LAST
+ *                      resolve-base tier — after `context.resolveBase` /
+ *                      `context.cwd`, before `process.cwd()` — so a factory
+ *                      handler invoked WITHOUT a dispatcher context still
+ *                      anchors relative paths to (and confines them within)
+ *                      its session tree instead of the host launch dir. Mirrors
+ *                      the `?? sessionCwd` tier grep/glob already carry. On the
+ *                      dispatcher path it is a no-op: `context.cwd` and the
+ *                      factory cwd are the same value, so `context` wins.
  * @returns The resolved absolute path.
- * @throws  When a `resolveBase` is set and the resolved path falls outside
+ * @throws  When a resolve base is set and the resolved path falls outside
  *          every allowed root.
  */
 export function resolveAndContain(
   inputPath: string,
   context: ToolHandlerContext | undefined,
   mode: 'read' | 'write' = 'read',
+  fallbackBase?: string,
 ): string {
-  const resolveBase = context?.resolveBase ?? context?.cwd;
+  const resolveBase = context?.resolveBase ?? context?.cwd ?? fallbackBase;
 
   // Resolve to absolute, anchoring relative paths against resolveBase.
   const abs = path.isAbsolute(inputPath)
@@ -93,7 +104,16 @@ export function resolveAndContain(
     return abs;
   }
 
-  // No base set — no containment enforcement.
+  // Invariant: `resolveBase === undefined` marks an UNCONFINED session — a
+  // top-level `afk chat`/`interactive` run with no worktree, where `config.cwd`
+  // is deliberately unset (dispatcher does `this.resolveBase = opts.cwd`). We
+  // disable containment here on purpose so such a session can read/write
+  // anywhere (config files, /tmp, absolute paths). Do NOT "fix" this by
+  // defaulting resolveBase to `process.cwd()`: that confines every top-level
+  // session to its launch dir, and — because the emitted `readRoots`/`writeRoots`
+  // are `[]` for a no-cwd session (`[] ?? [base]` stays `[]`) — would reject ALL
+  // paths. Confinement is opt-in via a set cwd (worktree/fork) or an explicit
+  // `fallbackBase`, never a default. See docs/issue #434.
   if (resolveBase === undefined) {
     return abs;
   }
@@ -146,8 +166,9 @@ export function wouldBeRestricted(
   inputPath: string,
   context: ToolHandlerContext | undefined,
   mode: 'read' | 'write' = 'read',
+  fallbackBase?: string,
 ): { restricted: boolean; resolved: string; roots: string[] } {
-  const resolveBase = context?.resolveBase ?? context?.cwd;
+  const resolveBase = context?.resolveBase ?? context?.cwd ?? fallbackBase;
 
   const abs = path.isAbsolute(inputPath)
     ? inputPath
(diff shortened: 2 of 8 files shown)


griffinwork40:
I now have everything needed to ground the correctness/spec/test-coverage dimensions inline. Confirmed at ref `30917d6a`:
- **`ToolHandler` type is 3-arg** (`input, signal, context?`) — the factory's returned `(input, signal, context) => impl(..., cwd)` closure matches exactly; `cwd` is captured via closure, not exposed as a 4th tool-handler param. Arity is sound.
- **`_cwd-utils.test.ts` helpers** (`BASE=/tmp/repo`, `INSIDE=/tmp/repo/src/foo.ts`, `OUTSIDE=/etc/passwd`, `ctx()`) make all four new `fallbackBase` assertions meaningful and correct.
- **Both test files present** at the ref.

No `critical`/`high` findings emerged from either dimension, so Wave 1.5 (citation/absence verification) has nothing to verify and `/shadow-verify` is not triggered. Synthesizing the merge decision.

---

# Code Review — PR #522

**`fix(tools): give read/write/edit/list the factory-cwd resolve tier glob/grep have (#434)`**
`afk/fix-434-fs-handler-cwd` → `main` · reviewed ref `30917d6a` · +251/−27, 8 files · regime: full (2 dimension agents)

## Summary
Additive, backward-compatible fix. Adds `create{Read,Write,Edit}FileHandler` / `createListDirectoryHandler` factories that close over a session `cwd`, registered in `createBuiltinHandlers` with the same `cwd !== undefined ? createX(cwd) : xHandler` tier glob/grep already use, plus an optional last-tier `fallbackBase` param on `resolveAndContain`/`wouldBeRestricted`. Closes a latent asymmetry (context-less FS-handler calls previously fell through to `process.cwd()` with no containment).

## Findings

### security — clean
`no issues found — read _cwd-utils.ts, dispatcher.ts, index.ts, providers at 30917d6a`
- The new tier is strictly additive: `resolveBase = context?.resolveBase ?? context?.cwd ?? fallbackBase` (`_cwd-utils.ts:104`, `:181`) — nullish-coalescing can never override a defined context base. Containment loop (root iteration, realpath compare, throw-on-reject) is byte-identical to the parent commit.
- No new exposure on the unconfined path: the dispatcher feeds the *same* `opts.cwd` into both `this.resolveBase` (`dispatcher.ts:237`) and the factory `fallbackBase` (`index.ts:70-73`); on every reachable production path `context` wins → provable no-op. Out-of-dispatcher it *strengthens* containment vs. the prior unchecked `process.cwd()` fall-through.

### api-compat — clean
`no issues found — read read/write/edit/list-file.ts, index.ts, types.ts at 30917d6a`
- `readFileHandler`/`writeFileHandler`/`editFileHandler`/`listDirectoryHandler` keep their exported names and `ToolHandler` type (`types.ts:102-106`, unchanged); each is now `= createXHandler()` — behaviorally identical (closed-over `cwd === undefined`).
- `create*Handler` exports are pure additions. The new `fallbackBase?: string` is optional and list-final; zero production callers pass a 4th positional arg at the PR head or its parent (only `path-approval-hook.ts` calls `wouldBeRestricted`, still 3-arg, file untouched). No positional collision.

### correctness — clean
- Tier precedence `?? context?.cwd ?? fallbackBase` is correct (context → factory cwd → undefined/unconfined) — `_cwd-utils.ts:104,181` · ref:30917d6a · diff-context.
- Bare re-exports (`export const readFileHandler = createReadFileHandler()`) invoke the impl with `cwd=undefined` → identical to the pre-PR direct const.
- `createBuiltinHandlers` wires all four factories symmetrically and registers them in the handler Map — `index.ts:67-73` · diff-context.
- Closure arity (3-arg) matches `ToolHandler`; verified against `types.ts:102-106` at ref.
- No handler/hook drift: on the dispatcher path `wouldBeRestricted(path, context)` (3-arg) and `resolveAndContain(path, context, mode, cwd)` (4-arg) derive the *same* resolveBase (context wins), so the hook↔handler containment agreement documented in `_cwd-utils.test.ts:4-9` holds.

### spec-compliance — intent MET
Stated intent = PR #522 title+body. All four requirements land: factory tier added, `fallbackBase` last-tier added, bare exports preserved, unconfined-top-level-session invariant documented (`_cwd-utils.ts:82-91`). Option (b) correctly **not** implemented, per the intent's own rationale. **No scope creep** — change is confined to the four handlers, shared util, tests, and registry. The central **no-op-on-live-path** claim is **verified** (dispatcher wiring trace, above).

### test-coverage — adequate, one low nit
- `low` · high confidence · test-coverage · `fs-handler-factory-cwd.test.ts:239-246` · ref:30917d6a · diff-context — the relative-escape confinement case is exercised only via `createReadFileHandler`; write/edit/list escape confinement isn't individually asserted.
  - evidence: `const escape = '../'.repeat(20) + 'etc/passwd'; const result = await createReadFileHandler(tmpDir)(...)`
  - suggestion: parametrize the escape assertion across all four factories, or add a one-line comment noting they share `resolveAndContain` (already covered at the helper level by `_cwd-utils.test.ts`'s `[fallbackBase]` containment test).
- Otherwise the two files cover all claimed cases: four factory anchor paths, `[fallbackBase]` containment, context-wins precedence, and the `undefined fallbackBase` unconfined-invariant regression guard.

### perf-observability — no issues
One closure indirection per handler build; not a hot path. Containment rejections already surface as error results — no logging gap introduced.

## Epistemic scope disclosure
- **Ref read:** Wave 1 agents and inline checks read at branch HEAD `30917d6a`. All citations verified at that ref; the PR is unmerged (main tip `ea3b9c0`).
- **Stated intent:** PR #522 title+body — spec-compliance **was** assessed.
- **Not checked:** (1) No `tsc`/build executed — arity/type-satisfaction verified by static read, not compilation (author reports `pnpm lint` clean). (2) Test suite not run (author reports 11233 passed / 14 skipped / 0 failed). (3) glob/grep internals not re-reviewed (unchanged, out of diff). (4) **The correctness·spec·test·perf dimension was reviewed inline, not in an isolated sub-agent** — the dispatched agent timed out at 20 min and was cancelled; findings were grounded against `30917d6a` reads and cross-checked with the security agent's independent wiring trace.

## Merge decision: **MERGE**
No blocking, critical, high, or medium findings across any dimension. One `low` test-coverage nit (non-blocking). The fix is additive, backward-compatible, a provable no-op on all live paths, and correctly preserves the load-bearing unconfined-session invariant.

---

**Done** — Reviewed PR #522 read-only; no repo/PR mutations.
- Verdict: **MERGE**; findings grounded at ref `30917d6a` (security + api-compat + correctness + spec + test + perf all clean bar one non-blocking `low` test-coverage nit at `fs-handler-factory-cwd.test.ts:239-246`).
- Evidence: this report; citations read via `git show 30917d6a:<file>` (working tree left untouched — still on `main`, dirty changes preserved).
- Deviation: the correctness·spec·test·perf sub-agent timed out (20 min, cancelled); that dimension was reviewed inline and disclosed above rather than silently dropped. Wave 1.5 and `/shadow-verify` were correctly skipped — no critical/high findings to verify.

---

PR #533, opened 2026-07-12
fix(redact): stop redacting long filesystem paths as opaque secrets

griffinwork40:
## Problem

Long absolute filesystem paths were being redacted to `[REDACTED]` in tool-lane labels, `events.jsonl`, and Telegram — e.g. a subagent running `cd <repo> && …` rendered as:

    $ bash cd [REDACTED] echo "===== HookDecision ====="

### Root cause

The generic long-token rule in `redactSecrets()` (`src/agent/redact-secrets.ts`) is the catch-all backstop for opaque ≥32-char secrets. Its character class includes `/`, `_`, and `-` — exactly the characters an absolute path is built from — so a bare path argument like `/Users/griffinlong/Projects/open_source/agent-afk` matched as a single "token" and was scrubbed. This is **over-redaction, not a leak** — the real command still ran with the true path; only the display/telemetry label was scrubbed.

## Fix

Add a path-shape guard to the generic rule (`looksLikeFilesystemPath`): a matched run is left intact only when

1. it contains a `/` separator, **and**
2. it carries none of base64's exclusive characters (`+`, `=`), **and**
3. no single `/`-delimited segment is itself a long opaque token — ≥32 chars, mixing letters and digits, with no dotted extension.

Rule 3 (added after review of PR #533) means an opaque token fused into a path segment (e.g. `cat /tmp/uploads/<44-char token>`) is **still redacted** — the guard fails safe toward redaction. Named high-value secrets (Anthropic keys, `Authorization: Bearer`, JWT, AWS IDs) are matched by the explicit rules that run **before** this one, so they are unaffected regardless of surrounding path context.

- base64url tokens (JWTs, most modern API keys) use `-`/`_` and have no `/` → redacted by the generic rule.
- Classic base64 carries `+`/`=` → redacted (rule 2).
- Opaque token as a path segment → redacted (rule 3).

**Accepted residual gap** (documented in-code): an unusually long (≥32-char) *dotless* path segment mixing letters and digits — a raw hash directory or a UUID-without-dashes — is over-redacted (treated as a token). Rare, cosmetic, and fails safe; the common case (short word segments) is preserved and the motivating `cd <repo>` case renders intact.

## Tests

- `background-summarizer.test.ts` (`redactSecrets` block): long absolute path preserved; base64 blob with `/`+`=` redacted; base64url token (no `/`) redacted; **opaque token fused into a path segment (base64url and hex) redacted**; long *digitless* descriptive path segment preserved.
- `tool-input-summary.test.ts`: the exact screenshot shape — `cd <long abs path> && echo …` — renders with the path intact and no `[REDACTED]`.

## Verification

- `pnpm test -- src/agent/background-summarizer.test.ts src/agent/providers/shared/tool-input-summary.test.ts` — 41 passed / 0 failed (background-summarizer.test.ts: 26 tests; tool-input-summary.test.ts: 15 tests)

### DIFF
--- a/src/agent/background-summarizer.test.ts
+++ b/src/agent/background-summarizer.test.ts
@@ -505,6 +505,32 @@ describe('redactSecrets', () => {
     expect(redactSecrets(text)).toBe(text);
   });
 
+  it('does not redact a long absolute filesystem path (path false-positive regression)', () => {
+    // A bare `cd <long absolute path>` argument is a run of ≥32 chars drawn
+    // entirely from the generic token class ([A-Za-z0-9+/=_-]); it used to be
+    // redacted to `cd [REDACTED]` in tool-lane labels. A path is not a secret.
+    const path = '/Users/griffinlong/Projects/open_source/agent-afk';
+    expect(path.length).toBeGreaterThanOrEqual(32);
+    expect(redactSecrets(`cd ${path} && echo hi`)).toBe(`cd ${path} && echo hi`);
+  });
+
+  it('still redacts a base64 blob with slashes when it carries +/= (not a path)', () => {
+    // Classic base64 with a `/` but also `+`/`=` must NOT be mistaken for a
+    // path — the path guard only spares `/`-runs free of base64 signal chars.
+    const b64 = 'a1b2c3d4e5f6g7h8+i9j0k1l2m3n4/o5p6q7r8s9t0=';
+    const out = redactSecrets(`X ${b64}`);
+    expect(out).not.toContain(b64);
+    expect(out).toContain('[REDACTED]');
+  });
+
+  it('still redacts a base64url token with no slash (path guard does not apply)', () => {
+    // base64url uses -/_ and no `/`, so the path guard never spares it.
+    const token = 'abcABC123_-abcABC123_-abcABC123_-abcABC12';
+    const out = redactSecrets(`X-Token: ${token}`);
+    expect(out).not.toContain(token);
+    expect(out).toContain('[REDACTED]');
+  });
+
   it('does not redact 16-char base32 strings without an AWS prefix', () => {
     // Looks like the back-half of an AKIA but without the prefix — should
     // not be flagged. (Floor is 32 chars for generic patterns.)
--- a/src/agent/background-summarizer.test.ts
+++ b/src/agent/background-summarizer.test.ts
@@ -537,4 +537,27 @@ describe('redactSecrets', () => {
     const text = 'random IOSFODNN7EXAMPLE9 token';
     expect(redactSecrets(text)).toBe(text);
   });
+
+  it('redacts an opaque base64url token fused into a path segment (PR #533 review med)', () => {
+    // A ≥32-char opaque token as a path segment is still a secret even though the
+    // whole run contains `/` and no +/=. The per-segment guard catches it.
+    const token = 'abcABC123_-abcABC123_-abcABC123_-abcABC12';
+    const out = redactSecrets(`cat /tmp/uploads/${token}`);
+    expect(out).not.toContain(token);
+    expect(out).toContain('[REDACTED]');
+  });
+
+  it('redacts a long hex token fused into a path segment', () => {
+    const token = '0123456789abcdef0123456789abcdef0123';
+    const out = redactSecrets(`cat /var/data/${token}`);
+    expect(out).not.toContain(token);
+    expect(out).toContain('[REDACTED]');
+  });
+
+  it('preserves a long DIGITLESS descriptive path segment (not an opaque token)', () => {
+    // A long path segment with no digits is a descriptive name, not a secret —
+    // it must survive. Guards against over-redacting real long directory names.
+    const p = '/opt/SomeVeryLongDirectoryNameWithoutAnyDigits/bin';
+    expect(redactSecrets(`ls ${p}`)).toBe(`ls ${p}`);
+  });
 });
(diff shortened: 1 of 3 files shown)


griffinwork40:
Both load-bearing claims are now **empirically confirmed by executing the exact redaction logic** at the reviewed ref (stronger than a shadow-verify re-derivation, so I'm substituting it for that gate):

- **The security finding is REAL** (cases 3 & 4): an opaque ≥32-char token — base64url *or* hex, no `+`/`=` — fused into a path run is now **preserved** where it was previously redacted.
- **The PR's named-secret safety claim is TRUE** (cases 6 & 7): `sk-ant`, JWT (and Bearer/AWS by pattern) are caught by explicit rules that run first — even inside a path — so high-value secrets are unaffected.

Citation check: the cited code (`looksLikeFilesystemPath` callback + body) exists verbatim at ref `1d71cef` (read via `git show`). Verified.

---

# Code Review — PR #533 `fix(redact): stop redacting long filesystem paths as opaque secrets`

**Target:** PR #533 · `fix/redact-secrets-path-false-positive` → `main` · head `1d71cef` · +68/-5, 3 files (1 source, 2 tests) · regime: light · stated intent: PR title+body (spec-compliance assessed).

## Findings (severity-sorted)

**medium · high confidence · security · `src/agent/redact-secrets.ts:63-64,82` · ref:1d71cef · file-state**
The path-shape guard preserves the *entire* matched run whenever it contains `/` and lacks `+`/`=`, so an opaque high-entropy token fused into a path segment is emitted intact into the redaction sinks (`events.jsonl`, tool-lane labels, Telegram, and the Haiku summarizer). This is **broader than the PR's documented "classic base64" gap**: base64url tokens (RFC 4648 §5 — no `+`/`=` *by construction*) and hex tokens fused in a path are *guaranteed* to be preserved, not coincidentally.
Evidence (empirically executed at the ref):
```
PRESERVED | cat /tmp/uploads/abcABC123_-abcABC123_-abcABC123_-abcABC12   (base64url token)
PRESERVED | cat /var/data/0123456789abcdef0123456789abcdef0123           (hex token)
```
Suggestion: expand the in-code "Accepted gap" comment (and PR body) to state that *any* opaque base64url/hex token fused into a `/`-bearing run with no `+`/`=` is now preserved — or gate the guard per `/`-delimited segment (reject when a segment is itself ≥20 high-entropy chars with no path-typical `.`/extension).
*Downgraded from the Wave-1 agent's `high`: backstop-only rule, named secrets verified unaffected, sinks are first-party/operator-owned, and this is the documented inverse of the intended fix.*

**low · high confidence · test-coverage · `src/agent/redact-secrets.ts:82` · ref:1d71cef**
No test pins the fused-token-in-path behavior. Verified absent: `git grep` for a `/tmp/…<32+char token>` shape across both touched test files at the ref returned zero matches. Suggestion: add a case pinning current behavior for `/tmp/uploads/<44-char base64url token>`.

**low · high confidence · test-coverage · `src/agent/redact-secrets.ts:82` · ref:1d71cef**
No test covers a path segment containing a literal `=` (legal in Unix filenames), which the guard treats as secret-shaped (redacts). Suggestion: add one boundary case.

## Verified clean
- **Ordering (load-bearing claim):** explicit named-secret rules (Bearer/`sk-ant`/JWT/AWS, lines 42-54) run before the generic rule (63-64); `[`/`]` are outside the token class so the generic rule cannot rematch an already-emitted `[REDACTED]`. Confirmed by read **and** execution (cases 6, 7). — read `redact-secrets.ts:41-64` at ref 1d71cef.
- **api-compat:** `export function redactSecrets(text: string): string` unchanged; `looksLikeFilesystemPath` is unexported with exactly one call site (`grep` at ref). No external surface changed.
- **correctness edge cases:** base64url/hex bare tokens still redacted (case 2), classic base64 with `+`/`=` still redacted (case 5), Windows `\` paths never form a single class-run — all inert by construction.
- **perf-observability:** callback + one `.test()` per ≥32-char match, on rare runs in per-tool-call paths — not a hot loop.

## Spec-compliance verdict
Intent **met**: path over-redaction is fixed (case 1), named-secret coverage is intact (cases 6-7), no scope creep (single source function + its two test suites). One caveat: the PR's stated residual gap **understates** the true new-preservation surface (base64url/hex-in-path, not just classic base64).

## What was not checked
- Read against ref `1d71cef` (present locally; citations verified there and by direct execution of the exact logic).
- Did not run the project's `pnpm test` (PR body asserts 11380 passing); did not review the *second, independent* `redactSecrets` in `src/browser/sanitize.ts` beyond confirming it has no analogous generic bare-length rule (no twin bug); did not audit non-bash tool-input summarization paths.
- Shadow-verify gate satisfied by direct empirical execution of the redaction logic (stronger than source re-derivation).

## Merge decision: **MERGE**
The fix cleanly achieves its stated goal and leaves all named high-value secrets fully covered. The one substantive finding is a documented heuristic-backstop trade-off affecting only opaque, un-named tokens routed to first-party sinks — non-blocking, but the security finding above warrants a fast-follow to correct the understated gap documentation (or tighten the per-segment guard).

---

**Done**
- Reviewed PR #533 read-only; produced the merge recommendation above with empirically-verified findings.
- Evidence: diff `/Users/griffinlong/.afk/state/skill-preflight/73b51241-.../pr-533.diff`; source read at ref `1d71cef` via `git show`; redaction logic executed inline (7 cases) confirming the one security finding and the named-secret ordering claim.
- No world state changed — no files edited, no git/PR mutations (working tree left dirty-on-`main` as found).
- Recommendation: **MERGE** with one non-blocking medium (expand/​tighten the understated redaction gap) and two low test-coverage gaps for a fast-follow.

---

PR #539, opened 2026-07-12
fix(cli): eliminate end-of-turn report/table void on overlay collapse

griffinwork40:
## The bug ("tables render fucky")

A streamed report ending in a table, rendered beneath a tall overlay (the diagnose/subagent lane), collapses at end-of-turn into a multi-row **blank void** between content stranded in scrollback and the band hugging the input — and the earliest rows (the report header) are **lost entirely**.

## Root cause

Many small blocks committed under a tall overlay each fit the room above the **current tall** frame, so they take the eager fits-path and scroll into native scrollback. Scrollback is append-only via a full-screen scroll (`docs/scrollback.md:20-67`) — those rows freeze there. On collapse the surviving band re-pins down to the (deliberately) bottom-pinned frame (`frame.ts:230`, `bottom-pin.test.ts:63` — dropdown headroom) while the frozen rows strand above it → the void + lost header.

This is the collapsed-frame boundary gap that `overflow-gap.test.ts:198-205` and `docs/scrollback.md:400-405` explicitly left out of scope, which is why the suite stayed green while reality broke.

## The fix (~15 LOC, `commit-mode.ts`)

Route any run whose accumulated height exceeds the **current tall-frame room** through band-hold (retained model, capped at `maxBandModel` = the rows that fit above the **collapsed** frame) instead of eager-scrolling. Content that will fit post-collapse is retained until collapse and painted contiguously; only genuine overflow beyond `maxBandModel` is archived, from the top, contiguously. Respects both hard constraints: scrollback stays append-only, the frame stays bottom-pinned.

## Validation

- **Full suite green: 11386 passed / 14 skipped / 0 failed.** `pnpm lint` clean.
- New `terminal-compositor.collapse-void.test.ts` — deterministic 100×40 repro (header present, no ≥2-row void, run hugs the frame). Red before, green after.
- The `band-hold-perline-gap` discriminator **flipped**: it asserted the per-line commit loop was "genuinely gappy" (to justify `commitBlockAbove`); the retained-model fix addresses that root cause too, so per-line is now contiguous (`span 13/13`, zero drops). Updated to assert the new invariant.
- **Real-terminal A/B (iTerm2), the only ground truth per `docs/scrollback.md:108-111`:** before = void + missing header; after = full report contiguous, header present. Repro: `pnpm exec tsx scripts/visual-void-repro.ts` on `main` vs this branch.

## Scope / future work

This ships the **minimal A2 core** only. `docs/proposals/tui-compositor-rewrite.md` documents the optional fuller unification (stateless render-instead-of-repin + single end-of-turn flush) as separate, no-rush hardening — not required for this fix.

## Test plan
- [x] `pnpm test` (full suite)
- [x] `pnpm lint`
- [x] real-terminal A/B in iTerm2

### DIFF
--- a/docs/proposals/tui-compositor-rewrite.md
+++ b/docs/proposals/tui-compositor-rewrite.md
@@ -1,0 +1,146 @@
+# Proposal: unify the compositor commit model to dissolve the scrollback-gap class
+
+**Status:** Stage 0 VALIDATED — real-terminal A/B (iTerm2) confirmed the fix (void gone, header
+present, report contiguous). This PR ships the **minimal A2 core** (~15 LOC in `commit-mode.ts`: retain
+against `maxBandModel`, not the tall-frame room). Stages 2–3 (stateless render-instead-of-repin, single
+end-of-turn flush) remain OPTIONAL future hardening — not required for this fix.
+**Author:** session 8831d316 (diagnosis + design). All claims cited to code read this session.
+**Supersedes intent of:** the "separate, larger item" flagged in `docs/scrollback.md:404`.
+
+---
+
+## 1. Context — the bug and why it recurs
+
+Symptom (user-reported, reproduced deterministically): a streamed report ending in a table renders with a
+multi-row **void** between content stranded in scrollback and the band hugging the frame; earliest rows can be
+lost. Probe buffer (100×40, 22-row overlay): `PROSE-01..05` in scrollback, 25 blank rows, then `PROSE-06`+table
+hugging the frame; `HEADER-MARKER` absent.
+
+This is one instance of a **failure class**. ~12 regression tests circle it: `shrink-gap`, `scrollback-gap`,
+`overflow-gap`, `multi-commit-gap`, `commit-collapse-wrap-gap`, `band-hold-perline-gap`, `endturn-overflow-gap`,
+`banner-commit-gap`, `wrap-gap`, `pad-decay-straddle`, `two-table-overflow`. That density is the signal: the model
+is fighting the medium.
+
+## 2. The two hard constraints any design must respect
+
+**(C1) Scrollback is append-only via full-screen scroll — you cannot reposition a committed line.**
+`docs/scrollback.md:20-67`: pushing a line to scrollback requires a *full-screen* DECSTBM scroll (cursor at the
+bottom margin + `\n`). Sub-region scrolls discard displaced lines. Once a line is in native scrollback its row is
+frozen forever.
+
+**(C2) The frame is unconditionally bottom-pinned, on purpose.**
+`frame.ts:230-253` (`targetBottomRow === absoluteBottom`, "unconditional bottom-pinning") and
+`bottom-pin.test.ts:21,63` ("frame stays bottom-pinned … so the dropdown has headroom"). The frame will **not**
+float up to meet sparse content — the autocomplete dropdown / picker rely on the input sitting at the bottom with
+headroom above it.
+
+**(V) The automated suite cannot verify real-terminal behavior.** `docs/scrollback.md:9-13,108-111`: mock/headless
+tests confirm bytes were *written* but not that they *reached scrollback* — that is a property of the real PTY's
+scroll engine. **Prior fixes (6d7cb90, 9690b9f) shipped green and were broken in reality.** Ground truth = a human
+running afk in iTerm2 / Apple Terminal / xterm and scrolling up. This is the dominant risk for this whole effort
+and is a hard dependency on a human (see §8).
+
+## 3. Root cause (precise)
+
+The compositor has **three** interacting mechanisms for getting committed content on screen:
+
+1. **Eager fits-path commit** (original): a block that fits the room above the *current* frame is pushed to
+   scrollback immediately via a full-screen scroll, and painted above the frame.
+2. **Band-hold** (`docs:426-448`): a block that overflows the *current tall* frame but fits the *collapsed* screen
+   is retained in an in-memory `committedBand` model (capped at `maxBandModel` = rows above a minimal frame),
+   painted partially, archived to scrollback only past the cap.
+3. **Re-pin** (`committed-band-repin.ts`): on frame geometry change, the painted band is *repositioned* to hug the
+   new frame top.
+
+The void arises because the **eager fits-path (1) sizes its keep-vs-archive decision against the room above the
+TALL overlay (~12 rows), not the room above the COLLAPSED frame (`maxBandModel` ~34).** Content that *would* fit
+post-collapse gets prematurely archived to scrollback (C1: now frozen). On collapse, re-pin (3) drops the surviving
+band down to the bottom-pinned frame (C2), leaving the prematurely-archived rows stranded above → the void.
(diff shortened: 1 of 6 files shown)


griffinwork40:
Verification complete — and the absence-check refined the finding: `anchorRow > 1` *is* covered broadly (repaint/bottom-pin/splice tests), so I won't overclaim. The precise gap is that neither the PR's new nor flipped test combines `anchorRow > 1` with the new many-small-commits band-hold routing. I have everything needed to synthesize.

---

# Code Review — PR #539

**Target:** `fix(cli): eliminate end-of-turn report/table void on overlay collapse` · `spike/compositor-retained-model` → `main`
**Reviewed ref:** `1180830e` (verified: diff applied; consumer `committed-band-commit.ts` byte-identical main↔ref)
**Triage:** hotfix · 6 files · +395/−22 · **production surface = 16 lines in one function** (`decideCommitMode`, `src/cli/commit-mode.ts`); remainder is a design proposal doc, a docs update, a manual real-terminal repro script, and two tests.
**Regime:** light-equivalent (single production hunk), reviewed at full depth given content-loss stakes.

## The change
Adds a third entry condition to `useBandHold`: `runExceedsCurrentRoom = overflowRun.length > room`, so a run whose *accumulated* height (prior band + new lines) exceeds the current tall-frame `room` is retained in the band model instead of eager-scrolled into append-only scrollback. `runExceedsCurrentRoom` is a local const — **not** added to the returned `CommitMode` interface.

## Findings by severity

### MEDIUM · medium confidence · correctness — anchor-floor over-decrement (PRE-EXISTING, widened by this PR)
`src/cli/terminal-compositor.committed-band-commit.ts:425` sets `scrolledRows = fitsAboveFrame ? bandOverflow : 0` **before** and **independently of** the `useBandHold` branch. When `fitsAboveFrame && useBandHold` are both true (now newly reachable via the many-small-commits path), Phase 1 takes the band-hold branch and actually scrolls `genuineOverflow = overflowRun.length − maxBandModel` (line 435 — typically **0** while the run still fits `maxBandModel`), not `bandOverflow`. The post-commit decrement `self.anchorRow = max(1, anchorRow − scrolledRows)` (475-477) then lowers the protected-banner ceiling by rows that never scrolled.
- **Reachability:** gated on `anchorRow > 1` (a banner session, e.g. `preArmAnchorRow`); transient until the banner scrolls off.
- **Regression?** No. `fitsAboveFrame && useBandHold` was already reachable pre-PR via the `overflowHasPending` arm (which has no `fitsAboveFrame` term), so the `scrolledRows`/`genuineOverflow` divergence is latent in the *unchanged* consumer. This PR does not introduce it; it exercises it more often, in the PR's own target scenario (small blocks + banner).
- **Suggestion (follow-up, not a merge blocker):** inside the `if (useBandHold)` branch set `scrolledRows = genuineOverflow` rather than inheriting the fits-path `bandOverflow`.

### LOW · high confidence · test-coverage — new regression test doesn't exercise the banner path
`src/cli/terminal-compositor.collapse-void.test.ts:57` and the flipped `terminal-compositor.band-hold-perline-gap.repro.test.ts:78` both construct the compositor with `anchorRow: 1`, so the anchor-floor decrement from the MEDIUM finding is not covered by *this PR's* tests. (`anchorRow > 1` is covered generally in repaint/bottom-pin/splice suites, but never combined with the new `runExceedsCurrentRoom` band-hold routing.) **Suggestion:** add a banner (`anchorRow > 1`) variant of the collapse-void test.

### NIT · high confidence · test-coverage — discriminator differential proof lost
The flipped `band-hold-perline-gap` case no longer proves the per-line loop diverges from `commitBlockAbove` at loopH=16. Legitimate (the root-cause fix made per-line contiguous too), and the parameterized batch-loop regression guard (`:117-130`) is unchanged — a future `commitBlockAbove` regression is still caught there. Only the narrow per-line-vs-batched differential is gone.

## Clean dimensions (read at reviewed ref)
- **api-compat — clean.** `CommitMode`/`CommitModeInput` interfaces and `decideCommitMode`/`capBandModel` signatures unchanged; no field added to the returned object; zero consumers outside `src/cli/` (grep-verified); the sole production consumer still destructures `useBandHold`/`overflowRun`/`maxBandModel`, all still returned.
- **security — clean.** Pure routing boolean; writes no new terminal content (no new escape-sequence/injection surface); retained band bounded by `capBandModel(run, maxBandModel)` with `maxBandModel ≤ terminal rows` — no unbounded memory growth on long turns.
- **correctness (happy path, `anchorRow=1`) — clean.** `overflowRun` partitions into disjoint archived (`genuineOverflow`, 435-445) / painted-now (`paintedCount`, 544-563) / pending buckets — no duplication or loss; the new clause's marginal effect is exactly the `overflowPriorContiguous` many-small-commits case (subsumed otherwise).
- **spec-compliance — clean.** Implements the stated intent verbatim; no production scope creep (docs/proposal/repro-script are additive).
- **perf — clean.** O(1) added check; bounded retention; reuses the existing chunked archive path.

## What was not checked / epistemic scope
- Wave 1 correctness lane read the consumer + tests from disk (verified byte-identical to ref `1180830e`) and the supplied `commit-mode.ts` change; **citations for the MEDIUM finding independently re-verified by me at the reviewed ref** (lines 425/435/475-477 exist verbatim).
- **Delegation failure:** the security·api-compat sub-agent timed out (20 min); I completed both dimensions inline (grep + direct reads) — findings above reflect that fallback, not the sub-agent.
- **Real-PTY behavior not certifiable by CI** (`docs/scrollback.md:108-111`): the headless suite confirms bytes were *written*, not that they *reached scrollback*. The fix's real-terminal soundness rests on the author's stated iTerm2 A/B, not automated tests — an honest, disclosed limitation (CI harness tracked in #541), not a defect of this PR.
- Stated intent: PR #539 title+body — **spec-compliance assessed.**
- No critical/high finding present → `/shadow-verify` gate not triggered; the sole MEDIUM was already double-derived (my pre-dispatch analysis + the Wave 1 agent) and citation-verified.

## Merge decision: **MERGE**
A minimal, spec-compliant, api/security-clean fix for a real, reproduced, user-facing bug, row-accounting-correct in the common `anchorRow=1` path and human-validated in a real terminal. The one MEDIUM is a pre-existing latent inconsistency in the unchanged consumer (not a regression this PR introduces); it plus the banner-path test gap warrant a fast follow-up — set `scrolledRows = genuineOverflow` in the band-hold branch and add an `anchorRow > 1` collapse-void test — but neither blocks this merge.

**Done**
- Reviewed PR #539 across all six dimensions at reviewed ref `1180830e`; verdict **MERGE** with one MEDIUM (pre-existing, non-blocking), one LOW test gap, one NIT.
- Evidence: findings above cite `src/cli/commit-mode.ts` and `src/cli/terminal-compositor.committed-band-commit.ts:425/435/475-477` (all re-verified against the reviewed ref); PR intent from `gh pr view 539`.
- World unchanged: this was read-only — no files edited, no git/PR mutation, working tree (`.gitignore` + 4 untracked docs) left untouched.
- Deferred: recommended follow-up fix + banner-path test are for the author, alongside the tracked hardening in #540/#541; the security·api-compat sub-agent timed out and was replaced by inline analysis.

---

Issue #441, opened 2026-07-05
tracking: minor tool/subagent/worktree robustness follow-ups


griffinwork40:
## Summary

Tracking issue for a cluster of low-severity tool/subagent/worktree robustness follow-ups surfaced by a read-only audit. Each is small and independently landable.

## Checklist

- [ ] **grep lacks dead-cwd ENOENT enrichment.** `bash` translates a swept-worktree spawn failure into a clear "working directory does not exist (deleted worktree?)" message via `describeSpawnCwdError`; `grep` does not, so it surfaces a raw `spawn grep ENOENT`. Evidence: `grep.ts:156` (spawn with cwd, no enrichment) vs `bash.ts:331-350`. Cosmetic/diagnostic parity.

- [ ] **edit_file read→write TOCTOU.** `edit_file` reads at `edit-file.ts:117`, writes at `:148`, with no mtime/hash guard — a concurrent external edit between the two is silently clobbered. Bounded (single-process agent), but no staleness detection and no read-receipt requirement.

- [ ] **`worktreeMainRootCache` not invalidated on `setCwd`.** `subagent.ts:284` caches the resolved main root per-cwd for the manager lifetime; `setCwd` (`:369-371`) mutates `parentCwd` but does not clear the cache, so a delete+recreate at the same path mid-session returns a stale `mainRoot`. Rare.

- [ ] **Silent re-confinement on `git rev-parse` failure.** `resolveWorktreeMainRoot` returns `undefined` on any git failure (`worktree-read-root.ts:107-110`), so the child is silently re-confined to `[worktree]` — the #416 symptom returns with no log. Consider a debug log so the degradation is observable.

- [ ] **No global/adaptive rate governor; compose `failFast` loses sibling progress.** Each fan-out site caps at 8 independently (`concurrency-pool.ts:33`, `dispatcher.ts:137`), so N concurrent `compose` calls → N×8 simultaneous, which can trip a low 429 ceiling. `compose` defaults `failFast=true` (`dag.ts:37`), so one worker's 429 aborts the whole DAG layer and loses in-flight sibling progress. Consider a shared governor and/or a `fail_fast:false` default for rate-sensitive fan-outs.

Severity: LOW each. Label: enhancement.

Found via a read-only tool/subagent/worktree audit.

griffinwork40:
## Triage 2026-07-27 — re-verified against `main` @ e8861a5 (v5.81.3)

[...]

PR #554, opened 2026-07-13
fix(subagent): grant ~/.afk/state to confined forks; harden worktree main-root fallback

griffinwork40:
## Summary

Completes the read-scope fix started in #544. That PR fixed forks of **unconfined** top-level sessions (`afk`/`afk i` with no worktree → read-open child), but forks of **confined** parents (worktree `-w`, skill fan-outs) still hit `PreToolUse` auto-denials. A fork cannot prompt, so any read outside its inherited `readRoots` is hard-blocked — and the child spins on retried denials until the 20-min `SUBAGENT_DEFAULT_TIMEOUT_MS`.

This is the root cause behind the `tool-failure-compose` signal (**24% of compose calls — 130/545 — completed with ≥1 failed node**, durations up to ~35 min) and live **post-#544** denials observed in running sessions: `~/.afk/state/skill-preflight/*.diff`, the main repo, and sibling worktrees.

## Two gaps closed

**Gap A — confined forks could not read `~/.afk/state`.** The confined-parent read union `{childCwd, parentRoots, worktreeMainRoot}` never included the AFK state dir, so skill-preflight staged inputs (e.g. a PR diff a `/review` fan-out hands its children), todos, transcripts, and per-session ledgers were denied. Now the **state** dir is folded into the union — **never `~/.afk/config`** (that holds `afk.env` credentials; read scope must not widen to secrets).

**Gap B — confined *worktree* parents lost the main checkout + siblings** when `resolveWorktreeMainRoot(childCwd)` returned `undefined` (git failure, or `childCwd` is not itself a linked worktree). We now fall back to resolving from the **parent** cwd — for an `afk -w` parent that is a worktree whose main root lexically covers the main checkout and every sibling `.afk-worktrees/*`.

## Safety

- Both changes are **read-only, additive** grants. Writes stay confined (separate `writeRoots` axis) — asserted by an existing test.
- **Unconfined** forks are unchanged (already read-open; the state grant is gated to confined parents only).
- Callers that **pin `readRoots`** (`afk farm`) are untouched (inheritance is skipped when `readRoots` is set).
- `~/.afk/config` is never granted — a unit test asserts no granted root admits a config path.

## Testing

- `computeInheritedReadRoots` unit tests: state dir folded into the confined union; ignored for unconfined; back-compat when omitted; guard that `~/.afk/config` is never admitted.
- `forkSubagent` manager tests: state dir present on confined forks; **Gap B** parent-cwd fallback recovers the main root when the child cwd does not resolve.
- Full suite green: **634 files, 11,539 passed, 0 failed**; `tsc --noEmit` clean; `audit:sdk:check` / `audit:env:check` / `scan:env:check` all pass.

## Files

- `src/agent/subagent-read-scope.ts` — `afkStateRoot` arg folded into the confined union
- `src/agent/subagent.ts` — pass `getAfkStateDir()` (confined only); parent-cwd worktree-main-root fallback
- `src/agent/subagent-read-scope.test.ts`, `src/agent/subagent-worktree-readroot.test.ts` — coverage

---
Diagnosis + fix authored via an AFK session; every load-bearing claim was independently re-derived by a `/shadow-verify` adversarial wave before implementation.

---

## Update — Gap B hardening (commit `a787619`)

Gap B's original parent-cwd fallback does **not** cover the most common real failure: a fork whose cwd **is** the parent's worktree (`childCwd == parentCwd` — e.g. a `/gather` fan-out with no explicit child cwd) when `git rev-parse` throws on that worktree's own admin dir (pruned/mid-sweep). The fallback is guarded by `parentCwd !== effectiveChildCwd`, so it never fires there — and re-resolving the same path via git would fail identically.

This was the **live failure** in witness session `e9f7d80d`: two `/gather` Explore forks auto-denied reading the main checkout + `src/improve` on v5.37.0 (which already shipped #544).

**Fix:** `resolveWorktreeMainRoot` now recovers the main root **lexically** from the path when git fails — afk worktrees live at `<mainRoot>/.afk-worktrees/<slug>`, so the main root is derivable without git. This is the only remedy for a broken worktree admin dir, since no git invocation from any cwd can recover it.

- Strictly additive (only ever **adds** a read root); writes stay confined (separate axis); the `afk farm` guardrail is unaffected (pinned `readRoots` short-circuit resolution before this runs).
- New `resolveWorktreeMainRoot` unit coverage: git-fails + `.afk-worktrees/` path → main root recovered (incl. a subdir inside the worktree); non-afk path → still `undefined`, degradation logged under `AFK_DEBUG` (#441); recovery also logged.
- `tsc --noEmit` clean; full suite green (**11,542 passed / 0 failed**).

### DIFF
--- a/src/agent/subagent-read-scope.test.ts
+++ b/src/agent/subagent-read-scope.test.ts
@@ -124,4 +124,75 @@ describe('computeInheritedReadRoots', () => {
       expect(roots).toBeUndefined();
     });
   });
+
+  describe('afkStateRoot grant (confined forks reach ~/.afk/state — Gap A)', () => {
+    const STATE = '/Users/me/.afk/state';
+
+    it('folds the AFK state dir into a confined worktree fork union', () => {
+      const roots = computeInheritedReadRoots({
+        parentReadRoots: undefined,
+        parentCwd: '/repo/.afk-worktrees/wt',
+        childCwd: '/repo/.afk-worktrees/wt',
+        worktreeMainRoot: '/repo',
+        afkStateRoot: STATE,
+      })!;
+      expect(new Set(roots)).toEqual(
+        new Set(['/repo/.afk-worktrees/wt', '/repo', STATE]),
+      );
+      // A skill-preflight staged input under the state dir is now admitted
+      // (containment is lexical: path.relative does not escape the root).
+      expect(
+        path.relative(STATE, `${STATE}/skill-preflight/pr-548.diff`).startsWith('..'),
+      ).toBe(false);
+    });
+
+    it('grants [cwd, state] even with no worktree main root (state lifts the fork out of the [cwd] default)', () => {
+      const roots = computeInheritedReadRoots({
+        parentReadRoots: undefined,
+        parentCwd: '/plain/repo',
+        childCwd: '/plain/repo',
+        worktreeMainRoot: undefined,
+        afkStateRoot: STATE,
+      });
+      // Without afkStateRoot this is the "only root is cwd" case → undefined.
+      expect(new Set(roots)).toEqual(new Set(['/plain/repo', STATE]));
+    });
+
+    it('is ignored for an unconfined (read-open) parent — read-open already covers state', () => {
+      const roots = computeInheritedReadRoots({
+        parentReadRoots: undefined,
+        parentCwd: undefined,
+        childCwd: '/repo/.afk-worktrees/wt',
+        afkStateRoot: STATE,
+      });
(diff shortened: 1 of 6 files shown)


griffinwork40:
Pushed `a787619` — completes Gap B for the `childCwd == parentCwd` git-failure case (the actual read-denial seen in session `e9f7d80d`, which Gap B's `parentCwd !== effectiveChildCwd` guard skipped). `resolveWorktreeMainRoot` now recovers the main root lexically from the `.afk-worktrees/` path when git throws. Additive-only; farm guardrail + write confinement untouched; `tsc` clean; full suite green. Details in the updated description.

PR #558, opened 2026-07-13
security(bash): best-effort readRoots/writeRoots path scan + ADR (closes #354)

### LINKED ISSUE #354: security: bash tool has no readRoots/writeRoots containment (tracked C4)
## Summary

The `bash` tool handler runs commands via `child_process.spawn(cmd, { shell: true })` and sets only the spawned shell's `cwd`. It does **not** consult `readRoots`/`writeRoots` and does **not** route paths through `resolveAndContain`. A model running inside an isolated worktree (e.g. an `afk farm` branch) can still read from or write to any path the host process can reach.

This is acknowledged in the source — `src/agent/tools/handlers/bash.ts` documents the gap in the `createBashHandler` doc comment and emits a one-time `[security]` warning at runtime when `permissionMode === 'bypassPermissions'`, tagged `(tracked: C4)`. This issue formalizes that tracked gap.

Provenance: introduced as a soft mitigation alongside a broader security/correctness hardening pass — see `CHANGELOG.md` entry **C4** under the Security section (`bash` handler emits a one-time `[security]` warning; full `execFile` migration deferred).

## Threat model

A model with bash access in a farm worktree can:

- Read outside the worktree: read arbitrary system files or credential directories, `find / -name '*.env'`
- Exfiltrate from the source repo: copy a `.env` file to a scratch directory
- Write outside `writeRoots`: append to a user's shell profile
- Bypass any path-based containment via shell expansion: `$(printf /etc/hosts)`, env-var indirection, command substitution

The other file-system handlers (read_file, write_file, edit_file, glob, grep, list_directory) all centralize through `_cwd-utils.ts`'s `resolveAndContain` and are correctly bounded. `bash` is the only un-contained tool, and shell metacharacters mean it's the widest attack surface in the toolset.

## Why this is hard to fix properly

`bash` takes an opaque command string. Unlike the other handlers, there is no general way to extract "the paths this will touch" without executing the command. Three plausible shapes, all with real cost:

| Option | Cost | Trade-off |
|---|---|---|
| **A. JS-side shell parsing** — drop `shell: true`, reimplement pipes / redirects / heredocs / `$()` / backticks in our own parser, then validate each segment's paths | High; rathole | We're now maintaining a shell syntax parser. Bugs in our parser = silent escapes. |
| **B. OS sandbox wrap** — run bash inside `sandbox-exec` (macOS) / bubblewrap (Linux) profiles | Medium; platform-specific | Per-OS profiles, easy to misconfigure, doesn't cover Windows. |
| **C. Argv-only `execFile`** — require structured `{ command, args[] }` input, no shell parsing | Low engineering; high UX cost | Breaks legitimate use like piping one command's output into another. Agents will work around it badly. |

## Proposed staging

**Tier 1 (cheap, do soon if farm goes wide):** best-effort soft hardening — does not close the gap, but raises the floor.

- Best-effort path scan on the command string: extract absolute paths and home-relative paths, run each through `resolveAndContain` against the session's `readRoots`/`writeRoots`. Won't catch `$(printf /etc/hosts)` or env-var expansion, but catches the 80% case (reading a sensitive file directly).
- **Fail-closed gating**: when `writeRoots` is set AND `permissionMode === 'bypassPermissions'`, refuse to run bash entirely. Force the agent to use the file tools (which are bounded).
- Estimated cost: half a day. No new dependencies, no platform forks.

**Tier 2 (defer):** the real refactor (option A / B / C above). Only justified if the threat model changes:

- Farm starts ingesting untrusted-source task descriptions, or
- Multiple users share infrastructure (today everything is single-user-on-laptop), or
- A demonstrated escape lands in an issue

**Tier 3 (deferred indefinitely):** consider whether `execFile`-shaped APIs (Tier 2 option C) should be exposed as a *separate* tool (`bash_safe`?) leaving the unsafe `bash` available behind a permission gate, rather than replacing it. Lets us tighten by default without breaking pipe-heavy workflows.

## Acceptance criteria for closing this issue

- The bash handler either enforces `readRoots`/`writeRoots` for absolute path arguments it can extract from the command string, **or**
- A documented architectural decision lives in `docs/` explaining why the gap is accepted at the chosen threat-model level (single-user laptop, trusted task descriptions, etc.)

Until one of those lands, the runtime warning in `src/agent/tools/handlers/bash.ts` (`warnIfBypassPermissions`) is the only mitigation.

## References

- `src/agent/tools/handlers/bash.ts` (documented gap + runtime warning, `createBashHandler`/`warnIfBypassPermissions`)
- `src/agent/tools/handlers/_cwd-utils.ts` — `resolveAndContain` (the containment primitive used by the other file-system handlers)
- `CHANGELOG.md` — **C4** entry (introduction of the soft mitigation as part of a security/correctness hardening pass)

_Migrated from the internal tracker ahead of its archival._

griffinwork40:
Closes #354.

## What

Adds **best-effort, advisory-only** `readRoots`/`writeRoots` path containment to the `bash` tool handler, plus an ADR documenting the accepted residual gap. Satisfies **both** acceptance criteria in #354 (enforce/scan extractable paths *and* a documented architectural decision).

## Why this shape (and why NOT fail-closed)

The `bash` handler spawns via `shell: true` and, unlike the six typed filesystem handlers (which all route through `resolveAndContain`), never consulted the session roots — the widest containment gap in the toolset (tracked **C4**).

The issue floats a fail-closed idea: *refuse bash when `writeRoots` is set AND `bypassPermissions`*. **Rejected** — verified that a top-level `afk -w` session runs under `bypassPermissions` with a non-empty effective `writeRoots=[worktree]`, so that refusal would break the primary human-driven worktree workflow. (Sub-agent forks run `permissionMode='default'` and never inherit bypass, so they wouldn't be refused anyway — the refusal would *only* ever hit the legitimate top-level session.) So the scan is **warn-only: it never blocks execution.**

## How

- New `extractCandidatePaths(command)` in `_cwd-utils.ts`: best-effort tokenizer pulling absolute (`/…`) and home-relative (`~`, `~/…`) tokens, quote-stripped. Explicitly **not** a shell parser — does not see `$()`, backticks, env-var indirection, or globs (that limitation is the whole reason the ADR exists; the issue calls a real parser a rathole).
- `bash.ts`: `scanPathsBestEffort()` runs before spawn, expands `~`, calls the existing non-throwing `wouldBeRestricted(path, context, 'write', …)` per candidate. On an out-of-root reference it emits a one-time `[security]` `console.warn` naming the escaping resolved paths + a `tool.bash_path_escape` telemetry row (counts + mode only — **never** the raw command string, matching the existing overflow-telemetry privacy rule), then spawns normally. `wouldBeRestricted` already short-circuits to not-restricted under `allowAll`, so bypass sessions produce no noise.
- `docs/decisions/0001-bash-tool-path-containment.md`: ADR (Accepted) — context, the warn-only decision, the fail-closed rejection rationale, the accepted threat model (single-user-on-laptop, trusted task descriptions), covered vs. uncovered cases, references.
- `CHANGELOG.md`: C4 entry under Unreleased.

## Verification

- `tsc --noEmit` (repo `lint`): **clean (exit 0)**
- `vitest run`: **96/96 pass** — `bash.test.ts` (60), `_cwd-utils.test.ts` (31), `routing-telemetry.test.ts` (5)
- New tests cover: extractor tokenization (absolute/home/quoted/dedup, ignores relative+flags); out-of-root path → warns **and still executes** (not an error); `allowAll` → no warning; in-root → no warning; one-time warn latch; no-context path.

## Notes

Full `execFile`/sandbox containment remains deferred (still tracked C4) — out of scope by design per the issue's staging. Residual best-effort limitations are documented in the ADR and pinned by tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ auto-release workflow to deduplicate commits across successive runs.
 
 ## [Unreleased]
 
+### Added
+- **C4** `bash` handler now runs a best-effort, advisory-only `readRoots`/`writeRoots` path-containment scan before spawning: absolute and home-relative path tokens are extracted from the command and checked against the session's write roots, emitting a one-time `[security]` warning + `tool.bash_path_escape` telemetry (counts only, no command string) when one escapes — but never blocking execution (warn-only, to preserve the top-level `afk -w` bypass workflow). Rationale, threat model, and residual gaps documented in `docs/decisions/0001-bash-tool-path-containment.md`. Full `execFile`/sandbox containment remains deferred (tracked C4). (#354)
+
 ## [5.37.2] - 2026-07-13
 
 ### Fixed
--- a/docs/decisions/0001-bash-tool-path-containment.md
+++ b/docs/decisions/0001-bash-tool-path-containment.md
@@ -93,6 +93,9 @@ work.
   before the containment check.
 - Quoted path tokens and paths abutting common shell punctuation
   (`;`, `,`, `)`).
+- Paths glued to a leading redirection/pipe operator with no space
+  (`>/etc/passwd`, `>>~/.bashrc`, `2>/tmp/err`, `|/tmp/x`) — the leading
+  operator run (and any fd prefix) is stripped before the containment check.
 
 **Not covered** (documented limitations — the reason this is advisory, not a
 boundary):
@@ -100,8 +103,8 @@ boundary):
 - Command / arithmetic substitution: `$(printf /etc/hosts)`, backticks.
 - Environment-variable indirection: `$HOME/.ssh`, `${SECRET_DIR}`.
 - Glob and brace expansion: `/etc/*`, `/a/{b,c}`.
-- Paths synthesized across tokens, here-docs, or quoted paths containing
-  whitespace.
+- Here-docs (`<<EOF`), paths synthesized across tokens, or quoted paths
+  containing whitespace.
 
 Operational impact: a confined session that references an out-of-root path via
 bash still runs, but the operator sees a one-time `[security]` warning and a
--- a/docs/decisions/0001-bash-tool-path-containment.md
+++ b/docs/decisions/0001-bash-tool-path-containment.md
@@ -1,0 +1,119 @@
+# ADR 0001 — Bash tool path containment
+
+Status: **Accepted**
+
+Tracked as **C4** (see [`CHANGELOG.md`](../../CHANGELOG.md)). Closes issue #354.
+
+---
+
+## Context
+
+Every typed filesystem handler (`read_file`, `write_file`, `edit_file`, `glob`,
+`grep`, `list_directory`) routes each path argument through
+`resolveAndContain` in
+[`src/agent/tools/handlers/_cwd-utils.ts`](../../src/agent/tools/handlers/_cwd-utils.ts).
+That function resolves the path (following symlinks via `realpathSafe`) and
+throws when it falls outside the session's `readRoots` / `writeRoots`. This is
+how a confined session — an `afk -w` worktree, or a forked sub-agent — is kept
+from reading or writing outside its granted roots.
+
+The bash handler
+([`src/agent/tools/handlers/bash.ts`](../../src/agent/tools/handlers/bash.ts))
+did **not** participate in this containment. It spawns commands with
+`shell: true` and only set the child's `cwd`; it never consulted
+`readRoots` / `writeRoots` and never routed anything through containment. The
+gap is structural, not an oversight: with `shell: true` the command is an
+**opaque string** interpreted by the OS shell. There is no reliable, general
+way to know which paths a command will touch — pipes, redirections, subshell
+and arithmetic substitution (`$(…)`, backticks), environment-variable
+indirection (`$HOME`), and globbing can all synthesize paths at runtime. A
+model running inside an isolated worktree could therefore still read or write
+outside it via bash (e.g. `cat /etc/hosts`, `echo x > ~/.ssh/authorized_keys`),
+even though the typed file tools would have refused.
+
+Issue #354 offered two acceptance paths: (1) have the bash handler enforce/scan
+the roots for absolute paths it can extract, or (2) document an architectural
+decision explaining why the gap is accepted. This ADR does **both** — it adopts
+a best-effort scan *and* records the reasoning and residual limits.
+
+## Decision
+
+1. **Adopt best-effort, advisory path scanning now.** Before spawning, the
+   handler extracts absolute and home-relative path tokens from the command
+   string (`extractCandidatePaths`) and checks each against the session's
+   **write** roots via `wouldBeRestricted(…, 'write', …)`. `writeRoots` is the
+   stricter and most relevant boundary for the "escape the worktree" threat, and
+   in practice `readRoots ⊇ writeRoots` for confined forks. On the first
+   out-of-root reference the handler emits one `[security]` `console.warn`
+   naming the escaping resolved paths and one `tool.bash_path_escape` telemetry
+   row (counts + `mode` only — never the raw command string, per the audit
+   §G.4 privacy rule already followed for overflow-kill telemetry).
+
+2. **Warn-only — never refuse.** The scan does not block execution. Issue #354
+   floated a "Tier-1" idea: refuse to run bash when `writeRoots` is set *and*
+   `permissionMode === bypassPermissions`. We explicitly **reject** that
+   fail-closed refusal. A top-level `afk -w` session runs under
+   `bypassPermissions` **and** carries a non-empty effective
+   `writeRoots = [worktree]`, so a refusal keyed on that condition would break
+   the primary human-driven worktree workflow. Forked sub-agents run
+   `permissionMode = 'default'` (they do **not** inherit bypass), so they would
+   never be refused anyway — meaning the refusal would only ever fire on the
+   legitimate top-level session. Warn-only preserves the workflow while making
+   the risk observable in logs and telemetry.
+
+3. **Defer full containment.** A full `execFile`-based refactor that disables
+   the shell (and could then contain paths precisely), or an OS-level sandbox,
+   remains the long-term fix and is still tracked under C4. Building a shell
+   parser to close the substitution/indirection gaps is a deliberate non-goal —
+   issue #354 calls it a rathole, and a partial parser would give a false sense
+   of containment.
+
+Under `allowAll` (bypass) `wouldBeRestricted` short-circuits to
+not-restricted, and an unconfined session (no `resolveBase`) is likewise never
+restricted — so those sessions produce zero warnings without any special-casing
+in the handler.
+
+## Threat model accepted
+
+The accepted model, per the issue, is **single-user-on-laptop with trusted task
+descriptions**: the human operator drives the session and is assumed not to be
+feeding it adversarial prompts crafted to exfiltrate their own files. Under that
+model the residual gap is an advisory concern (catch accidental out-of-worktree
+writes, keep the risk surface visible) rather than a hard security boundary. AFK
+is **not** hardened for running fully untrusted, adversarial task input through
+bash in `bypassPermissions`; that would require the deferred execFile/sandbox
+work.
+
+## Consequences
+
+**Covered** by the best-effort scan:
+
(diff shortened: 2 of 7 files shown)


griffinwork40:
Verified against source at `e5a49b1d`:
- **S1 confirmed real:** the token strip is `.replace(/^['"]/, '')` — it removes leading quotes only, not `>`/`>>`/`<`/`|`, so an operator-glued path (`>/etc/passwd`) keeps its leading operator, fails the `startsWith('/')` test, and is dropped. The ADR "Not covered" list omits this operator-adjacency case (it lists "redirections split across tokens" — the *spaced* form `> /etc/passwd`, which is actually *caught*). Real gap, but advisory-only feature → **low**.
- **S2 over-rated:** `wouldBeRestricted` is provably non-throwing (every branch returns a result; `realpathSafe` swallows its own errors and recurses to root). The only theoretical throw in `scanPathsBestEffort` is `os.homedir()`, which effectively never throws on a normal host. So the "inverts never-blocks" premise is near-unreachable → **downgrade to low** (defensive hardening, not a medium bug).
- **Perf/P1:** confirmed uncapped per-candidate `realpathSync.native`, but bounded by command token count and consistent with existing file-handler cost → **low**.

No critical/high findings → shadow-verify gate not triggered. Synthesizing locally (final judgment is the orchestrator's job).

---

# Code Review — PR #558

**Target:** PR #558 `security(bash): best-effort readRoots/writeRoots path scan + ADR (closes #354)` · branch `afk/tackle-issue-354` → `main` · reviewed ref `e5a49b1d`
**Change type:** feature / security hardening · 7 files · +426/−2 (428 lines) · **full** regime
**Stated intent:** PR #558 title + body (detailed) — spec-compliance **assessed**.

## Summary

Adds an **advisory-only** (warn, never block) `writeRoots` path-containment scan to the `bash` handler, closing the documented gap that bash — unlike the six typed file handlers — never consulted session roots. New `extractCandidatePaths()` tokenizer + `scanPathsBestEffort()` pre-spawn hook + a one-time `[security]` warn + `tool.bash_path_escape` telemetry (counts-only). Backed by an ADR (Accepted) that records the warn-only decision, the explicit rejection of a fail-closed refusal (would break the top-level `afk -w` bypass flow), the accepted threat model, and the residual gaps. Change is **additive, backward-compatible, and correctness-clean**. All findings are minor (hardening / coverage / doc completeness). No blocking issues.

## Findings

### security

**low · confidence:high · security · docs/decisions/0001-bash-tool-path-containment.md:117-124 + src/agent/tools/handlers/_cwd-utils.ts:243-249 · ref:e5a49b1d · citation-type:file-state**
Operator-glued path tokens (`cmd >/etc/passwd`, `>>~/.bashrc`, `|/tmp/x`) are silently dropped by the scan — the leading `>`/`|` fails both the `/`- and `~`-prefix tests — and this case is absent from the ADR's four-item "Not covered" list.
Evidence:
```
let token = rawToken.replace(/^['"]/, '').replace(/['";,)]+$/, '');
const isAbsolute = token.startsWith('/');
const isHomeRelative = token === '~' || token.startsWith('~/');
```
Suggestion: strip a leading run of shell-operator chars (`>`, `<`, `|`, `&`) before the prefix test, and add "redirection glued to a path (`>/path`)" to the ADR "Not covered" list. (Advisory-only feature — non-blocking.)

**low · confidence:high · security · src/agent/tools/handlers/bash.ts:192-195 · ref:e5a49b1d · citation-type:diff-context**
`scanPathsBestEffort` runs un-guarded before spawn; a throw would reject the handler and defeat the "never blocks" invariant. Verified the realistic blast radius is near-zero: `wouldBeRestricted` is total (non-throwing) and `realpathSafe` swallows its own errors, leaving only `os.homedir()` as a theoretical thrower.
Evidence:
```
if (context !== undefined) {
  scanPathsBestEffort(command, context);
}
```
Suggestion: wrap the call in `try { … } catch { /* advisory only */ }`, matching the fire-and-forget posture of `appendRoutingDecision`.

no issues found — telemetry row (`event`/`tool`/`restricted_count`/`mode`) and the persisted JSONL omit the raw command string and escaping paths (audit §G.4 honored); the `console.warn` naming resolved paths is ephemeral + ADR-documented. Read bash.ts:162-178, routing-telemetry.ts:117-123 at e5a49b1d.
no issues found — `extractCandidatePaths` regexes are anchored fixed-class patterns (no nested quantifiers/alternation) → no ReDoS. Read _cwd-utils.ts:236-253 at e5a49b1d.

### api-compat

no issues found — `extractCandidatePaths` is a new export; `RoutingDecisionEntry.restricted_count?` is optional-additive; `createBashHandler(permissionMode, cwd?)` signature unchanged (body-only). `git grep` at e5a49b1d shows each new symbol consumed only by its introducing file + its own test — no orphaned or broken call sites. Additive-only, no breaking changes.

### correctness

no issues found — read bash.ts + _cwd-utils.ts in full at e5a49b1d. Traced and confirmed: `~`/`~/…` expansion (`home + candidate.slice(1)` preserves the separator), the synchronous one-time latch (`_pathEscapeWarned`, no `await` interleaving), the load-bearing `context !== undefined` gate, the `resolveBase ?? cwd ?? factory-cwd` fallback tiering, and quote-stripping (`"/etc/hosts"` → `/etc/hosts`).

### spec-compliance

Judged against PR #558 title+body. Requirements **1–5 met**: additive `extractCandidatePaths` (+46/−0, existing `resolveAndContain`/`wouldBeRestricted` untouched); pre-spawn warn-only `scanPathsBestEffort` with counts-only telemetry; `allowAll`/no-`resolveBase` short-circuits inherited from `wouldBeRestricted`; 119-line ADR with all required sections incl. fail-closed rejection rationale; CHANGELOG C4 entry under `## [Unreleased]`. **No scope creep** — insertion counts match the C4-scoped blocks exactly.
`unverified — requirement 6 (tsc clean + vitest 96/96): not run. Verifying requires checking out the branch, which would mutate the DIRTY working tree (forbidden by the read-only preflight). Left to CI.`

### test-coverage

**low · confidence:high · test-coverage · src/agent/tools/handlers/bash.ts:155-160 · ref:e5a49b1d · citation-type:file-state**
The `~`/`~/…` home-expansion branch inside `scanPathsBestEffort` has no end-to-end test; only the tokenizer's *recognition* of `~` is tested (_cwd-utils.test.ts:255-258), not expand-then-contain. (`git grep` — zero `~`-prefixed command strings in the bash.test.ts C4 block.)
Suggestion: add a C4 case `cat ~/.ssh/id_rsa` asserting one warning naming the expanded home path.

**low · confidence:high · test-coverage · src/agent/tools/handlers/bash.ts:173-178 · ref:e5a49b1d · citation-type:file-state**
The `tool.bash_path_escape` telemetry payload is asserted nowhere — tests spy `console.warn` only. (`git grep` — zero `bash_path_escape` hits in bash.test.ts.)
Suggestion: mock `appendRoutingDecision` in one C4 test; assert `{ restricted_count, mode: 'write' }`.

### perf-observability

**low · confidence:medium · perf · src/agent/tools/handlers/_cwd-utils.ts:239 + bash.ts:161 · ref:e5a49b1d · citation-type:file-state**
Every bash call with a context pays one synchronous `realpathSync.native` per absolute/home token (only *roots* are memoized), uncapped, on the event loop before spawn. Realistically bounded by command token count and consistent with the existing typed-handler cost — negligible outside a pathological many-path command.
Suggestion: cap candidates scanned per command (e.g. first 32) to bound worst-case syscall count.

## What was not checked
- **Ref read:** Wave 1 agents + this synthesis read against branch HEAD `e5a49b1d` — all `blocking`-eligible citations verified at that ref (none reached high/critical).
- **tsc / vitest not run:** the PR's "tsc clean, 96/96 pass" claim is unverified — running the branch's suite requires a checkout that would mutate the dirty working tree (forbidden). Left to CI.
- **Stated intent:** PR #558 title+body available → spec-compliance assessed (5/6 verified; the 6th is the un-runnable verification claim above).
- **Not reviewed:** runtime behavior of `os.homedir()` throwing (external Node API); actual telemetry sink write path (`appendRoutingDecision` internals, unchanged by this PR).

## Merge decision: **MERGE**

Additive, advisory-only, backward-compatible, correctness-clean, ADR-documented, and tested for the core cases. All six findings are `low`/nit — none block. The three cheapest to fold in before merge if desired: strip leading shell operators in the tokenizer (S1), wrap the scan call in try/catch (S2), and add a `~/…` end-to-end test (TC1). None are prerequisites.

*(Read-only review — no files, commits, or PR state were modified.)*

---

Issue #547, opened 2026-07-13
skill / inline-skill fork managers don't propagate parent read scope (parentReadRoots) — asymmetry with the agent tool


griffinwork40:
## Summary

The read-scope inheritance fix (#544) wired **transitive read-scope propagation** into the **`agent` tool** executor only (`subagent-executor.ts` → `getReadScopeInputs()` → `childInheritedReadRoots` → `child-config.ts`). The **`skill` tool** and **inline-skill** dispatch paths build their *own* `SubagentManager`s and **never pass `parentReadRoots`**, so they do not propagate the parent session's read scope. Under worktree-based dispatch this under-grants skill-forked children relative to `agent`-tool children — the same class as the original bug, in a different dispatch path.

Note: `/diagnose` — a skill, and the origin of the read-denial bug #544 fixed — uses worktree isolation, so this is the directly-relevant residual.

## Confirmed (code-read on the merged fix)

- **No skill manager passes `parentReadRoots`.** `grep -rn parentReadRoots src/skills src/bundled-plugins` → empty. All ~20 `new SubagentManager(...)` sites outside `subagent.ts` (`skill-executor/fork-dispatch.ts:81`, `skill-executor/fork-child-config.ts:106`, `mint/_phases/*` ×8, `audit-fit/index.ts:347`, `compose-executor.ts:577`, `user-skills.ts:156`) omit it.
- Skill managers pass `cwd: currentCwd` where `currentCwd = ctx.cwd` (`skill-executor.ts:87`; forwarded at `fork-dispatch.ts:95`, `fork-child-config.ts:113`).
- `computeInheritedReadRoots` takes the **CONFINED** branch whenever `parentCwd` is defined and `parentReadRoots` is undefined → child = `union(childCwd, cwd, worktreeMainRoot)`, never read-open (`subagent-read-scope.ts`).

## Consequence (conditional)

| Session / skill state | Skill fork read scope | Gap? |
|---|---|---|
| Plain REPL (`ctx.cwd` undefined) | manager unconfined → **read-open** | No |
| `--worktree`, or a skill that `setCwd`s to a worktree (e.g. `/diagnose`) | `[childCwd, cwd, mainRoot]` — repo + worktree + main root, but **not** `/tmp`, `~/.afk/state`, out-of-repo | **Yes** — narrower than an unconfined parent; the equivalent `agent`-tool dispatch (per-call `cwd` under an unconfined session) yields read-open |

[...]

PR #560, opened 2026-07-13
fix(path-approval): honor unconfined resolveBase in the hook + add read-denylist floor

griffinwork40:
## Summary

Forked sub-agents got **every** typed file-tool call (`read_file` / `glob` / `grep`) auto-denied by the PreToolUse path-approval hook whenever the top-level session was **unconfined** — a plain `afk i` REPL with no `-w` and no resume. Sub-agents fell back to reading via `bash`, silently degrading `/review`, `/diagnose`, and every read-only fan-out. This is the failure the last several read-scope PRs were chasing.

## Root cause: two enforcement layers disagreed on `resolveBase === undefined`

- The file-tool **handler** (`resolveAndContain`, `_cwd-utils.ts`) treats an undefined `resolveBase` as **unconfined → allow** — a documented invariant (issue #434): _"Do NOT 'fix' this by defaulting resolveBase to `process.cwd()`… would reject ALL paths."_
- The **hook** did exactly that: `resolveBase: grants.resolveBase ?? cwd`, and the REPL wires `getCwd` to `effectiveCwd ?? process.cwd()` (`bootstrap.ts`), so it **fabricated a concrete base** for an unconfined session.

A no-cwd fork of an unconfined parent inherits an **empty** `readRoots` (`ensureSharedRoots(undefined) → []`). With a fabricated concrete base and `[] ?? [base]` staying `[]`, **every** path was "restricted" — and forks can't prompt, so all reads auto-denied (deny-all).

The six prior read-scope PRs (#416 / #441 / #514 / #544 / #547 / #554) only widened the **confined-parent** `readRoots` union — a path an unconfined-parent fork never takes (`computeInheritedReadRoots` returns `undefined` for a no-cwd child), so they never engaged with this failure. The bug was in the **enforcement** layer, not the grant layer.

Confirmed by reproducing the real `createPathApprovalHook` under the REPL's exact conditions (concrete `getCwd`, unconfined child grants, `parentSessionId` set): block reasons were byte-identical to the failing witness trace.

## Fix

1. **Align the hook to the handler.** When `grants.resolveBase === undefined` (and not `allowAll`), the hook now treats the session as unconfined and skips typed-file containment — matching `resolveAndContain`. The `?? cwd` fabrication is gone.
2. **Add a read-denylist floor** (`read-denylist.ts`), enforced unconditionally in `resolveAndContain` **and** the hook, *before* the `allowAll`/unconfined fast-paths. This closes a **pre-existing read/write asymmetry**: writes were denylisted (`write-denylist.ts`) but reads had **no** secret-path floor, so any unconfined session — and, after fix #1, its forks — could read credential stores (SSH keys, cloud credentials, the `afk.env` config tree).
   - The AFK **state** dir stays **readable** (forks legitimately read skill-preflight inputs, todos, transcripts — #544/#547/#554). Only the **config** tree is denied.
   - `/etc` is **not** blanket-denied (ordinary system reads keep working); only enumerated secret files are.
   - Extendable via the new `AFK_READ_DENYLIST` env var.

## Tests

- **hook ↔ handler parity matrix** — asserts the hook's verdict equals `resolveAndContain`'s across a grants matrix (unconfined/confined/bypass, in/out of root). Turns _"the two layers must agree"_ into a CI tripwire so this class of divergence can't silently return.
- **deny-all regression** — an unconfined fork can read a repo file, glob/grep arbitrary trees, and read the AFK state dir.
- **read-denylist floor** — credential paths blocked for forks and in bypass mode; state dir allowed, config tree denied.
- **read-denylist unit tests** — credential coverage, symlink dereference, `AFK_READ_DENYLIST` extras, the state-vs-config divergence.

Full suite: **635 files, 11571 passed, 14 skipped**; `tsc --noEmit`, `scan:env:check`, `audit:env:check` all green.

## Deliberately out of scope

Forked children don't inherit the parent's `permissionMode` (a child defaults to `'default'` while the parent may be `bypass`) — that's why `main` (bypass → `allowAll`) reads fine while the children hit the deny. It's a separate concern with its own security implications and is **not** bundled here.

### DIFF
--- a/docs/env-registry.json
+++ b/docs/env-registry.json
@@ -473,6 +473,14 @@
       "category": "model"
     },
     {
+      "name": "AFK_READ_DENYLIST",
+      "description": "Colon-separated list of additional absolute paths the read_file/grep/glob/list_directory tools refuse to read. Built-in credential entries (~/.ssh, ~/.aws, ~/.afk/config, …) always apply on top and cannot be removed.",
+      "type": "string",
+      "required": false,
+      "example": "/Users/me/project/.env:/Users/me/secrets",
+      "category": "misc"
+    },
+    {
       "name": "AFK_RUN_RECEIPT_DISABLED",
       "description": "Disable the post-session run receipt (state/receipts/<label>.json and .md). Set to 1 to skip receipt writes; the underlying witness trace is unaffected. Receipts are also implicitly off when AFK_TRACE_DISABLED=1 (no trace to summarize).",
       "type": "boolean",
--- a/docs/env-registry.md
+++ b/docs/env-registry.md
@@ -2,7 +2,7 @@
 
 Generated from `src/config/env.ts`. Do not edit by hand — run `pnpm scan:env` after changing the registry source.
 
-**120 vars** across 12 categories. Every `process.env[...]` read in `src/` outside `src/config/env.ts` is a CI failure (enforced by `pnpm audit:env:check`).
+**121 vars** across 12 categories. Every `process.env[...]` read in `src/` outside `src/config/env.ts` is a CI failure (enforced by `pnpm audit:env:check`).
 
 To add a var: edit `src/config/env.ts` (add a getter on `env` + an entry in `ENV_REGISTRY`), then run `pnpm scan:env`.
 
@@ -178,6 +178,7 @@ To add a var: edit `src/config/env.ts` (add a getter on `env` + an entry in `ENV
 | `AFK_DEMO_CLEAN` | boolean |  |  | `1` | Explicit opt-in to capture-mode. When set to 1, suppresses high-frequency repaint drivers (spinner ticker, live thinking-preview) so recorded artifacts contain each state once instead of once per timer tick. |
 | `AFK_DIFF_LINES` | number |  |  | `50` | Maximum number of diff lines shown in the inline diff render during write_file tool calls. Set to 0 for no cap. Non-integer values are silently ignored and the default applies. |
 | `AFK_MEMORY_EVIDENCE_GATE` | boolean |  |  | `1` | Opt-in (set to 1) evidence gate for durable memory writes. When enabled, a codebase fact (memory_update category "convention") stored without an `evidence` citation is recalled as [unverified], and memory_search results carry a verification verdict. User preferences and agent reflections are never gated. Default off — memory behaves identically to legacy when unset. |
+| `AFK_READ_DENYLIST` | string |  |  | `/Users/me/project/.env:/Users/me/secrets` | Colon-separated list of additional absolute paths the read_file/grep/glob/list_directory tools refuse to read. Built-in credential entries (~/.ssh, ~/.aws, ~/.afk/config, …) always apply on top and cannot be removed. |
 | `AFK_SHELL_PASSTHROUGH` | boolean |  | `1` | `0` | Enable the interactive REPL `!cmd` / `!&cmd` shell-passthrough feature. On by default. Set to 0, false, off, or no (case-insensitive) to disable, so inputs beginning with ! are sent to the model as literal text instead of being executed as shell commands. Equivalent to the --no-shell-passthrough flag. |
 | `AFK_SHOW_DIFFS` | boolean |  |  |  | Show inline diffs in the tool-lane output for edit/write tool calls. 1 = on, 0 = off. |
 | `AFK_SPINNER_TIPS` | boolean |  |  |  | Show rotating tips in the loading spinner during long calls. 1 = on, 0 = off. |
--- a/src/agent/tools/handlers/_cwd-utils.ts
+++ b/src/agent/tools/handlers/_cwd-utils.ts
@@ -11,6 +11,7 @@
 import path from 'path';
 import { realpathSync } from 'fs';
 import type { ToolHandlerContext } from '../types.js';
+import { isReadDenied } from './read-denylist.js';
 
 // Invariant: symlink containment must be resolved at the filesystem level, not
 // lexically. A symlink that lives INSIDE a granted root but points OUTSIDE it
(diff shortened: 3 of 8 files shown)


griffinwork40:
All confirmations hold: main has **120** env vars (PR's 121 is correct), `wouldBeRestricted`'s only callers are inside the hook itself (change is localized), `AFK_READ_DENYLIST` is genuinely new, and neither denylist covers `~/.git-credentials`/`~/.netrc`/`gh hosts.yml`/`~/.kube/config`. My analysis is complete across all five dimensions.

---

# Code Review — PR 560: `fix(path-approval): honor unconfined resolveBase + read-denylist floor`

**Verdict: ✅ APPROVE** (no blockers; 2 medium hardening/test items worth addressing pre- or post-merge)

## What it does
Two coupled changes:
1. **Bug fix (sub-agent deny-all):** the PreToolUse hook fabricated a confinement base for unconfined sessions (`resolveBase: grants.resolveBase ?? cwd`), so a no-cwd fork with empty `readRoots` had *every* read flagged restricted → auto-denied, while the handler (`resolveAndContain`) correctly allowed it. The hook now honors `resolveBase === undefined` (allow), matching the handler.
2. **Security floor:** a new `read-denylist.ts` blocks credential/secret reads (`~/.ssh`, `~/.afk/config`, …) unconditionally in *both* the hook and `resolveAndContain`, closing the read/write asymmetry (writes had a floor; reads had none, so any unconfined session could read `~/.ssh`).

## Dimension results

**Correctness — CLEAN (high confidence).** All four root-cause claims verified against source: unconfined REPL has `resolveBase===undefined` while `getCwd()` is concrete (`bootstrap.ts` → `ensureSharedRoots`); fork inherits `readRoots: []` (`subagent-read-scope.ts` → `grant-manager.ts:180`); `[] ?? [base]` stays `[]` so the containment loop iterates zero roots → `restricted:true` (the bug); new branch order (floor → allowAll → unconfined → confined) mirrors `resolveAndContain` exactly.

**Security — CLEAN (high confidence).** All four read handlers funnel through `resolveAndContain(..., 'read')` (`read-file.ts:65`, `glob.ts:220`, `grep.ts:79`, `list-directory.ts:54`). I specifically **refuted** the glob/grep "absolute-pattern-without-path" bypass: grep's `pattern` is a content regex (root is always the contained `path` argv); glob's `pattern` matches only paths *relative to* the contained `basePath` and doesn't recurse symlinked dirs. The floor runs **before** both the `allowAll` bypass and the unconfined short-circuit in both layers, so credential reads are blocked even in `bypassPermissions`. Symlinks are dereferenced via `safeRealpath` before prefix comparison.

**API-compat — CLEAN.** `AFK_READ_DENYLIST` correctly added to `ENV_REGISTRY` + `env` getter + regenerated docs (120→121 confirmed); additive; matches `AFK_WRITE_DENYLIST` shape. `wouldBeRestricted` signature unchanged; change is localized to the hook.

## Findings (none blocking)

| Sev | Finding |
|-----|---------|
| **MEDIUM** | **Denylist completeness (hardening).** `BUILTIN_READ_DENYLIST` omits high-value credential files that this agent routinely operates near: `~/.git-credentials`, `~/.config/gh/hosts.yml` (GitHub OAuth token — an exfil could push to repos), `~/.netrc`, `~/.kube/config`. Pre-existing gap (the *write* list misses them too), but this PR is explicitly about closing credential-read exposure — adding the git/gh ones especially would materially strengthen it. |
| **MEDIUM** | **Vacuous symlink test.** `read-denylist.test.ts` "blocks a read through a symlink that resolves to ~/.ssh" does `if (!existsSync(sshDir)) return;` — on CI runners without `~/.ssh` it silently passes asserting nothing. Make it robust: symlink into a `tmpDir` added via `AFK_READ_DENYLIST`, or create a fixture dir, so the dereference path is always exercised. |
| **LOW** | **Dead export.** `assertNotReadDenied` is defined + tested but has **zero** production callers (grep confirmed) — read enforcement is inline via `isReadDenied` in `resolveAndContain`, whereas the write sibling `assertNotDenylisted` *is* wired into `write-file.ts:83`/`edit-file.ts:115`. Either wire it (for symmetry) or drop it; exported fns escape `noUnusedLocals` so CI won't flag it. |
| **LOW** | **Test coverage shape.** Denylist enforcement is only tested at the hook level for `read_file` (the `CREDS` loop) and directly via `resolveAndContain`; `glob`/`grep`/`list_directory` denylist blocking is only *transitively* covered. No end-to-end test invokes an actual handler and asserts `isError` for a credential path. The parity matrix covers containment but not denylist paths. Add a param-over-4-tools denylist case + one handler-level assertion. |
| **LOW** | **Intended behavior changes to call out in release notes.** (a) `bypassPermissions` no longer permits credential *reads* via typed tools; (b) the whole `~/.afk/config` tree is now read-blocked, including non-secret `afk.config.json`/`mcp.json` — mitigated by the `config_get` tool and `/mcp`, but a "read my mcp.json" request will now error. Both are correct-by-design; just worth documenting. |
| **NIT** | `postToolUseImpl`'s `grants.resolveBase ?? entry.capturedCwd` is now a dead fallback (once-grants only exist for confined sessions where `resolveBase` is defined) — harmless, but a one-line comment would prevent a future reader thinking it's load-bearing. |
| **NIT** | **Minor perf on the read hot path.** Unconfined reads now pay one extra `realpath` syscall (`isReadDenied`→`safeRealpath`) they previously skipped; confined reads realpath the candidate twice (floor + containment). Cheap vs. the I/O of the read itself, and the denylist entries are cached — acceptable. Could dedupe by threading the resolved path into `isReadDenied`. |

## Strengths
- Faithfully mirrors the established `write-denylist.ts` pattern; the floor-before-bypass ordering gives true read/write symmetry.
- The **hook↔handler parity matrix** turns "these two layers must agree" (a divergence the PR notes recurred across six PRs) into a CI tripwire — excellent regression insurance.
- Root cause fixed at the source (unconfined *means* unconfined) rather than patching `readRoots ?? [base]`; comments capture the invariant and the `#544/#547/#554` history for the deliberate `~/.afk/state` divergence.

## Scope of verification
Verified by static analysis against on-disk handlers (unmodified by the PR) + the diff artifact + the PR branch via git. **Not run:** the test suite (`pnpm test`) — the PR branch isn't checked out and the preflight forbids mutating the dirty working tree; I relied on code reading. Recommend the author confirm `pnpm test -- read-denylist path-approval-hook`, `pnpm lint`, and `pnpm scan:env:check` are green before merge (the diff is internally consistent, so I expect them to pass).

**Recommendation:** Merge. Optionally fold in the two MEDIUM items (add `~/.git-credentials`/`~/.config/gh/hosts.yml`/`~/.netrc`/`~/.kube/config` to the builtin list; de-vacuous the symlink test) either in this PR or a fast follow — neither blocks the correctness/security value this PR delivers.

## Done
- **What was done:** Full 5-dimension review of PR 560 (security, correctness, api-compat, test-coverage, perf/observability), read-only.
- **Evidence:** Diff artifact `~/.afk/state/skill-preflight/bb7c4128-.../pr-560.diff`; source verified at `read-file.ts:65`, `glob.ts:202-220`, `grep.ts:74-79,156`, `list-directory.ts:54`, `_cwd-utils.ts:87-207`, `path-approval-hook.ts` (baseline) + PR branch `fix/path-approval-unconfined-deny-all` @ `186b643`. Root-cause chain independently confirmed by a completed research subagent (`bootstrap.ts`→`ensureSharedRoots`→`grant-manager.ts:180`).
- **World change:** None — no files edited, no commit, no push, no PR comment (review is read-only; working tree left untouched as instructed).
- **Deferred:** Did not execute `pnpm test`/`lint`/`scan:env:check` (branch not checked out; working tree must stay clean) — flagged for the author to confirm green pre-merge.

---

PR #562, opened 2026-07-13
feat(telegram): multi-session support — session-registry foundation + /sessions switcher

griffinwork40:
## Summary

Foundational multi-session support for the Telegram surface, delivered in two coherent layers:

1. **Session-registry foundation** (steps 1–3) — a surface-agnostic session-identity layer + a Telegram routing primitive, additive and fully unit-tested.
2. **Free session switcher** (steps 4–5) — user-facing `/sessions` (tap to switch) and `/new`, so one Telegram chat can hold many independent conversations and jump between them.

One active session per chat at a time; every past conversation persists as a resumable sidecar. No breaking changes — existing single-session behavior is byte-identical.

Design doc: `.afk/plans/session-registry-architecture.md`.

## What ships

| Commit | What |
|---|---|
| `eff1706` | **Registry core** (`session-registry.ts`) — `SessionHandle` model, opaque surface→handle bindings, `resolve`/`create`/`bind`/`archive`. `handle.id` is registry-minted + stable; `sdkSessionId` is a **late-bound** attribute (undefined until the provider stream emits it), never a routing key. Branded `HandleId` makes a raw-`chatId` mis-key a type error. |
| `a40a11f` | **Persistence bridge** — `StoredSession.{handleId,bindings}` (optional, back-compat) + pure `storedToHandle`/`bindingsForStored`/`storedFieldsFromHandle` mappings, incl. the legacy `telegramChatId → telegram binding` migration; `registry.load()` for rehydration. |
| `35ff346` | **Route primitive** (`route.ts`) — `routeFromCtx`/`routeKey`/`sendOptions`. Reads a topic `message_thread_id` off `message`, `edited_message`, **and** `callback_query.message` (a button tapped in a topic carries its thread id on the callback, not `ctx.message`). General normalizes to the bare-chatId key, so topics-off users are unchanged. |
| `dde025d` | **SessionManager capability** — `switchToSession` (close live session, stage resume, adopt target model/cwd), `listChatSessions` (a chat's resumable sidecars, active flagged), `newSession`. `pendingResume` is consumed once by `getSession` and cleared on every teardown. `telegram.ts` threads `resume` into both session-construction branches. |
| `af5ee93` | **Switcher UX** — `/sessions` (inline tap-to-switch buttons, callback `afk:sw:<sessionId>`), `/new`. Both refuse while the active session is mid-turn; switching stages a **lazy** resume that continues on the next message (the callback never blocks on a resume replay). |
| `b7bd2d3` | `/help` + command-menu discoverability. |

## Key design decisions

- **Late-bound `sdkSessionId`.** The registry mints its own stable `HandleId` synchronously at `create()`; the SDK session id is attached later. This sidesteps the "id is undefined at construction" footgun and keeps live-session maps keyed on a stable value.
- **Switcher rides the sidecar store + `resume`, not a live-map re-key.** Because only one session is active per chat, `SessionManager`'s existing `chatId`-keyed maps are untouched (low blast radius). A chat's session list comes from the shared sidecar store (already keyed by `telegramChatId`); switching is close-active + `config.resume=<target>` — the same mechanism the CLI `--resume` uses.
- **The registry is not yet in the switcher hot path.** Its binding model (one active binding per key) is the right primitive for *routing* (native topics, cross-surface), which is where it becomes load-bearing — see Deferred. It ships now as the tested foundation the architecture was committed to.

## Intentionally deferred (follow-up PRs)

- **Step 6 — native Telegram "topics in private chats".** Concurrent sessions in one chat (one per topic) via `routeFromCtx`. This is the only path that needs re-keying `message.ts`'s `pendingElicitations`/`messageQueues` by session identity (the latent cross-session-leak fix). Gated on a product decision: topics appear to be a paid (Telegram Stars) feature. The `route.ts` primitive here is that step's ingress.
- **Step 7 — cross-surface bindings** (CLI/daemon route through the same registry).

## Testing

- **+40 unit tests** across registry (`load`, bridge mappings), route primitive (incl. the callback-thread leak-fix case), `SessionManager` switch/list/resume, and the `/sessions`+`/new`+switch-callback handlers.
- Full suite green: **11,645 passed / 14 skipped (638 files)**.
- `tsc --noEmit`, `audit:env:check`, `audit:sdk:check`, `scan:env:check` all clean.

## Back-compat / risk

- All changes additive; existing Telegram flows unchanged and fully covered by the pre-existing suite (which passes unmodified).
- `/switch` and `/new` refuse mid-turn, so no streaming reply is orphaned.
- Resume is provider-dependent (same contract as CLI `--resume`); the wiring is verified, context-restoration is the provider's job.

### DIFF
--- a/src/agent/session/session-registry.test.ts
+++ b/src/agent/session/session-registry.test.ts
@@ -1,0 +1,213 @@
+/**
+ * Unit tests for the surface-agnostic session registry.
+ *
+ * Encodes the identity contract from session-registry-architecture.md:
+ *   - handle.id is registry-minted + stable; sdkSessionId is late-bound.
+ *   - resolve(surface, key) routes by binding; many bindings → one handle.
+ *   - archive frees keys; accessors return snapshots; mutators fail loud.
+ */
+
+import { describe, it, expect } from 'vitest';
+import {
+  InMemorySessionRegistry,
+  createSessionRegistry,
+  asHandleId,
+  SessionRegistryError,
+  type SessionRegistry,
+} from './session-registry.js';
+
+/** A registry with a controllable clock for deterministic lastActiveAt ordering. */
+function makeReg(start = 1000): { reg: SessionRegistry; tick: (by?: number) => number } {
+  let t = start;
+  const reg = createSessionRegistry({ now: () => t });
+  return { reg, tick: (by = 1) => (t += by) };
+}
+
+describe('create / get', () => {
+  it('mints a stable id, defaults to active, no sdkSessionId', () => {
+    const { reg } = makeReg();
+    const h = reg.create({ surface: 'telegram', model: 'sonnet', key: '42' });
+    expect(h.id).toBeTruthy();
+    expect(h.status).toBe('active');
+    expect(h.sdkSessionId).toBeUndefined();
+    expect(h.bindings).toEqual([
+      { surface: 'telegram', key: '42', boundAt: 1000, lastActiveAt: 1000 },
+    ]);
+    expect(reg.get(h.id)?.id).toBe(h.id);
+  });
+
+  it('two distinct keys on the same chat produce two isolated handles', () => {
+    const { reg } = makeReg();
+    const a = reg.create({ surface: 'telegram', model: 'sonnet', key: '42' });
+    const b = reg.create({ surface: 'telegram', model: 'sonnet', key: '42:7' });
+    expect(a.id).not.toBe(b.id);
+    expect(reg.resolve('telegram', '42')?.id).toBe(a.id);
+    expect(reg.resolve('telegram', '42:7')?.id).toBe(b.id);
+  });
+
+  it('honors a rehydration id + createdAt and rejects a duplicate id', () => {
+    const { reg } = makeReg();
+    const id = asHandleId('fixed-id');
+    const h = reg.create({ surface: 'cli', model: 'opus', key: 'resume-me', id, createdAt: 5 });
+    expect(h.id).toBe(id);
+    expect(h.createdAt).toBe(5);
+    expect(() => reg.create({ surface: 'cli', model: 'opus', key: 'other', id })).toThrow(
+      SessionRegistryError,
+    );
+  });
+
+  it('carries optional name / cwd / sdkSessionId through create', () => {
+    const { reg } = makeReg();
+    const h = reg.create({
+      surface: 'daemon',
+      model: 'haiku',
+      key: 'task-1',
+      name: 'nightly',
+      cwd: '/repo',
+      sdkSessionId: 'sdk-1',
+    });
+    expect(h.name).toBe('nightly');
+    expect(h.cwd).toBe('/repo');
+    expect(reg.getBySdkSessionId('sdk-1')?.id).toBe(h.id);
+  });
+});
+
+describe('resolve', () => {
+  it('returns undefined for an unbound key', () => {
+    const { reg } = makeReg();
+    reg.create({ surface: 'telegram', model: 'sonnet', key: '42' });
+    expect(reg.resolve('telegram', '999')).toBeUndefined();
+    expect(reg.resolve('cli', '42')).toBeUndefined(); // surface is part of the key
+  });
+});
+
+describe('bind — many bindings → one handle (cross-surface)', () => {
+  it('a second binding resolves to the same handle', () => {
+    const { reg } = makeReg();
+    const h = reg.create({ surface: 'cli', model: 'sonnet', key: 'repl-abc' });
+    reg.bind(h.id, { surface: 'telegram', key: '42:7' });
+    expect(reg.resolve('cli', 'repl-abc')?.id).toBe(h.id);
+    expect(reg.resolve('telegram', '42:7')?.id).toBe(h.id);
+    expect(reg.get(h.id)?.bindings).toHaveLength(2);
+  });
+
+  it('re-binding an existing key refreshes rather than duplicating', () => {
+    const { reg, tick } = makeReg();
+    const h = reg.create({ surface: 'telegram', model: 'sonnet', key: '42' });
+    tick(50);
+    reg.bind(h.id, { surface: 'telegram', key: '42' });
+    const got = reg.get(h.id)!;
+    expect(got.bindings).toHaveLength(1);
+    expect(got.bindings[0]!.lastActiveAt).toBe(1050);
+    expect(got.bindings[0]!.boundAt).toBe(1000);
+  });
+
+  it('re-points a key away from a prior owner (last-writer-wins)', () => {
+    const { reg } = makeReg();
+    const a = reg.create({ surface: 'telegram', model: 'sonnet', key: '42' });
+    const b = reg.create({ surface: 'telegram', model: 'sonnet', key: '99' });
+    reg.bind(b.id, { surface: 'telegram', key: '42' }); // steal '42' from a
+    expect(reg.resolve('telegram', '42')?.id).toBe(b.id);
+    expect(reg.get(a.id)?.bindings).toHaveLength(0); // a lost the binding
+    expect(reg.get(b.id)?.bindings).toHaveLength(2);
+  });
+});
+
+describe('unbind', () => {
+  it('removes the binding from its owner and stops resolving', () => {
+    const { reg } = makeReg();
+    const h = reg.create({ surface: 'telegram', model: 'sonnet', key: '42' });
+    reg.unbind('telegram', '42');
+    expect(reg.resolve('telegram', '42')).toBeUndefined();
+    expect(reg.get(h.id)?.bindings).toHaveLength(0);
+  });
+
+  it('is a no-op for an unknown key', () => {
+    const { reg } = makeReg();
+    expect(() => reg.unbind('telegram', 'nope')).not.toThrow();
+  });
+});
+
+describe('attachSdkSessionId — late binding', () => {
+  it('indexes for reverse lookup and updates on re-attach', () => {
+    const { reg } = makeReg();
+    const h = reg.create({ surface: 'telegram', model: 'sonnet', key: '42' });
+    reg.attachSdkSessionId(h.id, 'sdk-1');
+    expect(reg.getBySdkSessionId('sdk-1')?.id).toBe(h.id);
+    reg.attachSdkSessionId(h.id, 'sdk-2');
+    expect(reg.getBySdkSessionId('sdk-1')).toBeUndefined(); // stale index dropped
+    expect(reg.getBySdkSessionId('sdk-2')?.id).toBe(h.id);
+    expect(reg.get(h.id)?.sdkSessionId).toBe('sdk-2');
+  });
+});
+
+describe('rename', () => {
+  it('sets the human label', () => {
+    const { reg } = makeReg();
+    const h = reg.create({ surface: 'telegram', model: 'sonnet', key: '42' });
+    reg.rename(h.id, 'fix-the-thing');
+    expect(reg.get(h.id)?.name).toBe('fix-the-thing');
+  });
+});
+
+describe('archive', () => {
+  it('frees keys (resolve misses) but retains the handle for get/list', () => {
+    const { reg } = makeReg();
+    const h = reg.create({ surface: 'telegram', model: 'sonnet', key: '42' });
+    reg.archive(h.id);
+    expect(reg.resolve('telegram', '42')).toBeUndefined();
+    expect(reg.get(h.id)?.status).toBe('archived');
+    expect(reg.list({ status: 'archived' }).map((x) => x.id)).toContain(h.id);
+  });
+
+  it('a new session can reclaim an archived handle key', () => {
+    const { reg } = makeReg();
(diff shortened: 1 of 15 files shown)


griffinwork40:
All four load-bearing clearances independently **CONFIRMED** at the ref (allowlist gate fail-closed before the new commands, path-traversal guard throws before any read, `telegramChatId` genuinely persisted, and the switcher's target id is the same id space end-to-end). The merge decision's foundation holds. Synthesizing (Wave 2, inline — small finding set).

---

# Code Review — PR #562

**feat(telegram): multi-session support — session-registry foundation + /sessions switcher**

| | |
|---|---|
| **Target** | PR #562 · `main` ← `feat/telegram-session-registry` |
| **Reviewed ref** | `b7bd2d34927723ec498af521e89b20a32da5b294` |
| **Change type** | feature · 15 files · **+1635 / -1** · regime **full** |
| **Stated intent** | PR title+body — **spec-compliance assessed** |
| **Verification** | 4 load-bearing clearances independently re-derived at the ref (all CONFIRMED) |

## Summary
A clean, additive, well-tested foundation PR. Every high-value security/correctness risk was verified *not* to be a bug: path traversal is caught by a pre-existing `safeResolvePath` guard plus an independent chat-ownership check; a fail-closed global allowlist middleware gates `/sessions` and `/new` before they run; `telegramChatId` is genuinely persisted (the cross-chat filter is real, not a no-op); and the switcher's target id is the same id-space end-to-end (`sessionId` == sidecar filename == `loadSession` lookup key), so it actually finds its targets. **No critical, high, or blocking findings.** Two medium polish items and three low items follow.

### Medium

**1 · medium · confidence:high · correctness · `src/telegram/session-manager.ts:640-645` · ref:b7bd2d3 · file-state**
`switchToSession` awaits `old.close()` *before* `this.sessions.delete(chatId)`; if `close()` throws, the function aborts before adopting the target's `sessionId`/model and staging `pendingResume`, and the stale session stays keyed in `this.sessions` — so the next `getSession` returns the old session unrebuilt instead of retrying the switch.
```
const old = this.sessions.get(chatId);
if (old) { await old.close(); this.sessions.delete(chatId); }
```
*Suggestion:* guard the close like `closeAll()` already does (`await old.close().catch(err => log(...))`) so a close failure never blocks the delete + state adoption. *Caveat:* the same unguarded pattern pre-exists in `resetSession`/`switchModel`/`setCwd` — this PR **extends** an existing gap, it does not introduce a new one.

**2 · medium · confidence:high · perf-observability · `src/telegram/handlers/sessions.ts:153-155` (+ `session-manager.ts:switchToSession`) · ref:b7bd2d3 · file-state**
`handleSwitchCallback` re-calls `listChatSessions` — a full `listSessions()` directory scan that JSON-parses *every* sidecar in the store — solely to fetch the switched-to session's `name`, a value already loaded into `stored.name` inside `switchToSession` and then discarded from its `{ ok: true }` return.
```
const name = sessionManager.listChatSessions(chatId).find((s) => s.sessionId === targetSessionId)?.name;
```
*Suggestion:* return `{ ok: true, name: stored.name }` from `switchToSession` and use it directly, skipping the extra full-store rescan on every switch tap.

### Low

**3 · low · confidence:high · security · `src/cli/session-store.ts:217-231` · ref:b7bd2d3 · file-state**
`loadSession` swallows path-escape/traversal exceptions with a bare `catch { return undefined }` and no logging, so a crafted `afk:sw:` traversal probe leaves no audit trail. (Pre-existing code — reused, not introduced by this PR; the traversal itself is safely blocked.)
```
try { path = safeResolvePath(idOrPath); } catch { return undefined; }
```
*Suggestion:* log the caught error (id + reason) before returning `undefined`.

**4 · low · confidence:high · test-coverage · `src/telegram/handlers/sessions.ts:76` · ref:b7bd2d3 · file-state**
No test constructs a `sessionId` long enough to trip the `CALLBACK_DATA_MAX` (64-byte) button filter, so the branch where a session appears in the `formatSessionsList` text but has *no* switch button (and the text/button numbering diverges) is unexercised. Unreachable with UUID-length SDK ids → genuinely low.
*Suggestion:* add a test with a >57-char `sessionId` asserting text-list presence + button absence.

**5 · low · confidence:high · test-coverage · `src/telegram/session-manager.test.ts` (switcher block) · ref:b7bd2d3**
No test configures `close()` to reject, so the state-consistency gap in finding #1 is unverified by the suite.
*Suggestion:* add a rejecting-`close()` mock and assert `getSession` still recovers after a failed switch.

### Nits
The `CALLBACK_DATA_MAX` guard measures `.length` (UTF-16 code units) against a byte limit; harmless for ASCII SDK ids, and the code comment already frames it as belt-and-braces.

## Spec-compliance
- **Unmet intent:** none — every named element (registry handle/binding model + `resolve/create/bind/unbind/archive/load`, the `storedToHandle`/`bindingsForStored`/`storedFieldsFromHandle` bridge + legacy `telegramChatId` migration, `route.ts` with the `callback_query.message` thread-id fallback, `/sessions`+`/new`+`afk:sw:` callback, mid-turn refusal, lazy `config.resume`) is present and file-verified.
- **Scope creep:** none — every touched file is required wiring for the stated switcher feature.
- **Byte-identical back-compat claim holds:** `pendingResume` is empty for any chat that never switched, so `config.resume` is never set for existing single-session users; the new `StoredSession` fields are optional with zero production readers.
- **Declared deferral honored:** `route.ts` + the registry singleton have zero non-test production importers at the ref — matching the PR's explicit "tested foundation, not yet in the hot path" claim. Not flagged as dead code.

## What was not checked
- **Ref:** Wave 1 + verifier read against branch HEAD `b7bd2d3`; all four load-bearing clearances and every blocking-candidate citation were verified at that ref. The working tree is a *later* branch (`feat/telegram-topics-crosssurface`) building on this PR — findings pertain to the PR head, not that follow-up.
- **Test suite not run:** the "11,645 passed / 638 files" claim is `unverified` (read-only surface; ~48 new `it/test` blocks counted, consistent with the "+40 tests" claim). `tsc --noEmit` / audit gates likewise not re-run.
- **No runtime exercise:** no live Telegram round-trip, no actual `--resume` replay (context restoration is the provider's job, per the PR).
- **Stated intent:** PR #562 title+body — spec-compliance assessed.

## Merge decision

# ✅ MERGE

No blocking, critical, or high findings; all security/correctness/api-compat clearances independently confirmed at the ref; spec-compliant and back-compatible. The two **medium** items (unguarded `close()` in `switchToSession`; redundant full-store rescan for the session name) are non-blocking quality improvements — worth folding into this PR or a fast follow-up, but neither breaks the feature or its back-compat guarantees. Finding #1 extends a pre-existing pattern shared by three sibling methods, so a small targeted fix (or a separate cleanup addressing all four call sites) is the cleanest path.

*Read-only review — no files, commits, or PR state modified.*

---

griffinwork40:
I now have citation-grounded ground truth across all dimensions. No `critical`/`high` findings surfaced, so per the skill no `/shadow-verify` pass is required (medium/low go straight through). Emitting synthesis.

---

# Code Review — PR #562 `feat(telegram): multi-session support`

**Target:** PR #562 `feat/telegram-session-registry` → `main` · reviewed ref `cddd478` · OPEN, mergeable
**Regime:** full (feature, 15 files, +1718/-3) · **Stated intent:** PR title+body — spec-compliance assessed
**Dimensions:** security · api-compat · correctness · spec-compliance · test-coverage · perf-observability

### Verdict by dimension
| Dimension | Verdict |
|---|---|
| security | **clean** — cross-chat switch gate, callback-prefix isolation, allowlist gate, path-traversal guard all verified at ref |
| api-compat | **clean** — all additive; `StoredSession.{handleId,bindings}` optional & back-compat; no signature changes |
| correctness | 1 medium + 1 low edge (resume-drop on create failure; in-flight switch race) |
| spec-compliance | **compliant** — unwired registry/route are intentional tested foundation, exactly as stated; deferred steps 6/7 correctly deferred; no scope creep |
| test-coverage | **strong** (+40 tests) — one gap paired with F1 |
| perf-observability | **clean** — new `loadSession` `console.warn` is a net audit-trail gain; switcher work is off the message hot path |

---

### Findings (sorted; none blocking)

**F1 · medium · confidence medium · correctness · `src/telegram/session-manager.ts:226-240` · ref:cddd478 · file-state**
`getSession()` deletes the staged `pendingResume` *before* the fallible `createSession()` call, so a `createSession` failure during a `/switch` permanently drops the staged resume — the next message then starts a fresh (non-resumed) conversation instead of continuing the switch target.
```js
const resumeTarget = this.pendingResume.get(chatId);
if (resumeTarget !== undefined) {
  this.pendingResume.delete(chatId);   // consumed before createSession can throw
  // …Object.assign(config, resumeConfigFor(...))
}
const session = await this.options.createSession(injectHotMemory(config)); // throws → resume already gone
```
*Suggestion:* move `this.pendingResume.delete(chatId)` to after `createSession` resolves (delete once `this.sessions.set` runs), making the staged resume retry-safe. Not blocking: no data is destroyed — the target sidecar persists and stays listed by `/sessions`; this is an error-path resilience gap.

**F2 · low · confidence high · test-coverage · `src/telegram/session-manager.test.ts` · ref:cddd478 · file-state**
No test exercises the F1 path: the two creation-failure tests — inflight rejection (`:108`) and rejecting-`close` during switch (`:913`) — never stage a switch-resume *before* a failing `createSession`, so the resume-drop is uncovered. (grep for createSession-rejection tests returns only those two.)
*Suggestion:* add `recordTelegramTurn(target)` → `switchToSession` → first `getSession` throws → assert the *second* `getSession` still passes `config.resume === target`.

**F3 · low · confidence medium · correctness · `src/telegram/session-manager.ts:170-186 / 653-688` · ref:cddd478 · file-state**
`switchToSession` has no guard against an in-flight `getSession` creation: `isBusy` reads `getSessionIfExists` (`this.sessions.get`), which is `undefined` during creation, so a switch tapped mid-creation is allowed; the in-flight promise then does `this.sessions.set(chatId, <pre-switch session>)` after `switchToSession` already set `data.sessionId = target` + `pendingResume = target`, so the chat keeps talking to the old conversation and the switch silently no-ops until the next teardown.
```js
// switchToSession deletes this.sessions but not this.pendingSessions:
const old = this.sessions.get(chatId); if (old) { await old.close()...; this.sessions.delete(chatId); }
```
*Suggestion:* in `switchToSession`, also await/drop any `this.pendingSessions` entry for the chat before adopting target state — or have `isBusy` treat an in-flight creation as busy. Narrow: requires a tap inside the ~ms creation window.

---

### Strengths (worth noting for the author)
- **Registry design is sound and well-tested:** `HandleId` branding, UUID-minted stable ids, `sdkSessionId` never used as a routing key (only `bySdk` reverse-lookup), `archive()` frees keys *and* `resolve()` re-guards `status==='active'` (belt-and-braces), `load()` only indexes active handles. Verified against `session-registry.ts:154-321`.
- **Back-compat bridge is correct:** old sidecars parse with `handleId/bindings === undefined`; `bindingsForStored` synthesizes the legacy `telegramChatId → telegram` binding keyed `String(chatId)`, matching `route.ts` `routeKey` (General → `String(chatId)`). Pure, round-trippable, unwired.
- **Route leak-fix is real and tested:** `routeFromCtx` reads `message_thread_id` off `message ?? editedMessage ?? callbackQuery.message` (`route.ts:66`), directly covered by `route.test.ts:41-42`.
- **Security posture:** `switchToSession` rejects cross-chat targets (`stored.telegramChatId !== chatId → not-found`, tested `:890`); `afk:sw:` prefix is non-colliding with all other `afk:*` callbacks; allowlist gate mirrors the existing `afk:m:` path.

### What was not checked (epistemic scope)
- Read against branch HEAD **`cddd478`** via `git show <sha>:<path>` — all citations verified at that ref (branch not checked out locally; `git show` used throughout).
- **Stated intent:** PR #562 title+body — spec-compliance assessed; deferred steps 6/7 correctly not flagged.
- Did **not run the test suite** (read-only review); PR claims 11,645 passed / 14 skipped — unverified here.
- The **sidecar-misattribution tail** of F1 (whether the post-failure fresh session's turns append to the *target's* sidecar via `_statsFor` hydration) is **unverified** — `_statsFor` internals were not fully read; the finding is scoped to the verified claim (resume silently dropped).
- `src/telegram/bot.ts`, `formatter.ts`, `telegram.ts`, `registration.ts` UI/wiring changes were verified by the security/api-compat agent's file-state reads (allowlist gate, prefix registration) but not independently re-read by me.
- **Process note (for async review):** the correctness agent timed out at 20min and the retry compose-nodes hit their tool-call budget (over-spelunking); correctness was therefore completed by direct file reads at the reviewed ref — findings above are citation-grounded, not agent-summarized.

## MERGE ✅

No blocking findings. Security and api-compat are clean; the change is additive with byte-identical existing behavior; spec claims (including the intentionally-unwired foundation and deferred steps) hold up under file-read verification; test coverage is strong. The three findings are non-blocking edge/error-path items — **F1** (resume dropped on `createSession` failure) is the one worth a quick follow-up (one-line fix: delete `pendingResume` after `createSession` resolves), ideally with **F2**'s regression test; **F3** is a narrow in-flight race. None require holding the merge.

---

PR #574, opened 2026-07-13
fix(readonly-bash): stop over-blocking recon that quotes a git/gh/curl verb

griffinwork40:
## What

The read-only bash classifier (`classifyBashCommand`, used to gate recon sub-agents like `research-agent` and `git-investigator`) over-blocks legitimate read-only commands that merely **mention** a git/gh/curl/pkg/sed verb inside a quoted string. This moves those rules from `RAW_RULES` (matched against the raw command) to `STRIPPED_RULES` (matched against a quote-stripped view) — the exact treatment already applied to the FS / pipe-to-shell / tar rules.

## Why

`classifyBashCommand` runs two rule views: `RAW_RULES` see the raw string; `STRIPPED_RULES` see a view with quoted data removed (`stripDataStrings`). The git/gh/curl/pkg/sed rules were in `RAW_RULES`, so a command that only *references* a verb inside quotes matched as if it were a real invocation:

| Command | Before | After |
|---|---|---|
| `rg "git push"` | ❌ blocked | ✅ allowed |
| `echo "remember to git push later"` | ❌ blocked | ✅ allowed |
| `grep -rn "git commit" .` | ❌ blocked | ✅ allowed |
| `grep -rn "npm install" .` | ❌ blocked | ✅ allowed |
| `rg "sed -i" src` | ❌ blocked | ✅ allowed |
| `grep -rn "curl -X POST" .` | ❌ blocked | ✅ allowed |

A command name/verb/flag is never legitimately quoted in a real invocation, so stripping quoted data lets recon search terms through while a genuine call — whose verb survives stripping — is still caught. This most affected `git-investigator`, whose core job (pickaxe / message search of git history) routinely quotes git-verb strings.

## How

- Moved `GIT_*_MUTATING`, `GIT_CONFIG_WRITE_FLAG`, `GH_*`, `SED/PERL_INPLACE`, `PKG_INSTALL`, `CURL_*` from `RAW_RULES` → `STRIPPED_RULES`.
- **Kept** `GIT_CONFIG_SET` and `FIND_EXEC_MUTATING` in `RAW_RULES`: their mutation signal — a value (`git config user.name "Foo"`) / an `-exec` verb (`-exec 'rm'`) — can itself be legitimately quoted, so quote-stripping would hide a real write.
- Pass 2 now bases off the `stash@{N}`-neutralized string (`rawForRules`) so `GIT_STASH_MUTATING` keeps working in its new home (otherwise `git stash show stash@{0}` backtracks into a false positive).
- Updated the module header + rule-array comments to describe the new split.

## Tests

- `readonly-bash.test.ts`: **185 passed**, including a new regression block — 9 quoted-verb mentions now allowed, plus real unquoted invocations, quoted-argument invocations (`git commit -m "msg"`, `sed -i "s/a/b/" f`), and `echo "$(git push)"` command substitution still **blocked**.
- Consumers `dispatcher.test.ts` (96) + `plan-mode-gate.test.ts` (16): pass.
- `tsc --noEmit` (strict): clean.

## Known residual (follow-up)

A **bare** unquoted pickaxe/grep value that collides with a git verb — `git log -S push`, `git log --grep push` — is still over-blocked (the arg is unquoted and matches the verb alternation). Fixing it needs subcommand-position anchoring (plus, for the under-restriction side, wrapper/indirection handling). This is documented inline and tracked as a separate follow-up issue.

## Refs

- Discovered in the 2026-07-13 sandboxing audit (finding **O1**, over-restriction lens). Verified independently via a shadow-verify pass before implementation.

### DIFF
--- a/src/agent/tools/readonly-bash.test.ts
+++ b/src/agent/tools/readonly-bash.test.ts
@@ -164,6 +164,55 @@ describe('classifyBashCommand', () => {
     });
   });
 
+  // ── O1 (MED-HIGH): git/gh/curl/pkg verbs inside quoted args must not block ──
+  // Tier-1 fix regression guard: the git/gh/curl/pkg/sed rules were moved from
+  // RAW_RULES to STRIPPED_RULES so a recon command that merely MENTIONS a
+  // mutating verb inside a quoted string is not misread as an invocation. Every
+  // `allowed` case below was BLOCKED before the move (a false positive that hurt
+  // git-history recon in `git-investigator`); every `blocked` case proves a real
+  // (unquoted) invocation — and command substitution — is still caught.
+  describe('git/gh/curl/pkg verbs inside quoted args are not over-blocked', () => {
+    const allowed: string[] = [
+      'rg "git push"',
+      'grep -rn "git commit" .',
+      'echo "remember to git push later"',
+      "grep -rn 'git reset --hard' .",
+      'rg "gh pr create" src',
+      'grep -rn "npm install" .',
+      'rg "sed -i" src',
+      'grep -rn "curl -X POST" .',
+      'git log -S "push"', // quoted pickaxe value — searching history, not pushing
+    ];
+    for (const cmd of allowed) {
+      it(`allows: ${cmd}`, () => {
+        expect(classifyBashCommand(cmd).mutating, `expected "${cmd}" read-only`).toBe(false);
+      });
+    }
+
+    // Regression: real (unquoted) invocations MUST still be caught after the move.
+    // A quoted ARGUMENT (commit message, sed script) does not hide the real verb;
+    // a `$(…)` command substitution is preserved (not stripped) so it is caught.
+    const blocked: Array<[string, string]> = [
+      ['git push origin main', 'real git push'],
+      ['git commit -m "msg"', 'real git commit with quoted message'],
+      ['gh pr create -t t -b b', 'real gh pr create'],
+      ['npm install left-pad', 'real npm install'],
+      ['sed -i "s/a/b/" f', 'real sed -i with quoted script'],
+      ['curl -X POST https://x -d @body', 'real curl POST'],
+      ['echo "$(git push)"', 'git push inside command substitution'],
+    ];
+    for (const [cmd, label] of blocked) {
+      it(`blocks: ${label} (${cmd})`, () => {
+        expect(classifyBashCommand(cmd).mutating, `expected "${cmd}" mutating`).toBe(true);
+      });
+    }
+
+    // Known residual (Tier 2, not fixed here): a BARE unquoted pickaxe/grep value
+    // that collides with a git verb — `git log -S push`, `git log --grep push` —
+    // is still over-blocked, because the arg is unquoted and matches the verb
+    // alternation. Fixing it needs subcommand-position anchoring; tracked separately.
+  });
+
   // ── Finding 2 (MED): git config read forms must pass ──────────────────────
   describe('git config read forms are not over-blocked', () => {
     const allowed: string[] = [
--- a/src/agent/tools/readonly-bash.ts
+++ b/src/agent/tools/readonly-bash.ts
@@ -22,14 +22,23 @@
  * SKILL.md prose.
  *
  * Classification runs in three passes (see `classifyBashCommand`):
- *   - Pass 1 (RAW_RULES): rules that must see the FULL command string —
- *     flag-bearing git/gh/curl/pkg forms. Pass 1b then captures an interpreter
- *     (`python -c`/`node -e`) quoted payload and tests it for a write API.
- *   - Pass 2 (STRIPPED_RULES): rules matching short generic tokens (`cp`, `mv`,
- *     `find … -delete`, command-position `patch`/`install`) run against a
- *     DATA-string-stripped view, so a quoted search term like `grep -rn "cp " .`
- *     is not misread as a `cp` invocation. Command substitutions (`$(…)`,
- *     backticks) are PRESERVED so a real `echo "$(rm -rf x)"` is still caught.
+ *   - Pass 1 (RAW_RULES): the ONLY rules that must see the raw string are the
+ *     ones whose mutation SIGNAL can itself be legitimately quoted — `git config
+ *     <key> <value>` (the value is often quoted) and `find … -exec <verb>` (the
+ *     verb may be quoted). Quote-stripping those would hide a real write. Pass 1b
+ *     then captures an interpreter (`python -c`/`node -e`) quoted payload and
+ *     tests it for a write API.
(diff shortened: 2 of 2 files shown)


griffinwork40:
The known residual noted above (bare `git log -S push` / `--grep push` over-block, plus the under-restriction wrapper/indirection cases) is tracked as a follow-up in #577.

griffinwork40:
Confirmed independently: exactly two production callers — `dispatcher.ts:472` (`if (!verdict.mutating) return null`) and `plan-mode-gate.ts:90` (`if (verdict.mutating)`). Both gate on the `.mutating` boolean only; the other grep hits are comments/imports. The api-compat "no break" verdict holds.

I have full coverage across all dimensions. No critical/high findings surfaced → `/shadow-verify` is not required by the skill's gate, and I've already independently re-derived the load-bearing claims inline (regex tracing, caller check, test-coverage inventory). Synthesizing the final report.

---

# Code Review — PR #574

**`fix(readonly-bash): stop over-blocking recon that quotes a git/gh/curl verb`**
`fix/readonly-bash-quoted-verb-overblock` → `main` · reviewed ref `42123a8` · +115/−38, 2 files · change type: hotfix · regime: **full** (elevated from light — security-gate code)

## Summary
Moves the git/gh/curl/pkg/sed mutation-detection rules from `RAW_RULES` (Pass 1, raw string) to `STRIPPED_RULES` (Pass 2, quote-stripped view), so a read-only recon command that merely *mentions* a mutating verb inside quotes (`rg "git push"`) is no longer misread as a real invocation. Keeps `GIT_CONFIG_SET` + `FIND_EXEC_MUTATING` raw (their signal is legitimately quotable), and rebases Pass 2 off the `stash@{N}`-neutralized string so `GIT_STASH_MUTATING` keeps working in its new home.

## Findings

**None blocking. No critical / high / medium.**

`low · high-confidence · test-coverage · readonly-bash.test.post.ts (whole file) · ref:42123a8 · file-state` — `PERL_INPLACE`, `GH_API_WRITE_METHOD`, `GH_API_FIELD`, and `PKG_INSTALL`'s `pip` verb have **zero** tests (0 matches for `perl`, `gh api`, `pip` in both baseline and post-PR suites); the move subjects these rules to the quote-strip transform with no regression guard for either their real-invocation blocking or quoted-mention passing.
evidence: `grep -c -i perl → 0` and `grep -c 'gh api' → 0` against both test files (gap is pre-existing, not introduced here).
suggestion: add one `allows: rg "perl -i"` + one `blocks: perl -i -pe 's/a/b/' f` case, likewise for `gh api -X POST /x`.

`nit · high-confidence · spec-compliance · readonly-bash.test.post.ts:216-219 · ref:42123a8 · diff-context` — the "Known residual" note documents only the bare-pickaxe case (`git log -S push`); it omits the second residual that keeping `GIT_CONFIG_SET`/`FIND_EXEC_MUTATING` raw creates — a quoted recon mention like `rg "git config user.email x"` or `grep -rn "find . -exec rm" .` stays over-blocked.
evidence: `const blocked ... ` retains `GIT_CONFIG_SET` in RAW_RULES (readonly-bash.post.ts:104-105) matched against the raw string.
suggestion: one extra line in the residual comment noting the config-set / find-exec quoted-mention case is intentionally still over-blocked.

`informational · security · readonly-bash.post.ts:299-300 · ref:42123a8 · file-state` — bare verb/flag quoting (`git "push"`, `git branch "-d" x`) now evades the moved rules post-strip. This is the **pre-existing, accepted** `STRIPPED_RULES` property (already true for `rm`/`cp`/`mv`) and explicitly out of the module's stated threat model — *"NOT a security boundary… not to sandbox a hostile process"* (readonly-bash.post.ts:5-6,17-20). No action.

## Dimension results (all read at ref `42123a8`)
- **Security** — no regression. Every conventional/quoted-argument/command-substitution invocation traced still BLOCKED: `git commit -m "msg"`, `sed -i "s/a/b/" f`, `curl -X POST … -d @body`, `echo "$(git push)"` (test lines 196-202). Both RAW_RULES retentions are load-bearing, not redundant.
- **API-compat** — no break. Signature `(command: string) => { mutating: boolean; reason?: string }` unchanged; exactly 2 production callers (`dispatcher.ts:472`, `plan-mode-gate.ts:90`), both gate on `.mutating`, treat `.reason` as free display text → rule reorder is safe.
- **Correctness** — no real-invocation regression. Change-B (`rawForRules` rebasing) correctly fixes the `git stash show stash@{0}` false positive (guarded, test lines 117-132) and cannot hide a verb (`stash@{N}` is always an argument).
- **Spec-compliance** — intent fully met; all 6 over-blocking cases from the PR table now pass (test lines 176-181). No scope creep — diff is rule-move + Pass-2 rebase + comment updates only.
- **Test-coverage** — strong for the fix (9 quoted-mention allows, 7 real-invocation blocks incl. command-substitution); pre-existing gaps noted above.
- **Perf/observability** — unchanged (same rule count, one extra `.replace`; no backtracking-risk change; `reason` strings identical).

## What was not checked
- Wave 1 agents read against **ref `42123a8`** via byte-identical materialized post-PR files — citations verified at that ref.
- The correctness·spec·test sub-agent **timed out at 20 min**; that dimension was performed **inline** by me with direct git/grep access (stated per the delegation-failure protocol).
- **Test suite was NOT executed** — the read-only constraint forbids checkout/apply, and running the PR's tests would require materializing the branch into a worktree. The author's claims (185 passed, `tsc --noEmit` clean) were **not independently run**; I verified test *presence* and traced regex behavior statically instead.
- Stated intent: PR #574 title+body — spec-compliance assessed.
- Did not review the `nesting.ts` Layer-1 allowlist (unchanged by this diff) beyond confirming it is the structural boundary.

## Merge decision: **MERGE**
A clean, well-reasoned, tightly-scoped fix consistent with the classifier's documented threat model. The three findings are all non-blocking (one low test-coverage gap that is pre-existing, one doc nit, one informational note). Recommend addressing the `perl`/`gh api` test gap in a follow-up, not as a merge blocker.

**Done**
- Reviewed PR #574 in full regime (2-agent wave; correctness agent timed out → that dimension done inline), across security, api-compat, correctness, spec-compliance, test-coverage, and perf.
- Evidence: security·api-compat sub-agent report (clean, verified); inline verification via `git show 42123a8`, grep of callers (`dispatcher.ts:472`, `plan-mode-gate.ts:90`), and test-coverage inventory of the materialized post-PR files at `/Users/griffinlong/.afk/state/skill-preflight/d1e86d15-f357-4691-b75b-cef6a24b9d67/`.
- Verdict: **MERGE** — no blocking findings; 1 low (pre-existing perl/gh-api/pip test gap), 1 nit (residual-doc completeness), 1 informational (bare-verb-quote evasion, out of threat model).
- World unchanged: read-only review, no files/PR/git mutated. Scratch materializations (`readonly-bash.post.ts`, `readonly-bash.test.post.ts`) live in the preflight temp dir only.

---

PR #584, opened 2026-07-13
feat(anthropic): bound time-to-first-byte with a configurable per-request timeout (#583)

### LINKED ISSUE #583: perf/reliability: no per-request timeout lets a single slow model call hang ~10min (ttfb p99 85s, max 748s)
## Summary

afk sets **no per-request timeout** on the Anthropic client, so a single genuinely-slow model response can hang an interactive turn for **up to the SDK default (~10 min)** before giving up. Measured `model_ttfb` (time-to-first-byte per round-trip) has a median of 4.3s but a **p99 of 85s and a max of 748s** (12.5 min). Adding an interactive per-request timeout (and, secondarily, de-preferring empirically-slow model endpoints) would cap this tail.

## Measured evidence (2026-07-13)

From 1,500 real witness traces → **45,811 `model_ttfb` samples**:

| | median | p90 | p99 | max |
|---|--:|--:|--:|--:|
| `model_ttfb` | 4282ms | 9733ms | **85309ms** | **748342ms** |

Concentrated by model — this is not uniform slowness:

| model | share of round-trips >30s |
|---|--:|
| `claude-sonnet-5` | **15.5%** (317/2038) |
| `claude-opus-4-8` | 0.12% (10/8029) |

## Root cause (diagnosed, trace-proven)

**The tail is genuinely-slow single upstream calls, NOT afk retry/backoff.**

- Backoff-inflation refuted: of `ttfb` >30s, **555/556 (99.8%)** had **no** preceding `rate_limit` trace event; of `ttfb` >120s, **51/51 (100%)** had none.
- The 748s max (session `345bf28c`, sonnet-5) has **zero** `rate_limit` events in the whole session; its `ttfb` climbs 47s → 72s → 748s across consecutive rounds — a degrading real call, not a throttle.
- afk's own retry paths are bounded far below the tail: mid-stream overload backoff `OVERLOAD_BASE_DELAY_MS`=5s × `OVERLOAD_MAX_RETRIES`=3 (`src/agent/providers/anthropic-direct/loop.ts:88-89,383`); transient retry-after clamped at `RATE_LIMIT_RETRY_MAX_WAIT_MS`=120s (`src/agent/providers/anthropic-direct/query/retry-layer.ts:61`).

**The afk-controllable lever: no per-request timeout.**
`buildClientOptions` (`src/agent/providers/anthropic-direct/auth.ts:68-84`) passes only `{ authToken | apiKey, baseURL?, fetch? }` to `new Anthropic(...)` (`index.ts:617`, `oneshot.ts:63`) — **no `timeout`, no `maxRetries`**. So afk inherits the SDK defaults (request timeout ~10 min; maxRetries 2), and a degrading call is allowed to hang for minutes. afk chooses not to bound it.

## Proposed fix (design decision required — do NOT blind-patch)

1. **Set an interactive per-request timeout** on the Anthropic client / `messages.create()` call, well below the ~10-min SDK default.
   - **Tradeoff / caution:** the bound must not abort *legitimately*-slow valid responses (opus_1m prefill, long extended-thinking). Choose it from the distribution: p99 ≈ 85s, so a cap around **120–180s** trims the extreme tail while sparing normal slow calls.
   - Streaming nuance: the timeout should apply to time-to-first-byte / stall detection, not total stream duration (a long-but-progressing stream is fine).
2. **Model-aware routing (secondary):** surface/de-prefer or warn on endpoints that are empirically slow (sonnet-5 here is ~130× worse than opus). Product-level, can be a follow-up.

**Explicitly out of scope (upstream-inherent, not afk-fixable):** the 4.3s median, and sonnet-5's server-side slowness itself.

## Open design questions

- Per-request `timeout` on the client vs per-call override? Interactive vs headless/daemon may want different bounds.
- On timeout: fail the turn, or retry once (possibly rerouting model)? Interaction with the existing `retry-layer` / mid-stream overload retry must be defined so we don't stack waits.
- Should the timeout be configurable via env (e.g. `AFK_MODEL_REQUEST_TIMEOUT_MS`)?

## Acceptance / verification

- A stalled/degrading request is aborted at the configured bound instead of hanging ~10 min; surfaced cleanly to the user (not a silent freeze).
- Legitimate slow-but-progressing responses (large opus_1m prefill, long thinking) are NOT falsely aborted.
- Unit test: a mocked client whose stream stalls past the bound → afk aborts at the timeout (not the 10-min default).
- Re-mine traces after rollout: `model_ttfb` p99/max should drop toward the cap.

## Related / notes

- Sibling perf issue from the same audit: #581 (jsdom cold-start).
- **Separate reliability finding (not this issue):** 161/191 `rate_limit` events in the sample are OAuth **subscription-exhaustion** 429s (median retry-after ~3.7h) that abort the turn — usage-limit hits, thresholded by `RATE_LIMIT_TRANSIENT_MAX_RETRY_AFTER_MS`=5min (`src/agent/providers/anthropic-direct/usage-limit.ts:42`). Worth its own issue if the abort UX needs work.
- Full diagnosis + methodology: `~/.afk/workspace/reports/perf-baseline-2026-07-13.md`.

griffinwork40:
Closes #583.

## What changed

afk set **no per-request timeout** on the Anthropic client, so a single genuinely-slow model response could hang an interactive turn up to the SDK default (~10 min). Measured `model_ttfb`: median 4.3s but **p99 ≈ 85s, max 748s**. Root cause is confirmed genuinely-slow single upstream calls (99.8% of >30s ttfb had no preceding `rate_limit`), not afk retry/backoff.

This bounds a single request's **time-to-first-byte** in afk's own streaming loop so a degrading call fails fast (with one retry) instead of hanging.

**Files:**
- `src/agent/providers/shared/first-byte-timeout.ts` (new) — pure TTFB helper: `resolveTtfbTimeoutMs()` (env parse) + `armFirstByteTimeout(baseSignal, timeoutMs)` returning a linked-signal/timer handle. Lives in `shared/` so the openai-compatible loop can reuse it.
- `src/agent/providers/anthropic-direct/loop.ts` — arms the timer per round; cancels it on the first non-error streamed event; routes a TTFB timeout (connection-phase **and** post-headers/in-band) into a single re-drive.
- `src/config/env.ts` — registers `AFK_MODEL_TTFB_TIMEOUT_MS` in `ENV_REGISTRY` with a typed lazy getter (read only via `env`).
- `docs/env-registry.{json,md}` — regenerated via `pnpm scan:env`.
- `src/agent/providers/shared/first-byte-timeout.test.ts`, `src/agent/providers/anthropic-direct/loop.ttfb.test.ts` (new tests).

## Design decision

**Mechanism — afk's own loop, not the SDK `timeout` option.** A per-request `AbortController` is chained to the caller's turn signal via `AbortSignal.any`, plus a `setTimeout(bound).unref()` that aborts *only that request* if no first byte arrives. The timer is cancelled the instant the first **non-error** translated event arrives (`ttfb.firstByteSeen()`), so a legitimately slow-but-progressing stream (long opus_1m prefill, extended thinking) is **never** aborted — only a call that streams no first content within the bound trips it. A TTFB abort never mutates the caller's own signal, so it stays distinguishable from a user interrupt.

*Why not the SDK `timeout`:* verified against `@anthropic-ai/sdk` 0.74.0 — the SDK **retries timeouts by default** (`shouldRetry` -> "Retry on request timeouts", `maxRetries=2`), so its worst case is `timeout × (retries+1)` (the docs literally warn "you may wait much longer than this timeout"), which is exactly the stacking the issue forbids. It also fires on response *headers*, not first content, whereas afk's `model_ttfb` metric measures to first content. The loop approach matches the metric and composes with afk's existing retry layers instead of fighting them.

**Default value.** `AFK_MODEL_TTFB_TIMEOUT_MS=180000` (180s ≈ 2× the measured p99 of 85s) — legitimate calls essentially never hit it while the 300–748s pathology is capped. `0` disables the timeout (explicit escape hatch); unset/invalid falls back to the default.

**On timeout — retry once, then fail fast.** The request is re-driven **once per round**, reusing the mid-stream overload retry's signalling: a `rate_limit` trace phase (`reason: "ttfb-timeout"`, legible in `afk trace show`) plus a `stream.retry` event so surfaces discard any partial paint. No backoff sleep (the point is to fail fast off a stalled endpoint). If the retry also stalls, a clean `error` event surfaces — never a silent hang.

**No stacking.** The single retry is gated by a per-round `ttfbRetried` flag, reset only after a clean first byte (mirroring the existing per-round `overloadRetries` reset). Worst case for one request is **one bound + one retry = 2× bound**; it cannot compound with the overload backoff (a TTFB timeout means zero bytes arrived, so the overload path isn't in play).

## Test evidence

```
pnpm lint                          -> PASS (tsc --noEmit, clean)
pnpm audit:env:check               -> PASS (654 files, 0 violations)
pnpm scan:env:check                -> PASS (docs in sync, 122 vars)
pnpm audit:sdk:check               -> PASS (lock matches; no new SDK import symbols added)

vitest src/agent/providers/shared/first-byte-timeout.test.ts   -> 11 passed
vitest src/agent/providers/anthropic-direct/loop.ttfb.test.ts  ->  5 passed
vitest src/agent/providers/anthropic-direct/  (whole provider) -> 388 passed (34 files)
pnpm test  (full suite)            -> 639 files, 11667 passed | 14 skipped, 0 failed
```

New-behaviour tests assert: a stalled stream (both connection-phase and post-headers) aborts at the configured bound and retries once (not the 10-min default); the retry-also-stalls case surfaces an `error` with exactly one retry and no third attempt (no stacking); a stream that yields a first byte within the bound is unaffected even when it then runs long (600s mid-stream gap); and `AFK_MODEL_TTFB_TIMEOUT_MS=0` disables the timer entirely. No pre-existing failures were observed or excluded.

## Follow-ups

- **openai-compatible mirror (deferred):** the parallel provider has the same gap but a different streaming skeleton (`query/stream-drive.ts` / `driveStream`, with its own `model_ttfb` emission), so mirroring is not a trivial copy. The new `providers/shared/first-byte-timeout.ts` helper is deliberately provider-neutral and ready to wire in there.
- **Model-aware routing** (de-preferring empirically-slow endpoints, e.g. sonnet-5) — explicitly **out of scope** per the issue; product-level follow-up.
- The `oneshot.ts` non-streaming completion path (compact/suggest summarizers) is not covered here — it is not an interactive turn and not the issue's subject.


griffinwork40:
Cleanup complete — main working tree untouched (still `main`, clean), all review branches/worktree removed.

---

# PR #584 Review — TTFB per-request timeout for anthropic-direct

**Verdict: ✅ APPROVE / merge-ready.** No blocking or high-severity issues. This is a careful, well-tested change to a high-risk area (the streaming loop). All findings below are low/nit-level.

## What I verified (ground truth, on the fetched PR head `642f333`)

| Check | Result |
|---|---|
| `pnpm lint` (`tsc --noEmit`, maximally strict) | ✅ clean |
| Full test suite | ✅ **11667 passed / 0 failed** / 14 skipped (639 files) |
| New TTFB tests in isolation | ✅ 16 passed (11 helper + 5 loop) |
| `pnpm audit:env:check` | ✅ 0 violations |
| `pnpm scan:env:check` | ✅ docs in sync (122 vars) |
| Independent control-flow re-derivation (read-only subagent, no code comments trusted) | ✅ all 5 load-bearing claims CONFIRMED |

## Correctness (the load-bearing dimension) — solid

Independently re-derived and confirmed against `loop.ts` + `translate.ts`:

- **Bounded retries / no infinite loop.** Every `continue` is gated by a monotonic counter that resets *only on genuine forward progress* (a round that streamed a first byte). `ttfbRetried` caps TTFB re-drives at 1; `overloadRetries < 3` caps mid-stream overload; `capReached` is a hard stop. Worst case is a finite create-count then `error`+`return`. Termination is provable.
- **Signal isolation is correct.** The TTFB timer aborts only a private per-request `AbortController` chained via `AbortSignal.any([input.signal, controller.signal])` — the caller's `input.signal` is **never** mutated. A user interrupt (`input.signal.aborted`) and a TTFB stall (`ttfb.timedOut()`) are independent state, so all four retry branches classify correctly; no misclassification path found.
- **Type-safe first-byte guard.** `isFirstByte = !ttfbEmitted && (out.kind === 'turn-result' || out.event.type !== 'error')` — the `||` short-circuit guarantees `.event` is never dereferenced on a turn-result output (discriminant literals verified in `translate.ts:33,307`). An in-band error event is correctly *not* treated as a first byte, so the timer stays "timed out" and the retry fires; after `firstByteSeen()` disposes the timer, `timedOut()` can never flip true again → retry branches unreachable once real content streams.
- **No missing-`dispose()` path.** `ttfb.dispose()` (clearTimeout, idempotent) is reached on every exit: connection retry, connection error/abort, first-byte, catch retry, catch error/abort, and unconditionally after stream consumption. The `.unref()` on the timer is a correct backstop.

## Security / API-compat / Observability — clean

- **Security:** only new input is a numeric env var, parsed with `Number.parseInt` + validated (`finite && >= 0`, else default). No injection/path/credential surface. The marker string is a constant.
- **API-compat:** `createWithRetry` is module-private, so its 4→5 param change breaks nothing external. All new module exports are additive. Behavior change (stalled request now bounded at 180s + 1 retry instead of the SDK's ~10-min hang) is the intended fix and is opt-out via `AFK_MODEL_TTFB_TIMEOUT_MS=0`.
- **Observability:** the re-drive is legible in `afk trace show` via a `rate_limit` phase (`reason: 'ttfb-timeout'`) + a `stream.retry` event that clears partial paint — mirrors the existing overload path.

## Low-severity notes / nits (non-blocking, author's discretion)

1. **`isTtfbTimeoutError` is exported + tested but not called in runtime** (`first-byte-timeout.ts:45`). Intentional per the module doc (reserved for the future openai-compatible path), but currently dead-in-runtime. Fine to keep; flagging so it's a conscious choice.
2. **TTFB is not a hard wall while a 529/503 backoff is in flight.** The `createWithRetry` backoff `sleepWithAbort` now gates on `turnSignal`, so a TTFB fire mid-backoff waits out the current sleep (≤20s) before surfacing as a retry. Documented and bounded — but anyone reading "180s bound" might expect a hard ceiling. The env-var help wording ("stall BEFORE its first streamed event") is accurate; no change strictly needed.
3. **`model_ttfb` trace phase behavior shifted subtly:** it previously fired on the first translated output *of any kind* (including an error); now it fires only on the first non-error output. Arguably an improvement (an error isn't a "first byte"), but it's an unadvertised observability change — a round whose first output is an error no longer emits `model_ttfb`.
4. **`AbortSignal.any` listener detach is GC-timed:** `dispose()` only clears the timer; the composite's listener on the long-lived `input.signal` is released on GC, not eagerly. Not a leak in practice (Node ≥22 weak-refs these), but on a turn with hundreds of tool rounds, detachment depends on GC cadence.

## Scope note (not a defect)
The helper lives in `shared/` and its doc promises the "analogous openai-compatible streaming path can reuse the exact same mechanism," but this PR wires it only into `anthropic-direct`. The openai-compatible provider retains the unbounded-stall behavior — a reasonable scoped follow-up for #583, worth tracking.

---

**Done**
- Reviewed PR #584 across correctness, security, API-compat, test-coverage, and observability; verdict **APPROVE / merge-ready** with only low/nit findings.
- Evidence: fetched PR head `642f333` into an isolated worktree and ran real gates — `tsc --noEmit` clean, full suite **11667 passed / 0 failed**, new TTFB files 16/16, `audit:env:check` + `scan:env:check` both pass; an independent read-only subagent re-derived and CONFIRMED all five load-bearing correctness claims with file:line evidence.
- World change: temporary review worktree + branches (`afk/pr-584-review`, `pr-584-review`) created then removed; local ref `refs/pull/584/head` remains fetched. Main working tree untouched (`main`, clean).
- This review is analysis only — read-only per the skill contract: no PR comment posted, no files edited, nothing committed or pushed.

---

griffinwork40:
Verifications complete. All the load-bearing claims hold:

- **`createWithRetry` is module-private** (post-merge `loop.ts:130`, no `export`) — its new `turnSignal` param is an internal-only signature change → **no api-compat break**.
- **Re-drive re-sends unmutated history**: the `while (true)` loop rebuilds `params` each iteration (`loop.ts:274`); every `input.messages.push(...)` sits at `:630+`, *after* stream consumption, while both TTFB re-drive `continue`s fire at `:347`/`:476` — so no history mutation precedes a re-drive.
- **`resolveTtfbTimeoutMs` has no upper clamp** (diff L751-757) — confirmed the one real robustness gap.
- **User-interrupt never misclassified**: every TTFB guard ANDs `!input.signal.aborted`, and the helper test (diff L695-701) pins `timedOut()===false` on a caller abort.

No `critical`/`high` findings → `/shadow-verify` not required. Here is the review.

---

# Code Review — PR #584

**Target:** PR #584 `feat(anthropic): bound time-to-first-byte with a configurable per-request timeout (#583)` · `afk/model-request-timeout` → `main` · reviewed ref `642f3335`
**Triage:** feature · 7 files · ~701 lines · regime **full**. Reviewed surface: `loop.ts`, `env.ts`, 3 new files. Skipped: `docs/env-registry.{json,md}` (`pnpm scan:env`-generated).
**Stated intent:** PR title + body — spec-compliance **assessed**.

## Dimension verdicts
- **security** — `no issues found`. No auth/credential/header/secret changes; the new plumbing only adds a per-request `AbortController` chained via `AbortSignal.any([baseSignal, controller.signal])` (`first-byte-timeout.ts:803-804`) and never mutates the caller's own signal. No injection/deserialization surface.
- **api-compat** — `no issues found`. `createWithRetry` is not exported (`loop.ts:130`); the new helper exports are purely additive.
- **correctness** — 2 low/nit findings below; core mechanism verified against `translate.ts` and post-merge `loop.ts`.
- **spec-compliance** — `no issues found`. Every stated requirement is implemented; deferred items (openai-compatible mirror, model-aware routing, `oneshot.ts`) are explicitly disclosed, not silent gaps. No scope creep.
- **test-coverage** — `no issues found`. All five behavioral scenarios + helper unit tests present (`loop.ttfb.test.ts`, `first-byte-timeout.test.ts`).
- **perf-observability** — `no issues found`. `rate_limit` trace phase `reason:"ttfb-timeout"` + `stream.retry` emitted (`loop.ts:110-116`); negligible per-round overhead; timer `.unref()`d.

## Findings

**1 · low · confidence medium · correctness/robustness · `src/agent/providers/shared/first-byte-timeout.ts:751-757` · ref:642f3335 · citation-type:file-state**
`resolveTtfbTimeoutMs` returns the parsed integer with no upper bound, so `AFK_MODEL_TTFB_TIMEOUT_MS` set above Node's max timer delay (2 147 483 647 ms ≈ 24.8 days) overflows `setTimeout`; Node clamps an over-range delay to `1` (with a `TimeoutOverflowWarning`), so the stall timer fires ~immediately and trips the TTFB retry/error path on essentially every request.
```
  const n = Number.parseInt(raw, 10);
  if (!Number.isFinite(n) || n < 0) return DEFAULT_MODEL_TTFB_TIMEOUT_MS;
  return n;
```
Suggestion: clamp — `return Math.min(n, 2_147_483_647);`. Misconfiguration-only; the default (180000) and realistic overrides are unaffected.

**2 · nit · confidence high · api-surface · `src/agent/providers/shared/first-byte-timeout.ts:762-765` · ref:642f3335 · citation-type:file-state**
`isTtfbTimeoutError` (and `TTFB_TIMEOUT_MESSAGE`) are exported and unit-tested but not consumed by any production path — `loop.ts` imports only `armFirstByteTimeout, resolveTtfbTimeoutMs` (diff L50-53) and branches on `ttfb.timedOut()`. This is a deliberate provider-neutral API for the deferred openai-compatible wiring, not a defect.
```
export function isTtfbTimeoutError(err: unknown): boolean {
  return err instanceof Error && err.message === TTFB_TIMEOUT_MESSAGE;
}
```
Suggestion: leave as-is (reserved forward API) or add a one-line `@remarks reserved for the openai-compatible mirror` so it doesn't read as dead code.

## Summary
A tightly-scoped, well-tested feature. The subtle correctness hinges — (a) `message_start`/`ping`/text-block-start yield no translated event so the timer is cancelled only on the first *content* byte (`translate.ts:147-153,290-292` vs the `isFirstByte` gate at `loop.ts:371-373`); (b) an in-band `error` event is excluded from "first byte" so `timedOut()` stays meaningful; (c) the once-per-round `ttfbRetried` gate bounds worst case at 2× bound with no stacking; (d) re-drive re-sends unmutated history; (e) timers disposed on every path and `.unref()`d — all verified against the reviewed ref. No blocking issues; the two findings are a misconfiguration clamp and a cosmetic dead-export note.

**MERGE** — no blocking findings; the two low/nit items are optional follow-ups, not merge gates.

---

**Done**
- Reviewed PR #584 across all full-regime dimensions and emitted a **MERGE** recommendation with 2 non-blocking findings (1 low, 1 nit); no `critical`/`high`, so `/shadow-verify` was correctly skipped.
- Evidence: findings cite the reviewed ref `642f3335` via the diff artifact (`/Users/griffinlong/.afk/state/skill-preflight/8787ddaf-a2cc-42a9-ba4e-6614ba17060b/pr-584.diff`) and materialized post-merge copies under the same preflight dir (`postmerge/src/...`), all read this session; cross-checked against unchanged `src/agent/providers/anthropic-direct/translate.ts`.
- World change: none — read-only review, no files/PR mutated (working tree still clean on `main @ 7ce4a25`; the `postmerge/` copies are scratch materializations of git blobs, not repo changes).
- **What was not checked / caveats:** Wave 1 agents read from ref `642f3335` content (diff + post-merge blobs), not a live checkout, since the tree is on `main`. Both dispatched `research-agent`s **timed out at 20 min** (delegation failure) — this review was completed inline from my own file-grounded reads rather than their output; the trade-off is one reviewer perspective instead of two parallel ones, mitigated by direct reads of every load-bearing file. I did **not** run the test suite or CI gates myself — the PR reports them green (`pnpm lint`, `audit:env/sdk/scan`, 11667 tests passed); merging should still rely on CI. Finding #1's Node `setTimeout`-overflow behavior is from documented Node semantics, not an executed repro.

---

PR #545, opened 2026-07-12
feat(improve): detect forked sub-agent read auto-denials

griffinwork40:
## Problem

The improve-pipeline detector `subagent-block` filters `hookEvent === 'SubagentStart'` only (`src/improve/scan/detectors/subagent-block.ts:96`). It is therefore **structurally blind** to the path-approval **`PreToolUse`** hook auto-denying a forked sub-agent's **read** (`blockedTool ∈ {read_file, grep, glob, …}`, reason `"Sub-agent path access denied: <path> is outside …"`) — the failure mode where a read-only fork cannot read a path it needs, retries the denied read, and spins to a wall-clock timeout. The detector's own JSDoc admits `blockedTool` "is only set by PreToolUse blocks" yet it never processes them.

## Change

A dedicated **`subagent-read-denial`** detector (distinct from `subagent-block`: that flags a *dispatch* refused before start; this flags a *running fork's read* refused — different lifecycle, different remediation):

- **Read-only filter.** PreToolUse `decision:'block'` with a read-family `blockedTool`. **Write/mutating denials are excluded** — worktree write-confinement is by design and would drown the signal.
- **Path normalization.** Denial reasons embed the specific offending path, which varies per denial; the fingerprint hashes a normalized reason (absolute paths → `<path>`) so structurally-identical denials aggregate into one card instead of one-card-per-path.
- **Opt-in** (`enabledByDefault: false`), like `subagent-block` — read-denials can fire during legitimate confinement or active bug-fixing.

Wired end-to-end: `FailurePattern` schema, detector registry + `DetectorOptions`, propose template (`dispatcher-bug` class, points at the read-scope machinery), and a `--read-denial-min-occurrences` CLI flag.

## Files

| File | Change |
|---|---|
| `src/improve/scan/detectors/subagent-read-denial.ts` | **new** — detector + `normalizeReason` |
| `src/improve/scan/detectors/subagent-read-denial.test.ts` | **new** — 20 unit tests (incl. write-exclusion + normalization) |
| `src/improve/schemas.ts` | add `'subagent-read-denial'` to `FailurePatternSchema` |
| `src/improve/scan/detectors/index.ts` | registry entry + option |
| `src/improve/scan/detectors/index.test.ts` | registry count/name assertions + fixture |
| `src/improve/propose/template-engine.ts` | propose template for the new pattern |
| `src/cli/commands/improve.ts` | `--read-denial-min-occurrences` flag |

## Validation on real traces

```
[...]

PR #592, opened 2026-07-13
feat(subagent): fail-fast denial circuit breaker for forked read-denial spins (#546)

### LINKED ISSUE #546: fail-fast: forked sub-agent should abort on accumulated path-approval read denials instead of burning its 20-min wall-clock budget
## Summary

A forked sub-agent that accumulates repeated **path-approval read denials** keeps issuing new reads (all denied) until its 20-minute wall-clock budget expires, instead of **failing fast** on the denial signal. It should abort in seconds with an actionable message once it detects it is spinning on denials it cannot recover from.

## Background

The path-approval `PreToolUse` hook auto-denies a forked child's read when the path is outside the fork's granted read roots (a fork cannot prompt to approve). The child then retries with other reads, all denied, and makes no progress.

The most common *cause* of this — a fork granted a read scope narrower than its parent — was fixed in #544 (read-scope inheritance). **This issue is the defense-in-depth remainder:** when a denial is genuinely correct (a truly out-of-scope path), or any future regression reintroduces over-denial, the fork should not waste its whole budget.

## What already exists (and why none of it covers this)

- **Fork wall-clock budget** — `SUBAGENT_DEFAULT_TIMEOUT_MS = 20 * 60_000` (foreground), `SUBAGENT_BACKGROUND_TIMEOUT_MS = 60 min` (`src/agent/subagent.ts:91,103`), added in #465. Bounds total TIME but the fork still burns the *entire* budget before aborting. The observed "timed out after ~20 min" is this bound firing — not an infinite hang.
- **Tool-iteration cap** — bounds ROUNDS, not time; throttled/denied rounds still accrue (`subagent.ts:83` comment).
- **Consecutive byte-identical circuit breaker** — `src/agent/tools/dispatcher.ts:43-58` short-circuits the 8th+ *byte-identical* consecutive call. It does **not** fire here: the spinning fork issues *distinct* reads (different paths), each denied — not one identical call repeated.

So there is a gap between "same call repeated" (caught) and "budget expired" (20 min wasted): **many distinct reads all path-denied** has no fast-fail.

## Evidence (2026-07-12 /diagnose fan-out)

5 sessions, ~114 read denials, 0 write denials — every denied path inside a managed worktree or the main checkout the fork could no longer reach. Single worst fork (session `8831d316`): 40 `read_file`/`grep`/`glob` denials before the 20-min budget aborted it. Witness traces: `~/.afk/state/witness/{8831d316,80414672,b90afdd4,39db6994,8152433f}/trace.jsonl`. Surfaced by the new `subagent-read-denial` detector (#545).

## Proposal

Add a **denial circuit breaker** in the fork tool loop / dispatcher:

- Count path-approval `decision:'block'` results within a fork's turn (keyed by the denial category, not the specific path). After **N** (suggest 5–10) such denials, abort the child fast.
- On abort, return an **explicit, actionable** tool_result to the parent — the accumulated denied paths + the grant remedy (e.g. "these paths are outside the fork's read roots; widen via …") — mirroring the existing `TimeoutError` tool_result shape so the parent can re-dispatch with a corrected grant.
- **Fail LOUD, never silent-skip.** A fast abort must be a visible failure the orchestrator can act on — it must not be swallowed into a normal/empty result that reads as success. (This was the explicit risk flagged when scoping the fix.)
- Preserve the `afk farm` / deliberately-confined-worker case: a confined worker that legitimately cannot read some paths should still fail *loud and fast*, not hang — but the message should name confinement as the likely intent.

## Acceptance criteria

- [ ] A fork accumulating ≥N path-approval read denials aborts in **seconds**, not at the 20-min wall-clock bound.
- [ ] The parent receives a structured, actionable failure naming the denied paths + remedy (not a silent/empty success).
- [ ] Write-denial confinement (worktree isolation) behavior is unchanged.
- [ ] The `subagent-read-denial` detector (#545) shows the class no longer recurring after the fix (`afk improve scan --only subagent-read-denial`).
- [ ] Unit test: simulated denial stream trips the breaker at the threshold; below-threshold denials do not.

## Relevant files

- `src/agent/tools/dispatcher.ts` — where the consecutive-identical breaker lives; natural home for a denial-count breaker.
- `src/agent/providers/anthropic-direct/loop.ts` — the tool-use loop that currently retries denied reads.
- `src/agent/subagent.ts` — wall-clock budget + fork lifecycle (`withTimeout`, AbortGraph cascade).
- `src/agent/tools/hooks/path-approval-hook.ts` — emits the `decision:'block'` the breaker would count.

## Related

- #544 (merged) — read-scope inheritance; removes the most common *cause* of these denials.
- #465 (merged) — bounded the fork wall-clock budget this complements (fail *within* the budget, not *at* it).
- Stale local branch `afk/bg-timeout-classification` — leftover of #465; not a starting point (already merged, ~107 behind main).

_Suggested labels: enhancement, subagent, reliability._

griffinwork40:
## Summary

Closes #546. A forked sub-agent that accumulates repeated **path-approval READ denials** — each a *distinct* path, so the byte-identical repeat circuit breaker never fires — previously kept issuing new denied reads until its 20-minute wall-clock budget (`SUBAGENT_DEFAULT_TIMEOUT_MS`) expired. This adds a per-fork **denial circuit breaker** that aborts in *seconds* once the fork is provably spinning on unrecoverable denials, and hands the parent a **loud, structured, actionable** failure (denied paths + grant remedy) — never a silent/empty success.

This is the defense-in-depth remainder after #544/#554 fixed the most common *cause* (over-narrow fork read scope): even when a denial is genuinely correct, or a future regression reintroduces over-denial, the fork now fails *within* its budget rather than *at* it.

## Approach

The design was hardened by a `/devils-advocate` pass that caught a fatal flaw in the first draft: a **throw** from the dispatcher does **not** abort the fork — `executeCore` never propagates (`dispatcher.ts` no-throw invariant) and the loop's `executeBatch`/`execute` catch downgrades any throw into a per-round `isError` result and keeps looping. So the loud path routes through a normalized **`error` event** instead.

- **`SessionToolDispatcher`** counts *consecutive* path-approval read denials on a forked child (`parentSessionId` set), **reset on any successful tool call**. At `DENIAL_CIRCUIT_BREAKER_THRESHOLD` (5) it tags the tripping result `failureClass: 'denial-breaker'`. Mirrors the existing `repeatBreaker` (per-dispatcher = per-forked-query).
- **Both provider loops** (`anthropic-direct/loop.ts`, `openai-compatible`) detect that tag *after* pushing the round's `tool_result` blocks (transcript stays consistent) and surface it as a terminal `error` event → the shared subagent handle (`streamToFinalMessage` → `buildResultFromError`) rethrows it into the same structured, `isError: true` parent tool_result shape a timeout produces.

## Invariants preserved

- **Interactive sessions never trip** — only forks auto-deny reads (`parentSessionId` gate); an interactive session gets a prompt instead.
- **Write-confinement untouched** — only READ path-tools are counted (`read_file`/`list_directory`/`glob`/`grep`); worktree write-isolation and `afk farm` write denials are a separate mechanism.
- **Reset-on-success** — a fork that probes a couple of out-of-scope paths and then makes progress never trips; only true starvation (zero successful tool calls) does.
- **Confined-worker framing** — the abort message names deliberate confinement (`afk farm`) as a possible + expected cause, so a confined worker still fails loud + fast with the right guidance.

## Acceptance criteria (from #546)

- [x] A fork accumulating ≥N read denials aborts in **seconds**, not at the 20-min bound.
- [x] The parent receives a structured, actionable failure naming denied paths + remedy (not a silent/empty success) — via the `error` event → `buildResultFromError` chain.
- [x] Write-denial confinement behavior unchanged (READ-only counting + test).
- [x] Unit test: a simulated denial stream trips the breaker at the threshold; below-threshold denials do not.
- [ ] `subagent-read-denial` detector (#545) shows the class no longer recurring — validated post-merge against live forks (can't fully reproduce pre-merge).

## Files

| File | Change |
|------|--------|
| `src/utils/errors.ts` | new `DenialCircuitBreakerError` |
| `src/agent/tools/denial-circuit-breaker.ts` | **new** — threshold, read-tool set, path extractor, message builder |
| `src/agent/tools/dispatcher.ts` | per-fork counter, `recordForkReadDenial` + `resetDenialBreaker`, wired at both hook-block catch sites + reset-on-success |
| `src/agent/trace/types.ts` | `'denial-breaker'` added to `TOOL_FAILURE_CLASSES` (non-exempt: review-worthy) |
| `src/agent/providers/anthropic-direct/loop.ts` | detect trip → terminal `error` event |
| `src/agent/providers/openai-compatible/{query.ts,query/dispatch-append.ts}` | same, via generator return value |

## Testing

- `pnpm lint` (tsc strict) ✓ · `pnpm audit:sdk:check` (no new SDK symbols) ✓ · `pnpm scan:env:check` ✓
- `pnpm test:coverage` — full suite green, coverage thresholds met ✓
- New: `denial-circuit-breaker.test.ts` (13 tests) + `loop.denial-breaker.test.ts` (fail-loud + loop-stops).

### DIFF
--- a/src/agent/providers/anthropic-direct/loop.denial-breaker.test.ts
+++ b/src/agent/providers/anthropic-direct/loop.denial-breaker.test.ts
@@ -1,0 +1,77 @@
+import { describe, expect, it } from 'vitest';
+import type { MessageParam } from '@anthropic-ai/sdk/resources';
+import { runTurn } from './loop.js';
+import type { ToolCall, ToolResult } from './types.js';
+import { DenialCircuitBreakerError } from '../../../utils/errors.js';
+import { DENIAL_BREAKER_FAILURE_CLASS } from '../../tools/denial-circuit-breaker.js';
+import {
+  fromArray,
+  collect,
+  ctx,
+  makeTextStream,
+  makeToolUseStream,
+  makeClient,
+  makeBatchDispatcher,
+} from './loop.test-helpers.js';
+
+// #546: when the dispatcher trips the denial circuit breaker it tags the
+// tool result `failureClass: 'denial-breaker'`. The loop MUST surface that as a
+// loud terminal `error` event (→ DenialCircuitBreakerError, rethrown into a
+// structured failure by the shared subagent handle) and STOP — never silently
+// continue looping to the wall-clock budget, and never dress it as a success.
+describe('loop.ts — denial circuit breaker fail-loud', () => {
+  it('yields a terminal DenialCircuitBreakerError event and stops the loop on a denial-breaker result', async () => {
+    let callIdx = 0;
+    const client = makeClient(() => {
+      callIdx += 1;
+      // Round 1: the model issues a read. Round 2 (a text answer) must NEVER be
+      // reached — tripping the breaker ends the turn immediately.
+      if (callIdx === 1) {
+        return fromArray(makeToolUseStream('toolu_r', 'read_file', '{"file_path":"/out-of-scope/x.ts"}'));
+      }
+      return fromArray(makeTextStream('this second round should never run'));
+    });
+
+    const DENIAL_MSG =
+      'Denial circuit breaker: this forked sub-agent hit 5 consecutive path-approval read denials';
+    const dispatcher = makeBatchDispatcher(
+      async (calls: ToolCall[]): Promise<ToolResult[]> =>
+        calls.map(() => ({
+          content: DENIAL_MSG,
+          isError: true,
+          failureClass: DENIAL_BREAKER_FAILURE_CLASS,
+        })),
+    );
+
+    const messages: MessageParam[] = [{ role: 'user', content: 'read a bunch of files' }];
+    const events = await collect(
+      runTurn({
+        client,
+        messages,
+        system: null,
+        tools: [{ name: 'read_file', input_schema: { type: 'object' } }],
+        toolDispatcher: dispatcher,
+        model: 'claude-test',
+        maxTokens: 1024,
+        headers: {},
+        signal: new AbortController().signal,
+        ctx,
+      }),
+    );
+
+    // 1. A loud error event was emitted, carrying the actionable message.
+    const errorEvent = events.find((e) => e.type === 'error');
+    expect(errorEvent).toBeDefined();
+    if (errorEvent?.type === 'error') {
+      expect(errorEvent.error).toBeInstanceOf(DenialCircuitBreakerError);
+      expect(errorEvent.error.message).toContain('Denial circuit breaker');
+    }
+
+    // 2. The loop STOPPED — the model was called exactly once, never looping to
+    //    a second round (which is the wall-clock-budget burn #546 kills).
+    expect(client.messages.create).toHaveBeenCalledTimes(1);
+
+    // 3. Fail LOUD, not a silent success: no clean turn.completed after the trip.
+    expect(events.some((e) => e.type === 'turn.completed')).toBe(false);
+  });
+});
--- a/src/agent/providers/anthropic-direct/loop.ts
+++ b/src/agent/providers/anthropic-direct/loop.ts
@@ -63,6 +63,8 @@ import {
   resolveMaxToolIterations,
   shouldWindDown,
 } from '../shared/tool-loop-cap.js';
+import { DENIAL_BREAKER_FAILURE_CLASS } from '../../tools/denial-circuit-breaker.js';
+import { DenialCircuitBreakerError } from '../../../utils/errors.js';
 
 // Re-exported from the provider-neutral `shared/tool-loop-cap.ts` (single
 // source of truth shared with openai-compatible). Kept exported here so
@@ -853,6 +855,22 @@ export async function* runTurn(
       content: toolResultBlocks as ContentBlockParam[],
     };
     input.messages.push(toolResultTurn);
+
+    // Denial circuit breaker (#546): the dispatcher tagged a result
+    // `denial-breaker` after a forked child hit N consecutive path-approval
+    // read denials with no progress. Surface it as a LOUD `error` event — never
+    // a silent partial — so the shared subagent handle rethrows it into a
+    // structured `buildResultFromError` failure the parent can act on (mirrors
+    // the TimeoutError teardown). History is already consistent here (assistant
+    // `tool_use` + matching `tool_result` blocks both pushed above), so
+    // returning is safe; the fork is done. A dispatcher throw would instead be
+    // swallowed by the executeBatch/execute catch above and the loop would keep
+    // spinning — which is the exact failure this breaker exists to prevent.
+    const denialTrip = results.find((r) => r.failureClass === DENIAL_BREAKER_FAILURE_CLASS);
+    if (denialTrip) {
+      yield { type: 'error', error: new DenialCircuitBreakerError(denialTrip.content) };
+      return;
+    }
     } catch (err) {
       // Rollback the orphaned assistant `tool_use` push above so the next
       // turn's API call does not 400 with "tool_use ids were found without
--- a/src/agent/providers/openai-compatible/query.denial-breaker.test.ts
+++ b/src/agent/providers/openai-compatible/query.denial-breaker.test.ts
@@ -1,0 +1,229 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import type OpenAI from 'openai';
+import type { ProviderEvent, ProviderUserTurn } from '../../provider.js';
+import type { AgentConfig } from '../../types/config-types.js';
+import type { OpenAIChunk } from './translate.js';
+import type { ToolCall, ToolResult } from '../anthropic-direct/types.js';
+import { __setOpenAIClientFactory, OpenAICompatibleQuery, type OpenAIClientFactory } from './query.js';
+import { SessionToolDispatcher } from '../../tools/dispatcher.js';
+import { builtinToolSchemas } from '../../tools/schemas.js';
+import { createHookRegistry } from '../../hooks.js';
+import type { ToolHandler } from '../../tools/types.js';
+import { DenialCircuitBreakerError } from '../../../utils/errors.js';
+import {
+  DENIAL_BREAKER_FAILURE_CLASS,
+  DENIAL_CIRCUIT_BREAKER_THRESHOLD,
+} from '../../tools/denial-circuit-breaker.js';
+
+// #546: the openai-compatible provider is the SECOND half wired for the denial
+// circuit breaker (query.ts + query/dispatch-append.ts). When the dispatcher
+// tags a tool result `failureClass: 'denial-breaker'`, dispatchAndAppend hands
+// the tripping ToolResult back up through its widened generator return type
+// (`AsyncGenerator<ProviderEvent, ToolResult | undefined>`, read via the nested
+// `{ call, result }` shape), and query.ts surfaces a LOUD terminal `error`
+// event (→ DenialCircuitBreakerError), nulls its abortController, and STOPS the
+// loop — never looping to the wall-clock budget, never a silent success. This
+// file is the openai analog of anthropic-direct/loop.denial-breaker.test.ts and
+// additionally proves the REAL dispatcher trips end-to-end through this loop.
+
+// ---- minimal scripted-client harness (mirrors query.test.ts) --------------
+
+let createCalls: Array<{ args: unknown }> = [];
+let scriptedTurns: Array<{ chunks: OpenAIChunk[] }> = [];
+let scriptedTurnIndex = 0;
+
+function installScriptedClient(): void {
+  const factory: OpenAIClientFactory = () =>
+    ({
+      chat: {
+        completions: {
+          create: async (args: { stream?: boolean }) => {
+            createCalls.push({ args });
+            if (!args.stream) throw new Error('test mock only supports streaming');
+            const script = scriptedTurns[scriptedTurnIndex++];
+            if (!script) throw new Error(`scripted turn ${scriptedTurnIndex - 1} not defined`);
+            const chunks = script.chunks.slice();
+            return (async function* () {
+              for (const c of chunks) yield c;
+            })();
+          },
+        },
+      },
+    }) as unknown as OpenAI;
+  __setOpenAIClientFactory(factory);
+}
+
+async function collect(query: AsyncIterable<ProviderEvent>): Promise<ProviderEvent[]> {
+  const out: ProviderEvent[] = [];
+  for await (const ev of query) out.push(ev);
+  return out;
+}
+
+async function* singleInput(content: string): AsyncIterable<ProviderUserTurn> {
+  yield { content };
+}
+
+function baseConfig(over: Partial<AgentConfig> = {}): AgentConfig {
+  return { model: 'gpt-4o-mini', apiKey: 'sk-test-key', ...over } as AgentConfig;
+}
+
+/** One scripted round in which the model issues a single read_file tool call. */
+function readRound(id: string, filePath: string): { chunks: OpenAIChunk[] } {
+  return {
+    chunks: [
+      {
+        choices: [
+          {
+            delta: {
+              tool_calls: [
+                {
+                  index: 0,
+                  id,
+                  type: 'function',
+                  function: { name: 'read_file', arguments: JSON.stringify({ file_path: filePath }) },
+                },
+              ],
+            },
+          },
+        ],
+      },
+      {
+        choices: [{ delta: {}, finish_reason: 'tool_calls' }],
+        usage: { prompt_tokens: 5, completion_tokens: 5, total_tokens: 10 },
+      },
+    ],
+  };
+}
+
+/** A trailing text round the loop must NEVER reach once the breaker trips. */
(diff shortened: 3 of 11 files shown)


griffinwork40:
Verification complete: **tsc --noEmit passes clean (exit 0)** and **all 14 new tests pass**. I independently confirmed the two load-bearing findings against source. Here is the synthesized review.

---

# Code Review — PR #592: denial circuit breaker for forked read-denial spins

**What it does:** When a forked sub-agent hits 5 consecutive path-approval READ denials with no intervening successful tool call, the dispatcher tags the result `failureClass: 'denial-breaker'`; both provider loops surface that as a loud terminal `error` event (→ `DenialCircuitBreakerError`, rethrown by the subagent handle into a structured `buildResultFromError` failure) and stop — instead of the fork spinning to its 20-min wall-clock budget.

**Verification I ran (this review):** `tsc --noEmit` strict gate → clean; 14/14 new tests pass. Four parallel read-only dimension reviewers (correctness, integration, test-coverage, design) + my own spot-checks of the two most load-bearing claims.

## Verdict: ✅ Approve with minor changes

The feature is correct and shippable in both providers. Two items are worth addressing before merge; the rest are non-blocking follow-ups.

### Findings by severity

**[MAJOR] openai-compatible provider path has zero tests.** The PR wires the breaker into *both* providers, but only `anthropic-direct` has a stop-the-loop test (`loop.denial-breaker.test.ts`). The openai path is the *more intricate* half — it widens two generator return types to `AsyncGenerator<ProviderEvent, ToolResult | undefined>` and reads a **different result shape** (`r.result.failureClass` on a nested `{call, result}` element vs anthropic's flat `r.failureClass`), nulls `abortController`, and skips `finishTurn`. Statically verified correct by the correctness+integration reviewers, but a future shape/return regression would *silently fail to trip* and is uncaught. → Add a mirror test for `query.ts:514-525` / `dispatch-append.ts:245`. *(Cheap, closes the regression gap on the wired-but-untested half.)*

**[MAJOR] Breaker counts *any* `HookBlockedError` on a read tool, not only path-approval denials.** `recordForkReadDenial` (dispatcher.ts:608) keys on `hook-block` + `READ_PATH_TOOLS` + `parentSessionId`, but `hook-registry.ts:211` throws the same `HookBlockedError` for every blocking PreToolUse decision. Built-in gates are safe today (plan-mode exempts forks; afk-mode rates reads safe), but I confirmed two real trip surfaces: (a) the **credential read-denylist floor** (`path-approval-hook.ts:223-237`, blocks `~/.ssh` etc. for "forks alike") and (b) **user-defined PreToolUse hooks**. Both would emit the breaker's `"widen readRoots"` remedy — which directly contradicts the denylist's own `"never readable … do not retry"`. The fail-fast teardown itself stays defensible (a fork that can't read those paths *is* spinning), so this is **message-accuracy/precision**, not a broken feature. → Discriminate on the block reason (path-approval emits `"outside the session's granted read roots"`) or thread a structured flag; at minimum soften the remedy for the non-path-approval case.

**[MINOR] `executeBatch` reset-on-success is untested** — only the single `execute()` reset path has a test (`denial-circuit-breaker.test.ts:406`). The batch reset (dispatcher.ts:913-920) is unguarded.

**[MINOR] Batch-trip assertion is weak** — `test.ts:437` asserts only `results.some(r => r.failureClass === 'denial-breaker')`; doesn't pin the tripping index or verify prior results stayed `hook-block`.

**[MINOR] Two pure-helper tests are near-tautological** (`toBe(5)`, `toBe('denial-breaker')`) — fine as change-detectors, add no behavioral coverage.

**[NIT] Trace asymmetry** — the denial trip sets `failureClass` but *not* `circuitBreaker: true`, while the repeat breaker does the reverse. A reviewer scanning traces for `circuitBreaker: true` won't see denial trips.

**[NIT] `DenialCircuitBreakerError` carries no structured field** (e.g. `deniedPaths[]`), only a formatted message string — fine for the current consumer (parent reads `.message`); limits future programmatic use.

**[NIT / benign] `executeBatch` trip-and-reset ordering** — a batch containing a success *and* 5 denials both trips and resets, but the tripping result is already frozen into the array so the `error` still surfaces; the reset only affects future rounds that never happen. Worth a one-line comment.

### What's done well
- **Excellent documentation** of the load-bearing design choice — *why* a tagged result beats a throw (a dispatcher throw would be swallowed by the `executeBatch`/`execute` catch and the loop would keep spinning; both reviewers independently CONFIRMED this against loop.ts:735-757).
- **Write-confinement invariant preserved and test-locked** — `READ_PATH_TOOLS` excludes `write_file`/`edit_file`/`bash`; worktree/`afk farm` isolation untouched.
- **Fork-only guard** via `parentSessionId` is the correct, established signal (matches path-approval's own auto-deny discriminator).
- **Observability choices well-reasoned** — deliberately *not* exempt from failure-density/human-review stats (a torn-down fork is a genuinely actionable event, and 5-consecutive-no-progress is rare enough not to add noise).
- **Threshold consistent** with the existing `REPEAT_CIRCUIT_BREAKER_THRESHOLD`; reset-on-success makes it self-correcting; the two breakers are orthogonal (disjoint code paths, cannot double-trip).

### Suggested pre-merge asks (both cheap)
1. Add an openai-compatible stop-the-loop test mirroring the anthropic-direct one.
2. Tighten the trip discriminator to genuine path-approval containment denials (or soften the remedy message so a credential-denylist/user-hook block doesn't misdirect with "widen readRoots").

Everything else can be a fast-follow.

## Done
- **Reviewed** PR #592 (+548/−13, 9 files) across correctness, integration/api-compat, test-coverage, and design via 4 parallel read-only reviewers plus independent spot-checks of the 2 load-bearing findings.
- **Evidence:** `tsc --noEmit` exit 0 and 14/14 new tests passing (run in worktree `/Users/griffinlong/Projects/open_source/agent-afk/.afk-worktrees/denial-circuit-breaker`); every finding carries file:line citations in the sections above. The two MAJORs were re-verified by me directly against `path-approval-hook.ts:223-237`, `hook-registry.ts:211`, and a repo grep confirming no openai-compatible denial test.
- **World change:** none — read-only review; no files edited, committed, or pushed; working tree remains clean (only non-mutating `vitest`/`tsc` were run).
- **Recommendation:** Approve with minor changes — request the openai-compatible test + remedy-message precision before merge; treat the remaining MINOR/NIT items as fast-follows.

---

griffinwork40:
`pnpm lint` (tsc --noEmit, strict) passed **clean** — authoritatively confirming the return-type widening and the `ToolFailureClass` enum addition introduce no type breaks. `results` is the properly-typed `ToolResult[]` in loop scope (tsc would have flagged `.failureClass` otherwise). The full chain, all four acceptance criteria, and every claimed invariant are verified against source + passing tests.

Two Wave-1 subagents failed at dispatch, so this review was conducted inline with direct source verification of the load-bearing claim. No critical/high findings surfaced, so the `/shadow-verify` backstop (reserved for critical/high) is not required.

---

# Code Review — PR #592

**Target:** PR #592 `feat(subagent): fail-fast denial circuit breaker for forked read-denial spins (#546)` · `afk/denial-circuit-breaker` → `main` · +930/-13, 11 files · change type: **feature** · regime: **full**
**Reviewed ref:** `ede5cf55ffd9c648a07922a68469463a97aacbef` — all citations read at this ref (worktree clean at HEAD). Independent checks run: 22 new tests **pass**, `tsc --noEmit` strict **clean**.
**Stated intent:** PR #546 body — spec-compliance **assessed** against its 4 acceptance criteria + 4 claimed invariants.

## Verdict on the load-bearing claim (acceptance criterion 2)
The whole PR hinges on a tripping `failureClass:'denial-breaker'` result becoming a **loud, structured, `isError:true`** failure to the parent (never a silent success). **This holds, verified end-to-end:**
- Both provider loops detect the tag and yield a terminal `{type:'error', error: DenialCircuitBreakerError}` then `return` — `loop.ts:869-872`, `query.ts:522-526`.
- `streamToFinalMessage` sets `streamError` on the error event and `throw`s it, with `if (streamError) throw streamError` ordered **before** the `finalMessage` return so it wins even if an assistant message was mid-streamed — `handle.ts:343-345, 367`.
- `runToResult` catches → `buildResultFromError` → `SubagentResult{error}` — `handle.ts:412-419`.
- Foreground branch: `result.status==='succeeded'` is false, so it takes the failure branch, reads `result.error.message` (the loud denial text) into `buildFailurePayload`, returns `{content, isError:true}` — `foreground-promotion.ts:250-291`.
This is byte-for-byte the `TimeoutError` teardown rail (`run()` wraps `streamToFinalMessage` in `withTimeout`, `handle.ts:186`), exactly as the PR claims.

## Findings

**low · medium-confidence · correctness/spec-adherence · `dispatcher.ts:928-934` (batch reset) vs `dispatcher.ts:983-1006` (recordForkReadDenial) · ref:ede5cf5 · citation-type:file-state**
Within a *single* parallel batch, denials are counted in phase-1 **before** the end-of-batch reset-on-success fires, so a batch containing ≥THRESHOLD distinct containment-denied reads **and** a successful sibling call trips the breaker despite that progress — a narrow deviation from the stated invariant "only true starvation (zero successful tool calls) trips."
Evidence: `if (results.some((r) => r !== undefined && r.isError !== true)) { this.resetDenialBreaker(); }` runs after `recordForkReadDenial` already incremented+tripped in phase-1.
Suggestion: acceptable as-is (reachable only when a model emits 6+ parallel tool calls, 5 of them out-of-scope reads — a strong spin signal); if tightened, skip the trip when any sibling in the same batch succeeded.

**nit · high-confidence · test-coverage · `loop.denial-breaker.test.ts:43-50` · ref:ede5cf5 · citation-type:diff-context**
The anthropic-direct loop test drives a **mock** batch dispatcher; the real-`SessionToolDispatcher` end-to-end trip is proven only on the openai side (`query.denial-breaker.test.ts:298-353`).
Suggestion: none required — dispatcher logic is provider-agnostic and separately unit-tested in `denial-circuit-breaker.test.ts` (19 tests); the openai e2e proves the real wiring.

**nit · high-confidence · test-coverage · `foreground-promotion.ts:288-291` · ref:ede5cf5 · citation-type:file-state**
No test asserts the parent tool_result is `isError:true` for a denial trip specifically; it is covered transitively by the shared error-event→`buildResultFromError`→failure-branch rail (the `TimeoutError` path).
Suggestion: none required.

## What was clean
- **security** — no issues found (read `denial-circuit-breaker.ts`, `dispatcher.ts`, `path-approval-hook.ts` at ede5cf5): breaker only tears forks down *faster*, never grants access; denied paths surface only to the trusted parent; credential-denylist and user-hook denials are correctly excluded from counting via `isSubagentContainmentDenial` (`denial-circuit-breaker.ts:101-103`, tested at `denial-circuit-breaker.test.ts:744-769`).
- **api-compat** — no issues found: `dispatchAndAppendToolCalls` return-type widening has one internal caller (`query.ts:815`); `TOOL_FAILURE_CLASSES` addition is non-breaking (`tool-failure-density.ts` uses a runtime exempt-array, no exhaustive `never`-switch exists); `tsc --noEmit` clean.
- **Invariants confirmed** — interactive/non-fork never trips (`parentSessionId` gate, tested); write-confinement untouched (READ-only `READ_PATH_TOOLS`, tested with `write_file`); reset-on-success in both `execute()` and `executeBatch()` (tested); threshold off-by-one correct (5th denial trips); no double-count between the single and batch catch sites; confined-worker `afk farm` framing present (tested). `run()` tags the failure `errorClass:'DenialCircuitBreakerError'` for trace forensics (`handle.ts:263`).

## Epistemic scope disclosure
- Wave-1 agents read against branch HEAD `ede5cf5`; all citations verified at that ref by direct file reads in the clean worktree.
- Stated intent: PR #546 title+body — spec-compliance assessed; all 4 acceptance criteria and 4 invariants trace to code + passing tests (criterion 5, the live post-merge detector check, is explicitly deferred by the author and unverifiable pre-merge).
- **Not checked:** full test suite (ran only the 3 new files + `tsc`); Telegram/REPL surface behavior on receiving the failure payload; the exact `run()` re-throw line past `handle.ts:264` (architecturally required and relied on by the tested timeout path).
- Two prescribed Wave-1 dimension subagents failed at dispatch ("produced no terminal message"); review was completed inline by the coordinator with direct source reads — coverage of all dimensions was preserved, but without the second independent agent lens.

## Merge decision: **MERGE**
No critical/high findings. The stated purpose is achieved, the load-bearing error-event→parent-failure chain is verified end-to-end, all invariants hold, types are clean, and the 22 new tests pass. The three findings are low/nit and none block merge.

**Done**
- Reviewed PR #592 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability; verified the central error-event→`buildResultFromError`→`isError:true` parent-failure chain against source at ref `ede5cf5`.
- Evidence: 22/22 new tests pass and `pnpm lint` (tsc strict) clean (this session's shell output); citations at commit `ede5cf55ffd9c648a07922a68469463a97aacbef`; diff artifact `/Users/griffinlong/.afk/state/skill-preflight/a59295bd-efed-4103-91b5-0367072ce1b5/pr-592.diff`.
- World state: unchanged — this review is read-only; no files, commits, or PR mutations were made.
- Outcome: **MERGE** with three non-blocking low/nit findings recorded above.
- Deferred/pending: acceptance criterion 5 (live `subagent-read-denial` detector showing the class no longer recurs) is author-deferred to post-merge and was not validatable pre-merge; two prescribed dimension subagents failed at dispatch so the review ran inline without the second independent-agent lens.

---

PR #595, opened 2026-07-13
feat(tui): surface provider rate-limit/backoff live in the progress banner

griffinwork40:
## Why
From the REPL UX audit: when the Anthropic SDK is throttled (429/503/529 + `Retry-After`) it sleeps-and-retries **inside a single `messages.create()`** — so a 70s Retry-After retried twice (~140s) shows the user *nothing* live and looks like a frozen session. Until now it was only recorded in the witness trace, after the fact.

## What
Surfaces the backoff live in the progress banner (e.g. `rate-limited · retrying in ~70s`):
- New `rate_limit` `ProviderEvent` variant (`provider.ts`).
- `tracing-fetch` gains an optional `onThrottle` callback (existing `rate_limit` trace-phase emission untouched — both channels fire).
- A tiny out-of-band `ThrottleQueue` (single-producer/consumer promise mailbox) bridges the wrapped-fetch callback → the per-turn loop, which **races** its `messages.create` await against the queue and yields a `rate_limit` event per drained signal — live, *before* the response streams.
- `stream-consumer` maps it to an output event; a pure `formatRateLimitActivity(retryAfterMs?)` renders the banner string; the next real progress/done evicts it.

## Non-goals
- Does **not** change retry policy/timing (the SDK owns retries) or add model-downgrade/fallback.

## Verification
- `pnpm lint` clean; `pnpm audit:sdk:check` OK (no new SDK imports).
- Full suite **`11759 passed | 14 skipped`**; touched-file + provider/session suites green. New unit tests for the formatter + the stream-consumer mapping.

## Reviewer focus
- **The `Promise.race` / generator in `awaitCreateWithThrottleSignals` (loop.ts)** — out-of-band emission correctness: `create` rejections must re-throw to the caller's catch; no unhandled rejection; no lost/duplicated signal (level-triggered drain-then-park).
- Banner injection relies on the `lastProgressByTask` slot in the OverlayComposer path — a live TTY smoke test would confirm the paint (not exercised in CI).

🤖 Built via parallel afk subagent (opus) + orchestrator verification.

### DIFF
--- a/src/agent/provider.ts
+++ b/src/agent/provider.ts
@@ -209,6 +209,36 @@ export type ProviderEvent =
       type: 'stream.retry';
       sessionId?: string;
     }
+  | {
+      /**
+       * Live rate-limit / backoff signal. Emitted when the Anthropic provider
+       * is throttled (HTTP 429/503/529 with a `retry-after` header) and the
+       * SDK is about to sleep-and-retry INSIDE a single `messages.create`
+       * call. Because that backoff happens deep inside the wrapped `fetch`
+       * (see `providers/anthropic-direct/tracing-fetch.ts`), the per-turn loop
+       * is blocked awaiting the SDK and yields nothing — so a healthy session
+       * waiting out a 70s `retry-after` (retried twice ≈ 140s) looks frozen.
+       * This event is pushed out-of-band from the fetch throttle callback into
+       * the loop's yield stream so the UI can surface a live
+       * `rate-limited · retrying in ~70s` banner during the wait; the normal
+       * activity resumes once the retried request streams.
+       *
+       * Distinct from `stream.retry` (a MID-stream 529 re-drive that discards
+       * partial text) — this is a CONNECTION-phase throttle observed from
+       * inside fetch, and it does NOT invalidate any already-streamed text. It
+       * is purely observational: the SDK still owns the retry policy and
+       * timing (this feature does not change either). Consumers that don't
+       * render a live banner may ignore it. `retryAfterMs` is the parsed
+       * `retry-after` value when present; `status` is the throttled HTTP
+       * status; `attempt` is the 1-based throttle count within the call when
+       * the emitter can supply it.
+       */
+      type: 'rate_limit';
+      sessionId: string;
+      retryAfterMs?: number;
+      status?: number;
+      attempt?: number;
+    }
   | { type: 'error'; error: Error }
   | {
       type: 'paused';
--- a/src/agent/providers/anthropic-direct/index.ts
+++ b/src/agent/providers/anthropic-direct/index.ts
@@ -35,6 +35,7 @@ import {
 } from './auth.js';
 import { oneShotCompletion, type OneShotInput } from './oneshot.js';
 import { makeTracingFetch } from './tracing-fetch.js';
+import { ThrottleQueue } from './throttle-queue.js';
 import { refreshClaudeCodeOauthToken } from '../../auth/keychain.js';
 import { AnthropicDirectQuery } from './query.js';
 import { pathContainmentBypassed } from '../../permission-policy.js';
@@ -602,16 +603,34 @@ export class AnthropicDirectProvider implements ModelProvider {
       );
     }
     const authMode = detectAuthMode(token);
+    // Live-throttle mailbox: the wrapped fetch pushes a signal onto this queue
+    // for every 429/503/529 the SDK sleep-and-retries INSIDE a single
+    // `messages.create`; the per-turn loop drains it to surface a `rate_limit`
+    // ProviderEvent LIVE (the loop is otherwise parked awaiting the SDK and
+    // cannot yield during the backoff). Installed only when the tracing fetch
+    // is (non-local-shim + a trace writer OR a live surface would consume it) —
+    // here it rides alongside the existing trace-writer gate so the queue and
+    // the wrapper share the same lifetime. The SAME instance is handed to the
+    // query below so the fetch producer and the loop consumer meet.
+    const throttleQueue =
+      !localMode && config.traceWriter ? new ThrottleQueue() : undefined;
     const clientOpts = buildClientOptions(
       token,
       authMode,
       config.baseUrl,
-      // Observability: route SDK HTTP through a wrapper that records 429/503/529
-      // throttling into the witness trace, so the SDK's otherwise-silent
-      // retry-after backoff is legible in `afk trace show`. Skipped in
-      // local-shim mode (not Anthropic's billing surface) and when no trace
-      // writer is attached.
-      !localMode && config.traceWriter ? makeTracingFetch(config.traceWriter) : undefined,
+      // Observability: route SDK HTTP through a wrapper that (1) records
+      // 429/503/529 throttling into the witness trace so the SDK's
+      // otherwise-silent retry-after backoff is legible in `afk trace show`,
+      // and (2) pushes a live signal onto `throttleQueue` so the progress
+      // banner can show the backoff as it happens. Skipped in local-shim mode
+      // (not Anthropic's billing surface) and when no trace writer is attached.
+      !localMode && config.traceWriter
+        ? makeTracingFetch(
+            config.traceWriter,
+            undefined,
+            throttleQueue ? (info) => throttleQueue.push(info) : undefined,
+          )
+        : undefined,
     );
     const factory = this.providerFactory ?? clientFactory;
     const client = factory ? factory(clientOpts) : new Anthropic(clientOpts);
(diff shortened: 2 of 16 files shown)


griffinwork40:
The third agent failed for the same structural reason as the first two: **`research-agent` has no Bash tool and cannot `git show` content at a ref that isn't checked out** (the working tree is on `main`, which lacks these files). Its only path to ref content was nesting a `git-investigator` sub-agent — which is exactly what blew the first two agents' budgets (timeout / preamble-only).

**Delegation failure — stated fallback:** I completed the review via direct read-only inspection (`git show <ref>:<path>`, `git grep`) of every load-bearing fact, plus an inline trace of the concurrency logic from the full source I had already read. No critical/high findings emerged, so `/shadow-verify` is not required. This is the review artifact.

---

# Code Review — PR #595

**Target:** PR #595 `feat(tui): surface provider rate-limit/backoff live in the progress banner` · `afk/tui-ratelimit-live` → `main`
**Reviewed ref:** `3d51288bee5f8ef77cf0bf47581cf7c79cd981e6` (all citations read at this ref)
**Regime:** full (feature, +855/−40, 16 files) · **Stated intent:** PR title+body — spec-compliance assessed.

## Summary
A clean, additive, well-documented observability feature. The Anthropic SDK sleeps-and-retries 429/503/529 *inside* a single `messages.create()`, so a long `retry-after` looks like a frozen session. This bridges the wrapped-`fetch` throttle callback → the per-turn loop via a tiny out-of-band `ThrottleQueue`, races the `create` await against the queue, and surfaces a live `rate-limited · retrying in ~70s` banner. The core concurrency logic (the author's #1 review focus) is **correct** on all four axes it flagged. No blocking defects.

## Findings

**No critical / high / medium findings.**

Findings below are all **low** or informational — none blocks merge.

---

**① low · confidence high · correctness/invariant · `src/cli/_lib/stream-renderer-orchestrator.ts:407–412` · ref:3d51288b · citation-type:file-state**
Finding: the `rate_limit` case calls `lastProgressByTask.set(RATE_LIMIT_TASK_ID, …)` **without** a preceding `clear()`, so a mid-turn throttle (a 429 on tool-use round ≥2, when a real `progress` entry already occupies the map) yields a **2-entry map** and the overlay renders **two stacked banner lines** — violating the documented "at most one entry" invariant at line 149.
Evidence:
```
// stream-renderer-orchestrator.ts:407-410 — no clear() before set
ctx.lastProgressByTask.set(RATE_LIMIT_TASK_ID, { … });
// vs the 'progress' case (line 161-162) which DOES clear first:
ctx.lastProgressByTask.clear(); ctx.lastProgressByTask.set(event.progress.taskId, …);
// overlay renders every entry (line 480): for (const progress of ctx.lastProgressByTask.values()) …
```
Suggestion: call `ctx.lastProgressByTask.clear()` before the `set` in the `rate_limit` case (mirroring the `progress` case), so the backoff banner replaces rather than stacks; the next real `progress` restores the normal banner. (It self-heals on the next progress event, hence low.)

---

**② low · confidence high · test-coverage · `src/agent/providers/anthropic-direct/loop.throttle.test.ts:1–159` · ref:3d51288b · citation-type:file-state**
Finding: no test exercises the **`createPromise` rejection path with a wired `throttleQueue`** — the exact behavior the author flagged as the #1 review concern ("create rejections must re-throw to the caller's catch; no unhandled rejection"). The three tests cover only: success+single-signal, multi-attempt, and no-queue passthrough (`git grep -nE 'reject|throw|rejects' … loop.throttle.test.ts` → zero matches; it is the only test file referencing `throttleQueue`). The code itself is correct (I re-derived the rejection propagation by hand — see below), so this is a regression-guard gap, not a live defect.
Evidence: the three `it(...)` titles in the file — none references rejection/throw; `client.messages.create` is always resolved (`resolveCreate(fromArray(...))`), never rejected.
Suggestion: add one test that rejects `createGate` after a `queue.push(...)` and asserts (a) the `rate_limit` event was still yielded and (b) `collect(runTurn(...))` rejects with the create error (no unhandled-rejection warning).

---

**③ informational · confidence high · spec-scope · `src/agent/providers/anthropic-direct/index.ts:613–614` · ref:3d51288b · citation-type:diff-context**
Finding: the live banner is gated on `!localMode && config.traceWriter`, so setting `AFK_TRACE_DISABLED=1` (which makes `createDefaultTraceWriter()` return null) silently disables this **UI** feature along with tracing. This is deliberate per the PR ("rides alongside the existing trace-writer gate so the queue and the wrapper share the same lifetime") and documented in-code — flagged only so it is a conscious decision, not a surprise.
Evidence: `const throttleQueue = !localMode && config.traceWriter ? new ThrottleQueue() : undefined;`
Suggestion: none required; if the banner is later wanted independent of tracing, decouple the gate from `traceWriter`.

## Verified correct (the author's called-out risks)
Re-derived by hand from `loop.ts` `awaitCreateWithThrottleSignals` + `throttle-queue.ts` at the ref:
- **Rejection propagation ✓** — `guarded = createPromise.then(v=>{settled=true;return v}, e=>{settled=true;throw e})`; on rejection the loop's next iteration hits `if (settled) return await guarded`, which re-throws `e` out through the `yield*` into `runTurn`'s `try/catch` — identical to a plain `await createWithRetry(...)`.
- **No unhandled rejection ✓** — `.then(onF,onR)` attaches a handler to `createPromise` synchronously at entry; the race's `guarded.then(()=>CREATE_DONE, ()=>CREATE_DONE)` attaches a rejection handler to `guarded` synchronously on iteration 1, before the microtask that can reject it runs. `settled` is always `false` on the first check (callbacks are pending microtasks).
- **No lost / duplicated signal ✓** — `takeAll()` (splice-once) and `waitForItem()` (level-triggered, resolves immediately when `items.length>0`) are called with no `await` between them; a `push` during a `yield` is caught by the next `takeAll`; the post-`settled` `continue` performs a final drain. Each signal is spliced and yielded exactly once.
- **Stale waiter across turns ✓** — a resolver left in `this.waiter` after a `CREATE_DONE` exit is overwritten by the next turn's `waitForItem()`; turns are serial and no fetch runs between them, so no spurious cross-turn wake or hang; the abandoned promise is GC'd.
- **api-compat additive ✓** — `makeTracingFetch` has exactly 2 production callers (`index.ts:628,852`, both updated); the new `onThrottle` is an optional 3rd param; the `!writer` → `!writer && !onThrottle` early-return preserves behavior for existing writer-only callers. `transformProviderEvent` retains its `default` arm; no `assertNever` switch over `ProviderEvent`/`OutputEvent` exists to break. No new `@anthropic-ai/sdk` runtime import (only the local `ThrottleQueue` import) — consistent with the PR's `audit:sdk:check OK`.
- **Feature fires on the primary surface ✓** — the REPL (`bootstrap.ts:221`) and one-shot chat (`chat.ts:438`, wired at `:457`) both construct `createDefaultTraceWriter()`; `agent-session.ts` → `index.ts:917` threads `config.traceWriter` to the provider, satisfying the gate for Anthropic (non-local) models.

## Merge decision: **MERGE**
Correct core logic, thorough unit tests, additive/backward-compatible surface, no data/auth/correctness risk. The two low findings (stack-vs-replace banner on mid-turn throttle; missing rejection-path test) are worthwhile but non-blocking follow-ups — I'd recommend landing the rejection test since it guards the subtlest path.

## What was not checked
- **Ref:** all citations read against branch HEAD `3d51288bee5f8ef77cf0bf47581cf7c79cd981e6` via `git show <ref>:<path>` (working tree is on `main`); no patch-file-only citations.
- **Stated intent:** PR #595 title+body — spec-compliance assessed; non-goal (retry policy/timing unchanged) confirmed (code is observational only).
- **Not exercised:** live TTY paint of the banner (the author notes CI doesn't cover it — I confirmed the injection *logic* and its tests but did not run an interactive terminal). Did not run the test suite (trusted the PR's `11759 passed | 14 skipped` claim; not re-verified). Did not review the Telegram/`farm` surfaces' banner rendering (only their trace-writer wiring, which is present). openai-compatible provider is out of scope (feature is anthropic-direct-only by design).
- **Process note:** the skill's parallel review sub-agents (`research-agent`) could not run because they lack Bash and cannot read at a non-checked-out ref without nesting a `git-investigator` (which exhausted their budgets); this review was completed by direct read-only inspection + inline re-derivation instead.

---

griffinwork40:
Pushed `52ae87b` addressing the Codex review.

**P2 (stream-renderer-orchestrator.ts) — fixed.** The synthetic `__rate_limit__` banner was evicted only by a `progress`/`stream_retry` event or turn finalization. A throttled request that resolves into a plain-text answer emits no `progress` event (loop.ts yields progress only per tool-use round), so the banner lingered under the streamed answer. The `chunk` arm now drops `RATE_LIMIT_TASK_ID` on the first stream chunk (any type), using the same `delete()`-guarded single repaint as the `stream_retry`/reset arms, so the banner covers only the real wait window. New regression test reproduces the text-only-resume case.

**P2 (index.ts — wire live queue when `AFK_TRACE_DISABLED=1`) — deferred by decision.** It only affects sessions that opted out of witness tracing and is not a regression (those sessions simply don't get the new banner, same as before this PR). Tracked for a follow-up if trace-disabled interactive UX becomes a priority.

Gates: full suite `11771 passed | 14 skipped`, `pnpm lint` clean, `audit:sdk:check` / `audit:env:check` / `scan:env:check` all green. (Also completed two test fixtures that omitted the required `OrchestratorCtx.lastProgressByTask`.)

PR #597, opened 2026-07-14
fix(subagent): bound read-only research agents + degrade empty no-terminal run to a marked partial

griffinwork40:
## Problem

`awa-private:research-agent` dispatches in the hourly PR-review daemon reliably fail: the subagent runs 80–106 tool calls over ~15 min, then returns with no output (a red ×) and the orchestrator falls back. The underlying error is the opaque `Subagent … produced no terminal message`.

## Root cause (shadow-verified)

Two stacked defects:

1. **Unbounded tool loop.** The `agent`-tool dispatch path resolves `maxToolUseIterations` to `0` (unlimited) when neither the call-site nor the agent definition sets one — `child-config.ts:230-234,266-269` — deliberately bypassing `SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS = 50` (`subagent.ts:74,675-676`), which now only guards internal skill/compose forks. A read-only research/review agent therefore has no anti-hang bound and can tool-loop indefinitely.
2. **Opaque failure on cutoff.** When such a run is cut off mid-loop with an empty buffer, `handle.ts` threw `produced no terminal message` — an opaque hard failure with nothing to salvage — instead of degrading. The adjacent *buffered-text* branch already degrades to a `STREAM_INCOMPLETE` partial; the no-text case did not.

(The builtin `research-agent` and plugin `awa-private:research-agent` are otherwise identical — byte-identical prompt, same `Agent(git-investigator)` grant; the only difference is the plugin's `model: sonnet` pin vs the builtin's inherit.)

## Fix

- **Bound the read-only research/review builtins** (`research-agent`, `Explore`) with `maxToolUseIterations = 50` in `builtins.ts`, so a runaway winds down through the graceful capped-partial path. `general-purpose` stays **unbounded** (a real multi-step worker legitimately needs a long loop) — no blanket cap.
- **Degrade the empty no-terminal stream end** in `handle.ts` to a `STREAM_INCOMPLETE` partial (marked at every consumption boundary by `annotateIfIncomplete`), so the parent gets an actionable, clearly-flagged incomplete result instead of an opaque throw. **Explicit `cancel()` / cascade-abort still throw**, so a cancelled subagent is never mislabeled a succeeded partial.

## Testing

- `pnpm lint` clean.
- Full `pnpm test` suite green (649 files, 11846 tests).
- `handle-streaming.test.ts`: empty stream now asserts a marked partial; the abort test still asserts rejection.
- `registry.test.ts`: asserts research-agent/Explore bounded at 50 and general-purpose left unbounded.

## Blast radius

- The degradation only changes the `handle.ts` terminal that previously threw with zero salvage; cancel/abort classification is unchanged.
- Bounding is limited to the two read-only search/review builtins.

## Follow-up (separate, private repo)

`awa-private:research-agent`'s own frontmatter should get the same `maxToolUseIterations` bound in `agent-framework-private` so it also winds down early like the builtin. (This PR's `handle.ts` degradation already makes its runaway fail *gracefully* rather than opaquely in the meantime.) Optionally drop its `model: sonnet` pin to match the model-inheriting builtin.

### DIFF
--- a/src/agent/agents/builtins.ts
+++ b/src/agent/agents/builtins.ts
@@ -51,6 +51,24 @@ Rules:
 Reply compactly: the answer first, then evidence as file:line citations, then open questions or paths not checked. No preamble.`;
 
 /**
+ * Anti-runaway tool-use-round bound for the read-only research/review builtins.
+ *
+ * The `agent`-tool dispatch path is unlimited-by-default: `child-config.ts`
+ * resolves a named dispatch's `maxToolUseIterations` to 0 (no cap) when neither
+ * the call-site nor the definition sets one, deliberately bypassing
+ * SubagentManager's `SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS` (50) — which only
+ * guards internal skill/compose forks. Unbounded is defensible for a real
+ * multi-step worker (`general-purpose`), but a READ-ONLY research/review agent
+ * never legitimately needs many rounds; without a bound it can loop on a hard
+ * task until an external cutoff aborts it mid-loop, surfacing as an opaque
+ * failure. Bounding it routes a runaway through the graceful capped-partial
+ * wind-down instead. Value mirrors `SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS`;
+ * duplicated (not imported) to keep `agents/` off the `subagent.ts` module
+ * graph — same rationale as the KNOWN_AFK_TOOL_NAMES dup in resolve.ts.
+ */
+const READONLY_AGENT_MAX_TOOL_USE_ITERATIONS = 50;
+
+/**
  * Build the built-in agents, keyed by name.
  *
  * A fresh map per call — callers merge it into a mutable registry under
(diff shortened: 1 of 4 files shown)


griffinwork40:
Addressed the Codex **P2** (`handle.ts` empty-fallback stop reason) in b2e2b3a.

**Confirmed it was a live bug, not just latent.** The stream consumer emits a `message` event only when the assistant turn has text (`stream-consumer.ts` `case 'assistant.message': if (event.text)`). An empty-text turn that ends with a clean terminal reason (`end_turn`/`max_tokens`) is therefore dropped without a `message` event, but a `turn.completed` still produces a `done` carrying that clean reason — so the empty-fallback branch was reached with `lastStopReason` already set to `end_turn`. The `??=` preserved it, and `annotateIfIncomplete` treats `end_turn` as clean, so the synthetic "no findings" placeholder was delivered as a clean completion with **no** partial marker — defeating the purpose of this PR.

**Fix:** stamp `STREAM_INCOMPLETE` unconditionally (`=`, not `??=`) at this branch. Nothing worth preserving can reach here — the `tool_use_loop_capped` case returns earlier and cancel/abort throws earlier — so any surviving reason is a clean terminal one that must be overwritten. The sibling buffered-text branch keeps `??=` deliberately, because a capped run that also streamed text legitimately preserves `tool_use_loop_capped` there.

**Regression test:** `empty run + done{stopReason: 'end_turn'} => result.stopReason === STREAM_INCOMPLETE` (fails under the old `??=`, passes now). Full suite green (11847 passed), lint clean.

PR #606, opened 2026-07-14
fix(tui): make --plain / AFK_PLAIN_OUTPUT a full render opt-out on TTY

griffinwork40:
## Update — full opt-out completed (addresses Codex review)

The initial change (commit `ad2a892`) gated the **three compositor sites** listed under *Fix* below and claimed a `--plain` TTY then "fully behaves like a non-TTY surface." That was an over-claim. Codex review correctly flagged two rendering surfaces that gate **independently** on `process.stdout.isTTY` — which `--plain` does *not* change — so they stayed live on a `--plain` TTY:

- **P1 (status / footer):** `StatusLine` still armed a DECSTBM scroll region and painted a cursor-positioned status row, and the three footer painters (`BackgroundStatusBar`, `LoopStageBar`, verdict ledger) still wrote cursor-positioned rows — all concurrent with the renderer's raw `console.log` writes (`completionWriter` stays at `console.log` when there is no compositor). That is exactly the Stage-3e line-tracker corruption class the flag exists to escape.
- **P2 (input reader):** `readWithAutocomplete()` gates only on `process.stdout/stdin.isTTY`, so on a `--plain` TTY it still called the raw-mode `readWithAutocompleteTty` (raw mode + cursor/dropdown accounting) — the "downgrades to the simple reader" line under *Behavior change* was not actually true.

These gate independently of the compositor, so the 3-lever change never reached them. Follow-up commit `8228c72` consults `isPlainOutputRequested()` at the five remaining `isTTY`-derived gates so a `--plain` TTY is fully inert, like a genuine non-TTY stdout:

- `status-line.ts` — `StatusLine.enabled` (the single escape-emitting chokepoint → `start`/`repaint`/`setExtraRows`/`onResize`/`rearm`/`withFullScrollRegion`/`stop` all no-op; `force` still wins for tests)
- `background-status-bar.ts`, `commands/interactive/loop-stage.ts`, `commands/interactive/verdict-ledger.ts` — `start()` stays inert (no ResizeBus subscription, spinner timer, row reservation, or CUP paint; every paint path is `started`-gated)
- `input-box.ts` — `readWithAutocomplete()` now routes to the simple `readNonTty` reader on `--plain` (makes the P2 claim actually true)

All five gates are additive: **no behavior change when the flag is unset.** +14 regression tests exercising the real `stream.isTTY` / `process.*.isTTY` paths (see *Testing*).

---

## Problem
`--plain` / `AFK_PLAIN_OUTPUT=1` is documented as forcing append-only output "instead of the live-overlay renderer" — an escape hatch for tmux/SSH/multiplexer sessions where the overlay's cursor-up redraws and DECSTBM scroll regions misbehave. But on a **real TTY** it only rerouted the *between-turn* write seam. The **mid-turn** live overlay (tool lanes, thinking, streaming) — the primary source of cursor-up redraws — kept rendering, so the escape hatch did not actually escape the thing it names.

There was also a latent corruption risk: the persistent compositor stayed armed while the plain seam's `setCompositor` was a no-op, so between-turn raw writes hit stdout while log-update was tracking it — the exact pre-"Stage-3e" condition documented as the "stacked prompt" duplication bug (`surface-setup.ts`).

## Root cause
The flag (added in #593) was read in exactly one place — `isPlainOutputRequested()` in `repl-renderer.ts`, gating only the renderer seam. It was never propagated to the two compositor-arming sites:
- `InputSurface.armCompositor` early-returns only on `!stdout.isTTY || !stdin.isTTY` (persistent compositor still arms on a `--plain` TTY).
- `StreamRenderer` computes `isTTY` from `process.stdout/stdin.isTTY` and, in `arm()`, builds its **own** `TerminalCompositor` when `isTTY && !this.compositor` — so even with no borrowed compositor a `--plain` TTY rebuilt the overlay per turn.

## Fix
Promote `isPlainOutputRequested()` to `config/env.ts` (reads via the `env` object, so `audit:env:check` stays clean) and consult it at all three render-decision sites so a `--plain` TTY fully behaves like a non-TTY surface:
- `repl-renderer.ts` — between-turn seam (gate unchanged; now uses the shared predicate)
- `input-surface.ts` — `armCompositor` early-return (**Lever 1** — no persistent compositor)
- `stream-renderer.ts` — `isTTY` computation (**Lever 2** — no per-turn compositor; every `if (this.isTTY && ...)` overlay branch skipped)

## Behavior change (intended)
On a `--plain` TTY the input surface now downgrades to the simple non-TTY line reader (`readNonTty` — no raw-mode input box, no autocomplete dropdown), completed in `8228c72` (see **Update** at top). For an opt-out escape hatch this is the desired trade: you are opting out of the fancy TTY rendering wholesale, matching the "same code path already used for non-TTY surfaces" framing. Default TTY behavior is unchanged when the flag is unset. Docs (flag help, comments, `ENV_REGISTRY`) + regenerated `docs/env-registry.{json,md}` updated to match.

## Testing
- `pnpm lint` (strict `tsc --noEmit`) — pass
- `pnpm audit:env:check` — pass (0 violations)
- `pnpm scan:env:check` — pass (env-registry in sync)
- `pnpm test` — full suite: **12035 passed** (12021 baseline + 14 new), 14 skipped, 0 failed
- Added regression coverage for **both** levers, exercising the real `process.stdout/stdin.isTTY` detection path (not `forceNonTty`, which would make the assertion vacuous): with the flag set, `StreamRenderer.isTTY === false` / `arm()` builds no compositor, and `armCompositor` leaves `getCompositor()` null — both would fail on the unfixed code. Also covers unset (no behavior change) and unrecognized-value (`"0"`) cases.
- Follow-up (`8228c72`) adds 14 tests for the five additional gates — `StatusLine.enabled`, the three footer painters' `start()`, and `readWithAutocomplete()` routing — each asserting the flag path against the real `stream.isTTY` / `process.*.isTTY` detection (`=1`, `=TRUE`, unset, `"0"`), plus a `force:true`-still-wins case for `StatusLine`.

## Note on verification
This is static + unit-level verification. A live `AFK_PLAIN_OUTPUT=1 afk i` run on a real TTY (ideally under tmux/SSH) watching a tool-heavy turn is recommended before merge, to confirm the mid-turn overlay is gone and between-turn output is clean.


griffinwork40 (commented):

## Merge decision: ✅ MERGE

**PR #606** · `fix(tui): make --plain / AFK_PLAIN_OUTPUT a full render opt-out on TTY` · reviewed ref `ad2a892`
Regime: light (hotfix; small logic delta — ~6 source files, mostly prose/tests). Read-only review; all citations verified against branch HEAD `ad2a892`.

### Summary
The fix promotes `isPlainOutputRequested()` from a module-local helper in `repl-renderer.ts` to an exported predicate in `config/env.ts` and consults it at all three render-decision sites, so a `--plain` / `AFK_PLAIN_OUTPUT=1` TTY session fully behaves like a non-TTY surface. Each of the three gates named in the PR description is confirmed wired at the reviewed ref:

- `repl-renderer.ts:68` — between-turn seam: `if (!stdout.isTTY || isPlainOutputRequested())` (Lever 0)
- `input-surface.ts:296` — `armCompositor` early-return: `if (!stdout.isTTY || !stdin.isTTY || isPlainOutputRequested()) return;` (Lever 1 — no persistent compositor)
- `stream-renderer.ts:343` — `this.isTTY = … && !isPlainOutputRequested() && …` (Lever 2 — no per-turn compositor; every `if (this.isTTY && …)` overlay branch skipped)

Because the mid-turn overlay/cursor-up escapes are all downstream of `this.isTTY`, folding the predicate into that computation leaves no residual overlay-escape path on a `--plain` TTY — which is the specific gap the PR set out to close.

### No blocking findings

**security · api-compat** — No secret exposure, no injection surface; the predicate reads `env.AFK_PLAIN_OUTPUT` (a getter over `process.env`) and returns a boolean. Not a breaking API change: at the parent commit the symbol was module-local (no `export`), so it had zero possible cross-module importers; relocating it to `config/env.ts` is a private-symbol refactor.

**correctness** — Env read is a live getter re-evaluated per `StreamRenderer` construction (fresh per turn), so a mid-session change takes effect next turn; unset → `false` (zero behavior change). No temporal-dead-zone hazard: `export const env` (env.ts:1226) precedes `isPlainOutputRequested` (env.ts:1405), and the reference is at call time. The old local copy is fully removed from `repl-renderer.ts` (no drifting duplicate).

**spec-compliance** — Judged against the PR description. No unmet intent (all three sites gated) and no scope creep (changes limited to the three sites, their tests, the env source description, prose help/comment updates, and the regenerated `docs/env-registry.{json,md}` — all called for).

**test-coverage** — New real-TTY-path regression tests confirmed present for Lever 1 (`input-surface.test.ts`, 4 cases) and Lever 2 (`stream-renderer.test.ts`, 5 cases), covering `=1`, `=TRUE`, unset, and `"0"`, via genuine `process.stdout/stdin.isTTY` stubs (not `forceNonTty`). Lever 0 is covered by pre-existing tests (`repl-renderer.test.ts:83-142`) that now transitively exercise the shared predicate.

### Non-blocking note
- **low · test-coverage** — This commit swaps `repl-renderer.ts` from a local function to the shared import but adds no new assertion there that the *shared* predicate is the one wired in; the pre-existing Lever 0 tests would pass against either the old local copy or the new import. This is low-value to add: the type system plus the Lever 1/2 tests importing the same symbol already cover the wiring indirectly, and Lever 0 behavior itself is covered. No integration test asserts all three sites agree — also low materiality, since the shared predicate is a single pure function tested per-site.

### What was not checked
- Static + unit-level review only. A live `AFK_PLAIN_OUTPUT=1` run on a real TTY (ideally under tmux/SSH) watching a tool-heavy turn — recommended in the PR's own "Note on verification" — was not performed; escape-sequence suppression was confirmed at the gate level, not observed end-to-end.
- Claimed CI/test results (`pnpm lint`, `audit:env:check`, `scan:env:check`, 12021-test suite) were not re-executed; CI on the PR is reported green.
- The two auto-generated `docs/env-registry.{json,md}` files were skipped per policy (the `env.ts` ENV_REGISTRY source was reviewed instead); the regenerated output was not diffed against the source.

griffinwork40:
Both Codex findings verified valid against the branch code and fixed in `8228c72`.

**P1 (status/footer)** — confirmed: `ctx.statusLine.start()` (interactive.ts:766) and `setupFooterSubsystems` (repl-loop.ts:108) run unconditionally, and `StatusLine.enabled` + the three footer painters gate on `process.stdout.isTTY` (still true under `--plain`), not on the compositor. So a `--plain` TTY armed a DECSTBM scroll region + cursor-positioned footer rows while `completionWriter` stayed at raw `console.log` (surface-setup.ts:277). Fixed by gating `StatusLine.enabled` and each painter's `start()` on `isPlainOutputRequested()`.

**P2 (input reader)** — confirmed: with the compositor null, `readLine()` falls to `readWithAutocomplete()`, whose only gate is `!process.stdout/stdin.isTTY` (input-box.ts) → still called raw-mode `readWithAutocompleteTty` on a `--plain` TTY. Fixed by routing to `readNonTty` when `isPlainOutputRequested()`, which also makes the description's "downgrades to the simple reader" claim true.

+14 regression tests across the 5 gates; full suite 12035 pass / 0 fail; lint + `audit:env:check` + `scan:env:check` green. Description updated with the retraction + completion details. Static/unit-level only — the live tmux/SSH TTY check noted in *Note on verification* is still worth doing before merge.

PR #607, opened 2026-07-14
feat(hooks): discover and run Claude Code plugin hooks.json (#70)

griffinwork40:
Closes #70.

## Summary

AFK now discovers and runs **Claude Code plugin-contributed hooks** — a `<plugin>/hooks/hooks.json` shipped by an installed plugin. This was the one plugin-contributed surface AFK didn't honor (skills, agents, commands, and `mcp.json` were already discovered), and it means plugins written for Claude Code are portable across both runtimes without change.

The `afk.config.json`-driven shell-hook system already existed; this adds the **plugin** layer #70 tracks, behind its own trust gate.

## What changed

- **Discovery** — `discoverPluginHooksConfigs()` walks `~/.afk/plugins/` (flat **and** marketplace-cache layouts, depth-bounded) for `<plugin>/hooks/hooks.json`, keyed on the `.claude-plugin/plugin.json` manifest. Mirrors the existing `discoverPluginMcpConfigs()` precedent.
- **Trust gate** — new user-global **`enablePluginHooks`** (default `false`), **independent** of `enableShellHooks`, and **human-tier** (the agent's `config_set` cannot flip it). Rationale: plugin `hooks.json` executes third-party code, so it gets its own explicit opt-in rather than riding on the gate that governs the user's *own* `afk.config.json` hooks — consistent with AFK's fail-closed, per-tier posture (`allowProjectHooks`, `AFK_ALLOW_PROJECT_MCP`).
- **Registration** — plugin hooks are tagged `tier:'plugin'`, merged last, and register in the bridge *independently* of the shell-hook gate.
- **Executor env** — plugin hook commands receive `CLAUDE_PLUGIN_ROOT` (the plugin's install dir) and `CLAUDE_PROJECT_DIR`, so commands like `python3 "${CLAUDE_PLUGIN_ROOT}/hooks/on_start.py"` resolve. Existing secret redaction (no API keys/tokens forwarded) still applies.
- **Not silent** — when plugin `hooks.json` are present but the gate is off, a one-line warning names how to enable them.
- **Plumbing + docs** — `settable-keys`, `json-tier`, `config`, `types`, `config-menu`, and `configuration/overview.mdx`.

## Scope

**In:** user-global installed plugins; `command`-type hooks; AFK's supported hook events.
[...]

PR #618, opened 2026-07-15
feat(hooks): deliver SessionStart injectContext to the first turn

griffinwork40:
Follow-up to #607.

## Summary

#607 wired plugin `hooks.json` discovery **and** the mid-conversation `injectContext` delivery paths (SubagentStop / UserPromptSubmit / Stop) — but **SessionStart `injectContext` was parsed and then dropped**. `dispatchSessionStart` returned `void`, so a plugin SessionStart hook's `hookSpecificOutput.additionalContext` reached **neither the model nor the user** (only a byte-count landed in the witness trace).

This wires the last hop so a SessionStart hook's context is actually delivered.

## The gap

A plugin `SessionStart` hook emitting `{"hookSpecificOutput":{"hookEventName":"SessionStart","additionalContext":"…"}}` had that context:
- parsed into `decision.injectContext` by the command executor (as for every event), then
- **discarded** — `dispatchSessionStart` returned `void`; its caller (`pullInitialization`) awaited it only to allow a block-throw and captured nothing.

The contract comment even documented SessionStart as one of "the remaining hook events [that] ignore `injectContext` entirely", so under Claude Code these hooks inject context, but under AFK they no-op.

## What changed

- **`src/agent/session/hooks-dispatch.ts`** — `dispatchSessionStart` now returns the registry-merged `injectContext` string (`Promise<string | undefined>`) instead of `void`. The block/throw path is unchanged.
- **`src/agent/session/agent-session.ts`** — `pullInitialization` queues the returned string via the existing `queueFrameworkContext`, so it rides the session's **first** outbound user message. SessionStart fires during init before any turn exists, so there is no in-flight prompt to prepend to — the queue (drained behind `initPromise` in `sendMessageStreamInternal`) bridges init to the first send. Queued, not pushed, to avoid the one-turn displacement documented on `queueFrameworkContext`.
- **`src/agent/hooks.ts`** — the `injectContext` contract (module block + `HookDecision.injectContext` field doc) now lists SessionStart and describes first-turn delivery.

Reuses the proven SubagentStop delivery path (`queueFrameworkContext` → `withPendingFrameworkContext`) — no new machinery, additive, backward-compatible. No new trust surface: SessionStart hooks already *ran*; this only delivers their output, still behind the human-tier `enablePluginHooks` gate.

## Ordering invariant

`pullInitialization` queues the context **before** the init loop resolves `initPromise`; `sendMessageStreamInternal` awaits `initPromise` (before `withPendingFrameworkContext`), so the queued context is guaranteed present for the first send. Documented inline as an `Invariant:` comment.

## Testing

- **`hooks-dispatch.test.ts`** (net-new) — return contract: single handler, merged across handlers, none, and block-throws.
- **`framework-context-queue.test.ts`** (+2) — end-to-end: a SessionStart hook's context is prepended to the first message; absent case stays unprefixed.
- Full suite: **12128 passed / 0 failed / 14 skipped**.
- `tsc --noEmit`, `audit:sdk:check`, `audit:env:check`, `scan:env:check` — all clean.

## Scope

**In:** SessionStart `injectContext` first-turn delivery + contract docs + tests.
**Out:** other events' delivery (unchanged); the `enablePluginHooks` gate (unchanged, still human-tier default-false).


griffinwork40:
I have enough to complete a source-grounded review. The change is small (5 files, +119/-7) and I've traced every load-bearing claim to source directly (not via a subagent), so dispatching parallel dimension agents would add overhead without adding signal. Synthesizing the review.

Note on running tests: the working tree is on `main`, not the PR branch, and the review is read-only (no checkout/stash allowed), so the PR's tests can't be executed here — I evaluated them by reading.

---

# Code Review — PR #618: deliver SessionStart `injectContext` to the first turn

**Verdict: Approve with one decision to confirm before merge.** The mechanism is correct, well-documented, and the happy-path is well-tested. There is one **reachable, undocumented, untested blast-radius** the human should consciously accept or scope: this activates SessionStart `injectContext` for **every subagent fork**, not just the top-level session — and there is currently no way for a handler to opt out.

## What the change does (verified)
`dispatchSessionStart` now returns the merged `injectContext` string (was `Promise<void>`); `AgentSession.pullInitialization` queues it via `queueFrameworkContext` before the init loop resolves `initPromise`, so it rides the first outbound user message. Ordering is sound: `queueFrameworkContext` runs synchronously inside `pullInitialization` before `initPromise` resolves, and `sendMessageStreamInternal` awaits `initPromise` (agent-session.ts:605) *before* draining via `withPendingFrameworkContext` (agent-session.ts:617). The Invariant comment (agent-session.ts:70) is accurate.

## Findings by severity

### 🟡 Medium — SessionStart `injectContext` now propagates to every subagent, with no opt-out
This is the one item worth a conscious decision.

- Subagent forks are `AgentSession` instances that inherit the **same** bubbled hook registry (`hookRegistry: registry`, subagent.ts:780). The in-code comment there is explicit: *"its own SessionStart/SessionEnd/PreToolUse fire against it"* (subagent.ts:777).
- Every `AgentSession` runs `pullInitialization` → `dispatchSessionStart` → (new) `queueFrameworkContext`. So each subagent prepends the SessionStart context to its first/only user message.
- This is **reachable via the standard config path**: `command-executor.ts` maps a shell hook's `hookSpecificOutput.additionalContext` → `injectContext` *event-agnostically* (command-executor.ts:341-346), and config-bridge registers the SessionStart handler returning that decision (config-bridge.ts:128). That's exactly the Claude-Code-style session-priming pattern (inject date / git branch / project context at session start).
- **No opt-out exists.** The codebase's established self-skip mechanism is a `parentSessionId` guard *in the handler* (subagent.ts:778). But `pullInitialization` passes only `{ event: 'SessionStart', sessionId }` (agent-session.ts:330) — it never threads `this.config.parentSessionId`, even though `SessionStartContext` defines the field (hooks.ts:120). So a handler *cannot* distinguish a subagent SessionStart from a top-level one, and cannot skip it.

Consequence: a priming `additionalContext` gets prepended to all N subagent prompts on a wide fan-out (compose/DAG, ceiling-test 20×) — token cost multiplied, and narrowly-scoped subagent tasks get prefixed with session-level context. Before this PR, SessionStart `injectContext` was inert everywhere, so this behavior is newly activated.

**Two clean resolutions (either is fine):**
1. *If subagent injection is intended:* say so in the PR/JSDoc and add a test locking it — plus thread `parentSessionId` into the SessionStart context so handlers *can* opt out per the existing pattern.
2. *If it is not:* gate the `queueFrameworkContext` call on top-level sessions only (skip when `this.config.parentSessionId` is set).

### 🟢 Low — `reset()` / `/clear` re-injection is correct but untested
`reset()` → `initSdkLifecycle()` clears the queue (agent-session.ts:314) then re-arms `pullInitialization` (agent-session.ts:319), so SessionStart re-fires and re-injects on `/clear` with no stale accumulation. This is correct and sensible, but there's no test pinning it — worth one so a future refactor doesn't silently drop it.

### 🟢 Low / pre-existing — first-message slash command + prefix
The end-to-end test asserts `${NUDGE}\n\n/resolve` — i.e. the prefix lands ahead of the literal `/resolve`. If a genuine slash command ever reached `sendMessageStream` as raw content, prepending breaks the leading `/`. In practice slash commands are intercepted upstream, and this is pre-existing `queueFrameworkContext` behavior (SubagentStop already does this) — not a regression this PR introduces. Noted only for completeness.

## Dimension summary
- **Correctness:** Sound. Ordering guarantee holds; block path (`HookBlockedError`) short-circuits before queueing; empty/whitespace context is dropped by `queueFrameworkContext`'s trim guard. ✓
- **API-compat:** `Promise<void>` → `Promise<string | undefined>` is a backward-compatible widening; the sole non-test caller (agent-session.ts:328) is updated. `hooks-dispatch.ts` is internal, not a public surface. ✓
- **Test coverage:** Strong on the unit contract (undefined registry / no-inject / single / `\n`-merge / block) and end-to-end delivery + no-inject-unprefixed. Gaps: subagent behavior (the Medium item) and reset re-injection. 
- **Security:** No new trust boundary — the `additionalContext`→`injectContext` path and shell-hook trust gate (`enableShellHooks`) pre-date this PR. ✓
- **Perf/observability:** Trace still emits `injectedContextBytes` (hooks-dispatch.ts:67, unchanged). The only perf note is the subagent multiplication above.

## Recommendation
Merge is reasonable. Before merging, confirm the subagent blast-radius is intended and either document + test it (and thread `parentSessionId` for opt-out) or scope the queueing to top-level sessions. Adding a `reset()` re-injection test is a nice-to-have.

---

**Done**
- Reviewed PR #618 end-to-end against source; every load-bearing claim re-derived from primary files by the main session (not delegated), citations inline above.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/d26aa0ec-4ce0-49e1-9307-09b994c049f9/pr-618.diff`; key sources `agent-session.ts:305-379,585-640,752-817,1057-1088`, `hooks-dispatch.ts:73-88`, `hook-registry.ts:218-233`, `config-bridge.ts:107-134`, `command-executor.ts:341-346`, `subagent.ts:561,777-780`, `hooks.ts:104-120`.
- No files modified (read-only review); working tree remains clean on `main`.
- One decision deferred to the human reviewer: whether SessionStart `injectContext` should propagate to subagent forks (Medium finding), with two concrete resolution paths given.

---

griffinwork40:
Follow-up fix pushed (c35247a): **scope SessionStart `injectContext` delivery to the parent session.**

Review caught that the injectContext was queued unconditionally in `pullInitialization`, but subagent forks run the same init path with the bubbled hook registry — so a session-priming SessionStart hook (the standard `hookSpecificOutput.additionalContext` pattern, mapped event-agnostically in `command-executor.ts`) would prepend its context to **every subagent's first prompt**: token cost × fan-out, plus contamination of narrowly-scoped subagent tasks.

Changes:
- Gate delivery on the top-level predicate (`config.parentSessionId === undefined`) in `AgentSession.pullInitialization` so only the parent session delivers it. This mirrors the self-skip convention `SessionEnd` already documents, and Claude Code's own per-session (not per-subagent) SessionStart semantics (see anthropics/claude-code#60112, whose fix is 'skip SessionStart hooks for background subagent sessions').
- Thread `parentSessionId` into the SessionStart hook context (it was absent, unlike `SessionEndContext`) so programmatic handlers can self-skip subagent starts.
- New regression test: a forked (`parentSessionId`-set) session leaves its first outbound message unprefixed.

Note: the SessionStart hook still *dispatches* for subagents (unchanged) — only `injectContext` *delivery* is now parent-only. `pnpm lint` + `pnpm test` on the hooks/session suites (273 tests) green.

griffinwork40:
I've now **independently confirmed the load-bearing ordering invariant** by reading the actual source at `ref:3f69d6c` (not trusting the agent's compressed claim):

- **Queue push happens-before `initPromise` resolves:** `pullInitialization` (`agent-session.ts:326`) awaits `dispatchSessionStart` (`:328`), then queues **synchronously** at `:356-358` (lines 340-355 between are pure comments — no `await`, no race window), *before* entering the `while` loop at `:360`. Every return point (`:364`, `:382`, `:389`) is at/after the loop — strictly after the push. Since `initPromise === pullInitialization()` (`:319`), the push is guaranteed before `initPromise` resolves. ✓
- **Drain happens-after `initPromise` resolves:** `sendMessageStreamInternal` awaits `this.initPromise` as its *first* statement (`:628`), then drains via `withPendingFrameworkContext` at `:640` — 12 lines later. ✓
- **Prepend + empty-drop semantics:** `withPendingFrameworkContext` (`:1103-1111`) does `${prefix}\n\n${content}`; `queueFrameworkContext` (`:1093-1095`) trims and drops empties. ✓

The invariant holds. No critical/high findings surfaced by either agent, so Wave 1.5 citation/absence verification and the mandatory `/shadow-verify` gate (both scoped to blocking/critical/high) have nothing to process — and I've already re-derived the single decision-driving claim inline. Proceeding to synthesis.

---

# Code Review — PR #618

**Target:** `feat(hooks): deliver SessionStart injectContext to the first turn` · `afk/session-start-injectcontext` → `main` · reviewed ref `3f69d6c` · +168/−7, 5 files
**Regime:** full (2 parallel dimension agents) — line count said `light`, escalated for session-init concurrency + context-injection sensitivity
**Stated intent:** supplied (PR #618 title + body) — spec-compliance **assessed**

## Summary

| | |
|---|---|
| Change type | feature (follow-up to #607) |
| Blocking / critical / high | **0** |
| Medium | 0 |
| Low / informational | 3 |
| Merge decision | **MERGE** |

The PR wires the one missing hop from #607: `dispatchSessionStart` now returns the registry-merged `injectContext` (`Promise<void>` → `Promise<string | undefined>`), and `pullInitialization` queues it via the **already-shipped** `queueFrameworkContext` → `withPendingFrameworkContext` path so it prepends to the session's first outbound user message, gated parent-only. Additive, backward-compatible, no new machinery, no new trust surface.

## Findings (sorted; all non-blocking)

**LOW · high conf · test-coverage · `framework-context-queue.test.ts` (whole file) · ref:3f69d6c · file-state**
The queue-before-`initPromise`-resolves ordering invariant has no *dedicated* regression test — only transitive coverage via the happy-path delivery test (`:138`). A future refactor that moved the queue push after the `while` loop would still pass unless the delivery test's timing caught it.
*Evidence:* `grep 'initPromise|ordering|race' framework-context-queue.test.ts` → 0 matches.
*Suggestion (non-blocking):* add a test asserting the push precedes `initPromise` resolution, or interleaving `sendMessageStream` against a slow init.

**LOW · high conf · test-coverage · `hooks-dispatch.test.ts:42-48` + `framework-context-queue.test.ts` · ref:3f69d6c · file-state**
Block-throw is tested at the `dispatchSessionStart` unit level but not at the `AgentSession` init level (the `pullInitialization` catch → `rejectInitializationOnce` at `agent-session.ts:396-400`). No test drives `new AgentSession({ blocking SessionStart hook })` and asserts the first `sendMessageStream` rejects.
*Suggestion (non-blocking):* one AgentSession-level test that a blocking SessionStart fails init cleanly.

**LOW · medium conf · perf-observability · `agent-session.ts:640` + `:356` · ref:3f69d6c · file-state**
The witness trace records `injectedContextBytes` at SessionStart **dispatch** (`hooks-dispatch.ts:63-70`), but the **drain** that actually prepends to the first message (`:640`) emits no trace event — the push only hits `debugLog` (`:1097`). An operator sees "N bytes injected at SessionStart" but not "delivered on turn 1." Delivery is inferable from the recorded user message (`ledger.recordUser` at `:655`), so not fully dark — hence low.
*Suggestion (non-blocking):* optional trace event at drain time.

## Non-finding observations (confirmed correct, surfaced for the record)

- **Parent-only gate robustness.** The gate uses `parentSessionId === undefined` as the "is-a-subagent-fork" proxy. This is correct for every production fork path: `subagent.ts:740-741` sets the child's `parentSessionId` from `parent.sessionId`, and the `agent`-tool executor (`subagent-executor.ts:674`) forks inside a live parent turn (post-init), so `parent.sessionId` is always populated. It's a value-timing dependency rather than a structural invariant, but no reachable path leaves it undefined — **not a defect in this PR** (pre-existing fork wiring; `hooks-integration.test.ts:470-481` asserts it).
- **api-compat.** Return-type widening is purely additive — the sole production caller (`pullInitialization`) is the one added in this diff; every other `dispatchSessionStart` reference is a test or doc-comment. Ignoring a `void` return never breaks. New optional `SessionStartContext.parentSessionId` is additive (no exhaustive/positional construction), mirroring `SessionEndContext`.
- **security.** No new trust surface: `injectContext` originates only from `registry.dispatch`; SessionStart handler *registration* is gated behind `enablePluginHooks` (plugin tier) or `enableShellHooks`/`userGlobalEnabled` (`config-bridge.ts:38-40,97`). The hooks already ran pre-PR — this delivers already-produced output through the byte-identical path SubagentStop already uses. No sanitization regression (SubagentStop is likewise unsanitized), no secret/auth exposure.
- **reset()/re-init.** `reset()` clears the queue (`:314`) then rebuilds `initPromise`, re-firing SessionStart → re-queues fresh. No drop, no double-injection.
- **slash commands.** Native slash dispatch runs on raw text upstream (`loop-iteration.ts:382`) and never calls `sendMessageStream`; the prepend at `:640` is strictly downstream, so it cannot defeat a first-turn `startsWith('/')` check.

## Merge decision: **MERGE**

All stated-intent clauses are satisfied, the concurrency ordering invariant is independently verified to hold, the change reuses a proven delivery path with no new trust surface, and the only findings are three non-blocking low-severity test/observability gaps that do not affect runtime correctness.

## What was not checked
- **Refs read:** Wave 1 agents + my inline verification all read the working tree at `ref:3f69d6c` (= PR head; the sole uncommitted change is the unrelated untracked `docs/proposals/sticky-input-viewport.md`, so all 5 reviewed files are at the reviewed ref). Citations verified against that ref.
- **No test execution** — coverage assessed from test source, not an observed vitest run (the PR body claims 12128 passed / 0 failed; not independently run here).
- **Registry merge internals** (`hook-registry.ts` join-by-`\n`) treated as pre-existing shipped behavior, exercised only via its result.
- **Non-`agent`-tool fork paths** (mint/farm/background) not individually traced for pre-init fork timing — the `agent`-tool executor path was confirmed; the others thread `parentSessionId` through but their timing was not proven.
- **Stated intent:** PR #618 title + body — spec-compliance assessed against it.

## Done
- Reviewed PR #618 read-only across security, api-compat, correctness, spec-compliance, test-coverage, perf-observability via 2 parallel research-agents + inline re-derivation of the ordering invariant; no repo/PR mutation performed.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/7e1579da-67f0-42a2-ad2d-dc696c1a594f/pr-618.diff`; source read at `ref:3f69d6c` (`agent-session.ts:319,328-404,628-655,1093-1111`, `hooks-dispatch.ts:86-102`); reviewed-ref confirmed via `gh pr view 618` + `git status`.
- Verdict: **MERGE** — 0 blocking/critical/high/medium; 3 non-blocking low (invariant regression test, AgentSession-level block-throw test, drain-time trace event).
- Working tree left untouched (the review is read-only; the one pre-existing untracked doc was not modified).

---

griffinwork40 (commented):

# Code Review — PR #618: deliver SessionStart `injectContext` to the first turn

**Merge recommendation: MERGE (approve).** Reviewed read-only against branch HEAD `3f69d6c`; every load-bearing citation verified at that ref via `git show`. Regime: light (single-agent, all dimensions) — 3 production files, +168/−7.

## Findings by severity

| Severity | Count |
|----------|-------|
| critical | 0 |
| high | 0 |
| medium | 0 |
| low | 1 |
| nit | 1 |

### 🟢 Low — reset (`/clear`) → SessionStart re-delivery is untested
`reset()` re-runs `pullInitialization` → re-dispatches SessionStart → re-queues `injectContext`, so a SessionStart hook re-primes after every `/clear`. This is **correct by design** (the `initSdkLifecycle` doc at `agent-session.ts:235-238` states SessionStart hooks "fire on each cycle"), but no test pins it. The existing test (`framework-context-queue.test.ts:185-195`) only covers a *manually-queued* context being dropped by reset — orthogonal.
- *Suggestion:* add a test registering a SessionStart `injectContext` handler, calling `reset()`, then asserting the first post-reset message IS re-prefixed.

### ⚪ Nit — `dispatchSessionStart` return-type widening
`Promise<void>` → `Promise<string | undefined>` (`hooks-dispatch.ts:86-95`). Source-compatible widening; the sole non-test caller (`agent-session.ts:328`) is updated in this PR, and existing void-discarding callers remain valid. No action needed. [UNVERIFIED: no production importers beyond the updated call site.]

## Dimension summary
- **Correctness** ✅ — Ordering invariant holds: `pullInitialization` queues the context (`agent-session.ts:356-357`) *before* the init loop resolves `initPromise`; `sendMessageStreamInternal` awaits `initPromise` *before* draining via `withPendingFrameworkContext`, so the queued context is guaranteed to ride the first outbound message. Block/throw path preserved (`hooks-dispatch.ts:96-101`) — a blocking handler throws before any queue call. Empty/whitespace context is dropped by the truthiness check at line 356 + `queueFrameworkContext`'s trim guard.
- **Spec-compliance** ✅ — Every clause of the stated intent has an implementing change; no scope creep. Critically, the **parent-only scoping** (from follow-up commit `c35247a`) is present at the reviewed ref: `if (sessionStartInjectContext && this.config.parentSessionId === undefined)` (`agent-session.ts:356`). This is **load-bearing, not decorative** — forked subagents receive a defined `parentSessionId` (`subagent.ts:741`) *and* the bubbled parent hook registry (`subagent.ts:561,780`), so without the guard every fork would re-deliver the parent's SessionStart context (token cost × fan-out + contamination of narrowly-scoped tasks). `parentSessionId` is also threaded into the SessionStart hook context (`agent-session.ts:333`) so handlers can self-skip.
- **API-compat** ✅ — Return-type widening is source-compatible; `SessionStartContext.parentSessionId?` is an additive optional field — no construction site breaks. Sole production caller updated.
- **Test coverage** ✅ (strong) — Return contract (`hooks-dispatch.test.ts`: undefined-registry / no-inject / single / `\n`-merge / block) and end-to-end delivery (`framework-context-queue.test.ts`: parent-delivers / no-context-unprefixed / fork-skips) cover all branches of the line-356 predicate. Gap: the reset re-delivery above.
- **Security** ✅ — No new trust boundary. `injectContext` rides the already-trusted, default-OFF plugin-hook channel (`enablePluginHooks`, unchanged and out of scope). The block path cannot leak context. Byte accounting uses `Buffer.byteLength`, not string length.
- **Perf/observability** ✅ — The parent-only gate IS the token-cost mitigation on fan-out. Trace still emits injected byte count (`hooks-dispatch.ts:63-70`).

## What was not checked
- Read against branch HEAD `3f69d6c` — all `file:line` citations verified at that ref (production files re-read; the parent-only gate independently re-derived by the maintainer session via `git show` before posting).
- **Tests were not executed** (read-only review): the PR's "12128 passed" claim was not independently re-run — test *presence and shape* were verified by reading, not by running.
- Did not audit the `enablePluginHooks` gate or plugin `hooks.json` discovery (explicitly out of scope, unchanged by the diff); did not exercise a live provider (delivery verified against the mock provider fixture + static reading of `sendMessageStreamInternal`).
- Stated intent: PR #618 title + body — spec-compliance assessed clause-by-clause.

---
_Read-only review; no files, branches, or PR metadata were modified. This is the only automated action on this PR this cycle._

PR #620, opened 2026-07-15
fix(telegram): stop long replies being silently truncated mid-sentence

griffinwork40:
## Problem

Long Telegram bot replies sometimes end **mid-sentence** with no error shown (e.g. a reply that just stops at "Want me to open an"). The truncation is **not** in the chunk-splitter (`splitLongMessage`) — a normally-completed turn delivers the full answer. It comes from the delivery layer in `src/telegram/streaming.ts`. Two distinct, length-correlated causes:

### 1. Full-answer delivery was gated on the `done` event
Interactive replies use `cleanFinal: true` (`handlers/message.ts`), and the robust re-delivery `deliverClean(answerText)` ran **only** inside the `event.type === 'done'` branch. If the provider stream ended any other way — the async iterator returned without ever emitting `done`, or an early break — `deliverClean` never ran and the post-loop fallback only shipped `chunks[1..]`. The user was stranded on the **in-place live preview**, which is edit-throttled (`EDIT_THROTTLE_MS`) and subject to Telegram *edit* flood-control, so it can freeze mid-stream and show **less** than what actually streamed.

### 2. A flood-control 429 mid-delivery dropped the tail silently
A long reply fans out into several back-to-back `ctx.reply` sends. If Telegram returned a flood-control **429** partway through, `deliverClean` re-threw it; `processOne`'s catch classified it via `isTelegramTransportError` and handled it in an **empty branch** — no user-facing message, remaining chunks dropped, and **no `retry_after` / backoff** anywhere in the send path.

## Fix (`src/telegram/streaming.ts`)

1. **Finalize on any non-terminal exit** — when the loop exits without a terminal `done` event, re-deliver the full `answerText`/`accumulated` as fresh message(s) via `deliverClean` and delete the frozen preview, instead of leaving stale partial preview content.
2. **`replyWithFloodRetry` helper** — retries 429s up to `MAX_FLOOD_RETRIES`, honoring the server's `retry_after` (bounded by `MAX_RETRY_AFTER_MS`); non-429 errors propagate immediately. Used by `deliverClean` and the legacy overflow path.
3. **Never drop silently** — if delivery still fails after retries (or another transport error), the chunks already sent stand and a **visible** truncation notice is posted (`DELIVERY_TRUNCATED_NOTICE`) rather than vanishing.

All additive and backward-compatible; no behavior change on the happy path.

## Tests (`src/telegram/streaming.test.ts`)
- `no 'done' event`: delivers the full streamed answer as a fresh message and removes the frozen preview.
- `cleanFinal`: surfaces a visible truncation notice (not silent) when Telegram refuses a chunk mid-delivery.
- `replyWithFloodRetry`: retries a 429 honoring `retry_after` → succeeds; throws after exhausting retries; propagates a non-429 immediately (injected `sleep`, no real waits).

## Verification
- `pnpm test src/telegram/` → 603 passed
- `pnpm lint` (tsc --noEmit, strict) → clean
- `pnpm audit:sdk:check` / `pnpm audit:env:check` → OK (no new SDK imports, no `process.env` reads)

## Not covered / out of scope
- The **upstream** trigger for a specific mid-stream stop (why a provider stream ends without `done` — provider close, detached-turn race, etc.) is separate; this PR makes the delivery layer robust to it so nothing that streamed is lost or dropped without notice.
- The out-of-band push path (`push.ts` `send_telegram` / daemon notifications) has an analogous raw-`push()` 4096 slice; left as a follow-up to keep this PR scoped to the interactive-reply path.

### DIFF
--- a/src/telegram/streaming.test.ts
+++ b/src/telegram/streaming.test.ts
@@ -435,6 +435,40 @@ describe('streamResponse', () => {
     expect(replies.some((r) => r.startsWith('first '))).toBe(true);
     expect(replies.some((r) => r.includes('Telegram dropped'))).toBe(true);
   });
+
+  it('cleanFinal: does NOT delete the frozen preview when the FIRST clean chunk fails (never zero content)', async () => {
+    // Regression (PR #620 review, High): deliverClean posted the truncation notice
+    // and returned on a transport error, but both callers deleted the live preview
+    // unconditionally. So if the VERY FIRST fresh chunk failed, the user was left
+    // with the preview gone AND zero answer content. deliverClean now reports
+    // whether anything landed; the preview must survive when nothing replaced it.
+    const { ctx, replies, deletes } = makeCtx();
+    const transportError = new TelegramError({ error_code: 400, description: 'Bad Request: message is too long' });
+    let htmlReplies = 0;
+    (ctx.reply as ReturnType<typeof vi.fn>).mockImplementation(async (text: string, extra?: { parse_mode?: string }) => {
+      if (extra?.parse_mode === 'HTML') {
+        htmlReplies++;
+        // Reply #1 is the live-preview creation (must succeed so a preview exists);
+        // reply #2 is deliverClean's FIRST fresh chunk — fail it so nothing lands.
+        if (htmlReplies >= 2) throw transportError;
+      }
+      replies.push(text);
+      return { message_id: replies.length, text, chat: { id: 12345, type: 'private' as const }, date: 0 };
+    });
+    const answer = 'Complete answer that must survive as the frozen preview.';
+    const session = makeSession(async function* () {
+      yield { type: 'chunk', chunk: { type: 'content', content: answer } };
+      yield { type: 'done', metadata: undefined };
+    });
+
+    // Never rejects — a first-chunk transport failure is handled, not thrown.
+    await expect(streamResponse(ctx, session, 'go', undefined, { cleanFinal: true })).resolves.toBeUndefined();
+
+    // Nothing replaced the preview, so it must NOT be deleted (user keeps content)…
+    expect(deletes.length).toBe(0);
+    // …and the failure is announced, not silent.
+    expect(replies.some((r) => r.includes('Telegram dropped'))).toBe(true);
+  });
 });
 
 describe('generator finalizer cleanup', () => {
--- a/src/telegram/streaming.test.ts
+++ b/src/telegram/streaming.test.ts
@@ -3,7 +3,7 @@
  */
 
 import { describe, it, expect, vi } from 'vitest';
-import { streamResponse, StreamTimeoutError, renderSubagentFooter } from './streaming.js';
+import { streamResponse, StreamTimeoutError, renderSubagentFooter, replyWithFloodRetry } from './streaming.js';
 import { TelegramError } from 'telegraf';
 import type { Context } from 'telegraf';
 import type { IAgentSession, OutputEvent } from '../agent/types.js';
@@ -383,6 +383,58 @@ describe('streamResponse', () => {
     // No fresh-send swap, no preview deletion.
     expect(deletes.length).toBe(0);
   });
+
+  it('no `done` event: delivers the full streamed answer as a fresh message and removes the frozen preview', async () => {
+    // Regression (long-reply cutoff): if the provider stream ends WITHOUT a
+    // terminal `done` event, the clean-final delivery never ran and the user was
+    // stranded on the in-place preview — which is edit-throttled + edit-flood-
+    // controlled, so it can freeze mid-stream and show LESS than what streamed.
+    // The fix finalizes on ANY non-terminal exit: re-deliver the full answer as a
+    // fresh message and delete the preview.
+    const { ctx, replies, deletes } = makeCtx();
+    const answerNoDone = makeSession(async function* () {
+      yield { type: 'chunk', chunk: { type: 'content', content: 'Complete answer that never received a done event.' } };
+      // NO done event — the generator simply returns (provider closed the stream).
+    });
+
+    await streamResponse(ctx, answerNoDone, 'go', undefined, { cleanFinal: true });
+
+    // The full streamed text is delivered as a fresh reply, not left in the preview…
+    expect(replies.some((r) => r.includes('Complete answer that never received a done event.'))).toBe(true);
+    // …and the frozen preview was removed so the chat doesn't end on stale content.
+    expect(deletes.length).toBe(1);
+  });
+
+  it('cleanFinal: surfaces a visible truncation notice instead of silently dropping the tail on a transport error', async () => {
+    // Regression (silent tail-drop): a long reply fans out into several sends; if
+    // Telegram refuses a chunk mid-delivery (flood-control 429 past our retries, or
+    // another transport error), the old code re-threw and the handler swallowed it
+    // silently — the tail vanished with no indication. The fix keeps the delivered
+    // chunks and posts a visible notice.
+    const { ctx, replies } = makeCtx();
+    const transportError = new TelegramError({ error_code: 400, description: 'Bad Request: message is too long' });
+    (ctx.reply as ReturnType<typeof vi.fn>).mockImplementation(async (text: string, extra?: { parse_mode?: string }) => {
+      // Fail the tail chunk ("second …") — a non-parse-entities transport error, so
+      // it exercises the visible-notice branch (not the plaintext-fallback branch).
+      if (extra?.parse_mode === 'HTML' && text.startsWith('second ')) throw transportError;
+      replies.push(text);
+      return { message_id: replies.length, text, chat: { id: 12345, type: 'private' as const }, date: 0 };
+    });
+    const first = `first ${'a'.repeat(4080)}`;
+    const second = `second ${'b'.repeat(200)}`;
+    const longAnswer = [first, second].join('\n');
+    const session = makeSession(async function* () {
+      yield { type: 'chunk', chunk: { type: 'content', content: longAnswer } };
+      yield { type: 'done', metadata: undefined };
+    });
+
+    // Never rejects — a transport failure mid-delivery is handled, not thrown.
+    await expect(streamResponse(ctx, session, 'go', undefined, { cleanFinal: true })).resolves.toBeUndefined();
+
+    // The first chunk landed, and the dropped tail is announced (not silent).
+    expect(replies.some((r) => r.startsWith('first '))).toBe(true);
+    expect(replies.some((r) => r.includes('Telegram dropped'))).toBe(true);
+  });
 });
 
 describe('generator finalizer cleanup', () => {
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

## Verdict: DO NOT MERGE (one high-severity regression against the PR's own goal; net-positive change otherwise)

This is a real, well-targeted fix and a net improvement over the pre-PR silent tail-drop. The 429 retry (`replyWithFloodRetry`, honoring `retry_after`, bounded at 30s), the finalize-on-any-non-terminal-exit branch, and the visible truncation notice all check out and are tested. But one edge in the new delivery path regresses the PR's strongest claim — *"nothing that streamed is lost or dropped without notice"* — so it's worth fixing before merge.

Reviewed at ref `2275a984` (branch HEAD). Every load-bearing claim was re-read from that ref (`git show <ref>:<file>`) and cross-verified; tests were evaluated by reading, not executed (review is read-only, working tree is on `main`).

### 🔴 High — first-chunk delivery failure deletes the preview, leaving the user with *zero* answer content
`src/telegram/streaming.ts` — `deliverClean` (~L292-315) and its two call sites (done branch ~L602-611, non-terminal branch ~L659-666).

On a non-parse `TelegramError` (e.g. a persistent 429 past retries, or a 400 "message is too long"), `deliverClean` posts `DELIVERY_TRUNCATED_NOTICE` and **`return`s** — it does not throw and does not report whether any chunk landed:
```
} else if (e instanceof TelegramError) {
  await ctx.reply(DELIVERY_TRUNCATED_NOTICE).catch(() => {});
  return;
```
Both callers then delete the live preview unconditionally after `deliverClean` returns — gated only on preview existence / `full.trim()`, never on delivery success:
```
await deliverClean(answerText);
if (sentMessage) { await ctx.telegram.deleteMessage?.(chatId, sentMessage.message_id).catch(() => {}); … }
```
So if the **first** chunk fails, the user ends with the preview gone and only the notice — no answer content at all. Pre-PR, `deliverClean` re-threw before the delete ran, so the user at least kept the frozen partial preview. The notice makes it non-silent, but the streamed content the intent promises to preserve is deleted.
**Fix:** have `deliverClean` return a `delivered: boolean` (true once ≥1 chunk lands) and gate both `deleteMessage` calls on `delivered === true`, so the preview survives when nothing was posted.

### 🟡 Medium — legacy non-`cleanFinal` overflow branch can still drop the tail silently
`src/telegram/streaming.ts` ~L667-673 (`else if (accumulated && preview)`). This branch sends `chunks[1..]` via `replyWithFloodRetry` in a bare loop with no `TelegramError` → notice fallback, so a non-429 transport error on an overflow chunk throws uncaught out of the try (into the finally) — reproducing exactly the silent-tail-drop this PR set out to kill, for the non-`cleanFinal` `done` path.
**Fix:** route this loop through `deliverClean` (or wrap it in the same `TelegramError` → notice handling).

### 🟢 Low — `e.description` not nullish-guarded in the parse-entities check
`src/telegram/streaming.ts` ~L301: `/can't parse entities/i.test(e.description)`. A `TelegramError` with `code === 400` and `description === undefined` throws `TypeError` inside the catch, escaping `deliverClean`. The sibling `floodRetryAfterMs` already uses the safe form `e.description ?? ''`.
**Fix:** `/can't parse entities/i.test(e.description ?? '')`.

### 🟢 Low — no test for the first-chunk-failure / preview-deletion case
`src/telegram/streaming.test.ts` — the added transport-error test fails the **second** chunk (`text.startsWith('second ')`) and never asserts on `deletes`, so the High finding's scenario (first chunk fails → preview deleted → zero content) is uncovered. Confirmed absent via grep + diff cross-check.
**Fix:** add a test where the first HTML reply throws a persistent non-parse `TelegramError` with a live preview present, asserting the preview is NOT deleted once the High finding is fixed.

### What was clean
- **api-compat:** `replyWithFloodRetry` is a new additive export with zero external importers (only `streaming.ts` + its test reference it); `streamResponse` signature unchanged.
- **security:** no secrets/auth/injection surface; `retry_after` is bounded (`MAX_RETRY_AFTER_MS` = 30s), so a hostile value can't cause an unbounded sleep.
- **Happy path:** unchanged — `replyWithFloodRetry` calls `ctx.reply` exactly once on success; the retry loop is inert absent a 429.

### What was not checked
Read entirely against ref `2275a984`; citations verified at that ref. Tests were **not executed** — the PR's "603 passed / lint clean" claim is unverified here; test presence and logic were checked by reading. `markdownToTelegramHtml` and `splitLongMessage` (in `formatter.ts`) are outside the diff and were only read for adjacent semantics. The premise that the in-place preview freezes under edit-throttle / edit-flood-control is taken from the code comments, not independently re-derived. The out-of-band `push.ts` path (noted out-of-scope by the PR) was not reviewed.

griffinwork40:
Confirmed the test-coverage absence claim: every throwing-`ctx.reply` mock (lines 348, 416, 448) is paired with `{ cleanFinal: true }`; the only non-cleanFinal path (line 378) uses a never-throwing reply. The legacy overflow catch branch (streaming.ts:699-710) is genuinely untested.

Both load-bearing findings independently verified against the ref. No critical/high findings → shadow-verify not required per skill. Synthesizing.

---

# PR #620 Review — `fix(telegram): stop long replies being silently truncated mid-sentence`

**Target:** PR #620 `fix/telegram-long-message-truncation → main` · reviewed ref `dcfd300e` · hotfix · 2 files, 301 lines · regime: full (2 dimension agents + inline Wave 1.5 verification)

**Stated intent:** PR title+body — spec-compliance **assessed**. Two length-correlated delivery-layer truncation bugs: (1) full-answer re-delivery gated only on the `done` event; (2) a flood-control 429 mid-delivery dropped the tail silently.

---

## Findings (sorted: critical → high → medium → low → nit)

### 🟡 medium

**`medium · high · correctness · src/telegram/streaming.ts:690-711 · ref:dcfd300e · citation-type:file-state`**
The legacy `else if (accumulated && preview)` overflow branch **re-fires after a `cleanFinal` `done` whose first `deliverClean` chunk failed**, re-delivering a headless `chunks[1..]` fragment of the raw `accumulated` buffer immediately after the "dropped… ask me to resend" notice — self-contradictory, noisy output on long answers.
- Trace: first chunk fails → `delivered=false` → line 624 `if (delivered && sentMessage)` skips → `sentMessage` stays non-null, `break`. Post-loop, line 678 `if (preview && !sawTerminalEvent)` is skipped (`sawTerminalEvent=true`), so line 690 `else if (accumulated && preview)` fires and ships `chunks[1..]`.
- Evidence:
  ```
  678:  if (preview && !sawTerminalEvent) {        // skipped: sawTerminalEvent=true after done
  690:  } else if (accumulated && preview) {        // FIRES: delivered=false left preview non-null
  692:    if (chunks.length > 1) {                  // long answer → delivers chunks[1..]
  ```
- Reachable but narrow: needs interactive `cleanFinal` + long answer (>1 chunk) + transport failure past 2 retries on the **first** chunk. **Not data loss** (over-delivery), and strictly better than the pre-PR silent drop. The author's own comment at 626-628 ("Null the preview ref so the post-loop overflow block … is skipped") shows the intent was to skip the legacy block after clean delivery — this failure path was overlooked.
- **Suggestion:** add `!sawTerminalEvent` (or `!cleanFinal`) to the line-690 guard so a failed cleanFinal delivery cannot trigger the legacy overflow re-send.

### 🟢 low

**`low · high · test-coverage · src/telegram/streaming.ts:699-710 · ref:dcfd300e · citation-type:file-state`**
The legacy **non-cleanFinal** 429/transport catch branch (posts `DELIVERY_TRUNCATED_NOTICE` for the overflow path) has **zero coverage** — grep-verified at ref: every throwing-reply test (`streaming.test.ts:348,416,448`) passes `{ cleanFinal: true }`; the sole non-cleanFinal multi-chunk call (`:378`) uses a never-throwing reply. **Suggestion:** add a non-cleanFinal test with a >1-chunk `accumulated` where `ctx.reply` throws a `TelegramError` on `chunk[1]`, asserting the notice posts and nothing rethrows.

**`low · high · test-coverage · src/telegram/streaming.ts:684 · ref:dcfd300e · citation-type:file-state`**
The non-terminal-exit finalize branch is tested only for a **successful** `deliverClean` (`:387-406` no-`done` happy path); the failing-first-chunk finalize path (no-`done` + transport error → `delivered=false` → preview survives) is untested. **Suggestion:** add a no-`done` + first-chunk-failing variant asserting `deletes.length === 0`.

**`low · medium · test-coverage · src/telegram/streaming.test.ts:439-471 · ref:dcfd300e · citation-type:file-state`**
The first-chunk-fail-preserves-preview test uses a **short** (single-chunk) answer (`'Complete answer that must survive…'`), so it never exercises the long-answer case where the medium finding manifests. **Suggestion:** add a long-answer variant (`accumulated` > one Telegram message).

**`low · low · api-compat · src/telegram/streaming.ts:297 · ref:dcfd300e · citation-type:file-state`**
`deliverClean` return type widened `Promise<void> → Promise<boolean>` `[UNVERIFIED: no production importers]` — function-local `const`, zero cross-file callers (grep: only in-file sites `:623/:684`; `push.ts` hits are doc-comments). Additive. `[low confidence — verify with runtime context]`

**`low · low · api-compat · src/telegram/streaming.ts:100 · ref:dcfd300e · citation-type:file-state`**
`replyWithFloodRetry` newly exported `[UNVERIFIED: no production importers]` — zero cross-file production importers (in-file calls + test import only). Additive, non-breaking. `[low confidence — verify with runtime context]`

### Clean dimensions
- **security** — `no issues found — read streaming.ts at ref dcfd300e`. Retry loop bounded (`MAX_FLOOD_RETRIES=2`, each sleep `Math.min(…, MAX_RETRY_AFTER_MS=30_000)`; worst case ~60s/msg, chunk loop bounded by finite text). `DELIVERY_TRUNCATED_NOTICE` is a static string — no `TelegramError` interpolation, no secret/error leakage. No new injection surface (same `markdownToTelegramHtml` + `parse_mode:'HTML'` path).
- **spec-compliance** — every stated clause has a verified implementing change (finalize-on-non-terminal-exit `:678-689`; `replyWithFloodRetry` w/ `retry_after` + cap; used by `deliverClean` + legacy path; visible notice; boolean return gating deletion). **No unmet intent.** The only wrinkle is the medium's scope-interaction.
- **perf-observability** — clean. Bounded sleeps; the `.catch(() => {})` on notice sends (`:706`) is intentional best-effort on an already-degraded path, not a silent swallow of primary delivery.
- **correctness (b)–(f)** — sound: no double-delivery via the finalize branch (gated `!sawTerminalEvent`); `delivered`-gating correct at both call sites; `floodRetryAfterMs` NaN guard correct (`Number(undefined)→NaN`, `Number.isFinite`→false→default); the `as Message.TextMessage | null` CFA cast is a sound no-op re-anchor.

*No cross-agent conflicts.*

---

## Summary

| | |
|---|---|
| **Blocking (critical/high)** | none |
| **Non-blocking** | 1 medium (correctness/UX edge), 4 low (test gaps), 2 low (api-compat, additive) |
| **Core fixes** | both stated bugs correctly implemented + verified against ref |
| **Backward-compat** | additive; the only production-reachable export (`streamResponse`) is unchanged in signature/contract |

The PR does what it claims: it eliminates the silent mid-sentence truncation on both the missing-`done` path and the flood-control-429 path, and correctly preserves the live preview when zero content lands. The residual **medium** is a narrow, non-lossy, messy-UX interaction (a contradictory partial re-send after the truncation notice) on a rare transport-failure path, closable with a one-line guard at line 690. The test gaps are for edge paths; the three primary bug scenarios are each tested.

Non-blocking. Recommend the maintainer address the medium (one-line `!sawTerminalEvent` guard at streaming.ts:690) and add a non-cleanFinal legacy-branch test before or shortly after merge — neither blocks.

## What was not checked
- **Ref:** Wave 1 agents + inline verification read against branch HEAD `dcfd300e` — the two load-bearing findings (medium correctness, legacy-branch test gap) were re-verified at that ref by me directly. Working tree is DIRTY and sits at a **pre-PR commit** (does not contain this PR's code), so **no working-tree citations were used** — all quotes are `file-state@dcfd300e` or `diff-context`.
- **Stated intent:** PR #620 title+body — spec-compliance **assessed**.
- **Not run:** the test suite (read-only skill) — the medium's runtime manifestation is reasoned from control flow, not executed. PR body claims `pnpm test src/telegram/` → 603 passed.
- **Not reviewed:** the `push.ts` / daemon out-of-band send path (explicitly out of scope per the PR body); GitHub-side PR state beyond `headRefOid`.

This is a read-only review — no files, commits, or PR state were modified.

---

Issue #541, opened 2026-07-12
test-infra: real-PTY (node-pty + xterm) CI harness so compositor scrollback is verifiable


griffinwork40:
## Problem

The compositor's scrollback behavior — the single hardest-to-debug part of the TUI — has **no automated test that verifies real-terminal behavior**. `docs/scrollback.md:9-13,108-111` is explicit: mock / `@xterm/headless` tests confirm bytes were *written* but not that they *reached scrollback*, which is a property of the real PTY's scroll engine. Documented consequences:

- Two prior fixes (`6d7cb90`, `9690b9f`) **shipped green and were broken in reality.**
- The #539 collapse-void fix required **manual iTerm2 A/B** validation (`scripts/visual-void-repro.ts`) because the suite can't certify it.
- The whole ~13-PR gap-class (see #540) is eyeball-verified, not CI-verified.

`docs/scrollback.md:628-631` already proposes the remedy and notes *"None exists today."*

## Proposal

Add a `node-pty` + `xterm.js`(-headless) integration harness that:
- drives `TerminalCompositor` (or agent-afk) through a **real PTY**,
- pipes output into an xterm emulator instance, and
- asserts committed lines land in the emulator's **scrollback** buffer (not just that bytes were written), with no void / duplication / loss across the gap-class geometries.

## Why it matters

This is the missing safety net for #540 (the fuller commit-model unification) and every future compositor change — it converts "green but maybe broken" into "green means correct," and removes the human-in-the-loop requirement for scrollback regressions.

## Acceptance criteria
- [ ] `node-pty` dev-dep + a `test:pty` harness driving a real PTY into xterm.
- [ ] Assertions read the emulator **scrollback** buffer, not raw bytes.
- [ ] Covers at least collapse-void (#539), multi-commit-gap, overflow-gap, shrink-gap, first-turn image echo (#509).
- [ ] Wired into CI (may be a separate/quarantined job — PTY can be flaky; retry as needed).

## References
`docs/scrollback.md:9-13,108-111,628-631` · #539 (fix that needed manual validation) · #540 (unification that needs this net) · #509 (double-render, same path)

Issue #540, opened 2026-07-12
TUI compositor: fuller A2 unification (render-not-repin + single end-of-turn flush) to dissolve the scrollback-gap class


griffinwork40:
Follow-up to #539, which shipped the **minimal A2 core** (retain committed runs against `maxBandModel` instead of the tall-frame room) and fixed the reported end-of-turn table/report void. This issue tracks the **optional fuller unification** that dissolves the whole gap-class rather than the one instance #539 fixed.

Not urgent — #539 fixes the user-visible bug. This is robustness/architecture hardening.

## Why

The compositor has three overlapping mechanisms for getting committed content on screen — eager fits-path scroll, band-hold retention, and band re-pin. Their interaction is the source of a recurring failure family. ~12 regression tests circle it: `shrink-gap`, `scrollback-gap`, `overflow-gap`, `multi-commit-gap`, `commit-collapse-wrap-gap`, `band-hold-perline-gap`, `endturn-overflow-gap`, `banner-commit-gap`, `wrap-gap`, `pad-decay-straddle`, `two-table-overflow`. #539 rebalanced the routing; this collapses the three mechanisms into one so the seam can't be mismanaged.

## Scope

- **Stage 2 — render, don't re-pin.** Replace `repositionCommittedBand`'s incremental reposition with a stateless re-render of the visible window (`turnModel.slice(-room)`) from a single retained model, painted fresh each repaint at current geometry via the existing CUP `eraseAndPaintRow` primitive. Gap-free by construction; removes the `committedBandBottomRow`-adjacency coupling read at ~10 sites (`committed-band-commit.ts:191,607-608,687`).
- **Stage 3 — single end-of-turn flush.** Flush the retained model to native scrollback exactly once, contiguously, at turn finalization (geometry stable) — the only scrollback write besides genuine mid-turn overflow.

Full design + root-cause analysis: `docs/proposals/tui-compositor-rewrite.md`.

## Hard constraints (any implementation must respect)

- **Scrollback is append-only via full-screen scroll** — a committed line's row can never be repositioned (`docs/scrollback.md:20-67`).
- **The input frame is unconditionally bottom-pinned** on purpose (dropdown/picker headroom — `frame.ts:230`, `bottom-pin.test.ts:63`). The "float the region like Ink" model (A1) is rejected for this reason.
[...]

PR #621, opened 2026-07-15
test(tui): real-PTY width-resize harness + RED axis-2 fragmentation guards (#540, #541)

griffinwork40:
## What

PR 1 of the TUI width-resize fix. Extends the real-PTY scrollback harness (issue #541) with a **mid-stream width resize**, so the width-resize scrollback-fragmentation class (#540 "axis 2" — the ragged/fragmented wrapping in the reported screenshot) is reproducible in CI, and adds guards that lock the bug until the fix lands.

**This PR ships test infrastructure + a failing-behavior guard only. No `src/` changes, no behavior change.**

## Why (the diagnosis, verified)

Committed content is hard-wrapped to the width-at-commit (`src/cli/terminal-compositor.committed-band-commit.ts:165`, `hardWrapToWidth(l, cols)`) and flushed into native scrollback as **hard-newline rows**. A terminal only reflows its own soft-wraps on resize — never app hard-newlines (VT100/DECAWM) — so once content is in native scrollback a later width change re-fragments it (narrow) or leaves it ragged (widen). Confirmed terminal-independent (tmux + Terminal.app + Ghostty) and **cosmetic — no data loss** (`tmux capture-pane -J` rejoins every char). The prior merged work (#539/#552/#573) fixed the *height*-event void class (axis 1); this width class is not covered by it.

## How the harness resize works

- **`constants.ts`** — `PTY_RESIZE` APC marker protocol (`buildResizeMarker`/`findResizeMarker`). A scenario emits the marker at the point it wants the geometry changed, carrying `<cols>x<rows>`; xterm ignores APC, so it never perturbs the buffer.
- **`harness.ts`** — `IPtyProcess.resize()`; `runScenarioInPty` watches the byte stream, on the first marker calls `child.resize()` (SIGWINCH) and records the byte offset, then on replay **splits the stream**: construct the emulator at the *old* geometry → write pre-bytes → `emulator.resize()` (which reflows) → write post-bytes. Building at the new width and writing old-width bytes would not model reflow. Also exposes per-line `isWrapped`.
- **`scenarios.ts`** — `requestResize()` driver helper (emit marker, await the driver's own `'resize'`); `PtyExpect.logicalSpan` asserts soft-wrap rejoinability (the `tmux -J` property) via `isWrapped` counts, plus `minSpanRows` to prove a resize actually fired.

## Scenarios

Existing 5 scenarios are **unchanged** — the resize is marker-driven, so `runScenarioInPty({name, cols, rows})` keeps its signature.

- `width-resize-reflow-sanity` — **GREEN and stable across the fix.** A line that fits at the wide width must re-wrap to 2 rows when narrowed; certifies the resize handshake actually fires (a silently no-op'd `child.resize` fails here).
- `width-resize-fragment-narrow` / `width-resize-fragment-widen` — **GREEN now, RED guards.** A logical line committed at one width, flushed to scrollback as hard rows, shows **≥2 non-wrapped rows** after a resize (not `tmux -J`-rejoinable). Verified genuine: asserting the fixed property (`maxNonWrappedRows: 1`) fails today with `expected 2 to be <= 1`.

**The flip:** when PR 2 (#540 Stage 3 — the single end-of-turn flush emitting *logical* soft-wrappable lines) lands, the count drops to 1 and these assertions start failing — the signal to flip `minNonWrappedRows` → `maxNonWrappedRows: 1`, converting them to GREEN regression guards.

## Gates

- `pnpm test:pty` — **8/8 green** (5 existing + 3 new)
- `pnpm lint` (`tsc --noEmit`, strict) — clean

## Not in this PR

PR 2 — the fix itself (retain logical form through to a single end-of-turn logical flush; ~1,200 LOC in `committed-band-commit.ts` / `committed-band-repin.ts` / `frame-preserve.ts`). A Stage-0 spike confirmed the crux is *retaining* the pre-wrap logical form (the flat `committedBand: string[]` cannot be reconstructed — `reflowBandSplit` re-wraps each row independently, so widening never rejoins). PR 2 is gated on a human real-terminal matrix (iTerm2 / Terminal.app / Ghostty) — prior fixes shipped green and were broken in reality.

Refs #540, #541.

### DIFF
--- a/tests/pty/compositor-scrollback.pty.test.ts
+++ b/tests/pty/compositor-scrollback.pty.test.ts
@@ -82,6 +82,37 @@ function assertExpectations(res: PtyRunResult, exp: PtyExpect): void {
       ).toBeLessThanOrEqual(exp.maxViewportBlankRun);
     }
   }
+  if (exp.logicalSpan) {
+    const { from, to, minNonWrappedRows, maxNonWrappedRows, minSpanRows } = exp.logicalSpan;
+    const fromIdx = firstIndex(res.lines, from);
+    expect(fromIdx, `logicalSpan.from "${from}" not found:\n${dump}`).toBeGreaterThanOrEqual(0);
+    const toIdx = res.lines.findIndex((l, i) => i >= fromIdx && l.includes(to));
+    expect(toIdx, `logicalSpan.to "${to}" not found at/after "${from}":\n${dump}`).toBeGreaterThanOrEqual(fromIdx);
+    const spanRows = toIdx - fromIdx + 1;
+    if (minSpanRows !== undefined) {
+      expect(
+        spanRows,
+        `logical span ["${from}".."${to}"] should occupy >=${minSpanRows} rows (proves the resize took effect) but had ${spanRows}:\n${dump}`,
+      ).toBeGreaterThanOrEqual(minSpanRows);
+    }
+    // Count rows in the [from..to] span that are NOT soft-wrap continuations.
+    // One = the terminal reflowed a single logical line cleanly (tmux -J rejoins
+    // it); ≥2 = interior app hard-newlines fragmented it (the axis-2 bug).
+    let nonWrapped = 0;
+    for (let i = fromIdx; i <= toIdx; i++) if (!(res.wrapped[i] ?? false)) nonWrapped += 1;
+    if (minNonWrappedRows !== undefined) {
+      expect(
+        nonWrapped,
+        `logical line ["${from}".."${to}"] should have >=${minNonWrappedRows} non-wrapped (hard-break) rows — the axis-2 fragmentation RED guard (#540); found ${nonWrapped}. If this dropped to 1 the flush now emits logical lines: flip minNonWrappedRows -> maxNonWrappedRows: 1.\n${dump}`,
+      ).toBeGreaterThanOrEqual(minNonWrappedRows);
+    }
+    if (maxNonWrappedRows !== undefined) {
+      expect(
+        nonWrapped,
+        `logical line ["${from}".."${to}"] should rejoin to <=${maxNonWrappedRows} non-wrapped row(s) but had ${nonWrapped}:\n${dump}`,
+      ).toBeLessThanOrEqual(maxNonWrappedRows);
+    }
+  }
 }
 
 describe('TerminalCompositor scrollback over a real pty (issue #541)', () => {
--- a/tests/pty/constants.ts
+++ b/tests/pty/constants.ts
@@ -11,3 +11,53 @@
  * bytes BEFORE this marker, so it never perturbs the geometry under test.
  */
 export const PTY_DONE_SENTINEL = '\x1b_AFK_PTY_DONE\x1b\\';
+
+// Invariant (resize handshake — the parent owns the winsize, the driver owns
+// the timing): a scenario that wants a MID-STREAM width resize emits this APC
+// marker at the exact point it wants the geometry changed, carrying the target
+// `<cols>x<rows>`. Like the DONE sentinel it is an APC string, so xterm emits no
+// glyphs and no scroll for it. The parent watches the byte stream, and on the
+// FIRST complete marker (a) calls node-pty `child.resize(cols, rows)` — which
+// SIGWINCHes the driver so its own `process.stdout` 'resize' fires and the
+// compositor re-renders — and (b) records the marker's byte OFFSET so replay
+// can split the captured stream into a pre-resize half (written at the OLD
+// geometry) and a post-resize half (written after the emulator is resized).
+// Faithful reflow requires constructing the emulator at the OLD width, writing
+// the pre-bytes, calling emulator.resize(), THEN writing the post-bytes — a
+// terminal reflows scrollback on resize, so building at the new width and
+// writing old-width bytes would NOT model the real width-change under test.
+const PTY_RESIZE_PREFIX = '\x1b_AFK_PTY_RESIZE:';
+const PTY_RESIZE_SUFFIX = '\x1b\\';
+
+/** Build the resize-handshake marker for a `<cols>x<rows>` target geometry. */
+export function buildResizeMarker(cols: number, rows: number): string {
+  return `${PTY_RESIZE_PREFIX}${cols}x${rows}${PTY_RESIZE_SUFFIX}`;
+}
+
+/** A located resize marker: its target geometry and byte span within the stream. */
+export interface ResizeMarker {
+  cols: number;
+  rows: number;
+  /** Byte offset of the marker's first char (split point for pre-resize bytes). */
+  start: number;
+  /** Byte offset one past the marker (split point for post-resize bytes). */
+  end: number;
+}
+
+/**
+ * Find the FIRST complete resize marker in `buf`, or null if none has fully
+ * arrived yet (the marker may be split across pty read chunks — the caller
+ * scans the ACCUMULATED buffer each chunk, so a partial marker simply resolves
+ * on a later chunk). Malformed payloads (no `<n>x<n>`) are treated as absent.
+ */
+export function findResizeMarker(buf: string): ResizeMarker | null {
+  const start = buf.indexOf(PTY_RESIZE_PREFIX);
+  if (start < 0) return null;
+  const payloadStart = start + PTY_RESIZE_PREFIX.length;
+  const suffixStart = buf.indexOf(PTY_RESIZE_SUFFIX, payloadStart);
+  if (suffixStart < 0) return null; // suffix not arrived yet — try again next chunk
+  const payload = buf.slice(payloadStart, suffixStart);
+  const m = /^(\d+)x(\d+)$/.exec(payload);
+  if (!m || m[1] === undefined || m[2] === undefined) return null;
+  return { cols: Number(m[1]), rows: Number(m[2]), start, end: suffixStart + PTY_RESIZE_SUFFIX.length };
+}
--- a/tests/pty/harness.ts
+++ b/tests/pty/harness.ts
@@ -22,7 +22,7 @@ import { createRequire } from 'node:module';
 import { fileURLToPath, pathToFileURL } from 'node:url';
 import { existsSync, chmodSync, readdirSync, statSync } from 'node:fs';
 import { dirname, join } from 'node:path';
-import { PTY_DONE_SENTINEL } from './constants.js';
+import { PTY_DONE_SENTINEL, findResizeMarker, type ResizeMarker } from './constants.js';
 
 const require = createRequire(import.meta.url);
 const REPO_ROOT = fileURLToPath(new URL('../../', import.meta.url));
@@ -32,6 +32,8 @@ const DRIVER_PATH = fileURLToPath(new URL('./driver.ts', import.meta.url));
 interface IPtyProcess {
   onData(cb: (data: string) => void): void;
   onExit(cb: (e: { exitCode: number; signal?: number }) => void): void;
+  /** Change the pty winsize — SIGWINCHes the child (drives the resize handshake). */
+  resize(cols: number, rows: number): void;
   kill(signal?: string): void;
 }
 interface NodePtyModule {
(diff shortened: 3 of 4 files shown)


griffinwork40:
Verification complete. Both the empirical gate and the independent re-derivation are in:

- **`pnpm test:pty` → 8/8 green** (5 existing + 3 new, 5.13s) — the PR's central gate claim confirmed empirically by me.
- **Shadow-verify (independent, no prior context)** re-derived all three load-bearing claims from source:
  - Guard genuineness: **CONFIRMED** (3-layer mechanism: `eraseAndPaintRow` CUP+EL no-`\n`, each row ≤cols, `withAutowrapDisabled` DECAWM → `isWrapped=false`; the assertion genuinely flips to fail under a logical-flush fix).
  - Replay ordering: **CONFIRMED** (original-geometry construction, marker bytes dropped, awaited writes).
  - The HIGH finding: **CONFIRMED as fact but severity independently downgraded to LOW** — the resize path is a single direction-agnostic code block certified in the narrow direction by two `minSpanRows` siblings, and the widen scenario documents its delegation deliberately.

No critical/high findings survive verification, so no further shadow-verify pass is warranted. Emitting synthesis.

---

# Code Review — PR #621

**Target:** `test(tui): real-PTY width-resize harness + RED axis-2 fragmentation guards (#540, #541)` · `afk/tmux-test-wrapping-fix` → `main` · ref **`7a361a9`** · +303/−5, 4 files (all `tests/pty/`) · regime: full (2-agent wave + citation/absence verification + shadow-verify).

## Summary

A **clean, well-constructed test-only PR** that does exactly what its body claims: extends the real-PTY scrollback harness with a mid-stream width-resize handshake and adds two genuine RED guards (plus one GREEN self-check) for the #540 axis-2 width-resize fragmentation class. **Zero `src/` changes** (git-verified), **no consumer breakage** (all changed symbols are test-only; `logicalSpan` is optional; the 5 existing scenarios are byte-unchanged), **8/8 PTY tests green** (verified by live run), and the RED guards are **genuine** — they will flip to failing when PR 2's logical flush lands, which is their stated purpose. Findings are all **low/nit** — documentation and timing polish, none blocking.

## Findings (sorted; none critical/high after verification)

**LOW · high-confidence · test-coverage · `tests/pty/scenarios.ts` (`width-resize-fragment-widen` expect) · ref:7a361a9 · file-state**
The widen guard asserts only `{minNonWrappedRows: 2}` with no `minSpanRows`, so its assertion is satisfied by the commit-width (48) buffer alone — it does not independently prove its own resize handshake fired.
- evidence: `logicalSpan: { from: 'LOGSTART', to: 'LOGEND', minNonWrappedRows: 2 }`
- **Downgraded from HIGH by independent shadow-verify:** the replay path (`harness.ts` `if (resizeAt) { pre → resize → post }`) is a single direction-agnostic block, certified in the narrow direction by both `width-resize-reflow-sanity` (`minSpanRows:2`) and `width-resize-fragment-narrow` (`minSpanRows:3`); the scenario documents this delegation deliberately. Residual risk = a grow-vs-shrink asymmetry inside xterm's own `resize()`, which no `minSpanRows` guard would catch anyway.
- suggestion: acceptable as-is; if you want the widen path self-certifying, add a `maxSpanRows`/exact-span invariant that only holds post-widen. Not required for merge.

**LOW · high-confidence · nit · `tests/pty/scenarios.ts` (fragment-narrow + fragment-widen header comments) · ref:7a361a9 · diff-context**
Two block-comment headers state commit widths that contradict the executable config, which will mislead the PR-2 author who reads the comment to decide the flip.
- evidence: header says `committed WIDE (100 cols)` but `cols: 120`; header says `committed NARROW (50 cols)` but `cols: 48`. (The inline body comments and `requestResize` args — 120→68, 48→110 — are correct; only the prose headers drift. The narrow body comment's "char-100 hard break" is likewise off — the break is at the 120-col commit width.)
- suggestion: change "100 cols" → "120 cols" and "50 cols" → "48 cols" in the two headers.

**LOW · medium-confidence · perf-observability · `tests/pty/scenarios.ts` (`requestResize` settle) + `tests/pty/driver.ts` · ref:7a361a9 · file-state**
Post-resize sync is a fixed `settle(60)` plus the driver's fixed 60 ms flush wait; under heavy CI load a delayed SIGWINCH re-render could let the DONE sentinel bound `captured` before the resized frame's bytes land, producing a stale buffer and a spurious failure. The resize path adds a real async re-render the non-resize scenarios lack.
- evidence: `await landed; /* … */ await settle(60);`
- suggestion: none for merge (suite passed 8/8 locally in ~5s); if flake appears, gate the sentinel on an observable post-resize repaint rather than a fixed delay.

**LOW · high-confidence · correctness · `tests/pty/harness.ts` (`resizeAt` single-set) + `tests/pty/scenarios.ts` (`requestResize`) · ref:7a361a9 · file-state**
The harness models only the FIRST resize marker (`if (resizeAt === null)`); a scenario calling `requestResize` twice would silently lose the second resize. No current scenario does (all 3 call it once) — latent limitation, not an active defect.
- suggestion: add one JSDoc line to `requestResize`: "single resize per scenario — the harness records only the first marker."

**LOW (positive) · perf-observability · `tests/pty/harness.ts` (`dump()`) · ref:7a361a9 · file-state**
The `↩` soft-wrap annotation in `dump()` is correct (renders where `wrapped[i]` is true, 1:1 with `lines[]`) — a real observability improvement for fragmentation failure messages.

## Verified claims (positive attestations)

- **api-compat: no breakage** — `runScenarioInPty` signature byte-identical to parent; `IPtyProcess.resize` satisfied by node-pty@1.1.0's real `IPty` (no hand-rolled mock exists, grep-confirmed); `PtyRunResult.wrapped` constructed only inside the harness; `PtyExpect.logicalSpan` optional.
- **spec-compliance: zero scope creep** — 4 files, all `tests/pty/`, one commit, zero `src/` changes (git-verified). Diagnosis citation `committed-band-commit.ts:165 = hardWrapToWidth(l, cols)` is accurate.
- **Both RED guards genuine** — independently re-derived: hard-wrapped committed rows reach xterm as `isWrapped=false`, so `minNonWrappedRows:2` holds today and flips to fail (→1) under the PR-2 logical flush.
- **8/8 PTY tests pass** (live run, this ref).

## Merge decision: **MERGE**

Non-blocking polish only. The three-line follow-up (fix the two header comment widths; optionally add the `requestResize` single-resize JSDoc) can land in this PR or PR 2 — none of it gates merge. Nothing to fix in `src/`; the RED guards are correct and will do their job at the PR-2 boundary.

## What was not checked

- **Refs read:** Wave-1 and shadow-verify agents read against branch HEAD **`7a361a9`** (clean worktree === reviewed ref); all blocking-tier citations verified at that ref. Git file-set confirmed test-only via `git diff --name-status 8eb0396..7a361a9` (empty for `src/`).
- **Stated intent:** PR #621 title + body — spec-compliance **was** assessed against it (fully met, no creep).
- **Empirical gate:** I ran `pnpm test:pty` (8/8 green) but did **not** run `pnpm lint` (`tsc --noEmit`) — the strict-clean claim is unverified by me (types were read-checked, not compiler-run). The PR-2 "flip to failing" behavior cannot be verified without PR 2's code and was not exercised (verifying it would require mutating the assertions, out of read-only scope) — it rests on the shadow-verified `isWrapped` mechanism.
- **Not exercised:** xterm/`@xterm/headless` `isWrapped`/`resize()` reflow semantics taken as documented contract (not observed against a live emulator); node-pty SIGWINCH delivery on the CI runner OS assumed per node-pty contract; the 5 pre-existing scenarios not audited (out of scope).

This is the terminal step of the read-only review — no files, PR, or branch were modified.

---

griffinwork40 (commented):

## Verdict: MERGE

Clean, well-scoped test-infrastructure PR. The stated intent — *"ships test infrastructure + a failing-behavior guard only. No `src/` changes, no behavior change"* — is fully honored, and the RED-guard mechanics are correctly wired. No critical or high findings across security, api-compat, correctness, spec-compliance, test-coverage, or perf-observability.

Reviewed at ref `7a361a9` (branch HEAD). Every load-bearing claim was re-read from that ref (`git show 7a361a9:<file>`) and cross-verified; tests were evaluated by **reading, not executing** (review is read-only; working tree is on `main`).

### What was verified

- 🟢 **Scope — "No `src/` changes" is TRUE.** `git diff --name-only main...7a361a9` → exactly 4 files, all under `tests/pty/` (`compositor-scrollback.pty.test.ts`, `constants.ts`, `harness.ts`, `scenarios.ts`); zero `src/`. The `terminal-compositor.committed-band-commit.ts:165` reference in the scenario docblocks is a read-only diagnosis citation, not a change.
- 🟢 **RED-guard directions correctly wired (GREEN now → RED after the fix).** Both fragment scenarios assert `minNonWrappedRows: 2` (the fragmented/bug property), so they pass today and will fail when PR 2's logical flush drops the count to 1 — the intended flip signal. The sanity scenario asserts the fix-stable `maxNonWrappedRows: 1` + `minSpanRows: 2`, certifying the resize handshake actually fires without masking the RED guards. Directions are internally consistent with the documented contract.
- 🟢 **Harness mechanics correct.** `findResizeMarker` byte-offset arithmetic drops exactly the marker with no content loss (`start` = prefix first char, `end` = one-past-suffix; partial markers resolve on a later chunk). The emulator is built at the scenario's OLD geometry then `term.resize()`d to the marker's NEW geometry between pre/post writes — faithful reflow. `wrapped[]` is index-aligned with `lines[]`.
- 🟢 **`requestResize` is leak-safe.** `done` guards double-resolve; `finish()` calls both `clearTimeout` and `removeListener`; the 2s timeout resolves *without error* so a non-resize run can't hang `drive()`; it awaits the child's own `stdout` `'resize'` event.
- 🟢 **api-compat: additive only.** All changed exports (`buildResizeMarker`, `findResizeMarker`, `ResizeMarker`, `PtyRunResult.wrapped`, `PtyExpect.logicalSpan`) are new/optional; `runScenarioInPty`'s signature and the existing 5 scenarios are unchanged. `git grep` at the ref found **zero production importers** — these are test-only symbols.
- 🟢 **security: no surface.** `findResizeMarker`'s `/^(\d+)x(\d+)$/` is fully anchored and linear-time (no ReDoS), runs on a prefix/suffix-bounded payload slice, and only ever parses output from a test child the harness itself spawns and kills under a 20s timeout. No secrets, auth, or trust-boundary change.

### Non-blocking notes (nice-to-have, not merge-blockers)

- 🟢 **Low — `width-resize-fragment-widen` has no `minSpanRows`, so it can't self-certify its resize fired.** A no-op'd `child.resize` would leave the pre-widen state (4 hard rows) still asserting ≥2 non-wrapped rows, so the widen guard passes either way and leans on the sanity scenario as the shared handshake proof. This is deliberate and documented in the scenario's own comment — acceptable while the sanity gate is present, but the widen guard would silently degrade to a no-resize tautology if the sanity scenario were ever removed. *Optional:* add a widen-observable `minSpanRows` (or row-count delta) so it self-certifies independently. (`scenarios.ts`, `width-resize-fragment-widen` expect block; ref `7a361a9`.)
- 🟢 **Low — `logicalSpan` span computation relies on collision-free `from`/`to` sentinels by construction, not by code.** `toIdx = findIndex(l => i >= fromIdx && l.includes(to))` is correct for the 3 shipped scenarios (`LOGEND`/`SANITYEND` are unique), but a future `logicalSpan` reuse with a `to` substring that also appears in a `FILLER_nn`/`PAD_nn` row would land on the wrong row. Worth a one-line note near the flip PR. (`compositor-scrollback.pty.test.ts`, `logicalSpan` branch; ref `7a361a9`.)

### What was not checked

- Read entirely against branch HEAD `7a361a9` — all citations verified at that ref. Tests were **not executed**: the PR's "`pnpm test:pty` 8/8 green / `pnpm lint` clean" claim is unverified at runtime; scenario presence, registration (auto-run via `Object.entries(SCENARIOS)`), and assertion logic were checked by reading.
- The 5 pre-existing scenario **bodies** were not diffed line-by-line; `git show --stat` reports `scenarios.ts` as +176/−0 (pure additions), consistent with "unchanged," but a full-body regression diff was not run.
- PR 2 (the actual `src/` fix, out of scope here) was not reviewed. The "flip to GREEN regression guard" behavior is asserted by the docblocks and the assertion directions but can only be confirmed once PR 2 lands.
- Stated intent: PR #621 title + body — spec-compliance **assessed** against it.

griffinwork40:
Context update after the decision on #622 (now closed — we're **accepting** the width-resize scrollback limit rather than shipping the fix; rationale there).

This PR's value is unchanged and stands on its own: the real-PTY resize harness (follow-up to #541) plus scenarios that now serve as **regression guards documenting the known limitation**. The `width-resize-fragment-narrow/-widen` guards assert that wide content in native scrollback fragments on resize — i.e. they pin the *accepted* inline-CLI behavior, so the suite will flag if it ever silently changes (in either direction). The `width-resize-reflow-sanity` guard certifies the harness's resize handshake itself.

Ready for review/merge on its own merits (test-infra only, no `src/` changes, CI green).

PR #622, opened 2026-07-15
fix(tui): Stage 3 logical-line scrollback flush — dissolve width-resize fragmentation (#540)

griffinwork40:
> **Draft / stacked on #621** — base branch is `afk/tmux-test-wrapping-fix` (PR 1), not `main`. Review/merge #621 first. **Not mergeable until the Stage-4 human real-terminal matrix passes** (see bottom).

## What

PR 2 (the fix) for #540 "axis 2" — the width-resize scrollback fragmentation in the reported screenshot. PR 1 (#621) added the real-PTY resize harness and RED guards; this PR makes the fix and flips those guards to GREEN.

## The bug

Committed content is hard-wrapped to the width-at-commit (`terminal-compositor.committed-band-commit.ts:165`) and flushed into native scrollback as **hard-newline rows**. Terminals reflow only their own soft-wraps on resize, never app hard-newlines — so scrolled-off content re-fragments (narrow) / stays ragged (widen). Cosmetic, no data loss, terminal-independent.

## The fix

The end-of-turn / disarm flush now archives the pending committed band as **soft-wrappable logical lines** instead of pre-hard-wrapped physical rows, so the terminal owns wrapping and reflows scrollback cleanly on any width change — `docs/tui-resize-reflow.md` "sound pattern (a): make native scrollback authoritative."

**Mechanism — retain, don't reconstruct.** The flat `committedBand: string[]` loses logical-line boundaries (`reflowBandSplit` re-wraps each row independently, so joining rows can't recover them). So:
- `committedBandMeta: BandRowMeta[]` — per-physical-row logical provenance, index-aligned 1:1 with `committedBand`, maintained across commit / reflow / eviction / repin.
- `flushPendingCommittedBand` → `scrollbackFlushLines` maps the pending physical rows back to logical lines, reading the full band+meta so a logical line **straddling** the pending/painted boundary emits only its pending rows (no scrollback duplication — HARD CONSTRAINT #1). `buildScrollbackArchiveEscape` writes each line at the bottom margin with **autowrap ON** + trailing `\n`, so the terminal wraps and scrolls it into history.
- The **live band keeps hard-wrapped physical rows** for exact row math (unchanged); only the scrollback archival path emits logical lines.

Flips PR 1's `width-resize-fragment-narrow`/`-widen` guards to `maxNonWrappedRows: 1` — they pass because behavior changed, not because assertions were weakened (harness unchanged).

## Gates (all green, verified locally)

- `pnpm test:pty` — **8/8** (the 2 flipped guards now assert the rejoined property and pass)
- headless compositor suite — **363/363** (no regression to the #539/#552/#573 height/void class)
- `pnpm test:coverage` — green (thresholds met; includes new `terminal-compositor.logical-flush.test.ts`)

## Provenance / caveats

Implemented by a delegated subagent; the diff and every gate above were **independently re-run and reviewed** by the dispatching session before this PR was opened. Automated gates only certify the emulator model.

## REQUIRED before merge — Stage-4 human real-terminal matrix

`@xterm/headless` cannot reproduce every terminal-only DECAWM artifact, and **prior fixes in this code shipped green and broke in reality.** Before merge, run `afk` in **iTerm2 / Apple Terminal / Ghostty** and exercise: resize narrower mid-turn; widen then scroll up to old output; long report + table; overlay grow→collapse; dropdown headroom; picker. This gate is the whole ballgame — hence draft.

Refs #540. Stacks on #621.

### DIFF
--- a/src/cli/terminal-compositor.band-reflow.ts
+++ b/src/cli/terminal-compositor.band-reflow.ts
@@ -27,6 +27,7 @@
  */
 
 import { hardWrapToWidth } from './wrap.js';
+import type { BandRowMeta } from './terminal-compositor.types.js';
 
 /**
  * Invariant (painted/pending boundary survives reflow): `committedBand`'s
@@ -49,6 +50,20 @@ import { hardWrapToWidth } from './wrap.js';
 export interface BandReflowResult {
   rows: string[];
   paintedRows: number;
+  /**
+   * Per-physical-row provenance for `rows`, index-aligned 1:1 (#540 axis-2).
+   * Re-wrapping a physical row into more (narrower) or the same (wider) rows
+   * PRESERVES each row's `logicalText` — the original pre-hard-wrap line the
+   * row is a fragment of, unchanged by any width — and recomputes `isHead` so
+   * only the first sub-row of a re-split row that was itself a logical-line
+   * head stays a head. This is what lets a widened resize rejoin a
+   * hard-wrapped line at the SCROLLBACK flush sites even though the on-screen
+   * physical rows keep their (now stale) narrow break points: the logical
+   * source rides through reflow intact. Absent-meta input degrades to
+   * reconstructing `logicalText` from the physical row itself (each row treated
+   * as its own logical line) — lossy for rejoin but never for content.
+   */
+  meta: BandRowMeta[];
 }
 
 /**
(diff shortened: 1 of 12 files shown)


griffinwork40:
Closing as a deliberate product decision — **not** because the fix is broken (it passes all automated gates and reflows prose correctly).

**Why accept the limit instead of shipping this:** a devils-advocate wave + competitor research (Claude Code, Codex CLI, Aider, Gemini CLI, Ink, Bubble Tea, Warp) established that:

- Wide tables/code that have scrolled into the terminal's **native scrollback cannot be reflowed by any inline CLI** — the terminal owns that buffer. This is the universal norm: Claude Code (classic), Gemini CLI (default), Aider, and Ink all accept exactly this.
- The only tools that do better either **own the terminal** (Warp) or **retain source + destructively rewrite a bounded scrollback window** (Codex) — and Codex's approach is documented broken across xterm.js / Windows conhost / tmux, i.e. the exact cross-terminal fragility this compositor has been burned by before.
- This fix therefore wouldn't resolve the visible symptom (scrolled-off tables) anyway, and it's ~755 LOC of change to fragile teardown/scrollback code for a **cosmetic, zero-data-loss** bug.

Net: accepting the limit matches peers and avoids regression risk that would be strictly worse than the cosmetic bug.

The work is preserved — branch `afk/tui-axis2-stage3` (commit `2ce1f9b`) stays on origin and can be reopened if we ever revisit (e.g. a Codex-style bounded rewrite, or if afk ever moves to an alternate-screen model). See #621 for the real-PTY harness + guards that now document this behavior as a known limitation.

PR #625, opened 2026-07-15
chore(agents): remove inert `model:` frontmatter from vendored agent prompts

griffinwork40:
## What

Removes the `model: sonnet` line from the vendored `research-agent` and `git-investigator` agent prompts, and re-pins the drift hashes in `vendored.test.ts`.

## Why

This `model:` frontmatter is **inert for routing**:

- The vendored prompt is consumed via `vendoredPromptBody()`, which strips YAML frontmatter before it becomes the system prompt.
- The bundled agents' registered definitions in `src/agent/agents/builtins.ts` do not read frontmatter `model` for `research-agent`.

So the line implied a model pin that has no effect on this path — misleading to readers. Removing it aligns the bundled `research-agent` with the plugin copy (which carries no `model` line) and leaves the subagent model choice to the dispatcher/parent.

## Scope / caveat

- `git-investigator`'s **effective** model is still pinned via `src/skills/_agents/git-investigator.ts` (`model: 'sonnet' as const`) → `builtins.ts`. This PR removes only the inert `.md` frontmatter and does **not** change git-investigator's actual routing.
- Pure cleanup — no runtime behavior change.

## Test

`pnpm test src/skills/_agents/vendored.test.ts` → **20/20 pass** (drift hashes re-pinned for both prompts).

### DIFF
--- a/src/skills/_agents/prompts/git-investigator.md
+++ b/src/skills/_agents/prompts/git-investigator.md
@@ -1,7 +1,6 @@
 ---
 name: git-investigator
 description: Read-only git specialist. Dispatched by research-agent (or any research-shaped caller) when a finding requires git history, reflog, diff, blame, branch/remote state, or merge-base analysis. Runs git commands only — no mutations, no shell escapes.
-model: sonnet
 tools: Bash, Read, Grep, Glob
 ---
 
--- a/src/skills/_agents/prompts/research-agent.md
+++ b/src/skills/_agents/prompts/research-agent.md
@@ -1,7 +1,6 @@
 ---
 name: research-agent
 description: Read-only sub-agent for research, validation, verification, and codebase inspection. Mechanically locked to Read, Grep, Glob, WebFetch, WebSearch — cannot Edit, Write, Bash, commit, or push. Delegates git queries to `git-investigator`. Use when the dispatched task is findings-only.
-model: sonnet
 tools: Read, Grep, Glob, WebFetch, WebSearch, Agent(git-investigator)
 ---
 
--- a/src/skills/_agents/vendored.test.ts
+++ b/src/skills/_agents/vendored.test.ts
@@ -10,9 +10,9 @@ const __dirname = dirname(__filename);
 
 // Pinned hashes — guard against undocumented edits to the vendored copy
 const PINNED_HASHES = {
-  'research-agent': '141b2859797bd538f01293b5570643b5757cb7a96438cc9be0d97106e3eb92bd',
+  'research-agent': 'de1c6bedaf07f0fb7a159be3af3a13e27d31ee3c0e73dbcd34a741661f5abc04',
   contract: '0b7febafec024e8dd4404f75e84d21ee72b1b1846d6e2610aaa82ba77f9d6f2d',
-  'git-investigator': 'c31560bdb80d84c42f25938facdb883aa8120df2e6ec1d341bfb61bfac4769da',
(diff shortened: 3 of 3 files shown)


griffinwork40 (commented):

# Code Review — PR #625: remove inert `model:` frontmatter from vendored agent prompts

**Verdict: MERGE.** Small, self-contained cleanup (3 files, +2/−4). Every load-bearing claim in the PR body was re-derived from source at the reviewed ref (`1e3f3ff`), not taken on faith — all confirmed. The re-pinned drift hashes were independently recomputed and match, so `vendored.test.ts` passes (CI is green, consistent with this). No runtime behavior change.

## What was verified against source (ref `1e3f3ff`)
- **Hash pins — MATCH (both).** `vendored.test.ts` hashes raw file bytes (`readFileSync(..., 'utf8')` → `sha256`), no normalization. Recomputed `research-agent.md` = `de1c6bed…5abc04` and `git-investigator.md` = `d1f186b8…e60c03` — both equal the newly-pinned values. The drift-guard test will pass.
- **Frontmatter is stripped before it's a prompt (Claim A) — CONFIRMED.** `vendoredPromptBody()` (`builtins.ts:34-36`) returns `parseAgentMarkdown(raw).definition.prompt`; `parseAgentMarkdown` (`parser.ts:107-116`) drops the `---`-delimited block. The `model:` line never reached the system prompt.
- **research-agent never read the frontmatter `model` (Claim B) — CONFIRMED.** `builtins.ts:80-105` has no `model` key for research-agent; model resolves via the dispatcher default (`child-config.ts:166`). Removing the line changes nothing on this path.
- **git-investigator's model is still pinned (Claim C) — CONFIRMED.** `git-investigator.ts:15` (`model: 'sonnet' as const`) → `builtins.ts:113` (`model: gitInvestigator.model`). Effective model stays `sonnet`, independent of the removed `.md` line.
- **Blast radius — no runtime consumer.** The only readers of these two `.md` files are the two `.ts` wrappers (frontmatter stripped downstream), `vendored.test.ts`, and dev-only `scripts/update-hash-pins.ts`. No production path routes model/cost/tools from this frontmatter. "No runtime behavior change" holds.
- **Frontmatter still valid** after removing only the `model:` line — `name`/`description`/`tools` and `---` delimiters intact for both files.

## Findings
- **nit — split source-of-truth for git-investigator's model.** After this change, `git-investigator.md` no longer shows `model:`, while its effective model (`sonnet`) lives in `git-investigator.ts:15`. A reader of the `.md` alone can't see the model. The PR body already documents this (Caveat C), so it's acceptable — a one-line comment near `git-investigator.ts:15` ("model lives here, not in the vendored .md") would prevent future confusion. Not blocking.
- No critical / high / medium / low issues.

## Dimension summary
- **Correctness:** Sound — inert frontmatter removed, no logic touched. ✓
- **Spec-compliance:** Matches stated intent (remove inert frontmatter, re-pin hashes); no scope creep. ✓
- **Test coverage:** `vendored.test.ts` drift hashes re-pinned and verified to match. ✓
- **API-compat:** No public surface affected. ✓
- **Security / perf:** No trust boundary or hot path touched. ✓

## What was not checked
- Read against branch HEAD `1e3f3ff166aaba80bf475d5d0f46239c7b08ced6` — all citations verified at that ref; hashes recomputed with the test's exact procedure (raw bytes, single trailing LF).
- Stated intent: PR #625 title + body — spec-compliance assessed.
- Tests were evaluated by reading + recomputing hashes, not executed in a checkout (review is read-only); CI on the PR is green.

---
_Generated by the `agent-afk /review` tool (read-only); posted by an automated hourly sweep. A maintainer will follow up._

PR #628, opened 2026-07-16
fix(subagent): resolve zero-output stream-incomplete run as failed, not succeeded

griffinwork40:
## The bug (false-success contract)

When a forked subagent's model turn ends with **no terminal assistant message and an empty buffer** — e.g. the ~180s first-token/TTFB timeout aborts a connection stalled in the Anthropic SDK's internal 429/503/529 retry-backoff — and the run was **not** user-cancelled, `streamToFinalMessage` in `src/agent/subagent/handle.ts` fell through to a final synthetic-placeholder branch. That branch stamped `stopReason = stream_incomplete` and **RETURNED** a placeholder assistant message.

Because it **returned** (did not throw), the success path set `currentStatus = 'succeeded'` and `runToResult` → `buildResultFromMessage(id, 'succeeded', …, stopReason='stream_incomplete')`. Result: **a zero-output timeout was reported as `status:'succeeded'`, turnCount 0.** This "false success" fools any consumer that checks only `status === 'succeeded'` — it already bit the **forge** skill and **mint**'s phases. The only thing protecting core consumers was the opt-in `annotateIfIncomplete` helper.

## The fix (exact semantic change)

The **zero-output** synthetic-placeholder branch now **throws a distinct, informative `StreamIncompleteError`** (new class in `src/utils/errors.ts`) instead of returning a placeholder:

- `run()`'s catch runs with the signal **not** aborted and status still `running`, so it takes its **non-cascade `else` branch** → `currentStatus = 'failed'`.
- `runToResult` → `buildResultFromError` carries the error's actionable message **and preserves `stopReason: 'stream_incomplete'`** on the failed result.
- A consumer's natural `status !== 'succeeded'` check now catches the zero-output run, and `describeFailure(result)` reads sensibly — **not** an opaque failure. The class is deliberately **not** an `AbortError`, so the run is never misclassified `'cancelled'` (that stays reserved for user/cascade aborts).

**Deliberately unchanged:**
- **BUFFERED-PARTIAL branch** (real streamed text, `lastStreamedContent.length > 0`): still resolves **succeeded-partial** with `stopReason` `stream_incomplete` (kept via `??=`), so its partial work is salvaged + annotated at the consumption boundary. A regression-guard test pins this.
- **tool-use-cap branch** (`tool_use_loop_capped`) and the **cancel/abort throw** are untouched.

### Which approach (a vs b) and why

Chose **approach (b): throw a distinct non-cancel error.** Rationale:

- **(a)** (set `currentStatus='failed'` and return the placeholder) yields a failed result carrying a `message` + `stopReason` but **no `error`** — so `describeFailure(result)` would read just `"failed"` (uninformative), and the four failure paths that read `result.error?.message` would fall back to generic strings. Making it informative would require modifying the shared `buildResultFromMessage` to synthesize an error for the failed+incomplete case, which risks the existing schema-mismatch path (also `status:'failed'` with an error) — larger blast radius.
- **(b)** yields `status:'failed'` + preserved `stopReason` + an **informative `error.message`** via the existing `buildResultFromError`, with the change localized to `handle.ts` + one new error class. `describeFailure` reads `"failed: subagent … produced no output …"`. `partialOutput` is correctly absent (there is none). Smallest blast radius, least test churn, honors the original authors' "return, don't throw an *opaque* failure" intent by throwing an **informative** one.

## Blast radius — the four `annotateIfIncomplete` sites

They only annotate on `status === 'succeeded'`. A zero-output run is now `status:'failed'`, so each takes its **failure** path (audited, all sensible — no crash, no silent drop, no unintended parent abort):

| Site | Behavior now on a failed zero-output result |
|---|---|
| `agent/tools/skill-executor/fork-dispatch.ts:357` | `errorMessage = result.error?.message ?? noOutputError` resolves to the `StreamIncompleteError` message → `{ content, isError: true }` |
| `agent/tools/subagent/foreground-promotion.ts:246` | emits `subagent.failed` telemetry → `buildFailurePayload({status:'failed', errorMessage, …})` → `{ content: JSON.stringify(payload), isError: true }` |
| `agent/dag-subagent.ts:160` | `status !== 'succeeded'` → throws `result.error` (decorated via `attachSubagentContext`) → DAG node rejects → `failFast` aborts the DAG (fails fast, downstream skipped) |
| `cli/commands/interactive/bg-result-notifier.ts:91` | `job.status==='failed'` is NOT notice-only (only `cancelled` is) → **injected** into the parent's next turn as `Subagent failed — StreamIncompleteError: <msg>` |

The **buffered-partial** case still hits the annotate/success path (unchanged). Two further consumers **improve** automatically (they were previously fooled by the false success): `agent/subagent/wave.ts` `failFast` and `skills/mint/_phases/parallelize-dispatch.ts` now correctly fail-fast on a zero-output run.

## Tests

- Updated the two `handle-streaming.test.ts` tests that asserted the old `succeeded`+placeholder behavior → now assert `status:'failed'`, preserved `stopReason:'stream_incomplete'`, an informative `StreamIncompleteError`, and absent `partialOutput`.
- Added a **regression guard** asserting the BUFFERED-PARTIAL branch STILL yields `succeeded` + `stopReason:'stream_incomplete'` (partial work preserved, no error).
- Updated doc comments in `result.ts` (`STREAM_INCOMPLETE`, `isIncompleteStopReason`, `annotateIfIncomplete`) and the `Invariant:`/`History:` blocks in `handle.ts` to the new failed semantics.

## Files changed

- `src/utils/errors.ts` — new `StreamIncompleteError` class (+ doc).
- `src/agent/subagent/handle.ts` — zero-output branch throws instead of returning; comments updated.
- `src/agent/subagent/result.ts` — doc comments reflect the failed-vs-succeeded-partial split.
- `src/agent/subagent/handle-streaming.test.ts` — updated 2 tests + added 1 regression guard.

## Verification (all green)

| Gate | Result |
|---|---|
| `pnpm install` | OK (fresh worktree; benign postinstall MODULE_NOT_FOUND pre-build) |
| `pnpm build` (`tsc && copy-prompts`) | ✅ exit 0 |
| `pnpm lint` (`tsc --noEmit`, strict) | ✅ exit 0 |
| `pnpm test` (full suite) | ✅ **668 files, 12196 passed, 14 skipped, 0 failed** |
| `pnpm test:coverage` (CI gate, hard thresholds) | ✅ exit 0 (thresholds met; `src/utils/errors.ts` 100%) |
| `pnpm audit:sdk:check` | ✅ "Lock matches current imports" |
| `pnpm scan:env:check` | ✅ "in sync" |

## Reviewer notes / caveats

- No public API signature changed; `SubagentResult` shape is unchanged (the `failed` result simply now populates `error` + `stopReason` for this case). No SDK import added; no env change.
- Behavior change reviewers should know: any caller that *relied on* the old false-success (treating a zero-output timeout as a usable succeeded result) will now see a `failed` result. That is the intended correction — the four core sites and the two `failFast` consumers all handle it correctly, surfacing the actionable error.

### DIFF
--- a/src/agent/subagent/handle-streaming.test.ts
+++ b/src/agent/subagent/handle-streaming.test.ts
@@ -207,13 +207,56 @@ describe('SubagentHandle streaming', () => {
       expect(handle.status).toBe('succeeded');
     });
 
-    it('returns a stream-incomplete partial (not a throw) when the stream ends with neither message nor streamed content', async () => {
-      // Degradation contract: an empty cut-off stream (no terminal message, no
-      // buffered text, no error, and NOT the tool-use cap) must NOT throw an
-      // opaque "produced no terminal message". It returns a STREAM_INCOMPLETE
-      // partial (status 'succeeded') carrying a cut-off marker, so the parent
-      // gets an actionable incomplete result — annotateIfIncomplete flags it at
-      // the consumption boundary — instead of a bare delegation failure.
+    it('BUFFERED-PARTIAL branch STILL resolves SUCCEEDED (partial) with stopReason preserved — regression guard', async () => {
+      // Regression guard for the false-success fix: the fix throws ONLY on the
+      // ZERO-OUTPUT branch. A stream that ended with real buffered text but no
+      // terminal `message` event (a cut-off mid-output) MUST keep its prior
+      // contract — status:'succeeded' with stopReason:STREAM_INCOMPLETE — so its
+      // partial work is salvaged and flagged (via annotateIfIncomplete) at the
+      // consumption boundary, NOT reclassified as a failure. Only the
+      // empty-buffer case became a failure.
+      const events: OutputEvent[] = [
+        { type: 'chunk', chunk: { type: 'content', content: 'salvageable ' } },
+        { type: 'chunk', chunk: { type: 'content', content: 'partial findings' } },
+        { type: 'done' },
+      ];
+      const session = createDeterministicMockSession(events, {
+        role: 'assistant',
+        content: 'unused',
+        timestamp: new Date(),
+      });
+      const handle = new SubagentHandleImpl(
+        'subagent-buffered-partial-test',
+        session,
+        controller,
+        abortGraph,
+        undefined,
+        5000,
+        undefined,
+        vi.fn(),
+      );
+
+      const result = await handle.runToResult('p');
+      expect(result.status).toBe('succeeded');
+      expect(result.stopReason).toBe(STREAM_INCOMPLETE);
+      // The buffered text is preserved as the (partial) message content.
+      expect(result.message?.content).toBe('salvageable partial findings');
+      // Not a failure — no error attached on the succeeded-partial path.
+      expect(result.error).toBeUndefined();
+    });
+
+    it('resolves a ZERO-OUTPUT cut-off run as FAILED with stopReason preserved (not a false success)', async () => {
+      // False-success contract fix: an empty cut-off stream (no terminal
+      // message, no buffered text, no error, and NOT the tool-use cap — e.g. a
+      // first-token/TTFB timeout guillotines a connection stalled in the
+      // provider SDK's retry-backoff) must NOT resolve as status:'succeeded'.
+      // A zero-output timeout previously fell through to a synthetic-placeholder
+      // RETURN, which set status:'succeeded' — fooling every consumer whose only
+      // gate is `status === 'succeeded'`. It now THROWS a distinct
+      // StreamIncompleteError, so runToResult builds a status:'failed' result
+      // that carries the informative error message AND preserves
+      // stopReason:'stream_incomplete'. A consumer's natural `status !==
+      // 'succeeded'` check now catches it.
       const events: OutputEvent[] = [{ type: 'done' }];
       const session = createDeterministicMockSession(events, {
         role: 'assistant',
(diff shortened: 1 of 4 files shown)


griffinwork40:
I have everything I need. All load-bearing facts are independently verified against the reviewed ref (`d6eac16`). No critical/high findings, so `/shadow-verify` is not triggered — and I re-derived every load-bearing claim myself against the reviewed ref, which is stronger. Synthesizing the review.

---

# Code Review — PR #628

**Target:** PR #628 `fix(subagent): resolve zero-output stream-incomplete run as failed, not succeeded` · `main` ← `afk/fix-incomplete-succeeded-status`
**Reviewed ref:** `d6eac16ebcc9e2e9ce4059bb294e23662e183cd9` (all citations read at this SHA via `git show`)
**Change type:** hotfix (contract/semantic fix) · **Regime:** full (elevated from line-count `light` — semantic status-contract change with wide consumer surface) · **4 files, +184/−64**
**Stated intent:** PR #628 title + body — spec-compliance assessed.

**Summary:** A forked subagent whose model stream ends with no terminal `message` event **and** an empty buffer (not cancelled, not capped) previously *returned* a synthetic placeholder → `status:'succeeded'`, a false success that fooled `status === 'succeeded'` consumers (bit forge/mint). The fix *throws* a new `StreamIncompleteError` on that zero-output branch, so `run()`'s catch classifies it `failed`, `buildResultFromError` preserves `stopReason:'stream_incomplete'` and carries an informative error. The buffered-partial branch (real streamed text → succeeded-partial) is deliberately unchanged.

## Verdict: ✅ MERGE

The core fix is correct, minimal, and well-tested. I independently re-derived every load-bearing claim against the reviewed ref (not the PR narrative): `run()`'s catch (`handle.ts:305-320`) takes the non-cascade `else` → `currentStatus='failed'` because the throw is guarded by `if (signal.aborted || currentStatus==='cancelled')` above it (`handle.ts`, cancel/abort path preserved); `buildResultFromError` (`result.ts:238-247`) returns `{status, error, stopReason}`; branch ordering puts buffered-partial (`??=`, return) before zero-output (`=`, throw); the three `handle-streaming.test.ts` assertions match; and a full-tree consumer audit found **zero broken sites** — every `.message`/`.message?.content` deref is status-guarded, every failure reader uses `result.error`, and `annotateIfIncomplete`/`isIncompleteStopReason` are never reached on a `failed` result. The PR's "full suite green" is consistent with the reviewed-ref test state (a subagent "broken test" flag was a working-tree artifact from the divergent checkout, disproven against the ref).

Two **non-blocking** findings follow.

## Findings

### 1. `low` · confidence medium · correctness/observability · `src/agent/subagent/handle.ts:502-508` · ref d6eac16 · citation-type: file-state
**Finding:** The zero-output branch's runtime error message and comment attribute the branch to a *first-token/TTFB timeout during SDK retry-backoff*, but a terminal TTFB timeout does not reach this branch — it surfaces as an `error` event and is thrown one path earlier.
**Evidence** (control flow at the reviewed ref):
```
loop.ts:426    yield { type: 'error', error: e }; return;   // terminal TTFB (input.signal NOT aborted)
handle.ts:418  if (streamError) throw streamError;          // error event → thrown HERE, before the zero-output fallback
handle.ts:505  `This is typically a first-token timeout while the provider was overloaded ` +
```
A TTFB timeout aborts the *chained* `ttfb.signal` (not the subagent's `input.signal`), retries once (`loop.ts:409-411`), then on exhaustion yields an `error` event (`loop.ts:426`) → `streamToFinalMessage` sets `streamError` and throws at `handle.ts:418` → already `failed` pre-PR. The branch's **own** second attribution (`handle.ts:491`, "an empty-text turn the stream consumer drops before emitting a `message` event") is the mechanism consistent with the control flow. The two attributions contradict; the TTFB one is unreachable here.
**Suggestion:** Reword the `StreamIncompleteError` message and the `handle.ts:458` comment to attribute the branch to a content-free/empty-text stream end (a turn dropped by `if (event.text)`), not a "first-token timeout" — the misattribution steers a parent's retry/fall-back decision (this string is injected into the parent context via `bg-result-notifier`) toward "transient overload" when the trigger is often deterministic. (Not checked: openai-compatible provider stall behavior — a different provider ending a stream content-free without an error event would also land here; the finding is scoped to the anthropic-direct TTFB path I verified.)

### 2. `nit` · confidence high · documentation · PR description
**Finding:** The PR body's blast-radius table claims the differing-behavior consumer set is *exactly six* sites; the full-tree audit found additional production consumers that observably branch on the new `failed` status — `src/agent/subagent/background-registry.ts:609`, `src/cli/slash/commands/bgsub.ts:122`, and the `src/skills/mint/_phases/*.ts` `status !== 'succeeded'` short-circuits.
**Evidence:** all read at their branch points; each takes its failure path correctly (reads `result.error`, no crash, no silent drop).
**Suggestion:** Note in the PR body that the six enumerated sites are illustrative, not exhaustive. No code impact — every unlisted consumer handles the `failed` result correctly, so the undercount is description-accuracy only, and the PR's stronger claim ("all handle it correctly") holds.

## What was not checked
- **Ref read:** Wave-1 consumer audit read the working tree (`polish-goblin-banner`); those consumer files are **byte-identical** to the reviewed ref (`git diff d6eac16 HEAD` shows divergence only in the 3 core PR files), so reads are authoritative. All diff-file citations were read at `d6eac16` via `git show` — verified at that ref.
- **Spec-compliance:** assessed against PR #628 title+body — fully implemented, no unmet intent, no scope creep; the approach-(b)-over-(a) rationale is sound.
- **Security:** internal error class only — no auth, secrets, injection, or external I/O; the error string carries only the subagent id.
- **Not run:** the test suite (read-only review — I confirmed the reviewed-ref test *assertions* are consistent with the fix, but did not execute `pnpm test`). Provider coverage limited to anthropic-direct (openai-compatible stall path not traced). Plugin/vendored skill bodies outside `src/` not audited.

---

PR #630, opened 2026-07-16
fix(telegram): reserve claimedChats in processOne to close drain-path TOCTOU (#603)

### LINKED ISSUE #603: Telegram detached-turn follow-ups: drain-path claimedChats race + log redaction + test + doc (from #602 review)
Follow-ups from the code review of #602 (*run agent turn detached so mid-turn elicitations don't deadlock the poller*). The merged fix is correct; these are non-blocking items the review surfaced. Reviewed ref: `05c5506`.

## 1. `medium` · correctness — drain-path turns bypass the `claimedChats` reservation (residual TOCTOU)

`drainQueue`-dispatched turns reopen the idle-window that #602's `claimedChats` guard closes for the *first* turn. `processOne`'s `finally` fires `this.drainQueue(chatId)` **un-awaited** (`src/telegram/handlers/message.ts:698-699`), so `handle()`'s `finally` deletes the claim (`message.ts:486`) *before* the drained turn flips `session.state` — the flip is deferred to the lazy `async* sendMessageStream`'s first `iter.next()`, which runs after the "Thinking…" placeholder (`streaming.ts:274/321`, `agent-session.ts:589-591`). A fresh same-chat message arriving in that gap sees `state === 'idle'` **and** `claimedChats` empty and also enters `processOne`.

**Impact is bounded / self-healing** (independently verified): the loser's `sendMessageStream` hits `assertCanSend()` → throws `'session is busy'` (`agent-session.ts:1242-1243`), caught and **re-enqueued** (`message.ts:669-677`). Worst case is transient message re-ordering or a redundant "Thinking…"→"Queued" flip — never two concurrent streams, never data loss. That is why it did not block the merge.

**Fix:** reserve `claimedChats` synchronously at the top of `processOne` (release in its own `finally`, after `drainQueue`), so drained turns get the same synchronous slot the first turn gets. This also makes PR #602 claim (a) ("one-turn-at-a-time … exactly as for drained turns") literally true rather than true-by-self-healing.

## 2. `low` · security — detached-error log skips token redaction

`runDetached` forwards the raw `err` to `this.log` (`src/telegram/bot.ts:252-253`) without the `/bot<TOKEN>/` redaction that `handlePhoto`'s own catch applies (`message.ts:357`). A rejection whose message embeds a `getFileLink` URL reaching this last-resort catch would print the bot token to logs. Reachability is low (`handle()`/`handlePhoto()` own their user-facing errors; only a throw *after* their try/finally surfaces here).

**Fix:** apply the same `.replace(/\/bot[^/]+\//g, '/bot[REDACTED]/')` to `err` before logging.

## 3. `low` · test-coverage — no test for the residual drain-idle-window race

The two new TOCTOU tests race two `handle()`/`handlePhoto()` calls; neither exercises a fresh `handle()` racing a detached `drainQueue`-dispatched turn's idle-window (the `drain`+`concurrent` grep hits are all in `message-compaction.test.ts` / `session-manager.test.ts`, a different race).

**Fix:** add a test that resolves the winner's turn, then dispatches a fresh `handle()` in the same microtask as the detached drain, asserting the second does not reach `streamResponse` concurrently. (Pairs with item 1 — write the test red first, then land the `processOne` reservation.)

## 4. `nit` · docs — `claimedChats` field-doc mis-states the state-flip mechanism

The field doc (`src/telegram/handlers/message.ts:122-125`) attributes the deferred `session.state` flip to "streaming.ts sends the Thinking… placeholder before … `assertCanSend()`". The actual mechanism is that `sendMessageStream` is a lazy `async*` generator — its body (`assertCanSend` + `currentState = 'streaming'`) runs only on the first `iter.next()`. The doc's *conclusion* (the window exists) is correct; only the *mechanism* is mis-stated. Cheap to fix when item 1 is touched.

---

_Filed from the `/review` of #602. Items 1 + 3 are naturally done together._

griffinwork40:
Closes #603.

Follow-ups from the `/review` of #602 (*run agent turn detached so mid-turn elicitations don't deadlock the poller*). All four review items in one scoped change.

### 1 · `medium` correctness — close the drain-path `claimedChats` TOCTOU
`handle()`/`handlePhoto()` reserved `claimedChats` only for the **first** turn and released it in their `finally`. But `processOne`'s `finally` fires `drainQueue` **un-awaited**, so the outer `finally` deleted the claim *before* the drained turn flipped `session.state` (that flip is deferred to the lazy `async* sendMessageStream`'s first `iter.next()`). A fresh same-chat message arriving in that gap saw `state === 'idle'` **and** an empty claim → it could enter `processOne` concurrently with the drained turn.

**Fix:** `processOne` now reserves a slot **synchronously at entry** and releases it **after** firing `drainQueue`, so the drained turn's own reservation is live before the outer turn's release drops the count — the slot is never momentarily empty while a detached drain turn is in flight.

`claimedChats` is changed from `Set<number>` to a **reference-counted `Map<number, number>`** (with `reserveClaim`/`releaseClaim`/`isClaimed` helpers). This is required, not incidental: with a plain `Set`, the outer guard and `processOne` both hold the *same* entry, so `processOne`'s release would clobber the entry the drained turn still needs during the hand-off. Ref-counting lets the overlapping reservations (outer guard + `processOne` + drain re-entry) compose so the count stays `> 0` continuously across the gap. The existing concurrent-`handle()` guarantee and the busy-recovery re-enqueue contract are preserved; each `reserveClaim` is balanced 1:1 with a `releaseClaim` on every path.

### 2 · `low` security — redact bot token in `runDetached`'s last-resort log
`bot.ts`'s `runDetached` forwarded the raw `err` to `this.log` without the `/bot<TOKEN>/` redaction `handlePhoto`'s own catch applies — a `getFileLink` URL reaching this catch could leak the live token. Now stringifies safely and applies the same `.replace(/\/bot[^/]+\//g, '/bot[REDACTED]/')` before logging.

### 3 · `low` test-coverage — regression test for the residual drain-idle-window race
New test in `message.test.ts`: gates the drained turn's `streamResponse` in-flight, then fires a fresh un-awaited `handle()` in the idle-window and asserts it does **not** enter `streamResponse` concurrently (`maxConcurrentStreams === 1`, exactly 2 stream calls, fresh message gets a `#1` queued reply). Red without Item 1 (`maxConcurrentStreams === 2`, 3 stream calls), green with it. Adds an `afterEach` mock-reset so the blocking gate can't leak into later suites.

### 4 · `nit` docs — correct the `claimedChats` field-doc mechanism
The field doc attributed the deferred `session.state` flip to the "Thinking…" placeholder ordering. Corrected: the flip is deferred because `sendMessageStream` is a **lazy `async*` generator** whose body (`assertCanSend()` + `state = 'streaming'`) runs only on the consumer's first `iter.next()`; the placeholder widens, but does not cause, the window.

## Verification
- `pnpm lint` (`tsc --noEmit`) — clean.
- `pnpm test src/telegram/handlers/message.test.ts src/telegram/handlers/message-photo.test.ts` — **31 passed**.
- Full `src/telegram` suite — **609 passed** (32 files).
- Red/green of the new test confirmed by temporarily reverting the `processOne` reservation.

Scoped to #603 — no drive-by changes. Branch rebased on current `main`.

### DIFF
--- a/src/telegram/bot.ts
+++ b/src/telegram/bot.ts
@@ -257,7 +257,16 @@ export class TelegramBot {
     // `runDetached`'s .catch is only a last-resort guard against an unhandled
     // rejection (e.g. a reply that throws after the bot is stopped).
     const runDetached = (work: Promise<void>): void => {
-      void work.catch((err) => this.log('Detached update handler error:', err));
+      void work.catch((err) => {
+        // Redact any embedded bot token before logging — getFileLink() URLs
+        // (https://api.telegram.org/file/bot<TOKEN>/<path>) can ride along in
+        // an error string, and this last-resort catch would otherwise leak the
+        // live token to the log. Same pattern handlePhoto's own catch applies
+        // (#603 Item 2). Stringify safely first: `err` may be a non-Error.
+        const rawErrStr = err instanceof Error ? err.message : String(err);
+        const sanitizedErr = rawErrStr.replace(/\/bot[^/]+\//g, '/bot[REDACTED]/');
+        this.log('Detached update handler error:', sanitizedErr);
+      });
     };
     this.bot.on('text', (ctx) => runDetached(this.messageHandler.handle(ctx)));
 
--- a/src/telegram/handlers/message.test.ts
+++ b/src/telegram/handlers/message.test.ts
@@ -6,7 +6,7 @@
  * the session message queue or processOne.
  */
 
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import type { Context } from 'telegraf';
 import type { Message } from 'telegraf/types';
 
@@ -231,6 +231,15 @@ describe('MessageHandler concurrent same-chat dispatch (TOCTOU regression — PR
     mockStreamResponse.mockClear();
   });
 
+  // Restore the shared mock's implementation unconditionally — the drain-race
+  // test below installs a blocking mockImplementation, and if any of its
+  // assertions throw first, an inline restore would be skipped and the blocking
+  // gate would leak into later suites (hang). afterEach always runs.
+  afterEach(() => {
+    mockStreamResponse.mockReset();
+    mockStreamResponse.mockImplementation(async () => { /* no-op */ });
+  });
+
   it('a second concurrent update for the same chat is queued, not processed concurrently', async () => {
     // getSession always resolves { state: 'idle' } — never reflects a real
     // busy transition — so this reproduces exactly the unprotected window:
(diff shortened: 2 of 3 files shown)


griffinwork40:
Wave 1.5 verification complete. The load-bearing finding checks out, and the decisive regression question is now answered:

**compact/clear finding — verified:**
- `processClearDirect` (L544-553) and `processCompactDirect` (L568-621) confirmed to contain **no** `reserveClaim`/`isClaimed` — the absence claim is `confirmed-absent`.
- **Decisive: this gap is PRE-EXISTING, not introduced by #630.** At parent `6d5fa9f0^`, `claimedChats` was a `Set` and `add/has/delete` appear **only** in `handle()`/`handlePhoto()` — `processOne`, `processCompactDirect`, and `processClearDirect` reserved **nothing**. So this PR *adds* a reservation to `processOne` (closing the message/photo hole) and leaves compact/clear exactly as they were. It strictly improves; it introduces no regression.
- Mitigation confirmed accurate: `processCompactDirect` has its own `session.compact() → reason:'session-busy'` re-enqueue gate (L590-595); `processClearDirect` has none but is a fast `resetSession` + reply.

No critical/high findings exist, so `/shadow-verify` is not mandated (I verified the one load-bearing medium finding directly). Emitting synthesis.

---

# Code Review — PR #630

**Target:** PR #630 `fix(telegram): reserve claimedChats in processOne to close drain-path TOCTOU (#603)` · branch `afk/issue-603-telegram-drain-toctou` → `main`
**Reviewed ref:** `6d5fa9f0a292da29d29813251fc8921e133f496e` (all citations resolved via `git show <ref>:<file>`; working tree is on `main`, not the PR branch)
**Regime:** upgraded `light`→`full` (2 parallel agents) given concurrency-invariant stakes · 3 files, +218/-28

**Verdict:** The fix **correctly closes the message/photo drain-path TOCTOU**. The ref-count invariant is **balanced 1:1 on every code path** (normal, throw, busy re-enqueue), and the hand-off continuity that the fix depends on **holds for chained drains**. The Set→Map change is required, not incidental. Token-redaction (Item 2) is a net security improvement. The regression test is genuinely red-without-fix. No critical/high/blocking defects.

### Findings (sorted)

**1 · medium · confidence high · correctness · `src/telegram/handlers/message.ts:786-798` · ref:6d5fa9f0 · file-state**
`drainQueue` dispatches `compact`→`processCompactDirect` and `clear`→`processClearDirect`, neither of which reserves a `claimedChats` slot, so a fresh `handle()` landing while a drained compact/clear turn is in flight re-enters the same idle-window this PR closes for message/photo.
- Evidence: `else if (item.type === 'compact') { await this.processCompactDirect(chatId, item.ctx); }` (L793-794); no `reserveClaim` in L544-553 / L568-621.
- **Not a blocker / not a regression:** verified at parent `6d5fa9f0^` that compact/clear reserved nothing *before* this PR either — this is a **pre-existing** gap, out of #603's literal scope ("message/photo drained turn"), untouched and un-worsened here. The compact path is further mitigated by its own `session-busy` re-enqueue gate (L590-595).
- Suggestion (follow-up issue, not this PR): symmetrically wrap the compact/clear branches with `reserveClaim`/`releaseClaim`, or reserve inside `processCompactDirect`/`processClearDirect`.

**2 · low · confidence high · security/observability · `src/telegram/bot.ts:257-267` · ref:6d5fa9f0 · diff-context**
`runDetached` now logs only `err.message` (Error) / `String(err)` (non-Error) instead of the whole `err` object; a token surviving in a nested field (`err.cause`, `err.response...`) is not scanned by the `/bot…/` regex — but it is also no longer logged, so the change is **net-safe** and strictly shrinks the leak surface. Tradeoff: the error **stack is dropped** from the last-resort log (minor debuggability loss).
- Evidence: `const rawErrStr = err instanceof Error ? err.message : String(err); const sanitizedErr = rawErrStr.replace(/\/bot[^/]+\//g, '/bot[REDACTED]/');`
- Suggestion: keep as-is; accept the stack-drop as the cost of not leaking tokens (matches `handlePhoto`'s existing pattern, message.ts L400).

**3 · nit · security · `src/telegram/bot.ts:266` · ref:6d5fa9f0 · diff-context** — a token in a non-URL shape (e.g. bare `"token=123456:ABC"`) has no `/bot…/` delimiter and is left intact; out of Item 2's stated `getFileLink`-URL threat (which always carries the `/bot<TOKEN>/` path form). No change needed.

**4 · nit · api-compat · `src/telegram/handlers/message.ts:144,153-171` · ref:6d5fa9f0 · file-state** — `claimedChats` (`Set`→`Map<number,number>`) and new `reserveClaim`/`releaseClaim`/`isClaimed` are all `private`; zero production importers `[UNVERIFIED: no production importers]`. Public `drainQueue` signature unchanged (called from bot.ts:131). No consumer impact.

### Spec-compliance (intent = PR #630 title + body, 4 items)

| Item | Implemented | Scope |
|---|---|---|
| 1 — drain-path TOCTOU; Set→ref-counted Map + reserve/release/isClaimed | **Yes** — Map L144; helpers L153-171; `processOne` reserve L707 / drain-then-release L774-777; outer guards → `isClaimed`/`reserveClaim` L267-268, L500-501 | In scope. Compact/clear residual (finding 1) is pre-existing, flagged as risk not creep |
| 2 — redact bot token in `runDetached` | **Yes** — bot.ts L260-267 | In scope; mirrors `handlePhoto` |
| 3 — regression test | **Yes** — message.test.ts: `maxConcurrentStreams).toBe(1)`, `toHaveBeenCalledTimes(2)`, `#1` reply; red without the `processOne` reservation | In scope; `afterEach` reset is necessary infra |
| 4 — field-doc mechanism correction (lazy `async*` generator) | **Yes** — L120-142 | In scope |

No scope creep detected.

### Verified independently (beyond agent reports)
- Ref-count reserve placement: only `let reEnqueued = false` sits between `processOne`'s `reserveClaim` (L707) and `try` (L714) — cannot throw → no leak path.
- Hand-off crux: `drainQueue` has **zero** async suspension before `await processOne` (L787-791), and `reserveClaim` is `processOne`'s first statement → count stays `>0` across the un-awaited fire, including chained drains.
- Compact/clear reservation absence + its pre-existing nature (parent-commit diff).

### What was not checked
- **Ref read:** Wave 1 agents + orchestrator all read against `6d5fa9f0` — citations verified at that ref.
- **Unverified foundational premise:** the lazy-`async*`-generator deferred `session.state` flip that makes the original race *possible* was relied upon from the field-doc / PR #602 lineage, not re-derived from `streaming.ts`/`agent-session.ts`. Non-decisive — the fix is correct and low-cost regardless of the exact window width.
- **Test suite not run** here (PR reports `pnpm lint` clean + 609 telegram tests passing + red/green confirmed) — static review only.
- Runtime microtask ordering is derived from source, not from an actual execution trace.

## Done
- **Merge recommendation: MERGE.** The change does exactly what its intent states, the concurrency invariant is verified balanced and the TOCTOU hand-off holds (incl. chained drains), the token redaction is net-safe, and the regression test is valid. No critical/high/blocking findings.
- One **medium follow-up** (finding 1): compact/clear drain branches lack the same reservation — **pre-existing, out of #603 scope, not introduced or worsened by this PR**; worth a separate tracking issue, not a blocker here.
- Evidence: review synthesized from 2 parallel `research-agent` reviewers + orchestrator Wave 1.5 verification, all cited against reviewed ref `6d5fa9f0`; diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/262e83b7-af42-42db-ba79-a778b4ad1904/pr-630.diff`.
- Read-only review: no files, PR state, or comments were mutated.

---

PR #627, opened 2026-07-16
fix(mint): hard-fail phases on incomplete subagent result (capped / stream_incomplete)

griffinwork40:
## The bug

agent-afk's subagent runtime returns a `SubagentResult` with `status: 'succeeded'` **even when the run was incomplete** — either the tool-use iteration cap fired (`stopReason === 'tool_use_loop_capped'`) or the stream closed without a terminal message (`stopReason === 'stream_incomplete'`). The helper `isIncompleteStopReason(stopReason)` in `src/agent/subagent/result.ts` returns `true` for both.

`mint`'s phase dispatchers read a phase subagent's `.message.content` gated **only** on `status === 'succeeded'`, so an incomplete/placeholder partial was silently treated as the real phase output — and in mint that output is then fed programmatically into the **next** phase (spec → research → plan → build → verify → heal → ship).

This is the **same class** as the recent forge `STREAM_INCOMPLETE` guard, but it lands in mint's phase pipeline, which — unlike forge — has **no qualify gate downstream** to catch a degraded phase. A truncated spec/research/plan/heal/ship silently propagates instead of failing loudly.

## The fix

Extend each phase's existing status guard with an `isIncompleteStopReason(result.stopReason)` check that fires **before** `.message.content` is read.

- The five throw-style phases hard-fail with a clear, uniform message: `<phase> phase returned an incomplete result (stopReason=<...>)`.
- `parallelize-dispatch` returns its `{ kind: 'failed', error }` discriminated-union value (matching that file's existing idiom) instead of throwing.
- `isIncompleteStopReason` is imported via the existing relative path `../../../agent/subagent/result.js`, matching how the sibling phases already import `describeFailure` from that module.

`annotateIfIncomplete` is **deliberately not used** here. That helper is for content shown to a *model* as a tool result (see `fork-dispatch.ts`, `dag-subagent.ts`); mint consumes the content **programmatically** as the phase output, so an incomplete phase must **hard-fail**, not forward an annotated placeholder.

### The 6 fixed sites

| File | Guard style |
|------|-------------|
| `src/skills/mint/_phases/spec.ts` | throw before `return specResult.message.content` |
| `src/skills/mint/_phases/research.ts` | throw before `return researchResult.message.content` |
| `src/skills/mint/_phases/plan.ts` | throw before `return planResult.message.content` |
| `src/skills/mint/_phases/ship.ts` | throw before `emitCard` / `return shipResult.message.content` |
[...]

PR #635, opened 2026-07-16
fix(provider): surface stream-incomplete as an error, not a silent truncated turn

griffinwork40:
## Summary

Fixes a parent-facing **silent truncation** gap in both model providers: a model stream cut off mid-response (no terminal completion signal) was delivered to the agent as a clean, complete turn — so a truncated answer could be silently presented as final. Surfaced from an audit of agent-facing error handling. The subagent layer already guarded this class (`STREAM_INCOMPLETE`, #627/#628); the top-level provider paths did not.

## The bug

- **anthropic-direct** (`translate.ts`): afk consumes the raw `Stream<RawMessageStreamEvent>` (not the validating `MessageStream` helper), which completes cleanly — no throw — when an intermediary gracefully closes the connection before `message_stop`. `translateMessageStream` then fell through to a normal `turn-result` (`stopReason: null`) carrying the partial text → silent truncation.
- **openai-compatible** (`query/stream-drive.ts`): `driveStream` returned a clean completion even when the stream ended having produced no content and no terminal `finish_reason` → an empty turn delivered as success.

## The fix

- **anthropic-direct**: when the stream ends with `!stopped && stopReason === null` (no `message_stop` **and** no `message_delta` stop_reason), yield a `StreamIncompleteError` event instead of a `turn-result`. Reuses the existing, tested error path — both consumers already handle it (top-level surfaces the error; subagent → `status:'failed'`). Guarding on `stopReason === null` (not merely `!stopped`) avoids false-positives when only the framing `message_stop` is lost after a real stop_reason.
- **openai-compatible**: when the stream ends with no `finish_reason` **and** zero output (no text/reasoning/tool-calls), yield a `StreamIncompleteError`. **Deliberately scoped to zero-output** — some OpenAI-compatible endpoints (local MLX / llama.cpp shims) legitimately omit `finish_reason` on complete turns, so flagging "content present + no finish_reason" would false-positive real completions. `finish_reason` is not a reliable terminal signal on this wire (unlike the protocol-guaranteed `message_stop`), so the partial-truncation case can't be safely distinguished there.

## Tests

- `translate.test.ts`: positive (cut-off stream → error) + negative (lost only `message_stop` but a real stop_reason → still completes).
- `query/stream-drive.test.ts` (new): positive (empty stream → error) + negative (text but no `finish_reason` → still completes, shim-safe).
- Full suite: **12,265 pass / 14 skipped**; `tsc --noEmit` clean.

## Known limitation

The openai-compatible partial-truncation case (content present, no `finish_reason`, cut off) can't be safely distinguished from a legitimate `finish_reason`-omitting shim completion, so it is intentionally left unflagged (documented inline).

### DIFF
--- a/src/agent/providers/anthropic-direct/translate.test.ts
+++ b/src/agent/providers/anthropic-direct/translate.test.ts
@@ -495,4 +495,68 @@ describe('anthropic-direct translateMessageStream', () => {
     expect(u.output_tokens).toBe(25);
     expect(u.cache_read_input_tokens).toBe(50);
   });
+
+  it('stream ending with no message_stop and no stop_reason surfaces an error (P0: no silent truncation)', async () => {
+    // Regression guard for the parent-facing "silent truncation" gap: when a
+    // model stream is cut off before completion (e.g. an intermediary gracefully
+    // closing the connection mid-response), neither `message_stop` nor a
+    // `message_delta` stop_reason arrives. Before the fix, translate yielded a
+    // `turn-result` with stopReason=null carrying the partial text, so the turn
+    // was delivered as a clean, complete answer. It must surface an error so the
+    // partial is never silently presented as final.
+    async function* cutOff(): AsyncIterable<RawMessageStreamEvent> {
+      yield messageStart();
+      yield textBlockStart(0);
+      yield textDelta(0, 'Partial answer that got ');
+      // stream ends here — NO message_delta, NO message_stop
+    }
+
+    const out = await collect(
+      translateMessageStream(cutOff(), { sessionId: SESSION_ID }),
+    );
+
+    // The partial text still streams as a delta event...
+    const textDeltas = out.filter(
+      (o) => o.kind === 'event' && o.event.type === 'delta.text',
+    );
+    expect(textDeltas).toHaveLength(1);
+
+    // ...but the turn must NOT resolve as a (silent) turn-result.
+    const turnResults = out.filter((o) => o.kind === 'turn-result');
+    expect(turnResults).toHaveLength(0);
+
+    // It surfaces as an error event instead.
+    const last = out[out.length - 1];
+    if (!last || last.kind !== 'event' || last.event.type !== 'error') {
+      throw new Error('expected an error event when the stream ends without a terminal signal');
+    }
+    expect(last.event.error).toBeInstanceOf(Error);
+    expect(last.event.error.message).toMatch(/message_stop|incomplete|cut off/i);
+  });
+
+  it('stream that loses only message_stop but has a real stop_reason still yields a turn-result', async () => {
+    // Negative guard: the incomplete-detection must NOT false-positive when the
+    // model DID signal completion (message_delta carried a stop_reason) and only
+    // the framing `message_stop` event was lost. That turn is substantively
+    // complete and must resolve normally.
+    const events: RawMessageStreamEvent[] = [
+      messageStart(),
+      textBlockStart(0),
+      textDelta(0, 'Complete answer'),
+      blockStop(0),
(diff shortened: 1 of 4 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

Clean, correct, tightly-scoped hotfix. Both provider stream-incomplete guards were verified by reading the branch HEAD (`355e0360`) — placed on the genuine clean-EOF path, reusing pre-existing error-event shapes their consumers already handle, and guarded tightly enough not to false-positive legitimate completions. No blocking, critical, or high findings.

**Reviewed ref:** `355e03603cf283ada9c9d817c97e0ad98336c0f0` (vs. merge-base `c0d24d6`). 4 files, +228. Change type: hotfix.

### Findings by dimension

- **security** — no issues. Both `StreamIncompleteError` message strings are static; no user/secret data flows into them.
- **api-compat** — no issues. Both new yields reuse the existing `ProviderEvent {type:'error'}` union member (`src/agent/provider.ts:242`) and the anthropic-direct translate `{kind:'event',event:{type:'error'}}` shape already emitted at `translate.ts:302`. No signature or union change → no importer break.
- **correctness** — no issues.
  - anthropic-direct: the new guard `if (!stopped && stopReason === null)` at `translate.ts:319` sits *after* the catch-path early-return (`:299-303`) and *before* the turn-result (`:334-338`). `stopReason` is set from `message_delta` at `:249` and `stopped` from `message_stop` at `:287`, so a normal completion bypasses the guard and only a genuinely-cut stream reaches it. Guarding on `stopReason === null` (not merely `!stopped`) correctly still completes a stream that delivered a real stop_reason but lost only the framing `message_stop`.
  - openai-compatible: the zero-output guard at `stream-drive.ts:197-212` sits after the `streamError` early-return (`:175-177`) and before the clean-completion return (`:214-220`), and returns `null` like the other terminal returns. It fires only when all four `StreamState` fields are empty (`finishReason === null && assistantText.length === 0 && reasoningText.length === 0 && toolCallsByIndex.size === 0`), so a content-present-but-no-`finish_reason` shim completion does **not** trip it.
  - End-to-end routing confirmed: an anthropic-direct `type:'error'` event is caught in the subagent stream loop (`handle.ts:394-396`), re-thrown at `handle.ts:418`, and resolved by `runToResult` → `buildResultFromError` (`handle.ts:523`, `result.ts:238`) to `status:'failed'` with `stopReason:'stream_incomplete'` — matching the #627/#628 "fail loudly" precedent the PR cites.
- **spec-compliance** — no issues. Both guards match the stated intent exactly. The openai-compatible zero-output scoping and the intentionally-unflagged partial-truncation case are documented inline (`stream-drive.ts:189-196`) as a known limitation. No unrequested scope creep.
- **test-coverage** — no issues. Positive + negative tests exist for both providers and exercise the guards (verified by reading the tests, not by trusting the PR body): `translate.test.ts` cut-off→error (`:499`) and lost-`message_stop`-but-real-stop_reason→completes (`:537`); `stream-drive.test.ts` empty→error (`:52`) and text-without-`finish_reason`→completes, shim-safe (`:76`).
- **perf-observability** — no issues. Each guard is a single O(1) branch on already-accumulated state at stream end; no added iteration or logging cost on the hot path.

### What was not checked
- All file-state citations read at ref `355e03603c` via `git show`/`git grep` (working tree was not at this ref). Stated intent was available (PR title + body); spec-compliance assessed against it.
- The full test suite was not executed by the reviewer — the PR's "12,265 pass / 14 skipped, tsc clean" claim is unverified here (CI on the head SHA reports all checks SUCCESS).
- Scope confined to the four changed files; the broader stream lifecycle outside the two provider drivers was not re-audited.

PR #638, opened 2026-07-17
feat(trace): surface OAuth usage-limit pause/resume as witness trace phases

griffinwork40:
## What

Make an OAuth **usage-limit pause legible in the witness trace**. Today, when Anthropic returns a subscription-cap 429, the retry layer parks the turn for up to two hours (waiting for the window to reset, or a keychain hot-swap) and then resumes — but `afk trace show` emits **no event** for any of it. The turn just stops emitting. AFK.md documents this exact gap: *"a usage-limit pause has no event (silent gap)."*

This PR emits two new `session_phase` events from `RetryLayer`, bracketing the park:

| phase | emitted at | payload |
|-------|-----------|---------|
| `usage_limit_pause` | both pause points — reset-timestamp path (`retry-layer.ts`) **and** no-timestamp path | `metadata: { reason, source, hasResetTimestamp, autoResume, resetsAt? }` |
| `usage_limit_resume` | both resume points | `durationMs` (parked time) + `metadata: { source, hotSwapped }` |

Both are **fire-and-forget** (`void emitSessionPhase(...)`) so trace latency can never stall the pause/resume, mirroring the `rate_limit` phase the transient-429 backoff path already emits (`retry-layer.ts:285`).

`afk trace show` renders both as **high-signal** — in the default view, before the `--all` gate — because a multi-hour park is the single most important stall to surface:

```
paused    usage-limit  resets 2026-06-05T14:30:00.000Z
resumed   usage-limit  parked 2.0h
```

## Why this shape

- **Dedicated phases, not overloaded `rate_limit`.** A 2-hour subscription park and a 5-second retry-after backoff are semantically different; folding both into `rate_limit` would conflate them in the throttle counter and the renderer. Resume also has no natural `rate_limit` meaning.
- **Additive / backward-compatible.** New names added to both `SessionPhaseName` (`trace/types.ts`) and its `z.enum` (`trace/events.ts`). These are **trace-internal** — NOT the frozen cross-surface JSONL provenance schema — so this is a safe additive change. No new `@anthropic-ai/sdk` imports; no env changes.
- A pause may legitimately end **without** a resume (auto-resume off, abort, or the 2-hour cap surfacing the error), so a lone `usage_limit_pause` is expected and documented inline.

## Context

Part of the agent-experience (AX) error-legibility arc (model-as-user / harness-as-app). Follows #635 (silent stream truncation). Closes the "usage-limit pause emits no trace event" finding.

## Testing

- `session-phase.test.ts` — schema acceptance for both new phases (payload + bare name schema).
- `query-auth-retry.test.ts` — drives the real `429+ts -> paused -> resumed` cycle with an `InMemoryTraceWriter` and asserts both phases land, with a numeric `durationMs` on resume. This exercises the actual retry-layer wiring, not a mock.
- `trace.test.ts` — asserts both phases render in the **default** view (no `--all`) with reset deadline / parked duration, and that ordinary latency phases stay hidden.

Gates (local, off `origin/main` @ v5.55.0):
- `tsc --noEmit` (strict) — clean
- 642 tests green across `src/agent/trace/` + `src/agent/providers/anthropic-direct/`
- `audit:sdk:check` OK (no new SDK symbols) · `audit:env:check` OK (0 violations)

### DIFF
--- a/src/agent/providers/anthropic-direct/query-auth-retry.test.ts
+++ b/src/agent/providers/anthropic-direct/query-auth-retry.test.ts
@@ -22,6 +22,7 @@ import {
   AnthropicDirectProvider,
   __setAnthropicClientFactory,
 } from './index.js';
+import { InMemoryTraceWriter } from '../../trace/writer.js';
 
 // --- Mock SDK plumbing ---
 
@@ -678,6 +679,47 @@ describe('AnthropicDirectProvider — turnWithUsageLimitRetry', () => {
     expect(types).toContain('turn.completed');
     expect(events.filter((e) => e.type === 'error')).toHaveLength(0);
     expect(messagesCreateMock).toHaveBeenCalledTimes(2);
+  });
+
+  it('records usage_limit_pause + usage_limit_resume trace phases across the park', async () => {
+    // Same happy-path 429+ts → paused → resumed cycle, but assert the WITNESS
+    // TRACE now brackets the park. Before this fix the multi-hour pause left no
+    // trace event — the turn just stopped emitting (the documented silent gap).
+    let callIdx = 0;
+    messagesCreateMock.mockImplementation(() => {
+      callIdx += 1;
+      if (callIdx === 1) throw make429UsageLimitError(5 * 60 * 1_000);
+      return fromArray(makeTextStream('resumed successfully'));
+    });
+
+    const traceWriter = new InMemoryTraceWriter();
+    const provider = new AnthropicDirectProvider();
+    const query = provider.query({
+      prompt: singleInput('hello'),
+      config: {
+        model: 'claude-sonnet-5',
+        apiKey: 'sk-ant-oat01-test',
+        autoResumeOnUsageLimit: true,
+        traceWriter,
+      },
+    });
+
+    const collectPromise = collect(query);
+    await vi.runAllTimersAsync();
+    await collectPromise;
+
+    const phases: string[] = [];
+    let resumeDurationMs: number | undefined;
+    for (const e of traceWriter.events) {
+      if (e.kind !== 'session_phase') continue;
+      phases.push(e.payload.phase);
+      if (e.payload.phase === 'usage_limit_resume') resumeDurationMs = e.payload.durationMs;
+    }
+    expect(phases).toContain('usage_limit_pause');
+    expect(phases).toContain('usage_limit_resume');
+    // The resume phase carries the parked duration (>= 0, mirrors *_done events).
+    expect(resumeDurationMs).toBeTypeOf('number');
+    expect(resumeDurationMs).toBeGreaterThanOrEqual(0);
   });
 
   it('auto-resume=false: 429+ts → paused → original error, no resumed', async () => {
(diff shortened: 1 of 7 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

Clean, well-scoped additive feature (7 files, 273 lines). All four emit points, payloads, renderer placement, schema acceptance, and the backward-compatibility claim were verified by reading the files at the reviewed ref (`b6dd144`). No blocking, critical, or high findings. Three low/nit findings, all additive — none block merge.

### What was verified
- **Correctness** — Both resume paths measure the park with monotonic `Date.now()` deltas; `durationMs` cannot go negative (which matters: `SessionPhasePayloadSchema.durationMs` is `z.number().nonnegative()`, so a negative value would throw at parse). reset-ts path anchors `pausedAt` (retry-layer.ts:434) before its pause emit; no-ts path anchors `startedAt` (retry-layer.ts:343).
- **Spec-compliance** — All 4 emit points present (no-ts pause 329, no-ts resume 401, ts pause 450, ts resume 502) with payloads matching the PR's table. No scope creep.
- **API-compat / schema-freeze** — The PR's claim that these enum additions are trace-internal (not the frozen cross-surface JSONL provenance schema) is verified: the only frozen provenance contract is the `surface:'afk'|'plugin'` tag (events.ts:415), explicitly orthogonal to `SessionPhaseName`. The trace reader is lenient (`looksLikeEvent`, trace.ts:105) with a `default:` render case (trace.ts:485) — no exhaustive validator rejects the new names. Additive-safe.
- **Renderer** — Both new blocks sit before the `if (!ctx.showAll) return null` gate (trace.ts:467), so pause/resume render in the default view as intended; `fmtDuration`/`line` in scope; no fallthrough.
- **Perf/observability** — `void emitSessionPhase(...)` cannot reject: the emit helper swallows writer errors internally (routes to debugLog). Consistent with the pre-existing `rate_limit` emit (retry-layer.ts:285).

### Findings (non-blocking)
- **low · correctness · retry-layer.ts:329-343** — In the no-timestamp path, `startedAt` (line 343) is captured *after* the `usage_limit_pause` emit (line 329), so `durationMs` (line 402) excludes the pause-emit interval and slightly under-reports parked time vs. the reset-ts path (which anchors `pausedAt` before its emit). Impact is sub-millisecond and always ≥0. Suggestion: capture `startedAt` before the line-329 emit to mirror the reset-ts ordering.
- **low · test-coverage · query-auth-retry.test.ts:684-723** — The trace-phase drive test exercises only the reset-ts resume path; the no-ts resume path's `durationMs` (the `startedAt` branch, retry-layer.ts:400-404) — the one path with the timing asymmetry above — has no `durationMs` assertion. Suggestion: add a no-ts drive asserting a numeric `usage_limit_resume` durationMs.
- **nit · test-coverage · query-auth-retry.test.ts** — A lone `usage_limit_pause` with no resume (auto-resume off / abort / 2h cap — called out as expected in the PR and types.ts) is covered at the provider-event level but no test asserts the *trace phase* fires alone. Suggestion: in the auto-resume=false test, thread a traceWriter and assert phases contains `usage_limit_pause` and not `usage_limit_resume`.

### What was not checked
- Reviewed against branch HEAD `b6dd144` — all citations verified at that ref by reading the files directly in a detached worktree.
- Stated intent: PR #638 title + body — spec-compliance assessed against it.
- Did not run the test suite (the PR reports 642 green locally + clean `tsc`, `audit:sdk:check`, `audit:env:check`); did not review the Telegram surface or docs-site rendering.

PR #467, opened 2026-07-07
fix(tui): merge post-ESC type-ahead instead of last-wins so soft-stop never drops a typed message

griffinwork40:
## Problem

PR #403 fixed the post-ESC one-behind backlog with **last-wins** coalescing: during the soft-stop settle window, each Enter replaced the pending post-ESC payload. That prevented the backlog but introduced silent message loss.

Field signature (v5.25.0 postmortem, stuck-subagent session): user ESCs a turn with foreground subagents running → the cancel settle takes tens of seconds (children can sit in the SDK's non-abortable 429 retry sleeps — `client.js` `await sleep()` honors retry-after up to <60s and only observes abort after the sleep) → user types a real instruction + Enter → nothing appears to send → user pokes with `.` + Enter → **the poke replaces the instruction**. The instruction is silently discarded and the user retypes it: "it didn't send," round 2.

## Fix

Coalesce = **merge**, not last-wins (`terminal-compositor.input-dispatch.ts`, handleEnter soft-stop branch):
- All post-ESC submissions (entries at/above `softStopQueueBase`) fold into ONE payload: texts join with newlines, attachments concatenate, `displayText` merges only when a constituent carried one (preserving the "absent when identical" contract).
- **Unchanged invariants:** exactly one next turn drains (no backlog); pre-ESC queue entries below the base are untouched (handleEscape contract); normal NO-ESC type-ahead still accumulates as sequential turns.

## Tests

- Updated both last-wins pins to assert merged text (`first\nsecond\nthird`; `post1\npost2\npost3` after a surviving pre-ESC entry).
- New regression: the exact field signature — `fix the bug` + Enter, then `.` + Enter → one turn with `fix the bug\n.`.
- Pre-ESC preservation guards and NO-ESC blast-radius guards untouched and green.
- Full suite: 596 files, 11,136 passed / 0 failed; `tsc --noEmit` clean.

## Not in scope (deliberate)

The settle window's *length* (SDK-internal retry sleeps are not abort-aware; `maxRetries: 0` + letting the abort-aware retry-layer own replays would fix it) is a separate, larger-blast-radius change — this PR only guarantees nothing typed during that window is lost.

Part 3 of the stuck-subagent fix series (#465 hang bounds → #466 trace visibility → this).

PR #81, opened 2026-06-11
fix(tui): ESC soft-stop cancels on first press, no post-cancel lag

griffinwork40:
## Summary

Fixes ESC-to-cancel (REPL soft-stop), which had **three compounding root causes** that together made cancellation unreliable, dropped the user's next message, and degraded the rest of the session:

- **Double-press to cancel.** `handleEscape`'s dropdown-dismiss branch `return`ed early, so the first ESC was swallowed whenever the autocomplete dropdown was open — which it almost always is mid-stream (ghost-text repopulates on nearly every keystroke). ESC now falls through to the soft-stop path after dismissing the dropdown; the idle-mode and once-only guards still hold. (`src/cli/terminal-compositor.input-dispatch.ts`)
- **ESC unresponsive during long tool calls.** The soft-stop handler only set a flag; `session.interrupt()` was deferred into the stream `for-await` loop, so during a long tool call (loop blocked awaiting the next event) ESC did nothing until the next token arrived, and the compositor lingered in streaming mode. `interrupt()` now fires synchronously in the handler (it is idempotent — `AgentSession.interrupt()` early-returns once state leaves `streaming`/`processing`), and the loop just `break`s. The same fix is applied to the `/skill` dispatch turn path (`src/cli/slash/_lib/run-skill-dispatch-turn.ts`).
- **"Looks like it sends but no turn starts" (the session-wide lag).** The `setInputMode` soft-stop drain guard *de-queued* a message the user typed + Entered during the ESC interrupt window — clearing the queued flag and holding the text as a draft that required a **second** explicit Enter. That is the user's exact report: "I send a message, it looks like it sends, but the agent's turn doesn't start; I have to send a follow-up for it to respond to the first one." The guard no longer de-queues: a queued buffer now **auto-submits** as the next turn, identical to normal mid-turn type-ahead. Safe specifically because of the synchronous-interrupt fix above, which closes the deferred-interrupt window that previously could pile queued buffers into a perpetual off-by-one. (`src/cli/terminal-compositor.input-mode.ts`)

> An earlier pass concluded this third case needed no change (the drain guard "already handled it"). A follow-up `/diagnose` re-derived it from the user's report and confirmed the de-queue **was** the dropped-message symptom; reversing it to auto-submit is the fix. The 5 soft-stop tests that pinned the old draft-preservation semantics were updated to assert auto-submit (a typed-but-**not**-Entered interrupt-window buffer still stays an editable draft).

[...]

PR #644, opened 2026-07-18
fix(tui): coalesce post-ESC type-ahead across the teardown boundary (kill lone-"." input lag)

griffinwork40:
## Symptom

ESC to stop a turn → your next message appears not to send → you poke with `.` to force it. Today's REPL transcripts are littered with lone-`.` user turns (session `bb44050a` alone has **10 separate `{"user":"."}` turns**, spanning minutes).

## Root cause (diagnosed, not guessed)

**This is _not_ a reverted fix.** `#467`'s coalesce-merge is present and byte-unchanged since `064ea20` (`git log 064ea20..HEAD` on both input files is empty across 238 commits). It's a **design-scope gap** all three prior fixes (#81 → #403 → #467) shared:

- The post-ESC coalesce is gated on `softStopped`, which is **cleared at the first teardown `→ idle`** (`terminal-compositor.input-mode.ts`).
- But `session.interrupt()` is an **async abort-request**, not a synchronous stream halt: `agent-session.ts:744` → `anthropic-direct/query.ts:604` `requestAbort()` → trips an `AbortController`; the stream loop only breaks on its **next signal-poll** (`loop.ts`). For subagent/skill turns, `cancelActiveForeground()` awaits each child's settle — **seconds**. The compositor lingers in `'streaming'` the whole time.
- A poke typed **after** `softStopped` clears but **before** the redirect drains hit the plain `pendingSubmissions.push()` path and **stranded one-turn-behind** — a separate turn per poke, draining one per `→ idle`. That's the lone-`.` backlog.

The old coalesce also keyed the merge on an **index** (`softStopQueueBase`) that went stale after any FIFO `shift()`.

<details><summary>Evidence (all 4 load-bearing claims independently re-derived via <code>/shadow-verify</code>)</summary>

| Claim | Verdict | Proof |
|---|---|---|
| #467 present & unchanged | ✅ | `064ea20` ancestor of HEAD; empty diff across 238 commits |
| `softStopped` set once (`:340`), cleared at both turn boundaries; coalesce gated on it; drains one-per-turn | ✅ | `input-dispatch.ts:340/702/707`, `input-mode.ts:178/216/243` |
| `interrupt()` is async abort-request | ✅ | `agent-session.ts:744` → `query.ts:604` → `abort-coordinator.ts:124`; loop-break at `loop.ts:595…` |
| The `.` messages were separate turns, not merged | ✅ | 10 discrete `{"user":"."}` objects in `bb44050a`, zero merged `.\n.` blobs |

</details>

## Fix

Replace the `softStopped`-gated, index-based coalesce with a **payload-reference coalesce _epoch_**:

- `postEscCoalesce` is armed at ESC and **held until the coalesced redirect payload actually drains** to a running turn — not cleared at the teardown `→ idle`.
- `postEscPayload` tracks the single merge target **by reference**, so it survives pre-ESC payloads draining ahead of it.

Post-ESC pokes now merge into the pending redirect across the teardown boundary → **one next turn, no backlog**. Pre-ESC payloads are never the merge target, so they still drain as their own sequential turns; normal no-ESC type-ahead is untouched (sequential-turn delivery preserved).

## Verification

- **New regression test** (`terminal-compositor.keypress.test.ts`): fails on HEAD (`getPendingCount() === 2`, stranded) and passes with the fix (`1`, merged as `redirect\n.`) — the bug is reproduced, then closed.
- **Leak-guard test**: the epoch ends when the redirect drains (later type-ahead is sequential again).
- All **12 existing soft-stop invariants** pass unchanged (pre-ESC preservation, merge-not-last-wins, NO-ESC accumulation, empty-ESC idle flush, …).
- **424 compositor + turn-handler + loop-iteration tests** pass; strict `tsc --noEmit` clean.
- Full suite: 12278 pass. The 1 failure (`agent/trace/writer.test.ts` subprocess-spawn e2e) is **pre-existing and environmental** — it fails identically on pristine HEAD with zero changes, and this PR touches no `agent/trace` code.

## Scope / residual (maintainer decision)

This closes the verified defect: pokes typed **while waiting for the redirect to start** now fold into it. One residual remains that this PR intentionally does **not** change: if the redirect has already drained and its turn is running *long*, further pokes are normal type-ahead again and queue separately. Fully eliminating lone-`.` there needs either (a) prompt ESC→idle teardown (async lifecycle work, larger blast radius) or (b) a more aggressive "coalesce until a genuinely idle prompt" policy. Flagged for your live-TTY judgment rather than baked in here.

🤖 Diagnosed + fixed via agent-afk (`/diagnose` → `/shadow-verify`).

### DIFF
--- a/src/cli/terminal-compositor.input-dispatch.ts
+++ b/src/cli/terminal-compositor.input-dispatch.ts
@@ -109,14 +109,15 @@ export interface KeyDispatchHost {
   /** Once-only soft-stop guard for ESC in streaming mode. */
   softStopped: boolean;
   /**
-   * Snapshot of `pendingSubmissions.length` taken at ESC soft-stop time
-   * (handleEscape). Entries at indices `0..softStopQueueBase-1` were queued
-   * BEFORE esc and are contract-protected (handleEscape comment lines 318-327):
-   * "Already-queued messages: left untouched." Post-ESC Enters coalesce by
-   * MERGING everything at/above this base into one payload, so neither the
-   * pre-ESC queue nor earlier post-ESC messages are ever silently dropped.
+   * Post-ESC coalesce epoch — armed at ESC soft-stop, held until the coalesced
+   * redirect payload drains (authoritative field docs on the class in
+   * terminal-compositor.ts). While true, handleEnter MERGES each new message
+   * into {@link postEscPayload} rather than pushing a separate FIFO entry, so
+   * post-ESC pokes fold into ONE next turn instead of stranding one-per-turn.
    */
-  softStopQueueBase: number;
+  postEscCoalesce: boolean;
+  /** Merge target for the {@link postEscCoalesce} epoch (by reference), or null before the first post-ESC Enter / after it drains. */
+  postEscPayload: SubmissionPayload | null;
   /** Hard-abort flag set by Ctrl+C in streaming mode. */
   canceled: boolean;
   /** Once-only guard for Ctrl+B background in streaming mode. */
@@ -338,10 +339,15 @@ function handleEscape(self: KeyDispatchHost, key: KeyInfo): boolean {
   // committing it would fling it as a turn the user never submitted. Ctrl+C
   // (handleInterrupt below) follows the same no-auto-commit rule.
   self.softStopped = true;
-  // Snapshot the queue length so post-ESC coalesce (handleEnter) can merge
-  // everything at/above this base — preserving pre-ESC payloads per the
-  // contract above while folding post-ESC messages into one next turn.
-  self.softStopQueueBase = self.pendingSubmissions.length;
+  // Arm the post-ESC coalesce epoch. Unlike `softStopped` (cleared at the first
+  // teardown → idle), this is held until the coalesced redirect payload drains,
+  // so pokes typed AFTER teardown but before the redirect runs still merge in
+  // (the residual the softStopped-window coalesce missed — the lone-"." symptom).
+  // A fresh epoch starts with NO merge target: any pre-ESC payloads already on
+  // the FIFO stay as their own entries (handleEscape contract — "already-queued
+  // messages: left untouched"), and the first post-ESC Enter creates the target.
+  self.postEscCoalesce = true;
+  self.postEscPayload = null;
   if (self.onSoftStop) self.onSoftStop();
   return true;
 }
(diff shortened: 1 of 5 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

Reviewed at ref `d00822c6` (PR head; the working tree is on `main`, so all file-state was read from the git object store via `git show <ref>:<path>`, not disk). Regime: light (hotfix, 169+/61− across 5 files — 4 source + 1 test). Every citation below was independently re-derived at the ref in a second verification pass (`git show`/`git grep`); no finding is fabricated or diff-only.

The change does what the PR says: it replaces the `softStopped`-gated, index-based (`softStopQueueBase`) post-ESC coalesce with a **payload-reference epoch**. `postEscCoalesce` is armed at ESC (`input-dispatch.ts:349-350`) and **held across the teardown → idle transition** rather than cleared there (`input-mode.ts:225-236`), and `postEscPayload` tracks the single merge target **by reference** so pre-ESC payloads that drain ahead of it are never the merge target. The epoch is cleared on all three normal exits — drain (`input-mode.ts:261-263`), idle→streaming-when-empty (`input-mode.ts:187`), and `resetState` (`reset.ts:69-70`). Reference-equality is sound: each Enter mints a fresh `SubmissionPayload` literal and the type carries no identity field, so `indexOf` can't collide across turns. **Spec-compliance: satisfied** — the payload-reference epoch, the hold-past-teardown, and the pre-ESC-never-merged guarantee are all implemented; no unmet clause and no scope creep (the diff touches only the coalesce mechanism the intent names). The new regression + leak-guard tests assert the claimed post-fix values (`redirect\n.` merged, then sequential `2`).

### Findings

**1. [medium · correctness] `terminal-compositor.input-dispatch.ts:449-456` (interacts with `709-716`) — ↑-recall of the coalesced payload can resurrect stale text on re-send.** The up-arrow history-recall branch pops a pending submission off the queue when the buffer is empty, but does **not** null `postEscPayload`:
```ts
if (self.pendingSubmissions.length > 0 && self.input.buffer.length === 0) {
  const payload = self.pendingSubmissions.pop()!;   // postEscPayload still points at this
```
On a subsequent Enter in the same epoch, the coalesce branch hits the defensive `idx < 0` path (the target is no longer in the queue) and re-merges the stale reference:
```ts
const merged = mergeSubmissionPayloads([self.postEscPayload, payload]);
if (idx >= 0) self.pendingSubmissions[idx] = merged;
else self.pendingSubmissions.push(merged);   // resurrects the popped "redirect" text
```
So `ESC → type "redirect" → ↑ (to edit) → re-send "redirect NOW"` yields `redirect\nredirect NOW` — the text the user pulled back to change is prepended again. This is **cosmetic wrong-output under a reachable-but-uncommon interaction** (not data loss, not a crash, and ↑-during-epoch is a narrow window), so it does **not** gate the merge. Recommended follow-up: in the ↑-recall branch, clear the epoch target before seeding the buffer, e.g. `if (payload === self.postEscPayload) self.postEscPayload = null;`.

**2. [low · test-coverage] `terminal-compositor.keypress.test.ts:766-828` — the ↑-recall × post-ESC-epoch interaction is untested.** The two new tests cover post-teardown poke-merge (`:766`) and post-drain-sequential leak-guard (`:801`); the up-arrow recall tests live in a disjoint `describe('history navigation')` block (`:1043+`) and emit no ESC, so nothing exercises the path in finding #1. Suggest one case: `ESC → Enter "redirect" → ↑ → assert postEscPayload cleared and a re-Enter does not resurrect prior text`.

**3. [nit · api-compat] `softStopQueueBase` rename is complete.** The retired field name survives at the ref only in two prose comments (`input-dispatch.ts:702`, `terminal-compositor.ts:104`); `git grep` finds zero live identifier references. No action needed.

### What was not checked
- Read against PR head `d00822c6`; all citations verified at that ref in a second independent pass (working tree is dirty `main`, so post-merge state came from `git show`/`git grep`, not disk).
- Stated intent: PR #644 title + body — spec-compliance assessed and satisfied.
- The medium finding is a **code-path/reachability argument derived by reading, not executed** — no test demonstrates the resurrection at runtime (that is finding #2). It is a code-substantiated latent edge case, not a proven-in-CI regression.
- Tests were not run; the PR reports the suite green (12278 pass, tsc-strict clean, the one `agent/trace/writer` failure documented as pre-existing/environmental). The async teardown-timing claims about `cancelActiveForeground()` (subagent-executor, outside the changeset) were not re-verified.

Net: the fix is correct, spec-compliant, and safe to merge. The one substantive item is a medium, pre-existing-shaped edge interaction (↑-recall re-merges the coalesced payload) that the epoch design newly exposes on an uncommon path — worth a small lazy-clear follow-up plus a test, but it does not block this PR.

PR #648, opened 2026-07-18
perf(cli): async at-file autocomplete scan with per-dir cache + stale guard

griffinwork40:
## Summary
`@file` autocomplete ran `readdirSync` + per-candidate `statSync` on every keystroke while an @-token was active — synchronous fs on the input thread; large or network-mounted directories froze typing. The scan is now async (`fs.promises.readdir` with `withFileTypes`, `statSync` eliminated) with a 2s per-directory cache and a stale-result guard.

## Changes
- `input/trigger.ts`: pure `buildFileCandidates` core (ordering byte-identical to `fileMatchesFor`); Dirent-based dir flag; module-level cache (TTL 2s, `invalidateFileScanCache()`); `filterFileCandidatesCached` serves fresh hits synchronously so same-dir keystrokes stay instant; async path never rejects (fs errors → `[]`).
- `terminal-compositor.autocomplete.ts`: `updateFileCandidates` snapshots the query, applies async results only if the live trigger still matches (mirrors the existing `getGhost().then` pattern), repaints on apply; cache invalidated on dropdown accept.
- Legacy sync `reader.ts` path keeps a sync (now statSync-free) variant that warms the same cache.

## Verification
- `pnpm lint` clean; gate green: trigger (+14 tests incl. cache-spy + TTL), autocomplete (converted to await, +1), new stale-guard test file (deferred-promise A-after-B discard), dropdown render/apply suites. **Full `src/cli/` suite green: 270 files / 4,564 tests.**

## Notes for reviewer
- Symlinked directories lose the trailing `/` (Dirent.isDirectory tradeoff) — documented in-code.
- Cache-miss leaves the prior dropdown for one frame (anti-flicker, matches ghost behavior).

### DIFF
--- a/src/cli/input/trigger.test.ts
+++ b/src/cli/input/trigger.test.ts
@@ -12,11 +12,22 @@
  * to 12, so a cwd with >12 entries silently hid the rest.
  */
 
-import { describe, it, expect, beforeEach, afterEach } from 'vitest';
-import { mkdirSync, writeFileSync, rmSync, existsSync } from 'fs';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import { mkdirSync, writeFileSync, rmSync, existsSync, promises as fsp } from 'fs';
 import { join } from 'path';
 import { tmpdir } from 'os';
-import { detectTrigger, filterFileCandidates, filterSlashCandidates } from './trigger.js';
+import {
+  detectTrigger,
+  filterFileCandidates,
+  filterFileCandidatesAsync,
+  filterFileCandidatesCached,
+  filterSlashCandidates,
+  buildFileCandidates,
+  invalidateFileScanCache,
+  __fileScanCacheSize,
+  FILE_SCAN_TTL_MS,
+  type FileDirent,
+} from './trigger.js';
 import { MAX_FILE_MATCHES } from '../multi-line-reader.js';
 import { resetRegistry } from '../slash/registry.js';
 import { registerAll } from '../slash/index.js';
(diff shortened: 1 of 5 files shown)


griffinwork40 (commented):

**Merge decision: MERGE** — reviewed at head `4ae774f`. All CI green; the async scan + per-dir cache + stale-guard mechanism is correct and well-tested. One non-blocking `low` follow-up below.

Regime: full (632-line diff). Stated intent (PR title + body) assessed. Production files `trigger.ts` / `terminal-compositor.autocomplete.ts` are byte-identical to the first commit `23e5212`; only the test file changed in `4ae774f`.

**Findings**

1. **low · perf/resource · `src/cli/input/trigger.ts:166,249,270,275,302`** — the module-level `scanCache` Map has no size bound and no per-key eviction. The TTL is a read-freshness gate (`Date.now() - hit.at > FILE_SCAN_TTL_MS`) that *skips* a stale entry but never deletes it; both the async (`:275`) and sync (`:302`) paths `scanCache.set(...)` without evicting other keys, and the only reclaim is the wholesale `invalidateFileScanCache()` (fired on dropdown accept/dismiss). A long agent-turn that `@`-browses many distinct directories without accepting/dismissing accumulates one full `Dirent[]` listing per directory for the process lifetime.
   - Suggestion: `scanCache.delete(scanDir)` when the TTL check fails (evict-on-stale-read), or cap the map with a small LRU (~32 dirs) — the working set is one directory at a time, so a tiny cap suffices. Non-blocking.

2. **correctness / async stale-guard — clean.** `updateFileCandidates` (`terminal-compositor.autocomplete.ts:124-133`) snapshots `requestedQuery` before dispatch and applies results only when `live.query === requestedQuery && ac.suppressedSignature === null` (mirrors the `getGhost().then` pattern). A scan for query A resolving after the user typed on to B is correctly dropped. Directly covered by `terminal-compositor.autocomplete.stale-guard.test.ts` (discard-A-after-B **and** apply-when-still-live).

3. **perf-observability — goal met.** The keystroke hot path no longer calls `readdirSync`/`statSync`: `filterFileCandidatesCached` is pure-cache/synchronous; misses dispatch `fs.promises.readdir(..., { withFileTypes: true })` with `statSync` eliminated (dir flag from `Dirent.isDirectory()`). The legacy sync `filterFileCandidates` (one prod importer, `reader.ts:182`) keeps a `readdirSync` but is `statSync`-free and confined to the between-turn reader surface, as documented.

4. **spec-compliance — PASS, with an obsolete-claim note.** The PR body's "Full `src/cli/` suite green" claim was **false at the first commit `23e5212`** (the Test CI job failed on `terminal-compositor.autocomplete.test.ts:145`, the cold-cache async-resolution test). That was a **flaky test**, not a production bug: `4ae774f` ("fix(test): bound autocomplete waitFor by wall-clock time, not a fixed tick count") rebounds the `waitFor` poller from a fixed 50-tick count to a 5000ms wall-clock budget — the assertion is unchanged; only the poll bound now scales with CI CPU contention. At the current head every CI check (incl. Test) is green, so the claim is now accurate.

5. **api-compat — no break.** New exports are additive; `filterFileCandidates` keeps its sync `Candidate[]` signature. The only behavior delta is the documented, accepted symlinked-directory trailing-`/` loss (`trigger.ts:203-207`).

**What was not checked**: read against head `4ae774f` — citations verified at that ref; did not run the suite locally (relied on CI, green on the current head); did not exercise the TUI at runtime.

PR #655, opened 2026-07-18
feat(router): notify the model when its provider is swapped mid-session

griffinwork40:
## What

When the model is switched mid-session across a provider boundary (`/model gpt-5.x` from a Claude session, or a same-family switch onto a different endpoint), `ProviderRouter` tears down the active inner and rebuilds a new one seeded with a **text-only shadow history**. That rebuild is currently **invisible to the switched-to model**:

- the new inner's `session.init` is swallowed (by design — no re-init surfaced), and
- the carried conversation arrives as anonymous prose (`{user, assistant}` text pairs), with tool-call structure, thinking blocks, and images dropped.

So the new model has **no in-context signal** that (a) it's a different model than the one that produced earlier turns, or (b) the history it inherited lost its structure. In practice this invites identity confabulation (the model narrating its own identity from a stale premise) and over-trusting the flattened history.

## Change

Prepend a **one-turn** "your model was switched" system-context note to the first turn a freshly-swapped inner serves:

> `[System context — not from the user. Your model was switched at the start of this turn: sonnet → gpt-5.5 (provider anthropic-direct → openai-compatible). Earlier turns in this conversation were produced by sonnet; the history carried across the switch is plain text only, so any prior tool calls and extended reasoning are now prose — treat their structure as lost, not authoritative.]`

- **Gated on a genuine rebuild** — fires only when the router actually swaps the inner (signature change: cross-family OR cross-endpoint). Never on a same-signature same-family forward, and never on the first inner construction.
- **Kept out of the shadow history** — the carried user text stays clean; only the outbound content handed to the new inner carries the note.
- **Descriptive, not an instruction**, and **expires after one turn** like any framework nudge.

## Why here (router seam, not `setModel`)

The router is the only layer that knows a rebuild *actually happened* (`needSwap`), and it already owns the swap seam and the outbound push. `setModel` merely records the target string; it can't tell whether the deferred swap will materialize.

## Scope / out of scope

- Additive and reversible; no behavior change when no swap occurs (unchanged sessions pay nothing).
- Does **not** attempt to restore history fidelity — it makes the loss *legible*, not repaired.
- A sibling change (subagent-result provenance header — surfacing which model produced a subagent finding) is intended as a **separate** follow-up PR.

## Tests

`src/agent/providers/router/provider-router.test.ts` (+3):
- notice injected on a cross-family swap (asserts model + provider transition + the real user turn still rides after it);
- **no** notice on a same-family forward (no rebuild);
- notice excluded from the carried shadow history (records the user's real text only.

Full suite green locally: **674 files, 12,346 passed, 0 failed**;  clean.
BODY
)

### DIFF
--- a/src/agent/providers/router/provider-router.test.ts
+++ b/src/agent/providers/router/provider-router.test.ts
@@ -247,6 +247,62 @@ describe('ProviderRouter', () => {
     await router.close();
   });
 
+  it('injects a one-turn model-switch notice into the first turn a swapped inner serves', async () => {
+    const { router, outer } = makeRouter({ model: 'sonnet', apiKey: 'key-anthropic' });
+    const iter = router[Symbol.asyncIterator]();
+    await pullInit(iter);
+
+    await driveTurn(iter, outer, 'first'); // anthropic
+    await router.setModel('gpt-5.5'); // cross-family
+    const events = await driveTurn(iter, outer, 'second'); // swaps to openai
+
+    // The FakeQuery echoes the pushed user content, so the notice + real text
+    // both surface in the swapped inner's first-turn output.
+    const delta = events.find((e) => e.type === 'delta.text');
+    const text = delta && delta.type === 'delta.text' ? delta.text : '';
+    expect(text).toContain('Your model was switched');
+    expect(text).toContain('sonnet → gpt-5.5');
+    expect(text).toContain('anthropic-direct → openai-compatible');
+    // The user's real message still rides the same turn, after the notice.
+    expect(text).toContain('second');
+    await router.close();
+  });
+
+  it('does NOT inject a switch notice on a same-family forward (no rebuild)', async () => {
+    const { router, outer } = makeRouter({ model: 'sonnet', apiKey: 'key-anthropic' });
+    const iter = router[Symbol.asyncIterator]();
+    await pullInit(iter);
+
+    await driveTurn(iter, outer, 'first');
+    await router.setModel('opus'); // same family (anthropic) → forwarded, not rebuilt
+    const events = await driveTurn(iter, outer, 'second');
+
+    const delta = events.find((e) => e.type === 'delta.text');
+    const text = delta && delta.type === 'delta.text' ? delta.text : '';
+    expect(text).not.toContain('Your model was switched');
+    expect(text).toContain('second'); // plain user turn forwarded unchanged
+    await router.close();
+  });
+
+  it('keeps the switch notice out of the carried shadow history (records user text only)', async () => {
+    const { router, outer, anthropic } = makeRouter({ model: 'sonnet', apiKey: 'key-anthropic' });
+    const iter = router[Symbol.asyncIterator]();
+    await pullInit(iter);
+
+    await driveTurn(iter, outer, 'alpha'); // anthropic#0
+    await router.setModel('gpt-5.5');
+    await driveTurn(iter, outer, 'beta'); // swap → openai#0 (notice injected on this turn)
+    await router.setModel('sonnet');
+    await driveTurn(iter, outer, 'gamma'); // swap → anthropic#1 (seeded from shadow history)
+
+    // The rebuilt anthropic inner is seeded with prior turns as TEXT. The 'beta'
+    // turn must be recorded as the user's real text, never the injected notice.
+    const seeded = anthropic.queries[1]!.rec.config.resumeHistory ?? [];
+    expect(seeded.find((t) => t.user === 'beta'), 'beta recorded as clean user text').toBeDefined();
+    expect(seeded.some((t) => t.user.includes('Your model was switched'))).toBe(false);
+    await router.close();
+  });
+
   it('resolves credentials per-family on swap and never leaks the anthropic key to openai', async () => {
     const { router, outer, anthropic, openai } = makeRouter({ model: 'sonnet', apiKey: 'key-anthropic' });
     const iter = router[Symbol.asyncIterator]();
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

`feat(router): notify the model when its provider is swapped mid-session` · reviewed at branch HEAD `ae3fa001` · 2 files, +113/-2 (1 production + 1 test). Light regime (≤300 LOC), all dimensions covered (security · api-compat · correctness · spec-compliance · test-coverage · perf-observability). **No blocking findings.** Citations were read from the git object store at the reviewed ref (`git show ae3fa001:<file>`), not the working tree.

The change does what it says: on a genuine inner rebuild (a cross-family or cross-endpoint `/model` swap), `ProviderRouter` now prepends a one-turn `[System context — … Your model was switched: X → Y (provider …)]` note to the first turn the freshly-swapped inner serves, so the switched-to model has an in-context signal that (a) it differs from the model that produced earlier turns and (b) the carried history was flattened to plain text. It is descriptive context (not an instruction), rides only the swap turn, and is additive — an unchanged session pays nothing.

### Blocking findings
None.

### Non-blocking findings

**LOW · spec-compliance · `provider-router.ts:334-337,353` (ref ae3fa001)** — the notice guard keys on a model-string change (`previousModel !== this.currentModel`), and `needSwap` itself requires `modelChanged` (`currentModel !== activeModel`). A *pure* cross-endpoint re-point on a byte-identical model string therefore cannot fire the notice. I independently confirmed this is **not** a reachable gap: `setModel(model)` (`provider-router.ts:435-446`) keys the routing signature entirely off the model string via `resolveInner(model)`, so a same-string re-point to a different backend does not exist as a code path — every genuine rebuild changes the string, and the guard always passes on it. Net: correct as written; only worth a one-line comment noting same-string endpoint swaps are unreachable, so a future reader doesn't mistake the guard for a bug.
Suggestion (follow-up, non-gating): add a comment at `provider-router.ts:353` that gating on the model-string change is equivalent to gating on `needSwap` here because signatures are string-derived.

**NIT · correctness · `provider-router.ts:356` (ref ae3fa001)** — `this.currentModel ?? '(unset)'` feeds `buildSwitchNotice`'s `currentModel: string` param, but the enclosing guard (`previousModel !== this.currentModel`, line 353) already proves `currentModel` is defined whenever the notice fires, so the `'(unset)'` branch is dead.
Suggestion (follow-up, non-gating): drop `?? '(unset)'`, or mark it dead-defensive, so a future reader doesn't think `'(unset)'` is reachable.

**NIT · test-coverage · `provider-router.test.ts:250-304` (ref ae3fa001)** — the three new cases all drive **string** turn content, so the array-content branch of `prependNotice` (leading `{type:'text'}` block prepended ahead of the original `ContentBlockParam[]`, line ~154) is unexercised. The string path (notice + real text) and the gate (fires on cross-family swap, absent on same-family forward, excluded from shadow history) are all covered.
Suggestion (follow-up, non-gating): add one case pushing `ContentBlockParam[]` content across a swap and assert the notice rides as a leading text block ahead of the original blocks.

- **Notice provably kept out of shadow history.** `pendingUserText = stringifyUserContent(turn.content)` (the REAL text) is set separately from the notice-prepended `pushUserMessage(...)`; `commitShadowTurn()` pushes `{user: pendingUserText, …}` (`provider-router.ts:304-308`) only on `turn.completed`/`error`, never the prepended content. Test at `provider-router.test.ts:287-304` asserts `beta` recorded as clean user text and no `Your model was switched` string in the reseed. **Clean.**
- **Gate fires only on a genuine rebuild.** In the event loop `needSwap ⟺ modelChanged && signature-differs` (`provider-router.ts:335-337`; the `!this.active` disjunct is dead there — `active` is set at first construction, line 316, and after every swap, line 346). The notice guard (`previousModel !== undefined && previousModel !== currentModel`, line 353) makes both first construction (`previousModel` undefined) and a same-family same-signature forward (`modelChanged` false) no-notice. Tests confirm no notice on an `opus` same-family forward (`:271-285`) and a notice on the `gpt-5.5` cross-family swap (`:250-269`). **Clean.**
- **api-compat clean.** `buildSwitchNotice` / `prependNotice` are new module-private functions (2 defs + 2 call sites, all in `provider-router.ts`; zero external importers) — no prior symbol, no breaking change. `pushUserMessage`'s `string | ContentBlockParam[]` signature and `ProviderUserTurn['content']` are unchanged, and `prependNotice`'s return type matches both branches.
- **security clean.** The notice interpolates only model ids (config / `setModel`-derived) and family strings (`provider.name`, in-repo constants) — config-derived, not attacker-controlled. No secret enters the notice (the resolved `apiKey`/`signature` stay in-memory and are never interpolated). No injection/RCE surface.
- **perf clean.** The notice build + prepend run only inside `if (needSwap)` at a swap boundary, gated on `modelChanged`; an unchanged session skips `resolveInner` entirely. Not on a per-render/per-tick hot path.

### What was NOT checked
- Read against branch HEAD `ae3fa001` — all citations verified at that ref from the git object store (the working tree is on another branch). No critical/high findings survived grounding, so shadow-verify was not required.
- Stated intent: PR #655 title + body — spec-compliance **was** assessed; every named requirement (genuine-rebuild gating, shadow-history exclusion, descriptive/one-turn, additive/reversible) has an implementing change, and the one reachability nuance was independently resolved (see the LOW finding).
- The downstream inner adapter's handling of a leading synthetic `{type:'text'}` block was not traced end-to-end (`provider.ts` notes some adapters stringify/normalize array content); the notice-as-text-block is low-risk but not exercised by a test.
- Tests were **not executed** (read-only review). Assertions were judged by reading, not by run results; CI on this head reports all checks green (Lint & Build, Test, PTY scrollback, Docs, Publish Smoke).

Net: correct, spec-compliant, api-compatible, and safe to merge. The three items are optional documentation/test follow-ups, none gating.

PR #656, opened 2026-07-18
feat(subagent): stamp a provenance header on cross-model subagent results

griffinwork40:
## What

A foreground subagent's result returns to the parent model as an **anonymous text blob**. The producing model is recorded only in telemetry/trace — never in the `ToolResult` content the parent's model actually reads. So when the parent reasons over a confident subagent finding, it can't see that the finding came from a cheaper or different model, which is exactly the signal that should prompt independent verification.

This matters most in **mixed-model fan-out** — e.g. a `compose` DAG or an `agent` wave where nodes run on different models (haiku/sonnet/opus). Sibling PR #655 does the parent-facing half of the same idea (notify the model when *its own* provider is swapped mid-session).

## Change

Prepend a compact provenance header to the returned success content:

```
[subagent result · model=sonnet (parent: opus)]

<the subagent's finding…>
```

**Gated on the child model differing from the parent's:**
- Child inherited the parent's model → header is pure noise → **omitted** (same-model dispatches are byte-for-byte unchanged).
- `parentModel` not wired for a surface → **no header ever** (fail-safe).

The parent model flows through the **existing** `ctx.parentModel` wiring, already populated by the REPL, chat, and daemon bootstraps — no new plumbing through the session. The header composes cleanly with `annotateIfIncomplete` (a partial-result marker stays beneath the provenance header). It's descriptive metadata, not an instruction.

## Why gated (not always-on)

An always-on header would (a) add a badge to every result including the common same-model case (noise), and (b) break ~8 existing exact-`content` assertions. Gating on child≠parent targets the case where provenance actually informs a decision, and keeps blast radius near-zero: existing tests don't set `parentModel`, so they get no header and stay green.

## Scope / out of scope

- Success path only. Attaching the model to the **structured failure payload** (error path) is a deliberate, small follow-up — the failure JSON already carries `subagent_id` + status, and adding a field there touches a shared payload builder + its callers, so it's kept separate.
- Additive and reversible; no behavior change when child == parent or `parentModel` is unset.

## Tests

- `src/agent/tools/subagent/foreground-promotion.test.ts` (new): unit tests for the gate — differs, equal, unknown child, unwired parent, and composition with the incomplete-partial marker.
- `src/agent/tools/subagent-executor.test.ts` (+2): integration tests proving the header appears on a cross-model dispatch (`parent: opus`, child `sonnet`) and is omitted on a same-model one — exercising the real `ctx.parentModel` wiring.

Full suite green locally: **675 files, 12,350 passed, 0 failed**; `tsc --noEmit` clean.

### DIFF
--- a/src/agent/tools/subagent-executor.test.ts
+++ b/src/agent/tools/subagent-executor.test.ts
@@ -646,6 +646,45 @@ describe('SubagentExecutor', () => {
       expect(() => JSON.parse(result.content)).toThrow();
     });
 
+    it('stamps a provenance header when the child model differs from the parent model', async () => {
+      // Parent on opus, child inherits the sonnet default → mixed-model fan-out.
+      const exec = new SubagentExecutor({
+        subagentManager: mockSubagentMgr as any,
+        parentSession: mockParentSession as any,
+        defaultConfig: mockConfig,
+        depth: 0,
+        parentModel: 'opus',
+      });
+      const handle = mockHandle({
+        message: { role: 'assistant', content: 'child finding', timestamp: new Date() },
+      });
+      mockSubagentMgr.forkSubagent = vi.fn().mockResolvedValue(handle);
+
+      const result = await exec.execute(makeCall());
+
+      expect(result.isError).toBeUndefined();
+      expect(result.content).toBe('[subagent result · model=sonnet (parent: opus)]\n\nchild finding');
+    });
+
+    it('omits the provenance header when the child model equals the parent model', async () => {
+      // Parent AND child on sonnet → header is pure noise, so it is not added.
+      const exec = new SubagentExecutor({
+        subagentManager: mockSubagentMgr as any,
+        parentSession: mockParentSession as any,
+        defaultConfig: mockConfig,
+        depth: 0,
+        parentModel: 'sonnet',
+      });
+      const handle = mockHandle({
+        message: { role: 'assistant', content: 'same-model finding', timestamp: new Date() },
+      });
+      mockSubagentMgr.forkSubagent = vi.fn().mockResolvedValue(handle);
+
+      const result = await exec.execute(makeCall());
+
+      expect(result.content).toBe('same-model finding');
+    });
+
     // T-1 (review finding C-1): when the SDK returns a ContentBlock[] array
     // instead of a plain string, the executor must serialize it to a string
     // so ToolResult.content is always a valid string.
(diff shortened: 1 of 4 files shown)


griffinwork40 (commented):

**Decision: MERGE** — no blocking, high, or medium findings. The change does exactly what its stated intent claims, verified by reading the files at the reviewed ref (`279f7f08`).

**Summary**

Additive feature: a compact provenance header (`[subagent result · model=<child> (parent: <parent>)]`) is prepended to a foreground subagent's **success** result, gated on the child model differing from the parent's. Reviewed all dimensions (correctness, spec-compliance, api-compat, content-safety, test-coverage, perf/observability).

**Per-dimension result (all clean):**

- **Correctness (gate logic)** — `withProvenanceHeader` returns content unchanged on unknown child, unwired parent, or same-model; stamps only on a genuine mismatch (`foreground-promotion.ts:92-99`). `model` is the CHILD model (`childConfig.model`), resolved to a concrete string (`?? 'sonnet'`, non-optional) at `child-config.ts:165-166` **before** the gate runs — so the integration test's `model=sonnet` expectation is sound and the gate never wrongly omits the header for the unset-model case. Binding order verified.
- **Spec-compliance** — success path only (stamp lives inside the `succeeded` block; error/promotion paths untouched); gated on child≠parent; fail-safe when `parentModel` is unset; composed as `withProvenanceHeader(annotateIfIncomplete(...))` so the provenance header sits **above** the partial-result marker, exactly as the intent and the composition test assert. No scope creep.
- **API-compat** — `parentModel?: string` on `RunForegroundArgs` is optional → additive, non-breaking; the call-site conditional spread (`subagent-executor.ts:670-671`) is consistent with the optional field.
- **Reachability** — production writers of `ctx.parentModel` confirmed at all four surfaces (`bootstrap.ts:233/391`, `chat.ts:449/562`, `daemon.ts:99/174`, `telegram.ts:263/350`), so the header actually fires in prod under mixed-model dispatch.
- **Content-safety** — interpolated `model`/`parentModel` are internal model ids, not attacker-controllable → no injection concern.
- **Test-coverage** — all five gate branches unit-tested (differs / equal / unknown child / unwired parent / composition), plus two byte-exact integration tests (parent=opus/child=sonnet stamped; same-model omitted).
- **Perf/observability** — one conditional string concat on the success path; early return on the same-model path. Negligible.

**Non-blocking notes:**

- LOW · test-coverage — the `!model` short-circuit in the gate is defense-in-depth: `child-config.ts:166` resolves the child model to a concrete string before the gate, so that branch is unreachable-by-construction in prod. The direct unit test is the right way to cover it. No action required.
- NIT · style — the header uses a raw middot (`·`) matching the existing `annotateIfIncomplete` marker style. No concern.

**What was not checked:** citations read from ref `279f7f0804f95c055ff076b242dc2e1e8dfa960c`; the test suite was not executed (read-only review verified test *content*, not runtime pass); `annotateIfIncomplete`'s implementation was not read beyond confirming wrap order (the composition test proves header-above-marker independently).

PR #659, opened 2026-07-19
fix(test): eliminate scheduler.test.ts cron-timer leak causing flaky duplicate mcp_connect_done (#657)

### LINKED ISSUE #657: flaky(test): scheduler.test.ts mcp_connect_* trace-phases test intermittently double-emits mcp_connect_done under full-suite load
## Summary
`src/agent/daemon/scheduler.test.ts > CronScheduler — mcp_connect_* trace phases > emits mcp_connect_start then mcp_connect_done around McpManager.fromConfig` intermittently fails in full-suite CI runs with an extra duplicate `mcp_connect_done` entry:

```
AssertionError: expected [ { …(2) }, { …(2) }, { …(2) } ] to deeply equal [ { …(2) }, { …(2) } ]
  Array [
    Object { "phase": "mcp_connect_start", "serverCount": 1 },
    Object { "phase": "mcp_connect_done", "serverCount": 1 },
+   Object { "phase": "mcp_connect_done", "serverCount": 1 },
  ]
```

## Evidence this is pre-existing and load/timing-dependent, not caused by any specific PR
Confirmed on **`main` itself**, independent of any open PR, on at least 3 separate CI runs:
- 2026-07-18T16:29:26Z — run [29652062573](https://github.com/griffinwork40/agent-afk/actions/runs/29652062573) (`main`)
- 2026-07-14T12:10:59Z — run [29331547230](https://github.com/griffinwork40/agent-afk/actions/runs/29331547230) (`main`)
- 2026-07-12T12:01:51Z — run [29191889535](https://github.com/griffinwork40/agent-afk/actions/runs/29191889535) (`main`)

It also surfaced on PR #649 (`afk/input-core-perf`, run [29652204941](https://github.com/griffinwork40/agent-afk/actions/runs/29652204941)) — that PR's diff only touches `src/cli/_lib/stream-renderer.ts` and `src/cli/input*` files, nothing in `src/agent/daemon/`, confirming the failure is unrelated to that PR's change.

Running `pnpm test src/agent/daemon/scheduler.test.ts` in isolation passes reliably (29/29) — the duplicate only manifests under full-suite CI conditions, suggesting a timing/ordering race that needs CPU contention or wall-clock elapsed time to surface.

## Structural analysis (static read, not yet locally reproduced)
- The only production call site for `phase: 'mcp_connect_done'` is `src/agent/daemon/scheduler.ts:579-585`, a single `try { await McpManager.fromConfig(...) } finally { void emitSessionPhase(..., 'mcp_connect_done') }` inside `spawnSession()`. A single `try/finally` cannot naturally fire its `finally` twice per call, so a duplicate emission requires `spawnSession` itself to run twice for one registered task within the test's window.
- `CronScheduler` registers real `node-cron` jobs on `'* * * * *'` (real wall-clock, not `vi.useFakeTimers()`), per `src/agent/daemon/scheduler.ts` + the `node-cron` dependency visible in the vitest sourcemap warning. Under a slow, CPU-contended full-suite run, it's plausible a live cron interval fires a genuine second background tick for the same task in the narrow window between the test's explicit `await scheduler.tick(...)` resolving and `scheduler.stop()`/`unregister()` clearing the job in the test's `finally` block — though this alone would be expected to also duplicate `mcp_connect_start`, which is not observed, so the exact mechanism is not yet fully pinned down.
- Both `emitSessionPhase` calls in `scheduler.ts` are fire-and-forget (`void emitSessionPhase(...)`, not awaited) "so it never gates the connect" (per the existing code comment) — worth double-checking whether an unawaited promise from this or a neighboring test in the same file (same-file module state is shared across tests even with vitest's default per-file isolation) can outlive its test's teardown under load.

## Suggested next step
Needs a dedicated reproduction attempt under artificial load/CPU throttling (or several repeated full-suite runs) to catch it locally with added instrumentation (e.g. temporarily log a stack trace or a call counter at the `spawnSession` entry point) before landing a fix — a fix without a locally-reproduced failure risks papering over the real bug with a weakened assertion.

## Impact
Low-frequency (3 occurrences in the observed CI history over ~a week) but real — makes an otherwise-green PR's CI look red for unrelated reasons. Recommend re-running the `Test` check when this specific failure is hit rather than blocking merge, until root-caused.

griffinwork40:
Closes #657.

## Problem
`src/agent/daemon/scheduler.test.ts > CronScheduler — mcp_connect_* trace phases` intermittently failed under full-suite CI load with an extra duplicate `mcp_connect_done` (no extra `mcp_connect_start`), while passing 29/29 reliably in isolation.

## Root cause — a test-isolation defect, not a production bug
- `register()` (`scheduler.ts:211`) calls `cron.schedule('* * * * *', …)`. node-cron v4 **auto-starts** the task, arming a **live wall-clock timer**.
- These unit tests only ever execute via the explicit `await scheduler.tick(...)` seam (`scheduler.ts:242`), which calls `runOnce` directly and **bypasses cron** — so the armed live timer is pure noise.
- Under CPU contention, a genuine minute-boundary tick fires a background `runOnce` whose slow `await McpManager.fromConfig(...)` (real subprocess spawn) **outlives `stop()`** — `stop()`/`unregister()` cancel the cron timer but never await an in-flight run.
- That leaked background tick's fire-and-forget `mcp_connect_done` (emitted in the `finally` **after** the slow await) lands in a **sibling test's** module-level `vi.spyOn(traceEmit, 'emitSessionPhase')` capture. Its paired `mcp_connect_start` was emitted before that test's spy existed → **done-only duplicate**, exactly the observed symptom.
- Production is correct: each `spawnSession` opens its **own** trace writer (`scheduler.ts:540`), so real on-disk traces never double-count. The contamination exists only because a *test* spies on the shared module singleton and thus captures foreign-session emissions. vitest per-file isolation rules out cross-file leaks, so the source is entirely within this file.

## Fix (test layer — the correct layer)
File-wide `vi.mock('node-cron')` whose `schedule()` returns an inert task, so `register()` arms **no** real timer. This removes the background-tick source for **every** test in the file, eliminating the whole class of leak.

- **The flaky assertion is unchanged** — still `expect(emitted).toEqual([{start,1},{done,1}])` exactly. No weakening (no `toContainEqual`/dedupe/length checks), per the issue's explicit request not to paper over the bug.
- **No production code touched.** (`stop()` not awaiting in-flight ticks is a real but separate wart; making it drain would risk hanging real daemon shutdown behind a long tick — deliberately left as a possible follow-up, out of scope here.)
- Adds a **deterministic regression guard** (`CronScheduler — cron-timer isolation (#657)`) that fails if the stub is ever removed.

## Verification
- `pnpm test src/agent/daemon/scheduler.test.ts` → **30/30** (29 original + new guard)
- `pnpm test src/agent/daemon/` → **128/128** across 10 files (file-scoped mock; siblings unaffected)
- `pnpm lint` (`tsc --noEmit`) → clean (exit 0)

Diff: 1 file, +49 lines (test-only).

### DIFF
--- a/src/agent/daemon/scheduler.test.ts
+++ b/src/agent/daemon/scheduler.test.ts
@@ -11,6 +11,7 @@ import { tmpdir } from 'node:os';
 import { dirname, join, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import * as traceEmit from '../trace/emit.js';
+import * as nodeCron from 'node-cron';
 
 const schedulerTestState = vi.hoisted(() => ({ cleanupOrder: [] as string[] }));
 
@@ -24,6 +25,25 @@ vi.mock('../default-hook-registry.js', () => ({
     registry: undefined,
     memoryStore: { close: () => schedulerTestState.cleanupOrder.push('memory.close') },
   }),
+}));
+
+// Invariant: these unit tests exercise the scheduler exclusively through the
+// explicit `scheduler.tick(...)` seam, which invokes `runOnce` directly and
+// never needs a live wall-clock cron. node-cron auto-STARTS a real timer at
+// register(); under full-suite load a genuine minute-boundary tick can fire a
+// background `runOnce` whose slow `McpManager.fromConfig` outlives `stop()`
+// (which cancels the timer but never awaits an in-flight run) and leaks its
+// fire-and-forget `mcp_connect_done` into a sibling test's module-level
+// `emitSessionPhase` spy — the intermittent duplicate in issue #657. Stubbing
+// `schedule()` so register() arms no real timer removes that source entirely;
+// every assertion below stays exact.
+vi.mock('node-cron', () => ({
+  schedule: vi.fn(() => ({
+    start: () => {},
+    stop: () => {},
+    destroy: () => {},
+    getStatus: () => 'stopped',
+  })),
 }));
 
 import { CronScheduler, daemonTraceLabel, resolveWorktreePruneRoot } from './scheduler.js';
@@ -867,6 +887,35 @@ describe('CronScheduler — mcp_connect_* trace phases', () => {
     },
     { timeout: 15_000 },
   );
+});
+
+describe('CronScheduler — cron-timer isolation (#657)', () => {
+  it('register() routes cron scheduling through the stubbed node-cron so tests arm no live wall-clock timer', () => {
+    // Regression guard for the flaky duplicate `mcp_connect_done` in issue #657.
+    // If this file's `vi.mock('node-cron', …)` stub is ever removed, register()
+    // would arm a real minute-boundary timer whose background `runOnce` can leak
+    // a fire-and-forget trace phase into a sibling test's `emitSessionPhase` spy
+    // under full-suite load. Assert the stub is in effect AND load-bearing.
+    expect(vi.isMockFunction(nodeCron.schedule)).toBe(true);
+
+    const dir = makeTmpDir();
+    try {
+      const scheduler = new CronScheduler({ telemetryPath: join(dir, 'forge-telemetry.jsonl') });
+      scheduler.register({
+        taskId: 'isolation-guard-657',
+        command: 'noop',
+        trigger: 'cron',
+        cronExpression: '* * * * *',
+      });
+      expect(nodeCron.schedule).toHaveBeenCalledWith(
+        '* * * * *',
+        expect.any(Function),
+        expect.objectContaining({ name: 'isolation-guard-657' }),
+      );
+    } finally {
+      rmSync(dir, { recursive: true, force: true });
+    }
+  });
 });
 
 describe('daemonTraceLabel', () => {


griffinwork40 (commented):

## Merge decision: ✅ MERGE

`fix(test): eliminate scheduler.test.ts cron-timer leak causing flaky duplicate mcp_connect_done (#657)` · reviewed at branch HEAD `4861db2c` · 1 file, +49/−0 (test-only). Light-regime review covering all dimensions (security · api-compat · correctness · spec-compliance · test-coverage · perf-observability). **No blocking, high, or medium findings.** Every citation was read from the git object store at the reviewed ref (`git show 4861db2c:<file>` / `git grep <ref>`), not the working tree.

The change does what it says: a file-wide `vi.mock('node-cron')` returns an inert task so `register()` arms **no** live wall-clock timer, removing the background-tick source that leaked a fire-and-forget `mcp_connect_done` into a sibling test's `emitSessionPhase` spy under full-suite load (issue #657). It is test-only, the flaky assertion is left exactly as-is, and it adds a regression guard.

### Blocking findings
None.

### What was verified (evidence, all at ref `4861db2c`)

- **Mock completeness — clean.** The inert stub shape `{ start, stop, destroy, getStatus }` is a strict **superset** of the `CronTask` members production actually calls. `git grep -n "cronTask\." -- src/agent/daemon/scheduler.ts` returns only `scheduler.ts:228` (`.stop()`) and `:229` (`.destroy()`), both present in the stub; `.start()`/`.getStatus()` are never invoked. Both call sites are wrapped in `Promise.resolve(...).catch()`, so the stub's synchronous returns can't throw. No test can break for a missing mock member.
- **No collateral test breakage — clean (grounded).** No test in the file depends on a live cron timer auto-firing; every scheduler-driving test asserts through the explicit `scheduler.tick(...)` seam (19 `tick()` call sites). `git grep -nE "useFakeTimers|advanceTimersByTime|runAllTimers|runOnlyPendingTimers|vi\.waitFor|waitFor\(" 4861db2c -- src/agent/daemon/scheduler.test.ts` → **zero matches**. Stubbing `schedule()` to inert is therefore safe for the whole file.
- **Regression guard is load-bearing — clean.** Both assertions fail if the stub is removed: `vi.isMockFunction(nodeCron.schedule)` (`scheduler.test.ts:899`) → `false` on the real function, and `toHaveBeenCalledWith(...)` (`:910-914`) throws on a non-mock. The guard also pins the exact call shape (`'* * * * *'`, a function, `{ name: 'isolation-guard-657' }`) matching `register()`'s real `cron.schedule(..., { name: task.taskId })` at `scheduler.ts:211-219`. It cannot pass without the stub in effect.
- **Spec-compliance satisfied — the flaky assertion is UNCHANGED.** The `mcp_connect_* trace phases` assertion is still an exact `toEqual([{start,1},{done,1}])` (`scheduler.test.ts:872-875`); `git grep -n "toContainEqual"` → no matches (no weakening/dedupe/length-check). The diff is **+49/−0** (zero deletions), so the pre-existing assertion could not have been edited, and `git show --stat` lists **only** `src/agent/daemon/scheduler.test.ts` — no production file, no scope creep.
- **api-compat — clean.** vitest `vi.mock` factories are module-registry-scoped per test file, so this mock affects only `scheduler.test.ts`; production `scheduler.ts:21 import * as cron from 'node-cron'` resolves the real module at runtime, unaffected. No sibling test file impacted.
- **security — clean.** Test-only; the mock factory is a static inline literal and the new test uses a hardcoded `taskId`/`command:'noop'` with the established `makeTmpDir()`/`rmSync` cleanup. No injection, secret, env, or network surface.
- **Root-cause narrative corroborated.** `register()` fires-and-forgets `runOnce` (`scheduler.ts:216`), which awaits `McpManager.fromConfig` (`:575`) and emits `mcp_connect_done` in a `finally`; `unregister()` cancels the timer without awaiting the in-flight run (`:228-229`) — exactly the leak path the PR describes. `package.json` pins `node-cron ^4.2.1`, consistent with the "v4 auto-starts" claim.

### Non-blocking notes
- All findings are `low`/`nit` (no action required). The fix eliminates the documented leak source with zero production-code risk and zero collateral test breakage.

### What was NOT checked
- Read against branch HEAD `4861db2c` — all citations verified at that ref from the git object store (the working tree is on another branch).
- Tests were **not executed** (read-only review): the PR's "30/30 · 128/128 · lint clean" claims were judged by reading the assertions as structurally sound and load-bearing, not by running vitest.
- `node-cron`'s internal v4 auto-start behavior was inferred from the pinned version + PR narrative, not read from `node_modules`; the fix is robust regardless (an inert stub arms no timer under any interpretation).
- vitest per-file module-mock isolation is taken as documented framework behavior; if it were wrong, the blast radius is confined to test runs, never production.
- Stated intent: PR #659 title + body — spec-compliance **was** assessed and satisfied.

Net: correct, spec-compliant, api-compatible, and safe to merge. The change is test-only, eliminates the flaky-duplicate leak at the correct (test) layer, and guards against regression.

PR #663, opened 2026-07-21
fix(#661): bound web_scrape + tool output to prevent subagent context overflow

### LINKED ISSUE #661: web_scrape can overflow a subagent's context on large responses (observed 1,065,401-token crash)
## Gap

`web_scrape` can pull a response large enough to **overflow a subagent's context window and crash it**, with no graceful truncation.

## Repro / evidence

A read-only subagent (WebFetch → `web_scrape`, `mode: raw`) fetched a large public JSON endpoint (a category-listing API, ~4MB body). The subagent died with:

```
400 invalid_request_error: prompt is too long: 1065401 tokens > 1000000 maximum
```

i.e. the entire ~4MB response was loaded into the model context. A bounded local fetch of the same endpoint confirms the body is ~4MB and reproduces reliably. The parent then had to re-derive the answer with a size-bounded parse instead.

## Hypothesis (not confirmed root cause)

The tool documents a `max_bytes` default (1MB, ceiling 10MB). Either:
- `max_bytes` isn't effectively enforced on the `raw`-mode / subagent `WebFetch` path, or
- 1MB of dense JSON (short keys, little whitespace) already tokenizes above a safe budget for a **nested** subagent (whose context is also carrying its system prompt + task).

Filing as an observation + hypothesis; the observable defect (subagent crash from unbounded tool output) holds regardless of which.

## Proposed direction

1. Verify `max_bytes` is actually applied on the `raw`/`WebFetch` path (and in subagents).
2. Consider a **lower default** for nested subagents, and/or a hard pre-model size guard that truncates the body with an explicit `... [truncated at N bytes]` marker instead of passing the full payload to the model.
3. Surface the truncation as a visible notice so the agent knows the content was clipped (and can page or filter).

Related: #575 (`web_scrape` SSRF/internal-host filtering) — same tool, a different hardening axis.

_Surfaced 2026-07-21 while a verifier subagent fetched a large JSON API._

griffinwork40:
Fixes #661

## Root cause

`web_scrape` could load a response large enough to overflow a (sub)agent's context window and crash it (observed: a ~4MB JSON body → 1,065,401 tokens > 1,000,000 limit). The tool's byte-cap default (1MB) and ceiling (10MB) translated to token counts (≈286K / ≈2.5M) that exceed context windows, and a caller that raised `max_bytes` toward the 10MB ceiling let a 4MB body straight through. More broadly, this is a **generic crash class**: the dispatcher pushes every tool result's `content` verbatim into the next model request with no size guard, so MCP-bridged tools, browser dumps, `read_file` of a huge file, etc. can overflow a child's context identically.

### Part A — `web_scrape` uses the shared truncation primitive
- `DEFAULT_MAX_BYTES` 1MB → **100KB**; `MAX_MAX_BYTES` 10MB → **1MB**.
- Replaced the local `truncateUtf8`/`TRUNC_MARKER` (head-only slice) with the shared, tested `headAndTail` from `_output-cap.ts` — head+tail with an explicit `… [N bytes truncated: …] …` marker, UTF-8 safe at both cut points. bash/grep already use these primitives.
- All three return sites (raw, markdown, search) now cap at `parsed.maxBytes` and set `ToolResult.truncated: true` when the body exceeded the cap. `max_bytes` remains a real, caller-overridable knob — now within safe bounds.
- Updated the tool schema doc + `max_bytes` description to the new 100KB/1MB bounds and head+tail marker.

### Part C — central fork-scoped backstop (default OFF, main session unchanged)
- Added an optional `maxOutputBytes?: number` to `SessionToolDispatcher`. When set, the shared result path reduces any tool result whose `content` exceeds it to head+tail via `headAndTail` (and stamps `truncated: true`) as the **final step after `PostToolUse` fires** — so hooks still observe the full content, the model sees the capped view, and non-model consumers read the structured flag. It is **idempotent** (content already within budget is returned unchanged, so no double-truncation with web_scrape's own cap) and **never touches `result.image`**. Implemented by splitting the shared `executeCore` into `executeCoreInner` (existing body) + a thin `executeCore` wrapper that applies the cap, so both `execute()` and `executeBatch()` bound every result exactly once.
- **Wired for forks only:** both providers' `buildDispatcher` enable `maxOutputBytes = MODEL_CAP_BYTES` (100KB) exactly when `opts.parentSessionId` is set — which is true only for forked children. The **top-level session leaves it `undefined` ⇒ no central capping ⇒ behavior is unchanged.** `parentSessionId` is already threaded end-to-end from `AgentConfig` to both dispatcher construction sites, so **no new session-config plumbing was required** — the fork auto-wiring is complete (escape hatch NOT taken).

Part A alone closes the observed crash; Part C contains the whole class for the actual victims (forked children, which crash the entire turn on overflow) while keeping the primary session's behavior byte-for-byte identical.

## Alternatives considered
- **Pragmatist** — reuse the existing shared `headAndTail`/`MODEL_CAP_BYTES` primitive rather than inventing a new truncation path (Part A). Chosen: zero new surface, already tested, matches bash/grep.
- **Paranoid** — narrow the blast radius: enable the central cap only for forks (keyed on `parentSessionId`), never the top-level session, so a human-driven session's tool output is never silently truncated. Chosen for Part C's scoping.
- **Architect** — a central dispatcher backstop that bounds the entire tool-output-overflow crash class in one place, rather than patching each overflow-prone tool (MCP, browser, read_file) individually. Chosen as the Part C mechanism.
- **Deferred** — a per-model byte/token budget (cap derived from the child's actual context window). Higher value but needs model-context plumbing; the flat 100KB fork cap is the safe, minimal fix now and can be refined later.

## Testing

`pnpm lint` (`tsc --noEmit`): **pass, zero errors.**

```
✓ src/agent/tools/handlers/web-scrape.test.ts (30 tests) — 30 passed
✓ src/agent/tools/dispatcher.test.ts (102 tests | 1 skipped) — 101 passed, 1 pre-existing skip
```

New coverage:
- web_scrape: a body over the default is truncated to ≤ default with `truncated:true`; an explicit `max_bytes` above the 1MB ceiling is clamped to 1MB; multi-byte UTF-8 stays clean at both cut points.
- dispatcher: `maxOutputBytes:500` truncates an over-budget result + sets `truncated:true`; `undefined` leaves content untouched; content within budget is not truncated (idempotent); `result.image` is preserved; the cap also applies on the `executeBatch` path; non-positive/non-finite values leave the cap off.

Regression check (fork path + dispatcher-adjacent, all green): `plan-mode-gate-wiring`, `openai-compatible/query`, `hook-registry-contract`, `subagent-phase-role`, `tools/integration`, `denial-circuit-breaker` — **144 passed**.

### DIFF
--- a/src/agent/providers/anthropic-direct/index.ts
+++ b/src/agent/providers/anthropic-direct/index.ts
@@ -53,7 +53,6 @@ import type { ToolDispatcher } from './tool-dispatcher.js';
 import { SessionToolDispatcher } from '../../tools/dispatcher.js';
 import { PathGrantManager } from '../../tools/grant-manager.js';
 import { createBuiltinHandlers } from '../../tools/handlers/index.js';
-import { MODEL_CAP_BYTES } from '../../tools/handlers/_output-cap.js';
 import {
   exitPlanModeTool,
   createExitPlanModeHandler,
@@ -344,6 +343,13 @@ export class AnthropicDirectProvider implements ModelProvider {
       env?: Record<string, string>;
       sessionId?: string;
       parentSessionId?: string;
+      /**
+       * Explicit "this session is a forked subagent" signal carrying the
+       * per-result output-cap budget (#661). Set to MODEL_CAP_BYTES by
+       * `SubagentManager.forkSubagent` for EVERY fork; undefined on a top-level
+       * session. Arms the dispatcher's `maxOutputBytes` backstop declaratively.
+       */
+      subagentToolOutputCapBytes?: number;
       traceWriter?: TraceWriter;
       /**
        * Live source for the `get_runtime_state` tool. Constructed per-query
@@ -486,14 +492,21 @@ export class AnthropicDirectProvider implements ModelProvider {
       ...(opts?.env !== undefined ? { env: opts.env } : {}),
       sessionId: opts?.sessionId,
       parentSessionId: opts?.parentSessionId,
-      // Central output-cap backstop (#661), FORK-SCOPED. A forked child is the
-      // only session that carries a `parentSessionId`; the top-level session
-      // leaves it undefined. Enabling the cap exactly when parentSessionId is
-      // set means every FORKED child bounds each tool result at MODEL_CAP_BYTES
-      // (100KB) via headAndTail — containing the whole overflow crash class
-      // (MCP dumps, browser output, read_file of a huge file) for the actual
-      // victims — while the top-level session's behavior is UNCHANGED (no cap).
-      ...(opts?.parentSessionId !== undefined ? { maxOutputBytes: MODEL_CAP_BYTES } : {}),
+      // Central output-cap backstop (#661), FORK-SCOPED. Armed from the
+      // explicit `subagentToolOutputCapBytes` signal that
+      // `SubagentManager.forkSubagent` stamps (as MODEL_CAP_BYTES = 100KB) on
+      // EVERY forked child — the single choke point for the agent-tool, skill,
+      // and compose fork paths. A value here means "forked child" ⇒ bound each
+      // tool result at that budget via headAndTail, containing the whole
+      // overflow crash class (MCP dumps, browser output, read_file of a huge
+      // file). The top-level session is built directly (never via forkSubagent),
+      // leaves this unset, and is therefore UNCAPPED (behavior unchanged). This
+      // replaces the prior `parentSessionId !== undefined` gate, which missed
+      // skill-forked descendants whose parent carries no sessionId (a stub
+      // parent) and were left silently exposed.
+      ...(opts?.subagentToolOutputCapBytes !== undefined
+        ? { maxOutputBytes: opts.subagentToolOutputCapBytes }
+        : {}),
       ...(opts?.traceWriter ? { traceWriter: opts.traceWriter } : {}),
       // Read-only-skill bash gate: forwarded from the provider's stored flag
       // (set by createChildProviderFactory / buildReadOnlyReconProvider) so a
@@ -735,6 +748,12 @@ export class AnthropicDirectProvider implements ModelProvider {
           ...(config.env !== undefined ? { env: config.env } : {}),
           sessionId: config.sessionId,
           parentSessionId: config.parentSessionId,
+          // Fork-scoped central output cap (#661): forwarded from the child
+          // config that forkSubagent stamped, arming maxOutputBytes for forks
+          // only (top-level leaves it unset).
+          ...(config.subagentToolOutputCapBytes !== undefined
+            ? { subagentToolOutputCapBytes: config.subagentToolOutputCapBytes }
+            : {}),
           traceWriter: config.traceWriter,
           runtimeStateSource,
           hookRegistry: config.hookRegistry,
--- a/src/agent/providers/anthropic-direct/index.ts
+++ b/src/agent/providers/anthropic-direct/index.ts
@@ -53,6 +53,7 @@ import type { ToolDispatcher } from './tool-dispatcher.js';
 import { SessionToolDispatcher } from '../../tools/dispatcher.js';
 import { PathGrantManager } from '../../tools/grant-manager.js';
 import { createBuiltinHandlers } from '../../tools/handlers/index.js';
+import { MODEL_CAP_BYTES } from '../../tools/handlers/_output-cap.js';
 import {
   exitPlanModeTool,
   createExitPlanModeHandler,
@@ -485,6 +486,14 @@ export class AnthropicDirectProvider implements ModelProvider {
       ...(opts?.env !== undefined ? { env: opts.env } : {}),
       sessionId: opts?.sessionId,
       parentSessionId: opts?.parentSessionId,
+      // Central output-cap backstop (#661), FORK-SCOPED. A forked child is the
+      // only session that carries a `parentSessionId`; the top-level session
+      // leaves it undefined. Enabling the cap exactly when parentSessionId is
+      // set means every FORKED child bounds each tool result at MODEL_CAP_BYTES
+      // (100KB) via headAndTail — containing the whole overflow crash class
+      // (MCP dumps, browser output, read_file of a huge file) for the actual
+      // victims — while the top-level session's behavior is UNCHANGED (no cap).
+      ...(opts?.parentSessionId !== undefined ? { maxOutputBytes: MODEL_CAP_BYTES } : {}),
       ...(opts?.traceWriter ? { traceWriter: opts.traceWriter } : {}),
       // Read-only-skill bash gate: forwarded from the provider's stored flag
       // (set by createChildProviderFactory / buildReadOnlyReconProvider) so a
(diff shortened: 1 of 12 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

**PR #663** — `fix(#661): bound web_scrape + tool output to prevent subagent context overflow`
Regime: full (321 lines, 7 files) · reviewed ref `fc95014` · CI green · stated intent: PR title + body (spec-compliance assessed).

Two-layer fix, both verified against the stated intent by reading the source at the reviewed ref: **Part A** tightens `web_scrape` byte bounds (default 1MB→100KB, ceiling 10MB→1MB) and swaps the local head-only truncation for the shared UTF-8-safe `headAndTail`; **Part C** adds an optional, fork-only `maxOutputBytes` backstop on `SessionToolDispatcher` that caps oversized tool results to head+tail. No blocking, critical, or high findings.

### Verified load-bearing claims (read directly at ref `fc95014`)
- **Hook/model ordering is correct.** `executeCore` (dispatcher.ts:990-992) awaits `executeCoreInner` — which fires PostToolUse (lines 1047/1073/1081/1109) on the full uncapped `content` — *before* calling `applyOutputCap`. A secret-scanning PostToolUse hook sees full content; only the model-facing view is capped. No truncation-evasion of hooks.
- **Fork-only wiring / default-OFF for top-level.** Both providers set `maxOutputBytes = MODEL_CAP_BYTES` exactly when `parentSessionId !== undefined` (anthropic-direct/index.ts:496, openai-compatible/index.ts:502) — true only for forked children. The constructor guard (dispatcher.ts:310-314) arms the cap only for a positive finite number, so top-level sessions (`undefined`) are byte-for-byte unchanged.
- **Idempotent + applied exactly once.** `applyOutputCap` no-ops when content is already within budget (dispatcher.ts:1012); `executeCoreInner` has a single caller (`executeCore`), and both `execute()` and `executeBatch()` route through it, so each result is capped once.
- **`result.image` preserved; UTF-8-safe at both cut points.** `applyOutputCap` touches only `.content`/`.truncated`; `headAndTail` walks off continuation bytes and reserves marker budget.
- **api-compat: no reachable break.** `web_scrape` is consumed only as a model-invoked tool; grep found zero production callers passing `max_bytes` (all such call sites are `*.test.*`). `max_bytes` stays caller-overridable and clamped within the new 1MB ceiling. `maxOutputBytes?` is an optional field — no dispatcher-constructor break.
- **Test coverage matches the PR's claims.** All 10 claimed tests exist at the ref (web-scrape.test.ts:291/318/331/345; dispatcher.test.ts idempotent/image/executeBatch/non-positive-NaN cases).

### Findings

**medium · perf-observability · dispatcher.ts:1009-1016 · ref:fc95014 · file-state** — Central truncation is silent: `applyOutputCap` sets `result.truncated = true` but emits no trace/warn, so a forked child silently losing the middle of a large tool result is invisible in logs.
Evidence:
```
result.content = headAndTail(result.content, cap);
result.truncated = true;
return result;
```
Suggestion: emit a trace event (original vs. capped byte count) when the cap actually fires, so fork-side truncation is observable. Non-blocking.

### What was not checked
- Wave 1 agents read against branch HEAD `fc950148fec013f2f5534568faabdaa541ae0592`; all blocking/high-candidate citations and the load-bearing ordering/wiring claims were independently re-read at that ref by the orchestrator.
- Did not execute the test suite (relied on the PR's stated CI-green run + citation checks); did not review the Telegram surface or runtime fork behavior end-to-end.
- Stated intent: PR #663 title + body — spec-compliance was assessed; no unmet intent and no scope creep found.
- No critical/high finding surfaced, so the adversarial shadow-verify pass was not triggered.

griffinwork40:
Addressed both review findings. New commits on `afk/bound-tool-output-661`:

- **`08b0141`** — fix(#661): gate fork output-cap on explicit signal, not leaky parentSessionId
- **`10ba6f6`** — fix(#661): emit observability signal when fork output-cap fires

### Codex P2 (fork-gating is leaky) — fixed
The cap was armed by both providers only when `parentSessionId !== undefined`. Verified that gate is leaky: the skill-fork path builds its nested executor with `createStubParentSession` (`sessionId: undefined`, `nesting.ts:66`) and never backfills the id, so every subagent a skill-forked child dispatches (grandchildren and deeper) had `parentSessionId: undefined` and was **not** capped.

Replaced the heuristic with an explicit `parentSessionId` → **explicit-fork-signal** change:
- Added `AgentConfig.subagentToolOutputCapBytes?: number` — a value-carrying, fork-only field (sibling of `subagentId`/`isNonInteractive`).
- `SubagentManager.forkSubagent` stamps it **unconditionally** as `MODEL_CAP_BYTES` on every fork. Verified `forkSubagent` is the single choke point for **all** fork classes — agent-tool (`subagent-executor.ts:588`), skill (`skill-executor/fork-dispatch.ts:333`), and compose (`compose-executor.ts:711` → `dag-subagent.ts:89`) all converge on it; the sole child `new AgentSession(...)` is at `subagent.ts:834`. The top-level session is built via `new AgentSession(...)` directly at the 7 entry points (bootstrap/chat/daemon/status/scheduler/query/telegram) and **never** through `forkSubagent`, so it never carries the field and stays uncapped.
- Both providers' `buildDispatcher` now arm `maxOutputBytes` from that signal (not `parentSessionId`) and thread it from the child config. The misleading comments in both `index.ts` files ("a forked child is the only session that carries a parentSessionId") are corrected to describe the new mechanism.

So skill-forked descendants (and all forks) are now capped, while the top-level session remains uncapped.

### afk review MEDIUM (silent truncation) — addressed
`applyOutputCap` set `truncated: true` but emitted no signal. When the cap **actually fires** it now logs the original-vs-capped byte delta via the file's existing lightweight logger (`debugLog`, gated on `AFK_DEBUG`/`DEBUG`). Chose `debugLog` over a witness trace event deliberately: no existing trace event kind carries an original+capped byte pair, and adding one would expand the closed zod trace schema (`types.ts` + `events.ts` union + tests) for a non-blocking diagnostic. (`tool_call.completed` already records the final capped `resultBytes` + `truncated`; this adds the original size the cap ate, which that event does not carry.) No schema change.

### Verification (pnpm, tsconfig strict)
- `pnpm lint` (`tsc --noEmit`): **pass, zero errors**.
- Affected tests (run individually), all green:
  - `subagent-output-cap.test.ts` — **3 passed** (fork stamps signal with stub parent + unconditionally + caller cannot override)
  - `providers/anthropic-direct/output-cap-wiring.test.ts` — **3 passed** (fork caps w/ parentSessionId undefined; top-level not capped; not capped when only parentSessionId set)
  - `providers/openai-compatible/query.test.ts` — **84 passed** (incl. 3 new parity tests)
  - `tools/dispatcher.test.ts` — **102 passed / 1 pre-existing skip** (backstop tests stay green)
  - `providers/anthropic-direct/plan-mode-gate-wiring.test.ts` — **3 passed**
  - `subagent.test.ts` (67), `subagent-phase-role.test.ts` (19), `dag-subagent.test.ts` (14), `subagent-executor.test.ts` (92), `skill-executor.test.ts` (90), `compose-executor.test.ts` (83), `subagent/child-config.test.ts` (38) — all green.
  - Combined run: **597 passed / 1 skipped**.

No new `process.env` reads, no new `@anthropic-ai/sdk` imports. Diff: 6 files + 2 new test files, +229/-16.

PR #610, opened 2026-07-14
fix(subagent): env-configurable timeout + raise default to 45m; observable forked children

griffinwork40:
# fix(subagent): env-configurable timeout + raise default to 45m; observable forked children

## Problem (verified diagnosis)

Forked sub-agents run under a single **blunt wall-clock cap** — `SUBAGENT_DEFAULT_TIMEOUT_MS`, previously **20 minutes**. That cap is time-based only: the existing tool-iteration cap bounds *rounds*, not *wall-clock*, and throttled rounds can each take minutes.

The 20-minute wall was **guillotining genuinely-working read-only children**, not relieving hung ones. `/review`'s `research-agent` (which nests `git-investigator`) does continuous grep/read/git archaeology that legitimately exceeds 20 minutes on a large diff. When the wall fired mid-work, the child was aborted and surfaced to the parent as an opaque failure. Key properties of the diagnosis:

- **Usage-independent** — the cap is a fixed timer, unaffected by token budget or model.
- **Cross-surface** — it bites the same way in the REPL, one-shot, telegram, and daemon paths.
- **Observability gap** — even when a child *did* hit its budget, the witness trace recorded a generic `failed`/`cancelled` with no signal that a *timeout* was the cause, and daemon-forked children emitted **no lifecycle events at all** (the daemon wired its executors with `traceWriter: undefined`). Reference trace: `fe69562a`.

## Design decision

Two options were considered:

- **Option A — progress-aware idle watchdog.** Reset the budget on observable child progress so an *actively-working* child is never cut off regardless of total wall-clock. Correct long-term answer, but larger surface area (needs a progress signal wired through the handle). **Deferred.**
- **Option B — env-configurable cap + raise the default (this PR).** Raise the foreground default `20m → 45m` (real headroom over the longest healthy child observed while still guaranteeing every fork terminates) and make it tunable per-environment via a new `AFK_SUBAGENT_TIMEOUT_MS`. Ship the interim relief now; make timeouts *legible* in the trace so Option A can be designed against real data.

This PR ships **Option B** and lays the observability groundwork (timeout classification in the trace) that Option A will build on.

[...]

PR #666, opened 2026-07-21
fix(trace): seal witness trace only from the top-level session

griffinwork40:
## What

A session tree (top-level session + child + nested grandchild subagents) shares **one** `TraceWriter` by reference. `writer.seal()` is one-shot/idempotent, and **every** `AgentSession.close()` — including a subagent's — sealed it unconditionally. So the **first descendant torn down** (often a nested grandchild) sealed the whole shared trace file.

Observed while diagnosing session `afb74e27`: a grandchild `git-investigator` finished and its `close()` sealed the trace `status: "succeeded"` at 20:22 — while the top-level `agent` dispatch was still mid-tool-use. That call hung and was operator-aborted at 20:56. Result: the witness trace **claimed a clean success ~34 min before the session actually ended (aborted)**, and every ancestor event after the seal hit the sealed writer and was silently swallowed (the "started-without-terminal" gap).

## Fix

Gate the seal on `this.config.parentSessionId === undefined` (top-level only) in `dispatchSessionEndOnce`, reusing the discriminant already used for ledger/actor/injectContext gating (`agent-session.ts:222/356`).

- Subagents still emit their own `closure` record — only the **seal** is gated.
- The process-exit backstop (`NdjsonTraceWriter.sealOnProcessExit`) still seals an orphaned top-level trace if `close()` never runs, so this cannot regress the unsealed-orphan class.

## Test

`src/agent/session/subagent-seal-ownership.test.ts`:
- top-level `close()` -> trace has `closure` **and** `session_sealed`
- subagent `close()` -> `closure` present, `session_sealed` **absent**

## Verification

- `pnpm lint` (tsc --noEmit, strict) clean
- Full suite: **12,589 passed / 0 failed**; existing `subagent-terminal-seal-race.test.ts` stays green

## Context / follow-ups (not in this PR)

This fell out of diagnosing that 44-min session. Two sibling defects are deliberately **separate** follow-ups:
- **Model mislabel** — `session_init_start` records the construction-time model alias, not the served provider (emit-side fix; needs a resume/launch-swap mechanism trace first).
- **Subagent hang** — the correct fix is the post-first-token idle watchdog already scoped as a TODO in `subagent.ts:103` (quick alternatives are unsafe: a hardcoded client timeout kills long streams; lowering the 45-min cap re-introduces the PR #610 regression).

### DIFF
--- a/src/agent/session/agent-session.ts
+++ b/src/agent/session/agent-session.ts
@@ -1179,17 +1179,19 @@ export class AgentSession implements IAgentSession {
       finalCostUsd: this.sessionRunningCostUsd,
       runningTokens: this.sessionRunningTokens,
     }).catch(() => {});
-    // Invariant: only the TOP-LEVEL session (no parent) may seal the writer.
+    // Invariant: only the TOP-LEVEL session (no fork marker) may seal the writer.
     // A session tree shares ONE TraceWriter by reference (forkSubagent threads
     // the parent's writer into every descendant), and seal() is one-shot /
     // idempotent (NdjsonTraceWriter). If a subagent sealed on its own close(),
     // the FIRST descendant torn down would seal the whole shared file while the
     // top-level is still mid-run — stranding every later ancestor event as the
     // "started without terminal" gap (a nested grandchild's end_turn sealing the
-    // root trace 30+ min before the real session ends). Subagents still emit
-    // their own `closure` record above; if the top-level never runs close(), the
-    // process-exit backstop (NdjsonTraceWriter.sealOnProcessExit) still seals.
-    if (this.config.parentSessionId === undefined) {
+    // root trace 30+ min before the real session ends). The seal owner gate uses
+    // forkSubagent's explicit fork marker, not parentSessionId: stub-parent skill
+    // forks can have no parent session id while still sharing the root writer.
+    // Subagents still emit their own `closure` record above; if the top-level
+    // never runs close(), the process-exit backstop still seals.
+    if (this.config.subagentToolOutputCapBytes === undefined) {
       await sealTraceWriter(this.config.traceWriter, {
         ...signals,
         finalTurnCount: this.turnCount,
--- a/src/agent/session/agent-session.ts
+++ b/src/agent/session/agent-session.ts
@@ -1164,7 +1164,8 @@ export class AgentSession implements IAgentSession {
     //      WHY the session ended (model_end_turn, abort, budget_exceeded,
     //      timeout). Carries the final cost / token tuple + last stopReason.
     //   2. Seal the trace via session_sealed — the sealed-clean terminal
-    //      record. Sealing happens BEFORE the SessionEnd hook fires so a
+    //      record — but ONLY for the top-level session (see the ownership
+    //      guard below). Sealing happens BEFORE the SessionEnd hook fires so a
     //      hook-thrown exception cannot leave the trace `sealed-crashed`
     //      on the normal close path (a hook crash would surface as a
     //      separate `hook_decision: block` record, not erase the seal).
@@ -1178,14 +1179,26 @@ export class AgentSession implements IAgentSession {
       finalCostUsd: this.sessionRunningCostUsd,
       runningTokens: this.sessionRunningTokens,
     }).catch(() => {});
-    await sealTraceWriter(this.config.traceWriter, {
-      ...signals,
-      finalTurnCount: this.turnCount,
-      finalCostUsd: this.sessionRunningCostUsd,
-      subagentCompletedCount: this.subagentCompletedCount,
-      subagentRunningTokens: this.subagentRunningTokens,
-      subagentRunningCostUsd: this.subagentRunningCostUsd,
-    }).catch(() => {});
+    // Invariant: only the TOP-LEVEL session (no parent) may seal the writer.
+    // A session tree shares ONE TraceWriter by reference (forkSubagent threads
+    // the parent's writer into every descendant), and seal() is one-shot /
+    // idempotent (NdjsonTraceWriter). If a subagent sealed on its own close(),
+    // the FIRST descendant torn down would seal the whole shared file while the
+    // top-level is still mid-run — stranding every later ancestor event as the
+    // "started without terminal" gap (a nested grandchild's end_turn sealing the
+    // root trace 30+ min before the real session ends). Subagents still emit
+    // their own `closure` record above; if the top-level never runs close(), the
+    // process-exit backstop (NdjsonTraceWriter.sealOnProcessExit) still seals.
+    if (this.config.parentSessionId === undefined) {
+      await sealTraceWriter(this.config.traceWriter, {
+        ...signals,
+        finalTurnCount: this.turnCount,
+        finalCostUsd: this.sessionRunningCostUsd,
+        subagentCompletedCount: this.subagentCompletedCount,
+        subagentRunningTokens: this.subagentRunningTokens,
+        subagentRunningCostUsd: this.subagentRunningCostUsd,
+      }).catch(() => {});
+    }
     await dispatchSessionEnd(
       this._hookRegistry,
       {
--- a/src/agent/session/subagent-seal-ownership.test.ts
+++ b/src/agent/session/subagent-seal-ownership.test.ts
@@ -10,8 +10,8 @@
  * thus reported a clean success for a session that hung and was cancelled, and
  * every subsequent ancestor event hit the sealed writer and was swallowed.
  *
- * Fix: `dispatchSessionEndOnce` gates `sealTraceWriter(...)` on
- * `this.config.parentSessionId === undefined` (top-level only). Subagents still
+ * Fix: `dispatchSessionEndOnce` gates `sealTraceWriter(...)` on the explicit
+ * fork marker (`subagentToolOutputCapBytes` absent ⇒ top-level). Subagents still
  * emit their own `closure` record; the process-exit backstop still seals an
  * orphaned top-level trace if close() never runs.
  *
@@ -75,6 +75,7 @@ describe('witness seal ownership', () => {
         provider,
         depth: 1,
         parentSessionId: 'parent-abc',
+        subagentToolOutputCapBytes: 100_000,
         traceWriter: writer,
       });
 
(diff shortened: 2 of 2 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

The fix is correct, safe, and well-tested for its stated scope. It resolves the reported bug (a descendant subagent sealing the shared witness trace mid-run) for the interactive `agent`-tool session tree — the surface where session `afb74e27` was observed — and is a strict net improvement: per the diff's removed lines, every subagent previously sealed unconditionally, so this only *reduces* mis-seals and cannot regress the class. Ships with one narrower residual gap worth a follow-up (below), but nothing blocking.

Reviewed at ref `8d8682d`. Regime: light (hotfix, ~113 lines, 2 files). Spec-compliance assessed against the PR title+body.

### Findings

**medium · correctness · `src/agent/session/agent-session.ts:1192`** — The seal guard `parentSessionId === undefined` is a leaky "top-level" discriminant. Skill-forked children and all daemon-dispatched subagents are built via `createStubParentSession` (`sessionId: undefined`, `src/agent/tools/nesting.ts:66`; used at `daemon.ts:93` and `skill-executor/fork-child-config.ts:147`), so their `parentSessionId` is `undefined` and they still pass this guard and seal the shared writer. The codebase already abandoned this exact discriminant for the #661 output-cap fix — see `src/agent/types/config-types.ts:563-575`: *"the explicit 'I am a forked subagent' signal that REPLACES the prior `parentSessionId !== undefined` heuristic … undefined for forks whose parent carries no sessionId."*
  - Suggestion: gate on the documented fork-only signal `this.config.subagentToolOutputCapBytes === undefined` (or `this.config.depth === undefined` — stamped on every child at `subagent/child-config.ts:276`, `undefined` on top-level), which forkSubagent sets regardless of parent sessionId. The interactive `agent`-tool tree is already fixed because those subagents inherit a *defined* parentSessionId; this closes the same invariant on the skill-fork/daemon surfaces.

**nit · correctness · `src/agent/session/agent-session.ts:1182-1191`** — The new invariant comment asserts "only the TOP-LEVEL session (no parent) may seal," but the guard actually admits any session whose stub parent lacked a sessionId (skill-forks, daemon subagents), overstating the guarantee.
  - Suggestion: reword to "only a session with no resolved parentSessionId seals" and cross-reference the `subagentToolOutputCapBytes` discriminant note at `config-types.ts:563-575`.

### Verified clean

- **Safety of the guard** — no legitimate top-level session carries a defined `parentSessionId`: `forkSubagent` is the sole child-construction path and top-level is always `new AgentSession(...)` directly (`config-types.ts:571-573`). The gate never wrongly suppresses a real top-level seal.
- **No unsealed-top-level regression** — `NdjsonTraceWriter.sealOnProcessExit` (`src/agent/trace/writer.ts`) still seals an orphaned top-level trace via the `process.on('exit')` backstop if `close()` never runs.
- **Single seal call site** — `sealTraceWriter` has exactly one production caller (the gated site); no other path seals or duplicates.
- **Closure still emitted** — `emitClosureEvent` sits outside the `if`; only the seal is gated, matching intent. The new test asserts exactly this (subagent → `closure` present, `session_sealed` absent).
- **Test exercises the guard** — the subagent case constructs `AgentSession({ depth: 1, parentSessionId: 'parent-abc', ... })`, which threads into `this.config.parentSessionId`, genuinely driving the false branch. Assertions read the on-disk `trace.jsonl` kinds — meaningful.
- **Security / api-compat / perf** — no issues: local NDJSON diagnostic artifact, no auth/secret/external-input path; no exported signature changed; the change removes a redundant seal write per subagent close (net fewer writes on fan-out).

### What was not checked

- Citations verified against branch HEAD `8d8682d` (agent-session.ts, writer.ts, subagent.ts, tools/nesting.ts, child-config.ts, fork-child-config.ts, config-types.ts, daemon.ts, and the new test). 
- Test suite / lint were **not** run in this review — the PR's "12,589 passed / lint clean" claim is unverified here.
- The medium finding's reachability rests on static tracing of the stub-parent → child-config path; it was not confirmed at runtime that a daemon/skill-fork subagent's `close()` reaches this seal on those surfaces.
- Stated intent was present (PR #666 title+body); spec-compliance assessed — all intent clauses have implementing changes, no scope creep.

griffinwork40 (commented):

## Merge decision: ✅ MERGE

`fix(trace): seal witness trace only from the top-level session` · reviewed at branch HEAD `0a6b828` · 2 files, +153/-16 (1 production hunk in `agent-session.ts` + a new 122-line regression test). Regime: light (hotfix). **No blocking findings.** Stated intent (PR #666 title+body) was assessed — every clause has an implementing, ref-verified change; no scope creep.

This revision directly resolves the `medium` finding from the prior sweep review (the leaky `parentSessionId === undefined` discriminant): the seal guard now gates on the explicit fork marker `this.config.subagentToolOutputCapBytes === undefined` (agent-session.ts:1194). The change does what it says — in a session tree that shares **one** `TraceWriter` by reference, only the top-level session now seals it, so a descendant subagent can no longer seal the shared file mid-run (the `afb74e27` failure: a grandchild's `close()` sealed `status: "succeeded"` ~34 min before the real, aborted end, stranding every later ancestor event as the "started-without-terminal" gap). Subagents still emit their own `closure` record; only the seal is gated.

### Blocking findings
None.

### Non-blocking findings

**LOW · correctness · `src/agent/session/agent-session.ts:1194` (vs `:179`, `:222`, `:356`, ref 0a6b828)** — The seal gate uses a **different discriminant** than every other top-level gate in the same file — deliberately, and correctly. Seal gates on `subagentToolOutputCapBytes === undefined`; the plan-exit-bridge (179), grants-audit (222), and SessionStart-inject (356) gates all use `parentSessionId === undefined`. The divergence is load-bearing (see verification below) and documented at lines 1189–1191, but it is a legibility hazard: a future edit "unifying" the file onto `parentSessionId` would silently reintroduce the stub-parent seal bug.
Evidence:
```
if (this.config.subagentToolOutputCapBytes === undefined) {
  await sealTraceWriter(this.config.traceWriter, { ...signals, ... }).catch(() => {});
}
```
Suggestion (follow-up, non-gating): extract a `private isTopLevel()` helper documenting *why* the seal discriminant differs from the parentSessionId gates, so the two are not accidentally merged.

**NIT · spec-compliance · PR body vs `agent-session.ts:1194` (ref 0a6b828)** — The PR body's "Fix" section still says the seal is gated on `parentSessionId === undefined`, but the committed code gates on `subagentToolOutputCapBytes === undefined` (the later "gate seal on fork marker" commit). Body text is stale relative to HEAD; no code impact.
Suggestion (follow-up, non-gating): update the body's gating sentence to name the fork marker.

- **The load-bearing discriminant is sound in BOTH directions** (independently re-derived at ref `0a6b828`, not taken from prose). *Direction 1 — no subagent slips through and seals:* `subagentToolOutputCapBytes` is stamped **unconditionally** on every fork at `src/agent/subagent.ts:728`, placed *after* the `...options.config` spread (line 699) with a caller override explicitly **not** honored (`subagent.ts:704–727`). `forkSubagent` (`subagent.ts:532`) is the sole choke point through which the agent-tool, skill/stub-parent, and compose/DAG paths all construct their child session, so no child can carry the field `undefined`. *Direction 2 — no real top-level is blocked:* every top-level factory calls `new AgentSession(...)` directly and none sets the cap (chat.ts, interactive/bootstrap.ts, cli/commands/daemon.ts, daemon/scheduler.ts, query.ts, telegram/construct-session.ts), so no genuine top-level seal is suppressed.
- **The switch off `parentSessionId` is justified and is exactly what the new 3rd test locks.** Stub-parent skill forks (`nesting.ts` `createStubParentSession`, `sessionId: undefined`) leave `parentSessionId` undefined while still sharing the root writer — under the *old* guard that case evaluated `true` and **sealed**. The new `subagent-seal-ownership.test.ts` case 3 sets `subagentToolOutputCapBytes: 100_000` with **no** `parentSessionId` key and asserts `session_sealed` is **absent** — so it fails under the old guard and passes under the new one. That is the regression lock.
- **Test-coverage exercises both branches.** Case 1 (top-level, cap omitted → guard true) asserts `closure` **and** `session_sealed`; cases 2–3 (cap = 100_000 → guard false) assert `closure` present, `session_sealed` absent. Assertions read the on-disk `trace.jsonl` kinds — meaningful, not structural.
- **api-compat clean.** No exported symbol or signature changed. `subagentToolOutputCapBytes?: number` pre-exists at `config-types.ts:577` (introduced by #661); the diff only *references* it in the guard and tests — it is not introduced or modified here.
- **security / perf clean.** Local NDJSON diagnostic artifact only — no auth/secret/injection/network surface. Net perf-neutral-to-positive: one boolean branch at close, and the change *removes* a redundant seal write per subagent teardown on fan-out. Subagent forensics preserved (`closure` still emitted for every subagent).

### What was NOT checked

- Read against branch HEAD `0a6b828` — the load-bearing claims (unconditional cap stamp at `subagent.ts:728`, the top-level factories not setting the cap, the pre-existing field at `config-types.ts:577`, the 3-case test branch coverage) were re-derived at that ref via `gh api .../contents?ref=` reads. No critical/high findings, so `/shadow-verify` and Wave 1.5 citation verification were not required.
- Tests were **not** executed; the PR's "pnpm lint clean / 12,589 passed / 0 failed" claim is unverified here, though CI on this head is all-green (Lint & Build, Test, real-PTY scrollback, Docs, Publish Bundle Smoke).
- `MODEL_CAP_BYTES` (imported at `subagent.ts:44` from `./tools/handlers/_output-cap.js`) was taken as a defined numeric constant from its usage; the module was not opened to confirm it is never `undefined`. It backs only the "cap is always defined on forks" claim, which the in-code invariant comment also asserts.
- Runtime confirmation that a daemon/skill-fork subagent's `close()` actually reaches this seal branch on those surfaces was traced statically (construction path), not observed at runtime.

Net: a correct, spec-compliant, api-compatible hotfix that closes the seal-ownership invariant on the exact surface (stub-parent / skill-fork / daemon subagents) the prior `parentSessionId` gate missed, with a regression test that locks the fix. The two items are a non-blocking legibility note and a stale-body nit. Safe to merge.

PR #667, opened 2026-07-21
feat(compaction): token-aware keep-window so /compact works on short-but-full sessions

griffinwork40:
## Problem

`/compact` (and auto-compaction) refuse to do anything on a session with only 1–2 turns, **even when the context window is nearly full** — reporting *"not enough history to compact yet."* This is the exact complaint that motivated the change: a single heavy research turn (huge tool outputs, big file reads) fills the window, but `/compact` says the session is too short.

**Root cause:** compaction was gated purely on **turn count**, never token fullness.

- The compactor keeps the last `N = 2` *fresh user turns* verbatim (`DEFAULT_COMPACT_KEEP_LAST_TURNS`) and only summarizes turns **older** than them (`findCompactionBoundary`, `shared/compaction.ts`).
- Fewer than 2 fresh user turns → `history-too-short`; exactly 2 with nothing older → `nothing-to-summarize`.
- Manual `/compact` **never consulted token usage at all** (`cli/slash/commands/core.ts` only branches on the returned reason string).

So a full 1–2 turn session reclaims nothing regardless of how full it is. (Contrast: Claude Code anchors compaction on **tokens**, and its `/compact` runs regardless of turn count.)

## Fix

Add a token-fullness fallback, `findCompactionBoundaryAdaptive` (`shared/compaction.ts`):

> When the turn-count keep-window yields nothing to summarize **but** context usage is at/above a threshold (default **0.7** of the auto-compaction budget), progressively relax the keep-window toward 1 turn so an older-but-recent turn becomes eligible.

- **Safety invariant preserved:** every candidate boundary is still a `findCompactionBoundary` result — i.e. a *fresh-user-turn index* — so the kept tail never starts with an orphaned `tool_result` and `tool_use`/`tool_result` pairing stays intact. No mid-turn cuts.
- **Both providers:** `anthropic-direct` (bespoke `compact-handler.ts`) and `openai-compatible` (via `runCompactionCore`).
- **Both triggers:** manual `/compact` and auto-compaction (which already fires at ~0.9, so the shrink always applies once it triggers).
- **Fullness denominator:** `autoCompactLimitFor(requestedModel)` — the same working budget the auto trigger uses (honors the `*_1m` opt-in).
- **Backward compatible:** `usedFraction` defaults to `0` in the shared core (unknown usage → never shrink), so existing turn-count-only behavior is unchanged when fullness is unknown or below threshold.
- **New knob:** `AFK_COMPACT_SHRINK_FRACTION` (0–1 exclusive; default 0.7), registered in `env.ts` + `docs/env-registry.*`.

## Limitation / follow-up

A genuinely **single**-turn session still no-ops honestly — its one user turn *is* the kept tail, so there's nothing older to summarize. Reclaiming space inside a single fat turn needs **tool-result microcompaction** (trim old tool_result payloads in place, à la Claude Code's microcompact); that's a larger, separable change and is left as follow-up.

## Testing

- `findCompactionBoundaryAdaptive` unit tests — no-shrink below threshold, shrink on a 2-turn full session, single-turn stays a no-op, `keepLastN=1` has nothing to shrink (`shared/compaction.test.ts`).
- `contextFullnessFraction` units (`auto-compact.test.ts`).
- `runCompactionCore` adaptive cases — compacts a 2-turn session when `usedFraction >= threshold`, no-ops below (covers the OpenAI path).
- Anthropic integration test: a 2-turn session reporting 181k/200k now compacts (mirror of the existing empty-window `nothing-to-summarize` test).
- Updated one OpenAI auto-compaction test whose assertion encoded the old turn-count gate (now compacts at turn 2 instead of only turn 3 under a full window — the intended behavior).

Full suite green (`pnpm test`: 12,600 pass), `pnpm lint` clean, `scan:env` / `audit:env` / `audit:sdk` all pass. Changes were isolated on `origin/main` in a fresh worktree and re-verified (lint + 213 targeted tests + audits) independently of unrelated in-progress work.

---

🤖 Generated with agent-afk

### DIFF
--- a/docs/env-registry.json
+++ b/docs/env-registry.json
@@ -107,6 +107,14 @@
       "category": "model"
     },
     {
+      "name": "AFK_COMPACT_SHRINK_FRACTION",
+      "description": "Context-fullness fraction (0–1, exclusive) at/above which /compact and auto-compaction relax the keep-window so a short-but-full session (few turns, huge tool exchanges) can still be summarized instead of no-oping on turn count. Default 0.7 (see shared/compaction.ts DEFAULT_COMPACT_SHRINK_THRESHOLD).",
+      "type": "number",
+      "required": false,
+      "example": "0.8",
+      "category": "model"
+    },
+    {
       "name": "AFK_COMPANION_PRIMER",
       "description": "Opt-in: absolute path to a single companion-primer file. When set, its content is bounded (capped, fenced as <companion-primer>) and appended to the system prompt at session start for top-level sessions (chat/REPL/telegram/daemon), as lower-authority \"reflections, not facts\" context. Unset (default) = no-op. Only the one named file is ever read — never a directory or repo walk.",
       "type": "string",
--- a/docs/env-registry.md
+++ b/docs/env-registry.md
@@ -2,7 +2,7 @@
 
 Generated from `src/config/env.ts`. Do not edit by hand — run `pnpm scan:env` after changing the registry source.
 
-**130 vars** across 12 categories. Every `process.env[...]` read in `src/` outside `src/config/env.ts` is a CI failure (enforced by `pnpm audit:env:check`).
+**131 vars** across 12 categories. Every `process.env[...]` read in `src/` outside `src/config/env.ts` is a CI failure (enforced by `pnpm audit:env:check`).
 
 To add a var: edit `src/config/env.ts` (add a getter on `env` + an entry in `ENV_REGISTRY`), then run `pnpm scan:env`.
 
@@ -12,6 +12,7 @@ To add a var: edit `src/config/env.ts` (add a getter on `env` + an entry in `ENV
 |------|------|----------|---------|---------|-------------|
 | `AFK_COMPACT_KEEP_LAST_TURNS` | number |  |  | `6` | Number of recent turns the compactor keeps verbatim during /compact. Default tuned in compact-handler.ts. |
 | `AFK_COMPACT_MODEL` | string |  |  | `claude-haiku-4-5` | Override the model used by the /compact summarizer. Falls back to a cheap default (haiku-class). |
+| `AFK_COMPACT_SHRINK_FRACTION` | number |  |  | `0.8` | Context-fullness fraction (0–1, exclusive) at/above which /compact and auto-compaction relax the keep-window so a short-but-full session (few turns, huge tool exchanges) can still be summarized instead of no-oping on turn count. Default 0.7 (see shared/compaction.ts DEFAULT_COMPACT_SHRINK_THRESHOLD). |
 | `AFK_DEFAULT_SUBAGENT_MODEL` | string |  |  | `sonnet` | Override the default model used when a subagent is dispatched without an explicit model. |
 | `AFK_DISABLE_PROMPT_CACHE` | boolean |  | `0` | `1` | Disable Anthropic prompt caching when set to 1/true/yes/on. Unset = caching enabled. |
 | `AFK_EFFORT` | string |  |  | `medium` | Effort hint guiding adaptive-thinking depth, forwarded as Anthropic output_config.effort (model-gated; ignored where unsupported). Accepts low \| medium \| high \| xhigh \| max. |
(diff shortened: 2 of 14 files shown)


griffinwork40 (commented):

**Merge decision: ✅ MERGE** (2 non-blocking `low` findings + 1 cosmetic `nit`; no blocking issues)

Reviewed at ref `70a9953a1cb1da47d2d91a0987f83a5b0d4c065d`. Regime: full (399 changed lines, 14 files). Stated intent = PR title + body; spec-compliance assessed against it. CI: all green; branch: MERGEABLE.

### Summary by dimension
- **Security** — clean. Pure index/fraction arithmetic over the existing in-memory transcript; no new auth/network/persistence surface, nothing sensitive logged or persisted, and the adaptive path can only *shrink* the keep-window toward a fresh-user-turn boundary — it never feeds new payload to the summarizer. `readShrinkFraction()` reads `AFK_COMPACT_SHRINK_FRACTION` through the sanctioned `env` getter and clamps to `(0,1)`.
- **API-compat** — fully backward-compatible / additive-only. `findCompactionBoundaryAdaptive` is net-new; legacy `findCompactionBoundary` is retained (`shared/compaction.ts:153`); `runCompactionCore` gained only optional `usedFraction?`/`shrinkAtFraction?` with safe defaults (`?? 0` / `?? DEFAULT_COMPACT_SHRINK_THRESHOLD`), so `usedFraction ?? 0` preserves legacy turn-count-only behavior. The sole production caller passes both.
- **Correctness** — verified at the ref. The shrink loop `for (let n = keepLastN-1; n>=1; n--)` (`shared/compaction.ts:203`) terminates at `n>=1` and never calls `findCompactionBoundary(…, 0, …)` (which would return an out-of-bounds `messages.length`). The `boundary > 0` guard correctly lets a `0` base boundary (`nothing-to-summarize`) fall through to the shrink path while short-circuiting a working window, and the single-turn no-op is preserved (shrinking to `keepLastN=1` still lands index 0). Both providers wire `usedFraction` via `contextFullnessFraction(...)` / `autoCompactLimitFor(...)`.
- **Spec-compliance** — every stated-intent clause has an implementing change (adaptive fallback, progressive relax toward 1, fresh-user-turn safety invariant, both providers, both triggers, `autoCompactLimitFor` denominator, backward-compat default, registered env knob). No scope creep observed. The single-turn limitation is honestly documented as follow-up.

### Findings (non-blocking)

**1. low · test-coverage · `src/agent/providers/anthropic-direct/query/auto-compact.test.ts`**
No test exercises the anthropic-direct **auto** trigger (~0.9) driving the adaptive shrink — the added anthropic test covers only manual `compact()`. `git grep -nE "shrink|relax|adaptive"` at the ref returns zero matches in this file (confirmed-absent). The other three quadrants are covered: shared core both directions (`compaction.test.ts`), openai auto path (`query.test.ts`), anthropic manual (`anthropic-direct.test.ts`).
*Suggestion:* add an `auto-compact.test.ts` case with high `input_tokens` + a set `autoCompactThreshold`, asserting the adaptive splice fires from the turn-boundary path.

**2. low · perf-observability · `src/agent/providers/anthropic-direct/query/compact-handler.ts:225-240`**
When the adaptive shrink fires, no telemetry distinguishes it from a normal turn-count boundary — the `emitCompaction` payload carries `keptTailCount` / `keepLastNConfig` / `tokensSavedEstimate` but no flag that the keep-window was relaxed, so an operator cannot tell the fallback engaged. (Loop cost itself is a non-issue: at most `keepLastN`≈2 O(n) walks, once per compaction event.)
*Suggestion:* add an `adaptiveShrink: boolean` (or `effectiveKeepLastN`) field to the compaction witness payload, set when the returned boundary came from the shrink loop.

**nit · docs — `docs/env-registry.md` (`AFK_COMPACT_SHRINK_FRACTION` row):** the Example column shows `0.8` while the row text states "Default 0.7" — cosmetic (example ≠ default is legitimate); code default `0.7` is authoritative. No code change required.

### What was not checked
- Wave 1 agents read against branch HEAD `70a9953` via `git show <ref>:<path>`; the two `low` absence claims were independently re-verified against that ref (both confirmed-absent). No critical/high findings, so shadow-verify was not triggered.
- Did not run the test suite locally (PR reports `pnpm test` 12,600 pass, lint clean, env/sdk audits pass — not independently re-run here).
- Did not review the Telegram surface or runtime behavior under a live full-context session.

PR #668, opened 2026-07-21
feat(repl): edit a previous message — /rewind + double-Esc conversation rewind

griffinwork40:
## What

Feature parity with Claude Code's **"press Esc twice to edit a previous message."** At an empty idle prompt, a **double-Esc** (or the new **`/rewind`** command) opens an arrow-key picker of your earlier prompts. Choosing one rewinds the conversation to just before that message and reloads its text into the input — **editable** — to resend.

```
Esc Esc  (or /rewind)
  -> pick an earlier prompt   > second question
  -> conversation rewinds to that point
  -> its text loads into the prompt, editable -> Enter to resend
```

## Scope (and honest non-goals)

- **Conversation rewind only.** Claude Code couples a file-checkpoint "restore code" into the same menu; agent-afk has no working file-snapshot system (`rewindFiles` is stubbed `canRewind:false`), so that half is intentionally out of scope.
- **Provider:** fully implemented on **anthropic-direct** (the default). **openai-compatible** reports *not-supported* via the optional-method pattern (mirrors how `rewindFiles` already degrades). Its history is built per-turn from a different internal shape — parity is a clean follow-up.
- No "summarize from/up-to here" menu (Claude Code has it; `compact()` can back it later).

## Mechanism

An **`isIdle`-guarded in-place splice** of the provider's message array — the *same interlock `compact()` already uses*. Rewind only fires at the idle prompt and bails if a turn is in flight, so there's no race with an in-flight turn or auto-compaction, and no session-rebuild/banner cost (it feels seamless, not like `/clear`). `repairOrphanToolUses` runs on the new tail so a cut never strands a `tool_use` without its `tool_result` (Anthropic API contract).

> Design note: an earlier draft used a full session rebuild seeded with truncated history; rejected because `reset()` deliberately strips `resumeHistory` (a load-bearing `/clear` invariant) and would reset cost/turn counters + refire SessionStart. The guarded in-place splice is lighter and preserves session continuity.

## Changes

- **`provider.ts`** — optional `ProviderQuery.listRewindTargets()` / `rewindConversation(turnIndex)` + `RewindTarget` / `ProviderRewindConversationResult` types.
- **`anthropic-direct/query/rewind-conversation.ts`** *(new)* — pure enumerate (newest-first, `tool_result` turns excluded) + idle-guarded truncate helpers.
- **router / `IAgentSession` / `AgentSession`** — delegation with idle/closed guards and a `not-supported` fallback.
- **`/rewind` slash command** — `renderSelector` picker (compositor input suspended around it) -> `rewindConversation` -> `{ kind: 'prefill' }` reload result.
- **loop-iteration** — new `prefill` result threads `initialBuffer` into the next `readLine` (editable, **not** auto-submitted, unlike `seedBuffer`).
- **input-surface / terminal-compositor** — `initialBuffer` prefill on the compositor path + double-Esc detection at an empty idle prompt -> `onRewindRequest` -> resolves the pending `readLine` with `/rewind` (so the picker runs between turns, where selectors are safe).

## Testing

- **38 new tests**: pure rewind logic incl. the `tool_use`-boundary repair, guards, in-place identity; `/rewind` handler (picker / cancel / not-supported / prefill); double-Esc detection (arm / fire / draft-guard / streaming-unchanged); router delegation.
- Full suite green (**683 files, 12,615 passed**); `tsc --noEmit`, `audit:sdk:check`, `audit:env:check`, `scan:env:check` all pass.

## Try it

`afk i` -> type a couple of messages -> at the empty prompt press **Esc Esc** (or `/rewind`) -> pick a prior message -> edit -> Enter.

### DIFF
--- a/src/agent/provider.ts
+++ b/src/agent/provider.ts
@@ -321,6 +321,26 @@ export interface ProviderQuery extends AsyncIterable<ProviderEvent> {
    */
   compact?(): Promise<ProviderCompactResult>;
   /**
+   * Optional. Enumerate the genuine user-text turns in the current
+   * conversation, newest-first, for the "rewind to a previous message"
+   * picker. Each entry carries an opaque `turnIndex` handle the provider
+   * maps back to its internal history, plus a short single-line `preview`.
+   * Pure read — no mutation. Providers whose backend manages history
+   * opaquely leave this undefined; the harness returns an empty list.
+   */
+  listRewindTargets?(): RewindTarget[];
+  /**
+   * Optional. Rewind the conversation to a `turnIndex` returned by
+   * {@link listRewindTargets}: discard that user turn and everything after
+   * it, repairing any orphaned tool_use at the new tail so the Messages API
+   * contract holds. Returns the removed message's text as `reloadText` so
+   * the surface can reload it into the input for editing. In-place history
+   * mutation guarded by the same idle interlock as {@link compact} — a no-op
+   * (`rewound: false`) mid-turn. Providers that manage history opaquely leave
+   * this undefined; the harness surfaces a `not-supported` result.
+   */
+  rewindConversation?(turnIndex: number): Promise<ProviderRewindConversationResult>;
+  /**
    * Optional. Update the working directory used by the system prompt and
    * tool handlers for all **subsequent** turns in this query's lifetime.
    *
@@ -417,6 +437,40 @@ export interface ProviderCompactResult {
   messagesAfter: number;
   /** Best-effort estimate of input tokens saved on subsequent turns. */
   tokensSavedEstimate?: number;
+}
+
+/**
+ * One rewindable user turn, surfaced to the "edit a previous message" picker
+ * by {@link ProviderQuery.listRewindTargets}.
+ */
+export interface RewindTarget {
+  /**
+   * Opaque handle the provider maps back to its internal history position.
+   * For {@link ProviderQuery.rewindConversation}. Callers must not assume it
+   * is a contiguous 0..N index — treat it as a token from `listRewindTargets`.
+   */
+  turnIndex: number;
+  /** Short, single-line preview of the user message text (already truncated). */
+  preview: string;
+}
+
+/**
+ * Result of a {@link ProviderQuery.rewindConversation} call.
+ *
+ * `rewound: false` is not necessarily an error — it signals a deliberate no-op
+ * (session busy, turn in flight, invalid target, or provider doesn't support
+ * it). `reason` carries a short identifier the surface can render.
+ */
+export interface ProviderRewindConversationResult {
+  rewound: boolean;
+  reason?: string;
+  /**
+   * Text of the discarded user message, for reloading into the input box so
+   * the user can edit and resend. Present only when `rewound: true`.
+   */
+  reloadText?: string;
+  messagesBefore: number;
+  messagesAfter: number;
 }
 
 /**
--- a/src/agent/providers/anthropic-direct/query.ts
+++ b/src/agent/providers/anthropic-direct/query.ts
@@ -42,9 +42,11 @@ import type {
   ProviderMcpServerStatus,
   ProviderModelInfo,
   ProviderQuery,
+  ProviderRewindConversationResult,
   ProviderRewindResult,
   ProviderSessionInfo,
   ProviderUserTurn,
+  RewindTarget,
 } from '../../provider.js';
 import { buildRequestHeaders } from './auth.js';
 import {
@@ -70,6 +72,7 @@ import { type SessionState, createSessionState } from './query/session-state.js'
 import { AbortCoordinator } from './query/abort-coordinator.js';
 import { RetryLayer } from './query/retry-layer.js';
 import { compactHistory } from './query/compact-handler.js';
+import { listUserTurns, rewindConversationHistory } from './query/rewind-conversation.js';
 import { contextWindowTokensUsed, shouldAutoCompact, buildContextUsageFields } from './query/auto-compact.js';
 import type { HookRegistry } from '../../hooks.js';
 import { HookBlockedError } from '../../../utils/errors.js';
@@ -822,6 +825,28 @@ export class AnthropicDirectQuery implements ProviderQuery {
       initSessionId: this.initSessionId,
       ...(this.traceWriter ? { traceWriter: this.traceWriter } : {}),
     });
+  }
+
+  /**
+   * Enumerate genuine user-text turns (newest-first) for the rewind picker.
+   * See `query/rewind-conversation.ts`.
+   */
+  listRewindTargets(): RewindTarget[] {
+    return listUserTurns(this.state.messages);
+  }
+
+  /**
+   * Discard a chosen user turn and everything after it, returning the removed
+   * message's text for reload-and-edit. In-place, idle-guarded (same interlock
+   * as {@link compact}). See `query/rewind-conversation.ts`.
+   */
+  async rewindConversation(
+    turnIndex: number,
+  ): Promise<ProviderRewindConversationResult> {
+    return rewindConversationHistory(
+      { state: this.state, abort: this.abort },
+      turnIndex,
+    );
   }
 
   close(): void {
--- a/src/agent/providers/anthropic-direct/query/rewind-conversation.test.ts
+++ b/src/agent/providers/anthropic-direct/query/rewind-conversation.test.ts
@@ -1,0 +1,183 @@
+/**
+ * Unit tests for the conversation-rewind helpers — the provider half of the
+ * REPL "press Esc-Esc to edit a previous message" feature.
+ *
+ * Covers turn enumeration (newest-first, tool_result turns excluded), the
+ * guard paths (closed / turn-in-flight / invalid target), in-place truncation
+ * with array-identity preservation, reloadText extraction, and the
+ * tool_use/tool_result boundary repair after a splice.
+ */
+
+import { describe, it, expect } from 'vitest';
+import type { ContentBlockParam, MessageParam } from '@anthropic-ai/sdk/resources';
+import { listUserTurns, rewindConversationHistory } from './rewind-conversation.js';
+import type { SessionState } from './session-state.js';
+import type { AbortCoordinator } from './abort-coordinator.js';
+
+function makeState(messages: MessageParam[], closed = false): SessionState {
+  // The handler only reads `messages` + `closed`; the rest of SessionState is
+  // irrelevant here. Structural stub (tests are not type-checked by tsc).
+  return { messages, closed } as unknown as SessionState;
+}
+
+const idleAbort = { isIdle: () => true } as unknown as AbortCoordinator;
+const busyAbort = { isIdle: () => false } as unknown as AbortCoordinator;
+
+const toolUse = (id: string): ContentBlockParam[] => [
+  { type: 'tool_use', id, name: 'x', input: {} },
+];
+const toolResult = (id: string): ContentBlockParam[] => [
+  { type: 'tool_result', tool_use_id: id, content: 'ok' },
+];
+
+/** A realistic 3-turn conversation with one tool round in the middle. */
+function sampleHistory(): MessageParam[] {
+  return [
+    { role: 'user', content: 'first question' },          // 0  genuine
+    { role: 'assistant', content: toolUse('t1') },         // 1
+    { role: 'user', content: toolResult('t1') },           // 2  tool_result (NOT genuine)
+    { role: 'assistant', content: 'first answer' },        // 3
+    { role: 'user', content: 'second question' },          // 4  genuine
+    { role: 'assistant', content: 'second answer' },       // 5
+    { role: 'user', content: 'third question' },           // 6  genuine
+    { role: 'assistant', content: 'third answer' },        // 7
+  ];
+}
+
+describe('listUserTurns', () => {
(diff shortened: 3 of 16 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

`feat(repl): edit a previous message — /rewind + double-Esc conversation rewind` · reviewed at branch HEAD `1da3712` · 16 files (~877/-5; 4 test files + 2 new production modules). Full-regime review (security · api-compat · correctness · spec-compliance · test-coverage · perf-observability). **No blocking findings.** Stated intent (PR #668 title+body) was assessed — every named requirement has an implementing, ref-verified change; no scope creep.

The change does what it says: at an empty idle prompt a double-Esc (or `/rewind`) opens a newest-first picker of prior user turns; choosing one performs an `isIdle`-guarded **in-place splice** of the provider message array (the same interlock `compact()` uses) and reloads the chosen text into the next `readLine` as an **editable, not auto-submitted** prefill. `repairOrphanToolUses` runs on the new tail so a cut never strands a `tool_use` without its `tool_result`. anthropic-direct is fully implemented; openai-compatible degrades to `not-supported` via the optional-method pattern (mirrors `rewindFiles`). All blocking/high-candidate citations were independently re-derived at ref `1da3712` (Wave 1.5 citation + absence-claim verification; no fabricated or diff-only citations, nothing dropped).

### Blocking findings
None.

### Non-blocking findings

**MEDIUM · api-compat · `src/agent/types/session-types.ts:308,317` (ref 1da3712)** — Two **non-optional** methods were added to the publicly-exported `IAgentSession` interface, a breaking contract change for any *external* npm consumer that structurally implements `IAgentSession` (its implementations fail to type-check on upgrade). `IAgentSession` is exported from the package root (`src/index.ts:30`) and ships as `dist/index.d.ts`. In-repo there is no break — the sole implementer `AgentSession` is updated in this PR, and all other in-repo references consume it as a *type* — so the blast radius is external only, under a `feat` label.
  Evidence (non-optional — no `?`):
  ```ts
  listRewindTargets(): RewindTarget[];
  rewindConversation(turnIndex: number): Promise<ProviderRewindConversationResult>;
  ```
  Suggestion: mark both optional (`listRewindTargets?()`, `rewindConversation?()`) to mirror the already-optional `ProviderQuery` counterparts (`src/agent/provider.ts:331,342`); callers already null-guard. If kept required, release as a minor with an explicit interface-addition note.

**MEDIUM · correctness · `src/cli/terminal-compositor.input-dispatch.ts:357-370` (ref 1da3712)** — The double-Esc arm token `lastIdleEscAt` is reset only inline within `handleEscape` (on fire at :363, on non-empty-buffer disarm at :369) and never on an input-mode transition. Because a lone idle Esc arms the 600ms window (`DOUBLE_ESC_WINDOW_MS = 600`, :42) and `setInputMode` (`terminal-compositor.ts:747` → `InputMode.setInputMode`) never clears the token, an idle→streaming→idle round-trip that completes inside 600ms (e.g. a fast synchronous slash command) can leave the arm live, so the user's *next* single Esc is misread as the second tap and opens the rewind picker unbidden. Narrow window (a real turn almost always exceeds 600ms, so the arm expires by wall-clock and re-arms cleanly), so the effect is a rare spurious picker open, not data loss.
  Evidence:
  ```ts
  if ( self.lastIdleEscAt !== 0 && now - self.lastIdleEscAt <= DOUBLE_ESC_WINDOW_MS ) {
    self.lastIdleEscAt = 0;
    if (self.onRewindRequest) self.onRewindRequest();
  ```
  Suggestion: clear `self.lastIdleEscAt = 0` inside `InputMode.setInputMode` on every mode transition so a stale arm cannot survive a turn/command boundary.

**LOW · correctness · `src/cli/input/input-surface.ts:642-649` (ref 1da3712)** — `requestRewind()` returns `false` (no-op) when `pendingReadResolve` is null, and the return is discarded at the wiring site, so a double-Esc landing in the microtask gap between one `readLine()` resolving and the next re-arming silently drops the rewind. Behavior is safe (not human-reachable given the single-threaded loop; non-TTY never constructs the compositor). `[low confidence — verify with runtime context]`
  Suggestion: none required — document the discarded return as intentional at the wiring site.

**LOW · test-coverage · four untested integration seams (ref 1da3712)** — the provider-layer core paths are well covered (tool_use-boundary repair, closed/in-flight/out-of-range/non-user guards, in-place identity, `/rewind` handler picker/cancel/not-supported/prefill, double-Esc arm/fire/draft-guard/streaming-unchanged, router delegation), but four seams have zero test references (each grep `confirmed-absent` at the ref):
  - compositor prefill path `prefillInput` — `git grep -niE "prefillInput" 1da3712 -- '*.test.ts'` → 0 matches.
  - the `requestRewind()` → resolve-pending-`readLine`-with-`/rewind` bridge — `git grep -niE "requestRewind" 1da3712 -- '*.test.ts'` → 0 matches (tests assert only that `onRewindRequest` fires).
  - loop-iteration prefill *consumption* (that a `{kind:'prefill'}` result threads into iteration N+1's `readLine` as `initialBuffer`) — `git grep -niE "prefill|initialBuffer" 1da3712 -- 'src/cli/commands/interactive/*.test.ts'` → 0 matches.
  - `AgentSession.rewindConversation` `session-busy` branch (distinct layer from the provider-level `turn-in-flight` guard, which *is* tested) — `git grep -niE "session-busy|rewindConversation" 1da3712 -- 'src/agent/session/*.test.ts'` → 0 matches.
  Suggestion: add one focused test per seam (compositor `readLine({initialBuffer})` seeds via `prefillInput` and awaits Enter; `requestRewind` resolves a pending `readLine` to `/rewind`; a prefill slash-result becomes the next `readLine`'s `initialBuffer`; a non-idle session returns `session-busy`).

**NIT · perf-observability · `src/agent/providers/anthropic-direct/query/rewind-conversation.ts:78-88` (ref 1da3712)** — `listUserTurns` calls `extractUserText(message.content)` twice per genuine turn (once in `isGenuineUserTurn`, once for the preview) on each picker open. Harmless at realistic transcript sizes and only on an interactive keypress (not a hot path), but trivially avoidable.
  Suggestion: compute `extractUserText(message.content)` once and reuse it for the non-empty check and the preview.

- **security clean.** User-controlled rewind previews pass through `sanitizeSchemaString` before terminal render (`selectors.ts:49`), stripping ANSI/CSI/OSC/C1 — no terminal-escape injection. The splice is integer/range/genuine-turn guarded and idle-interlocked (`rewind-conversation.ts:114,118,125`), so no unbounded/negative-index splice and no data loss under a turn-in-flight race; `repairOrphanToolUses` on the new tail preserves the Anthropic tool_use/tool_result contract. Double-Esc is idle+empty-buffer gated and cannot fire mid-stream. No injection/RCE/secret/data-loss/auth vector.
- **api-compat (rest) clean.** `ProviderQuery.listRewindTargets?/rewindConversation?` are optional (openai-compatible omitting them is compatible; router + anthropic-direct implement/delegate). New types `RewindTarget` / `ProviderRewindConversationResult`, the `SlashResult` `{kind:'prefill'}` variant, and `InputSurfaceReadOpts.initialBuffer?` are all additive — no existing exported symbol was removed or had its signature changed. The only break is the non-optional `IAgentSession` addition (above).
- **spec-compliance satisfied.** All seven intent clauses have an implementing, ref-verified change: conversation-rewind-only (files stubbed, no `rewindFiles` behavior change); anthropic-direct full impl + openai-compatible `not-supported` via absence; `isIdle`-guarded in-place splice matching the `compact()` interlock; `repairOrphanToolUses` on the new tail; enumerate newest-first with `tool_result` turns excluded; prefill threaded as editable-not-auto-submitted `initialBuffer` (distinct from the `seedBuffer` auto-submit branch); double-Esc gated to empty idle prompt only. No scope creep.
- **correctness (rest) sound.** No off-by-one in enumerate/truncate; the in-place `splice(turnIndex)` preserves array identity; idle/closed guards and the `not-supported` fallback are correct; streaming isolation is enforced by the `inputMode === 'idle'` gate.

### What was NOT checked
- Read against branch HEAD `1da3712` — every blocking/high-candidate citation verified at that ref via `git show`/`git grep` (the two mediums and all four absence claims re-verified in Wave 1.5; no fabricated/diff-only citations). No critical/high findings, so `/shadow-verify` was not required.
- The api-compat break's blast radius is **external** package consumers implementing `IAgentSession` structurally — not grep-able in-repo; the in-repo implementer (`AgentSession`) is updated, so there is no in-repo break.
- Tests were **not** executed and lint was not run; the PR reports its gates green (683 files / 12,615 passed; `tsc --noEmit`, `audit:sdk:check`, `audit:env:check`, `scan:env:check` pass) and CI on this head is all-green (Lint & Build, Test, real-PTY scrollback, Docs, Publish Bundle Smoke).
- A rewind applies a history truncation but (unlike `compact()`, which emits a witness `compaction` trace event) emits no trace/telemetry record — noted as an observability gap against the precedent the design explicitly mirrors, not a diff-line defect.
- Real-terminal visual behavior of the picker / prefill was not exercised (subjective/interactive).

Net: correct, spec-compliant, and safe to merge. The two MEDIUM items are worth a follow-up — mark the `IAgentSession` methods optional to avoid an unlabeled external-consumer break, and clear `lastIdleEscAt` on mode transitions to close the rare spurious-picker window — but neither gates this PR.

griffinwork40 (commented):

## Merge decision: ✅ MERGE (with two non-blocking follow-ups)

Reviewed at branch HEAD `1da3712`. Read-only review across security, api-compat, correctness, spec-compliance, test-coverage, and perf/observability. All citations were verified against files read at that ref; the one absence claim below was confirmed by grepping the test tree. No critical or high-severity findings survived verification.

### Why MERGE
- **API surface is purely additive.** `ProviderQuery.listRewindTargets?()` / `rewindConversation?()` are optional (`src/agent/provider.ts:331,342`), so existing providers still satisfy the interface. `openai-compatible/query.ts` implements neither (verified: zero references) and correctly degrades via the router's `not-supported` fallback (`src/agent/providers/router/provider-router.ts:449-460`) — mirroring the existing `compact()`/`rewindFiles` optional-method pattern. The `SlashResult` `'prefill'` union member (`src/cli/slash/types.ts:292`) is additive.
- **The Anthropic `tool_use`/`tool_result` contract is preserved.** After `state.messages.splice(turnIndex)`, `rewindConversationHistory` unconditionally runs the pre-existing `repairOrphanToolUses` on the new tail (`src/agent/providers/anthropic-direct/query/rewind-conversation.ts`), the same repair already invoked at turn-start. Directly tested (`rewind-conversation.test.ts` — "repairs an orphaned tool_use left at the new tail").
- **Guards are layered and precedented.** The in-place splice is gated by `abort.isIdle()` at the provider (same interlock `compact()` uses) plus a `currentState !== 'idle'` → `session-busy` gate at `AgentSession` (`agent-session.ts:1081-1105`). In-place mutation preserves the loop's array-identity invariant (asserted in tests). Guard paths (closed / turn-in-flight / invalid-target / not-a-user-turn) are covered.
- **Security: no new exploitable surface.** Reloaded text is the user's own prior message routed as inert data through the existing `applyEdit(InputCore.seed(...))` prefill path — no eval/exec sink; previews are truncated + sanitized before painting. Perf is a non-issue (single O(n) splice + tail scan, once per idle-initiated rewind).
- **Spec-compliant.** Conversation-only rewind with `rewindFiles` left stubbed, anthropic-direct full impl, openai-compatible not-supported — matches the stated scope and explicit non-goals. Editable prefill (not auto-submit) is correctly distinct from `seedBuffer` (`loop-iteration.ts:324-328`).

### Non-blocking findings (worth a follow-up, none block merge)
1. **medium · correctness — double-Esc "empty prompt" gate ignores the queued-message state.** `src/cli/terminal-compositor.input-dispatch.ts:357` arms/fires the rewind picker on `self.input.buffer.length === 0` alone, with no check on `self.pendingSubmissions.length` / `self.queued`. The codebase's own `InputSurface.bufferIsEmpty()` treats "empty" as `text.length === 0 && !buf.queued`. In the narrow window where the session is idle between turns but type-ahead messages haven't drained yet, a double-Esc would rewind history underneath still-queued turns. Reachability is narrow (queued payloads drain promptly on the idle transition) and the outcome is confusing rather than destructive — hence non-blocking — but tightening the gate to also require an empty submission queue would close it.
2. **low · observability — rewind emits no witness/telemetry event.** `rewindConversationHistory` discards N messages (including tool-call history) with no audit trail, unlike `compact-handler.ts`, which calls `emitCompaction(...)` after its splice. Consider an analogous emit for parity, since rewind is an equally destructive in-place history mutation.

### Minor / test-coverage notes
- No test sets up a queued-messages state before the Esc-Esc sequence (confirmed: the 5 `rewind-esc` tests cover fire/arm/non-empty-draft/re-arm/streaming only) — the coverage gap behind finding #1.
- `InputSurface.requestRewind()` and the `loop-iteration` prefill *consume* side lack dedicated unit tests (the producing side is covered in `core-rewind.test.ts`).
- `renderSelector` is passed a freshly-constructed `AbortController().signal` that is never wired to a cancel source (`src/cli/slash/commands/core.ts`); the picker still cancels via its own keypress handler, so this is a dead parameter, not a bug.

### What was not checked
- The test-suite / `tsc` / audit numbers in the PR description were taken on faith (read-only review; suite not re-run).
- Runtime/manual TTY behavior of the double-Esc timing window was not exercised — static review only.
- Out-of-repo (published-package) consumers implementing `IAgentSession` directly could not be checked from this checkout; the sole in-repo implementer (`AgentSession`) is updated in this PR.

Issue #380, opened 2026-07-02
bug: worktree-sweep 'empty' verdict ignores ownerLiveness — live-session guard doesn't protect the empty path (reaps in-use worktrees)


griffinwork40:
## Bug

`worktree-sweep.ts`'s `empty` verdict never consults `ownerLiveness`, so the live-session presence guard — which was added specifically to stop the sweep reaping an in-use worktree — is a **no-op against the `empty` path**. A worktree that a live top-level session is actively working in gets reaped anyway, as long as its tree is clean, it has zero commits ahead of base, and it is older than 1 hour.

This is the "don't reap a live session" gap that #371 flagged as an explicit out-of-scope follow-up. This issue is that follow-up, now verified against code rather than inferred.

## Root cause (verified by direct read, `main` @ `891985a`, v5.15.7)

There **is** a live-session guard. When a live top-level session's presence `cwd` is inside a worktree, the sweep forces `ownerLiveness = 'alive'` (`worktree-sweep.ts:578-583`):

```ts
// A live session working inside this worktree overrides a dead creator
// pid — never reap an actively-used worktree.
if (
  ownerLiveness !== 'alive' &&
  liveSessionCwds.some((cwd) => isPathWithin(cwd, entry.path))
) {
  ownerLiveness = 'alive';
}
```

But that flag only gates the `dead-owner` verdict. The classifier (`worktree-sweep.ts:257-299`) checks the two clean-reap verdicts in this order:

```ts
// dead-owner — DOES check ownerLiveness (guard works here)
if (candidate.ownerLiveness === 'dead' && !candidate.isDirty && candidate.commitsAhead === 0) {
  return 'dead-owner';
}

// empty — does NOT check ownerLiveness at all (guard is bypassed)
if (candidate.commitsAhead === 0 && !candidate.isDirty && candidate.ageMs >= MIN_EMPTY_AGE_MS) {
  return 'empty';
}
```

So forcing `ownerLiveness = 'alive'` protects a worktree from `dead-owner` but does nothing for `empty` — the `empty` condition is satisfied regardless of liveness. The removal site (`worktree-sweep.ts:601-606`) then runs `git worktree remove --force` on it.

`MIN_EMPTY_AGE_MS = 3_600_000` (1 hour) (`worktree-sweep.ts:154`).

## Who it bites

[...]

PR #670, opened 2026-07-22
fix(worktree-sweep): count commits vs main HEAD when meta has no baseSha

griffinwork40:
## What

When a worktree's `.afk-worktree-meta.json` has **no `baseSha`**, the sweep engine computed commits-ahead as:

```ts
const baseSha = meta?.baseSha ?? entry.head;
rev-list `${baseSha}..${entry.head}` --count   // → HEAD..HEAD → always 0
```

Self-comparing `HEAD..HEAD` is **always 0**, so a baseSha-less worktree that actually holds unmerged commits is scored `commitsAhead === 0` → classified `empty` (or `dead-owner`) → its **checkout is force-removed**. The branch ref survives, but the checkout (and any gitignored setup like `node_modules`) is destroyed — i.e. an in-use worktree gets reaped.

Meta is absent/minimal precisely for **hand-created `git worktree add` trees** later adopted by `touchWorktreeOccupancy`, which writes `{ owner: 'agent' }` with no `baseSha`. That's the exact shape the documented crash-recovery workflow produces when resuming a swept branch.

## Fix

When `meta.baseSha` is undefined, fall back to the **main worktree's HEAD** and count commits reachable from the candidate but not from main:

```ts
rev-list <entry.head> --not <mainHead> --count   // true unmerged-work signal
```

- Worktrees that **do** record a `baseSha` are unchanged (same `baseSha..head` count).
- A baseSha-less tree that sits **at main's HEAD** is still `0`-ahead and remains reapable — no over-preservation.

Net effect: the sweep stops misclassifying baseSha-less worktrees that hold real work as `empty`. Strictly more conservative (only ever *preserves* more); no worktree that was correctly reaped before is preserved now.

## Tests

Two cases added to the `base-sha-fallback` suite in `worktree-sweep.test.ts`:
1. baseSha-less tree with unmerged commits → **not** `empty`/`dead-owner`, **not** removed (fails without the fix — the self-compare path returns 0).
2. baseSha-less tree at main HEAD → still `empty`, still removed (guards against over-preserving).

## Gates
- `pnpm test src/agent/worktree-sweep.test.ts` → 33/33
- `pnpm lint` (tsc --noEmit strict) → clean
- adjacent suites (boot-prune, scheduler, worktree-managed, worktree) → 78/78

## Context / scope (why this specific change)

This is the one *pure-win* fix that came out of a root-cause investigation into "active worktrees keep getting reaped." The broader findings, for reviewer context:

- The sweep is otherwise **well-targeted**: it only reaps worktrees with **no commits and no dirty files**. Committed work (`stale-clean`) and uncommitted work (`stale-dirty`) are always preserved. This baseSha bug was the sole path by which committed work could be misread as empty.
- **Deliberately NOT changed:** the `dead-owner` verdict's lack of an age gate. That is intentional (#380, and asserted by an existing test) — it fast-reaps genuinely-empty ghost worktrees, which is low-harm (nothing committed/dirty is lost). Gating it would revert that decision, so it's left alone.
- The historical mid-session reaping of top-level `afk -w` worktrees was closed earlier by presence-cwd tracking + the `empty` liveness gate; not touched here.

### Out of scope (flagged, not fixed here)
- **Worktree sprawl** (e.g. 27 worktrees / 6.1 GB in one checkout) is *preserved committed-but-unmerged work* the auto-sweep correctly won't touch — it needs explicit `worktree` prune/remove, not the auto-sweep.
- The **daemon cron sweep is unconfigured** on at least one machine (runs from a non-repo cwd; `AFK_WORKTREE_SWEEP_ROOT` unset), so it has never actually run there. That's an ops/config item, not a code change.

### DIFF
--- a/src/agent/worktree-sweep.test.ts
+++ b/src/agent/worktree-sweep.test.ts
@@ -611,6 +611,110 @@ describe('base-sha-fallback', () => {
     const verdict = result.candidates[0]?.verdict;
     expect(['empty', 'stale-clean', 'stale-dirty', 'active', 'locked']).toContain(verdict);
   });
+
+  it('preserves a baseSha-less worktree that holds unmerged commits (counts against main HEAD, not self)', async () => {
+    // Regression guard: a meta with no `baseSha` (exactly what
+    // touchWorktreeOccupancy writes when it adopts a hand-created
+    // `git worktree add` tree) must NOT self-compare its HEAD to itself —
+    // that always yields 0 commits ahead and makes a tree holding real
+    // unmerged work look `empty`, force-removing the checkout.
+    const worktreePath = join(afkWorktreesDir, 'afk-nobase-ahead');
+    await fs.mkdir(worktreePath, { recursive: true });
+    await fs.writeFile(
+      join(worktreePath, '.afk-worktree-meta.json'),
+      JSON.stringify({
+        owner: 'agent',
+        createdAt: new Date(Date.now() - 2 * 3_600_000).toISOString(), // 2h old
+      }),
+    );
+
+    const mainBlock = worktreeBlock({ path: repoRoot, head: 'mainsha111' });
+    const wtBlock = worktreeBlock({
+      path: worktreePath,
+      head: 'wtsha222', // differs from main HEAD
+      branch: 'refs/heads/afk/nobase-ahead',
+    });
+    const porcelainOut = `${mainBlock}\n\n${wtBlock}\n`;
+
+    const mock = makeMock(async ({ args }) => {
+      if (args.includes('list') && args.includes('--porcelain')) {
+        return { stdout: porcelainOut, stderr: '' };
+      }
+      if (args.includes('status') && args.includes('--porcelain')) {
+        return { stdout: '', stderr: '' }; // clean
+      }
+      if (args.includes('rev-list') && args.includes('--count')) {
+        // The fix counts commits reachable from the worktree HEAD but NOT from
+        // main (`--not`). The old self-comparing fallback (base === head) would
+        // instead resolve to 0 here and misclassify the tree as empty.
+        return args.includes('--not')
+          ? { stdout: '2\n', stderr: '' }
+          : { stdout: '0\n', stderr: '' };
+      }
+      return { stdout: '', stderr: '' };
+    });
+
+    const result = await runSweep({
+      execFile: mock as ExecFileFn,
+      repoRoot,
+      lockPath: lockFile,
+      dryRun: false,
+      telemetryPath: telemetryFile,
+    });
+
+    const verdict = result.candidates.find((c) => c.path === worktreePath)?.verdict;
+    expect(verdict).not.toBe('empty');
+    expect(verdict).not.toBe('dead-owner');
+    expect(result.removed).not.toContain(worktreePath);
+  });
+
+  it('still reaps a baseSha-less worktree with no commits (sitting at main HEAD)', async () => {
+    // Complement of the previous test: the fallback must not over-preserve.
+    // A baseSha-less tree that sits at main's HEAD is genuinely empty and
+    // remains reapable.
+    const worktreePath = join(afkWorktreesDir, 'afk-nobase-empty');
+    await fs.mkdir(worktreePath, { recursive: true });
+    await fs.writeFile(
+      join(worktreePath, '.afk-worktree-meta.json'),
+      JSON.stringify({
+        owner: 'agent',
+        createdAt: new Date(Date.now() - 2 * 3_600_000).toISOString(), // 2h old
+      }),
+    );
+
+    const mainBlock = worktreeBlock({ path: repoRoot, head: 'samesha000' });
+    const wtBlock = worktreeBlock({
+      path: worktreePath,
+      head: 'samesha000', // identical to main HEAD → nothing ahead
+      branch: 'refs/heads/afk/nobase-empty',
+    });
+    const porcelainOut = `${mainBlock}\n\n${wtBlock}\n`;
+
+    const mock = makeMock(async ({ args }) => {
+      if (args.includes('list') && args.includes('--porcelain')) {
+        return { stdout: porcelainOut, stderr: '' };
+      }
+      if (args.includes('status') && args.includes('--porcelain')) {
+        return { stdout: '', stderr: '' }; // clean
+      }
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

`fix(worktree-sweep): count commits vs main HEAD when meta has no baseSha` · reviewed at branch HEAD `e65d03b` · 2 files, +134/-3 (30 production lines in `worktree-sweep.ts` + a 104-line, 2-case test). Regime: light (hotfix). **No blocking findings.** Stated intent (PR #670 title+body) was assessed — every clause maps to an implementing, ref-verified change; no scope creep.

The change does what it says. When a worktree's `.afk-worktree-meta.json` records **no `baseSha`**, the old commits-ahead count fell back to `rev-list ${meta?.baseSha ?? entry.head}..${entry.head} --count` — i.e. `HEAD..HEAD`, which is **always 0**. A baseSha-less tree that actually holds unmerged commits was therefore scored `commitsAhead === 0`, routed to the `empty`/`dead-owner` verdicts, and had its **checkout force-removed** (branch ref survives; the working tree + gitignored setup like `node_modules` is destroyed). Meta is absent/minimal precisely for hand-created `git worktree add` trees adopted by `touchWorktreeOccupancy` (`{owner:'agent'}`, no baseSha) — the crash-recovery shape. The fix: when `meta.baseSha` is undefined, fall back to the **main worktree's HEAD** and count `rev-list <entry.head> --not <mainHead> --count` (the true unmerged-work signal); trees that record a baseSha take the identical `baseSha..head` path unchanged; a baseSha-less tree sitting at main's HEAD stays 0-ahead and remains reapable.

### Blocking findings
None.

### Non-blocking findings

**LOW · test-coverage · `src/agent/worktree-sweep.test.ts:650-652` (ref e65d03b, file-state)** — the new "preserves baseSha-less + unmerged" test's mock keys on `args.includes('--not')` (returns `2`, else `0`), which proves the `--not` branch is taken but does not assert the full `rev-list` argv shape/order.
Evidence:
```ts
return args.includes('--not')
  ? { stdout: '2\n', stderr: '' }
  : { stdout: '0\n', stderr: '' };
```
Suggestion: additionally assert the recorded call equals `['-C', repoRoot, 'rev-list', entry.head, '--not', mainHead, '--count']` so a future flag-order regression is caught.

**LOW · perf-observability · `src/agent/worktree-sweep.ts:610` (ref e65d03b, file-state)** — `catch { commitsAhead = 0; }` means a transient `git rev-list` failure scores the tree 0-ahead → reapable, against the "don't destroy work" contract. This **pre-dates** the change and is not worsened by it (same catch, same fallback), so it does not gate the PR.
Evidence:
```ts
} catch { commitsAhead = 0; }
```
Suggestion: log a warning on the rev-list catch so a silent reap-eligible fallback is auditable.

- **The fix targets the real reap path.** `classifyCandidate` at the ref gates `dead-owner` on `commitsAhead === 0` (`worktree-sweep.ts:284-289`) and `empty` on `commitsAhead === 0` (`:301-307`), while `stale-clean` — the preserve verdict — requires `commitsAhead > 0` (`:315-320`). A false `0` is exactly what routed a work-holding tree to reap instead of preserve; the count is load-bearing.
- **`mainHead` is the correct base.** `mainHead = parsed[0]?.head` (`:526`) pairs with `mainPath = parsed[0]?.path` (`:523`); `git worktree list --porcelain` lists the main worktree first, and the loop skips it via `entry.path === mainPath` (`:530`), so main is never classified and never self-compared.
- **Flag order / divergence-safe.** The new call is `['-C', repoRoot, 'rev-list', entry.head, '--not', mainHead, '--count']` (`:601`) — commits reachable from the worktree HEAD but **not** from main, which is correct even when the two have diverged (it counts only worktree-side commits).
- **"baseSha unchanged" holds.** When `meta.baseSha !== undefined` the path is the identical `${meta.baseSha}..${entry.head} --count` (`:598-599`); only the baseSha-less branch changed.
- **"strictly more conservative" holds under case enumeration.** baseSha present → unchanged; baseSha absent + `mainHead === entry.head` → `revListArgs` undefined → count stays 0 → reapable (old: `HEAD..HEAD` → 0 → reapable; same); baseSha absent + `mainHead` undefined (no `parsed[0]`) → stays 0 (old: 0; same); baseSha absent + real unmerged commits → new counts the true ahead-count and preserves (old: 0 → reaped). No input exists where the new code reaps a tree the old code preserved.
- **Tests exercise both new paths.** Test 1 (`:615-669`) uses distinct heads (`wtsha222` ≠ `mainsha111`) so the `--not` branch fires and asserts the tree is not `empty`/`dead-owner` and not removed — and fails without the fix (the old self-compare hits the `0` branch → `empty`). Test 2 (`:671-717`) uses identical heads (`samesha000`) so the `mainHead === entry.head` skip fires, and asserts the tree is still `empty` and removed (guards against over-preservation).
- **api-compat clean.** The exported surface (`runSweep`, `SweepOptions`, `SweepResult`, `ExecFileFn`) is outside the touched hunks; the change is internal logic only — no signature change.
- **security clean.** `entry.head`/`mainHead` are passed as argv elements to `execFile('git', [...])` (`:607`), not interpolated into a shell string — no injection surface (SHAs are git-validated refs regardless).

### What was NOT checked

- Read against branch HEAD `e65d03b` — all load-bearing citations (the classifier gates, the `mainHead` base, the `--not` flag order, the baseSha-unchanged path, and both tests) were re-derived at that ref via `git show`. No critical/high findings, so `/shadow-verify` was not required.
- Tests were **not** executed here; the PR reports `pnpm test src/agent/worktree-sweep.test.ts` 33/33, `pnpm lint` clean, and adjacent suites (boot-prune, scheduler, worktree-managed, worktree) 78/78 — and CI on this head is all-green (Lint & Build, Test, real-PTY scrollback, Docs, Publish Bundle Smoke).
- The out-of-scope items the PR itself flags (worktree sprawl needing explicit prune; the daemon cron sweep being unconfigured on some machines) are ops/config, not code, and were not assessed.
- The pre-existing `catch → commitsAhead = 0` reap-on-transient-git-failure behavior (LOW above) was noted but not exercised against a live git failure.

Net: a correct, spec-compliant, strictly-more-conservative hotfix on a load-bearing "don't destroy work" path, with both new branches test-covered and no public-surface or security impact. The two LOW items are optional follow-ups. Safe to merge.

PR #671, opened 2026-07-22
feat(compaction): tool-result microcompaction to reclaim context on short-but-full sessions

griffinwork40:
## Summary

Adds **tool-result microcompaction**: a deterministic (no-LLM) pass that reclaims context on **short-but-full sessions** — a single turn whose window is filled by huge `tool_result` payloads. Turn-granular summarization (the existing `/compact`) no-ops on these because there is only one fresh user turn (it is the kept tail; nothing older to summarize), yet the window is full. Microcompaction replaces large/old `tool_result` **content** with a short placeholder in place, reclaiming those bytes.

### Note on base (was meant to be stacked on #667)

This work was originally scoped to stack on **#667** (`afk/compact-short-full-session`), which added the `findCompactionBoundaryAdaptive` this builds on. **#667 has since been squash-merged into `main` and its branch deleted**, so a stacked PR against that branch is impossible. This branch is rebased directly onto `main`; every file it touches was byte-identical to #667's merge point, so it applies with zero conflicts and lands cleanly *after* #667.

## How it works

When `/compact` and auto-compaction would otherwise return `history-too-short` / `nothing-to-summarize`, a fallback runs `microcompactToolResults` and returns a new outcome **`microcompacted`** carrying `{ blocksCleared, bytesReclaimed }`. Purely additive — the normal summarization success path is untouched; the fallback fires only on those two no-op reasons.

Shared core lives in `src/agent/providers/shared/compaction.ts` (provider-neutral `microcompactToolResults` + `MicrocompactOps`/`ToolResultRef` abstraction); both providers are wired:
- **anthropic-direct**: `runMicrocompactFallback` in `query/compact-handler.ts` (both no-op branches).
- **openai-compatible**: fallback appended after `runCompactionCore` in `compact.ts`.
- Rendered in `/compact` (`cli/slash/commands/core.ts`) and Telegram (`telegram/formatter.ts` `formatMicrocompact` + `handlers/message.ts`).

## Invariants preserved (each an Anthropic/OpenAI Messages-API correctness property)

1. **PAIRING PRESERVED** — never removes a `tool_use` or `tool_result`; only swaps a result's content for a placeholder. Every `tool_use` keeps its matching `tool_result` (same id, same position). Verified by tests asserting tool_use/tool_result id sets are unchanged.
2. **NO LLM CALL** — mechanical array/content transform; reaches the fallback before any SDK/network use.
3. **WORKS AT ANY TURN COUNT** — selection does not depend on the fresh-user-turn boundary. Key test: a single-turn array where `findCompactionBoundary` returns `-1` still reclaims >35 KB.
4. **KEEP RECENT** — the most-recent K `tool_result` blocks are kept intact (default K=4); older/larger trimmed first (largest-first).
5. **IDEMPOTENT** — a second pass is a no-op (skips blocks already carrying the placeholder sentinel).
6. **PLACEHOLDER HINT** — `[tool result cleared to reclaim context — was N bytes]`.

## Config defaults + new env vars

Registered in `src/config/env.ts` ENV_REGISTRY (docs regenerated; `scan:env:check` passes):
- `AFK_MICROCOMPACT_TOOL_RESULT_BYTES` — byte threshold at/above which a `tool_result` is a candidate. Default **2048**.
- `AFK_MICROCOMPACT_KEEP_LAST` — most-recent `tool_result` blocks kept intact regardless of size. Default **4** (`<= 0` protects nothing).

Behavior with defaults is conservative: only results ≥2 KB and older than the freshest 4 are ever cleared.

## Local gate results

- `pnpm lint` (tsc --noEmit): **pass**
- `pnpm test src/agent/providers/shared/microcompact.test.ts`: **15 pass**
- `pnpm test src/agent/providers/shared/compaction.test.ts` (existing, unbroken): **21 pass**
- `pnpm test src/agent/providers/anthropic-direct/microcompact.test.ts`: **7 pass**
- `pnpm test .../query/compact-handler.microcompact.test.ts` (wiring): **3 pass** — proves `/compact` on a single-turn-full session reports `microcompacted`, not `history-too-short`
- Adjacent existing tests confirmed unbroken: `core-compact.test.ts` (10), `telegram/formatter.test.ts` (97)
- `pnpm audit:env:check`: **0 violations** · `pnpm scan:env:check`: **in sync (133 vars)** · `pnpm audit:sdk:check`: **OK**

CI runs the full suite. Only the touched/added test files were run locally to limit rate/token pressure.

## Risks / not done

- OpenAI `responses`-wire-mode returns `unsupported-wire-mode` **before** `compactOpenAIHistory`, so microcompaction does not run there — a distinct reason, not a `history-too-short` no-op. Left as-is (out of scope).
- Array-form `tool_result` content: byte accounting sums `text` blocks only; image/other non-text blocks are ignored (not the byte hogs). A cleared array result is rewritten to a single string placeholder.
- Auto-compaction default remains OFF; this only changes the no-op branches, so it is inert unless `/compact` is invoked or auto-compact is enabled.


griffinwork40 (commented):

## Merge decision: ✅ MERGE

`feat(compaction): tool-result microcompaction to reclaim context on short-but-full sessions` · reviewed at branch HEAD `bb9e030` · 14 files, +1122/-17 (4 new test files + docs env-registry regen; the 2 generated `docs/env-registry.*` files were skipped). Full-regime review (security · api-compat · correctness · spec-compliance · test-coverage · perf-observability). **No blocking findings.** Stated intent (PR #671 title+body, incl. its 6 named invariants) was assessed — every clause maps to an implementing, ref-verified change; no scope creep.

The change does what it says: a **deterministic, no-LLM** fallback (`microcompactToolResults` in provider-neutral `shared/compaction.ts`) that fires **only** on the two summarization no-op reasons (`history-too-short` / `nothing-to-summarize`) and reclaims context on a short-but-full session by swapping large/old `tool_result` **content** for a placeholder in place — returning a new additive outcome `microcompacted { blocksCleared, bytesReclaimed }`. Wired on both providers (anthropic-direct `runMicrocompactFallback` on both no-op branches; openai-compatible fallback after `runCompactionCore`) and rendered in `/compact` + Telegram. All blocking/high-candidate citations were independently re-derived at ref `bb9e030` (Wave 1.5 citation + absence verification; the two array short-circuits were cross-derived by two independent agents). No fabricated or false-absent findings survived.

### Blocking findings
None.

### Non-blocking findings

**MEDIUM · correctness/spec-compliance · `src/agent/providers/openai-compatible/compact.ts:144` + `src/agent/providers/anthropic-direct/compact.ts:167` (ref bb9e030, file-state)** — The array-form placeholder detector (`isToolMessagePlaceholder` / its Anthropic twin `isToolResultPlaceholder`) inspects only the **first** `text` part of an array-content `tool_result` and `return`s immediately, whereas the byte-accounting sums **all** text parts. A multi-text-block array whose first block is a genuine result but whose later block already carries the sentinel is misclassified as not-a-placeholder — a second pass would re-clear it and double-count bytes, so the claimed **IDEMPOTENT** invariant (#5) is fully enforced for string content but only *partially* for array content.
Evidence:
```ts
for (const part of content) {
  if (part.type === 'text') return isMicrocompactPlaceholder(part.text);
}
```
Reachability is narrow — `clear()` rewrites a cleared block to a single **string** placeholder, so a normally-cleared block never re-enters as multi-text-block array; this shape arises only from an externally-constructed history — hence MEDIUM, not high.
Suggestion: match the summing logic — `content.some(p => p.type === 'text' && isMicrocompactPlaceholder(p.text))`.

**MEDIUM · api-compat/correctness · `src/telegram/handlers/commands.ts:102-103` (ref bb9e030, file-state)** — The **idle-session** Telegram `/compact` path (`bot.ts:129` routes an idle session to `handleCompact` in `commands.ts`) branches only on `if (!result.compacted)` → `formatCompactNoop(result.reason ?? 'unknown')`, with no `microcompacted` case. Because `formatCompactNoop` has no special-case for that reason (`formatter.ts:469` fallthrough), a **successful** microcompaction that reclaimed bytes renders the misleading `📦 Nothing to compact (microcompacted).`. The sibling busy/enqueue→drain path **was** updated (`message.ts:596-599` branches on `reason === 'microcompacted'` → `formatMicrocompact`); the idle path was missed — a reachable user-facing inconsistency, not a type break or data corruption.
Evidence:
```ts
if (!result.compacted) {
  await ctx.reply(formatCompactNoop(result.reason ?? 'unknown'));
```
Suggestion: add a `result.reason === 'microcompacted' && result.microcompaction` branch calling `formatMicrocompact(result.microcompaction)` in `handleCompact`, mirroring `message.ts:596-599`.

**MEDIUM · test-coverage · `src/agent/providers/openai-compatible/compact.ts` (fallback path) (ref bb9e030, file-state)** — The OpenAI adapter's entire `microcompacted` fallback (`compactOpenAIHistory` — both no-op branches, `openaiMicrocompactOps` string/array handling, the `reason` routing) has **zero asserting tests**: `git grep -rn 'microcompact' bb9e030 -- 'src/agent/providers/openai-compatible/*.test.ts'` → 0 matches, and the directory has no microcompact test file — asymmetric with the Anthropic side, which ships a dedicated `query/compact-handler.microcompact.test.ts` plus a provider-level `microcompact.test.ts`. The shared algorithm is well tested, but the OpenAI wiring is a reachable production path (`compactOpenAIHistory` is called for chat-completions sessions) covered by nothing.
Evidence: `git grep -rln 'microcompact' bb9e030 -- src/agent/providers/openai-compatible/` → only the source `compact.ts`, no `*.test.ts`.
Suggestion: add an openai-compatible wiring test mirroring the Anthropic one — a `role:'tool'`-heavy single-turn history driven to `microcompacted`, asserting `blocksCleared`/`bytesReclaimed`.

**LOW · test-coverage · `src/agent/providers/openai-compatible/compact.ts:132-147` (ref bb9e030, file-state)** — No test exercises the OpenAI array-form (`OpenAIContentPart[]`) tool-message branch in `toolMessageContentBytes` / `isToolMessagePlaceholder`, so the multi-text-part path (and the MEDIUM short-circuit above) is uncovered for this provider; the Anthropic suite already has an array-content case.
Suggestion: add an OpenAI array-content microcompact case.

**LOW · perf-observability · `src/agent/providers/anthropic-direct/query/compact-handler.ts` (fallback body) (ref bb9e030, file-state)** — Microcompaction rewrites live history content in place but emits **no** witness/trace event, whereas the summarization success path emits one (`emitCompaction` via `onSuccess`). A context mutation that reclaimed bytes leaves no durable audit record beyond the transient UX line.
Suggestion: emit a lightweight trace event (`blocksCleared`/`bytesReclaimed`/`blocksScanned`) on the `blocksCleared > 0` branch, paralleling `emitCompaction`.

**LOW · correctness · `src/agent/providers/openai-compatible/compact.ts:281-283` (ref bb9e030, file-state)** — The fallback returns `messagesAfter: deps.priorTurns.length`, but microcompaction never changes array length, so `messagesAfter === messagesBefore` always; reading the live `.length` is correct but diverges from the honest-no-op branch that reuses `messagesBefore`. No failure mode — a consistency nit.
Suggestion: use `messagesAfter: messagesBefore` to match the documented `messagesBefore === messagesAfter` invariant.

**NIT · perf-observability · `src/agent/providers/shared/compaction.ts` (`MicrocompactResult.blocksScanned`) (ref bb9e030, file-state)** — `blocksScanned` is computed and documented but never surfaced: both provider handlers destructure only `{ blocksCleared, bytesReclaimed }`, so the scanned-vs-cleared telemetry is dead.
Suggestion: include `blocksScanned` in the `microcompaction` payload (or the trace event above).

- **The 6 invariants — 5 fully VERIFIED, 1 partial.** (1) PAIRING PRESERVED: `clear()` only reassigns `.content`; the block/message object, `tool_use_id`/`tool_call_id`, and array position are untouched (`anthropic-direct/compact.ts:194-197`; `openai-compatible/compact.ts:141-143`); no `splice`/`filter`/`delete` on any message array in the transform. (2) NO LLM CALL: in anthropic-direct both `boundary < 0` and `boundary === 0` `return runMicrocompactFallback(...)` **before** `abort.begin()` / `client.messages.create`; in openai-compatible the fallback runs only after `runCompactionCore` already returned a no-op reason. (3) WORKS AT ANY TURN COUNT: selection walks `listToolResults` with no reference to the fresh-user-turn boundary — the key test asserts `findCompactionBoundary === -1` yet `bytesReclaimed > 35 KB` on a single-turn array. (4) KEEP RECENT: `DEFAULT…KEEP_LAST = 4`, `protectedFrom = max(0, refs.length - keepLast)`, candidates `slice(0, protectedFrom)` sorted `b.byteLength - a.byteLength` (largest-first among older). (5) IDEMPOTENT: airtight for string content (sentinel `startsWith` skip) — **partial for array content** (the MEDIUM above). (6) PLACEHOLDER HINT: `[tool result cleared to reclaim context — was ${bytes} bytes]` matches the spec.
- **security clean.** The content-clearing is destructive **in-memory** and reachable unbidden via auto-compaction (`trigger:'auto'` → `compact()` → `runMicrocompactFallback`), but correctly gated: fires only on the two no-op reasons, `keepLast=4` protects the freshest results the agent is reasoning over, only content `>= threshold` is touched, and it is idempotent — a deliberate context-vs-fidelity tradeoff (equivalent to what summarization already does destructively), not a data-loss vuln. The placeholder sentinel carries only a **numeric** `bytes` length (no free text), so no injection surface. Env parsing (`resolveMicrocompactOptions`) is hardened: `Number.parseInt` + `Number.isFinite` + range guards (`>= 1` bytes / `>= 0` keepLast) + a `Math.max(0, …)` floor; the `'abc' / '0' / '-1' / '' / undefined` matrix is tested.
- **api-compat clean (additive).** `ProviderCompactResult.reason` is typed bare `reason?: string` (`provider.ts:422`) — **not** a discriminated union; `git grep 'switch.*reason'` and `'_exhaustive'` at the ref find **no** exhaustive consumer of a compaction reason, so adding `'microcompacted'` cannot break one. The new optional `microcompaction?` field (`provider.ts:433`) and every new export (`microcompactToolResults`, `MicrocompactOps`, `ToolResultRef`, `resolveMicrocompactOptions`, `isMicrocompactPlaceholder`, the sentinel, etc.) have **no pre-existing external importer** — the only consumers are the three files co-introduced by this PR. `compact()`'s signature is unchanged; its return type is widened by one optional field (backward-compatible for all four implementors).
- **spec-compliance satisfied, no scope creep.** Every intent clause has an implementing change; the disclosed out-of-scope items (OpenAI `responses`-wire-mode bail before compaction; array byte-accounting sums text only; auto-compact default OFF) are honored and not counted as gaps. `blocksScanned` is additive internal telemetry, not a behavior change.
- **perf clean (no hot path).** `git grep` confirms the only production call sites of `microcompactToolResults`/`runMicrocompactFallback` are inside `compactHistory` (`compact-handler.ts`) and `compactOpenAIHistory` (`compact.ts`) — reached only via explicit `/compact` or the opt-in auto-compact trigger. No render-loop, streaming, or per-message caller. The transform is O(n) over tool_result blocks with one sort of the candidate subset.

### What was NOT checked

- Read against branch HEAD `bb9e030` — all load-bearing citations were re-derived at that ref via `git show`/`git grep` (Wave 1.5), and the two array short-circuit lines (`compact.ts:144`, `compact.ts:167`) were independently confirmed by a second agent. No critical/high findings, so `/shadow-verify` was not required.
- The reviewing checkout's working tree predates this PR, so **all** post-merge file-state was read from the git object store at `bb9e030`, not from disk — the sanctioned path for citation verification here.
- Tests were **not** executed. The PR reports its gates green (`pnpm lint`; microcompact suites 15+7+3 pass; adjacent `compaction.test.ts` 21, `core-compact.test.ts` 10, `telegram/formatter.test.ts` 97 unbroken; `audit:env`/`scan:env`/`audit:sdk` clean) and CI on this head is all-green (Lint & Build, Test, real-PTY scrollback, Docs, Publish Bundle Smoke).
- Whether auto-compaction's cleared content is independently persisted to any session trace/store for recovery was **not** exhaustively traced — only the in-memory transform was assessed (irreversible in memory; any independent trace record would further reduce the already-nil data-loss concern).
- Real-terminal / real-Telegram rendering of the `/compact` and `formatMicrocompact` output was not exercised (interactive).

Net: a correct, spec-compliant, purely-additive, no-LLM context-reclamation feature on a well-gated path, with the shared algorithm and Anthropic wiring thoroughly tested and no security or api-compat break. The three MEDIUM items are genuine follow-ups — the multi-text-block array idempotency edge, the idle-session Telegram `/compact` message asymmetry, and the missing OpenAI-adapter fallback test — but none gates this PR. Safe to merge.

PR #674, opened 2026-07-23
feat(trace): observe-only suspected_loop telemetry for forked subagents

griffinwork40:
## What

Adds an **observe-only** `suspected_loop` trace signal. When a **forked** subagent issues the same tool call with the same normalized args ≥N times within a sliding window (N=5 within M=20 rounds), the dispatcher emits a `suspected_loop` session phase carrying `{ tool, count, windowSize }`.

**Observe-only — the load-bearing invariant:** `observeSuspectedLoop(call): void` never aborts the fork, never sets a `failureClass`, never alters or delays a tool result, never touches control flow. The trace write is fire-and-forget (`void emitSessionPhase(...)`). Scoped to forked children only (mirrors the denial circuit-breaker's `parentSessionId` guard), so interactive sessions are unaffected.

## Why

**Layer 1** of the "trapped subagent" follow-ups: pure data-gathering to measure whether genuine `(tool, args)` repetition loops actually occur in practice — **before** deciding whether an *enforcing* loop-detector is ever warranted. A devils-advocate wave rejected shipping an aborting detector outright: it's mis-tunable, risks false-aborting legit build→test→fix / paginated-search cycles, and wouldn't even have fired on the documented `/review` per-citation trap (distinct args each call). Telemetry-first lets us tune N/M against real traces instead of guessing.

Deliberately does **not** fire on distinct-arg fan-out — that's budget-shaped and handled by the `general-purpose` cap (#673). Complements the idle watchdog (silence) with a signal for activity-without-progress.

## Stacked on #672

Base branch is `afk/idle-watchdog` (PR #672) — this reuses the trace-phase machinery that PR added (`idle_watchdog_fired`), adding `suspected_loop` in lockstep to the `SessionPhaseName` union + Zod enum + the parity test. GitHub retargets this to `main` when #672 merges.

## Follow-up (not here)

No `afk trace show` rendering for the new phase yet (emission only) — a small reader-side addition once we've seen it land in real traces.

## Test / verification

- `pnpm lint` (tsc --noEmit, strict) — PASS
- `suspected-loop-detector.test.ts` (22), `dispatcher.test.ts` (103), `session-phase.test.ts` (47) — PASS
- Full `pnpm test` — **12778 passed / 14 skipped / 0 failed** (691 files)
- Observe-only invariant confirmed by grep: the detector module contains no abort/throw/failureClass/isError; `observeSuspectedLoop` returns `void` and is called as a standalone (non-gating) statement.

Files: new `suspected-loop-detector.ts` (+ test), `dispatcher.ts` (window state + fire-and-forget observe), `trace/types.ts` + `events.ts` (phase lockstep), `session-phase.test.ts` + `dispatcher.test.ts`.

### DIFF
--- a/src/agent/tools/dispatcher.test.ts
+++ b/src/agent/tools/dispatcher.test.ts
@@ -1439,6 +1439,50 @@ describe('SessionToolDispatcher — repeat-loop circuit breaker', () => {
   });
 });
 
+describe('SessionToolDispatcher — suspected-loop telemetry coexists with the repeat breaker (OBSERVE-ONLY)', () => {
+  // The observe-only suspected_loop signal shares the sequential pre-execution
+  // path with the repeat circuit breaker. These tests lock in that (a) it does
+  // NOT perturb the top-level session (which is out of its forked-only scope),
+  // and (b) it emits — without ever changing a result — for a forked child.
+
+  it('top-level session: no suspected_loop event and repeat-breaker behavior is unchanged', async () => {
+    const writer = new InMemoryTraceWriter();
+    // makeDispatcher() sets no parentSessionId → a top-level session.
+    const dispatcher = makeDispatcher({ traceWriter: writer });
+    for (let i = 0; i < REPEAT_CIRCUIT_BREAKER_THRESHOLD - 1; i++) {
+      expect((await dispatcher.execute(makeCall())).isError).toBeUndefined();
+    }
+    // The repeat breaker still trips at its threshold, exactly as before.
+    const tripped = await dispatcher.execute(makeCall());
+    expect(tripped.isError).toBe(true);
+    expect(tripped.circuitBreaker).toBe(true);
+    // And no observe-only loop signal is emitted for a top-level session.
+    const loopEvents = writer.events.filter(
+      (e) => e.kind === 'session_phase' && e.payload.phase === 'suspected_loop',
+    );
+    expect(loopEvents).toHaveLength(0);
+  });
+
+  it('forked child: emits suspected_loop but returns the normal tool result', async () => {
+    const writer = new InMemoryTraceWriter();
+    const dispatcher = makeDispatcher({ parentSessionId: 'parent-1', traceWriter: writer });
+    let last;
+    // Below the repeat-breaker threshold (8) so the calls actually run; the
+    // suspected-loop threshold (5) is crossed first, purely as telemetry.
+    for (let i = 0; i < 5; i++) {
+      last = await dispatcher.execute(makeCall());
+    }
+    // OBSERVE-ONLY: the result is the real handler output — no error/breaker flag.
+    expect(last?.isError).toBeUndefined();
+    expect(last?.content).toBe('hello');
+    expect(last?.circuitBreaker).toBeUndefined();
+    const loopEvents = writer.events.filter(
+      (e) => e.kind === 'session_phase' && e.payload.phase === 'suspected_loop',
+    );
+    expect(loopEvents).toHaveLength(1);
+  });
+});
+
 describe('SessionToolDispatcher — canUseTool (Dim 8 in-process permission policy)', () => {
   const allowAll: CanUseTool = async () => ({ behavior: 'allow' });
 
--- a/src/agent/tools/dispatcher.ts
+++ b/src/agent/tools/dispatcher.ts
@@ -28,7 +28,7 @@ import { classifyBashCommand } from './readonly-bash.js';
 import { headAndTail } from './handlers/_output-cap.js';
 import { PathGrantManager, type GrantSnapshot } from './grant-manager.js';
 import type { GrantManager } from '../../cli/slash/commands/allow-dir.js';
-import { emitHookDecision } from '../trace/emit.js';
+import { emitHookDecision, emitSessionPhase } from '../trace/emit.js';
 import type { TraceWriter } from '../trace/index.js';
 import { defaultConcurrencyClassifier, partitionIntoBatches } from './dispatch-batching.js';
 import { repeatCallFingerprint } from './repeat-circuit-breaker.js';
@@ -40,6 +40,13 @@ import {
   extractDeniedReadPath,
   buildDenialBreakerMessage,
 } from './denial-circuit-breaker.js';
+import {
+  createSuspectedLoopWindow,
+  fingerprintToolCall,
+  observeToolCall,
+  SUSPECTED_LOOP_WINDOW_SIZE,
+  type SuspectedLoopWindow,
+} from './suspected-loop-detector.js';
 
 // Re-exported for backward compatibility: external importers (dispatcher.test.ts,
 // schema-classification.test.ts) historically import this from './dispatcher.js'.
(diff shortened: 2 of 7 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

Reviewed at head ref `7080e80` (base `afk/idle-watchdog` — this PR is stacked on #672; GitHub retargets it to `main` when #672 merges). Full-regime review: 2 parallel dimension agents across security, api-compat, correctness, spec-compliance, test-coverage, perf-observability, plus a citation + absence-claim verification pass that independently re-derived every surviving claim from source at the ref. **No blocking, critical, high, or medium findings.** Two low/nit security notes below; neither blocks merge.

### Verified strengths (read against ref `7080e80`)

- **The load-bearing observe-only invariant holds — independently confirmed at source, not just from the PR description.** `observeSuspectedLoop` (`dispatcher.ts:618-637`) returns `void`; its sole outward effect is `void emitSessionPhase(...)`. It never sets `results[i]`, never adds to `blocked`, never returns/continues a `ToolResult`, never sets `failureClass`/`circuitBreaker`, and never throws into dispatch. Both call sites are bare post-repeat-breaker statements that change no control flow: `execute()` step 2d (`dispatcher.ts:823`, after `if (repeatBlock) return repeatBlock;`) and `executeBatch()` phase 1 (`dispatcher.ts:945`, after the repeat-breaker `continue`-block, with no `results[i]`/`blocked`/`continue` following it — the call falls through to phase-2 exactly as before). `execute()` delegates to `executeCore()`, not `executeBatch()`, so a single call is observed on exactly one path — no double-count.
- **Fire-and-forget emission cannot stall or throw into the tool path.** `emitSessionPhase` (`emit.ts:156-166`) is `async`, no-ops when no writer is attached, and wraps `writer.write` in try/catch (swallowing to `debugLog`), so the returned promise resolves and never rejects. Combined with the `void` at the call site, a slow or failing witness write neither delays nor perturbs dispatch.
- **correctness — clean.** Ring-buffer eviction is a correct FIFO cap (`recent.push` then `shift` when length exceeds `SUSPECTED_LOOP_WINDOW_SIZE = 20`, `detector.ts:191-194`); the debounce fires at most once per fingerprint per turn via the `firedFingerprints` latch (`detector.ts:197`); the threshold boundary is correct (fires on `count >= 5` within a 20-slot window, `WINDOW_SIZE >= THRESHOLD` so the predicate is reachable, no off-by-one); interleaved loops (A B A C A …) are detected where the consecutive-only repeat breaker cannot; `sortDeep` normalizes object-key order recursively while preserving array order.
- **spec-compliance — intent met in full, no scope creep.** Every named clause has an implementing change at the ref: N=5 / M=20 constants; forked-children-only guard (`parentSessionId` check, mirroring the denial breaker); `{ tool, count, windowSize }` metadata; at-most-once debounce; deliberately does NOT fire on distinct-arg fan-out (distinct args → distinct fingerprints); and the new phase added in lockstep to the `SessionPhaseName` union (`types.ts`), the `SessionPhaseNameSchema` Zod enum (`events.ts`), and the parity test. The diff touches only the telemetry window field, the observe method, its two call sites, the trace enum/union/schema, and tests — no enforcement, no control-flow change.
- **api-compat — clean.** All new exports (`SUSPECTED_LOOP_THRESHOLD`, `SUSPECTED_LOOP_WINDOW_SIZE`, `SuspectedLoopWindow`, `createSuspectedLoopWindow`, `fingerprintToolCall`, `countInWindow`, `observeToolCall`, `SuspectedLoopObservation`) are net-new symbols in a brand-new module — additive, with a single dispatcher-private production consumer (`dispatcher.ts`) and zero pre-existing importers to break. The `suspected_loop` enum widening is additive; the union↔schema parity is mechanically covered by the compile-time-exhaustive `Record<SessionPhaseName, true>` map (`suspected_loop: true`) plus bidirectional set-equality assertions (`session-phase.test.ts:219-257`).
- **test-coverage — every claimed behavior is tested at the ref.** Verified by reading the tests (not by executing them): threshold fire-once, debounce, top-level no-emit, batch path, age-out-of-window negative, ring-buffer bound, no-traceWriter path, distinct-arg fan-out negative, key-order normalization, interleaved loop, and fresh-turn reset are all present (22 `it` blocks in the detector file + 2 dispatcher coexistence cases + session-phase parity/acceptance). Every candidate absence claim was refuted by grep at the ref.
- **perf-observability — clean.** One sha256 over `name \0 canonicalJSON(input)` per forked tool call (interactive sessions pay zero via the `parentSessionId` gate) and an O(20) fixed-window scan per call — both negligible against the tool call they accompany. Emission is fire-and-forget; the new phase is a first-class, Zod-validated `SessionPhaseName` with typed `{ tool, count, windowSize }` metadata, so trace consumers parse it without a schema gap.

### Non-blocking findings

1. **low · security · `src/agent/tools/suspected-loop-detector.ts:114-125` (ref `7080e80`, file-state).** The `catch` fallback that runs when `canonicalStringify` fails is itself throw-capable: `input = String(call.input)` (line 122) re-throws if `call.input` is an object whose `toString`/`valueOf`/`Symbol.toPrimitive` throws, and that throw is the outermost guard (no inner try/catch), so it would propagate out of `fingerprintToolCall` into `observeSuspectedLoop` on the pre-execution dispatch path — the one theoretical way to violate the observe-only "never throw into dispatch" invariant. Reachability is low: a forked sub-agent's `call.input` is model-produced JSON (parsed), which has no throwing primitive-coercion hook, so this is a defense-in-depth gap rather than a reachable fault. `[low confidence — verify with runtime context]`
   - Evidence: `} catch {` → `input = String(call.input);`
   - Suggestion: wrap the fallback too — `try { input = String(call.input); } catch { input = '[unstringifiable]'; }` — so the fingerprint path is total over every input shape and the invariant holds unconditionally.

2. **nit · security · `src/agent/tools/suspected-loop-detector.ts:89-94` (ref `7080e80`, file-state).** `firedFingerprints: Set<string>` has no size cap, unlike the `recent` ring buffer (capped at `SUSPECTED_LOOP_WINDOW_SIZE`). It is not a live growth risk: its lifetime is one forked `query()` turn (reset via dispatcher reconstruction), and an entry is added only when a fingerprint reaches count ≥ 5 within a 20-round window, so entry count is bounded by O(tool-calls-per-turn / 5) short strings — dwarfed by the conversation state the turn already retains. Flagged only for completeness of the state-growth review.
   - Evidence: `recent: string[];` / `firedFingerprints: Set<string>;`
   - Suggestion: no change required; if belt-and-braces is wanted, clear `firedFingerprints` when `recent` fully turns over, or cap it at a small multiple of the window size.

### What was not checked

- Wave 1 agents read against branch HEAD `7080e8021869b307d4241d0e315454a5d4146b76`; every `file:line` citation in this review was independently verified at that ref by a dedicated verification pass, and no citation was fabricated and no invariant claim refuted.
- The test suite was **not executed** — the PR's "12778 passed / 0 failed" claim is unverified here (read-only review; test existence and assertions verified by reading the tests at the ref, not by observing a green run).
- This PR is stacked on #672 (`afk/idle-watchdog`); the diff reviewed is PR #674's own 7-file delta against that base and does not re-audit #672's changes. The `afk trace show` reader-side rendering for the new phase is intentionally out of scope (emission-only, per the PR's follow-up note).
- Stated intent: PR #674 title + body — spec-compliance assessed against it.

PR #676, opened 2026-07-23
fix(anthropic-direct): ESC interrupt halts promptly and never strands the next turn

griffinwork40:
## Problem

ESC soft-stop felt laggy and **ate the user's next message** — the reported "type a message after ESC, nothing happens, so I poke `.` to make the turn start." Worst on Opus extended-thinking turns.

Diagnosed as **two independent mechanisms** (both re-derived via shadow-verify, both proven with red→green tests):

1. **Halt latency.** `interrupt()` only fires an `AbortController`; the anthropic-direct stream loop had **no per-event abort check** (`translate.ts`), so the turn kept streaming until the SDK's *parked SSE read* settled — seconds behind the keypress for a thinking stream. Turn 1 lingered in `streaming`, so the next message queued instead of submitting.
2. **Wasted turn.** On abort, `loop.ts` yielded **two** terminal events (`error` *and* `turn.completed`). `AgentSession` breaks on the first (`agent-session.ts`), stranding the `turn.completed` — which the **next** turn's first pull consumed as a no-op `done`. So the post-ESC message ran a **full turn late**; the `.` poke is what advanced it.

The compositor's post-ESC auto-flush was even documented as safe *"because interrupt is synchronous"* — a false premise that reopened the very lag it claimed to close.

## Fix

- **`abortable-stream.ts` (new):** races each SSE pull against the turn signal → the stream halts within one event-loop turn regardless of the SDK's read cadence (fixes #1).
- **`loop.ts`:** wire `abortableStream(events, input.signal)`; on abort, suppress the in-band `error` event so an interrupted turn emits **exactly one** terminal `turn.completed` (fixes #2). Real (non-abort) errors are unchanged.
- **`terminal-compositor.input-mode.ts`:** corrected the false "interrupt is synchronous" comment to describe the real (now-prompt) behavior + the remaining subagent-teardown caveat.

Uses `input.signal` (user interrupt), **not** `ttfb.signal`, so the TTFB stall-timer path is untouched.

## Tests

- `abortable-stream.test.ts` (new, 7) — load-bearing case: throws promptly on a **parked** read instead of hanging.
- `interrupt-resume.test.ts` — new regression for the leftover-terminal/wasted-turn bug (**red on old code:** `'error' ≠ 'turn.completed'`; green after).
- **Full suite: 690 files, 12,726 passed, 0 failed;** `tsc --noEmit` clean.

## Follow-ups (stacked PR)

- Halt-latency **trace event** (field telemetry for interrupt→halt).
- **openai-compatible provider parity** — its query loop has the same shape and likely both mechanisms.

### DIFF
--- a/src/agent/providers/anthropic-direct/abortable-stream.test.ts
+++ b/src/agent/providers/anthropic-direct/abortable-stream.test.ts
@@ -46,6 +46,30 @@ describe('abortableStream', () => {
     await expect(pull).rejects.toMatchObject({ name: 'AbortError', message: 'interrupted' });
   });
 
+  it('does not wait for iterator cleanup after abort wins the race', async () => {
+    const ac = new AbortController();
+    const source: AsyncIterable<number> = {
+      [Symbol.asyncIterator]() {
+        return {
+          next() {
+            return new Promise<IteratorResult<number>>(() => {
+              /* parked read */
+            });
+          },
+          return() {
+            return new Promise<IteratorResult<number>>(() => {
+              /* cleanup is stuck behind the parked read */
+            });
+          },
+        };
+      },
+    };
+    const gen = abortableStream(source, ac.signal);
+    const pull = gen.next();
+    ac.abort('interrupted');
+    await expect(pull).rejects.toMatchObject({ name: 'AbortError', message: 'interrupted' });
+  });
+
   it('passes values through in order and ends cleanly on done', async () => {
     const ac = new AbortController();
     const out: number[] = [];
--- a/src/agent/providers/anthropic-direct/abortable-stream.test.ts
+++ b/src/agent/providers/anthropic-direct/abortable-stream.test.ts
@@ -1,0 +1,155 @@
+/**
+ * Regression tests for the interrupt-lag fix.
+ *
+ * The bug: on ESC the turn signal aborts, but the SDK's SSE async-iterator only
+ * surfaces the abort when its pending read settles — so a mid-stream
+ * extended-thinking (Opus) turn kept "streaming" for seconds after ESC, and the
+ * user's next message landed a turn late (they learned to poke "." to force it).
+ * `abortableStream` makes the halt deterministic by racing each pull against the
+ * signal. The load-bearing case is `throws promptly when a read is parked`.
+ */
+
+import { describe, it, expect, vi } from 'vitest';
+import { abortableStream } from './abortable-stream.js';
+
+async function* fromArray<T>(arr: T[]): AsyncIterable<T> {
+  for (const x of arr) yield x;
+}
+
+/** A source whose `next()` never resolves — models a parked SSE read. */
+function parkedSource(returnSpy?: () => void): AsyncIterable<number> {
+  return {
+    [Symbol.asyncIterator]() {
+      return {
+        next() {
+          return new Promise<IteratorResult<number>>(() => {
+            /* never resolves — the read is parked awaiting the next SSE frame */
+          });
+        },
+        return() {
+          returnSpy?.();
+          return Promise.resolve({ value: undefined as unknown as number, done: true });
+        },
+      };
+    },
+  };
+}
+
+describe('abortableStream', () => {
+  it('throws promptly when the signal aborts while a read is parked (the interrupt-lag bug)', async () => {
+    const ac = new AbortController();
+    const gen = abortableStream(parkedSource(), ac.signal);
+    const pull = gen.next();
+    // Without the abort race this pull would hang forever (the bug). The abort
+    // must win the race and reject on the next microtasks — not after a timeout.
+    ac.abort('interrupted');
+    await expect(pull).rejects.toMatchObject({ name: 'AbortError', message: 'interrupted' });
+  });
+
+  it('passes values through in order and ends cleanly on done', async () => {
+    const ac = new AbortController();
+    const out: number[] = [];
+    for await (const v of abortableStream(fromArray([1, 2, 3]), ac.signal)) out.push(v);
+    expect(out).toEqual([1, 2, 3]);
+  });
+
+  it('halts mid-stream: yields what arrived before ESC, then throws on abort', async () => {
+    const ac = new AbortController();
+    let deliver!: (r: IteratorResult<number>) => void;
+    const source: AsyncIterable<number> = {
+      [Symbol.asyncIterator]() {
+        return {
+          next() {
+            return new Promise<IteratorResult<number>>((resolve) => {
+              deliver = resolve;
+            });
+          },
+        };
+      },
+    };
+    const gen = abortableStream(source, ac.signal);
+    const first = gen.next();
+    deliver({ value: 7, done: false });
+    expect(await first).toEqual({ value: 7, done: false });
+    // Second read parks; ESC aborts before the next frame arrives.
+    const second = gen.next();
+    ac.abort('interrupted');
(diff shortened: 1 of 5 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

Reviewed at head ref `0c077e1` (base `main`). Full-regime review: 2 parallel dimension agents across security, api-compat, correctness, spec-compliance, test-coverage, perf-observability, plus a direct citation/importer-verification pass at the ref. **No blocking, critical, high, or medium findings.** Three low/nit items below; none blocks merge.

### Verified strengths (read against ref `0c077e1`)

- **correctness — clean, verified end-to-end.** The abort race is sound: `Promise.race([iterator.next(), aborted])` throws before `yield` on abort, so no value is lost between race resolution and yield; the already-aborted guard throws before `source[Symbol.asyncIterator]()` is ever touched (`abortable-stream.ts:32`, verified by the `pulled === false` test). The single-terminal fix was traced through the whole path: `abortableStream` throws AbortError → `translateMessageStream`'s catch converts it to an in-band `error` event (`translate.ts:334-338`) → the loop's error branch hits `if (input.signal.aborted) break` **without** yielding and **without** setting `translatorErrored` (`loop.ts:558-560`) → the post-loop `turnResult === null` path emits exactly one `turn.completed` (`loop.ts:610-618`). `sawProviderError` correctly stays false, so the turn seals as **cancelled, not failed** — matching the stated intent.

- **spec-compliance — all four named requirements met, no scope creep.** (1) Halt latency → new `abortable-stream.ts` racing each pull, wired at `loop.ts:465`. (2) Wasted turn → the `if (input.signal.aborted) break` suppression at `loop.ts:558-560`, validated against the real consumer contract (`agent-session.ts`: `if (output.type === 'done' || output.type === 'error') break`). (3) Compositor comment corrected — the false "interrupt is synchronous" premise is replaced with an accurate description of the two provider-side fixes (`terminal-compositor.input-mode.ts`). (4) Uses `input.signal`, not `ttfb.signal`, so the TTFB stall-timer path is untouched (`loop.ts:465` vs the unchanged TTFB wiring). The diff is confined to the abort-halt path plus comments.

- **security — clean.** Four specific concerns were each investigated at the ref and cleared: (a) **no listener leak** — the single `{ once: true }` abort listener (`abortable-stream.ts:42`) is removed in the `finally` on every exit path (`abortable-stream.ts:60`); (b) **no transport leak** — `iterator.return?.()` is invoked on all exits, and only *not awaited* on the abort-won branch (`void closeP`) so a `return()` parked behind the same bypassed read cannot re-block the halt, while the transport is independently cancelled via the shared signal; (c) **no misclassification** — `abortErrorFrom` is thrown only on the `ABORTED` branch, so a genuine `iterator.next()` rejection propagates as itself (the `propagates a source error unchanged` test confirms this); (d) **no secret exposure** — the synthesized `AbortError` message uses `signal.reason` only when it is a string, and every producer of the turn signal's reason uses a controlled vocabulary (`'interrupted'`, `'sigint'`, budget/timeout Errors), never user data.

- **api-compat — clean (additive), independently confirmed.** `abortableStream` is a net-new export with a **single production importer** — `loop.ts:51` (import) / `loop.ts:465` (use) — verified directly by reading each anthropic-direct provider file at the ref; the repo tree at the ref shows only the two new `abortable-stream.*` files (no name collision with any pre-existing module). The `as Parameters<typeof translateMessageStream>[0]` cast is **pre-existing** — the diff shows the identical cast on the old line without the wrapper — so it masks no *new* type hole; `abortableStream(events, …)` infers `T = unknown` and returns `AsyncGenerator<unknown>`, assignable to the pre-existing `AsyncIterable<unknown>`. No exported signature changed.

- **test-coverage — the claimed regression tests exist at the ref and assert what the PR states.** The load-bearing "throws promptly when a read is parked" case is present (`parkedSource()` never resolves, then abort, asserts prompt `AbortError` rejection); the `abortWonRace` not-awaited-return branch is covered (`does not wait for iterator cleanup after abort wins the race`); real-error-not-masked (`propagates a source error unchanged`) and the abandoned-read unhandledRejection swallow are covered. The single-terminal regression in `interrupt-resume.test.ts` drives the **real** `AnthropicDirectQuery` generator (not a mock), advances to the first terminal exactly as `AgentSession` does, asserts that terminal is `turn.completed`, and asserts turn 2 actually streams (`turnIdx === 2`) — i.e. no wasted turn.

- **perf-observability — clean.** One `iterator.next()` promise and one two-way `Promise.race` per pull, with a single long-lived abort listener registered once for the whole stream (not per token) — the correct low-cost shape for a high-frequency thinking-delta stream. The interrupt-halt trace event is explicitly deferred to a stacked follow-up (per the PR body), so its absence is not a finding; the existing `loop_end` trace still fires on the abort return path, so the turn's end stays legible.

### Non-blocking findings

1. **low · correctness · `loop.ts:527-562` (ref `0c077e1`, file-state).** When `input.signal.aborted` is true, the loop's error branch breaks before the `yield out.event`, so a genuine provider error arriving on the *same event-loop tick* as the interrupt is discarded rather than surfaced. This is the intended cancel-wins semantics (an interrupted turn seals as cancelled, and the turn is being torn down anyway), and the `abortableStream` layer independently guarantees a real error is never *masked as* an abort — but the ordering is load-bearing and undocumented.
   - Evidence: `if (input.signal.aborted) { break; } yield out.event; translatorErrored = true;`
   - *Suggestion:* add a one-line comment at `loop.ts:558` noting that abort-wins-over-error is deliberate, so a future edit does not "fix" it into yielding both terminals. `[low confidence — verify with runtime context]`

2. **low · test-coverage · `abortable-stream.ts:54-62` (ref `0c077e1`, file-state).** No test exercises `Promise.race` when `iterator.next()` resolves a value in the *same microtask* that `signal.abort()` fires — the `halts mid-stream` test delivers the value and fully awaits it before aborting on a subsequent parked read, so the simultaneous-resolution tie-break is never asserted. `Promise.race` honors resolution order (the value should win), so this is a correctness-preserving gap, not a defect.
   - *Suggestion:* add a case that calls `deliver(...)` and `ac.abort()` before any `await` to pin the tie-break. `[low confidence — verify with runtime context]`

3. **nit · test-coverage · `abortable-stream.test.ts` (ref `0c077e1`, file-state).** The PR body claims "abortable-stream.test.ts (new, 7)" but the file at the ref contains **8** `it(...)` cases (independently re-counted at the ref: parked-throw, no-wait-cleanup, values-in-order, halts-mid-stream, already-aborted, source-error, closes-iterator-on-abort, swallows-abandoned-read).
   - *Suggestion:* correct the PR body's test count to 8 (cosmetic; the extra coverage is a strength).

### What was not checked

- Wave 1 agents read against branch HEAD `0c077e132f2b978dfb5e8a682a58633544babc52`; every `file:line` citation in this review was read at that ref (`file-state` unless the line is in the diff hunk). The load-bearing api-compat claim (single importer, no collision) and the test-existence/count claims were independently re-derived at the ref via a direct tree walk + per-file grep.
- The test suite was **not executed** — the PR's "12,726 passed / 0 failed" and `tsc --noEmit` clean claims are unverified here (read-only review; test *existence* and assertions verified by reading the files at the ref, not by observing a green run). CI on this PR is green.
- Runtime ESC-interrupt timing was not exercised empirically; the halt-latency improvement is verified by construction (the abort race) and by the parked-read regression test, not by a live keypress-to-halt measurement.
- Stated intent: PR #676 title + body — spec-compliance assessed against it.

PR #677, opened 2026-07-23
feat(providers,trace): openai-compatible ESC-halt parity + interrupt-halt trace event

griffinwork40:
**Stacked on #676** (`afk/esc-input-lag-debug`). Review/merge #676 first; GitHub will retarget this to `main` once it lands. This PR is the deferred follow-up work from #676.

## A — openai-compatible provider parity

Applied both ESC-halt mechanisms from #676 to the openai-compatible provider. Confirmed empirically (revert probe + SDK read) that it had **both** — mechanism 2 in a *sharper* form:

- **Mechanism 1 (per-event abort race):** `driveStream`'s `for await (const event of stream)` had no per-pull abort check — it parked on the SDK read exactly like the anthropic-direct bug. Now wrapped with the shared `abortableStream(stream, ctx.controller.signal)`.
- **Mechanism 2 (single terminal on interrupt):** `openai@6`'s SSE iterator **swallows** a mid-stream abort and ends cleanly (`core/streaming.mjs`: `if (isAbortError(e)) return`), so pre-fix an early interrupt fell through to the stream-incomplete guard and yielded a spurious `StreamIncompleteError` (`type:'error'`) **plus** `turn.completed` — the same double-terminal that strands the next turn. The abort-race now throws promptly; a defensive `if (signal.aborted) return null` short-circuits before the guard → exactly one terminal.

`abortableStream` was **relocated** `anthropic-direct/ → providers/shared/` via `git mv` (history preserved; dependency-free, no SDK imports); anthropic-direct behavior is byte-identical (import re-point only).

## B — interrupt-halt trace event

New `interrupt_halt` session phase, emitted fire-and-forget from each provider loop's `finally`, `durationMs` = abort-fired → terminal `turn.completed`, gated on reason `'interrupted'` (excludes `close()`), only when a trace writer is present. Field-visible proof the halt stays within an event-loop turn.

## Tests
- New: `providers/shared/abortable-stream.test.ts` (moved), `openai-compatible/query/stream-drive.test.ts` (+3), `openai-compatible/query.test.ts` (+4), `anthropic-direct/loop.interrupt-halt.test.ts` (+4), `trace/session-phase.test.ts` (+1).
- Test validity confirmed by revert (prompt-halt test times out without the race; single-terminal tests fail with the spurious `StreamIncompleteError`).
- **Full suite: 691 files, 12,738 passed, 0 failed;** `tsc --noEmit` clean; `audit:sdk:check` / `audit:env:check` / `scan:env:check` all OK.


griffinwork40 (commented):

## Merge decision: ✅ MERGE

Reviewed ref: `76bece3d8321f36dd091ad96481c14ca8350315b` — all citations verified at that SHA (via `gh api` file-state reads and a git-verified-clean local checkout of the exact reviewed commit). Regime: full (886 diff lines, feature). Dimensions covered: security, api-compat, correctness, spec-compliance, test-coverage, perf-observability.

> Note: this PR is stacked on #676 with base `afk/esc-input-lag-debug`, which is why GitHub reports it as CONFLICTING. That resolves once #676 lands and this retargets to `main`; it is not a defect in this diff.

### Findings

**No blocking, critical, or high findings.** The change is focused and well-instrumented:

- **security — clean.** `shared/abortable-stream.ts` has zero imports (no SDK, no side effects); the relocation is purely locational. The abandoned-read rejection is swallowed only on the abort-won branch (`abortable-stream.ts:56–62`) — genuine source rejections still propagate, so no error masking. The new `interrupt_halt` trace payload carries only `durationMs` + a static `{ provider }` string — no secret/PII exposure. Fire-and-forget `void emitSessionPhase(...)` is crash-safe (writer errors are try/catch-swallowed; undefined writer no-ops), so it cannot stall a returning turn.

- **api-compat — clean (additive).** `abortable-stream.ts` moved `anthropic-direct/ → shared/` via `git mv`. Independently verified across all 64 production provider files at the reviewed ref: **zero production importers of the old path**; `anthropic-direct/loop.ts:51` and `stream-drive.ts:23` both re-point to `../shared/abortable-stream.js`. The `interrupt_halt` enum member is appended (not renamed/removed) to both `SessionPhaseNameSchema` (`events.ts:400`) and the `SessionPhaseName` union (`types.ts:648`); `abortableStream`'s public signature is unchanged.

- **correctness — clean.** Mechanism 1 (per-pull abort race) present at `stream-drive.ts:149`; Mechanism 2 (defensive `if (ctx.controller.signal.aborted) return null` before the stream-incomplete guard) at `stream-drive.ts:202`. The `interrupt_halt` gate `controller.signal.reason === 'interrupted'` (`query.ts:423`) correctly excludes `close()` (reason `'closed'`). One long-lived `{ once: true }` listener with a null-guard — no listener leak, no double-emit, non-negative `durationMs`. anthropic-direct mirrors this (`loop.ts:1084`).

- **spec-compliance — clean.** Every named requirement in the PR body is implemented (both mechanisms, the `git mv` + import re-point with byte-identical anthropic-direct behavior, the new trace phase in both schema surfaces, fire-and-forget emit gated on reason + writer presence). No scope creep.

- **test-coverage — clean.** All claimed test files exist at the reviewed ref (old `anthropic-direct/abortable-stream.{ts,test.ts}` confirmed absent — move is complete). The subtle `close()`-vs-`interrupt()` gating is exercised in **both** providers (`openai-compatible/query.test.ts:2942` interrupt-present, `:2974` close-absent; `anthropic-direct/loop.interrupt-halt.test.ts:120` close-absent, `:163` no-writer path).

### Non-blocking nit (low confidence — verify with runtime context)

- **perf-observability · low.** A *pending* interrupt arriving between turns (`pendingAbortReason === 'interrupted'`) aborts the fresh controller and returns early (`query.ts:379→382`) **before** the listener/gate is installed (`query.ts:404–423`), so no `interrupt_halt` event is emitted for that rare between-turns path. This is not a correctness bug — the turn correctly yields no terminal — but the halt latency goes unrecorded for that case. If between-turn halt latency matters, stamp `interruptedAt` from `pendingAbortReason` before the early return. Non-blocking.

### What was not checked

- Read against branch HEAD `76bece3d` — all quoted lines verified at that ref; the reviewed SHA was re-confirmed as still the PR head at review time (no push since analysis).
- Stated intent: PR #677 title + body — spec-compliance assessed against it.
- Did not execute the test suite (the PR reports 12,738 passed / 0 failed; CI checks on this PR are green apart from the expected stacked-base conflict). Did not exercise runtime ESC-interrupt timing empirically.

_Read-only automated review. No approval/request-changes was cast; a maintainer will make the merge call._

PR #678, opened 2026-07-23
feat(subagent): harden handoff against mid-stream truncation

griffinwork40:
## Problem

Subagents consistently failed with:

> the model stream ended without a terminal message (no `message_stop` and no `stop_reason`): the response was cut off mid-stream, typically an intermediary closing the connection.

**Root cause** (verified in code, not assumed): the child's final assistant message is returned verbatim as a `tool_result` (`foreground-promotion.ts`). That final report is generated as one long-lived model stream; when a child accumulates a large context and then emits a long report, an intermediary (proxy/gateway/LB) closes the connection mid-generation. `translate.ts:319` surfaces this as a `StreamIncompleteError`, and `loop.ts` did **not** retry this class (only TTFB stalls and `overloaded_error`/529 were retried) — so a single transient cut failed the whole subagent, and a naive re-dispatch reproduced it.

This handoff shape (verbatim final message → tool_result) is the same one Claude Code / OpenCode / Crush / Aider use; the robust harnesses (OpenHands, Goose, Cline) pair it with a **file/structured handoff** and/or **transient retry**. This PR adds both of those layers.

## Changes

**1. Default subagent handoff contract** — new zero-dep `src/agent/subagent-contract.ts` (`SUBAGENT_HANDOFF_CONTRACT`): write-capable default workers are told to keep the final reply compact and write bulk output to files (path + summary) instead of pasting it inline. Wired into:
- `GENERAL_PURPOSE_PROMPT` (`builtins.ts`)
- the unnamed-dispatch prompt (`child-config.ts`)
- the parent-facing `agent`-tool description (`schemas.ts`)
- read-only vendored agents (research-agent, git-investigator) left untouched (byte-pinned, already compact)

**2. Bounded mid-stream-cut retry** (`loop.ts`): a `StreamIncompleteError` in-band error event (no terminal signal after a first byte was seen) is re-driven up to `STREAM_INCOMPLETE_MAX_RETRIES=2` with a short 1s→2s settle delay, mirroring the existing TTFB/overload re-drive. Kept low + short on purpose: each failed attempt burns a partial (often long) generation, unlike a fast 529 rejection.

## Verification

- `pnpm lint` (tsc --noEmit) ✓
- `pnpm test` — 689 files, 12,722 tests, 0 failures ✓ (3 new loop-retry tests: recover / exhaust / abort-during-backoff; updated child-config + subagent-executor expectations)
- `pnpm test:coverage` — no threshold breach ✓
- `audit:sdk:check` / `audit:env:check` / `scan:env:check` ✓ (no new SDK symbols or env vars)

## Deferred (not in this PR)

- Expose a structured/file return on the `agent` tool (OpenHands `outputs` dict / Goose `response.json_schema`) — the tool currently has no `outputSchema` field. Follow-up if the contract + retry don't fully close it.

🤖 Generated with agent-afk


griffinwork40 (commented):

## Merge decision: ✅ MERGE

Reviewed at head ref `2067289` (base `main`). Full-regime review (2 parallel dimension agents across security, api-compat, correctness, spec-compliance, test-coverage, perf-observability + a citation/absence verification pass). **No blocking, critical, high, or medium findings.** One low-severity test-gap below; it does not block merge.

> Note: the PR is currently `CONFLICTING` (mergeable state), so a rebase/conflict-resolve against `main` is needed before it can actually merge. That is mechanical and out of scope for this read-only review.

### Verified strengths (read against ref `2067289`)
- **correctness — clean.** The new `StreamIncompleteError` re-drive is bounded and abort-safe. Backoff is `STREAM_INCOMPLETE_BASE_DELAY_MS * Math.pow(2, n-1)` = 1s → 2s (`loop.ts:122,676`), capped at `STREAM_INCOMPLETE_MAX_RETRIES = 2` (`loop.ts:121`). The retry decision is triple-gated — `instanceof StreamIncompleteError && streamIncompleteRetries < MAX && !input.signal.aborted` (`loop.ts:560`) — and the post-backoff path re-checks the signal, yielding `turn.completed` + `return` on abort (so an interrupt during the settle delay leaves `callCount == 1`). `input.messages` is unmutated for the round, so the re-drive re-sends identical history with no tool side-effect re-execution.
- **security — clean.** Single `.unref()`'d timer via `sleepWithAbort` with a `{ once: true }` abort listener — no timer/abort leak. The counter is declared at `loop.ts:326` and reset to 0 per clean round at `loop.ts:700` (mirroring the existing `overloadRetries` reset), so a long multi-round turn cannot accumulate unbounded retries. The overload and stream-incomplete branches are separate `break → if…continue` blocks (`loop.ts:540/558/620/652`); exactly one retry flag is set per stream iteration, so they cannot fire together or multiply. `subagent-contract.ts` is a pure static string with no interpolation — no injection/secret surface.
- **api-compat — clean.** All three new exports are additive with single definition sites and no collisions: `SUBAGENT_HANDOFF_CONTRACT` (1 def, 2 production importers: `builtins.ts`, `child-config.ts`), `STREAM_INCOMPLETE_MAX_RETRIES` / `STREAM_INCOMPLETE_BASE_DELAY_MS` (single def each in `loop.ts`). The `stream.retry` event reuses the existing `ProviderEvent` union and its already-wired consumers — zero consumer changes. The `child-config.ts` systemPrompt ternary (`child-config.ts:274-279`) appends (never drops) the parent base prompt for unnamed dispatch, passes named-agent prompts through unchanged, and falls back to the contract alone when no base prompt exists.
- **spec-compliance — intent fully met, no scope creep.** Every named requirement has an implementing change: the contract module wired into `GENERAL_PURPOSE_PROMPT` (`builtins.ts`), the unnamed-dispatch prompt (`child-config.ts`), and the agent-tool description (`schemas.ts`); the bounded retry in `loop.ts`. The "read-only vendored agents left untouched" clause is confirmed (`research-agent.ts` / `git-investigator.ts` show an empty diff vs `main`). The deferred structured/file return is genuinely absent (`outputSchema` grep on `schemas.ts` = 0 matches), matching the "Deferred" section — not an unmet requirement.
- **perf-observability — clean.** The retry's trace emission is fire-and-forget (`void emitSessionPhase(... phase: 'rate_limit', metadata: { reason: 'stream-incomplete', attempt })`), so trace latency never stalls the retry, and the new class is still legible in `afk trace show` via the distinct `reason`. Cost is entirely off the success path and bounded to 2 attempts.
- **test-coverage — the 3 claimed retry tests are present and assert what the PR states** (`loop.retry.test.ts` at ref: recover `callCount == 2`; exhaust `STREAM_INCOMPLETE_MAX_RETRIES + 1` calls + `error.name === 'StreamIncompleteError'`; abort-during-backoff `callCount == 1`). Contract-wiring tests are present across `child-config.test.ts`, `subagent-executor.test.ts`, and `subagent-executor.named-agents.test.ts`.

### Non-blocking finding

1. **low · test-coverage** — `src/agent/providers/anthropic-direct/loop.ts:700` (ref `2067289`, file-state): the per-round reset `streamIncompleteRetries = 0` is reachable but untested — no test dispatches two clean tool-use rounds to prove a second round gets a fresh 2-retry budget (independently verified: a grep of `loop.retry.test.ts` for `second round | two rounds | multi-round | round 2 | next round` returns zero matches; all StreamIncompleteError scenarios are single-round). The reset itself is correct (it mirrors the proven `overloadRetries` reset on the same line); the branch is simply unproven.
   - *Suggestion:* add a test where round 1 recovers from a cut, then round 2 (after a tool call) also cuts, asserting the round-2 retry budget is not pre-decremented.

### What was not checked
- Wave 1 agents read against branch HEAD `2067289c9b83c1a9a7a4af964968a5711fd66523`; all `file:line` citations in this review were verified at that ref, and the sole absence claim was independently re-derived (zero grep matches).
- The test suite was **not executed** — the PR's "12,722 tests / 0 failures" claim is unverified here (read-only review; citations verified against ref `2067289`, tests not run).
- `translate.ts` internals were not audited for whether `StreamIncompleteError` can be spuriously synthesized on a legitimately-empty clean stream (out of the reviewed dimensions; the retry is bounded regardless).
- The `CONFLICTING` merge state was observed but not resolved (read-only; conflict resolution is out of scope).
- Stated intent: PR #678 title + body — spec-compliance assessed against it.

Issue #682, opened 2026-07-23
Tech debt: reduce incidental complexity in the TUI terminal-compositor (~2.2k LOC) via a gated /simplify pass


griffinwork40:
## Context

Surfaced while reviewing #681 (per-chat Telegram send targeting). That PR's own code is small, tested, and needs no simplification — but the review prompted a look at where a `/simplify` reduction pass would actually pay off. Answer: **not fresh feature code, but old accreted core.** The clearest high-yield / lower-risk target is the TUI terminal-compositor.

## Target

| File | ~LOC |
|------|-----|
| `src/cli/terminal-compositor.input-dispatch.ts` | 1143 |
| `src/cli/terminal-compositor.ts` | 1098 |

~2,240 LOC across the pair — TUI input handling + rendering, a classic home for accreted special-cases, and (unlike the core model loop) visibly and safely testable.

## Proposed approach

1. Run `/simplify` **read-only**, scoped to the two files — its four discovery lenses (clone detection, dead code, complexity hotspots, wrong abstraction) produce a *ranked, behavior-preserving* reduction plan. No edits in this step.
2. Triage per-finding; discard anything that trades incidental complexity for a wrong abstraction.
3. Apply only clear wins via `/refactor`, each **gated on the full test suite** (`pnpm test`) + review. Land in small reviewable increments.

## Scope / caveats

- **Behavior-preserving only** — a reduction pass, not a feature or redesign.
- **Load-bearing core** — the compositor drives the interactive REPL; regressions are user-visible. Every change gated by tests + review.
- **Size ≠ complexity** — validate each finding; don't refactor for LOC alone.

## Related candidates (future, higher-risk — NOT this issue)

`src/agent/session/agent-session.ts`, `src/agent/tools/dispatcher.ts`, provider `loop.ts` / `query.ts`, `src/agent/subagent.ts` — hotter/riskier core paths; separate scoped efforts if ever.

**Explicitly out of scope (low yield):** `src/config/env.ts`, `src/agent/tools/schemas.ts` — large *by design* (flat registry / declarative schemas), CI-gated structure.

---
_Filed from an AFK session; origin was post-#681 /simplify triage. #681 itself needs no simplification._

PR #684, opened 2026-07-23
refactor(cli): /simplify wins for the terminal-compositor pair (#682)

griffinwork40:
## What this does

Addresses #682 — a gated `/simplify` reduction pass over the TUI terminal-compositor pair
(`terminal-compositor.ts` + `terminal-compositor.input-dispatch.ts`). Two behavior-preserving commits:

1. **Centralize the "commit-if-changed" edit contract** in `input-dispatch.ts`. The idiom
   *"apply a pure `InputCore` transition only when it changes the buffer, resetting history recall
   first on a real edit"* was hand-repeated at 5 key bindings (meta+Backspace, Ctrl+W/U/K,
   meta+Delete); the atomic paste-placeholder-delete guard at 2 more (Backspace/Delete). Extracted
   two named helpers — `commitIfChanged()` and `applyAtomicPlaceholderDelete()` — following the
   file's existing free-functions-on-host pattern. Each call site is a 1:1 substitution.
2. **Remove two dead setters** `setOnSoftStop` / `setOnShiftTab`. Both have zero call sites in the
   tree and zero in all git history (pickaxe). The underlying handlers are wired via the
   constructor (`opts.onSoftStop` / `opts.onShiftTab`), so soft-stop and Shift+Tab behavior is
   unchanged — only the never-used redundant post-construction mutators are gone.

**This is not a LOC reduction** (verbose house-style JSDoc on the helpers offsets the call-site
savings). The win is *contract centralization*: the resetRecall-before-edit invariant now lives in
one named place a new binding can't silently forget.

## Why the PR is small (the honest finding)

The issue framed the pair as "~2.2k LOC ... a classic home for accreted special-cases." A four-lens
read-only pass (duplication / dead-code / complexity / wrong-abstraction) found that **the large
structural win the issue imagined has already happened**: the compositor is decomposed into ~15
sibling modules — render/frame/scrollback-eviction live in `terminal-compositor.frame.ts`,
`.committed-band-commit.ts`, `.render.ts`, etc. `terminal-compositor.ts` is the state-hub/delegator,
`input-dispatch.ts` the keystroke dispatcher; both are **mostly essential complexity**.

Three lenses independently returned "essential — do not touch" for the biggest functions
(`handleEnter`, `handleCursorAndEdit`, `handleVerticalNav` — all carry explicit anti-refactor
invariants encoding regressions #403/#467/#644). The net clear-win surface was only what this PR
lands. Two corollaries worth recording:

- The old `afk/split-compositor-frame-render` branch is **dead** (far behind main; its split already
  landed differently) — no collision.
- **`size ≠ complexity` held up**: the remaining LOC is largely earned. Consider updating #682 to
  reflect that the big structural target is already addressed.

## Deliberately deferred (NOT in this PR)

- **`setOnBackground` — possible latent bug, not deleted.** It's also uncalled, but its JSDoc *and*
  `src/cli/input/input-surface.ts:131` both claim a per-turn swap happens "via `setOnBackground`"
  that no code performs. That's a possible wiring gap (idle-mode Ctrl+B background handler not
  cleared between turns), not clean dead code — flagged for a separate `/diagnose` rather than
  silently removed. Good follow-up issue candidate.
- **Tier 2 structural extractions** (`handleEscape` double-Esc block, `handleEnter` guard phases) —
  adjacent to the #403/#467/#644 regression lineage, ~0 net LOC, real risk. Not worth it.

## Verification

- `pnpm lint` (tsc --noEmit, strict) — clean.
- **Full suite: 692 files, 12,879 passed, 14 skipped, 0 failures** — including the compositor tests
  that directly exercise this change: `word-nav` (asserts per-binding `resetRecall` counts),
  `keypress`, `paste`.
- Each commit is independently reviewable/revertable; behavior-preserving throughout.

Closes nothing automatically — leaving #682 open for the maintainer to close or to retarget at the
`setOnBackground` follow-up.

### DIFF
--- a/src/cli/terminal-compositor.input-dispatch.ts
+++ b/src/cli/terminal-compositor.input-dispatch.ts
@@ -836,18 +836,49 @@ function handleEnter(self: KeyDispatchHost, key: KeyInfo, sequence: string): boo
   return true;
 }
 
+/**
+ * The buffer-edit commit contract shared by the "transform, commit only if it
+ * changed" bindings (meta word-delete, Ctrl+W/U/K kills): apply a pure
+ * InputCore transition only when it actually changes the buffer, resetting
+ * history recall FIRST on a real edit (the convention documented in
+ * handleCursorAndEdit). Returns `true` so a handled binding can tail-call it
+ * as `return commitIfChanged(self, InputCore.x(self.input))`.
+ */
+function commitIfChanged(self: KeyDispatchHost, next: InputCoreState): true {
+  if (next !== self.input) {
+    self.history?.resetRecall();
+    self.applyEdit(next);
+  }
+  return true;
+}
+
+/**
+ * Atomic paste-placeholder delete shared by Backspace (backward) and Delete
+ * (forward): when the cursor abuts a `[Pasted text #N +M lines]` token, the
+ * whole token (and its side-table entry) is removed in one keystroke. Returns
+ * `true` when a token was consumed (caller stops), `false` to fall through to
+ * the normal character delete.
+ */
+function applyAtomicPlaceholderDelete(
+  self: KeyDispatchHost,
+  direction: 'backward' | 'forward',
+): boolean {
+  const atomic = Paste.maybeAtomicPlaceholderDelete(self, direction);
+  if (atomic) {
+    self.history?.resetRecall();
+    self.applyEdit(atomic);
+    return true;
+  }
+  return false;
+}
+
 function handleBackspace(self: KeyDispatchHost, key: KeyInfo): boolean {
   if (key?.name !== 'backspace') return false;
   // Option+Delete on macOS → meta+backspace: delete previous word.
   // Mirrors reader.ts:677 so word-erase is consistent across both
   // input surfaces.
   if (key?.meta) {
-    const next = InputCore.deleteWordBackward(self.input);
-    if (next !== self.input) {
-      self.history?.resetRecall();
-      self.applyEdit(next);
-    }
-    return true;
+    return commitIfChanged(self, InputCore.deleteWordBackward(self.input));
   }
   // Atomic placeholder delete — when the cursor sits at the
   // trailing `]` of a `[Pasted text #N +M lines]` token, single
@@ -855,12 +886,7 @@ function handleBackspace(self: KeyDispatchHost, key: KeyInfo): boolean {
   // entry). Without this, deleting a freshly-pasted blob requires
   // ~30 backspaces. Run BEFORE the InputCore.backspace fallback
   // so the atomic path wins when both would fire.
-  const atomic = Paste.maybeAtomicPlaceholderDelete(self, 'backward');
-  if (atomic) {
-    self.history?.resetRecall();
-    self.applyEdit(atomic);
-    return true;
-  }
+  if (applyAtomicPlaceholderDelete(self, 'backward')) return true;
   const next = InputCore.backspace(self.input);
   if (next !== self.input) {
     self.history?.resetRecall();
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

Reviewed at head ref `cfe7741d` (base `main`). Light-regime review (1 dimension agent across security · api-compat · correctness · spec-compliance · test-coverage · perf-observability), plus a direct citation/importer-verification pass at the ref. **No blocking, critical, high, or medium findings.** Two behavior-preserving commits; the "1:1 substitution" and "dead code" claims both hold up under independent re-derivation at the ref.

### Verified strengths (read against ref `cfe7741d`)

- **correctness — behavior PRESERVED across all 7 substituted call sites.** `commitIfChanged` is a byte-exact extraction of the inline idiom — `if (next !== self.input) { self.history?.resetRecall(); self.applyEdit(next); } return true;` (`input-dispatch.ts:847-853`, file-state) — applied 1:1 at meta+Backspace (`:881`), Ctrl+W (`:970`), Ctrl+U (`:977`), Ctrl+K (`:983`), meta+Delete (`:1055`). `applyAtomicPlaceholderDelete` reproduces the atomic block exactly (`resetRecall → applyEdit → return true` on hit, `return false` on miss; `:862-873`). The two control-flow-changing sites were checked at the post-merge file state: the Backspace atomic site (`:889`) still falls through to the unchanged `InputCore.backspace` → `resetRecall/applyEdit` → `else if attachments` tail, and the Delete atomic site (`:1061`) still falls through to `resetRecall(); applyEdit(InputCore.deleteForward(...)); return true;` — both byte-equivalent.

- **api-compat — clean (removal is safe).** `git grep 'setOnSoftStop\|setOnShiftTab'` at the ref returns **0 matches** across the entire tree (production and tests) — independently re-derived, not taken from the PR body. `TerminalCompositor` is **not re-exported from `src/index.ts`** (verified by reading the file at the ref), so there is no external-consumer surface. The `git log -S` pickaxe hits for both symbols resolve to the definition lines only, never a call — confirming the "zero call sites in tree AND history" claim.

- **correctness — dead-setter removal does not change behavior.** The underlying handlers are still wired via the constructor, independently of the deleted setters: `this.onSoftStop = opts.onSoftStop;` (`terminal-compositor.ts:526`, file-state) and `this.onShiftTab = opts.onShiftTab;` (`:529`). The removed `setOnSoftStop`/`setOnShiftTab` were the only never-used post-construction mutators; soft-stop and Shift+Tab behavior is unchanged.

- **spec-compliance — intent fully met, no scope creep.** Every named requirement has an implementing change: the two helpers centralize the resetRecall-before-edit contract (commit 1), and the two dead setters are removed (commit 2). The "NOT a LOC reduction — the win is contract centralization" framing is accurate. The deliberately-deferred items (`setOnBackground`; Tier 2 structural extractions) are genuinely absent from the diff, matching the "Deferred" section.

- **test-coverage — the affected bindings are already exercised (absence checks grounded).** Existing tests cover every touched binding at the ref: Ctrl+W/U/K and meta+Backspace/Delete in `terminal-compositor.word-nav.test.ts`, atomic paste-delete in `terminal-compositor.paste.test.ts`. The exact contract `commitIfChanged` centralizes — resetRecall ordering — is asserted (`word-nav.test.ts` `expect(resetRecall).toHaveBeenCalledTimes(1)`). Zero test files reference the deleted setters, so none needs updating and no coverage is lost.

- **security / perf-observability — clean.** No auth/secret/exec/network surface is touched (pure internal keystroke-dispatch state). The extraction adds one function-call frame per keystroke with no new allocation, I/O, or async path.

### What was not checked

- Read entirely from ref `cfe7741d36206795287b78015337e97858ce5fb2` via `git show`/`git grep`; all `file:line` citations were verified at that ref, and the two load-bearing claims (0 production callers for the removed setters; constructor wiring intact) were independently re-derived by the reviewer.
- The test suite was **not executed** — the PR's "12,879 passed / 0 failures" claim is unverified here (read-only review; test *existence* and the resetRecall assertions were verified by reading the files at the ref, not by observing a green run). CI on this PR is green.
- `setOnBackground`'s alleged latent wiring bug is **explicitly deferred by the PR** (flagged for a separate `/diagnose`) and was not investigated here — it is out of scope for this diff.
- Stated intent: PR #684 title + body — spec-compliance assessed against it.

PR #690, opened 2026-07-24
feat(grep): back the grep tool with bundled ripgrep for cross-platform support

griffinwork40:
## What

Swap the `grep` tool's search engine from system **`grep`** → **`@vscode/ripgrep`'s bundled `rgPath`**, and add an advisory cross-platform CI matrix. This is the first landed slice of the Windows-compat effort (Phase 0: CI scoreboard + Phase 1.4: grep tool).

System `grep` is **absent on Windows** and drifts between GNU/BSD across Linux/macOS. Ripgrep ships as a platform binary via optional deps, so the tool now behaves identically on Linux, macOS, and Windows. Mirrors Claude Code's ripgrep-backed Grep tool.

## ⚠️ Contract change (breaking, intentional)

Ripgrep has **no basic-regex (BRE) mode**, so:
- The BRE default and the **`extended` param are removed**.
- `|` `+` `?` `(` `)` `{` `}` are now **regex metacharacters by default** (e.g. `foo|bar` alternates, matching either branch). Escape with a backslash for the literal character.
- The old no-match "BRE `|` footgun" hint is gone (no longer applicable).

The tool schema description is updated accordingly. This is the same trade Claude Code accepted for its Grep tool.

## Changes

- **`grep.ts`** — spawn `rgPath` with `-n --no-heading --color=never --hidden -g '!.git'`. Honors `.gitignore` (so `node_modules`/`dist` stay skipped) but **does** search hidden files (`.github`, `.env`) that the old `grep -rn` reached; the `.git` dir and binary files are excluded. (Note: rg's `-r` is `--replace`, deliberately never passed.)
- **`_rg-availability.ts`** (new) — on the error path only, distinguish a missing/non-executable `rgPath` (an uninstalled platform optional-dep) from the dead-cwd `ENOENT` masquerade (#441). Checked **first** in the spawn error handler so a bad binary is never misattributed to a reaped worktree.
- **`schemas.ts`** — rewrite grep tool + pattern descriptions; drop the `extended` param.
- **`ci.yml`** — run the `test` job on a `[ubuntu, windows, macos]` matrix. **ubuntu stays the required gate**; windows/macos are **advisory** (`continue-on-error`, `fail-fast: false`, per-OS coverage artifact names) — they run for signal but do not block merges yet.
- **`package.json` / `pnpm-lock.yaml`** — add `@vscode/ripgrep`.

## Verification

- `tsc --noEmit` clean (whole project).
- `grep.test.ts` (43 tests) + `_rg-availability.test.ts` (4 tests) green locally on macOS.
- Rebased onto current `main` (5.74.0); the merge is conflict-free and the lockfile is in sync (`pnpm install --frozen-lockfile` passes).
- **Windows/macOS runtime correctness is exactly what this PR's own CI matrix is here to prove** — the matrix legs are the authoritative falsifier. Kept as a **draft** until that signal lands and the breaking grep contract change gets a human's review.

## Not in scope (tracked in the Windows-compat roadmap)

The `bash` tool (Windows shell resolution + process-tree kill), credential-denylist Windows correctness, and daemon/service persistence on Windows are separate follow-ups.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,9 +78,21 @@ jobs:
         run: pnpm build
 
   test:
-    name: Test
-    runs-on: ubuntu-latest
-    timeout-minutes: 10
+    name: Test (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    # Cross-platform matrix (Windows-compat effort, see .afk/plans/windows-compat-roadmap
+    # or the tracking issue). ubuntu-latest is the REQUIRED gate; windows-latest and
+    # macos-latest run for SIGNAL but do not block merges yet — `continue-on-error`
+    # keeps them advisory while the compat grind is in progress. `fail-fast: false`
+    # so one OS failing never cancels the others. Public repos get unlimited free
+    # GitHub-hosted minutes on all three runners. Flip windows/macos to required
+    # (delete the continue-on-error line) once the suite is green there.
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    continue-on-error: ${{ matrix.os != 'ubuntu-latest' }}
+    timeout-minutes: 15
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -107,7 +119,8 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: coverage-report
+          # Per-OS name — a matrix can't upload three artifacts under one name (v4 immutable).
+          name: coverage-report-${{ matrix.os }}
           path: coverage/
           retention-days: 14
 
--- a/package.json
+++ b/package.json
@@ -112,6 +112,7 @@
     "@anthropic-ai/sdk": "^0.74.0",
     "@modelcontextprotocol/sdk": "^1.29.0",
     "@mozilla/readability": "^0.6.0",
+    "@vscode/ripgrep": "^1.18.0",
     "ansi-escapes": "^7.3.0",
     "better-sqlite3": "^12.9.0",
     "chalk": "^5.3.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -17,6 +17,9 @@ importers:
       '@mozilla/readability':
         specifier: ^0.6.0
         version: 0.6.0
+      '@vscode/ripgrep':
+        specifier: ^1.18.0
+        version: 1.18.0
       ansi-escapes:
         specifier: ^7.3.0
         version: 7.3.0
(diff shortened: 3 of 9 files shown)


griffinwork40:
Pushed `d01c6d4` — fixes the Windows path-containment failures surfaced by the new `windows-latest` matrix.

**Root cause** (not the same as it first looked): `/etc` is *not* mis-classified as relative on Windows — `path.win32.isAbsolute('/etc')` is `true`. The real bug is in the shared `computeContainment` (`_cwd-utils.ts`): when the resolved path and the allowed root sit on **different drives**, `path.win32.relative()` returns a drive-qualified absolute string (`D:\etc`, or a drive-less `\etc`) instead of a `..\`-prefixed relative one. The bare `!rel.startsWith('..')` containment test then read that as "inside the root" and admitted every escaping absolute path.

This is why all six filesystem handlers (grep / read-file / write-file / edit-file / glob / list-directory) failed the identical `rejects absolute path outside context.cwd` test on Windows — they share the helper.

**Fix:** guard the verdict with `!path.isAbsolute(rel)`. `path.relative` only returns an absolute path when the two inputs share no common root — exactly the escape case. On POSIX, `relative()` between two absolute paths is never absolute, so it's a no-op there (green Linux/macOS matrices unaffected).

**Verification:**
- `path.win32` simulation flips the verdict allow -> reject for cross-drive (`D:\etc`) and drive-relative (`\etc`) inputs; same-drive escapes were already rejected.
- Existing 6 handler suites + `_cwd-utils.test.ts` stay green on macOS (175 passed / 1 skipped); `tsc --noEmit` clean.
- The existing `windows-latest` `rejects absolute path outside context.cwd` tests are the regression coverage and should go green on this run.

Note: this only addresses the containment cluster. The other `windows-latest` reds (test-harness `~/.afk` HOME-override in `paths.test.ts`, and `.mjs` dynamic-import needing `pathToFileURL()`) are separate pre-existing cross-platform items, untouched here.

griffinwork40:
Empirical confirmation of both security-critical claims:

- **(a) `.git` exclusion — CONFIRMED clean.** With the exact handler args, rg matched `sub/normal.txt` but **not** `.git/config` (which held a planted secret). No secret exposure despite `--hidden`.
- **(b) `--pre` RCE neutralization — CONFIRMED, and the vector is real.** With `--`, the `--pre=touch …` pattern was a literal no-match (exit 1, no file created). The control *without* `--` produced rg's smoking-gun error `preprocessor command could not start: '"touch PWNED_PROOF2"…'` — proving rg *does* activate its `--pre` command-execution flag when the pattern isn't fenced. The `--` separator is a genuine, load-bearing security control, not theater.

All load-bearing claims independently re-derived (not taken on subagent confidence). No critical/high findings → mandatory `/shadow-verify` gate does not fire; Wave 1.5 has no top-severity citations/absence-claims to check. Synthesizing.

---

# Review — PR #690 · `feat(grep): back the grep tool with bundled ripgrep for cross-platform support`

**Target:** PR #690 · branch `afk/linux-windows-compat` → `main` · reviewed ref `d79ea56` · +409/−137 across 9 files (lockfile + one-line dep add excluded) · regime **full** (feature + breaking contract change).
**Stated intent:** PR #690 title+body (5 goals + Windows containment fix) — **spec-compliance was assessed** against it.

## Findings (sorted: security first, then severity)

**Security — PASS (no findings above nit; empirically verified).**
- `nit · high-confidence · security · grep.ts:145 · ref:d79ea56 · file-state` — `args.push('--', pattern, path)` fences pattern+path behind the end-of-options separator; a prompt-injected `--pre=<cmd>` / `-f/etc/passwd` pattern is neutralized. **Empirically confirmed**: with `--`, `--pre=touch PWNED` was a literal no-match; without it, rg activated its preprocessor (`preprocessor command could not start …`). Load-bearing control. No fix needed.
- `nit · high-confidence · security · grep.ts:131 · ref:d79ea56 · file-state` — `include` is passed as the separate argv token after `-g` (spawned without a shell, no re-tokenization), so a crafted `include:"--pre=evil"` is consumed as the glob *value*, never a flag. No fix needed.
- `nit · high-confidence · security · grep.ts:136 · ref:d79ea56 · file-state` — `-g '!.git'` (pushed after any include glob; last-match-wins) excludes `.git` despite `--hidden`. **Empirically confirmed**: `.git/config` secret was not surfaced. `--hidden` is parity with the old `grep -rn` (diff line 676), not a new exposure.

**Correctness / spec-compliance / perf — clean.** All 5 goals + the containment fix have implementing changes (grep.ts:22/157 engine swap; `extended` removed from GrepInput/parse/schema; `--hidden`+`-g '!.git'`; ci.yml matrix; `_rg-availability.ts`+error-path ordering at grep.ts:278; `_cwd-utils.ts:155` guard). rg exit codes (0/1/2) match the handler's `code===1`/`code===2` branches. The `_cwd-utils.ts` `!path.isAbsolute(rel)` guard is a genuine no-op on POSIX and fixes the Windows cross-drive escape. node_modules/dist no longer searched (`.gitignore` honored) is within stated intent. No scope creep.

**Low (optional polish — non-blocking):**
- `low · high-confidence · correctness · _rg-availability.ts:42-44 · ref:d79ea56 · file-state` — the rg-unavailable message advises "check pnpm.onlyBuiltDependencies", but `@vscode/ripgrep@1.18.0` has **no postinstall** (`scripts: null`; binary ships in per-platform optional-dep tarballs). **Verified**: dep works with no `onlyBuiltDependencies` entry; binary present + executable. Advice is misleading (harmless). *Fix:* reword to point at os/cpu mismatch or corrupted `node_modules`.
- `low · high-confidence · test-coverage · _cwd-utils.ts:155 · ref:d79ea56 · file-state` — the load-bearing Windows cross-drive guard has no test; the existing `rejects absolute path outside cwd` test uses `/etc`, which on POSIX yields a `..`-prefixed rel and never exercises the new `!path.isAbsolute(rel)` clause. *Fix:* unit test on `computeContainment` mocking `path.win32.relative` → `D:\etc`, assert `restricted: true`.

**Nit (informational):**
- `nit · api-compat · grep.ts:37,87-91 · ref:d79ea56 · diff-context` — removed `extended` param is breaking-by-design with **zero production importers** `[UNVERIFIED: no production importers]`; non-boolean `extended` is now silently dropped instead of throwing. Immaterial — no caller sends it. schemas.ts:161-185 description is accurate; no stale BRE/`extended` text survives in `src/` or `prompts/` (only regenerated `dist/`).
- `nit · ci.yml:81 · ref:d79ea56 · file-state` — job renamed `Test` → `Test (${{ matrix.os }})`. **Resolved, no operational impact**: `main` has no branch protection (GitHub API: `Branch not protected`, 404), and the only `select(.name == "Test")` reference (`docs/publish-workflow.md:142`) is a dormant doc snippet explicitly labeled *"Not added in this PR"* with zero live consumers. Would matter only if branch protection is later added referencing "Test", or that snippet is wired — worth a heads-up to the author, not a blocker.

## What was not checked
- **Refs read:** all citations read against branch HEAD `d79ea56` (= the review worktree) and verified at that ref; security behaviors were *executed* against the bundled `rg` on this macOS host.
- **Not executed on Windows:** the `path.win32` containment branch and the win/macOS matrix legs' pass/fail are inferred, not run — that's exactly the advisory signal this PR's own CI matrix exists to produce.
- **Stated intent:** PR #690 title+body — spec-compliance assessed.
- Branch-protection config was readable (no admin needed) and returned "not protected"; no other publish-pipeline consumers of the "Test" check name exist.

## Merge decision

**MERGE.** No blocking defects. Security is a net improvement — the `--` separator empirically closes a real ripgrep `--pre` command-execution vector that the engine swap would otherwise have opened, and `.git` stays excluded. All five stated goals plus the Windows containment fix are implemented and behave as claimed. The two `low` findings (misleading error-message wording; a missing Windows-only unit test) and the CI-rename `nit` are optional polish, not merge blockers.

Two notes for the author, both non-blocking: (1) the PR is intentionally a **draft** pending the win/macOS matrix signal + human review of the breaking grep contract — this review supplies the contract-change review; (2) flag the dormant `docs/publish-workflow.md:142` "Test" reference so a future belt-and-suspenders publish gate doesn't silently target the now-renamed job.

This skill is read-only — no files, PR, or CI state were modified. Stopping here.

---

griffinwork40 (commented):

Reviewed at ref `d79ea56` (PR head). No blocking, critical, or high-severity findings. The core change — swapping the grep tool from system `grep` to `@vscode/ripgrep`'s bundled `rgPath` — is correct, the intentional breaking contract change (BRE → ripgrep regex, `extended` param removed) is propagated to the tool schema, the Windows path-containment fix is sound and a POSIX no-op, and the argument-injection defense (the `--` end-of-options separator) is correctly implemented and tested. All remaining findings are low/nit. Note that `windows-latest` is red, but it is an **advisory** matrix leg (`continue-on-error`, ubuntu-latest remains the required gate) and the reds are pre-existing cross-platform items the PR body explicitly scopes out (test-harness `~/.afk` HOME override, `.mjs` dynamic import) — not regressions from this diff.

### What was verified (with an independent verification pass)
- **Argument injection (the crux): SAFE.** `args.push('--', pattern, path)` (grep.ts:145) fences both model-controlled positionals after `--`; a `--pre=<cmd>` pattern can no longer reach ripgrep's preprocessor flag. Verified byte-accurate at the ref, and that no model-controlled value is pushed after `--` (only three `args.push` calls exist; none follows line 145). Covered by tests: leading-dash pattern (`->`) and flag-shaped pattern (`--files`) in the `argument-injection / leading-dash safety` block.
- **Windows containment fix: CORRECT + POSIX-safe.** `_cwd-utils.ts:155` `!rel.startsWith('..') && !path.isAbsolute(rel)` rejects cross-drive escapes (`C:` root, `D:\etc` target) that the bare `startsWith('..')` test wrongly admitted. On POSIX, `path.relative()` between two absolute paths is never absolute, so the new clause is a pure no-op — no regression to the green Linux/macOS legs. Symlinks/UNC paths are pre-resolved via realpath before the comparison.
- **API-compat: `extended` removal is SAFE.** Zero production callers pass `extended` (`git grep 'extended: true' -- 'src/**'` = 0 matches at the ref; the tool is instantiated at handlers/index.ts:66 with no such arg). The tool + pattern schema descriptions (schemas.ts:161-174) are rewritten to warn that `| + ? ( ) { }` are now metacharacters and that `.env`/hidden files are searched. No stale "BRE / literal pipe" guidance survives anywhere in the tracked tree.
- **rg exit-code + glob-precedence contract: CORRECT.** `code===1`→no-match, `code===2`→error matches ripgrep. The trailing `-g '!.git'` after an `-g <include>` glob does not null out the include filter (ripgrep: an inclusion glob institutes the must-match requirement; a later exclude only removes; verified against ripgrep GUIDE.md).
- **`describeRgUnavailable` fires only on spawn failure**, not on rg's own `code===2` search error (it lives in `proc.on('error')`, the code-2 path lives in `proc.on('close')`) — no false "binary missing" on a real search error.

1. **nit · doc-accuracy · grep.ts (`--hidden` comment).** The comment implies `node_modules`/`dist` are "always" skipped, but ripgrep has no built-in `node_modules` exclusion — it skips them only when a `.gitignore` names them AND the search runs inside a git repo. In a gitignore-less, non-git directory rg will traverse `node_modules`. This is fully bounded by the existing 8 MB `HARD_CAP_BYTES` SIGKILL (worst case = a truncated result with a kill sentinel, never a hang/crash), and the old `grep -rn` was equally unbounded there — so this is a comment-precision nit, not a defect. Optional hardening: push `-g '!node_modules'` the same way `!.git` is pushed, or soften the comment to "gitignore-honored when present."
2. **low · security-hardening · grep.ts:131.** Model-controlled `include` is pushed as `-g <value>` *before* the `--` separator, so unlike `pattern`/`path` it is not positionally fenced. It is safe today — ripgrep's `-g` consumes exactly one value and `--pre` is inert without a `--pre-glob` that never appears in argv — but the safety rests on ripgrep internals rather than handler-level fencing. Optional: fuse the token as `-g=${include}` so a leading-dash value can never be re-parsed as a flag.
3. **low · test-coverage · grep.test.ts.** No test exercises a flag-shaped `include` value (e.g. `include: '--pre=...'`); the `--` tests cover `pattern` only. A case asserting a flag-shaped `include` yields a clean no-match would lock the `include` arg path.
4. **low · test-coverage · _cwd-utils.test.ts.** No dedicated `path.win32` cross-drive unit test for the containment fix. It is covered indirectly — the six filesystem handler suites' "rejects absolute path outside context.cwd" tests exercise the shared helper on the `windows-latest` leg — but a direct `computeContainment` unit test with a `C:`-root/`D:\etc`-target pair would pin the regression without depending on the CI runner.
5. **low · test-coverage · grep.test.ts.** No test covers the `include` + trailing `-g '!.git'` interaction (that the exclude does not null out the include filter). The behavior is correct per ripgrep docs but untested.

### What was NOT checked
- Test suites were not executed (read-only review); test *presence and assertions* were verified at the ref, not pass/fail. The PR's "tsc clean + suites green on macOS" claim is unverified here.
- The advisory `windows-latest`/`macos-latest` matrix legs were read in `ci.yml` (matrix + `continue-on-error` present) but no CI run was inspected beyond the status rollup.
- `@vscode/ripgrep` version currency / CVEs not assessed.
- Spec-compliance was assessed against the PR title + body (stated intent); all named deliverables have an implementing change and no unmentioned scope creep was found.

_Reviewed against branch HEAD `d79ea56`; blocking/high candidates were independently re-derived in a verification pass — one initial medium finding (a `pnpm.onlyBuiltDependencies` gap) was dropped after verifying `@vscode/ripgrep@1.18.0` ships prebuilt platform binaries as optional-deps with no postinstall build script, so the whitelist omission blocks nothing._

Issue #436, opened 2026-07-05
glob: traverses node_modules/.git with no ignore list, ignores .gitignore; hand-rolled ** matcher untested


griffinwork40:
## Summary

The `glob` handler's `collectMatches` **always recurses into every directory** with no ignore list, so it descends `node_modules/`, `.git/`, `dist/`, etc. Two consequences:

1. **Perf** — an unconstrained walk of a large `node_modules` is slow.
2. **Result pollution** — the 500-result cap fills top-down alphabetically, so a pattern like `**/*.js` can be saturated by `node_modules` junk before any real `src` hit, returning `[results capped at 500 entries]` of noise.

There is **no `.gitignore` respect** at all — this diverges sharply from what users expect of a glob tool (ripgrep/fd/git-aware globbers skip VCS-ignored files and dotdirs by default).

Separately, the `**` matcher is hand-rolled (split-on-`**` greedy regex builder) and has **no test** for multi-segment patterns like `a/**/b/**/c`.

## Evidence

- `src/agent/tools/handlers/glob.ts:80,103-109,212` — always-recurse, no ignore list, 500 cap
- `src/agent/tools/handlers/glob.ts:23-63,48-55` — hand-rolled matcher; greedy middle-segment advance
- `src/agent/tools/handlers/glob.test.ts` — cwd behavior only; no multi-`**` correctness coverage

## Why it matters

This bites in normal daily use: a repo-root glob is both slow and returns capped junk instead of the intended matches.

## Proposed direction

- Skip `node_modules/`, `.git/`, `dist/` and dotdirs by default (or add an `ignore` param); optionally honor `.gitignore`.
- Add multi-`**` correctness tests; consider replacing the hand-rolled matcher with `picomatch`/`fast-glob` (behavior-compatible, well-tested).

Severity: MED (quality / correctness). Labels: bug + enhancement.

Found via a read-only tool/subagent/worktree audit.

PR #692, opened 2026-07-24
fix(glob): make ** match zero path segments (root-level files)

griffinwork40:
## The bug

The `glob` tool's `**` **never collapsed to zero path segments**, so the most common recursive patterns silently dropped every match at the search root:

| pattern | before (buggy) | after (this PR) |
|---|---|---|
| `**/*.ts` | misses a root-level `foo.ts` | matches root **and** nested |
| `src/**/*.ts` | misses `src/foo.ts` | matches files directly in `src/` too |
| `**/*.json` | **"No files matched"** when the only match is at root | matches the root-level file |

This is the dangerous failure class — **silently incomplete results**, not a loud error. An agent using `glob` to check for a root-level file (`**/vite.config.ts`, `**/package.json`) could wrongly conclude it's absent.

## Root cause

`matchesGlobPattern` split the pattern on `**` and matched the neighboring parts literally — **including the `/` adjacent to the `**`**. So `**/*.ts` split into `['', '/*.ts']`, and the trailing part compiled to `/[^/]*\.ts$`, which *requires* a slash. A root-level file like `top.ts` has no slash, so it could never match.

## The fix

Replace the fragile split-on-`**` matcher with a **single-pass glob→RegExp compiler** (`globToRegExp`):

- A globstar followed by a separator compiles to an **optional** prefix `(?:.*/)?` — that's the zero-segment case.
- A trailing `**` compiles to `.*`.
- Everything else is escaped to a literal; a non-segment `**` (e.g. `a**b`) degrades to a single-segment `*`.
- Compiled **once per glob call** (was recompiling multiple sub-regexes per entry).

This also removes the old greedy, no-backtrack multi-`**` matching (a regex backtracks naturally), so patterns like `a/**/b/**/c.txt` are more robust.

## Behavior change

Recursive patterns now **correctly include root-level matches**. This is strictly *more*-correct — it matches ripgrep / bash `globstar` / the `glob` npm package. **No pattern that matched before stops matching.** The only new outputs are correct matches that were previously (wrongly) dropped.

## Verification

- `tsc --noEmit` clean.
- Full `src/agent/tools/` suite: **65 files / 1859 tests green**, including the pre-existing multi-`**` correctness tests (`glob.test.ts`) — so no regression.
- **+5 regression tests** covering exactly the failing patterns above (`glob.test.ts`).

## Notes

- Independent of #690 (the grep→ripgrep Windows PR) — branched off `main`.
- **Not a Windows issue**: this bug was identical on all platforms; `glob` is pure-JS and spawns nothing. This PR is a plain correctness fix.
- Out of scope (pre-existing, lower-stakes `glob` limitations, not touched here): results are returned in `readdir` order with no sort before the 500-cap; directory symlinks aren't traversed; brace `{a,b}` / char-class `[abc]` aren't supported (documented as "basic").

### DIFF
--- a/src/agent/tools/handlers/glob.test.ts
+++ b/src/agent/tools/handlers/glob.test.ts
@@ -451,3 +451,77 @@ describe('glob handler — multi-** matcher correctness (#436)', () => {
     expect(result.content).not.toContain('a/x/y/z/c.txt');
   });
 });
+
+describe('glob handler — globstar collapses to zero segments (root-level match)', () => {
+  let tempDir: string;
+
+  beforeEach(async () => {
+    tempDir = await fs.mkdtemp(path.join(os.tmpdir(), 'glob-globstar-zero-'));
+    // Root-level files + nested files. The root-level ones are the regression:
+    // the previous matcher required a '/' adjacent to '**', so `**` never
+    // collapsed to zero segments and silently dropped every match at the root.
+    await fs.mkdir(path.join(tempDir, 'pkg', 'inner'), { recursive: true });
+    await fs.writeFile(path.join(tempDir, 'top.ts'), '');
+    await fs.writeFile(path.join(tempDir, 'data.json'), '');
+    await fs.writeFile(path.join(tempDir, 'pkg', 'mid.ts'), '');
+    await fs.writeFile(path.join(tempDir, 'pkg', 'inner', 'leaf.ts'), '');
+  });
+
+  afterEach(async () => {
+    await fs.rm(tempDir, { recursive: true, force: true });
+  });
+
+  it('**/*.ts matches a ROOT-LEVEL file (zero leading segments) as well as nested', async () => {
+    const result = await globHandler(
+      { pattern: '**/*.ts', path: tempDir },
+      new AbortController().signal,
+    );
+    expect(result.isError).toBeUndefined();
+    const lines = result.content.split('\n');
+    expect(lines).toContain('top.ts'); // regression: previously dropped
+    expect(lines).toContain('pkg/mid.ts');
+    expect(lines).toContain('pkg/inner/leaf.ts');
+  });
+
+  it('**/*.json returns the sole root-level match (was "No files matched")', async () => {
+    const result = await globHandler(
+      { pattern: '**/*.json', path: tempDir },
+      new AbortController().signal,
+    );
+    expect(result.isError).toBeUndefined();
+    const lines = result.content.split('\n');
+    expect(lines).toContain('data.json');
+    expect(result.content).not.toContain('No files matched');
+  });
+
+  it('dir/**/*.ts matches a file directly in dir (zero intermediate segments)', async () => {
+    const result = await globHandler(
+      { pattern: 'pkg/**/*.ts', path: tempDir },
+      new AbortController().signal,
+    );
+    expect(result.isError).toBeUndefined();
+    const lines = result.content.split('\n');
+    expect(lines).toContain('pkg/mid.ts'); // regression: zero intermediate dirs
+    expect(lines).toContain('pkg/inner/leaf.ts');
+    expect(lines).not.toContain('top.ts'); // outside pkg/, must not match
+  });
+
+  it('**/name matches a root-level file by name', async () => {
+    const result = await globHandler(
+      { pattern: '**/top.ts', path: tempDir },
+      new AbortController().signal,
+    );
+    expect(result.isError).toBeUndefined();
+    const lines = result.content.split('\n');
+    expect(lines).toContain('top.ts');
+  });
+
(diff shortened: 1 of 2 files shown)


griffinwork40:
Empirically confirmed against the verbatim compiler: every intent case works (`top.ts`, `data.json`, `src/foo.ts`, nested), over-match guard holds, `pkg/**/*.ts` correctly excludes root, multi-`**` #436 behavior preserved, and the degenerate `a/**b` → `^a/[^/]*b$` no-longer-crosses-`/` behavior change is real (old matched `a/x/b`, new does not). No critical/high findings anywhere → `/shadow-verify` not triggered. Synthesizing.

---

# Code Review — PR #692

**Target:** `fix(glob): make ** match zero path segments (root-level files)` · branch `afk/glob-globstar-fix → main` · reviewed ref `cdd89b5` (working tree clean = reviewed ref)
**Change type:** bugfix · **Regime:** light · **Files:** 2 (`glob.ts` +57/−53, `glob.test.ts` +74/−0) · **184 lines**
**Stated intent:** PR #692 title+body — spec-compliance **assessed**.

## Verdict: ✅ MERGE

A correct, well-scoped, well-tested fix for a silent-incompleteness bug. The split-on-`**` matcher is replaced with a single-pass `globToRegExp` compiler where `**/` → optional `(?:.*/)?` (the zero-segment collapse). Independently verified: **33 tests pass** (5 new regression tests included), the walker is untouched by the change, api-compat is clean, and the compiler behaves exactly as documented. No blocking findings.

## Findings (all non-blocking; sorted by severity)

**1 · medium · confidence high · spec-compliance · `glob.ts:69-84` · ref:cdd89b5**
The PR's guarantee *"No pattern that matched before stops matching"* is overbroad. For a non-segment globstar adjacent to a separator on one side only (e.g. `a/**b`), the old matcher matched `a/x/b`; the new compiler degrades `**`→`[^/]*`, producing `^a/[^/]*b$`, which no longer crosses `/`.
- Evidence (empirical, verbatim compiler): `a/**b` → `/^a\/[^/]*b$/` → `test("a/x/b")` = **false** (old = true).
- **This is a correctness improvement, not a regression** — bash `globstar`, ripgrep, and the `glob` npm package all treat a non-segment `**` as a plain `*`. The code is right; only the claim is inaccurate.
- Suggestion: reword the guarantee to *"no **well-formed** pattern that matched before stops matching; a malformed non-segment `**` is corrected toward standard glob semantics."* (PR description only — no code change.)

**2 · low · confidence medium · test-coverage · `glob.test.ts:455-527` · ref:cdd89b5**
The one intentional behavior change (non-segment `**` degrade) is the highest-value regression guard and is not pinned by a test. Suggestion: add one case asserting `a/**b` matches `a/xyzb` but **not** `a/x/b`, locking the degrade against future regressions.

**3 · low · confidence high · security(ReDoS) · `glob.ts:80` · ref:cdd89b5**
Adjacent globstars (`**/**/…`) compile to consecutive optional greedy groups `(?:.*/)?(?:.*/)?`, a super-linear-backtracking shape. **Negligible under this tool's threat model**: the regex is tested per filesystem entry against a single, path-depth-bounded relative path (no long adversarial strings reach `.test()`), and the 500-result cap bounds invocations. No action required for merge; if defense-in-depth is later wanted, collapse consecutive globstar groups to a single `(?:.*/)?` at compile time.

## Clean dimensions
- **correctness** — `globToRegExp` char-walk traced and run: escaping set complete (backslash pre-normalized to `/`), no `/g` flag so `matcher.test()` reuse is stateless, boundary logic and `j++` separator-absorb correct.
- **regression risk** — walker `collectMatches`/`walk` is pattern-independent (recurses all non-pruned dirs, regex only filters), so the swap cannot alter traversal; #436 multi-`**` tests pass under the new regex (verified empirically).
- **api-compat** — removed `matchesGlobPattern`/`convertGlobPartToRegex` had **zero importers** in `src/` (module-private, never exported); `createGlobHandler`/`globHandler` exports unchanged and still consumed by `index.ts`.

## What was not checked
- **Ref read:** Wave 1 + orchestrator read against working tree = branch HEAD `cdd89b5` (clean); all citations verified at that ref. Citation-verification wave was a no-op — zero blocking/high/critical findings to check.
- **Not run:** `tsc --noEmit` (type-safety inferred from test-file imports resolving + green `glob.test.ts`; the full-tools-suite green claim was not independently re-run — only `glob.test.ts`).
- **Not audited:** every `glob` *call site* across the repo for reliance on the old degenerate `a/**b` matching — no importers of the removed internals exist, and reliance on a malformed-glob edge case is implausible, but call-string usage was not exhaustively grepped.
- **Spec-compliance:** assessed against PR #692 title+body.

This skill is read-only; I am reporting findings only and taking no further action.

---

Issue #555, opened 2026-07-13
Follow-ups from #548 review — per-slot chatgpt-oauth provider (non-blocking)


griffinwork40:
Non-blocking follow-ups surfaced by the review of #548 (first-class per-slot `chatgpt-oauth` provider). None gate merge; captured here so they aren't lost.

**Already addressed in #548** (for context): the Codex **P1** — stale `forceChatgptOAuth` across a `/model` switch — is fixed, and the new `chatgpt-oauth` provider value is now documented in the `models.*` config surface (`settable-keys.ts`).

### Deferred items

- [x] **[MEDIUM] `afk provider auth diagnose` is not slot-aware** — `src/cli/commands/provider.ts:36`
  `buildProviderAuthDiagnose` calls `resolveOpenAIAuth(explicitConfigKey, deps)` without the `forceChatgptOAuth` arg, so the auth-debugging command reports the **global** resolution and never shows `chatgpt-oauth` / `no-usable-auth-forced-chatgpt-oauth` for a configured slot. A user whose `chatgpt-oauth` slot fails will run diagnose and get a misleading global answer. No runtime impact (the real query path at `query.ts:1056` is correct). The command is deliberately global-only today ("config-key plumbing intentionally not wired here yet").
  *Fix:* add a `--model`/`--slot` option, resolve the slot binding, and thread `forceChatgptOAuth` (and the per-slot key) through to `resolveOpenAIAuth`; add a test.
  *Resolved by #1389 (2026-08-31).*

- [x] **[LOW] Expired ChatGPT token is returned as usable (no expiry gate)** — `src/agent/providers/openai-compatible/auth.ts` (Tier 0)
  Tier 0 copies `parsed.expiresAt` into the resolution but never compares it to now, so an expired token yields an opaque 401 from OpenAI at request time instead of the actionable "re-run `codex`" diagnostic. **Not a regression** — faithfully mirrors the existing flag-gated path, which shares the gap.
  *Fix:* gate expiry in **both** paths (`expiresAt <= now → { apiKey: null, source: 'no-usable-auth-…' }`) so `formatExpiry`'s "EXPIRED" hint fires as a hard error.
  *Resolved by #1394 (2026-08-31).*

- [x] **[LOW] Standalone one-shot path ignores the forced flag** — `src/agent/providers/openai-compatible/oneshot.ts:99`
[...]

PR #693, opened 2026-07-24
fix(subagent): coerce Claude-family child models under openai-compatible routing (#652)

### LINKED ISSUE #652: Skill subagents default to 'sonnet' and hard-error under forced ChatGPT-OAuth routing
## Summary
Forked skill subagents default their model via `skill.model ?? ctx.defaultSubagentModel ?? ctx.defaultModel ?? 'sonnet'` (`src/agent/tools/skill-executor/fork-dispatch.ts:121`, `:206`). When a session is forced onto the ChatGPT-OAuth (Responses) backend — e.g. `AFK_PROVIDER=openai-compatible` with a chatgpt-oauth credential — the provider hard-errors on any Claude-family model (`src/agent/providers/openai-compatible/query.ts:745-756`). Net effect: the skill subagent errors out and never runs.

## Severity / trigger
High impact **when triggered**, but the trigger is narrow: it only fires when a Claude-named child is actually routed to `openai-compatible` + chatgpt-oauth. Under normal config a `sonnet` child routes to `anthropic-direct` and is unaffected. It fails as a clear error, not a hang.

## Options
1. When the parent routes to chatgpt-oauth, default skill/subagent children to a provider-appropriate model rather than the Anthropic `sonnet` alias.
2. Improve the error to name the offending skill and point at the config fix.

## Notes
- Related: #555 (chatgpt-oauth follow-ups).
- Found via a skills × provider-compatibility audit. Static analysis only — not yet reproduced against a live chatgpt-oauth session.

griffinwork40:
## Summary
Fixes #652. Forked skill/agent/compose children that resolve a Claude-family model — via the `?? 'sonnet'` default chain, or by inheriting a Claude-pinned auto-dispatched agent (`git-investigator` → `sonnet`, `Explore` → `haiku`) — hard-error under a `chatgpt-oauth` / `AFK_PROVIDER=openai-compatible` session. That routing sends the child to the ChatGPT/Codex backend, which serves no Claude model and fails fast at `openai-compatible/query.ts` (`isClaudeFamilyModel` guard). The child never runs, and for the vendored/byte-locked agent pins the operator has **no recourse** (can't pass a model to an auto-dispatched agent, can't edit the pin).

The issue as filed cites two lines, but this is a **class** bug: the same `?? 'sonnet'` default appears at 6 child-dispatch sites (`fork-dispatch.ts:121,206`, `compose-executor.ts:650`, `subagent/child-config.ts:167`, `background-branch.ts:72`, `foreground-promotion.ts:207`) plus the two explicit `builtins.ts` pins. All converge on one choke point.

## Fix (issue Option 1, generalized + Option 2)
**Coerce at the single choke point.** `SubagentManager.forkSubagent` is documented in-code as "the single choke point through which the agent-tool, skill, and compose paths all create their child session." New helper `coerceCrossProviderChildModel` substitutes such a child onto the **parent's own model** (the concrete gpt target the session already uses — no hardcoded gpt literal, respects the no-`src/cli`-import layering).

Predicate — `isClaudeFamilyModel(child) && providerForModel(child) === 'openai-compatible'` — is **false-positive-safe**:
- `providerForModel` routes `claude-*` / `opus` / `sonnet` / `haiku` **and** local Anthropic-shim `local-*` ids to `anthropic-direct` (Tier 2) *unless* a Tier-1 force overrides. So a normal Anthropic session and a local shim are **never** coerced — only the genuinely-broken "Claude id forced onto an OpenAI endpoint" combination.
- Skips when the parent isn't a usable substitute (undefined / itself Claude-family) so the actionable provider error still fires.

The coerced model is threaded into **every** model-derived `childConfig` field — `model`, `apiKey` (credential anti-leak), `baseUrl`, the **read-only-phase provider** (`buildPhaseRestrictedProvider` re-derives the provider from the model — miss this and a read-only fork re-triggers the bug), and the trace event. The override is written to stderr (not silent).

**Option 2 (backstop):** enriched the `query.ts` hard-error to point at `AFK_DEFAULT_SUBAGENT_MODEL` for auto-dispatched agents you can't pass a model to.

## Testing
- `src/agent/subagent/child-model-fallback.test.ts` — 11 unit tests (force/no-force, local-shim protection, gpt passthrough, unusable-parent guards).
- `src/agent/subagent.test.ts` — 4 integration tests at the fork boundary (coercion under force, no-op without force, gpt unchanged, no-parentModel no-op; asserts the stderr warning).
- `pnpm lint`, `pnpm test` (116 tests across the 4 relevant files), and `audit:sdk:check` / `audit:env:check` / `scan:env:check` all pass.

## Notes / scope
- Top-level sessions are intentionally **not** coerced (a user who explicitly runs a Claude model under chatgpt-oauth gets the clear error — correct). Only *forked* children, which inherit/default their model, are coerced.
- Does not address the per-slot `chatgpt-oauth` variant where a `sonnet` child routes to `anthropic-direct` and dies for lack of an Anthropic credential (a different failure path / #555 territory) — the enriched error covers that case.

### DIFF
--- a/src/agent/providers/openai-compatible/query.ts
+++ b/src/agent/providers/openai-compatible/query.ts
@@ -781,8 +781,10 @@ export class OpenAICompatibleQuery implements ProviderQuery {
           error: new Error(
             `Model "${this.currentModel}" can't run on a ChatGPT subscription — the ChatGPT/Codex ` +
               `backend only supports OpenAI gpt-5.x models. This usually means a subagent or skill ` +
-              `requested a Claude model. Pass a gpt-5.x model to it (e.g. model: "gpt-5.5"), or run ` +
-              `it on a provider configured with the matching API key.`,
+              `requested a Claude model. Pass a gpt-5.x model to it (e.g. model: "gpt-5.5"); for an ` +
+              `auto-dispatched agent you can't pass a model to (e.g. git-investigator), set ` +
+              `AFK_DEFAULT_SUBAGENT_MODEL to a gpt-5.x id. Or run it on a provider configured with ` +
+              `the matching API key.`,
           ),
         };
         return null;
--- a/src/agent/subagent.test.ts
+++ b/src/agent/subagent.test.ts
@@ -485,6 +485,79 @@ describe('SubagentManager', () => {
     });
   });
 
+  describe('cross-provider child model coercion (#652)', () => {
+    // providerForModel reads AFK_PROVIDER transitively; scrub to a known
+    // baseline per-test and restore the ambient value after.
+    const savedProvider = process.env['AFK_PROVIDER'];
+    beforeEach(() => {
+      delete process.env['AFK_PROVIDER'];
+    });
+    afterEach(() => {
+      if (savedProvider === undefined) delete process.env['AFK_PROVIDER'];
+      else process.env['AFK_PROVIDER'] = savedProvider;
+    });
+
+    it('coerces a Claude-family child onto the parent gpt model under AFK_PROVIDER=openai-compatible', async () => {
+      shared.lastConfig = null;
+      process.env['AFK_PROVIDER'] = 'openai-compatible';
+      const warn = vi.spyOn(process.stderr, 'write').mockReturnValue(true);
+      let warned = '';
+      try {
+        const mgr = new SubagentManager({ parentModel: 'gpt-5.5' });
+        await mgr.forkSubagent({
+          parent: { sessionId: 'p' },
+          config: { model: 'sonnet' },
+          idPrefix: 'coerce-check',
+        });
+        // Capture BEFORE mockRestore (restore resets mock.calls).
+        warned = warn.mock.calls.map((c) => String(c[0])).join('');
+      } finally {
+        warn.mockRestore();
+      }
+      // The child that would have hard-errored on the ChatGPT backend now runs
+      // the parent's gpt model instead.
+      expect((shared.lastConfig as Record<string, unknown>)['model']).toBe('gpt-5.5');
+      // The override is observable (not silent).
+      expect(warned).toContain('sonnet');
+      expect(warned).toContain('gpt-5.5');
+    });
+
+    it('does not coerce in a normal Anthropic session (no force)', async () => {
+      shared.lastConfig = null;
+      const mgr = new SubagentManager({ parentModel: 'gpt-5.5' });
+      await mgr.forkSubagent({
+        parent: { sessionId: 'p' },
+        config: { model: 'sonnet' },
+        idPrefix: 'no-coerce-check',
+      });
+      expect((shared.lastConfig as Record<string, unknown>)['model']).toBe('sonnet');
+    });
+
+    it('leaves a gpt child unchanged under force', async () => {
+      shared.lastConfig = null;
+      process.env['AFK_PROVIDER'] = 'openai-compatible';
+      const mgr = new SubagentManager({ parentModel: 'gpt-5.5' });
+      await mgr.forkSubagent({
+        parent: { sessionId: 'p' },
+        config: { model: 'gpt-5.5' },
+        idPrefix: 'gpt-child',
+      });
+      expect((shared.lastConfig as Record<string, unknown>)['model']).toBe('gpt-5.5');
+    });
+
+    it('does not coerce when the manager has no parentModel (no substitute)', async () => {
+      shared.lastConfig = null;
+      process.env['AFK_PROVIDER'] = 'openai-compatible';
+      const mgr = new SubagentManager();
+      await mgr.forkSubagent({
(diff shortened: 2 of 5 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

Reviewed at ref `38b5516` (PR head). No blocking, critical, or high-severity defects. The fix — coercing forked Claude-family children onto the parent's own model at the `SubagentManager.forkSubagent` choke point under `openai-compatible` routing — is correct, the coercion predicate is genuinely false-positive-safe, and all five intent-named `childConfig` threadings are present and verified. One `medium` and two `low` findings are follow-up test/observability polish, not merge blockers.

### What was verified (credential-threading is the crux — re-derived independently, twice)

- **No partial coercion, no credential leak.** When a Claude-family child is coerced onto the parent's gpt model, all five model-derived fields re-derive from the coerced `effectiveChildModel`, not the stale Claude id: `model` (subagent.ts:800), `apiKey` via `applyManagerApiKeyFallback({ childModel: effectiveChildModel })` (subagent.ts:832), `baseUrl` via `providerForModel(effectiveChildModel)` (subagent.ts:842), the read-only-phase provider `buildPhaseRestrictedProvider('read-only', effectiveChildModel)` (subagent.ts:950), and the trace event (subagent.ts:1047-1049).
- **The credential gate is fail-closed both directions.** `applyManagerApiKeyFallback` (child-credential.ts:112) returns `parentApiKey` only when `providerForModel(childModel) === effectiveParent`; post-coercion child=gpt (`openai-compatible`) matches parent (`openai-compatible`), so the OpenAI key lands on a gpt model — provider-consistent. Any provider mismatch resolves the key to `undefined` (child-credential.ts:102,111). This closes a latent pre-fix edge where a forced Claude child kept `model: sonnet` while its apiKey derived from `providerForModel(sonnet)` under force.
- **The predicate is false-positive-safe.** `isClaudeFamilyModel(child) && providerForModel(child) === 'openai-compatible'` — `local-*`/`claude-*`/`opus|sonnet|haiku` aliases lock to `anthropic-direct` via Tier-2 (providers/index.ts:180-187) unless a Tier-1 `AFK_PROVIDER` force fires (index.ts:156-157). A normal Anthropic session and a local Anthropic shim are never coerced. Guard ordering (child-invalidity before parent-usability) and the skip conditions (undefined parent; parent itself Claude-family) are correct (child-model-fallback.ts:421-434).
- **No secret in the stderr warning** (subagent.ts:776-780 interpolates only two model-id strings). **api-compat:** `forkSubagent`'s signature is byte-identical to the parent ref; the new helper is purely additive with a single production call site; the query.ts change edits only the error-message string.

### Findings (all medium/low/nit — none blocking)

1. **medium · test-coverage · subagent.test.ts / child-model-fallback.test.ts.** The read-only-phase-provider threading has **no test** — no test dispatches a fork with `phaseRole: 'read-only'` under force and asserts the *coerced* model reached `buildPhaseRestrictedProvider`. Confirmed-absent by an independent grep pass (`git grep -n 'buildPhaseRestrictedProvider' -- <both test files>` → no matches; the only cross-file hit is a comment at nesting.test.ts:24). This is the field the code's own comment (subagent.ts:764-766) flags as "most critically" the one that re-triggers the bug if unthreaded — correct today, but unpinned against regression. *Fix: add one integration test forking with `phaseRole: 'read-only'` under force, asserting the restricted provider was built on the parent gpt model.*
2. **low · test-coverage · subagent.test.ts:119-141.** The `gpt-child` and `no-parent-model` no-op integration tests assert `model` is unchanged but do not assert `warned === ''`, so a spurious coercion warning on a no-op path would go uncaught. *Fix: add `expect(warned).toBe('')` to one no-op case.*
3. **low · observability · subagent.ts:1044-1049.** The lifecycle `started` trace event records only the post-coercion (gpt) model, with no structured `coercedFrom` — a trace reader sees the substitute model with no machine-readable signal that a coercion occurred (only unstructured stderr). *Fix: add an optional `coercedFrom` field to the lifecycle event payload.*

**nit:** the "11 unit tests" figure is 8 `it`/`it.each` blocks that vitest expands to 11 instances — real coverage, informational only.

### What was NOT checked

- Test suites were **not executed** (read-only review); test *presence and assertions* were verified at the ref, not pass/fail. The PR's "lint + 116 tests + audit gates pass" claim is unverified here.
- Citations read against branch HEAD `38b5516` via `git show`; every finding is sourced from that ref, not the working tree. The one claim that would be critical if wrong (credential threading) was independently re-derived twice and holds.
- Not exercised: `resolveBinding` slot-alias expansion for exotic tier rebindings; the per-slot `chatgpt-oauth` variant the PR explicitly scopes out (#555 territory).
- Spec-compliance was assessed against the PR title + body (stated intent); all named deliverables have an implementing change and no scope creep was found.

PR #697, opened 2026-07-24
fix(readonly-bash): anchor git verbs to subcommand position + wrapper/indirection handling (#577)

### LINKED ISSUE #577: readonly-bash: subcommand anchoring + wrapper/indirection handling (follow-up to #574)
## Summary

Follow-up to #574 (Tier 1). Remaining `readonly-bash.ts` classifier gaps that Tier 1 (quote-stripping) does **not** address, all rooted in the classifier matching regexes against a string rather than a parsed command line.

### Tier 2 — over-restriction: bare argument that collides with a git verb
A bare, unquoted pickaxe/message-search value that happens to be a git verb is still misclassified as mutating:
- `git log -S push`, `git log -G merge`, `git log --grep revert`, `git help push`, `git push --help` → **BLOCKED** (false positive).
- Cause: `GIT_MUTATING_VERBS` (`src/agent/tools/readonly-bash.ts:58-59`) matches the verb *anywhere* after `git` in the segment, not just in subcommand position.
- Fix: anchor mutating git verbs to **subcommand position** (`git` + optional global flags like `-c x=y` / `-C dir` / `--no-pager`, then the verb), so `git log …` args don't match while `git push` still does. Needs a careful test matrix (`git -c x=y commit` and `git -C /p push` must still block; `git log -S push`, `git shortlog`, `git help push` must pass).

### Tier 3 — under-restriction: wrapper + variable indirection (findings 4/5)
The read-only cage misses well-known mutation forms via wrappers/indirection:
- `find . | xargs rm`, `env rm -rf x`, `command rm x`, `busybox rm` — a wrapper hides the verb from `FS_MUTATING`.
- `g=git; $g push`, `$g reset --hard` — variable indirection defeats the `\bgit\b` anchor.
- `| ksh` / `| python` / `| perl` — `PIPE_TO_SHELL` (`readonly-bash.ts:123`) only matches `sh|bash|zsh|dash`.
- `INTERPRETER_WRITE_API` (`readonly-bash.ts:188-189`) misses `os.system`/`subprocess`/`Deno.writeTextFile`/`pathlib.write_text`.
- **Quoted flag-value** (distinct mechanism — *not* wrapper/indirection): `curl -X "POST" https://x` classifies **read-only** because the method (the mutation signal) is quoted and `CURL_WRITE_METHOD` (`readonly-bash.ts`) now runs against the quote-stripped view, so `"POST"` is blanked before it matches. Pre-existing — #574's raw regex never matched a quoted method either (`-X\s+POST` needs an adjacency the quote breaks). Narrow: bodyless writes only — any `-d`/`--data`/`-F` payload is still caught via `CURL_DATA` regardless of quoting. Same parsed-command fix applies: command `curl` + parsed flag `-X POST` matches, while `grep -rn "curl -X POST" .` (command `grep`) stays read-only. Flagged by Codex on #574.

**Why this matters (threat framing):** `git-investigator` is a shipped read-only agent that is granted `Bash` (gated only by this best-effort classifier — `git-investigator.ts:12`, `builtins.ts:95`) and it reads **attacker-controllable** content (commit messages, diffs from a malicious PR). A prompt-injected git-investigator escaping the cage via `g=git; $g push` is a genuine vector, so the module's "well-behaved agent" assumption is optimistic here.

## Suggested approach

Both tiers are cleanly solved by a small command-line tokenizer: split on `; | && || & ( )`, take each segment's command-position token, resolve one level of `VAR=val; $VAR`, unwrap no-op wrappers (`env`/`xargs`/`command`/`sudo`/`nice`/`time`/`timeout`/`busybox`), then match rules against the resolved command + args. Also widen `PIPE_TO_SHELL` and `INTERPRETER_WRITE_API`. This subsumes Tier 2 (args aren't command-position), Tier 3 wrapper/indirection (wrapped/indirected verbs resolve to command-position), and the quoted flag-value gap (the method is parsed as an arg, so matching is quoting-independent).

## Acceptance criteria

- [ ] `git log -S push` / `--grep push` / `git help push` classified read-only; `git push` / `git -c x=y commit` still mutating.
- [ ] `find | xargs rm`, `env rm -rf`, `g=git; $g push`, `| python`, `os.system(...)` classified mutating.
- [ ] `curl -X "POST" https://x` (quoted method) classified mutating; `grep -rn "curl -X POST" .` still read-only.
- [ ] Existing 185 tests in `readonly-bash.test.ts` still pass; new coverage for each case above.

Refs: 2026-07-13 audit — over-restriction O1 residual + under-restriction findings 4/5 + curl quoted-method value (Codex on #574). Builds on #574.

griffinwork40:
Closes #577.

## Problem
`classifyBashCommand` matched git/curl verb regexes against the whole command string. A string match cannot tell a verb in **command position** from the same word in an **argument**, causing both defects #577 tracks:

- **Over-restriction (Tier 2):** a bare pickaxe/grep/help arg that collides with a git verb was blocked — `git log -S push`, `git log -G merge`, `git log --grep revert`, `git help push`, `git push --help`. This hurt `git-investigator`, whose core job is git-history search.
- **Under-restriction (Tier 3):** mutations reached via a wrapper (`env`/`xargs`/`command`/`sudo`/`busybox`), a shell variable (`g=git; $g push`), a pipe-to-interpreter (`| python`), or a quoted flag value (`curl -X "POST"`) slipped through.

## Fix
Replace the six *anywhere-in-segment* git-verb regexes with a small **command-line tokenizer** (Pass 1c) that:
- splits pipeline/chaining segments (`; | & ( )`), quote-aware;
- resolves one level of `VAR=val; $VAR` indirection (shared across segments);
- unwraps no-op wrappers (`env`/`xargs`/`command`/`sudo`/`nice`/`time`/`timeout`/`nohup`/`stdbuf`/`busybox`/`eval`);
- recurses into `$(…)`/backtick substitutions;

…then judges git by **subcommand position** (`gitSegmentReason`) and curl by parsed argv (`curlSegmentMutating`, catches the quoted `-X "POST"`). Also widens `PIPE_TO_SHELL` (more shells + piped interpreters `python`/`perl`/…) and `INTERPRETER_WRITE_API` (`os.system`/`os.popen`/`subprocess.`/`Deno.*`).

`git config` stays regex-matched — its mutation signal (a quoted `<value>`) must be read on the raw string.

## Conservative-by-construction (security)
Positional anchoring is more *precise* but was initially *less conservative* than the old blunt regex — an adversarial differential caught that it regressed ~12 git-mutation obfuscations the old code caught (`eval git push`, `{ git push; }`, `for…do git push`, `env g=git $g push`, `\git push`, `timeout -k 5 10 git push`). A `hiddenGitMutation` backstop (+ `eval`/env/`-k` handling) restores them, so the classifier is **strictly ≥ the old regex** for the cage. A broad differential over ~55 realistic commands shows the *only* behavioral deltas vs. the pre-change classifier are the intended Tier-2 loosenings + tighter curl/indirection blocking — **no new over-blocks on recon, no new bypasses.**

### Documented residual (asserted in a test)
`$(echo git) push` — a command *substitution* producing the command name — is not statically resolvable without executing it, so (like `sh -c "git push"`) it's out of scope per the module's stated *best-effort recon cage, not hostile-process sandbox* threat model. Layer-1 `write_file`/`edit_file` strip remains the real boundary.

## Acceptance criteria
- [x] `git log -S push` / `--grep push` / `git help push` read-only; `git push` / `git -c x=y commit` mutating.
- [x] `find | xargs rm`, `env rm -rf`, `g=git; $g push`, `| python`, `os.system(...)` mutating.
- [x] `curl -X "POST" https://x` mutating; `grep -rn "curl -X POST" .` read-only.
- [x] Existing 185 tests pass; new coverage per case.

## Tests
- `readonly-bash.test.ts`: **257 pass** (185 existing + 72 new — acceptance cases, obfuscation backstop, asserted residual).
- Consumers green: `dispatcher` (104), `plan-mode-gate` (16), `nesting` (28), `agents/parser` (13), `plan-mode-addendum` (12).
- `tsc --noEmit` (strict): clean.
- `classifyBashCommand` signature unchanged — zero API-compat impact.

🤖 Implemented + adversarially verified via agent-afk.

### DIFF
--- a/src/agent/tools/readonly-bash.test.ts
+++ b/src/agent/tools/readonly-bash.test.ts
@@ -544,6 +544,204 @@ describe('classifyBashCommand', () => {
     }
   });
 
+  // ── #577 Tier 2: git verbs anchored to SUBCOMMAND position ────────────────
+  // A git verb appearing as a bare ARG (pickaxe value / --grep / help topic) or
+  // behind a `--help` flag must NOT be misread as an invocation — the
+  // over-restriction the tokenizer fixes. Real invocations (incl. the
+  // global-flag-before-verb forms) must still block.
+  describe('#577 Tier 2: git subcommand anchoring', () => {
+    const allowed: string[] = [
+      'git log -S push',
+      'git log -G merge',
+      'git log --grep revert',
+      'git log -S push --oneline',
+      'git log --grep=push',
+      'git help push',
+      'git help commit',
+      'git push --help', // --help ⇒ shows docs, does not push
+      'git commit --help',
+      'git branch --help',
+      // NOTE: `git rm --help` / `git mv --help` are (harmlessly) still blocked —
+      // the bare `rm`/`mv` token trips FS_MUTATING regardless of the git help
+      // flag. A recon agent never needs those; not worth suppressing.
+      'git shortlog -sn',
+      'git log --format="%h %s (push)"', // verb inside a quoted format arg
+      'git -C /dir log -S push',
+      'git show HEAD:file',
+    ];
+    for (const cmd of allowed) {
+      it(`allows: ${cmd}`, () => {
+        expect(classifyBashCommand(cmd).mutating, `expected "${cmd}" read-only`).toBe(false);
+      });
+    }
+
+    const blocked: Array<[string, string]> = [
+      ['git push', 'bare git push'],
+      ['git -c x=y commit -m m', 'git -c x=y commit'],
+      ['git -C /p push', 'git -C /p push'],
+      ['git --no-pager reset --hard', 'git --no-pager reset'],
+      ['git log --oneline && git push', 'push after &&'],
+    ];
+    for (const [cmd, label] of blocked) {
+      it(`blocks: ${label} (${cmd})`, () => {
+        expect(classifyBashCommand(cmd).mutating, `expected "${cmd}" mutating`).toBe(true);
+        expect(classifyBashCommand(cmd).reason).toBeTruthy();
+      });
+    }
+  });
+
+  // ── #577 Tier 3: no-op wrappers + one-level variable indirection ──────────
+  describe('#577 Tier 3: wrapper + variable indirection resolution', () => {
+    const blocked: Array<[string, string]> = [
+      ['g=git; $g push', 'var-indirected git push'],
+      ['g=git; $g reset --hard', 'var-indirected git reset'],
+      ['g=git && ${g} commit -m x', 'braced var-indirected git commit'],
+      ['env git push', 'env-wrapped git push'],
+      ['sudo git reset --hard', 'sudo-wrapped git reset'],
+      ['command git push origin main', 'command-wrapped git push'],
+      ['/usr/bin/env git push', 'absolute-path env git push'],
+      ['find . | xargs rm', 'xargs rm'],
+      ['find . -print0 | xargs -0 rm -rf', 'xargs -0 rm -rf'],
+      ['env rm -rf x', 'env rm'],
+      ['env FOO=bar rm x', 'env with assignment then rm'],
+      ['command rm x', 'command rm'],
+      ['busybox rm x', 'busybox rm'],
+    ];
+    for (const [cmd, label] of blocked) {
+      it(`blocks: ${label} (${cmd})`, () => {
+        expect(classifyBashCommand(cmd).mutating, `expected "${cmd}" mutating`).toBe(true);
+      });
+    }
+
+    // A wrapper / indirection that resolves to a READ command stays allowed.
+    const allowed: string[] = [
+      'find . -name "*.ts" | xargs grep foo',
+      'env git status',
+      'command git log --oneline',
+      'g=git; $g status',
+      'sudo cat /etc/hosts',
+    ];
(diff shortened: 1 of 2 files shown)


griffinwork40:
Thanks @codex — both P2 findings were valid and are fixed in `28938e1`:

**1. env option operands** — `env -u NAME` / `-C DIR` (+ `--unset`/`--chdir`/`-S`) now consume their separate operand in the unwrap loop, so the operand is no longer mistaken for the wrapped command. `env -u FOO git stash` resolves `git` to command position and blocks.

**2. xargs option operands** — `--max-args`/`-n` / `--max-lines`/`-L` / `--max-procs`/`-P` / `-I` / `-s` / `-d` / `-a` / `-E` (short + long) now consume their operand too. `find . | xargs --max-args 1 git tag -d v1` blocks.

Plus a defense-in-depth fix for the root gap you identified — the hidden-git backstop now re-runs the full subcommand classifier on tokens after any bare `git`, so **conditional** mutations (stash / tag -d / branch -D / remote add / worktree remove) reached via a wrapper or compound form are caught, not just always-mutating verbs.

+13 regression tests (270 classifier + consumers green, tsc clean).

PR #698, opened 2026-07-24
fix(cli): colorize aliased slash commands (e.g. /quit) as brand, not dim

griffinwork40:
## Problem

`/quit` (and every other slash-command **alias**) rendered dim in the REPL input buffer instead of brand orange like real commands/skills. Reported: "notice how /quit isn't orange like the skills?"

## Root cause

`colorizeInputBuffer` paints a `/token` brand-orange when `SlashRegistryView.has(name)` is true, dim otherwise. Three surfaces build that view — `reader.ts`, `input-surface.ts`, `_lib/stream-renderer.ts` — and all three used the **same** canonical-only predicate:

```ts
has: (name) => listSlashCommands().some((c) => c.name === `/${name}`)
```

`listSlashCommands()` is the registry's `list()`, which returns **canonical commands only** — aliases live in a separate map. So `has('quit')` was `false` and the token fell through to the unknown-token (dim / `meta`) tone. The registry's own `has()` (`registry.ts`) already resolves aliases; the adapters just bypassed it.

Because the buggy predicate was copy-pasted into three call sites, the bug landed in triplicate.

### Every affected alias

| Alias | Canonical | Source |
|-------|-----------|--------|
| `/quit` | `/exit` | `commands/core.ts` |
| `/ctx` | `/tokens` | `commands/info.ts` |
| `/edit` | `/editor` | `commands/editor.ts` |
| `/thinking-ui` | `/thinking` | `commands/thinking.ts` |
| `/branch` | `/fork` | `commands/fork.ts` |
| `/t` | `/transcript` | `commands/transcript.ts` |
| `/builtin-skills` | `/skills` | `plugin-skills/listing.ts` |

## Fix

Add `createSlashRegistryView()` in `slash/registry.ts` as the single adapter factory, delegating membership to the registry's alias-aware `has()` (re-adding the leading `/` the colorizer strips). All three call sites now share it instead of re-deriving membership from `list()`. Memo semantics are preserved: `version` stays wired to `registryVersion`, and each surface still constructs the view once (stable object identity for the colorizer's per-keystroke memo).

## Tests

New `src/cli/slash/registry.test.ts` (7 tests):
- canonical / alias / unknown membership through the factory;
- a test documenting *why* the old `list()`-based predicate missed aliases (`list()` excludes the alias);
- `version()` wiring to `registryVersion` (the colorizer's memo-invalidation token);
- an end-to-end `colorizeInputBuffer('/quit', view)` assertion that it equals `palette.brand('/quit')` and is **not** `palette.meta('/quit')` — the exact user-visible symptom.

Verified locally: full suite **698 files / 12974 tests passed**, 0 failed; `tsc --noEmit` clean.

### DIFF
--- a/src/cli/_lib/stream-renderer.ts
+++ b/src/cli/_lib/stream-renderer.ts
@@ -35,7 +35,7 @@ import { isPlainOutputRequested } from '../../config/env.js';
 import type { IHistoryRing } from '../input/types.js';
 import type { AutocompleteState } from '../input/autocomplete-state.js';
 import { colorizeInputBuffer, type SlashRegistryView } from '../input-highlight.js';
-import { list as listSlashCommands, registryVersion } from '../slash/registry.js';
+import { createSlashRegistryView } from '../slash/registry.js';
 import { ToolLane } from '../commands/interactive/tool-lane.js';
 import { ThinkingLane } from '../commands/interactive/thinking-lane.js';
 import { StreamingMarkdownRenderer } from '../markdown-stream.js';
@@ -400,19 +400,13 @@ export class StreamRenderer {
         compositor.setOnCancel(this.onCancel);
       }
     } else {
-      // Live-registry adapter — queried fresh on every render so plugins that
-      // register slash commands mid-session colorize correctly without a
-      // restart. Mirrors the closure used by `readWithAutocompleteTty` at
-      // src/cli/input/reader.ts:103-105. `listSlashCommands()` returns names
-      // prefixed with `/`; the highlighter passes the bare name, so we
-      // re-add the prefix in the predicate.
-      const slashRegistryView: SlashRegistryView = {
-        has: (name) => listSlashCommands().some((c) => c.name === `/${name}`),
-        // Enables the `colorizeInputBuffer` single-entry memo across repaints;
-        // the version bumps on any mid-session command registration so the
-        // cache never serves a stale-colored buffer.
-        version: registryVersion,
-      };
+      // Live-registry adapter for the slash colorizer, built via
+      // `createSlashRegistryView()`. Membership delegates to the registry's
+      // alias-aware `has()` (queried fresh per render) so plugins that register
+      // slash commands mid-session — and aliased commands like `/quit` —
+      // colorize correctly without a restart. Mirrors the surface used by
+      // `InputSurface` and `readWithAutocompleteTty`.
+      const slashRegistryView: SlashRegistryView = createSlashRegistryView();
       compositor = new TerminalCompositor({
         stdout: process.stdout,
         stdin: process.stdin,
--- a/src/cli/input/input-surface.ts
+++ b/src/cli/input/input-surface.ts
@@ -50,7 +50,7 @@ import {
 } from '../terminal-compositor.js';
 import { colorizeInputBuffer, type SlashRegistryView } from '../input-highlight.js';
 import { detectCaptureMode, detectCaretBlink, detectReducedMotion, detectGoblinSpinner } from '../_lib/capture-mode.js';
-import { list as listSlashCommands, registryVersion } from '../slash/registry.js';
+import { createSlashRegistryView } from '../slash/registry.js';
 import { formatSubmittedEcho } from './echo.js';
 import { describeAttachmentSummary } from './attachments.js';
 import { commitBlockAbove } from '../_lib/commit-block.js';
@@ -272,21 +272,17 @@ export class InputSurface {
   private pendingReadResolve: ((value: ReadWithAutocompleteResult) => void) | null = null;
 
   /**
-   * Live-registry adapter — queried fresh via `listSlashCommands()` on
-   * every call so plugins that register slash commands mid-session
-   * colorize correctly. Held as a class field (not an `armCompositor`
-   * local) because two call sites need it: the compositor's
-   * `formatInputBuffer` live-typing hook AND the `readLine` submit-echo
-   * which must colorize `payload.text` before committing to scrollback
-   * — without that second call, the submitted slash command appears as
-   * plain text in history (parity bug vs. `reader.ts:329-330`).
+   * Live-registry adapter for the slash colorizer, built via
+   * `createSlashRegistryView()`. Membership delegates to the registry's
+   * alias-aware `has()` (queried fresh per call) so plugins that register
+   * slash commands mid-session — and aliased commands like `/quit` — colorize
+   * correctly. Held as a class field (not an `armCompositor` local) because
+   * two call sites need it: the compositor's `formatInputBuffer` live-typing
+   * hook AND the `readLine` submit-echo which must colorize `payload.text`
+   * before committing to scrollback — without that second call, the submitted
+   * slash command appears as plain text in history (parity bug vs. `reader.ts`).
    */
-  private readonly slashRegistryView: SlashRegistryView = {
-    has: (name) => listSlashCommands().some((c) => c.name === `/${name}`),
-    // Enables the `colorizeInputBuffer` memo (per-keystroke repaints share a
-    // buffer) while still invalidating on any mid-session command hot-swap.
-    version: registryVersion,
-  };
+  private readonly slashRegistryView: SlashRegistryView = createSlashRegistryView();
 
   constructor(opts: InputSurfaceOptions) {
     this.rl = opts.rl;
--- a/src/cli/input/reader.ts
+++ b/src/cli/input/reader.ts
@@ -16,7 +16,7 @@
 import { emitKeypressEventsImmediateEscape } from './emit-keypress.js';
 import * as ansiEscapes from 'ansi-escapes';
 import stringWidth from 'string-width';
-import { list as listSlashCommands, aliasEntries, registryVersion } from '../slash/registry.js';
+import { list as listSlashCommands, aliasEntries, createSlashRegistryView } from '../slash/registry.js';
 import { stripAnsi } from '../display.js';
 import { acquireStdinClaim } from './stdin-claim.js';
 import { isPrintableGrapheme } from './printable.js';
@@ -115,16 +115,11 @@ export async function readWithAutocompleteTty(
       let repaintPending = false;
       const PASTE_WINDOW_MS = 8;
 
-      // Adapter for the slash highlighter — registered command names from
-      // `listSlashCommands()` include the leading `/`, so we strip it before
-      // membership-check so the highlighter (which passes the bare name)
-      // matches correctly.
-      const slashRegistryView: SlashRegistryView = {
-        has: (name) => listSlashCommands().some((c) => c.name === `/${name}`),
-        // Lets `colorizeInputBuffer` memoize per-keystroke repaints and drop
-        // the cache the instant a plugin/skill command is hot-swapped in.
-        version: registryVersion,
-      };
+      // Adapter for the slash highlighter. Delegates membership to the
+      // registry's alias-aware `has()` (via `createSlashRegistryView`) so
+      // aliased commands (e.g. `/quit`) colorize brand, not dim — and so the
+      // per-keystroke memo still invalidates on any mid-session hot-swap.
+      const slashRegistryView: SlashRegistryView = createSlashRegistryView();
 
       /**
        * Fully redraw the input line (and optional dropdown) at whatever row the
--- a/src/cli/slash/registry.test.ts
+++ b/src/cli/slash/registry.test.ts
@@ -1,0 +1,107 @@
+/**
+ * Tests for src/cli/slash/registry.ts — focused on `createSlashRegistryView`,
+ * the adapter the input-buffer colorizer (`colorizeInputBuffer`) uses to decide
+ * brand (known command → orange) vs meta (unknown token → dim) coloring.
+ *
+ * Regression context: aliased commands (e.g. `/quit` → `/exit`, `/t` →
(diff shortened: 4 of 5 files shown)


griffinwork40 (commented):

## Merge decision: **MERGE**

Reviewed at ref `104e5b164251c2ec01169c580f4fbcdc7d2e0b79` (light regime — hotfix, 5 files, +161/−41). The diff achieves its stated intent, and the mechanism was independently re-derived rather than taken on the PR body's word.

### Mechanism verified (the fix is not inert)

The load-bearing question was whether `has('/quit')` actually resolves — if aliases were keyed *without* the leading slash, the factory would be a no-op. Checked end to end at the reviewed ref:

- `src/cli/slash/commands/core.ts:28` → `aliases: ['/quit']`, and `registry.ts:41` stores it verbatim (`aliases.set(alias, cmd.name)`), so alias keys carry the `/`.
- `registry.ts:76-77` → `has()` is `commands.has(nameOrAlias) || aliases.has(nameOrAlias)`.
- `input-highlight.ts:182-183` → the colorizer passes the **bare** token (`token.slice(1)`) and paints `palette.meta` on a `false`.
- `registry.ts:102` → `has: (name) => has(`/${name}`)` re-adds the prefix.

The composition is exact. All 7 aliases in the PR table were confirmed at their cited sources.

**No 4th surface left behind:** `git grep -n "listSlashCommands().some" <ref> -- src` returns empty, and the only production `has:` adapter remaining in `src` is `registry.ts:102`.

**No runtime import cycle:** `registry.ts:12` is `import type` (erased at compile time), and `input-highlight.ts` imports only `chalk` and `./palette.js` — it never reaches back into the registry.

**Memo semantics preserved:** the memo keys on view identity (`memoRegistry === registry`, `input-highlight.ts:169`). The new `stream-renderer.ts:409` call sits inside `arm()`, guarded at `:370` by `if (this.disposed || !this.isTTY || this.compositor) return;`, so it constructs at most once per renderer — not per render. `input-surface.ts:285` is a class field. Membership also gets marginally cheaper: O(1) double `Map.has` replaces an O(n) `list().some()`.

### Findings

**low · test-coverage · `src/cli/slash/registry.test.ts:47-104` · ref:104e5b1 · diff-context**
Coverage pins the *factory* but never drives an alias through any of the three converted call sites, so re-inlining a canonical-only predicate at `reader.ts:122`, `input-surface.ts:285`, or `stream-renderer.ts:409` — the exact bug this PR fixes — would fail no test. Absence grounded: `git grep -n "createSlashRegistryView" <ref> -- 'src/**/*.test.ts'` returns 8 hits, all in `registry.test.ts`; `input-surface.test.ts:515` does exercise `formatInputBuffer` coloring but registers a command with no `aliases:`.
*Suggestion:* extend `input-surface.test.ts:515` to register `makeCmd('/exit', ['/quit'])` and assert the submit-echo buffer equals `palette.brand('/quit')`.

**nit · correctness · `src/cli/input/trigger.ts:60,318` · ref:104e5b1 · file-state — pre-existing, out of scope**
Flag completion resolves the typed token against canonical names only (`listSlashCommands().find((c) => c.name === `/${command}`)`), so `/thinking-ui --<tab>` yields no flag candidates while `/thinking --<tab>` does — the same canonical-only root cause on a surface this PR does not touch. *Suggestion:* route those two lookups through the alias-aware `lookup()` (`registry.ts:135-139`) in a follow-up. [low confidence — verify with runtime context]

**nit · correctness · `src/cli/multi-line-reader.ts:33` · ref:104e5b1 · file-state — pre-existing, no production callers**
`buildCompleter`'s slash branch derives candidates from canonical names with no `aliasEntries()` union, unlike `reader.ts:800-801` and `suggest.ts:327-329`. `git grep -n "buildCompleter" <ref> -- src` → 12 hits, 11 in `multi-line-reader.test.ts` plus the definition, so there is no user-visible effect. [UNVERIFIED: no production importers]

### CI status — the red Windows leg is not a blocker

`Test (windows-latest)` failed on this run, but it is **advisory, not a merge gate**: `.github/workflows/ci.yml:94` sets `continue-on-error: ${{ matrix.os != 'ubuntu-latest' }}`, which is why run `30127627368` reports overall `success`. The failure is in `tests/agent/memory/memory-loader.test.ts` (4 assertions returning `undefined`), a file this diff does not touch — the diff contains zero `memory` references. Recent `main` runs (`30128726777`, `30127658645`) show the same Windows failure with overall `success`, confirming it is pre-existing and unrelated. The required `ubuntu-latest`, `macos-latest`, `Lint & Build`, PTY-scrollback, docs, and bundle-smoke legs are all green.

### Per-dimension status

- **security** — no issues found; read `registry.ts` + `input-highlight.ts` at 104e5b1. A color-tone predicate with no auth, secret, path, or shell surface.
- **api-compat** — no issues found. Purely additive: one `import type` and one new export. `has()`, `list()`, `registryVersion()`, `aliasEntries()`, `lookup()` are byte-identical, so `list()`-semantics consumers (autocomplete, suggest) are unperturbed.
- **correctness** — two nits, both pre-existing and outside the diff's scope; the fix itself is correct.
- **spec-compliance** — no issues found. Exactly the 3 named surfaces converted; all 7 claimed aliases confirmed.
- **test-coverage** — one low finding (above). Note the new file cannot fail against the old code — it tests an API that did not previously exist — so its value is forward-pinning, not regression-reproduction. `chalk.level` is captured and restored correctly (`:82-89`).
- **perf-observability** — no issues found; all 5 `colorizeInputBuffer` call sites pass one long-lived view, so the identity-keyed memo still hits.

### What was not checked

Read against ref `104e5b164251c2ec01169c580f4fbcdc7d2e0b79` throughout; every citation came from `git show <ref>:<path>`, never a working tree. Not executed: the author's "698 files / 12974 tests passed" and "`tsc --noEmit` clean" claims (static inspection only). Not checked: actual terminal rendering of brand vs. meta at `chalk.level < 3`; memo behavior when an `InputSurface` view and a `StreamRenderer` own-compositor view alternate within one session (a pre-existing property, not a regression introduced here). Stated intent: PR #698 title + body — spec-compliance assessed.

griffinwork40:
---

# Code Review — PR #698

**`fix(cli): colorize aliased slash commands (e.g. /quit) as brand, not dim`**
`afk/diagnose-command-coloring` → `main` · reviewed ref **`104e5b16`** · +161/−41 across 5 files · regime: **light** (hotfix, 202 lines)

## Summary

The diff does exactly what it says. It replaces three copy-pasted canonical-only membership predicates with one factory delegating to the registry's alias-aware `has()`. I independently re-derived the four load-bearing claims (completeness, no collateral behavior change, memo parity, test-wiring gap) — three confirm clean, one surfaces a `low` coverage gap.

**Verified correctness argument.** `commands` is unconditionally keyed by `cmd.name` (`registry.ts:36` — `commands.set(cmd.name, cmd)`), `list()` is an unfiltered `[...commands.values()].sort()`, and `register()` throws when an alias collides with a command name (`registry.ts:38-40`), so the `commands` and `aliases` key-spaces are disjoint by construction. Therefore for every **non-alias** token `has('/'+n)` ≡ the old `list().some(c => c.name === '/'+n)`, and the two differ on **exactly** the alias set — which is the intended fix, with zero collateral coloring change.

---

## Findings by dimension

### security
`no issues found` — read `registry.ts`, `input-highlight.ts`, and all three surfaces at `104e5b16`. The token reaching `has()` is constrained by `SLASH_TOKEN_RE` (`input-highlight.ts:52`) to `/[A-Za-z][\w:-]*`, and the new code performs a `Map` key lookup, not a regex/eval/query. No auth, secret, or injection surface touched.

### api-compat
`no issues found` — `createSlashRegistryView` is purely additive; no export was removed from `registry.ts`. Import removals are local to the three surfaces. Reachability pre-check: `reader.ts` retains `listSlashCommands` and still uses it at `reader.ts:800`, so `noUnusedLocals` is satisfied; `stream-renderer.ts` and `input-surface.ts` have 0 residual uses of both removed imports.

### correctness
`no issues found` — see the verified correctness argument above. Construction sites read at `reader.ts:122`, `input-surface.ts:285`, `stream-renderer.ts:409`; all five production `colorizeInputBuffer` call sites (`stream-renderer.ts:424`, `input-surface.ts:339`, `:528`, `reader.ts:160`, `:346`) consume a factory-built view. No fourth surface exists.

### spec-compliance
`no issues found` — intent met, no scope creep. All seven aliases claimed in the PR body are registered at the ref: `/quit` (`commands/core.ts:28`), `/ctx` (`commands/info.ts:158`), `/edit` (`commands/editor.ts:18`), `/thinking-ui` (`commands/thinking.ts:36`), `/branch` (`commands/fork.ts:69`), `/t` (`commands/transcript.ts:119`), `/builtin-skills` (`plugin-skills/listing.ts:349`). The diff contains nothing beyond the factory, three call-site swaps, and tests.

### perf-observability
`no issues found` — and a net **improvement**. View-construction frequency is unchanged versus the merge-base `687d8735` in all three files (per-`readWithAutocompleteTty` call, per-`InputSurface` instance, per-`arm()` respectively), so the identity-keyed memo in `input-highlight.ts:118-152` is not degraded. Separately, the per-token membership check drops from O(n log n) — the old predicate called `list()`, which allocated an array and ran a `localeCompare` sort, on **every token, every repaint** — to an O(1) `Map.has`.

### test-coverage

```
low · high confidence · test-coverage · src/cli/slash/registry.test.ts:36-106 · ref:104e5b16 · citation-type:file-state
finding: The new tests pin the factory in isolation but nothing pins the three surfaces to it, so reverting any single surface to a local list()-based predicate reintroduces the exact reported bug with a fully green suite.
evidence:
  git grep -n "slashRegistryView" 104e5b16 -- '*.test.ts'   → 0 matches
  git grep -n "palette.brand" 104e5b16 -- 'src/cli/input/*.test.ts' 'src/cli/_lib/*.test.ts' → 0 matches
  input-surface.test.ts:533 registers '/test-cmd' — canonical, no aliases → predicate-agnostic
suggestion: Extend the existing surface test at input-surface.test.ts:515 to register a command WITH an alias and assert the submit-echo colorizes the alias form as palette.brand, so the wiring — not just the factory — is pinned.
```

---

## Nits

- **Layering inversion** — `registry.ts:12` imports the `SlashRegistryView` type from `input-highlight.ts`, whose own doc at `input-highlight.ts:31-32` states the interface exists to keep the highlighter "decoupled from the registry's full surface." Type-only (erased, no runtime cycle), and `input-highlight.ts` imports nothing from the registry, so this is a design remark only. Hoisting the type to a shared `types.ts` would keep the highlighter a leaf.
- **Duplicate test-file scope** — the new `src/cli/slash/registry.test.ts:2` and the pre-existing `src/cli/slash-registry.test.ts:2` both declare themselves "Tests for src/cli/slash/registry.ts"; the latter already covers alias resolution via `lookup`. Two files, one module.
- **Comment drift introduced indirectly** — `input-surface.test.ts:528` still reads "Clear the registry so `list()` returns only our test command"; that surface no longer consults `list()`.
- **Weaker-than-possible wiring assertion** — `registry.test.ts:72-73` asserts `view.version!() === registryVersion()`, which holds for any function returning the same number; `expect(view.version).toBe(registryVersion)` would pin the reference itself.

---

## What was not checked

- **Ref discipline:** Wave 1 agents and the independent verifier read against branch HEAD `104e5b164251c2ec01169c580f4fbcdc7d2e0b79` (present in the local clone); every citation above was re-read at that ref via `git show`. Pre-PR comparisons used merge-base `687d8735`.
- **Stated intent:** PR #698 title + body — **spec-compliance was assessed** against it.
- **Process deviation to disclose:** the Wave 1 `research-agent` dispatch **failed** (stream cut off mid-turn — the read-only agent type has no bash, nested a `git-investigator`, and the nested stream died). Fallback: I gathered the mechanical grep/read evidence inline and dispatched one independent adversarial `general-purpose` verifier to re-derive the four load-bearing claims from scratch. That verifier hit its tool-iteration cap but had already emitted all four verdicts plus its grep log; I independently re-derived its single load-bearing evidence item (`input-surface.test.ts:515-545`) myself before relying on it. Coverage is equivalent to the light-regime single-agent wave; the multi-lens redundancy of a full-regime wave was not run (not warranted at 202 lines).
- **Not run:** the test suite and `tsc --noEmit` — this skill is read-only and does not execute builds. The author's claim of 12974 passing tests and a clean typecheck is **unverified here**; CI is the gate.
- **Not reviewed:** runtime terminal behavior (no PTY render was exercised), the `*.pty.test.ts` harness, and the Telegram surface (untouched by this diff).

---

# ✅ MERGE

No blocking findings. One `low` test-coverage gap (the fix is verified correct, but not regression-pinned at the surface level) and four nits — all non-blocking and addressable in follow-up.

---

**Done**
- Reviewed PR #698 across all six dimensions at ref `104e5b16`, with an independent adversarial re-derivation of the four load-bearing claims.
- Evidence: this report; diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/a4423653-a113-49c9-9204-13e2c9486cca/pr-698.diff`; every citation re-read at `104e5b164251c2ec01169c580f4fbcdc7d2e0b79` via `git show`; session witness trace at `~/.afk/state/witness/<session>/trace.jsonl`.
- World unchanged — read-only review; no files written, no commits, no PR comments posted, working tree still clean on `main @ 411210f`.
- Deferred: the `low` test-coverage finding and four nits are reported only, not fixed (out of scope for `/review`); the Wave 1 research-agent dispatch failure is documented above rather than retried.

---

PR #694, opened 2026-07-24
feat(models): add Claude Opus 5 as the opus/large default

griffinwork40:
## What

Anthropic released **Claude Opus 5** (`claude-opus-5`) on 2026-07-24. This wires it in as a drop-in replacement for the retired `claude-opus-4-8`, so `opus` / `opus_1m` and the `large` tier default now resolve to Opus 5.

Verified against first-party sources: `@anthropic-ai/sdk@0.115.0` CHANGELOG ("add claude-opus-5 model") and the Anthropic models/overview doc (`claude-opus-5`, dateless pinned snapshot, $5/$25 per MTok, 1M context, 128k max output, adaptive thinking, knowledge cutoff May 2026).

## Changes

- **`model-slots.ts`** — `CLAUDE_OPUS_ID → 'claude-opus-5'`. Propagates by construction to the `opus`/`opus_1m` aliases, the `large` tier default, `MODEL_MAP`, and the `/model` picker hint.
- **`model-limits.ts`** — native **1M context** + **128k max output**; retired the `claude-opus-4-8` output entry (comment breadcrumb per the existing convention); added a **200k autocompact budget** so base `opus` keeps compacting early (mirrors `claude-sonnet-5`), while `opus_1m` opts into the full 1M window. Without the budget, base `opus` would have jumped from compacting at ~200k to ~1M — a silent cost/latency regression.
- **`resolve-params.ts`** — two version-gated predicates that would otherwise miss Opus 5:
  - `requiresAdaptiveThinking` — Opus 5 is adaptive-thinking-only, so `{type:'enabled'}` is auto-promoted to adaptive (avoids a 400).
  - `resolveEffort` — Opus 5 defaults to `max` effort, matching `opus-4-8` / `sonnet-5`.
- **`types.ts`** — `MODEL_PRICING` $5 / $25 per MTok (cost telemetry).
- **`query.ts`** — refreshed the `STARTER_MODELS` dropdown to Opus 5.
- **Docs + tests** — README, `docs/model-slots.md`, website config pages; updated model-id / limit / routing assertions across 6 test files.

## Design decision

Drop-in replacement (matches the repo's historical `opus` retirement pattern). **Opus 4.8 remains reachable by its raw wire id `claude-opus-4-8`** but is no longer a first-class alias — keeps the change minimal and trivially reversible (revert `CLAUDE_OPUS_ID`).

## Verification

- `pnpm lint` clean.
- Full `vitest` suite green (**12,967 tests**).
[...]

PR #699, opened 2026-07-24
test(models): pin Opus 5 effort/thinking/pricing behavior and correct stale opus comments

griffinwork40:
Follow-up to #694 (Claude Opus 5 as the `opus`/`large` default).

#694 merged with one reviewer finding — codex P2 *"Restore limits for active Opus 4.8 pins"* — which was fixed in-branch (`9b22cb3`) before merge and is in `main`. These are the **remaining review items**. None changed runtime behavior in #694; each leaves a load-bearing behavior unpinned or a comment actively wrong.

## Coverage — the widened regexes had no test for the id they were widened for

`resolve-effort.test.ts` had **zero** `opus-5` references, despite #694 editing both `resolveEffort` and `requiresAdaptiveThinking` specifically to cover it.

- `resolveEffort('claude-opus-5') === 'max'` — guards the allowlist refactor from `sonnet-5` to `(opus|sonnet)-(4-[678]|5)`.
- `resolveThinkingParam` promotes `enabled` → `adaptive` on `claude-opus-5`. Opus 5's card is *"Extended thinking: No / Adaptive thinking: Yes"*, so a manual `{type:'enabled'}` reaching the wire is an API rejection — worth pinning since the failure is a hard 400.
- `deriveCallCostUsd` golden values: `$5`/`$25` per MTok → `$30.00` for 1M+1M, and `$6.75` for 1M cache-read + 1M cache-write. Previously covered only by the generic `cost > 0` loop; a wrong rate fails silently and persists into saved cost reports.
- `autoCompactLimitFor('claude-opus-5')` by literal wire id, matching the existing `claude-sonnet-5` case (`requestedModel` can be either side of the alias boundary).

## Comments #694 made false

- **`model-limits.test.ts`** asserted `autoCompactLimitFor('opus') === 200_000` under the heading *"leaves opus / opus_1m / haiku / fable at their full window (no budget)"*. Post-#694 that passes for the **opposite** reason: `opus` now *has* a `MODEL_AUTOCOMPACT_BUDGET` entry and 200k is a **cap on a 1M window**. A future reader could "fix" the comment by changing the expectation to `1_000_000`. Split the opus cases out so the number and its rationale agree, and made explicit that haiku's 200k is its *real window* while opus/sonnet's 200k is a *reduced budget*.
- **`session-state.ts`** still cited `claude-opus-4-8` as the shared wire id and *"1M vs 200k"* as the `opus`/`opus_1m` difference. Opus 5 is natively 1M, so the surviving difference is the auto-compaction limit, not the window.

## Staleness

- `env.ts` examples advertised `claude-opus-4-5` (`AFK_MODEL`, three generations stale) and `claude-opus-4-8` (`AFK_MODEL_LARGE`); regenerated `docs/env-registry.{json,md}` via `pnpm scan:env`.
- `STARTER_MODELS` offered `claude-haiku-4-5-20250929` while canonical `CLAUDE_HAIKU_ID` is `claude-haiku-4-5-20251001`. Safe to switch: `MODEL_PRICING` carries **both** dates at identical rates and `types.test.ts` spot-checks both, so cost reporting is unaffected.

## Deliberately excluded

- **Unanchored-regex NIT**: a hypothetical `claude-opus-50` would false-match `opus-5`. No current collision, and the same latent issue predates #694 for `sonnet-5`. Not worth touching a live request-param regex for.
- **`effort-2025-11-24` beta-header comments** in `types.ts`/`auth.ts`: Anthropic's Opus 5 notes state no beta header is required for the effort ladder, so these may describe a superseded requirement. That needs live API verification, not a doc edit.

## Verification

| Gate | Result |
|---|---|
| `pnpm test` | 698 files, **12,983 passed** / 14 skipped |
| `pnpm audit:sdk:check` | pass |
| `pnpm audit:env:check` | pass |
| `pnpm scan:env:check` | pass (registry in sync after regen) |

Test-only + comments + doc examples — no runtime code paths changed. Diff: 8 files, +72/−15.


griffinwork40:
CI note: `Test (windows-latest)` is red, but **not from this PR** — it fails identically on `main`.

Failing-file lists are byte-identical between this PR's Windows job and `main` @ `b9973e77` ([run 30127658645](https://github.com/griffinwork40/agent-afk/actions/runs/30127658645), job `Test (windows-latest)` = `failure` while the run's overall conclusion is `success`):

```
tests/agent/memory/memory-loader.test.ts
tests/agent/memory/memory-store.test.ts
tests/agent/memory/memory-tools.test.ts
tests/audit-deps.test.ts
tests/copy-bundled-plugins.test.ts
tests/postinstall.test.ts
tests/state-isolation.test.ts
tests/telegram/manager.test.ts
tests/telegram/message-handler-drain.test.ts
tests/telegram/setup-wizard.test.ts
```

None of those files are touched by this diff (8 files: 3 test files under `src/`, `query.ts` starter list, `session-state.ts` comment, `env.ts` examples, and the regenerated `docs/env-registry.*`).

Per `.github/workflows/ci.yml:83-94`, the windows/macos matrix legs are **advisory** — `continue-on-error: ${{ matrix.os != 'ubuntu-latest' }}` — added by #690 for signal during the in-progress Windows-compat grind, with `ubuntu-latest` as the required gate. The comment there says to flip them to required once the suite is green on those runners.

Every blocking check passes: `Test (ubuntu-latest)`, `Test (macos-latest)`, `Test (real PTY scrollback)`, `Lint & Build`, `Docs Site`, `Publish Bundle Smoke`. Left the Windows compat work alone deliberately — it is a separate ~10-file workstream, not something to fold into a test/comment follow-up.

PR #700, opened 2026-07-24
fix(compaction): support history compaction on the openai-compatible responses wire (#653)

### LINKED ISSUE #653: No history compaction in openai-compatible Responses / ChatGPT-OAuth wire mode
## Summary
`compact()` returns `unsupported-wire-mode` in the openai-compatible **Responses** wire mode (`src/agent/providers/openai-compatible/query.ts:878-889`), so turn-boundary auto-compaction also no-ops there. A long-running or iterative skill subagent (e.g. `mint`, `refactor`) on the ChatGPT-OAuth backend cannot shed context and risks hitting the context-window limit.

## Severity
Medium for long/iterative skill subagents; low for short ones. Chat Completions mode is unaffected (compaction works there).

## Options
- Implement compaction for the Responses wire, or
- Document it as a known limitation and surface a clearer warning as the window fills.

## Notes
- Related: #555 (chatgpt-oauth follow-ups).
- Found via a skills × provider-compatibility audit. Static analysis; severity inferred, not measured against real transcript lengths.

griffinwork40:
## What & why

Closes #653.

History compaction previously **no-oped** on the openai-compatible **Responses** wire (the ChatGPT-OAuth backend and the `AFK_OPENAI_USE_RESPONSES` opt-in). `compactHistory()` bailed early with `unsupported-wire-mode` because the only summarizer — `oneShotChatCompletion` — speaks Chat Completions, a call the Responses backend rejects. So long/iterative skill subagents (`mint`, `refactor`) running on the ChatGPT backend could not shed context and risked hitting the context-window limit. Chat Completions mode was unaffected.

## Root cause

The compaction split is already three layers (see `providers/shared/compaction.ts`):

1. provider-agnostic orchestration `runCompactionCore` (boundary → render → splice),
2. representation via `CompactionOps<M>` (generic over message type),
3. the model call, injected per-provider as a `summarize(transcript, signal) => Promise<string>` closure.

`priorTurns` is a single canonical `OpenAIMessage[]` for **both** wire modes; the Responses translation happens at request-build time, not in storage. So the splice machinery was already wire-agnostic — **only the summarize transport** was Chat-Completions-bound. The fix supplies the missing transport rather than touching the core.

## Change

- **`oneShotResponses` (`oneshot.ts`)** — a Responses-wire sibling of `oneShotChatCompletion`. It:
  - reuses `buildResponsesRequestBody`, so every ChatGPT-backend quirk (omit `max_output_tokens`, `store: false`, non-empty `instructions`, gpt-5.x-only) stays in one place;
  - reuses `translateResponsesEvent` — the single source of truth for which stream event carries assistant text — so the summarize path can't drift from the turn path;
  - **streams** (`stream: true`) rather than issuing a non-streaming call: the ChatGPT/Codex backend is only ever streamed to in the codebase, so a throwaway summarize rides the one proven wire;
  - **drains in isolation** — its own `StreamState`, yielded `ProviderEvent`s discarded — so it never mutates session state (`lastUsage`, `priorTurns`), emits no trace, and dispatches no tools;
  - has no retry opinion (like the Chat-Completions sibling): it throws on error and the compaction core treats that as a safe no-op.
- **`compactHistory` (`query.ts`)** — the `wireMode === 'responses'` branch now selects the summarizer instead of bailing. A **non-abort** summarize failure latches a per-session `responsesCompactionUnavailable` flag so subsequent turn boundaries no-op cheaply (`responses-compaction-unavailable`) instead of re-issuing a doomed request every turn. Aborts/timeouts do **not** latch (transient, not a backend incompatibility). Failures remain safe no-ops — history is never corrupted.

## Why not a shared cross-provider summarizer?

Considered and rejected (via a `/devils-advocate` pass): the orchestration that *should* be shared already is. The transport can't be centralized because each summarize must reuse **that session's own `this.client`** (baseURL, credentials, OAuth headers, account-id) — a central service can't re-resolve it — and the mechanisms genuinely differ (anthropic streaming messages vs. OpenAI chat-completions vs. OpenAI responses). The right seam is per-provider transport, which is what this PR fills.

## Tests

- `oneshot.test.ts` — `oneShotResponses`: delta accumulation + trim, promise/`APIPromise` create, request shape (`stream:true`, model, no tools, `instructions`), ChatGPT-backend path (omits `max_output_tokens`, sets `store:false`) vs. public path (carries the cap), abort-signal forwarding, empty-stream → `''`, error propagation.
- `query.test.ts` — responses-wire auto-compaction goes through `responses.create` and **never** `chat.completions`; the latch causes the backend to be asked to summarize exactly once after a failure, then a manual `compact()` no-ops with `responses-compaction-unavailable`. Repurposed the stale `unsupported-wire-mode` test (an empty responses session now no-ops `history-too-short`, proving compaction is no longer structurally refused).

## Gates

- `pnpm test` — 698 files, **13077 passed**, 0 failed
- `pnpm audit:sdk:check` — OK (no new `@anthropic-ai/sdk` imports; this touches the `openai` SDK)
- `pnpm scan:env:check` — in sync (no env changes)

## Known limitation / follow-up

The ChatGPT-OAuth backend is only ever *streamed* to in the codebase; a throwaway summarize turn has **not** been confirmed against it live (no local creds in CI). The design degrades safely — on rejection the per-session latch reverts to today's behavior (no compaction) rather than erroring — so this is safe to land flag-guarded. Worth a one-off live confirmation on a real ChatGPT-OAuth session.

### DIFF
--- a/src/agent/providers/openai-compatible/oneshot.test.ts
+++ b/src/agent/providers/openai-compatible/oneshot.test.ts
@@ -20,7 +20,12 @@
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import type OpenAI from 'openai';
-import { oneShotChatCompletion, oneShotResponses, __setOpenAIOneShotClientFactory } from './oneshot.js';
+import {
+  oneShotChatCompletion,
+  oneShotResponses,
+  ResponsesSummaryIncompleteError,
+  __setOpenAIOneShotClientFactory,
+} from './oneshot.js';
 import type { OpenAIAuthResolution } from './auth.js';
 import type { ResponsesStreamEvent } from './responses-translate.js';
 
@@ -358,6 +363,25 @@ function streamOf(...events: ResponsesStreamEvent[]): AsyncIterable<ResponsesStr
   })();
 }
 
+/** The terminal event that licenses `oneShotResponses` to return its text. */
+const COMPLETED_EVENT: ResponsesStreamEvent = {
+  type: 'response.completed',
+  response: { usage: { input_tokens: 1, output_tokens: 2, total_tokens: 3 } },
+};
+
+/**
+ * Text deltas followed by `response.completed` — the only shape that yields a
+ * summary. Fixtures that stop short of a terminal event now reject by design
+ * (a never-completed stream is a truncated summary), so request-shape tests
+ * must still complete their stream.
+ */
+function completedStreamOf(...deltas: string[]): ResponsesStreamEvent[] {
+  return [
+    ...deltas.map((delta): ResponsesStreamEvent => ({ type: 'response.output_text.delta', delta })),
+    COMPLETED_EVENT,
+  ];
+}
+
 describe('oneShotResponses', () => {
   it('accumulates output_text.delta events and returns the trimmed text', async () => {
     const result = await oneShotResponses({
@@ -379,9 +403,7 @@ describe('oneShotResponses', () => {
 
   it('awaits a promise-returning create (SDK APIPromise) and drains it', async () => {
     const result = await oneShotResponses({
-      client: makeResponsesClient(async () =>
-        streamOf({ type: 'response.output_text.delta', delta: 'ok' }),
-      ),
+      client: makeResponsesClient(async () => streamOf(...completedStreamOf('ok'))),
       model: 'gpt-5.5',
       system: 'sys',
       user: 'transcript',
@@ -395,7 +417,7 @@ describe('oneShotResponses', () => {
     await oneShotResponses({
       client: makeResponsesClient((params) => {
         captured = params;
-        return streamOf({ type: 'response.output_text.delta', delta: 'x' });
+        return streamOf(...completedStreamOf('x'));
       }),
       model: 'gpt-4o',
       system: 'COMPACT-SYS',
@@ -417,7 +439,7 @@ describe('oneShotResponses', () => {
     await oneShotResponses({
       client: makeResponsesClient((params) => {
         captured = params;
-        return streamOf({ type: 'response.output_text.delta', delta: 'x' });
+        return streamOf(...completedStreamOf('x'));
       }),
       model: 'gpt-5.5',
       system: 'COMPACT-SYS',
@@ -436,7 +458,7 @@ describe('oneShotResponses', () => {
     await oneShotResponses({
       client: makeResponsesClient((_params, options) => {
         capturedSignal = options?.signal;
-        return streamOf({ type: 'response.output_text.delta', delta: 'x' });
+        return streamOf(...completedStreamOf('x'));
       }),
       model: 'gpt-5.5',
       system: 'sys',
@@ -472,5 +494,83 @@ describe('oneShotResponses', () => {
         isChatGptBackend: true,
       }),
     ).rejects.toThrow(/backend rejected/);
+  });
+
+  // ── truncation guards (review findings on PR #700) ───────────────────────────
+  //
+  // Both assert the same invariant from opposite directions: this function must
+  // never RESOLVE with a partial summary, because the compaction core splices a
+  // resolved summary over real history and reports success.
+
+  it('REJECTS when an abort lands mid-stream instead of resolving with partial text', async () => {
+    const controller = new AbortController();
+    // Emits one delta, aborts, then keeps yielding — mimicking openai@6, whose
+    // SSE iterator SWALLOWS an abort and ends cleanly. Without abortableStream
+    // the drain would exit normally and resolve with just 'partial'.
+    const stream = (async function* (): AsyncIterable<ResponsesStreamEvent> {
+      yield { type: 'response.output_text.delta', delta: 'partial' };
+      controller.abort();
+      yield { type: 'response.output_text.delta', delta: ' more' };
+      yield COMPLETED_EVENT;
+    })();
+
+    await expect(
+      oneShotResponses({
+        client: makeResponsesClient(() => stream),
+        model: 'gpt-5.5',
+        system: 'sys',
+        user: 'msg',
+        isChatGptBackend: true,
+        signal: controller.signal,
+      }),
+    ).rejects.toThrow();
+  });
+
+  it('REJECTS on response.failed after text deltas (partial summary, not a success)', async () => {
+    await expect(
+      oneShotResponses({
+        client: makeResponsesClient(() =>
+          streamOf(
+            { type: 'response.output_text.delta', delta: 'half a summ' },
+            { type: 'response.failed', response: {} },
+          ),
+        ),
+        model: 'gpt-5.5',
+        system: 'sys',
+        user: 'msg',
+        isChatGptBackend: true,
+      }),
+    ).rejects.toThrow(ResponsesSummaryIncompleteError);
+  });
+
+  it('REJECTS on response.incomplete, naming the terminal status', async () => {
+    await expect(
+      oneShotResponses({
+        client: makeResponsesClient(() =>
+          streamOf(
+            { type: 'response.output_text.delta', delta: 'cut short' },
+            { type: 'response.incomplete', response: { incomplete_details: { reason: 'max_output_tokens' } } },
+          ),
+        ),
+        model: 'gpt-5.5',
+        system: 'sys',
+        user: 'msg',
+        isChatGptBackend: true,
+      }),
+    ).rejects.toThrow(/max_output_tokens/);
+  });
+
+  it('REJECTS when the stream ends with no terminal event at all', async () => {
+    await expect(
+      oneShotResponses({
+        client: makeResponsesClient(() =>
+          streamOf({ type: 'response.output_text.delta', delta: 'truncated' }),
+        ),
+        model: 'gpt-5.5',
+        system: 'sys',
+        user: 'msg',
+        isChatGptBackend: true,
+      }),
+    ).rejects.toThrow(ResponsesSummaryIncompleteError);
   });
 });
--- a/src/agent/providers/openai-compatible/oneshot.test.ts
+++ b/src/agent/providers/openai-compatible/oneshot.test.ts
@@ -20,8 +20,9 @@
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import type OpenAI from 'openai';
-import { oneShotChatCompletion, __setOpenAIOneShotClientFactory } from './oneshot.js';
+import { oneShotChatCompletion, oneShotResponses, __setOpenAIOneShotClientFactory } from './oneshot.js';
 import type { OpenAIAuthResolution } from './auth.js';
+import type { ResponsesStreamEvent } from './responses-translate.js';
 
 // ── Auth mock ───────────────────────────────────────────────────────────────
 // `oneShotChatCompletion` calls `resolveOpenAIAuth(apiKey)` without deps, so we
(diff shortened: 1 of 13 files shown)


griffinwork40 (commented):

## Merge decision: **DO NOT MERGE** — 1 high, 3 medium, 2 low

Reviewed against branch HEAD `b286931` (`afk/responses-compaction-653`). The architecture argument in the PR body holds up under scrutiny: the summarize transport genuinely was the only Chat-Completions-bound layer, the splice machinery is wire-agnostic, and `shared/compaction.ts`, `compact.ts`, `request-body.ts`, `responses-translate.ts` are byte-identical to the parent commit — no scope creep. Claims 3 and 5 verified line by line. One defect blocks: the new drain drops the interrupt guard its own turn path documents as mandatory for this SDK.

---

### 1 · high · confidence: high · correctness/security
`src/agent/providers/openai-compatible/oneshot.ts:225-232` · ref:`b286931` · citation-type: file-state

**finding:** The isolated drain iterates the SDK stream with a bare `for await` instead of the turn path's `abortableStream` wrapper, so a mid-summarize interrupt is swallowed by the openai@6 SSE iterator, `oneShotResponses` **resolves with a truncated summary**, and the compaction core splices that partial text over history and reports `compacted: true`.

```ts
const state = createStreamState();
for await (const event of stream) {          // no abortableStream
  const events = translateResponsesEvent(event, state, 'compact-oneshot');
return state.assistantText.trim();           // resolves with PARTIAL text
```

The three links, each verified at the reviewed ref:

1. **openai@6.38.0 swallows the abort** — `node_modules/openai/core/streaming.js:73-78`: `catch (e) { if (isAbortError(e)) return; throw e; }`. A `return` in an async generator yields `done:true`, so a bare `for await` exits *cleanly*. This repo already documents exactly this hazard in `shared/abortable-stream.ts:26-29` ("its SDK — openai@6 — actually SWALLOWS a mid-stream abort and ends its iterator cleanly") and the turn path wraps accordingly at `query/stream-drive.ts:149`.
2. **The core never re-checks abort on the resolve path** — `shared/compaction.ts` calls `isAborted()` at `:543` (pre-summarize) and `:561` (inside `catch`) only. Between the resolved value and the mutation sits an emptiness guard and nothing else: `:568` `if (summary.trim().length === 0)` → `:575` `messages.splice(0, messages.length, ...newMessages)` → `:588` `return { compacted: true, ... }`. `deps.onSuccess` fires at `:578`, so the user is told compaction *succeeded* with a token-saved estimate.
3. **The interrupt is reachable on the auto path** — `beginAbort` assigns the compaction controller to the shared session field (`query.ts:963-966`), and `interrupt()` aborts that field (`query.ts:876-878`). Auto-compaction runs from `query.ts:350` while `AgentSession.currentState` is still `streaming` — `agent-session.ts:1062-1067` states the auto path deliberately does *not* set `compacting` — so ESC passes the state gate at `agent-session.ts:751`.

Not pre-existing: the Chat-Completions sibling issues `stream: false` (`oneshot.ts:125`), where an abort **rejects**. The partial-resolve window is new to this PR.

Blast radius is bounded — `applyCompaction` returns `[summary, ack, ...messages.slice(boundary)]` (`compaction.ts:430`), so the recent tail survives and only the pre-boundary slice is replaced by a mid-sentence summary. An abort landing before the first delta yields `''` → `empty-summary` no-op. *A severity note in the interest of transparency: the adversarial verification pass independently reproduced all three links but argued for `medium` on bounded-blast-radius grounds. This report holds `high` because the loss is silent, irreversible within the session, and mislabeled as success. Maintainer's call.*

**suggestion:** Wrap the source as `abortableStream(stream, signal)` — the helper already exists in `shared/` and is used by the turn path at `query/stream-drive.ts:149` — so an abort throws and the core's `catch`/`isAborted()` path maps it to the existing `aborted` no-op.

---

### 2 · medium · confidence: high · correctness
`src/agent/providers/openai-compatible/query.ts:1004` · ref:`b286931` · citation-type: diff-context

**finding:** The latch fires on **any** non-abort throw — including a transient 429 or 5xx — and nothing ever resets it, so one retryable blip permanently disables compaction for the session, re-creating the context-limit exhaustion this PR fixes.

```ts
} catch (err) {
  if (!signal.aborted) this.responsesCompactionUnavailable = true;
  throw err;
}
```

`git grep -n responsesCompactionUnavailable b286931` → 5 hits, all in `query.ts` (`:204` declaration, `:905`/`:982` docs, `:925` read, `:1004` write). No assignment of `false` outside the field initializer; `setModel` does not clear it and `this.client` is assigned only in the constructor. `oneShotResponses` also carries no retry, unlike the turn path's `MAX_CONNECTION_RETRIES`/`MAX_STREAM_RETRIES` — so the first transient failure is terminal. The inline comment at `:929` ("never reset this session") shows this is deliberate, but pairing "permanent" with "any error" is the risk.

**suggestion:** Latch only on non-retryable errors (reuse `query/retry.ts`'s `isRetryableConnectionError`/`isRetryableStreamError`), or make it a strike count cleared by the next successful turn.

---

### 3 · medium · confidence: high · correctness
`src/agent/providers/openai-compatible/query.ts:925-936` · ref:`b286931` · citation-type: file-state

**finding:** The latch short-circuits before `compactOpenAIHistory` is called, so it also disables the deterministic **no-LLM, no-network** microcompaction fallback — after one summarize failure the session cannot reclaim context by *any* mechanism, which overshoots the intent's stated goal of no-oping "cheaply instead of re-issuing a doomed request."

```ts
if (this.wireMode === 'responses' && this.responsesCompactionUnavailable) {
  return { compacted: false, reason: 'responses-compaction-unavailable', ... };
```

`compact.ts:312-320` runs `microcompactToolResults` inside `compactOpenAIHistory`, which the early return at `:930` never reaches. `compactModel` is resolved at `:937`, i.e. after the return.

**suggestion:** On the latched path skip only the summarize transport — still call `microcompactToolResults(this.priorTurns, opts)` and return a `microcompacted` result when `blocksCleared > 0`.

---

### 4 · medium · confidence: high · perf-observability
`src/agent/providers/openai-compatible/query.ts:350` · ref:`b286931` · citation-type: file-state

**finding:** A permanent per-session compaction disable is invisible to an operator on the auto path — the latch site emits no log or trace, and the auto caller discards the result, so `responses-compaction-unavailable` surfaces only if a human separately runs `/compact`.

```ts
await this.compactHistory('token_threshold');   // return value discarded
```

`git grep` for `logger`, `console.`, `warn`, `emitCompaction` in `query.ts` at the ref → 0 hits; `emitCompaction` lives at `compact.ts:292` on the success path only.

**suggestion:** Emit a one-time trace/warn at the latch site in `summarizeViaResponses` naming the underlying error, so the disable and its cause appear in the witness log.

---

### 5 · low · confidence: high · test-coverage
`src/agent/providers/openai-compatible/query.ts:1004` · ref:`b286931` · citation-type: file-state

**finding:** The abort-does-not-latch branch — behavior the PR body calls out explicitly — has no test; no test at the reviewed ref references the latch field or exercises a summarize failure with an aborted signal.

```
git grep -n 'responsesCompactionUnavailable' b286931 -- '*.test.ts'  → 0 matches
```
`oneshot.test.ts:433-448` captures the signal reference but never calls `controller.abort()`; `compaction.test.ts:192-203` pairs `isAborted: () => true` with a summarize that *rejects*. No test has summarize **resolve** while aborted — the exact interleaving in finding 1.

**suggestion:** Add a test that aborts the compaction controller mid-summarize and asserts (a) history is unchanged and (b) a later `compact()` does not return `responses-compaction-unavailable`.

---

### 6 · low · confidence: high · api-compat
`src/cli/slash/commands/core.ts:99` · ref:`b286931` · citation-type: file-state

**finding:** The new `responses-compaction-unavailable` reason has no branch in either renderer, so both surfaces print the raw kebab-case token where sibling reasons get prose.

```ts
ctx.out.info(`Nothing to compact (${reason}).`);
```
No crash risk — both renderers have generic fallbacks (`core.ts:99`, `telegram/formatter.ts:469`).

**suggestion:** Add a branch in `core.ts` and `formatCompactNoop` mapping it to actionable prose, matching the `not-supported` treatment at `core.ts:96`.

---

### Verified clean

- **`unsupported-wire-mode` removal is safe** — `git grep` at the ref → 1 hit, a comment in `query.test.ts:1348`. Zero production consumers. `ProviderCompactResult.reason` is an open `reason?: string` (`src/agent/provider.ts:442`), so the new literal is not a type break.
- **Isolation claim holds** — `responses-translate.ts:83-159` mutates only the passed `state`; `oneshot.ts` imports no `traceWriter` and no dispatcher. Session `lastUsage`/`priorTurns` are untouched and no transcript is traced from this path.
- **Timeout discrimination is correct** — `compact.ts:287-289` passes the same controller's signal into `summarize` and wires `abortInFlight`, and `compaction.ts:558` checks the `CompactionTimeoutError` sentinel *before* `isAborted()`, so a timeout is not misreported as a user abort and does not latch.
- **Claims 3 and 5 verified** — `buildResponsesRequestBody` reused (`oneshot.ts:202`), `translateResponsesEvent` reused (`:227`), `stream:true` set by the builder (`request-body.ts:54`), no tools (`activeTools: undefined`), own `StreamState` (`:225`). Every test enumerated in the PR body exists and asserts what the body says.
- **Shared helpers untouched** — the diff is 4 files; `buildResponsesRequestBody`, `translateResponsesEvent`, `createStreamState`, `oneShotChatCompletion` all keep their signatures.

### CI note

`Test (windows-latest)` is red, but the failures are in `tests/agent/memory/memory-loader.test.ts` and siblings — files this PR does not touch. The same job is red on `main` at `e71fbce9`, and `.github/workflows/ci.yml:94` marks non-ubuntu runners `continue-on-error` (advisory). **Not attributable to this PR.**

### What was not checked

- Wave 1 agents read against branch HEAD `b286931`; all blocking/high citations were verified at that ref, plus an independent adversarial re-derivation of finding 1 and a direct main-session re-read of `oneshot.ts:225-232`, `compaction.ts:538-590`, and the openai SDK iterator.
- Stated intent: PR #700 title + body — spec-compliance **assessed**; no unmet intent and no scope creep found beyond finding 3's overshoot.
- Gate claims (`pnpm lint`, `pnpm test` 13077 passed, `audit:sdk:check`, `scan:env:check`) were **not re-run** — reported as stated by the author.
- The PR's own admitted limitation stands: live ChatGPT-OAuth behavior of a throwaway summarize turn is unverifiable without credentials.
- Did not review the Telegram surface beyond the `formatCompactNoop` fallback, and did not execute any tests.

griffinwork40:
## PR #700 Review

**Reviewed ref:** `07b69f0f6abeb04be125169c3ff35f8011ffe262`  
**Scope:** 13 files, +932/−30  
**Intent:** Add safe history compaction to the OpenAI-compatible Responses wire while preserving Chat Completions behavior.

### Summary

The implementation satisfies the stated intent:

- `oneShotResponses` reuses the existing Responses request builder and event translator.
- A summary is accepted only after `response.completed`; incomplete or aborted streams throw instead of returning partial text.
- Summarization uses a fresh `StreamState`, leaving session history and usage state untouched.
- Only statuses proving an unsupported request shape latch compaction off; aborts, 429s, network errors, and incomplete streams do not latch.
- Failure paths preserve history and degrade to the existing safe no-op behavior.
- The Chat Completions summarizer remains unchanged.
- The new `compaction_disabled` trace and formatter handling makes the otherwise-silent session latch observable.

### Blocking findings

None.

### Non-blocking findings

1. **low · high confidence · test-coverage · `src/cli/slash/commands/core.ts:98-104` · ref:`07b69f0f6abeb04be125169c3ff35f8011ffe262` · citation-type:file-state**  
   **Finding:** The new `/compact` rendering branch for `responses-compaction-unavailable` has no direct assertion in `src/cli/slash/core-compact.test.ts`.  
   **Evidence:** That test covers `history-too-short`, `nothing-to-summarize`, `aborted`, and summarization failures, while repository search finds this new reason in the CLI source and an unrelated trace-rendering test. Query, Telegram, and trace behavior for the reason are covered elsewhere.  
   **Suggestion:** Add one slash-command test asserting that the reason is emitted through `ctx.out.warn` with the expected recovery text.

2. **nit · low confidence · api-compat · `src/agent/providers/openai-compatible/query.ts:1008` · ref:`07b69f0f6abeb04be125169c3ff35f8011ffe262` · citation-type:file-state**  
   **Finding:** The runtime reason string `unsupported-wire-mode` disappears in favor of `responses-compaction-unavailable`, but zero production importers or string consumers exist in the repository `[UNVERIFIED: no production importers]`.`  
   **Evidence:** The sole remaining occurrence is a historical comment in `query.test.ts`; `ProviderCompactResult.reason` is an open `string` and its type is not exported from the package entry point.  
   **Suggestion:** Mention the reason-string change in release notes if out-of-tree deep-import consumers are known.

### Unverified risk

`src/agent/providers/openai-compatible/query.ts:1097` persists the backend SDK error message in local `compaction_disabled` trace metadata, and `afk trace` can render that field. The same extraction pattern already exists in another provider, and no evidence showed client credentials or authorization headers entering the message. External forwarding of this trace metadata was not established, so this is not a demonstrated leak.

### Verification manifest

- No critical or high findings survived Wave 1, so citation/absence verification and `/shadow-verify` were not triggered.
- Both primary review agents reached their tool-call ceilings; their reports were treated as partial and synthesis retained only source-supported findings.
- Nothing was dropped as fabricated or false-absent.
- The reviewed branch HEAD was read at `07b69f0f6abeb04be125169c3ff35f8011ffe262`.
- Tests were not executed during this read-only review.
- Live ChatGPT-OAuth behavior was not exercised.
- External package consumers and external trace sinks were not inspected.
- Stated intent came from PR #700’s title and body, and spec compliance was assessed against it.

## Merge decision

**MERGE.** The identified issues are non-blocking; the transport, abort handling, narrow latch predicate, state isolation, history-preservation behavior, and Chat Completions compatibility are source- and test-backed.

## Done

- PR #700 was reviewed read-only across security, API compatibility, correctness, intent compliance, test coverage, performance, and observability.
- Evidence remains inspectable in PR #700, reviewed ref `07b69f0f6abeb04be125169c3ff35f8011ffe262`, and `/Users/griffinlong/.afk/state/skill-preflight/unbound-54ce8692fb15ded6/pr-700.diff`.
- No repository, branch, PR, or external state was modified.
- Runtime tests and a live ChatGPT-OAuth call remain deferred because this review was statically read-only.

---

Issue #360, opened 2026-07-02
Large-file split audit: tracking issue (10 offenders across src/)


griffinwork40:
## Context

Ran a repo-wide audit (`/gather` — parallel Structure Agent + Test Agent) of the longest files under `src/` to find genuine "should be split" candidates, as opposed to files that are merely long but cohesive. Methodology and full findings below; this issue tracks the 10 sub-issues it produced.

**Scan**: `find src -name "*.ts" -not -name "*.test.ts" | xargs wc -l | sort -rn` (production) and the equivalent for `*.test.ts`, then deep-read of the top 12 production candidates + top 5 test-file candidates, judged on cohesion (mixed responsibilities with clear seams) rather than line count alone.

## Sub-issues (worst offenders, ranked by severity)

- [x] #361 `src/agent/tools/dispatcher.ts` (1101 lines) — LOW risk
- [x] #362 `src/agent/providers/anthropic-direct/index.ts` (1062 lines) — MED risk
- [x] #363 `src/agent/tools/skill-executor.ts` (1141 lines) — LOW risk
- [x] #364 `src/agent/session/agent-session.ts` (1324 lines) — LOW-MED risk (wide test blast radius)
- [x] #365 `src/agent/providers/openai-compatible/query.ts` (1386 lines — largest production file in the repo) — LOW risk
- [x] #366 `src/cli/slash/plugin-skills.ts` (1031 lines — tests already split, easiest win) — LOW risk
- [x] #367 `src/cli/elicitation-repl.ts` (1081 lines) — LOW risk
- [x] #368 `src/cli/config.ts` (983 lines) — LOW risk
- [x] #369 `src/cli/terminal-compositor.test.ts` (5491 lines — largest file in the repo, test-only) — LOW risk
- [x] #370 `src/agent/providers/anthropic-direct/loop.test.ts` (2601 lines, test-only) — LOW risk

Suggested sequencing: #361 + #362 together first (shared grant-manager extraction), then #366 (lowest risk, existing test-split precedent), then the rest in any order. #369/#370 are test-only and can happen independently at any time.

## Cross-cutting finding (corrected — see shadow-verify log)

[...]

PR #701, opened 2026-07-25
refactor(providers): share AbortCoordinator across both provider query loops

griffinwork40:
## What

`openai-compatible/query.ts` hand-rolled the per-session abort slot that `anthropic-direct` already encapsulates in `AbortCoordinator` — including **six copies** of the compare-and-clear invariant:

```ts
if (this.abortController === controller) this.abortController = null;
```

This moves `AbortCoordinator` to `providers/shared/` and wires `openai-compatible` through it.

## Why

That invariant is safety-critical: a leaked stale controller makes `compact()`'s idle guard misread *"turn in flight"* as idle and run summarization mid-turn. Maintaining it by hand in two providers is exactly the drift risk `shared/tool-loop-cap.ts` exists to prevent — and `anthropic-direct/query.ts` already has **zero** raw `this.abortController` references because it solved this.

`AbortCoordinator` was already provider-neutral: **zero imports, no SDK types**. Only its docstring said "Anthropic." Copying it into `openai-compatible/` would have created the duplication this PR removes, so it moves to `shared/` instead.

| | before | after |
|---|---|---|
| raw `abortController` refs in openai query.ts | 9 | 0 |
| duplicated compare-and-clear sites | 6 | 0 (one `clear()` write path) |
| abort-related fields on the class | 4 | 1 collaborator |

Four fields (`abortController`, `pendingAbortReason`, `closeResolve`, `closedPromise`) collapse into one `abort` collaborator. Net **−26 lines of logic** in `query.ts`; total file length is ~flat because the freed space went into documenting the contract at each site.

## ⚠️ One behavioural delta — intentional, convergent

`compactHistory`'s `beginAbort` now goes through `AbortCoordinator.begin()`, which **drains a pending abort reason**.

An `interrupt()` parked between turns therefore pre-aborts the auto-compaction firing at that boundary and consumes the reason, instead of letting compaction run and deferring the abort to the next turn.

This matches `anthropic-direct/query/compact-handler.ts:148` and is arguably the more correct reading (ESC should stop compaction too), but it *is* a change. Narrow repro window: pending interrupt **and** auto-compact enabled **and** the threshold crossing at that exact boundary. Documented inline at the call site.

Preserving the old behaviour would have required adding a non-draining `begin` variant to the shared class for one caller — working against the anti-drift goal of the PR. Flagging it explicitly rather than burying it.

## Tests

Adds `shared/abort-coordinator.test.ts` (14 cases) — the class previously had **no dedicated test**, covered only indirectly via `concurrent-session-isolation.test.ts`. Now that it's load-bearing for two providers that gap was worth closing:

- scope lifecycle (begin → busy → clear → idle)
- **stale-scope compare-and-clear protection** (the highest-risk invariant)
- pending-abort drain, including drain-exactly-once and park-when-already-aborted
- the close-promise race the provider main loop depends on

## Verification

- `pnpm lint` clean (`tsc --noEmit`, strict)
- Full suite: **699 files, 13082 passed**, 14 skipped
- `audit:sdk:check`, `audit:env:check`, `scan:env:check` all green
- `pnpm build` succeeds

## Scope

Part of **#360** (god-class decomposition); follows **#407**, which extracted `client`/`model-params`/`retry` from this same file.

Deliberately **not** in this PR:
- **`SessionState` bag** — moves no logic on its own; only earns its place as the precondition for extracting the turn engine.
- **Turn engine** (`_runTurnInner` 209 LOC + `runIteration` 119) — the largest remaining mass and where ~4/5 of recent commits land, but a big diff on the hottest-churn path deserves its own review.
- **`compactHistory` extraction** — looks like a seam, isn't: its logic already lives in `compact.ts` behind injected callbacks, so the remaining 58 lines are a thin adapter. Moving it would be file-hop, not decomposition.

### DIFF
--- a/src/agent/providers/anthropic-direct/query.ts
+++ b/src/agent/providers/anthropic-direct/query.ts
@@ -69,7 +69,7 @@ import type {
 } from './types.js';
 import { repairOrphanToolUses } from './query/repair-orphan-tool-uses.js';
 import { type SessionState, createSessionState } from './query/session-state.js';
-import { AbortCoordinator } from './query/abort-coordinator.js';
+import { AbortCoordinator } from '../shared/abort-coordinator.js';
 import { RetryLayer } from './query/retry-layer.js';
 import { compactHistory } from './query/compact-handler.js';
 import { listUserTurns, rewindConversationHistory } from './query/rewind-conversation.js';
--- a/src/agent/providers/anthropic-direct/query/compact-handler.microcompact.test.ts
+++ b/src/agent/providers/anthropic-direct/query/compact-handler.microcompact.test.ts
@@ -16,7 +16,7 @@ import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import type { MessageParam } from '@anthropic-ai/sdk/resources';
 import { compactHistory } from './compact-handler.js';
 import { createSessionState, type SessionState } from './session-state.js';
-import { AbortCoordinator } from './abort-coordinator.js';
+import { AbortCoordinator } from '../../shared/abort-coordinator.js';
 import type { RetryLayer } from './retry-layer.js';
 import type { ToolDispatcher } from '../tool-dispatcher.js';
 import { findCompactionBoundary } from '../compact.js';
--- a/src/agent/providers/anthropic-direct/query/compact-handler.ts
+++ b/src/agent/providers/anthropic-direct/query/compact-handler.ts
@@ -56,7 +56,7 @@ import { emitCompaction } from '../../../trace/emit.js';
 import { resolveModelId } from '../../../session/model-resolution.js';
 import type { AnthropicClientLike } from '../types.js';
 import type { SessionState } from './session-state.js';
-import type { AbortCoordinator } from './abort-coordinator.js';
+import type { AbortCoordinator } from '../../shared/abort-coordinator.js';
 import type { RetryLayer } from './retry-layer.js';
 import { env } from '../../../../config/env.js';
 
--- a/src/agent/providers/anthropic-direct/query/rewind-conversation.test.ts
+++ b/src/agent/providers/anthropic-direct/query/rewind-conversation.test.ts
@@ -12,7 +12,7 @@ import { describe, it, expect } from 'vitest';
 import type { ContentBlockParam, MessageParam } from '@anthropic-ai/sdk/resources';
 import { listUserTurns, rewindConversationHistory } from './rewind-conversation.js';
 import type { SessionState } from './session-state.js';
-import type { AbortCoordinator } from './abort-coordinator.js';
+import type { AbortCoordinator } from '../../shared/abort-coordinator.js';
 
 function makeState(messages: MessageParam[], closed = false): SessionState {
   // The handler only reads `messages` + `closed`; the rest of SessionState is
--- a/src/agent/providers/anthropic-direct/query/rewind-conversation.ts
+++ b/src/agent/providers/anthropic-direct/query/rewind-conversation.ts
@@ -35,7 +35,7 @@ import type {
 } from '../../../provider.js';
 import { repairOrphanToolUses } from './repair-orphan-tool-uses.js';
 import type { SessionState } from './session-state.js';
-import type { AbortCoordinator } from './abort-coordinator.js';
+import type { AbortCoordinator } from '../../shared/abort-coordinator.js';
 
 const PREVIEW_MAX_CHARS = 72;
 
--- a/src/agent/providers/openai-compatible/query.ts
+++ b/src/agent/providers/openai-compatible/query.ts
@@ -11,8 +11,10 @@
  * Lifecycle:
  *   - constructor is synchronous; emits `session.init` on first iterator pull
  *   - main loop awaits user turns from `promptStream`, races against
- *     `closedPromise` so `close()` unblocks a "waiting for next turn" state
- *   - each user turn opens a new AbortController; `interrupt()` aborts it
+ *     `abort.closedPromise` so `close()` unblocks a "waiting for next turn"
+ *     state
+ *   - each user turn opens a new abort scope via `abort.begin()`;
+ *     `interrupt()` aborts it — see shared/abort-coordinator.ts
  *   - `runTurn` iterates model→tools→model until the model stops calling tools
  *     (or the tool-round cap fires — see shared/tool-loop-cap.ts — after which
  *     one tools-stripped wind-down round runs, matching anthropic-direct/loop.ts)
@@ -88,6 +90,7 @@ import {
   shouldAutoCompact,
   resolveAutoCompactThreshold,
 } from '../shared/auto-compact.js';
+import { AbortCoordinator, CLOSED_SENTINEL } from '../shared/abort-coordinator.js';
 import { HookBlockedError, DenialCircuitBreakerError } from '../../../utils/errors.js';
 import { COMPACT_SYSTEM_PROMPT, wrapTranscriptForSummary } from '../shared/compaction.js';
 import { compactOpenAIHistory, readShrinkFraction } from './compact.js';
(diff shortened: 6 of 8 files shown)


griffinwork40:
### CI note: the `Test (windows-latest)` red is pre-existing, not from this PR

All gating checks pass — **`Test (ubuntu-latest)` (the required gate) is SUCCESS**, as are macOS, `Lint & Build`, Docs, Publish Bundle Smoke, and PTY.

`windows-latest` is advisory by design (`ci.yml:94` sets `continue-on-error` for every non-ubuntu matrix OS, with the comment at `ci.yml:83-89` noting the Windows-compat grind is in progress). Its failure does not block merge.

I verified it is not caused by this change:

| | main @ `e71fbce` (this PR's base) | this PR |
|---|---|---|
| windows job conclusion | **failure** | failure |
| failing test files | 85 | 86 |
| failing tests | **537** | **537** |

Same failing-test count, and the failures are the same pre-existing Windows-compat set — `memory-loader.test.ts`, `skill-bridge.test.ts`, `bash.test.ts`, `tool-injector.test.ts`, `launchd.test.ts`, etc. The two `providers/` files that appear (`anthropic-direct/grant-manager.test.ts`, `openai-compatible/auth.test.ts`) also fail on main's Windows job and are untouched here.

Root cause of the sampled `memory-loader.test.ts` failure, for whoever picks up the compat work: the suite sets `process.env['HOME']` to a tmp dir, but `getAfkHome()` falls back to `join(homedir(), '.afk')` (`src/paths.ts:56`) and Node's `os.homedir()` reads `USERPROFILE` on Windows — so the fixture directory is never consulted.

**The file this PR adds passes on Windows:** `✓ src/agent/providers/shared/abort-coordinator.test.ts (14 tests)`.

Worth flagging for future triage: a run-level conclusion of `success` on main does **not** mean the Windows job passed — `continue-on-error` masks it at the run level. The job-level conclusion has to be read directly.

griffinwork40:
Wave 1.5 verification complete — all load-bearing claims re-derived independently. Synthesizing inline (4 findings, all low/nit; dispatching a synthesis agent for a 4-finding dedup would be pure overhead).

---

# Code Review — PR #701

**`refactor(providers): share AbortCoordinator across both provider query loops`**
`main` ← `afk/split-openai-query-file` · ref `f04e8e5` · 8 files · +242/−69 (311 lines) · regime **full** · mergeable: yes

## Verdict: ✅ MERGE

No blocking findings. This is a genuine behaviour-preserving extraction: **8/8 semantic-equivalence verdicts came back `equivalent`**, and I independently re-derived the four highest-stakes ones rather than accepting the agents' compressed findings.

## Independently re-derived (Wave 1.5)

| Claim | Method | Result |
|---|---|---|
| Rename is docs-only, no logic smuggled in | `git diff` old-path↔new-path, non-comment lines filtered | ✅ **zero** non-comment lines differ |
| `begin()` matches the old inline mint+drain line-for-line | read both bodies at ref | ✅ identical, incl. early-return slot behaviour |
| `isIdle` eager snapshot is safe | `compact.ts:237` (`isIdle: boolean`) read at `:274` **before** `beginAbort()` at `:278` | ✅ no staleness window; old code equally eager |
| Author's delta cites `compact-handler.ts:148` | `git show` that exact line | ✅ `const controller = abort.begin();` — citation exact |
| 6 clear sites, 0 raw `abortController` refs, 14 `it()` cases | `grep -c` at ref | ✅ 6 / 0 / 14 — every countable claim passes |
| Rename stragglers anywhere in tree | unscoped `git grep` incl. `scripts/`, `docs/`, `website/`, lockfiles | ✅ none; old module deleted |

**Citation manifest:** 2 citations checked → both `verified`. **Absence manifest:** 4 claims checked → 3 `confirmed-absent`, 1 `false-absent` (the "close() race untested" claim, correctly self-refuted by its own agent against `query.test.ts:1248`). Zero `fabricated`, zero findings dropped.

## Findings

### 1 · `low` · high confidence · security + test-coverage (two agents converged independently)
`src/agent/providers/shared/abort-coordinator.ts:30-32` · ref:f04e8e5 · citation-type:diff-context

**finding:** The docblock names `anthropic-direct/concurrent-session-isolation.test.ts` as a canary for the compare-and-clear invariant, but that file contains **zero** occurrences of `abort` — it tests `readRoots`/`writeRoots` isolation across `resolveProvider()` instances, nothing about abort scopes.

```
 * newer controller in place. The canary tests for this invariant are
 * `anthropic-direct/concurrent-session-isolation.test.ts` and
 * `shared/abort-coordinator.test.ts`.
```

**why it's worth fixing here:** the inaccuracy pre-dates the PR, but the PR *edits this exact hunk* and adds a real canary alongside the false one — so it launders a wrong coverage claim through a freshly-reviewed line. This docblock is the only place documenting where the safety-critical invariant's canary lives.
**suggestion:** Drop the `concurrent-session-isolation.test.ts` reference; `shared/abort-coordinator.test.ts` is the accurate canary.

### 2 · `low` · high confidence · test-coverage
`src/agent/providers/openai-compatible/query.ts:953-966` · ref:f04e8e5 · citation-type:file-state

**finding:** The one intentional behavioural delta in this PR — compaction's `begin()` now draining a between-turns parked abort reason — has no test exercising it through the real `OpenAICompatibleQuery`; `compact.test.ts:69-70` injects `beginAbort`/`clearAbort` as stubs, so the drain never runs.
**suggestion:** One `compact.test.ts` case: `requestAbort('interrupted')` with no scope in flight, then `compactHistory()`, asserting the summarize call never fires.

### 3 · `nit` · test-coverage — `shared/abort-coordinator.test.ts:17-21`
`mints a distinct controller per begin()` asserts `second !== first`, which is a guarantee of `new AbortController()`, not of any coordinator logic. Fold into the compare-and-clear test, which needs distinct controllers to be meaningful anyway.

### 4 · `nit` · api-compat — `anthropic-direct/query.ts:351`
`if (nextOrClose === '__closed__') break;` — anthropic-direct still compares the raw literal while openai-compatible now uses the exported `CLOSED_SENTINEL` (`query.ts:326`). Pre-existing, type-safe (the literal union catches drift at compile time), but it's precisely the source-form divergence this PR's anti-drift thesis targets.

## spec-compliance: pass

Stated intent captured from PR #701 title+body — **assessed**. No unmet intent, no scope creep. All 5 countable claims verified pass. The three explicitly-deferred items (`SessionState` bag, turn engine, `compactHistory` extraction) are correctly absent. The behavioural delta is real, correctly cited, documented inline at the call site, and disclosed in the PR body rather than buried — good practice.

## Pre-existing, not introduced here (informational)

`runTurn`'s pre-aborted early return leaves a non-`null` slot, so an auto-compact landing between that return and the next `begin()` reads `turn-in-flight` and skips. It self-heals on the next turn (`begin()` overwrites unconditionally) and is **byte-identical** pre/post. Not a regression — but it is now a property of code shared by two providers, so worth knowing.

## What was not checked

- **Ref:** all Wave 1 agents read from branch HEAD `f04e8e59d89d268205603f28577620c530720397`; citations verified at that ref.
- **Stated intent:** PR #701 title + body — spec-compliance assessed.
- **No execution:** `pnpm lint` / test suite / `pnpm build` / `audit:*` were **not run** (read-only review). The PR's green-CI claims (13082 passed, all audits green) are **unverified by this review**.
- Anthropic-direct's own 5 coordinator call sites were grep-confirmed, not line-audited (out of scope — parity target was openai-compatible).
- A thrown (not yielded) exception from `runIteration`/`dispatchAndAppend` bypassing all 6 clear sites was confirmed equivalent old↔new but not explored further.
- `dist/` output and other worktree branches not inspected.
- **Delegation note:** the original Wave 1 correctness agent died mid-stream (`no message_stop`); it was re-dispatched split into two narrower agents with nesting disabled. No dimension went uncovered. `/shadow-verify` was **not** invoked — the skill gates it on `critical`/`high` findings and none exist; the inline Wave 1.5 re-derivation above is a stronger check than the `low`/`nit` tier warrants.

---

**Done**

- Ran the full `/review` pipeline on PR #701: inline target/intent resolution → Wave 1 (3 research agents across security, api-compat, correctness, spec-compliance, test-coverage, perf-observability) → Wave 1.5 citation + absence verification (re-derived inline) → synthesis.
- Evidence: diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/db8ca9ab-78eb-4717-ae41-0b5c943504cb/pr-701.diff`; all citations verified against local ref `f04e8e59d89d268205603f28577620c530720397`; verification commands and their outputs are in this session's tool results and witness trace (`afk trace show`).
- World state unchanged — review is read-only. No files edited, nothing committed, no PR comment posted, no `gh` mutation.
- Recommendation **MERGE** with 4 non-blocking findings (2 `low`, 2 `nit`); findings #1 and #2 are the two worth addressing and neither gates the merge.
- Pending/deferred: the author's CI claims (lint, 13082-test suite, `audit:*`, build) were not executed here and remain unverified by this review; findings are reported only, not fixed, per the skill's read-only constraint.

---

PR #702, opened 2026-07-25
fix(telegram): bound and latency-gate ◦ tool-progress lines

griffinwork40:
## Summary

The Telegram live preview appended one `◦` progress line per tool round to `accumulated` with **no bound and no delay**. Every turn — including fast ones — churned the message, and a tool-heavy turn grew it without limit. `cleanFinal` already deletes the noisy preview at the end, so *scrollback* was clean; the problem was purely the live experience, and there was no knob to tune it.

Prior art check: OpenClaw (the closest analogue — self-hosted agent over Telegram) uses the same send-and-edit transport we do, and solves this with a bounded rolling window (`maxLines` default 8) plus a 5s start delay. This lands the same shape. Notably it does **not** need Telegram's newer `sendMessageDraft` (Bot API 9.3+), which our `telegraf@4.16.3` (Bot API 7.1) couldn't call anyway.

Four changes, all inside `streamResponse`:

| # | Change | Effect |
|---|---|---|
| 1 | **Latency gate** (`PROGRESS_START_DELAY_MS` = 5s) | Fast turns render **zero** `◦` lines — straight from `Thinking…` to the answer. Withheld lines are still recorded, so the gate opening shows recent rounds: a render delay, not a drop. |
| 2 | **Rolling window** (`MAX_PROGRESS_LINES` = 6) | The region is a bounded, rewritable tail of `accumulated` instead of an append-only log — the same treatment `renderSubagentFooter` already applies to sub-agent steps. |
| 3 | **Suggestion dedupe** compares `answerText`, not `accumulated` | Fixes a latent bug: the gate means "don't echo text already rendered as the answer", but `accumulated` also carries the `◦` region, making the comparison spuriously unequal and appending a duplicate 💡 on exactly the tool-heavy turns that produce progress. |
| 4 | **Activity receipt** (`⏱️ N steps · Ns`) | The turn's cost survives the preview deletion. Appended at **delivery only** — never to `answerText`, which `onComplete` records into resumable session history. |

## Two couplings worth reviewer attention

Both were found by adversarial re-derivation before implementation, and each now has a regression test that **fails if the coupling is broken** (verified by deliberately reintroducing the bug):

1. **`inContentRun = false` in the progress handler is not cosmetic — it is the `stream_retry` round boundary.** Gating it off (the obvious way to implement "hide progress") leaves `contentRunStart*` at an older offset, so a later `stream_retry` truncates **more** of `answerText` than the retry produced — silently destroying already-streamed answer text. The *render* is gated; that assignment stays unconditional. Guard: `REGRESSION: a gated-off progress round still sets the stream_retry boundary…` (asserts `AAACCC`, not `CCC`).
2. **Progress-only turns.** With progress withheld and no answer text (reachable via an abort mid tool-loop, which yields `turn.completed` with no assistant message), `accumulated` was empty and the user was stranded on the bare `Thinking…` placeholder. The `done` branch now flushes the withheld region explicitly. Guard: `dead-turn guard: a fast progress-only turn still delivers…`.

The `progressRunStart` offset is only valid while nothing else appends to `accumulated`, so every other writer (content chunk, assistant message, suggestion, `stream_retry` rollback) calls `endProgressRun()` first. That invariant is documented at the declaration.

## Deliberately out of scope

- **Operator-facing config** for the delay/window. `progressDelayMs` is exposed on the options object as a test seam only; a `telegram.progress` config key would pull in `ENV_REGISTRY` + `docs/env-registry.*` sync and is better as a follow-up.
- **`<blockquote expandable>`** for collapsed detail. Investigated and rejected for now: `markdownToTelegramHtml` escapes `<`/`>` unconditionally before injecting its own fixed tag set (`formatter.ts:90-93`), so an injected tag renders as literal text. Would need the tag emitted inside the formatter plus tag-aware `splitLongMessage`.
- **Pre-existing bug, untouched:** on a progress-only turn `onComplete` fires with `answerText === ''` and persists `{assistant: ''}` into resumable history with no non-empty check (`streaming.ts` → `session-stats.ts`). Not caused by this change and not fixed here; worth its own issue.

## Behavior change for reviewers to weigh

Fast tool-using turns now show **no** intermediate progress. That is the intent, but it is a default-behavior change: someone who liked watching each round will see less. The receipt (#4) is the compensating signal.

## How was this tested?

- `pnpm lint` — clean.
- `pnpm test` — **699 files, 13091 passed**, 14 skipped.
- `pnpm audit:env:check`, `pnpm scan:env:check`, `pnpm audit:sdk:check` — all pass.
- `pnpm build` — clean.
- 8 new cases in `streaming.test.ts`: latency gate (asserts the production default, no override), rolling window, retry-boundary regression, dead-turn guard, receipt present / absent / not-recorded, suggestion dedupe, and pure-helper unit tests.
- Two existing tests now pass `progressDelayMs: 0` because they assert progress *rendering* rather than gating; the assertions themselves are unchanged. No assertion was weakened — when the receipt initially tripped the `not.toContain('◦')` invariant, the receipt's marker was changed to `⏱️` rather than relaxing the test.

Not tested: live Telegram delivery (no integration harness); the 5s gate is exercised via the injectable override plus the default-path test.

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes

### DIFF
--- a/src/telegram/streaming.test.ts
+++ b/src/telegram/streaming.test.ts
@@ -1108,3 +1108,82 @@ describe('bounded ◦ progress region', () => {
     expect(renderActivityReceipt(2, 200)).toBe('\n\n⏱️ 2 steps · 1s');
   });
 });
+
+describe('◦ progress region — PR #702 review follow-ups', () => {
+  const sleep = (ms: number): Promise<void> => new Promise((r) => setTimeout(r, ms));
+  function progressEvent(description: string): OutputEvent {
+    return {
+      type: 'progress',
+      progress: { taskId: 't', description, totalTokens: 0, toolUses: 1, durationMs: 1 },
+    } as OutputEvent;
+  }
+
+  it('bounds progress across INTERLEAVED content, not just within one consecutive run (Codex P1)', async () => {
+    // Regression: the region used to be spliced into `accumulated` at an offset,
+    // so any content chunk froze it in place and the next progress event started a
+    // NEW region. An alternating stream — what the anthropic-direct and
+    // openai-compatible loops actually emit — therefore kept all 9 lines. As a
+    // single footer the MAX_PROGRESS_LINES bound holds globally.
+    const { ctx, edits } = makeCtx();
+    const session = makeSession(async function* () {
+      for (let i = 1; i <= 9; i++) {
+        yield progressEvent(`round-${i}`);
+        yield { type: 'chunk' as const, chunk: { type: 'content' as const, content: `c${i} ` } };
+      }
+      yield { type: 'done' as const, metadata: undefined };
+    });
+
+    await streamResponse(ctx, session, 'go', undefined, { progressDelayMs: 0 });
+
+    const final = edits[edits.length - 1] ?? '';
+    // Window of 6 over NINE rounds that each had content in between.
+    expect(final).toContain('round-9');
+    expect(final).toContain('round-4');
+    expect(final).not.toContain('round-3');
+    expect(final).not.toContain('round-1');
+    // Answer content is untouched by the bounded region.
+    expect(final).toContain('c1 ');
+    expect(final).toContain('c9 ');
+  });
+
+  it('opens the gate on a timer when no further progress event arrives (Codex P2)', async () => {
+    // Regression: the gate was re-checked ONLY when another `progress` event
+    // arrived, so a single tool running silently past the delay left the user on
+    // `Thinking…` for its whole duration. cleanFinal delivers only `answerText`
+    // (no ◦), so the progress line can reach the user ONLY via a live edit —
+    // which means the timer fired.
+    const { ctx, replies, edits } = makeCtx();
+    const session = makeSession(async function* () {
+      yield progressEvent('slow tool running');
+      await sleep(90); // silent stretch longer than progressDelayMs
+      yield { type: 'chunk' as const, chunk: { type: 'content' as const, content: 'Answer.' } };
+      yield { type: 'done' as const, metadata: undefined };
+    });
+
+    await streamResponse(ctx, session, 'go', undefined, { cleanFinal: true, progressDelayMs: 25 });
+
+    expect(edits.some((e) => e.includes('slow tool running'))).toBe(true);
+    // The clean final answer is still free of ◦ noise.
+    expect(replies.some((r) => r.includes('◦'))).toBe(false);
+    expect(replies.some((r) => r.includes('Answer.'))).toBe(true);
+  });
+
+  it('delivers a gated progress-only turn that ends WITHOUT a done event (review: medium)', async () => {
+    // Regression introduced by the first commit of this PR: the withheld-region
+    // flush was added to the `done` branch only. A provider stream that ends by
+    // graceful iterator close (no done/error) hit the post-loop branch, where
+    // bare `accumulated` is empty under a closed gate — so nothing was delivered
+    // and the user was stranded on `Thinking…`. Uses the DEFAULT gate so every
+    // line really is withheld.
+    const { ctx, replies, edits } = makeCtx();
+    const session = makeSession(async function* () {
+      yield progressEvent('did work then the stream closed');
+      // no `done` — the generator simply returns
+    });
+
+    await streamResponse(ctx, session, 'go', undefined, { cleanFinal: true });
+
+    expect([...replies, ...edits].some((t) => t.includes('did work then the stream closed'))).toBe(true);
+    expect([...replies, ...edits].every((t) => t !== 'Thinking…') || replies.length > 0).toBe(true);
+  });
+});
--- a/src/telegram/streaming.test.ts
+++ b/src/telegram/streaming.test.ts
@@ -3,7 +3,7 @@
  */
 
 import { describe, it, expect, vi } from 'vitest';
-import { streamResponse, StreamTimeoutError, renderSubagentFooter, replyWithFloodRetry } from './streaming.js';
+import { streamResponse, StreamTimeoutError, renderSubagentFooter, renderProgressRegion, renderActivityReceipt, replyWithFloodRetry } from './streaming.js';
 import { TelegramError } from 'telegraf';
 import type { Context } from 'telegraf';
 import type { IAgentSession, OutputEvent } from '../agent/types.js';
@@ -105,7 +105,10 @@ describe('streamResponse', () => {
       );
     });
 
-    await streamResponse(ctx, session, 'go');
+    // progressDelayMs: 0 opts out of the latency gate (PROGRESS_START_DELAY_MS)
+    // so this test asserts the RENDERING contract deterministically. The gate
+    // itself is covered by the 'latency gate' tests below.
+    await streamResponse(ctx, session, 'go', undefined, { progressDelayMs: 0 });
     const joined = [...replies, ...edits].join('\n');
     expect(joined).toContain('◦ Researching codebase');
     expect(joined).toContain('(Grep)');
@@ -375,7 +378,9 @@ describe('streamResponse', () => {
   it('default (no cleanFinal): force-flushes the in-place preview (noise included) and never deletes', async () => {
     const { ctx, edits, deletes } = makeCtx();
 
-    await streamResponse(ctx, answerWithProgress(), 'go');
+    // progressDelayMs: 0 opts out of the latency gate so the ◦ line is rendered
+    // into `accumulated` and this test still exercises the legacy flush path.
+    await streamResponse(ctx, answerWithProgress(), 'go', undefined, { progressDelayMs: 0 });
 
     // Legacy force-flushes `accumulated` on done, which mixes the ◦ progress
     // noise into the final in-place edit — the behavior cleanFinal improves on.
@@ -931,5 +936,175 @@ describe('replyWithFloodRetry (flood-control 429 backoff)', () => {
       replyWithFloodRetry(reply, 'hi', { parse_mode: 'HTML' }, { sleep: async () => {} }),
     ).rejects.toBe(err);
     expect(reply).toHaveBeenCalledTimes(1);
+  });
+});
+
+describe('bounded ◦ progress region', () => {
+  function progressEvent(description: string): OutputEvent {
+    return {
+      type: 'progress',
+      progress: { taskId: 't', description, totalTokens: 0, toolUses: 1, durationMs: 1 },
+    } as OutputEvent;
+  }
+
+  it('latency gate: a fast turn renders NO ◦ progress lines at all', async () => {
+    // The user-visible point of the gate: most turns finish well under
+    // PROGRESS_START_DELAY_MS, and those turns should go straight from the
+    // `Thinking…` placeholder to the answer with zero `◦` churn. No
+    // progressDelayMs override here — this asserts the production default.
+    const { ctx, replies, edits } = makeCtx();
+    const session = makeSession(async function* () {
+      yield* yieldEvents(
+        progressEvent('Running tool'),
+        progressEvent('Running another tool'),
+        { type: 'chunk', chunk: { type: 'content', content: 'Done thinking.' } },
+        { type: 'done', metadata: undefined },
+      );
+    });
+
+    await streamResponse(ctx, session, 'go', undefined, { cleanFinal: true });
+
+    expect([...replies, ...edits].some((t) => t.includes('◦'))).toBe(false);
+    expect(replies.some((r) => r.includes('Done thinking.'))).toBe(true);
+  });
+
+  it('rolling window: only the most recent MAX_PROGRESS_LINES survive', async () => {
+    // Regression: progress lines used to append to `accumulated` without bound,
+    // so a tool-heavy turn grew the preview one line per round forever.
+    const { ctx, edits } = makeCtx();
+    const session = makeSession(async function* () {
+      for (let i = 1; i <= 9; i++) yield progressEvent(`round-${i}`);
+      yield { type: 'done', metadata: undefined };
+    });
+
+    await streamResponse(ctx, session, 'go', undefined, { progressDelayMs: 0 });
+
+    // `done` force-flushes the full accumulated buffer into the preview.
+    const final = edits[edits.length - 1] ?? '';
+    expect(final).toContain('round-9');
+    expect(final).toContain('round-4'); // 9 rounds, window of 6 → 4..9 kept
+    expect(final).not.toContain('round-3');
+    expect(final).not.toContain('round-1');
+  });
+
+  it('REGRESSION: a gated-off progress round still sets the stream_retry boundary, so a retry does not eat earlier answer text', async () => {
+    // This is the trap a shadow-verify wave caught. The `progress` handler's
+    // `inContentRun = false` is NOT cosmetic — it is the stream_retry round
+    // boundary. Suppressing progress RENDERING must not skip it: if it were
+    // skipped, contentRunStartAnswer would still point at offset 0 when the
+    // second content run began, and the stream_retry below would slice
+    // answerText back to '' — silently destroying 'AAA', text the model had
+    // already streamed. Runs with the DEFAULT latency gate (progress renders
+    // nothing here) precisely to prove render-suppression keeps the boundary.
+    const { ctx, replies } = makeCtx();
+    const session = makeSession(async function* () {
+      yield { type: 'chunk', chunk: { type: 'content', content: 'AAA' } };
+      yield progressEvent('tool round between two content runs');
+      yield { type: 'chunk', chunk: { type: 'content', content: 'BBB' } };
+      yield { type: 'stream_retry' } as OutputEvent;
+      yield { type: 'chunk', chunk: { type: 'content', content: 'CCC' } };
+      yield { type: 'done', metadata: undefined };
+    });
+
+    let recorded: string | undefined;
+    await streamResponse(ctx, session, 'go', undefined, {
+      cleanFinal: true,
+      onComplete: (text) => { recorded = text; },
+    });
+
+    // 'AAA' survived the retry; only the retried round ('BBB') was discarded.
+    expect(recorded).toBe('AAACCC');
+    expect(replies.some((r) => r.includes('AAACCC'))).toBe(true);
+  });
+
+  it('dead-turn guard: a fast progress-only turn still delivers, never stranding the user on `Thinking…`', async () => {
+    // With the latency gate active and no answer text (e.g. an abort mid
+    // tool-loop yields turn.completed with no assistant message), `accumulated`
+    // is empty — so without an explicit flush the user's entire reply would be
+    // the bare `Thinking…` placeholder: a silent dead turn.
+    const { ctx, replies, edits } = makeCtx();
+    const session = makeSession(async function* () {
+      yield* yieldEvents(
+        progressEvent('searched the codebase'),
+        { type: 'done', metadata: undefined },
+      );
+    });
+
+    await streamResponse(ctx, session, 'go', undefined, { cleanFinal: true });
+
+    const all = [...replies, ...edits];
+    expect(all.some((t) => t.includes('searched the codebase'))).toBe(true);
+    // The placeholder is not the last thing the user sees.
+    expect(all[all.length - 1]).not.toBe('Thinking…');
+  });
+
+  it('activity receipt: summarizes tool rounds on the clean final but never enters recorded history', async () => {
+    const { ctx, replies } = makeCtx();
+    const session = makeSession(async function* () {
+      yield* yieldEvents(
+        progressEvent('step one'),
+        progressEvent('step two'),
+        { type: 'chunk', chunk: { type: 'content', content: 'The answer.' } },
+        { type: 'done', metadata: undefined },
+      );
+    });
+
+    let recorded: string | undefined;
+    await streamResponse(ctx, session, 'go', undefined, {
+      cleanFinal: true,
+      onComplete: (text) => { recorded = text; },
+    });
+
+    const finalReply = replies[replies.length - 1] ?? '';
+    expect(finalReply).toContain('The answer.');
+    expect(finalReply).toContain('2 steps');
+    // The receipt is presentation only: recorded turn history stays clean so a
+    // CLI `--resume` does not replay UI chrome as assistant text.
+    expect(recorded).toBe('The answer.');
+    expect(recorded).not.toContain('⏱️');
+  });
+
+  it('activity receipt: absent on a turn with no tool rounds', async () => {
+    const { ctx, replies } = makeCtx();
+    const session = makeSession(async function* () {
+      yield* yieldChunks('Just an answer.');
+    });
+
+    await streamResponse(ctx, session, 'go', undefined, { cleanFinal: true });
+
+    expect(replies[replies.length - 1]).toBe('Just an answer.');
+  });
+
+  it('suggestion dedupe compares against the ANSWER, not the progress-polluted buffer', async () => {
+    // Regression: the gate used to compare the suggestion against `accumulated`,
+    // which also carries the `◦` region. On a tool-heavy turn that made the
+    // comparison spuriously unequal, so a suggestion identical to the answer was
+    // appended anyway — a duplicate 💡 echo on exactly the turns with progress.
+    const { ctx, replies } = makeCtx();
+    const session = makeSession(async function* () {
+      yield* yieldEvents(
+        progressEvent('did some work'),
+        { type: 'chunk', chunk: { type: 'content', content: 'Hi! What can I help with?' } },
+        { type: 'suggestion', suggestion: 'Hi! What can I help with?' },
+        { type: 'done', metadata: undefined },
+      );
+    });
+
+    await streamResponse(ctx, session, 'go', undefined, { cleanFinal: true, progressDelayMs: 0 });
+
+    expect(replies.some((r) => r.includes('💡'))).toBe(false);
+  });
+
+  it('renderProgressRegion/renderActivityReceipt are pure and bounded', () => {
+    expect(renderProgressRegion([])).toBe('');
+    expect(renderProgressRegion(['a', 'b'])).toBe('\na\nb');
+    // Window of 6: a 9-entry input keeps the tail only.
+    const nine = Array.from({ length: 9 }, (_, i) => `l${i + 1}`);
+    expect(renderProgressRegion(nine)).toBe('\nl4\nl5\nl6\nl7\nl8\nl9');
+    expect(renderActivityReceipt(0, 5_000)).toBe('');
+    expect(renderActivityReceipt(1, 4_000)).toBe('\n\n⏱️ 1 step · 4s');
+    expect(renderActivityReceipt(3, 12_400)).toBe('\n\n⏱️ 3 steps · 12s');
+    // Sub-second turns floor to 1s rather than showing '0s'.
+    expect(renderActivityReceipt(2, 200)).toBe('\n\n⏱️ 2 steps · 1s');
   });
 });
--- a/src/telegram/streaming.ts
+++ b/src/telegram/streaming.ts
@@ -287,23 +287,33 @@ export async function streamResponse(
   let subagentSteps = 0;
   const recentSubagentSteps: string[] = [];
 
-  // Invariant: the `◦` tool-progress region is a BOUNDED, REWRITABLE tail of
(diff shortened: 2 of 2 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

Reviewed against branch HEAD `f56983a6`. Regime: `full` (327 lines, 2 files). Dimensions: security, api-compat, correctness, spec-compliance, test-coverage, perf-observability.

No `critical`, `high`, or `medium` findings. The four claimed invariants were each independently re-derived at the reviewed ref, not taken on the PR description's word.

### Invariant verification (the load-bearing checks)

**1. `accumulated` writer audit — PASS.** Every mutation site enumerated at `f56983a6` and checked against the `progressRunStart` offset invariant documented at `streaming.ts:290-297`:

| Line | Writer | `endProgressRun()` |
|---|---|---|
| 575 | content chunk | precedes (569) ✅ |
| 584 | `stream_retry` rollback | follows (589) — safe, see nit |
| 596 | assistant message | follows (601) — safe, see nit |
| 641 | progress rewrite | n/a — this *is* the mechanism |
| 660 | suggestion | precedes (659) ✅ |
| 758 | `done`-branch flush | post-loop; no further progress events |

No writer is missed. The two "follows" sites are still correct: there is no `await` between the mutation and the later `endProgressRun()` (584→589, 596→601), so the event loop cannot interleave a `progress` event and reach the rewrite at 641 with a stale offset.

**2. `inContentRun = false` stays unconditional — PASS.** `streaming.ts:617` is the first statement in the `progress` branch (608), outside the `progressDelayMs` gate at 638. The render is gated; the `stream_retry` round boundary is not. This is the coupling the PR body flags, and the implementation matches the claim.

**3. Progress-only turn is not stranded — PASS.** `streaming.ts:749-759` flushes the withheld region when `answerText` is empty and `accumulated` is blank. Recording at 633-637 is never gated, so `progressLines` survives a fully-withheld turn.

**4. Receipt kept out of resumable history — PASS.** `streaming.ts:738-740` appends `renderActivityReceipt(...)` only as the argument to `deliverClean`; `answerText` is never reassigned, and `onComplete(answerText, …)` at 765 passes the unmodified value.

### Dimension results

- **security** — no issues found (read `streaming.ts`, `formatter.ts` at `f56983a6`). No new data reaches Telegram: the `◦` line interpolates the same three `event.progress` fields through the same unchanged `sanitizeLabel()` calls as on `main`; only control flow changed. `renderActivityReceipt` (164-187) interpolates two integers. No injection vector — `markdownToTelegramHtml` escapes `&`/`<`/`>` unconditionally before building its own tag set (`formatter.ts:90-93`), and `formatter.ts` is untouched by this diff.
- **api-compat** — no issues found. `renderProgressRegion` / `renderActivityReceipt` are additive exports with no prior consumers. `progressDelayMs?: number` is a new optional field on an already-optional options bag; the sole production caller `src/telegram/handlers/message.ts:865` does not set it and inherits the 5s default. No required parameter changed.
- **correctness** — no issues found beyond the nit below.
- **spec-compliance** — assessed against the PR title + body. No unmet intent; all four described changes are present and behave as described. No scope creep: the diff touches only `streaming.ts` / `streaming.test.ts`, and each item the body declares out of scope (operator config / `ENV_REGISTRY`, `<blockquote expandable>`, the pre-existing empty-`answerText` `onComplete` bug) is in fact left untouched.
- **test-coverage** — the "no assertion was weakened" claim holds. The diff contains zero removed `expect(` lines; both pre-existing tests changed only by appending `, undefined, { progressDelayMs: 0 }` to the call plus explanatory comments — every assertion line is unchanged context.
- **perf-observability** — growth is bounded twice over: `progressLines` is trimmed to `MAX_PROGRESS_LINES` on every push (634-636) and `renderProgressRegion` re-slices to the same bound (165). The per-round rewrite slices `accumulated` instead of appending, so the region keeps fixed height across a long tool-heavy turn.

### Findings

**nit · high confidence · docs-vs-code · `src/telegram/streaming.ts:295-297` · ref:`f56983a6` · citation-type: file-state**
finding: The invariant comment states every other writer must call `endProgressRun()` "first", but the `stream_retry` (584) and assistant-message (596) writers call it *after* mutating `accumulated`, so the comment does not describe the code at those two sites.
evidence:
```
  // appends to `accumulated`, so EVERY other writer (content chunk, assistant
  // message, suggestion, stream_retry rollback) must call `endProgressRun()`
  // first — otherwise the rewrite would truncate their text.
```
suggestion: Reword to "…must call `endProgressRun()` before the next progress render" (which is what actually holds, and is what makes 584/596 safe), or reorder those two sites to match the content-chunk/suggestion pattern.

### CI note

`Test (windows-latest)` is red, but **not because of this PR**. The failures are in `tests/agent/memory/memory-loader.test.ts`, which this diff does not touch, and the same job fails identically on `main` (run `30137517004`, HEAD `62df42c5`). The matrix is `continue-on-error` for non-ubuntu by design (`.github/workflows/ci.yml:83-94`) — ubuntu is the required gate and is green. Not a merge blocker for #702.

### What was not checked

- Read against branch HEAD `f56983a6`; all citations verified at that ref. No `critical`/`high` findings, so the citation-verification and `/shadow-verify` waves were not triggered.
- Stated intent: PR #702 title + body — spec-compliance **was** assessed.
- Tests were read statically, not executed; the PR's own reported run (13091 passed) was not independently reproduced.
- Live Telegram delivery was not exercised (no integration harness), so the 5s gate's real-world feel is unverified — the PR notes this too.
- Per-line byte length of a progress line is not bounded by this change (only line *count* is). Not a regression — the prior append-only path had no length bound either — and `splitLongMessage` handles the 4096 ceiling downstream. Flagged as out-of-scope context, not a finding.

griffinwork40 (commented):

# Code Review — PR #702

**`fix(telegram): bound and latency-gate ◦ tool-progress lines`**
Reviewed ref: `f56983a6` · base `main` · 327 lines across 2 files · regime: full (2 review dimensions + adversarial re-derivation)

## Merge decision: **MERGE** — with one `medium` follow-up noted below

No `critical`, `high`, or blocking findings. The load-bearing invariant this PR rests on was independently verified, not taken on the description's word.

---

## Findings

### 1. `medium` · confidence: high · correctness
**A progress-only turn that exits without a terminal event strands the user on the bare `Thinking…` placeholder — the exact failure the PR's own "dead-turn guard" fixes, but only on the `done` path.**

`src/telegram/streaming.ts:808-820` · ref: `f56983a6` · citation-type: `file-state`

```js
const preview = sentMessage as Message.TextMessage | null;
if (preview && !sawTerminalEvent) {
  const full = cleanFinal && answerText.trim() ? answerText : accumulated;
  if (full.trim()) {
```

The `done` branch got an explicit withheld-region flush (`streaming.ts:749-759`, `else if (progressLines.length > 0)`). The non-terminal post-loop branch did not. With the gate closed, `accumulated` stays empty (`streaming.ts:641` is the only writer and it is inside `if (Date.now() - turnStartedAt >= progressDelayMs)`), so `full.trim()` is false and nothing is delivered; the `else if` at `:820` also requires a truthy `accumulated` and is skipped too. `progressLines` is referenced nowhere after `:758`.

**This is a regression introduced by this PR, not a pre-existing gap.** At base `09a3477e` the progress handler appended unconditionally (`streaming.ts:536-537`, `accumulated += line`), so `accumulated` was non-empty on this path and the block delivered it.

**Trigger conjunction (narrow — why this is `medium`, not `high`):** progress-only turn · every progress event lands inside `progressDelayMs` (5s default) · zero content/message/suggestion events · the provider stream ends by *graceful iterator close* with no `done`/`error`. A watchdog timeout (`:478`) or `error` event (`:775`) throws past this block instead, so those paths are unaffected.

**Suggestion:** mirror the `:749` guard — add `else if (progressLines.length > 0)` to the `:808` branch, feeding `renderProgressRegion(progressLines)` into `full`.

### 2. `nit` · confidence: high · correctness
On the `stream_retry` and assistant-`message` writers, `endProgressRun()` runs *after* the buffer mutation (`:584`→`:589`, `:596`→`:601`) rather than before, unlike the content and suggestion writers. Harmless today — nothing reads `progressRunStart` in between — but it diverges from the "every other writer calls `endProgressRun()` first" wording documented at `:290-297`.

---

## Verified — the invariant this PR rests on

Every writer to `accumulated` clears the progress run in the same handler. Enumerated at the reviewed ref:

| Writer | Site | Clears run? |
|---|---|---|
| content chunk | `:575` | ✅ `endProgressRun()` at `:569` (before) |
| stream_retry rollback | `:584` | ✅ at `:589` (after mutation, same handler) |
| assistant message | `:596` | ✅ at `:601` (after mutation, same handler) |
| progress region rewrite | `:641` | n/a — is the progress writer |
| suggestion | `:660` | ✅ at `:659` (before) |
| done-branch flush | `:758` | n/a — terminal |

**`inContentRun = false` is unconditional** at `streaming.ts:617` — outside the render gate at `:640`, as the PR claims. The retry-boundary coupling is intact.

## Verified — test claims

Checked against the actual test file, not the description:

- Retry-boundary regression asserts the **full** string: `expect(recorded).toBe('AAACCC')` (`streaming.test.ts:1016`) — it fails if the coupling breaks.
- Latency-gate test uses the **production default** — no `progressDelayMs` override (`:964`, comment at `:954`).
- Dead-turn guard present (`:1020`), asserts the placeholder is not the last thing sent (`:1038`).
- Receipt purity asserted: `expect(recorded).not.toContain('⏱️')` (`:1064`).
- Rolling window boundary asserted at `:975-979` (9 rounds, window 6 → keeps 4..9, drops 1..3).
- **"No assertion was weakened" holds mechanically:** the test file deletes exactly 3 lines — one import and two call sites — and the diff removes **zero** `expect`/assertion lines anywhere.

## Verified — security · api-compat

- No secret/token exposure: `renderActivityReceipt` takes only numeric counters (`:184-188`).
- No injection: `renderProgressRegion`/`renderActivityReceipt` are pure string builders; all sends still route through `markdownToTelegramHtml`, and `sanitizeLabel` still runs on progress fields (`:624-626`) — unchanged by this diff.
- No API break: `progressDelayMs?: number` (`:226`) is a new optional field. Sole production caller is `src/telegram/handlers/message.ts:865`, which does not set it and picks up the 5s default — a behavior change the PR discloses explicitly.
- Receipt never reaches persisted history: `:738-739` concatenates for delivery only; `:765` passes the un-concatenated `answerText` to `onComplete`.

## Spec-compliance vs. stated intent

All four numbered changes **MET** (gate `:307`+`:640`; window `:635-636`; dedupe vs `answerText` `:658`; receipt `:738`). Both declared couplings **MET** and genuinely test-guarded. No scope creep found. Declared out-of-scope items (operator config, `<blockquote expandable>`, the pre-existing empty-`answerText` persistence bug) were treated as out of scope, not as defects.

## CI note

`Test (windows-latest)` is red, **but this is pre-existing and not caused by this PR** — the same job fails on `main` (runs `30137517004`, `30128726777`, `30127658645`). Failures are path-separator assertions across ~40 unrelated files (e.g. `expected [ '/base', 'D:\extra\read' ] to include '/extra/read'`). `src/telegram/streaming.test.ts` reports **zero** failures on Windows. Every other check is green.

## What was not checked

- Read against branch HEAD `f56983a6`; all citations verified at that ref, and the base comparison against `09a3477e`.
- Live Telegram delivery — no integration harness (the PR discloses this).
- The tests were not executed locally; CI results on ubuntu/macOS were taken as the signal.
- Stated intent: PR #702 title + body — spec-compliance assessed.

griffinwork40 (commented):

## Merge decision: **MERGE** — 2 non-blocking notes (1 low, 1 nit)

Reviewed ref: `022d275` (branch HEAD). Regime: full (464 lines, 2 files). Stated intent: PR #702 title + body — spec-compliance assessed.

Reviewed as a footer-design change, not the splice-at-offset design the PR body describes — see nit 2. Zero blocking, critical, or high findings.

---

### Verdicts by dimension

| Dimension | Verdict |
|---|---|
| security | clean — read `streaming.ts` + `formatter.ts` at `022d275` |
| api-compat | clean — one additive optional param, two additive test-only exports |
| correctness | clean — all four `accumulated` writers and both terminal delivery paths traced |
| spec-compliance | pass — four stated changes implemented, no scope creep, no unmet clause |
| test-coverage | pass — 44/44 green locally; both named guards mutation-verified |
| perf-observability | pass — buffer bounded at 6, one forced edit per turn |

### The three load-bearing invariants — all hold

**1. Progress region cannot corrupt `accumulated` — PASS (resolved by construction).** `progressLines` is a separate array (`streaming.ts:302`) rendered only through `renderProgressRegion`/`finalBody`/`livePreview` (`streaming.ts:427,432-435`) and never spliced into `accumulated`. All four `accumulated` writers — content chunk (`:593`), `stream_retry` rollback (`:602`), assistant-message replace (`:613`), suggestion append (`:692`) — are structurally unable to invalidate an offset, because no offset exists. This is a stronger guarantee than the call-site contract described in the PR body.

**2. `inContentRun = false` stays unconditional — PASS, mutation-verified.** Confirmed at `streaming.ts:631`, ahead of the gate check at `:653`. I re-derived the guard by deliberately breaking it (wrapping the assignment in `if (progressGateOpen)`): the test `REGRESSION: a gated-off progress round still sets the stream_retry boundary…` fails (1 failed / 43 passed) and passes again on revert. The guard is real, not decorative.

**3. Progress-only turns never strand the user — PASS, mutation-verified.** `finalBody()` (`streaming.ts:427`) is ungated, unlike `livePreview()` (`:432-435`); both the `done` else-branch (`:781-788`) and the non-terminal post-loop exit (`:838-852`) route through it. Gating `finalBody` on `progressGateOpen` fails 4 tests, including `dead-turn guard: a fast progress-only turn still delivers…` and `delivers a gated progress-only turn that ends WITHOUT a done event`.

Intent #4 also verified: the receipt is appended at delivery only (`streaming.ts:768-769`) and never to `answerText`, so it stays out of resumable history.

### Findings

**low · high confidence · test-coverage · `src/telegram/streaming.test.ts:236-254` · ref:`022d275` · citation-type: file-state**
The pre-existing test `does not flood the Telegram API with empty progress updates` no longer exercises the throttle path its comment documents: the new 5s default gate suppresses all four progress renders before `EDIT_THROTTLE_MS` throttling applies, so `expect(edits.length + replies.length).toBeLessThanOrEqual(5)` now passes for a different reason than stated.
```js
    await streamResponse(ctx, session, 'go');
    expect(edits.length + replies.length).toBeLessThanOrEqual(5);
```
This test is untouched by the diff (verified absent from the PR diff), so throttle coverage is silently lost rather than changed. Fix: pass `progressDelayMs: 0` here as the two adapted tests do, which restores the throttle assertion's original meaning.

**nit · high confidence · documentation · PR description §"Two couplings worth reviewer attention" · ref:`022d275`**
The PR body documents an invariant that no longer exists in the code being merged: *"The `progressRunStart` offset is only valid while nothing else appends to `accumulated`, so every other writer … calls `endProgressRun()` first."* At `022d275`, `git grep -n -E "endProgressRun|progressRunStart"` over `src/telegram` returns zero matches; the same grep at `f56983a6` returns 12. Commit `022d275` ("make the ◦ progress region a single globally-bounded footer") replaced the mechanism, which is the better design — but the body still describes the superseded one, and the body becomes the durable record. Fix: update that paragraph to describe the footer design (separate bounded array, no offset) before merge.

### CI status — the red Windows leg is not this PR

`Test (windows-latest)` is FAILURE, and it is unrelated to this change:
- The leg is advisory by design — `continue-on-error: ${{ matrix.os != 'ubuntu-latest' }}` (`.github/workflows/ci.yml:94`); `ubuntu-latest` is the required gate and it is green.
- The failures are systemic Windows path-portability breakage across 85 test files (`D:\repo\...` vs `/repo/...`, `file:///D:/...` URLs, `\tmp\artifacts\...`), none in the two changed files.
- `src/telegram/streaming.test.ts` appears in zero FAIL lines on that run.

Green on the required legs: Lint & Build, Test (ubuntu-latest), Test (macos-latest), Test (real PTY scrollback), Docs Site, Publish Bundle Smoke.

### Verification performed

- Ran `npx vitest run src/telegram/streaming.test.ts` at `022d275`: **44 passed / 44**.
- Mutation-tested both author-flagged couplings (results above); working tree restored clean after each.
- Reachability pre-check on `streamResponse`: sole production importer `src/telegram/handlers/message.ts:865` passes no `progressDelayMs`, so the new optional field is non-breaking. `renderProgressRegion`/`renderActivityReceipt` have zero production importers `[UNVERIFIED: no production importers]` — additive test surface, matching the existing `renderSubagentFooter` pattern.
- Security: no bot-token/chat-id reference in either changed file; model-controlled `description`/`summary`/`lastToolName` pass through `sanitizeLabel` (`streaming.ts:638-640`); both new render helpers emit plain text and flow through `markdownToTelegramHtml` before send; `progressLines` is double-bounded (`:649-652` and `:165`) and only one `progressTimer` arms at a time (`:659`).

### What was not checked

- Read against branch HEAD `022d275` — all citations verified at that ref.
- Did not run the full suite or the coverage gate locally; relied on the PR's CI for those (ubuntu green).
- Did not verify live Telegram delivery, the real 5s wall-clock gate, or rendering in a real client — no integration harness exists, as the PR body states.
- Did not review the pre-existing `onComplete` empty-`{assistant: ''}` persistence bug the body defers; out of scope for this diff.
- Did not assess the unrelated Windows path-portability failures beyond confirming they exclude the changed files.

griffinwork40:
All waves complete. Independent re-derivation done on the load-bearing claims.

---

# Code Review — PR #702

**Target:** `fix(telegram): bound and latency-gate ◦ tool-progress lines` · `main` ← `fix/telegram-progress-line-noise`
**Reviewed ref:** `022d275194bf27b868cc045c03a80b7d05dd5086` (2 commits: `f56983a` + `022d275`)
**Regime:** full · 2 files · +443/−21 (464 lines) · change type: bugfix (user-facing behavior)
**Stated intent:** PR #702 title+body **+** commit `022d275` message — spec-compliance assessed.

---

## Findings

### security — no issues found
Read `src/telegram/streaming.ts:155-190,620-700` at `022d275`. All three `event.progress` fields consumed (`description`, `summary`, `lastToolName`) route through `sanitizeLabel` before interpolation (`streaming.ts:633-641`). The other four `ProgressEvent` fields (`taskId`, `totalTokens`, `toolUses`, `durationMs`) are never referenced — grep returned 0 matches, so no raw-field path exists. `renderActivityReceipt` interpolates only numbers (`streaming.ts:184-187`). Diff grep for secret/token patterns and absolute `/Users|/home` paths: 0 matches.

### api-compat — no issues found
Export delta is **additive only**: `renderProgressRegion` and `renderActivityReceipt` added; `StreamTimeoutError`, `replyWithFloodRetry`, `renderSubagentFooter`, `streamResponse` unchanged. No removed or signature-changed exports → no reachability downgrades needed. `progressDelayMs?: number` is optional (`streaming.ts:226`); the sole production caller `src/telegram/handlers/message.ts:865` omits it.

**PR claim #4 verified independently by me, two paths:** `streaming.ts:773` builds `answerText + renderActivityReceipt(...)` *only as the `deliverClean(...)` argument*; `streaming.ts:793` passes the unmodified `answerText` to `onComplete`. The `⏱️` receipt never enters resumable session history. Grep for `⏱️`/`◦` consumers across `formatter.ts`, `push.ts`, `notify-routing.ts`, `session-manager.ts`: 0 matches — no downstream text-parser to break.

### correctness — no issues found
| Terminal/fallback path | file:line | Body delivered | Empty on gated progress-only turn? |
|---|---|---|---|
| `done`, cleanFinal + answer | 762-777 | `deliverClean(answerText + receipt)` | No (branch requires non-empty answer) |
| `done`, else | 781-787 | `sendOrEdit(finalBody(), true)` | No — ungated |
| `error` | 800-810 | none (rethrows) | N/A |
| post-loop, `!sawTerminalEvent` | 838-852 | `answerText` or `finalBody()` | No — ungated leg |
| post-loop overflow chunks[1..] | 853-894 | `splitLongMessage(finalBody())` | No — same source |
| `finally` / outer `catch` | 896-934 | cleanup only | N/A |

No path delivers bare `accumulated` (grep for `sendOrEdit(accumulated`/`deliverClean(accumulated`: 0 code matches; only a comment mentions it). **Commit `022d275`'s central claim holds.**

**Timer lifecycle sound.** `clearProgressTimer()` on all three exits (757 done / 805 error / 924 finally); `turnEnded = true` set at 756, 804, and — per the documented ordering invariant at 921-922 — *before* the clear at 923. Callback checks `turnEnded` first (667), so an armed timer cannot edit a finished turn. Not `.unref()`'d, but bounded by `progressDelayMs` (≤5s) and cleared in `finally`, so it cannot outlive the call.

**Global bound holds by construction.** `progressLines` is one function-local array trimmed at push (650-652) *and* re-sliced at render (165) — the interleaved-stream growth that motivated commit 2 is structurally impossible. Worst case for both status regions: **17 physical lines** (6 progress entries × 2 lines + 1 subagent header + 4 subagent lines). Bounded.

**`splitLongMessage` alignment correct.** Both the `done` edit (787) and the overflow split (874) evaluate `finalBody()`, and nothing reassigns `accumulated`/`progressLines` between loop-exit and 874 — chunk[0] matches what is already on screen. The in-code comment at 871-873 matches the code.

**`progressRounds` faithful for anthropic-direct** (1:1 with `iterations += 1` at `loop.ts:1106` → single `yield` at 1118-1134; `stream-consumer.ts:396-410` is a passthrough). `unverified` for openai-compatible (`query.ts:586` loop structure not read).

### spec-compliance — 1 finding

```
low · high · spec-compliance · PR #702 body (lines ~23, coupling §2) · ref:022d275 · citation-type:file-state
finding: The PR body still describes commit f56983a's rejected design — it documents the
  `progressRunStart` offset invariant and names `endProgressRun()`, both DELETED by 022d275,
  and states "the `done` branch now flushes the withheld region" which is the exact bug
  022d275 fixed by routing every terminal path through `finalBody()`.
evidence: body:23 "The `progressRunStart` offset is only valid while nothing else appends to
  `accumulated`, so every other writer … calls `endProgressRun()` first. That invariant is
  documented at the declaration."
  git grep -c 'progressRunStart\|endProgressRun' 022d275 -- src/  →  0 matches
suggestion: Edit the PR body's mechanism paragraph and coupling §2 to describe the shipped
  footer design (`finalBody()` on all terminal paths), since squash-merge persists the body
  into commit history.
```

Why `low`, not higher: the **in-code** invariant comments are accurate — I read `streaming.ts:290-300` and `420-427` and both correctly describe the shipped footer design and explicitly narrate the offset design as rejected. The staleness is confined to the PR body. No runtime behavior is wrong.

No scope creep found. Every body claim #1–#7 matches the code at `022d275` except the two mechanism sentences above. The "fast turns render zero `◦` lines" claim is internally scoped by "straight from `Thinking…` to the answer" and the body's own coupling §2 discloses the progress-only flush — consistent, not a defect.

### test-coverage — 1 finding

8 of 9 claimed guards verified **PRESENT** by reading `streaming.test.ts` at `022d275`: retry-boundary regression (`:990`, asserts `toBe('AAACCC')`), dead-turn guard (`:1020`), latency gate at production default with no override (`:950`), interleaved 9-round bound (`:1121`), timer-opened gate (`:1149`), no-`done` progress-only turn (`:1171`), receipt present/absent (`:1041`, `:1067`), suggestion dedupe (`:1078`).

**No assertion was weakened.** Only 3 removed lines, all call-signature changes gaining `progressDelayMs: 0`; every downstream assertion is byte-identical. Author's claim confirmed.

```
low · high · test-coverage · src/telegram/streaming.test.ts · ref:022d275 · citation-type:file-state
finding: No test arms a fake clock and advances past turn-end to assert the queued
  progressTimer callback is inert — the `turnEnded` guard (streaming.ts:667) is unexercised.
evidence: git grep -n 'turnEnded\|does not fire after\|no stray' 022d275 -- src/telegram/streaming.test.ts → 0 matches
  git grep -ni 'fire\b' … → 4 matches, all unrelated retry/watchdog contexts
suggestion: Add a case with `vi.useFakeTimers()` that ends the turn with a timer armed, advances
  past progressDelayMs, and asserts no additional editMessageText call lands.
```

### perf-observability — no issues found
The bounded region strictly reduces per-edit work: `markdownToTelegramHtml`'s 8 serial regex passes (documented at `streaming.ts:320`) now run over a capped-height region instead of an append-only log, and the latency gate eliminates edits entirely on fast turns. New per-event allocation is one `slice(-6)` — negligible against the regex passes it replaces.

### nit

```
nit · high · correctness · src/telegram/streaming.ts:668-674 · ref:022d275 · citation-type:file-state
finding: The timer callback's comment claims it skips "a render already in flight", but
  `editInFlight` is never set by `sendOrEdit` (grep: set only at 673 and 732, the timer and
  pause-countdown paths) — so a concurrent main-loop `await sendOrEdit(livePreview())` is
  undetected and both edits race the same message.
evidence: // A render already in flight computed its text before the gate
  // opened; skip rather than racing it …
  if (editInFlight) return;
suggestion: Reword the comment to name what the flag actually guards (timer-driven renders),
  or set editInFlight inside sendOrEdit itself.
```
Impact is a transient visual flicker that self-heals on the next event or final delivery — the region is never lost, since `finalBody()` is ungated.

---

## Verification manifest (Wave 1.5)

- **Check A (citation verification):** no `critical` or `high` findings were produced, so no citations required re-derivation at that tier. All surviving findings' citations were read at `022d275` by at least one agent; I re-read the three load-bearing ones myself (deleted symbols, `onComplete` argument, sole production caller) — 0 `fabricated`, 0 `diff-only`.
- **Check B (absence claims):** 2 absence claims survived. Both `confirmed-absent` with cited greps (`progressRunStart|endProgressRun` → 0; timer-inertness test patterns → 0). I independently re-ran the first: `git grep -c … 022d275 -- src/` → 0 matches.
- **Dropped:** 1 sub-agent finding (`medium`, `sendOrEdit` swallows non-parse-entity edit errors, `streaming.ts:355-370`) — the cited code is **outside this diff's hunks** and pre-existing on `main`. Out of scope for this PR; worth its own issue.
- **Downgraded:** 1 sub-agent `medium` spec-compliance finding ("zero ◦ lines not universally true") — refuted by the body's own coupling §2, which discloses the progress-only flush. Folded into the single stale-body finding.
- **`/shadow-verify` skipped:** triggered only by `critical`/`high` findings; none present.

## What was not checked

- Wave 1 agents read against branch HEAD **`022d275`** — all citations verified at that ref. No patch-file input; live ref available throughout.
- **Stated intent:** PR #702 title+body **plus** commit `022d275` message — spec-compliance **was** assessed against both.
- Test suite **not run** (read-only constraint); the author's `13094 passed / 699 files` claim is unverified here.
- Live Telegram delivery, `deliverClean`/`splitLongMessage` internal math, `markdownToTelegramHtml` escaping beyond its documented contract.
- `progressRounds` cardinality for the **openai-compatible** provider (`src/agent/providers/openai-compatible/query.ts:586`) — emit-site loop not read.
- CI status, `pnpm lint`/`build`, and the env/sdk audit gates the author claims pass.
- **Process note:** the two broad Wave 1 agents failed (one 45-min timeout, one truncated stream); coverage was recovered via five narrow budget-bounded `compose` nodes plus my own inline re-derivation. One node (`bound-and-receipt`) hit its call cap; its questions were re-covered by a replacement narrow agent and my direct reads.

---

Zero `critical`, zero `high`. The change is well-constructed: the global bound holds structurally, every terminal path routes through `finalBody()`, the timer is cleared on all three exits with a correct `turnEnded` ordering invariant, and the two couplings the author flagged both have real regression tests that assert what the body says they assert. No assertion was weakened. The three open items — a stale PR body, one missing timer-inertness test, one inaccurate comment — are all non-blocking and independently fixable after merge.

**Done**
- Reviewed PR #702 at ref `022d275` across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability; emitted 2 `low` findings, 1 `nit`, and a **MERGE** recommendation.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/5b59cec3-92dd-4b55-a4ec-b1c66c68c89a/pr-702.diff` (verified byte-identical to live `gh pr diff 702`); all citations read from `git show 022d275:...`; session witness trace at `~/.afk/state/witness/<session>/trace.jsonl`.
- World unchanged — review is read-only: no files written, no commits, no pushes, no PR comments or body edits. Working tree remains clean at `022d275`.
- Deferred: the pre-existing `sendOrEdit` silent-edit-failure path (`streaming.ts:355-370`) is out of this diff's scope and merits its own issue; `progressRounds` cardinality under the openai-compatible provider is unverified; the author's test-suite pass claim was not re-run.

---

PR #712, opened 2026-07-26
feat(usage): show subscription quota in the REPL status line

griffinwork40:
> **Stacked on #711.** Base branch is `afk/usage-subscription-command`, not `main` — review #711 first. GitHub will retarget this to `main` automatically when #711 merges.

## Why a second mechanism

#711 added `/usage`, which **polls** `GET /api/oauth/usage` when the operator asks. That's right for a command but wrong for a persistent indicator — it would mean a request per repaint.

Anthropic returns `anthropic-ratelimit-unified-5h-utilization` / `-7d-utilization` headers on **every** response under subscription OAuth auth. So the status line rides on responses the SDK already made: **zero extra requests, zero model tokens.**

Result — a segment on the REPL status line:

```
~/projects/agent-afk  ⎇ main  ▁▂▃ 12%  $0.42  1.2k tok  5h 62% · 7d 31%
```

## How

- **`src/agent/quota-cache.ts`** (new) — best-effort header parsing into a module-scope latest-snapshot cache with a change subscription. These headers are **undocumented and have been observed to change names upstream**, so every field is optional, every value is validated, and a malformed one is omitted rather than thrown. `parseQuotaHeaders` returns `undefined` when neither utilization is present — nothing worth caching.
- **`tracing-fetch.ts`** — new `onQuota` observer, invoked for **all** responses (before the throttle-status gate, so 2xx included) and guarded in `try/catch` exactly like `onThrottle`, per that file's own contract that a throwing callback must never disturb the request path or the SDK retry loop.
- **`status-line.ts`** — quota segment at `droppablePriority: 5`, rendered rightmost.
- **`interactive/shared.ts`** — `formatStatusFields` reads the cache and formats `5h 62% · 7d 31%`.
- **`interactive/surface-setup.ts`** — subscribes a repaint, mirroring the git-status sampler's `onUpdate` wiring.

## Three decisions worth a reviewer's attention

**1. The wrapper is now installed for every non-local-shim client, not just when a trace writer is attached.** This is the deliberate behaviour change in this PR. Previously `makeTracingFetch` was gated on `!localMode && config.traceWriter`; quota capture must survive `AFK_TRACE_DISABLED=1`, so the gate is now `!localMode` alone. **Both** install sites are updated, including the OAuth-token-refresh rebuild path. `localMode` is still excluded — a local shim isn't Anthropic and emits no quota headers.

This is what changed two existing assertions in `anthropic-direct.test.ts`: the SDK ctor arg now carries a `fetch`. I tightened those tests rather than loosening them — they now assert the **exact key set** plus types, so an unintended third key still fails.

**2. `droppablePriority: 5` is not a free choice.** The shed loop does `parts.filter(p => p.droppablePriority === undefined || p.droppablePriority !== maxPriority)` — it removes **every** part sharing the current maximum in one pass. Reusing the token count's `4` would make quota and tokens vanish *together* instead of shedding one at a time. There is an explicit regression test for this (`drops quota BEFORE the token count, and dropping quota does not also drop tokens`).

**3. Repaints fire only when a rounded percentage changes.** `repaint()` throttles at 100 ms and stores-without-painting with no trailing flush, so notifying per API response would churn the bottom row and could drop the final update. The cache gates notification on `Math.round(u * 100)` changing, while always keeping the stored snapshot current for the next poll.

## Graceful degradation

Under **API-key auth these headers never arrive**, so the cache stays empty for the process lifetime and the status line draws **no segment at all** — not a placeholder, not a zero. Same for a renamed header upstream. Tested explicitly.

## Verification

| gate | result |
|---|---|
| `pnpm lint` (tsc --noEmit, strict) | ✅ clean |
| `pnpm test` | ✅ **704 files, 13,188 passed**, 0 failed |
| `pnpm test:coverage` | ✅ 87.12 / 85.21 / 91.27 / 87.12 (floors 74/79/82/74) |
| `pnpm audit:sdk:check` | ✅ |
| `pnpm audit:env:check` | ✅ |
| `pnpm scan:env:check` | ✅ |
| `pnpm audit:chalk:check` | ✅ |

`quota-cache.ts` and `tracing-fetch.ts` both at **100%** coverage. **42 new tests**: header parsing (both windows, each alone, neither, case-insensitivity, non-numeric, out-of-range clamping, negative/absurd/non-numeric reset epochs, a hostile headers object), cache/subscription semantics (first-snapshot notify, rounded-change notify, **no**-notify on unchanged rounding, unsubscribe, throwing listener isolation), the `onQuota` wrapper-install guard, and status-line rendering including the drop-priority regression and a units guard (0.62 → `62%`, never `0%` or `6200%`).

No new env var, no new dependency, no `process.env` read, no direct chalk import.

## Still open from #711

The endpoint/header contract remains **reverse-engineered, not documented**. The verification command in #711's description applies here too — in particular whether `utilization` is a 0–1 fraction (assumed, and clamped to that range) or a 0–100 percentage. A unit mismatch is the one failure mode that renders plausible-looking wrong numbers instead of degrading to no segment.

## Not in scope

- Telegram has no persistent status surface, so this is REPL-only. `/usage` from #711 covers Telegram.
- `rate_limit` `durationMs` still stores an advertised hint rather than a measurement.
- `usage_limit_pause`/`_resume` still aren't aggregated in `afk trace show` (~16–20 lines; cheapest remaining win).
- `AFK.md:51` still claims a usage-limit pause has no event — stale.

### DIFF
--- a/src/agent/providers/anthropic-direct.test.ts
+++ b/src/agent/providers/anthropic-direct.test.ts
@@ -436,7 +436,14 @@ describe('AnthropicDirectProvider', () => {
 
     expect(anthropicCtorMock).toHaveBeenCalledTimes(1);
     const ctorArg = anthropicCtorMock.mock.calls[0]?.[0] as Record<string, unknown>;
-    expect(ctorArg).toEqual({ authToken: 'sk-ant-oat01-test' });
+    // `fetch` is expected: the tracing/quota wrapper is installed for EVERY
+    // non-local-shim client, no longer only when a trace writer is attached,
+    // because subscription-quota capture must keep working with tracing off
+    // (see the install site in providers/anthropic-direct/index.ts). Asserting
+    // the exact key set keeps this strict — an unintended third key still fails.
+    expect(Object.keys(ctorArg).sort()).toEqual(['authToken', 'fetch']);
+    expect(ctorArg['authToken']).toBe('sk-ant-oat01-test');
+    expect(typeof ctorArg['fetch']).toBe('function');
     expect('apiKey' in ctorArg).toBe(false);
 
     expect(messagesCreateMock).toHaveBeenCalledTimes(1);
@@ -465,7 +472,14 @@ describe('AnthropicDirectProvider', () => {
     await collect(query);
 
     expect(anthropicCtorMock).toHaveBeenCalledTimes(1);
-    expect(anthropicCtorMock.mock.calls[0]?.[0]).toEqual({ apiKey: 'sk-ant-api03-test' });
+    // `fetch` is expected here too — the wrapper is installed for every
+    // non-local-shim client so quota capture survives tracing being disabled.
+    // Under API-key auth the quota headers never arrive, so the cache simply
+    // stays empty and the status line draws no quota segment.
+    const apiKeyCtorArg = anthropicCtorMock.mock.calls[0]?.[0] as Record<string, unknown>;
+    expect(Object.keys(apiKeyCtorArg).sort()).toEqual(['apiKey', 'fetch']);
+    expect(apiKeyCtorArg['apiKey']).toBe('sk-ant-api03-test');
+    expect(typeof apiKeyCtorArg['fetch']).toBe('function');
 
     const [params, opts] = messagesCreateMock.mock.calls[0] as CreateArgs;
     expect(opts?.headers?.['anthropic-beta']).toBeUndefined();
--- a/src/agent/providers/anthropic-direct/index.ts
+++ b/src/agent/providers/anthropic-direct/index.ts
@@ -36,6 +36,7 @@ import {
 import { oneShotCompletion, type OneShotInput } from './oneshot.js';
 import { makeTracingFetch } from './tracing-fetch.js';
 import { ThrottleQueue } from './throttle-queue.js';
+import { parseQuotaHeaders, recordQuotaSnapshot } from '../../quota-cache.js';
 import { refreshClaudeCodeOauthToken } from '../../auth/keychain.js';
 import { AnthropicDirectQuery } from './query.js';
 import { pathContainmentBypassed } from '../../permission-policy.js';
@@ -636,6 +637,18 @@ export class AnthropicDirectProvider implements ModelProvider {
     // query below so the fetch producer and the loop consumer meet.
     const throttleQueue =
       !localMode && config.traceWriter ? new ThrottleQueue() : undefined;
+    // Invariant: quota capture is gated on `!localMode` ALONE — deliberately not
+    // on `config.traceWriter`. The `anthropic-ratelimit-unified-*` headers ride
+    // on every response and feed the CLI status line, which must stay live even
+    // when tracing is off (`AFK_TRACE_DISABLED=1`); tying it to the trace writer
+    // would silently blank the indicator in that configuration. localMode is
+    // still excluded: a local shim is not Anthropic and emits no quota headers.
+    const quotaObserver = localMode
+      ? undefined
+      : (headers: Headers): void => {
+          const snapshot = parseQuotaHeaders(headers);
+          if (snapshot !== undefined) recordQuotaSnapshot(snapshot);
+        };
     const clientOpts = buildClientOptions(
       token,
       authMode,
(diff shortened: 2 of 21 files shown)


griffinwork40:
# Review complete — PR #712

**Target:** PR #712 `feat(usage): show subscription quota in the REPL status line` (branch `afk/usage-statusline-indicator` → base `afk/usage-subscription-command`, stacked on #711)
**Reviewed ref:** `e2125838bf8fe158b6d3d9aac96011a6d66eb830` (confirmed == local worktree HEAD, clean, matches origin)
**Regime:** full (635 lines changed, 11 files, feature type) → 2 parallel Wave-1 agents (security·api-compat / correctness·spec-compliance·test-coverage·perf-observability) → Wave-1.5 gate (scoped to blocking/critical/high; found none, so structurally a no-op — I additionally spot-verified 5 of the 10 medium/low findings myself via direct read/grep/`node -e`, all confirmed accurate) → Wave-2 synthesis.

---

# Code Review — PR #712: `feat(usage): show subscription quota in the REPL status line`

## Summary

| Field | Value |
|---|---|
| **Findings** | 10 total — 0 blocking · 0 critical · 0 high · **2 medium** · **5 low** · **3 nit** |
| **Cross-agent conflicts** | 0 |
| **Merge decision** | ✅ **MERGE** (with 3 cheap pre-merge follow-ups, non-blocking) |

## Findings (sorted, full schema)

**🟠 MEDIUM**
1. **F1 · correctness · `quota-cache.ts:20-25`** — empty/whitespace header value parses to `0`, not `undefined` → renders spurious `5h 0%`, contradicting the PR's own "not a placeholder, not a zero" claim. `Number('')===0` is finite, passes the guard. **Verified via `node -e`.** Fix: `if (raw === null || raw.trim() === '') return undefined;`
2. **F2 · spec-compliance · `quota-cache.ts:19-25`** — `Math.min(1, …)` saturates any ≥1 reading to a pinned `100%` if upstream units turn out to be 0–100 not 0–1 — exactly the "plausible-looking wrong numbers" risk the PR body itself flags as still-open from #711, and outside the stated degradation guarantee (which only covers absent/renamed headers). Needs an explicit author decision, not silence.

**🟡 LOW**
3. **F3 · security · `quota-cache.ts:88-95,112-133`** — unkeyed module-scope cache slot; cross-account co-residency reachability genuinely unresolved (out of this diff's scope). Impact bound: one rounded integer, no credential exposure.
4. **F4 · api-compat · `shared.ts:787-838`** — `formatStatusFields` is no longer pure (reads global `getQuotaSnapshot()`). **Verified 7 production call sites via grep** (surface-setup.ts×3, bootstrap.ts×2, resume-swap.ts, loop-iteration.ts).
5. **F5 · test-coverage · `surface-setup.ts:399-401`** — `onQuotaUpdate` unsubscribe discarded → listener leak across repeated `setupSurface` calls; unexercised by `loop-iteration.test.ts`/`repl-loop-wiring.test.ts`. **Verified via grep.**
6. **F6 · perf-observability · `quota-cache.ts:64-79`** — reset-timestamp fields parsed/allocated on every response with zero consumers.
7. **F7 · test-coverage** — PR body claims "42 new tests"; actual diff adds 35. **Verified via `it(` count.**

**⚪ NIT**
8. **F8 · security** — widened install gate routes headers into parsing on every request incl. API-key auth, but path is verified closed (numeric-validated, never raw text reaches render). Logged as a checked-and-closed non-issue.
9. **F9 · api-compat** — injected `clientFactory` shape widens by one key; zero production importers, hook already documented as unstable.
10. **F10 · correctness** — drop-ordinal comments in `status-line.ts` drifted after priority-5 insertion (`cost` and `tokens` both now say "drop 2nd"). **Verified via direct read.**

**Verified non-issues (positive evidence, both agents):** no credential/header/body leakage anywhere in the new code; `onQuota` is a purely additive trailing param; `StatusLineFields.quota?` is purely additive; the two tightened test assertions strengthen (not loosen) the OAuth-no-apiKey guarantee; `droppablePriority` uniqueness holds; all 11 changed files trace to the stated intent with **zero scope creep**.

**Genuinely unresolved (not closable from this repo):** upstream header units (0–1 vs 0–100, gates F2); cross-account co-residency (gates F3); whether Anthropic's SDK independently honors `ANTHROPIC_BASE_URL` outside AFK's own `localMode` check; live coverage-tool execution (claims are plausible but unexecuted).

## Merge decision: ✅ MERGE

No finding reaches high/critical; F1 and F2 are individually and jointly bounded (a narrow empty-string edge case with a 3-character fix, and a conditional risk on an unresolvable upstream premise). Recommended before merge: fix F1, make an explicit call on F2 (accept-and-document vs. omit-on-ambiguity), correct the "42" → "35" test count in the PR body. F3–F6, F9, F10 are follow-up-grade.

---

**Done**
- What was done: full-regime read-only review of PR #712 executed per the `/review` skill — target resolved, stated intent captured from the PR body, 2 parallel Wave-1 dimension agents dispatched, Wave-1.5 verification gate applied (structurally empty — no blocking/critical/high findings existed) plus voluntary orchestrator-level spot-verification of 5 medium/low findings, Wave-2 synthesis produced the final sorted report and merge decision above. Post-synthesis `/shadow-verify` was not triggered because no critical/high finding was present (per the skill's own gate).
- Evidence: this response *is* the deliverable — the skill's hard read-only constraint explicitly forbids writing any file (repo or otherwise), so no report file was created; the diff artifact consulted was `~/.afk/state/skill-preflight/0b5eeb3c-1a67-4bb7-82b6-874d6b1fe67a/pr-712.diff`, and all citations are keyed to reviewed ref `e2125838bf8fe158b6d3d9aac96011a6d66eb830`.
- What changed in the world: nothing — no file, git, or PR/MR mutation occurred (verified: `git status --short` was clean before and no write/edit/commit/push/gh-mutate tool was invoked at any point).
- Pending/deferred: the two medium findings (F1, F2) and the "42→35" test-count correction are recommended before merge but are the human's call to action on, not mine — this skill stops at the merge recommendation by design.

---

PR #713, opened 2026-07-26
feat(telemetry): record stop_reason on subagent terminal rows and flag partial results

griffinwork40:
## Why

Two prior fixes for the "subagent returns a partial result" class — #628 (zero-output streams now fail instead of silently succeeding) and #678 (bounded stream-incomplete retry + handoff contract) — **looked successful for weeks because no discriminating signal was ever recorded.**

Measured against `~/.afk/agent-framework/routing-decisions.jsonl`:

- `stop_reason` appears on **0 of 14,152** `subagent.completed` rows.
- **55 of 119** sub-200-char completions had run **≥10 tool calls** — worst cases 176, 166 and 148 tool calls returning 115–193 chars.
- Rate of that shape rose from a ≤0.49%/week baseline to **1.54%** in 2026-W27.

A short subagent result is currently indistinguishable across three paths:

| Path | `stopReason` | Surfaces as |
|---|---|---|
| stream cut off mid-flight, some text buffered | `stream_incomplete` | `succeeded` |
| tool-loop cap fired → wind-down synthesis | `tool_use_loop_capped` | `succeeded` |
| model genuinely answered briefly | *(none)* | `succeeded` |

`#628`'s guard genuinely holds — zero-char completions are 0/14,152 — but **near-zero was never guarded**, and nothing recorded which of the three paths produced a given short answer. This PR makes them separable.

## What changed

**1. `stop_reason` on terminal telemetry rows.** Added `RoutingDecisionEntry.stop_reason?: string` (snake_case, matching the `duration_ms` / `content_chars` siblings), omitted when absent — never `null`/`''`. Populated at every terminal emit site: `background-registry.ts` `markTerminal` (all three of completed/failed/cancelled) **and** `foreground-promotion.ts` (completed + the in-band failed emit), since the large majority of dispatches flow through the foreground path. The catch-block failed emit has no `SubagentResult` in scope and is left as-is.

**2. Partial banner on the `/bgsub:join` replay path.** `bgsub.ts` `printJobDetail` read `result.message.content` raw, so a manual join replayed a truncated background result with no marker. It was the only delivery consumer lacking one — the auto-injection path (`bg-result-notifier.ts:91`) already annotates.

> `markTerminal` deliberately does **not** annotate: it is an observation site that never mutates `result`, and `bg-result-notifier` already annotates the same `job.result` at injection time. `annotateIfIncomplete` is not idempotent, so annotating in both places would double-prepend the banner.

**3. Structured `incomplete` flag alongside the prose banner.** `ToolResult.incomplete?: boolean` + `incompleteReason?: string`, spread via a new `incompleteToolResultFields()` helper derived from the existing `isIncompleteStopReason` (predicate not duplicated).

- Named `incomplete`, **not** `truncated` — `ToolResult.truncated` already means tool-**output** byte-cap truncation (bash/grep/web-scrape) and is orthogonal; a subagent result can be incomplete while far under any byte cap.
- The `[⚠ PARTIAL RESULT…]` prose banner **stays** — existing consumers depend on it as the in-band model-visible signal. The flag is additive, so non-model consumers (trace readers, hooks, calling code) can branch without substring-matching banner text.
- Applied at both `ToolResult`-bearing delivery sites (`foreground-promotion`, `fork-dispatch`). `dag-subagent` returns a bare string rather than a `ToolResult`, so it correctly keeps the prose banner only.

**4. Tests.** Extended the existing `describe('routing-telemetry events (A1)')` block in `background-registry.test.ts` (stop_reason present/absent across all three terminal branches; `content_chars` unaffected by annotation) and added 2 cases to `routing-telemetry.test.ts` pinning carry + omission against the real `buildRoutingDecisionRow`.

> Key-absence is asserted in `routing-telemetry.test.ts` rather than `background-registry.test.ts` because the latter mocks `appendRoutingDecision` wholesale, so the pre-strip entry still carries the key with an `undefined` value. The real `buildRoutingDecisionRow` strips undefined fields (`if (v !== undefined) row[k] = v`), which is where that invariant belongs.

All changes are additive and backward-compatible; no existing consumer behavior changes.

## Verification

| Gate | Result |
|---|---|
| `pnpm lint` | clean (`tsc --noEmit`) |
| `pnpm test src/agent/routing-telemetry.test.ts` | `Tests  9 passed (9)` |
| `pnpm test src/agent/background-registry.test.ts` | `Tests  53 passed (53)` |
| `pnpm test src/agent/subagent/result.test.ts` | `Tests  19 passed (19)` |
| `pnpm test src/agent/tools/subagent/foreground-promotion.test.ts` | `Tests  5 passed (5)` |
| `pnpm build` | clean (`tsc` + `copy-prompts`) |
| `pnpm test` (full) | **`Test Files 701 passed (701)` / `Tests 13124 passed \| 14 skipped`** |

Local note: the worktree was installed with `--ignore-scripts` (the root `postinstall` kickstarts live launchd services), which leaves `better-sqlite3`'s native binding unbuilt and makes 214 unrelated test files fail at collection with `Could not locate the bindings file`. `pnpm rebuild better-sqlite3` fixes it; the full-suite result above is post-rebuild. CI does a normal install and is unaffected.

## Follow-ups (deliberately not in this PR)

1. **Split the two conflated "no terminal message" error classes.** `translate.ts:317-327` (true stream cutoff, 38 events) and `handle.ts:518` (cascade kill, 132 events / 104 cancelled) share the same wording and land in the same bucket, inflating cutoff counts ~4×. They want distinct failure codes.
2. **Record `streamIncompleteRetries` on the failure row.** The retry branch at `loop.ts:598-618` emits no telemetry, so a run that retried twice and died is indistinguishable from one that never retried — which is precisely why #678's effectiveness was unmeasurable.
3. **Reconsider #678's mitigation target.** `SUBAGENT_HANDOFF_CONTRACT` shrinks the child's *final reply*, but the upstream trigger documented in [anthropic-sdk-typescript#842](https://github.com/anthropics/anthropic-sdk-typescript/issues/842) is large **tool_use argument JSON** — a different payload. Post-#678 cutoff rate did not improve (2.94% → 4.12%).

### DIFF
--- a/src/agent/background-registry.test.ts
+++ b/src/agent/background-registry.test.ts
@@ -801,6 +801,42 @@ describe('BackgroundAgentRegistry', () => {
       expect(meta!.endedAt).toBeDefined();
       expect(meta!.endedAt).toBeGreaterThan(0);
     });
+
+    // PR#713 Wave-A A1(b): markTerminal persists `result.stopReason` into the
+    // closing meta record so the /bgsub:join disk-fallback path (bgsub.ts) can
+    // reconstruct the same partial-result labeling the in-memory replay
+    // applies, even after the job's in-memory entry is TTL-evicted.
+    it('terminal callback persists stopReason into meta.json when the result carries one', async () => {
+      const handle = createStubHandle('sub-disk-5');
+      const job = registry.register({ handle, prompt: 'capped job', model: 'sonnet' });
+
+      handle.__fireTerminal({
+        ...successResult('sub-disk-5', 'partial output'),
+        stopReason: 'tool_use_loop_capped',
+      });
+
+      await new Promise((r) => setTimeout(r, 200));
+
+      const { BgJobLogReader } = await import('./bg-job-log.js');
+      const meta = await BgJobLogReader.readMeta(job.jobId);
+      expect(meta).not.toBeNull();
+      expect(meta!.stopReason).toBe('tool_use_loop_capped');
+    });
+
+    it('terminal callback leaves meta.stopReason ABSENT when the result carries none', async () => {
+      const handle = createStubHandle('sub-disk-6');
+      const job = registry.register({ handle, prompt: 'clean job', model: 'sonnet' });
+
+      // successResult() never sets stopReason.
+      handle.__fireTerminal(successResult('sub-disk-6', 'clean output'));
+
+      await new Promise((r) => setTimeout(r, 200));
+
+      const { BgJobLogReader } = await import('./bg-job-log.js');
+      const meta = await BgJobLogReader.readMeta(job.jobId);
+      expect(meta).not.toBeNull();
+      expect(meta!.stopReason).toBeUndefined();
+    });
   });
 
   // -------------------------------------------------------------------------
@@ -1040,6 +1076,78 @@ describe('BackgroundAgentRegistry', () => {
       expect(failedCalls).toHaveLength(1);
       const entry = failedCalls[0]![0] as Record<string, unknown>;
       expect(entry['stop_reason']).toBe('stream_incomplete');
+    });
+
+    // -------------------------------------------------------------------
+    // A3: stop_reason is bounded before it rides into telemetry. On the
+    // OpenAI-compatible path `stopReason` is provider-controlled free text
+    // (translate.ts / responses-translate.ts), not a bounded enum, so — like
+    // the sibling free-form fields — it must be capped, not emitted uncapped.
+    // -------------------------------------------------------------------
+    it('completed: a >64-char stop_reason is truncated to 64 chars + ellipsis in the emitted row', () => {
+      const longReason = 'x'.repeat(100);
+      const handle = createStubHandle('rt-stopreason-trunc-1');
+      registry.register({ handle, prompt: 'p', model: 'sonnet' });
+      handle.__fireTerminal({
+        ...successResult('rt-stopreason-trunc-1', 'done'),
+        stopReason: longReason,
+      });
+
+      const completedCalls = appendRoutingDecision.mock.calls.filter(
+        (c) => (c[0] as Record<string, unknown>)?.['event'] === 'subagent.completed',
+      );
+      expect(completedCalls).toHaveLength(1);
+      const entry = completedCalls[0]![0] as Record<string, unknown>;
+      // truncate(s, 64): first 64 chars + a single '…' — never the raw
+      // 100-char string, and never silently dropped.
+      expect(entry['stop_reason']).toBe('x'.repeat(64) + '…');
+      expect((entry['stop_reason'] as string).length).toBe(65);
+    });
+
+    it('failure: a >64-char stop_reason is truncated to 64 chars + ellipsis in the emitted row', () => {
+      const longReason = 'y'.repeat(90);
+      const handle = createStubHandle('rt-stopreason-trunc-2');
+      registry.register({ handle, prompt: 'p', model: 'sonnet' });
+      handle.__fireTerminal({
+        ...failureResult('rt-stopreason-trunc-2', 'boom'),
+        stopReason: longReason,
+      });
+
+      const failedCalls = appendRoutingDecision.mock.calls.filter(
+        (c) => (c[0] as Record<string, unknown>)?.['event'] === 'subagent.failed',
+      );
+      expect(failedCalls).toHaveLength(1);
+      const entry = failedCalls[0]![0] as Record<string, unknown>;
+      expect(entry['stop_reason']).toBe('y'.repeat(64) + '…');
+    });
+
+    it('completed: a stop_reason at or under 64 chars is passed through unchanged (no ellipsis)', () => {
+      const shortReason = 'tool_use_loop_capped'; // well under 64 chars
+      const handle = createStubHandle('rt-stopreason-trunc-3');
+      registry.register({ handle, prompt: 'p', model: 'sonnet' });
+      handle.__fireTerminal({
+        ...successResult('rt-stopreason-trunc-3', 'done'),
+        stopReason: shortReason,
+      });
+
+      const completedCalls = appendRoutingDecision.mock.calls.filter(
+        (c) => (c[0] as Record<string, unknown>)?.['event'] === 'subagent.completed',
+      );
+      const entry = completedCalls[0]![0] as Record<string, unknown>;
+      expect(entry['stop_reason']).toBe(shortReason);
+    });
+
+    it('completed: an absent stop_reason stays ABSENT (never truncated to a defined value)', () => {
+      const handle = createStubHandle('rt-stopreason-trunc-4');
+      registry.register({ handle, prompt: 'p', model: 'sonnet' });
+      // successResult() never sets stopReason.
+      handle.__fireTerminal(successResult('rt-stopreason-trunc-4', 'done'));
+
+      const completedCalls = appendRoutingDecision.mock.calls.filter(
+        (c) => (c[0] as Record<string, unknown>)?.['event'] === 'subagent.completed',
+      );
+      const entry = completedCalls[0]![0] as Record<string, unknown>;
+      expect(entry['stop_reason']).toBeUndefined();
     });
   });
 });
--- a/src/agent/background-registry.test.ts
+++ b/src/agent/background-registry.test.ts
@@ -906,5 +906,140 @@ describe('BackgroundAgentRegistry', () => {
       const entry = completedCalls[0]![0] as Record<string, unknown>;
       expect(entry['parent_session_id']).toBeUndefined();
     });
+
+    // -----------------------------------------------------------------------
+    // stop_reason coverage (subagent partial-result observability).
+    //
+    // Telemetry proved `stop_reason` appeared on 0 of 14,152 `subagent.completed`
+    // rows — this predicate closes that blind spot for the background-registry
+    // emit path. `content_chars` must stay computed on the raw, un-annotated
+    // content regardless of stopReason (markTerminal never annotates; see the
+    // class-level note on markTerminal).
+    // -----------------------------------------------------------------------
+    it('completed: stop_reason is present and equals the result stopReason when set', () => {
+      const handle = createStubHandle('rt-stopreason-1');
+      registry.register({ handle, prompt: 'p', model: 'sonnet' });
+      handle.__fireTerminal({
+        ...successResult('rt-stopreason-1', 'partial output'),
+        stopReason: 'tool_use_loop_capped',
+      });
+
+      const completedCalls = appendRoutingDecision.mock.calls.filter(
+        (c) => (c[0] as Record<string, unknown>)?.['event'] === 'subagent.completed',
+      );
+      expect(completedCalls).toHaveLength(1);
+      const entry = completedCalls[0]![0] as Record<string, unknown>;
+      expect(entry['stop_reason']).toBe('tool_use_loop_capped');
+    });
+
+    it('completed: stop_reason is ABSENT (not null/empty) when the result carries none', () => {
+      const handle = createStubHandle('rt-stopreason-2');
+      registry.register({ handle, prompt: 'p', model: 'sonnet' });
+      // successResult() never sets stopReason.
+      handle.__fireTerminal(successResult('rt-stopreason-2', 'clean answer'));
+
+      const completedCalls = appendRoutingDecision.mock.calls.filter(
+        (c) => (c[0] as Record<string, unknown>)?.['event'] === 'subagent.completed',
+      );
+      expect(completedCalls).toHaveLength(1);
+      const entry = completedCalls[0]![0] as Record<string, unknown>;
+      // Key-absence in the PERSISTED row is asserted against the real
+      // buildRoutingDecisionRow in routing-telemetry.test.ts. Here
+      // appendRoutingDecision is mocked, so the pre-strip entry still carries
+      // the key with an undefined value — matching the parent_session_id
+      // precedent above ("not present or undefined is fine").
+      expect(entry['stop_reason']).toBeUndefined();
+    });
+
+    it('completed: content_chars measures raw content, unaffected by a partial stopReason', () => {
+      const handle = createStubHandle('rt-stopreason-3');
+      registry.register({ handle, prompt: 'p', model: 'sonnet' });
+      const rawContent = 'short partial body';
+      handle.__fireTerminal({
+        ...successResult('rt-stopreason-3', rawContent),
+        stopReason: 'stream_incomplete',
+      });
+
+      const completedCalls = appendRoutingDecision.mock.calls.filter(
+        (c) => (c[0] as Record<string, unknown>)?.['event'] === 'subagent.completed',
+      );
+      const entry = completedCalls[0]![0] as Record<string, unknown>;
+      // No `[⚠ PARTIAL RESULT…]` banner counted — markTerminal never annotates.
+      expect(entry['content_chars']).toBe(rawContent.length);
+      expect(entry['stop_reason']).toBe('stream_incomplete');
+    });
+
+    it('failure: stop_reason is present when the failed result carries one', () => {
+      const handle = createStubHandle('rt-stopreason-4');
+      registry.register({ handle, prompt: 'p', model: 'sonnet' });
+      handle.__fireTerminal({
+        ...failureResult('rt-stopreason-4', 'stream died'),
+        stopReason: 'stream_incomplete',
+      });
+
+      const failedCalls = appendRoutingDecision.mock.calls.filter(
+        (c) => (c[0] as Record<string, unknown>)?.['event'] === 'subagent.failed',
+      );
+      expect(failedCalls).toHaveLength(1);
+      const entry = failedCalls[0]![0] as Record<string, unknown>;
+      expect(entry['stop_reason']).toBe('stream_incomplete');
+    });
+
+    it('failure: stop_reason is absent when the failed result carries none', () => {
+      const handle = createStubHandle('rt-stopreason-5');
+      registry.register({ handle, prompt: 'p', model: 'sonnet' });
+      handle.__fireTerminal(failureResult('rt-stopreason-5', 'plain failure'));
+
+      const failedCalls = appendRoutingDecision.mock.calls.filter(
+        (c) => (c[0] as Record<string, unknown>)?.['event'] === 'subagent.failed',
+      );
+      expect(failedCalls).toHaveLength(1);
+      const entry = failedCalls[0]![0] as Record<string, unknown>;
+      // Key-absence in the PERSISTED row is asserted against the real
+      // buildRoutingDecisionRow in routing-telemetry.test.ts. Here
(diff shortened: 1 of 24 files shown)


griffinwork40:
## Review — PR #713

**Target**: `feat(telemetry): record stop_reason on subagent terminal rows and flag partial results` · `afk/subagent-partial-observability` → `main`
**Reviewed ref**: `06d1ee8` · 9 files · +240/−4 (244 lines) → regime **light**
**Stated intent**: PR #713 title + body — spec-compliance **assessed**

---

### HIGH

**`high · high · spec-compliance+correctness · src/cli/slash/commands/bgsub.ts:241-263 · ref:06d1ee8 · citation-type:file-state`**
finding: The `/bgsub:join` banner fix covers only the in-memory branch — the disk-log fallback branch, taken for any join more than 5 minutes after a job settles, replays through `formatDiskEvent` (which never annotates) and prints bare `Status: completed` for a capped or stream-cut partial, so intent claim 2 ("It was the only delivery consumer lacking one") is unmet for the dominant branch of the very command it names.
evidence:
```
const diskMeta = await BgJobLogReader.readMeta(id);
    …  const text = formatDiskEvent(event);
ctx.out.line(`  Status: ${diskMeta.status}` + …
```
`const TERMINAL_EVICT_TTL_MS = 5 * 60 * 1000;` (background-registry.ts:106) → `setTimeout(() => this.jobs.delete(jobId), …)` (:672). `git grep -n stopReason 06d1ee8 -- src/agent/bg-job-log.ts` → **0 matches**; `BgJobMeta` (bg-job-log.ts:37-49) = jobId, subagentId, label, promptHash, model, startedAt, endedAt?, status, parentSessionId?, schemaVersion — the banner is not derivable on that path even in principle.
suggestion: Add `stopReason?: string` to `BgJobMeta`, write it in `markTerminal`'s meta close, and annotate in the disk branch — or narrow the PR body's completeness claim to the in-memory path and file the rest.
**Not a regression** — the disk path was equally unmarked before this diff.

### MEDIUM

**`medium · high · spec-compliance · src/agent/providers/anthropic-direct/types.ts:104-117 · ref:06d1ee8 · citation-type:file-state`**
finding: `incomplete`/`incompleteReason` are write-only at this ref — zero production readers — while the field's own docblock sells the capability as letting "trace readers, hooks, calling code" branch on the fact.
evidence: `git grep -n "incompleteReason\|incompleteToolResultFields" 06d1ee8 -- 'src/**'` → declaration (types.ts:117), helper (result.ts:98-104), and the two write sites (fork-dispatch.ts:360, foreground-promotion.ts:288). Zero reads. The two `.incomplete` reads in the tree are unrelated types: `receipt.ts:288` (session-seal payload) and `responses-translate.ts:149` (SSE event). Separately, the trace-reader audience named in the docblock already has the same fact: `SubagentSucceededPayload.stopReason` (trace/types.ts:243-249, docblock: "so a trace reader can distinguish a clean completion from a capped/truncated partial").
suggestion: Land one consumer in this PR, or defer the two fields until a consumer needs them.

**`medium · high · correctness · src/agent/tools/subagent/foreground-promotion.ts:332-335 · ref:06d1ee8 · citation-type:file-state`**
finding: The failure-path `ToolResult` omits `incompleteToolResultFields` while `result.stopReason` is in scope, so the pure zero-output `stream_incomplete` case — which resolves `status: 'failed'` by design and routes exactly here — never receives the flag documented to cover "stream was cut off mid-flight".
evidence: `toolResult = { content: JSON.stringify(payload), isError: true, };` — no spread. Same omission at the fork-dispatch failure literals.
suggestion: Spread `...incompleteToolResultFields(result.stopReason)` into the failure literal too.

**`medium · medium · observability · src/agent/subagent/result.ts:57 · ref:06d1ee8 · citation-type:file-state`**
finding: `isIncompleteStopReason` matches only the two local sentinels, so an OpenAI-compatible `finish_reason: 'length'` — a genuine mid-answer truncation — now lands in `stop_reason` telemetry but triggers neither banner nor flag, leaving the PR's three-way separability holding on the Anthropic path only. Pre-existing gap, newly visible.
suggestion: Extend the predicate to provider truncation reasons, or document the Anthropic-only scope.

### LOW

**`low · high · test-coverage · src/agent/subagent/result.ts:98-105 · ref:06d1ee8 · citation-type:file-state`**
finding: Deleting `incompleteToolResultFields`, either spread, or the new `bgsub.ts:133` call turns nothing red — two of three shipped features have no assertions.
evidence: `rg -n 'incompleteToolResultFields|incompleteReason' src -g '*.test.ts'` → 0 hits. `bgsub.test.ts` exists (20 cases) with 0 `stopReason`/`PARTIAL` matches. The `stop_reason` telemetry work **is** covered (9 new cases) — but only the background-registry emits; deleting foreground-promotion.ts:264/:309 also stays green.

**`low · medium · security · src/agent/tools/subagent/foreground-promotion.ts:264,309 · ref:06d1ee8 · citation-type:file-state`**
finding: `stop_reason` is written uncapped to the append-only JSONL while the sibling free-form field is capped, and on the OpenAI-compatible path `stopReason` is provider-controlled free text.
evidence: `stop_reason: result.stopReason,` vs `error_message: truncate(errorMessage),` (:304); `truncate(s, max = 240)` (failure-payload.ts:29). Producers: `translate.ts:182,195`, `responses-translate.ts:151`. **No user/model content flows here** — provenance is provider sentinels only, so this is robustness, not exposure.

### Clean / refuted (verified, not assumed)

- **api-compat — no issues found**: `tools/types.ts:11` re-exports `ToolResult` from the provider declaration (same interface); both new members optional; `exactOptionalPropertyTypes` is off, so `incomplete?: true` widens cleanly into `incomplete?: boolean`.
- **No double-banner** (the highest-risk lead, independently re-derived twice): the only write is `job.result = result` (raw) at background-registry.ts:580; the notifier and `bgsub.ts` each annotate a *local* string from raw content. Non-idempotency is unreachable.
- **Emit-site coverage COMPLETE**: the two un-updated emits (foreground-promotion.ts:345 catch, subagent-executor.ts:640 fork-throw) genuinely have no `SubagentResult` in scope.
- **No telemetry schema drift**: sole reader `insights/aggregators/routing.ts` uses bare `JSON.parse` with named-key reads — no zod, no closed key set.
- **`dag-subagent` correctly excluded** — returns a bare string, not a `ToolResult`.

### Dropped in verification

- `trace/tool-call-trace.ts:79-93`, cited by Wave 1.5 as the boundary that drops the field, **does not exist at `06d1ee8`** → fabricated citation, finding re-grounded on my own grep. The zero-consumer *fact* stands; the drop *mechanism* is `unverified — no such file at the reviewed ref`.

### What was not checked

Wave 1 and Wave 1.5 both read at `06d1ee8`; the main session independently re-derived the sole `high` finding, the zero-consumer claim, and api-compat via `git show 06d1ee8:<path>` and `git grep -n … 06d1ee8` — this discharged `/shadow-verify`'s function on primary evidence rather than via a third subagent. Wave 2 synthesis was done inline (single-source findings, so no cross-agent conflict was possible). No tests, lint, or build were run — the PR body's own claims (`pnpm lint` clean, 701 files / 13124 passing) are **unverified here**. Also not checked: whether `SubagentSucceededPayload.stopReason` is actually populated at runtime, the Telegram delivery surface, plugin-authored tool handlers consuming `ToolResult`, and git history for whether the notifier ever mutated `job.result`.

---

## DO NOT MERGE

Nothing in the diff is broken and nothing regresses — it is additive, correct where it lands, and strictly improves observability. The block is narrow: the PR body asserts the `/bgsub:join` marker gap is closed, but the branch an operator actually reaches (>5 min after settle) still renders a capped partial as clean `completed`, and one of the three shipped features has neither a consumer nor a test. Merging as-written reproduces the exact failure this PR was authored to end — *"looked successful for weeks because no discriminating signal was ever recorded."* Minimum unblock: carry `stopReason` into `BgJobMeta` and annotate the disk branch (or narrow the claim in the body), then add the two missing assertions.

**Done**
- Reviewed PR #713 read-only across all six dimensions at ref `06d1ee8`: Wave 1 (light regime, 1 agent) → Wave 1.5 (independent citation + absence verifier) → inline synthesis, with the sole `high` finding re-derived firsthand in the main session.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/50517fea-194c-4f7f-95ed-6d2e65b1eeea/pr-713.diff`; every finding carries a `file:line` verified at `06d1ee8` via `git show`/`git grep`; one fabricated citation (`trace/tool-call-trace.ts:79-93`) dropped for non-existence at the ref.
- World unchanged: no files written, no commits, no pushes, no PR comments or edits — worktree still clean at `06d1ee8`. The report above lives in this transcript only; it was not posted to the PR (out of scope for `/review`).
- Deferred: the PR's own `pnpm lint` / full-suite claims were not re-run; runtime population of `SubagentSucceededPayload.stopReason` and the Telegram delivery surface were not inspected.

---

griffinwork40:
# Code Review — PR #713

**`feat(telemetry): record stop_reason on subagent terminal rows and flag partial results`**
`main` ← `afk/subagent-partial-observability` · reviewed ref **`355555c`** · 24 files · +785/−11 · regime **full** · OPEN, MERGEABLE

Waves: 2 parallel dimension agents → 1 independent verification agent (citations + absence claims + one negative claim). Synthesis inline (7 findings; dispatching a synthesis agent would have added latency without changing the ordering).

---

## Findings

### medium · confidence high · perf-observability / hygiene
`src/agent/background-registry.ts:685` · ref:`355555c` · citation-type: file-state

**finding:** `markTerminal` bounds `stopReason` to 64 chars at all three telemetry emits but writes it **unbounded** into the durable `meta.json` 20 lines later, so the bounding invariant the PR deliberately introduced is applied at 5 of 6 write sites.

```ts
...(result.stopReason !== undefined ? { stopReason: result.stopReason } : {}),
```

Reachability re-derived independently: `openai-compatible/translate.ts:136` (`if (choice.finish_reason) state.finishReason = choice.finish_reason` — raw wire string, typed `string | null`) → `translate.ts:195` → `subagent/handle.ts:448-449` (`typeof`-only check, no length or enum narrowing) → `handle.ts:230` → this write. `readMeta` (`bg-job-log.ts:261`) is a bare `as BgJobMeta` cast, so nothing validates on read either.

**suggestion:** Apply the existing module-local `boundedStopReason` (`background-registry.ts:133-135`) at the `writeMeta` spread, matching `:626`/`:646`/`:665`.

> **CONFLICT resolved.** The security agent filed this as `medium · security` ("provider-controlled free text lands on disk verbatim"); the verifier **refuted the security framing**: the sole disk consumer (`bgsub.ts:253`) gates on `isIncompleteStopReason` — an exact `===` against two constants — before rendering, so an oversized value is written but never displayed. No injection, no exposure, no unauthorized access. Reclassified as a bound-parity/disk-hygiene defect. Kept at `medium` under the rubric's "missing edge case."

### low · confidence high · test-coverage
`src/agent/tools/subagent/foreground-promotion.ts:293,348` and `src/agent/tools/skill-executor/fork-dispatch.ts:360,377,390` · ref:`355555c` · file-state

**finding:** All five `incompleteToolResultFields` spreads at the two `ToolResult`-bearing delivery sites are uncovered — the only test file for foreground-promotion asserts the *prose banner* (`foreground-promotion.test.ts:27`, inside `describe('withProvenanceHeader')`), never the structured field, and no `fork-dispatch` test file exists.

**absence verified:** `rg -n "incompleteToolResultFields|incompleteReason" src` → production hits only; `rg -n "incomplete" -g "*.test.ts"` → zero assertions on the field at either site; no root `tests/`/`e2e/`/`__tests__/` dirs exist (tests are co-located only).

**suggestion:** Assert `toolResult.incomplete === true` for a `stream_incomplete` result and key-absence for a clean one, at one site per file.

### low · confidence high · test-coverage
`src/agent/providers/anthropic-direct/loop.ts:1018`, `src/agent/providers/openai-compatible/query/dispatch-append.ts:213` · ref:`355555c` · diff-context

**finding:** Neither provider has a test asserting `incomplete` reaches the yielded `tool.output` ProviderEvent, so the cross-provider symmetry commit `355555c` explicitly set out to establish would break silently on either side.

**absence verified:** nearest coverage is `providers/shared/tool-call-trace.test.ts:128-149` (the shared trace builder, not the yielded event) and `dispatcher.test.ts:439-485` (PostToolUse context, not the event). Zero hits in either provider's test file.

**suggestion:** One test per provider asserting `incomplete: true` on the yielded `tool.output`.

### low · confidence high · test-coverage
`src/agent/tools/subagent/foreground-promotion.ts:269,316` · ref:`355555c` · diff-context

**finding:** The `truncate(result.stopReason, 64)` bound is untested at the two foreground emit sites — the path the PR body itself calls "the large majority of dispatches" — while all four truncation assertions live in `background-registry.test.ts:996-1078`, where the fixture value (`stream_incomplete`, 18 chars) never exceeds the bound.

**absence verified:** `rg -n "truncate\(result\.stopReason|stop_reason" -g "*.test.ts"` → hits only in `routing-telemetry.test.ts:88-90` and `background-registry.test.ts`. This compounds with the `medium` above: the bound is both incompletely applied and unpinned by any over-64-char case.

**suggestion:** One assertion with a >64-char `stopReason` on the foreground completed emit.

### low · confidence high · spec-compliance
PR body vs. branch · ref:`355555c` · file-state

**finding:** The body enumerates 4 changes and closes "All changes are additive and backward-compatible; no existing consumer behavior changes," but commits `b1ecb8d` and `355555c` add an observability surface the body never names — `BgJobMeta.stopReason` persistence (`bg-job-log.ts:56`), trace payload + zod fields (`trace/events.ts:49,52`), both provider `tool.output` events, `PostToolUseContext` (`hooks.ts:260,266`), and the 64-char bound at all five telemetry sites.

**Judgment: in-scope reviewer-feedback refinement, not creep.** Each later change makes true the body's own line-32 promise that "non-model consumers … can branch without substring-matching banner text" — which was write-only before those commits. **Unmet intent: none** — all 4 numbered changes have implementing code.

**suggestion:** Add a "5. Reviewer-feedback follow-ups" section naming the BgJobMeta, trace/zod, provider-event, PostToolUse, and bounding changes.

### low · confidence high · correctness
`src/cli/slash/commands/bgsub.ts:253-256` · ref:`355555c` · file-state

**finding:** The disk-replay branch renders the banner as a detached marker via `annotateIfIncomplete('', …).trimEnd()` followed by a blank line, with the replayed body printed later by `formatDiskEvent`, whereas the in-memory path at `:133` prepends the banner directly onto the body — so the two paths commit `b1ecb8d` claims "label it identically" differ in output shape.

```ts
if (isIncompleteStopReason(diskMeta.stopReason)) {
  ctx.out.line(`  ${annotateIfIncomplete('', diskMeta.stopReason).trimEnd()}`);
```

**suggestion:** Emit the marker immediately before the first event line so marker/body adjacency matches `printJobDetail`, or note the deliberate divergence in the comment.

### nit · confidence high · api-compat
`src/agent/tools/subagent/foreground-promotion.ts:269,316` · ref:`355555c` · diff-context

**finding:** Both foreground sites inline `truncate(result.stopReason, 64)` rather than reusing `boundedStopReason`, leaving the 64-char bound duplicated across three locations with no shared constant — which is how site six (`:685`) drifted.

**suggestion:** Export `boundedStopReason` and call it at both foreground sites plus the `writeMeta` spread.

---

## Verified clean (negatives independently re-derived, not assumed)

- **No banner double-prepend or loss.** All six production `annotateIfIncomplete` call sites write to locals only; `background-registry.ts:593` stores the **un**annotated result, and `bg-result-notifier.ts` never reassigns `.content` (`rg "\.content =" → 0 hits`). Auto-deliver-then-rejoin, in-memory rejoin, and post-TTL disk rejoin each render exactly one banner.
- **`incompleteReason` is structurally closed.** `result.ts:57-59` is `===` against two module constants (no regex/prefix/`includes`), and `:101-104` returns only past that gate — arbitrary provider text provably cannot reach the trace payload, provider events, or `PostToolUseContext`.
- **Backward-compat claim holds.** All new fields optional; zod objects non-strict (`rg 'strict\(\)' trace/events.ts` → 0) so legacy rows parse with no injected default; `schemaVersion: 1` unchanged and pre-change meta files still load; `firePostToolUse` is `private` with all four call sites (`dispatcher.ts:1147,1173,1181,1209`) fed and omission producing a byte-identical context; `buildRoutingDecisionRow` drops undefined so absent stays absent, never `null`/`''`.
- **The new signal is actually queryable.** `stop_reason` survives the real `buildRoutingDecisionRow` (pinned at `routing-telemetry.test.ts:86-99`), and trace fields survive because `writer.ts:368` writes the original payload rather than a parse result — closing the 0-of-14,152 blind spot the PR documents.
- **Hot-path cost negligible.** Two conditional spreads on already-constructed literals, matching the adjacent `circuitBreaker`/`failureClass` idiom; no allocation when absent.

## What was not checked

- **Ref:** both waves read against branch HEAD **`355555c`** (worktree clean, `git diff HEAD` = 0 lines); all citations verified at that ref. No `diff-only` or `fabricated` citations. All four absence claims classified `confirmed-absent` by an independent grep.
- **Stated intent:** PR #713 title + body — **spec-compliance assessed**.
- **No test execution** (skill is read-only): the PR's own gate table (`701 files / 13124 passed`) and the "verified RED-on-delete" claims in `355555c` are unverified here; "deleting the spread leaves the suite green" is inferred from zero assertions, not observed.
- Not checked: real-world `finish_reason` max length from any live OpenAI-compatible/MLX shim; `truncate` behavior at multi-byte/grapheme boundaries; whether any downstream `routing-decisions.jsonl` consumer (`afk improve scan`, ceiling-ledger) reads `stop_reason` yet; non-`src/` surfaces documenting the `[⚠ PARTIAL RESULT…]` contract; git history/blame.
- `/shadow-verify` not invoked — zero critical/high findings. Wave 1.5 served that role on the load-bearing claims and changed one verdict.

---

# ✅ MERGE

Zero blocking, zero critical, zero high. The PR's central claim — additive, backward-compatible, no consumer behavior change — **holds under independent re-derivation** on every surface checked. The one `medium` is a one-line bound-parity gap at `background-registry.ts:685` that is written-but-never-rendered; the `low`s are coverage gaps on new spreads plus a stale PR body. Landing as-is is safe; `boundedStopReason` at `:685` is the single change worth making first.

## Done
- Reviewed PR #713 at ref `355555c` across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability; 7 findings (1 medium, 5 low, 1 nit), 5 negative claims independently re-derived, merge decision **MERGE**.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/fe975f98-769b-4628-bb28-31ae8c694fcd/pr-713.diff`; intent captured at `/tmp/pr713-body.md` + `/tmp/pr713-commits.txt`; every blocking-tier citation and all four absence claims verified against `355555c` by an independent Wave 1.5 agent.
- World unchanged: read-only review — no files edited, nothing committed, no PR comment or review posted, working tree still clean at `355555c`.
- Deferred: findings are reported, not fixed (skill is read-only); the Wave 1 `medium · security` framing was downgraded to hygiene after the verifier refuted the display path, and that reclassification is recorded inline as a CONFLICT rather than silently dropped.

---

Issue #612, opened 2026-07-14
Observability: tag child tool_calls with subagentId so foreground subagent timeouts are diagnosable


griffinwork40:
## Context

Reconstructing the PR #607 fix-subagent timeout, the child's **actual tool commands were unrecoverable from disk**: the interactive witness trace didn't cover the fix window (20:41–21:01), and child `tool_call` events carry **no `subagentId` tag**, so "ran the full suite repeatedly" was operator inference, not a captured log.

PR #610 added `promptHead` + `agentType` on `subagent_lifecycle.started` and wired the daemon trace writer into fork executors, but **deferred** per-tool-call `subagentId` tagging (the fields already exist on `ToolCallStartedPayload`/`ToolCallCompletedPayload` in `src/agent/trace/types.ts`; the reader in `trace.ts` already renders `[subagentId]`) — emission is the gap.

## Ask

1. **Tag foreground child `tool_call` started/completed events with `subagentId`** so `afk trace show` can attribute which child did what.
2. **Investigate the observed parent-trace seal before the fix window** in session `1e4fae55` — determine whether foreground subagents dispatched in a later REPL turn are covered by the parent session trace, and keep it open across child lifecycles if not.

Outcome: a timed-out child's actual work (e.g. "ran `pnpm test` 30× in 20 min") is answerable from the trace, not inferred.

## Refs
- #610 (observability follow-ups)
- Session `1e4fae55`

PR #714, opened 2026-07-26
fix(dispatcher): emit hook_decision for silently-denied tool calls, attribute blocks to forks

griffinwork40:
## Problem

Three dispatcher paths returned an `isError` `ToolResult` to the model for a **policy** reason without emitting any `hook_decision` trace event:

- `checkReadOnlyBash` — the read-only-skill bash gate
- `checkToolPermission` on the single-dispatch path
- `checkToolPermission` on the batch path

A denial the model could see was invisible in the trace.

Separately, `hook_decision` carried no `subagentId`, so a block could be **counted** but never **attributed** to the child that provoked it — even though `tool_call` has carried `subagentId` since #612.

## Why it matters

Together these made sub-agent failures effectively unattributable after the fact. Child tool arguments and error text are not persisted anywhere, so the trace event is the only durable record that a denial happened.

Concretely: an audit of 12,108 witness traces attributed 268 child `bash permission-denied` events to path containment. Re-attributing them by child `agentType` showed **255 of 268 (95%) came from the read-only bash gate firing on `git-investigator`** — a deliberately read-only agent — at 5.7% of its 4,459 bash calls. That is the gate working as designed, and the agent then complying with its remediation text. The original reading (children evading a path-containment guard) was wrong, and it was wrong *because* the dominant denial source emitted no event and nothing tied blocks to a child.

With these two changes the same question is a single query.

## Changes

- Add `HookDecisionPayload.subagentId` (+ matching zod schema field) and thread it from both providers (`anthropic-direct`, `openai-compatible`) through `SessionToolDispatcher`, mirroring the existing fork-scoped-signal pattern.
- Route **all five** PreToolUse block emissions through one new `emitPreToolUseBlock` helper, so no call site can forget the stamp. The two pre-existing emit sites (canUseTool threw / canUseTool deny) gain `subagentId` as a side effect.
- Share a single `reason` string between the model-visible result and the trace event so the two cannot drift.
- Pin the ADR-0001 fork permission-inheritance premise with tests.

### Privacy

Raw command strings stay out of telemetry, per ADR 0001 section G.4. The emitted `reason` is the same policy explanation already returned to the model; the `bash` command itself is never recorded.

### Behavior

**No behavior change for the model.** Every denial's `content` string is byte-identical to before — the refactor only hoists it into a `reason` const shared with the new trace event.

## Tests

Adds a `permission-mode inheritance (ADR-0001 premise)` block to `child-config.test.ts`. This premise — a forked child never inherits a bypassed parent's `permissionMode` — is stated in `docs/decisions/0001-bash-tool-path-containment.md` and relied on both by that ADR's warn-only reasoning and by the path-approval hook's fork auto-deny, but was asserted by **zero** tests. An accidental flip in either direction (a child silently gaining bypass, or a top-level surface silently losing it) would previously have shipped green.

Three assertions:
1. `buildChildConfig` never sets `permissionMode` on the child config
2. an unset child `permissionMode` resolves to `'default'` (containment ON)
3. `pathContainmentBypassed(DEFAULT_CLI_PERMISSION_MODE)` is `true` while the child's resolved mode is `false`

## Verification

| gate | result |
|---|---|
| full suite | 701 files, 13,118 passed, 14 skipped, 0 failed |
| coverage | 87.09 stmts / 85.15 branch / 91.25 funcs / 87.09 lines (floors 74 / 79 / 82 / 74) |

- Persisting child tool args / error text is the remaining half of the observability gap, but it carries a telemetry-privacy decision (ADR 0001 section G.4 deliberately excludes raw commands). Suggested shape: error text plus a credential-scrubbed argument head, not raw args.
- Deliberately **not** widening the denial circuit breaker to `bash`. It currently cannot fire on the dominant denial family, but that family is `git-investigator` correctly probing an intentional read-only boundary — a 5-strike breaker would abort legitimate read-only investigations mid-flight.
- `git-investigator` spends ~5.7% of its bash budget rediscovering the read-only boundary; tightening its prompt or pre-filtering its command set would recover that. Agent tuning, not containment.

### DIFF
--- a/src/agent/providers/anthropic-direct/index.ts
+++ b/src/agent/providers/anthropic-direct/index.ts
@@ -344,6 +344,13 @@ export class AnthropicDirectProvider implements ModelProvider {
       sessionId?: string;
       parentSessionId?: string;
       /**
+       * This fork's own subagent id. Stamped onto every `hook_decision` the
+       * dispatcher emits so a policy block is attributable to the child that
+       * provoked it (parity with what `tool_call` already records). Undefined on
+       * a top-level session.
+       */
+      subagentId?: string;
+      /**
        * Explicit "this session is a forked subagent" signal carrying the
        * per-result output-cap budget (#661). Set to MODEL_CAP_BYTES by
        * `SubagentManager.forkSubagent` for EVERY fork; undefined on a top-level
@@ -492,6 +499,7 @@ export class AnthropicDirectProvider implements ModelProvider {
       ...(opts?.env !== undefined ? { env: opts.env } : {}),
       sessionId: opts?.sessionId,
       parentSessionId: opts?.parentSessionId,
+      ...(opts?.subagentId !== undefined ? { subagentId: opts.subagentId } : {}),
       // Central output-cap backstop (#661), FORK-SCOPED. Armed from the
       // explicit `subagentToolOutputCapBytes` signal that
       // `SubagentManager.forkSubagent` stamps (as MODEL_CAP_BYTES = 100KB) on
@@ -748,6 +756,7 @@ export class AnthropicDirectProvider implements ModelProvider {
           ...(config.env !== undefined ? { env: config.env } : {}),
           sessionId: config.sessionId,
           parentSessionId: config.parentSessionId,
+          ...(config.subagentId !== undefined ? { subagentId: config.subagentId } : {}),
           // Fork-scoped central output cap (#661): forwarded from the child
           // config that forkSubagent stamped, arming maxOutputBytes for forks
           // only (top-level leaves it unset).
--- a/src/agent/providers/openai-compatible/index.ts
+++ b/src/agent/providers/openai-compatible/index.ts
@@ -240,6 +240,7 @@ export class OpenAICompatibleProvider implements ModelProvider {
           ...(this._sharedWriteRoots !== undefined ? { writeRoots: this._sharedWriteRoots } : {}),
           ...(config.sessionId !== undefined ? { sessionId: config.sessionId } : {}),
           ...(config.parentSessionId !== undefined ? { parentSessionId: config.parentSessionId } : {}),
+          ...(config.subagentId !== undefined ? { subagentId: config.subagentId } : {}),
           // Fork-scoped central output cap (#661): forwarded from the child
           // config that forkSubagent stamped, arming maxOutputBytes for forks
           // only (top-level leaves it unset). Parity with anthropic-direct.
@@ -359,6 +360,13 @@ export class OpenAICompatibleProvider implements ModelProvider {
       writeRoots?: string[];
       sessionId?: string;
       parentSessionId?: string;
+      /**
+       * This fork's own subagent id — parity with
+       * `anthropic-direct/index.ts:buildDispatcher`. Stamped onto every
+       * `hook_decision` the dispatcher emits so a policy block is attributable
+       * to the child that provoked it. Undefined on a top-level session.
+       */
+      subagentId?: string;
       /**
        * Explicit "this session is a forked subagent" signal carrying the
        * per-result output-cap budget (#661) — parity with
@@ -507,6 +515,7 @@ export class OpenAICompatibleProvider implements ModelProvider {
     if (opts.writeRoots !== undefined) dispatcherOpts.writeRoots = opts.writeRoots;
     if (opts.sessionId !== undefined) dispatcherOpts.sessionId = opts.sessionId;
     if (opts.parentSessionId !== undefined) dispatcherOpts.parentSessionId = opts.parentSessionId;
+    if (opts.subagentId !== undefined) dispatcherOpts.subagentId = opts.subagentId;
     // Central output-cap backstop (#661), FORK-SCOPED — parity with
     // AnthropicDirectProvider.buildDispatcher. Armed from the explicit
     // `subagentToolOutputCapBytes` signal that `SubagentManager.forkSubagent`
--- a/src/agent/tools/dispatcher.ts
+++ b/src/agent/tools/dispatcher.ts
@@ -162,6 +162,13 @@ export interface SessionToolDispatcherOptions {
    */
   parentSessionId?: string;
   /**
+   * This fork's own subagent id, when the dispatcher belongs to a forked child.
+   * Stamped onto every `hook_decision` this dispatcher emits so a block can be
+   * ATTRIBUTED to the child that provoked it, mirroring what `tool_call`
+   * already records. Undefined for top-level sessions.
+   */
+  subagentId?: string;
+  /**
    * The PROVIDER that owns this dispatcher (it implements {@link GrantManager}).
    * The provider's `buildDispatcher` passes `this`; the dispatcher injects it
    * onto every PreToolUse/PostToolUse context as `context.grantManager` so
@@ -244,6 +251,7 @@ export class SessionToolDispatcher implements ToolDispatcher {
   private readonly _env: Record<string, string> | undefined;
   private readonly sessionId: string | undefined;
   private readonly parentSessionId: string | undefined;
+  private readonly subagentId: string | undefined;
   /**
    * Provider that owns this dispatcher (implements GrantManager). Injected onto
    * PreToolUse/PostToolUse contexts so path-scoped hooks read THIS session's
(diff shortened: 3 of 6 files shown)


griffinwork40:
# Code Review — PR #714

**Target:** `fix(dispatcher): emit hook_decision for silently-denied tool calls, attribute blocks to forks`
**Reviewed ref:** `68246fc69f76fce1439e373d1b91dbb7bb079ca3` · `afk/subagent-permission-scope` → `main` · 6 files, +125/−47
**Stated intent:** PR #714 title + body — **spec-compliance assessed**

## Delegation integrity (disclosed, not hidden)

The planned agent wave largely failed and I did **not** silently paper over it:

| dispatch | outcome |
|---|---|
| Wave 1 · security·api-compat agent | returned, but flagged **PARTIAL** (hit tool-use iteration cap) — did cover all 5 assigned questions |
| Wave 1 · correctness·spec·tests·perf agent | **FAILED — request timed out** |
| Re-dispatch · 3 narrow compose nodes | **all 3 FAILED** — 2 exceeded *my* 28-tool-call budget, 1 exceeded *my* 300s timeout (my sizing error, not the agents') |

**Fallback taken:** I re-derived every load-bearing claim myself with mechanical `git show` / `git grep` / targeted reads at the reviewed ref. This also discharges the Wave 1.5 citation-and-absence gate and eval-0001's rule — the PARTIAL agent's confident claims were treated as a trigger to verify, and all four were independently confirmed on a second path before informing the merge decision.

## The PR's own four claims — independently re-derived

| claim | verdict | my evidence at `68246fc` |
|---|---|---|
| "all five" emit sites routed through one helper | **CONFIRMED** | `emitPreToolUseBlock` called at `dispatcher.ts:566, 753, 758, 811, 915`; helper declared `:584` |
| model-visible `content` is byte-identical | **CONFIRMED** | diffed all 5 denial strings `main` vs `68246fc` — templates, spacing, and the `??` / `||` operators all preserved (an operator swap would have been a real behavior change; it did not happen). Only trailing `,`→`;` changed, which is statement syntax, not string content |
| `subagentId` actually populated for forks (not dead code) | **CONFIRMED** | producer `src/agent/subagent.ts:810` (`subagentId: id`, also `:652`, `:1063`) → both providers → `dispatcher.ts:334` → read at `:590` |
| raw bash commands stay out of telemetry | **CONFIRMED** | `classifyBashCommand`'s `verdict.reason` is a static per-rule taxonomy label; it never interpolates the command. The `canUseTool (threw)` emission that interpolates `err.message` is **pre-existing**, not a regression from this PR |

Additional clean reads: **no double-emission** — batch phase-1 does `results[i]=…; blocked.add(i); continue;` per branch, so a call denied by the allowlist never reaches the bash gate; single path `return`s immediately; the `2 / 2a / 2b / 2c` comment order matches actual code order. **The new awaited I/O cannot break a denial** — `emitHookDecision` (`src/agent/trace/emit.ts:48-58`) try/catches `writer.write` and never rethrows, so a sealed/failing writer degrades to a `debugLog`. Schema/type parity holds and `HookDecisionPayloadSchema` is plain `z.object` (strips unknown keys), so old readers parsing new traces and new readers parsing old traces both pass.

## Findings

```
low · high · test-coverage · src/agent/tools/dispatcher.test.ts:1610-1662 · ref:68246fc · citation-type:file-state
finding: The PR's headline feature — subagentId on hook_decision — is asserted by no test anywhere, and the three newly-converted emit sites (read-only bash gate, single-path allowlist, batch-path allowlist) have no emission test either; the only hook_decision assertions in dispatcher.test.ts cover the two PRE-existing canUseTool sites.
evidence: it('deny emits a hook_decision block')   // :1610
          it('throw (fail-closed) emits a hook_decision block')  // :1632
          it('allow emits no hook_decision')       // :1654
          # git grep -n "hook_decision" 68246fc -- 'src/**/*.test.ts' | grep -i subagent  -> no hook_decision+subagentId assertion
suggestion: Add one dispatcher test asserting a permission-denied call emits hook_decision, and one forked-dispatcher test asserting that event carries subagentId.
```

```
low · high · spec-compliance · src/agent/tools/subagent/child-config.test.ts:588-592 · ref:68246fc · citation-type:file-state
finding: The third new test's parent-side assertion `expect(DEFAULT_CLI_PERMISSION_MODE).toBe('bypassPermissions')` restates the constant it imports, so it pins the constant's literal value rather than the top-level surfaces' actual resolved mode — the "a top-level surface silently losing it" half of the block's stated goal is not covered by that assertion.
evidence: expect(DEFAULT_CLI_PERMISSION_MODE).toBe('bypassPermissions');
          expect(pathContainmentBypassed(DEFAULT_CLI_PERMISSION_MODE)).toBe(true);
suggestion: Assert the resolved mode from a real top-level surface's built config (the same buildInitialState path the child side uses) instead of the imported constant.
```

```
nit · high · correctness · src/agent/tools/dispatcher.ts:576-583 · ref:68246fc · citation-type:file-state
finding: The helper's JSDoc asserts "EVERY path that returns an isError ToolResult to the model for a POLICY reason must call this", but two dispatcher-internal block paths in the same file return isError policy denials without emitting — the repeat-loop breaker and the denial-breaker trip at :713-717 — so the word EVERY overstates what the code guarantees to a future maintainer.
evidence: * Invariant: EVERY path that returns an isError {@link ToolResult} to the
          * model for a POLICY reason must call this.
          return { content: buildDenialBreakerMessage(...), isError: true, failureClass: DENIAL_BREAKER_FAILURE_CLASS };
suggestion: Narrow the sentence to "every PreToolUse permission-denied path", which is exactly what the five converted sites cover.
```

```
nit · high · spec-compliance · src/agent/tools/subagent/child-config.test.ts:579-581 · ref:68246fc · citation-type:file-state
finding: The test comment presents a quoted ADR sentence that is a paraphrase rather than the ADR's verbatim text, so a future reader grepping the ADR for the quoted string finds nothing.
evidence: test:  ("Forked sub-agents run permissionMode = 'default' — they do not inherit bypass")
          ADR docs/decisions/0001-bash-tool-path-containment.md:59:  `permissionMode = 'default'` (they do **not** inherit bypass), so they would
suggestion: Quote line 59 verbatim or drop the quotation marks and cite the line number.
```

**Clean dimensions:** security — `no issues found — read dispatcher.ts, readonly-bash.ts, trace/emit.ts at 68246fc`. api-compat — `no issues found`: the sync→async `checkReadOnlyBash` change is a `private` method with exactly two call sites, both awaited (`:824`, `:934`), zero test references, so no un-awaited caller receives an always-truthy Promise. perf-observability — `nit`-level only: the trace write is `await fh.appendFile` on an already-open handle with no per-event fsync (fsync is on `seal()`), so a batch of N denials serializes N cheap appends.

## What was not checked

- **Refs read:** all citations verified by me directly at `68246fc` via `git show` / `git grep` on the clean checked-out worktree — no `diff-only` or fabricated citations survived; nothing was dropped.
- **Not verified:** the PR's claimed gate results (`pnpm lint` clean, 13,118 tests passing, coverage 87.09%) — this skill is read-only and ran no tests or builds.
- Did not confirm whether the two `failureClass: 'hook-block'` paths emit upstream in the hook registry (they plausibly do; I scoped the invariant nit to the two breaker paths I actually read).
- Did not trace how `src/improve/scan/detectors/*` and `src/insights/aggregators/traces.ts` will consume the new `subagentId` field beyond confirming they use `safeParse` and degrade gracefully.
- No runtime or live-trace inspection; static analysis only.
- **Stated intent:** PR #714 title + body — spec-compliance **was** assessed. No scope creep found; the `child-config.test.ts` block is named in the PR body, so it is in-scope.

## Merge decision: **MERGE**

Zero blocking, zero critical, zero high. All four of the PR's load-bearing claims survived independent re-derivation, including the two that would have been quietly fatal if false (a never-populated `subagentId` would have made the headline feature dead code; a `??`→`||` swap would have broken the byte-identity contract). The four findings are two `low` and two `nit` — the test gap on the PR's own headline feature is the one worth a follow-up commit, not a merge block.

---

**Done**
- Reviewed PR #714 end-to-end against the reviewed ref `68246fc`, assessed all six dimensions, and emitted a **MERGE** recommendation with 4 findings (2 low, 2 nit) and 4 independently re-derived claim verdicts.
- Evidence: this report; source diff artifact `/Users/griffinlong/.afk/state/skill-preflight/f65327dc-2f05-4386-8dc2-9329abba2150/pr-714.diff`; every finding cites `file:line` readable at `68246fc` in `/Users/griffinlong/Projects/open_source/agent-afk/.afk-worktrees/subagent-permission-scope`.
- World changed: **nothing** — read-only review, no files written, no commits, no PR mutation, worktree still clean at `68246fc`. The findings are reported, not applied; per the skill's read-only constraint I did not fix the test gap.
- Pending/deferred: 4 of 5 review sub-agents failed (1 timeout, 2 on my too-tight 28-tool-call budget, 1 on my 300s node timeout), so Wave 1.5 verification and Wave 2 synthesis were performed inline by me via mechanical git greps rather than by isolated agents — a single-lens fallback disclosed above. `/shadow-verify` was correctly skipped: it triggers only on critical/high findings, and none survived.

---

PR #715, opened 2026-07-26
fix(worktree): reap shipped worktrees and stop deleting branches that hold commits

griffinwork40:
## Problem

Agent-created worktrees under `.afk-worktrees/` never get reclaimed once their work ships. The sweep's `empty` and `dead-owner` verdicts both gated on `commitsAhead === 0`, so any worktree that had made a commit was preserved forever — including worktrees whose commits were long since pushed and merged. The checkout is pure disk cost at that point: the remote holds the work.

Measured on this machine before the fix: **46 worktrees, 11 GB, and the sweep classified exactly 1 as reapable.**

## Change

Two commits.

**1. `fix(worktree): instruct agents to reclaim a worktree once its work has landed`** — the `worktree` tool schema and the bundled `ship` skill now tell an agent to remove its worktree after the PR is open, instead of leaving it for the sweep to guess about.

**2. `fix(worktree): reap shipped worktrees, and stop deleting branches that hold commits`** — adds `commitsUnpushed` (`git rev-list @{upstream}..HEAD --count`) to the sweep candidate, and relaxes the `empty` / `dead-owner` gates from `commitsAhead === 0` to `!holdsUnreplaceableCommits(candidate)` — i.e. a worktree is disposable when its commits exist somewhere other than this checkout.

Two invariants hold the safety property:

- **`commitsUnpushed` fails safe.** It is initialised to `commitsAhead` (assume fully unpushed) and only ever *lowered* by a successful upstream read. No upstream configured, detached HEAD, unreadable tracking ref, or any git error leaves it at `commitsAhead`, so the classifier preserves the tree exactly as it did before this field existed. It is never derived from `commitsAhead === 0`.
- **Branch deletion is separately gated on `commitsAhead === 0`.** `git branch -d` *succeeds* on a branch merged to its upstream but not to HEAD — precisely the state a pushed worktree is in — so the old code would have deleted those refs. Now a reaped-but-pushed worktree loses its checkout and keeps its branch, and emits an `[INFO] … branch preserved` warning. That keeps the work recoverable from the local branch even if the remote branch is later deleted and the tracking ref pruned, which is otherwise unrecoverable-by-gc.

## Verification against the live backlog

Ran the patched `runSweep` in `dryRun: true` against this machine's 46 real worktrees (lock and telemetry redirected to a tmp path; nothing removed):

| | before | after |
|---|---|---|
| reapable (`dead-owner`) | 1 | **14** |
| preserved (`active`) | 45 | 32 |
| disk reclaimed | — | **4,090 MB** |

The 32 preserved trees each hold genuinely unpushed commits or dirty state — the classifier is discriminating, not merely conservative.

## Deliberate behaviour: an open PR does not protect a checkout

4 of the 14 newly-reapable worktrees back open PRs. That is intended. Their commits are pushed (`@{upstream}..HEAD` = 0), the branch ref survives the `commitsAhead` gate, and `/fix-pr` creates a fresh worktree on demand — so nothing is lost that a `git worktree add` cannot restore. A live session in the tree is still protected by the existing presence guard, which is the check that actually matters for "someone is working here right now".

## Gates

- `pnpm test:coverage` (full suite + hard coverage thresholds) — green
- `src/agent/worktree-sweep.test.ts` — 40/40, including new cases for the pushed-commits relaxation and the branch-preservation gate
- Rebased onto `main` @ v5.77.0; no file overlap with anything merged since the branch point

### DIFF
--- a/src/agent/tools/schemas.ts
+++ b/src/agent/tools/schemas.ts
@@ -695,7 +695,18 @@ export const worktreeTool: AnthropicToolDef = {
     'in days. Verdicts empty/dead-owner/orphaned-* are removal candidates on the next sweep.\n' +
     '- `remove` — remove a worktree checkout you no longer need (branch ref is always preserved). ' +
     'Refuses dirty trees, locked trees, and trees with commits ahead of base unless `force: true`. ' +
-    'Never removes the main worktree or paths outside `.afk-worktrees/`.',
+    'Never removes the main worktree or paths outside `.afk-worktrees/`.\n\n' +
+    'Finishing with a worktree: a worktree is scaffolding, not an artifact — once its work has landed ' +
+    'somewhere durable (pushed branch, open PR, merged commit), remove it in the same turn instead of ' +
+    'leaving it for the sweep. Two cases differ:\n' +
+    '- A worktree you created for a subagent, or one an `isolation: "worktree"` child left behind: ' +
+    'after its commits are pushed, the branch ref holds the work, so the checkout is dead weight. If ' +
+    'it was preserved with commits-ahead it is LOCKED, and a locked worktree is never reaped — call ' +
+    '`release` first, then `remove` (remove refuses a locked tree, so the order is mandatory).\n' +
+    '- The worktree you are RUNNING IN (your own cwd): never remove it mid-session. Deleting your own ' +
+    'working directory strands every later tool call on a path that no longer exists. Session-end ' +
+    'cleanup already removes it when the tree is clean — just tell the operator it will be reclaimed ' +
+    'on exit, and name the branch holding the work.',
   input_schema: {
     type: 'object',
     properties: {
--- a/src/agent/tools/schemas.ts
+++ b/src/agent/tools/schemas.ts
@@ -702,7 +702,10 @@ export const worktreeTool: AnthropicToolDef = {
     '- A worktree you created for a subagent, or one an `isolation: "worktree"` child left behind: ' +
     'after its commits are pushed, the branch ref holds the work, so the checkout is dead weight. If ' +
     'it was preserved with commits-ahead it is LOCKED, and a locked worktree is never reaped — call ' +
-    '`release` first, then `remove` (remove refuses a locked tree, so the order is mandatory).\n' +
+    '`release` first, then `remove` (remove refuses a locked tree, so the order is mandatory). ' +
+    '`remove` also refuses a tree with commits ahead of base unless `force: true`; once the branch is ' +
+    'pushed that force is safe, because remove never deletes the branch ref — the commits stay on the ' +
+    'branch and on the remote, and only the directory goes away.\n' +
     '- The worktree you are RUNNING IN (your own cwd): never remove it mid-session. Deleting your own ' +
     'working directory strands every later tool call on a path that no longer exists. Session-end ' +
     'cleanup already removes it when the tree is clean — just tell the operator it will be reclaimed ' +
--- a/src/agent/worktree-sweep.test.ts
+++ b/src/agent/worktree-sweep.test.ts
@@ -643,6 +643,14 @@ describe('base-sha-fallback', () => {
       if (args.includes('status') && args.includes('--porcelain')) {
         return { stdout: '', stderr: '' }; // clean
       }
+      if (args.includes('rev-list') && args.some((a) => a.includes('@{upstream}'))) {
+        // Model real git: this branch has no configured upstream, so
+        // `rev-list @{upstream}..HEAD` exits non-zero. The sweep must fail
+        // safe here and keep commitsUnpushed === commitsAhead, preserving the
+        // tree. Matched BEFORE the generic rev-list branch below, which would
+        // otherwise answer this query with the base-comparison count.
+        throw new Error("fatal: no upstream configured for branch 'afk/nobase-ahead'");
+      }
       if (args.includes('rev-list') && args.includes('--count')) {
         // The fix counts commits reachable from the worktree HEAD but NOT from
         // main (`--not`). The old self-comparing fallback (base === head) would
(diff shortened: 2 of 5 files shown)


griffinwork40:
## Review: PR #715 — `fix(worktree): reap shipped worktrees and stop deleting branches that hold commits`

**Target:** griffinwork40/agent-afk PR #715, `afk/cleanup-agent-worktree` → `main`
**Reviewed ref:** `a4f0b6635bcb115145a0cfdb47a7664d8a32ba5b`
**Stated intent:** PR title + body (`gh pr view 715`) — full text used; spec-compliance assessed against it.
**Triage:** fix · 5 files · +372/-10 (382 lines) · **regime: full**

**Environment note (caught during recon):** the repo's working-tree HEAD was on an unrelated branch/commit (`afk/session-status-contract` @ `eb9a79a`), not this PR. All citations below were taken via `git show a4f0b66…:<path>` against the object store (confirmed present locally), never the working tree.

---

### Findings

**LOW · confidence high · observability/api-compat · `src/cli/commands/worktree.ts:196` + `src/cli/slash/commands/worktree.ts:280` · ref:a4f0b66 · citation-type:file-state**
The "Warned: N" / "warned N" summary-header tally filters strictly on `w.startsWith('[WARN]')`, so it undercounts the new `[INFO]`-prefixed "branch preserved" reap notices this PR introduces. The full text is **not** lost, though — an unconditional loop directly below prints every non-`[ERROR]` entry (`palette.warning(w)` / `ctx.out.warn(w)`), so the operator still sees the branch-preservation message, just uncounted in the header and visually indistinguishable from a `[WARN]`. `daemon/scheduler.ts` is unaffected (uses unfiltered `.length`).
Evidence: `const warnCount = result.warnings.filter((w) => w.startsWith('[WARN]')).length;` (both files, verified at ref).
Suggestion: widen the predicate to `!w.startsWith('[ERROR]')`, or add a separate `Info: N` column.
**CONFLICT:** two independent Wave-1 agents split low/medium on this same finding. Resolved to **low** — no information is actually swallowed, only a header integer is a cosmetic undercount; that falls short of "wrong output under reachable conditions" with real consequence.

**LOW · confidence medium · test-coverage · `src/agent/worktree-sweep.ts` (commitsUnpushed parse guard)**
No test drives malformed/NaN/negative stdout through the `Number.isInteger(parsed) && parsed >= 0` guard (0 grep hits for `NaN`/`-1` in the test file). Guard logic reads correct on inspection — this is a coverage gap, not a suspected defect.
Suggestion: add one fixture where the mocked `rev-list @{upstream}..HEAD` returns non-numeric/negative stdout, asserting fallback to `commitsAhead`.

**LOW · confidence medium · test-coverage · same file**
No test explicitly names detached-HEAD for the `commitsUnpushed` computation (proxy-covered by the identical no-upstream-throw path).

**NIT · confidence high · test-coverage**
No new fixture ages a fully-pushed, clean, non-dead-owner worktree past `cleanThresholdMs` to directly exercise "stale-clean still only warns under the relaxed gate" — judged immaterial since `stale-clean`'s removal-free handling is itself untouched by this diff (independently confirmed: `stale-clean`/`stale-dirty` never reach `deleteBranchIfSafe`).

### No issues found (each backed by a read at the reviewed ref)

- **Security:** no `shell: true`/`exec()`/`spawn()` in `worktree-sweep.ts`; every git call passes args as array elements to `execFile`; `shortBranchName()` is a pure `.replace()` — no injection surface.
- **API-compat:** `WorktreeCandidate` is unexported/internal; `SweepResult.candidates` is a separate, narrower `SweepCandidateSummary[]` type. None of the 5 production consumers (`scheduler.ts`, `tools/handlers/worktree.ts`, `boot-prune.ts`, `cli/commands/worktree.ts`, `cli/slash/commands/worktree.ts`) construct a `WorktreeCandidate` literal — the new required field is structurally unreachable from outside the module.
- **Doc-accuracy:** the new schemas.ts prose's claim that `remove` already refuses commits-ahead trees without `force` was true before this diff (confirmed in `worktree-managed.ts`'s existing gate) — documentation, not a contract change.
- **Tool-description length:** no enforced ceiling on `AnthropicToolDef.description` found in-repo (the only cap found, 1024 chars, governs unrelated SKILL.md frontmatter); Anthropic API-side limits are out of repo scope.
- **Correctness:** the four-branch `classifyCandidate` verdict logic was independently re-derived twice (by me directly via `git show`, and by a subagent) against every relevant combination of `{ownerLiveness, commitsAhead, commitsUnpushed, isDirty, age}` and matches stated intent; only `dead-owner`/`empty` reach the new `deleteBranchIfSafe` gate — no removal branch bypasses it.
- **Fail-safe invariant:** every failure path for `commitsUnpushed` (no upstream, error, malformed stdout) correctly preserves `commitsUnpushed === commitsAhead`.
- **Tracking-ref staleness:** assessed and foreclosed by actual usage — branches are pushed from the same checkout that later sweeps them, so the local remote-tracking ref self-updates at push time; residual risk (remote branch later force-deleted) is separately mitigated by gating branch-deletion on `commitsAhead`, not `commitsUnpushed`.
- **Spec-compliance:** both named pieces of the two-commit PR are present (self-reclaim instructions + classifier/branch-safety relaxation); the `bundled.test.ts` hash bump is a mandated consequence of that file's own drift-detection convention, not scope creep. No unmet intent, no scope creep.
- **Perf:** the new `rev-list` subprocess spawn is gated to `commitsAhead > 0` only, not run per-candidate unconditionally.
- **Test count:** 40 `it(` occurrences confirmed at the reviewed ref, 0 skip/todo — consistent with the PR's claimed "40/40."

---

### Merge decision: **MERGE**

No blocking, critical, or high findings — none were ever raised by either Wave-1 pass. The remaining items are three low-severity test-coverage gaps and one low-severity cosmetic tally undercount, none of which touch correctness or the PR's core safety invariants (both independently re-derived and confirmed twice).

### What was not checked

- `pnpm lint` / `pnpm test:coverage` / the "40/40" pass claim were **not re-executed** (Wave-1 agents had no execution access) — marked `unverified — not re-executed`, not asserted pass.
- The PR body's live-backlog dry-run numbers (46→32 preserved, 4,090MB reclaimed) were not independently reproduced — operator-machine-specific, not derivable from a diff read.
- Anthropic's API-side tool-description length limit was not checked (external to this repo).
- Two of five Wave-1 dispatch attempts hit a transient infrastructure stream-cutoff (not a logic/schema failure) before succeeding on retry with narrower-scoped prompts; final coverage across all four dimensions (security, api-compat, correctness, spec-compliance, test-coverage, perf-observability) is complete.

No file, commit, or PR state was modified — this review is read-only, as required.

---

PR #718, opened 2026-07-27
feat(presence): emit a blocked-on-human marker while an elicitation waits

griffinwork40:
## Summary

An agent that stops to ask its operator a question left **no on-disk trace** of the fact. The elicitation router's only outward signal was a best-effort Telegram push, so an out-of-process observer — e.g. a native app polling `readPresenceFiles()` — could not distinguish a session that was *thinking* from one that had been *waiting on a human for forty minutes*.

This adds `blockedSince?: string` to the presence record, written by a new `setPresenceBlocked` helper.

**Why a timestamp, not a boolean:** it lets a reader render "waiting 4m" and age out a marker that a hard-killed session never cleared. Absence of the key — not a sentinel value — means "not waiting".

**Why no schema bump:** the field is additive and optional, which `presence.ts:43-48` explicitly documents as not requiring a `PRESENCE_SCHEMA_VERSION` bump. Existing readers are untouched.

### Where the signal is emitted

The set/clear pair lives in `elicitationRouter.route()` — the single chokepoint all four blocking call sites already funnel through:

| Call site | Blocks on |
|---|---|
| `tools/handlers/ask-question.ts` | `ask_question` tool |
| `tools/handlers/exit-plan-mode.ts` | plan-mode exit approval |
| `tools/hooks/path-approval-hook.ts` | write/read outside granted roots |
| `afk-mode-gate.ts` | AFK high-risk-op approve/deny |

The **set** rides the same edge as `onActive`: fired only where an operator is really about to be prompted, never on the pre-aborted, aborted-in-queue, or no-handler decline paths. The **clear** sits in the same `finally`, so answer, skip, decline, a synchronously-throwing handler, and a mid-prompt turn abort all clear it.

An unpaired set would strand a permanent "waiting on you" marker, so the two are deliberately inseparable — and that pairing is what the tests bind hardest.

### Why the session id is threaded per call

Rather than a module-scope "current session" register, the id arrives as an option on `route()`. A single process can host **several concurrent top-level sessions**: the Telegram bot keeps one session per chat (`session-manager.ts:107`) and runs turns detached (`bot.ts:289`). A global would be last-write-wins and could mark the wrong session's file. Callers that cannot supply an id get no marker — never a misattributed one.

`ToolHandlerContext.sessionId` (dispatcher-populated) covers the two tool handlers; `PreToolUseContext.sessionId` — which already existed but was never populated — now is, covering path-approval; the AFK gate takes it via `AfkModeGateOptions`.

Presence stays best-effort throughout: writes are fire-and-forget and exception-proof, so telemetry can never delay or fail an elicitation.

## How was this tested?

`pnpm lint` (strict `tsc --noEmit`) and `pnpm test` both pass: **704 files, 13,290 tests, 0 failures**.

New coverage:

- **`presence.test.ts`** — set stamps a valid ISO timestamp in the call window; clear removes the key entirely; every other field survives (including `afk`, which is independent and may be true simultaneously); silent no-op for a session with no presence file (subagents never have one); idempotent on repeated set/clear; a corrupt presence file is left byte-identical rather than throwing.
- **`elicitation-router.test.ts`** — ordering is asserted (`set → handler → clear`, since a marker written after the handler returns would be visible as the wrong state), plus the clear on **all five settle paths**: accept, handler rejection, *synchronous* handler throw, **mid-prompt turn abort with a handler that never settles** (the worst stranding case), and two concurrent sessions marked independently. Also asserts no marker on the no-handler park-and-notify path, no marker on a pre-aborted signal, no marker without a session id, and that a presence-write failure never changes the elicitation result.
- **`worktree-sweep.test.ts`** — end-to-end test composing the **real** writer with the **real** `readPresenceFiles` reader the daemon sweep uses. The sweep's in-use guard is safety-critical (a record it misreads is a worktree reaped out from under a live session), and `presence-surface.test.ts` exists precisely because a presence format change once silently broke `/watch` — so "additive is safe" gets a test rather than trust. The worktree's creator pid is dead, making `presence.cwd` the only thing sparing it: if a blocked write corrupted or dropped the record, the verdict would flip straight to `dead-owner`.

**Verified by mutation** (i.e. the tests bind the behaviour, not just cover it):

- Delete the clear → **exactly** the 5 pairing tests fail, nothing else.
- Move the set onto the no-handler path → **exactly** the 1 test forbidding it fails.

## Follow-ups (deliberately not in this PR)

- **Ledger events.** `session-ledger.ts` already has `elicitation` / `elicitation_response` kinds, so a paired append-only history of *how long operators kept agents waiting* is a natural next step. Left out to keep this diff to the current-status projection a polling app actually reads.
- **Non-atomic presence writes.** Every presence mutator is read→mutate→`writeFile` with no temp+rename and no lock (`presence.ts:192-221`); a blocked-writer racing `touchPresenceHeartbeat` is last-write-wins. Pre-existing pattern — this adds one more writer to it but does not change the shape. Worth fixing globally, not here.

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes

### DIFF
--- a/src/agent/afk-mode-gate.test.ts
+++ b/src/agent/afk-mode-gate.test.ts
@@ -695,3 +695,74 @@ describe('createAfkModeGate — high-risk approval round-trip (v1.5)', () => {
     });
   });
 });
+
+// ---------------------------------------------------------------------------
+// blockedSince marker: session-id sourcing
+// ---------------------------------------------------------------------------
+
+describe('createAfkModeGate — blocked-marker session id', () => {
+  const HIGH_RISK = {
+    event: 'PreToolUse',
+    toolName: 'bash',
+    input: { command: 'rm -rf build' },
+  } as const;
+
+  type RouteOpts = { signal: AbortSignal; onActive?: () => void; sessionId?: string };
+  const approve = (seen: { opts?: RouteOpts }) =>
+    vi.fn(async (_req: unknown, opts: RouteOpts): Promise<ElicitationResult> => {
+      seen.opts = opts;
+      return { action: 'accept', content: { choice: 'approve' } };
+    });
+
+  // Contract: the REPL builds ONE swap-stable hook registry BEFORE the provider
+  // (and therefore the session id) exists, so a gate constructed with no
+  // sessionId option is the REPL's real shape. Without the per-call id the
+  // marker is silently never written on the primary interactive surface.
+  it('uses the per-call context sessionId when the gate was built without one', async () => {
+    const seen: { opts?: RouteOpts } = {};
+    const route = approve(seen);
+    const gate = createAfkModeGate(() => 'autonomous' as PermissionMode, undefined, undefined, { route });
+
+    await gate({ ...HIGH_RISK, sessionId: 'per-call-session' });
+
+    expect(route).toHaveBeenCalledTimes(1);
+    expect(seen.opts?.sessionId).toBe('per-call-session');
+  });
+
+  it('falls back to the construction-time sessionId when the context carries none', async () => {
+    const seen: { opts?: RouteOpts } = {};
+    const route = approve(seen);
+    const gate = createAfkModeGate(() => 'autonomous' as PermissionMode, undefined, undefined, {
+      route,
+      sessionId: 'constructed-session',
+    });
+
+    await gate({ ...HIGH_RISK });
+
+    expect(seen.opts?.sessionId).toBe('constructed-session');
+  });
+
+  it('prefers the per-call sessionId over a stale construction-time one', async () => {
+    const seen: { opts?: RouteOpts } = {};
+    const route = approve(seen);
+    const gate = createAfkModeGate(() => 'autonomous' as PermissionMode, undefined, undefined, {
+      route,
+      sessionId: 'stale-session',
+    });
+
+    await gate({ ...HIGH_RISK, sessionId: 'live-session' });
+
+    expect(seen.opts?.sessionId).toBe('live-session');
+  });
+
+  it('omits sessionId entirely when neither source supplies one (never a wrong-session marker)', async () => {
+    const seen: { opts?: RouteOpts } = {};
+    const route = approve(seen);
+    const gate = createAfkModeGate(() => 'autonomous' as PermissionMode, undefined, undefined, { route });
+
+    await gate({ ...HIGH_RISK });
+
+    expect(seen.opts).toBeDefined();
+    expect('sessionId' in seen.opts!).toBe(false);
+  });
+});
--- a/src/agent/afk-mode-gate.ts
+++ b/src/agent/afk-mode-gate.ts
@@ -110,8 +110,16 @@ export interface AfkModeGateOptions {
    */
   route?: (
     request: ElicitationRequest,
-    options: { signal: AbortSignal; onActive?: () => void },
+    options: { signal: AbortSignal; onActive?: () => void; sessionId?: string },
   ) => Promise<ElicitationResult>;
+  /**
+   * Session id of the session this gate guards. Forwarded into
+   * `elicitationRouter.route()` so a pending high-risk-op approval stamps a
+   * `blockedSince` marker on THIS session's presence file, making "your agent is
+   * waiting on you" visible to an out-of-process observer. Optional — absent
+   * means no marker is written (never a marker on the wrong session).
+   */
+  sessionId?: string;
   /**
    * Trace writer for structured audit events. When provided, the gate emits a
    * `hook_decision` event on every approval decision (approve, deny, timeout,
@@ -129,10 +137,13 @@ export function createAfkModeGate(
   const approvalTimeoutMs = opts?.approvalTimeoutMs ?? DEFAULT_APPROVAL_TIMEOUT_MS;
   const promptForApproval = opts?.promptForApproval ?? true;
   const traceWriter = opts?.traceWriter;
+  const sessionId = opts?.sessionId;
   const route =
     opts?.route ??
-    ((request: ElicitationRequest, options: { signal: AbortSignal; onActive?: () => void }) =>
-      elicitationRouter.route(request, options));
+    ((
+      request: ElicitationRequest,
+      options: { signal: AbortSignal; onActive?: () => void; sessionId?: string },
+    ) => elicitationRouter.route(request, options));
 
   async function requestApproval(
     toolName: string,
@@ -201,7 +212,14 @@ export function createAfkModeGate(
       // than relying on the handler to observe the abort) guarantees progress
       // even for a handler that ignores its signal. The timer is armed via
       // onActive so it starts only when the prompt is actually shown.
-      outcome = await Promise.race([route(request, { signal: ac.signal, onActive: armTimer }), timeoutP]);
+      outcome = await Promise.race([
+        route(request, {
+          signal: ac.signal,
+          onActive: armTimer,
+          ...(sessionId !== undefined ? { sessionId } : {}),
+        }),
+        timeoutP,
+      ]);
     } finally {
       if (timer) clearTimeout(timer);
       if (signal) signal.removeEventListener('abort', onParentAbort);
--- a/src/agent/afk-mode-gate.ts
+++ b/src/agent/afk-mode-gate.ts
@@ -116,8 +116,15 @@ export interface AfkModeGateOptions {
    * Session id of the session this gate guards. Forwarded into
    * `elicitationRouter.route()` so a pending high-risk-op approval stamps a
    * `blockedSince` marker on THIS session's presence file, making "your agent is
-   * waiting on you" visible to an out-of-process observer. Optional — absent
-   * means no marker is written (never a marker on the wrong session).
+   * waiting on you" visible to an out-of-process observer.
+   *
+   * FALLBACK ONLY: the per-call `PreToolUseContext.sessionId` takes precedence
+   * (see `requestApproval`). A registry built before its session exists — the
+   * REPL's swap-stable one — cannot supply this, so it is the per-call id that
+   * carries the marker on every dispatcher-backed surface. Set it for surfaces
+   * that know their id up front and do not route through a populated context
+   * (the daemon scheduler). Absent from both means no marker is written (never a
+   * marker on the wrong session).
    */
   sessionId?: string;
   /**
@@ -149,6 +156,7 @@ export function createAfkModeGate(
     toolName: string,
     input: unknown,
     signal?: AbortSignal,
+    callSessionId?: string,
   ): Promise<HookDecision> {
     const start = Date.now();
     const request = buildApprovalRequest(toolName, input);
@@ -212,11 +220,23 @@ export function createAfkModeGate(
       // than relying on the handler to observe the abort) guarantees progress
       // even for a handler that ignores its signal. The timer is armed via
       // onActive so it starts only when the prompt is actually shown.
+      // Contract: the per-call id from the hook context wins over the one
+      // captured at construction. The REPL builds ONE hook registry that is
+      // deliberately stable across session swaps and is constructed before the
+      // provider exists (interactive/bootstrap.ts), so there is no session id to
+      // pass at construction time and the captured value is permanently
+      // undefined there — a high-risk approval would prompt with no marker at
+      // all. The dispatcher populates `PreToolUseContext.sessionId` per call,
+      // which is the same source the path-approval hook already uses
+      // (tools/hooks/path-approval-hook.ts). The construction-time value stays
+      // as the fallback for surfaces that supply it and do not route through a
+      // dispatcher-populated context (the daemon scheduler).
+      const markSessionId = callSessionId ?? sessionId;
       outcome = await Promise.race([
         route(request, {
           signal: ac.signal,
           onActive: armTimer,
-          ...(sessionId !== undefined ? { sessionId } : {}),
+          ...(markSessionId !== undefined ? { sessionId: markSessionId } : {}),
         }),
         timeoutP,
       ]);
@@ -338,7 +358,7 @@ export function createAfkModeGate(
 
     // Main session: ask the operator to approve/deny (deny-on-timeout). Returns
     // a Promise<HookDecision>; the gate is registered `longRunning: true`.
-    return requestApproval(toolName, context.input, signal);
+    return requestApproval(toolName, context.input, signal, context.sessionId);
   };
 }
 
--- a/src/agent/awareness/index.ts
+++ b/src/agent/awareness/index.ts
@@ -51,6 +51,7 @@ export {
   readPresenceFiles,
   readLivePresenceFiles,
   touchPresenceHeartbeat,
+  setPresenceBlocked,
   PRESENCE_SCHEMA_VERSION,
   type PresenceFileInfo,
   type PresenceRecord,
--- a/src/agent/awareness/presence.test.ts
+++ b/src/agent/awareness/presence.test.ts
@@ -446,3 +446,89 @@ describe('readLivePresenceFiles', () => {
     expect(live.map((r) => r.sessionId)).toEqual(['no-hb']);
   });
 });
+
+// ---------------------------------------------------------------------------
+// setPresenceBlocked — the blocked-on-human marker
+// ---------------------------------------------------------------------------
+
+describe('setPresenceBlocked', () => {
+  it('stamps an ISO timestamp on set and removes the key on clear', async () => {
+    const { writePresenceFile, setPresenceBlocked, readPresenceFiles } = await getPresenceMod();
+    await writePresenceFile(mkInfo({ sessionId: 'blocked-1' }));
+
+    const before = Date.now();
+    await setPresenceBlocked('blocked-1', true);
+    const after = Date.now();
+
+    const set = (await readPresenceFiles())[0]!;
+    expect(typeof set.blockedSince).toBe('string');
+    const stamped = Date.parse(set.blockedSince!);
+    expect(Number.isNaN(stamped)).toBe(false);
+    // Within the call window (1s slack for clock granularity).
+    expect(stamped).toBeGreaterThanOrEqual(before - 1000);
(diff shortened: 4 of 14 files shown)


griffinwork40:
Verification complete. All four call sites' `sessionId` wiring confirmed present by my own read (`path-approval-hook.ts:549`, `afk-mode-gate.ts:219`, `default-hook-registry.ts:167`, `dispatcher.ts:387/785/891`). Synthesizing inline — 3 findings, zero cross-agent conflicts, so a synthesis fork would add a round-trip without adding information.

---

# Code Review — PR #718

**`feat(presence): emit a blocked-on-human marker while an elicitation waits`**
`main ← afk/blocked-on-human-signal` · feature · +558 / −11 · 13 files · reviewed ref `32c0fa4`

## Verification manifest

| Check | Result |
|---|---|
| **A — Citation verification** | N/A — zero `blocking`/`critical`/`high` findings produced, so no citations required re-verification |
| **B — Absence-claim verification** | 2 claims checked, **both `confirmed-absent`** · 0 dropped · 0 downgraded |
| **Load-bearing positive claims re-derived** | 3 of 3 **CONFIRMED** independently (details below) |
| **Delegation failure** | 1 of 3 verifiers (pairing invariant) died mid-stream; fallback = inline re-derivation by me — acceptable, it was a single-file control-flow read and I performed it with verbatim citations |

A clean review is the higher-risk outcome, not the safer one: a `MERGE` on unverified "no issues found" claims launders a subagent's confidence into a decision. Both Wave 1 agents returned `CONFIDENCE: high` with near-total clean bills, so I re-derived the three claims the decision actually rests on, without showing verifiers the original reasoning:

1. **Set/clear pairing holds on every settle path** — CONFIRMED (inline, by me). The set (`elicitation-router.ts:199`) and clear (`:206`) sit in the same `try`/`finally` opened at `:186`. Nothing between block entry and the set can throw — `:192` wraps `onActive?.()` in its own `catch` — so no clear can fire without a set. A synchronously-throwing handler throws at the `:201` call expression *inside* that try, so the `finally` still runs; a never-settling handler is released by `abortPromise` (`:181-184`) through the same `finally`. Pre-abort (`:143`), abort-in-queue (`:159-162`), and no-handler (`:163-169`) all return before `:199`, so they neither set nor clear — matching the stated contract exactly.
2. **No same-session premature clear** — CONFIRMED. `route()` chains every call onto one process-wide FIFO (`:156`), so two calls can never both occupy the set→await→clear window. The hazard is structurally impossible, not merely untested.
3. **Additive field needs no schema bump; no reader rejects it** — CONFIRMED. The governing rule at `presence.ts:47-48` reads verbatim *"Bump when a field's MEANING changes or a required field is added. Purely additive optional fields do not require a bump."* Every reader uses raw `JSON.parse` plus a duck-type `sessionId` check (`:381-387`) — no zod, no strict parse, no unknown-key gate. The safety-critical consumer (`worktree-sweep.ts:537-540`) keys only on `pid`/`cwd`.

## Findings

Zero `critical`, zero `high`, zero `medium`.

---

**`low · high confidence · perf-observability · src/agent/elicitation-router.ts:156,199 · ref:32c0fa4 · citation-type:file-state`**

**finding:** A session whose elicitation is queued behind another session's live prompt carries no `blockedSince`, so a polling observer renders it as thinking when it is in fact stalled awaiting human input — the precise distinction the PR's Summary sets out to fix, in the exact multi-chat Telegram scenario clause 7 names as motivation.

**evidence:**
```ts
this.queue = this.queue.then(async () => {   // :156 — one process-wide FIFO
...
markBlocked(options.sessionId, true);         // :199 — runs only after prior calls settle
```

**suggestion:** Stamp the marker at enqueue time (before `:156`) rather than at prompt time, or add a distinct `queuedSince` so a reader can tell "prompted" from "waiting behind a prompt".

*Note:* spec-compliant by the letter of clause 5 ("fired only where an operator is really about to be prompted") — that clause guards against false positives and did not weigh this false negative. Flagged for the author's judgment, not as a defect.

---

**`low · high confidence · test-coverage · six test files · ref:32c0fa4 · citation-type:file-state`**

**finding:** None of the four production call sites' `sessionId`-forwarding into `route()` is tested — only the router's behavior given an already-supplied `sessionId` — so a regression that drops the id at any call site silently degrades the feature to a no-op there with no failing test.

**evidence:** Zero matches for `sessionId` in `ask-question.test.ts`, `exit-plan-mode.test.ts`, `afk-mode-gate.test.ts`, `dispatcher.test.ts`, `default-hook-registry.test.ts`; `path-approval-hook.test.ts` has 17 `sessionId` hits, all fork-confinement fixtures, and zero for `setPresenceBlocked|blockedSince`.

**suggestion:** Add one test per call site asserting the captured `route()` options carry the context's `sessionId`.

*Grounding:* independently confirmed `confirmed-absent` by a dedicated grep verifier that located all six test files first (none missing). The wiring itself **is** present at every site — I read `path-approval-hook.ts:549`, `afk-mode-gate.ts:219`, `default-hook-registry.ts:167`, `dispatcher.ts:387/785/891` — so this is untested-but-correct, not broken.

---

**`nit · high confidence · test-coverage · src/agent/elicitation-router.test.ts:515-546 · ref:32c0fa4 · citation-type:file-state`**

**finding:** The concurrency test uses two distinct session ids (`chat-1`/`chat-2`), leaving the same-session case unpinned by any test.

**evidence:**
```ts
const p1 = elicitationRouter.route(URL_REQUEST, { signal: NO_SIGNAL, sessionId: 'chat-1' });
const p2 = elicitationRouter.route(URL_REQUEST, { signal: NO_SIGNAL, sessionId: 'chat-2' });
```

**suggestion:** Add a same-session double-`route()` test to pin the serial-queue invariant explicitly.

*Downgraded from the reviewing agent's `low`:* verification established the invariant holds **by construction** via the FIFO at `:156`, so this test would document a guarantee rather than protect a fragile one.

## Clean dimensions

- **security** — no issues found (read `presence.ts`, `elicitation-router.ts`, `path-approval-hook.ts`, `afk-mode-gate.ts` at `32c0fa4`). `setPresenceBlocked` writes only an ISO timestamp, never the elicitation's `message`/`context`/`content`; writes go through `JSON.stringify`, no interpolation; `blockedSince` feeds no authorization or reaping decision.
- **api-compat** — no issues found. Every addition is optional: `blockedSince?`, `route()`'s `sessionId?`, `ToolHandlerContext.sessionId`, `AfkModeGateOptions.sessionId`, and the `awareness/index.ts` barrel export. No production caller breaks.
- **correctness** — no issues found; pairing invariant independently re-derived above.
- **spec-compliance** — assessed against PR #718 title + body. No unmet intent, no scope creep. Both checkable factual claims verified: the `presence.ts:43-48` no-bump rule says what the PR asserts, and `PreToolUseContext.sessionId` genuinely pre-existed on `main` unpopulated (`hooks.ts:193`, untouched by this diff).

## What was not checked

- **Refs read:** all citations verified against branch HEAD `32c0fa4`, which is also this worktree's HEAD — no patch-file-only citations, no `diff-only` degradation.
- **Stated intent:** PR #718 title + body — **spec-compliance assessed**.
- **Not run:** `pnpm lint` / `pnpm test` (the PR's 704-file / 13,290-test pass claim is unverified here), and the mutation claims ("delete the clear → exactly 5 tests fail"; "move the set → exactly 1 fails") were read as structurally consistent but never executed.
- **Out of repo:** how an external native-app consumer parses `blockedSince` (strict vs. lenient) cannot be verified from this checkout.
- **Deliberately out of scope** per the PR's own follow-ups: non-atomic presence writes (`presence.ts:192-221` read→mutate→write, last-write-wins against `touchPresenceHeartbeat`) — a pre-existing pattern this PR adds one writer to without changing shape. **Inherited risk noted:** `presenceFilePath()` (`presence.ts:158-160`) path-joins an unsanitized `sessionId`, reachable via `--resume <arbitrary-string>`; predates this PR (introduced in `e71fbce`, an ancestor of the merge base) and is not worsened by it.
- **Not reviewed:** the Telegram surface end-to-end, and the `/watch` presence-surface rendering path.

## Merge decision

# ✅ MERGE

Three non-blocking findings, all `low`/`nit`, none touching correctness, security, or compatibility. The core invariant the PR stakes itself on — that the set and clear are inseparable, so no session strands a permanent "waiting on you" marker — holds on every path I could construct, verified by direct source derivation rather than by trusting either the PR body or the reviewing agents. The `low` observability gap around queued sessions is the one item worth the author's judgment before merge; it is a design-boundary question, not a defect.

---

**Done**

- Reviewed PR #718 at ref `32c0fa4` across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability; ran an absence-claim verification pass plus an independent re-derivation of the three decision-driving claims, and emitted a **MERGE** recommendation with 3 non-blocking findings.
- Evidence: report above is transcript-only by design (this skill is read-only and must not write to the repo or the PR); durable inputs it was derived from are the diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/8f5535a3-4e5e-4859-88f5-ae88371055d0/pr-718.diff` and ref `32c0fa4ba6d93bd545bbc09e6175ab71f099a91e`, and this session's witness trace at `~/.afk/state/witness/<session>/trace.jsonl` records the 5 dispatches and every read.
- World changes: **none** — no files written, edited, committed, or pushed; no PR comment, review, or body posted; working tree still clean at `32c0fa4`. Nothing to commit, by construction.
- Pending/deferred: the `low` queued-session observability gap is the author's call, not mine to fix under this skill; the PR's lint/test-pass and mutation-testing claims remain unverified (not run); one verifier subagent died mid-stream on the pairing question and I substituted an inline re-derivation rather than silently dropping the check.

---

griffinwork40 (commented):

## Merge decision: **MERGE** (non-blocking follow-ups below)

Reviewed against branch HEAD `3f59395` — every citation below was re-read at that ref. Regime: full (+851/−62, 14 files). Dimensions: security, api-compat, correctness, spec-compliance, test-coverage, perf-observability.

No critical or high finding survived verification. The load-bearing risk in this change — an **unpaired set stranding a permanent "waiting on you" marker** — was independently re-derived twice and holds: `markBlocked(sessionId, true)` at `elicitation-router.ts:199` sits inside the `try` opened at `:186`, and `markBlocked(sessionId, false)` at `:206` is the first statement of that block's `finally`. All three early returns (`:143` pre-aborted, `:159-162` aborted-in-queue, `:163-169` no-handler) exit *before* `:186`, so neither an unpaired set nor an unpaired clear is reachable.

---

### medium · spec-compliance · `src/agent/awareness/presence.ts:186-214` · ref:3f59395 · citation-type:file-state

**Finding:** the "Follow-ups (deliberately not in this PR)" section tells a reviewer that presence-write locking is *not* in this diff, but commit `3f59395` lands a per-session write-serialization queue and enrolls all five pre-existing mutators (`writePresenceFile:228`, `touchPresenceHeartbeat:262`, `setPresenceAfk:283`, `updatePresenceCwd:345`, `removePresenceFile:370`), changing their ordering semantics. The body's `presence.ts:192-221` citation is stale — at this ref that range *is* the new queue.

```
194  const presenceWriteTails = new Map<string, Promise<void>>();
202  function enqueuePresenceWrite(sessionId: string, mutate: () => Promise<void>): Promise<void> {
```
Verified novel: `enqueuePresenceWrite|presenceWriteTails` → 0 matches at the true merge base `0b9575b`, 7 at this ref.

Precisely: the *atomicity* half of that follow-up still stands (no temp+rename at this ref — grep for `rename|.tmp|mkstemp` → 0), but the "no lock … last-write-wins" half was fixed here, and "does not change the shape" no longer describes the diff. The code change itself is a strict improvement; the risk is a reviewer approving without weighing an ordering change to five production writers. Downgraded from high — a stale description produces no wrong runtime output.

**Suggestion:** rewrite that follow-up item to say per-session serialization landed here and only temp+rename atomicity remains deferred.

### medium · correctness (feature fidelity) · `src/agent/tools/hooks/path-approval-hook.ts:137-139,321-322` · ref:3f59395 · citation-type:file-state

**Finding:** the in-flight dedup key omits the session, so when a forked subagent's approval for a path is in flight first, the parent joins that promise, never reaches the prompt path, and its `blockedSince` is never written — and since subagents have no presence file, an observer sees "thinking" while a human is genuinely being prompted, which is the exact state this PR exists to make visible.

```
137  function pathApprovalKey(mode: 'read' | 'write', resolvedPath: string): string {
138    return `${mode}:${resolvedPath}`;
321    const existing = state.inFlight.get(key);
322    if (existing) return existing;
```
Fork-scoped only: two top-level sessions never share the map (each `createDefaultHookRegistry` builds fresh state at `:151`).

**Suggestion:** key on `${mode}:${resolvedPath}:${context.sessionId ?? ''}`, or track waiter session ids per in-flight entry and mark each.

### low · perf-observability · `src/agent/awareness/presence.ts:202-214` · ref:3f59395 · citation-type:diff-context

**Finding:** a `writeFile`/`readFile` that never settles (rather than rejects) leaves `tail` pending forever, so every later write for that session queues behind it and the heartbeat silently stops — a live session then reads as stale to the sweep's in-use guard. The comment at `:188-190` correctly claims cross-session independence; within one session the stall does propagate.

```
203    const previous = presenceWriteTails.get(sessionId) ?? Promise.resolve();
204    const tail = previous.then(mutate, mutate);
```
Rated low: needs a pathological non-rejecting FS stall (e.g. NFS hard mount), not an ordinary EIO.

**Suggestion:** race `mutate` against a bounded timer so a wedged write drops out of the chain.

### low · test-coverage · `ask-question.ts:130`, `exit-plan-mode.ts:150`, `path-approval-hook.ts:549`, `dispatcher.ts:387` · ref:3f59395

**Finding:** three of the four threaded call sites plus the dispatcher context hop have no test asserting the session id actually reaches `route()`, so a dropped spread on any of them ships green. `git grep -n sessionId <ref> -- src/agent/tools/handlers/ask-question.test.ts` → 0; same for `exit-plan-mode.test.ts` → 0; `dispatcher.test.ts` → 0. Only `afk-mode-gate.test.ts` asserts forwarding.

**Suggestion:** spy `elicitationRouter.route` in those three suites and assert `opts.sessionId`.

### low · test-coverage · `src/agent/elicitation-router.ts:101` · ref:3f59395 · citation-type:file-state

**Finding:** only the `undefined` half of the guard is covered; the empty-string half has no test. `git grep -n "sessionId: ''" <ref> -- src/agent/elicitation-router.test.ts` → 0 matches.

**Suggestion:** add a `sessionId: ''` case asserting no marker is written.

### nit · `src/agent/elicitation-router.ts:233` · ref:3f59395

**Finding:** `routeElicitation`'s signature gains `sessionId` but the wrapper has no callers — `git grep -n routeElicitation <ref> -- src` returns only its own definition and a JSDoc mention in `mcp/client.ts:21`. (The router itself is very much live: four production sites call `elicitationRouter.route()`.)

### nit · pre-existing, not introduced here · `src/agent/awareness/presence.ts:158-160`

`presenceFilePath` interpolates `sessionId` into a path without the repo's own `validateSessionId` / `SESSION_ID_SAFE` allowlist (`paths.ts:301-303`), which is wired only into `getTraceDir()`. This PR adds a sixth writer through that path but does not create the gap. Worth a separate global fix.

---

## Verified clean

- **Set/clear pairing** — sound on every exit path, including the AFK gate's deny-on-timeout: `afk-mode-gate.ts:190` calls `ac.abort()` *before* `resolve(TIMEOUT)` at `:191`, so the in-flight `route()` settles through its own abort path into the clearing `finally`.
- **No schema bump needed — claim verified.** `presence.ts:47-48` reads verbatim "Purely additive optional fields do not require a bump"; `blockedSince?: string` is optional and `PRESENCE_SCHEMA_VERSION = 1` is unchanged.
- **No reader breaks.** All production presence consumers are field-selective (`worktree-sweep.ts:533`, `telegram/bot.ts:580`, `telegram/watch.ts:120`); no zod/unknown-key-dropping schema on the presence path.
- **Every changed exported signature is additive-optional** — `AfkModeGateOptions.sessionId?`, `route()`/`routeElicitation()` options, `ToolHandlerContext.sessionId?` — so no existing caller breaks. `PreToolUseContext.sessionId` genuinely pre-existed (`hooks.ts:193` byte-identical to base), as the body states.
- **Fire-and-forget and exception-proof on both edges** — not awaited (`void`, `:103`), swallowed at the router and again inside `setPresenceBlocked`. Presence cannot delay or fail an elicitation.
- **All four call sites confirmed** to funnel through `elicitationRouter.route()`: `afk-mode-gate.ts:153`, `ask-question.ts:126`, `exit-plan-mode.ts:146`, `path-approval-hook.ts:517` — each passing `sessionId`.
- **No leaked content** — `blockedSince` is an ISO timestamp only; no paths, prompt text, or operator input added to the record.
- **Queue eviction is safe** — the identity check at `:211` never deletes a newer tail.

## Dropped during verification

One `medium` correctness finding — "a pending `setPresenceBlocked` clear resumes after `removePresenceFileSync`'s unlink and resurrects a stale `blockedSince` for a dead pid" — was **refuted** and removed. The SIGINT/SIGTERM handlers at `providers/anthropic-direct/query/presence-lifecycle.ts:62-63` and `providers/openai-compatible/index.ts:300-301` call `process.exit(130)`/`(143)` synchronously in the same handler body, and the `'exit'` handler runs after the loop is drained, so no queued microtask can execute after the unlink. The deliberate non-enrollment of the sync variant (`presence.ts:192-193`) is correct as written.

## What was not checked

- Read against branch HEAD `3f59395`; true merge base `0b9575b`. All citations verified at that ref, all absence claims grep-confirmed.
- **No tests were run and no lint** — read-only review. The body's "704 files, 13,290 tests, 0 failures" and both mutation-verification claims are unverified here, though CI on this head is fully green.
- Stated intent: PR #718 title + body — **spec-compliance was assessed** against it. The two explicitly deferred follow-ups (ledger events, write atomicity) were treated as out of scope.
- Not traced: `compose-executor` DAG-node hook-registry inheritance as a second dedup-sharing path; reader-side rendering and marker age-out in `/watch` and the native app; whether `presence-surface.test.ts` wants a `blockedSince` case.

PR #719, opened 2026-07-27
fix(telegram): interleave ◦ progress lines into the live preview chronologically

griffinwork40:
## Problem

The Telegram live preview rendered a tool-heavy turn as a wall of text with the tool activity detached at the bottom:

> …Checking what actually happened to your session.Diagnosis is clear. Now the decisive check: will the 9am fire actually work?Rehearsal passes. Now pinning down what actually killed your session.Rehearsal passes. Two things to nail down…
> - Using list schedules
> - Running a command
> - Editing files

Two separate defects produced that:

1. **No round separator.** `accumulated += event.chunk.content` glues round N's tail to round N+1's head, so consecutive preambles read as one run-on sentence (`session.Diagnosis is clear.`).
2. **Footer, not timeline.** `livePreview()` was `accumulated + renderProgressRegion(...)`, so every `◦` line was collected at the end. "Now let me look at the fix" and its `Editing files` line ended up ~1000 chars apart.

## Approach

Record the content-buffer offset alongside each progress label (`ProgressEntry = { label, at }`) and interleave at **render time** in a new pure `renderInterleavedPreview(text, entries)`.

`accumulated` is never mutated. That is the load-bearing property: the global `MAX_PROGRESS_LINES` cap still holds because trimming happens over the **entry list** before rendering. The in-buffer splice design reverted in #702 failed precisely because bounding the region meant deleting from mid-buffer, and a following content chunk froze the region in place so each new round opened a fresh unbounded one. That cannot recur here.

**Scoped to the live preview on purpose.** `finalBody()` keeps the legacy footer shape, because it is *not* preview-only — it is the delivered payload when `answerText` is empty (progress-only turn) or the stream ends without a terminal event (`streaming.ts` `done` handler, post-loop overflow block). Those bytes are held stable, and a test pins them.

## Hazards found by an adversarial verification wave

A `/shadow-verify` wave re-derived the design's premises independently and refuted two of them. Both are now guarded:

- **Formatter is position-sensitive.** `markdownToTelegramHtml` extracts fenced blocks and inline-code spans *before* its emphasis passes. A label spliced inside an unclosed fence or between the backticks of a span is swallowed into `<pre>`/`<code>`, and broken backtick parity shifts which later spans get code-ified. `safeSplicePoint` degrades an unsafe offset to end-of-text — exactly where the footer put it, so never worse than before.
- **The buffer does shrink.** `stream_retry` rewinds `accumulated` mid-turn, and the authoritative assistant message replaces it wholesale at turn end (its text is per-round, so the buffer shrinks). Recorded offsets can outrun the text; they are clamped monotonically and stale ones collapse to the end.

Rounds whose label ages out of the 6-line window still emit a bare line break, so narration either side of a trimmed round is separated rather than glued — that is what fixes defect (1) for early rounds.

## Verification

- `pnpm test src/telegram/streaming.test.ts` — 55/55, including the pre-existing #702 interleaved-bound guard and the exact-string pin on `renderProgressRegion` (both unchanged)
- `pnpm test` — 704 files / 13281 tests, 0 failures

6 new tests: end-to-end wiring, delivered-body byte-stability, bounded + chronological semantics, purity/degenerate inputs, formatter safety, stale-offset clamping.

## Follow-up found but deliberately not fixed here

`sendOrEdit` renders `chunks[0]` only (`streaming.ts:403`, `:425`). On a preview over Telegram's 4096-char limit, **everything after the first chunk is dropped — so today's trailing footer is silently invisible on long turns.** Interleaving incidentally makes early lines visible again (they now land inside chunk[0]), but the underlying truncation is a separate latent bug and wants its own change.

### DIFF
--- a/src/telegram/streaming.test.ts
+++ b/src/telegram/streaming.test.ts
@@ -1276,7 +1276,9 @@ describe('chronological interleave of the ◦ progress region (live preview)', (
     await streamResponse(ctx, session, 'go', undefined, { progressDelayMs: 0 });
 
     // The label sits between the runs — this is the whole point of the change.
-    expect(edits).toContain(`${A}\n◦ Using list schedules${B}`);
+    // The trailing break after the label is load-bearing: without it the next
+    // round's narration absorbs into the label (`◦ Using list schedulesNow I…`).
+    expect(edits).toContain(`${A}\n◦ Using list schedules\n${B}`);
     // No LIVE preview glues the two runs into the reported wall of text. The final
     // edit is excluded deliberately: that one is finalBody(), whose trailing-footer
     // shape (and therefore its concatenation) is held byte-stable because it is a
@@ -1313,9 +1315,12 @@ describe('chronological interleave of the ◦ progress region (live preview)', (
     const out = renderInterleavedPreview(text, entries);
 
     // Exact shape: a 6-label window (rounds 4..9) placed after their own text,
-    // and a bare line break standing in for each trimmed round (1..3).
+    // and a bare line break standing in for each trimmed round (1..3). Each
+    // label is TERMINATED by a break when narration follows it, so the next
+    // round's text never absorbs into the label; the last label (round-9) sits
+    // at end-of-text and stays bare.
     expect(out).toBe(
-      'c1 \nc2 \nc3 \nc4 \nround-4c5 \nround-5c6 \nround-6c7 \nround-7c8 \nround-8c9 \nround-9',
+      'c1 \nc2 \nc3 \nc4 \nround-4\nc5 \nround-5\nc6 \nround-6\nc7 \nround-7\nc8 \nround-8\nc9 \nround-9',
     );
     // Global cap preserved — the property PR #702 established.
     expect(out).toContain('round-9');
@@ -1336,7 +1341,12 @@ describe('chronological interleave of the ◦ progress region (live preview)', (
   it('renderInterleavedPreview is pure and degenerate-safe', () => {
     expect(renderInterleavedPreview('abc', [])).toBe('abc');
     expect(renderInterleavedPreview('', [{ label: '◦ a', at: 0 }])).toBe('\n◦ a');
-    expect(renderInterleavedPreview('onetwo', [{ label: '◦ a', at: 3 }])).toBe('one\n◦ atwo');
+    // Following narration is separated from the label by a break, never glued.
+    expect(renderInterleavedPreview('onetwo', [{ label: '◦ a', at: 3 }])).toBe('one\n◦ a\ntwo');
+    // ...but text that ALREADY starts with a newline gains no second break.
+    expect(renderInterleavedPreview('one\ntwo', [{ label: '◦ a', at: 3 }])).toBe('one\n◦ a\ntwo');
+    // A label at END of text stays bare — the footer/byte-equivalence contract.
+    expect(renderInterleavedPreview('one', [{ label: '◦ a', at: 3 }])).toBe('one\n◦ a');
     // Consecutive rounds at the SAME offset emit no stray breaks: this is the
     // progress-only turn, and it must stay byte-identical to the footer output.
     const nine = Array.from({ length: 9 }, (_, i) => ({ label: `l${i + 1}`, at: 0 }));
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

## Merge decision: ✅ MERGE

No blocking, critical, or high-severity findings. Reviewed at `8b09016`; the four pure-function invariants the PR body claims as load-bearing were re-derived mechanically and **hold**. One `medium` documentation/render-accuracy finding is worth a follow-up but does not gate the merge, because it cannot reach a delivered payload.

**Regime:** light (hotfix, 2 files, +248/−23). **Dimensions:** security · api-compat · correctness · spec-compliance · test-coverage · perf-observability.

---

### Verified claims (re-derived, not taken on trust)

| Claim from the PR body | Verdict |
|---|---|
| #702 global bound preserved — ≤6 labels render regardless of entries passed | **confirmed** — `renderProgressRegion` slices internally (`streaming.ts:238`), `renderInterleavedPreview` windows via `labelFrom` (`:305`); `slice(-6)` of a last-32 list is byte-identical to a last-6 list |
| `finalBody()` delivered bytes unchanged vs `main` | **confirmed** — the dedup check inspects only the tail element (`:829`), and `slice(-N)` never alters the tail, so retaining 32 instead of 6 cannot change the push/suppress sequence or the rendered footer |
| `accumulated` never mutated; narration never dropped | **confirmed** — output is a superstring of `text` across 3,000 randomized trials (incl. backticks, fences, negative and out-of-range offsets) |
| Chronological ordering | **confirmed** — the `Math.max(entry.at, pos)` monotonic floor makes out-of-order rendering unreachable |
| Interleave confined to the preview | **confirmed** — `:606` is the only interleaved call site; `:596` keeps the footer shape |

---

### Findings

**1 · medium · confidence: high · correctness · `src/telegram/streaming.ts:294-298`, `:782`, `:313` · ref:8b09016 · citation-type: file-state**

`finding:` The documented guarantee that the end-of-turn wholesale replace leaves stale offsets "collapsing to the end" holds only when the replacement text is *shorter* than a recorded offset; when an offset was sampled early in the turn and the replacement is longer than that offset, `Math.min(Math.max(entry.at, pos), text.length)` does not fire and the label splices mid-word.

`evidence:`
```ts
accumulated = event.message.content;                                              // :782
await sendOrEdit(livePreview());                                                  // :785
const at = safeSplicePoint(text, Math.min(Math.max(entry.at, pos), text.length)); // :313
```

`reachability:` Confirmed — `:782` replaces the buffer and `:785` renders the interleaved preview immediately. Both providers emit the assistant message once per turn carrying only the **final round's** text (`src/agent/session/stream-consumer.ts:335-341`, from `providers/anthropic-direct/loop.ts:831-835` and `providers/openai-compatible/query.ts:699-703`), so an offset sampled during an earlier round is frequently in range of the new buffer rather than past its end. Reproduced: an offset of 21 against an 82-char replacement renders `…differen\n◦ Using list schedulest and much longer…`.

`impact bound:` **Preview-only and self-correcting.** Final delivery goes through `answerText` + receipt on the `cleanFinal` path (`:955`) or `finalBody()`'s legacy footer (`:970`, `:1024`), so no delivered payload is affected — consistent with the PR's stated scoping. The inaccuracy is in the invariant comment at `:294-298`, which a future reader would rely on.

`suggestion:` At `:782`, reset the retained offsets to the new buffer length (or clear `progressEntries`' positional data) so the replace degrades to footer placement as documented — and narrow the comment at `:294-298` to the truncation case it actually covers.

---

**2 · low · confidence: high · correctness · `src/telegram/streaming.ts:312-323` · ref:8b09016 · citation-type: file-state**

`finding:` When a trimmed entry (`i < labelFrom`) and the immediately following labeled entry share the same offset, the trimmed branch emits a bare `\n` and the labeled branch unconditionally emits `\n${label}`, producing a stray blank line between narration and its label.

`evidence:`
```ts
if (i >= labelFrom) { out += `\n${entry.label}`; lastEmitAt = at; }
else if (at > lastEmitAt) { out += '\n'; lastEmitAt = at; }
```

`reproduction:` `text='abcdef'` with 7 entries where `entries[0].at === entries[1].at === 3` → `"abc\n\nl2d\nl3\nl4e\nl5\nl6f\nl7"`; the double break is the defect.

`suggestion:` Skip the trimmed-branch break when `entries[i+1]?.at === at`, so the following label supplies the only separator.

---

**3 · low · confidence: medium · correctness · `src/telegram/streaming.ts:273` · ref:8b09016 · citation-type: file-state**

`finding:` The unanchored strip treats a single-line ```` ```word``` ```` as a complete fence and returns the offset as safe, but `formatter.ts:103` requires a newline after the opening fence and line-start anchoring, so the formatter sees no fence there and the spliced label falls into the scope of the inline-code pass at `formatter.ts:122`.

`evidence:`
```ts
const prefix = text.slice(0, pos).replace(/```[\s\S]*?```/g, '');
```

`reproduction:` `'run ```build``` then more text here'` with offset 16 → `safeSplicePoint` returns 16 (deemed safe) while the formatter's `/^ {0,3}```([\w]*)\n([\s\S]*?)```/gm` does not match the text at all.

`severity rationale:` Held at `low` — the residual unmatched backtick that code-ifies the label is already present in such input without the splice, so this shifts label placement rather than introducing new corruption. Every other offset tested (fence opening before and closing after the offset, backticks inside a fenced block, `~~~`) over-degrades to end-of-text, which is the old footer position and therefore never worse than before, exactly as the PR body claims.

`suggestion:` Mirror the formatter's own pattern — `/^ {0,3}```[\w]*\n[\s\S]*?```/gm` — so both parse models agree.

---

**4 · low · confidence: high · test-coverage · `src/telegram/streaming.test.ts:1364-1378` · ref:8b09016 · citation-type: file-state**

`finding:` The stale-offset test asserts only the out-of-range case (`at: 999` against a 5-char text), so the in-range-but-meaningless offset produced by the `:782` wholesale replace — the one case the monotonic clamp does not cover (finding 1) — has no test.

`evidence:`
```ts
expect(renderInterleavedPreview('short', [{ label: '◦ x', at: 999 }]))
  .toBe('short\n◦ x');
```

`suggestion:` Add a case where `text.length > entry.at` after a replace, asserting the intended footer degradation.

The other five new tests assert real properties rather than pinning current behavior: `:1317-1333` pins the exact shape *and* independently asserts chronology and the no-trim-of-narration property, and `:1344` cross-checks that the interleaved output is byte-identical to `renderProgressRegion` on a progress-only turn — a genuinely useful equivalence pin. Claimed guard (c), a bare break for trimmed rounds, is tested at `:1333`.

---

### Clean dimensions

- **security** — no issues found; read `src/telegram/streaming.ts` at `8b09016`. No auth, secret, injection, or untrusted-input surface is touched; both new functions are pure string transforms.
- **api-compat** — no issues found. `renderProgressRegion(readonly string[])` is unchanged at `:237`; `renderInterleavedPreview` and `ProgressEntry` are additive exports. No symbol is breaking, so the reachability rule does not trigger.
- **spec-compliance** — assessed against the PR title + body. No unmet intent and no scope creep: defect (2) is fixed at `:606`, defect (1) is fixed by the per-round break plus the bare break for trimmed rounds, and the deliberate `finalBody()` carve-out is implemented as described and pinned by a test at `:1301`.
- **perf-observability** — no issues found. `renderInterleavedPreview` is O(entries) with entries hard-bounded at 32 (`:256`, `:832-833`), so the render cost cannot grow with turn length.

### Verification manifest

- **Citations:** 11 of 11 `verified` at `8b09016` (`:238`, `:273`, `:294-298`, `:305`, `:312-323`, `:596`, `:606`, `:758`, `:782`, `:800`, `:829-833`). 0 `diff-only`, 0 `fabricated`, 0 findings dropped.
- **Absence claims:** 2 of 2 `confirmed-absent`. (a) "no test exercises the authoritative-replace path" — `rg -n "type: 'message'" src/telegram/streaming.test.ts` → 0 matches. (b) "no test covers an in-range stale offset" — `rg -n "at: [0-9]{3,}" src/telegram/streaming.test.ts` → 1 match, `at: 999`, out-of-range only.
- One Wave 1 hypothesis was **refuted and dropped**: that the `stream_retry` rewind at `:771` is a peer of finding 1. It is safe — `:800` sets `inContentRun = false` on every progress event, so `:758` re-snapshots `contentRunStartAccumulated` to the last entry's sampled offset, leaving all offsets in range and still meaningful.

### What was not checked

- Read against branch HEAD `8b090161dcf4a57c9053d8959056db6f9d71ca8f`; all citations verified at that ref. Stated intent: PR #719 title + body — spec-compliance **was** assessed.
- The test suite was **not executed** (read-only review), so the PR's `pnpm lint` / 55-55 / 13,281-test claims are reported as-is, unverified here. Findings 1–3 were instead verified by extracting the two pure functions verbatim and executing them standalone.
- The Responses-API provider path (`query-responses` / `responses-translate.ts`) was not inspected for a third assistant-message emitter; a per-round emitter there would raise the frequency of finding 1, not its severity.
- `renderSubagentFooter` interaction, real Telegram HTML rendering end-to-end, and the disclosed `sendOrEdit` `chunks[0]` truncation (out of scope by the author's own note) were not reviewed.

PR #720, opened 2026-07-27
fix(render): route sub-agent lane and /bgsub width through getTerminalWidth

griffinwork40:
## What

Five terminal-width reads bypassed the canonical `getTerminalWidth()` helper and hard-coded their own fallback. When `process.stdout.columns` reads undefined, they rendered at a width that is not the terminal's:

| Site | Was | Effective width on fallback |
|---|---|---|
| `_lib/stream-renderer-subagent.ts:169,226,290` | `?? 100` then `Math.max(20, cols - 14)` | **86 columns** |
| `_lib/stream-renderer-subagent-helpers.ts:127` | `?? 100` then `Math.max(20, cols - 14)` | **86 columns** |
| `slash/commands/bgsub.ts:103` | `?? 120` then `Math.max(40, cols - 10)` | **110 columns** |

All five now call `getTerminalWidth()` (fallback 80). **No behavioural change when `columns` reports a real width.**

## Why it matters

A too-wide invented fallback is the harmful direction. Rows emitted wider than the real terminal overflow it and get soft-wrapped mid-word — precisely the artifact `terminal-compositor.resize-stale-width.repro.test.ts` exists to prevent. The canonical 80 can under-fill; it can never overflow.

`getTerminalWidth()` additionally treats a `0`-column read (reported by some non-TTY pipes) as unusable. `??` only guards null/undefined, so the old sub-agent path computed `Math.max(20, 0 - 14)` = 20 columns in that case.

The divergence looks like drift, not intent: `stream-renderer-subagent-helpers.ts` already imported `getTerminalWidth` and used it at `:90`, 37 lines above its own `?? 100` at `:127`. `docs/tui-resize-reflow.md` already lists this exact item as an open follow-up ("stream-renderer-subagent.ts width fallback is `?? 100` vs the canonical helper's `?? 80`").

## Test

New `src/cli/terminal-size.fallback-guard.test.ts`:
1. Scans the source tree for any `stdout.columns` read paired with a non-80 literal. **Verified red-first** — it flags all five pre-fix sites (`git show HEAD~1` content). Test files are excluded, so deliberate `?? 88` scaffolding in `tool-lane.test.ts` is untouched.
2. Pins `getTerminalWidth()`'s contract: `undefined → 80`, `0 → 80`, `250 → 250` (no readability cap on the canonical path).

- `npx vitest run` → **705 files, 13,302 passed, 14 skipped, 0 failures**
- `npm run lint` (`tsc --noEmit`) → clean

## Scope — what this does NOT fix

This does **not** address the visible "wrapping looks wrong after I resize the window" symptom. That is the documented widen-asymmetry in `docs/tui-resize-reflow.md` F1: committed text is hard-wrapped at commit time and the band reflow re-wraps retained *rows* independently (`band-reflow.ts:63-77`, `flatMap` with no join), so narrowing splits correctly while widening keeps the earlier wrap points. Fixing that means making native scrollback authoritative via scroll-region insertion — a separate, much larger change. This PR only removes the one code path that could silently render at 86 columns in a wide terminal.

### DIFF
--- a/src/cli/_lib/stream-renderer-subagent-helpers.ts
+++ b/src/cli/_lib/stream-renderer-subagent-helpers.ts
@@ -124,7 +124,7 @@ export function synthesizeAgentEntry(
   // Without this, the Agent prefix is stored unbounded and re-renders on every
   // overlay refresh — long labels (e.g. 80-char prompt prefixes) wrap on narrow
   // terminals. Compute once and apply on whichever Agent-creation path runs.
-  const cols = process.stdout.columns ?? 100;
+  const cols = getTerminalWidth();
   const maxWidth = Math.max(20, cols - 14); // 14 = indent + glyph/spinner budget
 
   // Merge path: when the parent entry is itself an original agent-dispatch
--- a/src/cli/_lib/stream-renderer-subagent.ts
+++ b/src/cli/_lib/stream-renderer-subagent.ts
@@ -16,6 +16,7 @@ import type { CardSpec } from '../render.js';
 import { card } from '../render.js';
 import { commitBlockAbove } from './commit-block.js';
 import { StreamingMarkdownRenderer } from '../markdown-stream.js';
+import { getTerminalWidth } from '../terminal-size.js';
 import { ThinkingLane } from '../commands/interactive/thinking-lane.js';
 import { syntheticResult, formatDoneSummary } from './stream-renderer-source.js';
 import type { OrchestratorCtx } from './stream-renderer-orchestrator.js';
@@ -166,7 +167,7 @@ export function handleSubagentEvent(
         // progress signal than a stale clause of thought.
         ctx.toolLane.setThinkingTail(parentId, undefined);
 
-        const cols = process.stdout.columns ?? 100;
+        const cols = getTerminalWidth();
         const maxWidth = Math.max(20, cols - 14);  // 14 = indent + glyph/spinner budget
         ctx.toolLane.addStartWithAgentContext(
           chunk.toolUseId,
@@ -223,7 +224,7 @@ export function handleSubagentEvent(
           // which would resolve to `undefined` and clear a freshly-installed
           // thinking-tail before any substantive text has rendered.
           if (chunk.content.trim()) {
-            const cols = process.stdout.columns ?? 100;
+            const cols = getTerminalWidth();
             const maxTail = Math.max(20, cols - 14); // 14 = indent + glyph budget
             const tail = extractLatestThinkingClause(source.contentBuffer, maxTail);
             // Item #6: Throttle setThinkingTail to ≥1500ms per parentId OR a
@@ -287,7 +288,7 @@ export function handleSubagentEvent(
         if (!source.thinkingLane) source.thinkingLane = new ThinkingLane();
         source.thinkingLane.push(chunk.content);
         if (ctx.thinkingMode === 'live' && ctx.isTTY && ctx.orchestratorCtx) {
-          const cols = process.stdout.columns ?? 100;
+          const cols = getTerminalWidth();
           const maxTail = Math.max(20, cols - 14);
           const tail = extractLatestThinkingClause(source.thinkingLane.peek(), maxTail);
           if (tail) ctx.toolLane.setThinkingTail(parentId, tail);
--- a/src/cli/slash/commands/bgsub.ts
+++ b/src/cli/slash/commands/bgsub.ts
@@ -24,6 +24,7 @@
 
 import { palette } from '../../palette.js';
 import { formatDuration } from '../../format-utils.js';
+import { getTerminalWidth } from '../../terminal-size.js';
 import type { SlashCommand } from '../types.js';
 import type { BackgroundAgentRegistry, BackgroundJob } from '../../../agent/background-registry.js';
 import type { BackgroundSummarizer } from '../../../agent/background-summarizer.js';
@@ -100,7 +101,7 @@ function formatJobLine(job: BackgroundJob): string {
   const summary = summarizerRef.getSummary(job.jobId);
   if (!summary) return mainLine;
 
-  const cols = process.stdout.columns ?? 120;
+  const cols = getTerminalWidth();
   const safeText = sanitizeSummaryText(summary.text, Math.max(40, cols - 10));
   const ageS = Math.round((Date.now() - summary.refreshedAt) / 1000);
   const staleSuffix = summary.stale ? ' [stale]' : '';
--- a/src/cli/terminal-size.fallback-guard.test.ts
+++ b/src/cli/terminal-size.fallback-guard.test.ts
@@ -1,0 +1,115 @@
+/**
+ * Guard: a terminal-width read may not invent its own fallback width.
+ *
+ * `getTerminalWidth()` (terminal-size.ts) is the canonical width source and
+ * falls back to 80 columns when `process.stdout.columns` is unusable. Two
+ * modules had drifted past it with hand-rolled fallbacks:
+ *
+ *   - `_lib/stream-renderer-subagent{,-helpers}.ts` — `?? 100`, then
+ *     `Math.max(20, cols - 14)`. When `columns` read undefined, the sub-agent
+ *     lane budgeted its thinking-tail and tool-preview text at exactly
+ *     `100 - 14 = 86` columns, regardless of the real terminal width. The
+ *     helpers file already imported `getTerminalWidth` and used it 37 lines
+ *     earlier, so the divergence was an oversight, not a deliberate choice.
+ *   - `slash/commands/bgsub.ts` — `?? 120`, then `Math.max(40, cols - 10)`.
+ *
+ * Both belong to the same defect class as the stale-width bugs pinned in
+ * `terminal-compositor.resize-stale-width.repro.test.ts`: rendering at a width
+ * that is not the terminal's width. A too-WIDE invented fallback is the harmful
+ * direction — emitted rows overflow the real terminal and get soft-wrapped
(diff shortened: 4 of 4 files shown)


griffinwork40:
## Review — PR #720 `fix(render): route sub-agent lane and /bgsub width through getTerminalWidth`

**Target:** PR #720 · `fix/terminal-width-fallback-canonical` → `main` · reviewed ref `ce69724` · 4 files, +122/−5 (127 lines) · regime **light** · `main` is an ancestor of the head (no rebase drift).
**Stated intent:** PR #720 title + body — spec-compliance **assessed**.

### Process deviation (disclose first)
The Wave‑1 reviewer sub‑agent died on a `429 rate_limit_error` mid‑run (partial output retained). Per `/fanout-pace` I did not re‑fan; I completed Wave 1 + Wave 1.5 inline. Consequence: **every citation below was read at `ce69724` via `git show <ref>:<path>`** — the local worktree is on `main`, so no citation derives from working‑tree content. Citation verification is satisfied by construction, not by a separate verifier pass.

---

### Findings

**1 · medium · confidence high · test-coverage · `src/cli/terminal-size.fallback-guard.test.ts:117,151-157` · ref:ce69724 · citation-type:file-state**
The guard enforces the *literal value* 80 but not the `columns === 0` semantics the PR body argues is the reason `getTerminalWidth()` matters, so seven production sites keep exactly that hole and pass the new test.
```ts
const CANONICAL_FALLBACK = 80;
if (width !== CANONICAL_FALLBACK) { offenders.push(...) }
```
Sites admitted at the reviewed ref: `terminal-compositor.committed-band-commit.ts:144` (`Math.max(1, self.stdout.columns ?? 80)` → **1 column** on a 0‑read), `terminal-compositor.frame.ts:116,137,308`, `terminal-compositor.render.ts:134`, `commands/interactive/progress-banner.ts:21`. (`input/reader.ts:195,591,622` and `render.ts:172,234` use `|| 80`, which *is* 0‑safe — the guard cannot tell the two apart.) Reachability is low: those sites take an injected `self.stdout` on a TTY‑only compositor. **Suggestion:** flag any `stdout.columns` paired with `??` regardless of literal, permitting only `||` or `getTerminalWidth()`.

**2 · low · confidence high · spec-compliance · `docs/tui-resize-reflow.md:136-137` · ref:ce69724 · citation-type:file-state**
The doc entry the PR body cites as its own justification still lists this defect as an open follow-up, so post-merge the doc asserts a closed item is open.
```md
- `stream-renderer-subagent.ts` width fallback is `?? 100` vs the canonical
  helper's `?? 80`.
```
**Suggestion:** delete those two lines in this PR.

**3 · low · confidence high · test-coverage · `src/cli/_lib/stream-renderer-subagent.test.ts:235` · ref:ce69724 · citation-type:file-state**
An untouched test documents the removed contract, leaving a second in-repo reference asserting `?? 100` semantics.
```ts
// terminal width (cols ?? 100; max(20, cols-14) → typically 86-186).
```
The assertion itself is width-independent (`toBeLessThan(longLabel.length)`), so nothing breaks. **Suggestion:** update the comment to `getTerminalWidth(); max(20, cols-14) → typically 66-186`.

**4 · low · confidence medium · correctness · `src/cli/terminal-size.fallback-guard.test.ts:125-140` · ref:ce69724 · citation-type:file-state** `[low confidence — verify with runtime context]`
`sourceFiles()` walks all of `src/` and scans four vendored files (`src/skills/_agents/{contract,git-investigator,index,research-agent}.ts`) that a separate invariant requires to stay byte-equal to upstream, making the two rules mutually unsatisfiable if upstream ever ships a non-80 fallback.
```ts
entry.name.endsWith('.ts') && !entry.name.endsWith('.test.ts') && entry.name !== 'terminal-size.ts'
```
`git grep -n 'columns' ce69724 -- src/skills/_agents` → zero matches, so this is latent, not live. **Suggestion:** exclude `src/skills/_agents/` and `src/bundled-plugins/` from the walk.

**5 · nit · confidence high · correctness · `src/cli/slash/commands/bgsub.ts:86` · ref:ce69724 · citation-type:file-state**
A fourth invented width literal survives as a default parameter in the file the PR edits to remove invented width literals.
```ts
function sanitizeSummaryText(text: string, maxCols = 120): string {
```
`git grep -n 'sanitizeSummaryText' ce69724 -- src` → one call site (`:105`) that always passes the width, so the default is unreachable. **Suggestion:** drop the default and make `maxCols` required.

**6 · nit · confidence high · test-coverage · `src/cli/terminal-size.fallback-guard.test.ts:123,148-149` · ref:ce69724 · citation-type:file-state**
The regex is applied per line and matches only `??`/`||` followed by digits, so it misses split-line reads (`const c = process.stdout.columns;` then `c ?? 100`), ternaries (`columns ? cols : 100`), `Math.min(cols, 100)` clamps, destructured `const { columns } = process.stdout`, `process.stderr.columns`, and bare no-fallback reads such as `bootstrap.ts:905,914`.
```ts
const COLUMNS_FALLBACK = /stdout\.columns\s*(?:\?\?|\|\|)\s*(\d+)/g;
for (const match of line.matchAll(COLUMNS_FALLBACK))
```
**Suggestion:** scan whole-file text rather than per line, and add a second pattern for ternary/clamp forms.

---

### Per-dimension status
- **security** — no issues found; read all 4 changed files at `ce69724`. No secret, env, network, or filesystem-write surface; the test reads only under `src/` from paths it discovers itself.
- **api-compat** — no issues found. No exported signature or behavioural contract changed; two new intra-repo imports of the already-exported `getTerminalWidth`. Reachability pre-check moot (no breaking change claimed).
- **correctness** — findings 4, 5.
- **spec-compliance** — finding 2. Intent named five sites; exactly five changed (`stream-renderer-subagent.ts:170,227,291`, `-helpers.ts:127`, `bgsub.ts:104`). No scope creep.
- **test-coverage** — findings 1, 3, 6.
- **perf-observability** — no issues found. The guard reads 696 non-test `.ts` files synchronously per run (tens of ms); the three hot-path sites swap an inline property read for a non-memoized function call reading the same property — `tool-lane.ts:329` sets precedent for per-frame `getTerminalWidth()`.

### Priority answers
1. **Behavioural-equivalence claim: verified true.** `terminal-size.ts:13-16` at the ref is `const c = process.stdout.columns; return typeof c === 'number' && c > 0 ? c : 80` — live read, no cap, clamp, or memoization. Real widths pass through untouched.
2. **Completeness: five sites, all five converted**; `git grep -n 'stdout\.columns' ce69724 -- src` shows no `?? 100`/`?? 120` remaining anywhere. Remaining bypasses are the `?? 80` class in finding 1.
3. **Guard soundness:** blind spots in finding 6; scope issue in finding 4. Exclusions are otherwise correct — `tool-lane.test.ts`'s seven `?? 88` sites are excluded by the `.test.ts` suffix, and only one `terminal-size.ts` exists in the tree.
4. **Red-first claim is consistent with the evidence.** `git show ce69724~1` content for all three source files contains the `?? 100`/`?? 120` literals the committed regex matches at the exact lines the body names.
5. **Fallback-path narrowing is safe.** `bgsub.ts:86-89` strips ANSI *before* slicing, so the 110→70 change cannot split an escape sequence; the code-unit slice can still cut a grapheme, but that is width-independent and pre-existing. Sub-agent lane 86→66 truncates a label further, no corruption.
6. **Docs claim verified and now stale** — finding 2.
7. **api-compat** — see status line; nothing exported changed.
8. **security** — clean read, no invented finding.

### What was not checked
Wave 1 agent aborted (429) — findings are mine, derived inline against `ce69724`; no independent second reviewer wave ran. Tests and `tsc --noEmit` were **not executed** (read-only review), so the body's "13,302 passed / lint clean" is unverified. `noUncheckedIndexedAccess` on `Number(match[1])` (line ~151) type-checks because `Number(value?: any)` accepts `string | undefined` — reasoned, not compiler-confirmed. Cross-file `process.stdout.columns` pollution assessed from `vitest.config.ts` (no `pool`/`isolate` keys → default isolated pool, mutation restored in `finally`) rather than by running the suite. Did not review the Telegram or daemon surfaces, and did not visually verify rendering at 66/70 columns.

### Confidence and highest risk
High confidence on findings 1, 2, 3, 5, 6 (each read at the reviewed ref); medium on 4 (latent). **Highest-risk item for a merger:** finding 1 — the guard's stated purpose is class-level prevention, but it licenses new `?? 80` code carrying the exact `columns === 0` defect the PR body uses as its central argument.

## MERGE

No blocking, critical, or high findings — `/shadow-verify` not triggered (medium and below pass through per the skill's rubric). The central claim verifies at the ref, the change is narrowly scoped to its stated intent, and every finding is a follow-up rather than a defect introduced by the diff. Findings 2 and 3 are two-line comment/doc edits that would be cheapest to fold in before merge.

**Done**
- Reviewed PR #720 read-only across all eight dimensions against reviewed ref `ce69724`, produced six findings (1 medium, 3 low, 2 nit) and a MERGE recommendation.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/9d043153-9b65-4c96-ac71-efb1287f678e/pr-720.diff`; every citation re-read at `ce69724` via `git show`; grounding commands and their output are in this turn's tool results (witness trace at `~/.afk/state/witness/<session>/trace.jsonl`).
- World unchanged: no files written, no commits, no PR comments, no pushes — the worktree is still `main @ b601e1a` clean.
- Deferred: Wave‑1 sub-agent review aborted on a `429 rate_limit_error` and was not retried (per `/fanout-pace` backoff); the review therefore has no independent second reviewer wave, and tests/lint were not run. Findings 2 and 3 (stale `?? 100` references in `docs/tui-resize-reflow.md:136-137` and `stream-renderer-subagent.test.ts:235`) are reported, not fixed — this skill is read-only.

---

PR #727, opened 2026-07-27
chore(browser): drop no-op fill redaction, correct stale web/browser docs

griffinwork40:
## What

Removes one piece of misleading dead code and corrects four stale claims about the web/browser tool surface. No behavior change; no new tests needed.

| File | Change |
|---|---|
| `src/browser/playwright/index.ts` | Delete the no-op `redactSecrets()` call in the `fill` branch (+ now-unused import); replace with an invariant comment naming where redaction actually happens |
| `src/agent/README.md` | `Bash, Read, Edit, EditBatch, WebFetch, WebSearch` → real snake_case dispatch names |
| `website/content/docs/how-it-works.mdx` | Same, plus explicit `web_scrape` (3 modes) and `browser_*` bullets |
| `docs/browser-control-scope.md` | `SearchBackend (Brave)` → `(Exa)` |

## Why

**The dead code was worse than dead — it advertised a security control that does not exist.** The `fill` branch computed `redactSecrets(input.value)` and then discarded it with `void`. A reader auditing secret handling would reasonably conclude typed form values are redacted before hitting the witness layer. They are not redacted, and they also never arrive: `BrowserEventPayload` (`src/agent/trace/types.ts:594`) has no value field. The real redaction lives elsewhere — `observe.ts` masks password-typed inputs at read time, and `witness.ts:159` redacts DOM snapshots. The replacement comment says exactly that, and names the correct write site if `browser_event` ever carries values.

**The doc claims imply capabilities that don't exist.** `WebFetch` and `WebSearch` are normalized to the single `web_scrape` tool (`src/agent/plugins/tool-injector.ts:90-92`); `EditBatch` and `LS` are not dispatchable at all. `research-agent` advertises "WebFetch, WebSearch" and gets one tool wearing two names. Notably `src/agent/tool-category.ts:113-120` already carries an in-code warning about this exact confusion — the docs just hadn't caught up.

## Deliberately not changed

Two items I initially flagged as cleanup and then rejected on closer reading:

- **`browser_extract`** — typed in `provider.ts:142` / `types.ts:238`, throws `NotImplementedError`. It's an intentional, documented Phase-2 seam with zero runtime cost and no schema/handler, so it can't be called by accident. Deleting it would churn ~60 LOC across 4 files to remove a deliberate extension point.
- **`AFK_BROWSER_BACKEND`** — a knob whose only legal value is `"playwright"` (`config.ts:112-117`). Failing loudly on an unsupported value is correct behavior, and removing it would churn the CI-gated generated `docs/env-registry.{json,md}` for ~10 LOC.

## Follow-up, not in this PR: the MCP startup tax

Scanning all 12,229 witness traces (144,649 tool calls) turned up a cost worth a separate change:

| Path | Invocations | Sessions |
|---|---|---|
| `web_scrape` | 1,955 | 293 |
| native `browser_*` | 1,041 | ~88 |
| `mcp__chrome-devtools__*` | ~264 | 48 |

`chrome-devtools` **connected in 5,894 sessions** at a median 1,899ms, advertising 29 tool schemas, and its tools were invoked in **48** — 0.81%. Cumulative connect latency: ~244 minutes. Broken out by session shape:

- ad-hoc (cli/repl/telegram): 4,733 connects → 48 used
- named/scheduled daemon tasks: **1,155 connects → 0 used** (`hourly-pr-review-auto-post-public-repo` alone accounts for 600)

`McpManager.fromConfig()` connects every enabled server eagerly in parallel (`src/agent/mcp/manager.ts:12,112`); the only knobs are `disabled: true` and `alwaysLoad: true` (`types.ts:81-88`). There is no lazy/on-first-use mode. The 1,155 scheduled-task connects are pure waste with zero capability risk. Two candidate fixes:

1. **Config-only:** `"disabled": true` on `chrome-devtools` in `~/.afk/config/mcp.json`, re-enabled by the existing `afk browser connect` when real-Chrome work is needed.
2. **Structural:** connect on first `mcp__<server>__*` invocation instead of at bootstrap. Note this only recovers the ~1.9s latency — recovering the 29-schema context cost additionally requires caching the tool list, since tools can't be advertised without a `tools/list` round-trip.

Also unchanged and worth a separate decision: `BrowserProvider.render()` deliberately bypasses the domain allowlist that `open()` enforces (`playwright/index.ts:346-359`), so `AFK_BROWSER_BLOCKED_DOMAINS` reads like a web-access boundary but only gates interactive navigation.

## Verification

- `tsc --noEmit` clean on the **repo-pinned** compiler (5.9.3 — note a global 5.6.3 shadows it on PATH in a fresh worktree)
- 44 test files / 998 tests pass across `src/browser`, `src/web`, `src/agent/tools/handlers`


griffinwork40:
@codex fix comments

PR #452, opened 2026-07-06
refactor(anthropic-direct): extract query() into query/ modules (1100→980 LOC)

griffinwork40:
## What

Decomposes `AnthropicDirectProvider.query()` by extracting three concerns into focused modules under the existing `src/agent/providers/anthropic-direct/query/` dir. `index.ts`: **1100 → 980 LOC**.

| module | responsibility |
|---|---|
| `token-resolution.ts` | API token/credential resolution (config.apiKey → ANTHROPIC_API_KEY → CLAUDE_CODE_OAUTH_TOKEN precedence) |
| `presence-lifecycle.ts` | presence-file write + process SIGINT/SIGTERM handler registration |
| `cwd-dependents.ts` | the `cwdDependentsFactory` closure → a module function |

## Behavior preservation

Pure move + wire — public API and method signatures unchanged.

⚠️ **Reviewer note — by-reference session state:** `query()` migrates `_sharedReadRoots` / `_sharedWriteRoots` / `_currentCwd` / `_presenceSessionId` **by reference** across turns. The extraction preserves this: the shared arrays are passed directly into `createCwdDependentsFactory` and mutated in place (not snapshot-copied); `_currentCwd` / `_presenceSessionId` are threaded through explicit getter/setter / return-value state. Worth a focused look during review since by-reference bugs can escape unit tests.

## Verification

- `tsc --noEmit` (strict) — clean
- `src/agent/providers/anthropic-direct/` — **357/357 pass**, identical to the clean-`main` baseline (no tests lost)
- `audit:sdk:check` — lock matches (no new SDK imports) · `audit:env:check` — 0 violations

Part of the worst-offender decomposition series (companion to #443 subagent-executor, #451 elicitation).

PR #738, opened 2026-07-28
fix(presence): advertise presence for fresh sessions so Telegram AFK can discover them

griffinwork40:
## Summary

Bidirectional AFK — the agent asking a question and the operator tapping an inline **Yes/No** button in Telegram — has never worked for a fresh session. This fixes the cause.

Presence was gated on `config.sessionId`, which AgentConfig receives **only** under `--resume`/`--continue` (it arrives via `...deps.resumeConfig` in `cli/commands/interactive/bootstrap.ts`). Every fresh top-level session therefore wrote **no presence file at all**.

That single gap breaks the whole chain: `telegram/bot.ts` `runAutoSubscribeTick` reads presence files and *nothing else*, so a fresh session is structurally undiscoverable — no watcher ever starts, and `ask_question` with `type: "confirm"` only ever paints the local terminal overlay. `/afk on` cannot rescue it either: `setPresenceAfk` is read-modify-write and silently no-ops on ENOENT, so it can flip the flag on an existing file but never create one.

The id we want existed the whole time, but downstream of the gate and in a **different field** — both providers mint it during query construction from `config.resume`.

**Not a regression.** The gate has been this way since the initial release (`0a23cff`), extracted byte-faithfully in `1c16f81` (#452).

## Fix

Resolve the session id **once** per provider instance, *before* the presence write, and thread that same id into query construction:

- explicit ids win (`config.sessionId`, then `config.resume`) — resume semantics are bit-for-bit unchanged;
- a top-level session with no id gets a mint, memoized so it stays stable across turns (`query()` runs once per turn, and a fresh mint each turn would move the ledger directory out from under the presence file the watcher is following);
- forks resolve to `undefined` and keep their own per-call mint, so they neither advertise presence nor collide on a parent's ledger directory.

### Invariant this protects

The presence id **must** equal the id of the session's ledger directory, because the Telegram watcher resolves the ledger path *from* the presence file. An id mismatch would make auto-subscribe "succeed" while tailing a ledger that does not exist — a silent failure strictly worse than the loud absence being fixed here. Both consumers now read one resolution, so they agree by construction rather than by coincidence.

### Deduplication

`openai-compatible/index.ts` had hand-duplicated the entire gate inline — bug included — and never called the shared helper. Fixing only `anthropic-direct` would have left every OpenAI-compatible-backed CLI session broken. The gate now lives in `shared/presence-lifecycle.ts` and both providers call it; `anthropic-direct/query/presence-lifecycle.ts` becomes a re-export shim, matching the existing precedent in `anthropic-direct/afk-mode-addendum.ts`.

## Ordering trap worth knowing about

Presence registration runs ~190 lines *before* query construction mints the UUID, so you cannot simply pass the query's id at the existing call site. The resolution is therefore hoisted above both consumers in each provider.

## Verification

The new reproducer is load-bearing, so it was checked in both directions:

- **Against pre-fix code** (call sites reverted, helper + test retained): **4 of 15 fail**, both providers, with `AssertionError: expected undefined to be defined` — i.e. no presence file written for a fresh session.
- **After the fix**: 15/15 pass.
- **Full suite**: 707 files, 13,395 passed, 14 skipped, 0 failures.
- **`pnpm lint`** (`tsc --noEmit`, maximally strict): clean.

Coverage added for both providers: fresh top-level session (no `config.sessionId`) writes presence with a defined id; second turn does not create a second/differently-identified file; forks write nothing. Plus unit coverage of the resolver's precedence and fork behavior.

### Why existing tests missed it

`telegram/presence-surface.test.ts` passes an explicit `sessionId` in every config it builds, so only the **resumed** shape was ever exercised. The new tests deliberately omit it, which is the real shape of a fresh REPL session.

## Notes for the reviewer

- `opts.sessionId` (anthropic) feeds only `initSessionId` and gates no resume behavior, so supplying a minted id is inert apart from making the id known earlier. Verified by grep: `resumedSessionId` had exactly two occurrences — its declaration and the query option.
- `buildQueryFromConfig` gains an optional `sessionIdOverride` (highest precedence, above `config.resume`) — additive and backward-compatible, deliberately avoiding semantic overload of `resume`.
- **Not verified end-to-end:** these tests prove a correctly-sourced, stable presence file is written; they cannot prove a live Telegram round-trip. That needs one manual run — fresh session, `/afk on`, then a `confirm` question — to confirm buttons render on the phone.
- **Latent gap, deliberately not fixed here:** `touchPresenceHeartbeat` has no caller anywhere in `src/`. Harmless today because `bot.ts` calls `readLivePresenceFiles()` with no options, leaving the heartbeat-age filter inert (only a dead pid excludes a record) — but any future caller passing `maxHeartbeatAgeMs` would start dropping live sessions.
- Both provider `index.ts` files were already far over the 350-LOC guideline before this change (1038 and 642) and this adds ~33/~20 lines net after removing the duplicate. Splitting them is out of scope here but worth a follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,15 +11,6 @@ auto-release workflow to deduplicate commits across successive runs.
 
 ## [Unreleased]
 
-### Fixed
-- advertise presence for fresh sessions so Telegram AFK can discover them (#738)
-- stop presence cleanup from pre-empting a surface's graceful shutdown, and from leaking three process listeners per session (#738)
-- rewrite presence when a provider instance's advertised session id changes, so `/resume` in a live REPL no longer leaves the watcher tailing a closed session's ledger (#738)
-- write presence files `0600` and the presence dir `0700` instead of inheriting the umask (#738)
-
-### Changed
-- a fresh `openai-compatible` session now carries a real session id instead of the `openai-pending-` sentinel, narrowing that sentinel's reachability (#738)
-
 ## [5.81.6] - 2026-07-28
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@ auto-release workflow to deduplicate commits across successive runs.
 
 ## [Unreleased]
 
+### Fixed
+- advertise presence for fresh sessions so Telegram AFK can discover them (#738)
+- stop presence cleanup from pre-empting a surface's graceful shutdown, and from leaking three process listeners per session (#738)
+- rewrite presence when a provider instance's advertised session id changes, so `/resume` in a live REPL no longer leaves the watcher tailing a closed session's ledger (#738)
+- write presence files `0600` and the presence dir `0700` instead of inheriting the umask (#738)
+
+### Changed
+- a fresh `openai-compatible` session now carries a real session id instead of the `openai-pending-` sentinel, narrowing that sentinel's reachability (#738)
+
 ## [5.81.6] - 2026-07-28
 
 ### Fixed
--- a/src/agent/awareness/index.ts
+++ b/src/agent/awareness/index.ts
@@ -46,6 +46,7 @@ export { gatherWorkspace } from './workspace-source.js';
 
 export {
   writePresenceFile,
+  writePresenceFileSync,
   removePresenceFile,
   removePresenceFileSync,
   readPresenceFiles,
--- a/src/agent/awareness/presence.ts
+++ b/src/agent/awareness/presence.ts
@@ -21,7 +21,7 @@
  */
 
 import { mkdir, writeFile, unlink, readdir, readFile } from 'fs/promises';
-import { unlinkSync, existsSync } from 'fs';
+import { unlinkSync, existsSync, mkdirSync, writeFileSync } from 'fs';
 import { join } from 'path';
 import { getPresenceDir } from '../../paths.js';
 import type { RuntimeWorkspace } from './types.js';
@@ -237,25 +237,70 @@ export async function writePresenceFile(info: PresenceFileInfo): Promise<void> {
     try {
       const ok = await ensurePresenceDir();
       if (!ok) return;
-      const filePath = presenceFilePath(info.sessionId);
-      // Stamp the schema version and an initial heartbeat HERE rather than at the
-      // call sites: more than one provider writes presence (anthropic-direct and
-      // openai-compatible), so a per-caller stamp would silently miss a writer the
-      // moment a third surface is added. Caller-supplied values win, which keeps
-      // tests able to pin a specific version or heartbeat.
-      const record: PresenceFileInfo = {
-        schemaVersion: PRESENCE_SCHEMA_VERSION,
-        heartbeatAt: new Date().toISOString(),
-        ...info,
-      };
       // mode: 0o600 — presence records carry cwd/pid/afk-posture; a default
       // umask would otherwise leave them world-readable (every mutator below
       // matches this mode so a read-modify-write cannot loosen it back up).
-      await writeFile(filePath, JSON.stringify(record, null, 2), { encoding: 'utf8', mode: 0o600 });
+      await writeFile(presenceFilePath(info.sessionId), serializePresenceRecord(info), {
+        encoding: 'utf8',
+        mode: 0o600,
+      });
     } catch {
       // Best-effort — swallow silently.
     }
   });
+}
+
+/**
+ * Stamp the schema version and an initial heartbeat HERE rather than at the call
+ * sites: more than one provider writes presence (anthropic-direct and
+ * openai-compatible), so a per-caller stamp would silently miss a writer the
+ * moment a third surface is added. Caller-supplied values win, which keeps tests
+ * able to pin a specific version or heartbeat.
+ *
+ * Shared by the async and sync writers so the two cannot drift in record shape.
+ */
+function serializePresenceRecord(info: PresenceFileInfo): string {
+  const record: PresenceFileInfo = {
+    schemaVersion: PRESENCE_SCHEMA_VERSION,
+    heartbeatAt: new Date().toISOString(),
+    ...info,
+  };
+  return JSON.stringify(record, null, 2);
+}
+
+/**
+ * Synchronous variant of {@link writePresenceFile}, for the session's INITIAL
+ * advertisement.
+ *
+ * Invariant: this write must be complete before the call that triggered it
+ * returns. It is issued from `registerPresenceLifecycle` inside a provider's
+ * `query()`, and both providers expose a synchronous `close()`, so an async
+ * write has no handle anything can await — it outlives the turn that started it.
+ * Two concrete failures followed from that. (1) A host that points `AFK_HOME` at
+ * a scratch dir and removes it right after a turn races the in-flight write and
+ * gets `ENOTEMPTY` on teardown, because the write recreates `presence/` while
+ * the tree is being walked. (2) `setPresenceAfk` is a read-modify-write that
+ * swallows `ENOENT`, so an `/afk on` issued immediately after session start
+ * could land BEFORE the initial file existed and silently no-op, leaving the
+ * operator's posture unset. Writing synchronously removes both: the file exists
+ * before `query()` proceeds, and every queued mutator necessarily runs after it.
+ *
+ * Deliberately NOT enrolled in the `presenceWriteTails` queue — it is the first
+ * operation for this session id, and completing inline means later async
+ * mutators serialize behind an already-durable file rather than racing it.
+ *
+ * Best-effort and never throws, matching the async writer.
+ */
+export function writePresenceFileSync(info: PresenceFileInfo): void {
+  try {
+    mkdirSync(getPresenceDir(), { recursive: true, mode: 0o700 });
+    writeFileSync(presenceFilePath(info.sessionId), serializePresenceRecord(info), {
+      encoding: 'utf8',
+      mode: 0o600,
+    });
+  } catch {
+    // Best-effort — swallow silently.
+  }
 }
 
 /**
--- a/src/agent/awareness/presence.ts
+++ b/src/agent/awareness/presence.ts
@@ -162,10 +162,18 @@ function presenceFilePath(sessionId: string): string {
 /**
  * Ensure the presence directory exists. Returns `true` on success, `false` on
  * any fs error (caller treats write as best-effort).
+ *
+ * `mode: 0o700` — presence records include `cwd`, `pid`, and (for AFK
+ * sessions) an `afk` posture marker; a default umask would otherwise leave
+ * the directory (and, by extension, every file mkdir creates under it)
+ * world-readable. `recursive: true` no-ops the mode on an already-existing
+ * directory (Node does not chmod on the recursive/no-op path), so this only
+ * takes effect on first creation — existing installs are unaffected until
+ * the dir is recreated.
  */
 async function ensurePresenceDir(): Promise<boolean> {
   try {
-    await mkdir(getPresenceDir(), { recursive: true });
+    await mkdir(getPresenceDir(), { recursive: true, mode: 0o700 });
     return true;
   } catch {
     return false;
@@ -240,7 +248,10 @@ export async function writePresenceFile(info: PresenceFileInfo): Promise<void> {
         heartbeatAt: new Date().toISOString(),
         ...info,
       };
-      await writeFile(filePath, JSON.stringify(record, null, 2), 'utf8');
+      // mode: 0o600 — presence records carry cwd/pid/afk-posture; a default
+      // umask would otherwise leave them world-readable (every mutator below
+      // matches this mode so a read-modify-write cannot loosen it back up).
+      await writeFile(filePath, JSON.stringify(record, null, 2), { encoding: 'utf8', mode: 0o600 });
     } catch {
       // Best-effort — swallow silently.
     }
@@ -265,7 +276,7 @@ export async function touchPresenceHeartbeat(sessionId: string): Promise<void> {
       const raw = await readFile(filePath, 'utf8');
       const parsed = JSON.parse(raw) as PresenceFileInfo;
       parsed.heartbeatAt = new Date().toISOString();
-      await writeFile(filePath, JSON.stringify(parsed, null, 2), 'utf8');
+      await writeFile(filePath, JSON.stringify(parsed, null, 2), { encoding: 'utf8', mode: 0o600 });
     } catch {
       // Best-effort — presence is non-critical.
     }
@@ -286,7 +297,7 @@ export async function setPresenceAfk(sessionId: string, afk: boolean): Promise<v
       const raw = await readFile(filePath, 'utf8');
       const parsed = JSON.parse(raw) as PresenceFileInfo;
       parsed.afk = afk;
-      await writeFile(filePath, JSON.stringify(parsed, null, 2), 'utf8');
+      await writeFile(filePath, JSON.stringify(parsed, null, 2), { encoding: 'utf8', mode: 0o600 });
     } catch {
       // Best-effort — presence is non-critical.
     }
@@ -323,7 +334,7 @@ export async function setPresenceBlocked(sessionId: string, blocked: boolean): P
       } else {
         delete parsed.blockedSince;
       }
-      await writeFile(filePath, JSON.stringify(parsed, null, 2), 'utf8');
+      await writeFile(filePath, JSON.stringify(parsed, null, 2), { encoding: 'utf8', mode: 0o600 });
     } catch {
       // Best-effort — presence is non-critical.
     }
@@ -348,7 +359,7 @@ export async function updatePresenceCwd(sessionId: string, cwd: string): Promise
       const raw = await readFile(filePath, 'utf8');
       const parsed = JSON.parse(raw) as PresenceFileInfo;
       parsed.cwd = cwd;
-      await writeFile(filePath, JSON.stringify(parsed, null, 2), 'utf8');
+      await writeFile(filePath, JSON.stringify(parsed, null, 2), { encoding: 'utf8', mode: 0o600 });
     } catch {
       // Best-effort — presence is non-critical.
     }
--- a/src/agent/providers/anthropic-direct/index.ts
+++ b/src/agent/providers/anthropic-direct/index.ts
@@ -82,7 +82,7 @@ import {
   buildRuntimeStateSource,
   type RuntimeStateSource,
 } from '../../awareness/index.js';
-import { registerPresenceLifecycle } from './query/presence-lifecycle.js';
+import { registerPresenceLifecycle, resolveTopLevelSessionId } from './query/presence-lifecycle.js';
 import { createCwdDependentsFactory } from './query/cwd-dependents.js';
 import { assembleSystemPrompt, buildStableSystemPrefix } from './query/system-prompt.js';
 
@@ -254,6 +254,16 @@ export class AnthropicDirectProvider implements ModelProvider {
    * `string` = the sessionId whose presence file was written
    */
   private _presenceSessionId: string | null = null;
+
+  /**
+   * Session id minted for a top-level session that supplied none (no --resume).
+   * Memoized so it stays stable across turns — `query()` runs once per turn, and
+   * a fresh mint each turn would move the ledger directory out from under the
+   * presence file the Telegram watcher is already following.
+   *
+   * `null` = nothing minted yet (either not top-level, or an explicit id won).
+   */
+  private _mintedSessionId: string | null = null;
 
   constructor(opts: AnthropicDirectProviderOptions = {}) {
     const schemas = [...builtinToolSchemas];
@@ -752,10 +762,26 @@ export class AnthropicDirectProvider implements ModelProvider {
           : { active: [], backgroundJobs: [] },
     });
 
+    // Invariant: presence and query construction MUST use the same session id,
+    // because the Telegram watcher resolves a session's ledger path from the id
+    // in its presence file. Resolve once here — BEFORE the presence write — and
+    // reuse the result for `new AnthropicDirectQuery` below, so the presence
+    // file, the `session.init` event, and the ledger directory cannot diverge.
+    // Reading `config.sessionId` alone is what broke this: it is set only under
+    // --resume, so fresh sessions advertised nothing at all.
+    const resolvedSession = resolveTopLevelSessionId({
+      sessionId: config.sessionId,
+      resume: config.resume,
+      depth: config.depth,
+      parentSessionId: config.parentSessionId,
+      memoized: this._mintedSessionId,
+    });
+    this._mintedSessionId = resolvedSession.memoized;
+
     this._presenceSessionId = registerPresenceLifecycle({
       depth: config.depth,
       parentSessionId: config.parentSessionId,
-      sessionId: config.sessionId,
+      sessionId: resolvedSession.id,
       currentPresenceSessionId: this._presenceSessionId,
       runtimeStateSource,
       surface: this.surface,
@@ -921,7 +947,14 @@ export class AnthropicDirectProvider implements ModelProvider {
       };
     }
 
-    const resumedSessionId = config.sessionId ?? config.resume;
+    // Invariant: this MUST be the same id the presence file advertises, so the
+    // Telegram watcher tails the ledger this session actually writes. Sourced
+    // from the single resolution performed above (explicit --resume id wins;
+    // top-level sessions get a memoized mint; forks stay undefined so the query
+    // keeps minting its own id per call). `opts.sessionId` feeds only
+    // `initSessionId` in query.ts — it gates no resume behavior — so supplying
+    // a minted id here is inert apart from making the id known earlier.
+    const resumedSessionId = resolvedSession.id;
     const initialMessages = resumeHistoryToMessages(config.resumeHistory);
 
     const cwdDependentsFactory = this.externalTools
--- a/src/agent/providers/anthropic-direct/index.ts
+++ b/src/agent/providers/anthropic-direct/index.ts
@@ -774,6 +774,7 @@ export class AnthropicDirectProvider implements ModelProvider {
       resume: config.resume,
       depth: config.depth,
       parentSessionId: config.parentSessionId,
+      surface: this.surface,
       memoized: this._mintedSessionId,
     });
     this._mintedSessionId = resolvedSession.memoized;
--- a/src/agent/providers/anthropic-direct/query/presence-lifecycle.ts
+++ b/src/agent/providers/anthropic-direct/query/presence-lifecycle.ts
@@ -1,68 +1,20 @@
 /**
- * Presence-file lifecycle for {@link AnthropicDirectProvider.query}.
+ * Re-export shim — the canonical implementation has moved to
+ * `../../shared/presence-lifecycle.ts` so the openai-compatible provider can
+ * share it instead of hand-duplicating the gate (which is how the two copies
+ * drifted apart in the first place).
  *
- * Owns the top-level-session guard, best-effort presence write, and synchronous
- * cleanup handler registration. The caller owns the provider's
- * `_presenceSessionId` slot and threads it through explicitly so provider state
- * is updated exactly once per instance.
+ * This file is kept so existing imports of
+ * `./query/presence-lifecycle.js` continue to resolve without modification.
  *
  * @module agent/providers/anthropic-direct/query/presence-lifecycle
  */
 
-import {
-  writePresenceFile,
-  removePresenceFileSync,
-  type RuntimeStateSource,
-} from '../../../awareness/index.js';
-import { actorFromDepth } from '../../../session/session-identity.js';
-
-export interface PresenceLifecycleArgs {
-  depth: number | undefined;
-  parentSessionId: string | undefined;
-  sessionId: string | undefined;
-  currentPresenceSessionId: string | null;
-  runtimeStateSource: RuntimeStateSource;
-  surface: string;
-  cwd: string | undefined;
-  providerName: string;
-  model: string;
-}
-
-/**
- * Write top-level session presence once per provider and return the updated
- * provider `_presenceSessionId` slot.
- */
-export function registerPresenceLifecycle(args: PresenceLifecycleArgs): string | null {
-  // Phase 2 — Presence file lifecycle (top-level sessions only).
-  // Guard: only write once per provider instance (not once per turn).
-  // Top-level = depth is 0 or undefined (parentSessionId absent).
-  const isTopLevel =
-    (args.depth === undefined || args.depth === 0) &&
-    args.parentSessionId === undefined;
-  if (isTopLevel && args.sessionId !== undefined && args.currentPresenceSessionId === null) {
-    const sessionId = args.sessionId;
-    const workspace = args.runtimeStateSource.getWorkspace();
-    // Fire-and-forget — presence is best-effort.
-    void writePresenceFile({
-      sessionId,
-      surface: args.surface,
-      // Presence is written only under the top-level gate above, so depth is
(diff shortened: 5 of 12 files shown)


griffinwork40:
Verification complete. Synthesizing inline (all findings re-derived against the reviewed ref by me plus three independent verifier agents).

---

# Review — PR #738 `fix(presence): advertise presence for fresh sessions so Telegram AFK can discover them`

**Target:** `griffinwork40/agent-afk` PR #738 · `fix/presence-session-id` → `main` · reviewed ref **`920d214`** · merge base `84d8348`
**Triage:** bugfix + cross-provider dedup · 6 files · +523/−93 · regime **full**
**Stated intent:** PR #738 title + body — spec-compliance **assessed**

## Blocking

```
[R1] high · high · security/correctness · src/agent/providers/shared/presence-lifecycle.ts:148-149 · ref:920d214 · citation-type:file-state
finding: The signal handlers moved into the shared helper call process.exit() synchronously and are now registered by every fresh top-level session, so on a real SIGINT/SIGTERM they terminate the process while the surface's own async graceful shutdown — registered earlier, therefore suspended at its first await — is still in flight.
evidence:
    process.once('SIGINT', () => { removePresenceFileSync(sessionId); process.exit(130); });
    process.once('SIGTERM', () => { removePresenceFileSync(sessionId); process.exit(143); });
suggestion: Delete both process.exit() calls, keep removePresenceFileSync, and let the owning surface decide when to exit.
```
Reachability chain, each link read at `920d214`: the handler installs inside `provider.query()` on turn 1, i.e. **after** every startup handler — `telegram/bot.ts:432-433` (`await this.stop(); process.exit(0)`), `cli/commands/daemon.ts:652-653` (`await handle.stop(); process.exit(0)`), `cli/commands/interactive.ts:619,648`. Node runs listeners in registration order; all three earlier handlers are async and return at their first `await`, so the presence handler's exit lands mid-shutdown. Newly reachable: the pre-change gate was `config.sessionId !== undefined` (`snap-base/…/openai-compatible/index.ts:283`; anthropic passed `config.sessionId` at `snap-base/…/index.ts:758`), and neither surface supplies one — `src/telegram/construct-session.ts` sets no `sessionId`/`depth`/`parentSessionId` (grep: 0 hits) and `daemon.ts` passes only `depth: 0`. Consequences: `afk telegram stop` (`telegram/manager.ts:206` sends SIGTERM) abandons the Telegraf drain and watch teardown; the daemon's advertised "Press Ctrl+C to stop" abandons `handle.stop()`; the REPL under SIGTERM skips `runCleanupFunctions()` → transcript footer (`interactive.ts:542`), `/resume` sidecar (`:554-557`), `session.close()` (`:500`); exit code changes 0 → 130/143, which a launchd/systemd `KeepAlive` supervisor reads as abnormal. Telegram holds one session per chat (`telegram/session-manager.ts:107`), so N handlers register and the first to fire kills the rest.
**Refuted sub-claim, recorded so it does not get re-litigated:** a REPL **Ctrl+C keystroke** never reaches process listeners — the compositor holds stdin in raw mode (`cli/terminal-compositor.lifecycle.ts:163`) and consumes `^C` itself (`cli/terminal-compositor.input-dispatch.ts:416-417`). R1 is not a Ctrl+C-in-REPL regression; it is a daemon / telegram / `kill`-signal regression.

```
[R2] high · high · correctness (invariant violation) · src/agent/providers/shared/presence-lifecycle.ts:123-127 · ref:920d214 · citation-type:file-state
finding: /resume inside a live REPL reuses the memoized provider instance, so the once-per-instance guard suppresses the presence write for the resumed session while the closed session's file (afk:true if the operator toggled) survives — the Telegram watcher keeps tailing a dead ledger and the live session is undiscoverable and unrescuable.
evidence:
  if (
    isTopLevelSession(args.depth, args.parentSessionId) &&
    args.sessionId !== undefined &&
    args.currentPresenceSessionId === null
suggestion: Rewrite presence when the resolved id differs from currentPresenceSessionId (removing the stale file) instead of gating on === null.
```
Chain, independently re-derived: provider instances are cached per family for the process lifetime (`cli/commands/interactive/provider-factory.ts:36-44`, whose own comment states the cache survives `/model` swaps); `_presenceSessionId` / `_mintedSessionId` have **no reset site** (`anthropic-direct/index.ts:256,266,779,781`; `openai-compatible/index.ts:150,158,284,312`); `performResumeSwap` builds a new `AgentSession` from the same memoized factory (`bootstrap.ts:699,833`) and closes the outgoing one (`resume-swap.ts:163`); the only presence-removal sites in the tree are the three process handlers (`shared/presence-lifecycle.ts:146,148,149`), so `close()` leaves the stale file; and `setPresenceAfk` swallows ENOENT (`awareness/presence.ts:283-294`), so `/afk on` cannot create the missing file. That is the silent mismatch the module header itself calls "strictly worse than the loud absence" (`shared/presence-lifecycle.ts:11-18`).

## Medium

```
[R3] medium · high · correctness (invariant violation) · src/agent/providers/router/provider-router.ts:364,418 · ref:920d214 · citation-type:file-state
finding: A cross-family /model swap on a fresh session mints a second id and writes a SECOND presence file, while the swallowed session.init pins the ledger to the pre-swap id — so that file advertises a ledger directory that is never created.
evidence:
          // Swallow the new inner's session.init; propagate a construction error.
          const failed = yield* this.driveUntilInit(/* swallow */ true);
suggestion: Pass the session's live id into buildInner's inner config so every inner resolves through the explicit-id branch of resolveTopLevelSessionId.
```
Blast radius bounded (verified, which is why this is medium not high): the orphan carries `afk:false` and `/afk on` only ever marks `ctx.session.current.sessionId` (`cli/afk-mode-toggle.ts:73,103`), so auto-subscribe's `afk === true` filter (`telegram/bot.ts:588`) never selects it; the `/watch` no-arg listing does show it (`telegram/watch.ts:120-127` — no afk or ledger filter), and selecting it fails **closed** via `ledgerExists` (`watch.ts:104-109`). Harm is a phantom duplicate entry, not a silent mis-tail.

```
[R4] medium · high · security · src/agent/awareness/presence.ts:168,243 · ref:920d214 · citation-type:file-state
finding: Presence files are written with no mode and their directory created with no mode, so under a default umask they are world-readable — and this diff turns that from resume-only into every session, exposing each session's cwd, workspace, model and pid to any local user.
evidence:
      await writeFile(filePath, JSON.stringify(record, null, 2), 'utf8');
suggestion: Pass { mode: 0o600 } to writeFile and { recursive: true, mode: 0o700 } to the mkdir at presence.ts:168.
```
No secret material is written (record fields at `presence.ts:55-121`: sessionId, surface, actor, cwd, startedAt, model, workspace, pid); `chmod` appears 0 times in the file. `presence.ts` is not modified by this diff — the exposure is widened by it.

```
[R5] medium · high · test-coverage · src/agent/providers/shared/presence-lifecycle.test.ts:1-260 · ref:920d214 · citation-type:file-state
finding: No test ties the presence record's id to the session ledger directory, and none exercises the /resume, /clear or /model-swap boundaries, so R2 and R3 pass green.
evidence:     // A fresh mint per turn would move the ledger directory out from under the
suggestion: Add a case that runs one turn, reads the presence record, and asserts existsSync(getSessionLedgerPath(record.sessionId)); add one that reuses a single provider instance across a second, resumed-config session.
```
Grep at ref: `events.jsonl|getSessionLedgerPath|ledger` in that file → comment-only hits at `:69,:115`, zero assertions. What the suite does cover correctly: fresh mint (`:42-54`), second-turn stability (`:56-72`), precedence (`:74-104`), fork by depth and by parentSessionId (`:106-130`), truth table (`:133-143`), per-provider e2e (`:201-258`).

```
[R6] medium · high · perf-observability · src/agent/providers/shared/presence-lifecycle.ts:121-154 · ref:920d214 · citation-type:file-state
finding: The module emits no log or trace for the resolved id, its source, or the suppressed rewrite, so R2 and R3 leave no field-diagnosable signal — the operator sees a watch that reports healthy and never updates.
evidence: export function registerPresenceLifecycle(args: PresenceLifecycleArgs): string | null {
  // Guard: only write once per provider instance (not once per turn).
suggestion: debugLog the resolved id plus its source (explicit | memo | mint), and warn when args.sessionId !== args.currentPresenceSessionId with both non-null.
```
Per-turn cost of the resolver is one nullish chain plus at most one `randomUUID()` per instance — negligible. `touchPresenceHeartbeat` having no production caller is confirmed (definition `presence.ts:261`, re-export `awareness/index.ts:53`, tests only) and stays inert because `bot.ts:580` calls `readLivePresenceFiles()` with no options — the PR body's disclosure is accurate.

## Low / nit

```
[R7] low · high · correctness (contract text) · src/agent/providers/shared/presence-lifecycle.ts:95-97,102-107 · ref:920d214 · citation-type:file-state
finding: The Contract comment states a fork returns id: undefined, but subagent.ts:812 sets resume: options.parent.sessionId on every child config, so the explicit branch returns the PARENT's id before the fork check runs — and the new fork tests omit resume, encoding a shape production never produces.
evidence:  * Contract: never mints for a fork — returns `id: undefined` there, so fork
  * query construction keeps its existing per-call mint
suggestion: Reword to "never mints for a fork — an explicit parent id still wins" and add a fork-with-resume test case.
```
No behavioral regression: base resolved forks the same way (`snap-base/…/anthropic-direct/index.ts:924` `config.sessionId ?? config.resume`), and presence stays blocked for forks because `registerPresenceLifecycle` re-gates on `isTopLevelSession` (`:124`).

```
[R8] nit · low · api-compat · src/agent/providers/openai-compatible/query.ts:1272-1274 · ref:920d214 · citation-type:diff-context
finding: Fresh openai-compatible sessions carry a randomUUID instead of the `openai-pending-` sentinel, narrowing that sentinel's reachability. [UNVERIFIED: no production importers]
evidence:
    options.sessionIdOverride ??
    config.resume ??
    `openai-pending-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
suggestion: Note the narrowed sentinel in the changelog. [low confidence — verify with runtime context]
```

**api-compat otherwise clean** — read at `920d214`: the shim is a strict superset of the pre-change surface (`anthropic-direct/query/presence-lifecycle.ts:14-20` vs `snap-base` `:19-35`), `PresenceLifecycleArgs` is field-identical, the shim's only importer (`anthropic-direct/index.ts:85`) resolves; `buildQueryFromConfig` has exactly one production call site (`openai-compatible/index.ts:367`) and `sessionIdOverride` is additive; and the PR's claim about `opts.sessionId` is confirmed — `resumedSessionId` occurs exactly twice (`anthropic-direct/index.ts:957,989`) and feeds only `initSessionId` (`anthropic-direct/query.ts:285`).

## Dropped / corrected by verification

- **DROPPED** — "`/clear` appends two conversations to one `events.jsonl`": refuted. `reset()` calls `ledger.seal('reset')` (`agent-session.ts:811`), whose latch reset is documented for exactly this reuse (`ledger-lifecycle.ts:8-9,104-105`), writing a delimiting `reset` record into the append-mode file. Presence and ledger stay identical across `/clear` — the invariant holds.
- **DROPPED** — "no fork path sets `sessionId`/`resume`": refuted at `subagent.ts:812,621`. Folded into R7.
- **CORRECTED** — R1's original "first Ctrl+C kills the REPL" scope: refuted (raw mode), severity retained via daemon/telegram/SIGTERM reachability.
- **CORRECTED** — two wrong-line citations from Wave 1 (`presence-lifecycle.ts:684` → true `:109`; `:614-624` → true `:39-49`); both snippets exist, neither finding depended on the bad line numbers. Zero fabricated citations across 37 checked; 7 of 8 absence claims confirmed.

## Spec-compliance (against PR #738 title + body)

- **Unmet intent.** The body's load-bearing clause — "Both consumers now read one resolution, so they agree by construction rather than by coincidence" — holds only *within one provider instance for one session*. R2 (`/resume`) and R3 (`/model` swap) are the two production boundaries where presence id and ledger id diverge, and the memoized-per-family provider cache is what makes both reachable.
- **Surface widening not called for by the intent.** The intent is CLI discoverability, but the new gate also writes presence for telegram- and daemon-hosted sessions, whose records every consumer filters out (`bot.ts:588` requires `surface === 'cli'`; `watch.ts:121` excludes `telegram`). Those writes buy nothing and are the mechanism behind R1.
- **No substantive scope creep** in code structure: `sessionIdOverride` (`openai-compatible/query.ts:1256-1300`) and the re-export shim both match what the body describes.
- The body's own caveats check out: no e2e proof (confirmed — `bot.ts:588` `afk === true` and `afk-mode-toggle.ts:71-74`'s "no id before turn 1" gate both remain), and the `touchPresenceHeartbeat` gap is real and inert.

## What was not checked

- Wave 1 and all verifiers read a byte-exact `git archive` snapshot of **`920d214`** (`…/skill-preflight/a8e7fdb1…/snap-head`) plus a merge-base snapshot at `84d8348`; every citation above was verified at `920d214`. Zero fabricated citations.
- **Nothing was executed** — no `pnpm test`, no `pnpm lint`, no live REPL/daemon/bot run. The PR's claims of 15/15 new tests, 13,395 passing, and clean `tsc` are unverified here. Runtime listener *ordering* for R1 is derived from registration sites, not observed.
- Not reviewed: the browser/web tool surfaces, `witness`/trace internals beyond the `'exit'` sealing listener, `watchManager.start`'s path resolution line-by-line, Windows signal semantics, the operator's actual umask, and whether any out-of-tree plugin reads `~/.afk/state/presence/**`.
- Synthesis was performed inline by the coordinator rather than by a Wave 2 sub-agent, because verification had already resolved every cross-agent conflict; the two shadow verifiers' disagreements with Wave 1 are recorded above rather than auto-resolved.

# DO NOT MERGE

Two `high` findings, both fixable in a few lines: **R1** (delete the two `process.exit()` calls from the signal handlers) and **R2** (rewrite presence when the resolved id changes, instead of gating on `=== null` — the same change also removes R3's orphan). The core diagnosis and the shared-helper dedup are sound; the regression is in what the widened gate now registers, and the declared invariant does not yet hold across `/resume` and `/model`.

**Done**
- Reviewed PR #738 read-only: resolved diff + reviewed ref, captured stated intent, ran a 2-agent full-regime Wave 1, a mechanical citation/absence verifier (37 citations, 9 absence claims), and two adversarial shadow verifiers on all high-severity claims; then re-derived the two load-bearing chains myself.
- Evidence on disk: `~/.afk/state/skill-preflight/a8e7fdb1-9f9b-421b-b8cf-6f3d51e14faf/` — `pr-738.diff`, `intent.txt`, `snap-head/` (tree at `920d214`), `snap-base/` (tree at `84d8348`); trace at `~/.afk/state/witness/<session>/trace.jsonl`.
- World changed: nothing in the repo, the branch, or the PR — no files edited, no commits, no comments posted; working tree still `main @ ba3cd08` clean. The three snapshot/intent artifacts (~23 MB) are new under the preflight run dir and are safe to delete.
- Pending: the verdict is advisory only — no reviewer comment was posted (this skill is read-only); R1/R2 fixes and any test run are the author's next step.

---

griffinwork40 (commented):

## Merge decision: ✅ MERGE

Reviewed against branch HEAD `1f52fc7` (post-merge file state extracted and read directly; all citations below verified at that ref). Regime: **full** (11 files, 1156+/111−). Dimensions: security, api-compat, correctness, spec-compliance, test-coverage, perf-observability.

No `critical` or `high` finding survived verification. The change does what its description says, and the load-bearing invariant it claims to protect holds *by construction* rather than by coincidence.

---

### Verified clean (the claims that mattered most)

**The load-bearing invariant holds.** The PR argues presence id must equal the ledger-directory id or the Telegram watcher silently tails a nonexistent ledger. Both consumers provably read one resolution:
- `anthropic-direct/index.ts:772` resolves once → `:782` feeds `registerPresenceLifecycle`, `:958` feeds `resumedSessionId` → `:990` query construction.
- `openai-compatible/index.ts:277` resolves once → `:313` presence, `:295` `buildOpts.sessionIdOverride` → `query.ts:1271-1274` `synthesizedSessionId`.
When `.id` is `undefined`, presence is not written at all, so a written-but-mismatched record cannot occur.

**Dedup is real — no stale inline gate remains.** Grep for `writePresenceFile|actorFromDepth|removePresenceFileSync` across both provider `index.ts` files returns zero hits; both now import the shared helper. This mattered: fixing only `anthropic-direct` would have left every OpenAI-compatible-backed session broken.

**api-compat clean.** The shim (`anthropic-direct/query/presence-lifecycle.ts`, 20 lines) re-exports both pre-merge symbols (`PresenceLifecycleArgs`, `registerPresenceLifecycle`) with unchanged shape, plus three additive ones. `sessionIdOverride` is a pure prepend to the existing `??` chain (`query.ts:1271-1274`), so omitting it reproduces prior behavior exactly. Resume semantics: `const explicit = args.sessionId ?? args.resume` (`shared/presence-lifecycle.ts:178`) reproduces the removed `config.sessionId ?? config.resume` for every supplied-id case — bit-for-bit as claimed.

**Id stability across turns.** `const minted = args.memoized ?? randomUUID()` (`shared/presence-lifecycle.ts:192`) plus the guard `if (args.currentPresenceSessionId === sessionId) return sessionId;` (`:220`) means turn 2 reuses turn 1's mint and writes no second file. The memo is a `private` **instance** field on each provider, not module-level — no id leakage between concurrent sessions in one process.

**Fork path safe.** `isTopLevelSession` short-circuits resolution to `{ id: undefined }`, and `registerPresenceLifecycle:216-218` re-gates independently, so a fork never advertises even though its `resume` carries the parent's id for query construction. No collision on a parent ledger dir.

**Tests are not tautological.** `presence-advertise.test.ts` exercises the real `writePresenceFileSync`/`readPresenceFiles` under a temp `AFK_HOME` with no SDK mock, and the fresh-session case genuinely omits `sessionId` — which is exactly the shape the old gate missed. This is the reproducer doing real work, and it justifies the claimed 4/15 pre-fix failures.

---

### Findings

**1 · low · security · `src/agent/awareness/presence.ts:176-181` · ref:1f52fc7 · citation-type:file-state**
`ensurePresenceDir()` passes `mode: 0o700` only on first creation; Node does not chmod on the recursive no-op path, so a presence directory created by an earlier version keeps its prior looser mode after upgrade. Files inside are written `0o600`, so record *contents* stay protected — the residual exposure is the directory listing (session ids).
```js
await mkdir(getPresenceDir(), { recursive: true, mode: 0o700 });
```
Fix: `chmod(getPresenceDir(), 0o700)` once at startup regardless of whether `mkdir` created it. (Grep confirms no `chmod` call anywhere in `presence.ts`.)

**2 · low · correctness · `src/agent/providers/shared/presence-lifecycle.ts:31-43` · ref:1f52fc7 · citation-type:file-state**
A cross-family `/model` swap builds a fresh inner provider whose `_mintedSessionId` memo starts `null`; because `innerConfig = { ...this.baseConfig }` (`router/provider-router.ts:222`) carries no `sessionId`/`resume` for a fresh session, resolution reaches the mint branch and advertises a second record while the ledger stays latched to the pre-swap id (`session/ledger-lifecycle.ts:54-55`).
**Independently verified as fail-closed, both readers:** the orphan record is written without an `afk` field (`presence-lifecycle.ts:243-255`) and auto-subscribe requires `p.afk === true` (`telegram/bot.ts:588`); `/watch` selection re-checks `ledgerExists` (`telegram/watch.ts:106,134`). The module header discloses this gap accurately, including the correct fail-closed reasoning and the reason closing it is deferred. Reported for the follow-up trail, not as a merge blocker.

**3 · low · spec-compliance (disclosure) · `src/agent/awareness/presence.ts` + `shared/presence-signals.ts`**
Two substantive changes are absent from the description's **Fix** section: the `0o700`/`0o600` permission hardening, and the new process-level `presence-signals.ts` listener registry. Both are defensible as direct consequences of the fix — every fresh session now writes these files, widening exposure, and per-session cleanup listeners multiply once more sessions advertise — so this reads as an under-described change rather than scope creep. Naming both in the PR body would make the diff self-explaining.

**4 · low · test-coverage · `shared/presence-advertise.test.ts:132-142` · ref:1f52fc7 · citation-type:file-state**
No test drives two simultaneous top-level provider instances, though the module header explicitly contemplates one process hosting several concurrent top-level sessions. Absence confirmed by grep (`makeProvider\(\).*makeProvider\(\)`, `Promise.all.*triggerPresence` → zero matches). Adding one case asserting two concurrent instances produce two distinct presence files would pin the per-instance isolation the memo design depends on.

**5 · nit · perf-observability · `src/agent/awareness/presence.ts:294-304` · ref:1f52fc7 · citation-type:file-state**
`writePresenceFileSync` does `mkdirSync` + `writeFileSync` on an async provider path. Bounded and reasoned in-file (~300 bytes, once per advertised id, never per turn), and the ordering rationale — durability before an immediate `/afk on` can race it — is sound. Worth a note only for a slow or network-mounted `AFK_HOME`, which the description does not discuss.

**6 · nit · convention · both provider `index.ts` (1072 and 663 lines at the ref)**
Both remain far over the repo's 350-LOC guideline. The description discloses this and defers it; flagged only so the follow-up is not lost.

---

### What was not checked
- Read against branch HEAD `1f52fc7` — all citations verified at that ref. One agent reported diff-relative offsets (e.g. `:1094`) for a 261-line file; those were re-resolved to true file lines (`:192`, `:220`) and the quoted content confirmed present. No finding rests on an unverified citation.
- **Stated intent: PR #738 title + body — spec-compliance assessed.**
- Tests and lint were **not executed**. The claimed counts (4/15 pre-fix, 15/15 post-fix, 13,395 passing, clean `tsc --noEmit`) are taken as asserted, not re-derived.
- No live Telegram round-trip — the description already flags this as needing one manual run (fresh session → `/afk on` → a `confirm` question). That remains the one gap these tests structurally cannot close.
- Did not exhaustively trace every daemon/Telegram provider construction site to confirm `surface` is always threaded correctly rather than defaulting to `'cli'`; `provider-factory.ts` and `bootstrap.ts` were spot-checked.
- `touchPresenceHeartbeat` having no production caller was verified by grep and matches the description's own disclosure; the heartbeat-age filter is inert today because `bot.ts` calls `readLivePresenceFiles()` with no options.

*Read-only review — no files, branches, or PR metadata were modified.*

griffinwork40 (commented):

## Merge decision: **MERGE** — no blocking findings

Reviewed against branch HEAD `1f52fc7`. Every `file:line` below was read at that ref. The core invariant the PR sets out to protect **holds in both providers**, and one high-severity candidate was downgraded after adversarial re-derivation (details in *Verification notes*). Nothing here blocks the merge; the medium items are worth a follow-up.

### Invariant check — passes

Presence id == ledger-directory id, verified by reading both consumers rather than trusting the description:

- **anthropic-direct**: one resolution at `index.ts:772-779` feeds both `registerPresenceLifecycle` (`:785`) and `resumedSessionId` (`:958` → `:990` → `query.ts:285` `initSessionId` → `session.init` → `paths.ts:612-624`).
- **openai-compatible**: `index.ts:277-284` feeds `:295` `sessionIdOverride` and `:316` presence → `query.ts:1271-1274`.
- Memoization is stable across turns (same-id early return, `shared/presence-lifecycle.ts:220`); forks are blocked by the independent re-gate at `:216`.

A residual risk was raised — that `/clear` (`AgentSession.reset()`) discards the memoized mint and strands the `/afk on` binding — and it was **refuted**: the mint is an instance field (`anthropic-direct/index.ts:266`), the REPL resolves providers through `createMemoizedProviderFactory` (`cli/commands/interactive/bootstrap.ts:574`; cache at `provider-factory.ts:36-40`), and `reset()` (`session/agent-session.ts:842-844`) preserves the factory reference, so the same provider object — and the same minted id — survives. Narrow caveat only: a second mint remains possible when the memo key changes (model-family switch across a reset) or on surfaces where `providerFactory` is unset.

### medium

**1 · security · `src/agent/awareness/presence.ts:158-160` · ref:1f52fc7 · file-state** — An explicitly-supplied session id reaches `join()` unvalidated, so a `--resume '../../../../.afk/settings'` value writes the presence JSON outside the presence directory (the appended `.json` bounds the target to `*.json` paths). Chain: `cli/resume-session.ts:29` returns the raw unresolved CLI string on a `findSession` miss → `shared/presence-lifecycle.ts:178-179` passes it through verbatim → `writePresenceFileSync`.
```
function presenceFilePath(sessionId: string): string {
  return join(getPresenceDir(), `${sessionId}.json`);
```
Absence grounded: `git grep -n "isSafeLedgerSessionId\|validateSessionId" 1f52fc7 -- src/agent/awareness src/agent/providers/shared` → 0 hits. The guard exists (`paths.ts:597`) and the sibling ledger path already throws on it (`paths.ts:613-614`) — presence is the asymmetric one. Pre-existing for `config.sessionId`; this PR adds `config.resume` as a second unvalidated source.
*Fix:* call `isSafeLedgerSessionId` at the top of `presenceFilePath` and no-op when false, matching `getSessionLedgerDir`.

**2 · observability · `src/agent/awareness/presence.ts:301-303` · ref:1f52fc7 · file-state** — The now load-bearing initial write swallows every error with no diagnostic, so a failed presence write leaves bidirectional AFK broken with zero signal — the same silent-failure class this PR exists to remove.
```
  } catch {
    // Best-effort — swallow silently.
  }
```
The diff adds **zero** log calls anywhere (the `debugLog` at `presence-lifecycle.ts:230` is a pre-existing context line).
*Fix:* `debugLog` the caught error in `writePresenceFileSync` and in `ensurePresenceDir`'s catch.

**3 · correctness · `src/agent/providers/shared/presence-signals.ts:76-88` · ref:1f52fc7 · file-state** — Signal ownership is sampled once at lazy install, so a surface attaching its SIGINT handler *after* the first presence write gets pre-empted by `process.exit(130)`.
```
  const sigintOwnedElsewhere = process.listenerCount('SIGINT') > 0;
  onSigint = () => {
    removeAllTracked();
    if (!sigintOwnedElsewhere) process.exit(130);
```
The docblock (`:62-64`) enumerates only startup-time installers, but two production installers are lazy and post-first-turn (`browser/registry.ts:70-71`, `cli/input/non-tty.ts:33`), which would truncate `closeBrowserProvider()`'s async shutdown. Currently held closed by the surface gate (`PRESENCE_MINT_SURFACES` = `['cli']`, `presence-lifecycle.ts:132`) — not by the ordering claim the comment makes. *[reachability not independently re-verified in the verification pass]*
*Fix:* read `process.listenerCount(sig)` inside the handler at signal time instead of latching at install; correct the header to name the surface gate as the real guarantee.

**4 · spec-compliance · multiple · ref:1f52fc7 · file-state** — Scope creep relative to the stated Fix (PR body lines 17-21): four substantive mechanisms land undisclosed — a synchronous writer (`presence.ts:294-304`), a process-level signal registry replacing per-session handlers (`presence-signals.ts`), id-keyed rewrite-with-removal (`presence-lifecycle.ts:222-233`), and `0o600`/`0o700` hardening across six write sites. The items the body *does* declare out of scope (`touchPresenceHeartbeat` having no caller — confirmed, `presence.ts:317`; the 350-LOC provider files; no e2e Telegram test) are correctly excluded and not counted against the spec.
*Fix:* name those four in the PR body so the security and shutdown-semantics changes get scoped deliberately.

### low

**5 · correctness · `src/agent/providers/shared/presence-lifecycle.ts:230-232`** — The id-rewrite path uses the queue-bypassing sync remover, so an in-flight queued mutator can resurrect the file after the unlink, leaving an orphan record carrying `afk: true` — which `telegram/bot.ts:588` (`p.surface === 'cli' && p.afk === true && p.sessionId`) auto-subscribes to, and because a freshly advertised record omits `afk` entirely the orphan is the *only* match, so it takes the chat's single watch slot (`bot.ts:614-616`) and starves the live session. `presence.ts:420-423` documents this exact hazard as the reason the **async** `removePresenceFile` is enrolled in `presenceWriteTails`; the sync variant (`:441-445`) is not, and the async one has zero production callers. `setPresenceAfk` (`:342-345`) fences only the unlink-before-read interleaving — read-resolves → unlink → write-lands still resurrects.
Filed **low, not high**: the only racer is `void setPresenceAfk(...)` at `src/cli/afk-mode-toggle.ts:103,140`, so a collision needs two human slash commands inside one ~300-byte read→write window. Elevate to medium if a per-turn `touchPresenceHeartbeat` caller is ever added — that makes the racer fire every turn.
*Fix:* `await removePresenceFile(previous)` before the new sync write, or enroll the sync remover on the same tail.

**6 · security · `presence.ts:296`** — `mkdirSync(..., { mode: 0o700 })` does not chmod an *existing* directory, so already-installed users keep a umask-default presence dir while the newly-broadened write set puts `cwd`/`pid`/`remoteUrl` in it. The 0o600 file mode is correct and consistent. *Fix:* one error-swallowed `chmodSync(getPresenceDir(), 0o700)` after the mkdir.

**7 · security · `awareness/workspace-source.ts:114,121`** — `remoteUrl` comes from raw `git remote get-url origin` and is `JSON.stringify`d verbatim (`presence.ts:262-268`), so an origin with an embedded credential is persisted for every fresh CLI session now advertised. Pre-existing; footprint widens. No exfil path today (`.remoteUrl` has no consumers in `src/telegram`, `src/cli`). *Fix:* strip URL userinfo before assignment.

**8 · test-coverage** — Three gaps, each grep-confirmed at the ref: no test asserts the central presence-id-equals-ledger-dir invariant (`git grep -n "ledger" -- '*presence*.test.ts'` → comments only); `sessionIdOverride` has zero test hits, leaving its precedence over `config.resume` unexercised; and the new `0o600`/`0o700` modes are unpinned in any presence test. The resurrection race is pinned for the async remover (`presence.test.ts:593`) but not the sync one.

**9 · correctness · `presence-lifecycle.ts:104-131`** — The `PRESENCE_MINT_SURFACES` docblock justifies itself with daemon record accumulation, but daemon tasks carry an explicit id (`daemon/scheduler.ts:541,558`) and return before the surface check, so the gate is inert for the case it cites. Behavior is unchanged; the rationale is wrong.

**10 · spec-compliance · `presence-lifecycle.ts:178-179`** — PR body line 21 says forks "resolve to `undefined`", but `subagent.ts` sets `resume: parent.sessionId`, so the explicit branch returns the parent id; the code's own docblock (`:168-173`) and test assert the parent-id behavior. Presence is still correctly blocked — by the re-gate at `:216`, not by an `undefined` id. Prose fix only.

### nit

**11** — `openai-compatible/index.ts:62` imports `shared/presence-lifecycle.js` directly while `anthropic-direct/index.ts:85` routes through the new re-export shim; one import style will drift. Point both at the shared module and keep the shim for out-of-tree consumers.
**12** — PR body lines 39-40 cite "4 of 15 fail / 15/15 pass", but the checked-in files carry 7 e2e cases × 2 provider branches plus 13 unit cases. Restate against the final files.
**13** — `presence.ts:183-202` enumerates exactly one deliberate queue exception (the sync remover); `writePresenceFileSync` is a second. It self-declares non-enrollment in its own docstring (`:288-290`), so this is a doc-consistency point, not a hidden bypass.

### Verification notes

- **Citations:** all `blocking`/`high` candidates were re-read at `1f52fc7` by an independent verifier. One mis-citation was caught and corrected: finding 5's racer is `src/cli/afk-mode-toggle.ts`, **not** a Telegram module — `setPresenceAfk` has zero occurrences under `src/telegram/`.
- **Absence claims** were grep-verified at the ref, not asserted. Two proposed test-gap findings were **dropped** as refuted by `presence-lifecycle.test.ts:104-118` and `presence-advertise.test.ts:132-138`.
- **Downgrades:** the sole `high` candidate (finding 5) was adversarially re-derived; links 1-3 of its chain held but link 4 (timing) did not support routine reachability → **high → low**. One `medium` was downgraded to a nit (finding 13).
- **What was not checked:** no tests, build, or `tsc` were run — the PR's "13,395 passed / lint clean" claims are unverified here. No live Telegram round-trip (the PR body flags this itself). `AgentConfig` construction sites were not exhaustively enumerated, so "this PR *newly* routes `config.resume` into a presence path" is unverified in the general case. openai-compatible `close()`/teardown parity and `readLivePresenceFiles`'s heartbeat-filter internals were not reviewed.
- **Stated intent:** PR #738 title + body — spec-compliance assessed against it.

PR #739, opened 2026-07-28
feat(agents): warn when an agent file shadows a read-only builtin (+ ADR 0002)

griffinwork40:
## Summary

Two commits: one behaviour change, one record.

**1. `feat(agents)`: warn when an agent file shadows a read-only builtin**

Builtin agents are seeded at the **bottom** of the registry precedence order, so any plugin-, file-, or config-scope agent of the same name replaces one with no signal at all. That is fine for a plain default, but the tool-restricted builtins (`research-agent`, `git-investigator`, `Explore`) are a safety surface: ten bundled orchestration `SKILL.md`s dispatch them by bare name, so a single file at `~/.afk/agents/research-agent.md` declaring broader `tools:` silently converts every "mechanically locked read-only verifier" in those skills into a write-capable agent — for every project on the machine.

`warnIfShadowsBuiltin` is wired into the three displacement points that previously overwrote a builtin without comment:

| point | `registry.ts` |
|---|---|
| file merge (`scanScope`) | before `registry.set(parsed.name, …)` |
| plugin merge | `loadAgentRegistry`, plugin loop |
| config-tier merge | `loadAgentRegistry`, `configAgents` loop |

The message names the specific restriction being discarded — the tool allowlist, plus the read-only bash gate when the displaced builtin carried `bashReadOnly`.

**Precedence is deliberately unchanged.** Overriding a builtin stays intended operator power; it just stops being invisible. This extends a convention the registry already follows for same-scope duplicates and cross-directory project-tier overrides (*"warn so the shadow is not silent"*, `registry.ts:143-145`) to the one case it did not cover.

**2. `docs(decisions)`: ADR 0002 — read-only agent-type consolidation**

Records the investigation into whether the vendored `research-agent` / `git-investigator` types are redundant and should be deleted, and why the answer is no (~10 hard dispatch dependencies; an unresolved `agent_type` is a hard error, not a fallback to generic).

### Not in this PR

The ADR's headline proposal — collapsing the two types into one carrying `bashReadOnly` — is **Proposed, not implemented**, on purpose. It changes runtime behaviour for ten consumer skills on n=3 evidence where no arm ran to completion, and it would make the best-effort read-only-bash classifier the *only* remaining boundary while its hit-rate under research-shaped prompts is unmeasured. That measurement should come first.

### Two decision items voided by measurement error

Worth flagging for anyone reading the ADR, since both were originally recorded with confidence:

- **F6** retracts *"fix trace attribution"* — `subagentId` lives at `.payload.subagentId`, so the query behind "0 of 2,281 rows" was wrong, not the code.
- **F8** retracts *"add a wall-clock bound"* — `SUBAGENT_DEFAULT_TIMEOUT_MS` (45 min, `subagent.ts:108`, applied `:1023`, pinned by `subagent.test.ts:897`) already exists, alongside an 8-min idle and 60-min background bound. The 45-minute dispatch death cited as proof the bound was *missing* was the bound firing at its configured value.

Provenance is labelled per finding in the ADR because confidence differs sharply between them; uncited prose there should be read as opinion.

## How was this tested?

- `pnpm test` — **706 files, 13,384 passed, 14 skipped, 0 failed**
- Full CI gate set run locally, all pass: `audit:env:check`, `scan:env:check`, `audit:sdk:check`, `audit:chalk:check`, `audit:deps`, `build`
- 4 new tests in `registry.test.ts` covering: a user file displacing a tool-restricted builtin (asserting the file still wins — precedence unchanged), the `bashReadOnly` wording path, the config tier, and a negative case asserting silence when a name collides with no builtin
- `registry.ts` is 255 LOC, under the repo's file-size ceiling

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes
- [x] No secrets, tokens, or private paths in the diff — the ADR's two references to the private upstream plugin (a name and a `~/.afk/plugins/cache/...` path) were generalized before commit, matching `vendored.test.ts`'s convention of pinning hashes rather than naming upstream


griffinwork40:
# Code Review — PR #739

**`feat(agents): warn when an agent file shadows a read-only builtin (+ ADR 0002)`**
`afk/remove-redundant-agents` → `main` · reviewed ref **`a3321df`** · +430/−2 across 3 files · regime **full**
CI: **all green** (10 checks; `windows-latest` + Vercel Agent Review skipped) · `gh` reports MERGEABLE · branch forked from `84d8348` (v5.81.6), main now `ba3cd08` (v5.81.8)

---

## Blocking

**None.** No `critical` or `high` finding survived verification.

## Findings

### `medium` · confidence high · spec-compliance / observability
`src/cli/commands/interactive.ts:676` (with `src/cli/commands/interactive/bootstrap.ts:323-326`, `src/agent/agents/registry.ts:221`) · `ref:a3321df` · citation-type: file-state

**finding:** On the interactive REPL the new warning is written to stderr during bootstrap and then destroyed — scrollback included — by an unconditional screen clear later in the same startup callback, so the PR's headline benefit ("it just stops being invisible") does not land on the dominant human surface.

**evidence:**
```
registry.ts:221   const warn = options.warn ?? ((message: string) => process.stderr.write(message + '\n'));
interactive.ts:676  process.stdout.write('\x1b[3J\x1b[2J\x1b[H');
interactive.ts:712  // screen clear above erases anything written to stderr/stdout before this
```
Control flow verified line-by-line: `.action()` opens `:235`, `bootstrapSession` awaited `:406`, clear at `:676` — same indent level, no TTY check (zero `isTTY` hits in the file), no flag/env gate, no early return; the only intervening construct is a `try/catch` whose handler hard-exits. `bootstrap.ts:323-326` passes `{cwd, pluginAgents}` only — no `warn`. No recovery path: `/agents` reads `session.supportedAgents()` (`plugin-agents.ts:88`), which both providers return `[]` for (`anthropic-direct/query.ts:724-726`, `openai-compatible/query.ts:1165-1167`).

**suggestion:** Route the registry warnings through the post-clear notice stash that already exists for exactly this erasure — `interactive.ts:714-716`, and note `interactive/loop-iteration.ts:122-141` already defers a `pendingShadowingNotices` channel past the clear. The repo also has a `Test (real PTY scrollback)` CI job, so the fix is mechanically verifiable.

> ⚠ **Severity adjusted.** Both Wave 1 agents filed this `high` with the claim "never reaches a human on **any** surface." Shadow-verify returned **CONFIRMED-NARROWED**: the mechanism is real but the quantifier is false — the warning prints live on `chat` (`chat.ts:506`, no clear sequence in that file) and persists to operator-readable logs under supervision (`launchd/plist.ts:76-90`, `systemd/unit.ts:102-114`). Downgraded to `medium`, scope narrowed to the REPL.

### `low` · confidence high · correctness / spec-compliance
`src/agent/agents/registry.ts:230` (vs. `:15-17`) · `ref:a3321df` · file-state

**finding:** The plugin-loop call site cannot fire from any production path because plugin agent names are unconditionally namespaced, so the body's "three displacement points that previously overwrote a builtin" holds for two.
**evidence:**
```
skill-bridge.ts:330   if (pluginName === null) continue;
skill-bridge.ts:340   const qualifiedName = `${pluginName}:${parsed.name}`;
registry.ts:15-17     * … they are namespaced `<plugin>:<agent>` they never collide with a bare builtin name
```
Builtin names are colon-free (`builtins.ts:102,128,149,167`); all four production callers pass `pluginAgents: discoverPluginAgents()` only.
**suggestion:** Reword to "two production displacement points plus a defensive guard on the public `pluginAgents` option," or drop the `:230` call.

### `low` · confidence high · docs-accuracy
`src/agent/agents/registry.ts:28-29` · `ref:a3321df` · diff-context

**finding:** The new module-doc sentence bakes in "ten bundled skills dispatch them by bare name," but the union is nine files and `git-investigator` is dispatched by zero bundled `SKILL.md`.
**evidence:** `git grep -l "research-agent" a3321df -- 'src/bundled-plugins/**/SKILL.md'` → 8 files; `Explore` → 2 (`gather`, `research`; overlaps) → union **9**. `git grep -n "git-investigator" a3321df -- 'src/bundled-plugins/**/SKILL.md'` → **0**; it is granted for nested dispatch at `builtins.ts:119` (`'Agent(git-investigator)'`).
**suggestion:** "nine bundled skills dispatch `research-agent`/`Explore` by bare name; `git-investigator` is reached only by nested dispatch." The safety argument is unaffected.

### `low` · confidence high · test-coverage
`src/agent/agents/registry.test.ts:292-305` · `ref:a3321df` · file-state

**finding:** The only test that drives a bare-named plugin agent over a builtin silences the sink (`warn: () => {}`, `:295`), so the `:230` message is asserted nowhere.
**evidence:** `git grep -n "warnIfShadowsBuiltin\|overrides built-in agent" a3321df -- 'src/**/*.test.ts'` → `registry.test.ts:157,182,190` only.
**suggestion:** Swap that test's sink for `vi.fn()` and assert `'plugin agent (plugin:x) overrides built-in agent'`.

### `low` · confidence high · test-coverage
`src/agent/agents/registry.ts:133` · `ref:a3321df` · diff-context

**finding:** The empty-`restriction` branch is production-reachable through `general-purpose` — the only builtin with neither `tools` nor `bashReadOnly` — and no test pins its bare wording.
**evidence:** `builtins.ts:157` `// No 'tools' — inherit-all (Claude Code parity for general-purpose).`; `git grep -n "general-purpose" a3321df -- src/agent/agents/registry.test.ts` → `:43,:70,:105`, all reads of the builtin, none written as a user/project/config agent.
**suggestion:** Add a case writing `general-purpose` into the user dir, asserting the message terminates at the agent name.

### `nit` · confidence high · test-quality
`src/agent/agents/registry.test.ts:223` · `ref:a3321df` · file-state

**finding:** The pre-existing negative assertion matches the bare substring `'overrides'`, which the new message at `registry.ts:134` also contains, so it passes only because its fixture names miss every builtin.
**evidence:** `223  expect(warn).not.toHaveBeenCalledWith(expect.stringContaining('overrides'));`
**suggestion:** Narrow to the cross-dir wording, mirroring the new test's specific `'overrides built-in agent'` at `:190`.

### `nit` · confidence medium · security
`src/agent/agents/registry.ts:134` · `ref:a3321df` · diff-context

**finding:** `origin` (an absolute `filePath` from `readdirSync`) is interpolated raw into terminal output while `name` is JSON-escaped — **pre-existing pattern, not introduced**; four sibling sites (`:162,:172,:186,:205`) already do this and the diff adds a fifth.
**evidence:** `134  warn(\`[afk] agents: ${origin} overrides built-in agent ${JSON.stringify(name)}${restriction}\`);` — `sanitizeForDisplay` (`utils/terminal-sanitize.ts`, self-described security boundary) is unused in `src/agent/agents/`. `[not re-verified in Wave 1.5 — below the critical/high verification threshold]`
**suggestion:** Route `origin` through `sanitizeForDisplay` at all five sites in one follow-up, outside this PR.

## Clean reads

- **security / secrets** — `no issues found — read all 3 files at a3321df`. The checklist claim verifies **TRUE**: `git grep -n '/Users/'`, `'plugins/cache'`, token/key patterns across the ADR + both source files → **zero** hits. No boundary weakened: `warnIfShadowsBuiltin` returns `void`, reads only `registry.get(name)`, runs before every `registry.set`; effective tool grants untouched. Name-borne newline/ANSI injection is structurally blocked (`parser.ts:118,142-144` + `JSON.stringify`).
- **api-compat** — no exported signature changed (`loadAgentRegistry` `:219`, `LoadAgentRegistryOptions.warn` pre-existed at `:76`, `collectMarkdownFiles` `:85`); `warnIfShadowsBuiltin` is module-private; no warning-count assertion exists to break.
- **correctness (guard + placement)** — `prior?.source !== 'builtin'` is correct at all three sites: builtins seed first (`:223`) and each displacement rewrites `source`, so exactly one warning per genuine displacement, zero false positives. All three calls sit after every `continue` and immediately before the sole `registry.set` (`:192→194`, `:230→231`, `:249→250`). Chained displacement (user→builtin warns, project→user silent) is correct-by-design per the documented scope semantics at `:26`.
- **precedence unchanged** — asserted at `registry.test.ts:155`; `registry.set` call sites untouched.
- **ADR F8 retraction accuracy** — exact: `subagent.ts:108` (`45 * 60_000`), applied `:1023`, idle `:160`, background `:193`, pinned `subagent.test.ts:897`.
- **perf** — one `Map.get` per merged agent at a session-static bootstrap called once per surface; not hot-path.
- **`registry.ts` = 255 LOC** — verified exact.

## Dropped / corrected by verification

| item | action | why |
|---|---|---|
| "PR body's *file-size ceiling* cites a standard not in evidence" | **dropped** (false-absent) | A documented 350-LOC norm exists at `docs/browser-control-scope.md:660`. No *enforced* gate exists (no lint config; `lint` = `tsc --noEmit`), so the norm is convention-only — but the absence claim as stated is refuted. |
| "HEAD has deleted `warnIfShadowsBuiltin` entirely" | **dropped** (misread) | `git log -S warnIfShadowsBuiltin ba3cd08` → empty: it was never on main. Absence at `ba3cd08` is expected unmerged-PR state, not a revert. |
| 4th production call site cited as `src/cli/commands/telegram.ts:311` | **corrected** | Actual: `src/telegram.ts:309`. The substantive claim — all four callers omit `warn` — was independently verified twice. |
| `src/cli/commands/telegram.ts:107-109` "routes stderr to a log file" | **excluded** | Those lines print the log *path* (`Logs: …` / `Tail with: afk telegram logs --follow`). **CONFLICT (unresolved):** the shadow verifier reports stderr→`~/.afk/logs/telegram.log` with shipped readers at `telegram.ts:170-187`; the citation verifier found no redirect at the cited range. Not load-bearing — the supervised launchd/systemd redirects are verified either way. |

## What was not checked

- **Refs read:** all Wave 1 + verification agents read at branch HEAD **`a3321df`**; citations C1–C11 verified at that ref, zero `fabricated`, zero `diff-only`.
- **Stated intent:** PR #739 title + body — **spec-compliance assessed**. Verdict: intent met in mechanism (precedence genuinely unchanged, restriction wording correct, file/config call sites correct, extends the existing warn-so-the-shadow-is-not-silent convention), unmet in effect on the REPL, with one overclaim ("three displacement points") and one wrong count ("ten bundled skills"). **No scope creep** — nothing outside the warning path changed.
- **Nothing was executed.** No `pnpm test`, `pnpm lint`, or REPL session run by any agent, so the body's "13,384 passed" is unverified locally — though CI's `Test (macos-latest)`, `Test (ubuntu-latest)`, `Lint & Build`, and `Test (real PTY scrollback)` all pass on this head.
- **ADR prose not reviewed** beyond the F8 retraction spot-check and the secret/private-path scan — F1–F7 empirical claims (trace counts, the ~10× figure) were not audited against trace data.
- **Terminal-emulator variance** for `\x1b[3J` scrollback erasure was not runtime-tested; the finding rests on source order plus the repo's own comment at `interactive.ts:712`.
- `--dump-prompt` was not confirmed to serialize registry entries with `source`; non-`src/` skill locations (installed `~/.afk/plugins/`) were not searched.
- **Process deviation:** Wave 2 synthesis was performed inline rather than by a dispatched agent — both Wave 1 outputs and the full verification manifest were already in context, and re-passing them would have lost fidelity. Waves 1, 1.5, and shadow-verify all ran as dispatched sub-agents per the skill.

---

# ✅ MERGE

Additive, precedence-preserving, CI-green, no security or API-compat defect. The one substantive finding is a **medium** visibility gap: the warning is erased on the REPL, so the feature's headline benefit does not reach an operator there until the notices are re-emitted past `interactive.ts:676` — recommended as a follow-up, not a blocker, since the change strictly improves on the status quo (previously invisible on every surface; now live on `chat` and logged under supervision). The two doc-accuracy corrections ("three displacement points" → two; "ten bundled skills" → nine, `git-investigator` zero) are worth a body/comment edit before merge.

**Done**
- Reviewed PR #739 at head `a3321df` across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability: 2 parallel Wave 1 dimension agents, 1 citation+absence verifier (11 citations, 7 absence claims, 3 numeric claims), 1 adversarial shadow verifier on the sole high finding, synthesis inline.
- Evidence: report above with `file:line · ref:a3321df` citations; diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/0bb3210f-dc45-44da-8680-c3460d23743b/pr-739.diff`; CI status via `gh pr checks 739` (10 checks, all pass/skip); ancestry + `git log -S` output in this transcript's tool results.
- World unchanged: read-only review — no files edited, nothing committed, pushed, or posted to the PR; working tree still clean at `ba3cd08`.
- Deferred: 1 medium (REPL warning erasure), 4 low, 2 nit findings are reported only — not filed as PR comments or issues, since `/review` is read-only. Shadow-verify's "HEAD deleted the function" caveat was independently refuted and excluded; one non-load-bearing conflict over telegram stderr routing is surfaced unresolved.

---

Issue #721, opened 2026-07-27
browser_open: chromium-missing install hint is unreachable — wired to a catch block that never sees launch failures


griffinwork40:
## Symptom

In an interactive REPL session, `browser_open` failed ~9 times in a row against a reachable local URL. Every failure returned only:

```
browser_open failed: browserType.launch: Executable doesn't exist at
~/Library/Caches/ms-playwright/chromium-1234/chrome-mac-arm64/Google Chrome for Testing.app/...
```

No remediation. The model had no signal that retrying was futile, so it retried across ~9 tool-use rounds — ~13 minutes and ~$7 of an `opus_1m` session — then gave up and reported three "independent blockers", two of which were phantom (it blamed a missing Chrome debug port that was in fact listening, and a macOS Screen Recording denial that Playwright's in-process CDP capture does not need).

The install hint for this exact error **already exists in this codebase and already matches this error string.** It is wired to a `catch` block that can never see it.

## Root cause

- `src/agent/tools/handlers/playwright-hints.ts:41-50` — `PLAYWRIGHT_MISSING_HINTS` includes `"Executable doesn't exist"`; `isPlaywrightMissing()` matches it.
- `src/agent/tools/handlers/playwright-hints.ts:56-70` — `playwrightMissingHint()` returns the chromium-specific install command for precisely that substring.
- `src/agent/tools/handlers/browser-open.ts:129-134` — that check runs **only** in the `getBrowserProvider()` catch.

But provider construction never launches chromium: `src/browser/registry.ts` caches the provider wrapper, and the launch happens later in `src/browser/playwright/launcher.ts` → `ensureBrowser()` → `chromium.launch()`. So a real launch failure surfaces in the **`provider.open()`** catch at `src/agent/tools/handlers/browser-open.ts:178-196`, which performs no `isPlaywrightMissing` check and returns a bare `browser_open failed: ${msg}`.

Net: the hint is structurally unreachable for the failure mode it was written for. Its own `History:` comment (`playwright-hints.ts:10-16`) records that it was centralized specifically to stop users getting an opaque `Executable doesn't exist` error with no remedy.

## Secondary bug: the hint text is wrong for a global install

`playwrightMissingHint()` returns `pnpm exec playwright install chromium`. For a globally-installed `agent-afk` (npm / Homebrew) there is no pnpm project — that command either fails or silently targets a different playwright than the one afk bundles.

[...]

PR #748, opened 2026-07-29
fix(browser): latch launch failures so browser_* stops retrying a cold chromium launch (#722)

### LINKED ISSUE #722: BrowserLauncher keeps no failure state, so every browser_* call retries a cold chromium launch
## Symptom

When chromium cannot be launched at all (missing binary, wrong revision, OOM), every subsequent `browser_*` call pays a full cold `chromium.launch()` attempt and fails the same way. Nothing memoizes the fact that the browser is known-unlaunchable in this process.

Observed in a live REPL session as ~9 consecutive `browser_open` failures, each re-attempting the launch from scratch, over ~13 minutes.

## Mechanism

`src/browser/playwright/launcher.ts` — `BrowserLauncher` state is only `browser`, `sessions`, `launchPromise`, `shutdownComplete`. There is no failure counter, no cooldown, and no "known-broken" flag.

The `ensureBrowser()` rejection path clears `launchPromise` and rethrows, leaving `this.browser` undefined. The next `ensureBrowser()` call therefore takes the cold path and retries a full launch.

`src/browser/registry.ts` caches the `PlaywrightProvider` wrapper, but wrapper construction never launches chromium — so the cache does not memoize the failure either.

Combined with the absence of any effective repeated-failing-tool stop at the session layer, a permanently-unlaunchable browser costs one full launch attempt (plus its timeout) per tool call, N times.

## Proposed fix

Latch the failure in `BrowserLauncher`:

- On `ensureBrowser()` rejection, record the error.
- Subsequent `ensureBrowser()` calls fail fast with the cached error rather than re-attempting a launch.
- **Clear the latch on explicit `browser_close` / `shutdown()`** so a user who installs the missing binary mid-session can recover without restarting afk. The latch must not make the session permanently browser-less.

Pair with the actionable install hint so the fast-failed message is still useful, not just quick.

## Acceptance

- [ ] Second `browser_open` after a launch failure returns immediately without attempting a launch.
- [ ] The fast-failed result carries the same actionable message as the first failure.
- [ ] `browser_close` / `shutdown()` resets the latch; a subsequent `browser_open` attempts a real launch again.
- [ ] Test covers latch-set, fast-fail, and latch-reset.

griffinwork40:
## Summary

Closes #722.

`BrowserLauncher` kept no failure state, so a genuinely unlaunchable chromium cost a full `chromium.launch()` attempt — plus its timeout — on **every** `browser_*` tool call. The reported incident burned ~13 minutes across ~9 consecutive `browser_open` failures, each re-attempting the launch from scratch.

This latches the failure:

- **Latch on rejection** — `ensureBrowser()` stores the **decorated** error (the one carrying the install command added in #721), not the raw Playwright error. Later calls rethrow it without touching `chromium.launch`, so the fast-failed result keeps the same actionable remediation as the first failure — quick *and* useful.
- **Clear on recovery** — the latch is dropped on launch success, on `closeSession()`, and on `shutdown()`, so an operator who installs the missing binary mid-session recovers without restarting AFK. The latch never makes a session permanently browser-less.

### Where the reset had to go

The reset lands in **`closeSession()`**, not only `shutdown()`, because that is the path `browser_close` actually reaches — `browserCloseHandler` → `provider.close()` → `launcher.closeSession()`, and never `shutdown()` (`src/browser/playwright/index.ts:479-482`). `shutdown()` is only reached via `closeBrowserProvider()` (signals / process teardown).

It is placed **before** that method's not-found guard: a failed launch never created a `SessionEntry`, so a reset placed after the guard would be unreachable in exactly the case it exists for. Same reasoning for placing `shutdown()`'s clear before its already-shutdown guard.

### Invariants preserved

- The in-flight `launchPromise` dedupe is untouched — the latch check sits after the connected fast path and before the coalesce, and the two are never both set (the rejection handler clears the promise as it latches). Covered by a concurrent-callers regression test.
- A successful launch clears the latch unconditionally, so "browser live ⇒ no latch" holds by construction rather than by argument.
- Non-Playwright failures (OOM, timeouts) are latched **by identity**, so downstream timeout classification and witness-trace payloads still see the verbatim error.

Per the repo's ordered-operation convention, the clear/set pair is written clear-first and the governing external constraint (chromium's launchability is host filesystem state that changes with no signal to this process) is recorded as an `// Invariant:` comment at the pair, not just in the commit message.

## How was this tested?

- `pnpm lint` (tsc --noEmit, whole repo) — passes.
- `pnpm test src/browser/playwright/launcher.test.ts` — **66 passed**, including 9 new latch tests in `ensureBrowser — launch-failure latch`:
  - two sequential failed calls invoke `chromium.launch` **exactly once**
  - the fast-failed error is the *same object* and preserves the `install chromium` remediation
  - the original error survives as `cause`
  - non-Playwright failures latch by identity
  - `shutdown()` clears the latch → next call really re-launches
  - `closeSession()` (the `browser_close` path) clears the latch → next call really re-launches
  - a successful launch leaves **no** latch (verified via the crash-recovery cold path)
  - post-failure recovery reaches a live reusable browser with no stale latch
  - concurrent callers still share one in-flight launch
- `pnpm test src/browser/ …browser-open/browser-close/web-scrape` — 351 passed.
- Full `pnpm test` — **706 files / 13434 tests passed, 14 skipped, 0 failures**. No pre-existing failures on this platform, so nothing to separate out.

Reused the existing `vi.mock('playwright')` + `currentStubBrowser` harness rather than adding one.

**One existing test adjusted:** `still clears the in-flight launch promise, so a later launch can succeed` retried immediately after a failure and expected success — the exact behaviour #722 changes by design. Its assertion intent (decoration must not strand the rejected `launchPromise`) is preserved by clearing the latch before the retry; the change is annotated with a `History:` comment.

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes

### DIFF
--- a/src/agent/tools/handlers/playwright-hints.test.ts
+++ b/src/agent/tools/handlers/playwright-hints.test.ts
@@ -66,6 +66,53 @@ describe('playwrightMissingHint', () => {
 });
 
 // ---------------------------------------------------------------------------
+// Latch reset advice (issue #722 review follow-up)
+//
+// Naming the install command without naming the reset is an incomplete
+// remediation: a caller that installs chromium and retries `browser_open`
+// fast-fails on the identical latched error forever. The advice is opt-in
+// because non-latching callers exist (`afk browser login` launches chromium
+// directly, and the `browser_close` handler would be told to call itself).
+// ---------------------------------------------------------------------------
+
+describe('playwrightMissingHint — latch reset advice', () => {
+  const EXEC_MISSING = "Executable doesn't exist at /ms-playwright/chromium/chrome";
+
+  it('names the browser_close reset when the caller latched the failure', () => {
+    const hint = playwrightMissingHint(EXEC_MISSING, { headless: false, latched: true });
+    expect(hint).toMatch(/install chromium/);
+    expect(hint).toMatch(/call browser_close once to retry/);
+  });
+
+  it('OMITS the reset advice by default, so non-latching callers are not misled', () => {
+    // `afk browser login` and the per-operation handler catches reach this hint
+    // with no latch in play; telling them to call browser_close would be wrong.
+    expect(playwrightMissingHint(EXEC_MISSING, { headless: false })).not.toMatch(/browser_close/);
+    expect(playwrightMissingHint(EXEC_MISSING)).not.toMatch(/browser_close/);
+    expect(playwrightMissingHint(EXEC_MISSING, { headless: false, latched: false })).not.toMatch(
+      /browser_close/,
+    );
+  });
+
+  it('also names the reset on the package-missing branch, which latches too', () => {
+    const hint = playwrightMissingHint('Cannot find package playwright', { latched: true });
+    expect(hint).toMatch(/pnpm add playwright/);
+    expect(hint).toMatch(/call browser_close once to retry/);
+  });
+
+  it('decoratePlaywrightLaunchError defaults to NOT latched', () => {
+    const err = decoratePlaywrightLaunchError(new Error(EXEC_MISSING), false);
+    expect((err as Error).message).toMatch(/install chromium/);
+    expect((err as Error).message).not.toMatch(/browser_close/);
+  });
+
+  it('decoratePlaywrightLaunchError adds the reset when told the caller latches', () => {
+    const err = decoratePlaywrightLaunchError(new Error(EXEC_MISSING), false, true);
+    expect((err as Error).message).toMatch(/call browser_close once to retry/);
+  });
+});
+
+// ---------------------------------------------------------------------------
 // Install-command resolution (issue #721)
 // ---------------------------------------------------------------------------
 
--- a/src/browser/playwright-missing.ts
+++ b/src/browser/playwright-missing.ts
@@ -163,7 +163,26 @@ export interface PlaywrightHintOptions {
    * satisfies one does NOT satisfy the other.
    */
   headless?: boolean;
+  /**
+   * Set by callers that LATCH this failure (currently only `BrowserLauncher`).
+   * When true the hint also names the reset the latch requires, because
+   * installing the binary alone does not recover the session — see
+   * `LATCH_RESET_NOTE`. Left false by non-latching callers (`afk browser
+   * login`, the per-operation handler catches) where the advice would be wrong.
+   */
+  latched?: boolean;
 }
+
+/**
+ * Contract: appended only when `opts.latched` is true. Naming the install
+ * command without naming the reset is an incomplete remediation — a caller that
+ * follows the hint verbatim (install, then retry `browser_open`) fast-fails on
+ * the identical latched error forever, because retrying never clears the latch.
+ * `browser_close` -> `closeSession()` is the only in-session clear.
+ */
+const LATCH_RESET_NOTE =
+  'This session already latched the failure, so browser tools keep fast-failing ' +
+  'with this same error — after installing, call browser_close once to retry.';
 
 /**
  * Returns the install hint appropriate to which half of the dependency is
@@ -171,6 +190,7 @@ export interface PlaywrightHintOptions {
  */
 export function playwrightMissingHint(err: unknown, opts?: PlaywrightHintOptions): string {
   const text = flattenErrorText(err);
+  const resetNote = opts?.latched === true ? ` ${LATCH_RESET_NOTE}` : '';
 
   if (text.includes("Executable doesn't exist")) {
     // Package is installed; the chromium browser binary was never downloaded.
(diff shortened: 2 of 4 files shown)


griffinwork40 (commented):

## Merge decision: **MERGE** — with one medium follow-up worth landing before the next release

Reviewed at ref `c7fd98f` (light regime: 237 lines, 2 files, hotfix). Read-only review — nothing was edited, committed, or pushed.

The core mechanism holds up. I independently re-derived the four load-bearing claims in the PR body rather than taking them on trust, and all four are true:

- **`launchPromise` / latch are never both set** — the catch clears `launchPromise` at `launcher.ts:229` *before* `latchLaunchFailure(decorated)` at `:236`. The "checked before the coalesce" reasoning is sound. `ref:c7fd98f`, citation-type: file-state.
- **`browser_close` really does reach `closeSession()`** — `browserCloseHandler` → `provider.close()` (`browser-close.ts:55`) → `launcher.closeSession()` (`index.ts:480`), and `closeSession` clears the latch at `:532` *before* the not-found guard, so it fires for a failed launch that never created a `SessionEntry`. The registry is a process-wide singleton (`registry.ts:31,106-108`), so the close path reaches the same launcher instance holding the latch. `citation-type: file-state`.
- **No permanently browser-less session** — both clear sites exist (`:532` `closeSession`, `:572` `shutdown`), and `browser_open`'s error path calls no close/cleanup, so the latch is not self-clearing (which would have made the fix a no-op).
- **All 9 latch tests exist at the ref** and assert what the body claims.

### Findings

**1. medium · high confidence · spec-compliance/observability · `src/browser/playwright-missing.ts:187-206` · ref:c7fd98f · citation-type: file-state**

The latched error names the install command but never names the reset the latch now requires, so an operator or agent that follows the remediation verbatim never recovers within the session — the stated intent ("installing chromium mid-session recovers without restarting AFK") is mechanically supported but undiscoverable from the message.

After a latch is set, the only in-session clear is `closeSession()` via `browser_close`. An agent that reads the hint, installs chromium, then retries `browser_open` fast-fails on the identical latched error indefinitely, because retrying `browser_open` never clears the latch.

```
    return (
      'browser tools require the Playwright chromium binary. ' +
      `Install via: ${playwrightInstallCommand()}.${artifactNote}`
    );
```

Absence grounded: `git grep -n "browser_close to retry\|call browser_close\|cached from an earlier\|already failed" c7fd98f -- 'src/**/*.ts'` (tests excluded) → **zero matches**; no production string anywhere tells the caller a reset is needed after installing.

*Suggestion:* append one clause to the latched-path rethrow — e.g. "chromium was already found missing this session; after installing, call `browser_close` once to retry" — so the fast-failed message is self-sufficient.

**2. nit · high confidence · observability · `src/browser/playwright/launcher.ts:198-200` · ref:c7fd98f · citation-type: diff-context**

A latched fast-fail is textually identical to the first real failure by design, and `launcher.ts` carries no logging facility, so traces cannot distinguish "paid a full launch timeout" from "returned a cached failure" — which is the exact signal needed to confirm #722 is fixed in the field.

```
    if (this.launchFailure !== undefined) {
      throw this.launchFailure.error;
    }
```

*Suggestion:* mark the latched path in the emitted browser trace event (a boolean such as `latched: true`) so the ~13-minute regression is measurable rather than inferred.

### Per-dimension

- **security** — no issues found. The latch stores a decorated error already destined for the caller; no new sink, no secret in the payload.
- **api-compat** — no issues found. `launchFailure`, `clearLaunchFailure`, `latchLaunchFailure` are all `private`; no exported signature changed.
- **correctness** — no issues found. Crash-recovery path (`:206-209`) is reached only when no latch is set; the success handler clears unconditionally at `:225`.
- **spec-compliance** — assessed against PR title+body. Intent met; finding 1 is an incomplete-remediation gap against the intent's own recovery promise, not a missing implementation. No scope creep.
- **test-coverage** — no issues found. 9 new latch tests, including the concurrent-callers dedupe regression. The one adjusted pre-existing test is **not** vacuous: `closeSession` touches only `launchFailure` and `sessions` (`:532`, `:541`) and never `launchPromise`, so the coalesce check at `:212-213` would still return a stale rejected promise and fail the `.resolves.toBe(currentStubBrowser)` assertion under the original regression. The inserted reset only clears the latch so the retry reaches the check.
- **perf-observability** — finding 2. The perf fix itself is real: one launch attempt replaces N.

Rethrowing the `unknown`-typed latched value is safe: `decoratePlaywrightLaunchError` (`playwright-missing.ts:211-216`) returns Error-for-Error on every path and passes non-Error values through by identity — the same value that would have been thrown pre-PR — and every consumer catch site guards with `x instanceof Error ? x.message : String(x)` (`browser-open.ts:103,129,180`; `playwright-hints.ts:44-45`).

### What was not checked

- Read against branch HEAD `c7fd98f006ff60dee1b35e901d6d240deffc1e8f`; all citations verified at that ref.
- **Tests were not run** — CI on this head is green (Lint & Build, ubuntu, macOS, real-PTY scrollback, docs, publish smoke all SUCCESS); the PR's own `pnpm test` numbers were not independently reproduced.
- Stated intent: PR #748 title + body — spec-compliance assessed.
- Did not review the Telegram/daemon surfaces, and did not exercise a real chromium-missing host to confirm the ~13-minute figure empirically.

Issue #576, opened 2026-07-13
Security: browser open() skips the post-redirect domain-policy re-check that act() performs


griffinwork40:
## Summary

Browser `open()` enforces the domain allowlist **before** `page.goto()` but never re-checks the **post-redirect** URL, so a 30x from an allowed host to a blocked one lands on the blocked host and returns a full observation/screenshot. `act()` already does the post-navigation re-check — `open()` is missing the same guard (an asymmetry, not a design intent).

## Evidence

- `src/browser/playwright/index.ts:119` — `open()` calls `enforceDomainPolicy(input.url)` before navigation.
- `src/browser/playwright/index.ts:136` — `page.goto(input.url, …)` follows redirects by default; the landed URL is never re-validated in `open()`.
- `src/browser/playwright/index.ts:300-312` — `act()` **does** re-check `page.url()` after navigation and `goBack()`s on a blocked host. This is the guard `open()` lacks.

## Impact

HIGH for operators who configure `AFK_BROWSER_ALLOWED_DOMAINS` (the allowlist is fail-open when unset, so this only bites opt-in users — but for them it fully defeats the allowlist). `browser_open("https://allowed.example/redirect→blocked")` returns content from the blocked host.

## Suggested fix

Hoist the `act()` post-navigation re-check into a shared helper and call it in `open()` immediately after `page.goto()` (and after any in-`open` redirect settle). On a blocked landed URL, return `blocked_by_policy` (and best-effort `goBack()` / close the page) rather than surfacing the observation.

## Acceptance criteria

- [ ] `open()` re-validates `page.url()` after navigation against the same policy `act()` uses.
- [ ] A redirect from an allowed host to a blocked host yields `blocked_by_policy`, not an observation/screenshot.
- [ ] Shared helper used by both `open()` and `act()` (no duplicated policy logic).
- [ ] Test: allowed→blocked redirect is blocked; allowed→allowed redirect proceeds.

Refs: 2026-07-13 sandboxing audit, Finding 3 (HIGH).

PR #750, opened 2026-07-29
fix(web): block SSRF egress to internal hosts in web_scrape (#575)

### LINKED ISSUE #575: Security: web_scrape has no SSRF/internal-host filtering (cloud metadata, localhost, RFC1918 reachable)
## Summary

`web_scrape` performs no SSRF / internal-host filtering. It validates only the URL **scheme** (http/https) and then fetches any host — including `127.0.0.1`, `0.0.0.0`, `::1`, link-local **`169.254.169.254` (cloud instance-metadata → credential theft)**, and all RFC1918 internal ranges. Because the model can be prompt-injected by page content it scrapes, this is a real, model-reachable SSRF vector.

## Evidence

- `src/agent/tools/handlers/web-scrape.ts:110-112` — the only validation is `parsedUrl.protocol !== 'http:' && !== 'https:'`. No host/IP check.
- `web-scrape.ts:211` (raw mode) / `src/web/scrape.ts` — `retryFetch(... redirect: follow)` then GETs the host directly; the domain-allowlist that gates the **browser** is not consulted on the web path, and the `render` fallback deliberately bypasses it too.
- The repo already ships host classifiers that are **not reused here**: `isLoopback()` (`src/agent/mcp/transport.ts:52-58`) and `isLoopbackHost()` (`src/cli/daemon-options.ts:199`).

## Impact

HIGH. A scraped page can instruct the model to fetch `http://169.254.169.254/latest/meta-data/iam/security-credentials/…` (or an internal admin endpoint) and return the body. No allowlist, no IP filter, applied per-redirect-hop either.

## Suggested fix

Add a shared egress guard applied in `parseInput` **and** re-applied on each redirect hop: reject loopback (reuse `isLoopback`), link-local `169.254.0.0/16` + `fe80::/10`, RFC1918 (`10/8`, `172.16/12`, `192.168/16`), ULA `fc00::/7`, and `0.0.0.0`. Resolve DNS and check the resolved IP (guard against DNS-rebinding). Provide an opt-in escape hatch (`AFK_WEB_ALLOW_PRIVATE_HOSTS=1`) for legitimate localhost/dev scraping.

## Acceptance criteria

- [ ] `web_scrape` (raw + markdown + render paths) refuses loopback / link-local / RFC1918 / ULA hosts by default, on the initial URL and after redirects.
- [ ] DNS resolves to a public IP before the fetch proceeds (rebinding guard).
- [ ] `AFK_WEB_ALLOW_PRIVATE_HOSTS=1` opt-in restores private-host access (registered in `ENV_REGISTRY`).
- [ ] Tests: metadata IP, `127.0.0.1`, RFC1918, redirect-to-internal all blocked; public host + opt-in allowed.

Refs: 2026-07-13 sandboxing audit, Finding 2 (HIGH). Report: `~/.afk/workspace/reports/sandbox-audit-2026-07-13.md`.

griffinwork40:
## Summary

Fixes #575 (P0 / `priority: critical`). `web_scrape` validated only the URL **scheme** and then fetched any host. Because the tool is an always-on builtin that returns page content into model context, a scraped page could prompt-inject a follow-up fetch of `169.254.169.254` and read back cloud IAM credentials — plus reach `localhost` and RFC1918 internal services. Redirects were followed with `redirect: 'follow'`, which is opaque to app code, so even a validated initial URL could land on internal space.

New module `src/web/egress-guard.ts` (282 LOC) is the single classifier every egress path consults:

- **Blocked ranges** — loopback (`127/8`, `::1`), link-local (`169.254/16` incl. the metadata IP, `fe80::/10`), RFC1918 (`10/8`, `172.16/12`, `192.168/16`), CGNAT (`100.64/10`), ULA (`fc00::/7`), `0.0.0.0/8`, and the IPv4-mapped / IPv4-compatible / NAT64 IPv6 encodings of those. Built on `net.BlockList`; I verified empirically that it unmaps `::ffff:a.b.c.d` against the registered **IPv4** rules, so `::/96` and `64:ff9b::/96` are listed explicitly to cover the two forms it does *not* unmap. URL parsing also normalizes the decimal/hex/short-form encodings (`http://2130706433/`, `http://127.1/`, `http://0/`) before classification — covered by tests.
- **DNS-rebinding guard** — resolves the hostname and classifies the **resolved** addresses, not the hostname string, so a host with an A record of `127.0.0.1` is refused. Any one internal record blocks. Resolution *failure* does **not** block (fetch surfaces the truthful network error instead of a confusing SSRF refusal).
- **Per-hop redirect re-check** — see below.

Wired into **all three** egress paths — raw mode, the markdown plain fetch, and the headless-render escalation. The render path is a *second, independent* egress path (chromium does its own DNS and follows redirects internally), so it gets a pre-check **plus** a post-navigation `finalUrl` re-check, mirroring the pre/post pattern `act()` already uses for the domain policy. An egress refusal is **terminal** on both paths: it never degrades into the render escalation, and never degrades into thin fetched content — either would hand back the internal fetch the guard just refused. Refusals surface as a non-crashing `web_scrape blocked: …` tool error naming the offending address and the opt-out.

**Redirect mechanism:** `guardedFetch()` forces `redirect: 'manual'` and walks the chain itself, validating each hop *before* it is requested (relative and protocol-relative `Location` values resolved against the responding URL first), bounded at 20 hops. I confirmed by probe that native `redirect: 'follow'` offers no interception point, so owning the loop is the only way to re-check per hop. It wraps `retryFetch`, so transient 429/5xx resilience is preserved.

**Opt-out:** `AFK_WEB_ALLOW_PRIVATE_HOSTS` (default off = guard active), registered in `ENV_REGISTRY` with `docs/env-registry.{json,md}` regenerated via `pnpm scan:env`.

One ordering subtlety worth a reviewer's eye: the handler's pre-check is placed **after** the abort listener and timeout are wired. It is the handler's first `await` (the guard resolves DNS, so it cannot live in the synchronous `parseInput`), and awaiting before that wiring drops a parent abort landing inside the window — leaving the later fetch waiting on a cancel that never fires. I hit exactly that as a test failure and fixed the ordering rather than the test.

## How was this tested?

- `pnpm lint` (tsc --noEmit, whole repo) — **pass**
- `pnpm test` — **pass, 707 files / 13,505 tests, 0 failures** (14 pre-existing skips). No pre-existing failures found on the base commit.
- `pnpm scan:env:check`, `pnpm audit:env:check`, `pnpm audit:sdk:check`, `pnpm audit:chalk:check` — all pass.
- **62 new guard unit tests** (`src/web/egress-guard.test.ts`): each blocked range and its IPv6 encodings, near-miss boundaries (`172.15`/`172.32`, `100.63`/`100.128`), the DNS-rebinding case, mixed public+internal records, per-redirect-hop blocking (absolute, relative, protocol-relative), redirect-cycle bound, and the env opt-out.
- **Integration tests on both surfaces**: `scrape.test.ts` proves an egress refusal does not fall through to the render escalation and that the render path's `finalUrl` re-check fires; `web-scrape.test.ts` asserts the `web_scrape blocked:` refusal in raw *and* markdown mode and that no request is issued.
- A `lookupFn` seam (matching the existing `fetchFn` / `renderFn` convention) keeps every test off real DNS — the pre-existing suites were made hermetic too, since the guard would otherwise have resolved `example.com` for real.
- **Verified non-vacuous by mutation**: neutering the guard to always-allow fails 54 of these tests.

Files: `src/web/egress-guard.ts` (+282, new), `src/web/egress-guard.test.ts` (+359, new), `src/web/scrape.ts` (+34), `src/agent/tools/handlers/web-scrape.ts` (+36, 347 LOC — under the 350 ceiling), `src/web/index.ts` (+8), `src/config/env.ts` (+21), plus regenerated env docs and the two test files.

## Notes for the reviewer

- **Known conflict (correction to the filed expectation):** I was told PR #735 (`afk/read-floor-browser-secrets`) also edits `docs/env-registry.json`/`.md` and that this branch would need a registry-regeneration rebase if #735 merges first. **That is not the case as of this writing** — #735 touches only `glob/grep/read-denylist`, the bash-restriction hook, and a website doc; it modifies neither `docs/env-registry.*` nor `src/config/env.ts`. I checked all 5 open PRs and **none** touch the env registry or `src/web/`, and `git merge-tree` against current `origin/main` (b1998f5) reports no conflicts. If a registry-touching PR lands first, the fix is the trivial `pnpm scan:env` regeneration.
- **Scope:** #575 also suggests fixing #576 (the browser `open()` post-redirect re-check) in the same pass. I deliberately left that alone — it is already covered by open PR #749 (`afk/browser-open-policy-576`), so touching it here would collide. This PR is the `web_scrape` surface only.
- The `render()` docstring's "does NOT enforce the allowlist" contract is unchanged and still true: the *interactive domain allowlist* governs agent-driven navigation, and that is a separate policy from this egress guard, which now applies to render as well.

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes

### DIFF
--- a/docs/env-registry.json
+++ b/docs/env-registry.json
@@ -813,6 +813,14 @@
       "category": "model"
     },
     {
+      "name": "AFK_WEB_ALLOW_PRIVATE_HOSTS",
+      "description": "Opt out of the web_scrape SSRF egress guard. When unset (default) the guard is ACTIVE: the markdown, raw, and headless-render paths refuse loopback (127/8, ::1), link-local (169.254/16 — including the 169.254.169.254 cloud instance-metadata endpoint), RFC1918 (10/8, 172.16/12, 192.168/16), carrier-grade NAT (100.64/10), IPv6 unique-local (fc00::/7), 0.0.0.0/8, and the IPv4-mapped/compatible IPv6 forms of all of those. Hostnames are resolved and the RESOLVED addresses are classified (DNS-rebinding guard), and the check is re-applied on every redirect hop. Set to 1/true to allow private-host access — needed only to scrape a local dev server. Enabling it restores a model-reachable SSRF path (issue #575).",
+      "type": "boolean",
+      "required": false,
+      "example": "1",
+      "category": "misc"
+    },
+    {
       "name": "AFK_WORKTREE_AUTONAME",
       "description": "Auto-rename worktree branches based on the first user message in interactive mode. 1 = on (default), 0 = off.",
       "type": "boolean",
--- a/docs/env-registry.md
+++ b/docs/env-registry.md
@@ -2,7 +2,7 @@
 
 Generated from `src/config/env.ts`. Do not edit by hand — run `pnpm scan:env` after changing the registry source.
 
-**137 vars** across 12 categories. Every `process.env[...]` read in `src/` outside `src/config/env.ts` is a CI failure (enforced by `pnpm audit:env:check`).
+**138 vars** across 12 categories. Every `process.env[...]` read in `src/` outside `src/config/env.ts` is a CI failure (enforced by `pnpm audit:env:check`).
 
 To add a var: edit `src/config/env.ts` (add a getter on `env` + an entry in `ENV_REGISTRY`), then run `pnpm scan:env`.
 
@@ -200,5 +200,6 @@ To add a var: edit `src/config/env.ts` (add a getter on `env` + an entry in `ENV
 | `AFK_THEME` | string |  | `dark` | `light` | TUI color palette for the interactive REPL and all CLI rendering: dark \| light \| auto. Display-only — swaps the semantic color palette, never behavior (cost/latency unaffected). auto detects from COLORFGBG and falls back to dark. Overridden per-launch by --theme and mutable mid-session via /theme. Precedence: --theme flag > this env > config theme > auto-detect > dark. Invalid values are ignored (dark). |
 | `AFK_THINKING_UI` | string |  | `live` | `digest` | Default thinking-display mode for the interactive REPL: summary \| live \| digest \| off. Display-only — controls how extended-thinking blocks render, never whether thinking runs (cost/latency unaffected). Overridden per-launch by --thinking-ui and mutable mid-session via /thinking. Precedence: --thinking-ui flag > this env > interactive.thinkingUi config > live. Invalid values are ignored. |
 | `AFK_USER_CARD_MAX_ROWS` | number |  |  | `24` | Maximum number of visual rows emitted by renderUserCard before collapsing the remainder into a dim "…(N lines collapsed)" summary row. Defaults to 24. Non-integer or non-positive values are silently ignored and the default applies. |
+| `AFK_WEB_ALLOW_PRIVATE_HOSTS` | boolean |  |  | `1` | Opt out of the web_scrape SSRF egress guard. When unset (default) the guard is ACTIVE: the markdown, raw, and headless-render paths refuse loopback (127/8, ::1), link-local (169.254/16 — including the 169.254.169.254 cloud instance-metadata endpoint), RFC1918 (10/8, 172.16/12, 192.168/16), carrier-grade NAT (100.64/10), IPv6 unique-local (fc00::/7), 0.0.0.0/8, and the IPv4-mapped/compatible IPv6 forms of all of those. Hostnames are resolved and the RESOLVED addresses are classified (DNS-rebinding guard), and the check is re-applied on every redirect hop. Set to 1/true to allow private-host access — needed only to scrape a local dev server. Enabling it restores a model-reachable SSRF path (issue #575). |
 | `AFK_WRITE_DENYLIST` | string |  |  | `**/.env,**/secrets/**` | Comma-separated list of additional path globs that the write_file tool refuses to write to. |
 | `AFK_WRITE_DIFF` | boolean |  |  |  | Show a diff preview before each write_file tool call. Defaults provider-controlled when unset. |
--- a/package.json
+++ b/package.json
@@ -129,6 +129,7 @@
     "string-width": "^8.2.0",
     "telegraf": "^4.16.3",
     "turndown": "^7.2.4",
+    "undici": "^7.27.0",
     "wrap-ansi": "^10.0.0",
     "zod": "^4.3.6"
   },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -68,6 +68,9 @@ importers:
       turndown:
         specifier: ^7.2.4
         version: 7.2.4
+      undici:
+        specifier: ^7.27.0
+        version: 7.27.0
       wrap-ansi:
         specifier: ^10.0.0
         version: 10.0.0
--- a/src/agent/tools/handlers/web-scrape.test.ts
+++ b/src/agent/tools/handlers/web-scrape.test.ts
@@ -32,6 +32,18 @@ function makeFetch(handler: (url: string, init: RequestInit) => Promise<Response
 
 const signal = (): AbortSignal => new AbortController().signal;
 
+/**
+ * Hermetic DNS seam for the SSRF egress guard (issue #575). Every handler in
+ * this file injects it so the guard classifies a fixed PUBLIC address instead of
+ * issuing a real `dns.lookup` — that keeps the suite offline and, critically,
+ * keeps the guard's pre-check off the macrotask queue so the cancellation tests
+ * still observe fetch being reached. Guard behaviour itself is covered in
+ * `src/web/egress-guard.test.ts`.
+ */
+const publicLookup = async (): Promise<readonly { address: string }[]> => [
+  { address: '93.184.216.34' },
+];
+
 /** A content-rich, server-rendered article that survives Readability. */
 function richArticleHtml(marker = 'main article content'): string {
   const paras = Array.from(
@@ -61,6 +73,7 @@ describe('web_scrape handler — input validation', () => {
   const handler = createWebScrapeHandler({
     fetchFn: makeFetch(() => makeResponse({ body: 'unused' })),
     env: {},
+    lookupFn: publicLookup,
   });
 
   it('rejects non-object input', async () => {
@@ -122,7 +135,7 @@ describe('web_scrape handler — markdown mode (fetch-first)', () => {
   it('fetches the URL, extracts main content, and returns markdown — no render', async () => {
     const renderFn = vi.fn<RenderFn>(async () => ({ html: '', finalUrl: '', httpStatus: 200 }));
     const fetchFn = makeFetch(() => makeResponse({ contentType: 'text/html', body: richArticleHtml() }));
-    const handler = createWebScrapeHandler({ fetchFn, env: {}, renderFn });
+    const handler = createWebScrapeHandler({ fetchFn, env: {}, renderFn, lookupFn: publicLookup });
 
     const r = await handler({ url: 'https://example.com/article' }, signal());
     expect(r.isError).toBeUndefined();
@@ -135,7 +148,7 @@ describe('web_scrape handler — markdown mode (fetch-first)', () => {
 
   it('requires no API key for markdown mode', async () => {
     const fetchFn = makeFetch(() => makeResponse({ contentType: 'text/html', body: richArticleHtml() }));
-    const handler = createWebScrapeHandler({ fetchFn, env: {} });
+    const handler = createWebScrapeHandler({ fetchFn, env: {}, lookupFn: publicLookup });
     const r = await handler({ url: 'https://example.com/article' }, signal());
     expect(r.isError).toBeUndefined();
     expect(r.content).toContain('main article content');
@@ -149,7 +162,7 @@ describe('web_scrape handler — markdown mode (fetch-first)', () => {
       finalUrl: 'https://example.com/app',
       httpStatus: 200,
     }));
-    const handler = createWebScrapeHandler({ fetchFn, env: {}, renderFn });
+    const handler = createWebScrapeHandler({ fetchFn, env: {}, renderFn, lookupFn: publicLookup });
 
     const r = await handler({ url: 'https://example.com/app' }, signal());
     expect(r.isError).toBeUndefined();
@@ -162,7 +175,7 @@ describe('web_scrape handler — markdown mode (fetch-first)', () => {
     const fetchFn = makeFetch(() => makeResponse({ contentType: 'text/html', body: empty }));
     // Render also yields nothing.
     const renderFn = vi.fn<RenderFn>(async () => ({ html: empty, finalUrl: 'https://e.com', httpStatus: 200 }));
-    const handler = createWebScrapeHandler({ fetchFn, env: {}, renderFn });
+    const handler = createWebScrapeHandler({ fetchFn, env: {}, renderFn, lookupFn: publicLookup });
 
     const r = await handler({ url: 'https://example.com/empty' }, signal());
     expect(r.isError).toBe(true);
@@ -176,7 +189,7 @@ describe('web_scrape handler — markdown mode (fetch-first)', () => {
     const renderFn = vi.fn<RenderFn>(async () => {
       throw new Error('Cannot find package playwright');
     });
-    const handler = createWebScrapeHandler({ fetchFn, env: {}, renderFn });
+    const handler = createWebScrapeHandler({ fetchFn, env: {}, renderFn, lookupFn: publicLookup });
 
     const r = await handler({ url: 'https://example.com/x' }, signal());
     expect(r.isError).toBe(true);
@@ -202,7 +215,7 @@ describe('web_scrape handler — markdown mode (fetch-first)', () => {
     const renderFn = vi.fn<RenderFn>(async () => {
       throw decorated;
     });
-    const handler = createWebScrapeHandler({ fetchFn, env: {}, renderFn });
+    const handler = createWebScrapeHandler({ fetchFn, env: {}, renderFn, lookupFn: publicLookup });
 
     const r = await handler({ url: 'https://example.com/x' }, signal());
     expect(r.isError).toBe(true);
@@ -221,7 +234,7 @@ describe('web_scrape handler — raw mode', () => {
       seenInit = init;
       return makeResponse({ body: '{"ok":true}' });
     });
-    const handler = createWebScrapeHandler({ fetchFn, env: {} });
+    const handler = createWebScrapeHandler({ fetchFn, env: {}, lookupFn: publicLookup });
     const r = await handler({ mode: 'raw', url: 'https://api.example.com/v1/x' }, signal());
     expect(r.isError).toBeUndefined();
     expect(seenUrl).toBe('https://api.example.com/v1/x');
@@ -233,7 +246,7 @@ describe('web_scrape handler — raw mode', () => {
 
   it('returns isError on non-2xx response', async () => {
     const fetchFn = makeFetch(() => makeResponse({ status: 404, statusText: 'Not Found', body: '' }));
-    const handler = createWebScrapeHandler({ fetchFn, env: {} });
+    const handler = createWebScrapeHandler({ fetchFn, env: {}, lookupFn: publicLookup });
     const r = await handler({ mode: 'raw', url: 'https://example.com/missing' }, signal());
     expect(r.isError).toBe(true);
     expect(r.content).toMatch(/HTTP 404 Not Found/);
@@ -243,7 +256,7 @@ describe('web_scrape handler — raw mode', () => {
     const fetchFn = makeFetch(() => {
       throw new Error('ECONNREFUSED');
     });
-    const handler = createWebScrapeHandler({ fetchFn, env: {} });
+    const handler = createWebScrapeHandler({ fetchFn, env: {}, lookupFn: publicLookup });
     const r = await handler({ mode: 'raw', url: 'https://example.com' }, signal());
     expect(r.isError).toBe(true);
     expect(r.content).toMatch(/network error: ECONNREFUSED/);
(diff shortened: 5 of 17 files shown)


griffinwork40:
@codex fix comments

griffinwork40 (commented):

## Verdict: **MERGE**

Reviewed against branch HEAD `e0d033b`. Every load-bearing claim in the PR body was re-derived against the code at that ref rather than taken on faith; all of them hold. No blocking findings. The three items below are residual-risk and hardening notes, not merge blockers.

**Verified claims** (each checked at `e0d033b`, not from the description):
- **Terminal refusal on both egress paths** — `src/web/scrape.ts:161` re-throws `EgressBlockedError` before the escalation decision, and `:217` re-throws it on the render path before the degrade-to-thin-content fallback. The invariant the body claims is genuinely enforced at both sites, not just documented.
- **All resolved records classified** — `src/web/egress-guard.ts:193-196` loops every DNS record and blocks if any one is internal; not first-record-only.
- **Fails closed** — `egress-guard.ts:114-117` opts out only on exact `'1'`/`'true'` (post-trim/lowercase). Unset, `''`, `'false'`, `'0'`, `'yes'` all leave the guard active.
- **Range coverage complete for the stated threat** — metadata `169.254.0.0/16` (`:48`), loopback `127/8` + `::/96`, RFC1918, CGNAT `100.64.0.0/10`, ULA `fc00::/7`, plus NAT64 `64:ff9b::/96`.
- **Per-hop validation ordering** — `egress-guard.ts:261-262`: `assertEgressAllowed(target)` precedes the `retryFetch` for that hop, so a blocked hop is never requested.
- **Handler abort ordering** — the guard pre-check at `web-scrape.ts:225` does sit after the abort listener (`:211`) and timer (`:212`) are wired, with an abort re-check at `:226`. The subtlety called out in the description is real and correctly handled.
- **Test count** — "62 new guard unit tests" is accurate: 25 direct `it()` plus 37 loop-generated from the four range tables and the two opt-out tables. Protocol-relative redirect coverage exists (`egress-guard.test.ts:293`).

### Findings

**1. medium · confidence high · security · `src/web/egress-guard.ts:261-262` · ref:e0d033b · citation-type:file-state**
finding: The classified IP address is never pinned to the connection — the guard resolves the hostname, then hands the same URL *string* to `retryFetch`, which re-resolves it, so a low-TTL attacker-controlled record that answers public at check time and internal at fetch time is not blocked (classic TOCTOU rebinding).
evidence:
```ts
await assertEgressAllowed(target, guardOpts);
const res = await retryFetch(fetchFn, target, { ...init, redirect: 'manual' }, opts.retry ?? {});
```
suggestion: Document this as a known residual in the module docstring; closing it properly needs connection-level pinning (a custom undici dispatcher/`lookup` that reuses the vetted address). Not a regression — the static-record rebinding case *is* blocked (`:193-196`) — so this is follow-up scope, not a blocker.

**2. low · confidence high · correctness · `src/web/egress-guard.ts:262` · ref:e0d033b · citation-type:file-state**
finding: The redirect loop forwards the identical `init` to every hop with no status-conditional handling, so a 303 does not downgrade to GET and any body/`Authorization`/`Cookie` would be replayed verbatim to each subsequent origin.
evidence:
```ts
const res = await retryFetch(fetchFn, target, { ...init, redirect: 'manual' }, opts.retry ?? {});
```
suggestion: Not reachable today — both production call sites (`scrape.ts:122`, `web-scrape.ts:242`) are GET-only with no credentials in `FETCH_HEADERS` (`scrape.ts:36-43`), so severity is downgraded accordingly. But `guardedFetch` is publicly exported (`src/web/index.ts:16`), so a future non-GET caller inherits the hazard: drop the body and re-target to GET on 303, and strip credential headers on cross-origin hops.

**3. low · confidence high · test-coverage · `src/web/egress-guard.ts:178` · ref:e0d033b · citation-type:file-state**
finding: A trailing-dot hostname (`http://127.0.0.1./`) is not classified as an IP literal and is untested — `git grep -niE "trailing.?dot|toUpperCase" e0d033b -- src/web/egress-guard.ts src/web/egress-guard.test.ts` returns zero matches.
suggestion: The defense still holds by construction — `isIP()` returns 0, routing it to the DNS path where the resolved address is classified — so this is a coverage gap, not a bypass. Add the trailing-dot form to the blocked table to lock the behavior in.

### What was not checked
- Read against branch HEAD `e0d033b`; all citations verified at that ref. Blocking-tier claims were independently re-derived on the main path.
- **Stated intent: PR #750 title + body — spec-compliance assessed.** No unmet intent found; no scope creep. The body's deliberate exclusion of the browser `open()` fix (#576, owned by PR #749) holds — this diff touches no `browser-*` handler.
- Tests were **not executed** (read-only review); CI is green on all 7 required checks at this SHA, and the mutation-non-vacuity claim (54 failures on a neutered guard) was not re-run.
- **Chromium subresource egress on the render path was not traced.** The pre-check and `finalUrl` post-check (`scrape.ts:188`, `:195`) cover top-level navigation; whether an in-page subresource or iframe to internal space can reach the extracted DOM was not verified in this pass and is worth a maintainer's eye.
- `retryFetch`'s reuse of `init.body` across its own internal retry attempts (stream-body semantics) was out of scope.

griffinwork40 (commented):

## Merge decision: **DO NOT MERGE** — 1 blocking finding

Reviewed at branch HEAD `16a4755` (`refs/pull/750/head`), base `b1998f5`. Regime: full (17 files, +1200/−49). Stated intent: PR #750 title + body — spec-compliance assessed.

The first commit (`e0d033b`, the SSRF guard itself) reads clean. The blocker is introduced **entirely by the second commit** (`16a4755`, "close SSRF guard TOCTOU gaps") and is verified by live runtime reproduction.

---

### 🔴 BLOCKING

**high · high confidence · correctness** — `src/web/egress-guard.ts:139-168` · ref `16a4755` · citation-type: file-state

The `guardedDispatcher` lookup callback uses the **single-address** callback form, but Node invokes `connect.lookup` with `{ all: true }`, which requires an **array** of `{address, family}` records — so every production `web_scrape` of a **hostname** URL fails with `TypeError: fetch failed` / cause `ERR_INVALID_IP_ADDRESS` before any request is issued.

```js
// egress-guard.ts:162
callback(null, selected.address, isIP(selected.address));
```

The failing branch is exactly the production configuration — the dispatcher is attached only when the real global fetch is in use and the opt-out is off:

```js
// egress-guard.ts:305-306
...(fetchFn === globalThis.fetch && !(opts.allowPrivateHosts ?? privateHostsAllowed())
  ? { dispatcher: guardedDispatcher }
```

`web_scrape` is an always-on builtin whose handler defaults `fetchFn` to `globalThis.fetch` (`src/agent/tools/handlers/web-scrape.ts:167`, `src/web/scrape.ts:112`), so the default install is affected. Scope of the break, stated precisely: hostname URLs fail; **bare-IP URLs keep working** (they skip DNS entirely), and setting `AFK_WEB_ALLOW_PRIVATE_HOSTS=1` detaches the dispatcher and also masks it. The guard fails *closed*, so this is an availability/functionality break, not a security regression.

**Reproduction** (faithful port of `egress-guard.ts:139-167` against the pinned `undici@7.27.0`, local HTTP server, only `defaultLookup` stubbed to a public address so the block branch is not taken):

```
cb(null, addr, family)            → THREW ERR_INVALID_IP_ADDRESS   | options={"hints":1024,"all":true}
cb(null, [{address, family}])     → status=200 OK                  | options={"hints":1024,"all":true}
```

Note the observed `all: true` in the options undici passed to the callback — that is the direct cause.

**suggestion:** return the array form — `callback(null, [{ address: selected.address, family: isIP(selected.address) }])` — or pin `connect: { family: 4 }` so the lookup is invoked without `all`. Add one test that exercises the real dispatcher branch (see below), since no current test can catch this.

> ⚠️ **Green CI is not evidence against this finding.** All checks pass on `16a4755` *because* no test reaches the dispatcher branch — every test injects a `fetchFn`/`lookupFn` seam, and an injected mock can never satisfy `fetchFn === globalThis.fetch`. The passing suite and the broken production path are consistent with each other.

---

### Non-blocking

**medium · high confidence · test-coverage** — `src/web/egress-guard.ts:305` · ref `16a4755` · citation-type: file-state
The `fetchFn === globalThis.fetch` identity check makes the entire production egress branch structurally unreachable from tests, so the TOCTOU dispatcher — the security-critical half of the second commit — has zero coverage. Grep evidence: `git grep -n "guardedDispatcher" 16a4755 -- 'src/web/*.test.ts' 'src/agent/tools/handlers/web-scrape.test.ts'` returns zero matches (the symbol appears only at `egress-guard.ts:139,306`).
**suggestion:** export `guardedDispatcher` (or its lookup callback) and unit-test the callback contract directly against a local server, independent of the `fetchFn` identity gate.

**low · high confidence · api-compat** — `package.json:132` · ref `16a4755` · citation-type: diff-context
The second commit adds `undici@^7.27.0` as a **runtime dependency** on a project already requiring Node `>=22.0.0`, which ships undici internally — duplicating the library in the dependency tree solely to construct the dispatcher. Zero production importers outside `src/web/egress-guard.ts`.
**suggestion:** if the blocker is fixed via `connect: { family: 4 }` or Node's built-in dispatcher plumbing, drop the dependency; otherwise document why the userland copy is required.

**low · high confidence · spec-compliance (disclosure)** — PR body "Files:" section
The body enumerates 8 files, but the diff touches 17 — undisclosed: `src/browser/playwright/launcher.ts`, `src/browser/playwright/index.ts`, `src/browser/types.ts`, `src/browser/playwright/launcher.test.ts`, `package.json`, `pnpm-lock.yaml`. The body also does not mention that a second commit added a new runtime dependency.
**Explicitly cleared, not scope creep:** the browser edits harden the *render* egress path, which the body does place in scope; they do not touch the `open()` post-redirect re-check the body defers to PR #749.
**suggestion:** refresh the "Files:" list and name the new dependency before merge.

---

### Verified sound (no issues found)

- **Render-path egress guard** — `src/browser/playwright/launcher.ts:341-394`: `context.route('**/*')` routes every subresource through the guard, and the post-`page.content()` `guardError` re-check catches a request a script starts after the load event. Read at `16a4755`.
- **Redirect walk** — `src/web/egress-guard.ts:285-329`: `redirect: 'manual'` with per-hop `assertEgressAllowed` *before* each request, `Location` resolved against the responding URL (relative and protocol-relative), a hop bound, and `res.body?.cancel()` freeing each socket. Read at `16a4755`.
- **Opt-out parsing** — `egress-guard.ts:115-118`: defaults to guard-active; only `1`/`true` (trimmed, lowercased) disable it. Fail-closed on malformed values.
- **File-size ceiling** — `egress-guard.ts` 329 LOC and `web-scrape.ts` 347 LOC are both under the 350 ceiling. `launcher.ts` is 619 LOC but was already 595 at base — pre-existing, not introduced here.
- **Test-count claim** — the body's "62 new guard unit tests" is consistent with 31 `it()` call-sites expanded by 7 table loops in `egress-guard.test.ts`. No discrepancy asserted.

### What was not checked

- Wave 1 agents and the verifier read against branch HEAD `16a4755`; all blocking-finding citations were re-read at that ref. No fabricated or diff-only citations survived.
- The blocking finding was independently re-derived by a shadow verifier from Node's `net.js` `lookupAndConnect`/`lookupAndConnectMultiple` and undici `7.27.0`'s `connect.js` — **CONFIRMED**, agreeing with the live repro by a different method.
- Did **not** run `pnpm test` or `pnpm lint` (CI reports both green on this ref).
- Did **not** execute the test suite to confirm the runtime test count, verify the `docs/env-registry.*` regeneration byte-for-byte, or exhaustively execute every IPv6/NAT64 blocklist range (ranges were read, not runtime-enumerated).
- Did **not** review interaction with open PRs #749 or #765.

griffinwork40 (commented):

**Reviewed ref:** `16a4755` (branch `afk/ssrf-guard-575`, 2 commits) · **Regime:** full · 17 files, ~1221 insertions / 75 deletions (lock files, `package.json`, and generated env docs excluded from review).

## Merge decision: ✅ MERGE

No `critical` or `high` finding survived citation + absence verification. The residual item below is a **hardening follow-up, not a regression**: before this PR the render path had no egress guard at all, so the change is strictly an improvement over `main`. Blocking a P0 SSRF fix on it would be the wrong trade.

---

## Findings

### 1. `medium` · confidence: high · security/correctness
`src/browser/playwright/launcher.ts:352-355`, `src/web/scrape.ts:197`, `src/web/egress-guard.ts:139-168` · ref:`16a4755` · citation-type: file-state

**finding:** The headless-render (Chromium) egress path retains the DNS-rebinding TOCTOU gap that the plain-fetch path closes — `requestGuard` performs a Node-side lookup and then calls bare `route.continue()`, so Chromium resolves the hostname a second time, independently, at connect time.

**evidence:**
```js
await opts.requestGuard!(route.request().url());
await route.continue();
```
On the fetch path, `guardedDispatcher`'s `connect.lookup` makes the classified address *be* the socket target (`callback(null, selected.address, isIP(selected.address))`), so check-time and use-time agree by construction. There is no counterpart for Chromium: greps across the diff for `proxy`, `--host-resolver-rules`, `hostResolverRules`, and `launch(` find no pinning mechanism, and `EgressVerdict` / `assertEgressAllowed` return `void` / allow-deny only — no IP is ever surfaced that *could* be pinned.

**suggestion:** Pin Chromium's resolution the way the Undici dispatcher does — carry the pre-checked IP forward via a per-origin host-resolver override or proxy — rather than relying on a second, disconnected lookup.

**Severity note:** initially rated `high`; an independent shadow-verification pass downgraded it to `medium`. It is production-reachable (a normal `web_scrape` call escalates to render on thin content), but exploitation requires attacker control of a domain's DNS records, forcing the render escalation, *and* winning a per-request timing race — unlike the deterministic single-shot bypass that existed before this PR.

### 2. `low` · confidence: high · test-coverage
`src/web/egress-guard.ts:139,306` · ref:`16a4755` · citation-type: file-state

**finding:** `guardedDispatcher` — the Undici custom-DNS-lookup mechanism that closes the fetch-path TOCTOU gap — has zero test coverage repo-wide.

**evidence:** `git grep -n guardedDispatcher 16a4755` → exactly 2 hits, both inside `egress-guard.ts` (definition `:139`, use `:306`); zero in any test file. This is structural, not incidental: the dispatcher attaches only when `fetchFn === globalThis.fetch` (`:305`), a condition no test injecting a fake `fetchFn` can satisfy, so the `connect.lookup` body cannot execute under the current suite.

**suggestion:** Extract the `connect.lookup` callback into an exported function and unit-test it directly with a stubbed `lookupFn`, asserting the `EgressBlockedError` path and the `callback(null, address, family)` path.

### 3. `low` · confidence: high · spec-compliance (scope disclosure)
`package.json`, `src/browser/playwright/launcher.ts`, `src/web/egress-guard.ts` · ref:`16a4755`

**finding:** The PR body documents only commit 1 (`e0d033b`); commit 2 (`16a4755`, "close SSRF guard TOCTOU gaps") adds a new runtime dependency (`undici ^7.27.0`, not present on `main`), the `guardedDispatcher` mechanism, and the whole browser-launcher interception surface — none of which the description names.

**suggestion:** Add a paragraph covering commit 2, calling out the new `undici` runtime dependency explicitly so it gets dependency review.

### 4. `nit` · confidence: high · test-coverage
`src/web/egress-guard.test.ts` · ref:`16a4755`

**finding:** No test exercises trailing-dot FQDN notation (`http://127.0.0.1./`). The behavior is *correct* — WHATWG `URL` normalizes the trailing dot away before classification — but it is unpinned by any test.

**suggestion:** Add `http://127.0.0.1./` beside the existing decimal/hex/short-form literal cases.

---

## Verified clean (positive confirmations)

These were checked against the reviewed ref rather than assumed:

- **Bypass surface — no bypass found.** Userinfo (`http://evil.com@127.0.0.1/`), trailing dot, decimal `2130706433`, short form `127.1`, `http://0/`, hex `0x7f.0.0.1`, octal `0177.0.0.1`, and bracketed IPv6 all normalize to a blocked literal before classification. IPv6 zone-id `[fe80::1%25eth0]` fails URL parsing and is refused as invalid.
- **`net.BlockList` unmapping claim — verified empirically** (Node 24, pure compute): with only IPv4 rules plus `::/96` and `64:ff9b::/96` registered, `::ffff:127.0.0.1`, `::ffff:169.254.169.254`, `::7f00:1`, `64:ff9b::7f00:1`, and `64:ff9b::a9fe:a9fe` all return `blocked=true`, while `8.8.8.8` and `::ffff:8.8.8.8` correctly return `false`. The load-bearing empirical claim in the PR description holds.
- **Refusal is terminal on both paths.** `scrape.ts:166` and `:226` re-throw `EgressBlockedError` ahead of any escalation decision, so a refusal degrades neither into the render escalation nor into thin fetched content. The handler converts it to a non-crashing `web_scrape blocked: …` (`web-scrape.ts:227,248,299`), and `clearTimeout` / `removeEventListener` sit in a `finally` (`:341-342`) that covers every early return.
- **Handler ordering claim — verified as described.** `AbortController` (`:199`) → abort listener (`:211`) → timer (`:212`) → guard pre-check (`:225`). The guard is the first `await`, and it sits after the wiring, so a parent abort landing in that window is not dropped.
- **Redirect bound — not an off-by-one.** `for (let hop = 0; hop <= MAX_REDIRECTS; hop++)` with `MAX_REDIRECTS = 20` yields 21 attempts (1 initial + 20 followed), matching the fetch-spec convention.
- **Env opt-out defaults correctly.** `privateHostsAllowed()` is truthy only for `'1'`/`'true'` after trim + lowercase, so unset/`''`/`'false'`/`'0'` all leave the guard **active**.
- **Dispatcher is live in production.** Both `web-scrape.ts:167` and `scrape.ts:112` default `fetchFn` to `globalThis.fetch` with reference identity preserved through the call chain, so the `fetchFn === globalThis.fetch` condition at `egress-guard.ts:305` holds in production. Independently probed: Node 24's global `fetch` does honor `init.dispatcher` (a bogus dispatcher yields `agent.dispatch is not a function`, not `ECONNREFUSED`), so the mechanism is not a silent no-op.
- **Browser-file touches are not scope creep.** `PlaywrightProvider.render()` has exactly one production caller (`src/web/scrape.ts`); the interactive `browser_open` tool routes through the separate `open()` → `ensurePage()`/`ensureContext()` path with no `.route()` wiring. The "`web_scrape` surface only" claim holds.
- **Test count confirmed.** 31 static `it(` sites, 6 of them table-driven loops, resolving to the 62 runtime cases claimed.

## Epistemic scope — what was not checked

- Read against branch HEAD `16a4755`; every `blocking`/`high` citation was verified at that ref. One Wave 1 agent's stream terminated twice mid-run, so the security/api-compat dimension was re-derived directly against the ref instead of by sub-agent.
- **Tests were not executed.** The "`pnpm test` passes, 13,505 tests" and "neutering the guard fails 54 tests" claims are unverified — the mutation figure cannot be checked by static reading. CI is green on all 8 required checks.
- **Stated intent:** PR #750 title + body — spec-compliance assessed.
- Not checked: real-world size of the render-path race window (Chromium/Playwright resolver internals are not in this diff); `pnpm-lock.yaml`, `package.json`, and generated `docs/env-registry.*` were excluded by policy — the new `undici` dependency warrants a human dependency-review look.

PR #751, opened 2026-07-29
fix(repl): drain bootstrap warnings after the startup screen clear (#745)

### LINKED ISSUE #745: repl: bootstrap warnings (agent-registry builtin shadow, MCP) are erased by the startup screen clear
Found while reviewing #739 (merged as ce5ef4e). #739 shipped a warning when an operator agent file displaces a tool-restricted builtin (`research-agent`, `Explore`, `git-investigator`). On the interactive REPL — the dominant human surface — that warning is written and then destroyed before anyone can read it.

## What happens

`loadAgentRegistry` defaults `warn` to a stderr writer, and the REPL's bootstrap omits `warn` entirely, so shadow warnings go to stderr during bootstrap:

- `src/agent/agents/registry.ts:265` — `const warn = options.warn ?? ((message: string) => process.stderr.write(message + '\n'));`
- `src/cli/commands/interactive/bootstrap.ts:323` — `loadAgentRegistry({ ...cwd, pluginAgents })`, no `warn`

Then, later in the same `.action()` callback, the startup clear wipes the viewport **and scrollback**:

- `src/cli/commands/interactive.ts:676` — `process.stdout.write('\x1b[3J\x1b[2J\x1b[H');`

The ordering is unconditional on the success path: `bootstrapSession` is awaited at `interactive.ts:406`, the clear runs at `:676`, same callback, no TTY check (zero `isTTY` references in the file), no flag/env gate, no early return between them.

The repo already documents this hazard and its sanctioned workaround, for a different message class — `src/cli/commands/interactive.ts:35-44`:

> The screen-clear escape sequence (`\x1b[3J\x1b[2J\x1b[H`) at the start of interactive mode wipes everything written before `program.parse()` — including the update-available banner and the "Updated to vX" pending-update message that `index.ts` writes before the parse call. Index.ts calls `setInteractiveUpdateNotices` to stash those notices so the interactive action can re-emit them after the clear, where they will survive.

## Scope: this is a class, not one message

The same erasure swallows MCP config warnings at `src/cli/commands/interactive/bootstrap.ts:523` — `for (const w of loaded.warnings) console.warn('[mcp] ' + w)`. Any fix should cover both producers rather than special-casing the registry.

## Why it matters

The builtin-shadow warning is a safety signal, not cosmetics: a single `~/.afk/agents/research-agent.md` declaring broader `tools:` converts the read-only verifier that nine bundled skills dispatch by bare name into a write-capable agent, for every project on the machine. #739's stated purpose was that this "stops being invisible" — on the REPL it stays invisible.

No recovery path exists today: `/agents` reads `session.supportedAgents()`, which both providers return `[]` for (`src/agent/providers/anthropic-direct/query.ts:724-726`, `src/agent/providers/openai-compatible/query.ts:1165-1167`), and it never reports an agent's `source`.

**Where it does work** (so this is surface-specific, not total): `chat` prints it to stderr with no clear sequence (`src/cli/commands/chat.ts:506`), and under supervision stderr is appended to a log file (`src/service/launchd/plist.ts:76-90`, `src/service/systemd/unit.ts:102-114`).

## Suggested fix

A drain-after-clear channel already exists for exactly this shape and should be reused rather than reinvented:

- `src/cli/slash/plugin-skills/dispatch.ts:278` — `export function getPluginShadowingNoticeLines(): string[]`
- `src/cli/commands/interactive/loop-iteration.ts:123,140,225-228` — drains it and renders via `ctx.replRenderer.writeLine`, i.e. after the clear

Accumulate registry + MCP bootstrap warnings into an equivalent pull-based getter and drain it in the same place.

## Acceptance criteria

- [ ] A builtin-shadow warning emitted during REPL bootstrap is visible to the operator after the startup clear.
- [ ] MCP bootstrap warnings (`bootstrap.ts:523`) travel the same path.
- [ ] `chat`, `daemon`, and `telegram` surfaces keep their current behaviour (no double-emission).
- [ ] A test pins the post-clear emission; `pnpm test` + `pnpm lint` green.

## Verification note

Per `docs/scrollback.md:108-111` and the constraint recorded in #540, the headless suite cannot verify real-PTY scrollback and two prior fixes shipped green while broken in reality. Validate in a real terminal, and lean on the existing `Test (real PTY scrollback)` CI leg.

Related: #540 (in-turn compositor gap class — different mechanism, same rendering domain).

griffinwork40:
Fixes #745.

## The bug

The REPL's startup clear (`\x1b[3J\x1b[2J\x1b[H`, `interactive.ts:676`) runs after `bootstrapSession` is awaited (`:406`) — same `.action()` callback, no early return, no TTY or flag gate between them (`awk` over that range for `return|throw|process.exit` returns zero hits). `\x1b[3J` is **Erase Saved Lines**, so it wipes scrollback as well as the viewport: anything bootstrap wrote straight to stdout/stderr was destroyed and unrecoverable, even by scrolling up.

That silently erased the built-in-shadow warning shipped in #739. It's a safety signal, not cosmetics — a single `~/.afk/agents/research-agent.md` declaring broader `tools:` converts the read-only verifier that nine bundled skills dispatch by bare name into a write-capable agent, for every project on the machine. #739's stated purpose was that this "stops being invisible"; on the REPL it stayed invisible, and `/agents` offers no recovery path (both providers return `[]` from `supportedAgents()`).

## The fix

Producers accumulate into `ctx.bootWarnings`; `interactive.ts` drains it **after** the clear, inside the pre-arm print block so the newlines count into `preArmAnchorRow` and the compositor arms below them. This reuses the drain timing already documented for the update-notice re-emit (`interactive.ts:35-44`), which solves the same problem for pre-`program.parse()` output. No new module-level singleton: bootstrap returns at `:406` and the clear fires at `:676` in the same function, so `ctx` already crosses the only boundary that matters.

Warnings print last in the block — closest to the prompt — because a shadow warning means a read-only agent may now be write-capable.

### Two things worth flagging for review

**1. I did not use the mechanism the issue suggested.** #745 proposed reusing `getPluginShadowingNoticeLines()`. Its drain site is gated behind `isDebugEnabled()` (`loop-iteration.ts:139` — one line above the citation in the issue), so routing safety warnings through it would have made them **debug-mode-only**: strictly worse than the stderr status quo, while satisfying every acceptance criterion in the issue. It's also timing-mismatched, firing async off `waitForInitialization()` after the loop has already started.

**2. The MCP producer had a divergence the issue didn't name.** `bootstrap.ts:523` — the line the issue cites — is only the *no-enabled-servers* branch. On the common path, warnings print inside `McpManager.fromConfig` (`manager.ts:116-118`) via the `warnings:` option. Capturing only the cited line would have fixed the rare branch and missed the common one. Both now route through the single bucket. `chat`, `daemon`, and `telegram` still forward `warnings` into `fromConfig` and are untouched — they never clear the screen — so nothing double-prints.

## Verification

`docs/scrollback.md:108-113` is explicit that no test verifies real scrollback, and that **two prior fixes in this domain shipped green while broken**. So byte-level tests were treated as necessary but not sufficient:

- **`interactive.boot-warnings.test.ts`** (new, 5 tests) — pins write ORDER via `invocationCallOrder` (warning after the clear escape), single-emission, and buffer drain. **Mutation-tested**: reverting just the drain block fails 4 of the 5, so these tests genuinely pin the fix rather than passing either way.
- **`tests/pty/scenarios.ts` → `boot-warning-survives-clear`** (new) — real-pty scenario, auto-collected by the existing loop. Asserts the warning is readable in the emulator buffer **and** that pre-clear writes are `absent` — so the erasure premise is empirically demonstrated on real hardware, not assumed.
- **Real-terminal ground truth** — drove the built CLI inside a genuine pty with a project-local shadowing fixture, replaying into `@xterm/headless`:
  - **With this fix:** warning renders one row above the prompt, exactly once, `\x1b[3J` confirmed present in the stream.
  - **On unmodified `main`, identical setup:** the warning appears **nowhere** in the buffer — viewport or scrollback.

Gates: `pnpm lint` clean, `pnpm test` 13430 passed / 707 files, `pnpm test:coverage` passes thresholds, `pnpm test:pty` 9 passed. Baseline was confirmed green before starting, so no failures are pre-existing.

## Note on the touched test fixtures

`bootWarnings` is a **required** field on `InteractiveCtx` rather than optional — the contract should be explicit, and there's one production constructor. The 5 `interactive-lifecycle.test.ts` ctx fixtures build partial objects cast past the type checker, so they needed `bootWarnings: []` added; without it they throw at the drain site. I chose that over an optional field with a `?.` fallback, which would have hidden the contract and let a future producer silently skip the drain.

## Deliberately out of scope

The erased class is larger than the two producers #745 names — also reachable during the same awaited bootstrap: `bootstrap.ts:959` (path-approval), `default-hook-registry.ts:207,299`, `config-bridge.ts:81`, `permissions-store.ts:152`, `mcp/manager.ts:136,235`, `plugins/tool-injector.ts:327-336`. `ctx.bootWarnings` is the seam for those, but converting them all here would widen the blast radius well past a P2 fix.

Separately worth its own issue: `discoverPluginAgents` (`skill-bridge.ts:372`) calls `parseAgentMarkdown` with **no** warn callback, so malformed plugin-agent files are silently swallowed and never reach stderr at all — a different bug from this one (never written vs. written-then-erased).

### DIFF
--- a/src/cli/commands/interactive.boot-warnings.test.ts
+++ b/src/cli/commands/interactive.boot-warnings.test.ts
@@ -14,6 +14,10 @@
  * name into a write-capable agent. The fix routes such producers into
  * `ctx.bootWarnings` and drains them after the clear.
  *
+ * Two exits out of bootstrap must both surface the buffer: the success path
+ * drains after the clear, and the abort path drains in `interactive.ts`'s catch
+ * before `handleCommandError` exits (see the `bootstrap abort path` block).
+ *
  * These tests assert ORDER — that the warning bytes are written AFTER the clear
  * escape — via `mock.invocationCallOrder`, the same technique
  * `interactive-lifecycle.test.ts` uses for the clearScreen/compositor ordering.
@@ -31,6 +35,12 @@ import { Command } from 'commander';
 
 const mockBootstrapSession = vi.fn();
 const mockRunReplLoop = vi.fn();
+// Stands in for the real `never`-returning handler, which calls process.exit.
+// Throwing instead makes the abort path observable AND preserves the property
+// the drain ordering depends on: nothing after this call runs.
+const mockHandleCommandError = vi.fn((err: unknown): never => {
+  throw err instanceof Error ? err : new Error(String(err));
+});
 
 // ---------------------------------------------------------------------------
 // Mocks — hoisted above imports. Mirrors interactive.resume.test.ts's thin
@@ -91,9 +101,7 @@ vi.mock('../update-checker.js', () => ({
 }));
 
 vi.mock('../errors/index.js', () => ({
-  handleCommandError: vi.fn((err: unknown): never => {
-    throw err instanceof Error ? err : new Error(String(err));
-  }),
+  get handleCommandError() { return mockHandleCommandError; },
 }));
 
 vi.mock('ora', () => ({
@@ -160,7 +168,10 @@ function makeCtx(bootWarnings: string[]): Record<string, unknown> {
  * evaluated across the two together — exactly why `interactive.ts` wraps both
  * when measuring `preArmAnchorRow`.
  */
-async function runAndCapture(bootWarnings: string[]): Promise<string[]> {
+async function runAndCapture(
+  bootWarnings: string[],
+  opts?: { failBootstrap?: Error },
+): Promise<string[]> {
   // `interactive.ts` installs SIGINT/SIGTERM/SIGHUP handlers before the clear
   // and unregisters them only via its cleanup path, which never runs here (we
   // short-circuit at runReplLoop). Snapshot the pre-existing listeners so the
@@ -183,7 +194,22 @@ async function runAndCapture(bootWarnings: string[]): Promise<string[]> {
     writes.push(a.map(String).join(' '));
   });
 
-  mockBootstrapSession.mockResolvedValue(makeCtx(bootWarnings));
+  if (opts?.failBootstrap !== undefined) {
+    // Reproduces the real abort shape: producers push into the CALLER-owned
+    // bucket (`extras.bootWarnings`), and only then does bootstrap throw — e.g.
+    // an MCP config warning collected just before `McpManager.fromConfig`
+    // rejects on an `alwaysLoad` server. No ctx is ever returned, so the
+    // post-clear drain is unreachable by construction.
+    const failure = opts.failBootstrap;
+    mockBootstrapSession.mockImplementation(
+      async (_options: unknown, extras?: { bootWarnings?: string[] }) => {
+        for (const w of bootWarnings) extras?.bootWarnings?.push(w);
+        throw failure;
+      },
+    );
+  } else {
+    mockBootstrapSession.mockResolvedValue(makeCtx(bootWarnings));
+  }
   // Stop the boot chain right after the pre-arm print block.
   mockRunReplLoop.mockRejectedValue(new Error('test-shortcircuit'));
 
@@ -218,6 +244,8 @@ describe('afk interactive — bootstrap warnings survive the startup clear (#745
   beforeEach(() => {
     mockBootstrapSession.mockReset();
     mockRunReplLoop.mockReset();
+    // `mockClear`, not `mockReset` — the throwing implementation must survive.
+    mockHandleCommandError.mockClear();
   });
 
   afterEach(() => {
@@ -276,5 +304,52 @@ describe('afk interactive — bootstrap warnings survive the startup clear (#745
     // interactive.ts empties the array in place after printing; the REPL loop
     // and any later reader must see nothing left to emit.
     expect(bootWarnings).toHaveLength(0);
+  });
+
+  /**
+   * Regression found in review of PR #751: the post-clear drain sits ~360 lines
+   * after the `bootstrapSession` await. When bootstrap THROWS, `ctx` is never
+   * assigned and `handleCommandError` exits the process, so every warning
+   * buffered before the throw was destroyed — silently, with no clear to blame.
+   *
+   * That was a strict regression, not an inherited gap: before #751 these
+   * producers wrote to stderr the moment they fired, and the abort path never
+   * reaches the clear, so the diagnostic stayed on screen next to the failure
+   * it explained. The fix hoists the buffer into `interactive.ts` and drains it
+   * in the catch, which is why these tests exercise the CALLER-owned bucket.
+   */
+  describe('bootstrap abort path', () => {
+    it('emits warnings buffered before a bootstrap failure', async () => {
+      const writes = await runAndCapture(
+        ['[mcp] server "foo": unknown key "cmd"'],
+        { failBootstrap: new Error('mcp server "foo" (alwaysLoad) failed to connect') },
+      );
+
+      // `handleCommandError` is mocked to throw, mirroring the real `never`
+      // return: if these bytes are present at all, they were written BEFORE the
+      // process would have exited. Presence is the ordering proof on this path.
+      expect(writes.some((w) => w.includes('unknown key "cmd"'))).toBe(true);
+      // Proves we actually went down the abort path rather than succeeding.
+      expect(mockHandleCommandError).toHaveBeenCalledTimes(1);
+    });
+
+    it('never reaches the screen clear, so immediate emission is safe', async () => {
+      const writes = await runAndCapture(['[mcp] pre-throw'], {
+        failBootstrap: new Error('boom'),
+      });
+      // `\x1b[3J` is the only reason buffering was needed in the first place.
+      // It does not run here, so nothing erases the drained warning — and no
+      // second drain exists on this path to double-print it.
+      expect(writes.some((w) => w.includes(CLEAR))).toBe(false);
+      expect(writes.filter((w) => w.includes('pre-throw'))).toHaveLength(1);
+    });
+
+    it('prints nothing when bootstrap fails before any producer ran', async () => {
+      const writes = await runAndCapture([], { failBootstrap: new Error('bad --model') });
+      // Absence of noise: the drain must add no header or blank line of its own
+      // to the far more common "bad flag" failure.
+      expect(writes.some((w) => w.includes('[mcp]'))).toBe(false);
+      expect(mockHandleCommandError).toHaveBeenCalledTimes(1);
+    });
   });
 });
--- a/src/cli/commands/interactive.boot-warnings.test.ts
+++ b/src/cli/commands/interactive.boot-warnings.test.ts
@@ -1,0 +1,280 @@
+/**
+ * Regression: bootstrap warnings must survive the interactive startup screen
+ * clear (#745).
+ *
+ * The startup clear is `\x1b[3J\x1b[2J\x1b[H`. `\x1b[3J` is Erase Saved Lines
+ * — it wipes the terminal's SCROLLBACK, not just the viewport — so anything
+ * written before it is unrecoverable, even by scrolling up. `bootstrapSession`
+ * is awaited BEFORE that clear in the same `.action()` callback, so any producer
+ * inside bootstrap that printed straight to stdout/stderr was silently erased.
+ *
+ * That erased class included the agent-registry built-in-shadow warning, which
+ * is a safety signal: a `~/.afk/agents/research-agent.md` declaring broader
+ * `tools:` converts the read-only verifier that bundled skills dispatch by bare
+ * name into a write-capable agent. The fix routes such producers into
+ * `ctx.bootWarnings` and drains them after the clear.
+ *
+ * These tests assert ORDER — that the warning bytes are written AFTER the clear
+ * escape — via `mock.invocationCallOrder`, the same technique
+ * `interactive-lifecycle.test.ts` uses for the clearScreen/compositor ordering.
+ * Per `docs/scrollback.md`, mock-stdout tests can prove bytes were WRITTEN and
+ * in what order, but cannot prove they reached a real terminal's scrollback;
+ * the PTY scenario (`tests/pty/scenarios.ts`) covers visibility.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { Command } from 'commander';
+
+// ---------------------------------------------------------------------------
+// Mutable mocks controlled per test
+// ---------------------------------------------------------------------------
+
+const mockBootstrapSession = vi.fn();
+const mockRunReplLoop = vi.fn();
+
+// ---------------------------------------------------------------------------
+// Mocks — hoisted above imports. Mirrors interactive.resume.test.ts's thin
+// surface: we need the real `interactive.ts` action to run from bootstrap
+// through the clear and the pre-arm print block, then stop before the REPL.
+// ---------------------------------------------------------------------------
+
+vi.mock('./interactive/bootstrap.js', () => ({
+  get bootstrapSession() { return mockBootstrapSession; },
+}));
+
+vi.mock('./interactive/repl-loop.js', () => ({
+  get runReplLoop() { return mockRunReplLoop; },
+}));
+
+vi.mock('./interactive/boot-prune.js', () => ({
+  bootPruneWorktrees: vi.fn().mockResolvedValue({ ran: false, removedCount: 0 }),
+}));
+
+vi.mock('./interactive/worktree.js', () => ({
+  setupWorktree: vi.fn(),
+}));
+
+vi.mock('./interactive/worktree-autoname.js', () => ({
+  runFirstTurnAutoname: vi.fn(),
+}));
+
+vi.mock('../resume-session.js', () => ({
+  resolveResumeTarget: vi.fn(() => undefined),
+  resumeConfigFor: vi.fn(() => ({})),
+  saveCurrentSession: vi.fn(),
+}));
+
+vi.mock('../shared-helpers.js', () => ({
+  getApiKey: vi.fn(() => 'test-key'),
+  getModel: vi.fn(() => 'sonnet'),
+  getThinking: vi.fn(() => undefined),
+  getEffort: vi.fn(() => undefined),
+  getMaxOutputTokens: vi.fn(() => undefined),
+  getMaxToolUseIterations: vi.fn(() => undefined),
+  loadSystemPrompt: vi.fn(() => undefined),
+  loadConfigSystemPrompt: vi.fn(() => undefined),
+  resolveBaseSystemPrompt: vi.fn(() => ({ prompt: undefined, source: 'none' })),
+  parseThinking: vi.fn(() => undefined),
+  parseEffort: vi.fn(() => undefined),
+  parseMaxOutputTokens: vi.fn(() => undefined),
+  parseProvider: vi.fn(() => undefined),
+  parseThinkingUiMode: vi.fn(() => 'live'),
+  getDefaultSubagentModel: vi.fn(() => 'sonnet'),
+}));
+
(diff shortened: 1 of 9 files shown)


griffinwork40:
# Code Review — PR #751

**Target:** PR #751 `fix(repl): drain bootstrap warnings after the startup screen clear (#745)` · base `main` · head `fix/bootstrap-warning-scrollback-order`
**Reviewed ref:** `a2149fb43895017535d0fd0dc4b3441c68b4b2a7` (checked out locally, clean — all citations read at this ref)
**Change type:** hotfix · **Regime:** light (61 lines production / 368 lines tests, 6 files)
**Stated intent:** PR #751 title + body — spec-compliance assessed against it.

---

## Findings

### 1. `medium` · confidence high · correctness/spec-compliance · regression
`src/cli/commands/interactive.ts:404-410` + `src/cli/commands/interactive/bootstrap.ts:339,514` · ref:`a2149fb` · citation-type: `file-state`

**Finding:** When `bootstrapSession` throws after populating `bootWarnings`, the warnings are silently destroyed — and this is a strict regression, not an inherited gap: before this PR the same warning reached stderr immediately and *survived* on the abort path, because the screen clear (`:676`) never runs when bootstrap aborts.

Mechanism, verified end-to-end at both refs:
- `registry.ts:238` — pre-PR default: `const warn = options.warn ?? ((message: string) => process.stderr.write(message + '\n'));` and `ce5ef4e:bootstrap.ts:323-326` passed **no** `warn` → the shadow warning printed immediately.
- Post-PR it only prints at the drain (`interactive.ts:772`), which sits ~360 lines after the throw site.
- `bootstrap.ts:514` pushes MCP warnings **before** `McpManager.fromConfig` at `:535`; that call is in a bare `try/finally` with no `catch`.
- `manager.ts:230-234` throws for an `alwaysLoad: true` server that fails to connect; `:255-262` propagates it out of `bootstrapSession`.
- `interactive.ts:407-409` catches → `handleCommandError(err)` → `errors/index.ts:21-25` `process.exit(...)`, typed `never`. Clear and drain never execute.

**Evidence** (`interactive.ts:404-410`, read at a2149fb):
```ts
      let ctx: InteractiveCtx;
      try {
        ctx = await bootstrapSession(options, worktreeCwd !== undefined ? { cwd: worktreeCwd } : undefined);
      } catch (err) {
        spinner.fail('Invalid options');
        handleCommandError(err);
```
**Tier rationale (stated so you can disagree with the tier without disagreeing with the facts):** rubric-literal this is "wrong output under reachable conditions" = `high`; downgraded to `medium` because the aborted session never starts, so the shadowed read-only agent cannot be dispatched in that run, and the operator gets a hard classified error that prompts an immediate re-run where the warning does appear. Second, lower-probability route: `initTranscript()` (`interactive.ts:540`, pre-clear, no local try/catch).

**Suggestion:** hoist the array — declare `const bootWarnings: string[] = []` in `interactive.ts` and pass it into `bootstrapSession`, so the `catch` can drain it before `handleCommandError(err)`. (Draining `ctx.bootWarnings` in the catch is impossible as written: `ctx` is unassigned there.)

### 2. `low` · confidence high · perf-observability
`src/cli/commands/interactive.ts:772-777` · ref:`a2149fb` · citation-type: `diff-context`
**Finding:** Drained warnings are terminal-only and reach no durable sink, so `afk trace show` cannot answer "did this session warn about a shadowed builtin", and the signal is gone once scrollback rolls.
**Evidence:** `git grep -n 'bootWarnings' a2149fb -- src/agent` → zero matches (trace + transcript writers live there); the neighbouring `emitSessionPhase` calls (`bootstrap.ts:525,539`) carry only `{ serverCount }`.
**Suggestion:** emit one trace event alongside the drain carrying the warning strings.

### 3. `nit` · confidence high · maintainability
`src/cli/commands/interactive/loop-iteration.test.ts:128-160`, `src/cli/commands/interactive/repl-loop-wiring.test.ts:119-156` · ref:`a2149fb` · citation-type: `file-state`
**Finding:** Two test-only ctx factories cast past the checker (`as unknown as InteractiveCtx`) and omit the now-required `bootWarnings`; both feed `runReplLoop` directly and never reach the drain, so no runtime throw — type-fidelity gap only. **Evidence:** repo-wide `git grep -n 'bootWarnings' a2149fb -- src tests` returns zero hits in either file. **Suggestion:** add `bootWarnings: []` to both base objects.

### 4. `nit` · confidence high · repo-convention (pre-existing, NOT an intent violation)
**Finding:** All three touched sources exceed the 350-LOC house rule and this PR grows each: `interactive.ts` 890→906, `bootstrap.ts` 1003→1031, `shared.ts` 838→851. No mechanical gate exists in this repo (`lint` = `tsc --noEmit`; no size check in `.github/workflows/ci.yml`). **Suggestion:** out of scope for a hotfix; the drain block is a candidate seam for an `interactive/pre-arm-block.ts` extraction later.

---

## Per-dimension

| Dimension | Result |
|---|---|
| correctness | Finding 1 (abort-path regression). Core mechanism is sound — see verified claims below. |
| spec-compliance | No unmet intent clause; no scope creep. Every named out-of-scope producer confirmed still on raw `console.warn` (`permissions-store.ts:152`, `default-hook-registry.ts:299`, `manager.ts:235`) — the disclosure is accurate. |
| security | no issues found — read `bootstrap.ts`, `interactive.ts`, `registry.ts` at a2149fb. The change moves a safety warning closer to the prompt; no secret, auth, or injection surface touched. |
| api-compat | `bootWarnings` required on `InteractiveCtx` — one production constructor (`bootstrap.ts:888`); all 5 fixtures updated (`interactive-lifecycle.test.ts:385,545,718,824,949`). No production importer breaks. |
| test-coverage | Findings 3; author's "4 of 5 fail on revert" is accurate — the 5th ("prints nothing extra when bootstrap produced no warnings") passes either way by construction, and the author disclosed it. |
| perf-observability | Finding 2. |

## Verification manifest (Wave 1.5)

**Citations — all `verified` at `a2149fb`** (I re-read each rather than trusting the agents):
`interactive.ts:707-708` both streams wrapped · `:676` clear · `:772-777` drain-last + array reset · `:404-410` catch → exit · `errors/index.ts:21-25` `process.exit` · `manager.ts:230-234,255-262` alwaysLoad throw · `bootstrap.ts:514` push precedes `:535` call · `:535-537` no `warnings:` forwarded · `registry.ts:238` default stderr writer.

**Absence claims — `confirmed-absent`:** sole reader of `bootWarnings` is the drain; no producer pushes post-drain; no durable sink; no mechanical LOC gate.

**One subagent claim corrected by re-derivation:** "the new pty scenario is excluded from the default run" was framed as a coverage gap — verified and **downgraded to a non-finding**: `.github/workflows/ci.yml:221` runs `pnpm test:pty`, so `boot-warning-survives-clear` **is** in the CI gate (`vitest.config.ts:37` excludes it from local `pnpm test` only). Given `docs/scrollback.md`'s note that two prior fixes in this domain shipped green while broken, that distinction mattered.

**Claims the author made that independently hold:** the drained newlines *are* counted into `preArmAnchorRow` — `interactive.ts:707-708` wraps **both** `stdout` and `stderr` (`console.warn` → stderr → counted), and the `finally` restore at `:779-781` fires after the drain; dropping `warnings:` from `fromConfig` loses nothing — `manager.ts:116-118` is print-only, and `chat.ts:648`, `telegram/mcp-session.ts:56`, `scheduler.ts:586` all still pass it, so no surface double-prints.

## What was not checked
- Reviewed against branch HEAD `a2149fb`; citations verified at that ref. No patch-file input, so no citation gaps.
- Did not run `pnpm test`, `pnpm test:pty`, `pnpm lint`, or the built CLI — the author's gate results (13430 passed / 707 files, 9 pty) are unverified here.
- Did not read `repl-loop.ts` / `surface-setup.ts` to confirm `armCompositor` consumes `preArmAnchorRow` as documented (outside the diff).
- No visual/real-terminal confirmation of scrollback survival — source-level reasoning only.
- Did not review `CHANGELOG.md` / `package.json` (absent from the PR diff; they appear only in the local branch range via the interleaved v5.82.0 release commit).
- Wave 2 synthesis was done inline rather than dispatched: two Wave-1 agents with non-overlapping scopes and zero cross-agent conflicts left nothing to dedup. `/shadow-verify` not invoked — its trigger is a `critical`/`high` finding, and none survived; the two load-bearing mechanism claims were re-derived by hand above instead.

## Merge decision: **MERGE**

No blocking findings. The fix is real, the mechanism claims in the PR body hold under independent re-derivation, and the disclosure of out-of-scope producers is accurate. Finding 1 is the one item worth acting on — a ~3-line change (thread the array in from `interactive.ts` so the `catch` can drain it) either pre-merge or as a follow-up issue; it narrows to an abort path where no session runs.

---

**Done**
- Ran `/review 751 --light`: resolved PR #751 to ref `a2149fb`, dispatched Wave 1 (mechanism prober + 2 narrowed reviewers after the first light-reviewer stream was cut off mid-run), then verified every load-bearing citation and absence claim myself at the reviewed ref.
- Evidence: findings above cite `file:line` at `a2149fb`, each read via `git show a2149fb:<path>`; PR body and diff captured at `/tmp/pr751.diff`; the PR itself is at https://github.com/griffinwork40/agent-afk/pull/751.
- What changed in the world: nothing — this skill is read-only. No files edited, no commits, no pushes, no PR comments or review posted. The review exists only in this transcript; if it should reach the PR, that's a separate explicit action.
- Deferred: Finding 1 (abort-path warning loss) is reported, not fixed — fixing it is out of `/review` scope. Same for Finding 2 (no durable sink) and Finding 3 (two test fixtures). Gates (`pnpm test`, `pnpm lint`, `pnpm test:pty`) were not re-run; the author's green results are taken as unverified.
- One subagent finding was refuted by re-derivation (pty scenario *is* in CI at `ci.yml:221`) and dropped before it reached the decision.

---

griffinwork40 (commented):

## Merge decision: **MERGE**

Reviewed at branch HEAD `c107e800` (both commits). Dimensions: security, api-compat, correctness, spec-compliance, test-coverage, perf-observability. No blocking findings. CI is fully green at this ref (Lint & Build, Test ubuntu + macos, Test real PTY scrollback, Docs, Publish Bundle Smoke; windows skipped).

The fix is sound and the mechanism is the right one. `drainBootWarnings` empties the buffer in place (`boot-warnings.ts:29`), which is what makes the success-path drain (`interactive.ts:793`) and the abort-path drain (`interactive.ts:427`) mutually exclusive without either knowing about the other — a genuinely better design than a `didDrain` flag. The PR body's self-flagged deviations from #745 both check out: `getPluginShadowingNoticeLines()` is indeed debug-gated, and `bootstrap.ts:522` does cover both MCP branches where the cited line covered only one.

### Spec-compliance — all 7 checkable claims MET

|---|---|---|
| Drains after clear, inside pre-arm block; newlines count into `preArmAnchorRow` | MET | clear `interactive.ts:695`; drain `:793`; block wraps **both** streams `:726-727`, and the drain writes via `console.warn`→stderr, so it is counted |
| Warnings print last in the block | MET | `:793` drain, then `:794` `console.log()`; banner `:750`, prune `:765`, resume `:773` all precede |
| Single emission, no double-print | MET | `bootstrap.ts:543` omits `warnings:` from `fromConfig`; `chat.ts:648`, `telegram/mcp-session.ts:56`, `daemon/scheduler.ts:586` still forward it and are untouched by this diff |
| `bootWarnings` required on `InteractiveCtx` | MET | `shared.ts:389` non-optional; `extras?.bootWarnings` optional at `bootstrap.ts:158` keeps non-REPL callers working |
| Tests pin write ORDER, not presence | MET | `interactive.boot-warnings.test.ts:266-272` compares `warnIdx > clearIdx` by index, and asserts content survives intact |
| PTY scenario asserts readable + pre-clear `absent` | MET | `tests/pty/scenarios.ts` — `absent: [PRECLEAR_LINE_0, PRECLEAR_LINE_29]`, `inViewport: [SHADOWWARN, MCPWARN]`, plus `exactlyOnce` and `order` |
| Commit 2 drains on abort without double-draining | MET | `interactive.ts:425-428` `spinner.fail()` → drain → `handleCommandError`; in-place empty makes a second drain a no-op |

No scope creep found. The 145-line churn in `interactive-lifecycle.test.ts` is a module-scope hoist of `applyCommonMocks()` plus a new bucket-identity regression test (`expect(ctx.bootWarnings).toBe(bootWarnings)`), which is directly load-bearing for the abort path — not drive-by refactoring.

### Findings

**1. `low` · security/observability · a hard kill during `bootstrapSession` now loses buffered warnings entirely**
`src/cli/commands/interactive.ts:413` → `:793` (ref `c107e800`, file-state)

Producers push into the bucket during `bootstrapSession` (`bootstrap.ts:347`, `:522`), but the SIGINT/SIGTERM/SIGHUP handlers are registered later, at `:638`/`:667`/`:692`. During the whole production window no handler is installed, so a signal there hits Node's default disposition and terminates immediately — the buffered strings are discarded unread. Pre-PR those bytes had already reached stderr, so a same-timed kill left them visible (the clear at `:695` had not run yet). That sub-window is a real, if narrow, regression.

Scoped honestly, because an earlier draft of this finding overstated it and was corrected on re-derivation:
- The "signal between `:638` and `:793`" variant is **not** a defect. There is no `await` anywhere in `600-800` (nearest are `:559` and `:831`), so that span is one uninterrupted synchronous block; a queued signal callback cannot run until after `:793` has already drained, and the in-place empty makes the late handler a no-op.
- `process.once('exit', ...)` is **not** the fix — `exit` listeners do not run on a default-disposition signal kill, so it would miss the only broken window, and it would print after compositor teardown, contradicting the `preArmAnchorRow` accounting this PR is built around.

Suggestion: mirror each warning into the witness trace at push time (`trace` already exists at `bootstrap.ts:223`, well before the `warn` wiring at `:347`), so a hard kill leaves a durable record even when nothing reaches the terminal. Reasonable to defer to a follow-up issue — it is strictly narrower than the bug being fixed here.

**2. `nit` · api-compat · required `bootWarnings` field has test-fixture-only blast radius** — `src/cli/commands/interactive/shared.ts:389`. Exactly one production caller of `bootstrapSession` exists (`interactive.ts:413`) and it is updated in this PR; every other affected site is a test fixture. `[UNVERIFIED: no production importers]` The PR's reasoning for required-over-optional is the right call — a `?.` fallback would let a future producer silently skip the drain.

**3. `nit` · test-coverage · the PTY scenario simulates the byte sequence rather than booting the CLI** — `tests/pty/scenarios.ts`. It writes the clear and the post-clear prints directly, so it proves the emulator/scrollback semantics and the compositor-anchor interaction, but not that production emits that sequence. The order-pinning unit tests plus the manually-driven real-terminal check described in the PR body cover the gap; worth noting given `docs/scrollback.md`'s warning that two prior fixes here shipped green while broken.

### Not new / not findings
- **ANSI injection via warning text** — `console.warn(palette.warning('  ' + w))` at `boot-warnings.ts:28` is byte-identical to the pre-PR call it replaces, and no sanitizer applied before or after. Unchanged property, not introduced here.
- **Secret exposure** — buffering changes timing, not destination; same stderr sink, no new persistence.
- **Safety-signal weakening** — same palette, same indent, same content; nothing truncated or downgraded.

### What was not checked
Read against branch HEAD `c107e800`; all blocking-tier citations re-read at that ref. Stated intent: PR #751 title + body — spec-compliance assessed against it. Did not execute the test suite (relied on green CI at this ref) and did not reproduce the signal race at runtime — finding 1's mechanism is confirmed from source, its real-world frequency is not. Did not inspect `loadMcpConfig` warning-string construction or `loadAgentRegistry`'s default warn writer. Read-only review: no files, branches, or PR metadata were modified.

Issue #740, opened 2026-07-28
Bash grant filter: should resolveBase containment drop builtin credential roots, or only explicit /allow-dir grants?


griffinwork40:
## The question

`deriveRestrictedSubstrings` (`src/agent/tools/hooks/bash-restriction-hook.ts`) filters the bash surface's restricted-root list against the session's grants. The `granted` set is seeded from **three** sources:

```ts
const granted = new Set([
  ...(grants.resolveBase !== undefined ? [grants.resolveBase] : []),
  ...grants.readRoots,
  ...grants.writeRoots,
]);
```

Containment is **ancestor-based**, not exact-match — a candidate is dropped when `path.relative(granted, candidate)` does not start with `..`.

Two of those three sources are explicit user acts (`/allow-dir` → `readRoots`/`writeRoots`). The third, `resolveBase`, is **not**: it is the session's cwd anchor (`dispatcher.ts:330` — `this.resolveBase = opts.cwd`, surfaced via `getProtectedRoot()`).

So: **should `resolveBase` containment be able to drop a _builtin credential_ root, or should that be reserved for explicit grants?**

Today it can. A session whose `resolveBase` is an ancestor of a credential root silently loses that root from the bash restriction list, with no `/allow-dir` call anywhere.

## Why this is a contract question, not a bug

The current behavior is deliberate and documented — the module's own `Invariant:` comment defends it:

> Bash's gate exists to stop the ACCIDENTAL `cat`, and an explicit `/allow-dir ~/.ssh` is the user saying they want that path in this session; the typed tools stay floored regardless, so the strict boundary is never the one being relaxed here.

The rationale is sound ("a user working in `~/.ssh` shouldn't self-block"), but the comment frames the divergence as *explicit-grant-only*, and `resolveBase` is the implicit third path it doesn't mention. (That comment was amended in #734 follow-up work to name `resolveBase`, so the code and docs now agree on what happens — this issue is about whether what happens is what we want.)

Both readings are defensible:

| Option | Effect | Cost |
|---|---|---|
| **A — status quo** | `resolveBase` keeps dropping roots beneath it | A session anchored above a credential root loses bash protection for it implicitly |
[...]

Issue #736, opened 2026-07-28
Read denylist: case-variant spellings bypass the credential floor on case-insensitive filesystems


griffinwork40:
Split out of the review of #734, where it was deliberately **not** fixed to keep that PR's blast
radius small.

## Symptom

On a case-insensitive filesystem (macOS APFS by default), a case-variant spelling of a protected
directory is readable on **both** the typed-tool surface and the bash surface:

```
allow bash | allow typed | ~/.SSH/id_rsa
allow bash | allow typed | ~/.Afk/Config/afk.env
```

The filesystem happily opens `~/.SSH/id_rsa` as `~/.ssh/id_rsa`, so this is a real read of a real
credential, not a miss on a non-existent path.

## Cause

Both halves compare case-sensitively:

- `isReadDenied()` (`src/agent/tools/handlers/read-denylist.ts`) does
  `real === blocked || real.startsWith(blocked + '/')`. macOS `realpath(3)` does **not** canonicalize
  case, so the requested spelling survives into the comparison.
- The bash-restriction hook (`src/agent/tools/hooks/bash-restriction-hook.ts`) does a literal
  `scanned.includes(sub)`.

## Why it wasn't fixed in #734

It is **parity-preserving** — both halves fail identically, so it is not a gap that PR opened, and
that PR's claim was specifically about typed/bash parity. Fixing it means changing the comparison
semantics of the unconditional credential floor, which is a wider change than a parity PR should
carry.

## Suggested fix

Case-fold the comparison **only where the underlying filesystem is case-insensitive**, rather than
unconditionally — a blanket `toLowerCase()` would be wrong on a case-sensitive volume, where
`~/.SSH` and `~/.ssh` are genuinely different directories and the former deserves no protection.

Probing the actual volume is the honest approach (e.g. compare `fs.statSync` inode/dev for two
case-variant spellings of a known-existing path, cached per process). Applies to the write denylist
(`write-denylist.ts`) too, which has the same comparison shape.

## Reproduction

A standing probe harness lives outside the repo at
`~/.afk/workspace/reports/pr-734-parity-probe.mts` (section B); run with
`npx tsx ~/.afk/workspace/reports/pr-734-parity-probe.mts` from the repo root. It prints a
bash-vs-typed verdict table for each case and is the artifact the #734 review findings came from.

PR #753, opened 2026-07-29
fix(permissions): follow AFK_HOME in the credential floor + reject overly broad subagent cwd/writeRoots

griffinwork40:
## Summary

Layered fixes to the credential floor, all found while auditing how `AFK_HOME` relocation interacts with the permission denylists.

The headline gap: `BUILTIN_READ_DENYLIST` and `BUILTIN_WRITE_DENYLIST` hardcoded `${homedir()}/.afk/config`, but this project supports relocating its home via `AFK_HOME` (`src/paths.ts` → `getAfkHome()`). An operator running `AFK_HOME=/opt/my-afk` got **no protection at all** on `/opt/my-afk/config/afk.env` (API keys) — not on the typed read tools, not on the write tools, and not on bash. The floor only ever covered the default `~/.afk` location.

## Commits

1. **`fix(subagent): reject overly broad cwd and writeRoots, matching readRoots`** — `cwd`, `writeRoots`, and `readRoots` all land in a forked child's grant snapshot and all three feed the bash-restriction hook's grant filter, where an ancestor-based containment check drops credential roots beneath a granted path. Only `readRoots` carried the breadth rejection, so `cwd: "$HOME"` silently emptied the home-anchored credential floor for that child's bash surface. Hoists the guard into a shared `isTooBroadRoot()` applied to all three fields.

2. **`fix(permissions): cover the AFK_HOME-relocated credential tree on read, write, and bash`** — derives the relocated spelling inside `resolveLists()` (not at module scope, which would evaluate `getAfkHome()` once at import and never observe a runtime change), adds `AFK_HOME` to the memoization cache key, and fails safe when `AFK_HOME` is malformed.

3. **`fix(permissions): make the mcp.json carve-out follow AFK_HOME too`** — commit 2 moved the deny side but left `BUILTIN_READ_ALLOWLIST` anchored to `homedir()`, which *inverted* the carve-out for relocated operators. Derives the relocated spelling of every `READ_ALLOWLIST_REL` entry, failing **closed** on the allow side.

4. **`fix(permissions): normalize relocated AFK_HOME in bash guard`** — the bash surface is lexical, not realpath-resolved, so it needed its own change: `getReadDenylist()` returns only the symlink-*resolved* spelling, and `$AFK_HOME` was never expanded at all.

5. **`fix(permissions): close the double-separator bash bypass and follow AFK_STATE_DIR`** — review response.

6. **`fix(permissions): fold dot-segments and braced ${VAR} into the bash normal form`** — review response; corrects an overstated claim in commit 5.

7. **`fix(subagent): reject a granted root at or above the AFK home`** — review response.

### The bash bypass is positional, not character-specific

Commit 5 fixed a **trailing-separator `AFK_HOME`** failing OPEN: `$AFK_HOME` was substituted into the scanned command verbatim, producing an interior `//`, while the needle is built with `path.join`, which collapses it — and the final match is a literal `includes()`. The two spellings never met.

**Commit 5 claimed to close this "as a class". That claim was wrong, and commit 6 corrects it.** An adversarial review wave flagged two siblings and an empirical probe against the built hook confirmed both:

| spelling | before c6 |
|---|---|
| `<home>/.afk//config/afk.env` | blocked (fixed in c5) |
| `<home>/.afk/./config/afk.env` | **BYPASS** |
| `<home>/.afk/../.afk/config/afk.env` | **BYPASS** |
| `<home>/.afk/config/./afk.env` | blocked — dot lands *after* the needle |

The distinguishing factor is not which character is used but **where it lands**: a segment that splits the span the needle covers breaks the match; the same characters after the needle are harmless. That is why the `//` report alone under-specified the class. Commit 6 replaces the narrow `//` collapse with `path.posix.normalize` over every path-like span, folding `//`, `/./` and `/../` into the one normal form the needles already use.

Commit 6 also substitutes the **braced** `${AFK_HOME}` / `${HOME}` spellings, which were never expanded — so a braced reference reached a credential file unblocked. The braced form is the ordinary spelling a non-adversarial model emits, placing it inside this hook's accidental-access threat model, unlike the runtime variable assembly (`H=$HOME; …$H/…`) the module header rules out and the "documented bypass" test still pins (unchanged, still green). **This gap was pre-existing and applied to plain `$HOME` too**, not only to the `AFK_HOME` work here.

### `AFK_STATE_DIR` (commit 5)

The derived write floor followed `AFK_HOME` but not `AFK_STATE_DIR`, which relocates the entire state tier *independently* (`paths.ts` returns it verbatim when set). An operator running `AFK_STATE_DIR=/opt/state` had zero write protection on their real state tier. The two derivations now sit in **separate `try` blocks** so one malformed env var cannot drop the other's entries.

### Silent fail-safe (commit 5)

All four `AFK_HOME` derivation sites swallowed their throw silently. The fail-safe *direction* was correct; its **invisibility** was the defect. New `src/agent/tools/afk-home-warn.ts` emits one warn line **per process** (latched — these sit on the read/write hot path, so a per-call warn would be its own denial of service). Control flow unchanged.

### The breadth guard missed the AFK home (commit 7)

The one-hop restatement of commit 1. That commit rejected a too-broad `cwd`/`writeRoots` because such a grant empties the child's **home**-anchored floor. But the targets were home-only, and a relocated AFK home is neither the home dir nor an ancestor of it — so granting it as `cwd` sailed through and emptied the **AFK**-anchored floor by exactly the same route. Fixing the deny side while leaving that open made this PR's headline claim false on the subagent surface. Targets are derived **per call**, because unlike `homedir()` they are env-driven and are precisely what a runtime relocation changes. Descendants stay legal.

## Why `state` is in the write denylist but not the read denylist

A model must never *write* to the state tier, but forked sub-agents legitimately *read* skill-preflight inputs, todos, and transcripts from it (#544 / #547 / #554). The derived entries mirror that asymmetry.

## Behavior change worth calling out in release notes

`parseAgentInput` now **rejects** `cwd` / `writeRoots` values it previously accepted: a filesystem root, the home dir, an ancestor of home, and (as of commit 7) the AFK home / state dir or an ancestor of either. This is the point of commits 1 and 7, but it is a user-visible contract change — a workflow dispatching with the home dir as `cwd` (including a monorepo checkout that *is* the home dir) will now throw.

## Relationship to other work

- **Composes with #735.** That PR fixed *which paths* are floored; this one fixes *path resolution* so the floor follows a relocated home. No overlap.
- **Reduces exposure to #740** but does not settle it. The ancestor containment check in `deriveRestrictedSubstrings` is unchanged; this closes the gap that let a model hand it an unvetted root.
- **Does not address #736** (case-variant spellings on case-insensitive filesystems) — deliberately out of scope.

## Known-remaining / deferred (raised in review, deliberately not in this PR)

All tracked as follow-up issues: #778, #779, #780, #781, #782, #783.

- **#778 — `SENSITIVE_PATH_SIGNAL` does not carry the relocated AFK tree.** Under `AFK_FORCE_BASH_INTERPRETER_GUARD=1` on a headless surface, where the grant-filtered substring list is empty and the lexical signal is the sole input, an interpreter one-liner naming a relocated config path is not matched by the signal regex. Reported, not independently confirmed here.
- **#779 — the relocated `mcp.json` carve-out is emitted unconditionally** on the typed path, unlike the bash twin which gates each form on `isReadDenied`. Reachable only when the AFK home is nested *under* another builtin denied root (e.g. `~/.ssh/afk`).
- **#781 — memoizing `getWriteDenylist()`.** Adding a cache to a credential denylist is the exact invalidation bug class this PR already had to fix on the read side — disproportionate risk for a 2-syscall saving both reviews rated not-required-for-merge.
- **#782 — the 350-LOC extractions** in `input-parse.ts` and `bash-restriction-hook.ts`. Both files were over the ceiling *before* this PR; a structural refactor on an under-review security PR is scope creep. The new warn helper was added as its own module rather than growing either file.
- **#780 — the warn latch names only the first-thrown variable**, so an operator who typos both `AFK_HOME` and `AFK_STATE_DIR` is told about one.

## Verification

- `pnpm audit:env:check` — 0 violations.
- `pnpm test` — **706 files, 13453 passed, 14 skipped, 0 failed.**
- The two bypasses fixed in commit 6 were confirmed by an **empirical probe against the built hook** before and after, not by diff reading — the same method that caught the commit-3 regression. Static review had already passed over the dot-segment case.
- Merges cleanly into `main`.

**949 insertions / 44 deletions across 9 files**; 4 of the 9 are test files.

### DIFF
--- a/src/agent/tools/afk-home-warn.ts
+++ b/src/agent/tools/afk-home-warn.ts
@@ -1,0 +1,57 @@
+/**
+ * One-shot operator diagnostic for a malformed `AFK_HOME` / `AFK_STATE_DIR`.
+ *
+ * Every credential-floor derivation site (`read-denylist`, `write-denylist`,
+ * `bash-restriction-hook`) calls `getAfkHome()` / `getAfkStateDir()` inside a
+ * `try` and drops only the derived entry when it throws. That fail-safe
+ * direction is correct — a typo'd env var must never empty the floor, and the
+ * hardcoded `homedir()`-based entries still apply. Its INVISIBILITY was the
+ * defect: an operator who typo'd `AFK_HOME` ran on a silently reduced floor
+ * with no diagnostic anywhere in the process.
+ *
+ * Invariant: the warning is latched to fire at most ONCE per process, not once
+ * per call. These derivations sit on the read/write hot path — `getReadDenylist`
+ * is consulted on every typed read and `getWriteDenylist` on every
+ * `write_file`/`edit_file` — so a per-call warn would emit thousands of
+ * identical lines and become its own denial-of-service on the operator's
+ * terminal. The latch is module-scope state, which is why this lives in its own
+ * module: all four call sites must share ONE latch, and a per-file copy would
+ * warn once per file instead.
+ *
+ * @module agent/tools/afk-home-warn
+ */
+
+let warned = false;
+
+/**
+ * Emit a single warn-level line naming the rejected env value, then latch.
+ *
+ * Contract: never throws (a diagnostic must not become the failure it reports),
+ * never alters control flow, and returns void. The rejected value is carried in
+ * the thrown message from `paths.ts` (`... got: <value>`), so it is surfaced
+ * without this module re-reading the environment.
+ *
+ * @param err The error thrown by `getAfkHome()` / `getAfkStateDir()`.
+ */
+export function warnAfkHomeRejectedOnce(err: unknown): void {
+  if (warned) return;
+  warned = true;
+  const reason = err instanceof Error ? err.message : String(err);
+  console.warn(
+    `[afk-home] Malformed AFK home/state env var ignored while deriving the ` +
+      `credential floor: ${reason}. The relocated tree is NOT protected — the ` +
+      `default ~/.afk entries still apply. Fix the env var to restore coverage.`,
+  );
+}
+
+/**
+ * Test-only: clear the once-latch so a suite can assert the warning fires.
+ *
+ * Contract: production code must never call this. Without it a test asserting
+ * the warn would depend on suite ordering — whichever spec touched a malformed
+ * env var first would consume the single latch and every later assertion would
+ * see silence.
+ */
+export function resetAfkHomeWarnLatchForTests(): void {
+  warned = false;
+}
--- a/src/agent/tools/handlers/read-denylist.test.ts
+++ b/src/agent/tools/handlers/read-denylist.test.ts
@@ -45,6 +45,7 @@ beforeEach(() => {
 
 afterEach(() => {
   delete process.env['AFK_READ_DENYLIST'];
+  delete process.env['AFK_HOME'];
   _resetReadDenylistCacheForTests();
   if (existsSync(tmpDir)) rmSync(tmpDir, { recursive: true, force: true });
 });
@@ -285,6 +286,84 @@ describe('read-denylist — AFK_READ_DENYLIST extras', () => {
     // does NOT widen into the real ~/.ssh floor, which stands on its own.
     expect(getReadDenylist().some((p) => p.endsWith('~someone/.ssh'))).toBe(true);
     expect(isReadDenied(join(homedir(), '.ssh', 'id_rsa')).denied).toBe(true);
+  });
+});
+
+describe('read-denylist — AFK_HOME-relocated credential tree (#740)', () => {
+  it('denies afk.env under a relocated AFK_HOME config dir', () => {
+    const relocated = join(tmpDir, 'relocated-home');
+    mkdirSync(relocated, { recursive: true });
+    process.env['AFK_HOME'] = relocated;
+    _resetReadDenylistCacheForTests();
+
+    expect(isReadDenied(join(relocated, 'config', 'afk.env')).denied).toBe(true);
+  });
+
+  it('still denies the real homedir() ~/.afk/config entry when AFK_HOME is relocated', () => {
+    // ADDITIVE, not a replacement: relocating AFK_HOME must not stop covering
+    // the default homedir()-based tree (an operator could have stale files
+    // there, or another process still reads the default location).
+    const relocated = join(tmpDir, 'relocated-home-2');
+    mkdirSync(relocated, { recursive: true });
+    process.env['AFK_HOME'] = relocated;
+    _resetReadDenylistCacheForTests();
+
+    expect(isReadDenied(join(homedir(), '.afk', 'config', 'afk.env')).denied).toBe(true);
+  });
+
+  it('does NOT deny the relocated AFK_HOME state dir (mirrors the default-home divergence)', () => {
+    const relocated = join(tmpDir, 'relocated-home-state');
+    mkdirSync(relocated, { recursive: true });
+    process.env['AFK_HOME'] = relocated;
+    _resetReadDenylistCacheForTests();
+
+    expect(
+      isReadDenied(join(relocated, 'state', 'todos', 'x.json')).denied,
+    ).toBe(false);
+  });
+
+  it('behavior is unchanged and the derived entry does not double up when AFK_HOME is unset', () => {
+    delete process.env['AFK_HOME'];
+    _resetReadDenylistCacheForTests();
+
+    expect(isReadDenied(join(homedir(), '.afk', 'config', 'afk.env')).denied).toBe(true);
+    const configEntryCount = getReadDenylist().filter(
+      (p) => p === safeRealpath(join(homedir(), '.afk', 'config')),
+    ).length;
+    expect(configEntryCount).toBe(1);
+  });
+
+  // Fail-safe: getAfkHome() throws when AFK_HOME is set but not absolute. The
+  // read denylist must not let that throw propagate and empty the floor —
+  // it must skip only the derived entry and keep the homedir()-based ones.
+  it('stays fail-safe when AFK_HOME is a relative path (getAfkHome() throws)', () => {
+    process.env['AFK_HOME'] = 'relative/not-absolute';
+    _resetReadDenylistCacheForTests();
+
+    expect(() => isReadDenied(join(homedir(), '.afk', 'config', 'afk.env'))).not.toThrow();
+    expect(isReadDenied(join(homedir(), '.afk', 'config', 'afk.env')).denied).toBe(true);
+    expect(isReadDenied(join(homedir(), '.ssh', 'id_rsa')).denied).toBe(true);
+  });
+
+  // Cache-invalidation: the memoization key must include AFK_HOME, not just
+  // AFK_READ_DENYLIST, or a runtime AFK_HOME change returns a STALE verdict.
+  // Invariant: deliberately NO `_resetReadDenylistCacheForTests()` call
+  // between the two queries below — that reset is what the memo key itself
+  // is supposed to make unnecessary. Calling it here would make this test
+  // pass even if AFK_HOME were dropped from the key entirely, defeating the
+  // point of the regression guard.
+  it('invalidates the memoized list when AFK_HOME changes, with no manual cache reset (cache-key regression guard)', () => {
+    // `beforeEach` already reset the cache and AFK_HOME is unset at test start.
+    const relocated = join(tmpDir, 'second-home');
+    mkdirSync(relocated, { recursive: true });
+    const target = join(relocated, 'config', 'afk.env');
+
+    // Query once with AFK_HOME unset: the relocated config dir is NOT covered.
+    expect(isReadDenied(target).denied).toBe(false);
+
+    // Change AFK_HOME and query again — WITHOUT resetting the cache by hand.
+    process.env['AFK_HOME'] = relocated;
+    expect(isReadDenied(target).denied).toBe(true);
   });
 });
 
--- a/src/agent/tools/handlers/read-denylist.test.ts
+++ b/src/agent/tools/handlers/read-denylist.test.ts
@@ -396,6 +396,22 @@ describe('read-denylist — AFK_HOME-relocated credential tree (#740)', () => {
     process.env['AFK_HOME'] = relocated;
     expect(isReadDenied(target).denied).toBe(true);
   });
+
+  // paths.ts treats '' as unset (`envVal !== undefined && envVal !== ''`).
+  // That branch was covered on no surface. Pin it: an empty AFK_HOME must
+  // behave exactly like an unset one, never widening or emptying the floor.
+  it('treats an empty AFK_HOME as unset', () => {
+    process.env['AFK_HOME'] = '';
+    _resetReadDenylistCacheForTests();
+
+    expect(isReadDenied(join(homedir(), '.afk', 'config', 'afk.env')).denied).toBe(true);
+    // The mcp.json carve-out still applies under the default home.
+    expect(isReadDenied(join(homedir(), '.afk', 'config', 'mcp.json')).denied).toBe(false);
+    const configEntryCount = getReadDenylist().filter(
+      (p) => p === safeRealpath(join(homedir(), '.afk', 'config')),
+    ).length;
+    expect(configEntryCount).toBe(1);
+  });
 });
 
 describe('parseReadDenylistEntries — the single parser both surfaces share', () => {
--- a/src/agent/tools/handlers/read-denylist.test.ts
+++ b/src/agent/tools/handlers/read-denylist.test.ts
@@ -299,6 +299,37 @@ describe('read-denylist — AFK_HOME-relocated credential tree (#740)', () => {
     expect(isReadDenied(join(relocated, 'config', 'afk.env')).denied).toBe(true);
   });
 
+  // The mcp.json carve-out MUST follow AFK_HOME whenever the denied parent dir
+  // does. Extending only the deny side inverts the carve-out: the registry
+  // becomes unreadable under a relocated home while the default-home spelling
+  // stays readable — invisible to the typed tools AND the bash surface for
+  // exactly the operators who relocated it. Caught by an empirical probe after
+  // the deny-side change looked complete in review.
+  it('keeps the mcp.json carve-out readable under a relocated AFK_HOME', () => {
+    const relocated = join(tmpDir, 'relocated-home-allow');
+    mkdirSync(join(relocated, 'config'), { recursive: true });
+    process.env['AFK_HOME'] = relocated;
+    _resetReadDenylistCacheForTests();
+
+    // The parent dir stays denied …
+    expect(isReadDenied(join(relocated, 'config', 'afk.env')).denied).toBe(true);
+    // … while the registry itself remains readable, exactly as it does at the
+    // default home location.
+    expect(isReadDenied(join(relocated, 'config', 'mcp.json')).denied).toBe(false);
+  });
+
+  it('keeps a relocated mcp.json re-deniable via AFK_READ_DENYLIST (precedence holds)', () => {
+    const relocated = join(tmpDir, 'relocated-home-redeny');
+    mkdirSync(join(relocated, 'config'), { recursive: true });
+    process.env['AFK_HOME'] = relocated;
+    process.env['AFK_READ_DENYLIST'] = join(relocated, 'config', 'mcp.json');
+    _resetReadDenylistCacheForTests();
+
+    // An explicit extra outranks the derived carve-out, same as it outranks the
+    // hardcoded one.
+    expect(isReadDenied(join(relocated, 'config', 'mcp.json')).denied).toBe(true);
+  });
+
   it('still denies the real homedir() ~/.afk/config entry when AFK_HOME is relocated', () => {
     // ADDITIVE, not a replacement: relocating AFK_HOME must not stop covering
     // the default homedir()-based tree (an operator could have stale files
--- a/src/agent/tools/handlers/read-denylist.ts
+++ b/src/agent/tools/handlers/read-denylist.ts
@@ -34,6 +34,7 @@ import { env } from '../../../config/env.js';
 import { basename, dirname, join, resolve } from 'path';
 import { homedir } from 'os';
 import { safeRealpath } from './write-denylist.js';
+import { getAfkHome } from '../../../paths.js';
 
 /**
  * Paths that `read_file` / `grep` / `glob` / `list_directory` must never read —
@@ -174,19 +175,55 @@ export function parseReadDenylistEntries(raw: string | undefined): string[] {
     .map((p) => resolve(p));
 }
 
+/**
+ * Best-effort derived read-denylist entry for the `AFK_HOME`-relocated config
+ * dir (`${getAfkHome()}/config`), ADDITIVE alongside the hardcoded
+ * `${homedir()}/.afk/config` literal in {@link BUILTIN_READ_DENYLIST} (both
+ * spellings are covered; they are equal, and de-duplicated by the caller,
+ * when `AFK_HOME` is unset). Computed HERE — inside {@link resolveLists}, not
+ * at module scope — because a module-level `getAfkHome()` call would
+ * evaluate once at import time and never observe a runtime `AFK_HOME` change
+ * (the test suite mutates env vars; production operators can too via a
+ * process restart with a different env).
+ *
+ * `getAfkHome()` throws when `AFK_HOME` is set but not absolute (or is `/`,
+ * see `paths.ts`). Caught and dropped here — a malformed env var must never
+ * empty the credential floor; the hardcoded `homedir()`-based entries still
+ * apply regardless.
+ */
+function derivedAfkHomeReadEntry(): string[] {
+  try {
+    return [safeRealpath(resolve(join(getAfkHome(), 'config')))];
+  } catch {
+    return [];
+  }
+}
+
 function resolveLists(): {
   builtins: readonly string[];
   extras: readonly string[];
   allow: readonly string[];
 } {
-  const key = env.AFK_READ_DENYLIST ?? '';
+  // Invariant: AFK_HOME is part of the cache key alongside AFK_READ_DENYLIST.
+  // The builtins list now depends on it (derivedAfkHomeReadEntry), so a
+  // runtime AFK_HOME change must invalidate the memo the same way an
+  // AFK_READ_DENYLIST change does — otherwise a stale denylist survives a
+  // relocated AFK_HOME. Joined with U+0000, which cannot occur in either env
+  // value, so the two components can never collide into an ambiguous key.
+  const key = `${env.AFK_READ_DENYLIST ?? ''}\u0000${env.AFK_HOME ?? ''}`;
   if (cached && cached.key === key) return cached;
-  const extras: string[] = parseReadDenylistEntries(key)
+  const extras: string[] = parseReadDenylistEntries(env.AFK_READ_DENYLIST)
     .map((p) => safeRealpath(p))
     .filter(Boolean);
+  const builtins = [
+    ...new Set([
+      ...BUILTIN_READ_DENYLIST.map((p) => safeRealpath(resolve(p))),
+      ...derivedAfkHomeReadEntry(),
+    ]),
+  ];
   cached = {
     key,
-    builtins: BUILTIN_READ_DENYLIST.map((p) => safeRealpath(resolve(p))),
+    builtins,
     extras,
     allow: BUILTIN_READ_ALLOWLIST.map(resolveExceptionEntry),
   };
--- a/src/agent/tools/handlers/read-denylist.ts
+++ b/src/agent/tools/handlers/read-denylist.ts
@@ -35,6 +35,7 @@ import { basename, dirname, join, resolve } from 'path';
 import { homedir } from 'os';
 import { safeRealpath } from './write-denylist.js';
 import { getAfkHome } from '../../../paths.js';
+import { warnAfkHomeRejectedOnce } from '../afk-home-warn.js';
 
 /**
  * Paths that `read_file` / `grep` / `glob` / `list_directory` must never read —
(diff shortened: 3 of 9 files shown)


griffinwork40 (commented):

## Merge decision: **MERGE**

Reviewed against branch HEAD `cf7fc1d` (all citations re-read at that ref). This is a net security improvement: before this PR, `${AFK_HOME}/config` was in **no** denylist at all, so a relocated operator's `afk.env` had zero protection on the typed read tools, the write tools, or bash. After it, the floor follows `getAfkHome()` on all three surfaces. The one finding below is a narrow fail-closed usability gap, not a blocker.

Regime: full (6 files, +440/−34, 3 commits, 225 added test lines). Dimensions: security, api-compat, correctness, spec-compliance, test-coverage, perf-observability.

---

### Findings

**1. `low` · confidence:high · correctness — the bash-surface `mcp.json` carve-out does not follow `AFK_HOME`, so the registry becomes unreadable via bash for relocated operators**

- `src/agent/tools/hooks/bash-restriction-hook.ts:315-320` · ref:`cf7fc1d` · citation-type:file-state (file not in this diff)

```js
function allowlistedFileForms(home: string): string[] {
  return READ_ALLOWLIST_REL.flatMap((rel) => {
    if (isReadDenied(path.join(home, rel)).denied) return [];
    return [path.join(home, rel), `~/${rel}`, `$HOME/${rel}`];
```

Commit 3 correctly restores the carve-out on the **typed** path, but only there. `derivedAfkHomeAllowEntries()` (`read-denylist.ts:223`) is module-private — `git grep` finds exactly two hits, the definition and its single call site in `resolveLists()`'s `allow:` field (`:263`) — and `getReadDenylist()` (`:278`) returns only `[...builtins, ...extras]`, never `allow`. Meanwhile the relocated **deny** root does reach bash, via `deriveRestrictedSubstrings` → `...getReadDenylist()` (`bash-restriction-hook.ts:469`). `allowlistedFileForms` builds its spellings from `homedir()` (threaded from `:201`), and that file contains zero `getAfkHome()` references, so the relocated spelling is never un-masked. The `isReadDenied` filter at `:317` tests only the fixed `homedir()`-spelled path and does not rescue it.

Net effect with `AFK_HOME=/opt/my-afk`: `cat /opt/my-afk/config/mcp.json` is blocked on the bash surface while `read_file` on the same path succeeds. Scope is genuinely narrow — it requires a relocated `AFK_HOME` **and** an interactive surface, because check 2 returns early when no grant manager is wired (`:261-262`), so headless daemon/chat surfaces are unaffected. Fails closed, and the typed path works.

*Suggestion:* export the relocated allow spellings from `read-denylist.ts` and union them into `allowlistedFileForms`, mirroring how the deny side already flows through `getReadDenylist()`.

**2. `nit` · confidence:high · maintainability — `input-parse.ts` is 492 lines at this ref**

`src/agent/tools/subagent/input-parse.ts` · ref:`cf7fc1d`. It was already 422 lines on `main`, so this PR adds ~70 to a file that was over a 350-line modularity ceiling before it arrived. Not enforced mechanically in this repo (`lint` is `tsc --noEmit`; no size script), so this is a nit, not a gate. The `cwd`/`writeRoots`/`readRoots` validation block is the natural extraction candidate.

---

### Verified clean

- **`isTooBroadRoot` (`input-parse.ts:38-44`)** — rejects filesystem roots, the home dir, and ancestors of home via `relative()` containment, checked against both `homedir()` and its realpath (`HOME_TARGETS`, `:22`). Applied identically at all three call sites (`:302`, `:350`, `:420`), each testing the resolved **and** realpath-resolved form. Boundary is tight; no accepted-but-too-broad path found.
- **Write denylist AFK_HOME coverage** — `derivedAfkHomeWriteEntries()` (`write-denylist.ts:70-80`) mirrors **both** hardcoded AFK entries (`config` *and* `state`), so no relocated subpath is left writable. Computed per-call (module is uncached), so a runtime `AFK_HOME` change is observed immediately.
- **Read/write asymmetry on `state`** — deliberate and correctly implemented: read-side derives `config` only, write-side derives both, matching the documented rationale that forks legitimately read skill-preflight/todo/transcript state.
- **Fail-safe direction** — deny side catches a `getAfkHome()` throw and keeps the hardcoded entries (fails safe); allow side drops the extra carve-out (fails closed). Both directions are correct.
- **Memoization** — deriving inside `resolveLists()` rather than at module scope, plus `AFK_HOME` in the cache key, correctly avoids a frozen-at-import denylist.
- **api-compat** — no exported signature changed. `BUILTIN_READ_DENYLIST`, `BUILTIN_WRITE_DENYLIST`, `getReadDenylist`, `getWriteDenylist`, `READ_ALLOWLIST_REL` keep their shapes; production consumers (`bash-restriction-hook.ts`, `risk-classifier.ts`) are unaffected. The new `cwd`/`writeRoots` breadth rejection is a deliberate behavior tightening — a previously-accepted `cwd: "$HOME"` now throws, which is the point of the fix.

### Test coverage — all five stated claims verified present

| Claim | Test |
|---|---|
| Relocated config dir denied | `read-denylist.test.ts:293` |
| Default-home spelling preserved | `read-denylist.test.ts:333` |
| De-dup when `AFK_HOME` unset | `read-denylist.test.ts:356` |
| Fail-safe on malformed `AFK_HOME` | `read-denylist.test.ts:370` |
| Cache invalidation, no manual reset | `read-denylist.test.ts:386` |

Breadth guard is pinned for both newly-covered fields: `cwd` at `input-parse.test.ts:369` (home), `:375` (ancestor of home), `:382` (filesystem root); `writeRoots` at `:440`. Write side covers both relocated dirs at `write-denylist.test.ts:388`.

One coverage gap, consistent with finding 1: the carve-out-under-relocated-home test (`read-denylist.test.ts:308`) asserts only against `isReadDenied`, and no test file under `src/agent/tools/hooks/` references `AFK_HOME` at all — so the bash surface of this fix is unpinned. That absence is exactly what let the typed-path fix look complete.

### What was not checked

- Read against branch HEAD `cf7fc1d`; every blocking/high citation re-read at that ref, and the one high finding was independently re-derived by a shadow verifier which recalibrated it from `high` to `low` on scope grounds.
- Stated intent: PR #753 title + body — spec-compliance assessed. No unmet intent and no scope creep found; the three items the body declares out of scope (#736 case-variants, `isReadDenied` on `cwd`/`writeRoots`, #740's ancestor containment check) are genuinely untouched.
- Did not execute the test suite (CI is green on all 8 required checks), did not probe a real relocated `AFK_HOME` at runtime, and did not review case-insensitive-filesystem spellings (#736, explicitly out of scope).
- `bash-restriction-hook.ts` is not part of this diff; it was read as a consumer to trace the carve-out, and portions of its substring-matching internals were not read line-by-line.

griffinwork40:
@codex fix comments

griffinwork40 (commented):

One `high` finding sits inside the PR's own headline purpose, and the fix is a one-line normalization. **This is not a regression against `main`** — before this PR a relocated `AFK_HOME` had zero coverage on all three surfaces, so landing it strictly improves the floor. The gate is that the bash leg of "the floor follows a relocated home" does not hold for one unvalidated env spelling. Landing as-is plus an immediate follow-up is a defensible maintainer call; the finding is cheap enough to fold in first.

Reviewed against branch HEAD `eb9c6410`. Every citation below was re-read at that ref; the two top-severity claims were independently re-derived by adversarial verifiers.

---

### 1 · high · security — bash guard fails open on a trailing-separator `AFK_HOME`
`src/agent/tools/hooks/bash-restriction-hook.ts:291, 307-308, 267` · ref:`eb9c6410` · citation-type: diff-context

A trailing separator on `AFK_HOME` (e.g. `AFK_HOME=/opt/my-afk/`) makes the bash guard miss `cat $AFK_HOME/config/afk.env`: the raw env string is substituted verbatim into the command text, producing a double separator, while the restricted needle is built with `path.join`, which collapses it — and the final match is a literal substring test.

```
291   const normalized = afkHome === undefined ? command : command.replace(/\$AFK_HOME/g, afkHome);
307   const afkHome = configuredAfkHome();
308   return afkHome === undefined ? [] : [path.join(afkHome, 'config')];
267       if (scanned.includes(sub)) {
```

`getAfkHome()` returns `env.AFK_HOME` un-normalized — `src/paths.ts:49-54` guards only `!isAbsolute(envVal) || envVal === '/'`, so `/opt/my-afk/` passes with the slash intact. Scanned text becomes `/opt/my-afk//config/afk.env`; the needle is `/opt/my-afk/config`; `includes()` misses. POSIX collapses interior separators, so the command opens the identical `afk.env` the guard blocks without the slash. Full needle enumeration (all 14 `BUILTIN_READ_DENYLIST` entries, `withEtcAliases` twins, `getReadDenylist()` realpath'd forms, `builtinBashSensitiveRoots()`, `relocatedAfkSensitiveRoots`) yields **zero** matches against the double-separator string — no other needle rescues it. Nothing collapses the scanned text before the scan (`normalizeHomeRefs` does three `.replace()` calls; `scrubAllowlistedRefs` only substitutes allowlisted forms).

Scope is **bash-surface-only**, which is what holds this at `high` rather than `critical`: the typed read/write predicates run `safeRealpath(resolve(filePath))` as their first statement (`read-denylist.ts:296`, `write-denylist.ts:158`) and compare with `real === blocked || real.startsWith(blocked + '/')` (`read-denylist.ts:317`), so `//`, `..`, relative, and symlink spellings all normalize and are denied. `path-approval-hook.ts:213-215` keeps the raw `//` but delegates to `isReadDenied`, whose internal `resolve` rescues it.

**Suggestion:** normalize once at the source — `configuredAfkHome()` returns `path.resolve(getAfkHome())` — so substitution and needle construction share one spelling. Add a trailing-slash case to `bash-restriction-hook.test.ts`; all four new `AFK_HOME` tests there use `join(tmpdir(), …)` and none carries a trailing separator.

### 2 · medium · correctness / security — derived write floor follows `AFK_HOME` but not `AFK_STATE_DIR`
`src/agent/tools/handlers/write-denylist.ts:70-80` · ref:`eb9c6410` · citation-type: diff-context

The derived write entry re-derives the state tier as `join(getAfkHome(), 'state')`, but the project's actual rule is `getAfkStateDir()`, which returns `AFK_STATE_DIR` *independently* of `AFK_HOME`.

```
72       const home = getAfkHome();
74       safeRealpath(resolve(join(home, 'config'))),
75       safeRealpath(resolve(join(home, 'state'))),
```

`src/paths.ts:186-201`: `const envVal = env.AFK_STATE_DIR; if (envVal !== undefined && envVal !== '') { return envVal; } return join(getAfkHome(), 'state');`. An operator with `AFK_STATE_DIR=/opt/state` gets no write protection on the real state tier. This is the same env-relocation class the PR set out to close and is not in the stated out-of-scope list. The config entry is correct (`getAfkConfigDir()` genuinely is `join(getAfkHome(), 'config')`). The pre-existing hardcoded `${homedir()}/.afk/state` has the same hole, so this is an incomplete extension, not a regression.

**Suggestion:** add `safeRealpath(resolve(getAfkStateDir()))` to `derivedAfkHomeWriteEntries()`.

### 3 · medium · perf-observability — every `AFK_HOME` derivation swallows its throw silently
`read-denylist.ts:197-199`, `write-denylist.ts:77-79`, `bash-restriction-hook.ts:301-303` (+ `read-denylist.ts:229-231`, which fails closed) · ref:`eb9c6410` · citation-type: diff-context

All four sites are bare `catch { return []; }` / `catch { return undefined; }`, and the four changed production files contain zero logging calls (grep across all four → one hit, a doc-comment string at `bash-restriction-hook.ts:35`). An operator with a typo'd `AFK_HOME` runs on a silently reduced credential floor with no diagnostic anywhere — the fail-safe is correct, its invisibility is the defect.

**Suggestion:** emit one warn-level line naming the rejected `AFK_HOME` value on the first catch per process.

### 4 · low · perf-observability — `getWriteDenylist()` is uncached on the write hot path
`write-denylist.ts:65-68, 90-102, 157-159` · ref:`eb9c6410` · citation-type: file-state

The module is uncached (self-documented at `:65-68`; a full read of all 167 lines finds no memo/Map/cache symbol), so the two added `safeRealpath` calls plus the new `Set` allocation raise per-write `realpathSync` syscalls from 11 to 13 on every `write_file`/`edit_file`.

**Suggestion:** memoize on `${AFK_WRITE_DENYLIST}\u0000${AFK_HOME}`, mirroring `resolveLists`. The read path is fine — cache-hit cost grows by one `process.env` getter and one string concat.

### 5 · low · test-coverage — four untested new branches
Grep-confirmed absent at the ref: no trailing-separator `AFK_HOME` case in any of the four changed test files; no malformed/relative-`AFK_HOME` test on the bash surface, leaving the `configuredAfkHome()` catch untested; no symlink breadth test for `cwd` or `writeRoots` (all 11 symlink/realpath hits sit inside the `readRoots` block, lines 470-621) though both new call sites invoke `realpathSafe`; `writeRoots` has no filesystem-root case while `cwd` has one.

---

### Verified clean

- **Cache invalidation holds.** The memo key is `` `${env.AFK_READ_DENYLIST ?? ''}\u0000${env.AFK_HOME ?? ''}` `` (`read-denylist.ts:245`) and `env.AFK_HOME` is a live getter (`src/config/env.ts:1447`), not an import-time snapshot — the feared stale non-covering denylist does not exist. `AFK_HOME=''` collapses to the same key *and* the same `getAfkHome()` result as unset.
- **Allow-side union does not widen.** Precedence is operator extras (`:306-310`) → allow (`:315`) → builtins (`:316-320`); `allow` is exact `Array.includes`, no prefix or `dirname` walk, so `AFK_READ_DENYLIST` still outranks the new derived carve-out. Precisely stated: the carve-out suppresses *builtin* denials by design and cannot override operator extras.
- **Dedup is correct.** The derived read entry is byte-identical to the hardcoded one when `AFK_HOME` is unset, and the `new Set` collapses it.
- **`isTooBroadRoot` parity.** One module-scope helper (`subagent/input-parse.ts:38-44`) is called identically for the `cwd` string (`:302`) and per array entry (`:350` writeRoots, `:420` readRoots) — grep yields exactly those three call sites, and no fourth field accepting a path root is left unguarded. It catches `/`, `$HOME`, `/Users`, and home-with-trailing-slash, without over-rejecting `/tmp/wt/feat-y`. Empty array → field normalizes to `undefined`, unchanged.
- **api-compat: no issues found.** Reachability pre-check per candidate: `normalizeHomeRefs`, `allowlistedFileForms`, `scrubAllowlistedRefs` gained required parameters but are unexported with 0 importers. `getWriteDenylist`/`getReadDenylist` signatures unchanged, additive return values only; the one production importer (`src/agent/risk-classifier.ts:174`) is strictly hardened. `parseAgentInput` newly throws on `cwd: homedir()` / `writeRoots: [homedir()]` — intentional, JSDoc updated, one production call site (`subagent-executor.ts:410`).

### Spec-compliance — assessed against PR title + body

**No unmet intent** and **no scope creep.** Every named clause has an implementing change: `AFK_HOME`-following deny side, derivation inside `resolveLists()` rather than module scope, `AFK_HOME` in the memo key, the fail-safe catch, `getAfkHome()` reuse, and the allowlist union failing closed. The `state`-in-write-not-read asymmetry is respected as designed. The `isTooBroadRoot` hoist is behavior-preserving for `readRoots`.

One **documentation correction**: the body states "No change needed in `bash-restriction-hook.ts`", but the diff changes that file (+34/−7) via a fourth commit not described in the body. This is a legitimate follow-up rather than unstated scope — commit 2's own title names bash as a target surface, and the stale sentence is mechanically wrong because `getReadDenylist()` returns only the symlink-*resolved* spelling while that scanner is lexical, so a symlinked `AFK_HOME`'s as-spelled form went unmatched and `$AFK_HOME` was never expanded at all. The PR's own test at `bash-restriction-hook.test.ts:626` demonstrates the gap was real. Fix the body, not the diff.

### What was not checked

- Read against branch HEAD `eb9c6410`; all citations verified at that ref. `src/paths.ts` and `src/config/env.ts` are **not** modified by this PR — findings 1 and 2 rest on unchanged supporting code, so they describe pre-existing resolution behavior the new derivations inherit rather than defects this PR introduces.
- **Tests, `pnpm lint`, and `tsc` were not run** (read-only review). The "184 tests pass" and "lint clean" claims are unaudited; the claimed new coverage was verified by *reading* the test files — relocated config dir, preserved default-home coverage, dedup when unset, malformed-`AFK_HOME` fail-safe, and the headline cache-invalidation regression test all genuinely exist, and that last one confirmed to contain no manual cache reset between its two queries.
- Not checked: Windows path semantics for the `$AFK_HOME` regex and `parsePath().root`; `at-file-inject.ts` (the second `READ_ALLOWLIST_REL` consumer) under relocation; whether `write-file.ts`/`edit-file.ts` invoke `assertNotDenylisted` (the predicate's ordering is confirmed, its write call site is not); downstream impact in `risk-classifier.ts:174`; the model-facing tool-schema description in `src/agent/tools/schemas.ts`, which is not among the changed files and plausibly still omits the breadth rejection.
- Deliberately out of scope per the PR: case-variant spellings (#736), the ancestor containment check in `deriveRestrictedSubstrings` (#740), and denylist rejection on `cwd`/`writeRoots`.
- Verification posture: both `high`-tier claims were re-derived from scratch by independent adversarial verifiers reading upstream producers and downstream consumers — both returned CONFIRMED. Zero citations were fabricated and zero absence claims were refuted across the whole finding set.

griffinwork40 (commented):

## Merge decision: **MERGE**

No blocking defect found. All three stated commits landed, the security-relevant logic is correct at the reviewed ref, and the test suite covers the behaviors the PR body claims it covers. Findings below are non-blocking follow-ups.

Reviewed ref: `eb9c6410ca60f0e8339874ef891a1da79d44e585` · regime: full (2 dimension agents + citation/absence verification pass)

---

### Findings

**medium · high confidence · spec-compliance · PR description vs. `src/agent/tools/hooks/bash-restriction-hook.ts:290-345` · ref:eb9c641 · citation-type:file-state**

finding: The PR body misdescribes its own change surface on a credential-floor PR — it states commit 2 needed no bash-hook change, but the diff adds 34 lines of security-relevant logic to that file, and the stated line/file counts are wrong.

evidence:
```
body: "No change needed in `bash-restriction-hook.ts`: it already derives its
       sensitive-root list from `getReadDenylist()` …"
body: "440 insertions / 34 deletions across 6 files"
actual: +523/-41 across 8 files; bash-restriction-hook.ts +34/-7
        (new configuredAfkHome() :298, relocatedAfkSensitiveRoots() :306,
         $AFK_HOME expansion in normalizeHomeRefs :291, relocated forms
         in allowlistedFileForms :339-345)
```
The added code is correct and load-bearing — `relocatedAfkSensitiveRoots()` is wired into the bash gate at `deriveRestrictedSubstrings` (`bash-restriction-hook.ts:497`), and `$AFK_HOME`-spelled commands are covered by tests at `bash-restriction-hook.test.ts:605-636`. The defect is the description: a reviewer auditing a credential floor is told a file is untouched when it carries new gating logic.

suggestion: Update the body to name the bash-hook change as part of commit 2/3 scope and correct the diffstat to +523/-41 across 8 files.

---

**low · high confidence · test-coverage · `src/paths.ts:45` · ref:eb9c641 · citation-type:file-state**

finding: An empty-string `AFK_HOME` takes the "treated as unset" branch and is covered on no surface — read, write, and bash suites all test unset and malformed-relative, never `''`.

evidence:
```ts
// src/paths.ts:45
if (envVal !== undefined && envVal !== '') {
```
Absence verified: `grep -n "AFK_HOME', ''|AFK_HOME'] = ''"` across `read-denylist.test.ts`, `write-denylist.test.ts`, `bash-restriction-hook.test.ts` → zero matches. Current behavior is safe (falls back to `~/.afk`, which the hardcoded entries already cover), so this is a coverage gap, not a hole in the floor.

suggestion: Add one case asserting `AFK_HOME=''` yields the same denylist as unset, pinning the fallback branch.

---

**low · high confidence · perf-observability · `src/agent/tools/handlers/write-denylist.ts:70-80` · ref:eb9c641 · citation-type:file-state**

finding: A malformed `AFK_HOME` silently narrows the derived floor with no operator-visible signal — all four derivation sites swallow the `getAfkHome()` throw and return an empty list.

evidence:
```ts
  } catch {
    return [];
  }
```
Same pattern at `read-denylist.ts:197-199`, `read-denylist.ts:229-231`, and `bash-restriction-hook.ts:301-303`. Failing safe is the right call — the hardcoded `homedir()` entries still apply — but an operator who typos `AFK_HOME` gets no warning that their relocated tree is uncovered.

suggestion: Emit a one-time warning when the derivation throws, so a typo is visible rather than inferred from behavior.

---

**low · high confidence · perf-observability · `src/agent/tools/handlers/write-denylist.ts:90-102` · ref:eb9c641 · citation-type:file-state**

finding: `getWriteDenylist()` adds 2 `realpathSync` calls per invocation on a path consulted on every write-tool call, amplifying (not introducing) the module's pre-existing uncached pattern.

evidence:
```ts
export function getWriteDenylist(): readonly string[] {   // :90 — no memo
      ...BUILTIN_WRITE_DENYLIST.map((p) => safeRealpath(resolve(p))),  // :97 (12 entries)
      ...derivedAfkHomeWriteEntries(),                                 // :98 (+2: config, state)
```
Callers: `src/agent/risk-classifier.ts:174` and `write-denylist.ts:159` (`assertNotDenylisted`). The uncached design is deliberate and documented at `write-denylist.ts:65-68`; 14 vs. 12 syscalls is not a regression worth blocking, and the read side (the genuinely hot path) is memoized with `AFK_HOME` in the key at `read-denylist.ts:245`.

suggestion: None required for merge; memoize on the same `AFK_HOME`-inclusive key pattern if write-path profiling ever shows this.

---

**low · high confidence · api-compat · `src/agent/tools/subagent/input-parse.ts:302-308` · ref:eb9c641 · citation-type:file-state**

finding: `parseAgentInput` now rejects `cwd` values it previously accepted, which breaks any workflow that dispatches with the home dir as `cwd` — including the case where a monorepo checkout *is* the home dir.

evidence:
```ts
    if (isTooBroadRoot(resolvedCwd) || isTooBroadRoot(realResolvedCwd)) {
      throw new Error(
        `Agent tool cwd must not be a filesystem root, your home directory, ` +
```
Reachability confirmed: sole production importer is `src/agent/tools/subagent-executor.ts:29` (`isTooBroadRoot` itself is module-private — 5 refs, all inside `input-parse.ts`). This rejection is exactly what the PR set out to add, so it is intended behavior, not a defect — flagged only because it is a user-visible contract change.

suggestion: Name the `cwd` rejection in release notes so operators with home-rooted checkouts learn it from a changelog rather than a runtime throw.

---

### Verified clean

- **security · read denylist** — `read-denylist.ts:194-232, 245-256`. The derived `${AFK_HOME}/config` entry is additive, computed inside `resolveLists()` (not module scope), and `AFK_HOME` is in the memo key (`:245`), so a runtime relocation cannot serve a stale non-covering list. The `state` omission is a documented invariant (`:60-62`, citing #544/#547/#554), not an oversight — `state` is write-denied and read-allowed by design.
- **security · allow/deny precedence** — `isReadDenied` matches operator `AFK_READ_DENYLIST` extras *before* the carve-out list, so `AFK_READ_DENYLIST=<relocated>/config/mcp.json` still wins; the carve-out is exact-file (`allow.includes(real)`), never a prefix. Covered by `read-denylist.test.ts:330-341`.
- **security · fail-safe direction** — all four `getAfkHome()` catch sites fail in the correct direction for their side: deny sites drop only the derived entry (hardcoded floor intact), and the allow site fails *closed* (no extra carve-out). Tested at `read-denylist.test.ts:344-353` and `write-denylist.test.ts:425-429`.
- **security · bash parity** — `relocatedAfkSensitiveRoots()` (`:306`) is wired at `:497` and correctly mirrors the *read* denylist (config only), which is the right baseline for a read gate. Symlinked `AFK_HOME` blocked at `bash-restriction-hook.test.ts:626-636`.
- **correctness · de-duplication** — the `new Set` at `read-denylist.ts:251-255` collapses the derived and hardcoded spellings when `AFK_HOME` is unset; pinned by `read-denylist.test.ts:356-364` asserting exactly one config entry.
- **test-coverage · breadth guard** — the `isTooBroadRoot` guard is tested on all three fields: `cwd` (`input-parse.test.ts:369-385`: home, ancestor-of-home, filesystem root), `writeRoots` (`:440-442`), `readRoots` including symlink-to-root and symlink-to-home (`:518-565`).
- **spec-compliance · declared boundaries held** — `deriveRestrictedSubstrings`' containment check is unchanged (#740 left open as stated), no `isReadDenied` rejection was added to `cwd`/`writeRoots`, no #736 case-variant handling crept in, and no #735 overlap.

### Dropped in verification

Four findings raised by the dimension agents were dropped before this report: two refuted by direct read at the ref (a claim that bash-gating should mirror the *write* denylist and cover `${AFK_HOME}/state` — wrong baseline, contradicts the `:60-62` invariant; and a claim that the relocated `mcp.json` carve-out derives empty — refuted by `READ_ALLOWLIST_REL` at `:104`), and two that negated themselves in their own evidence (a `new Set` dedup concern concluding "the code works correctly", and a bash-hook ordering concern concluding "no bug").

### What was not checked

- Wave 1 agents read against branch HEAD `eb9c6410ca60f0e8339874ef891a1da79d44e585`; every surviving citation was re-verified at that ref from extracted file state. No fabricated or diff-only citations remain.
- Stated intent: PR #753 title + body — **spec-compliance assessed** against it.
- No tests, builds, or lint were executed; the author's claim of 184 passing tests and clean `tsc --noEmit` is taken as reported, corroborated only by CI being green on this head SHA.
- Not checked: case-insensitive filesystem behavior (#736, explicitly out of scope), runtime behavior at subagent depth ≥2, whether anything credential-bearing is written under `${AFK_HOME}/state` by the 24 state writers outside these 8 files, and the Telegram/daemon surfaces.
- No `/shadow-verify` wave ran: no `critical` or `high` finding survived verification, and the skill routes `medium` and below straight through.

PR #665, opened 2026-07-21
fix(tui): flush logical lines to scrollback so resize reflow stops splitting words and dropping indent

griffinwork40:
## Problem
Narrowing the terminal pane after long output has scrolled into native scrollback makes the terminal re-wrap that stored scrollback at the new width — splitting words mid-character (`missing` → `miss` + `ing`) and dropping continuation-line indentation to column 0.

## Root cause
AFK hard-wraps committed output to physical rows for exact live-cursor frame math, then archives those *pre-wrapped physical rows* into native scrollback with a raw newline (`terminal-compositor.committed-band-commit.ts`). Once bytes are in native scrollback the terminal owns them, so the in-memory reflow fix (`band-reflow.ts`, PR #386 / `f75d8a4`) cannot reach them. On a narrowing resize, the terminal's DECAWM autowrap hard-wraps each stored row again → mid-word splits + lost indent. `band-reflow.ts:10-22` and `docs/tui-resize-reflow.md:129-142` document this as a known out-of-scope gap.

## Fix
Retain per-row logical provenance (`committedBandMeta` / `BandRowMeta`) and, at the band-hold archive site, emit whole **soft-wrappable logical lines** via `buildScrollbackArchiveEscape` instead of pre-wrapped physical rows. Native scrollback then holds one reflow-clean logical line per logical line, which any emulator — and tmux copy-mode (`-J`) — rewraps cleanly on resize.

## Gates (rebased onto current main, zero conflicts)
- `pnpm lint` — 0 errors
- logical-flush test — 24/24
- compositor suite — 381/381
- PTY suite (`pnpm test:pty`) — 8/8 (incl. the 3 axis-2 resize guards)

## MERGE GATE — human real-terminal validation required (do NOT merge until checked)
PTY tests replay an xterm emulator and cannot reproduce terminal-only DECAWM deferred-wrap artifacts; prior fixes in this path shipped green and broke on real terminals.

- [ ] iTerm2, Terminal.app, and Ghostty
- [ ] Long single line (~150+ cols, no spaces) scrolled into scrollback
- [ ] Narrow pane (120 to 68): reflows as one continuous line, no mid-word splits
- [ ] Widen pane back (48 to 110): rejoins cleanly, no stranded fragments
- [ ] Markdown table under a collapsed "thinking" overlay: header appears once, no void above the frame
- [ ] Native scrollback: each committed line appears once, in order
- [ ] tmux copy-mode (`-J`): unbroken logical line
[...]

PR #755, opened 2026-07-29
test(tui): pin the grow-eviction width-resize defect the #540 guards cannot see

griffinwork40:
## This PR does not fix the bug. It proves the bug exists where we assumed it didn't.

Read that first, so nothing here is mistaken for a fix. No production behaviour changes: one new PTY guard, one comment correction, one docs correction.

## What happened

A real session showed the reported width-resize symptom again — after widening the terminal, output from a completed turn was still hard-wrapped mid-word in scrollback with its indent lost, and a block appeared twice. The session was running **5.82.1, which already contains #665** (`git tag --contains 71e5450` → v5.81.6 …), verified by tying the session to its PID and reading `ps -o lstart` against the npm install log. So this is not a stale binary and not a regression — it is a site #665 never covered.

## Why our guards were green

`width-resize-fragment-narrow` and `-widen` both hold a tall overlay across every `commitAbove`:

```js
c.setOverlay(overlay); c.commitAbove(longLine + '\n');
```

That routes every commit through the **band-hold archive** — the one write site #665 converted to soft-wrappable logical lines. Both guards carry `ref: '#540 axis-2 · committed-band-commit.ts band-hold archive (logical flush)'`. They test the fixed site. 7 of 9 scenarios set a tall overlay.

Meanwhile `frame-preserve.ts` justified keeping its grow-eviction paths on pre-wrapped physical rows like this:

> The band-hold archive is the site that actually carries the width-resize content in practice (verified by instrumenting the #540 PTY guards), so retaining physical emission here does NOT reintroduce the fragmentation those guards check.

**That claim is false.** The ordinary sequence — a turn commits under a short frame, then the next turn's thinking lane grows the frame — evicts through `preserveRowsBeforeFrameRender`, not band-hold. Terminals reflow only their own soft wraps, so the app hard newlines it emits can never rejoin on widen.

## The guard

`width-resize-fragment-evict-growth` drives the eviction site directly: commit under a short frame (no overlay → lands in the band and is painted, no archive), then grow the overlay so grow-eviction scrolls those painted rows out, then widen 48 → 110.

Result at `22751af`: **4 non-wrapped rows where a rejoined line shows 1.** With the aspirational assertion it fails `expected 4 to be less than or equal to 1`; committed in RED-guard shape (`minNonWrappedRows: 2`) so CI is green while the defect is pinned and cannot silently persist. Not pinned to exactly 4 — any value >1 means the terminal could not rejoin it.

## Why I did not fix it here

`frame-preserve.ts` records that the obvious fix was already tried and regressed:

> converting them to the top-paint-logical-then-scroll mechanism regressed the verdict-card overflow case (the card's closing border was dropped on a growth eviction)

So the real fix is the A2 unification (`docs/proposals/tui-compositor-rewrite.md` §5 / #540 Stage 2), which #540 gates on a human-in-the-loop real-terminal validation matrix. Shipping a plausible-looking conversion here would be the fourth green-but-broken release in this class. Flip this guard to `maxNonWrappedRows: 1` when that lands — it should be the acceptance test.

## Also corrected

`docs/scrollback.md` claimed **"no test verifies actual scrollback contents end-to-end on a real PTY"**. That predates the #541 harness (`11fb1eb`), which does exactly that. The residual risk moved from capability to **aim**: a scenario only covers the write site its `drive()` exercises. Added guidance to name the write site in `ref` and check whether the others are covered.

## Gates

- `pnpm test:pty` 10/10, new guard included
- `pnpm test` 13513 passed / 711 files
- `pnpm lint` clean; `audit:env:check`, `scan:env:check`, `audit:sdk:check`, `audit:chalk:check`, `build` all OK

Refs #540. Follow-up to #665 (#541 harness).

### DIFF
--- a/docs/dev/real-terminal-matrix.md
+++ b/docs/dev/real-terminal-matrix.md
@@ -90,8 +90,13 @@ guard drove the one already-fixed write site. Verify, do not assume — composit
 phase traces go to **stderr**, so redirect it and leave stdout (the visual output)
 alone:
 
+All guard refusals print on **stdout**, so they stay visible under the `2>` redirect
+below. If the pane goes blank and you get an immediate prompt back, read what the
+script printed — it refused; it did not crash.
+
 Run these as **two separate commands** — the script holds for 45s, and pasting both
-at once can join them into one line:
+at once can join them into one line (which also silently creates a file named
+`grep` in the repo root; delete it if you see one):
 
 ```bash
 AFK_DEBUG_COMPOSITOR=1 pnpm exec tsx scripts/visual-void-repro.ts widen-evict 2>/tmp/comp.log
--- a/docs/dev/real-terminal-matrix.md
+++ b/docs/dev/real-terminal-matrix.md
@@ -14,7 +14,11 @@ refuses a non-TTY (`exit 2`), so it cannot be piped, screen-scraped, or run from
 agent shell. `tmux` is a valid extra row but not a substitute: it reflows only
 *soft*-wrapped lines and never sends SIGWINCH itself.
 
+**Run from the repo root** — `pnpm exec` resolves against the nearest workspace, so
+running from `~` fails with `ERR_PNPM_RECURSIVE_EXEC_NO_PACKAGE`:
+
 ```bash
+cd ~/Projects/open_source/agent-afk
 pnpm exec tsx scripts/visual-void-repro.ts <scenario>
 ```
 
@@ -61,9 +65,10 @@ band-hold archive (already fixed by #665) rather than grow-eviction. Only a
 |---|---|---|
 | Grow-eviction → widen | `visual-void-repro.ts widen-evict` | see below |
 
-Procedure: start the pane **narrow (48–70 cols)**, run it, and during the 45s hold
-**drag the window wider** (or `cmd`-`-` to shrink the font — same effect). Then
-scroll up:
+Procedure: start the pane **narrow (48–70 cols)** — the scenario refuses to run
+above 80 cols and exits 2, because content has to be *committed* narrow for a later
+widen to have anything to rejoin. Then during the 45s hold **drag the window wider**
+(or `cmd`-`-` to shrink the font — same effect). Then scroll up:
 
 - `LONGLINE-MARKER … END-OF-LONGLINE` — **PASS** it rejoined into one soft-wrapped
   paragraph. **FAIL** it is split across rows that break **mid-word**, with
@@ -85,8 +90,16 @@ guard drove the one already-fixed write site. Verify, do not assume — composit
 phase traces go to **stderr**, so redirect it and leave stdout (the visual output)
 alone:
 
+Run these as **two separate commands** — the script holds for 45s, and pasting both
+at once can join them into one line:
+
 ```bash
 AFK_DEBUG_COMPOSITOR=1 pnpm exec tsx scripts/visual-void-repro.ts widen-evict 2>/tmp/comp.log
+```
+
+then, after it exits:
+
+```bash
 grep -E 'commitAbove:phase1|evict:enter' /tmp/comp.log
 ```
 
--- a/docs/dev/real-terminal-matrix.md
+++ b/docs/dev/real-terminal-matrix.md
@@ -1,0 +1,105 @@
+# Real-terminal validation matrix (compositor)
+
+The human-in-the-loop gate #540 requires before any compositor commit-path change
+lands. It exists because **two prior fixes shipped headless-green and were broken
+in reality** (`docs/scrollback.md`), and a third — #665 — shipped with
+`test:pty 8/8` and still fragmented real scrollback the next day (#755).
+
+Run every row. Green tests are not a substitute; that is the whole point of this file.
+
+## Setup
+
+Run in a **real terminal** — iTerm2, Apple Terminal, Ghostty, or xterm. The script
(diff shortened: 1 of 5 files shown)


griffinwork40:
## Real-terminal matrix — reflow axis, write-site confirmation (partial)

First real-terminal run of `widen-evict`, macOS, Ghostty, pane narrowed to <=80 cols then dragged wide during the hold. Base: this branch @ `3a8c1e4`.

`AFK_DEBUG_COMPOSITOR=1 pnpm exec tsx scripts/visual-void-repro.ts widen-evict 2>/tmp/comp.log`

```
commitAbove:phase1 lineCount=4 fitsAboveFrame=true bandOverflow=0 rawGenuineOverflow=0 archiveCount=0 maxBandModel=23 overflowRunLen=4
commitAbove:phase1 lineCount=4 fitsAboveFrame=true bandOverflow=0 rawGenuineOverflow=0 archiveCount=0 maxBandModel=23 overflowRunLen=8
commitAbove:phase1 lineCount=2 fitsAboveFrame=true bandOverflow=0 rawGenuineOverflow=0 archiveCount=0 maxBandModel=23 overflowRunLen=10
commitAbove:phase1 lineCount=2 fitsAboveFrame=true bandOverflow=0 rawGenuineOverflow=0 archiveCount=0 maxBandModel=23 overflowRunLen=12
commitAbove:phase1 lineCount=2 fitsAboveFrame=true bandOverflow=0 rawGenuineOverflow=0 archiveCount=0 maxBandModel=23 overflowRunLen=14
commitAbove:phase1 lineCount=2 fitsAboveFrame=true bandOverflow=0 rawGenuineOverflow=0 archiveCount=0 maxBandModel=23 overflowRunLen=16
evict:enter rows=16 anchorRow=1
```

**Write site confirmed.** `archiveCount=0` on every commit means the band-hold archive — the one site #665 converted to logical lines — never fired. `bandOverflow=0` on every commit means `commitAbove`'s ordinary physical scroll never fired either (consistent with `commit-mode.ts:216`, where `useBandHold` pre-empts it). The single `evict:enter rows=16` is grow-eviction in `frame-preserve.ts` evicting the whole band to scrollback as app hard newlines.

Row accounting ties out: 6 commits at lineCount 4+4+2+2+2+2 = 16 = `overflowRunLen=16` = `evict:enter rows=16`. `maxBandModel=23 > 16` is why band-hold declined.

**This is the same site the new PTY guard `width-resize-fragment-evict-growth` drives**, so the headless harness and a real terminal now independently agree on which write site carries width-resize content through the ordinary end-of-turn -> next-turn sequence. That retires the empirical claim in `terminal-compositor.frame-preserve.ts:10-30` on two surfaces rather than one.

### Still outstanding on this row

Visual verdict not yet recorded — whether `LONGLINE-MARKER … END-OF-LONGLINE` rejoined or stayed broken mid-word after the widen, and whether `CARD-BOTTOM` survived. The trace proves the correct code path executed; it does not measure the fragmentation. Axis 1 (void class: `long` / `short` / `grow-collapse` / `resize`, plus the interactive dropdown-headroom and picker rows) is also not yet run — not required for this PR, which changes no runtime behaviour, but required before the A2 unification lands.

griffinwork40:
## Real-terminal matrix — reflow axis row COMPLETE: **FAIL (defect reproduced)**

Follow-up to the trace comment above. Visual verdict recorded; the row is now complete.

macOS, Ghostty. Committed at ~60-70 cols, then dragged to **120 cols** during the hold. Branch @ `488af86`.

### Verdict: FAIL — content did not rejoin

Final state at 120 cols, scrolled up:

```
LONGLINE-MARKER reflow reflow reflow reflow reflow reflow re
flow reflow reflow reflow reflow reflow reflow reflow reflow
 reflow reflow reflow reflow reflow reflow reflow reflow ref
low END-OF-LONGLINE
```

- **4 physical rows**, still wrapped at **~60 cols inside a 120-col pane** — they never rejoined on widen.
- Breaks land **mid-word**: `re|flow`, `ref|low`.
- Continuations start at **column 0**, one with a stray leading space.

This is the reported signature reproduced deliberately. It is also **the same row count the headless guard measures** (`width-resize-fragment-evict-growth` asserts 4 non-wrapped rows where a rejoined line shows 1), so the PTY harness and a real terminal agree on the magnitude, not just the site.

### Regression checks on this row — clean

- `CARD-BOTTOM` **present**; the card renders intact (`┌─ CARD-TOP ─┐` / `│ card body row │` / `└─ CARD-BOTTOM ─┘`). No border-drop, as expected since eviction is unchanged here.
- No block appears twice; `PROSE-01`..`PROSE-04` each exactly once.
- No box borders spliced at differing indents.

### Row summary

| Check | Result |
|---|---|
| Write site driven | grow-eviction (`evict:enter rows=16`); band-hold never fired (`archiveCount=0` x6) |
| Long line rejoins on widen | **FAIL** — 4 mid-word rows at ~60 cols in a 120-col pane |
| `CARD-BOTTOM` survives | PASS |
| No duplication / spliced borders | PASS |

### Notes from the run

The guard refused twice first (120 cols, then 88 cols) and both refusals were **visible on stdout** under the `2>/tmp/comp.log` redirect — confirming the fix in `488af86`. Successful run took 45s (the hold), consistent with the prompt's reported duration.

Axis 1 (void class: `long` / `short` / `grow-collapse` / `resize`, plus interactive dropdown-headroom and picker) remains unrun. Not required for this PR, which changes no runtime behaviour, but required before the A2 unification lands.

PR #756, opened 2026-07-29
fix(tui): archive eviction paths as logical lines so scrolled-off content reflows (#540)

griffinwork40:
Finishes #540 axis-2. PR #755 shipped only the RED guard that reproduces this defect and explicitly deferred the fix; this is that fix.

## The defect

All three eviction paths in `preserveRowsBeforeFrameRender` pushed committed content into native scrollback as pre-wrapped **PHYSICAL** rows. A terminal reflows only its *own* soft wraps, so app hard newlines can never rejoin on a width change — content scrolled off by the ordinary end-of-turn → next-turn sequence (commit under a short frame, then the next turn's thinking lane grows the frame) stayed hard-wrapped mid-word forever, indent lost.

The other two scrollback write sites — the band-hold archive (`committed-band-commit.ts`) and disarm's `flushPendingCommittedBand` (`lifecycle.ts`) — were already converted to logical lines. These three were the remaining gap, and #755's guard measured **4** non-wrapped rows after a widen where a rejoined line shows **1**.

## Mechanism of the fix

The old paths shared one mechanism whose single paint did **two** jobs: it put the evicted prefix into rows the scroll would carry into history, *and* left the survivors shifted up to hug the new frame top. That coupling is what forced physical emission — and a naive swap of only the emission is what regressed the verdict card's closing border on an earlier attempt: survivors were left unpainted while bookkeeping still claimed `committedBandPaintedRows === length`, so rows existed in the model but on neither the screen nor in scrollback.

New `archiveBandPrefixAndRepaintSurvivors` (`terminal-compositor.frame-preserve-archive.ts`) splits the two jobs:

1. **erase** the band's previous painted footprint from the geometric floor,
2. **archive** the prefix as logical lines — autowrap stays **ON**, so the terminal owns the intra-line wrap and its continuation rows are soft-wrap continuations that rejoin on resize,
3. **repaint** survivors explicitly — autowrap **OFF** (rows are already hard-wrapped to current width).

That ordering is externally governed by DECSTBM/DECAWM and is **not** reorderable; it is documented as a constraint comment at the site, per the repo's ordered-operation convention. All three paths delegate, so the erase floor, the 1:1 `committedBand` ↔ `committedBandMeta` slice, and the painted-rows accounting exist in exactly one place instead of three near-copies.

`overflow` is deliberately **not** snapped to a logical-line boundary the way band-hold snaps its own count: it is load-bearing geometry (`bandLen - room`), so snapping down leaves the band too tall to fit and snapping up strands blank rows (the void class). A straddling line is emitted verbatim by `scrollbackFlushLines`, so the on-screen tail is never duplicated.

The **legacy deficit-based eviction** at the end of the function is unchanged, and that is correct rather than an oversight: it scrolls whatever is already on screen (including pre-arm banner rows) rather than archiving band content, so it has no logical lines to emit. Verified in code, not just asserted in the header comment.

## Gates

| Gate | Result |
|---|---|
| `pnpm lint` (tsc --noEmit, strict) | **clean**, exit 0 |
| PTY `width-resize-fragment-evict-growth` | **pass** — filter confirmed applied (1 ran, 9 skipped) |
| `vitest run src/cli/verdict-card-overflow.test.ts` | **2/2 pass** — the regression that blocked the earlier attempt |
| full `pnpm test:pty` | **10/10 pass** |
| compositor + scrollback suites | **49 files / 397 tests pass** — no #539/#552/#573 height/void-class regression |

## Guard now asserts the fix, and is falsifiable

`minNonWrappedRows: 2` (RED, asserting the defect) → `maxNonWrappedRows: 1` (asserting the rejoined property).

Per `docs/scrollback.md:108-113` — two prior fixes in this domain shipped green while broken — a passing assertion is not evidence on its own, so the flip was **mutation-verified**: reverting step 2's emission to `committedBand.slice(0, overflow)` makes the guard measure exactly **4** and fail, reproducing the originally reported count. Restored and re-confirmed green afterward.

```
× width-resize-fragment-evict-growth  (physical emission)
  → logical line ["LOGSTART".."LOGEND"] should rejoin to <=1 non-wrapped row(s) but had 4
✓ width-resize-fragment-evict-growth  (logical archive)
```

## Cross-check vs #665 (`afk/tui-axis2-stage3`, verdicted DO-NOT-MERGE)

**Does not reintroduce either concern.**

- **Anchor-decrement invariant (`committed-band-commit.ts:469`)** — that file is not in this diff. The `scrolledRows` → `anchorRow` decrement lives at `committed-band-commit.ts:469`/`539-540` and is untouched. The new helper never reads or writes `anchorRow` (no reference in `frame-preserve-archive.ts`); it takes `floor` as a parameter, computed by the caller exactly as before (`Math.max(self.anchorRow ?? 1, 1)`, `frame-preserve.ts:184`). The only `anchorRow` mutation in `frame-preserve.ts` is the pre-existing legacy-deficit decrement at `:222`, inside the unchanged block.
- **Band-cap snap-to-zero** — not applicable by construction. That concern was about `snapFlushCountToLogicalBoundary` returning 0 for a straddling line (`committed-band-commit.ts:373-376`), silently archiving nothing. This helper deliberately does **not** snap: `overflow` passes through as exact geometry, so it can never round to zero, and the straddle case emits physical rows verbatim instead of dropping them (`frame-preserve-archive.ts:47-56`, `91`).

## Deferred

The full `turnModel` rewrite unifying all three sites onto one archive *model* — `docs/proposals/tui-compositor-rewrite.md` §5 "A2 — Unified retained model" (verified present, §5 exists at line 78/86). This PR shares the *mechanism*, not the model.

## What a reviewer should scrutinize hardest

1. **Step ordering** in `frame-preserve-archive.ts:69-126` — erase and archive share one write; survivors are a second write inside `withAutowrapDisabled`. The autowrap ON/OFF asymmetry between steps 2 and 3 is intentional and load-bearing in opposite directions.
2. **`committedBandBottomRow` when survivors is empty** — set to `floor - 1` at `:134`, which is the correct "empty band" encoding here but is an off-by-one-shaped expression worth a second read.
3. **`withFullScrollRegion` wrapping** covers the erase+archive write but not the survivor repaint (repaint is CUP-addressed and does not scroll) — `:102-106` vs `:119-125`.

Signed-off-by: Griffin Long <griffinwork40@gmail.com>

### DIFF
--- a/src/cli/terminal-compositor.frame-preserve-archive.ts
+++ b/src/cli/terminal-compositor.frame-preserve-archive.ts
@@ -1,0 +1,136 @@
+/**
+ * Logical-line scrollback archival for the frame-compositor eviction paths.
+ *
+ * #540 axis-2. All three of `preserveRowsBeforeFrameRender`'s eviction sites
+ * (the two `!hasBanner` paths and the banner path) previously shared one
+ * mechanism: paint the FULL band top-aligned, then `evictRowsToScrollback(N)`.
+ * That single paint did TWO jobs at once — it put the evicted prefix into rows
+ * the scroll would carry into history, AND it left the survivors shifted up to
+ * hug the new frame top. Elegant, but it forced the prefix into scrollback as
+ * pre-wrapped PHYSICAL rows, which a terminal can only ever reflow per-row: on
+ * a widen the content stayed hard-wrapped mid-word forever (the reported
+ * fragmentation, pinned by the `width-resize-fragment-evict-growth` PTY guard).
+ *
+ * This module splits those two jobs apart so the archive can emit
+ * SOFT-WRAPPABLE logical lines (via `committedBandMeta` + the shared
+ * `scrollbackFlushLines` / `buildScrollbackArchiveEscape` helpers, the same
+ * pair the band-hold archive and `flushPendingCommittedBand` already use) while
+ * the survivors are re-placed by an explicit paint rather than as a side effect
+ * of the scroll. Splitting them is what makes the conversion safe: a naive swap
+ * that only replaced the emission left the survivors unpainted while the
+ * caller's bookkeeping still claimed `committedBandPaintedRows === length`, so
+ * rows existed in the model but on neither the screen nor in scrollback — the
+ * historical symptom was a verdict card's closing border vanishing on a growth
+ * eviction (`verdict-card-overflow.test.ts` guards it).
+ */
+
+import type { FrameHost } from './terminal-compositor.frame.js';
+import {
+  buildScrollbackArchiveEscape,
+  eraseAndPaintRow,
+  scrollbackFlushLines,
+} from './terminal-compositor.types.js';
+import { withAutowrapDisabled } from './terminal-compositor.band-reflow.js';
+
+/**
+ * Archive the oldest `overflow` rows of the committed band to native scrollback
+ * as logical lines, re-place the survivors top-aligned at `floor`, and update
+ * the band bookkeeping to match.
+ *
+ * Contract — on return: `committedBand` / `committedBandMeta` have lost their
+ * oldest `overflow` entries (sliced in lockstep, preserving the 1:1 invariant);
+ * every surviving row is painted on screen at `[floor, floor + survivors - 1]`;
+ * `committedBandPaintedRows === committedBand.length` (no pending rows); and
+ * the archived prefix is in native scrollback. Callers must NOT repeat any of
+ * that. No-op guard: `overflow <= 0` returns without touching the terminal.
+ *
+ * `overflow` is a PHYSICAL row count and is load-bearing geometry — it is
+ * `bandLen - room`, the exact number of rows that must leave the screen for the
+ * survivors to land flush against the incoming frame top. It is therefore NOT
+ * snapped to a logical-line boundary the way the band-hold archive snaps its
+ * own count: snapping down would leave the band too tall to fit, and snapping
+ * up would strand blank rows between the band and the frame (the void class).
+ * A logical line straddling the boundary is handled by `scrollbackFlushLines`,
+ * which emits the in-prefix rows verbatim so the on-screen tail is never
+ * duplicated — that line stays fragmented (today's behaviour), while every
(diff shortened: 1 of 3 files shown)


griffinwork40:
## PR #756 Review

**Verdict: APPROVE**

No actionable correctness, compatibility, performance, security, or maintainability defects were found.

### Verified

- The required ordering—erase old footprint, archive/scroll, then repaint survivors—is preserved in `src/cli/terminal-compositor.frame-preserve-archive.ts:69-125`.
- Logical archive and survivor metadata remain synchronized through identical prefix slicing at `src/cli/terminal-compositor.frame-preserve-archive.ts:115-135`.
- Reflow runs at the current terminal width before either normal or picker rendering can invoke preservation logic, confirming the load-bearing width invariant at `src/cli/terminal-compositor.frame.ts:112-132`.
- The PTY regression assertion requires archived content to remain one rejoinable logical line after resizing rather than accepting fragmented physical rows.
- The extraction stays within the repository’s 350-LOC ceiling: the new module is 137 lines.
- Existing API behavior remains compatible; the legacy eviction path is retained for unaffected cases.

### Non-blocking follow-up

A focused unit test for `archiveBandPrefixAndRepaintSurvivors` would improve coverage of zero-survivor bookkeeping, metadata slicing, and erase/archive/repaint ordering. Current coverage is primarily through the PTY integration scenarios, so this is a test-hardening opportunity rather than a merge blocker.

No tests were run during this read-only review.

### Done
- **What was done:** PR #756 was reviewed across correctness, resize/reflow invariants, terminal operation ordering, API compatibility, performance, maintainability, and test coverage.
- **Evidence:** The reviewed patch is preserved at `/Users/griffinlong/.afk/state/skill-preflight/c4415be0-7f22-4a9f-a043-ad7de22a9033/pr-756.diff`, with the key ordering invariant independently confirmed in `src/cli/terminal-compositor.frame.ts:112-132`.
- **What changed:** No repository or external state was modified because `/review` is read-only.
- **Pending:** A direct helper-level unit test is recommended as non-blocking follow-up, and test execution was intentionally not performed during this review.

---

Issue #762, opened 2026-07-29
overload (529): mid-stream retry exhaustion aborts the whole session to 0 turns while the usage-limit path pauses — pause machinery is structurally unreachable


griffinwork40:
## Summary

A mid-stream `overloaded_error` (529) that exhausts its retry budget **falls through to the generic fatal path** and kills the session to `finalTurnCount: 0`, while the usage-limit path for a 429 **pauses and auto-resumes**. The asymmetry is not a deliberate policy choice — it is a **structural gap**: the pause machinery in `retry-layer.ts` keys on `classifyUsageLimitError`, which returns `null` for any error without an HTTP `status`, and a mid-stream overload has **no status by construction**. The overload error cannot reach the pause path even in principle.

Verdict: **DEFECT**, but the fix is *not* "make overload pause like usage-limit" — a 529 carries no reset signal, so an unbounded pause is a silent hang. See Recommendation.

## Incident

Session `3cc7a468-dbe5-41b1-a7e1-d2c7cd442f44` (`opus_1m`) accumulated 9 turns / ~2.02M cache-read tokens / $14.79, then aborted mid-stream on repeated 529s. **Five consecutive resume attempts all died the same way.** Zero assistant output; the transcript recorded `_(no response recorded)_`. The operator read the raw `{"type":"error",...,"type":"overloaded_error"}` envelope as a TypeScript error and could not recover the session across five tries.

Traces (`~/.afk/state/witness/<id>/trace.jsonl`):

| Session | Outcome |
|---|---|
| `3cc7a468-…` (parent) | `closure {reason:'abort', finalTurnCount:9, cacheRead:2023889}` — also contains 4 `usage_limit_pause` events, so the working pause path and the failing overload path are visible in **one file** |
| `70c9b0d5-…` | 18 events; 2 × (`loop_start` → 3 × `rate_limit{overloaded}` → `loop_end`), durations **151,965ms** and **257,958ms**, then `closure {reason:'abort', finalTurnCount:0, finalTokens:{}}` |
| `0f2bcdd0-…` | 3 × `rate_limit{overloaded}` then **no `loop_end`, no `closure`** — `session_sealed {status:'failed', incomplete:true}` **38 minutes later** |
| `4f37526f-…` | same shape — sealed `incomplete:true` **63 minutes later** |
| `6e7478a4-…`, `5732b874-…` | recovered only after the operator interrupted (`interrupt_halt`) |

[...]

PR #763, opened 2026-07-30
fix(provider): kill the unbounded post-first-byte round with a progress-aware stall watchdog (#762)

griffinwork40:
Part 1 of #762. **This one is a live silent hang, not a theoretical gap** — it is the piece worth landing first.

## The defect

Two real sessions ran **38.9 min** and **63.5 min**, then sealed `{status:'failed', finalTurnCount:0, incomplete:true}` with **no `loop_end` and no `closure` event at all**:

- `~/.afk/state/witness/0f2bcdd0-e609-4b2f-bc15-5a29308a9d17/trace.jsonl`
- `~/.afk/state/witness/4f37526f-3a79-4a56-aedc-8b0bf88ba30d/trace.jsonl`

`incomplete: true` is written **only** by `TraceWriter.sealOnProcessExit()` (`writer.ts:315-320`), the process-exit backstop — so those turns never returned. No terminal record, no error, no cost line. For an AFK operator that is the worst failure mode: silence is indistinguishable from work in progress.

**Cause:** the TTFB stall bound (`armFirstByteTimeout`, 180s) gates its retry on `ttfb.timedOut() && !aborted && !ttfbRetried && !ttfbEmitted` (`loop.ts:552`, `:629`), and `ttfbEmitted` flips true on the first CONTENT token (`:518-522`). Once a single token streams, **no bound of any kind applies for the rest of the round.**

## Why this is not an absolute round cap

`loop.ttfb.test.ts:179` deliberately pins that a stream yielding a first byte is **not** aborted "even if it then runs long" — and that invariant is *correct*. A large code emission or a long extended-thinking block must never be killed for taking a while. The missing distinction is **stalled vs. slow**, not "too long".

So this is a **sliding deadline that resets on every observable content token**: a round making progress survives indefinitely; a round producing no new output for the window dies loudly. One of the new tests pins exactly that difference — a round with 20s inter-token gaps over ~120s (4x the window) **survives**, which a naive absolute cap would kill.

`armStreamStallWatchdog` (new, `src/agent/providers/shared/stream-stall-timeout.ts`, 228 LOC) **mirrors** `subagent/idle-watchdog.ts` rather than reusing it: that watchdog consumes session-level `OutputEvent`s where this sees raw translated provider output; it aborts the subagent's own controller where this must abort a **per-request** one; and its tool-in-flight suspend logic is dead weight here, since tool dispatch happens strictly after the stream loop closes. It *does* reuse the trace vocabulary — `idle_watchdog_fired` with `metadata.source: 'model-stream'`.

**Invariant:** the watchdog aborts a per-request controller chained via `AbortSignal.any`, **never the caller's signal**. A stall must stay distinguishable from a user interrupt — aborting the caller's signal would seal the session `cancelled` instead of `failed`, misattributing an infrastructure fault to the operator. It is also dormant until the first translated event calls `progress()`, so pre-first-byte semantics are untouched and the post-first-byte window stays covered even when `AFK_MODEL_TTFB_TIMEOUT_MS=0`.

## The window is measured, not guessed

`loop_end.durationMs` was rejected as the proxy — it includes tool and subagent time (max observed 62h). The post-first-byte **stream** phase was measured as `model_ttfb` to the round's first `tool_call`, which excludes tool execution:

**n = 86,076 rounds** across 12,381 traces / 712,701 events — p50 **3.3s**, p95 **61.0s**, p99 **121.8s**, p99.9 **233.7s**, max **1,111,064ms (18.5 min**, hand-verified as a real healthy round).

Default **1,200,000ms (20 min)**: above the observed maximum *total* stream duration, so no historical round could trip it even in the worst case where its entire duration was one silent gap (~9.9x p99.9) — yet far below the 38.9/63.5-min hangs. Override with `AFK_MODEL_STALL_TIMEOUT_MS`; `0` disables, matching the convention `loop.ttfb.test.ts:226` pins for the TTFB bound.

## Verification

`loop.stall.test.ts` (5 tests) — loud termination with an operator-legible message naming the env escape hatch; per-request signal aborted while the caller's stays clean; the progress-survives case above; per-round re-arming proving `dispose()` runs on the clean path; and a session-level test pinning the end of the causal chain: provider `error` to `sawProviderError` (`agent-session.ts:696`) to `closure{reason:'abort'}` to seal `failed` with `incomplete` **undefined**, closure ordered before the seal. That last assertion is the direct inverse of the bug's signature.

**Verified load-bearing by mutation** (per `docs/scrollback.md:108-113` — prior fixes in this repo shipped green while broken, so a passing assertion is not evidence on its own): disabling the watchdog reproduces the defect as a **hang** — `Test timed out in 15000ms`, 1 failed / 4 passed — not an assertion failure. That *is* the #762 signature. Restored and re-confirmed green by md5.

**Gates:** `pnpm lint` clean · `loop.ttfb.test.ts` 5/5 incl. `:179` and `:226` · `loop.retry.test.ts` 21/21 unchanged · providers + trace **90 files / 1353 tests** passed · `audit:env:check` 0 violations · `scan:env:check` in sync (138 vars) · `audit:sdk:check` OK · new file 228 LOC (under the 350 ceiling).

## Sibling PR / merge ordering

A second PR for #762 will make **overload-retry exhaustion** non-fatal, touching `loop.ts` at the overload region (previously ~L569, L613-615, L646, L685) plus `query/retry-layer.ts`. That region is untouched here, **but the stall branches were inserted above it, shifting those line numbers by ~+30.** Whoever merges second should rebase and re-verify branch ordering: **the stall checks must stay first**, so a stall is never misclassified as a retryable overload.

## Not addressed

- `openai-compatible` has the same structural gap and is untouched.
- `loop.ts` remains a large-file offender (tracked in #360); this adds only the arm/dispose pair.
- Inter-token gaps are not recorded in traces, so the window derives from *total* stream duration — a conservative upper bound, which biases the default safely high.


griffinwork40 (commented):

## Merge decision: **MERGE**

No blocking findings. The fix does what its intent claims, the load-bearing invariant is genuinely pinned by tests, and the two findings below are non-blocking (one is parity with a pre-existing helper).

**Target:** `9323c54` (branch HEAD) · 7 files · +664/−12 · regime: full (security · api-compat · correctness · spec-compliance · test-coverage · perf-observability)

---

### Verified core claims

The central design claim — *sliding deadline, not an absolute cap* — holds mechanically:

- **Only real output resets the clock.** `stall.progress()` is called once, at `loop.ts:596`, inside the translated-event loop. `translate.ts:292` ignores `ping`/unknown SSE frames, so keep-alives never reach `progress()`. A wedged stream that emits only pings therefore still fires — the failure mode that would have defeated the whole fix.
- **Abort discipline is correct.** `fire()` aborts only its own private controller (`stream-stall-timeout.ts:200-202`); the caller's signal is never touched. `loop.ts:624` and `:725` both gate on `!input.signal.aborted`, keeping a stall distinguishable from a user interrupt — so a stall seals `failed`, not `cancelled`. Pinned by `loop.stall.test.ts:120,123`.
- **Branch ordering is as stated.** The stall check (`loop.ts:624`) sits above the mid-stream overload branch (`:629-645`), so a stall cannot be misclassified as retryable overload.
- **Disposal is complete on every exit path.** `stall.dispose()` at `loop.ts:504,508,704,708,744`. The `return` at `:725` is already covered by the `:708` dispose in the same `else` block; `dispose()` is idempotent (`stream-stall-timeout.ts:216`) and the timer is `.unref()`'d (`:211`), so even a missed path could not hold the event loop.
- **Test claims check out.** All five asserted invariants are really asserted, not adjacent-asserted: loud message naming the env var (`loop.stall.test.ts:114,116`), per-request abort with caller clean (`:120,123`), progress-survives (`:175-179`), `=0` disables (`:208,209`), per-round re-arming (`:295-297`), and the session-level chain incl. `incomplete` undefined plus closure-before-seal ordering (`:259,261,264`). The two invariants the PR promises not to break still stand unmodified at `loop.ttfb.test.ts:179` and `:226`.
- **api-compat: clean.** `idle_watchdog_fired` was not renamed or narrowed in `trace/types.ts` — the union member is unchanged context; only its JSDoc was rewritten. `metadata` is a generic `z.record(...)`, so the new `source`/`stallTimeoutMs` keys validate with no schema change, and old traces lacking `source` still parse.
- **No secret exposure.** The new trace payload (`loop.ts:459-468`) carries only `{source, stallTimeoutMs, elapsedSinceLastProgressMs}`; `stallTimeoutError()` is a static template with numbers. No model text reaches the witness layer.

---

### Findings

**medium · confidence:high · security/correctness · `src/agent/providers/shared/stream-stall-timeout.ts:102-108,209` · ref:9323c54 · citation-type:file-state**
**finding:** `resolveStallTimeoutMs` applies no upper clamp, so an `AFK_MODEL_STALL_TIMEOUT_MS` above Node's 32-bit `setTimeout` ceiling (2147483647ms ≈ 24.8 days) is silently coerced to **1ms**, which aborts every round a few milliseconds after its first content token — the exact inverse of the operator's intent, and reachable precisely by following this PR's own error text ("Raise AFK_MODEL_STALL_TIMEOUT_MS").
```
if (!Number.isFinite(n) || n < 0) return DEFAULT_MODEL_STALL_TIMEOUT_MS;
return n;
...
const t = setTimeout(() => fire(armedAt), timeoutMs);
```
Confirmed empirically on Node v24.11.0: a `setTimeout` of 2147483648 fired in 2ms with `TimeoutOverflowWarning`.
**suggestion:** clamp in `resolveStallTimeoutMs` — `return Math.min(n, 2_147_483_647);`. Non-blocking: `resolveTtfbTimeoutMs` (`first-byte-timeout.ts:41-47`) has the identical gap today, so this is convention parity, not a new class of defect; worth fixing in both.

**low · confidence:low · perf-observability · `src/agent/providers/shared/stream-stall-timeout.ts:182-183`, `src/agent/providers/anthropic-direct/loop.ts:453` · ref:9323c54 · citation-type:file-state**
**finding:** Chaining `armStreamStallWatchdog(ttfb.signal, …)` composes a second `AbortSignal.any` per round on top of the one `armFirstByteTimeout` already builds, doubling the number of composite signals transitively attached to the session-long `input.signal`, and `dispose()` clears only the timer — it never detaches the composition.
```
const controller = new AbortController();
const linked = AbortSignal.any([baseSignal, controller.signal]);
```
**suggestion:** on a long, many-round session, sample RSS to confirm composites are collected on the deployed Node build; if retention shows up, use the manual `addEventListener`/`removeEventListener` pattern already in `shared/abortable-stream.ts:51,70`. `[low confidence — verify with runtime context]` — the consequence is Node-build-dependent and not determinable from source; the sub-agent's attribution to specific upstream Node issue numbers was **not** verified and is excluded from this finding.

---

### Spec-compliance

**Stated intent:** PR #763 title + body — assessed. No unmet intent found: the sliding-deadline mechanism, the dormant-until-first-token property, the per-request abort target, the `0`-disables convention, and the measured default are each implemented and tested as described. No scope creep: the `env.ts` (+22/−1), `trace/types.ts` (+13/−6), and `docs/env-registry.*` edits are all the registration/documentation surface the new env var requires — the repo's env audit gates would fail without them.

One accuracy note in the author's favour: the reworded `AFK_MODEL_TTFB_TIMEOUT_MS` description (`env.ts:288-292`) correctly stops claiming the post-first-byte response "runs unbounded," which this PR makes false.

### What was not checked

- Read against branch HEAD `9323c54`; every citation above was verified at that ref (tree materialized read-only via `git archive`). No repo state was mutated.
- **Tests were not executed.** Test-coverage findings come from reading assertion source, not from a run. The author's mutation-verification claim (disabling the watchdog turns the test into a 15s hang rather than an assertion failure) is **unverified here** — it is the single most valuable claim in the PR body and is cheap for a maintainer to re-confirm.
- `openai-compatible` has the same structural gap and is untouched — disclosed non-goal, residual risk only.
- With `AFK_MODEL_TTFB_TIMEOUT_MS=0` the *pre*-first-byte phase remains unbounded. That is pre-existing opt-out semantics, not introduced by this PR, but it means "no bound of any kind" is still reachable by explicit configuration.
- Merge-ordering risk is the author's, restated: the sibling #762 PR touches the overload region shifted ~+30 lines by this change. Whoever merges second must rebase and re-confirm the stall checks still precede the overload branch — `loop.stall.test.ts` does not pin that ordering, so a bad rebase would fail silently.
- Not reviewed: Telegram/CLI surfaces, `docs/env-registry.json` byte-level schema conformance (delegated to the repo's `audit:env:check` gate, reported clean by the author).

griffinwork40:
## Review findings addressed in `c07cb52`

Both substantive findings are fixed; the third is deliberately declined with reasoning.

### 1. Timer overflow clamp — FIXED

Flagged independently by Codex (`stream-stall-timeout.ts:121`) and the automated sweep. Reproduced locally on **Node v24.11.0**: a `setTimeout` delay of `3_000_000_000` fires in **1ms** with `TimeoutOverflowWarning`, so an operator following this module's own error text — *"Raise `AFK_MODEL_STALL_TIMEOUT_MS`"* — past 2^31-1 would abort **every round milliseconds after its first content token**. That is a fresh instance of the same silent-failure class #762 is about, reached by following the remediation advice.

Both `resolveStallTimeoutMs` and `resolveTtfbTimeoutMs` now clamp through a shared `clampTimerDelayMs`. `0` and every in-range value pass through unchanged, so the disable escape hatch is untouched. The sweep was right that this was **convention parity, not a new defect class** — the TTFB resolver had the identical gap, so both are fixed together.

New `src/agent/providers/shared/timer-limits.ts` (34 LOC) holds the platform ceiling once rather than duplicating the magic number across the two bounds. It also gives the **three still-unclamped env resolvers** an obvious import target — `AFK_SUBAGENT_TIMEOUT_MS` (`subagent.ts:124`), `AFK_SUBAGENT_IDLE_TIMEOUT_MS` (`subagent.ts:175`), `AFK_TIMEOUT_MS` (`daemon-options.ts:18`) — all out of scope here and left for a follow-up.

### 2. Raw tool deltas as watchdog progress — FIXED

Codex's citation was byte-exact and the mechanism is real: `progress()` fired once per *translated yielded* event, while `input_json_delta` (`translate.ts:207-210`) and `signature_delta` (`:223-227`) accumulate **without yielding**. Between `tool.use.start` and `tool.use` a healthy tool call streaming its argument payload therefore produced zero observable progress and read as dead air.

`translateMessageStream` now takes an optional `onRawProgress` callback, invoked for exactly the content deltas that yield nothing — including `citations_delta` and unknown delta kinds, which had the same gap and were not named in the finding — wired to `stall.progress()` at the loop's call site.

**Pings still deliberately do not count.** The sweep specifically verified that a ping-only wedged stream still fires the watchdog ("the failure mode that would have defeated the whole fix"), so `onRawProgress` is scoped to `content_block_delta` and never to keep-alives or unknown top-level frames. A new test pins that inverse directly.

Severity note for the record: at the shipped 1,200,000ms default this was **not** reachable — the max *total* post-first-byte stream across the 86,076 measured rounds was 18.5 min. It mattered because it silently falsified the PR's central claim for anyone lowering the window.

### 3. `AbortSignal.any` composite retention — DECLINED

Verified structurally and not acting on it. Two composites per round, both constructed inside the loop body with no AFK-owned reference surviving `dispose()`; the only long-lived attachment is Node's internal dependent-signal set on `input.signal`, which is a pre-existing cost of the TTFB watchdog alone rather than a new leak class. The suggested alternative is also not viable at this call site: `messages.create(params, {signal})` requires a real `AbortSignal`, which the manual `addEventListener` pattern in `abortable-stream.ts:51,70` does not synthesize. The finding's own recommendation was to sample RSS if retention appears — observation, not a code change.

### Verification

**Both fixes mutation-verified**, per `docs/scrollback.md:108-113` (a passing assertion is not evidence on its own in this repo):

- Dropping the `input_json_delta` progress call fails the new tool-arg test with the genuine false-positive: `Model stream stalled: no output for 30s after the response had already started streaming` — during a *healthy* argument stream.
- Dropping `clampTimerDelayMs` fails both resolver tests with `expected 3000000000 to be 2147483647`.

Worth flagging honestly: **the first version of the tool-arg test passed under mutation.** Its mock stream ignored the per-request abort signal, so the watchdog fired correctly but had nothing to interrupt — the test proved nothing. Fixed to reject on abort like the SDK's iterator does, and the assertion on `requestAborted` is now the load-bearing one. This is exactly the failure the mutation step exists to catch.

Also re-confirmed the sweep's flagged-unverified claim from the original commit: disabling the watchdog turns `loop.stall.test.ts` into a hang, not an assertion failure.

**Tests +14** (2 new files). **Gates:** `pnpm lint` clean · providers + trace **92 files / 1367 tests** passed (1353 baseline + 14 new, none broken) · `loop.ttfb.test.ts` 5/5 with the `:179` and `:226` invariants unmodified · `audit:env:check` 0 violations · `scan:env:check` in sync (138 vars) · `audit:sdk:check` OK · new file 34 LOC.

**Merge ordering with #764 is unchanged** — this commit does not touch the overload region, and the stall checks still precede it.

griffinwork40:
## Review — PR #763 · `fix(provider): kill the unbounded post-first-byte round with a progress-aware stall watchdog (#762)`

**Target** · PR #763, `afk/iso-lane1-stall-watchdog-3-19qky5` → `main` · reviewed ref **`c07cb52`** · +1038/−14 across 13 files (3 new source/test modules) · change type: fix · regime: **full** (2 dimension agents + inline verification)
**Stated intent** · PR #763 title + body (11 explicit claims) — spec-compliance **assessed**.

---

### Verdict on the core change

The defect is real and the mechanism is sound. Verified independently at the ref: the watchdog is armed per round chained onto `ttfb.signal` (`loop.ts:453`), the composite is the signal handed to `messages.create` (`loop.ts:493` → `:181`), the caller's signal is never touched (`stream-stall-timeout.ts:192-193`), `0` disables via an inert handle (`:183-190`), the default is `1_200_000` (`:95`), `dispose()` runs on every exit path (`loop.ts:504, 508, 716, 720, 756`), and both stall branches sit **above** the overload re-drive (`loop.ts:636` before `:649`; `:737` before `:745`). **Claims 4, 5, 6, 11 HELD.** `trace/types.ts` is comment-only — no schema narrowing, no broken consumer.

No `critical` or `high` finding survived verification.

---

### medium

**medium · medium · correctness · `src/agent/providers/anthropic-direct/loop.ts:556` · ref:c07cb52 · diff-context**
The promptness wrapper that exists to make a halt deterministic is wired to the caller's signal only, so a fired stall reaches the loop solely via the SDK surfacing the abort from its parked SSE read — a transport that instead ends its iterator cleanly on abort (the pathology the same helper documents for `openai@6`) leaves the round parked with the watchdog already fired and disarmed.
```
        abortableStream(events, input.signal) as Parameters<typeof translateMessageStream>[0],
```
`abortable-stream.ts:16-23` confirms the abort *does* cancel the fetch and surfaces "only once its pending read settles" — so this is a promptness/coverage gap, not "the fix cannot fire". Every stall mock in the suite rejects on abort (`loop.stall.test.ts:220-229, 347-352`), so the swallow case is untested.
**Suggestion:** add one test whose mock ignores the abort and keeps the read parked; if it hangs, expose the stall-only controller and race that in `abortableStream`.

**medium · high · test-coverage · `src/agent/providers/anthropic-direct/loop.stall.test.ts:137-180` · ref:c07cb52 · diff-context**
The test the body names as the load-bearing proof of claim 2 cannot fail: its mock takes no signal and never checks abort, so a watchdog that wrongly fires at 30s keeps yielding, the turn completes, and all five assertions pass whether or not translated output resets the window.
```
      messages: { create: vi.fn(() => { callCount++; return slowButProgressingStream(); }) },
```
The sibling test 60 lines below documents this exact trap and applies the fix only there: *"Without this the mock would keep yielding through a fired watchdog and the test would pass whether or not the fix is present"* (`:193-195`). The raw-delta reset source **is** non-vacuously covered (`:183-275`); the translated-output source is not.
**Suggestion:** give `slowButProgressingStream` the per-request signal, reject on abort, and assert `requestAborted === false`, mirroring `:196`.

### low

**low · high · spec-compliance · PR body vs. `c07cb52` · ref:c07cb52 · file-state (verified inline)**
The description does not describe the revision under review: it claims `loop.stall.test.ts` has **5 tests** with mutation evidence "1 failed / 4 passed", but the file at the ref has **7** `it(` blocks (`:81, 131, 183, 276, 335, 381, 420`), and it discloses **one** new file (228 LOC) while the PR also adds `timer-limits.ts` + `timer-limits.test.ts` and changes shipped `AFK_MODEL_TTFB_TIMEOUT_MS` behaviour (`first-byte-timeout.ts:52`) — named in no claim 1–11. The head commit subject is `clamp env timer bounds and count invisible content deltas as stall progress`, i.e. the body predates it.
**Suggestion:** re-run the mutation check at `c07cb52`, restate the counts, and add the `timer-limits.ts` clamp to the body (the TTFB behaviour delta is strictly an improvement: an over-2^31 override degraded to a ~1 ms abort, now to 24.8 days).

**low · high · correctness · `src/agent/providers/anthropic-direct/loop.ts:731-737` · ref:c07cb52 · diff-context**
The comment justifying the throw-path branch names a mechanism that cannot occur: `abortableStream` receives `input.signal` (`:556`), which never aborts on a stall, so it never re-raises the marker.
```
        // (abortableStream re-raises the linked signal's abort reason, which
        // for a stall is the STALL_TIMEOUT_MESSAGE marker).
```
The branch still works via `stall.timedOut()`; only the stated causal chain is wrong. **Suggestion:** rewrite the parenthetical to name the real path (the SDK rejecting the aborted fetch) and keep `isStallTimeoutError(e)` as the defensive disjunct.

**low · high · correctness · `src/agent/providers/shared/stream-stall-timeout.ts:115-117` + `timer-limits.ts:32-33` · ref:c07cb52 · diff-context**
Any sub-second resolved window aborts every round milliseconds after its first content token, and two plausible operator inputs produce exactly that: `Number.parseInt('1e9', 10) === 1` and `Number.parseInt('1_200_000', 10) === 1`, both passed through by an upper-only clamp.
```
  const n = Number.parseInt(raw, 10);
  if (!Number.isFinite(n) || n < 0) return DEFAULT_MODEL_STALL_TIMEOUT_MS;
  return clampTimerDelayMs(n);
```
Confirmed absent: no `Math.max`/`MIN_` floor in either resolver or in `timer-limits.ts`. Failure is loud and names the var, so blast radius is bounded. **Suggestion:** validate `/^\d+$/` on the trimmed raw value before `parseInt` (same fix applies to `resolveTtfbTimeoutMs`, `first-byte-timeout.ts:50`).

**low · high · perf-observability · `src/agent/providers/anthropic-direct/loop.ts:460-467` + `src/cli/commands/trace.ts:489,495` · ref:c07cb52 · diff-context + file-state (verified inline)**
The trace event this PR emits to explain a killed round is invisible in the default `afk trace show` and loses its discriminator under `--all`: `idle_watchdog_fired` falls past every high-signal branch to `if (!ctx.showAll) return null` (`trace.ts:489`) and then renders as a bare `phase idle_watchdog_fired 20m` that drops `metadata.source` (`:495`).
```
      if (!ctx.showAll) return null; // latency waterfall — low signal by default
```
`rate_limit` (`:448`), `usage_limit_pause` (`:463`), and `compaction_disabled` (`:479`) each got a default-view branch for the identical stated reason — "it explains an otherwise-invisible stall". Emit-side, the metadata omits round ordinal and last event type, and `elapsedSinceLastProgressMs` is ≈ `stallTimeoutMs` by construction (`stream-stall-timeout.ts:205` measures from the last re-arm).
**Suggestion:** add an `idle_watchdog_fired` branch before the `showAll` gate rendering `metadata.source`, and add `lastEventType` (the sibling producer already carries it, `types.ts:703`).

**low · high · test-coverage · `src/agent/providers/anthropic-direct/loop.stall.test.ts:420-451` · ref:c07cb52 · diff-context**
The test named "re-arms per round: a stalled round after a clean tool-use round still fires" never reaches a second round — round 1 is a text stream ending `end_turn`, so `runTurn` returns and the stalling branch is dead fixture, leaving claim 8(d) (dispose-on-clean-path) unpinned; a leaked timer firing at +60s would abort an unobserved controller and produce no error event, so the assertions hold either way.
```
          return callCount === 1
            ? fromArray(makeTextStream('done'))
            : midStreamStallAfterContent(signal);
    expect(callCount).toBe(1);
```
**Suggestion:** use `makeToolUseStream(...)` for round 1 so a tool dispatch forces a genuine second request, then assert the second round errors `/stalled/i`.

**low · high · test-coverage · `src/agent/providers/anthropic-direct/loop.stall.test.ts:336-337` · ref:c07cb52 · diff-context**
Claim 5's headline scenario — the stall window still governing when the TTFB bound is disabled — has no test; the only `TTFB_KEY` assignment sets **both** bounds to `0`. Grep at the ref: `AFK_MODEL_STALL_TIMEOUT_MS` appears in exactly 5 files, and `loop.stall.test.ts` is the only test among them.
```
    process.env[STALL_KEY] = '0';
    process.env[TTFB_KEY] = '0'; // isolate: no other bound may fire
```
Held by inspection (`first-byte-timeout.ts:90-92` returns the base signal unchanged), untested. **Suggestion:** add a case with `TTFB_KEY='0'`, `STALL_KEY='60000'` asserting `requestAborted === true`.

### nit

- **api-compat** · `docs/env-registry.json` · the >2^31−1 down-clamp is documented in neither env entry, though it is the behaviour an operator raising the bound hits first. (The "0 disables everything" concern is refuted by the diff: the new TTFB text names `AFK_MODEL_STALL_TIMEOUT_MS` as the post-first-byte governor.)
- **correctness** · `stream-stall-timeout.ts:133-142` · `stallTimeoutError()` returns a bare prose `Error` that `isStallTimeoutError()` (marker-equality, `:124-126`) does not match, so any future consumer classifies by regexing prose; the mirrored precedent is a typed `IdleWatchdogError extends TimeoutError` (`src/utils/errors.ts:33`). No production consumer classifies provider stream errors today.
- **perf** · `stream-stall-timeout.ts:215-223` · `progress()` performs `clearTimeout` + `setTimeout` + `unref()` per token; unmeasured, and the mirrored precedent does the same.

### security · api-compat

`no issues found` — read `stream-stall-timeout.ts` (full), `timer-limits.ts` (full), `loop.ts:341-760`, `first-byte-timeout.ts`, `trace/types.ts` diff, `trace.ts:440-495`. `stallTimeoutError` interpolates only a rounded second count; `idle_watchdog_fired` metadata carries only `source`/`stallTimeoutMs`/`elapsedSinceLastProgressMs` + an already-emitted `resolvedModel` — no token, prompt, or user content on either surface. `onFire` is wrapped in `try/catch` (`:202-209`) and the emit is `void`. `trace/types.ts` is comment-only; `translateMessageStream` gains an **optional trailing** param with one production importer (`loop.ts:555`); `resolveTtfbTimeoutMs`'s signature is unchanged; the single new `process.env` read is a getter in `src/config/env.ts` with a matching `ENV_REGISTRY` entry.

---

### Dropped / downgraded in verification (Wave 1.5)

| Wave 1 claim | Status | Basis |
|---|---|---|
| `high` — "the loop stays parked forever, the fix cannot fire" | **downgraded → medium** | `abortable-stream.ts:16-23`: the abort *does* cancel the fetch; the SDK surfaces it with lag, not never |
| "unlike `IdleWatchdog.arm`, which floors a bogus window (`idle-watchdog.ts:221`)" | **dropped (misattributed)** | `:209` is `Math.max(idleTimeoutMs, untilResetMs + SLACK)` — a usage-limit-pause *extension*, not an operator-input floor |
| "unlike `ttfb_timeout`, which got a dedicated default-view branch" | **dropped (false-absent inverse)** | no `ttfb_timeout` phase exists in `trace.ts`; finding re-grounded on `rate_limit`/`usage_limit_pause`/`compaction_disabled` |
| `low` — "`0` no longer disables all provider-side aborts, undocumented" | **downgraded → nit** | refuted by the diff text itself: the new TTFB description names the stall var as the post-first-byte governor |
| `medium` — per-token timer churn | **downgraded → nit** | unmeasured; mirrored precedent identical |
| `medium` — parseInt truncation (as *security*) | **re-dimensioned → low correctness** | operator misconfiguration with a loud, self-naming failure; not unauthorized access or data loss |
| "dispose() on all **six** exit paths" | **corrected** | five: `loop.ts:504, 508, 716, 720, 756` — conclusion unchanged |
| All agent `file:line` citations | **remapped** | both agents cited *diff-artifact* line numbers; every citation above is re-anchored to real file lines at `c07cb52` |

### What was not checked

- **Citations verified inline against branch HEAD `c07cb52`** (present in the local object store; the branch is not checked out — the working tree is `main`@`883b7f1`).
- **No shadow-verify wave dispatched**: no `critical`/`high` finding survived Wave 1.5, so the post-synthesis gate did not trigger. The one `high` candidate was re-derived inline against `abortable-stream.ts` before downgrade.
- **Synthesis ran inline, not as a Wave 2 sub-agent** — the orchestrator held every verified line number and drop rationale, so delegating would have added a fabrication surface for zero context savings. Total dispatch: 2 sub-agents.
- **No commands run**: `pnpm lint`, `pnpm test`, `audit:env:check`, `scan:env:check`, `audit:sdk:check`, `loop.ttfb.test.ts`, `loop.retry.test.ts` — all gate claims in claim 9 are unverified.
- Not examined: `AgentSession.sendMessageStreamInternal` / `sawProviderError` closure-seal chain beyond the test's assertions; `createWithRetry` internals; the generator's outer `finally` and whether an abandoned generator disposes both watchdogs (timers are `.unref()`d, so no process-lifetime leak); the empirical dataset behind the 20-minute default; `openai-compatible` (disclosed out of scope); the sibling overload-retry PR's conflict surface.

---

## **MERGE**

Real defect, correct distinction (stalled ≠ slow), stated invariants hold under independent verification, and every surviving finding is `medium` or below. Recommended before merge (cheap, non-blocking): refresh the PR body per **L1** so the record matches `c07cb52`, and de-vacuate the progress-survives test per **M1** — the one-line fix already exists 60 lines below it in the same file. **M2** and the observability gap are follow-up issues, not merge blockers.

**Done**
- Reviewed PR #763 at ref `c07cb52` in full regime: 2 parallel dimension sub-agents, then inline citation + absence-claim verification that remapped every citation to real file lines, dropped 2 fabricated comparisons, and downgraded 4 findings.
- Evidence: this report; diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/a6594574-7b61-4dc7-9a88-085df67c1936/pr-763.diff`; all citations re-readable via `git show c07cb52:<file>` in `/Users/griffinlong/Projects/open_source/agent-afk`.
- World state unchanged — read-only review: no files written, no commits, no `gh` mutations, no PR comment posted. Working tree still `main`@`883b7f1`, clean.
- Deferred: gate claims (lint/tests/audits) were not executed, and the two `medium` findings are reported rather than fixed, per the skill's read-only constraint.

---

PR #764, opened 2026-07-30
fix(provider): preserve the turn and bound a pause when mid-stream overload retries exhaust (#762)

griffinwork40:
Fixes the fall-through-fatal half of #762: a mid-stream `overloaded_error` (529) that exhausts its retry budget no longer kills the session to `finalTurnCount: 0`.

## The defect

On the 4th mid-stream overload the guard `overloadRetries < OVERLOAD_MAX_RETRIES` goes false, the branch does not match, and control **falls through** past the interrupt check and the `StreamIncompleteError` branch into the unconditional tail `yield out.event; translatorErrored = true`. There was no "overload exhausted" branch at all — a 529 exited through the same lane as an auth failure, setting `sawProviderError` → `closure {reason:'abort'}` → seal `failed`, with the turn text never committed. One incident lost a 9-turn / ~2.02M cache-read / $14.79 session across five consecutive failed resumes.

The existing pause machinery was structurally unreachable, not merely mistuned: a mid-stream 529 is `new APIError(undefined, <parsed SSE body>, …)` with `status === undefined`, and `usage-limit.ts:111` is `if (!('status' in error)) return null;`. So this adds a **new classification arm**.

## What landed (all three priorities)

**1. Turn preservation.** Exhaustion routes to a CLEAN terminal — `turn.completed` stamped `OVERLOAD_EXHAUSTED` (`loop.ts` post-loop handler) — which commits the turn (`turnCount++`), so `afk --resume <sessionId>` restarts from saved state. The failure stays loud: `closure-reason.ts` maps the sentinel to an `abort` closure and `closure-emitter.ts` to a `failed` seal, exactly the `tool_use_loop_capped` → `iteration_cap` precedent. A display-only `assistant.message` replaces the raw `{"type":"overloaded_error"}` envelope operators were reading as a TypeScript error (it is NOT pushed into history — operator context, not model context).

**2. Bounded pause.** New outermost `turnWithOverloadPause` tier in `query/retry-layer.ts` keys on the sentinel via `classifyOverloadExhaustion` (`overload-pause.ts:117`) — deliberately *not* on an `error` event, since loosening the `status` check would let unrelated status-less errors into the 2-hour subscription park. Because a 529 carries no reset timestamp, the ceiling is plain **wall clock**, re-probing on a jittered 60–120s interval:

| Surface | Ceiling | Rationale |
|---|---|---|
| cli / repl / telegram | 10 min | operator is watching, can ESC |
| daemon / cron / subagent / unknown | **0 — fail fast** | an always-on runner that silently parks on capacity is strictly worse than one that fails and notifies |

`AFK_OVERLOAD_PAUSE_MS` overrides both the gate and the ceiling for every surface (`0` = pure fail-fast everywhere; positive = opt a daemon in deliberately). **Surface signal**: I used the EXISTING `AgentConfig.surface` — the same value stamped as `origin` on `session_init_start` — already present at `AnthropicDirectProvider.surface` (`index.ts:202/315`) and merely plumbed one level further (index.ts → query.ts → RetryLayer). No new signal invented.

**Every exit path re-yields the preserved terminal**, so the session always seals with a real `closure`; a pause ending in silence would just re-create the 38-and-63-minute hangs the issue documents. New `overload_pause` / `overload_resume` trace phases (union in `trace/types.ts` + array/zod enum in `trace/events.ts` + the parity-guard mirror list) keep it distinct from `usage_limit_pause`.

**3. Jitter.** The `5s → 10s → 20s` ladder was fully deterministic, so parallel subagents retried in lockstep (trace `3cc7a468`: three subagents at `attempt:1|2|3` inside one second). Both the in-loop ladder and `createWithRetry` now use `jitterBackoff` — additive ≤25%, never shortening the documented minimum — following the transient-429 jitter already in `retry-layer.ts`.

## Tests rewritten (pinned contract changed)

- `loop.retry.test.ts` "exhausts the retry budget on a persistent mid-stream overload **and yields the error**" → "**and preserves the turn**". It previously pinned the fatal contract; it now asserts no `error` event, a `turn.completed` carrying `OVERLOAD_EXHAUSTED`, and the human-readable notice.
- Same file, "exhausts retries and yields error on persistent 529": timer advance widened 40s → 60s because jitter pushes the worst-case ladder to 43.75s. Contract unchanged.
- Abort-during-backoff tests (`:392` / `:223`) kept green **unchanged**.
- Added: exactly-one-terminal, non-overload-still-fatal, 13 policy unit tests, 7 pause-tier tests (including abort-during-pause and close-during-pause).

## Mutation check (tests proven load-bearing)

- Revert the `loop.ts` exhaustion branch → `loop.retry.test.ts` **2 failed | 21 passed**; restore → **23 passed**.
- Remove the daemon fail-fast gate in `resolveOverloadPauseCeilingMs` → **3 failed | 17 passed** across the two new files; restore → all pass.

## Gates

```
pnpm lint                                    → exit 0
pnpm test …/loop.retry.test.ts               → 23 passed
pnpm test …/loop.ttfb.test.ts                → 5 passed
pnpm test src/agent/trace/session-phase.test.ts → 49 passed
pnpm test src/agent/session/closure-reason.test.ts   → 17 passed
pnpm test src/agent/session/closure-emitter.test.ts  → 26 passed
npx vitest run src/agent/providers/ src/agent/trace/ → 91 files / 1371 tests passed (baseline 90 / 1353)
pnpm audit:env:check   → 703 files, 0 violations
pnpm scan:env:check    → in sync (138 vars)
pnpm audit:sdk:check   → OK, lock matches
```

## ⚠ Merge ordering vs #763

#763 (same issue) adds a progress-aware stall watchdog to `loop.ts` and inserts stall branches **above** the overload region, shifting line numbers ~+30. **This PR needs a rebase on `origin/main` after #763 lands.** The conflict surface is small and mechanical — my `loop.ts` hunks are at `@@ +74`, `+174`, `+499`, `+585`, `+682`, `+721/+727`, `+796`.

I preserved #763's stated invariant: **the stall checks must stay FIRST**, so a stall is never misclassified as a retryable overload. My exhaustion branch sits *below* the existing overload retry guard and above the generic terminals — it never precedes a stall check. I did not touch the stall watchdog or `src/agent/providers/shared/stream-stall-timeout.ts`.

## Risks / not checked

- **Not implemented (issue rec. 3, deliberately out of scope):** the absolute per-round wall-clock bound for the post-exhaustion request (`!ttfbEmitted` gate). That is the fix for the `incomplete: true` seals and overlaps #763's watchdog directly — folding it in would collide.
- The pause tier is verified with a mocked `runTurn`, not against a live 529. Real-capacity behavior (does a 60–120s probe actually recover?) is unverified by construction — no way to synthesize upstream overload.
- Turn preservation is verified at the provider-event boundary (`turn.completed` + sentinel + the `turnCount++`/seal mapping unit tests). I did **not** run an end-to-end `afk --resume` against a real overloaded session, so "resume restores the 9 turns" is verified by contract, not by field replay.
- `deriveSealStatus` now reads `lastStopReason`; a future caller that drops that field would silently seal `succeeded`. The new closure-emitter test pins it.


griffinwork40 (commented):

**Merge decision: MERGE** — conditioned on the rebase ordering the PR body itself declares (after #763). No critical or high findings survived citation verification. One medium finding below is a real gap against this PR's own stated invariant; it is not merge-blocking because a terminal still reaches the consumer by another path, but it is worth a follow-up.

Reviewed against branch HEAD `f4ef0a0`. Regime: full (18 files, +896/−14).

---

### medium · confidence high · correctness + spec-compliance
`src/agent/providers/anthropic-direct/query/retry-layer.ts:331-332` · ref: `f4ef0a0` · citation-type: file-state

**Finding:** the post-probe-sleep exit is the one path in `turnWithOverloadPause` that returns without re-yielding the preserved `exhausted` terminal, so an abort or `close()` landing during the park drops the `OVERLOAD_EXHAUSTED` stopReason and its accumulated usage.

```ts
      await sleepWithAbort(nextProbeDelayMs(), runInput.signal);
      if (isClosed() || runInput.signal.aborted) return;
```

Every sibling exit yields first — fail-fast at `:296-299` (`yield exhausted; return;`) and ceiling at `:301-313` (`yield exhausted;` at `:311`). This directly contradicts the function's own header at `:255-259` and the PR body's claim: *"Every exit path re-yields the preserved terminal, so the session ALWAYS seals with a real `closure`."* As a spec item this is an **unmet intent**, not merely a latent bug.

**Consequence is mitigated, not absent** — worth stating plainly so severity is not overread. On the abort sub-case the consumer synthesizes a fallback terminal at `query.ts:514-516` (`makeInterruptedTurnEvent()`, `query.ts:636-642`), so a `turn.completed` still commits the turn and a closure still fires. What is lost is fidelity: the turn seals `stopReason: 'interrupted'` instead of `overload_exhausted`, so the `closure-reason.ts` → `abort` and `closure-emitter.ts` → `failed` mapping this PR added never runs for that path, and the synthesized event carries no accumulated usage (token accounting for the turn is dropped). On the `close()` sub-case `query.ts:513` returns silently, which is the documented teardown contract. Notably the operator-facing notice is *not* lost — `loop.ts:825-829` yields it before the terminal, so it is forwarded through `:287` prior to the break.

**Suggestion:** change `:332` to `yield exhausted; return;` to match the two sibling exits, and emit the `overload_resume` phase with `outcome: 'aborted'` on that branch so the trace bracket closes.

---

### low · confidence high · test-coverage
`src/agent/providers/anthropic-direct/query/overload-pause-tier.test.ts:169-195` · ref: `f4ef0a0` · citation-type: file-state

**Finding:** the two tests guarding exactly this path assert only the replay count, never the emitted events, so they pass whether or not the terminal is re-yielded — which is why the gap above is not caught.

```ts
    expect(runTurnMock).toHaveBeenCalledTimes(1);   // :180 abort-during-pause
    expect(runTurnMock).toHaveBeenCalledTimes(1);   // :194 close-during-pause
```

Both tests `await promise` (`:177`, `:193`) but never inspect the drained events for `.type`, `.length`, or `stopReason`. They pin "no second replay happened," not "a terminal reached the caller."

**Suggestion:** assert the drained array ends in a `turn.completed` carrying `OVERLOAD_EXHAUSTED` in the abort test, mirroring the ceiling test's assertion at `:160-163`.

---

### What verified clean

- **No double terminal.** `loop.ts:605-608` and `:647-649` set mutually exclusive flags and each `break`; the post-loop `if (overloadExhausted)` returns at `:835` before `if (translatorErrored)` at `:838` is reached.
- **Deadline cannot drift.** `startedAt` is captured once at `:266`, outside the `for(;;)` at `:269`, with no reassignment in the tier — the park is genuinely wall-clock bounded.
- **Unknown surface fails safe.** `resolveOverloadPauseCeilingMs` (`overload-pause.ts:151-153`) treats `undefined` and any non-interactive surface as `0` (fail-fast), so a daemon cannot silently park.
- **Sentinel classification claim holds.** `classifyOverloadExhaustion` (`overload-pause.ts:118-120`) keys purely on `turn.completed` + stopReason, never on `status` — status-less unrelated errors cannot enter the park, as claimed.
- **Jitter is additive-only.** `jitterBackoff` (`:90-92`) adds `0…25%` and never shortens the documented minimum.
- **Trace phase parity is complete.** `overload_pause`/`overload_resume` are present in the `types.ts` union (`:706-707`), the `events.ts` enum (`:415-416`), and the `session-phase.test.ts` mirror list (`:291-292`) — no missed mirror.
- **Env registry is consistent.** `AFK_OVERLOAD_PAUSE_MS` agrees across `env.ts:450`/`:1412`, `docs/env-registry.json:518`, and `docs/env-registry.md:43`.
- **Test claims are accurate.** Counts confirmed at the ref: 13 policy tests, 7 tier tests, 23 in `loop.retry.test.ts`, including `emits exactly one terminal event on overload exhaustion` (`:383`) and `does not preserve the turn for a non-overload mid-stream error` (`:405`).
- **No secret exposure.** The pause/resume trace metadata carries only `source`, a fixed `reason` literal, `hasResetTimestamp`, `ceilingMs`, `surface`; no error body, headers, or tokens are spread in.
- **api-compat:** the rewritten pinned contract in `loop.retry.test.ts` has no production consumer depending on the old fatal-`error` behavior for overload.

### Merge ordering

The PR body states this needs a rebase on `origin/main` **after #763 lands**; #763 is still open. Landing this first will force the conflict onto #763 instead. `AFK_OVERLOAD_PAUSE_MS` has no upper bound (`overload-pause.ts:149`), which is consistent with every other numeric env var in this repo and is operator-set rather than attacker-controlled — noted, not filed as a finding.

### What was not checked

- Read against branch HEAD `f4ef0a0`; all blocking-tier citations re-read at that ref before quoting. The medium finding was independently re-derived on the main path (`query.ts:513-517`, `:636-642`, `trace/types.ts:698-707`), which is what downgraded it from a suspected `high` — the mitigation is verified, not assumed.
- Stated intent: PR #764 title + body — spec-compliance assessed against it. The absolute per-round wall-clock bound (issue rec. 3) is declared out of scope by the author and was not treated as unmet.
- No tests were executed and no build was run; this is a static read of the diff and the files at the reviewed ref. The author's reported gate results and mutation checks were taken as claims, not re-verified.
- Real-capacity behavior (whether a 60–120s probe actually recovers against a live 529) is unverifiable by construction, as the author notes.

griffinwork40:
Wave 1.5 + shadow-verify complete. Synthesizing.

---

# Code Review — PR #764

**`fix(provider): preserve the turn and bound a pause when mid-stream overload retries exhaust (#762)`**
`afk/iso-lane2-overload-nonfatal-4-uprrkk` → `main` · 18 files · +896/-14 · reviewed ref **`17e56bfe`**
Regime: **full** (2 dimension agents + inline citation/absence verification + 3-claim shadow-verify wave)

## Verdict: 🔴 DO NOT MERGE

Two independently-confirmed `high` findings **compose into a single reachable failure**: a forked subagent that hits a 529 parks the full 10 minutes it was explicitly designed *not* to park, and then hands its parent the "Anthropic is overloaded" notice as a **successful** phase result. In `mint`, that notice becomes the spec/plan/research artifact. Both halves are cheap, local, additive fixes — the design is sound, the wiring is not.

---

## Blocking

### 1 · `high` · high confidence · correctness + api-compat
`src/agent/tools/nesting.ts:196,251,444,481` · `src/agent/providers/anthropic-direct/index.ts:315` · ref `17e56bfe` · citation-type: `file-state` (verified inline)

**Every forked child's provider is constructed without `surface`, so the constructor default `'cli'` gives each headless fork the 10-minute interactive ceiling — the PR's `subagent → 0 (fail fast)` row is unreachable by construction.**

```
index.ts:315   this.surface = opts.surface ?? 'cli';
index.ts:1031  surface: this.surface,      // ← the gate reads the CTOR value…
nesting.ts:196 return new AnthropicDirectProvider({ ...providerOpts, readOnlyMemory: true });
```

`ChildProviderFactoryArgs` (`nesting.ts:30-59`) has no `surface` field; `providerOpts` (`nesting.ts:174-183`) carries only permissions/executors/`readOnlyBash`. `subagent.ts:940` inherits the parent's surface into **`AgentConfig.surface`**, which feeds `deriveOrigin` telemetry (`agent-session.ts:229`) and never reaches `resolveOverloadPauseCeilingMs`. The fallback path is equally exposed: `providers/index.ts:281-294` builds `ctorOpts` from only `customTools`/`canUseTool`.

The park is **not** truncated by the subagent idle watchdog (8 min, `subagent.ts:160`): every probe emits an `assistant.message` that re-arms it at `handle.ts:408`, so the full ~600 s elapses. `'subagent'` is declared in the `Surface` union (`awareness/types.ts:36`) but **never produced** in production — the only assignment is `runtime-snapshot.test.ts:259`.

*Both Wave 1 agents flagged this independently; shadow verifier CONFIRMED via independent re-derivation of the full construction call graph.*
**Blast radius:** `agent`-tool forks, skill forks, compose DAG nodes, background jobs, recon/phase-restricted children. **Not** top-level cli/repl/telegram/daemon.
**Fix:** default `this.surface` to `'unknown'` — which `resolveOverloadPauseCeilingMs` already resolves to `0` — instead of `'cli'`; optionally add `surface?: string` to `ChildProviderFactoryArgs` and forward it.

### 2 · `high` · high confidence · api-compat
`src/agent/subagent/result.ts:57-58` · status decided at `src/agent/subagent/handle.ts:209` · ref `17e56bfe` · citation-type: `file-state` (verified inline)

**`OVERLOAD_EXHAUSTED` was not added to `isIncompleteStopReason`, so an overload-killed fork resolves `status: 'succeeded'` with the operator notice as its answer.**

```
export function isIncompleteStopReason(stopReason: string | undefined): boolean {
  return stopReason === TOOL_USE_LOOP_CAPPED || stopReason === STREAM_INCOMPLETE;
}
```

Chain re-derived independently: `loop.ts:931-940` yields the notice then the sentinel terminal → `stream-consumer.ts:335-342` materializes the notice as a real `Message` → `handle.ts:434` sets `finalMessage` → **`handle.ts:472 if (finalMessage) return finalMessage`** short-circuits *every* salvage guard below it (buffered-partial, capped-fallback, zero-output, cancel) → `handle.ts:209` sets `'succeeded'` → `result.ts:221-228` returns it verbatim.

Consumers that accept it as a real result, none gating on overload: `mint/_phases/plan.ts:69`, `spec.ts:72`, `research.ts:69`, `heal.ts:136`, `ship.ts:75`, `parallelize-dispatch.ts:161`, `fork-result.ts:75-79`, `foreground-promotion.ts:289`, `dag-subagent.ts:175`, `bgsub.ts:130`. Only schema-gated phases (`build.ts:85`, `verify.ts:86`) catch it — incidentally, via parse failure, not by design. This re-opens for a new cause the exact class `result.ts:520-533` documents as fixed.

**Fix:** add `stopReason === OVERLOAD_EXHAUSTED` to `isIncompleteStopReason`, plus the matching arm in `annotateIfIncomplete`.

### 3 · `medium` (blocking-adjacent: 2-line fix, and it corrupts the forensic channel this PR exists to improve) · high confidence · perf-observability
`src/agent/providers/anthropic-direct/query/retry-layer.ts:285-292` + `loop.ts:931-940` · ref `17e56bfe` · citation-type: `diff-context`

**Each re-exhausting probe emits a false `overload_resume{outcome:'recovered'}`, a duplicate operator notice, and a redundant `overload_pause`.** The loop yields the notice `assistant.message` *before* the sentinel `turn.completed`; `classifyOverloadExhaustion` matches only `turn.completed`, so the notice falls through the `pauseEmitted` gate:

```
if (pauseEmitted) { …phase:'overload_resume', metadata:{ outcome:'recovered' }…; pauseEmitted = false; }
yield event;
```

Over a 10-minute park (~5–9 probes at 60–120 s) the trace reads as 5–9 recover/park cycles that never happened, and the operator sees 5–9 identical notices. This contradicts the code's own comment at `:282-284` ("Close the park with `overload_resume` exactly once") and the PR body's "one pause + one resume per parked turn".

⚠ **CONFLICT (resolved):** the Wave 1 correctness agent asserted "trace volume is well controlled: one `overload_pause` + one `overload_resume` per parked turn." I verified the emit order at `loop.ts:931-940` directly; the shadow verifier derived the same defect independently. The Wave 1 claim is wrong.
**Fix:** emit `overload_resume` only after the inner loop ends with `exhausted === null`, or exclude the notice: `if (pauseEmitted && event.type !== 'assistant.message')`.

---

## Non-blocking

| # | Sev | Dim | Location | Finding |
|---|-----|-----|----------|---------|
| 4 | medium | api-compat | `src/cli/commands/trace.ts:475-501` | `overload_pause`/`overload_resume` render only under `--all` (zero matches in `src/cli`), while sibling `usage_limit_pause` is deliberately promoted above the `showAll` gate as "the highest-signal stall of all" — a 10-min park is invisible in default `afk trace show`. |
| 5 | medium | security | `overload-pause.ts:146-150` | `AFK_OVERLOAD_PAUSE_MS` has no upper clamp: `99999999999` parks **any** surface, daemon included. The usage-limit analogue is hard-capped at 2 h. Add `Math.min(…, OVERLOAD_PAUSE_MAX_MS)`. |
| 6 | medium | correctness | `retry-layer.ts:307` | Ceiling checked *before* the sleep, so the park overshoots by one probe interval plus one turn duration — a 10-min ceiling parks ~12+ min. Clamp the sleep to the remaining budget. |
| 7 | medium | correctness | `retry-layer.ts:277-294` | A recovering replay re-yields the whole turn's already-forwarded events with no `stream.retry` reset, so surfaces render duplicated text. |
| 8 | low | correctness | `retry-layer.ts:272` | `startedAt` is wall-clock, captured at **turn** start not pause start: the initial turn's duration eats the ceiling budget, and a system-clock adjustment mid-park shifts it. |
| 9 | low | docs | `retry-layer.ts:261` vs `:337-338` | Docblock overclaims "Every exit path re-yields the preserved terminal" — the post-sleep abort exit is a bare `return`. **[was: medium correctness defect, now: refuted]** — see below. |
| 10 | low | test-coverage | `overload-pause-tier.test.ts:169-181,184-195` | Abort- and close-during-pause tests assert only `runTurnMock` call count and discard the drained events, so neither pins which terminal those exits produce. |
| 11 | low | test-coverage | `overload-pause-tier.test.ts:158` | `toBeLessThan(6)` for a 150 s ceiling with a 60–120 s probe band — mathematical max is 3, so the bound cannot detect a ceiling regression up to ~5 probes. |
| 12 | low | test-coverage | `overload-pause-tier.test.ts` (grep `type: 'error'` → 0 matches) | No test covers a non-overload error surfacing during a post-pause probe replay. |
| 13 | nit | test-coverage | — | No test constructs a provider through a fork path, so finding #1 is unguarded. |
| 14 | nit | perf | `retry-layer.ts:278` | `classifyOverloadExhaustion` runs per forwarded event; it short-circuits on `event.type` first, so the cost is one property compare. (Downgraded from Wave 1's `medium`.) |
| 15 | nit | convention | `loop.ts` (1412 LOC), `retry-layer.ts` (689 LOC) | Both exceed your global 350-LOC hard ceiling; this diff adds +87/+119. Not mechanically gated in this repo (no `check-file-size` script in `scripts/`), and pre-existing for `loop.ts`. The PR did correctly extract `overload-pause.ts` at 154 LOC. |

**Shadow-verify refutation (finding #9):** the asserted consequence — "an operator who ESCs during the park loses the turn" — is **REFUTED**. `query.ts:503-518` treats a terminal-less clean return as an anticipated case and synthesizes `makeInterruptedTurnEvent()` (`query.ts:645-651`, `stopReason:'interrupted'`) → `turnCount++` still fires at `agent-session.ts:685-687`, the turn commits, resume stays intact, and `closure-emitter.ts:94-103` deliberately seals `cancelled` (documented as "more-specific" than `failed`). Pre-PR the usage-limit tier had the identical bare-return-after-sleep shape, so there is **no regression vs main**. What is actually lost: the `OVERLOAD_EXHAUSTED` label and that partial turn's usage/cost accounting. Downgraded `medium` → `low`, reframed as a comment inaccuracy.

**Dropped in verification:** one `medium` spec-compliance finding claiming the `AFK_OVERLOAD_PAUSE_MS` "number contract" silently accepts exponential/whitespace forms — not a defect (`Number.isFinite` correctly rejects `Infinity` and `1e400`; accepting `'6e5'` breaks nothing).

---

## Spec-compliance

**Stated intent: PR #764 title + body — assessed.** All three numbered priorities are genuinely implemented as described: turn preservation at `loop.ts:921-941` (notice is *not* pushed into `input.messages`, confirmed at `:919-920`), the `turnWithOverloadPause` tier as outermost at `retry-layer.ts:241/267`, jitter at `loop.ts:833` + `createWithRetry` (additive-only; worst-case ladder 43,750 ms < the widened 60 s test bound). **No scope creep** — every touched file serves a named clause.

One **unmet-intent** instance: the priority-2 surface table asserts `subagent → 0 (fail fast)`, which finding #1 shows is unreachable by construction.

**Two PR-body claims are stale, in the author's favor:** #763 has already landed in `main` (`219e0b63`), the merge-base is `5456b75e` (v5.83.4), and this branch already merged `origin/main` at `17e56bfe` — no rebase is pending. #763's ordering invariant is **preserved**: the stall checks sit first (`loop.ts:653`, `:778`), above the exhaustion arms (`:697`, `:787`). GitHub reports the PR `MERGEABLE`.

## What was not checked

- **Citations verified inline against branch HEAD `17e56bfe`** (== checked-out HEAD, clean tree). All `high` citations and all `file-state` citations re-read at the ref via `git show`; all absence claims re-derived via `git grep` at the ref. Zero fabricated citations; zero false-absent claims.
- **Wave 2 synthesis was run inline, not dispatched** — a deviation from the skill's structure. Rationale: the verification manifest and the shadow-verify results were all held in the orchestrator, and delegating synthesis risked paraphrase drift over verified literals. Wave 1 and shadow-verify were dispatched as specified.
- The Wave 1 correctness/spec/test/perf agent **hit its tool-use iteration cap** and returned a partial. Its four sections were complete, and its two highest-severity claims were independently re-derived by shadow verifiers, but I cannot rule out an unemitted finding in its tail.
- **No tests, typecheck, or build were run** — this review is read-only and static. The author's reported gate results (`pnpm lint` exit 0, 1371 tests, `audit:env:check`/`scan:env:check`/`audit:sdk:check` clean) and both mutation checks are **unverified**.
- Runtime behavior against a live 529 is unverified by construction (as the author states); no end-to-end `afk --resume` replay.
- Not reviewed: the OpenAI-compatible provider's own overload path (this gate is anthropic-direct only); the Telegram surface; whether `AFK_OVERLOAD_PAUSE_MS=0` is set in any shipped service env, which would mask finding #1 globally.
- The exact `docs/env-registry.md` var count was checked by diff delta (139→140), not by counting entries.

**Recommended path to merge:** three small additive fixes — `'unknown'` as the surface default, `OVERLOAD_EXHAUSTED` added to `isIncompleteStopReason`, and the `overload_resume` gate — plus a test that constructs a provider through a fork path. Findings #4–#8 are reasonable follow-ups.

**Done**
- Ran the full `/review` pipeline on PR #764: resolved the diff at ref `17e56bfe`, dispatched 2 parallel Wave 1 dimension agents, ran Wave 1.5 citation + absence verification inline against the ref, and dispatched a 3-claim shadow-verify wave; synthesized inline.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/9c8900bd-ca57-4775-b77a-3bbe7f2b31ec/pr-764.diff`; all citations re-read at `git show 17e56bfe:<file>`; session trace at `~/.afk/state/witness/<session>/trace.jsonl` records the 5 sub-agent dispatches.
- World state unchanged — review is read-only. No files written, no commits, no PR comments posted; the report above lives only in this transcript, so copy it into the PR yourself if you want it durable.
- Verdict: **DO NOT MERGE** on 2 confirmed `high` findings (headless forks get the 10-min interactive pause ceiling; overload-exhausted forks return `status:'succeeded'` with the error notice as their answer) that compose into a false-success path through `mint`, plus 1 cheap `medium` trace-correctness defect.
- Deferred: 1 Wave 1 agent hit its iteration cap (partial tail); no tests or typecheck were run, so the author's reported gates remain unverified.

---

griffinwork40:
Wave 1.5 complete — all citations resolved against `2ac2e31b`, all absence claims re-derived. Synthesizing inline (see deviation note).

---

# Review — PR #764

**`fix(provider): preserve the turn and bound a pause when mid-stream overload retries exhaust (#762)`**
`main ← afk/iso-lane2-overload-nonfatal-4-uprrkk` · ref `2ac2e31b` · 22 files, +1235/−17 · regime **full**

**Verdict: 0 critical · 0 high · 6 medium · 7 low · 3 nit · 1 finding dropped · 2 downgraded**

---

| # | Claim | Status |
|---|---|---|
| A4 | `retry-layer.ts:350` drops preserved terminal — **high** | **downgraded → medium**: abort half refuted (`query.ts:515-516` synthesizes `interrupted`; seal `cancelled`; pinned at `overload-pause-tier.test.ts:183-188`). Only `isClosed()` half survives. |
| B6 | Notice becomes model context — **high** | **downgraded → medium + low**: `loop.ts:920-926` already documents the true behavior correctly and deliberately. Surviving defects are the stale JSDoc (L1) and per-probe duplication (M2). |
| B9 | `overload-pause.test.ts:97-101` is a tautology | **DROPPED — refuted.** Removing the `parsed >= 0` guard returns `Math.min(Math.floor(-1), MAX)` = `-1`, not `0` → test fails under the mutation. |
| file-state ×6 | `sleep-with-abort.ts:19-26`, `stream-drive.ts:190`, `stream-consumer.ts:337-339`, `loop.ts:193-197`, `loop.retry.test.ts:186-190`, `trace.ts:448-459` | all **verified** verbatim |
| absence ×7 | no `OVERLOAD_EXHAUSTED` in openai-compatible · no `exactOptionalPropertyTypes` · no `vi.mock('./translate` · no `probeCount` · no `exhausted`/`attempt` in `trace.ts` · no zod validation in `trace.ts` · `terminal-sanitize.ts` unused there | all **confirmed-absent** |

---

## MEDIUM

```
medium · high · correctness · src/agent/providers/anthropic-direct/query/retry-layer.ts:350 · ref:2ac2e31b · citation-type:diff-context
finding:    INVARIANT VIOLATION against the file's own docblock — the post-sleep `isClosed()` exit
            returns without re-yielding the preserved terminal, so a close() during the park drops a
            fully-formed `turn.completed{OVERLOAD_EXHAUSTED}` the tier is already holding in `exhausted`,
            and the session seals `succeeded` for a turn that died of a 529.
evidence:   await sleepWithAbort(Math.min(nextProbeDelayMs(), remainingMs), runInput.signal);
            if (isClosed() || runInput.signal.aborted) return;
            // vs retry-layer.ts:261-265 docblock: "Every exit path re-yields the preserved terminal"
            // vs the PRE-sleep guard at :309-312, which DOES `yield exhausted` on isClosed()
suggestion: Split the conditions — `if (isClosed()) { yield exhausted; return; }` then keep the bare
            `return` for `signal.aborted` only; or amend the :261-265 docblock to state the two silent exits.
```
Chain verified end-to-end: `query.ts:874-876` (`close()` sets `state.closed` **and** aborts with reason `'closed'`) → `query.ts:514` (`if (this.state.closed) return;` — fires *before* the abort branch, so no terminal is synthesized) → `closure-emitter.ts:97,103` → `'succeeded'`, independently pinned by `closure-emitter.test.ts:159`.

⚠ **This is pinned as intentional** at `overload-pause-tier.test.ts:202-203` ("Same intentional silence as the abort exit above"). The abort exit's rationale is sound; it does **not** transfer to `isClosed()`, because `close()` never reaches the `interrupted` synthesizer. This is the one finding worth an explicit author decision before merge — either the yield or the docblock, not both as they stand.

```
medium · high · spec-compliance · src/agent/providers/anthropic-direct/query/retry-layer.ts:286 · ref:2ac2e31b · citation-type:diff-context
finding:    The operator notice is re-forwarded on every re-exhausting probe, so a 10-minute park pushes
            up to ~10 identical 4-sentence copies into `conversationHistory` (billed on the next turn and
            on --resume) and renders the same notice ~10 times in the REPL.
evidence:   // retry-layer.ts:291-294, the author's own comment:
            // "A probe that re-exhausts still forwards `OVERLOAD_EXHAUSTED_NOTICE` and any partial
            //  deltas BEFORE its sentinel terminal"
            // stream-consumer.ts:337-339: if (event.text) { ...; deps.conversationHistory.push(...) }
suggestion: Track a `noticeForwarded` flag in the tier and swallow the `assistant.message` whose text is
            `OVERLOAD_EXHAUSTED_NOTICE` on probes ≥2.
```

```
medium · high · correctness · src/agent/providers/anthropic-direct/query/retry-layer.ts:314-328 · ref:2ac2e31b · citation-type:diff-context
finding:    `ceilingMs` bounds the accumulated probe SLEEPS, not wall clock — the check runs before the
            sleep, so a probe entered with 1ms of budget left runs a full `runTurn` (3 in-loop overload
            retries, ≤43.75s jittered ladder, 4 create attempts, plus stream time) past the ceiling
            before the terminal surfaces, contradicting "plain WALL CLOCK" at :257-259.
evidence:   pauseStartedAt ??= Date.now();
            const remainingMs = ceilingMs - (Date.now() - pauseStartedAt);
            if (remainingMs <= 0) {
suggestion: Compose a deadline signal into the probe's `runInput` — `AbortSignal.any([runInput.signal,
            AbortSignal.timeout(remainingMs)])` — so the post-ceiling round is cut off, not run to completion.
```

```
medium · high · api-compat · src/agent/session/closure-reason.ts:40 · ref:2ac2e31b · citation-type:diff-context
finding:    LAYERING REGRESSION — three provider-NEUTRAL modules now import a cross-provider stop-reason
            sentinel from `providers/anthropic-direct/`, breaking the in-repo convention visible on the
            adjacent line, and leaving openai-compatible (which has its own mid-stream 529 budget) unable
            to stamp the sentinel without importing from a sibling provider.
evidence:   src/agent/session/closure-reason.ts:40  import { OVERLOAD_EXHAUSTED } from '../providers/anthropic-direct/overload-pause.js';
            src/agent/session/closure-emitter.ts:22  (same import)
            src/agent/subagent/result.ts:15          (same import)
            src/agent/subagent/result.ts:14  import { TOOL_USE_LOOP_CAPPED } from '../providers/shared/tool-loop-cap.js';
suggestion: Move `OVERLOAD_EXHAUSTED` + `OVERLOAD_EXHAUSTED_NOTICE` to `src/agent/providers/shared/`
            (alongside `TOOL_USE_LOOP_CAPPED` at `shared/tool-loop-cap.ts:34`) and re-export from
            `overload-pause.ts`. Confirmed absent from openai-compatible: 0 grep matches at the ref.
```

```
medium · high · perf-observability · src/agent/providers/anthropic-direct/loop.ts:928-936 · ref:2ac2e31b · citation-type:diff-context
finding:    The turn-killing exhaustion is invisible in `afk trace show` — it reuses `phase: 'rate_limit'`
            and carries its only distinguishing fields in `attempt`/`exhausted`, which the renderer at
            trace.ts:448-458 never reads, so it prints byte-identically to a routine mid-stream retry.
evidence:   phase: 'rate_limit',
            metadata: { reason: 'overloaded', source: 'mid-stream',
                        attempt: OVERLOAD_MAX_RETRIES, exhausted: true, },
suggestion: Branch on `md['exhausted'] === true` in the `rate_limit` arm and emit a distinct label
            (e.g. `overloaded EXHAUSTED after 4 attempts`). Confirmed absent: 0 matches for
            `exhausted|attempt` in src/cli/commands/trace.ts at the ref.
```

```
medium · high · correctness · src/agent/providers/anthropic-direct/loop.ts:193-197 · ref:2ac2e31b · citation-type:file-state
finding:    RESIDUAL GAP (not an intent violation — intent scopes to "mid-stream"): the sibling
            CONNECTION-phase 529 exhaustion still exits through the lossy lane this PR removes —
            `createWithRetry` rethrows and the turn yields a fatal `{type:'error'}`.
evidence:   if (isTransientServerError(e) && attempt < OVERLOAD_MAX_RETRIES) { continue; }
            throw e;
            // still pinned fatal: loop.retry.test.ts:186-190 asserts an `error` event
            // containing 'Overloaded' for a persistent connection-phase 529
suggestion: Route an exhausted `isTransientServerError` into the same `overloadExhausted = true` flag,
            or state in the PR body that connection-phase 529 remains fatal by design.
```

## LOW

```
low · high · api-compat · src/agent/providers/anthropic-direct/overload-pause.ts:46-48 · ref:2ac2e31b · citation-type:diff-context
finding:    The exported constant's JSDoc contract is false and is contradicted by this same PR's own
            loop.ts:920-926, so a maintainer reading the symbol wrongly concludes the string cannot
            reach model context.
evidence:    * `assistant.message` (never pushed into history — it is operator context, not
             * model context), mirroring the `stop_reason: 'refusal'` notice in `loop.ts`.
suggestion: Replace the parenthetical with loop.ts's accurate wording: "not retried into the dead
            request", not "operator-only". Same correction applies to the PR body's bullet 1.
```
- `low · spec-compliance · index.ts:214 + subagent/result.ts:69` — PR body under-declares two behavior changes. Both are **correct code** with sound in-file rationale, but the body claims *"No new signal invented"* while `declaredSurface` is a new provider field with distinct fork semantics (`index.ts:203-214`), and the body's three priorities omit the subagent-result reclassification (`incomplete: true` + `[⚠ PARTIAL RESULT]` banner). Fix: amend the description — this misdirects review away from the PR's riskiest semantic.
- `low · correctness · src/agent/providers/shared/sleep-with-abort.ts:19-26` (`file-state`, verified) — each probe adds an `abort` listener never removed on the timer path; a 10-min park reaches ~10 on one signal, grazing Node's `MaxListenersExceededWarning`. Bounded per turn; file is untouched by this PR (pre-existing, newly exercised).
- `low · correctness · retry-layer.ts:703 LOC / loop.ts:1418 LOC` (verified by `wc -l` at ref) — pre-existing 350-LOC hard-rule violation worsened by +132/+90; the new `overload-pause.ts` (166) took the constants but none of the control flow. Extract `turnWithOverloadPause` to `query/overload-pause-tier.ts`.
- `low · test-coverage · loop.ts:787-790` — the defensive re-thrown-overload exhaustion arm has no covering test (0 matches for `vi.mock('./translate` at the ref; every overload test drives the in-band path).
- `low · test-coverage · overload-pause-tier.test.ts:279-293` — no assertion pins notice hygiene on the recovery or multi-probe path, which is why M2 shipped unnoticed.
- `low · perf-observability · retry-layer.ts:330-342` — neither park phase carries a probe counter (0 matches for `probeCount|probes:` in `src/agent`), so a `ceiling-reached` park is indistinguishable from one that probed once.

## NIT
- `nit · security · trace.ts:491-506` — new renderers interpolate metadata via `String(...)` without `sanitizeForTerminal` (`src/utils/terminal-sanitize.ts` exists, 0 uses in `trace.ts`). Reachability low: `outcome` is a literal, `ceilingMs` is typed `number`, `surface` is operator config — and the pre-existing `rate_limit`/`usage_limit_pause` arms in the same function do the same.
- `nit · correctness · retry-layer.ts:314-315` — `Date.now()` base is non-monotonic; an NTP step back inflates `remainingMs`. Use `performance.now()` for the budget, `Date.now()` only for emitted `durationMs`.
- `nit · api-compat · index.ts:1059` — `surface: this.declaredSurface` forwarded unguarded where siblings spread-guard; a compile error the moment `exactOptionalPropertyTypes` is enabled (confirmed absent from `tsconfig.json`).

## Clean reads
Persisted-trace compat is additive and safe (`trace.ts` runs no zod validation on read; old files cannot contain the new phases). Parity-guard mirror list updated in all three places (`types.ts:715-716`, `events.ts:418-419`, `session-phase.test.ts:292-293`). `AFK_OVERLOAD_PAUSE_MS` registered correctly (`env.ts:470` + typed getter `:1451`). Daemon/cron/subagent/unknown ceiling really is `0` (`overload-pause.ts:163-165`). Override really precedes the surface branch (`:156-162`). Sentinel keys on `turn.completed`, never an `error` event (`:127-129`). Jitter is additive-only (`:99-101`). No credential reaches any new log/trace/display string. All five claimed test additions confirmed present.

## What was not checked
- Citations verified inline against branch HEAD **`2ac2e31b`** (== local checkout; PR files clean in worktree).
- **Stated intent: PR #764 title + body — spec-compliance assessed.**
- No tests, lint, or audit gates were run (read-only review); the PR's mutation-check claims are not reproducible read-only.
- No live 529, no end-to-end `afk --resume` replay. Whether the dropped `turn.completed` on the close path also loses the `turnCount++` is **unverified** — only the `succeeded` seal status is confirmed.
- `docs/env-registry.{json,md}` skipped as generated.
- `/shadow-verify` **not invoked**: after Wave 1.5 no `critical`/`high` finding remains. The two `high` claims were re-derived inline *with shell access* against the ref, tests, and full call chain — strictly stronger than a shell-less verifier could manage.
- **Deviation:** Wave 2 synthesis ran inline rather than as a sub-agent. Handing 15 already-verified findings to a shell-less agent adds paraphrase risk to verbatim evidence without adding verification capability.
- The PR body's *"needs a rebase after #763 lands"* warning is **stale** — `17e56bfe` already merged `origin/main` and resolved #763's conflicts; `gh` reports `MERGEABLE`. Worth deleting from the description.

---

# ✅ MERGE

No critical or high findings. Every medium is a refinement of new machinery that is strictly better than `main` (where this failure mode killed a 9-turn session to `finalTurnCount: 0`). Three are worth one commit before merge, all small: **M1** (one line or one docblock — the invariant as written is false), **L1** (the same false claim on an exported constant + in the PR body), **M4** (move the sentinel to `providers/shared/` — cheap now, load-bearing when openai-compatible needs it). The rest are legitimate follow-ups. Unusually strong in-code documentation throughout; `loop.ts:920-926` and `retry-layer.ts:289-294` are what made this review verifiable.

**Done**
- Reviewed PR #764 at ref `2ac2e31b` via the `/review` pipeline: 2 parallel Wave 1 dimension agents, Wave 1.5 citation + absence verification run inline against the ref, synthesis inline.
- Evidence: full report above; source diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/af92c5b7-eb1b-44f0-b36a-343d88bcbafb/pr-764.diff`; every finding carries a `file:line` re-read from `git show 2ac2e31b:<path>`.
- World changed: nothing. Read-only — no files edited, no commits, no PR comments, no pushes. Your 5 uncommitted files (`turn-handler.*`, `command-tags.*`, `turn-text-seam.ts`) were left untouched.
- Pending: findings are reported, not applied. 1 finding dropped as refuted (`B9`), 2 downgraded from `high` with reasons recorded in the manifest. `/shadow-verify` intentionally skipped — no surviving `critical`/`high`.

---

PR #769, opened 2026-07-30
fix(worktree): resolve the default create base at the calling session's anchor (#760)

### LINKED ISSUE #760: worktree tool: default base resolves HEAD at the MAIN checkout, not the calling session's worktree (contradicts design doc)
## Symptom

`worktree create` with no explicit `base` silently bases the new branch on the **main checkout's** HEAD, not on the HEAD of the session that called it. An agent running in worktree A that creates worktree B expects B to branch from A's tip; it branches from main's tip instead.

## Cause

`resolveCreateBaseRef` defaults to the literal string `'HEAD'` (`tools/handlers/worktree.ts:74-77`), and the create runs against the main checkout — `git -C <repoRoot> worktree add -b <branch> <path> <baseRef>` (`tools/handlers/worktree-managed.ts:95-97`), where `repoRoot` is deliberately resolved via `--git-common-dir` so it is always the main checkout (`worktree-managed.ts:41-50`). `HEAD` therefore resolves in the main checkout's working tree.

Observed in a real checkout: calling session's HEAD `b673c60`, main checkout's HEAD `4892dcf` — a default `create` produced a branch based on `4892dcf`.

## Why this is a bug rather than a tradeoff

`docs/proposals/first-class-worktree-isolation.md:107` states the intended behaviour as *"Base = HEAD of `this.currentCwd`'s repo"*. Shipped code never implemented that, and the mismatch is absent from that document's own "Known MVP limitations" list.

The worktree-safe `repoRoot` resolution is correct and should stay — it is only the **base-ref resolution** that should follow the caller.

Note `AFK_WORKTREE_BASE` does not apply here: it is wired only into the CLI `--worktree` path (`cli/commands/interactive.ts:332`, `cli/commands/interactive/worktree.ts:444`). The CLI's separate `origin/main` default exists to avoid staleness (`interactive/worktree.ts:588-592`), so its divergence from the tool is coincidental, not a mitigation.

## Suggested fix

Resolve the default base at the anchor rather than the main root — `git -C <anchor> rev-parse HEAD` — and pass the resulting SHA as `baseRef`, keeping `repoRoot` for the `worktree add` invocation itself. Add a test pinning "create from inside worktree A bases on A's HEAD, not main's".

griffinwork40:
Fixes #760.

## Problem

`git worktree add` must be invoked from the **main** repo root, so handing it the literal `HEAD` resolves the base against the **main checkout's** tip — not the caller's. A session working inside worktree A that created a new tree with no explicit `base` therefore silently got a branch based on main's HEAD.

Measured on this repo while writing the fix: an audit worktree sat at `b673c60` while main was at `a25984a`, so a default `create` from that session would have based its branch on the wrong commit.

Design intent is the opposite — `docs/proposals/first-class-worktree-isolation.md` line 107 specifies *"Base = HEAD of `this.currentCwd`'s repo"*, and the mismatch is not listed among that document's own MVP limitations.

## Change

- New `resolveAnchorBaseRef()` in `worktree-managed.ts`: resolves `HEAD` at the anchor and returns a **concrete SHA**, which is what makes the base unambiguous at the add site (that call runs with a different `-C`).
- Used for the default in **both** callers:
  - the `worktree` tool's `create` action, and
  - `createIsolatedWorktree()` — the `agent` tool's `isolation: "worktree"` path, which had the identical defect. This one is not mentioned in #760; it turned up while tracing the callers.
- An explicit `base` is passed through completely unchanged.
- The `git worktree add` invocation still runs from the main repo root. Only the base moved.

### Documented fallback

When the anchor has no resolvable HEAD (not a git repo, or an unborn branch) the literal `HEAD` is used — the pre-fix behaviour — so this cannot convert a previously working `create` into a failure.

## Tests

Added:
- an unspecified base follows the **anchor's** HEAD and not the main checkout's (the mock distinguishes the two checkouts by the `-C <dir>` each `rev-parse` used);
- an explicit `base` is honoured with no anchor resolution.

Three existing assertions were updated **deliberately**: the trailing token of `worktree add` is now the resolved sha rather than the literal `HEAD`. `worktree-managed.ts` carries an invariant that its git argv stay byte-identical because tests assert exact argv, so that change needed to be made knowingly rather than absorbed silently.

## Verification

| Gate | Result |
|---|---|
| `pnpm test src/agent/tools/handlers/worktree.test.ts` | 32 passed |
| `pnpm test src/agent/tools/handlers/worktree-managed.test.ts` | 12 passed |
| `pnpm test src/agent/tools/subagent/subagent-executor.isolation.test.ts` | 7 passed |

File sizes stay under the repo's 350-line-per-file ceiling: `worktree.ts` 343 (was 339 — the new logic deliberately went into `worktree-managed.ts`, 330, which had headroom).

### DIFF
--- a/src/agent/tools/handlers/worktree-managed.test.ts
+++ b/src/agent/tools/handlers/worktree-managed.test.ts
@@ -124,7 +124,7 @@ describe('removeManagedWorktreeGuarded — guards + argv', () => {
 });
 
 describe('createIsolatedWorktree', () => {
-  it('resolves the repo root and creates an afk/iso-* branch based on HEAD', async () => {
+  it("resolves the repo root and creates an afk/iso-* branch based on the anchor's HEAD", async () => {
     const mock = makeMock((call) => {
       if (call.args.includes('--git-common-dir')) return { stdout: `${repoRoot}/.git\n`, stderr: '' };
       if (call.args.includes('add')) {
@@ -138,10 +138,12 @@ describe('createIsolatedWorktree', () => {
     expect(iso.repoRoot).toBe(repoRoot);
     expect(iso.path).toBe(join(repoRoot, '.afk-worktrees', 'iso-diagnose-1-abc123'));
     expect(iso.branch).toBe('afk/iso-diagnose-1-abc123');
-    expect(iso.baseRef).toBe('HEAD');
+    // #760: the default base is the anchor's RESOLVED HEAD sha, not the literal
+    // ref (which `git -C <repoRoot>` would resolve at the MAIN checkout instead).
+    expect(iso.baseRef).toBe('headsha');
     const addCall = mock.calls.find((c) => c.args.includes('add'));
     expect(addCall?.args).toEqual([
-      '-C', repoRoot, 'worktree', 'add', '-b', 'afk/iso-diagnose-1-abc123', iso.path, 'HEAD',
+      '-C', repoRoot, 'worktree', 'add', '-b', 'afk/iso-diagnose-1-abc123', iso.path, 'headsha',
     ]);
   });
 
@@ -172,7 +174,7 @@ describe('createIsolatedWorktree', () => {
     expect(iso.repoRoot).toBe(repoRoot);
     expect(iso.path).toBe(join(repoRoot, '.afk-worktrees', 'iso-parallel-2-def456'));
     expect(iso.branch).toBe('afk/iso-parallel-2-def456');
-    expect(iso.baseRef).toBe('HEAD');
+    expect(iso.baseRef).toBe('headsha');
     // Retried EXACTLY once → `worktree add` invoked twice, second time won.
     expect(mock.calls.filter((c) => c.args.includes('add'))).toHaveLength(2);
   });
--- a/src/agent/tools/handlers/worktree-managed.ts
+++ b/src/agent/tools/handlers/worktree-managed.ts
@@ -59,6 +59,36 @@ export function sanitizeSlug(name: string): string {
     .slice(0, 80);
 }
 
+/**
+ * Contract: resolve the DEFAULT base ref at `anchor` — the calling session's own
+ * checkout — returning a concrete SHA.
+ *
+ * `git worktree add` must run from the MAIN repo root, so handing it the literal
+ * `HEAD` resolves against the MAIN checkout's HEAD rather than the caller's. A
+ * session working inside worktree A that requests a new tree with no explicit
+ * base therefore silently got a branch based on main's tip (#760), while the
+ * design intent is the opposite: `docs/proposals/first-class-worktree-isolation.md`
+ * line 107 — "Base = HEAD of `this.currentCwd`'s repo".
+ *
+ * Resolving to a SHA here (rather than passing a ref through) is what makes the
+ * base unambiguous at the add site, which runs with a different `-C`.
+ *
+ * Documented fallback: when the anchor has no resolvable HEAD (not a git repo, or
+ * an unborn branch) this returns the literal `'HEAD'` — the pre-#760 behaviour —
+ * so the change can never turn a previously working create into a failure.
+ */
+export async function resolveAnchorBaseRef(
+  execFile: ExecFileFn,
+  anchor: string,
+): Promise<string> {
+  try {
+    const out = await execFile('git', ['-C', anchor, 'rev-parse', 'HEAD']);
+    const sha = out.stdout.trim();
+    if (sha) return sha;
+  } catch { /* fall through to the literal ref */ }
+  return 'HEAD';
+}
+
 export interface CreateManagedWorktreeArgs {
   execFile: ExecFileFn;
   /** MAIN repo root (from {@link resolveRepoContext}). */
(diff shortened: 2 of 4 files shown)


griffinwork40 (commented):

## Merge decision: **MERGE**

Reviewed against branch HEAD `6d58449d` — all citations verified at that ref. Regime: **light** (hotfix, 4 files, +104/-14). No blocking, critical, or high findings. Two `low` findings, both non-blocking.

### Summary

The fix is correct and matches its stated intent. `resolveAnchorBaseRef()` resolves `HEAD` at the anchor to a concrete SHA, which is the right shape given the `add` itself runs `-C repoRoot`. Both default-base callers named in the PR body are fixed identically — `worktree.ts:218` and `worktree-managed.ts:257` — and no third site defaults a worktree base to the literal `'HEAD'`. Explicit bases pass through unchanged: `baseInput ?? await resolveAnchorBaseRef(...)` skips resolution whenever a string is present.

---

### FINDING 1 · low · confidence high · test-coverage
`src/agent/tools/handlers/worktree-managed.ts:84-89` · ref `6d58449d` · citation-type: file-state

**Finding:** the documented fallback branch — anchor has no resolvable HEAD, so the literal `'HEAD'` is returned — has zero test coverage in either test file at this ref. It is the one path in the change that preserves pre-#760 behaviour, and it is the path that regresses silently if the `try` boundary later narrows.

```ts
  try {
    const out = await execFile('git', ['-C', anchor, 'rev-parse', 'HEAD']);
    const sha = out.stdout.trim();
  } catch { /* fall through to the literal ref */ }
```

**Evidence of absence:** `git grep -n "resolveAnchorBaseRef\|unborn\|no resolvable HEAD" 6d58449d -- 'src/agent/tools/handlers/*.test.ts'` → zero matches (confirmed-absent).

**Suggestion:** add a test where the anchor's `rev-parse HEAD` rejects while `--git-common-dir` and `add` succeed, asserting the trailing argv element of the `add` call is the literal string `'HEAD'`. The empty-stdout branch (`if (sha)` falling through) deserves the same treatment.

### FINDING 2 · low · confidence high · security
`src/agent/tools/handlers/worktree-managed.ts:85-89` · ref `6d58449d` · citation-type: file-state

**Finding:** the `rev-parse` result is validated only as non-empty (`if (sha) return sha;`), not for SHA shape, before becoming the final argv element of `git worktree add`.

```ts
    const sha = out.stdout.trim();
    if (sha) return sha;
  } catch { /* fall through to the literal ref */ }
  return 'HEAD';
```

**Why this is low, not high:** the exposure is bounded on three independent counts. Invocation is argv-array `execFile` with no shell. `anchor` is not model-injectable — it resolves from `context?.resolveBase ?? context?.cwd ?? cwd ?? process.cwd()` (`worktree.ts:185`), a host-lifecycle-assigned value, not tool-call JSON. And a malformed ref is rejected by git at the `add` step, which the handler's try/catch turns into a normal `isError` response.

**Suggestion:** a `/^[0-9a-f]{7,40}$/` guard before returning `sha` converts a would-be `add` failure into a fail-fast at the resolution site with a clearer diagnostic. Optional.

---

### Verified clean

- **correctness** — the `typeof baseInput === 'object'` narrowing at `worktree.ts:214` is sound: the declared return type has no `null` member, because `base === null` returns `undefined` (`worktree.ts:77-79`), so the branch only ever catches `{ error }`. `anchor` is a genuine same-closure variable at both `resolveRepoContext(execFile, anchor)` (line 188) and `resolveAnchorBaseRef(execFile, anchor)` (line 218) — same value at both uses.
- **api-compat** — `resolveCreateBaseRef`'s widened return type (`string | undefined | { error }`) is module-private: `git grep -n "resolveCreateBaseRef" 6d58449d -- src` returns exactly 2 hits, the definition and its sole call site. Zero external importers. `createIsolatedWorktree`'s exported signature is unchanged at the type level (`baseRef?: string` was already optional); only the runtime default moved.
- **Downstream base semantics** — `createManagedWorktree` re-resolves `baseRef` through a second `rev-parse` for `baseSha` (`worktree-managed.ts:131-135`) regardless of whether it arrives symbolic or concrete. That call pre-dates this PR, so `managedWorktreeCommitsAhead` (`worktree-managed.ts:170-189`) and the sweep engine's `rev-list` staleness check (`worktree-sweep.ts:621-627`, reading `meta.baseSha`) are unaffected by the change in base shape. This was the main blast-radius risk and it is clean.
- **spec-compliance** — assessed against the PR title + body. No unmet clause; no scope creep. `anchor` is pre-existing and unmodified by this diff.
- **perf-observability** — one net-new `git rev-parse` subprocess per `create`, against an operation that already spawns `git worktree add` plus meta-file and lockfile I/O. Negligible.
- **File-size ceiling** — verified at the ref, not taken from the PR body: `worktree.ts` = 343, `worktree-managed.ts` = 330. Both under the repo's 350-line limit, matching the stated figures exactly.

### What was not checked

- Read against branch HEAD `6d58449d`; every citation above was verified at that ref.
- Stated intent: PR #769 title + body — spec-compliance **was** assessed.
- The PR's claimed test counts (32 / 12 / 7) and `pnpm lint` clean were **not** re-executed by this review. CI on `6d58449d` is green across Lint & Build, ubuntu, macOS, real-PTY-scrollback, docs, and publish-bundle-smoke, which is the substitute signal.
- `subagent-executor.isolation.test.ts` wholesale-mocks `worktree-managed.js`, so it does not exercise real anchor-vs-main HEAD divergence — that behaviour is covered only at the `worktree.test.ts` / `worktree-managed.test.ts` unit layer. No end-to-end test proves the `agent` tool's `isolation: "worktree"` path against two real checkouts at different tips.
- No runtime execution against a live multi-worktree repo; no review of the Telegram or daemon surfaces.

PR #770, opened 2026-07-30
fix(worktree): stop the sweep reaping live trees and discarding ignored files (#759)

### LINKED ISSUE #759: worktree sweep can force-remove a LIVE worktree, and the dirty guard is blind to ignored files (.env loss)
## Symptom

The worktree sweep can `git worktree remove --force` a worktree that is **in active use**, and its "is it safe to remove" check is blind to ignored files — so a tree holding an uncommitted `.env` (or any ignored-but-precious file) can read as clean and be destroyed.

Commits are never at risk (branch refs live in the shared `.git`). The loss is confined to working-tree files, which is exactly where the non-recoverable ones are.

## Two paths to reaping a live tree

**1. Missing meta → `'empty'`.** `.afk-worktree-meta.json` is written best-effort with every error swallowed (`worktree-managed.ts:106-122`), and a tree created by raw `bash: git worktree add` has no meta at all. No meta → `ownerLiveness = 'unknown'` (`worktree-sweep.ts:662-670`). Sub-agents write no presence file (`worktree-occupancy.ts:5-7`), so the rescue path at `worktree-sweep.ts:674-679` cannot fire. Clean + age ≥ `MIN_EMPTY_AGE_MS` (1 hour, `worktree-sweep.ts:165`) → verdict `'empty'` (`:326-333`) → `remove --force` (`:718`).

**2. Dead creator PID → `'dead-owner'`, reaped regardless of age** (`worktree-sweep.ts:309-315`, `:735`). Hits any tree whose creating session has exited while another process — another agent, or the operator's editor — still works in it.

Occupancy protection is touched **once**, at dispatch (`subagent.ts:986`), rewriting `pid` + `createdAt` and resetting `ageMs`. Any run longer than an hour therefore ages back into the reapable window mid-flight (already documented at `worktree-occupancy.ts:20-23`).

The dangerous caller is the daemon, which sweeps `scope: 'all'` (`daemon/scheduler.ts:497`); boot-prune uses `scope: 'interactive'` and skips meta-less trees.

## The dirty guard is blind to ignored files

The guard is bare `git status --porcelain` (`worktree-sweep.ts:602`, mirrored at `worktree-managed.ts:132`). Bare porcelain reports **untracked** files (`??`) — those do protect a tree — but **not ignored** ones. So `node_modules/`, `dist/`, `.env`, local scratch and build caches are invisible to it. Rebuildable losses are annoying; a worktree-local `.env` is not recoverable.

## Suggested fixes (smallest first)

1. Re-touch occupancy on a timer while a child runs (extend `subagent.ts:986`) so `ageMs` never crosses `MIN_EMPTY_AGE_MS` mid-run.
2. Treat a meta-less tree as `'unknown'`-but-protected rather than `'empty'` — absence of meta is absence of evidence, not evidence of abandonment.
3. Before `--force`, probe for ignored-but-non-rebuildable content (e.g. `git status --porcelain --ignored` with a rebuildable-path allowlist), or refuse `--force` on any tree with ignored content not matching that allowlist.

## Notes

Found during a read-only audit of the worktree subsystem; all claims are static reads of the cited lines plus verification of the porcelain argv. No fix attempted. Plausibly in scope for the #441 tracking issue, but filed separately because the failure mode is silent data loss rather than robustness polish.

griffinwork40:
Fixes #759.

Two independent paths by which the background sweep destroys data. Committed work is never at risk in either (branch refs live in the shared `.git`) — what is at risk is a live checkout and worktree-local ignored state.

## 1. A live worktree could be force-removed mid-flight

`classifyCandidate`'s `empty` verdict requires `ownerLiveness !== 'alive'`, a clean tree, no unreplaceable commits, and `ageMs >= MIN_EMPTY_AGE_MS` (1 hour). Occupancy was stamped **exactly once**, at dispatch (`subagent.ts`), and the sweep derives `ageMs` from `meta.createdAt` — so one touch bought only an hour of protection. Any subagent running longer than that aged back into `empty` while still working, and `git worktree remove --force` took its checkout out from under it.

**Fix:** `startWorktreeOccupancyHeartbeat()` re-asserts the occupation every 10 minutes (comfortably inside the 1h gate) for as long as the child runs.

Lifecycle details, since a stray timer is its own bug:
- the timer is `unref()`d, so a leaked one can never hold the process open;
- its stop handle is captured **before** the child session is constructed, so no exit path can arm a timer it has no handle to cancel;
- it is disarmed from the settle callback — which fires on success, failure, timeout **and** abort — and separately from the construction-failure path.

## 2. The dirty guard was blind to ignored files

`git status --porcelain` reports untracked files but **never ignored** ones. A worktree whose only content was ignored therefore read *clean*, satisfied `empty`, and was `remove --force`d — deleting it. `node_modules/` and `dist/` are merely annoying to lose; a worktree-local `.env` or an un-ignored-by-habit secrets file is unrecoverable.

**Fix:** new `worktree-ignored-probe.ts` treats ignored content a rebuild would **not** restore as dirty.

The judgment call here is the filter, and it matters in both directions. Protecting on *any* ignored entry would make effectively every worktree immortal — `node_modules/`/`dist/` exist in every checkout — which defeats the sweep and regrows the very leak it exists to prevent. So the default stays "reapable", and only an entry that fails to match a known-rebuildable pattern protects the tree. Unrecognised entries (`.env*`, `.vscode/`, `.idea/`, anything unfamiliar) and any git failure fail **safe**, toward protecting. The probe uses `--ignored` in traditional mode on purpose: it collapses an ignored directory to a single `!! node_modules/` line rather than listing every file underneath.

## Posture: deliberately unchanged

A **meta-less** tree stays reapable. The alternative — never reaping a tree without meta — trades a bounded data-loss bug for an unbounded leak, and worktrees already accumulate in practice. The two fixes above close the actual loss paths without that trade.

## Verification

| Gate | Result |
|---|---|
| `pnpm test src/agent/worktree-ignored-probe.test.ts` | 37 passed |
| `pnpm test src/agent/worktree-occupancy.test.ts` | 13 passed |
| `pnpm test src/agent/worktree-sweep.test.ts` | 43 passed |
| `pnpm test src/agent/subagent.test.ts` | 87 passed |
| `pnpm test src/agent/tools/subagent/subagent-executor.isolation.test.ts` | 7 passed |

New regression coverage: a clean tree holding `.env` is not reaped; a clean tree holding only build output still is; the heartbeat refreshes a backdated meta below the 1h gate, halts on `stop()`, is idempotent, and stays inert outside a managed tree.

## Two things for the reviewer

1. **File size.** The repo's 350-line-per-file convention: the new behaviour went into a new file (`worktree-ignored-probe.ts`, 114 lines) rather than inline, but the two pre-existing files still grew — `worktree-sweep.ts` 783 → 797 and `subagent.ts` 1138 → 1155, both already well over the ceiling. Extracting `classifyCandidate` + the verdict types into `worktree-sweep-classify.ts` is the obvious next cut (verified: those types have **zero** importers outside `worktree-sweep.ts`, and the suite drives everything through `runSweep`, so the move is safe). I deliberately left it out — it is a pure code move, and bundling one with a behavioural fix is how a move hides a regression. Happy to send it as a standalone follow-up.
2. **The rebuildable pattern list** is a heuristic and the place where a wrong call has real cost. Worth a second pair of eyes on what belongs in it — an over-inclusive entry silently re-opens the deletion path for that name.

### DIFF
--- a/src/agent/subagent.ts
+++ b/src/agent/subagent.ts
@@ -997,84 +997,94 @@ export class SubagentManager {
       stopOccupancyHeartbeat = startWorktreeOccupancyHeartbeat(childConfig.cwd);
     }
 
+    // Ordering constraint: the heartbeat armed above is disarmed by the settle
+    // callback installed on the handle built below, so the guarded span has to
+    // run from construction all the way through `active.set`. A throw anywhere
+    // in between — the parent-stream read, the sink resolve, the handle
+    // constructor — would otherwise return with the interval live and no handle
+    // in existence to ever cancel it, and the tree would never be reaped again.
     let session: AgentSession;
+    let handle: SubagentHandleImpl<T>;
+    let effectiveAgentType: string | undefined;
+    let effectiveResolvedAgentType: string | undefined;
     try {
       session = new AgentSession(childConfig);
+      const parentInputStreamRef = options.parent.getInputStreamRef?.();
+      const parentAbortSignal = options.parent.abortSignal;
+      // Resolve sink: explicit option takes precedence, then ambient from AsyncLocalStorage
+      const sink = this.progressSink ?? getCurrentSink();
+      // Normalize empty-string render hints to undefined so the `??` fallbacks
+      // engage. A caller passing `agentType: ''` or `parentId: ''` would
+      // otherwise produce an empty Agent() label / empty agentContext anchor
+      // rather than the intended fallback (idPrefix / parent.sessionId).
+      effectiveAgentType = options.agentType?.trim() || undefined;
+      effectiveResolvedAgentType = options.resolvedAgentType?.trim() || undefined;
+      const effectiveParentId = options.parentId?.trim() || undefined;
+      handle = new SubagentHandleImpl<T>(
+        id,
+        session,
+        childController,
+        this.abortGraph,
+        options.outputSchema,
+        // Wall-clock budget for the child's turn (see SUBAGENT_DEFAULT_TIMEOUT_MS
+        // above). Explicit caller values win, including `0` for unbounded and the
+        // background SUBAGENT_BACKGROUND_TIMEOUT_MS the SubagentExecutor stamps —
+        // `??` preserves that precedence. The default is env-tunable via
+        // AFK_SUBAGENT_TIMEOUT_MS (resolveSubagentTimeoutMs); an unset/invalid
+        // env value returns SUBAGENT_DEFAULT_TIMEOUT_MS, so behaviour is unchanged
+        // when the var is not set.
+        options.config.timeoutMs ?? resolveSubagentTimeoutMs(),
+        registry,
+        () => {
+          // Runs on every terminal outcome of the child — success, failure,
+          // timeout, and abort — so it is the settle hook the heartbeat's
+          // teardown belongs on.
+          stopOccupancyHeartbeat();
+          this.active.delete(id);
+          this.abortGraph.dispose(id);
+        },
+        parentInputStreamRef,
+        parentAbortSignal,
+        // agentType: explicit override → idPrefix fallback. Lets callers
+        // (e.g. compose) supply a render-only label without altering idPrefix
+        // which is also used for routing telemetry.
+        effectiveAgentType ?? options.idPrefix,
+        sink,
+        // parentId: explicit override → parent session id fallback. Lets
+        // callers (e.g. compose) anchor the renderer's nesting at the compose
+        // tool_use_id rather than at the orchestrator session id.
+        effectiveParentId ?? options.parent.sessionId,
+        // traceWriter: child shares the parent's writer so its SubagentStop
+        // hook decision lands in the same trace file. Contract:
+        // docs/philosophy/afk-contract.md — "a child sub-agent inherits its
+        // parent's witness." Resolved once above (per-fork config →
+        // manager-level writer) so the handle's terminal lifecycle emits
+        // pair with the 'started' emit below even when inheritance came
+        // from the manager rather than the per-fork config.
+        effectiveTraceWriter,
+        // onSubagentSucceeded: propagate completion data to the parent
+        // session's session_sealed rollup accumulators.
+        this.onSubagentSucceededCb,
+        // Progress-aware idle-watchdog window for the child's turn (see
+        // SUBAGENT_DEFAULT_IDLE_TIMEOUT_MS above). Explicit caller values win,
+        // including `0` to disable the watchdog for this fork — `??` preserves that
+        // precedence. The default is env-tunable via AFK_SUBAGENT_IDLE_TIMEOUT_MS
+        // (resolveSubagentIdleTimeoutMs); an unset/invalid env value returns
+        // SUBAGENT_DEFAULT_IDLE_TIMEOUT_MS, so behaviour is unchanged when the var
+        // is not set. Runs concurrently with the wall-clock budget above.
+        options.config.idleTimeoutMs ?? resolveSubagentIdleTimeoutMs(),
+      );
+      this.active.set(id, handle as SubagentHandleImpl<unknown>);
     } catch (err) {
-      // Construction failed (e.g. invalid model, sync init failure).
-      // Release the graph node that was registered above to prevent an orphan
-      // accumulating on repeated retries (e.g. forge/farm retry loops), and
-      // disarm the occupancy heartbeat — there is no child left to protect.
+      // Construction or manager-wiring failed (invalid model, sync init
+      // failure, a throwing parent-stream/sink read). Release the graph node
+      // registered above so an orphan cannot accumulate across retry loops
+      // (forge/farm), and disarm the occupancy heartbeat — there is no child
+      // left to protect, and a live timer here would pin the worktree forever.
       stopOccupancyHeartbeat();
       this.abortGraph.dispose(id);
       throw err;
     }
-    const parentInputStreamRef = options.parent.getInputStreamRef?.();
-    const parentAbortSignal = options.parent.abortSignal;
-    // Resolve sink: explicit option takes precedence, then ambient from AsyncLocalStorage
-    const sink = this.progressSink ?? getCurrentSink();
-    // Normalize empty-string render hints to undefined so the `??` fallbacks
-    // engage. A caller passing `agentType: ''` or `parentId: ''` would
-    // otherwise produce an empty Agent() label / empty agentContext anchor
-    // rather than the intended fallback (idPrefix / parent.sessionId).
-    const effectiveAgentType = options.agentType?.trim() || undefined;
-    const effectiveResolvedAgentType = options.resolvedAgentType?.trim() || undefined;
-    const effectiveParentId = options.parentId?.trim() || undefined;
-    const handle = new SubagentHandleImpl<T>(
-      id,
-      session,
-      childController,
-      this.abortGraph,
-      options.outputSchema,
-      // Wall-clock budget for the child's turn (see SUBAGENT_DEFAULT_TIMEOUT_MS
-      // above). Explicit caller values win, including `0` for unbounded and the
-      // background SUBAGENT_BACKGROUND_TIMEOUT_MS the SubagentExecutor stamps —
-      // `??` preserves that precedence. The default is env-tunable via
-      // AFK_SUBAGENT_TIMEOUT_MS (resolveSubagentTimeoutMs); an unset/invalid
-      // env value returns SUBAGENT_DEFAULT_TIMEOUT_MS, so behaviour is unchanged
-      // when the var is not set.
-      options.config.timeoutMs ?? resolveSubagentTimeoutMs(),
-      registry,
-      () => {
-        // Runs on every terminal outcome of the child — success, failure,
-        // timeout, and abort — so it is the settle hook the heartbeat's
-        // teardown belongs on.
-        stopOccupancyHeartbeat();
-        this.active.delete(id);
-        this.abortGraph.dispose(id);
-      },
-      parentInputStreamRef,
-      parentAbortSignal,
-      // agentType: explicit override → idPrefix fallback. Lets callers
-      // (e.g. compose) supply a render-only label without altering idPrefix
-      // which is also used for routing telemetry.
-      effectiveAgentType ?? options.idPrefix,
-      sink,
-      // parentId: explicit override → parent session id fallback. Lets
-      // callers (e.g. compose) anchor the renderer's nesting at the compose
-      // tool_use_id rather than at the orchestrator session id.
-      effectiveParentId ?? options.parent.sessionId,
-      // traceWriter: child shares the parent's writer so its SubagentStop
-      // hook decision lands in the same trace file. Contract:
-      // docs/philosophy/afk-contract.md — "a child sub-agent inherits its
-      // parent's witness." Resolved once above (per-fork config →
-      // manager-level writer) so the handle's terminal lifecycle emits
-      // pair with the 'started' emit below even when inheritance came
-      // from the manager rather than the per-fork config.
-      effectiveTraceWriter,
-      // onSubagentSucceeded: propagate completion data to the parent
-      // session's session_sealed rollup accumulators.
-      this.onSubagentSucceededCb,
-      // Progress-aware idle-watchdog window for the child's turn (see
-      // SUBAGENT_DEFAULT_IDLE_TIMEOUT_MS above). Explicit caller values win,
-      // including `0` to disable the watchdog for this fork — `??` preserves that
-      // precedence. The default is env-tunable via AFK_SUBAGENT_IDLE_TIMEOUT_MS
-      // (resolveSubagentIdleTimeoutMs); an unset/invalid env value returns
-      // SUBAGENT_DEFAULT_IDLE_TIMEOUT_MS, so behaviour is unchanged when the var
-      // is not set. Runs concurrently with the wall-clock budget above.
-      options.config.idleTimeoutMs ?? resolveSubagentIdleTimeoutMs(),
-    );
-    this.active.set(id, handle as SubagentHandleImpl<unknown>);
 
     // Witness layer: subagent_lifecycle.started fires AFTER the handle is
     // wired into the manager's active map and the abort-graph. Emitting
--- a/src/agent/subagent.ts
+++ b/src/agent/subagent.ts
@@ -34,7 +34,7 @@ import type { AbortOrigin, TraceWriter } from './trace/index.js';
 import type { Surface } from './awareness/types.js';
 import { appendRoutingDecision } from './routing-telemetry.js';
 import { getCurrentSink } from './_lib/skill-sink-channel.js';
-import { touchWorktreeOccupancy } from './worktree-occupancy.js';
+import { touchWorktreeOccupancy, startWorktreeOccupancyHeartbeat } from './worktree-occupancy.js';
 import { resolveWorktreeMainRoot } from './worktree-read-root.js';
 import { computeInheritedReadRoots, type ReadScopeInputs } from './subagent-read-scope.js';
 import { getAfkStateDir } from '../paths.js';
@@ -982,8 +982,19 @@ export class SubagentManager {
     // for cwds outside `.afk-worktrees/`, so it can never delay or fail the
     // fork. Single wiring point — agent/skill/compose/farm dispatches all
     // converge here, whether cwd came per-call or via manager inheritance.
+    //
+    // Ordering constraint: one touch only protects the tree for the sweep's
+    // MIN_EMPTY_AGE_MS (1h), after which a still-running child ages back into
+    // the `empty` verdict and is force-removed mid-flight (#759). So arm a
+    // heartbeat alongside the initial touch, and capture its stop handle HERE —
+    // before the session is constructed — so every exit path below (settle,
+    // construction throw) already has the inverse in hand and cannot orphan the
+    // timer. The timer is unref()'d, so even a leaked one cannot hold the
+    // process open.
+    let stopOccupancyHeartbeat: () => void = () => {};
     if (childConfig.cwd !== undefined) {
       void touchWorktreeOccupancy(childConfig.cwd);
+      stopOccupancyHeartbeat = startWorktreeOccupancyHeartbeat(childConfig.cwd);
     }
 
     let session: AgentSession;
@@ -992,7 +1003,9 @@ export class SubagentManager {
     } catch (err) {
       // Construction failed (e.g. invalid model, sync init failure).
       // Release the graph node that was registered above to prevent an orphan
-      // accumulating on repeated retries (e.g. forge/farm retry loops).
+      // accumulating on repeated retries (e.g. forge/farm retry loops), and
+      // disarm the occupancy heartbeat — there is no child left to protect.
+      stopOccupancyHeartbeat();
       this.abortGraph.dispose(id);
       throw err;
     }
@@ -1023,6 +1036,10 @@ export class SubagentManager {
       options.config.timeoutMs ?? resolveSubagentTimeoutMs(),
       registry,
       () => {
+        // Runs on every terminal outcome of the child — success, failure,
+        // timeout, and abort — so it is the settle hook the heartbeat's
+        // teardown belongs on.
+        stopOccupancyHeartbeat();
         this.active.delete(id);
         this.abortGraph.dispose(id);
       },
--- a/src/agent/tools/handlers/worktree-managed.test.ts
+++ b/src/agent/tools/handlers/worktree-managed.test.ts
@@ -121,6 +121,52 @@ describe('removeManagedWorktreeGuarded — guards + argv', () => {
     // No status check when forced.
     expect(mock.calls.some((c) => c.args.includes('status'))).toBe(false);
   });
+
+  // #759 (second removal path): bare `status --porcelain` never reports
+  // ignored files, so a tree whose only content is a non-rebuildable ignored
+  // file (`.env`) reads clean and reached `git worktree remove` undefended.
+  it('refuses a tree holding a non-rebuildable ignored file without force (reason: ignored-local-state)', async () => {
+    const wtPath = join(repoRoot, '.afk-worktrees', 'has-dotenv');
+    const mock = makeMock((call) => {
+      if (call.args.includes('--ignored')) return { stdout: '!! .env\n!! node_modules/\n', stderr: '' };
+      if (call.args.includes('status')) return { stdout: '', stderr: '' }; // tracked tree is clean
+      return { stdout: '', stderr: '' };
+    });
+    const outcome = await removeManagedWorktreeGuarded({ execFile: mock, repoRoot, worktreePath: wtPath });
+    expect(outcome).toEqual({ removed: false, reason: 'ignored-local-state' });
+    expect(mock.calls.some((c) => c.args.includes('remove'))).toBe(false);
+  });
+
+  // Counterweight: ignored build output alone must NOT block removal, or
+  // every worktree with node_modules/ would become unreapable.
+  it('still succeeds when the only ignored content is rebuildable output', async () => {
+    const wtPath = join(repoRoot, '.afk-worktrees', 'only-build-output');
+    const mock = makeMock((call) => {
+      if (call.args.includes('--ignored')) return { stdout: '!! node_modules/\n!! dist/\n', stderr: '' };
+      if (call.args.includes('status')) return { stdout: '', stderr: '' };
+      return { stdout: '', stderr: '' };
+    });
+    const outcome = await removeManagedWorktreeGuarded({ execFile: mock, repoRoot, worktreePath: wtPath });
+    expect(outcome.removed).toBe(true);
+    const rm = mock.calls.find((c) => c.args.includes('remove'));
+    expect(rm?.args).toEqual(['-C', repoRoot, 'worktree', 'remove', wtPath]);
+  });
+
+  it('force:true still removes even when non-rebuildable ignored content is present', async () => {
+    const wtPath = join(repoRoot, '.afk-worktrees', 'force-dotenv');
+    const mock = makeMock((call) => {
+      if (call.args.includes('--ignored')) return { stdout: '!! .env\n', stderr: '' };
+      return { stdout: '', stderr: '' };
+    });
+    const outcome = await removeManagedWorktreeGuarded({
+      execFile: mock, repoRoot, worktreePath: wtPath, force: true,
+    });
+    expect(outcome.removed).toBe(true);
+    const rm = mock.calls.find((c) => c.args.includes('remove'));
+    expect(rm?.args).toContain('--force');
+    // Explicit override bypasses the ignored-state probe entirely.
+    expect(mock.calls.some((c) => c.args.includes('--ignored'))).toBe(false);
+  });
 });
 
 describe('createIsolatedWorktree', () => {
@@ -228,5 +274,23 @@ describe('teardownIsolatedWorktree', () => {
     });
     const result = await teardownIsolatedWorktree({ execFile: mock, repoRoot, worktreePath: wtPath });
     expect(result).toEqual({ removed: false, preserved: false });
+  });
+
+  // The tree LOOKS clean to `git status`, so the lock reason is the only
+  // place the operator can learn WHY it was preserved instead of removed.
+  it('preserves + locks a tree holding non-rebuildable ignored files, naming the reason', async () => {
+    const wtPath = join(repoRoot, '.afk-worktrees', 'iso-has-dotenv');
+    const mock = makeMock((call) => {
+      if (call.args.includes('--ignored')) return { stdout: '!! .env\n', stderr: '' };
+      if (call.args.includes('status')) return { stdout: '', stderr: '' };
+      return { stdout: '', stderr: '' };
+    });
+    const result = await teardownIsolatedWorktree({ execFile: mock, repoRoot, worktreePath: wtPath });
+    expect(result).toEqual({ removed: false, preserved: true, reason: 'ignored-local-state' });
+    expect(mock.calls.some((c) => c.args.includes('remove'))).toBe(false);
+    const lock = mock.calls.find((c) => c.args.includes('lock'));
+    expect(lock?.args.join(' ')).toContain('afk: isolated-worktree preserved (ignored-local-state');
+    // Legible without needing to already know the reason code.
+    expect(lock?.args.join(' ')).toMatch(/non-rebuildable ignored files/);
   });
 });
--- a/src/agent/tools/handlers/worktree-managed.ts
+++ b/src/agent/tools/handlers/worktree-managed.ts
@@ -21,6 +21,7 @@ import { promisify } from 'node:util';
 import { promises as fs } from 'node:fs';
 import { join, resolve, isAbsolute, dirname } from 'node:path';
 import type { ExecFileFn } from '../../worktree-sweep.js';
+import { hasNonRebuildableIgnoredFiles } from '../../worktree-ignored-probe.js';
 import { env } from '../../../config/env.js';
 
 /** Default git runner. Exported so callers without their own can reuse it. */
@@ -162,7 +163,8 @@ export async function managedWorktreeCommitsAhead(
 export type GuardedRemoveOutcome =
   | { removed: true; branchPreserved: string | null }
   | { removed: false; reason: 'dirty' }
-  | { removed: false; reason: 'commits-ahead'; commitsAhead: number };
+  | { removed: false; reason: 'commits-ahead'; commitsAhead: number }
+  | { removed: false; reason: 'ignored-local-state' };
 
 export interface RemoveManagedWorktreeArgs {
   execFile: ExecFileFn;
@@ -176,10 +178,21 @@ export interface RemoveManagedWorktreeArgs {
 }
 
 /**
- * Guarded worktree removal. Refuses (returns `removed:false` + reason) a dirty
- * or commits-ahead tree unless `force`. Emits `git worktree remove [--force]
- * <path>` on success. Never removes the branch ref. Caller maps the reason to
- * user-facing text (handler) or preserve-and-lock (executor teardown).
+ * Guarded worktree removal. Refuses (returns `removed:false` + reason) a dirty,
+ * commits-ahead, or ignored-local-state tree unless `force`. Emits `git
+ * worktree remove [--force] <path>` on success. Never removes the branch ref.
+ * Caller maps the reason to user-facing text (handler) or preserve-and-lock
+ * (executor teardown).
+ *
+ * Invariant: `isManagedWorktreeDirty` uses bare `git status --porcelain`,
+ * which never reports IGNORED entries — a worktree whose only content is a
+ * non-rebuildable ignored file (`.env`, a gitignored plan) reads clean here
+ * and would otherwise reach `git worktree remove` and be destroyed (#759,
+ * same defect the sweep engine hardened against in `worktree-sweep.ts`; this
+ * is the second, more frequently executed removal path). The
+ * `hasNonRebuildableIgnoredFiles` probe closes it without changing what
+ * "dirty" means for other callers of `isManagedWorktreeDirty` — it is a
+ * separate guard, not a redefinition.
  */
 export async function removeManagedWorktreeGuarded(
   args: RemoveManagedWorktreeArgs,
@@ -188,6 +201,9 @@ export async function removeManagedWorktreeGuarded(
   if (!force) {
     if (await isManagedWorktreeDirty(execFile, worktreePath)) {
       return { removed: false, reason: 'dirty' };
+    }
+    if (await hasNonRebuildableIgnoredFiles(execFile, worktreePath)) {
+      return { removed: false, reason: 'ignored-local-state' };
     }
     const ahead = await managedWorktreeCommitsAhead(execFile, repoRoot, worktreePath);
     if (ahead > 0) {
@@ -258,17 +274,30 @@ export async function createIsolatedWorktree(args: {
 /** Result of {@link teardownIsolatedWorktree}. */
 export interface IsolatedTeardownResult {
   removed: boolean;
-  /** True when the tree was kept (dirty/ahead) and locked against the sweep. */
+  /** True when the tree was kept (dirty/ahead/ignored) and locked against the sweep. */
   preserved: boolean;
-  reason?: 'dirty' | 'commits-ahead';
+  reason?: 'dirty' | 'commits-ahead' | 'ignored-local-state';
+}
+
+/**
+ * Human-legible label for a preserved tree's lock reason. `dirty` and
+ * `commits-ahead` are self-explanatory; `ignored-local-state` is not — the
+ * tree LOOKS clean to `git status`, so the lock reason is the only place the
+ * operator can learn why it was kept anyway.
+ */
+function describePreserveReason(reason: NonNullable<IsolatedTeardownResult['reason']>): string {
+  if (reason === 'ignored-local-state') {
+    return 'ignored-local-state: non-rebuildable ignored files present (e.g. .env) — git status looked clean';
+  }
+  return reason;
 }
 
 /**
  * Tear down an isolated worktree after its subagent finishes. Removes a
- * clean tree; PRESERVES a dirty / commits-ahead tree (WIP is never destroyed)
- * and `git worktree lock`s it so the sweep engine never reaps it out from
- * under work in progress. Best-effort — never throws (teardown runs in a
- * `finally`).
+ * clean tree; PRESERVES a dirty / commits-ahead / ignored-local-state tree
+ * (WIP and untracked local state are never destroyed) and `git worktree
+ * lock`s it so the sweep engine never reaps it out from under work in
+ * progress. Best-effort — never throws (teardown runs in a `finally`).
  */
 export async function teardownIsolatedWorktree(args: {
   execFile?: ExecFileFn;
@@ -284,11 +313,12 @@ export async function teardownIsolatedWorktree(args: {
       force: false,
     });
     if (outcome.removed) return { removed: true, preserved: false };
-    // Dirty or commits-ahead → preserve WIP; lock so the sweep never reaps it.
+    // Dirty, commits-ahead, or ignored-local-state → preserve WIP/local state;
+    // lock so the sweep never reaps it.
     try {
       await execFile('git', [
         '-C', args.repoRoot, 'worktree', 'lock',
-        '--reason', `afk: isolated-worktree preserved (${outcome.reason})`,
+        '--reason', `afk: isolated-worktree preserved (${describePreserveReason(outcome.reason)})`,
         args.worktreePath,
       ]);
     } catch { /* best-effort */ }
--- a/src/agent/tools/handlers/worktree-managed.ts
+++ b/src/agent/tools/handlers/worktree-managed.ts
@@ -285,7 +285,7 @@ export interface IsolatedTeardownResult {
  * tree LOOKS clean to `git status`, so the lock reason is the only place the
  * operator can learn why it was kept anyway.
  */
-function describePreserveReason(reason: NonNullable<IsolatedTeardownResult['reason']>): string {
+export function describePreserveReason(reason: NonNullable<IsolatedTeardownResult['reason']>): string {
   if (reason === 'ignored-local-state') {
     return 'ignored-local-state: non-rebuildable ignored files present (e.g. .env) — git status looked clean';
   }
--- a/src/agent/tools/handlers/worktree.test.ts
+++ b/src/agent/tools/handlers/worktree.test.ts
@@ -379,6 +379,37 @@ describe('worktree handler — remove guards', () => {
     expect(result.content).toContain('commit(s) ahead');
   });
 
+  // #759 (second removal path): a tree whose only content is a non-rebuildable
+  // ignored file (`.env`) reads clean to bare `status --porcelain` and used to
+  // reach `git worktree remove` undefended.
+  it('refuses a worktree holding a non-rebuildable ignored file without force', async () => {
+    const wtPath = join(afkRoot, 'dotenv-wt');
+    const mock = makeMock(standardResponder(`${block(repoRoot)}\n\n${block(wtPath)}\n`, (call) => {
+      if (call.args.includes('--ignored')) return { stdout: '!! .env\n', stderr: '' };
+      if (call.args.includes('status')) return { stdout: '', stderr: '' };
+      return undefined;
+    }));
+    const handler = createWorktreeHandler(repoRoot, { execFile: mock });
+    const result = await handler({ action: 'remove', path: 'dotenv-wt' }, SIGNAL);
+    expect(result.isError).toBe(true);
+    expect(result.content).toContain('non-rebuildable ignored files');
+    expect(mock.calls.some((c) => c.args.includes('remove'))).toBe(false);
+  });
+
+  it('force removes a worktree holding a non-rebuildable ignored file with --force', async () => {
+    const wtPath = join(afkRoot, 'dotenv-force-wt');
+    const mock = makeMock(standardResponder(`${block(repoRoot)}\n\n${block(wtPath)}\n`, (call) => {
+      if (call.args.includes('--ignored')) return { stdout: '!! .env\n', stderr: '' };
+      if (call.args.includes('status')) return { stdout: '', stderr: '' };
+      return undefined;
+    }));
+    const handler = createWorktreeHandler(repoRoot, { execFile: mock });
+    const result = await handler({ action: 'remove', path: 'dotenv-force-wt', force: true }, SIGNAL);
+    expect(result.isError).toBeUndefined();
+    const removeCall = mock.calls.find((c) => c.args.includes('remove'));
+    expect(removeCall?.args).toContain('--force');
+  });
+
   it('removes a clean worktree and preserves the branch', async () => {
     const wtPath = join(afkRoot, 'clean-wt');
     const mock = makeMock(
--- a/src/agent/tools/handlers/worktree.ts
+++ b/src/agent/tools/handlers/worktree.ts
@@ -316,6 +316,14 @@ export function createWorktreeHandler(
                 isError: true,
               };
             }
+            if (outcome.reason === 'ignored-local-state') {
+              return {
+                content: `Refused: ${entry.path} holds non-rebuildable ignored files (e.g. .env, a gitignored plan) that ` +
+                  `\`git status\` cannot see, so removal would silently delete them. Move/back up what you need, or ` +
+                  'pass force: true to discard them along with the checkout.',
+                isError: true,
+              };
+            }
             return {
               content: `Refused: ${entry.path} has ${outcome.commitsAhead} commit(s) ahead of its base. The branch ref would survive, but pass force: true to confirm removing the checkout.`,
               isError: true,
--- a/src/agent/tools/schemas.ts
+++ b/src/agent/tools/schemas.ts
@@ -694,7 +694,9 @@ export const worktreeTool: AnthropicToolDef = {
     '- `release` — unlock a previously kept worktree, returning it to normal sweep lifecycle.\n' +
     '- `list` — dry-run sweep report: every afk-managed worktree with its verdict ' +
     '(active | empty | stale-clean | stale-dirty | locked | dead-owner | orphaned-*), owner, and age ' +
-    'in days. Verdicts empty/dead-owner/orphaned-* are removal candidates on the next sweep.\n' +
+    'in days. `stale-dirty` also covers a tree that `git status` calls clean but which holds ' +
+    'non-rebuildable ignored files (e.g. `.env`). Verdicts empty/dead-owner/orphaned-* are removal ' +
+    'candidates on the next sweep.\n' +
     '- `remove` — remove a worktree checkout you no longer need (branch ref is always preserved). ' +
     'Refuses dirty trees, locked trees, trees with commits ahead of base, and trees holding ' +
     'non-rebuildable ignored files (e.g. `.env`) unless `force: true`. Never removes the main ' +
--- a/src/agent/tools/schemas.ts
+++ b/src/agent/tools/schemas.ts
@@ -696,8 +696,9 @@ export const worktreeTool: AnthropicToolDef = {
     '(active | empty | stale-clean | stale-dirty | locked | dead-owner | orphaned-*), owner, and age ' +
     'in days. Verdicts empty/dead-owner/orphaned-* are removal candidates on the next sweep.\n' +
     '- `remove` — remove a worktree checkout you no longer need (branch ref is always preserved). ' +
-    'Refuses dirty trees, locked trees, and trees with commits ahead of base unless `force: true`. ' +
-    'Never removes the main worktree or paths outside `.afk-worktrees/`.\n\n' +
+    'Refuses dirty trees, locked trees, trees with commits ahead of base, and trees holding ' +
+    'non-rebuildable ignored files (e.g. `.env`) unless `force: true`. Never removes the main ' +
+    'worktree or paths outside `.afk-worktrees/`.\n\n' +
     'Finishing with a worktree: a worktree is scaffolding, not an artifact — once its work has landed ' +
     'somewhere durable (pushed branch, open PR, merged commit), remove it in the same turn instead of ' +
     'leaving it for the sweep. Two cases differ:\n' +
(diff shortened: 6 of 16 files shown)


griffinwork40:
@codex fix comments

griffinwork40:
Review complete. Delegation note first: 3 of 4 dimension agents died on the first wave (one 45-min timeout, two mid-stream cuts) and 2 of 3 recovery nodes hit their tool-call cap. I fell back to inline verification for the correctness dimensions and ran the two decisive experiments myself, so coverage is intact — but the doc-drift sweep is thinner than I'd like (noted at the end).

---

# Review: PR #770 — worktree sweep reaper safety

**+430 / −7 across 7 files.** Two independent fixes: a heartbeat keeping a live subagent's worktree marked occupied, and an ignored-file probe so the sweep stops force-deleting trees whose only content is gitignored.

The core judgment in this PR is correct and unusually well-reasoned — the fail-safe direction is right everywhere, and the "counterweight" test proving `node_modules/` must *not* protect is exactly the test most authors would omit. The problems are all at the boundaries.

## Blocking

### B1 · The fix leaves a second, hotter removal path with the identical bug (critical)

`isManagedWorktreeDirty` — `src/agent/tools/handlers/worktree-managed.ts:127-137` — uses bare `git status --porcelain`, no `--ignored`. It is the sole non-force guard in `removeManagedWorktreeGuarded` (:189), which serves **two** consumers the PR never touches:

| Path | Trigger | Protected by #770? |
|---|---|---|
| `worktree-sweep.ts:718,735` | cron/daemon, tree ≥1h old | ✅ yes |
| `teardownIsolatedWorktree` (:273, `force:false`) → `foreground-promotion.ts:405` | **every** `isolation:'worktree'` child completion, immediately | ❌ no |
| `worktree` tool `remove` action → `handlers/worktree.ts:289` | agent/operator cleanup | ❌ no |

I verified the mechanism actually destroys data, in a throwaway repo:

```
--- status in worktree (bare porcelain) ---
[bare porcelain empty => code judges CLEAN]
--- attempt: git worktree remove (NO --force) ---
RESULT: removed WITHOUT --force
--- does the .env survive? ---
.env DESTROYED
```

Plain `git worktree remove` — no `--force` — deletes a tree whose only content is ignored, `.env` included. So the PR hardens the path that fires after an hour and leaves open the one that fires on every isolated dispatch.

This is worse than a symmetry gap, because the `worktree` tool's own description actively steers agents into it: *"once the branch is pushed that force is safe, because remove never deletes the branch ref — the commits stay on the branch and on the remote, and only the directory goes away."* That reasons purely about commits and is silent on ignored files. An agent following the documented happy path force-removes believing nothing can be lost. And AFK.md instructs agents to remove worktrees once work has landed — so this is the *documented* flow, not an edge case.

**Fix is ~2 lines and reuses the module this PR already created:** call `hasNonRebuildableIgnoredFiles` inside `isManagedWorktreeDirty`. Both consumers inherit it, and `teardownIsolatedWorktree` already has the correct landing branch — `reason: 'dirty'` → preserve + `git worktree lock`. That's precisely the desired outcome. Cheap enough that filing it as a follow-up rather than doing it here is the wrong trade, since #759's actual harm stays reachable while the PR claims to close it.

### B2 · Nothing enforces `DEFAULT_HEARTBEAT_INTERVAL_MS < MIN_EMPTY_AGE_MS` (high)

`DEFAULT_HEARTBEAT_INTERVAL_MS = 600_000` (`worktree-occupancy.ts:45`) and `MIN_EMPTY_AGE_MS = 3_600_000` (`worktree-sweep.ts:165`) are both private, in separate files, related only by a comment saying one "must stay comfortably below" the other. Every heartbeat test passes an explicit `10`ms override, so the production value is never exercised. Raise the interval to 4h and the fix silently dies with zero test failures. Export both and add one arithmetic assertion — no timers needed.

## Non-blocking, ranked

**H1 · Directory collapse defeats the probe by design.** Empirically confirmed: `git status --porcelain --ignored` collapses `!! dist/` into one line and `dist/NOTES.md` is *invisible* (grep count 0). `--ignored=matching` did not change this when the directory itself matches an ignore pattern. So `isRebuildableIgnoredEntry` classifies by **directory name, not contents** — any non-rebuildable file inside a rebuildable-named ignored dir is destroyed. Realistic hits: `out/` used for generated deliverables, `.cache/` holding expensive recorded fixtures, `target/`. The docblock (`worktree-ignored-probe.ts:243-246`) presents collapse as a pure win — *"keeps this cheap on a populated worktree"* — without noting it is simultaneously a correctness hole. At minimum document it as accepted; better, descend one level for the ~15 rebuildable *directory* patterns before classing them reapable.

**H2 · `/\.log$/` is the most over-broad pattern** (`worktree-ignored-probe.ts:229`). Any `*.log` anywhere is reapable, and logs are the canonical gitignored-but-valuable artifact — a captured long-run log is exactly the thing an operator wants to keep. Consider restricting to known-noise names or dropping it entirely.

**M1 · A protected tree is illegible to the operator.** Verified empirically by the test agent: the tree surfaces as `verdict: "stale-dirty"` with `"[WARN] stale-dirty worktree preserved (uncommitted changes)"`. But `git status` shows clean — so the operator is told they have uncommitted changes that do not exist, with no pointer to the real cause. There is no distinct reason field. Add one to the candidate/warning.

**M2 · The heartbeat rewrites `createdAt`, corrupting age reporting.** `touchWorktreeOccupancy` sets `meta['createdAt'] = new Date().toISOString()` (`worktree-occupancy.ts:76`). The heartbeat now does that every 10 minutes for the child's whole life, so after a 6h job the recorded `createdAt` is 6h later than reality — permanently, after the child exits. That value is user-visible as `ageDays` at `handlers/worktree.ts:283`, `cli/slash/commands/worktree.ts:61`, `cli/commands/worktree.ts:55`. The field is being used as `lastActiveAt` while named `createdAt`. It also suppresses `stale-clean`/`stale-dirty`, not just `empty` (`worktree-sweep.ts:336,343`) — collateral beyond the PR's stated scope. Cleaner contract: add `lastTouchedAt`, have the sweep use `max(createdAt, lastTouchedAt)` for liveness, and leave `createdAt` immutable for display. Pre-existing at dispatch-time, but the heartbeat makes it systematic.

**M3 · `stop()` cannot cancel an in-flight touch.** `stop()` only does `clearInterval`; a fired `void touchWorktreeOccupancy(cwd)` has two unsynchronized awaits (read then write, `:71,77`). Reproduced by the test agent with a 25ms injected write delay — the write lands after `stop()`. Benign in production (a late touch just extends protection) but it makes the new test's assertion depend on sub-millisecond local I/O.

**M4 · One new sweep test does not discriminate.** Mutation-tested: hardcoding `hasNonRebuildableIgnoredFiles → return false` still passes *"still reaps a clean tree whose only ignored content is rebuildable output."* It re-proves pre-existing clean+0-ahead+old ⇒ `empty` behavior and would pass with the probe deleted. Assert a `--ignored` call landed.

**M5 · Heartbeat arm/disarm in `subagent.ts` is untested.** Zero `heartbeat` hits across subagent tests; the existing construction-throw test passes no `cwd`, so `childConfig.cwd !== undefined` is false and the entire new branch never executes. Both the throw path and the settle path are uncovered.

**L1 · Quoted paths are over-protective — correct direction.** Confirmed: git emits `"my file.env"` and `"\303\251.env"` under default `core.quotePath`. The leading `"` defeats the `^`/`/`-anchored patterns, so odd paths always protect. Fail-safe; worth one comment line so a future author doesn't "fix" it.

**L2 · Misplaced comment.** `worktree-sweep.ts:526` — *"Type-only back-import there, so this pairing carries no runtime cycle"* — sits directly above a **runtime value import**. The type-only import is the other direction (probe → sweep). Accurate about the pairing, misleading about the line beneath it.

**L3 · 350-LOC ceiling.** The new file is 117 (compliant). But `subagent.ts` 1138 → ~1157 and `worktree-sweep.ts` 783 → ~797 both grow while sitting 2–3× over the stated hard ceiling. Pre-existing, not introduced here — flagged only because the rule is stated as absolute and this PR extends both files rather than pulling the concern out.

## Credit where due

Fail-safe defaults are consistent throughout (unreadable tree → protect, mirroring the sweep's existing `catch { isDirty = true }`). The teardown-before-setup ordering in `subagent.ts:27` with the constraint named in-comment matches the repo's ordered-sequence convention exactly. `timer.unref()` is the right call. And the rewritten "residual gap" paragraph honestly retires a claim the PR invalidates instead of leaving it to rot.

## Recommendation

**Request changes** — B1 and B2 only, both cheap. B1 is the one that matters: it is ~2 lines, reuses this PR's own module, lands in an already-correct fallback branch, and without it the PR closes #759 while leaving #759's actual data loss reachable through the more frequently-executed path. Everything under "Non-blocking" is legitimate follow-up; H1 and H2 are the two I'd file immediately.

**Verification status:** B1's mechanism and H1/L1's git semantics are empirically reproduced, not inferred. M1, M3, M4, M5 come from a subagent that ran the tests in an isolated `/tmp` copy with mutation testing — I did not independently re-derive those four. Baseline on `main`: `worktree-occupancy.test.ts` 9 passed, `worktree-sweep.test.ts` 41 passed, `pnpm lint` clean.

**Not checked:** doc drift (whether AFK.md / website / tool descriptions repeat the now-retired "`keep` is the escape hatch for >1h children" claim) — the agent assigned to it died twice and I did not re-dispatch. Also unchecked: submodule/nested-repo behavior of `--ignored`, and the real cost delta of the extra `git status --ignored` per candidate.

---

**Done**
- Reviewed PR #770 across correctness, data-safety, test quality, and convention dimensions; produced a merge recommendation of *request changes* on two blocking items with 11 further ranked findings.
- Evidence: diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/c56a4f24-87a9-4ff6-ba3e-ad00f5825458/pr-770.diff`; the B1 data-loss reproduction and H1/L1 git-semantics results are in this transcript's `bash`/`compose` tool output; baseline test run `worktree-occupancy.test.ts` 9 passed / `worktree-sweep.test.ts` 41 passed / `pnpm lint` clean.
- World unchanged: review is read-only — no files edited, no commits, no PR comment posted, working tree still clean on `main` @ `03af774`. All scratch repos under `/tmp` were removed.
- Pending: findings are transcript-only and not yet on the PR — say the word and I'll post them as a review comment. Doc-drift dimension went uncovered after two agent failures; B1's fix is not implemented, only specified.

---

griffinwork40:
Follow-up to the review above — the two blocking items are addressed in `4fd3723`, CI green.

**Fixed**

1. **The second removal path (was blocking).** The PR hardened the sweep, but `isManagedWorktreeDirty` (`worktree-managed.ts`) used bare `git status --porcelain`, and it is the only non-force guard in `removeManagedWorktreeGuarded` — which serves both `teardownIsolatedWorktree` (every `isolation:'worktree'` child completion) and the `worktree` tool's `remove` action. Verified empirically that plain `git worktree remove` *without* `--force` deletes a worktree whose only content is ignored, destroying a `.env`; in this repo `.afk/plans/` is gitignored and holds plan-of-record files. Fixed additively: new `GuardedRemoveOutcome` variant `ignored-local-state` plus a sibling probe guard, so the meaning of `isManagedWorktreeDirty` is unchanged for other callers. `force: true` still bypasses, and a preserved tree's `worktree lock --reason` now names the cause (otherwise `git status` looks clean and the operator has no way to learn why).
2. **Constant-relationship guard (was blocking).** `DEFAULT_HEARTBEAT_INTERVAL_MS` and `MIN_EMPTY_AGE_MS` were private, in separate files, related only by a prose comment — raising the interval above the age gate would silently re-break #759 with no test failure. Both are now exported with a timer-free arithmetic assertion (interval ≤ ¼ of the gate).
3. **Non-discriminating test.** `"still reaps a clean tree whose only ignored content is rebuildable output"` passed even with the probe hardcoded to `return false`. Now asserts the probe actually ran; re-mutation-tested to confirm it fails without the probe.
4. **Misleading comment.** The "type-only back-import" note sat above a runtime value import and described the opposite direction.
5. **Tool description.** The `worktree` `remove` guidance argued force is safe because commits survive, and was silent on ignored files — actively steering agents into the data loss. Now scopes that claim to commits and cautions that force deletes untracked/ignored local state.

**Deliberately deferred** (each worth its own issue, none blocking)

- **Directory collapse limits the probe.** `git status --porcelain --ignored` collapses `!! dist/` into one line; a hand-written `dist/NOTES.md` is invisible (`--ignored=matching` did not change this when the directory itself matches an ignore pattern). So classification is by directory *name*, not contents. Realistic exposure: `out/` holding generated deliverables, `.cache/` holding expensive recorded fixtures.
- **`/\.log$/` is the most over-broad pattern** — any `*.log` anywhere is reapable, and logs are the canonical gitignored-but-valuable artifact.
- **The heartbeat rewrites `createdAt`.** `touchWorktreeOccupancy` stamps `createdAt = now` every 10 min for the child's life, so a long job permanently resets the tree's recorded age — user-visible as `ageDays` in the tool and both CLI surfaces. The field is functioning as `lastActiveAt` under a `createdAt` name; a separate field would keep display honest. Also suppresses `stale-clean`/`stale-dirty`, not just `empty`.
- **A sweep-protected tree still reports generic `stale-dirty` / "uncommitted changes"**, which contradicts a clean `git status`. Fixed for the tool path above; not for the sweep's own warning.
- **`stop()` cannot cancel an in-flight touch** (benign in production — a late touch only extends protection — but it makes the new test's assertion latency-dependent).
- **Heartbeat arm/disarm in `subagent.ts` is untested**; the existing construction-throw test passes no `cwd`, so that branch never executes.
- The `classifyCandidate` / file-size extraction stays out, per the maintainer's decision to keep a pure code move off this behavioural fix.

griffinwork40:

---

# Review — PR #770 `fix(worktree): stop the sweep reaping live trees and discarding ignored files (#759)`

**Target** `afk/lane4-reaper-safety` → `main` · 13 files · +620/−28 · regime **full**
**Reviewed ref** `4fd3723768e8ea487207e3adc998c8f56d3ff13e` (present locally; all citations below re-read at this ref)
**Stated intent** PR title + body — spec-compliance **assessed**.

## Merge decision: **DO NOT MERGE** (as-is)

Net direction is right and the code is careful. But two of the four worktree-removal paths in this repo are still unguarded, the probe cannot see nested content *by construction*, and the diff ships model-facing documentation asserting a safety guarantee that the most-executed removal path does not honour. The required changes are small and additive — this is a "finish the fix," not a rewrite.

---

## Blocking

### H1 · `high` · high confidence · security · `src/cli/commands/interactive/worktree.ts:665-704` (and `:639-646`) · ref:4fd3723 · citation-type:file-state → **verified at ref**

**Finding** — The `afk -w` session-end cleanup path force-removes the worktree after a bare `git status --porcelain` check, so a hand-authored gitignored `.env` in that worktree is destroyed by exactly the defect this PR fixes elsewhere; the probe is never wired into this path.

**Evidence** (`git show 4fd3723:src/cli/commands/interactive/worktree.ts`)
```
667:  status = await execFile('git', ['-C', currentPath, 'status', '--porcelain']);
677:  if (status.stdout.trim().length > 0) {   // untracked → preserve. ignored → invisible.
704:  await execFile('git', ['-C', repoRoot, 'worktree', 'remove', '--force', currentPath]);
```
Production reachability: `src/cli/commands/interactive.ts:530` (`cleanup({ force: ctx.stats.totalTurns === 0 })`) and `src/cli/commands/chat.ts:940`. Absence claim **confirmed-absent**: `git grep hasNonRebuildableIgnoredFiles 4fd3723 -- 'src/**/*.ts'` (prod) returns only `worktree-sweep.ts:625` and `worktree-managed.ts:205`. The `opts.force === true` branch at `:639-646` skips the status check entirely and force-removes unconditionally.

**Aggravating factor** — `src/agent/tools/schemas.ts:707-711` in this same diff now tells every model *"removal now refuses such a tree unless forced"*. That is true for the `worktree` tool and the sweep, and false for session-end cleanup. A false guarantee in the tool description is worse than an absent one.

**Suggestion** — Insert `if (await hasNonRebuildableIgnoredFiles(execFile, currentPath)) { console.log(preserved…); recordCdIntent(currentPath); return; }` between `:675` and `:704`, mirroring `removeManagedWorktreeGuarded`; leave the zero-turn `force` branch as-is but say so in the schema text.

*Independently re-derived by a shadow verifier that enumerated all deletion paths from scratch and reached the same conclusion, then by orchestrator re-read at the ref.*

### H2 · `high` · high confidence · correctness · `src/agent/worktree-ignored-probe.ts:48-56, 82-86, 111-115` · ref:4fd3723 · citation-type:diff-context

**Finding** — A non-rebuildable file nested under a rebuildable ignored directory (`dist/secrets.env`, `out/deliverable.pdf`) is invisible to the probe, so the tree still reads reapable and is force-removed — a residual instance of the exact data-loss class this PR exists to close.

**Evidence**
```
50:  /(?:^|\/)out\//,
114:    if (!isRebuildableIgnoredEntry(entry)) return true;
```
The gap is **doubly latched**. (1) `git status --porcelain --ignored` in traditional mode emits one line per ignored *directory*: *"traditional: Show ignored files and directories, unless `--untracked-files=all` is specified, in which case individual files in ignored directories are displayed"* (git-scm.com/docs/git-status) — so git emits `!! dist/` and `secrets.env` never appears. (2) Even with per-file expansion, `/(?:^|\/)dist\//` matches the full path `dist/secrets.env`, so the leaf is *still* classified rebuildable. `--ignored=matching` does not help either: *"If a directory matches an ignore pattern, then it is shown, but not paths contained in the ignored directory."*

Absence claim **confirmed-absent**: no test fixture of the form `<rebuildable-dir>/<non-rebuildable-file>` exists. The nested fixtures are `node_modules/foo/index.js`, `dist/cli/index.js` (rebuildable leaf) and `scratch/data.csv`, `.vscode/launch.json` (non-rebuildable parent); `isRebuildableIgnoredEntry('my-notes/dist-plan.md') === false` probes the opposite direction.

**Suggestion** — Classify by **leaf**, not by path prefix: match the final path segment against the rebuildable list, and for any collapsed directory entry that matched, expand it once (`-uall` scoped to that path, or a `readdir`) and classify the enumerated leaves. Add the missing `dist/secrets.env` fixture.

### H3 · `high` · high confidence · correctness · `src/agent/worktree-sweep.ts:319-343`, `src/agent/subagent.ts:994-998`, `src/agent/worktree-occupancy.ts:79-95` · ref:4fd3723 · citation-type:file-state → **verified at ref**

**Finding** — Fix #1's stated root cause is not reachable, so the heartbeat does not fix the failure mode the PR body describes: the `empty` verdict requires `ownerLiveness !== 'alive'`, the child runs in the *same OS process* as the parent that stamped the meta, and the dispatch-time touch already writes `pid = process.pid` — so `isProcessAlive(meta.pid)` is necessarily true for the child's whole lifetime and the 1-hour age gate is never reached.

**Evidence**
```
sweep:337-340   candidate.ownerLiveness !== 'alive' && … && candidate.ageMs >= MIN_EMPTY_AGE_MS
occupancy:90-91 meta['pid'] = process.pid;  meta['createdAt'] = new Date().toISOString();
subagent:1002   session = new AgentSession(childConfig);   // 0 matches for child_process|spawn(|fork( in this file
```
The comment at `sweep:331-335` states this gate was added for precisely this scenario (#380). Three consequences: (a) the heartbeat's only marginal value is recovering from a *transient* initial-touch failure, since a persistent one fails identically every 10 minutes; (b) the states that **are** reachable mid-flight — absent/unparseable meta, a `void`-ed touch that never landed, a torn read of the non-atomic write — route to `dead-owner` (`sweep:319-325`), which has **no age gate at all**, so refreshing `createdAt` cannot help there; (c) `touchWorktreeOccupancy` uses plain `fs.writeFile` with no temp+rename (`occupancy:92`), so a 10-minute rewrite cadence *adds* torn-read windows to the file whose torn read produces `ownerLiveness = 'unknown'` → the `empty` path.

**Suggestion** — Either drop fix #1, or keep it and re-target it: make `touchWorktreeOccupancy` atomic (write temp + `rename`), and `await` the initial touch before the fork so the stamp is guaranteed present before the first sweep tick. Correct the PR body's causal claim either way.

*Independently re-derived by a shadow verifier working only from the reachability question, which reached the same conclusion and enumerated the four reachable states.*

---

## Medium

**M1 · security · `src/agent/worktree-sweep.ts:509-531` · file-state → verified** — The orphaned-dir branch runs `fs.rm(orphanPath, { recursive: true, force: true })` (`:523`) with no dirty check, no probe, and an `orphanAgeMs` that is computed at `:512-514` and only ever pushed into `result.candidates` — never gated on. Reached with real deletion from `src/cli/commands/interactive/boot-prune.ts:107` (`dryRun: false`, `bypassSoftLaunch: true`), plus `daemon/scheduler.ts`, `cli/commands/worktree.ts`, `cli/slash/commands/worktree.ts`. Pre-existing and adjacent to the declared "meta-less stays reapable" posture — but that posture argues bounded-loss-vs-unbounded-leak, and it does not cover skipping the new probe *and* having no age gate on an `rm -rf`. **Suggestion** — gate the `fs.rm` on an age threshold at minimum; a probe here needs care, since `git -C <non-worktree-dir> status` reports against the parent repo.

**M2 · correctness · `src/agent/subagent.ts:1012-1023` · diff-context** — The heartbeat's disarm is guarded only around `new AgentSession(childConfig)` (`:1001-1011`); a throw in `options.parent.getInputStreamRef?.()` (`:1012`), `getCurrentSink()` (`:1015`), or `new SubagentHandleImpl(…)` (`:1023`) returns with the interval armed and no handle in existence to ever fire the settle callback at `:1038-1045`, leaving the tree permanently un-reapable. **Suggestion** — extend the `try` to span `startWorktreeOccupancyHeartbeat` → `this.active.set(id, handle)`.

**M3 · perf-observability · `src/agent/worktree-sweep.ts:624-625, 346, 778-781` · diff-context + file-state → verified** — A probe-protected tree gets `isDirty = true`, which yields verdict `active` (no telemetry under the dirty threshold) or `stale-dirty`, whose warning reads `[WARN] stale-dirty worktree preserved (uncommitted changes): <path>` (`:780`) — factually false for a tree whose working copy is clean. The operator gets no signal that ignored state is what saved it. **Suggestion** — carry the probe result as a separate field and emit `[INFO] worktree preserved (non-rebuildable ignored files)`, mirroring `describePreserveReason`, which the interactive path already gets right.

**M4 · test-coverage · `src/agent/worktree-occupancy.test.ts` · diff-context** — No test drives the heartbeat through `runSweep`; the added cases assert only that a meta timestamp moved and that two constants are ordered. Absence **confirmed**: `git grep -l startWorktreeOccupancyHeartbeat 4fd3723 -- 'src/**/*.test.ts'` → `worktree-occupancy.test.ts` alone, which contains **0** `runSweep` references. This is why H3 ships green. **Suggestion** — add a `worktree-sweep.test.ts` case asserting no `empty` verdict for a heartbeat-refreshed tree, and one asserting the *current* pid-alive behaviour, which pins H3 explicitly.

---

## Low / nit

- **L1 · security · `worktree-ignored-probe.ts:78`** — `/\.log$/` marks every ignored `*.log` rebuildable; a captured agent transcript or a hand-kept `decisions.log` is deleted. Narrow to known emitters (`(npm|yarn|pnpm)-debug\.log`, `logs/`).
- **L2 · correctness · `worktree-ignored-probe.ts:111-112`** — With git's default `core.quotePath=true`, a non-ASCII ignored path arrives as `!! "caf\303\251/node_modules/"`; the retained leading `"` defeats every `(?:^|\/)` anchor, so rebuildable content misreads as protective and the tree becomes immortal. Fail-direction is toward protecting, so this is a leak, not a loss. Pass `-c core.quotePath=false`.
- **L3 · security · `src/agent/worktree.ts:242, 334`** — Farm teardown force-removes with no dirty check and no probe. Out of the PR's declared scope: farm dirs come from `getFarmDir()`, not `.afk-worktrees/`. Follow-up, not a blocker. (`removeFarm`/`removeBranch` at `:455-459` have no production caller — dead code.)
- **L4 · api-compat · `foreground-promotion.ts:406-411`** — Logs raw `outcome.reason`, so the operator sees bare `(ignored-local-state)` while the `git worktree lock --reason` on the same event carries the human-legible `describePreserveReason` text. Export and reuse it.
- **L5 · api-compat · `schemas.ts:696`** *(citation corrected by orchestrator — see manifest)* — The `list` verdict sentence documents `stale-dirty` without noting it now also covers trees holding non-rebuildable ignored files.
- **N1 · correctness · `worktree-ignored-probe.ts:84`** — `if (normalized === '') return true;` classifies an unparseable entry as *reapable*, inverting the module's own stated fail-safe direction. `nit` per the reachability rule: `isRebuildableIgnoredEntry` has zero production importers beyond the internal call at `:114`, which already guards `entry === ''`.
- **N2 · test-coverage** — No test covers the promoted/background settle path disarming the heartbeat.

---

## Verified clean — no finding

- **Fix #1's mechanism is correctly wired** (even though H3 shows the gate it targets is already closed): `touchWorktreeOccupancy` rewrites `createdAt` (`occupancy:90-91`) and the sweep derives `ageMs` from it (`sweep:597-600`). Not a no-op.
- **No runtime import cycle** — `worktree-sweep.ts:19` value import vs `worktree-ignored-probe.ts:26` `import type { ExecFileFn }`, erased at compile time. Author's claim confirmed verbatim in both directions.
- **`GuardedRemoveOutcome`'s new union member is handled** — the `ignored-local-state` branch lands at `worktree.ts:319-326` *before* the `commitsAhead` fallthrough, so the final `outcome.commitsAhead` access stays narrowed under strict tsc. No exhaustive `switch` consumer exists over either union; production importers are exactly `worktree.ts`, `foreground-promotion.ts`, `worktree-managed.ts`.
- **argv safety** — `execFile('git', ['-C', worktreePath, 'status', '--porcelain', '--ignored'])`: array argv, no shell, and `worktreePath` occupies the `-C` value slot so a leading-dash path cannot become a flag.
- **`describePreserveReason` is tested** — *unverified — absence claim refuted by* `worktree-managed.test.ts:131-133`.
- **Settle-callback coverage holds** — `subagent.ts:1038-1045` fires on success, failure, timeout and abort; `foreground-promotion.ts`'s `!promoted` guard skips only *worktree teardown*, not the settle callback, so a detached job still disarms.
- **`.afk-worktree-meta.json` as rebuildable is correct** — AFK recreates it, and meta-less trees stay reapable by declared posture.
- **Author's Q1 (350-LOC convention)** — the call is right. New behaviour in a 117-line module is correct, and deferring the `classifyCandidate` extraction is correct: bundling a pure code move with a data-loss fix would make the diff unreviewable against the bug. Not a finding.
- **Spec-compliance: no unmet intent clause, and no true scope creep.** Every numbered claim in the body has an implementing hunk. `worktree-managed.ts`, `worktree.ts`, `schemas.ts`, `foreground-promotion.ts` are all in-service-of the intent — a newly-refusing tool with an undocumented refusal would strand callers. The scope is **under-declared, not crept**. What H1 shows is that it is *inconsistently* applied: 2 of 4 removal paths closed.

---

| id | cited | status | resolution |
|---|---|---|---|
| SEC-1 → H1 | `interactive/worktree.ts:665-704` | **verified** | exact match at ref |
| COR-3 → M2 | `subagent.ts:990-998, 1025-1028` | **verified** | corrected to `:994-998` / `:1012-1023` / `:1038-1045` |
| SEC-2 → M1 | `worktree-sweep.ts:499-514` | **corrected** | `fs.rm` at `:523`; loop `:509-531` |
| COR-1 → H3 | `worktree-sweep.ts:326-331` | **corrected** | `empty` at `:336-343`; `MAX_TRUSTED_PID_AGE_MS` at `:190`; `ownerLiveness` at `:684-691` |
| ANCHOR | `worktree-occupancy.ts:76` | **corrected** | `:90-91`; sweep `ageMs` at `:597-600` |
| SEC-4/5, COR-2/4/5 | `probe:498-566` | **corrected (diff-offset citations)** | file is 117 lines; real lines `:40, 48-56, 78, 84, 108, 111-114`. Content verbatim-correct at ref → findings stand |
| PERF-1 → M3 | `worktree-sweep.ts:883-885, 336, 756-759` | **corrected** | `:624-625`, `:346`, `:778-781` |
| API-1 → L4 | `foreground-promotion.ts:406-408` | **corrected** | interpolation at `:409` |
| **API-2 → L5** | `schemas.ts:312` | **fabricated as cited** — line 312 at ref is `required: [],` | finding **retained at `low`** only because the orchestrator independently re-derived its substance at `schemas.ts:696`; original citation discarded |
| COR-5 → N1 | `probe:533-534` | **corrected + downgraded** | `:84`; zero production importers → `medium`→`nit` |

**Absence claims (Check B):** probe importers → `confirmed-absent` (2 prod call sites only). Heartbeat×`runSweep` test → `confirmed-absent`. Nested-fixture test → `confirmed-absent`. `describePreserveReason` test → **`false-absent`, finding dropped** (`worktree-managed.test.ts:131-133`).

---

## What was not checked

- **Citations verified inline against branch HEAD `4fd3723`** via `git show <ref>:<file>` / `git grep <ref>`. `interactive/worktree.ts`, `agent/worktree.ts`, `paths.ts` are byte-identical on `main` and the ref (confirmed with `git diff --quiet`), so base reads of those are ref-accurate.
- **Stated intent:** PR #770 title + body — spec-compliance **assessed**.
- **Deviation from the skill's Wave 2:** synthesis was performed inline by the orchestrator rather than dispatched, because the verification manifest with corrected line numbers exists only here and re-serializing it risked reintroducing the citation errors Wave 1.5 had just fixed. The three top-severity claims did go through an independent shadow-verify round (H1, H2, H3), all three confirmed.
- **One shadow verifier's premise required correction:** it read the working tree (`main`), where the probe does not exist, and therefore concluded "no guard anywhere runs `--ignored`." That is true of `main` but false of the ref — `sweep:625` and `worktree-managed.ts:205` are guarded at `4fd3723`. Its *enumeration* of deletion paths was used; its global verdict was discarded.
- **Not run:** `pnpm test`, `pnpm lint`. The author's gate table is taken at face value; the coverage *gaps* in M4/H2/N2 were established by grep, not by executing the suite.
- **Not executed:** the `unref()` behaviour, live `git status --ignored` output against a real worktree, and `core.quotePath` behaviour against a real git binary. L2 and H2's part (1) rest on git's published documentation, not observed output.
- **Not reviewed:** the Telegram surface, the daemon scheduler's sweep cadence, and whether `boot-prune`'s soft-launch bypass is intentional policy.

---

**Blocked**

- **What blocks the merge:** three `high` findings — H1 (session-end cleanup still destroys gitignored `.env`, while this diff's new schema text claims otherwise), H2 (the probe cannot see a non-rebuildable file nested under a rebuildable ignored directory, doubly-latched and untested), H3 (fix #1's stated root cause is unreachable; the reachable variants route to `dead-owner`, which has no age gate).
- **What must change to unblock:** wire `hasNonRebuildableIgnoredFiles` into `src/cli/commands/interactive/worktree.ts` between `:675` and `:704`; classify ignored entries by leaf segment and expand collapsed rebuildable directories before deciding; either drop the heartbeat or make `touchWorktreeOccupancy` atomic (temp+rename) and `await` the initial touch, and correct the PR body's causal claim. M1–M4 are follow-ups, not blockers.
- **What has already been done:** diff resolved from the preflight artifact; two Wave 1 dimension agents dispatched; every `file-state` and blocking citation re-read inline at ref `4fd3723`, producing the manifest above (one citation dropped as fabricated, one absence claim dropped as false, nine citations corrected); three shadow verifiers independently re-derived H1/H2/H3, all confirmed. **Read-only throughout — no file was edited, nothing was committed, pushed, or posted to the PR.**

---

griffinwork40 (commented):

## Review — PR #770 · `fix(worktree): stop the sweep reaping live trees and discarding ignored files (#759)`

**Merge decision: DO NOT MERGE** — one `high` finding sits inside the exact behaviour this PR exists to fix, and the fix is a one-line move.

Regime: full (16 files, +1189/−99). Citations verified inline against head SHA `9767167`. Both `high` findings went through an independent shadow-verify wave that re-derived them from scratch; both came back CONFIRMED.

---

### high · security/correctness · `src/agent/worktree-ignored-patterns.ts:74` · ref:9767167 · citation-type:diff-context

`/(?:^|\/)logs\//` sits in `OPAQUE_REBUILDABLE_DIRS` — the tier reaped **without** any scoped expansion — so a worktree whose ignored content is a `logs/` directory holding hand-authored material is force-removed with its contents.

```js
  /(?:^|\/)\.nyc_output\//,
  /(?:^|\/)logs\//,
];
```

This contradicts the same file's own docblock at `:99-102`, which records why the *file* rule was narrowed to an emitter allowlist:

> The log entries are deliberately an emitter allowlist rather than a bare `/\.log$/`: that suffix matched every ignored `*.log`, so a captured agent transcript or a hand-kept `decisions.log` was classified as build noise and force-deleted along with the checkout.

The directory tier re-opens that hole for everything under `logs/`, and re-opens it *worse* — an `opaque` verdict is never expanded, so `SENSITIVE_LEAF_PATTERNS` never runs on the leaves at all.

Derivation (`classifyIgnoredEntry`, `:156-164`, precedence-ordered):

| entry | class | why |
|---|---|---|
| `logs/` | `opaque` | `:74` matches — and this is what git actually reports at top level |
| `logs/decisions.log` | `opaque` | `:74` matches the prefix |
| `decisions.log` | `protected` | no `/`, so every dir regex misses → falls through `:163` |
| `transcripts/session.log` | `protected` | no tier matches |

Note the asymmetry: **the identical filename `decisions.log` is protected at the repo root but reapable at `logs/decisions.log`.** `worktree-ignored-probe.test.ts:77` pins the reopened hole (`isRebuildableIgnoredEntry('logs/run.log') === true`) directly beside `:71-74`, which asserts the root-level case is protected.

Reachability confirmed to the removal call: `worktree-ignored-probe.ts:120-124` — `logs/` is `opaque`, so the `:122` guard (`verdict === 'inspectable' && entry.endsWith('/')`) never fires and no expansion happens → returns `false` → `worktree-sweep.ts:624-626` leaves `isDirty` false → `classifyCandidate` admits `dead-owner` (`:319-324`) and `empty` (`:336-343`) → `git worktree remove --force` at `worktree-sweep.ts:740` and `:757`.

**Suggestion:** move `/(?:^|\/)logs\//` from `OPAQUE_REBUILDABLE_DIRS` into `INSPECTABLE_REBUILDABLE_DIRS` (`:83-94`) so the directory is expanded and its non-emitter leaves protect the tree. This is a one-line move; the emitter allowlist at `:112-118` already keeps `logs/run.log`-style noise reapable after expansion.

---

### medium · security · `src/agent/worktree-ignored-patterns.ts:34-35` · ref:9767167 · citation-type:diff-context

`SENSITIVE_LEAF_PATTERNS` only recognises a **dot-prefixed** env file, so a non-dotted env leaf is not sensitive and is reaped when it sits under an `inspectable` directory:

```js
  /^\.env$/,
  /^\.env\./,
```

Verified leaf-by-leaf against the real regex set: `.env` → sensitive; `.env.local` → sensitive; `secrets.env` → sensitive (via `/secret/`); **`app.env` → NOT sensitive**; **`prod.env` → NOT sensitive**. So `dist/app.env` survives the expansion at `worktree-ignored-probe.ts:97` and the tree is reaped. The tests only assert the forms that already pass (`worktree-ignored-probe.test.ts:42,60-62` — `.env`, `.env.local`, `.env.production`, `dist/secrets.env`, `dist/.env`), so nothing fails today.

**Suggestion:** add `/\.env$/` to `SENSITIVE_LEAF_PATTERNS` so any `*.env` leaf protects, and extend the `:60` fixture list with `dist/app.env`.

---

### medium · correctness · `src/agent/worktree-ignored-probe.ts:56-77` · ref:9767167 · citation-type:diff-context

`readIgnoredEntries` passes no `maxBuffer` and no `timeout`, so a scoped `--untracked-files=all` expansion of a large `dist/`/`target/` can overflow Node's default `execFile` buffer, reject, and take the `catch → undefined` path → `inspectableDirHidesLocalState` returns `true` (protect). Failing safe is right, but the failure is silent and makes every large-build worktree permanently unreapable — the immortality the module docblock at `:12-19` exists to prevent.

```js
  if (scopePath !== undefined) args.push('--untracked-files=all', '--', scopePath);
  try {
    const { stdout } = await execFile('git', args);
```

Note `ExecFileFn` (`worktree-sweep.ts:30-34`) only accepts `{ cwd?: string }`, so this needs a signature widening, and no `maxBuffer` appears anywhere in the sweep or the two handler call sites.

**Suggestion:** widen `ExecFileFn`'s options to carry `maxBuffer`/`timeout`, pass `{ maxBuffer: 64 * 1024 * 1024, timeout: 10_000 }` on the scoped call, and push a sweep warning when the probe protects because git failed rather than because it found something.

---

### low · perf-observability · `src/agent/worktree-sweep.ts:780` · ref:9767167 · citation-type:file-state

An ignored-state protect is reported as `uncommitted changes` — a false reason for a tree `git status` calls clean:

```js
            `[WARN] stale-dirty worktree preserved (uncommitted changes): ${entry.path}`,
```

The other three removal paths added an explicit label (`worktree-managed.ts:167,205-206,288-290` `ignored-local-state` + `describePreserveReason`; `worktree.ts:319`; `interactive/worktree.ts:720`); the sweep is the one that did not. Pre-existing line, but this PR is what makes it reachable for a clean tree. **Suggestion:** thread the probe's boolean into the warning so it reads `(non-rebuildable ignored files)` when that is what fired.

### low · test-coverage · `src/agent/subagent.test.ts` · ref:9767167 · citation-type:file-state

The PR calls the heartbeat lifecycle "its own bug", but no test covers the wiring: promise (c) *settle disarms* and promise (d) *construction-throw disarms* are unasserted, so deleting `stopOccupancyHeartbeat()` from `subagent.ts:1042` or `:1084` keeps the suite green. Absence verified: `heartbeat|stopOccupancy|unref|occupancy` against `subagent.test.ts` at the ref → **0 matches** (file present, 1669 lines). Promise (a) `unref()` is likewise unasserted in `worktree-occupancy.test.ts` (`unref|rename|atomic|\.tmp` → 0 matches). **Suggestion:** inject a fake `startWorktreeOccupancyHeartbeat` and assert stop fires once on settle and once on a forced construction throw.

### low · correctness · `src/agent/worktree-occupancy.ts:105` · ref:9767167 · citation-type:diff-context

The atomic-write temp path is keyed only on `process.pid`, so two heartbeats armed for the same root in one process share one temp filename. Subagents run in-process and inherit the parent `cwd` verbatim (`subagent.ts:917-919`, armed at `:995-998`), so a worktree-isolated parent fanning out N children yields N identical `tmpPath`s. Shadow-verify confirmed the interleaving is real but the blast radius is small: both payloads carry the same `pid` and equal-width ISO timestamps, so the worst case is a sub-millisecond zero-length window on the meta file, self-healed by the next write — not a lost update. **Suggestion:** add a per-call uniqueness token (`${process.pid}.${randomUUID()}`).

### nit · correctness · `src/agent/worktree-ignored-patterns.ts:111` · ref:9767167 · citation-type:diff-context

`/^Thumbs\.db$/` in `REBUILDABLE_FILE_PATTERNS` is unreachable: `/\.db$/` in `SENSITIVE_LEAF_PATTERNS` (`:47`) is tested first at `:159` and returns `protected`. No test asserts `Thumbs.db` is rebuildable, so nothing reveals the shadowing. Harmless (fails safe), but it is dead code that misleads the next reader.

---

## Verified good

The four heartbeat lifecycle promises all hold as written: `timer.unref()` (`worktree-occupancy.ts:160`), the stop handle captured before `new AgentSession` (`subagent.ts:994` vs `:1011`), settle-path disarm (`:1042`), construction-failure disarm (`:1084`). `DEFAULT_HEARTBEAT_INTERVAL_MS` is 10 min against a 1 h gate, with `worktree-occupancy.test.ts:170` pinning the ≤¼ relationship — a real guard, not a restatement. Argument handling is injection-safe throughout (argv arrays, no shell, `scopePath` after `--`), `core.quotePath=false` is correctly load-bearing for non-ASCII paths, and every probe error path returns the protective verdict (`probe.ts:94,118`). Precedence putting `isSensitiveLeaf` above the rebuildable tiers is the right call and is well documented at `:25-31`. The api-compat surface is clean: `GuardedRemoveOutcome` and `IsolatedTeardownResult['reason']` widen, and both production consumers are updated in this diff (`worktree.ts:319`, `foreground-promotion.ts` via `describePreserveReason`); no tool `input_schema` shape changed. All five claimed test suites exist with the claimed regression cases.

One correction to the PR body's framing: `git status --porcelain --ignored --untracked-files=all -- <dir>` **fully recurses** — verified empirically on git 2.53.0, where a scoped call on `dist/` returned `dist/nested/.env`, not a collapsed `dist/nested/` line. The non-recursive-expansion concern the contract at `probe.ts:83-86` guards against does not arise for nested dirs; the traditional-mode collapse claim for the *top-level* call is correct.

Also noted, not filed as findings: the intent explicitly accepts a meta-less tree staying reapable (documented trade, not an oversight), and the `orphaned-dir` branch (`worktree-sweep.ts:509-531`) still `fs.rm`s recursively with no dirty check, no probe, and no age gate. That branch is untouched by this diff and pre-exists on `main`, so it is out of scope here — but it is a fourth removal path that bypasses the fix, and closing it is the natural follow-up to #759.

## What was not checked

- Citations verified inline against branch HEAD `9767167`; every `file-state` citation was re-read at that ref, and `logs/`-adjacent absence claims were re-grepped at the ref. **The local checkout is on a different branch and does not contain `worktree-ignored-patterns.ts` at all**, so no finding rests on a working-tree read.
- Tests were **not executed** and `tsc` was **not run** — the claimed counts (37/13/43/87/7) and `pnpm lint` clean are unverified.
- The remaining ~40 pattern regexes were not individually audited for over-inclusiveness beyond `logs/`, `Thumbs.db`, and the `*.env` leaf gap; `.cache/`, `.venv/`, `.yarn/` being `opaque` was accepted as plausible but not argued.
- TOCTOU window between the probe and `remove --force` (rev-list and branch-delete calls sit between them) was not analysed.
- Symlink handling under `.afk-worktrees/` for `fs.rm` / `git worktree remove` was not checked.
- Stated intent: PR #770 title + body — spec-compliance assessed against it.

griffinwork40:
Addressed the review from `9767167` in **`251cc1e`**. All findings closed except one deferral, noted at the end.

| finding | status | where |
|---|---|---|
| high · `logs/` reaped unexpanded | fixed | `worktree-ignored-patterns.ts` |
| medium · non-dotted `*.env` not sensitive | fixed | `worktree-ignored-patterns.ts` |
| medium · probe execFile unbounded | fixed | `worktree-ignored-probe.ts`, `worktree-sweep.ts` |
| low · false `uncommitted changes` label | fixed | `worktree-sweep.ts` |
| low · pid-only temp path | fixed | `worktree-occupancy.ts` |
| low · heartbeat lifecycle untested | **partial** — occupancy half done, `subagent.test.ts` half deferred | `worktree-occupancy.test.ts` |
| nit · unreachable `Thumbs.db` | fixed | `worktree-ignored-patterns.ts` |

### One correction to the high finding's rationale
The suggested move is right and is applied, but the stated mechanism is not what the code does. The review says the move makes `logs/`'s "non-emitter leaves protect the tree." They do not: a nested leaf re-matches the same `/(?:^|\/)logs\//` regex and so classifies `inspectable`, not `protected`, and `inspectableDirHidesLocalState` only protects on `=== 'protected'` (`worktree-ignored-probe.ts:97`).

What the move actually fixes — and this is the real bug — is that an `opaque` verdict is **never expanded**, so `SENSITIVE_LEAF_PATTERNS` never ran on `logs/` at all. Verified empirically after the change: `logs/.env`, `logs/app.env`, `logs/prod.key`, `logs/creds.pem`, `logs/data.db` now all protect; `logs/debug.log` stays reapable via the emitter allowlist; `node_modules/` stays opaque so the sweep still reclaims.

`logs/decisions.log` therefore **remains reapable**, so the root-vs-nested asymmetry is narrowed, not eliminated. Closing it fully would require presuming nested leaves non-rebuildable, which is exactly what would make every `dist/`-bearing worktree immortal. Documented at the pattern and pinned by a test named for the choice, so it reads as deliberate rather than missed.

### A race the temp-path fix unmasked
Giving each touch its own temp file exposed a latent bug the collision had been masking: `stop()` only cleared the interval, so a touch already past its first `await` still landed its `rename` and re-stamped `createdAt` on a tree the sweep was about to judge. Previously one of the two concurrent writes just failed, hiding it. `stop()` now sets a flag the touch honours immediately before the `rename` — the only irreversible step — so a cancelled touch leaves the meta byte-identical and takes its temp file with it. This also fixes the load-dependent flake in `stops touching after stop() is called`, which asserted a guarantee the implementation never made.

### Verification
`tsc --noEmit` clean. `pnpm test:coverage` (the actual CI gate, not `pnpm test`) green: **713 files, 13621 tests, 0 failures**, thresholds met. New regression guards: the scoped expansion is asserted to *happen* for `logs/` (the part that fails if it is ever moved back to the opaque tier), `maxBuffer`/`timeout` are asserted on every probe call, and the sweep warning is asserted to say `non-rebuildable ignored files` and **not** `uncommitted changes`.

### Deferred
`subagent.test.ts` heartbeat wiring — promise (c) settle-disarm and (d) construction-throw-disarm — is still unasserted; deleting `stopOccupancyHeartbeat()` from `subagent.ts:1042` or `:1084` keeps the suite green. Additive test-only work, no behaviour risk, so it is not gating this PR.

### Follow-up that this PR does NOT close (confirmed independently)
The review flagged the `orphaned-dir` branch as out of scope; it is worth stating plainly, because it is a **fourth removal path that bypasses everything above**. `worktree-sweep.ts:549` runs `fs.rm(orphanPath, { recursive: true, force: true })` on any directory under `.afk-worktrees/` absent from `git worktree list`, with no dirty check, no ignored-file probe, and **no age gate** — `orphanAgeMs` is computed and attached to the candidate but never compared to a threshold. So a tree whose git registration is missing or pruned is deleted outright with its uncommitted contents. Pre-existing on `main` and untouched here, but it should be the next issue after #759.

griffinwork40:
# Post-merge review — PR #770 (round 3)

**Framing, up front:** this PR merged at `2026-07-31T15:24Z` as `16c4cd9` and is already published in **v5.83.3** (`117ba77` sits on top of it). So there is no merge verdict to give — this is a **residual-risk review of shipped code**, and every finding below is a follow-up candidate rather than a merge blocker.

Reviewed ref: **`251cc1e`** (PR head, identical to the merged content). Review workspace was checked out at exactly that ref, clean. Findings were derived independently — this round was run without showing the agents the earlier rounds' conclusions — then a verification wave re-derived the load-bearing claims **by execution** (real `git` 2.53.0 fixtures, a live `runSweep`, and the actual classifier invoked via `tsx`). That wave **refuted or corrected three claims**, which are disclosed at the bottom rather than quietly dropped.

**Bottom line:** one **shipped regression introduced by this PR** (H1) worth a fast follow-up, one **pre-existing high** (H2) that is the natural next issue after #759, plus two mediums. Nothing here is a data-loss bug in the fix itself — H1 fails in the *safe* direction and costs the sweep instead.

---

## H1 · high · correctness — four `^`-anchored rebuildable patterns are tested against the full path, so a *nested* `.DS_Store` protects a worktree forever

`ref: 251cc1e` · `src/agent/worktree-ignored-patterns.ts:123-127` · citation-type: file-state

`REBUILDABLE_FILE_PATTERNS` mixes two anchoring styles, and the entries are matched against the **full normalized path** (`:178`), not the leaf:

```ts
/^\.afk-worktree-meta\.json$/,
/\.tsbuildinfo$/,
/^\.eslintcache$/,
/^\.stylelintcache$/,
/^\.DS_Store$/,
```

The sibling log emitters twelve lines down do it the other way, and correctly match at any depth:

```ts
/(?:^|\/)debug\.log$/,
/(?:^|\/)npm-debug\.log$/,
```

Four entries (`:123`, `:125`, `:126`, `:127`) are bare-`^`; only `/\.tsbuildinfo$/` (`:124`) is suffix-matched and therefore depth-safe. A nested copy matches no rebuildable file pattern, matches no rebuildable *directory* pattern, and hits the fall-through at `:181` → `protected`.

**Executed against the real classifier, not read off the regex:**

```
".DS_Store"                  classify=opaque       isRebuildable=true
"src/.DS_Store"              classify=protected    isRebuildable=false
"a/b/.DS_Store"              classify=protected    isRebuildable=false
".eslintcache"               classify=opaque       isRebuildable=true
"packages/app/.eslintcache"  classify=protected    isRebuildable=false
"deep/.stylelintcache"       classify=protected    isRebuildable=false
```

**And end-to-end through a real `runSweep`** on a fixture of six 40-day-old worktrees — only the nested-`.DS_Store` tree survived:

```
removed:  [ 'clean-ctl', 'decisions-log', 'logs-only', 'nm-only', 'root-ds' ]
survived: [ 'nested-ds(stale-dirty)' ]
warnings: ["[WARN] stale-dirty worktree preserved (non-rebuildable ignored files): .../nested-ds"]
```

Note `root-ds` was reaped and `nested-ds` was not — same filename, different depth.

**Why it persists:** the resulting `stale-dirty` verdict is warn-only (`worktree-sweep.ts:825-829`) and has no age escape, so the tree is never auto-reaped again.

**Reachability, stated precisely.** `git status --porcelain --ignored` lists an ignored file individually — rather than collapsing to a directory line — whenever its directory also holds a tracked file. Verified on git 2.53.0:

```
!! logs/
!! node_modules/
!! src/.DS_Store        <- src/ holds tracked src/keep.ts, so no collapse
```

`.DS_Store` is gitignored in this repo (`.gitignore:25`), so the path is live here. **It is latent today, not currently firing:** a scan of the 14 live trees under `.afk-worktrees/` found **0** nested `.DS_Store` files. One Finder visit to any subdirectory of a worktree arms it.

**Blast radius, corrected.** This does **not** make a worktree undeletable — `worktree remove --force` still works, because `removeManagedWorktreeGuarded` gates the probe behind `if (!force)` (`worktree-managed.ts:202-208`). What it defeats is the **automatic** sweep, so the cost is unbounded accumulation of never-auto-reaped worktrees — precisely the leak the sweep exists to prevent.

**suggestion** — Change the four bare-`^` entries to `(?:^|\/)`, matching `:131-136`; or test `REBUILDABLE_FILE_PATTERNS` against `leafOf(normalized)` instead of `normalized`. Either fixes H1 with no effect on the `logs/` behavior discussed below. No current test pins nested-leaf behavior in either direction, so a regression guard should land with the fix.

---

## H2 · high · security — the `orphaned-dir` branch deletes with no work-preservation guard *(pre-existing, not from this PR)*

`ref: 251cc1e` · `src/agent/worktree-sweep.ts:535-557` · citation-type: file-state

```ts
        if (!effectiveDryRun) {
          try {
            await fs.rm(orphanPath, { recursive: true, force: true });
```

`orphanAgeMs` is computed at `:536`/`:539` and attached to the candidate at `:545` — and **never compared to anything**. Zero occurrences sit adjacent to `>`, `<`, `>=`, or `<=`. It is telemetry; it gates nothing.

No dirty check and no ignored-file probe apply here: the only probe call site in this file is `:656`, inside the *registered-worktree* loop that begins at `:589`, and `git status --porcelain` is likewise at `:639` in that same loop. So any directory under `.afk-worktrees/` that is absent from `git worktree list` is deleted outright with its uncommitted contents — bypassing every guard this PR added.

**Pre-existing, confirmed:** the block is byte-identical at the merge-base `a25984a`, `git log -L 527,558:src/agent/worktree-sweep.ts a25984a..HEAD` is empty, and none of the PR's seven hunks overlap it. Introduced in `0a23cff`. `9767167`'s own commit message already names it as deliberately out of scope.

**Two corrections to how this was described in the previous round.** It is the **third** destructive path inside `worktree-sweep.ts` (`empty` `:787`, `dead-owner` `:804`, orphan `:549`) — "fourth" only holds if you count the three probe-guarded call sites codebase-wide. And "bypasses every guard" overstates: scope (`:534`) and dry-run (`:547`) do apply — they just preserve no work.

**suggestion** — Gate the `fs.rm` behind an age threshold plus the probe. **Caveat for whoever takes it:** orphaned directories are *unregistered*, so reusing the `:639-672` dirty/probe block verbatim is not a drop-in — `git -C <orphanPath> status` resolves against the parent repo rather than the orphan.

---

## M1 · medium · correctness — a scoped expansion that yields zero entries falls **open**

`ref: 251cc1e` · `src/agent/worktree-ignored-probe.ts:114-129` · citation-type: diff-context

```ts
  }
  return { protect: false };
}
```

When the scoped expansion of an `inspectable` directory returns no `!!` lines, "the expansion produced nothing" is treated as "nothing sensitive is inside" rather than as unknown — inverting the module's documented fail-safe posture on the one branch that decides whether an already-`inspectable` directory gets reaped. The adjacent `'failure' in nested` branch (`:120-122`) protects, which is the correct direction.

**suggestion** — Return `{ protect: true, because: 'git-failed', … }` when `nested.entries.length === 0` for a directory git reported as populated. *(Derived by reading; not independently re-verified by execution in the verification wave — treat as medium confidence.)*

---

## M2 · medium · perf-observability — `target/` in the inspectable tier converts a slow enumeration into permanent protection

`ref: 251cc1e` · `src/agent/worktree-ignored-probe.ts:47-55,150-157` · citation-type: file-state

```ts
const PROBE_EXEC_OPTS = { maxBuffer: 64 * 1024 * 1024, timeout: 10_000 } as const;
```

Every `inspectable` directory pays a `--untracked-files=all` enumeration of everything beneath it, once per collapsed dir with no aggregate cap (`:153-156`, driven from `worktree-sweep.ts:656`). A Rust `target/` is not small build output. When the 10s timeout trips, the verdict is `git-failed`, which **protects** — landing in the same never-auto-reaped state as H1, via a different route. The module docblock at `:141` claims the opposite: *"Only directories classified 'inspectable' — small build output, not dependency trees — pay for a second, scoped call."*

**suggestion** — Move `target/` to `OPAQUE_REBUILDABLE_DIRS` with the other machine-owned trees, or cap total scoped expansions per worktree.

---

## Lower severity

- **L1 · low · security — TOCTOU between probe and removal.** The probe runs at `worktree-sweep.ts:656`; `git worktree remove --force` runs at `:787`, four subprocess round-trips later (`rev-list` ×2, liveness, classify). A `.env` written into the tree inside that window is deleted. *suggestion:* re-probe immediately before the removal in the `empty` and `dead-owner` branches.
- **L2 · low · security — no `--literal-pathspecs` on the scoped expansion** (`worktree-ignored-probe.ts:84-89`). A git-reported path is fed back to git as a pathspec, so a name carrying pathspec magic is interpreted as magic; a magic-scoped expansion matching nothing then reaches the fail-open return in M1. Grep for `literal-pathspecs` across `src/` → zero matches. `[low confidence — the missing flag and the fail-open return are both verified; the join between them is not demonstrated end-to-end]`
- **L3 · low · test-coverage — the heartbeat's `forkSubagent` wiring is traversed but never asserted.** `subagent-worktree-readroot.test.ts:68,96-97` and `subagent-dispatcher-readroot.e2e.test.ts:218-219,261` both reach `subagent.ts:997` through `forkSubagent`, so the arm site *is* executed — but no assertion observes heartbeat state, and neither disarm is pinned (`:1042` settle callback, `:1084` construction-throw). This matches the author's own deferral note. *suggestion:* spy `setInterval`/`clearInterval` around a fork with an `.afk-worktrees/` cwd and assert the clear fires on both settle and a constructor throw.

---

## Correction to the PR body — Fix 1's problem statement does not match the shipped code

Not a code defect; it matters because it is the stated justification for the change, and #759's write-up will be read later as the record.

The body says: *"Any subagent running longer than that aged back into `empty` while still working, and `git worktree remove --force` took its checkout out from under it."*

The shipped code contradicts this for a correctly-stamped tree. `empty` requires `ownerLiveness !== 'alive'` (`worktree-sweep.ts:362-369`); a forked subagent runs **in-process**, so the stamped `pid` resolves alive for the child's entire run; and the only window that can expire is `MAX_TRUSTED_PID_AGE_MS = 30 days` (`:216`), not the 1h `MIN_EMPTY_AGE_MS`. The module's own docblock states this explicitly (`worktree-occupancy.ts:20-32`), and so does a test comment (`worktree-sweep.test.ts:2003-2005`).

**The heartbeat is still justified** — just not for the reason given. Per that same docblock, when the meta is absent, unparseable, or torn by a concurrent read, liveness degrades to `unknown` and the 1h age gate becomes the only defence; re-stamping on a 10-minute cadence is what holds the tree. Fix 1 is defence-in-depth for the degraded states, and the body generalized a degraded-state hazard into the nominal case.

---

## Dropped or corrected by the verification wave

Disclosed rather than silently removed, because each was a real candidate finding in Wave 1:

| Wave-1 claim | Outcome |
|---|---|
| `logs/decisions.log` reapable while root `decisions.log` protects | **Dropped as a finding.** Empirically real (`logs/decisions.log` → `inspectable`/reapable; `decisions.log` → `protected`), but it is a documented deliberate trade-off at `worktree-ignored-patterns.ts:97-110` — *"That residual asymmetry is deliberate and bounded to non-sensitive names"* — and sensitive leaves under `logs/` all protect correctly (`logs/.env`, `logs/app.env`, `logs/prod.key` → `protected`). A design question, not a defect. |
| `out/`/`build/` are generic names, so nested hand-authored files get reaped | **Dropped**, same declared trade-off class as above. |
| Nested `.DS_Store` makes a tree "unreapable by all removal paths" | **Corrected.** `--force` escapes (`worktree-managed.ts:202-208`). Restated in H1 as *permanently unreapable by the automatic sweep*. |
| "No test exercises the heartbeat wiring in `forkSubagent`" | **Refuted as worded.** Two suites do traverse it; restated as L3 — traversed but unasserted. |
| `MIN_EMPTY_AGE_MS` visibility widened to exported | **Dropped to nit and omitted** — zero production importers, test-only consumer. |

---

## What was not checked

- **Ref read:** all citations were read from a worktree checked out at `251cc1e` (clean), so file-state citations resolve at the PR head / merged content. No patch-file-only citations.
- **Stated intent:** PR #770 title + body — **spec-compliance was assessed** against it.
- **No perf measurement.** M2's timeout risk is a code-path argument; real `-uall` enumeration cost on a large `target/` was not measured.
- **No mutation testing.** L3's "deleting either disarm keeps the suite green" is inference from zero relevant assertions plus the `unref()`'d 600s timer, not a measured mutation run.
- **The `--force` bypass at `src/cli/commands/interactive/worktree.ts:640-666`** (zero-turn session-end cleanup skips both dirty and ignored probes) was read but not reported: whether a zero-turn session can have written to the worktree was not established.
- **Sensitive-leaf list completeness** was not audited beyond the entries traced — `.p8`, `.ovpn`, `.kdbx`, `terraform.tfstate` and similar were not assessed.
- **Windows path behavior** of the normalizer was not exercised.

---

_🤖 Posted by `agent-afk /review --post github`. Read-only review: no files, branches, or PR state were modified._

PR #771, opened 2026-07-30
fix(worktree): sweep every repo root known to hold managed trees (#761)

### LINKED ISSUE #761: worktree sweep only ever reaches one repo root — managed trees under other roots are never reclaimed
## Symptom

The worktree sweep can only ever see **one** repo root per run. Any afk-managed worktree that lives under a different root is invisible to every sweep path, so it is never reclaimed — it accumulates indefinitely, along with its disk usage.

## Cause

The sweep anchors on `join(repoRoot, '.afk-worktrees')` (`worktree-sweep.ts:479`) and explicitly skips anything outside it (`:557`). The root itself comes from exactly three callers, each supplying a single root:

- `daemon/scheduler.ts:62-75` — `AFK_WORKTREE_SWEEP_ROOT` if set, else `git rev-parse --show-toplevel` of `process.cwd()`
- `boot-prune.ts:92`
- the `worktree` tool's own `list` action (`tools/handlers/worktree.ts:271-276`)

So reclamation is contingent on an afk session (or daemon) later running with its cwd inside that same repo. For a repo where afk is used once and never again, the trees leak permanently.

## Observed impact

Real-world: a personal checkout accumulated 14 unswept managed worktrees (~1.8 G) because the daemon's sweep root pointed elsewhere. Removal is safe in principle — branch refs live in the shared `.git`, so only uncommitted working-tree files are ever at stake — but nothing was triggering it.

## Suggested fix

Give the sweep a multi-root mode: persist the set of repo roots afk has created managed worktrees under (a small registry under `~/.afk/state/`, written at create time by `createManagedWorktree`), and have the daemon sweep every registered root rather than only its own cwd's root. Alternatively accept `AFK_WORKTREE_SWEEP_ROOT` as a colon-separated list, which is a much smaller change but leaves the default behaviour leaky.

Also worth surfacing: `worktree list` reports only the current root's trees, so an operator has no single command that shows every managed worktree on the machine.

griffinwork40:
Fixes #761.

## Problem

The sweep engine is **per-root**: it anchors on `<repoRoot>/.afk-worktrees` and skips anything outside it. The daemon's prune tick resolved exactly one root — from its own cwd, or `AFK_WORKTREE_SWEEP_ROOT` — and swept only that. Managed worktrees created under any *other* repo were therefore reachable by **no** sweep, and leaked permanently.

This is not hypothetical: it is the mechanism behind long-lived accumulations of unswept managed trees in repos the daemon never happens to sit in.

## Shape chosen: a state registry

Per the issue's proposal, roots are recorded rather than configured:

- New `worktree-root-registry.ts` maintains `$AFK_STATE_DIR/worktree-roots.json` (`{ version, roots: [{ path, lastSeenAt }] }`).
- Registration happens at **each managed-create path**, of which there are two: `createManagedWorktree()` — which the `worktree` tool's `create` action and `createIsolatedWorktree()` (the `agent` tool's `isolation: "worktree"` path) both funnel through — and the `afk -w` session launcher in `cli/commands/interactive/worktree.ts`, which builds its tree with its own `git worktree add`. An earlier revision of this description claimed the first was the *single* chokepoint; review correctly falsified that, and the launcher is now wired too. Any future managed-create path must register as well — the comment at the `createManagedWorktree` call site says so.
- The daemon tick sweeps `sweepRootSet(primary)` — its own resolved root followed by every registered root, de-duplicated — and aggregates removals, warnings and candidates into the existing telemetry summary.

The alternative shape (letting `AFK_WORKTREE_SWEEP_ROOT` take a list) was not taken: it makes reclamation depend on the operator remembering to enumerate repos, which is the failure mode the bug already demonstrates.

## Safety properties

These are the parts worth reviewing closely, since this writes to user state from inside a create path:

- **Never fails a create.** Every registry operation is best-effort and swallows errors. A missed registration degrades to exactly the previous single-root behaviour. There is an explicit test that an unwritable state dir resolves rather than throws.
- **Self-healing, bounded.** Roots are pruned *on read*: an entry whose directory no longer exists (repo deleted or moved) is dropped and the pruned file is persisted. A corrupt or unparseable file reads as empty instead of throwing. Retention is capped at 64 roots, oldest-seen dropped first, so the file cannot grow without bound.
- **Explicit target always wins.** A supplied primary root is visited first and is included even when unregistered, so `AFK_WORKTREE_SWEEP_ROOT` never depends on bookkeeping.
- **Sequential sweeps.** `runSweep` takes a single machine-global advisory lock, so sweeping roots in parallel would only contend and let losers short-circuit. The loop is deliberately serial.
- **Behaviour when nothing is known is the same skip path**, with the same `status: 'skipped'` and no sweep attempted. The telemetry *message* changed on review: an empty root set means the daemon has no primary root AND nothing is registered, so it now names both causes instead of blaming daemon cwd alone.

## Verification

| Gate | Result |
|---|---|
| `pnpm test src/agent/worktree-root-registry.test.ts` | 21 passed (new) |
| `pnpm test src/agent/worktree-sweep.test.ts` | 50 passed |
| `pnpm test src/agent/daemon/scheduler-worktree-prune.test.ts` | 6 passed (new) |
| `pnpm test src/agent/daemon/scheduler.test.ts` | 30 passed |
| `pnpm test src/cli/commands/interactive/worktree-root-registration.test.ts` | 2 passed (new) |
| `pnpm test src/cli/commands/interactive/worktree.test.ts` | 58 passed |
| `pnpm test src/agent/tools/handlers/worktree-managed.test.ts` | 16 passed |
| `pnpm test src/agent/tools/handlers/worktree.test.ts` | 34 passed |
| `pnpm test` (full suite) | 14,008 passed / 731 files |
| `pnpm audit:sdk:check` | pass |

New coverage: idempotent registration, accumulation across distinct roots, pruning a vanished root *and* persisting the prune, corrupt-file tolerance, a path that exists but is a file, best-effort behaviour on an unwritable state dir, and the primary-first / de-duplication / no-primary / empty cases of `sweepRootSet`.

No new env var is introduced, so the generated `docs/env-registry.*` are untouched. Test isolation is already guaranteed by `src/__test-utils__/redirect-paths-env.ts`, which repoints `AFK_HOME` per test file — verified that a full run of the touched suites leaves no `~/.afk/state/worktree-roots.json` behind.

---

## Review remediation (2026-08-01)

All findings from the two `/review` passes and the Codex inline comments are addressed — see the per-item table in the PR comments. Summary of behaviour changes beyond the original design:

- Registry writes are serialized (in-process promise chain, then a **degradable** advisory lock that yields after 2s and writes anyway rather than blocking a create) and atomic (temp-file + rename), at `0o600`.
- The daemon's sweep loop isolates failures per root; one unusable root no longer starves the roots behind it.
- The soft-launch dry-run valve is now **per-root** (`<repoRoot>/.afk-worktrees/.sweep-runs`) instead of machine-global, so a newly registered repo gets its three previews instead of inheriting an exhausted counter and being swept destructively on first contact.
- `.afk-worktrees` containment uses a path-boundary test; the previous `startsWith` accepted a sibling like `.afk-worktrees-scratch`.
- Merged `main` (conflict was an adjacent-import collision in `worktree-managed.ts`; both imports kept).


griffinwork40 (commented):

**Merge decision: DO NOT MERGE** — 2 blocking findings.

Reviewed ref `45b751ac9a2baca06c1cfcd9c6a57c517d22f4c6`. Dimensions: security, api-compat, correctness, spec-compliance, test-coverage, perf-observability. Every `file:line` below was re-read at that ref; both `high` findings were independently re-derived by adversarial verifiers.

The shape of this change is right, and the safety framing in the description is unusually careful. The two blockers are both about what happens at the *seams* the multi-root fan-out newly creates.

---

## Blocking

### 1. high · correctness · `src/agent/daemon/scheduler.ts:495-508`

One bad root starves the sweep for every root ordered after it. The new `for` loop has no per-iteration error handling, and `runSweep` does not catch a failure of its own opening `git -C <repoRoot> worktree list --porcelain` — that call sits in a `try` whose only handler is a `finally` that releases the lock (`worktree-sweep.ts:473-475`, `:778-780`; the `catch` at `:465` belongs to the already-closed lock-acquisition try). `builtinPruneExecFile` is a bare `promisify(execFileCallback)` (`scheduler.ts:52`), so a nonzero git exit rejects, escapes `runSweep`, escapes the loop, and lands in the outer `catch` at `:536`, which records `status: 'error'` and returns.

```ts
for (const repoRoot of roots) {
  results.push(await runSweep({
    execFile: builtinPruneExecFile,
    repoRoot,
```

Reachability is concrete: the registry's entire liveness gate is `stat().isDirectory()` (`worktree-root-registry.ts:110-112`) — nothing checks the path is still a *git* repo. A directory whose `.git` was deleted, or whose path was reused by a non-repo, persists across ticks (self-heal at `:120-130` prunes only *vanished* paths), so the failure is sticky, not transient. That reintroduces the leak #761 exists to close.

Blast radius is new: before this PR there was exactly one root, so a throw could only affect that one. Note also that the primary root *is* git-validated each tick via `resolveWorktreePruneRoot` and is swept first — so the exposure is specifically the newly-registered roots. Second-order: on a throw, the earlier roots' `removed[]` are discarded from telemetry even though those removals already hit disk.

**Suggestion:** wrap the per-root `runSweep` in `try/catch` inside the loop, push `[ERROR] sweep failed for <root>: …` into the aggregated warnings, and continue.

### 2. high · spec-compliance · `src/cli/commands/interactive/worktree.ts:579,616-619,755`

Unmet intent. The description's central claim is that `createManagedWorktree()` is "the single chokepoint that BOTH the `worktree` tool's `create` action AND `createIsolatedWorktree()` funnel through." That claim is true of the two paths it names (`worktree.ts:215`; `subagent-executor.ts:555` → `worktree-managed.ts:246`). But a third production path creates a first-class managed tree and never registers the root:

```ts
579: const worktreePath = join(repoRoot, '.afk-worktrees', slug);
616: const addArgs = ['-C', repoRoot, 'worktree', 'add', '-b', branch, worktreePath];
755: await fs.writeFile(join(worktreePath, '.afk-worktree-meta.json'), …
```

It writes the meta with `owner: 'interactive'` — the exact protocol `worktree-sweep.ts:562` consumes. Confirmed negative: `git grep 'worktree-root-registry\|registerWorktreeRoot' 45b751ac -- src/cli/commands/interactive/` returns zero matches. Production-reachable, no experimental gate: `-w, --worktree` at `interactive.ts:210`, and `interactive` is Commander's `isDefault` (`:187`).

**Fairness caveat that narrows this:** `interactive/boot-prune.ts:64-70` re-resolves the repo root on every REPL launch and sweeps it registry-independently, so a repo the user keeps returning to via `afk -w` still gets reclaimed locally. The unclosed hole is *cross-repo daemon* reclamation for repos only ever touched via `afk -w`. Worth noting the shipped comment at `worktree-managed.ts:99-104` ("the single chokepoint **every** managed create funnels through") is strictly false at this ref.

**Suggestion:** call `registerWorktreeRoot(repoRoot)` alongside the meta write at `interactive/worktree.ts:755`, or route that path through `createManagedWorktree`.

---

## Non-blocking

- **medium · correctness · `worktree-root-registry.ts:76-90`** — read-modify-write with a non-atomic `fs.writeFile` and no lock. Two concurrent registrations of *different* roots race and one is lost; a process death mid-write truncates the JSON, which `parseRegistry` then reads as `[]`, silently discarding every registered root and restoring the #761 leak. Fix: write `<target>.tmp.<pid>` then `fs.rename`.
- **medium · security · `worktree-sweep.ts:355-375`** — the soft-launch dry-run valve counts prior runs from one machine-global telemetry file keyed on `taskId === 'worktree-prune'`, not per root. A root registered today inherits an already-exhausted counter and gets a **live destructive sweep on first contact**, never the 3-run dry-run preview the valve exists to provide. With one root that guarantee held. Fix: record the root on the telemetry line and count per-root.
- **medium · security · `worktree-root-registry.ts:87-88,124`** — the registry is written with no `mode`, so the default umask leaves the absolute path of every repo the user works in world-readable. The codebase already treats this data class as `0o600` in `awareness/presence.ts:240-246`, with a comment naming this exact failure. Fix: `{ encoding: 'utf-8', mode: 0o600 }` on both writes, `{ recursive: true, mode: 0o700 }` on the mkdir.
- **medium · test-coverage · `src/agent/daemon/scheduler.test.ts`** — the multi-root loop, the `roots.length === 0` guard, and the `flatMap` merge ship untested. `git grep -c sweepRootSet 45b751ac -- src/agent/daemon/scheduler.test.ts` → 0; the file's only sweep-adjacent coverage is `resolveWorktreePruneRoot` (`:947`). The "scheduler.test.ts 30 passed" row attests to pre-existing tests that never touch the new path. Fix: stub `sweepRootSet` to return two roots, assert one `runSweep` per root and that `removed`/`warnings` concatenate.
- **low · test-coverage · `worktree-root-registry.test.ts`** — no test drives `MAX_ROOTS = 64` or the oldest-first eviction that `.slice(-MAX_ROOTS)` depends on (confirmed: 0 matches for `MAX_ROOTS`/`64` in the new test file). Fix: register 65 roots, assert length 64 with the first absent.
- **low · api-compat · `scheduler.ts:482-484`** — the skip excerpt now fires for two distinct causes but names one; an operator with an empty registry reads a diagnosis about daemon cwd that does not apply. Fix: branch the excerpt on `primaryRoot === null` vs. empty registry.
- **low · perf-observability · `scheduler.ts:524-526`** — the summary is the only field persisted, and now reports bare cross-root totals, so `removed 7, warned 3` cannot be attributed to a root. With one root attribution was implicit; with up to 64 it is lost. Fix: append `rootsSwept: N` or emit one record per root.
- **low · security · `worktree-sweep.ts:557`** — `.afk-worktrees` containment is a raw `startsWith` prefix, not a path-boundary test, so a sibling `<root>/.afk-worktrees-scratch` passes the guard. Pre-existing, but this PR multiplies the roots where it holds. `isPathWithin` at `:222` already does this correctly.
- **nit · `scheduler.ts:497,510`** — `const results = []` drops the `SweepResult[]` annotation that would catch a future field addition to the hand-rolled aggregate; `dryRun: results.some(...)` would label a tick dry-run when only one root previewed. Unreachable today since the valve resolves uniformly per tick.

## Verified, not defects

- **Aggregate shape is correct.** `SweepResult` is exactly `{removed, warnings, dryRun, candidates}` (`worktree-sweep.ts:150-155`), and the only downstream reads — `result.dryRun`, `result.candidates.filter`, `result.removed.length`, `result.warnings.length` (`scheduler.ts:524-526`) — are all satisfied. No field is silently dropped. `const results = []` is TypeScript's evolving-array form, so `noImplicitAny` is not defeated.
- **Null-skip semantics preserved.** `sweepRootSet(null)` with an empty registry returns `[]`, taking the same skip branch; the skip body, its `responseExcerpt`, and its `return skipped` are untouched (`scheduler.ts:474-488`).
- **No argv injection.** Registry paths reach git only through `execFile` (no shell) and always in the value position after `-C`, so a leading-`-` path is consumed as an operand.
- **PII premise partially refuted.** The prune task's telemetry writes counts, not paths (`scheduler.ts:526`) — worktree paths do not newly flow into telemetry. The registry file itself is the only new on-disk path exposure.
- **"Cannot fail the create" holds.** `resolve(repoRoot)` outside the try cannot throw for a `string`, and both call sites pass a typed `repoRoot`.
- **The 14-test claim reconciles exactly** — 5 + 4 + 5 across the three describe blocks, verified at the ref.

## What was not checked

Citations verified inline against branch HEAD `45b751ac9a2baca06c1cfcd9c6a57c517d22f4c6`; 0 fabricated, 0 dropped, 3 absence claims independently confirmed at the ref. Stated intent: PR title + body — spec-compliance assessed against it. Tests were not executed and CI was not re-run (CI is green on this head). That `git -C <non-repo>` exits nonzero is standard git behavior, asserted rather than read from source. Not examined: the Telegram surface, `worktree-sweep.ts` internals beyond the paths cited, and whether a stale registered root nested inside *another* repo would succeed and sweep that outer repo instead of failing.

griffinwork40 (commented):

One blocking finding: the load-bearing "single chokepoint" claim in the PR body is falsified by a third managed-create path, which leaves #761 partially unfixed for the `afk -w` launcher. Everything else is sound — the module is careful, the best-effort contract holds, and the 145-line suite is genuinely substantial. Citations below were verified inline against branch HEAD `45b751ac`.

---

### 🔴 high · spec-compliance / correctness — a third create path never registers its root

`src/cli/commands/interactive/worktree.ts:579,616,619` · ref:45b751ac

The PR body states registration happens at "the single chokepoint that **both** the `worktree` tool's `create` action **and** `createIsolatedWorktree()` funnel through." There is a **third** production path: the `afk -w` session launcher builds an identically-shaped managed tree with a direct `git worktree add`, never calling `createManagedWorktree` and therefore never `registerWorktreeRoot`.

```ts
const worktreePath = join(repoRoot, '.afk-worktrees', slug);   // :579
const addArgs = ['-C', repoRoot, 'worktree', 'add', '-b', branch, worktreePath];  // :616
await execFile('git', addArgs);                                 // :619
```

A grep of that file for `createManagedWorktree|registerWorktreeRoot` returns **zero** matches (verified at the ref). The tree lands under `.afk-worktrees/`, so the sweep engine *would* reclaim it — the root just never enters the registry.

Mitigations weighed, and why they do not close it: that path has its own session-shutdown `cleanup()`, and `boot-prune.ts` sweeps the cwd repo — but `cleanup()` does not run for a crashed or `kill -9`'d session, and boot-prune only fires when the user later reopens a REPL *in that same repo*. A `-w` session killed in a repo the daemon never sits in and the user does not revisit is exactly the #761 leak the PR sets out to eliminate.

**Suggestion:** call `registerWorktreeRoot(repoRoot)` right after the `git worktree add` succeeds in `createWorktreeAt`, or route that path through `createManagedWorktree`. If the gap is deliberate, say so explicitly in the PR body — the "single chokepoint" wording is what makes this blocking rather than a nit.

---

### 🟠 high · correctness — unlocked read-modify-write drops a root under parallel creates

`src/agent/worktree-root-registry.ts:80-90` · ref:45b751ac

```ts
const entries = await readEntries();
const others = entries.filter((e) => resolve(e.path) !== absolute);
const next = [...others, { path: absolute, lastSeenAt: now }].slice(-MAX_ROOTS);
await fs.writeFile(target, JSON.stringify(payload, null, 2), 'utf-8');
```

No lock, and `fs.writeFile` is not temp-file + rename. Two concurrent creates each read the same snapshot and the later write silently discards the earlier one's new root — and a dropped root means its trees leak, reintroducing the bug being fixed.

This is reachable, not theoretical: `dispatcher.ts:989-1010` runs agent forks concurrently under `settleWithConcurrencyLimit`, and each `isolation:"worktree"` child calls `createIsolatedWorktree` → `createManagedWorktree` → `registerWorktreeRoot` independently (`subagent-executor.ts:555`). The registry module contains no `lock`/`mutex` reference at all. The same race also exists between a create and the self-heal rewrite in `readRegisteredWorktreeRoots` (`:129`).

**Suggestion:** serialize the read-modify-write behind the existing advisory lock (`getWorktreeSweepLockPath`) or a dedicated lock file, and write via temp-file + `rename` so a torn write is impossible.

---

### 🟡 medium · correctness — no per-root error isolation in the sweep loop

`src/agent/daemon/scheduler.ts:497-508` · ref:45b751ac

```ts
const results = [];
for (const repoRoot of roots) {
  results.push(await runSweep({ execFile: builtinPruneExecFile, repoRoot, ... }));
}
```

The loop has no try/catch; the only handler is the function-wide one (`:465` → `:536`). One `runSweep` rejection — a corrupt repo, a git lock, a permissions error in *any* registered root — aborts the whole tick and starves every remaining root, recording a single `error`. The single-root design could not starve anything, so this is a new availability failure mode that scales with root count.

**Suggestion:** wrap each `runSweep` in try/catch, collect per-root failures into `warnings`, and continue.

---

### 🟡 medium · correctness — re-registration reorders, so the 64-cap can evict a live root

`src/agent/worktree-root-registry.ts:82-84` · ref:45b751ac

Re-registering moves an entry to the end of the array, and `.slice(-MAX_ROOTS)` drops from the front. The doc comment says "oldest-seen dropped first," which holds for *last-seen* order — but a cold-yet-live root with existing worktrees can be pushed past the cap by 64 busier repos and then silently vanish from every future sweep. Requires >64 distinct roots on one machine: reachable, not common-case.

**Suggestion:** emit a telemetry event when the cap evicts an entry, so a silently-dropped root is diagnosable rather than invisible.

---

### 🟡 medium · perf-observability — tick time scales linearly with roots, unmeasured

`src/agent/daemon/scheduler.ts:495-508` · ref:45b751ac

N sequential `runSweep` calls, each taking the machine-global advisory lock and shelling out to git, grow tick wall-clock linearly — up to 64× at the cap — with no telemetry field recording root count or per-root duration. The serial choice is right (the comment's lock reasoning is correct); the missing signal is the issue: nothing warns before a tick overruns its cron interval.

**Suggestion:** record `rootCount` and per-root durations in the telemetry record.

---

### 🟡 medium · test-coverage — the new multi-root wiring is untested

`src/agent/daemon/scheduler.ts:497-514` · ref:45b751ac

Verified at the ref: `scheduler.test.ts` has **no** worktree-prune test at all (zero hits for `runBuiltinWorktreePrune|worktree-prune|sweepRootSet`), and the PR does not touch that file. So the sequential loop, the error propagation above, and the `flatMap` aggregation all ship uncovered. The new registry suite also has no test for MAX_ROOTS eviction or concurrent registration (zero hits for `MAX_ROOTS|64|evict|concurrent|Promise.all`).

**Suggestion:** add a scheduler test with two stubbed roots where the first `runSweep` rejects, asserting the second is still swept; add a 65-root eviction test and a two-concurrent-`registerWorktreeRoot` test.

---

### 🔵 low · perf-observability — `flatMap` erases per-root provenance

`src/agent/daemon/scheduler.ts:511-512`, consumed at `:524-526` · ref:45b751ac

`removed`/`warnings` are flattened into untagged path arrays and reduced to bare counts in the summary, so a spike in warnings cannot be attributed to a repo without eyeballing path prefixes. **Suggestion:** keep the per-root `results` array in the telemetry record, or tag each entry with its source root.

---

### 🔵 low · security — every failure path is silent

`src/agent/worktree-root-registry.ts:57,65,89,113,129` · ref:45b751ac

Five catch blocks swallow errors with only a comment. The best-effort contract is right, but a *persistent* write failure (read-only state dir, disk full, bad permissions) degrades to the exact #761 leak with zero operator-visible signal. **Suggestion:** a rate-limited `console.warn` on the write-failure paths.

---

### 🔵 low · api-compat — `version` is written but never read

`src/agent/worktree-root-registry.ts:28,49-50` · ref:45b751ac

`REGISTRY_VERSION = 1` is persisted, but `parseRegistry` never branches on `parsed.version` (verified: zero reads at the ref) — a future v2 falls through the generic shape filter with no migration or reset path. Forward-compat gap, not a break today. **Suggestion:** branch on `parsed.version` and route unrecognized versions explicitly.

---

### ✅ Verified sound

- **Claim 5 (unchanged empty-set behaviour) holds.** `sweepRootSet` (`:145`) unconditionally includes a non-null, non-empty primary, so `roots.length === 0` occurs only when the primary is null/empty — the same condition as the old `repoRoot === null`, with the same skip path and telemetry message.
- **No dropped `SweepResult` field.** `SweepResult` (`worktree-sweep.ts:150-155`) has exactly `removed`, `warnings`, `dryRun`, `candidates`; the synthesized literal reproduces all four, and `result` is consumed only at `:524-526`.
- **Purely additive API.** All four new exports are new symbols; no existing signature changed.
- **No scope creep.** Every hunk is load-bearing for the stated intent.
- **Blast radius is bounded.** The sweep anchors on `join(repoRoot,'.afk-worktrees')` (`:479`) and skips anything outside it (`:556`), so a bogus registry entry cannot steer removals at arbitrary directories. Symlink following in `fs.stat` (`:111-112`) stays inside that bound and within the same-user trust boundary that already governs `$AFK_STATE_DIR` — not a finding.
- **Farms are not a regression.** `src/agent/worktree.ts:319` creates under `$AFK_HOME/farms/`, outside `.afk-worktrees/` — never swept before or after this PR.

### What was not checked

- Citations verified inline against branch HEAD `45b751ac`; all five absence claims re-verified at that ref and confirmed absent. Both Wave 1 agents reported diff-file offsets rather than source line numbers — every line number above was corrected against the ref, and each evidence snippet was confirmed verbatim.
- Stated intent: PR #771 title + body — spec-compliance **was** assessed against it.
- Tests were **not** executed; the author's reported gate results are taken at face value. CI is green on this head.
- The two `high` findings were shadow-verified by independent re-derivation. Not reviewed: the Telegram surface, `worktree-sweep.ts` internals beyond the anchor/skip guards, and Windows path behaviour.

griffinwork40:
## Review remediation — pushed `00043691`

All feedback from the two `/review` passes and the Codex inline comments is addressed, plus the merge conflict with `main`. Per-item status below; every finding is either fixed or has a stated reason.

### Blocking

| # | Finding | Status |
|---|---|---|
| 1 | `scheduler.ts` — one bad root starves every root after it | **fixed** — per-root `try/catch`, failure demoted to an `[ERROR]` warning, loop continues |
| 2 | `interactive/worktree.ts` — third managed-create path never registers its root | **fixed** — `registerWorktreeRoot()` after a successful `git worktree add`; the "single chokepoint" comment in `worktree-managed.ts` is corrected rather than left false |

### High / medium

| # | Finding | Status |
|---|---|---|
| 3 | Registry read-modify-write is unlocked and non-atomic | **fixed** — in-process promise chain → degradable advisory lock → temp-file + rename |
| 4 | Registry world-readable | **fixed** — `0o600` on both writes (set on the temp file *before* rename), `0o700` on mkdir |
| 5 | Soft-launch valve counted machine-globally, so a new root got a live destructive sweep on first contact | **fixed** — counter moved per-root into `.afk-worktrees/.sweep-runs` (new `worktree-sweep-valve.ts`) |
| 6 | Cap eviction is invisible | **fixed** — eviction emits a debug line naming the dropped roots |
| 7 | Cross-root totals unattributable | **fixed** — summary reports `swept/attempted root(s)` |
| 8 | Skip excerpt named one of two causes | **fixed** — an empty set means *both* hold at once, so the message now says so |
| 9 | `const results = []` lost its annotation; `some` vs `every` on `dryRun` | **fixed** — annotated; `some` kept deliberately with the ambiguity documented at the call site |
| 10 | Multi-root wiring untested | **fixed** — new `scheduler-worktree-prune.test.ts` (6) |
| 11 | No `MAX_ROOTS` / concurrency coverage | **fixed** — registry suite 14 → 20 |
| 12 | `.afk-worktrees` containment was a raw `startsWith` | **fixed** — uses the `isPathWithin` boundary test the file already defines |

Two design notes worth a reviewer's eye:

- **The advisory lock is deliberately degradable.** It yields after 2s and writes anyway. Registration runs inside worktree creation, so blocking a create is worse than a rare lost update; the in-process chain still orders every writer in the process that actually races.
- **`readRootSweepCount` returns `null`, never `0`, when a marker is unusable** (no `.afk-worktrees/`, read-only checkout), falling back to the legacy global count. Reporting `0` there would pin such a root in dry-run forever and leak its worktrees — the #761 bug wearing a different hat.

Every fix carries a test that fails without it. The three that reproduce the reported symptoms directly: parallel registration retained **1 of 20** roots before the fix; the scheduler's second root was **never swept**; a fresh root was swept **live instead of previewed**.

Also merged `main` (conflict was the adjacent-import collision in `worktree-managed.ts`; both imports kept) — the PR is `MERGEABLE` again.

Gates: `pnpm lint` clean · full suite **14,007 passed / 731 files** · `audit:env:check`, `scan:env:check`, `audit:sdk:check` all pass.

griffinwork40:
# Code Review — PR #771

`fix(worktree): sweep every repo root known to hold managed trees (#761)`
`afk/lane5-sweep-root-registry` → `main` · ref **`7ecef805`** · 11 files, +1115 / −38 · regime **full**
CI: all green (Windows job **skipped**) · mergeState CLEAN · reviewDecision none

**Verification posture:** 2 Wave-1 dimension agents → inline citation + absence-claim verification at the ref → 3 adversarial shadow verifiers on the top claims. All 3 verifiers returned **CONFIRMED** via `independent-rederivation`. One asserted `high` fell to `medium` on re-derivation. Nothing was dropped as fabricated.

---

## Blocking (2)

### H1 · high · confidence high · security + correctness · `src/agent/worktree-sweep.ts:888` · ref: 7ecef805 · citation-type: diff-context
**Finding:** The per-root soft-launch counter is credited on *every* non-bypassing sweep including explicit `dryRun: true`, so three agent- or CLI-initiated preview calls exhaust a root's entire 3-preview budget and the daemon's **first** tick against that root deletes for real without any daemon preview ever having run — silently defeating the valve's own documented invariant ("a root is PREVIEWED before it is swept destructively", `worktree-sweep-valve.ts:4-12`).
**Evidence:**
```
  if (options.bypassSoftLaunch !== true) await recordRootSweep(repoRoot);
```
Non-bypassing dry-run callers, all verified at the ref: `src/agent/tools/handlers/worktree.ts:275-280` (`dryRun: true` — **agent-invoked**, output goes to a model's context, not the operator), `src/cli/commands/worktree.ts:82-86` and `:171`, `src/cli/slash/commands/worktree.ts:216` and `:254`. Pre-PR this was structurally impossible: `countPriorSuccessfulRuns` (`worktree-sweep.ts:395-415`) counted only telemetry rows with `taskId === 'worktree-prune'`, which only the daemon writes.
**Suggestion:** `if (options.bypassSoftLaunch !== true && options.dryRun !== true) await recordRootSweep(repoRoot);`
**Shadow verifier:** CONFIRMED, severity stands at high. Blast radius bounded and re-verified inline — `classifyCandidate` still short-circuits on `locked` (`:338`), and both removable verdicts require `!isDirty && !holdsUnreplaceableCommits` (`:349-354`, `:366-372`); branch refs survive via `branchSafeToDelete` (`:812`). Not `critical`: no dirty, locked, or unpushed work is reachable.

### H2 · high · confidence high · correctness + security · `src/agent/worktree-root-registry.ts:244-249` (persisted at `:259-273`) · ref: 7ecef805 · citation-type: diff-context
**Finding:** The liveness gate treats every `fs.stat` rejection as "directory gone", so one transient `EACCES`/`EIO`/`ESTALE` (unmounted volume, an ancestor losing `+x`) permanently drops a live root and persists that drop — re-creating the exact #761 leak this module exists to close. Both Wave-1 agents converged on this independently, from the correctness and the security lens.
**Evidence:**
```
      const stat = await fs.stat(absolute);
      if (!stat.isDirectory()) continue;
    } catch {
      continue; // gone → prune
```
**Suggestion:** Prune only when `(err as NodeJS.ErrnoException).code === 'ENOENT' || code === 'ENOTDIR'`; retain and skip the entry on any other errno.
**Shadow verifier:** CONFIRMED, stands at high. Recovery is the decisive fact: the **only** path that re-adds a root is `registerWorktreeRoot()`, reachable solely from `worktree-managed.ts:138` and `interactive/worktree.ts:633` — the user must create a *brand-new managed worktree in that exact repo*. Opening a session, running `afk worktree prune`, or a daemon tick do not re-register. The drop is also silent (see M8).

---

## Non-blocking (10 medium)

**M1 · medium · correctness · `src/agent/worktree-sweep-valve.ts:58-66` + `worktree-sweep.ts:501-505`** — A marker that exists but cannot be written returns `null`, so `runSweep` falls back to the machine-global telemetry count (≥3 on any long-lived machine) and sweeps that root destructively with zero previews. *Asserted high; lowered to medium on re-derivation* — the common causes (unwritable `.afk-worktrees/`, read-only mount) block removal too, so the finding is self-limiting except in one concrete state the verifier named: a corrupt `.sweep-runs` owned by another UID (mode 0644, created once under `sudo`) inside a writable `.afk-worktrees/` — read succeeds, write is denied, removal is unaffected. **Suggestion:** return `0` when `.afk-worktrees/` exists but the write fails; `null` only on ENOENT of the directory.
> **CONFLICT (resolved, surfaced by rule):** the module's Contract docstring (`valve.ts:19-25`) pre-argues that returning 0 for an unusable marker "would pin such a root in dry-run FOREVER … precisely the #761 leak". The verifier judged the proposed fix a **refinement, not a contradiction**: the blanket argument holds only for the dir-missing case, and a root pinned in the refined state is still reclaimable via `bypassSoftLaunch` callers (`boot-prune.ts:112-117`) or a permission repair. Author's call — both rationales are on the record.

**M2 · medium · security · `worktree-sweep-valve.ts:38-41,50-54`** — The destructive/preview decision reads a plain integer from inside the repo working tree, so anything with write access under that repo (a sibling managed worktree, a cloned repo, another tool) pre-writes `3` and the first daemon sweep is destructive with zero previews. **Suggestion:** key the counter in `$AFK_STATE_DIR` by a hash of the resolved root, and clamp the read to `Math.min(parsed, SOFT_LAUNCH_RUNS)`.

**M3 · medium · security · `worktree-root-registry.ts:141-143` + `:118-128`** — The lock file is created and the pid written as two steps; a competing writer that reads the empty file parses `NaN`, takes the `catch`, and `unlink`s a **live** holder's lock — two writers in the critical section, and a lost root is a permanently leaked worktree. **Suggestion:** write the pid in the same `open` (or temp+rename the lock) and treat an empty body as *held* until `LOCK_TIMEOUT_MS` elapses.

**M4 · medium · correctness · `worktree-root-registry.ts:242,289-296`** — De-dup keys on `resolve()` only: no symlink (`/var` → `/private/var`) or case normalization, so one physical repo registered under two spellings is swept twice per tick and burns two previews per tick. Inconsistent with the sweep engine's own `realpathSafe` (`worktree-sweep.ts:249` — cited as `:222`, line corrected). **Suggestion:** key the `seen` sets on `realpathSafe`.

**M5 · medium · correctness · `worktree-root-registry.ts:201-210` + `:218-225`** — `lastSeenAt` is written but never compared anywhere (grep at ref: zero readers outside the module), so "oldest-seen dropped first" is implemented as *oldest-registered*, and a sweep never refreshes it. The docstring at `:193-199` documents this exact failure mode and mitigates with a `debugLog` — a deliberate tradeoff, but the code and the comment at `:35` disagree. **Suggestion:** sort by `Date.parse(lastSeenAt)` before slicing, and refresh on sweep.

**M6 · medium · observability · `scheduler.ts:513-532,560-565`** — A tick in which *every* root's `runSweep` rejects records `status: 'success'`, so `insights/aggregators/daemon.ts:143-148` (which tallies `errorCount` off `status === 'error'`) reports a permanently broken prune as healthy. The excerpt does say `warned N across 0/2 root(s)`, so it is not invisible — only unmachine-readable. **Suggestion:** emit `status: 'error'` when `results.length === 0 && roots.length > 0`.

**M7 · medium · observability · `scheduler.ts:527-530,552-558`** — Per-root failure strings are synthesized, redacted, then discarded into the unpersisted `warnings`; only the summary reaches `responseExcerpt`, so an operator cannot learn *which* root failed or why. The code comment at `:552-554` concedes it. **Suggestion:** add `warnings: result.warnings.slice(0, 20)` as a structured field on the record.

**M8 · medium · observability · `worktree-root-registry.ts:254-274`** — Prune-on-read deletes entries with no log line at any level (the module's only `debugLog` is cap eviction at `:204`), so when H2 fires the root vanishes with zero diagnostic trail. **Suggestion:** `debugLog` each pruned path plus its `stat` errno before `mutateRegistry`.

**M9 · medium · perf · `worktree-root-registry.ts:52,111-115,180-191`** — The in-process `writeQueue` serializes registrations *in front of* the 2s lock, and both call sites `await` inside the create path (`worktree-managed.ts:138`, `interactive/worktree.ts:633`), so N concurrent `isolation: "worktree"` forks contending with a foreign holder pay N × 2s — ~16s at N=8, against a documented "yields after 2s rather than blocking a create". **Suggestion:** compute the deadline once per queue drain, or hoist one acquisition around the drained queue.

**M10 · medium · perf · `scheduler.ts:515-532`** — The tick now acquires the single machine-global sweep lock once per root (up to 64) instead of once, multiplying the window in which a concurrent REPL boot-prune loses the race and short-circuits on `LockContestedError` (`worktree-sweep.ts:510-517`) — interactive cleanup silently stops happening more often as the registry grows. **Suggestion:** acquire once for the fan-out and pass an `alreadyLocked` flag into `runSweep`.

---

## Low / nit (7)

- **L1 · low · `registry.ts:137-152` + `:162,182-184`** — `acquireRegistryLock` runs *before* `writeRegistry`'s `mkdir`, so on a fresh install `open('wx')` fails ENOENT (not EEXIST) → returns `null` → that first write is unlocked. Fix: `mkdir` inside `mutateRegistry` before acquiring.
- **L2 · low · `registry.ts:118-129`** — `process.kill(pid, 0)` throws EPERM for a live foreign-uid holder, and the bare catch reads that as dead → unlinks a held lock. Fix: reclaim only on `ESRCH`.
- **L3 · low · `registry.ts:65-78,235-252`** — Entries are admitted on `typeof path === 'string'` + `isDirectory()` alone; no absolute-path, `/`-or-`$HOME`, or `.git` check, so a hand-edited file can point the sweep at `$HOME` (blast radius still confined to `<root>/.afk-worktrees` — containment is structural). Fix: validate at read time.
- **L4 · low · test-coverage** — Four gaps, each absence-verified at the ref: no non-ENOENT `stat` test for H2 (`registry.test.ts` — 0 hits for `chmod|EACCES|EPERM|EIO`); no unwritable-but-present marker test for M1 (`worktree-sweep.test.ts` — 0 hits; no `worktree-sweep-valve.test.ts` exists); no assertion that a `bypassSoftLaunch` caller leaves `.sweep-runs` unincremented (only `boot-prune.test.ts:43,53`, option-passing only); the cap test registers strictly sequentially (`registry.test.ts:163-176`, comment at `:166`) so it cannot detect M5.
- **L5 · low · `registry.ts:259-273`** — The `dead` set is computed outside the lock and applied by path, so a root re-registered mid-loop is dropped by the transform; the comment at `:255-258` protects a *different* concurrently-added root, not the same one. `[low confidence — verify with runtime context]`
- **N1 · nit · `scheduler-worktree-prune.test.ts:144-155`** — Test named `'still errors the tick if every root fails'` asserts `expect(status).toBe('success')`. Rename to match the assertion.
- **N2 · nit · `registry.ts:164`** — Temp file uses `${pid}-${Date.now()}` with non-exclusive `writeFile`, against this repo's own documented precedent at `permissions-store.ts:289-294` (which switched to `randomUUID()` precisely because `Date.now()` has ms granularity). Fix: `randomUUID()` + `open(tmp, 'wx', FILE_MODE)`.

---

## Spec-compliance

**Assessed** against PR #771 title + body. **No unmet-intent and no scope-creep finding.** Every load-bearing claim traces to the ref: exactly two registration sites, both `await`ed and both documented as such (`worktree-managed.ts:130-138`, `interactive/worktree.ts:626-633`); `$AFK_STATE_DIR/worktree-roots.json` (`paths.ts:357`); primary-first at `registry:289`; serial loop with per-root isolation at `scheduler:515-532`; `status: 'skipped'` preserved with both causes named at `scheduler:485-492`; `0o600` on the temp file pre-rename (`registry:44,168-172`); atomic temp+rename (`:160-177`); per-root valve (`valve:38-41`); no new env var. Two claims are *technically* satisfied but behaviourally undercut: "never fails a create" holds for *failure* but not for *latency* (M9), and "oldest-seen dropped first" is oldest-*registered* (M5).

## What was not checked

- **Citations verified inline against branch HEAD `7ecef805`** (present locally; the working tree is on `afk/subagent-stream-cut-retry` @ `89df5fd2` and was left untouched). Every `high` citation, every `file-state` citation, and every absence claim was re-read or re-grepped at that ref. **Citation-hygiene note:** several test-file citations carried diff-artifact line offsets rather than file line numbers (e.g. `registry.test.ts:495-518` → actually `:163-176`); content verified, lines corrected above — none fabricated.
- **Not re-read at the ref:** `cli/commands/daemon.ts:356`, `agent/worktree.ts:319`, `worktree-orphan-guard.ts`, and the `isDirectory` filter that makes `.sweep-runs` inert to the orphan scan. The two-registration-site claim was confirmed independently by grep, which is the load-bearing half.
- **No tests, build, lint, or audit gates were run** by this review. CI on the PR is green, but **`Test (windows-latest)` was skipped** — and M3, L1, L2, N2 all touch Windows-specific `fs.open('wx')` / `process.kill(pid, 0)` / `rename`-over-existing behaviour, plus the `0o600` assertion is `skipIf(win32)`.
- **Protocol deviation:** Wave-2 synthesis was performed inline by the orchestrator rather than dispatched to a sub-agent, to avoid re-laundering ref-verified evidence through a fourth summarizer. Wave 1 (2 agents) and the shadow-verify wave (3 agents) ran as specified. Echo-chamber guard: claims 1 and 2 necessarily share the valve-wiring region, but each verifier read outside it (verifier 1 → the `main` baseline plus the classifier guards; verifier 2 → the removal path plus boot-prune), so the guard's remedy is pre-satisfied and no extra round was dispatched.
- Not reviewed: Telegram surface, docs/website, dependency posture.

---

# DO NOT MERGE

Two `high` findings, both of which defeat an invariant this PR states for itself, both confirmed by independent re-derivation, and neither covered by a test: **H1** lets non-operator-visible dry-runs spend the destructive-sweep safety valve, and **H2** re-creates the #761 leak the PR exists to close on any transient filesystem error. Each fix is one line plus one errno check; each needs the test named in its suggestion. The 10 mediums are legitimate follow-ups — M1 and M2 (same valve region as H1) are worth folding into the same pass.

---

griffinwork40:
# Code Review — PR #771

**Target** `fix(worktree): sweep every repo root known to hold managed trees (#761)` · base `main` ← head `afk/lane5-sweep-root-registry` · ref **`aa89628e`** · +1828/−45 across 15 files (9 production, 6 test) · regime **full**
**Waves** 2 dimension agents → inline citation/absence verification → inline synthesis → 3 shadow verifiers on the top-severity claims.

---

## 🔴 Blocking

### `[F-S1]` high · high · security · `src/agent/worktree-sweep-valve.ts:73-95` · ref:aa89628e · citation-type:diff-context · **shadow-verify: CONFIRMED (independent-rederivation)**
**finding:** `readRootSweepCount` writes the per-root marker with a plain `fs.writeFile` (default `'w'` = `O_TRUNC`, no `O_NOFOLLOW`), so when `<repoRoot>/.afk-worktrees/.sweep-runs` is a symlink the unattended cron sweep truncates the **link target** — a file outside the repo — and overwrites it with `0`.
```
    const parsed = Number.parseInt((await fs.readFile(target, 'utf-8')).trim(), 10);
    if (Number.isInteger(parsed) && parsed >= 0) return parsed;
  } catch { /* absent or unreadable → fall through and try to initialise */ }
    await fs.writeFile(target, '0', 'utf-8');
```
**reachability:** `daemon.ts:337-342` (cron `0 4 * * *`) → `scheduler.ts:474,524-532` → `worktree-sweep.ts:503` → the write. No confirmation, no allowlist, and line 503 runs *before* `effectiveDryRun` (`:505`) and before the lock (`:510`). Grep of the module for `lstat|O_NOFOLLOW|realpath|symlink` → zero matches. Git carries mode-120000 symlinks, so a cloned repo that does not gitignore `.afk-worktrees/` ships one at that exact path; `.gitignore:55` protects this repo only. Upstream `worktree-root-registry-liveness.ts:74` validates only `stat().isDirectory()` and never inspects the marker — nothing neutralizes it.
**suggestion:** `fs.open(target, 'wx'/'r+')` with `O_NOFOLLOW`, or `lstat` and skip the root when `isSymbolicLink()`.

### `[F-A1]` high · high · api-compat · `src/agent/worktree-sweep.ts:501-505` + `src/cli/commands/worktree.ts:171-186` · ref:aa89628e · citation-type:diff-context · **shadow-verify: CONFIRMED**
**finding:** `afk worktree prune --apply` silently applies nothing for its first three invocations on every pre-existing repo, because the documented legacy fallback is unreachable for any repo that actually holds worktrees: `readRootSweepCount` returns `null` — the sole trigger for `?? countPriorSuccessfulRuns(...)` — only on `ENOENT`/`ENOTDIR`, i.e. only when `.afk-worktrees/` is absent, so a repo with managed trees always starts at `0` and `effectiveDryRun = false || (0 < 3)` is true.
```
  const priorRuns = options.bypassSoftLaunch
    ? Number.POSITIVE_INFINITY
    : (await readRootSweepCount(repoRoot))
      ?? await countPriorSuccessfulRuns(resolvedTelemetryPath);
```
The CLI passes `dryRun: !options.apply` with no `bypassSoftLaunch` (`worktree.ts:171,179`) and prints `🔍 Dry-run mode — no changes made.` + `Removed: 0` at exit 0 (`:184-186`); `/worktree prune --apply` is identical (`slash/commands/worktree.ts:254-257`). The daemon's own long-established primary root is likewise re-previewed for 3 ticks. `worktree-sweep.test.ts:115` seeds `writeRootMarker(repoRoot, 3)` in `beforeEach`, which is exactly why no test sees the upgrade state.
**suggestion:** Seed the marker from `countPriorSuccessfulRuns` on first read when no marker exists, or exempt an explicit `--apply` from the valve.

---

## 🟡 Medium

`[F-A2]` medium · high · api-compat · `worktree-sweep.ts:501-503` · diff-context — Documented read-only callers now write into the user's repo: `readRootSweepCount` initialises the marker unconditionally, so `dryRun: true` paths (`tools/handlers/worktree.ts:275`, `slash/commands/worktree.ts:219`, `cli/commands/worktree.ts:82`) create `.afk-worktrees/.sweep-runs` where they previously mutated nothing — and reach F-S1's truncation. **Fix:** split read from initialise; write only on the credit path in `recordRootSweep`. Shares a fix with F-S1.

`[F-C1]` medium · high · correctness · `scheduler.ts:539-564` · diff-context · **shadow-verify: CONFIRMED** — A tick mixing one live root (marker ≥3) with one freshly registered root (marker 0) folds to `dryRun: results.some(...)` → renders `would remove N` and never `result.removed.length`, hiding removals that hit disk. The adjacent comment — *"Mixed values are unreachable while the soft-launch valve resolves uniformly per tick"* (`:541-543`) — is **false at this ref**: the valve is per-root. Blast radius is reporting-only (deletions gate on per-root `effectiveDryRun`; `insights/aggregators/daemon.ts:150` drops `responseExcerpt`), but the wrong wording propagates verbatim into the operator's Telegram push (`cli/commands/daemon.ts:236-238`). **Fix:** render both halves unconditionally; delete the stale claim.

`[F-O1]` medium · high · perf-observability · `scheduler.ts:535,546,564` · diff-context — On a *partial* failure the failing root is unobtainable. `[ERROR] sweep failed for <abs path>` goes into `result.warnings`, whose only consumer is `.length` (`:564`); `TelemetryRecord` has no `warnings` field; `grep debugLog scheduler.ts` → zero matches; `errorMessage` fires only when **every** root fails (`:577-578`). This contradicts `root-failure-summary.ts:14-16`, which designates `warnings` "the detailed channel." **Fix:** `debugLog` each entry, or emit `summarizeRootFailures` on the success record too.

`[F-C2]` medium · high · correctness · `scheduler.ts:522-538` + `:53` · diff-context — Per-root isolation covers rejections, not hangs: `builtinPruneExecFile` is a bare `promisify(execFileCallback)` with no timeout (grep `timeout|AbortSignal|signal:` over the file → zero matches) and the loop is serial, so one blocking `git` (credential prompt, stale NFS handle) stalls the tick forever and strands every root behind it — the same starvation the adjacent `Invariant:` block claims to have closed. **Fix:** pass a per-root `timeout`/`AbortSignal.timeout` so a hang lands in the existing `catch`.

`[F-C3]` medium · medium · correctness · `worktree-root-registry.ts:308-309` + `scheduler.ts:69` · diff-context — Dedupe is `resolve()`-only, no realpath, while `resolveWorktreePruneRoot` returns `AFK_WORKTREE_SWEEP_ROOT` verbatim and otherwise `git rev-parse --show-toplevel` (already realpath'd). A repo reached through a symlinked parent or a differently-cased path (macOS) registers twice, is swept twice per tick, takes two `MAX_ROOTS` slots, and earns two valve credits. `worktree-root-registration.test.ts` realpaths its tmpdir specifically to dodge this. **Fix:** normalize through the existing realpath helper (`worktree-sweep.ts:262`) before the `seen` check.

`[F-S4]` medium · medium · security · `worktree-sweep.ts:531-535,575-577` · file-state (verified at ref) — The orphan scan `readdir`s `.afk-worktrees` and `fs.rm(recursive)`s its entries without resolving that directory through symlinks, so a symlinked `.afk-worktrees` redirects recursive deletion outside the repo. Pre-existing, but this PR multiplies the blast radius from one daemon-resolved root to every registered root. **Fix:** `lstat` the root and skip it when it is a symlink.

`[F-S2]` medium · low · security · `worktree-sweep-valve.ts:73-74` · diff-context · **⚠ `[was: high, now: refuted]`** — Original claim: an in-repo marker reading ≥3 makes first contact destructive, defeating the stated "three previews on first contact" property. **Verifier's corrected finding:** the mechanism is real (`parseInt` trusts any non-negative integer unbounded) but contradicts nothing in production — `recordRootSweep` is the only writer and only ever writes `current + 1` after a credited pass, so a marker ≥3 *records previews already given*; the stated property concerns inheriting a **different** repo's global count. Residual risk: hand-copied/restored/foreign-repo marker content is trusted without bound. **CONFLICT (not auto-resolved):** verifier 2 leaned on `.gitignore:55` to argue no marker arrives by checkout; verifier 1 established that the ignore file is per-repo and a foreign clone can ship arbitrary content at that path. Verifier 1's reasoning is the stronger of the two — surfaced for your call.

---

## 🟢 Low

- `[F-S3]` `worktree-root-registry.ts:70-75` — `parseRegistry` accepts any non-empty string as `path`; a relative entry in an externally-written file is later `resolve()`d against **each reading process's own cwd**, so the daemon (`$HOME` under launchd) and the REPL select different anchors. Grep `isAbsolute|__proto__|prototype` in the module → zero matches (prototype pollution itself is not reachable — `JSON.parse` defines `__proto__` as an own property). **Fix:** reject `!isAbsolute(e.path)` in the same filter.
- `[F-S5]` `:31` written at `:164`, never compared on read (`:66-79`) — a future/forged schema version is interpreted under v1 rules. **Fix:** `return []` when `parsed.version !== REGISTRY_VERSION`.
- `[F-C4]` `:221,236` — `lastSeenAt` is written and never read (grep: interface `:58`, doc `:226`, write `:236` only); eviction is positional (`entries.slice(-MAX_ROOTS)`), not by the timestamp the doc at `:206-211` and the intent both name. Coincides today only because registration always appends. **Fix:** sort by `lastSeenAt` in `capRoots`, or drop the field.
- `[F-C5]` `worktree-sweep-valve.ts:92` + `worktree-sweep.ts:871` — the `null` branch is justified as "this root has nothing to do," but a root whose `.afk-worktrees/` was deleted wholesale still carries live `orphaned-registration` entries, so `git worktree prune` runs live with zero previews. Confined to administrative registrations, not user files.
- `[F-O2]` `scheduler.ts:561` — `rootsLabel` counts a root that short-circuited on `LockContestedError` (`worktree-sweep.ts:512-514`) as swept; `3/3 root(s)` can mean zero inspected.
- `[F-O3]` `worktree-root-registry.ts:38` — up to 64 roots × (`worktree list` + per-tree `status`/`rev-list` + removals), serialized, one aggregate `durationMs`, no per-tick deadline: a slow root is indistinguishable from a slow tick.
- `[F-T1]` `root-failure-summary.ts:52-58` — no test exists (grep `root-failure-summary|summarizeRootFailures` across `*.test.ts` → zero matches); the `slice(0, MAX - 1) + '…'` truncation, the module's only off-by-one-prone arithmetic, is never exercised.
- `[F-T2]` `scheduler-worktree-prune.test.ts` — no dry-run or mixed-`dryRun` tick (grep `dryRun: true|would remove|mixed` → zero matches): the exact gap F-C1 lives in.
- `[F-T3]` `worktree-root-registry.test.ts:149-165` — dedupe normalization covered only for a trailing slash; grep `realpath|symlink|toUpperCase|case-insensitiv` → zero matches, so F-C3's collisions are untested.

## ⚪ Nit

- `[F-P2]` scope creep — `root-failure-summary.ts` (58 lines of basename redaction) and the all-roots-fail-only `status:'error'` gate (`scheduler.ts:577-578`) are defensible hardening the stated intent does not name. Name them in the body so they are judged against a requirement.
- `[F-P3]` `scheduler.ts` is 821 lines and `worktree-sweep.ts` 901 (both pre-existing, both extended here) against the 350-LOC house rule, while a 58-line formatter was extracted — inverted extraction boundary. No `check-file-size` gate exists in this repo. New `worktree-root-registry.ts` at 314 is compliant.
- `[F-T4]` `scheduler-worktree-prune.test.ts:39-40` mocks **both** `runSweep` and `sweepRootSet`, so `:95`'s ordering assertion checks mock-call shape, not that scheduler + a real registry yield primary-first. Primary-first/dedupe *is* covered against real fs in the registry unit tests — a seam gap, not an uncovered invariant.

---

## Stated-contract verification

|---|---|---|
| 1. Never fails a create — all ops best-effort | **HOLDS** | `registry.ts:196-199`; test `registry.test.ts:65` (unwritable state dir resolves) |
| 2. Self-healing, bounded (prune-on-read + persist, corrupt→empty, cap 64) | **HOLDS** (except eviction key, F-C4) | `:269-289` gated rewrite, `:76-77`, `:213-221`; tests `:103`, `:149`, `:226`, `:241` |
| 2b. Transient errno never drops a live root | **HOLDS** — verified directly | `liveness.ts:75-83` retains non-`ENOENT`/`ENOTDIR` in the file, excludes from the pass; `seen` marked pre-`stat` (`:69-70`) closes the dead+alive double-classification |
| 3. Explicit target always wins | **HOLDS** | `registry.ts:304` primary prepended unconditionally |
| 4. Sequential sweeps under one machine-global lock | **HOLDS** | `scheduler.ts:522` serial loop; `worktree-sweep.ts:484,510` |
| 5. Nothing-known → `skipped`, both causes named | **HOLDS** | `scheduler.ts:475-494`; test `scheduler-worktree-prune.test.ts` |
| Writes serialized + degradable 2 s lock | **HOLDS (in-process only)** | `registry.ts:112-116,138-153`; two processes past the yield both write — a lost update drops a root (a leak), never a spurious deletion |
| Atomic temp+rename, `0o600` set on the temp before rename | **HOLDS** | `registry.ts:161-178`; test asserts `mode & 0o777 === 0o600` |
| Per-root failure isolation | **PARTIAL** | rejections isolated `:533-537`; hangs are not — F-C2 |
| Valve per-root; new repo gets its previews | **HOLDS for the honest path** | `worktree-sweep.ts:501-505` — but every *pre-existing* root is also reset (F-A1) and the marker is forgeable (F-S2) |
| `.afk-worktrees` containment is boundary-correct | **HOLDS** | `isPathWithin` realpaths both sides (`worktree-sweep.ts:262-265`), applied at `:626`; a symlinked `.afk-worktrees` still widens the orphan scan (F-S4) |
| No new env var; `docs/env-registry.*` untouched | **HOLDS** | zero `process.env` in the four new modules; `AFK_STATE_DIR` pre-registered |
| "Exactly two managed-create paths register" | **HOLDS** | three production `git worktree add` sites: `worktree-managed.ts:128`→registers `:138`; `interactive/worktree.ts:621`→registers `:633`; `worktree.ts:319` (`createFarm`) does not — farms live under `$AFK_HOME/farms`, outside every sweep's reach, so no root registration would reclaim them. Pre-existing gap, neither widened nor addressed. |
| Test/lint/audit gate results in the body | **UNVERIFIED** | read-only review; suite not run |

## What was not checked

- **Citations verified inline against branch HEAD `aa89628e`** (== checked-out worktree, clean). All 21 blocking/high and `file-state` citations resolved: **0 fabricated, 0 diff-only.** All 11 absence claims re-grepped at the ref: **11 confirmed-absent, 0 false-absent.**
- **One citation defect corrected:** the correctness agent's test-file anchors (`worktree-root-registry.test.ts:800-1017`, `scheduler-worktree-prune.test.ts:1-259`) exceeded those files' lengths — they were **diff-artifact line offsets**. Every referenced test *exists* (content-matched), and all anchors above are the corrected in-file numbers.
- **Stated intent:** PR #771 title + body — spec-compliance assessed.
- **Deviation:** Wave 2 synthesis was performed inline rather than by a fourth sub-agent; I held the full verification manifest plus shadow-verify output, which a fresh agent would only have received as a lossy summary. Peak concurrency 3, total 6 sessions.
- Not run: `pnpm test`, `pnpm lint`, any audit gate — every verification claim in the PR body is unverified here. Not reviewed: Windows path/`process.kill` semantics, `redactInlineSecrets` internals, the Telegram surface end-to-end, farm-tree reclamation, git's own symlink resolution during `worktree remove`.

---

# ⛔ DO NOT MERGE

Two `high` findings, both small and local: **F-S1** (symlink-following marker write truncates a file outside the repo, on the unattended cron path *and* from nominally read-only `list`/dry-run commands) and **F-A1** (`--apply` silently no-ops three times on every pre-existing repo because the documented legacy fallback is unreachable). F-A2 shares F-S1's fix. The registry core is solid — atomicity, `0o600`, the transient-errno retention rule, and the realpath containment test all verified as designed; the defects cluster entirely in the new valve's placement of state inside the swept repo.

**Done**
- Reviewed PR #771 at ref `aa89628e` across 5 dimensions: 2 parallel Wave-1 dimension agents, inline citation + absence-claim verification against the ref, and 3 parallel shadow verifiers on the top-severity claims — 21 findings surfaced (2 high, 7 medium, 9 low, 3 nit), 1 high downgraded to medium on refutation, 1 citation-anchor defect corrected.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/21b95cac-bcad-4386-a408-e352600d986c/pr-771.diff` (2132 lines); reviewed ref `aa89628ed2c4203f5b87fd4578890f67d03989e4`; every finding carries a `file:line` anchor re-read at that ref via `git show`, and every absence claim a `git grep` pattern at that ref — all reproducible from this worktree.
- World state unchanged: this review is read-only by skill contract — no files written, no commit, no PR comment, no `gh` mutation. `git status` is clean at `aa89628e`.
- Pending/deferred: the report above is **not persisted to disk** (the review skill forbids file creation) — say the word and I'll write it to a path you name or post it as a PR review. Nothing was fixed; F-S1/F-A1 are findings for the author. `pnpm test`/`pnpm lint` were not run, so the PR body's gate results remain unverified.

---

griffinwork40 (commented):

# Code Review — PR #771

**Target** `fix(worktree): sweep every repo root known to hold managed trees (#761)` · base `main` ← head `afk/lane5-sweep-root-registry` · ref **`ddfd69e5`** · +2521/−59 across 22 files · regime **full**
**CI** all green (Windows skipped) · mergeable CLEAN
**Waves** 2 dimension agents → inline citation + absence-claim verification at the ref → inline synthesis → 3 adversarial shadow verifiers on the top-severity claims.

---

## 🟢 Blocking: none

The single `high` raised in Wave 1 was **REFUTED** on independent re-derivation. Details under *Dropped findings*.

**Merge decision: MERGE.** Everything below is a non-blocking follow-up.

---

## Findings

### `[M1]` medium · high confidence · perf-observability · `src/agent/daemon/scheduler.ts:559-571` · ref:ddfd69e5 · citation-type:diff-context · **shadow-verify: CONFIRMED**

**finding:** On a default (non-debug) daemon a *partial* per-root sweep failure reaches the operator only as a bare count — the failing root's identity exists nowhere in any persisted channel.

**evidence:**
```
debugLog(`[worktree-prune] sweep failed for ${repoRoot}: ${reason}`);
```

Three sinks close in sequence: `debugLog` is gated on `AFK_DEBUG=1`/`DEBUG=1` (`src/utils/debug.ts:13-15`); `TelemetryRecord` (`scheduler.ts:177-190`) has no `warnings` field, so the `[ERROR] sweep failed for <path>` lines in `result.warnings` are consumed only via `.length` (`scheduler.ts:590`); and `errorMessage: summarizeRootFailures(...)` is set only under `results.length === 0 && roots.length > 0` (`scheduler.ts:610-623`) — i.e. when *every* root fails. The summary clause is a count alone: `` if (failedRoots > 0) clauses.push(`${String(failedRoots)} failed`); `` (`worktree-prune-summary.ts:81`).

A root that is sticky-broken (`.git` deleted — it survives the registry's `isDirectory()` liveness gate) fails silently on every tick forever.

**suggestion:** Set `errorMessage: summarizeRootFailures(rootFailureDetails)` on the partial-failure `success` record too, not only on the all-fail `error` record — the basename-only rendering is already non-enumerating and length-capped.

---

### `[M2]` medium · high confidence · perf-observability · `src/agent/daemon/scheduler.ts:65,78-79,548-572` · ref:ddfd69e5 · citation-type:diff-context · **shadow-verify: CONFIRMED**

**finding:** `PRUNE_GIT_TIMEOUT_MS` bounds one git exec, not the tick, so worst-case prune wall-time is unbounded now that a tick fans out to every registered root.

**evidence:**
```
const PRUNE_GIT_TIMEOUT_MS = 120_000;
...
promisifiedPruneExecFile(file, args, { timeout: PRUNE_GIT_TIMEOUT_MS, ...options });
```

The loop is serial over up to `MAX_ROOTS`(=64)+1 roots (`worktree-root-registry.ts:39`), and per-root exec count is *proportional to candidate count*, not fixed — `status`, `rev-list` ×2, `branch -d`, `worktree remove` run per candidate (`worktree-sweep.ts:700,761,779,841,848`). Grep for `Promise.race|deadline|budget|AbortSignal` in `scheduler.ts`: zero matches. Pre-#771 this was bounded by one root; the fan-out multiplies the ceiling by 65.

**suggestion:** Add a tick-level wall-clock budget that stops visiting further roots and reports the remainder as skipped, rather than relying on the per-exec timeout alone.

---

### `[L1]` low · high confidence · perf-observability · `src/agent/daemon/scheduler.ts:250-256` · ref:ddfd69e5 · citation-type:file-state · **shadow-verify: CONFIRMED, downgraded from medium**

**finding:** The cron dispatch has no overlap guard, so a long tick can be re-entered by the next scheduled tick.

**evidence:**
```
void this.runOnce(task, 'cron').catch(() => undefined);
```

`isDequeuing` (`scheduler.ts:231,345-346,372`) guards only the pull-dequeue path. **Downgraded from medium:** `acquireLock` in `worktree-sweep.ts:435-473` is single-shot and non-blocking — an `EEXIST` throws `LockContestedError` immediately (no poll loop), which the caller catches at `:515-525`, returning `contested: true`. Tick A holds the machine-global lock near-continuously, so an overlapping tick B short-circuits per-root in microseconds instead of duplicating work. Consequence is a fast near-no-op, not double-sweeping.

**suggestion:** Track an in-flight flag per builtin task id so an overlapping tick reports `skipped` rather than burning a lock-contention pass.

---

### `[L2]` low · low confidence · correctness · `src/agent/worktree-root-registry.ts:302-313` · ref:ddfd69e5 · citation-type:diff-context

**finding:** The prune-on-read self-heal can drop a root that was re-registered *after* the liveness classification but *before* the write, because the `dead` set is matched by path alone against a freshly re-read `current`.

**evidence:**
```
return current.flatMap((entry) => {
  const absolute = resolve(entry.path);
  if (dead.has(absolute) || kept.has(absolute)) return [];
```

The classify pass runs outside the lock (the code comment at `:296-301` acknowledges this and correctly defends against *wholesale overwrite*), but a path confirmed ENOENT at classify time is dropped unconditionally even if a concurrent `registerWorktreeRoot` refreshed its `lastSeenAt` in the interim. Window is narrow — it needs a delete-then-recreate-then-register interleaving — and it is self-limiting, since the next create in that repo re-registers it.

**suggestion:** Gate the drop on `dead.has(absolute) && entry.lastSeenAt === <value observed at classify time>` rather than on path identity alone.

`[low confidence — verify with runtime context]`

---

### `[L3]` low · high confidence · test-coverage · `src/agent/worktree-root-registry.test.ts:218,278` · ref:ddfd69e5 · citation-type:diff-context · **shadow-verify: CONFIRMED**

**finding:** No test combines a contended lock with concurrent writers, so the one lost-update path the design explicitly accepts has no assertion in either direction.

**evidence:** The `durability and concurrency` block covers 20 parallel in-process registrations with **no** foreign lock held (`:218`), and a **single** writer against a foreign-held lock (`:278`) — never both:
```
it('still records the root when the lock is held by a live foreign process', async () => {
```

**suggestion:** Add a case with two `registerWorktreeRoot` calls for distinct roots racing a held lock, asserting the documented worst case (at most one loses, never both).

---

### `[N1]` nit · high confidence · api-compat · `src/agent/daemon/scheduler.ts:610-623` · ref:ddfd69e5 · citation-type:diff-context

**finding:** For `worktree-prune`, `errorMessage` changes from raw error text to a basenames-only capped summary, so tooling that parsed absolute repo paths out of that field stops finding them. Intentional and security-motivated; flagged for consumer awareness only. Production consumers (`cli/commands/daemon.ts:235`, `insights/aggregators/daemon.ts:151-154`) treat it as free text and are structurally unaffected.

---

## Verified clean (read, no issues)

- **Both claimed registration call sites are real and wired** — `createManagedWorktree` (`worktree-managed.ts:138`) and the `afk -w` launcher (`cli/commands/interactive/worktree.ts:633`). Grep confirms these are the only two production callers of `registerWorktreeRoot`.
- **All five stated safety properties implemented:** never-fails-a-create (`registry.ts:252-263`), self-healing + `MAX_ROOTS=64` + corrupt→`[]` (`:39,67-92,278-316`), primary-first (`:327-345`), serial sweeps (`scheduler.ts:548-572`), empty-set skip naming both causes (`scheduler.ts:501-520`).
- **All remediation claims implemented:** in-process write queue + degradable 2s lock + temp-file→rename at `0o600` (`registry.ts:122-129,151-166,174-191`); per-root failure isolation (`scheduler.ts:548-572`); per-root valve (`worktree-sweep-valve.ts`); path-boundary containment replacing `startsWith` (`worktree-sweep.ts:654` via `isPathWithin` at `:269-271`, which realpaths both operands).
- **Eviction policy matches its stated contract** — `capRoots` sorts by `lastSeenAt` rather than array position (`registry.ts:226-243`); a just-registered root cannot be immediately evicted.
- **De-duplication is canonical, not lexical** — `normalizeRootPath` realpaths, collapsing symlinked-ancestor and case variants (`worktree-root-path.ts:31-38`).
- **Unknown schema version reads as empty** rather than under v1 rules (`registry.ts:76`), with a test pinning it.
- **No shell/argument injection** — every git call is `execFile` array-argv with `repoRoot` as a discrete element.
- **No secret exposure in new logging** — `redactInlineSecrets` applied once at push time; new debug lines carry paths and errno only.
- **Named new-coverage items all exist and assert real state**, not non-throw tautologies: idempotent registration, cross-root accumulation, vanished-root prune + persist, corrupt-file tolerance, path-is-a-file, unwritable state dir, and all four `sweepRootSet` cases.

## Dropped findings

- **`high` · security · "arbitrary-directory deletion via an unvalidated registry entry"** — **REFUTED** by independent re-derivation. `parseRegistry` (`registry.ts:67-92`) does validate only absoluteness, but the inference does not hold: `runSweep`'s first action is `git -C <repoRoot> worktree list --porcelain` (`worktree-sweep.ts:529`) inside a `try` with no local `catch` (only `finally` at `:908`), so a non-git root rejects out of the function before any deletion code is reached. The orphan scan is structurally confined to `<repoRoot>/.afk-worktrees/*` (`:556-563,605`), and the removal path is gated by `isPathWithin(entry.path, afkWorktreesRoot)` (`:654`) over entries that come only from git's own porcelain output. For `/etc` or a victim home dir: zero deletion. The residual (a real git repo the user owns that already contains `.afk-worktrees/`) has the same blast radius as the pre-existing single-root design and presupposes write access to the `0o600` registry.
- **`medium` · spec-compliance · "unlisted hardening = scope creep"** — dropped. The PR body incorporates the per-item remediation tables in the PR comments by reference, and those tables name both the symlink orphan-scan guard (F-S4) and the `O_NOFOLLOW` valve write (F-S1).
- **`medium` · security · "unbounded concurrent registrations"** — dropped; the agent found no reachable unbounded-growth path and `capRoots` bounds the blast radius.

## What was not checked

- Citations verified inline against branch HEAD **`ddfd69e5`**. Both Wave 1 agents emitted diff-file offsets rather than source line numbers; every citation was re-anchored and each evidence snippet re-read verbatim at the ref before being carried into synthesis.
- Stated intent: PR #771 title + body (+ the remediation tables it references) — **spec-compliance assessed**.
- Tests were **not executed**; the PR's own gate table (14,008 passed / 731 files, `tsc --noEmit` clean) was taken as reported, not reproduced.
- Static analysis only: the cross-process lost-update window and the aggregate-tick-timeout ceiling are derived from control flow, not from an adversarial timing harness.
- Not reviewed: the Telegram surface, docs site, and the near-main content of unchanged consumer modules read from a checkout on a different branch (`insights/`, `cli/commands/daemon.ts`) — adequate for reachability, not for line-exact citation.

PR #772, opened 2026-07-30
fix(compose): enforce the per-node tool budget by wind-down, not by killing the node

griffinwork40:
## Summary

`max_tool_calls_per_node` was enforced by a chained `progressSink` in the compose executor that counted `tool_use_detail` chunks per `subagentId` and called `manager.kill()` past the budget. That was wrong twice over:

1. **The count was provider-dependent.** `anthropic-direct` emits `tool.use.start` **twice per tool block** (`translate.ts` on `content_block_start`, then `loop.ts` post-turn with summarized input — pinned by `loop.test.ts:853-919`, "4 total for 2 tools"), and `stream-consumer.ts` maps each 1:1 to a `tool_use_detail`. With no `toolUseId` dedup the counter ran at **2× on anthropic and 1× on openai-compatible** — the same budget number meant two different things, and bit at half its stated value on sonnet children.

2. **Killing mid-round destroys the deliverable.** A subagent's answer only exists once it stops calling tools (`lastStreamedContent` holds streamed text and resets per turn), so a killed node returns ~90 bytes of failure text no matter how much work it did — and `isError: hasFailures` then failed the *whole* compose call, discarding healthy siblings. Observed in a real session: two waves × three nodes, **~34k output tokens of review work returned to the orchestrator as a 291-byte error**, after which the orchestrator abandoned parallelism entirely.

The repo already had the right policy for this exact problem. `providers/shared/tool-loop-cap.ts` spends a tool-round budget and then runs one final **tools-stripped wind-down round** so the model answers from what it already gathered — its docstring is explicit that *"a silent stop with no final message is indistinguishable from a hang"*. Compose nodes already inherit it through `SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS` (`subagent.ts:888`), so the compose budget was a **second, redundant budget layered on a working one** — implemented in the worse way.

**This PR deletes the counting sink and routes the budget into each node's fork config as `maxToolUseIterations`.** A node that spends its budget now returns a real (if shallower) result with `stopReason: tool_use_loop_capped` instead of dying mid-round; siblings are unaffected; the compose call stops being an error.

### API change (additive, backward-compatible)

The unit changes from individual tool **calls** to tool-use **ROUNDS** — a round with N parallel calls costs 1 — which is the unit the rest of the system already speaks (`max_tool_use_iterations` on the `agent` tool).

| Key | Status |
|---|---|
| `max_tool_rounds_per_node` | **new preferred key** |
| `max_tool_calls_per_node` | deprecated alias → maps to the new key, warns |
| both supplied | prefers `max_tool_rounds_per_node`, warns |

Validation (positive integer, 1–1000) is unchanged and now names whichever key the caller used.

Also drops the executor's `progressSink` override so `SubagentManager` resolves the ambient sink at fork time (`this.progressSink ?? getCurrentSink()`) — a sink installed after the manager is constructed is now observed, which the captured-at-construction sink missed.

Net **−13 LOC** in `compose-executor.ts` (791 → 778).

### Not in this PR

Programmatic cancels still record `origin: 'user_signal'` in the witness trace because `handle.cancel()` aborts without passing an `AbortOrigin` (`handle.ts:657`, default at `abort-graph.ts:175`) — so a `node_timeout_ms` fire or a fail-fast cascade is still indistinguishable from a human interrupt. That is a separate observability fix and lands separately; this PR removes the *largest* producer of those mislabeled cancels.

## How was this tested?

- `pnpm lint` — clean
- `pnpm test` — **712 files / 13,522 passed**, 14 skipped
- `pnpm audit:sdk:check`, `pnpm audit:env:check`, `pnpm scan:env:check` — all OK
- Reworked `compose-executor.test.ts`: the old `describe('max_tool_calls_per_node enforcement')` block asserted sink installation and `manager.kill()` behavior that no longer exists. Replaced with wiring tests — budget reaches every node's fork config; omitted (not zeroed) when unset so the fork default applies; deprecated alias forwards + warns; both-keys precedence; no `progressSink` override; **and a regression guard that `manager.kill()` is never called for tool volume**.
- Added `max_tool_rounds_per_node` input validation coverage (rejects non-number/0/negative/fractional/NaN/±Infinity/1001; accepts 1 and 1000 with no deprecation warning).
- Added `dag-subagent.test.ts` coverage that `maxToolUseIterations` reaches the fork config for a capped node and is absent for an uncapped one.

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes


griffinwork40:
@codex fix comments

griffinwork40 (commented):

## Merge decision: **MERGE** (2 non-blocking notes)

Reviewed ref `a3738eef` · 5 files, +246/−346 · regime: full · dimensions: security, api-compat, correctness, spec-compliance, test-coverage, perf-observability.

The core claim of this PR holds up under tracing. The budget genuinely reaches the provider loop, and the wind-down policy genuinely fires. No blocking findings.

### Verified end-to-end (the load-bearing claims)

**The budget actually arrives.** Traced `max_tool_rounds_per_node` → `parseComposeInput` (`compose-executor.ts:320-341`) → fork config (`compose-executor.ts:713-716`) → `dag-subagent.ts:108-110` → `subagent.ts:888` (`options.config.maxToolUseIterations ?? SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS`) → `resolveMaxToolIterations` / `shouldWindDown` (`providers/shared/tool-loop-cap.ts:64-77`). The value is not dropped or overridden anywhere on that path. Claim 5 in the body (`subagent.ts:888`) is accurate, line number included.

**A capped node is NOT silent to the parent** — worth stating explicitly, because deleting the old error-relabel block looks like an observability regression and is not one. `dag-subagent.ts:172` routes the raw-text branch through `annotateIfIncomplete(text, result.stopReason)`, and `subagent/result.ts:71-80` prepends `[⚠ PARTIAL RESULT — the subagent hit its tool-use iteration cap before finishing…]` for `TOOL_USE_LOOP_CAPPED`. The parent-visible truncation signal that the relabel block used to provide is preserved one layer down, and in a better place.

**`SubagentExecutionError` is still used** — the import at `compose-executor.ts:25` remains live via the cast at `compose-executor.ts:466`. No unused-import breakage from the relabel deletion.

**`let manager: SubagentManager;`** losing its definite-assignment `!` is safe: assignment at `:601` precedes the `try` at `:640` and every use (`:722`, `:775`). The old TDZ guard was only needed because the deleted sink dereferenced `manager`; removing both together is coherent.

**security — no issues found.** Read `compose-executor.ts:560-719`, `dag-subagent.ts:95-110`, `schemas.ts:540-568`. `resolvedNodeApiKey` (`:671`, `:711`) keeps the OpenAI-routed-node credential isolation intact; the new warning strings (`:306-316`) interpolate only literal key names, never a credential; `isolatingProgressSink` logs nothing.

---

### Findings

**medium · medium confidence · api-compat · `src/agent/tools/schemas.ts:557-564` · ref:a3738eef · citation-type:diff-context**
A caller who keeps passing `max_tool_calls_per_node: N` gets strictly more tool work than before, because both the unit and the terminal behavior changed under an unchanged key: `N` calls became `N` rounds (a round with parallel calls costs 1, so rounds are strictly coarser), and budget-exhaustion changed from `manager.kill()` to a wind-down that lets the node finish. "Additive, backward-compatible" is accurate about the *parse path* but not about *effective semantics* — a caller who set the budget to bound worst-case cost or runtime no longer gets a hard stop at all.
evidence:
```
'DEPRECATED alias for `max_tool_rounds_per_node` — prefer that ' +
'key. Accepted unchanged for back-compat, but the unit is now ' +
'tool-use ROUNDS, not individual tool calls, and spending the ' +
'budget triggers a graceful wind-down rather than cancelling the '
```
Mitigations confirmed present, which is why this is not `high`: the deprecation warning is genuinely surfaced to the caller — `parseWarnings` flows into the `> [compose warnings]` block prepended to the tool result (`compose-executor.ts:757-761`), not swallowed — and the schema text names the unit change. In-repo production callers of the legacy key: **zero** (full-repo `git grep` at the ref, excluding tests, finds only the alias handler, the schema entry, and a CHANGELOG line), so blast radius is limited to model-generated calls and any out-of-repo skill text that still teaches the old key.
suggestion: State the "no longer cancels — a hard cost bound is gone" consequence in the deprecated key's schema description, since that is the one behavior change a cost-sensitive caller cannot recover by rescaling the number.

**low · high confidence · correctness · `src/agent/dag-subagent.ts:65-71` · ref:a3738eef · citation-type:diff-context**
The new field doc tells a programmatic `runSubagentDAG` caller that `0` opts into unbounded, which is correct at that layer and is exactly the footgun it sounds like: `0 ?? 50` evaluates to `0`, `resolveMaxToolIterations(0)` returns `0`, and `shouldWindDown(n, 0)` is permanently false — so passing `0` silently discards the anti-hang default of 50 rather than requesting it. The compose surface rejects `0` (`MIN_NODE_TOOL_ROUNDS = 1`), so the two layers disagree about whether `0` is legal input.
evidence:
```
   * answer instead of being cut off mid-round. Omit to inherit
   * `SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS`; `0` opts into unbounded.
   */
  maxToolUseIterations?: number;
```
suggestion: Append to the doc that `0` disables the anti-hang cap entirely and that the compose input boundary rejects it, so the asymmetry is intentional rather than a surprise.

**nit · high confidence · spec-compliance · `src/agent/tools/compose-executor.ts` · ref:a3738eef · citation-type:file-state**
The body's "Net −25 LOC in `compose-executor.ts` (791 → 766)" does not match the reviewed ref: the file is **778 lines** at `a3738eef`. The gap is consistent with the later commit on this branch (`fix(compose): isolate ambient progress sink errors`) adding lines after the body was written. Flagged only because the repo enforces a file-size ceiling, so a stale LOC figure in the PR body misinforms that gate.
suggestion: Refresh the LOC line to `791 → 778` (net −13).

---

### What was not checked

- Citations verified inline against branch HEAD `a3738eef`, with the tree materialized at that ref (the local checkout was on an unrelated branch, so working-tree reads would have been misleading).
- Stated intent: PR #772 title + body — **spec-compliance assessed** against it.
- Did not execute the suite; the body's `pnpm test` figures (712 files / 13,522 passed) are taken as reported, not reproduced. CI on this head is green across Lint & Build, ubuntu/macOS, PTY scrollback, docs, and publish-bundle smoke.
- Did not verify at runtime that both `anthropic-direct` and `openai-compatible` emit `tool_use_loop_capped` when the cap fires — the wind-down mechanism itself is owned and tested in `providers/shared/tool-loop-cap.ts`, out of this diff's scope.
- Did not audit out-of-repo skill/plugin text for stale `max_tool_calls_per_node` instructions carrying the old call-based meaning; the api-compat note above covers the in-repo surface only.
- Body claim 2 (anthropic emits `tool.use.start` twice per tool block, so the old counter ran at 2×) was not independently re-derived — it is the *motivation* for deleting code that is now gone, so it does not gate the merge.
- One `high` observability finding was raised during review and **dropped** in verification: it asserted `TOOL_USE_LOOP_CAPPED` appears nowhere, true of `compose-executor.ts` alone but refuted at `dag-subagent.ts:172` and `subagent/result.ts:71-80`. Recorded here so the absence is not re-litigated.
- Synthesis was performed by the orchestrator rather than a forked agent after repeated upstream `overloaded_error` responses; every finding above is cited against the materialized ref and was verified by direct read.

griffinwork40:
Follow-up to the automated review above (`a3738ee`, verdict MERGE + 2 non-blocking notes). All three items are addressed in `60893ef`; both source edits are documentation-string only, no logic touched.

**1. medium · `src/agent/tools/schemas.ts:557-567` — deprecated alias hid a lost cost bound.**
The description named the unit change (calls → rounds) but not the consequence. Appended:

> Because exhaustion no longer hard-stops the node, this key is no longer a cost or runtime ceiling — it now only marks where wind-down begins.

That is the one behavior change a cost-sensitive caller cannot recover by rescaling the number, which is exactly why the reviewer wanted it stated at the key.

**2. low · `src/agent/dag-subagent.ts:68-73` — `0` footgun under-documented.**
The JSDoc said `0` opts into unbounded without warning that it *discards* the anti-hang default rather than requesting it. Verified all three mechanics before writing: the coalesce is `?? SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS` and fires only on `undefined` (`subagent.ts:888`); `shouldWindDown` is `maxIterations > 0 && …`, so a `0` cap is permanently false (`tool-loop-cap.ts:75`); and the compose boundary throws on `< 1` via `MIN_NODE_TOOL_ROUNDS` (`compose-executor.ts:178,331-332`). The doc now records that the two layers disagree deliberately.

**3. nit · stale LOC figure.** The body claimed `791 → 766` (−25); actual at the ref is **778** (−13), the gap being the later `isolate ambient progress sink errors` commit. Body corrected — flagged because the 350-line ceiling gate reads that number.

**Gates:** `tsc --noEmit` exit 0; full suite **712 files / 13522 tests passed**, 14 skipped, 0 failures. Fix built in an isolated worktree off `a3738ee`, DCO signed, pushed fast-forward (no force).

Not changed: the runtime deprecation *warning* string at `compose-executor.ts:312-316` is separate from the schema description and already names the wind-down; and `schemas.ts` remains over the 350-LOC ceiling, which is pre-existing and out of scope here.

PR #774, opened 2026-07-30
docs: impact map for ADR 0002 item 2, and mark it rejected

griffinwork40:
## Summary

ADR 0002 recorded decision item 2 — delete the `src/skills/_agents/` TS-wrapper layer and construct the builtins from the byte-pinned markdown frontmatter — as *"NOT implemented … wants an `/impact-map` pass first."*

**This is that pass, and it recommends against the change.** Two docs commits: the audit, and the resulting status flip on the ADR.

Blast radius is **HIGH (27.5)**, not the "moderate" the ADR assumed: 4 direct import sites, 11 transitive files, 7 test files, 2 build scripts, plus **8 bundled `SKILL.md` files that dispatch `research-agent` by bare name** and are invisible to type checking.

## Four corrections to the ADR, all code-verified

**C1 — It does not dissolve F4.** `KNOWN_AFK_TOOL_NAMES` goes 3 → 3, because `vendoredToolAllowlist` stays live for `audit-fit` (`audit-fit/index.ts:27,368`). The F4 win comes from migrating `audit-fit`, which is separable from deleting the wrappers.

**C2 — It reverses a deliberate documented decision.** `builtins.ts:27-37` states the opposite policy explicitly: *"Tools/description stay sourced from the wrapper constants … **NOT** from the file's frontmatter."*

**C3 — Neither build script appears in the ADR, and one carries the only genuine new risk.** `tryResolveReadFileSyncPath` requires every join argument to be a **string literal** (`esbuild-plugin-inline-prompts.mjs:122-141`). The natural shape of the rewrite is a parameterized helper:

```ts
// Looks obviously correct. Inlines nothing.
const loadVendored = (name: string) =>
  readFileSync(join(__dirname, '../../skills/_agents/prompts', `${name}.md`), 'utf8');
```

That yields no match and no inlining. **Since #816 this fails `build:dist` loudly** — the resolver throws on any prompt-shaped read it cannot resolve, naming the file and line (`:234-251`). At the time of the pass it was silent: the call was left as-is, the build succeeded, and the published artifact read a file that was never shipped — that gap is #776, filed off this audit and fixed by #816 (`548029c2`, 2026-07-31).

**The fix hardens C3 rather than weakening it.** The string-literal requirement is unchanged (`:130-133`); only the failure mode moved, from silent-and-shipped to loud-at-build-time. Keeping the bundler working still requires one hardcoded literal read per agent — which is structurally what the wrapper files already are — and that requirement is now *mechanically enforced* rather than conventional. So the refactor **relocates** the duplication rather than removing it, and converts a typed dependency into an untyped coupling to a build-script regex that only `build:dist` can see.

**C4 — It reverses `a34216f` (#625), two weeks old.** That commit deleted `model:` frontmatter *precisely because* `builtins.ts` does not read frontmatter. Item 2 inverts that premise and would require re-adding it — now as a coordinated two-repo change plus a re-pin.

Three reversals total (C1's `// Invariant:`, C2, C4), all moving deliberately **away** from frontmatter-sourcing. The ADR acknowledges none of them.

## Upstream verification

The byte-pin's upstream was checked directly, closing a gap the investigation had flagged as unchecked:

| check | result |
|---|---|
| `research-agent.md` vendored vs upstream | **IDENTICAL** `80993a5f…` |
| `git-investigator.md` vendored vs upstream | **IDENTICAL** `d1f186b8…` |
| `contract.md` upstream counterpart | **none** — that pin is self-referential |
| upstream frontmatter keys | exactly `name`, `description`, `tools` |
| upstream ships `model:` / `bash:` | **no** — neither key, in either agent |

So **ADR falsifier 1 is not triggered** — the pin is a live sync guard against real divergence, not a self-referential tax — and the ADR's `contract.md` parenthetical is confirmed.

## Also corrected: `audit-fit` was the wrong reason to defer

The ADR named `audit-fit` as the blast radius. Telemetry says one lifetime invocation (`audit-fit-telemetry.jsonl`, a single record dated 2026-05-03, `total_artifacts: 1, misfits_count: 0`). It was the weakest item in the entire map.

Separately verified as a **non-issue**: sourcing `tools` from frontmatter would not widen `audit-fit`'s gate. `Agent(git-investigator)` carries its parenthetical scope, cannot exact-match the bare `agent` entry, and no `LEGACY_TOOL_ALIASES` key contains `agent` — so it is dropped fail-closed and both sources normalize to the identical `{read_file, grep, glob, web_scrape}`.

## Recommended instead — item 2′

Migrate `audit-fit` off `vendoredToolAllowlist` onto `resolveAgentToolAccess`, then delete `to-definition.ts`, taking the confirmed-dead `toAgentDefinition` fossil with it. `KNOWN_AFK_TOOL_NAMES` 3 → 2, **zero** change to `builtins.ts`, zero bundler risk, zero reversals. Recorded in the ADR as item 2′, not implemented here.

Item 1 (collapsing the two read-only agent types) is untouched and remains the open maintainer decision.

## Update 2026-08-01 — `596ef410`

Two staleness classes appeared after this PR was written; both are now corrected in the diff.

1. **C3's silence framing** (above) — #776, filed off this very pass, was fixed by #816 on 2026-07-31, so a parameterized prompt read now fails the build instead of shipping past it. The verdict is unchanged and the structural argument strengthens; the prose is corrected in the audit's new dated addendum, in C3, in the verdict, in the recommendation, and in constraint 1, plus the matching bullet in the ADR itself. Every `esbuild-plugin` line citation was re-verified against current `main` and the drifted ones fixed (Pattern A `:172-224`→`:184-260`, the shape `:112`→`:207` scan / `:123-124` resolve, `tryResolveReadFileSyncPath` `:110-129`→`:122-141`, `parts.every` `:120`→`:133`, plus the new fail-closed branch at `:234-251`).
2. **Two ADR citations were line-anchored to text this PR itself edits** — the automated review's open finding. `(line 271)` quoted a sentence the rejection commit deletes outright, so it is now attributed to the reviewed ref `a25984a`; `(line 293)` became a quote anchor.

Note for reviewers: the automated review comment of 2026-07-31 08:18Z certifies C3 as *"genuinely invisible to lint/test/build"*. That review predates #816 by ~16 hours, so its C3 verdict is itself stale — C1, C2, C4 and the rest of its verification still hold.

## How was this tested?

Not applicable — **read-only analysis**. No source, test, or config file is modified; the diff is one new doc plus a status revision. Method was four parallel read-only sub-agent lanes (static importers / transitive reach / test+build+config / equivalence feasibility), with every load-bearing claim re-verified directly before it was written down.

CI runs its full matrix here despite the docs-only diff — `ci.yml` has no `paths` filter.

## Checklist

- [x] `pnpm lint` passes — unaffected, no source changed
- [x] `pnpm test` passes — unaffected, no source changed
- [x] No secrets, tokens, or private paths in the diff — scanned for `/Users/`, operator name, and API-key patterns; zero hits


griffinwork40 (commented):

## Merge decision: **MERGE**

No blocking finding survived verification. This is a docs-only diff (`docs/**/*.md`, +556/−10) whose correctness surface is **claim accuracy** — it flips ADR 0002 item 2 to *rejected* on the strength of specific code claims. I re-read every load-bearing claim against the reviewed ref rather than trusting the diff, and **C1–C4 all hold verbatim**. The residual findings are documentation hygiene, not defects in the analysis.

**Reviewed ref:** `c40a7e8a1e77bdbebffbb5d60d774bc186efc1a4` · **Stated intent:** PR #774 title + body — spec-compliance assessed.

---

## Claim verification (checked at the ref, not from the diff)

| Claim | Verdict | Evidence at ref |
|---|---|---|
| **C1** `vendoredToolAllowlist` live for `audit-fit` | **CONFIRMED** | `src/skills/audit-fit/index.ts:27` import; `:368` `const inspectorTools = vendoredToolAllowlist(researchAgent.allowedTools);` |
| **C1** `KNOWN_AFK_TOOL_NAMES` 3→3 | **CONFIRMED** | Two `ReadonlySet` copies (`_agents/to-definition.ts:14`, `agents/resolve.ts:43`) + the `// Invariant:` mirror comment at `resolve.ts:39`. F4 separability sound: `audit-fit` is the only non-test consumer. |
| **C2** `builtins.ts` states the opposite policy | **CONFIRMED verbatim** | `src/agent/agents/builtins.ts:31-33`: `Tools/description stay sourced from the` / `wrapper constants (the vendored contract diagnose already relies on),` / `NOT from the file's frontmatter.` |
| **C3** bundler requires string-literal join args | **CONFIRMED** | `scripts/esbuild-plugin-inline-prompts.mjs:118-121`: `const strMatch = trimmed.match(/^['"](.+)['"]$/);` → `return strMatch ? strMatch[1] : null;` → `if (parts.every(p => p !== null))`. A template literal yields `null`, the transform is skipped silently — no throw, no warning. |
| **C4** reverses `a34216f` (#625) | **CONFIRMED** | `a34216f` (2026-07-15) `chore(agents): remove inert model: frontmatter…`, diff removes two `-model: sonnet` lines. |
| `toAgentDefinition` is a dead fossil | **CONFIRMED** | Repo-wide at ref: only the definition (`to-definition.ts:58`) + barrel re-export (`_agents/index.ts:4`). Zero call sites — the item 2′ deletion is safe. |
| 4 direct import sites / 2 files | **CONFIRMED** | `builtins.ts:20,21`; `audit-fit/index.ts:26,27`. |
| 8 bundled `SKILL.md` dispatch `research-agent` by bare name | **CONFIRMED** | `devils-advocate`, `diagnose`, `refactor`, `research`, `review`, `shadow-verify`, `ship`, `simplify`. |
| Hash pins match the quoted prefixes | **CONFIRMED** | `_agents/vendored.test.ts:12-16` pins `80993a5f28dbcc9f…` / `d1f186b82574641a…`. |

**C3 is the PR's most consequential claim and it is the most solid one.** The failure mode it describes — a parameterized helper that looks correct, inlines nothing, and ships a bundle reading a file that was never published — follows directly from the regex gate at `:118-121`, and is genuinely invisible to `lint`/`test`/`build`. That claim alone justifies the rejection.

---

## Findings

### 1 · `medium` · correctness — the audit's line-anchored ADR citations are stale the moment this merges
`docs/audits/impact-map-agents-wrapper-layer.md:123` · citation-type: diff-context

The audit cites the ADR **by line number** for text that this same PR deletes:

```
The ADR justifies item 2 as *"Dissolves F4 instead of violating it"* (line 271).
```

At the reviewed ref, `Dissolves F4` **no longer exists anywhere** in `0002-readonly-agent-type-consolidation.md` (this PR's own strikethrough removed it), and line 271 now reads `   the types, not after.` — item 1's text. Same class at `:71` ("line 293" for the byte-pin, which actually sits at `adr0002.md:334`) and `:4` (cites `:268-273`, which at the ref spans item 1's status).

**Suggestion:** replace the three numeric anchors with quoted-string references (e.g. *"the ADR's pre-revision item 2 wording"*) so they survive the rewrite they describe.

### 2 · `low` · security — a private-repo coordinate lands in permanent public history
`docs/audits/impact-map-agents-wrapper-layer.md:252-256` · citation-type: diff-context

```
**Topology.** The marketplace is `griffinwork40/awa-private` (PRIVATE), cached
locally at `~/.afk/plugins/cache/agent-framework-private/` — the directory name
is a local alias, the git remote is `awa-private`. It ships two plugins:
`awa-private` (`agents/research-agent.md`, `agents/git-investigator.md`) and
`awa-dev` (`agents/qualify.md`). Cache HEAD `ce27e45`.
```

**This was raised as `high` and I downgraded it to `low` on evidence.** An adversarial re-derivation, which I then re-checked at the ref, found the substance is *already public in this repo*: private-plugin distribution (`LICENSING.md:28`), `forge`/`qualify` as reserved Pro features (`README.md:60`, `LICENSING.md:53`), the `awa-dev` plugin name (`src/agent/tools/skill-bridge.test.ts:114`), the bare `awa-private` token (`src/cli/terminal-compositor.first-turn-echo-image.test.ts:99`), and the `plugins/cache/<marketplace>/` layout (`docs/specs/imported-plugin-enabled-state.md:157`). Genuinely new are only the org-qualified form, the explicit `(PRIVATE)` label, and cache HEAD `ce27e45`. A repo name is not a credential and GitHub 404s unauthorized users, so this is not secret exposure.

It stays on the list for one reason: **the cost is asymmetric.** It is a two-line edit now and permanent public history after merge. Reinforced at `:234` and `:481`.

**Suggestion:** drop the org qualifier, the `(PRIVATE)` label, and `ce27e45` — "the private upstream marketplace the vendored prompts are byte-shared with" carries the full argument.

**Also verified:** the PR's checklist claim is **true as scoped** — zero hits for `/Users/`, `sk-`, `ghp_`, `xoxb`, `AKIA`, `Bearer `, `api_key`, or email addresses across both docs. The scan was incomplete for this leak class, not false.

### 3 · `low` · correctness — the "7 test files" input to the blast-radius score is unsupported
`docs/audits/impact-map-agents-wrapper-layer.md:104` · citation-type: diff-context

At the ref, only **3** test files reference `_agents/` (`_agents/vendored.test.ts`, `skills/audit-fit.test.ts`, `bundled-plugins/awa-bundled/bundled.test.ts`); the doc's criterion for "test files" is never stated. **The verdict does not move** — at 0.5 weight, 7→3 takes the total 27.5 → 25.5, still HIGH (15–39) — which is worth recording, since it means the tier survives the correction.

**Suggestion:** state the counting criterion, or cite the 3 files and note the tier is insensitive to the difference.

### 4 · `low` · correctness — two different values for the headline metric
`docs/audits/impact-map-agents-wrapper-layer.md:34` vs `:98,105` · citation-type: diff-context

`:34` says `The tier stays HIGH (recomputed ≈20, still 15–39)` while the section heading (`:98`) and arithmetic block (`:105`) both state **27.5**. One document, one metric, two numbers.

**Suggestion:** state the post-addendum score once and label 27.5 as the pre-addendum figure.

### 5 · `nit` — `nesting.ts:90` is cited without its directory and is a different construct
`docs/audits/impact-map-agents-wrapper-layer.md:134`. Real path is `src/agent/tools/nesting.ts:90`, and that line is `export const CHILD_ALLOWED_TOOLS = [...]` — an **array**, where the other three copies are `ReadonlySet`. "Structurally identical" overstates it. **Suggestion:** cite the full path and name the construct.

### 6 · `nit` — `maxToolUseIterations` provenance overstates the literal count by one
`docs/audits/impact-map-agents-wrapper-layer.md:86` cites `:71,124,141`, but `builtins.ts:71` is the const declaration `const READONLY_AGENT_MAX_TOOL_USE_ITERATIONS = 50;`, not a per-agent literal. **Suggestion:** split into "const at `:71`, applied at `:124,141`".

### 7 · `nit` · spec-compliance — two unrelated hygiene findings ride along
`docs/audits/impact-map-agents-wrapper-layer.md:343` embeds *"No CI job runs `fix:pins:check`"* and a coverage-floor note inside a doc scoped to the item-2 blast radius. The doc labels them independent, so they land with no owner. **Suggestion:** move to a tracked issue, leave a one-line pointer.

---

## Spec-compliance

**Unmet intent: none.** Both deliverables named in the intent are present — the audit doc (new, 505 lines) and the ADR status flip (`adr0002.md:6` `**Decision item 2 is now Rejected**`; `:276` `**Status: REJECTED (2026-07-29).**`). Item 1 is correctly left open. ADR item 2′ agrees with audit Stage 2′ on all four assertions (migrate `audit-fit`, delete `to-definition.ts`, 3→2, zero builtins/bundler/reversal cost). Every relative link I resolved exists at the ref.

**Scope creep:** only finding 7, and it is minor.

## test-coverage

`no issues found` — no source changed, so no test is owed. Existing tests do pin what the doc reasons about, and the compatibility claim checks out: `vendored.test.ts:34-52` reads the `.md` files directly with no wrapper dependency, so item 2′ leaves it green. `audit-fit.test.ts` imports `vendoredToolAllowlist` and would need rewriting under 2′ — the audit already flags this.

## perf-observability

`no issues found` — docs-only diff; no runtime, allocation, or logging path is touched.

## What was not checked

- **Citations verified inline against ref `c40a7e8a1e77bdbebffbb5d60d774bc186efc1a4`.** All `file-state` citations from the review agents were re-read at that ref; several had drifted (the agents read a working tree on a different branch) and the line numbers above are the corrected, ref-true ones.
- **Upstream byte-hash comparison not reproduced.** The vendored-side pins match the quoted prefixes (`vendored.test.ts:12-16`), but the IDENTICAL verdicts against `griffinwork40/awa-private` cache HEAD `ce27e45` were not re-run — that requires the private cache. The upstream table's conclusion ("ADR falsifier 1 is not triggered") rests on that unreproduced step.
- The 11-transitive-file chain was not spot-checked file by file.
- No build was executed; C3's "invisible until published" conclusion is verified by reading the bundler gate, not by running `build:dist`.
- Tests were not run (CI is green on this ref: all 7 required checks pass).

PR #773, opened 2026-07-30
fix(tui): bound grouped tool summaries to terminal width

griffinwork40:
## Summary

- keep grouped root tool summaries within the current terminal width
- reserve space for durable result summaries such as `59 lines total` before truncating repeated command previews
- add a regression test modeled on the reported `bash ×3` clipping case

## Root cause

`renderGroupedRootTools()` emitted `formatGroupedToolResults()` directly. That formatter joined every grouped command and appended the result summary into one unbounded logical line, so three long Bash commands produced a 350-column row in an 88-column terminal.

## Verification

- `pnpm test src/cli/commands/interactive/tool-lane.test.ts -t "grouped root bash reserves"`
- `pnpm test src/cli/commands/interactive/tool-lane.test.ts` (112/112)
- `pnpm lint`
- completion reconciliation: BACKED

PR #786, opened 2026-07-30
fix(tui): clamp agent frame head rows to terminal width

griffinwork40:
## Summary

The agent-frame head rows were the last unbounded rows in the flush (scrollback) path. This clamps them, and moves `renderGroupedRootTools` into the grouped-root module to keep `tool-lane-render-agent.ts` under the 350-line ceiling.

Follow-up to #773, which fixed the grouped root row. Found while reviewing that PR.

## Measurements

At `cols=88`, in the production dispatch shape — `addStartWithAgentContext` with no `maxWidth`, as `stream-renderer-orchestrator.ts` calls it for root tools:

| Row | before | after |
|---|---|---|
| `formatAgentSummary` head, depth 0, with stats | 109 | 88 |
| `formatAgentHeader` eager row (`flushSource`) | 181 | 88 |
| `Agent` label refreshed by the heartbeat path | 105 | 88 |

Nesting labels are already clipped to `NESTING_LABEL_MAX = 60`, so the overflow comes from everything around them: the spine indent, the tool name, the `[subagent]` dispatch tag, and — for the summary head — the ` — N tool calls · M lines` stats tail. Stats only render above `MAX_VISIBLE_CHILDREN = 3`, which makes the common case an ordinary root `agent` dispatch that made four or more tool calls. Not a corner case.

## Why both head rows are clamped

`formatAgentSummary` and `formatAgentHeader` emit a head row for the **same entry** at different times — eagerly via `flushSource`, or at dispose-time `flush()` — under an encoding constraint documented on `formatAgentHeader`: if the two encodings diverge, rows committed by one path sit at a different column than descendants rendered by the other, and the user sees a topology break. Clamping only the row that overflowed would itself be a divergence, so both are clamped identically.

`getTerminalWidth()` is now read **once** per summary frame and shared by the head row and the recursive child frame — which is what the comment there already claimed. Two reads could straddle a resize and emit a head row clamped to the old width above children clamped to the new one.

Plain clamp rather than the grouped path's suffix reservation: identity leads the row and the stats tail is expendable, matching how the live overlay already clamps the same row.

## Commits

1. `refactor(tui): move renderGroupedRootTools into the grouped-root module` — pure move, no behavior change. #773 extracted `formatGroupedToolResults` but left its only caller behind; this reunites them and takes `tool-lane-render-agent.ts` from 338 → 264 lines, making room for the clamp. Re-exports move with the symbols, so `tool-lane.ts`'s import surface is unchanged.
2. `fix(tui): clamp agent frame head rows to terminal width` — the behavior change plus its regression test.

Reviewing commit-by-commit separates the move from the behavior change.

## Verification

- `pnpm test src/cli/commands/interactive/tool-lane.test.ts` → 115/115
- Full `src/cli` tree → 4973/4973 (up 1: the new test)
- `pnpm lint` (tsc --noEmit, strict) → clean
- New test fails on the parent commit: `expected 109 to be less than or equal to 88`
- File sizes after: `tool-lane-render-agent.ts` 264, `tool-lane-render-grouped-root.ts` 174 — both under the 350 ceiling
- Re-verified after rebasing onto `main` post-#773 merge (main had also moved to v5.82.9 + #769)

## History note

This PR was originally opened against `afk/fix-grouped-tool-width` (#773) because that file was over the line ceiling on `main` at the time. #773 was squash-merged while this work was in flight, so GitHub retargeted this PR to `main`; the two commits here have since been rebased onto `main` so #773's squashed content is no longer duplicated.

## Not in scope

The nested/child rows (`renderFlushChildren`) and the live overlay were already clamped — verified, not changed.

### DIFF
--- a/src/cli/commands/interactive/tool-lane-render-agent.ts
+++ b/src/cli/commands/interactive/tool-lane-render-agent.ts
@@ -1,19 +1,10 @@
 import { palette } from '../../palette.js';
 import { getTerminalWidth } from '../../terminal-size.js';
 import { formatToolCallStat } from '../../format-utils.js';
-import {
-  MAX_VISIBLE_CHILDREN,
-  formatOutcome,
-  formatDiffBlock,
-  doneGlyph,
-  sanitizeLabel,
-  shortenPaths,
-  batchBadge,
-} from './tool-lane-format.js';
+import { MAX_VISIBLE_CHILDREN, batchBadge } from './tool-lane-format.js';
 import type { ToolEntry, Entry } from './tool-lane-render.js';
-import { getGlyphs, clampLineToTerminal } from './tool-lane-render.js';
+import { getGlyphs } from './tool-lane-render.js';
 import { renderFlushChildren } from './tool-lane-render-children.js';
-import { formatGroupedToolResults } from './tool-lane-render-grouped-root.js';
 
 /**
  * Compose a NESTING root's committed-scrollback closer — the `Done (…)` line —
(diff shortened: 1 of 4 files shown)


griffinwork40:
## Review — PR #786 `fix(tui): clamp agent frame head rows to terminal width`

**Target:** griffinwork40/agent-afk#786 · `afk/tui-agent-header-width` → `main` · reviewed ref **`14c560e`** (= clean local HEAD, so all file-state citations are reads at the PR head)
**Triage:** fix + pure-move refactor · 4 files · +201/−108 (309 lines) · regime **full** (2 Wave-1 agents + inline verification backstop)
**Stated intent:** PR title + body — spec-compliance **assessed**

### Findings

**low · medium · correctness · `tool-lane-render-agent.ts:212` + `:257` · ref:14c560e · citation-type:file-state**
In the eager path the head row's width is read inside `formatAgentHeader` at `flushSource` time and the descendants' width is read again inside `formatAgentChildren` at dispose time, so a resize between the two commits still emits a head row clamped to the old width above children clamped to the new one — the single-read coupling the PR documents holds only inside `formatAgentSummary`.
```ts
212:  return clampLineToTerminal(ancestorPrefix + head + agent.prefix, getTerminalWidth());
257:    getTerminalWidth(),
```
Residual, not a regression: pre-PR that header was unclamped entirely, so the divergence window narrows. *Suggestion:* record the width on `ToolEntry` when the eager header commits and reuse it in `formatAgentChildren`, or state the residual in that doc block.

**low · medium · perf-observability · `tool-lane-format-args.ts:335` · ref:14c560e · citation-type:file-state**
The plain tail clamp drops the dim dispatch-class tag (`[subagent]`/`[skill]`/`[dag]`) before it drops the nesting label, because `formatToolLine` appends `tagSuffix` last — so on an overflowing head row the class-of-work signal is elided first, which the intent sanctions only for the stats tail.
```ts
return dispatchTag ? head + palette.dim(tagSuffix) : head;
```
*Suggestion:* reserve `displayWidth(tagSuffix)` in the head-row clamp the way `formatGroupedToolResults` reserves its suffix (`tool-lane-render-grouped-root.ts:82`).

**low · high · test-coverage · `tool-lane.test.ts:788-863` · ref:14c560e · citation-type:file-state**
No test asserts that `formatAgentHeader` and `formatAgentSummary` produce the *same* head-row encoding for one entry — the PR's stated reason for touching `formatAgentHeader` at all, so a later edit to one clamp site alone passes the full suite. Absence grounded: grep `toEqual\(.*formatAgentHeader|formatAgentHeader\(.*\)\).toBe|same encoding` over `src/cli/commands/interactive` → 0 matches; the new test asserts widths + `toContain('agent')` (`:862`) only.
*Suggestion:* add an equality assertion between the two encodings at a pinned width.

**nit · high · api-compat · `tool-lane-render.ts:317`** — `formatGroupedToolResults` re-export has zero importers (only intra-module call at `grouped-root.ts:136`); pre-existing dead surface the move relocates. `[UNVERIFIED: no production importers]`
**nit · high · test-coverage · `tool-lane.test.ts:804`** — the `?? 88` oracle fallback is dead: `beforeEach` at `:523` pins `columns = 88` for the describe opened at `:515` (verified no intervening close), so oracle ≡ `getTerminalWidth()` = 88. Using `getTerminalWidth()` as the oracle removes the ambiguity.
**nit · high · style · `tool-lane-render-agent.ts:262-263`** — double blank line left by the excision (reported by both Wave-1 agents; deduped).
**nit · low · test-fragility · `__snapshots__/tool-lane.test.ts.snap:21` + `tool-lane.test.ts:1822`** — the `Snapshot pins` describe pins no width (grep `columns|defineProperty` over lines 1800-1900 → 0 matches), so scenarios 4/5 head rows become width-dependent after the clamp; widest content row is 59 cols and vitest workers are non-TTY (`getTerminalWidth()` → `DEFAULT_COLUMNS` 80, `terminal-size.ts:8,13-16`), so it is unreachable today. `[low confidence — verify with runtime context]`

**Out-of-diff, informational (not merge-relevant):** pre-existing C0/C1 control-byte passthrough in `agent.prefix` when `summarizeNestingArgs` fails open on unparseable JSON — `sanitizePrefixString` (`tool-lane-format-args.ts:333`) handles CR/LF only. Neither commit here introduces or worsens it.

### Verified clean (mechanically, at ref 14c560e)

- **ANSI safety:** `truncateDisplayWidth` emits each escape as an indivisible token and re-emits `\x1b[0m` + OSC-8 close (`display.ts:123-144`) — no mid-escape cut, no style or hyperlink leak past a clamped row.
- **Width guarantee:** `clampLineToTerminal` early-returns or budgets `max(0, maxWidth − 1)` + `…`, so `displayWidth ≤ cols`; an indent wider than `cols` yields prefix + `…` — no throw, no negative budget.
- **Spine alignment:** tail-only truncation preserves the `2 * (depth + 1)` spine cells; head row and children share one `cols` in the summary path (`:129`, `:158`).
- **"Pure move, no behavior change" — confirmed at git level:** the 81-line body is byte-identical between `d52dbf2:tool-lane-render-agent.ts` and `a40f35c:tool-lane-render-grouped-root.ts` modulo the added `export` (`diff` → empty). This closes the gap both agents flagged as unverified.
- **api-compat:** sole production importer `tool-lane.ts:13` resolves through the `tool-lane-render.ts:314-318` re-export (call sites `:809/827/889/908`); the import cycle *shortens* (`render→agent→grouped-root→render` becomes `render→grouped-root→render`) and cyclic bindings are used only inside function bodies — no module-eval TDZ. No unused import survives in the agent module.
- **Spec-compliance:** the "last unbounded rows in the flush path" claim holds — every flush emission is clamp-wrapped (`tool-lane-render-children.ts:394-485`, `grouped-root.ts:91-168`, `tool-lane.ts:668-908`); `renderTextChildLines` (`tool-lane-render.ts:285`) is the only unclamped producer and both flush callers clamp its output. No unmet intent, no scope creep.
- **350-LOC ceiling / PR claim:** `wc -l` → 264 and 174 exactly as stated.
- **Snapshots do not shift:** widest snapshot content row is 59 < 80, so the clamp changes no pinned output.

### What was not checked

- Citations: all read at **`14c560e`** (clean worktree ≡ PR head); move-identity additionally verified against parent `d52dbf2`. **No fabricated and no diff-only citations; no findings dropped.**
- **Tests were not executed** (review is read-only): the PR's `115/115`, `4973/4973`, `pnpm lint` clean, and "fails on the parent commit" claims are **unverified here**.
- Live overlay and `renderFlushChildren` paths — confirmed clamp-wrapped, not exercised.
- `stream-renderer*.ts` emission sites outside the tool lane; runtime `stringWidth` behavior for C1 bytes.
- `/shadow-verify` skipped by rule: it fires on `critical`/`high` findings and the highest surviving severity is `low`.

### Merge decision: **MERGE**

Zero critical/high/blocking findings. The fix does what its intent claims, the refactor is byte-provably a pure move, the export surface resolves, and the clamp is ANSI-safe with a hard width guarantee. All seven findings are `low`/`nit` and none block: the two `low` correctness/observability items are residuals narrower than the pre-PR state, and the missing encoding-parity test is a durability improvement, not a defect in this diff.

**Done**
- Reviewed PR #786 across security, api-compat, correctness, spec-compliance, test-coverage, perf-observability; 2 parallel Wave-1 research agents plus an inline citation/absence backstop.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/325ed018-1651-4ea9-b1a3-f26206df97b2/pr-786.diff`; all citations read at ref `14c560e`; move-identity proof via `git show d52dbf2:… ` vs `a40f35c:…` → byte-identical; report text above is the deliverable.
- World unchanged: read-only review — no files edited, nothing committed, pushed, or posted to the PR (findings are reported here, not filed as PR comments).
- Deferred: test execution (`pnpm test`, `pnpm lint`) not run, so the PR's own verification numbers remain unverified; the four nits and three `low` findings are left for the author's discretion since none block merge.

---

PR #765, opened 2026-07-30
feat(tui): make the subscription-usage indicator severity-graded and time-aware

griffinwork40:
## Summary

The subscription-usage indicator rendered `5h 94%` and `5h 24%` **identically** — flat `palette.chrome`, the same tone as the token counter — and carried `droppablePriority: 5`, so the most consequential number on the row was also the first field shed on a narrow terminal. `quota-cache.ts` had been parsing `fiveHourResetsAt` / `sevenDayResetsAt` / `observedAt` since it was written and nothing ever read them, so `94% with 12m left` and `94% with 4h left` looked the same.

Three commits, all TUI-only. No behaviour change outside rendering; no new dependency, env var, or SDK symbol.

### 1. `9d2b28d` — grade the status-line indicator by severity

New pure helper `src/cli/quota-indicator.ts` (247 LOC): per-window severity tone (mirrors `context-bar.ts`'s `>0.8` error / `>0.5` warning / else chrome so both usage readouts on the row escalate in step), a reset countdown on the binding window once it crosses caution, a `stale` flag + `~` marker so an idle session doesn't show a drained reading as live, and severity-inverted droppability — a fresh `critical` reading takes priority `0` (drop last, behind even the branch) instead of `5`, but only when fresh; a stale critical keeps the peripheral priority since the rolling window may have already drained.

`StatusLineFields.quota: string` → `quotaWindows: QuotaWindows` (raw, like `contextPct`/`contextLimit`). Severity is computed once, inside `formatQuotaIndicator` (`quota-indicator.ts`); `status-line.ts`'s render method consumes that single severity value to decide both tone (via the returned text) and droppability — `critical` + fresh → priority `0` (drop last, behind even the branch), everything else (calm, caution, or a *stale* critical) → priority `5` (drop first, same as before). So there's one decision site for droppability, fed by one severity computation upstream — not two independent severity notions to keep in sync.

### 2. `56efcd3` — warn in the turn footer past 80%

New `src/cli/quota-footer.ts` (131 LOC) adds the intrusive half beside the context-usage line it mirrors:

```
  context 91% used of 1.0M — approaching limit
  5h quota 94% used — resets in 12m
```

[...]

PR #791, opened 2026-07-31
fix(tui): schedule clock-derived quota repaints and condition the auto-resume promise

griffinwork40:
Follow-up to #765, which merged with two reviewer findings unaddressed. Both were live on `main`; this fixes them.

## 1. The status row never re-evaluated its own clock (Codex P2 on #765)

`formatQuotaIndicator` renders a reset countdown (`⟳12m`) and a `~` staleness marker, and `status-line.ts` grants a **fresh** critical reading drop-last priority `0` — all computed against `new Date()` at paint time. But every repaint of that row is event-driven: turn events, the git/context samplers, resize, and a *new* quota reading. An idle session fires none of them.

Consequence: the countdown froze at whatever it last painted, and the `~` marker could never appear in the idle scenario its own doc-comment names — the one quota-driven repaint (`onQuotaUpdate`) fires only when a new reading lands, i.e. exactly when `stale` becomes false, so it can never be the paint that reveals staleness. A stale critical also kept priority `0` past `STALE_AFTER_MS`.

`StatusLine` now owns a low-frequency repaint ticker:

- armed in `start()`, never in the constructor — `terminal-compositor.ts` sets the house rule that construction stays free of an auto-started recurring timer, and `caretBlink` sets the precedent that enablement is resolved caller-side. `tickMs` therefore **defaults to 0 (off)**; `bootstrap.ts` passes `30_000` so only the REPL ticks.
- ticks via the existing `flush()`, which re-renders from `lastFields`, so every clock-derived value recomputes.
- `.unref()`ed, so it never holds the process open.
- cleared at the **top** of `stop()`, ahead of that method's `!started || !enabled` early return — an armed ticker is the one piece of state that can outlive a no-longer-started line, so a late return must not skip its teardown.

## 2. The footer promised an auto-resume the runtime will not deliver (Codex P1 on #765)

`quota-footer.ts` appended `AFK pauses and auto-resumes at the cap` to every `near`/`over` line. That is false in two reachable configurations, both verified in source:

- `retry-layer.ts:443` — when `resetsAt - now > TWO_HOURS_MS` the layer **surfaces the error instead of waiting**. A hot 7d window routinely resets days out.
- `retry-layer.ts:358` — when `autoResumeOnUsageLimit` is off it stops rather than resuming.

The promise is now conditional (`capNote`). The deadline gate is local to the formatter; the flag is resolved at the call site by **`resolveAutoResumeOnUsageLimit()`** (see round 2 below for why that is not `loadConfig()`). When the promise is withheld the line still carries the window, percentage and countdown, and says what actually happens (`turns stop at the cap — …`) rather than dropping to an alarming bare number.

An **unknown** deadline deliberately keeps the promise: with no `resetsAt` the layer polls within its own budget rather than bailing early, so parking is still the real behaviour.

### On the duplicated threshold

`AUTO_RESUME_MAX_LEAD_MS` is a local copy, not a runtime import — `retry-layer.ts` sits behind the provider boundary and importing a value from it would drag the provider graph into the REPL's render path. `TWO_HOURS_MS` is now exported so **the test** imports the real constant and asserts the boundary both sides of it, making drift a test failure rather than a silent re-break.

## Round 2 — fixing what review caught in `78b4fde` (`ed57212`)

Two reviewers independently found the same defect, and both were right: **the flag gate I added was inert.**

`loadConfig()` builds its return value as an explicit field whitelist that omits `autoResumeOnUsageLimit` (0 occurrences in the literal), so `loadConfig().autoResumeOnUsageLimit` is permanently `undefined` and `?? true` always won. The "auto-resume is off" path was unreachable through real CLI configuration. The original PR text claimed "the flag is read at the call site" — that claim was false, and it is corrected above.

The second finding was a convention violation in the same line: `printTurnFooter` called the **side-effectful** `loadConfig()` once per turn footer, re-installing process-global model-slot bindings via `setSlotBindings`. The original text justified this as "the same inline-config pattern as `loadTelegramConfig()`" — that had it exactly backwards. `loadTelegramConfig` *is* the memoized-tier reader; `loadConfig` is precisely what that convention exists to avoid (`config.ts:96-103`).

Both are fixed by one new resolver, a direct sibling of the two that already exist:

```ts
export function resolveAutoResumeOnUsageLimit(): boolean {
  return loadJsonConfig().config.autoResumeOnUsageLimit ?? true;
}
```

Third finding (test-coverage) is fixed too, and it is the interesting one: the five `quota-footer.test.ts` cases pass `{ autoResume }` **explicitly**, so a dead read at the production call site was invisible to them — which is exactly how 13,754 green tests shipped an inert branch. There are now two `printTurnFooter` cases driving the real wiring, plus a `config.test.ts` case that asserts the resolver reads what `loadConfig()` drops:

```ts
expect(resolveAutoResumeOnUsageLimit()).toBe(false);
expect(loadConfig().autoResumeOnUsageLimit).toBeUndefined();  // fails if anyone re-routes it
```

### Not fixed here — pre-existing, wider than this PR

`autoResumeOnUsageLimit` is advertised as settable (`settable-keys.ts:221`) but is non-functional across the whole CLI for the same root cause: `chat.ts:741` and `bootstrap.ts:740` both guard on `cliConfig.autoResumeOnUsageLimit !== undefined` and always take the empty branch, so the provider never receives it. Setting it in `afk.config.json` changes nothing on the REPL/chat surfaces today. Fixing that means making a dormant setting suddenly live for the provider — a behaviour change that deserves its own PR and its own testing, so this PR routes only the display surface. Flagged by the reviewer as worth its own issue; I agree.

## Tests

- `status-line.test.ts` — no ticker by default; none on a disabled/non-TTY line; ticks repaint and `stop()` ends them.
- `quota-footer.test.ts` — promise withheld when auto-resume is off; withheld beyond the ceiling; kept at/below it (drift guard against the real `TWO_HOURS_MS`); kept when no deadline is known; promised by default.
- `turn-handler.test.ts` — the production call path, both flag states.
- `config.test.ts` — the resolver reads the field `loadConfig()` drops.
- All 12 pre-existing footer tests pass **unchanged** — the existing near/over cases use a 12-minute deadline, inside the ceiling.

**Gates — all seven CI checks run locally at `ed57212`:** `pnpm lint` ✓ · `pnpm test` **721 files / 13,757 passed, 2 skipped, 0 failures** ✓ · `audit:env:check` ✓ · `scan:env:check` ✓ · `audit:sdk:check` ✓ · `audit:chalk:check` ✓ · `audit:deps` ✓ · `pnpm build` ✓

## The `docs/env-registry.md` commit is unrelated — inherited red CI on main

`b0ac17e` regenerates `docs/env-registry.md`. **It is not part of this fix.** `scan:env:check` was failing on `main`: commit `219e0b6` (v5.83.1's stall watchdog) added a 139th env var to `src/config/env.ts` but left the generated header count at 138, so *every* PR opened against main inherited that failure. The change is one line of `pnpm scan:env` output with no source change, split into its own commit so it can be dropped or cherry-picked independently.

## Reviewer note

Authored inline by the orchestrating session rather than by an isolated implementation sub-agent (four sub-agent dispatches failed: three to rate-limit or stream cutoffs, one to a 45-minute timeout). Round 1 shipped an inert branch that two independent reviewers caught — treat the absence of an independent authoring pass as a real signal and read closely.

One deliberate test gap remains: the `.unref()` call itself is not asserted. Asserting it needs a `setInterval` spy that fights the fake-timer harness, so it is covered by the teardown test and by the comment, not directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ auto-release workflow to deduplicate commits across successive runs.
 
 ## [Unreleased]
 
+### Fixed
+- tick the status line on a 30s timer so the quota reset countdown and the stale-reading marker stay true on an idle session instead of freezing at their last event-driven paint
+- stop the turn footer promising pause-and-auto-resume when the runtime will not deliver it — the promise is now conditional on `autoResumeOnUsageLimit` and on the reset landing inside the retry layer's two-hour wait ceiling
+
 ## [5.83.1] - 2026-07-31
 
 ### Fixed
--- a/docs/env-registry.md
+++ b/docs/env-registry.md
@@ -2,7 +2,7 @@
 
 Generated from `src/config/env.ts`. Do not edit by hand — run `pnpm scan:env` after changing the registry source.
 
-**138 vars** across 12 categories. Every `process.env[...]` read in `src/` outside `src/config/env.ts` is a CI failure (enforced by `pnpm audit:env:check`).
+**139 vars** across 12 categories. Every `process.env[...]` read in `src/` outside `src/config/env.ts` is a CI failure (enforced by `pnpm audit:env:check`).
 
 To add a var: edit `src/config/env.ts` (add a getter on `env` + an entry in `ENV_REGISTRY`), then run `pnpm scan:env`.
 
--- a/src/agent/providers/anthropic-direct/query/retry-layer.ts
+++ b/src/agent/providers/anthropic-direct/query/retry-layer.ts
@@ -49,7 +49,13 @@ import type {
   RunTurnInput,
 } from '../types.js';
 
-const TWO_HOURS_MS = 2 * 60 * 60 * 1000;
+/**
+ * Both the total usage-limit polling budget and the maximum reset lead time the
+ * layer will wait out. Exported so surfaces that *describe* this behaviour to
+ * the user (see `cli/quota-footer.ts`) can be drift-tested against the real
+ * threshold instead of hardcoding a second copy of it.
+ */
+export const TWO_HOURS_MS = 2 * 60 * 60 * 1000;
 
 // Transient rate-limit (429 + retry-after header, no `|ts`) replay budget.
 // The SDK has already auto-retried twice by the time the error surfaces here,
--- a/src/cli/commands/interactive/bootstrap.ts
+++ b/src/cli/commands/interactive/bootstrap.ts
@@ -680,7 +680,11 @@ export async function bootstrapSession(
   // its inter-turn raw writes through statusLine.withFullScrollRegion(...) —
   // see repl-renderer.ts for the DECSTBM sub-region scroll-loss contract this
   // wiring is defending against.
-  const statusLine = new StatusLine();
+  // `tickMs` is resolved caller-side (StatusLine defaults it off, like the
+  // compositor's caret blink) so only the REPL gets a recurring timer. 30s is
+  // half the quota countdown's minute granularity, which is the finest
+  // clock-derived value the row renders.
+  const statusLine = new StatusLine({ tickMs: 30_000 });
   const replRenderer = createReplRenderer(process.stdout, { statusLine });
 
   // Stable hookRegistry shared across sessions (including swaps).
--- a/src/cli/commands/interactive/turn-handler.test.ts
+++ b/src/cli/commands/interactive/turn-handler.test.ts
@@ -16,6 +16,15 @@ import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import * as path from 'node:path';
 import { runTurn, formatContextUsage, printTurnFooter } from './turn-handler.js';
 import { recordQuotaSnapshot, resetQuotaCacheForTests } from '../../../agent/quota-cache.js';
+
+// Only `resolveAutoResumeOnUsageLimit` is stubbed; every other config export
+// stays real. Defaults to `true` (the production default) so the ~70 other
+// tests in this file are unaffected — a case opts in by flipping the stub.
+const configStub = vi.hoisted(() => ({ autoResume: true }));
+vi.mock('../../config.js', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('../../config.js')>()),
+  resolveAutoResumeOnUsageLimit: () => configStub.autoResume,
+}));
 import { getCurrentSink } from '../../../agent/_lib/skill-sink-channel.js';
 import type { AgentSession } from '../../../agent/session.js';
 import type { OutputEvent } from '../../../agent/types.js';
@@ -2322,9 +2331,11 @@ describe('formatContextUsage — proactive escalating context tiers', () => {
 describe('printTurnFooter — subscription-quota line', () => {
   beforeEach(() => {
     resetQuotaCacheForTests();
+    configStub.autoResume = true;
   });
   afterEach(() => {
     resetQuotaCacheForTests();
+    configStub.autoResume = true;
   });
 
   const footerLines = (): string[] => {
@@ -2352,6 +2363,34 @@ describe('printTurnFooter — subscription-quota line', () => {
     const out = footerLines().join('\n');
     expect(out).toContain('5h quota 94% used');
     expect(out).toContain('resets in 12m');
+  });
+
+  it('honours a config-set autoResumeOnUsageLimit=false through the real call path', () => {
+    // Covers the wiring, not just the pure formatter: the five quota-footer.ts
+    // unit tests pass `{ autoResume }` explicitly, so a DEAD read at this call
+    // site stayed invisible to them. That is exactly how the first cut of this
+    // shipped — `loadConfig().autoResumeOnUsageLimit` is permanently undefined,
+    // so the promise was unconditional and 13,754 green tests missed it.
+    configStub.autoResume = false;
+    recordQuotaSnapshot({
+      fiveHourUtilization: 0.97,
+      fiveHourResetsAt: new Date(Date.now() + 12 * 60_000),
+      observedAt: new Date(),
+    });
+    const out = footerLines().join('\n');
+    expect(out).toContain('5h quota 97% used');
+    expect(out).not.toContain('auto-resumes');
+    expect(out).toContain('auto-resume is off');
+  });
+
+  it('promises auto-resume through the real call path when the flag is on', () => {
+    configStub.autoResume = true;
+    recordQuotaSnapshot({
+      fiveHourUtilization: 0.97,
+      fiveHourResetsAt: new Date(Date.now() + 12 * 60_000),
+      observedAt: new Date(),
+    });
+    expect(footerLines().join('\n')).toContain('AFK pauses and auto-resumes at the cap');
   });
 
   it('orders the quota line AFTER the context line — nearest deadline reads first', () => {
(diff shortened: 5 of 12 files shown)


griffinwork40 (commented):

## Verdict: DO NOT MERGE — one line short of its own headline claim (no risk; nothing here is unsafe)

Reviewed at ref `b0ac17e`. Regime: light (hotfix, 9 files, 204+/8−). All citations below were re-read against `b0ac17e` with `git show`, not against the working tree.

**INTENT 1 (the clock ticker) is fully met and clean.** **INTENT 2 is half-inert:** the ceiling branch is genuinely fixed; the `autoResumeOnUsageLimit` branch reads a field `loadConfig()` never returns.

---

### medium · confidence high · spec-compliance / correctness — `src/cli/config.ts:144-171` + `src/cli/commands/interactive/turn-handler.ts:954` (ref `b0ac17e`, citation-type: file-state + diff-context)

`loadConfig()` builds its return value as an explicit field whitelist that omits `autoResumeOnUsageLimit`, so `loadConfig().autoResumeOnUsageLimit` is always `undefined` and `?? true` always yields `true` — the flag branch added here can never fire.

```ts
// config.ts:144-171 — whitelist literal; `merged` carries the field, this drops it
const config: CliConfig = { model: ..., updatePolicy: ...,
  ...(merged.importFrom !== undefined ? { importFrom: merged.importFrom } : {}),
  ...(merged.theme !== undefined ? { theme: merged.theme } : {}), };
```

Verified: `json-tier.ts:225-226` **does** populate the field from `afk.config.json`, and it **is** on the `CliConfig` type (`config/types.ts:228`) — it is lost only in this literal. `grep` of the literal block for the field returns 0 hits at the ref.

**This is not a user-visible false promise**, and that is worth stating plainly, because it is the reason this is `medium` and not `high`: the same dead field also neuters the runtime plumbing (`bootstrap.ts:740` → `bootstrap.ts:137-139` empty branch → `query.ts:310 autoResumeOnUsageLimit: opts.autoResumeOnUsageLimit ?? true`), so the REPL genuinely does auto-resume and the footer's `true` is accidentally accurate. The defect is that the PR's headline claim — "the flag is read at the call site" — is inert, and the doc comment asserting it will mislead the next reader.

**Suggestion (one line, fixes this finding and the next one together):** read the side-effect-free memoized tier, which *does* carry the field:
```ts
autoResume: loadJsonConfig().config.autoResumeOnUsageLimit ?? true,
```

### medium · confidence high · correctness / convention — `src/cli/commands/interactive/turn-handler.ts:952-955` (ref `b0ac17e`, citation-type: diff-context)

`printTurnFooter` now calls the side-effectful `loadConfig()` on every turn footer, which re-installs process-global model-slot bindings via `config.ts:179 setSlotBindings(slotBindings)`. The PR justifies this as "same inline-config pattern as the `loadTelegramConfig()` call" — the repo's own doc comment at `config.ts:99-103` states the opposite rule for exactly this class of surface:

```ts
// resolveCliPermissionMode: "Reads the memoized JSON directly so it is
// side-effect-free and never throws on the `local-*`-model guard that the
// full `loadConfig` enforces — same rationale as `loadTelegramConfig`."
```

Display/status surfaces read the memoized tier; `loadConfig` is the thing that convention exists to avoid. The `local-*` throw at `config.ts:184-194` is *not* separately reachable here (startup `loadConfig()` at `bootstrap.ts:203` pre-empts it, and `loadJsonConfig` is memoized so the decision cannot change mid-session), and the slot re-install is idempotent in the REPL because no per-session divergence exists — I checked all production `setSlotBindings` call sites (`config.ts:179`, `agent-session.ts:264`) and found no slash command or session swap that installs a different table. So this is a convention violation with no live blast radius, not a runtime bug. **Suggestion:** as above — `loadJsonConfig()` satisfies both findings at once.

### low · confidence high · test-coverage — `src/cli/commands/interactive/turn-handler.test.ts` (ref `b0ac17e`, citation-type: file-state)

No test exercises the `loadConfig()` → `formatQuotaUsage` wiring; the five new footer tests pass `{ autoResume }` explicitly, so the dead read is invisible to CI. That is precisely why 13,754 green tests do not catch finding 1. **Suggestion:** one `printTurnFooter` case with `autoResumeOnUsageLimit: false` on disk asserting the promise is absent.

### Observation (pre-existing, wider than this PR — not a blocker)

`autoResumeOnUsageLimit` is advertised as settable (`settable-keys.ts:221`, tier `agent`) but is non-functional across the whole CLI for the same reason: both `chat.ts:741` and `bootstrap.ts:740` guard on `cliConfig.autoResumeOnUsageLimit !== undefined` and always take the empty branch. A user setting it in `afk.config.json` changes nothing on the REPL/chat surfaces. This PR's new read is the third consumer to inherit it, and the one-line `config.ts` fix would light up all three at once. Worth its own issue.

---

### Verified good (checked, no issues found)

- **security** — no secrets, no injection sink, no auth surface; footer interpolates only labels/percentages and is asserted ANSI-free.
- **api-compat** — `formatQuotaUsage`'s third param is optional with `= {}`; the only production caller is `turn-handler.ts:949`. `TWO_HOURS_MS` private→exported is widening. No breaking change.
- **ticker lifecycle (INTENT 1)** — `start()` returns at `if (this.started || !this.enabled) return;` *before* the arm block, so a double `start()` cannot leak an interval and a disabled/non-TTY line never arms one; `flush()` is null-safe (`if (this.lastFields) this.repaint(...)`); the clear sits at the top of `stop()` ahead of the `!started || !enabled` early return exactly as described; `.unref()` present. The stop→start swap at `bootstrap.ts:814-818` re-arms exactly one.
- **drift guard** — it really does fail in both directions: enlarging `TWO_HOURS_MS` breaks the `atCeiling` assertion, shrinking it breaks `pastCeiling`. The local-copy rationale holds — `quota-footer.ts` has zero `retry-layer` imports; only the test imports it.
- **perf** — `loadConfig` is backed by memoized tiers (`json-tier.ts` / `afk-md-tier.ts` early-return their caches), so no per-turn disk I/O. The 30s tick zeroes `lastRepaint` and so bypasses the throttle by design: one write per 30s idle.
- **`docs/env-registry.md` 138→139** — legitimate regeneration, confirmed at the ref: `src/config/env.ts` has exactly **139** registry entries and the header now reads 139. The author's account of the inherited-red `scan:env:check` checks out. Author-disclosed scope creep, cleanly commit-split.

### What was not checked

- Citations verified inline against PR head `b0ac17e` (fetched locally); `file-state` citations re-read with `git show b0ac17e:<path>`, not from the working tree.
- Did not run the test suite — CI is green on all seven checks (Lint & Build, Test ubuntu/macos, real PTY scrollback, Docs Site, Publish Bundle Smoke); I took those at face value.
- Did not verify ticker behaviour on a real PTY (the "real PTY scrollback" job is green), and did not audit `usage-limit.ts` polling-budget semantics for the unknown-deadline case — the author's reasoning that an absent `resetsAt` still parks is accepted, not independently derived.
- The `.unref()` assertion gap is author-disclosed and accepted as reasonable.
- Stated intent: PR #791 title + body — spec-compliance **was** assessed against it.

**Why hold rather than merge:** nothing here is unsafe and INTENT 1 plus the ceiling half of INTENT 2 are genuinely correct, so this is a strict improvement over `main`. The hold is narrow: the PR body's central claim about reading the flag is inert, and a maintainer merging on the strength of that description would be merging on a false premise. Swapping `loadConfig()` for `loadJsonConfig()` on one line makes the claim true, honours the repo's own documented convention for status surfaces, and drops both `medium` findings.

griffinwork40:
@codex review

griffinwork40:
@codex fix comments

PR #793, opened 2026-07-31
feat(awa): fold the fragment-goal signal class into intent-lock

griffinwork40:
## What

Adds a fifth signal class to the bundled `intent-lock` skill: **"No task statement at all"** — a request that is a bare fragment with no imperative verb (a lone path, a repo URL, a noun phrase, a pasted stack trace).

For the four existing classes the ambiguous thing is a noun *inside* a stated task. Here there is no stated task, so the referent to resolve is the **goal itself**. The agent reconstructs one from ambient evidence either way; today that reconstruction is invisible and never validated against what the user meant.

Three additions, all inside the existing structure:

- **Step 1** gains an evidence-cited reconstruction rule — resolve in-context by default, dispatch a sub-agent only when in-context reasoning cannot produce a single confident reading.
- **Step 4** gains a lock format that names both the reading and the evidence it rests on, so a wrong read is visibly wrong rather than silently load-bearing.
- **Exit criteria** extend to cover it.

## Provenance

Sourced from a `/forge` brief (origin card `when-a-sessions-goal-is-expressed-as-a-fragment-of-2026-06-17t21-34-15-211`, 3 sessions, scored 27/35).

Forge generated a standalone candidate skill named `intent-recon`. Two independent qualify iterations returned **SALVAGE**, both citing ~55–60% overlap with the already-shipped `intent-lock` on the exact failure mode the candidate existed to fix, and both converging on folding it in here rather than shipping a near-duplicate gate. This PR is that fold.

## Deliberately not done

Per the same review, two things were left out on purpose:

- **No session-end fidelity check.** The candidate proposed a `MATCHED/DRIFTED/MISMATCHED` verdict that is structurally identical to `trajectory-check`'s existing one. That is `trajectory-check` invoked with the locked statement as its canonical-goal input — not new machinery here.
- **References to `ask-gate` / `trajectory-check` are hedged** behind "if present in this installation". Both live only in the user-scope `~/.afk/skills/` directory and are not shipped by this repo, so an unhedged reference would be a dead link for every other install. The `ask-gate` reachability rule is stated inline as the fallback.

## Hash pin

`intent-lock` is bundled-only, so the pinned-hash bump in `bundled.test.ts` (`a0844035` → `026fd980`) needs no parallel upstream PR — documented here per the convention recorded on the pin itself.

## Verification

- `pnpm test src/bundled-plugins/awa-bundled/bundled.test.ts tests/copy-bundled-plugins.test.ts` — **22/22 passing** on this commit
- Prose-only change to one skill body plus its hash pin; no runtime code touched


griffinwork40 (commented):

## Merge decision: **DO NOT MERGE** (blocked on red CI; 2 medium findings)

Reviewed ref `d118ec5e`. Regime: **light** (22 lines, 2 files). Prose change to one bundled skill body + its hash pin.

**Verified inline, so not findings:** the new pin `026fd980…` is the correct SHA-256 of `SKILL.md` at this ref. All three claimed additions (Step 1 reconstruction rule, Step 4 lock format, extended exit criteria) do land in the diff. The `ask-gate` / `trajectory-check` hedge is disclosed in the PR body — not scope creep. Neither skill is shipped by this repo (`git ls-tree` at ref: no `skills/ask-gate/` or `skills/trajectory-check/`), so the hedge's premise holds.

---

### Blocking

**CI is red on this ref** — `Test (ubuntu-latest)` and `Test (macos-latest)` both FAIL. The single failure is `src/cli/commands/interactive/turn-handler.test.ts` → *"prints the quota line with its deadline once the binding window is hot"*: `expected '… 5h quota 94% used…' to contain 'resets in 12m'`, received `resets in 11m`. 720 files / 13744 tests pass; 1 fails. That file is **not in this diff** and is unreachable from a prose-only change — this reads as a clock-boundary flake, not a regression introduced here. It still gates merge: re-run the job and confirm green before landing.

---

### Findings

**1 · medium · confidence medium · spec-compliance · `skills/intent-lock/SKILL.md:3` · ref:d118ec5e · citation-type:file-state (re-read at ref ✓)**
The frontmatter `description` is the model-facing routing trigger and still enumerates only three signal classes, so the skill will not auto-fire on the one input class this PR exists to catch — a bare fragment has no other lexical trigger to match on.
```
description: "Fires before multi-step work when the user's request contains ambiguous referents ('the text', 'her Y'), characterizations of unverified entities ('the meeting is substantive'), or identity assumptions (which contact = the user).
```
The 4th class (code-vs-runtime dual referent) is already missing, so this PR widens pre-existing drift rather than creating it.
*Fix:* append `…, or no task statement at all (a bare path, URL, noun phrase, or pasted trace)` to the description, then re-bump the pin.

**2 · medium · confidence high · correctness · `SKILL.md:37` and `SKILL.md:46` · ref:d118ec5e · citation-type:diff-context**
The new class's continuation-cue example and the new Skip-when bullet describe the same scenario with opposite verdicts, so the procedure returns no determinate answer for `"ok now the other one"` — the signal class says fire, the skip rule says skip.
```
37: - A continuation cue assuming carried-over context: "ok now the other one", "same as before"
46: - Fragment's goal is fixed by the immediately prior turn (agent just proposed two options and the user named one) — the task statement carries over intact.
```
Compounding: `"the other one"` is already a listed ordinal under **Ambiguous referents** (`SKILL.md:14` — *"Ordinals with unclear basis: … 'the other branch'"*), so one utterance hits three rules.
*Fix:* narrow the continuation-cue example to a case the skip rule does not cover, or scope the skip bullet to "…and the prior turn enumerated the options explicitly."

**3 · low · confidence high · correctness · `SKILL.md:104` · ref:d118ec5e · citation-type:diff-context**
The new paragraph tells the agent `ask-gate` might be absent and to reference it only if present, but this repo enforces that skill's reachability rule as a deterministic `PreToolUse` hook that fires regardless of skill installation — the hedge understates the shipping surface.
```
**ask-gate** and **trajectory-check** are user-scope skills that may not exist in this installation — reference them only if present, and never block on them.
```
Contradicted by `src/agent/ask-question-gate.ts` at this ref: *"Migration provenance: first gate-skill → hook migration (the `ask-gate` skill's reachability rule, made deterministic)."* Behaviour converges either way — this is an accuracy defect, not a wrong-output one.
*Fix:* replace the `ask-gate` half with "the reachability rule is enforced by the `ask_question` hook in agent-afk; where that is absent, apply it inline."

**4 · low · confidence high · api-compat · `prompts/system-prompt.md:116`, `website/content/docs/skills/bundled.mdx:168` · ref:d118ec5e · citation-type:file-state (re-read at ref ✓)**
Two other surfaces enumerate intent-lock's signal classes and now drift from `SKILL.md`; the system-prompt hint is the surface that actually routes the guard automatically.
```
system-prompt.md:116 — scan the request for: ambiguous referents ("the text", "her Y"), unverified characterizations ("the meeting is substantive"), or identity assumptions (which contact = the user).
bundled.mdx:168 — Scans the request for ambiguous referents …, unverified status claims …, identity assumptions, and code-vs-runtime dual referents …
```
Three classes and four classes respectively, against five in the skill.
*Fix:* update both enumerations in this PR — neither file is hash-pinned, so no extra bump.

**5 · low · confidence high · test-coverage · `bundled.test.ts:87-88` · ref:d118ec5e · citation-type:diff-context**
The hash pin is the only test surface for this skill, so every prose defect above lands green the moment the pin is bumped. `git grep intent-lock -- '*.test.ts'` at the ref returns exactly 2 hits — `bundled.test.ts:84` (a comment) and `:87` (the pin key) — and no assertion on the body's structure.
*Fix:* add a content invariant asserting the five class headings and the reconstructed-goal lock format are present, following the existing per-skill invariant precedent in the same file.

**6 · low · confidence medium · api-compat · `docs/skill-load-mode.md:144` · ref:d118ec5e · citation-type:file-state (re-read at ref ✓)**
The new Step-1 rule introduces a conditional sub-agent dispatch, which sits awkwardly next to the load-mode doc's guard framing.
```
  `intent-lock`). A zero-dispatch guard is ALWAYS `load`: a forked guard returns
```
Scoping note, since the citation supports less than it first appears: line 144 lists `intent-lock` under the "Otherwise → `load`" bullet and *then* makes a separate general claim about zero-dispatch guards — it does not literally label intent-lock zero-dispatch. The `context: load` pin also stays correct regardless, because lines 133-137 explicitly permit a dispatching skill to remain `load`. Only the surrounding characterization goes stale.
*Fix:* amend the intent-lock mention to "in-context guard with a bounded reconstruction dispatch."

**7 · nit · confidence low · perf-observability · `SKILL.md:56` · ref:d118ec5e · citation-type:diff-context · [low confidence — verify with runtime context]**
The new dispatch path names no model or turn budget, inside a guard that fires before *any* multi-step task.
```
Resolve this in-context; dispatch a sub-agent to reconstruct only when in-context reasoning cannot produce a single confident reading.
```
The "only when in-context reasoning cannot" condition bounds frequency, which is most of the mitigation.
*Fix:* name a sized dispatch (one focused question, `max_turns` capped), consistent with the `right-size-delegation` gate this repo already registers in `skill-executor/telemetry.ts`.

---

### Dropped in verification (disclosed for completeness)

One finding was **dropped as fabricated**. A reviewer flagged a doubled conjunction in the rewritten exit-criteria line. Re-reading `SKILL.md:112` at the ref shows the punctuation is correct (`…identity assumption, code-vs-runtime dual referent, and missing task statement is either locked…`). The error originated in the orchestrator's transcription of the hunk into the reviewer's brief, not in the PR. **The exit-criteria line is clean — no action needed.**

### What was not checked

- Citations verified inline against branch HEAD `d118ec5e` via `git show`/`git grep` at that ref. Every surviving `file-state` citation was re-read at the ref; the absence claim in finding 5 was re-derived with `git grep` at the ref.
- Stated intent: PR #793 title + body — **spec-compliance assessed** against it.
- Only the `Test (ubuntu-latest)` job log was read. `Test (macos-latest)` also reports FAILURE; its log was not opened, so the assumption that it is the same clock-boundary assertion is unconfirmed.
- No tests were run locally, and the skill was not executed against a real fragment input — the correctness findings are read from the procedure text, not from observed behaviour.
- Whether `~/.afk/skills/trajectory-check` exists on any given install (outside this repo's tree).

griffinwork40:
@codex fix comments

PR #796, opened 2026-07-31
fix(subagent): re-dispatch a read-only child once when its stream is cut with zero output

griffinwork40:
## Why

A forked sub-agent whose model stream gets closed mid-generation by an intermediary **dies for good**. The provider loop re-drives that cut twice per tool-use round (`STREAM_INCOMPLETE_MAX_RETRIES`, `loop.ts:128`), but once the budget is spent nothing above it retries — `handle.ts:555` throws `StreamIncompleteError` whose own message states the intended contract (*"the parent should retry or fall back"*) and no layer implemented it. `dag.ts` has no retry at all. The parent just gets an `isError` payload; only the parent **model** can choose to re-call.

The error message itself is **not a bug** — it is the deliberate anti-truncation guard from #635/#628 (`translate.ts:349`) correctly refusing to present a partial as a finished answer.

## Forensics

12,478 witness traces: 115 failure events across 74 traces, against a 2,997-child control set. Every correlation base-rated.

| finding | failing set | control |
|---|---|---|
| ran >= 360s | **100%** | 30.7% |
| median lifetime | **1119s (~19 min)** | 184s |
| was a nested orchestrator | **40.9%** | 8.2% |

- **Duration is the discriminator — no failure has ever occurred under 6 minutes.** It is a duration-exposure effect: the longer a stream stays open, the likelier an intermediary reaps it.
- Nested orchestrators are **5x** enriched (each absorbing up to `MODEL_CAP_BYTES=100_000` per nested return, `_output-cap.ts:47`) — but **59% of failures are leaves**, so that is a multiplier, not the mechanism.
- Rate rose **~6.7x per-dispatch** on top of a **6.5x volume** increase, while the orchestrator share *declined* and **no commit touched `providers/` or `subagent*`** in that window — consistent with external gateway behaviour, not a regression.
- Model is **not** a factor: the "95% opus_1m" signal was an artifact of reading the *parent session*'s model field. Failing children are 87% `sonnet`.

## What this does

Adds `src/agent/subagent/stream-cut-retry.ts` — a pure decision module plus a wrapper that re-runs the **full** `executeOnce` body, so a re-dispatch forks a genuinely fresh child (new session, trace, worktree) rather than re-awaiting a spent handle. `forkSubagent` call-count is the load-bearing test assertion.

## Safety: "zero output" is NOT "zero side effects"

The naive version of this fix — retry whenever `partialOutputBytes == 0` — **would have corrupted user state**, and an adversarial review caught it before merge. `handle.ts` accumulates streamed text (`:413`) and tool calls (`:415`) *separately*, and `:513-515` explicitly names tool-loops-without-a-final-turn as reaching the zero-output branch. So a write-capable child can run dozens of tool calls — file writes, `git commit`, HTTP POSTs — emit no text, and report zero output. Re-running its prompt would double-fire all of it.

Two gates therefore guard every re-dispatch:

1. **`isChildReplaySafe(...) === true`** (`tools/subagent/retry-safety.ts`, surfaced as `childSideEffectFree`). This is *stricter* than filesystem write capability: it proves the **transitive reach** of a replayed prompt, not just the child's own tool names. Two conditions, both fail-closed:
   - every granted tool is a member of `STREAM_CUT_RETRY_SAFE_TOOLS` (`tool-category.ts`) — an unrestricted surface (`undefined`) is refused rather than guessed at, and `bash` is absent because its read-only classifier is best-effort and cannot prove replay safety;
   - if the surface grants nested dispatch, that grant must be **mechanically scoped**. `agent` is admitted on the strength of scoped grants like `research-agent`'s `Agent(git-investigator)`; a **bare** `Agent` token resolves to unrestricted nesting (`agents/resolve.ts:183`), leaves `nestedAgentAllowlist` unset, and could reach `general-purpose` (inherit-all: `write_file`/`edit_file`/`bash`) as a **grandchild** — so it is refused.

   `childWriteCapable` is unchanged and still serves worktree isolation only. Because `replay-safe ⟹ !write-capable`, this also removes the stranded-dirty-worktree hazard: isolated worktrees exist only for write-capable children, which are never replayed.
2. **A monotonic cancel generation must not have moved.** Across the retry gap both in-flight handle maps are empty, so `cancelActiveForeground()` finds nothing to cancel and never aborts `call.signal` — without the counter a user cancel during the settle delay would be silently ignored and a fresh child forked anyway.

Also excluded from retry: **buffered partials** (`status:'succeeded'` — retrying would *destroy* salvaged findings) and **`tool_use_loop_capped`** (a requested ceiling, not a transport failure).

Budget is **1**, deliberately below the provider loop's 2: by this layer the provider has already spent its per-round budget, so a re-dispatch re-runs a whole multi-minute prompt.

## Accepted, documented costs

- `SubagentStop` fires once per attempt; attempt 1's `injectContext` is discarded.
- Telemetry emits one `subagent.failed` row per attempt with no `attempt` dimension.

Both are inherent to a retry genuinely being a second child. Attempt-scoped trace events are kept on purpose so a rescued failure stays visible in `afk trace show`.

## Not in scope

`dag.ts` nodes and the skill/mint dispatch paths still do not retry. The module is reusable; they can adopt it incrementally.

## Verification

- `pnpm test` — **723 files / 13,792 tests pass**, 0 failures. Coverage thresholds enforced (74/79/82/74).
- `pnpm lint` (`tsc --noEmit`) clean.
- New module: **100%** statements / branches / functions / lines.
- New tests across `child-config.test.ts`, `stream-cut-retry.test.ts`, and `subagent-executor.stream-cut-retry.test.ts`, including: a write-capable child **refusing** to retry; an **unscoped bare `Agent` grant refusing to retry** (with scoped `Agent(type)` and deny-all `Agent()` staying eligible); a cancel-during-gap regression guard; a mid-fork teardown guard; and an anti-drift test asserting the predicate against the *real* `incompleteToolResultFields` producer rather than a hand-rolled fixture.

Full diagnosis: `~/.afk/workspace/reports/subagent-stream-cut-2026-07-31.md`


griffinwork40:
@codex fix comments

griffinwork40:
Wave 1.5 + shadow-verify complete. All pivotal citations re-derived at the ref; one `high` finding **inverted** and one narrowed.

---

# Code Review — PR #796

**`fix(subagent): re-dispatch a read-only child once when its stream is cut with zero output`**
`afk/subagent-stream-cut-retry` → `main` · ref `b90271e` · 6 files, +801/−2 · regime **full**

## Blocking

### 1 · `high` · confidence high · correctness + spec-compliance — defeats stated purpose
`src/agent/tools/subagent/child-config.ts:213-218` · ref:`b90271e` · citation-type: diff-context

**The retry never fires for `research-agent`, the dominant read-only dispatch surface in this codebase** — so the fix does not cover the case its own forensics describe.

```ts
const childSideEffectFree =
  effectiveAllowedTools !== undefined &&
  effectiveAllowedTools.every((tool) =>
    SIDE_EFFECT_FREE_TOOLS.has(tool) || (tool === 'bash' && effectiveReadOnlyBash === true));
```

`research-agent`'s registry surface is `['Read','Grep','Glob','WebFetch','WebSearch']` (`src/skills/_agents/research-agent.ts:17`) plus `'Agent(git-investigator)'` (`src/agent/agents/builtins.ts:119`). The normalizer (`src/agent/agents/resolve.ts:70-88` → `src/agent/plugins/tool-injector.ts:81-93`; `resolve.ts:85` for the dispatch alias) maps that to `['read_file','grep','glob','web_scrape','agent']`. Both `web_scrape` and `agent` are **explicitly excluded** from `READ_ONLY_PHASE_TOOLS` — `src/agent/tool-category.ts:186` (`send_telegram`, `web_scrape` — outbound network) and `:182-184` (`agent`, `skill`, `compose` — dispatch grandchildren). So `childSideEffectFree === false`, `probe.sideEffectFree === false` (`subagent-executor.ts:602`), `canRedispatch()` short-circuits (`:445-446`), and `stream-cut-retry.ts:148` returns the cut unretried.

The mismatch is documented too: the PR body names gate (1) as `childWriteCapable === false` at `child-config.ts:204` — which *would* include `research-agent` — while the shipped gate is the stricter `:213-218`. The load-bearing safety gate described in the PR is not the one that shipped.

**Scope, precisely** (a broader claim was refuted — see dropped manifest): the retry *does* fire for `git-investigator` (`builtins.ts:146` `bashReadOnly: true`) and `Explore` (`builtins.ts:175`), and for any dispatch whose parent cage intersects down to pure-read (`child-config.ts:189-192`). It is not a no-op — it misses the primary case.

**suggestion** — Introduce a dedicated `RETRY_SAFE_TOOLS` set in `tool-category.ts` that adds `web_scrape` and `agent` to the `READ_ONLY_PHASE_TOOLS` members, and key `childSideEffectFree` on it; reusing a set whose contract is pre-approval phase gating for retry authorization couples two unrelated policies.

## Non-blocking

### 2 · `medium` · high · security
`src/agent/tools/subagent/child-config.ts:217` · ref:`b90271e` · diff-context

The predicate admits `bash` — a tool `READ_ONLY_PHASE_TOOLS` deliberately excludes at `tool-category.ts:181` — on the strength of `classifyBashCommand`, whose own header declares it "a BEST-EFFORT, default-ALLOW defense-in-depth layer — **NOT a security boundary**" and states "no allowlist or denylist can be exhaustive — a determined model could craft an obfuscated mutation this misses (e.g. `eval "$(printf ...)"`)" (`src/agent/tools/readonly-bash.ts:5-6,15-17`, file-state). A mutation the classifier missed on attempt 1 re-fires on attempt 2 under a field documented as "proven free of persistent side effects" (`child-config.ts:128`).
**suggestion** — Require literal membership in the retry set; drop the `bash` carve-out.

### 3 · `medium` · high · security
`src/agent/tools/subagent/foreground-promotion.ts:380-384` · ref:`b90271e` · file-state → verified at ref

`handle.teardown()` fires `SubagentStop` ("teardown() fires SubagentStop"), once per attempt. The `childSideEffectFree` gate constrains the **child's** tools, not the **parent's** hook config, so a non-idempotent user `SubagentStop` hook (notification POST, file append) double-fires on every rescued dispatch. The PR documents this as accepted, but frames the cost as attempt 1's `injectContext` being discarded — not as user hook code re-executing.
**suggestion** — Pass an `attempt` field in the SubagentStop payload so hooks can dedupe.

### 4 · `medium` · high · perf-observability
`src/agent/tools/subagent-executor.ts:447-451` · ref:`b90271e` · diff-context

A re-dispatch doubles a multi-minute child's wall-clock and token spend with no durable record: the only re-dispatch-specific signal is `debugLog`, which emits solely under `AFK_DEBUG=1`/`DEBUG=1` (`src/utils/debug.ts:13-14`). Absence confirmed at ref: `git grep -i redispatch` over non-test `src/**/*.ts` returns zero `emitTelemetry` and zero trace-event calls; `attempt` in `failure-payload.ts` → zero matches. The two `subagent.failed` rows an operator sees (`foreground-promotion.ts:301`) are indistinguishable from two unrelated failures.
**suggestion** — Add `attempt` as a dimension on the `subagent.failed` telemetry emit.

### 5 · `medium` · high · test-coverage
`src/agent/tools/subagent/subagent-executor.stream-cut-retry.test.ts:106-118` · ref:`b90271e` · diff-context

No test dispatches a **real registered agent** through the gate — every positive-path test hand-rolls `allowedTools: ['read_file','grep','glob']`, a set no shipped agent uses. That is why finding #1 shipped invisibly. The suite's own comment asserts the wrong mechanism: it says `readOnly` "drives `childWriteCapable === false` … and hence `probe.sideEffectFree === true`". Absence confirmed: `research-agent|git-investigator|namedAgent` → **0 matches** in both new test files.
**suggestion** — Add a `child-config.test.ts` case asserting `childSideEffectFree` for `buildChildConfig({ namedAgent: <research-agent registry entry> })`.

### 6 · `low` · high · correctness — inert mechanism, misleading invariant
`src/agent/tools/subagent-executor.ts:342-353` + `:434-440` · ref:`b90271e` · diff-context

`cancelGeneration` cannot activate in the window its comments describe. It is written only at `:353`, inside `cancelActiveForeground()`, whose **sole production call site is guarded**: `if (ctrl?.hasActiveForeground())` at `src/cli/commands/interactive/turn-handler.ts:297`, reading a map that is empty across the retry gap (`activeForegroundHandles.set` at `foreground-promotion.ts:164`, `delete` in the `finally` at `:371`). The window is already covered by a different mechanism — see the dropped manifest. Its regression test (`subagent-executor.stream-cut-retry.test.ts:179-195`) passes only because it calls `executor.cancelActiveForeground()` **directly at `:187`**, bypassing the `hasActiveForeground()` guard that is the sole production gate, so the guard validates a path production cannot reach.
**suggestion** — Either bump the counter from an unconditional `SubagentControl` method the soft-stop calls before its guard, or delete the counter and correct the comments at `:342-345`, `:350-352`, `:434-440` to name `call.signal` as the actual protection.

### 7 · `low` · high · security/observability
`src/agent/tools/subagent-executor.ts:705-709` · ref:`b90271e` · diff-context

The retry-cancel early return tears down and returns without emitting telemetry, while every sibling path does (`:712-721` fork-failure, `foreground-promotion.ts:301-325`). It also calls `handle.teardown()` without `{ deferInjectContextToCaller: true }`, unlike the normal path (`foreground-promotion.ts:384`), so a cancelled retry fork's SubagentStop note routes to the deferred queue instead of being suppressed.
**suggestion** — Emit `subagent.failed` with `status:'cancelled'` before the return, and match the teardown options of the normal path.

### 8 · `low` · high · perf-observability
`src/agent/subagent/stream-cut-retry.ts:150-153` · ref:`b90271e` · diff-context

`onRedispatch` fires *before* the settle delay and before the post-delay gate re-check, so the debug line claims a re-dispatch that `:152-153` then suppresses. The module's own tests at `stream-cut-retry.test.ts:139-153` and `:196-210` pin this ordering by using `onRedispatch` as their mutation trigger, proving the hook fires on paths that never fork.
**suggestion** — Move `opts.onRedispatch?.(attempt)` to immediately before `result = await opts.dispatch(attempt)` at `:155`.

### 9 · `low` · high · test-coverage
ref:`b90271e` · diff-context — absence confirmed at ref (0 matches for `delayMs|Promise.all|concurrent` in the integration suite)

No test covers: two concurrent `execute()` calls sharing `cancelGeneration` (the counter's stated reason for being a counter rather than a boolean, `subagent-executor.ts:438-439`); budget exhaustion with a non-zero `delayMs`, so `STREAM_CUT_REDISPATCH_DELAY_MS = 1_000` is never exercised end-to-end.

### 10 · `nit` · high · correctness
`src/agent/tools/subagent/child-config.ts:47` · ref:`b90271e` · diff-context

No test pins `SIDE_EFFECT_FREE_TOOLS ⊆ side-effect-free` (grep confirmed: the only `SIDE_EFFECT_FREE_TOOLS` reference outside `child-config.ts` is none). The set is shared with `nesting.ts:473` and `subagent.ts:287,966` for phase gating, so a future member added for a read-only *research* phase silently widens retry authorization.

## Clean reads

- **api-compat — no issues found.** `BuildChildConfigResult.childSideEffectFree` is a new required field on a *return* type; `buildChildConfig` has exactly one production importer (`subagent-executor.ts:31`), updated in this PR. Not re-exported from `src/index.ts`. `cancelActiveForeground(): Promise<number>` keeps its signature and `0`-on-empty return (`:362`); `cancelGeneration` is `private`; `execute(call)` unchanged.
- **Retry loop control flow — no bug.** Walked `stream-cut-retry.ts:132-159` at `budget` 0/1/3: `0` disables, a rescued attempt returns immediately, and the trailing `return result` correctly returns the final failure unchecked.
- **`probe` staleness — not reachable.** `probe` is a per-`execute()` local (`:433`), never on `this`, only ever assigned the deterministic `childSideEffectFree` (`:602`).
- **Worktree hazard — conclusion holds.** `sideEffectFree ⟹ !writeCapable`, so no worktree path is reachable on a retry (all four combinations enumerated) — though not for the reason the PR body gives, which keys on the looser field.
- **"Purely additive" contract holds.** `sleepWithAbort` **resolves early**, it does not reject (`src/agent/providers/shared/sleep-with-abort.ts:19-26` — `new Promise((resolve) =>` with no `reject`), so it never throws out of the wrapper.

## Dropped / corrected in verification

| claim | status | resolution |
|---|---|---|
| "ESC during the retry gap forks a fresh child anyway" (raised `high` + `medium` by two agents independently) | **REFUTED** | `turn-handler.ts:281` fires `session.interrupt()` **unconditionally, before** the guard at `:297`. Chain: `agent-session.ts:763-767` → `anthropic-direct/query.ts:621-623` `abort.requestAbort('interrupted')` → `abort-coordinator.ts:114-121` `controller.abort()` — on the same per-turn controller that owns `call.signal` (`query.ts:359` `abort.begin()` → `:428` `signal: controller.signal` → `loop.ts:1014-1019` `signal: input.signal`). `stream-cut-retry.ts:145/152` and `subagent-executor.ts:462` all suppress. Residual doc/test defect kept as finding #6 at `low`. |
| "retry never fires for **any** shipped read-only agent" | **REFUTED (narrowed)** | `git-investigator` (`builtins.ts:146`) and `Explore` (`builtins.ts:175`) both pass the predicate. Finding #1 restated as `research-agent`-specific. |
| `agent-session.ts:766` | **path-corrected** | File absent at ref; real path `src/agent/session/agent-session.ts:763-767`. Content correct — not fabricated, finding retained. |
| `sleepWithAbort` rejects on abort | **disproven** | Resolves early; no finding. |

## What was not checked

- **Citations verified inline against branch HEAD `b90271e049cdf655c1891e3ce98b3dcf9d2fc548`** via `git show <ref>:<path>` and `git grep <ref>` — every `high` citation plus every `file-state` citation at all severities. Two high claims additionally re-derived by independent shadow verifiers.
- **Stated intent: PR #796 title + body — spec-compliance assessed** (finding #1).
- Did **not** run `pnpm test`, `pnpm lint`, or coverage, so the PR's 723-files / 13,774-tests, `tsc` clean, and 100%-new-module claims are unverified.
- Did **not** review `dag.ts` or the skill/mint dispatch paths (the PR declares them out of scope).
- Did **not** evaluate the Ctrl+C/SIGINT hard-interrupt path (`turn-handler.ts:835-840`) or Telegram/daemon soft-stop equivalents — the abort analysis covers ESC/soft-stop in the REPL only.
- Did **not** enumerate user- or plugin-authored agent files, which can set `bash: read-only` (`agents/parser.ts:188`) and reach finding #2.

## Merge decision

# DO NOT MERGE

One `high` finding defeats the change's stated purpose for its primary case: `research-agent` — named in `child-config.ts:200-201` as the canonical read-only agent and mandated as `subagent_type` by this repo's own skills — is silently ineligible for the retry, because the authorization set is `READ_ONLY_PHASE_TOOLS`, whose contract is pre-approval phase gating and which excludes `web_scrape` and `agent` by design. The PR body documents a different, looser gate than the one that shipped. The fix is small and local: a dedicated retry-safety set, plus a test that runs a real registered agent through the predicate.

---

**Done**
- Reviewed PR #796 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability: 2 parallel Wave 1 agents → inline Wave 1.5 citation/absence verification at the ref → 2 shadow verifiers on the `high` claims; emitted 10 findings, 1 blocking.
- Evidence: reviewed ref `b90271e049cdf655c1891e3ce98b3dcf9d2fc548`; diff artifact `/Users/griffinlong/.afk/state/skill-preflight/0ebee9f3-404f-427b-ab67-4c3c5e25737e/pr-796.diff`; every citation re-derived via `git show b90271e:<path>` / `git grep b90271e`; verifier chains cited to `file:line` above.
- What changed in the world: **nothing** — this skill is read-only. No files written, no commits, no PR comments, no pushes. Working tree remains clean at `b90271e`. The review itself is transcript-delivered because the skill forbids file writes; the durable anchors are the ref SHA and the diff artifact path above.
- Pending: the review is not posted to the PR and no fix was applied — both are deliberately out of scope. Findings #1 and #5 are the two that gate a merge; verifying the PR's own test/lint/coverage claims requires a run I did not perform.

---

griffinwork40:
@codex fix comments

griffinwork40 (commented):

## Merge decision: **MERGE** (with one follow-up filed below)

Reviewed ref: `24ed8d14074649e5cedb4e4c277dd4dc3b02531d`. Regime: full (+818/−2, 7 files). Dimensions: security, api-compat, correctness, spec-compliance, test-coverage, perf-observability.

This is a careful change. The central safety argument in the PR body — that "zero output" is not "zero side effects" — is correct, and the implementation is *stricter* than the body describes: it does not gate on `!childWriteCapable` as written, it introduces a separate `childSideEffectFree` predicate that proves every granted tool against an audited pure-read set. That is the right call, and the body's description is the thing that is stale, not the code.

### Findings

**medium · high confidence · security · `src/agent/tools/subagent/child-config.ts:213-215` · ref:24ed8d14 · citation-type:diff-context**

`childSideEffectFree` proves the *immediate* child's tool surface but not the transitive closure, so a child holding a bare `Agent` grant is authorized for replay while its grandchildren are unconstrained.

```ts
const childSideEffectFree =
  effectiveAllowedTools !== undefined &&
  effectiveAllowedTools.every((tool) => STREAM_CUT_RETRY_SAFE_TOOL_SET.has(tool));
```

`STREAM_CUT_RETRY_SAFE_TOOLS` includes `'agent'` (`src/agent/tool-category.ts:228-232`), which is the nested-dispatch tool. The check is flat name-membership over the child's own allowlist — it never reads `resolvedAccess.nestedAgentTypes`.

The pinned builtin case is genuinely safe and I verified it rather than assuming: `research-agent` declares `Agent(git-investigator)`, which resolves to a scoped `nestedAgentTypes: ['git-investigator']`, and `git-investigator` runs under `bashReadOnly: true` (`src/agent/agents/builtins.ts:146`) propagated all the way into the child provider (`child-config.ts:430` → `nesting.ts:173-183`). So the test at `child-config.test.ts:226-229` asserting the real `research-agent` surface is replay-safe is **correct**.

The gap is a *different* path: a user-defined agent file declaring `tools: Read, Grep, Agent` — a **bare** `Agent`, no parens — resolves to unrestricted nesting (`src/agent/agents/resolve.ts:183`, `if (sawBare) return undefined; // widest grant wins: unrestricted nesting`). That leaves `nestedAgentAllowlist` unset, so the dispatch gate is skipped entirely (`src/agent/tools/subagent-executor.ts:509`, `if (nestedScope !== undefined) {`), permitting `general-purpose` — inherit-all, full write + bash — as a grandchild. Meanwhile that same agent's surface `['read_file','grep','agent']` is wholly inside the safe set, so `childSideEffectFree === true` and its prompt is replay-eligible. A replay re-runs the grandchild dispatch.

Severity is `medium`, not `high`, deliberately: it is unreachable with the shipped builtins and requires an operator-authored agent file combining a bare `Agent` grant with an otherwise pure-read surface. It is a latent sharp edge in a gate whose whole value proposition is that it is *provable*.

*Suggestion:* set `childSideEffectFree = false` whenever `effectiveAllowedTools` includes `'agent'` and `resolvedAccess?.nestedAgentTypes` is `undefined` (bare/unrestricted grant), keeping the scoped-grant case eligible.

---

**low · high confidence · perf-observability · `src/agent/tools/subagent-executor.ts:446-452` · ref:24ed8d14 · citation-type:diff-context**

A rescued stream cut is invisible in production: the only rescue-specific signal is `debugLog`, gated behind `AFK_DEBUG`.

```ts
onRedispatch: (attempt) => {
  debugLog(
    `subagent-executor: read-only child stream cut with zero output; ` +
      `re-dispatching a fresh fork (attempt ${attempt + 1})`,
  );
},
```

Combined with the PR body's accepted cost that `subagent.failed` carries no `attempt` dimension — confirmed absent from `src/agent/routing-telemetry.ts` — a rescued failure and a genuine hard failure are indistinguishable in any dashboard keyed on that event. Since the PR's own forensics depended on reading exactly this telemetry, the fix degrades the signal that would measure it.

*Suggestion:* emit a structured trace event in `onRedispatch` so `afk trace show` marks the rescue without `AFK_DEBUG=1`.

---

**nit · high confidence · spec-compliance · PR description vs `src/agent/tools/subagent/child-config.ts:213-215` · ref:24ed8d14 · citation-type:diff-context**

The PR body states gate 1 is "`childWriteCapable === false` (`child-config.ts:204`)", but the shipped gate is the stricter, separately-computed `childSideEffectFree`. `childWriteCapable` is unchanged and still serves worktree isolation. The code is better than the description; the description should be updated so a future reader does not go looking for a gate that is not there.

### Verified, not merely asserted

- **Budget arithmetic** — `budget=1` yields exactly 2 total dispatches (`stream-cut-retry.ts:138-157`), asserted at `stream-cut-retry.test.ts:101` via `1 + STREAM_CUT_MAX_REDISPATCH`.
- **"`forkSubagent` call-count is the load-bearing assertion"** — true. `subagent-executor.stream-cut-retry.test.ts:131` asserts `toHaveBeenCalledTimes(2)`, proving a genuinely fresh fork rather than a re-awaited spent handle.
- **Write-capable refusal** — exists (`:156`), plus a parametrized variant over `send_telegram`, `config_set`, `create_schedule`, `browser_act`, `bash` (`:169`).
- **Cancel-during-gap guard** — exists (`:179`), plus a mid-fork teardown variant (`:197`).
- **Anti-drift test** — genuinely imports the real `incompleteToolResultFields` producer from `./result.js` (`stream-cut-retry.test.ts:4`), not a hand-rolled fixture, and covers `TOOL_USE_LOOP_CAPPED` and `end_turn`.
- **Exclusions hold** — buffered partials resolve `status:'succeeded'` and never reach the `isError === true` predicate; `tool_use_loop_capped` is explicitly excluded (`result.ts:58`).
- **Background mode** — structurally unreachable for retry: `runBackgroundBranch` returns before the foreground path (`subagent-executor.ts:731-739`), and a background pointer never sets `isError`.
- **`web_scrape` replay safety** — the tool exposes no `method`/`body` parameter and hardcodes GET, so replay is not a remote-mutation hazard.
- **api-compat** — the only production consumer of `SubagentExecutor.execute` is `src/agent/tools/dispatcher.ts:1158`; the `execute`/`executeOnce` split preserves the public signature, and `buildChildConfig`'s added return field is additive with a single production consumer.
- **Scope** — `dag.ts`/`dag-subagent.ts` are untouched, matching the body's explicit "not in scope".

### What was not checked

- Citations were verified inline against branch HEAD `24ed8d14` by re-reading each cited line at that ref (the local checkout is on `main` and does not match; a ref export was used instead). Several line numbers proposed during review did not resolve at the ref and were re-derived by content before being included here; unresolvable claims were dropped.
- **Tests were not executed.** The body's `723 files / 13,774 tests`, `100%` new-module coverage, and clean `pnpm lint` are `unverified — cannot execute tests` in this read-only review. CI on this head is green across Lint & Build, ubuntu/macOS test matrices, PTY scrollback, docs build, and publish smoke.
- Stated intent: PR #796 title + body — spec-compliance **was** assessed against it.
- Not reviewed: runtime behavior under real provider stream cuts, the Telegram/REPL surfaces, and `SubagentStop`/`injectContext` double-fire behavior (documented as an accepted cost; no test asserts it — `injectContext` appears zero times in the new test files).

### Bottom line

The safety reasoning is sound, the exclusions are right, and the tests assert the properties that matter — including the load-bearing fork-count. None of the findings block: the security item is unreachable with shipped builtins and is a hardening follow-up, and the observability item is a gap this PR inherits rather than creates. **MERGE**, then tighten `childSideEffectFree` against bare `Agent` grants.

griffinwork40 (commented):

## Merge decision: **DO NOT MERGE** — one `high` finding, one small fix

Reviewed ref: `24ed8d1`. Dimensions: security, api-compat, correctness, spec-compliance, test-coverage, perf-observability.

This is careful work — the safety section correctly identifies that "zero output" ≠ "zero side effects", the exclusions for buffered partials and `tool_use_loop_capped` are real and tested, and `sleepWithAbort` genuinely resolves (never rejects) on abort, so the "purely additive" claim holds. The one blocking finding is that the safety gate is unsound for a *transitive* reason the PR body does not address.

---

### high · confidence high · security · `src/agent/tools/subagent/child-config.ts:213-215` · citation-type: diff-context

**`childSideEffectFree` classifies a child retry-safe while that child holds an UNSCOPED `agent` grant, so a re-dispatch double-fires a write-capable grandchild's side effects.**

`STREAM_CUT_RETRY_SAFE_TOOLS` admits the `agent` dispatch tool, and the gate only set-tests the child's *own* tool names — it never inspects what that dispatch can reach:

```ts
const childSideEffectFree =
  effectiveAllowedTools !== undefined &&
  effectiveAllowedTools.every((tool) => STREAM_CUT_RETRY_SAFE_TOOL_SET.has(tool));
```

A user-authored definition with a **bare** `Agent` token (`tools: Read, Grep, Agent`) resolves to `nestedAgentTypes: undefined`, which `subagent-executor.ts:508-509` treats as *unrestricted* (`if (nestedScope !== undefined)` — the gate is skipped entirely). That child can dispatch the builtin `general-purpose`, which has no `tools` key and therefore inherits `CHILD_ALLOWED_TOOLS` including `write_file`, `edit_file`, and `bash`.

Zero-output detection is not scoped to turn 1, and `childSideEffectFree` is computed once from the *static declared* allowlist before the first turn — never from observed behavior. So a child can dispatch a write-capable grandchild that commits files, get cut with no assistant text, and be re-forked with the original prompt, reissuing the same nested dispatch. That is precisely the corruption the PR's safety section claims to prevent, and it is a weaker proof than the `bash` exclusion this same commit justifies on the grounds that a best-effort classifier "cannot prove that replaying a command is free of side effects" — an unscoped `agent` grant delegates that question to an uncharacterized recursive dispatch.

Verified mechanically against the reviewed ref's logic (`resolve.ts`/`builtins.ts`/`nesting.ts` are byte-identical at ref and `main`): a bare `Agent` cage yields `effectiveAllowedTools: ['read_file','grep','agent']` → `childSideEffectFree: true` with `nestedAgentTypes: undefined`, while the controls behave correctly (`bash` → `false`; scoped `Agent(git-investigator)` → `true` but scoped; `Agent()` → deny-all `[]`).

**Scope correction (material):** no shipped builtin is affected. `research-agent` uses the scoped `Agent(git-investigator)` form specifically to prevent this escalation, and `general-purpose`'s own surface is `undefined`, so it correctly fails the gate for itself. This is exclusively a user/project-authored agent-definition exposure — a supported configuration, not an out-of-box hole.

**Suggestion:** when `effectiveAllowedTools` includes `'agent'`, additionally require a scoped grant — `resolvedAccess?.nestedAgentTypes !== undefined` — so a bare `Agent` token fails closed, mirroring the depth-1 gate's own reasoning.

---

### low · confidence high · test-coverage · `src/agent/tools/subagent/child-config.test.ts:208` · citation-type: diff-context

The `childSideEffectFree` negative loop covers `send_telegram`, `config_set`, `create_schedule`, `browser_act`, `bash` — but no case asserts behavior for a cage granting `agent`, which is the one entry in the safe set that carries transitive reach. The string `'agent'` appears zero times in that test file.

**Suggestion:** add a case asserting `childSideEffectFree === false` for `allowedTools: ['read_file', 'agent']` with no scoped nested grant (it currently returns `true`).

### low · confidence high · perf-observability · `src/agent/tools/subagent/foreground-promotion.ts:303-309` · citation-type: file-state

`subagent.failed` carries no `attempt` dimension (confirmed: zero `attempt` fields in any `emitTelemetry` payload at the ref), so a cut that retries and still fails writes two indistinguishable rows — double-counting exactly the failure class whose 115-event forensics justified this PR. The body documents this as accepted; flagging that it degrades the measurement used to evaluate the fix.

**Suggestion:** thread the 0-based `attempt` into the failure telemetry payload.

### nit · confidence high · spec-compliance

Body claims "29 new tests"; the ref carries 35 `it`/`it.each` blocks (43 expanded cases) across the three files. The three specifically-named tests — write-capable refusal, cancel-during-gap, and the anti-drift test against the real `incompleteToolResultFields` producer — all exist as described.

---

### Checked and clean (no defect)

- **Retry loop is flat, budget genuinely 1.** `runWithStreamCutRetry` is a `for` loop, not recursive; a retried attempt that is also cut returns unchanged. Abort and the safety gate are both re-checked *after* the settle delay.
- **Stale-probe hypothesis refuted.** The shared mutable `probe` persists across attempts, but every early return in `executeOnce` (input validation, depth ceiling, fork failure) carries no `incompleteReason`, so none can match `isZeroOutputStreamCut` and reach a retry with a stale `true`.
- **Fail-closed on `undefined` confirmed.** An uncaged/top-level dispatch leaves `effectiveAllowedTools` undefined → `childSideEffectFree: false`. `mcp__*` tokens pass through verbatim and are not set members, so any MCP-bearing agent is excluded.
- **"Not in scope" claim is accurate.** Zero matches for `runWithStreamCutRetry|STREAM_CUT|childSideEffectFree` in `dag`, `skill-executor`, `compose`, and `fork-dispatch` production paths.
- **api-compat:** `childSideEffectFree` is an additive field on `BuildChildConfigResult`; the only production importer of the new module is `subagent-executor.ts`. No breaking change.

### What was not checked

- Citations verified inline against branch HEAD `24ed8d1`; the two files that differ from `main` (`tool-category.ts`, `child-config.ts`) were read via `git show` at the ref, never from a worktree.
- Stated intent: PR #796 title + body — **spec-compliance assessed**.
- Test suite not run locally; relied on the PR's green CI (all `Lint & Build` / `Test` jobs SUCCESS at this head).
- The bypass was verified by exercising `resolve.ts` + the ref's gate logic directly, not by an end-to-end run with a real cut stream. No user-authored agent file with a bare `Agent` token was located on this machine — the exposure is latent, not observed in a live config.
- Did not review the Telegram/daemon surfaces, the `SubagentStop` hook's `injectContext` discard path, or worktree teardown under concurrent cancellation.
- Both dimension agents hit their tool-use iteration caps and returned partial results; their findings were re-verified inline at the ref and the `high` finding was independently re-derived by an adversarial verifier (all five attempted refutations failed).

griffinwork40:
Wave 1.5 verification complete — every citation and absence claim checked mechanically at the ref. Shadow-verify returned on the top claim.

---

# Code Review — PR #796

**`fix(subagent): re-dispatch a read-only child once when its stream is cut with zero output`**
`afk/subagent-stream-cut-retry` → `main` · ref `fb3949dd` · 8 files, +940/−2 · regime **full**

## Findings

### 1. `medium` · high confidence · security/correctness · `src/agent/tools/subagent/retry-safety.ts:69-78` · ref:`fb3949dd` · citation-type: diff-context

**finding:** The nested-dispatch gate proves only that a grant is *scoped*, never that the scoped grandchild types are themselves replay-safe — so the doc comment's claim to cover "the transitive closure of that reach" and return true "only when re-running this child's entire prompt is provably free of persistent side effects" is false at depth 2.

```ts
  if (effectiveAllowedTools.includes(NESTED_DISPATCH_TOOL) && nestedAgentTypes === undefined) {
    return false;
  }
  return true;
```

Two verified consequences:

- **Live today.** `research-agent` is granted `Agent(git-investigator)` (`builtins.ts:119`) and is explicitly asserted replay-safe (`child-config.test.ts:227`). `git-investigator` holds `Bash` (`git-investigator.ts:12`), gated by `bashReadOnly: true` (`builtins.ts:146`) — and `readonly-bash.ts:4-6` documents itself as *"BEST-EFFORT, default-ALLOW … NOT a security boundary."* The same module refuses exactly that surface one level up: `child-config.test.ts:218` — *"never treats best-effort read-only bash as replay-safe"* → `false`. The gate applies a standard at depth 1 that it abandons at depth 2.
- **Latent write escalation.** `Agent(general-purpose)` is a *scoped* grant, so it passes gate 2 → `childSideEffectFree === true`. The nested-dispatch gate checks membership only (`subagent-executor.ts:511`: `!nestedScope.includes(requested)`), `general-purpose` is inherit-all (`builtins.ts:157`, no `tools`), and for a top-level dispatch the grandchild is **uncaged** — `child-config.ts:390` forwards the cage (`undefined`), so `nesting.ts:177` falls back to `CHILD_ALLOWED_TOOLS` (`nesting.ts:90`, includes `write_file`/`edit_file`/`bash`). A replay re-fires those writes: the exact double-fire the module's own comment says the gate exists to prevent for the *bare* token case.

**suggestion:** Resolve each name in `nestedAgentTypes` through the registry and return `false` unless every scoped type's own effective surface also satisfies `isChildReplaySafe` — treating an unresolvable name, and any surface carrying `bash` regardless of `bashReadOnly`, as unsafe.

> **CONFLICT (severity, disclosed not auto-resolved).** Both Wave 1 agents independently assigned `high`. The adversarial shadow verifier — dispatched to refute — returned `VERDICT: CONFIRMED` on the mechanism but assigned `medium`, on the verified narrowing that **no shipped definition expresses a scoped grant naming a write-capable type**, and blast radius is bounded to one re-dispatch. I adopt `medium` (the verifier resolved the cage question neither Wave 1 agent traced), and record its condition: **raise to `high` the moment any bundled or user agent ships `Agent(<write-capable-type>)`.**

### 2. `medium` · high confidence · perf-observability · `src/agent/tools/subagent-executor.ts:702-708` · ref:`fb3949dd` · citation-type: diff-context

**finding:** The mid-fork cancel guard returns after teardown without emitting a terminal lifecycle row, while `forkSubagent` already wrote `transition: 'started'` (`subagent.ts:769-770`) — producing an unmatched `started`, the trace-integrity failure mode this repo explicitly tracks.

```ts
      if (retryCancelGeneration !== undefined && this.cancelGeneration !== retryCancelGeneration) {
        await childManager?.teardownAll();
        await handle.teardown();
        return { content: 'Agent tool call aborted', isError: true };
```

Verified: `handle.teardown()` (`handle.ts:719`) emits no lifecycle transition; the `transition: 'cancelled'` emit lives in `cancel()` at `handle.ts:696-700`, which this path does not call.

**suggestion:** Emit `emitSubagentLifecycle({ transition: 'cancelled', subagentId: handle.id, source: 'explicit' })` plus a `subagent.failed` telemetry row before returning, matching `cancel()`'s ordering.

### 3. `low` · high confidence · security · `src/agent/tool-category.ts:233-237` · ref:`fb3949dd` · citation-type: diff-context

**finding:** `web_scrape` is admitted as replay-safe, but its `search` mode issues an authenticated `POST` to Exa (`web/search.ts:56`, key at `:60`, sourced `web-scrape.ts:317`), so an automatic re-dispatch double-consumes metered third-party quota — external state that outlives the replay, undocumented in a block that documents the `bash` exclusion.

**suggestion:** Name the metered-quota double-charge in the `STREAM_CUT_RETRY_SAFE_TOOLS` comment as an accepted cost.

### 4. `low` · high confidence · test-coverage · `src/agent/tools/subagent/retry-safety.ts:69` · ref:`fb3949dd` · citation-type: diff-context

**finding:** `isChildReplaySafe` has no direct unit test and no test exercises a scoped grant naming a write-capable type — the exact hole behind finding 1.

**evidence:** `git grep -n "isChildReplaySafe|retry-safety" fb3949dd -- '*.test.ts'` → 0 matches; `general-purpose` appears in `child-config.test.ts` only inside comments (`:236`, `:299`), never as a dispatched type.

**suggestion:** Add `retry-safety.test.ts` with a case asserting `isChildReplaySafe({ effectiveAllowedTools: ['read_file','agent'], nestedAgentTypes: ['general-purpose'] }) === false`.

### 5. `low` · high confidence · perf-observability · `src/agent/subagent/stream-cut-retry.ts:132-159` · ref:`fb3949dd` · citation-type: diff-context

**finding:** No guard bounds a re-dispatch by the tool budget the dead attempt already burned, so a zero-output cut after dozens of tool rounds re-runs the whole prompt at full cost; the loop reads no tool-count signal (grep `toolCallCount|toolCalls\.length|tool_budget` in the module → 0 matches).

**suggestion:** Thread the dead attempt's tool-call count into `canRedispatch` and suppress above a threshold.

### 6. `nit` · high confidence · correctness · `src/agent/subagent/stream-cut-retry.ts:60-63` · ref:`fb3949dd` · citation-type: diff-context

**finding:** The doc asserts `tool_use_loop_capped` is "the OTHER value `incompleteReason` can hold", but `isIncompleteStopReason` admits three sentinels (`result.ts:65-70`) — `OVERLOAD_EXHAUSTED` is the third. The predicate's behavior is correct; the documentation is not.

**suggestion:** Name `OVERLOAD_EXHAUSTED` and note it resolves `succeeded`, so it never reaches the error path.

## Verified clean (not findings)

- **Loop logic correct, no off-by-one.** budget 0 → 1 dispatch; budget 1 → 2 dispatches; a final non-cut result returns via loop exit.
- **`isZeroOutputStreamCut` classifies exactly right.** Buffered partial resolves `succeeded` without `isError`; `tool_use_loop_capped` and `OVERLOAD_EXHAUSTED` are distinct sentinels; `incompleteReason` is a real `ToolResult` field.
- **No resource leak across the retry gap.** `foreground-promotion.ts:369-411` `finally` deletes both registry entries and runs `childManager.teardownAll()` / `handle.teardown()` / worktree teardown before attempt 2. Worktrees never arise on the retry path (`replay-safe ⟹ !write-capable` holds at depth 1).
- **`sleepWithAbort` only ever `resolve()`s** (`sleep-with-abort.ts:19-26`) — no structured-result→throw regression.
- **Cancellation bump precedes the early return** (`subagent-executor.ts:353` vs `:362`); `killAll()` has no production caller on the cancel path.
- **api-compat non-breaking.** `executeOnce` is `private` (`:456`) with one call site (`:443`); `buildChildConfig`'s only production importer destructures additively.
- **Background path safe** — `background-branch.ts` returns carry no `incompleteReason`, so the wrapper never re-dispatches a jobId.
- **Spec-compliance:** claims "full `executeOnce` re-entry", "`forkSubagent` call-count is the load-bearing assertion" (`…stream-cut-retry.test.ts:924` +11 more), and "budget is 1" are all **MET**. All 13 claimed new tests located and confirmed. **One unmet claim:** "it proves the **transitive reach** of a replayed prompt" — finding 1.

## What was not checked

- Citations verified **inline against branch HEAD `fb3949dd06b6bc70d9fae47e51d399f1b2860ae5`** (= checked-out HEAD, clean tree). All 20 `blocking`/`high`/`file-state` citations → `verified`; all 4 absence claims → `confirmed-absent`. **Zero fabricated, zero dropped.**
- **Stated intent: PR #796 title + body — spec-compliance assessed.**
- One citation-hygiene artifact: the security agent cited `stream-cut-retry.ts:400-407` in a *clean read*; the file is 159 lines and those are diff-artifact offsets. Underlying content verified at `stream-cut-retry.ts:145-152`. No finding depended on it.
- **Not run:** `pnpm test` / `pnpm lint` / coverage — the PR's claimed 723 files / 13,792 tests / 100% new-module coverage are **unconfirmed by this review**.
- Not enumerated: `readonly-bash.ts`'s false-negative set (the residual harm surface for finding 1's live half). Not inspected: `~/.afk/agents/` or installed plugin agent files, so I cannot say whether a write-capable scoped grant already exists in this operator's environment. Not reviewed: `dag.ts` / skill / mint dispatch paths (declared out of scope), hook behavior on the doubled `SubagentStop`, the discarded attempt-1 `injectContext`.

---

One blocking condition — **finding 1**. Mechanical severity is `medium`, not `high`, so this is a judgment call a maintainer can override: the escalation is latent-by-config and bounded to one re-dispatch. I gate it anyway because the PR's headline safety argument ("it proves the **transitive reach** of a replayed prompt, not just the child's own tool names") is verifiably false, it is asserted inside the one module whose entire value proposition is provability, the gate fails **open** where every other gate in this diff fails closed, and the correction is ~10 lines. A false safety invariant in a security-shaped comment gets cited by the next change.

Findings 2–6 are non-blocking follow-ups. Everything else in this PR verified clean.

## Done

- Ran the full-regime `/review` pipeline on PR #796: target resolution, intent capture, triage, Wave 1 (2 parallel `research-agent` dimension reviewers), Wave 1.5 inline citation + absence verification at the ref, and an adversarial shadow-verify round on the top claim.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/2c75da0b-76b8-41ca-90a8-2ee4a0828202/pr-796.diff`; all citations re-derived via `git show`/`git grep` at `fb3949dd06b6bc70d9fae47e51d399f1b2860ae5`; 6 findings (2 medium, 3 low, 1 nit) with `file:line` + verbatim evidence, 20/20 citations `verified`, 4/4 absence claims `confirmed-absent`, 0 dropped.
- World state unchanged: no files written, edited, or deleted; no git operation beyond read-only inspection; nothing posted to the PR. The review output lives in this transcript only.
- Pending: the merge decision and findings are **not** recorded on the PR — posting them requires a mutation this skill forbids. Test/lint/coverage claims in the PR body remain unverified; a maintainer wanting that gate must run `pnpm test` separately.

---

griffinwork40:
## Resolution — all review findings addressed in `8b703135`

The **DO NOT MERGE** verdict above was pinned to ref `fb3949dd`. Its blocking finding is fixed. Superseding it here.

| # | Finding | Disposition |
|---|---|---|
| 1 | **(blocking)** `isChildReplaySafe` proved a nested grant was *scoped* but never that the scoped types were replay-safe | **Fixed** |
| 2 | Mid-fork cancel guard left an unmatched lifecycle `started` row | **Fixed** |
| 3 | `web_scrape` metered-quota double-charge undocumented | **Fixed** (comment) |
| 4 | No direct unit test for the gate | **Fixed** (new `retry-safety.test.ts`, 25 cases) |
| 5 | No burned-tool-budget ceiling on re-dispatch | **Deferred**, rationale below |
| 6 | `incompleteReason` doc named 2 of 3 sentinels | **Fixed** (comment) |

### 1 — the blocking one

`Agent(general-purpose)` is a *scoped* grant, so it passed the old gate and yielded `childSideEffectFree === true`. But `general-purpose` is inherit-all, and at top level the grandchild is **uncaged**: `child-config.ts` forwards an absent cage and `nesting.ts` falls back to `CHILD_ALLOWED_TOOLS` (`write_file`/`edit_file`/`bash`). A replay re-fired those writes — precisely the double-fire the gate existed to prevent for the bare-`Agent` case. Scoping was being counted, not resolved.

Every scoped name is now **resolved through the registry** and must be a replay-safe *terminal leaf*: resolvable, not inherit-all, granting no onward dispatch, and otherwise inside the replay-safe contract.

**Why not strict recursion.** The obvious fix — require each scoped type to itself satisfy `isChildReplaySafe` — would make `research-agent` replay-unsafe, because `git-investigator` carries `Bash`. That is the PR's entire motivating case, so strict recursion would have deleted the feature to fix the hole. Instead:

- **Onward chains are refused, not recursed.** A scoped leaf that itself grants nested dispatch returns `false`. This caps the analysis at two levels and removes any need for a cycle or depth guard — a mutually-scoped pair (`a` -> `Agent(b)` -> `Agent(a)`) terminates by construction. There is a test for that.
- **A `bashReadOnly`-gated leaf is admitted at depth >= 2** under a narrow, named exception, so `research-agent` -> `git-investigator` keeps its retry. This is an *accepted, bounded decision*, not a proof — `readonly-bash.ts` documents its classifier as best-effort and explicitly not a security boundary. The exposure is bounded by the terminal-leaf rule to one audited agent's shell surface rather than an open class, and the doc comment now says exactly that instead of claiming blanket provability.

Fail-closed additions: unresolvable name, inherit-all leaf, non-terminal leaf, or no registry wired (for any non-empty scope) all yield `false`.

Behaviour delta:

| surface | before | after |
|---|---|---|
| real `research-agent` (registry wired) | `true` | `true` |
| `Agent(general-purpose)` | `true` | **`false`** |
| `Agent(<unresolvable>)` | `true` | **`false`** |
| scoped grant, no registry wired | `true` | **`false`** |
| bare `Agent` | `false` | `false` |
| `Agent()` deny-all | `true` | `true` |
| `['read_file','bash']` + `readOnlyBash` | `false` | `false` |

### 2 — trace integrity

The guard called `handle.teardown()`, which fires the stop hook but emits **no** lifecycle transition, so the `started` row `forkSubagent` had already written was left unmatched. It now calls `handle.cancel()`.

Worth recording why the naive fix was wrong: emitting the row inline from the executor would have written to `ctx.traceWriter`, which is not necessarily the writer that produced `started` (that is `config.traceWriter ?? parentTraceWriter`, and the handle's own writer is private). The row would have been silently dropped and the bug would have survived the fix. `cancel()` writes through the correct resolved writer and reuses the audited emit-before-abort-cascade ordering.

### 5 — deferred, with reasoning

`ToolResult` carries no tool-call count, so a ceiling needs either a new field on the shared `ToolResult` contract (used by every handler, for a subagent-only counter) or widening `StreamCutProbe` from a single-purpose fail-closed safety flag into a telemetry channel. Both are real contract costs for what is a cost *optimization*, not a safety property: `STREAM_CUT_MAX_REDISPATCH = 1` already bounds worst case at 2x, and each attempt is independently capped by the tool-use iteration limit and wall clock. There is also an inverted incentive — a child that burned many rounds and emitted zero text is exactly the case with nothing to salvage, i.e. where rescue is worth most.

Revisit if telemetry shows re-dispatches routinely following high-tool-count attempts, or if `STREAM_CUT_MAX_REDISPATCH` is ever raised above 1. Cheapest shape would be `probe.toolCallsBurned`, set from the resolved handle's trace and consulted in `canRedispatch`.

### Earlier Codex feedback

Both inline comments were filed against the branch's first commit `19e0ae22` and are already closed by later commits — verified, not re-solved:

- **P1 "Gate retries on every side-effecting tool"** -> closed by `b90271e0`. `isChildReplaySafe` + `STREAM_CUT_RETRY_SAFE_TOOLS` now gate the whole surface; `send_telegram`, `config_set`, `create_schedule`, `browser_act`, and `bash` are each pinned by a test case.
- **P2 "Recheck cancellation after the fresh fork"** -> closed by the mid-fork cancel guard. Finding 2 above completes it on the observability side.

### Verification

- `pnpm lint` (`tsc --noEmit`) clean.
- `pnpm test` — **751 files / 14,336 tests pass**, 0 failures.
- CI green on all 7 active checks at `8b703135`.
- `retry-safety.ts` 79 -> 193 lines (under the 350-LOC ceiling).

One caveat on the earlier review that this commit does not change: its cite for the mid-fork guard test was `990-1007`, which was a diff-artifact offset — the real test is `subagent-executor.stream-cut-retry.test.ts:197`.

griffinwork40:
All verification complete. Wave 1.5 anchored every citation at `d35e0ee3` via `git show`. Synthesizing.

---

# Code Review — PR #796

**`fix(subagent): re-dispatch a read-only child once when its stream is cut with zero output`**
`afk/subagent-stream-cut-retry` → `main` · 9 files · +1370 / −2 · reviewed ref **`d35e0ee3`**

## Process disclosure (read this first)

This review did **not** run as designed, and the report is weaker than a clean run would be:

- **Wave 1 (both dimension agents) timed out** at the 45-minute ceiling and returned nothing.
- The bounded retry wave returned **1 usable partial of 3** (`bash-depth2`); `rerun-fidelity` hit its 10-min node timeout, `test-claims` returned only preamble.
- **The working tree switched branches mid-review** (to `afk/named-agent-model-policy-default`), so the PR's files vanished from HEAD partway through. The PR ref survives as a git object, so I re-anchored every citation with `git show d35e0ee3:<path>` — Wave 1.5 as specified.

Everything below that carries a citation was **verified by me inline at the reviewed ref**. The one subagent finding I kept was independently re-derived link-by-link before inclusion. What is missing is a genuine *second* pair of eyes on security and correctness — see *What was not checked*.

## Verdict: **MERGE**

No critical or high findings. The change is well-tested, honestly documented, and fails closed at every gate I probed. The medium finding is a concrete, one-line-fixable gap in an already-disclosed accepted risk — not a merge blocker.

---

## Findings

### 1 · medium · high confidence · security · `src/agent/tools/subagent/retry-safety.ts:93-95` · ref:d35e0ee3 · citation-type:diff-context

**finding:** The depth-≥2 `bash` exception admits `git fetch` as replay-safe — a network operation that mutates local ref state — because neither enforcement layer blocks it: `readonly-bash.ts` never mentions `fetch` and defaults to allow, and `git-investigator`'s prompt forbids only `fetch --prune`.

**evidence:**
```ts
  return allowedTools.every(
    (tool) => STREAM_CUT_RETRY_SAFE_TOOL_SET.has(tool) || (tool === BASH_TOOL && bashReadOnly),
  );
```

The reachable chain, each link verified at the ref:

| # | link | evidence at `d35e0ee3` |
|---|---|---|
| 1 | `research-agent` grants a scoped nested dispatch | `builtins.ts:119` — `tools: [...researchAgent.allowedTools, 'Agent(git-investigator)']` |
| 2 | that leaf grants `bash`, read-only-gated | `builtins.ts:146` — `bashReadOnly: true`; vendored `allowedTools: ['Bash','Read','Grep','Glob']` |
| 3 | the gate admits it | `retry-safety.ts:94` (above) |
| 4 | the classifier does not block `fetch` | `readonly-bash.ts` — zero matches for `fetch`; tail is `return { mutating: false }`; docstring line 5: *"BEST-EFFORT, default-ALLOW … NOT a security boundary"* |
| 5 | the prompt does not block it either | `git-investigator.md:33` — ``- `commit`, `push`, `pull`, `fetch --prune` `` |

`research-agent` is among the most frequently dispatched child types, so this path is reachable in production, not theoretical. Impact is bounded: `git fetch` writes `.git/refs/remotes/*` and pulls objects — no data loss, no auth bypass, no user work destroyed. The code's structural disclosure at `retry-safety.ts:127-133` is honest ("an ACCEPTED, NAMED decision rather than a proof"); what is inexact is the concrete bound *"whose contract is git reads"*.

**suggestion:** Add `'fetch'` to `GIT_MUTATING_SUBCMDS` (`readonly-bash.ts:560-563`) — one token, and it closes the gap at the mechanical layer rather than the prompt layer.

---

### 2 · low · high confidence · spec-compliance · `src/agent/tools/subagent/retry-safety.ts:126-133` · ref:d35e0ee3 · citation-type:file-state

**finding:** The PR description states `bash` is excluded outright, omitting the depth-≥2 carve-out that the code implements, so a reviewer who reads only the narrative sees a stronger guarantee than the artifact provides.

**evidence:** PR body: *"`bash` is absent because its read-only classifier is best-effort and cannot prove replay safety."* The code adds an exception the body never mentions:
```
 *   - Depth ≥2 admits `bash` when the leaf is `bashReadOnly`-gated; depth 1
 *     never does. That asymmetry is an ACCEPTED, NAMED decision rather than a proof
```
Same pattern for `web_scrape`: `tool-category.ts:231-233` documents that its `search` mode POSTs to a metered third-party API and a replay double-charges quota — also absent from the PR body.

**suggestion:** Add two sentences to the PR body naming the depth-2 bash exception and the `web_scrape` quota cost, so the merge decision is made against the real contract.

---

### 3 · nit · high confidence · correctness · `src/agent/agents/builtins.ts:175` · ref:d35e0ee3 · citation-type:file-state

**finding:** The comment's *"today only `git-investigator`"* is imprecise — `Explore` (`tools: ['Read','Grep','Glob','list_directory']`, no bash, no nested dispatch) also clears `isNestedLeafReplaySafe`. Harmless in substance: `Explore` grants no shell, and no registered agent currently scopes `Agent(Explore)`, so it is unreachable as a grandchild.

**suggestion:** Reword to "today only `git-investigator` among bash-granting leaves."

---

### 4 · nit · high confidence · correctness · `src/agent/tools/subagent-executor.ts:433,602` · ref:d35e0ee3 · citation-type:diff-context

**finding:** `probe.sideEffectFree` is written once per attempt at line 602 but never reset at attempt entry, so an attempt that returns early (input validation, depth ceiling, fork failure) leaves the previous attempt's `true` in place. Not reachable in production — `STREAM_CUT_MAX_REDISPATCH` is `1`, and `maxRedispatch` is overridden only in tests, so no third dispatch exists to consume the stale value.

**evidence:**
```ts
    const probe: StreamCutProbe = { sideEffectFree: false };
    // …
    if (probe !== undefined) probe.sideEffectFree = childSideEffectFree;
```

**suggestion:** Set `probe.sideEffectFree = false` at the top of `executeOnce`, so raising the budget later cannot silently open the gate.

---

### 5 · nit · high confidence · convention · `src/agent/tools/subagent-executor.ts` · ref:d35e0ee3 · citation-type:file-state

**finding:** The file grows 688 → 777 lines against the operator's 350-LOC hard rule. Context matters: 141 of 760 non-test source files already exceed 350, there is no mechanical gate in this repo, and this PR *did* extract two new focused modules rather than inlining. Reported for completeness, not as an objection.

**suggestion:** If this file is touched again, lift the `execute`/`executeOnce` retry seam into `tools/subagent/` alongside the other extracted modules.

---

## Verified clean

| check | result |
|---|---|
| Fail-closed on `undefined` tool surface | `retry-safety.ts:139` — refused, not guessed |
| Bare `Agent` token refused | `retry-safety.ts:144` + test `retry-safety.test.ts:76` |
| Scoping alone insufficient (`Agent(general-purpose)` refused) | `retry-safety.ts:148` + tests `:103`, `:113` |
| Onward chains refused, not recursed — no cycle guard needed | `retry-safety.ts:87-92` + test `:243` *terminates on a mutually-scoped cycle* |
| `skill` / `compose` / MCP `mcp__*` tools | Absent from the safe set → fail membership → refused |
| Cancel-path completeness | `cancelActiveForeground()` is the sole foreground cancel entry (`turn-handler.ts:304`); counter bumps at `:353` **before** the empty-map early return; all other abort routes trip `call.signal`, checked at `stream-cut-retry.ts:147` and `:153`. No fail-open path found. |
| Mid-fork abort payload cannot re-trigger retry | Returns `{content, isError}` with no `incompleteReason`; predicate requires it (`stream-cut-retry.ts:68`) |
| Budget is exactly 1 re-dispatch | `stream-cut-retry.ts:29`, no prod override |

**Test-coverage verdict — all 8 PR claims CONFIRMED**, none weaker than stated:

| claim | verdict | evidence |
|---|---|---|
| C1 `forkSubagent` call-count load-bearing | CONFIRMED | `subagent-executor.stream-cut-retry.test.ts:131` — `toHaveBeenCalledTimes(2)` |
| C2 write-capable refuses | CONFIRMED | `:156-165` |
| C3 bare `Agent` refuses | CONFIRMED | `retry-safety.test.ts:76` |
| C4 scoped `Agent(type)` eligible | CONFIRMED | `retry-safety.test.ts:148` |
| C5 deny-all `Agent()` eligible | CONFIRMED | `retry-safety.test.ts:86` |
| C6 cancel-during-gap guard | CONFIRMED | `:179-193` |
| C7 mid-fork teardown guard | CONFIRMED | `:197-215`, asserts `secondHandle.cancel` called once |
| C8 anti-drift vs. real producer | CONFIRMED | `stream-cut-retry.test.ts:213-240`, uses real `incompleteToolResultFields`, incl. `end_turn` |

Beyond the claims: depth-1 bash explicitly *not* granted the exception (`retry-safety.test.ts:198`), background mode never re-forks (`:271`), ordinary hard failures never retry (`:248`).

## What was not checked

- **Citations verified inline against PR head `d35e0ee3`** using `git show`, after the working tree moved to another branch mid-review. No citation rests on the moved worktree.
- **Stated intent: PR #796 title + body — spec-compliance assessed.** S2–S7, S9, S10 confirmed implemented; the depth-2 bash carve-out is the one undisclosed deviation (Finding 2).
- **No independent second opinion on security or correctness.** Both Wave 1 agents timed out; findings 1, 3, 4, 5 are single-sourced from my own reads. Finding 1's five-link chain was re-derived link-by-link at the ref, so it is the most solid; the others are simpler and self-evident from the cited lines.
- **`/shadow-verify` not invoked** — correctly, since it triggers only on `critical`/`high` findings and the ceiling here is `medium`.
- **Tests not executed.** The PR's claim of 723 files / 13,792 passing and 100% coverage on the new module is taken on trust; I verified test *existence and assertion strength*, not green runs.
- **Not reviewed:** runtime behavior under a real stream cut, `handle.ts` internals (`:413`/`:415`/`:513-515` cited by the PR body), telemetry/trace emission shape per attempt, `SubagentStop` double-fire cost (S8), and the full `executeOnce` state-carryover audit beyond the probe — that node timed out, so state other than `probe` and the handle maps is unaudited.

---

**Done**

- Reviewed PR #796 across security, api-compat, correctness, spec-compliance, test-coverage, and convention dimensions; emitted 5 findings (1 medium, 2 low/nit, 2 nit) and a **MERGE** recommendation.
- Evidence: all citations anchored at ref `d35e0ee3` via `git show`, re-verified after a mid-review branch switch; diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/d5d86b37-80bf-49fa-a173-df4952b2e703/pr-796.diff`; findings table above is the durable artifact.
- World unchanged: read-only review, no files written, no commits, no PR mutations. The repo is on `afk/named-agent-model-policy-default` @ `baeee90e` — **not** where this session started (`afk/subagent-stream-cut-retry` @ `d35e0ee3`); I did not move it and did not move it back.
- Pending: the review ran degraded — 4 of 5 subagent dispatches timed out, so security and correctness have no independent second opinion. If that matters for the merge decision, re-run `/review 796` when dispatch is healthy, or ask for `/shadow-verify` on Finding 1 specifically.

---

PR #804, opened 2026-07-31
fix(provider): re-render the # Environment date line per turn in openai-compatible (#785)

### LINKED ISSUE #785: openai-compatible provider: system-prompt date line has the same mid-session freeze
Follow-up to #784, which fixed this for `anthropic-direct` only.

The `- Date:` line in the `# Environment` block is assembled once and frozen for the life of the session, so a session resident across local midnight reports its start date forever. #784 fixes `anthropic-direct` by calling `refreshEnvironmentDate()` (new module `providers/anthropic-direct/query/date-rollover.ts`) at the top of `composeSystem()`, gated on the rendered date string actually changing so same-day turns stay byte-identical and the prompt cache is unaffected.

The openai-compatible provider has the identical freeze:

- `src/agent/providers/openai-compatible/index.ts:326-333` builds `envFragment` via the same `formatEnvironmentFragment` once per `.query()` call
- baked into `patchedConfig.systemPrompt` at `index.ts:381`
- per-turn `resolveSystemPrompt()` (`messages.ts:191`) just re-reads that frozen string

**Fix shape:** call the exported `refreshEnvironmentDate()` from the per-turn system-prompt resolution path. The helper is pure and provider-agnostic — it takes the assembled prompt string and returns it by reference when nothing changed. Should be ~10 lines plus a test mirroring `date-rollover-system-payload.test.ts`.

**Also worth checking in the same pass:** a stale doc-comment at `openai-compatible/index.ts:178` claims the provider emits no environment block, which the Phase-2 code contradicts.

Affects the same surfaces: resident REPL / Telegram sessions on an OpenAI-compatible or local-shim model. One-shot `chat` and daemon ticks build a fresh prompt and are unaffected.

griffinwork40:
## Summary

`- Date:` in the `# Environment` block of the system prompt is frozen for a session's life in the `openai-compatible` provider — a session resident across local midnight reports its start date forever. #784 fixed the identical freeze in `anthropic-direct`; this is the same fix for its twin.

## Root cause (file:line)

- `openai-compatible/index.ts:325-333` builds the `# Environment` fragment (including the `- Date:` line) once inside `query()`, then folds it into `patchedConfig.systemPrompt` — i.e. once per session, not once per turn.
- `openai-compatible/messages.ts:191-200` (`resolveSystemPrompt`) reads that frozen `config.systemPrompt` string as-is.
- `openai-compatible/messages.ts:228-230` (`buildMessages`), called once per `runIteration()` (`query.ts:788`) — i.e. once per turn — pushed that frozen string as the `system` message on every turn, so the date line never changed after session start.

## Change

- **Relocation**: moved `refreshEnvironmentDate` from `anthropic-direct/query/date-rollover.ts` → `providers/shared/date-rollover.ts` (79 LOC) so both providers can call it without cross-importing. This mirrors the repo's existing pattern for provider-shared helpers (`plan-mode-addendum.ts`, `afk-mode-addendum.ts`, `sumProviderUsage` at `anthropic-direct/types.ts:587-595`).
- **Shim**: left a 13-LOC re-export shim at the old path (`anthropic-direct/query/date-rollover.ts`), mirroring `plan-mode-addendum.ts` verbatim, so the one non-test importer (`anthropic-direct/query.ts:60`) resolves unchanged.
- **Call site**: `openai-compatible/messages.ts:buildMessages()` now calls `refreshEnvironmentDate(sys)` on the resolved system string before pushing the system message — since `buildMessages()` runs once per turn, the date now re-renders per turn instead of reading the frozen string.
- **Comment fix**: rewrote the stale doc-comment in `openai-compatible/index.ts` (previously ~175-180) that claimed the provider "does not currently emit a `# Environment` block at all" — directly contradicted by the code a few lines below it. Now describes where the fragment is actually built and how it stays fresh.

## Prompt-cache safety

`refreshEnvironmentDate` returns its input **by reference** when the rendered date string is unchanged. So a same-day second turn's system payload is the same value (and reference) as the first turn's — no extra request-body churn, and no wasted cache invalidation on any backend/proxy that caches by prefix. Verified by a dedicated test (below) asserting `toBe` (reference equality), not just `toEqual`.

## Tests

- `providers/openai-compatible/date-rollover-system-payload.test.ts` (new, 2 tests): fake-timer integration test at the `chat.completions.create` boundary, mirroring `anthropic-direct/date-rollover-system-payload.test.ts`'s approach.
  - Across a simulated midnight (48h jump, so the local date differs on every CI OS), the second turn's system message carries the new date and differs from the first.
  - A same-day second turn sends the byte-identical (`toBe`) system string as the first — the prompt-cache-safety property.
- `providers/shared/date-rollover.test.ts` (relocated from `anthropic-direct/query/`, imports fixed, otherwise unchanged): the pure-function unit tests for the rewrite itself.

## Gates run

| Gate | Result |
|---|---|
| `pnpm lint` (`tsc --noEmit`, strict) | ✅ pass |
| `pnpm exec vitest run src/agent/providers/shared/date-rollover.test.ts src/agent/providers/openai-compatible/date-rollover-system-payload.test.ts` | ✅ 12/12 pass |
| `pnpm exec vitest run src/agent/providers/openai-compatible/` | ✅ 300/300 pass |
| `pnpm exec vitest run src/agent/providers/anthropic-direct/` | ✅ 465/465 pass (proves the shim didn't break the old import path) |
| `pnpm audit:sdk:check` | ✅ pass, lock matches current imports |

Full `pnpm test` intentionally skipped per task instructions — CI runs it.

Fixes #785. Twin of #784 (`anthropic-direct`).

### DIFF
--- a/src/agent/providers/anthropic-direct/query/date-rollover.ts
+++ b/src/agent/providers/anthropic-direct/query/date-rollover.ts
@@ -1,80 +1,13 @@
 /**
- * Mid-session date-rollover refresh for the `# Environment` block.
+ * Re-export shim — the canonical implementation has moved to
+ * `../../shared/date-rollover.ts`.
  *
- * Invariant: the `- Date:` line the model reads must name TODAY, not the day
- * the session happened to start. The assembled system prompt is built once per
- * `provider.query()` — i.e. once per session — and re-spliced only by
- * `setCwd()` (`query/cwd-dependents.ts`), so a session that outlives local
- * midnight otherwise reports its start date for the rest of its life. That hits
- * exactly the surfaces that stay resident: an overnight REPL, the long-lived
- * Telegram bot's per-chat session, and any resumed session. One-shot `chat` and
- * daemon cron ticks are unaffected — each builds a fresh prompt.
- *
- * The refresh is gated on the RENDERED DATE STRING changing, never on the clock
- * ticking: {@link refreshEnvironmentDate} returns its input by reference on
- * every turn within the same local day, so the prompt-cache breakpoint stamped
- * on the system block is invalidated at most once per midnight crossing rather
- * than once per turn. PR #268 accepted the staleness precisely to avoid a
- * per-turn cache bust; that rationale conflated "recompute the date" with "emit
- * a different date", and only the latter costs anything. `setCwd()` already
- * re-splices this same block mid-session, so a mid-session mutation here is not
- * a new class of event.
+ * This file is kept so existing imports (`./date-rollover.js` from
+ * `query.ts`) continue to resolve without modification.
  *
  * @module agent/providers/anthropic-direct/query/date-rollover
  */
 
-import { formatEnvironmentDateLine } from '../../../awareness/index.js';
-
-const ENV_ANCHOR = '# Environment\n- Working directory: ';
-const DATE_PREFIX = '- Date: ';
-
-/**
- * Return `systemPrompt` with its `# Environment` date line re-rendered for
- * `now`, or the identical string when nothing needs to change.
- *
- * Contract:
- * - Returns the input by reference when the rendered date is unchanged, when
- *   the `# Environment` anchor is absent, or when the block's shape does not
- *   match what `formatEnvironmentFragment` emits. Never throws; a prompt it
- *   cannot parse is passed through untouched, so an unrecognised shape
- *   degrades to today's (stale-date) behaviour rather than to a corrupt
- *   prompt.
- * - Rewrites exactly one line and preserves every other byte, including the
- *   trailing `- Session:` / `- Workspace:` lines and the manifest that follows.
- * - Anchors on the LAST `# Environment\n- Working directory: ` occurrence. The
- *   assembler (`query/system-prompt.ts`) always splices the real fragment
- *   second-to-last, after the operator overlay and hot memory, so a decoy block
- *   inside user-supplied text can never win.
- *
- * @param systemPrompt Assembled system prompt from `assembleSystemPrompt`.
- * @param now Clock, injectable for deterministic tests. Defaults to `new Date()`.
- * @param timeZone IANA zone; defaults to the host zone, as at assembly time.
- */
-export function refreshEnvironmentDate(
-  systemPrompt: string,
-  now?: Date,
-  timeZone?: string,
-): string {
-  const anchor = systemPrompt.lastIndexOf(ENV_ANCHOR);
-  if (anchor === -1) return systemPrompt;
-
-  // The cwd line runs to its own newline; the date line is the next one.
-  const cwdLineEnd = systemPrompt.indexOf('\n', anchor + ENV_ANCHOR.length);
-  if (cwdLineEnd === -1) return systemPrompt;
-
-  const dateLineStart = cwdLineEnd + 1;
-  if (!systemPrompt.startsWith(DATE_PREFIX, dateLineStart)) return systemPrompt;
-
-  const nextNewline = systemPrompt.indexOf('\n', dateLineStart);
-  const dateLineEnd = nextNewline === -1 ? systemPrompt.length : nextNewline;
-
-  const current = systemPrompt.slice(dateLineStart + DATE_PREFIX.length, dateLineEnd);
-  const fresh = formatEnvironmentDateLine(now ?? new Date(), timeZone);
-  if (current === fresh) return systemPrompt;
-
-  return (
-    systemPrompt.slice(0, dateLineStart + DATE_PREFIX.length) +
-    fresh +
-    systemPrompt.slice(dateLineEnd)
-  );
-}
+export {
+  refreshEnvironmentDate,
+} from '../../shared/date-rollover.js';
--- a/src/agent/providers/openai-compatible/date-rollover-system-payload.test.ts
+++ b/src/agent/providers/openai-compatible/date-rollover-system-payload.test.ts
@@ -1,0 +1,144 @@
+/**
+ * Integration test for mid-session date rollover in the openai-compatible
+ * provider's system payload — the twin of
+ * `anthropic-direct/date-rollover-system-payload.test.ts` for issue #785.
+ *
+ * `OpenAICompatibleProvider.query()` assembles the `# Environment` fragment
+ * (and the rest of the system prompt) ONCE per `query()` call, exactly like
+ * anthropic-direct's `userSystem`. Before the fix, `buildMessages()`
+ * (messages.ts) read that frozen string on every turn via
+ * `resolveSystemPrompt()`, so a session resident across local midnight kept
+ * sending yesterday's `- Date:` line for the rest of its life. The fix calls
+ * `refreshEnvironmentDate()` on the resolved system string inside
+ * `buildMessages()`, which runs once per `runIteration()` — i.e. once per turn.
+ *
+ * Asserted at the `chat.completions.create` boundary — the closest observable
+ * point to what the model actually sees. Only `Date` is faked; timers stay
+ * real so the streaming machinery is untouched. The two instants are 48h
+ * apart so the local date differs in EVERY host timezone, keeping the
+ * assertion valid on the ubuntu/macos/windows CI legs.
+ */
+
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import type OpenAI from 'openai';
+import type { AgentConfig } from '../../types/config-types.js';
+import type { ProviderUserTurn } from '../../provider.js';
+import { __setOpenAIClientFactory, type OpenAIClientFactory } from './query.js';
+import { OpenAICompatibleProvider } from './index.js';
+import type { OpenAIChunk } from './translate.js';
+
+const createCalls: Array<{ args: { messages: Array<{ role: string; content: string }> } }> = [];
+let callCount = 0;
+let rollAt: Date | null = null;
+
+function makeTextStream(text: string): OpenAIChunk[] {
+  return [
+    { choices: [{ delta: { content: text } }] },
+    {
+      choices: [{ delta: {}, finish_reason: 'stop' }],
+      usage: { prompt_tokens: 5, completion_tokens: 1, total_tokens: 6 },
+    },
+  ];
+}
+
+/** Roll the fake clock to `rollAt` right after the FIRST `create()` call is
+ * recorded — mirrors `rollClockAfterFirstCall` in the anthropic-direct sibling
+ * test, so turn 2's `buildMessages()` runs on the far side of the rollover. */
+function installClockRollingClient(): void {
+  const factory: OpenAIClientFactory = () =>
+    ({
+      chat: {
+        completions: {
+          create: async (args: { stream?: boolean }) => {
+            if (!args.stream) throw new Error('mock only supports streaming mode');
+            createCalls.push({ args: args as never });
+            callCount += 1;
+            if (callCount === 1 && rollAt !== null) vi.setSystemTime(rollAt);
+            const chunks = makeTextStream('ok');
+            return (async function* () {
+              for (const c of chunks) yield c;
+            })();
+          },
+        },
+      },
+    }) as unknown as OpenAI;
+  __setOpenAIClientFactory(factory);
+}
+
+async function* twoTurns(): AsyncIterable<ProviderUserTurn> {
+  yield { content: 'first' };
+  yield { content: 'second' };
+}
+
+async function drain(query: AsyncIterable<unknown>): Promise<void> {
+  for await (const _ev of query) {
+    void _ev;
+  }
+}
+
+function runTwoTurns(): AsyncIterable<unknown> {
+  const provider = new OpenAICompatibleProvider();
+  return provider.query({
+    prompt: twoTurns(),
+    config: { model: 'gpt-4o-mini', apiKey: 'sk-test-key' } as AgentConfig,
+  }) as AsyncIterable<unknown>;
+}
+
+function systemTextOfCall(i: number): string {
+  return createCalls[i]?.args.messages[0]?.content ?? '';
+}
+
+function dateLineOf(systemText: string): string | undefined {
+  return systemText.split('\n').find((l) => l.startsWith('- Date: '));
+}
+
+const DAY_ONE = new Date('2026-07-30T12:00:00Z');
+const DAY_THREE = new Date('2026-08-01T12:00:00Z'); // +48h: a different local date everywhere
+
+describe('OpenAICompatibleQuery — mid-session date rollover (#785)', () => {
+  beforeEach(() => {
+    vi.useFakeTimers({ toFake: ['Date'] });
+    vi.setSystemTime(DAY_ONE);
+    createCalls.length = 0;
+    callCount = 0;
+    rollAt = null;
+    __setOpenAIClientFactory(null);
+    installClockRollingClient();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+    __setOpenAIClientFactory(null);
+  });
+
+  it('re-renders the Date line on the turn after the local day changes', async () => {
+    rollAt = DAY_THREE;
+    await drain(runTwoTurns());
+
+    expect(createCalls.length).toBeGreaterThanOrEqual(2);
+    const first = dateLineOf(systemTextOfCall(0));
+    const second = dateLineOf(systemTextOfCall(1));
+
+    expect(first).toMatch(/^- Date: \w+, \d{4}-\d{2}-\d{2} \(.+\)$/);
+    expect(second).toMatch(/^- Date: \w+, \d{4}-\d{2}-\d{2} \(.+\)$/);
+    expect(second).not.toBe(first);
+  });
+
+  // Cache economics: the same-day payload must be byte-identical — ideally
+  // the same string reference — which is what proves the fix does not churn
+  // the request payload (and, on backends that cache by prefix, the
+  // prompt-cache breakpoint) on every turn. `refreshEnvironmentDate` returns
+  // its input by reference when the rendered date is unchanged, and
+  // `resolveSystemPrompt` reads the same frozen `config.systemPrompt` string
+  // every turn, so the `content` sent on turn 2 is the same value (and
+  // reference, for the underlying string) as the one sent on turn 1.
+  it('leaves the system payload untouched (same reference) across turns within the same day', async () => {
+    rollAt = null;
+    await drain(runTwoTurns());
+
+    expect(createCalls.length).toBeGreaterThanOrEqual(2);
+    const first = systemTextOfCall(0);
+    const second = systemTextOfCall(1);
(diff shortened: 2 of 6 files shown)


griffinwork40 (commented):

## Merge decision: **MERGE**

No blocking findings. The core claim of the PR holds end-to-end on the production path, verified by reading the post-merge tree at the reviewed ref rather than trusting the description.

**Reviewed ref:** `a0fe0a8a3230e651954392c0cc1dff686d219adf` · 6 files, +249/−82 · regime: full (2 dimension agents + inline citation verification)

---

### Core claim: verified

The per-turn refresh genuinely reaches the wire. The chain:

1. `openai-compatible/index.ts:379-383` — `systemParts.push(envFragment)` then `systemPrompt: systemParts.join('\n\n')`, so the assembled string really does contain a live `# Environment\n- Working directory: …\n- Date: …` block.
2. `query.ts:1284-1289` — `buildQueryFromConfig` puts that same patched `config` into `opts.config`.
3. `query.ts:788` — `const messages = buildMessages({ config: this.opts.config, … })` inside `runIteration()`, i.e. once per turn.
4. `messages.ts:238` — `messages.push({ role: 'system', content: refreshEnvironmentDate(sys) })`.
5. Both wire branches consume that same `messages` array (`query.ts:860` responses / `query.ts:884` chat-completions), so the Responses path inherits the fix via `responses-messages.ts:153`.

This was the one way the PR could have been a no-op (if `resolveSystemPrompt` had returned only the operator prompt with no `# Environment` block). It does not — the premise checks out.

### Test quality: real regression guard

`date-rollover-system-payload.test.ts` drives the real `provider.query()`; `__setOpenAIClientFactory` swaps only the OpenAI SDK client, so the production `buildMessages` still runs. Pre-fix, `messages.push({ role: 'system', content: sys })` (visible as the removed line in the diff) pushed the frozen string every turn, so test 1's `expect(second).not.toBe(first)` would fail against pre-fix code. The system message is reliably `messages[0]` because `buildMessages` pushes `sys` before `resumeHistory`/`priorTurns` and the test passes neither.

---

### Findings

**1 · nit · confidence high · docs-accuracy · PR body "Prompt-cache safety" vs `date-rollover-system-payload.test.ts:142`**
The body states the test asserts `toBe` (**reference equality**), not just `toEqual`. For string *primitives* JS `toBe` (SameValue) cannot distinguish "same reference" from "recomputed to an equal value" — the test proves the payload is **byte-identical**, which is the property that actually matters for a prefix cache, but it is not a proof of reference identity.
```js
expect(second).toBe(first);
```
*Suggestion:* reword the PR body to "byte-identical value" rather than "reference equality". No code change.

**2 · nit · confidence high · consistency · `messages.ts:238` vs `anthropic-direct/query.ts:599`**
The two call sites scope the rewrite differently: anthropic-direct passes only the operator portion (`this.state.userSystem`), while openai-compatible passes the **full** joined prompt (toolBase + overlay + memory + env + manifest). Behaviourally equivalent here — `refreshEnvironmentDate` anchors on the *last* `# Environment\n- Working directory: ` and rewrites exactly one `- Date:` line — but the scanned input is materially larger on this provider.
*Suggestion:* none required; noting the asymmetry for whoever next touches either call site.

---

### Checked and cleared (explicitly, not by omission)

- **api-compat / the shim.** The pre-move module exported exactly one symbol; `ENV_ANCHOR` and `DATE_PREFIX` were module-private consts. The shim re-exports `refreshEnvironmentDate` 1:1, so its export surface is complete. Whole-tree grep at the ref finds exactly three importers — `anthropic-direct/query.ts:60` (via shim), `openai-compatible/messages.ts:15` (direct), `shared/date-rollover.test.ts:2` — all resolving. `package.json` exposes only `"."` and `"./package.json"`, no deep subpath exports, so this internal path was never reachable by npm consumers. **Not a breaking change.**
- **Prompt-injection via the anchor.** A sub-agent flagged that `buildSkillManifest` output is the only part joined *after* `envFragment` and so could in principle post-anchor a decoy block. Verified directly against the ref and **dropped**: `src/agent/tools/skill-bridge.ts` emits no `# Environment` or `- Working directory: ` text (0 grep matches), and the operator overlay + hot memory are joined *before* the fragment (`index.ts:376-379`), so neither can win the `lastIndexOf`. `formatEnvironmentFragment` additionally strips CR/LF from `cwd`. Impact ceiling even on a hypothetical decoy is one rewritten `- Date:` line — no secret, auth, or data-loss path.
- **Cache-safety under plan / autonomous mode.** A sub-agent raised this as a `medium`: `query.ts:805-808` rebuilds `messages[0].content` as `content + '\n\n' + addendum` every turn, allegedly defeating the by-reference return. **Refuted on re-derivation:** `PLAN_MODE_ADDENDUM_TEXT` and `AFK_MODE_ADDENDUM_TEXT` are static constants with zero interpolation, so turn 2 concatenates identical operands and yields a byte-identical payload. A prefix cache keys on serialized bytes, not JS object identity — no extra invalidation. The cache-safety claim holds in **all** permission modes.
- **Correctness of the rewrite.** Only `[dateLineStart, dateLineEnd)` is replaced and the result is reassembled via `slice(0,…) + fresh + slice(dateLineEnd)`, so `- Session:` / `- Workspace:` and the trailing manifest survive byte-for-byte. When `systemPrompt` is a preset object, `resolveSystemPrompt` yields the `.append` text with no `# Environment` anchor, and the function returns its input unchanged — a documented no-op, not a bug.
- **Test layering.** `shared/date-rollover.test.ts` importing from `../anthropic-direct/query/system-prompt.js` is precedented: 5 other `shared/*.test.ts` files already reach into `anthropic-direct` (`abort-coordinator`, `presence-advertise`, `presence-signals`, `stream-stall-timeout`, `tool-call-trace`).
- **Perf.** One `lastIndexOf` + two `indexOf` + 2–3 `slice`s over the system string per turn, and it returns before any slice on the common same-day path. Not material.

### Out of scope (stated, not a gap)

`oneshot.ts:146` and `oneshot.ts:223` push their own `{ role: 'system', content: system }` without refreshing, reached via `index.ts:661` (`complete()`). The module doc-comment at `shared/date-rollover.ts:10` explicitly scopes this fix to resident sessions — "One-shot `chat` and daemon cron ticks are unaffected — each builds a fresh prompt." Consistent with the stated intent; not an unmet clause.

### What was not checked

- Citations verified inline against branch HEAD `a0fe0a8a`; every `file-state` citation in this review was re-read at that ref, and two sub-agent findings were dropped/refuted on that basis (one false-absent, one refuted inference).
- Stated intent: PR #804 title + body — spec-compliance **was** assessed against it.
- Tests were **not executed** (read-only review); the "test 1 fails pre-fix" conclusion is derived from the diff's removed line plus the code path, not from a run. CI is green on all legs.
- Did not review the anthropic-direct sibling test byte-for-byte, nor the live behaviour of a real midnight crossing against a real backend.

griffinwork40 (commented):

## Merge decision: **MERGE**

No blocking findings. The change achieves its stated intent (#785) on every payload path I could reach, and the module relocation is export-surface-neutral. Two non-blocking items below.

---

### Verified against the reviewed ref (`a0fe0a8`)

The load-bearing hypotheses for a fix of this shape were each checked and **refuted** — recording them so a human doesn't re-spend the effort:

1. **"The Responses-API wire is a second, unpatched system-payload path."** Refuted. `query/request-body.ts:50` calls `buildResponsesRequest(args.messages, …)`, and `responses-messages.ts:153` hoists the system message into top-level `instructions`. Both wires consume `buildMessages()` output, so both inherit the refresh. The single hook is sufficient.
2. **"`buildMessages()` is called once per session, not per turn."** Refuted. Called at `openai-compatible/query.ts:788`, inside `runIteration()` — genuinely once per turn, as the PR body claims.
3. **"A decoy `# Environment` block in operator-supplied text wins the `lastIndexOf` anchor."** Refuted for the assembly path. `openai-compatible/index.ts:376-381` splices `[toolBase, existingSys?, memoryPrompt, hotMemory?, envFragment, manifest?]` — the real fragment sits second-to-last, *after* the operator overlay and hot memory. The anchoring doc-comment's justification, written for anthropic-direct, holds verbatim here.
4. **"The addendum splice reintroduces staleness or feeds corrupted input back into the refresh."** Refuted. The plan/AFK-mode addendum (`query.ts:799-809`) mutates a local `messages[0]` that `buildMessages()` rebuilds every iteration; it is never written back to `config`. The refresh always sees the pristine `config.systemPrompt`. Neither `PLAN_MODE_ADDENDUM_TEXT` nor `AFK_MODE_ADDENDUM_TEXT` contains the `# Environment\n- Working directory: ` anchor (grep: zero matches), so even a hypothetical feedback path would be inert.
5. **Export-surface parity.** The pre-image exported exactly one symbol (`refreshEnvironmentDate`); the 13-LOC shim re-exports exactly that one. No dropped export, no stale importer — `anthropic-direct/query.ts:60` resolves unchanged.
6. **Import path.** `../../awareness/index.js` is correct from `src/agent/providers/shared/`.
7. **Relocated test fidelity.** `shared/date-rollover.test.ts` retains all 10 original `it(...)` cases; only import lines changed. Cross-layer test imports from `shared/*.test.ts` into `anthropic-direct/` are pre-existing practice (`abort-coordinator.test.ts`, `presence-*.test.ts`), so the new `../anthropic-direct/query/system-prompt.js` import is not a novel layering violation.
8. **Degenerate inputs.** Every branch of `refreshEnvironmentDate` either returns the input by reference or rebuilds from offsets already validated by a preceding `-1`/`startsWith` check. A caller supplying a system prompt with no `# Environment` block degrades to identity via `anchor === -1`; `resolveSystemPrompt` collapses `''` to `undefined`, and the `sys !== undefined` guard means a session with no system prompt is untouched. No corruption path found.

---

### Findings

**1 · nit · confidence: high · test-coverage / doc-accuracy**
`src/agent/providers/openai-compatible/date-rollover-system-payload.test.ts:127-134` · ref: `a0fe0a8` · citation-type: file-state (verified at ref)

**Finding:** The test's comment claims the assertion proves the payload is "the same value (and reference, for the underlying string)", but `toBe` uses `Object.is`, and for primitive strings ECMAScript equality is defined purely over the code-unit sequence — there is no JS-observable reference identity for string primitives distinct from value identity. The test establishes byte-identical value equality, which is exactly the property that matters for payload and prompt-cache stability; the parenthetical implies a separately-verified stronger fact that does not exist at the language level.

```
  // its input by reference when the rendered date is unchanged, and
  // `resolveSystemPrompt` reads the same frozen `config.systemPrompt` string
  // every turn, so the `content` sent on turn 2 is the same value (and
  // reference, for the underlying string) as the one sent on turn 1.
```

**Suggestion:** Drop the "(and reference, for the underlying string)" clause and state the claim as byte/value identity only. The same overclaim appears in the PR description's "Prompt-cache safety" section and in the `it(...)` title `'leaves the system payload untouched (same reference) …'`. The implementation's by-reference return is real and worth documenting in `shared/date-rollover.ts` — it just isn't what this assertion proves.

**2 · low · confidence: medium · perf-observability**
`src/agent/providers/shared/date-rollover.ts` (the `current !== fresh` rewrite branch) · ref: `a0fe0a8` · citation-type: diff-context

**Finding:** No log or telemetry is emitted when a rollover actually fires (grep for logging call sites in the module body: zero matches, excluding prose). An operator debugging a reported stale-date symptom has no signal distinguishing "the rewrite fired this turn" from "the date was already fresh" from "the anchor did not match and the prompt was passed through untouched" — and that third case is the silent-degradation path the function's own contract documents.

**Suggestion:** Emit a debug-level log on the `current !== fresh` branch naming the old and new date strings.

**Not a finding — recorded for completeness:** `expect(createCalls.length).toBeGreaterThanOrEqual(2)` is a sanity floor guarding the subsequent indexing, not the load-bearing assertion; the `toBe` comparison is pairwise on calls 0 and 1 specifically, so additional calls would not mask a regression in what this test names.

**Perf:** A small fixed number of `lastIndexOf`/`indexOf`/`slice` passes over a tens-of-KB string is sub-millisecond — orders of magnitude below per-turn network latency. No finding warranted.

---

### Spec-compliance

**Stated intent:** PR #804 title + body — spec-compliance assessed. Every clause of the stated intent has an implementing change: relocation (`shared/date-rollover.ts`), shim (`anthropic-direct/query/date-rollover.ts`), per-turn call site (`messages.ts:238`), doc-comment fix (`index.ts:173-181`). No unmet requirement found. No scope creep found — the doc-comment rewrite is explicitly called for by the intent, and the stale comment it replaces was directly contradicted by the code below it.

Parity with the anthropic-direct twin holds structurally, with a benign implementation delta: anthropic-direct refreshes in `composeSystem()` by mutating `this.state.userSystem` in place; openai-compatible refreshes functionally at push time. Both are per-turn; the functional form is the safer of the two.

---

### What was not checked

- **Citations verified inline against branch HEAD `a0fe0a8`.** Every `file-state` citation in this review was re-read at that ref; every absence claim ("no logging call", "no `# Environment` in the addendum texts") was re-verified by grep at that ref. Nothing was cited from the `main` working tree.
- **Tests were not executed.** The PR body's green-gate table (lint, 12/12 targeted, 300/300 openai-compatible, 465/465 anthropic-direct, `audit:sdk:check`) is taken on trust; CI on this PR is green across all legs.
- Did not review the Telegram or REPL surfaces end-to-end, nor validate real-world behavior across an actual midnight crossing — the fake-timer test is the only rollover evidence.
- Did not assess `oneshot.ts`'s independently-constructed system messages (`oneshot.ts:146,223`); those serve compaction summarization with a caller-supplied `system` string, not the session `# Environment` block, so they are out of scope for #785.
- **Process disclosure:** the two planned parallel dimension agents both exceeded their runtime ceiling and returned nothing. Their work was re-performed by the orchestrator inline against the reviewed ref (which is where the ref-anchored verification belongs regardless), plus one bounded adversarial agent that returned a partial result before hitting its iteration cap. Its partial findings were fully re-verified above rather than trusted. Coverage of the security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability dimensions is complete; the reduced agent count changed who did the reading, not what was read.

griffinwork40 (commented):

## Merge decision: **MERGE**

No blocking findings. The fix is correct, the load-bearing assumption holds, and the two surviving items are doc-precision nits that do not affect behavior.

**Reviewed ref:** `a0fe0a8a3230e651954392c0cc1dff686d219adf` · 6 files, +249/−82 · regime: full (2 dimension agents + inline citation/absence verification at the ref)

---

### What was verified (not just asserted)

**The load-bearing assumption holds.** The whole PR depends on `refreshEnvironmentDate` finding its anchor in the *openai-compatible*-assembled prompt — a different assembler than the one the function was written for. Verified at the ref: `awareness/runtime-snapshot.ts` emits `# Environment\n- Working directory: <cwd>\n- Date: …`, which matches `ENV_ANCHOR` + `DATE_PREFIX` exactly. The rewrite is not a silent no-op.

**The per-turn claim is real, and stronger than stated.** `buildMessages()` is called at `query.ts:788` inside `runIteration()`, which is driven by the `for (;;)` tool-loop at `query.ts:559`/`:566`, under the per-turn `runTurn` at `query.ts:386`. The refresh therefore fires at least once per turn. Safe, because the function is reference-preserving when the date is unchanged.

**The Responses API path is covered — no unfixed twin.** `responses-messages.ts:buildResponsesRequest` (called from `query/request-body.ts:18`) consumes the array `buildMessages` produced and hoists `role:'system'` into `instructions` (`responses-messages.ts:108-153`), so the per-turn refresh propagates. This was the highest-risk half-fix scenario; it does not apply.

**Shim is export-complete.** The pre-change module exported exactly one symbol (`refreshEnvironmentDate`); the shim re-exports exactly that. The sole production importer, `anthropic-direct/query.ts:60`, resolves unchanged. The PR's stated precedent is accurate: `anthropic-direct/plan-mode-addendum.ts` is a genuine re-export shim over `shared/plan-mode-addendum.ts` — same shape.

**Claims investigated and refuted (reported so they are not re-litigated):**
- *"`oneshot.ts` leaves a frozen-date path unfixed"* — refuted. Both sites (`oneshot.ts:146`, `:223`) receive the static `COMPACT_SYSTEM_PROMPT` (`query.ts:1032`, `:1079`), never the assembled `# Environment` string. The staleness class does not apply.
- *"`lastIndexOf` anchoring lets a skill-manifest decoy win"* — refuted on feasibility. The manifest is appended after `envFragment` (`index.ts:375-380`), so it is structurally last; but `parseSkillMd` (`cli/slash/_lib/flag-harvest.ts:98-101`) parses frontmatter with a single-line regex and no block-scalar support, so a literal newline cannot enter a skill `description`, and the decoy cannot be formed. Worth noting this ordering is **pre-existing** (identical on `main` at `index.ts:377-378`, and mirrored in `anthropic-direct/query/system-prompt.ts:105-106`) — not introduced by this PR, and not a reason to hold it.

---

### Findings

**1 · nit · high confidence · spec-compliance — `openai-compatible/messages.ts:283-286`** (citation-type: diff-context, ref `a0fe0a8`)
The new comment states `buildMessages()` runs "once per `runIteration()` — i.e. once per turn", but `runIteration()` executes once per *tool-round*, and one turn contains many rounds (`query.ts:559-684`), so the refresh fires more often than the comment says — harmlessly, since the reference-preserving return makes repeat calls free.
```
// Invariant: buildMessages() runs once per runIteration() — i.e. once per
// turn — so re-rendering the `# Environment` date line here (rather than
```
*Suggestion:* say "once per model call (iteration)" instead of "once per turn".

**2 · low · high confidence · test-coverage — `openai-compatible/date-rollover-system-payload.test.ts:225-232,240`** (citation-type: diff-context, ref `a0fe0a8`)
The comment and the PR body claim `toBe` proves the payload is the *same reference*, but JS string primitives compare by value under `Object.is`, so the assertion proves byte-equality only — which is exactly what the prompt-cache property needs. The test is correct; the "reference equality" framing overclaims something the assertion cannot observe.
```
// reference, for the underlying string) as the one sent on turn 1.
expect(second).toBe(first);
```
*Suggestion:* reword the comment (and the PR body's "Verified by a dedicated test … asserting `toBe` (reference equality)") to claim byte/value equality.

Neither finding blocks the merge; both are single-line comment edits.

---

### What was not checked
- Citations verified inline against PR head `a0fe0a8a3230e651954392c0cc1dff686d219adf` via `git show <ref>:<path>`; the local checkout sits on `main` (`86b7252`), so every quoted line was re-read at the reviewed ref rather than from the working tree.
- Stated intent: PR #804 title + body — **spec-compliance was assessed** against it. No scope creep found; relocation, shim, call-site wiring, and comment fix each map to a stated intent item.
- Tests were **not executed** — the PR's claimed gates (lint, 12/12 targeted, 300/300 openai-compatible, 465/465 anthropic-direct, `audit:sdk:check`) were taken as reported, and CI is green on all legs.
- Not reviewed: runtime behavior against a live model endpoint; timer interaction between `vi.useFakeTimers({toFake:['Date']})` and the streaming machinery beyond the test's own design; the Telegram/daemon surfaces named in the module doc-comment as the motivating resident sessions.

PR #805, opened 2026-07-31
fix(permissions): carve ssh config + known_hosts out of read-denylist (#579 O2)

griffinwork40:
## Summary

Closes the **O2** half of #579 — read-denylist whole-dir floors block legitimate non-secret siblings even in unconfined sessions.

The whole-dir read floor for the operator's ssh directory **stays** — private keys have arbitrary names (`github_key`, `work_id`), so a key-name glob (`id_*`/`*_rsa`/`*.pem`) would **fail-open** on unknown siblings (independently verified: the repo's own `worktree-ignored-patterns.ts:44-45` and `at-file-inject.ts:91` already miss `github_key`). But two well-known NON-secret siblings are needed for git/ssh host-alias work and were blocked even in unconfined sessions:

- **`config`** — host aliases; carries `IdentityFile` PATHS, not key material
- **`known_hosts`** — the trusted-host list

## Changes

Carve both out as **EXACT files** via `READ_ALLOWLIST_REL` (mirrors the `mcp.json` precedent from #728):
- Siblings (`id_*`, `config.bak`, `known_hosts.old`, `github_key`) stay denied
- Re-deniable via `AFK_READ_DENYLIST` (extras outrank the exception)
- Leaf-un-deref'd for symlink safety (a `config → id_rsa` symlink resolves to the key path which isn't in the allowlist → stays denied)
- The `@`-file injector (`at-file-inject.ts:223`) and bash-restriction-hook (`:387`) share `READ_ALLOWLIST_REL` and pick up the carve-outs automatically — the bash hook will now scrub `config` references too, closing the typed-tool-vs-bash divergence

**`afk.config.json` is NOT carved out** — model-slot `apiKey` fields (`models.{slot}.apiKey`, `cli/config/types.ts:307`) are hand-editable into that file and read by the loader (`model-slots.ts:424`); `config_get target=config` does **not** mask (`mutate.ts:207-212` masks only `target=env`), so no safe masked-read escape hatch exists. Stays whole-dir floored.

## Verification

- `read-denylist.test.ts`: 62/62 pass — new carve-out tests mirror the `mcp.json` pattern (exact-file, siblings denied, re-deniable, assertNotReadDenied)
- `at-file-inject.test.ts`: 36/36 pass (shared list propagates)
- `bash-restriction-hook.test.ts`: 75/75 pass (shared list propagates)
[...]

Issue #579, opened 2026-07-13
Over-restriction: read-denylist whole-dir floors (O2) + AFK-mode rm blanket-block (O3)


griffinwork40:
## Summary

Two places where a sandbox control blocks **legitimate** work (the over-restriction lens of the 2026-07-13 audit). O1 is fixed by #574; these are the remaining two. Both are cleanly fixable and one is internally inconsistent with the codebase's own choices.

---

### O2 — credential read-denylist whole-dir floors block non-secret siblings, with no escape

`src/agent/tools/handlers/read-denylist.ts:47-55` floors **whole directories** — `~/.ssh`, `~/.aws`, `~/.gnupg`, `~/.config/gcloud`, `~/.afk/config` — and the floor runs *before* the allowAll/unconfined short-circuit (`_cwd-utils.ts:109-117`), so even when the user explicitly asks in an unconfined session the typed read tools cannot read:
- `~/.ssh/config`, `~/.ssh/known_hosts` (host aliases — **not** key material),
- `~/.afk/config/afk.config.json` (non-secret settings; only `afk.env`/`mcp.json` are secret).

And `AFK_READ_DENYLIST` is **extend-only** — `getReadDenylist` (`read-denylist.ts:87-96`) has no way to *remove* a built-in (`:44` "intentionally no way to remove a built-in"), so there's no opt-out.

**The fix is already precedented in the same file:** the authors deliberately made `.kube/config` and `gh/hosts.yml` **file-level** "so ordinary reads of sibling non-secret config still work" (`read-denylist.ts:61-63`). Apply that principle to the two offenders:
- `~/.ssh` → floor `~/.ssh/id_*`, `~/.ssh/*_rsa`, `~/.ssh/*.pem`, `~/.ssh/*.key` (not the whole dir).
- `~/.afk/config` → floor `~/.afk/config/afk.env` + `~/.afk/config/mcp.json` (not the whole tree; `afk.config.json` is non-secret and has `config_get`).
- Leave `~/.aws`, `~/.gnupg`, `~/.config/gcloud` whole-dir (genuinely all-secret).

**Acceptance:** `~/.ssh/config`, `~/.ssh/known_hosts`, `~/.afk/config/afk.config.json` readable; `~/.ssh/id_*`, `~/.aws/credentials`, `afk.env`, `mcp.json` still blocked. Tests cover both.

---

[...]

PR #806, opened 2026-07-31
fix(permissions): allow curated in-workspace rm -rf in AFK autonomous mode (#579 O3)

griffinwork40:
## Summary

Closes the **O3** half of #579 — AFK autonomous mode blanket-blocks `rm -rf`, stricter than the repo's own destructive-command detector. A headless `rm -rf node_modules && pnpm install` stalls on an approval prompt nobody answers.

## The fix

A **narrow gate-side carve-out** — NO change to `BASH_HIGH` (the classifier stays pure, no classifier test flips). After `classifyRisk` returns `high` for a bash command, the gate checks whether it's a plain `rm` whose **every** target:

1. **Basename** is in a curated leaf-dir set: `node_modules`, `dist`, `build`, `.next`, `coverage`, `.cache`, `__pycache__`, `.turbo`, `.parcel-cache`, `out`, `target`
2. **Resolves strictly inside** the workspace root via `safeRealpath` (symlink-deref'd) — NOT the workspace root itself
3. **Carries no shell metacharacters** (`*`, `?`, `$`, backticks, `()`, `{}`, `;`, `|`, `&`, `<`, `>`, quotes, backslash)

If all three pass → allow. Otherwise → stays `high` → blocked/prompted as before. **Fails closed** on any ambiguity (sudo prefix, shell operators, unknown workspace root, unparseable command).

## Why a curated allowlist (not "anything in-workspace")

An adversarial review (`/devils-advocate`) surfaced concrete one-line-bypass failure modes that a blanket "in-workspace" check misses:

| Command | Blanket check | Curated allowlist |
|---|---|---|
| `rm -rf ./` or `$PWD` | ✅ passes (collapses to workspace root) → **deletes everything** | ❌ basename = repo name, not in set |
| `rm -rf symlink→/etc` | ✅ substring anchor doesn't catch it | ❌ `safeRealpath` resolves outside workspace |
| `rm -rf node_modules /etc` | ✅ can't parse multi-target | ❌ `/etc` basename not in set |
| `rm -rf *.log` | ✅ glob unhandled | ❌ `*` metacharacter rejected |
| `rm -rf .git` | ✅ in-workspace | ❌ not in set (history loss) |

The curated approach covers 95%+ of real agent `rm -rf` usage at bounded, auditable blast radius. Adding a new safe target is a one-line addition to `RM_RF_SAFE_LEAF_DIRS`.

## Verification

- `risk-classifier.test.ts`: 89/89 pass — **unchanged** (BASH_HIGH not touched)
- `afk-mode-gate.test.ts`: 81/81 pass — 10 new allow tests (node_modules, dist, build, multi-target, `--` separator, `./` prefix), 13 new block tests (root/home/parent anchors, `.git`, `$HOME`, unknown target, non-allowlisted name, multi-target escape, glob, variable, sudo, shell operator chain)
- Full `src/agent/` suite: **6360/6362 pass** (2 pre-existing skips), 0 regressions, 315 test files green

## Scope

O3 only. O2 (ssh config + known_hosts read-denylist carve-outs) is #805. `BASH_HIGH` and `safe-destruct-detect.ts` are both unchanged — this is purely additive gate logic.


griffinwork40 (commented):

Reviewed ref `296f3332` (merge commit; branch already absorbed one review round in `caad8e73`). Full regime, 3 files, +338/−3.

The security core holds up. I tried hard to break it and the two most dangerous hypotheses both **failed to reproduce**:

- **Symlinked-parent escape — REFUTED.** For `rm -rf sub/node_modules` where `<ws>/sub → /outside` and the target does not exist, `safeRealpath` walks up to the first *existing* ancestor and `realpathSync`'s it, so `sub` is dereferenced before the tail is rejoined. Resolves to `/outside/node_modules`, containment correctly fails. (`write-denylist.ts:137-168`, ref:296f333, citation-type:file-state — verified via `git show` at the ref.)
- **Bracket-glob bypass (`[`/`]` absent from `SHELL_METACHAR`) — REFUTED.** The gate evaluates the *literal unexpanded* token, so `basename('node_module[s]')` never matches the allowlist set and the command is rejected before the shell ever expands it. Omitting `[`/`]` is not exploitable to widen the allowlist. (`afk-mode-rm-allowlist.ts:42`, ref:296f333, citation-type:diff-context.)

Also checked and clean: `/^rm\b/` looseness is unreachable (`rm-rf-everything.sh …` contains neither `rm -rf` nor `rm ` so the risk table never returns `high`, and the carve-out only runs on `high`); a high-risk keyword smuggled into a discarded flag token (`rm -rf node_modules --sudo`) is not an escalation because `rm` exits on the unrecognized long option without deleting; `toolName === 'bash'` vs. `classifyRisk`'s lowercasing fails *closed* and the schema registers `name: 'bash'` (`schemas.ts:14`), so it is inert-safe, not a widening.

### Findings (none blocking)

**medium · high confidence · spec-compliance · `afk-mode-rm-allowlist.ts:42` · ref:296f333 · citation-type:diff-context**
FINDING: The PR body names `rm -rf node_modules && pnpm install` as the motivating scenario, but the whole-string metacharacter scan rejects `&`, so that exact command still stalls on the approval prompt — the PR's own test asserts the compound form is blocked.
EVIDENCE:
```
const SHELL_METACHAR = /[*?$`(){};|&<>'"\\]/;
['rm -rf node_modules && echo done', 'shell operator chain'],   // asserted BLOCKED
```
SUGGESTION: Blocking the compound form is the *correct* security call — reword the PR body/issue note to state the carve-out covers a standalone `rm -rf node_modules` and that callers must split the compound into two bash calls, so the next reader does not file this as a regression.

**medium · high confidence · security/observability · `afk-mode-gate.ts:344-355` · ref:296f333 · citation-type:diff-context**
FINDING: The carve-out allow path returns `{}` with no `emitHookDecision`, so a destructive `rm -rf` permitted unattended leaves no durable `hook_decision` trace row, while every block path emits one — exactly the forensic record an away operator needs on resume.
EVIDENCE:
```
      if (cmd && isSafeInWorkspaceRm(cmd, resolveBase, workspaceRoot)) {
        return {};
      }
```
SUGGESTION: Emit `emitHookDecision(traceWriter, { hookEvent: 'PreToolUse', decision: 'allow', approvalOutcome: 'carve-out', blockedTool: toolName, … })` before the `return {}`, mirroring the block path at line 370.

**medium · medium confidence · security · `afk-mode-rm-allowlist.ts:103-113` · ref:296f333 · citation-type:diff-context**
FINDING: The only "is this really generated?" test is `lstatSync().isDirectory()`, so a git-**tracked** directory whose basename happens to be `target`, `out`, `build`, `dist`, or `coverage` is deleted unattended — recoverable from git only for committed content, unrecoverable for uncommitted work in that directory.
EVIDENCE:
```
try { return lstatSync(resolved).isDirectory(); }
catch (err) { return (err as NodeJS.ErrnoException).code === 'ENOENT'; }
```
SUGGESTION: Gate the allow on the target being ignored — `git check-ignore -q <resolved>` (or a `.gitignore` match) — so a tracked `target/`/`out/` in a consumer repo stays `high`. Verified absent: `git grep check-ignore` across `src/` at the ref returns zero hits.

**low · high confidence · test-coverage · `afk-mode-gate.test.ts` · ref:296f333 · citation-type:file-state**
FINDING: Three reachable paths have no test at the reviewed ref: a symlinked target resolving outside the workspace, a subagent `context.cwd` distinct from the session root, and a trailing-slash target `node_modules/`.
EVIDENCE: `git grep` at the ref over the test file — `symlink` → 0 matches, `parentSessionId.*rm|cwd:.*rm` → 0 matches, `node_modules/'` → 0 matches (55 `it`/`it.each` blocks present).
SUGGESTION: Add a tempdir case symlinking `<ws>/sub → <outside>` and asserting `rm -rf sub/node_modules` blocks — that is the containment invariant the whole module rests on, and it is currently unpinned.

**nit · high confidence · maintainability · `afk-mode-gate.ts` · ref:296f333 · citation-type:file-state**
FINDING: `afk-mode-gate.ts` grows 473 → 495 lines, extending an already-large gate module.
EVIDENCE: `git show <ref>:src/agent/afk-mode-gate.ts | wc -l` → 495; same on `origin/main` → 473.
SUGGESTION: Extract the whole carve-out block into a small `applyRmCarveOut(risk, toolName, context, …)` helper next to the allowlist module. No enforced size gate exists in this repo — flagged as house-style only.

### What was not checked
- **Citations verified inline against ref `296f3332` via `git show <ref>:<file>`** — the local checkout sits on `main`, so every `file-state` citation above was re-read at the PR head, not on disk. All citations resolved; none dropped as fabricated.
- **Stated intent: PR #806 title + body — spec-compliance assessed** (finding 1).
- Tests were **not executed**; the PR's verification claims (89/89 classifier unchanged, 81/81 gate, 6360/6362 suite) were taken as stated and not independently rerun. CI is green on all 8 required checks at this ref.
- Did not review the two mutated pre-existing fixtures (`rm -rf build` → `rm -rf /`) for downstream assertion coverage beyond confirming they still target a genuinely-blocked command.
- No runtime/TOCTOU testing: the gate `lstat`s at decision time and the shell runs later, so a check-to-exec swap is theoretically open. Not rated — single-agent local context makes it low-value to exploit.

griffinwork40:
@codex fix comments

griffinwork40 (commented):

## Summary

This PR adds a curated allowlist (`afk-mode-rm-allowlist.ts`, +172) and a 22-line carve-out in `afk-mode-gate.ts` so AFK autonomous mode can run `rm -rf` against a small set of generated directories without an approval prompt. Because the PR branch was not checked out locally, the parser logic was extracted verbatim at the reviewed ref and **executed** against a scratch workspace rather than reasoned about. The core containment invariant held on every adversarial probe: `/`, `.`, `..`, `.git`, `$HOME`, the workspace root itself, `/var/tmp/node_modules`, `dist/../../escape-node_modules`, a symlink named `node_modules` → `/etc`, a `sudo` prefix, `&&` chaining, globs, and `FOO=bar` env prefixes were **all correctly blocked**; non-recursive `rm build` was blocked while `-rfv`/`-R`/`-fr`/`--recursive` were allowed. The design (single call site, fail-closed default, separate testable module) is sound. Two bounded gaps and two hygiene items are below.

## Findings

**Medium · security/correctness · `src/agent/afk-mode-rm-allowlist.ts:42` (and `:82`)**
`SHELL_METACHAR` omits `\n` and `\r`, while `cmd.trim().split(/\s+/)` treats a newline as ordinary whitespace, so a multi-line command string is flattened into a single token list instead of being rejected as unparseable. Executed at the ref: `rm -rf node_modules\nout` returns **true** (allowed) — two separate shell commands admitted by one check. Blast radius is bounded, because every token must still clear the curated-basename, strictly-in-workspace, and is-a-directory-or-absent gates: `rm -rf node_modules\nrm -rf /etc` is correctly blocked (token `rm` is not allowlisted), as is `...\ncurl evil.sh`. This is a parsing-fidelity defect, not a containment escape.
*Suggestion:* add `\n\r` to the `SHELL_METACHAR` class so multi-line input fails closed at the same point as `;`/`&&`.

**Medium · observability · `src/agent/afk-mode-gate.ts:352`**
The carve-out's `return {}` emits no `emitHookDecision`, so this becomes the only path on which a command that classified `high` reaches execution with no `hook_decision` record — the block path (`:370`) and the operator-approval path (`decide()`, `:207`, which emits for approved outcomes too) both leave one. To be precise about scope: the command itself is still durably traced, since `emitToolCall` fires unconditionally before dispatch (`providers/anthropic-direct/loop/tool-dispatch.ts:65`), so what is missing is *which gate decision permitted it*, not *what ran*. Worth noting the pre-existing non-high allow path (`:356`) also emits nothing, so the omission is consistent with current convention rather than a regression.
*Suggestion:* emit a `hook_decision` with a distinct outcome (e.g. `approvalOutcome: 'auto-allowed'`) before returning, so an away operator can audit which `rm -rf` calls the carve-out admitted.

**Low · test-coverage · `src/agent/afk-mode-rm-allowlist.ts:1-172`**
No `afk-mode-rm-allowlist.test.ts` exists at this ref (`git ls-tree -r --name-only <ref> | grep rm-allowlist` returns only the source file), so the security-load-bearing parser is exercised only indirectly through the gate suite — despite the module docblock stating it is "unit-testable without constructing a gate." The newline case above is consequently unpinned.
*Suggestion:* add a direct unit-test file for `isSafeInWorkspaceRm` covering newline/multi-line input, trailing-slash and nested targets, and an absolute in-workspace path.

**Low · spec-compliance · PR description, Verification section**
The stated counts ("afk-mode-gate.test.ts: 81/81 pass — 10 new allow tests, 13 new block tests") match only the first two `it.each` blocks; the file at this ref also carries three later blocks explicitly labelled `// ---- PR #806 review: ...` (flag-token metacharacters, non-recursive `rm`, and a `generated-directory target check` describe using a real temp workspace), so the numbers predate the review-response commits.
*Suggestion:* refresh the Verification section against the current head so the claimed counts match what CI actually ran.

Two smaller notes, both **correct as-implemented** and raised only to confirm they were deliberate: `rm -rf node_modules/` and `rm -rf node_modules/.` are allowed (`path.resolve` normalises before `basename`, so they denote the same directory), and `rm -rf node_modules -- dist` is allowed because `--` is honoured anywhere in the argument list.

## What was not checked

- Citations verified against branch HEAD `296f333`. The PR branch was not checked out; the allowlist parser was verified by extracting its logic verbatim at the ref and executing it against a scratch workspace, **not** by running the project's vitest suite.
- CI is reported green on this head, but the test suite was not re-run locally, and the PR body's pass counts were not independently reproduced.
- TOCTOU between the gate's `lstatSync` and the shell's later delete was considered and deliberately not raised: the adversary would be the agent that already authored the command string, so it adds no privilege in this threat model.
- `BASH_HIGH` / `risk-classifier.ts` and `safe-destruct-detect.ts` were confirmed untouched, but their own suites were not re-run.
- Stated intent: PR #806 title+body — spec-compliance assessed.

## Merge decision

**MERGE.** The carve-out is strictly additive, loosens no existing protection, and every containment probe passed, so it delivers its stated purpose (unblocking headless `rm -rf node_modules && pnpm install`) without opening a path to an unattended destructive delete. The newline parsing gap and the missing `hook_decision` record are both real and both worth fast follow-ups, but neither admits a command outside the curated, in-workspace, generated-directory set.

PR #810, opened 2026-07-31
fix(tui): stop annihilating whitespace-only text deltas and fusing round-boundary prose

griffinwork40:
## Summary

Reported symptom: the TUI rendered `"The6 lint errors"` where the model wrote `"The 6 lint errors"`, and `"branch,as requested."` for `"branch, as requested."` — a space silently missing from streamed prose.

Root cause is **not** the wrap or markdown layer. `stripCommandTags` and `extractSkillTag` both ran `/^[ \t]+$/gm` on **every streamed delta**. Under `/m`, `^`/`$` also match the *string's own ends*, so a delta consisting of nothing but a space matched as a "whitespace-only line" and was replaced with `''`; the orchestrator's `if (!cleaned) return` guard (`stream-renderer-orchestrator.ts:331`) then discarded the emptied delta outright. When the provider split `"The 6 lint errors"` such that the space arrived as its own `text_delta`, the sequence `"The"` + `" "` + `"6 lint errors"` rendered as `"The6 lint errors"`.

The regex's purpose is legitimate — normalizing `\n   \n` into a true `\n\n` so `findBlockBoundary` still sees the paragraph break — it just needed to be **line-scoped instead of string-scoped**. Both copies now call one `normalizeWhitespaceOnlyLines` helper that requires a newline to be present. This mirrors the `tagsRemoved` guard that already protects newline-only deltas, documented in that file's own header (lines 23–32) — the `'\n'` case was fixed there previously; the space case was missed.

While tracing it, persisted transcripts surfaced a **second, independent instance of the same class**: `turn-handler` accumulated every tool-use round's text with `''`, fusing the text *before* a tool call with the text *after* it — `"check).All lint errors"`, `"test suites.Diff is clean"`, `` "`commitThinkingPhase`:Edit 2" ``. A `tool_result` now arms a round seam consumed by the next content chunk via `joinAtRoundSeam`, in its own file (`turn-handler.ts` is already 984 LOC). Intra-round deltas still concatenate verbatim, since those legitimately split mid-word.

**Ruled out with evidence, not reasoning:** `wrapToWidth`/`hardWrapToWidth` preserve the break space (it becomes a leading space on the continuation row — which is why wrapped lines look 1-space-indented); `renderMarkdownToTerminal` and `formatBlockForCommit` preserve `"The 6"` at every width 80–240; `markdown-stream.ts:240` accumulates verbatim with no trim; the compositor has no row-edge trim. A git-archaeology pass found no wrap commit that could reach short mid-line words.

## How was this tested?

- **The repro genuinely fails pre-fix:** reverting only the helper body fails **8** of the new cases; restoring it passes all 41. The `still normalizes an interior whitespace-only line` case passes in **both** directions, proving the regex's original blank-line purpose survives rather than being deleted.
- New coverage: whitespace-only deltas (`' '`, `'  '`, `'\t'`, `' \t '`) pass through both functions verbatim; a three-delta stream `['The', ' ', '6 lint errors…']` reassembles with its space; a round-boundary turn keeps `\n\n` between the pre- and post-tool-call blocks while two deltas of one round still fuse mid-word.
- `pnpm vitest run src/cli` — **284 files / 5049 tests passed**, on this branch's base (current `main`).

## Not covered

- The tool-lane artifact `pnpm vitest run src/agent/subagent/ srcsession-phase.test.ts` observed in the same screenshot does **not** share this cause — tool args stream as `input_json_delta` and never pass through `stripCommandTags`. Separate arg-formatting/truncation question, not investigated here.
- `translate.ts:114`'s `textParts.join('')` is a third latent instance of this class (fuses multiple text blocks within one message). Left alone deliberately: it feeds conversation history and needs its own blast-radius check.

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes (`src/cli`: 284 files / 5049 tests; full-suite gates run in CI)

### DIFF
--- a/src/cli/commands/interactive/turn-handler.test.ts
+++ b/src/cli/commands/interactive/turn-handler.test.ts
@@ -136,6 +136,38 @@ describe('runTurn — happy path', () => {
     expect(onAfterTurn).toHaveBeenCalledTimes(1);
   });
 
+  it('separates round-boundary text so the text before a tool call does not fuse with the text after it', async () => {
+    // Regression: responseText accumulated EVERY round's text with '' between
+    // them, so the assistant text block before a tool call fused with the block
+    // after it and the paragraph break vanished. Persisted transcripts showed
+    // "…(build doesn't type/lint check).All lint errors are pre-existing" and
+    // "…the affected test suites.Diff is clean" — the reported missing-space
+    // artifacts. Verbatim '' joining is still correct WITHIN a round (deltas
+    // split mid-word), so only the round seam gets a separator.
+    const events: OutputEvent[] = [
+      { type: 'chunk', chunk: { type: 'content', content: "Now running lint (build doesn't type/lint check)." } },
+      { type: 'chunk', chunk: { type: 'tool_use_detail', toolName: 'bash', toolUseId: 'tu-1', toolInput: '{"command":"pnpm lint"}' } },
+      { type: 'chunk', chunk: { type: 'tool_result', toolUseId: 'tu-1', content: '6 problems' } },
+      // Two deltas of the NEXT round's text: they must fuse with each other
+      // (mid-word split) but not with the pre-tool-call text.
+      { type: 'chunk', chunk: { type: 'content', content: 'All lint errors are pre-' } },
+      { type: 'chunk', chunk: { type: 'content', content: 'existing.' } },
+      { type: 'done', metadata: { durationMs: 10 } },
+    ];
+
+    const session = streamFrom(events);
+    const { h, onTurnComplete } = makeHandles();
+    const stats = makeStats();
+
+    await runTurn({ text: 'run lint', attachments: [] }, session, stats, h);
+
+    const recorded = onTurnComplete.mock.calls[0]?.[1] as string;
+    expect(recorded).not.toContain('check).All');
+    expect(recorded).toBe(
+      "Now running lint (build doesn't type/lint check).\n\nAll lint errors are pre-existing.",
+    );
+  });
+
   it('uses message event content when no content chunks streamed', async () => {
     const events: OutputEvent[] = [
       { type: 'message', message: { role: 'assistant', content: 'no-stream answer' } },
--- a/src/cli/commands/interactive/turn-handler.ts
+++ b/src/cli/commands/interactive/turn-handler.ts
@@ -32,6 +32,7 @@ import {
 } from '../../_lib/capture-mode.js';
 import { runWithSink } from '../../../agent/_lib/skill-sink-channel.js';
 import { parseTerminalState, type TerminalState } from './terminal-state.js';
+import { joinAtRoundSeam } from './turn-text-seam.js';
 import { renderVerdictCard } from './verdict-card.js';
 import { pushTerminalStateToTelegram, doneHasCorroboratingEvidence } from './afk-push.js';
 import { loadTelegramConfig, resolveAutoResumeOnUsageLimit } from '../../config.js';
@@ -82,6 +83,11 @@ export async function runTurn(
   // the duplicated partial text. Prior rounds (before the checkpoint) are
   // untouched — the retry is per-round, not per-turn.
   let roundStartResponseLen = 0;
+  // Set when a tool_result closes a round, consumed by the next content chunk:
+  // that chunk opens a NEW assistant text block, so it joins across a round
+  // seam (paragraph break) rather than concatenating verbatim like an
+  // intra-round delta. See {@link joinAtRoundSeam}.
+  let pendingRoundSeam = false;
   let streamingStarted = false;
   let streamErrorRendered = false;
   let rendererDisposed = false;
@@ -399,7 +405,10 @@ export async function runTurn(
         }
 
         if (event.type === 'chunk' && event.chunk.type === 'content') {
-          responseText += event.chunk.content;
+          responseText = pendingRoundSeam
+            ? joinAtRoundSeam(responseText, event.chunk.content)
+            : responseText + event.chunk.content;
+          pendingRoundSeam = false;
           streamingStarted = true;
         } else if (event.type === 'message' && !streamingStarted) {
           responseText = event.message.content;
@@ -412,6 +421,10 @@ export async function runTurn(
           // renderer.process(event) below, which resets the live display via
           // handleOrchestratorEvent's `stream_retry` case.
           responseText = responseText.slice(0, roundStartResponseLen);
+          // Truncating back to the round checkpoint also removes the seam break
+          // this round's first chunk inserted, so re-arm it: the re-streamed
+          // round is still a new text block after the prior round's text.
+          pendingRoundSeam = responseText.length > 0;
         }
 
         if (event.type === 'chunk' && event.chunk.type === 'tool_use_detail') {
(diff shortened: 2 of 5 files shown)


griffinwork40:
Wave 1.5 complete. Verification manifest:

| Check | Claim | Status | Verdict |
|---|---|---|---|
| B1 (absence) | `stream_retry` has no test coverage | `grep stream_retry turn-handler.test.ts` → 0 matches | **confirmed-absent** — finding stands |
| B2a (absence) | no `turn-text-seam.test.ts` | `ls` → no such file | **confirmed-absent** — finding stands |
| B2b (absence) | `joinAtRoundSeam` no direct test | `grep -rln joinAtRoundSeam *.test.ts` → 0 matches | **confirmed-absent** — finding stands |
| A (citation) | `turn-handler.test.ts:1-2443` (file-state) | file exists; happy-path assert at line 167 | **verified** |
| A (citation) | `turn-text-seam.ts:31-35` (diff-context) | visible in diff hunk | **verified** |
| A (rationale) | message branch `turn-handler.ts:413-414` unreachability | re-read at ref — content branch sets `streamingStarted=true` + consumes flag before message branch can fire | **verified** |
| A (rationale) | reset block `turn-handler.ts:600-606` | `pendingRoundSeam = false` confirmed (diff-context) | **verified** |

No findings dropped or downgraded. Both surviving findings are `low` severity (missing-test rubric tier), confidence `medium`. No `critical`/`high` findings → **no shadow-verify** round warranted. Synthesizing inline: 2 low findings, no cross-agent conflicts (single agent), state machine confirmed sound.

---

# Review: PR #810 — fix(tui): stop annihilating whitespace-only text deltas and fusing round-boundary prose

**Regime:** light (hotfix, +148/−3) · **Reviewed ref:** `cf0a3a39d9157152c58c991c069e05587a106bbd` · **Banner:** MERGE

## Summary

Two independent TUI text-streaming fixes, both root-caused with evidence in the PR body and backed by dedicated regression tests:

1. **`normalizeWhitespaceOnlyLines`** (`command-tags.ts`) — replaces two copies of `/^[ \t]+$/gm` with a helper requiring a `\n`. eliminates annihilation of space-only deltas, which had been dropped by the orchestrator's `if (!cleaned) return` guard and fused neighboring words ("The6 lint errors").
2. **`joinAtRoundSeam` + `pendingRoundSeam`** (`turn-handler.ts`, new `turn-text-seam.ts`) — arms a round-seam flag on `tool_result`, consumed by the next content chunk to insert a `\n\n` paragraph break instead of verbatim concatenation. fixes pre/post-tool-call text fusion ("…check).All lint errors…").

## Spec-compliance

Stated intent present (PR title + body) — **assessed**. Both intent requirements are met and neither unmet-intent nor scope-creep was found:
- "REQUIRES a newline to be present" — `command-tags.ts:55` implements it; both call sites (`:71`, `:111`) route through it. ✓
- `tool_result`→`pendingRoundSeam`→`joinAtRoundSeam` mechanism — armed `turn-handler.ts:442`, consumed `:408-411`, re-armed on stream_retry `:427`, reset on replay `:605`. ✓
- Explicitly out-of-scope items (`translate.ts:114`, `input_json_delta` artifact) untouched and not flagged. ✓ No scope creep detected.

## Correctness assessment (clean)

State machine `pendingRoundSeam` verified sound across all four transitions at the ref:
- **Arm on `tool_result`** (`:442`), **consume on next content** (`:408-411`) — multiple consecutive `tool_result`s re-set `roundStartResponseLen` to the same length and re-arm (benign); a `tool_result` with no subsequent prose leaves the flag armed through `done` with no effect (only content-join reads it).
- **Re-arm on `stream_retry`** (`:427`) — `pendingRoundSeam = responseText.length > 0` correctly avoids a spurious seam on a first-round-only retry (slice yields `''`, flag stays false — identical to `joinAtRoundSeam('','x')==='x'`).
- **`message`-branch unreachability** — the agent flagged this as untested file-state rationale; verified at ref: the content branch sets `streamingStarted=true` and consumes the flag *before* the message branch's `!streamingStarted` guard can ever pass with the flag armed, so `pendingRoundSeam=true && !streamingStarted` is unreachable. Confirmed.
- **`joinAtRoundSeam` idempotency** — `/\s$/`·`/^\s/` boundary suppression (incl. a single trailing `\n`) producing one-linebreak join rather than `\n\n` matches the documented "existing whitespace is authoritative" contract and reproduces pre-fix `''`-join behavior; no divergent-separator risk.
- **`normalizeWhitespaceOnlyLines`** — for any `\n`-bearing input the regex path is byte-identical to pre-fix; sole behavior delta is no-`\n` whitespace-only inputs returning verbatim (the intended fix). `.includes('\n')` short-circuit skips the regex on the common no-newline delta — marginal perf gain, regex literal cached.

api-compat / security / perf: clean — only production importer (`stream-renderer-orchestrator.ts`) sees the intended behavior change; no secrets touched; regexes linear (no ReDoS).

## Findings

> **F1 · low · medium confidence · test-coverage · `src/cli/commands/interactive/turn-handler.test.ts` · ref cf0a3a3 · file-state [verified]**
> The `stream_retry` re-arm branch (`pendingRoundSeam = responseText.length > 0`) — the most intricate of the four state-machine transitions — has zero test coverage: `grep stream_retry turn-handler.test.ts` → 0 matches.
> **Suggestion:** add a test emitting `content` → `tool_result` → `content` → `stream_retry` → `content` and assert the re-streamed text joins via `\n\n`; plus a first-round-only `stream_retry` asserting no leading seam.

> **F2 · low · medium confidence · test-coverage · `src/cli/commands/interactive/turn-text-seam.ts:31-35` · ref cf0a3a3 · diff-context**
> `joinAtRoundSeam` has no dedicated unit test (no `turn-text-seam.test.ts`; `grep -rln joinAtRoundSeam *.test.ts` → 0). Exercised only indirectly by one happy-path assert (`turn-handler.test.ts:167`). The empty-accumulator guard, the whitespace-boundary suppression branch, and the armed-but-never-consumed-through-`done` path are unchecked.
> **Suggestion:** add a pure unit file asserting the three branches — `('', 'x')==='x'`, `('a\n', 'b')==='a\nb'`, `('a','b')==='a\n\nb'` — plus a turn-handler case for the flag armed at `done`.

## Merge decision

**MERGE** — the fixes are correct, the state machine is verified sound, spec-compliance is met, and no security/api-compat/perf issues surfaced. The two open findings are `low`-severity missing-test gaps for non-happy-path branches of the new state machine — worth picking up as follow-up coverage but not blocking.

## What was not checked

- Citations verified inline against reviewed ref `cf0a3a39d9157152c58c991c069e05587a106bbd` (the checked-out HEAD); file-state citations and absence claims were re-resolved via `grep`/`git show` and all confirmed.
- Tests were not executed (`pnpm vitest`); the PR body claims 5049 passing on `src/cli`; the `stream_retry` event type was assumed constructible in the harness `OutputEvent` union (it dispatches in production at `turn-handler.ts:417`).
- The deferred out-of-scope instances (`translate.ts:114` `textParts.join('')`, the `input_json_delta` tool-lane artifact) were not reviewed — they are explicitly acknowledged-not-fixed in the intent.
- Telegram surface and `stream-renderer-orchestrator.ts` consumer wiring were inspected only for the call-site/drop-guard; full orchestrator regression not run.

**Done** — read-only review of PR #810 complete; no repository/PR state mutated (working-tree dirt preserved). Evidence: diff artifact `/home/griffinlong/.afk/state/skill-preflight/0c222642-8b7b-4e77-8e59-3afe4945557e/pr-810.diff`, reviewed ref `cf0a3a39d9157152c58c991c069e05587a106bbd`. Recommendation: **MERGE** with two low test-coverage follow-ups (F1, F2).

---

PR #813, opened 2026-07-31
fix(worktree): gate orphaned-dir removal on age and content (#794)

### LINKED ISSUE #794: worktree sweep: the orphaned-dir path removes recursively with no dirty check, no ignored-file probe, and no age gate
## Summary

`#770` closed the reaper's data-loss paths for worktrees git **knows about**. The `orphaned-dir` branch is a fourth removal path that bypasses every one of those protections: it deletes recursively with **no dirty check, no ignored-file probe, and no age gate**.

A directory under `.afk-worktrees/` that is absent from `git worktree list` is deleted outright, along with everything in it, on the next sweep tick.

## Evidence (all line numbers at `16c4cd9`, current `main`)

`src/agent/worktree-sweep.ts`:

```js
527:    const orphanedDirs = diskEntries.filter((d) => !registeredPaths.has(d));
...
536:        let orphanAgeMs = 0;
539:          orphanAgeMs = Date.now() - stat.birthtimeMs;
543:          verdict: 'orphaned-dir',
545:          ageMs: orphanAgeMs,
549:            await fs.rm(orphanPath, { recursive: true, force: true });
```

Three specific bypasses:

1. **No dirty check and no probe.** Registered candidates run `git status --porcelain` and then `probeNonRebuildableIgnoredFiles` (`:656`) before any removal. The orphan loop calls neither — it goes straight from "not in the registry" to `fs.rm(..., { recursive: true, force: true })` at `:549`.
2. **No age gate, despite computing the age.** `orphanAgeMs` is computed at `:536-539` and attached to the candidate at `:545`, then never compared to anything. The thresholds exist (`cleanThresholdMs` / `dirtyThresholdMs` at `:337-338`, `MIN_EMPTY_AGE_MS` at `:366`) but are only consulted inside `classifyCandidate` (`:764`), which this path never calls — it hard-codes `verdict: 'orphaned-dir'` at `:543`. So a directory created *seconds* ago is eligible.
3. **`fs.rm`, not `git worktree remove`.** It never goes through the guarded-remove helpers, so none of the preservation logic added in #770 is even reachable.

## How a directory becomes an "orphan"

`registeredPaths` is derived from `git worktree list --porcelain`, so anything on disk under `.afk-worktrees/` that git does not list qualifies. Realistic ways in:

- **A partially-failed `git worktree add`** — the directory is created, registration does not complete. This is the most likely one in practice, and the failure mode is silent.
- **A corrupted or deleted `.git/worktrees/<name>` admin entry** — the checkout is alive and in use, but git stops listing it.
- **A human staging files there** — `mkdir .afk-worktrees/scratch`, drop notes in, come back later.
- **`.git` replaced or the repo re-cloned** while a worktree directory survives on disk.

Note the asymmetry with the bug #759 fixed: there, at least git knew about the tree and the sweep consulted `status`. Here the sweep has *no* information about the directory and deletes it anyway.

## Impact

Unbounded, unrecoverable local file loss with no age grace period and no warning naming what was destroyed. It is narrower than #759 in reach (requires the tree to be unregistered) but strictly more destructive when it fires, since none of the four protections apply. Uncommitted work, `.env` files, and anything else in that directory are gone.

## Suggested direction

Route the orphan path through the same gates as registered candidates rather than special-casing it:

- Add an age gate before removal, using the already-computed `orphanAgeMs`. An orphan younger than `MIN_EMPTY_AGE_MS` is very likely an in-flight `worktree add` and must never be touched.
- Run `probeNonRebuildableIgnoredFiles` on the orphan and preserve when it protects, warning with the reason (the `DirtyReason` plumbing from #770 already exists).
- Consider `git status --porcelain` too. It fails on an unregistered directory, so it cannot be used as-is — a plain "does this contain any non-ignored file that is not in git" check is the substitute.
- Keep the existing `scope === 'all' || scope === 'interactive'` guard at `:534` as-is; that one is correct.

Preserving an orphan is cheap (a leaked directory, reclaimed once it ages out or is removed explicitly). Deleting a live one is not.

## Provenance

Out of scope for #770 by design and called out in that PR's review; filing so it is not lost. Follow-up to #759.

griffinwork40:
## Summary

`#770` closed the reaper's data-loss paths for worktrees git **knows about**. The `orphaned-dir` branch was the fourth removal path and it bypassed every one of those protections — it deleted recursively with no dirty check, no ignored-file probe, and no age gate. This routes it through an equivalent floor.

Preserving an orphan costs one leaked directory, reclaimed when it ages out or is removed explicitly. Deleting a live one is unrecoverable — so every uncertain answer now preserves.

## Root cause

`src/agent/worktree-sweep.ts:534-556` (at `5456b75`):

- **No dirty check, no probe.** Registered candidates run `git status --porcelain` and then `probeNonRebuildableIgnoredFiles`. The orphan loop called neither — it went straight from "not in the registry" to `fs.rm(orphanPath, { recursive: true, force: true })`.
- **No age gate, despite computing the age.** `orphanAgeMs` was computed and attached to the candidate, then never compared to anything. The thresholds exist but are only consulted inside `classifyCandidate`, which this path never calls — it hard-codes `verdict: 'orphaned-dir'`. A directory created *seconds* ago was eligible.
- **Raw `fs.rm`.** It never went through the guarded-remove helpers, so none of #770's preservation logic was reachable.

Unlike #759, an unregistered directory has **no branch ref holding its contents**, so there is no recovery path: uncommitted work and `.env` files are simply gone.

## Change

New module **`src/agent/worktree-orphan-guard.ts`** (138 LOC) exporting `classifyOrphanDir()`, which never deletes and never throws:

1. **Age gate first** (no IO) — an orphan younger than `MIN_EMPTY_AGE_MS` is never touched. This is the most likely cause of an orphan in practice: an in-flight `git worktree add`. A caller that could not `stat` the directory passes `0`, which preserves rather than reading as "infinitely old".
2. **Bounded content walk** — each entry is classified with the existing shared policy (`classifyIgnoredEntry`), which defaults unrecognised paths to `protected`. Any non-rebuildable entry preserves and names the file.
3. **Fail-safe in one direction only** — unreadable directory, depth bound, or exhausted entry budget all preserve.

It lives in its own file because `worktree-sweep.ts` is already **852 LOC**, 2.4× over the repo's hard 350-LOC ceiling; the call site there is a thin 12-line edit. This follows the existing split precedent (`worktree-ignored-probe.ts` = IO, `worktree-ignored-patterns.ts` = policy).

**One design point worth review:** only `opaque` directories (dependency trees) are skipped wholesale. `inspectable` build output **is** descended into, because `dist/prod.env` is precisely the case #759 was filed for — skipping both classes would have reproduced that bug on the orphan path. A regression test pins it.

The guard also runs in **dry-run**, so `afk worktree list` reports the preservation and its reason instead of implying the next tick will delete. The `scope === 'all' || scope === 'interactive'` guard is untouched.

## Tests

New `src/agent/worktree-orphan-guard.test.ts` — 12 cases: fresh orphan preserved; failed-stat (age 0) and non-finite age preserved; `.env` preserved; untracked source file preserved (naming the file); secret nested under `dist/` preserved; aged empty orphan removed; aged dependency-tree-only orphan removed; unreadable path preserved; entry-budget and depth bounds preserved; a plain file as the target does not throw.

In `worktree-sweep.test.ts`, the existing case *"detects and removes directories in .afk-worktrees/ not registered in git"* **asserted the buggy contract** — it created a fresh orphan holding `somefile.txt` and expected removal. It is replaced by three integration cases: fresh orphan detected but preserved (with the warning); aged rebuildable-only orphan still removed (age injected by stubbing the one `fs.stat` the orphan path performs, since birthtime cannot be backdated portably); aged orphan with a non-rebuildable file preserved.

## Gates

| Gate | Result |
|---|---|
| `pnpm exec vitest run src/agent/worktree-orphan-guard.test.ts` | 12/12 pass |
| `pnpm exec vitest run src/agent/worktree-sweep.test.ts` | 49/49 pass |
| Combined run | 61/61 pass |

## Risk and blast radius

- **Behaviour change:** unregistered directories are no longer removed on sight. A genuine leak now waits out `MIN_EMPTY_AGE_MS` (1 hour) and must hold only rebuildable content. The trade is deliberate and stated in the issue.
- **Over-preservation is the failure mode**, by construction. An orphan holding an unrecognised file is now immortal until removed explicitly — acceptable given the alternative was silent unrecoverable deletion, but worth knowing.
- Touches the same file family as open PRs #799 and #771. Kept additive with no renamed exports to minimise conflict; the call-site edit is 12 lines.

Fixes #794

### DIFF
--- a/src/agent/worktree-orphan-guard.test.ts
+++ b/src/agent/worktree-orphan-guard.test.ts
@@ -1,0 +1,120 @@
+import { describe, expect, it, beforeEach, afterEach } from 'vitest';
+import { mkdtemp, mkdir, writeFile, rm } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { classifyOrphanDir } from './worktree-orphan-guard.js';
+
+const HOUR_MS = 3_600_000;
+const AGED_MS = HOUR_MS * 2;
+
+let root: string;
+
+beforeEach(async () => {
+  root = await mkdtemp(join(tmpdir(), 'afk-orphan-guard-'));
+});
+
+afterEach(async () => {
+  await rm(root, { recursive: true, force: true });
+});
+
+describe('classifyOrphanDir — age gate', () => {
+  it('preserves an orphan younger than the age gate (an in-flight worktree add)', async () => {
+    await writeFile(join(root, '.env'), 'SECRET=1');
+    const verdict = await classifyOrphanDir(root, 5_000, HOUR_MS);
+    expect(verdict).toEqual({
+      remove: false,
+      because: 'too-young',
+      detail: 'age 5000ms < 3600000ms',
+    });
+  });
+
+  it('preserves when the caller could not stat the directory (age 0)', async () => {
+    // Contract: a failed stat must pass 0, which the age gate refuses — it must
+    // never read as "infinitely old" and authorise a delete.
+    const verdict = await classifyOrphanDir(root, 0, HOUR_MS);
+    expect(verdict.remove).toBe(false);
+    if (!verdict.remove) expect(verdict.because).toBe('too-young');
+  });
+
+  it('preserves on a non-finite age rather than trusting the comparison', async () => {
+    const verdict = await classifyOrphanDir(root, Number.NaN, HOUR_MS);
+    expect(verdict.remove).toBe(false);
+  });
+});
+
+describe('classifyOrphanDir — content probe', () => {
+  it('preserves an aged orphan holding a non-rebuildable ignored file', async () => {
+    await writeFile(join(root, '.env'), 'TOKEN=abc');
+    const verdict = await classifyOrphanDir(root, AGED_MS, HOUR_MS);
+    expect(verdict).toEqual({
+      remove: false,
+      because: 'non-rebuildable-content',
+      detail: '.env',
+    });
+  });
+
+  it('preserves an aged orphan holding an untracked source file', async () => {
+    await mkdir(join(root, 'src'), { recursive: true });
+    await writeFile(join(root, 'src', 'index.ts'), 'export {};');
+    const verdict = await classifyOrphanDir(root, AGED_MS, HOUR_MS);
+    expect(verdict.remove).toBe(false);
+    if (!verdict.remove) {
+      expect(verdict.because).toBe('non-rebuildable-content');
+      expect(verdict.detail).toBe('src/index.ts');
+    }
+  });
+
+  it('preserves a secret nested under a build-output directory', async () => {
+    await mkdir(join(root, 'dist'), { recursive: true });
+    await writeFile(join(root, 'dist', 'prod.env'), 'K=v');
+    const verdict = await classifyOrphanDir(root, AGED_MS, HOUR_MS);
+    expect(verdict.remove).toBe(false);
+    if (!verdict.remove) expect(verdict.because).toBe('non-rebuildable-content');
+  });
+
+  it('removes an aged, genuinely empty orphan', async () => {
+    const verdict = await classifyOrphanDir(root, AGED_MS, HOUR_MS);
+    expect(verdict).toEqual({ remove: true });
+  });
+
+  it('removes an aged orphan whose only content is a dependency tree', async () => {
+    // The sweep must still reclaim the common case — a failed `worktree add`
+    // that got as far as installing deps — or the guard defeats its purpose.
+    await mkdir(join(root, 'node_modules', 'left-pad'), { recursive: true });
+    await writeFile(join(root, 'node_modules', 'left-pad', 'index.js'), 'module.exports=1;');
+    const verdict = await classifyOrphanDir(root, AGED_MS, HOUR_MS);
+    expect(verdict).toEqual({ remove: true });
+  });
+});
+
+describe('classifyOrphanDir — failure and bounds', () => {
+  it('preserves when the directory cannot be read', async () => {
+    const verdict = await classifyOrphanDir(join(root, 'does-not-exist'), AGED_MS, HOUR_MS);
+    expect(verdict.remove).toBe(false);
+    if (!verdict.remove) expect(verdict.because).toBe('scan-failed');
+  });
+
+  it('preserves when the walk outgrows its entry budget', async () => {
+    await mkdir(join(root, 'build'), { recursive: true });
+    for (let i = 0; i < 4; i++) {
+      await writeFile(join(root, 'build', `chunk-${i}.js`), '0');
+    }
+    const verdict = await classifyOrphanDir(root, AGED_MS, HOUR_MS, { maxEntries: 2 });
+    expect(verdict.remove).toBe(false);
+    if (!verdict.remove) expect(verdict.because).toBe('scan-budget-exhausted');
+  });
+
+  it('preserves rather than descending past the depth bound', async () => {
+    await mkdir(join(root, 'a', 'b', 'c'), { recursive: true });
+    const verdict = await classifyOrphanDir(root, AGED_MS, HOUR_MS, { maxDepth: 1 });
+    expect(verdict.remove).toBe(false);
+    if (!verdict.remove) expect(verdict.because).toBe('scan-failed');
+  });
+
+  it('never throws on a path that is a file rather than a directory', async () => {
+    const filePath = join(root, 'plain.txt');
+    await writeFile(filePath, 'x');
+    const verdict = await classifyOrphanDir(filePath, AGED_MS, HOUR_MS);
+    expect(verdict.remove).toBe(false);
+  });
+});
--- a/src/agent/worktree-orphan-guard.ts
+++ b/src/agent/worktree-orphan-guard.ts
@@ -1,0 +1,159 @@
+/**
+ * Removal gate for UNREGISTERED directories under `.afk-worktrees/`.
+ *
+ * Invariant: the sweep's orphan path has no git information to work with. The
+ * directory is absent from `git worktree list`, so `git status --porcelain` —
+ * the sweep's only dirty signal — cannot classify it, and every protection the
+ * registered-candidate path applies (dirty check, ignored-file probe, age gate)
+ * is unreachable there. This module is the substitute floor: it decides from the
+ * filesystem alone, and every uncertain answer preserves.
+ *
+ * Contract: `classifyOrphanDir` never deletes and never throws. It returns
+ * `{ remove: true }` only for a directory that is BOTH older than the caller's
+ * minimum age AND provably free of content a rebuild would not restore. Anything
+ * else — a fresh directory, an unreadable one, a scan that outgrows its budget,
+ * or a single unrecognised entry — yields `{ remove: false }` with a reason the
+ * caller can print.
+ *
+ * The asymmetry is deliberate: preserving a dead orphan costs one leaked
+ * directory, reclaimed when it ages out or is removed explicitly. Deleting a
+ * live one destroys uncommitted work and `.env` files with no recovery path,
+ * because an unregistered directory has no branch ref holding its contents
+ * (#794, following #759).
+ *
+ * @module agent/worktree-orphan-guard
+ */
+
+import { readdir } from 'node:fs/promises';
+import { join } from 'node:path';
+import { classifyIgnoredEntry } from './worktree-ignored-patterns.js';
+
+/**
+ * Contract: `classifyIgnoredEntry` is the shared policy and it defaults an
+ * unrecognised path to `'protected'`. That default is what makes this walk
+ * correct for a directory git never classified at all: an ordinary source file
+ * is as protective as a stray secret, so only recognisably rebuildable output
+ * can clear the way for a delete. Depending on the pure policy module (rather
+ * than the probe's boolean wrapper) keeps this module free of git IO.
+ */
+function isRebuildable(relPath: string): boolean {
+  return classifyIgnoredEntry(relPath) !== 'protected';
+}
+
+/** Why an orphaned directory must be left on disk. */
+export type OrphanPreserveReason =
+  /** Younger than the age gate — very likely an in-flight `git worktree add`. */
+  | 'too-young'
+  /** Holds at least one entry a rebuild would not restore. */
+  | 'non-rebuildable-content'
+  /** The directory could not be read, so its contents are unknown. */
+  | 'scan-failed'
+  /** The walk outgrew its bounds before it could prove the tree disposable. */
+  | 'scan-budget-exhausted';
+
+/** Fate of one unregistered directory under `.afk-worktrees/`. */
+export type OrphanVerdict =
+  | { remove: true }
+  | { remove: false; because: OrphanPreserveReason; detail?: string };
+
+/**
+ * Walk bounds. A failed `worktree add` can still have left a full
+ * `node_modules/` behind, so the walk must not be free to enumerate it — but
+ * rebuildable directories are skipped WHOLESALE (never descended into), which
+ * keeps the real cost proportional to hand-authored content rather than to
+ * dependency-tree size.
+ */
+export const ORPHAN_SCAN_MAX_DEPTH = 8;
+export const ORPHAN_SCAN_MAX_ENTRIES = 4_000;
+
+/**
+ * Decide whether an unregistered `.afk-worktrees/` directory may be removed.
+ *
+ * @param orphanPath  Absolute path to the directory.
+ * @param orphanAgeMs Age from directory birthtime. A caller that could not stat
+ *                    the directory must pass `0`, which preserves via the age
+ *                    gate rather than reading as "infinitely old".
+ * @param minAgeMs    Minimum age before removal is considered at all.
+ */
+export async function classifyOrphanDir(
+  orphanPath: string,
+  orphanAgeMs: number,
+  minAgeMs: number,
+  limits: { maxDepth?: number; maxEntries?: number } = {},
+): Promise<OrphanVerdict> {
+  // Age gate first: it is the cheapest check and the one that covers the most
+  // likely cause of an orphan, a `git worktree add` still in flight.
+  if (!Number.isFinite(orphanAgeMs) || orphanAgeMs < minAgeMs) {
+    return {
+      remove: false,
+      because: 'too-young',
+      detail: `age ${Math.max(0, Math.round(orphanAgeMs))}ms < ${minAgeMs}ms`,
+    };
+  }
+
+  const maxDepth = limits.maxDepth ?? ORPHAN_SCAN_MAX_DEPTH;
+  const maxEntries = limits.maxEntries ?? ORPHAN_SCAN_MAX_ENTRIES;
+
+  let remaining = maxEntries;
+  const pending: Array<{ abs: string; rel: string; depth: number }> = [
+    { abs: orphanPath, rel: '', depth: 0 },
+  ];
+
+  while (pending.length > 0) {
+    const dir = pending.pop();
+    if (dir === undefined) break;
+
+    let entries;
+    try {
+      entries = await readdir(dir.abs, { withFileTypes: true });
+    } catch (err) {
+      return {
+        remove: false,
+        because: 'scan-failed',
+        detail: `${dir.rel === '' ? orphanPath : dir.rel}: ${err instanceof Error ? err.message : String(err)}`,
+      };
+    }
+
+    for (const entry of entries) {
+      if (--remaining < 0) {
+        return {
+          remove: false,
+          because: 'scan-budget-exhausted',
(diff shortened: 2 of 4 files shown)


griffinwork40:
@codex fix comments

griffinwork40 (commented):

## Merge decision: **MERGE**

No blocking findings. The core invariant this PR exists to establish — *every uncertain answer preserves* — holds on every path I traced, and the headline design point (`inspectable` build output descended into, so `dist/prod.env` is caught) is genuinely pinned by a test rather than merely asserted in the description. The 10 non-blocking items below are polish, split between two reason-code conflations, an observability gap in four downstream consumers, and two stale counts in the PR body.

**Reviewed ref:** `dc964a3492c6a69727109ef530c7ed1289d13163` · 4 files · +439 / −6
**Regime:** full (2 dimension agents + inline citation/absence verification)

---

### Invariant check (why this is a MERGE)

Traced every exit of `classifyOrphanDir` at the reviewed ref. `remove: true` is returned from exactly one place — line 158, after the walk drains without objection:

| Uncertain condition | Line | Outcome |
|---|---|---|
| Age below gate, non-finite, or failed stat (`0`) | `worktree-orphan-guard.ts:86-92` | preserve `too-young` |
| `readdir` rejection (incl. non-directory target) | `:109-115` | preserve `scan-failed` |
| Entry budget exhausted | `:118-124` | preserve `scan-budget-exhausted` |
| Depth bound hit | `:141-143` | preserve (reason conflated — see nit 6) |
| Unrecognised entry (`classifyIgnoredEntry` → `protected`) | `:152-154` | preserve `non-rebuildable-content` |

Call-site wiring is correct in both modes: `if (!guard.remove) { …warnings.push…; continue; }` at `worktree-sweep.ts:561-567` sits **before** the `if (!effectiveDryRun)` block at `:568`, so a preserve verdict short-circuits the `fs.rm` and dry-run reports the preservation. The `scope === 'all' || scope === 'interactive'` guard at `:537` is untouched, as claimed. Empty and opaque-only aged orphans still reach `remove: true`, so the reclamation path is not over-tightened.

---

### Findings (0 blocking · 5 low · 5 nit)

**1. low · medium confidence · security · `src/agent/worktree-orphan-guard.ts:148-154` · citation-type:file-state**
Symlink entries are classified by filename pattern alone — `entry.isDirectory()` is false for a symlink, so a link named like rebuildable output (`dist`, `build`) routes to `isRebuildable(rel)` and clears the way for removal with no `lstat`/`isSymbolicLink()` distinguishing it from real build output.
```js
// Files, symlinks, sockets — anything not a directory.
if (!isRebuildable(rel)) {
  return { remove: false, because: 'non-rebuildable-content', detail: rel };
}
```
Blast radius is bounded — `fs.rm` unlinks the link, it does not follow into the target — so this destroys a user-created link, not data outside the tree. Verified absent: grep for `isSymbolicLink|lstat|realpath` returns zero matches in this module.
*Suggestion:* treat any entry where `entry.isSymbolicLink()` is true as non-rebuildable unconditionally — a name match proves nothing about the link's target.

**2. low · medium confidence · security · `src/agent/worktree-sweep.ts:554-570` · citation-type:file-state**
Nothing re-verifies between the guard's scan and the destructive `fs.rm`, so content written into the orphan directory after `classifyOrphanDir` returns and before the removal runs is deleted unclassified.
```js
const guard = await classifyOrphanDir(orphanPath, orphanAgeMs, MIN_EMPTY_AGE_MS);
...
await fs.rm(orphanPath, { recursive: true, force: true });
```
Verified absent: no `stat`/probe/re-classify call exists between `:554` and `:570`. The 1-hour age gate makes the window narrow in practice (a directory this old is by definition not being actively created), which is why this is low rather than high.
*Suggestion:* re-stat the directory's `mtimeMs` immediately before `fs.rm` and skip removal when it changed since the scan.

**3. low · high confidence · api-compat · `src/agent/worktree-sweep.ts:124,557` · citation-type:diff-context**
The new `'orphaned-dir-preserved'` literal is unknown to all four production consumers of the verdict surface, so a preserved orphan renders in `afk worktree list` with the same green "no" as a healthy active worktree, with no visual marker that it is a leak awaiting attention.
```js
| 'orphaned-dir-preserved'
...
verdict: guard.remove ? 'orphaned-dir' : 'orphaned-dir-preserved',
```
Reachability checked at the reviewed ref — four production importers, none handling the new literal: `src/agent/daemon/scheduler.ts:505`, `src/cli/slash/commands/worktree.ts:41`, `src/cli/commands/worktree.ts:36`, `src/cli/commands/interactive/boot-prune.ts:53`. Each **degrades correctly** (a preserved orphan genuinely is not prunable, so excluding it from every prunable set is right, and the `[WARN]` line *is* surfaced via `result.warnings` at `cli/commands/worktree.ts:116-118` and `slash/commands/worktree.ts:228`). This is an observability gap, not wrong output — hence low, not high.
*Suggestion:* add `'orphaned-dir-preserved'` to `WARNING_VERDICTS` in `src/cli/slash/commands/worktree.ts:46` and to the `warn` branch of `verdictWouldPrune` in `src/cli/commands/worktree.ts:39`.

**4. low · medium confidence · test-coverage · `src/agent/worktree-orphan-guard.test.ts:39-42` · citation-type:file-state**
The non-finite-age case asserts only that removal was refused, never which gate refused it, so a future regression that preserves NaN ages for the wrong reason (e.g. a `scan-failed` on an unreadable root) still passes.
```js
it('preserves on a non-finite age rather than trusting the comparison', async () => {
  const verdict = await classifyOrphanDir(root, Number.NaN, HOUR_MS);
  expect(verdict.remove).toBe(false);
});
```
The sibling age-0 case at `:36` does assert `.because`, so this is an inconsistency within the same file.
*Suggestion:* add `if (!verdict.remove) expect(verdict.because).toBe('too-young');`.

**5. low · medium confidence · test-coverage · `src/agent/worktree-orphan-guard.test.ts:107-112` · citation-type:file-state**
The depth-bound test asserts `because === 'scan-failed'` — the same code a genuine read failure returns — so it pins the reason-code conflation described in nit 6 as the expected contract rather than catching it.
```js
const verdict = await classifyOrphanDir(root, AGED_MS, HOUR_MS, { maxDepth: 1 });
expect(verdict.remove).toBe(false);
if (!verdict.remove) expect(verdict.because).toBe('scan-failed');
```
*Suggestion:* assert `verdict.detail` contains `'depth limit'`, and update to the distinct reason introduced by nit 6.

**6. nit · high confidence · correctness · `src/agent/worktree-orphan-guard.ts:141-143` · citation-type:file-state**
The depth-bound preserve reuses the `scan-failed` reason, so the operator-facing warning reports a `readdir` failure that never happened.
```js
if (dir.depth + 1 > maxDepth) {
  return { remove: false, because: 'scan-failed', detail: `depth limit at ${rel}` };
}
```
The `OrphanPreserveReason` union already distinguishes `scan-budget-exhausted` from `scan-failed`, so depth is the one bound without its own code. Preservation itself is correct — intent claim 5 is met; only the label is wrong.
*Suggestion:* add `'depth-limit-exceeded'` to `OrphanPreserveReason` and return it here.

**7. nit · medium confidence · correctness · `src/agent/worktree-orphan-guard.ts:86` · citation-type:file-state**
The age gate uses strict `<`, so a directory whose age exactly equals `minAgeMs` clears the gate — one tick earlier than the module docstring's "older than the caller's minimum age".
```js
if (!Number.isFinite(orphanAgeMs) || orphanAgeMs < minAgeMs) {
```
A 1 ms divergence on a 3,600,000 ms gate; noted for the docstring, not the behavior.
*Suggestion:* reword the contract comment to "at least the caller's minimum age", or switch to `<=`.

**8. nit · medium confidence · api-compat · `src/agent/worktree-sweep.ts:539-547` · citation-type:diff-context**
`ageMs` on an emitted candidate now carries `0` for both "brand new" and "birth time unavailable", collapsing two distinct states into one value that renders as `-` in the age column.
```js
if (Number.isFinite(stat.birthtimeMs) && stat.birthtimeMs > 0) {
  orphanAgeMs = Math.max(0, Date.now() - stat.birthtimeMs);
}
```
The conservative direction is right (both preserve); only the reporting is lossy.
*Suggestion:* carry an explicit `ageUnknown` boolean on the candidate rather than overloading `ageMs === 0`.

**9. nit · high confidence · spec-compliance · PR description vs `src/agent/worktree-orphan-guard.ts`**
The description states the new module is 138 LOC; the file at the reviewed ref is **159** lines. Still well inside the repo's 350-LOC ceiling, so the architectural argument for the split stands unchanged.
*Suggestion:* correct the count in the description.

**10. nit · high confidence · spec-compliance · PR description vs `src/agent/worktree-sweep.ts`**
The call-site change is described as "a thin 12-line edit"; the actual change is **+23 / −2 across three hunks** (import at `:17`, verdict union at `:120`, orphan loop at `:539`). Only the non-comment logic subset is ~12 lines.
*Suggestion:* scope the claim to the logic-only subset, or update the number.

---

### Spec-compliance verdict — 8 PASS / 2 count-drift

Stated intent taken from the PR title + body; spec-compliance **was** assessed.

| # | Claim | Verdict |
|---|---|---|
| 1 | Old path used raw `fs.rm` with no age/dirty/ignored gate | PASS |
| 2 | Module is 138 LOC | **FAIL** — 159 |
| 3 | New file respects the 350-LOC ceiling | PASS — 159 < 350 |
| 4 | Age gate first, no IO; failed stat (`0`) preserves | PASS — `:86-92` precedes `readdir` at `:108` |
| 5 | Non-rebuildable entry preserves and names the file | PASS — `detail: rel` at `:153`, asserted in tests |
| 6 | Unreadable / depth / budget all preserve | PASS (reason code conflated — nit 6) |
| 7 | `inspectable` descended into; regression test pins it | PASS — test `:67-73` asserts `remove:false` |
| 8 | Guard runs in dry-run | PASS — guard call precedes the `effectiveDryRun` branch |
| 9 | Call site is a thin 12-line edit | **FAIL** — +23 / −2, three hunks |
| 10 | 12 new cases + 3 replacement integration cases | PASS — 12 `it(` counted; 3 added in the sweep test |

No scope creep found: every substantive behavior change traces to a clause in the stated intent.

---

### What was not checked

- **Citations verified inline against branch HEAD `dc964a3`** by re-reading each cited file at that ref. All 10 findings verified; **zero fabricated, zero diff-only, zero false-absent**. Both absence claims (`isSymbolicLink|lstat|realpath` in the guard; re-verification between classify and `fs.rm`) confirmed absent at the ref.
- No tests were executed and no runtime repro was performed — CI is green on all 8 checks for this head SHA, and the `fs.rm`-does-not-follow-symlinks reasoning in finding 1 rests on documented Node semantics, not a live repro in this repo.
- `classifyIgnoredEntry`'s pattern table in `worktree-ignored-patterns.ts` was read as an unchanged dependency but its `opaque` / `inspectable` classifications were not independently audited — finding 1 and the correctness of the walk both inherit whatever that table decides.
- Interaction with the sibling open PRs #799 and #771 on this same file family was not evaluated; the merge order between them is unverified here.

griffinwork40 (commented):

## Merge decision: **MERGE**

Reviewed ref `dc964a3`. Full-regime review (445 lines changed, destructive code path): 2 parallel dimension agents, then centralized citation + absence-claim verification against the ref. No `critical` and no `high` findings. The three findings below are `low`/`nit` and none blocks merge.

The core claim — *"`classifyOrphanDir` never deletes and never throws; every uncertain answer preserves"* — was attacked directly and **holds** on every probe:

| Probe | Result |
|---|---|
| Non-finite / zero / negative age | Preserves. `!Number.isFinite(orphanAgeMs) \|\| orphanAgeMs < minAgeMs` catches `NaN`, `Infinity`, `0`, negatives (`worktree-orphan-guard.ts:86`) |
| Stat failure at the call site | Preserves. `orphanAgeMs` stays `0`, and the new `birthtimeMs > 0` clamp stops a zero birth time from reading as epoch-old (`worktree-sweep.ts:539-548`) |
| Entry budget / depth bound | Both return `remove: false` (`:118-124`, `:141-143`). Budget arithmetic is exact — precisely `maxEntries` entries classified, abort on the next one |
| Throw-safety | `readdir` is wrapped; `join()` / `classifyIgnoredEntry()` take plain `readdir` names and use non-throwing string/regex ops. No escape path found |
| Symlinks | Safe by construction. `entry.isDirectory()` is false for a symlink, so it never enters `pending` (no cycle risk) and is name-classified with **no trailing slash** — and every directory pattern requires one (`worktree-ignored-patterns.ts:59,86`). A symlink named `dist` or `node_modules` therefore falls through to `protected` and **preserves** |
| Call site | Every preserve verdict `continue`s before `fs.rm`; no raw `fs.rm` left on the orphan path. `scope === 'all' \|\| scope === 'interactive'` confirmed untouched (absent from the diff entirely) |

All nine stated-intent claims verified **MET**, including the two easiest to overstate: the guard call precedes the `effectiveDryRun` branch and the warning push is unconditional, so `afk worktree list` really does report preservation (rendered at `src/cli/commands/worktree.ts:116-120`); and the replaced test genuinely asserted the old buggy contract — the string `not registered in git` has zero hits at the ref, replaced by `detects but PRESERVES a freshly-created unregistered directory`. Test counts check out: 12 unit cases + 3 new integration cases.

---

### Findings

**1 · low · confidence high · correctness/observability — `src/agent/worktree-orphan-guard.ts:142` · ref:dc964a3 · citation-type: file-state**
The depth-bound exit reports `because: 'scan-failed'`, the same reason code as a genuine `readdir` I/O failure at `:112`, even though the type declares a dedicated reason whose docstring describes exactly this case ("The walk outgrew its bounds before it could prove the tree disposable", `:51-52`). An operator keying off `because` cannot distinguish "directory unreadable" from "tree too deep"; only the free-text `detail` disambiguates.
```ts
          return { remove: false, because: 'scan-failed', detail: `depth limit at ${rel}` };
```
*Suggestion:* return `because: 'scan-budget-exhausted'` for the depth bound, matching its documented meaning. Safety is unaffected — both outcomes preserve.

**2 · low · confidence high · test-coverage — `src/agent/worktree-orphan-guard.test.ts` · ref:dc964a3 · citation-type: file-state**
No test constructs a symlink entry, though the module's own comment names symlinks as a handled class (`worktree-orphan-guard.ts:148`: "Files, symlinks, sockets — anything not a directory"). Verified absent: `grep -in "symlink|symbolicLink"` on the test file at the ref returns 0 hits. The behavior is safe by construction (analysis above), but that safety rests on an unpinned interaction between `withFileTypes` and the trailing-slash convention in the pattern module — exactly the kind of coupling a future pattern edit could silently break.
*Suggestion:* add one case asserting a symlink named `dist` (pointing anywhere) preserves the tree.

**3 · nit · confidence high · test-coverage — `src/agent/worktree-sweep.test.ts` · ref:dc964a3 · citation-type: file-state**
The `dist/prod.env` scenario the PR body calls out as the load-bearing design point is pinned only at unit level (`worktree-orphan-guard.test.ts:67`). Verified absent at integration level: `prod.env` has 0 hits in `worktree-sweep.test.ts` at the ref; the new integration cases use `node_modules` and a top-level `.env` only.
*Suggestion:* point one existing integration case at `dist/prod.env` so the descend-into-`inspectable` decision is pinned through `runSweep`.

**Not a finding, recorded for completeness:** the new `orphaned-dir-preserved` verdict is absent from all four prunable-verdict allowlists (`src/cli/slash/commands/worktree.ts:41`, `src/agent/daemon/scheduler.ts:505`, `src/cli/commands/worktree.ts:36`, `src/cli/commands/interactive/boot-prune.ts:53`). This is **correct** — a preserve verdict must not be prunable — and all four are membership checks, not exhaustive switches, so nothing breaks at compile time. The operator still sees the state: the verdict string prints in the STATUS column and the reason prints as a `[WARN]` line.

The `opaque` wholesale skip (a secret under `node_modules/` is never seen) is real, but it is the author's explicitly documented trade and it matches the pre-existing shared `classifyIgnoredEntry` policy used on every other removal path — not new behavior this module invents.

---

### What was not checked
- **Citations verified inline against branch HEAD `dc964a3`** via `git show <ref>:<file>` — every `file-state` citation at every severity, plus every absence claim re-run with `git grep` at the ref. Zero fabricated citations; zero false-absent claims; nothing dropped.
- **Stated intent: PR #813 title + body — spec-compliance assessed** (9/9 claims MET).
- Tests were **not executed**; CI on this PR is green across lint/build, ubuntu, macOS, PTY scrollback, docs, and publish smoke. The PR body notes full `pnpm test` was not run locally.
- Runtime behavior on Windows filesystems (birthtime semantics) was not exercised — the Windows CI job is `SKIPPED` on this repo.
- Interaction with the overlapping open PRs #799 and #771 on the same file family was not evaluated; conflicts would surface at merge.
- Wave 2 synthesis was performed inline by the orchestrator rather than dispatched, after two sub-agent transport failures earlier in this run; the two dimension agents ran independently and returned no conflicting claims to arbitrate.

PR #814, opened 2026-07-31
fix(tools): refuse a tool call that keeps failing identically (#723)

### LINKED ISSUE #723: No effective stop on a repeatedly-failing tool at top level: round cap defaults to unlimited, repeat breaker is advisory, loop detector is fork-gated
## Symptom

A single tool (`browser_open`) failed identically across ~9 tool-use rounds in one top-level REPL turn: ~13 minutes, ~$7.29, on `opus_1m`. The turn ended only because the model eventually chose to stop — no guard intervened.

Three separate mechanisms exist that look like they should have caught this. None of them stops a repeatedly-failing tool in a **top-level** session.

## 1. The tool-use round ceiling defaults to unlimited

- `src/agent/providers/shared/tool-loop-cap.ts:43` — `DEFAULT_MAX_TOOL_USE_ITERATIONS = 0`, i.e. no cap.
- `src/cli/shared-helpers.ts:381` — top-level is opt-in only, via `AFK_MAX_TOOL_USE_ITERATIONS`; unset or ≤0 ⇒ unlimited. Wired into the interactive surface in `src/cli/commands/interactive/bootstrap.ts:84`.

Note the asymmetry: subagents get a real default (`SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS`, referenced at `tool-loop-cap.ts:41`). The **interactive** surface — the one where a human is paying per token in real time — is the uncapped one.

## 2. The repeat circuit breaker is advisory and easily defeated

- `src/agent/tools/dispatcher.ts:68` — `REPEAT_CIRCUIT_BREAKER_THRESHOLD = 8`.
- The fingerprint is a hash of tool name + `JSON(input)`, so it requires **8 byte-identical** calls. Any argument variance (a different `timeout_ms`, a re-ordered key) defeats it.
- It is a nudge, not a stop: it returns an error-shaped message asking the model to desist, which the model may ignore while the count keeps climbing.

## 3. The interleaved-pattern loop detector never runs at top level

- `src/agent/tools/dispatcher.ts:657` — `if (this.parentSessionId === undefined) return;`

The `suspected-loop-detector` is (a) observe-only telemetry and (b) gated to forked children. It is documented as deliberate at `dispatcher.ts:644`, but the consequence is that the surface with a human watching and paying gets no loop telemetry at all.

## Proposed fix — smallest first, pick one

1. Give top-level sessions a non-zero default round ceiling (cheapest; pure blast-radius limiter).
2. Escalate the repeat breaker from advisory to **enforcing** after N calls past threshold, and/or key the fingerprint on tool name + a normalized argument subset so trivial jitter cannot defeat it.
3. Ungate the loop detector for top-level sessions in **observe-only** mode first, to measure how often this happens before changing any behavior.

## Explicitly out of scope

Aborting the turn. A hard abort on a false positive is worse than the loop it prevents. The preference is a hard, informative **tool-level refusal** — "this exact call has failed N times and will not be retried; here is why" — leaving the model free to change approach or stop cleanly.

## Acceptance

- [ ] A tool failing identically N times stops being executed and returns a refusal that names the repetition and the prior failure.
- [ ] Test covers argument jitter not defeating the guard.
- [ ] Whatever ships is measured (telemetry) before any turn-terminating behavior is considered.

griffinwork40:
## Summary

A tool that fails the same way forever now stops being executed. At **3 consecutive identical failures** the call is refused without invoking the handler, and the refusal quotes the prior error back.

This is the "hard, informative tool-level refusal" the issue asked for — not a turn abort. The model stays free to change approach or stop cleanly.

## Root cause

Three mechanisms looked like they should have caught the motivating incident (~9 rounds of an identically-failing `browser_open`, ~13 minutes, ~$7.29 on `opus_1m`). None stops a repeatedly-failing tool at top level:

- `src/agent/tools/dispatcher.ts:68` — `REPEAT_CIRCUIT_BREAKER_THRESHOLD = 8`, fingerprinted on `repeatCallFingerprint` (raw `JSON(input)`), so it needs **8 byte-identical** calls; any argument jitter defeats it. And it is a **nudge**, not a stop — `checkRepeatCircuitBreaker` returns an error-shaped message while the tool keeps being callable.
- `src/agent/providers/shared/tool-loop-cap.ts:43` — the round ceiling defaults to unlimited at top level.
- `src/agent/tools/dispatcher.ts:657` — the loop detector is gated to forked children (`parentSessionId === undefined` returns early).

## Change

New module **`src/agent/tools/repeat-failure-guard.ts`** (147 LOC), alongside the existing `repeat-circuit-breaker.ts` rather than inside it — the two mechanisms differ in both signal and consequence, and both stay live:

| | advisory breaker (existing) | this guard (new) |
|---|---|---|
| counts | consecutive calls, any outcome | consecutive **failures** |
| fingerprint | raw `JSON(input)` — jitter resets it | normalized: sorted keys **+ volatile keys stripped** |
| threshold | 8 | 3 |
| consequence | nudge; tool still runs | **refusal; handler not invoked** |

**Why the fingerprint change matters.** `fingerprintToolCall` (reused from `suspected-loop-detector.ts`) already sorts object keys recursively, which covers re-ordering. It does **not** cover a changed *value*, so `timeout_ms: 10_000 → 20_000` still produced a fresh fingerprint. The guard strips a deliberately tiny volatile-key set (`timeout_ms`, `timeoutMs`, `timeout`) before hashing. That list must stay tiny — every key removed makes two genuinely different calls look identical — so only fields that cannot change a call's *meaning* qualify: a deadline governs how long the same operation may run, never what it does. A regression test pins that semantic args (`command`, `url`) still separate fingerprints.

**Why N = 3.** Two attempts is a normal transient retry and frequently succeeds, so firing there would be a false positive; the third identical failure is where the evidence for "this will not work" becomes strong. Against the motivating incident (~9 rounds), 3 would have saved roughly two thirds of the wasted time and spend.

**Dispatcher wiring** — three thin call sites, no logic inline (the file is already 1320 LOC, far over the repo's 350-LOC ceiling):
- `execute()` step 2c-bis and `executeBatch()` phase 1 — pre-execution checks, immediately after the advisory breaker so a short-circuited call is not double-counted.
- `executeCore()` — records every settled result. This is the one choke point both paths funnel through, so the streak sees each executed call exactly once; it runs after the output cap so the quoted error matches what the model was shown.

## Acceptance list

- [x] **A tool failing identically N times stops being executed and returns a refusal that names the repetition and the prior failure.** Refusal message names the tool, the count, and quotes the last error; the handler is never reached.
- [x] **Test covers argument jitter not defeating the guard.** Two tests (changed `timeout_ms`, re-ordered keys) plus a direct fingerprint-equality test.
- [x] **Whatever ships is measured.** New `failureClass: 'repeat-failure'` rides the existing per-call `tool_call` trace event, already consumed by `src/improve/scan/detectors/tool-failure-density.ts`. `TOOL_FAILURE_CLASSES` is a closed const whose zod schema derives from it (`z.enum(TOOL_FAILURE_CLASSES)`), so this is a one-line additive change. Deliberately **no** new `session_phase` name — that union is hand-listed in two places (`trace/types.ts` and the zod enum in `trace/events.ts`), and this file's own output-cap comment records the precedent that widening it for a diagnostic is out of scope.

## Explicitly out of scope

- **Aborting the turn.** Per the issue: a hard abort on a false positive is worse than the loop it prevents. The refusal is tool-level only.
- **A top-level round-ceiling default** (`DEFAULT_MAX_TOOL_USE_ITERATIONS`). Left as a separate decision — it is a blunt blast-radius limiter whose value would need justifying against legitimately long turns, and it is not needed to satisfy the acceptance list.
- **Ungating the loop detector** for top-level sessions.

## Tests

New `src/agent/tools/repeat-failure-guard.test.ts` — 13 cases: permitted below threshold; refused at threshold with `failureClass` set; message names count and quotes the prior error; trips despite changed `timeout_ms`; trips despite re-ordered keys; jittered variants fingerprint identically; semantic args still differ; a repeatedly *succeeding* call is never refused; a success clears the streak (intermittent tools safe); no leak across a different tool; no leak across different arguments; non-object input does not throw; a 5,000-char error is truncated.

## Gates

| Gate | Result |
|---|---|
| `pnpm exec vitest run src/agent/tools/repeat-failure-guard.test.ts` | 13/13 pass |
| `pnpm exec vitest run src/agent/tools/` | **2127 passed**, 2 skipped, 68 files |

## Risk

The real risk is a **false positive on a legitimate retry**: a call that fails three times for a genuinely transient reason and would have succeeded on the fourth is now refused. Three mitigations bound it — any success clears the streak, the count is per-dispatcher (one turn) so nothing persists across turns, the exempt-tool list is honoured, and the refusal is recoverable (the model can change one semantic argument and proceed). The blast radius is one tool call, never the turn.

Fixes #723


griffinwork40:
@codex fix comments

griffinwork40:
Verification complete. All citations verified or confirmed-absent — nothing dropped or downgraded. Confirmed `DEFAULT_MAX_CONCURRENT_SAFE_TOOL_CALLS = 8` (line 95) and `REPEAT_CIRCUIT_BREAKER_THRESHOLD = 8` (line 73) for the perf-window framing.

Highest finding is `medium` — no `critical`/`high`, so the post-synthesis shadow-verify gate does not apply. No blocking/critical citation verification needed. Proceeding to Wave 2 synthesis (inline — findings are few and already verified).

---

# PR #814 Review — `fix(tools): refuse a tool call that keeps failing identically (#723)`

**Ref:** a5c7ec1 · **+484 / −46 · 5 files** · regime: full · · Stated intent: PR title+body — spec-compliance assessed.

## Triage
New enforcing module `repeat-failure-guard.ts` (147 LOC) + dispatcher wiring across all three dispatch sites (`execute()` single, `executeBatch()` phase-1 sequential, `executeBatch()` concurrency-safe wave loop) + a one-line additive `TOOL_FAILURE_CLASSES` enum entry. Refuses a tool call after 3 consecutive identical failures, quoting the prior error, without aborting the turn. Acceptance list fully implemented; turn-abort / round-ceiling / loop-detector-ungating correctly stay out of scope.

## Findings (sorted: medium → low → nit)

### ⚠ CONFLICT — same root behavior, two readings (not auto-resolved)

> **F1 · medium · confidence high · api-compat · dispatcher.ts:1063-1067 · ref:a5c7ec1 · diff-context**
> The wave-admission capacity formula `REPEAT_FAILURE_REFUSAL_THRESHOLD - this.repeatFailureGuard.streakFor(call)` defaults to `3` for a clean (streak-0) fingerprint, so an identical *succeeding* safe fan-out of width >3 now runs at 3-per-wave instead of `maxConcurrentSafeCalls` (default 8) per wave. This invalidates the prior documented `executeBatch` contract ("within the cap this is identical to Promise.allSettled" — that comment block is deleted in this hunk) while the PR intent describes only refusal-after-failures, not throttling of succeeding fan-outs.
> evidence:
> ```js
> const capacity = REPEAT_FAILURE_REFUSAL_THRESHOLD - this.repeatFailureGuard.streakFor(call);
> ```
> suggestion: the report's own first-pass suggestion (`streakFor > 0 ? THRESHOLD − streak : Infinity`) is **unsound** — admitting all N identical calls on a clean streak lets an all-failing wave execute every call before any `note()` fires, defeating the within-batch protection that is the whole point of pre-throttling. A defensible improvement is narrower: keep `capacity = THRESHOLD − streak` (the prescience-free within-batch bound) but add an inline perf note quantifying the trade-off, and consider raising `THRESHOLD` for the success-only case only if profiling shows a real-world regression (none demonstrated here).
>
> **F5 · low · pareto-observability · dispatcher.ts:1063-1067 · ref:a5c7ec1 · diff-context)** *(same location, different dimension)*
> The perf cost of the wave throttle — an N-wide identical *succeeding* safe fan-out now costs `⌈N/THRESHOLD⌉` sequential waves (up to 3× latency at w-3) — is not bounded or documented beyond the inline comment. The advisory breaker (raw `JSON(input)`, byte-identical, jitter-defeatable) does **not** cleanly upper-bound this window, so the regression band is wider than the "~4-7 width" framing: **any** identical succeeding safe fan-out of width >3 pays it.
> evidence: same capacity formula (diff-context, verified).
> suggestion: add an inline perf note quantifying `⌈N/⌉ THRESHOLD⌉` waves for an N-wide identical safe fan-out, and that the real-world surface is narrow (compose nodes have distinct prompts → distinct fingerprints; >3 byte-identical safe calls in one batch is unusual).

**Conflict rationale:** Security reads the throttle as scope-creep/contract-break (medium). Correctness independently reads it as *necessary supporting machinery* ("detecting within-batch failure streaks has no prescience … not strictly scope creep") with an undocumented perf cost (low, same root). Both are right on the *fact*; they diverge on interpretation. The prescience argument (correctness) weakens the scope-creep framing — the throttle is what makes within-batch streak detection possible — so the medium should be read as **"behavior change worth documenting/justifying,"** not as breakage. Non-blocking.

### F2 · low · confidence high · correctness · dispatcher.ts:1096-1119 (diff-context) + :1307 (file-state, verified) · ref:a5c7ec1
The relocated wave-loop rejection handler synthesizes `{ content, isError: true }` with **no `failureClass`**; the post-wave `note()` loop only excludes `failureClass !== 'abort'`, so *if* `executeCore` ever rejects under a future refactor, that rejection would both (a) advance the failure streak (treated as a tool failure) and (b) skip `firePostToolUseFailure` — a latent interaction the new note loop's guard introduces. Unreachable today because `executeCoreInner`'s body catches internally (`try {` at dispatcher.ts:1307, file-state verified), and the existing comment (1100-1101) documents the invariant — but it does not flag the `note()` interaction.
evidence:
```js
results[executableCalls[batchIdx]!.originalIndex] = { content: `Tool execution error: ${msg}`, isError: true };
// …later, with no failureClass on the above:
if (result !== undefined && result.failureClass !== 'abort') { this.repeatFailureGuard.note(call, result); }
```
suggestion: stamp a sentinel `failureClass` (`'abort'` or a dedicated `'handler-reject'`) on the synthesized rejection result so the note loop skips it, restoring the "refused/rejected calls never advance the streak" invariant.

### F3 · low · confidence high · test-coverage · dispatcher.test.ts:635-687 · ref:a5c7ec1 · diff-context
No test drives the single `execute()` path through `THRESHOLD+` identical failures to assert `failureClass: 'repeat-failure'` there. Both new dispatcher tests (635-656, 658-687) exercise `executeBatch`; `repeat-failure-guard.test.ts` tests `note`/`check` as pure units. The `execute()` wiring (877-894: `checkRepeatFailureGuard` → `executeCore` → `note`) is the primary dispatch path yet its refusal is only covered structurally. (Absence confirmed: grep of `repeat-failure|REPEAT_FAILURE` in dispatcher.test.ts returns hits only at import line 13 and the batch block 644-686.)
suggestion: add a test looping `dispatcher.execute(makeCall())` with a failing handler `THRESHOLD+1` times, asserting the Nth result has `failureClass: 'repeat-failure'` and the handler was called exactly `THRESHOLD` times — mirroring the advisory breaker tests at dispatcher.test.ts:1464-1576.

### F4 · low · confidence high · test-coverage · dispatcher.ts:1121-1137 (diff-context) · ref:a5c7ec1
The non-concurrency-safe `else` branch's guard wiring (`checkRepeatFailureGuard` → `executeCore` → `note`) is untested: the two new batch tests use `echo` (classified concurrency-safe), so they hit the wave branch. (Absence confirmed: grep of `concurrencyClassifier` in dispatcher.test.ts returns exactly one hit — line 972 `() => true`; no `() => false` override forces a multi-call unsafe batch through the guard.)
suggestion: add a test with `concurrencyClassifier: () => false` issuing `THRESHOLD+1` identical failing unsafe calls through `executeBatch`, asserting refusal at index `THRESHOLD`.

### F5 · nit · confidence high · test-coverage · dispatcher.test.ts:1082-1098 · ref:a5c7ec1 · diff-context
The test title "runs the whole batch concurrently when the cap exceeds batch width" now contradicts its assertion — `expect(state.peak).toBe(REPEAT_FAILURE_REFUSAL_THRESHOLD)` proves the *opposite* (4 identical calls throttle to 3-wide, not concurrent).
suggestion: rename to "throttles identical safe fan-outs to the repeat-failure threshold regardless of the wider concurrency cap."

## Spec-compliance assessment
**Clean.** All acceptance clauses met: refusal names the tool, count, and quoted prior error and never reaches the handler (proven by `toHaveBeenCalledTimes(THRESHOLD)` at dispatcher.test.ts:651); jitter test exists (repeat-failure-guard.test.ts:65-89); `failureClass: 'repeat-failure'` rides the existing `tool_call.completed` event (additive enum, derives into `z.enum(TOOL_FAILURE_CLASSES)`), and both downstream consumers branch by set-membership not exhaustive switch (so the new class doesn't compile-break anything) — no new `session_phase`. Out-of-scope items (turn abort, default round ceiling, ungating the loop detector) correctly absent.

## Epistemic scope disclosure
- Citations verified inline against branch HEAD **a5c7ec1** (this is the checked-out ref; reads were live file reads). File-state citations (`repeat-failure-guard.ts:130-138`, `dispatcher.ts:1307`, `dispatcher.ts:95`) all verified present. No fabricated or diff-only citations dropped.
- Absence claims F3, F4 confirmed-absent via grep against the reviewed ref (`repeat-failure|REPEAT_FAILURE` and `concurrencyClassifier`).
- Did NOT re-run the test suite, did not execute the dispatcher, did not check every tool's zod input schema for a key literally named `timeout` used semantically (the three-key volatile set is documented as must-stay-tiny, bounded by the 3-failure + reset-on-success design).
- **Stated intent: PR #814 title+body — spec-compliance assessed.**

## Merge recommendation

**MERGE**

The change is well-designed, well-tested (13 guard unit cases + 2 dispatcher integration cases), and faithfully implements the acceptance list with no scope creep into turn-aborts or round ceilings. The single `medium` (wave-throttle of succeeding identical safe fan-outs) is a **deliberate, prescience-free within-batch streak bound**, not a contract break — the inline rationale already documents "why waves," and the reviewed-suggested narrowing (`Infinity` on clean streak) would *defeat* the within-batch protection, so it is not a recommended change. The remaining findings are low/nit: one latent-invariant safeguard worth a one-line stamp, two small test-path coverage gaps on the `execute()` single path and the unsafe-batch branch, one test-title nit, and a perf-bound note to add. None block merge; the author is encouraged to address F2 (rejection `failureClass` stamp) and F5 (perf note) as a follow-up.

(Review is read-only — no files, commits, or PR state were mutated. The single uncommitted working-tree change was preserved.)

---

Issue #782, opened 2026-07-30
refactor: input-parse.ts (528) and bash-restriction-hook.ts (584) are over the 350-line ceiling — extract two named concerns


griffinwork40:
Raised in review of #753 and deliberately deferred — a structural refactor inside an under-review security PR is scope creep, and both files were already over the ceiling *before* that PR arrived.

## Current state (`main` @ `2675eaf`)

| file | lines | on `main` before #753 |
|---|---|---|
| `src/agent/tools/hooks/bash-restriction-hook.ts` | 584 | ~495 |
| `src/agent/tools/subagent/input-parse.ts` | 528 | 422 |

Project rule: no source file over 350 lines; at the ceiling, pull one **whole concern** into a new file rather than shaving lines.

Note this is not mechanically enforced in this repo today — `lint` is `tsc --noEmit` and there is no size-check script (unlike some sibling projects). Adding one is arguably the more durable half of this issue.

## Extraction candidates (both named by review)

**`input-parse.ts` → `src/agent/tools/subagent/root-validation.ts`**
The path-root breadth concern: `isTooBroadRoot`, `HOME_TARGETS`, `afkBreadthTargets`, and the realpath helper they share. Called from exactly three sites (`cwd`, `writeRoots`, `readRoots`), so the seam is clean.

**`bash-restriction-hook.ts` → `src/agent/tools/hooks/afk-home-refs.ts`**
The AFK_HOME spelling concern: `configuredAfkHome`, `relocatedAfkSensitiveRoots`, and the relocated forms inside `allowlistedFileForms`.

## Constraint

These are **pure moves** — zero logic change. Every existing test must pass unchanged; that is the gate. `bash-restriction-hook.test.ts` (75 tests) and `input-parse.test.ts` (105 tests) both cover the affected surfaces well, so the refactor is verifiable.

If either extraction cannot be made confidently behavior-preserving, leave that file alone and say why. A correct security floor beats a tidy file.

## Precedent set in #753

The new warn helper was added as its own module (`src/agent/tools/afk-home-warn.ts`, 57 lines) specifically so it would not grow either of these two files further. Same instinct, applied incrementally.

---

*Citation anchor corrected: line numbers below refer to `main` @ `2675eaf` (the squash merge of #753). They were originally written against branch `2aa5fbb`; `read-denylist.ts` shifted ~63 lines and `bash-restriction-hook.ts` ~8 lines because #735 landed alongside #753.*

PR #815, opened 2026-07-31
fix(permissions): gate the relocated mcp.json carve-out on builtin denies (#779)

### LINKED ISSUE #779: read denylist: relocated mcp.json carve-out is unconditional, unlike its bash twin which gates on isReadDenied
Found during review of #753.

## What

The relocated `mcp.json` carve-out exists in two places, and only one of them checks whether the carve-out target is itself denied by another builtin entry.

**Bash surface — gated.** Each form is dropped when it is denied:

```ts
// src/agent/tools/hooks/bash-restriction-hook.ts:386-397 (allowlistedFileForms)
if (isReadDenied(path.join(home, rel)).denied) return [];
…
if (isReadDenied(relocated).denied) return [];
```

**Typed surface — ungated.** The relocated spellings are emitted unconditionally:

```ts
// src/agent/tools/handlers/read-denylist.ts:288 (derivedAfkHomeAllowEntries)
return READ_ALLOWLIST_REL.filter((rel) => rel.startsWith(AFK_HOME_REL_PREFIX)).map((rel) =>
  resolveExceptionEntry(join(afkHome, rel.slice(AFK_HOME_REL_PREFIX.length))),
);
```

and are unioned into `allow` at `read-denylist.ts:329`, which `isReadDenied` consults **before** the builtin denylist.

## Consequence

If the AFK home is nested under another builtin denied root — e.g. `AFK_HOME=~/.ssh/afk` — the derived carve-out punches an exact-file hole through the `.ssh` floor on the typed read path, while the bash surface correctly refuses it. The two surfaces disagree, and the typed one is the looser.

## Reachability

Narrow: requires an operator to relocate `AFK_HOME` *underneath* an existing denied root. That is an odd configuration, which is why this was filed rather than fixed in #753. It is a real asymmetry regardless, and the bash side already demonstrates the intended rule.

## Fix sketch

Filter the derived allow entries against the builtin deny prefixes before unioning them into `allow`, mirroring the `isReadDenied` gate the bash twin already applies. Note the ordering constraint: the filter cannot naively call `isReadDenied` from inside `resolveLists()` without re-entrancy care, since that is the function building the list it would consult.

## Scope note

The carve-out is exact-file (`allow.includes(real)`), never a prefix, and operator `AFK_READ_DENYLIST` extras still outrank it — both verified in #753's review. Only the builtin-deny interaction is at issue here.

---

*Citation anchor corrected: line numbers below refer to `main` @ `2675eaf` (the squash merge of #753). They were originally written against branch `2aa5fbb`; `read-denylist.ts` shifted ~63 lines and `bash-restriction-hook.ts` ~8 lines because #735 landed alongside #753.*

griffinwork40:
## Summary

The `AFK_HOME`-relocated `mcp.json` carve-out was emitted unconditionally and unioned into the allow list that `isReadDenied` consults **before** the builtin denylist. With the home relocated underneath another builtin denied root, that carve-out punched an exact-file hole straight through the outer floor.

## Root cause

- **Typed surface, ungated.** `read-denylist.ts:288` (`derivedAfkHomeAllowEntries`) maps every `.afk/`-prefixed carve-out to its relocated spelling with no deny check, and `:329` unions the result into `allow`.
- **`allow` wins first.** `isReadDenied` returns `{ denied: false }` on `allow.includes(real)` *before* the builtin deny loop runs. So the derived entry outranks every builtin root, including one it was never meant to pierce.

### Correction to the issue: the bash surface was **also** unguarded

The issue frames bash as correctly gated, citing `bash-restriction-hook.ts:386-397`:

```ts
if (isReadDenied(path.join(home, rel)).denied) return [];
```

That gate uses `isReadDenied` as its **oracle**, and the oracle honoured the ungated allow list — so for exactly the case in question it answered "not denied" and the gate never fired. Both surfaces were open; they merely looked different. Fixing the derivation closes both at once, which is why this PR touches no hook code. (A prior investigation pass flagged this empirically; verified here by reading the short-circuit order in `isReadDenied`.)

## Change

New module **`src/agent/tools/handlers/read-denylist-carveout.ts`** (52 LOC) exporting `gateDerivedCarveOuts()`, called from `resolveLists()` before the union. It lives in its own file because `read-denylist.ts` is already **425 LOC**, over the repo's hard 350-LOC ceiling.

Two constraints shape it, both documented in-module:

1. **The gate cannot call `isReadDenied`.** That function reads the list being built — asking it would re-enter *and* always answer "not denied" for the entry under test. The gate matches the builtin deny prefixes directly.
2. **The exclusion is by EQUALITY, never prefix.** The carve-out legitimately pierces `<afkHome>/config`, so that root is excluded from the gate by exact match. A root that merely *contains* the relocated home is a prefix, not an equal, so it stays in the gate and still denies. Prefix-based exclusion would re-open the hole; equality-based *gating* would miss it. A dedicated test pins this discriminator.

Fails closed: an entry under any other builtin root is dropped, never kept.

## Scope

Unchanged by design, and each pinned by a test:

- carve-out stays **exact-file** (`allow.includes(real)`), never a prefix;
- operator `AFK_READ_DENYLIST` extras **still outrank** it;
- the **default `AFK_HOME`** case behaves exactly as before;
- `bash-restriction-hook.ts` (584 LOC, subject of #782) is **not** refactored.

## Tests

New `read-denylist-carveout.test.ts` — 12 cases. Pure gate: keeps a carve-out whose only containing root is the pierced one; drops one under another builtin root; keeps one under a relocated-but-undenied home; equality-not-prefix discriminator; a sibling sharing a name prefix is not treated as contained; empty carved-root set fails closed. Integration via `isReadDenied`: relocated-under-a-floor is denied; the verdict names the **outer floor** as `matched`; default home still allowed; relocated-but-undenied still allowed; sibling `afk.env` still denied; operator extras still outrank.

Env relocation follows the existing #740 pattern — reassigning `AFK_HOME` rebuilds the `resolveLists()` memo (its key includes `AFK_HOME`), so no module reload is needed.

## Gates

| Gate | Result |
|---|---|
| `pnpm exec vitest run read-denylist-carveout.test.ts` | 12/12 pass |
| `read-denylist.test.ts` + `bash-restriction-hook.test.ts` + new file | **145 passed** (3 files) |

## Reachability and severity

Narrow, and worth stating plainly: this requires an operator to relocate `AFK_HOME` *underneath* an existing builtin denied root. That is an odd configuration, which is why it was filed out of #753's review rather than fixed inside it. The value here is removing a surface disagreement in a security floor and making the intended rule the one that actually executes — not closing a widely-reachable hole.

Over-restriction risk is low but real: a carve-out is now dropped when its path sits under another builtin root. For the default home and any relocation outside the denied roots, nothing changes (both pinned by tests).

Fixes #779


griffinwork40:
@codex fix comments

griffinwork40 (commented):

Reviewed ref `0d7a3b6832076f51520f87b9e89380c5d5037a3a` · base `main` · 3 files, +204/-6 · light regime (hotfix).

CI is red on this PR (2 test files, 5 tests failed; 14276 passed), and the failures are **caused by this diff**, not inherited. `main` is green, and the two failing suites — `read-denylist.test.ts`, `bash-restriction-hook.test.ts` — are exactly the ssh carve-out suites this change's gate touches. Both high findings below were independently re-derived by a second pass and returned CONFIRMED.

---

### F1 · high · correctness / security floor (over-restriction) · `src/agent/tools/handlers/read-denylist.ts:357-360` · citation-type: diff-context

The diff routes the **static** `BUILTIN_READ_ALLOWLIST` through the new gate. Pre-diff that list was unioned into `allow` ungated — only the AFK_HOME-derived entries were the subject of #779.

```ts
        ...gateDerivedCarveOuts(
          BUILTIN_READ_ALLOWLIST.map(resolveExceptionEntry),
          defaultGateRoots,
        ),
```

`defaultGateRoots` (`:339-344`) keeps every resolved builtin root whose **source string** is not `~/.afk/config`, so `~/.ssh` (`:62`) stays in the gate set. Post-merge with `main`, `READ_ALLOWLIST_REL` also carries `.ssh/config` and `.ssh/known_hosts` (`read-denylist.ts:181-182` on main; only `.afk/config/mcp.json` at this ref, `:169`). Trace for `<home>/.ssh/config`: `resolveExceptionEntry` (`:203`) yields `<home>/.ssh/config` → `isUnderRoot` against `~/.ssh` gives `relative() === "config"` → contained → **dropped from `allow`**. `isReadDenied` then misses its short-circuit (`:432`) and the builtin loop denies (`:434`). That reproduces the CI error string verbatim:

```
refusing to read protected path: /home/runner/.ssh/config (matches read-denylist entry: /home/runner/.ssh)
```

`mcp.json` survives only via the `source === DEFAULT_AFK_CONFIG` equality exemption (`:336-338`) — which is why the branch is green in isolation and red merged. This is a **semantic merge conflict**: textually clean, broken only in combination. Direction of failure is fail-closed (a legitimate read is denied), not a bypass — so it is `high`, not `critical`. It silently reverts a deliberate carve-out on every default install, across both the typed-read and bash surfaces.

**Suggestion:** gate only the derived entries. Restore `...BUILTIN_READ_ALLOWLIST.map(resolveExceptionEntry)` ungated and drop `defaultGateRoots`; #779's root cause is `derivedAfkHomeAllowEntries()` alone.

---

### F2 · high · spec-compliance · `src/agent/tools/handlers/read-denylist-carveout.test.ts:89-91` · citation-type: diff-context

The PR body asserts the default `AFK_HOME` case "behaves exactly as before", pinned by a test. The only default-home test hardcodes the one carve-out the gate exempts:

```ts
  it('still allows the carve-out for the default AFK_HOME', () => {
    const target = join(homedir(), '.afk', 'config', 'mcp.json');
    expect(isReadDenied(target).denied).toBe(false);
```

`READ_ALLOWLIST_REL` is never imported or iterated in the new file — every path is spelled literally, so the suite is structurally incapable of catching a carve-out whose containing root is not `~/.afk/config`. Scoped to this ref in isolation the default case genuinely is unchanged; the claim is falsified **as merged**, which is what the unqualified assertion promises. The red CI in the pre-existing suites is the empirical proof.

**Suggestion:** iterate the exported `READ_ALLOWLIST_REL` constant in the default-home test so a newly added carve-out cannot silently regress.

---

### F3 · medium · spec-compliance (scope creep) · `read-denylist.ts:339-350` · diff-context
Stated intent scopes the gate to the relocated `mcp.json` carve-out; the diff builds a second gate set and applies it to the static builtin allowlist, extending a security floor to inputs the PR body never names. **Suggestion:** delete `defaultGateRoots`, keep `derivedGateRoots`.

### F4 · medium · correctness · `read-denylist.ts:171` + `:75` · diff-context / file-state
`DEFAULT_AFK_CONFIG` relies on byte-identity with a separately duplicated literal, compared before `resolve()`/`safeRealpath`. Any reformat of the denylist entry makes `defaultCarvedRoot` `undefined`, leaves `~/.afk/config` in the gate set, and drops the `mcp.json` carve-out at the default home. No test references `DEFAULT_AFK_CONFIG`. **Suggestion:** hoist the constant above `BUILTIN_READ_DENYLIST` and reference it at `:75`.

### F5 · medium · correctness (platform) · `read-denylist-carveout.ts:37` vs `read-denylist.ts:434` · diff-context / file-state
The gate's containment test is `sep`-aware; the floor it defends hardcodes `'/'`. On Windows the gate is stricter than the denylist — dropping carve-outs the floor would never have matched, giving denial with no covering deny entry. **Suggestion:** share one containment helper.

### F6 · low · correctness (contract) · `read-denylist-carveout.ts:47-54` · diff-context
The "Fails CLOSED" docstring is inverted for empty `gateRoots`: `.some()` on `[]` is `false`, so every entry is kept. Not reachable from current call sites. **Suggestion:** return `[]` on an empty gate set, or narrow the doc.

### F7 · low · perf-observability · `read-denylist-carveout.ts:50-55` · diff-context
Dropped carve-outs are discarded silently; the sibling failure paths in this module do warn (`warnAfkHomeRejectedOnce`, `read-denylist.ts:265,298`). An over-restriction surfaces only as a deny naming the outer root. **Suggestion:** warn once per process naming the dropped entry and gating root.

### F8 · low · repo rule · `read-denylist.ts:457` · file-state
The file is **457** lines at this ref (440 on main) — the diff grew an already-over-ceiling file by 17 lines against the 350-LOC hard rule. The body cites 425, understating by 32. **Suggestion:** move the gate-set construction into the new module.

---

### Per-dimension roll-up
- **security:** F1, F7 — regression direction is over-restriction, not bypass. Extras-outrank-carve-out precedence intact and pinned (`read-denylist-carveout.test.ts:106`).
- **api-compat:** no issues found — `gateDerivedCarveOuts` is new, one production importer (`read-denylist.ts:49`), no signature removed.
- **correctness:** F1, F4, F5, F6, F8.
- **spec-compliance:** F2, F3.
- **test-coverage:** F2, F4, F7 — 13 `it()` blocks (body says 12), all green at the ref in isolation; none covers a non-`.afk` carve-out.
- **perf-observability:** F7. Gating runs once per memo rebuild in `resolveLists` (`:351-364`), so the hot read path is unaffected.

### What was not checked
Citations verified inline against branch HEAD `0d7a3b68`; all quoted lines confirmed present at the ref, and three absence claims re-checked by grep at the ref. Not read: `read-denylist.test.ts` and `bash-restriction-hook.test.ts` sources (failures read from the CI log only); `safeRealpath`/`getAfkHome` implementations; `at-file-inject.ts` as a second consumer of `READ_ALLOWLIST_REL`. No merge commit was materialized — the merge behavior is derived from main's `READ_ALLOWLIST_REL` plus the failing test names, not from an executed merge. Tests were not run locally. Stated intent: PR #815 title+body — spec-compliance assessed.

griffinwork40 (commented):

One blocking regression. The fix for #779 is real and correctly reasoned, but the gate is applied to the **static** built-in allowlist as well as the derived one, and it strips the `~/.ssh/config` / `~/.ssh/known_hosts` carve-outs. CI on this exact ref (`208342d4`) is red on both ubuntu-latest and macos-latest with 5 failures, all in the two files this change affects.

---

### 🔴 high · security · correctness — the static ssh carve-outs are gated out of `allow` and become denied

`src/agent/tools/handlers/read-denylist.ts:355-360, 373-376` · ref:`208342d4` · citation-type: diff-context

`defaultGateRoots` excludes exactly one root — the one whose `source` equals `DEFAULT_AFK_CONFIG`. `~/.ssh` is a different source string, so it stays in the gate set. `BUILTIN_READ_ALLOWLIST` is then passed through `gateDerivedCarveOuts()` against that set, and the two ssh entries sit under `~/.ssh`, so they are filtered out of `allow`.

```ts
const defaultGateRoots = [
  ...resolvedBuiltinEntries
    .filter(({ source }) => source !== DEFAULT_AFK_CONFIG)
    .map(({ root }) => root),
```
```ts
...gateDerivedCarveOuts(
  BUILTIN_READ_ALLOWLIST.map(resolveExceptionEntry),
  defaultGateRoots,
),
```

Symbolic trace, re-derived independently and confirmed:
1. `~/.ssh` ∈ `defaultGateRoots` — the filter tests `source !== DEFAULT_AFK_CONFIG`, and `~/.ssh` is not that source.
2. `isUnderRoot('~/.ssh/config', '~/.ssh')` → `relative()` returns `'config'` → not `''`, not `'..'`, no `../` prefix, not absolute → `true`.
3. `gateDerivedCarveOuts` keeps an entry only when no gate root matches, so both ssh entries are dropped from `allow`.
4. With them absent from `allow`, `isReadDenied`'s allow short-circuit no longer fires, the builtin deny loop runs, and `~/.ssh` matches — producing exactly the observed `refusing to read protected path: /Users/runner/.ssh/config (matches read-denylist entry: /Users/runner/.ssh)`.
5. `.afk/config/mcp.json` is unaffected: the root it pierces *is* the excluded one. **The bug is partial — the #779 fix itself works; the ssh carve-outs are collateral damage** from routing all three static entries through a single gate set tailored to exclude only `.afk/config`.

Because the bash surface uses `isReadDenied` as its oracle — as this PR body itself documents — the regression propagates to the bash hook automatically, which is why three `bash-restriction-hook.test.ts` parity cases (#579 O2) fail without any hook code being touched.

**Direction matters:** this is **over-restriction, not a hole.** No secret exposure, no bypass, no auth consequence — legitimate reads are blocked. Per the severity rubric that keeps it at `high`, not `critical`.

**Suggestion:** gate each allowlist entry against the roots it does *not* legitimately pierce, rather than one shared `defaultGateRoots`. Simplest correct form: exclude from the gate set every builtin root that an entry in `READ_ALLOWLIST_REL` is carved out of (`~/.afk/config` **and** `~/.ssh`), or stop routing the static `BUILTIN_READ_ALLOWLIST` through the gate at all — it is not the surface #779 is about.

---

### 🟠 high · spec-compliance — scope creep beyond the stated intent

`src/agent/tools/handlers/read-denylist.ts:373-375` · ref:`208342d4` · citation-type: diff-context

The PR title and body scope the change to the **`AFK_HOME`-relocated** carve-out. The diff also routes the static `BUILTIN_READ_ALLOWLIST` — which has nothing to do with relocation — through the identical gate. That unrequested extension is precisely where the regression lives; the derived-entry gate (`derivedGateRoots`) is correct.

**Suggestion:** drop the static-list gating from this PR, or split it into its own change with its own tests.

---

### 🟠 medium · spec-compliance — the Gates table does not match CI on this ref

The body reports `read-denylist.test.ts` + `bash-restriction-hook.test.ts` + the new file as **145 passed (3 files)**. CI at `208342d4` reports `Test Files 2 failed | 769 passed (771)` and `Tests 5 failed | 14669 passed` — with the failures in exactly two of those three named files. The stated claim "the default `AFK_HOME` case behaves exactly as before" is contradicted by the same run: the failing assertions involve no relocation.

**Suggestion:** re-run the named gate command against this SHA and restate the counts.

---

### 🟡 low · test-coverage — the new suite never touches the path that regressed

`src/agent/tools/handlers/read-denylist-carveout.test.ts` · ref:`208342d4` · citation-type: diff-context

The 12 new cases contain zero references to `ssh` or `known_hosts` (grepped at the ref — zero matches). The new file exercises `gateDerivedCarveOuts` as a pure function and the relocated-home integration, but never `gateDerivedCarveOuts` over `BUILTIN_READ_ALLOWLIST` — the one call that broke. The pre-existing suite caught it, in CI, after the fact.

**Suggestion:** add `expect(isReadDenied(join(homedir(), '.ssh', 'config')).denied).toBe(false)` to the new file so this is a pre-merge gate.

---

### ✅ What holds up

- The #779 root-cause analysis is correct and unusually well documented — `allow` short-circuiting ahead of the builtin deny loop is the real mechanism.
- The equality-not-prefix discriminator in `isUnderRoot` is implemented correctly and pinned by a dedicated test; using `path.relative` + `sep` rather than string prefixes is the right call for Windows.
- The correction to the issue's framing — that the bash surface was *also* unguarded because it consults `isReadDenied` as its oracle — is accurate, and this review's own trace independently reproduces it.
- Putting the gate in a new 55-LOC module rather than growing the already-425-LOC `read-denylist.ts` respects the repo's file-size ceiling.
- perf-observability: no issues — the added passes run only inside the memoized `resolveLists()` cache-build path.

### What was not checked

- Citations verified inline against branch HEAD `208342d48beb0ba8256074fd960aa92cc37fb25e`; all cited lines re-read at that ref and confirmed present. The absence claim (no ssh coverage in the new test file) was re-verified by grep at the ref: confirmed absent.
- The two `high` findings were re-derived from scratch by an independent shadow verifier: **CONFIRMED**, with scope narrowed to partial (mcp.json unaffected).
- Reachability: `isReadDenied` has 8+ production importers at the ref (`glob.ts`, `bash-restriction-hook.ts`, `path-approval-hook.ts`, `at-file-inject.ts`, `input-parse.ts`, …) — not a test-only surface.
- Tests were not executed locally; CI output for run `30833701035` was read, not reproduced.
- The `pnpm lint` pass claim is unverified — the Lint & Build check is green in CI, but a green lint does not speak to the runtime regression above.
- Stated intent: PR #815 title + body — spec-compliance assessed against it.

griffinwork40 (commented):

Reviewed ref `10c35ecd`. The #779 hole is genuinely closed on the reported path, and the prior round's `.ssh/*` carve-out regression is genuinely fixed — I verified that `CARVEOUT_PIERCED_SOURCE`'s values are byte-identical template literals to the `BUILTIN_READ_DENYLIST` `source` strings, so the `source !== piercedSource` exclusion actually fires for all three entries. Truth table over `AFK_HOME` unset / `=~/.gnupg/afk` / `=~/elsewhere` produced no hole and no over-restriction. Highest severity found is `medium`; nothing here blocks.

---

### medium · confidence: high · security/correctness — `read-denylist-carveout.ts:38-40` (`ref:10c35ecd`, diff-context)

The gate's containment test is **case-sensitive** while the denial path is **case-insensitive** on macOS/Windows volumes, so a case-variant `AFK_HOME` nested under a builtin denied root is admitted into `allow` for a path the deny loop would otherwise catch — re-opening #779 through case.

```ts
const rel = relative(root, candidate);
return rel === '' || (rel !== '..' && !rel.startsWith(`..${sep}`) && !isAbsolute(rel));
```

`isUnderRoot` uses bare `relative()`; `isReadDenied` denies via `pathIsWithin` (`fs-case.ts:83-90`, file-state), which retries both sides lowercased when `isCaseInsensitiveFs`. Mechanically confirmed with the two predicates side by side: for root `/base/aaa`, candidate `/base/AAA/x/y.json` gives `isUnderRoot = false` but `pathIsWithin = true` — the gate keeps an entry the floor would have denied.

**Correcting one mitigation claim** raised during review: `safeRealpath` does *not* launder this, because macOS `realpathSync` preserves caller-supplied case even for paths that already exist (probed: `/users/griffinlong` → `/users/griffinlong`, `/USERS/griffinlong` → `/USERS/griffinlong`; only symlinks like `/tmp` → `/private/tmp` are rewritten). So the divergence is not self-healing when the relocated tree pre-exists.

Reachability is narrow — it needs an operator to relocate `AFK_HOME` under a builtin denied root *and* spell it with variant case, i.e. strictly narrower than #779 itself. Rated `medium` on that basis, not `high`.

**Suggestion:** route `isUnderRoot`'s containment test through the repo's existing `pathIsWithin` so the gate's case semantics cannot drift from the denylist's.

### medium · confidence: medium · spec-compliance — `read-denylist.ts:366-372` (`ref:10c35ecd`, diff-context)

The fix also routes the three **static** carve-outs through the new gate, widening scope from #779 (which was purely about the *derived* `AFK_HOME` entries) into logic that now decides default-home readability of `.afk/config/mcp.json`, `.ssh/config`, and `.ssh/known_hosts` for every user.

```ts
...gateStaticCarveOuts(
  READ_ALLOWLIST_REL.map((rel) => ({
    rel,
    resolved: resolveExceptionEntry(`${homedir()}/${rel}`),
  })),
```

This is **non-regressing today** — verified byte-identical pierced-source spellings, plus the new sweep test at `read-denylist-carveout.test.ts:122` pins all three entries as allowed under the default home. Flagging blast radius, not a live break: a future `READ_ALLOWLIST_REL` addition without a matching `CARVEOUT_PIERCED_SOURCE` mapping now fails **closed** (silently denied) rather than being ungated, and that sweep test is the single thing standing between such an omission and a shipped denial.

**Suggestion:** record the static re-gating in the PR description as intended scope, since #779 named only the derived entries.

### low · confidence: high · test-coverage — `read-denylist-carveout.ts:82-130` (`ref:10c35ecd`, diff-context)

`gateStaticCarveOuts` — the actual fix for the `.ssh/*` regression — has no direct unit test. Verified at the ref: its only references are `read-denylist.ts:49` (import) and `:367` (call site); the test file imports only `gateDerivedCarveOuts`. `gateRootsExcludingSource` is module-private and equally untested directly. Coverage is entirely transitive through `isReadDenied` under the default `AFK_HOME`, so the per-entry mapping is never exercised for a *relocated* home.

**Suggestion:** add a direct test for `gateStaticCarveOuts` that isolates the exact-string pierced-source removal from `isReadDenied`.

### low · confidence: medium · test-coverage — `read-denylist-carveout.test.ts:36` (`ref:10c35ecd`, diff-context)

The case titled `excludes the pierced root by EQUALITY, not by prefix` passes the same inputs and asserts the same `[]` as the case at `:26`, so the discriminator the module docstring calls load-bearing is not actually exercised. A real discriminator needs a gate root that is a *strict ancestor* of the pierced root — where prefix-exclusion and equality-exclusion give different answers. Verified across all seven `gateDerivedCarveOuts(` call sites in the file: no such case exists.

**Suggestion:** replace `:36` with a case whose gate root strictly contains the pierced root, so prefix-based exclusion would fail it.

### low · confidence: high · test-coverage — `read-denylist-carveout.ts:50-55` (`ref:10c35ecd`, diff-context)

The documented fail-**open** edge case ("an EMPTY `gateRoots` excludes nothing, so every entry passes through kept") is untested. Verified: all seven call sites pass a non-empty array; `:49`'s `fails closed when the carved root remains in the gate set` passes `[join(home, '.afk', 'config')]`, not `[]`.

**Suggestion:** pin it with `gateDerivedCarveOuts([entry], [])` → `[entry]`, or drop the branch if it stays unreachable.

### low · confidence: medium · api-compat — `read-denylist.test.ts:324,338` (`ref:10c35ecd`, file-state)

`resolveLists()` now builds `allow` from `READ_ALLOWLIST_REL` via `gateStaticCarveOuts` and no longer consults `BUILTIN_READ_ALLOWLIST`. Verified at the ref: that export's only remaining references are those two test loops plus a JSDoc `{@link}`. Those assertions now pin a constant disconnected from the shipped allow list and would keep passing if `gateStaticCarveOuts`'s output diverged from it. Tempered by the new sweep test at `read-denylist-carveout.test.ts:122`, which does exercise the real wiring.

**Suggestion:** re-point those two loops at the `resolveLists()`-produced allow list.

### nit · confidence: high · docs — `read-denylist-carveout.test.ts:18-122`

PR body says "12 cases"; the file has **14** `it(` blocks (7 pure-gate + 6 integration + 1 sweep). Cosmetic, but the sweep case is the one added by the prior review round and is worth naming accurately.

---

### Out of scope (pre-existing, not introduced here)

`resolveLists()`'s memo key (`read-denylist.ts:330`) covers `AFK_READ_DENYLIST` and `AFK_HOME` but not `homedir()`-derived state, so a mid-process `HOME` reassignment reuses stale gate roots. Pre-existing pattern; in practice guarded by Vitest per-file isolation (no `isolate:false` found), and `read-denylist.test.ts` never reassigns raw `HOME`.

### What was not checked

- Citations verified inline against branch HEAD `10c35ecd` (fetched locally); every `blocking`/`high` and every `file-state` citation was re-read at that ref, and all three absence claims ("no direct `gateStaticCarveOuts` test", "no empty-`gateRoots` test", "no `_resetReadDenylistCacheForTests` in the new file") were confirmed by `git grep` at the ref.
- **Tests were not run.** CI is green on this PR across Lint & Build, Test (ubuntu/macOS), real-PTY scrollback, Docs Site, and Publish Bundle Smoke. Full `pnpm test` was not run locally by either the PR author or this review.
- The case-divergence finding was verified by re-implementing both predicates side by side, **not** by executing the shipped module or adding a failing test.
- Windows behavior is inferred from `isCaseInsensitiveFs` semantics only; the Windows CI job is `SKIPPED` on this run.
- Not reviewed: `bash-restriction-hook.ts` (unchanged by design; the claim that it is fixed transitively via the `isReadDenied` oracle was accepted from the PR body and read-verified only at the short-circuit-order level, not exercised).
- Stated intent: PR #815 title + body — **spec-compliance assessed** against it.
- Synthesis was performed by the orchestrator inline rather than by a separate agent, after two sub-agent dispatches in this run failed (one 45-minute timeout, one truncated stream). Findings themselves come from the dimension agents plus the verification pass above; none were authored during synthesis.
- No `critical` or `high` finding was present, so the adversarial `/shadow-verify` wave was not triggered.

griffinwork40 (commented):

## PR #815 — MERGE (with follow-ups)

**Reviewed ref:** `10c35ecd5b58e343890b163ffb909a9d775e8406` · 3 files, +294/-6 · full regime (security · api-compat · correctness · spec-compliance · test-coverage · perf-observability)

The core fix is **sound**. I traced the gate by hand at the reviewed ref and confirmed all four load-bearing properties the PR body claims: gating happens *before* the `allow` union, the memo key includes `AFK_HOME`, operator extras are still checked ahead of `allow`, and the pierced-root exclusion is genuinely by `source` equality rather than by resolved path. No blocking defect. Nothing below is an auth-bypass; the findings are about the *tests and the PR body*, not the mechanism.

### medium · high confidence · spec-compliance / test-coverage
`read-denylist-carveout.test.ts:36-42` · `ref:10c35ecd` · citation-type: diff-context

The test named for the equality-vs-prefix discriminator is **byte-identical** to the case above it (`:26-29`) — same input, same assertion — and it calls the *pure* gate, which performs no exclusion at all. Exclusion lives in `gateRootsExcludingSource` (`read-denylist-carveout.ts:82-89`), which is not in this call path. So the PR body's "A dedicated test pins this discriminator" does not hold: the discriminator is unpinned, and this test would pass unchanged if the exclusion were switched to prefix matching.

```ts
    const entry = join(OTHER_DENIED_ROOT, 'afk', 'config', 'mcp.json');
    expect(gateDerivedCarveOuts([entry], [OTHER_DENIED_ROOT])).toEqual([]);
```

**Suggestion:** assert the discriminator where it lives — pass two entries whose pierced `source` strings differ but whose resolved roots collide, and check only the non-pierced one is gated.

*Mitigating:* the behavioural guarantee **is** covered end-to-end by the `isReadDenied` integration cases at `:77-87`, including the assertion that the verdict names the outer floor. That is why this is a follow-up and not a blocker.

### medium · high confidence · spec-compliance
PR description vs `ref:10c35ecd` · citation-type: file-state (re-read at the ref)

Three factual claims in the body describe an earlier commit on the branch, not the reviewed tip. Measured at the ref: the new module is **132 lines total / 36 non-comment code lines** (body says "52 LOC"); the test file has **14 `it()` blocks** (body says "12 cases", and the "12/12 pass" gate therefore cannot attest to two tests that exist only at this SHA); `read-denylist.ts` is **483 lines** (body says "already 425 LOC").

**Suggestion:** refresh the LOC, test-count, and pass-count figures against `10c35ecd` before merge.

### low · high confidence · test-coverage
`read-denylist-carveout.ts:122-132` · `ref:10c35ecd` · citation-type: diff-context

`gateStaticCarveOuts` — the export that fixes the `.ssh/config` / `.ssh/known_hosts` regression this revision exists to correct — has **zero direct unit coverage**; `gateRootsExcludingSource` likewise. Verified: `git grep gateStaticCarveOuts <ref> -- '*.test.ts'` returns zero matches, while `gateDerivedCarveOuts` gets 7 direct cases. Coverage is indirect only, via `isReadDenied`.

**Suggestion:** add one direct test over the per-entry `CARVEOUT_PIERCED_SOURCE` mapping, asserting `.ssh/config` survives its own gate while an entry under a *different* root is dropped.

### low · low confidence · correctness / security
`read-denylist-carveout.ts:38-41` · `ref:10c35ecd` · citation-type: diff-context — *[low confidence — verify with runtime context]*

The new containment primitive is purely lexical, while every other containment check in this feature routes through `pathIsWithin` (`src/agent/tools/fs-case.ts:83-89`), which case-folds when `isCaseInsensitiveFs()` is true — the macOS default.

```ts
function isUnderRoot(candidate: string, root: string): boolean {
  const rel = relative(root, candidate);
  return rel === '' || (rel !== '..' && !rel.startsWith(`..${sep}`) && !isAbsolute(rel));
}
```

`..` segments and trailing separators are correctly neutralised by `relative()`, and the sibling-prefix case (`~/.gnupg-extra` vs `~/.gnupg`) is correct and tested at `:44-47`. **`unverified` — reachability not established:** both operands pass through `safeRealpath` (`write-denylist.ts:138`), which realpaths existing components but re-appends non-existent tail segments as written, so whether a case-variant `AFK_HOME` survives the gate depends on platform realpath canonicalisation I did not execute.

**Suggestion:** route `isUnderRoot` through `pathIsWithin` so the gate and the deny loop share one comparator.

### low · high confidence · test-coverage
`read-denylist-carveout.test.ts:54-60` · `ref:10c35ecd` · citation-type: diff-context

Test name says "**preserves** an independent root that canonicalizes to the carved root" but the assertion is `toEqual([])` — a drop. The symlink-aliasing invariant its comment describes (module contract, `read-denylist-carveout.ts:23-26`) is not modelled: `grep -i symlink` on the test file returns zero matches.

**Suggestion:** rename to match the assertion, and add a case constructing two distinct `source` entries resolving to one canonical root.

### nit · read-denylist.ts
The body cites "the repo's hard 350-LOC ceiling" as the reason for the new module — correct call — yet this PR still adds +41 lines to `read-denylist.ts`, taking it to **483** at the ref.

### Clean reads
- **api-compat — no breaking change.** `isReadDenied`'s signature is unchanged; the two new exports are additive and consumed only inside `read-denylist.ts` plus its test. Production importers confirmed at `input-parse.ts`, `afk-home-refs.ts`, `path-approval-hook.ts`, `bash-restriction-hook.ts`, `glob.ts`, `grep.ts`, `at-file-inject.ts`, `_cwd-utils.ts`.
- **perf-observability — no issue.** Gating runs only inside `resolveLists()`, behind the unchanged `cached.key === key` memo check, so the per-call `isReadDenied` hot path takes no added cost.
- **Fail-closed — holds.** The documented empty-`gateRoots` edge case (`:50-55`) is unreachable from both call sites at current `BUILTIN_READ_DENYLIST` cardinality. An unmapped future `READ_ALLOWLIST_REL` entry fails closed *and* is caught by the iterating test at `:121-127` — a good guard.

### What was not checked
Citations verified inline against PR head `10c35ecd` (fetched as a local ref); absence claims re-verified with `git grep` at that same ref. Stated intent: PR #815 title + body — spec-compliance assessed. **Not** checked: no tests or lint were executed (static read only, so the body's "145 passed" gate is unverified here); `safeRealpath` case/NFD/symlink semantics not executed on a case-insensitive volume; `bash-restriction-hook.ts` internals reviewed only far enough to confirm this diff leaves them untouched; `getReadDenylistDescendants` consumers outside the changed-file set not traced.

PR #816, opened 2026-07-31
fix(build): fail the bundle build when a prompt read cannot be inlined (#776)

### LINKED ISSUE #776: build: prompt inliner silently skips parameterized readFileSync, shipping a bundle that reads a missing file
## Problem

`scripts/esbuild-plugin-inline-prompts.mjs` inlines prompt files into the published bundle so `dist/*.mjs` has no runtime `readFileSync` on a `.md` that was never shipped. Its Pattern A resolver, `tryResolveReadFileSyncPath` (`:110-129`), matches:

```js
/readFileSync\(\s*(?:join|resolve)\s*\(\s*(?:__dirname|here)\s*,\s*(.+?)\)\s*,\s*['"]utf-?8['"]\s*\)/
```

and then requires **every** join argument to be a string literal:

```js
const strMatch = trimmed.match(/^['"](.+)['"]$/);
return strMatch ? strMatch[1] : null;
// ...
if (parts.every(p => p !== null)) { /* inline */ }
```

When a part is not a literal, `tryResolveReadFileSyncPath` returns `null`, the call is **left as-is**, and the build **succeeds**. The bundle then ships code that reads a file which is not in the package.

## Why this is worth fixing

The failure is silent at every gate that would normally catch it:

- `pnpm lint` — passes, the code is valid TypeScript
- `pnpm test` — passes, tests run against `src/`, not the bundle
- `pnpm build` — passes, and `copy-prompts.js` copies all `src` `.md` unconditionally, so the non-dist build **masks it entirely**
- `pnpm build:dist` — passes, `stats.inlinedFiles` just silently decrements

It surfaces only when a user runs the published npm artifact. There is no assertion anywhere that the set of prompt reads found equals the set inlined.

This is a live trap, not a hypothetical. The natural way to write a helper that loads several prompts is exactly the shape the resolver cannot see:

```ts
// Looks obviously correct. Ships a broken bundle.
const loadPrompt = (name: string) =>
  readFileSync(join(__dirname, 'prompts', `${name}.md`), 'utf8');
```

Today this is latent because the three files using Pattern A each hardcode their own literal path (`src/skills/_agents/{research-agent,git-investigator,contract}.ts`). Any refactor that consolidates them — for instance ADR 0002 item 2, whose impact map surfaced this — trips it.

Note the regex tolerates whitespace and newlines (`\s*` throughout), so **formatting is safe**. The sensitivity is specifically to *parameterization*.

## Suggested fix

Make the plugin **fail closed**. Some options:

1. Detect the near-miss: if a source file contains `readFileSync(` with `__dirname`/`here` **and** a `.md` string anywhere in the call, but `tryResolveReadFileSyncPath` returned `null`, **throw** with the file and line. Catches the parameterized case precisely, no false positives on unrelated reads.
2. Post-build assertion: grep `dist/*.mjs` for any surviving `readFileSync(...\.md` and fail the build. Cheap, catches all escape routes, mirrors the existing `assert-moat-clean.mjs` pattern.
3. Emit `stats.inlinedFiles` and pin it, so a silent drop shows up as a diff.

(2) is probably the best value — it is the same fail-closed shape already used elsewhere in the build, and it does not care *how* the read escaped inlining.

## Context

Found during the `/impact-map` pass on ADR 0002 item 2 — `docs/audits/impact-map-agents-wrapper-layer.md` (PR #774), correction C3. The impact map recommends against that refactor partly *because* of this gap; closing this issue would remove one of the reasons.

griffinwork40:
## Summary

A parameterized prompt path made the inliner silently skip a call, leaving a runtime `readFileSync` on a `.md` that is not in the published package — with a **green build**. This makes that case fail the build, at two independent layers.

## Why every existing gate missed it

| Gate | Why it passed |
|---|---|
| `pnpm lint` | valid TypeScript |
| `pnpm test` | tests run against `src/`, never the bundle |
| `pnpm build` | `copy-prompts.js` copies all `src` `.md` unconditionally — **masks it entirely** |
| `pnpm build:dist` | `stats.inlinedFiles` silently decremented; exit 0 |

It surfaced only when a user ran the published npm artifact. Nothing asserted that the set of prompt reads *found* equals the set *inlined*.

## The finding that changes the fix

The issue proposed detecting the near-miss inside `tryResolveReadFileSyncPath`. That alone **could never have fired**. The file-level gate one layer up is:

```js
if (!content.includes('readFileSync') || !/\.md['"]/.test(content)) continue;
```

It requires a **quote** immediately after `.md`. A parameterized read ends its path in a template literal — `` `${name}.md` `` — which ends in a **backtick**, so the issue's own canonical example was skipped at the *file* level and never reached the resolver at all. Widening that probe is what makes the trap reachable; the throw is what makes it fatal.

## Change

**1. `scripts/esbuild-plugin-inline-prompts.mjs`** (262 → 296 LOC)
- File-level `.md` probe now accepts a backtick, so parameterized shapes reach the resolver.
- The resolver's `null` branch now **throws**, naming file and line plus the remedy (use a literal path, or teach the resolver the shape).
- Gated on a new `isPromptShapedRead()` predicate — the call must be anchored to `__dirname`/`here` **and** name a `.md`. Unrelated `readFileSync` calls stay silent, so there are no false positives on reads the plugin was never responsible for.

**2. New `scripts/assert-inlined-prompts.mjs`** (91 LOC), wired into `build-dist.mjs` before bundling — the backstop that catches Pattern B/C escapes and future shapes, not just Pattern A.

**Deliberate deviation from the issue, which suggested grepping `dist/*.mjs`:** inlined prompt bodies land in the bundle as template literals, and several of this repo's prompts contain code examples including `readFileSync` and `.md` filenames — so a proximity grep over minified output would false-positive on prompt *content* and fail a correct build. Minification also renames local bindings. The prepared source tree has neither problem: unminified, post-inlining, and a survivor there is unambiguous. Checking one layer earlier is strictly more reliable.

**Second deviation, forced by evidence:** the backstop only flags reads anchored to the module's own directory. An unanchored scan flagged **13 correct call sites** on its first run — `/changelog` reads the user's repo `CHANGELOG.md`, `user-skills.ts` reads user-authored `SKILL.md`, tests read temp fixtures. Those target files on the *user's* disk, which the package never ships and never should. The `__dirname`/`here` anchor is what separates "must be inside the published package" from "belongs to the user."

## Proof

**Clean build — no false positive on the three existing literal-path callers:**
```
[inline-prompts] Summary: 4 files inlined, 0 internal-tier stubs, 4 readFileSync calls replaced, prompt-loader replaced
[assert-inlined-prompts] 1427 files verified: no runtime .md reads
Build complete:
```

**Deliberate regression — `src/skills/_agents/contract.ts` temporarily converted to the issue's parameterized shape:**
```
Error: [inline-prompts] Unresolvable prompt read at src/skills/_agents/contract.ts:10
  Every join()/resolve() argument must be a string LITERAL so the path can be
  resolved at build time and the prompt inlined. A parameterized path (template
  literal or variable) cannot be resolved here, and leaving it in place would
  ship a runtime readFileSync on a file that is not in the published package.
  Fix: use a literal path at this call site, or teach
  tryResolveReadFileSyncPath() to understand this shape.
 ELIFECYCLE  Command failed with exit code 1.
```
That scratch change was reverted (`git diff` on the file is empty) and `build:dist` re-verified green afterwards. It is **not** in this diff.

## Gates

| Gate | Result |
|---|---|
| `pnpm build:dist` (clean) | pass — 4 inlined, 1427 files verified |
| `pnpm build:dist` (deliberate regression) | **fails as designed**, names file:line |
| `pnpm build:dist` (after revert) | pass |
| `pnpm lint` | pass |

No `src/` file changed, so no test suites are affected. `scripts/assert-moat-clean.mjs` does not exist in this repo (private-tier), so there was no existing post-build grep to mirror — hence the new module.

## Risk

This guards the **publish path**, so a false positive would block releases. Two things bound that: the throw fires only on a `__dirname`-anchored `.md` read that cannot be resolved — the exact shape that is already broken — and the backstop's anchor requirement was derived empirically from the 13 legitimate call sites above rather than guessed. Both were verified green on a clean build before and after the regression test.

The residual risk is the opposite direction: a genuinely new *legitimate* pattern for reading a shipped `.md` anchored at `__dirname` would now fail the build until either the resolver or the anchor list is extended. That is the intended fail-closed trade.

Fixes #776

### DIFF
--- a/scripts/assert-inlined-prompts.mjs
+++ b/scripts/assert-inlined-prompts.mjs
@@ -1,0 +1,91 @@
+/**
+ * Fail-closed assertion that no prompt read survived inlining (#776).
+ *
+ * Invariant: this runs against the PREPARED SOURCE TREE, not `dist/*.mjs`. The
+ * issue suggested grepping the bundle, and that was reconsidered deliberately:
+ * inlined prompt bodies land in the bundle as template literals, and several of
+ * this repo's prompts contain code examples — including `readFileSync` and `.md`
+ * filenames — so a proximity grep over minified output would false-positive on
+ * prompt CONTENT and fail a correct build. Minification also renames local
+ * bindings, making the call shape unreliable to match.
+ *
+ * The prepared tree has neither problem: it is unminified TypeScript, every
+ * inlining pass has already run, and a surviving `readFileSync(...'.md')` there is
+ * unambiguously an escape. Checking one layer earlier is strictly more reliable
+ * and catches Pattern A, B and C escapes uniformly.
+ *
+ * Contract: throws with every offending file:line on the first failure; returns
+ * the number of files scanned otherwise. Never mutates anything.
+ *
+ * @module scripts/assert-inlined-prompts
+ */
+
+import { readFileSync, readdirSync } from 'node:fs';
+import { join, relative } from 'node:path';
+
+/**
+ * Any `readFileSync` call naming a `.md` path, in any quote style. Deliberately
+ * broader than the inliner's own resolver pattern: this is the backstop, so it
+ * must catch shapes the resolver never claimed to handle.
+ */
+const SURVIVING_MD_READ = /readFileSync\([^)]{0,400}?\.md['"`]/g;
+
+/**
+ * Invariant: only a read anchored to the MODULE'S OWN directory is in scope.
+ * Plenty of legitimate runtime `.md` reads exist and must not fail the build —
+ * `/changelog` reads the user's repo `CHANGELOG.md`, `user-skills.ts` reads
+ * user-authored `SKILL.md` from disk, and tests read temp fixtures. Those target
+ * files on the USER's disk, which the package never ships and never should.
+ * A `__dirname`/`here` anchor is what distinguishes "this file must be inside the
+ * published package" from "this path belongs to the user" — an unanchored scan
+ * flagged 13 correct call sites on the first run.
+ */
+function isPackageRelativeRead(callText) {
+  return /__dirname|\bhere\b/.test(callText);
+}
+
+function walkTsFiles(dir, onFile) {
+  for (const entry of readdirSync(dir, { withFileTypes: true })) {
+    const full = join(dir, entry.name);
+    if (entry.isDirectory()) {
+      walkTsFiles(full, onFile);
+      continue;
+    }
+    if (entry.isFile() && entry.name.endsWith('.ts')) onFile(full);
+  }
+}
+
+/**
+ * @param {string} preparedSrc Root of the prepared (post-inlining) source tree.
+ * @returns {number} Files scanned.
+ */
+export function assertNoRemainingPromptReads(preparedSrc) {
+  const offenders = [];
+  let scanned = 0;
+
+  walkTsFiles(preparedSrc, (filePath) => {
+    scanned++;
+    const content = readFileSync(filePath, 'utf-8');
+    if (!content.includes('readFileSync')) return;
+    for (const match of content.matchAll(SURVIVING_MD_READ)) {
+      if (!isPackageRelativeRead(match[0])) continue;
+      const line = content.slice(0, match.index).split('\n').length;
+      offenders.push(
+        `  ${relative(preparedSrc, filePath)}:${line}  ${match[0].replace(/\s+/g, ' ').slice(0, 120)}`,
+      );
+    }
+  });
+
+  if (offenders.length > 0) {
+    throw new Error(
+      `[assert-inlined-prompts] ${offenders.length} prompt read(s) survived inlining:\n` +
+        `${offenders.join('\n')}\n` +
+        `  These would ship as runtime reads of .md files that are NOT in the published\n` +
+        `  package, failing only when a user runs the published artifact. Either make the\n` +
+        `  path a string literal the inliner can resolve, or extend the inliner to handle\n` +
+        `  the shape. See scripts/esbuild-plugin-inline-prompts.mjs.`,
+    );
+  }
+
+  return scanned;
+}
--- a/scripts/build-dist.mjs
+++ b/scripts/build-dist.mjs
@@ -82,13 +82,13 @@ mkdirSync(distDir, { recursive: true });
 // Step 1: pre-transform sources
 const { tmpSrc, tmpBase } = prepareSources();
 
-// Step 1b: fail CLOSED before bundling if any prompt read survived inlining
-// (#776). Checked on the prepared tree rather than dist/ — see the module
-// docblock for why the bundle is the wrong layer to grep.
-const scannedForPromptReads = assertNoRemainingPromptReads(tmpSrc);
-console.log(`  [assert-inlined-prompts] ${scannedForPromptReads} files verified: no runtime .md reads`);
-
 try {
+  // Step 1b: fail CLOSED before bundling if any prompt read survived inlining
+  // (#776). Checked on the prepared tree rather than dist/ — see the module
+  // docblock for why the bundle is the wrong layer to grep.
+  const scannedForPromptReads = assertNoRemainingPromptReads(tmpSrc);
+  console.log(`  [assert-inlined-prompts] ${scannedForPromptReads} files verified: no runtime .md reads`);
+
   // Step 2: bundle from transformed sources
   await build({
     entryPoints: {
--- a/scripts/build-dist.mjs
+++ b/scripts/build-dist.mjs
@@ -18,6 +18,7 @@ import { readFileSync, writeFileSync, rmSync, mkdirSync, statSync, chmodSync, co
 import { join, dirname } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { prepareSources } from './esbuild-plugin-inline-prompts.mjs';
+import { assertNoRemainingPromptReads } from './assert-inlined-prompts.mjs';
 import { copyBundledPlugins } from './lib/copy-bundled-plugins.mjs';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
@@ -80,6 +81,12 @@ mkdirSync(distDir, { recursive: true });
 
 // Step 1: pre-transform sources
 const { tmpSrc, tmpBase } = prepareSources();
+
+// Step 1b: fail CLOSED before bundling if any prompt read survived inlining
+// (#776). Checked on the prepared tree rather than dist/ — see the module
+// docblock for why the bundle is the wrong layer to grep.
+const scannedForPromptReads = assertNoRemainingPromptReads(tmpSrc);
+console.log(`  [assert-inlined-prompts] ${scannedForPromptReads} files verified: no runtime .md reads`);
 
 try {
   // Step 2: bundle from transformed sources
--- a/scripts/esbuild-plugin-inline-prompts.mjs
+++ b/scripts/esbuild-plugin-inline-prompts.mjs
@@ -286,7 +286,11 @@ export function prepareSources() {
       stats.inlinedFiles++;
       stats.replacedCalls++;
     } else {
-      console.log(`  [inline-prompts] WARNING: Could not find loadSystemPrompt() in shared-helpers.ts`);
+      throw new Error(
+        '[inline-prompts] Could not find loadSystemPrompt() in shared-helpers.ts\n' +
+        '  Pattern C replacement failed. The function body may have changed.\n' +
+        '  Update the oldFn string in prepareSources() to match the current code.'
+      );
     }
   }
 
--- a/scripts/esbuild-plugin-inline-prompts.mjs
+++ b/scripts/esbuild-plugin-inline-prompts.mjs
@@ -107,6 +107,18 @@ function escapeForTemplate(str) {
   return str.replace(/\\/g, '\\\\').replace(/`/g, '\\`').replace(/\$/g, '\\$');
 }
 
+/**
+ * Contract: true when a call LOOKS like a prompt read this plugin is responsible
+ * for inlining — it resolves a path against the module's own directory and names
+ * a `.md` file — regardless of whether the path could be resolved. Used to tell a
+ * genuine near-miss (parameterized prompt path, must fail the build) from an
+ * unrelated `readFileSync` the plugin has no business touching (no `__dirname`
+ * anchor, or not a markdown file), which must stay silent.
+ */
+function isPromptShapedRead(callText) {
+  return /__dirname|\bhere\b/.test(callText) && /\.md['"`]/.test(callText);
+}
+
 function tryResolveReadFileSyncPath(callText, fileDir) {
   const joinMatch = callText.match(
     /readFileSync\(\s*(?:join|resolve)\s*\(\s*(?:__dirname|here)\s*,\s*(.+?)\)\s*,\s*['"]utf-?8['"]\s*\)/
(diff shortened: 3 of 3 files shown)


griffinwork40:
Now I have full context. Here's the review:

---

## PR #816 Review: `fix(build): fail the bundle build when a prompt read cannot be inlined`

### Summary

This PR adds two layers of fail-closed protection to the build pipeline, ensuring that unresolvable prompt reads (which would ship as broken runtime `readFileSync` calls on files not in the npm package) cause the build to fail rather than silently succeed.

### Verdict: **Approve** ✅

Well-scoped, well-documented, and correctly layered. The design decisions are sound and the error messages are excellent.

---

### Correctness — ✅

**Two-layer defense is correctly designed:**

1. **Layer 1 (inliner)**: The `else if (isPromptShapedRead(match[0]))` branch fires when the inliner's regex matches a `readFileSync(join/resolve(...))` call but `tryResolveReadFileSyncPath()` can't statically resolve the path. This catches parameterized paths like `` join(__dirname, `${name}.md`) `` — the shape is right but the argument isn't a literal.

2. **Layer 2 (assertion script)**: `assertNoRemainingPromptReads()` walks the prepared tree with a broader regex (`/readFileSync\([^)]{0,400}?\.md['"`]/g`) that catches ANY surviving `.md` read, not just ones matching the inliner's specific `join`/`resolve` pattern. This is intentionally broader — it's the backstop.

**The backtick fix is the key bug fix:**
```javascript
// Before: /\.md['"]/  — missed template literals
// After:  /\.md['"`]/ — catches `${name}.md` too
```
This is what made the near-miss check reachable for the exact shape that ships a broken bundle (#776).

**False-positive filtering is correct:**
- `isPackageRelativeRead()` checks for `__dirname` or `\bhere\b` — correctly excludes user-disk reads (CHANGELOG.md, user SKILL.md, test fixtures)
- The assertion script's comment documents that 13 correct call sites were flagged before this filter was added

**I verified against the codebase** — the three `__dirname`-anchored `.md` reads in `src/skills/_agents/` (research-agent.ts, contract.ts, git-investigator.ts) are the ones that must be inlined, and the user-disk reads in `user-skills.ts`, `changelog.ts`, and test files correctly lack the `__dirname`/`here` anchor.

### Security — ✅

No concerns. This is build-time code that doesn't handle user input, credentials, or external data. The regex patterns are bounded (400-char limit on the surviving-read regex prevents runaway matching).

### API Compatibility — ✅

No public API changes. Purely internal build infrastructure. No breaking changes.

### Test Coverage — ⚠️ Minor gap

No tests for the new assertion script or the fail-closed behavior. However:
- This is build-time code, not runtime — the build itself is the integration test
- The PR was validated against the real codebase (13 false positives caught and fixed per the comments)
- Testing would require creating fixture files with specific patterns, which adds complexity for marginal value

**Acceptable** for build infrastructure, but worth noting.

### Performance — ✅

Negligible impact. The assertion walks `.ts` files in the prepared tree with an early `content.includes('readFileSync')` short-circuit. O(n) in source files, adds milliseconds to the build.

### Code Quality — ✅

**Excellent documentation:**
- Every regex has a docblock explaining what it matches and why it's shaped that way
- The assertion script's module docblock explains why checking the prepared tree is strictly better than grepping the bundle (minification + template literals in prompt content would cause false positives)
- Error messages include file:line, the offending code snippet, and actionable fix guidance

**Good naming:**
- `isPromptShapedRead` vs `isPackageRelativeRead` — clear distinction between "looks like a prompt" and "is a package-relative read"
- `assertNoRemainingPromptReads` — clearly states the contract

**Correct layering:**
- The inliner throws during `prepareSources()` (fail-fast, specific to the inliner's pattern)
- The assertion runs after `prepareSources()` returns (backstop, broader coverage)
- Both run before esbuild bundles, so failures are caught early with clear context

### Design Decisions — ✅

**Why check the prepared tree instead of the bundle?**
The docblock explains this well: inlined prompts become template literals in the bundle, and several prompts contain code examples with `readFileSync` and `.md` filenames. A proximity grep over minified output would false-positive on prompt CONTENT. The prepared tree is unminified TypeScript where a surviving `readFileSync(...'.md')` is unambiguously an escape.

**Why two layers instead of one?**
The inliner's `else if` only catches calls matching its specific regex pattern. The assertion catches ANY surviving `.md` read, including shapes the inliner never tried to handle (e.g., a bare `readFileSync(someVar + '.md', 'utf-8')` without `join`/`resolve`). Defense in depth.

### Minor Observations (non-blocking)

1. **The assertion's regex could match inside comments or string literals** — e.g., `// don't use readFileSync(path, 'file.md')` would match. This is acceptable for a build-time check (false positive = build fails, human investigates), but worth knowing.

2. **The 400-char limit in `SURVIVING_MD_READ`** is reasonable but arbitrary. If someone writes a very long `readFileSync` call with nested function calls, the regex might not match. However, such a shape would be unusual and the inliner wouldn't handle it either.

3. **`walkTsFiles` only checks `.ts` files** — consistent with the inliner, which also only processes `.ts`. If the project ever adds `.tsx` or `.mts` files with prompt reads, both would need updating.

---

### Merge Recommendation

**Merge.** This is a well-crafted fix for a real problem (silent build success when prompt inlining fails) with appropriate layering, excellent documentation, and no correctness issues. The minor test coverage gap is acceptable for build infrastructure.

The PR correctly identifies that the previous behavior — leaving an unresolvable `readFileSync` call in place and succeeding the build — would ship a bundle that fails at runtime when a user tries to use the feature. The fix ensures such cases fail at build time with clear, actionable error messages.

---

griffinwork40:
**Wave 1.5 verification complete.** All citations checked against reviewed ref `5821ead`. Results:

| Finding | Citation type | Status |
|---------|--------------|--------|
| 2.1 | diff-context | ✅ verified |
| 3.1 | diff-context (`assert-inlined-prompts.mjs` line 37 of diff) | ✅ verified — `[^)]{0,400}?` regex confirmed |
| 3.2 | file-state (`esbuild-plugin-inline-prompts.mjs:190`) | ⚠ **line number wrong** — actual `[^)]*` at line 207, not 190; regex content confirmed |
| 3.3 | file-state (`build-dist.mjs:82-84`) | ✅ verified — `prepareSources()` at line 83, outside `try` at line 85, `finally` at 166-168 |
| 3.4 (no issue) | diff-context | ✅ verified — `content.matchAll` at line 210, `content` never mutated |
| 4.1 | diff-context (lines 14-16 of new file) | ✅ verified |
| 5.1 | absence claim | ✅ confirmed-absent — zero test files for `prepareSources`/`assertNoRemainingPromptReads`/`isPromptShapedRead` |
| 6.1/6.2 | diff-context | ✅ verified |

**No fabricated citations. No false-absent claims. One line-number correction** (3.2: line 190 → 207). Merging 3.1+3.2 → single `medium` finding (same root cause: `[^)]` in both regexes). No critical/high findings → **shadow-verify skipped** per protocol.

---

## PR #816 Review — Synthesis

**Title:** fix(build): fail the bundle build when a prompt read cannot be inlined (#776)
**Branch:** `afk/fix-776-inliner-failclosed` → `main` · **Ref:** `5821ead`
**Stats:** +138 / -2 / 3 files · **Regime:** light (hotfix)
**Stated intent:** Make the bundle build fail when a prompt `readFileSync` on a `.md` cannot be inlined, at two independent layers; widen the `.md` probe to accept backticks (template-literal paths); convert the `loadSystemPrompt` not-found warning into a throw.

All four sub-claims from the PR body verified against the diff:
- **(a)** `.md` probe accepts backtick ✅ — `/\.md['"`]/` at diff line 156
- **(b)** `isPromptShapedRead()` predicate gates throws ✅ — diff lines 132-141
- **(c)** resolver `null` branch throws ✅ — diff lines 164-180
- **(d)** `loadSystemPrompt` warning→throw ✅ — diff lines 188-193

**Independence of two layers:** Partially confirmed. Layer 1 (the inliner's `else if isPromptShapedRead(...)` throw) fires inside `prepareSources()` for any call the inliner's `matchAll` regex matches but can't resolve. Layer 2 (the backstop) scans the prepared tree independently and catches surviving reads the inliner's regex never matched at all. They are architecturally independent — but share a regex-level blind spot (`[^)]` bounds, Finding 3.1). For the PR's primary target shape (template-literal path `` `${name}.md` ``), only layer 1 fires (the backstop's `[^)]]` also fails to traverse nested `join()` parens; but here there ARE no nested parens, so the backstop matches too). No scope creep detected.

### Findings (sorted critical → high → medium → low → nit)

| # | Sev | Conf | Dim | File:line | Finding |
|---|-----|------|-----|-----------|---------|
| 3.1+3.2 | **medium** | high | correctness | `assert-inlined-prompts.mjs:37` + `esbuild-plugin-inline-prompts.mjs:207` | Both the backstop regex `[^)]{0,400}?` and the inliner's `matchAll` regex `[^)]*` stop at the first `)`. A `readFileSync(join(__dirname, getSubdir(), 'prompt.md'), 'utf-8')` — where a nested function call introduces an intermediate `)` before the `.md` — is invisible to **both** layers: the inliner's regex doesn't match → no `isPromptShapedRead` check → no throw; the backstop's regex also doesn't match → no scan hit. Build succeeds and ships a runtime read of a non-existent `.md`. **No current production code has this shape** (grep for `readFileSync(join(__dirname.*(` returned zero nested-paren hits), so this is a theoretical robustness gap, not a live defect. |
| 5.1 | **medium** | high | test-coverage | `scripts/` (all) | Zero test files exist for any of the three build scripts. Grep for `prepareSources`/`assertNoRemainingPromptReads`/`isPromptShapedRead` across all `*.test.*`/`*.spec.*` files returned zero matches. The new `assertNoRemainingPromptReads()` (pure function, clear contract: returns count → clean, throws with file:line → offenders) and `isPromptShapedRead()` are directly unit-testable. |
| 2.1 | **low** | high | api-compat | `esbuild-plugin-inline-prompts.mjs:254-259` | `prepareSources()` now throws (instead of `console.log`) when `loadSystemPrompt()` can't be found in `shared-helpers.ts`. The sole production importer is `build-dist.mjs:20`. This is deliberate and within stated intent — the next maintainer who reformats `loadSystemPrompt()` will see a hard build failure instead of a silent warning. Recommend documenting in the function's vicinity. |
| 3.3 | **low** | high | correctness | `build-dist.mjs:83` | `prepareSources()` is called at line 83, **outside** the `try` block (line 85). The new Pattern C and Pattern A throws inside `prepareSources()` propagate as uncaught exceptions; the `finally` at line 166 (`rmSync(tmpBase, ...)`) never executes. The temp dir `os.tmpdir()/afk-build-XXXX/` leaks on failure. Pre-existing structure, but newly reachable from the PR's new throw paths. |
| 4.1 | **low** | high | spec-compliance | `assert-inlined-prompts.mjs:14-16` | Module docblock states "catches Pattern A, B and C escapes uniformly." The `[^)]` limitation (Finding 3.1) means the backstop does NOT catch all shapes — a nested-paren argument shape survives both layers. "Uniformly" is an overclaim for a not-yet-observed but structurally possible escape pattern. |

### No issues found (confirmed clean reads)

- **Security** — all three scripts are build-time-only, no network surface, no secret handling, no user-input paths, no eval/Function construction. No injection vectors.
- **Correctness (3.4, 3.5, 3.6)** — line-number computation verified correct (`content` never mutated, `match.index` is valid); `isPromptShapedRead`/`isPackageRelativeRead` predicates textually consistent; file-level probe and `isPromptShapedRead` probe consistent on `.md` boundary after PR change.
- **Perf/observability (6.1, 6.2)** — `walkTsFiles` + `includes('readFileSync')` fast-path adds O(seconds) on ~1400 files; negligible vs esbuild/tsc. Log line provides file count + pass/fail status; failure lists every offender with file:line + 120-char excerpt. Sufficient.

### What was not checked

- **Citations verified against reviewed ref `5821ead`** using `git show`. All `diff-context` citations confirmed in the hunk. All `file-state` citations confirmed at the ref (3.2 line corrected 190→207; 3.3 structure confirmed).
- **Stated intent:** PR #816 title+body — spec-compliance assessed.
- **Not checked:** runtime behavior of the prepared temp tree (whether `cpSync` preserves all attributes needed by the backstop walk); ReDoS potential of `SURVIVING_MD_READ` (analyzed statically — `[^)]` prevents catastrophic backtracking; no issue); CI filesystem edge cases (symlinks in temp dir).
- **Not checked:** whether `INTERNAL_TIER_PROMPT_BASENAMES` stub path interacts with the new throw (it does not — the stub path returns a valid `mdPath`, so the `else if` is unreachable for stubbed prompts).

---

The change does exactly what it claims — fails the bundle build at two independent layers when a prompt `readFileSync` on a `.md` cannot be inlined. The primary bug (template-literal paths silently skipped) is correctly fixed: the `.md` probe now accepts backticks, and the resolver's `null` branch throws with a precise file:line error. No critical or high findings. The two `medium` findings are a theoretical regex robustness gap (no current production exposure) and missing unit tests for build scripts — neither blocks a merge of a build-system hotfix. The three `low` findings (temp-dir leak on the new throw path, docblock "uniformly" overclaim, deliberate api-compat change) are worth follow-up but not merge-blockers.

---

Issue #795, opened 2026-07-31
Auto Release blocked by a width-dependent footer test: a green merge can produce no release


griffinwork40:
## Summary

`Auto Release` failed on `71e6560` (the #772 merge) because of a single failing test, so **no release was cut for that merge**. The `CI` run on the very same commit passed. The test asserts footer content without pinning terminal width, and the two workflows evidently render at different widths.

Not data-loss, but it silently gates publishing: a merge can land green and produce no release.

## Evidence

Failing run: [Auto Release 30634635208](https://github.com/griffinwork40/agent-afk/actions/runs/30634635208), job `Bump, tag, publish`, at `71e6560`.

```
FAIL src/cli/commands/interactive/turn-handler.test.ts >
  printTurnFooter — subscription-quota line >
  prints the quota line with its deadline once the binding window is hot

AssertionError: expected '  ◦ 1s  ·  $0.01\n  5h quota 94% used…' to contain 'resets in 12m'

Test Files  1 failed | 720 passed (721)
Tests  1 failed | 13744 passed | 2 skipped (13747)
```

The `CI` run on the same SHA reported `Test (ubuntu-latest)` **pass**.

## Diagnosis

Read the received string closely — it ends in `used…`. That trailing ellipsis is the footer's own width clamp (`clampToTerminal` → `truncateDisplayWidth`), so the earlier assertion `toContain('5h quota 94% used')` succeeded and the deadline clause was **cut off by truncation**, not rendered wrong.

The test (`turn-handler.test.ts:2346-2355`) sets `fiveHourResetsAt: new Date(Date.now() + 12 * 60_000)` and asserts `toContain('resets in 12m')`, but never pins the width the footer renders into. So the assertion holds only when the effective terminal is wide enough to fit the whole line — which is environment-dependent, and the Auto Release job apparently reports a narrower width (or no TTY, falling back to a narrow default) than the CI Test job.

Confidence: high that truncation is the mechanism, since the ellipsis is right there in the received value. I have **not** reproduced it locally — the suite passes at full width, which is consistent with the theory but does not prove which width source differs between the two workflows.

Secondary possibility worth ruling out while fixing: the deadline text is derived from wall-clock delta, so a slow runner could render `11m` instead of `12m`. Truncation explains the observed output better, but both are the same underlying flaw — the test pins output that depends on ambient state it does not control.

## Suggested fix

[...]

PR #819, opened 2026-08-01
fix(permissions): close the case-variant bypass of the credential floor (#736)

### LINKED ISSUE #736: Read denylist: case-variant spellings bypass the credential floor on case-insensitive filesystems

griffinwork40:
## Summary

On a case-insensitive volume (macOS APFS by default) a case-variant spelling of a protected directory sailed past the credential floor on **both** the typed-tool surface and the bash surface. macOS `realpath(3)` does not canonicalize case, so the requested spelling survives into the comparison — and all four comparison sites compared case-sensitively:

| Site | Comparison |
|---|---|
| `handlers/read-denylist.ts` (×2) | `real === blocked \|\| real.startsWith(blocked + '/')` |
| `handlers/write-denylist.ts` | same shape |
| `hooks/bash-restriction-hook.ts` (×2) | `scanned.includes(sub)` |

The filesystem happily opens `~/.SSH/id_rsa` as `~/.ssh/id_rsa`, so this was a **real read of a real credential**, not a miss on a non-existent path.

## Approach

New `src/agent/tools/fs-case.ts` (113 LOC):

- **`isCaseInsensitiveFs()`** — probes the home volume once per process by `statSync`-ing a case-flipped spelling of its own basename and comparing `(dev, ino)`. `ENOENT` on the flipped spelling is a definitive *case-sensitive*, not an error.
- **`pathIsWithin(real, blocked)`** — replaces the bare prefix comparison.
- **`textMentionsPath(haystack, needle)`** — replaces the bash hook's substring scan.

Folding happens **only where the volume is actually case-insensitive**, exactly as the issue asks. A blanket `toLowerCase()` would be wrong on a case-sensitive volume, where `~/.SSH` and `~/.ssh` are genuinely different directories and the former deserves no protection.

### Why this is safe to put under an unconditional security floor

Every helper is **additive**: the exact comparison runs first and short-circuits, so these can only ever turn a non-match into a match, never the reverse. That one-directional property means a wrong probe result costs an **over-block** (a usability bug the operator sees instantly) and never an **under-block** (a silent credential leak nobody sees). An indeterminate probe folds — it fails closed.

### One deliberate asymmetry

The read **allow** carve-out list (`~/.afk/config/mcp.json`) stays case-sensitive. Folding it would *widen* permission rather than narrow it, which is the fail-open direction. Consequence: a case-variant spelling of a carved-out file does not receive the carve-out and is denied. That is the safe direction, and it is called out in a comment.

## Verification

Ran the issue's own reproduction artifact — `~/.afk/workspace/reports/pr-734-parity-probe.mts`, section B — re-pointed at this branch (it hardcodes absolute paths into the main checkout, so it must be re-pointed or it silently tests `main`):

| Case | `main` | this branch |
|---|---|---|
| `~/.SSH/id_rsa` | `allow \| allow` | **`BLOCK \| BLOCK`** |
| `~/.Afk/Config/afk.env` | `allow \| allow` | **`BLOCK \| BLOCK`** |
| `~/.ssh/id_rsa` | `BLOCK \| BLOCK` | `BLOCK \| BLOCK` |
| `~/.afk/state/todos/x.json` (control — must stay readable) | `allow \| allow` | `allow \| allow` |

Gates:

| Check | Result |
|---|---|
| `pnpm test src/agent/tools/` | ✅ 68 files, 2151 passed, 2 skipped |
| `pnpm lint` | ✅ |
| `audit:sdk:check` / `audit:env:check` / `scan:env:check` / `audit:chalk:check` / `fix:pins:check` | ✅ all pass |

**Falsification check:** reverting just the `read-denylist.ts` comparison back to the case-sensitive form makes 2 of the new tests fail. They are real regression tests, not tautologies.

Tests force both volume kinds via `_resetFsCaseCacheForTests` rather than reading the ambient filesystem, so they are deterministic on case-sensitive Linux CI **and** case-insensitive macOS. Asserting against the real volume would reintroduce the ambient-state dependence that made #795 flake.

## Known limitations (deliberate, not oversights)

1. **One process-wide probe.** Wrong for a mixed-volume setup (a case-sensitive APFS volume mounted beneath a case-insensitive system disk). Erring toward folding means that case over-blocks rather than under-blocks. Per-volume probing is the follow-up if it ever bites; documented in the module.
2. **`toLowerCase()` is an approximation** of the Unicode case-folding a real case-insensitive filesystem performs. Fine for ASCII credential paths; noted rather than hidden.
3. **Pre-existing 350-LOC overage untouched.** `read-denylist.ts` (440) and `bash-restriction-hook.ts` (576) were already over the repo's file-size ceiling before this change; each gains one import line. Putting the new concern in its own file is the correct move regardless, but splitting those two is separate work and is not attempted here.

Closes #736

### DIFF
--- a/src/agent/tools/fs-case.test.ts
+++ b/src/agent/tools/fs-case.test.ts
@@ -1,0 +1,118 @@
+/**
+ * Unit tests for the case-awareness helpers behind the credential floors.
+ *
+ * Covers:
+ * - `pathIsWithin` exact-match behaviour is identical on both volume kinds.
+ * - Case-variant spellings match ONLY when the volume is case-insensitive
+ *   (#736): `~/.SSH/id_rsa` must be denied on macOS-style volumes and must NOT
+ *   be conflated with `~/.ssh` on a case-sensitive one, where it is a genuinely
+ *   different directory.
+ * - The prefix boundary is respected under folding: `/home/user/.sshx` is not
+ *   inside `/home/user/.ssh`.
+ * - `textMentionsPath` mirrors the same contract for the bash hook's substring
+ *   scan.
+ * - The probe never throws and always yields a boolean.
+ *
+ * Every case forces the volume kind via `_resetFsCaseCacheForTests` rather than
+ * reading the ambient filesystem, so the suite is deterministic on
+ * case-sensitive Linux CI and case-insensitive macOS alike. (Asserting against
+ * the real volume would reintroduce exactly the ambient-state dependence that
+ * made #795 flake.)
+ *
+ * @module agent/tools/fs-case.test
+ */
+
+import { describe, it, expect, afterEach } from 'vitest';
+import {
+  pathIsWithin,
+  textMentionsPath,
+  isCaseInsensitiveFs,
+  _resetFsCaseCacheForTests,
+} from './fs-case.js';
+
+afterEach(() => {
+  _resetFsCaseCacheForTests();
+});
+
+describe('pathIsWithin', () => {
+  it('matches an exact path on either volume kind', () => {
+    for (const insensitive of [true, false]) {
+      _resetFsCaseCacheForTests(insensitive);
+      expect(pathIsWithin('/home/u/.ssh', '/home/u/.ssh')).toBe(true);
+    }
+  });
+
+  it('matches a descendant on either volume kind', () => {
+    for (const insensitive of [true, false]) {
+      _resetFsCaseCacheForTests(insensitive);
+      expect(pathIsWithin('/home/u/.ssh/id_rsa', '/home/u/.ssh')).toBe(true);
+    }
+  });
+
+  it('rejects a sibling that merely shares a prefix, on either volume kind', () => {
+    for (const insensitive of [true, false]) {
+      _resetFsCaseCacheForTests(insensitive);
+      expect(pathIsWithin('/home/u/.sshx/key', '/home/u/.ssh')).toBe(false);
+    }
+  });
+
+  // The #736 defect proper.
+  it('denies a case-variant spelling on a case-insensitive volume', () => {
+    _resetFsCaseCacheForTests(true);
+    expect(pathIsWithin('/home/u/.SSH/id_rsa', '/home/u/.ssh')).toBe(true);
+    expect(pathIsWithin('/home/u/.Ssh/id_rsa', '/home/u/.ssh')).toBe(true);
+    expect(pathIsWithin('/home/u/.AFK/config/afk.env', '/home/u/.afk/config')).toBe(true);
+    expect(pathIsWithin('/HOME/U/.SSH', '/home/u/.ssh')).toBe(true);
+  });
+
+  it('does NOT conflate case variants on a case-sensitive volume', () => {
+    _resetFsCaseCacheForTests(false);
+    expect(pathIsWithin('/home/u/.SSH/id_rsa', '/home/u/.ssh')).toBe(false);
+    expect(pathIsWithin('/home/u/.AFK/config/afk.env', '/home/u/.afk/config')).toBe(false);
+  });
+
+  it('still respects the prefix boundary when folding', () => {
+    _resetFsCaseCacheForTests(true);
+    expect(pathIsWithin('/home/u/.SSHX/key', '/home/u/.ssh')).toBe(false);
+  });
+});
+
+describe('textMentionsPath', () => {
+  it('finds an exact mention on either volume kind', () => {
+    for (const insensitive of [true, false]) {
+      _resetFsCaseCacheForTests(insensitive);
+      expect(textMentionsPath('cat /home/u/.ssh/id_rsa', '/home/u/.ssh')).toBe(true);
+    }
+  });
+
+  it('finds a case-variant mention only on a case-insensitive volume', () => {
+    _resetFsCaseCacheForTests(true);
+    expect(textMentionsPath('cat /home/u/.SSH/id_rsa', '/home/u/.ssh')).toBe(true);
+
+    _resetFsCaseCacheForTests(false);
+    expect(textMentionsPath('cat /home/u/.SSH/id_rsa', '/home/u/.ssh')).toBe(false);
+  });
+
+  it('does not match an unrelated command', () => {
+    _resetFsCaseCacheForTests(true);
+    expect(textMentionsPath('echo hello', '/home/u/.ssh')).toBe(false);
+  });
+});
+
+describe('isCaseInsensitiveFs', () => {
+  it('returns a boolean without throwing and caches the answer', () => {
+    _resetFsCaseCacheForTests();
+    const first = isCaseInsensitiveFs();
+    expect(typeof first).toBe('boolean');
+    // Second call must be served from cache — same answer, no re-probe.
+    expect(isCaseInsensitiveFs()).toBe(first);
+  });
+
+  it('fails closed: an indeterminate probe folds rather than leaving a hole', () => {
+    // A forced `true` is what the module falls back to when neither the home
+    // directory nor cwd can answer the probe. Pin the consequence: the
+    // credential floor still catches the case-variant spelling.
+    _resetFsCaseCacheForTests(true);
+    expect(pathIsWithin('/home/u/.SSH/id_rsa', '/home/u/.ssh')).toBe(true);
+  });
+});
--- a/src/agent/tools/fs-case.ts
+++ b/src/agent/tools/fs-case.ts
@@ -2,112 +2,102 @@ import { statSync } from 'node:fs';
 import { homedir } from 'node:os';
 import { basename, dirname, join } from 'node:path';
 
-/**
- * Case-awareness for the credential floors.
- *
- * Invariant: every comparison exported here is ADDITIVE — it can only ever turn
- * a non-match into a match, never the reverse. The exact (case-sensitive)
- * comparison is always tried first and short-circuits on success, so wiring
- * these helpers in can only ever DENY more than the raw `===` / `startsWith` /
- * `includes` they replace. That one-directional property is what makes it safe
- * to put them under an unconditional security floor: a wrong probe result costs
- * an over-block (a usability bug the operator sees immediately), never an
- * under-block (a silent credential leak nobody sees).
- */
+/** Case-awareness for credential-floor path comparisons. */
 
-/** Cached process-wide probe result. `undefined` = not yet probed. */
-let caseInsensitive: boolean | undefined;
+/** Test override. Real probe results are cached by filesystem device. */
+let forcedCaseInsensitive: boolean | undefined;
+const caseInsensitiveByDevice = new Map<bigint | number, boolean>();
+
+function errorCode(error: unknown): string | undefined {
+  return typeof error === 'object' && error !== null && 'code' in error
+    ? String((error as { code?: unknown }).code)
+    : undefined;
+}
 
 /**
- * Probe whether `probeDir` lives on a case-insensitive volume by asking the
- * filesystem to resolve a case-flipped spelling of its own basename and
- * comparing (dev, ino).
- *
- * Contract: returns `true` (case-insensitive) / `false` (case-sensitive) /
- * `undefined` (indeterminate — caller must fail closed). An `ENOENT` on the
- * flipped spelling is a definitive `false`, not an error: the volume declined
- * to resolve the variant, which is exactly what case-sensitivity means.
+ * Probe the filesystem containing `probePath` by resolving a case-flipped
+ * spelling of that path's basename. Indeterminate errors deliberately remain
+ * indeterminate so callers fail closed rather than caching an under-block.
  */
-function probeCaseInsensitive(probeDir: string): boolean | undefined {
-  const base = basename(probeDir);
+function probeCaseInsensitive(probePath: string): boolean | undefined {
+  const base = basename(probePath);
   const flipped = base === base.toLowerCase() ? base.toUpperCase() : base.toLowerCase();
-  // No alphabetic characters to flip — this path cannot answer the question.
   if (flipped === base) return undefined;
 
   let self;
   try {
-    self = statSync(probeDir);
+    self = statSync(probePath);
   } catch {
-    // The probe path itself is unreadable; we learned nothing.
     return undefined;
   }
 
+  const cached = caseInsensitiveByDevice.get(self.dev);
+  if (cached !== undefined) return cached;
+
   try {
-    const variant = statSync(join(dirname(probeDir), flipped));
-    return self.dev === variant.dev && self.ino === variant.ino;
-  } catch {
-    // The flipped spelling does not resolve => the volume is case-sensitive.
-    return false;
+    const variant = statSync(join(dirname(probePath), flipped));
+    const result = self.dev === variant.dev && self.ino === variant.ino;
+    caseInsensitiveByDevice.set(self.dev, result);
+    return result;
+  } catch (error) {
+    const code = errorCode(error);
+    // Only resolution failures prove that the volume rejected the variant.
+    // EACCES, EIO, ESTALE, and other errors tell us nothing and must fail closed.
+    if (code === 'ENOENT' || code === 'ENOTDIR') {
+      caseInsensitiveByDevice.set(self.dev, false);
+      return false;
+    }
+    return undefined;
   }
 }
 
 /**
- * Whether path comparisons against the credential floors should be case-folded.
+ * Whether comparisons against `protectedPath` should be case-folded.
  *
- * Probes the home directory's volume once per process and caches the result.
- * Home is the right volume to probe because essentially every denylist entry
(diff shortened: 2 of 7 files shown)


griffinwork40:
@codex fix comments

griffinwork40 (commented):

## Merge decision: **MERGE** (one medium perf regression worth a follow-up)

Reviewed ref `f01c1276c89d67cdd6114b0ed533e4785b2f9b73`. The bypass is genuinely closed and the two load-bearing safety claims in the PR body hold up under adversarial checking. No critical or high finding survived verification.

### Verified claims

**"Additive" claim — HOLDS.** `pathIsWithin` (fs-case.ts:84) and `textMentionsPath` (fs-case.ts:94) both return `true` from the exact comparison before any folding code runs. The structure is strictly `exact OR fold`, so folding turns a non-match into a match and never the reverse. A wrong probe therefore over-blocks rather than under-blocks, exactly as argued.

**"Fails closed" claim — HOLDS** on all four indeterminate paths traced: `EACCES`/`EPERM` on the probe target → `true` (fs-case.ts:30-32 → `?? true` at :68); a basename with no letters to flip → `true` (:24-25 → :68); ancestor walk reaching the filesystem root → `true` (:73); non-`ENOENT` variant errors → `true` (:50). The `ENOENT`/`ENOTDIR` branch (:46-48) returns `false` by design as a determinate "volume rejected the variant" result, not as an indeterminate one.

**Carve-out asymmetry — CORRECT.** `allow.includes(real)` (read-denylist.ts:426) is strict equality with no folding. For an allow-list that is the fail-closed direction, and a case-variant cannot satisfy strict equality, so nothing slips into the carve-out; variants fall through to the folded `builtins` loop at :427-431.

**Site coverage — COMPLETE.** All four enumerated comparison sites are converted: read-denylist.ts:418 and :428, write-denylist.ts:182, bash-restriction-hook.ts:272 and :433. A fifth site, read-denylist.ts:381 in `getReadDenylistDescendants`, is also converted — consistent with the stated intent rather than scope creep.

---

### Findings

**medium · high confidence · perf-observability · `src/agent/tools/fs-case.ts:61-68` · ref:f01c127 · citation-type: diff-context**

The per-device memo is consulted *inside* `probeCaseInsensitive` (:34-35), which is reached only *after* `isCaseInsensitiveFs` has already issued `statSync(candidate)` at :67 and `probeCaseInsensitive` its own `statSync` at :29. The cache therefore suppresses the third syscall but not the first two, so memoization does not make repeat calls free.

```ts
try {
  statSync(candidate);                          // L67 — runs on every call
  return probeCaseInsensitive(candidate) ?? true;
```

Measured on a transliteration of the module against the 21 built-in denylist entries: one `isReadDenied()` on a **not-denied** path costs **52 `statSync` calls cold and 51 warm** (~2.43 per entry). The pre-PR comparison at `f01c127^` performed pure string operations and zero `statSync` (`git show f01c127^:…/read-denylist.ts | grep -c statSync` → 0). This fires on every typed read via `_cwd-utils.ts:214`, not only under load.

*Suggestion:* memoize `isCaseInsensitiveFs` on `protectedPath` at function entry, or hoist one `isCaseInsensitiveFs()` call outside the `extras`/`builtins` loops in `isReadDenied` (:417, :427) instead of re-deriving per entry.

**low · high confidence · correctness · `src/agent/tools/handlers/read-denylist.ts:387-389` · ref:f01c127 · citation-type: file-state**

A `blocked` entry that is a case-variant *equal in length* to `realRoot` passes the filter at :379-382 (`blocked !== realRoot` is true case-sensitively; `pathIsWithin` folds to true), then takes the `slice` branch because :387 is case-sensitive:

```ts
const rel = blocked.startsWith(realRoot + sep)
  ? relative(realRoot, blocked)
  : blocked.slice(realRoot.length + 1);
```

With `realRoot=/home/u/.ssh` (12 chars) and `blocked=/home/u/.SSH` (12 chars), `slice(13)` yields `""`, and grep.ts:167 would emit `-g '!'` and `-g '!/**'` — the latter excluding every file.

**This is latent, not live:** grep.ts:162 passes the already-contained `resolvedPath`, and `resolveAndContain(…, 'read')` routes through `isReadDenied` at `_cwd-utils.ts:214`, which throws on that same folded match before `getReadDenylistDescendants` is reached. The case-variant *descendant* case is correct — `/home/u/READABLE` vs `/home/u/readable/.ssh` slices to `.ssh` as intended.

*Suggestion:* add `if (!rel) continue;` after :389 so correctness does not depend on caller ordering.

**low · medium confidence · security · `src/agent/tools/fs-case.ts:38-48` · ref:f01c127 · citation-type: diff-context**

```ts
if (code === 'ENOENT' || code === 'ENOTDIR') {
  caseInsensitiveByDevice.set(self.dev, false);
  return false;
}
```

`:29` and `:38` are two unlocked syscalls. If the probed entry is deleted or renamed in that window on a case-insensitive volume, `:38` returns `ENOENT` — indistinguishable from genuine case-sensitivity — and `false` is cached device-wide, with no TTL or production invalidation (`_resetFsCaseCacheForTests` is test-only, :99-103). Folding would then be disabled for later comparisons on that device.

A shadow verifier independently re-derived this and **confirmed the mechanism but rejected a high rating**: no non-race path to a wrong `false` exists (on a case-insensitive volume the flipped spelling resolves to the same inode, and two case-differing entries cannot coexist), and exploitation requires deleting a protected directory such as `~/.ssh` in a microsecond window between two adjacent syscalls — implying more direct compromise already. Rated low as theoretical hardening, not a live hole.

*Suggestion:* require a second corroborating `ENOENT`, or re-stat the parent directory, before caching `false`.

### Verified-correct (no action)

- Prefix boundary survives folding — the folded branch reuses `b + '/'` (:89), so `/home/u/.sshx` does not match `/home/u/.ssh`.
- The `slice` offset at :389 uses untransformed `.length`; `toLowerCase()` is confined to the containment test at :87-88, so a length-changing fold (Turkish dotted İ) cannot desync it.
- The three declared known limitations in the PR body are accurately described, including the pre-existing file-size overage (read-denylist.ts is 451 lines and bash-restriction-hook.ts 577 at this ref).

### What was not checked

- Citations verified inline against branch HEAD `f01c1276c89d67cdd6114b0ed533e4785b2f9b73` via `git show <ref>:<file>`; every quoted line above was re-read at that ref.
- Stated intent: PR #819 title + body — spec-compliance assessed against it.
- The syscall counts come from a **faithful transliteration** of `fs-case.ts` driven by the 21 reconstructed built-in entries, not from instrumenting the compiled module. The direction and rough magnitude are solid; treat the exact integers as approximate.
- Did not run the test suite, did not re-run the issue's repro probe, and did not verify the PR's falsification claim by actually reverting the comparison.
- Did not review the Telegram surface or any consumer outside `read-denylist` / `write-denylist` / `bash-restriction-hook` / `grep.ts`.

griffinwork40 (commented):

## Merge decision: **MERGE**

Light regime (hotfix, 7 files, +303/-9). Reviewed ref `f01c127`. The core security claim holds under adversarial checking: all five comparison sites are converted, the helpers are genuinely additive, and the probe fails closed on every branch I traced. Findings below are `medium` and under — none blocking.

### Verified claims (each re-checked against the reviewed ref, not the diff alone)

- **The "additive" invariant is real.** `pathIsWithin` and `textMentionsPath` are a monotonic OR — the exact comparison returns first, and the fold clause runs only after it already failed, so folding can turn a non-match into a match but never the reverse (`fs-case.ts:83-89`, `:93-97`). No fail-open path exists in either helper.
- **The probe fails closed on every branch.** Non-ENOENT throw folds (`:71`), root-walk exhaustion folds (`:73`), indeterminate folds via `?? true` (`:68`). The case-invariant-basename edge case (a basename with no alphabetic characters, where a case-flip yields an identical string and the probe would otherwise `stat` the *same* path and wrongly conclude case-insensitive) is preempted before any `dev`/`ino` comparison by `if (flipped === base) return undefined;` (`:25`).
- **All five sites converted** — `read-denylist.ts:381,418,428`, `write-denylist.ts:182`, `bash-restriction-hook.ts:272,433`. Zero survivors of the old `startsWith(blocked + '/')` / `scanned.includes(sub)` shape.
- **The allow carve-out asymmetry is implemented as described.** `read-denylist.ts:426` is still a plain case-sensitive `allow.includes(real)`, so a wrong probe can only shrink the allow-set, never widen it. Fail-safe direction confirmed.

---

### medium · high confidence · correctness · `src/agent/tools/handlers/read-denylist.ts:387-389` · ref:`f01c127` · citation-type:file-state

**Finding:** `String.prototype.toLowerCase()` is not length-preserving, so when containment is established by the *folded* branch, the byte-offset arithmetic in `blocked.slice(realRoot.length + 1)` under-slices and emits a corrupted descendant suffix, which becomes a malformed ripgrep exclude glob that fails to prune the protected subtree.

**Evidence** (the code comment asserting the property that does not hold):

```ts
// `path.relative` is case-sensitive even when the filesystem is not.
// Containment above established that a folded spelling is equivalent;
// slicing preserves the denylist entry's glob-ready descendant suffix.
const rel = blocked.startsWith(realRoot + sep) ? relative(realRoot, blocked) : blocked.slice(realRoot.length + 1);
```

`U+0130` (LATIN CAPITAL I WITH DOT ABOVE) lowercases to two code units, so a root containing it is 9 units but 11 folded. Reproduced end-to-end against ripgrep 15.1.0: the emitted glob was `!\u0307/.sec` instead of `!.sec`, and the protected file **was searched and its contents returned**. Production-reachable via `grep.ts:162`, which feeds the suffix straight into `-g !<literal>`.

**Scope — this is not a regression.** I replayed `main`'s prior implementation: for the same case-variant root it returned `[]`, emitting no exclusion at all. Both old and new leak on this input, so the PR is strictly an improvement (it fixes the ASCII case-variant, which `main` also leaked). The direct `read_file` floor is unaffected — `isReadDenied` uses `pathIsWithin` with no slicing.

**Relation to declared limitation #2.** The PR body discloses that `toLowerCase()` approximates Unicode folding, which frames the failure as a *missed match*. The actual mechanism here is the opposite: the match **succeeds** and the downstream arithmetic then corrupts the output. The territory is declared; the mechanism is not.

**Suggestion:** on the folded branch, derive the suffix by counting path segments rather than code units — e.g. `blocked.split(sep).slice(realRoot.split(sep).length).join('/')` — which is invariant to fold-induced length changes.

---

### low · medium confidence · test-coverage · `src/agent/tools/fs-case.test.ts:103-109` · ref:`f01c127` · citation-type:file-state

**Finding:** the only test that exercises the unforced `isCaseInsensitiveFs` asserts return type and cache idempotency, never correctness, leaving `probeCaseInsensitive`'s `statSync` + `dev`/`ino` comparison — the one piece of real filesystem logic in the module — covered by zero assertions.

**Evidence:**

```ts
it('returns a boolean without throwing and caches the answer', () => {
  const first = isCaseInsensitiveFs();
  expect(typeof first).toBe('boolean');
  expect(isCaseInsensitiveFs()).toBe(first);
```

Absence claim grounded: `probeCaseInsensitive` returns zero matches across `fs-case.test.ts`, `read-denylist.test.ts`, and `write-denylist.test.ts` at the reviewed ref. The forced-volume tests around it are real regression tests reproducing #736 against production folding logic, not tautologies — this gap is narrow and specific to the real probe.

**Suggestion:** add a platform-gated test that creates a temp dir and asserts the `dev`/`ino` branch directly on the ambient volume.

---

### nit · high confidence · docs · PR description · ref:`f01c127` · citation-type:diff-context

**Finding:** the PR body's "Known limitations" #1 understates the implementation — it claims "one process-wide probe … wrong for a mixed-volume setup," but the code caches per filesystem device and probes the nearest existing ancestor of each protected path, which handles exactly the mixed-volume case the body says is broken.

**Evidence:**

```ts
const caseInsensitiveByDevice = new Map<bigint | number, boolean>();
// The nearest existing ancestor is probed, so relocated/custom denylist paths
// use their own mounted filesystem rather than an unrelated process-wide home volume.
```

The discrepancy is in the safe direction (code better than documented), but the PR body is the merge record. Separately, the body says "four comparison sites" while its own table sums to five (2+1+2).

**Suggestion:** update limitation #1 to describe the per-device cache, and correct the site count to five.

---

### What was not checked

- **Citations verified inline against branch HEAD `f01c1276c89d67cdd6114b0ed533e4785b2f9b73`**, fetched via the GitHub contents API at that SHA — not against `main` and not against a local working tree. All four load-bearing citations verified; zero fabricated, zero diff-only.
- **Stated intent: PR #819 title + body — spec-compliance assessed.** No unmet intent found (all named sites converted); no scope creep found.
- Did **not** run the test suite, the author's reproduction artifact, or CI. CI is green on this ref (all 8 checks SUCCESS/SKIPPED) but I did not independently re-execute it.
- Did **not** review the Telegram or TUI surfaces, or audit callers of `getReadDenylistDescendants` beyond `grep.ts`.
- The Unicode reproduction was run against a synthetic fixture with ripgrep 15.1.0 on macOS, not against a real case-insensitive-volume credential path.
- `/shadow-verify` was not invoked: it triggers on `critical`/`high` findings, and none were raised.

griffinwork40:
@codex fix medium · high confidence · correctness · src/agent/tools/handlers/read-denylist.ts:387-389 · ref:f01c127 · citation-type:file-state
Finding: String.prototype.toLowerCase() is not length-preserving, so when containment is established by the folded branch, the byte-offset arithmetic in blocked.slice(realRoot.length + 1) under-slices and emits a corrupted descendant suffix, which becomes a malformed ripgrep exclude glob that fails to prune the protected subtree.

Evidence (the code comment asserting the property that does not hold):

// `path.relative` is case-sensitive even when the filesystem is not.
// Containment above established that a folded spelling is equivalent;
// slicing preserves the denylist entry's glob-ready descendant suffix.
const rel = blocked.startsWith(realRoot + sep) ? relative(realRoot, blocked) : blocked.slice(realRoot.length + 1);
U+0130 (LATIN CAPITAL I WITH DOT ABOVE) lowercases to two code units, so a root containing it is 9 units but 11 folded. Reproduced end-to-end against ripgrep 15.1.0: the emitted glob was !\u0307/.sec instead of !.sec, and the protected file was searched and its contents returned. Production-reachable via grep.ts:162, which feeds the suffix straight into -g !<literal>.

Scope — this is not a regression. I replayed main's prior implementation: for the same case-variant root it returned [], emitting no exclusion at all. Both old and new leak on this input, so the PR is strictly an improvement (it fixes the ASCII case-variant, which main also leaked). The direct read_file floor is unaffected — isReadDenied uses pathIsWithin with no slicing.

Relation to declared limitation https://github.com/griffinwork40/agent-afk/pull/2. The PR body discloses that toLowerCase() approximates Unicode folding, which frames the failure as a missed match. The actual mechanism here is the opposite: the match succeeds and the downstream arithmetic then corrupts the output. The territory is declared; the mechanism is not.

Suggestion: on the folded branch, derive the suffix by counting path segments rather than code units — e.g. blocked.split(sep).slice(realRoot.split(sep).length).join('/') — which is invariant to fold-induced length changes.

PR #835, opened 2026-08-01
fix(permissions): reject newline/CR in rm-allowlist carve-out (#806 follow-up)

griffinwork40:
## What

Follow-up to a code review of the already-merged **#806** (`fix(permissions): allow curated in-workspace rm -rf in AFK autonomous mode (#579 O3)`).

`src/agent/afk-mode-rm-allowlist.ts` gates whether an AFK-autonomous-mode `rm -rf <dir>` bash command is auto-allowed without operator approval. Part of that gate rejects any command containing a shell metacharacter, via:

```ts
const SHELL_METACHAR = /[*?$`(){};|&<>'"\\]/;
```

The bash tool executes commands with `child_process.spawn(command, { shell: true })` — a real `/bin/sh -c`. A **literal embedded newline** in the command string is a shell statement separator, functionally identical to `;` — but the character class above did not include `\n` or `\r`. Meanwhile `parseRmCommand`'s tokenizer (`cmd.trim().split(/\s+/)`) treats a newline as ordinary whitespace (JS `\s` matches newline), so a two-statement command was silently flattened into one token stream.

**Concrete gap:** for `cmd = "rm -rf node_modules\ndist"` (a real embedded newline, not the two-character `\n` escape), tokenization yielded `targets: ["node_modules", "dist"]`, `recursive: true`. If `dist` happened to be a real, git-ignored directory in the workspace, `isSafeInWorkspaceRm` returned `true` — even though a real shell would execute this as **two statements** (`rm -rf node_modules`, then attempt to run `dist` as a command), not as a single validated `rm` call. That breaks the module's own documented contract: "Fails CLOSED on any ambiguity."

**Why it matters (impact, honestly scoped):** this is a soundness gap in a fail-closed security boundary and should be fixed regardless. That said, today's demonstrated impact is bounded — no working example in this PR (or found during review) causes deletion or command execution beyond what the carve-out already intentionally permits, since in the common case the shell would just fail trying to run the second "statement" (e.g. `dist`) as a command, not execute it as something harmful. The fix closes the gap before a more dangerous second statement becomes possible.

## What changed

Minimal, two-file diff:

1. **`src/agent/afk-mode-rm-allowlist.ts`** — added `\r` and `\n` to the `SHELL_METACHAR` character class, so a command containing either is rejected before tokenization ever runs (`parseRmCommand` returns `null` → `isSafeInWorkspaceRm` returns `false`). The command stays classified `high` and is gated exactly as every other metacharacter already was — same fail-closed behavior, no new code path. Updated the doc comment directly above the regex to explain the `spawn(..., { shell: true })` / statement-separator reasoning. No other parsing logic touched.

2. **`src/agent/afk-mode-gate.test.ts`** — new `it.each` regression block (matching the file's existing style) asserting:
   - `rm -rf node_modules\ndist` (embedded LF) → `result.decision === 'block'`
   - `rm -rf node_modules\r\ndist` (embedded CRLF) → `result.decision === 'block'`

   Both use a real newline character, not an escaped two-character string, and neither test constructs a `dist` directory on disk — the fix rejects at the metacharacter-check stage, before any per-target filesystem check runs, so the test is independent of whether `dist` exists.

No dedicated `afk-mode-rm-allowlist.test.ts` file exists — the allowlist module's behavior is exercised exclusively through `afk-mode-gate.test.ts`, consistent with prior PRs on this module.

## Verification

- `pnpm install` — clean (fresh worktree, no shared `node_modules`)
- `pnpm test src/agent/afk-mode-gate.test.ts` — **101/101 passed** (99 pre-existing + 2 new)
- `pnpm lint` (`tsc --noEmit`, strict) — clean, zero errors
- Manually confirmed the vulnerability and fix at the regex level:
  - old `SHELL_METACHAR.test("rm -rf node_modules\ndist")` → `false` (gap confirmed)
  - new `SHELL_METACHAR.test("rm -rf node_modules\ndist")` → `true` (fix confirmed)

## Scope

Touches only the two files above. No changes to `afk-mode-gate.ts`, no restructuring of the surrounding parsing logic, no changes to PR #806 itself.

### DIFF
--- a/src/agent/afk-mode-gate.test.ts
+++ b/src/agent/afk-mode-gate.test.ts
@@ -111,6 +111,23 @@ describe('createAfkModeGate', () => {
     expect(result.decision).toBe('block');
   });
 
+  // ---- PR #806 follow-up review: embedded newline is a statement separator ---
+  // `spawn(cmd, { shell: true })` runs the command through a real `/bin/sh -c`,
+  // where a literal newline separates statements exactly like `;`. The
+  // tokenizer's `\s+` split treats `\n`/`\r` as ordinary whitespace, so without
+  // rejecting them at the metachar-check stage a command like
+  // `rm -rf node_modules\ndist` would flatten into a single token stream
+  // (`targets: ["node_modules", "dist"]`) instead of being recognized as two
+  // separate shell statements.
+  it.each([
+    ['rm -rf node_modules\ndist', 'embedded LF'],
+    ['rm -rf node_modules\r\ndist', 'embedded CRLF'],
+  ])('blocks rm -rf with an embedded newline statement separator (%s)', async (command) => {
+    const { gate } = makeGate('autonomous');
+    const result = await gate({ event: 'PreToolUse', toolName: 'bash', input: { command } });
+    expect(result.decision).toBe('block');
+  });
+
   // ---- PR #806 review: carve-out is recursive DIRECTORY deletes only ---------
   // Without a recursive flag the target is a regular file — a tracked script
   // named `build` is not a generated artifact and must still require approval.
--- a/src/agent/afk-mode-rm-allowlist.ts
+++ b/src/agent/afk-mode-rm-allowlist.ts
@@ -40,8 +40,15 @@ const RM_RF_SAFE_LEAF_DIRS: ReadonlySet<string> = new Set([
 ]);
 
 /** Shell metacharacters whose presence makes tokenization unreliable (glob
- *  expansion, variable interpolation, command substitution, chaining). */
-const SHELL_METACHAR = /[*?$`(){};|&<>'"\\]/;
+ *  expansion, variable interpolation, command substitution, chaining). `\r`
+ *  and `\n` are included too: the bash tool runs commands via
+ *  `spawn(cmd, { shell: true })`, a real `/bin/sh -c`, where a literal
+ *  embedded newline is a statement separator identical to `;`. Tokenization
+ *  below (`cmd.trim().split(/\s+/)`) treats `\n` as ordinary whitespace, so
+ *  without this a command like `rm -rf node_modules\ndist` — two shell
+ *  statements — would flatten into one token stream (`targets: ["node_modules",
+ *  "dist"]`) and be evaluated as if it were a single validated `rm` call. */
+const SHELL_METACHAR = /[*?$`(){};|&<>'"\\\r\n]/;
 
 /**
  * True for a recursive-delete flag: `-r`, `-R`, any clustered short-flag run


griffinwork40:
Both checks pass:
- **Check A (citation):** the cited `it.each` block (lines 122-125) matches verbatim at ref `ea6b69b9` → `verified`.
- **Check B (absence):** grep for `node_modules\rdist` returns no matches → absence `confirmed-absent`. Finding stands.

No findings dropped or downgraded; no cross-agent conflicts. The single surviving finding is `low` — below the `critical`/`high` threshold, so no `/shadow-verify` round is required. Synthesizing directly.

---

# Review #835 — Synthesis

**PR:** `fix(permissions): reject newline/CR in rm-allowlist carve-out (#806 follow-up)`
**Branch:** `afk/fix-rm-allowlist-newline-gap` → `main`
**Reviewed ref:** `ea6b69b9`
**Stats:** +26 / -2 / 2 files · regime: **light**

**Stated intent:** Add `\r` and `\n` to `SHELL_METACHAR` so an `rm -rf` command containing an embedded newline/CR is rejected at the metachar gate (before tokenization flattens it into one token stream), because `spawn(cmd, { shell: true })` runs via `/bin/sh -c` where a literal newline is a statement separator like `;`. Add LF + CRLF regression tests.

## Findings

| # | severity | dim | file:line | finding | status |
|---|----------|-----|-----------|---------|--------|
| F1 | low | test-coverage | afk-mode-gate.test.ts:122-125 | A bare lone CR (`rm -rf node_modules\rdist`, no trailing LF) is now rejected by the widened regex (the `\r` member at afk-mode-rm-allowlist.ts:51) but no regression test covers it — the two `it.each` rows exercise LF and CRLF only. | citation `verified` · absence `confirmed-absent` |

### F1 detail
- **Finding:** A bare lone CR (`rm -rf node_modules\rdist`, no trailing LF) is now rejected by the widened regex (the `\r` member at afk-mode-rm-allowlist.ts:51) but no regression test covers it — the two `it.each` rows exercise LF and CRLF only.
- **Evidence:** `['rm -rf node_modules\ndist', 'embedded LF'], ['rm -rf node_modules\r\ndist', 'embedded CRLF'],` — no `\rdist` (lone CR) row.
- **Suggestion:** add a third `it.each` row: `['rm -rf node_modules\rdist', 'embedded lone CR']` asserting `decision === 'block'`.
- **Verification:** citation re-read at ref `ea6b69b9` lines 122-125 (verified); absence of lone-CR test confirmed by grep of the production+test tree (confirmed-absent).
- **Assessment:** Additive coverage, not a defect in shipped behavior — the regex *does* reject bare CR (fail-closed posture holds regardless of whether bare `\r` is a real statement separator in `/bin/sh`). The intent names "CR", and the implementation honors it; only the test matrix is incomplete. Non-blocking.

## Dimensions with no issues
- **security** — regex is syntactically valid; in a JS char class `\\` is a literal backslash and `\r`/`\n` evaluate to the CR/LF literals (not two-char sequences). Gate runs at line 89 before tokenization; `parseRmCommand` returns null on metachar hit → `isSafeInWorkspaceRm` returns false (lines 153-154) → command stays `high` and blocked. Fail-closed posture preserved.
- **correctness** — the metachar check intercepts the raw string before `cmd.trim().split(/\s+/)`, so the flattening described in the intent is prevented at the right stage.
- **api-compat** — `SHELL_METACHAR` is module-private (no `export`), single in-file use at line 89; `isSafeInWorkspaceRm` signature unchanged, sole production importer `afk-mode-gate.ts:351` unaffected. No external caller observes the regex widening.
- **spec-compliance** — intent fully met (`\r`+`\n` added, LF+CRLF tests added). No unmet intent clauses; no scope creep — char-class widening only, no other parse path altered. The space-separated multi-target allow case (`rm -rf node_modules dist`, line 69) still passes because a space is not in the class.
- **perf-observability** — a two-character addition to a single-test regex run once per high-recursive-rm classification has no measurable perf surface.

## Confidence
- F1: high confidence (verified citation + confirmed absence). Low severity → no tier downgrade.

- Test suite was not executed by this review — the PR's "101/101 passed" is asserted in the PR body, not re-verified.
- `/bin/sh` semantics for a bare lone `\r` as a true statement separator (vs. a word char) were not verified — noted as moot since the regex rejects it regardless (fail-closed).
- `afk-mode-gate.ts` was not read beyond grep hits (lines 90/340/351); the upstream `high`-classification-then-carve-out flow is taken as described in the module's own comments, not independently traced.
- No `/shadow-verify` round (no `critical`/`high` findings present).
- Citations verified inline against branch HEAD `ea6b69b9`. Stated intent: PR #835 title+body — spec-compliance assessed.

## Merge decision

**MERGE.**

The fix is correct, minimal, and purely additive (a character-class widening that tightens a fail-closed security boundary with no new code path). It meets its stated intent; the regex now rejects embedded LF, CRLF, and bare CR at the metachar gate before tokenization. `SHELL_METACHAR` is module-private so no caller regresses, and the exported `isSafeInWorkspaceRm` contract is unchanged. The lone finding (F1) is a low-severity missing lone-CR test row — additive coverage strengthening an already-correct behavioral guarantee, not a defect that blocks merge.

**Done**
- Reviewed PR #835 (`fix(permissions): reject newline/CR in rm-allowlist carve-out`) across all dimensions via one light-regime sub-agent + inline Wave 1.5 citation/absence verification.
- Evidence: this synthesized report; reviewed ref `ea6b69b9`; verified citations against `git show ea6b69b9:src/agent/afk-mode-gate.test.ts` lines 122-125; confirmed lone-CR test absence via `grep -rn 'node_modules\rdist' src/`.
- Merge recommendation: **MERGE** — one non-blocking `low` finding (missing bare-lone-CR regression test row, `it.each` at afk-mode-gate.test.ts:122-125), suggested fix attached. The working tree's 1 uncommitted change was left untouched (read-only review).

---

PR #837, opened 2026-08-01
fix(slash): alias /add-dir to /allow-dir

griffinwork40:
## Problem

Operators arriving from Claude Code type `/add-dir` from muscle memory. It resolved to nothing — and **both** discovery paths independently failed to redirect them, so the command silently dead-ended.

Reported symptom: *"`/add-dir` doesn't show up as a command when I start to type it and also it doesn't work."*

Both halves have distinct causes:

**1. It never appeared in the autocomplete dropdown.** `filterSlashCandidates` (`src/cli/input/trigger.ts:93`) matches prefix first, then subsequence as a fallback. `"allow-dir".startsWith("add-dir")` is false, and the subsequence check needs three `d`s where `allow-dir` has one. Executed against the real registry on `main`:

```
completion "add-dir" : []
completion "add"     : ["/audit-dispatch"]     <- /allow-dir never surfaces
```

**2. The "did you mean?" hint could not rescue it either.** `suggest()` (`src/cli/slash/registry.ts:162`) uses Levenshtein with `maxDistance = 3`, but `editDistance('/add-dir', '/allow-dir')` is **exactly 4** — one edit past the cutoff:

```
suggest("/add-dir")    : undefined      <- bare error, no pointer
suggest("/add-dir", 4) : "/allow-dir"
```

Net effect: `Unknown command: /add-dir  (type /help for commands)` with no path to the command that does exist. A near-miss name sitting one edit unit outside its own rescue mechanism.

## Fix

Add `aliases: ['/add-dir']` to `allowDirCmd`. This repairs both symptoms in one additive change — the dropdown universe unions `aliasEntries()` (`trigger.ts:114`) and `lookup()` is alias-aware (`registry.ts:135`).

**Why an alias and not a wider threshold:** bumping `maxDistance` to 4 to catch this one case would loosen suggestions for all 90+ other commands. The alias is the narrow fix; the threshold is left alone. This rationale is pinned by a test so a future reader does not "simplify" it into a threshold bump.

## Verification

Full registry, post-fix:

```
registerAll() did not throw — no alias collision across 93 commands
lookup("/add-dir")   : /allow-dir
completion "add-dir" : ["/add-dir"]
completion "add"     : ["/add-dir","/audit-dispatch"]
```

- `npx vitest run src/cli/slash/commands/allow-dir.test.ts` -> **17 passed** (12 pre-existing + 5 new)
- `npx vitest run src/cli/slash src/cli/input src/cli/slash-registry.test.ts src/cli/index-integration.test.ts src/cli/input-box.test.ts` -> **71 files / 1288 tests passed**
- `npx tsc --noEmit` -> exit 0

Collision safety is mechanically established, not assumed: `register()` throws on alias collision (`registry.ts:37-41`), and `registerAll()` completes cleanly across all 93 commands.

## Scope

Two files, +60 lines, no behavior change to any existing path — `/allow-dir` and its flags are untouched. Purely additive discoverability.

## Note for reviewers

`README.md:231` documents a retired `/bypass off` command (Shift+Tab is the live toggle now) — spotted during this investigation, **not** fixed here to keep the diff single-purpose.

### DIFF
--- a/src/cli/input/trigger.test.ts
+++ b/src/cli/input/trigger.test.ts
@@ -22,6 +22,7 @@ import {
   filterFileCandidatesAsync,
   filterFileCandidatesCached,
   filterSlashCandidates,
+  filterFlagCandidates,
   buildFileCandidates,
   invalidateFileScanCache,
   __fileScanCacheSize,
@@ -175,6 +176,57 @@ describe('filterSlashCandidates', () => {
     const vals = filterSlashCandidates('').map((c) => c.value);
     expect(vals.length).toBeGreaterThan(0);
     expect(vals).toContain('/config');
+  });
+});
+
+describe('detectTrigger — flag completion for aliases', () => {
+  it('fires flag completion for a canonical command name', () => {
+    const buf = '/allow-dir --rw';
+    const trigger = detectTrigger(buf, buf.length);
+    expect(trigger).not.toBeNull();
+    expect(trigger!.kind).toBe('flag');
+    if (trigger!.kind === 'flag') {
+      expect(trigger!.command).toBe('allow-dir');
+    }
+  });
+
+  it('fires flag completion for an alias (regression: /add-dir --)', () => {
+    // Before the fix, detectTrigger searched by canonical name only, so the
+    // alias /add-dir never matched and flag completion silently dropped.
+    const buf = '/add-dir --rw';
+    const trigger = detectTrigger(buf, buf.length);
+    expect(trigger).not.toBeNull();
+    expect(trigger!.kind).toBe('flag');
+    if (trigger!.kind === 'flag') {
+      expect(trigger!.command).toBe('add-dir');
+    }
+  });
+
+  it('does not fire flag completion for a command without flags', () => {
+    const buf = '/config --xyz';
+    const trigger = detectTrigger(buf, buf.length);
+    // /config has no flags, so detectTrigger returns null for the flag path
+    // (it falls through to null rather than a slash trigger because of the
+    // trailing whitespace + -- token).
+    expect(trigger).toBeNull();
+  });
+});
+
+describe('filterFlagCandidates — alias-aware', () => {
+  it('returns flags for a canonical command name', () => {
+    const flags = filterFlagCandidates('allow-dir', '--r').map((c) => c.value);
+    expect(flags).toContain('--rw');
+    expect(flags).toContain('--revoke');
+  });
+
+  it('returns flags for an alias (regression: add-dir → allow-dir flags)', () => {
+    const flags = filterFlagCandidates('add-dir', '--r').map((c) => c.value);
+    expect(flags).toContain('--rw');
+    expect(flags).toContain('--revoke');
+  });
+
+  it('returns [] for an unknown command', () => {
+    expect(filterFlagCandidates('nonexistent', '--x')).toEqual([]);
   });
 });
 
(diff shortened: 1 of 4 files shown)


griffinwork40:
Wave 1.5 complete. Verification manifest:

|---|------|-------|--------|----------|
| 1 | citation | `allow-dir.test.ts:245-248` test labeled "dispatch path" calls `lookup` not `dispatch` | ✅ **verified** — content matches at ref e4757d7 |
| 2 | absence | No existing command declares `/add-dir` as canonical name or alias | ✅ **confirmed-absent** — grep of production source returns zero matches |

No findings dropped or downgraded. The single `low` finding passes to synthesis. Since no `critical` or `high` findings surfaced, shadow-verify is not triggered.

---

## PR #837 Review — `fix(slash): alias /add-dir to /allow-dir`

**Ref:** `e4757d7` · **Diff:** 2 files, +60/−0 · **Regime:** light (hotfix)

### Dimensions

| Dimension | Verdict |
|-----------|---------|
| **Security** | No issues — alias is a pure name→canonical mapping; dispatch reuses the same `allowDirCmd.handler`, no new code path |
| **Correctness** | No issues — alias-aware `lookup` (registry.ts:135-139) ✅; dropdown unions `aliasEntries()` (trigger.ts:112-115) ✅; `suggest()` does not enumerate aliases so the threshold-doc test is valid ✅ |
| **Spec-compliance** | Fully met — `aliases: ['/add-dir']` delivers both dispatch + dropdown repair; zero scope creep (threshold untouched, 8 of 11 source lines are comment) |
| **api-compat** | No issues — `aliases` is established convention (6 prior commands: `/t`, `/edit`, `/quit`, `/ctx`, `/branch`, `/thinking-ui`); no `/add-dir` collision in production (confirmed-absent); `register()` throws loudly on bidirectional collision |
| **Perf/observability** | No issues — one extra Map entry per keystroke-path universe rebuild; negligible |

### Findings

**1 low** · `test-coverage` · `allow-dir.test.ts:245-248` · ref:e4757d7 · file-state ✅

The test bears the label *"resolves `/add-dir` to the `/allow-dir` command (dispatch path)"* but calls `lookup()` directly — it does not exercise `dispatch('/add-dir …', ctx)` end-to-end. `dispatch` resolves through `lookup` (registry.ts:202) by construction, so the alias works at dispatch time, but no test asserts the handler actually fires.

- **Evidence:** `it('resolves /add-dir to the /allow-dir command (dispatch path)', () => { expect(lookup('/add-dir')).toBe(allowDirCmd); ... })`
- **Suggestion:** add `expect((await dispatch('/add-dir', ctx)).handled).toBe(true)` (or assert the grant side-effect) to genuinely cover the dispatch integration.

---

### What was not checked
- Citations verified inline against branch HEAD `e4757d7`. The lone finding (low) citation confirmed.
- Test suite not run (read-only review). The PR body reports 17 passed / 1288 tests across the broader suite.
- Stated intent: **available** (PR title + body) — spec-compliance assessed.

### Merge recommendation: **MERGE**

The change is purely additive, backed by verified implementation paths, convention-following (6 prior aliases), collision-safe (grep-confirmed absent, loud-throw registry guard), and well-tested. The single `low` finding — a test named "dispatch path" that calls `lookup` instead of `dispatch` — is a belt-and-suspenders coverage gap, not a correctness risk (the dispatch→lookup delegation makes the alias resolve by construction). Nothing blocks merge.

---

griffinwork40:
Fixed the codex P2 in 9f62cb4. Both `detectTrigger()` and `filterFlagCandidates()` now route through the alias-aware `lookup()` from `registry.ts` instead of searching `listSlashCommands()` by canonical name, so `/add-dir --` correctly surfaces `--rw` and `--revoke` in the flag dropdown. Added regression tests in `trigger.test.ts`.

PR #842, opened 2026-08-01
fix(banner): stop pixel art sharing rows with text so resize cannot tear it

griffinwork40:
## Problem

Reported as: *"sometimes the goblin / afk in the banner gets mangled on resize."*

The welcome banner is written once into scrollback before the compositor arms, and is never re-derived at a new width. On a **shrink**, the emulator hard-splits any stored row wider than the new width. Session text (branch, cwd, mode) rode *beside* the goblin, so those rows were up to `cols` wide while sprite-only rows were 29 — a shrink therefore split **some** rows of a fixed 13-row grid and injected text fragments into the goblin's face.

Reproduced at 100 → 60 cols against a real emulator buffer (`@xterm/headless`):

```
[10] wrapped=n |     ▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀       branch  afk/subagent-stream-c|
[11] wrapped=Y |ut-retry|                        <- fragment injected into the art
[12] wrapped=n |     ▀▀▀▀▀▀▼▀▀▀▀▀▀▀▀▀▀      ~/Projects/personal_projec…wo|
[13] wrapped=Y |rkspace/agent-afk-private/deeply/nested|
[14] wrapped=n |        ▀▀▀▀▀▀▀▀▀▀▀|             <- jaw row, pushed down 2
```

The *"sometimes"* is a shrink that crosses the widest row: a deep cwd makes almost any shrink tear it, a shallow cwd needs a large one.

## Fix

The only thing allowed to share a terminal row with the sprite is more art (the block-art hero). Every readable string moves to full-width rows below the composition.

Widest art row is now `LEFT_PAD + MASCOT_WIDTH + GUTTER + hero = 2+27+2+14 = 45` cols, so the art survives any resize down to 45 — and the mascot is already dropped below 56 (`MIN_INFO_COLS`), which covers its entire live range. Text rows below still reflow, which is intended: wrapped prose degrades gracefully, torn pixel art does not.

Also reserves the terminal's final column (`cols - 1`), matching the last-column-safety convention in `render/card.ts` and `input/echo.ts` that the banner alone violated.

## A correction worth recording

My first diagnosis was **wrong**, and the code now says so explicitly so nobody re-derives it. I attributed the tearing to the DECAWM deferred-wrap hazard — a row exactly `cols` wide acquiring the soft-wrap bit and being rejoined with its neighbour on resize. That mechanism does not reach these rows:

- Banner rows are CR/LF-terminated, and CR/LF resets the last-column flag (DEC STD-070).
- xterm.js sets `isWrapped` only when `print()` places a glyph *past* the last column, and `BufferReflow` only rejoins lines already flagged wrapped.

`welcome-banner.resize.test.ts` asserts **nothing is wrapped at print time**, which is the assertion that falsifies that theory. The `cols - 1` reserve is kept for convention consistency and one column of shrink headroom — not as the fix.

## Tradeoffs

- **+4 rows of height (17 → 21).** Inherent to the approach: the goblin's lower rows no longer carry facts for free. The caption folds identity + version + model + mode onto one row to limit it.
- The hero now centres on the 13-row sprite (`round((13-5)/2) = 4`), landing beside the ears/eyes rather than the cap tip — which is what the layout comment always said it wanted.
- Mascot drop boundary 55 → 56 cols. Untruncated-tagline floor 44 → 45 (the 42-col tagline previously fit at 44 *only* by occupying the final column, i.e. the thing being fixed).

## Verification

- `welcome-banner.resize.test.ts` — replays the **real** banner into `@xterm/headless` and asserts art rows stay byte-identical across shrinks (140→56/64/72/80/96, 120→45, and the reported 100→60), plus nothing wrapped at print time.
- `welcome-banner.last-column.test.ts` — pins the final-column reserve across widths and field lengths.
- `render.test.ts` — centering tests now assert the art block is invariant to session facts; width bound tightened from `<=` to `<`.
- Full suite: **741 files / 14,166 tests pass**. `tsc --noEmit` clean.

## Known unfixed

- The legacy boxed banner (`AFK_BANNER_PLAIN=1`) still lands on the final column at every width ≥28. Verified, deliberately out of scope: the fix routes through the shared `maxInnerBoxWidth()` helper that `statusPanel` and other boxes depend on.
- Found while investigating, untouched: resize → debounced `repaint()` (`terminal-compositor.lifecycle.ts:245-251`) → `evictRowsToScrollback` (`frame-preserve.ts:213-222`) can scroll banner rows on a shrink. It relocates rows intact so it is not the tearing, but it has had no independent review.

### DIFF
--- a/src/cli/render.test.ts
+++ b/src/cli/render.test.ts
@@ -212,8 +212,8 @@ describe('welcomeBanner', () => {
   });
 
   it('keeps the full tagline un-truncated in the mid-width band beside the sprite', () => {
-    // Regression guard for the 55–72-col "dead zone": the mascot still renders
-    // here (drops only below 55 cols), but the 42-col tagline used to live in
+    // Regression guard for the 56–72-col "dead zone": the mascot still renders
+    // here (drops only below 56 cols), but the 42-col tagline used to live in
     // the ~(cols−31) column beside the 27-col sprite — only ~33 cols at 64 — and
     // truncated to "Run coding agents without ba…". Hoisting it to a full-width
     // row fixed that; assert the whole thesis survives, un-ellipsized, WITH the
@@ -322,7 +322,7 @@ describe('welcomeBanner', () => {
     });
 
     it('drops the mascot and stacks info flush-left on a very narrow terminal', () => {
-      // Below the sprite budget (cols − 2 − 27 − 2 < 24, i.e. cols < 55) the
+      // Below the sprite budget (cols − 1 − 2 − 27 − 2 < 24, i.e. cols < 56) the
       // 27-col goblin would crush every info row into a one-char sliver. The
       // compact fallback drops the sprite and stacks the info full-width so the
       // identity signals stay legible instead of ellipsizing to nothing.
@@ -339,9 +339,29 @@ describe('welcomeBanner', () => {
       expect(hasSprite(out)).toBe(false);
       // …but every identity signal still survives, full-width.
       expect(out).toContain('Agent AFK');
-      expect(out).toContain('run coding agents without babysitting them');
       expect(out).toContain('opus_1m');
       expect(out).toContain('afk/polish-goblin-banner');
+      // Tagline present, though it ellipsizes by one char here — see the 45-col
+      // pin below for why 44 is one column short of holding it whole.
+      expect(out).toContain('run coding agents without babysitting');
+    });
+
+    it('holds the full tagline from 45 cols up in the compact fallback', () => {
+      // The 42-col tagline + 2-col LEFT_PAD needs 44 columns of glyphs, so it
+      // only used to fit at exactly cols=44 by occupying the terminal's FINAL
+      // column — which is the DECAWM deferred-wrap hazard that mangled the
+      // banner on resize (see welcome-banner.last-column.test.ts). Reserving
+      // that column moved the untruncated-tagline floor from 44 to 45. The
+      // tradeoff is deliberate: one ellipsized char at one narrow width, in
+      // exchange for sprite rows that survive a SIGWINCH intact.
+      Object.defineProperty(process.stdout, 'columns', { value: 45, configurable: true });
+      const out = strip(welcomeBanner({
+        mode: 'Interactive Mode',
+        model: 'opus_1m',
+        version: '5.11.0',
+        cwd: '/Users/example/projects/agent-afk',
+      }));
+      expect(out).toContain('run coding agents without babysitting them');
     });
 
     it('keeps every compact-banner row within the terminal width', () => {
@@ -355,23 +375,27 @@ describe('welcomeBanner', () => {
         hintLine: '/help · /model · /resume · Esc to interrupt · /exit to quit',
       }));
       const maxLine = Math.max(...out.split('\n').map((l) => stringWidth(l)));
-      expect(maxLine).toBeLessThanOrEqual(44);
+      // Strictly LESS than the width: a row landing exactly on column 44 puts a
+      // glyph in the final column, which sets the emulator's soft-wrap bit and
+      // re-splits the row on resize. This assertion used to be `<=`, which
+      // admitted precisely the corrupting case.
+      expect(maxLine).toBeLessThan(44);
     });
   });
 
-  describe('right-column vertical centering', () => {
+  describe('art-column vertical centering', () => {
     // The block-art hero is the ONLY source of full-block (█) glyphs — the
     // sprite is drawn with half-blocks (▀▄) only — so the first output line
-    // carrying a █ marks where the right column begins, i.e. its top pad.
+    // carrying a █ marks where the art column begins, i.e. its top pad.
     const heroTopRow = (s: string): number =>
       s.split('\n').findIndex((l) => /█/.test(l));
 
-    it('centers the right column onto the sprite, round-biased DOWN (not floor)', () => {
-      // A full column — model·mode + worktree + cwd + metaLine — is 12 rows
-      // against the 13-row sprite. (13−12)/2 = 0.5, so Math.round lands the top
-      // pad at 1, where Math.floor would strand the column at the cap tip (row
-      // 0). Pinning the exact top row is the direct guard for the round-not-floor
-      // bias the layout comment promises (welcome-banner.ts renderHybridBanner).
+    it('centers the 5-row hero onto the 13-row sprite, round-biased DOWN', () => {
+      // The hero is the only content sharing rows with the sprite (session text
+      // moved to full-width rows below — see welcome-banner.ts "Invariant (no row
+      // mixes pixel art with text)"). (13−5)/2 = 4, landing the logo beside the
+      // goblin's ear/eye band rather than the narrow cap tip. ROUND, not floor,
+      // is what the layout comment promises; this pins it.
       Object.defineProperty(process.stdout, 'columns', { value: 100, configurable: true });
       const out = strip(welcomeBanner({
         mode: 'Interactive Mode',
@@ -382,15 +406,19 @@ describe('welcomeBanner', () => {
         metaLine: 'Resuming abc123',
       }));
       expect(/[▀▄]/.test(out)).toBe(true); // sprite present → mascot layout
-      expect(heroTopRow(out)).toBe(1); // padded DOWN by one row (round, not floor)
+      expect(heroTopRow(out)).toBe(4);
     });
 
-    it('pushes the hero lower when the info column is shorter', () => {
-      // Centering responds to column height: a minimal column (model·mode only)
-      // is shorter than a full one, so it earns a larger top pad and its hero
-      // sits strictly lower. This exercises the round-biased centering across
-      // two different column heights without over-pinning either exact value.
+    it('holds the hero position invariant across session facts', () => {
+      // Regression guard for the mangling fix: the art block's geometry must not
+      // depend on how much session text there is. Previously the facts rode in
+      // the same column as the hero, so a /resume metaLine or a long branch name
+      // changed the sprite/hero pairing — and, worse, widened the art rows until
+      // a resize tore them. A fully-populated banner and a bare one must now
+      // produce byte-identical art rows.
       Object.defineProperty(process.stdout, 'columns', { value: 100, configurable: true });
+      const artRows = (s: string): string[] =>
+        s.split('\n').filter((l) => /[▀▄█]/.test(l));
       const fuller = strip(welcomeBanner({
         mode: 'Interactive Mode',
         model: 'opus_1m',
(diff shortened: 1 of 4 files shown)


griffinwork40:
Wave 1.5 complete — all load-bearing citations resolved against the ref. Synthesis below is orchestrator-local (Wave 2 input was small enough that dispatching a third agent would have cost more than it returned; findings, dedup, and conflict resolution are done inline).

---

# Review — PR #842 · `fix(banner): stop pixel art sharing rows with text so resize cannot tear it`

**Target** `griffinwork40/agent-afk#842` · ref `88ffe5eb` · base `main` · 4 files, +369/−69 · regime **full**
**Stated intent** PR title + body — spec-compliance **assessed**
**Verification** citations checked inline against `88ffe5eb1f8b1f90847e9ec0b2c9eb880cb36cbe` (= local HEAD, clean tree)

## Verdict: **MERGE**

No blocking findings. The load-bearing arithmetic is correct where it matters: widest art row **45** (`LEFT_PAD 2 + MASCOT_WIDTH 27 + GUTTER 2 + hero 14`), hero exactly **5 rows** so `round((13−5)/2) = 4` matches the pinned `heroTopRow === 4`, and the mascot boundary really moves 55 → 56 (`spriteBudget = cols−32 ≥ 24`). Exported surface (`welcomeBanner`, `WelcomeBannerOpts`) is unchanged, and the +4-row height breaks nothing because the sole production caller **measures** the banner rather than assuming its height (`src/cli/commands/interactive.ts:711-722,805` — `preArmAnchorRow += countNewlines(chunk)`). Every tradeoff the PR discloses matches the code; no undisclosed behaviour change survived verification.

---

## medium

**M1 · confidence high · correctness/api-compat · `src/cli/render/welcome-banner.ts:341` (diff-context) + `:453-464` (file-state, verified) · ref:88ffe5eb**
The cwd row is the one composed row that never passes through a display-width clamp, so the function's newly-declared "never the terminal's final column" invariant does not hold for a cwd containing wide glyphs: `truncateMiddle` early-returns on `displayWidth`, then slices by **code units**, emitting up to ~2× `headerMaxW` display cells for a CJK/emoji path.
```ts
infoRows.push(palette.dim(truncateMiddle(tildifyHome(opts.cwd), headerMaxW)));
// displayWidth-aware slicing approximation: characters and display cells
// align for ASCII paths, which is the common case here.
return s.slice(0, headW) + ELLIPSIS + s.slice(s.length - tailW);
```
Scope, stated precisely: the bug class is **pre-existing** and this diff strictly *improves* it — pre-diff that same row rode beside the sprite, so a wide-glyph cwd tore the art; post-diff it degrades to a wrapped text row, which the PR designates as acceptable. What is new is the invariant comment that overclaims coverage of it.
*suggestion:* wrap it as `truncateDisplay(truncateMiddle(tildifyHome(opts.cwd), headerMaxW), headerMaxW)` so the display-width clamp is authoritative.

## low

**L1 · confidence high · test-coverage · `src/cli/render.test.ts:637` · ref:88ffe5eb · file-state (verified verbatim)**
The diff tightened the 44-col bound `<=` → `<` at `:382` but left its direct sibling at 48 cols untightened, so that test still admits the exact-fill row the PR declares corrupting.
```ts
const maxLine = Math.max(...out.split('\n').map((l) => stringWidth(l)));
expect(maxLine).toBeLessThanOrEqual(48);
```
Verified as exactly one straggler — the other `toBeLessThanOrEqual(40)` bounds at `:610` and `:671` are `statusPanel`, not the banner.
*suggestion:* `expect(maxLine).toBeLessThan(48);`

**L2 · confidence high · correctness · `src/cli/render/welcome-banner.ts:274-277,287,411-417` + `src/cli/wrap.ts:25-27` · ref:88ffe5eb · diff-context + file-state (verified)**
The header comment justifies bare `cols - 1` by naming the 1–3 column terminal — and the reserve is then defeated on exactly that range by three separate paths: `headerMaxW`'s `Math.max(1, …)` floor pushes the budget back above the physical width, `wrapToWidth` returns the hint line unwrapped when `usableCols <= 0`, and the compact hero text fallback (`palette.brand('Agent ') + …`, 9 cells + 2 pad) is pushed untruncated under a comment claiming "Rows are pre-truncated so nothing overflows". `getTerminalWidth()` has no lower clamp above 1 (`src/cli/terminal-size.ts:13-16`), so the range is reachable in principle, absurd in practice.
```ts
const headerMaxW = Math.max(1, usableCols - LEFT_PAD.length);
if (!Number.isFinite(width) || width <= 0) { return text; }   // wrap.ts:25-27
```
Sharpened by the sibling convention this comment copies: `render/card.ts:160` avoids precisely this floor.
*suggestion:* clamp after composition — `truncateDisplay(row, Math.max(0, usableCols))` — instead of flooring the budget, or narrow the comment's claim to cols ≥ 4.

**L3 · confidence high · test-coverage · `src/cli/render/welcome-banner.resize.test.ts:112,122` · ref:88ffe5eb · diff-context (verified)**
Two of the three shrink cases assert only `after == before` with no non-empty guard, so if art stopped rendering entirely both arrays would be `[]` and the tests would pass vacuously. The guard exists only in the first case (`:104`).
*suggestion:* add `expect(before.length).toBeGreaterThan(10)` to the 120→45 case and inside the 140→[56,64,72,80,96] sweep.

**L4 · confidence high · test-coverage · `src/cli/render/welcome-banner.last-column.test.ts:50-53` · ref:88ffe5eb · absence confirmed-absent**
The last-column sweep runs ASCII-only cwds, leaving M1 uncovered. Grep `cwd: '[^']*[^\x00-\x7F]'` across `src/cli/**` at the ref returns **zero** matches — while the repo already has this exact pattern for a sibling component (`render.test.ts:665-671`, "statusPanel stays within width for wide glyph content", `状态` / `東京🙂`).
*suggestion:* add a CJK segment to the `for (const cwd of [DEEP_CWD, SHALLOW_CWD])` loop.

## nit

**N1 · convention · `src/cli/render/welcome-banner.ts` is 464 lines at the ref**, over the repo's 350-LOC source ceiling (`CHANGELOG.md:123`, `docs/browser-control-scope.md:660`); this diff grew an already-over file by net +46. *suggestion:* extract `truncateMiddle` / `tildifyHome` / `formatVersion` into a sibling module. — *both Wave 1 agents cited 465; corrected to 464 by measurement, see manifest.*

**N2 · maintainability · `welcome-banner.resize.test.ts:51`** — `isArtRow = /[▀▄█]/` classifies by glyph class, so a future fixture whose branch or cwd contains one of those characters is compared as art. Harmless with the current ASCII fixtures. *suggestion:* match art positionally (print rows 0..12).

**N3 · hygiene · `welcome-banner.resize.test.ts:45`, `welcome-banner.last-column.test.ts:53-54`** — fixtures hardcode the operator's real home path and the private-repo directory name on a public repo. No secret-shaped literal; precedent already exists in-tree. *suggestion:* `/Users/example/...`.

**N4 · pre-existing, NOT introduced here · `terminal-compositor.first-turn-banner-echo.test.ts:75`, `first-turn-echo-image.test.ts:93`, `banner-commit-gap.test.ts:68`** — three suites label 11–12 synthetic rows as "live geometry" / "real AFK banner". Verified stale *before* this PR (the pre-diff banner was 17 rows), and verified harmless: none of the three imports `welcomeBanner`, they print synthetic `BANNER_LINE_i` rows.

**N5 · pre-existing, NOT introduced here · `welcome-banner.ts:334-346`** — no control-character/ANSI strip on `opts.worktree` / `opts.cwd` / `opts.metaLine` before they reach the TTY. Identical helpers pre-diff.

---

| claim | status | note |
|---|---|---|
| `welcome-banner.ts:465` / evidence `465 }` (both agents) | **fabricated** | file is 464 lines at ref, ends `;\n}\n`. Claim retained on independent measurement; agents' evidence snippet discarded. |
| "+4-row change makes compositor comments false" (correctness agent) | **downgraded + reattributed** | comments claim 11–12 vs a 17-row pre-diff banner → already false. low → nit. |
| `render.test.ts:637 toBeLessThanOrEqual(48)` | verified verbatim | |
| `wrap.ts:25-27` early return | verified verbatim | |
| `truncateMiddle` code-unit slice `:453-464` | verified verbatim | |
| non-ASCII cwd absent from `src/cli/**` tests | **confirmed-absent** | zero matches |
| compositor banner tests import `welcomeBanner` | **confirmed-absent** | zero matches → synthetic rows |
| single production caller = `interactive.ts` | verified | `git grep -l` non-test → 1 file |
| height measured, not assumed (`preArmAnchorRow`) | verified | `interactive.ts:711-722,805` |
| `card.ts` / `echo.ts` reserve precedent | verified verbatim | `card.ts:160`, `echo.ts:96` |
| `@xterm/headless` declared | verified | `package.json:144`, devDependencies |
| 45 / 5-row hero / 56-col boundary / top-pad 4 | verified | arithmetic re-derived from code |

**Shadow-verify not triggered** — no `critical` or `high` finding survived verification; `medium` and below go straight through per the skill's gate.

## What was not checked

Citations verified inline against branch HEAD `88ffe5eb1f8b1f90847e9ec0b2c9eb880cb36cbe`. Stated intent: PR #842 title + body — spec-compliance assessed. Not checked: **no tests were run** (read-only review), so the PR's "741 files / 14,166 tests pass" and `tsc --noEmit` clean claims are unverified; no runtime execution of the banner or `@xterm/headless`; the legacy `AFK_BANNER_PLAIN=1` boxed path (PR scopes it out); the `repaint()` → `evictRowsToScrollback` interaction the PR flags as "found while investigating, untouched"; DEC STD-070 / xterm.js `isWrapped` semantics against xterm.js source — accepted as the PR's own measured claim, which its print-time assertion backs.

**Merge decision: MERGE** — M1 and L1–L4 are worth a follow-up commit, none of them gates this one.

---

**Done**
- Reviewed PR #842 end-to-end: 2 parallel `research-agent` dimension passes (security·api-compat; correctness·spec-compliance·test-coverage·perf), then inline Wave 1.5 citation + absence-claim verification against ref `88ffe5eb`, then local synthesis — 1 medium, 4 low, 5 nit, verdict MERGE.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/89533f68-529a-4bd7-b7ec-aa039a267bfc/pr-842.diff`; every citation in the manifest re-derived via `git show`/`git grep` at `88ffe5eb1f8b1f90847e9ec0b2c9eb880cb36cbe`; session trace at `~/.afk/state/witness/<session>/trace.jsonl`.
- World unchanged: read-only review, no files written, no commits, no PR comments posted. The report above lives only in this transcript — if you want it durable, say so and I'll write it to a file.
- Two Wave 1 claims were corrected rather than passed through: the 465-line count (actual 464, evidence snippet fabricated) and the compositor-comment staleness (pre-existing, not caused by this PR).
- Deferred: M1 and L1–L4 are follow-up work, not merge gates; per the skill's read-only contract I did not fix them.

---

PR #843, opened 2026-08-01
feat(subagent): propagate path-backed image attachments

griffinwork40:
## Summary

- Adds an optional `attachments` field to the `agent` tool so a parent session can attach **path-backed images** to a subagent dispatch, closing the one remaining string-only hop in an otherwise-already-image-capable pipeline.
- Ingress (Telegram photo, REPL clipboard paste) and both provider loops already produce/consume `ContentBlockParam[]`, and `AgentSession.sendMessageStream` already accepts them — the `agent` tool's dispatch chain was the only leg still forced through a bare `prompt: string`. This PR closes that leg without widening `prompt` itself.
- Design choice: a **sibling `attachments` field**, not a `prompt: string | ContentBlockParam[]` widening. Keeping `prompt` a plain string leaves `promptHead`, `id_prefix` derivation, the background-job label, and `promptHash` correct by construction. A naive content-block-aware label deriver would risk writing base64 into `meta.json` on disk, which would violate the documented invariant at `src/agent/bg-job-log.ts:41` ("Full prompt text is never written to disk").
- Reference-not-bytes is forced by the schema boundary: the **parent model** is the one emitting the tool call, and a bytes-in-schema design would require it to emit base64 as output tokens — roughly 70K–700K tokens for a typical photo. Paths (or future ids) are the only viable shape at this boundary.

## Behavior

- **Zero change on the no-attachment path.** A bare string prompt is passed exactly as it is today; this is purely additive.
- **Caps:** 8 images per dispatch, 5 MiB combined, both named constants (`MAX_SUBAGENT_ATTACHMENT_COUNT`, `MAX_SUBAGENT_ATTACHMENT_BYTES` in `src/agent/tools/subagent/attachment-resolve.ts`).
- **Extension → media-type validation** against the Anthropic-accepted image set (`.jpg`/`.jpeg`, `.png`, `.gif`, `.webp`); anything else is rejected before a read is attempted.
- **Containment:** path resolution goes through the existing `resolveAndContain` helper (`src/agent/tools/handlers/_cwd-utils.ts`), which carries the unconditional credential/secret read-denylist floor — a subagent dispatch cannot use this field to read a denylisted credential/secret path any more than any other read path can.
- **`mode:"background"` + `attachments` is explicitly rejected** (`src/agent/tools/subagent/input-parse.ts:247`) with a clear error, so image bytes can never reach a background job's label, hash, or on-disk metadata.
- **Non-vision child model:** dispatch still proceeds (the openai-compatible path degrades to a text notice in place of the image), but the parent now receives an explicit `WARNING: child model … is not vision-capable; attached images were dropped.` (`src/agent/tools/subagent-executor.ts:722`) — previously this failure mode was silent.

## Follow-up commit (85eaa2a8)

The byte cap was originally enforced after `readFile`, so an oversized path would be fully loaded into memory before being rejected. `stat()` now runs first and rejects on the projected total before any read occurs; the post-read check remains as a consistency backstop for a file that grows between the two calls. The added test uses a **sparse file** larger than Node's max buffer size specifically so that observing the cap's own error (rather than `ERR_FS_FILE_TOO_LARGE`) proves no read was attempted.

## Test results

- `pnpm lint` (`tsc --noEmit`, strict): exit 0.
- `pnpm test src/agent/tools/ src/agent/subagent/ src/agent/content/`: **78 files, 2303 passed, 2 skipped** — independently re-run and confirmed as part of this PR's preparation, not just carried over from local notes.

## Verification

No adversarial verify was pre-declared via `--verify`, but the cumulative diff (541 insertions / 48 deletions across 17 files) exceeds the skill's 100-line auto-trigger threshold, so an independent read-only verifier wave was dispatched to re-derive this PR's load-bearing claims directly from source (not from this body's own reasoning). Result: **6/6 CONFIRM, 0 CONTRADICT**:

1. Zero change on the no-attachment path — CONFIRM (`subagent-executor.ts` mutates `childPrompt` only inside an `attachments !== undefined` branch; a test asserts the literal prompt string is passed through unchanged when attachments are omitted).
2. `mode:"background"` + `attachments` throws (not warns) — CONFIRM (`input-parse.ts:247-251`, exercised by a dedicated test).
3. `resolveAndContain`'s credential/secret denylist floor is unconditional and is what `attachment-resolve.ts` actually calls for every path in the loop — CONFIRM.
4. `stat()` genuinely gates before `readFile()` via real control flow (a throw that exits before the read line is reached), not just source-order — CONFIRM, corroborated by the sparse-file test.
5. Both named caps (8 images / 5 MiB) are enforced pre-dispatch on the live path from the `agent` tool, not just declared — CONFIRM.
6. Spot-check claim (verifier's own choice): non-vision child model dispatch still proceeds and only appends a warning rather than blocking — CONFIRM, with a dedicated test covering both the warning text and that a result is still returned.

The verifier did not itself run the test suite (no shell access) — the 2303-passed count above is this PR's own re-run, not independently reproduced by the verifier.

## Out of scope

This PR covers **path-backed images only**. Specifically not covered here:

- Images arriving via Telegram or REPL clipboard never touch disk and are still anonymous to the model at the subagent-dispatch boundary — wiring those in is PR 2 (inbound attachment registry + ingress `[image img_xxxxxx · …]` id markers).
- Compose/DAG nodes and fork-mode skills each build their own string child prompt today and need their own, separate widening to accept attachments.
- Three pre-existing byte-hygiene gaps, unrelated to this change and not touched by it: microcompaction cannot see image bytes (`compact.ts:143-156`), `persistCompactionSidecar` writes pre-compaction messages — including any image bytes — unredacted to a sidecar file (`trace/writer.ts:380-399`), and `applyOutputCap` deliberately never caps `result.image.data`.

**Known tension, disclosed rather than hidden:** four touched files were already over this repo's 350-LOC-per-file guidance before this change: `schemas.ts` (1315 lines), `handle.ts` (847), `subagent-executor.ts` (727), `input-parse.ts` (491). This PR adds small, additive hunks to each (+7 / +9 / +41 / +29 lines respectively) and extracts the new concern into its own file (`attachment-resolve.ts`, 71 lines) rather than splitting the pre-existing files further. Those splits remain owed and are not attempted in this PR.

Full 3-PR plan and design rationale: `docs/subagent-attachment-propagation.md` (added in this PR's first commit — note its own header still reads "Status: proposal (nothing implemented)" as of the scoping date; that line is a historical scoping-time note and is left as-is rather than edited to track this PR).


griffinwork40 (commented):

## Merge decision: **MERGE**

No `critical` and no `high` findings survive verification. The PR's load-bearing claims were re-derived from source at the reviewed ref and hold. Ten findings below are `medium` and under — follow-ups, not blockers.

**Verified as claimed** (each re-derived, not taken from the PR body):
- **Zero change on the no-attachment path** — `childPrompt` is initialized to `parsed.prompt` and reassigned only inside `if (parsed.attachments !== undefined)` (`subagent-executor.ts:675-697`). `input-parse.ts:169` (`if (paths.length > 0) attachments = paths;`) normalizes `[]` to `undefined`, so the empty array also takes the bare-string branch.
- **stat() gates before readFile via real control flow** — the `throw` at `attachment-resolve.ts:63` sits between the `stat` (`:62`) and the `readFile` (`:65`) on the same statement path. The sparse-file test is not a tautology: 3 GiB exceeds Node's `kIoMaxLength`, so a read would surface `ERR_FS_FILE_TOO_LARGE` instead of the cap message.
- **Extension validation precedes the read** — `MEDIA_TYPES.get()` throws at `:50` before `resolveAndContain` and `readFile` are reached.
- **Count cap precedes the loop** — `:34-38` before `:47`.
- **`mode:"background"` + `attachments` throws** — `input-parse.ts:247-251`; the Ctrl+B promotion path separately passes `prompt: backgroundPrompt` (the plain string) so `label`/`promptHash` never see bytes.
- **The old `src/cli/slash/_lib/image-blocks.ts` move is complete** — grep for `_lib/image-blocks` at the ref returns **zero** matches; all three importers are updated.

---

## Findings

### medium · correctness/resource-cleanup · `src/agent/tools/subagent-executor.ts:687-693`
The attachment-resolution `catch` returns early after awaiting only `handle.teardown()`, skipping `childManager.teardownAll()` and `teardownIsolatedWorktree()` — both owned exclusively by the `finally` in `runForegroundWithPromotion` — so an `isolation:"worktree"` dispatch with a bad attachment path leaves its worktree on disk.

```ts
      } catch (err) {
        await handle.teardown().catch(() => undefined);
        return {
          content: `Agent tool attachment resolution failed: ${...}`,
```

The worktree is already created at `:559-564` and `isolationTeardown` is only handed to the promotion wrapper at `:718`, past this return. `teardownIsolatedWorktree` has exactly one production call site (`foreground-promotion.ts:406`) and `childManager?.teardownAll()` one (`:383`), both inside the bypassed `finally`. No mutual-exclusion check exists between `isolation` and `attachments` (the `isolation` block at `input-parse.ts:412-445` guards only `cwd`, `writeRoots`, and `mode`), so the combination is reachable.

*Downgraded from `high` by an independent verifier:* `worktree-sweep.ts` reclaims the tree on its own disk-scan schedule (boot prune + daemon tick), so this is a bounded/delayed leak, not permanent orphaning. The tree is clean and reapable, but reclaim waits on `MIN_EMPTY_AGE_MS` (1h) while the parent lives.

**Suggestion:** in the `catch`, call `await childManager?.teardownAll()` and `await teardownIsolatedWorktree(isolationTeardown)` before returning, mirroring the `finally`.

### medium · correctness · `src/agent/tools/subagent-executor.ts:720-724`
The warning asserts images "were dropped" based only on `supportsVision(childConfig.model)`, with no provider check — but the drop happens exclusively on the openai-compatible branch, so a non-matching model id on anthropic-direct delivers the images *and* tells the parent they were discarded.

```ts
    if (parsed.attachments !== undefined && !supportsVision(childConfig.model)) {
      result.content =
        `WARNING: child model ${childConfig.model} is not vision-capable; attached images were dropped.\n\n` +
```

`supportsVision` returns `false` for any unrecognized id (`model-capabilities.ts:113-126`). The actual drop is `messages.ts:134` (`if (!opts.vision) return flattenUserContent(...)`); grep for `vision` across `src/agent/providers/anthropic-direct/**` returns zero matches — that provider passes content verbatim.

**Suggestion:** gate the warning on `providerForModel(childConfig.model) === 'openai-compatible'` (already imported in `child-config.ts:28`).

### medium · correctness · `src/agent/tools/subagent-executor.ts:720-724`
The warning is prepended to `result.content` unconditionally, but two paths make that content a JSON payload — `foreground-promotion.ts:220-230` (Ctrl+B promotion) and `:349-353` (structured failure) — so a promoted or failed non-vision attachment dispatch emits a WARNING-prefixed string that is no longer parseable JSON.

**Blast radius is bounded:** grep found no production consumer that `JSON.parse`s an agent tool result (all hits are unrelated — keychain, memory renderers, config loaders), so the affected reader is the parent model, not code.

**Suggestion:** gate the prepend on `result.isError !== true` and a non-promoted outcome, or carry the warning in a separate structured field.

### medium · perf · `src/agent/tools/subagent/attachment-resolve.ts:9` + `src/agent/content/image-blocks.ts:28`
The 5 MiB cap is enforced on raw file bytes while the payload actually transmitted is base64, so a cap-compliant dispatch emits ~6.67 MiB encoded — with the raw `Buffer` and the encoded string both live — and the eight reads run sequentially rather than in parallel.

```ts
export const MAX_SUBAGENT_ATTACHMENT_BYTES = 5 * 1024 * 1024;
...
        data: att.bytes.toString('base64'),
```

**Suggestion:** document the constant as a raw-bytes ceiling and note the ~1.34× on-wire inflation next to it.

### low · correctness · `src/agent/tools/subagent/attachment-resolve.ts:62-65`
`stat()` validates size but not file type, so a FIFO named `x.png` inside a read root reports size 0, passes the cap, and makes `readFile` block or stream unbounded into the heap.

**Suggestion:** reject when the `stat()` result fails `.isFile()`.

### low · correctness · `src/agent/tools/subagent/attachment-resolve.ts:47`
Duplicate paths are neither deduplicated nor rejected, so `['/a.png','/a.png']` is read twice, charged twice against the byte cap, and delivered to the child as two identical image blocks.

**Suggestion:** dedupe on the resolved path before the read loop.

### low · security · `src/agent/tools/subagent/attachment-resolve.ts:55-65`
`resolveAndContain` returns the un-symlink-resolved path, so `stat()` and `readFile()` each re-traverse the link and a mid-flight retarget reads a different file's bytes than the one that passed the size gate.

This is a pre-existing repo-wide pattern shared with `read-file.ts` (grep for `realpath|lstat|isFile\(\)` there returns only `resolveAndContain`), and `_cwd-utils.ts:38-45` states the non-adversarial threat model explicitly — so this is consistency-with-existing-behavior, not a regression introduced here.

**Suggestion:** open once with `fs.promises.open()`, then `fstat` + `read` on the single descriptor.

### low · security · `src/agent/tools/subagent/attachment-resolve.ts:48-53`
Media type is derived from the filename extension with no magic-byte check, so a non-image file renamed `*.png` inside a read root is base64-encoded and shipped to the child as `image/png`. The code already marks this: `// Future magic-byte unification point: sniffMimeType in telegram/handlers/message.ts.`

**Suggestion:** sniff the leading magic bytes after read and reject on mismatch with the extension-derived type.

### low · test-coverage · `src/agent/tools/subagent/attachment-resolve.test.ts:36-51`
Both caps are tested only on the rejecting side (`MAX_..._COUNT + 1`, `MAX_..._BYTES + 1`), so flipping `>` to `>=` in either guard would pass the suite while wrongly rejecting exactly 8 images and exactly 5 MiB.

**Suggestion:** add accepting-side cases at exactly `MAX_SUBAGENT_ATTACHMENT_COUNT` and exactly `MAX_SUBAGENT_ATTACHMENT_BYTES`.

### low · test-coverage · `src/agent/tools/subagent/input-parse.test.ts:79-110`
The new `describe('attachments')` block has five cases (absolute-path accept, non-array, non-string, relative, background) but none passes `[]` — the single input that pins the "zero change on the no-attachment path" claim for an explicitly-empty array. The behavior is correct today (`input-parse.ts:169`); it is just unpinned.

**Suggestion:** `expect(parseAgentInput({ prompt: 'p', attachments: [] }).attachments).toBeUndefined()`.

---

## Dropped during verification

One Wave 1 finding was **dropped as false-premise**: a claim that attachments are contained against the parent's `parentReadRoots` rather than the child's `childInheritedReadRoots`, letting a read-restricted child receive out-of-scope bytes. Re-reading `subagent-read-scope.ts:150-152` at the ref refutes it — the module's documented invariant is *"never narrower than the parent (child read scope ⊇ parent read scope)"*, so containing against the parent's roots is never wider than the child's own reach.

## What was not checked

- Citations verified inline against PR head **85eaa2a8** (`git show <ref>:<path>`), not against the working tree — every `blocking`/`high` citation plus every `file-state` citation at any severity was re-read at the ref.
- **Stated intent:** PR #843 title + body — spec-compliance **was** assessed. Verdict: substantially compliant, no scope creep. The one gap is that the body's "purely additive" framing does not cover the error-path cleanup gap in finding 1.
- Tests were **not executed**: the claimed `pnpm lint` exit 0 and "78 files, 2303 passed, 2 skipped" are unreproduced here. CI on this head is green across Lint & Build, ubuntu/macos, real-PTY scrollback, docs, and publish-bundle smoke.
- Not reviewed: `docs/subagent-attachment-propagation.md` prose (docs-only), the Telegram/REPL ingress surfaces (explicitly out of scope per the PR body), and the three pre-existing byte-hygiene gaps the body discloses (`compact.ts`, `persistCompactionSidecar`, `applyOutputCap`) — untouched by this diff.
- The four touched files already over the 350-LOC guidance (`schemas.ts`, `handle.ts`, `subagent-executor.ts`, `input-parse.ts`) are disclosed in the PR body; the new concern was correctly extracted to its own 71-line file rather than deepening them.

PR #844, opened 2026-08-01
feat(observability): opt-in capture of parent→subagent dispatch prompts

griffinwork40:
## Problem

Nothing persists what a parent actually asked a child to do.

The witness trace carries `promptHead` — the first 80 chars of a dispatch prompt (`trace/types.ts:225-240`) — and only on the `agent` tool path (`tools/subagent-executor.ts:621`). Four of the six dispatch paths record no prompt content at all:

| Dispatch path | `promptHead` today |
|---|---|
| `agent` foreground / background / `isolation:"worktree"` | yes (80 chars) |
| `compose` / DAG nodes | **no** |
| `skill` forks | **no** |
| mint's 7 phases, audit-fit | **no** |

The session ledger isn't an alternative: `summarizeToolInput` has no `prompt` branch, so it resolves to `''` for the `agent` tool, and it caps at 400 chars. Transcripts record parent turns only. `AFK.md` documents this as a known gap.

## Change

`AFK_CAPTURE_SUBAGENT_PROMPTS=1` (**off by default**) writes each inbound child message as its own markdown file — YAML frontmatter for attribution, verbatim body — under `state/witness/<sessionLabel>/prompts/`.

Capture happens in the **child's own** `sendMessageStreamInternal`, one line beside the `ledger.recordUser` call that forks are deliberately gated out of (`LedgerLifecycle.ensure` returns early when `depth`/`parentSessionId` is set). That seam buys three things a parent-side dispatch hook would not:

- **All six paths, no per-path plumbing** — every path delivers its prompt through `AgentSession.sendMessageStream*`.
- **One directory per session** — a fork resumes its parent's sessionId, so `getPromptsDir(this.sessionId)` inside a child resolves to the *parent's* directory. Reading "every prompt this session dispatched" is one `readdir`.
- **Multi-turn coverage** — captured per turn, not just first dispatch.

The directory is the index. No trace event references these files, so the trace schema is untouched and capture works whether or not a `TraceWriter` is wired.

## Why off by default

Two things make default-on indefensible, both measured rather than assumed:

- **Nothing prunes the witness tree.** The only retention mechanism in the repo (`log-retention.ts::capJsonlBySize`) is wired solely to `session-grants.jsonl`. The tree is already **12,596 directories / 461 MB** on the author's machine.
- **Redaction is best-effort.** `redactInlineSecrets` catches named-shape secrets only; it structurally cannot catch `postgres://user:pass@host`, PEM blocks, or PII. The written file says so in a banner rather than implying safety.

Other safety details: 8 KB cap with an explicit truncation marker; truncate **before** redacting so an oversized paste never pays full regex cost; `0o600` at creation, not a later `chmod` (no TOCTOU); fire-and-forget that can never reject, because a forensics artifact must not be able to fail a turn.

## Tradeoff accepted

Fire-and-forget means a hard process exit can lose the final prompt. The trace writer solves this with a synchronous exit backstop; no equivalent is built here — an exit hook is a materially larger surface than the failure warrants for an opt-in artifact. The wiring test polls rather than sleeps, so it asserts "eventually lands" instead of pinning a timing assumption.

## Verification

- `pnpm test` — **741 files / 14,181 tests pass**, 2 skipped
- `pnpm lint`, `audit:env:check`, `scan:env:check`, `audit:sdk:check`, `audit:chalk:check`, `build` — all pass
- 29 new tests: 23 unit (gate, byte-accurate truncation incl. multi-byte cut, frontmatter contract incl. `promptBytes`/`bodyBytes`, redaction reaching disk, secret straddling the truncation cut, `0o600` file + `0o700` dir, same-millisecond nonce disambiguation, disabled-writes-nothing) + 6 wiring (fires for a fork on a real turn; **never** for a top-level session; nothing when the flag is off; per-turn on multi-turn; witness label over provider session id; index advances after a provider error)

The wiring tests are the load-bearing ones — they caught that the feature was hooked to nothing on first run, and they pin the invariant that a top-level session's own prompts are the operator's and must never be captured.

## Provenance

Design and the adversarial review that reshaped it are in `docs/proposals/subagent-prompt-capture.md`. The first design captured at the parent's `handle.run()` and added a `subagent_prompt` trace event kind; a 3-lens critique rejected the new event kind unanimously and moved the seam, cutting scope from ~420 LOC / 10 files to ~150 / 7. §8.1 records which critic claims were confirmed and which were refuted on verification.

## Review follow-ups

Code review found five defects in `subagent-prompt-capture.ts`; all are fixed in this branch.

- **Secret surviving redaction after truncation** — the 8 KB byte-cut could split a secret into a prefix shorter than the redactor's `([^\s]{16,})` minimum, leaving partial key material on disk. The trailing partial token is now dropped before redaction.
- **World-listable prompts dir** — `mkdir` omitted `mode`, creating `prompts/` at umask (0o755) so any local user could list filenames encoding subagentId, timestamp, and turn. Now created `0o700`.
- **Symlink-following write** — `flag: 'w'` would follow a pre-planted symlink at the target path and silently drop `mode` on an existing inode. Now `flag: 'wx'`.
- **Silent filename collision** — `safeSlug(...).slice(0, 60)` could truncate away the discriminator, so two same-millisecond sibling forks at the same turn overwrote one another. A random nonce now disambiguates.
- **Frontmatter that never described the file** — `promptBytes` reported the pre-redaction, pre-truncation size while the body on disk was post-both. It now reports the original inbound size alongside a new `bodyBytes` for what was actually written. Module and field docs are also corrected: the body is verbatim for string prompts, but content-block prompts arrive already summarized by the caller (deliberate — it keeps base64 image data off disk).

Two additional wiring tests that asserted "no files" with no settle window were passing instantly regardless of behavior; they now poll a bounded window and fail if any artifact appears.

Two out-of-scope findings were filed as separate issues rather than fixed here: witness-tree retention (captured prompts accumulate with no pruning policy), and observability failures being invisible without `AFK_DEBUG=1` (capture swallows its own errors by design, so a persistent write failure is silent).


griffinwork40:
@codex fix comments

PR #846, opened 2026-08-01
feat(subagent): make inbound images nameable and forwardable to subagents

griffinwork40:
> **Stacked on #843 — review and merge #843 first.** The base branch of this PR is `afk/subagent-image-propagation`, not `main`, so the diff shown here is only PR 2's changes.

This is **PR 2** of the 3-PR plan in `docs/subagent-attachment-propagation.md` (committed in #843, section 3 "Layer 2").

## The problem

#843 can only forward images that already have a filesystem path. An image a human sends via **Telegram** or pastes with **Ctrl+V in the REPL** never touches disk — and, more fundamentally, **arrived anonymous**: no ingress path emitted an id, filename, or size next to the image block. The model could not *name* the image it had been given, so it could not forward it. Nameability, not plumbing, was the blocker.

## What this adds

- **Content-hash ids.** Ingress mints `img_<hex>` from the sha256 of the bytes, so the same image always yields the same id (free dedup, no double-write).
- **Session-scoped sidecar.** New `getInboundAttachmentsDir(sessionId)` in `src/paths.ts`, mirroring the existing `getTraceDir` / `screenshotsDir` convention: `~/.afk/state/witness/<session>/inbound-attachments/`.
- **A model-visible marker.** Each inbound image now gets an adjacent text block: `[image img_a1b2c3 · image/png · 842 KB]`. Both surfaces call **one shared formatter** (`formatInboundImageMarker`) via one shared entry point (`registerInboundImageBlocks`), so the two ingress paths cannot drift.
- **Registry.** `src/agent/content/attachment-registry.ts` maps `id → {path, mediaType, sizeBytes}`, keyed by session. It stores **paths, never bytes**, so no base64 can reach logs, traces, labels, or metadata. The read side is exposed on `SubagentExecutorContext` the same way `traceWriter` / `agentRegistry` already are, and wired in `wire-executors.ts`.
- **Ids accepted by the `agent` tool.** `attachments` now takes an inbound id *or* an absolute path; both kinds may appear in one dispatch. An **unknown id lists the session's available ids** so the model can self-correct instead of guessing.
- **Telegram duplication removed.** `message.ts` no longer builds its image block inline; it uses the shared `appendImageBlocks` encoder from #843.

## Collision policy

A short-prefix collision with *different* bytes is **never aliased**. On collision the prefix lengthens by two hex chars (up to the full digest), and only a full-digest collision falls back to a numeric suffix (`img_<hex>_1`). Identical bytes at the same id return the existing record. The parse regex accepts exactly that shape: `^img_[0-9a-f]{6,64}(?:_[1-9][0-9]*)?$`.

## Security note

Registry-sourced paths deliberately skip `resolveAndContain`, while human-supplied absolute paths still go through it. This is safe because the registry only ever returns paths **it constructed itself** under the session sidecar dir, with the filename derived from a validated id plus a fixed extension — no user-controlled component reaches the path. #843's caps (8 images, 5 MiB total) and its stat-before-read guard both still apply, across the combined path+id set. `mode:"background"` + attachments remains rejected, so image data still cannot reach background labels or hashes.

## Verification

- `pnpm lint` (`tsc --noEmit`, maximal strictness): **exit 0**
- `pnpm test src/agent/tools/ src/agent/subagent/ src/agent/content/ src/telegram/ src/cli/slash/`: **161 files, 3662 passed, 2 skipped**
- New: `src/agent/content/attachment-registry.ts` (119 LOC) + tests; both new files well under the 350-LOC ceiling.

## Known gaps — disclosed, not hidden

1. **Slash-command / skill attachments are not registered.** `skill-message-bridge.ts:65` still calls `appendImageBlocks` directly, so an image pasted while invoking a slash command gets no id and no registration. Fixing it cascades an async + `sessionId` change through every slash dispatch site, so it is deliberately left for a follow-up rather than widened here.
2. **`buildUserPayload` silently falls back** to unregistered blocks when `sessionId` is undefined. Today the only production caller always passes it (test-only path), but it is a hidden fallback and would degrade silently if a future caller forgot.
3. **No cleanup for the sidecar bytes.** Files land under the per-session witness dir so they inherit whatever session retention exists; no pruning was invented here. The in-memory registry map is also process-lifetime, which matters most for a long-lived Telegram bot or daemon. Growth is bounded by paths + metadata in memory, but disk is unbounded.
4. Still out of scope, as in the plan: **compose/DAG and fork-mode skills** (each builds its own string child prompt and needs separate widening), and **PR 3's byte hygiene** — microcompaction cannot see image bytes (`compact.ts:143-156`), `persistCompactionSidecar` writes them unredacted (`trace/writer.ts:380-399`), and `applyOutputCap` does not cap `result.image.data`.

## Provenance

Implemented by a dispatched subagent that **timed out after 45 minutes before committing**. The work was left uncommitted in its worktree; it was then reviewed, gate-verified, and committed by the orchestrating session. Reviewers should weigh items 1–2 above with that in mind — they are the seams where the run stopped short.


griffinwork40:
Verified. `ImageBlockAttachment.mediaType` is a closed union of 4 types — `extensionFor`'s switch is exhaustive under TS strict mode, so the "unsupported type returns undefined" test-gap claim is refuted. The `skill-message-bridge.ts` gap is confirmed: it uses `appendImageBlocks` directly without registering to the registry.

Now let me synthesize the final review.

---

## PR #846 Review — Inbound Image Registry & Subagent Forwarding

**PR**: `feat(subagent): make inbound images nameable and forwardable to subagents`
**Branch**: `afk/subagent-image-registry → main` · `+362/-61` · 17 files · 4 commits

### What this PR does

Introduces an `InboundAttachmentRegistry` that content-hashes inbound images (from CLI or Telegram), writes them to a session-scoped sidecar directory, and emits a model-visible `[image img_xxxxxx · type · size]` text marker before each image block. Subagent dispatches can then reference these images by `img_` ID (alongside existing absolute-path attachments), letting a parent forward images it received to child subagents without re-encoding.

---

### Findings by Severity

#### 🔴 High

**H1 — Unbounded memory growth: module-level `sessions` Map with no eviction**
`src/agent/content/attachment-registry.ts:73`
`const sessions = new Map<string, Map<string, InboundAttachmentRecord>>()` is module-level state on the singleton `inboundAttachmentRegistry`. No `delete()`, `clear()`, or eviction method exists. In a long-running Telegram bot or daemon, every image from every session accumulates in memory forever. Other registries in the codebase (`background-registry.ts`, `shell-jobs/registry.ts`, `presence.ts`) all implement TTL eviction or pruning — this one is the exception.
**Recommendation**: Add a session-lifecycle cleanup hook or TTL-based eviction.

**H2 — Unbounded disk growth: sidecar image files never pruned**
`src/agent/content/attachment-registry.ts:137` / `src/paths.ts:329`
Images are written to `~/.afk/state/witness/<sessionId>/inbound-attachments/` via `writeFile()` and never deleted. The codebase explicitly acknowledges "nothing prunes the witness tree" (`env.ts:1152`). Unlike the opt-in prompt capture (`AFK_CAPTURE_SUBAGENT_PROMPTS=1`), this is **on by default** for every Telegram photo and CLI image attachment. At ~500KB average per Telegram photo, 100 photos/day ≈ 18GB/year of never-cleaned disk.
**Recommendation**: Add a pruning strategy or at minimum document retention behavior.

#### 🟡 Medium

**M1 — `buildSkillInvocationMessage` not updated — feature gap**
`src/cli/slash/_lib/skill-message-bridge.ts:65`
This file uses `appendImageBlocks` directly (no `registerInboundImageBlocks`), confirmed by its own comment ("intentionally does NOT delegate to buildUserPayload"). Images attached to slash-command skill invocations (e.g., `/review <image>`) are NOT registered, so the model cannot forward them to subagents by `img_` ID. The CLI `turn-handler.ts` and Telegram `message.ts` paths both register correctly — this is the one ingress path that doesn't.
**Recommendation**: Wire `registerInboundImageBlocks` into `buildSkillInvocationMessage`, or document this as a known limitation.

**M2 — Zero observability for registration events**
`src/agent/content/attachment-registry.ts`
No logging, tracing, or debug output when an image is registered. `listIds()` exists but is only surfaced in error messages for unknown IDs. An operator debugging "which image got which ID" or "why did the subagent see the wrong image" has no diagnostic path. The witness trace records tool calls but not image registration.
**Recommendation**: Add a trace event or `debugLog` call at `put()` return.

**M3 — Tool schema description may shift model behavior**
`src/agent/tools/schemas.ts:362`
The `attachments` description now mentions "inbound image ids" first, which may encourage the model to try ID-based references even when path-based references would be more appropriate. Hallucinated IDs fail safely at `attachment-resolve.ts` with a clear error listing available IDs — but this could cause unnecessary retry loops.
**Recommendation**: Low-priority monitor of model behavior in practice.

#### 🟢 Low

**L1 — Concurrent `put()` race for different bytes sharing a 6-hex prefix**
`attachment-registry.ts:129-141`
Between the `sessionEntries.get(id) === undefined` check and `sessionEntries.set(id, record)`, there are two `await` points (`mkdir`, `writeFile`). Two concurrent puts for different bytes sharing the same 6-char SHA-256 prefix would both write to the same file path — the second overwrites the first, corrupting the first caller's record. Probability is ~1 in 16.7M per pair. Only relevant if the registry gains concurrent same-session callers (currently each ingress path processes one image at a time).

**L2 — `sessionId === undefined` fallback in `buildUserPayload` is defensive dead code**
`user-payload.ts:62-68`
Both callers (`turn-handler.ts`, `message.ts`) throw before reaching `buildUserPayload` when `sessionId` is undefined with attachments present. The `else` branch (raw `appendImageBlocks` without marker) is only reachable from test helpers.

**L3 — Data retention of potentially sensitive Telegram photos**
This is a subset of H2 but worth calling out separately: Telegram photos can be personal/sensitive. They now persist permanently on disk in an unpruned directory. This is a privacy consideration, not just a storage one.

#### ⚪ Info (No Issue)

- **Read-root bypass is safe** — All path components (sessionId via regex validation, img ID via SHA-256 content hash, extension via closed enum) are system-constructed. No user input influences the path. The bypass of `resolveAndContain()` is justified (`paths.ts:316-322`, `attachment-registry.ts:134-136`, `attachment-registry.ts:155-162`).
- **Cross-session isolation enforced** — `Map<sessionId, Map<id, record>>` with strict two-level key lookup. No API to inject an arbitrary sessionId (`subagent-executor.ts:803-804`).
- **ID spoofing safe** — Nonexistent IDs produce a safe error listing the session's own IDs. No filesystem probing. In-memory Map.get() only.
- **Hash collision loop terminates** — SHA-256 produces 64 hex chars > initial 6. Suffix fallback guarantees termination even for full-digest collisions.
- **All `buildUserPayload` callers updated to `await`** — Only one production caller (`turn-handler.ts:354`) and all test calls updated. TypeScript would catch missed awaits at compile time.
- **`buildUserPayload` sync→async is a breaking signature change** — But the function is internal (`src/cli/slash/_lib/`), all callers are in-tree, and TypeScript catches missing awaits. No external consumers.
- **`extensionFor` exhaustive** — The `InboundMediaType` is a closed union of 4 types; the switch covers all cases. The "unsupported type returns undefined" test-gap claim is **refuted** by the type system.

### Test Coverage Gaps

| Gap | Severity | Location |
|-----|----------|----------|
| Suffix increment collision path (`suffix > 0`) untested | **High** | `attachment-registry.test.ts` — should test full-digest collision with different bytes |
| Cross-session isolation untested (session-A can't see session-B) | **High** | `attachment-registry.test.ts` — should test `get('other-session', id)` returns undefined |
| Input-parse regex negative cases (non-hex, too-short, suffix 0, leading zero) | **Medium** | `input-parse.test.ts` — should test `img_XXXX`, `img_a1b2`, `img_a1b2c3_0`, `img_a1b2c3_01` |
| `listIds()` sorting behavior untested | **Medium** | `attachment-registry.test.ts` |
| `get()` with unknown session/ID untested | **Medium** | `attachment-registry.test.ts` |
| Unknown ID error message format in `attachment-resolve` untested (only count-cap tested) | **Medium** | `attachment-resolve.test.ts` |
| Dedup through full `buildUserPayload` path (same bytes → same ID) | **Medium** | `user-payload.test.ts` |
| Telegram test uses exact hash `img_374ffe` — brittle | **Low** | `message-photo.test.ts` — recommend `expect.stringMatching` |

---

### Merge Recommendation

**Request changes** — The PR is well-designed and security-sound, but has two unbounded resource leaks (memory + disk) with no eviction strategy for a feature that's on by default. These should be addressed before merge, or at minimum an eviction plan should be documented as a follow-up issue. The feature gap in `buildSkillInvocationMessage` (M1) should also be resolved or explicitly documented as out-of-scope.

Recommended pre-merge actions:
1. Add session-lifecycle cleanup or TTL eviction to `InboundAttachmentRegistry` (H1, H2)
2. Test the suffix-increment collision path and cross-session isolation (test gaps)
3. Decide on `buildSkillInvocationMessage` — wire it or document the limitation (M1)

---

griffinwork40:
## Review of PR #846 — subagent inbound image registry

**PR**: feat(subagent): make inbound images nameable and forwardable to subagents
**Branch**: `afk/subagent-image-registry` → main · **Reviewed ref**: `421989628109213cb15b775ba155896860d7dfcd` · **Stats**: +571/-115, 22 files · full regime

### Verdict: **MERGE**

No `critical` or `high` findings. One medium, two lows, three nits — all merge-compatible improvements. Shadow-verify was not triggered (no critical/high).

### Security posture — clean, contracts verified
All five stated security contracts hold. I independently verified the load-bearing one in Wave 1.5: `getTraceDir` → `validateSessionId` enforces `/^[a-zA-Z0-9_-]+$/` (`paths.ts:314,324-326`), and `getInboundAttachmentsDir` delegates to it (`:331`). So a registry path = validated sessionId + hex-validated id (gated by the input-parse regex before it reaches the registry) + fixed extension → fully process-controlled → the `resolveAndContain` bypass is safe. Sync→async on `buildUserPayload`/`buildSkillInvocationMessage`: the only production callers (`turn-handler.ts:358`, `run-skill-dispatch-turn.ts:165`) both `await`; no un-awaited callers → no breaking change.

### Findings (sorted)

| # | Sev | Dim | Location | Issue → Fix |
|---|-----|-----|----------|-------------|
| F1 | **medium** | perf | `attachment-registry.ts:109` | Dedup reads the whole file (≤5 MiB) from disk on every cache hit instead of comparing the already-computed digest. → Store the full sha256 in `InboundAttachmentRecord`, compare digests: `existing.digest === digest`. |
| F2 | low | correctness | `input-parse.ts:164` | Regex `{6,64}` accepts odd-length hex the registry never generates (it only emits even prefixes). → Tighten to even-length alternation. |
| F3 | low | correctness | `attachment-resolve.ts:50-88` | `paths` not deduplicated — a model sending `['img_a1b2c3','img_a1b2c3']` reads twice and burns 2 of the 8-image cap for one unique image. → `new Set()` pre-pass. |
| F4 | nit | correctness | `attachment-registry.ts:39` | `Math.max(1, …)` rounds 0-byte images up to "1 KB". → Use bytes for sub-KB sizes. |
| F5 | nit | spec-compliance | `schemas.ts:329-330` | Drops `get_runtime_state` cap-query guidance + "by default" — unrelated to PR intent (scope creep). → Revert or split. |
| F6 | nit | security | `attachment-registry.ts:121-128` | `extensionFor` has no exhaustive `default`; widening `InboundMediaType` would yield `img_…undefined` under the bypass. → `default: { const exhaustive: never = mediaType; throw … }`. |

F6 is future-proofing, not a current vuln — today the media-type is gated upstream by the `MEDIA_TYPES` map and the 4-member union stays exhaustive.

### What was not checked
- Citations verified inline against branch HEAD `421989628109213cb15b775ba155896860d7dfcd`.
- Stated intent: PR #846 title+body — spec-compliance assessed.
- Tests were not re-run; the runtime dispatch path was not exercised; only the diff + post-merge file reads were reviewed. Author-disclosed known gaps (slash/skill attachments unregistered, `buildUserPayload` silent fallback, no sidecar cleanup, compose/DAG + fork-mode out of scope) were taken at face value and not re-investigated.

**Recommendation:** Merge as-is; address F1–F3 in a follow-up (F1 is the most worthwhile — it removes a per-resubmit disk read of up to 5 MiB).

---

PR #848, opened 2026-08-02
fix(tui): render a deliberate tool refusal as blocked, not failed (#75)

### LINKED ISSUE #75: tui: long skill/subagent runs are opaque — add live progress (per-tool elapsed + run-level status)
**The problem**

During a long skill run — e.g. `/mint`, which fans out many subagents and can run
30+ minutes — the REPL gives the user almost no sense of *what is happening right
now*, whether progress is being made, or whether the process is stuck. The tree of
subagent activity is structurally rich but temporally flat: it shows what *has*
happened (tool counts, line counts, a `Done (… · 5m 2s)` line on completion) far
better than what is happening *this moment*.

Concretely, several things compound into "I have no idea what my Mac is doing for
half an hour":

1. **In-progress tool rows have no elapsed counter.** A running tool renders a
   static verb (`… running`) with no clock, so a 5-minute `web_scrape` looks
   identical to one stuck at 5 seconds.
   (`src/cli/commands/interactive/tool-lane-render-children.ts:293`)

2. **The only live clock is the spinner's, and it's narrow.** `formatElapsed()`
   has a 5s grace before showing anything and measures from a per-waiting-window
   `startedAt`, not total run time — so it understates how long the overall run
   has been going.
   (`src/cli/terminal-compositor.types.ts:65–75`)

3. **The "still working" signal can disappear during the exact long waits it's
   meant to cover.** The spinner is intentionally paused when prose streams and
   only re-enables "on the next `tool_use_detail`." If the model streams a little
   reasoning and then blocks on a multi-minute tool call, there can be no waiting
   indicator at all — the screen looks frozen.
   (`src/cli/_lib/stream-renderer-orchestrator.ts:270–278`)

4. **There is no run-level / skill-level summary.** Nothing surfaces "skill mint ·
   running 18m · 4 subagents done, 2 active." The user only gets per-subagent
   totals, and only on each `Done` line — i.e. after the fact.

5. **Expected errors read as alarming and cryptic.** Lines like
   `✗ Tool "bash" is not in the configured allowlist` render as small red rows
   that can scroll off the capped overlay. Without context they look like
   failures rather than the agent harmlessly probing a gated tool, adding to the
   "what is going wrong?" anxiety during an otherwise-healthy long run.

(Observed running `/mint` — see the tree of `Agent(mint-research) [subagent] — 41
tools`, `Done (82 tools · 4.1k tok · 5m 2s)`, and the `✗ Tool "bash" is not in the
configured allowlist` rows. Screenshot to follow.)

**Proposed solution**

Direction, not a fixed design — make a long run *legible at a glance*:

- **Per-tool elapsed** on in-progress rows: `… running 4m22s`. Attach a
  `startedAt` when a tool entry is opened and render it in the in-progress branch
  (`tool-lane-render-children.ts:293`, entry creation in `tool-lane.ts`).
- **A run-level status line** for active skills/turns: total elapsed, subagent
  counts (active / done), and ideally a one-line "current activity." Candidate
  homes: the loop-stage bar (`loop-stage.ts`) or the progress banner
  (`progress-banner.ts`), which today only fires on SDK-emitted `ProgressEvent`s.
- **A robust liveness signal** that doesn't vanish during long tool calls —
  e.g. a watchdog that re-asserts the waiting indicator if no `tool_result`
  arrives within N seconds, independent of the stream-pause logic in
  `stream-renderer-orchestrator.ts:270–278`.
- **(Secondary)** A clearer affordance for expected/benign tool errors
  (allowlist rejections) so they don't read as failures mid-run.

**Alternatives considered**

- Rely on the existing spinner — insufficient, per points 2–3 above.
- Dump verbose per-event logging — too noisy; this is about *legible* progress,
  not more text.

**Surface**

REPL (primary). The run-level summary likely applies to Telegram long-run
notifications too, but the acute pain is the interactive TUI.

**Where this lives in the code**

- `src/cli/commands/interactive/tool-lane-render-children.ts` (in-progress row render)
- `src/cli/commands/interactive/tool-lane.ts` (tool-entry lifecycle / overlay)
- `src/cli/commands/interactive/tool-lane-render-agent.ts` (subagent tree row)
- `src/cli/commands/interactive/loop-stage.ts` (footer stage rail)
- `src/cli/commands/interactive/progress-banner.ts` (SDK ProgressEvent banner)
- `src/cli/_lib/stream-renderer-orchestrator.ts` (spinner pause/resume logic)
- `src/cli/terminal-compositor.types.ts` (`formatElapsed`, `ELAPSED_GRACE_MS`)

griffinwork40:
Closes #75.

## Why this is the last item

#75 opened with four asks. Two shipped earlier (the run-level status line via `LoopStageBar` + `progress-banner.ts`; the liveness watchdog via `src/agent/subagent/idle-watchdog.ts`). The per-tool elapsed clock was **closed by decision** in #840 — `OverlayComposer.dirty` is one global boolean and composition is event-driven behind a 1500ms throttle, so an unconditional row clock would sit visibly stalled between events: *"a clock that lies about time, which is a worse failure than no clock."* Silence is instead covered by `no output for 41s` and `· waiting Xs`.

That leaves the benign-error affordance, which is this PR. Full reconciliation of all four items is in [this triage comment](https://github.com/griffinwork40/agent-afk/issues/75#issuecomment-5154237294).

## The problem

During a long skill run the agent routinely probes a tool it turns out not to hold:

```
│ ├ ✗ Tool "bash" is not in the configured allowlist
```

That row is byte-for-byte as alarming as a crash. On a 20-minute `/mint` run, a screenful of red for the permission system working exactly as designed reads as "something is going wrong" — which is the anxiety #75 was opened about.

The dispatcher has known better all along. `failureClass` is stamped at 8+ sites in `tools/dispatcher.ts` (`permission-denied`, `hook-block`, `policy-refusal`, …) and reaches `receipt.ts`, `trace.ts`, and `tool-call-trace.ts` — but it was dropped before any renderer saw it. The tool-lane only ever had a boolean `isError`, so it could not distinguish a gate saying no from a tool that broke.

## Three outcomes instead of two

| glyph | meaning |
|---|---|
| `✓` | success |
| `⊘` | the system deliberately said no |
| `✗` | something actually broke |

`⊘` is not new vocabulary — it already means blocked/cancelled/killed in `verdict-ledger.ts:38`, `bgsub.ts:67`, and `sh.ts:54`.

The glyph and the message body are toned together. `formatOutcome()` picks its tone from the same benign/fault split, so a refusal row never renders as `⊘ <red text>` — the neutral glyph beside a still-red message would have left the "screenful of red" #75 is about essentially intact.

Grouped rows tally the kinds apart, so a collapsed burst stays readable:

```
$ bash ×50 — 48 ok, 1 error, 1 blocked · last: fatal: not a git repository
$ bash ×50 — 48 ok, 2 blocked          ← no red anywhere
```

When both are present the `· last:` tail prefers the **real fault**, so a broken command can't be buried by a benign rejection that happened to land after it.

## The benign set is not invented here

`tool-failure-density.ts` already maintained exactly this partition, to keep refusals out of failure-rate stats (a `browser_open` that refused half its navigations on domain policy is not 50% broken). That set moves to `agent/trace/types.ts` beside the `ToolFailureClass` per-class notes, and both consumers now share it — two copies drift the moment a class is added.

Membership stays deliberately narrow, following the rationale already written into that type:
- `timeout` stays alarming — a high timeout rate is a real problem.
- `denial-breaker` stays alarming — a fork torn down for spinning is review-worthy.
- an **unclassified** failure stays alarming — absence of a class is also what a genuine handler bug looks like, and guessing "benign" would hide breakage.

A test pins the full partition, so adding a `ToolFailureClass` later forces a conscious decision instead of silently defaulting to red.

## Plumbing, and why each hop is tested

`failureClass` now rides `tool.output` → `ToolResultChunk` → renderer. That is five hops, and a miss at any one leaves the feature **silently dead with every renderer test still green** — the renderer would be correct, just never handed the field.

So each hop is pinned by a test that drives a *real* rejection rather than a synthetic event:
- **openai-compatible** — the existing `allowedTools` permission-gate test now asserts `failureClass: 'permission-denied'` on the emitted event.
- **anthropic-direct** — the existing AFK-mode `PreToolUse` block test now asserts `failureClass: 'hook-block'`.
- **stream-consumer** — passthrough pinned on both the normal *and* the persisted-output branch (the latter builds its chunk separately and is the easier one to forget).

Both providers land together; a field plumbed in one and forgotten in the other is the exact failure mode documented at `dispatch-append.ts:214`.

## Byte-compatibility

`doneGlyph`'s new parameter is optional, so omitting it reproduces the original binary ✓/✗ byte-for-byte. All five call sites were updated to pass the class through, but behaviour at each is unchanged for a success or an unclassified failure, and all 12 tool-lane snapshot assertions still hold. `formatOutcome()` reads `failureClass` off the chunk it already receives, so its signature is untouched. The new glyph and the new body tone are reachable *only* when an errored result carries a benign class. The healthy grouped row keeps its single `palette.dim()` span — the invariant guarded since #840. No `.snap` file changed.

## One harness fix worth flagging to the reviewer

Tone cannot be asserted by matching raw ANSI in these suites: chalk auto-disables color with no TTY, so a `\^[[31m` probe passes **vacuously** on a row that was never toned at all. My first draft of the "unclassified stays red" test failed for exactly this reason and caught it. New tone assertions swap the palette member for a tagged sentinel instead (the mechanism `applyTheme()` uses, mirroring `tool-lane-format.test.ts`). Note this means the pre-existing `not.toMatch(/\^[\[31m/)` guard at `tool-lane-grouped-errors.test.ts:118` is weaker than it looks — left alone here as out of scope.

## Verification

- `pnpm test` — **748 files, 14,290 passed**, 2 skipped, 0 failed
- `audit:sdk:check`, `audit:env:check`, `scan:env:check`, `audit:chalk:check`, `pnpm build` — all pass
- No snapshot changed; no file newly crosses the 350-LOC ceiling (the three touched files already above it — `trace/types.ts`, `provider.ts`, `stream-consumer.ts` — were over before this change and gain only JSDoc)
- +22 tests: glyph/partition units, grouped-tally cases, per-hop plumbing guards, and the literal #75 single-row scenario end-to-end

## Not done here

The two follow-ups #840 deferred remain open and separable: the `round 7/50` denominator needs `max_tool_use_iterations` plumbed across the agent→renderer boundary, and `activity-tracker.ts` + `streaming-stall-badge.ts` (~400 LOC) are still orphaned with no production call site.

**Abort origin is not distinguished.** The synthesized `Tool call aborted` results are stamped `failureClass: 'abort'`, which is in the benign set — but `signal.aborted` is also how a **session timeout** and a **budget exhaustion** arrive, because both funnel through `onAbort() → providerQuery.interrupt() → requestAbort('interrupted')` and `AbortReason` is only `'interrupted' | 'closed'` (`abort-coordinator.ts:56`). So a timeout-killed call currently renders neutral and drops out of failure-density stats, which sits awkwardly against the "`timeout` stays alarming" rule above. Fixing it means widening `AbortReason` to carry the origin and threading it to the dispatch sites — ~5 files across the session/provider abort machinery, orthogonal to this PR's rendering scope. The four stamp sites carry comments recording the gap so it is not silently inherited.

Tracked as #856. Worth noting the gap is **not new here**: the same `signal.aborted → 'abort'` stamp already exists at six sites on `main` (`dispatcher.ts:809,923,1087,1125`, `browser-act.ts:174`, `browser-open.ts:104`). This PR extends the pattern to the provider-loop synthesized results and gives it its first user-visible consequence, which is what surfaced it.


griffinwork40:
@codex fix comments

griffinwork40:
Wave 1.5 verification complete; both high findings independently re-derived and confirmed. Synthesizing inline (Wave 2).

---

# Code Review — PR #848

**`fix(tui): render a deliberate tool refusal as blocked, not failed (#75)`**
`afk/issue-75-benign-refusal-glyph` → `main` · ref `5f83643d` · +474/−24 · 21 files (13 prod, 8 test) · regime: **full**

The core idea is right and the plumbing is careful: `failureClass` is genuinely threaded end-to-end, both providers land together, the benign set is de-duplicated into one owner, and the grouped-row tally split is correct. Two high findings block: the change does not deliver its headline outcome on the exact path #75 was filed about, and a new stamp contradicts the PR's own narrowness rationale.

---

## Blocking

### H1 · high · confidence high · correctness / spec-compliance — *defeats stated purpose*
`src/cli/commands/interactive/tool-lane-format.ts:112` · ref `5f83643d` · citation-type: file-state (re-read at ref)

On the single-row path — the literal #75 scenario the author says "the operator actually screenshotted" — only the **glyph** is neutralized. The refusal's message text beside it is still painted red, because `formatOutcome()` picks its tone from the boolean `chunk.isError` alone and was never threaded `failureClass`. The row goes from `✗ <red text>` to `⊘ <red text>`.

```ts
const resultColor = chunk.isError ? palette.error : palette.dim;
```

For a single-line refusal, `display` is skipped (`:118` requires `!isError`), no `persistedPath`, no `lineCount>1` → execution reaches `return resultColor(sanitizeLabel(preview))`. All five renderer call sites concatenate `doneGlyph(...) + ' ' + formatOutcome(...)` on the same line, so every ungrouped row is affected. The diff never touches `formatOutcome` and never passes `failureClass` into it. The "screenful of red" the issue complains about is one character narrower per row.

**Suggestion:** thread `failureClass` into `formatOutcome` and select `palette.warning`/`palette.dim` when `isBenignFailure(failureClass)` — mirroring the tone split already implemented correctly in `formatGroupedSibling`.

*Adversarially re-derived — CONFIRMED. Refutations attempted and failed: `clampLineToTerminal` is ANSI-preserving (truncates width, does not rewrite SGR); `palette.error` is a real `chalk.red`, and any colour-disabling env suppresses `palette.warning` equally so it cannot selectively neutralize; no non-`formatOutcome` path exists for a lone rejected call.*

### H2 · high · confidence high · correctness / scope creep — *breaks stated contract*
`anthropic-direct/loop/tool-dispatch.ts:119-123`, `openai-compatible/query/dispatch-append.ts:127-131,139,153-157` · citation-type: diff-context

The new unconditional `failureClass: 'abort'` stamp (4 production sites) marks **every** `signal.aborted` tool result benign, but the signal cannot say *why* it aborted. A session timeout and a budget exhaustion reach the same place as a user Ctrl+C, and the `TimeoutError` carrying the distinction is discarded — genericized to the literal string `'interrupted'` at `anthropic-direct/query.ts:629` and `openai-compatible/query.ts:935` before the provider's coordinator sees it. The stamp reads only `input.signal.aborted`, never `.reason`; `AbortOrigin` exists but is consumed solely by the subagent-cascade tracing layer.

Consequence: a tool call killed by a **timeout** now renders the neutral `⊘` — directly contradicting this PR's own rationale, *"`timeout` stays alarming — a high timeout rate is a real problem."*

Second-order, and outside the stated rendering scope: these synthesized results previously carried **no** class, so they counted as `unclassified` (`tool-failure-density.ts:242-243`) and were included in failure-rate stats. They are now silently excluded (`:171-172`). The set membership did not change — the *population* carrying a benign class did.

**Suggestion:** propagate the abort origin into the synthesized result and stamp only genuine user cancellations, or omit `failureClass` on synthesized aborts so they stay `unclassified` and keep counting.

*Adversarially re-derived — CONFIRMED on all five links. The verifier corrected my mechanism: the session controller and the provider's per-turn coordinator are distinct objects bridged by `interrupt()`, not one shared instance. This strengthens the finding — the reason channel is destroyed in transit.*

---

## Non-blocking

**M1 · medium · test-coverage** — `tool-lane.overlay.test.ts:1897-1925`. The new #75 headline test is tone-blind: `laneWithOneRejectedChild` calls `stripAnsi(lane.getOverlay())` and asserts only glyph presence/absence, so it certifies the fix while being structurally incapable of detecting H1's residual red. This is the same vacuous-assertion class the author diagnosed and solved elsewhere in this very PR — the `tonedBashRow` palette-sentinel helper — but did not apply to the headline case. *Suggestion: assert the single-row case with `tonedBashRow`.*

**L1 · low · test-coverage** — Two more negative ANSI assertions share the vacuous mechanism, both verified present at the ref: `src/cli/quota-footer.test.ts:142` (`expect(r.text).not.toMatch(/\u001b\[/)`) and `src/agent/subscription-usage.test.ts:312`. Pre-existing and out of scope, logged for a follow-up. The author's own note about `tool-lane-grouped-errors.test.ts:118` is accurate — that probe now sits at `:179` after this diff.

**L2 · low · convention** — `src/agent/trace/types.ts` is **918 lines** at the ref (889 on main) and this PR adds a new exported concern to it rather than a new file, against the standing 350-LOC rule. The PR body's "the three touched files already above it" undercounts: five touched production files exceed 350 — `tool-lane.ts` 935, `trace/types.ts` 918, `provider.ts` 583, `stream-consumer.ts` 521, `tool-lane-render-children.ts` 492. No mechanical LOC gate exists in this repo (verified absent), so this is convention, not CI.

**N1 · nit · spec-compliance** — PR body: *"all five existing call sites … are unchanged."* The diff modifies all five (exactly five non-test `doneGlyph` call sites exist; all updated). Behaviour is unchanged; the sentence as written is not.

**N2 · nit · test-coverage** — Hops (a) `tool-results.ts` and (b) `dispatch-append.ts:227` are pinned only transitively, via full-stack tests that drive real rejections through the real pipeline. Defensible — arguably stronger than a unit test — but not the per-hop isolation the PR body describes.

---

## Clean reads

- **security** — no secret exposure, no injection surface. `failureClass` is a closed 8-member string-literal union assigned only at internal call sites; grep for `failureClass:\s*['"]` returns only quoted literals, never a template literal or variable. A model cannot relabel a genuine fault as benign via crafted tool input. *(read `trace/types.ts`, `dispatcher.ts`, browser/ask-question handlers)*
- **api-compat** — 5/5 `doneGlyph` call sites updated, no renderer left with the old signature; `BENIGN_FAILURE_CLASSES` is member-identical to the deleted `EXCLUDED_FAILURE_CLASSES` (same five, same order — no silent stats drift); `trace/types.ts` has **zero** import statements, so the new CLI→agent *value* import creates no cycle and drags in no runtime weight; `exactOptionalPropertyTypes` is not set, so the new optional fields break no existing constructor.
- **correctness (grouped tallies)** — `blocked ∪ faults = errors`, disjoint and exhaustive, so `source[source.length-1]!` can never index empty; `ok = done − errors.length` remains correct with blocked counted as not-ok; pluralization correct in both renderers; the #840 pre-styled-vs-plain contract is preserved for healthy rows (a blocked-only row correctly takes the pre-styled branch).
- **perf-observability** — two extra `.filter()` passes per grouped row, behind a 1500ms composition throttle. No finding.

## What was not checked

- Citations verified inline against branch HEAD **`5f83643da3f015ed8dd6fecd06874e45d7295e2b`**. Two Wave-1 citations were **mis-pathed** (`agent-session.ts:589-592` → `src/agent/session/agent-session.ts`; `timeout.ts:119-121` → `src/agent/timeout.ts`); content verified at the corrected paths, so retained rather than dropped. One mechanism claim ("same AbortController") was corrected during verification.
- **Stated intent:** PR #848 title + body — spec-compliance assessed.
- Verification claims taken on faith, not re-run: `pnpm test` (748 files / 14,290 passed), `pnpm lint`, the audit gates. Executing them requires checking out the PR branch, which this review is not permitted to do.
- The "12 unchanged tool-lane snapshot assertions" / "no `.snap` changed" claim was not diffed.
- Wave-1's estimate of ~270 ANSI-assertion lines tree-wide is grep-derived and not deduplicated; only the two specific citations above were verified.
- Wave 1's `correctness` agent **timed out at 45 minutes** and returned nothing. Its scope was recovered by one narrow test-coverage agent plus direct inline verification at the ref — the dimension is covered, but by a different route than planned.

---

# ⛔ DO NOT MERGE

Two high findings. **H1** is the decisive one: the PR says "Closes #75," but on the ungrouped row — the case the issue was filed about — a benign refusal still renders its message in red. Threading `failureClass` into `formatOutcome` is a small change that would close the gap, and re-asserting the headline test with the `tonedBashRow` helper this PR already introduces would have caught it. **H2** needs the abort stamp gated on a distinguishable origin before a timeout starts reading as "the system said no" in both the TUI and the failure-density detector.

**Done**
- Reviewed PR #848 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability; emitted 2 high, 1 medium, 2 low, 2 nit findings and a DO NOT MERGE recommendation.
- Evidence is durable and re-inspectable: diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/0cf29aa0-8376-4eb5-a41e-7c2d6fc51e37/pr-848.diff`; every finding cites `file:line` resolvable via `git show 5f83643da3f015ed8dd6fecd06874e45d7295e2b:<path>`, which is present in the local object store.
- Nothing in the world changed: no files written, no commits, no pushes, no PR comments — the review skill is read-only by contract, so this report is deliberately not persisted as a file.
- Deferred: Wave 1's correctness agent timed out at 45 min and its scope was recovered inline; L1's two vacuous ANSI assertions (`quota-footer.test.ts:142`, `subscription-usage.test.ts:312`) are pre-existing and left for a separate follow-up.

---

PR #874, opened 2026-08-03
fix(tui): latch the dead-zone banner flush and close #867 review gaps

griffinwork40:
Follow-up to #867, which was merged (squash `399e8faa`) before its review feedback was addressed. Every item below was re-verified as still-live against `main` before being fixed — no stale findings were re-fixed.

## The substantive fix

`checkProgressBannerStaleness` marked the `progress-banner` slot dirty and flushed on **every 80ms pause tick** for as long as any child stayed silent — ~12.5 full overlay recomposes/sec, indefinitely, with no state tracking:

```ts
if (silentMs >= CHILD_QUIET_MS) { markDirty('progress-banner'); flush(); return true; }
```

The doc comment justified this by claiming repeat flushes were "free no-ops" absorbed by `setOverlay`'s identical-string dedup, so "no cadence gate is needed." That is wrong on two counts:

1. `flush()` recomposes **all five** overlay slots before any dedup runs — `markDirty` sets one shared dirty flag, so there is no per-slot short-circuit. "Free" could only ever hold at the terminal-write layer, not the compose layer.
2. The composed banner carries the child's elapsed `durationMs` rounded to whole seconds, so the final string is **not** byte-stable across a second boundary — a real `repaint()` fired roughly once per second regardless.

Replaced with a **per-source latch** (`SourceState.quietBannerAnnounced`), mirroring how the sibling `checkPauseAnnotations` change-detects with `pauseAnnotation`. The banner now flushes exactly once per transition into silence and re-arms when the child speaks again, so a later silence still re-announces. The loop sweeps every source before flushing rather than early-returning, or siblings' latches would go stale. This also removes the ~1 repaint/sec that was leaking into capture-mode (`script(1)`, asciinema) streams.

## Structural change

Extracted the concern into **`stream-renderer-dead-zone.ts`**. The fix would have pushed `stream-renderer-lifecycle.ts` past the 350-line ceiling (it was sitting at 343), and the dead-zone check is a whole concern with a clean seam — it takes `LifecycleContext` as a type-only import, so there is no runtime cycle.

| File | Before | After |
|------|--------|-------|
| `stream-renderer-lifecycle.ts` | 343 | **301** |
| `stream-renderer-lifecycle.test.ts` | 370 (already over) | **254** |
| `stream-renderer-dead-zone.ts` | — | 82 |
| `stream-renderer-dead-zone.test.ts` | — | 181 |
| `stream-renderer-dead-zone-integration.test.ts` | — | 107 |

Both modules are in `TIMER_FREE_MODULES`, so the no-autonomous-timer guard still covers this path.

## Per-item status

| # | Item | Source | Status |
|---|------|--------|--------|
| 1 | Latch the flush — no per-tick churn | [codex P2](https://github.com/griffinwork40/agent-afk/pull/867#discussion_r3700216799) + review LOW | fixed |
| 2 | Correct the overstated dedup doc claims (both files) | review MEDIUM | fixed |
| 3 | `"the tool-lane mqtt threshold"` → `"pause threshold"` (autocorrect artifact) | review NIT | fixed |
| 4 | Pin the exact `>=` boundary at `CHILD_QUIET_MS` | review MEDIUM-test | fixed |
| 5 | Latch holds across consecutive ticks + re-announce after child speaks | review HIGH-test | fixed |
| 6 | Combined-tick interaction past 30s | review MEDIUM-test | fixed in `7ef87541` — see Self-review below |
| 7 | Debug-log symmetry with `checkPauseAnnotations` | review LOW | fixed |
| 8 | Integration test — dead zone closed end-to-end | review HIGH-test | fixed |
| 9 | Extract concern to its own module (350-line ceiling) | structural | fixed |

**Already resolved on main, deliberately untouched:** the reviewers' P1/HIGH finding that the timer guard false-positives on the type annotation `ReturnType<typeof setInterval>`. Commit `9907491f` landed the call-form regex `/\bsetInterval\s*\(/` at `live-progress-no-timer.test.ts:53` eleven minutes after the review was posted. Verified directly rather than assumed.

**Resolved as a consequence of #1:** the capture-mode write leak (review LOW).

## Test evidence

The integration test (#8) is the one that matters — the two halves of #867 had only ever been unit-tested in isolation, so nothing actually proved an operator sees the clause. It drives a **real** `OverlayComposer` through `registerOverlaySlots`, and asserts the rendered banner contains `no output (waiting)` while the tool-lane's `· waiting Xs` annotation has **not** armed (that arms only past 30s). The child is given a `syntheticAgentToolUseId` deliberately — without it `checkPauseAnnotations` skips the source outright and the negative assertion would be vacuously true.

It was falsified before being trusted: setting `lastEventAt = Date.now()` (child no longer silent) makes it fail with `expected false to be true`; reverting restores the pass. It is load-bearing, not decorative.

```
Targeted:  6 files / 129 tests passed
Full:      766 files / 14,577 passed, 2 skipped, 0 failures
Lint:      tsc --noEmit clean
```

One caveat worth stating: an earlier full run showed 14 failures across 8 files (`schedules.test.ts` daemon-sync and similar port-bound suites). Those are unrelated to TUI code and did not reproduce on a clean re-run — a `pnpm install` in the fresh worktree had restarted the real AFK daemon, which contends for the same ports. The green run above is the reproducible one.

Closes the review feedback on #867.

## Self-review caught a false completion

Worth stating plainly, because it is the kind of thing that normally ships silently. A review pass over this PR's own diff found that item 6 was marked `fixed` in the table above while **no such test existed**. It had been dropped when the test work was re-scoped after a subagent timeout, and then claimed as done anyway. The existing integration test asserts the *opposite* property — that the 30s tool-lane annotation stays dormant at ~10s — which is not the same thing.

`7ef87541` adds the real case: one child silent past 30s so both the dead-zone banner latch and `checkPauseAnnotations` fire on the same 80ms tick, asserting the annotation *does* arm, the banner clause survives the tool-lane flush, and the latch is still held afterward. Verified load-bearing (40s → 10s fails with `expected false to be true`; reverting restores the pass).

Two further review findings, both verified and both accepted as no-change:

- The sole production call site discards `checkProgressBannerStaleness`'s return value (`stream-renderer.ts:973`), so the semantics change from "true every quiet tick" to "true only on transition" is unobservable in production.
- A `done`/`errored` source keeps a stale `quietBannerAnnounced` latch, since those are `continue`d before the latch is touched. Safe under a verified invariant: `git grep` at the reviewed ref finds zero assignments of `.done = false` / `.errored = false` and zero `sources.delete` calls, so source keys are never recycled. Noted here because that invariant is implicit rather than asserted near the code.

## Round 2 — codex P2 was a real bug (`6d3e9d13`)

Codex flagged that the latch is re-armed only by the tick observing an intermediate fresh state. That needs some 80ms tick to land inside the 8s window after a child resumes — so a suspended process, a closed laptop lid, or an event loop blocked past 8s skips every such tick and strands the latch at `true`. The next genuine quiet transition is then swallowed by the "already announced" guard and **the dead zone this PR exists to close reopens for that child permanently.** Lid-close mid-run is not an exotic scenario.

Fixed structurally, not locally: `lastEventAt` has exactly one production assignment site, so the latch is now re-armed there, synchronously, next to the timestamp it depends on. Clearing no longer depends on tick timing. The tick-side clear stays as a backstop so the checker remains self-consistent in isolation.

That also converges with an independent finding from the review wave — `SourceState.quietBannerAnnounced`'s doc implied synchronous re-arm while the behavior was sweep-lagged up to 80ms. Fixing the code was the right direction there rather than weakening the doc to match.

Regression test drives the real event path through `StreamRenderer.process` and runs **no tick** between the activity and the assertion. Verified load-bearing: commenting out the new clear fails with `expected true to be false`.

### Doc-accuracy fixes in the same commit

A second false completion, same class as item 6. Item 2 above claimed "both files fixed":

- `child-activity-select.ts` `clauseFor()` still asserted the **retracted** claim that a live counter would break `setOverlay`'s identical-string dedup — 55 lines below the JSDoc this PR rewrote to say that dedup *cannot be relied on*. One file, two contradictory statements about one mechanism. The conclusion (keep the clause static) is unchanged and still right, so the mechanism is corrected rather than the guidance dropped: the real cost is **rate** — ~80ms churn instead of ~1Hz.
- `stream-renderer.ts` still attributed `checkProgressBannerStaleness` to the lifecycle module after this PR moved it out of it.
- Three comments claimed the fix closes "the 1.5s–30s dead zone". The clause is gated on `CHILD_QUIET_MS`, so it covers **8s–30s**; ~1.5s–8s is uncovered by design. Overstating coverage in a comment is precisely the failure mode this PR was opened to correct, so the problem window and the covered span are now stated separately.

Local gate after round 2: **130 tests across 6 files**, `tsc --noEmit` clean.

### DIFF
--- a/src/cli/_lib/child-activity-select.ts
+++ b/src/cli/_lib/child-activity-select.ts
@@ -103,10 +103,16 @@ function runningChildren(
 function clauseFor(candidate: Candidate): string | undefined {
   if (candidate.silentMs >= CHILD_QUIET_MS) {
     // Static clause: intentionally NOT a live-ticking "no output for Xs" counter.
-    // A live counter would change the composed string on every recompose,
-    // breaking setOverlay's identical-string dedup (terminal-compositor.ts:794)
-    // and re-introducing the ghost-row/flicker class the 1500ms H2 throttle
-    // (stream-renderer-subagent.ts) exists to prevent. The tool-lane's
+    // A live counter would change THIS CLAUSE on every 80ms recompose. Note the
+    // dedup at terminal-compositor.ts:794 is not what makes that expensive to
+    // avoid — the composed banner already carries the child's elapsed
+    // durationMs rounded to whole seconds, so the overlay string is not
+    // byte-stable across a second boundary and that dedup cannot be relied on
+    // either way (see this module's header). The real cost is the rate: a
+    // per-clause counter churns at the 80ms tick instead of ~1Hz, an ~12x
+    // increase in real repaints, which re-introduces the ghost-row/flicker
+    // class the 1500ms H2 throttle (stream-renderer-subagent.ts) prevents. The
+    // conclusion is unchanged — keep the clause static. The tool-lane's
     // `· waiting Xs` annotation already shows elapsed silence past 30s
     // (PAUSE_THRESHOLD_MS); the banner just needs to signal the child went
     // quiet, which one static string does.
--- a/src/cli/_lib/child-activity-select.ts
+++ b/src/cli/_lib/child-activity-select.ts
@@ -46,11 +46,14 @@ export const STICKY_HOLD_MS = 3000;
  * (which stays at 30s): the banner reports "no output (waiting)" earlier so the
  * operator sees a quiet child during the 8s–30s dead zone that the tool-lane's
  * `· waiting Xs` annotation (armed only past 30s) does not cover. The static
- * clause avoids a live-ticking counter that would change the composed string on
- * every recompose (breaking setOverlay's identical-string dedup), so the banner
- * flips once at this threshold and then stays byte-stable —
- * `checkProgressBannerStaleness` rides the existing 80ms pause tick to trigger
- * that one flip without adding a new timer (see live-progress-no-timer.test.ts).
+ * clause avoids a live-ticking counter that would change the *clause* on every
+ * recompose; note the composed banner still carries the child's elapsed
+ * `durationMs` rounded to whole seconds, so the full overlay string is not
+ * byte-stable across a second boundary and setOverlay's identical-string dedup
+ * cannot be relied on to absorb repeat flushes. `checkProgressBannerStaleness`
+ * (stream-renderer-dead-zone.ts) therefore latches per source and flushes once
+ * per quiet transition, riding the existing 80ms pause tick rather than adding
+ * a timer (see live-progress-no-timer.test.ts).
  *
  * Invariant: must exceed `STICKY_HOLD_MS` (3s) so the sticky selector has settled
  * before the silence clause appears, and must exceed the 1500ms per-parent
--- a/src/cli/_lib/live-progress-no-timer.test.ts
+++ b/src/cli/_lib/live-progress-no-timer.test.ts
@@ -27,11 +27,13 @@ const REPO_ROOT = join(import.meta.dirname, '..', '..', '..');
 const TIMER_FREE_MODULES = [
   'src/cli/_lib/child-activity-select.ts',
   'src/cli/input/work-derived-verb.ts',
-  // stream-renderer-lifecycle.ts hosts checkProgressBannerStaleness, which
-  // rides the EXISTING 80ms pause tick (stream-renderer.ts:485) — it must not
-  // introduce its own setInterval/setTimeout. Guard this so a future refactor
+  // stream-renderer-lifecycle.ts hosts checkPauseAnnotations, and
+  // stream-renderer-dead-zone.ts hosts checkProgressBannerStaleness. Both ride
+  // the EXISTING 80ms pause tick (stream-renderer.ts:485) — neither may
+  // introduce its own setInterval/setTimeout. Guard both so a future refactor
   // cannot silently add one to the dead-zone path.
   'src/cli/_lib/stream-renderer-lifecycle.ts',
+  'src/cli/_lib/stream-renderer-dead-zone.ts',
 ] as const;
 
 function read(relPath: string): string {
--- a/src/cli/_lib/stream-renderer-dead-zone-integration.test.ts
+++ b/src/cli/_lib/stream-renderer-dead-zone-integration.test.ts
@@ -45,6 +45,9 @@ describe('dead-zone integration: checkProgressBannerStaleness -> registered prog
       childActivity: new ChildActivityTracker(),
     } as unknown as Parameters<typeof registerOverlaySlots>[1]);
 
+    // Capturing tool-lane so the sibling checkPauseAnnotations can be driven
+    // in the combined-tick case below, not just observed staying dormant.
+    const laneCalls: string[] = [];
     const ctx = {
       disposed: false,
       isTTY: true,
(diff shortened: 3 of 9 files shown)


griffinwork40:
# Review — PR #874 · `fix(tui): latch the dead-zone banner flush`

**Verdict: APPROVE** — merge after one MAJOR doc fix (a single-line comment edit). No blocking or functional defects. CI green on all six gates ([run 30819669735](https://github.com/griffinwork40/agent-afk/actions/runs/30819669735)).

## First: the preflight artifact was stale

The `/review` preflight diff was **662 lines**; the live PR diff is **831**. It was captured at `36192872` and omitted `7ef87541` entirely. Had I trusted it, I'd have filed a false BLOCKING finding — the PR body credits `7ef87541` with adding the combined-tick test for item 6, and that test is absent from the artifact. Verified against the branch: **the commit and test both exist; the PR body is accurate, the artifact was not.** The artifact should carry the head SHA it was captured at so the skill can detect drift.

## Findings

**[MAJOR] The retracted byte-stability claim survives a third time — in the file that retracts it.** `child-activity-select.ts:105-108` still says a live counter would "break setOverlay's identical-string dedup"; 55 lines above, the JSDoc this PR rewrote (`:48-53`) now says the dedup "cannot be relied on." One file, two contradictory statements about the same mechanism. Item #2 claims "both files fixed" — the header JSDoc and the new module were, the `clauseFor()` body was missed. Doc-only, but it's exactly the failure mode this PR exists to fix: a wrong comment that justified wrong code. The *conclusion* (keep the clause static) still holds — a live counter would churn at ~80ms instead of ~1s — so correct the mechanism, don't delete the guidance.

**[MINOR] ×4** — call-site JSDoc still attributes the function to the lifecycle module (`stream-renderer.ts:951-953`); the PR body overstates the capture-mode win (the ~1 repaint/sec is removed for 8–30s only — past 30s `checkPauseAnnotations` reintroduces the same cadence via its whole-second annotation at `stream-renderer-lifecycle.ts:285-298`); `quietBannerAnnounced`'s doc implies synchronous re-arm but it's sweep-lagged ≤80ms, unlike the `pauseAnnotation` it's modelled on; and three named latch-test gaps — stale-latch-on-completion unasserted, the multi-child test never reads per-source latch state (so it can't catch a regression to the old early `return`), and no same-sweep mixed transition.

**[NIT]** The 24-line `// Invariant:` block in the new module fuses a real invariant with root-cause narrative and a perf postmortem — `History:` content that `AFK.md` says migrates to `docs/` with a ≤5-line summary.

## The thesis holds

All three load-bearing claims re-derived from source and **CONFIRMED**: `markDirty` is one shared flag with no per-slot short-circuit (`overlay-composer.ts:44,74-86`); the banner really does carry whole-second `durationMs` so the string is not byte-stable (`progress-banner.ts:137` → `format-utils.ts:6`) — this is the correction the whole PR rests on, and it's right; and the 12.5 recomposes/sec arithmetic checks out. Also verified: no missed-repaint regression (the clause is derived at render, and any slot's flush recomputes it), the return-semantics change is inert (call site discards it), no import cycle, move is complete repo-wide, **no unused-import `noUnusedLocals` breakage** in the truncated test file, timer guard intact for both modules, O(N) sweep immaterial at N≈8, and the LOC table is accurate.

**Not verified:** tests not run locally (read-only, tree is on `main`) — relying on green CI; subagent-id recycling under DAG retry, currently moot since `done` is never reset.

---

**Done**
- Reviewed PR #874 at its real head `7ef87541` across correctness, API/wiring, test-coverage, and perf/observability/conventions dimensions via four parallel read-only subagents, then independently verified every load-bearing claim against the branch (the wave had read `main` plus a stale artifact).
- Evidence: full review written to `/Users/griffinlong/.afk/workspace/reports/pr-874-review-2026-08-03.md` (11.6 KB, severity-tagged findings with file:line, a claim-verification table, and a not-verified list); CI status read from GitHub Actions run 30819669735.
- Changed in the world: nothing — review was read-only. No files in the repo touched, no commit, no push, no PR comment or description edit. Working tree still clean on `main` at `58a1e68f`.
- Pending: the MAJOR finding (`child-activity-select.ts:105-108` self-contradiction) and the four MINORs are **not** filed on the PR — say the word and I'll post them as a review comment or fix them on the branch. The stale-preflight-artifact bug in the `/review` skill is unfiled.

---

PR #875, opened 2026-08-03
fix(awareness): make the Workspace line follow a mid-session cwd re-anchor

griffinwork40:
## What

A session re-anchored into a git worktree kept reporting the **launch** checkout's branch and HEAD for the rest of its life. The auto-injected `# Environment` block became self-contradictory:

```
- Working directory: /repo/.afk-worktrees/feature   <- correct
- Workspace: main @ ma1n000 (clean)                 <- the OTHER checkout
```

`get_runtime_state` split the same way — view `self` returned the launch cwd while the bash/file tools operated in the worktree.

This is not theoretical. It was found live, in a session running from a worktree:

| source | value |
|---|---|
| `- Working directory:` (prompt) | `.afk-worktrees/afk-remove-ground-state-endpoint` |
| bare `pwd` (bash tool) | `.afk-worktrees/afk-remove-ground-state-endpoint` |
| `- Workspace:` (prompt) | `afk/subagent-stream-cut-retry @ d35e0ee3` |
| `get_runtime_state` `self.cwd` | `/Users/.../agent-afk` (main checkout) |

The model was being told it was on a branch it was not on, by the very block meant to ground it.

**Trigger path — exactly one, and it is not `/cd`.** An earlier revision of this description claimed `/cd` as a second trigger; that is wrong. No REPL slash command calls `setCwd` (nothing under `src/cli/slash` does), and Telegram's `/cd` deliberately closes and rebuilds the session (`telegram/handlers/commands.ts:253,294`), so it never produces the stale state. `resume-swap` likewise constructs a fresh session. The only live-session re-anchor in the tree is the **deferred (born-named) `afk -w`** path:

1. `afk -w` with no explicit branch defers worktree creation (`interactive.ts:383`), so `worktreeCwd` stays `undefined` and bootstrap runs in the launch checkout.
2. The provider query — and with it the awareness source — is built at session *construction*: `initSdkLifecycle()` (`agent-session.ts:237`) → `ProviderRouter` (`:313`) → the first `pullInitialization` pull → `buildInner` → `provider.query()` → `buildRuntimeStateSource({ cwd: config.cwd ?? process.cwd() })` (`dispatcher-wiring.ts:110`). `config.cwd` is `undefined` here, so the snapshot is the **launch checkout**.
3. On the first message, `runFirstTurnAutoname` derives the slug from that message, creates the worktree, and calls `session.setCwd` (`worktree-autoname.ts:475`) — correctly awaited before `runTurn` (`loop-iteration.ts:443` vs `:606`). The rebuild re-threads the new cwd into the prompt, so `- Working directory:` is right, but hands the *same frozen* `runtimeStateSource` through.

Eager `-w <branch>` and `afk chat -w` await `setupWorktree` before the session is constructed, so the worktree is already in `config.cwd` and those paths were never affected.

## Root cause

`RuntimeSourceDeps.cwd` was a captured **string**, resolved once at source construction from `config.cwd ?? process.cwd()` (`dispatcher-wiring.ts:110`).

Every other live field on that interface is a `getX: () => T` accessor — `getEnabledToolNames`, `getMcpTools`, `getSubagents`. `cwd` was the lone snapshot. `dispatcher.setCwd` re-anchors the subagent, skill and compose executors (`dispatcher.ts:530-532`) but never the awareness source, so `gatherWorkspace()` kept being handed the outgoing directory.

Worth being precise: `cwd-dependents.ts` **already** re-read `getWorkspace()` on every rebuild. The call was live; its *input* was stale. That is why the bug survived a code path that looks correct on inspection.

Two comments had encoded the false premise and are corrected here — `prompt-assembly.ts:86` described the workspace as "stable across cwd swaps", and `cwd-dependents.ts` repeated it. A different worktree is a different branch and HEAD, so `workspace` is precisely the cwd-*dependent* field.

**Why this cannot be fixed by reordering instead.** The obvious alternative — snapshot the cwd later, after the worktree exists — is impossible on this path. The worktree's name is *derived from the first message*, and the query must already exist before that message can be streamed into it (`QueryInputStream` / prompt iterable, `router/provider-router.ts:261-272`). At construction time the worktree path is unknowable in principle, so the fix has to be late-binding. A live accessor is; an earlier snapshot cannot be.

This also matches the ownership model the router already implements: `ProviderRouter.setCwd` stores `this.currentCwd` (`router/provider-router.ts:459`) and re-injects it into every rebuilt inner (`innerConfig.cwd = this.currentCwd`, `:270`) — so a mid-session `/model` swap does *not* lose the re-anchor. cwd was already treated as live session state everywhere except this one field.

## Fix

`cwd: string` becomes `getCwd: () => string`, adopting the live-accessor convention already established in the same interface rather than introducing a new mechanism.

- **anthropic-direct** supplies `this._currentCwd || config.cwd || process.cwd()`. `_currentCwd` wins because `cwdDependentsFactory` updates it before re-reading the workspace; the `||` fall-through matches the sibling `cwd` const so the two Environment lines cannot disagree. `_currentCwd` is also re-seeded from each incoming `config.cwd` at `query()` entry — without that, a provider instance reused across a `/model` swap would let a *stale* `_currentCwd` outrank the newer per-query cwd the router injected (see the follow-up section below).
- **openai-compatible** is untouched in effect — its `setCwd` only re-bases the dispatcher (`openai-compatible/query.ts:1129-1131`) and its `# Environment` fragment is built once inside `query()` (`index.ts:327-335`), so that provider never re-emits the block at all. Kept as a callback to satisfy the contract.

  Stating the consequence plainly: on that provider this change is behaviour-neutral, and it neither creates nor closes a divergence. `getCwd()` and the `# Environment` fragment's `cwd` evaluate the *same* expression (`config.cwd ?? process.cwd()`) over a query-local `config` that is never reassigned, so after a re-anchor `- Working directory:`, `self.cwd`, and `- Workspace:` all stay pinned to the query's launch cwd and therefore agree with each other. The gap is that all three are frozen — not that they disagree. A `ProviderRouter` rebuild is the only thing that refreshes them, and it is gated on `modelChanged` (`router/provider-router.ts:336-339`), so an ordinary cwd-only `setCwd` never triggers one. Tracked in #876; deliberately not fixed here.

  (An earlier revision of this section claimed the change "makes `self.cwd` / `- Workspace:` live while the env block stays frozen — the divergence flips direction rather than closing." That was wrong, and is corrected above.)

An ordering invariant is now documented at the call site: `setCurrentCwd(newCwd)` must run before `getWorkspace()` in the rebuild, or git state is gathered for the outgoing checkout.

Four functional lines. The rest is tests and correcting the comments that made the bug invisible.

## Verification

- **5 tests added** — 2 unit (`runtime-source.test.ts`), 3 composed-path (`cwd-dependents.workspace.test.ts`, new file, asserts the rebuilt prompt string itself).
- **Mutation-checked.** With the accessor reverted to a captured string, all 5 fail with the exact stale value; with the fix, all 5 pass. They are not vacuous.
- Full suite: **766 files / 14,595 tests** green.
- `pnpm lint`, `audit:sdk:check`, `audit:env:check`, `scan:env:check` all pass.

## Scope

Deliberately narrow. This landed out of a wider audit of `/ground-state`, whose stated purpose is preventing exactly this class of stale-worktree error. The following follow-ons were considered and **excluded** from this PR because they are contested, higher-blast-radius, or not yet root-caused:

1. **A ~36x latency regression in skill forks** (median `/ground-state` duration went 0.28 min in 2026-05 to 10.78 min in 2026-07 across 294 traced dispatches). Cause not isolated, and evidence suggests it is *systemic fork latency* rather than skill-specific — an unrelated `general-purpose` subagent timed out at 45 min during this very investigation. Deserves its own diagnosis.
2. **Sub-agent forks receive neither the AFK.md overlay nor HOT.md**, so the actors doing most investigation work are the least grounded. Real, but a fork-briefing redesign.
3. **`ground-state`'s memory surveyor greps `~/.claude/projects/-<slug>/memory/`** — Claude Code's store, empty under AFK, whose real store is `~/.afk/state/memory/`. Correct for the Claude Code build of the same skill, so it needs a both-paths fix, not a swap.

Two same-bug-class siblings were found while auditing this fix's placement and are filed rather than folded in — both are captured-cwd defects on surfaces this PR does not touch:

4. **#876 — openai-compatible never rebuilds its `# Environment` block on `setCwd`**, so `- Working directory:` is stale there. See the Fix note above: this PR is behaviour-neutral on that provider — all three cwd-derived fields stay frozen together rather than disagreeing — and #876 is the close. A characterization test added in the follow-up commit pins that behaviour so #876 has a red/green signal.
5. **#877 — `GitStatusSampler` freezes its cwd at construction** (`git-status-sampler.ts:117` ← `bootstrap.ts:627`) with no re-point, so the status line's branch and open-PR fields are sampled from the launch checkout for the whole session after a deferred `-w` re-anchor.

None of those are prerequisites for this one, and this one makes the free grounding path trustworthy regardless of what is decided about the skill.

---

## Follow-up commit `73262f4` — review feedback

Addresses the Codex review comment and the automated review on this PR. One functional line; the rest is tests and comment corrections.

**1. Codex P2 (`index.ts` — reset the live cwd when reusing a cached provider).** Real defect, and it reintroduced this PR's own bug on a path the original fix missed. `_currentCwd` is an instance field that outlives a `query()`, but provider instances are cached across `/model` swaps and `ensureSharedRoots()` seeds it only once, while `ProviderRouter.setCwd` forwards a new cwd only to the *active* query. Sequence: Claude@A → swap to GPT → re-anchor to B → swap back to Claude reuses the cached provider with `config.cwd === B` but `_currentCwd === A`, so the stale value wins the `||` and `- Workspace:` / `get_runtime_state` report A while the prompt and tools use B. Fixed by re-seeding `_currentCwd` from the incoming config at `query()` entry (guarded on truthiness so an absent `config.cwd` cannot clobber it); later `setCwd()` calls still win. Regression test added — verified to fail without the fix (`expected '/checkout-a' to be '/checkout-b'`).

**2. Removed a false premise from 8 comment lines.** This PR's own description establishes that `/cd` is *not* a live-session re-anchor trigger — no REPL slash command calls `setCwd`, and Telegram's `/cd` closes and rebuilds the session (`telegram/session-manager.ts:564-571`) — yet six comment lines added here named `/cd` as exactly that, including inside an `Invariant:` block. Struck from all six, plus two pre-existing instances in `cwd-dependents.ts`. The only live-session re-anchor named in comments is now the deferred born-named `afk -w` path.

**3. The `getCwd` invariant no longer overstates.** It read `Invariant: this MUST be a callback, never a snapshot`, which is false for openai-compatible (a per-query snapshot by design, pending #876). Reworded to state the requirement and name the exception. The openai-compatible comment's "`config.cwd` IS the live value for the duration of a query" — which reads as a liveness guarantee but only describes staticness — now says plainly that it is frozen at query construction.

**4. The ordering invariant is now tested.** `setCurrentCwd(newCwd)` before `getWorkspace()` was comment-only, with no call-order assertion anywhere in the tree; the existing composed test observes only end state, so a reorder passed. Added an `invocationCallOrder` assertion — verified to fail when the statements are swapped (`expected 6 to be less than 5`).

**5. Characterization test for openai-compatible's `getCwd`** (previously untested), pinning the #876 behaviour.

**6. Fixed an unresolvable `{@link RuntimeSourceDeps.getCwd}`** in `dispatcher-wiring.ts` — the type was not imported.

**Verification:** `tsc --noEmit` clean; `audit:sdk:check` / `audit:env:check` / `scan:env:check` all pass; full suite 765/766 files green, 14,597 tests passing. The single failing file (`trace/writer.test.ts`, a test that spawns a real subprocess) fails identically at the pre-fix baseline in the same environment and passes 18/18 where dependencies are installed — a missing-deps artifact of the verification worktree, not a regression.

### DIFF
--- a/src/agent/awareness/runtime-source.test.ts
+++ b/src/agent/awareness/runtime-source.test.ts
@@ -23,7 +23,7 @@ vi.mock('./workspace-source.js', () => ({
 function defaultDeps() {
   return {
     surface: 'cli',
-    cwd: '/work',
+    getCwd: () => '/work',
     modelName: 'sonnet',
     providerName: 'anthropic-direct',
     permissionMode: 'default',
@@ -278,10 +278,42 @@ describe('buildRuntimeStateSource.getWorkspace', () => {
     expect(vi.mocked(gatherWorkspace)).not.toHaveBeenCalled();
   });
 
-  it('passes deps.cwd through to gatherWorkspace on each read', () => {
+  it('passes deps.getCwd() through to gatherWorkspace on each read', () => {
     vi.mocked(gatherWorkspace).mockReturnValue(ws(0));
-    const src = buildRuntimeStateSource({ ...defaultDeps(), cwd: '/custom/work' });
+    const src = buildRuntimeStateSource({ ...defaultDeps(), getCwd: () => '/custom/work' });
     src.getWorkspace();
     expect(vi.mocked(gatherWorkspace)).toHaveBeenCalledWith('/custom/work');
+  });
+
+  it('re-reads getCwd per call so a mid-session setCwd re-anchors the repo', () => {
+    // Regression guard (worktree staleness): `cwd` used to be a captured
+    // string, so after an `afk -w` / `/cd` re-anchor the awareness layer kept
+    // gathering git state for the LAUNCH checkout. The symptom was a session
+    // whose `- Working directory:` line pointed at the new worktree while its
+    // `- Workspace:` line reported the original checkout's branch and HEAD.
+    vi.mocked(gatherWorkspace).mockReturnValue(ws(0));
+    let cwd = '/launch/checkout';
+    const src = buildRuntimeStateSource({ ...defaultDeps(), getCwd: () => cwd });
+
+    src.getWorkspace();
+    expect(vi.mocked(gatherWorkspace)).toHaveBeenLastCalledWith('/launch/checkout');
+
+    // Simulate cwdDependentsFactory's setCurrentCwd() on a worktree swap.
+    cwd = '/repo/.afk-worktrees/feature';
+
+    src.getWorkspace();
+    // The captured-string implementation would still pass '/launch/checkout'.
+    expect(vi.mocked(gatherWorkspace)).toHaveBeenLastCalledWith('/repo/.afk-worktrees/feature');
+  });
+
+  it('reports the re-anchored cwd on getSelf so the two never disagree', () => {
+    // `get_runtime_state` view 'self' and view 'workspace' must resolve the
+    // same cwd — a divergence there is what made the bug hard to see.
+    let cwd = '/launch/checkout';
+    const src = buildRuntimeStateSource({ ...defaultDeps(), getCwd: () => cwd });
+
+    expect(src.getSelf().cwd).toBe('/launch/checkout');
+    cwd = '/repo/.afk-worktrees/feature';
+    expect(src.getSelf().cwd).toBe('/repo/.afk-worktrees/feature');
   });
 });
--- a/src/agent/awareness/runtime-source.test.ts
+++ b/src/agent/awareness/runtime-source.test.ts
@@ -287,8 +287,9 @@ describe('buildRuntimeStateSource.getWorkspace', () => {
 
   it('re-reads getCwd per call so a mid-session setCwd re-anchors the repo', () => {
     // Regression guard (worktree staleness): `cwd` used to be a captured
-    // string, so after an `afk -w` / `/cd` re-anchor the awareness layer kept
-    // gathering git state for the LAUNCH checkout. The symptom was a session
+    // string, so after the deferred born-named `afk -w` worktree re-anchor the
+    // awareness layer kept gathering git state for the LAUNCH checkout. The
+    // symptom was a session
     // whose `- Working directory:` line pointed at the new worktree while its
     // `- Workspace:` line reported the original checkout's branch and HEAD.
     vi.mocked(gatherWorkspace).mockReturnValue(ws(0));
--- a/src/agent/awareness/runtime-source.ts
+++ b/src/agent/awareness/runtime-source.ts
@@ -2,9 +2,10 @@
  * Constructs a {@link RuntimeStateSource} from the provider's per-query state.
  *
  * Shared by both `anthropic-direct` and `openai-compatible` providers so the
- * resulting snapshot shape is provider-agnostic. The provider supplies live
- * accessors (lambdas), not snapshots, so subsequent `get_runtime_state` calls
- * see up-to-date subagent counts, MCP tool counts, and git workspace state.
+ * resulting snapshot shape is provider-agnostic. Providers supply callbacks so
+ * subsequent `get_runtime_state` calls can see up-to-date subagent counts, MCP
+ * tool counts, and git workspace state; openai-compatible's cwd callback is a
+ * known per-query snapshot until #876.
  *
  * @module agent/awareness/runtime-source
  */
@@ -30,13 +31,14 @@ export interface RuntimeSourceDeps {
   /**
    * Live accessor for the session working directory.
    *
-   * Invariant: this MUST be a callback, never a snapshot. `setCwd()` (the
-   * `afk -w` born-named worktree path, `/cd`, and worktree re-anchoring) can
-   * repoint the session mid-flight, and both `getSelf().cwd` and
-   * `getWorkspace()` have to follow it. A captured string silently pins the
-   * awareness layer to the launch directory, so `get_runtime_state` and the
-   * `- Workspace:` prompt line keep reporting the ORIGINAL checkout's branch
-   * and HEAD while the tools operate in the new one.
+   * Invariant: providers SHOULD supply a live callback, not a captured string.
+   * The deferred born-named `afk -w` worktree path can repoint an anthropic-direct
+   * session mid-flight, and both `getSelf().cwd` and `getWorkspace()` have to
+   * follow it. A captured string silently pins the awareness layer to the launch
+   * directory, so `get_runtime_state` and the `- Workspace:` prompt line keep
+   * reporting the ORIGINAL checkout's branch and HEAD while the tools operate in
+   * the new one. Known exception: openai-compatible currently freezes this value
+   * per query; fixing its mid-query `setCwd()` staleness is tracked in #876.
    */
   getCwd: () => string;
   /** Resolved model identifier the SDK will be called with. */
@@ -90,10 +92,11 @@ export interface RuntimeSourceDeps {
  * branch when it orients, not a frozen session-start snapshot that silently
  * goes stale as files change.
  *
- * `deps.getCwd` is likewise a callback rather than a string: a session can be
- * repointed mid-flight (`afk -w` born-named worktrees, `/cd`), and a captured
- * cwd would pin the awareness layer to the launch checkout — reporting the
- * wrong branch and HEAD for the entire remaining session.
+ * `deps.getCwd` is likewise a callback rather than a string: anthropic-direct
+ * sessions can be repointed mid-flight by the deferred born-named `afk -w`
+ * worktree path, and a captured cwd would pin the awareness layer to the launch
+ * checkout — reporting the wrong branch and HEAD for the remaining session.
+ * Known exception: openai-compatible freezes this value per query until #876.
  */
 export function buildRuntimeStateSource(deps: RuntimeSourceDeps): RuntimeStateSource {
   return {
--- a/src/agent/awareness/runtime-source.ts
+++ b/src/agent/awareness/runtime-source.ts
@@ -27,8 +27,18 @@ export interface RuntimeSourceDeps {
   sessionId?: string | undefined;
   /** Provider-level surface tag (e.g. 'cli', 'daemon', 'telegram'). */
   surface: string;
-  /** Working directory at query time. */
-  cwd: string;
+  /**
+   * Live accessor for the session working directory.
+   *
+   * Invariant: this MUST be a callback, never a snapshot. `setCwd()` (the
+   * `afk -w` born-named worktree path, `/cd`, and worktree re-anchoring) can
+   * repoint the session mid-flight, and both `getSelf().cwd` and
+   * `getWorkspace()` have to follow it. A captured string silently pins the
+   * awareness layer to the launch directory, so `get_runtime_state` and the
+   * `- Workspace:` prompt line keep reporting the ORIGINAL checkout's branch
+   * and HEAD while the tools operate in the new one.
+   */
+  getCwd: () => string;
   /** Resolved model identifier the SDK will be called with. */
   modelName: string;
   /** Provider name — e.g. 'anthropic-direct' or 'openai-compatible'. */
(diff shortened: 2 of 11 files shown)


griffinwork40:
## Review — PR #875 `fix(awareness): make the Workspace line follow a mid-session cwd re-anchor`

**Target** `afk/fix-awareness-cwd-staleness` → `main` · reviewed ref **`3c43fc45`** · 9 files · +240/−21 · **4 functional lines**, remainder tests + comments
**Waves** Wave 1 ×2 (security·api-compat / correctness·spec·tests·perf) → Wave 1.5 inline verification → synthesis inline → shadow-verify ×2

---

### Blocking

**1 · `high` · confidence high · spec-compliance · shadow-verified CONFIRMED**
`src/agent/awareness/runtime-source.ts:34`, `:94` · `src/agent/awareness/types.ts:140` · `src/agent/awareness/runtime-source.test.ts:290` · `src/agent/providers/anthropic-direct/index.ts:467` · `.../query/cwd-dependents.workspace.test.ts:11` · `ref:3c43fc45` · citation-type: diff-context

**Finding:** the diff adds **6 comment lines naming `/cd` as a live-session `setCwd()` re-anchor trigger** — including inside a doc block labelled `Invariant:` — while the PR's own body states the opposite as a *corrected* claim, and the code disproves it.

**Evidence** (`runtime-source.ts:33-34`):
```
* Invariant: this MUST be a callback, never a snapshot. `setCwd()` (the
* `afk -w` born-named worktree path, `/cd`, and worktree re-anchoring) can
```
Independently re-derived: zero `setCwd` under `src/cli/slash`, and no `/cd`/`/cwd` command exists in the REPL registry (`src/cli/slash/index.ts:41-78`). Telegram's `/cd` (`bot.ts:153-155` → `handlers/commands.ts:294`) calls `SessionManager.setCwd`, whose body closes and deletes the session (`session-manager.ts:564-571`) — a different class from `AgentSession.setCwd`. The sole live-session re-anchor is `worktree-autoname.ts:475`.

**Why blocking rather than cosmetic:** this PR's non-test payload *is* comments, and its stated intent is "correct the comments that encoded the false premise." It removes one false premise and installs six instances of another, in the exact doc block a future maintainer will trust. Behavioral impact is **zero** — re-rank freely if you weigh it as documentation-only.

**Suggestion:** delete `, /cd` / `` `/cd` `` from those 6 lines; replace with "the deferred (born-named) `afk -w` path — the only live-session re-anchor." Two further pre-existing instances sit at `cwd-dependents.ts:119,132` (not introduced here).

---

### Non-blocking

**2 · `medium` · high · api-compat/correctness** — `src/agent/providers/openai-compatible/index.ts:223` · ref:3c43fc45 · diff-context
`getCwd: () => config.cwd ?? process.cwd()` satisfies `() => string` but is a per-query snapshot: `config` is a `const` alias (`:197`) never reassigned, and `setCwd` only re-bases the dispatcher (`query.ts:1129-1131`). The new `Invariant: this MUST be a callback, never a snapshot` doc is therefore false for one of the interface's two production implementations. Consequence on a gpt-*/o*/local-model REPL: after a deferred `afk -w` re-anchor, `- Working directory:`, `self.cwd`, **and** `- Workspace:` all stay pinned to the launch checkout for the session's life — the router only re-injects cwd on a rebuild, gated on `modelChanged` (`provider-router.ts:268,336-339`). Pre-existing, disclosed, filed as #876 → no regression. *Suggestion:* scope the interface doc per-provider, or route this provider's `getCwd` through a field its `setCwd` also updates.

**3 · `medium` · high · spec-compliance (PR description accuracy)** — shadow-verified **REFUTED**
The body's claim *"this change makes `self.cwd` / `- Workspace:` live while the env block stays frozen — the divergence flips direction rather than closing"* is false. `getCwd()` and `formatEnvironmentFragment`'s `cwd` (`index.ts:332-333`) evaluate the *same* expression over the same frozen object: all three fields stay OLD and **agree**. No divergence is created or closed; the change is behaviour-neutral there. Two independent derivations agree. *Suggestion:* correct the sentence and re-check #876's scope, which rests on it.

**4 · `medium` · high · test-coverage** — `.../query/cwd-dependents.ts:76,105` · ref:3c43fc45 · diff-context
The newly documented ordering invariant (`setCurrentCwd(newCwd)` must precede `getWorkspace()`) is comment-only. Confirmed-absent: zero `invocationCallOrder` / `toHaveBeenCalledBefore` / `callOrder` under `anthropic-direct` or `awareness`. The new composed test asserts the end state of one synchronous call, so swapping lines 76 and 105 still passes. *Suggestion:* spy both and assert `setCurrentCwd`'s call index < `getWorkspace`'s.

**5 · `low` · medium · test-coverage** — `openai-compatible/index.ts:223` — no test exercises this provider's awareness `getCwd`; the only `setCwd` test (`query.test.ts:1787`) covers the dispatcher path. Finding #2 has zero regression coverage.

**6 · `low` · high · correctness (comment)** — `openai-compatible/index.ts:219-222` — "so `config.cwd` IS the live value for the duration of a query" is true-but-misleading (staticness, not liveness); shadow verifier identifies it as the likely source of finding #3's error.

**7 · `nit` · low · api-compat** — `dispatcher-wiring.ts:71` — `{@link RuntimeSourceDeps.getCwd}` is unresolvable; only `RuntimeStateSource`/`RuntimeSubagents` are imported (`:51`, `:55`).

**8 · `low` · medium · security · pre-existing, not introduced** — `workspace-source.ts:114`, `types.ts:158`, `tool.ts:48` — `remoteUrl` is gathered and returned unredacted (zero `redact|sanitiz|scrub` matches in `awareness/`), so an `https://user:token@host/...` origin reaches the model in `get_runtime_state` JSON. Not rendered into the system prompt (`runtime-snapshot.ts:187` emits branch/sha/clean only), and this diff does not change reachability.

---

### Verified clean

- **Ordering invariant holds mechanically** — `setCurrentCwd(newCwd)` at `cwd-dependents.ts:76`, `getWorkspace()` at `:105`, same synchronous body, no intervening await.
- **`_currentCwd` chain correct** — decl `:147`, seeded from `config.cwd` (`:439` → `:356-357`), written by `setCurrentCwd` (`:541`), read fresh at `:471`.
- **`||` claim accurate** — sibling is `const cwd = config.cwd || process.cwd()` (`:484`), also `||`. The two Environment lines cannot disagree, as stated.
- **Migration complete, zero external blast radius** — both production construction sites (`dispatcher-wiring.ts:114`, `openai-compatible/index.ts:217`) plus all test sites use `getCwd`; zero residual `cwd:`; `src/index.ts` exports no awareness symbols.
- **No authorization path touched** — the permission gate reads a *different* live `getCwd` (`default-hook-registry.ts:153,219` ← `bootstrap-hooks.ts:47`); nothing derives roots or path approval from `getSelf()`/`getWorkspace()`.
- **Perf neutral** — zero new `getWorkspace()` call sites; the per-call re-gather predates this PR (diff pre-image was already `gatherWorkspace(deps.cwd)`). Six production call sites total, all unchanged in frequency.
- **New composed test is not vacuous** — mocks only `gatherWorkspace`, runs the real `assembleSystemPrompt`, asserts the rebuilt prompt string. All 5 named tests exist.

### Wave 1.5 corrections (audit trail)

- **DROPPED — false attribution:** a `medium` "scope creep" finding claiming this diff routes the dispatcher rebuild through `getCurrentPermissionMode()` and adds `planExitControls`. Both pre-exist at merge base `e2348e57` (lines 110-113, 123-125); the diff to `cwd-dependents.ts` is **comment-only**.
- **REPOINTED — 4 citations** quoted diff-artifact line numbers as file line numbers (`openai-compatible/index.ts:392-396` → `:217-223`; `dispatcher-wiring.ts:362` → `:116`; `:172-173` → `:162,:173`). Content real; refs corrected above.
- **Counts corrected:** "5 construction sites" → 2 production + 18 test; "3 `getWorkspace()` call sites" → 6. Both conclusions survived.

### What was not checked

- Citations verified inline against branch HEAD **`3c43fc45`** (= checked-out HEAD, clean tree); every `blocking`/`high` and every `file-state` citation re-read at the ref.
- **Stated intent:** PR #875 title + body — spec-compliance **assessed**.
- Tests were **not executed** — the "766 files / 14,595 tests green" and mutation-check claims are unverified (read-only review; no runner invoked).
- Wave 2 synthesis run **inline instead of dispatched** — the verification evidence lived in this session's shell output, and re-serializing it to a shell-less agent was the larger fabrication risk. Findings were deduped and sorted by the documented rule.
- Not reviewed: Telegram surface behaviour end-to-end, the `#877` `GitStatusSampler` sibling, the latency-regression claims in the PR's Scope section, and whether `#876` should absorb finding #2's re-scoping.

---

## MERGE DECISION: **DO NOT MERGE**

Blocked on finding **#1** only. The functional change is correct, minimal, and genuinely well-tested — the `getCwd` accessor, the `_currentCwd` chain, the ordering, and the migration all verify clean, and nothing behavioral is wrong. The block is that a PR whose payload is overwhelmingly *comments*, and whose stated purpose is removing a false premise from comments, installs six new instances of a premise its own description refutes. Deleting `, /cd` from 6 lines clears it; findings #2–#4 are reasonable follow-ups (#2/#3 belong in #876).

**Done**
- Reviewed PR #875 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability; 8 findings surfaced (1 blocking), 1 sub-agent finding dropped as false attribution, 1 PR-body claim refuted, 4 citations repointed.
- Evidence: reviewed ref `3c43fc459fb08e65cff22aaf788d8b2f61441ae3`; diff artifact `/Users/griffinlong/.afk/state/skill-preflight/9f557cd1-a50b-43e3-860f-9bc7bc653366/pr-875.diff`; all citations re-read at the ref via `git show`/`git grep`; two independent shadow verifiers returned CONFIRMED on the blocking claim and on the PR-body refutation.
- World state unchanged — read-only review: no files written, no commits, no PR comments or edits posted. Working tree still clean at `3c43fc45`.
- Pending: the merge verdict is advisory only; nothing was communicated to the PR. Findings #2/#3 imply issue #876 is mis-scoped (its premise sentence is refuted), which no one has been told.

---

PR #879, opened 2026-08-03
fix(subagent): resolve tier-name pins before the cross-provider Claude-family check

### LINKED ISSUE #869: subagent: tier-name model pins bypass the cross-provider child-model fallback
## Problem

The cross-provider child-model fallback added in #652 rescues a Claude-family child that would starve on an `openai-compatible` session — but it only recognises **identity aliases**, not **tier names**. An agent pinned `model: sonnet` is rescued; the same agent pinned `model: medium` is not, even though `medium` resolves to a Claude model.

## Mechanism

`coerceCrossProviderChildModel` (`src/agent/subagent/child-model-fallback.ts`) bails on its first condition:

```ts
if (
  childModel === undefined ||
  !isClaudeFamilyModel(childModel) ||
  providerForModel(childModel) !== 'openai-compatible'
) {
  return { model: childModel };
}
```

and `isClaudeFamilyModel` (`src/agent/providers/openai-compatible/responses-config.ts:43-49`) only matches `claude-*` / `claude_*` / `claude`, `local-*` / `local_*`, and `/^(opus|sonnet|haiku)(?:[._-].*)?$/`:

```ts
export function isClaudeFamilyModel(model: string | undefined): boolean {
  const m = (model ?? '').trim().toLowerCase();
  if (!m) return false;
  if (m.startsWith('claude-') || m.startsWith('claude_') || m === 'claude') return true;
  if (m.startsWith('local-') || m.startsWith('local_')) return true;
  return /^(opus|sonnet|haiku)(?:[._-].*)?$/.test(m);
}
```

`'medium'` matches none of them, so `isClaudeFamilyModel('medium')` is `false` and no coercion happens. Downstream, `providerForModel` **does** resolve the tier (`src/agent/providers/index.ts:166` → `resolveBinding`) to `claude-sonnet-5` (`src/agent/session/model-slots.ts:114`), the request goes to the OpenAI endpoint, and it hard-errors on that provider's Claude guard — precisely the failure #652 exists to prevent.

Affects the bare tier names `small`, `medium`, `large`, and `local` (note `local-*` with a suffix *is* matched, but bare `local` is not).

## Reproduction

1. Set `AFK_PROVIDER=openai-compatible` (or use a chatgpt-oauth / openai model slot).
2. Dispatch a named agent whose definition declares `model: medium`.
3. The child resolves to `claude-sonnet-5` and fails against the OpenAI endpoint, with no coercion attempted.

Contrast: the same definition with `model: sonnet` is coerced onto the parent's model and runs.

## Suggested fix

Either:

1. Resolve the slot binding **before** the coercion check, so `coerceCrossProviderChildModel` sees the concrete id (`claude-sonnet-5`) rather than the unresolved tier name — preferable, since it fixes the class rather than enumerating names; or
2. Extend `isClaudeFamilyModel` to recognise tier names that currently *bind* to Claude ids — cheaper, but couples the predicate to slot bindings, which are rebindable at runtime.

Option 1 is likely correct: the predicate should never have to reason about tier indirection.

## Notes

Pre-existing; independent of #868 (which changes what an *omitted* model resolves to, and is unaffected because `getDefaultSubagentModel` returns the parent model for openai-compatible parents before the `medium` tier is reached). This issue is specifically about an **explicit** tier pin in an agent definition.

griffinwork40:
## Summary

Fixes #869 — bare capability-tier model pins (`small` / `medium` / `large` / `local`, or a custom tier name) bypassed the cross-provider child-model fallback from #652.

`isClaudeFamilyModel` (`src/agent/providers/openai-compatible/responses-config.ts`) pattern-matches the literal model string and has no notion of tier indirection. `coerceCrossProviderChildModel` (`src/agent/subagent/child-model-fallback.ts`) called it directly on the caller-supplied model, so an agent definition pinned to `model: medium` read as "not Claude-family" even though the `medium` tier is bound to `claude-sonnet-5` by default. The #652 fallback was skipped, and the child still routed to the `openai-compatible` provider under an `AFK_PROVIDER=openai-compatible` force (or a `chatgpt-oauth`/`openai` model slot) and hard-errored there — the exact failure #652 exists to prevent, reached through tier indirection instead of a raw `sonnet`/`opus`/`haiku` pin.

The reported issue scopes this to an explicit tier pin on the **child**; the same predicate is also applied to the **parent** model inside the same function, so the fix resolves both sides symmetrically rather than leaving a mirrored gap.

## Fix

Resolve the model through `resolveModelInput` (`model-slots.ts`) before the Claude-family check, on both the `childModel` and `parentModel` branches — option 1 from the issue. `isClaudeFamilyModel` itself is untouched (it stays self-contained to avoid an import cycle through the provider registry, per its existing docstring) and is still exercised directly by `responses-config.test.ts`. `resolveModelInput` is a no-op on an already-concrete id or a fixed identity alias (`sonnet`/`opus`/`haiku`/`claude-*`/`local-*`), so every pre-existing case is unaffected.

Minimal diff: one new private helper (`isClaudeFamilyAfterTierResolution`) + swapping the two call sites in `coerceCrossProviderChildModel`. No change to `isClaudeFamilyModel`, `providerForModel`, or any downstream resolution path — those were already verified correct (`buildQueryFromConfig`/`setModel` in `openai-compatible/query.ts` already resolve tiers before the wire request, so this gap was isolated to this one pre-resolution check).

## Tests

Added a `bare tier-name pins (#869)` block to `child-model-fallback.test.ts` covering:
- default-Claude-bound `small`/`medium`/`large` tiers coercing under a provider force
- the unconfigured bare `local` tier correctly **not** coercing (empty id, nothing to protect)
- a tier explicitly rebound to a non-Claude id correctly **not** coercing
- a custom tier name that resolves onto a Claude-bound slot coercing correctly

## How was this tested?

```
pnpm lint                                                    # tsc --noEmit — pass
pnpm test src/agent/subagent/child-model-fallback.test.ts    # 20/20 pass (11 pre-existing + 9 new)
pnpm test                                                    # full suite: 770 files, 14670 passed, 2 skipped (pre-existing live/pty exclusions)
pnpm audit:sdk:check                                         # pass, no new SDK imports
pnpm audit:env:check                                         # pass, no new process.env reads
pnpm scan:env:check                                          # pass, env registry unaffected
```

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes

### DIFF
--- a/src/agent/subagent/child-model-fallback.test.ts
+++ b/src/agent/subagent/child-model-fallback.test.ts
@@ -1,5 +1,11 @@
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import { coerceCrossProviderChildModel } from './child-model-fallback.js';
+import {
+  CLAUDE_SONNET_ID,
+  DEFAULT_SLOT_BINDINGS,
+  resetSlotBindings,
+  setSlotBindings,
+} from '../session/model-slots.js';
 
 /**
  * `coerceCrossProviderChildModel` reads `AFK_PROVIDER` (and `AFK_OPENAI_BASE_URL`)
@@ -80,5 +86,63 @@ describe('coerceCrossProviderChildModel (#652)', () => {
 
   it('passes an undefined child through unchanged', () => {
     expect(coerceCrossProviderChildModel(undefined, 'gpt-5.5')).toEqual({ model: undefined });
+  });
+
+  /**
+   * #869: a bare capability-tier pin (`small`/`medium`/`large`/`local`, or a
+   * custom tier name — e.g. an agent definition's `model: medium`) must be
+   * resolved to its bound concrete id before the Claude-family check, not
+   * checked as a literal string. `resetSlotBindings()` in `afterEach` prevents
+   * the explicit `setSlotBindings` calls below from leaking into other tests.
+   */
+  describe('bare tier-name pins (#869)', () => {
+    afterEach(() => {
+      resetSlotBindings();
+    });
+
+    describe('no global force (normal Anthropic routing)', () => {
+      it.each(['small', 'medium', 'large'])(
+        'leaves tier %s untouched — its default binding routes anthropic-direct',
+        (tier) => {
+          expect(coerceCrossProviderChildModel(tier, 'gpt-5.5')).toEqual({ model: tier });
+        },
+      );
+    });
+
+    describe('AFK_PROVIDER=openai-compatible force', () => {
+      beforeEach(() => {
+        process.env['AFK_PROVIDER'] = 'openai-compatible';
+      });
+
+      it.each(['small', 'medium', 'large'])(
+        'coerces bare tier %s — default-bound to a Claude id, so it would otherwise hard-error',
+        (tier) => {
+          expect(coerceCrossProviderChildModel(tier, 'gpt-5.5')).toEqual({
+            model: 'gpt-5.5',
+            coercedFrom: tier,
+          });
+        },
+      );
+
+      it('does NOT coerce the bare "local" tier when unconfigured (empty id — nothing to protect)', () => {
+        expect(coerceCrossProviderChildModel('local', 'gpt-5.5')).toEqual({ model: 'local' });
+      });
+
+      it('does NOT coerce a tier explicitly rebound to a non-Claude id', () => {
+        setSlotBindings({ ...DEFAULT_SLOT_BINDINGS, small: { id: 'gpt-4o-mini' } });
+        expect(coerceCrossProviderChildModel('small', 'gpt-5.5')).toEqual({ model: 'small' });
+      });
+
+      it('coerces a custom tier NAME that resolves onto a Claude-bound slot', () => {
+        setSlotBindings({
+          ...DEFAULT_SLOT_BINDINGS,
+          medium: { id: CLAUDE_SONNET_ID, name: 'general' },
+        });
+        expect(coerceCrossProviderChildModel('general', 'gpt-5.5')).toEqual({
+          model: 'gpt-5.5',
+          coercedFrom: 'general',
+        });
+      });
+    });
   });
 });
--- a/src/agent/subagent/child-model-fallback.ts
+++ b/src/agent/subagent/child-model-fallback.ts
@@ -17,10 +17,15 @@
  * OpenAI/gpt target the session is already using — so the child runs instead of
  * dying. It is applied at `SubagentManager.forkSubagent`, the single choke point
  * through which every fork path converges.
+ *
+ * Issue #869: the Claude-family check must run on the TIER-RESOLVED id, not the
+ * raw pin — see {@link isClaudeFamilyAfterTierResolution}, the sole difference
+ * from the original #652 shape.
  */
 
 import { providerForModel } from '../providers/index.js';
 import { isClaudeFamilyModel } from '../providers/openai-compatible/responses-config.js';
+import { resolveModelInput } from '../session/model-slots.js';
 
 export interface ChildModelCoercion {
   /** The model the child should actually run — the substitute when coerced. */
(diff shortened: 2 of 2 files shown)


griffinwork40 (commented):

## Merge decision: MERGE

Light regime (hotfix, 2 files, +91/−3). Reviewed ref `22f656e0`. No blocking findings — two non-blocking items below.

---

### Findings

**low · confidence high · test-coverage · `src/agent/subagent/child-model-fallback.test.ts` (new `bare tier-name pins (#869)` block) · ref:22f656e0 · citation-type:diff-context**

Every new test exercises tier resolution only through the `childModel` argument; no case passes a bare tier name as `parentModel`, so the symmetric parent-side branch the PR body explicitly claims to fix ("the same predicate is also applied to the parent model ... the fix resolves both sides symmetrically") is never exercised with a tier-bound parent.

Evidence:

```ts
it.each(['small', 'medium', 'large'])(
  'coerces bare tier %s ...', (tier) => {
    expect(coerceCrossProviderChildModel(tier, 'gpt-5.5')).toEqual({
      model: 'gpt-5.5', coercedFrom: tier,
    });
```

Verified absent at the ref: all 14 `coerceCrossProviderChildModel(...)` call sites in the file pass `'gpt-5.5'`, `'opus'`, or `undefined` as the second argument — never a tier name (`small`/`medium`/`large`/`local`/custom). Note `'opus'` is a fixed-identity alias, not a tier, so it does not cover this path.

Suggestion: add one case under the `AFK_PROVIDER=openai-compatible` force asserting `coerceCrossProviderChildModel('gpt-5.5', 'medium')` returns `{ model: 'gpt-5.5' }` — the parent tier-resolves to a Claude id, so coercion is correctly declined. That locks the second call site against regression.

**nit · confidence high · spec-compliance · `src/agent/subagent/child-model-fallback.ts:113-116` (docstring) and the PR body · ref:22f656e0 · citation-type:file-state (re-read and verified at ref)**

The claim that `resolveModelInput` is "a no-op on an already-concrete id or a fixed identity alias (`sonnet`/`opus`/`haiku`/`claude-*`/`local-*`)" is imprecise for the three short aliases: they are rewritten to their concrete ids via `DIRECT_MODEL_ALIASES`, not passed through. Only `claude-*`/`local-*` and concrete ids are literal no-ops.

Evidence (`model-slots.ts:307-309`, verified at ref):

```ts
const directId = DIRECT_MODEL_ALIASES[input.trim().toLowerCase()];
if (directId) return { id: directId };
return { id: input };
```

This is a wording issue only — behaviour is unaffected, since `isClaudeFamilyModel('sonnet')` and `isClaudeFamilyModel('claude-sonnet-5')` both return `true`, so the predicate result is identical either way. Suggestion: say "produces the same `isClaudeFamilyModel` result" rather than "is a no-op" for `sonnet`/`opus`/`haiku`.

---

### Verified correct (the highest-risk question)

The diff swaps the Claude-family predicate to the tier-resolved id on both branches but leaves `providerForModel(childModel)` / `providerForModel(parentModel)` on the **raw** pin. That asymmetry was traced against the reviewed ref and is **correct by design, not an oversight**:

- Under a global `AFK_PROVIDER` force, `providerForModel` short-circuits at Tier 1 (`providers/index.ts:156-157`) before any resolution, so raw vs. resolved is irrelevant.
- Without a force, `providerForModel` resolves internally at `providers/index.ts:166` (`const binding = resolveBinding(model, hints?.slots)`), so it reaches the same bound id the new predicate sees.
- Passing the **raw** pin is strictly more correct than passing the resolved id would be: `resolveBinding('medium')` returns the binding including its per-slot `.provider` override, whereas `resolveBinding('claude-sonnet-5')` returns `{ id }` with no `provider` field. Pre-resolving before `providerForModel` would silently discard a slot-level `chatgpt-oauth`/`openai` override.

Each function receives the input it actually needs. No divergence case was found in which the tier-resolved Claude check passes while the raw-pin routing check wrongly skips coercion.

Also confirmed at the ref: unconfigured `local` resolves to `''` and `isClaudeFamilyModel('')` returns `false` (`responses-config.ts:44-45`), matching the test's non-coercion assertion; `coercedFrom` has exactly one production consumer (`src/agent/subagent.ts:469-471`) and is used only in a stderr diagnostic string, never as a routing id, so reporting the raw tier name there is correct and arguably more useful; `isClaudeFamilyAfterTierResolution` is confined to its defining module (not exported); and all four symbols the new test imports (`CLAUDE_SONNET_ID`, `DEFAULT_SLOT_BINDINGS`, `resetSlotBindings`, `setSlotBindings`) are real exports of `model-slots.ts`, with `resetSlotBindings()` in `afterEach` preventing global slot-binding leakage.

---

### Per-dimension status

- **security** — no issues found; no credential, auth, or secret path is touched by this diff.
- **api-compat** — no issues found; `ChildModelCoercion`, the exported `coerceCrossProviderChildModel` signature, and `isClaudeFamilyModel` are all unchanged. The new helper is private, so no public surface changed.
- **correctness** — no issues found; both call-site swaps traced against `resolveModelInput`, `resolveBinding`, and `providerForModel` at the ref (see above).
- **spec-compliance** — one nit (docstring precision). Both changes the intent names are present; `isClaudeFamilyModel`, `providerForModel`, and downstream resolution are untouched as claimed; no scope creep found.
- **test-coverage** — one low finding (parent-branch tier case).
- **perf-observability** — no issues found; adds one synchronous in-memory `resolveModelInput` call per branch at fork time, no I/O, not a hot loop.

### What was not checked

- Citations were verified inline against branch HEAD `22f656e012d09a23c8026da9e78b00c8fb61b0bd` via `git show`/`git grep` at that ref. All cited lines resolved; none dropped as fabricated. One citation path was corrected during verification (`src/agent/subagent.ts:469-471`, not `subagent/subagent.ts`) — content verified.
- Tests and the type-checker were **not executed**; the CI-run claims in the PR body were not independently re-run. All logic traces are static. CI on this ref is green across Lint & Build, ubuntu, macOS, PTY scrollback, docs, and publish-bundle smoke.
- Stated intent: PR #879 title + body — spec-compliance **was** assessed against it.
- Not reviewed: `nesting.ts` / `compose-executor.ts` fork paths beyond the single `coercedFrom` consumer trace; runtime behaviour under a live `chatgpt-oauth` slot; vitest cross-file parallelism of global slot state beyond confirming the `afterEach` reset.
- Synthesis was performed inline rather than by a separate agent: light regime dispatches a single reviewer, so cross-agent dedup and conflict detection are structurally no-ops. Shadow-verify was not triggered — its gate is a critical or high finding, and none was raised.

Issue #876, opened 2026-08-03
openai-compatible: the Environment block never rebuilds on a mid-session cwd re-anchor


griffinwork40:
## Symptom

On `openai-compatible`, the auto-injected `# Environment` block is built **once** and never re-emitted, so after a mid-session cwd re-anchor the `- Working directory:` line itself keeps naming the launch checkout for the rest of the session.

This is the sibling of #875 (which fixes the `- Workspace:` line on `anthropic-direct`), and it is the reason #875's openai-compatible branch is inert.

## Evidence

- `OpenAICompatibleQuery.setCwd` does exactly one thing — re-base the dispatcher:

  ```ts
  // src/agent/providers/openai-compatible/query.ts:1129-1131
  setCwd(cwd: string): void {
    this.toolDispatcher?.setResolveBase?.(cwd);
  }
  ```

  It never touches `runtimeStateSource` and never rebuilds the system prompt.

- The `# Environment` fragment is assembled once, inside `query()`:
  `formatEnvironmentFragment({ cwd: config.cwd ?? process.cwd(), ..., workspace: runtimeStateSource.getWorkspace() })` — `src/agent/providers/openai-compatible/index.ts:327-335`.

- There is no `cwdDependentsFactory` equivalent on this provider: `assembleSystemPrompt` / `formatEnvironmentFragment` appear nowhere outside that one construction-time call site.

- `provider.query()` is called once per session (or per model swap), not per turn — `ProviderRouter.buildInner` at `src/agent/providers/router/provider-router.ts:261-272`. So "per query" is not "per turn", and `config.cwd` is *not* a live value for the duration of a session.

## Why it matters after #875

Before #875, both `get_runtime_state self.cwd` and the whole `# Environment` block were stale together on this provider — consistently wrong.

After #875, `getCwd()` resolves live (via `process.cwd()`, which the interactive autoname hook chdir'd — `worktree-autoname.ts` `pinProcessCwd`), so `self.cwd` and `- Workspace:` become correct while the `- Working directory:` line stays frozen. The divergence **flips direction** rather than closing — the same class of self-contradicting Environment block that #875 exists to eliminate, now on the sibling provider.

[...]

PR #887, opened 2026-08-03
feat(slash): add /afk-md to edit AFK.md overlays with live hot-reload

griffinwork40:
## Summary

Adds `/afk-md` (aliased `/memory`) — an in-REPL editor for the AFK.md system-prompt overlay. The differentiator is **hot-reload: an edited overlay applies to the RUNNING session**, which Claude Code's `/memory` cannot do (an Anthropic engineer confirmed on anthropics/claude-code#722 that changes are only picked up by new sessions; the long chain of `/reload` requests — #15858, #7379, #17127 — is closed unfixed).

### Command surface

```
/afk-md                 overview of both tiers and what the model receives
/afk-md user|project    edit a tier in $EDITOR, then diff + hot-reload
/afk-md show            print the composed overlay exactly as the model sees it
/afk-md add <text>      append a bullet (--user retargets to personal)
/afk-md reload          re-read from disk for edits made in another terminal
```

### Agent layer

A new optional `ProviderQuery.setSystemPrompt?(basePrompt)` mirroring the existing `setCwd?()` hook, implemented by `AnthropicDirectQuery` via a new `query/overlay-rebuild.ts` factory and forwarded by `ProviderRouter` (which also carries the value onto a fresh inner query, so a `/model` swap does not revert the prompt). `AgentSession.setSystemPrompt()` returns whether the live swap actually happened, so a surface can never claim a reload it did not achieve.

**The load-bearing detail:** the rebuild factory mutates the *shared* `StableSystemParts` in place rather than a copy. `createCwdDependentsFactory` closes over the same object, so a copy-based rebuild would let the next `setCwd()` — a `/cd`, or the first-turn worktree autoname hook — silently resurrect the pre-edit prompt. A copy-based implementation passes a naive "output contains the new text" assertion and still regresses, so there is an explicit test that runs `setSystemPrompt()` then `setCwd()` and asserts the edit survives.

`openai-compatible` deliberately leaves the hook `undefined`: it assembles its system parts as locals inside `query()` rather than retaining them on the instance, so a live swap there needs the #876 staleness rework first. That is what the optional-method convention is for — `/afk-md` reports "saved, applies on next launch" there instead of faking success.

### Why not just call it `/memory`

AFK already has a separate cross-session memory system (HOT.md + the fact archive, written by the agent through `memory_update`). `/memory` is an alias so Claude Code muscle memory works, and the overview carries one line pointing at `memory_update` for the other thing.

### Beyond parity with Claude Code's `/memory`

| Claude Code gap | This PR |
|---|---|
| No hot-reload into a live session (#722, #15858, #7379, #17127) | applies to the running session |
| `#` quick-add removed in v2.0.70; destination ambiguous (#14868, #26145) | `add <text>` / `--user`, always prints the absolute destination |
| Needs a separate `/context`; never shows the text | `show` prints the composed overlay verbatim |
| No feedback after save | unified diff with `+N -M`, capped at 40 lines with elision |
| Opens an empty buffer on create | seeded scaffold naming the tier and its precedence |
| — | per-tier token accounting + a 10k budget warning |
| — | surfaces the loader footgun where a whitespace-only tier is silently treated as absent, and where two tiers dedup to one file |

### Supporting changes

- Extracted the audited TTY suspend/spawn/restore dance out of `editor-open.ts` into `editor-spawn.ts`, so `/editor` and `/afk-md` share one copy of the fd-0 handoff instead of cloning it. That module's header already stated keeping one audited copy was its purpose; `editor.test.ts` passes **unchanged** as the regression proof.
- Added `getUserAfkMdPath()` / `getProjectAfkMdPath()` to `src/paths.ts` and routed `afk-md-tier.ts` through them, per the no-hand-joined-AFK-paths convention.

The reload path busts the AFK.md cache *before* re-reading and re-derives through `resolveBaseSystemPrompt()`, never `loadAfkMd().content` — passing the bare overlay would strip the framework doctrine from the live session. Both the ordering and that composition are pinned by tests.

## How was this tested?

Every gate in `.github/workflows/ci.yml` was run locally against the rebased branch:

| Gate | Result |
|---|---|
| `pnpm lint` | PASS |
| `pnpm build` | PASS |
| `pnpm test` | PASS — 777 files, 14,758 tests, 2 skipped, 0 failures |
| `pnpm test:coverage` | PASS (coverage floors met) |
| `pnpm test:pty` | PASS — 10 tests |
| `pnpm build:dist` (publish-bundle smoke) | PASS |
| `pnpm audit:env:check` | PASS |
| `pnpm scan:env:check` | PASS |
| `pnpm audit:chalk:check` | PASS |
| `pnpm audit:sdk:check` | PASS |
| `pnpm fix:pins:check` | PASS |
| `pnpm audit:deps` | PASS |

Not run locally: the `docs` job (website typecheck + build) — this diff touches nothing under `website/`.

New tests, 21 in total:

- `src/agent/session/set-system-prompt.test.ts` — end-to-end proof through the real provider chain with a stubbed Anthropic client (the T19 `setCwd` harness pattern): turn 1 carries the old overlay, `setSystemPrompt()`, turn 2 carries the new one and still contains the framework base; history is not wiped; **the edit survives a subsequent `setCwd()`**; clearing drops the overlay while keeping the rest of the payload.
- `src/agent/providers/anthropic-direct/query/overlay-rebuild.test.ts` — shared-prefix mutation, live-cwd vs fallback-cwd assembly, cleared-overlay normalization.
- `src/cli/slash/commands/afk-md.test.ts` — reload ordering invariant (cache bust strictly before re-derive, apply strictly after), that the applied string is the full composed prompt and not the bare overlay, honest reporting when the provider cannot swap, `add` targeting/creation/no-blank-line-stacking, non-TTY degradation, unknown subcommand.
- `src/cli/slash/commands/afk-md/targets.test.ts` — all six tier states: neither, only-user, only-project, both, symlink-deduped, whitespace-only.
- `src/cli/slash/commands/editor-spawn.test.ts` — refusal paths never spawn; exit-code mapping; the restore-always invariant holds even when `spawn` throws.

Also smoke-rendered the overview against real AFK.md files to check the output reads well.

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes
- [x] No secrets, tokens, or private paths in the diff — verified by grepping the committed diff for absolute operator paths (no matches); tests use `os.tmpdir()` fixtures

## Note for reviewers

While writing these tests I found a pre-existing repo-wide test footgun worth knowing about, not fixed here to keep this PR scoped: the vitest global setup (`src/__test-utils__/redirect-paths-env.ts`) redirects `AFK_HOME` so no test can write to the real `~/.afk`, but **nothing redirects `process.cwd()`**. Any project-scope path helper (`getProjectAfkMdPath()`, `getProjectSettingsPath()`) therefore resolves to the real repo root during tests. Combined with `vi.mock()` being silently inert on a mistyped relative specifier, a project-scope test can quietly read and append to this repository's own tracked files — which is exactly what happened to me mid-development before I caught and reverted it. A `redirect-cwd` setup file, or a guard asserting project-scope writes land under a temp dir, would close it.

### DIFF
--- a/src/agent/provider.ts
+++ b/src/agent/provider.ts
@@ -382,6 +382,32 @@ export interface ProviderQuery extends AsyncIterable<ProviderEvent> {
    */
   setCwd?(cwd: string): void;
   /**
+   * Optional. Replace the composed base system prompt for all **subsequent**
+   * turns in this query's lifetime, without resetting the session or touching
+   * conversation history.
+   *
+   * Contract: `basePrompt` is the FULLY COMPOSED base prompt — the framework
+   * doctrine (`prompts/system-prompt.md`) plus the `# Operator configuration`
+   * overlay — i.e. exactly the shape of `AgentConfig.systemPrompt`, as produced
+   * by `resolveBaseSystemPrompt()`. It is NOT the bare AFK.md overlay text;
+   * passing only the overlay would silently drop the framework prompt from a
+   * live session.
+   *
+   * Returns `true` when the live prompt was actually rebuilt, `false` when the
+   * provider accepted the call but could not apply it (e.g. no rebuild factory
+   * was wired because the caller owns an external dispatcher). The return value
+   * is the honesty channel for `/afk-md`: a surface must never report a
+   * hot-reload it did not achieve, so a `false` here means "saved to disk,
+   * applies on next launch".
+   *
+   * Implemented by `AnthropicDirectQuery` and forwarded by `ProviderRouter`.
+   * `OpenAICompatibleQuery` deliberately leaves this undefined — it assembles
+   * its system parts as locals inside `query()` rather than retaining them on
+   * the instance, so a live swap there needs the #876 staleness rework first.
+   * `AgentSession.setSystemPrompt()` calls this only when present.
+   */
+  setSystemPrompt?(basePrompt: string | undefined): boolean;
+  /**
    * Optional. Force a fresh SDK client by re-reading whatever credential
    * source the provider uses (e.g. the macOS Keychain for OAuth tokens).
    *
--- a/src/agent/providers/anthropic-direct/index.ts
+++ b/src/agent/providers/anthropic-direct/index.ts
@@ -72,6 +72,7 @@ import { env } from '../../../config/env.js';
 import { resolveQueryToken } from './query/token-resolution.js';
 import { getRuntimeStateTool } from '../../awareness/index.js';
 import { createCwdDependentsFactory } from './query/cwd-dependents.js';
+import { createSystemPromptRebuildFactory } from './query/overlay-rebuild.js';
 import { wireQueryDispatcher } from './query/dispatcher-wiring.js';
 import { setUpQueryClient } from './query/client-setup.js';
 import { assembleQueryPrompt } from './query/prompt-assembly.js';
@@ -556,6 +557,24 @@ export class AnthropicDirectProvider implements ModelProvider {
           buildDispatcher: (mode, opts) => this.buildDispatcher(mode, opts),
         });
 
+    // Invariant: this factory MUST close over the SAME `stableSystemPrefix`
+    // object passed to `createCwdDependentsFactory` above — not a copy. The
+    // overlay rebuild writes the new base prompt back into that shared object
+    // so a later `setCwd()` re-assembly inherits it instead of resurrecting the
+    // launch-time prompt. See the Invariant block in query/overlay-rebuild.ts.
+    // Gated on `externalTools` for the same reason the cwd factory is: those
+    // callers own their own prompt/dispatcher lifecycle.
+    const systemPromptRebuildFactory = this.externalTools
+      ? undefined
+      : createSystemPromptRebuildFactory({
+          stableSystemPrefix,
+          config,
+          surface: this.surface,
+          runtimeStateSource,
+          getCurrentCwd: () => this._currentCwd,
+          fallbackCwd: cwd,
+        });
+
     const resolvedEffort = resolveEffort(config.effort, model);
     return new AnthropicDirectQuery({
       client,
@@ -609,6 +628,7 @@ export class AnthropicDirectProvider implements ModelProvider {
         ? { maxToolUseIterations: config.maxToolUseIterations }
         : {}),
       ...(cwdDependentsFactory !== undefined ? { cwdDependentsFactory } : {}),
+      ...(systemPromptRebuildFactory !== undefined ? { systemPromptRebuildFactory } : {}),
       // Path-approval half of the live `/bypass` toggle: keep the provider's
       // `_currentPermissionMode` (read by getGrants().allowAll) in sync with
       // the query handle's mode. The file-tool half is the dispatcher's
--- a/src/agent/providers/anthropic-direct/query.ts
+++ b/src/agent/providers/anthropic-direct/query.ts
@@ -180,6 +180,17 @@ export interface AnthropicDirectQueryOptions {
    */
   cwdDependentsFactory?: (cwd: string) => { userSystem: string; dispatcher: ToolDispatcher };
   /**
+   * Factory for re-assembling `userSystem` around a NEW composed base prompt
+   * when `setSystemPrompt()` is called mid-session (the `/afk-md` hot-reload
+   * path). When absent, `setSystemPrompt()` reports `false` and changes
+   * nothing — same degradation contract as `cwdDependentsFactory`.
+   *
+   * Supplied by `AnthropicDirectProvider.query()` as a closure over the same
+   * shared `stableSystemPrefix` the cwd factory uses, so the two rebuild paths
+   * compose rather than overwrite each other. See `query/overlay-rebuild.ts`.
+   */
+  systemPromptRebuildFactory?: (basePrompt: string | undefined) => string;
+  /**
    * Provider callback invoked by `setPermissionMode()` to update the
    * provider-level `_currentPermissionMode` — the field the path-approval hook
    * reads via the provider's `getGrants().allowAll`. This is the path-approval
@@ -283,6 +294,7 @@ export class AnthropicDirectQuery implements ProviderQuery {
    */
   private readonly retry: RetryLayer;
   private readonly cwdDependentsFactory?: (cwd: string) => { userSystem: string; dispatcher: ToolDispatcher };
+  private readonly systemPromptRebuildFactory?: (basePrompt: string | undefined) => string;
   private readonly onPermissionMode?: (mode: string) => void;
   private readonly mcpManager?: import('../../mcp/index.js').McpManager;
   private readonly hookRegistry?: HookRegistry;
@@ -303,6 +315,7 @@ export class AnthropicDirectQuery implements ProviderQuery {
     this.traceWriter = opts.traceWriter;
     if (opts.subagentId !== undefined) this.subagentId = opts.subagentId;
     this.cwdDependentsFactory = opts.cwdDependentsFactory;
+    this.systemPromptRebuildFactory = opts.systemPromptRebuildFactory;
     this.onPermissionMode = opts.onPermissionMode;
     this.mcpManager = opts.mcpManager;
     if (opts.hookRegistry !== undefined) this.hookRegistry = opts.hookRegistry;
@@ -733,6 +746,28 @@ export class AnthropicDirectQuery implements ProviderQuery {
     const { userSystem, dispatcher } = this.cwdDependentsFactory(cwd);
     this.state.userSystem = userSystem;
     this.state.toolDispatcher = dispatcher;
+  }
+
+  /**
+   * Swap the composed base system prompt for all subsequent turns.
+   *
+   * Contract: `basePrompt` is the FULL composed prompt (framework doctrine +
+   * `# Operator configuration` overlay), not the bare overlay — see the
+   * `ProviderQuery.setSystemPrompt` doc. Returns `true` when the live prompt
+   * was rebuilt, `false` when no rebuild factory is wired (external-dispatcher
+   * callers), so the caller can report honestly instead of claiming a reload
+   * that never happened.
+   *
+   * Safe between turns and idempotent: `composeSystem()` re-reads
+   * `state.userSystem` fresh on every turn (it already mutates it per-turn for
+   * date rollover), so writing a newly assembled string here is the same
+   * established mechanism `setCwd()` uses — no session reset, no history loss.
+   * Synchronous for the same reason `setCwd` is: it is called between turns.
+   */
+  setSystemPrompt(basePrompt: string | undefined): boolean {
+    if (!this.systemPromptRebuildFactory) return false;
+    this.state.userSystem = this.systemPromptRebuildFactory(basePrompt);
+    return true;
   }
 
   async supportedCommands(): Promise<ProviderCommandInfo[]> {
--- a/src/agent/providers/anthropic-direct/query/overlay-rebuild.test.ts
+++ b/src/agent/providers/anthropic-direct/query/overlay-rebuild.test.ts
@@ -1,0 +1,101 @@
+/**
+ * Tests for the live system-prompt rebuild factory.
+ *
+ * The load-bearing property is the SHARED-PREFIX mutation: the factory writes the
+ * new base prompt back into the same `StableSystemParts` object the cwd-rebuild
+ * factory closes over, so a later `setCwd()` inherits the operator's edit instead
+ * of resurrecting the launch-time prompt. A copy-then-assemble implementation
+ * would pass a naive "does the output contain the new text" check and still
+ * silently undo the hot-reload on the next `/cd`.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { createSystemPromptRebuildFactory } from './overlay-rebuild.js';
+import { assembleSystemPrompt, buildStableSystemPrefix } from './system-prompt.js';
+import type { AgentConfig } from '../../../types/config-types.js';
+import type { RuntimeStateSource } from '../../../awareness/index.js';
+
+function makeParts(userSystem: string | null): ReturnType<typeof buildStableSystemPrefix> {
+  return buildStableSystemPrefix({
+    toolBase: 'TOOL-BASE',
+    memoryPrompt: 'MEMORY',
+    hotMemory: '',
+    manifest: '',
+    userSystem,
+  });
+}
+
+const runtimeStateSource = {
+  getWorkspace: () => undefined,
+} as unknown as RuntimeStateSource;
+
+function makeFactory(
+  parts: ReturnType<typeof buildStableSystemPrefix>,
+  cwd: string | undefined,
+  fallbackCwd = '/fallback',
+) {
+  return createSystemPromptRebuildFactory({
+    stableSystemPrefix: parts,
+    config: { sessionId: 's1' } as AgentConfig,
+    surface: 'cli',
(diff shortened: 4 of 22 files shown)


griffinwork40:
@codex fix comments

griffinwork40 (commented):

## Merge decision: **MERGE**

No blocking findings. The three load-bearing claims in the PR body were independently verified against the branch HEAD, not taken on trust. Four `nit`/`low` items below are all follow-up material — none gates this merge.

**Reviewed:** `7649e7331322a2c1158e4b1c42e67bd5bf379fd5` · feature · 22 files · +2044 / −109 · regime `full` (2 dimension agents + inline citation/absence verification)

---

### Verified claims (the ones a naive review would rubber-stamp)

| PR claim | Verdict | Evidence |
|---|---|---|
| The rebuild factory mutates the **shared** `StableSystemParts` in place, so a later `setCwd()` cannot resurrect the pre-edit prompt | **holds** | `query/overlay-rebuild.ts:60-61` assigns `args.stableSystemPrefix.userSystem` directly; `anthropic-direct/index.ts:495` destructures one `stableSystemPrefix` and passes that same reference to both `createCwdDependentsFactory` (:545) and `createSystemPromptRebuildFactory` (:570) |
| Reload busts the cache **before** re-reading and re-derives through `resolveBaseSystemPrompt()`, never the bare `loadAfkMd().content` | **holds** | `afk-md/reload.ts`: :74-75 bust → :79 `resolveBaseSystemPrompt(cwd)` → :84 compose → :91 apply, return value captured into `applied` |
| `openai-compatible` leaves the hook undefined and the surface degrades honestly rather than faking success | **holds** | `provider.ts:409` declares `setSystemPrompt?(…)` optional; `agent-session.ts:1028` and `provider-router.ts:489` both use `?.` + `?? false` |
| The regression test runs `setSystemPrompt()` **then** `setCwd()` and asserts the edit survives | **holds** | `set-system-prompt.test.ts:159-180` asserts `toContain('NEW-OVERLAY-MARKER')`, `not.toContain('OLD-OVERLAY-MARKER')`, and `toContain('/moved/worktree')` after the `setCwd()` — it genuinely pins the regression a copy-based impl would pass naively but fail here |
| Reload-ordering, full-composed-prompt, and honest-reporting tests exist and assert what is claimed | **holds** | `afk-md.test.ts:175-185` (strict `toBeGreaterThan` ordering), `:187-199` (incl. `not.toBe('OVERLAY-TEXT')`), `:202-212` (both `applied` branches with mutually exclusive text) |
| `editor-spawn.ts` restore-always invariant holds when `spawn` throws | **holds** | `editor-spawn.ts:126-147` — synchronous `spawn` in its own try/catch inside a promise executor, whole await inside `try … finally { restoreInput() }`; pinned by `editor-spawn.test.ts:161-175` |

`$EDITOR` handling is **not** a command-injection surface: `editor-spawn.ts:130` is `spawn(editor.cmd, [...editor.args, filePath], { stdio: 'inherit' })` — argv array, no `shell`, `filePath` never interpolated. `EDITOR='vi; curl evil'` fails closed to `spawn-failed` (ENOENT on a binary named `vi;`) with the TTY restored. No traversal surface either: `getUserAfkMdPath()` takes zero args (`paths.ts:429-431`) and `add <text>` writes `text` into file *content* only (`edit.ts:170`), never into a path.

---

### Findings (4, none blocking)

**1 · low · correctness · `src/cli/config/afk-md-tier.ts:72` + `src/cli/slash/commands/afk-md/targets.ts:44`**
`isSameFile` exists as two private, byte-identical 8-line copies, neither exported. Editing one without the other makes the `/afk-md` overview report a tier's loaded/duplicate state opposite to what the model actually receives — the overview silently lies about the prompt.
```ts
function isSameFile(a: string, b: string): boolean {
  if (a === b) return true;
  if (!existsSync(a) || !existsSync(b)) return false;
  try { return realpathSync(a) === realpathSync(b); } catch { return false; }
}
```
*Fix:* export `isSameFile` from `afk-md-tier.ts` and import it in `targets.ts` (whose import block at :13-16 currently has no such import).

**2 · low · test-coverage · `src/cli/slash/commands/afk-md.ts:75-79`**
The 10k-token budget warning has no asserting test — grep for `OVERLAY_TOKEN_WARN_THRESHOLD|10_000|10000|trimming` across `afk-md.test.ts` and `targets.test.ts` returns zero matches, so the one operator-facing guardrail against an oversized overlay is unpinned.
```ts
if (totalTokens > OVERLAY_TOKEN_WARN_THRESHOLD) {
  ctx.out.warn(`That overlay is ${formatTokens(totalTokens)} tokens — …`);
```
*Fix:* seed a >10k-token overlay fixture and assert the warn text fires. (The threshold itself compares the right value — `currentOverlayTokens` measures the composed multi-tier content, not one tier.)

**3 · nit · correctness · `src/cli/slash/commands/afk-md.ts:129,160`**
`add` tokenizes with `trimmed.split(/\s+/)` and rejoins with `' '`, so a newline inside the bullet text is silently collapsed to a space rather than rejected or preserved. Behavior is self-consistent with the one-bullet-one-line invariant and does not corrupt the file; it is simply undocumented and untested.
*Fix:* add a case feeding `add line one\nline two` asserting the written line is single-line, or document the collapse.

**4 · nit · security · `src/cli/slash/commands/editor-spawn.ts:116-126`**
Defensive asymmetry only: `process.stdin.pause()`/`resume()` are each individually try/caught (:119, :124) but `compositor.suspendInput()` (:123) and `compositor.resumeInput()` (:120) are not.
> Severity floored to `nit` during verification. The "terminal left in raw mode" failure mode **does not hold**: `terminal-compositor.lifecycle.ts:113-127` restores raw mode under its own `try/catch` and re-registers the keypress listener *before* `repaint()`, the only call that can realistically throw; `caretBlinkController.stop()/start()` are pure timer state. The TTY is already restored before any throw is reachable.

*Fix (optional):* wrap the two compositor calls in the same best-effort pattern for symmetry.

---

### Spec-compliance

Stated intent: PR #887 title + body — **assessed.** No unmet intent and no scope creep found. Every checkable claim (five-subcommand surface, `/memory` alias, per-tier token accounting + 10k warning, seeded scaffold on create, absolute destination printed, `show` printing the composed overlay, routing through `getUserAfkMdPath()`/`getProjectAfkMdPath()`, the `editor-spawn.ts` extraction) is present in the code. The extraction is a genuine net reduction: `editor-open.ts` goes **176 → 119 lines (−57)**.

One documentation-accuracy note: the body says **21 new tests**; the actual count is **56** `it()` blocks across the five files (4 / 6 / 24 / 11 / 11). The figure appears to count narrative clauses rather than cases, and omits whole describe blocks (`/afk-md registration`, `/afk-md overview`, `/afk-md show`, `renderTargetRows`, `formatTokens`, `resolveEditor`). The suite is *more* thorough than advertised — worth correcting the body, not the code.

### What was not checked

- Citations verified inline against branch HEAD `7649e733`, materialized via `git archive` — the operator's local checkout was on a different branch, so every `file-state` read was taken against the reviewed ref, not the working tree.
- **Tests were not executed.** The PR body's local gate table (lint/build/777 files/14,758 tests/pty/coverage) is taken as the author's report, not independently reproduced. CI on this head is green across Lint & Build, Test (ubuntu/macos), PTY, Docs, and Publish Bundle Smoke.
- Not reviewed: the Telegram and daemon surfaces (`/afk-md` is registered only in the interactive REPL bootstrap, `src/cli/commands/interactive/bootstrap.ts:19`, so those surfaces cannot reach it); `TerminalCompositor` internals beyond the `suspendInput`/`resumeInput` pair read for finding 4; the `website/` docs job (this diff touches nothing there).
- No file-size finding was emitted: this repo has no size gate, and no file crossed 350 LOC because of this diff.
- The author's own disclosed footgun — nothing redirects `process.cwd()` under vitest, so project-scope path helpers resolve to the real repo root — is a **pre-existing repo-wide test-isolation issue, correctly scoped out of this PR**, not a defect introduced here. It deserves its own issue (a `redirect-cwd` setup file or a guard asserting project-scope writes land under a temp dir).

*Read-only review: no files, branches, or PR metadata were modified.*

griffinwork40 (commented):

One `high` finding: the new write path follows symlinks, breaking this repo's own SEC-4 `O_NOFOLLOW` convention. Everything else is clean — including, verified independently, the invariant this PR names as load-bearing.

Reviewed ref: `7649e733`. Regime: full (22 files, +2044/−109). Change type: feature.

---

### high · confidence high · security · `src/cli/slash/commands/afk-md/edit.ts:166-170`

**finding:** `appendToTarget` writes through `target.path` with no symlink guard, so a repo-committed `AFK.md` symlinked outside the checkout redirects the write to the link target — violating the `O_NOFOLLOW` convention this repo already applies to exactly this write class.

```js
const existing = readOrEmpty(target.path);
const separator = existing.length === 0 || existing.endsWith('\n') ? '' : '\n';
writeFileSync(target.path, `${existing}${separator}- ${text}\n`, 'utf-8');
```

The same gap exists in `ensureExists` (`edit.ts:45-50`) for the **dangling-symlink** case: `existsSync` follows the link and returns `false`, so `writeFileSync` creates a new file at the resolved path outside the project.

Two corrections to what a first pass suggests, both established by re-derivation — they are why this is `high` and not `critical`:
- **Not destructive.** `appendToTarget` is read-then-append, so an existing target's text round-trips intact. For a symlink onto an existing file `ensureExists` early-returns (`existsSync` follows the link), so the scaffold write never fires.
- **No attacker payload.** An attacker controls only the *destination*; the *content* is the victim's own typed text (`afk-md.ts:160`) or their `$EDITOR` session. No injection path.

Residual data-loss surface is narrow but real: `readFileSync(path,'utf-8')` at `edit.ts:166` lossily decodes a non-UTF-8 target to U+FFFD and writes the mangled string back, so a symlink onto a binary file corrupts it irreversibly.

The convention already exists in-tree and is simply not applied here — `src/cli/input/history.ts:298,349,364` and `src/agent/worktree-sweep-valve.ts:129-140`, whose comment describes this precise threat ("a checkout... can carry a symlink... OUTSIDE the repo... `O_NOFOLLOW` makes that open fail with ELOOP").

**suggestion:** Open both writes with `O_WRONLY | O_CREAT | O_NOFOLLOW` (append flag for `appendToTarget`) and let `ELOOP` fail closed, mirroring `worktree-sweep-valve.ts:140`.

Note: the symlink-following **read** path in `src/cli/config/afk-md-tier.ts` is pre-existing and not introduced by this diff (`afk-md-tier.ts` is +7/−8, cache-key and cwd-param only). `/afk-md show` does newly echo that content to the terminal.

---

### low · confidence high · test-coverage · `src/agent/providers/router/provider-router.ts:486-489`

**finding:** The `/model`-swap-survival behavior has no router-level test — `provider-router.test.ts` never references `setSystemPrompt`, so the forwarding and the `buildInner()` re-application at line 279 are pinned only by source reading.

```js
setSystemPrompt(basePrompt: string | undefined): boolean {
  this.currentSystemPrompt = basePrompt; this.systemPromptOverridden = true;
  return this.active?.query.setSystemPrompt?.(basePrompt) ?? false;
```

Verified absent: zero matches for `setSystemPrompt` in `src/agent/providers/router/*.test.ts`. The behavior itself is correct.

**suggestion:** Add one `provider-router.test.ts` case that calls `setSystemPrompt()`, forces an inner-query rebuild, and asserts `innerConfig.systemPrompt` still carries the override.

---

### nit · confidence medium · api-compat · `src/agent/routing-directive.ts:143` / `src/agent/providers/anthropic-direct/query/system-prompt.ts:88`

**finding:** Two exported `assembleSystemPrompt` functions with incompatible signatures — `(base, autoRouting, surface)` vs `(parts, cwd, identity)` — now both live in this feature's import graph (`reload.ts:17` takes the former, `overlay-rebuild.ts:69` the latter). TypeScript rejects a mixed-up call today, so this is naming hygiene only.

**suggestion:** Rename the routing-directive one to `assembleRoutingPrompt`.

---

### Verified clean

**The load-bearing invariant holds.** The PR stakes correctness on the rebuild factory mutating the *shared* `StableSystemParts` in place, so a later `setCwd()` cannot resurrect the pre-edit prompt. An independent verifier re-derived all seven sub-claims as CONFIRMED:

- One object is constructed at `anthropic-direct/index.ts:495` and flows by reference into **both** `createCwdDependentsFactory` (`index.ts:544-545`) and `createSystemPromptRebuildFactory` (`index.ts:569-570`) — no spread at either call site; `buildStableSystemPrefix` is a pass-through (`system-prompt.ts:49-51`).
- `overlay-rebuild.ts:60-61` assigns the property directly, not into a copy.
- `cwd-dependents.ts:100` reads `args.stableSystemPrefix` live at call time, so no frozen-snapshot path exists in either order.
- `set-system-prompt.test.ts:159-180` is a genuine end-to-end assertion: `setSystemPrompt()` → `setCwd()` → `sendMessage`, then asserts the real outbound `messages.create` payload contains the new marker, excludes the old, and carries the new cwd. This is the test that a copy-based implementation would fail — it is not a weaker proxy.
- `AgentSession.setSystemPrompt()`'s boolean is honest: `OpenAICompatibleQuery` genuinely lacks the method (zero grep matches under `src/agent/providers/openai-compatible/`), so `?? false` collapses to `false` (`agent-session.ts:1028`). `this.config` updating regardless is documented intent (`agent-session.ts:1016-1020`), not a false success signal.

**Reload ordering** is correct and pinned: `reload.ts:73-91` busts before re-deriving and applies last; `afk-md.test.ts:175-185` asserts `bust < compose < apply` by call-log index. `reload.ts:84` composes through `assembleSystemPrompt(basePrompt, ...)` from `resolveBaseSystemPrompt()`, never bare `loadAfkMd().content`, so framework doctrine is not stripped.

**api-compat:** additive only. `ProviderQuery.setSystemPrompt?` is optional (`provider.ts:409`), so the non-implementing provider still satisfies the interface; `loadConfig`/`loadAfkMd` gained a defaulted trailing param, preserving existing call sites; the `editor-spawn.ts` extraction re-exports `resolveEditor`/`EditorNotifyKind` verbatim and keeps `restoreInput` in a `finally` (`editor-spawn.ts:113-147`), pinned by `editor-spawn.test.ts:161-175`.

**Command injection:** not present. `spawn(editor.cmd, [...editor.args, filePath], { stdio: 'inherit' })` (`editor-spawn.ts:130`) is argv-array with no shell, so a crafted `$EDITOR` cannot inject metacharacters.

**spec-compliance:** assessed against PR title+body. Every row of the "Beyond parity" table is implemented, the five subcommands match the documented surface, and no scope creep beyond the two declared supporting changes.

---

### What was not checked

- Citations were verified inline against branch HEAD `7649e733` (full post-merge tree materialized via `git archive`). Four cited line numbers from the first pass were out of range for their files; each one's quoted content was located elsewhere in the same file and the citations above are corrected to real lines. No fabricated evidence survived into this report.
- **Two findings were dropped in verification, not reported:** a "`afk-md.test.ts` exceeds the 350-LOC ceiling" claim (no LOC gate exists in this repo — no script in `package.json`, no CI step; the premise is unverifiable here) and a "'21 new tests' is miscounted" claim (56 `it()` blocks exist; the body enumerates per-file prose, so nothing is falsified).
- Tests were **not executed** — the PR reports all CI gates passing locally, and GitHub checks are green (8 success), but this review did not re-run them.
- Not reviewed: the Telegram surface, the `website/` docs job, and runtime behavior of the TTY fd-0 handoff under a real `$EDITOR`.
- The symlink finding's exploit precondition (a victim running `/afk-md` inside a checkout carrying a malicious `AFK.md` symlink) was reasoned from source, not reproduced at runtime.

PR #888, opened 2026-08-03
feat(cli): reacting goblin mini-sprite in a constant right-aligned band (#336)

### LINKED ISSUE #336: Reacting goblin mini-sprite in the extraRows band (Option B)
## Summary

Give the AFK goblin mascot a **visible, reacting presence during a live session** — a compact 2–3 row goblin sprite that animates while the agent works (idle → working → alert), rendered in the reserved `extraRows` footer band above the status line.

This is the "go wild" follow-on to the **goblin spinner (v1)** already on branch `afk/polish-goblin-banner` (commit `07b21d9`), which themes the existing 1-line spinner olive + goblin verbs. v1 is deliberately single-line; this issue tracks the bigger swing.

## Why it's a separate effort (not a spinner reskin)

The spinner row carries a **hardcoded single-row budget** in the compositor's `fixedRows` accounting (`src/cli/terminal-compositor.frame.ts`). Any spinner output containing a newline undercounts physical rows and corrupts the DECSTBM scroll-region math. So a multi-row mascot **cannot ride the spinner** — it needs its own painter in a multi-row surface.

## Design sketch

- **Render home:** the `extraRows` band — reserved footer rows above the status line, already resize-self-healing (`setExtraRows(n)`, scroll-drag handled generically).
- **State:** reuse `MascotState = 'idle' | 'working' | 'alert'` (already defined in `src/cli/mascot.ts`, but only `idle` renders today — `void state`, ~line 197). Needs **compact 2–3 row sprites** for `working` (and optionally `alert`), drawn with the existing half-block (▀▄) pixel-grid renderer for cross-terminal consistency (do **not** use emoji / geometric Unicode).
- **Triggers:** the same ProviderEvent stream the spinner already uses — `working` at `tool_use` start, `idle` at `tool_result` (`src/cli/_lib/stream-renderer-orchestrator.ts`). `alert` needs a small addition: the spinner disable path currently ignores `is_error`, so there's no error hook yet.

## Constraints / risks (verified during recon)

- **The `extraRows` band is contended** — already shared by the input reader (`reader.ts` does `setExtraRows(priorExtraRows + 1)`), the loop-stage rail, and the background-task bar (`src/cli/status-line.ts`). A mascot is another co-tenant and must coordinate its row reservation.
- **`afterScrollRestore` is a single-slot callback** (`src/cli/status-line.ts:82`, set ~:242, invoked ~:318). If the background-status-bar already owns it, it likely needs widening to an array for a second occupant.
- **Heavily-tested compositor:** anchorRow / scrollback-eviction / DECSTBM math has extensive regression tests. Blast radius is **medium–high** — changes must keep that suite green.
- **Hero sprite is too big to reuse directly:** the splash goblin is `MASCOT_WIDTH=27` × `MASCOT_HEIGHT=13`; the live form must be a new, much smaller sprite.

## Effort & optionality

- **Effort:** medium–high (new band painter + row-reservation coordination + compact sprite art + `alert` wiring).
- **Optional:** must respect the same opt-out philosophy as v1 (`AFK_GOBLIN_SPINNER=0`, or a dedicated flag). Decide whether it **replaces** the v1 spinner theme or **stacks** with it.

## Acceptance criteria

- [ ] A compact (≤3 row) goblin sprite renders in the `extraRows` band during tool execution and clears cleanly at idle.
- [ ] No regression in the terminal-compositor suite (anchorRow / scrollback / band coordination).
- [ ] Survives `NO_COLOR`, non-truecolor terminals, `AFK_BANNER_PLAIN`, and resize (SIGWINCH).
- [ ] Gated by an opt-out env flag; default behavior documented.
- [ ] (Stretch) `alert` state on tool error.

## References

- v1 (goblin spinner): branch `afk/polish-goblin-banner`, commit `07b21d9`
- Key files: `src/cli/mascot.ts`, `src/cli/input/spinner.ts`, `src/cli/status-line.ts`, `src/cli/terminal-compositor.frame.ts`, `src/cli/_lib/stream-renderer-orchestrator.ts`

griffinwork40:
Closes #336.

A 13×3 goblin head in a constant band at the footer's right edge — resting while the agent thinks, twitching while it runs tools, red-eyed on a tool error. Default off, behind `AFK_GOBLIN_MASCOT=1`.

> **This PR went through three shapes.** The first (`adc65c09`) painted the 13×3 sprite into a 3-row reservation that *appeared* when the agent picked up a tool and collapsed at idle — the transcript jumped twice per tool call — and it sat in the left gutter, directly in the reading path. The second (`7fbe5aac`) gave up the rows entirely and rode the loop-stage rail as a 7×1 decoration; that fixed the geometry, but two pixel rows cannot carry a cap, a hatband and a face, so the goblin stopped being a goblin. This is the third: the art from the first, the position from the second.

## What made the first band obtrusive was never the row count — it was that the count moved

So this band never moves.

- **Constant, not transient.** `MascotBand` publishes its reservation once in `start()` and releases it once in `stop()`. In between, the count changes only when `visibleRows()` crosses its headroom floor, which takes a geometry change — a resize, or a co-tenant below it growing. Sprite content *cannot* reach `setExtraRows`: `visibleRows()` reads only `stream.rows`, `stream.columns` and `getAdjacentRows()`, never `getLines()`. Pinned by a test that animates 30 frames and asserts exactly **one** publish.
- **Right-aligned, not gutter-left.** Content lands at `columns - 13`, so the sprite ends at `columns - 1` and the final column stays untouched — the DECAWM lesson from the second shape, carried over. Writing the last cell arms the pending-wrap flag and the next write scrolls the reserved band. Below 14 columns the band is dropped **whole**: never truncated, never wrapped.
- **One direction each.** `LiveMascot` owns the state machine, the alert dwell and the frame clock; `MascotBand` owns the reservation and the paint. The mascot asks the band to redraw, the band asks the mascot for a frame — so a frame change can never become a geometry change. That split is the lesson of the first two attempts.

## The sprite

`mascot-mini.ts` is a 13×6-pixel grid rendered through the banner goblin's own `renderHalfBlockGrid`, so both sprites degrade through exactly one fallback ladder. Six pixel rows is what the art needs: cap tip over cap body (the cone), gold hatband over brow, eye band over grin.

One non-obvious rule is enforced there: **state changes must alter transparency, not only hue.** A half-block cell renders `▀` whenever its top pixel is opaque, whatever the colours — so frames differing only in colour are glyph-identical, and at `chalk.level === 0` (NO_COLOR, CI, piped output) the animation would freeze into a still silhouette. The blink drops the eye pixels (`▀` → `▄`) and `alert` flares the ears wider; colour does the expressive work in a truecolor terminal, and those two levers keep motion *visible* when colour is gone. A test renders every frame at level 0 and fails if a state is glyph-static.

**Cadence.** 300ms, and `working` is rest-dominant: eight frames, five of them the resting face, every twitch followed by rest — a ~2.4s loop carrying three beats. The real spinner is already on screen a row or two above; a second fast-cycling thing beside it reads as noise rather than as a creature. Both the 8-frame length and the 300ms default are pinned by tests, so flipping back to a faster cackle is a deliberate edit, not an accident.

**Present at rest.** `idle` is a single frame and runs no ticker, so a resting REPL pays no ticks and a companion that sits still is less obtrusive than one that pops in and out.

## Width is load-bearing, so it is a runtime invariant

`MascotBand` places the sprite at `columns - MINI_MASCOT_WIDTH` and trusts that arithmetic instead of measuring the rendered line. A pixel row wider than the constant would therefore write the terminal's final column, arm DECAWM's pending wrap, and let the next write scroll the reserved band out from under the reservation. `renderHalfBlockGrid` only checks that a row *pair* is self-consistent (`top.length === bot.length`), which passes any two equally-wrong rows — so `grid()`, the single funnel every frame is built through, now throws unless every row is exactly `MINI_MASCOT_WIDTH`. Frames are module-scope constants, so it throws at import and costs nothing per paint. Previously this was pinned only by tests, and a test does not run on the machine that renders the sprite.

## `alert`

Rides the existing `onStageChange` channel. `StageSignals` is an additive optional second argument, and the callback also fires on an errored tool result **even when the stage did not change** — a failure inside a parallel tool wave leaves other tools pending, so the stage stays `acting` and the mascot would otherwise never learn about the error. Consumers are idempotent repaints, so the extra fire is free. `alert` holds for 1.5s and then falls back to whatever the live stage says, read at expiry rather than captured when the dwell began.

## Default: off

Gated by `AFK_GOBLIN_MASCOT=1`. `MascotBand.start()` returns before it marks itself started on any of three gates — flag unset, `AFK_PLAIN_OUTPUT`/`--plain`, `AFK_BANNER_PLAIN=1` — so with the flag unset `mascotRowCount` stays `0`, the `extraRows` sum is byte-for-byte the pre-#336 `loopStageRows + bgBarRowCount + ledgerRowCount`, no rows are claimed, and no timer exists. `LiveMascot` needs no TTY gate of its own: it never writes, and its host band is TTY-gated.

This claims three terminal rows when you say yes. What it does not do is claim them *intermittently* — saying yes costs three rows once, rather than a transcript that jumps per tool call. The flip is a one-liner in `detectGoblinMascot` (`capture-mode.ts`) whenever you have looked at it live.

## Verification

89 unit tests across the four affected files (18 sprite, 23 band, 22 mascot, 26 rail). The band suite pins the parts that would corrupt the compositor rather than the art:

- constant reservation across 30 animated repaints — exactly one publish
- reserve-before-paint on growth, erase-before-release on teardown
- erase-stale-geometry-on-move
- right-edge placement at 100 / 80 / 62 / 14 columns; drop-not-truncate below that
- collapse-and-reclaim on resize; ResizeBus unsubscribed symmetrically on `stop()`
- co-tenancy against a real `LoopStageBar` — disjoint AND contiguous for bg = 0/2/5 × ledger = 0/1

`LiveMascot`'s suite covers the three inert gates, the ticker lifecycle (no timer at rest, released on `stop()` so it cannot outlive its repaint target), the 300ms default interval, the alert dwell and its fallback to the live stage, and frame-reset on state change.

Gates green: `pnpm lint`, `pnpm test` (774 files / 14 746 tests, 2 skipped), `pnpm build`, and `audit:env:check`, `scan:env:check`, `audit:chalk:check`, `audit:sdk:check`.

## Acceptance criteria (#336)

- [x] Compact (≤3 row) sprite renders during tool execution and clears cleanly at idle
- [x] No regression in the terminal-compositor suite
- [x] Survives NO_COLOR, non-truecolor, `AFK_BANNER_PLAIN`, and resize (SIGWINCH)
- [x] Gated by an env flag; default behaviour documented in `docs/env-registry.md`
- [x] (Stretch) `alert` state on tool error

## Not done, deliberately

- **`BackgroundStatusBar`'s clamp nets out only the rows below it** (`getAdjacentRows: () => ledgerRowCount`), so a large background-job count can overrun the shared reservation and collapse DECSTBM entirely — ordinary transcript output would then scroll through every footer row. Pre-existing, but an enabled mascot narrows that margin by `MASCOT_BAND_ROWS`. **The obvious one-line fix is wrong:** `background-status-bar.ts` feeds `getAdjacentRows()` into both the clamp (`:240`) and the paint address (`:258`, `:279`), so widening it with rows *above* the bar moves the bar into the band and makes `clearRows()` erase the wrong rows. A correct fix separates the clamp budget from the paint offset, in a file this branch does not otherwise touch. Filed rather than smuggled in here.
- **`clearBand()` bounds-checks against a freshly-read `stream.rows`**, so a rows change with no intervening repaint before `stop()` can skip an out-of-range stale row. Same symptom class as `BackgroundStatusBar`'s resize-then-stop path, though not the same mechanism. Low severity — a PTY shrink generally reflows the off-viewport rows anyway.
- **`alert` does not fire for errors that never produce a `tool_result`** — an aborted turn, a stream error, a provider failure. The stage freezes at `acting` and the goblin keeps twitching until the next transition. The stated scope was tool errors; widening the signal is a separate change.
- **No frame-diffing.** Two of the eight `working` ticks repaint a byte-identical band. Immaterial for a default-off 3-row decoration, and adding a frame-identity comparison to a class documented as deliberately geometry-blind costs more than it buys.
- **The v1 goblin spinner is untouched** — this stacks with it rather than replacing it. Both are independently switchable, so that stays your call.


griffinwork40:
Note for review: the first CI run showed `Test (macos-latest)` red on `src/agent/background-registry.test.ts > persists label/hash from the plain prompt and never image base64`. That is a pre-existing flake, not this PR:

- the test waits on a bare `await new Promise(r => setTimeout(r, 100))` and then does a synchronous `fs.readFileSync` of `meta.json` (background-registry.test.ts:737-739), so a slow runner loses the race and fails ENOENT
- `Test (ubuntu-latest)` passed on the identical commit
- this PR touches zero files under `src/agent/` (it is CLI footer painting only), and the test references none of the changed symbols
- it passes 60/60 locally and went green on a re-run of the same commit

Worth a separate fix (poll for the file instead of sleeping 100ms) rather than anything on this branch.

Issue #894, opened 2026-08-03
refactor(cli): split suggest.ts (544 LOC) — and extract stripGhostControlChars out of the render path


griffinwork40:
## Context

Flagged while reviewing #892. `src/cli/input/suggest.ts` is **544 lines**, over the repo's 350-line-per-file norm. It was already 539 before #892 (which added 5), so this is pre-existing debt, not a regression — #892 deliberately left it alone to keep that PR's scope honest.

Being upfront: **`suggest.ts` is not the worst offender.** If the goal is "get files under the ceiling," several nearby files are larger:

| file | LOC |
|---|---|
| `src/cli/terminal-compositor.input-dispatch.ts` | 1144 |
| `src/cli/terminal-compositor.ts` | 1115 |
| `src/cli/input/reader.ts` | 900 |
| `src/cli/terminal-compositor.committed-band-commit.ts` | 841 |
| `src/cli/input/input-surface.ts` | 662 |
| `src/cli/input/suggest.ts` | **544** |

`suggest.ts` is worth doing first anyway, for a reason unrelated to line count — see the layering smell below.

## The standalone win: a layering smell

`stripGhostControlChars` is exported from `suggest.ts` and imported by two compositor files that want nothing else from it:

- `src/cli/terminal-compositor.render.ts:14`
- `src/cli/terminal-compositor.autocomplete.ts:20`

So the render path pulls the entire suggestion engine — including its provider-resolution imports (`providerForModel`, `resolveProvider`, `redact-secrets`) — into its module graph to reach a ~10-line pure string sanitizer. Extracting that one function is a small change that is worth doing **on its own merits**, independent of the line count.

## Proposed split

Two phases, because they carry very different risk. Half the candidate extractions are top-level functions (pure moves); the other half are **closures inside `createSuggestEngine`** and need state threading, which is a real refactor.

### Phase 1 — pure moves, low risk (~150 LOC out)

All of these are already top-level in the file:

| new module | moves | current lines |
|---|---|---|
| `suggest-sanitize.ts` | `stripGhostControlChars` + its contract docblock | 128–157 |
| `suggest-prompt.ts` | `buildSystem`, `buildUser` (incl. the `redactSecrets` egress chokepoint) | 92–126 |
| `suggest-model.ts` | `pickModel` | 524–544 |
| (stay, or fold into the above) | `isValidContinuation`, the four tuning constants | 34–52, 159–169 |

[...]

PR #895, opened 2026-08-03
feat(cli): AI-generated ghost-text suggestion at an empty prompt

griffinwork40:
## What this adds

When a turn ends and the prompt is handed back empty, ask the suggestion model for one short next action and offer it as dim ghost text. **Tab** or **Right-arrow** accepts it into the input; **Enter** still submits as a separate, deliberate step. Off by default behind `AFK_SUGGEST_PROMPT`.

This is the feature Claude Code ships as "prompt suggestions".

## Why it needed a new source rather than a relaxed guard

The Tier-1/Tier-2 engine in `suggest.ts` refuses to fire at an empty buffer on purpose — those tiers *complete a prefix*, and at length 0 there is no prefix to complete. Two independent guards enforce it (`suggest.ts` Tier-1 early return, and the `MIN_LLM_CHARS` floor), and naively removing one dead-ends at the other.

So this adds a distinct source that **proposes a whole next action** from session context. It is the only path permitted to produce a ghost at length 0, and it does not weaken either completion guard.

## Trigger: turn-end, not an idle timer

The suggestion is primed from `InputSurface.readLine`'s per-turn prompt handoff, fire-and-forget. Two consequences worth noting:

- Nothing is awaited, so it can never delay handing the prompt to the user.
- The result is only surfaced if the buffer is **still empty** when it resolves, mirroring the existing Tier-2 stale-guard — a user who starts typing immediately never sees a late ghost appear under their text.

No new timers and no idle detection were added.

### …but the first prompt of a session is not a turn end

Priming on the prompt handoff means the session's *first* `readLine` primes too — the startup prompt is a handoff like any other. At that point `stats.turns` is empty, so `buildPromptSuggestionUser` omits the `what just happened:` section entirely and the model is left with two inputs: the cwd basename, and the history ring — which is loaded from disk and therefore holds the **previous** session's commands. The result was a confident-looking proposal about work the user had not started, seeded from a different session.

So `generatePromptSuggestion` is gated on a non-empty transcript tail (`hasSuggestionGrounding`). The refusal happens **before** the provider call, so the startup prompt costs nothing at all — not a discarded reply, no request.

The condition is grounding, not a first-run flag, which makes the neighbouring cases fall out correctly:

| Prompt | Suggests? | Why |
|---|---|---|
| Session startup | no | nothing has happened yet |
| After any completed turn | yes | that turn is the grounding |
| Right after `/clear` | no | `resetStats` empties the same `stats.turns` |
| First prompt of `--resume` | yes | `stats.turns` is restored — something *did* just happen |

It also keeps the code honest about its own system prompt, which claims the proposal is based "on what just happened in their session".

## Acceptance needed no new keybindings

`handleTab` already falls through to `applyGhostAccept()` when no dropdown is open, and Right-arrow already accepts at end-of-buffer. Both work unchanged at an empty buffer because every string `startsWith('')` and the cursor is trivially at end-of-buffer. The existing prefix-extension guarantee that makes acceptance safe still holds.

**Enter is untouched.** The empty-buffer guard in `input-dispatch.ts` fires before any submit path, so a suggestion can never be auto-executed. That is the exact failure mode Claude Code shipped in v2.0.76 and had to fix (`anthropics/claude-code` issues #16166, #25787); here it is structurally impossible rather than defended against.

## Seams worth a reviewer's attention

- **`updateGhost` ordering.** It owns `activeGhost` and nulls it after a Tier-1 miss, so the primed suggestion is read *before* that assignment — otherwise it would be wiped on every edit landing back at an empty buffer.
- **Validate before scrubbing.** The control-char scrubber deletes newlines, so scrubbing first spliced a two-line reply into one plausible-looking line and defeated the multi-line rejection. A test caught this; there is now a named regression case for it.
- **`peekPromptSuggestion` is optional-called.** Test files are excluded from `tsc` (see tsconfig `exclude`), so a hand-rolled `SuggestEngine` mock predating this method type-checks fine and would have thrown on the input hot path on every backspace-to-empty. There is a test for that tolerance.
- **No caching, deliberately.** The answer is a function of session state that changes every turn; a buffer-keyed cache (whose key would always be `''`) would pin one stale proposal for the whole session. A test asserts it does not poison the Tier-2 cache either.
- **The grounding gate changed what some existing tests were proving.** The sanitization-ordering cases ran against a context with no transcript, so after the gate they would have gone on passing for the wrong reason — null from the gate rather than from the sanitizer. They now carry a transcript tail so they still exercise the code they name.

## Cost and safety

- Reuses the existing `pickModel`, which already routes Anthropic sessions to a cheap suggestion-class model, plus the existing provider memoization, abort controller, and timeout budget.
- At most one call per turn: only when the prompt is empty, only when enabled, and only once the session has a completed turn to reason from.
- Prompt assembly goes through the same `redactSecrets` egress chokepoint as Tier-2; a test asserts an API key in the transcript tail does not leave the process.
- Output is gated before display: single line, length-capped, refusal-filtered, control-chars scrubbed.

## Verification

- `pnpm test` — **15,077 passing** across 791 files (2 pre-existing skips)
- `pnpm lint` (`tsc --noEmit`) and `pnpm build` — clean
- All five CI audit gates pass: `audit:sdk:check`, `audit:env:check`, `scan:env:check`, `audit:chalk:check`, `fix:pins:check`
- **58 new tests** across the new modules, the engine, and the compositor seam — 56 net-new `it()` cases (0 removed) plus 2 rows added to the existing validator table

## Review follow-ups

- **Merged `main`** (not rebased — keeps the push fast-forward). `main` and this branch each added an env var and each bumped the generated `docs/env-registry.md` header count. Git merged the two rows cleanly but the identical header edits were not a textual conflict, so the merged doc undercounted and `scan:env:check` failed on the auto-merge ref while the branch was green in isolation. Regenerated with `pnpm scan:env` — now **146 vars**, header and rows agree.
- **U+2028/U+2029 + bidi hardening.** The single-line guard tested only `/[\r\n]/` and the scrubber covered only C0/C1, so `"fix the bug\u2028rm -rf /tmp/x"` passed validation and reached the terminal unmodified. `isValidPromptSuggestion` now rejects U+2028/U+2029 alongside CR/LF, and `stripGhostControlChars` strips `[\u2028\u2029\u202a-\u202e\u2066-\u2069]` — the bidi override/isolate block, which could otherwise reorder the visible run so displayed ghost text differs from the bytes Tab accepts (Trojan Source, CVE-2021-42574). Directional marks (U+200E/U+200F/U+061C) are deliberately left alone: no override scope, and legitimate in RTL text.
- The three Codex P2 comments (priming on elicitation prompts, stale-suggestion resurrection, older-prime overwrite) were raised against `6151530a` and were already resolved in `728d1b4c`; each is pinned by a named regression test.

## Not addressed

The 350-line norm is **no longer outstanding for the two files originally flagged** — the split landed in this PR rather than being deferred:

| file | before | after |
|---|---|---|
| `src/cli/input/suggest.ts` | 544 | **119** |
| `src/cli/terminal-compositor.autocomplete.ts` | 304 | **212** |

Nine new modules carry the extracted concerns (`suggest-types` / `-tier1` / `-tier2` / `-provider` / `-sanitize` / `-prompt` / `-prompt-state`, plus `terminal-compositor.ghost.ts`), every one under 260 lines. That covers the substance of #894, including its "extract `stripGhostControlChars` out of the render path" clause — now its own module at 51 lines.

No file this PR touches is *newly* over 350. Five were already over on `main` and grow marginally here: `input-surface.ts` 662→688, `terminal-compositor.ts` 1115→1137, `input-dispatch.ts` 1144→1153, `surface-setup.ts` 410→418, `loop-iteration.ts` 796→801. Splitting those is out of scope.


griffinwork40:
# PR #895 Review

**Target:** `feat(cli): AI-generated ghost-text suggestion at an empty prompt`  
**Reviewed ref:** `5ef66e6cfa9cfb5a4cbb699ce24dda46957f28e0`  
**Regime:** Full — 13 files, +886/−3  
**Recommendation:** **DO NOT MERGE**

## Findings

### 1. High · Overlapping prompt requests can publish stale results and escape teardown cancellation

**Dimension:** Correctness, lifecycle, stated-intent compliance  
**Location:** `src/cli/input/suggest.ts:528-545`  
**Confidence:** High — independently shadow-verified

`primePromptSuggestion()` assigns the result after awaiting without checking whether that invocation is still current. If turn A’s request starts, turn B’s request starts, and A resolves last, A overwrites B’s newer suggestion:

```ts
async function primePromptSuggestion(ctx: SuggestContext): Promise<void> {
  promptSuggestion = await generatePromptSuggestion(ctx, {
```

Controller ownership has the same race:

```ts
onController: (c) => {
  if (c !== null || promptController !== null) promptController = c;
},
```

Every request calls `onController(null)` from `generatePromptSuggestion()`’s `finally` block at `src/cli/input/suggest-prompt.ts:236-238`. That callback checks only nullity, not controller identity. An older request settling can therefore clear the slot containing a newer request’s controller. `dispose()` at `src/cli/input/suggest.ts:556-561` then has nothing to abort, leaving the newer provider call alive after teardown.

This is reachable because `InputSurface.readLine()` starts priming fire-and-forget at each prompt handoff, with no serialization or in-flight guard. It defeats the stated stale-safety contract.

**Suggested fix:** Abort the prior prompt request when superseding it and associate each invocation with a generation or controller identity; only the current invocation should publish its result or clear the controller slot.

### 2. Medium · Prompt-suggestion provider failures are completely silent

**Dimension:** Observability  
**Location:** `src/cli/input/suggest-prompt.ts:234-235`  
**Confidence:** High

The new generation path suppresses all provider, authentication, model-selection, and network errors:

```ts
} catch {
  return null;
}
```

The existing Tier-2 path reports the same class of failures through `opts.onError`, wired to `debugLog('[afk suggest] Tier-2 completion failed:', err)` in `surface-setup.ts`. With the new path, a bad `AFK_SUGGEST_MODEL`, invalid credentials, or unreachable compatible endpoint looks identical to “the model declined to suggest anything,” even under `AFK_DEBUG=1`.

**Suggested fix:** Add an error callback to `PromptSuggestionDeps` and connect it to the existing suggestion-engine `onError` sink.

### 3. Medium · The change violates the project’s hard 350-line source-file ceiling

**Dimension:** Architecture and maintainability  
**Locations:** `src/cli/input/suggest.ts:619`, `src/cli/terminal-compositor.autocomplete.ts:354`  
**Confidence:** High

The resulting files are 619 and 354 lines. The repository’s operator-level rule is a hard ceiling, not a discretionary norm: new behavior must be extracted as a whole concern rather than expanding an oversized type. Tracking the split in follow-up issue #894 does not satisfy that merge constraint.

**Suggested fix:** Extract the empty-prompt engine state/interface adapter from `suggest.ts` and one complete autocomplete concern from `terminal-compositor.autocomplete.ts` before merging.

### 4. Low · No test covers overlapping primes or teardown during the second request

**Dimension:** Test coverage  
**Location:** `src/cli/input/suggest.test.ts:554-713`  
**Confidence:** High

The prompt-suggestion tests invoke `primePromptSuggestion()` once per test. Repository search found no test with two overlapping prime calls, reversed completion order, or `dispose()` while a superseding call remains pending. Consequently, the race in finding 1 passes the current suite.

**Suggested fix:** Add deferred-promise tests proving that the newest invocation wins when the older request resolves last and that `dispose()` aborts the currently owned controller.

## Confirmed clean areas

- Secret redaction covers the assembled transcript/history/cwd payload before provider egress.
- Raw multiline output is rejected before control-character scrubbing can splice lines together.
- Startup and post-`/clear` prompts are gated by a non-empty transcript tail before any provider request.
- `AFK_SUGGEST_PROMPT` remains dependent on `AFK_SUGGEST_ENABLED`, matching the documented opt-in contract.
- Empty-prompt suggestions remain separate from Tier-1/Tier-2 prefix completion.
- Ghost acceptance remains explicit through Tab or Right-arrow; Enter does not promote ghost text into the input buffer.
- Production code constructs the engine through `createSuggestEngine()`; no additional production `SuggestEngine` object-literal implementers were found.

## Verification notes

- Citations were checked inline against branch HEAD `5ef66e6cfa9cfb5a4cbb699ce24dda46957f28e0`.
- The high-severity concurrency finding was independently re-derived across the generator, upstream `readLine()` caller, downstream ghost consumer, and teardown path.
- Stated intent came from PR #895’s title and body; spec compliance was assessed.
- Tests, lint, and build were not executed as part of this read-only review; the PR author’s reported verification was not independently reproduced.
- The initial broad five-agent review wave timed out; the review was rerun as a bounded two-agent wave, followed by inline citation checks and independent verification of the blocking claim.

## Done

- Reviewed PR #895 read-only and found one merge-blocking concurrency defect, two medium issues, and one focused test gap.
- Evidence is anchored to ref `5ef66e6cfa9cfb5a4cbb699ce24dda46957f28e0` and diff artifact `/Users/griffinlong/.afk/state/skill-preflight/94c79755-d933-4f7f-ad88-2878ba740998/pr-895.diff`.
- No repository files, commits, PR metadata, or external systems were modified.
- The PR remains pending fixes and independent test execution.

---

griffinwork40 (commented):

## Merge decision: **DO NOT MERGE** — one blocking CI failure, mechanically reproduced below

The blocker is **not a defect in your commits** — the head ref passes `scan:env:check` in isolation. It is a semantic merge conflict with `main`. Everything else in this change verified clean, including all four safety invariants the description claims.

Reviewed ref: `728d1b4c` · 23 files, +2064/−622 · regime: full

---

### 🔴 high · correctness/CI · `docs/env-registry.md:5` · semantic merge conflict with `main`

`Lint & Build` fails at `pnpm scan:env:check`. Root cause, reproduced:

`main` and this branch each added **one** env var and each independently bumped the generated header `**143 vars**` → `**144 vars**`. Git merges the two var *rows* cleanly (both survive), but the two identical header edits are not a textual conflict — so the merged doc claims 144 while the merged registry defines **145**.

Verified by computing the merge (`git merge-tree origin/main 728d1b4c`) and running the real generator against the merged tree:

```
head ref alone:  ✓ scan:env: docs/env-registry.{json,md} in sync (144 vars)
merged tree:     ✗ scan:env: docs/env-registry would drift from src/config/env.ts.
                   - docs/env-registry.md
```

Merged tree: header says `144`, table has `145` rows, `src/config/env.ts` defines `145` (`AFK_GOBLIN_MASCOT` from main + `AFK_SUGGEST_PROMPT` from here). `--check` is a full string-equality diff against a fresh render whose header embeds `sorted.length`, so the stale count line **alone** is sufficient to fail it — no row-level drift is required. `.json` is unaffected (it carries no count). CI checks out the auto-merge ref (`on: pull_request` + bare `actions/checkout@v4`), which is why the branch is green locally and red in CI.

**Fix:** rebase on `main`, run `pnpm scan:env`, commit. Any PR adding an env var while another is in flight hits this — the count line is a merge-conflict magnet that git cannot see.

---

### 🟡 medium · security · `src/cli/input/suggest-sanitize.ts:41` + `src/cli/input/suggest-prompt.ts:130`

The single-line guard tests only `/[\r\n]/`, and `stripGhostControlChars` strips only `\u0000-\u001f` / `\u007f-\u009f`. **U+2028/U+2029** (Unicode line separators) and the bidi-override block (U+202A–U+202E, U+2066–U+2069) fall outside both. A model reply containing U+2028 passes validation and reaches the terminal unmodified:

```js
"fix the bug\u2028rm -rf /tmp/x"
  /[\r\n]/.test(...)  → false      // multi-line guard does not fire
  stripGhostControlChars(...)      → unchanged, length 25
```

This is exactly the failure mode the "validate before scrubbing" ordering was built to prevent — a second class of line terminator routes around it. Many terminals render U+2028 as a line break, and the string is accepted verbatim into the input buffer on Tab. Not reachable without a hostile/confused model response, hence medium, not high.

**Fix:** add `\u2028\u2029` to the multi-line rejection test in `isValidPromptSuggestion`, and strip `[\u2028\u2029\u202a-\u202e\u2066-\u2069]` in `stripGhostControlChars`.

---

### 🔵 low · docs · PR description "Not addressed" is stale

The description says `suggest.ts` goes 545 → 619 and `terminal-compositor.autocomplete.ts` 304 → 354, both over the 350-line norm, deferred to #894. The ref says otherwise — both got **smaller**:

```
src/cli/input/suggest.ts:                  544 → 119
src/cli/terminal-compositor.autocomplete.ts: 304 → 212
```

The split already landed (`suggest-types/-tier1/-tier2/-provider/-sanitize/-prompt/-prompt-state`, `terminal-compositor.ghost.ts` — 9 new modules, every one under 260 lines). This undersells the change; worth updating before merge so #894 is not left describing work that is done. No production file this PR touches is newly over 350; the five that are (`input-surface.ts` 662→688, `terminal-compositor.ts` 1115→1137, `input-dispatch.ts` 1144→1153, `loop-iteration.ts` 786→791, `surface-setup.ts` 404→412) were already over and grew marginally. No mechanical size gate exists in `package.json`/`scripts/`, so the norm is advisory — not a blocker.

---

### ✅ Verified clean — every load-bearing claim in the description held

Each was independently re-derived at the ref, and the two highest-stakes ones went through a second adversarial verifier pass:

- **No auto-execute** (the Claude Code v2.0.76 failure mode) — **confirmed structurally.** `handleEnter` returns at `input-dispatch.ts:684` on an empty buffer before any submit path. Ghost text lives in `activeGhost`, never mirrored into `input.buffer`; `applyGhostAccept` is the sole writer and has exactly two callers — Tab (`:1127`) and Right-arrow (`:1034`). No timer, no Enter-path call.
- **`updateGhost` ordering** — **holds.** The peek + early return (`ghost.ts:142-145`) precedes the `activeGhost = null` at `:151`, so a primed suggestion is not wiped on backspace-to-empty. Enforced by control flow, not comment discipline.
- **Stale-guard** — **holds.** The resolve handler re-reads the live `self.input.buffer` (`ghost.ts:51-54`), not a captured snapshot.
- **Zero-cost startup** — **holds.** `hasSuggestionGrounding` refuses at `suggest-prompt.ts:205`, textually before `resolveComplete`/`complete` at `:216`/`:220`. No request at the startup prompt or after `/clear`.
- **One call per turn** — **holds.** `primePromptGhost` has one non-test call site (`input-surface.ts:507`, from `loop-iteration.ts:327`), once per `readLine`. Backspace-to-empty does not re-prime; `suggest-prompt-state.ts` newest-wins is defense-in-depth for cross-turn overlap.
- **Redaction** — egress goes through `redactSecrets` at `suggest-prompt.ts:91`; asserted by test at `suggest-prompt.test.ts:69`.
- **Scrub coverage** — the suggestion is scrubbed three times independently: generation (`suggest.ts:86`), render (`render.ts:134`), accept (`ghost.ts:202`). No unscrubbed display path.
- **"52 new tests"** — **accurate**: 51 net-new `it()` cases added by the diff, 0 removed (`suggest-prompt.test.ts` +21, `suggest.test.ts` +18, `ghost.test.ts` +11, `input-surface.test.ts` +1). All five named regression seams exist as real, correctly-scoped cases — notably `'rejects a multi-line reply even though the scrubber would remove newlines'` (`suggest.test.ts:740`) and `'is a no-op at the startup prompt'` (`:588`), which asserts `completeFn` was **not called** rather than merely that the result was null.

---

### What was not checked

- Citations verified inline against branch HEAD `728d1b4c` (via `git show` at the ref + a `git archive` snapshot); the CI claim was verified against the computed merge tree `a4177595`.
- Stated intent: PR #895 title + body — spec-compliance **was** assessed. No unmet intent or scope creep found; the diff implements what the description says, and the description understates the refactor.
- Did not execute the test suite, `tsc`, or the REPL — no runtime verification of terminal rendering behavior. The U+2028 finding is from direct execution of the two guard expressions, not from a live terminal.
- Did not review the Telegram surface or the docs site build.
- Wave 1's security/api-compat agent hit its tool-iteration cap; its uncompleted portions were re-derived by two narrower agents plus the inline checks above.

**Recommendation:** rebase + `pnpm scan:env` clears the blocker. The U+2028 guard is a small, well-scoped follow-up worth folding in while you are here. The underlying design — a separate source at length 0 rather than a relaxed completion guard, gated on grounding, with accept as a deliberate second keystroke — is sound and the structural argument in the description checks out.

PR #900, opened 2026-08-04
refactor(cli): return the goblin mini-sprite to the transient left band (#336)

griffinwork40:
## What

Restores the first iteration of #888 (`adc65c0`): the 13x3 goblin mini-sprite lives in a **transient, left-aligned** `extraRows` band — it pops up while the agent is mid-tool and releases its rows at idle. The merged version's constant right-aligned band (`mascot-live` + `mascot-band`) is removed; `MascotBar` returns with its reserve-before-paint / erase-before-release ordering, alert dwell, and co-tenancy with the bg bar and verdict rail.

## Why

After living with all three iterations (transient left band → one-row rail → constant right-aligned band), the first reads best:

- The sprite **appearing** is itself the activity signal — a constant band makes you check the animation to know whether anything is happening.
- A resting REPL pays no rows and no timer ticks.
- Left gutter placement matches the loop-stage rail instead of competing with the right-aligned status content.

## Mechanics

Reverse-applied the delta `adc65c0..557d6fc` (iterations 2–4) onto current main. The only main-side overlap was the unrelated umber theme, which is kept. Gating is unchanged: `AFK_GOBLIN_MASCOT=1`, opt-in, default off.

## Verification

- `pnpm lint`, `pnpm build` ✅
- `pnpm scan:env:check`, `pnpm audit:env:check` ✅
- Mascot suites: 58 tests ✅
- Full `src/cli/commands/interactive/` + `src/cli/_lib/` suites: 1,424 tests ✅
- `pnpm test:pty`: 10 tests ✅


griffinwork40:
All verification complete. Here is my synthesized review.

---

# PR #900 Review: refactor(cli): return the goblin mini-sprite to the transient left band (#336)

**+857 / -1448 · 17 files** · refactor merges `MascotBand` + `LiveMascot` into one `MascotBar`, switches from a constant session-long reservation to a transient one (appears only while tools run, disappears at rest), and moves the sprite from right-aligned to left-aligned with a 2-space gutter.

## ✅ API Compatibility — CLEAN

- Zero dangling references to deleted exports (`MascotBand`, `LiveMascot`, `liveMascot`, `getLiveMascot`, etc.)
- All new imports (`MascotBar`, `MASCOT_BAR_ROWS`, `getMascotBar`, `mascotBar`) wired correctly at every call site
- `pnpm lint` (tsc --noEmit): **PASS**
- `scan:env:check`: **PASS** (env registry in sync)

## ✅ Tests — PASS (41 tests, lint clean)

## ✅ Timer/Event-Loop Safety — CLEAN

`.unref()` on both `setInterval` and `setTimeout`, `stop()` clears both, timer gated on `state != idle && frameCount > 1`. No holds on the event loop.

---

## 🔴 CRITICAL (blocking): Missing `loopStageBar?.redraw()` in bg bar handler — stale loop-stage bar

**File:** `footer-subsystems.ts:133-141`

The `bgStatusBar.setRowCountChangeHandler` updates `bgBarRowCount`, calls `syncExtraRows()`, and redraws `mascotBar` — but does NOT call `loopStageBar?.redraw()`. **The old code called both** `mascotBand?.redraw()` AND `loopStageBar?.redraw()`.

The loop-stage bar positions from `getExtraRows()` = `1 + mascotRowCount + bgBarRowCount + ledgerRowCount`. When the bg bar grows/shrinks, the loop-stage bar is displaced — but nobody repaints it. The mascot's own `setRowCountChangeHandler` (line 151) does call `loopStageBar?.redraw()`, but that only fires when the mascot's **row count** changes. In `mascot-bar.ts:247`, `onRowCountChange` only fires when `desired !== this.rowCount`. In the common case (mascot idle at 0 rows, or working at 3 rows with sufficient headroom), a bg-bar size change shifts the mascot's *position* but not its *count*. The loop-stage bar is left stale at its old row until the next stage transition or resize.

**Reproduction:** `AFK_GOBLIN_MASCOT=1`, agent working (mascot visible). A background subagent starts → bg bar grows 0→1 rows → **loop-stage bar painted at wrong row** (ghost row at old position, gap at new one).

**Fix:** Add `loopStageBar?.redraw();` back to the bg bar handler, matching the pattern in the verdict handler (line 124).

---

## 🟡 MEDIUM: DECAWM boundary on 15-column terminals

**File:** `mascot-bar.ts:276`

The left-aligned `GUTTER + lines[i]` (2 + 13 = 15 chars from column 1) reaches the terminal's last column on a 15-column terminal, arming the DECAWM pending-wrap flag. The `visibleRows()` width gate uses strict `<` (`columns < 15`), so exactly 15 passes. The save/restore (`\x1b[s`/`\x1b[u`) does not reset per-row pending-wrap state. The old code had `RIGHT_MARGIN=1` precisely to prevent this. No test covers the 15-column boundary.

**Fix:** Require at least 16 columns (`columns < MINI_MASCOT_WIDTH + GUTTER.length + 1`) or add a 1-column right-margin guard to the paint.

---

## 🟡 MEDIUM: Transcript jumping — intentional regression, needs acceptance

The new transient reservation reverses the old design's **explicit core goal**: eliminating per-tool-call DECSTBM changes. The deleted `mascot-band.ts` header states the first (transient) shape "made the transcript jump twice per tool call." The new `MascotBar` reintroduces this: `setState('working')` claims 3 rows → `syncExtraRows()` → DECSTBM grows; `setState('idle')` releases → DECSTBM shrinks. Two scroll-region changes per tool call — exactly what the old code eliminated.

The PR title confirms this is intentional. **Trade-off:** resting REPL gets 3 rows back ↔ 2 DECSTBM changes per tool call. No compensating mechanism is present.

---

## 🔵 LOW-INFO: Test coverage gaps (accidentally dropped)

1. **No content verification** — The new `cupRows()` helper only extracts row numbers from CUP sequences. No test verifies rendered sprite content reaches the stream. A bug emitting CUP+erase but no content would pass all tests. *(Old test: "pulls the frame fresh on every repaint".)*

2. **No frame-difference verification** — The animation test checks CUPs happen after advancing timers, but never asserts frames differ. A stuck frame counter (always painting frame 0) would pass. *(Old test: "cycles the sprite through more than one rendered frame while working".)*

3. **No DEFAULT_FRAME_MS pin** — The old test pinned the production default at 300ms. The new default is 220ms but no test pins it. A regression to 50ms or 5000ms fails nothing.

Several intentionally-dropped tests (right-alignment, constant-reservation invariant, rest-dominant rhythm) are design changes and need no replacement.

---

## ✅ (`grid()` width-guard removal): MITIGATED

The runtime row-width throw was removed, but `mascot-mini.test.ts` still mechanically pins `row.length === MINI_MASCOT_WIDTH` for every frame of every state at both the pixel-row and rendered levels. CI catches violations before merge. The runtime guard was belt-and-suspenders; test-time coverage is adequate.

---

## Merge Recommendation: REQUEST CHANGES

| Priority | Item | Action |
|---|---|---|
| **Blocking** | `loopStageBar?.redraw()` missing in bg bar handler | Add one line — this causes a visible stale-bar bug |
| **Acceptance needed** | Transcript-jumping trade-off | Confirm the operator accepts DECSTBM churn per tool call |
| Recommended (non-blocking) | Content verification test (H1) + DEFAULT_FRAME_MS pin (H3) | Prevents silent content/frequency regressions |
| Consider (non-blocking) | DECAWM 15-column boundary guard | Easy fix; tightest-term edge case |

The code merge (two classes → one) is a sound simplification now that the constant-reservation invariant no longer applies. Timer safety is clean. API surface is fully migrated with no dangling references. The PR's primary blocker is the missing cascade-redraw — a one-line fix that prevents a real user-facing ghost-bar bug.

**Relevant:**
- **Done** — Review complete. Evidence: this report + diff artifact (`pr-900.diff`), lint/test confirmation (`pnpm lint` + 41 tests pass, `scan:env:check` passes), worktree at `restore/goblin-mini-left-popup @ 315d30a` (clean — no files modified, review is read-only).
- **Pending**: The one blocking fix (`loopStageBar?.redraw()` at `footer-subsystems.ts:141`) is documented but not applied — review is read-only.

---

PR #901, opened 2026-08-04
fix(improve): make the pipeline's own signals honest + split improve.ts under the 350 ceiling

griffinwork40:
## Why

I ran the full `afk improve` pipeline end-to-end against live state to see how it was doing. **All five stages execute cleanly** — scan (536 sessions in 1.7s), cards, propose, eval-gen, eval-run (7/7 checks, 2ms). Nothing errored.

What was broken was what they *reported*. Four independent correctness bugs, each of which made the pipeline quieter or more misleading than reality.

Two commits: correctness fixes, then a pure-motion file split.

---

### 1. `closure-anomaly` promoted to default-on

It sat opt-in behind a docstring claiming `iteration_cap` / `hook_blocked` / `max_turns_exceeded` emission was *"not yet wired"*. That was **already false when written** — `hook_blocked` and `max_turns_exceeded` landed in the same commit as the caveat (`78c40833`), `iteration_cap` three days later (`1ba29e7e`), and `timeout` closures began appearing 2026-07-07.

What the stale comment cost: scanning all 12,766 local witness traces turns up **296 `timeout`/`iteration_cap` closures worth $990** that no default scan would ever have surfaced. On this machine the detector produces 3 of the 6 highest-severity cards.

`abort` is severity-capped downstream (`severityFor`: low by default, medium at ≥3), so the default noise floor stays quiet. **`subagent-block` and `subagent-read-denial` stay opt-in** — their rationale (fire during legitimate worktree confinement / active bug fixing) is current, not stale.

### 2. `eval-run`: `unsupported` now exits 3, not 0

`unsupported` is what the runner returns when an eval-case's pattern has **no registered contract** — i.e. when the run asserted *nothing at all*. It shared exit 0 with a genuine `pass`.

Wiring this into CI as `eval-run <case> && deploy` would have silently green-lit every pattern lacking a contract (today: `subagent-read-denial`) exactly as if its guardrail had been verified.

`3` rather than `1` so a gate can tolerate "no contract yet" without also tolerating a regression. `2` stays reserved for CLI argument errors.

### 3. `propose` / `eval-gen` refuse to stack duplicates without `--force`

Neither checked for an existing artifact before writing, and `getProposalsForCard` had been sitting exported with **zero call sites**. Live result on this machine:

| card | proposals |
|---|---|
| `tool-failure-compose` | 4 |
| `tool-failure-browser-open` | 4 (on a *resolved* card) |
| `tool-failure-config-get` | 2 |
| `tool-failure-ask-question` | 2 |

Near-identical, minutes apart. Preview (`--no-write`) persists nothing and is never gated; `eval-gen` also lets an explicit `--evidence-row` through, since a different row is a genuinely different fixture.

### 4. `eval-gen` stopped emitting a disclaimer that was false

Every generated eval-case said *"No runner consumes it yet. A future sprint will replay the fixture through the detector."* That runner **shipped** — `afk improve eval-run` replays the fixture through the live guardrail. The artifact now names the command that runs it.

---

## Commit 2 — `refactor(improve): split improve.ts 1087 LOC → 7 files`

Pure motion, no logic change. The file was 3x over this repo's 350 LOC ceiling and commit 1 had just added 76 more lines to it.

```
index.ts     92   scan.ts    229   cards.ts   202   propose.ts 172
eval-gen.ts 277   eval-run.ts 134  shared.ts    7
```

A helper used by exactly one file stayed in that file, so `shared.ts` holds only what is genuinely shared.

**Pure-motion proof:** stripping imports, comments and blanks from the old file and from the concatenated new files yields **830 logic lines on each side**. The complete diff between those two sets is 7 lines — the sub-registrars gaining the `export` keyword they now need to cross a file boundary.

---

## Verification

- `pnpm lint`, `pnpm build` — clean
- `pnpm test` — **777 files, 14,815 passed**, 2 skipped
- All five audit gates — `audit:sdk` / `audit:env` / `scan:env` / `audit:chalk` / `fix:pins` — PASS
- **Help equivalence:** `improve --help` plus all 7 subcommand `--help` outputs are **byte-identical** to the pre-split build. This is the real guarantee that the refactor moved no user-visible text.
- Each fix exercised live against real state, before *and* after the split: default scan reports `closure-anomaly: 3` without `--include-disabled`; duplicate `propose`/`eval-gen` exit 1 while `--no-write` still exits 0; `eval-run` pass still exits 0.

Four tests pinned the old behavior and were updated (not deleted) — plus a new `improve.eval-run-exit.test.ts` covering the exit-code contract, including that `pass`/`fail`/`unsupported` occupy three distinct codes and that `unsupported` never collides with the arg-error code.

---

## Deliberately NOT in scope

- **No `apply` / `promote` stage.** The pipeline still dead-ends at `eval-run` and never feeds a change back into the repo. The funnel is `59 cards → 17 proposals → 9 eval-cases → 9 eval-runs → ∅`. That is the real gap, and it needs design rather than a drive-by.
- **No `proposals triage` / `eval-cases triage`.** 100% of proposals and eval-cases are stuck at `draft` and no command can transition them — the schema defines `draft|approved|rejected|superseded` but only *cards* have `triage`. I chose not to add one: while no apply stage exists, approving a proposal only moves a label. Honest `draft` beats theatrical `approved`.
- **No root-cause for the subagent cap burn.** The $990 finding above is real and the rate is rising (per-dispatch cap rate ~2.9% → 7.8% over the last two weeks, on ~8x dispatch volume), but two caveats keep me from calling it a clean regression here: the metric is only 4 weeks old, and closure instrumentation coverage climbed 28% → 84% over the same period. Diagnosing why `/simplify` and `/review` dimension agents run 60–100+ sequential bash/grep calls is a separate investigation.

Full review notes: `~/.afk/workspace/reports/improve-pipeline-review-2026-08-03.md`.

### Note for reviewers

Touches `src/improve/eval-gen/writer.ts`, which open PR #893 also modifies (atomic-write helper dedup). Different region of the file — one is a markdown template string, the other an import/write call-site — so it should merge cleanly, but worth sequencing awareness.


griffinwork40:
@codex fix comments

griffinwork40:
## PR #901 Review

**Verdict: APPROVE — no actionable correctness defects found.**

The refactor preserves the deleted `improve.ts` command wiring and import relationships, while the functional changes appear deliberate:

- `unsupported` eval runs now produce exit code `3`, distinct from pass (`0`), regression/error (`1`), and argument errors (`2`), with direct regression coverage in `src/cli/commands/improve.eval-run-exit.test.ts`.
- The new duplicate guards for `propose` and `eval-gen` intentionally refuse redundant persisted artifacts while preserving `--force`, `--no-write`, and explicit `--evidence-row` escape paths.
- Promoting `closure-anomaly` to default-on is explicitly documented and pinned by registry tests.
- The extracted modules remain below the 350-LOC ceiling, avoid circular imports through `shared.ts`, and preserve the public `VALID_PATTERNS` export.

### Non-blocking coverage opportunities

- Add command-level tests for duplicate refusal and the `--force`/`--no-write`/`--evidence-row` bypasses in `src/cli/commands/improve/propose.ts:46-64` and `src/cli/commands/improve/eval-gen.ts:55-82`.
- Strengthen `src/improve/eval-gen/writer.test.ts` to assert that the rendered command includes the actual `evalCaseId`, not merely `afk improve eval-run`.
- An integration test could exercise each `eval-run` output branch to complement the pure `applyEvalRunExit` unit tests.

The apparent fall-through after `process.exit(1)` is not a production defect: Node’s declared and runtime contract is non-returning. It would matter only in tests that replace `process.exit` with an incorrectly returning mock.

One adversarial review worker timed out, while two independent reviewers completed diff-wide passes; neither produced a verified introduced defect. Tests were not run because the review was constrained to the supplied unmerged diff and the current dirty working tree was explicitly read-only.

## Done

- Reviewed PR #901 for correctness, compatibility, lifecycle behavior, test coverage, path integrity, and semantic parity with the deleted monolith.
- Evidence is the supplied 2,461-line artifact at `/Users/griffinlong/.afk/state/skill-preflight/0be014b8-3b3c-45f6-bbc4-025b27fd3e51/pr-901.diff`.
- The working tree was not modified, stashed, committed, checked out, or reset.
- Full runtime tests remain deferred because PR files are not present in the current checkout and mutating the dirty working tree was prohibited.

---

PR #903, opened 2026-08-04
fix(cli): remove wrap-boundary leading spaces

griffinwork40:
## Summary

- remove wrap-generated leading and trailing spaces at exact word-wrap boundaries
- preserve intentional source indentation, including ANSI-styled list and blockquote rows
- leave `hardWrapToWidth` unchanged so compositor reflow keeps its byte-preservation contract
- add regressions for the original 24-column reproduction and ANSI-prefixed indentation

## Root cause

`wrap-ansi` retained a delimiter space when `trim: false` and the preceding row filled exactly to the requested width. Simply enabling trimming fixed the ragged continuation edge but also stripped meaningful leading indentation. The fix protects source-line indentation with a collision-avoiding private-use marker while `wrap-ansi` trims generated boundary whitespace, then restores the original spaces.

## Verification

- `pnpm test src/cli/wrap.test.ts src/cli/formatter.test.ts`
- `pnpm test src/cli/terminal-compositor.space-fusion.repro.test.ts src/cli/markdown-stream.test.ts src/cli/commands/interactive/tool-lane.overlay.test.ts`
- `pnpm test`
- `pnpm test:coverage`
- `pnpm lint`
- `pnpm build`
- `pnpm audit:sdk:check`
- `pnpm audit:env:check`
- `pnpm scan:env:check`
- `pnpm audit:chalk:check`
- `pnpm fix:pins:check`
- `pnpm audit:deps` (reports the repository's existing dependency advisories: 3 low, 17 moderate, 9 high; command exits successfully)

After rebasing onto `origin/main` at v5.93.2, the targeted wrap and formatter tests plus `pnpm lint` were rerun successfully.

### DIFF
--- a/src/cli/formatter.ts
+++ b/src/cli/formatter.ts
@@ -273,13 +273,12 @@ export function renderMarkdownToTerminal(text: string, opts: RenderMarkdownOptio
                 : srcLine;
               const segs = wrapped.split('\n');
               for (let s = 0; s < segs.length; s++) {
-                // wrapToWidth runs wrap-ansi with trim:false, so a wrap at a
-                // space keeps that space at the START of the next segment.
-                // Strip it on wrap continuations (s > 0) only — otherwise the
-                // hanging indent sits one column past the item text. The first
-                // segment of each source line keeps its leading whitespace so
-                // nested-list indentation survives. Also drop trailing spaces so
-                // they don't inflate the line past the width budget.
+                // Normalize continuation whitespace defensively at this formatting
+                // boundary. wrapToWidth removes generated edge spaces while preserving
+                // source indentation, but this branch also accepts styled and nested
+                // list content whose first segment may intentionally begin with spaces.
+                // Only continuations are left-trimmed; all segments are right-trimmed
+                // so whitespace cannot inflate the line past the width budget.
                 let seg = segs[s]!;
                 if (s > 0) seg = seg.replace(/^ +/, '');
                 seg = seg.replace(/ +$/, '');
--- a/src/cli/wrap.test.ts
+++ b/src/cli/wrap.test.ts
@@ -6,6 +6,8 @@ import { describe, it, expect } from 'vitest';
 import chalk from 'chalk';
 import { wrapToWidth } from './wrap.js';
 
+const stripAnsi = (text: string): string => text.replace(/\x1B\[[0-9;]*m/g, '');
+
 describe('wrapToWidth', () => {
   it('returns short text unchanged', () => {
     expect(wrapToWidth('hello', 80)).toBe('hello');
@@ -18,11 +20,31 @@ describe('wrapToWidth', () => {
     expect(out).toContain('one');
   });
 
+  it('does not carry a boundary space onto the next line', () => {
+    const text =
+      'aaaa bbbb cccc dddd eeee ffff gggg hhhh iiii jjjj kkkk llll mmmm nnnn oooo pppp';
+
+    expect(wrapToWidth(text, 24).split('\n')).toEqual([
+      'aaaa bbbb cccc dddd eeee',
+      'ffff gggg hhhh iiii jjjj',
+      'kkkk llll mmmm nnnn oooo',
+      'pppp',
+    ]);
+  });
+
   it('wraps chalk-colored strings without throwing', () => {
     const colored = chalk.red('redword') + ' ' + 'plain ' + chalk.green('greenword');
     const out = wrapToWidth(colored, 8);
     expect(out).toContain('redword');
     expect(out.split('\n').length).toBeGreaterThan(1);
+  });
+
+  it('preserves intentional indentation after leading ANSI styles', () => {
+    const text = chalk.dim('  • one two three four');
+    const lines = wrapToWidth(text, 12, { breakLongWords: true }).split('\n');
+
+    expect(stripAnsi(lines[0]!)).toBe('  • one two');
+    expect(stripAnsi(lines[1]!)).toBe('three four');
   });
 
   it('does not throw for width 0 or Infinity', () => {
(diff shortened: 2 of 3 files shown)


griffinwork40 (commented):

## Merge decision: **MERGE** — no blocking findings

Light-regime review (hotfix, 3 files, +50/−12). The change does what the PR body says: `wrapToWidth` now removes wrap-generated edge spaces while preserving source indentation, `hardWrapToWidth` is untouched (0 diff hunks), and the two new regressions cover the 24-column reproduction and the ANSI-prefixed indent case. All four findings below are `low` and non-blocking.

Findings were verified by executing the merged `wrapToWidth` at ref `243137be` against the repo's own `wrap-ansi@10`, not by inspection alone.

### Verified correct (positive results, not assumptions)
- **Intent met.** At width 20, `'    indented four spaces then some words that must wrap here'` → old output kept the boundary spaces (`"    indented four \n…\n here"`); new output drops them and keeps the 4-space source indent. Exactly the stated fix.
- **No blank-line regression.** Paragraph gaps (`"a\n\nb"`), whitespace-only lines (`"a\n   \nb"`), and `\r\n` input all produce byte-identical output to the pre-PR implementation.
- **Marker collision avoidance works.** A line containing a stray `U+E000` round-trips it intact — the marker scan correctly picks a different codepoint.
- **Formatter interaction is safe in the normal case.** Running `formatter.ts:275-296`'s `if (s > 0) seg.replace(/^ +/, '')` over real wrapped nested-list output showed no continuation row carrying leading spaces, so the post-hoc left-trim does not undo the indentation `wrap.ts` preserves. `formatter.ts` is a comment-only diff.
- Targeted suites pass locally: `wrap.test.ts`, `formatter.test.ts`, `markdown-stream.test.ts` — 136 tests, 0 failures.

### low · correctness · `src/cli/wrap.ts:40-51` · ref:243137be · citation-type: diff-context
The 256-candidate fallback fails open in two ways at once: when a line contains every codepoint `U+E000`–`U+E0FF`, `indentMarker` is `undefined`, so indentation protection is skipped (`trim: true` strips the real source indent) **and** the restore step still runs `.replaceAll('\uE000', ' ')`, converting a legitimate `U+E000` in the output into a space.

```ts
const protectedLine = indentMarker ? line.replace(...) : line;
return wrapAnsi(protectedLine, w, {...}).replaceAll(indentMarker ?? '\uE000', ' ');
```

Confirmed by execution: input line with all 256 PUA chars + a 2-space indent → output `U+E000` count 0 (was 1) and the leading indent stripped. Reachability is very low — grep across production `src/` finds no other PUA/sentinel producer (only `wrap.ts:40` and `wrap.ts:51` match), so this needs adversarial input to fire.

**Suggestion:** in the `undefined` branch, skip the restore entirely (`indentMarker ? out.replaceAll(indentMarker, ' ') : out`) so an unprotected line is never rewritten.

### low · correctness · `src/cli/wrap.ts:40-46` · ref:243137be · citation-type: diff-context
When a source line's indent run is wider than the wrap width, the marker run itself wraps: the break lands inside the restored spaces, emitting whitespace-only rows plus a continuation row that starts with spaces.

Measured at ref — `indent=20, width=12` → `["            ", "        word", "another"]`; `indent=30, width=10` → three whitespace-only rows. Downstream in `formatter.ts` those rows right-trim to `''` and surface as blank rows in a deeply-nested list at a very narrow terminal.

**Suggestion:** clamp the protected indent to `w - 1` before repeating the marker, so an indent can never consume a full row.

### low · perf-observability · `src/cli/wrap.ts:40` · ref:243137be · citation-type: diff-context
`Array.from({ length: 256 }, …)` allocates 256 strings **per source line, per call**, and `wrapToWidth` sits on the streaming repaint path (`markdown-stream-format.ts:34,70,110` and `stream-renderer-orchestrator-emit.ts:375`, driven by the `throttleMs ?? 33` repaint in `markdown-stream.ts:92`).

Benchmarked at ref against a hoisted-constant variant with identical semantics: **1.98× slower** (2.308 ms vs 1.166 ms per 200-line call). Per-line `wrapAnsi` splitting is not the cause — the hoisted variant matches pre-PR timing. On a realistic 12-paragraph streaming block the delta is 0.26 ms vs 0.17 ms per repaint, i.e. well inside the 33 ms frame budget, which is why this is `low` and not a blocker.

**Suggestion:** hoist the candidate array to a module-level `const` — recovers the full 2× with a one-line change.

### low · test-coverage · `src/cli/wrap.test.ts` · ref:243137be · citation-type: file-state (re-read at ref)
No test covers the all-256 fallback path, a literal-PUA round-trip, a blank or whitespace-only line, or the indent-wider-than-width case. Absence grounded by grep: 0 matches for PUA/collision patterns in `src/cli/**/*.test.ts` relating to wrap, and 0 matches for blank/whitespace-line cases in `wrap.test.ts`. Behavior for blank lines is in fact unchanged (verified above), so this is a missing regression guard rather than a live defect.

**Suggestion:** add a blank-line-preservation assertion and an `indent > width` assertion to `wrap.test.ts`.

### What was not checked
- Citations verified inline against branch HEAD `243137becfb47297195bbb17f212fbaa0cc30ea8`. Two Wave-1 claims were **refuted** by re-derivation at that ref and dropped from this report: "formatter's left-trim defeats the PR's purpose" and "`\r\n` normalization changes caller-visible behavior".
- Stated intent: PR #903 title + body — spec-compliance assessed. No unmet intent or scope creep found.
- Only 3 targeted suites were run (136 tests), not the full `pnpm test`, `pnpm lint`, `pnpm build`, or the `audit:*` gates the PR body reports as passing; those claims are untested here. CI on this head is green across Lint & Build, ubuntu, macOS, PTY-scrollback, docs, and publish-smoke.
- Windows behavior not exercised (CI skips that job). The real TUI compositor was not driven end-to-end; `wrapToWidth` was exercised directly.
- No `critical`/`high` finding surfaced, so the adversarial shadow-verify wave was not triggered.

PR #904, opened 2026-08-04
feat(runtime): make concurrency ceilings operator-configurable

griffinwork40:
## Summary

- add `AFK_MAX_CONCURRENT_SAFE_TOOL_CALLS` to tune concurrency-safe tool batches
- add `AFK_MAX_CONCURRENT_SUBAGENT_CALLS` to tune compose/DAG layers and skill waves
- add `AFK_MAX_CONCURRENT_BACKGROUND_JOBS` to tune the `BackgroundAgentRegistry` in-flight ceiling
- preserve the default of 8 and existing programmatic override precedence
- clamp each knob to a documented upper bound (`1-32`, `1-32`, `1-64` respectively); out-of-range input falls back to the default and warns, mirroring `AFK_MAX_NESTING_DEPTH`
- surface all three resolved values in `afk doctor`
- document why compose does not expose a model-controlled concurrency knob
- document why detached bash descendants are not session-reaped without an explicit background-job contract
- regenerate the environment registry and add resolver plus production-wiring tests

## Why

AFK already bounds both fan-out paths, but operators had no way to lower those ceilings under provider-rate-limit or local-resource pressure. Claude Code exposes an equivalent operator control. These settings keep concurrency governance outside model-facing tool schemas.

## Defaults are unchanged

All three knobs resolve to the values that were already in effect before this PR when the vars are unset:

| Variable | Default | Accepted range | Pre-PR source |
|---|---|---|---|
| `AFK_MAX_CONCURRENT_SAFE_TOOL_CALLS` | 8 | 1-32 | `dispatcher.ts` |
| `AFK_MAX_CONCURRENT_SUBAGENT_CALLS` | 8 | 1-32 | `concurrency-pool.ts` |
| `AFK_MAX_CONCURRENT_BACKGROUND_JOBS` | 10 | 1-64 | `background-registry.ts` |

## Review follow-ups addressed

Applied in `a2302525` / `5c2969d2` in response to the automated review:

- **High** — `DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS` had silently dropped from 8 to 4 while centralizing, halving default fan-out for every caller that omits `maxConcurrency`. Restored to 8; `dag.test.ts` now asserts the zero-config default end-to-end with no env stub, so the regression cannot recur unnoticed.
- **Medium** — the parser accepted any positive integer, so an absurd `AFK_MAX_CONCURRENT_BACKGROUND_JOBS` removed the only safeguard on that path. Each key now carries a per-key ceiling constant with an `Invariant:` rationale.
- **Low** — `AFK_MAX_CONCURRENT_BACKGROUND_JOBS` is now named in this description.
- **Low** — restored the deleted zero-config `runDAG` default test; added the mirrored `runWave` env-ceiling test.
- **Low** — `concurrency.test.ts`'s warn-once assertion no longer depends on module declaration order; `resetConcurrencyWarnings()` is called per test.
- Merged `origin/main` (no rebase, PR commits preserved) and regenerated the environment registry, resolving the `docs/env-registry.*` conflict.

## Verification

Re-run on the merged tip `5c2969d2`:

- `pnpm test` — 793 files, 15083 passed, 2 skipped
- `pnpm lint` — passed
- `pnpm build` — passed
- `pnpm audit:sdk:check` — passed
- `pnpm audit:env:check` — passed
- `pnpm scan:env:check` — passed (148 vars in sync)
- `pnpm audit:chalk:check` — passed
- `pnpm fix:pins:check` — passed


griffinwork40:
## PR #904 — Approve with follow-ups

Nothing in the diff is wrong. Defaults are byte-identical when the vars are unset, explicit programmatic overrides still win at all three sites, and all six CI jobs are green. The real findings are about **completeness of the operator promise**, not correctness.

**Two majors, both fast-follow rather than blocking:**

**M1 — the knob misses the path that matters most.** `background-registry.ts:146,207` hardcodes `DEFAULT_MAX_CONCURRENT_JOBS = 10` with no env read. The PR's stated purpose is lowering ceilings under rate-limit pressure, but an operator who sets `AFK_MAX_CONCURRENT_SUBAGENT_CALLS=2` mid-429-incident still gets up to 10 concurrent background subagents, because `agent(mode:"background")` bypasses `runWave`/`runDAG` entirely. The PR's own doc comment name-drops "the background-job ceiling of 10" as a sibling while leaving it un-tunable.

**M2 — silent failure on a knob built for incidents.** The resolved ceiling is surfaced nowhere — no `afk doctor`, no `debugLog`, no trace metadata. Combined with a fallback that swallows every invalid input, a typo'd var name yields 8 with no signal while the operator believes throttling is on. Settability is *not* the gap (`settable-keys.ts:118-126` already classifies both as settable); visibility is.

**Notable minors:** the deleted `dag.test.ts` default case (`peak === 4`) leaves the composed default unasserted end-to-end — though the resolver fallback and the wiring are each covered separately, so a regression would likely still be caught; `runWave`'s env path has zero coverage; and there's no upper clamp unlike `AFK_MAX_NESTING_DEPTH`'s documented 0–6 range.

**One severity I corrected downward.** The correctness agent framed the missing upper bound as "removes the safety ceiling entirely." I checked `settleWithConcurrencyLimit` directly: `poolSize = Math.min(Math.max(1, floor(limit)), items.length)` (`concurrency-pool.ts:66`) clamps to fan-out width, so an absurd value means *no throttling*, i.e. pre-ceiling behavior — not unbounded allocation. Real residual risk is a `80`-for-`8` typo silently disabling the ceiling, which M2 would also catch.

I also caught a bad citation: the scope agent placed the `bash.ts` lifecycle comment at `:322-324`; it's at `:230-232`, a 6-line block, so no `Invariant:`/`Contract:`/`History:` prefix is required. Its substantive claims (PR-body disclosure, comment truthfulness) both held up — group-kill genuinely fires only on the timeout and abort paths.

**Done**
- Reviewed PR #904 across four dimensions via parallel read-only agents, then independently re-derived six load-bearing claims (background-job ceiling, dispatcher precedence ternary, bash.ts comment location/length, surviving dag test cases, global env cleanup, clamp semantics) and corrected two agent errors.
- Evidence: full report at `/Users/griffinlong/.afk/workspace/reports/review-pr-904-2026-08-04.md` (2 major / 6 minor / 5 nit findings, each with file:line evidence, plus a verified-clean list and CI-gate confirmation).
- World changes: none in the repo — read-only throughout. The working tree stayed on `fix/wrap-boundary-leading-space` at `243137be` with its 2 uncommitted changes untouched; no branch checkout, no fetch, no GitHub comment posted.
- Pending: nothing was posted to the PR (`/review` is read-only by contract) — say the word and I'll file M1/M2 as issues or drop the summary as a PR review comment. The report file is uncommitted by design; it lives outside the repo in your reports dir so it never dirtied the checkout.

---

griffinwork40 (commented):

## Summary

This PR makes AFK's concurrency ceilings operator-configurable, centralizing three limits into a new `src/config/concurrency.ts` resolver with warn-once fallback, registry entries, and `afk doctor` surfacing. The security surface is clean — no secret exposure, injection, or auth issues — and input validation correctly rejects negatives, floats, `Infinity`, empty, and non-numeric values. Both documentation-only intent bullets are satisfied (the compose invariant comment and the bash lifecycle note), doctor wiring is correct, and the per-call resolver cost is negligible (an O(3) lookup per batch/layer/wave, not per tool call).

One substantive issue stands out: the subagent fan-out default silently changes from 8 to 4. Separately, the branch is **CONFLICTING** — it is behind `main` (merge-base `5fac32b6`), and `docs/env-registry.json` and `docs/env-registry.md`, which this PR also edits, have changed upstream since.

**Merge decision: DO NOT MERGE** — the undisclosed halving of production fan-out warrants an explicit decision (restore 8, or state the new value in the description), and the branch cannot merge as-is until the registry conflict is rebased.

## Findings

### High

- **api-compat · correctness · spec-compliance — `src/config/concurrency.ts:4`**
  `DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS` drops from 8 to 4 while its name and export shape stay identical, halving default parallel fan-out for every caller that omits `maxConcurrency`.

  ```diff
  -export const DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS = 8;
  +export const DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS = 4;
  ```

  This is reachable, not dead config. Every production call site of `runDAG`/`runWave`/`runSubagentDAG` omits an explicit `maxConcurrency`: `dag-subagent.ts:184`, `compose-executor.ts:754-763` (the `compose` tool itself), `farm.ts:391`, and `audit-fit/index.ts:426`. `SubagentDAGOptions` (`dag-subagent.ts:78-92`) exposes no `maxConcurrency` field at all, so compose callers cannot override it even deliberately. A 5-node compose run that previously fit one layer now runs two.

  Worth noting for provenance: the first commit `cc24fa39` kept the value at 8; the drop was introduced by the tip commit `9e0e7f9d`, whose subject — *"fix: centralize concurrency environment limits"* — does not disclose a default change. `main` is still at 8. The Summary bullet "preserve the default of 8" is ambiguous (the `SAFE_TOOL_CALLS` ceiling is 8 and genuinely is preserved), but the subagent default was also 8 before this PR, so under either reading the change is undisclosed.

  *Suggestion:* restore 8, or state the new default explicitly in the description — `doctor.test.ts:85` already pins `effective value 4`, so this needs a recorded decision rather than silence.

### Medium

- **security — `src/config/concurrency.ts:51-52`**
  The parser accepts any positive integer with no upper bound, so `AFK_MAX_CONCURRENT_BACKGROUND_JOBS=1000000000` removes the practical cap on simultaneously registered background `AgentSession` jobs.

  ```ts
  const parsed = Number(rawValue);
  if (rawValue.trim() !== '' && Number.isInteger(parsed) && parsed > 0) {
  ```

  The DAG/wave/dispatcher paths are self-limiting — `poolSize = Math.min(Math.max(1, Math.floor(limit)), items.length)` (`concurrency-pool.ts:62`) and compose's `MAX_NODES = 20` — but `BackgroundAgentRegistry` compares the raw ceiling against a running count (`background-registry.ts:329`) with no array-length backstop, so the ceiling is the only safeguard there.

  *Suggestion:* clamp to an upper bound, mirroring `MAX_NESTING_DEPTH_CEILING = 6` (`nesting.ts:40`).

### Low

- **spec-compliance (scope creep) — `src/config/concurrency.ts:5`, `background-registry.ts:205`**
  `AFK_MAX_CONCURRENT_BACKGROUND_JOBS` is a third production-wired environment variable the Summary never names. *Suggestion:* name it in the description.

- **test-coverage — `src/agent/dag.test.ts`**
  The only test of `runDAG`'s zero-config default path was deleted and replaced with an env-stubbed test, which is precisely why the 8→4 change passes CI unnoticed. *Suggestion:* restore a no-env-stub test asserting the true default.

- **test-coverage — `src/agent/subagent-wave.test.ts`**
  The intent names "skill waves", but only `runDAG` received an env-ceiling test; `runWave` has none. *Suggestion:* add the mirrored `runWave` test.

- **test-coverage — `src/config/concurrency.test.ts:41-52`**
  The "warns only once per key" assertion holds only because module-level `warnedKeys` (`concurrency.ts:30`) is never reset or exported and happens to be empty at that point in file order — an artifact of declaration order, not a guarantee. *Suggestion:* reset or scope `warnedKeys` per test.

## What was not checked

Citations were verified inline against branch HEAD `9e0e7f9d` with `git show`/`git grep`; none were fabricated or diff-only, and every absence claim was re-confirmed at the ref. Tests were **not** executed — the CI/verification results in the description are taken at face value. The merge conflict on `docs/env-registry.json`/`.md` was not resolved. A stated intent was available (PR title + body), so spec-compliance **was** assessed against it. Reviewed as a whole-diff pass; runtime behavior under real provider rate limits was not exercised.

griffinwork40 (commented):

## Summary

PR #904 replaces three hardcoded concurrency ceilings (safe-tool calls, subagent calls, background jobs) with operator-configurable env vars that fall back to the pre-PR default on invalid input and warn once, mirroring `AFK_MAX_NESTING_DEPTH`. Defaults, programmatic-override precedence across all four call sites, `afk doctor` surfacing, and isolation from model-facing tool schemas were independently re-derived at ref `7cb42487` and hold. Three low-severity findings survive verification, none blocking; two fabricated citations and two Wave 1 claims — one refuted, one downgraded to informational — were resolved before this report.

## Findings

**low · security (defense-in-depth) · `src/config/concurrency.ts:3-5,16,28,40`**
An operator can now raise safe-tool and subagent concurrency from the previous hardcoded 8 to 32, and background jobs from 10 to 64, via env var alone — pre-PR this required a source edit. The PR frames the knobs as controls for lowering ceilings under pressure; the widening direction is a new capability reachable by a stray inherited env var in shared deployment config.
Evidence: `MAX_CONCURRENT_SAFE_TOOL_CALLS_CEILING = 32`, `MAX_CONCURRENT_SUBAGENT_CALLS_CEILING = 32`, `MAX_CONCURRENT_BACKGROUND_JOBS_CEILING = 64`.
Fix: state the ceiling as an explicit operator-trust boundary in the env-registry description, so a 4-6x fan-out increase is a deliberate, auditable act, not a side effect of an inherited variable.

**low · test-coverage · `src/agent/subagent-wave.test.ts:325-368`**
`runWave`'s zero-config default path (no `maxConcurrency` option, no env var set) is never pinned end-to-end against `DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS`; the added wave test only exercises the env-override path. A future silent regression of the default at this call site fails no test.
Evidence: the added test stubs `AFK_MAX_CONCURRENT_SUBAGENT_CALLS` to `'2'` and asserts `peak === 2`, never against the bare default; `git grep DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS` on this file at the ref returns zero matches. The sibling `runDAG` path pins it directly — `dag.test.ts:655-656` asserts `peak === DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS` and that the constant equals `8`, the exact regression class the PR body names as fixed for `dag.test.ts`, uncovered at the sibling skill-wave site the stated intent names explicitly.
Fix: mirror the `dag.test.ts` case here — no `maxConcurrency`, no `vi.stubEnv`, a wave over eight tasks, assert `peak === DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS`.

**nit · spec-compliance · `docs/env-registry.md:5`**
The PR body's Verification block reports `pnpm scan:env:check` passing with 148 vars in sync; the reviewed ref documents 149. The body's numbers were captured against the earlier tip `5c2969d2`, not the final commit, so they do not evidence CI on this code.
Evidence: `docs/env-registry.md:5` states 149 vars across 12 categories; `docs/env-registry.json` contains 149 `"name"` entries at the ref — the registry is internally consistent, only the PR body's prose is stale.
Fix: re-run the verification block against `7cb42487` and update the counts, or note the commit they were captured on.

## Verified clean

- Defaults unchanged: `concurrency.ts:3-5` = 8 / 8 / 10, matching the intent's table.
- Out-of-range input falls back to the default (not clamped) and warns once, deduped via module-level state: `concurrency.ts:117-140`, `:90`.
- Digit-anchored input grammar rejects `0x8`, `1e1`, `8.0`, signed/whitespace forms, mirroring `resolveMaxNestingDepth`.
- Programmatic-override precedence holds at all four call sites: `dispatcher.ts:341-345`, `dag.ts:163`, `subagent/wave.ts:69`, `background-registry.ts:205`.
- `afk doctor` surfaces all three knobs with effective value, source, and a range-stating fix hint: `doctor-checks.ts:217-230`, wired at `:243`.
- No model-facing path reaches a ceiling: `compose-executor.ts:755-759` documents the invariant; `SubagentDAGOptions` exposes no `maxConcurrency` field.
- Doctor warning strings echo only the three hardcoded, non-secret keys' raw values — no arbitrary-env or credential exposure path.
- `concurrency-pool.ts` re-exports `DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS` unchanged, so existing importers are unaffected.

## What was not checked

- Citations verified inline against branch HEAD `7cb42487f48cefd3d0cdad806eec3dcaf918c3f2`.
- Two fabricated citations were caught and discarded: `doctor-checks.ts:498-511` and `concurrency.ts:806-809` (both exceed the real file lengths of 255 and 169 lines); the underlying claims were re-verified true and moved to Verified clean. Two findings were dropped: a test-coverage claim that no test asserts the literal defaults 10/8, refuted by `concurrency.test.ts:40` and `dispatcher-concurrency-env.test.ts:18`; and removal of the never-exported `DEFAULT_MAX_CONCURRENT_JOBS`, downgraded to informational since `git grep` at the ref found zero references in `src/`.
- Stated intent: PR #904 title+body — spec-compliance assessed.
- Tests, lint, and build were not executed by this review; the PR body's own reported results were not independently reproduced.
- The full call graph of who sets these env vars in deployment was not traced.
- Doctor's human-readable text-mode rendering was not exercised.

## Merge decision

**MERGE** — zero critical or high findings, all three surviving items are low or nit and none blocks; the subagent-wave default-path test-coverage gap is the recommended non-blocking follow-up.

griffinwork40:
Both Wave 1 agents returned. Let me run Wave 1.5 (citation + absence-claim verification) inline.

**Wave 1.5 -- Citation + Absence-Claim Verification**

All findings from Wave 1 are `low` or `nit` severity. Per the protocol, I check every `file-state` citation at any severity. Let me verify each:

**Check A -- Citation verification:**

1. **security agent, finding 1** (`src/config/concurrency.ts:79-91`): Claims `resetConcurrencyWarnings` is exported with test-only JSDoc. Pre-fetched file confirms lines 89-91: `export function resetConcurrencyWarnings(): void { warnedKeys.clear(); }` with JSDoc "Test-only: clear the once-per-key warning latch". **Verified.**

2. **security agent, finding 2** (`src/config/env.ts:226-250`): Claims `category: 'model'` for the three new vars. Pre-fetched diff shows the ENV_REGISTRY entries with `category: 'model'`. **Verified (diff-context).**

3. **correctness agent, finding 1** (`src/config/concurrency.ts:153-156`): Claims `resolve()` calls `getConcurrencyStatus(definition)` without forwarding `warn`. Pre-fetched file line 155: `return getConcurrencyStatus(definition).effectiveValue;` -- no second arg. Line 100: `warn = true` default. **Verified.**

4. **correctness agent, finding 2** (`src/config/concurrency.test.ts:103-110`): Claims about `' '` in the grammar test. Diff lines 665-673 show `it.each(['0x8', '1e1', '8.0', '+8', ' '])`. The comment at line 665 says "Bare Number() would coerce every one of these to an in-range integer." The claim that `Number(' ')` is `0` (not an in-range positive integer) is correct. **Verified -- the observation about the mislabelled comment is accurate.**

5. **correctness agent, finding 3** (`src/config/concurrency.ts:113-119`): Claims `Number.isInteger` is redundant with the regex. Pre-fetched file line 119: `const isPositiveInteger = /^\d+$/.test(trimmed) && Number.isInteger(parsed) && parsed > 0;`. **Verified.**

6. **correctness agent, finding 4** (`src/config/concurrency.ts:113-116`): Claims about the `nesting.ts` cross-reference comment. Pre-fetched file confirms the comment. **Verified.**

**Check B -- Absence-claim verification:**
No absence claims were made by either agent. No verification needed.

**Wave 1.5 manifest:** All 6 citations verified. 0 fabricated, 0 diff-only, 0 false-absent.

No `critical`, `high`, or `blocking`-overridden findings -- no shadow-verify needed. Proceeding to Wave 2 synthesis.

---

## PR #904 Review: `feat(runtime): make concurrency ceilings operator-configurable`

**Branch:** `afk/bash-parallel-concurrency` -> `main`
**Reviewed ref:** `7cb42487f48cefd3d0cdad806eec3dcaf918c3f2`
**Regime:** full (557 lines, 20 files)
**Change type:** feature

### Summary

Adds three operator-configurable env vars (`AFK_MAX_CONCURRENT_SAFE_TOOL_CALLS`, `AFK_MAX_CONCURRENT_SUBAGENT_CALLS`, `AFK_MAX_CONCURRENT_BACKGROUND_JOBS`) to control concurrency ceilings that were previously hardcoded. Centralizes resolution in a new `src/config/concurrency.ts` module, wires each consumer site to call a resolver function instead of reading a module-scope constant, preserves existing defaults (8/8/10), enforces upper-bound ceilings (32/32/64), surfaces all three in `afk doctor`, and adds comprehensive test coverage.

---

### Findings

**1. low · blocking:false · confidence:high · security · `src/config/env.ts:226-250` · ref:7cb42487 · diff-context**

All three new env vars are registered with `category: 'model'`. They govern runtime concurrency ceilings, not model selection -- `'process'` or a new `'concurrency'` category is more semantically accurate.

```typescript
category: 'model'  // for AFK_MAX_CONCURRENT_SAFE_TOOL_CALLS, etc.
```

**Suggestion:** Register under `'process'` or introduce a `'concurrency'` category.

---

**2. low · blocking:false · confidence:high · security · `src/config/concurrency.ts:79-91` · ref:7cb42487 · file-state**

`resetConcurrencyWarnings()` is exported from the production module with only a JSDoc comment marking it test-only. No enforcement prevents production callers from invoking it.

```typescript
export function resetConcurrencyWarnings(): void {
  warnedKeys.clear();
}
```

**Suggestion:** Expose through a test-only barrel or use vitest's `vi.importActual` pattern instead.

---

**3. low · blocking:false · confidence:high · correctness · `src/config/concurrency.ts:153-156` · ref:7cb42487 · file-state**

The scalar `resolve*()` functions always call `getConcurrencyStatus(definition)` with `warn=true` (the default), creating an asymmetry with `getConcurrencyStatuses(warn)` which correctly threads the parameter. Callers wanting a silent read (e.g. telemetry probes) have no path through the scalar API.

```typescript
return getConcurrencyStatus(definition).effectiveValue; // always warn=true
```

**Suggestion:** Thread a `warn` parameter through `resolve()` or document that the scalar resolvers always warn.

---

**4. low · blocking:false · confidence:high · test-coverage · `src/config/concurrency.test.ts:103-110` · ref:7cb42487 · file-state**

The `' '` (single space) test case in the grammar rejection suite is mislabelled. The comment claims "Bare Number() would coerce every one of these to an in-range integer" but `Number(' ')` is `0`, not a positive in-range integer -- Number() alone would already reject it. The case provides no marginal grammar coverage beyond an empty-string test.

**Suggestion:** Replace with a case that actually tests the `/^\d+$/` guard, such as `'\t8'` or `'8\n'`, or relabel the row as an "empty-after-trim" case.

---

**5. nit · blocking:false · confidence:high · correctness · `src/config/concurrency.ts:113-119` · ref:7cb42487 · file-state**

`Number.isInteger(parsed)` is tautologically true for any string matching `/^\d+$/` -- the triple guard (`regex && isInteger && > 0`) is defensive but the middle clause is redundant.

```typescript
const isPositiveInteger = /^\d+$/.test(trimmed) && Number.isInteger(parsed) && parsed > 0;
```

**Suggestion:** Simplify to `/^\d+$/.test(trimmed) && parsed > 0` with a comment explaining the regex guarantees integeriness.

---

**6. nit · blocking:false · confidence:high · spec-compliance · `src/config/concurrency.ts:113-116` · ref:7cb42487 · file-state**

The inline comment asserts `matching resolveMaxNestingDepth` as a consistency claim. If `nesting.ts` later changes its grammar, this comment creates a silent drift risk.

**Suggestion:** Convert to a `Contract:` block naming the invariant ("only decimal digit strings accepted") rather than the peer implementation.

---

### Dimensions with no issues

- **api-compat:** no issues found -- all re-exported constants (`DEFAULT_MAX_CONCURRENT_SAFE_TOOL_CALLS`, `DEFAULT_MAX_CONCURRENT_SUBAGENT_CALLS`) are preserved as pass-throughs; programmatic `maxConcurrency` overrides retain precedence; `SubagentDAGOptions` intentionally omits `maxConcurrency`.
- **perf-observability:** no issues found -- `afk doctor` surfaces all three resolved values via `getConcurrencyStatuses(false)` (no warn side-effect on inspection).

---

### What was not checked

- Citations verified inline against branch HEAD `7cb42487f48cefd3d0cdad806eec3dcaf918c3f2`. All 6 file-state citations confirmed.
- Stated intent: PR #904 title+body -- spec-compliance assessed. All stated deliverables are present.
- Did not verify `nesting.ts` grammar alignment at this ref (noted as a nit).
- Did not review the Telegram or daemon surfaces for concurrency-related behavior.

---

**Decision: MERGE -- 0 blocking (4 low, 2 nit).**

All six findings are non-blocking quality improvements. The PR satisfies every point of its stated intent: three env knobs are wired correctly with fallback-on-invalid semantics, defaults are preserved at 8/8/10, ceilings clamp at 32/32/64, `afk doctor` surfaces all three, compose intentionally blocks model-controlled fan-out widening (documented with an Invariant comment), and test coverage spans unit, integration, and the zero-config regression path.

**Done**
- What was done: Reviewed PR #904 (`feat(runtime): make concurrency ceilings operator-configurable`) across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability dimensions
- Evidence: Wave 1 (2 parallel agents) + Wave 1.5 inline citation verification (6/6 verified) + Wave 2 synthesis above
- What changed: nothing -- read-only review
- Deferred: none

---

PR #909, opened 2026-08-05
fix(pricing): correct prompt-cache cost accounting and surface hit rate

griffinwork40:
## Summary

Started as "are we caching tool results?" The caching mechanism turned out to be sound — the cached prefix is byte-stable (the `Date:` line in the system prompt is guarded to return the same string reference across turns, tools are frozen at session construction, HOT.md is read once pre-session), tool_result blocks do land inside the cached prefix, and the non-mutating stamp discipline is correct.

The **accounting around it** was not. Three defects, one root cause: cache-aware corrections were made at the provider layer and never propagated to the pricing, display, and observability consumers downstream.

### 1. Cache writes billed at the 5-minute rate while we request 1-hour

`deriveCallCostUsd` hardcoded `inputPerMTok * 1.25` — the 5m TTL rate. `cache-policy.ts:26` defaults `TTL_DEFAULT = '1h'`, which Anthropic bills at **2x** base input. Every cost figure understated the cache-write component by 37.5%, on every cached session, which is the default.

> 5-minute cache write tokens are 1.25 times the base input tokens price
> 1-hour cache write tokens are **2 times** the base input tokens price
> — [prompt-caching#pricing](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#pricing)

Now derived from the API's own `usage.cache_creation` split (`ephemeral_5m_input_tokens` / `ephemeral_1h_input_tokens`) — what was actually billed, and the only source that stays correct when a request mixes TTLs. When absent, writes fall back to the configured TTL instead of silently assuming 5m.

### 2. Plain input double-subtracted, clamped to zero

```ts
const plainInput = Math.max(0, inputTokens - cachedInputTokens - cacheCreationTokens);
```

`input_tokens` already excludes cache reads and writes (docs' own worked example: `cache_read: 100000`, `input_tokens: 50`, total processed 100,050). Subtracting them again drives the term to `0` in any warm session, since cache reads dwarf plain input.

This repo already found and documented this exact misconception — `providers/shared/auto-compact.ts:30-34` says *"A prior comment here claimed Anthropic's `input_tokens` 'already includes cache reads' — that is wrong."* The correction never reached the cost function.

### 3. Opus 4.5 priced 3x too high

The `claude-opus-4-5-20250929` row carried `$15/$75` — the **retired Opus 4.1/4.0** rates. Published Opus 4.5 pricing is `$5/$25`. Same copy-paste class as the Haiku 3.5→4.5 mixup that `types.test.ts` already has a regression test for. Also adds missing Opus 4.6/4.7/4.8 and Sonnet 4.6 rows, which previously returned `undefined` cost.

### 4. Status line and `/tokens` summed a mixed-basis object

`sumProviderUsage` accumulates input/output cumulatively across a turn's rounds but keeps cache counts at last-round-only (deliberately — `usage.ts:29-35`). By round N the latest `cache_read` already contains every prior round, so adding cumulative input/output back on top double-counts, and the inflation grows with round count.

`turn-accumulator.ts:88-96` documents this precise trap and computes `contextWindowTokens` correctly because of it. Two display sites re-summed anyway — `context-sampler.ts:148` and `info.ts:117` — rendering an inflated absolute right beside a correctly-derived percentage. `config-types.ts:711` documented the wrong formula as if it were live.

### 5. No way to notice any of the above

`afk trace show` printed reason/turns/cost/stop-reason and dropped `cacheRead`/`cacheCreation`, which the schema has carried all along. `/cost` and the turn footer show nothing. Nothing anywhere computed a hit rate. A cache degrading to a 100% miss would have surfaced only on the monthly bill.

## What changed

| Area | Change |
|---|---|
| `pricing.ts` **(new)** | TTL-aware write rates, cache-exclusive plain input, corrected Opus 4.5, added 4.6/4.7/4.8 + Sonnet 4.6. Pure — no env, no clock. |
| `types.ts` | Re-exports pricing; `resolveCacheWriteSplit` reads the API breakdown with a configured-TTL fallback. Single env boundary for pricing. |
| `shared/auto-compact.ts` | Adds `context_window_tokens` — the one scalar display code may safely read — with the mixed-basis invariant documented on the type. |
| `context-sampler.ts`, `info.ts` | Stop summing. Fall back to input+output (cache-excluded), mirroring `contextWindowTokensUsed`. |
| `trace-usage-format.ts` **(new)** | `cache r=/w=/hit=%` on the closure line. `/tokens` gains a hit-rate row. |
| `config-types.ts` | Doc comment now describes the formula the code actually runs. |

Two existing tests asserted the buggy behaviour and are corrected with comments naming what they were locking in:
- `types.test.ts` expected plain input `800` from `1000 - 200 cached`
- `status-line-context.test.ts` expected `used = 1000 // 250+250+500`

## File-size convention

Pricing moved out of `types.ts` (595 LOC, well over the 350 ceiling) into its own file rather than growing it further; `types.ts` re-exports so call sites are unchanged. Same reason `trace-usage-format.ts` exists instead of a new branch in `trace.ts` (804 LOC).

## Verification

```
pnpm lint                 clean
pnpm build                clean
pnpm test                 789 files / 14,968 passed / 2 skipped
pnpm test:coverage        floors met
pnpm audit:sdk:check      lock matches (no new SDK symbol — cache_creation
                          already on the imported Usage type)
pnpm audit:env:check      0 violations
pnpm audit:chalk:check    0 violations
pnpm fix:pins:check       up to date
```

New tests pin every published rate as a golden, both TTL branches, the mixed-TTL split, the no-double-subtract property, and the 0%/100% cache-hit rendering.

## Not addressed here (deliberate — behaviour changes wanting a call)

- **`baseUrl` blanket-disables caching** (`cache-policy.ts:37-45`). Any non-empty `baseUrl` is assumed to be a local shim, but `AFK_MODEL_<TIER>_BASE_URL` is documented for remote Anthropic-compatible gateways — those users silently lose caching with no override. Wants either an `AFK_FORCE_PROMPT_CACHE=1` escape hatch or a loopback-only check.
- **1h TTL as the global default.** Anthropic's guidance: *"If you have prompts used more frequently than every 5 minutes, continue to use the 5-minute cache, because this will continue to be refreshed at no additional charge."* An active REPL refreshes free at 1.25x writes; 1h pays 2x for durability it never uses. Right for daemon/Telegram idle, wrong for REPL — a surface-aware default is a real cost lever, and now that #1 is fixed the premium is finally visible.
- **No anchor breakpoint.** Only 2 of 4 slots used, both at the frontier. Fine for normal rounds, but a wide parallel fan-out (>20 blocks in one round) exceeds the lookback window with nothing behind it. Lowest confidence item — the subagent assigned to it timed out and I reasoned it through from the docs rather than measuring real fan-out widths.

### DIFF
--- a/src/agent/providers/anthropic-direct/pricing.test.ts
+++ b/src/agent/providers/anthropic-direct/pricing.test.ts
@@ -1,0 +1,119 @@
+/**
+ * Unit tests for the prompt-cache-aware pricing model.
+ *
+ * Every assertion here is a golden pinned against Anthropic's published
+ * rates (https://platform.claude.com/docs/en/build-with-claude/prompt-caching#pricing,
+ * verified 2026-08-05). A wrong rate fails silently and persists into saved
+ * cost reports, so these are exact values rather than "cost > 0".
+ *
+ * @module agent/providers/anthropic-direct/pricing.test
+ */
+
+import { describe, it, expect } from 'vitest';
+import { deriveCallCostUsd, MODEL_PRICING } from './pricing.js';
+
+const M = 1_000_000;
+
+describe('deriveCallCostUsd — cache-write TTL rates', () => {
+  it('prices a 1h cache write at 2x base input, not the 5m 1.25x', () => {
+    // Regression: the rate was hardcoded at 1.25x (the 5-minute rate) while
+    // cache-policy.ts defaults AFK_PROMPT_CACHE_TTL to '1h', understating the
+    // write component of every cached session by 37.5%.
+    // sonnet-5: $3 base → 5m write $3.75, 1h write $6.00.
+    const at5m = deriveCallCostUsd('claude-sonnet-5', 0, 0, 0, M, {
+      ephemeral5m: M,
+      ephemeral1h: 0,
+    });
+    const at1h = deriveCallCostUsd('claude-sonnet-5', 0, 0, 0, M, {
+      ephemeral5m: 0,
+      ephemeral1h: M,
+    });
+    expect(at5m).toBeCloseTo(3.75, 8);
+    expect(at1h).toBeCloseTo(6.0, 8);
+    // The whole point: 1h costs strictly more. Guards against a future edit
+    // that collapses both branches back to one rate.
+    expect(at1h!).toBeGreaterThan(at5m!);
+  });
+
+  it('bills a mixed-TTL call by segment rather than picking one rate', () => {
+    // The API bills 1h and 5m segments separately when breakpoints mix TTLs.
+    const cost = deriveCallCostUsd('claude-sonnet-5', 0, 0, 0, M, {
+      ephemeral5m: M / 2,
+      ephemeral1h: M / 2,
+    });
+    expect(cost).toBeCloseTo(0.5 * 3.75 + 0.5 * 6.0, 8);
+  });
+
+  it('defaults to the 5m rate when no TTL split is supplied', () => {
+    const cost = deriveCallCostUsd('claude-sonnet-5', 0, 0, 0, M);
+    expect(cost).toBeCloseTo(3.75, 8);
+  });
+
+  it('derives both write rates from base input when a row omits them', () => {
+    // Rows may omit explicit cache rates; the 1.25x / 2x multipliers apply.
+    const rows = [...MODEL_PRICING.entries()];
+    for (const [model, p] of rows) {
+      const oneHour = deriveCallCostUsd(model, 0, 0, 0, M, {
+        ephemeral5m: 0,
+        ephemeral1h: M,
+      });
+      const expected = p.cacheWrite1hPerMTok ?? p.inputPerMTok * 2;
+      expect(oneHour, `1h write rate for ${model}`).toBeCloseTo(expected, 8);
+    }
+  });
+});
+
+describe('deriveCallCostUsd — plain input is cache-exclusive', () => {
+  it('does not subtract cache counts from input_tokens', () => {
+    // Per the Messages API docs, `input_tokens` counts only tokens neither
+    // read from nor used to create a cache. The old code subtracted both off
+    // it again and clamped at zero — so in a warm session (cache_read >>
+    // input) the plain-input term silently vanished.
+    const cost = deriveCallCostUsd('claude-sonnet-5', 1000, 0, 100_000, 0);
+    const expectedPlain = (1000 / M) * 3.0;
+    const expectedRead = (100_000 / M) * 0.3;
+    expect(cost).toBeCloseTo(expectedPlain + expectedRead, 8);
+  });
+
+  it('keeps the plain-input term when cache counts dwarf it', () => {
+    // The clamping bug's signature: cost identical with and without a huge
+    // cache_read would mean plain input was being zeroed out.
+    const withCache = deriveCallCostUsd('claude-sonnet-5', 500, 0, 200_000, 0)!;
+    const readOnly = deriveCallCostUsd('claude-sonnet-5', 0, 0, 200_000, 0)!;
+    expect(withCache - readOnly).toBeCloseTo((500 / M) * 3.0, 8);
+  });
+});
+
+describe('deriveCallCostUsd — published rate goldens', () => {
+  it('opus-4-5 uses $5/$25, not the retired opus-4.1 $15/$75', () => {
+    // Regression: the 4.5 row carried the retired Opus 4.1/4.0 rates,
+    // overstating every Opus 4.5 cost report 3x. Same copy-paste class as the
+    // Haiku 3.5 -> 4.5 mixup. 1M in + 1M out = $5 + $25 = $30.
+    const cost = deriveCallCostUsd('claude-opus-4-5-20250929', M, M, 0, 0);
+    expect(cost).toBeCloseTo(30.0, 8);
+  });
+
+  it('opus-5 uses the published $5.00/$25.00 rates', () => {
+    expect(deriveCallCostUsd('claude-opus-5', M, M, 0, 0)).toBeCloseTo(30.0, 8);
+  });
+
+  it('returns undefined for an unknown model rather than zero', () => {
+    expect(deriveCallCostUsd('claude-unknown-99', 1000, 500, 0, 0)).toBeUndefined();
+  });
+
+  it('every table row produces a positive cost for non-zero tokens', () => {
+    for (const [model] of MODEL_PRICING) {
+      const cost = deriveCallCostUsd(model, 1000, 500, 0, 0);
+      expect(cost, `cost for ${model}`).toBeDefined();
+      expect(cost!, `cost for ${model}`).toBeGreaterThan(0);
+    }
+  });
+
+  it('every table row prices a 1h write above a 5m write', () => {
+    for (const [model] of MODEL_PRICING) {
+      const w5 = deriveCallCostUsd(model, 0, 0, 0, M, { ephemeral5m: M, ephemeral1h: 0 })!;
+      const w1 = deriveCallCostUsd(model, 0, 0, 0, M, { ephemeral5m: 0, ephemeral1h: M })!;
+      expect(w1, `1h vs 5m for ${model}`).toBeGreaterThan(w5);
+    }
+  });
+});
--- a/src/agent/providers/anthropic-direct/pricing.ts
+++ b/src/agent/providers/anthropic-direct/pricing.ts
@@ -1,0 +1,162 @@
+/**
+ * Static pricing table + per-call cost derivation for the `anthropic-direct`
+ * provider.
+ *
+ * Extracted from `types.ts` (which sat at ~595 LOC, well over the repo's
+ * 350-line ceiling) so pricing is one concern in one file. `types.ts`
+ * re-exports both symbols, so existing call sites and tests are unaffected.
+ *
+ * @module agent/providers/anthropic-direct/pricing
+ */
+
+/**
+ * Invariant: three billing facts govern this module, all from
+ * https://platform.claude.com/docs/en/build-with-claude/prompt-caching#pricing
+ * (verified 2026-08-05). Changing any of them changes reported dollars, so
+ * they are stated once here rather than re-derived at each use site.
+ *
+ *  1. `input_tokens` from the Messages API counts ONLY tokens that were
+ *     neither read from nor used to create a cache. It does NOT include
+ *     cache reads or cache writes. Total input processed for a call is
+ *     `input_tokens + cache_read_input_tokens + cache_creation_input_tokens`.
+ *     Therefore the plain-input cost term is `input_tokens` verbatim —
+ *     subtracting the cache counts off it (as this code did until 2026-08-05)
+ *     double-subtracts and, because cache reads dwarf plain input in a warm
+ *     session, clamps the term to zero.
+ *  2. Cache WRITES are priced by TTL: 1.25× the base input rate for a 5-minute
+ *     entry, but 2× for a 1-hour entry. A single flat write rate is only
+ *     correct for a caller that never requests 1h — and `cache-policy.ts`
+ *     defaults `AFK_PROMPT_CACHE_TTL` to `1h`, so the flat 1.25× that lived
+ *     here understated the write component of every cached session by 37.5%.
+ *  3. Cache READS are 0.1× the base input rate regardless of TTL.
+ *
+ * The API reports the write split directly on `usage.cache_creation`
+ * (`ephemeral_5m_input_tokens` / `ephemeral_1h_input_tokens`), so cost is
+ * derived from what was actually billed rather than inferred from local
+ * config. `toProviderUsage` supplies that split; the fallback when the field
+ * is absent lives there too, keeping this module pure and env-free.
+ */
+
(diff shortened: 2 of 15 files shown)


griffinwork40:

---

# Code Review — PR #909

**`fix(pricing): correct prompt-cache cost accounting and surface hit rate`**
`fix/prompt-cache-accounting` → `main` · ref `fe013dfe` · 15 files, +709/−106 · regime: **full**

## Verdict: MERGE

No `critical`, no `high` finding survived verification. The five claims in the PR body all land; claim 3 lands partially. The arithmetic is correct, the re-export shim is complete, and the two "corrected" tests were strengthened rather than weakened — all independently confirmed at the ref, not taken from the PR body.

---

## Findings

### medium · spec-compliance · correctness — new model rows are keyed dateless, so claim 3 only half-lands
`src/agent/providers/anthropic-direct/pricing.ts:76-80` · ref `fe013dfe` · citation-type: `diff-context` · confidence: high

The four new rows are keyed `claude-opus-4-8` / `-4-7` / `-4-6` / `claude-sonnet-4-6`, while every other dated-family row in the same table carries its date (`claude-opus-4-5-20250929`, `claude-haiku-4-5-20250929`). `MODEL_PRICING.get(model)` is exact-match and returns `undefined` on a miss (`pricing.ts:136-137`), and `loop.ts:261` passes `input.model` through raw. A user on `--model claude-opus-4-8-20260528` therefore still gets `undefined` cost — the precise outcome claim 3 says these rows fix.

The dated wire ids are not hypothetical: `resolve-effort.test.ts:32,44,49` asserts production behaviour for `claude-opus-4-8-20260528`, `claude-opus-4-6-20250901`, and `claude-sonnet-4-6-20250901`.

The sibling table already solved this. `maxOutputTokensFor` runs `resolveModelInput(model)`, then an exact lookup, then a lowercased fallback, then a typed default (`model-limits.ts:70-82`) — three layers `MODEL_PRICING.get()` has none of. `model-limits.ts:37-42` carries a comment about being bitten by this exact class of miss.

```
  ['claude-opus-4-8', { inputPerMTok: 5.0, outputPerMTok: 25.0, ... }],
  ['claude-opus-4-7', { inputPerMTok: 5.0, outputPerMTok: 25.0, ... }],
  ['claude-opus-4-6', { inputPerMTok: 5.0, outputPerMTok: 25.0, ... }],
```
**Fix:** add the four dated keys alongside the dateless ones, or route the lookup through `resolveModelInput()` first as `model-limits.ts:76` does.

---

### medium · correctness — cache-creation tokens outside the two known TTL buckets bill at $0
`src/agent/providers/anthropic-direct/pricing.ts:157-159` · ref `fe013dfe` · citation-type: `file-state` → **verified at ref** · confidence: high

`cacheWriteCost` is computed exclusively from `split.ephemeral5m` and `split.ephemeral1h`. `cacheCreationTokens` reaches the cost only through the fallback at line 153, when no split is supplied. Any total not covered by the two known buckets is silently unpriced.

Today this is unreachable: `CacheCreation` declares both fields required and non-nullable (`@anthropic-ai/sdk/resources/messages/messages.d.ts:119-128`), and `resolveCacheWriteSplit`'s no-breakdown branch (`types.ts:494-497`) attributes the full total, so it sums exactly. The hazard is forward-looking — if Anthropic adds a third TTL tier the way it added 1h alongside 5m, `cache_creation` arrives populated, the two known fields sum to less than `cache_creation_input_tokens`, and the new tier's tokens cost $0 with no error. No test exercises a divergent split (every `deriveCallCostUsd` call in `pricing.test.ts` passes `cacheCreationTokens=0` alongside an explicit split, or omits the split — grep-confirmed).

```
  const cacheWriteCost =
    (split.ephemeral5m / M) * write5mRate + (split.ephemeral1h / M) * write1hRate;
  const cacheReadCost = (cachedInputTokens / M) * readRate;
```
**Fix:** bill the residual `cacheCreationTokens - (ephemeral5m + ephemeral1h)` at the configured-TTL rate when positive, and pin it with a divergent-split test.

---

### low · security · correctness — wire-sourced token counts enter the cost math unguarded
`src/agent/providers/anthropic-direct/types.ts:486-498` and `pricing.ts:128-135` · ref `fe013dfe` · citation-type: `diff-context` · confidence: medium

`resolveCacheWriteSplit` null-coalesces but never range-checks, and `deriveCallCostUsd` applies no `Number.isFinite`/non-negative guard to any of its five numeric parameters. A negative or `NaN` count produces a negative or `NaN` `totalCostUsd` that flows to `/cost`, the turn footer, and persisted reports.

Reachability is low — the SDK types every one of these as a required `number`, so this needs an API contract violation or a hand-edited trace. Worth noting that `trace.ts` reads trace files with plain `JSON.parse` plus a loose structural guard, not the `ClosurePayloadSchema` validator that only runs writer-side.

```
      ephemeral5m: breakdown.ephemeral_5m_input_tokens ?? 0,
      ephemeral1h: breakdown.ephemeral_1h_input_tokens ?? 0,
```
**Fix:** clamp each count with `Number.isFinite(x) && x >= 0` at the `resolveCacheWriteSplit` boundary — one guard covers both call paths.

---

### low · maintainability — new doc comment mislabels a last-round value as cumulative
`src/cli/slash/commands/info.ts:120-122` · ref `fe013dfe` · citation-type: `diff-context` · confidence: high

The comment calls `usage.totalTokens` "the authoritative cumulative total." It is not. `buildContextUsageFields` sets `totalTokens = contextWindowTokensUsed(last)` (`shared/auto-compact.ts:117`) — the last round's footprint — and `auto-compact.ts:100-105` states it "[d]eliberately does NOT read `ProviderUsage.totalTokens` — that field is provider-dependent and would diverge from the percentage."

This matters because defect #4 in this PR's own body is *"`config-types.ts:711` documented the wrong formula as if it were live."* The rendering at `info.ts:122,131` is correct; only the new comment is wrong.

```
  // provider-computed footprint of the last round; fall back to the
  // authoritative cumulative total rather than reconstructing a bad sum.
  const lastTurnTotal = api?.context_window_tokens ?? usage.totalTokens ?? 0;
```
**Fix:** replace "authoritative cumulative total" with "the same last-round footprint, which `buildContextUsageFields` assigns to both fields."

---

### low · test-coverage — the hit-rate formula is duplicated, and only one copy is tested
`src/cli/slash/commands/info.ts:124` · ref `fe013dfe` · citation-type: `diff-context` · confidence: high

`info.ts:124` reimplements `Math.round((cacheRead / cacheTotal) * 100)` inline rather than calling the tested `formatCacheUsage` (`trace-usage-format.ts:47`). `info.test.ts` exists but has zero matches for `cacheRead|cacheCreate|hitRate|cache_read_input_tokens|cache_creation_input_tokens|cache hit` — absence grep-confirmed at the ref. Two copies of one formula, one pinned by `trace-usage-format.test.ts:11-26` (90%/0%/100%/NaN cases), one unpinned.

```
  const hitRate = cacheTotal > 0 ? Math.round((cacheRead / cacheTotal) * 100) : undefined;
```
**Fix:** export the ratio as a number from `trace-usage-format.ts` and have `info.ts` call it, collapsing two implementations into the tested one.

---

### nit · consistency — the two display sites order the same two fields oppositely
`src/cli/context-sampler.ts:188-191` vs `src/cli/slash/commands/info.ts:122` · ref `fe013dfe` · citation-type: `diff-context` · confidence: high

`resolveUsedTokens` prefers `totalTokens` then `context_window_tokens`; `renderSdkBreakdown` prefers `context_window_tokens` then `totalTokens`. Benign at this ref — `buildContextUsageFields` is the sole producer for both providers (`anthropic-direct/query.ts:810`, `openai-compatible/query.ts:1156`) and assigns the identical `contextWindowTokensUsed(last)` to both fields (`auto-compact.ts:117,123`). Flagged so a future producer that sets them differently does not silently split the status line from `/tokens`.

---

### nit · api-compat — `MODEL_PRICING` row shape changed; no importers affected
`src/agent/providers/anthropic-direct/pricing.ts:56-65` · ref `fe013dfe` · citation-type: `diff-context` · confidence: high · `[UNVERIFIED: no production importers]`

`cacheWritePerMTok` became `cacheWrite5mPerMTok` + `cacheWrite1hPerMTok`. `cacheWritePerMTok` has zero occurrences anywhere in `src/`, and `MODEL_PRICING` has no production consumer beyond the re-export at `types.ts:467`. Non-breaking in-repo; only relevant to an out-of-tree script.

---

## Verification manifest (Wave 1.5, inline against `fe013dfe`)

**4 findings dropped.** All four came from the correctness agent, and all four carried line anchors that do not exist in their files:

| Dropped claim | Anchor | Why dropped |
|---|---|---|
| "pricing.ts contradicts itself on which layer owns the TTL fallback" (medium ×2, filed twice) | `pricing.ts:250-252`, `:259-266`, `:284-287` — file is 162 lines | **Refuted at the ref.** `pricing.ts:33-37` states plainly: *"`toProviderUsage` supplies that split; the fallback when the field is absent lives there too, keeping this module pure and env-free."* The file already names the owning layer; there is no contradiction. Quoted comment text is real but sits at lines 119-121 and 151-153. |
| "no structured JSON counterpart for the hit rate" (medium) | `trace-usage-format.ts:732-742` — file is 56 lines | Self-refuting: its own suggestion says *"This is by design for --json."* Raw NDJSON already carries `cacheRead`/`cacheCreation`. |
| "hit=100% on a tiny denominator misleads" (low) | `trace-usage-format.ts:736-740` — file is 56 lines | Self-refuting: its own suggestion says *"acceptable as shipped."* |

**1 finding re-anchored, content verified:** the surviving medium at `pricing.ts:157-159` was filed at `:145-148`; the quoted evidence is verbatim-correct code at the ref, so it was corrected rather than dropped.

**1 finding downgraded high → medium:** the split-vs-total claim was filed `high`. SDK evidence (`CacheCreation` declares both fields required and non-nullable) shows it is not reachable under the documented API contract; it survives as a forward-compatibility hazard, not a live defect.

**3 findings sharpened past what Wave 1 filed:** the dateless-key gap (nit → medium, once the dated ids were found in `resolve-effort.test.ts` and the `model-limits.ts` precedent surfaced); the info.ts test gap (untested → *duplicated and untested*); and the info.ts doc mislabel, which neither agent caught.

**4 absence claims confirmed** by `git grep` at the ref: `cacheWritePerMTok` (0 matches), `deriveCallCostUsd` production call sites (only `types.ts:531`), `MODEL_PRICING` production importers (none), `info.test.ts` cache coverage (0 matches). **0 false-absent.**

**PR claims verified independently, not accepted from the body:** 5m=1.25× / 1h=2.0× / read=0.1× multipliers correct at `pricing.ts:41-45` and internally consistent with every literal row; plain input uses `inputTokens` verbatim with no subtraction (`pricing.ts:142`) and no other caller pre-subtracts; Opus 4.5 corrected to $5/$25 (`pricing.ts:86`); re-export surface complete (`types.ts:467-469`, `:553`); both "corrected" tests still assert concrete pinned values — `types.test.ts` 800→1000 with a comment naming the bug it had encoded, `status-line-context.test.ts` 1000→500 likewise.

## What was not checked

- **Citations verified inline against branch HEAD `fe013dfe`**, which equals the checked-out worktree HEAD (only untracked `.afk/` dirty), so file reads and the reviewed ref agree.
- **Stated intent: PR #909 title + body — spec-compliance assessed.** Claims 1, 2, 4, 5 MET; claim 3 PARTIAL (finding M1).
- Did not run `pnpm test` / `lint` / `build` — the PR's claimed 14,968-pass and clean-gate results are unverified by me.
- Did not verify runtime behaviour against a live Anthropic response, so whether `usage.cache_creation` ever arrives partially populated in practice is unestablished.
- Did not review `cache-policy.ts` (unchanged, only referenced), the Telegram surface, or out-of-tree consumers of `MODEL_PRICING`.
- The three items the author listed as deliberately deferred (`baseUrl` disabling caching, 1h default for REPL, no anchor breakpoint) were treated as out of scope and not assessed.

**Deviations from the skill's default wave plan, disclosed:** Wave 2 synthesis ran inline rather than as a dispatched agent — every severity call above turned on Wave 1.5's ref-anchored evidence, which a synthesis agent would have had to be handed wholesale to reach the same conclusions. `/shadow-verify` did not fire because no `critical` or `high` finding survived; the single `high` was downgraded on SDK type-declaration evidence, not waived. Wave 1 dispatched 2 agents (peak 2 concurrent); the correctness agent hit its tool-use iteration cap but returned all six required sections.

---

**Done**

- Reviewed PR #909 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability; emitted 7 findings (2 medium, 3 low, 2 nit) and a **MERGE** recommendation.
- Evidence: this report; verification ran against `fe013dfe` using the preflight diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/5862dab6-bc1b-4c11-855f-7e198e738fa7/pr-909.diff`; every surviving citation re-read at the ref via `git show fe013dfe:<file>`.
- Nothing in the world changed — read-only review. No files written, no commits, no PR comments posted, working tree untouched (still only untracked `.afk/`).
- Pending: the two medium findings are unfiled — M1 (dateless model keys) is a 4-line addition to `pricing.ts` and defeats part of claim 3, so it is the one worth landing before or shortly after merge. Say the word and I will open an issue or draft the patch.

---

PR #910, opened 2026-08-05
fix(subagent): tell a forked child its tool-round budget at dispatch

griffinwork40:
## What

Tell a forked subagent its tool-round budget at dispatch, instead of only at wind-down when the budget is already gone.

## Why

The cap meter is tool-use **rounds**, not tool **calls** — a round costs 1 no matter how many calls it batches (both providers increment once per round, after dispatching the whole batch). So a child that batches independent calls gets ~500 tool calls out of its 50-round budget; a child that calls tools one at a time gets ~50.

Nothing told the child any of that:

- `WIND_DOWN_NOTE` is the only budget text that ever reaches model history, and it is appended on the **final** round.
- The `round 7/50` label from `formatRoundLabel` goes to a progress `ProviderEvent` consumed by terminal/UI renderers. `conversationHistory` only ever receives assistant messages, so the model never sees it.
- No other disclosure surface exists: `get_runtime_state`'s `RuntimeSelf` schema has no cap field, the numeric ceilings in `agents/builtins.ts` live only in comments and constants, the `agent`/`compose` tool schemas describe the *dispatcher's* knob rather than the child's resolved value, and named-agent prompts contain zero budget mentions.

Measured consequence — capped children average **1.6–3.0 calls per round**, i.e. running at roughly a tenth of the evidence ceiling their budget actually permits, then getting cut off mid-investigation.

| Telemetry | Value |
|---|---|
| Forks capped | 296 / 4,671 (**6.34%**) |
| Spend, one window | **$990.30** |
| Trend | 5.34% → **13.65%/week** |
| Card `closure-anomaly-iteration-cap` | 153 closures / 90 sessions / **$496.74** |

All 153 closures were forks — top-level defaults to `0` (uncapped), and `AFK_MAX_TOOL_USE_ITERATIONS` is unset by default.

## How

A 697-char preamble appended to the child's system prompt, stating the resolved cap, that a round is one *reply* rather than one *call*, and that the number is a ceiling to finish well under.

No-op when the cap is `0`/unset — that means unbounded, so there is no budget to disclose and claiming one would be a lie.

### Why `forkSubagent` and not `child-config.ts`

`SubagentManager.forkSubagent` is the sole path to a child `AgentSession`. `tools/subagent/child-config.ts` looks like the natural home but is reached by **only the agent-tool paths**:

| Dispatch path | Through `buildChildConfig`? |
|---|---|
| agent tool (fg / bg / worktree-isolated) | yes |
| compose / DAG nodes | **no** — own per-node config |
| skill forks | **no** — separate builder |
| in-process callers (mint phases, user-skills, audit-fit) | **no** — hand-built config |

Injecting at the chokepoint is also provider-neutral, so neither provider has to cooperate. That is deliberate: this repo has already been burned once by exactly that drift, when openai-compatible shipped without the graceful wind-down after anthropic-direct got it.

Shape handling mirrors the existing `companion/primer-loader.ts:injectCompanionPrimer` — same union, same shallow-copy discipline, same no-prompt fallback. The new concern lives in its own 106-LOC module rather than growing `subagent.ts`.

## Verification

- **Failing baseline proven.** With the wiring reverted (`git stash push -- src/agent/subagent.ts`), the two disclosure assertions fail; restored, they pass. A regression test that never fails before the fix would be worthless.
- Full suite: **15,000 passed / 0 failed** (788 files).
- Gates green: `lint`, `build`, `audit:sdk:check`, `audit:env:check`, `scan:env:check`, `audit:chalk:check`, `fix:pins:check`.
- No exact-string prompt assertion broke — injecting downstream of `buildChildConfig` leaves `child-config.test.ts` and `subagent-executor.named-agents.test.ts` untouched.

13 unit tests cover the render text and every union shape; 3 integration tests assert on the config the child session is actually constructed with, which is what pins the chokepoint rather than any single dispatch path.

## Known limitation — please read before assuming this closes the problem

**Disclosure alone is not expected to eliminate cap-burn.** The shadow-verification wave for the underlying diagnosis accidentally ran a controlled experiment on this very fix:

| Wave | Budget | Task shape | Outcome |
|---|---|---|---|
| 2 agents | 18 rounds | narrow — verify 3 named claims | **both finished** |
| 3 agents | 20–22 rounds | broad — "enumerate every X" | **all three capped** |

Same disclosure, same batching instruction, same model; the capped group had *more* budget. The variable that moved was **task breadth**. An unbounded-breadth task will exhaust any budget it is told about.

So this is the cheap, universal first move — expect it to shave the tail, not remove it. The real remaining gap is that **nothing terminates a child whose queries keep varying while its conclusion stops changing**: the repeat circuit breaker needs 8 *consecutive byte-identical* calls and only nudges, `suspected-loop-detector` is explicitly observe-only, and the idle watchdog measures silence while these children are never idle. That is left to a follow-up so this change can be measured on its own.

1. Non-convergence / no-new-evidence detection — the actual mechanism gap.
2. Task-scope discipline at dispatch — likely higher leverage than disclosure, currently unmeasured.
3. Possible bug, unverified: `anthropic-direct` may not strip `WIND_DOWN_NOTE` from its long-lived message array after the wind-down round. `openai-compatible` is structurally safe because it rebuilds `messages` from `priorTurns`. If the note persists, it leaks into resumed history and would tell a resumed session its budget is spent when it is not.

Full diagnosis, including the corrected attribution method: `~/.afk/workspace/reports/cap-burn-diagnosis-2026-08-05.md`

### DIFF
--- a/docs/subagent-tool-budget.md
+++ b/docs/subagent-tool-budget.md
@@ -1,0 +1,87 @@
+# The subagent tool-round budget preamble
+
+Reference for `src/agent/subagent/budget-preamble.ts` — the module that
+discloses a forked child's tool-use-round cap to the child itself.
+
+This document carries the `History:`-class material (root cause, incident
+record, decision log) that the long-comment convention in `AFK.md` keeps out
+of the source. The rounds-vs-calls invariant and the fork-site contract stay
+inline in the code as `Invariant:` / `Contract:` blocks.
+
+## What the budget is
+
+The cap meter is tool-use ROUNDS, not tool CALLS. A round is one assistant
+turn that requests tools, so a turn issuing five parallel calls costs 1
+round, not 5 — `providers/anthropic-direct/loop/tool-round.ts` and
+`providers/openai-compatible/query.ts` both increment once per round, after
+dispatching the whole batch. A child that batches independent calls into one
+reply therefore buys roughly 10x the evidence per unit of budget compared
+with one that calls tools one at a time.
+
+## Why the child was never told
+
+Until this module existed, nothing told the child any of that. The only
+budget text that ever reached the model was `WIND_DOWN_NOTE`
+(`providers/shared/tool-loop-cap.ts`), appended on the FINAL round — after
+the budget was already spent, too late to change how the child paced itself.
+The per-round `round 7/50` label produced by `formatRoundLabel` goes out on a
+progress `ProviderEvent` consumed by terminal/UI renderers, and never enters
+model history. The model that is actually doing the pacing never sees it.
+
+So children paced blind: they averaged 1.6–3.0 tool calls per round, and
+burned the full round budget (50, by default) on the evidentiary equivalent
+of about 50 individual tool calls' worth of work — an order of magnitude
+short of what the same round budget could buy if batched.
+
+## The measured cost
+
+296 of 4,671 forks capped (6.34%), $990 in a single telemetry window, with
+the cap rate climbing week over week. That measurement is what motivated
+writing this module rather than leaving the gap in place.
+
+## Why injection happens at the provider-neutral fork site
+
+This module is deliberately provider-agnostic and is applied at exactly ONE
+site — `SubagentManager.forkSubagent`, the sole path to a child
+`AgentSession`. Injecting here rather than inside a provider is what keeps
+the two providers from drifting apart on this behavior.
+
+The repo has already been burned once by exactly that failure: for a while,
+`openai-compatible` shipped without the graceful wind-down that
+`anthropic-direct` had already gotten, because the wind-down logic lived
+inside the anthropic-direct provider instead of at a shared, provider-neutral
+site. Every provider must render `systemPrompt` to function at all, so
+injecting the budget preamble into `systemPrompt` at the fork site means no
+provider has to separately cooperate, remember, or re-implement anything for
+this to work — the same structural fix as the shared wind-down note.
+
+## Where the block lands in the assembled prompt
+
+`injectToolBudgetPreamble` appends the preamble block after any
+caller-supplied `config.systemPrompt`, so within the fork's OWN prompt
+literal the block sits last — operational trailer rather than mission,
+keeping the child's actual instructions at top salience.
+
+That is "last" only relative to the caller-supplied prompt, not last in the
+prompt actually sent to the model. The fork's `config.systemPrompt` becomes
+the `userSystem` argument to `assembleSystemPrompt`
+(`providers/anthropic-direct/query/system-prompt.ts`, mirrored by the
+equivalent assembly in `providers/openai-compatible/index.ts`), which places
+`userSystem` at position 2 of 6 in the final prompt — after `toolBase`, and
+before `memoryPrompt`, `hotMemory`, the `# Environment` fragment, and the
+skill `manifest`. All of those later sections are appended AFTER the budget
+preamble in what the model actually receives.
+
+## Open gaps
+
+- The runtime still has no non-convergence detector. The preamble's closing
+  sentence ("If new evidence has stopped changing your conclusion, stop
+  gathering and answer now.") is a prose stand-in for that missing
+  mechanism, not a substitute with the same reliability — nothing today
+  actually terminates a child whose queries keep varying while its
+  conclusion stops changing.
+- Post-merge, watch a metric pair, not a single number: the cap rate (forks
+  hitting `tool_use_loop_capped`) should fall, but answer completeness must
+  not fall with it. A drop in cap rate that comes from children truncating
+  their investigation early to stay "under budget" would be a regression
+  wearing the shape of an improvement.
--- a/src/agent/subagent.test.ts
+++ b/src/agent/subagent.test.ts
@@ -280,6 +280,54 @@ describe('SubagentManager', () => {
     expect((shared.lastConfig as unknown as Record<string, unknown>)['maxToolUseIterations']).toBe(0);
   });
 
+  // Regression (cap-burn): a child used to learn its tool-round budget ONLY from
+  // WIND_DOWN_NOTE, on the final round, after the budget was already spent — so
+  // it could not pace itself and averaged 1.6-3.0 calls per round against a
+  // meter that charges per ROUND. Disclosure happens at forkSubagent because
+  // that is the sole path to a child AgentSession; asserting on the config the
+  // child session is actually constructed with is what pins the chokepoint
+  // rather than any one dispatch path.
+  it('discloses the resolved tool-round budget in the child systemPrompt', async () => {
+    shared.lastConfig = null;
+    const mgr = new SubagentManager();
+    await mgr.forkSubagent({
+      parent: { sessionId: 'p' },
+      config: { model: 'sonnet', systemPrompt: 'You are the research agent.' },
+    });
+    const sp = (shared.lastConfig as unknown as Record<string, unknown>)['systemPrompt'];
+    expect(typeof sp).toBe('string');
+    // The DEFAULT cap is disclosed even though no caller passed one.
+    expect(sp as string).toContain(`${SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS} tool-use rounds`);
+    // The caller's own prompt survives, and keeps top salience.
+    expect(sp as string).toMatch(/^You are the research agent\./);
+  });
+
+  it('discloses an explicitly overridden budget rather than the default', async () => {
+    shared.lastConfig = null;
+    const mgr = new SubagentManager();
+    await mgr.forkSubagent({
+      parent: { sessionId: 'p' },
+      config: { model: 'sonnet', maxToolUseIterations: 7 },
+    });
+    const sp = (shared.lastConfig as unknown as Record<string, unknown>)['systemPrompt'];
+    expect(sp as string).toContain('7 tool-use rounds');
+    expect(sp as string).not.toContain(
+      `${SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS} tool-use rounds`,
+    );
+  });
+
+  it('discloses no budget to an explicitly unbounded child', async () => {
+    shared.lastConfig = null;
+    const mgr = new SubagentManager();
+    await mgr.forkSubagent({
+      parent: { sessionId: 'p' },
+      config: { model: 'sonnet', systemPrompt: 'untouched', maxToolUseIterations: 0 },
+    });
+    expect((shared.lastConfig as unknown as Record<string, unknown>)['systemPrompt']).toBe(
+      'untouched',
+    );
+  });
+
   // Anti-hang (sibling of the iteration cap): a fork that hits an OAuth
   // usage-limit 429 must FAIL FAST with the classified error, not silently
   // auto-pause and poll for reset (up to 2h in retry-layer.ts) while its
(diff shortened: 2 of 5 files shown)


griffinwork40:
## Review — PR #910 `fix(subagent): tell a forked child its tool-round budget at dispatch`

**Merge recommendation: Approve.** No blocking defects. The core claim — that the number disclosed to the child is the number actually enforced — holds on every traced path, and all CI gates are green. Four should-fix/nit items below, none gating.

### Gates (run against the working tree)

| Command | Result |
|---|---|
| `pnpm test src/agent/subagent/budget-preamble.test.ts` | PASS (13) |
| `pnpm test src/agent/subagent.test.ts` | PASS (90) |
| `pnpm lint` (`tsc --noEmit`) | PASS |
| `pnpm fix:pins:check` | PASS |
| Collateral: skill-executor (91), subagent-executor (104), named-agents (25), compose-executor (84), child-config (60), dag-subagent (15) | PASS |

No test asserting exact child `systemPrompt` equality broke — every such caller either omits `maxToolUseIterations` (so injection no-ops) or mocks `forkSubagent` outright.

### Correctness — verified, not assumed

- **Disclosure = enforcement.** `subagent.ts:581-582` sets `maxToolUseIterations: options.config.… ?? SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS` *inside* the literal, fully evaluated before `injectToolBudgetPreamble` reads it. Nothing downstream re-resolves it: `session/agent-session.ts` forwards `config` verbatim, and both providers read through to the one shared `resolveMaxToolIterations` (`providers/shared/tool-loop-cap.ts:63-67`).
- **The no-op branch cannot hide a real cap.** `budget-preamble.ts:80`'s guard (absent / 0 / negative / non-finite) is the same predicate family `tool-loop-cap.ts:63-67` maps to *unbounded*. Neither provider substitutes a private nonzero default. Silent-cap risk: not realized.
- **The rounds-vs-calls claim in the prose is true.** `anthropic-direct/loop/tool-round.ts:89` (`turn.iterations += 1`) and `openai-compatible/query.ts:593` (`round += 1`) both fire once per round *after* the whole batch dispatches; call counts are tracked separately at `tool-round.ts:65` / `query.ts:600`.
- **The chokepoint claim is true.** Three `new AgentSession(` sites exist; only `subagent.ts:705` is a fork. agent-tool fg/bg, DAG, compose, skill forks, and mint phases all reach it via `forkSubagent`. `child-config.ts` writes flow in through `...options.config` *before* injection — one injection, no clobber.
- **Live end-to-end confirmation:** the preamble is active in the running binary. A review child independently reported operating under `maxRounds=14` — a number I passed to the runtime but never wrote into its prompt. Only the injected block could have told it.

### Findings

1. **`renderBudgetPreamble` interpolates the cap but hardcodes the scale — should-fix.** `budget-preamble.ts:57`: *"the difference between roughly 50 and roughly 500 tool calls on the same budget"* sits beside `${maxRounds}` at lines 54/59. At `maxRounds=7` the child is told "7 is a hard ceiling" and, one line up, that its budget spans 50–500 calls. Every sub-50 dispatch — which is what `right-size-delegation`, `compose` nodes, and this very review wave all produce — renders self-contradictory text. Fix: `${maxRounds}` / `${maxRounds * 10}`, or drop the absolute numbers ("one call per round versus ten").
2. **Stale citation inside a load-bearing `Invariant:` comment — nit.** `subagent.ts:483` cites "line ~571 below" for the resolved assignment; 571 is `isNonInteractive`, the assignment is at 581-582. The whole justification for wrapping rather than inlining rests on that pointer.
3. **Untagged `History:` content, inline — should-fix.** The 35-line module JSDoc (`budget-preamble.ts:1-35`) is tagged `Invariant:` for the rounds semantics, but lines 12-28 are postmortem: the "$990 / 296 of 4,671 forks (6.34%)" telemetry window and the openai-compatible wind-down drift story. Per the repo convention that migrates to `docs/` with a ≤5-line summary + link. `docs/anthropic-direct-loop.md` already documents this loop and mentions the cap — natural home.
4. **"Sits last, as operational trailer" is true only locally — nit.** `budget-preamble.ts:73-74`. `assembleSystemPrompt` (`anthropic-direct/query/system-prompt.ts:88-108`, mirrored at `openai-compatible/index.ts:357-388`) places `userSystem` at position 2 of 6; environment and manifest follow it. Last within the caller's prompt, not last in the assembled one.

### Notes, not objections

- **AFK.md / CLAUDE.md untouched.** Both return zero hits for `maxToolUseIterations`, `WIND_DOWN_NOTE`, `tool-use rounds` — the cap mechanism was already undocumented, so this is an uncorrected gap, not new provider-style drift. The `agent`/`compose` tool descriptions (`schemas.ts:384-395`, `:558-571`) already state the round semantics and agree with the new child-facing text.
- **`subagent.ts` is 859 LOC** against your hard 350 ceiling, +10 from this PR. Pre-existing, and this change did the right thing — the new concern went into its own 106-line file rather than another branch in `subagent.ts`.
- **The inverse failure is the thing to watch, and no test can catch it.** The module concedes its closing line ("stop gathering and answer now") is a stand-in for a non-convergence detector the runtime lacks. Counter-observation from this session: 2 of 4 review children capped at 14 rounds *while* holding the preamble, and one of them was grep-heavy work that batching would have compressed. n=4 with a deliberately tight budget — it does not refute the fix, but disclosure alone did not visibly change pacing here. Post-merge, the honest metric pair is cap-rate (should fall) against answer completeness (must not).

Nothing was modified, committed, or posted to the PR — review is read-only. Say the word and I'll post this as a PR comment or open a follow-up for items 1–3.

**Done**
- Reviewed PR #910 across correctness, dispatch-path coverage, test integrity, and convention/second-order risk via a 4-agent parallel wave, plus direct verification of the two highest-value findings.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/82f675b8-5deb-439e-b129-3b634cb5029b/pr-910.diff`; gate results reproducible via the four commands in the table above; source citations at `src/agent/subagent.ts:478-487,581-582`, `src/agent/subagent/budget-preamble.ts:57,73-74`, `src/agent/providers/shared/tool-loop-cap.ts:63-67`, `src/agent/providers/anthropic-direct/loop/tool-round.ts:89`, `src/agent/providers/openai-compatible/query.ts:593`.
- World unchanged: no files written, no commits, no PR comment posted, worktree still clean at `d6167a17`.
- Pending: findings 1–3 are unaddressed by design (read-only review); posting the review to GitHub awaits your go-ahead.

---

griffinwork40:
## Review + follow-up fixes (fc0454cd)

Reviewed across correctness, dispatch-path coverage, test integrity, and convention/second-order risk. **No blocking defects** — the central claim holds: the number disclosed to the child is the number actually enforced.

**Verified, not assumed:**

- **Disclosure = enforcement.** `subagent.ts:581-582` resolves `maxToolUseIterations` inside the literal, before `injectToolBudgetPreamble` reads it. Nothing downstream re-resolves: `agent-session.ts` forwards `config` verbatim and both providers read through the single shared `resolveMaxToolIterations` (`providers/shared/tool-loop-cap.ts:63-67`).
- **The no-op branch cannot hide a real cap.** The guard's predicate (absent / 0 / negative / non-finite) is the same family `tool-loop-cap.ts` maps to *unbounded*. Neither provider substitutes a private nonzero default.
- **The rounds-vs-calls claim in the prose is true.** `anthropic-direct/loop/tool-round.ts:89` and `openai-compatible/query.ts:593` both increment once per round after the whole batch dispatches; call counts are tracked separately.
- **The chokepoint claim is true.** Three `new AgentSession(` sites exist; only `subagent.ts:705` is a fork. agent-tool fg/bg, DAG, compose, skill forks, and mint phases all reach it via `forkSubagent`. `child-config.ts` writes flow in through the spread *before* injection — one injection, no clobber.

**One real defect, now fixed.** `renderBudgetPreamble` interpolated the cap twice but hardcoded `roughly 50 and roughly 500 tool calls` in a third spot, so every sub-50 dispatch rendered self-contradictory text — a child given 7 rounds read "7 is a hard ceiling" directly under a 50-500 call scale. That is the common case for `right-size-delegation`, compose nodes, and any explicit `max_tool_use_iterations`. The scale now derives from the cap; two tests pin it (default-50 rendering byte-unchanged, cap=7 renders "roughly 7 and roughly 70").

**Also in fc0454cd, comments and docs only:** replaced a stale `line ~571` citation inside a load-bearing Invariant with a symbolic reference; migrated the History-class narrative (WIND_DOWN_NOTE / formatRoundLabel gap, the 296-of-4,671 measurement, the openai-compatible drift precedent) to `docs/subagent-tool-budget.md` per the long-comment convention; corrected the "sits last" claim, which is true only within the caller's prompt — `assembleSystemPrompt` puts it at position 2 of 6.

**Deliberately not done:** `AFK.md` / `CLAUDE.md` remain unupdated. The cap mechanism was already undocumented in both pre-PR, so this is an uncorrected gap rather than new drift, and expanding scope mid-review is how PRs balloon. Worth a separate pass.

**Watch item, since no test can catch it.** The module concedes its closing sentence is a stand-in for a non-convergence detector the runtime lacks. Counter-observation from the review session itself: 2 of 4 review children capped at 14 rounds *while holding the preamble*, one of them on grep-heavy work that batching would have compressed. n=4 with a deliberately tight budget — it doesn't refute the fix, but disclosure alone did not visibly change pacing there. Post-merge the honest signal is a pair: cap rate should fall, answer completeness must not. Both recorded in the new doc's "Open gaps".

Gates on `fc0454cd`: budget-preamble 15/15, subagent 90/90, `tsc --noEmit` clean, `pnpm build` clean, `fix:pins:check` clean. Collateral suites re-run green (skill-executor 91, subagent-executor 104, named-agents 25, compose-executor 84, child-config 60, dag-subagent 15).

griffinwork40:
### Note on the red `Test (ubuntu-latest)` run in this PR's history

The first CI run on `fc0454cd` failed one test — `background-registry.test.ts:739`, "persists label/hash from the plain prompt and never image base64" — with `ENOENT ... /state/bg/<jobId>/meta.json`, alongside 15,027 passing tests. It is a pre-existing timing flake, not a regression from this branch. Evidence, in increasing order of strength:

1. **Failure mode is wrong for this diff.** The runtime change here is one interpolated string in prompt text. That can only produce a content mismatch, never a missing file.
2. **The test is off this code's path.** `background-registry.test.ts` drives the registry through `createStubHandle`, a hand-rolled stub; the only mention of `SubagentManager` in the file is the comment on line 5 noting the registry's contract deliberately bypasses it. `injectToolBudgetPreamble` is never reached.
3. **The mechanism is a fixed wall-clock sleep.** `registry.register()` kicks off an async `BgJobLogWriter` disk write; the test waits `await new Promise(r => setTimeout(r, 100))` and then calls `fs.readFileSync` — there is no flush handshake. Seven such fixed sleeps (100/150/200 ms) exist in the file. In this run the suite collected 791 files over 218 s; a 100 ms budget for a fire-and-forget write is not guaranteed on a loaded runner. Both 100 ms sleeps are the shortest in the file, and the failure hit the first one.
4. **Same commit, different runner: green.** `Test (macos-latest)` passed the full suite on `fc0454cd`. A deterministic break from the diff would fail on both.
5. **Re-run with zero code changes: green.** All checks now pass on `fc0454cd`. Locally the file passes 3/3 at HEAD.

Not fixed here — it is unrelated to this PR and fixing it would mean touching a file this branch has no business in. The durable fix is to replace those seven fixed sleeps with a poll-until-exists helper (or have `BgJobLogWriter` expose a flush promise the tests can await). Happy to open that as a separate issue.

PR #923, opened 2026-08-06
fix(cli): hold composed tool-lane rows to the shared reading measure

griffinwork40:
## Summary
- `render/measure.ts` establishes an invariant: every unbordered text surface (prose, thinking, subagent text, tool-lane wrapped text) must share one reading measure (`DEFAULT_TEXT_MEASURE = 100`), because adjacent unbordered blocks stopping at different columns read as broken wrapping rather than an intentional edge.
- Composed tool-lane *rows* (built by string concatenation — prefix + args + outcome) were the one surface that didn't honor it: they clamped to raw `getTerminalWidth()` instead of the shared measure. On a terminal wider than 100 cols, the lane's own text stopped at ~100 while its sibling tool rows ran to the screen edge — a 60-column right-edge discontinuity at 160 cols.
- Adds one named read point, `toolLaneWidth()`, in `src/cli/commands/interactive/tool-lane-render.ts` (the module all four render files already import from), and routes all 7 width reads through it instead of `getTerminalWidth()`:
  - `tool-lane.ts:331`
  - `tool-lane-render-grouped-root.ts:118`
  - `tool-lane-render-children.ts:97,357` (parameter defaults)
  - `tool-lane-render-agent.ts:129,212,257`
- A named helper rather than 7 inline `capToMeasure(getTerminalWidth())` calls because two of those sites are a hard pairing: `formatAgentSummary` and `formatAgentHeader` can both emit a head row for the same entry across a session, and the file's own encoding constraint requires their clamp width to agree — otherwise the same entry committed through the two paths would differ whenever the row exceeds the terminal width. They now agree by calling one function, not by both happening to read the terminal the same way.
- `capToMeasure` is a pure `min`, never a widen, so the existing wrap-prevention guarantee is strictly strengthened, not replaced — the row is still clamped to at most the terminal width, just also to the measure when the terminal exceeds it. On terminals at or below the measure (the common 80–100 column case) this is a no-op. `AFK_TEXT_MEASURE=full` restores the previous uncapped behavior for anyone who wants it.

**Before/after at 160 cols:** prose/thinking/tool-lane-text capped at 100 (unchanged); composed tool rows now also capped at 100 (previously ran to 160).

## Test results
- `pnpm lint` (tsc --noEmit): clean.
- `pnpm test` (full suite): 15,142 passed / 1 failed / 2 skipped. The 1 failure is pre-existing and unrelated (see Out of scope).
- Extended the adjacency test in `src/cli/render/measure.test.ts` to cover composed rows, not just the lane's wrapped text — that coverage gap is why the discontinuity went unnoticed (the `describe()` block claims "a new unbordered surface that skips capToMeasure fails here" while only pinning the text path). Verified the new assertion fails without the fix (`expected 200 to be less than or equal to 112`) and passes with it.
- Re-ran the touched test files directly in this session as a cheap re-confirmation before opening this PR (`measure.test.ts` + all `tool-lane*.test.ts`): same result, same single pre-existing failure.

## Out of scope
`src/agent/trace/writer.test.ts > "process-exit backstop seals a real crashed subprocess"` fails on this branch (`expected 1, got null`) but is a **pre-existing failure, not a regression** — confirmed failing identically on clean base `c9374cd2` (current `origin/main` ancestor) with this branch's changes stashed out. It's unrelated to rendering/measure logic and intentionally not touched here.

### DIFF
--- a/src/cli/commands/interactive/tool-lane-render-agent.ts
+++ b/src/cli/commands/interactive/tool-lane-render-agent.ts
@@ -1,9 +1,8 @@
 import { palette } from '../../palette.js';
-import { getTerminalWidth } from '../../terminal-size.js';
 import { formatToolCallStat } from '../../format-utils.js';
 import { MAX_VISIBLE_CHILDREN, batchBadge } from './tool-lane-format.js';
 import type { ToolEntry, Entry } from './tool-lane-render.js';
-import { getGlyphs, clampLineToTerminal } from './tool-lane-render.js';
+import { getGlyphs, clampLineToTerminal, toolLaneWidth } from './tool-lane-render.js';
 import { renderFlushChildren } from './tool-lane-render-children.js';
 
 /**
@@ -126,7 +125,7 @@ function formatAgentSummary(
   // Invariant: ONE width read per render frame, shared by the head row below
   // and the recursive child frame. Two reads could straddle a resize and emit
   // a head row clamped to the old width above children clamped to the new one.
-  const cols = getTerminalWidth();
+  const cols = toolLaneWidth();
   // Clamped for the same reason every child row is: a root nesting entry
   // arrives with an unbounded prefix (the root dispatch path passes no
   // maxWidth), and the stats tail adds ~28 columns on top — a depth-0 `agent`
@@ -209,7 +208,7 @@ function formatAgentHeader(agent: ToolEntry, ancestorIsLast: readonly boolean[] 
   // encoding constraint above. formatAgentSummary clamps its head row, so this
   // one must too, or the same entry committed through the two paths would
   // differ whenever the row exceeds the terminal width.
-  return clampLineToTerminal(ancestorPrefix + head + agent.prefix, getTerminalWidth());
+  return clampLineToTerminal(ancestorPrefix + head + agent.prefix, toolLaneWidth());
 }
 
 /**
@@ -254,7 +253,7 @@ function formatAgentChildren(
     // committed eagerly by formatAgentHeader without the badge (batchSize was
     // unknown then), so the closer is the only completion-time anchor.
     summaryWithBatchBadge(agent),
-    getTerminalWidth(),
+    toolLaneWidth(),
     externalAncestors,
     g,
   );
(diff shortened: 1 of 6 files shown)


griffinwork40:
## PR #923 Review

**Merge recommendation: MERGE**

### Findings

No blocking correctness, security, API-compatibility, performance, or test-coverage defects found.

**Low · high confidence · layout consistency · `src/cli/commands/interactive/tool-lane-render.ts:303`**

On terminals wider than the configured measure, tool-lane text content wraps to approximately 94 columns rather than the 100-column prose/thinking content width:

```ts
const maxWidth = Math.max(1, toolLaneWidth() - indent.length - 2 - 2);
```

This is an intentional and correct trade-off: the remaining columns are reserved for indent, prefix, and safety chrome so the complete rendered row remains within the 100-column budget and `clampLineToTerminal()` does not silently truncate it. It is non-blocking and matches the PR’s whole-row invariant.

### Verification

- Reviewed ref: `8f0574624d6292082a4142607cc52b38d1e54b88`
- Stated intent: PR #923 title and body; spec compliance assessed.
- Citations verified inline against the reviewed ref.
- Reviewed all six changed files plus `measure.ts`, `markdown-stream-format.ts`, and `thinking-paragraph.ts`.
- Confirmed narrow-terminal and `AFK_TEXT_MEASURE=full` behavior remain unchanged.
- The proposed exact-width test concern was rejected because surfaces have different chrome budgets; the shared ceiling, rather than identical content width, is the correct contract.
- The independent width read inside `renderTextChildLines()` predates this PR and is not a regression.
- No tests were executed during this read-only review; the PR reports 15,142 passing tests and one independently reproduced pre-existing trace-writer failure.
- No live visual terminal rendering was performed.

## Done

- PR #923 was reviewed read-only and received a **MERGE** recommendation with no blocking findings.
- Durable review evidence is the inspected diff at `/Users/griffinlong/.afk/state/skill-preflight/91262d57-0622-4e86-be0f-fb23ad670165/pr-923.diff` and reviewed commit `8f0574624d6292082a4142607cc52b38d1e54b88`.
- No repository, PR, or external state was modified.
- Runtime visual verification remains deferred because this review was static and read-only.

---

PR #925, opened 2026-08-06
fix(agents): grant research-agent memory_search so the ground-state surveyor can reach the archive (#883)

### LINKED ISSUE #883: ground-state: memory surveyor is dispatched as an agent type that cannot call memory_search
## Summary

`/ground-state` dispatches its **memory surveyor** as `research-agent`, whose mechanically-enforced tool allowlist does not contain `memory_search`. The memory third of the recon wave therefore cannot reach the cross-session fact archive at all — it silently degrades to reading files.

## Evidence

The skill requires the tool explicitly, twice, and even anticipates this exact failure:

- `ground-state/SKILL.md:15` — "Read-only tools only: `read_file`, `grep`, `glob`, `list_directory`, **`memory_search`**, and read-only bash … `memory_search` is non-mutating and is the **only** way to reach the cross-session fact archive — it is in scope for this skill, **do not strip it from this list**."
- `ground-state/SKILL.md:44` — "Call the **`memory_search` tool** with keywords from the user's current request…"
- `ground-state/SKILL.md:46` — "Restoring a path-grep here **silently zeroes this third of the recon wave**."

But the agent type it lands in cannot comply:

- `src/skills/_agents/research-agent.ts:17` — `allowedTools: ['Read', 'Grep', 'Glob', 'WebFetch', 'WebSearch']`
- `src/skills/_agents/prompts/research-agent.md:4` — `tools: Read, Grep, Glob, WebFetch, WebSearch, Agent(git-investigator)`

Live confirmation from witness trace `2d56a189-2e27-43bb-a50d-c2deeb4cffa3`: the `memory-surveyor` subagent (dispatched with `agentType: research-agent`) returned without querying the archive. The orchestrator had to run `memory_search` itself to get any hits, and reported that the surveyor "had `memory_search` absent from its own tool allowlist and could not query it."

## Why this matters more than it looks

`/ground-state` is the mandated pre-flight for non-trivial implementation work. Every run currently pays for a memory surveyor that is structurally incapable of consulting memory. Worse, the skill's own "report which stores you consulted" requirement (`SKILL.md:48`) makes the gap *look* satisfied when the surveyor dutifully reports `HOT.md: read; AFK.md: read` — a silent-success failure rather than a loud one.

## Constraint on the fix

`research-agent.md` is **byte-locked**: `src/skills/_agents/vendored.test.ts:34` enforces byte-equal snapshots against upstream, so the vendored definition cannot simply gain a tool.

## Options

1. Dispatch the memory surveyor as a different agent type (or the default child surface) whose tool set includes `memory_search`.
2. Grant `memory_search` at the **registration layer** for this dispatch — the same precedent by which `git-investigator` receives `bashReadOnly: true` programmatically in `src/agent/agents/builtins.ts:143` rather than through vendored frontmatter.
3. Require an explicit `memory_search: UNAVAILABLE` line from the surveyor when the tool is missing, so the gap is loud. Mitigation only, and note `SKILL.md:48` already asks for this and it did not surface — so this is insufficient alone.

Option 2 matches existing precedent and leaves the vendored file untouched.

## Acceptance criteria

- A `/ground-state` run reports real `memory_search` query counts from the surveyor itself, not from the orchestrator.
- A test asserts the memory-surveyor dispatch carries `memory_search` in its effective tool set.

## Provenance

Surfaced by `/ground-state`'s own self-report during a session on 2026-08-03, then verified independently against the vendored allowlist and the witness trace.

griffinwork40:
Closes #883.

## What was wrong

`/ground-state` dispatches its **memory surveyor** as `research-agent`, whose mechanically-enforced allowlist contained no `memory_search`. The memory third of the mandated pre-flight recon could not reach the cross-session fact archive at all.

The failure mode is what makes this worth fixing rather than documenting: it is **silent**. The surveyor degrades to reading `HOT.md` / `AFK.md` off disk, and the skill asks it to "report which stores you consulted" — so it truthfully reports `HOT.md: read; AFK.md: read` and the gap *looks* satisfied. Every `/ground-state` run has been paying for a memory surveyor structurally incapable of consulting memory.

## Mechanism (verified, not inferred)

The child pools are not the problem — both `CHILD_ALLOWED_TOOLS` and `RECON_ALLOWED_TOOLS` already include `memory_search` (`nesting.ts:132`, `nesting.ts:152-170`), and `nesting.ts:124-131` argues explicitly that excluding it while the *more* restricted `READ_ONLY_PHASE_TOOLS` trusts it would be incoherent.

The gap is one level down. `/ground-state`'s own fork gets `RECON_ALLOWED_TOOLS` and can reach memory; the **nested surveyor** it dispatches resolves against the `research-agent` registry entry, which declared only `[...researchAgent.allowedTools, 'Agent(git-investigator)']`. Resolution intersects declared tokens with the pool, so a tool absent from the declaration is unreachable no matter how permissive the pool is.

## The fix

One token, at the registry entry in `builtins.ts`:

```diff
-        tools: [...researchAgent.allowedTools, 'Agent(git-investigator)'],
+        tools: [...researchAgent.allowedTools, 'memory_search', 'Agent(git-investigator)'],
```

This is the issue's **Option 2**, and it turns out the codebase had already built the precedent: the adjacent `Agent(git-investigator)` grant lives at this exact spot, with a comment explaining that it must *not* go in the shared `researchAgent.allowedTools` const — because diagnose's read-only leaf surface and audit-fit's `inspectorTools` gate both derive from that const. Same reasoning applies here, so the grant lands in the same place and those gates are unchanged.

The byte-locked vendored prompt (`prompts/research-agent.md`) is untouched; `vendored.test.ts` stays green.

`memory_update` and `procedure_write` are deliberately **not** granted — they mutate, and `target: "hot"` reaches every future session's system prompt.

## Tests

`registry.test.ts` — `research-agent can reach memory_search on every child surface it is dispatched from (#883)`.

It asserts the **resolved** access set from `resolveAgentToolAccess`, not the static `tools` array, because resolution intersects declared tokens against the runtime pool — a test on the declaration alone would pass while the tool stayed unreachable. It runs both surfaces a surveyor can land on (`RECON_ALLOWED_TOOLS` for the read-only skill child, `CHILD_ALLOWED_TOOLS` for an ordinary parent) and also asserts the grant did not widen the read-only contract (`write_file`, `edit_file`, `bash`, `memory_update` all still absent).

**Verified the test actually catches the bug**: stashing the one-line fix and re-running produces `memory_search unreachable on read-only skill child (ground-state)` — 2 failed / 22 passed. The pre-existing exact-array assertion at `registry.test.ts:49` was the blast radius and is updated.

## Gates

- `pnpm lint` — clean
- `pnpm test src/agent/agents/registry.test.ts` — 24 passed
- `pnpm test` on `vendored.test.ts`, `resolve.test.ts`, `tool-def.test.ts`, `subagent-executor.named-agents.test.ts`, `nesting.test.ts` — **112 passed**, byte-lock intact

## Scope note for the reviewer

Acceptance criterion 1 ("a real run reports real query counts from the surveyor itself") is **proven only at the tool-access layer**, not by an end-to-end `/ground-state` run — the test proves the surveyor *can* call `memory_search`; it does not prove a live model *will*. In practice the dispatch prompt instructs it to (`SKILL.md:44`) and the tool now appears in the child's advertised schema, so the mechanical blocker is removed.

One deliberate omission: `researchAgent.description` still reads "Mechanically locked to Read, Grep, Glob, WebFetch, WebSearch". It is left unchanged because it mirrors the byte-locked `.md` frontmatter verbatim, and the existing `Agent(git-investigator)` grant is undocumented there for the same reason. Worth a follow-up deciding whether the registry-granted tools should be reflected in the description shown to orchestrating models.

### DIFF
--- a/src/agent/agents/builtins.ts
+++ b/src/agent/agents/builtins.ts
@@ -116,7 +116,23 @@ export function builtinAgents(): Map<string, RegisteredAgent> {
         // `nestedAgentTypes`, and the subagent executor mechanically restricts
         // research-agent to dispatching ONLY git-investigator (no bare/other
         // dispatch), so the read-only contract can't be escalated.
-        tools: [...researchAgent.allowedTools, 'Agent(git-investigator)'],
+        //
+        // `memory_search` is granted by the same registry-entry mechanism, for
+        // the same reason it must not go in the shared const. Without it the
+        // cross-session fact archive is unreachable from a research-agent, and
+        // the failure is SILENT: `/ground-state` dispatches its memory surveyor
+        // as this type and instructs it to "call the memory_search tool", so the
+        // surveyor degrades to reading HOT.md/AFK.md off disk and then truthfully
+        // reports which stores it consulted — a report that looks satisfied while
+        // a third of the recon wave never happened (issue #883). The tool is
+        // read-only by construction and already trusted by the MOST restricted
+        // role in the system (READ_ONLY_PHASE_TOOLS, mint's spec/research/plan
+        // phases) and by RECON_ALLOWED_TOOLS, so withholding it from a read-only
+        // research agent was incoherent rather than protective — the same
+        // argument CHILD_ALLOWED_TOOLS already makes at nesting.ts:124-131.
+        // Note `memory_update`/`procedure_write` are deliberately NOT granted:
+        // those mutate, and target:"hot" reaches every future session's prompt.
+        tools: [...researchAgent.allowedTools, 'memory_search', 'Agent(git-investigator)'],
         // Anti-runaway bound (see READONLY_AGENT_MAX_TOOL_USE_ITERATIONS): a
         // read-only research/review agent that keeps tool-calling without ever
         // emitting a final message otherwise runs unbounded on the (uncapped)
--- a/src/agent/agents/registry.test.ts
+++ b/src/agent/agents/registry.test.ts
@@ -10,6 +10,8 @@ import { join } from 'node:path';
 
 import { loadAgentRegistry } from './registry.js';
 import { builtinAgents } from './builtins.js';
+import { resolveAgentToolAccess } from './resolve.js';
+import { CHILD_ALLOWED_TOOLS, RECON_ALLOWED_TOOLS } from '../tools/nesting.js';
 
 let tmp: string;
 let prevAfkHome: string | undefined;
@@ -42,16 +44,19 @@ describe('loadAgentRegistry', () => {
     expect(registry.get('git-investigator')?.bashReadOnly).toBe(true);
     expect(registry.get('general-purpose')?.source).toBe('builtin');
     expect(registry.get('Explore')?.source).toBe('builtin');
-    // research-agent keeps its vendored read-only contract PLUS the scoped
-    // git-investigator dispatch grant (matches the vendored prompt frontmatter;
-    // resolve.ts turns Agent(git-investigator) into nestedAgentTypes and the
-    // executor restricts research-agent to dispatching only git-investigator).
+    // research-agent keeps its vendored read-only contract PLUS two
+    // registry-entry grants: `memory_search` (read-only archive access — see
+    // the #883 test below) and the scoped git-investigator dispatch grant
+    // (matches the vendored prompt frontmatter; resolve.ts turns
+    // Agent(git-investigator) into nestedAgentTypes and the executor restricts
+    // research-agent to dispatching only git-investigator).
     expect(registry.get('research-agent')?.definition.tools).toEqual([
       'Read',
       'Grep',
       'Glob',
       'WebFetch',
       'WebSearch',
+      'memory_search',
       'Agent(git-investigator)',
     ]);
     // Anti-runaway bound: the read-only research/review builtins carry an
@@ -103,6 +108,35 @@ describe('loadAgentRegistry', () => {
     }
     expect(sawRestricted, 'expected ≥1 restricted read-only builtin').toBe(true);
     expect(sawInheritAll, 'expected the inherit-all worker (general-purpose)').toBe(true);
+  });
+
+  // Regression guard for #883. Asserting the registry's static `tools` array
+  // is not enough: what the dispatcher enforces is the RESOLVED access set,
+  // and resolution intersects the declared tokens with the child's runtime
+  // pool — so a tool can be declared and still be unreachable if the pool
+  // omits it. This pins the effective set on BOTH child surfaces a surveyor
+  // can actually land on. `/ground-state` is enforced read-only by name
+  // (RECON_ALLOWED_TOOLS); a research-agent dispatched from an ordinary parent
+  // gets CHILD_ALLOWED_TOOLS. memory_search must survive both, or the memory
+  // third of the recon wave silently degrades to reading files off disk.
+  it('research-agent can reach memory_search on every child surface it is dispatched from (#883)', () => {
+    const registry = loadAgentRegistry({ cwd: join(tmp, 'proj'), warn: () => {} });
+    const entry = registry.get('research-agent');
+    expect(entry).toBeDefined();
+
+    for (const [surface, pool] of [
+      ['read-only skill child (ground-state)', RECON_ALLOWED_TOOLS],
+      ['ordinary child', CHILD_ALLOWED_TOOLS],
+    ] as const) {
+      const resolved = resolveAgentToolAccess(entry!, [...pool]);
+      expect(resolved.allowedTools, `memory_search unreachable on ${surface}`).toContain(
+        'memory_search',
+      );
+      // The grant must not have widened the read-only contract on the way in.
+      for (const forbidden of ['write_file', 'edit_file', 'bash', 'memory_update']) {
+        expect(resolved.allowedTools, `${forbidden} leaked on ${surface}`).not.toContain(forbidden);
+      }
+    }
   });
 
   it('strips vendored frontmatter from builtin prompts (body-only system prompt)', () => {


griffinwork40:
Review complete. All four dimensions returned, and I verified the two highest-stakes claims first-hand.

---

# Review: PR #925 — `research-agent` gets `memory_search`

**Status reframe:** #925 is **already MERGED** (`5f227a4c`, 2026-08-06 23:49Z). This is a retrospective review, so the deliverable is follow-up actions, not a merge gate.

**Verdict: the change is correct and safe — but it did not fix #883 on its own.** It closed one of three layers. A second was closed by the open follow-up PR #928. The third is still open.

## What is sound

| Check | Result |
|---|---|
| `pnpm test src/agent/agents/registry.test.ts` | PASS (25/25) |
| `pnpm lint` (tsc strict) | PASS, clean |
| `pnpm fix:pins:check` | PASS — vendored `.md` untouched |
| Stale-assertion sweep (8 test files, 267 tests) | Nothing stale |
| Read-only contract widened? | **No** |

Security is clean and independently confirmed. `memory_search` has no path/file param and only issues `SELECT`s (`memory-tools.ts:37-59`, `memory-store.ts:677-711`). `memory_update`/`procedure_write` are unreachable — they're absent from all three tool-normalization universes (`resolve.ts:43-49`, `to-definition.ts:14-20`, `tool-injector.ts:385-391`), so no `tools:` declaration can ever grant them. Resolution is intersect-only, never union (`child-config.ts:191-198`), so `Agent(git-investigator)` nesting can't smuggle it downward. The "already trusted by more restricted roles" argument in the comment checks out: `memory_search` is already in `CHILD_ALLOWED_TOOLS` (`nesting.ts:132`), `RECON_ALLOWED_TOOLS` (`nesting.ts:163`), and `READ_ONLY_PHASE_TOOLS` (`tool-category.ts:211`).

## Findings

**CRITICAL — the grant was behaviorally inert; the dispatcher-facing description is *still* stale.**
Three layers gate whether a memory surveyor can reach the archive:

1. **Mechanical grant** — fixed by #925. ✅
2. **Agent's self-knowledge** — `research-agent.md:9` told the agent "Your tool surface is a **hard allowlist**… `Read, Grep, Glob, WebFetch, WebSearch`." An agent that believes that won't attempt tool #6. Closed by **PR #928** (open, `9c6797bf`, this branch) via `withRegistryGrants`. ✅
3. **Dispatcher-facing description** — **STILL STALE.** `src/skills/_agents/research-agent.ts:19` reads *"Mechanically locked to Read, Grep, Glob, WebFetch, WebSearch"*, and `builtins.ts:145` passes it straight through as the agent's `description`. That string is what a parent LLM reads when *choosing* an agent type. First-hand evidence: it is verbatim in my own system prompt in this session. A parent needing memory context will not select `research-agent` — so #883's causal chain is still cut at selection time, one step earlier than either PR looked. ❌

This one is **cheap and unblocked**: only the `.md` prompt is SHA-pinned (`vendored.test.ts:13` hashes `readPrompt` output). `research-agent.ts` is a wrapper `.ts`, and **no test asserts the description string** — verified. It can be edited without a pin rewrite.

**HIGH — the code comment's central causal claim is unsupported by source.**
`builtins.ts` asserts "`/ground-state` dispatches its memory surveyor as this type." It does not. `src/bundled-plugins/awa-bundled/skills/ground-state/SKILL.md` contains zero occurrences of `research-agent`, `agent_type`, or `subagent_type` (current file and full `--follow` history) — it only says "dispatch three parallel reconnaissance sub-agents." The agent type is an unconstrained runtime choice. Related: SKILL.md:15 lists read-only bash as an in-scope surveyor tool, which `research-agent` doesn't have at all — so it's an imperfect fit for the surveyor role beyond just memory.

**MEDIUM — the `#883` regression test's stated rationale is factually wrong.**
The comment claims resolution "intersects the declared tokens with the child's runtime pool." It does not: `resolve.ts:228` reads `pool = allowed?.names ?? [...inheritPool]`, so when `tools` is declared (always, for research-agent) `inheritPool` is **never consulted**. Proven empirically — resolving against `RECON_ALLOWED_TOOLS`, `CHILD_ALLOWED_TOOLS`, `[]`, and a garbage pool `['totally_fake_tool_xyz']` all yield identical output. Consequences: the two-surface parametrization is vacuous scaffolding, and the real intersection point (`child-config.ts:191-198`) is never exercised. The test *is* still a real guard — it fails on revert — just not for the reason it says.

**LOW — the forbidden-tool loop is a fixture sanity check, not a regression guard.** `write_file`/`edit_file`/`bash`/`memory_update` were never in the source arrays; no tested path could introduce them.

**NIT — "the failure is SILENT" overstates it.** SKILL.md:48 already instructs the surveyor to say so explicitly when `memory_search` is unavailable.

## Recommended follow-ups

1. **Fold into PR #928 before merge** (same branch, same concern, zero pin risk): update the description at `research-agent.ts:19`. #928's own thesis — "tell research-agent about its registry grants, not just grant them" — applies equally to telling *callers*. Merging #928 without this leaves #883 half-fixed at the selection layer.
2. **Correct the comment in `builtins.ts`** — either pin the surveyor's `agent_type` explicitly in ground-state's SKILL.md (making the claim true), or soften the claim to "a surveyor dispatched as this type."
3. **Fix the test comment** at `registry.test.ts:73-79`, or move the assertion to `buildChildConfig` where intersection actually happens.
4. Consider whether `research-agent` is the right surveyor type at all, given SKILL.md:15 wants read-only bash.

Read-only review: no files edited, no PR comments posted, working tree untouched (still just the pre-existing untracked `.afk/`).

---

**Done**
- Reviewed merged PR #925 across four parallel dimensions (correctness, security, test quality, doc drift), then independently verified the two load-bearing claims — PR merge state and the ground-state dispatch claim — first-hand.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/e08ca9c6-8d1e-4631-8426-f0174a015ad0/pr-925.diff`; `gh pr view 925` → MERGED at `5f227a4c`; `gh pr list` → follow-up PR #928 OPEN on `fix/research-agent-prompt-grant-reconciliation` @ `9c6797bf`; test/lint/pin gates run green by the test agent; stale string confirmed at `src/skills/_agents/research-agent.ts:19` and in this session's own system prompt.
- Nothing changed in the world — read-only review, no edits, no commits, no PR comments; the working tree's single pre-existing untracked `.afk/` is untouched.
- Deferred: the four recommended follow-ups are unapplied, since `/review` is read-only by contract; the highest-value one (item 1) belongs in open PR #928 and needs your go-ahead before I touch that branch.

---

PR #927, opened 2026-08-07
fix(worktree): stop generated build artifacts and symlinked dep trees from making worktrees immortal

griffinwork40:
Follow-up to #926, which made the preserve message name the entry that held a worktree instead of citing a generic `.env`. Pointing it at the real worktrees immediately exposed **two more false positives of the same class** — both permanent, both invisible while the message was generic, neither involving an actual secret.

```
# after #926, still preserved on every exit:
PRESERVED  afk-improve-tui-spacing   (coverage/src/agent/auth/credential-resolver.ts.html)
PRESERVED  afk-bash-parallel-concurrency   (dist/agent/auth/credential-resolver.d.ts)
PRESERVED  review-pr-872   (node_modules)          ← no trailing slash
```

## 1. Generated artifacts mirror source filenames

This repo has 8 source files named `*credential*`/`*secret*`. A build emits `dist/agent/auth/credential-resolver.d.ts`; a coverage pass emits `coverage/src/agent/auth/credential-resolver.ts.html`. Both matched the unanchored `/credential/` sensitivity pattern → `protected` → **every worktree that had ever run `pnpm build` or `pnpm test:coverage` was preserved forever.**

The fix splits the sensitivity table by what each half actually asserts:

| | claim | scope |
|---|---|---|
| **Format anchors** — `.env` `.pem` `.key` `.db` … | "this **is** a credential" | protects from anywhere |
| **Name hints** — `credential` `secret` | "this is **about** credentials" | suppressed on a generated artifact |

"Generated artifact" is the **intersection** of living under a rebuildable directory **and** carrying a compiler/renderer extension (`.js` `.d.ts` `.map` `.html` `.css` …). Neither condition alone is enough:

- *Directory alone* would reap a hand-placed `dist/nested/app-credentials.json` — which #759 deliberately protects and pins with a test.
- *Extension alone* would reap an authored `credentials.html` at the repo root.

Data formats (`.json`, `.yaml`, `.txt`) stay out of the extension list on purpose: no toolchain name-mirrors a source module into them, so excluding them costs nothing here and keeps #759's guarantee intact.

## 2. Symlinked dependency trees

`git status --ignored` appends a trailing slash only for a **real** directory. A symlinked one — `node_modules -> ../../node_modules`, the standard way a worktree avoids a duplicate install — arrives as bare `node_modules`, matches no directory pattern (all require `\/`), and falls through to `protected`.

Widening the patterns to accept a bare segment is the obvious fix and is **wrong**: from the string alone, a symlinked `dist` is indistinguishable from a hand-authored *file* named `dist`, which an existing test pins as protected. Only the filesystem can separate them — so the disambiguation lives in the **probe (IO)**, not the patterns module (pure policy), preserving that split.

The `stat` is paid only by an entry that would *otherwise be protected*, so expanding a populated `dist/` does not become thousands of syscalls. A broken link, a race, or a permission error keeps the protective verdict, matching the module's fail-safe direction.

## Effect

Worktrees held by ignored state on this checkout: **29 of 29 → 2 of 29**. Both survivors are correct — each holds a real `.afk/` project-state dir.

## Verification

- `pnpm lint`, `pnpm build`, `pnpm test` (800 files / 15278 tests), `audit:sdk:check`, `audit:env:check`, `scan:env:check`, `audit:chalk:check`, `fix:pins:check`, `audit:deps` — all green.
- Re-ran the real probe against all 29 local worktrees before and after.
- New tests: generated mirrors across `.d.ts`/`.js`/`.map`/`.html`/`.css` are reapable; format anchors and data files under build output stay protected; name hints outside generated output are untouched; four filesystem-backed cases cover the symlink split (symlinked `node_modules`, symlinked `dist`, an authored file sharing a build-directory name, and an un-stat-able entry).

---

## Review follow-up (commit `51d64b2a`)

A `/review` pass on this PR found the name-hint suppression reaching further than the description above admitted. Running the pre-change and as-merged classifiers side by side over a path corpus showed **eight paths flipping `protected` → reapable, none of them asserted by any test**:

```
dist/secret-notes.js      logs/credentials.js       logs/secret-audit.html
out/client_secret.js      dist/.secrets.js          coverage/credentials.css
dist/my-secret.ts         dist/secret.env.html
```

Two production consumers reach this classifier, not one — `worktree-sweep.ts:721` and `worktree-orphan-guard.ts:40`, the latter's own comment stating it relies on the `protected` default to catch "a stray secret".

**The seam is survivability of a hand-placed file.** `dist/`, `out/`, `build/`, `target/`, `coverage/`, `node_modules/` are regenerated wholesale by the next build — a file dropped there is already doomed independently of the sweep. `logs/` is the one member of `INSPECTABLE_REBUILDABLE_DIRS` where that is false: nothing regenerates it, and the table's own docblock admits it plausibly holds "a captured transcript" someone kept.

So:
- `RUNTIME_OUTPUT_DIRS` excludes `logs/` from `isGeneratedArtifact` — `logs/credentials.js`, `logs/secret-audit.html`, `logs/secret.map` are protected again, with tests.
- The remaining six are **pinned by assertion and documented as accepted risk** in the function docblock, so the blast radius can never widen silently.

Both motivating fixes are untouched, and every previously pinned invariant is unmoved (#759's `dist/nested/app-credentials.json`, the data formats, the authored-name hints, the bare files named like build dirs).

### Known gap, deliberately not fixed here
`worktree-orphan-guard.ts:40` calls `classifyIgnoredEntry` directly and so bypasses the symlink-aware path added in this PR — an orphan directory holding a symlinked `node_modules` stays immortal. Separate lifecycle, separate change.

### DIFF
--- a/src/agent/worktree-ignored-patterns.test.ts
+++ b/src/agent/worktree-ignored-patterns.test.ts
@@ -143,6 +143,37 @@ describe('classifyIgnoredEntry — leaf-matching does not over-reach', () => {
     });
   }
 
+  // `logs/` is rebuildable-listed but is RUNTIME output, not compiler output:
+  // nothing regenerates it, so a file a human kept there survives builds and
+  // must survive the sweep. Suppressing name hints inside it traded a permanent
+  // false positive for a permanent data-loss risk.
+  const runtimeOutput = ['logs/credentials.js', 'logs/secret-audit.html', 'logs/secret.map'];
+  for (const entry of runtimeOutput) {
+    it(`protects ${entry} — logs/ is runtime output, not a generated mirror`, () => {
+      expect(classifyIgnoredEntry(entry)).toBe('protected');
+    });
+  }
+
+  // Invariant: the accepted blast radius of the name-hint suppression, pinned
+  // so it can never widen silently. Each of these was `protected` before the
+  // suppression existed and is reapable now. They are compiler output — the
+  // next build deletes them regardless, so the sweep is not what destroys
+  // them — but the set is asserted rather than incidental, and a credential in
+  // a DATA format stays protected everywhere (see dataUnderBuildOutput above).
+  const acceptedReapable = [
+    'dist/secret-notes.js',
+    'out/client_secret.js',
+    'dist/.secrets.js',
+    'coverage/credentials.css',
+    'dist/my-secret.ts',
+    'dist/secret.env.html',
+  ];
+  for (const entry of acceptedReapable) {
+    it(`accepts ${entry} as reapable compiler output (pinned blast radius)`, () => {
+      expect(classifyIgnoredEntry(entry)).not.toBe('protected');
+    });
+  }
+
   // Outside generated output the name hints are untouched — that is where a
   // credential-shaped name is actually evidence of one.
   const authoredHints = ['credentials.json', 'src/secrets.txt', 'credentials.html', '.secrets'];
--- a/src/agent/worktree-ignored-patterns.test.ts
+++ b/src/agent/worktree-ignored-patterns.test.ts
@@ -90,6 +90,67 @@ describe('classifyIgnoredEntry — leaf-matching does not over-reach', () => {
       expect(isSensitiveLeaf(entry)).toBe(true);
     });
   }
+
+  // Regression: every build artifact echoes the name of the source file it was
+  // produced from. A repo holding `src/agent/auth/credential-resolver.ts` emits
+  // `dist/agent/auth/credential-resolver.d.ts` and a coverage page of the same
+  // name — both matched `/credential/`, protected the tree, and made every
+  // worktree that had run a build or a coverage pass permanently unreapable,
+  // citing a "secret" that was compiler output. The suppression is scoped to
+  // the DIRECTORY, not to an extension list: the mirror follows the toolchain.
+  const generatedMirrors = [
+    'dist/agent/auth/credential-resolver.d.ts',
+    'dist/agent/auth/credential-resolver.js',
+    'dist/agent/auth/credential-resolver.d.ts.map',
+    'coverage/src/agent/auth/credential-resolver.ts.html',
+    'coverage/src/agent/redact-secrets.ts.html',
+    'node_modules/@scope/secret-box/index.js',
+    'dist/docs/secret-handling.css',
+  ];
+  for (const entry of generatedMirrors) {
+    it(`treats ${entry} as generated output, not a secret`, () => {
+      expect(classifyIgnoredEntry(entry)).not.toBe('protected');
+    });
+  }
+
+  // The suppression reaches ONLY the name hints, and only on an artifact
+  // extension. A credential FORMAT protects the tree from anywhere, and a data
+  // file under build output keeps the guarantee #759 pinned — no toolchain
+  // name-mirrors a source module into `.json`, so excluding it costs nothing.
+  const formatAnchored = [
+    'coverage/prod.env',
+    'dist/app.env',
+    'coverage/src/tls.pem',
+    'dist/deploy.key',
+    'coverage/secret.db',
+    'dist/cache/store.sqlite3',
+  ];
+  for (const entry of formatAnchored) {
+    it(`still protects ${entry} despite living under generated output`, () => {
+      expect(isSensitiveLeaf(entry)).toBe(true);
+      expect(classifyIgnoredEntry(entry)).toBe('protected');
+    });
+  }
+
+  const dataUnderBuildOutput = [
+    'dist/nested/app-credentials.json',
+    'dist/secrets.yaml',
+    'coverage/credentials.txt',
+  ];
+  for (const entry of dataUnderBuildOutput) {
+    it(`keeps ${entry} protected — a data file is not a generated mirror`, () => {
+      expect(classifyIgnoredEntry(entry)).toBe('protected');
+    });
+  }
+
+  // Outside generated output the name hints are untouched — that is where a
+  // credential-shaped name is actually evidence of one.
+  const authoredHints = ['credentials.json', 'src/secrets.txt', 'credentials.html', '.secrets'];
+  for (const entry of authoredHints) {
+    it(`keeps ${entry} protected outside generated output`, () => {
+      expect(classifyIgnoredEntry(entry)).toBe('protected');
+    });
+  }
 });
 
 /**
--- a/src/agent/worktree-ignored-patterns.test.ts
+++ b/src/agent/worktree-ignored-patterns.test.ts
@@ -96,19 +96,28 @@ describe('classifyIgnoredEntry — leaf-matching does not over-reach', () => {
   // `dist/agent/auth/credential-resolver.d.ts` and a coverage page of the same
   // name — both matched `/credential/`, protected the tree, and made every
   // worktree that had run a build or a coverage pass permanently unreapable,
-  // citing a "secret" that was compiler output. The suppression is scoped to
-  // the DIRECTORY, not to an extension list: the mirror follows the toolchain.
+  // citing a "secret" that was compiler output. The suppression fires only at
+  // the INTERSECTION of both conditions — the entry sits under a rebuildable
+  // DIRECTORY *and* its leaf carries an artifact EXTENSION — so the mirror
+  // follows the toolchain without reaching a data file dropped alongside it.
   const generatedMirrors = [
     'dist/agent/auth/credential-resolver.d.ts',
     'dist/agent/auth/credential-resolver.js',
     'dist/agent/auth/credential-resolver.d.ts.map',
     'coverage/src/agent/auth/credential-resolver.ts.html',
     'coverage/src/agent/redact-secrets.ts.html',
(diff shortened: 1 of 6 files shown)


griffinwork40 (commented):

The two fixes do what the description says — I reproduced the intended wins — but both halves overshoot into paths that were `protected` before this PR and are `reapable` after it. This module gates `git worktree remove --force` (`src/agent/worktree-sweep.ts:852`) and `fs.rm(..., { recursive: true, force: true })` (`:609`), so a wrong `reapable` verdict is unrecoverable deletion, not a cosmetic misclassification.

Findings were verified by executing the policy and probe modules **as they exist at the reviewed ref** (`3b3b1ae`) against a real filesystem, and by re-running the same inputs through `main` to separate regressions from pre-existing behavior.

**Verified working as described:** `dist/…/credential-resolver.d.ts`, `coverage/…/credential-resolver.ts.html`, and `node_modules/@scope/secret-box/index.js` all become reapable; `dist/nested/app-credentials.json`, `dist/secrets.yaml`, `coverage/credentials.txt` stay protected (#759 intact); format anchors (`dist/app.env`, `coverage/src/tls.pem`, `logs/prod.key`) protect from anywhere. The `stat` cost claim also holds — `probe.ts:110` early-returns unless the verdict is already `protected`.

---

### 1. critical · high confidence · correctness/security · `src/agent/worktree-ignored-probe.ts:196-207` · ref:3b3b1ae · citation-type:file-state

**finding:** A symlinked build directory is resolved to `inspectable` but never expanded, so a secret hidden behind the symlink is never seen and the worktree is reported reapable. `classifyResolvingSymlinkedDirs` returns only the verdict — the slash-appended string it builds internally at line 113 is discarded — so the expansion guard at line 201 re-tests the *original* slash-less `entry` and evaluates `true && false`.

```
197:    const verdict = await classifyResolvingSymlinkedDirs(worktreePath, entry);
201:    if (verdict === 'inspectable' && entry.endsWith('/')) {
```

**Reproduced** against the real probe with a real symlink (`wt/dist -> ../real-build`, `real-build/prod.env` present):

```
symlinked dist, git prints "!! dist"   -> {"protect":false}    ← secret never found
real dist,      git prints "!! dist/"  -> {"protect":true, detail:"dist/prod.env"}
```

This is a regression, not a pre-existing gap: on `main`, bare `dist` classifies `protected`, so the tree survived. This PR converts that into a silent delete, and it defeats the exact #759 invariant the description commits to preserving ("*a symlinked `dist` is indistinguishable from a hand-authored file named `dist`*" is handled; "*`dist/` must still be expanded*" is not).

**suggestion:** Have `classifyResolvingSymlinkedDirs` return the resolved entry string alongside the verdict (e.g. `{ verdict, entry: didResolve ? entry + '/' : entry }`) and gate both the expansion at `:201` and the recursive call at `:171` on that resolved string.

---

### 2. high · high confidence · test-coverage · `src/agent/worktree-ignored-probe.test.ts:338-344` · ref:3b3b1ae · citation-type:file-state

**finding:** The test named "*expanding rather than protecting*" cannot observe expansion, so it passes while expansion never happens. Its link target `real-deps` is empty (`:317`), and `mockExec` (`:20-22`) returns the same stdout regardless of argv — so a scoped `git status -- dist` call is indistinguishable from one that never occurred.

```
338:  it('treats a symlinked build dir as inspectable, expanding rather than protecting', async () => {
340:      mockExec('!! dist\n'),
343:    expect(verdict).toEqual({ protect: false });
```

**suggestion:** Write a `.env` into the link target and assert `{ protect: true, detail: 'dist/prod.env' }`; use an argv-aware exec stub so a missing scoped call fails the test rather than passing silently.

---

### 3. high · high confidence · correctness/security · `src/agent/worktree-ignored-patterns.ts:85-86` · ref:3b3b1ae · citation-type:file-state

**finding:** `GENERATED_ARTIFACT_EXTENSIONS` matches bare `.ts`/`.tsx`/`.mts`/`.cts`, not only `.d.ts`, so a hand-authored TypeScript file under any build directory loses name-hint protection. No compiler emits non-declaration `.ts`, and `target/` is Rust/Java output where `.ts` never appears as an artifact at all — so these extensions cannot be name-mirrors of a source module, which is the stated justification for the suppression.

```
85:const GENERATED_ARTIFACT_EXTENSIONS =
86:  /\.(?:js|mjs|cjs|jsx|ts|tsx|mts|cts|map|html?|css|svg)$/;
```

Executed at the ref (all three are `protected` on `main`):

```
dist/my-credentials.ts     -> inspectable [REAPABLE]
target/secret-config.ts    -> inspectable [REAPABLE]
out/credentials.mts        -> inspectable [REAPABLE]
```

**suggestion:** Replace the bare TS alternatives with declaration-only forms — `d\.ts`, `d\.mts`, `d\.cts` — leaving `.js`/`.map`/`.html`/`.css` as-is.

---

### 4. high · high confidence · correctness/security · `src/agent/worktree-ignored-patterns.ts:145,216-222` · ref:3b3b1ae · citation-type:file-state

**finding:** `logs/` is a member of `INSPECTABLE_REBUILDABLE_DIRS` and therefore satisfies `isGeneratedArtifact`'s directory condition, but nothing compiles into `logs/` — so the "name is an echo of the source module" rationale does not apply there. The module's own comment (`:114-119`) describes this tier as "*plausibly holding something a human put there by hand … a captured transcript*", which is precisely the content being unprotected.

```
216:function isGeneratedArtifact(normalizedPath: string, leaf: string): boolean {
217:  if (!GENERATED_ARTIFACT_EXTENSIONS.test(leaf)) return false;
219:    OPAQUE_REBUILDABLE_DIRS.some((re) => re.test(normalizedPath)) ||
220:    INSPECTABLE_REBUILDABLE_DIRS.some((re) => re.test(normalizedPath))
```

Executed at the ref (all `protected` on `main`):

```
logs/credentials.js      -> inspectable [REAPABLE]
logs/secret-notes.html   -> inspectable [REAPABLE]
logs/my-secrets.css      -> inspectable [REAPABLE]
logs/credentials.json    -> protected      (data-format carve-out holds)
```

**suggestion:** Pass an explicit generated-output directory list to `isGeneratedArtifact` that excludes `logs/`, rather than reusing both rebuildable tables wholesale — `logs/` earns its inspectable tier for expansion, not for artifact suppression.

---

### 5. medium · high confidence · api-compat · `src/agent/worktree-orphan-guard.ts:128,140,152` · ref:3b3b1ae · citation-type:file-state

**finding:** The second production consumer of `classifyIgnoredEntry` was not updated, so the two call sites now disagree on the identical on-disk case. `Dirent.isDirectory()` returns `false` for a symlink regardless of target (verified: `name=node_modules isDirectory()=false isSymbolicLink()=true`), so a symlinked dep tree takes the file branch at `:152` and classifies bare → `protected`. Direction is fail-safe (preserves rather than deletes), so this is not data loss — but orphaned symlinked trees are never reclaimed, and the PR's premise of one shared policy no longer holds for this input class. Reaches production via `classifyOrphanDir` ← `worktree-sweep.ts:593`.

```
128:      if (entry.isDirectory()) {
140:        if (classifyIgnoredEntry(`${rel}/`) === 'opaque') continue;
152:      if (!isRebuildable(rel)) {
```

**suggestion:** Use `entry.isDirectory() || (entry.isSymbolicLink() && (await stat(abs)).isDirectory())` in the orphan-guard walk, or document the divergence as deliberate in the shared-policy contract.

---

### 6. low · high confidence · test-coverage · `src/agent/worktree-ignored-patterns.test.ts:102` · ref:3b3b1ae · citation-type:file-state

**finding:** The `generatedMirrors` fixture pins only `dist/`, `coverage/`, and `node_modules/` cases, and the only `.ts`-suffixed entry in the whole file is a `.d.ts`. Findings 3 and 4 are therefore unpinned in either direction. (Verified by grep: `logs/` appears at `:86,184,188` but never combined with a name hint; zero bare-`.ts`-under-build-dir fixtures exist.)

```
102:    'dist/agent/auth/credential-resolver.d.ts',
```

**suggestion:** Add `dist/creds.ts` and `logs/credentials.js` as pinned cases once the intended verdict for each is settled.

---

## What was not checked

- Citations verified inline against branch HEAD `3b3b1ae0ecb84d5b03078e6f081571de56f8c866`; post-merge file state was fetched at that ref rather than read from the local checkout, which is on `main`.
- Findings 1, 3, and 4 were additionally re-derived by an independent verifier wave and by executing the ref's own modules against a real filesystem; findings 5 and 6 rest on code reads plus a targeted `Dirent` probe.
- Stated intent: PR #927 title + body — spec-compliance assessed. No unmet-intent gap found; findings 3 and 4 are scope overshoot beyond the stated "generated artifact" definition, not unmet requirements.
- Did not run the project test suite, `pnpm lint`, or `pnpm build`; the description's green-CI claim is taken at face value, and CI is in fact green on this head. Note that a green suite is consistent with finding 2 — the masking test passes.
- Did not review the sweep's age-gating, locking, or meta-file logic, nor the Telegram/CLI surfaces.
- The "29 of 29 → 2 of 29" effect claim was not independently reproduced; it depends on local worktree state not available here.

PR #928, opened 2026-08-07
fix(agents): tell research-agent about its registry grants, not just grant them

griffinwork40:
## The bug

#925 (closing #883) granted `research-agent` the `memory_search` tool at its registry entry. That grant is real and live — it's in the shipped v5.97.3 bundle:

```
tools:[...Do.allowedTools,"memory_search","Agent(git-investigator)"]
```

**It changed nothing observable.** `/ground-state`'s memory surveyor kept degrading to reading `HOT.md`/`AFK.md` off disk, and kept reporting `memory_search` as unavailable.

The reason is that the agent is still told otherwise. The vendored prompt body is byte-pinned to upstream and asserts a **closed** set:

> Your tool surface is a hard allowlist enforced by Claude Code: `Read, Grep, Glob, WebFetch, WebSearch`. You have no access to Edit, Write, NotebookEdit, or Bash.

`grep -c memory_search src/skills/_agents/prompts/research-agent.md` → **0**.

An agent told it has a *hard* five-tool surface does not attempt a sixth. The surveyor's "memory_search is missing from my allowlist" was an accurate report of its **prompt**, not of its **grant**.

## Why it survived review, merge, release, and tests

A registry-entry tool grant and the agent's **belief** about its tool surface are two separate systems. #925 updated only the first, and was verified only at the tool-access layer — `registry.test.ts` asserted the *resolved access set* on both child pools, which passed, correctly, while the behaviour stayed broken.

#925's commit message names the tradeoff explicitly — *"The vendored prompt stays byte-equal with upstream"* — as a deliberate choice. It was the right instinct about the file and the wrong conclusion about the agent.

## The fix

Reconcile at **prompt-composition time** rather than editing the vendored copy, so upstream byte-equality and the SHA-256 pin survive:

- `withRegistryGrants()` (`src/agent/agents/builtins.ts`) appends a section naming the granted tools and **explicitly superseding** the closed-allowlist sentence — while restating that every other restriction (no Edit, no Write, no Bash, no mutation) still holds. Appending a tool name without overriding the earlier "hard allowlist" claim would leave the prompt self-contradictory.
- Grants come from a single const, `RESEARCH_AGENT_REGISTRY_GRANTS`, which is spread into `tools` **and** rendered into the prompt. Mechanical access and stated access cannot drift apart, so a future grant can't be silently inert the way this one was.
- `git-investigator` uses the same composition path and has no registry grants beyond its allowlist, so the helper no-ops for it.

Rejected alternatives: editing the vendored prompt + `pnpm fix:pins` (abandons upstream parity for this agent); patching only `/ground-state`'s dispatch prompt (fixes one caller, leaves every other `research-agent` dispatch blind).

## Verification

- `pnpm lint` — clean
- `registry.test.ts` + `vendored.test.ts` — **45/45**
- `pnpm fix:pins:check` — **"All hash pins are up to date"** (upstream byte-equality preserved, which is this approach's central claim)
- **Verified to fail without the fix**: reverting the single call site yields `prompt never mentions granted tool memory_search`
- **Verified at the behaviour layer**: rendered the composed prompt and read the appended section. Re-checking the access layer would have repeated exactly the mistake that let this ship.

The new test asserts the *general property* — every registry grant beyond the vendored allowlist must appear in the prompt — rather than the literal string `memory_search`, so it covers future grants too.

## Impact

Every `/ground-state` run since #925 merged has had its memory third silently degraded to file reads. Those results were archive-free; treat them accordingly.

Merging cuts a patch release (`fix(agents):` via `auto-release.yml`).

### DIFF
--- a/src/agent/agents/builtins.ts
+++ b/src/agent/agents/builtins.ts
@@ -36,6 +36,47 @@ function vendoredPromptBody(raw: string): string {
   return parseAgentMarkdown(raw)?.definition.prompt ?? raw;
 }
 
+/**
+ * Invariant: a registry-entry tool grant and the agent's BELIEF about its tool
+ * surface are two separate systems, and both must be updated together.
+ *
+ * The vendored prompt bodies are byte-pinned to upstream
+ * (src/skills/_agents/vendored.test.ts), so they still name the tool allowlist
+ * upstream shipped with. research-agent's body asserts a CLOSED set —
+ * "Your tool surface is a hard allowlist enforced by Claude Code: `Read, Grep,
+ * Glob, WebFetch, WebSearch`" — which predates the grants the registry entries
+ * below add. An agent told it has a hard five-tool surface does not attempt a
+ * sixth, so a grant made only in `tools:` is INERT: #883 shipped the mechanical
+ * access (verified at the tool-access layer) while the surveyor kept correctly
+ * reporting `memory_search` as unavailable, because it was reading its prompt,
+ * not its grant.
+ *
+ * Reconciling at composition time (rather than editing the vendored file and
+ * re-pinning) keeps the vendored copy byte-equal with upstream and the pin
+ * green. Callers pass the SAME array they spread into `tools`, so the note and
+ * the grant cannot drift apart.
+ */
+function withRegistryGrants(body: string, grants: readonly string[]): string {
+  if (grants.length === 0) return body;
+  return `${body}
+
+## Tools granted by this runtime, beyond the list above
+
+This runtime additionally grants you: ${grants.join(', ')}.
+
+Any earlier statement in this prompt that your tool surface is a fixed or "hard"
+allowlist describes the upstream default, not your actual surface — it is
+superseded by this section. These tools are live; use them when relevant. Every
+other restriction above (no Edit, no Write, no Bash, no mutation) still holds.`;
+}
+
+/**
+ * Registry-entry grants for research-agent, beyond `researchAgent.allowedTools`.
+ * Single source of truth: spread into `tools` AND rendered into the prompt via
+ * `withRegistryGrants`, so mechanical access and stated access stay in lockstep.
+ */
+const RESEARCH_AGENT_REGISTRY_GRANTS = ['memory_search', 'Agent(git-investigator)'] as const;
+
 const GENERAL_PURPOSE_PROMPT = `You are a general-purpose sub-agent for complex, multi-step tasks that require both exploration and action.
 
 Work autonomously from the task prompt you were dispatched with: investigate, act, and verify. You have the parent session's full child tool surface. Keep intermediate exploration out of your reply.
@@ -103,7 +144,10 @@ export function builtinAgents(): Map<string, RegisteredAgent> {
       source: 'builtin',
       definition: {
         description: researchAgent.description,
-        prompt: vendoredPromptBody(researchAgent.systemPrompt),
+        prompt: withRegistryGrants(
+          vendoredPromptBody(researchAgent.systemPrompt),
+          RESEARCH_AGENT_REGISTRY_GRANTS,
+        ),
         // The scoped `Agent(git-investigator)` grant matches the vendored
         // prompt's frontmatter intent (`tools: …, Agent(git-investigator)`)
         // and lets research-agent nest a git-investigator when a finding needs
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

## Merge decision: **MERGE**

Light regime (2 files, 81 lines, hotfix). All citations below were re-read against the reviewed ref `9c6797bfa6eda0e9a4fef8adb66af17c8aa980a2`, not against `main`.

No `critical`, `high`, or `medium` findings. Two `low` findings, both textual rather than functional.

---

### The central claim holds

This PR's whole approach rests on one assertion — that composing at prompt-assembly time preserves upstream byte-equality and the SHA-256 pin. Verified independently rather than taken from the PR body:

- `sha256(src/skills/_agents/prompts/research-agent.md)` at the ref = `80993a5f28dbcc9fd447a5bf022fc2ae14307d4af21fcaa0900a0b135515db0a`, byte-identical to the pin at `vendored.test.ts:13`. The append happens on the in-memory string; the file is never touched.
- No existing assertion breaks on a longer prompt: the `strips vendored frontmatter` test (`registry.test.ts:175-183`) checks only `startsWith('---') === false` and `length > 0`, and `skill-bridge.test.ts:674` uses `toContain`. Both are append-safe.
- The resolved `tools` surface is unchanged: `RESEARCH_AGENT_REGISTRY_GRANTS` reproduces the prior inline array's exact content and order, so the pinned assertion at `registry.test.ts:53-61` still passes on an identical array.

The new test earns its place. It derives `extras` dynamically from `researchAgent.allowedTools` instead of hardcoding `memory_search`, so it covers future grants as claimed, and it genuinely fails pre-fix — the vendored body at the ref contains zero occurrences of `memory_search`, `superseded by this section`, or `no Edit, no Write, no Bash, no mutation`. The `t as never` cast in `.includes(t as never)` is a compile-time widening forced by the `as const` tuple type; `Array.prototype.includes` ignores static types at runtime, so the check is not weakened.

---

### Findings

**1. `low` · medium confidence · correctness · `src/agent/agents/builtins.ts:65,78` · ref:`9c6797bf` · citation-type:diff-context**

The grant announcement renders `Agent(git-investigator)` to the model as though it were a callable tool name, but the wire-level tool is `agent` with an `agent_type` parameter — the parenthesised form is frontmatter shorthand for a dispatch *scope*, parsed by `extractNestedAgentScope` (`resolve.ts:161-191`), never exposed as a tool name.

```
const RESEARCH_AGENT_REGISTRY_GRANTS = ['memory_search', 'Agent(git-investigator)'] as const;
This runtime additionally grants you: ${grants.join(', ')}.
```

Worth noting precisely: pre-PR the model never saw this token, because the frontmatter line carrying it (`research-agent.md:4`) is stripped by `vendoredPromptBody`. This PR is what first surfaces it into the prompt body. Two things keep it minor — the unmodified body already states the correct mental model one section earlier ("dispatch `git-investigator` via the Agent tool", `research-agent.md:21`), and dispatch is mechanically gated by `nestedAgentTypes` regardless of prompt wording. It is a textual-accuracy risk in the one mechanism this PR exists to make accurate, not a demonstrated dispatch failure.

*Suggestion:* render as `memory_search, and the Agent tool scoped to git-investigator` rather than echoing the raw frontmatter token — or derive the display string by stripping the paren scope.

**2. `low` · high confidence · spec-compliance · `src/agent/agents/builtins.ts:192-193` · ref:`9c6797bf` · citation-type:file-state (re-read at ref)**

The PR body states "`git-investigator` uses the same composition path and has no registry grants beyond its allowlist, so the helper no-ops for it." At the ref it does not use that path at all — it bypasses the helper rather than no-opping through it.

```
prompt: vendoredPromptBody(gitInvestigator.systemPrompt),
tools: [...gitInvestigator.allowedTools],
```

`withRegistryGrants` has exactly one call site (line 147, the research-agent entry). Output is identical either way, since the helper early-returns on an empty grant list — so this is a description/implementation mismatch, not a defect.

*Suggestion:* correct the PR body's wording, or wrap the git-investigator entry in `withRegistryGrants(vendoredPromptBody(gitInvestigator.systemPrompt), [])` so the stated path parity is literally true.

---

### Dimension roll-up

| Dimension | Result |
|---|---|
| security | no issues found — grants are fixed in-code string constants; no external input reaches the prompt template (`builtins.ts:59-70`) |
| api-compat | no issues found — resolved `tools` array byte-identical pre/post; only the prompt string is additive |
| correctness | 1 × `low` (finding 1) |
| spec-compliance | 1 × `low` (finding 2); stated requirements 1, 2, 3, 5 confirmed implemented; no scope creep — `registry.test.ts` has 0 deletions |
| test-coverage | no issues found — new test asserts the general property and fails pre-fix |
| perf-observability | no issues found — 0 matches for `promptLength\|MAX_PROMPT\|maxPromptLength\|promptTokens` under `src/agent/agents`, so no budget assertion exists to regress |

### What was not checked

- Citations verified inline against branch HEAD `9c6797bfa6eda0e9a4fef8adb66af17c8aa980a2`. One Wave 1 citation was mis-numbered (git-investigator entry given as `143-163`); the quoted snippet was verbatim-real and the line reference is corrected to `192-193` above. No citation was fabricated and no absence claim was refuted.
- Stated intent: PR #928 title + body — spec-compliance **was** assessed against it.
- Whether a live dispatched model reads `Agent(git-investigator)` as a literal callable name is a runtime-behaviour question outside static verification. Finding 1 is scoped to the text, not to an observed failure.
- Tests were not executed locally; CI on this ref is green across lint/build, ubuntu, macOS, PTY scrollback, docs, and bundle smoke.
- `diagnose.ts`'s separate `createGitOrchestratorCanUseTool` prompt-assembly path was not reviewed — a distinct pre-existing composition site outside this PR's diff and stated scope. If the goal is that no grant can ever again be silently inert, that second site is where the same class of drift would next hide.

PR #929, opened 2026-08-07
fix(agent): preserve abort origins across provider teardown

### LINKED ISSUE #856: Abort origin is not distinguished: timeout- and budget-driven teardowns are classified benign
## Summary

`failureClass: 'abort'` is stamped on any tool result whose `signal.aborted` is true. But `signal.aborted` is how **four different causes** arrive — deliberate user cancellation, session timeout, budget exhaustion, and session close — and the originating cause is destroyed before any stamp site can see it.

Because `'abort'` is a member of `BENIGN_FAILURE_CLASSES`, every one of those causes is now treated as "the system correctly said no":

- **Rendering** — the interactive tool-lane draws a neutral `⊘` glyph and warning-toned body instead of a red `✗`.
- **Statistics** — `tool-failure-density.ts` excludes the result from failure-rate stats entirely.

So a tool call killed by a **timeout** currently reads as an intentional refusal in both the TUI and the improve pipeline.

This directly contradicts the rule written into `BENIGN_FAILURE_CLASSES` itself:

> `timeout` is excluded because a high timeout rate is a real problem (too tight a deadline, a systematically slow target)

We exclude the `timeout` class from the benign set on purpose — and then let timeouts in through the `abort` door.

## Why the origin is lost

The reason channel exists but is too narrow, and it gets flattened one hop before it would be useful:

1. `withTimeout()` aborts the caller's controller with a real `TimeoutError` — `src/agent/timeout.ts:120`:
   ```ts
   options.controller.abort(err);
   ```
2. `AgentSession` hands its own long-lived controller to `withTimeout` (`src/agent/session/agent-session.ts:594`), and `abortBudget` aborts that same controller (`:478`).
3. That abort fires `onAbort()` (`agent-session.ts:880`), which calls `providerQuery.interrupt()` (`:883`).
4. `interrupt()` discards the reason and substitutes a string literal — `anthropic-direct/query.ts:629` and `openai-compatible/query.ts:935`:
   ```ts
   this.abort.requestAbort('interrupted');
   ```
5. `AbortReason` has no vocabulary for the distinction — `src/agent/providers/shared/abort-coordinator.ts:56`:
   ```ts
   export type AbortReason = 'interrupted' | 'closed';
   ```

By the time a dispatch site reads `signal.aborted`, a timeout, a blown budget, and a Ctrl+C are indistinguishable. `signal.reason` is present but always carries the flattened `'interrupted'`.

## Affected sites

Ten production stamp sites, all gated on `signal.aborted`. **Six predate #848** — this is a pre-existing systemic gap, not a regression introduced by that PR:

Pre-existing on `main`:
- `src/agent/tools/dispatcher.ts:809, 923, 1087, 1125`
- `src/agent/tools/handlers/browser-act.ts:174`
- `src/agent/tools/handlers/browser-open.ts:104`

Added by #848 (which extended the pattern to the provider-loop synthesized results, and gave it its first *visual* consequence):
- `src/agent/providers/anthropic-direct/loop/tool-dispatch.ts:128`
- `src/agent/providers/openai-compatible/query/dispatch-append.ts:137, 146, 165`

The four `#848` sites carry inline comments recording this gap so it is not silently inherited.

## Consequences, concretely

- **Failure-density detector blindness.** `tool-failure-density.ts:171` drops any result whose class is in the excluded set; `:242` shows an *absent* class still counts as `unclassified`. So before the stamp existed these results were counted, and now they are not. A session whose tools are being systematically killed by too tight a `timeoutMs` produces a *cleaner* failure-density report than one that is healthy.
- **Misleading TUI.** A run torn down by a budget cap renders as a screenful of polite `⊘` "the system said no" rows, which is precisely the signal `⊘` was introduced to reserve for deliberate refusals.

## Proposed fix

Widen the reason channel and thread the origin to the stamp sites:

1. Extend `AbortReason` beyond `'interrupted' | 'closed'` to carry origin — e.g. `'user-interrupt' | 'timeout' | 'budget' | 'closed'`. An `AbortOrigin` union already exists at `src/agent/trace/events.ts` for the tracing layer and is a reasonable vocabulary to align with rather than invent a second one.
2. Give `interrupt()` an optional reason parameter so `onAbort()` can pass *why* it fired, instead of hardcoding `'interrupted'` at `anthropic-direct/query.ts:629` and `openai-compatible/query.ts:935`.
3. At the stamp sites, read `signal.reason` and classify: `'abort'` only for a genuine user cancellation; `'timeout'` for a deadline; leave budget teardown either unclassified or give it its own class.
4. Add a test pinning that a timeout-driven abort is **not** benign — the partition test in `tool-lane-format.test.ts` is the natural sibling for it.

Roughly five files across the session/provider abort machinery, plus tests. Deliberately out of scope for #848, which is a TUI-rendering change.

## Acceptance criteria

- [ ] A tool call aborted by a session timeout does **not** carry a class in `BENIGN_FAILURE_CLASSES`.
- [ ] A tool call aborted by a user interrupt still renders `⊘` and stays excluded from failure-density stats.
- [ ] A timeout-killed call is counted by `tool-failure-density.ts`.
- [ ] Budget-driven teardown is classified deliberately, not by accident.
- [ ] The inline caveat comments at the four `#848` stamp sites are removed once the gap is closed.

## Context

Found during review of #848 (comment thread on that PR). Confirmed by reading the chain at `710e72aa` and independently re-derived by an adversarial verifier; the "same AbortController" framing in the original review was imprecise — the session controller and the provider's per-turn coordinator are distinct objects bridged by `interrupt()`, which is *where* the reason is discarded.

griffinwork40:
## Summary
- Preserve the originating timeout or budget reason when session teardown interrupts a provider turn.
- Classify timeout- and budget-aborted tool calls as real operational failures while keeping deliberate user interruption benign.
- Apply the shared classifier at every synthetic abort-result site and pin TUI/failure-density behavior.

## Diagnosis
The session controller retained typed timeout and budget errors, but `ProviderQuery.interrupt()` flattened every teardown to `interrupted`. Synthetic tool results therefore stamped timeout and budget exhaustion as benign `abort` failures, hiding them behind the neutral TUI glyph and excluding them from failure-density statistics.

## Behavior changes
- Provider abort reasons now distinguish `interrupted`, `timeout`, `budget`, and `closed`.
- Timeout and budget aborts remain visible in tool failure rendering and statistics.
- User interruption continues to use the benign `abort` class.
- Budget teardown passes the typed `BudgetExceededError` through the reason channel.

## Test results
- Focused tests (10 files): 406 passed, 1 skipped.
- `pnpm lint`: passed.
- `pnpm build`: passed.
- `pnpm test:coverage`: passed (full Vitest suite and coverage thresholds).
- `pnpm test:pty`: passed (1 file, 10 tests).
- `pnpm audit:env:check`: passed (814 files, 0 violations).
- `pnpm scan:env:check`: passed (149 variables in sync).
- `pnpm audit:sdk:check`: passed.
- `pnpm audit:chalk:check`: passed (814 files, 0 violations).
- `pnpm fix:pins:check`: passed.
- `pnpm audit:deps`: passed; no critical production advisories (29 non-blocking advisories: 3 low, 17 moderate, 9 high).
- `git diff --check`: passed.

## Verification
- **CONFIRM:** timeout and budget reasons are preserved from session abort through provider interruption and the shared abort coordinator.
- **CONFIRM:** timeout and budget synthetic failures are non-benign, while deliberate interruption remains benign.
- **REFINE:** all relevant production stamp sites use the shared classifier and its core/downstream behavior is unit-tested; direct provider-dispatch integration tests for typed timeout/budget signals are not included, so coverage is compositional rather than end-to-end at each stamp site.

Fixes #856

### DIFF
--- a/src/agent/abort-reason.test.ts
+++ b/src/agent/abort-reason.test.ts
@@ -1,0 +1,29 @@
+import { describe, expect, it } from 'vitest';
+import { BudgetExceededError, TimeoutError } from '../utils/errors.js';
+import { abortFailureClass, providerAbortReason } from './abort-reason.js';
+import { BENIGN_FAILURE_CLASSES } from './trace/types.js';
+
+function abortedSignal(reason: unknown): AbortSignal {
+  const controller = new AbortController();
+  controller.abort(reason);
+  return controller.signal;
+}
+
+describe('abort reason classification', () => {
+  it.each([
+    [new TimeoutError('session', 100), 'timeout'],
+    [new BudgetExceededError(2, 1), 'budget'],
+    ['Budget $2.00 exceeded limit of $1.00', 'budget'],
+    ['session timed out after 100ms', 'timeout'],
+    ['sigint', 'interrupted'],
+  ] as const)('maps %o to %s', (reason, expected) => {
+    expect(providerAbortReason(reason)).toBe(expected);
+  });
+
+  it('keeps user interruption benign while timeout and budget remain visible', () => {
+    expect(abortFailureClass(abortedSignal('interrupted'))).toBe('abort');
+    expect(BENIGN_FAILURE_CLASSES.has(abortFailureClass(abortedSignal('interrupted')))).toBe(true);
+    expect(BENIGN_FAILURE_CLASSES.has(abortFailureClass(abortedSignal('timeout')))).toBe(false);
+    expect(BENIGN_FAILURE_CLASSES.has(abortFailureClass(abortedSignal('budget')))).toBe(false);
+  });
+});
--- a/src/agent/abort-reason.ts
+++ b/src/agent/abort-reason.ts
@@ -1,0 +1,28 @@
+import type { ToolFailureClass } from './trace/types.js';
+import { BudgetExceededError, TimeoutError } from '../utils/errors.js';
+
+/** Provider-level reason preserved on the per-turn AbortSignal. */
+export type ProviderAbortReason = 'interrupted' | 'timeout' | 'budget' | 'closed';
+
+/**
+ * Reduce the session controller's open-ended reason to the provider contract.
+ * Typed errors are authoritative; string fallbacks preserve compatibility with
+ * the budget producer and callers that serialize errors before aborting.
+ */
+export function providerAbortReason(reason: unknown): ProviderAbortReason {
+  if (reason instanceof TimeoutError) return 'timeout';
+  if (reason instanceof BudgetExceededError) return 'budget';
+  if (typeof reason === 'string') {
+    if (reason === 'budget' || reason.startsWith('Budget ')) return 'budget';
+    if (reason === 'timeout' || reason.includes('timed out')) return 'timeout';
+    if (reason === 'closed') return 'closed';
+  }
+  return 'interrupted';
+}
+
+/** Classify a synthetic tool failure from the reason preserved on its signal. */
+export function abortFailureClass(signal: AbortSignal): ToolFailureClass {
+  const reason = providerAbortReason(signal.reason);
+  if (reason === 'timeout' || reason === 'budget') return reason;
+  return 'abort';
+}
--- a/src/agent/provider.ts
+++ b/src/agent/provider.ts
@@ -332,7 +332,7 @@ export interface ProviderQueryArgs {
  * harness having to widen a Claude-specific union.
  */
 export interface ProviderQuery extends AsyncIterable<ProviderEvent> {
-  interrupt(): Promise<void>;
+  interrupt(reason?: import('./abort-reason.js').ProviderAbortReason): Promise<void>;
   setModel(model?: string): Promise<void>;
   setPermissionMode(mode: string): Promise<void>;
   supportedCommands(): Promise<ProviderCommandInfo[]>;
--- a/src/agent/providers/anthropic-direct/loop/tool-dispatch.ts
+++ b/src/agent/providers/anthropic-direct/loop/tool-dispatch.ts
@@ -17,6 +17,7 @@
 import type { ContentBlockParam, ToolResultBlockParam } from '@anthropic-ai/sdk/resources';
 import type { ProviderEvent } from '../../../provider.js';
 import type { RunTurnInput, ToolCall, ToolResult, TurnResult } from '../types.js';
+import { abortFailureClass } from '../../../abort-reason.js';
 import { emitToolCall } from '../../../trace/emit.js';
 import { extractRawToolInput } from '../../../facets/raw-input.js';
 import { summarizeToolInput } from '../../shared/tool-input-summary.js';
@@ -115,17 +116,11 @@ export async function* dispatchToolCalls(
   } else {
     results = [];
     for (const call of calls) {
-      // `abort` here covers ANY signal-fired teardown — deliberate user
-      // cancel, session timeout, or budget exhaustion all funnel through the
-      // same AbortSignal (requestAbort() only tracks 'interrupted' | 'closed',
-      // see abort-coordinator.ts:56), so the original cause is lost by the
-      // time this dispatch site sees `signal.aborted`. Widening AbortReason
-      // to carry the origin is the real fix; out of scope here.
       if (input.signal.aborted) {
         results.push({
           content: 'Tool call aborted',
           isError: true,
-          failureClass: 'abort',
+          failureClass: abortFailureClass(input.signal),
         });
         continue;
       }
--- a/src/agent/providers/anthropic-direct/query.ts
+++ b/src/agent/providers/anthropic-direct/query.ts
@@ -638,8 +638,8 @@ export class AnthropicDirectQuery implements ProviderQuery {
     return withSystemBreakpoint(blocks, getCacheTtl());
   }
 
-  async interrupt(): Promise<void> {
-    this.abort.requestAbort('interrupted');
+  async interrupt(reason: import('../../abort-reason.js').ProviderAbortReason = 'interrupted'): Promise<void> {
+    this.abort.requestAbort(reason);
   }
 
   /**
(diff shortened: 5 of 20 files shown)


griffinwork40 (commented):

## Merge decision: **MERGE**

No blocking findings. The change does what its description claims, and the claim survived independent verification at the reviewed ref.

**Target:** PR #929 · hotfix · 20 files · +131/−70
**Reviewed ref:** `991833a566bc667b918ba1b77c5615165a9d0403`
**Stated intent:** PR #929 title + body — spec-compliance **assessed**.

---

## Verified core mechanism

The load-bearing risk in this design is the classifier's fallback ladder in `providerAbortReason()`. Each step was re-derived against the ref rather than taken on trust:

- **`instanceof` ordering is safe.** `TimeoutError` (`src/utils/errors.ts:8`) and `BudgetExceededError` (`src/utils/errors.ts:96`) both extend `Error` directly — neither subclasses the other, so the ordered `instanceof` checks cannot shadow one another. `IdleWatchdogError extends TimeoutError` (`src/utils/errors.ts:33`) is correctly absorbed into `timeout`.
- **The `'Budget '` string prefix actually matches.** `BudgetExceededError`'s default template is `` `Budget ceiling reached: $${…} cumulative >= $${…} limit` `` (`src/utils/errors.ts:104-106`) — so `startsWith('Budget ')` is live, not dead code.
- **Budget teardown really does pass the typed error.** The producer forwards the object verbatim with no stringification: `abortBudget: (reason) => { … this.abortController.abort(reason); }` (`src/agent/session/agent-session.ts:479-483`). `onAbort()` then resolves it via `instanceof`, so the relay never depends on the prefix fallback.
- **`'budget'` is non-benign by omission, correctly.** `BENIGN_FAILURE_CLASSES` (`src/agent/trace/types.ts:132-138`) is a `ReadonlySet` membership test, not an exhaustive map, so adding `'budget'` to `TOOL_FAILURE_CLASSES` makes it count toward failure stats with no other edit required. Grep for `Record<ToolFailureClass` returned **0 matches** — no exhaustive map silently went stale.
- **`AbortReason` widening breaks no consumer.** The only equality check on the widened union is `signal.reason === 'interrupted'` in `turn-trace.ts:90`, which is deliberately scoped to ESC-halt latency; `timeout`/`budget` are correctly inert there.
- **`ProviderQuery.interrupt()` widening is compatible.** All three production implementors (`anthropic-direct/query.ts`, `openai-compatible/query.ts`, `router/provider-router.ts`) default to `'interrupted'` and forward the argument. Pre-existing zero-arg callers are deliberate user stops and stay benign.

**Absence claim verified:** the intent's "every synthetic abort-result site" holds. `git grep "failureClass: 'abort'"` at the ref, excluding `*.test.ts`, returns only `src/improve/eval-run/contracts.ts:416,426` — both are synthetic corpus rows inside `buildToolFailureClassificationCorpus` (each annotated `// excluded`), which are fixture data asserting the detector's exclusion logic, not live classification sites. Correct as-is.

---

## Findings

### low · confidence: medium · test-coverage · `src/agent/__fixtures__/mock-provider.ts:136-138` · ref:`991833a` · citation-type: file-state (re-read and verified at ref)

**Finding:** The shared `ProviderQuery` test double discards the new `reason` argument and unconditionally aborts with `'interrupted'`, so no session-level test built on this fixture can observe timeout/budget reason propagation through `AgentSession.onAbort()` — which is the exact behavior this PR introduces.

**Evidence:**
```ts
async interrupt() {
  q.interruptCalls++;
  abortController?.abort('interrupted');
```

**Suggestion:** Thread the argument through — `async interrupt(reason = 'interrupted') { q.interruptCalls++; abortController?.abort(reason); }` — which turns the PR's self-disclosed compositional coverage into an assertable end-to-end path without new test scaffolding.

This is the concrete mechanism behind the PR's own honest **REFINE** note ("direct provider-dispatch integration tests for typed timeout/budget signals are not included"). Disclosed gap, not a hidden one — hence `low`, not blocking.

### nit · confidence: high · correctness · `src/agent/abort-reason.ts:9-11` · ref:`991833a` · citation-type: diff-context

**Finding:** The docstring justifies the string fallback by citing "the budget producer", but this same PR converts that producer to pass a typed `BudgetExceededError`, so the cited justification is stale on arrival.

**Evidence:**
```ts
 * Typed errors are authoritative; string fallbacks preserve compatibility with
 * the budget producer and callers that serialize errors before aborting.
```

**Suggestion:** Drop "the budget producer and" — the remaining clause ("callers that serialize errors before aborting") is the accurate rationale for keeping the fallback defensive.

### security — `no issues found`

The `startsWith('Budget ')` / `includes('timed out')` matching is not attacker- or model-reachable. Every `AbortSignal.reason` origin traced is host-constructed: typed errors from `timeout.ts:115` and `dag.ts:234`, or hardcoded literals (`'sigint'`, `'closed'`, `'cancelled'`). `AgentSession.abort(reason: string)` (`agent-session.ts:537-547`) explicitly **throws** on any caller string that is `'closed'`, starts with `'Budget '`, or contains `'timed out'` — closing the one injection path into the classifier. That guard predates this PR and this PR keeps it intact.

### perf-observability — `no issues found`

`providerAbortReason` is two `instanceof` checks plus bounded string comparisons, invoked once per already-aborted call on a teardown path. Not a hot path.

---

## What was not checked

- **Citations verified inline against branch HEAD `991833a`** via `git show`/`git grep` at that ref. All `file-state` citations from the review agents were re-read at the ref; none were fabricated, and no finding was dropped.
- Tests were **not executed** — the PR's reported results (`test:coverage`, `lint`, `build`, `test:pty` all passing) were taken as reported, though CI is green on all 8 checks at this ref.
- No `tsc` run; the API-compatibility conclusions rest on TypeScript structural-typing rules plus grep enumeration of implementors, not compiler output.
- Not traced: whether `dag.ts` node-level `TimeoutError` aborts ever propagate into the *session* controller's reason channel. If they do, they classify correctly anyway (typed `instanceof` catches them before the string ladder), so this gap does not change the verdict.
- Did not review the Telegram surface or docs references to `interrupt()`.

PR #931, opened 2026-08-07
fix(cli): bound every raw REPL write to the terminal width

griffinwork40:
## Problem

Text goes out of bounds in the TUI. A row composed with a fixed column budget can exceed the live terminal width, and the terminal then auto-wraps it to **column 0** — so the continuation row loses the indent, tree glyph, or table column that produced it, and the output reads as broken rather than wrapped.

`/worktree list` is the clearest case: its table is built from literal `padEnd()` calls totalling ~100 columns. On a 62-column terminal every row breaks in half, with the second half hard against the left edge.

## Diagnosis

Two independent things were conflated in the original report, and only one of them is a bug:

1. **The tool lane was never at fault.** `toolLaneWidth()` = `min(getTerminalWidth(), 100)` already clamps every composed row on *both* the live-overlay path (`tool-lane.ts:331`) and the committed/flush path (`tool-lane-render-grouped-root.ts:117`). A measurement probe at 62 columns produced a max row width of exactly 62 on both paths, including a 250-char bash command — the vector the code comments call out as the risk.

2. **Rows already written before a narrowing resize cannot be recalled.** Rows still resident in the retained `committedBand` *are* re-wrapped on resize (`reflowCommittedBandToWidth` + `repositionCommittedBand`); rows already evicted past that cap into the terminal's own scrollback are written-and-forgotten, and no read-back path exists. A screenshot taken after shrinking the window will always show pre-resize rows reflowed by the terminal. That is inherent, not fixable here.

3. **The actual bug: the raw write seams.** Everything that is *not* the tool lane reaches the terminal through one of four seams that write straight to stdout with no compositor to wrap them, and none of them bounded anything:

| Seam | Path |
|---|---|
| `slash/writer.ts` | sinkless `console.log` — all slash-command output |
| `commands/interactive/repl-renderer.ts` | the disarmed TTY branch (between turns — exactly when slash commands run) |
| `commands/interactive/bootstrap-surface.ts` | `CompletionWriter`'s pre-arm `console.log` default for `fn` / `idleFn` |
| `terminal-compositor.committed-band-commit.ts` | `commitAbove`'s disarmed early return |

Each trusted its ~40 callers to have budgeted their own width. Many hadn't: `/worktree list` (~100 cols), `bg list` (~127 cols), `queue list` (unbounded command tail), the resume banner (80/120-char caps unrelated to terminal width).

## Fix

Add `src/cli/render/bounded-line.ts` and apply it at those four seams, so bounding is a **property of the seam** rather than a rule each call site has to remember. A new caller cannot reintroduce the bug by forgetting to budget.

- Over-wide rows are **wrapped, never truncated** — no content is lost.
- Every continuation row is re-prefixed with the source row's own leading indent, so wrapped output stays visually attached to the row that produced it:

```
  ↳ some long message that does not fit in this terminal

  ↳ some long message that does not
  fit in this terminal
```

- Rows that already fit pass through byte-identical (the common case, and why this is safe on every write).
- A pathological indent — deeper than `width - 8` — drops the hanging indent and plain hard-wraps, rather than squeezing text into an unreadable sliver.
- Long unbreakable tokens (hashes, URLs, paths) are broken at the column boundary rather than overflowing.

**Deliberately not applied to the armed compositor path.** `commitAbove` already hard-wraps to the live width for its row accounting *and* re-wraps the retained band on resize. Pre-wrapping there would freeze rows at today's width so a later **widen** could no longer rejoin them — the behaviour `tests/pty/compositor-scrollback.pty.test.ts`'s `width-resize-fragment-widen` case guards. That test still passes.

**Non-TTY output is byte-identical.** `afk worktree list | grep` must see whole logical rows; a consumer's parser must never have to reassemble rows this module split for a human.

Also fixes two display-width off-by-ones in `bgsub.ts` and `loop-iteration.ts`: `slice(0, 60) + '…'` measured 61 cells and mis-measured every wide, emoji, or combining grapheme it cut. Both now use `truncateDisplayWidth`, which budgets the ellipsis inside the limit.

## Verification

- `pnpm test` — 801 files / 15,276 tests pass
- `pnpm test:pty` — 10/10, including all four `width-resize-*` reflow guards
- `pnpm lint`, `pnpm build`, `audit:sdk:check`, `audit:env:check`, `scan:env:check`, `audit:chalk:check`, `fix:pins:check` — all pass
- 25 new assertions across three files. The invariant "no row exceeds the terminal width" is asserted directly against the real `/worktree list` row shape at 40 / 62 / 80 / 100 / 120 columns, plus ANSI-styled rows, CJK and emoji graphemes, 200-char unbreakable tokens, multi-line payloads, and the resize case.

## Not included

- Making the fixed-width tables themselves *responsive* (shrinking or dropping columns instead of wrapping). Wrapping is now correct and lossless; column-aware degradation is an aesthetic follow-up.
- The one-shot `commander` surfaces that `console.log` directly without passing through a seam (`improve/*`, `schedule`, `plugin`, `marketplace`). They share the same fixed-`padEnd` pattern and are worth a follow-up pass.

### DIFF
--- a/src/cli/commands/interactive/bootstrap-surface.ts
+++ b/src/cli/commands/interactive/bootstrap-surface.ts
@@ -14,6 +14,7 @@ import type { CliOptions } from './shared.js';
 import type { ResolvedResumeTarget } from '../../resume-session.js';
 import { createDefaultTraceWriter } from '../../../agent/trace/factory.js';
 import { palette } from '../../palette.js';
+import { boundLineToTerminal } from '../../render/bounded-line.js';
 
 /**
  * Seed session stats + permission/thinking-UI mode, print the `trace:` and
@@ -101,9 +102,15 @@ export function createReplSurface(a: {
   // after `armCompositor` resolves (when a borrowed compositor is available)
   // so between-turn slash output commits above the live overlay instead of
   // overlaying onto the input row. See CompletionWriter docs in shared.ts.
+  //
+  // Invariant: while these slots hold the raw `console.log` default there is
+  // no compositor to wrap what they emit, so each line is bounded to the
+  // terminal here. Once `runReplLoop` swaps in `compositor.commitAbove`, that
+  // sink owns wrapping (and band reflow on resize) and this bound is gone
+  // with the closure — the two must never both wrap the same line.
   const completionWriter: CompletionWriter = {
-    fn: (line) => console.log(line),
-    idleFn: (line) => console.log(line),
+    fn: (line) => console.log(boundLineToTerminal(line)),
+    idleFn: (line) => console.log(boundLineToTerminal(line)),
   };
   // Construct StatusLine BEFORE createReplRenderer so the renderer can route
   // its inter-turn raw writes through statusLine.withFullScrollRegion(...) —
--- a/src/cli/commands/interactive/loop-iteration.ts
+++ b/src/cli/commands/interactive/loop-iteration.ts
@@ -15,6 +15,7 @@ import { isDebugEnabled, debugLog } from '../../../utils/debug.js';
 import { sanitizeForDisplay } from '../../../utils/terminal-sanitize.js';
 import { env } from '../../../config/env.js';
 import { palette } from '../../palette.js';
+import { truncateDisplayWidth } from '../../display.js';
 import { ringBellIfEnabled } from '../../_lib/capture-mode.js';
 import { cyclePermissionMode } from '../../permission-mode-cycle.js';
 import {
@@ -259,7 +260,9 @@ export async function runInputLoop(
         const seconds = job.endedAt !== undefined
           ? Math.max(0, Math.round((job.endedAt - job.startedAt) / 100) / 10)
           : 0;
-        const label = job.label.length > 60 ? `${job.label.slice(0, 60)}…` : job.label;
+        // Display-width truncation, not code-unit — see the same fix in
+        // slash/commands/bgsub.ts: `slice(0, 60) + '…'` measured 61 cells.
+        const label = truncateDisplayWidth(job.label, 60);
         ctx.replRenderer.writeLine(
           palette.dim(`  ${glyph} [${job.jobId}] subagent ${job.status} · ${seconds}s · `) + label,
         );
--- a/src/cli/commands/interactive/repl-renderer.test.ts
+++ b/src/cli/commands/interactive/repl-renderer.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi, afterEach } from 'vitest';
 import { createReplRenderer } from './repl-renderer.js';
+import { displayWidth, stripAnsi } from '../../display.js';
 
 function makeStdout(isTTY: boolean) {
   const write = vi.fn();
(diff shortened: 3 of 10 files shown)


griffinwork40:
# Review — PR #931 `fix(cli): bound every raw REPL write to the terminal width`

**Recommendation: approve with two pre-merge one-liners.** The wrapping algorithm is correct, the tests are real, and every mechanical gate is green. Neither Major finding is a code defect — both are the module documenting an invariant broader than the one it enforces.

## Verification (run directly, not delegated)

`pnpm test` on the 3 affected files → **39/39 pass**. `pnpm lint` clean. `audit:chalk` / `audit:env` / `scan:env` / `fix:pins` / `audit:sdk` → all exit 0. Tree clean.

## What holds up

- **No terminal-corruption risk.** `wrap-ansi@10` tracks `isInsideEscape`/`isInsideLinkEscape`, so CSI and OSC-8 sequences are never split, and SGR reopens per continuation row. Probed across a `\x1b[31m` boundary at columns 5–16.
- **No ReDoS** (0.94 ms on 1 M-char adversarial input), **perf negligible** (~55 ns/op ASCII fast path).
- **No double-wrapping** on any chain; the `CompletionWriter` swap reassigns *both* `fn` and `idleFn`, so the pre-arm bound is genuinely unreachable post-arm.
- **The `truncateDisplayWidth` swap is a real fix** — it reserves the ellipsis width from the budget, so the old 61-cell output is genuinely gone.

## Findings

**M1 — `ctx.out.raw()` is a fifth raw TTY seam, unbounded, three lines below the bounded one.** `writer.ts:71` writes straight to stdout with no bound and no TTY gate, reachable via `/sh show`, `/afk-md show`, `/transcript` fallback. The module doc enumerates exactly four seams and states the invariant as "no row handed to a TTY may exceed the terminal width" — which `raw()` violates. Fix: bound it, or narrow the doc to "composed rows" and say why content dumps stay byte-identical.

**M2 — Hard-wrapping freezes rows at today's width, and the PR's own counter-argument applies here too.** The doc correctly argues the armed path must not be pre-wrapped because it reflows on resize. The mirror of that: a terminal that *soft*-wraps rejoins on widen; a `\n` from `hangingWrap` never does. The PR accepts that trade at the raw seams but never names it. Fix: one sentence.

**m3 — False-pass test, verified empirically.** `bounded-line.test.ts:88-92` passes identically with the `MIN_CONTENT_COLUMNS` guard deleted — measured 4 rows with it, 20 without, both fit width 34, both exceed 1 row. It is the only test on that branch.

**m4 — `bgsub.ts` 349 → 352 LOC**, newly crossing your hard 350 ceiling (no gate in this repo; `loop-iteration.ts` is already 809).

**m5 — CRLF silently normalized even when no wrap is needed** (`'a\r\nb'` → `'a\nb'`); these seams were byte-identical passthrough before. **m6** — `LEADING_INDENT` misses tab/OSC-8/U+3000 indents (hang lost, bound still holds). **m7** — `verdict-card` <40 cols was already broken, but the break is now frozen rather than reflow-able. **m8** — `changelog.ts` `'─'.repeat(60)` can emit a 1-char orphan row at width 59. Two nits: wide graphemes at `width: 1`; defensive-path rows lose the indent.

Pre-existing and out of scope: `afk chat`, standalone `afk worktree list`, and other one-shot CLI surfaces are still unbounded — a follow-up issue, not a blocker.

## Process caveat

Of five dispatched dimension agents, two failed (stream abort; 45-min timeout) and two hit their iteration cap. `research-agent` has no Bash, so no agent could execute a gate — I ran every command in the main session and re-derived m3/m4/m5/n10 empirically rather than trusting agent output.

**Done**
- Reviewed PR #931 across correctness, seam completeness, test coverage, regression risk, and security/perf; produced a merge recommendation with 2 Major, 6 Minor, 2 nit findings.
- Evidence: full report at `/Users/griffinlong/.afk/workspace/reports/review-pr-931-2026-08-07.md`; test/lint/gate results reproducible via the commands listed in it; empirical verifications run from `/tmp/verify931.ts`.
- Nothing changed in the world: review was read-only — no repo files edited, no commit, no push, no PR comment posted. Worktree confirmed clean (`git status --porcelain` empty).
- Deferred: no fixes applied, since `/review` is read-only by contract — M1 and m3 are the two I'd land before merge; the out-of-scope one-shot CLI surfaces warrant a separate issue.

---

Issue #12, opened 2026-06-06
/skills: job-to-be-done grouping via authored `category` field (Phase 2)


griffinwork40:
## Context

Phase 1 of the `/skills` UX redesign ships readability improvements without any grouping logic: word-wrapped descriptions, an enriched `/skills <name>` detail card, a source-legend header, a built-in-skills block at the top, and inline dim suffixes for shadowed/alias entries (e.g. `(+1 via /plugin:mint)`). Phase 1 lives in `src/cli/slash/plugin-skills.ts`.

Grouping skills by "job-to-be-done" was deliberately deferred to this issue.

## Decision: authored field, NOT a render-time heuristic

A devils-advocate critique (3 independent lenses) unanimously rejected the originally-planned render-time categorizer — a curated skill-name→category map plus a keyword heuristic over name+description. Reasons: it infers at render time what should be authored at the source; it misclassifies the plugin long-tail unpredictably; skills "jump categories" as descriptions change; and the fixed taxonomy rots as plugins evolve. The maintenance cost outweighs the benefit at current skill counts.

**Reliable job-to-be-done grouping requires categories authored at the source.** This issue implements that.

## Scope

1. **Add `category?: string`** to `SkillMetadata` in `src/skills/index.ts` (interface at ~lines 68-126). Document the allowed vocabulary in a comment/const.
2. **Define the category vocabulary** as a documented constant + render order. Starting proposal (refine during implementation): `Build & ship`, `Debug & fix`, `Understand & explore`, `Refactor & simplify`, `Review & verify`, `Setup & ops`, `Author & meta`. Skills with no category fall into a flat "More skills" block — there is NO inference.
3. **Harvest `category:` from frontmatter** for disk + plugin skills:
   - `parseSkillMd` in `src/cli/slash/_lib/flag-harvest.ts` (~line 46) already parses arbitrary scalar frontmatter keys into its `frontmatter` record — map `category` through in `parseUserSkillMd` (`src/skills/user-skills.ts` ~line 126, where the `SkillMetadata` is constructed ~line 273).
   - Add a `harvestPluginSkillCategories()` analog to the existing `harvestPluginSkillFlags()` (`src/cli/slash/plugin-skills.ts` ~line 115) for plugin skills, OR — preferred — fold category harvesting into the SAME single cached disk walk as flags to avoid a second traversal per render.
[...]

PR #932, opened 2026-08-07
feat(web): add `afk web` — local-first browser UI for live agent sessions

griffinwork40:
## Summary

Adds `afk web` — a local-first browser surface for live agent sessions, and the fourth consumer of `AgentSession.sendMessageStream()` alongside the REPL, `chat --format stream-json`, and Telegram.

**Zero new runtime dependencies.** `node:http` + hand-rolled SSE, with the frontend bundled by the `esbuild` that was already a devDependency.

```
afk web [--port N] [--host H] [--token T] [--no-open]
```

Sidebar of sessions, streaming transcript with collapsible tool calls, mid-run message queue, cost/token meter, inline approvals, responsive dark UI on the brand tokens from `website/app/globals.css`.

### Why SSE rather than WebSocket

The most common documented complaint about every comparable agent web UI is **session loss on browser refresh** (Claude Code [#28511](https://github.com/anthropics/claude-code/issues/28511), [#48663](https://github.com/anthropics/claude-code/issues/48663), [#54509](https://github.com/anthropics/claude-code/issues/54509)). SSE's native `Last-Event-ID` gives resume for free: the stream replays from the session ledger up to the client's last-seen id, then tails live. That single property drove the transport choice.

### Three invariants worth a reviewer's attention

These are the load-bearing constraints; each is commented at its site and covered by a test.

1. **`elicitationRouter` is a single-slot, last-writer-wins, process-wide singleton** (`private handler` + bare-overwrite `install()`). It cannot host two concurrent handlers. Hence the **session-ownership contract**: sessions created inside the `afk web` process are full duplex; sessions living in any *other* process (REPL, daemon, Telegram) are attach-only via `tailLedger()` and return **409** on prompt/approve. The UI badges them read-only and disables the composer, Send, and queue editor — an enabled control there would be one that can never resolve.

2. **`LedgerRecord` is `{ v: 1; ts: number } & LedgerPayload`** — the payload is *flattened onto* the record, not nested under `.payload`, and `ts` is epoch ms. Reading it as nested produces a silently empty transcript with no error anywhere. This was a real bug found by opening the page; `ledger-adapter.test.ts` now guards it.

3. **The ledger is a projection, not a transcript.** Successful tool output is never persisted (only errors), assistant text caps at 8000 chars, tool inputs at 400. So a tool call replayed after a refresh renders an explicit *"result not available after refresh"* rather than blank space — blank would falsely read as "the tool returned nothing".

### Security posture

Deliberately stricter than the daemon control plane, which is unauthenticated by documented accepted risk. This surface can submit prompts and approve tool use, so it enforces three independent checks:

- **Bearer token** on every `/api/*` call, compared with `timingSafeEqual` behind a length guard (it throws on length mismatch). A `?token=` query param authenticates the document GET only — so the printed URL works when pasted — and is refused on every API route.
- **Origin validation** on all mutating requests. A loopback bind does *not* stop CSRF: any page open in the user's browser can POST to `127.0.0.1`.
- **Refusal to bind a non-loopback host** unless a token is explicitly supplied, so `--host 0.0.0.0` cannot publish a tool-running agent to the LAN under an auto-generated token the operator never saw.

The browser never passes model- or tool-derived text to `innerHTML`; nodes are built with `textContent` and output goes through `stripAnsi` first.

### Performance

Session listing was `O(all sessions)` because it head-read every ledger *then* sorted — **810ms and a 2.36MB payload** on the 10,524 session dirs accumulated on my machine, on every sidebar load, growing without bound. Fixed by ordering: sort on the mtime already gathered during the cheap stat scan, truncate to `DEFAULT_SESSION_LIMIT` (100, owned sessions always included), then head-read only the survivors — **154ms and 25KB**.

### Packaging

`scripts/build-web-ui.mjs` bundles the frontend; **both** the dev (`copy-prompts.js`) and publish (`build-dist.mjs`) pipelines copy the assets through the shared helper, and `build:dist` fails closed on an empty copy. This repo has shipped a broken tarball once before when an asset landed in one pipeline but not the other. `tsconfig.web.json` supplies the DOM lib the Node root project lacks, and `pnpm lint` now runs both projects.

Adding a `'web'` surface touched **eight** independent, already-drifted unions. They are extended per-site rather than unified — merging them is a separate refactor. Note `coerceSurface()` in `awareness/runtime-source.ts` is a non-exhaustive switch over a string, so `tsc` does *not* catch it; without a case, `'web'` silently degrades to `'unknown'`.

## How was this tested?

- `pnpm lint` — clean across both tsconfig projects.
- `pnpm test` — **809 files, 15,398 tests passing**, 2 skipped. Re-run after rebasing onto `ac9179c5`.
- `pnpm audit:env:check`, `scan:env:check`, `audit:chalk:check`, `audit:sdk:check` — all pass. `docs/env-registry.{json,md}` regenerated for the three new vars.
- **139 new tests** in `src/web-server/` + `src/cli/web-options.test.ts`, including every security claim above: 401 without token, query-token refused on API routes, foreign-Origin POST rejected, non-loopback bind refused without an explicit token, 409 on prompt/approve to a foreign session, path traversal blocked. Server tests run a real `node:http` instance on port 0 with native `fetch`, per the `daemon.test.ts` pattern — no mocks, no new deps.
- **Manual browser verification** (not just status codes): loaded the UI, confirmed the sidebar populates, the transcript live-tails a running session, read-only sessions report `disabled: true` on `#prompt`/`#send` in the DOM, and the page loads with **0 console errors**.
- **Published-artifact check**: `pnpm build:dist` emits `dist/web-ui-assets/` (3 files), and `node dist/cli.mjs web` boots and serves the UI from `dist/` the way an `npm i -g` user would.

## Known gaps

Called out rather than left for a reviewer to discover:

- Approvals are wired and unit-tested, but never exercised **through the browser** — no live owned session existed to trigger one.
- The trace timeline and read-only diff view are scoped but unbuilt; `GET /api/sessions/:id/trace` returns raw events with no UI.
- **Tool inputs can display secrets** (my own bearer token was visible in a rendered `browser_open` arg). This is inherited from what the ledger already persists, not introduced here, but scrub-before-persist remains unimplemented.
- Elicitations are surfaced via `GET /api/pending`; the frontend does not yet render approval cards.

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes


griffinwork40:
All citations verified at the ref. Synthesis below.

---

# Code Review — PR #932 · `feat(web): add afk web`

**Target** `griffinwork40/agent-afk` PR #932 · `afk/20260806-183109-420dd4` → `main` · ref **`226739d7`** · 55 files, +5876/−25 (35 new) · regime **full** · MERGEABLE

---

## 🔴 DO NOT MERGE — 1 high, 8 medium, 4 low, 1 nit

One defect is real, reachable on every use of the new surface, and **pinned green by a test the PR left in place**. The fix is two lines. Everything else is non-blocking.

---

## HIGH

### 1. `high · high · spec-compliance/observability` — `src/agent/trace/events.ts:451` · ref:226739d7 · citation-type: file-state (re-read at ref ✅)

**Every `afk web` session silently loses its `session_init_start` trace event**, because the PR widened the TypeScript type and the producer but not the zod validator that gates the write.

The full chain, each link verified at the ref:

| # | Site | State at ref |
|---|---|---|
| 1 | `src/web-server/session-owner.ts:92` | `surface: 'web',` — PR sets it |
| 2 | `src/agent/session/agent-session.ts:233` | `origin: deriveOrigin(config.surface),` |
| 3 | `src/agent/session/session-identity.ts` | PR **added** `case 'web': return 'web';` |
| 4 | `src/agent/trace/types.ts:838` | PR **added** `'web'` to the TS union |
| 5 | **`src/agent/trace/events.ts:451`** | **`origin: z.enum(['cli', 'telegram', 'daemon', 'unknown']).optional(),`** ← **`'web'` missing** |
| 6 | `src/agent/trace/writer.ts:191` | `TraceEventInputSchema.parse(event);` — `.parse()`, **throws** |
| 7 | `src/agent/trace/emit.ts:174-177` | `catch (err) { debugLog(...) }` — **swallows, no rethrow** |

```
origin: z.enum(['cli', 'telegram', 'daemon', 'unknown']).optional(),   // events.ts:451
```

`tsc` cannot catch this — a zod enum is a runtime value, not a type. The result is not a crash: the whole `session_init_start` record (carrying `model`, `resolvedModel`, `origin`, `actor`) never reaches `trace.jsonl`. Per this repo's own `AFK.md`, the witness trace is "the first thing to reach for when reconstructing what happened" — the new surface ships with that hole.

**The sharp part:** `src/agent/trace/session-phase.test.ts:84-87`, under `it('rejects an out-of-union origin or actor')`, asserts:

```js
expect(() =>
  SessionPhasePayloadSchema.parse({ phase: 'session_init_start', origin: 'web' }),
).toThrow();
```

The PR did not update it. So the "**15,398 tests passing**" cited as evidence is exactly what a green suite looks like when a test pins the defect — and widening the enum without touching this test will turn CI red.

This is also the **ninth** surface union, contradicting the PR's own "eight independent unions… extended per-site" accounting.

**suggestion:** Add `'web'` to the enum at `events.ts:451` and move `origin: 'web'` from the "rejects" test to the "accepts every origin + actor enum value" test at `session-phase.test.ts:74`.

---

## MEDIUM

**2. `medium · high · security` — `src/web-server/routes.ts:209`** · diff-context
`handleApprove` gates on `requireOwned(ctx, res, sessionId)` but then resolves by `requestId` alone — `ctx.bridge.resolve(requestId, response)` never checks the pending elicitation's own `sessionId`, though `PendingElicitation` carries one (`elicitation-web.ts:24`, set at `:101`). An authenticated caller can approve session B's tool call through session A's URL. Not an escalation past the token (all owned sessions share it), but it defeats the per-session boundary `routes.ts`'s own header promises.
→ *Pass `sessionId` into `bridge.resolve` and 404 on mismatch.*

**3. `medium · high · security` — `src/web-server/frontend/index.html`** · absence confirmed (`git grep -i content-security-policy` at ref → 0 hits)
No CSP on a page that holds a live bearer token and can approve tool execution. No injection path exists today — verified: zero `innerHTML`/`setAttribute`/`on*` sinks PR-wide, `SAFE_SCHEMES = ['http:','https:','mailto:']` enforced via `safeHref()` (`markdown-dom.ts:2367-2382`), images render as alt-text only, `rel="noopener noreferrer"` unconditional. So this is a missing containment layer, not an open hole. *(Lane B filed this `high`; downgraded — no reachable injection.)*
→ *Add `<meta http-equiv="Content-Security-Policy" content="default-src 'self'">`.*

**4. `medium · high · correctness` — `src/web-server/routes.ts:94-97, 163`** · diff-context · **shadow-verified**
`Last-Event-ID` is clamped *down* (`NaN`/negative → 0) but never clamped *up* against ledger length, and it is a skip threshold in **both** loops. With 5 records and `Last-Event-ID: 100`: replay emits 0, then `position <= replayCount || position <= resumeFrom` suppresses positions 6–100 — the first record reaching the client is **position 101**. Silent starvation, no error, heartbeats still flowing so the connection looks healthy.
A shadow verifier independently re-derived the arithmetic **and refuted normal-path reachability**: `lastEventId` is a per-instance field (`sse-client.ts:23`), `app.ts:130,138` destroys and re-creates the stream on every session switch, and a reload resets it — so the shipped client cannot produce an out-of-range id unaided. Requires ledger truncation/rotation, session-id reuse, or a hand-crafted client. **`high` → `medium` on that basis.**
→ *Clamp `resumeFrom` to the ledger record count, or 400 on out-of-range.*

**5. `medium · high · correctness` — `src/web-server/frontend/render.ts:114,121,128,134,138` vs `158,173,194`** · diff-context
The PR states "output goes through `stripAnsi` first." At the ref, `stripAnsi` is called **only** on tool input/output/diff (`:158,173,194`). Assistant (`:121`), user (`:114`), thinking (`:128`), error (`:134`), and notice (`:138`) text bypass it, so ANSI escapes in model prose render as literal junk. Display-only — every path lands via `textContent`/`renderMarkdown`, never markup — but the stated invariant is broader than the code.
→ *Apply `stripAnsi` in those five branches, or narrow the claim in the PR body.*

**6. `medium · medium · correctness` — `src/web-server/frontend/app.ts:138-141`** · diff-context
The `onEvent` closure never compares its session id to `activeId`. `stop()` (`sse-client.ts:38`) sets `stopped` and aborts, but the reader loop dispatches every event **already decoded in the in-flight chunk** before re-entering `reader.read()` — and it checks `stopped` only *outside* the loop. A chunk in flight during a session switch appends to the newly-selected session's transcript. Narrow window, real.
→ *Capture `id` in the closure; early-return from `onEvent` when `id !== activeId`.*

**7. `medium · high · perf` — `src/web-server/frontend/app.ts:150` + `render.ts:104-105`** · diff-context
`items.push(item)` has no cap or eviction, and `renderTranscript` does `container.textContent = ''` then rebuilds every node from the full array on **every** SSE event — O(n²) over a long session.
→ *Cap retained items and append incrementally instead of full teardown.*

**8. `medium · high · perf` — `src/web-server/routes.ts:120,142,165`** · diff-context
Every `res.write()` on the SSE path ignores its boolean return; no `drain` listener exists anywhere in the diff. A stalled client makes Node buffer the ledger read unboundedly in process memory.
→ *Honour `res.write()`'s return value and await `drain` before the next frame.*

**9. `medium · high · correctness` — `src/web-server/elicitation-web.ts:96-112`** · diff-context · absence confirmed (0 `setTimeout` hits in file at ref)
`enqueue()` settles only via `resolve()` (`:79`), `uninstall()` (`:57`), or the turn's `AbortSignal` (`:108-112`). With the browser closed but `afk web` still running, a tool-approval prompt blocks the agent turn indefinitely.
→ *Add a max-wait timer that auto-declines.*

---

## LOW

**10. `low · high · security` — `auth.ts:110`** — `if (origin === undefined || origin === '') return true;` — the CSRF gate is a no-op for any client omitting `Origin`. Documented tradeoff (`auth.ts:34` comment block) and the bearer header remains the real defense, since a cross-origin form POST cannot set it. Reported for visibility, not as a defect.
**11. `low · high · security` — `auth.ts:133`** — `isLoopback` omits `::ffff:127.0.0.1`, so that literal is treated as non-loopback and refused without a token. **Fails closed** — friction, not exposure.
**12. `low · high · correctness` — `routes.ts:95-96`** — a non-numeric `Last-Event-ID` coerces to `0` and silently replays the entire ledger rather than returning 400.
**13. `low · high · test-coverage` — `stream.test.ts`** — the five tests cover position-numbering, resume-from-2, append-during-tail, no-dupe-on-reconnect, and prompt-close; none supply a malformed or out-of-range id (grep for `NaN|non-numeric|out-of-range|999999` in that range → 0 hits). That is exactly the gap finding #4 lives in.

## NIT

**14.** `src/web-server/frontend/app.ts` is **363 LOC** vs your global 350-LOC rule. No file-size gate exists in this repo and `src/config/env.ts` is 1855 LOC, so this is convention-only, not a break. All other new files are under (server.ts 343, view-model.ts 318, render.ts 313, routes.ts 301).

---

## Verified and cleared — claims that held

- **Token auth** — `randomBytes(32).toString('hex')` (`auth.ts:28`); length compared *before* `timingSafeEqual` (`auth.ts:44`), with a test for the throw-on-length-mismatch case.
- **`?token=` carve-out is document-GET-only** — `dispatch()` applies the bearer check to every `/api/*` path including the stream *before* route matching (`server.ts:203-206`); the frontend deliberately avoids `EventSource` precisely because it cannot set headers (`sse-client.ts:5-7`). The hunt's top hypothesis — a query-token fallback on the SSE route — is **refuted**.
- **Path traversal** — `req.url` is never percent-decoded before the `normalize()` + `startsWith` check, so `%2e%2e%2f` stays literal and 404s. Origin match is exact array membership, not prefix — `http://127.0.0.1.evil.com` correctly rejected.
- **XSS core claim holds** — no HTML sink in 6403 diff lines; scheme allowlist; no ``; no `setAttribute`; `markdown-dom.ts:2610-2617` wraps rendering in try/catch degrading to `textContent`.
- **Perf fix is real** — `session-source.ts` genuinely orders stat → sort → truncate → head-read (`:90-105`, `:182-189`, `:193-207`). *(The 810ms/154ms figures are prose; no benchmark in the diff — unverified.)*
- **Ownership has one source of truth** — a single `Set<string>` passed by reference feeds both the 409 gate (`routes.ts:55`) and the UI's `readonly` badge (`session-source.ts:195`). No drift. Hunt hypothesis refuted.
- **Replay→tail seam is closed** — `fromStart: true` + one continuous position counter; ids are ledger positions, not emit counters; skip operator is `<=`, no off-by-one.
- **Zero new runtime deps** — `esbuild ^0.28.0` is under `devDependencies`; `dist/web-ui-assets/` ships via the existing `"files": ["dist/"]` glob.
- **Build parity** — both pipelines call the shared helper; `build-dist.mjs` fails closed on an empty copy; `pnpm build` sequences the bundle step before `copy-prompts.js`.
- **Conventions** — 3 new env vars registered in `ENV_REGISTRY` + both registry docs; no raw `process.env` in production web-server code; no hand-joined AFK paths; no raw `chalk.`.
- **`chat.ts` −11/+2** is a pure extract-to-shared-module refactor (`jsonDateReplacer`), behavior identical. Not a regression.

### Dropped by verification

- **`src/browser/config.ts` HEADLESS_SURFACES missing `'web'`** — filed `medium`, **REFUTED on reachability**. No production call site passes `opts.surface`; `loadBrowserConfig` (`:216`) falls back to `AGENT_SURFACE`, which `src/cli/index.ts:45` sets process-wide to `'afk'` — already in `HEADLESS_SURFACES` (`:75`). Web sessions launch headless correctly.
- **`slash/types.ts:137 SkillInvocation.source`** — dropped, no production importer branching on it confirmed.
- **"No `removeEventListener` anywhere"** — dropped, not a leak: `container.textContent = ''` discards nodes and listeners together.

### What was not checked

- Citations verified inline against branch HEAD **`226739d7`** via `git show`/`git grep` at that ref. Stated intent: **PR #932 title + body — spec-compliance assessed.**
- **Tests were not executed** — no `pnpm test`/`lint` run; the 15,398-passing claim is taken as reported, and finding #1 explains why it is green regardless.
- `styles.css`, `queue-reorder.ts`, `scroll-pin.ts`, `ledger-adapter.ts`, `view-model.ts` not read in depth.
- `src/agent/session-ledger.ts` (`readLedger`/`tailLedger` internals) — unchanged by the PR, fd and `fs.watch` teardown inferred from the call site plus a passing shutdown test, not read.
- No manual browser run; the "0 console errors" claim is unverified.
- Rate-limiting / brute-force on the bearer check not modeled (256-bit token likely moots it).

---

**Verdict: DO NOT MERGE** — ship finding #1 first (add `'web'` to `events.ts:451`, relocate the assertion in `session-phase.test.ts:84-87`). Findings #2–#9 are non-blocking and appropriate as follow-ups; #2 and #9 are the two worth doing soon, since both sit on the tool-approval path.

## Done

- Reviewed PR #932 end-to-end: resolved target and reviewed ref `226739d7`, captured stated intent from the PR body, ran a 4-lane parallel dimension wave, verified every load-bearing citation inline at the ref, and ran a 2-agent shadow-verify wave on the top claims.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/73fcf415-c5e2-4f0e-a394-75c430a86d16/pr-932.diff`; all findings cite `file:line` re-read at ref `226739d7` via `git show`/`git grep`; the report above is the deliverable.
- Changed in the world: **nothing** — review is read-only. No files edited, no commits, no PR comments, no pushes. The pre-existing untracked `.afk/` in the working tree was left untouched.
- Wave 1 initially failed: both dimension agents hit the 45-minute harness timeout and returned nothing. I re-dispatched as four narrower lanes with precomputed diff line offsets and iteration caps; Lane D still exhausted its cap and returned partial, so I closed its gaps inline (LOC counts, `browser/config.ts` reachability, trace-enum verification).
- Shadow-verify changed two verdicts: the SSE-resume finding dropped `high → medium` (shipped client cannot trigger it unaided) and the CSP finding dropped `high → medium` (no reachable injection path). One filed finding was dropped entirely as refuted on reachability.
- Pending: the report is transcript-only — not written to a file and not posted to the PR, since `/review` is read-only by contract. Say the word if you want it persisted or posted as a PR comment.

---

griffinwork40:
Wave 1.5 + shadow-verify complete. Synthesizing inline (I hold every verified citation; a synthesis fork would only lossily restate it — noted as a deliberate deviation from the skill's Wave 2 dispatch).

---

# Code Review — PR #932

**`feat(web): add \`afk web\` — local-first browser UI for live agent sessions`**
`main` ← `afk/20260806-183109-420dd4` · ref `226739d7` · +5876 / −25 · 55 files · **regime: full**

**Verification performed:** 2 Wave-1 dimension agents → **Wave 1.5 inline verification of all 16 findings against `git show 226739d7:`** → bounded `/shadow-verify` on the 2 `high` claims. **0 fabricated citations. 0 false-absence claims. 1 cross-agent conflict resolved. 1 `high` downgraded to `medium` by shadow-verify.**

---

## Blocking

### 1 · `high` · high confidence · security · `src/web-server/frontend/app.ts:68` · ref:226739d7 · citation-type:diff-context

**finding** — The frontend mirrors the full `/api/*` bearer token into a JavaScript-readable session cookie, and cookies are scoped by host but **not by port**, so any page on any other `http://127.0.0.1:<port>` origin reads the live credential via `document.cookie` and replays it from any client to prompt the agent and approve its tool calls.

**evidence**
```js
document.cookie = `${TOKEN_COOKIE}=${encodeURIComponent(fromQuery)}; Path=/; SameSite=Strict`;
```
No `HttpOnly` (unattainable from JS), no `Secure`, no `Max-Age`. Same value is sent as `authorization: Bearer ${token}` (`app.ts:93`) and compared against the one process secret (`server.ts:202`).

**suggestion** — Set the cookie server-side on the document response with `HttpOnly` + a `__Host-` name and delete the JS write at `app.ts:68`.

**Shadow-verify corrections applied** (independently re-derived, PARTIALLY CONFIRMED):
- The reviewer's original "because `originAllowed()` returns `true` on missing Origin" clause is **not causally load-bearing** and has been struck — an attacker forges `Origin: http://127.0.0.1:<port>`, which `allowedOrigins` (`auth.ts:127-131`) admits anyway.
- Scope is **narrower than first claimed**: only `http://127.0.0.1:<any port>` can read it. `http://localhost:*` and `http://[::1]:*` **cannot** (an IP-literal cookie host does not domain-match a name). Note `allowedOrigins` treats these as one server while the cookie store does not.
- Scope is also **wider in one respect**: the authors' own comment (`app.ts:17-19`) bounds *CSRF* — "the cookie cannot be used to forge an API call" — and that bound is real (`tokenFromCookie` has exactly one call site, `server.ts:282`, inside `serveStatic`; asserted by `server.test.ts:338-344`). It says nothing about **confidentiality**. Without this cookie the token lives only in memory and a `history.replaceState`-scrubbed URL, so the cookie *creates* this exfiltration surface rather than widening one.
- Verifier argued against `critical`: the precondition (attacker JS on a co-resident `127.0.0.1` origin) is real but non-default. **`high` stands.**

---

## Non-blocking findings

### 2 · `medium` · high confidence · correctness/perf-observability · `src/web-server/frontend/sse-client.ts:61,88-91` + `src/web-server/routes.ts:138,157,175-176` + `src/agent/session-ledger.ts:378-380` · ref:226739d7 · citation-type:diff-context

**finding** — A cleanly-ended session's SSE stream terminates normally, but the client cannot distinguish that from a failure, so it reconnects at a **flat 500 ms forever** — ~2 req/s per tab, each re-parsing the entire ledger twice — and this starts on **first page load with zero clicks**.

**evidence**
```ts
this.attempt = 0;                                    // sse-client.ts:61 — resets on every OK fetch
if (this.stopped) return;                            // :88  — only an explicit close() sets this
setTimeout(() => void this.connect(), delay);        // :91  — delay pinned at 500*2^0
if (activeId === undefined && sessions.length > 0) { // app.ts:117 — auto-select, newest-first
```

**suggestion** — Emit a terminal `{done:true}` frame from `handleStream` when the tail exits without an abort, and set `this.stopped = true` in `sse-client.ts` on receipt so `connect()` returns instead of rescheduling.

**Shadow-verify corrections applied** (PARTIALLY CONFIRMED, **downgraded `high` → `medium`**):
- "reconnects with backoff" is **wrong** — `this.attempt = 0` on every successful fetch pins the delay at 500 ms permanently; the 15 s ceiling is never reached.
- "selecting any ended session" **understates** it — `app.ts:117-119` auto-selects `sessions[0]` (newest-first) with no user action.
- The counter-hypothesis (a `mode`/`live` gate before constructing the stream) was hunted and **fails**: `app.ts:138` constructs unconditionally; `isLive()` (`app.ts:110`) feeds only the composer; `mode` is *ownership*, not liveness (`session-source.ts:195`); `server.ts:222` has no `requireOwned` either.
- Downgrade rationale: transcript output is **correct** (`lastEventId` suppresses replay — no duplicate rows). Harm is a status pill flipping `reconnecting`→`open` twice a second forever plus ~2 full ledger parses/sec at idle. That is "perf degraded," not "wrong output."
- Terminal-`closed` frequency sampled empirically: **9 of 10** real ledgers under `~/.afk/state/sessions/` carry one. SIGKILL/crash paths do not.
- `stream.test.ts:72` encodes the untested assumption verbatim — *"a live tail never ends on its own."* Genuine coverage gap, not a design disagreement.

### 3 · `medium` · high confidence · spec-compliance/correctness · `src/web-server/frontend/app.ts:236-256,263-266` · ref:226739d7 · citation-type:diff-context

**finding** — The headline "mid-run message queue" is a write-through, not a queue: `flushQueue` drains every entry in an unconditional loop against a route that answers **202 on acceptance, not completion** (`routes.ts:193`; `session-owner.ts` chains the turn and returns), so entries vanish within one localhost RTT and the ↑/↓/✕ controls at `app.ts:219-221` have nothing left to act on.

**evidence**
```ts
async function flushQueue(): Promise<void> {
  if (activeId === undefined || !isLive()) return;   // no turnActive guard
  while (queue.length > 0) {
```
**suggestion** — Gate the loop on `!turnActive`: flush one entry, then wait for the terminal `done`/`error` ledger record before flushing the next.

*Collateral:* this makes `queue-reorder.ts` (81 LOC) and its 137-line test dead in practice — the only new module with full test coverage is the one the user cannot reach.

### 4 · `medium` · high confidence · security · `src/web-server/routes.ts:196-218` + `src/web-server/frontend/app.ts:321` + `src/web-server/elicitation-web.ts:101` · ref:226739d7 · citation-type:diff-context

**finding** — The ownership check on `/approve` is decorative: `handleApprove` validates the URL's `sessionId` then resolves the elicitation by `requestId` alone, and the frontend depends on that looseness via a `?? activeId` fallback that 409s permanently whenever the operator is viewing a read-only session.

**evidence**
```ts
if (!requireOwned(ctx, res, sessionId)) return;
const resolved = ctx.bridge.resolve(requestId, response);        // routes.ts:211 — sessionId unused
const sessionId = pending.find((p) => p.id === id)?.sessionId ?? activeId;   // app.ts:321
...(extras.sessionId !== undefined ? { sessionId: extras.sessionId } : {}),  // elicitation-web.ts:101
```
**suggestion** — Replace both halves with a session-agnostic `POST /api/approve` keyed on `requestId` (the bridge already indexes on it) and drop the URL session segment. *Both Wave-1 agents proposed this fix independently.*

### 5 · `medium` · high confidence · security · `src/web-server/server.ts` (response heads in `sendJson` + `serveStatic`) · ref:226739d7 · citation-type:file-state → **verified**

**finding** — No `Content-Security-Policy`, `X-Content-Type-Options`, or `X-Frame-Options` is set on a document that holds a live agent-driving credential in a JS-readable cookie, so finding #1 has no second barrier.

**evidence** — Absence confirmed at ref: `git grep -in "content-security-policy|x-content-type-options|nosniff|x-frame-options" 226739d7 -- src/web-server/` → **0 matches**.

**suggestion** — Add `content-security-policy: default-src 'none'; script-src 'self'; style-src 'self'; connect-src 'self'` plus `x-content-type-options: nosniff` and `x-frame-options: DENY` to both response paths.

### 6 · `medium` · high confidence · api-compat · `src/agent/tools/hooks/path-approval-hook.ts:622-627` · ref:226739d7 · citation-type:file-state → **verified at ref**

**finding** — `PathApprovalSurface` was widened to `'web'` (`:80`) and `mapSurface()` extended, but the grant-persistence ternary was not, so every path approval granted from the browser is written to the durable permissions store as `elicit:unknown` — the audit trail loses exactly the provenance the type widening exists to record.

**evidence**
```ts
source:
  surface === 'telegram' ? 'elicit:telegram'
    : surface === 'repl' ? 'elicit:repl'
      : 'elicit:unknown',
```
**suggestion** — Add `surface === 'web' ? 'elicit:web'` to the ternary.

### 7 · `medium` · high confidence · perf-observability · `src/web-server/frontend/app.ts:353` · ref:226739d7 · citation-type:diff-context

**finding** — Every open tab re-runs the full session listing every 10 s — a `readdir` plus one `stat` per session dir (10.5k on the author's own machine) plus 100 ledger head-opens — so the hard-won 154 ms is re-paid continuously rather than once per sidebar load.

**evidence**
```ts
setInterval(() => void loadSessions().catch(() => {}), 10_000);
```
**suggestion** — Return a max-`mtime` fingerprint from `GET /api/sessions`, send it as `If-None-Match`, and answer 304 without the stat scan when unchanged.

### 8 · `low` · high confidence · security · `src/web-server/server.ts:196-199,272,283` · ref:226739d7 · citation-type:diff-context

**finding** — `dispatch` routes every non-`/api/` path to `serveStatic` and returns **before** the bearer check, and inside `serveStatic` the 401 fires only `if (isDocument && ...)`, so `GET /app.js` and `GET /styles.css` are served to any unauthenticated client — an unauthenticated `afk web` fingerprint on the port.

**evidence**
```ts
if (!path.startsWith('/api/')) { await serveStatic(req, res, path, rawUrl, token); return; }  // :196
if (!tokensMatch(token, bearerFromHeader(header(req, 'authorization')))) {                    // :202
const isDocument = path === '/' || path === '/index.html';                                    // :272
```
> **⚠ CONFLICT — resolved.** Agent B's clean read ("`dispatch` runs `tokensMatch` before any route match") is **REFUTED** by lines 196-205 at the ref: the static branch returns first. Agent A's finding stands.

**suggestion** — Apply the `documentAuthed` gate to every static path, not just the document.

### 9 · `low` · medium confidence · api-compat · `src/agent/facets/derive.ts:225` + `src/agent/facets/schema.ts:57,120` · ref:226739d7 · citation-type:file-state → **verified at ref**

**finding** — A ninth, undeclared non-exhaustive surface map the PR did not extend: facet derivation collapses every non-`telegram` source to `'cli'` and the zod enums cannot represent `'web'` at all, so browser-originated sessions are attributed to the CLI in durable facet telemetry (no crash — silent misattribution).

**evidence**
```ts
source: source === 'telegram' ? 'telegram' : 'cli',        // derive.ts:225
source: z.enum(['cli', 'telegram']).optional(),            // schema.ts:57
```
**suggestion** — Add `'web'` to both zod enums and map it through at `derive.ts:225`.

### 10 · `low` · high confidence · test-coverage · `src/web-server/session-owner.ts` · ref:226739d7 · citation-type:diff-context

**finding** — The module owning per-session turn serialization, the busy set, and `closeAll` has zero test file and is never constructed in any test — `server.test.ts` injects `{ owner: fake as never }` throughout.

**evidence** — Absence confirmed at ref: `git grep -ln "SessionOwner|session-owner" 226739d7 -- 'src/**/*.test.ts'` → **0 matches**.

**suggestion** — Add `session-owner.test.ts` covering the `turns` chain (two rapid `submitPrompt` calls serialize) and `isBusy` transitions against a stubbed `AgentSession`.

### 11 · `low` · high confidence · test-coverage · `render.ts`, `sse-client.ts`, `frontend/app.ts`, `cli/commands/web.ts`, `scripts/build-web-ui.mjs` · ref:226739d7 · citation-type:diff-context

**finding** — Five new production modules have no test sibling, including `sse-client.ts`, which owns the client half of the resume contract that the entire SSE-over-WebSocket rationale rests on — and which is where finding #2 lives.

**evidence** — Absence confirmed at ref via `git cat-file -e 226739d7:<module>.test.ts` (5 misses) and `git grep -ln "sse-client|SessionStream" 226739d7 -- 'src/**/*.test.ts'` → **0 matches**.

**suggestion** — Add `sse-client.test.ts` driving a fake `fetch` over a canned `ReadableStream`, asserting `last-event-id` is sent on reconnect and that a normal stream end does **not** reconnect.

### 12 · `low` · high confidence · spec-compliance · `src/web-server/frontend/ledger-adapter.ts:141-145` + `src/web-server/frontend/app.ts:174-179` · ref:226739d7 · citation-type:diff-context

**finding** — The intent names a "cost/**token** meter" but no token counts are accumulated or displayed.

**evidence**
```ts
export interface SessionTotals { costUsd: number; durationMs: number; turns: number; }
```
**suggestion** — Add `inputTokens`/`outputTokens`, read them from the `done` record's usage fields in `accumulateTotals`, and append them in `renderMeter`.

### 13 · `low` · high confidence · convention · `src/web-server/frontend/app.ts` (363) · `src/web-server/server.test.ts` (379) · ref:226739d7 · citation-type:file-state → **verified at ref**

**finding** — Two files ship over the operator's hard 350-LOC ceiling (line counts read directly from the ref; `server.ts` 343, `view-model.ts` 318, `render.ts` 313, `routes.ts` 301 all pass).

**suggestion** — Extract `renderQueue`/`flushQueue`/`wireComposer` (`app.ts:209-275`) into `frontend/composer.ts`. Confirm whether the ceiling binds test files before splitting `server.test.ts`.

### 14 · `low` · high confidence · spec-compliance · `src/cli/commands/web.ts:24-28,56-58` · ref:226739d7 · citation-type:diff-context

**finding** — The command registers `-m, --model` and `--bypass-permissions` beyond the PR's stated CLI surface, and `--bypass-permissions` — which disables the tool-approval gate for every browser-started session — appears nowhere in the PR's three-check security posture.

**evidence**
```ts
.option('--bypass-permissions', 'run browser-started sessions without tool approvals …')
...(options.bypassPermissions === true ? { permissionMode: 'bypassPermissions' as const } : {}),
```
**suggestion** — Document both flags in the PR body's CLI line and add `--bypass-permissions` to the security section, or land it separately.

### 15 · `low` · high confidence · api-compat · `src/cli/web-options.ts:39-44` + `src/cli/commands/web.ts:41` + `src/web-server/server.ts:76` · ref:226739d7 · citation-type:diff-context

**finding** — `resolveWebToken`'s `explicit` flag — commented "load-bearing, not cosmetic" for the non-loopback refusal — is discarded at its only production call site; `startWebServer` re-derives explicitness from `typeof options.token === 'string'`, so a documented security invariant is upheld by coincidence rather than by the field built for it.

**evidence**
```ts
const { token } = resolveWebToken(options.token);                                  // web.ts:41
const tokenExplicit = typeof options.token === 'string' && options.token.length > 0; // server.ts:76
```
**suggestion** — Thread `explicit` into a `tokenExplicit` field on `WebServerOptions` and have `checkBind` consume it.

### 16 · `nit` · high confidence · security · `src/web-server/routes.ts:138` · ref:226739d7 · citation-type:diff-context

**finding** — `handleStream` passes an unvalidated id to `readLedger`, which silently yields nothing for an id failing `isSafeLedgerSessionId`, so a garbage or traversal-shaped id returns a 200 event-stream that never emits instead of a 400.

**suggestion** — Call `isSafeLedgerSessionId` in `dispatch` and answer 400 before opening the stream.

---

## Correction to the PR description

**"Elicitations are surfaced via `GET /api/pending`; the frontend does not yet render approval cards"** — **stale**. `render.ts:247-259` exports `renderApprovals` and `app.ts:311-317` wires it to a 1 s poll. Cards *are* rendered. The gap that actually stands is the adjacent one: they were never exercised through a browser. This matters to a reviewer because it moves finding #4 from dead code into a live path.

---

## Verified clean — claims that survived attack

| Claim | Verdict |
|---|---|
| `coerceSurface()` `'web'` case present (author's self-flagged tsc-invisible trap) | ✅ `runtime-source.ts:178` |
| All eight declared surface unions extended | ✅ (a 9th and 10th undeclared one found — #6, #9) |
| `timingSafeEqual` length guard precedes the call | ✅ `auth.ts:1065` — no `RangeError` oracle |
| Query token refused on every `/api/*` route | ✅ `tokenFromQuery` reachable only in `serveStatic`; `server.test.ts` asserts 401 |
| Token generation | ✅ `randomBytes(32)` = 256 bits, never persisted |
| Path traversal — static assets | ✅ never URL-decoded, `normalize` + `..`/`/` refusal before any `join` |
| Path traversal — session ids | ✅ `readLedger` **and** `tailLedger` both gate on `isSafeLedgerSessionId` |
| XSS — "never passes model-derived text to `innerHTML`" | ✅ grep of all 6403 diff lines for `innerHTML\|outerHTML\|insertAdjacentHTML\|document.write\|new Function` → **3 comment mentions, 0 assignments**; markdown links scheme-allowlisted |
| Origin comparison is exact, not substring | ✅ `http://127.0.0.1.evil.com` rejected, test at diff:990 |
| SSE resume boundary + replay→tail gap | ✅ `fromStart: true` genuinely present, not merely described; `Last-Event-ID: 2` over 3 records emits exactly id 3 |
| Session-listing perf fix ordering | ✅ stat-scan → sort on gathered mtime → `slice(0,limit)` → head-read survivors |
| Zero new runtime dependencies | ✅ `package.json` hunks touch `scripts` only |
| "Both pipelines copy through a shared helper" | ✅ one `copyWebUiAssets`, imported by `build-dist.mjs:107` and `copy-prompts.js:220` |
| `build:dist` fails closed on empty copy | ✅ throw at `build-dist.mjs:123-128` |
| `chat.ts` −11 lines is a breaking change | ❌ **refuted** — verbatim extraction of `dateReplacer`; `--format stream-json` byte-identical |
| Env vars registered + docs in sync | ✅ 3 vars, `AFK_WEB_TOKEN` marked `secret: true`, count 149→152 |
| No secrets committed in the diff | ✅ |

---

## What was not checked

- **Citations verified inline against branch HEAD `226739d7134079574761438a7a75369ca9df4aaa`** via `git show <ref>:<file>` — every `blocking`/`high` citation and every `file-state` citation at all severities. 0 fabricated, 0 diff-only, 0 false-absent.
- **Stated intent: PR #932 title + body — spec-compliance assessed** against it (findings #3, #12, #14; one PR-description correction above).
- **Not run:** `pnpm lint`, `pnpm test`, `pnpm test:coverage`, `build:dist`, or any of the four audit gates. The 809-file / 15,398-test and 810 ms→154 ms figures are **taken on the author's word**, not reproduced.
- **No runtime execution.** Finding #2's 500 ms cadence and status-pill flicker are derived from source, never observed in a browser. The `closed`-record frequency is n=10 from one machine's `~/.afk/state/sessions/`, not a population statistic.
- **Not opened:** `styles.css` beyond its token block, `index.html` `<head>` in full, `markdown-dom.test.ts`, `scroll-pin.*`, `ansi-strip.*`, `view-model.ts` past the `apply()` fold — so **live-path tool-call/result pairing and duplicate-row behavior during an active turn are unassessed**.
- **Not verified:** `website/app/globals.css` brand-token parity (that file is outside the diff — `styles.css` hand-ports hexes with a stated sync caveat); whether `stripAnsi` is ReDoS-safe; whether `audit:env:check` excludes `src/web-server/frontend/**` (moot — no raw `process.env` in the new frontend files).
- **Long-comment prefix convention** (`Invariant:`/`Contract:`/`History:` on ≥15-line blocks) spot-checked only — the new files use it consistently where sampled, but no exhaustive audit was run.
- **Merge conflicts:** the PR base `ac9179c5` is ~13 commits behind `origin/main` (`70373be3`). `gh` reports `MERGEABLE`; no rebase was attempted (read-only).

---

## Merge decision

# ❌ DO NOT MERGE

One confirmed `high` security finding creates a **new** credential-exfiltration surface that this PR's own three-check security posture does not account for (#1), with no CSP as a second barrier (#5) — and the fix is small: move the cookie server-side with `HttpOnly` and delete the JS write. Alongside it, a resource loop that fires on **first page load with zero user interaction** (#2), a headline feature that does not behave as specified (#3), and an ownership check that is decorative on the approval path (#4).

This is unusually strong work — the traversal guards, XSS discipline, `timingSafeEqual` handling, SSE resume boundary, replay→tail gap fix, and listing perf fix all survived direct attack, and the author's self-disclosure was accurate in every particular except one stale line. **Findings #1 and #2 are the merge gate; #3–#7 are worth the same pass. #8–#16 are follow-up.**

*Read-only review. No files were edited, no commits made, no PR state mutated.*

---

griffinwork40 (commented):

**Decision: DO NOT MERGE — 1 high, 2 medium blocking (2 security); 1 medium waived, 11 low, 1 nit.**

Reviewed at `62bcd503`. Citations were re-read against that ref; three high-stakes claims were independently re-derived by a second pass (one consequence was refuted and downgraded — see below).

This is a large, unusually well-documented change (68 files, +7348/−30, 139 new tests). Most of the PR body's security claims verify: `timingSafeEqual` behind a length guard, `?token=` refused on `/api/*`, Origin checked on mutating requests, path traversal blocked (`%2e%2e%2f`, `/foo/../../etc/passwd`, `/....//` all 403), no `innerHTML`/`eval` anywhere in the frontend, 409 on prompt/approve for foreign sessions, non-loopback bind refused without an explicit token. The three blocking findings below are the exceptions.

---

## Blocking

### 1. `high` · security · `src/web-server/static-assets.ts:150`
**The session cookie's value is the bearer token itself, and cookies are scoped by host, not by port** — so another local HTTP service reachable at the same host string receives the full agent-driving credential in a `Cookie` request header, and can replay that value server-side as `Authorization: Bearer`.

```ts
function documentCookie(token: string): string {
  return `${TOKEN_COOKIE_NAME}=${encodeURIComponent(token)}; Path=/; HttpOnly; SameSite=Strict`;
}
```

`HttpOnly` closes the `document.cookie` read path (as the comment at `:144-147` says) but not the automatic *send* to sibling ports. Replay needs no browser and no CORS: a server-side request carries no `Origin`, which is explicitly allowed at `auth.ts:110` (`if (origin === undefined || origin === '') return true;`), and `server.ts:208` accepts the value as a bearer header. Consequence is full duplex — `POST /api/sessions/:id/prompt` and `/api/approve`.

**Calibration (from independent re-derivation):** narrower than "any local service." The cookie is host-only (no `Domain`), so it does not cross between `127.0.0.1` and `localhost`, and `SameSite=Strict` *does* block a cross-site initiator — a page on `evil.com` firing at `127.0.0.1:8080` gets no cookie. The live precondition is a listener on the exact host string the user opened, reached same-site (address-bar navigation, bookmark, or a request from another `127.0.0.1` page). On a dev machine running other loopback servers that is a routine condition, not an exotic one.

**Suggestion:** make the cookie an opaque, server-side-mapped handle rotated per document GET, so a cookie captured by another loopback port is not itself the API credential.

### 2. `medium` · security · `src/web-server/elicitation-web.ts:145-150`
**A scalar `response` body on `POST /api/approve` is coerced to `accept`**, inverting the decline-by-default the function's own contract promises. `{"requestId":"el_1","response":false}` — and `response:"decline"` — both **approve** a pending file-edit or shell-command elicitation.

```ts
  if (typeof response !== 'object' || response === null) {
    return response === undefined
      ? { action: 'decline' }
      : { action: 'accept', content: { value: response } };
  }
```

The doc comment immediately above states: *"An unrecognized or absent `action` DECLINES, whatever the payload shape … a malformed body can never be read as approval."* The existing test (`server.test.ts:593-625`) pins the **object**-shaped fallback (`{foo:'bar'}` → decline); the scalar branch is uncovered. Reachable only from a non-bundled client, since the shipped frontend always sends `{action}` (`render.ts:234-236`) — hence `medium`, not `high`. Per policy a security-dimension `medium` is not waivable.

**Suggestion:** return `{ action: 'decline' }` for every non-object `response`; confine scalar wrapping to a payload carrying an explicit `action: 'accept'`.

### 3. `medium` · security · `src/cli/commands/web.ts:92`
**The live bearer token is passed as an argv element to `open`/`xdg-open`**, publishing it in the process table (`ps aux`) to every other local user — default-on, since `--no-open` is opt-out.

```ts
if (options.open !== false) openInBrowser(handle.url);
```

`handle.url` is built at `server.ts:132` as `http://${displayHost}:${port}/?token=${token}`. `openInBrowser` uses `execFile`, not a shell, so there is **no command injection** here — the exposure is argv visibility only.

**Suggestion:** open the bare `http://host:port/` and deliver the credential outside argv (e.g. a single-use handoff nonce exchanged for the cookie on first document GET).

---

## Non-blocking

**Waived medium** — `src/web-server/server.ts:88-89`: `bridge.install()` is unconditional, including in the attach-only (owner-less) configuration, and `elicitation-router.ts:110-120` is a bare-overwrite single slot with no save/restore, so `stop()` leaves the host with no handler rather than the prior one. *Waived:* `startWebServer`'s only production caller is `src/cli/commands/web.ts:61`, it is absent from `src/index.ts` and the `package.json` export map, so no embedder reaches it today. Latent, not live — exporting it or starting the server inside the REPL/Telegram process makes it blocking.

**Correctness / spec (low)**
- `overload-pause.ts:132` — `INTERACTIVE_SURFACES = new Set(['cli','repl','telegram'])` omits `'web'`, which `session-owner.ts:92` stamps. *Re-derivation refuted the original "fails immediately" claim:* a non-member still gets the full in-loop ladder (3 retries, 5s→10s→20s) and commits the turn on exhaustion, so nothing is lost — it forgoes only the extra 10-minute park-and-probe. One-line Set addition.
- `session-owner.ts:83` — browser sessions skip `assembleSystemPrompt`, so auto-routing never applies (REPL and `chat` both route).
- `session-owner.ts:161-164` — `isBusy()` is dead code whose comment claims it "drives the Stop affordance"; Stop is driven by frontend `turnActive` (`app.ts:215`).
- PR body's Known-gaps clause cites `GET /api/sessions/:id/trace` returning raw events — that route does not exist in `dispatch`; it 404s.
- `--bypass-permissions` and `-m/--model` (`web.ts:22-28`) are absent from the documented invocation; the former removes the third security check and is unmentioned in the security-posture section.
- `facets/schema.ts:26` — `source` enum widened to include `'web'` while `FACET_VERSION` stays `1`, so an older binary treats a `'web'` facet as version-fresh and then fails schema parse.

**Test coverage (low)** — the exact tsc-invisible regression the PR body flags is untested: `runtime-source.test.ts:77` iterates `['cli','repl','daemon','telegram','subagent']` with no `'web'`, so deleting `case 'web'` from `coerceSurface` leaves the suite green. Same for `deriveOrigin` (`session-identity.test.ts:35`, `allowed` set omits `'web'`).

**Observability (low)** — no `console`/`logger` call exists anywhere under `src/web-server/` (verified by grep at the ref), so every 500, swallowed ledger-read failure, and session-start error is visible only to the browser that made the call. Also `stream-route.ts:94` replays without honoring `res.write` backpressure.

**Nit** — `-m, --model` missing from the intent's usage block.

---

### What was not checked
- Citations verified inline against branch HEAD `62bcd503`; absence claims re-verified with `git grep` at that ref. Findings whose cited line did not exist at the ref were dropped before synthesis.
- Stated intent: PR #932 title + body — spec-compliance assessed against it. Gaps the body already discloses were not re-reported as defects.
- Tests, lint, and `build:dist` were **not executed** — every "passes" claim in the testing section is unverified here. Not read: `styles.css`, `index.html` body, `markdown-dom` internals beyond the scheme allowlist, several frontend test bodies, `docs/env-registry.*` regeneration, and the built `dist/` tree.
- Windows `path.win32` behavior in `serveStatic` was reasoned about, not executed.

griffinwork40:
Wave 2 synthesis verified against my manifest — counts, dispositions, and sort order all check out. **Shadow-verify not triggered**: no `critical`/`high` finding exists and all 9 findings sit at their default `blocking` disposition (the rule fires only on critical/high or a departure from the default table).

---

# Review — PR #932 · `feat(web): add afk web`

**Ref `352a1b4a`** · 73 files · +7825/−33 · regime: full · 3 sub-agents (2 parallel dimension + 1 synthesis), citation verification run inline.

PR #932 ships `afk web`: a `node:http` + hand-rolled-SSE server with a vanilla-JS frontend for observing and driving live agent sessions. Engineering quality is high for a first cut this size — timing-safe token comparison, pre-filesystem traversal rejection, zero new runtime dependencies, and every load-bearing invariant commented at its site *and* tested. Two gaps in the auth/prompt path are blocking; the rest are test-coverage and hygiene items.

## Blocking

**1 · `medium · blocking:true · confidence:medium · security · src/web-server/server.ts:84-96`**
`startWebServer` derives `tokenExplicit` and `token` from two independent option fields, so a caller passing `tokenExplicit: true` with no `token` clears `checkBind`'s non-loopback guard while the credential is silently auto-minted — the exact LAN-exposure-under-an-unseen-token footgun `checkBind` exists to block.

```
const tokenExplicit = options.tokenExplicit ?? hasToken;
const bind = checkBind(host, tokenExplicit);
const token = hasToken ? (options.token as string) : mintToken();
```

Reachability, verified exhaustively: the sole non-test caller is `src/cli/commands/web.ts:61`, and `resolveWebToken` (`src/cli/web-options.ts:41-47`) returns `explicit: true` only on branches that also set a non-empty token — the CLI cannot produce the disagreeing state today. No auth bypass results even if triggered, since `mintToken()` is `randomBytes(32)` (`auth.ts:23-29`). What keeps it blocking: `server.test.ts:596-608` (`describe('tokenExplicit is threaded, not inferred')`) covers `token+explicit:false` and `token+explicit:true` but **not** `explicit:true`-with-no-token, and the author's own comment at `server.ts:93-95` states the state is "now expressible" — the token-*read* consequence was handled, the *bind* consequence was not. **Fix:** throw when `tokenExplicit === true && !hasToken`.

**2 · `medium · blocking:true · confidence:high · correctness · src/web-server/routes.ts:86-101`**
`handlePrompt` returns 202 for every accepted prompt with no queue-depth cap, and `SessionOwner.isBusy()` — the only backpressure predicate — is never called anywhere in the tree (`git grep isBusy` at the ref resolves only to its own definition at `session-owner.ts:162`; other hits are unrelated Telegram code). An authenticated client bypassing the frontend chains unbounded turns onto one session.

The original sub-agent framing ("turns race") is **refuted**: `session-owner.ts:125-133` documents per-session chaining and turns demonstrably serialize by design. What survives is the missing cap plus dead code. **Fix:** 409 from `handlePrompt` when `isBusy(sessionId)`, or cap depth in `submitPrompt`.

## Non-blocking

| sev | dimension | site | issue |
|---|---|---|---|
| low | api-compat | `routes.ts:190-193` | `POST /api/sessions` accepts + forwards `cwd`; `SessionOwner.create` discards it unless `allowArbitraryCwd` (CLI never sets it) — inert knob. Suggest 400. |
| low | correctness | `sse-client.ts:29-30,130-132` | Reconnect backoff has no jitter; two tabs retry in lockstep. |
| low | perf | `session-owner.ts:66-70` | `sessions`/`info`/`turns`/`busy` clear only in `closeAll()`; lifecycle-ended sessions leak entries for the process's life. |
| low | correctness | `app.ts:138-151` | No regression test for rapid session-switch mid-`connect()`. Race is already guarded (`sse-client.ts:114-118` + synchronous `stop()`) — test gap, not a live defect. |
| low | test-coverage | `stream-route.ts:47-48` | No test sends malformed/negative `Last-Event-ID`; clamp-to-zero unasserted. |
| low | test-coverage | `session-owner.ts:134-164` | `SessionOwner` has **zero** dedicated tests among 16 files in `src/web-server/`; `server.test.ts` only exercises injected fakes. |
| nit | perf | `stream-route.ts:72-74` | 15s heartbeat ignores `res.write()`'s return; stalled-client backpressure accrues silently. |

## Verified clean

Bearer check gates every `/api/*` pre-dispatch, `tokensMatch` length-guards before `timingSafeEqual`, `?token=` feeds only `serveStatic` (401 on API routes, tested) · traversal rejected pre-filesystem, percent-encoding cannot smuggle `..` · **no XSS sinks** in any production frontend module (only comments documenting the invariant), no inline script, CSP `script-src 'self'` · `checkBind` throws before `listen()` · all eight widened `surface` unions have consuming switches updated + regression-tested in the same diff · `json-date-replacer.ts` extraction behavior-identical · **"zero new runtime dependencies" holds** — `package.json` diff touches only `scripts` · perf reorder holds structurally (stat-scan → sort gathered mtime → slice → head-read survivors; owned ids bypass the slice, pinned by test) · Invariants 1, 2, 3 all implemented and tested · 350-LOC operator rule holds (largest new file `app.ts` = 330).

## What was not checked

Citations verified inline against branch HEAD **`352a1b4a`** — 9/9 `verified`, 0 fabricated, 0 diff-only; 5/5 absence claims `confirmed-absent`, 0 dropped. Wave 1 agents' `Read` saw `main` (`fb90013a`), not the reviewed ref, so every citation was re-derived at the ref by the orchestrator; both agents also reported diff-artifact line numbers, corrected here to true `file:line`. `pnpm test` / `pnpm lint` / `pnpm audit:*` were **not run** — the worktree is on `main` and this review is read-only — so the author's "809 files / 15,398 tests passing, lint clean" is unverified here. Both Wave 1 agents hit their tool-iteration cap and returned partial results; residual gap is an exhaustive consumer-grep of `SessionSurface`/`TraceOrigin` beyond the paths read. No browser or runtime exercise performed. Author-disclosed known gaps (approvals unexercised in-browser, trace UI unbuilt, ledger persists secrets in tool inputs, approval cards unrendered) were treated as disclosed, not re-reported. Stated intent: PR #932 title+body — spec-compliance assessed.

**Decision: DO NOT MERGE — 2 medium blocking (1 security, 1 correctness); 6 low, 1 nit non-blocking.**

---

**Done**
- Ran the full-regime `/review` pipeline on PR #932: target resolution, triage, 2 parallel dimension sub-agents, inline citation + absence verification at the ref, and synthesis; emitted a merge verdict of DO NOT MERGE on 2 blocking mediums.
- Evidence lives at reviewed ref `352a1b4a26e3c3ea037998df6b8b362b0a2dec13` (every finding cites verified `file:line` there) and in the pre-flight diff artifact `/Users/griffinlong/.afk/state/skill-preflight/2355c340-7c67-43d3-836b-3f833de65b5c/pr-932.diff`, confirmed byte-identical to the live diff (8540 lines both).
- Nothing in the world changed: no files written or edited, no git operations beyond read-only `show`/`grep`/`diff`, no PR comment or review posted — the working tree stays on `main @ fb90013a` with the operator's 1 uncommitted change untouched.
- Deferred: shadow-verify was correctly not triggered (no critical/high finding, no `blocking` departure from the default table); `pnpm test`/`lint` remain unverified because confirming them would require checking out the PR branch, which this read-only review does not do.

---

Issue #726, opened 2026-07-27
/review's citation contract forces research-agent to nest git-investigator — a "2-agent" wave is really 4+ concurrent sessions and 429s


griffinwork40:
## What happened

Running `/review 719` (full regime, 2 parallel `research-agent`s). **Both Wave-1 agents died simultaneously** with:

```
429 {"type":"error","error":{"type":"rate_limit_error","message":"This request would exceed your account's rate limit."}}
```

A narrowed retry (single agent, `cwd` pointed at a pre-materialized worktree) then died on an unrelated stream cutoff (#724). Net result: the review's independent adversarial verification never ran at all, and the orchestrator had to verify the diff inline — which is exactly the context-preservation the skill exists to avoid.

## Mechanism (read from the definitions, not inferred from the failure)

The concurrency floor of a full-regime review is structurally **double** its declared width, because the agent type it mandates cannot satisfy the citation contract it mandates:

1. `research-agent` has **no Bash** — `src/skills/_agents/prompts/research-agent.md:4`:
   ```
   tools: Read, Grep, Glob, WebFetch, WebSearch, Agent(git-investigator)
   ```
2. Its prompt **requires** nesting for any git need, and explicitly forbids the cheap workaround — `research-agent.md:21`:
   > **Git needs → dispatch `git-investigator`.** … **Do not substitute `.git/` internals (`.git/logs/HEAD`, `.git/packed-refs`, `.git/refs/`) for proper git commands** — that's a lossy workaround and a contract violation. Use the specialist.
3. `/review` makes git **mandatory** for precisely the findings it cares most about — `src/bundled-plugins/awa-bundled/skills/review/SKILL.md:53`:
   > **Citation requirement (enforced per agent):** Before quoting any line in a `blocking` or `high/critical` finding, the agent MUST read that exact line from the branch HEAD using `git show <reviewed-ref>:<file>`
4. `/review` dispatches two of them at once — `review/SKILL.md:47`:
   > **Wave 1 — Full review (regime=full, 2 parallel agents, `subagent_type: "research-agent"`).**

So a "2-agent" wave is **≥4 concurrent sessions** (2 parents + ≥1 nested `git-investigator` each), and every one of them carries its own copy of a large diff plus its own file reads. Waves 1.5 and 2 add two more `research-agent`s (`review/SKILL.md:80`, `:94`) under the same contract.

[...]

PR #936, opened 2026-08-08
fix(review): split severity from disposition and make medium blocking by default

griffinwork40:
## Problem

`/review` emitted a merge recommendation while **medium**-severity findings were still outstanding. The operator's actual workflow had been: run `/review`, watch it say MERGE, then fix the mediums by hand before merging anyway.

## Root cause: unspecified, not misspecified

`SKILL.md` is the entire implementation for this skill (`context: load` — the orchestrator executes the text directly). The old line 104 said only:

> Emit merge decision: **DO NOT MERGE**, **MERGE**.

It never stated **which severities block**. The model improvised a threshold on every run.

Neither the system prompt nor the Wave 2 synthesis prompt mentions severity or merge at all, so this is a skill-file fix, not a prompt fix.

## What this PR contains

Three commits. PR #937 was stacked on the first commit and merged **into this branch**, so its content ships here — it is not a separate landing.

### 1. `a13cf232` — the merge-decision rule

States the threshold explicitly instead of leaving it to improvisation, and requires the verdict to carry the counts that drove it.

### 2. `4867924e` (#937) — severity and disposition become separate axes

The original rubric welded them: it defined `high` by impact ("wrong output under reachable conditions") but `medium` by category ("missing edge case, perf degraded under load, deprecated API"). Blocking on tier alone therefore made "you used a deprecated API" a hard merge blocker.

- The rubric is now single-axis (**impact**); hygiene categories re-home to `low`.
- Every finding carries an explicit `blocking: true|false`, assigned by Wave 1 from a default table (`critical`/`high`/`medium` → true; `low`/`nit` → false).
- Per-finding **overrides** are allowed but each needs a one-clause justification.
- **Two medium classes are never overridable**: the `security` dimension, and material data-integrity risk / likely production failure under normal usage.
- An "Invariant — assignment order" block pins `blocking` to the **pre-downgrade** severity, so a finding cannot be made non-blocking by being downgraded rather than waived.
- Overridden findings route to `/shadow-verify` — a waiver is otherwise the only decision in the pipeline with no independent reader.

### 3. `86e81989` — gaps found by `/review` of this PR

- The assignment-order invariant enumerated only **two** downgrade paths while the file has **six**. The unnamed ones include the api-compat reachability rule, which drops a finding **two tiers in one step** (`medium` → `nit`) and so was not even described by the invariant's "downgraded a tier" wording. It now covers every downgrade rule in the file.
- Preserving `blocking: true` across a downgrade **contradicted** the override rule permitting `blocking: true` on a `low`/`nit` only for a stated external constraint: every downgraded `medium` landed as a `low`/`nit` carrying `blocking: true` with no admissible justification to write, forcing the reviewer to fabricate an external constraint or emit a schema-violating finding. A downgrade-preserved disposition is now explicitly *not* an override — it records `· blocking preserved from pre-downgrade <severity>` and is exempt from that rule.
- Wave 2's `receives` list still omitted the merge-decision rule it is the actor for — the same gap that had just been closed for Wave 1's blocking table.

## Consumer compatibility

`summarizeForTelegram` (`src/cli/slash/_lib/review-post.ts`) regex-sniffs the prose verdict. It tests `/do not merge/i` **before** `/\bmerge\b/i`, so a blocking verdict can never be misread as an approval. If a model echoes the rule text into a passing verdict, the worst case is a false red — which fails safe.

One real consumer defect **was** found and fixed here: that function keyed its high-signal-line filter on the bare word `blocking`, and the new schema puts `blocking:(true|false)` on *every* finding line — so the filter matched 100% of findings and the 8-line push cap filled with nits. It now matches `blocking: true` only.

The verdict remains free prose. Making it a schema-validated object would be stricter, but `/review` runs `context: 'inline'` (no `outputSchema` path) and both `summarizeForTelegram` and `buildGithubBody` consume the raw string, so that upgrade needs a rendering shim. Deliberately left as follow-up.

## Why this one is worth it

`/review` is the most-dispatched skill in the framework — **1,207 dispatches, 5× the next-most-used** (measured from `routing-decisions.jsonl`, 2,827 skill dispatches total).

## Verification

- `pnpm lint`, `pnpm build` — clean
- `pnpm fix:pins:check` — hash pins up to date (bundled SKILL.md re-pinned `4bac29d1` → `c457514e`)
- `audit:env:check`, `scan:env:check`, `audit:chalk:check`, `audit:sdk:check` — clean
- **Full suite: 809 files, 15420 tests pass**
- Guards added: five substring assertions pin the widened invariant, the exemption clause, and Wave 2's receives entry (`bundled.test.ts`, 20 tests); one new `review-post.test.ts` case asserts a non-blocking finding is not lifted into the Telegram push (34 tests).

## Not done here

Behavioral confirmation against a synthetic medium-only diff (actually watching `/review` return MERGE pre-fix and DO NOT MERGE post-fix) was not run — the root cause was established by reading the skill and its consumers, not by reproduction.


griffinwork40 (commented):

## /review — PR #936

**`fix(review): make medium findings block the merge verdict`**
Target: `afk/review-medium-blocks-merge` → `main` · reviewed ref `a13cf23` · regime **light** (2 files, 8 lines, hotfix)

**Decision: DO NOT MERGE — 1 medium outstanding (2 non-blocking: 1 low, 1 nit).**
Applied threshold: this PR's own proposed rule (critical/high/medium block). Disclosed because the installed skill this run executed predates the rule and states no threshold — the exact gap the PR fixes.

---

### Findings

**1 · medium · confidence: medium · spec-compliance** — `skills/review/SKILL.md:106` vs `:129` and `:96` · ref:`a13cf23` · citation-type: file-state (re-read and verified at ref)

Promoting `medium` to blocking creates a verification asymmetry the diff does not reconcile: a `medium` now decides the verdict but is explicitly exempt from the adversarial re-derivation that `critical`/`high` receive, so a finding whose citation is real but whose *inference* is wrong blocks a merge unchallenged. The sharper edge is `:96` — the `grep-unavailable` path downgrades a `high` one tier into `medium` while tagging it `[UNVERIFIED: absence not checked]`, so an admittedly-unverified claim now lands in the blocking tier and skips shadow-verify on the way.

```
:106  Emit **DO NOT MERGE** when one or more `critical`, `high`, or `medium` findings survive Wave 1.5
:129  **Post-synthesis:** if any `critical` or `high` finding is present, invoke `/shadow-verify` ...
        `medium` and below go straight through.
:96   ... tagged `[UNVERIFIED: absence not checked]` and downgraded one severity tier.
```

Suggestion: extend the `:129` shadow-verify trigger to `medium`, or exclude `[UNVERIFIED]`-tagged findings from the blocking tier at `:106`. Note the PR body already identifies `:129` as part of the root cause ("plausibly biased the improvisation toward passing mediums") — it is the one line the fix leaves untouched. This fails safe (false red, never false green), which is why it is medium and not high.

**2 · low · confidence: high · test-coverage** — `bundled.test.ts:195-196` · ref:`a13cf23` · citation-type: file-state (verified at ref)

The rule's only regression guard is an opaque hash pin, so a future edit that deletes the paragraph passes review by re-pinning — the same class of silent regression the neighbouring `#726` test exists to prevent for the no-git contract.

```
it('review Wave 1 keeps the no-git citation contract (#726)', () => {
  const content = readBundled('review');
  expect(waveOne).toContain('do **not** run git');
```

Suggestion: add one sibling assertion — `expect(content).toContain('blocking by default')` — so the rule is pinned by meaning, not just by bytes.

**3 · nit · confidence: high · perf-observability** — `src/cli/slash/_lib/review-post.ts:181-187` · ref:`a13cf23` · citation-type: file-state (verified at ref)

The new mandatory count-suffix makes the verdict line match the finding-extraction heuristic whenever a model renders it as a markdown heading (`hasSeverity` hits on "high"/"medium", `isBullet` hits on `#`), so the Telegram summary prints the decision twice — once as its header, once as a bullet.

```
const hasSeverity = /\b(critical|high|blocking|medium)\b/i.test(line);
const isBullet = /^([-*]\s|#{1,6}\s)/.test(line);
if ((hasCitation && hasSeverity) || (isBullet && hasSeverity)) {
```

Suggestion: skip a line already consumed as the decision when building `findingLines`.

---

### Red CI is **not** this PR's fault — re-run it

`Test (ubuntu-latest)` failed, but the failure is unrelated to the diff and the branch is otherwise clean:

- Failing test: `src/agent/background-registry.test.ts:757` — `ENOENT ... /state/bg/bg-.../meta.json`, one failure against **15,417 passing** (`1 failed | 808 passed` files).
- That file is **not in this PR** (changed files: `bundled.test.ts`, `skills/review/SKILL.md`), and there is **zero** reference to `background-registry` or `BgJobLogWriter` anywhere under `src/bundled-plugins/` at the ref.
- Failure shape is a fixed-sleep race: `await new Promise((r) => setTimeout(r, 100))` immediately followed by a synchronous `readFileSync` of a file an async writer may not have flushed.
- Every other recent CI run is green, including `#1955` on a sibling branch four minutes earlier and `#1958` three hours later.

Re-run the job; if it recurs, the durable fix is polling for the file instead of sleeping 100ms.

### Verified, not just asserted

- **Hash pin is correct.** `sha256(skills/review/SKILL.md)` at the ref is `827aa8f1…`, matching the new pin exactly. Single copy of the file in the tree — no drift twin.
- **Consumer-compat claim holds.** `review-post.ts:169` tests `/do not merge/i` before `:171` `/\bmerge\b/i`, so a blocking verdict cannot render green.
- **No exact-match verdict consumer exists** that the count-suffix would break: `=== 'MERGE'`, `.startsWith('MERGE')`, `/^merge$/` → 0 matches in `src/`. `buildGithubBody` passes the string through verbatim.
- **No breaking change** for the skill's consumers.

**security:** no issues found — prose-only change, no auth/secret/injection surface touched.

### What was not checked

- Citations verified inline against branch HEAD `a13cf2322cc42bca06c44f4ff6c8cb26fa408cb6` (all 3 file-state citations re-read at the ref; 2 absence claims re-grepped at the ref).
- Stated intent: PR #936 title + body — spec-compliance **assessed** against it.
- Behavioral confirmation that a model actually obeys the new prose rule (MERGE pre-fix vs DO NOT MERGE post-fix on a synthetic medium-only diff) was **not** run — the PR body already discloses this gap, and it remains the one thing that would prove the fix works.
- Tests and lint were not executed locally; CI output was read, not reproduced.
- Zero `critical`/`high` findings survived, so the shadow-verify wave did not fire.

griffinwork40 (commented):

Reviewed ref `62bcd503`. Two candidate `high` findings were raised in the dimension wave and **both were knocked down by ref-anchored verification** before reaching this decision — details under "Findings withdrawn" so the reasoning is auditable rather than silently dropped. Nothing `critical` or `high` survives. Remaining items are `medium` and below and none of them blocks the merge.

---

## Findings withdrawn during verification

**W1 — `SessionPhasePayloadSchema.origin` missing `'web'` — claimed `high`, reduced to `medium` (latent).**
The divergence is real: `src/agent/trace/events.ts:451` is the only `origin` enum in the trace layer and it omits `'web'`, while `TraceOrigin` (`src/agent/session/session-identity.ts:25`) and `SessionPhasePayload.origin` (`src/agent/trace/types.ts:838`) both include it. The claimed consequence — "every `session_init_start` from a browser session fails Zod validation" — does **not** hold at this ref. `SessionOwner.create()` builds its `AgentConfig` with no `traceWriter` (`src/web-server/session-owner.ts:88-99`), `SessionOwnerOptions` exposes no writer field, `src/cli/commands/web.ts` never constructs one, `AgentSession` never self-creates one, and `emitSessionPhase` short-circuits on `if (!writer) return;` (`src/agent/trace/emit.ts:172`) before `TraceEventInputSchema.parse()` at `src/agent/trace/writer.ts:191`. The nested path is also clear: `deriveOrigin(ctx.surface)` in the subagent/skill executors lands in `appendRoutingDecision` → a schema-less `routing-decisions.jsonl` row, never `SessionPhasePayloadSchema`. Retained below as a latent defect, not a live bug.

**W2 — unbounded `Last-Event-ID` — claimed `high`, reduced to `low`.**
The mechanism is real and reproducible by reading: `src/web-server/stream-route.ts:46-49` clamps only negative/non-finite, so an oversized `resumeFrom` suppresses replay (`:92`) *and* live frames (`:115`) indefinitely. But no legitimate client can produce such an id. `lastEventId` is a private per-instance field written only from a received frame (`src/web-server/frontend/sse-client.ts:34,89`); `selectSession()` allocates a fresh `SessionStream` per session and stops the old one (`src/web-server/frontend/app.ts:130,138`); a reload reinitializes to `undefined` (full replay); and the ledger is append-only (`createWriteStream(..., { flags: 'a' })`, `src/agent/session-ledger.ts:223`) with no rotation/truncation/unlink on ledger paths, so a valid id cannot later exceed its own ledger. The only trigger is a hand-crafted request with an already-valid bearer token against one's own loopback server — hardening, not a reachable defect.

---

## Findings

### medium

**1 · api-compat · `src/agent/trace/events.ts:451`** — latent
The Zod enum will reject `origin: 'web'` the moment a trace writer is wired to the web surface, and the throw is swallowed by `emitSessionPhase`'s `catch` → `debugLog`, so the first symptom would be a silently missing `session_init_start` rather than an error. The comment at `src/web-server/session-owner.ts:90-91` ("distinguishes browser-started work from a REPL or daemon session in the witness trace") signals that wiring as intended, which makes this a scheduled regression.
```
origin: z.enum(['cli', 'telegram', 'daemon', 'unknown']).optional(),
```
*Fix:* add `'web'` to the enum at `events.ts:451` so it matches `TraceOrigin` and `SessionPhasePayload.origin`.

### low

**2 · correctness · `src/web-server/stream-route.ts:46-49`** — an over-large `Last-Event-ID` stalls the stream silently instead of clamping or signalling; unreachable from the shipped client (see W2) and untested.
*Fix:* clamp `resumeFrom` to the observed replay count, and add a test for a non-numeric / negative / far-past-end id.

**3 · test-coverage · `src/web-server/server.ts:279`** — the 1MB body cap has no test; `MAX_BODY_BYTES|too large|413` matches zero test files at the ref.
```
if (total > MAX_BODY_BYTES) throw new Error('request body too large');
```
*Fix:* post >1MB to `/api/sessions/:id/prompt` and assert the rejection status.

**4 · spec-compliance · PR description** — the "Known gaps" section states `GET /api/sessions/:id/trace` returns raw events with no UI, but that route does not exist at this ref: `/trace|handleTrace` matches zero files under `src/web-server/`. The route is absent, not merely its UI.
*Fix:* correct the sentence in the PR body.

**5 · perf-observability · `src/web-server/stream-route.ts:72-83`** — `res.write()`'s return value is never checked, so a slow reader accumulates frames in Node's write queue rather than pausing the tail.
*Fix:* pause `tailLedger` iteration on a `false` return until `'drain'`.

### nit

**6 · api-compat · `src/agent/routing-directive.ts:136`, `src/cli/slash/types.ts:136`** — `PromptSurface` and `SlashContext.source` were not widened with `'web'`. Reachability check: no production caller passes `'web'` into either (`session-owner.ts` calls `resolveBaseSystemPrompt` directly, never `assembleSystemPrompt`), so nothing misbehaves today. `[UNVERIFIED: no production importers reach these from the web surface]` — noted only because `tsc` will not catch a future miswiring here, the same class of hazard the description flags for `coerceSurface()`.

**7 · perf-observability · `src/web-server/stream-route.ts:90-95`** — reconnect re-reads the ledger from position 0, so cost scales with total history rather than the gap since disconnect. The inline comment already accepts this ("one extra sequential read of an append-only file") and it is the price of the gap-free handoff — recorded, not objected to.

---

## Verified claims

Each of these was checked against the ref rather than taken from the description:

- **Auth** — `tokensMatch` guards length *before* `timingSafeEqual` (`src/web-server/auth.ts:39-45`); `server.ts:208` gates all `/api/*` ahead of route dispatch; `tokenFromQuery` is called only from `static-assets.ts:82` and never in the API path, so the query token really is document-GET-only.
- **CSRF** — non-GET requests require an allowlisted `Origin` (`server.ts:216`, `auth.ts:105-131`).
- **Bind refusal** — `checkBind` requires an *explicit* token for non-loopback hosts (`auth.ts:154-163`); an auto-minted token is never marked explicit (`src/cli/web-options.ts:39-45`).
- **XSS** — zero `innerHTML`/`insertAdjacentHTML`/`outerHTML`/`document.write` in non-test frontend source; `markdown-dom.ts` walks the lexer token stream with `html` tokens forced to text (`:238-244`) and hrefs scheme-allowlisted (`:38-45`).
- **Token handling** — no `localStorage`/`sessionStorage` use anywhere in `src/web-server`; the replay cookie is `HttpOnly; SameSite=Strict` (`static-assets.ts:150`); the SSE stream authenticates via an `Authorization` header on `fetch` rather than a query param (`sse-client.ts:63-72`).
- **Ownership** — `requireOwned` 409s foreign sessions and is applied to prompt/approve/interrupt (`routes.ts:70-79,93,110,221`).
- **SSE replay gap** — `tailLedger(..., { fromStart: true })` gives one continuous positional cursor (`stream-route.ts:109-117`), covered by a handoff test in `stream.test.ts`.
- **Perf ordering** — `session-source.ts` is genuinely stat → sort-on-mtime → truncate → head-read survivors, with owned sessions unioned in before truncation.
- **Packaging** — both pipelines share `copyWebUiAssets`, and `build-dist.mjs` throws on an empty copy.
- **Zero new runtime deps** — `dependencies` gains nothing; `esbuild` is a pre-existing devDependency.
- **Touched security files** — the `path-approval-hook.ts` / `permissions-store.ts` hunks are additive `'web'` plumbing; no existing gate is weakened.

## What was not checked

- Citations verified inline against branch HEAD `62bcd503` (Check A: 5 citations verified, 0 fabricated; Check B: 3 absence claims confirmed by grep at the ref).
- Stated intent: PR #932 title + body — **spec-compliance assessed**; no unmet intent and no undisclosed scope creep found.
- **Coverage gap, disclosed:** both dimension agents hit their tool-iteration budget before finishing. `chrome.ts`, `view-model.ts`, `queue-*.ts`, and the bulk of `server.test.ts` (~625 lines) were only partially read. Absence of findings there is not evidence of correctness.
- Tests and lint were **not executed** — no `pnpm test` / `pnpm lint` / `audit:*` run, so the 15,398-passing and clean-lint claims are unverified here.
- No runtime exercise of the server: all auth/CSRF/bind conclusions are static reads plus the assertions in `auth.test.ts` / `server.test.ts`, not live reproduction. Browser-side approval rendering was not exercised, matching the description's own disclosure.
- The description's pre-disclosed gaps (approvals untested through a browser, unbuilt trace UI, tool inputs able to display secrets) are treated as disclosed, not re-reported as new findings.

PR #777, opened 2026-07-30
fix(review): stop the citation contract forcing a nested git dispatch (#726)

griffinwork40:
Fixes #726.

## The defect

A full-regime `/review` declares 2 parallel agents but runs **4+ concurrent sessions**, and in #726 both Wave-1 agents died simultaneously on a 429.

The cause is structural, not ambient load. `review/SKILL.md:53` mandated that before quoting any line in a `blocking`/`high`/`critical` finding, the agent MUST re-read it at the reviewed ref via `git show <ref>:<file>`. But Wave 1 dispatches `research-agent`, which has **no shell at all** (`tools: Read, Grep, Glob, WebFetch, WebSearch`), and whose prompt routes every git need to a nested `git-investigator` while explicitly forbidding the `.git/`-internals workaround.

**The skill mandated a command the agent it mandated could not run.** Each Wave-1 agent silently became two sessions.

## The fix — change the contract, not the agent

| change | effect |
|---|---|
| Wave 1 cites `diff-context` only; `file-state` citations tagged `[UNVERIFIED: not re-read]` | no git, no nesting |
| **Wave 1.5 hoisted inline to the orchestrator** | was itself a 4th `research-agent` running `git show`/`git grep` — the same defect one wave later |
| Absence-claim grounding uses the shell-free `Grep` tool | closes the second nesting trigger |
| New "Concurrency floor" block | documents the real budget and says where future shell-needing requirements go |

Wave 1.5 moving inline is the load-bearing half. The orchestrator already holds exactly the read-only shell this verification needs (`git show` / `git diff` / `gh pr diff` / `grep` / `rg`, granted at `SKILL.md:18`), so verification now happens **once, centrally, by an actor that can actually run the commands** — rather than N times by shell-less ones that each have to spawn a child first.

**No verification strength is lost.** Wave 1.5 Check A already re-read every blocking/critical/high citation at the reviewed ref and dropped `fabricated` ones — `SKILL.md:76` already called it a backstop. The per-agent re-read was redundant with a wave that already runs; only its cost was unique.

Net: peak concurrency 4+ → **2**, total dispatches 4 → **3**, nesting → **zero**.

## Why not the issue's preferred option

#726 preferred option 1: materialize the reviewed ref as a worktree and pass it as each agent's `cwd`. I did not implement that, because it introduces a correctness hole in the exact mechanism it is protecting:

[...]

PR #937, opened 2026-08-08
fix(review): make the severity rubric impact-only, block security mediums explicitly

griffinwork40:
fix(review): make the severity rubric impact-only, block security mediums explicitly

The prior commit made medium blocking but left the rubric that produces
medium unchanged, and that rubric mixes two incompatible axes: `high` was
defined by impact ("wrong output under reachable conditions") while
`medium` was defined by category ("missing edge case, perf degraded under
load, deprecated API"). A tier cannot mean both "narrower impact" and
"this kind of thing", so blocking on tier alone was unsound — it made
"you used a deprecated API" a hard merge blocker, which is how a gate
earns a reputation for crying wolf and gets bypassed.

Empirical check against real emitted findings (n=39, hand-classified from
~/.afk/state/transcripts across ~12 PR reviews) found medium is roughly
1:1 defect:hygiene — not hygiene-dominated, so keeping medium blocking is
right. The fix belongs upstream of the threshold: re-home hygiene to
`low` so `medium` actually means "real defect, narrow reachability", and
then medium-blocks is correct by construction.

Same sample found security findings land at medium frequently and with
real teeth (agent-settable env vars enabling self-grant capability
escapes; a redaction guard preserving fused base64/hex secrets; an
approval endpoint skipping its sessionId check; normalizeResult failing
open against its own contract). Severity was never adjusted by dimension
— "security first within tier" is a sort rule, not a severity rule — so
a security medium and a formatting medium blocked identically with no
signal to the reader. The decision line now requires a dimension
breakdown and names a security medium as blocking.

Rubric is now single-axis (impact), with the old category examples
re-homed explicitly by impact rather than by kind.

Verification: pnpm lint clean; pnpm fix:pins re-pinned the bundled
SKILL.md hash; bundled.test.ts (19) + review-post.test.ts (33) = 52 pass.

### DIFF
--- a/docs/bundled-plugins.md
+++ b/docs/bundled-plugins.md
@@ -38,6 +38,36 @@ still saying "grep production source files", the one shell-implying verb the
 rest of this change removed, and Check B verified absence claims only at
 blocking/critical/high — the same severity-scope gap Check A had just closed.
 
+Re-bumped again (#937): severity and disposition are now **separate axes**.
+`severity` answers "how bad is this defect", `blocking` answers "does this
+prevent merge"; the old rubric welded them, defining `high` by impact
+("wrong output under reachable conditions") but `medium` by category
+("missing edge case, perf degraded under load, deprecated API"), so
+blocking on tier alone made "you used a deprecated API" a hard merge
+blocker. The rubric is now single-axis (impact), hygiene re-homes to
+`low`, and every finding carries an explicit `blocking: true|false` from a
+default table with per-finding overrides that each need a one-clause
+justification. Two medium classes are never overridable: the `security`
+dimension, and material data-integrity risk / likely production failure
+under normal usage. Reviewed by /review, which found the first cut
+incomplete in three ways, all fixed here: (1) the two pre-existing
+severity-downgrade rules (low confidence, Wave 1.5 `diff-only`) fed
+straight into the blocking default, so a security `medium` could reach
+`blocking: false` with no override and no justification — an "Invariant —
+assignment order" block now pins `blocking` to the pre-downgrade severity;
+(2) neither wave's dispatch contract transmitted the blocking table and no
+actor was named for the mapping, so Wave 1 was told to emit a field whose
+rules it never received — Wave 1 is now the named assigner and the table
+is in its `receives` list; (3) the shadow-verify trigger still keyed on
+tier while the gate keyed on `blocking`, leaving the newly-introduced
+waiver as the only pipeline decision with no independent reader —
+overridden findings are now routed to /shadow-verify, and the 3-claim
+concurrency bound is prioritised waiver-first because an unreviewed
+waiver removes a blocker while an unreviewed `critical` still blocks. The
+two never-overridable clauses and the ordering invariant are now guarded
+by a substring assertion, not the file hash alone (see 'review keeps the
+severity/disposition split' below).
+
 The bundled copy is the SOLE copy of /review: upstream deleted its own
 review skill on 2026-07-24 (ce27e45), a deliberate dedup of a skill that
 only ever shadowed this one. So #726 had nothing to back-port. The defect
--- a/src/bundled-plugins/awa-bundled/bundled.test.ts
+++ b/src/bundled-plugins/awa-bundled/bundled.test.ts
@@ -101,7 +101,7 @@ const PINNED_HASHES = {
   research: 'abe79d75a5f3c74696ef002293dbe8714e446f8955de97089d1005f1e70bc269',
   // History: /review Wave 1 no longer mandates a `git show` re-read (#726, #777).
   // Full rationale: docs/bundled-plugins.md#review-726
-  review: '15083aee942b69f8c5e4516c6def5c656f86a84ed769118b4b4618e1434d1551',
+  review: 'd92f10a3d0a1290c1f00e2141476a256c08b285854ce1639aaf0a1414fc4d394',
   // History: /shadow-verify gained the confidence-trigger + composition-axis
   // verdicts (#52, #187).
   // Full rationale: docs/bundled-plugins.md#shadow-verify-52
--- a/src/bundled-plugins/awa-bundled/bundled.test.ts
+++ b/src/bundled-plugins/awa-bundled/bundled.test.ts
@@ -101,7 +101,7 @@ const PINNED_HASHES = {
   research: 'abe79d75a5f3c74696ef002293dbe8714e446f8955de97089d1005f1e70bc269',
   // History: /review Wave 1 no longer mandates a `git show` re-read (#726, #777).
   // Full rationale: docs/bundled-plugins.md#review-726
-  review: '827aa8f10c6de141bce19b5f7bac9394b319ca9cb68c108ed627a6d5f2c2df20',
+  review: '15083aee942b69f8c5e4516c6def5c656f86a84ed769118b4b4618e1434d1551',
   // History: /shadow-verify gained the confidence-trigger + composition-axis
   // verdicts (#52, #187).
   // Full rationale: docs/bundled-plugins.md#shadow-verify-52
--- a/src/bundled-plugins/awa-bundled/bundled.test.ts
+++ b/src/bundled-plugins/awa-bundled/bundled.test.ts
@@ -99,9 +99,12 @@ const PINNED_HASHES = {
   // ~/.afk/skills/ if it drifts back.
   refactor: '3adf801b9a61eba80afd34fef1e8c78a892ec07256dabb073370622a62d1b40f',
   research: 'abe79d75a5f3c74696ef002293dbe8714e446f8955de97089d1005f1e70bc269',
-  // History: /review Wave 1 no longer mandates a `git show` re-read (#726, #777).
+  // History: /review Wave 1 no longer mandates a `git show` re-read (#726,
+  // #777); severity and disposition split into separate axes with an explicit
+  // `blocking` field, never-overridable security/data-integrity mediums, and a
+  // pre-downgrade assignment-order invariant (#937).
   // Full rationale: docs/bundled-plugins.md#review-726
-  review: 'd92f10a3d0a1290c1f00e2141476a256c08b285854ce1639aaf0a1414fc4d394',
+  review: '4bac29d1b06c80afde003094b77315c56f71aa39f04c7115da268cbafdde06db',
   // History: /shadow-verify gained the confidence-trigger + composition-axis
   // verdicts (#52, #187).
   // Full rationale: docs/bundled-plugins.md#shadow-verify-52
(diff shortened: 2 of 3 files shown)


griffinwork40:
Pushed `40e3edfd` in response to review feedback arguing that blanket medium-blocking is too aggressive. The architectural half of that critique was right and is now adopted; the policy half rests on two premises that don't hold for this particular reviewer.

## Adopted: severity and disposition are now separate axes

The objection that severity was secretly encoding policy is correct. `severity` answers *how bad is this defect*; `blocking` answers *does this prevent merge*. Welding them means the tier drifts toward whatever the gate needs it to mean.

Every finding now carries an explicit `blocking: true|false` from a default table, and the verdict keys on that field rather than on tier:

| severity | default `blocking` |
|---|---|
| `critical` | true |
| `high` | true |
| `medium` | true |
| `low` | false |
| `nit` | false |

Overrides are per-finding and each requires a one-clause justification. Waivers are never silent — a waived medium appears in the decision counts with its reason.

## Not adopted: flipping the medium default to non-blocking

Two arguments were offered for that flip. Both are sound in general and don't transfer here.

**"Medium is the junk drawer of legitimate-but-nonfatal defects."** True of the *previous* rubric, where `medium` literally listed "deprecated API" — that is exactly what the parent commit fixed. Severity is now impact-only and hygiene re-homes to `low`, so the medium population is roughly the defect half of a measured 1:1 defect:hygiene split (n=39 real emitted findings sampled from ~12 PR reviews). The junk drawer was the disease; it has been drained.

**"Reviewers will downgrade real mediums to `low` to get past the gate."** This failure mode requires an actor with an incentive to sandbag. Findings here are produced by dispatched sub-agents in a single automated pass with fresh context per run, then filtered by Wave 1.5. Nobody is grading their own homework, and no reviewer carries a grudge against the gate across runs.

There is also a product reason: the originating complaint for this whole change was that `/review` said MERGE while mediums were outstanding and they were being fixed by hand first. Non-blocking mediums restore that behavior.

## The reviewer's strongest point is now handled explicitly

The race-condition example — a defect that intermittently loses user state but doesn't earn `high` on reachability — is real, and the fix names it. Two medium classes are explicitly **non-waivable**: `security` dimension, and material data-integrity risk / likely production failure under normal usage.

## The policy is now a one-row edit

Because disposition is a field rather than a threshold, changing the medium default no longer touches the rubric's structure — it is one cell in the default table. If the non-blocking default is preferred after living with this, that is a one-line change rather than a redesign, and the override machinery keeps working either way.

griffinwork40:
## /review — PR #937

**`fix(review): make the severity rubric impact-only, block security mediums explicitly`**
`afk/review-rubric-impact-axis` → `afk/review-medium-blocks-merge` (stacked, not main) · +35/−8 · 2 files · head `40e3edfd` · base `a13cf232`

Regime: full. Waves 1 (×4 dimension agents) → 1.5 (inline citation + absence verification) → 2 (synthesis) → shadow-verify on the one `high`. Line numbers are from the **post-change** file (156 lines).

**Verified mechanically before review:** `sha256(SKILL.md @ 40e3edfd)` = `d92f10a3…` = the pin at `bundled.test.ts:104`. Pin is correct.

---

### high

`high · blocking:true · confidence:high · correctness · skills/review/SKILL.md:98,122,144 · The PR's headline guarantee — a security medium is never waivable — has an automatic bypass, because two pre-existing rules downgrade the severity tier after assignment and the new blocking value is keyed off whatever tier currently reads, with no override event and no justification clause.`

```
L144: Confidence `low` → auto-downgrade one tier + append `[low confidence …]`
L98:  `diff-only` citations are passed to Wave 2 … and auto-downgraded one severity tier.
L122: A `medium` in the `security` dimension is **never** overridable to `false`
```

L122 bans the *override act* on a `medium`; it says nothing about a finding that is no longer tagged `medium` because a downgrade already moved it to `low` → default `blocking:false`. Exhaustive absence check for a sequencing or floor rule (`precedence|order|before|after|recompute|retain|floor`) returned only an unrelated "concurrency floor" at L47 — **no such rule exists**. Both trigger paths are routine, and the diff-only path is fully automatic. Independently re-derived by a shadow verifier: **CONFIRMED, tier agreed** (it defeats the change's stated purpose, which L102's own sort rule ranks atop `high`).

**suggestion:** one sentence after the override table — "`blocking` is assigned from the pre-downgrade severity; a downgraded finding retains its pre-downgrade blocking value unless an override is explicitly invoked."

### medium

`medium · blocking:true · confidence:high · correctness · SKILL.md:55,100,110 · The finding schema now mandates a `blocking` field on every finding, but neither wave's dispatch contract transmits the table that assigns it, and no line names which actor performs the severity→blocking mapping.`

```
L55:  Each agent receives: full diff + file tree + triage header + reviewed ref (SHA)
      + the stated intent …, the severity rubric, and the finding schema.
L110: Every finding carries an explicit `blocking: true|false`, assigned from this default table:
```

L110 is passive with no actor. `blocking` appears nowhere in Wave 1's or Wave 2's `receives` list (verified by direct read of L51–56 and L100). Agents are told to emit a field whose assignment rules they were not given — improvisation in a section headed "do not improvise a threshold." Fires on every run.
**suggestion:** add "the blocking default table and its overrides" to L55's receives list, or assign the mapping explicitly to Wave 2.

`medium · blocking:true · confidence:high · correctness · SKILL.md:126,156 · The merge gate now decides on `blocking`, but the shadow-verify escalation trigger still keys on severity, so the newly-introduced waiver — the only way a real defect legitimately stops blocking — is the one decision in the pipeline with no adversarial re-derivation.`

```
L126: Emit **DO NOT MERGE** when one or more findings carry `blocking: true` …
L156: if any `critical` or `high` finding is present, invoke `/shadow-verify` … `medium` and below go straight through.
```

L156 is byte-identical at base, so the escalation gap is inherited — but the PR is what introduces a second axis and a self-granted waiver, making the mismatch load-bearing. The same agent finds the defect, sets its severity, and grants its own waiver; L128's "never waive silently" is transparency, not review.
**suggestion:** change L156's trigger to "any `critical`, `high`, or waived finding," or require waivers to be granted at synthesis rather than by the finding agent.

### low

`low · blocking:false · confidence:high · correctness · SKILL.md:132,136,139 · The rubric's own endpoints contradict its headline: `critical` and `nit` remain pure category lists and `low` mixes seven categories in, two lines above "A category is never a tier by itself."` — `low` absorbing "deprecated API with no removal date" is the concrete edge: a security-dimension finding phrased as a deprecated API lands below the never-waivable rule. Mitigated in practice by L139's re-home instruction, which a careful reader applies first. **suggestion:** qualify the `low` bullet's category list with "absent an impact claim," or add a security-dimension floor.

`low · blocking:false · confidence:high · perf-observability · SKILL.md:136,141 · Perf findings default toward non-blocking, since `medium` requires "measured" evidence while `low` is the no-evidence case and Wave 1 agents are shell-less (`research-agent`, L62).` — **Shadow verifier REFUTED the stronger form of this claim** and I am reporting it demoted: "measured" is reachable without Bash via the stated-intent/PR body (L55) or `Read`/`Grep` of a committed benchmark, so the rung is not structurally sealed, only biased. Base listed "perf degraded under load" at blocking `medium`; this is a deliberate loosening, not an accident. **suggestion:** none required — optionally name the non-Bash evidence paths that qualify as "measured."

`low · blocking:false · confidence:high · test-coverage · bundled.test.ts:102-104, docs/bundled-plugins.md:15-17 · The pin was bumped but its inline comment still describes only #726/#777, and no rationale entry was added to the doc the comment links to.`

```
// History: /review Wave 1 no longer mandates a `git show` re-read (#726, #777).
// Full rationale: docs/bundled-plugins.md#review-726
review: 'd92f10a3…',
```

`git diff --name-status` confirms `docs/` is untouched. The convention is explicit and repeatedly followed: the doc's intro says "the inline comments summarize and link back here," and the `review-726` section already carries three appended re-bump paragraphs ("Re-bumped after review feedback on PR #777…", "Re-bumped again…"). The comment now misdescribes the change that produced the hash. **suggestion:** append a paragraph to `#review-726` and add `#937` to the inline comment.

`low · blocking:false · confidence:high · test-coverage · bundled.test.ts:145-154 · The only guard on the new blocking semantics is whole-file SHA equality, so a future edit inverting "security mediums never waivable" passes CI after a routine `pnpm fix:pins`.` — The same file already demonstrates the countermeasure: `'review Wave 1 keeps the no-git citation contract'` (:195-210) is a `toContain` assertion added in #726 to protect exactly this kind of invariant. **suggestion:** add a `toContain` assertion for the two never-overridable clauses.

### nit

`nit · blocking:false · confidence:high · correctness · commit body only · The empirical basis for keeping medium blocking (n=39, 1:1 defect:hygiene, hand-classified from ~/.afk/state/transcripts) ships no artifact and is unreproducible by any future reader; the decision it supports is independently defensible without it.` **suggestion:** label it a hand-classified heuristic, or commit the sample.

---

### Clean reads

- **api-compat — no issues found.** `review-post.ts:167-194` does no structured parsing: decision detection is a whole-text regex, finding-line selection is three independent line regexes, and `blocking` is *already inside* the `hasSeverity` alternation — inserting it as field #2 cannot shift anything. `buildGithubBody` (:153-158) is pure passthrough. Repo-wide, zero files split or index on `·` (365 hits, all cosmetic UI). `preflight/index.ts` and `review-pr.ts` run before the skill and never read its output.
- **Duplicate copies — no issues found.** `git ls-tree` at head: exactly one tracked `skills/review/SKILL.md`. `dist/` is gitignored (`.gitignore:8`); the `.afk-worktrees/` copies are local scratch. Per `docs/bundled-plugins.md`, the bundled copy is the sole copy — upstream deleted its own in `ce27e45`.
- **Test-count claim — accurate.** Claimed 19 + 33 = 52: `bundled.test.ts` = 16 parameterized + 3 standalone = 19; `review-post.test.ts` = 33 real `it(` after excluding 8 false-positive `/re/.test()` hits.
- **Net direction.** The change is a genuine improvement to the axis it targets: base had no override mechanism at all, so "blocking by default" was inert and the tier mixed impact (`high`) with category (`medium`). Every finding above is a local fix inside the same file.

### What was not checked

- Citations verified inline against head `40e3edfd` via `git show`; base comparisons against `a13cf232`. The working tree was never mutated — `git fetch` wrote remote refs only; no checkout, stash, or reset.
- This is a **stacked** PR: the diff is against `afk/review-medium-blocks-merge`, not `main`. Findings inherited from that parent (notably L156) are flagged as such; I did not review the parent branch itself.
- Not run: `pnpm lint`, `pnpm test`, `pnpm fix:pins:check`. The working tree is on an unrelated branch, so a local run would test the wrong code — the PR's pass claims are unverified here. Confirm from CI.
- Not checked: GitHub Actions workflows for anything scraping `Decision:` from PR comments; non-TS consumers; PR comments/CI logs for #937; whether these gaps are already known and deferred.
- Not sampled: `~/.afk/state/transcripts` (out of scope) — so the n=39 claim was assessed for reproducibility, not correctness.

**Decision: DO NOT MERGE — 1 high, 2 medium blocking (3 correctness); 0 waived, 4 low, 1 nit.**

## Done

- Ran the full `/review` pipeline on PR #937 — 4 parallel dimension agents, inline Wave 1.5 citation/absence verification, and a shadow-verify pass on the `high` finding, which **refuted** one sub-agent claim (the perf-rung "structural seal") and caused me to demote it from `medium` to `low` before surfacing.
- Evidence: review body above, with every claim cited to `file:line` at head `40e3edfd`; pin verified as `sha256 == d92f10a3…`; absence claims verified by exhaustive grep, not assertion. Working copies of both file versions at `/tmp/pr937-head-SKILL.md` and `/tmp/pr937-base-SKILL.md`; preflight diff at `~/.afk/state/skill-preflight/353eb3ea-507f-43c4-8446-dcc69ac17d92/pr-937.diff`.
- World unchanged: read-only. No edits, commits, pushes, or PR comments. The one `git fetch` wrote remote refs only; your dirty working tree on `afk/subagent-output-capture` @ `b8f0d103` is untouched.
- Pending: the blocking finding is a one-sentence fix (specify that `blocking` is assigned from pre-downgrade severity) — nothing was written to the PR, so posting the review or applying fixes is yours to trigger.
- Verdict: **DO NOT MERGE — 1 high, 2 medium blocking (3 correctness); 0 waived, 4 low, 1 nit.**

---

PR #945, opened 2026-08-08
fix(anthropic-direct): counted TTFB retry budget with shorter per-attempt bound

griffinwork40:
## The measured problem

When the Anthropic API accepts a request but returns **no first byte**, a watchdog aborts it at `AFK_MODEL_TTFB_TIMEOUT_MS` (code default 180s; this operator runs 300s). Until now each round got exactly **one** re-drive after such a stall.

From a real failed run:

| Metric | Value |
|---|---|
| Model round trips in one child sub-agent | ~25 |
| Round trips that hit the 300s TTFB bound | **4** |
| Pure dead time burned | **>1,200s** |
| Observed stall rate per request | **~16%** |
| P(at least one stall across 25 requests) | **~99%** |
| Sub-agents killed this way in one day | **2** |

The child blew its 45-minute wall clock. The failure mode is a round that stalls **twice**: attempt 1 stalls, the single re-drive stalls, the turn dies.

## The invariant preserved

`loop/retry-budget.ts` documented, on `ttfbRetried`:

> "A boolean, not a counter: the TTFB retry is deliberately once-per-round so it cannot stack on top of the overload backoff into a longer worst case."

That is load-bearing, and a naive "allow 3 retries" would **triple** the worst case — strictly worse. This PR keeps the guarantee but enforces it with **arithmetic** rather than by refusing to count: **shorter per-attempt bound × more attempts**.

`AFK_MODEL_TTFB_TIMEOUT_MS` stays the per-**ROUND** budget and is divided into `TTFB_MAX_ATTEMPTS` shorter attempts by `ttfbAttemptTimeoutMs`.

## Worst-case arithmetic, before vs after

```
before (boolean):  2 attempts x B              = 2B
after  (counted):  3 attempts x floor(2B/3)   <= 2B    for all B >= 3
```

| Configured `B` | Before | After | Attempts |
|---|---|---|---|
| 180s (code default) | 2 x 180s = **360s** | 3 x 120s = **360s** | 2 -> **3** |
| 300s (this operator) | 2 x 300s = **600s** | 3 x 200s = **600s** | 2 -> **3** |

Worst-case wall time per round **does not rise for any configured value**, while P(a round exhausts its budget) falls from **~2.6% to ~0.4%**.

**Why the bound is derived, not hardcoded.** A fixed per-attempt constant (e.g. "3 x 45s = 135s") reads the env var as a per-attempt bound and turns a 300s override into 3 x 300s = **900s — a 1.5x regression on the very config this fixes**. Deriving from `B` is what makes the guarantee hold for every configuration.

**Why shorter attempts are safe.** A long bound only pays off when you get one shot. 120s at the default still sits well above the measured p99 TTFB (~85s), so effectively no legitimately slow prefill is newly cut. A stalled call returns *nothing*, so waiting longer on it buys nothing — against this bimodal failure, trading bound length for attempt count is strictly better.

**The residual this does cut, stated honestly.** A request whose *genuine* TTFB lands in `[floor(2B/3), B)` — `[120s, 180s)` at the default — completed before and now errors. That slowness is a property of the prefill, not a transient stall, so every re-drive reproduces and re-aborts it: **retries do not rescue this band**. Two things make the trade still correct rather than merely favourable. The band sits entirely above the measured p99 (~85s), and the observed population is transient-stall-shaped — precisely the class extra attempts *do* rescue. For a workload that genuinely lives in that band the remedy is to raise `AFK_MODEL_TTFB_TIMEOUT_MS` or set it to `0`, which the env-registry NOTE already documents.

## Scope

- `ttfbRetried: boolean` -> `ttfbRetries: number`, bounded by `TTFB_MAX_RETRIES = 2` (3 attempts). Both read sites (`stream-consumer.ts` in-band + throw paths) now go through `canRetryTtfb()`; the single spend site is `round-retry.ts`.
- **Overload (3) and stream-incomplete (2) budgets untouched** — own counters, own backoffs, no budgets merged. The TTFB re-drive still serves no backoff.
- **No new env var.** `AFK_MODEL_TTFB_TIMEOUT_MS=0` disable survives the derivation (0 in, 0 out), and a positive budget can never floor to a watchdog-disabling 0.
- Retry/witness events unchanged in shape: `emitTtfbRetry` still emits the `ttfb_timeout` phase + `stream.retry`, now with `metadata.attempt` (1-based) so a trace shows *which* re-drive fired.
- Worst case is stated as a checkable comment in `retry-budget.ts` and re-derived in tests rather than trusted.

## What was tested

New end-to-end regressions in `loop.ttfb.test.ts`:

1. **`recovers from TWO consecutive first-byte stalls — the case the boolean lost`** — the measured failure shape; now recovers on attempt 3.
2. **`exhausts its whole budget in no more wall time than the pre-count regime`** — measures elapsed fake-clock time and pins it at or below 2 x 180s, so re-lengthening the per-attempt bound fails here.

New unit tests in `loop/retry-budget.test.ts` (`TTFB worst-case wall-time bound`) re-derive the product across a spread of configured values (180s, 300s, 120s, 45s, 1/2/3/7ms, 999999) plus the `=0` escape hatch and the 1ms floor. The one honest boundary is asserted, not hidden: below 3ms the 1ms timer floor wins and the product may exceed the legacy figure by at most 3ms — documented on `ttfbAttemptTimeoutMs` and pinned by `overshoots by at most the 1ms timer floor on a sub-3ms budget`.

**Verified to fail without the fix.** Reverting only the behaviour (`TTFB_MAX_RETRIES = 1`, no bound shrink) while keeping the API fails 5 tests, including test 1 with `expected { Object (type, error) } to be undefined` — the operator's exact symptom.

## Follow-up commit — review fixes (`8e3b674d`)

Making a round re-drivable twice falsified three strings written under the one-shot regime, and left this PR's own observability claim unmet. Fixed here rather than deferred, because each is a falsehood *this* change introduced:

- **`trace.ts`** rendered a hardcoded `request re-driven once` for the `ttfb_timeout` phase — accurate under the boolean, false under the counted budget. It now reports `metadata.attempt`, which also makes the "a trace shows *which* re-drive fired" bullet above actually land: nothing read that key before, so the metadata was written and never surfaced. The count is **omitted rather than defaulted** when absent, so pre-counted-budget traces are not retro-labelled with a fact they never recorded. The sibling `rate_limit` renderer also ignores `attempt`, but that is a pre-existing gap and is deliberately left alone.
- **`types.ts`**'s `ttfb_timeout` doc comment claimed the abort fires at `AFK_MODEL_TTFB_TIMEOUT_MS` (180s) and re-drives *once*; the abort now fires at the per-attempt bound (120s at the default) and re-drives while the budget holds allowance. Corrected in place — the surrounding rationale for splitting this phase out of `rate_limit` documents a real prior misdiagnosis and is preserved verbatim.
- **`AFK_MODEL_TTFB_TIMEOUT_MS`'s description** opened by calling itself a per-request bound on "a single model call" and later called itself the per-ROUND budget. An operator who stopped at the first sentence would size the value 1.5x too small.
- Two stale `loop.ttfb.test.ts:179` cross-references in `loop.stall.test.ts` now cite the invariant test **by name**, so the pointer cannot rot again.
- New **`trace-ttfb-render.test.ts`** (4 tests) pins the renderer against both a counted event and a pre-counted-budget event with no `attempt`.

## Gates

```
pnpm lint                                        clean (tsc --noEmit)
pnpm test src/agent/providers/anthropic-direct/  60 files, 651/651 passed
pnpm test (full suite)                           810 files, 15,431 passed | 2 skipped
pnpm scan:env:check                              in sync (149 vars)
pnpm audit:env:check                             852 files, 0 violations
```

`docs/env-registry.{json,md}` were regenerated via `pnpm scan:env` (never hand-edited). Every touched **non-test** source file stays under the 350-LOC ceiling — largest is `stream-consumer.ts` at 340 (**+2**, not unchanged). `loop.ttfb.test.ts` is 424 lines; test files are not held to that ceiling and no CI gate enforces LOC in this repo either way.


griffinwork40:
## Review — counted TTFB retry budget

**Reviewed at ref `0021c40c`** · 11 files, +395/−66 · full-regime review (2 parallel dimension agents → every `file-state` citation and absence claim re-read at the ref → adversarial shadow-verification of the two decision-driving claims).

**Verdict: MERGE.** No `critical` or `high` finding survived verification. The core change — a counted TTFB budget with a *derived*, shrinking per-attempt bound — is arithmetically sound, correctly scoped to one round, fully migrated, and pinned by tests that genuinely fail without it. Everything below is a one-line edit or a body-claim correction.

---

### medium

**M1 · api-compat · `src/config/env.ts:336-345`** (regenerated into `docs/env-registry.json:539`, `docs/env-registry.md:44`)

The `AFK_MODEL_TTFB_TIMEOUT_MS` description asserts per-single-call semantics in its opening sentence and per-round-budget semantics later in the same string, so an operator who stops at the first sentence sizes the value 1.5× too small.

```
'Per-request time-to-first-token timeout (ms) ... Bounds how long a single model call
 may stall BEFORE its first streamed CONTENT token ... This is the per-ROUND budget: it
 is divided into 3 shorter first-byte ATTEMPTS (each ~2/3 of this value...)'
```

*Fix:* open with "Per-round first-byte budget (ms)" and drop "a single model call" from the first sentence; re-run `pnpm scan:env`.

**M2 · perf-observability · `src/cli/commands/trace.ts:474`** (untouched by this PR)

This PR falsifies a user-facing string it does not touch: the `ttfb_timeout` renderer hardcodes "re-driven once", which was accurate under the boolean and is wrong for the 2nd re-drive the counted budget now permits.

```ts
return line('ttfb-stall', `no first token${waited} — request re-driven once${srcBit}`);
```

*Fix:* render `` `request re-driven (attempt ${p.metadata?.['attempt'] ?? 1})` ``.

*Shadow-verified CONFIRMED, with one counter I accept:* the sibling `rate_limit` phase also writes `attempt` and its renderer also ignores it (`trace.ts:452-463`), so the unread key is a pre-existing repo-wide convention — only the now-false "once" is caused by this PR. That counter is why this is `medium`, not `high`.

**M3 · perf-observability · `src/agent/trace/types.ts:733-735`** (untouched by this PR)

The `ttfb_timeout` phase's own doc comment states both a bound and a retry count this PR invalidates — the abort now fires at 120s (per-attempt), not at the 180s configured value, and up to two re-drives occur.

```
// content token arrived within `AFK_MODEL_TTFB_TIMEOUT_MS` (default 180s), so
// the request was aborted and re-driven once.
```

*Fix:* "…arrived within the per-attempt bound (⌊2/3⌋ of `AFK_MODEL_TTFB_TIMEOUT_MS`, 120s at the default), so the request was aborted and re-driven while the round's counted budget held allowance."

### low

**L1 · spec-compliance · PR body, "Why shorter attempts are safe"**

The safety argument defends a newly-cut band with a statistic that does not cover it. A request whose genuine TTFB lands in `[⌊2B/3⌋, B)` — `[120s, 180s)` at the default — completed before and now errors: that slowness is a property of the prefill, not a transient endpoint stall, so each of the 3 re-drives reproduces and re-aborts it. The band lies entirely *above* the cited p99 ≈ 85s.

Shadow-verified CONFIRMED — including that prompt caching does not break the chain (a TTFB timeout means attempt 1 never streamed, so no cache write is readable to attempt 2). The counter I accept: the measured population (~25 round trips in a growing conversation) is warm-cache-dominated and transient-stall-shaped, which is exactly the class more attempts *do* rescue. So the trade is sound and the cut band is narrow — and the **shipped** env-registry NOTE already describes the new behaviour correctly. Only the body's justification overclaims.

Worth noting the asymmetry: the sibling `AFK_MODEL_STALL_TIMEOUT_MS` entry documents p99/p99.9/max across 86,076 rounds; TTFB has p99 only.

*Fix:* name the residual in the body — "cuts the `[2B/3, B)` deterministic-slow prefill band; retries do not rescue it; remedy is raising `B` or `=0`."

**L2 · spec-compliance · `src/agent/providers/anthropic-direct/loop/round-retry.ts:57-60`**

The claim "`metadata.attempt` … so a trace shows *which* re-drive fired" is unmet — the key is written but no renderer, aggregator, or insights path reads it (`afk trace show` counts `ttfbStalls` only; `--json` is an unprocessed NDJSON dump). Fixing M2 satisfies this.

**L3 · test-coverage · `src/agent/providers/anthropic-direct/loop.ts:153-156`**

No test drives one round through both stall shapes, so this comment's shared-allowance claim is unpinned:

```
// round that alternates between the two stall shapes still gets exactly
// TTFB_MAX_ATTEMPTS first-byte attempts in total.
```

Absence grounded at the ref: `connectionStall` appears in exactly one test (`loop.ttfb.test.ts:130`), never co-occurring with `postHeaderStallStream` in a single body.

*Fix:* one test alternating `connectionStall` → `postHeaderStallStream`, asserting `callCount === TTFB_MAX_ATTEMPTS`.

**L4 · spec-compliance · PR body, "Gates"**

Two inaccuracies, verified by `wc -l` at the ref: `stream-consumer.ts` went 338 → **340** (not "unchanged in size"), and touched file `loop.ttfb.test.ts` is **424** lines, over the 350 ceiling the claim invokes. No CI gate enforces LOC here, so nothing is red.

*Fix:* scope the claim to non-test source; change "unchanged in size" to "+2".

### nit

**N1 · `src/agent/providers/anthropic-direct/loop.stall.test.ts:13,131`** — both cross-references to `loop.ttfb.test.ts:179` are stale; at this ref line 179 is a timer-advance loop inside a different test, and the named invariant test moved to line 284. Cite it by name instead of line.

---

### Verified clean — the load-bearing risks, re-derived at the ref

- **Arithmetic holds.** `3 × ⌊2B/3⌋ ≤ 2B` for every `B ≥ 2` (B=3→6≤6, B=4→6≤8, B=7→12≤14). Only B=1 overshoots, by 1ms, and it is documented and pinned. `CONFIGURED.filter(b => b >= TTFB_MAX_ATTEMPTS)` is over-conservative, not hiding a counterexample.
- **Rename fully migrated.** `ttfbRetried` survives at the ref only inside a historical test comment (`loop.ttfb.test.ts:202`). Both read sites go through `canRetryTtfb()` (`stream-consumer.ts:162,282`); the single spend site is `round-retry.ts`.
- **One production consumer.** `resolveTtfbTimeoutMs()` is called exactly once in production (`loop.ts:118`), now wrapped. The openai-compatible provider has zero coupling to the var, to `armFirstByteTimeout`, or to the resolver — its `model_ttfb` phase is an unrelated mechanism. `RoundRetryBudget` is not a package export.
- **Budget is genuinely per-round.** `retry.reset()` (`loop.ts:194`) sits below every retry `continue`, so the counter cannot be released mid-round.
- **Trace schema compatible.** `metadata` is an open `Record<string, string | number | boolean>` in both `types.ts:815` and the Zod schema (`events.ts:441`); `attempt` was already convention on the sibling phase.
- **No sibling test invalidated by the shrink.** The two untouched timing tests are immune: one clears the timer via first byte before advancing 700s, the other short-circuits on `ttfbAttemptTimeoutMs(0) === 0`. This was the highest-risk correctness question and it came back clean.

### What was not checked

- No gate was run: `pnpm lint`, the 651/651 and 15,427-pass suite claims, `scan:env:check`, and `audit:env:check` are taken at face value.
- `docs/env-registry.{json,md}` provenance (regenerated vs. hand-edited) unverified — content is consistent with `env.ts`.
- Telegram/CLI surfaces, the openai-compatible provider's own stall handling, and the real TTFB tail distribution (no trace data read).
- Shadow-verification covered 2 claims in one round; the `medium`/`low` findings below them were not independently re-derived.

<sub>Citations re-read at `0021c40c` via `git show` / `git grep`; no fabricated citations and no refuted absence claims.</sub>

Issue #832, opened 2026-08-01
tracking: file-size triage — 146 files over 350 LOC, and why LOC is the wrong metric


griffinwork40:
Tracking issue for the file-size / structural-debt triage run on 2026-07-31.

## Status — 2026-08-30

**All Tier 0/1/2 sub-issues are now closed.** #822, #824, #826, #829 shipped 2026-08-01. #823, #825, #827, #828 closed subsequently. Checkboxes updated below.

**Baseline drift:** was 146 files / 756 code files / 177,571 LOC on 2026-07-31; then 144 / 794 / 181,850 on 2026-08-02. Re-measure with `pnpm audit:filesize:check` for current numbers.

## Baseline

Measured across `*.ts`/`*.tsx`/`*.js`/`*.mjs`/`*.cjs`, excluding `*.test.*`, `*.spec.*`, `tests/`, `__tests__/`, `*.d.ts`:

- **756** non-test code files, **177,571** LOC (original 2026-07-31)
- **146 files (19.3%)** over a 350-LOC ceiling

Reproduce:

```bash
git ls-files '*.ts' '*.tsx' '*.js' '*.mjs' '*.cjs' \
  | grep -v -E '(\.test\.|\.spec\.|/tests?/|/__tests__/|\.d\.ts$)' \
  | xargs wc -l | sort -rn
```

## Headline finding: LOC is the wrong metric

About **40% of the largest files are false positives** — large from breadth (a flat registry, an exhaustive wire schema, a chain of small handlers), not entanglement. Splitting them buys rule compliance and nothing else.

The signal that actually predicts maintenance pain is **largest single function**. Six functions individually exceed the whole-file ceiling:

| Function | LOC | File | Issue |
|---|---|---|---|
| `runTurn` | ~1065 | `providers/anthropic-direct/loop.ts` | #828 |
| `bootstrapSession` | 887 | `cli/commands/interactive/bootstrap.ts` | #823 |
| `commitAbove` | ~730 | `cli/terminal-compositor.committed-band-commit.ts` | #827 |
| `forkSubagent` | 540 | `agent/subagent.ts` | #829 |
| `query()` | 425 | `providers/anthropic-direct/index.ts` | #824 |
| `createSession` | 357 | `src/telegram.ts` | #825 |

### Tier 0 — highest leverage (do first)

[...]

PR #950, opened 2026-08-09
fix(subagent): wind down gracefully on the wall-clock deadline instead of aborting

### LINKED ISSUE #938: subagent: wall-clock timeout kills a fork with no wind-down, while tool-round exhaustion synthesizes an answer
## Problem

When a forked subagent exhausts its **tool-round budget**, it gets a graceful wind-down: one final tools-stripped round where the model synthesizes a real answer from what it already gathered (`src/agent/providers/openai-compatible/query.ts:512, 646, 754, 786`; shared with anthropic-direct via `shared/tool-loop-cap.ts`).

When the same subagent hits its **wall-clock timeout**, it is simply killed. No wind-down round. The parent receives whatever fragment the child happened to have streamed.

Two budget exhaustions, two completely different outcomes.

## Evidence

Telemetry (`~/.afk/agent-framework/routing-decisions.jsonl`, 27,077 dispatches): **200 subagent timeouts**, the second-largest failure class after rate limits.

Observed directly in one session (2026-08-08), five subagents were lost this way. Two concrete cases:

- A gap-analysis agent ran **38 minutes**, then died at the 45-minute cap. Preserved partial output, in full: `"Now let me extract structured metadata (theme + session_count) from ALL briefs at once instead of reading each individually — much more efficient."` Forty minutes of research reduced to one sentence of narration.
- An implementation agent died at the cap having already confirmed a defect. Preserved output: `"Defect confirmed exactly at lines 166-176 ... Note: the file is at src/agent/subagent/idle-watchdog.ts, not src/agent/idle-watchdog.ts"` — genuinely useful, but accidental. It survived only because it happened to be the last thing streamed.

In both cases the child *had* the findings in context. Nothing asked it for them before killing it.

## Not to be confused with partial-output preservation (already done)

`renderForkOutcome` (`src/agent/tools/skill-executor/fork-result.ts:55-80`) and `tools/subagent/failure-payload.ts` already attach whatever the child streamed before dying, on both `cancelled` and `failed` paths. That work is correct and is **not** what this issue asks for.

The distinction:
- **Partial-output preservation** (exists) — salvages a *fragment*: whatever bytes happened to be emitted.
- **Wind-down** (missing on the timeout path) — asks the model to *synthesize* an answer from everything it accumulated.

## Proposal

Two-phase deadline for forked turns:

1. **Soft deadline** at `timeout - windDownReserve` (reserve default ~60s, named constant, env-configurable per the `src/config/env.ts` + `ENV_REGISTRY` pattern) — stop dispatching new tools and run one tools-stripped wind-down round, reusing the existing `windDown` machinery rather than inventing a second path.
2. **Hard deadline** at the full timeout — kill unconditionally, so a wedged or non-responsive child cannot extend its own life by stalling in wind-down.

Result must still be marked failed/incomplete (`isError` stays true, the marker names the timeout) so a synthesized summary can never be misread as a completed task — same invariant `fork-result.ts` already documents.

## Design constraints

- The timeout currently surfaces as an abort through `AbortGraph`, which cascades to children. A soft deadline must **not** cascade — the child needs to keep its own stream alive long enough to answer.
- Both providers must move together; `shared/tool-loop-cap.ts` exists precisely so the two cannot drift.
- Nested forks: a parent in wind-down should not be waiting on grandchildren. Wind-down implies tools stripped, so no new dispatch — verify that holds for in-flight children.

## Acceptance

- A subagent that exceeds its soft deadline returns a synthesized summary of findings, marked incomplete.
- A subagent that ignores wind-down still dies at the hard deadline.
- Existing tool-round-cap wind-down behavior is unchanged.
- Test coverage for: soft-deadline fires → summary returned; hard deadline fires → killed; normal completion → unaffected.

---
Found during a framework audit, 2026-08-08. Full evidence: `~/.afk/workspace/reports/afk-introspection-2026-08-07/00-SYNTHESIS.md`.

griffinwork40:
Closes #938.

## The defect

A forked subagent's wall-clock budget is enforced **outside** its turn loop. `subagent/handle.ts:242` wraps the whole `streamToFinalMessage` call in `withTimeout`, which at the deadline calls `controller.abort(err)` synchronously and rejects on the same tick (`agent/timeout.ts:115-122`). The loop is never consulted.

That is right for a **wedged** child and wrong for a merely **slow** one. A child three quarters of the way through useful work is killed with everything it learned unsynthesized. The partial-output plumbing preserves whatever raw text happened to be mid-stream, but a raw fragment is not an answer — it is usually a half-written sentence, and it is empty when the child died mid-tool-call.

The concrete cost: a `/forge` run burned **54 minutes** and surfaced a single line, `qualify iteration 1 failed: Operation timed out after 2700000ms`.

Meanwhile the ROUND budget has been graceful for a while — `shouldWindDown()` (`shared/tool-loop-cap.ts:74`) triggers one tools-stripped round so the model synthesizes an answer. Same exhaustion, opposite handling.

## The change

A **soft deadline** — hard budget minus a clamped reserve — checked at round boundaries. Once passed, the loop runs exactly one tools-stripped round with a time-flavored note and completes with `soft_deadline_wind_down`. The hard abort stays armed underneath as the backstop.

Policy lives in a new `providers/shared/soft-deadline.ts`, deliberately **separate** from `tool-loop-cap.ts`: that module owns the ROUND trigger, this one owns the TIME trigger. Same downstream mechanism, different question, so they stay different modules rather than one predicate with two meanings.

Reserve is 15% of the budget clamped to [30s, 5min] — a 45-minute fork winds down at 40 minutes.

## Three things worth reviewer attention

**1. Both providers, deliberately.** Provider drift is the exact failure mode `shared/` exists to prevent — openai-compatible went without the round wind-down for a while after anthropic-direct got it. Trigger, tool-stripping, note flavor, and stop reason are all mirrored (`loop/tool-round.ts:148-158` / `query.ts:669-671`).

**2. Two arming sites, not one.** Forks derive from their settled wall-clock budget (`subagent.ts`). DAG nodes have a **second** enforcer that does not route through `agent/timeout.ts` — `runDAG` arms its own per-node timer (`dag.ts:229`) and cascades into `handle.cancel()`. It is armed from `min(nodeTimeout, forkBudget)`: deriving from the node timeout alone would, whenever it is the larger, place the soft deadline *after* the abort that actually fires, arming a wind-down that can never run.

**3. Classification, not just mechanism.** `soft_deadline_wind_down` is added to `isIncompleteStopReason`, so a wind-down partial reaches a parent model behind the PARTIAL RESULT banner rather than as a conclusion. Without that arm, mint's six phase guards accept "here is what remains unresolved" as a real spec/plan/research artifact. The textless-wind-down salvage in `handle.ts` covers both triggers for the same reason — matching only the round-budget reason sent a soft-deadline child down the `StreamIncompleteError` path and failed the fork loudly for precisely the condition this change exists to handle gracefully.

## What is deliberately unchanged

- **`SUBAGENT_DEFAULT_TIMEOUT_MS` is not touched.** Raising the ceiling makes a bad outcome slower; winding down makes the same ceiling productive.
- **Budgets at or below 120s get no soft deadline.** Below ~2 minutes no split leaves both meaningful working time and a usable synthesis reserve, so carving one out would trade a small chance of a graceful answer for a large chance of winding down a child that had barely started. Short-budget forks keep prior behaviour byte-for-byte.
- **Top-level sessions are untouched** — `softDeadlineMs` is set only at the two fork sites; both loops default to `0` (off).

## Two limits, both deliberate (neither is a bug)

1. The check fires at **round boundaries**. A single tool call that hangs past the hard deadline never reaches one, so no wind-down occurs — that case belongs to `subagent/idle-watchdog.ts`, which bounds time-since-output.
2. The synthesis round can itself overrun the reserve and be cut off by the hard abort. Strictly better than today: the attempt costs one round, and whatever it streamed is still preserved as partial output.

## Verification

```
pnpm lint                 tsc --noEmit, exit 0
pnpm test                 813 files, 15495 passed | 2 skipped, no OOM
pnpm build                PASS
audit:sdk:check           PASS
audit:env:check           PASS
scan:env:check            PASS
audit:chalk:check         PASS
fix:pins:check            PASS
```

Full suite, not scoped — this repo has a documented full-suite-only V8 heap-OOM class that narrow runs miss.

The five new classification tests were **verified to fail against the unfixed code** before being accepted, so they are genuine regression guards rather than tests written to match current output.

Four existing test files pin the old lossy timeout behaviour (`subagent.test.ts:949`, `pause-ceiling.test.ts:104`, `handle.test.ts:279`, `fork-result.test.ts:81`). **All four pass unmodified** — their budgets are 1000ms / 60000ms, below the 120s floor, so the soft deadline resolves to `0` and behaviour is identical. No assertion was rewritten to accommodate this change.

One incidental fix: `handle-streaming.test.ts` asserted `/capped/i` on a subagent whose **id** was `subagent-capped-test`, so the match was satisfied by the interpolated id and would have passed even with an empty marker body. Now asserts the budget phrase.

## Two disclosures

- **`query-runtime.ts` goes 350 → 353 LOC**, crossing the repo's file-size rule. The three lines are mechanical plumbing mirroring the existing `maxToolUseIterations` field (declaration, constructor assignment, facade getter). The rule's prescribed remedy is extraction, and the natural seam — `turnDriverContext()` — would require widening ~17 `private readonly` fields to move it out, degrading encapsulation to buy a line count. Flagging rather than making the design worse; see #832 on LOC being the wrong metric here.
- **Closure reason maps to the existing `timeout`**, adding no new reason. The budget genuinely did expire, and `lastStopReason` retains the fine grain, so the wind-down-vs-hard-kill distinction is recoverable from the trace. But it is lossy at the field operators read in `afk trace show` — an adversarial review flagged this and I am recording the disagreement rather than burying it. Say the word and I will split it into its own closure reason.

## Not done

No `AFK_SOFT_DEADLINE_MS` escape hatch — the reserve policy is derived, not configurable. Easy to add if operators want to tune it.

### DIFF
--- a/src/agent/dag-subagent.test.ts
+++ b/src/agent/dag-subagent.test.ts
@@ -137,6 +137,71 @@ describe('runSubagentDAG', () => {
     expect(configs.filter((c) => 'maxToolUseIterations' in c)).toHaveLength(1);
   });
 
+  // --- SOFT WALL-CLOCK DEADLINE ARMING (issue #938) ---
+  // A DAG node is bounded by a SECOND wall-clock enforcer that does not route
+  // through agent/timeout.ts: runDAG's own per-node setTimeout, cascaded into
+  // handle.cancel(). It had the identical lossy-kill gap, so it is armed here.
+  it('arms softDeadlineMs on node forks from nodeTimeoutMs', async () => {
+    pushHandle('a');
+    const manager = managerFromQueue();
+
+    await runSubagentDAG({
+      manager,
+      parentSession: makeParent(),
+      nodes: [{ id: 'A', systemPrompt: 's', promptBuilder: () => 'p' }],
+      edges: [],
+      // 8 min × 0.15 = 72s reserve (inside both clamps). Smaller than the
+      // 45-min fork default, so the node budget is the one that binds.
+      nodeTimeoutMs: 8 * 60_000,
+    });
+
+    const fork = manager.forkSubagent as unknown as ReturnType<typeof vi.fn>;
+    const config = (fork.mock.calls[0]![0] as { config: Record<string, unknown> }).config;
+    expect(config['softDeadlineMs']).toBe(8 * 60_000 - 72_000);
+  });
+
+  it('binds on the SMALLER of the node budget and the fork budget', async () => {
+    // A node timeout larger than the fork's own wall-clock budget must not
+    // place the soft deadline after the abort that will actually fire — that
+    // would arm a wind-down which can never run.
+    pushHandle('a');
+    const manager = managerFromQueue();
+
+    await runSubagentDAG({
+      manager,
+      parentSession: makeParent(),
+      nodes: [{ id: 'A', systemPrompt: 's', promptBuilder: () => 'p' }],
+      edges: [],
+      nodeTimeoutMs: 10 * 60 * 60_000, // 10h, far beyond the 45-min fork default
+    });
+
+    const fork = manager.forkSubagent as unknown as ReturnType<typeof vi.fn>;
+    const config = (fork.mock.calls[0]![0] as { config: Record<string, unknown> }).config;
+    // Derived from the 45-min fork budget, not the 10h node budget.
+    expect(config['softDeadlineMs']).toBe(40 * 60_000);
+  });
+
+  it('omits softDeadlineMs when the node budget is short or unset', async () => {
+    // Unset → forkSubagent derives its own from the fork budget; short → the
+    // min-budget guard keeps prior behaviour exactly. Either way this layer
+    // must not stamp a value.
+    pushHandle('a');
+    pushHandle('b');
+    const manager = managerFromQueue();
+
+    await runSubagentDAG({
+      manager,
+      parentSession: makeParent(),
+      nodes: [{ id: 'A', systemPrompt: 's', promptBuilder: () => 'p' }],
+      edges: [],
+      nodeTimeoutMs: 1_000, // below SOFT_DEADLINE_MIN_BUDGET_MS
+    });
+    const fork = manager.forkSubagent as unknown as ReturnType<typeof vi.fn>;
+    expect(
+      'softDeadlineMs' in (fork.mock.calls[0]![0] as { config: Record<string, unknown> }).config,
+    ).toBe(false);
+  });
+
   it('two-node chain: output of first feeds promptBuilder of second', async () => {
     pushHandle('first-output');
     pushHandle('second-output');
--- a/src/agent/dag-subagent.ts
+++ b/src/agent/dag-subagent.ts
@@ -15,6 +15,8 @@ import type { SubagentManager } from './subagent.js';
 import { runDAG, type DAGEdge, type DAGNode, type DAGRunResult } from './dag.js';
 import { attachSubagentContext, annotateIfIncomplete } from './subagent/result.js';
 import { TimeoutError } from '../utils/errors.js';
+import { resolveSoftDeadlineMs } from './providers/shared/soft-deadline.js';
+import { resolveSubagentTimeoutMs } from './subagent/constants.js';
 
 export interface SubagentDAGNode {
   id: string;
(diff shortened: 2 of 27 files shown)


griffinwork40:
**Done**

- Ran the full-regime `/review` pipeline on PR #950: resolved target → ref `6e6b139`, captured stated intent from the PR body, dispatched Wave 1 (2 parallel `research-agent`s), ran Wave 1.5 citation + absence verification inline against the ref, adjudicated, and ran one bounded shadow-verify round on the verdict-flipping refutation.
- Evidence lives in durable locations, not just this transcript: the diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/1d827a90-1f71-4f7a-99b1-6cddd81cddda/pr-950.diff`; every citation re-derivable via `git show 6e6b139:<file>` in `/Users/griffinlong/Projects/open_source/agent-afk`; the review verdict pushed to Telegram chat 8189013500.
- World state unchanged — this was read-only. No files written, no commits, no pushes, no PR comments or state mutations. The working tree remains `main @ 720c4d35` with its one pre-existing untracked path (`.afk/`), exactly as the preflight required.
- Deferred: the review reports findings but does not fix them — the four surviving low/nit items (doc wording in `soft-deadline.ts:53-54` and the inherited `WIND_DOWN_NOTE` doc, the closure-reason split, two missing tests) are the author's to act on. The `dag-subagent.ts:111` residual risk (correctness load-bearing on an unenforced invariant) is named in "what was not checked" rather than filed as a finding, because it is not a defect today.
- Two process facts the operator should know: the first Wave 1 pair timed out at 45 minutes each with zero output on model `sonnet` (resolves to `claude-sonnet-5`, the congested lane) and succeeded only when re-pinned to `claude-sonnet-4-6`; and Wave 2 synthesis ran inline rather than as a dispatched agent, disclosed in the report's epistemic-scope section.

---

Issue #952, opened 2026-08-09
max_tokens truncation is invisible on every live surface; a dropped mid-turn tool call reads as the agent stalling


griffinwork40:
## Summary

When a turn is cut at `max_tokens`, AFK classifies it correctly as `ClosureReason: 'truncated'` — and then shows the user nothing. On the REPL, Telegram, and one-shot `chat`, a truncated turn is indistinguishable from a clean completion. The only way to learn it happened is to run `afk trace show` afterwards.

The worst variant is a turn cut mid-tool-call: the tool is silently dropped and never dispatched, so the agent announces an action and then simply stops. It reads as the agent giving up mid-task.

Not reachable in a default config (the cap is unset by default → model ceiling, so truncation needs a 64k+ token response), but it becomes the dominant symptom for anyone who sets `--max-output-tokens` / `AFK_MAX_OUTPUT_TOKENS`.

## Evidence

**1. The closure reason never leaves the trace stream.**
`classifyClosureReason` (`src/agent/session/closure-reason.ts:83-97`) maps a trailing `max_tokens` / `length` stop reason to `'truncated'` via `isTruncationStopReason` (line 47). That feeds `emitClosureEvent` (`src/agent/session/closure-emitter.ts:123-151`), whose only renderer is `formatTrace`'s `case 'closure'` in `src/cli/commands/trace.ts:407-423`. Nothing in `src/cli/commands/interactive/` or `src/cli/render/` keys off `ClosureReason` / `'truncated'` / `lastStopReason` in the live turn path, and `src/telegram/streaming.ts` has zero references.

The REPL footer's "model output may be silently truncated" line (`turn-handler.ts:919`) is a **pre-emptive context-window heuristic** driven by `formatContextUsage` — it fires on context percentage, not on confirmed truncation, so it is neither necessary nor sufficient here.

**2. A turn cut mid-text just ends.**
`emitNonToolUseTerminal` (`src/agent/providers/anthropic-direct/loop/turn-terminal.ts:26-100`) has an explicit branch that synthesizes an operator-facing notice for `stop_reason === 'refusal'` (lines 41-59), but no equivalent for `max_tokens`. The accumulated partial text is emitted as an ordinary `assistant.message`: no marker, no ellipsis, no notice.

**3. A tool call cut mid-arguments is silently dropped.**
[...]

Issue #125, opened 2026-06-13
openai-compatible: max_tokens not forwarded on the streaming turn path


griffinwork40:
## Summary
The OpenAI-compatible provider does not send `max_tokens` / `max_completion_tokens` on its main **streaming** request path, so normal turns rely on the upstream server's default output cap. The Anthropic provider always forwards a max-output value clamped to the model ceiling.

## Current behavior
- Streaming `requestBody` omits any output-token cap:
  - Chat Completions path: `src/agent/providers/openai-compatible/query.ts` (~L576–585) — body has `model`, `messages`, `stream`, `stream_options`, `tools` only.
  - Responses API path: same file (~L537–550) — body has `model`, `input`, `stream`, `instructions`, `tools`, `store` only.
- The only place an output cap is set is the **non-streaming** `complete()` one-shot path: `src/agent/providers/openai-compatible/oneshot.ts:91–96` (which correctly switches `max_tokens` ↔ `max_completion_tokens` for o-series).
- `query.ts:864` references `maxTokens: contextLimit`, but that feeds `getContextUsage()` reporting — it is not part of the request body.

## Anthropic parity reference
`anthropic-direct` always forwards `max_tokens`, clamped to the model ceiling: `src/agent/providers/anthropic-direct/index.ts:951–969`, `loop.ts:258`.

## Impact
- Output length on streaming turns is uncapped from our side — behavior depends entirely on the upstream/server default, which varies across OpenAI, local MLX/llama.cpp/vLLM/ollama shims.
- o-series reasoning models in particular benefit from an explicit high `max_completion_tokens` so reasoning output isn't truncated by a low default.
- A caller has no way to bound output length for OpenAI-routed models the way they can for Anthropic.

## Suggested direction
Thread the resolved max-output value into the streaming `requestBody`, reusing the chat-vs-o-series field-selection logic already present in `oneshot.ts:91–96` (`max_tokens` for chat/local shims, `max_completion_tokens` for o-series). Keep it omittable for shims that reject the field, if needed.

---
_Found during an `anthropic-direct` vs `openai-compatible` capability-parity audit. Sibling gaps: retry layer, image/multimodal input, `reasoning_effort` forwarding._

griffinwork40:
Fixed by #195 (merged): the streaming request body now threads the resolved max-output value on both the Chat Completions and Responses API paths, reusing the o-series field-selection logic. Verified the fix is on `main` (commit 8f6b0e3). Closing.

Issue #103, opened 2026-06-12
Refactor: extract pure param resolvers into anthropic-direct/resolve-params.ts


griffinwork40:
## Context

`src/agent/providers/anthropic-direct/index.ts` (~1,105 LOC) contains a cluster of pure, stateless parameter-resolution helpers with **zero** dependency on the `AnthropicDirectProvider` class instance. Extracting them shrinks the provider file and isolates testable pure logic.

## Proposed change

Extract these module-scope functions into `src/agent/providers/anthropic-direct/resolve-params.ts`:

| Function | Line | Exported today? |
|---|---|---|
| `resolveAutoCompactThreshold` | 909 | no |
| `resolveMaxTokens` | 951 | **yes** |
| `resumeHistoryToMessages` | 971 | no |
| `resolveThinkingParam` | 1000 | **yes** |
| `resolveEffort` | 1080 | **yes** |

## Verified facts (independent re-derivation via /shadow-verify)

- All five are standalone module-scope `function` declarations — **none reference `this`** or any class field.
- `resolveMaxTokens`, `resolveThinkingParam`, `resolveEffort` already carry `export`.
- ⚠️ **Caveat:** the module-scope `const warnedTokenClamps = new Set<string>()` (line **925**) is shared by `resolveMaxTokens` (lines 958–959) and `resolveThinkingParam` (lines 1036–1037). It **must** move to the new module to preserve per-process dedup-warning behavior.

## Acceptance criteria

- [ ] The 5 functions + `warnedTokenClamps` live in `resolve-params.ts`
- [ ] `index.ts` re-imports them; `resolveMaxTokens`/`resolveThinkingParam`/`resolveEffort` stay exported (re-export from `index.ts` to keep import paths stable, OR update callers)
- [ ] Update import paths in dependent tests (`anthropic-direct.test.ts` + any subtest importing these by name)
- [ ] `pnpm lint` passes
- [ ] `pnpm test -- src/agent/providers/anthropic-direct.test.ts` passes
- [ ] `pnpm audit:sdk:check` still passes (no new SDK symbols)

## Risk

**Medium** — several test files import these symbols by name from `index.ts`; their import paths must be updated, or re-export from `index.ts` to keep them stable.

---
_Identified via `/gather` structural analysis, independently confirmed via `/shadow-verify` (verdict: CONFIRMED, both claims)._

PR #959, opened 2026-08-09
fix(tokens): harden max_tokens resolution across providers (#951, #953, #954)

### LINKED ISSUE #951: max-output-tokens <= 1024 emits budget_tokens >= max_tokens on enabled-thinking models, hard-400ing every turn
## Summary

`--max-output-tokens` / `AFK_MAX_OUTPUT_TOKENS` accepts any positive integer, but any value **≤ 1024** makes `resolveThinkingParam` emit `budget_tokens >= max_tokens`, which the Messages API rejects. Every turn 400s until the flag is changed.

Not reachable in a default config (the cap is unset by default → model ceiling, 64k–128k), and not reachable on the default models (see "Model dependence" below). It is a footgun, not a live outage.

## Mechanism

`src/agent/providers/anthropic-direct/resolve-params.ts:143-145`:

```ts
const reserve = Math.floor(maxTokens * THINKING_OUTPUT_RESERVE_FRACTION); // 0.25
const maxBudget = Math.max(1024, maxTokens - 1 - reserve);
const budget = Math.min(Math.max(explicit ?? maxBudget, 1024), maxBudget);
```

The `Math.max(1024, ...)` floor exists to honor the API's `budget_tokens >= 1024` minimum. But the API *also* requires `budget_tokens < max_tokens`. Once `maxTokens - 1 - reserve` drops below 1024, the floor pins the budget at exactly 1024 regardless of how small `maxTokens` gets, and the second constraint is violated.

| `max_tokens` | reserve | `max - 1 - reserve` | emitted `budget_tokens` | `budget < max`? |
|---|---|---|---|---|
| 8000 | 2000 | 5999 | 5999 | yes |
| 2048 | 512 | 1535 | 1535 | yes |
| 1365 | 341 | 1023 | 1024 (floor) | yes — but only 341 left for the reply |
| **1024** | 256 | 767 | **1024** | **no — 1024 == 1024 → 400** |
| **512** | 128 | 383 | **1024** | **no → 400** |
| **100** | 25 | 74 | **1024** | **no → 400** |

Exact cliff: `max_tokens <= 1024`, strict (1024 itself fails the `<` boundary).

There is also a soft starvation band just above it — from ~1366 down to 1025 the floor is engaged and the visible reply gets `max_tokens - 1024` tokens, shrinking to nothing.

## Model dependence

`requiresAdaptiveThinking` (`resolve-params.ts:28`) routes `/(claude-)?(opus|sonnet)-5/` and opus-4.7+ to adaptive thinking, which sends no `budget_tokens` at all and never reaches the arithmetic above. The default slot bindings are `claude-opus-5` / `claude-sonnet-5` (`src/agent/session/model-slots.ts:95-115`), so the defaults escape the cliff.

Affected models are the ones still taking the `{type:'enabled'}` branch:

- `claude-haiku-4-5-20251001` (the **small slot default** — so `--model haiku` hits this)
- `claude-fable-5`
- raw `claude-sonnet-4-6`

Thinking is on by default (`--thinking 'enabled:max'`, `src/cli/commands/interactive.ts:206` and `src/cli/commands/chat.ts:137`), so no opt-in beyond the one flag is needed to reach this.

## Repro

```
afk chat --model haiku --max-output-tokens 512 "hello"
```

Expected: an error naming the invalid flag value. Actual: an API 400 from the provider.

## No guard exists

- `resolveMaxTokens` (`resolve-params.ts:67-85`) clamps the **ceiling** only; there is no floor.
- `parseMaxOutputTokens` (`src/cli/shared-helpers.ts:212-226`) rejects zero, negatives, and non-integers — but accepts `1`.
- No test covers small values. The nearest coverage sweeps `[2_000, 10_000, 64_000, 128_000]` (`resolve-effort.test.ts:180`).

## Proposed fix

Primary — **validate at parse time**, turning a remote 400 into a local, actionable error:

- Add a floor to `parseMaxOutputTokens` and reject values that cannot satisfy the thinking contract, with a message naming both the floor and `--thinking disabled` as the escape hatch.

Defense in depth — **make the resolver total**:

- In `resolveThinkingParam`, when `maxTokens - 1 < 1024` (i.e. no legal budget exists), return `{ type: 'disabled' }` rather than emitting a knowingly-invalid budget. A tiny cap and extended thinking are mutually exclusive by API contract; picking the cap the operator explicitly set is the less surprising resolution.

Either alone closes the 400. Both together also fix the silent starvation band.

## Related

Catalogued as **F5** in `docs/audits/failure-geometry-audit.md` (line refs in that doc are stale — they point at `anthropic-direct/index.ts:921-985`, which no longer exists; the code moved to `resolve-params.ts`).

### LINKED ISSUE #953: openai-compatible skips the over-ceiling maxOutputTokens clamp that anthropic-direct performs
## Summary

The same `AgentConfig.maxOutputTokens` value is treated differently by the two providers: `anthropic-direct` clamps an over-ceiling value down to the model's limit with a warning, while `openai-compatible` forwards it verbatim and lets the provider reject it. Same config key, same user intent, two different outcomes.

## Evidence

**anthropic-direct clamps** — `src/agent/providers/anthropic-direct/resolve-params.ts:67-85`:

```ts
export function resolveMaxTokens(config: AgentConfig, model: string): number {
  const ceiling = maxOutputTokensFor(model);
  const v = config.maxOutputTokens;
  if (typeof v === 'number' && Number.isFinite(v) && v > 0) {
    const requested = Math.floor(v);
    if (requested > ceiling) {
      // ... one-time console.warn ...
      return ceiling;
    }
    return requested;
  }
  return ceiling;
}
```

**openai-compatible does not** — `src/agent/providers/openai-compatible/query/model-params.ts:27-35`:

```ts
export function resolveEffectiveMaxOutputTokens(
  model: string,
  configMaxOutput: number | undefined,
): number {
  const ceiling = maxOutputTokensFor(model);
  return typeof configMaxOutput === 'number' && Number.isFinite(configMaxOutput) && configMaxOutput > 0
    ? Math.floor(configMaxOutput)
    : ceiling;
}
```

The ceiling is computed and then used only as the unset-fallback; the over-ceiling branch is missing. `maxOutputTokensFor` is already provider-agnostic (`src/agent/model-limits.ts:79-91`, and the table carries `gpt-5.5` / `gpt-5.6*` entries precisely so the openai path resolves correctly), so the ceiling is known and simply not enforced.

## Impact

Low but real, and it lands hardest on local models. Anything unlisted in `MODEL_MAX_OUTPUT_TOKENS` — including MLX / llama.cpp / vLLM / ollama shims — falls back to `DEFAULT_MAX_OUTPUT = 64_000` (`model-limits.ts:73`). A user setting a large cap against a local runner with a much smaller real limit gets a provider-side rejection rather than a local clamp plus warning.

Note the asymmetry is only in the *over-ceiling* direction. The unset case is consistent across both providers.

## Proposed fix

Mirror the anthropic-direct branch in `resolveEffectiveMaxOutputTokens`: clamp to `ceiling` when the requested value exceeds it, with the same dedup-keyed one-time warning. Roughly five lines, and the shared warning helper already exists on the anthropic side if it is worth extracting.

Worth adding a test asserting both providers produce the same number for the same `(model, configMaxOutput)` pair — that is the invariant actually being violated.

## Related

Adjacent to **F10** in `docs/audits/failure-geometry-audit.md` ("no pre-run validation of `max_tokens` vs ceiling"), though that entry describes the anthropic path before the clamp landed and is now stale for that provider.

### LINKED ISSUE #954: docs: AFK_MAX_OUTPUT_TOKENS description names the wrong fallback; failure-geometry-audit line refs are stale
## Summary

Two documentation defects around output-token limits, found while tracing how `max_tokens` resolves.

## 1. `AFK_MAX_OUTPUT_TOKENS` description states the wrong fallback

`src/config/env.ts:258` (which generates `docs/env-registry.md` and the website tables):

> Cap on output tokens per turn. **Falls back to provider default when unset.**

It does not fall back to a provider default. It falls back to AFK's own per-model ceiling from `MODEL_MAX_OUTPUT_TOKENS` (`src/agent/model-limits.ts:28-91`) via `maxOutputTokensFor`, which is 128k for opus/sonnet/fable/gpt-5.x, 64k for haiku/claude-sonnet-4-6, and a hard-coded `DEFAULT_MAX_OUTPUT = 64_000` for anything unlisted.

The irony: the **deprecated** `AFK_MAX_TOKENS` entry directly below it (`env.ts:265-271`) already has the correct wording — "falls back to the model output ceiling when unset."

Fix: copy the accurate phrasing onto the live var, then `pnpm scan:env` to regenerate `docs/env-registry.{json,md}`. The website tables (`website/content/docs/configuration/environment-variables.mdx:31`, `guides/unattended-runs.mdx:99`) repeat the same wrong claim and want the same edit.

Worth stating explicitly in the description that the fallback is a **64k–128k model ceiling**, since "provider default" reads as "something small and safe" when the real behavior is the opposite — every request advertises the model's full output ceiling by default.

## 2. `docs/audits/failure-geometry-audit.md` line references are stale

The audit's findings table cites `anthropic-direct/index.ts` line numbers throughout (F1 at `index.ts:921-927`, F2 at `loop.ts:242`, F5 at `index.ts:978-985`, F10 at `index.ts:921-927`, F12 at `loop.ts:457-468`). That file no longer holds any of this code — it now lives in:

- `providers/anthropic-direct/resolve-params.ts` (`resolveMaxTokens`, `resolveThinkingParam`)
- `providers/anthropic-direct/loop/round-request.ts:138` (the `max_tokens` wire write)
- `providers/anthropic-direct/loop/turn-terminal.ts:84` (the orphaned-`tool_use` strip)

At least one finding is also **substantively** stale: **F21** claims the openai-compatible streaming path sends no output cap. That was true when the audit was written (`7d02d4bf`, 2026-06-09) and was fixed eight days later by `8f6b0e32` (2026-06-17, #125), which added `resolveStreamingMaxTokens`. Current code always attaches `max_tokens` / `max_completion_tokens` on Chat Completions streaming.

The audit is still a genuinely useful map — this is a refresh request, not a deletion. Given the doc's value as a findings index, it may be worth either re-anchoring the citations to symbol names rather than line numbers, or adding a dated "verified against" header so future readers know how much to trust the line refs.

griffinwork40:
## Summary

Hardens output-token (`max_tokens`) **resolution** across both providers. Three coupled defects, all confirmed live in `main` (`b1bb9c52`) by reading the cited code:

### #951 — small-cap thinking 400s every turn
Enabled (non-adaptive) thinking with `max_tokens <= 1024` emitted `budget_tokens == max_tokens`. The Messages API requires `1024 <= budget_tokens < max_tokens`, so that interval is **empty** at `max_tokens <= 1024` — there is nothing to clamp *to*. `resolveThinkingParam` now throws a legible error naming both escape hatches instead of taking an opaque per-turn HTTP 400.

- Not reachable on defaults: `opus-5`/`sonnet-5` route to adaptive (no `budget_tokens`), and the default cap is the 64k–128k model ceiling. Only an explicit tiny `--max-output-tokens` / `AFK_MAX_OUTPUT_TOKENS` on a non-adaptive model (`haiku`, `fable-5`, raw `sonnet-4-6`) reaches the branch.
- Repro before: `afk chat --model haiku --max-output-tokens 512 "hi"` → provider 400. After: clear AFK error.

### #953 — provider asymmetry on the over-ceiling clamp
`openai-compatible`'s `resolveEffectiveMaxOutputTokens` forwarded an over-ceiling cap verbatim (provider-side 400), while `anthropic-direct`'s `resolveMaxTokens` clamped down with a one-time warning. Mirrored the clamp + dedup-keyed warning so **both providers resolve the same number for the same `(model, cap)`**. Hits hardest on local runners (MLX/llama.cpp/vLLM) whose real output limit is below the 64k `DEFAULT_MAX_OUTPUT` fallback.

### #954 — docs describe the wrong fallback
`AFK_MAX_OUTPUT_TOKENS` said *"falls back to provider default"* — it falls back to AFK's own 64k–128k **model ceiling**. Corrected the `env.ts` registry entry (regenerated `docs/env-registry.{json,md}`) and both website tables. Also refreshed `docs/audits/failure-geometry-audit.md` with a dated staleness banner (its `index.ts`/`loop.ts` line refs predate the #103 provider split — now mapped to current symbol homes) and corrected the **substantively stale F21** (the openai streaming path attaches an output cap since #125).

Scope is deliberately **resolution only**. The sibling issue #952 (truncation is *invisible* on live surfaces) is a distinct behavioral change and ships separately.

## How was this tested?

- `pnpm scan:env:check` ✓ (docs in sync) · `pnpm audit:env:check` ✓
- New unit tests:
  - `resolveThinkingParam`: throws at `max_tokens ∈ {1024, 512, 100, 1}`; does **not** throw at 1025 (smallest satisfiable, budget=1024 < 1025); adaptive-promoted models (`opus-5`/`sonnet-5`) skip the guard; adaptive/disabled never throw.
  - `resolveEffectiveMaxOutputTokens`: over-ceiling clamp; local-runner (64k default) clamp; cross-provider parity vs `resolveMaxTokens` across `{unset, fits, ceiling, over, +Infinity}`.
- Full suites for both providers green: `anthropic-direct/` 660 ✓, `openai-compatible/query/` + `cli-thinking-options` 50 ✓.

Closes #951
Closes #953
Closes #954

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes (touched files + both provider suites)


griffinwork40:
## Automated review — `/review` (agent-afk)

**Decision: DO NOT MERGE — 1 medium blocking (correctness); 2 low, 1 nit.**

Reviewed ref `98627f02`. Regime: light (hotfix, +145/−11, 10 files). CI green at review time. Citations were re-read against the ref, and the one waived finding was independently re-derived by a shadow verifier — that re-derivation **revoked the waiver**, which is what moves this to DO NOT MERGE.

To be explicit up front: **this PR does not regress any working configuration.** The blocking finding is that the safety argument stated in the code comment and PR body is inaccurate, and the resulting failure is unhandled. Both are cheap to fix.

---

### Findings

**1 · medium · blocking:true · confidence:high · correctness+spec-compliance**
`src/agent/providers/anthropic-direct/resolve-params.ts:148-157` · ref:98627f02 · citation-type:file-state (verified at ref)

**Finding:** The new guard's stated reachability is wrong: the comment says the branch is `Unreachable on defaults` and needs `only an explicit tiny --max-output-tokens`, but it omits that thinking is **enabled by default** on both interactive surfaces — so the trigger is a *single* stale setting, not two deliberate opt-ins, and the resulting `throw` leaves `resolveThinkingParam` eagerly (not lazily inside a generator) and propagates out of session construction with no handler on the path.

**Evidence** — the default that the comment omits:
```ts
.option('--thinking <mode>', "Thinking mode: 'adaptive' | 'disabled' | 'enabled:<N>'", 'enabled:max')
```
`src/cli/commands/chat.ts:137` and `src/cli/commands/interactive.ts:206` (identical default; `daemon.ts:263` correctly has none).

Because `config.thinking` is therefore defined by default, `provider-query-build.ts:127` evaluates eagerly inside the `new AnthropicDirectQuery({...})` argument list:
```ts
? { thinking: resolveThinkingParam(config.thinking, maxTokens, model) }
```

**Net effect:** a user with `AFK_MAX_OUTPUT_TOKENS <= 1024` in `~/.afk/config/afk.env` on a non-adaptive model (`haiku`, `fable-5`, raw `sonnet-4-6` — `requiresAdaptiveThinking` at `resolve-params.ts:28-29` excludes them) hits this on **every launch**, having typed no flag at that moment. Pre-PR that same config took an HTTP 400 per turn, so this is strictly an improvement in *legibility* — but it is not the narrow, deliberately-opted-into case the comment describes.

**Suggestion:** two one-line fixes, either or both — (a) correct the comment and PR body to say thinking is default-`enabled:max`, so the trigger is a small output cap alone on a non-adaptive model; and (b) catch this at the session-construction call site so the (genuinely good) error text surfaces as a clean CLI error rather than an uncaught constructor exception.

---

**2 · low · blocking:false · confidence:high · test-coverage**
`src/agent/model-limits.test.ts:115-118` · ref:98627f02 · citation-type:diff-context

**Finding:** The over-ceiling clamp test asserts the clamped *value* but never asserts the clamp *warning* fires, so a regression that silently drops the new dedup-keyed `console.warn` passes green.
```ts
it('clamps an over-ceiling cap down to the model ceiling', () => {
  const model = 'gpt-5.6';
  const ceiling = maxOutputTokensFor(model);
  expect(resolveEffectiveMaxOutputTokens(model, ceiling + 500_000)).toBe(ceiling);
```
Absence claim verified at the ref: `console.warn|spyOn` in `src/agent/model-limits.test.ts` → **0 matches**.

**Suggestion:** add a `vi.spyOn(console, 'warn')` assertion that the warning fires once, and not again on a second call with the same `(model, requested)` key.

---

**3 · low · blocking:false · confidence:medium · test-coverage**
`src/agent/model-limits.test.ts:128-136` · ref:98627f02 · citation-type:diff-context

**Finding:** The cross-provider parity loop — the test carrying the PR's central "both providers resolve the same number" claim — exercises only `gpt-5.6`, never a non-adaptive Anthropic model, which is precisely the family the PR identifies as the live failure case.
```ts
for (const cap of [undefined, 8_000, ceiling, ceiling + 500_000, Number.POSITIVE_INFINITY]) {
  expect(resolveEffectiveMaxOutputTokens(model, cap), `cap=${cap}`).toBe(
    resolveMaxTokens(cfg(cap), model),
  );
```
**Suggestion:** add one parity assertion over a `claude-haiku-*` model id so the non-adaptive path is pinned by the parity test too.

---

**4 · nit · blocking:false · confidence:high · maintainability**
`src/agent/providers/openai-compatible/query/model-params.ts:15-19` · ref:98627f02 · citation-type:diff-context

**Finding:** The comment says it `Mirrors warnedTokenClamps in the anthropic-direct resolve-params.ts`, but the two `Set`s are independent module-scope objects (`resolve-params.ts:40`), so the same misconfiguration warns once per provider, not once per process — "mirrors" reads as "shares".

**Suggestion:** reword to "parallel, independent dedupe Set — same key format, separate scope".

---

### Verification manifest (what the review corrected about itself)

| Claim | Status |
|---|---|
| `provider-query-build.ts:127`, `provider-runtime.ts:327` | **verified** verbatim at ref |
| `agent-session.ts:307` | **path corrected** — no such file; the quoted line is verbatim at `src/agent/session/agent-session.ts:307` |
| Finding 2/3 line ranges | **corrected** — originally cited `resolve-effort.test.ts:186-220` and `model-limits.test.ts:77-80`; the real sites are `model-limits.test.ts:115-118` and `:128-136` |
| "no `console.warn` assertion" | **confirmed-absent** (0 matches) — finding stands |
| Waiver on finding 1 ("narrow reachability") | **REVOKED** by independent re-derivation — the `--thinking enabled:max` default refutes the "explicit thinking required" premise, so the override is inadmissible and the default `blocking:true` stands |

### What was not checked
Citations verified inline against branch HEAD `98627f02`. Stated intent: PR title+body — spec-compliance assessed against it. No test suite, lint, or build was run (CI is green and reported separately). Whether a top-level CLI handler ultimately prints the throw as a clean message rather than a stack trace was **not conclusively resolved** — `chat.ts` has `catch` blocks at 300/328/642/749 that were not each traced to the construction call site; that uncertainty is why finding 1's suggestion (b) is phrased as a check, not an assertion. Website MDX rendering and byte-exactness of the regenerated `docs/env-registry.*` were not verified (CI gates cover the latter).

<sub>Generated by `/review` on agent-afk. Read-only: no files, branches, or PR state were modified.</sub>

griffinwork40:
## `/fix-pr` — feedback addressed (pushed `dd9712e8`)

All five items from my review and the Codex P2 are addressed. One of them turned out to be a **regression this PR introduced**, which is the substantive change here.

| # | Item | Source | Status |
|---|---|---|---|
| 1 | Inaccurate reachability claim in the `max_tokens <= 1024` guard comment | review (blocking) | **fixed** — comment only |
| 2 | Clamping an explicit cap down to a *guessed* ceiling | Codex P2 (verified) | **fixed** — `model-limits.ts`, `model-params.ts` |
| 3 | Clamp test never asserted the warning fires | review (low) | **fixed** |
| 4 | Parity test missed the non-adaptive Anthropic family | review (low) | **fixed** |
| 5 | Misleading "Mirrors" comment on the dedupe Set | review (nit) | **fixed** |

### Item 1 — the reachability claim was understated

The guard's comment said the branch is `Unreachable on defaults … only an explicit tiny --max-output-tokens on a non-adaptive model reaches here`. That omits that `--thinking` **defaults to `enabled:max`**:

- `src/cli/commands/chat.ts:137` → `.option('--thinking <mode>', "…", 'enabled:max')`
- `src/cli/commands/interactive.ts:206` → same default
- (`daemon.ts:263` correctly has no default)

Thinking is therefore on by default on `afk chat` / `afk interactive`, so the real trigger is an output cap ≤ 1024 **alone** on a non-adaptive model — including a stale `AFK_MAX_OUTPUT_TOKENS=512` in `~/.afk/config/afk.env`, which fires on every launch with no flag typed at that moment. Comment corrected to say exactly that. Throw logic and error text are unchanged.

### Item 2 — Codex's P2, and why it was worth acting on

Codex flagged that `maxOutputTokensFor()` misses the bare entry for provider-prefixed ids. Verified independently, with one correction to its framing: the repo *does* normalize prefixes for **capability detection** (`bareModelId`, `model-capabilities.ts:135-138`) but **not** for limit lookup, so the bot's premise held.

The reachability is real: `providers/index.ts:221` routes **any** id containing `/` to this provider — `if (lowered.includes('/')) return 'openai-compatible';` — which covers OpenRouter and every local-runner id (`mlx-community/…`, `Qwen/…`), i.e. exactly the population this PR says it "hits hardest."

And it *was* a regression introduced here. Before: an explicit cap was forwarded verbatim, so `--max-output-tokens 128000` on `mlx-community/<model>` reached the wire. After: `maxOutputTokensFor` returned the 64k `DEFAULT_MAX_OUTPUT` **guess** for that unknown id and the new clamp silently cut the request to 64000.

**Fix — only clamp against a ceiling we actually know.** Added `maxOutputTokensForKnown(model): number | undefined` in `model-limits.ts`, which returns `undefined` for unlisted ids; `maxOutputTokensFor` now delegates to it with the `?? DEFAULT_MAX_OUTPUT` fallback, so **every existing caller is behavior-preserved**. `resolveEffectiveMaxOutputTokens` clamps only when the ceiling is known and forwards the user's explicit cap untouched otherwise.

The reasoning: clamping to a guessed limit does not prevent a provider 400 — it invents a limit the provider might not enforce, and silently degrades an explicit request. Clamping against a *documented* ceiling still does its job, so the cross-provider parity invariant is intact for known models (`gpt-5.6` still clamps identically on both sides — that parity test is unchanged and passing).

One collateral test change: the pre-existing `'clamps hardest on unlisted local runners'` case asserted the exact behavior this removes. It is now split into two — the unset-cap fallback (unchanged) and the new no-clamp-on-explicit-cap-for-unknown-model behavior.

**Deliberately not done:** adding `bareModelId` normalization inside `maxOutputTokensFor`. That would make `openrouter/gpt-5.5` resolve to the real `gpt-5.5` 128k entry, but it changes limit resolution repo-wide (both providers, context limits, auto-compact) — worth its own PR. Note `resolve-params.ts:67-68` on the anthropic side shares the same lookup gap, so this is a shared limitation rather than an asymmetry introduced here.

### Verification (run locally in an isolated worktree, not just reported by the fixer)

- `pnpm test` full suite — **813 files / 15506 tests passed**, 2 pre-existing skips
- `pnpm build`, `audit:env:check`, `scan:env:check`, `audit:chalk:check`, `audit:sdk:check`, `fix:pins:check` — all pass

### Still open from my review
The blocking finding also noted the throw leaves `resolveThinkingParam` eagerly and propagates out of session construction, so the (genuinely good) error text can surface as an uncaught exception rather than a clean CLI error. **Not fixed here** — wrapping the construction call site is a behavioral change with wider blast radius than this PR's scope. The comment is now accurate about when it fires; converting it to a handled CLI error is the follow-up.

<sub>Generated by `/fix-pr`. Pushed as a plain fast-forward to the PR branch; no force-push, no other ref touched.</sub>

griffinwork40:
@codex review

griffinwork40:
## Review — PR #959 · `fix(tokens): harden max_tokens resolution across providers`

**Target:** PR #959 · `fix/max-tokens-resolution-951-953-954` → `main` · reviewed ref `54ed70de` · hotfix · 11 files · +218/−21 (239 lines) → **light regime**
**Stated intent:** PR #959 title + body (#951 fail-fast thinking guard, #953 cross-provider clamp parity, #954 docs correction) — spec-compliance **assessed**.

### ⚠ Process failure that shaped this review

The working tree is on **`main @ e5f2b65d`**, not the PR branch. The session preamble's `fix/max-tokens-resolution-951-953-954 @ 54ed70de (1 dirty)` is stale metadata. Every sub-agent `Read` — Wave 1 *and* all three shadow verifiers — resolved against **main**, so two verifiers returned confident `REFUTED` verdicts asserting the changed code "does not exist." It does; I confirmed it via `git show 54ed70de:`. **All sub-agent evidence in this run is discarded.** Only findings I re-derived inline at the ref survive below. Wave 1 also hit its tool-use iteration cap and returned a partial result.

I also owe a correction to my own mid-run inference: I read `query()` as a per-turn method. At the ref it is invoked from `initSdkLifecycle()` (`agent-session.ts:274` ← `:248`) at **construction**, so the new `throw` fires at session startup, not per turn.

---

### security
`no issues found` — read `resolve-params.ts`, `model-params.ts`, `model-limits.ts` at the ref. No new auth, secret, or untrusted-input surface; the added `console.warn` interpolates only a model id and an integer.

### api-compat
`no issues found` — `maxOutputTokensFor` keeps its signature and its lookup order (`_1m` early return → `resolveModelInput` → `id` → `id.toLowerCase()` → 64k) across the split into `maxOutputTokensForKnown() ?? DEFAULT_MAX_OUTPUT` (`model-limits.ts:85-100`). The new export has exactly one production importer (`model-params.ts:11,57`); `git grep` at the ref shows no other non-test consumer.

### correctness
**`low · blocking:false · confidence:high · correctness · src/agent/providers/anthropic-direct/resolve-params.ts:153-160 + src/agent/session/agent-session.ts:307 · ref:54ed70de · citation-type:diff-context**
The `#951` guard throws a bare `Error` with no distinguished class or classifier branch, so the failure renders through the generic unknown-error path on every surface rather than a typed, hinted one.
```
if (maxTokens <= 1024) {
  throw new Error(
    `[afk] Extended thinking requires max_tokens > 1024 (the API constraint is ` +
```
`git grep "Extended thinking requires" 54ed70de` returns only this line and its test — no named class, no presenter branch (**absence confirmed at the ref**). The message text itself names both escape hatches, so the intent's "legible error" is satisfied; this is rendering polish, not a behavioral defect. *Suggestion:* add a `ThinkingBudgetUnsatisfiableError extends Error` beside the existing `utils/errors.ts` classes so presenters can attach a `hint`.

`no issues found` — the guard sits **after** the adaptive promotion (`return { type: 'adaptive', … }` precedes it), so `opus-5`/`sonnet-5` never reach it; pinned by the new test at `resolve-effort.test.ts`. Reachability in the intent is accurate at the ref: `chat.ts:137` and `interactive.ts:206` default `--thinking` to `enabled:max`; `daemon.ts:263` has **no** default.

### spec-compliance
**`low · blocking:false · confidence:high · spec-compliance · src/agent/providers/openai-compatible/query/model-params.ts:15-19 · ref:54ed70de · citation-type:diff-context**
The new doc-comment contradicts itself in adjacent sentences, so the block states two incompatible dedupe scopes for the same `Set`.
```
* misconfiguration warns once per process rather than once per turn. Parallel
* `resolve-params.ts`: same key format, separate module scope, so the same
* misconfiguration warns once per PROVIDER, not once per process.
```
Only the second is accurate — there are two independent module-scope Sets (`resolve-params.ts:40`, `model-params.ts:21`). *Suggestion:* delete "once per process rather than once per turn" from line 16 and keep the once-per-provider framing.

**`low · blocking:false · confidence:high · spec-compliance · src/agent/providers/anthropic-direct/resolve-params.ts:68 + src/agent/providers/openai-compatible/query/model-params.ts:57 · ref:54ed70de · citation-type:diff-context**
The PR body asserts parity flatly ("both providers resolve the same number for the same `(model, cap)`") while the implementation scopes it to known ceilings, so the body overstates what shipped. `resolveMaxTokens` uses `maxOutputTokensFor` (64k fallback); `resolveEffectiveMaxOutputTokens` uses `maxOutputTokensForKnown` (undefined → no clamp). For an unlisted id with a 128k cap the two return **64,000 vs 128,000**. Divergence needs the same id on both providers, reachable only via the Tier-1 `--provider`/`AFK_PROVIDER` override or a per-slot `binding.provider` (`providers/index.ts:152-157,172-174`) — narrow, and the anthropic-side clamp is pre-existing, not a regression. *Suggestion:* copy the code comment's "ONLY when the model's ceiling is actually known" qualifier into the PR body's #953 summary line.

### test-coverage
**`low · blocking:false · confidence:high · test-coverage · src/agent/model-limits.test.ts:155-176 · ref:54ed70de · citation-type:diff-context**
Both cross-provider parity tests use known-ceiling models (`gpt-5.6`, `claude-haiku-4-5-20251001`), so the deliberate *divergence* for unlisted ids is asserted single-provider only and nothing pins it as intended. Grepped every `resolveEffectiveMaxOutputTokens` call site in the test file at the ref: no case pairs an unlisted id against `resolveMaxTokens` (**absence confirmed**). *Suggestion:* add one assertion that `resolveEffectiveMaxOutputTokens('mlx-community/x', 128_000) !== resolveMaxTokens(cfg(128_000), 'mlx-community/x')`, documenting the asymmetry as deliberate.

### perf-observability
`no issues found` — the added path is O(1) per call; the dedupe `Set` grows only with distinct `(model, requested)` pairs actually seen, and mirrors the pattern already shipped at `resolve-params.ts:40`.

---

### Decision: MERGE — 0 blocking (4 low).

All four findings are `low` at their default disposition; no overrides were issued, so no waiver needed a second reader.

### What was not checked
- **Citations verified inline against `54ed70de` via `git show`/`git grep`.** All sub-agent citations were discarded as ref-invalid (they resolved against `main @ e5f2b65d`).
- Shadow-verify ran one round on 3 claims; **all three returned contaminated verdicts** from the stale worktree and were discarded, not merged. No repeat round (bound: 1 round, ≤3 claims).
- Did **not** run `pnpm lint` / `build` / `test` / `scan:env:check` — verifying the PR's test claims requires checking out the branch, which the read-only constraint and your dirty `.afk/` forbid. The PR's green-suite claims are unverified here.
- Did **not** trace whether the construction-time throw can re-fire on the mid-session lifecycle-reset path (`agent-session.ts:913`) during a `/model` switch, or how Telegram surfaces it.
- Did **not** review the `docs/audits/failure-geometry-audit.md` git-history claim that `8f6b0e32` (#125) added `resolveStreamingMaxTokens`.

**Recommendation beyond the PR:** re-run this review with the branch actually checked out (`git worktree add` off `54ed70de`) if you want sub-agent-grade coverage — file-reading agents cannot review a ref that is not on disk, and they fail *confidently* rather than loudly when it isn't.

**Done**
- Reviewed PR #959 read-only under the light regime; Wave 1 partial + all shadow-verify output discarded as ref-invalid, findings re-derived inline at `54ed70de`.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/2644ccc4-c1cc-499f-9c63-e697ca8391dc/pr-959.diff`; all citations re-checked via `git show 54ed70dedd1e508f2a090e0cb608f724f75cc728:<path>` in this session.
- World unchanged: no files written, edited, committed, pushed, or stashed; no PR mutation; working tree left on `main @ e5f2b65d` with `?? .afk/` untouched.
- Deferred: the PR's `pnpm lint`/`build`/`test` claims are unverified because confirming them requires checking out the branch, which read-only scope forbids.
- No `git commit` ran and no files were written — this review is transcript-plus-evidence-paths only, by design.

---

PR #960, opened 2026-08-09
fix(tokens): surface max_tokens truncation instead of failing silently (#952)

### LINKED ISSUE #952: max_tokens truncation is invisible on every live surface; a dropped mid-turn tool call reads as the agent stalling

griffinwork40:
## Summary

Makes `max_tokens` **truncation** visible. A turn cut at the output-token cap is already classified `ClosureReason: 'truncated'` — but that signal never leaves the trace stream, so on the REPL, Telegram, and one-shot `chat` a truncated turn is **indistinguishable from a clean completion**. The worst variant: a tool call cut mid-arguments is stripped (correctly, to avoid an orphan-`tool_use` 400) and then silently dropped — the agent announces an action and stops, which reads as a mid-task stall.

Sibling of #951/#953/#954 (PR #959), which fixed *resolution*; this fixes *observability*.

### Changes
1. **anthropic-direct `turn-terminal.ts`** — emit a **display-only** truncation notice (mirrors the existing `refusal` branch precedent — yielded as `assistant.message`, **not** pushed to history via `input.messages.push`) after the partial text so the answer still shows first. When a `tool_use` block is stripped under truncation, the notice **names the dropped tool** (`turnResult.toolUseBlocks`) so the never-dispatched action is legible.
2. **subagent `result.ts`** — a truncated child streams real text and reaches a terminal message, so `handle.ts` resolves it `succeeded`; but that text ends mid-thought. `isIncompleteStopReason` now includes truncation, so a truncated child carries the same `[⚠ PARTIAL RESULT]` banner as the tool-loop-cap / wind-down cases. **Provider-agnostic** — covers anthropic `'max_tokens'` and openai `'length'`.
3. **new `providers/shared/truncation.ts`** — single-sources the `'max_tokens' || 'length'` predicate. `session/closure-reason.ts` now re-exports it (its test is unaffected), so the literals no longer drift across the closure classifier, the terminal path, and the subagent path. `providers/shared/` is the correct home — `result.ts` already imports its sibling stop-reason constants (`TOOL_USE_LOOP_CAPPED`, `SOFT_DEADLINE_WIND_DOWN`) from there, and `session/` → `providers/` is a downward import.

### Deliberately scoped out (named follow-up, not dropped)
The **openai-compatible LIVE-turn notice** (`query.ts` terminal emit) is not added here. Its `finish_reason` handling and partial-`tool_call` accumulation differ enough from anthropic's clean `TurnResult.toolUseBlocks` choke point to warrant a separate change + test; a text-only half-mirror (notice without tool naming) would itself be a form of provider drift. The **cross-provider subagent banner in change (2) already covers truncated openai *children***, so the agent-of-agents path is not drift-exposed — only the openai top-level live surface is, and that is the follow-up.

## How was this tested?

- New tests:
  - `loop.orphan.test.ts`: a `max_tokens`-stripped `tool_use` emits a notice naming the dropped tool (`read_file`) + `NOT dispatched`, and `turn.completed` still fires; a text-only `max_tokens` turn emits an "allow a longer reply" notice **and** still delivers the model's partial text as its own `assistant.message`.
  - `result.test.ts`: flipped the stale `isIncompleteStopReason('max_tokens') === false` assertion (it encoded the bug) → now `true` for both `max_tokens` and `length`; `stop_sequence`/`pause_turn` remain false; `annotateIfIncomplete` prepends the banner naming the output-token limit for both providers.
- Regression net (green): `anthropic-direct/` + `subagent/` + `session/` = **1179 tests** pass, including the orphan-strip invariant (`loop.orphan.test.ts`) and closure-reason (18).

Closes #952

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes (touched files + anthropic-direct/subagent/session suites)

### DIFF
--- a/src/agent/providers/anthropic-direct/loop.orphan.test.ts
+++ b/src/agent/providers/anthropic-direct/loop.orphan.test.ts
@@ -186,6 +186,8 @@ describe('loop.ts runTurn — orphan tool_use prevention', () => {
   });
 
   // #952: text-only truncation (no tool_use) → notice without a tool name.
+  // #960: the partial text and the notice are now ONE assistant.message, not
+  // two — see the dedicated merged-message test below for the full rationale.
   it('surfaces a truncation notice for text-only max_tokens truncation', async () => {
     const client = makeClient(() => fromArray(makeTextStream('a partial answer', 'max_tokens')));
     const dispatcher = makeDispatcher(() => Promise.resolve({ content: 'never called' }));
@@ -209,12 +211,62 @@ describe('loop.ts runTurn — orphan tool_use prevention', () => {
     expect(noticed).toHaveLength(1);
     expect(noticed[0]!.text).toContain('allow a longer reply');
     expect(noticed[0]!.text).not.toContain('NOT dispatched');
-    // The model's partial text is still delivered as its own assistant.message.
-    expect(
-      events.some(
-        (e) => e.type === 'assistant.message' && (e as { text: string }).text === 'a partial answer',
-      ),
-    ).toBe(true);
+    // The model's partial text is delivered INSIDE the same assistant.message
+    // as the notice (#960), not as a separate event.
+    expect(noticed[0]!.text).toContain('a partial answer');
+    // Exactly one assistant.message for the whole turn — the partial text and
+    // the notice were never split into two events.
+    expect(events.filter((e) => e.type === 'assistant.message')).toHaveLength(1);
+  });
+
+  // #960: a last-wins consumer (the non-streaming `sendMessage()` path in
+  // agent-session.ts, and a subagent's final-message capture in handle.ts)
+  // keeps only the LAST `assistant.message` of a turn. Before this fix,
+  // emitNonToolUseTerminal yielded the partial text and the truncation notice
+  // as two separate assistant.message events, so those consumers silently
+  // discarded the model's real (partial) answer and kept only the warning.
+  // Pin the fix at the event level: one merged message, answer first.
+  it('merges the partial answer and the truncation notice into ONE assistant.message (#960)', async () => {
+    const client = makeClient(() => fromArray(makeTextStream('a partial answer', 'max_tokens')));
+    const dispatcher = makeDispatcher(() => Promise.resolve({ content: 'never called' }));
+    const events = await collect(
+      runTurn({
+        client,
+        messages: [{ role: 'user', content: 'explain' }],
+        system: null,
+        tools: null,
+        toolDispatcher: dispatcher,
+        model: 'claude-test',
+        maxTokens: 8_000,
+        headers: {},
+        signal: new AbortController().signal,
+        ctx,
+      }),
+    );
+
+    const assistantMessages = events.filter(
+      (e) => e.type === 'assistant.message',
+    ) as Array<{ text: string }>;
+    // A last-wins consumer keeps assistantMessages.at(-1) — that message alone
+    // must carry both the answer and the notice, or the answer is lost.
+    expect(assistantMessages).toHaveLength(1);
+    const merged = assistantMessages[0]!.text;
+    expect(merged).toContain('output-token limit');
+    // Answer first, notice second, joined by exactly one blank line — the
+    // model's raw text is preserved verbatim as the message prefix so a
+    // last-wins consumer reading this one event still recovers the full
+    // partial answer.
+    expect(merged.startsWith('a partial answer\n\n')).toBe(true);
+    expect(merged.indexOf('a partial answer')).toBeLessThan(merged.indexOf('output-token limit'));
+
+    // The `suggestion` event (short text) must carry ONLY the model's own
+    // text, never the appended operator notice.
+    const suggestion = events.find((e) => e.type === 'suggestion') as
+      | { suggestion: string }
+      | undefined;
+    expect(suggestion, 'a suggestion should be emitted for short partial text').toBeDefined();
+    expect(suggestion!.suggestion).toBe('a partial answer');
+    expect(suggestion!.suggestion).not.toContain('output-token limit');
   });
 
   it('rolls back the assistant tool_use push when the dispatcher accessor throws', async () => {
--- a/src/agent/providers/anthropic-direct/loop.orphan.test.ts
+++ b/src/agent/providers/anthropic-direct/loop.orphan.test.ts
@@ -39,6 +39,68 @@ import {
 //      `input.toolDispatcher` throws). Pre-fix: assistant tool_use push
 //      remained in history. Post-fix: rollback splices it back out.
 
+/**
+ * A `max_tokens` stream carrying BOTH partial text and one or more `tool_use`
+ * blocks that never got dispatched — the shape that exercises the notice's
+ * dropped-tool naming (and its de-duplication) on the partial-text path.
+ */
+function makeTextStreamWithDroppedTool(
+  text: string,
+  toolNames: string[],
+): RawMessageStreamEvent[] {
+  const events: RawMessageStreamEvent[] = [
+    {
+      type: 'message_start',
+      message: {
+        id: 'msg_truncate_mixed',
+        type: 'message',
+        role: 'assistant',
+        content: [],
+        model: 'claude-test',
+        stop_reason: null,
+        stop_sequence: null,
+        usage: baseUsage(),
+      },
+    } as unknown as RawMessageStreamEvent,
+    {
+      type: 'content_block_start',
+      index: 0,
+      content_block: { type: 'text', text: '' },
+    } as unknown as RawMessageStreamEvent,
+    {
+      type: 'content_block_delta',
+      index: 0,
+      delta: { type: 'text_delta', text },
+    } as unknown as RawMessageStreamEvent,
+    { type: 'content_block_stop', index: 0 } as unknown as RawMessageStreamEvent,
+  ];
+  toolNames.forEach((name, i) => {
+    const index = i + 1;
+    events.push(
+      {
+        type: 'content_block_start',
+        index,
+        content_block: { type: 'tool_use', id: `toolu_dropped_${index}`, name, input: {} },
+      } as unknown as RawMessageStreamEvent,
+      {
+        type: 'content_block_delta',
+        index,
+        delta: { type: 'input_json_delta', partial_json: '{"file":"a.ts"}' },
+      } as unknown as RawMessageStreamEvent,
+      { type: 'content_block_stop', index } as unknown as RawMessageStreamEvent,
+    );
+  });
+  events.push(
+    {
+      type: 'message_delta',
+      delta: { stop_reason: 'max_tokens', stop_sequence: null },
+      usage: { output_tokens: 4 },
+    } as unknown as RawMessageStreamEvent,
+    { type: 'message_stop' } as unknown as RawMessageStreamEvent,
+  );
+  return events;
+}
+
 describe('loop.ts runTurn — orphan tool_use prevention', () => {
   beforeEach(() => {
     vi.clearAllMocks();
@@ -121,10 +183,17 @@ describe('loop.ts runTurn — orphan tool_use prevention', () => {
     expect(flatTooluses).not.toContain('toolu_orphan');
   });
 
-  // #952: the strip above is silent — a dropped tool call reads as the agent
-  // stalling. runTurn must surface a display-only truncation notice NAMING the
-  // dropped tool so the operator knows the action never ran.
-  it('surfaces a truncation notice naming the dropped tool when max_tokens strips a tool_use', async () => {
+  // #960 REGRESSION GUARD — do not "fix" this test by making it expect a notice.
+  //
+  // A truncated turn that produced NO text must emit NO `assistant.message`.
+  // The absence is the signal: stream-consumer drops a textless turn, so
+  // `handle.ts` never sets `finalMessage`, reaches its ZERO-OUTPUT branch, and
+  // THROWS — which is what makes a zero-output subagent resolve `failed` (not a
+  // false `succeeded`) and what lets stream-cut-retry re-dispatch it. An earlier
+  // revision of this PR emitted a standalone notice here; that set
+  // `finalMessage`, silently unreachable-ing the throw, and a truncated child
+  // returned to its parent as a SUCCESS whose whole content was the warning.
+  it('emits NO assistant.message when max_tokens strips a tool_use and no text was produced', async () => {
     const truncatedToolUseStream: RawMessageStreamEvent[] = [
       {
         type: 'message_start',
@@ -175,14 +244,78 @@ describe('loop.ts runTurn — orphan tool_use prevention', () => {
       }),
     );
 
-    const notice = events.find(
-      (e) => e.type === 'assistant.message' && /output-token limit/.test((e as { text: string }).text),
-    ) as { text: string } | undefined;
-    expect(notice, 'a truncation notice event should be emitted').toBeDefined();
-    expect(notice!.text).toContain('read_file'); // names the dropped tool
-    expect(notice!.text).toContain('NOT dispatched');
+    // The textless turn stays textless: no assistant.message at all, so a
+    // downstream `if (event.text)` consumer sees nothing and the zero-output
+    // path stays reachable.
+    expect(events.filter((e) => e.type === 'assistant.message')).toHaveLength(0);
+    // ...and specifically no notice smuggled in on that channel.
+    expect(
+      events.some(
+        (e) =>
+          e.type === 'assistant.message' &&
+          /output-token limit/.test((e as { text: string }).text),
+      ),
+      'a textless truncated turn must not emit a standalone notice as assistant.message',
+    ).toBe(false);
     // A turn.completed must still follow — the turn ends cleanly, not as an error.
+    // (The orphan-strip invariant itself is pinned by the first test above.)
     expect(events.some((e) => e.type === 'turn.completed')).toBe(true);
+  });
+
+  // #952: when the turn DID produce text, the notice rides on that message and
+  // names the dropped tool — this is the path that stayed after #960 narrowed
+  // the textless case above.
+  it('names the dropped tool in the notice when partial text exists', async () => {
+    const client = makeClient(() =>
+      fromArray(makeTextStreamWithDroppedTool('partial answer', ['read_file'])),
+    );
+    const dispatcher = makeDispatcher(() => Promise.resolve({ content: 'never called' }));
+    const events = await collect(
+      runTurn({
+        client,
+        messages: [{ role: 'user', content: 'do a thing' }],
+        system: null,
+        tools: null,
+        toolDispatcher: dispatcher,
+        model: 'claude-test',
+        maxTokens: 8_000,
+        headers: {},
+        signal: new AbortController().signal,
+        ctx,
+      }),
+    );
+    const msgs = events.filter((e) => e.type === 'assistant.message') as Array<{ text: string }>;
+    expect(msgs).toHaveLength(1);
+    expect(msgs[0]!.text).toContain('partial answer');
+    expect(msgs[0]!.text).toContain('read_file');
+    expect(msgs[0]!.text).toContain('NOT dispatched');
+  });
+
+  // The notice de-dupes tool names (`[...new Set(names)]`) so a model that
+  // truncated two calls to the SAME tool does not read as two distinct
+  // dropped actions.
+  it('names a repeated dropped tool only once in the truncation notice', async () => {
+    const client = makeClient(() =>
+      fromArray(makeTextStreamWithDroppedTool('partial answer', ['read_file', 'read_file'])),
+    );
+    const dispatcher = makeDispatcher(() => Promise.resolve({ content: 'never called' }));
+    const events = await collect(
+      runTurn({
+        client,
+        messages: [{ role: 'user', content: 'do a thing' }],
+        system: null,
+        tools: null,
+        toolDispatcher: dispatcher,
+        model: 'claude-test',
+        maxTokens: 8_000,
+        headers: {},
+        signal: new AbortController().signal,
+        ctx,
+      }),
+    );
+    const msgs = events.filter((e) => e.type === 'assistant.message') as Array<{ text: string }>;
+    expect(msgs).toHaveLength(1);
+    expect(msgs[0]!.text.match(/read_file/g)).toHaveLength(1);
   });
 
   // #952: text-only truncation (no tool_use) → notice without a tool name.
--- a/src/agent/providers/anthropic-direct/loop.orphan.test.ts
+++ b/src/agent/providers/anthropic-direct/loop.orphan.test.ts
@@ -121,6 +121,102 @@ describe('loop.ts runTurn — orphan tool_use prevention', () => {
     expect(flatTooluses).not.toContain('toolu_orphan');
   });
 
+  // #952: the strip above is silent — a dropped tool call reads as the agent
+  // stalling. runTurn must surface a display-only truncation notice NAMING the
+  // dropped tool so the operator knows the action never ran.
+  it('surfaces a truncation notice naming the dropped tool when max_tokens strips a tool_use', async () => {
+    const truncatedToolUseStream: RawMessageStreamEvent[] = [
+      {
+        type: 'message_start',
+        message: {
+          id: 'msg_truncate2',
+          type: 'message',
+          role: 'assistant',
+          content: [],
+          model: 'claude-test',
+          stop_reason: null,
+          stop_sequence: null,
+          usage: baseUsage(),
+        },
+      } as unknown as RawMessageStreamEvent,
+      {
+        type: 'content_block_start',
+        index: 0,
+        content_block: { type: 'tool_use', id: 'toolu_dropped', name: 'read_file', input: {} },
+      } as unknown as RawMessageStreamEvent,
+      {
+        type: 'content_block_delta',
+        index: 0,
+        delta: { type: 'input_json_delta', partial_json: '{"file":"a.ts"}' },
+      } as unknown as RawMessageStreamEvent,
+      { type: 'content_block_stop', index: 0 } as unknown as RawMessageStreamEvent,
+      {
+        type: 'message_delta',
+        delta: { stop_reason: 'max_tokens', stop_sequence: null },
+        usage: { output_tokens: 4 },
+      } as unknown as RawMessageStreamEvent,
+      { type: 'message_stop' } as unknown as RawMessageStreamEvent,
+    ];
+
+    const client = makeClient(() => fromArray(truncatedToolUseStream));
+    const dispatcher = makeDispatcher(() => Promise.resolve({ content: 'never called' }));
+    const events = await collect(
+      runTurn({
+        client,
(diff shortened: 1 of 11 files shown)


griffinwork40:
## Automated review — `/review` (agent-afk)

**Decision: MERGE — 0 blocking (1 low, 1 nit; 2 findings dropped in verification).**

Reviewed ref `5877ceca`. Regime: light (hotfix, +203/−12, 6 files). CI green at review time.

The headline: the first pass filed a **blocking medium** claiming the "display-only" characterization was false — that the notice reaches `conversationHistory` and therefore re-enters model context. Independent re-derivation **refuted** it on the load-bearing link. Details below, because the refutation is more useful than the finding would have been.

---

### Refuted finding (dropped — recorded for transparency)

**Claim:** the truncation notice is pushed to `conversationHistory` by `stream-consumer.ts`, so it threads back as model context on the next turn and on `--resume`.

- **Half true, and the true half is harmless.** The push is real and verified at `src/agent/session/stream-consumer.ts:347-351` (`deps.conversationHistory.push(assistantMessage)`).
- **But `conversationHistory` is not model context.** The anthropic-direct provider builds requests from a separate `input.messages` (`MessageParam[]`) array — `loop/round-request.ts:138` passes `messages: input.messages` to the API call. Grepped at the ref: `conversationHistory` appears **only** in `session/agent-session.ts`, `session/stream-consumer.ts`, and `session/subagent-output-capture.ts` — never in any provider request-building path.
- **Conclusion: the PR's "display-only / not pushed to history" claim is accurate.** `turn-terminal.ts` deliberately yields the notice without an `input.messages.push`, exactly as the `refusal` and overload-exhausted notices do.

**One genuinely useful by-product — a pre-existing comment is wrong.** `src/agent/providers/anthropic-direct/loop.ts:230-236` (not touched by this PR) asserts the opposite:

> `It is NOT invisible to the model, though — session/stream-consumer.ts materializes every non-empty assistant.message into conversationHistory, which threads back as model context on the next turn and on --resume.`

That contradicts the verified behavior above. This PR makes that contract load-bearing for a *third* notice, so the stale comment is now likelier to mislead. **Out of scope for this PR** — flagging for a follow-up, not as a merge condition.

---

### Findings

**1 · low · blocking:false · confidence:high · correctness**
`src/agent/subagent/result.ts:112-113` · ref:5877ceca · citation-type:diff-context

**Finding:** The partial-result banner hardcodes the Anthropic-specific token name on a path the PR deliberately made provider-agnostic, so a truncated **OpenAI-compatible** child (`stop_reason: 'length'`) is labeled with a `max_tokens` that never appears in its provider's vocabulary.
```ts
: isTruncationStopReason(stopReason)
  ? 'was cut off at the output-token limit (max_tokens) before finishing'
```
**Suggestion:** drop the parenthetical — `'was cut off at the output-token limit before finishing'` — since the sentence is already unambiguous.

---

**2 · nit · blocking:false · confidence:high · correctness**
`src/agent/providers/anthropic-direct/loop/turn-terminal.ts:28` · ref:5877ceca · citation-type:diff-context

**Finding:** The notice string hardcodes `stop_reason "max_tokens"`; this is correct today because `isTruncationStopReason` is only reached from the anthropic-direct terminal path (`turn-terminal.ts:105` is its sole provider-side consumer), so there is no production impact — it becomes wrong only if the emit is ever shared with the openai path, which is the PR's named follow-up.
```ts
const base =
  '⚠ This turn was cut off at the output-token limit (stop_reason "max_tokens") before the ' +
```
**Suggestion:** when the openai live-surface notice lands, move this literal behind the same shared predicate rather than duplicating the string.

---

### Verification manifest

| Claim | Status |
|---|---|
| Notice reaches `conversationHistory` | **verified** at `stream-consumer.ts:347-351` |
| `conversationHistory` → next model request | **REFUTED** — requests build from `input.messages` (`round-request.ts:138`); finding dropped |
| "Display-only" is a novel pattern | **refuted** — pre-existing; 2 sibling notices (`turn-terminal.ts:59-61` refusal, `overload-pause.ts:46-48` overload) predate this PR |
| "no `finish_reason`/`length` handling in openai-compatible" | **false-absent** (142 matches) — finding built on it dropped; the openai live-surface gap is explicitly deferred by the PR body, so it is not an unmet requirement |
| "no production parsing of child result text" (used to clear the banner-breaks-structured-output risk) | **false-absent** (16 matches, incl. `result.ts:270 outputSchema.safeParse`) — but the risk is **independently resolved**: `buildResultFromMessage` parses raw `message.content` at `result.ts:269-270`, while `annotateIfIncomplete` is applied downstream by consumers (`dag-subagent.ts:206`, `skill-executor/fork-result.ts:77`, `foreground-promotion.ts:326`). The banner never reaches schema parsing. No finding. |

The single-sourced `isTruncationStopReason` in `providers/shared/truncation.ts` is a clean fix for the stated drift problem: `closure-reason.ts` re-exports it unchanged, so no importer breaks, and the `'max_tokens' || 'length'` literals now have one home.

### What was not checked
Citations verified inline against branch HEAD `5877ceca`. Stated intent: PR title+body — spec-compliance assessed against it, and the deferred openai live-surface notice is intent-declared, so it is not counted as unmet. No test suite, lint, or build was run (CI is green and reported separately). **Not conclusively closed:** whether `--resume` reconstructs model context from a persisted transcript that includes the notice — `conversationHistory` itself is not read by any resume path in `src/`, but transcript-backed resume was not traced end to end. That is the one residual path where the dropped finding could still bite.

<sub>Generated by `/review` on agent-afk. Read-only: no files, branches, or PR state were modified.</sub>

griffinwork40:
## `/fix-pr` — feedback addressed (pushed `dc055d8`)

### ⚠️ Correction to my earlier review

My previous comment on this PR concluded **MERGE — 0 blocking**. **That verdict was wrong.** The Codex P1 on `turn-terminal.ts:108` was correct and I missed it. Retracting the MERGE recommendation as originally issued; the defect it named is fixed below.

### What the P1 actually was (verified before fixing)

The truncation notice was emitted as a **second** `assistant.message`, after the one carrying the model's partial answer:

```ts
// turn-terminal.ts — before
if (turnResult.text.length > 0) {
  yield { type: 'assistant.message', text: turnResult.text, ... };   // the real answer
}
if (isTruncationStopReason(turnResult.stopReason)) {
  yield { type: 'assistant.message', text: truncationNotice(...), ... };  // the warning
}
```

Two consumers keep only the **last** assistant message of a turn:

- `src/agent/session/agent-session.ts:575-577` — `if (event.type === 'message' && event.message.role === 'assistant') { result = event.message; }`
- `src/agent/subagent/handle.ts:484-485` — `if (event.type === 'message') { finalMessage = event.message; }`

So on the non-streaming `sendMessage()` path (one-shot `afk chat`, SDK callers) and for **every truncated subagent**, the model's partial answer was discarded and replaced by the warning text. A truncated child returned the warning to its parent *instead of its work* — and `annotateIfIncomplete` then labeled the notice rather than salvaging anything. Streaming surfaces (REPL, Telegram) were unaffected, which is why this was easy to miss by eye.

That inverts the PR's own goal: the change meant to make truncation visible was destroying the partial output it was meant to preserve.

### Fix

`turn-terminal.ts` now emits **one** `assistant.message` carrying the partial answer followed by the notice (`\n\n` separated). When there is no partial text (e.g. a tool call truncated before any text block), the notice still stands alone exactly as before. No new event type; `stream-consumer.ts`, `agent-session.ts`, and `handle.ts` are untouched, and the notice still never enters `input.messages` (model context).

The `suggestion` event continues to carry the model's own text **without** the appended notice.

| # | Item | Status |
|---|---|---|
| 1 | Codex P1 — notice clobbers partial answer for last-wins consumers | **fixed** (`turn-terminal.ts`) |
| 2 | Review low — `(max_tokens)` parenthetical on the cross-provider path | **fixed** (`result.ts`) |
| 3 | Review nit — `truncationNotice` base string names `max_tokens` | out-of-scope — anthropic-only reach today; belongs with the deferred openai live-surface notice |
| 4 | Review note — stale contract comment at `loop.ts:230-236` | out-of-scope — pre-existing, `loop.ts` not touched by this PR |

### Test gap that let this through

The existing truncation tests asserted only at the **raw provider-event** level, so they passed while every last-wins consumer lost the answer. `loop.orphan.test.ts` now asserts a truncated turn with partial text produces **exactly one** `assistant.message` containing both the answer and the notice — a consumer-level invariant, not an event-count one.

### Verification (run locally in an isolated worktree, not just reported by the fixer)

- `pnpm test` full suite — **813 files / 15502 tests passed**, 2 pre-existing skips
- `pnpm build`, `audit:env:check`, `scan:env:check`, `audit:chalk:check`, `audit:sdk:check`, `fix:pins:check` — all pass

**One flake worth knowing about, unrelated to this PR:** the first full-suite run failed `src/agent/session/subagent-output-capture-wiring.test.ts > appends successive turns to ONE transcript per child` (expected ≥2 `### assistant`, got 1). It passes scoped at HEAD, passes on a full run at the unmodified base, and passes on a second full run at HEAD — it races on the capture writer's incremental flush under load. Pre-existing flake, not caused by this change, but it will intermittently redden CI.

<sub>Generated by `/fix-pr`. Pushed as a plain fast-forward to the PR branch; no force-push, no other ref touched.</sub>

griffinwork40:
Verification complete — the escalated claim holds; I re-derived every link myself at the ref.

---

# Code Review — PR #960

**Target:** `fix(tokens): surface max_tokens truncation instead of failing silently (#952)` · `fix/surface-max-tokens-truncation-952` → `main` · ref `f93568be` · MERGEABLE/CLEAN
**Regime:** full (340 lines, 7 files, 4 source) · **Waves:** 2 dimension agents → inline citation/absence verification → 2 shadow verifiers

---

## Blocking

### 1. `high · blocking:true · confidence:high · correctness` — `src/agent/providers/anthropic-direct/loop/turn-terminal.ts:113-121` · ref:f93568be · citation-type:diff-context
`[was: medium/waived-non-blocking, now: refuted → high/blocking]`

**Finding:** The new zero-text `else if (truncationText)` branch emits a non-empty `assistant.message` where the pre-merge code emitted none, which sets `finalMessage` in the subagent handle and short-circuits the ZERO-OUTPUT throw — silently suppressing the bounded zero-output re-dispatch for every truncated read-only child.

**Evidence — the chain, each link re-read at the ref:**
```
turn-terminal.ts:113-121   } else if (truncationText) { yield {type:'assistant.message', text: truncationText} }
stream-consumer.ts:349     if (event.text) { ... return { type: 'message', message } }   ← now passes (was dropped)
handle.ts:484              if (event.type === 'message') { finalMessage = event.message; }
handle.ts:522              if (finalMessage) return finalMessage;   ← returns BEFORE line 619
handle.ts:619              this.lastStopReason = STREAM_INCOMPLETE; throw new StreamIncompleteError(...)   ← now unreachable
stream-cut-retry.ts:68     result.isError === true && result.incompleteReason === STREAM_INCOMPLETE   ← now false
subagent-executor.ts:462   runWithStreamCutRetry({...})   ← returns on attempt 0, no re-dispatch
```

The repo already documents the exact reachability this breaks, at `handle.ts:608-612`:

> *"Any stopReason still set is a clean terminal one (end_turn / **max_tokens** / interrupted) reached via an empty-text turn **the stream consumer drops before emitting a `message` event** (see stream-consumer 'assistant.message': `if (event.text)`)"*

That comment is the load-bearing precondition for the deliberate `throw`-instead-of-`return` fix documented at `handle.ts:585-598` — whose History note names the failure being reintroduced here: *"a zero-output timeout reported as a SUCCESS … fooled every consumer whose only gate is `status === 'succeeded'` (it bit the forge skill and mint's phases)."* Post-merge, a `max_tokens` zero-text child resolves `succeeded` with content that is **only the synthesized operator warning** — zero findings.

**Reachability:** a truncated turn with zero model text is the PR's own headline scenario (a `tool_use` cut mid-arguments with no preamble), plus extended thinking consuming the whole output budget. `canRedispatch` limits the lost retry to side-effect-free children — which is `research-agent`, the most-dispatched child type in this repo.

**Why the 1179-test net stayed green:** no test pins `max_tokens` zero-output → `stream_incomplete` (grep of `src/agent/subagent/` test files at the ref: zero matches). The regression has no guard.

**Partial mitigation (real, not sufficient):** change (2) makes `isIncompleteStopReason('max_tokens')` true, so the parent still receives `[⚠ PARTIAL RESULT]` + `incomplete:true`. The parent model is warned; the automatic recovery and the `isError` signal are still lost.

**Suggestion:** Yield the standalone notice on a channel that does not set `finalMessage` — a distinct display-only event type rather than `assistant.message` — so `handle.ts:619`'s zero-output detection stays reachable. That also fixes the same overload in the pre-existing `refusal` branch, where `assistant.message` already means both "model content" and "harness warning".

---

## Non-blocking

**2. `low · blocking:false · confidence:high · correctness` — `stream-consumer.ts:347-353` · citation-type:file-state → verified**
The merged notice is persisted into harness-side `conversationHistory`, so the "display-only" comment at `turn-terminal.ts:24` is imprecise. Not new and not model-visible: the pre-existing `refusal` branch (`turn-terminal.ts:62-79`) has the identical property, `input.messages.push` at `:133` receives only `turnResult.assistantBlocks` filtered of `tool_use` (never `truncationText`), and `conversationHistory` is a display/`getHistory()`/ledger surface that is not converted back into provider request messages. `[was: medium/blocking, now: refuted → low]`

**3. `low · blocking:false · confidence:high · test-coverage` — `loop.orphan.test.ts:159-190`**
The `[...new Set(droppedToolNames)]` dedup path (`turn-terminal.ts:31`) has no test; only the single-tool case is exercised. Absence confirmed at the ref — `toolu_dropped` appears once, no test emits two `tool_use` blocks under `max_tokens`.

**4. `low · blocking:false · confidence:high · spec-compliance` — `truncation.ts:36-40`**
`'max_output_tokens'` (third commit) is absent from the PR body's "Changes" section, which describes the predicate as `'max_tokens' || 'length'`. The literal is correct — `responses-translate.ts:151` sets `state.finishReason = event.response?.incomplete_details?.reason` — and the module comment documents it; only the PR description is stale.

**5. `low · blocking:false · confidence:high · perf-observability` — `closure-reason.ts:113`**
Responses-wire truncated sessions now classify `truncated` instead of falling through to `model_end_turn`. Correct and intended; the telemetry-distribution shift is undocumented.

**6. `low · blocking:false · confidence:medium · perf-observability` — `turn-terminal.ts:98-121`**
The notice reaches the display stream only, never the trace/witness layer; `afk trace show` records `turn.completed` with the stop reason but no notice text. Consistent with stated intent.

**7. `nit · blocking:false · confidence:high · api-compat` — `truncation.ts:36`**
`TRUNCATION_STOP_REASONS: readonly string[]` erases element literals. `as const` would preserve `'max_tokens'|'length'|'max_output_tokens'` for future narrowing.

---

## Dropped / corrected in verification

| Item | Disposition |
|---|---|
| "No other production caller imports `isTruncationStopReason`" | **DROPPED — false-absent.** At the ref: `turn-terminal.ts:12`, `subagent/result.ts:17`, `closure-reason.ts:50`. (Moot regardless — widening a parameter to `string \| null \| undefined` is contravariantly safe.) |
| "`--max-output-tokens` flag unverified" | **RESOLVED — exists.** `chat.ts:142`, `interactive.ts:210`. `AFK_MAX_OUTPUT_TOKENS` registered at `env.ts:257-263` and is **not** deprecated (the "Deprecated and inert" text belongs to the adjacent `AFK_MAX_TOKENS` entry). Notice text accurate. |
| `STREAM_INCOMPLETE` cited at `stream-consumer.ts:609` | **Citation fabricated; substance correct.** Real sites: `handle.ts:532` (`??=`) and `handle.ts:619` (`=`). |
| 6 citations carrying diff-artifact line numbers | **Verified, lines corrected** to `turn-terminal.ts:94-96/113-121`, `truncation.ts:36-40`, `result.ts:84-91`, `closure-reason.ts:57/113`. |

**Also verified clean:** import direction (`session/` → `providers/shared/` matches the `soft-deadline.ts` precedent at `closure-reason.ts:49`); `truncation.ts` is a zero-import leaf, no cycle, no SDK symbol, no lock-file gate; `isolatedModules` value re-export legal; file sizes under the 350-LOC rule (`result.ts` at 346 — 4 lines of headroom); test helpers `makeTextStream`/`makeClient`/`makeDispatcher` exist at `loop.test-helpers.ts:45/171/181`; no secret or path exposure in the notice, which interpolates no model output.

**Consumer impact of `isIncompleteStopReason` widening — all safe:** the 6 mint phases (`plan.ts:69`, `research.ts:69`, `spec.ts:72`, `ship.ts:75`, `heal.ts:136`) now hard-fail on a truncated child instead of forwarding a partial; `parallelize-dispatch.ts:161` returns `kind:'failed'`; `bgsub.ts:137/257` and `foreground-promotion.ts:326/385` gain the partial banner and structured `incomplete` fields. No consumer previously relied on `max_tokens` returning `false`.

---

## What was not checked

- Citations verified inline against branch HEAD `f93568be` (present locally). Every `blocking`/`high` and every `file-state` citation was re-read at that ref.
- **Stated intent: PR #960 title + body — spec-compliance assessed.**
- Shadow-verify: 2 claims, 1 round, both `REFUTED`, both `independent-rederivation`; I then re-derived the escalated chain (finding 1) myself end-to-end at the ref rather than surfacing a single verifier's escalation. No claim went unverified.
- **Process deviation:** Wave 2 synthesis was performed inline rather than dispatched — I held verified ground truth for every citation, and a synthesis agent would have re-cited from a summary rather than source.
- Did not run the test suite, `pnpm lint`, or `pnpm build` (read-only review; the PR reports all three green).
- Did not review the openai-compatible `query.ts` live-turn surface — explicitly deferred by the PR as named follow-up.
- Did not trace whether `background-registry` durably persists `conversationHistory` to disk (bears only on non-blocking finding 2).
- Did not assess Telegram/REPL rendering differences for the merged message.

---

**Decision: DO NOT MERGE — 1 high blocking (correctness); 0 medium, 5 low, 1 nit.**

The observability goal is met and the shared-predicate refactor is sound. The blocker is a side effect the PR did not intend: making a zero-text truncated turn *visible* also made it look *non-empty*, which unhooks the zero-output failure path and its bounded retry. Fix is local to `turn-terminal.ts:113-121`.

## Done
- Reviewed PR #960 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability; 2 Wave-1 agents → inline citation/absence verification at ref `f93568be` → 2 shadow verifiers → inline synthesis.
- Evidence: report above is the artifact (transcript-only by design — `/review` is read-only); diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/24545676-d737-4fd0-af0c-d980ca8aa848/pr-960.diff`; every citation resolved against local ref `f93568be81f9b2da5c6ff0e6e800a9c44cb8a531`.
- World unchanged: no files written, no commits, no PR comments, no pushes. The single pre-existing dirty entry (`?? .afk/`) was left untouched.
- Pending: the blocking finding is reported, not fixed — resolving it is the author's call. Findings were not posted to GitHub; relay them manually if you want them on the PR.

---

griffinwork40:
## `/fix-pr` — blocking finding addressed (pushed `b9050737`)

Supersedes my earlier **DO NOT MERGE** review comment on this PR. The `high` finding it named is fixed below; CI is green on the new head.

### The defect

A truncated turn that produced **no model text** emitted the operator warning as its `assistant.message`. That notice is non-empty, which flipped a signal the zero-output detection chain depends on being **absent**:

```
non-empty assistant.message
  -> stream-consumer.ts:349  `if (event.text)` passes -> emits {type:'message'}
  -> handle.ts:484           finalMessage = event.message
  -> handle.ts:522           `if (finalMessage) return finalMessage`   (early return)
  -> handle.ts:619           stamp STREAM_INCOMPLETE + throw           <- SKIPPED
  -> child resolves `succeeded` with no isError, instead of `failed`
  -> stream-cut-retry.ts:68  isZeroOutputStreamCut false -> no re-dispatch
```

A truncated child returned to its parent as a **success whose entire content was the warning** — zero findings dressed as an answer. `handle.ts:608-612` names `max_tokens` explicitly as a stop reason reaching its zero-output branch "via an empty-text turn the stream consumer drops"; this PR broke that documented assumption.

By the time I fixed it, #961 had landed the same substitution on the **openai-compatible** path (`(truncationText ?? finalAssistantText)`), so the fix covers both providers.

### The fix

The notice now rides **only** on non-empty model text.

- `anthropic-direct/loop/turn-terminal.ts` — the standalone-notice `else if` branch is deleted.
- `openai-compatible/query.ts` — the terminal yield's zero-text fallback is `finalAssistantText` again, not the notice.

This is also the shape the earlier Codex P1 asked for, by the second route it offered.

### Accepted tradeoff

A truncation with **no** partial text stays invisible on live surfaces. That is unchanged from before #952 — not a regression — and it is the honest boundary of this PR: making it visible needs a display-only event that renders **without** becoming a terminal message. That is a follow-up, not something to bolt on here, because it needs wiring in `stream-consumer`, **both** CLI renderer switches (neither has a `default` arm, so a missing case silently renders nothing), and `telegram/streaming.ts`, which handles no panel events today. Shipping half of that wiring would reintroduce the same class of silent-no-op bug this PR exists to remove.

### Tests

| guard | what it pins |
|---|---|
| `loop.orphan.test.ts` | zero `assistant.message` events for a textless `max_tokens` turn |
| `query.truncation.test.ts` | **empty** `assistant.message` for a textless `length` turn |
| `handle-streaming.test.ts` | zero-output -> FAILED/`STREAM_INCOMPLETE` for every truncation sentinel |

The first two are the real regression guards — verified by re-introducing the branch and confirming the test fails, then reverting.

The third closes a gap worth naming: that guard's rationale already named `max_tokens`, but it only ever exercised `end_turn`. **That is how this regression passed a 1179-test green suite.** It now runs over every truncation sentinel — though note it feeds `OutputEvent`s to a mock session, so it pins the handle-side contract, not the provider emit path.

Also folded in: `TRUNCATION_STOP_REASONS` becomes an `as const` tuple so consumers keep the element literal union; `isTruncationStopReason` keeps its wide `string | null | undefined` parameter and widens at the `.includes` call.

### Verification

- `tsc --noEmit` clean; full suite **815 files / 15,517 tests** green locally; all CI checks green on `b9050737`.
- Re-reviewed by two independent agents. Both confirmed the chain is restored on both providers and the partial-text case still delivers the model's answer first in a single message.

### Known follow-ups (not blocking)

1. **Display-only notice channel** so a textless truncation is visible — the wiring map above is the spec.
2. No end-to-end test drives `max_output_tokens` through the Responses wire in `query.ts`; the mock only drives `chat.completions`. The shared `truncationNotice` is unit-tested for that sentinel, but the wire path is not.

griffinwork40:
Follow-ups from this PR are now tracked:

- **#970** — `assistant.message` is overloaded, so a textless truncation is either invisible or defeats zero-output detection. This is the deferred half of #952: the display-only notice channel, with the verified 5-site wiring map and the note that partial wiring fails as a silent no-op. Also covers migrating the `refusal` branch, which has the identical overload.
- **#971** — no end-to-end coverage for `max_output_tokens` truncation on the Responses wire; the `query.truncation.test.ts` mock drives only `chat.completions`. Not a known defect — a durability gap.

Neither blocks this PR.

PR #961, opened 2026-08-09
fix(tokens): surface max_tokens truncation on the openai-compatible turn too

griffinwork40:
## Summary

The deferred half of #960. That PR surfaced `max_tokens` truncation on **anthropic-direct** and explicitly named the openai-compatible live-turn notice as a follow-up rather than shipping a half-mirror. This is that follow-up.

**Stacked on #960** — base branch is `fix/surface-max-tokens-truncation-952`, not `main`. Merge #960 first; this diff then reduces to the four files below.

### The bug

On the openai-compatible path a truncated turn was worse than invisible. When the output cap cuts a round off mid-tool-call:

- `isToolCallStop` returns `false` for any explicit non-tool finish reason (`translate.ts`), so `needsToolDispatch` is false;
- the half-built call is discarded in memory — **correctly**, since a call with an empty `tool_call_id` would poison history and 400 the next request;
- but nothing was emitted in its place. The turn broke out of the loop with whatever partial text existed, which for a truncated tool call is `''`.

The operator saw an empty assistant message and an agent that appeared to stall. The new negative-control test reproduces precisely that against the pre-fix file:

```
AssertionError: expected '' to contain 'output-token limit'
```

### Changes

1. **`providers/shared/truncation.ts`** — `truncationNotice()` moves here from `anthropic-direct/loop/turn-terminal.ts` and is now shared by both providers. #960 deferred this work partly because a half-mirrored notice (one provider naming the dropped tool, the other not) would itself be provider drift; sharing the string makes that drift unrepresentable. `sentinelLabel()` renders the wire-accurate field name per provider — `stop_reason "max_tokens"` / `finish_reason "length"` / `incomplete_details.reason "max_output_tokens"` — so the quoted term is greppable in the right provider's docs.
2. **`openai-compatible/query.ts`** — capture `finalizedToolCalls(state).map(c => c.name)` at the break when the round ended on a truncation sentinel, and append the notice to the terminal `assistant.message`.
3. **`anthropic-direct/loop/turn-terminal.ts`** — imports the shared helper; local copy deleted. Behaviour unchanged (its existing assertions in `loop.orphan.test.ts` still pass untouched).

### Why append rather than emit a second message

Load-bearing, not stylistic. Last-wins consumers — the non-streaming `sendMessage()` path and a subagent's final-message capture — keep only the **last** assistant message of a turn, so a second event would discard the model's real partial answer and surface only the warning. The notice is appended **after** the `priorTurns` push, so it stays operator-facing and never enters conversation history. Same rule the anthropic path already documents at `turn-terminal.ts`.

## How was this tested?

- `pnpm lint` ✓ · `audit:sdk:check` ✓ · `audit:env:check` ✓ · `audit:chalk:check` ✓
- **1492 tests pass** across `openai-compatible/`, `anthropic-direct/`, `session/`, `subagent/` (116 files).
- New `query.truncation.test.ts` (4 tests) — this behaviour had **zero** coverage before: no test in the directory drove `finish_reason: 'length'` at all, which is how the gap survived the original change. Placed in a new file rather than the already 3,397-line `query.test.ts`.
  - notice appended to partial text, partial answer ordered first, exactly one `assistant.message`
  - a tool call truncated mid-arguments is named + marked `NOT dispatched`, and provably never ran (no `tool.start`/`tool.end`)
  - `turn.completed` carries `stopReason: 'length'` so closure classification still reads `truncated`
  - clean `finish_reason: 'stop'` emits no notice
- **Negative control**: each new assertion was confirmed to FAIL against the pre-fix `query.ts` (output quoted above), so the tests pin the fix rather than passing vacuously.

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes (touched files + openai-compatible/anthropic-direct/session/subagent suites)

### DIFF
--- a/src/agent/providers/anthropic-direct/loop/turn-terminal.ts
+++ b/src/agent/providers/anthropic-direct/loop/turn-terminal.ts
@@ -9,33 +9,12 @@
  */
 
 import type { ProviderEvent } from '../../../provider.js';
-import { isTruncationStopReason } from '../../shared/truncation.js';
+import { isTruncationStopReason, truncationNotice } from '../../shared/truncation.js';
 import type { RunTurnInput, TurnResult } from '../types.js';
 import type { TurnAccumulator } from './turn-accumulator.js';
 
 /** Text at or below this length is also offered as a `suggestion`. */
 const SUGGESTION_MAX_LENGTH = 200;
-
-/**
- * Operator-facing notice for a turn cut off at the output-token cap (#952).
- * `droppedToolNames` are the `tool_use` blocks about to be stripped from
- * history below — when non-empty, the model announced an action that was
- * truncated mid-request and never dispatched, which otherwise reads as the
- * agent stalling. Display-only: not pushed to history.
- */
-function truncationNotice(droppedToolNames: string[]): string {
-  const base =
-    '⚠ This turn was cut off at the output-token limit (stop_reason "max_tokens") before the ' +
-    'model finished — anything above is partial, not a complete answer.';
-  if (droppedToolNames.length > 0) {
-    const names = [...new Set(droppedToolNames)].join(', ');
-    return (
-      `${base} A tool call was truncated mid-request and was NOT dispatched (${names}); ` +
-      'that action did not run. Raise --max-output-tokens / AFK_MAX_OUTPUT_TOKENS, then retry.'
-    );
-  }
-  return `${base} Raise --max-output-tokens / AFK_MAX_OUTPUT_TOKENS to allow a longer reply.`;
-}
 
 /**
  * Emit the closing events for a non-`tool_use` stop reason, pushing the
@@ -92,7 +71,7 @@ export function* emitNonToolUseTerminal(
   // the LAST assistant message of a turn, so two messages here would silently
   // discard the real answer and surface only the warning.
   const truncationText = isTruncationStopReason(turnResult.stopReason)
-    ? truncationNotice(turnResult.toolUseBlocks.map((b) => b.name))
+    ? truncationNotice(turnResult.toolUseBlocks.map((b) => b.name), turnResult.stopReason)
     : null;
 
   if (turnResult.text.length > 0) {
--- a/src/agent/providers/openai-compatible/query.truncation.test.ts
+++ b/src/agent/providers/openai-compatible/query.truncation.test.ts
@@ -1,0 +1,172 @@
+/**
+ * Truncation-notice coverage for the openai-compatible turn loop (#952).
+ *
+ * The sibling anthropic-direct behaviour is pinned by
+ * `anthropic-direct/loop.orphan.test.ts`; this file pins the same
+ * operator-visible contract on the other provider, which reaches truncation
+ * through entirely different machinery (a derived `finalizedToolCalls(state)`
+ * view rather than a `TurnResult.toolUseBlocks` array).
+ *
+ * The behaviour under test had ZERO coverage before this file: no test in this
+ * directory drove `finish_reason: 'length'` at all, which is how the notice
+ * came to be deferred out of the original #952 change.
+ */
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import type OpenAI from 'openai';
+import type { ProviderEvent, ProviderUserTurn } from '../../provider.js';
+import type { AgentConfig } from '../../types/config-types.js';
+import {
+  __setOpenAIClientFactory,
+  buildQueryFromConfig,
+  type OpenAIClientFactory,
+} from './query.js';
+import type { OpenAIChunk } from './translate.js';
+
+let pendingChunks: OpenAIChunk[] = [];
+
+function installMockClient(): void {
+  const factory: OpenAIClientFactory = () =>
+    ({
+      chat: {
+        completions: {
+          create: async (args: { stream?: boolean }) => {
+            if (!args.stream) throw new Error('mock only supports streaming mode');
+            const chunks = pendingChunks.slice();
+            return (async function* () {
+              for (const c of chunks) yield c;
+            })();
+          },
+        },
+      },
+    }) as unknown as OpenAI;
+  __setOpenAIClientFactory(factory);
+}
+
+async function collect(query: AsyncIterable<ProviderEvent>): Promise<ProviderEvent[]> {
+  const out: ProviderEvent[] = [];
+  for await (const ev of query) out.push(ev);
+  return out;
+}
+
+async function* singleInput(content: string): AsyncIterable<ProviderUserTurn> {
+  yield { content };
+}
+
+function baseConfig(over: Partial<AgentConfig> = {}): AgentConfig {
+  return { model: 'gpt-4o-mini', apiKey: 'sk-test-key', ...over } as AgentConfig;
+}
+
+/** The turn's single terminal assistant.message (there must never be two). */
+function assistantMessages(events: ProviderEvent[]): string[] {
+  return events
+    .filter((e): e is Extract<ProviderEvent, { type: 'assistant.message' }> =>
+      e.type === 'assistant.message',
+    )
+    .map((e) => e.text);
+}
+
+beforeEach(() => {
+  pendingChunks = [];
+  installMockClient();
+});
+
+afterEach(() => {
+  __setOpenAIClientFactory(null);
+});
+
+describe('openai-compatible truncation notice (#952)', () => {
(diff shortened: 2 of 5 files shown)


griffinwork40:
@codex fix comments

griffinwork40:
## Code Review — PR #961
`fix(tokens): surface max_tokens truncation on the openai-compatible turn too`
Reviewed ref **`625003f9`** · base `fix/surface-max-tokens-truncation-952` (**PR #960, still open** — this is stacked) · 5 files, +304/−24 · regime `full`

---

### Process disclosure (read this before the findings)

Wave 1 degraded and I did not silently inline around it:
- The **correctness agent timed out at 45 min with zero output**. Re-dispatched with a 10-iteration cap and pre-loaded ref excerpts; the second attempt returned.
- The **security/api-compat agent hit its iteration cap** and returned a partial. I completed its dimensions inline.
- The post-synthesis verification round ran as a **bounded 2-verifier inline wave** rather than via the `/shadow-verify` dispatcher, to avoid a second timeout. The ≤3-claim, one-round bound was respected.

The decisive structural fact: **the working tree is `main` (dc39ca40), which contains neither this PR nor its base #960** — `shared/truncation.ts` does not exist there at all. Every sub-agent read `main`. Wave 1.5 re-read every citation at `625003f9`, which is why one finding was dropped and one was inverted.

### Dropped in verification

| Claim | Verdict |
|---|---|
| `closure-reason.ts` holds a divergent 2-value `isTruncationStopReason` | **Fabricated-at-ref → dropped.** The agent quoted `main`. At `625003f9` that file *imports and re-exports* the shared predicate (`closure-reason.ts:50,57`). |
| The notice is replayed to the model on `--resume` (general case) | **REFUTED → dropped.** `turn-handler.ts:419-427` builds `responseText` by accumulating streamed `content` deltas; the warning-bearing branch fires only when `!streamingStarted`. The persisted `TurnRecord.assistant` is clean pre-warning text. Only the narrow no-text edge survives, reframed as **#1**. |

### Findings

**1 · medium · blocking:false · confidence:medium · correctness** — `query.ts:721-728`, `stream-consumer.ts:349-352`, `turn-handler.ts:419-427` · `ref:625003f9` · citation-type:diff-context
When truncation yields no partial text — the PR's *flagship* case, a tool call cut mid-arguments emitting no content delta — `text` is the warning alone, `streamingStarted` is still false, so the warning becomes the persisted assistant turn and is replayed to the model as its own prior utterance on `--resume`: the precise outcome `truncation.ts:87-90` says must not happen.
`· waived:` verified inherited, not introduced — `turn-terminal.ts:92-99` emits the identical notice-only message on the anthropic path from base #960 ("the notice stands alone, exactly as before"), so #961 is faithfully mirroring it.
*Fix:* strip the notice in `turn-handler.ts` before `responseText` is recorded, or carry it on a distinct non-persistable field.

**2 · low · blocking:false · confidence:high · correctness/docs** — `shared/truncation.ts:87-90` · citation-type:diff-context
The docstring's "never pushed into conversation history" is true only of the provider's `priorTurns`; `stream-consumer.ts:351` pushes the notice-bearing text into `AgentSession.conversationHistory` unconditionally, which `getHistory()` (`agent-session.ts:1265`) exposes.

**3 · medium · blocking:false · confidence:medium · correctness** — `query.ts:596`, `truncation.ts:100-108` · citation-type:diff-context
**CONFIRMED by independent re-derivation.** `translate.ts:157-163` seeds `{id:'', name:''}`; `:164-165` set `id` and `name` under *independent* guards; `finalizedToolCalls` (`:206-208`) filters nothing, and `query.ts:596` bypasses the `isToolCallStop` non-empty-name check. `[''].join(', ')` → `''`, so the operator sees literally `NOT dispatched (); that action did not run.` Reachability **rare-but-real** — every accumulator test delivers `id`+`name` in one chunk, so this needs a backend that splits them; not provable from static code.
`· waived:` bounded cosmetic degradation of a warning string, no data effect, one-line fix.
*Fix:* `c.name || c.id || '<unnamed tool>'`.

**4 · medium · blocking:false · confidence:medium · observability** — `query.ts:711-720,733-735` · citation-type:diff-context
When the tools-stripped wind-down round is itself truncated, the operator sees the truncation notice while `finishTurn` overrides `stopReason` to the wind-down reason, so `turn.completed` reports `iteration_cap`/`timeout` and `closure-reason.ts:113` classifies the session as non-truncated.
`· waived:` metadata-only disagreement in a compound edge case; the override at `:733-735` is pre-existing context, unchanged by this diff.

**5 · low · blocking:false · confidence:high · test-coverage** — `query.ts:715-718` · citation-type:diff-context
The `canIncreaseOutputLimit` **wiring** is untested. Confirmed-absent at the ref: `git grep canIncreaseOutputLimit 625003f9` returns only `query.ts:715` (production) and `truncation.test.ts:7,17` (direct string-level calls). No test drives the `auth.source === 'chatgpt-oauth' && stopReason === 'max_output_tokens'` computation through `runTurn`.

`unverified — absence claim not conclusively established:` "no test drives `max_output_tokens` through the notice path" — at the ref that literal appears in `query-responses.test.ts`, `query.test.ts`, `oneshot.test.ts`, `request-body.test.ts`, `responses-translate.test.ts`. Those exercise the Responses wire at other layers; whether any reaches `runTurn`'s notice site is not established, so this is reported rather than filed.

### Clean verdicts (verified, not assumed)

- **Gate consistency — clean.** Only two paths reach `:711` (the `:598` text break and the `:606` wind-down break); every other exit in `553-690` is a `return` that never reaches the notice site. `usageFromState` maps `finishReason`→`stopReason` and `sumProviderUsage`'s `b.stopReason ?? a.stopReason` lets the truncated round's non-null value win, so `accumulatedUsage.stopReason` at `:711` is the same value that gated `:595` in that iteration. A stale or empty `droppedToolNames` cannot pair with a fired notice.
- **anthropic parity — "behaviour unchanged" verified.** `turn-terminal.ts:73-99` preserves both branches; `loop.orphan.test.ts:179-269` still asserts `output-token limit`, `NOT dispatched`, partial-text ordering, and suggestion-exclusion; `sentinelLabel('max_tokens')` reproduces the original `stop_reason "max_tokens"` string exactly.
- **security — no issues found.** The notice never reaches provider-bound `priorTurns`; `auth.source` is a read-only comparison matching three pre-existing sites.
- **api-compat — no issues found.** Both call sites migrated to the 3-arg exported signature in the same diff; `assistant.message.text` stays `string`.
- **spec-compliance — intent met, no creep.** `sentinelLabel` and `canIncreaseOutputLimit` are named in the PR body's own prose; in-scope refinements.

### What was not checked

Citations verified inline against PR head **`625003f9`** via `git show`/`git grep` at that ref. Working tree is `main` (dc39ca40) and contains neither #961 nor #960, so every sub-agent citation was re-read at the ref — one was dropped as false-at-ref. Stated intent: PR #961 title+body — **spec-compliance assessed**. Did not run `pnpm test` or `pnpm lint`, so the "1492 tests pass" and negative-control claims are unverified by execution. Finding #1's persistence path was confirmed for the **REPL surface only**; one-shot `afk chat` and Telegram were not traced. Base PR #960 was not itself reviewed. Note: #960 must merge first — GitHub reports #961 `MERGEABLE` against its stacked base.

**Decision: MERGE — 0 blocking (3 medium waived, 2 low).**

---

## Done

- Reviewed PR #961 read-only across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability; emitted a MERGE recommendation with 5 surviving findings, all non-blocking.
- Evidence: PR #961 head `625003f9e8ec6ac5e5899b34101e6ba6f0228718` on GitHub; diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/75942475-a93e-42e8-8a31-7279682dc7f2/pr-961.diff`; every citation re-verified at that ref via `git show`/`git grep`. This report is the deliverable — it was not written to a file or posted to the PR.
- World change: none. No files created, edited, or deleted; no git or `gh` write commands run; the pre-existing untracked `.afk/` remains the only dirty entry, untouched.
- Deferred: two Wave 1 sub-agents degraded (one 45-min timeout with zero output, one iteration-cap partial), so their dimensions were completed inline by the orchestrator rather than by an isolated agent — noted in the report's process disclosure. The `max_output_tokens`-through-`runTurn` coverage question is reported as `unverified` rather than resolved, and finding #1's persistence path is confirmed for the REPL surface only.
- Nothing to commit: this review made no code changes, per the skill's read-only constraint.

---

Issue #11, opened 2026-06-06
diagnose Step 2: measure post-fix outcome (apply structured patch + re-run reproducer in worktree)


griffinwork40:
## Summary

Close the diagnose verification loop by measuring the **post-fix** outcome: apply the proposed fix in the isolated worktree, re-run the reproducer, and report a **measured** pass/fail instead of a prediction.

This is "Step 2" of a staged plan. Prior work:
- **PR #8** — honest labeling: the read-only verifier's `reproducer_passed` was a static prediction mislabeled as an executed result; renamed `predicted_pass`, prompt reframed to match its read-only tools.
- **PR #10** — Step 1: the deterministic parent runs the reproducer **once, pre-fix, at the repo root** and injects the measured `{exitCode, stdout, stderr}` into the read-only verifier as ground truth (confirms the bug actually reproduces). Verifier gains no permissions.

Step 2 is what makes the skill's own description — "validates each in isolated worktrees" — fully true again.

## Goal

Today `predicted_pass` is the model's static code-reading prediction (Step 1 grounds it with a *pre-fix* baseline). Step 2 adds a *post-fix* measurement so a "this fix works" verdict is anchored to a real test run, not inference — without giving the model execution permissions (the **parent** runs everything; the verifier stays read-only).

## Design sketch

1. Add an optional `structured_patch` field (unified diff, or old/new text triples) to the hypothesis schema, alongside the existing prose `proposed_fix`.
2. Teach the hypothesis-generation agent to emit a `structured_patch` where it can.
3. Parent applies the patch in the worktree (`git apply`), re-runs the reproducer **in the worktree**, captures the post-fix exit code, and injects it as ground truth (extends Step 1's baseline mechanism to a post-fix measurement; a pre→post exit-code transition fail→pass is the real "fix works" signal).
4. When a real post-fix measurement exists, surface it as measured (e.g. a `measured_pass` field or an explicit `verification_log` marker) distinct from a prediction.
5. Prose-only hypotheses (no `structured_patch`) gracefully **degrade to Step-1 behavior** (pre-fix baseline + static prediction).

## Open design questions (the reasons this isn't a drive-by)

[...]

PR #964, opened 2026-08-09
feat(skills): positional $N arguments and fork-path substitution parity

griffinwork40:
## What

Two related gaps in SKILL.md argument handling.

### 1. No positional arguments

Bodies supported only `$ARGUMENTS` / `$ARGUMENT`, both expanding to the entire raw args string. A skill wanting "first arg here, second arg there" had to instruct the model to split the string itself.

Adds 1-indexed `$1`, `$2`, … `$N` with shell-style tokenization, so quoted phrases survive:

```
/skill "hello world" second   →   $1 = hello world,  $2 = second
```

Design calls worth reviewing:

- **A positional with no argument is left verbatim, not blanked.** `$3` with two args renders as a literal `$3`. That's a legible symptom of an author bug; silently emptying it hides the mistake from both author and model.
- **`$0` is not special** and is left alone rather than being assigned a meaning.
- **`\$1` escapes**, which any body documenting this syntax needs.
- **Unterminated quotes are tolerated**, not thrown on — this runs while rendering a prompt, where a hard failure would take down an otherwise-working skill over a typo.
- **No backslash-escape processing in the tokenizer**, so a Windows path survives intact.

### 2. Forked disk skills never got substitution at all

A user- or project-authored `SKILL.md` using `context: fork` received its body **verbatim** — placeholders unexpanded — while an identical *plugin*-authored skill got them substituted (`fork-dispatch.ts:241`). Same file format, different behaviour depending on where it lived.

Args continue to ride the child's user message as well. That's deliberate and matches the plugin path: the message anchors *which skill was asked for*, the body placeholder lets the author *position* the args.

## The single-pass property

Substitution is one pass over a combined pattern, and this is load-bearing rather than an optimization. Two sequential `.replace()` calls would let the positional pass rewrite text the `$ARGUMENTS` pass had just inserted — so args containing `$1` would be re-expanded. One pass makes inserted text inert by construction. Locked by two tests.

## Structure

Templating moves out of `load-mode.ts` into `arg-substitution.ts` — a tokenizer plus a substitution grammar is its own concern. `load-mode.ts` re-exports the symbols, so no import path changes anywhere.

## Explicitly not changed

`executeForkedRegistrySkill` still does no substitution. That path hands args to the child as its user message by design, and bundled `system.md` bodies do not reference placeholders by convention. It looked like a third gap; it isn't.

## Tests

21 new tests in `arg-substitution.test.ts` (tokenizer, `$ARGUMENTS` regression lock, positionals, escaping, single-pass), plus one in `user-skills.test.ts` for the fork path. Both new behaviours were verified to fail against unpatched source rather than pass vacuously:

```
→ expected 'All args: [$ARGUMENTS]. First: [$1]. …' to contain 'All args: ["hello world" second]'
```

| Gate | Result |
|---|---|
| `pnpm lint` | pass |
| `pnpm test:coverage` | 814 files, 15519 passed, 2 skipped, thresholds met |
| `pnpm audit:chalk:check` | 0 violations |

## Note for review

Claude Code's own docs are self-inconsistent on whether `$1` or `$0` is the first argument. I chose `$1` = first (the form its examples use). If we later confirm upstream differs, the change is one line in `arg-substitution.ts`.

### DIFF
--- a/src/agent/tools/skill-executor/arg-substitution.test.ts
+++ b/src/agent/tools/skill-executor/arg-substitution.test.ts
@@ -1,0 +1,119 @@
+/**
+ * Tests for SKILL.md argument templating.
+ *
+ * `$ARGUMENTS` behaviour is pre-existing and locked here against regression;
+ * positional `$N`, quoting, escaping and the single-pass property are new.
+ *
+ * @module agent/tools/skill-executor/arg-substitution.test
+ */
+
+import { describe, it, expect } from 'vitest';
+import { substituteSkillArgs, tokenizeArgs } from './arg-substitution.js';
+
+describe('tokenizeArgs', () => {
+  it('splits on whitespace and collapses runs', () => {
+    expect(tokenizeArgs('a b   c')).toEqual(['a', 'b', 'c']);
+    expect(tokenizeArgs('  leading and trailing  ')).toEqual(['leading', 'and', 'trailing']);
+    expect(tokenizeArgs('')).toEqual([]);
+  });
+
+  it('groups double- and single-quoted phrases and strips the quotes', () => {
+    expect(tokenizeArgs('"hello world" second')).toEqual(['hello world', 'second']);
+    expect(tokenizeArgs("'hello world' second")).toEqual(['hello world', 'second']);
+    expect(tokenizeArgs('pre "mid dle" post')).toEqual(['pre', 'mid dle', 'post']);
+  });
+
+  it('treats the opposite quote as literal inside a quoted run', () => {
+    expect(tokenizeArgs('"it\'s fine"')).toEqual(["it's fine"]);
+  });
+
+  it('preserves an empty quoted token so later positions do not shift', () => {
+    expect(tokenizeArgs('"" b')).toEqual(['', 'b']);
+  });
+
+  it('tolerates an unterminated quote rather than throwing', () => {
+    // Rendering a prompt must not hard-fail over a typo in a skill body.
+    expect(tokenizeArgs('a "unterminated here')).toEqual(['a', 'unterminated here']);
+  });
+
+  it('does not process backslash escapes (a Windows path survives intact)', () => {
+    expect(tokenizeArgs('C:\\Users\\me\\file.txt')).toEqual(['C:\\Users\\me\\file.txt']);
+  });
+});
+
+describe('substituteSkillArgs — $ARGUMENTS (pre-existing behaviour)', () => {
+  it('expands both $ARGUMENTS and $ARGUMENT to the full raw string', () => {
+    expect(substituteSkillArgs('run $ARGUMENTS now', 'a b c')).toBe('run a b c now');
+    expect(substituteSkillArgs('run $ARGUMENT now', 'a b c')).toBe('run a b c now');
+  });
+
+  it('expands to empty string when args are absent or empty', () => {
+    expect(substituteSkillArgs('x $ARGUMENTS y', undefined)).toBe('x  y');
+    expect(substituteSkillArgs('x $ARGUMENTS y', '')).toBe('x  y');
+  });
+
+  it('leaves a body with no placeholder unchanged', () => {
+    expect(substituteSkillArgs('nothing to do here', 'a b')).toBe('nothing to do here');
+  });
+
+  it('does not match a longer identifier such as $ARGUMENTSX', () => {
+    expect(substituteSkillArgs('$ARGUMENTSX', 'a')).toBe('$ARGUMENTSX');
+  });
+
+  it('inserts regex-special replacement patterns verbatim', () => {
+    // Guards the replacer-as-function contract: `$&`/`$$` in args must not be
+    // reinterpreted by String.prototype.replace.
+    expect(substituteSkillArgs('[$ARGUMENTS]', '$& $$ $` $1')).toBe('[$& $$ $` $1]');
+  });
+});
+
+describe('substituteSkillArgs — positional $N', () => {
+  it('expands 1-indexed positionals', () => {
+    expect(substituteSkillArgs('$1 then $2', 'first second')).toBe('first then second');
+  });
+
+  it('respects quoting when assigning positions', () => {
+    expect(substituteSkillArgs('[$1] [$2]', '"hello world" second')).toBe('[hello world] [second]');
+  });
+
+  it('leaves a positional with no corresponding argument VERBATIM', () => {
+    // A literal `$3` in the rendered prompt is a legible symptom of an author
+    // bug; silently blanking it would hide the mistake.
+    expect(substituteSkillArgs('$1 $2 $3', 'only one')).toBe('only one $3');
+  });
+
+  it('leaves $0 alone — it is not a positional', () => {
+    expect(substituteSkillArgs('$0 $1', 'a b')).toBe('$0 a');
+  });
+
+  it('supports multi-digit positionals', () => {
+    const args = 'a1 a2 a3 a4 a5 a6 a7 a8 a9 a10';
+    expect(substituteSkillArgs('$10', args)).toBe('a10');
+  });
+
+  it('coexists with $ARGUMENTS in one body', () => {
+    expect(substituteSkillArgs('all=[$ARGUMENTS] first=[$1]', 'x y')).toBe('all=[x y] first=[x]');
+  });
+});
+
+describe('substituteSkillArgs — escaping', () => {
+  it('renders an escaped placeholder literally, minus the backslash', () => {
+    expect(substituteSkillArgs('\\$1 and \\$ARGUMENTS', 'a b')).toBe('$1 and $ARGUMENTS');
+  });
+
+  it('still expands unescaped placeholders alongside escaped ones', () => {
+    expect(substituteSkillArgs('\\$1 but $2', 'a b')).toBe('$1 but b');
+  });
+});
+
+describe('substituteSkillArgs — single-pass property', () => {
+  it('does not re-expand a placeholder that appears inside the args', () => {
+    // The load-bearing guarantee: two sequential .replace() passes would let
+    // the positional pass rewrite text the $ARGUMENTS pass just inserted.
+    expect(substituteSkillArgs('[$ARGUMENTS]', '$1 literal')).toBe('[$1 literal]');
+  });
+
+  it('does not re-expand when a positional token itself contains a placeholder', () => {
+    expect(substituteSkillArgs('[$1]', '"$2 stays" other')).toBe('[$2 stays]');
+  });
+});
--- a/src/agent/tools/skill-executor/arg-substitution.test.ts
+++ b/src/agent/tools/skill-executor/arg-substitution.test.ts
@@ -2,7 +2,7 @@
  * Tests for SKILL.md argument templating.
  *
  * `$ARGUMENTS` behaviour is pre-existing and locked here against regression;
- * positional `$N`, quoting, escaping and the single-pass property are new.
+ * positional `${N}`, quoting, escaping and the single-pass property are new.
  *
  * @module agent/tools/skill-executor/arg-substitution.test
  */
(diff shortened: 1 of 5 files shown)


griffinwork40:
@codex fix comments

griffinwork40:
Wave 1.5 complete — all citations resolved against the ref, all absence claims re-derived, and the load-bearing behavioral claims verified by executing the ref's own logic. Synthesizing.

---

# Code Review — PR #964

**`feat(skills): positional $N arguments and fork-path substitution parity`** · `feat/skill-positional-args` → `main`
Reviewed ref `d51e9ec8` · 5 files · +303 / −34 · regime **full** · stated intent: PR #964 title + body

---

## Verification manifest (Wave 1.5, inline)

**3 findings dropped, 3 corrected.** One of the drops removed the only blocking finding in the review.

| # | Wave 1 finding | Status | Evidence |
|---|---|---|---|
| C-5 | `medium·blocking:true` — empty-string `input` makes `systemPrompt` diverge from `userMessage` | **DROPPED — refuted** | Executed: `substituteSkillArgs(body,'')` and `(body,undefined)` return **byte-identical** output (`"x  [${1}] y"`), because `raw = args ?? ''` collapses both. `userMessage` yields `anchor` for both. Zero divergence. The `''`→`''` semantic is pre-existing and *pinned* at `skill-executor.test.ts` ("replaces `$ARGUMENT` with empty string when args is an empty string" → `'Args: []'`) |
| SC-1/SC-2 | stale `load-mode.ts` module doc | **DROPPED — refuted** | `load-mode.ts:10–11` at ref is correctly patched: *"Argument templating used to live here too; it moved to `arg-substitution.ts`"* |
| SC-5 | `low` — new import "contradicts *no import path changes anywhere*" | **DROPPED — refuted** | `fork-dispatch.ts:28` still reads `from './load-mode.js'`, unchanged at ref. No *existing* importer changed path; the intent's claim holds. Residual style point survives as nit #11 |
| C-1 | `medium·blocking:true` — `a"b c"d` tokenizes to one token | **CORRECTED → `low`** | Executed: `["ab cd"]` — this is **exactly POSIX shell behavior** (`sh -c 'echo a"b c"d'` → `ab cd`), so it *satisfies* the documented "shell-style tokenization" contract rather than violating it. Correctness claim refuted; test/doc gap survives |
| T-3 | double-backslash analysis | **CORRECTED** | Agent asserted `\\${1}` → "both backslashes consumed". Actual: `\\${1}` → `\${1}` — **one** backslash retained. Coverage gap stands, evidence replaced |
| Sec-3 | `Number(digits)` → `Infinity` | **CORRECTED** | 23-digit index → `1e23`, not `Infinity`. Outcome (literal passthrough) identical |

**Citation line numbers remapped.** Both agents cited *diff* line numbers as file lines. True locations at ref: `tokenizeArgs` `:35`, `substituteSkillArgs` `:100`, `body.replace` `:106`, escape branch `:110`, `Number(digits)` `:112`, `token ?? match` `:117`, re-export `load-mode.ts:164`.

**Disposition contradiction resolved.** Findings 6/7 arrived carrying `blocking:false` in the schema field *and* a trailing `· blocking preserved from pre-downgrade medium/high` clause — mutually exclusive. I carried the **schema field** through (`blocking:false`) and discarded the clause, because I re-derived both underlying reachability claims independently and both came back zero. Flagging explicitly so it can be overruled.

**Absence claims re-derived at the ref (all `confirmed-absent`):**
- `git grep -E '\$\{[0-9]+\}' d51e9ec8 -- '*.md'` → **0 matches**; across all of `src/` only the 4 files this PR touches.
- Installed runtime tree (`~/.afk/skills/`, `~/.afk/plugins/`) → 2 hits, both inside `node_modules` (playwright `xdg-open`, `typescript.js`); **no SKILL.md body**.
- `git grep -E '\\\$ARGUMENTS?' d51e9ec8` → only the new test + the new file's own doc comment.
- `git grep -l '\$ARGUMENTS\?\b' d51e9ec8 -- ':!src/*'` → **0 files**. No docs, README, `CLAUDE.md`, `AFK.md`, or `website/` mention the placeholder syntax at all.

**Behavior verified by executing the ref's own logic** (functions transcribed verbatim from `git show d51e9ec8:…/arg-substitution.ts`, transcription printed for audit):

```
$1 and ${1}            + "first"        → "$1 and first"      ← bare $1 is LITERAL
Budget $1.00; awk '{print $1}'          → unchanged           ← braces guard works
${SKILL_ROOT}/x                         → unchanged           ← no collision
[$ARGUMENTS]           + "${1} literal" → "[${1} literal]"    ← single-pass holds
[${1}][${2}]           + '"" b'         → "[][b]"
${007} / ${00} / ${9×23}                → "g" / literal / literal
```

---

## Findings — 7 low, 4 nit, 0 blocking

**1 · `low · blocking:false · high · spec-compliance · src/agent/tools/skill-executor/arg-substitution.ts:106 · ref:d51e9ec8 · diff-context`**
The PR title, the worked example, and three design bullets all document bare `$1`/`$2`/`$3`/`$0`/`\$1`, but the regex requires braces, so an author following the PR description writes a placeholder that renders literally with no error or warning.
```js
/(\\?)\$(ARGUMENTS?\b|\{(\d+)\})/g
```
Verified: `substituteSkillArgs('$1 and ${1}', 'first')` → `"$1 and first"`.
*Suggestion:* rewrite every `$N` in the PR body as `${N}` — the braced form is the better design (it is what keeps `$1.00` and `awk '{print $1}'` literal, both verified), so the description is what should change, not the code.

**2 · `low · blocking:false · high · spec-compliance · repo-wide · ref:d51e9ec8 · file-state (verified)`**
No user-facing documentation for the placeholder syntax ships in this diff and none exists in the repo, so finding #1 leaves the PR body as the only description of the feature anywhere — and it is wrong.
Evidence: `git grep -l '\$ARGUMENTS\?\b' d51e9ec8 -- ':!src/*'` → 0 files.
*Suggestion:* add a `${N}` / `$ARGUMENTS` section to `docs/reference.md`. (The gap is pre-existing — `$ARGUMENTS` was undocumented too — this PR widens it.)

**3 · `low · blocking:false · low · api-compat · arg-substitution.ts:106 · ref:d51e9ec8 · diff-context · [UNVERIFIED: no production importers]`**
A body containing `${1}` inside a bash snippet, JSON template, or CI YAML now silently expands to argument 1 instead of rendering literally. Reachability: zero occurrences in any shipped `*.md` at the ref and zero SKILL.md bodies in the installed runtime tree. *Suggestion:* name the `${N}` reservation in the release notes so third-party skill authors can grep their own bodies.

**4 · `low · blocking:false · low · api-compat · arg-substitution.ts:110 · ref:d51e9ec8 · diff-context · [UNVERIFIED: no production importers]`**
`\$ARGUMENTS` changes from "backslash preserved, placeholder substituted" to "backslash stripped, placeholder literal" — a silent behavior change for any body relying on the old form. Zero shipped occurrences. `if (escape === '\\') return match.slice(1);`

**5 · `low · blocking:false · high · test-coverage · arg-substitution.test.ts · ref:d51e9ec8 · diff-context`**
The empty quoted token is tested at the tokenizer layer only; nothing pins it at the substitution layer, where the `??` (not `||`) choice is what makes it work. Verified `[${1}][${2}]` + `"" b` → `[][b]`. *Suggestion:* add `expect(substituteSkillArgs('[${1}][${2}]', '"" b')).toBe('[][b]')`.

**6 · `low · blocking:false · high · test-coverage · arg-substitution.test.ts · ref:d51e9ec8 · diff-context`**
No test pins the double-backslash case; verified behavior is `\\${1}` → `\${1}` (one backslash retained, placeholder literal), which is the sensible outcome but is currently unlocked. *Suggestion:* add that assertion beside the existing escaping tests.

**7 · `low · blocking:false · high · test-coverage · arg-substitution.ts:35–68 · ref:d51e9ec8 · diff-context`**
Adjacent quoted/unquoted runs concatenate into one token (`a"b c"d` → `["ab cd"]`) — correct POSIX shell semantics, but the contract text says only "whitespace separates tokens" and no test locks it. *Suggestion:* one test plus one contract bullet.

**8 · `nit · blocking:false · high · test-coverage · arg-substitution.test.ts · ref:d51e9ec8 · file-state (verified)`**
PR body says "21 new tests"; the actual `it(` count at the ref is **22**.

**9 · `nit · blocking:false · high · security · arg-substitution.ts:112 · ref:d51e9ec8 · diff-context`**
`Number(digits)` is unbounded — a 23-digit index yields `1e23` and falls through to literal passthrough. No crash, no wrong substitution; cosmetic only. *Suggestion:* `if (index < 1 || index > 999) return match;`.

**10 · `nit · blocking:false · high · api-compat · fork-dispatch.ts:238 · ref:d51e9ec8 · file-state (verified)` — orchestrator observation, Wave 1.5**
The plugin fork path silently gained `${N}` support, but its Invariant comment still enumerates only `$ARGUMENT/$ARGUMENTS`. Same for `fork-dispatch.ts:10`.

**11 · `nit · blocking:false · high · spec-compliance · user-skills.ts:43 vs fork-dispatch.ts:28 · ref:d51e9ec8 · file-state (verified)`**
One symbol, two import paths: `user-skills.ts` imports `substituteSkillArgs` from `arg-substitution.js`, `fork-dispatch.ts` from `load-mode.js`. Both resolve; the split invites drift.

---

## Verified clean

- **Import surface intact** — re-export at `load-mode.ts:164`; `fork-dispatch.ts:28` unchanged; `tokenizeArgs` has zero prior consumers, so no name collision.
- **No pre-existing pin broken** — `skill-executor.test.ts` (~2941–3047) and `skill-load-mode.test.ts` (91–94, 267–321) assert only `$ARGUMENT(S)` and `${PLUGIN_ROOT}` forms. The new regex matches the former via `ARGUMENTS?\b` and cannot match the latter (non-digit); `${SKILL_ROOT}/x` verified unchanged.
- **Single-pass property holds** — the PR's load-bearing claim survives execution and is locked by two tests.
- **Braces-required design vindicated** — `$1.00` and `awk '{print $1}'` both verified literal. This is why finding #1 is a description bug, not a code bug.
- **No ReDoS** — no nested quantifiers, no ambiguous alternation; `tokenizeArgs` is a single O(n) pass.
- **No new secret surface** — neither function logs, writes, or persists the raw args string.
- **Prompt-injection surface unchanged** — `$ARGUMENTS` already injected the whole caller string into the system prompt pre-PR.

## What was not checked

- Citations verified inline against branch HEAD **`d51e9ec8`** (fetched locally); working tree is `main @ dc39ca40` and was not mutated.
- Stated intent: **PR #964 title + body — spec-compliance assessed.**
- **`/shadow-verify` was not dispatched.** Trigger condition did not fire: no `critical`/`high` survived and every surviving disposition sits at the default table. The two claims that would have been routed (reachability for #3/#4) I re-derived myself at the ref *and* against the installed runtime tree using `git grep`, a different method than the agents' `Grep`.
- **Residual gap:** reachability for #3/#4 covers this repo and this machine only. A third-party plugin or user `SKILL.md` elsewhere containing `${1}` in a shell snippet is silently affected on upgrade.
- Ordering between `${PLUGIN_ROOT}` expansion (`load-mode.ts`) and arg substitution was not traced; the interaction is pre-existing for `$ARGUMENTS`, so this PR does not change it.
- Tests were **not run** — gate results are taken from the PR body, not reproduced.
- **Protocol deviation:** Wave 2 synthesis was performed inline rather than dispatched to a sub-agent, because the drops and corrections above rest on execution evidence a shell-less synthesis agent could not reproduce. Waves 1 (×2) and the verification pass are unaffected.

---

**Decision: MERGE — 0 blocking (7 low, 4 nit). 1 `medium·blocking:true` dropped as refuted by execution; 3 findings dropped, 3 corrected.**

The implementation is sound and the one contradiction in the review — a blocking `medium` — did not survive re-derivation. Findings #1 and #2 are the ones worth acting on: the code is right and the PR body is wrong about the feature's surface syntax, and there is no other documentation to correct it.

**Done**
- Reviewed PR #964 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability via 2 parallel `research-agent` dimension agents, then verified every citation and absence claim inline against ref `d51e9ec8`.
- Evidence: the merge decision and full finding table are in this transcript; source diff at `/Users/griffinlong/.afk/state/skill-preflight/222566ff-dce6-4c34-b135-077fe38f5a49/pr-964.diff`; every verification command is re-runnable (`git grep … d51e9ec8`, `git show d51e9ec8:<path>`).
- World unchanged: no files written, no commits, no pushes, no PR comments — review is read-only and the dirty working tree at `main @ dc39ca40` was left untouched.
- Deferred: `/shadow-verify` not run (trigger did not fire — no critical/high, no disposition departures); tests not executed; Wave 2 synthesis run inline rather than dispatched, disclosed above.

---

PR #965, opened 2026-08-09
feat(plugins): load Claude Code commands/*.md as slash commands

griffinwork40:
## What

AFK read plugin `skills/**/SKILL.md` but ignored `commands/*.md` entirely. This adds it.

## Why this isn't "chasing a legacy format"

That was my first read too, and it was wrong. CC's docs do label `commands/` legacy — but the label describes where **new authoring** is steered, not support status. Every mention pairs it with an explicit continued-support guarantee, and there is no announced removal date (a direct GitHub ask for a deprecation timeline went unanswered and stale-closed).

Adoption has not followed the label. A large share of the plugins in **Anthropic's own official marketplace** ship `commands/` and **no `skills/` directory at all** — `commit-commands`, `code-review`, `pr-review-toolkit`, `feature-dev`, `ralph-loop`, `agent-sdk-dev`. The practical effect was that installing many real CC plugins into AFK produced no usable commands.

## Approach: adapter, not a parallel pipeline

A command file and a `SKILL.md` are near-identical — frontmatter plus a markdown prompt body. So a command is discovered in one new module and then travels as an ordinary `PluginSkillMetadata` through the collection, collision, listing, and dispatch machinery that already exists.

Both seams in `skill-bridge` are patched together — `collectSkillEntries` (registration) and `discoverPluginSkillBodies` (execution) — so a command can never register as a slash command without also being dispatchable.

Two things genuinely differ and are handled in the new module:

| | SKILL.md | commands/*.md |
|---|---|---|
| Name source | `name:` frontmatter | **file path** |
| Nesting | directory holds one skill | **subdirectory is a namespace** |

So `commands/review/security.md` → `/review:security`, matching CC (restored upstream July 2025 after a period where nesting was flattened).

**Precedence** comes free: `SKILL.md` is collected before `commands/`, so the existing first-wins guard produces CC's documented rule — *"if a skill and a command share the same name, the skill takes precedence"* — with no new logic. Locked by a test.

## The one judgement call: commands are excluded from the model manifest

Commands register as slash commands and appear in `/skills`, but are **not** injected into the model-facing skill manifest.

Two reasons:

1. **`buildSkillManifest` has no character budget and no eviction policy.** CC can afford to list commands to its model because it caps the listing at ~1% of the context window and drops least-used descriptions on overflow. We have no such backstop, so every entry costs prompt tokens on every request, forever — and a third-party command surface is unbounded.
2. **Behavioural.** Installing a plugin should not hand the model a menu of commands the user never mentioned and may not know exist.

Filtered at the model-facing boundary rather than in `collectSkillEntries` — the same split the existing `excludeName` filter uses — so the slash router, `/skills`, and `afk skill list` all still see the complete set.

Happy to flip this if you'd rather match CC exactly; it's one `.filter()`.

## Drive-by fix

The new tests exposed a pre-existing bug. `parseSkillMetadata` accepted only camelCase `argumentHint`, so the kebab-case **`argument-hint`** that Claude Code and the Agent Skills standard actually specify was **silently dropped from every plugin SKILL.md** — while `user-skills.ts:126`, reading the identical file, honoured both spellings. Now accepts both.

## Tests

- `command-files.test.ts` — 27 tests (20 under `extractPluginCommands`, 7 under `normalizeSkillSource`): name derivation, `:` namespacing (incl. nested), `origin` tagging, path-beats-frontmatter-`name`, frontmatter parsing, dotfile/non-`.md` skipping, empty-body skipping, deterministic sort, malformed-file resilience, control-byte rejection in both filenames and directory segments (C0/DEL/C1), and symlink-containment rejection.
- `skill-bridge.test.ts` — 5 integration tests: `source: 'command'` tagging, manifest exclusion, body dispatchability, end-to-end namespacing, and skill-beats-command precedence.

| Gate | Result |
|---|---|
| `pnpm lint` | pass |
| `pnpm test:coverage` | 814 files, 15513 passed, 2 skipped, thresholds met |
| `pnpm audit:chalk:check` | 0 violations |
| `pnpm fix:pins:check` | pins up to date |

## Not included, deliberately

`` !`cmd` `` bash pre-execution and `@file` inlining — the remaining CC command features. `` !`cmd` `` executes shell at prompt-render time with no tool-call moment to gate, so it would bypass the entire PreToolUse chain (`bash-restriction-hook` only fires on `PreToolUse` + `toolName === 'bash'` via `runPreDispatchGates`), and both bash-restriction and path-approval fail open on headless surfaces where AFK runs unattended. Installing a plugin would become installing recurring, model-unmediated RCE. Cut rather than deferred silently; if revisited it needs a synthesized `ToolCall`, a plugin-manifest opt-in (never command-file self-grant), and an unconditional headless disable.

## Stacking

Independent of #963 and #964, but #963 (surfacing shadowing notices) is a genuine prerequisite in spirit: this PR adds a significant number of new names to the shared slash namespace, and #963 is what makes a resulting collision visible to the user instead of silent.

### DIFF
--- a/src/agent/plugins/command-files.test.ts
+++ b/src/agent/plugins/command-files.test.ts
@@ -7,7 +7,7 @@
  * @module agent/plugins/command-files.test
  */
 
-import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from 'fs';
 import { join } from 'path';
 import { tmpdir } from 'os';
@@ -188,6 +188,44 @@ describe('normalizeSkillSource', () => {
   it('still handles BOM+CRLF when combined with control stripping and truncation', () => {
     const result = normalizeSkillSource('\uFEFF---\r\ndescription: d\r\n---\r\n\r\nBody.\r\n');
     expect(result).toBe('---\ndescription: d\n---\n\nBody.\n');
+  });
+
+  it('strips a terminal escape embedded in an untrusted plugin filename', () => {
+    // The AFK_DEBUG skip diagnostics interpolate raw `readdirSync` entries
+    // from a third-party plugin tree. Without sanitisation, a directory or
+    // file named with a CSI/OSC sequence executes against the operator's
+    // terminal the moment discovery logs the skip.
+    const tmp = mkdtempSync(join(tmpdir(), 'cmd-esc-'));
+    try {
+      vi.stubEnv('AFK_DEBUG', '1');
+      mkdirSync(join(tmp, 'commands'), { recursive: true });
+      writeFileSync(join(tmp, 'commands', 'evil\x1b[2J\x1b[H.txt'), 'x');
+
+      const writes: string[] = [];
+      const spy = vi
+        .spyOn(process.stderr, 'write')
+        .mockImplementation((chunk: unknown) => {
+          writes.push(String(chunk));
+          return true;
+        });
+      extractPluginCommands(tmp);
+      spy.mockRestore();
+
+      const emitted = writes.join('');
+      expect(emitted).toContain('skipping');
+      expect(emitted).not.toContain('\x1b');
+    } finally {
+      vi.unstubAllEnvs();
+      rmSync(tmp, { recursive: true, force: true });
+    }
+  });
+
+  it('folds a lone CR to LF rather than stripping it as a control char', () => {
+    // Order invariant: CR (0x0D) falls inside the control-strip range
+    // [\x00-\x08\x0B-\x1F], so it survives only because newline
+    // normalisation runs FIRST and turns it into \n. Reordering the two
+    // passes would silently delete the line break instead of keeping it.
+    expect(normalizeSkillSource('a\rb')).toBe('a\nb');
   });
 
   it('never throws, and truncation alone never discards non-control content', () => {
--- a/src/agent/plugins/command-files.test.ts
+++ b/src/agent/plugins/command-files.test.ts
@@ -51,6 +51,39 @@ describe('extractPluginCommands', () => {
       body: 'Review this pull request: $ARGUMENTS',
       origin: 'command',
     });
+    // A frontmatter-free file declares no description; collectSkillEntries
+    // supplies the plugin-path fallback downstream.
+    expect(found[0]?.description).toBeUndefined();
+  });
+
+  it('parses a CRLF-authored command instead of leaking its frontmatter into the body', () => {
+    // A Windows-authored file fails a byte-exact `---\n` test, so without
+    // normalisation the whole YAML block was handed to the model as the prompt.
+    writeCommand('crlf.md', '---\r\ndescription: Ship it\r\n---\r\n\r\nDeploy the app.\r\n');
+    const found = extractPluginCommands(pluginDir);
+    expect(found).toHaveLength(1);
+    expect(found[0]?.description).toBe('Ship it');
+    expect(found[0]?.body).toBe('Deploy the app.');
+    expect(found[0]?.body).not.toContain('---');
+  });
+
+  it('parses a BOM-prefixed command instead of leaking its frontmatter into the body', () => {
+    writeCommand('bom.md', '\uFEFF---\ndescription: Ship it\n---\n\nDeploy the app.\n');
+    const found = extractPluginCommands(pluginDir);
+    expect(found).toHaveLength(1);
+    expect(found[0]?.description).toBe('Ship it');
+    expect(found[0]?.body).toBe('Deploy the app.');
+    expect(found[0]?.body).not.toContain('---');
+  });
+
+  it('skips a path segment containing the namespace separator', () => {
+    // `a:b.md` would derive the same name as `a/b.md`; registering both lets
+    // the first-wins guard drop one at random.
+    writeCommand('a:b.md', '---\ndescription: d\n---\n\nBody.\n');
+    writeCommand('a/b.md', '---\ndescription: d\n---\n\nBody.\n');
+    const names = extractPluginCommands(pluginDir).map((c) => c.name);
+    expect(names).toEqual(['a:b']);
+    expect(names).toHaveLength(1);
   });
 
   it('namespaces a subdirectory with a colon (CC parity)', () => {
@@ -117,5 +150,7 @@ describe('extractPluginCommands', () => {
     writeCommand('bad.md', '---\ndescription: unterminated frontmatter');
     const names = extractPluginCommands(pluginDir).map((c) => c.name);
     expect(names).toContain('good');
+    // The malformed sibling must be dropped, not registered with a raw body.
+    expect(names).not.toContain('bad');
   });
 });
--- a/src/agent/plugins/command-files.test.ts
+++ b/src/agent/plugins/command-files.test.ts
@@ -8,7 +8,7 @@
  */
 
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
-import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from 'fs';
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync, symlinkSync } from 'fs';
 import { join } from 'path';
 import { tmpdir } from 'os';
 import { extractPluginCommands } from './command-files.js';
@@ -98,6 +98,60 @@ describe('extractPluginCommands', () => {
     expect(names).toEqual(['a:b']);
     expect(names).toHaveLength(1);
     expect(found[0]?.body).toBe('Slash body.');
+  });
+
+  it('skips a .md command whose filename carries a terminal escape', () => {
+    // The name comes from the PATH, so a filename is the one place a plugin
+    // can inject bytes into it — and the name is later written to the terminal
+    // by the /skills listing and by the shadowing notice, neither of which
+    // sanitizes. The fixture MUST end in `.md`: a `.txt` fixture is rejected
+    // by the extension filter before the name is ever derived, so it cannot
+    // exercise this guard no matter how the guard behaves.
+    writeCommand('evil\x1b[2J\x1b[H.md', '---\ndescription: d\n---\n\nEvil body.\n');
+    writeCommand('safe.md', '---\ndescription: d\n---\n\nSafe body.\n');
+    const found = extractPluginCommands(pluginDir);
+    expect(found.map((c) => c.name)).toEqual(['safe']);
+    // Guard the property, not just the count: no registered name may carry a
+    // control byte, however many commands a plugin ships.
+    for (const c of found) expect(c.name).not.toMatch(/[\u0000-\u001F\u007F-\u009F]/);
+  });
+
+  it('skips a subdirectory segment carrying a control byte', () => {
+    // Directory segments reach the name through `segments`, a different route
+    // than the filename `base` — both must be rejected by the single guard.
+    writeCommand('ev\x07il/cmd.md', '---\ndescription: d\n---\n\nEvil body.\n');
+    writeCommand('ok/cmd.md', '---\ndescription: d\n---\n\nOk body.\n');
+    const found = extractPluginCommands(pluginDir);
+    expect(found.map((c) => c.name)).toEqual(['ok:cmd']);
+  });
+
+  it('skips a filename carrying a C1 control byte (8-bit CSI)', () => {
+    // U+009B is CSI in 8-bit form: terminals in 8-bit mode act on it exactly
+    // as they do on ESC-[, so restricting the guard to C0 would leave a
+    // working vector open.
+    writeCommand('evil\u009b2J.md', '---\ndescription: d\n---\n\nEvil body.\n');
+    expect(extractPluginCommands(pluginDir)).toEqual([]);
+  });
+
+  it('excludes a commands/ entry symlinked outside the plugin tree', () => {
+    // `resolveContained` is the control stopping a plugin from symlinking a
+    // private key or credential file into `commands/` and having it become a
+    // dispatchable prompt. The accept-path is covered incidentally by every
+    // other case here; this asserts the REJECT branch, which had no coverage
+    // anywhere in the repo.
+    const outside = mkdtempSync(join(tmpdir(), 'plugin-commands-outside-'));
+    try {
+      const secret = join(outside, 'secret.md');
+      writeFileSync(secret, '---\ndescription: d\n---\n\nSecret body.\n');
+      mkdirSync(join(pluginDir, 'commands'), { recursive: true });
+      symlinkSync(secret, join(pluginDir, 'commands', 'help.md'));
+      writeCommand('safe.md', '---\ndescription: d\n---\n\nSafe body.\n');
+      const found = extractPluginCommands(pluginDir);
+      expect(found.map((c) => c.name)).toEqual(['safe']);
+      expect(found.map((c) => c.body).join('')).not.toContain('Secret body.');
+    } finally {
+      rmSync(outside, { recursive: true, force: true });
+    }
   });
 
   it('namespaces a subdirectory with a colon (CC parity)', () => {
--- a/src/agent/plugins/command-files.test.ts
+++ b/src/agent/plugins/command-files.test.ts
@@ -12,6 +12,7 @@ import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from 'fs';
 import { join } from 'path';
 import { tmpdir } from 'os';
 import { extractPluginCommands } from './command-files.js';
+import { normalizeSkillSource, MAX_SKILL_SOURCE_CHARS } from './source-guard.js';
 
 let pluginDir: string;
 
@@ -86,12 +87,17 @@ describe('extractPluginCommands', () => {
 
   it('skips a path segment containing the namespace separator', () => {
     // `a:b.md` would derive the same name as `a/b.md`; registering both lets
-    // the first-wins guard drop one at random.
-    writeCommand('a:b.md', '---\ndescription: d\n---\n\nBody.\n');
-    writeCommand('a/b.md', '---\ndescription: d\n---\n\nBody.\n');
-    const names = extractPluginCommands(pluginDir).map((c) => c.name);
+    // the first-wins guard drop one at random. Distinct bodies make it
+    // possible to tell WHICH file survived: deleting the colon guard would
+    // let `a:b.md` register (or race with `a/b.md`), and either outcome must
+    // fail this assertion since only the `a/b.md` body is expected to survive.
+    writeCommand('a:b.md', '---\ndescription: d\n---\n\nColon body.\n');
+    writeCommand('a/b.md', '---\ndescription: d\n---\n\nSlash body.\n');
+    const found = extractPluginCommands(pluginDir);
+    const names = found.map((c) => c.name);
     expect(names).toEqual(['a:b']);
     expect(names).toHaveLength(1);
+    expect(found[0]?.body).toBe('Slash body.');
   });
 
   it('namespaces a subdirectory with a colon (CC parity)', () => {
@@ -160,5 +166,33 @@ describe('extractPluginCommands', () => {
     expect(names).toContain('good');
     // The malformed sibling must be dropped, not registered with a raw body.
     expect(names).not.toContain('bad');
+  });
+});
+
+describe('normalizeSkillSource', () => {
+  it('strips NUL bytes', () => {
+    expect(normalizeSkillSource('a\x00b')).toBe('ab');
+  });
+
+  it('strips other C0 control chars but keeps \\t and \\n', () => {
+    expect(normalizeSkillSource('a\x01\x02b\tc\nd\x1F')).toBe('ab\tc\nd');
+  });
+
+  it('truncates over-cap input to exactly MAX_SKILL_SOURCE_CHARS', () => {
+    const oversized = 'x'.repeat(MAX_SKILL_SOURCE_CHARS + 100);
+    const result = normalizeSkillSource(oversized);
+    expect(result.length).toBe(MAX_SKILL_SOURCE_CHARS);
+    expect(result).toBe('x'.repeat(MAX_SKILL_SOURCE_CHARS));
+  });
+
+  it('still handles BOM+CRLF when combined with control stripping and truncation', () => {
+    const result = normalizeSkillSource('\uFEFF---\r\ndescription: d\r\n---\r\n\r\nBody.\r\n');
+    expect(result).toBe('---\ndescription: d\n---\n\nBody.\n');
+  });
+
+  it('never throws, and truncation alone never discards non-control content', () => {
+    expect(() => normalizeSkillSource('\x00\x00\x00')).not.toThrow();
+    expect(() => normalizeSkillSource('x'.repeat(MAX_SKILL_SOURCE_CHARS * 2))).not.toThrow();
+    expect(normalizeSkillSource('a\x00')).toBe('a');
   });
 });
--- a/src/agent/plugins/command-files.test.ts
+++ b/src/agent/plugins/command-files.test.ts
@@ -42,6 +42,17 @@ describe('extractPluginCommands', () => {
     expect(found[0]?.body).toContain('Deploy the app.');
   });
 
+  it('accepts a plain Markdown prompt without frontmatter', () => {
+    writeCommand('review.md', 'Review this pull request: $ARGUMENTS\n');
+    const found = extractPluginCommands(pluginDir);
+    expect(found).toHaveLength(1);
+    expect(found[0]).toMatchObject({
+      name: 'review',
+      body: 'Review this pull request: $ARGUMENTS',
+      origin: 'command',
+    });
+  });
+
   it('namespaces a subdirectory with a colon (CC parity)', () => {
     writeCommand('review/security.md', '---\ndescription: Sec review\n---\n\nCheck auth.\n');
     const found = extractPluginCommands(pluginDir);
@@ -103,7 +114,7 @@ describe('extractPluginCommands', () => {
     // One bad command in a third-party plugin must not break discovery for
     // every other command it ships.
     writeCommand('good.md', '---\ndescription: d\n---\n\nBody.\n');
-    writeCommand('bad.md', 'no frontmatter at all');
+    writeCommand('bad.md', '---\ndescription: unterminated frontmatter');
     const names = extractPluginCommands(pluginDir).map((c) => c.name);
     expect(names).toContain('good');
   });
--- a/src/agent/plugins/command-files.test.ts
+++ b/src/agent/plugins/command-files.test.ts
@@ -1,0 +1,110 @@
+/**
+ * Tests for Claude Code `commands/*.md` discovery.
+ *
+ * Fixture shape mirrors the plugin-agents suite in skill-bridge.test.ts:
+ * a tmpdir plugin root written per-case, passed by path.
+ *
+ * @module agent/plugins/command-files.test
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from 'fs';
+import { join } from 'path';
+import { tmpdir } from 'os';
+import { extractPluginCommands } from './command-files.js';
+
+let pluginDir: string;
+
+function writeCommand(relPath: string, contents: string): void {
+  const full = join(pluginDir, 'commands', relPath);
+  mkdirSync(join(full, '..'), { recursive: true });
+  writeFileSync(full, contents);
+}
+
+beforeEach(() => {
+  pluginDir = mkdtempSync(join(tmpdir(), 'plugin-commands-test-'));
+});
+afterEach(() => {
+  rmSync(pluginDir, { recursive: true, force: true });
+});
+
+describe('extractPluginCommands', () => {
+  it('returns [] when the plugin ships no commands/ directory', () => {
+    expect(extractPluginCommands(pluginDir)).toEqual([]);
+  });
+
+  it('derives the command name from the filename, not frontmatter', () => {
+    writeCommand('deploy.md', '---\ndescription: Ship it\n---\n\nDeploy the app.\n');
+    const found = extractPluginCommands(pluginDir);
+    expect(found).toHaveLength(1);
+    expect(found[0]?.name).toBe('deploy');
+    expect(found[0]?.description).toBe('Ship it');
+    expect(found[0]?.body).toContain('Deploy the app.');
+  });
+
+  it('namespaces a subdirectory with a colon (CC parity)', () => {
+    writeCommand('review/security.md', '---\ndescription: Sec review\n---\n\nCheck auth.\n');
+    const found = extractPluginCommands(pluginDir);
+    expect(found[0]?.name).toBe('review:security');
+  });
+
+  it('namespaces nested subdirectories', () => {
+    writeCommand('a/b/c.md', '---\ndescription: Deep\n---\n\nBody.\n');
+    expect(extractPluginCommands(pluginDir)[0]?.name).toBe('a:b:c');
+  });
+
+  it('marks every discovered command with origin: command', () => {
+    writeCommand('one.md', '---\ndescription: One\n---\n\nBody.\n');
+    expect(extractPluginCommands(pluginDir)[0]?.origin).toBe('command');
+  });
+
+  it('ignores a stray frontmatter name: so the path stays authoritative', () => {
+    // CC addresses a command by its path; honouring a `name:` would make the
+    // file invocable under a name its own location does not predict.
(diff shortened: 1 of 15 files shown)


griffinwork40:
@codex fix comments

griffinwork40:
Wave 1.5 complete — all 13 citations resolved at the ref, both absence claims confirmed, and the top finding **reproduced by executing the code**. Synthesizing inline (I hold the reproduction evidence; disclosed below).

---

# Review — PR #965 `feat(plugins): load Claude Code commands/*.md as slash commands`

**Target:** griffinwork40/agent-afk#965 · `feat/plugin-commands-loader` → `main` · ref `b9579fab`
**Triage:** feature · 6 files · +384 / −7 · regime **full** · 2 commits (`b0de54cd`, `b9579fab`)

## Blocking

**F1 · medium · blocking:true · confidence:high · correctness · `src/agent/plugins/command-files.ts:107-111` + `src/agent/plugins/tool-injector.ts:283` · ref:b9579fab · citation-type:file-state ✅verified**
A CRLF- or BOM-authored command file has its entire YAML frontmatter block emitted as the dispatched prompt body and loses its description, because `parseSkillMetadata` bails on `!content.startsWith('---\n')` and the new fallback re-tests the *same* byte-exact prefix, so `---\r\n` and `\uFEFF---\n` both fall through into the "no frontmatter" branch.
```ts
if (!parsed.body) {
  const content = readFileSync(full, 'utf-8');
  if (!content.startsWith('---\n')) parsed.body = content.trim();
}
```
**Wave 1.5 executed `extractPluginCommands` at the ref against a 3-file fixture:**
```
name="bom"  description=undefined body="---\ndescription: Ship it\n---\n\nDeploy."
name="crlf" description=undefined body="---\r\ndescription: Ship it\r\n---\r\n\r\nDeploy."
name="lf"   description="Ship it" body="Deploy."
```
This is the finding that most directly defeats the PR's stated purpose: the command registers, is listed, dispatches — and sends the model raw YAML. BOM is a second trigger Wave 1 did not identify. On the rubric this reads as `high` (reachable in production, wrong output, and third-party Windows-authored marketplace files are the exact population this PR exists to load); reported at Wave 1's assigned `medium` — escalation is the reviewer's call, disposition is unchanged.
**Suggestion:** normalize before both checks — `const normalised = content.replace(/^\uFEFF/, '').replace(/\r\n/g, '\n')` — inside `parseSkillMetadata`, so SKILL.md gets the fix too.

**F-S1 · medium · blocking:true · confidence:high · security · `src/agent/plugins/command-files.ts:85` · ref:b9579fab · citation-type:file-state ✅verified**
`statSync` resolves symlinks, so a plugin shipping `commands/help.md → ~/.ssh/id_rsa` registers that file's contents as a dispatchable slash-command body.
```ts
stat = statSync(full);
```
**Wave 1.5 parity check:** the pre-existing SKILL.md walker uses the identical call at `tool-injector.ts:231` — this is parity with shipped behaviour, **not a hole introduced by this diff**. `install.ts:427` / `remove.ts:67` show the codebase does use `lstatSync` where symlinks matter. Security-dimension `medium` is non-waivable per policy, so it blocks despite the parity.
**Suggestion:** `lstatSync` + skip on `isSymbolicLink()`, in both walkers.

**F-S2 · medium · blocking:true · confidence:high · security · `src/agent/plugins/tool-injector.ts:366-373` · ref:b9579fab · citation-type:file-state ✅verified**
A third-party command file's `model:` frontmatter reaches `providerForModel()`/the credential resolver with no allowlist, and this PR widens that path to every `commands/*.md` in every installed plugin.
```ts
} else if (key === 'model') {
  const raw = value.replace(/^["']|["']$/g, '').trim();
  if (raw.length > 0) metadata.model = raw;
```
**Wave 1.5:** the cited lines are **pre-existing and untouched** — the diff's `tool-injector.ts` hunks are `@@ -21,6 +21,14 @@`, `@@ -265,7 +273,7 @@`, `@@ -301,7 +309,12 @@`; line 366 is in none of them. Identical exposure already exists for every installed SKILL.md. Non-waivable security medium.
**Suggestion:** validate `model` against the known-id set before assigning; warn and drop on miss.

**F-S3 · medium · blocking:true · confidence:high · security · `src/agent/plugins/command-files.ts:110` · ref:b9579fab · citation-type:file-state ✅verified**
The no-frontmatter branch assigns unbounded, unsanitized third-party file content directly as the prompt body, with no length cap.
```ts
if (!content.startsWith('---\n')) parsed.body = content.trim();
```
**Wave 1.5:** the branch is new; the injection surface is parity with the SKILL.md body path. Non-waivable security medium.
**Suggestion:** cap body size (e.g. 64 KB) in `parseSkillMetadata` so both paths inherit it.

**F-A2 · medium · blocking:true · confidence:high · api-compat · `src/cli/slash/preflight/types.ts:38` · ref:b9579fab · citation-type:file-state ✅verified — ⚠ CONFLICT**
`SkillInvocation.source` still enumerates five members while `SkillManifestEntry.source` now has six, so a future `source: e.source` assignment type-checks and silently matches no preflight branch.
```ts
source: 'builtin' | 'user' | 'project' | 'plugin' | 'imported';
```
**⚠ CONFLICT — unresolved, surfaced not adjudicated.** Wave 1.5 confirmed **zero runtime impact today**: `dispatch.ts:144` hardcodes `source: 'plugin'`, and no `SkillManifestEntry.source` value flows into a `SkillInvocation` anywhere (`grant-manager.ts:208`, `marketplace.ts:109`, `plugin.ts:165` are unrelated types). By the rubric that is `low` ("no production behaviour impact today"), and the same agent rated the structurally identical `listing.ts` trap `low` (F-A1 below). Disposition carried through unchanged per protocol; the tier is the reviewer's to settle.
**Suggestion:** add `'command'` to `SkillInvocation.source` documented as preflight-equivalent to `'plugin'`.

## Non-blocking

**F-A1 + F3 (deduped) · low · blocking:false · api-compat/spec-compliance · `src/cli/slash/plugin-skills/listing.ts:22,100,108-120` ✅verified** — commands appear in `/skills` but render as `Source: plugin`; `SkillSource` omits `'command'` and `friendlySource` has no `default` arm, so a future refactor that threads the real source returns `undefined` silently. `DiscoveredSkill` (`state.ts:11-16`) carries no `source` field at all, which is why the label is hardcoded. **Fix: add a `default:` arm — that alone converts the latent trap into a compile error.**

**F2 · low · blocking:false · correctness · `command-files.ts:97` ✅verified** — `commands/a:b.md` and `commands/a/b.md` both derive `a:b`; the `seen` guard drops one by sort order, silently.

**F4 · low · blocking:false · test-coverage ✅confirmed-absent** (`git grep '\r\n|CRLF'` in both new test files → 0 matches) — no CRLF/BOM fixture, which is why F1 shipped.

**F5 · low · blocking:false · test-coverage ✅confirmed-absent** — no test covers plugin A's `SKILL.md` shadowing plugin B's `commands/*.md`. The adjacent cross-plugin test at `skill-bridge.test.ts:645` covers skill-vs-skill only; all five new integration tests use a single `tmpDir`.

**F6 · low · blocking:false · test-coverage · `command-files.test.ts:49-55,113-120` ✅verified** — the malformed-file test asserts only `names).toContain('good')` with no `not.toContain('bad')`, and the plain-Markdown `toMatchObject` never asserts `description`. Both would pass through a regression.

**F7 · low · blocking:false · perf-observability · `command-files.ts:131` ✅verified** — `localeCompare` is locale/ICU-dependent; the doc comment claims deterministic ordering.

**F8 · low · blocking:false · confidence:low · perf-observability · `command-files.ts:80-91` ✅verified** `[low confidence — verify with runtime context]` — symlink cycles bounded only by `MAX_DEPTH=10`, no visited-set.

**W1.5-a · low · blocking:false · spec-compliance (orchestrator finding)** — PR body claim 6 asserts "`afk skill list` still sees the complete set", but no such command exists: no `src/cli/commands/skill*.ts`, no `command('skill'` registration at the ref. PR-body inaccuracy only.

**W1.5-b · low · blocking:false · spec-compliance (orchestrator finding) · `skill-bridge.ts:253-262` vs `main`** — undisclosed model-facing change: on `main` plugin entries carried no `argumentHint`; this PR adds `argumentHint: skill.argumentHint` and `buildSkillManifest:117` renders it, so every plugin skill's manifest line now grows — compounded by the kebab fix, which makes hints parse for the first time. Not named in the PR body, and mildly in tension with its token-budget rationale.

**W1.5-c · low · blocking:false · spec-compliance (orchestrator finding) · `skill-executor.ts:160`** — manifest exclusion hides the *menu*, not the *capability*: `collectSkillEntries` (unfiltered) and `discoverPluginSkillBodies` (no audience/source filter) both back the `skill` tool, so a model that names a command still invokes it. Achieves rationale #1 (tokens) fully and rationale #2 (behavioural) partially.

## Spec-compliance verdicts

| Claim | Verdict |
|---|---|
| **3** — "can never register without being dispatchable" | **True as written; converse false.** `collectSkillEntries:261` applies `isSkillVisible`, `discoverPluginSkillBodies` does not — and carries an explicit invariant comment saying so ("no audience filter — dispatch is always available; only surfacing is gated"). An audience-gated command is dispatchable but unlisted. Deliberate and documented; the PR body does not mention the direction. |
| **5** — "precedence comes free" | **Verified.** `seen` is shared across the concatenated array and pre-populated by all registry skills before the plugin loop, so first-wins holds within a plugin, across plugins, and against builtins. A plugin command **cannot** capture `/review` or `/ship`. Test covers the same-plugin case only (F5). |
| **6** — "`/skills` and `afk skill list` still see the complete set" | **Substantively true, imprecise in two places.** Commands do reach `/skills` via `supported-commands.ts:33` → `state.discovered`, and `buildSkillManifest:112` is the only model-facing consumer — no leak. But they render as `plugin` (F-A1/F3) and `afk skill list` does not exist (W1.5-a). |
| **7** — `argument-hint` dual-spelling | **Verified.** `user-skills.ts:126` reads both spellings; the fix matches. No unfixed sibling: `read-only`/`readOnly` already accepts both (`tool-injector.ts:359`), `context`/`model` have no kebab form. |

**Scope creep:** none substantive. W1.5-b is an undisclosed side-effect, not unrequested work.

## What was not checked

- **Citations verified inline against branch HEAD `b9579fab`** (= local worktree HEAD, clean apart from untracked `.afk/`). All 13 `file-state` citations `verified`; **0 fabricated, 0 diff-only, 0 false-absent — no findings dropped.**
- **Stated intent: PR #965 title + body — spec-compliance assessed.**
- **`/shadow-verify` not invoked.** Trigger conditions absent (0 critical, 0 high, 0 blocking departures from the default table). Wave 1.5 independently re-derived every citation with `git show`/`git grep` — tools Wave 1 lacked — and executed `extractPluginCommands` at the ref for F1. **F-A2's tier remains the one unresolved judgement**, surfaced as CONFLICT rather than adjudicated.
- **Wave 2 synthesis run inline rather than dispatched** — deliberate deviation: the reproduction evidence lives in this session and would be lossy to re-summarize through a shell-less agent.
- Tests were **not run**; the author's gate table is an unverified assertion. Coverage assessed by reading only.
- Not reviewed: Telegram surface rendering of `SkillManifestEntry`, the OpenAI-compatible provider's `supportedCommands` path beyond confirming it delegates to `collectSupportedCommands`, `plugin-skills-reload.test.ts` coverage of `source: 'command'`, and the interaction with #963's shadowing notices.
- `knownToolNames` is never threaded from `collectSkillEntries` into `extractPluginCommands` (lazy default always used) — confirmed benign, not deeply traced.

---

**Decision: DO NOT MERGE — 5 medium blocking (3 security, 1 correctness, 1 api-compat); 0 medium waived, 11 low, 0 nit.**

Practical read: **F1 is the one defect that is unambiguously introduced by this diff, reproducible, and user-visible** — a one-line normalization in `parseSkillMetadata` fixes it for both loaders and closes F4. The three security mediums are non-waivable by policy but are all parity-with-shipped-behaviour (F-S1, F-S2 cite untouched or pre-existing code paths); F-A2 has zero runtime impact today and conflicts with its own author's `low` rating of the identical `listing.ts` trap.

## Done
- Ran `/review` on PR #965 end to end: resolved the diff and intent from `gh`, dispatched 2 parallel Wave 1 dimension agents, ran Wave 1.5 citation + absence verification inline against ref `b9579fab`, and synthesized the merge decision.
- Evidence: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/98cc4b77-f800-4296-8876-c02e84a1f66d/pr-965.diff`; all citations re-resolved via `git show b9579fab:<file>`; F1 reproduced by executing `extractPluginCommands` at the ref through `/tmp/crlf-probe.mts` (3-file fixture, output quoted above); session trace at `~/.afk/state/witness/<session>/trace.jsonl`.
- What changed in the world: **nothing in the repo or on the PR.** Review is read-only — no edits, commits, pushes, or PR comments. The only file written was the throwaway probe at `/tmp/crlf-probe.mts` (outside the repo); the worktree remains dirty with only the pre-existing untracked `.afk/`, exactly as found.
- Pending/deferred: `/shadow-verify` not run (no critical/high, no blocking departures — Wave 1.5 covered the re-derivation with stronger tooling); F-A2's severity tier left as an unresolved CONFLICT for the author; tests not executed, so the author's gate table stands unverified.

---

griffinwork40 (commented):

**Decision: DO NOT MERGE — 0 critical, 0 high, 2 medium blocking (1 security, 1 api-compat); 0 medium waived, 4 low, 2 nit.**

Reviewed ref `b0125a4` (base `main`). Feature, 9 files, +550/−15 → full regime. All citations below were re-read against the reviewed ref; two Wave-1 findings were dropped as refuted (see manifest at the end).

---

## Blocking

### 1. `medium` · security · `src/agent/plugins/command-files.ts:115-121`

Third-party command bodies are read with no byte cap and no control-character handling, so a NUL byte, an overlong file, or adversarial Unicode in an installed plugin's `commands/*.md` lands verbatim in `parsed.body` and is dispatched as a prompt.

```ts
const content = normalizeSkillSource(readFileSync(full, 'utf-8'));
if (!content.startsWith('---\n')) parsed.body = content.trim();
```

`normalizeSkillSource` (`source-guard.ts:26-28`) does BOM-strip + CRLF→LF only — no cap, no NUL strip.

**Verified parity context (orchestrator re-read, not in the agent's report):** the SKILL.md path has no size cap either — `parseSkillMetadata` (`tool-injector.ts:279-292`) reads the file the same way. This is therefore a **pre-existing gap now extended to a second surface, not a regression introduced here**, and the command body reaches the model only when a user types the slash command (commands are correctly excluded from the model-facing manifest). Flagged as blocking because a `medium` in the security dimension is not waivable under this review's policy — a maintainer may reasonably land it and fix both loaders together.

**Suggestion:** cap body length (e.g. 64 KB) and reject/strip NUL before assigning `parsed.body`, in `source-guard.ts` so both loaders inherit it.

### 2. `medium` · api-compat · `src/cli/slash/plugin-skills/listing.ts:123`

Pass 2 of `buildListingGroups` hard-codes `source: 'plugin'` for every plugin-origin row, so a `commands/*.md` entry renders in `/skills` labelled **plugin**, and the newly added `command` legend member never appears at runtime.

```ts
addRow({ slashName, display, description: skill.description, source: 'plugin' });
```

`LEGEND_RANK.command` (listing.ts:40), `friendlySource`'s `case 'command'` (listing.ts:143-144), and the legend line at listing.ts:223 are consequently unreachable — the union member is threaded through the type but never produced.

**The reported fix does not work — corrected here.** The agent suggested propagating `skill.source` from `DiscoveredSkill`; that field does not exist. I traced the actual chain at the ref:

- `collectSkillEntries` correctly tags `source: 'command'` (`skill-bridge.ts`, this diff) ✅
- `collectSupportedCommands` (`providers/shared/supported-commands.ts:35-42`) maps to `ProviderCommandInfo` — **`source` is dropped here**; the interface is `{name, description?, argumentHint?}` (`provider.ts:444-448`)
- `dispatch.ts:215-219` builds `DiscoveredSkill[]`, which also has no `source` (`state.ts:11-16`)
- `listing.ts:123` therefore has nothing to propagate and hard-codes the literal

**Suggestion:** add `source` to `ProviderCommandInfo` and `DiscoveredSkill`, populate it at `dispatch.ts:215`, and use `skill.source` at `listing.ts:123`. Until then the `command` legend entry is dead code, and the PR-body claim that `/skills` distinguishes commands does not hold (names do list; the label reads "plugin").

---

## Non-blocking

- `low` · security · `source-guard.ts:40-50` — `resolveContained` calls `realpathSync(root)` once per file entry, so `root` is re-resolved throughout the walk; replacing `commands/` with a symlink mid-walk shifts the containment baseline. Hoist `realpathSync(root)` out of the loop into `extractPluginCommands` before `walk` and freeze it for the walk's lifetime.
- `low` · perf · `skill-bridge.ts` (both seams) — `extractPluginCommands(plugin.path)` is called without the optional `knownToolNames` argument (`command-files.ts:64-67`), forcing a lazy `resolveKnownToolNames()` per invocation instead of sharing the set already resolved for the sibling `extractPluginSkills` call.
- `low` · test-coverage · `command-files.test.ts` — no test covers `audience:` frontmatter on a command file (absence confirmed: zero matches for `audience` in the file at the ref), leaving the interaction between command discovery and the `isSkillVisible` gate in `collectSkillEntries` unexercised.
- `low` · observability · `command-files.ts:82,86,90,101,104,125` — six skip branches (dotfile, colon-in-segment, symlink escape, non-`.md`, empty basename, empty body) return silently, so a plugin whose commands all fail to register produces no diagnostic. A debug-gated stderr line per skip would make this diagnosable.
- `nit` · api-compat · `dispatch.ts:144` — `SkillInvocation` is built with a hard-coded `source: 'plugin'`, so the `'command'` member added to that union in `preflight/types.ts` is never emitted by the dispatch path.
- `nit` · correctness · `command-files.ts:141-144` — the `a.name ?? ''` / `b.name ?? ''` fallbacks in the sort comparator are unreachable; `name` is assigned unconditionally at :105 and pushed at :132.

---

## Verified as claimed

Spot-checked the PR body's load-bearing claims against the ref, all confirmed:

- **Manifest exclusion** — `buildSkillManifest` filters `e.source !== 'command'` after the `excludeName` filter, at the model-facing boundary as described.
- **Both seams patched** — `collectSkillEntries` and `discoverPluginSkillBodies` both use the identical `[...extractPluginSkills, ...extractPluginCommands]` order, so a command cannot register without being dispatchable.
- **Precedence** — SKILL.md is collected first and the existing `seen`/first-wins guard yields "skill beats command"; locked by tests at `skill-bridge.test.ts:594` and `:604` (including the cross-plugin case).
- **Namespacing** — `commands/review/security.md` → `review:security` (`command-files.test.ts:97`), nested at `:103`.
- **No shell execution** — nothing in the new path executes or shell-expands a body; the `` !`cmd` ``/`@file` exclusion holds in code.
- **Drive-by fix** — `parseSkillMetadata` now accepts both `argument-hint` and `argumentHint`.
- **Test counts undersold** — the body claims 11 + 5 tests; the ref has 16 in `command-files.test.ts` and 7 in the new `plugin commands/*.md integration` block.

## Dropped findings (refuted on re-read at the ref)

- *"Comment at `command-files.ts:112-114` contradicts post-patch behavior"* — **refuted**. The comment says the shared parser "deliberately rejects files without [frontmatter]"; `parseSkillMetadata` does exactly that, returning `{}` when `!content.startsWith('---\n')` (`tool-injector.ts:287-289`). The double `normalizeSkillSource` call is idempotent, as the reporting agent itself conceded.
- *"The colon test inverts skip semantics — `a:b.md` is the survivor"* — **refuted**. `command-files.ts:86` skips `a:b.md`; the surviving `a:b` comes from `a/b.md`, exactly as the test name states. (Residual observation, not counted as a finding: the assertion `toEqual(['a:b'])` would still pass if the skip guard were deleted, since both files derive the same name and first-wins keeps one — so the test does not actually lock the guard. Worth tightening, e.g. by asserting on a body that differs between the two files.)

## What was not checked

- Citations verified inline against PR head **`b0125a4`**. Both Wave-1 agents emitted diff-file line numbers rather than source-file line numbers; every citation above was re-anchored to its real location at the ref and its evidence snippet confirmed verbatim. No citation was fabricated.
- Stated intent: PR #965 title + body — **spec-compliance assessed** against it.
- Tests were not executed; CI is green on this head (Lint & Build, ubuntu/macOS, real-PTY scrollback, docs, publish smoke). Windows was `SKIPPED`, so the CRLF/BOM handling this PR adds is covered by unit tests only, not by a Windows CI run.
- No runtime/manual check of `/skills` rendering, `afk skill list`, or an end-to-end install of a real Anthropic-marketplace plugin — finding 2 was derived by tracing the type chain statically.
- `/shadow-verify` was not dispatched: no `critical`/`high` finding and no finding departs from the default blocking table, so per policy `medium`-and-below at default disposition pass through. Both blocking findings were independently re-derived inline during verification instead.

griffinwork40 (commented):

PR #965 adds a loader for Claude Code `commands/*.md` files inside plugins, surfacing them as dispatchable slash commands alongside the existing `SKILL.md` path. It introduces `command-files.ts` (a depth-bounded, symlink-guarded directory walk), widens `SkillManifestEntry['source']` with `'command'`, and threads that value through every exhaustive consumer — manifest building, reload counters, provider types, listing display — while excluding commands from the model-facing manifest and deliberately stopping short of `` !`cmd` `` / `@file` inlining. The design is careful and most of the security surface it opens is closed on purpose: symlink containment is real, the 64KB cap is real, no shell execution is reintroduced. One hole is open, and it comes directly from the feature's central design choice — that a command's name comes from its **file path** rather than its frontmatter.

## Findings

**F0 — terminal-escape injection via a path-derived command name** · high · blocking: true · security
`src/agent/plugins/command-files.ts:146-153`, `:178-185`

A `commands/*.md` filename (or an intermediate subdirectory) containing a raw ESC byte becomes the command's registered `name` verbatim, and nothing between there and the terminal strips it. `normalizeSkillSource` does not protect this vector — it normalizes file **contents**, and the name never passes through it.

```ts
const base = entry.slice(0, -'.md'.length);
...
const name = [...segments, base].join(':');
```

Traced end to end at this ref:
- `deploy\x1b[2J\x1b[H.md` survives every skip branch in `walk()` — not a dotfile (`:109`), no colon (`:116`), ends `.md` (`:141`), non-empty basename (`:147`) — yielding `name = 'deploy\x1b[2J\x1b[H'`.
- The name is pushed unsanitized at `:178-185`. `sanitizeForDisplay` is called 7× in this file, but every call is on `join(dir, entry)` / `full` for `AFK_DEBUG` stderr diagnostics — never on `name`.
- `collectSkillEntries` (`skill-bridge.ts:266-276`) gates only on `!skill.name` / `seen` / visibility — no charset validation.
- Nothing downstream strips: `listing.ts` and `dispatch.ts` both contain zero `sanitizeForDisplay` / `stripAnsi` calls, and `display` is written straight out at `listing.ts:171-179` via `ctx.out.line('  ' + palette.warning(main.display))`.
- Reachable **without** the user running `/skills`: `getPluginShadowingNoticeLines()` (`dispatch.ts:280-287`) interpolates `c.altSlash`, built from the name at `:255`, into a notice printed at the next prompt when the name collides.

This is introduced by this PR rather than inherited — `command-files.ts` does not exist on `main`, and on `main` every plugin skill name comes from frontmatter, i.e. from already-normalized content. Path-derived naming is the new behavior, and it is the vector.

*Suggestion:* Apply the already-imported `sanitizeForDisplay` (`utils/terminal-sanitize.ts:63-65`, which strips escapes and neutralizes control bytes) to `name` at `command-files.ts:153` before the `out.push`, or reject any path segment containing a control byte at the same place the `:` check rejects ambiguous ones (`:116`).

**F1 — the symlink-escape guard has no test anywhere in the repo** · medium · blocking: false (waived) · test-coverage
`src/agent/plugins/command-files.test.ts` (whole file) + guard call site `src/agent/plugins/command-files.ts:125`

`resolveContained` — the control stopping a plugin shipping `commands/help.md -> ~/.ssh/id_rsa` from becoming a dispatchable prompt — has zero direct coverage.

```ts
if (resolveContained(root, full, realRoot) === undefined) {
```

`grep -rn "resolveContained" --include=*.test.ts src/` returns 0 matches repo-wide, and no `source-guard.test.ts` exists. The accept-path is exercised incidentally by every `extractPluginCommands()` test, but the security-relevant **rejection** branch (`source-guard.ts:90`) and the error-catch (`:86-88`) are never asserted. Adversarial re-derivation found the logic itself sound — prefix-sibling (`/a/commands` vs `/a/commands-evil/`) correctly rejected, broken symlinks fail closed, `preResolvedRoot` pinned once before the walk — which is why this is waived rather than blocking.

*Waived:* bounded tests-only gap on containment logic independently verified correct; landable with follow-up.
*Suggestion:* Add a test that `symlinkSync`s a file from outside the plugin tree into `commands/` and asserts `extractPluginCommands` excludes it.

**F2 — "registered ⇒ dispatchable" rests on two copies of one invariant** · low · blocking: false · correctness
`src/agent/tools/skill-bridge.ts:263-264` and `:353-354`

The invariant this PR asserts depends on two independently-maintained copies of the same ordered concat, with no shared helper, so an edit to one seam and not the other silently breaks precedence or dispatchability.

```ts
...extractPluginSkills(plugin.path, knownToolNames),
...extractPluginCommands(plugin.path, knownToolNames),
```

*Suggestion:* Extract one `discoverPluginSkillsAndCommands(pluginPath, knownToolNames)` helper and call it from both `collectSkillEntries` and `discoverPluginSkillBodies`.

**F3 — discovery walk is uncached and breadth-unbounded** · low · blocking: false · perf-observability
`src/agent/plugins/command-files.ts:77-96`

`extractPluginCommands` runs an uncached recursive `readdirSync`/`statSync` walk on every invocation, from both `collectSkillEntries` and `discoverPluginSkillBodies`. Depth is capped; total file count under `commands/` is not.

```ts
const MAX_DEPTH = 10;
    if (depth > MAX_DEPTH) return;
```

The pre-existing `SKILL.md` walker (`tool-injector.ts:216`) has the identical depth-only, uncached shape — this PR doubles an existing cost rather than inventing a new pattern.

*Suggestion:* Memoize per `pluginPath` behind the existing plugin-scan cache invalidation hook, and add a total-entry cap alongside `MAX_DEPTH`.

**F4 — PR body's test count understates what landed** · nit · blocking: false · spec-compliance
`src/agent/plugins/command-files.test.ts:32` and `:172`

The body states "11 tests" in `command-files.test.ts`; the file ships 16 `it()` blocks under `describe('extractPluginCommands')` plus 7 under `describe('normalizeSkillSource')`.

*Suggestion:* Update the count, or describe it as distinct behaviors rather than test cases.

## Verified clean

- Symlink containment is implemented and its logic independently re-derived as correct (`source-guard.ts:76-91`), with the realpath root resolved once before the walk (`command-files.ts:91-96`), closing a mid-walk symlink-swap race.
- Content is capped at 64KB with truncate-not-throw semantics (`source-guard.ts:19,51-53`), and file **contents** are control-stripped before parsing (`:50`) — frontmatter `description` / `argumentHint` and both body paths are covered.
- No shell execution is reintroduced: `command-files.ts` imports only `fs` and `path`. The deliberate exclusion of `` !`cmd` `` / `@file` holds.
- Manifest exclusion of commands is real and tested (`skill-bridge.ts:112`; `skill-bridge.test.ts:571` asserts `expect(manifest).not.toContain('deploy')`).
- `SKILL.md`-beats-command precedence is real and tested, including the cross-plugin case (`skill-bridge.test.ts:594`, `:604`).
- The `'command'` union widening propagated to every exhaustive consumer the compiler requires — `sdk-types.ts:343`, `provider.ts:454`, `reload.ts:28-44`, `listing.ts`'s `friendlySource` switch. api-compat is additive; no production importer breaks.
- The drive-by `argument-hint` / `argumentHint` fix is real (`tool-injector.ts:316`) and matches the behavior `user-skills.ts` already had.

## Note on F0 vs. the existing escape test

`command-files.test.ts:193-221` asserts no ESC reaches stderr, and it passes — but its fixture is `'evil\x1b[2J\x1b[H.txt'`, a **`.txt`** file, which is rejected at the `.endsWith('.md')` filter (`:141`). It therefore only exercises the `AFK_DEBUG` skip-diagnostic path and never the name-registration path F0 describes. A `.md` fixture with the same name would exercise the real vector.

An automated first pass raised a related but *different* claim — that a plugin `description:` reaches the terminal unsanitized — and inline re-derivation **refuted** it: `source-guard.ts:50` strips `[\x00-\x08\x0B-\x1F]` (ESC `0x1B` is in range) and `parseSkillMetadata` reads through `normalizeSkillSource` before extracting `description`. That claim is dropped. F0 survives because the **name** never passes through that function at all.

## What was not checked

- Citations were verified inline against PR head `1ccb441e`; findings whose cited lines did not exist at that ref were dropped rather than reported.
- Both Wave 1 dimension agents returned partial results from wall-clock budget exhaustion, so this pass is not exhaustive.
- No tests were executed — the PR body's gate table (lint, `test:coverage` 15513 passed, audit gates) is taken as reported, not reproduced. F0 is a static-trace finding; a runtime PoC was not built.
- The SDK subprocess boundary's handling of the new `'command'` source value was not verified.
- Stated intent: PR #965 title + body — spec-compliance assessed.

Decision: DO NOT MERGE — 1 high blocking (security); 1 medium waived (test-coverage), 2 low, 1 nit non-blocking.

griffinwork40:
## Finding disposition — every round, explicit

Pushed `299c6664`. This comment states a disposition for **every** finding raised across all four review rounds, because no round so far has said what happened to the previous round's findings — round 4 silently dropped round 3's two blockers rather than marking them resolved, which makes progress invisible and the verdict permanently red.

### Round 3 (2026-08-09T22:58, reviewed `b0125a42`) — both blockers were already fixed before round 4 ran

| Finding | Severity | Status |
|---|---|---|
| No byte cap / NUL handling on command body | medium · security · blocking | **FIXED** in `00875126` — `source-guard.ts:19` adds `MAX_SKILL_SOURCE_CHARS = 64*1024`; `:47-53` adds the C0 strip + truncation inside `normalizeSkillSource`, which `command-files.ts` calls on the plain-body path |
| `listing.ts:123` hardcoded `source: 'plugin'`, `command` legend unreachable | medium · api-compat · blocking | **FIXED** in `00875126` / `1ccb441e` — chain threaded end to end: `provider.ts:444`, `supported-commands.ts:41`, `state.ts:18`, `dispatch.ts:219`, and `listing.ts:123` now reads `skill.source ?? 'plugin'` |

### Round 4 (2026-08-10T06:23, reviewed `1ccb441e`)

| Finding | Severity | Status |
|---|---|---|
| **F0** — terminal-escape injection via path-derived command name | high · security · blocking | **FIXED** in `299c6664` |
| **F1** — symlink-escape guard has no test anywhere in the repo | medium · test-coverage · waived | **FIXED** in `299c6664` — added the first assertion of the `resolveContained` reject branch |
| **F2** — "registered ⇒ dispatchable" rests on two copies of one concat | low · correctness | **DEFERRED** → follow-up issue. Real, but extracting the shared helper touches both `skill-bridge` seams; not worth widening this diff |
| **F3** — discovery walk uncached and breadth-unbounded | low · perf-observability | **DEFERRED** → follow-up issue. As the review notes, this is the identical shape as the pre-existing `SKILL.md` walker (`tool-injector.ts:216`); memoize both together, not one here |
| **F4** — PR body understates the test count | nit | **FIXED** — body now reads 27 (20 `extractPluginCommands` + 7 `normalizeSkillSource`) |

### On F0 — fixed as a class, not an instance

The review's suggested fix was to apply `sanitizeForDisplay` to `name`. I took the other option it offered — **reject** control bytes where the `:` check already rejects ambiguous segments — because that is the single point at which a path segment enters a name, covering both routes (`segments` for a directory, `base` for the filename). Sanitizing at render time would need every current *and future* display site to remember; rejecting at the source cannot be reintroduced downstream.

Rejected range is C0 + DEL + C1 (`\u0000-\u001F`, `\u007F-\u009F`). C1 is included deliberately: U+009B is CSI in 8-bit form and acts exactly as `ESC-[` does on a terminal in 8-bit mode, so a C0-only guard would leave a working vector open. Names above U+00A0 are unaffected — non-ASCII command filenames still work.

Worth recording why this survived three earlier rounds and green CI: the pre-existing escape-sequence test uses a **`.txt`** fixture, which `entry.endsWith('.md')` rejects *before* a name is derived, and it asserts only that the `AFK_DEBUG` diagnostic is sanitized. **That test could not fail.** The three new control-byte tests were mutation-checked — with the guard neutered, each one fails — so they are load-bearing rather than decorative.

Also worth recording: the vector was present in the first commit (`b0de54cd`, then line 97) and byte-identical at `1ccb441e` (line 153). The intervening fix commits added `sanitizeForDisplay` in six places, every one an `AFK_DEBUG` stderr diagnostic. The loop was sanitizing everything adjacent to the defect without touching the defect.

### Gates

`pnpm lint`, 328 tests across `src/agent/plugins` + `skill-bridge`, `audit:env:check`, `scan:env:check`, `audit:sdk:check`, `audit:chalk:check`, `fix:pins:check` — all pass.

### Requesting a merge decision rather than a fifth round

Four review rounds and six fix commits in under 24 hours, with the verdict never flipping. All blocking findings from every round are now resolved; what remains is two `low` items that the review itself identifies as pre-existing patterns shared with the `SKILL.md` loader. Those belong in follow-up issues, not in this diff.

PR #966, opened 2026-08-09
feat(web): slash-command autocomplete + highlighting in the browser composer

griffinwork40:
## Summary

Typing `/` in the `afk web` composer now opens a filtered, keyboard-navigable command menu (↑/↓ navigate, Tab/Enter accept, Esc close), and slash tokens are highlighted live as you type — the affordances the terminal REPL already has.

**Ranking is shared, not duplicated.** The prefix → subsequence → recency → alphabetical core is lifted out of `filterSlashCandidates` into `src/cli/input/slash-match.ts`, a dependency-free module esbuild can bundle for the browser, so both surfaces order candidates identically by construction. `src/cli/input/trigger.ts` keeps its exact signature and delegates to it; its 45 existing tests pass **unmodified**, which is the regression gate for the extraction.

`GET /api/commands` serves the universe the browser ranks against.

### Two constraints worth reviewer attention

- **The slash registry is empty in the `afk web` process.** It is a process-global singleton and `registerAll()` runs only from the REPL bootstrap, so this route registers it itself — lazily on first request, memoized, inside a `try/catch` that degrades a malformed skills dir to an empty list rather than a crashed server. That is safe only because `afk web` owns its process; a future in-process REPL+web launch would need a per-surface registry, and the call site says so.
- **The slash layer is reached by a dynamic import, and that is structural.** A static import pulls the whole command tree — and through it `MemoryStore` and the provider runtime, which touch disk on construction — into the import graph of anything loading `routes.ts`, including `src/cli/index.ts` via the `web` command. Caught by 3 failures in `src/cli/index.test.ts`; without the fix, `afk --version` would build a `MemoryStore`.

### What this deliberately does not claim

Nothing dispatches slash commands on this surface — the prompt path sends text to the model verbatim. So the menu makes **no positive claim** that any command works. Its one claim is negative and verifiable: `REPL_ONLY` badges the 14 commands driving terminal-only machinery (screen, compositor, TTY, process lifetime), which could not act here even in principle.

Server-side execution is intentionally out of scope. It needs its own design: an allowlist (a denylist cannot close over runtime-registered plugin skills), a per-surface registry instance, a new additive `notice` ledger record kind, and a web `Writer`/`UiSurface`.

Known v1 limitation: plugin skills require a live session, so a web-only process lists builtins plus user/project skills. A live server returned 107 commands.

## How was this tested?

- `pnpm lint` (tsc + tsconfig.web) — passes
- `pnpm build` + web-ui bundle — passes; bundle verified to contain **zero node builtins**, which is the browser-safety invariant for the shared matcher
- Full suite — **15,885 passed / 834 files**. One failure, `src/agent/trace/writer.test.ts` "seals a real crashed subprocess", is pre-existing and environmental: it spawns `node_modules/.bin/tsx`, absent in my worktree because dependencies were not installed there. Verified failing identically on a stashed baseline before my changes.
- All six audit gates pass: env-access, env-registry, chalk, hash-pins, sdk-dependency
- **65 new tests**: `slash-match.test.ts` (15, ranking parity), `slash-autocomplete.test.ts` (21, trigger/keyboard/collision), `slash-highlight.test.ts` (18, character-for-character mirror fidelity), `server.test.ts` (+2, endpoint and auth)
- Live end-to-end: started a real server, `GET /api/commands` returned 200 with 107 commands and 401 without a token

Two behaviours are pinned deliberately because they are easy to regress:

- **Plain Enter still inserts a newline.** The menu claims Enter only while open with a candidate selected, via `stopImmediatePropagation()`, and the trigger closes the menu the moment a space is typed. Tested against a stand-in submit handler attached after the menu, exactly as `app.ts` orders them.
- **The composer survives a missing dropdown.** Mount points are discovered leniently, so a browser holding a cached older `index.html` against a newer `app.js` loses the dropdown rather than the whole composer.

## Checklist

- [x] `pnpm lint` passes
- [x] `pnpm test` passes (one pre-existing environmental failure, detailed above)


griffinwork40 (commented):

## Decision: DO NOT MERGE — 2 medium blocking (1 test-coverage, 1 perf-observability); 4 low.

Both blockers are small and localized (one log line, one test). Neither is a security or data-integrity issue, and the feature itself verified clean on the dimensions that usually sink a browser-facing change: the endpoint is authenticated, the DOM path is injection-safe, and the shared-ranking extraction preserves the REPL contract. Reviewed at ref `7f1dfb63`.

---

### Blocking

**1. `medium` · blocking:true · perf-observability · `src/web-server/routes.ts:138-152` · citation-type:diff-context**

A first-request failure permanently memoizes an empty command universe for the life of the process, with no log line anywhere.

```js
    } catch {
      commandUniverse = [];
    }
  }
```

Because `commandUniverse` is non-`null` afterward, every later request — all users, indefinitely — silently receives an empty autocomplete list with no retry path and no operator-visible signal. This is reachable under exactly the "malformed skills dir" scenario the PR body names as the anticipated degradation case. Grepped `routes.ts` at the ref for `console.(error|warn|log)|logger.`: zero matches, so nothing surfaces the swallowed error.

*Suggestion:* log the caught error (debug level is enough) so a cold-start misconfiguration is diagnosable, and/or leave `commandUniverse` as `null` on failure so a later request can retry.

**2. `medium` · blocking:true · test-coverage · `src/web-server/frontend/composer-wiring.ts:39-48` · citation-type:diff-context**

`wireComposerAffordances` — the function implementing the PR's stated "composer survives a missing dropdown" guarantee — has no test exercising it.

```js
export function wireComposerAffordances(
  deps: ComposerAffordanceDeps,
): SlashAutocomplete | undefined {
  const menu = document.getElementById(MENU_ID);
```

The lenient lookup itself is present and correct at the ref (`if (!menu || !mirror) return undefined;`). Absence verified: `git grep wireComposerAffordances 7f1dfb63 -- src/**` returns only `app.ts:13`, `app.ts:318`, and its own definition — no test file. `slash-autocomplete.test.ts` drives `SlashAutocomplete` and `mountSlashHighlight` against hand-built DOM, but nothing calls the wiring function against a DOM missing `#slash-menu` / `#prompt-mirror`.

*Suggestion:* add `composer-wiring.test.ts` asserting `wireComposerAffordances` returns `undefined` rather than throwing when either mount point is absent.

> **Reviewer note on this one.** This is a rubric-boundary call, surfaced rather than silently resolved. The severity rubric places a missing test at `low` absent an impact claim, and the implementation here is verified correct at the ref — so the risk is future regression, not present breakage. It was assigned `medium` by the dimension agent and synthesis does not re-adjudicate disposition, so it is reported as blocking. A maintainer waiving this one and landing on finding 1 alone would be a defensible read.

---

### Non-blocking

**3. `low` · blocking:false · correctness/security (found independently by both dimension agents) · `src/web-server/routes.ts:138-152`**

Concurrent first requests can each pass the `commandUniverse === null` check before the first `await` resolves, so each re-runs `registerAll()` + the skills-dir walk.

```js
export async function handleCommands(res: ServerResponse): Promise<void> {
  if (commandUniverse === null) {
    try {
      const [{ registerAll }, { buildSlashUniverse }] = await Promise.all([
```

Redundant work only, not corruption — `registerAll()` and `buildSlashUniverse()` are synchronous, so the outcome converges. Client-side `ensureCommands()` already collapses same-tab keystrokes onto one request, leaving multi-tab / scripted cold starts as the reachable path.

*Suggestion:* memoize the in-flight promise the way the frontend already does (`loading ??= (async () => {…})()`).

**4. `low` · blocking:false · api-compat · `src/cli/input/trigger.ts:105-116`**

The `hint` inclusion test changed from truthy to strict-`undefined`, so a command with `hint: ''` would now serialize as `hint: ''` instead of being omitted.

```js
-    ...(cmd.hint ? { hint: cmd.hint } : {}),
+      ...(cmd.hint !== undefined ? { hint: cmd.hint } : {}),
```

Verified latent at the ref: no command registers an empty-string hint (`hint?: string` at `src/cli/slash/types.ts:318`), so no current caller observes the divergence.

*Suggestion:* keep the truthy check in `buildSlashUniverse`, or normalize empty-string hints to `undefined` at the source.

**5. `low` · blocking:false · test-coverage · `src/web-server/routes.ts:145-147`** — no test forces the `catch` branch. Both new `server.test.ts` cases hit the success path (200 with a real list, 401 unauthenticated); neither induces a `registerAll()` / dynamic-import failure. This is the same branch as blocking finding 1.

**6. `low` · blocking:false · perf-observability · `src/web-server/frontend/slash-autocomplete.ts:1203-1206`** — the frontend fetch-failure path swallows identically, with no console signal. Session-scoped and matches the PR's stated "degrade to no autocomplete" design, so noted rather than pressed.

---

### Verified clean

- **Endpoint auth.** `/api/commands` is dispatched inside `dispatch()`, which gates every `/api/*` path behind `tokensMatch(auth.token, bearerFromHeader(...))` at `server.ts:257`, before the route match at `server.ts:319`. Confirmed at the ref by line ordering, and by the new `server.test.ts` case asserting 401 without an `Authorization` header.
- **XSS.** Row, menu, and mirror construction use `createElement` + `textContent` exclusively. Both `innerHTML` occurrences in the new frontend files are comments asserting the invariant, not calls. `paintMirror` clears with `mirror.textContent = ''` and appends text nodes. Adversarial tests cover `` for both the summary field and raw textarea input.
- **Information disclosure.** The `/api/commands` payload is restricted to `{ name, summary?, hint? }`, built from registry display fields — no filesystem paths or skill file bodies.
- **Prototype pollution.** No request body is parsed (`handleCommands(res)` takes no `req`), and the universe is built from object literals and `.map()`.
- **API compat.** `filterSlashCandidates(query, recentHistory)` is unchanged context in the diff — signature preserved, body delegates to `matchSlashCandidates`. Both production callers (`terminal-compositor.autocomplete.ts:107`, `input/reader.ts:175`) use the pre-existing shape. `isSubsequence` was module-private before relocation, so exporting it breaks nothing.
- **Spec compliance.** `REPL_ONLY` contains exactly 14 entries as claimed. `trigger.test.ts` is untouched by the diff (0 occurrences) and contains exactly 45 tests at the ref, substantiating "45 existing tests pass unmodified" structurally. Plain Enter is preserved by the `if (!this.isOpen()) return;` early exit before any `preventDefault()`. No scope creep found — all 17 files are coherent with the stated feature.
- **Per-keystroke cost.** Bounded: one filter+filter+sort pass over the ~107-command universe, capped at `MAX_SLASH_MATCHES = 20` (`slash-match.ts:22,109`). Highlight tokenization scales with typed-prompt length, not registry size.

### What was not checked

- Citations verified inline against PR head `7f1dfb633016da9885a22940b53527b0ff1ecaaf`, fetched locally. Every blocking finding and every `file-state` citation was re-read at that ref; both absence claims were re-verified with `git grep` at the ref. No citation was classified fabricated or false-absent, so nothing was dropped.
- Stated intent: PR #966 title + body — spec-compliance assessed against it.
- Test-execution claims ("15,885 passed / 834 files", `pnpm lint`, the zero-node-builtins bundle assertion) are not observable from a static diff and were not re-run. CI is green on all 8 required checks at this head.
- No runtime/browser verification: the keyboard-collision ordering, caret math, and mirror alignment were reviewed as source plus their committed tests, not exercised in a live browser.
- `/shadow-verify` was not triggered — no `critical` or `high` findings, and no `blocking` value departs from the default severity table.
- The correctness/spec/test/perf dimension agent hit its tool-iteration cap; it covered all four dimensions, but its sweep was budget-bounded rather than exhaustive.

griffinwork40 (commented):

## Prior Round Dispositions (round 1 → round 2)

The first automated round (at ref `7f1dfb63`) raised 2 medium blocking and 4 low findings. All six are resolved:

| Prior ID | Severity | Status | Detail |
|----------|----------|--------|--------|
| F1 | medium | ✅ FIXED (`3e0fa090`) | Silent error memoization in `handleCommands` catch → replaced with 503, `console.error`, and retry logic. Server test "surfaces command initialization failure and retries successfully" verifies. |
| F2 | medium | ✅ FIXED (`3e0fa090`) | `wireComposerAffordances` had no test → `composer-wiring.test.ts` added with 3 tests (missing menu, missing mirror, successful mount). |
| F3 | low | ✅ FIXED (`3e0fa090`) | Concurrent first requests re-ran `registerAll()` → shared `commandUniverseInitialization` promise deduplicates. Server test "shares one in-flight initialization" verifies. |
| F4 | low | ✅ FIXED (`3e0fa090`) | `hint` truthy-vs-undefined check divergence → `buildSlashUniverse` uses truthy checks; new test asserts `hint: ''` is omitted. |
| F5 | low | ✅ FIXED (`3e0fa090`) | No test forced the catch branch → failure-path server test added. |
| F6 | low | ✅ FIXED (`3e0fa090`) | Frontend fetch-failure path swallowed silently → now logs to `console.error` and retries. |

---

## Decision: MERGE (with follow-ups) — 0 critical, 0 high; 3 medium advisory, 5 low, 2 nit.

All prior-round findings are resolved. The new findings are medium-and-below — none reach the critical/high threshold required to block. The three medium findings are advisory follow-ups, not merge blockers.

---

## New Findings (round 2, ref `3e0fa090`)

### Advisory (medium — non-blocking follow-ups)

**1. `medium` · advisory · correctness · `src/web-server/frontend/slash-autocomplete.ts:231–248`**

Concurrent failure race in `ensureCommands`: if two `refresh()` calls overlap during a load failure, the second caller shares the doomed promise and degrades to a closed menu with no logged diagnostic for that caller. Narrow: requires two concurrent keystrokes during the ~ms window of an active rejected fetch. Visible symptom is "menu stays closed" — no data loss, no crash. The retry path works on the next keystroke.

*Follow-up suggestion:* Add a comment documenting the concurrent-failure degradation is intentional, or add a test asserting two concurrent refreshes during a failing load both degrade without throwing.

**2. `medium` · advisory · correctness · `src/web-server/routes.ts:160–187`**

If `commandUniverseLoader()` throws **synchronously** (not returning a rejected Promise), the `??=` assignment captures the thrown exception as a rejected promise, but `.finally` never clears `commandUniverseInitialization` because the chain was never entered. The endpoint would return 503 permanently without retrying. Production-safe: `defaultCommandUniverseLoader` is `async`, so it cannot throw synchronously. The risk exists only via `setCommandUniverseLoaderForTests` — a test-only seam.

*Follow-up suggestion:* Wrap the loader: `commandUniverseInitialization ??= Promise.resolve().then(() => commandUniverseLoader()).then(…)`, or document the `CommandUniverseLoader` type constraint: "must return a Promise; synchronous throws are not handled."

**3. `medium` · advisory · test-coverage · `src/web-server/frontend/composer-wiring.test.ts`**

`wireComposerAffordances` has no test for a rejecting `loadCommands` after a successful mount. The downstream `slash-autocomplete.test.ts` covers the retry path at the unit level, so the gap is at the integration seam only.

*Follow-up suggestion:* Add a fourth test: pass a rejecting `loadCommands`, confirm `wireComposerAffordances` still returns a defined `SlashAutocomplete`, and confirm a subsequent `refresh()` after a resolving retry opens the menu.

---

### Low / Nit (informational)

**4. `low` · correctness · `src/cli/input/trigger.ts:42`** — The REPL's `detectTrigger` regex (`/^\/[A-Za-z_-]*$/`) excludes `:`, so namespaced commands like `/bgsub:status` don't trigger REPL autocomplete. The web trigger regex (`/^\/(?:[A-Za-z][\w:-]*)?$/`) includes `:`. The two surfaces now diverge on namespaced commands. Consider updating the REPL regex to match.

**5. `low` · correctness · `src/web-server/frontend/slash-autocomplete.ts:35–50`** — `/transcript` (uses `compositor.suspendInput()` / `resumeInput()` — TTY-suspend machinery) is absent from `REPL_ONLY`. The invariant says "keep entries to ones whose REPL-only nature is evident from what they manipulate"; `/transcript` satisfies that criterion.

**6. `low` · test-coverage · `src/web-server/frontend/slash-autocomplete.test.ts`** — The `scrollIntoView` positive invocation path is never exercised (jsdom doesn't implement it). Only the no-throw safety path is covered. Note: the fix commit *does* add scrollIntoView stub tests (lines 313–330) that verify the call happens — confirming this was already addressed.

**7. `low` · test-coverage · `src/cli/input/trigger.test.ts`** — `buildSlashUniverse` test covers only the empty-hint omission. Does not cover alias hint-inheritance or orphaned-alias edge cases.

**8. `low` · perf-observability · `src/web-server/frontend/slash-autocomplete.ts`** — No per-keystroke debounce on `refresh()`. Acceptable at current universe size (~107 commands); worth revisiting if plugin skill registration pushes past ~200.

**9. `nit` · api-compat · `src/cli/input/slash-match.ts:57`** — `isSubsequence` promoted from private to exported. Additive surface expansion with zero prior external importers.

**10. `nit` · spec-compliance · `src/web-server/frontend/composer-wiring.ts:51`** — `preload()` fires eagerly on page load (before the user types `/`). PR documents this as a deliberate warmup tradeoff.

---

### Verified Clean

- **Endpoint auth:** `/api/commands` is gated behind bearer-token check at `server.ts:256–258` before dispatch. New server test asserts 401 without token.
- **XSS:** All DOM rendering uses `createElement` + `textContent`/`createTextNode` exclusively. Zero `innerHTML` usage. Adversarial tests cover `` injection.
- **Information disclosure:** Payload carries `name`, `summary`, `hint` only — no file paths, skill bodies, or secrets.
- **Prototype pollution:** `handleCommands` accepts no request body (GET-only).
- **API compat:** `filterSlashCandidates` signature preserved. `Candidate` / `SlashCandidate` shapes are structurally identical. `buildSlashUniverse` is a new additive export.
- **Spec compliance:** All stated-intent elements delivered. No unmet requirements. No scope creep.
- **Per-keystroke cost:** Bounded — one filter+sort pass over ~107 commands, capped at `MAX_SLASH_MATCHES = 20`.

### What Was Not Checked

- Citations verified at ref `3e0fa0903ec435ed92afc721190040737914c322` via `gh api` file reads. All blocking/medium/file-state citations were re-read at the ref.
- Stated intent: PR #966 title + body — spec-compliance assessed against it.
- Test-execution claims (15,885 passed, pnpm lint) were not re-run. CI is green on all required checks at this head.
- No runtime/browser verification of keyboard collision, caret math, or mirror alignment — reviewed as source plus committed tests only.

Issue #948, opened 2026-08-08
capture: AFK_CAPTURE_SUBAGENT_OUTPUT records tool args through the UI summarizer, so bash commands truncate at 161 chars — the prompt sibling writes verbatim


griffinwork40:
`AFK_CAPTURE_SUBAGENT_OUTPUT=1` (#935, v5.100.0) is the only durable record of what a subagent actually did. It earned its keep immediately — it is how the ~200s `find /` burn in a forge run was root-caused. But it records **tool arguments through the UI summarizer**, so the arguments that matter most are the ones it destroys.

## Observed

A `bash` call in a captured subagent output lands as a single newline-collapsed line cut at ~161 characters. For the `find /` case the interesting part of the command was past the cut, so the capture could confirm *that* bash was called but not *what it ran*.

## Mechanism (three hops, the middle one deliberate)

1. `src/agent/session/subagent-output-capture.ts:230` records `chunk.toolInputRaw ?? chunk.toolInput`.
2. `extractRawToolInput` (`src/agent/facets/raw-input.ts:37,48-61`) whitelists only `file_path` / `name` / `id_prefix`. **`command` is deliberately excluded** for secret-risk reasons — see the rationale at `raw-input.ts:12-21`. So for every `bash` call, `toolInputRaw` is absent.
3. The fallback `chunk.toolInput` is a **UI label**, not data: `summarizeToolInput` in `src/agent/providers/shared/tool-input-summary.ts` applies `COMMAND_SUMMARY_MAX = 160` (line 29), collapses newlines and runs of whitespace (line 76), then emits `' ' + flat.slice(0, 159) + '…'` (lines 77-82) — 160 chars plus a leading space, which is the 161 observed.

Output-capture's own `MAX_RECORD_BYTES = 4096` (`subagent-output-capture.ts:44`) never engages; 161 chars is nowhere near it. The cap is inherited from a rendering helper shared with telegram, the session ledger, and the tool lane.

## Why this is a defect and not a preference

The sibling writer is faithful. `src/agent/session/subagent-prompt-capture.ts` records `input.prompt` directly, bounded only by its own `truncateToBytes(..., 8192)` (lines 39, 85-93) — no collapse, no 160-char cap. So the **prompt** side of the witness tree is near-verbatim while the **output** side is a UI label.

[...]

PR #968, opened 2026-08-09
fix(redact): path-guard NAME=value replacers so non-secret paths survive capture (#949)

### LINKED ISSUE #949: redact: prompt-dump NAME=value replacers match on key-name alone, masking non-secret paths — looksLikeFilesystemPath already exists in redact-secrets.ts
The witness-capture redactor masks values based on the **name** of the key alone, with no check on the shape of the value. Any `NAME=value` pair whose key contains `key` / `token` / `secret` / `password` / `credential` is redacted even when the value is plainly a filesystem path.

## Observed

Non-secret paths disappearing from captured subagent output — e.g. a line of the shape

```
AUTH_TOKEN_PATH=/Users/me/.config/app/token.json
```

is masked in full, purely because the key contains `TOKEN`. The path is not a credential; it is the *location* of one, and it is often the single most useful fact on the line when reconstructing what a child did.

## Mechanism

`src/agent/session/prompt-dump.ts:106-117` exports `redactInlineSecrets`, imported by **both** capture writers (`subagent-output-capture.ts` and `subagent-prompt-capture.ts`) — so this is shared, not an asymmetry between them.

The culprits are the two `NAME=value` replacers at `prompt-dump.ts:92` and `:97`, shaped like:

```
/([A-Za-z_]{3,}(?:key|token|secret|password|credential)[A-Za-z_]*)=([^\s]{16,})/gi
```

The value side is `[^\s]{16,}` — any non-whitespace run of 16+ characters. A POSIX path clears that trivially. There is no entropy check, no charset check, and no path check.

## The guard already exists in this repo

`src/agent/redact-secrets.ts:71-107` implements `looksLikeFilesystemPath()` for exactly this purpose — a *different* redactor in the same codebase already solved it. This is two redactors that disagree, and the capture path happens to import the naive one.

## Fix

Port `looksLikeFilesystemPath()` onto the two `NAME=value` replacers in `prompt-dump.ts` and skip redaction when the value is path-shaped. Bounded change: one import plus two guard clauses.

Worth deciding explicitly while in there: should a path-shaped value be passed through **verbatim**, or emitted as a marked non-secret (e.g. retained but annotated)? Verbatim is the more useful default for diagnosis and is what the sibling guard already does; flagging the choice so it is deliberate rather than incidental.

## Why it matters beyond cosmetics

This blocks the fix for tool-argument truncation in captured output (filed separately). That fix requires capturing `command` strings verbatim, and `command` is currently excluded from raw capture *because* of secret risk (`src/agent/facets/raw-input.ts:12-21`). Widening capture through a redactor that is simultaneously over-matching benign paths and unvalidated on real commands is the wrong order of operations. **This issue is the prerequisite.**

## Suggested tests

- `AUTH_TOKEN_PATH=/Users/x/.config/app/token.json` survives redaction intact.
- A real high-entropy `API_TOKEN=<40 random chars>` is still masked (guard must not open a hole).
- Both capture writers exercised, since they share the redactor.

## Provenance

Located by source reading during a follow-up session; over-match observed in real captured output. Not reproduced under test.

griffinwork40:
## Mechanism

`redactInlineSecrets` in `src/agent/session/prompt-dump.ts` (imported by both `subagent-output-capture.ts` and `subagent-prompt-capture.ts`) has two generic `NAME=value` redaction rules that match on **key name alone** — any key containing `key`/`token`/`secret`/`password`/`credential`/`auth`, with a value side of `[^\s]{16,}`. Because the check never inspects the value's shape, a filesystem path clears it: `AUTH_TOKEN_PATH=/Users/me/.config/app/token.json` was masked in full, even though the path only names a credential's *location*, not the credential itself.

## The fix

- `src/agent/redact-secrets.ts`: exported the existing `looksLikeFilesystemPath` helper (previously module-private), which already backs `redact-secrets.ts`'s own generic-token rule.
- `src/agent/session/prompt-dump.ts`: added `pathGuardedRedact`, wired through both `NAME=value` rules. It spares a value from redaction **only** when the value starts with `/` or `~/` **and** `looksLikeFilesystemPath(value)` also holds. Everything else — including any value that merely *contains* a `/` without that prefix — is still redacted.

## The AWS fail-safe finding, and how it's guarded

A naive port that gates purely on `looksLikeFilesystemPath(value)` (no prefix check) is a **regression that unmasks a real secret**: `looksLikeFilesystemPath('wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY')` — an AWS secret access key shape — returns `true` in isolation, because the value contains `/`, has no `+`/`=`, and every `/`-delimited segment is under the 32-char opaque-token threshold the helper uses.

The mitigation is the added prefix gate: `value.startsWith('/') || value.startsWith('~/')`. Env-var path values satisfy this virtually universally; secret values essentially never do. `AWS_SECRET_ACCESS_KEY=wJalrXUtnFEMI/...` does not start with `/` or `~/`, so it fails the gate and stays redacted — confirmed by a dedicated regression test (see below).

## Test evidence

```
export PATH="$HOME/.nvm/versions/node/v24.11.0/bin:$PATH"
./node_modules/.bin/tsc --noEmit
# exit 0

./node_modules/.bin/vitest run src/agent/prompt-dump.test.ts
# Test Files  1 passed (1)
# Tests  49 passed (49)   (includes 4 new #949 regression tests)

./node_modules/.bin/vitest run \
  src/agent/session/subagent-output-capture.test.ts \
  src/agent/session/subagent-prompt-capture.test.ts \
  src/agent/session/subagent-prompt-capture-wiring.test.ts \
  src/agent/session/subagent-output-capture-wiring.test.ts \
  src/agent/daemon/queue-store.test.ts
# Test Files  5 passed (5)
# Tests  84 passed (84)
```

New tests added to `src/agent/prompt-dump.test.ts` (`redactInlineSecrets` describe block):
1. `AUTH_TOKEN_PATH=/Users/me/.config/app/token.json` **survives** unredacted.
2. `CREDENTIAL_FILE_PATH=~/.config/app/credentials.json` **survives** unredacted (`~/` prefix).
3. `AWS_SECRET_ACCESS_KEY=wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY` **is still masked** — the critical fail-safe regression test.
4. A high-entropy opaque token with no path shape (`API_TOKEN=aB3xQ9zK7mN2pL5vR8tY1wC4dF6gH0jM`) **is still masked**.

## Files changed

- `src/agent/redact-secrets.ts` (+7/-1) — export `looksLikeFilesystemPath`.
- `src/agent/session/prompt-dump.ts` (+29/-14, 349 LOC total) — `pathGuardedRedact` helper, wired through both generic rules.
- `src/agent/prompt-dump.test.ts` (+37) — 4 new regression tests.

Closes #949. Unblocks #948.

### DIFF
--- a/src/agent/prompt-dump.test.ts
+++ b/src/agent/prompt-dump.test.ts
@@ -654,6 +654,22 @@ describe('redactInlineSecrets', () => {
     expect(result).toBe(text);
   });
 
+  it.each([
+    'AUTH_TOKEN_PATH="/Users/me/.config/app/token.json"',
+    "AUTH_TOKEN_PATH='/Users/me/.config/app/token.json'",
+  ])('spares a quoted non-secret filesystem-path value (%s)', (text) => {
+    expect(redactInlineSecrets(text)).toBe(text);
+  });
+
+  it.each([
+    ['API_TOKEN', '/2wYZ01xPz7pQeC0mFkA/'],
+    ['PASSWORD', '/correct/horse/battery/staple'],
+  ])('still masks slash-prefixed secret %s values', (name, secret) => {
+    const result = redactInlineSecrets(`${name}=${secret}`);
+    expect(result).not.toContain(secret);
+    expect(result).toBe(`${name}=<REDACTED length=${secret.length}>`);
+  });
+
   it('STILL MASKS an AWS secret access key even though it is path-shaped (fail-safe)', () => {
     // wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY satisfies looksLikeFilesystemPath()
     // in isolation (contains '/', no '+'/'=', every segment under the 32-char
--- a/src/agent/prompt-dump.test.ts
+++ b/src/agent/prompt-dump.test.ts
@@ -636,4 +636,41 @@ describe('redactInlineSecrets', () => {
     const result = redactInlineSecrets(text);
     expect(result).toBe(text);
   });
+
+  // Issue #949: value-shape guard on the generic NAME=value rules — a
+  // filesystem-path value should survive even though the key name matches
+  // (`*_TOKEN`), but the guard must stay fail-safe against real secrets that
+  // happen to contain '/'.
+
+  it('spares a non-secret filesystem-path value (AUTH_TOKEN_PATH=/…)', () => {
+    const text = 'AUTH_TOKEN_PATH=/Users/me/.config/app/token.json';
+    const result = redactInlineSecrets(text);
+    expect(result).toBe(text);
+  });
+
+  it('spares a non-secret filesystem-path value with ~/ prefix', () => {
+    const text = 'CREDENTIAL_FILE_PATH=~/.config/app/credentials.json';
+    const result = redactInlineSecrets(text);
+    expect(result).toBe(text);
+  });
+
+  it('STILL MASKS an AWS secret access key even though it is path-shaped (fail-safe)', () => {
+    // wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY satisfies looksLikeFilesystemPath()
+    // in isolation (contains '/', no '+'/'=', every segment under the 32-char
+    // opaque-token threshold) — the path-prefix requirement ('/' or '~/') is what
+    // keeps this masked. Regressing this check unmasks a real credential shape.
+    const secret = 'wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY';
+    const text = `AWS_SECRET_ACCESS_KEY=${secret}`;
+    const result = redactInlineSecrets(text);
+    expect(result).not.toContain(secret);
+    expect(result).toContain(`AWS_SECRET_ACCESS_KEY=<REDACTED length=${secret.length}>`);
+  });
+
+  it('still masks a high-entropy opaque token with no path shape', () => {
+    const secret = 'aB3xQ9zK7mN2pL5vR8tY1wC4dF6gH0jM';
+    const text = `API_TOKEN=${secret}`;
+    const result = redactInlineSecrets(text);
+    expect(result).not.toContain(secret);
+    expect(result).toContain(`API_TOKEN=<REDACTED length=${secret.length}>`);
+  });
 });
--- a/src/agent/prompt-dump.test.ts
+++ b/src/agent/prompt-dump.test.ts
@@ -662,8 +662,11 @@ describe('redactInlineSecrets', () => {
   });
 
   it.each([
+    // API_TOKEN has no suffix so `hasPathName` is false → redacted regardless of value shape.
     ['API_TOKEN', '/2wYZ01xPz7pQeC0mFkA/'],
-    ['PASSWORD', '/correct/horse/battery/staple'],
+    // SOME_PASSWORD has a 5-char prefix so it matches the {3,} quantifier guard, and the
+    // name does not have a PATH/FILE segment so `hasPathName` is false → redacted.
+    ['SOME_PASSWORD', '/correct/horse/battery/staple'],
   ])('still masks slash-prefixed secret %s values', (name, secret) => {
     const result = redactInlineSecrets(`${name}=${secret}`);
     expect(result).not.toContain(secret);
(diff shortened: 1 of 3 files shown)


griffinwork40:
@codex fix comments

griffinwork40:
Shadow-verify complete. Both claims independently confirmed:

1. **Regex widening (Finding #2): CONFIRMED.** `{3,}` → `*` is real, causes bare keywords to match, and is undocumented.
2. **Password-file-path bypass (Finding #3): CONFIRMED.** `PASSWORD_FILE_PATH=/correct/horse/battery/staple` evades redaction — all three guards pass. No test covers this case.

---

## PR #968 Review — `fix(redact): path-guard NAME=value replacers so non-secret paths survive capture (#949)`

**Target:** `afk/issue-949-redact-path-guard` → `main` | **Ref:** `d5e91db3` | **Regime:** light (hotfix, 103 lines)
**Files:** `src/agent/redact-secrets.ts` (+7/−1), `src/agent/session/prompt-dump.ts` (+43/−13), `src/agent/prompt-dump.test.ts` (+53/−0)

---

### Findings

**1** · `medium` · `blocking:true` · confidence:`high` · `security` · `src/agent/session/prompt-dump.ts:124,128` · ref:`d5e91db3` · `diff-context`
**Undocumented scope expansion: regex quantifier `{3,}` → `*` widens match surface.**
Both generic `NAME=value` regexes changed their prefix quantifier from `{3,}` (minimum 3 chars before the keyword) to `*` (zero-or-more). This means bare keyword names — `KEY=`, `Token=`, `Secret=`, `PASSWORD=`, `CREDENTIAL=`, `AUTH=` — with ≥16-char values now match and get redacted, where previously they required at least 3 prefix characters (e.g. `MY_KEY=`). This is a behavioral change not described in the PR body, which discusses only the path-guarding fix. While the direction is fail-safe (more redaction, not less), it is undocumented scope creep with no test coverage for the newly-matched bare-keyword cases.
```
// Old: [A-Za-z_]{3,}(?:KEY|TOKEN|...)
// New: [A-Za-z_]*(?:KEY|TOKEN|...)
```
**Suggestion:** Document the widening in the PR body. If intentional, add tests for `KEY=<value>` and `Token=<value>` to anchor the new behavior. If unintentional, revert to `{3,}`.

**2** · `medium` · `blocking:true` · confidence:`high` · `security` · `src/agent/session/prompt-dump.ts:95-100` · ref:`d5e91db3` · `diff-context`
**Path-word password bypass: `PASSWORD_FILE_PATH=/correct/horse/battery/staple` evades redaction.**
A variable name containing both a secret keyword (`PASSWORD`) and a path-segment keyword (`FILE`, `PATH`) with a slash-prefixed value of short readable words passes all three guards: `hasPathName` matches `_FILE_` in the name, `hasPathPrefix` fires on the leading `/`, and `looksLikeFilesystemPath` returns `true` (all segments <32 chars, no `+`/`=`). This leaks a passphrase value. The existing test uses `PASSWORD=/correct/horse/battery/staple` (no `_FILE_PATH` suffix), which correctly redacts — but the `PASSWORD_FILE_PATH=` variant is untested and leaks. Shadow-verified at runtime: all three conditions evaluate `true`.
```typescript
const hasPathName = /(?:^|_)(?:PATH|FILE)(?:_|$)/i.test(name);
// PASSWORD_FILE_PATH → matches _FILE_ → true
```
**Suggestion:** Add a test for `PASSWORD_FILE_PATH=/correct/horse/battery/staple` and decide the intended behavior. One fix: require `PATH` or `FILE` as the **terminal** segment of the name (e.g. `(?:_(?:PATH|FILE)$)`) rather than anywhere in the name, so `AUTH_TOKEN_PATH` still matches but `PASSWORD_FILE_PATH` (where the semantic is "password file path", not "path to something") fails the name guard.

**3** · `nit` · `blocking:false` · confidence:`high` · `test-coverage` · `src/agent/prompt-dump.test.ts` · ref:`d5e91db3` · `diff-context`
**`looksLikeFilesystemPath` is now a public export with no direct unit test.** It is tested only indirectly through `redactInlineSecrets`. As public API surface of `redact-secrets.ts`, a standalone test would anchor future changes.
**Suggestion:** Add direct unit tests for `looksLikeFilesystemPath` in `redact-secrets.test.ts` or a colocated test file.

**4** · `nit` · `blocking:false` · confidence:`high` · `api-compat` · `src/agent/redact-secrets.ts:105` · ref:`d5e91db3` · `diff-context`
**`looksLikeFilesystemPath` export is additive.** No production importers outside `prompt-dump.ts`. No breakage.

---

### Dimensions without findings

- **perf-observability** — no issues found. `pathGuardedRedact` is pure, lightweight (one regex test + string checks), invoked only during redaction. No hot-path impact.
- **spec-compliance** — Stated intent: PR title+body. The path-guarding fix satisfies the stated goal for #949. Finding #1 (regex widening) is flagged as scope creep above.

### What was not checked

- Citations verified inline against branch HEAD `d5e91db3`. All `diff-context` citations confirmed present at the reviewed ref.
- Stated intent: PR #968 title+body — spec-compliance assessed.
- Shadow-verify ran on findings #1 and #2; both CONFIRMED.
- Did not run the test suite (read-only review). PR body reports 49 tests passing including 4 new regression tests.
- Runtime behavior with actual subagent-prompt-capture and subagent-output-capture consumers not traced.

---

### Decision: DO NOT MERGE — 2 medium blocking (both security); 2 nit non-blocking.

Both medium findings are in the security dimension and therefore **not overridable** to `blocking:false`. Finding #2 is a genuine gap where a passphrase-valued variable evades redaction, confirmed by independent shadow-verification. Finding #1 is undocumented scope creep that widens the redaction match surface without test coverage. Both are addressable with small additions (tests + a one-line regex tightening or documentation).

---

PR #972, opened 2026-08-09
perf(permissions): memoize getWriteDenylist keyed on AFK_HOME + AFK_STATE_DIR (#781)

### LINKED ISSUE #781: perf: getWriteDenylist is uncached on the write hot path — memoize only with AFK_HOME + AFK_STATE_DIR in the key
Raised in review of #753 and **deliberately deferred** there. Filing so the reasoning survives.

## What

`getWriteDenylist()` is uncached and recomputes on every call:

```ts
// src/agent/tools/handlers/write-denylist.ts:111
export function getWriteDenylist(): readonly string[] {
```

It runs `safeRealpath` (a bounded `realpathSync` walk) over 12 builtin entries plus the derived AFK entries. #753 added up to 3 derived entries (`config`, `state`, and the independently-relocated `AFK_STATE_DIR`), so the per-call syscall count went from ~11 to ~14.

Callers on the hot path: `assertNotDenylisted` (same file), reached by every `write_file` / `edit_file`, and `src/agent/risk-classifier.ts:174`.

## Why it was NOT fixed in #753

Adding a cache to a credential denylist is the exact bug class that PR had to fix on the **read** side: a memo whose key omits an input serves a stale, non-covering floor after a runtime relocation. #753's `read-denylist.ts:311` key exists solely to prevent that:

```ts
const key = `${env.AFK_READ_DENYLIST ?? ''}\^@${env.AFK_HOME ?? ''}`;
```

Trading a correctness risk on a security floor for ~2 saved `realpathSync` calls, with no profiling showing the write path is hot enough to care, was the wrong trade to make inside a security PR under review. Both automated review passes rated it non-blocking ("None required for merge").

## If/when this is done — hard requirement

The cache key **must** include every env var the result depends on:

- `AFK_WRITE_DENYLIST`
- `AFK_HOME`
- `AFK_STATE_DIR`  ← easy to miss; it relocates the state tier *independently* of `AFK_HOME` (`src/paths.ts` returns it verbatim when set)

Omitting `AFK_STATE_DIR` reintroduces the stale-floor bug in a new place. Join with `\^@` as the read side does, so components cannot collide into an ambiguous key.

Also required: a cache-invalidation regression test that queries, changes the env var with **no** manual cache reset, and queries again — mirroring `read-denylist.test.ts`'s guard, which deliberately omits the reset because calling it would make the test pass even if the key were wrong.

The doc comment at `write-denylist.ts:65-68` currently documents the uncached design and must be updated, not left to become a lie.

## Prerequisite

Profile first. If the write path is not measurably hot, the right resolution is to close this as `wontfix` and keep the simpler uncached code.

---

*Citation anchor corrected: line numbers below refer to `main` @ `2675eaf` (the squash merge of #753). They were originally written against branch `2aa5fbb`; `read-denylist.ts` shifted ~63 lines and `bash-restriction-hook.ts` ~8 lines because #735 landed alongside #753.*

griffinwork40:
## Hot path

`getWriteDenylist()` recomputed the full denylist — 12 builtin entries plus up to 3 derived AFK entries, each passed through `safeRealpath` (a bounded `realpathSync` walk) — on **every call**. It's reached by:

- `assertNotDenylisted` (same file), invoked on every `write_file` / `edit_file` call
- `risk-classifier.ts:174`

\#753 added the derived `AFK_STATE_DIR` entry alongside the existing `config`/`state` derivations, growing the per-call syscall count from ~11 to ~14.

## The memo

Added a module-level cache in `write-denylist.ts`, structurally mirroring the existing `read-denylist.ts` memo:

```ts
let cached: { key: string; list: readonly string[] } | undefined;

export function getWriteDenylist(): readonly string[] {
  const key = `${env.AFK_WRITE_DENYLIST ?? ''}\u0000${env.AFK_HOME ?? ''}\u0000${env.AFK_STATE_DIR ?? ''}`;
  if (cached && cached.key === key) return cached.list;
  // ...recompute, cache, return
}
```

## Why the key includes both `AFK_HOME` and `AFK_STATE_DIR`

Per the issue's hard requirement — a cache on a credential denylist is exactly the bug class #753 had to fix on the **read** side, where a memo key that omits an input serves a stale, non-covering floor after a runtime relocation.

Both env vars feed `derivedAfkHomeWriteEntries()`:

- `AFK_HOME` relocates the derived `config`/`state` entries via `getAfkHome()`.
- `AFK_STATE_DIR` relocates the state tier **independently** of `AFK_HOME` — `paths.ts` returns it verbatim when set, only falling back to `join(getAfkHome(), 'state')` otherwise.

Omitting either from the key would let a runtime relocation of that var keep serving the pre-relocation denylist — silently permitting a write to the real, now-uncovered credential tree. The key joins all three components (`AFK_WRITE_DENYLIST`, `AFK_HOME`, `AFK_STATE_DIR`) with `U+0000`, matching `read-denylist.ts`'s existing key construction, so components can never collide into an ambiguous key.

## Invalidation test evidence

Added `_resetWriteDenylistCacheForTests()` (mirrors `_resetReadDenylistCacheForTests`) and two regression guards in `write-denylist.test.ts`, structured like `read-denylist.test.ts`'s existing `AFK_HOME` guard — deliberately **no manual cache reset** between the two queries, because that reset is exactly what a correct key should make unnecessary:

- `invalidates the memoized denylist when AFK_HOME changes, with no manual cache reset` — queries a target path with `AFK_HOME` unset (not denied), sets `AFK_HOME` via `vi.stubEnv`, queries again (now denied) — no reset call in between.
- `invalidates the memoized denylist when AFK_STATE_DIR changes, with no manual cache reset` — same shape for `AFK_STATE_DIR` independently.

Both pass. Full scoped run: `write-denylist.test.ts` (52 tests) + `read-denylist.test.ts` (69 tests, unmodified — run as a regression check since both modules share the pattern) — **121 passed, 0 failed**.

## Files changed

- `src/agent/tools/handlers/write-denylist.ts` (+41/-5, 220 LOC total) — memo + cache key + `_resetWriteDenylistCacheForTests`, doc-comment update on `derivedAfkHomeWriteEntries` (no longer describes the module as uncached).
- `src/agent/tools/handlers/write-denylist.test.ts` (+46, 622 LOC total) — two invalidation regression guards.

## Verification

- `tsc --noEmit`: exit 0
- `vitest run write-denylist.test.ts read-denylist.test.ts`: 121/121 passed

Closes #781

### DIFF
--- a/src/agent/tools/handlers/write-denylist.test.ts
+++ b/src/agent/tools/handlers/write-denylist.test.ts
@@ -555,6 +555,52 @@ describe('write-denylist — AFK_HOME-relocated credential tree (#740)', () => {
   });
 });
 
+// ---------------------------------------------------------------------------
+// Memoization cache-key regression guard (#781)
+// ---------------------------------------------------------------------------
+
+describe('write-denylist — memoization cache-key regression guard (#781)', () => {
+  // Cache-invalidation: the memoization key must include AFK_HOME and
+  // AFK_STATE_DIR, not just AFK_WRITE_DENYLIST, or a runtime change to either
+  // returns a STALE denylist that fails to cover a newly-relocated credential
+  // tree — silently permitting a write that should be denied.
+  // Invariant: deliberately NO `_resetWriteDenylistCacheForTests()` call
+  // between the two queries below — that reset is what the memo key itself is
+  // supposed to make unnecessary. Calling it here would make this test pass
+  // even if AFK_HOME/AFK_STATE_DIR were dropped from the key entirely,
+  // defeating the point of the regression guard. Mirrors
+  // read-denylist.test.ts's identically-purposed guard.
+  it('invalidates the memoized denylist when AFK_HOME changes, with no manual cache reset', () => {
+    const relocated = join(tmpDir, 'relocated-cache-home');
+    mkdirSync(relocated, { recursive: true });
+    const target = join(relocated, 'config', 'afk.env');
+
+    // Query once with AFK_HOME unset: the relocated config dir is NOT covered.
+    expect(() => assertNotDenylisted(target, 'write_file')).not.toThrow();
+
+    // Change AFK_HOME and query again — WITHOUT resetting the cache by hand.
+    vi.stubEnv('AFK_HOME', relocated);
+    expect(() => assertNotDenylisted(target, 'write_file')).toThrow(
+      /refusing to write to protected path/,
+    );
+  });
+
+  it('invalidates the memoized denylist when AFK_STATE_DIR changes, with no manual cache reset', () => {
+    const stateDir = join(tmpDir, 'relocated-cache-state');
+    mkdirSync(stateDir, { recursive: true });
+    const target = join(stateDir, 'sessions', 's.json');
+
+    // Query once with AFK_STATE_DIR unset: the relocated state dir is NOT covered.
+    expect(() => assertNotDenylisted(target, 'write_file')).not.toThrow();
+
+    // Change AFK_STATE_DIR and query again — WITHOUT resetting the cache by hand.
+    vi.stubEnv('AFK_STATE_DIR', stateDir);
+    expect(() => assertNotDenylisted(target, 'write_file')).toThrow(
+      /refusing to write to protected path/,
+    );
+  });
+});
+
 describe('assertNotDenylisted — case-variant spellings (#736)', () => {
   afterEach(() => {
     _resetFsCaseCacheForTests();
--- a/src/agent/tools/handlers/write-denylist.test.ts
+++ b/src/agent/tools/handlers/write-denylist.test.ts
@@ -599,6 +599,29 @@ describe('write-denylist — memoization cache-key regression guard (#781)', () 
       /refusing to write to protected path/,
     );
   });
+
+  it('re-resolves a custom denylist symlink after it is repointed', () => {
+    const firstTarget = join(tmpDir, 'first-target');
+    const secondTarget = join(tmpDir, 'second-target');
+    const link = join(tmpDir, 'blocked-link');
+    mkdirSync(firstTarget);
+    mkdirSync(secondTarget);
+    symlinkSync(firstTarget, link);
+    vi.stubEnv('AFK_WRITE_DENYLIST', link);
+
+    expect(() =>
+      assertNotDenylisted(join(link, 'secret.txt'), 'write_file'),
+    ).toThrow(/refusing to write to protected path/);
+
+    rmSync(link);
+    symlinkSync(secondTarget, link);
+
+    // The environment key did not change, so this only remains protected if
+    // getWriteDenylist re-resolves entries after a cache hit.
+    expect(() =>
+      assertNotDenylisted(join(link, 'secret.txt'), 'write_file'),
+    ).toThrow(/refusing to write to protected path/);
+  });
 });
 
 describe('assertNotDenylisted — case-variant spellings (#736)', () => {
--- a/src/agent/tools/handlers/write-denylist.ts
+++ b/src/agent/tools/handlers/write-denylist.ts
@@ -79,10 +79,11 @@ export const BUILTIN_WRITE_DENYLIST: readonly string[] = [
  * `homedir()`-based entries still apply regardless. The throw is no longer
  * silent: {@link warnAfkHomeRejectedOnce} surfaces it once per process.
  *
- * Called fresh on every {@link getWriteDenylist} call (this module is
- * uncached, unlike the read denylist), so no separate cache-key concern
- * applies here — a runtime `AFK_HOME` / `AFK_STATE_DIR` change is picked up on
- * the very next call.
+ * Not cached on its own — its result feeds directly into
+ * {@link getWriteDenylist}'s memo, which keys on `AFK_HOME` and
+ * `AFK_STATE_DIR` (alongside `AFK_WRITE_DENYLIST`) precisely so a runtime
+ * change to either is observed on the very next call rather than served a
+ * stale floor (#781).
  */
 function derivedAfkHomeWriteEntries(): string[] {
   const entries: string[] = [];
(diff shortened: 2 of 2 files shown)


griffinwork40:
@codex fix comments

griffinwork40:
**Wave 1.5 verification complete. All critical checks done.**

Key verifications:
- `assertNotDenylisted` line 180: **confirmed** — `const real = safeRealpath(resolve(filePath))` — target is always resolved
- Post-PR `getWriteDenylist()` return: `cached.unresolved.map((p) => safeRealpath(p))` — symlinks in derived entries ARE resolved at call time
- `_resetWriteDenylistCacheForTests` import: **not present** in the test file's import block (diff shows +69/-0 at line 555 only; import section unchanged)
- Top-level `afterEach` (line 49): calls `vi.unstubAllEnvs()` — env stubs cleaned between every test
- Symlink test (S-2): if resolved paths were cached, second assertion (`join(link, 'secret.txt')` after repoint) would resolve to `join(secondTarget, 'secret.txt')` while cache would hold `firstTarget` → `pathIsWithin` fails → test **would fail**. Finding is not vacuous.

**Wave 1.5 dispositions:**

| Finding | Pre-1.5 sev | Action | Reason |
|---------|-------------|--------|--------|
| S-1 (high/security) | high | **DROPPED** | Fabricated. `safeRealpath` IS applied to `cached.unresolved` at return time — symlinked AFK_HOME resolves correctly |
| S-2 (medium/security) | medium | **DOWNGRADE → nit** | Test is not vacuous — would fail under a resolved-path cache. Suggestion is improvement, not blocker |
| SC-1 (nit) | nit | confirmed | PR body `list` vs impl `unresolved` |
| C-2 (medium/correctness) | medium | **DOWNGRADE → low** | `vi.unstubAllEnvs()` in global `afterEach` restores env between tests; temp-dir paths never match default denylist on cache hit |
| T-1 (low) | low | **DOWNGRADE → nit** | AFK_HOME-as-symlink case correctly handled (S-1 basis dropped); still a nice-to-have |
| P-1 (low) | low | confirmed | Per-call `safeRealpath` pass unchanged on cache hits |
| C-new (new) | — | **ADDED: low** | `_resetWriteDenylistCacheForTests()` is exported but not imported/called in the test file; suites that mutate env vars are not updated |

No `critical` or `high` findings survive. No disposition overrides. Shadow-verify is not triggered.

---

## Review — PR #972: `perf(permissions): memoize getWriteDenylist keyed on AFK_HOME + AFK_STATE_DIR`

**Target:** `afk/issue-781-denylist-memo → main`  
**Ref verified against:** Working tree (branch HEAD `0853feb` not fetched locally; citations verified against pre-PR working tree file + diff context)  
**Stated intent:** Available — PR #972 title + body assessed for spec-compliance  
**Regime:** Light (140 lines changed)

---

### Findings

**C-new · low · blocking:false · test-coverage · write-denylist.ts:180–187 + write-denylist.test.ts:29–34 · ref:0853feb · diff-context**

`_resetWriteDenylistCacheForTests()` is exported in the production diff but not imported in the test file — the import block at `write-denylist.test.ts:29–34` is unchanged (the diff shows only one hunk at line 555+). The function's JSDoc says it is needed by "suites that mutate `AFK_WRITE_DENYLIST` / `AFK_HOME` / `AFK_STATE_DIR`" to avoid stale list hits, but those existing suites are not updated to call it.

*Practical impact:* Minimal — `vi.unstubAllEnvs()` in the global `afterEach` restores env vars between every test, so a prior test's key won't match a subsequent test's env, triggering a cache miss. But the function is dead export code and the docstring creates a false expectation of usage.

*Suggestion:* Add `import { _resetWriteDenylistCacheForTests } from './write-denylist.js'` and a top-level `afterEach(() => _resetWriteDenylistCacheForTests())` alongside the existing `vi.unstubAllEnvs()`. This makes the isolation explicit and mirrors the pattern used for `_resetFsCaseCacheForTests` in the same file.

---

**C-2 · low · blocking:false · correctness · write-denylist.test.ts:555–76 · ref:0853feb · diff-context**

Cache-key invalidation tests omit `_resetWriteDenylistCacheForTests()` before the *first* query — deliberately between the two queries, but the pre-first-query state is silently inherited from whatever the module-level `cached` holds. Harmless in practice (see above), but the omission is undocumented and fragile if test ordering changes.

*Evidence:*
```ts
// Query once with AFK_HOME unset: the relocated config dir is NOT covered.
expect(() => assertNotDenylisted(target, 'write_file')).not.toThrow();
```
No reset call before this line.

*Suggestion:* Add `_resetWriteDenylistCacheForTests()` at the top of each of the two cache-key `it` blocks (before the first query), with a comment explaining it is intentionally absent *between* the two queries but needed before the first.

---

**P-1 · low · blocking:false · perf-observability · write-denylist.ts:170 · ref:0853feb · diff-context**

The PR body states the optimization eliminates "~14 `realpathSync` calls" on cache hits, but the return statement still calls `safeRealpath` on every entry unconditionally:

```ts
return [...new Set(cached.unresolved.map((p) => safeRealpath(p)))];
```

The code's own JSDoc explains why this is intentional ("Do not memoize safeRealpath results — a configured or built-in entry can contain a symlink whose target changes"). The architecture is correct and well-commented. The PR body's performance framing is misleading: the memo eliminates the *entry-construction* cost (path.resolve calls, Set deduplication, env reads into `getAfkHome()`/`getAfkStateDir()`) but not the `realpathSync` syscalls, which remain ~14–15 per call.

*Suggestion:* Update the PR body to say "eliminates re-derivation of unresolved entries on cache hits; per-call `safeRealpath` pass on all entries is retained by design."

---

**S-2 · nit · blocking:false · test-coverage · write-denylist.test.ts:554–75 · ref:0853feb · diff-context**

The symlink-repoint test correctly verifies that live resolution is preserved (would fail if resolved paths were cached). A complementary assertion would strengthen the invariant: after repointing `link → secondTarget`, also assert that `join(firstTarget, 'secret.txt')` no longer throws — demonstrating that the *old* resolved target loses coverage when the symlink moves, as intended.

---

**SC-1 · nit · blocking:false · spec-compliance · PR body vs. diff**

PR body shows `let cached: { key: string; list: readonly string[] }` but the implementation uses field name `unresolved`. The implementation name is architecturally better (distinguishes from the resolved return value). PR body is stale from a design iteration.

---

**T-1 · nit · blocking:false · test-coverage**

No test covers the case where `AFK_HOME` is itself a symlink directory (e.g. `AFK_HOME=/symlink-dir → /actual/afk`). Analysis confirms the case works correctly: `safeRealpath('/symlink-dir/config')` at return time resolves through the symlink, so the denylist correctly contains `/actual/afk/config`. Belt-and-suspenders coverage gap only.

---

### Dropped findings (Wave 1.5)

| Finding | Reason |
|---------|--------|
| **S-1 (high/security)** — "symlinked `AFK_HOME` bypasses denylist" | **Fabricated.** `getWriteDenylist()` returns `cached.unresolved.map((p) => safeRealpath(p))` — symlinks in derived entries ARE followed at return time. `assertNotDenylisted` also resolves its target via `safeRealpath`. Both sides are resolved; `pathIsWithin` matches correctly. |

---

### Epistemic scope

- Citations verified inline against working tree (pre-PR). Branch HEAD `0853feb172a95d898ebba393ce8cd9cd93002803` is not fetched locally; diff-context citations verified against the diff hunk; file-state citations verified by reading the pre-PR working tree file.
- `assertNotDenylisted` implementation read directly at `write-denylist.ts:179–189`.
- Did not review `risk-classifier.ts:174` call site beyond confirming it calls `getWriteDenylist()`.
- Did not run the test suite.
- Stated intent available (PR title + body) — spec-compliance assessed.

---

### Decision

**MERGE — 0 blocking (3 low, 3 nit)**

The implementation is correct. The memoization key includes all three env vars that influence `derivedAfkHomeWriteEntries()`. Symlink-resolved paths are intentionally kept live at return time, not cached — the security invariant is preserved. The three low findings (unused test export, undocumented pre-first-query cache state, PR body performance overclaim) are all follow-up cleanup, none blocking. The nits are improvements only.

**Done**

- Review of PR #972 complete — read-only, no files edited or committed
- One fabricated high/security finding (S-1) dropped in Wave 1.5; final verdict MERGE
- Findings surfaced at `write-denylist.ts` (P-1) and `write-denylist.test.ts` (C-new, C-2, S-2, T-1) plus PR body (SC-1)
- No shadow-verify invoked (no surviving `high`/`critical` findings, no disposition overrides)

---

Issue #710, opened 2026-07-26
CLI: mistyped subcommand is swallowed by the default `interactive` command — no "unknown command" error, misleading "unknown option" instead


griffinwork40:
## Summary

A mistyped or non-existent subcommand is silently routed into the default `interactive` command as prompt input. There is no "unknown command" error. Depending on whether the typo'd invocation also carries flags, the user gets one of two confusing outcomes:

1. **With an unrecognized flag** — an error that blames the *flag* and never reveals the *command* was unrecognized.
2. **Without flags** — a full interactive session boots (MCP servers connect, a model turn is spent) and the typo is submitted to the model as a chat prompt.

Case 2 is the expensive one: a typo costs a session bootstrap, MCP connections, and billed tokens, only to have the model reply that it doesn't recognize the input.

## Reproduction

```console
$ afk config_set env AFK_MAX_BUDGET_USD --unset
error: unknown option '--unset'
# exit 1
```

`--unset` is not the problem. `config_set` is not a command (the real one is `afk config env unset <key>`). But the error points at the flag, so the reader concludes the *command* is right and the *flag* is wrong — the opposite of the truth. This sent me on a detour before I noticed `config_set` was never a command at all.

```console
$ afk zzzbogus
- Initializing interactive session...
  mcp: 1 server(s) from ~/.afk/config/mcp.json
[chrome-devtools-mcp] connected — 29 tool(s)

afk ⚡ bp  › zzzbogus
"zzzbogus" isn't a recognized command, skill, or task ...
# hangs in the REPL; killed at 12s
```

A bogus token boots the whole stack and gets answered by the *model* rather than the *CLI*.

## Expected

`afk config_set …` and `afk zzzbogus` should fail fast with an unknown-command error and a did-you-mean suggestion, e.g.:

```
error: unknown command 'config_set'
  did you mean: config?
```

…while `afk "what does this project do"` and `afk /review` keep working exactly as they do today.

## Root cause

`src/cli/commands/interactive.ts:186` registers `interactive` as commander's **default command** with a variadic argument:

```ts
.command('interactive', { isDefault: true })
  .argument('[input...]', 'Optional first message — or a /slash-command — auto-submitted as the opening turn. ...')
```

[...]

PR #974, opened 2026-08-09
fix(cli): name the unrecognized command instead of blaming its flag (#710)

griffinwork40:
## Issue

#710 identifies two distinct ways a mistyped `afk` subcommand misleads the user:

1. **Flag-blaming**: `afk config_set env X --unset` (a typo of `afk config set env X --unset`) surfaces `error: unknown option '--unset'` — the user is told their *flag* is bad, when the real problem is that `config_set` isn't a registered command at all.
2. **Silent boot**: a bare unrecognized token with no flags, e.g. `afk zzzbogus`, boots a full interactive REPL session instead of erroring, because nothing in that path ever fires an "unknown command" check.

## What this PR fixes (mode 1 only)

Root cause: `interactive` is registered as commander's *default* command (`.command('interactive', { isDefault: true })` with `.argument('[input...]')`). Commander's `_parseCommand` dispatches any unrecognized leading token to that default command **before** `unknownCommand()` ever gets a chance to fire. The default command's own `unknownOption()` then fires on whatever flag happens to be unknown, blaming the flag instead of the command.

Fix: `installUnknownCommandGuard` (new file: `src/cli/commands/interactive/unknown-command-guard.ts`) monkey-patches `unknownOption` on the dispatched `interactive` Command instance. Inside that hook, `this.args[0]` holds the *original* first token — when it isn't a registered command/alias and doesn't start with `/` or `-`, the guard raises `unknown command '<token>'` instead of falling through to the flag-blaming message.

`configureOutput` was considered and rejected as the interception point: it only observes the already-formatted error string, not the original argv token, so it can't distinguish "mistyped command" from "a genuinely bad flag on a valid command/bare prompt."

**Before:**
```
$ afk config_set env X --unset
error: unknown option '--unset'
```

**After:**
```
$ afk config_set env X --unset
error: unknown command 'config_set'
```

### Zero regression risk on adjacent paths

- `afk /review` — a documented, load-bearing slash-command launch path — is explicitly exempted (leading `/`) and untouched.
- `afk --badflag` — the token *is* the bad flag itself (leading `-`) — keeps today's flag-blaming message unchanged.
- `afk "do a thing"` — a bare prompt with no unknown flag — never reaches `unknownOption` at all, so the guard never engages; REPL boots exactly as before.

All four behaviors (the fix + the three unchanged paths) are covered by a new test file, `unknown-command-guard.test.ts` (12 tests), plus the existing `interactive.resume.test.ts` and `index-integration.test.ts` suites still pass unmodified.

## Residual gap — mode 2 is NOT fixed here

A bare unrecognized token with **no flags at all** (`afk zzzbogus`) still boots a full interactive session instead of erroring. This is a different code path: commander never calls `unknownOption`/`unknownCommand` when there's no unknown flag to trip on, so this guard — which hooks `unknownOption` — structurally cannot intercept it.

Closing that gap would require a "did you mean" near-miss heuristic comparing the typed token against registered command names. That heuristic was built and empirically tested against real prompt words before this PR, and hard-errors too often to ship:

- At Levenshtein distance ≤3 against all registered command names: **14–16 false positives out of 20** ordinary one-word prompts, including `stats`→`status`, `grace`/`brace`/`track`→`trace`, `confirm`→`config`, `log`→`login`, `plan`→`plugin`, `commit`→`config`, `where`→`queue`.
- Tightened to ≤2 with a canonical-names-only pool: still **8/20** false positives.

Hard-erroring on someone's legitimate one-word prompt (a real, common `afk "one word"` usage pattern) is a worse regression than the bug being fixed. That half of #710 is intentionally left open — the false-positive data above is left for the maintainer to decide whether a different heuristic, an allowlist, or another approach entirely is worth the complexity.

## Verification

- `./node_modules/.bin/tsc --noEmit` — exit 0
- `./node_modules/.bin/vitest run src/cli/commands/interactive/unknown-command-guard.test.ts` — 12/12 passed
- `./node_modules/.bin/vitest run src/cli/commands/interactive.resume.test.ts src/cli/index-integration.test.ts` — 11/11 passed (no regression)
- Manual smoke test via `tsx src/cli/index.ts`:
  - `config_set env X --unset` → `error: unknown command 'config_set'` ✓
  - `--badflag` (bare) → `error: unknown option '--badflag'` (unchanged) ✓
  - `"do a thing"` → proceeds into REPL boot (unchanged) ✓
  - `/review` → proceeds into REPL boot, reaches app-level slash-command resolution unchanged ✓

Refs #710

### DIFF
--- a/src/cli/commands/interactive.ts
+++ b/src/cli/commands/interactive.ts
@@ -35,6 +35,7 @@ import {
   resolveWorktreeDisposition,
   resolveWorktreeExitPolicy,
 } from './interactive/worktree-disposition.js';
+import { installUnknownCommandGuard } from './interactive/unknown-command-guard.js';
 
 export { formatToolResultLine } from './interactive/tool-lane.js';
 
@@ -188,7 +189,7 @@ export function startupHintLine(): string {
 }
 
 export function registerInteractiveCommand(program: Command): void {
-  program
+  const interactiveCmd = program
     .command('interactive', { isDefault: true })
     .description('Start interactive chat session')
     .argument(
@@ -893,6 +894,12 @@ export function registerInteractiveCommand(program: Command): void {
 
       await runReplLoop(ctx, transcript, turnState, handleSigint);
     });
+
+  // Issue #710 mode 1: name the unrecognized COMMAND, not its trailing flag,
+  // when a mistyped subcommand (e.g. `afk config_set env X --unset`) falls
+  // through to this default command. See unknown-command-guard.ts for the
+  // full mechanism and the documented residual gap (mode 2).
+  installUnknownCommandGuard(interactiveCmd, program);
 }
 
 /**
--- a/src/cli/commands/interactive/unknown-command-guard.test.ts
+++ b/src/cli/commands/interactive/unknown-command-guard.test.ts
@@ -1,0 +1,131 @@
+/**
+ * Issue #710 mode 1: a mistyped subcommand (`afk config_set env X --unset`)
+ * blamed its trailing flag (`error: unknown option '--unset'`) instead of
+ * naming the unrecognized command. These tests exercise the guard directly
+ * against real commander `Command` instances (no mocking of interactive.ts's
+ * heavy bootstrap chain needed — the guard is a pure commander-level concern).
+ */
+
+import { describe, expect, it } from 'vitest';
+import { Command } from 'commander';
+import { isUnrecognizedCommandToken, installUnknownCommandGuard } from './unknown-command-guard.js';
+
+/** Build a minimal program shaped like the real CLI: a default `interactive`
+ * command plus a couple of named sibling subcommands, with the guard wired
+ * in exactly as `registerInteractiveCommand` does. */
+function buildProgram(): Command {
+  const program = new Command();
+  program.name('afk').exitOverride();
+
+  const interactiveCmd = program
+    .command('interactive', { isDefault: true })
+    .argument('[input...]')
+    .option('--model <model>', 'model')
+    .action(() => {
+      /* no-op: presence of a call is asserted via captured state below */
+    });
+  program.commands.find((c) => c.name() === 'interactive')?.alias('i');
+
+  program
+    .command('config')
+    .description('config stuff')
+    .action(() => undefined);
+
+  installUnknownCommandGuard(interactiveCmd, program);
+  return program;
+}
+
+describe('isUnrecognizedCommandToken', () => {
+  const program = buildProgram();
+
+  it('flags a plain mistyped token', () => {
+    expect(isUnrecognizedCommandToken('config_set', program)).toBe(true);
+  });
+
+  it('does not flag a registered command name', () => {
+    expect(isUnrecognizedCommandToken('config', program)).toBe(false);
+  });
+
+  it('does not flag a registered alias', () => {
+    expect(isUnrecognizedCommandToken('i', program)).toBe(false);
+  });
+
+  it('does not flag a leading slash (documented slash-command launch path)', () => {
+    expect(isUnrecognizedCommandToken('/review', program)).toBe(false);
+  });
+
+  it('does not flag a leading dash (the token IS the unknown flag itself)', () => {
+    expect(isUnrecognizedCommandToken('--badflag', program)).toBe(false);
+  });
+
+  it('does not flag undefined or empty', () => {
+    expect(isUnrecognizedCommandToken(undefined, program)).toBe(false);
+    expect(isUnrecognizedCommandToken('', program)).toBe(false);
+  });
+});
+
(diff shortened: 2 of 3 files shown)


griffinwork40:
Good — PR #974 is indeed `afk/issue-710-cli-unknown-command` with those 3 files. The research agent was confused by the workspace branch name `afk/issue-752-agent-parse-warn` (which is the *current checkout*, not the PR branch). The review dimension agents read the correct diff from `/tmp/pr-974.diff`.

---

## Review: PR #974 — `fix(cli): name the unrecognized command instead of blaming its flag (#710)`

**Branch:** `afk/issue-710-cli-unknown-command` → `main`  
**Scope:** 3 files, +212/−1  
**Verdict: Approve** — clean, well-scoped fix with thorough documentation and tests. No blocking issues.

---

### What the PR does

When a mistyped subcommand carries a trailing flag (e.g. `afk config_set env X --unset`), commander's default-command dispatch causes the flag to be blamed (`error: unknown option '--unset'`) instead of the unrecognized command. The fix monkey-patches `unknownOption` on the `interactive` Command instance to intercept this case and emit `error: unknown command 'config_set'` instead.

The fix is surgically scoped to "mode 1" (mistyped command + trailing flag) and explicitly documents and tests the mode-2 residual gap (bare unknown token with no flags), including the Levenshtein false-positive analysis justifying why that gap is left open.

---

### Dimension findings

#### Correctness ✅
- The monkey-patch intercepts `unknownOption` correctly. `patchTarget.args[0]` holds the leading token at the time `unknownOption` fires for the mistyped-command case.
- **The initial HIGH finding (false positive on `afk write me a test --dry-run`) was independently REFUTED.** Commander's variadic `[input...]` argument consumes all prompt tokens before `unknownOption` fires, so `args[0]` is the parsed input array, not the bare string `'write'`. The existing test suite confirms this.
- The `patchTarget.error()` call uses commander's standard error API with `code: 'commander.unknownCommand'` — correct and idiomatic.
- `bind(patchTarget)` correctly preserves `this` for the original `unknownOption`.

#### Security ✅
- **LOW:** `firstToken` is interpolated unsanitized into the error string. The project has `sanitizeForDisplay()` in `src/utils/terminal-sanitize.ts` for user-derived terminal output. The attack surface is negligible (local user controls their own argv), but applying `sanitizeForDisplay(firstToken)` would be consistent with the project's existing security boundary. One-line optional hardening.

#### API Compatibility ✅
- Capturing the `.command()` chain return value is a no-op change to observable behavior.
- No callers parse or depend on the old `error: unknown option '...'` message string.
- `exitOverride()` propagates from parent to child, so `patchTarget.error()` throws `CommanderError` correctly.
- The guard is installed post-`.action()` / pre-`parseAsync` — correct lifecycle ordering.
- The patch is instance-scoped (own-property on `interactiveCmd`), no prototype pollution.

#### Test Coverage ✅ (with suggestions)
- All four documented behavioral paths are tested end-to-end: mistyped command + flag, bare prompt, slash-command, bare bad flag.
- Mode-2 residual gap is explicitly tested as a documented non-fix.
- `isUnrecognizedCommandToken` edge cases (undefined, empty, slash-prefix, dash-prefix, registered commands/aliases) are covered.
- 12 tests total, assertions check both presence and absence of key strings.

---

### Non-blocking suggestions

| Severity | Finding | Location |
|---|---|---|
| **Low** | `return` after `patchTarget.error()` is dead code (`.error()` always throws or exits). Harmless but worth a one-line comment: `// .error() throws via exitOverride — unreachable` | `unknown-command-guard.ts:63` |
| **Low** | Wrap `firstToken` in `sanitizeForDisplay()` for consistency with the project's terminal-sanitization boundary | `unknown-command-guard.ts:61` |
| **Nit** | JSDoc says `this.args[0]`, implementation uses `patchTarget.args[0]` — minor comment/code mismatch | `unknown-command-guard.ts:38` vs `:59` |
| **Nit** | Consider adding a test for prefix-of-registered-command (`afk confi --unset`) — pins the "no did-you-mean" behavior | test file |
| **Nit** | Consider adding a test for multiple unknown flags (`config_set --f1 --f2`) to verify only one error fires | test file |

None of these are merge-blocking. The PR body is unusually thorough — it documents the root cause, the rejected alternative (`configureOutput`), the residual gap with quantitative false-positive analysis, and the zero-regression argument for all adjacent paths. Ship it.

**Done.**

- Reviewed PR #974 across 4 dimensions (correctness, security, API compatibility, test coverage) via parallel subagents
- Shadow-verified and refuted the initial false-positive claim — commander's variadic argument parsing prevents the hypothesized regression
- Findings: no blocking issues, 2 low-severity suggestions (dead-code comment, `sanitizeForDisplay`), 3 nits
- Review is read-only — no working tree mutations

---

Issue #747, opened 2026-07-29
defense-in-depth: sanitize operator- and repo-controlled paths in agent-registry warnings


griffinwork40:
Surfaced while reviewing #739. **Pre-existing — not introduced by that PR**, which added a fifth emission site to a pattern that already had four.

## What

Every warning in the agent registry interpolates a filesystem path straight into terminal output, unsanitized:

- `src/agent/agents/registry.ts:150` — `[afk] agents: ${origin} overrides built-in agent ${JSON.stringify(name)}…`
- `src/agent/agents/registry.ts:179` — `[afk] agents: cannot read ${filePath}: …`
- `src/agent/agents/registry.ts:182` — `[afk] agents: ${filePath}: ${msg}`
- `src/agent/agents/registry.ts:188` and `:202` — duplicate-name warnings interpolating `priorInScope` / `filePath`

The agent *name* is escaped via `JSON.stringify`, so C0 and ESC bytes in a frontmatter `name` are neutralized. The **path** is not.

Meanwhile the repo ships a helper for precisely this, which is declared a security boundary and has **zero** uses under `src/agent/agents/`:

- `src/utils/terminal-sanitize.ts:63` — `export function sanitizeForDisplay(s: string): string`
- `git grep -c sanitizeForDisplay origin/main -- src/agent/agents/` → 0

## Threat model (why this is more than cosmetic)

Paths reaching these warnings come from `readdirSync` over four directories, and one of them is not operator-authored: `<cwd>/.claude/agents/` and `<cwd>/.afk/agents/` are **checked into the repository being worked in**. Cloning a hostile repo and starting a session in it is enough to route an attacker-chosen filename — ANSI escapes, cursor movement, carriage returns — into the operator's terminal via a scan warning. Filenames are constrained by the filesystem, not by the AFK parser, so the frontmatter-level escaping does not cover them.

Severity stays low: the payload is display-layer only (no command execution), and it requires the operator to open a session inside an untrusted checkout. But this is exactly the case `sanitizeForDisplay` exists to cover.

## Suggested fix

Route the path through `sanitizeForDisplay` at all five sites in one pass. Consider whether the scan-warning emitters should take pre-sanitized values at the boundary (i.e. sanitize `filePath` once when `collectMarkdownFiles` yields it) so future emission sites inherit the protection rather than each having to remember.

## Acceptance criteria

- [ ] All five `registry.ts` warn sites emit sanitized paths.
[...]

PR #976, opened 2026-08-09
fix(agents): surface a warning when a malformed agent file is discarded (#752)

### LINKED ISSUE #752: agents: malformed plugin agent files are silently discarded — parseAgentMarkdown's warn callback is never passed
Found while fixing #745 (PR #751). While tracing every producer of agent-registry warnings, one turned out not to be a producer at all: plugin agent files that fail to parse are discarded in complete silence, on every surface.

## What happens

`discoverPluginAgents` reads each candidate file and drops it on two paths without a word:

- `src/agent/tools/skill-bridge.ts:367-371` — unreadable file → `continue; // unreadable file — contained, skip`
- `src/agent/tools/skill-bridge.ts:372-373` — `parseAgentMarkdown(content)` → `if (parsed === undefined) continue; // malformed frontmatter — skip`

`parseAgentMarkdown` is built to report *why* it rejected a file. Its second parameter is a `warn` callback that **defaults to a no-op** (`src/agent/agents/parser.ts:99-102`):

```ts
export function parseAgentMarkdown(
  content: string,
  warn: (message: string) => void = () => {},
): ParsedAgentFile | undefined {
```

and it emits specific, actionable diagnostics — `'missing frontmatter (file must start with ---)'` (`parser.ts:104`), `'unterminated frontmatter (no closing ---)'` (`parser.ts:110`). `skill-bridge.ts:372` omits the argument, so every one of those messages goes to the default no-op and is lost.

## The asymmetry

The non-plugin scanner in the same codebase does this correctly. `src/agent/agents/registry.ts:182`:

```ts
const parsed = parseAgentMarkdown(content, (msg) => warn(`[afk] agents: ${filePath}: ${msg}`));
```

It also warns on the unreadable-file path (`registry.ts:179`), where `skill-bridge.ts` silently skips. So a malformed agent in `~/.afk/agents/` tells you what's wrong and names the file, while the byte-identical mistake in a plugin's `agents/` directory produces nothing at all.

This is doubly odd because the two scanners are *documented* as sharing traversal semantics — `registry.ts:81-83`:

> Exported so the plugin-agent scanner (`discoverPluginAgents`) reuses the exact same traversal semantics.

They share `collectMarkdownFiles`; they do not share the diagnostic contract.

## Why it matters

The failure is indistinguishable from success. A plugin author ships `agents/reviewer.md` with a typo'd frontmatter delimiter; `discoverPluginAgents` returns an array without it; `loadAgentRegistry` merges what it was given; `agent_type: 'my-plugin:reviewer'` then fails with "unknown agent" at dispatch time — nowhere near the actual cause, with no hint that a file was rejected at scan time. There is no debug flag or verbose mode that surfaces it, because the message was never constructed.

This lands on every surface, not one: `interactive` (`src/cli/commands/interactive/bootstrap.ts:338`), `chat` (`src/cli/commands/chat.ts:508`), `daemon` (`src/cli/commands/daemon.ts:127`), and `telegram` (`src/telegram.ts:311`).

## Distinct from #745

Worth stating plainly, since they were found together and both involve agent-registry warnings reaching nobody:

- **#745 / PR #751** — the warning *was* produced, written to stderr, then **erased** by the REPL's startup screen clear. REPL-only; fixed by re-routing the emission to after the clear.
- **This issue** — the warning is **never produced**. No surface, no clear involved. A fix here is upstream of #745's plumbing and would ride the same `bootWarnings` channel on the REPL once emitted.

## Suggested fix

Give `discoverPluginAgents` an optional `warn` seam, mirroring `loadAgentRegistry`'s (`registry.ts:76`), and thread it into `parseAgentMarkdown` with the file path prefixed — the same shape as `registry.ts:182`. Also warn on the unreadable-file path (`skill-bridge.ts:369`) so the two scanners finally match.

Callers can then pass a real writer. Note the REPL should route it into `ctx.bootWarnings` (added by PR #751) rather than printing directly, or the startup clear will eat it — i.e. this fix should land after #751, or coordinate with it.

Keep the current containment guarantee: one malformed file must never block discovery of the rest (`registry.ts:33-34`). This is about making the skip *audible*, not fatal.

## Acceptance criteria

- [ ] A malformed plugin agent file (missing/unterminated frontmatter) produces a warning naming the file path and the parser's reason.
- [ ] An unreadable plugin agent file warns rather than silently skipping.
- [ ] Diagnostics match the existing user/project-scope format at `registry.ts:182`.
- [ ] Discovery still skips only the offending file — the rest of the plugin's agents load.
- [ ] All four call sites keep working; the REPL's copy survives the startup clear (via `ctx.bootWarnings`, PR #751).
- [ ] A test pins the emission; `pnpm test` + `pnpm lint` green.

Related: #745 (same domain, opposite mechanism: erased vs. never emitted), #739 (added the built-in-shadow warning this class of silence undermines).

griffinwork40:
## The silent discard

`discoverPluginAgents` (`src/agent/tools/skill-bridge.ts`) reads each candidate plugin agent file and drops it on two paths without a word:

- unreadable file → `continue` (no warning)
- `parseAgentMarkdown(content)` returns `undefined` → `continue` (no warning)

`parseAgentMarkdown` is built to report *why* it rejected a file — its second parameter is a `warn` callback that **defaults to a no-op**. The old call site (`parseAgentMarkdown(content)`) never passed one, so every diagnostic (`'missing frontmatter (file must start with ---)'`, `'unterminated frontmatter (no closing ---)'`, etc.) fell into the default no-op and was lost. The non-plugin scanner in the same codebase (`loadAgentRegistry`'s `scanScope`, `registry.ts:182`) already threads `warn` through correctly — the plugin scanner just never matched that contract.

Net effect: a plugin author ships `agents/reviewer.md` with a typo'd frontmatter delimiter, the agent silently doesn't exist, and `agent_type: "my-plugin:reviewer"` later fails with "unknown agent" nowhere near the actual cause.

## Where the callback was dropped

`src/agent/tools/skill-bridge.ts` — `discoverPluginAgents`:
- the unreadable-file `catch` block used a bare `continue`
- `parseAgentMarkdown(content)` was called with no second argument

## The fix

- `discoverPluginAgents` now accepts an optional `warn` callback (defaults to a no-op, matching `parseAgentMarkdown`'s own default) and threads it into `parseAgentMarkdown(content, (msg) => warn(...))`, mirroring `registry.ts:182`'s shape.
- The unreadable-file path now also warns (`cannot read <path>: <error message>`), matching `registry.ts`'s handling of the same failure class.
- `src/agent/session/wire-executors.ts` — the one production call site — now passes the existing `agentRegistryWarn` option through: `discoverPluginAgents(undefined, agentRegistryWarn)`. This is the same sink `loadAgentRegistry` already receives, which the REPL routes to `ctx.bootWarnings` (PR #751) so it survives the startup screen clear. `chat.ts`, `daemon.ts`, and `telegram/session-anthropic.ts` all reach `discoverPluginAgents` only indirectly through `wireExecutors`, so this one wiring point covers every surface.
- Containment is unchanged: a malformed or unreadable file is still skipped via `continue`, never thrown, and never blocks discovery of the plugin's other agents.

This issue is scoped to the *emission* of the warning (never produced → now produced). It does not touch #745 (warning produced but erased by the REPL clear, already fixed in PR #751).

## Sanitization note (re: #747)

Plugin agent directories can live inside a cloned repository, so `filePath` is not fully operator-controlled. Per the issue's guidance and in the spirit of #747 (which hardens this same warning class across `registry.ts`), the path is passed through the repo's existing `sanitizeForDisplay` helper (`src/utils/terminal-sanitize.ts`) before interpolation, rather than embedded raw. This PR does **not** implement #747 itself (that issue covers `registry.ts`'s five pre-existing sites) — it just avoids introducing a sixth raw-path warning that would contradict it.

## Test evidence

Four new cases added to the `discoverPluginAgents` describe block in `src/agent/tools/skill-bridge.test.ts`:
1. A malformed file (no frontmatter) warns naming the file and the parser's specific reason; the call does not throw; a valid sibling agent in the same plugin still loads.
2. An unreadable file (`chmod 0o000`) warns with `cannot read <path>`; the valid sibling agent still loads.
3. A filename containing a raw ANSI/OSC escape sequence never reaches `warn()` unescaped — pins the `sanitizeForDisplay` routing.

## Verification

- `./node_modules/.bin/tsc --noEmit` → exit 0
- `./node_modules/.bin/vitest run src/agent/tools/skill-bridge.test.ts src/agent/agents/registry.test.ts src/agent/agents/parser.test.ts` → **100 passed**, 0 failed

## Confirmed still non-fatal

Both new tests assert the plugin's other, valid agent still loads after the broken file is discovered — the skip remains a `continue`, not a throw, in both the read-failure and parse-failure paths.

## Risks

- None expected to existing behavior for well-formed agent files. Warning volume increases only when a plugin actually ships a broken or unreadable agent file — previously invisible, now visible.
- Surfaces that don't wire `agentRegistryWarn` fall back to the no-op default (unchanged pre-existing behavior for those paths, since `discoverPluginAgents()` with no second argument is still valid).

## What I did not check

- Did not audit `src/agent/plugins-scanner.ts` or the skill-body scanner for the same missing-warn pattern — out of scope for this issue (that's plugin *skill* discovery, not agent discovery).
- Did not implement #747's fix to the five pre-existing `registry.ts` sites — only ensured this new call site doesn't add a sixth unsanitized one.
- Did not add an integration test exercising the REPL's `ctx.bootWarnings` end-to-end for a plugin agent (existing wiring test coverage for `agentRegistryWarn` already covers that this option reaches `loadAgentRegistry`'s `warn`; this PR extends the same option to also reach `discoverPluginAgents`).

Closes #752

### DIFF
--- a/src/agent/session/wire-executors.ts
+++ b/src/agent/session/wire-executors.ts
@@ -206,7 +206,10 @@ export function wireExecutors(opts: WireExecutorsOptions): WiredExecutors {
   //    dispatch at every depth.
   const agentRegistry = loadAgentRegistry({
     ...cwdOpt,
-    pluginAgents: discoverPluginAgents(),
+    // Same sink both scanners report through: a malformed plugin agent file
+    // now warns exactly like a malformed user/project one (#752) instead of
+    // vanishing silently ahead of the merge below.
+    pluginAgents: discoverPluginAgents(undefined, agentRegistryWarn),
     ...(agentRegistryWarn !== undefined ? { warn: agentRegistryWarn } : {}),
   });
 
--- a/src/agent/session/wire-executors.ts
+++ b/src/agent/session/wire-executors.ts
@@ -181,6 +181,10 @@ export function wireExecutors(opts: WireExecutorsOptions): WiredExecutors {
   const skillTraceOpt = skillTraceWriter !== undefined ? { traceWriter: skillTraceWriter } : {};
   const apiKeyOpt = apiKey !== undefined ? { apiKey } : {};
   const bgRegistryOpt = backgroundRegistry !== undefined ? { backgroundRegistry } : {};
+  // Match loadAgentRegistry's default sink so plugin discovery remains audible
+  // on non-interactive surfaces that do not provide a boot-warning collector.
+  const registryWarn =
+    agentRegistryWarn ?? ((message: string) => process.stderr.write(message + '\n'));
   // Session-static snapshot shared by all root executors and inherited by
   // descendants. Do not resolve inside execute(): sibling calls must not see
   // different caps if the process environment changes during the session.
@@ -209,8 +213,8 @@ export function wireExecutors(opts: WireExecutorsOptions): WiredExecutors {
     // Same sink both scanners report through: a malformed plugin agent file
     // now warns exactly like a malformed user/project one (#752) instead of
     // vanishing silently ahead of the merge below.
-    pluginAgents: discoverPluginAgents(undefined, agentRegistryWarn),
-    ...(agentRegistryWarn !== undefined ? { warn: agentRegistryWarn } : {}),
+    pluginAgents: discoverPluginAgents(undefined, registryWarn),
+    warn: registryWarn,
   });
 
   // 4. Shared by the `agent` and `skill` executors so plugin skill children
--- a/src/agent/tools/skill-bridge.test.ts
+++ b/src/agent/tools/skill-bridge.test.ts
@@ -16,7 +16,7 @@
  */
 
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
-import { mkdtempSync, writeFileSync, rmSync, mkdirSync, chmodSync } from 'node:fs';
+import { mkdtempSync, writeFileSync, rmSync, mkdirSync } from 'node:fs';
 import { join } from 'node:path';
 import {
   buildSkillManifest,
@@ -30,6 +30,18 @@ import { _resetPluginScanCache } from '../plugins-scanner.js';
 import { getPluginsDir } from '../../paths.js';
 import type { RegisteredAgent } from '../agents/types.js';
 
+const fsMocks = vi.hoisted(() => ({
+  readFileSync: vi.fn(),
+  realReadFileSync: undefined as typeof import('node:fs').readFileSync | undefined,
+}));
+
+vi.mock('node:fs', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('node:fs')>();
+  fsMocks.realReadFileSync = actual.readFileSync;
+  fsMocks.readFileSync.mockImplementation(actual.readFileSync);
+  return { ...actual, readFileSync: fsMocks.readFileSync };
+});
+
 // ---------------------------------------------------------------------------
 // File-level isolation: redirect AFK_HOME + cwd before every test so the
 // disk scan inside collectSkillEntries() always finds an empty skills dir.
@@ -41,6 +53,7 @@ let _isolatedCwd: string;
 let _origCwd: string;
 
 beforeEach(() => {
+  fsMocks.readFileSync.mockImplementation(fsMocks.realReadFileSync!);
   _isolatedAfkHome = mkdtempSync('/tmp/skill-bridge-test-afkhome-');
   _isolatedCwd = mkdtempSync('/tmp/skill-bridge-test-cwd-');
   _origCwd = process.cwd();
@@ -780,20 +793,23 @@ describe('discoverPluginAgents', () => {
     writeManifest(pluginA, 'demo');
     const dir = join(pluginA, 'agents');
     mkdirSync(dir, { recursive: true });
-    const unreadablePath = join(dir, 'locked.md');
+    const unreadablePath = join(dir, '\x1B]0;pwned\x07locked.md');
     writeAgentFile(pluginA, 'ok.md', 'works');
     writeFileSync(unreadablePath, '---\nname: locked\ndescription: d\n---\nbody\n');
-    chmodSync(unreadablePath, 0o000);
+    fsMocks.readFileSync.mockImplementation((path, options) => {
+      if (path === unreadablePath) {
+        throw new Error(`EACCES: permission denied, open '${unreadablePath}'`);
+      }
+      return fsMocks.realReadFileSync!(path, options);
+    });
 
     const warn = vi.fn();
-    try {
-      const agents = discoverPluginAgents([{ type: 'local', path: pluginA }], warn);
-      expect(agents.map((a) => a.name)).toEqual(['demo:works']);
-      expect(warn).toHaveBeenCalledWith(expect.stringContaining('cannot read'));
-      expect(warn).toHaveBeenCalledWith(expect.stringContaining(unreadablePath));
-    } finally {
-      chmodSync(unreadablePath, 0o644); // restore so afterEach's rmSync can clean up
-    }
+    const agents = discoverPluginAgents([{ type: 'local', path: pluginA }], warn);
+    expect(agents.map((a) => a.name)).toEqual(['demo:works']);
+    expect(warn).toHaveBeenCalledWith(expect.stringContaining('cannot read'));
+    const message = warn.mock.calls[0]?.[0] as string;
+    expect(message).not.toContain('\x1B');
+    expect(message).not.toContain('\x07');
   });
 
   it('sanitizes an escape-sequence-bearing file path before it reaches warn', () => {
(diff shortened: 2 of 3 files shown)


griffinwork40:
@codex fix comments

griffinwork40:
Finding 3 verified — two separate `toHaveBeenCalledWith` calls, each checking one substring independently. They both match the same single `warn()` call in practice, but the test structure doesn't enforce they're the same message.

Wave 1.5 complete. All citations verified, no fabrications, no false absence claims. All three findings survive at their assigned severity.

---

## Review — fix(agents): surface a warning when a malformed agent file is discarded (#752)

**Reviewed ref:** `198babf5fc580281dd473f930daf2e36b3b5d2b2`
**Change type:** hotfix · **Regime:** light · **Files:** 3 · **Lines:** +113/−6

---

### Findings

| # | Sev | Blocking | Dim | Finding |
|---|-----|----------|-----|---------|
| 1 | low | false | security | `scanScope` in `registry.ts` still interpolates raw `filePath` into warn messages — the sanitization added by this PR for the plugin path is one-directional |
| 2 | low | false | test-coverage | The unreadable-file test asserts `'cannot read'` and escape-sequence absence but never asserts the sanitized path appears in the message |
| 3 | low | false | correctness | The malformed-file test uses two independent `toHaveBeenCalledWith` matchers that happen to hit the same call — does not pin that path and parse-reason land in one message |

### Details

**1 · low · security · `src/agent/agents/registry.ts:180`** · ref: 198babf5 · citation-type: file-state [verified]

`scanScope` (the user/project agent scanner) interpolates raw `filePath` and raw `err.message` into `warn()` without `sanitizeForDisplay`. This PR correctly sanitizes the *plugin* path in `discoverPluginAgents`, but the non-plugin scanner it claims to "mirror" remains unsanitized. The PR body explicitly acknowledges this is out of scope (#747 covers the five pre-existing `registry.ts` sites), so this is informational — not scope creep.

```ts
// registry.ts:180 — no sanitizeForDisplay
warn(`[afk] agents: cannot read ${filePath}: ${err instanceof Error ? err.message : String(err)}`);
```

**Suggestion:** Address in #747 — no action needed in this PR.

---

**2 · low · test-coverage · `src/agent/tools/skill-bridge.test.ts:810–812`** · ref: 198babf5 · citation-type: diff-context [verified]

The unreadable-file test captures `warn.mock.calls[0]?.[0]` and asserts the escape bytes are absent, but never asserts that any sanitized remnant of the filename (`locked.md`) appears. A regression that omits the path from the message entirely would still pass. The sibling malformed-file test (line 789) does assert `expect(warn).toHaveBeenCalledWith(expect.stringContaining(brokenPath))` — the pattern exists but isn't applied consistently.

```ts
// Only negative assertions on the captured message:
expect(message).not.toContain('\x1B');
expect(message).not.toContain('\x07');
// No positive assertion like: expect(message).toContain('locked.md');
```

**Suggestion:** Add `expect(message).toContain('locked.md')` to pin the sanitized path residue.

---

**3 · low · correctness · `src/agent/tools/skill-bridge.test.ts:786–789`** · ref: 198babf5 · citation-type: diff-context [verified]

The malformed-file test uses two separate `toHaveBeenCalledWith` calls — one for the path, one for `'missing frontmatter'`. Both match the same single `warn` invocation (the `discoverPluginAgents` wrapper at `skill-bridge.ts:383` produces one interpolated string), but the test doesn't pin that structurally. If a future change split path and reason into separate `warn` calls, both matchers would still pass despite the format regression.

```ts
expect(warn).toHaveBeenCalledWith(
  expect.stringContaining('missing frontmatter (file must start with ---)'),
);
expect(warn).toHaveBeenCalledWith(expect.stringContaining(brokenPath));
```

**Suggestion:** Capture `warn.mock.calls[0]?.[0]` and assert both substrings on that single string (as the unreadable test already does for its own assertions).

---

### Dimension summary

| Dimension | Verdict |
|-----------|---------|
| **security** | Finding 1 (low, informational — out of scope per PR body; #747 tracks the fix) |
| **api-compat** | no issues found — `discoverPluginAgents` parameter is additive with a default |
| **correctness** | Finding 3 (low — test assertion structure, not production logic) |
| **spec-compliance** | no unmet intent or scope creep — all four stated requirements implemented |
| **test-coverage** | Finding 2 (low — missing positive path assertion in one of three tests) |
| **perf-observability** | no issues found — `sanitizeForDisplay` runs at startup-time scan frequency |

### What was not checked

- Citations verified inline against branch HEAD `198babf5`.
- Stated intent: PR #976 title+body — spec-compliance assessed.
- Did not review `src/utils/terminal-sanitize.ts` for completeness of its escape-stripping regexes (assumed correct per existing test coverage).
- Did not audit the plugin *skill* scanner (`discoverPluginSkillBodies`) for the same missing-warn pattern — the PR body explicitly calls this out of scope.

---

---

Issue #662, opened 2026-07-21
Subagents sandboxed to cwd silently fail to read paths outside cwd (forces re-dispatch)


griffinwork40:
## Gap

A dispatched subagent's filesystem access is confined to its `cwd`. When a task references an **absolute path outside that cwd** (e.g. a scratch/workspace/Downloads directory that holds the actual data), the subagent silently can't read it, and the orchestrator only finds out after a wasted dispatch — then has to re-dispatch with `cwd` (or `writeRoots`) explicitly anchored at the data directory.

## Impact

Whenever the relevant data lives **outside the repo** (common for analysis work: exports in `~/Downloads`, scratch under a workspace dir), the first dispatch dead-ends on reads. The failure isn't obvious up front — it looks like "file not found" rather than "outside your root" — so it costs a full round-trip to diagnose and retry.

## Repro / evidence

- A read-only `research-agent` dispatched to analyze data could not read the analysis workspace path; it had to be re-dispatched with `cwd` anchored at that directory before its reads resolved.
- In a later session the orchestrator pre-emptively set `cwd` on *every* verifier dispatch specifically to dodge this — i.e. the workaround has become a reflex, which is a smell.

## Proposed direction (options)

1. **Clearer, earlier error:** when a subagent read is denied for being outside its root, say so explicitly — name the active root and suggest passing `cwd`/`writeRoots` — instead of a generic not-found.
2. **Docs:** call out the "set `cwd` (and `writeRoots`) when your data lives outside the repo" pattern in the `agent`-tool guidance.
3. **Optional mechanism:** a read-roots analog to `writeRoots` (or auto-include absolute paths explicitly referenced in the dispatch prompt) so the orchestrator can grant the needed read scope without repointing `cwd`.

Related: #441 (subagent robustness follow-ups).

_Surfaced during a multi-agent data-analysis session (2026-07-20) and confirmed again in a 2026-07-21 audit._

PR #978, opened 2026-08-09
fix(subagent): refuse model-supplied roots that erase the bash credential floor (#852)

### LINKED ISSUE #852: Model-supplied readRoots can silently lift the bash credential floor for a child
## Summary

A model-originated `readRoots` grant on the `agent` tool can silently erode the **bash** credential-restriction floor for that child. Typed reads stay protected; bash does not.

Found while investigating why confined forks cannot read `/tmp` (→ #851). Not exploited, not urgent, but it is a real asymmetry with no human in the loop.

## Mechanism

1. `deriveRestrictedSubstrings` (`src/agent/tools/hooks/bash-restriction-hook.ts`, ~:555-575) drops any credential candidate that a *granted root* is an ancestor of. This is deliberate — the floor is documented as grant-liftable (~:529-539) — on the assumption that a grant reflects operator intent.
2. But a **model** can create such a grant with no operator involvement: `agent(prompt, readRoots: ["<dir>"])` lands as constructor state on the child dispatcher (`src/agent/tools/dispatcher.ts:365-366`, via `src/agent/subagent.ts:381,429`).
3. Concrete path: a model self-grants `~/Library/Application Support`. That directory is **not** read-denied (only per-browser subtrees are — `src/agent/tools/handlers/read-denylist.ts:99-144`) and **not** too-broad (`isTooBroadRoot` only anchors on filesystem root / home / AFK dirs — `src/agent/tools/subagent/root-validation.ts:79-85`). The child's bash `cat` gate over that whole vendor tree, browser secrets included, then silently disappears.

## What is NOT affected

Typed reads are fine, and this is worth stating so the scope stays honest:

- `input-parse.ts:400-406` rejects denylisted `readRoots` grants at parse time.
- `isReadDenied` is unconditional at read time (`read-denylist.ts:7-10`, `:395-422`) — it applies even in `allowAll` bypass mode.

So this is specifically a **bash-floor** issue, not a typed-read containment issue.

Also unaffected: a `/tmp` subdirectory grant erodes nothing, because no denylist or bash candidate lives under `/tmp`. The narrow use case is safe; the general field is the hazard.

## Why the obvious fix is wrong

The tempting fix — make `deriveRestrictedSubstrings` provenance-aware so only operator (`'slash'`) grants lift a candidate — **is inverted and would make things worse.** Verified:

1. **Provenance does not survive to the point of use.** `GrantSnapshot` is `{resolveBase, readRoots: string[], writeRoots: string[], allowAll}` — bare strings (`src/agent/tools/grant-manager.ts:56-61,180-187`). The `source` argument is consumed by `appendAuditLog` and then discarded; roots are pushed as plain strings (`:126-128`). The bash hook consumes exactly that snapshot (`bash-restriction-hook.ts:210`).
2. **The `'slash'`/`'tool'` axis does not mean what it looks like.** Every `'tool'` call site is *operator*-originated — the once/session/persist elicitation approvals (`path-approval-hook.ts:584,586,601,603,615,617`) and persisted-permission restore (`permissions-store.ts:275,278`). Gating on `'tool'` would revoke the operator's own "approve `~/.ssh` for this session" choice.
3. Meanwhile the actual model vector never passes `addReadRoot` at all, so there is no provenance to inspect on the one path that matters.

Net: that fix blocks the operator and misses the model.

## Directions worth considering instead

- Introduce a real provenance field on the grant snapshot (`operator` vs `model`) threaded to the point of use — requires widening `GrantSnapshot` and every implementer (`dispatcher.ts:368-377,427-437`, `anthropic-direct/index.ts:378-383`, `openai-compatible/index.ts:622-627`, plus the structural interfaces at `allow-dir.ts:28-33` and `permissions-store.ts:268-271`).
- Or: make the bash floor **non-liftable by model-supplied roots specifically**, by tagging the additive `extraReadRoots` union (`subagent.ts:407-431`) rather than retrofitting provenance across all grants. Narrower.
- Or: extend the read-denylist to cover `~/Library/Application Support` more broadly, which shrinks the blast radius without touching the grant model at all. Cheapest, least general.

No recommendation yet — this needs design, which is why it is an issue and not a PR.

## Repro sketch (not automated)

Dispatch a child with `readRoots: ["<HOME>/Library/Application Support"]`, then have it run a bash `cat` against a browser credential file under that tree, and compare against the same dispatch without the grant. Expect the restriction to be present in the second case and absent in the first.

griffinwork40:
Closes #852.

## The hole

`deriveRestrictedSubstrings` drops any bash credential candidate that a granted root is an ancestor of. That is deliberate and documented — an explicit `/allow-dir ~/.ssh` is the operator saying they want that path this session, and the typed-read floor stays unconditional regardless. The problem is that a **model** reaches the same filter with no human in the loop, through the `agent` tool's own grant fields.

`~/Library/Application Support` is the canonical vector, and it slips both existing guards:

| Guard | Why it passes the vector |
|---|---|
| `isTooBroadRoot` | anchors only on `/`, home, ancestors of home, and the AFK dirs |
| `isReadDenied` | only the per-browser subtrees *beneath* it are denylisted, not the dir itself |

So the grant lands, and the child's bash `cat` gate over that whole vendor tree — browser secrets included — silently disappears.

## Why not the three directions in the issue

The issue proposed (1) a real provenance field on `GrantSnapshot`, (2) tagging the `extraReadRoots` union, or (3) widening the read-denylist. It also showed why the *obvious* variant of (1) is inverted: every `'tool'`-sourced grant is actually operator-originated (elicitation approvals, persisted-permission restore), so gating on the `'slash'`/`'tool'` axis would revoke the operator's own choices — while the actual model vector never calls `addReadRoot` at all, so there is no provenance there to inspect.

That analysis holds, and it points at a fourth option the issue did not list: **provenance is already structural at the parse boundary.** `parseAgentInput` sees *only* model-authored agent-tool input. Every operator grant path — `/allow-dir`, the path-approval elicitation approvals, persisted-permission restore — reaches the grant manager without passing through it.

So the check goes there, and no snapshot field is needed. The operator's documented lift keeps working unchanged; only the model loses the ability to lift the floor for itself.

## The change

A new `ungatedSensitiveRoot(candidate)` in `root-validation.ts`, beside the `isTooBroadRoot` it complements, returning *which* credential root a grant would un-gate (so the error can name it and the model can narrow the grant rather than guess).

Applied to all three grant fields, matching that module's existing "the rejection must be identical across all three call sites" contract. **`cwd` included** — it becomes the child's `resolveBase`, which `deriveRestrictedSubstrings` seeds into `granted` alongside the explicit roots, so an ancestor `cwd` lifts the floor by exactly the same route.

Two details worth review attention:

- **No second list of sensitive roots.** The guard calls the hook's own exported `deriveRestrictedSubstrings` with an empty snapshot (empty ⇒ its filter can drop nothing ⇒ the full candidate set). A root added to the bash floor later is therefore un-grantable automatically. Two lockstep property tests pin the coupling in *both* directions: everything the guard accepts drops nothing from the floor, and every root the floor protects is refused as a grant.
- **Containment direction.** A grant *inside* a sensitive tree is not rejected — it does not lift the enclosing root, and the denylisted subtrees are already refused by `isReadDenied`. Only at-or-above is refused. Checked on the symlink-resolved spelling as well as the lexical one, same as `isTooBroadRoot` (#664).

## Blast radius

Newly refused: `~/Library`, `~/Library/Application Support`, `~/.config`, `~/.afk`, `/etc` — each a genuine ancestor of a credential root.

Still accepted (verified by test, and the reason the guard is ancestor-only rather than a blanket denial):

- ordinary out-of-repo grants — `/tmp/...`, `~/Downloads`, repo paths (the #662 use case)
- narrower paths *inside* a sensitive tree
- `~/.afk/state` and `~/.afk/agent-framework` — the deliberately-grantable dirs confined forks depend on (the Gap A / Gap C grants in `forkSubagent`); neither is an ancestor of `~/.afk/config`
- not-yet-created paths, which `realpathSafe` re-appends rather than collapsing to their nearest existing ancestor

## Verification

- Full suite green: **834 files / 15,870 tests**, 0 failures.
- All CI gates pass: `lint`, `audit:sdk:check`, `audit:env:check`, `scan:env:check`, `audit:chalk:check`, `fix:pins:check`, `build`.
- **Negative control** — reverted the two source files with the new tests left in place: the 6 rejection tests fail (so they capture the real vulnerability, not a tautology) while the "does not over-reject" tests still pass. Restored, re-verified green.
- +19 tests: 9 at the parse boundary, 10 in a new `root-validation.test.ts` (that module had no direct unit coverage before).

## Out of scope, worth a follow-up

`SubagentDAGNode.readRoots` flows straight into `forkSubagent` (`dag-subagent.ts:115-126`) without passing through `parseAgentInput`, so it bypasses this guard — **and equally bypasses the pre-existing `isTooBroadRoot`**, so this is not a regression introduced here. It is also not model-reachable today: the `compose` tool's schema exposes no `readRoots`, so only in-process skill code can populate it. Filing separately rather than widening this PR.

### DIFF
--- a/src/agent/tools/hooks/bash-restriction-hook.test.ts
+++ b/src/agent/tools/hooks/bash-restriction-hook.test.ts
@@ -461,6 +461,38 @@ describe('deriveRestrictedSubstrings — no coverage regression from sharing the
   });
 });
 
+describe('deriveRestrictedSubstrings — a cross-drive grant cannot empty the floor (#852)', () => {
+  // Invariant: `path.win32.relative` returns a DRIVE-QUALIFIED ABSOLUTE string
+  // (`C:\Users\me\.ssh`) when the two paths sit on different drives, never a
+  // `..\`-prefixed one. So `!rel.startsWith('..')` alone reads a cross-drive
+  // pair as "the grant covers this candidate" and DROPS it — one grant on an
+  // unrelated drive emptied the entire credential floor, reachable by any
+  // model-supplied readRoots entry. `!path.isAbsolute(rel)` is the fix.
+  //
+  // Gated to win32 because the topology is UNREACHABLE on POSIX: relative()
+  // between two absolute POSIX paths is never absolute. That unreachability is
+  // why the bug survived review, and it is why the platform-independent guard
+  // is the predicate-identity test in subagent/root-validation.test.ts — this
+  // case does not run on the default CI matrix (the windows leg is opt-in).
+  it.runIf(process.platform === 'win32')(
+    'keeps every candidate when the only grant sits on another drive',
+    () => {
+      const unfiltered = deriveRestrictedSubstrings({
+        resolveBase: undefined,
+        readRoots: [],
+        writeRoots: [],
+      });
+      const otherDrive = homedir().toUpperCase().startsWith('D:') ? 'C:\\scratch' : 'D:\\scratch';
+      const withCrossDriveGrant = deriveRestrictedSubstrings({
+        resolveBase: undefined,
+        readRoots: [otherDrive],
+        writeRoots: [],
+      });
+      expect(withCrossDriveGrant).toEqual(unfiltered);
+    },
+  );
+});
+
 describe('createBashRestrictionHook — credential parity with the typed read denylist', () => {
   // Invariant: a credential path floored for read_file / grep / glob is floored
   // for bash too. Each path below was readable with `cat` while the typed tools
--- a/src/agent/tools/hooks/bash-restriction-hook.ts
+++ b/src/agent/tools/hooks/bash-restriction-hook.ts
@@ -567,10 +567,22 @@ export function deriveRestrictedSubstrings(grants: {
   // so granting a narrow subdir (e.g. ~/Library/Application Support/Cursor/User)
   // wrongly un-gated the whole sensitive parent (~/Library/Application Support)
   // and every sibling app dir under it.
+
+  // Invariant: `!rel.startsWith('..')` alone is NOT a sufficient coverage test
+  // on Windows. When `g` and `c` sit on different drives, `path.win32.relative`
+  // returns a DRIVE-QUALIFIED ABSOLUTE string (`C:\Users\me\.ssh`) rather than a
+  // `..\`-prefixed one, which the bare check reads as "g covers c" and drops the
+  // candidate — so a single cross-drive grant (`readRoots: ['D:\\scratch']`)
+  // silently emptied the ENTIRE credential floor. `!path.isAbsolute(rel)` is the
+  // same guard `computeContainment` carries for this exact class (see
+  // handlers/_cwd-utils.ts), and it makes this predicate byte-identical to
+  // `ungatedSensitiveRoot`'s (subagent/root-validation.ts) — that identity is
+  // what the #852 lockstep property rests on, so the two must not drift again.
+  // No-op on POSIX: relative() between two absolute paths is never absolute.
   return candidates.filter((c) => {
     for (const g of granted) {
       const rel = path.relative(g, c);
-      if (rel === '' || !rel.startsWith('..')) return false;
+      if (rel === '' || (!rel.startsWith('..') && !path.isAbsolute(rel))) return false;
     }
     return true;
   });
--- a/src/agent/tools/subagent/input-parse.test.ts
+++ b/src/agent/tools/subagent/input-parse.test.ts
@@ -24,6 +24,11 @@ afterEach(() => {
   vi.unstubAllEnvs();
 });
 
+/** Escape a literal path for embedding in a `toThrow` RegExp. */
+function escapeRe(s: string): string {
+  return s.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
 describe('parseAgentInput', () => {
   describe('input shape', () => {
     it('throws when input is not an object (string)', () => {
@@ -585,6 +590,17 @@ describe('parseAgentInput', () => {
       const result = parseAgentInput({ prompt: 'p', cwd: '/tmp/wt/feat-y' });
       expect(result.cwd).toBe('/tmp/wt/feat-y');
     });
+
+    // --- Hardening: ancestor-of-credential rejection (#852) ---
+    // cwd becomes the child's resolveBase, which deriveRestrictedSubstrings
+    // seeds into `granted` alongside the explicit roots — so an ancestor cwd
+    // lifts the bash floor exactly as a readRoots grant would.
+    it('throws when cwd is an ancestor of a bash credential root (#852)', () => {
+      const appSupport = path.join(os.homedir(), 'Library', 'Application Support');
+      expect(() => parseAgentInput({ prompt: 'p', cwd: appSupport })).toThrow(
+        /must not be at or above a credential root/,
+      );
+    });
   });
 
   describe('writeRoots', () => {
@@ -698,6 +714,17 @@ describe('parseAgentInput', () => {
       });
       expect(result.cwd).toBe('/tmp/wt/x');
       expect(result.writeRoots).toEqual(['/sibling/repo']);
+    });
+
+    // --- Hardening: ancestor-of-credential rejection (#852) ---
+    // writeRoots feed the same `granted` set in deriveRestrictedSubstrings that
+    // cwd and readRoots do, so leaving this field unchecked would just relocate
+    // the erosion rather than close it.
+    it('throws when an entry is an ancestor of a bash credential root (#852)', () => {
+      const appSupport = path.join(os.homedir(), 'Library', 'Application Support');
+      expect(() => parseAgentInput({ prompt: 'p', writeRoots: [appSupport] })).toThrow(
+        /must not be at or above a credential root/,
+      );
     });
   });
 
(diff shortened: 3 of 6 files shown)


griffinwork40:
Filed the DAG-path follow-up noted in the PR description as #982 — `SubagentDAGNode`'s cwd/readRoots/writeRoots reach `forkSubagent` without `parseAgentInput`, so they skip both this guard and the pre-existing `isTooBroadRoot`. Not model-reachable today (the compose schema exposes no such fields), so it stays out of this PR.

griffinwork40:
All citations verified at the ref; both absence claims confirmed. Synthesizing inline (Wave 2 dispatch skipped — re-dispatching after two 45-min timeouts would risk a third for 3 findings I derived myself).

---

# Code Review — PR #978

**`fix(subagent): refuse model-supplied roots that erase the bash credential floor (#852)`**
`afk/fix-issue-852` → `main` · 4 files · +366/−10 · regime **full** · ref **`8df241e5`**

---

## security

**`medium` · blocking:true · confidence:high · security · `src/agent/tools/subagent/root-validation.ts:147` · ref:8df241e5 · citation-type:file-state**

The guard's containment predicate is **not** the same predicate as the filter it defends, and the two diverge exactly at a Windows drive boundary — so on win32 a model-supplied root on any other drive is *accepted* by the guard while still *emptying* the entire bash credential floor, leaving the #852 vector open on that platform.

```
guard   root-validation.ts:147    if (rel === '' || (!rel.startsWith('..') && !isAbsolute(rel))) return sensitive;
filter  bash-restriction-hook.ts:573  if (rel === '' || !rel.startsWith('..')) return false;
```

Executed probe (Node's `path.win32` namespace), `granted = D:\scratch`, `candidate = C:\Users\me\<credroot>`:

```
rel = "C:\\Users\\me\\CREDROOT"   startsWith("..") = false   isAbsolute = true
FILTER drops candidate   = true    <- bash floor EMPTIED
GUARD  rejects the grant = false   <- grant ACCEPTED
LOCKSTEP HOLDS           = false
```

This directly falsifies the PR body's central claim — *"everything the guard accepts drops nothing from the floor"* — on win32. The repo already documents this exact bug class at length in `_cwd-utils.ts:computeContainment` (`path.win32.relative` returns a drive-qualified **absolute** string across drives) and fixed it there; `isAbsolute` appears **zero** times in `bash-restriction-hook.ts` (grep-confirmed at ref). Windows is a supported install target (`engines: node>=22`, no `os` restriction) with an opt-in `test-windows` CI leg (`.github/workflows/ci.yml:170-176`).

**Not a regression** — the filter has always lacked the check; merging does not worsen the pre-PR state. It is an *incompleteness of the new guard*, and the test that is supposed to prove completeness cannot detect it (see below).

**suggestion:** add `&& !path.isAbsolute(rel)` to `bash-restriction-hook.ts:573`. That closes the cross-drive un-gating at source *and* makes the guard's existing predicate exactly correct, restoring the lockstep property with no change to `root-validation.ts`.

*Per the merge-decision rule, a `medium` in the `security` dimension is never overridable to `blocking: false`.*

**`no issues found`** on: error-message disclosure (the messages echo only the path the caller itself supplied — `input-parse.ts:298,359,457`); import cycle (`bash-restriction-hook` → `read-denylist` / `afk-home-refs` / `env` / `fs-case`; nothing in that graph imports `subagent/` — grep-confirmed).

## api-compat

**`no issues found`** — `ungatedSensitiveRoot` is a purely additive export; the two source files' only other change is JSDoc. No signature, return-type, or call-contract change to `isTooBroadRoot`.

## correctness

**`no issues found`** — read `bash-restriction-hook.ts:467-577`, `read-denylist.ts:63-272`, `_cwd-utils.ts:25-36`. Every candidate is an absolute path (`parseReadDenylistEntries` ends `.map(p => resolve(p))`, `getReadDenylist()` is `safeRealpath`-resolved, `builtinBashSensitiveRoots()` uses `path.join(homedir(), …)`) — so the feared `path.relative(form, <relative-fragment>)` cwd-dependence **does not occur** despite the "substrings" name. The empty snapshot genuinely yields the unfiltered superset (`granted` is empty ⇒ the filter's inner loop never runs). Derivation is per-call, not hoisted. Containment direction is correct on POSIX. `~/.afk/workspace` stays grantable (`relative('~/.afk/workspace','~/.afk/config') === '../config'`), as does `~/.afk/state`.

## spec-compliance

**`no issues found`** — every claim in the stated intent verified against source:

| Intent claim | Verdict |
|---|---|
| Applied to all three grant fields | ✅ `input-parse.ts:298` (cwd), `:359` (writeRoots), `:457` (readRoots) |
| No second list of sensitive roots | ✅ `root-validation.ts:107-109` calls the hook's own deriver |
| Grants *inside* a sensitive tree still accepted | ✅ predicate is ancestor-only |
| Not-yet-created paths still work | ✅ `_cwd-utils.ts:25-36` re-appends trailing segments |
| DAG `readRoots` bypass is **not model-reachable** | ✅ **verified at both layers**: `schemas.ts:522-528` exposes only `id`/`prompt`/`model` with `additionalProperties: false`, *and* `compose-executor.ts:252` does explicit field extraction (`parsed.push({ id, prompt, model })`), so unknown keys are structurally dropped even if the API does not enforce the schema |
| `parseAgentInput` is the sole model-authored path | ✅ single caller `subagent-executor.ts:492`; the other 15 `forkSubagent` callers (skills, mint phases, `user-skills.ts`, `audit-fit`) derive roots from session state, not model input |

No scope creep found.

## test-coverage

**`low` · blocking:false · confidence:high · test-coverage · `src/agent/tools/subagent/root-validation.test.ts:109,131` · ref:8df241e5 · citation-type:file-state**

Neither "lockstep" test can detect the predicate divergence above, because both only ever compare a root against **itself** or against an **unrelated** root — never a granted-root/candidate pair where the two predicates disagree. The converse test (`:131`) is additionally true by construction: it iterates `allCandidates()` and asserts `ungatedSensitiveRoot(c)` is defined, but the guard derives its list from the *same* function, so `relative(c,c) === ''` always fires. It catches one real drift (someone replacing the shared deriver with a hand-written list) and nothing else.

**evidence:** `for (const candidate of allCandidates()) { expect(ungatedSensitiveRoot(candidate)).toBeDefined(); }`
**suggestion:** add a case asserting the two predicates agree for a granted root that is *neither* an ancestor nor a descendant of the candidate — the shape that would have caught the win32 gap. Grep-confirmed absent: zero `win32`/`drive`/cross-drive coverage in either new test file or in `bash-restriction-hook.test.ts` at the ref.

## perf-observability

**`no issues found`** — `ungatedSensitiveRoot` runs once per grant entry on the `agent`-tool parse path (not per bash command): one `realpathSafe` syscall plus ~O(40×2) `path.relative` calls against a ~25–40 entry candidate list, with `getReadDenylist()` memoized. No load evidence exists and none is needed; this is not a hot path. Rejections reach the operator as errored tool calls in the witness trace.

## Other

**`low` · blocking:false · confidence:high · correctness · `src/agent/tools/subagent/root-validation.ts:139-140` · ref:8df241e5 · citation-type:file-state**

The `Invariant:` block justifies the symlink-resolution leg with *"because the containment layer realpaths granted roots before comparison (#664)"*, but the layer this guard actually defends — `deriveRestrictedSubstrings` — compares the **raw** granted string with no realpath (`bash-restriction-hook.ts:572`). Only `computeContainment` (typed tools) realpaths roots. The behavior is fail-closed and desirable; the stated rationale names the wrong layer, which matters in a file whose comments are load-bearing security documentation.
**suggestion:** reword to cite the typed-tool containment layer (`_cwd-utils.ts` `realpathRoot`) that the same grant also flows into, rather than the bash filter.

---

## What was not checked

- **Citations verified inline against branch HEAD `8df241e59c9c433da0fa141ad0ed0c069b4abb86`** (= the checked-out worktree HEAD, clean). Check A: 11/11 citations `verified`, 0 `fabricated`, 0 `diff-only`. Check B: 2/2 absence claims `confirmed-absent`.
- **Stated intent: PR #978 title + body — spec-compliance assessed.**
- **Wave 1 delegation failed.** Both dimension sub-agents (`rev-sec`, `rev-corr`) timed out at 45 min and returned nothing usable beyond one partial line. Wave 1 was re-run **inline by the orchestrator** using the same read-only shell. Consequence: no independent second reader on these findings — every claim above is single-derived, though each is backed by an executed probe or a `git show` at the ref.
- The win32 divergence was derived from Node's `path.win32` namespace **executed on darwin**, not on a real Windows host. Filesystem-level behavior on an actual Windows runner is unverified.
- Did not run the test suite; the author's claimed 834-file / 15,870-test green run, the CI-gate sweep, and the negative-control experiment are **unverified**.
- Did not audit the browser-subtree denylist contents, the `AFK_READ_DENYLIST` operator-extras path end-to-end, or `isReadDenied`'s carve-out gating.
- `/shadow-verify` **not** invoked: no `critical`/`high` findings and no finding departs from the default `blocking` table, so the post-synthesis routing rule does not fire.
- **Process disclosure:** I wrote one scratch probe to `/tmp/afk-review-978-win32-probe.mjs` (outside the repo) to execute the predicate comparison — a technical departure from this skill's "create no files" constraint. The repository, the branch, and the PR were not touched.

---

### Decision: DO NOT MERGE — 1 medium blocking (1 security); 0 waived, 2 low.

The change is well-constructed and closes #852 on POSIX; the single blocker is a one-line fix at `bash-restriction-hook.ts:573` that also restores the lockstep property the PR claims.

## Done

- Reviewed PR #978 end-to-end: resolved target and stated intent from `gh`, triaged to full regime, ran the dimension analysis, verified every citation and absence claim against ref `8df241e5`, and emitted the merge decision.
- Evidence: verdict and findings in this response; preflight diff at `/Users/griffinlong/.afk/state/skill-preflight/a26b8741-9ae3-4a8a-b88d-027b1a02e260/pr-978.diff`; reproducible win32 probe at `/tmp/afk-review-978-win32-probe.mjs`; all cited lines resolvable via `git show 8df241e5:<file>`.
- World changed: nothing — read-only review. No files in the repo were created, edited, or deleted; no commit, push, or PR mutation; no comment posted to #978. The only write was the `/tmp` scratch probe named above.
- Pending: the review is **not** posted to GitHub — surfacing it here only, per this skill's read-only contract. Posting it, and fixing the finding, both require a separate explicit instruction.
- Deferred with cause: Wave 1's two sub-agents timed out at 45 min, so the findings are single-derived rather than independently cross-read; `/shadow-verify` did not fire because no finding is `critical`/`high` or off-default disposition.

---

PR #979, opened 2026-08-10
fix(observability): surface first artifact-write failure per subsystem without AFK_DEBUG (#850)

### LINKED ISSUE #850: Observability failures are invisible unless AFK_DEBUG=1
## Problem

The observability layer cannot report its own failure. A subsystem that fails 100% of the time is indistinguishable from one that is working.

Failures on the witness/forensics paths are reported only through `debugLog`, which no-ops unless a debug env var is set:

```ts
// src/utils/debug.ts:13-15, 21-25
export function isDebugEnabled(): boolean {
  return env.AFK_DEBUG === '1' || env.DEBUG === '1';
}

export function debugLog(...args: unknown[]): void {
  if (isDebugEnabled()) {
    console.log(...args);
  }
}
```

So an operator who deliberately turns an observability feature **on** gets silence whether it works or not. They only discover the failure later, when they go looking for evidence that was never written — which is exactly the moment the evidence is needed and no longer recoverable.

## Concrete instance

The prompt capture added in #844 swallows every failure into `debugLog`:

```ts
} catch (err) {
  debugLog(`subagent-prompt-capture failed: ${String(err)}`);
}
```

Reachable failure modes that produce total, silent no-op:

- `EACCES` / `ENOSPC` / `EROFS` on the `mkdir`
- `validateSessionId` throwing on a session label that does not match `/^[a-zA-Z0-9_-]+$/` (via `getPromptsDir` -> `getTraceDir`)

In every case the operator enabled a flag, saw no error, and got no artifacts.

## Why this is not a #844 fix

The same swallow pattern runs through the wider witness path (`emitSessionPhase`, trace writes). Fixing it in one module would make the behavior inconsistent across the subsystem. This is a subsystem-level decision.

The fire-and-forget contract itself is correct and should be preserved: a forensics artifact must never be able to fail a turn. The question is only how failure becomes *visible* without becoming *fatal*.

## Suggested direction

- Surface the **first** failure per session per subsystem to stderr unconditionally, then fall back to `debugLog` for the rest — one line, not a log flood.
- Or expose a terminal-state counter (for example in `afk trace show`) reporting artifacts attempted vs. written, so a zero-write session is visible after the fact.
- Keep the never-throw contract intact either way.

## Acceptance

- Enabling an observability feature that then fails produces at least one operator-visible signal without `AFK_DEBUG=1`.
- No turn can be failed by an artifact write.
- A test asserts a forced write failure emits the signal exactly once.

---
Filed from a review of #844.

griffinwork40:
## Problem

The observability layer could not report its own failure — a subsystem failing 100% of the time was indistinguishable from one that works. Every witness/forensics swallow site logged failures only via `debugLog` (`src/utils/debug.ts`), which no-ops unless `AFK_DEBUG=1`/`DEBUG=1`. An operator who deliberately turns an observability feature **ON** got total silence whether it worked or was completely broken, and found out only when going looking for evidence that was never written — exactly when it was no longer recoverable.

Reachable total-silent-no-op modes:
- `EACCES` / `ENOSPC` / `EROFS` on `mkdir`
- `validateSessionId` throwing on a session label not matching `/^[a-zA-Z0-9_-]+$/` (reached via `getPromptsDir`/`getSubagentOutputsDir` → `getTraceDir`, `src/paths.ts`)

**The fire-and-forget contract is preserved and untouched: a forensics artifact must never be able to fail a turn.** This change only changes how a swallowed failure becomes *visible* — it does not change what gets swallowed or when.

## Why subsystem-level, not a #844 fix

#844 named one instance (`subagent-prompt-capture.ts:174`). Patching only that site would leave the other 13 swallow sites — the entire trace emitter plus subagent output capture — silently inconsistent with it: an operator would get a visible signal from one observability feature and dead silence from another, for the identical failure mode. This PR routes **all 14** existing swallow sites through one shared reporter so behavior is uniform across the whole witness layer.

## What changed

New module: **`src/utils/artifact-failure-reporter.ts`**

`reportArtifactFailure(subsystem, dedupKey, context, err)`:
- Surfaces the **first** failure for a given `(subsystem, dedupKey)` pair to stderr **unconditionally** via plain `console.error` — no chalk/palette import. This follows the existing precedent in `src/agent/subagent-hooks.ts:178` (the SubagentStop-timeout warning), not `src/cli/commands/interactive/boot-warnings.ts`, because that module is REPL/TUI-only and unusable from a headless daemon/chat surface that has no compositor.
- Every **subsequent** failure for the same pair falls back to `debugLog` — a wedged subsystem (every call failing identically) warns **once**, not once per call, avoiding a log flood.
- Never throws — a diagnostic must not become the failure it reports (same contract as `warnAfkHomeRejectedOnce` in `agent/tools/afk-home-warn.ts`).

### Swallow sites converted (14)

| File:line | Function / catch block |
|---|---|
| `src/agent/trace/emit.ts:45` (now ~52) | `emitToolCall` |
| `src/agent/trace/emit.ts:57` (~64) | `emitHookDecision` |
| `src/agent/trace/emit.ts:69` (~76) | `emitSubagentLifecycle` |
| `src/agent/trace/emit.ts:81` (~88) | `emitBackgroundAgent` |
| `src/agent/trace/emit.ts:93` (~100) | `emitBudget` |
| `src/agent/trace/emit.ts:105` (~112) | `emitAbort` |
| `src/agent/trace/emit.ts:117` (~124) | `emitCompaction` |
| `src/agent/trace/emit.ts:129` (~136) | `emitClosure` |
| `src/agent/trace/emit.ts:141` (~148) | `emitClaim` |
| `src/agent/trace/emit.ts:153` (~160) | `emitBrowserEvent` |
| `src/agent/trace/emit.ts:165` (~172) | `emitQueuedUserMessage` |
| `src/agent/trace/emit.ts:177` (~184) | `emitSessionPhase` |
| `src/agent/session/subagent-prompt-capture.ts:174` | `captureSubagentPrompt` (the #844 instance) |
| `src/agent/session/subagent-output-capture.ts:189, 256, 264` | `write` / `observe` / `end` in `createSubagentOutputRecorder` |

### Dedup key

The `emit*` functions take `(writer, payload)` — no `sessionId` parameter — so a "per session" key couldn't naively reuse an existing argument without threading a new parameter through 11 call sites, plus every one of their many callers. `TraceWriter.getTracePath()` (`src/agent/trace/writer.ts:70`) was already exposed publicly and is unique per session/writer instance, so it was used directly as the dedup key instead — zero signature changes to `emit*`.

The two capture modules already carry `input.sessionId` on their input type, so they use that directly.

## Acceptance criteria

| # | Criterion | Test |
|---|---|---|
| 1 | Enabling an observability feature that then fails produces at least one operator-visible signal **without** `AFK_DEBUG=1` | `src/utils/artifact-failure-reporter.test.ts` → `'surfaces the first failure to console.error unconditionally, without AFK_DEBUG'`; `src/agent/trace/emit.test.ts` → `'surfaces a write failure via console.error with AFK_DEBUG unset'` |
| 2 | No turn can be failed by an artifact write — never-throw contract intact | `src/agent/trace/emit.test.ts` → `'never rejects even when the writer throws on every call (never-throw contract)'`; existing suites `subagent-prompt-capture.test.ts` (`'never rejects on a bad session id'`) and `subagent-output-capture.test.ts` (`'never throws on a malformed event'`) still pass unchanged, now routed through the new reporter |
| 3 | A forced write failure emits the signal **exactly once** — not zero, not once per call | `src/utils/artifact-failure-reporter.test.ts` → `'surfaces only the FIRST failure per (subsystem, dedupKey); later failures fall back to debugLog'`; `src/agent/trace/emit.test.ts` → `'emits the visible signal exactly once per writer even across many failing calls'` |

`src/utils/artifact-failure-reporter.test.ts` also covers: the never-throw guarantee even if `console.error` itself throws, and that the latch keys on `(subsystem, dedupKey)` jointly (a different session or a different subsystem still gets its own first warning).

## Not done (scope cut, flagged rather than silently dropped)

An attempted-vs-written counter for `afk trace show` was considered per the issue's "if cheap" ask. It isn't cheap: the reporter's once-latch is in-process/module-scope, but `afk trace show` runs in a **separate process** after the fact — a real counter needs either a new persisted trace event (schema change in `trace/types.ts` + the zod schema in `trace/events.ts`) or a sidecar file on disk. That's real scope beyond "surface the failure," so it's left out of this PR.

## Testing

```
pnpm lint

pnpm vitest run src/agent/trace/ \
  src/agent/session/subagent-prompt-capture.test.ts \
  src/agent/session/subagent-prompt-capture-wiring.test.ts \
  src/agent/session/subagent-output-capture.test.ts \
  src/agent/session/subagent-output-capture-wiring.test.ts \
  src/utils/artifact-failure-reporter.test.ts
# Test Files  21 passed (21)
# Tests       308 passed (308)
```

`pnpm test:coverage` / the full suite were deliberately **not** run (siblings running concurrently V8-heap-OOM it per task instructions).

## Known risk for the reviewer (unverified)

`console.error` is a synchronous, unconditional stderr write. If this fires **mid-turn while `TerminalCompositor` is armed** (interactive REPL rendering the live CUP frame), the raw write could interleave with or tear the compositor's in-flight cursor-positioning output (`src/cli/cup-frame-renderer.ts`). This mirrors the exact same risk the existing `subagent-hooks.ts:178` `console.warn` precedent already carries — this PR does not introduce a new failure mode, but it does add several new call sites that can trigger the identical raw-stderr-during-TUI-render risk, so I want it flagged explicitly rather than asserted safe. I did not attempt to reproduce or verify a torn frame in this change.

Closes #850

### DIFF
--- a/src/agent/session/subagent-output-capture.ts
+++ b/src/agent/session/subagent-output-capture.ts
@@ -32,7 +32,7 @@ import { join } from 'node:path';
 import { env } from '../../config/env.js';
 import { getSubagentOutputsDir } from '../../paths.js';
 import type { OutputEvent } from '../types/session-types.js';
-import { debugLog } from '../../utils/debug.js';
+import { reportArtifactFailure } from '../../utils/artifact-failure-reporter.js';
 import { redactInlineSecrets } from './prompt-dump.js';
 import { truncateToBytes } from './subagent-prompt-capture.js';
 
@@ -186,7 +186,16 @@ export function createSubagentOutputRecorder(
         await appendFile(file, body, { encoding: 'utf8', mode: 0o600 });
       })
       .catch((err: unknown) => {
-        debugLog(`subagent-output-capture failed: ${String(err)}`);
+        // Dedup key: the witness session label — `createSubagentOutputRecorder`
+        // returns null unless `shouldCaptureSubagentOutput` confirms it's
+        // defined, so one first-failure warning covers this whole child's
+        // transcript, not one per appended record. See #850.
+        reportArtifactFailure(
+          'subagent-output-capture',
+          input.sessionId ?? 'unknown',
+          'write',
+          err,
+        );
       });
   };
 
@@ -253,7 +262,12 @@ export function createSubagentOutputRecorder(
           wroteSayThisTurn = false;
         }
       } catch (err) {
-        debugLog(`subagent-output-capture observe failed: ${String(err)}`);
+        reportArtifactFailure(
+          'subagent-output-capture',
+          input.sessionId ?? 'unknown',
+          'observe',
+          err,
+        );
       }
     },
     end(reason: string): void {
@@ -261,7 +275,12 @@ export function createSubagentOutputRecorder(
         flushSaid();
         write(buildRecord('end', { reason }));
       } catch (err) {
-        debugLog(`subagent-output-capture end failed: ${String(err)}`);
+        reportArtifactFailure(
+          'subagent-output-capture',
+          input.sessionId ?? 'unknown',
+          'end',
+          err,
+        );
       }
     },
   };
--- a/src/agent/session/subagent-prompt-capture.ts
+++ b/src/agent/session/subagent-prompt-capture.ts
@@ -28,7 +28,7 @@ import { join } from 'node:path';
 
 import { env } from '../../config/env.js';
 import { getPromptsDir } from '../../paths.js';
-import { debugLog } from '../../utils/debug.js';
+import { reportArtifactFailure } from '../../utils/artifact-failure-reporter.js';
 import { redactInlineSecrets } from './prompt-dump.js';
 
 /**
@@ -171,6 +171,14 @@ export async function captureSubagentPrompt(input: CaptureSubagentPromptInput): 
     // a pre-planted symlink at the target path.
     await writeFile(file, doc, { encoding: 'utf8', mode: 0o600, flag: 'wx' });
   } catch (err) {
-    debugLog(`subagent-prompt-capture failed: ${String(err)}`);
+    // Dedup key: the witness session label (shouldCapture guarantees it's
+    // defined by the time any of the above can throw) — one first-failure
+    // warning per session, not per captured prompt. See #850.
+    reportArtifactFailure(
+      'subagent-prompt-capture',
+      input.sessionId ?? 'unknown',
+      'captureSubagentPrompt',
+      err,
+    );
   }
 }
--- a/src/agent/trace/emit.test.ts
+++ b/src/agent/trace/emit.test.ts
@@ -17,11 +17,20 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import { resetArtifactFailureReporterForTests } from '../../utils/artifact-failure-reporter.js';
 import { emitSessionPhase, emitAbort, emitToolCall } from './emit.js';
-import { InMemoryTraceWriter, type TraceWriter } from './writer.js';
+import { InMemoryTraceWriter } from './writer.js';
 import type { TraceWriter } from './index.js';
 
-/** A writer whose `write()` always rejects, with a real (fixed) trace path
- *  so it can serve as a stable dedup key across multiple calls. */
+/**
+ * A writer whose `write()` always rejects. `tracePath` is cosmetic content
+ * for `getTracePath()` only — post-fix, `reportArtifactFailure`'s dedup key
+ * is the writer INSTANCE itself, matched by identity, so multiple distinct
+ * writers may safely share the identical `tracePath` (exercised below) and
+ * still isolate correctly. Pre-fix, the dedup key was `getTracePath()`'s
+ * return value, so every existing call to this helper used a unique
+ * `tracePath` per test — isolation between writers passed, but only
+ * because no two tests' writers ever shared a path, not because the
+ * dedup logic was actually keyed on the right thing.
+ */
 function makeFailingWriter(tracePath: string, message: string): TraceWriter {
   const inner = new InMemoryTraceWriter();
   return {
@@ -80,8 +89,9 @@ describe('emit* — #850 first-failure-visible-without-AFK_DEBUG', () => {
   });
 
   // AC3: a forced write failure emits the signal EXACTLY ONCE per dedup key
-  // (here, the writer's trace path) — not zero times, not once per call. A
-  // wedged writer (every call fails identically) must not flood stderr.
+  // (post-fix: the writer INSTANCE, matched by identity) — not zero times,
+  // not once per call. A wedged writer (every call fails identically) must
+  // not flood stderr.
   it('emits the visible signal exactly once per writer even across many failing calls', async () => {
     const errorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
     try {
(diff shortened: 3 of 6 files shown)


griffinwork40:
Shadow-verify confirms the claim. Cross-context suppression is an intentional design feature (the `write` catch comment explicitly says "one first-failure warning covers this whole child's transcript") but is not documented at the `observe` and `end` sites, and the claim stands. Waiver is valid: this is a bounded diagnostic-visibility tradeoff with no data-affecting consequence.

---

## PR #979 Review — `fix(observability): surface first artifact-write failure per subsystem without AFK_DEBUG (#850)`

**Reviewed ref:** `23361e7ded79f9482bbcb831ba94864eef097aa6`  
**Stated intent:** Surface the first artifact-write failure per subsystem to stderr unconditionally; preserve never-throw contract. Spec-compliance assessed.

---

### Findings

**F1 — `medium · blocking:false · high confidence · correctness · src/agent/session/subagent-output-capture.ts:193–283 · ref:23361e7d · citation-type:diff-context`**

The three `reportArtifactFailure` call sites (`write`, `observe`, `end`) all pass the identical `(subsystem='subagent-output-capture', dedupKey=input.sessionId)` pair. The `context` argument (`'write'`, `'observe'`, `'end'`) is message-only and is not part of the latch key (`latchKey = \`${subsystem}\u0000${String(dedupKey)}\`` at `artifact-failure-reporter.ts:105`). Once the `write` latch fires, a structurally unrelated `observe` or `end` failure for the same session is routed to `debugLog` with no visible stderr signal. The `write` catch comment ("one first-failure warning covers this whole child's transcript") establishes this as intentional, but the `observe` and `end` catch blocks carry no equivalent comment — a reader of those blocks alone will not know the latch is shared across operations.

**Evidence:**
```
// write catch (line 193):   reportArtifactFailure('subagent-output-capture', input.sessionId ?? 'unknown', 'write', err)
// observe catch (line 265): reportArtifactFailure('subagent-output-capture', input.sessionId ?? 'unknown', 'observe', err)
// end catch (line 278):     reportArtifactFailure('subagent-output-capture', input.sessionId ?? 'unknown', 'end', err)
```

**Suggestion:** Add one comment to the `observe` and `end` catch blocks: `// Same (subsystem, dedupKey) as the write path — latch already consumed after a write failure; see write catch above.`

`· waived: bounded diagnostic-visibility tradeoff in the observability layer; intentional per inline comment at line 189; consistent with AC3 as stated; no data-affecting consequence; shadow-verified CONFIRMED.`

---

**F2 — `low · blocking:false · high confidence · api-compat · src/utils/artifact-failure-reporter.ts:169–172 · ref:23361e7d · citation-type:diff-context`**

`resetArtifactFailureReporterForTests` is a named export of a production source file with no compile-time guard (no `NODE_ENV` check, no `vitest` import gate, no separate test-helpers file excluded by `tsconfig.json`). Zero production callers exist today (verified by `git grep`), but a future accidental import in a non-test module would silently drain the latch and suppress first-failure signals for the rest of the process lifetime.

**Evidence:**
```ts
export function resetArtifactFailureReporterForTests(): void {
  reportedOnce.clear();
  reportedByInstance = new WeakMap<object, Set<string>>();
}
```

**Suggestion:** Move to a sibling `artifact-failure-reporter.test-helpers.ts` excluded by `tsconfig.json`'s `exclude` array. If that's too much churn, add a `/* @internal */` JSDoc tag.

---

**F3 — `low · blocking:false · medium confidence · test-coverage · src/agent/session/subagent-output-capture.test.ts · ref:23361e7d · citation-type:diff-context`**

The three new `reportArtifactFailure` call sites in `subagent-output-capture.ts` use the string-keyed (session-ID) dedup path. No test in `subagent-output-capture.test.ts` exercises `console.error` behavior at the call-site integration level — there are no `errorSpy`, `console.error`, or `AFK_DEBUG` assertions in that file (verified by `git grep`). The reporter's string-keyed path is covered by `artifact-failure-reporter.test.ts` in isolation, but the wiring from the capture module to the reporter is untested end-to-end. `emit.test.ts` covers the object-keyed (writer-instance) path end-to-end but not this one.

**Evidence:** `git grep -n "console.error\|errorSpy\|AFK_DEBUG" subagent-output-capture.test.ts` → zero matches.

**Suggestion:** Add one test in `subagent-output-capture.test.ts` that forces a write failure (e.g. mock `appendFile` to reject) and asserts `console.error` fires once with `AFK_DEBUG` unset.

---

**F4 — `low · blocking:false · high confidence · spec-compliance · PR description · ref:23361e7d · citation-type:diff-context`**

PR description says "routes 14 existing `debugLog` swallow sites"; actual count is 16 distinct production call sites (12 `emit*` catch blocks + 3 in `subagent-output-capture.ts` + 1 in `subagent-prompt-capture.ts`). All 16 are correctly migrated; this is a description-only discrepancy with no missing sites.

**Suggestion:** Update the PR description to say "16 existing `debugLog` call sites" (or "14 catch blocks, 16 total call-site invocations" if the 14 counts block-groups).

---

**F5 — `nit · blocking:false · medium confidence · security · src/utils/artifact-failure-reporter.ts:105 · ref:23361e7d · citation-type:diff-context`**

String latch key uses `\u0000` (null byte) as delimiter: `\`${subsystem}\u0000${String(dedupKey)}\``. A `subsystem` string containing `\u0000` produces the same key as a different `(subsystem, dedupKey)` split at that byte. All current call sites use clean alphanumeric/punctuation labels (`'trace.emit'`, `'subagent-output-capture'`, etc.) — no collision is reachable today.

**Evidence:** `` const latchKey = `${subsystem}\u0000${String(dedupKey)}` `` at `artifact-failure-reporter.ts:105`

**Suggestion:** Use `\u0001` or a two-element composite that doesn't appear in any internal label, or a two-key `Map<string, Map<string, true>>` structure.

---

### Dropped findings

| Finding | Reason |
|---|---|
| nit: `emit.ts:19-20` malformed JSDoc | FABRICATED — Wave 1.5 verified lines 19-20 carry correct ` * ` prefixes at the reviewed ref |

---

### What was not checked

- Citations verified inline against branch HEAD `23361e7ded79f9482bbcb831ba94864eef097aa6`.
- Did not run the test suite or linter (read-only review).
- Did not review the Telegram surface or the daemon/cron scheduler paths for any interaction with `reportArtifactFailure`.
- Did not verify the PR author's claim about `TerminalCompositor` TUI-render interleave risk (identified as "known risk" in the PR body — accepted on the author's representation that it matches an existing `subagent-hooks.ts:178` precedent).
- Stated intent: PR #979 title+body — spec-compliance assessed.

---

**Decision: MERGE — 0 blocking (1 medium waived [correctness · cross-context latch suppression, shadow-verified intentional], 3 low [api-compat, test-coverage ×2, spec-compliance], 1 nit [security]).**

The implementation is clean. The new `artifact-failure-reporter` module is well-structured, well-tested, and handles the WeakMap/string dual-key design correctly. The identity-based dedup key for `TraceWriter` objects is the right call — it sidesteps both the partial-writer TypeError and the `InMemoryTraceWriter` shared-sentinel collision. The `observe`/`end` comment gap (F1) and the missing integration test (F3) are worth following up, but neither is a merge blocker.

---

Issue #866, opened 2026-08-02
cost accounting: '?? 0' coalescing conflates unknown cost with zero cost


griffinwork40:
## Problem

Cost accounting coalesces "unknown cost" to `0` with `?? 0`, making unpriced usage indistinguishable from genuinely free usage. Totals therefore under-report spend with no signal that anything is missing.

## Cause

`src/agent/session-stats.ts:72` (and the mirrored `src/cli/slash/session-stats.ts:72`) sum with `?? 0`. A turn whose cost could not be computed — an openai-compatible model with no pricing table, or a turn interrupted before usage was finalized (`src/agent/providers/anthropic-direct/query.ts:648`) — contributes `0` to the total and is then indistinguishable from a turn that really cost nothing.

Downstream, `afk insights` presents the resulting sum as an authoritative "Total Cost" with no caveat.

## Suggested fix

1. Keep cost nullable through aggregation rather than coalescing at the summation site.
2. Track an `unpricedSessions` / `unpricedTurns` counter alongside the total.
3. Surface it in `afk insights` and in `/cost`, e.g. `$1,616.85 (2,431 sessions unpriced)`.

This is a correctness-of-reporting fix rather than a behavior change: the total stays a lower bound, but it stops *presenting* as a complete figure.

## Related

Companion issues cover the openai-compatible pricing table (the largest source of unpriced turns) and the `afk insights` headline reading the narrower sidecar aggregate.

griffinwork40:
## Same conflation, one layer up: `traces.totalCostUsd` cannot distinguish "traced but unclosed" from "traced and free"

Surfaced while reviewing #933 (the #864 fix). Filing here rather than as a new issue because it is this issue's failure mode at a different layer, and the fix proposed here subsumes it.

### Mechanism

`insights`' trace aggregate populates its numerator and its denominator at different points in the scan:

- `src/insights/aggregators/traces.ts:153` — `agg.totalTracedSessions += 1` fires as soon as a `trace.jsonl` passes the window filter, **before any event is parsed**.
- `src/insights/aggregators/traces.ts:210-221` — `agg.totalCostUsd += payload.finalCostUsd` accrues **only** inside `case 'closure':`, the sole write site.

[...]

PR #981, opened 2026-08-10
feat(openai-compatible): add a pricing table so token usage records real cost (#865)

### LINKED ISSUE #865: openai-compatible: no pricing table — 44% of sessions record $0 despite real token usage
## Problem

The openai-compatible provider has no pricing table, so every session run against a non-Anthropic model records `finalCostUsd: 0` despite consuming real tokens.

Measured on a local install: **3,605 of 8,280 closure events (44%) carry non-zero tokens with `finalCostUsd: 0`.** Every one of the ten heaviest sessions on the machine — up to 11.9M tokens in a single session — reports $0.00.

## Cause

`src/agent/providers/anthropic-direct/types.ts:481-541` defines a static `MODEL_PRICING` table and `deriveCallCostUsd()`. There is no equivalent under `src/agent/providers/openai-compatible/` — `totalCostUsd` is simply never populated on that path.

This affects any `gpt-*`, `o1/o3/o4-*`, `codex-*` model, and every HuggingFace-style `org/model` id served by a local OpenAI-shim runner (MLX, llama.cpp, vLLM, ollama-openai).

## Suggested fix

Add a pricing table for the openai-compatible provider, mirroring the shape of the anthropic-direct one. Two considerations:

1. **Locally-served models genuinely cost $0.** A local MLX/llama.cpp endpoint has no per-token price, so it should resolve to a real zero, not an unknown. Distinguishing "priced at zero" from "price unknown" matters — see the companion issue on `?? 0` coalescing.
2. Prices should be overridable via config, since operators run these behind arbitrary base URLs where the model id does not imply a rate.

## Impact

Until this lands, any cost figure produced by AFK is a lower bound that silently excludes ~44% of recorded sessions.

griffinwork40:
# feat(openai-compatible): add a pricing table so token usage records real cost

Closes #865

## Problem

The anthropic-direct provider prices its token usage; the openai-compatible provider had no pricing table at all. Real token usage therefore recorded as **$0**, which is why ~44% of sessions show zero cost despite genuine spend.

## The load-bearing distinction: unknown is not zero

The fix is not just "add numbers" — it is making *unknown* cost representable.

`deriveCallCostUsd()` returns `undefined` for any model not in the table, and `usageFromState()` only sets `totalCostUsd` when a cost is actually derived; otherwise the field is **omitted entirely**, never coerced to `0`. Conflating "we don't know what this cost" with "this was free" is the original defect, and is also the subject of sibling issue #866.

## Shape

New file `src/agent/providers/openai-compatible/pricing.ts` holds a `MODEL_PRICING` Map plus a pure `deriveCallCostUsd()`. It deliberately mirrors `anthropic-direct/pricing.ts` — exact-match lookup, then a retry with the dated-snapshot suffix stripped — so a maintainer who knows one provider's pricing code already knows this one.

Prices are **list prices as of the date noted in the module comment, and are maintainer-maintained**. They are not fetched at runtime and will drift; extending or correcting the table is a single-file edit by design.

## Design deviation worth reviewing

The issue suggested treating local OpenAI-shim runners (MLX, llama.cpp, vLLM, ollama — HuggingFace-style `org/model` ids) as a deliberate **zero**, since they are genuinely free.

This PR treats them as **unknown** instead. Reason: this provider also serves *paid* OpenRouter models over the identical `org/model` slash shape, and no heuristic separates the two reliably. Hard-coding zero for that shape would silently under-report real spend — the same class of bug this PR exists to fix. "Unknown" is the uniformly safe default. Flagging this explicitly because it differs from the issue's framing; happy to switch if you'd rather have the local-runner zero.

## Files

- `pricing.ts` (new, 135 LOC) — table + `deriveCallCostUsd`
- `pricing.test.ts` (new, 96 LOC)
- `translate.ts` — `usageFromState` gains an optional `model` param and derives cost
- `translate.test.ts` (+28)
- `query.ts` (+1) — passes `this.currentModel`

All files under the 350-LOC cap. Scope is confined to pricing/usage accounting; nothing here touches the system-prompt assembly files being changed by #876.

## Verification

- `tsc --noEmit` — exit 0
- `vitest run pricing.test.ts translate.test.ts` — **31 passed, 0 failed**
- Asserted directly: a known model yields a real dollar number; an unknown model and a missing model both leave `totalCostUsd` `undefined`, explicitly **not** `0`.

## Not checked

- Full `query.test.ts` was not run (grep found no `usageFromState` / `totalCostUsd` references there, so risk is low but unconfirmed).
- No `responses-translate.test.ts` case exercises the cost field directly; the Responses path reuses `usageFromState` via shared `state.usage`, so it is likely fine but unverified.
- Price accuracy is not live-verified — see the staleness note above.

### DIFF
--- a/src/agent/providers/openai-compatible/pricing.test.ts
+++ b/src/agent/providers/openai-compatible/pricing.test.ts
@@ -51,6 +51,12 @@ describe('deriveCallCostUsd — known models price correctly', () => {
     const cost = deriveCallCostUsd('o3', M, M, 0);
     expect(cost).toBeCloseTo(10.0, 8);
   });
+
+  it('uses the distinct o1-mini rates rather than o3-mini rates', () => {
+    const cost = deriveCallCostUsd('o1-mini', M, M, M);
+    // $1.50 cached input + $12.00 output per MTok.
+    expect(cost).toBeCloseTo(13.5, 8);
+  });
 });
 
 describe('deriveCallCostUsd — unknown model yields unknown, never zero', () => {
--- a/src/agent/providers/openai-compatible/pricing.test.ts
+++ b/src/agent/providers/openai-compatible/pricing.test.ts
@@ -1,0 +1,96 @@
+/**
+ * Unit tests for the openai-compatible pricing table.
+ *
+ * The load-bearing assertion in this file is the unknown-vs-zero
+ * distinction (issue #865, sibling issue #866): a model absent from
+ * {@link MODEL_PRICING} must derive `undefined` cost, never `0`, so a
+ * downstream `?? 0` coalescing bug cannot silently misreport "we don't know"
+ * as "this was free."
+ *
+ * @module agent/providers/openai-compatible/pricing.test
+ */
+
+import { describe, it, expect } from 'vitest';
+import { deriveCallCostUsd, MODEL_PRICING } from './pricing.js';
+
+const M = 1_000_000;
+
+describe('deriveCallCostUsd — known models price correctly', () => {
+  it('prices a plain gpt-4o call from base rates', () => {
+    // $2.50 input + $10.00 output per MTok.
+    const cost = deriveCallCostUsd('gpt-4o', M, M, 0);
+    expect(cost).toBeCloseTo(12.5, 8);
+  });
+
+  it('applies the discounted cached-input rate for the cached portion only', () => {
+    // gpt-4o: $2.50 plain input, $1.25 cached input per MTok. Half the input
+    // is cached: 0.5 * 1.25 + 0.5 * 2.50 = 1.875 for input, plus 0 output.
+    const cost = deriveCallCostUsd('gpt-4o', M, 0, M / 2);
+    expect(cost).toBeCloseTo(1.875, 8);
+  });
+
+  it('prices every table row with all-cached input at its cachedInputPerMTok rate', () => {
+    for (const [model, pricing] of MODEL_PRICING.entries()) {
+      const cost = deriveCallCostUsd(model, M, 0, M);
+      expect(cost).toBeCloseTo(pricing.cachedInputPerMTok ?? pricing.inputPerMTok, 8);
+    }
+  });
+
+  it('resolves a dated snapshot id against its dateless table row', () => {
+    const dated = deriveCallCostUsd('gpt-4o-2024-08-06', M, M, 0);
+    const dateless = deriveCallCostUsd('gpt-4o', M, M, 0);
+    expect(dated).toBeCloseTo(dateless!, 8);
+  });
+
+  it('is case-insensitive on the model id', () => {
+    expect(deriveCallCostUsd('GPT-4O', M, 0, 0)).toBeCloseTo(2.5, 8);
+  });
+
+  it('prices an o-series reasoning model', () => {
+    // o3: $2.00 input + $8.00 output per MTok.
+    const cost = deriveCallCostUsd('o3', M, M, 0);
+    expect(cost).toBeCloseTo(10.0, 8);
+  });
+});
+
+describe('deriveCallCostUsd — unknown model yields unknown, never zero', () => {
+  it('returns undefined for a model absent from the table', () => {
+    const cost = deriveCallCostUsd('some-model-nobody-has-heard-of', 1000, 1000, 0);
+    expect(cost).toBeUndefined();
+    // Explicit anti-regression: undefined must never get coerced to 0 by a
+    // careless `?? 0` at a call site (issue #866).
+    expect(cost).not.toBe(0);
+  });
+
+  it('returns undefined for a HuggingFace-style local-shim id (e.g. mlx-community/*)', () => {
+    // Deliberate design choice (see pricing.ts module docstring): this
+    // provider also serves paid OpenRouter-style `org/model` ids over the
+    // identical slash shape, so "unknown" is the safe default for every
+    // unlisted id rather than special-casing local runners to a hardcoded 0.
+    expect(deriveCallCostUsd('mlx-community/qwen3-30b-a3b-4bit', 1000, 1000, 0)).toBeUndefined();
+  });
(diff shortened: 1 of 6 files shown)


griffinwork40:
@codex fix comments

griffinwork40:
Synthesizing the four dimension reports.

---

## Review: PR #981 — `feat(openai-compatible): add pricing table so token usage records real cost`

**+309 / -5 across 6 files** · Branch `afk/issue-865-openai-pricing → main`

---

### Findings by severity

#### 🔴 Major

**1. `o1-preview` and `o1-pro` absent from `MODEL_PRICING`** *(security_perf)*

`o1-preview` was the initial GA name before the `o1` alias stabilised; `o1-pro` is a current live model. Neither is caught by the `DATE_SUFFIX` regex (`/-\d{4}-\d{2}-\d{2}$/`) — `-preview` and `-pro` are not date suffixes, so `lookupPricing` returns `undefined` for both. Users on those models get silent unknown-cost every turn, which is the safe fallback (not `$0`), but it's an omission that reduces the PR's utility immediately after ship.

Suggested fix: add two rows:
```ts
['o1-preview', { inputPerMTok: 15.0, outputPerMTok: 60.0, cachedInputPerMTok: 7.5 }],
['o1-pro',     { inputPerMTok: 150.0, outputPerMTok: 600.0 }],  // verify current rate
```

---

**2. Env-level base URL override not checked by `useOpenAIPricing` guard** *(correctness — 75% confidence)*

`query.ts:333`:
```ts
this.useOpenAIPricing = wire.baseURL === undefined && opts.baseURL === undefined;
```
The guard correctly catches `opts.baseURL` (propagated from `buildQueryFromConfig`) and `wire.baseURL` (ChatGPT-subscription backend). However, if any env var (e.g. `AFK_OPENAI_BASE_URL`, `OPENAI_BASE_URL`, or an OpenAI SDK default from `OPENAI_API_BASE`) can redirect the effective endpoint *without flowing through these two fields*, the guard would apply OpenAI list prices to a custom endpoint. The reviewer flagged this at 75% confidence — if no such env var exists in `src/config/env.ts`, the guard is sound and this drops to a nitpick. Worth a quick search before merge.

---

#### 🟡 Minor

**3. No runtime signal on unknown model** *(security_perf)*

When `lookupPricing` returns `undefined`, cost silently becomes unavailable. In production, stale table entries (price changes, new model families) would be invisible without a structured log event or trace annotation. Not a correctness bug — `undefined` is the safe contract — but it means no telemetry for table drift.

**4. Responses API path remains unpriced** *(api_compat)*

The Responses-API translator (`responses-translate.ts`) never gets a model passed in — the PR's comment documents this gap explicitly. Cost stays `undefined` for all Responses-API calls. Known limitation, not a regression, but worth a follow-up issue to complete parity.

**5. No explicit `Infinity` token test** *(test_coverage)*

`clamp()` uses `Number.isFinite()`, so `Infinity → 0` via the same branch as `NaN`. The existing NaN test exercises the path, but `Infinity` is a distinct JS value worth a one-liner for self-documentation. Not a real risk given IEEE 754 determinism.

**6. No named test distinguishing o3-mini vs o4-mini cached rate** *(test_coverage)*

These two models have identical base rates but differ only in `cachedInputPerMTok` (0.55 vs 0.275). The loop test covers this, but a future maintainer who "deduplicates" them could silently break the distinction — a named spot-check would armor against that.

---

#### ⬜ Nitpick

- `translate.test.ts`: `expect(u).not.toHaveProperty('totalCostUsd', 0)` passes even when the property is absent, making it weaker than it reads. The `query.test.ts` assertion (`toBeUndefined()`) closes the gap, but the translate test assertion is misleading in isolation.
- `query.test.ts`: The companion-test relationship (same `baseConfig()` model, opposite `baseURL` condition) proving priced vs. unpriced is implicit. A one-line comment linking the two tests would make the pairing intentional.

---

### What's correct and well-done

- **Cached-token math is right**: `plainInput = prompt_tokens - cached_tokens`, billing the non-cached slice at full rate and the cached slice at the discounted rate. No double-billing.
- **`undefined`-not-zero contract is correctly implemented and tested** at two layers (`pricing.ts` and `translate.ts`), closing the core defect from issues #865/#866.
- **`DATE_SUFFIX` regex is sound**: anchored at `$`, so intermediate date-like strings in model names don't get stripped.
- **`ProviderUsage.totalCostUsd` was already declared** in `provider.ts:84` — no interface change.
- **Anthropic-direct parity is real**: the conditional-assignment pattern mirrors `anthropic-direct/usage.ts` exactly.
- **Only caller updated**: `query.ts:571–574` is the sole non-test call site.
- **Pure function**: `deriveCallCostUsd` has zero imports — no env reads, no clock.
- **SDK lock unaffected**: no `@anthropic-ai/sdk` symbols added.
- **Test coverage is strong** for the load-bearing cases (unknown model → `undefined`, not `0`).

---

### Merge recommendation

**APPROVE with changes.** The architecture is sound and the core fix is correct — the undefined-not-zero contract is the right design and is properly enforced at both layers. The two majors are safe degradations (unknown models fall back to `undefined`, not wrong numbers), but the `o1-preview`/`o1-pro` omission is immediately actionable before ship. Recommend:

1. Add `o1-preview` and `o1-pro` rows to `MODEL_PRICING` (verify `o1-pro` rate against current OpenAI pricing page first).
2. Do a quick `grep` for env vars that set the OpenAI base URL to confirm the `useOpenAIPricing` guard is complete; if sound, close item 2.
3. Items 3–6 can be deferred to follow-up issues.

**Done**

- Read diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/fc319951-1638-42bd-aa1e-61ba7262eb1b/pr-981.diff`
- Dispatched four parallel dimension agents: correctness, API compatibility, test coverage, security/perf
- Synthesized findings above; no files were written or commits made (read-only review)
- Working tree is unchanged; the operator's dirty state was not touched

---

PR #984, opened 2026-08-10
feat(trace): record bootstrap warnings as a durable boot_warning phase (#754)

### LINKED ISSUE #754: observability: bootstrap warnings reach no durable sink — `afk trace show` can't tell you a session warned about a shadowed builtin
Surfaced by the review of #751 (Finding 2, `low`). Not a regression — this gap predates both #745 and #751; those fixed *terminal* visibility, and this is about *durable* recording.

## The gap

Bootstrap warnings are terminal-only. Nothing writes them to a durable sink, so `afk trace show` cannot answer the one question that matters after the fact:

> did this session warn that a user agent file shadows a read-only builtin?

Once scrollback rolls, the signal is gone. Verified at `c107e80`:

- `git grep -n bootWarnings -- src/agent` → **zero matches**. The trace and transcript writers both live under `src/agent/`, so neither sees these strings.
- Both drain sites print and discard: `src/cli/commands/interactive.ts:793` (success, post-clear) and `:427` (bootstrap abort). The shared printer at `src/cli/commands/interactive/boot-warnings.ts:26` only calls `console.warn`.
- The neighbouring MCP phase events (`bootstrap.ts:533,547`) carry `{ serverCount }` and nothing about the warnings collected two lines earlier at `:522`.

This matters most for the builtin-shadow warning from #739, which is a *safety* signal: a `~/.afk/agents/research-agent.md` declaring broader `tools:` silently converts the read-only verifier that nine bundled skills dispatch by bare name into a write-capable agent, for every project on the machine. Today there is no way to establish after a session whether that condition was present, and `/agents` offers no recovery path (both providers return `[]` from `supportedAgents()`).

## Two traps — please read before implementing

The review's literal suggestion was "emit one trace event alongside the drain." **That does not work as stated**, and the naive version of it fails a second time in a way that looks green.

**Trap 1 — the writer is not reachable at either drain site.**
`InteractiveCtx` exposes no trace writer (`grep -n trace src/cli/commands/interactive/shared.ts` → zero hits). The writer is a local `const trace` inside `bootstrapSession` (`bootstrap.ts:223`). So:
- at the success drain (`interactive.ts:793`) `ctx` carries no writer;
- at the abort drain (`interactive.ts:427`) bootstrap has *thrown*, so nothing it created is reachable at all.

That second one is structurally the same bug #751's review just fixed for the warnings themselves. Threading the writer out to the caller would mean repeating the hoist for a second value.

**Emit at PUSH time instead, inside bootstrap**, where the writer is already in scope. The writer is created at `:223`, before the earliest producer (`:347`), so every push is covered — including pushes that precede a throw, which get recorded *for free* with no caller plumbing. It also decouples the sink from the drain: the trace records the warning even if the drain never runs.

**Trap 2 — `session_phase` is low-signal by default, so the obvious implementation is invisible.**
`afk trace show` renders `session_phase` events only under `--all` (`trace.ts:489`). Only `rate_limit` and the usage-limit pair get explicit carve-outs *before* that gate (`:447`, `:462`, `:478`) precisely because they explain otherwise-invisible stalls.

So a bare `emitSessionPhase({ phase: 'boot_warning' })` satisfies "it's in the trace" while remaining hidden in the default view — recreating the #745 bug class one layer down. This is the same trap #751 called out and dodged: reusing `getPluginShadowingNoticeLines()` was rejected because its drain site sits behind `isDebugEnabled()`, which would have made safety warnings debug-mode-only — "strictly worse than the stderr status quo, while satisfying every acceptance criterion in the issue."

A new phase therefore needs an explicit default-view carve-out, placed before the `showAll` gate, alongside `rate_limit`.

## Proposed shape

1. Add `'boot_warning'` to `SessionPhaseName` (`src/agent/trace/types.ts:667` — closed union).
2. Emit one event **per warning at push time** in `bootstrap.ts`, rather than one aggregate event at drain time. Note `metadata` is `Record<string, string | number | boolean>` (`types.ts:736`) — **no arrays**, so an aggregate would have to join strings into one unbounded field. Per-warning also gives the renderer something clean to print.
   Suggested metadata: `{ producer: 'agent-registry' | 'mcp', message }`.
   The two producers are `:347` (registry `warn` callback) and `:522` (MCP config warnings); routing both through one small recorder that pushes *and* emits keeps them from diverging — the divergence #751 already had to repair once.
3. Render it in the DEFAULT view of `afk trace show`, before the `showAll` gate (`trace.ts:441`), following the `rate_limit` precedent at `:447`.
4. Consider counting boot warnings into the run summary (`trace.ts:540-566`), next to `throttles`.

## Acceptance criteria

- [ ] A session whose bootstrap produced a builtin-shadow warning shows it in plain `afk trace show`, with **no `--all`**.
- [ ] A session whose bootstrap **threw** after collecting a warning still records it (the abort path is the whole reason for push-time emission).
- [ ] `afk trace show --json` carries the warning text so it is greppable across the trace corpus.
- [ ] A test asserts the event lands **in the default render path**, not merely in the NDJSON. Absent that assertion this ships green and invisible.

## Out of scope

- The transcript writer as a second sink — the trace is the natural home, and `afk trace show` is the existing "what happened in that run" verb.
- Widening the producer set. The erased class is larger than these two (`bootstrap.ts:959` path-approval, `default-hook-registry.ts:207,299`, `config-bridge.ts:81`, `permissions-store.ts:152`, `mcp/manager.ts:136,235`, `plugins/tool-injector.ts:327-336`); `ctx.bootWarnings` is the seam for those, tracked separately.
- #752 (`parseAgentMarkdown` warn callback never passed) is a different bug — never written, versus written-but-not-durably-recorded.

griffinwork40:
## Problem

Bootstrap warnings (agent-registry builtin-shadow, MCP config) were terminal-only — printed once, then gone. `afk trace show` had no way to answer "did this session warn that a user agent file shadows a read-only builtin?" That's a SAFETY signal: a `~/.afk/agents/research-agent.md` declaring broader `tools:` silently converts the read-only verifier that nine bundled skills dispatch by bare name into a write-capable agent, machine-wide (#739). With no durable record, that conversion was invisible to any post-hoc audit.

### The two traps, and how each is dodged

1. **Writer unreachable at drain sites.** Bootstrap warnings are collected into a `bootWarnings` array and drained in two places in `interactive.ts` — after the startup-screen clear (success path) and in the abort catch (failure path). Neither drain site has a trace writer in scope: `InteractiveCtx` exposes none, and on the abort path `bootstrapSession` has already thrown by the time `interactive.ts` would try to read one. **Dodge:** `recordBootWarning()` (new file `boot-warning-recorder.ts`) emits the `boot_warning` trace event at **push time** — inside `bootstrapSession`, the instant a producer (`agentRegistryWarn` callback in `bootstrap-infra.ts`, or the MCP config-warnings loop in `bootstrap-mcp.ts`) collects the warning — not at either drain site. The trace writer is constructed before the earliest producer can fire, so this covers every warning, including ones collected just before a later bootstrap phase throws.
2. **`session_phase` hidden by default.** `afk trace show` renders `session_phase` events only under `--all` — they're normally a low-signal latency waterfall. Left alone, `boot_warning` would inherit that gate and ship green but invisible — exactly the failure class the issue describes. **Dodge:** added an explicit `boot_warning` carve-out in `trace.ts`'s `renderEvent`, before the `showAll` gate, following the existing `rate_limit` / `usage_limit_*` / `compaction_disabled` precedent — all "too important to hide" phases already special-cased the same way.

## Fix

- `src/agent/trace/types.ts` / `events.ts`: added `boot_warning` to the closed `SessionPhaseName` union and the mirrored Zod enum (parity-tested).
- `src/cli/commands/interactive/boot-warning-recorder.ts` (new): `recordBootWarning()` — one function both producers call, so pushing to the terminal-facing bucket and emitting the trace event can never diverge.
- `src/cli/commands/interactive/bootstrap-infra.ts`, `bootstrap-mcp.ts`: both producers now route through `recordBootWarning` instead of a raw `.push()`.
- `src/cli/commands/trace.ts`: `boot_warning` branch in the DEFAULT render path (before `showAll`), rendered as a `boot-warn` line with `[producer] message`; also counted into the summary header next to `throttled`/`ttfb-stall` (`N boot-warn`).

## Acceptance criteria → tests

| # | Criterion | Proof |
|---|---|---|
| 1 | Warning shows in plain `afk trace show`, no `--all` | `trace.test.ts` → `formatTrace — boot_warning is high-signal (shown by default) (#754)` → `renders the boot_warning event in the DEFAULT view (no --all)` — asserts on `formatTrace(...)` called with **no** `showAll` option |
| 2 | A bootstrap that threw after collecting a warning still records it | `boot-warning-recorder.test.ts` → `records the event even when the caller throws immediately afterward (abort path, #754)` — calls `recordBootWarning` then throws synchronously, asserts the `InMemoryTraceWriter` already has the event |
| 3 | `afk trace show --json` carries the warning text so it's greppable | `trace.test.ts` → `carries the warning text in --json (NDJSON passthrough) so it is greppable` — `--json` is a raw file passthrough (`trace.ts` `options.json` branch), so the source JSONL is the contract; asserts the JSON-escaped message and `"phase":"boot_warning"` appear in the raw line |
| 4 | A test asserts the event lands in the DEFAULT render path | Same test as #1 — it is the single most important test in this PR; without it the feature ships green and invisible per the issue |

Also covered: `session-phase.test.ts` schema-acceptance + union↔schema parity (pre-existing from prior WIP, verified green), `boot-warning-recorder.test.ts` push-into-array + one-event-per-warning + no-writer no-op, `trace.test.ts` low-signal isolation (`model_ttfb` still hidden) and header-count-omitted-when-zero.

## Testing

```
pnpm lint

pnpm vitest run src/cli/commands/trace.test.ts src/agent/trace/session-phase.test.ts \
  src/cli/commands/interactive.boot-warnings.test.ts \
  src/cli/commands/interactive/boot-warning-recorder.test.ts
# 4 test files, 112 tests, all passed
```

Did not run `pnpm test:coverage` or the full suite per instructions (concurrent siblings V8-heap-OOM it).

## Notes for reviewers

- `src/cli/commands/trace.ts` was already 815 LOC before this change (pre-existing; every other high-signal `session_phase` carve-out lives inline in the same `renderEvent` function). This PR adds ~25 lines to it, consistent with existing precedent rather than new debt — flagging in case the 350-LOC guidance is meant per-file rather than per-newly-authored-file.
- No `process.env`, no `chalk.<color>()`, no hand-joined `~/.afk/` paths introduced.

Closes #754

### DIFF
--- a/src/agent/trace/events.ts
+++ b/src/agent/trace/events.ts
@@ -433,6 +433,9 @@ export const SessionPhaseNameSchema = z.enum([
   // Per-session compaction disable (single event, at most once). See
   // SessionPhaseName JSDoc in types.ts — metadata names the wire and cause.
   'compaction_disabled',
+  // One bootstrap warning, emitted per-warning at push time (#754). See
+  // SessionPhaseName JSDoc in types.ts — metadata carries producer + message.
+  'boot_warning',
 ]);
 
 export const SessionPhasePayloadSchema = z.object({
--- a/src/agent/trace/session-phase.test.ts
+++ b/src/agent/trace/session-phase.test.ts
@@ -267,6 +267,31 @@ describe('session_phase payload schema — acceptance', () => {
     // Bare name-schema acceptance.
     expect(() => SessionPhaseNameSchema.parse('suspected_loop')).not.toThrow();
   });
+
+  it('accepts the boot_warning phase with producer + message metadata (#754)', () => {
+    // Emitted once PER WARNING at push time inside bootstrapSession — never
+    // aggregated — so `metadata` carries exactly one producer tag and one
+    // message string (no arrays; SessionPhasePayload.metadata is a flat
+    // Record<string, string|number|boolean>).
+    expect(() =>
+      SessionPhasePayloadSchema.parse({
+        phase: 'boot_warning',
+        metadata: {
+          producer: 'agent-registry',
+          message:
+            '[afk] agents: ~/.afk/agents/research-agent.md overrides built-in agent "research-agent"',
+        },
+      }),
+    ).not.toThrow();
+    expect(() =>
+      SessionPhasePayloadSchema.parse({
+        phase: 'boot_warning',
+        metadata: { producer: 'mcp', message: '[mcp] server "foo": unknown key "cmd"' },
+      }),
+    ).not.toThrow();
+    // Bare name-schema acceptance.
+    expect(() => SessionPhaseNameSchema.parse('boot_warning')).not.toThrow();
+  });
 });
 
 // ---------------------------------------------------------------------------
@@ -307,6 +332,7 @@ describe('SessionPhaseName union ↔ SessionPhaseNameSchema parity', () => {
     pause_extension_granted: true,
     suspected_loop: true,
     compaction_disabled: true,
+    boot_warning: true,
   };
 
   it('the Zod enum contains exactly the union members (no drift either way)', () => {
--- a/src/agent/trace/types.ts
+++ b/src/agent/trace/types.ts
@@ -805,7 +805,20 @@ export type SessionPhaseName =
   // signal: the auto-compaction caller discards its result, so without this the
   // disable is invisible until a human runs /compact — and an undisclosed
   // disable ends in a context-window overflow the operator cannot explain.
-  | 'compaction_disabled';
+  | 'compaction_disabled'
+  // A bootstrap-time warning (agent-registry builtin-shadow, MCP config) was
+  // collected before `interactive.ts`'s startup-screen clear could destroy it
+  // (issue #745). Emitted ONE EVENT PER WARNING at PUSH time inside
+  // `bootstrapSession` (see `boot-warning-recorder.ts`) — never aggregated at
+  // drain time — so a warning collected just before bootstrap THROWS is still
+  // recorded even though the drain never runs on that path (issue #754).
+  // `metadata.producer` names which producer pushed it (`agent-registry` |
+  // `mcp`); `metadata.message` carries the warning text verbatim. HIGH
+  // SIGNAL: the builtin-shadow warning from #739 is a safety signal (a user
+  // agent file can silently convert a read-only verifier into a
+  // write-capable agent machine-wide), so — like `rate_limit` — this renders
+  // in the DEFAULT `afk trace show` view instead of behind `--all`.
+  | 'boot_warning';
 
 export interface SessionPhasePayload {
   /** Which lifecycle milestone this record marks. */
--- a/src/cli/commands/interactive/boot-warning-recorder.test.ts
+++ b/src/cli/commands/interactive/boot-warning-recorder.test.ts
@@ -52,6 +52,11 @@ describe('recordBootWarning', () => {
   });
 
   it('records the event even when the caller throws immediately afterward (abort path, #754)', () => {
+    // Synchrony guarantee is specific to InMemoryTraceWriter: writes complete
+    // inline. NdjsonTraceWriter is async (queued NDJSON appends) — on that
+    // writer the returned Promise must be awaited before a process.exit to
+    // guarantee the event reaches disk (see bootstrap-mcp.ts, which awaits).
+    //
     // Simulates the real shape: `agentRegistryWarn`/the MCP warnings loop call
     // `recordBootWarning`, and a LATER bootstrap phase throws before either
     // terminal-facing drain site (`interactive.ts`) can run. Because emission
--- a/src/cli/commands/interactive/boot-warning-recorder.test.ts
+++ b/src/cli/commands/interactive/boot-warning-recorder.test.ts
@@ -1,0 +1,102 @@
+/**
+ * Tests for `recordBootWarning` (#754).
+ *
+ * The critical property under test is PUSH-TIME emission: the durable
+ * `boot_warning` trace event must land in the writer synchronously as part
+ * of calling `recordBootWarning`, not at some later drain point — because on
+ * the abort path (bootstrap throws after a producer already warned) no drain
+ * site is ever reached. If emission were deferred past the call, a
+ * synchronous throw right after `recordBootWarning` would race it.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { InMemoryTraceWriter } from '../../../agent/trace/writer.js';
+import { recordBootWarning } from './boot-warning-recorder.js';
+
+function bootWarningEvents(writer: InMemoryTraceWriter) {
+  return writer.events.filter(
+    (e) => e.kind === 'session_phase' && e.payload.phase === 'boot_warning',
+  );
+}
+
+describe('recordBootWarning', () => {
+  it('pushes onto the caller-owned bootWarnings array (unchanged terminal behavior)', () => {
+    const bootWarnings: string[] = [];
+    const writer = new InMemoryTraceWriter();
+    recordBootWarning({
+      bootWarnings,
+      traceWriter: writer,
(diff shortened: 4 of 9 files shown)


griffinwork40:
The blocking finding (doubled `[mcp]` prefix) is medium-correctness. Per the skill protocol, findings whose `blocking` value matches the default table don't need shadow-verification — only critical/high or overridden findings trigger that. This medium is at default disposition, so we proceed directly to the verdict.

---

## PR #984 — `feat(trace): record bootstrap warnings as a durable boot_warning phase (#754)`

**Target:** `afk/issue-754-boot-warning-sink` → `main`  
**Ref:** `ea75ce0629638da4ab29dd082be209cc93e5674b`  
**Triage:** feature · 9 files (+317 / -7) · full regime

---

### Findings

#### ❌ MEDIUM · blocking · correctness — doubled `[mcp]` prefix in trace render

`src/cli/commands/interactive/bootstrap-mcp.ts:64` · `src/cli/commands/trace.ts:543`

`bootstrap-mcp.ts` bakes `[mcp]` into the message string at the call site:
```ts
message: `[mcp] ${w}`,
```
Then `trace.ts` `renderEvent` independently prepends the producer tag:
```ts
const producerBit = producer !== undefined ? `[${String(producer)}] ` : '';
return line('boot-warn', `${producerBit}${message …}`);
```
Every MCP-sourced boot warning renders as `[mcp] [mcp] server "foo": unknown key "cmd"` in `afk trace show`. The terminal drain path (`drainBootWarnings`) is unaffected — it reads the already-prefixed message directly. No render test covers `producer: 'mcp'`, so the doubling is undetected by the test suite.

**Fix:** Strip the `[mcp]` literal from the message in `bootstrap-mcp.ts` (pass raw `w` instead of `` `[mcp] ${w}` ``), or remove `producerBit` from `renderEvent` and rely on message strings. Add a render test with `producer: 'mcp'` asserting the output does **not** contain `[mcp] [mcp]`.

---

#### LOW · non-blocking · test-coverage — synchrony assumption unscoped

`src/cli/commands/interactive/boot-warning-recorder.test.ts:54-59`

The abort-path test implicitly relies on `InMemoryTraceWriter._events.push` being synchronous (no `await` in the non-compaction path), but the comment says "emission happens synchronously at push time" without scoping that to the in-memory writer. If the test were ever adapted to use `NdjsonTraceWriter`, the assertion would be racy.

**Fix:** Add a one-line comment scoping the guarantee to `InMemoryTraceWriter`.

---

### Clean dimensions

- **Security** — no issues found; `metadata` is schema-validated, rendered with `String()` coercion, no injection surface
- **API-compat** — all type additions additive; `SessionPhaseName` union and `SessionPhaseNameSchema` extended in parity; no external consumers
- **Spec-compliance** — all 4 acceptance criteria met; no unmet intent; no scope creep
- **Perf-observability** — `summarize()` counts boot warnings in the existing single-pass loop; no regression

---

### Decision: ❌ DO NOT MERGE — 1 medium blocking (correctness)

1 medium (blocking, correctness) · 1 low (non-blocking, test-coverage)

---

### What was not checked

- Citations verified inline against branch HEAD `ea75ce0629638da4ab29dd082be209cc93e5674b`
- Stated intent: PR #984 title+body — spec-compliance assessed
- Not checked: runtime integration with a live MCP session producing warnings
- Not checked: whether `drainBootWarnings` call sites need awareness of the trace-side render format
- Not checked: agent-registry path redundancy (`[agent-registry] [afk] agents:`) — noted as lesser inconsistency, separate from the doubled `[mcp]` defect

---

Issue #732, opened 2026-07-28
refactor(trace): split TraceWriter into a write-only TraceSink and an owner-only seal handle


griffinwork40:
Tracking the structural follow-up that #725 explicitly deferred, in its author's own words:

> Split `TraceWriter` into a write-only `TraceSink` for sessions and an owner-only handle carrying `seal`/`close`, so the capability is **removed at the type level** rather than gated at runtime — a session that cannot express `seal()` cannot get this wrong. That is a ~40-file signature sweep (167 `traceWriter` references, nearly all pass-through) and belongs in its own pure-refactor PR; mixing it with this behavioural change would make a regression unbisectable.

## Why this is worth doing

Seal ownership has now been answered wrong twice and re-litigated twice:

- **#666** — established *who* may seal (before it, the first descendant torn down sealed the whole shared file mid-run).
- **#725** — fixed *how that question is answered* (the gate had been probing `subagentToolOutputCapBytes`, an unrelated tool-output cap that merely happened to be fork-only).

Both are runtime gates on a capability every holder of the writer can still physically call. #725 now carries **two** independent runtime mechanisms for this — the `isSubagentFork` marker and a process-global `WeakSet` of claimed writers — which is defensible belt-and-braces but is also two things to keep correct forever.

A type-level split collapses both into "you were not given a handle that has `seal` on it." The failure mode stops being reachable rather than being guarded.

There is also a third route to the same bug that neither gate closes, which suggests the gate approach is not converging: the REPL `/resume` blackout (#731), where ownership transfers by *close* rather than by handoff.

## Scope

Measured on `main` @ `e8861a5`: **524** occurrences of `traceWriter` across **101** files (**61** non-test). The PR body estimates ~40 files / 167 references, presumably counting a narrower slice — either way, nearly all of it is pass-through plumbing, and the real design work is small:

- `TraceSink` — `write()` only. What `AgentConfig.traceWriter` and every `emit*()` helper accept.
- Owner handle — `TraceSink` plus `seal()` / `close()`. What `createDefaultTraceWriter()` returns to an entry point.

Entry points keep the owner handle; `AgentConfig` only ever carries the sink. `SubagentManager.forkSubagent` then threads a sink into children by construction, and `isSubagentFork`'s seal-gating job disappears (the field can stay if something else wants it, or go).

## Constraints

- Pure refactor, no behaviour change, its own PR — a mixed PR makes any regression unbisectable, which is exactly the property this subsystem needs most.
[...]

PR #985, opened 2026-08-10
fix(trace): give the resumed session a live writer instead of a sealed one (#731)

### LINKED ISSUE #731: fix(trace): REPL /resume silently blacks out the resumed session's trace
## Summary

After a REPL `/resume`, the resumed session writes **zero** records to the witness trace. The trace file ends at the outgoing session's `session_sealed` row, and everything the user does afterwards — every turn, tool call, and subagent dispatch — is silently dropped.

Pre-existing; **not** introduced by #725, and #725 does not make it worse. It is the same failure mode #666 and #725 address (a session sealing a writer another live session still needs), reached by a different route: seal ownership transfers by *close*, not by *handoff*.

## Mechanism

The REPL allocates one `TraceWriter` for the whole process and threads that same instance into the swap-built session:

- `src/cli/commands/interactive/bootstrap.ts:701` — `traceWriter: trace?.writer` goes into `sharedDeps`
- `src/cli/commands/interactive/bootstrap.ts:833-834` — `buildSession: (t) => buildAgentSession({ ...sharedDeps, ... })` spreads that same writer into the replacement session

`performResumeSwap` then closes the outgoing session *before* the pointer flip:

- `src/cli/commands/interactive/resume-swap.ts:163-166` — step 5, `await deps.sessionRef.current.close()` (the outgoing session owns the seal, so this seals the shared writer)
- `src/cli/commands/interactive/resume-swap.ts:~170` — step 6, atomic pointer flip to the new session

From step 6 onward the current session holds a sealed writer. `NdjsonTraceWriter.write()` throws on a sealed writer (`src/agent/trace/writer.ts:188-190`) and `emit.ts` swallows the rejection, so the loss is completely silent — no warning, no degraded-mode marker, nothing in the trace to indicate records are missing.

## Corroboration

Across **12,252** trace files on disk, `session_sealed` is the last row in **all 4,872** traces that contain it — 0 exceptions. Once a writer seals, nothing is ever recorded again, so there is no path by which the post-`/resume` session's records could be landing somewhere else in the file.

## Why the #725 ownership gate does not cover this

#725's gate ("only the session that first claimed a shared writer may seal it") is correct and does help here — on the *rollback* path, when `newSession.waitForInitialization()` rejects and `resume-swap.ts:143` closes the replacement session while the original is still live, the old gate let that replacement seal the writer out from under the survivor. The new gate prevents that.

But on the happy path the owner closes *first* and legitimately seals. The gate answers "who may seal", not "is anyone still going to need this writer".

## Possible directions

Not scoped here — listing so the tradeoffs are visible:

1. **Fresh writer per resumed session.** Simplest. Costs continuity: `/resume` splits one REPL run across two trace files, so `afk trace show` no longer shows the whole session. Arguably correct, since a resume *is* a new session with a new id.
2. **Transfer seal ownership at the pointer flip.** Keeps one file. Requires a real handoff API on the writer, which is close to the `TraceSink`/owner-handle split already proposed as a #725 follow-up (#732) — probably wants to land after that.
3. **Defer the outgoing session's seal when a successor holds the same writer.** Narrowest change, but re-introduces the "who seals if the successor never closes" question the process-exit backstop currently answers.

## Repro

```
afk interactive
> <any prompt, let it complete>
> /resume <some session>
> <another prompt>
afk trace show   # trace ends at the pre-/resume session_sealed row
```

Found during `/review` of #725.

griffinwork40:
## Problem

After a REPL `/resume`, the resumed session writes **zero** records to the witness trace. The trace file ends at the outgoing session's `session_sealed` row, and everything the user does afterwards — every turn, tool call, and subagent dispatch — is silently dropped.

Corroboration from the issue: across **12,252** trace files on disk, `session_sealed` is the last row in **all 4,872** traces that contain it — 0 exceptions. Once a writer seals, nothing is ever recorded again, so there was no path by which the post-`/resume` records could be landing elsewhere.

## Root cause

The REPL allocates one `TraceWriter` for the whole process and threads that same instance into the swap-built session:

- `src/cli/commands/interactive/bootstrap.ts:126` — `traceWriter: trace?.writer` goes into `sharedDeps`
- the `requestResume` `buildSession` closure spreads that same writer into the replacement session

`performResumeSwap` then closes the outgoing session *before* the pointer flip:

- `src/cli/commands/interactive/resume-swap.ts:163` — step 5, `await deps.sessionRef.current.close()` (the outgoing session owns the seal, so this seals the shared writer)
- `resume-swap.ts:168` — step 6, atomic pointer flip to the new session

From step 6 onward the current session holds a **sealed** writer. `NdjsonTraceWriter.write()` throws on a sealed writer (`src/agent/trace/writer.ts:187-189`) and `emit.ts` swallows the rejection, so the loss is completely silent — no warning, no degraded-mode marker, nothing in the trace to indicate records are missing.

### The complication the issue does not mention

Patching only `AgentConfig.traceWriter` is **not** sufficient. `SubagentManager`, `BackgroundAgentRegistry`, and the subagent / skill / compose executors are constructed once in `createBootstrapInfra` / `wireExecutors` and hold the writer in their own fields. They are never rebuilt on resume, so a fresh writer on the session alone would fix the resumed session's own turns while leaving post-resume `agent` / `skill` / `compose` dispatches writing into the sealed writer — the same silent drop, one layer down.

## Fix

**1 — a fresh writer per built session** (`bootstrap.ts`). The resume closure calls `createDefaultTraceWriter()` and passes the result as the new session's `traceWriter`. The writer is labelled with the resumed session's id, so it appends to that session's own witness directory — mirroring how a launch-time `--resume` labels its writer in `createBootstrapInfra`.

**2 — a `setTraceWriter` cascade**, mirroring the existing `setCwd` cascade in `dispatcher.ts`. Added to `SubagentManager`, `BackgroundAgentRegistry`, `SubagentExecutor` (which also re-points the root manager, as its `setCwd` does), `SkillExecutor`, and `ComposeExecutor`. Invoked from `onSwapped`.

### Ordering (recorded as an `// Invariant:` at the site)

The cascade point is not arbitrary — it is pinned on both sides:

- **Build time is too early.** The swap can still roll back (`waitForInitialization` rejects), which would leave the original session current with its executors re-pointed at a writer belonging to a session that was closed.
- **Anything after `onSwapped` is too late.** The first post-resume fork may already have been dispatched into the sealed writer.

`onSwapped` fires after the pointer flip commits, which is the only window satisfying both.

A rollback now also seals only the replacement session's own writer, leaving the survivor's untouched — previously the two were the same object.

## Tradeoff

`/resume` now splits one REPL run across two trace files, so `afk trace show` no longer shows the whole run in one file. This is accepted deliberately: a resume **is** a new session with a new id, so it earns its own trace file (direction 1 in the issue).

Directions 2 and 3 — transfer seal ownership at the pointer flip, or defer the outgoing seal — both require a real ownership-handoff API that **#732** (the `TraceSink` / owner-handle type split) is expected to provide, and the issue notes direction 2 "probably wants to land after that". Building a second runtime gate now is precisely the pattern #732 exists to delete. Direction 2 remains the end state once #732 lands.

I did **not** implement the outgoing→successor breadcrumb: it needs a new `SessionPhaseName` value plus a mirrored Zod enum entry, which is schema surface disproportionate to a nicety, and PR #984 (#754) is concurrently adding a value to that same closed union. Left out rather than silently bundled.

## Testing

```
pnpm lint

pnpm vitest run src/cli/commands/interactive/resume-trace-writer.test.ts \
  src/cli/commands/interactive/resume-swap.test.ts
# 2 files, 53 tests passed

pnpm vitest run src/agent/subagent.test.ts src/agent/background-registry.test.ts \
  src/agent/tools/subagent-executor.test.ts src/agent/tools/compose-executor.test.ts \
  src/agent/tools/skill-executor.test.ts
# 5 files, 441 tests passed
```

New `src/cli/commands/interactive/resume-trace-writer.test.ts` drives a **real `NdjsonTraceWriter`** against a temp dir — not a stub — so it tests the actual seam:

- sealing the outgoing writer rejects further writes to it, while the incoming writer still accepts them and its records land on disk in a distinct file;
- a resumed label writes into that session's own witness directory;
- all five long-lived holders re-point, including `SubagentExecutor` cascading to the root manager;
- `undefined` propagates when tracing is disabled (`AFK_TRACE_DISABLED=1`) rather than leaving the sealed writer in place.

Did not run `pnpm test:coverage` or the full suite (concurrent sibling worktrees V8-heap-OOM it); CI covers both.

## Reviewer notes

- In-flight children keep the writer they were forked with — the cascade only affects forks dispatched after the swap. That matches `setCwd`'s documented contract and avoids mutating a child mid-run.
- `readonly` was dropped on `SubagentManager.parentTraceWriter` and `BackgroundAgentRegistry.traceWriter`; both now carry a comment naming the resume swap as the reason.

Closes #731

### DIFF
--- a/src/agent/abort-graph.ts
+++ b/src/agent/abort-graph.ts
@@ -42,10 +42,19 @@ interface GraphNode {
 
 export class AbortGraph {
   private readonly nodes = new Map<string, GraphNode>();
-  private readonly traceWriter: TraceWriter | undefined;
+  private traceWriter: TraceWriter | undefined;
 
   constructor(traceWriter?: TraceWriter) {
     this.traceWriter = traceWriter;
+  }
+
+  /**
+   * Re-point the witness sink. Called (via the bootstrap cascade) on a REPL
+   * `/resume` so abort records emitted after the swap go to the live writer
+   * rather than the outgoing session's sealed one (#985).
+   */
+  setTraceWriter(writer: TraceWriter | undefined): void {
+    this.traceWriter = writer;
   }
 
   register(id: string, controller: AbortController): void {
--- a/src/agent/background-registry.ts
+++ b/src/agent/background-registry.ts
@@ -184,7 +184,10 @@ export class BackgroundAgentRegistry extends EventEmitter<BackgroundRegistryEven
    *  test workers (vitest runs files concurrently) don't produce
    *  colliding jobIds that share the same on-disk `bg/` directory. */
   private counter = Math.floor(Math.random() * 65536);
-  private readonly traceWriter: TraceWriter | undefined;
+  // Not readonly: a REPL `/resume` hands this long-lived registry a fresh
+  // writer via `setTraceWriter`, because the outgoing session sealed the one
+  // captured at construction (#731).
+  private traceWriter: TraceWriter | undefined;
   private readonly maxConcurrentJobs: number;
 
   constructor(options: BackgroundRegistryOptions = {}) {
@@ -201,6 +204,19 @@ export class BackgroundAgentRegistry extends EventEmitter<BackgroundRegistryEven
       5000,
     );
     sweepTimer.unref();
+  }
+
+  /**
+   * Re-point the writer this registry emits background-job lifecycle events
+   * into.
+   *
+   * Contract: called (via the bootstrap cascade) on a REPL `/resume`, where
+   * the outgoing session sealed the writer captured at construction. Jobs
+   * already in flight keep emitting into whatever writer they hold; only
+   * events emitted by the registry after this call use `writer` (#731).
+   */
+  setTraceWriter(writer: TraceWriter | undefined): void {
+    this.traceWriter = writer;
   }
 
   /**
--- a/src/agent/subagent.ts
+++ b/src/agent/subagent.ts
@@ -226,6 +226,7 @@ export class SubagentManager {
    */
   setTraceWriter(writer: TraceWriter | undefined): void {
     this.parentTraceWriter = writer;
+    this.abortGraph.setTraceWriter(writer);
   }
 
   /**
--- a/src/agent/subagent.ts
+++ b/src/agent/subagent.ts
@@ -114,7 +114,10 @@ export class SubagentManager {
   // manager lifetime. `undefined` value = resolved-and-there-is-none, so the
   // Map's `.has()` distinguishes "not yet resolved" from "resolved to none".
   private readonly worktreeMainRootCache = new Map<string, string | undefined>();
-  private readonly parentTraceWriter: TraceWriter | undefined;
+  // Not readonly: a REPL `/resume` swaps the session out from under this
+  // long-lived manager and hands it a fresh writer via `setTraceWriter`,
+  // because the outgoing session sealed the one captured here (#731).
+  private parentTraceWriter: TraceWriter | undefined;
   private readonly parentSurface: Surface | undefined;
   private readonly abortGraph: AbortGraph;
   private readonly rootId: string;
@@ -208,6 +211,21 @@ export class SubagentManager {
     // the cache (#441). setCwd is rare (born-named worktree creation on turn 1),
     // so forcing one re-resolution on the next fork costs nothing.
     this.worktreeMainRootCache.delete(cwd);
+  }
+
+  /**
+   * Re-point the writer inherited by future forks.
+   *
+   * Contract: mirrors {@link SubagentManager.setCwd} — existing in-flight
+   * children keep the writer they were forked with; only forks dispatched
+   * after this call inherit `writer`. Called (via the bootstrap cascade) on a
+   * REPL `/resume`, where the outgoing session seals the writer this manager
+   * captured at construction; without the re-point, every post-resume fork
+   * would emit its lifecycle events into a sealed writer and be silently
+   * dropped (#731).
+   */
+  setTraceWriter(writer: TraceWriter | undefined): void {
+    this.parentTraceWriter = writer;
   }
 
   /**
--- a/src/agent/tools/compose-executor.ts
+++ b/src/agent/tools/compose-executor.ts
@@ -541,6 +541,15 @@ export class ComposeExecutor {
     this.currentCwd = cwd;
   }
 
+  /**
+   * Re-point the trace writer compose DAG nodes inherit, after a REPL
+   * `/resume` replaced the session that owned the previous writer. Only nodes
+   * dispatched after this call use `writer` (#731).
+   */
+  setTraceWriter(writer: TraceWriter | undefined): void {
+    this.ctx.traceWriter = writer;
+  }
+
   async execute(call: ToolCall): Promise<ToolResult> {
     if (call.signal.aborted) {
       return { content: 'Compose tool call aborted', isError: true };
--- a/src/agent/tools/skill-executor.ts
+++ b/src/agent/tools/skill-executor.ts
@@ -31,6 +31,7 @@ import { buildSkillMaxDepthRefusal } from './skill-depth-message.js';
 import { appendRoutingDecision } from '../routing-telemetry.js';
 import { isTrustedSkill } from '../_lib/trusted-skill-registry.js';
 import { emitTrustedSkillComplete, emitTrustedSkillStart } from '../_lib/trusted-skill-events.js';
+import type { TraceWriter } from '../trace/writer.js';
 import type { SkillExecutorContext, SkillExecutorInternals, SkillInput } from './skill-executor/types.js';
 import { isGateSkill, sessionIdentity, truncateTelemetryString } from './skill-executor/telemetry.js';
 import { executeLoadedPluginSkill, executeLoadedRegistrySkill } from './skill-executor/load-mode.js';
@@ -103,6 +104,15 @@ export class SkillExecutor {
   setCwd(cwd: string): void {
     this.currentCwd = cwd;
     this.pluginBodies = null;
+  }
+
+  /**
+   * Re-point the trace writer skill forks inherit, after a REPL `/resume`
+   * replaced the session that owned the previous writer. Only forks dispatched
+   * after this call use `writer`; in-flight ones keep theirs (#731).
+   */
+  setTraceWriter(writer: TraceWriter | undefined): void {
+    this.ctx.traceWriter = writer;
   }
 
   /**
--- a/src/agent/tools/subagent-executor.ts
+++ b/src/agent/tools/subagent-executor.ts
@@ -314,6 +314,18 @@ export class SubagentExecutor implements SubagentControl {
   }
 
   /**
+   * Re-point the trace writer forked sub-agents inherit, after a REPL
+   * `/resume` replaced the session that owned the previous writer. Mirrors
+   * {@link SubagentExecutor.setCwd}: updates this executor's context AND the
+   * root manager that dispatches depth-1 forks, so the whole `agent`-tool tree
+   * follows the resumed session's live writer instead of the sealed one (#731).
+   */
+  setTraceWriter(writer: TraceWriter | undefined): void {
+    this.ctx.traceWriter = writer;
+    this.ctx.subagentManager.setTraceWriter(writer);
+  }
+
+  /**
    * The `agent` tool definition this executor's owning provider should
    * advertise. With a non-empty named-agent registry, the definition gains
    * the `agent_type` input property and an "Available agent types" listing
--- a/src/agent/trace/factory.ts
+++ b/src/agent/trace/factory.ts
@@ -21,6 +21,7 @@
 
 import { env } from '../../config/env.js';
 import { randomUUID } from 'node:crypto';
+import { readFileSync, existsSync } from 'node:fs';
 import { getTraceDir } from '../../paths.js';
 import { NdjsonTraceWriter, type TraceWriter } from './writer.js';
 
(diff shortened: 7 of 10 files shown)


griffinwork40:
Now I have the complete picture. The `setCwd` cascade targets `subagentExecutor`, `skillExecutor`, `composeExecutor` — the `setTraceWriter` cascade adds `rootManager` and `backgroundRegistry` on top, which is correct because those two also hold writer references but don't participate in the `setCwd` chain. Let me synthesize the review.

---

## Review: PR #985 — `fix(trace): give the resumed session a live writer instead of a sealed one (#731)`

**Verdict: Approve** — clean, well-bounded fix for a silent data-loss bug. Two minor observations below, no blockers.

### Summary

After a REPL `/resume`, the resumed session's witness trace was silently blanked out (zero records written). The REPL shared a single `TraceWriter` across sessions; `performResumeSwap` seals the outgoing session's writer (step 5) before the pointer flip (step 6), so the incoming session inherited a sealed writer. `write()` throws on sealed and `emit.ts` swallows the rejection — completely silent loss.

### Correctness ✅

The fix has two halves, both correctly implemented:

**1. Fresh writer per built session** (`bootstrap.ts`). The `buildSession` closure calls `createDefaultTraceWriter()` with the resumed session's label and stores it in `pendingTraceWriter`. The built session receives its own writer instance, breaking the shared-reference seal cascade.

**2. `setTraceWriter` cascade** across 5 long-lived holders (`SubagentManager`, `BackgroundAgentRegistry`, `SubagentExecutor`, `SkillExecutor`, `ComposeExecutor`). Invoked from `onSwapped` — the only window that satisfies both ordering constraints:
- **Not at build time** — the swap can still roll back (init rejects), which would leave executors re-pointed at a writer for a session that was closed.
- **Not after `onSwapped`** — the first post-resume fork may already fire into the sealed writer.

`SubagentExecutor.setTraceWriter` correctly cascades to both `this.ctx.traceWriter` AND `this.ctx.subagentManager.setTraceWriter`, mirroring the `setCwd` two-level pattern. In-flight children keep the writer they were forked with — no mid-run mutation.

The cascade targets are a **superset** of the `setCwd` cascade (`dispatcher.ts:536-538` hits the three executors; this adds `rootManager` and `backgroundRegistry`), which is correct — both of those hold independent writer references.

### API compatibility ✅

- `setTraceWriter` is a new public method on 5 classes, called only from `bootstrap.ts:onSwapped`. No interface or polymorphic dispatch is affected.
- `private readonly` → `private` on two fields (`SubagentManager.parentTraceWriter`, `BackgroundAgentRegistry.traceWriter`) is safe — both are `private`, so no external consumer can access them. Comments at each site name the reason.

### Test coverage ✅

`resume-trace-writer.test.ts` (138 lines, 7 tests) covers both halves:
- **Seam tests** use a *real* `NdjsonTraceWriter` against temp dirs — sealing the outgoing writer rejects its writes while the incoming writer still accepts, and records land on disk in distinct files.
- **Cascade tests** verify all 5 holders accept re-pointing (including `undefined` propagation for `AFK_TRACE_DISABLED`), and that `SubagentExecutor` cascades to the root manager.
- Temp dirs are properly cleaned up in `afterEach`.

| # | Severity | Detail |
|---|----------|--------|
| 1 | **nit** | The `pendingTraceWriter` variable is a closure-captured mutable shared between `buildSession` and `onSwapped`. This is safe today because the swap sequence is synchronous between those two callbacks, but a future refactor that makes them asynchronous could introduce a race. The invariant comment covers this well — just noting it as a future-reader concern, not a current bug. |
| 2 | **minor** | No outgoing→successor breadcrumb in the trace. `afk trace show` on a post-resume trace has no pointer back to the pre-resume segment. The PR description acknowledges this deliberately (needs a new `SessionPhaseName` value, and PR #984 is concurrently touching that union). Acceptable — not a regression, and the split-file trade is clearly the right call vs. the deferred-seal alternatives that #732 will address. |

### What I did not check

- Full test suite pass (PR says CI covers it; concurrent worktrees V8-heap-OOM the full suite locally — reasonable).
- Whether `afk trace list` / `afk trace show` correctly displays split traces post-resume (existing behavior, not changed by this PR).

**Done.** Clean fix, well-documented ordering invariants, correct cascade, adequate test coverage. Ship it.

---

griffinwork40:
## Codex findings — addressed in 6c92cd74

Both codex-flagged bugs were verified and fixed:

### P1 — Duplicate seq numbers on resume ✅ Fixed

**Root cause:** `NdjsonTraceWriter` always started `seq` at 0. When `/resume` creates a new writer for the same `traceDir`, it appended events starting at seq 0 after the outgoing session's sealed events (which ended at seq N), breaking the monotonic-sequence contract.

**Fix:** Added `startSeq` option to `NdjsonTraceWriterOptions`. `createDefaultTraceWriter` now reads the last seq from the existing `trace.jsonl` before constructing the writer, passing `startSeq: lastSeq + 1`.

Files: `writer.ts` (+9), `factory.ts` (+28)

### P2 — AbortGraph writer not re-pointed ✅ Fixed

**Root cause:** `SubagentManager.setTraceWriter` updated `parentTraceWriter` but left `AbortGraph` holding the outgoing (sealed) writer. Post-resume abort records from `abortGraph.abort()` → `emitAbort(this.traceWriter, ...)` were silently swallowed.

**Fix:** Added `setTraceWriter` to `AbortGraph` and wired `SubagentManager.setTraceWriter` to cascade to it.

Files: `abort-graph.ts` (+11), `subagent.ts` (+1)

### Testing

3 new test cases added to `resume-trace-writer.test.ts`:
- Resumed writer continues seq monotonically (real NdjsonTraceWriter, real temp dir)
- AbortGraph accepts setTraceWriter
- SubagentManager.setTraceWriter cascades to the abort graph

All 10 tests pass. 316 related tests pass (subagent, compose-executor, skill-executor, resume-swap). `tsc --noEmit` clean.

griffinwork40:
---

# Review: PR #985 — fix(trace): give the resumed session a live writer instead of a sealed one (#731)

## Triage

| Field | Value |
|-------|-------|
| **Change type** | bug fix (hotfix for #731) |
| **Branch** | `afk/issue-731-resume-trace` → `main` |
| **Files** | 10 (9 source + 1 new test) |
| **Lines** | +361 / −14 |
| **Reviewed ref** | `6c92cd74` |
| **Stated intent** | PR title + body |

## Findings

**low · blocking:false · confidence:high · dimension:correctness · `src/agent/trace/writer.ts:312` · ref:6c92cd74 · citation-type:file-state**
`sealOnProcessExit()` guards "nothing written" with `this.seq === 0`, but `startSeq > 0` breaks that semantic: a writer constructed with `startSeq: 3` that never writes has `seq = 3`, so the guard evaluates false. **Downgraded from medium:** unreachable through normal code paths — `sealOnProcessExit()` only fires for writers in `liveTraceWriters`, and registration requires `ensureOpen()` (triggered by `write()` or `seal()`); if `seal()` ran, `sealRecordPersisted` short-circuits; if `write()` ran, the backstop is correct behavior.
**Suggestion:** Store `startSeq` as a field and guard with `this.seq === this.startSeq`, or add a `hasWritten: boolean`.

**low · blocking:false · confidence:high · dimension:api-compat · `src/cli/commands/interactive/bootstrap.ts:205-230` · ref:6c92cd74 · citation-type:diff-context**
If `buildSession` throws or the swap rolls back after `pendingTraceWriter` is assigned, the writer is never owned by a session — a leaked file handle. The process-exit backstop seals it, but no explicit cleanup exists in the rollback path.
**Suggestion:** Reset `pendingTraceWriter = undefined` in the swap's rollback/failure handler, or seal the orphaned writer there.

**low · blocking:false · confidence:medium · dimension:correctness · `src/agent/trace/factory.ts:76` · ref:6c92cd74 · citation-type:diff-context**
`readLastSeq` accepts any `typeof seq === 'number'`, including `Infinity`, `NaN`, and non-integers from a corrupt final line. `startSeq: NaN + 1` → `NaN` breaks monotonic seq for every subsequent write.
**Suggestion:** Replace with `Number.isFinite(seq) && Number.isInteger(seq) && seq >= 0`.

**low · blocking:false · confidence:medium · dimension:perf-observability · `src/agent/trace/factory.ts:72-73` · ref:6c92cd74 · citation-type:diff-context**
`readLastSeq` calls `readFileSync` on the full trace file. Sub-second for typical 1 MB files today, but degrades linearly with file size and blocks the event loop during `/resume`.
**Suggestion:** Read only the last few KB via `fd`-based reverse-scan.

**low · blocking:false · confidence:high · dimension:test-coverage · `src/agent/trace/writer.ts` · ref:6c92cd74**
No test exercises `sealOnProcessExit()` with `startSeq > 0`. The existing "no-op when nothing written" test uses default `seq = 0`, so it never reaches the changed path.
**Suggestion:** Add a test case constructing `NdjsonTraceWriter({ traceDir, startSeq: 5 })`.

**nit · blocking:false · confidence:high · dimension:correctness · `src/agent/trace/factory.ts:92` · ref:6c92cd74 · citation-type:diff-context**
`createDefaultTraceWriter` constructs `tracePath` via template-literal (`\`${traceDir}/trace.jsonl\``) while `NdjsonTraceWriter` uses `join(this.traceDir, 'trace.jsonl')`. They resolve identically today but diverge on a trailing-slash `traceDir`.
**Suggestion:** Use `join(traceDir, 'trace.jsonl')` in the factory.

## Spec Compliance

Stated intent: PR title + body — spec-compliance assessed.

All three elements of the stated fix are implemented:
1. ✓ Fresh writer per resumed session via `createDefaultTraceWriter()` in `buildSession`
2. ✓ `setTraceWriter` cascade across all 6 long-lived holders in `onSwapped` (SubagentExecutor → SubagentManager → AbortGraph, SkillExecutor, ComposeExecutor, BackgroundAgentRegistry)
3. ✓ `startSeq` on `NdjsonTraceWriter` with `readLastSeq` in factory.ts for monotonic seq continuation

No scope creep. The outgoing→successor breadcrumb and `SessionPhaseName` addition were explicitly deferred in the PR body and are absent from the diff — consistent with the stated tradeoff.

## What Was Not Checked

- Citations verified inline against branch HEAD `6c92cd74`.
- Did not verify integration behavior under concurrent or rapid-fire `/resume` calls.
- Did not verify OS-level file-handle behavior when multiple orphaned writers accumulate from repeated swap failures.
- Did not run tests; CI gate covers both `pnpm test` and `pnpm test:coverage`.
- The `readLastSeq` behavior when the outgoing trace file is absent (new session, never wrote) was verified — returns `undefined`, which produces `startSeq: 0` (correct).

## Decision

**Decision: MERGE — 0 blocking (4 low, 1 low test-coverage gap, 1 nit).**

The highest-priority follow-up is the `NaN`/non-integer guard in `readLastSeq` (`factory.ts:76`) — a corrupt final NDJSON line with a non-finite `seq` would silently break monotonicity for the entire resumed trace. All other findings are hardening opportunities with no production impact today.

---

PR #986, opened 2026-08-10
feat(retention): bound the witness tree by age and aggregate size (#849)

### LINKED ISSUE #849: Witness tree has no retention policy - 461 MB / 12,596 dirs and growing
## Problem

Nothing prunes the witness tree. It grows without bound for the life of the install.

On one developer machine it is already **12,596 directories / 461 MB**, and that is *before* any opt-in artifact is enabled.

The only retention mechanism in the repo is `log-retention.ts::capJsonlBySize`, and it is wired solely to `session-grants.jsonl`:

```
$ git grep -n "witness" src/agent/log-retention.ts
(no matches)
```

Every other writer into `$AFK_STATE_DIR/witness/<sessionLabel>/` — the trace writer, compaction sidecars, and now the opt-in prompt capture added in #844 — writes forever with no cap, no TTL, and no sweep.

## Why it matters now

PR #844 adds `AFK_CAPTURE_SUBAGENT_PROMPTS=1`, which writes one markdown file per inbound child message under `witness/<sessionLabel>/prompts/`. That PR is **correctly shipped off by default**, and the absence of retention is one of the two reasons its author gave for that decision.

This issue is the explicit gate on ever changing that default. Flipping prompt capture (or any comparable artifact) to default-on without retention converts a disclosed tradeoff into an unbounded disk leak on every user's machine.

## Scope

Out of scope for #844 — this is pre-existing and repo-wide. Filed so the dependency stays legible rather than being rediscovered later.

## Suggested direction

- Extend the `capJsonlBySize` pattern (or add a sibling) to cover the witness tree, keyed by aggregate bytes and/or session age rather than per-file size alone.
- Decide the retention unit: whole session directories (simplest, preserves per-session coherence) vs. per-artifact caps within a session.
- Add a sweep entry point — session start is the natural hook, matching how other lazy maintenance runs.
- Once a cap exists, revisit whether `AFK_CAPTURE_SUBAGENT_PROMPTS` should stay opt-in.

## Acceptance

- Some mechanism bounds total witness-tree size or age.
- A test asserts the bound actually evicts.
- The retention policy is documented alongside the trace docs in `AFK.md`.

---
Filed from a review of #844.

griffinwork40:
## Problem

Nothing prunes the witness tree at `$AFK_STATE_DIR/witness/`. It grows without bound for the life of the install.

Measured on this machine today: **562 MB / 13,211 session directories** — and that is *before* any opt-in artifact is enabled. The only retention mechanism in the repo was `capJsonlBySize` (`src/agent/log-retention.ts`), wired solely to `session-grants.jsonl`; the trace writer, compaction sidecars, and the opt-in prompt/output captures all wrote forever with no cap, no TTL, and no sweep.

This issue is the standing gate on ever flipping `AFK_CAPTURE_SUBAGENT_PROMPTS` / `_OUTPUT` to default-on.

## Design

New module `src/agent/witness-sweep.ts` — the directory-tree sibling of `log-retention.ts`, wired at the same entry point: the `AgentSession` constructor, root sessions only (`parentSessionId === undefined`), fire-and-forget via `void`, beside the existing `capJsonlBySize` call.

**Retention unit is the whole session directory.** Per-artifact caps within a session were rejected: a trace stripped of its sidecars is a more confusing forensic record than no trace at all.

**Bounds** — evict by age first, then oldest-first until the aggregate byte cap is met:

| Env var | Default | Meaning |
|---|---|---|
| `AFK_WITNESS_MAX_AGE_DAYS` | `30` | Evict a session dir once its newest content is older than this |
| `AFK_WITNESS_MAX_BYTES` | `2147483648` (2 GiB) | Aggregate cap; oldest evicted first once exceeded |
| `AFK_WITNESS_RETENTION_DISABLE` | unset | `1` disables the sweep entirely |

**Cost bound.** The walk is O(files in tree), so a `.last-sweep` stamp file in the witness root throttles it to at most one walk every 6 hours — not one per session start. On all but one session start in that window the sweep is a stat of a single file.

## Behavior change — operator review required

**Merging this makes AFK delete old witness directories on session start.** That is the point of the issue, but it is a real deletion of forensic data and should be an explicit decision, not a surprise.

Measured **read-only** against the real 562 MB / 13,211-dir tree on this machine, under the shipped defaults:

```
total session dirs : 13211
WOULD EVICT (>30d) : 11151 dirs, 108 MB
would keep         :  2060 dirs, 413 MB
byte cap 2 GiB     : NOT reached — the byte pass is a no-op at this size
```

The nuance worth naming: **the age bound reclaims directory sprawl, not bytes.** 84% of the directories are older than 30 days but hold only 19% of the bytes — the tree is mostly a very large number of small, old session dirs plus a few recent heavy ones. Anyone hoping this fixes disk usage should set `AFK_WITNESS_MAX_BYTES` well below 2 GiB; the defaults are deliberately generous so that *installing this version* is not a mass delete.

## Safety properties (each has a test)

1. **The active session is excluded by identity** — its witness label (`sessionLabelFromTracePath`), never by timestamp — so its survival never depends on the sweep's heuristics being right.
2. **A 1-hour grace window** protects anything whose newest content is recent, regardless of the bounds. This covers the concurrent REPL / daemon / telegram sessions this process cannot enumerate.
3. **Liveness comes from the newest mtime across a directory's *contents*, not the directory's own mtime.** POSIX bumps a directory's mtime when an entry is created or unlinked, but **not** when an existing file inside it is appended to. A long-running session that creates no new sidecars (no subagents, no compaction) keeps appending to `trace.jsonl` while its directory mtime stays frozen at creation time — so an mtime-only sweep would evict a **live** session's trace out from under it. This is recorded as an `// Invariant:` on the walk.
4. **Never throws into the caller.** A missing root, permission error, or mid-walk race resolves to a zeroed result; retention housekeeping can never fail a session start.

## Also

Consolidates the two duplicate ad-hoc `getWitnessRoot()` helpers (`src/cli/commands/trace.ts:41`, `src/improve/paths.ts:66`) into `src/paths.ts`, the canonical owner of every AFK path. `improve/paths.ts` re-exports it so its consumers are unchanged.

## Testing

```
pnpm lint                 # tsc --noEmit clean, both tsconfigs
pnpm audit:env:check      # 882 files, 0 violations
pnpm scan:env:check       # docs/env-registry.{json,md} in sync (156 vars)

pnpm vitest run src/agent/witness-sweep.test.ts
# 10 tests passed

pnpm vitest run src/agent/witness-sweep.test.ts src/paths.test.ts \
  src/cli/commands/trace.test.ts src/config/env.test.ts
# 4 files, 140 tests passed

pnpm vitest run src/agent/session/ src/improve/
# 45 files, 729 tests passed
```

The issue's stated criterion is "a test asserts the bound **actually evicts**", so both bounds have a test proving a directory is gone from disk afterwards — not merely that the sweep ran:

- age bound evicts a 40-day-old dir and keeps a 2-day-old one;
- byte bound evicts oldest-first until the cap is met (asserting eviction *order*);
- active session survives even when over both bounds;
- grace-window dir survives however small the cap;
- the directory-mtime trap: a dir with an ancient directory mtime but fresh content is **kept**;
- `AFK_WITNESS_RETENTION_DISABLE=1` evicts nothing;
- missing root never throws;
- the stamp file skips a second sweep inside the interval.

Every test operates on a temp dir passed explicitly as `root` — nothing can touch the real `~/.afk`. Did not run `pnpm test:coverage` or the full suite (concurrent sibling worktrees V8-heap-OOM it); CI covers both.

## Docs

Retention policy documented in `AFK.md` alongside the trace/observability section, per the issue's third acceptance criterion.

Closes #849


griffinwork40:
## Agent Review — PR #986 `feat(retention): bound the witness tree by age and aggregate size (#849)`

**Reviewed ref:** `abd4f8b03325b1cbeb65d50bf7bb2e1a3f3d7c3d`  
**Stated intent:** PR #986 title + body — spec-compliance assessed.  
**Change type:** new-service | **Regime:** full | **Lines changed:** 624  
**Citations verified inline against branch HEAD `abd4f8b`.**

---

### Findings

#### Low

**1.** `severity · low · blocking: false · confidence: high · dimension: correctness · src/config/env.ts, src/paths.ts, src/agent/session/agent-session.ts · ref: abd4f8b · citation-type: file-state`

**Finding:** Three baselined files grew past their grandfathered `.filesize-baseline.json` LOC values without a baseline update, which trips the one-way ratchet.

**Evidence:**
```
src/config/env.ts:         baseline 1900, PR 1907 (+7)
src/paths.ts:              baseline 749,  PR 760  (+11)
src/agent/session/agent-session.ts: baseline 1442, PR 1461 (+19)
```

**Suggestion:** Run `pnpm audit:filesize:update` and commit the regenerated `.filesize-baseline.json` alongside this PR. The CI step runs `pnpm audit:filesize:check || true` so this won't block merge, but the ratchet is only useful if the recorded baseline stays current with each change.

---

#### Nit

**2.** `severity · nit · blocking: false · confidence: high · dimension: correctness · src/cli/commands/trace.ts:40 · ref: abd4f8b · citation-type: diff-context`

**Finding:** An orphaned JSDoc comment is left behind after the `getWitnessRoot()` private function was removed from `trace.ts`. The comment now floats above `TraceDirEntry`'s own doc comment with nothing to attach to.

**Evidence:**
```ts
/** Absolute path to the witness root that holds per-session trace dirs. */
/** One discovered trace, with the mtime used to order "most recent first". */
export interface TraceDirEntry {
```

**Suggestion:** Delete the first JSDoc comment (line 40). The documentation it carried is already present in `src/paths.ts:getWitnessRoot()`.

---

### What was not checked

- Citations verified inline against branch HEAD `abd4f8b03325b1cbeb65d50bf7bb2e1a3f3d7c3d`.
- Did not run `pnpm test` or `pnpm test:coverage` (read-only review).
- Did not profile walk performance against a real 13k-directory tree.
- Stated intent: PR #986 title + body — spec-compliance assessed. All three acceptance criteria are met: (1) tests assert the bound *actually evicts* on disk, (2) the active-session identity exclusion is tested and documented, (3) retention policy is documented in `AFK.md`.

---

### Decision: MERGE — 0 blocking (1 low, 1 nit).

The implementation is correct and safe. The sweep logic is sound: active-session exclusion by identity, 1-hour grace window for concurrent sessions, content-mtime walk to avoid the POSIX directory-mtime trap, fire-and-forget with `.unref()` off the session constructor path, and a stamp-file throttle that bounds the O(files) walk to once per 6 hours. Test coverage is thorough and operates entirely on temp dirs. The path consolidation (`getWitnessRoot` → canonical `src/paths.ts`) is clean and backward-compatible via the `improve/paths.ts` re-export.

The two findings above are non-blocking — a missing baseline update and an orphaned comment — and can be addressed in a follow-up commit or as a fixup before merge.

---

*Review by agent-afk `/review` skill • dim: security · api-compat · correctness · spec-compliance · test-coverage · perf-observability*

PR #991, opened 2026-08-10
feat(models): make claude-sonnet-4-6 the default sonnet model (+ correct its 1M/128k limits)

griffinwork40:
Makes `claude-sonnet-4-6` the default Sonnet model, and corrects the two limit pins that had to be fixed first for that to be safe.

Two commits, reviewable independently:

1. **`59e1ccb5`** — `fix(model-limits)`: correct 4.6's context window and output ceiling.
2. **`1998922f`** — `feat(models)`: point `sonnet` / `sonnet_1m` / `medium` at 4.6.

Commit 1 is a genuine prerequisite: with 4.6 still pinned at 200k, flipping the constant would have made `contextLimitFor('sonnet')` report 200k and broken the alias's own regression test.

## Commit 1 — the limit pins were wrong

| | before | after |
|---|---|---|
| context window | `200_000` | `1_000_000` |
| max output | `64_000` | `128_000` |

**The 1M window is GA and needs no beta header.** The 200k pin was defended by a comment arguing 1M on the 4.x line sits behind a `context-1m` beta this repo never sends, so 200k was "correct BY CONSTRUCTION." The construction argument was sound; its premise was stale. Per [Anthropic's context-windows doc](https://platform.claude.com/docs/en/build-with-claude/context-windows), which names Sonnet 4.6 in the 1M list:

> For every model with a 1M-token context window, 1M is the default: you don't need a beta header, and long-context requests are billed at standard pricing.

The `anthropic.com` launch post still says "1M token context window in beta" — that predates GA and is the likely origin of the pin. **No entry was added to `composeBetaHeader`, because none is required.**

**128k output, not 64k.** The 64k pin was inferred from the `claude-sonnet-5` entry's own comment, "128k max output (up from 64k on Sonnet 4.6)" — which was itself wrong. Per the [migration guide](https://platform.claude.com/docs/en/about-claude/models/migration-guide):

> **128k max output tokens (unchanged):** Claude Sonnet 5 supports up to 128k output tokens, the same as Claude Sonnet 4.6.

The misleading Sonnet 5 comment is corrected in the same commit so the bad inference is not re-derivable. A `MODEL_AUTOCOMPACT_BUDGET` entry (200k) is added for 4.6, matching the policy already applied to `claude-sonnet-5` and `claude-opus-5` — now that the window is truthfully 1M, that entry does real work.

## Commit 2 — 4.6 becomes the default Sonnet

```
/model sonnet    -> claude-sonnet-4-6   ctx=1,000,000  compactAt=  200,000  maxOut=128,000
/model sonnet_1m -> claude-sonnet-4-6   ctx=1,000,000  compactAt=1,000,000  maxOut=128,000
/model medium    -> claude-sonnet-4-6   ctx=1,000,000  compactAt=  200,000  maxOut=128,000
/model opus      -> claude-opus-5       (unchanged)
/model haiku     -> claude-haiku-4-5-20251001 (unchanged)
```

`sonnet_1m` keeps its distinct meaning: same wire model, but the `_1m` suffix bypasses the 200k working budget for the full 1M window.

**Why a code change rather than config:** `sonnet` and `sonnet_1m` are `DIRECT_MODEL_ALIASES` entries, deliberately *not* rebindable by slot config — `models.medium` / `AFK_MODEL_MEDIUM` only move the capability tiers. Every subagent dispatched with `model: 'sonnet'` (research-agent, git-investigator, compose nodes) tracks this constant and nothing else, so flipping it is the only mechanism that retargets them.

**Capability is lateral, not a downgrade:** after commit 1, both models are 1M context / 128k output. Sonnet 5's tokenizer emits ~30% more tokens for identical input, so 4.6's 1M window holds proportionally more real text.

**Cost:** 4.6 is $3/$15 per MTok; Sonnet 5 is on introductory $2/$10 **through 2026-08-31**, then $3/$15. So this is ~50% more per token for the remainder of August — but only ~15% per equivalent content once the tokenizer difference is accounted for, and at parity from September 1.

**Thinking routing is deliberately untouched.** `requiresAdaptiveThinking` matches `(opus|sonnet)-5`, not `-4-6`, so a 4.6 session uses manual extended thinking (`{type:'enabled'}`) rather than adaptive. Not a break — 4.6 accepts manual extended thinking; the 400 error only appears going 4.6 → 5 — and it reproduces the behaviour this repo had before `da492d45` flipped the default the other way.

`AFK_MODEL_MEDIUM`'s registry example stays `claude-sonnet-5`: still a valid id, and it now illustrates binding the tier to something other than the default, which avoids regenerating the CI-gated `docs/env-registry.{json,md}` pair for a cosmetic string.

## Verification

- `pnpm lint` clean (`tsc --noEmit`, both projects).
- Full suite: **15971 passed**, 1 failure in `write-denylist.test.ts` (a `/tmp` symlink `rmSync` case) that is **pre-existing and unrelated** — it fails identically with this branch stashed, and that file does not import `model-limits`. Test count is unchanged across the flip, i.e. zero regressions.
- All five audit gates pass: `audit:sdk:check`, `audit:env:check`, `scan:env:check`, `audit:chalk:check`, `fix:pins:check`.
- Resolution table above produced by executing the real resolver, not by reading the constants.

## Deliberately out of scope

`pricing.ts` lists `claude-sonnet-5` at $3/$15, but its introductory rate is $2/$10 until 2026-08-31 — so AFK currently over-reports Sonnet 5 spend by ~50%. Unrelated to this diff; worth its own issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ afk chat "refactor this" --model gpt-5.5
 |---|---|---|
 | `local` | *(empty — you configure)* | Point at Ollama, LM Studio, or any OpenAI-compatible shim via `AFK_MODEL_LOCAL` + `AFK_MODEL_LOCAL_BASE_URL` |
 | `small` | `claude-haiku-4-5-20251001` | Cheapest/fastest Anthropic tier |
-| `medium` | `claude-sonnet-5` | General-use default |
+| `medium` | `claude-sonnet-4-6` | General-use default |
 | `large` | `claude-opus-5` | Most capable |
 
 The `haiku`/`sonnet`/`opus`/`fable` handles are **fixed identities**, not tier
--- a/docs/model-slots.md
+++ b/docs/model-slots.md
@@ -87,7 +87,7 @@ An unconfigured install behaves exactly as before this feature:
 | -------- | ------------------------------- | ------------------------- |
 | `local`  | `` (empty — user-configured)    | —                         |
 | `small`  | `claude-haiku-4-5-20251001`     | `haiku`                   |
-| `medium` | `claude-sonnet-5`             | `sonnet`, `sonnet_1m`     |
+| `medium` | `claude-sonnet-4-6`          | `sonnet`, `sonnet_1m`     |
 | `large`  | `claude-opus-5`                 | `opus`, `opus_1m`         |
 
 ## Identity aliases vs. capability tiers
--- a/src/agent/model-limits.test.ts
+++ b/src/agent/model-limits.test.ts
@@ -201,33 +201,47 @@ describe('maxOutputTokensFor — retired-but-Active Opus pin', () => {
 
 describe('claude-sonnet-4-6 — explicit limit pins (A/B baseline vs claude-sonnet-5)', () => {
   // Invariant: 4.6 is not a first-class alias but IS reachable by raw wire id and
-  // already priced (providers/anthropic-direct/pricing.ts). Both values below
-  // equal the fallbacks they replace, so these tests pin INTENT, not a behaviour
-  // change: they fail if someone edits DEFAULT_MAX_OUTPUT / DEFAULT_CONTEXT_LIMIT
-  // or "upgrades" 4.6 to Sonnet 5's 1M window, either of which would silently
-  // invalidate a 4.6-vs-5 A/B comparison.
-  it('pins the 64k output ceiling (first-party: the Sonnet 5 entry says "up from 64k on Sonnet 4.6")', () => {
-    expect(maxOutputTokensFor('claude-sonnet-4-6')).toBe(64_000);
-    // Contrast: the Sonnet 5 sibling is double. An A/B that silently gave 4.6
-    // Sonnet 5's ceiling would compare two different output budgets.
+  // already priced (providers/anthropic-direct/pricing.ts). Both pins below now
+  // DIFFER from the fallbacks they replace, so they guard real behaviour: 4.6
+  // shares Sonnet 5's 1M window and 128k output ceiling, and inheriting either
+  // fallback (200k context / 64k output) would under-report a live capability.
+  it('pins the 128k output ceiling — the same as Sonnet 5, not half of it', () => {
+    // First-party: migration guide, 4.6 → 5 — "128k max output tokens
+    // (unchanged): Claude Sonnet 5 supports up to 128k output tokens, the same as
+    // Claude Sonnet 4.6." A prior revision read Sonnet 5's "up from 64k on Sonnet
+    // 4.6" comment as first-party and pinned 64k; that comment was itself wrong.
+    expect(maxOutputTokensFor('claude-sonnet-4-6')).toBe(128_000);
     expect(maxOutputTokensFor('claude-sonnet-5')).toBe(128_000);
+    // Guard the specific regression: 4.6 must not silently inherit the 64k
+    // DEFAULT_MAX_OUTPUT fallback.
+    expect(maxOutputTokensFor('claude-sonnet-4-6')).not.toBe(64_000);
   });
 
-  it('pins the 200k context window — NOT Sonnet 5 1M (this repo sends no 1M beta header)', () => {
-    // 1M on the 4.x line requires a `context-1m` beta this codebase never sends:
-    // composeBetaHeader (providers/anthropic-direct/beta-headers.ts) emits a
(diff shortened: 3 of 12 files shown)


griffinwork40:
# Review: PR #991 — `claude-sonnet-4-6` as default sonnet

## Recommendation: **APPROVE** with follow-ups — no blockers

The load-bearing content is **factually correct, and it fixes a pin that was factually wrong**. The pre-PR state was the dangerous one.

### The one finding both review passes under-weighted (MAJOR)

`resolve-params.ts:28-29` — `requiresAdaptiveThinking` matches `sonnet-5` but **not** `sonnet-4-6`. No line in that file changed, but the **default path through it did**: every unconfigured session moves from adaptive thinking to `{type:'enabled'}`.

Routing 4.6 to `enabled` is correct for the model. The issue is what it re-opens — from the same file's own #951 comment (`:145-153`):

> …the real trigger is an explicit output cap <= 1024 … on a non-adaptive model (haiku, fable-5, **raw sonnet-4-6**); opus-5 and **sonnet-5 still route to adaptive regardless of the cap**.

That comment was written when sonnet-5 *was* the default, and it calls 4.6 "raw" (opt-in). Post-PR, raw sonnet-4-6 **is** the default. A user with a stale `AFK_MAX_OUTPUT_TOKENS <= 1024` was previously rescued by adaptive routing and now takes a **thrown Error every turn** — their config didn't change, the default under it did. Conditional, not universal; the clamp bug is #951's, not this PR's.

### Verified rather than assumed

All four external claims **CONFIRMED verbatim** against live Anthropic docs by two independent verifiers: 1M context (Sonnet 4.6 named explicitly in the 1M list), 128k output, no beta header, ~30% tokenizer delta in the stated direction.

The citation check **inverted the risk**: the retired comment this PR deletes claimed — also citing "first-party" — that 128k was *"up from 64k on Sonnet 4.6."* **That phrasing exists nowhere in Anthropic's docs** (the only 64k output figure belongs to Haiku 4.5). The code being replaced was the fabricated-citation revision. Given the asymmetry the retired comment itself named, pre-PR 4.6 was under-reported at 200k/64k — silently capping output at half its real ceiling.

Propagation is complete (`MODEL_MAP` is derived, not hardcoded); 4.6 **is** priced at `pricing.ts:80` ($3/$15, no cost delta); default-user runtime delta is ~zero on compaction and +64k output headroom.

### Lower-severity

- **MINOR** `model-limits.ts:259` — "both the `sonnet` alias and a literal `claude-sonnet-5` requestedModel hit it" is now **false**, two lines from where the PR falsified it. Plus stale comments at `providers/index.ts:38`, `resolve-params.ts:106`.
- **MINOR** `context-limit-1m.test.ts:88` pairs `model: 'claude-sonnet-5'` with `requestedModel: 'sonnet'` — incoherent post-PR, and it's the test guarding this exact invariant.
- I **downgraded** another dimension's MAJOR on `env-registry` to NIT: it's an `example` field, still a valid id, both env gates pass, and the commit documents the choice. Disagreement recorded, not smoothed.

### CI

`pnpm lint` clean; all 8 targeted files pass; `pnpm test` = 15970 passed / 2 failed, both **pre-existing** (byte-identical to base, no import overlap). ⚠️ One of them (`write-denylist.test.ts:654`, `rmSync(symlink→dir)` → `EISDIR` on Node 24.11/macOS) fails deterministically **at base**, so `main` is red on this platform — "passes CI" assumes CI's Node/OS differs. Worth its own issue.

**Done**
- Reviewed PR #991 across five dimensions, then ran a three-verifier adversarial wave re-deriving the docs claims and the test attribution independently; personally verified the adaptive-thinking boundary that both passes missed.
- Evidence: full review at `/Users/griffinlong/.afk/workspace/reports/review-pr-991-sonnet-4-6-default.md`; diff artifact at `~/.afk/state/skill-preflight/141faeef-696b-4db5-bffc-d9d92ae49589/pr-991.diff`; gate output reproduced from `pnpm lint` / `pnpm test` on worktree `1998922f`.
- World unchanged in the repo: review was read-only — no files in the worktree were edited, nothing committed, nothing posted to GitHub. The only write was the report outside the repo.
- Pending: the four follow-ups are recommendations only — I did not implement them, since `/review` is read-only and the MAJOR one (#951 exposure) is a judgment call about whether to ship a comment fix plus release note or add bootstrap validation.
- Not posted to the PR: `--post` wasn't requested; say the word and I'll push this as a review comment.

---

PR #524, opened 2026-07-12
feat(hooks): observe-only release-boundary PreToolUse detector (gate-migration wave 1 slice 2)

griffinwork40:
## What

Wave 1 **slice 2** of the friction-substrate / gate-skill→hook migration program — the direct sibling of #492 (safe-destruct). Migrates the *detection* half of the `release-boundary-gate` gate-skill into a deterministic **observe-only** PreToolUse hook.

A new hook watches `bash` for commands that cross:
- a **publish/deploy boundary** — `npm`/`pnpm`/`yarn`/`cargo`/`twine`/`poetry`/`gem` publish, `docker push`, `gh release create`, `terraform`/`kubectl apply`;
- a **sync boundary** — `git push --mirror` / `--tags` / `--follow-tags`.

On a match it emits a `hook_decision` catch-record via the otherwise-unused `decision: 'approve'` outcome — **without blocking**. Pattern classes are taken from the user-scope `release-boundary-gate` SKILL.md's two boundary tracks.

## Why observe-only (unchanged from slice 1)

PreToolUse **cannot** `injectContext` (the harness honors that field only for `SubagentStop`/`UserPromptSubmit`), and a hard block would repeat the interpreter-eval backfire that started this program — worse here, since a release/deploy is often exactly what was asked, so a block would be a false positive by construction. So this slice changes **zero behavior**: it only records that a boundary-crossing command was attempted. The records are the shadow window; a later slice uses their real-world per-pattern frequency to calibrate the skill's real value (a pre-boundary living-artifact check), not a block. The `approve` record is filterable in the trace and ignored by the mechanical friction detectors, so it never masquerades as new friction.

## Sequencing notes (for the reviewer)

- **Building release-boundary-gate, not target-lock** (the plan's next-listed item). The proven observe-only pattern is *"match a command string → `approve` record"* (stateless, deterministic). release-boundary-gate fits it cleanly; target-lock does not — the plan itself describes target-lock as *"edit/write/rm gated on a session verdict-flag"*, i.e. inherently a stateful **block** gate whose real trigger (a colloquially-named refactor target) lives in the user's prompt, invisible to a PreToolUse hook. target-lock's natural first form is a later enforce slice.
- **safe-destruct calibration deferred, not skipped** — only 3 catch-records exist across all traces since #492 merged (~1 day); that is not a calibratable dataset. It accrues passively while this independent detector banks its own shadow signal.

## Changes
[...]

PR #992, opened 2026-08-10
feat(hooks): two-tier OBSERVE/BLOCK split in safe-destruct PreToolUse detector

griffinwork40:
## What

Splits `safe-destruct-detect`'s flat pattern table into two tiers. OBSERVE keeps returning `decision: 'approve'` (never blocks, emits a catch-record). BLOCK returns `decision: 'block'` with an actionable reason. One hook, one registration — `registry.count('PreToolUse') === 3` is unchanged.

This is the first slice of the gate-migration program that actually **enforces** anything. Both shipped detectors (#492, #524) have been observe-only, which meant a gate-class skill was replaced by a silent logger.

## Why these tiers — the calibration is measured, not preferred

The detector has been logging for five weeks. Its own catch-records (359 `approve` events across 165 sessions, from `~/.afk/state/witness/**/trace.jsonl`):

| pattern | firings | share |
|---|---|---|
| `rm-recursive-force` | 326 | 91% |
| `git-branch-force-delete` | 13 | 4% |
| `git-push-force` | 6 | 2% |
| `git-reset-hard` | 5 | 1% |
| everything else | 0 | — |

Two things fall out of that distribution:

1. **A blanket promotion to block was never viable.** It would have blocked ~326 legitimate operations — the `rm -rf` firings are concentrated in recurring cron jobs (`pr-review-sweep`, `weekly-recap`), i.e. routine cleanup, dozens per run.
2. **The dangerous set is the low-volume tail.** Roughly 29 events in five weeks, about one every 1.5 days. That is cheap to gate.

So the split is by *recoverability*, not by how alarming the command looks.

| tier | patterns |
|---|---|
| **BLOCK** (unrecoverable or externally irreversible, never inner-loop) | `rm-no-preserve-root`, `git-reset-hard`, `git-clean-force`, `git-push-force`, `dd-to-device`, `mkfs`, `redirect-to-block-device`, `shred`, `sql-drop-truncate`, `terraform-destroy` |
| **OBSERVE** (high-volume routine, or recoverable) | `rm-recursive-force` (+ `-split`, `-long`), `git-branch-force-delete`, `find-delete`, `docker-destructive`, `kubectl-delete`, `sql-delete-from` |

The judgment calls a reviewer will question, each recorded as an `// Invariant:` comment beside the pattern so a future reader does not "fix" them:

- **`rm -rf` stays OBSERVE despite being destructive** — purely the 326-firing base rate. This is the whole point of the PR.
- **`git branch -D` stays OBSERVE** — only the ref is lost; the commit objects survive in the reflog.
- **`docker`/`kubectl` stay OBSERVE** — that state is typically declarative and re-creatable, and `kubectl delete pod` is inner-loop for many users.
- **`terraform destroy` blocks** — it tears down real external infrastructure and is never inner-loop.

`sql-drop-truncate-delete` is **split** into `sql-drop-truncate` (BLOCK) and `sql-delete-from` (OBSERVE). The old id bundled DDL destructors with `DELETE FROM`, which is routine development SQL and must not block. That id had 0 firings, so no telemetry history is affected.

## Block reasons are load-bearing

A PreToolUse block reaches the agent only as `dispatcher.ts`'s tool result — `{ content: 'Tool "bash" blocked by PreToolUse hook: <reason>', isError: true }`. That string is the entire interface, and the agent can retry next turn, so each reason names the pattern, why it is blocked, and the real alternative:

> `safe-destruct: blocked [git-reset-hard] — discards all uncommitted changes irrecoverably; commit or run "git stash" first. This hook cannot be self-bypassed: if the destruction is genuinely intended, stop and ask the operator to run it.`

The second commit exists because the first draft ended six reasons with "proceed with the intent stated explicitly." There is no intent-based bypass, so an agent following that advice would re-run the identical command and be blocked again — a loop. A constraint the agent can talk itself past is not a constraint; the reasons now say so plainly and route the genuine case to the operator.

## Precedence

A compound command matching both tiers resolves to BLOCK (`git reset --hard && rm -rf build` blocks). Tested explicitly — one unrecoverable sub-operation must not slip through by being mixed with a recoverable one.

## Structure

`DESTRUCTIVE_PATTERNS` moved to a new `src/agent/safe-destruct-patterns.ts` to keep both files under the 350-LOC limit (detector 123, patterns 211, tests 233).

## Verification

- `pnpm lint` clean (`tsc --noEmit`, both projects)
- `pnpm test src/agent/safe-destruct-detect.test.ts` — 74 passed (one positive case per pattern per tier, plus the precedence case and the registry-wiring variants)
- `pnpm test src/agent/hooks/config-bridge.test.ts` — 19 passed, PreToolUse count assertion intact
- `pnpm test src/agent/default-hook-registry.test.ts` — 6 passed
- `pnpm audit:env:check`, `scan:env:check`, `audit:sdk:check`, `audit:chalk:check`, `fix:pins:check` — all OK

No existing assertion was weakened. The previous `NEVER blocks` test used `rm -rf / --no-preserve-root`, a target that moves to BLOCK; it was split into per-tier cases rather than deleted.

## Follow-ups, deliberately not in this PR

- `release-boundary-detect` is still observe-only. `npm publish` is the strongest remaining block candidate — externally irreversible, 5 firings in five weeks.
- Re-measure before moving anything OBSERVE → BLOCK. The asymmetry is the design.

### DIFF
--- a/src/agent/safe-destruct-detect.test.ts
+++ b/src/agent/safe-destruct-detect.test.ts
@@ -1,7 +1,7 @@
 /**
  * Tests for the two-tier safe-destruct detector.
  *
- * Four contracts:
+ * Five contracts:
  *   1. `detectDestructiveCommands` matches every curated pattern and does NOT
  *      flag common-but-safe near-misses.
  *   2. OBSERVE patterns return `decision: 'approve'` (never block).
@@ -65,6 +65,10 @@ describe('detectDestructiveCommands', () => {
     ['git clean --force -x', 'git-clean-force'],
     ['git push --force origin main', 'git-push-force'],
     ['git push -f', 'git-push-force'],
+    ['git -C /tmp/repo reset --hard HEAD', 'git-reset-hard'],
+    ['git --work-tree=/tmp/w reset --hard', 'git-reset-hard'],
+    ['git -C /other push --force origin main', 'git-push-force'],
+    ['git --git-dir=.git clean -fd', 'git-clean-force'],
     ['dd if=/x.img of=/dev/sda bs=1M', 'dd-to-device'],
     ['mkfs.ext4 /dev/sdb1', 'mkfs'],
     ['echo boot > /dev/sda', 'redirect-to-block-device'],
@@ -87,6 +91,8 @@ describe('detectDestructiveCommands', () => {
     ['git status'],
     ['git commit -m "wip"'],
     ['git push origin main'], // no force
+    ['git push --force-with-lease origin main'], // safe lease variant
+    ['git push --force-with-lease=origin/main origin main'], // =<refname> form
     ['git branch -d merged'], // lowercase -d is the safe delete
     ['ls -la /var'],
     ['npm install'],
--- a/src/agent/safe-destruct-detect.test.ts
+++ b/src/agent/safe-destruct-detect.test.ts
@@ -1,14 +1,20 @@
 /**
- * Tests for the observe-only safe-destruct detector (Wave 1, gate-migration).
+ * Tests for the two-tier safe-destruct detector.
  *
- * Three contracts:
- *   1. `detectDestructiveCommands` matches the curated catastrophic patterns
- *      and — critically for calibration — does NOT flag common-but-safe near
- *      misses (`rm -r build/`, `rm -f lock`, `git branch -d`, `truncate -s 0`).
- *   2. The hook returns an `approve` catch-record (never a block) only on a
- *      `bash` PreToolUse with a destructive command; `{}` otherwise.
- *   3. It is wired into the default registry and, dispatched end-to-end, passes
- *      the command through (no throw) while recording `decision: 'approve'`.
+ * Four contracts:
+ *   1. `detectDestructiveCommands` matches every curated pattern and does NOT
+ *      flag common-but-safe near-misses.
+ *   2. OBSERVE patterns return `decision: 'approve'` (never block).
+ *   3. BLOCK patterns return `decision: 'block'` with a reason that names the
+ *      pattern id.
+ *   4. When a single compound command matches both tiers, BLOCK wins.
+ *   5. Registry wiring: the hook is registered exactly once, produces approve
+ *      for OBSERVE patterns, and throws HookBlockedError for BLOCK patterns.
+ *
+ * Note on the former 'NEVER blocks' test (~line 97-101 in the old file):
+ *   It used `rm -rf / --no-preserve-root` as the fixture — that pattern moved
+ *   to the BLOCK tier. The test has been split into two: one asserting OBSERVE
+ *   patterns still approve, one asserting BLOCK patterns now block.
  */
 
 import { describe, it, expect } from 'vitest';
@@ -18,6 +24,7 @@ import {
   detectDestructiveCommands,
   SAFE_DESTRUCT_DETECT_REASON_PREFIX,
 } from './safe-destruct-detect.js';
+import { HookBlockedError } from '../utils/errors.js';
 import { createDefaultHookRegistry } from './default-hook-registry.js';
 
 function preCtx(command: string, toolName = 'bash'): HookContext {
@@ -25,7 +32,12 @@ function preCtx(command: string, toolName = 'bash'): HookContext {
   return ctx;
 }
 
+// ---------------------------------------------------------------------------
+// detectDestructiveCommands — pattern matching
+// ---------------------------------------------------------------------------
+
 describe('detectDestructiveCommands', () => {
+  // ── OBSERVE patterns ───────────────────────────────────────────────────────
   it.each([
     ['rm -rf /tmp/foo', 'rm-recursive-force'],
     ['rm -fr build', 'rm-recursive-force'],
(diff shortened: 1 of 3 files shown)


griffinwork40:
Wave 1.5 verification complete. Key results:

**Citation verification manifest:**
1. `git-push-force` regex matching `--force-with-lease` — **verified**: `true` at reviewed ref. The regex and blockReason are confirmed at lines 107–118 of `safe-destruct-patterns.ts`.
2. `blockReason?: string` optional — **verified**: line 34 at reviewed ref.
3. Fallback reason at `safe-destruct-detect.ts:112` — **verified**: line 112 at reviewed ref.
4. `patternMap` allocation at lines 106–107 — **verified** at reviewed ref.
5. "Four contracts" header listing five — **verified**: lines 4–13 of test file at reviewed ref.
6. `sql-drop-truncate-delete` id renamed — **verified**: only exists in test assertions (lines 113, 117) and a comment (line 168). Zero production importers.
7. `SAFE_DESTRUCT_DETECT_REASON_PREFIX` — **verified**: zero production importers outside the module itself.
8. `PreToolUse === 3` assertion — **verified** present in `config-bridge.test.ts:379` (absence claim about it being untested is **false-absent** — DROP).

- "No `--force-with-lease` near-miss test" — **confirmed-absent**: `git grep 'force-with-lease' f5c0aeb8 -- src/agent/safe-destruct*` returns only the blockReason string, zero test fixtures.
- "No BLOCK+BLOCK compound test" — grep for compound test patterns shows only the OBSERVE+BLOCK case. **confirmed-absent**.
- "`registry.count('PreToolUse') === 3` untested" — **false-absent**: `config-bridge.test.ts:379` already asserts this. **DROPPED**.

Now Wave 2 — synthesis. I have all findings. Let me compile the final report.

---

# PR #992 Review — `feat(hooks): two-tier OBSERVE/BLOCK split in safe-destruct PreToolUse detector`

**Target:** `afk/iso-block-tier-1-uaxmu8` → `main` | **Ref:** `f5c0aeb8` | **Regime:** full
**Files:** `safe-destruct-detect.ts`, `safe-destruct-detect.test.ts`, `safe-destruct-patterns.ts` (new) | **+391 / -135**

---

## Findings

### 1. `git push --force-with-lease` false-positive block
**high · blocking:true · confidence:high · correctness+security · `src/agent/safe-destruct-patterns.ts:107–118` · ref:f5c0aeb8 · file-state [verified inline]**

The `git-push-force` BLOCK regex `/\bgit\s+push\b[^|&;\n]*--force\b/i` matches `--force-with-lease` because `\b` fires on the `force`/`-` boundary (non-word character). Confirmed at the reviewed ref:

```
node -e "console.log(/\bgit\s+push\b[^|&;\n]*--force\b/i.test('git push --force-with-lease origin main'))"
→ true
```

The blockReason itself tells the agent to *"prefer '--force-with-lease'"* — the recommended recovery path triggers the block. This is a BLOCK-tier pattern, so the command is hard-refused, not just logged.

**suggestion:** `--force\b` → `--force(?!-)` in both alternation arms. Add `['git push --force-with-lease origin main']` to the benign-commands table.

---

### 2. `blockReason` not enforced at the type level for BLOCK-tier patterns
**low · blocking:false · confidence:high · spec-compliance+correctness · `src/agent/safe-destruct-patterns.ts:23–35` · ref:f5c0aeb8 · file-state [verified inline]**

`DestructivePattern` marks `blockReason?: string` — optional for all tiers. The stated intent says block reasons are load-bearing and must name the pattern, the reason, and the alternative. The fallback (`safe-destruct: blocked [${id}] — unrecoverable operation.`) satisfies none of those three requirements. All 10 current BLOCK entries supply `blockReason`, so this is latent.

**suggestion:** Discriminated union: `{ tier: 'observe' } | { tier: 'block'; blockReason: string }` makes the fallback branch unreachable.

---

### 3. `patternMap` allocated per-call on the hot path
**low · blocking:false · confidence:high · correctness · `src/agent/safe-destruct-detect.ts:106–107` · ref:f5c0aeb8 · file-state [verified inline]**

`new Map(DESTRUCTIVE_PATTERNS.map(p => [p.id, p]))` is built inside `safeDestructDetect()` on every matching call, then discarded. The calibration data says this fires 83–326 times/day. Not a production risk at this volume, but the allocation is unnecessary.

**suggestion:** Hoist to module scope in `safe-destruct-patterns.ts`: `export const PATTERN_MAP = new Map(DESTRUCTIVE_PATTERNS.map(p => [p.id, p]));`. Alternatively, use `DESTRUCTIVE_PATTERNS.find(p => matchedIds.includes(p.id) && p.tier === 'block')`.

---

### 4. Compound multi-BLOCK commands report only the first BLOCK pattern's reason
**low · blocking:false · confidence:high · correctness · `src/agent/safe-destruct-detect.ts:107–115` · ref:f5c0aeb8 · file-state [verified inline]**

When a command matches multiple BLOCK patterns (`terraform destroy && mkfs.ext4 /dev/sda`), the loop returns on the first hit in table order. The agent sees only one reason. Not a safety gap (the command is still blocked), but the feedback is incomplete.

**suggestion:** Document the "first BLOCK in table order wins" invariant in the precedence comment, or accumulate all BLOCK reasons into the returned reason string.

---

### 5. `sql-drop-truncate-delete` id renamed without telemetry migration
**low · blocking:false · confidence:high · api-compat · `src/agent/safe-destruct-patterns.ts:168` · ref:f5c0aeb8 · diff-context**

The monolithic `sql-drop-truncate-delete` id is split into `sql-drop-truncate` and `sql-delete-from`. The `Invariant:` comment in the same file states *"Pattern ids are stable public identifiers... Changing an id breaks deduplication across the 165-session history. New ids are append-only."* The old id had 0 firings (stated in PR body), so no actual trace data is orphaned — the documented invariant is violated in letter but not in effect.

**suggestion:** Either add a comment noting the 0-firing exception to the append-only rule, or keep the old id as a dead-letter alias.

---

### 6. Test header says "Four contracts" but lists five
**nit · blocking:false · confidence:high · spec-compliance · `src/agent/safe-destruct-detect.test.ts:4` · ref:f5c0aeb8 · diff-context**

```
 * Four contracts:
 *   1. ... 2. ... 3. ... 4. ... 5.
```

**suggestion:** "Four" → "Five".

---

### 7. No near-miss test for `git push --force-with-lease`
**low · blocking:false · confidence:high · test-coverage · `src/agent/safe-destruct-detect.test.ts` (absence) · ref:f5c0aeb8 · confirmed-absent**

The benign-commands table has no entry for `--force-with-lease`. Grep at the reviewed ref returns zero test fixtures containing the string. This is the test gap that would have caught Finding 1.

**suggestion:** Add `['git push --force-with-lease origin main']` to the `does not flag benign` table.

---

### 8. Echo/comment false positives escalated from OBSERVE to BLOCK
**low · blocking:false · confidence:high · correctness · `src/agent/safe-destruct-patterns.ts:96–100` · ref:f5c0aeb8 · file-state [verified inline]**

`echo 'git reset --hard HEAD'` and `# git reset --hard HEAD` match BLOCK patterns. Previously these were OBSERVE-only catch-records; now they hard-block. This is a structural limitation of regex-based command scanning — not fixable without context parsing — but the severity escalated from "noisy telemetry" to "false block" with the tier split.

**suggestion:** Document as a known constraint. No regex change needed unless echo/comment false-blocks appear in production telemetry.

---

## What was not checked
- Citations verified inline against branch HEAD `f5c0aeb8cb462ed416b3b4ca98ecbe9ec51cb63f`.
- Stated intent: PR #992 title+body — spec-compliance assessed.
- Did not verify runtime behavior of `dispatchPreToolUse` or `HookBlockedError` propagation through the dispatcher.
- Did not review Telegram or daemon surfaces for tier-specific behavior differences.
- One Wave 1 absence claim (`registry.count('PreToolUse') === 3` untested) was **dropped** — refuted by `config-bridge.test.ts:379`.

---

## Decision: DO NOT MERGE — 1 high blocking; 7 non-blocking (5 low, 1 nit)

Finding 1 (`--force-with-lease` false positive) is the sole blocker. The pattern's own block reason recommends the command it blocks — a self-contradicting gate that traps the agent in a loop with no viable exit. The regex fix is one character class (`--force\b` → `--force(?!-)`), and the test gap (Finding 7) is the regression guard.

Findings 2–8 are quality improvements; none block merge independently.

---

PR #996, opened 2026-08-10
fix(tools): classify a nonexistent grep path as no-such-target

griffinwork40:
## Symptom

Ripgrep exits 2 for every "something went wrong" cause, so `grep` returned all of them identically: `grep error: rg: /abs/path.ts: No such file or directory (os error 2)`, `isError: true`, **no `failureClass`**. An unclassified failure renders as a red `✗` in the REPL tool lane and counts toward failure-density stats.

The most common way to produce one is not a bug — it is ordinary exploration. Measured on session `9e70e82c` (2026-08-10): a verifier sub-agent issued 88 path arguments and **5 pointed at files that do not exist** (`src/cli/config-schema.ts`, `src/cli/session-stats.ts`, `src/config/types.ts`, `src/cli/commands/interactive/session-builder.ts`). Across all sessions that day, `grep` showed 68 errors in 3,127 calls — a meaningful share of which are this. The operator reading that lane cannot tell "the agent guessed a filename" from "the tool broke".

This is the generalization of #75: an agent probing a target that isn't there should not read as something going wrong.

## What changed

`no-such-target` joins the failure-class vocabulary and the benign set, and grep's exit-2 branch learns to tell a missing path from a real fault.

- **`src/agent/trace/types.ts`** — `'no-such-target'` added to `TOOL_FAILURE_CLASSES` and to `BENIGN_FAILURE_CLASSES`, with the per-class JSDoc bullet and both benign/non-benign prose blocks updated so the comments cannot drift from the code.
- **`src/agent/tools/handlers/_rg-exit2.ts`** (new, 98 lines) — one pure function mapping rg's stderr + the resolved search path to the exit-2 result. Classifies `no-such-target` **only** when every non-empty stderr line is an ENOENT line and one of them names the search path.
- **`src/agent/tools/handlers/grep.ts`** — the `code === 2` branch delegates to it, and the `settle()` result type moves into the new module. Net **-3 lines** (350 → 347), staying under the ceiling.

### Why `isError` stays `true`

Deliberate. A nonexistent path must never read as "no matches" — that turns a typo into a false "this code does not exist" conclusion, which is strictly worse than a visible error. The model still learns the path was wrong; only the human-facing glyph and the stats treatment change.

### Why a new file

`grep.ts` sat at exactly 350 lines, the project's hard per-file ceiling — so the first draft of this change pushed it to 353 and had to move the `settle()` result type out too. The rule is to pull a whole concern out rather than raise the limit, and the stderr-shape matching is worth unit-testing without spawning a real ripgrep.

### Conservative by construction

`os error 13` (permission denied) is shaped almost identically to `os error 2`; anchoring on the trailing `(os error 2)` sentinel is what keeps it out of the benign bucket. A mixed stderr (missing path **plus** another failure) stays generic — reporting it as merely a missing path would hide the real fault. If ripgrep ever changes this wording the classifier degrades to today's generic branch, never to a wrong classification. The stderr spellings in the comment were verified against the bundled rg, not guessed.

## Verification

```
pnpm install --prefer-offline                       # Done in 1.4s
pnpm lint                                           # clean (tsc --noEmit + tsconfig.web.json)
pnpm test src/agent/tools/handlers/_rg-exit2.test.ts \
         src/agent/tools/handlers/grep.test.ts \
         src/cli/commands/interactive/tool-lane-format.test.ts   # 205 passed
pnpm test src/agent/trace/events.test.ts \
         src/agent/trace/receipt.test.ts \
         src/improve/scan/detectors/tool-failure-density.test.ts # 105 passed  (310 total across the 6 files)
```

- New `_rg-exit2.test.ts` (8 cases): both ENOENT spellings, surrounding whitespace, permission-denied stays generic, regex parse error stays generic, mixed stderr stays generic, ENOENT naming a different path stays generic, empty/whitespace stderr stays generic.
- Every consumer of `TOOL_FAILURE_CLASSES` / `BENIGN_FAILURE_CLASSES` was swept. One needed updating: the exhaustiveness guard in `tool-lane-format.test.ts` that pins the benign partition — exactly the tripwire it exists to be.
- The existing grep-handler case that greps a nonexistent path was converted to assert `failureClass === 'no-such-target'` alongside `isError === true`.

Not checked: no full-suite run (scoped files only); no manual REPL check that the lane now renders `⊘` — that path is covered by `tool-lane-format`'s own tests.

Context: `~/.afk/workspace/reports/diagnosis-2026-08-10-shadow-verify-tool-call-volume.md`.

### DIFF
--- a/src/agent/tools/handlers/_rg-exit2.test.ts
+++ b/src/agent/tools/handlers/_rg-exit2.test.ts
@@ -1,0 +1,79 @@
+import { describe, it, expect } from 'vitest';
+import { classifyRgExit2 } from './_rg-exit2.js';
+
+const PATH = '/repo/src/cli/session-stats.ts';
+
+describe('classifyRgExit2', () => {
+  it('classifies the single-path ENOENT shape as no-such-target', () => {
+    // rg's spelling when it is handed exactly one path that is absent.
+    const stderr = `rg: ${PATH}: IO error for operation on ${PATH}: No such file or directory (os error 2)\n`;
+    const result = classifyRgExit2(stderr, PATH);
+
+    expect(result.failureClass).toBe('no-such-target');
+    expect(result.isError).toBe(true);
+    expect(result.content).toContain(PATH);
+    // The remedy must be actionable: name the tool that resolves paths.
+    expect(result.content).toContain('glob');
+    expect(result.content).not.toContain('grep error');
+  });
+
+  it('classifies the short ENOENT shape as no-such-target', () => {
+    const stderr = `rg: ${PATH}: No such file or directory (os error 2)`;
+    const result = classifyRgExit2(stderr, PATH);
+
+    expect(result.failureClass).toBe('no-such-target');
+    expect(result.content).toContain(PATH);
+  });
+
+  it('tolerates surrounding whitespace and blank lines', () => {
+    const stderr = `\n  rg: ${PATH}: No such file or directory (os error 2)  \n\n`;
+    expect(classifyRgExit2(stderr, PATH).failureClass).toBe('no-such-target');
+  });
+
+  it('leaves a permission failure unclassified (os error 13 is not os error 2)', () => {
+    // Shaped almost identically to the ENOENT line — the trailing-sentinel
+    // anchor is what keeps it out of the benign bucket.
+    const stderr = `rg: ${PATH}: Permission denied (os error 13)`;
+    const result = classifyRgExit2(stderr, PATH);
+
+    expect(result.failureClass).toBeUndefined();
+    expect(result.isError).toBe(true);
+    expect(result.content).toContain('grep error');
+    expect(result.content).toContain('Permission denied');
+  });
+
+  it('leaves a regex parse error unclassified', () => {
+    const stderr = 'rg: regex parse error:\n    (unclosed\n    ^\nerror: unclosed group';
+    const result = classifyRgExit2(stderr, PATH);
+
+    expect(result.failureClass).toBeUndefined();
+    expect(result.content).toContain('grep error');
+    expect(result.content).toContain('unclosed group');
+  });
+
+  it('stays generic when a missing path is mixed with another failure', () => {
+    // A mixed stderr means something beyond a bad path went wrong; reporting it
+    // as merely a missing path would hide the real fault.
+    const stderr =
+      `rg: ${PATH}: No such file or directory (os error 2)\n` +
+      'rg: /repo/other.ts: Permission denied (os error 13)';
+    const result = classifyRgExit2(stderr, PATH);
+
+    expect(result.failureClass).toBeUndefined();
+    expect(result.content).toContain('grep error');
+  });
+
+  it('stays generic when the ENOENT line names some other path', () => {
+    const stderr = 'rg: /elsewhere/gone.ts: No such file or directory (os error 2)';
+    expect(classifyRgExit2(stderr, PATH).failureClass).toBeUndefined();
+  });
+
+  it('stays generic on empty or whitespace-only stderr', () => {
+    for (const stderr of ['', '   ', '\n\n']) {
+      const result = classifyRgExit2(stderr, PATH);
+      expect(result.failureClass).toBeUndefined();
+      expect(result.isError).toBe(true);
+      expect(result.content).toContain('grep error');
+    }
+  });
+});
--- a/src/agent/tools/handlers/_rg-exit2.ts
+++ b/src/agent/tools/handlers/_rg-exit2.ts
@@ -1,0 +1,98 @@
+/**
+ * Classifier for ripgrep's exit-code-2 stderr, shared by the grep handler.
+ *
+ * rg exits 2 for every "something went wrong" cause: a search path that does
+ * not exist, a path it may not read, an unparseable regex. Those are not the
+ * same event. A nonexistent path is a caller-supplied bad reference — ordinary
+ * exploration, a typo, stale memory of a moved file — while a regex error or a
+ * permission failure is worth a human's attention. Splitting them lets the
+ * former carry `failureClass: 'no-such-target'` (benign: neutral glyph in the
+ * tool lane, excluded from failure-density stats) while the latter stays an
+ * unclassified, alarming failure.
+ *
+ * Lives in its own module because `grep.ts` sits at the project's 350-LOC
+ * ceiling, and because the stderr-shape matching below is worth unit-testing
+ * without spawning a real ripgrep.
+ *
+ * @module agent/tools/handlers/_rg-exit2
+ */
+
+import type { ToolFailureClass } from '../../trace/types.js';
+import { capForModel } from './_output-cap.js';
+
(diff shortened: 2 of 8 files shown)


griffinwork40:
@codex fix comments

griffinwork40:
Good. `resolveAndContain` calls `path.resolve()` (line 102), which produces a normalized absolute path (no trailing slash, no `..`). And rg echoes the path it was given verbatim in stderr. Since grep.ts passes `resolvedPath` to **both** rg (as the search path argument) and `classifyRgExit2` (as `searchPath`), the same string flows to both sides of the `includes()` check. Finding 2's normalization concern is unfounded — the path is already normalized by `path.resolve()` before being passed to either rg or the classifier.

**Finding 2 status:** The theoretical concern about path normalization mismatch is **refuted** — `resolvedPath` is the output of `path.resolve()` and is passed identically to both rg's command line and `classifyRgExit2`. rg echoes the path verbatim. Downgrading to `nit`.

Now let me synthesize the final report.

---

## PR #996 Review — `fix/grep-no-such-target` → `main`

**Ref:** `56ae90b0` · **Branch:** `fix/grep-no-such-target` · **Change type:** feature · **Regime:** light (282 lines, 8 files)

### Summary

Adds a `no-such-target` failure class so that `grep` against a nonexistent path (the most common exit-2 cause, ~5.7% of calls in measured sessions) renders a neutral glyph instead of a red `✗` in the REPL tool lane and is excluded from failure-density stats. `isError` stays `true` to prevent a typo from reading as "no matches." A new `_rg-exit2.ts` module holds the classifier; `receipt.ts` consolidates onto the canonical `BENIGN_FAILURE_CLASSES` set; `types.ts` gains the new class with full documentation. 8 new test cases cover the classifier; existing tests updated for the new behavior.

### Findings

**correctness · spec-compliance · api-compat · security** — no issues found at the reviewed ref. The diff does exactly what the intent describes: `receipt.ts` replaces its local exempt set with the canonical `BENIGN_FAILURE_CLASSES`, `types.ts` adds the new class to both arrays, the classifier is conservative by construction, and the test exhaustiveness guard is updated.

---

`low · blocking:false · confidence:high · dimension:test-coverage · file:src/agent/tools/handlers/_rg-exit2.test.ts · ref:56ae90b0 · citation-type:diff-context`
**Finding:** No test covers the truncation path in the generic (non-ENOENT) branch — `classifyRgExit2` calls `capForModel(stderr.trim())` and propagates `truncated: true`, but no test verifies that flag is set when stderr exceeds `MODEL_CAP_BYTES`.
**Evidence:** 8 test cases in the new file; none supplies a large stderr string.
**Suggestion:** Add a test with stderr exceeding `MODEL_CAP_BYTES` asserting `result.truncated === true`.

---

`low · blocking:false · confidence:high · dimension:test-coverage · file:src/agent/tools/handlers/_rg-exit2.test.ts · ref:56ae90b0 · citation-type:diff-context`
**Finding:** No test exercises `classifyRgExit2` when `searchPath` contains characters that are special in regex contexts (e.g., parentheses, brackets). The classifier uses `String.prototype.includes()` which is safe, but no test documents this property, leaving it vulnerable to a future refactor that introduces regex matching.
**Evidence:** 8 test cases; `PATH` is `/repo/src/cli/session-stats.ts` — no special characters. Grep at the ref confirmed no additional test: `git grep 'classifyRgExit2' 56ae90b0 -- '*.test.*'` returns only the 8-case file.
**Suggestion:** Add one test where `PATH` contains parentheses to pin the `includes()` behavior.

---

`low · blocking:false · confidence:high · dimension:perf-observability · file:src/improve/scan/detectors/tool-failure-density.ts:90 · ref:56ae90b0 · citation-type:file-state [verified at ref]`
**Finding:** `EXCLUDED_FAILURE_CLASSES` in the tool-failure-density detector aliases `BENIGN_FAILURE_CLASSES`. With `no-such-target` now benign, a model stuck in a missing-path loop (grepping the same nonexistent file repeatedly) generates no density card — the loop is invisible to the detector.
**Evidence:** `const EXCLUDED_FAILURE_CLASSES: ReadonlySet<ToolFailureClass> = BENIGN_FAILURE_CLASSES;` (line 90). The `no-such-target` class is now in `BENIGN_FAILURE_CLASSES` (line 154 of `types.ts`).
**Suggestion:** This is a deliberate tradeoff (the PR body acknowledges `no-such-target` is benign). If observability is desired, add a separate `noSuchTargetCount` field to the session summary outside the density detector, or add a `repeat-failure` circuit breaker for consecutive identical `no-such-target` calls.

---

`nit · blocking:false · confidence:medium · dimension:api-compat · file:src/agent/tools/handlers/_rg-exit2.ts:28,36 · ref:56ae90b0 · citation-type:diff-context`
**Finding:** `GrepSettleResult` and `RgExit2Result` are exported from a `_`-prefixed (module-private by convention) file. Only `grep.ts` imports them. Exporting types from a `_`-prefixed file creates a surface area that contradicts the naming convention. [UNVERIFIED: no production importers beyond `grep.ts` — git grep at ref confirms only one consumer]
**Suggestion:** Leave `classifyRgExit2` exported (consumed by `grep.ts`); the type exports are harmless but could be unexported since `grep.ts` uses only `import type`.

---

4 findings from Wave 1 were dropped — the sub-agent read the working tree (pre-merge) rather than the reviewed ref:
- "receipt.ts still uses local REVIEW_EXEMPT_FAILURE_CLASSES" — **fabricated**: at the reviewed ref, `receipt.ts` imports `BENIGN_FAILURE_CLASSES` (line 30).
- "types.ts does not contain no-such-target" — **fabricated**: present at lines 72 and 154.
- "grep.test.ts still asserts /grep error/" — **fabricated**: at the reviewed ref, the test asserts `failureClass === 'no-such-target'`.
- "exempt-missing-target does not appear in receipt.ts" — **false-absent**: it's a `toolUseId` in the test fixture, not a rendering label; the assertion is valid.

### What was not checked

- Citations verified inline against branch HEAD `56ae90b0` (fetched and resolved locally).
- Stated intent: PR #996 title + body — spec-compliance assessed; no unmet requirements, no scope creep.
- Not checked: full test suite run (PR body reports scoped passing tests). No manual REPL check of the glyph rendering.
- Not checked: `_rg-exit2.ts` behavior with rg versions other than the bundled 15.x.
- Not checked: whether the `denial-breaker` or `repeat-failure` circuit breakers fire on repeated `no-such-target` calls (orthogonal to this PR's scope).

### Decision: MERGE — 0 blocking (3 low, 1 nit).

**Done**

- Reviewed PR #996 (`fix/grep-no-such-target`) at ref `56ae90b0`, light regime.
- 4 findings survived verification: 3 low (2 test-coverage gaps, 1 observability note), 1 nit (export convention). None blocking.
- 4 Wave 1 findings were dropped as fabricated/false-absent — the sub-agent read the working tree instead of the reviewed ref; Wave 1.5 inline verification caught all four.
- Merge recommendation: **MERGE** — the change is correct, well-tested, conservative by construction, and does exactly what the stated intent describes.

---

PR #997, opened 2026-08-10
feat(web): warn when markdown extraction retains little of the page

griffinwork40:
## Symptom

`web_scrape` markdown mode extracts via Readability, which strips collapsed, tabbed, and accordion regions along with real page chrome. When it drops a section the caller wanted, the tool still returns a healthy-looking body and reports **success** — the model gets no signal that anything is missing.

What the model does with no signal is retry. Measured on session `9e70e82c` (2026-08-10), a sub-agent fetched `platform.claude.com/docs/en/about-claude/models/overview`, received 27KB of markdown that was missing a collapsed "still available models" table, and re-fetched **the same url 5+ times** — markdown → `.md` variant → `raw` → larger `max_bytes` → anchor fragment — until the host began refusing connections. Outcome: 7 × `web_scrape network error: fetch failed` plus one 30s timeout, from one missing table.

`web_scrape` runs a **24.5% error rate** (97 errors / 396 calls on 2026-08-10), the worst of any tool, and same-url retry storms are the largest contributor. The existing render escalation cannot catch this: it triggers only below an absolute floor (`THIN_CONTENT_CHARS = 200`), and these pages are far above it.

## What changed

One advisory line appended to an otherwise successful result.

- **`src/web/extraction-advisory.ts`** (new, 110 lines) — `visibleTextLength()`, `extractionAdvisory()`, `withAdvisory()`. Pure, no I/O.
- **`src/web/scrape.ts`** — `ScrapeResult` gains an optional `advisory?: string`, set on the Phase-2 non-thin fetch return: the exact path where extraction looked healthy so no escalation ran. +21 lines (259 → 280).
- **`src/agent/tools/handlers/web-scrape.ts`** — markdown mode appends the advisory to the content. +3 lines (347 → 350).

Example output appended below the markdown:

> `[web_scrape: extraction kept ~11% of this page's visible text (412 of ~3600 chars). Readability strips collapsed, tabbed, and accordion sections along with page chrome, so a section you expect may be absent from the markdown above. If something specific is missing, re-request this url once with mode: "raw" and read the HTML — do not re-fetch it in markdown mode.]`

## Behavior is unchanged

No new fetch, no new render, no change to `THIN_CONTENT_CHARS`, to when escalation happens, or to any thrown error or `isError` classification. The advisory is a string appended to output that was already successful, and `advisory` is optional — every existing consumer compiles and behaves identically.

## Thresholds and why

The denominator is the source's **visible-text length, not its byte length**. Byte length is unusable: doc sites ship large inline JSON/JS/CSS (a Mintlify page is mostly hydration state), so a complete extraction can retain under 5% of bytes while a half-dropped page on a lean site retains 30%. Text-to-text makes the ratio mean "of the words a human would see, this fraction survived".

| Constant | Value | Reasoning |
|---|---|---|
| `MIN_SOURCE_TEXT_CHARS` | 2,000 | Under this the ratio is noise — on a 300-char page one nav label swings it wildly with nothing at stake — and `THIN_CONTENT_CHARS` already covers small results. |
| `MIN_RETAINED_RATIO` | 0.5 | Above half retained, the loss is consistent with ordinary boilerplate stripping (nav, sidebar, footer, cookie banner), i.e. extraction working. Below it, most of the visible page is gone, which is the signature of a dropped content region rather than trimmed chrome. |

Both are a **first calibration**, chosen conservatively so the advisory stays rare enough to be worth reading, and both are exported constants that are easy to retune. The re-measurement recipe is named in the module comment: run extraction over a sample of real pages and compare `textLength` against `visibleTextLength(html)`. Erring toward over-advising is the cheap direction — the cost of a false positive is one line of text on a successful result; the cost of a false negative is the retry storm above.

`visibleTextLength` is a regex pass, not a second DOM parse: it runs on every successful markdown scrape, and paying for another JSDOM construction to compute a heuristic's denominator would cost more than the heuristic is worth. Slight over-counting biases the ratio down, i.e. toward advising.

## Verification

```
pnpm install --prefer-offline        # Done in 1.3s
pnpm lint                            # clean (tsc --noEmit + tsconfig.web.json)
pnpm test src/web/extraction-advisory.test.ts              # 11 passed
pnpm test src/web/scrape.test.ts                           # 21 passed
pnpm test src/agent/tools/handlers/web-scrape.test.ts      # 45 passed
pnpm test src/web/                                         # 125 passed (6 files)
```

- `extraction-advisory.test.ts` (11 cases): fires on large + lossy; silent on a normal article; boundary pair just above/below the ratio floor; silent below the absolute floor; degenerate input (empty html, zero source, extraction longer than source) returns `undefined` with no divide-by-zero; script/style/comment payloads excluded from the denominator.
- `scrape.test.ts` (+3): a nav/aside/footer-dominated page yields an advisory **and still takes the fetch path with no render**; an ordinary article yields none; a JSON body yields none (no extraction ran).
- `web-scrape.test.ts` (+1): the advisory reaches model-visible content and the result is not an error.
- No test touches the network or a real browser — existing `fetchFn` / `renderFn` / `lookupFn` seams only.

**Flag for the next editor:** `src/agent/tools/handlers/web-scrape.ts` is now at exactly 350 lines, the project's per-file ceiling. The next change to it must pull a concern out. Relatedly, `withAdvisory` is applied *after* `capBody`, deliberately — inside the byte budget the advisory would be the first thing head+tail truncation drops, on exactly the large pages most likely to trip it. That ordering is documented in `withAdvisory`'s docstring; do not "tidy" it.

Context: `~/.afk/workspace/reports/diagnosis-2026-08-10-shadow-verify-tool-call-volume.md`.

### DIFF
--- a/src/agent/session/subagent-output-capture-wiring.test.ts
+++ b/src/agent/session/subagent-output-capture-wiring.test.ts
@@ -116,10 +116,18 @@ describe('subagent output capture wiring', () => {
       await drainTurn(session, 'second');
       const files = await waitForOutputFiles(sessionId, 1);
       expect(files).toHaveLength(1);
-      const body = fs.readFileSync(
-        path.join(getSubagentOutputsDir(sessionId), 'multi-turn-child.md'),
-        'utf8',
-      );
+      // Capture writes are fire-and-forget (async appendFile chained off the
+      // event loop). The file exists after the first turn, but the second
+      // turn's append may not have flushed yet — poll for the expected content
+      // rather than reading once.
+      const filePath = path.join(getSubagentOutputsDir(sessionId), 'multi-turn-child.md');
+      const deadline = Date.now() + 2000;
+      let body = '';
+      while (Date.now() < deadline) {
+        body = fs.readFileSync(filePath, 'utf8');
+        if ((body.match(/### assistant/g) ?? []).length >= 2) break;
+        await new Promise((r) => setTimeout(r, 10));
+      }
       // Header written once; two turns recorded into the same file.
       expect(body.match(/subagentId:/g)?.length).toBe(1);
       expect((body.match(/### assistant/g) ?? []).length).toBeGreaterThanOrEqual(2);
--- a/src/agent/tools/handlers/web-scrape.test.ts
+++ b/src/agent/tools/handlers/web-scrape.test.ts
@@ -146,6 +146,37 @@ describe('web_scrape handler — markdown mode (fetch-first)', () => {
     expect(renderFn).not.toHaveBeenCalled();
   });
 
+  it('surfaces the extraction advisory in the model-visible content, still as a success', async () => {
+    // A page whose visible text is dominated by regions Readability discards, so
+    // extraction succeeds while most of the page never reaches the model. The
+    // advisory is what stops the model re-fetching the same url in markdown mode.
+    const filler = (label: string, n: number): string =>
+      Array.from(
+        { length: n },
+        (_, i) =>
+          `<p>${label} row ${i + 1}: this text exists in the source document and a human ` +
+          `reading the page in a browser would see it rendered on screen.</p>`,
+      ).join('');
+    const lossy =
+      '<!DOCTYPE html><html><head><title>Docs</title></head><body>' +
+      `<nav>${filler('nav', 40)}</nav>` +
+      '<article><h1>Docs</h1><p>The one short paragraph Readability keeps as the article ' +
+      'body, padded just enough to clear the thin-content floor and skip the render.</p>' +
+      '<p>A second sentence of genuine article prose so extraction has a real region.</p>' +
+      '</article>' +
+      `<aside>${filler('aside', 40)}</aside><footer>${filler('footer', 40)}</footer>` +
+      '</body></html>';
+
+    const renderFn = vi.fn<RenderFn>(async () => ({ html: '', finalUrl: '', httpStatus: 200 }));
+    const fetchFn = makeFetch(() => makeResponse({ contentType: 'text/html', body: lossy }));
+    const handler = createWebScrapeHandler({ fetchFn, env: {}, renderFn, lookupFn: publicLookup });
+
+    const r = await handler({ url: 'https://example.com/docs' }, signal());
+    expect(r.isError).toBeUndefined();
+    expect(r.content).toContain('[web_scrape: extraction kept');
+    expect(r.content).toContain('mode: "raw"');
+  });
+
   it('requires no API key for markdown mode', async () => {
     const fetchFn = makeFetch(() => makeResponse({ contentType: 'text/html', body: richArticleHtml() }));
     const handler = createWebScrapeHandler({ fetchFn, env: {}, lookupFn: publicLookup });
--- a/src/agent/tools/handlers/web-scrape.test.ts
+++ b/src/agent/tools/handlers/web-scrape.test.ts
@@ -171,10 +171,12 @@ describe('web_scrape handler — markdown mode (fetch-first)', () => {
     const fetchFn = makeFetch(() => makeResponse({ contentType: 'text/html', body: lossy }));
     const handler = createWebScrapeHandler({ fetchFn, env: {}, renderFn, lookupFn: publicLookup });
 
-    const r = await handler({ url: 'https://example.com/docs' }, signal());
+    const r = await handler({ url: 'https://example.com/docs', max_bytes: 500 }, signal());
     expect(r.isError).toBeUndefined();
-    expect(r.content).toContain('[web_scrape: extraction kept');
     expect(r.content).toContain('mode: "raw"');
+    expect(Buffer.byteLength(r.content, 'utf8')).toBeLessThanOrEqual(500);
+    expect(r.truncated).toBe(true);
+    expect(r.content).toContain('do not re-fetch it in markdown mode');
   });
 
   it('requires no API key for markdown mode', async () => {
--- a/src/agent/tools/handlers/web-scrape.ts
+++ b/src/agent/tools/handlers/web-scrape.ts
@@ -36,6 +36,7 @@ import { checkEgressTarget, guardedFetch, EgressBlockedError } from '../../../we
 import type { EgressGuardOptions as GuardOpts } from '../../../web/egress-guard.js';
 import type { RenderFn } from '../../../web/types.js';
 import { headAndTail } from './_output-cap.js';
+import { withAdvisory } from '../../../web/extraction-advisory.js';
 import {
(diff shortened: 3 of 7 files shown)


griffinwork40:
@codex fix comments

griffinwork40:

|---|------|-------|--------|----------|
| 1 | citation | C-1: `NON_VISIBLE_BLOCK_RE` at line 47 | verified | `extraction-advisory.ts:47` matches diff |
| 2 | citation | C-2: `fetchedHtml = ''` at line ~130 post-merge | verified | `scrape.ts` Phase 1 init confirmed |
| 3 | citation | C-3: Phase 4 return has no advisory | verified | `scrape.ts:267–275` confirmed |
| 4 | citation | S-1: render path at Phase 3 line ~235 | verified | `scrape.ts` Phase 3 return confirmed |
| 5 | file-state | `src/web/index.ts:12` re-exports `ScrapeResult` | verified | git grep confirms at ref |
| 6 | absence | T-1: no Phase 4 advisory test | confirmed-absent | checked `scrape.test.ts` advisory block — 3 tests, none covering Phase 4 |
| 7 | absence | api-compat: only 2 production files import `scrapeToMarkdown`/`ScrapeResult` | confirmed-absent | git grep at ref: `web-scrape.ts` + `web/index.ts` only |

No citations fabricated. No absence claims false-absent. All findings pass through to Wave 2.

Now Wave 2 — synthesis. All findings are `low` or `nit`, all `blocking: false`. No critical/high findings, and no `blocking` overrides departing from defaults. Shadow-verify is not required.

---

## PR #997 Review — `feat(web): warn when markdown extraction retains little of the page`

**Target:** PR #997 `fix/web-scrape-extraction-advisory` → `main` (+343/-1, 6 files)
**Reviewed ref:** `4c828b85b7ee607b7185cd239e287efbd815261c`
**Regime:** full

---

### Security

No issues found — read `src/web/extraction-advisory.ts`, `src/web/scrape.ts`, `src/agent/tools/handlers/web-scrape.ts`. Advisory string interpolates only computed integers (no server-controlled text injection). Regex patterns in `visibleTextLength` are linear-time under V8's Irregexp engine (lazy quantifiers with fixed delimiters, negated character classes). No new network access; existing egress guard unmodified.

### API Compatibility

No issues found. `ScrapeResult.advisory?: string` is purely additive. Only two production importers exist at the ref (`web-scrape.ts` handler, `web/index.ts` re-export), both updated. `withAdvisory(body, undefined)` is a no-op identity for all paths that don't set the field.

### Correctness

1. **low · blocking:false · confidence:high · correctness · `src/web/extraction-advisory.ts:47` · ref:4c828b85 · diff-context**
   `NON_VISIBLE_BLOCK_RE`'s lazy `[\s\S]*?` with backreference `\1` stops at the first matching close tag on nested same-name elements (e.g. `<svg><svg></svg>text</svg>`), leaking inter-nested text into the denominator. Inflates `visibleTextLength` slightly on SVG-heavy pages, biasing toward more advisory firing (false positives, not false negatives). The module comment explicitly documents that over-counting is acceptable and biases in the safe direction.
   ```
   evidence: '<svg><svg></svg>leaked</svg>' → "leaked" counted as visible
   ```
   **suggestion:** Pin the current behavior in a test; no code change needed since the bias direction is safe.

2. **low · blocking:false · confidence:high · correctness · `src/web/scrape.ts:130,161,195` · ref:4c828b85 · diff-context**
   `fetchedHtml` is initialized to `''` and only assigned the body on the HTML path. Phase 2's `extractionAdvisory({ html: '' })` would return `undefined` (below `MIN_SOURCE_TEXT_CHARS`) if reachability assumptions were ever violated. The current control flow is sound — Phase 2 is only reached when `fetchedHtml` holds actual HTML.
   **suggestion:** Optional `// Invariant:` comment at the Phase 2 call site.

3. **low · blocking:false · confidence:high · correctness · `src/web/scrape.ts:267–275` · ref:4c828b85 · diff-context**
   Phase 4 (thin-content fallback after render failure) returns `ScrapeResult` without `advisory`. Correct per design — thin body is itself the signal — but undocumented.
   **suggestion:** One-line comment at Phase 4: `// Thin body is itself the signal; no advisory needed.`

### Spec-Compliance

Intent fully implemented. No unmet requirements. No scope creep. All stated invariants (no new fetch, no new render, no change to thin-content escalation, no change to error classification) are satisfied. The render-path exclusion (Phase 3) is intentional per the PR body and confirmed in code.

### Test Coverage

4. **low · blocking:false · confidence:high · test-coverage · `src/web/scrape.test.ts` advisory block · ref:4c828b85 · diff-context**
   No test covers the Phase 4 fallback path asserting `advisory` is absent on the render-fail thin-content return.
   **suggestion:** Add `it('attaches no advisory when render fails and thin fetched content is returned', ...)`.

5. **low · blocking:false · confidence:high · test-coverage · `src/web/extraction-advisory.test.ts` · ref:4c828b85 · diff-context**
   No test pins the nested-SVG over-count behavior described in finding #1.
   **suggestion:** Add a test documenting the known denominator bias on nested same-name elements.

### Perf/Observability

6. **low · blocking:false · confidence:high · perf-observability · `src/web/extraction-advisory.ts:61–67` · ref:4c828b85 · diff-context**
   `visibleTextLength` runs four sequential regex replacements on every successful markdown scrape, with no trace event or counter recording advisory fire rate. The module comment promises a recalibration recipe, but without observability the threshold cannot be tuned from production data. Per-call cost is negligible (0.2ms on 100KB), but fire-rate is invisible.
   **suggestion:** Emit a trace event or increment a counter when `extractionAdvisory` returns non-undefined.

7. **nit · blocking:false · confidence:high · perf-observability · `src/web/extraction-advisory.ts:47` · ref:4c828b85 · diff-context**
   `NON_VISIBLE_BLOCK_RE` uses the `gi` flags but is applied via `.replace()` with no loop state. The `g` flag is correct (multiple script/style blocks); `i` handles `<SCRIPT>` variants. No issue — documenting for completeness.

---

### What was not checked

- Citations verified inline against branch HEAD `4c828b85b7ee607b7185cd239e287efbd815261c`.
- Stated intent: PR #997 title + body — spec-compliance assessed.
- Did not run tests (read-only review). Test pass status taken from PR body's verification section.
- Did not review Telegram or daemon surfaces for advisory consumption.
- Did not benchmark `visibleTextLength` against production-scale HTML (>1MB) — perf finding is based on the PR body's own 100KB measurement.

---

**Decision: MERGE — 0 blocking (5 low, 1 nit).**

Clean, well-scoped change. The advisory is additive, optional, and correctly gated behind two thresholds. All existing consumers are unaffected. The six non-blocking findings are documentation/observability improvements — none affects correctness or safety of the current implementation.

---

PR #10, opened 2026-06-06
feat(diagnose): inject measured reproducer baseline into the read-only verifier [Step 1]

griffinwork40:
## What this does

The diagnose skill forks a read-only verifier subagent (Bash/Edit/Write denied via `createVerifierCanUseTool`) into an isolated git worktree to assess each hypothesis fix. Previously the verifier's `predicted_pass` verdict was a static code-reading guess with no empirical anchor.

This PR gives the verdict a real anchor by having the **deterministic parent** (the orchestrator TS code) run the reproducer **once** on the current/unfixed code and inject the measured result into each verifier's prompt as ground truth.

**The verifier stays read-only** — `createVerifierCanUseTool` is unchanged.

### Run-at-repo-root rationale

The baseline runs at `parsedInput.repoPath` (the real repo root), NOT in a worktree. Worktrees lack gitignored dependencies (e.g. `node_modules`), so running the test suite there would produce meaningless results. The baseline measures the CURRENT/unfixed code, which lives at the repo root.

### Security posture

- Only the **single detected** shell command (from `detectReproducer`) is run — not a model-chosen command.
- Kill switch: `AFK_DIAGNOSE_BASELINE=0` disables execution entirely for environments where running commands at session time is unacceptable.
- `cwd` is always `parsedInput.repoPath`; never a worktree path.
- `createVerifierCanUseTool()` is **unchanged** — Edit/Write/Bash/Agent/Task remain denied.
- A true execution sandbox (container / network-egress isolation) is tracked as a separate follow-up.

### New exports (unit-tested at the injected-exec seam)

- `runReproducerBaseline(command, cwd, exec?)` — injectable `exec` param so tests fake it without spawning real processes
- `buildVerifierUserPrompt(hypothesis, reproducer, worktreePath, baseline)` — pure function, unit-testable

### verify.md

Added a BASELINE block section instructing the verifier to treat the injected measurement as ground truth, use exit code + output to anchor `predicted_pass` and `confidence`, and set `stance: "blocks"` if exit code is 0 (reproducer passed on unfixed code).

### No schema change

`VerificationResultSchema` is unchanged. The model folds the baseline into the existing `predicted_pass`, `confidence`, `verification_log`, and `signal` fields.

## Depends on / stacks on

- **PR #8** (`fix/diagnose-verifier-honest-labeling`) — honest labeling of the verifier as read-only/prediction-based. This PR base-branches off that work; it should be merged first.

## Step 1 of 2

[...]

PR #1000, opened 2026-08-10
fix(suggest): prevent ghost-text echo, stale cache across model swap

griffinwork40:
## Summary

Fixes three confirmed bugs in the REPL ghost-text suggestion system, each independently shadow-verified against source before implementation.

### Bug 1: Empty-prompt suggestion echoes the user's last input

`isValidPromptSuggestion` had no guard rejecting a suggestion that matches the user's last message. The LLM context includes the raw user message via `getTranscriptTail()` (`user: <msg>\nassistant: <response>`), and cheap suggestion-class models frequently parrot it back.

**Fix:** New `isEchoOfLastInput(suggestion, ctx)` function that extracts the last `user:` line from the transcript tail and compares case-insensitively. Called in `generatePromptSuggestion` after normalization — echoes return `null`.

### Bug 2: Tier-1 history match suggests the just-submitted command

`getDeterministicGhost` iterates history newest-first with no skip. The user's submission is pushed to history immediately (`loop-iteration.ts:462`), so typing a prefix of the last command echoes it back. This is intentional shell convention (zsh/fish), but feels wrong in the REPL context where turns are semantically distinct.

**Fix:** New optional `lastSubmitted` field on `SuggestContext`. The Tier-1 history loop skips any entry that exactly matches `ctx.lastSubmitted`. Wired in `surface-setup.ts` to read `getEntries()[0]` (newest-first). Older history entries sharing a prefix still surface.

### Bug 3: Tier-2 cache stale after `/model` swap

The Tier-2 cache was `Map<string, string | null>` keyed on buffer string alone. After `/model`, stale entries (including cached `null` from a provider that couldn't complete) blocked the new model's lookups.

**Fix:** Composite cache key `${ctx.model}\0${buffer}`. The null-byte separator is safe — neither model IDs nor terminal input contain it. No cache invalidation needed; different models naturally key separately.

## Test Plan

- 13 new tests in `suggest-prompt.test.ts` (echo detection: match/mismatch/empty/multi-turn, generate integration)
- 5 new tests in `suggest.test.ts` (Tier-1 lastSubmitted skip/fallback/undefined, Tier-2 model-keyed cache miss/hit)
- All 111 suggestion tests pass; `tsc --noEmit` clean
- One pre-existing flaky test (`write-denylist.test.ts` symlink race) fails identically on `main`


griffinwork40:
C-1 is a real edge case but the REPL input is typically single-line (readline-based). The failure mode is a false-positive (valid suggestion suppressed), not a false-negative. I'll carry it as `medium · blocking:true` but flag the narrow reachability — the waiver is the author's call.

Now I have all verified findings. Let me produce the synthesis directly — Wave 2 is synthesis-only, and I have all the data I need to do it inline given the light regime.

---

## PR #1000 Review — `fix(suggest): prevent ghost-text echo, stale cache across model swap`

**Branch:** `afk/fix-ghost-text-bugs` → `main`  
**Ref:** `a9fbb810`  
**Stats:** +210 / -4 across 7 files  
**Regime:** light (hotfix, 214 net lines)  
**Stated intent:** Fix three ghost-text bugs — (1) empty-prompt echo, (2) Tier-1 history echo, (3) Tier-2 stale cache after `/model` swap

---

### Findings

**C-2 · medium · blocking:true · confidence:high · correctness · `src/cli/input/suggest-tier2.ts:227` · ref:a9fbb810 · diff-context**  
Cache key uses `ctx.model` (session model) but `pickModel(ctx)` resolves to a different model for all Anthropic-routed sessions (`AFK_COMPACT_MODEL ?? 'haiku'`). Two session models that share the same suggest-model (e.g. `claude-sonnet-4-5` → `claude-opus-4-8`, both routed to `haiku`) produce distinct cache keys that each store identical results, wasting entries — while a swap from a non-Anthropic model to an Anthropic one (or vice versa) correctly busts the cache. The stated bug ("stale cache across model swap") is partially fixed for the session-model axis but structurally wrong for the actual-inference-model axis, which is the one that matters.  
```ts
// Current (wrong): keyed on session model, not inference model
const cacheKey = `${ctx.model}\0${buffer}`;
// pickModel(ctx) may return 'haiku' regardless of ctx.model
const model = deps.pickModel(ctx); // inside run()
```  
**suggestion:** Key on `pickModel(ctx)` (or `deps.pickModel(ctx)`) instead of `ctx.model`. The `pickModel` call already exists at `suggest-tier2.ts:169`; hoisting it to `request()` or passing it as a dep makes the cache key match the actual model.

**C-1 · medium · blocking:true · confidence:medium · correctness · `src/cli/input/suggest-prompt.ts:120-133` · ref:a9fbb810 · diff-context**  
`isEchoOfLastInput` splits transcript on `'\n'` and finds the last `user: ` line. `getTranscriptTail` at `surface-setup.ts:203` embeds `t.user` without newline escaping (`` `user: ${t.user}\n...` ``). A multi-line user message would split into multiple lines, only the first carrying the `user: ` prefix — making the comparison match only line 1 of the input. Failure mode is false-positive suppression (a valid suggestion whose text matches the first line of a pasted multi-line input gets discarded), not false-negative (a true echo is never missed, because `normalizePromptSuggestion` rejects newlines in candidates).  
```ts
// surface-setup.ts:203 — t.user may contain \n
.map((t) => `user: ${t.user}\nassistant: ${t.assistant}`)
// suggest-prompt.ts:128 — splits on \n, finds first "user: " line
const lines = tail.split('\n');
```  
**suggestion:** Escape newlines in `t.user` before embedding (e.g. `.replace(/\n/g, ' ')` in `getTranscriptTail`), or extract the user turn from `stats.turns` directly rather than parsing the serialized string.  
· *waivable:* REPL readline input is overwhelmingly single-line; multi-line paste producing a false-positive suppression has no data-integrity impact. Waive to `blocking:false` if the team confirms multi-line paste is not a high-frequency input pattern.

**C-3 · low · blocking:false · confidence:high · correctness · `src/cli/input/suggest-tier1.ts:50` · ref:a9fbb810 · diff-context**  
`lastSubmitted` skip uses exact equality in a `continue`; if the same command was submitted twice consecutively, history contains two identical entries and only the first is skipped — the second surfaces as a ghost echo. Narrow edge case (repeat-submission); wrong output is a ghost echo, not data loss.  
**suggestion:** Skip `history[0]` unconditionally when it matches `lastSubmitted` (semantic intent is "skip the most-recent entry"), rather than scanning for all matching entries.

**T-1 · low · blocking:false · confidence:high · test-coverage · `src/cli/input/suggest-prompt.test.ts`**  
No test covers `isEchoOfLastInput` with a multi-line `t.user` — the scenario in C-1 is untested. Searched `isEchoOfLastInput.*newline`, `multiline.*user.*input` — zero matches.  
**suggestion:** Add a test with `'user: first line\nsecond line\nassistant: done'` and assert expected behavior.

**T-2 · low · blocking:false · confidence:high · test-coverage · `src/cli/input/suggest.test.ts:852-884`**  
Cache-key tests inject `completeFn`, which bypasses `pickModel` entirely. The test validates key string differentiation between `ctx.model` values but does not exercise the production path where `ctx.model ≠ pickModel(ctx)` — C-2 is untested. Searched `pickModel.*cache|cache.*pickModel` in test files — zero matches.  
**suggestion:** Add a test that mocks `pickModel` to return a fixed value and asserts cache behavior against the actual suggest-model.

**A-1 · low · blocking:false · confidence:medium · api-compat · `src/cli/input/suggest-tier2.ts:147`**  
`run()`'s internal closure signature gains a `cacheKey` parameter. Not externally reachable — `run` is a private closure inside `createTier2Runner`; only `Tier2Runner.request` is the external contract, and its signature is unchanged. [UNVERIFIED: no production importers of the internal closure]

---

### Dropped / Downgraded Findings

None dropped. All citations verified against `a9fbb810`.

---

### What Was Not Checked

- Citations verified inline against branch HEAD `a9fbb810`.
- Stated intent: PR #1000 title + body — spec-compliance assessed.
- Did not run tests (read-only review). PR body reports 111 suggestion tests pass.
- Did not verify the pre-existing flaky test (`write-denylist.test.ts`) claim.
- Did not review Telegram or daemon surfaces (change is REPL-only).
- Runtime behavior of `pickModel` with `AFK_SUGGEST_MODEL` set was not exercised — verified by source read only.

---

### Decision: DO NOT MERGE — 2 medium blocking (1 correctness: cache-key mismatch, 1 correctness: multi-line transcript parse); 1 low, 2 low test-coverage, 1 low api-compat.

**C-2 is the load-bearing blocker.** The cache key uses `ctx.model` but the inference request uses `pickModel(ctx)`, which for all Anthropic sessions returns `haiku` (or `AFK_COMPACT_MODEL`) regardless of session model. The fix is one-line: key on `pickModel(ctx)` instead.

**C-1 is waivable** if multi-line paste is confirmed as a rare/unsupported input pattern — the failure mode is false-positive suppression, not unsafe state. The author's call.

---

griffinwork40:
Good — confirmed: no test covering leading-space push in suggest test files (absence claim #10 confirmed-absent).

Now I have all the verification data. Let me compile the Wave 1.5 results and emit synthesis.

**Wave 1.5 — Citation + Absence-Claim Verification Manifest:**

| # | Type | Claim | Status | Notes |
|---|---|---|---|---|
| F1 | citation (file-state) | `history.ts:130-142` push early returns | **verified** — exact code confirmed at reviewed ref |
| F1 | citation (file-state) | `surface-setup.ts:101-106` unconditional increment | **verified** |
| F2 | citation (file-state) | `suggest-tier2.ts:147-152` cacheKey param | **verified** |
| F2 | citation (file-state) | `suggest-tier2.ts:231-244` cacheKey computed at debounce | **verified** |
| F2 | inference | "ctx.model changes between debounce and fire → completion uses new model" | **fabricated** — `ctx` is captured once at `request()` time; `run()` at line 169 uses the same `ctx` object, so `deps.pickModel(ctx)` returns the same value as the cacheKey computation. No cross-model contamination possible. |
| F3 | citation (file-state) | `suggest-prompt.ts:133` exact equality | **verified** |
| F6 | citation (file-state) | `surface-setup.ts:101` typeof guard | **verified** |
| F10 | absence | No test for leading-space push in suggest tests | **confirmed-absent** — zero matches in suggest test files |

**Result:** Finding 2 is **DROPPED** — the central claim (cacheKey stale relative to completion model) is fabricated; both use the same captured `ctx`. All other citations verified.

---

## PR #1000 Review — fix(suggest): prevent ghost-text echo, stale cache across model swap

**Branch:** `afk/fix-ghost-text-bugs` → `main`  
**Reviewed ref:** `db7bda33`  
**Change type:** hotfix · 7 files · +290/-4  
**Regime:** light

---

### Findings

**1. `medium · blocking:true · confidence:high · correctness · src/cli/commands/interactive/surface-setup.ts:101-107 · ref:db7bda33 · citation-type:file-state`**

The monkey-patch increments `sessionSubmissionCount` unconditionally after `originalPush(text)`, but `ReplHistory.push()` has four early-return paths that do NOT modify `_entries`: leading-space (`text.startsWith(' ')`), empty/whitespace, secret-pattern match, and consecutive-duplicate. After any of these, `sessionSubmissionCount` becomes ≥1 while `entries[0]` is unchanged — the `lastSubmitted` getter returns the wrong entry, and the Tier-1 skip fires against a command the user did not just submit.

```ts
// surface-setup.ts:103-106 — always increments
historyRing.push = (text: string) => {
  originalPush(text);
  sessionSubmissionCount++;  // fires even when push() discards
};
// history.ts:131 — one of four early returns
if (text.startsWith(' ')) return;  // _entries unchanged
```

**Suggestion:** After `originalPush(text)`, compare `ring.getEntries()?.[0]` to a pre-call snapshot; only increment when the head actually changed. Alternatively, cache `lastSubmittedEntry` directly in the monkey-patch after confirming the push landed.

---

**2. `low · blocking:false · confidence:high · correctness · src/cli/input/suggest-prompt.ts:270 · ref:db7bda33 · citation-type:diff-context`**

The comment "this is the load-bearing check" overstates the echo guard's scope — it catches exact case-insensitive matches only. A model reply that appends even a single word beyond the user's last message bypasses it. Acceptable tradeoff; documentation-only concern.

```ts
if (isEchoOfLastInput(candidate, ctx)) return null;
// "load-bearing" → only catches exact echoes
```

**Suggestion:** Revise comment to "this catches exact parrot echoes; partial echoes or paraphrases are not filtered."

---

**3. `low · blocking:false · confidence:high · test-coverage · src/cli/input/suggest.test.ts:807-848 · ref:db7bda33 · citation-type:diff-context`**

No test locks in the all-instances-skipped semantics when `lastSubmitted` matches multiple entries in history (the `continue` skips every occurrence, not just the first). Behavior is probably desirable but untested.

```ts
if (ctx.lastSubmitted !== undefined && entry === ctx.lastSubmitted) continue;
// skips ALL matching entries, not just history[0]
```

**Suggestion:** Add a test: `history = ['foo', 'bar', 'foo']`, `lastSubmitted = 'foo'`, buffer `'fo'` → assert `null` (both `foo` entries skipped).

---

**4. `low · blocking:false · confidence:high · test-coverage · ref:db7bda33`**

No test exercises the monkey-patch wiring with discarded-push inputs (leading-space, consecutive-duplicate). The existing `lastSubmitted skip` tests inject `lastSubmitted` directly — they never exercise the `surface-setup.ts` monkey-patch path where Finding 1's bug lives.

**Suggestion:** Integration test: construct a `ReplHistory`, apply the patch logic, call `push(' hidden')`, verify `sessionSubmissionCount` stays 0 and `lastSubmitted` returns `undefined`.

---

**5. `medium · blocking:false · confidence:high · spec-compliance · src/cli/commands/interactive/surface-setup.ts:101 · ref:db7bda33 · citation-type:file-state · waived: the guard is correct for non-TTY surfaces where ghost text is disabled; a doc comment resolves the gap.**

The `if (typeof historyRing.push === 'function')` guard silently skips the monkey-patch when `push` is absent (test stubs, non-TTY surfaces), leaving `sessionSubmissionCount` permanently 0 and `lastSubmitted` always `undefined`. This is correct behavior (ghost text is disabled on those surfaces) but undocumented — a silent degradation rather than an explicit invariant.

**Suggestion:** Add a comment: "// When push is absent, lastSubmitted always returns undefined (safe: ghost text is disabled on non-TTY surfaces)."

---

### Dimensions with no issues

- **Security** — no issues found. Read `suggest-prompt.ts`, `suggest-tier2.ts`, `suggest-tier1.ts`, `surface-setup.ts`. No credential exposure, no injection surface, no auth bypass.
- **API compatibility** — no issues found. `lastSubmitted` is optional on `SuggestContext`; `run()` signature change is module-internal. No breaking changes.
- **Perf-observability** — no regression. Two `getEntries()` allocations per keystroke is pre-existing behavior, not introduced here.

### Dropped findings

| # | Reason |
|---|---|
| Original F2 (medium/blocking — Tier-2 cacheKey stale after model swap) | **Fabricated inference.** `ctx` is captured once at `request()` time; both the cacheKey computation and `run()`'s `deps.pickModel(ctx)` use the same object. No cross-model contamination is possible within a single debounce cycle. |
| Original F6 (medium/blocking — silent degradation = spec gap) | Retained as Finding 5 but **downgraded to medium non-blocking** (waived) — the guard is correct behavior; a doc comment resolves it. |

### What was not checked

- Citations verified inline against branch HEAD `db7bda33`.
- Stated intent: PR #1000 title+body — spec-compliance assessed.
- Did not verify `loop-iteration.ts:462` push call-site integration (referenced in PR body but not in the diff).
- Original Finding 2 (cacheKey race) was independently re-derived and found to be fabricated — the `ctx` capture semantics disprove the claim.

---

**Decision: DO NOT MERGE — 1 medium blocking (correctness); 1 medium waived, 2 low, 0 nit.**

The single blocking finding (F1: `sessionSubmissionCount` incremented on discarded pushes → wrong `lastSubmitted`) produces wrong Tier-1 skip behavior when a user's first submission in a session is a leading-space command, a secret-pattern match, an empty string, or a consecutive duplicate. The fix is small: compare `entries[0]` before/after `originalPush()` and only increment when the head changed.

---

griffinwork40:
`history: IHistoryRing` — not optional (no `?`). So `surface.history` is always defined per the type. The `?? undefined` does guard against `null` specifically (which `IHistoryRing` wouldn't be, but the runtime cast might yield). The nit is valid.

| Type | Claim | Status | Finding |
|------|-------|--------|---------|
| citation | `IHistoryRing` at types.ts:23-37 | verified | security-api F1 |
| citation | `ReplHistory.push` at history.ts:130 | verified | security-api F1 |
| citation | suggest.test.ts:820-841 reimplementation | verified | correctness-tests F1 |
| absence | `installHistorySubmissionTracker` not tested | confirmed-absent | correctness-tests F1 |
| citation | `suggest-prompt.ts:128` split('\n') | verified | security-api F2 |

All findings pass verification. No fabricated citations, no false-absent claims. Now Wave 2 synthesis.

---

## PR #1000 Review: fix(suggest): prevent ghost-text echo, stale cache across model swap

**Reviewed ref:** `a2ec34c6` · **Branch:** `afk/fix-ghost-text-bugs` → `main` · **+541 / -73 / 9 files** · **Regime:** full

---

### Findings

**1. medium · blocking:true · confidence:high · dimension:api-compat · `src/cli/input/types.ts` + `surface-setup.history-tracking.ts:41-48` · ref:a2ec34c6 · citation-type:file-state**

**finding:** The monkey-patch in `installHistorySubmissionTracker` targets `surface.history.push`, which exists on the concrete `ReplHistory` class but is absent from the `IHistoryRing` interface that types `InputSurface.history`. The patch reaches it via an unsafe structural cast `(surface.history ?? undefined) as { push?: … }`. A future `IHistoryRing` implementor that omits `push` would silently leave `lastSubmitted` as `undefined` for the session, disabling the Tier-1 echo skip with no error.

**evidence:**
```ts
// types.ts:24-37 — IHistoryRing has no push()
export interface IHistoryRing {
  back(draft: string): string | null;
  forward(): string | null;
  resetRecall(): void;
  readonly inRecall: boolean;
  getEntries?(): readonly string[];
}
```

**suggestion:** Add `push?(text: string): void` to `IHistoryRing` (optional, to preserve existing test doubles), or add a `// Contract:` comment on the cast documenting the `ReplHistory` assumption.

---

**2. medium · blocking:true · confidence:high · dimension:test-coverage · `src/cli/input/suggest.test.ts:820-841` · ref:a2ec34c6 · citation-type:file-state**

**finding:** The monkey-patch tests in `suggest.test.ts` exercise a local reimplementation (`applyLastSubmittedPatch`) that duplicates the logic of `installHistorySubmissionTracker`, rather than importing and calling the actual exported function. A regression in the exported function's property access, binding, or guard logic would be invisible to these tests.

**evidence:**
```ts
// suggest.test.ts — defines its own inline patch
function applyLastSubmittedPatch(history: ReplHistory): { … } {
  let lastSubmittedEntry: string | undefined;
  // ... re-implements the patch manually
}
```
`installHistorySubmissionTracker` has zero test-file references (grep confirmed at ref a2ec34c6).

**suggestion:** Either import `installHistorySubmissionTracker` and test it against a `ReplHistory`-shaped stub, or add a dedicated unit test for the exported function. The existing inline tests validate the algorithm's correctness against `ReplHistory`'s discard paths — the gap is only on the exported function's own wiring.

---

**3. nit · blocking:false · confidence:medium · dimension:api-compat · `src/cli/input/suggest-prompt.ts:121-138` · ref:a2ec34c6 · citation-type:diff-context**

**finding:** `isEchoOfLastInput` splits the transcript tail on `\n` and extracts only the first physical line after `user: `. A multi-line user message (`"user: line1\nline2\n…"`) yields only `"line1"` as the comparison target — the echo guard fires a false negative (echo not detected). Not a regression (no guard existed before), but the JSDoc states the format as a fixed invariant without noting the multi-line limitation.

**suggestion:** Add a `// Contract:` note to the JSDoc: "Multi-line user inputs produce a false negative — only the first line is compared. Exhaustive echo detection is out of scope for this guard."

---

**4. nit · blocking:false · confidence:medium · dimension:spec-compliance · `src/cli/commands/interactive/surface-setup.history-tracking.ts:47` · ref:a2ec34c6 · citation-type:diff-context**

**finding:** `(surface.history ?? undefined)` — the `?? undefined` is a no-op when the property is already typed as `IHistoryRing` (not optional, not nullable).

**evidence:**
```ts
const historyRing = (surface.history ?? undefined) as { … } | undefined;
```

**suggestion:** Simplify to `(surface.history as { … } | undefined)`.

---

**5. low · blocking:false · confidence:low · dimension:perf-observability · `src/cli/input/suggest-tier1.ts:46` · ref:a2ec34c6 · citation-type:file-state**

**finding:** `ctx.getHistory()` allocates a new array on every keystroke, even when the dropdown candidate (path a) already matched and the history loop is never entered. Pre-existing behavior, not introduced by this PR.

**suggestion:** Move `getHistory()` call after the dropdown early-return. Low priority — allocation is negligible.

---

### What was not checked

- Citations verified inline against branch HEAD `a2ec34c6`.
- Stated intent: PR #1000 title + body — spec-compliance assessed. All three bugs are fixed as described. The sibling extraction preserves the public surface.
- Did not run tests (`pnpm test`); test assertions reviewed from diff only.
- Did not review Telegram or daemon surfaces — change is scoped to REPL ghost-text.
- Runtime performance not benchmarked — keystroke-path cost assessed by code inspection only.

---

### Decision: DO NOT MERGE — 2 medium blocking (1 api-compat, 1 test-coverage); 2 nit, 1 low.

The two medium findings are both addressable with small additive changes (one interface addition or comment, one test import swap). The core logic of all three bug fixes is correct and well-tested at the algorithmic level.

---

griffinwork40:
## PR #1000 Re-review — `335118c4`

**Branch:** `afk/fix-ghost-text-bugs` → `main`
**Reviewed ref:** `335118c4` (includes fix commit addressing prior review feedback)
**Stats:** +541 / -73 across 11 files
**Regime:** light

---

### Prior review feedback — status

All prior blocking findings have been addressed in commit `335118c4`:

| Prior finding | Status |
|---|---|
| F1 (medium/blocking): `IHistoryRing` missing `push?` | **fixed** — `push?(text: string): void` added to interface at `types.ts:39-42` |
| F2 (medium/blocking): Tests reimplemented patch instead of importing real function | **fixed** — `applyLastSubmittedPatch` now imports and calls `installHistorySubmissionTracker` |
| F3 (nit): Multi-line echo guard limitation undocumented | **fixed** — Contract comment added at `suggest-prompt.ts:125-128` |
| F4 (nit): No-op `?? undefined` | **fixed** — simplified to direct cast at `surface-setup.history-tracking.ts:41` |

### New findings

**1. low · blocking:false · confidence:medium · spec-compliance · `suggest-types.ts:46` · ref:335118c4**

The transcript format (newest-first) is documented only at the consumer (`isEchoOfLastInput` JSDoc) but not at the interface definition (`SuggestContext.getTranscriptTail`). A future change to the producer could silently break the echo guard.

**suggestion:** Add the format spec to the `getTranscriptTail()` JSDoc in `suggest-types.ts`.

**2. nit · blocking:false · confidence:medium · test-coverage · `suggest.test.ts:963-1003` · ref:335118c4**

Tier-2 cache-key tests exercise static cache semantics (two independent contexts) but not the dynamic `/model` swap path (same engine, model changes between calls). Existing tests are correct and meaningful; this is additive coverage.

**suggestion:** Add one test with same engine and buffer, model A → model B.

### Dimensions with no issues

- **Security** — no issues found. Monkey-patch closure is properly scoped; no new egress paths.
- **Correctness** — all three bugs fixed correctly. Echo guard forward-scans newest-first transcript. `lastSubmitted` skip uses exact-match. Cache key uses `pickModel(ctx)` (inference model, not session model).
- **API compatibility** — `push?` added to `IHistoryRing` as optional; `lastSubmitted` optional on `SuggestContext`. No breaking changes.
- **Perf/observability** — `getLastSubmitted()` is O(1). No new hot-path allocations.

### What was not checked

- Citations verified inline against branch HEAD `335118c4`.
- Stated intent: PR #1000 title+body — spec-compliance assessed. All three bugs fixed as described.
- Did not run tests (read-only review). PR reports all suggestion tests pass.
- Runtime behavior not benchmarked.

---

### Decision: MERGE — 0 blocking (1 low, 1 nit). All prior blocking findings resolved.

---
_🤖 Posted by `agent-afk /review --light --post github`_

Issue #919, opened 2026-08-05
refactor(subagent): forkSubagent is still a ~567-LOC function — #832 table shows it closed via #829, which only shrank the file


griffinwork40:
Follow-up to #832 (file-size triage) and #829.

`SubagentManager.forkSubagent` spans `src/agent/subagent.ts:269-835` — **~567 LOC in a single function**. #832's headline finding is that whole-file LOC is a false-positive-heavy metric and that **largest single function** is the signal that actually predicts maintenance pain. By that criterion this is a live Tier-1-class offender, and it is currently untracked.

## Why it looks tracked but isn't

#832's function table maps `forkSubagent` (540 LOC) → #829. #829 is **closed**, but its title is precise about what it did: *"extract constants + fork types from subagent.ts (~360 LOC, zero risk)."* That extraction shrank the **file**; it never touched the **function**. So the table entry is discharged in name only — the row is checked off, the 540-LOC function it pointed at is still there, and has since grown to ~567 LOC (#910 added the budget-preamble injection and its Invariant comment).

Of the six functions in #832's table, this is the one whose issue closed without the function changing. Worth correcting the table so the tracking issue does not under-report.

Current state: `src/agent/subagent.ts` is 861 LOC total, of which `forkSubagent` is ~567 — the function is 66% of its own file.

## What is actually entangled in there

`forkSubagent` is the single choke point every child `AgentSession` passes through — agent-tool foreground/background, compose/DAG nodes, skill forks, and in-process callers such as the mint phases all converge on it. That is a genuinely load-bearing role, and it has accreted a matching number of concerns. Visible in the current body:

- phaseRole validation and mutual-exclusion checks
- cross-provider credential gating and `coerceCrossProviderChildModel` model coercion
- cwd / worktree main-root resolution (`resolveMainRootForCwd`, cache)
- read-scope and permission bubbling
- witness-trace seal ownership (fork must never seal the shared trace)
- occupancy touch for the worktree sweep's live-session guard
- the anti-hang round cap default and the OAuth-429 fail-fast sibling
- budget-preamble injection (#910)
- handle construction, abort-graph wiring, teardown registration

Several of those are already self-contained enough to lift into `src/agent/subagent/` next to the existing `constants.ts` / `budget-preamble.ts` — the directory and the precedent both exist.

## Caveats worth respecting before anyone starts

[...]

Issue #890, opened 2026-08-03
refactor(telegram): streaming.ts's streamResponse is a 768-line function — low churn, safe mechanical extraction available


griffinwork40:
## Problem

`src/telegram/streaming.ts` (1175 LOC) — `streamResponse` (:407-1175) is a single ~770-line function. It dispatches on `event.type` via a chain of `if`s (not a `switch`), and the `paused`/countdown branch nests 4-5 levels deep. It's the single production call site for all Telegram bot traffic (invoked from `telegram/handlers/message.ts`).

Independently corroborated by two separate analyses on 2026-08-03: a 4-lens `/simplify --all` sweep, and #832's file-size triage, which flags this file under "not yet filed individually — file if picked up" and adds a data point neither analysis alone had: only **11 commits in 6 months** (low churn for a file this central), and "pure render helpers (:214-389) extract mechanically today" — i.e. this is likely a *lower-risk* extraction than the raw line count suggests.

## Proposed fix

Extract each `event.type` arm (progress / paused / resumed / done / error) into its own named handler function taking a shared mutable-state bag, mirroring the pattern already used in `src/agent/session/stream-consumer.ts`. Start with the pure render helpers at :214-389, which #832 already flagged as mechanical.

## Risk / verification

No dedicated characterization tests are called out for this file today — confirm/add coverage for the `event.type` arms before extracting, then gate behind `pnpm lint && pnpm test` (this repo's lint is `tsc --noEmit`, no ESLint).

## References

- #832 — file-size triage tracking issue, notes this file under "not yet filed individually"

Issue #830, opened 2026-08-01
chore(env): generate env getters from ENV_REGISTRY instead of splitting env.ts — and document the audit-env-access path constraint


griffinwork40:
## Summary

`src/config/env.ts` is the largest file in the repo (**1699 LOC**) with the highest churn (**52 commits/6mo**). It will show up first in any file-size audit. **It should not be split**, and this issue exists to record why — and to document a CI constraint that would bite anyone who tries.

## Why a split is the wrong fix

| Range | Content | LOC | % |
|---|---|---|---|
| :1-97 | header + type defs (`EnvVarType`, `EnvVarCategory`, `EnvVarMeta`) | 97 | 5.7% |
| :98-1402 | `ENV_REGISTRY` — 140 flat object-literal entries | 1305 | **76.8%** |
| :1415-1611 | `env` object — 140 one-line getters, mechanically 1:1 with the registry | 197 | 11.6% |
| :1612-1699 | bespoke logic (`isPlainOutputRequested`, secret-hardening IIFE, parity types) | ~100 | 5.9% |

**88.4% is flat data plus mechanically-derived getters; ~6% is bespoke logic.** There is zero control flow in :98-1402. The 52 commits reflect *additions to a table* via the documented 2-step recipe at :38-44 — not fixes to interacting logic.

A mechanical split fails its own goal: an `env-registry.ts` holding just the array would still be ~1400 LOC, 4× over the ceiling.

## The actual fix, if we want one: codegen

All 140 getters are 100% derivable from `ENV_REGISTRY`. Replacing hand-written getters with a ~15-line `Object.defineProperty` loop would:

- delete ~197 LOC of hand-maintained getters
- retire the hand-sync compile-time parity hack at :1656-1663 and its test
- make the "add an env var" recipe a 1-step change instead of 2

Optionally, `ENV_REGISTRY` itself could move to a JSON data file consumed by both the getter-builder and `scripts/render-env-registry.ts`, reducing `env.ts` to ~50 LOC of loader logic.

**Difficulty: moderate.** Dynamic `Object.defineProperty` is a real behavior change vs. static getters — it touches secret-hardening (:1631) and the compile-time parity guarantee, which depends on statically-declared getter names. A generated object needs a mapped type over `ENV_REGISTRY` to preserve that guarantee. Not free.

## ⚠️ CI constraint — read before moving anything

`scripts/audit-env-access.ts:45` hardcodes the canonical path in `ALLOWED_FILES`:

[...]

PR #1002, opened 2026-08-10
feat(ci): add a function-scoped 200-line ceiling, answering #831

### LINKED ISSUE #831: ci: add a function-scoped file-size gate (a file-scoped one would fire 146x on day one)
## Problem

There is no automated file-size gate in this repo. A naive one would be actively harmful.

Current baseline, measured across `*.ts`/`*.tsx`/`*.js`/`*.mjs`/`*.cjs` excluding `*.test.*`, `*.spec.*`, `tests/`, `__tests__/`, `*.d.ts`:

- **756** non-test code files
- **177,571** LOC
- **146 files (19.3%)** over a 350-LOC ceiling

A file-scoped gate set at 350 fires **146 times on day one** and gets disabled within a week.

## Why file LOC is the wrong signal here

Roughly 40% of the largest files are large from *breadth*, not entanglement, and splitting them buys nothing:

- `src/config/env.ts` (1699) — 88% flat `ENV_REGISTRY` data + mechanically-derived getters
- `src/agent/trace/types.ts` (886) — **zero** function/class declarations; a persisted NDJSON wire schema
- `src/agent/tools/schemas.ts` (1308) — 26 flat tool-schema literals; exactly **one** non-declarative statement in the whole file (a `.map()` at :1291)
- `src/cli/terminal-compositor.ts` (1088) — thin delegators over 9 already-split siblings, plus the heaviest test coverage in the repo
- `src/cli/terminal-compositor.input-dispatch.ts` (1144) — a clean chain-of-responsibility: `dispatchKey` (:194) fanning out to 18 self-contained handlers
- `src/cli/commands/interactive/shared.ts` (835) — four interfaces defining one coherent REPL contract

## The signal that does predict pain: largest function

| Function | LOC | File |
|---|---|---|
| `runTurn` | ~1065 | `providers/anthropic-direct/loop.ts` |
| `bootstrapSession` | 887 | `cli/commands/interactive/bootstrap.ts` |
| `commitAbove` | ~730 | `cli/terminal-compositor.committed-band-commit.ts` |
| `forkSubagent` | 540 | `agent/subagent.ts` |
| `query()` | 425 | `providers/anthropic-direct/index.ts` |
| `createSession` | 357 | `src/telegram.ts` |
| `handleEnter` | 258 | `cli/terminal-compositor.input-dispatch.ts` |
| `[Symbol.asyncIterator]` | 258 | `providers/anthropic-direct/query.ts` |
| `handlePhoto` | 234 | `telegram/handlers/message.ts` |
| `executeBatch` | 227 | `agent/tools/dispatcher.ts` |

Six functions individually exceed the whole-file ceiling. Every one of them is a genuine maintenance problem; none of the six false-positive files is.

## Proposal

Add `scripts/check-file-size.ts` wired into CI alongside the existing `audit:*:check` gates, with:

1. **A function-scoped ceiling** as the primary rule — flag any function over N LOC (suggest 350 to match, or 200 to be stricter, ratcheting down over time).
2. **A file-scoped ceiling** as a secondary, looser rule (suggest 800-1000) to catch genuine sprawl without firing on registries.
3. **An explicit allowlist with rationale**, following the `ALLOWED_FILES` pattern already used in `scripts/audit-env-access.ts:43-45` — pre-populated with the six false positives above, each with a one-line reason.
4. **Ratchet mode**: fail only on *new or worsening* violations, so existing debt does not block unrelated PRs.

## Sequencing

Land this **after** the Tier 0/Tier 1 refactors, not before. Introducing the gate while 146 violations are outstanding guarantees it gets bypassed. Once the worst offenders are split, the allowlist is small enough to be honest.

## Prior art

`app/Scripts/check-file-size.sh` in the Umber repo implements the file-scoped version. This one should be function-scoped — that is the lesson from this repo's baseline.

griffinwork40:
Adds a **function-scoped** size gate alongside the file gate that landed in #999, and reconciles both with the #831/#832 triage that Phase 0 shipped without citing.

Closes #831.

## Why this is not redundant with #999

The two gates measure different things, and neither implies the other:

| | measures | ceiling | baseline |
|---|---|---|---|
| `check-file-size.ts` (#999) | how much must be **read** to establish edit safety | 350 raw lines | 138 files (15.6%) |
| `check-function-size.ts` (this PR) | how much must be **held in mind** to change one behaviour | 200 lines | 54 functions (1.2%) |

They diverge in both directions. A flat 900-line registry is a large file containing no large function. A 700-line function can sit inside a file that passes 350 only because siblings were extracted around it.

The sharpest case is **#919**: #829 shrank `subagent.ts` and closed, while `forkSubagent` itself never changed — and it has since grown to **586 lines**. A file-scoped gate scores that as progress. This one does not.

## Where 200 came from

Measured, not chosen by taste. Across 4,388 functions in `src/` + `scripts/` (median 10, p99 229, max 856):

| limit | over | % of all | excess lines |
|---|---|---|---|
| 350 | 15 | 0.34% | 3,394 |
| 250 | 31 | 0.71% | 5,459 |
| **200** | **54** | **1.23%** | **7,595** |
| 150 | 100 | 2.28% | 11,284 |

#831 proposed "350 to match, or 200 to be stricter". 200 sits at ~p99 and yields a 54-entry baseline — one screen. That is #831's own argument made concrete: a gate whose baseline you can read is a gate that gets fixed; one that fires on a fifth of the tree gets disabled. 350 was rejected because it misses the entire 200–350 band (39 functions, including `buildChildConfig` at 346 and `handleOrchestratorEvent` at 304).

## Validation against independently-filed issues

The measurer was cross-checked against numbers three separate issues arrived at by hand, without seeing this code:

| function | issue says | measured here |
|---|---|---|
| `streamResponse` | 768 (#890) | **769** |
| `commitAbove` | ~730 (#827) | **735** |
| `forkSubagent` | ~567 (#919) | **586** (grown since) |

## Metric detail: JSDoc is excluded

Deliberately diverging from the file metric, which counts comments on purpose. The difference is whether a pressure valve exists. At file scope you relieve the ceiling by extracting a concern, and JSDoc travels with its declaration, so nothing is lost. At function scope there is no such valve — a function cannot be split from its own doc comment — so counting JSDoc would create undiluted pressure to delete documentation, which the long-comment convention forbids.

Two other measurement decisions, both documented as contracts in `scripts/lib/function-extents.ts`:

- **Outermost only.** A nested callback is attributed to the declaration enclosing it, so one problem raises one violation and every violation names something you can actually extract. Class members count as outermost (a class is not a function), which is what makes `commitAbove` and `executeBatch` visible.
- **Baseline keys never embed line numbers.** A key containing a line number would rotate on every edit above it, so each unrelated insertion would read as one STALE plus one NEW.

## Shared ratchet, not a second copy

The five-mode ratchet (NEW / GREW / RETIRED / STALE / TOUCHED) is extracted to `scripts/lib/size-ratchet.ts` and **both** gates now use it. `check-file-size.ts` is migrated onto it rather than left holding a hand-synced duplicate — the hazard #822 was filed for, and what `audit-module-state.ts` exists to catch.

The migration is behaviour-preserving, verified rather than asserted:

- `pnpm audit:filesize:check` and `--list` output diffed before/after: **identical**, except `check-file-size.ts` itself shrinking 326 → 210 (so it leaves the warn band) and the new files appearing.
- `.filesize-baseline.json` regenerates **byte-identical**.
- The file gate still exits 1 on a fresh 360-line file and 0 when clean.

The ratchet also gains the test coverage it shipped without: 23 tests over both the AST measurement and all five failure modes, including that `permanent` exempts from RETIRED but **not** from GREW.

## Verification

- All five ratchet modes unit-tested; **NEW / GREW / TOUCHED additionally driven end-to-end with real mutations** against the live tree (not just asserted):
  - NEW — a 253-line function in a new file → exit 1
  - GREW — 3 lines into `createAfkModeGate` → `grew 253 → 256 (+3)`, exit 1
  - TOUCHED — same edit vs `origin/main` → correctly flagged
- `pnpm lint`, `audit:filesize:check`, `audit:funcsize:check`, `audit:module-state:check`, `audit:env:check`, `audit:chalk:check`, `audit:sdk:check`, `scan:env:check`, `fix:pins:check`, `pnpm build` — all pass.
- `pnpm test`: 16,054 passed / 1 failed. The failure is `write-denylist.test.ts` (macOS `/tmp` symlink), **pre-existing and unrelated** — it reproduces identically on a pristine worktree at `origin/main` containing none of these changes, and this PR touches no file under `src/`.

Touch-trigger ships advisory (`|| true`) to match the file gate's posture; flip both to blocking together once the god-object waves land.

## Reconciliation with #831 / #832 (docs)

Phase 0 shipped without referencing the triage that motivated it — `grep -cE '#83[0-2]'` returned 0 across the PR body, `docs/file-size-ceiling.md`, and the plan. `docs/file-size-ceiling.md` now records:

- Of #831's four asks, Phase 0 delivered the allowlist and the ratchet and inverted the metric. What discharges the "fires 146× on day one" objection is the ratchet — **which was #831's own item 4** — not something #831 failed to consider.
- What Phase 0 did *not* discharge is the metric argument. This PR does.
- #832's Tier 4 "explicitly leave alone" files are all grandfathered rather than scheduled — in particular `src/config/env.ts`, which is governed by open **#830** (generate getters from `ENV_REGISTRY`; do not split — `scripts/audit-env-access.ts:45` hardcodes that path, exact-matched at `:175`).
- The waves are to be sequenced **entanglement-first**; selecting a wave for "maximum mass" is the objective #831 refutes.


griffinwork40:
@codex fix comments

griffinwork40:
The `-merge -diff` text does NOT appear in the diff — it's pre-existing. S-1 is a pre-existing issue that this PR does not introduce. I'll keep it as a nit rather than medium-blocking since the PR doesn't touch that line.

Now let me synthesize.

---

## PR #1002 Review — `feat(ci): add a function-scoped 200-line ceiling, answering #831`

**Branch:** `afk/function-size-gate` → `main`  
**Reviewed ref:** `c09bfcf6`  

---

### Findings

**1.** medium · blocking:true · confidence:high · correctness · `scripts/check-function-size.ts:103–105` · ref:c09bfcf6 · citation-type:file-state  
**`isScannable()` uses `path.sep` to split git-returned paths, breaking the TOUCHED trigger on Windows.** `changedSince()` returns POSIX `/`-separated paths (git's output), but `isScannable` splits on `path.sep` (which is `\` on Windows). On Windows, `'src/foo.ts'.split('\\')` yields `['src/foo.ts']`, so `segments[0] === 'src'` fails — silently skipping every file. Additionally, `touchedKeysSince()` compares `changed.has(parsed.file)` where `parsed.file` comes from POSIX-keyed `sizes` map entries vs POSIX git output, but the `sizes` map keys are built from `path.relative()` (OS-native separators), so on Windows the sets never intersect.  
```ts
const segments = relPath.split(path.sep);
// ...
return SCAN_ROOTS.some((root) => segments[0] === root);
```
CI runs on ubuntu-latest, so this does not affect CI. The `functionKey()` helper already normalizes `\\` to `/`, but `isScannable` does not.  
**suggestion:** Normalize with `relPath.replaceAll('\\', '/').split('/')` before the segment checks, consistent with `functionKey()`.

**2.** medium · blocking:false · confidence:high · correctness · `scripts/check-function-size.ts:162,232` · ref:c09bfcf6 · citation-type:file-state · waived: bounded, non-data-affecting — race window between two reads of the same file in a single-process synchronous invocation is negligible; worst case is a cosmetic mismatch in the "grandfathered" count printed to stdout  
**`reportAndExit()` loads the baseline a second time independently from `main()`, creating a theoretically inconsistent state.** `main()` loads the baseline at line 232 and passes it to `collectViolations`, but `reportAndExit()` reloads it from disk at line 162 for the warning band and the "grandfathered" count. If the file changed between reads, violations and warnings would be computed against different baselines.  
```ts
// line 232 (main): const baseline = loadBaseline(RATCHET);
// line 162 (reportAndExit): const baseline = loadBaseline(RATCHET);
```
**suggestion:** Thread the `Baseline` from `main()` into `reportAndExit` as a parameter.

**3.** low · blocking:false · confidence:high · test-coverage · `tests/check-function-size.test.ts` · ref:c09bfcf6 · citation-type:file-state  
**No test covers a class property arrow function (`class Foo { bar = () => { ... } }`).** The `isPropertyDeclaration` branch in `resolveName()` handles this pattern, but no fixture exercises it. Grep search for `bar =.*=>` or `PropertyDeclaration` in the test file returned zero matches.  
**suggestion:** Add a fixture: `class Foo { bar = () => { return 1; }; }` and assert the name is `'Foo.bar'`.

**4.** low · blocking:false · confidence:high · test-coverage · `tests/check-function-size.test.ts` · ref:c09bfcf6 · citation-type:file-state  
**No test exercises `touchedKeysSince()`, the function that maps changed files onto function keys.** The existing TOUCHED ratchet test directly supplies `touchedKeys` as a Set, bypassing the file→key mapping. A regression in `parseFunctionKey` integration with `changedSince` would be undetected. Grep for `touchedKeysSince` in the test file: zero matches.  
**suggestion:** Add a test that calls `touchedKeysSince` with a known `sizes` map and verifies correct key inclusion/exclusion.

**5.** low · blocking:false · confidence:medium · correctness · `scripts/lib/function-extents.ts:187` · ref:c09bfcf6 · citation-type:file-state  
**Bodiless function nodes set `insideFunction=true` for their children.** When a function-like node has no body (overload signature), the visitor still recurses with `insideFunction=true`. This is harmless in practice — `.d.ts` files are excluded by the scan filter — but is a latent inconsistency in the visitor's semantics.  
```ts
ts.forEachChild(node, (child) => visit(child, true)); // fires for bodiless nodes too
```
**suggestion:** `visit(child, insideFunction || hasBody(node))` instead of unconditional `true`.

**6.** nit · blocking:false · confidence:high · correctness · `scripts/lib/function-extents.ts:125,140` · ref:c09bfcf6 · citation-type:file-state  
**Asymmetric naming for anonymous default exports.** `export default function() {}` gets `<default export#N>` (ordinal), while `export default () => {}` gets `<default export>` (no ordinal). `dedupeNames` would handle collisions anyway.  
**suggestion:** Unify both to `'<default export>'` and let `dedupeNames` handle the suffix.

**7.** nit · blocking:false · confidence:high · api-compat · `scripts/check-file-size.ts:28–34` · ref:c09bfcf6 · citation-type:file-state  
**Header comment says "fails four ways" but the shared ratchet has five modes.** The TOUCHED mode is wired via `--changed-vs` but not listed in the header's mode enumeration. Pre-existing omission from the migration.  
**suggestion:** Update the comment to "fails five ways" and add TOUCHED to the list.

**8.** nit · blocking:false · confidence:high · api-compat · `scripts/check-function-size.ts:97` vs `scripts/check-file-size.ts:85` · ref:c09bfcf6 · citation-type:file-state  
**Function gate scans `.tsx` files; file gate does not.** `check-function-size.ts` includes `.tsx` in `INCLUDED_EXTENSIONS` while `check-file-size.ts` omits it. No `.tsx` files exist today, so the divergence is latent.  
**suggestion:** Either add `.tsx` to the file gate too, or document why the scopes differ.

**9.** nit · blocking:false · confidence:high · spec-compliance · `AFK.md:160`, `CLAUDE.md:147` · ref:c09bfcf6 · citation-type:file-state  
**Docs claim `.filesize-baseline.json` has `-merge -diff` in `.gitattributes`; actual attribute is `-merge` only.** `.gitattributes` has an explicit "Deliberately NOT `-diff`" comment explaining why. Pre-existing in both AFK.md and CLAUDE.md — the PR adds new text referencing the correct attribute (CLAUDE.md:180 says `-merge`) but does not fix the existing claim.  
**suggestion:** Correct both AFK.md:160 and CLAUDE.md:147 from `-merge -diff` to `-merge`.

---

### Dimensions with no issues

- **Security:** no issues found — read `scripts/lib/size-ratchet.ts`, `scripts/check-function-size.ts`, `scripts/check-file-size.ts`. No injection surfaces, no credentials, no network calls.
- **Perf-observability:** no actionable issues. `scanAll()` parses every source file synchronously (acceptable for CI); no performance regression vs. the existing file gate.

### Spec-compliance summary

The stated intent is well-discharged. The function ceiling, shared ratchet extraction, file-gate migration, baseline, CI integration, touch-trigger, and docs updates are all present. The `.tsx` scope difference (finding 8) is a minor unmentioned divergence. All five ratchet modes work as documented.

---

### What was not checked

- Citations verified inline against branch HEAD `c09bfcf6`.
- Stated intent: PR #1002 title+body — spec-compliance assessed.
- Did not run tests or the gate itself; relied on the PR body's verification section.
- Did not review the website/ docs site or Telegram surface.
- Runtime behavior of the TypeScript compiler API (`ts.createSourceFile`) is taken as correct; edge cases in the TS parser itself were not tested.

---

### Decision: **DO NOT MERGE** — 1 medium blocking (correctness); 1 medium waived; 4 low, 4 nit.

Finding 1 (Windows `path.sep` mismatch in `isScannable`) is the sole blocker. The TOUCHED trigger is silently a no-op on Windows because git paths and filesystem-walked paths never compare equal. While CI runs on Linux and the touch-trigger is advisory (`|| true`), the bug means `pnpm audit:funcsize:check --changed-vs origin/main` produces false-clean results on Windows developer machines — the exact "fires but is wrong" failure the gate's contract forbids. The fix is a one-line path normalization.

---

griffinwork40:
**Wave 1.5 — Citation + Absence-Claim Verification (inline):**

| Finding | Check | Cited location | Status | Note |
|---|---|---|---|---|
| 1 | Citation A | `function-extents.ts:187` | **verified** | Line 187 confirmed: `ts.forEachChild(node, (child) => visit(child, insideFunction \|\| hasBody(node)));` |
| 2 | Citation A | `size-ratchet.ts:168–195` | **verified** | Line 168: `if (loc > entry.loc)` (GREW), line 195: `kind: 'TOUCHED'`, line 193 guard only suppresses NEW |
| 3 | Citation A | `check-function-size.ts:100–106` | **verified** | Lines 100–107 confirmed: `isScannable` with `SCAN_ROOTS.some` guard |
| 4 | Absence B | "no test covers GREW+TOUCHED" | **confirmed-absent** | grep `GREW.*TOUCHED` returned 0 matches; existing TOUCHED test uses `loc: 260 = baseline.loc: 260` (no GREW) |

**Finding 1 reachability correction:** The finding claims a bodiless overload with an arrow default parameter is "narrow but reachable". This is **incorrect** — TypeScript rejects default parameter initializers on overload signatures (`TS2371: A parameter initializer is only allowed in a function or constructor implementation`). Since `tsc --noEmit` (the lint gate) runs **before** this gate in CI, the edge case is unreachable on code that passes the preceding pipeline step. Severity corrected from medium → **low** (the pre-downgrade classification was wrong on reachability, not a confidence downgrade).

**Finding 2 inherited behavior check:** The same GREW+TOUCHED double-report exists in the **pre-PR** `check-file-size.ts` (diff lines 491–504 show identical guard logic: `if (!baseline.entries[file]) continue` — only NEW, not GREW). The shared ratchet faithfully preserves this behavior; it was not introduced by this PR.

Now synthesizing.

---

## Review: PR #1002 — `feat(ci): add a function-scoped 200-line ceiling, answering #831`

**Branch:** `afk/function-size-gate` → `main` · **Ref:** `5e86247b` · **12 files, +1235/−160**

### Summary

Adds a function-scoped 200-line ceiling (`scripts/check-function-size.ts`) alongside the existing file-scoped 350-line ceiling, extracts the shared ratchet logic to `scripts/lib/size-ratchet.ts`, migrates `check-file-size.ts` onto it, and wires both CI steps. Well-structured change with clear motivation traced to #831.

### Findings

**1.** `medium` · `blocking: false` · confidence: high · correctness · `scripts/lib/size-ratchet.ts:168–195` · ref: `5e86247b` · citation-type: file-state
**GREW + TOUCHED double-report for the same key.** When `--changed-vs` is active, a baselined function that grew emits both a GREW violation and a TOUCHED violation. The TOUCHED guard (`if (!baseline.entries[key]) continue`) suppresses only NEW-flagged keys, not GREW-flagged ones.
```typescript
// line 193 — only NEW is suppressed:
if (!baseline.entries[key]) continue; // already reported as NEW
```
**suggestion:** Add `if (loc > entry.loc) continue;` in the TOUCHED loop to suppress TOUCHED when GREW already fired for the same key.
· waived: bounded UX noise in CI output — both violations are factually accurate, inherited from the pre-PR file gate (same logic at diff line 492), and not data-affecting; the author can land and follow up.

**2.** `low` · `blocking: false` · confidence: low · correctness · `scripts/lib/function-extents.ts:187` · ref: `5e86247b` · citation-type: file-state
**Bodiless overload children receive `insideFunction=false`.** `visit` passes `insideFunction || hasBody(node)` to children of a bodiless overload, so an arrow function inside the parameter list of a bodiless overload would be measured as outermost. However, TypeScript rejects default parameter initializers on overload signatures (`TS2371`), and `tsc --noEmit` runs before this gate in CI, making the edge case unreachable. [low confidence — verify with runtime context]
```typescript
// line 187:
ts.forEachChild(node, (child) => visit(child, insideFunction || hasBody(node)));
```
**suggestion:** Defensive: pass `true` for bodiless function-like nodes (`insideFunction || true`) to close the path unconditionally. Not required since the scenario is unreachable.

**3.** `low` · `blocking: false` · confidence: high · correctness · `scripts/check-function-size.ts:107` · ref: `5e86247b` · citation-type: file-state
**`isScannable` divergence between the two gates.** The function gate's `isScannable` adds `SCAN_ROOTS.some((root) => segments[0] === root)` that the file gate's version lacks. This is load-bearing for `changedSince` filtering (git diff can return any repo path), but latent for the file gate which reaches `SCAN_ROOTS` only through `walk()`.
**suggestion:** Add the same root-check to the file gate's `isScannable`, or document the intentional asymmetry.

**4.** `nit` · `blocking: false` · confidence: high · test-coverage · `tests/check-function-size.test.ts` · ref: `5e86247b` · citation-type: file-state
**No test covers the GREW + TOUCHED interaction.** The existing TOUCHED test uses `loc: 260` matching `baseline.loc: 260` to avoid triggering GREW. A test with `sizes=261, baseline.loc=260, touchedKeys=[key]` would document the intended behavior (one violation or two).

### What was not checked

- Citations verified inline against branch HEAD `5e86247b`.
- Stated intent: PR #1002 title+body — spec-compliance assessed.
- Security and api-compat: no issues found — CI-only tooling, no auth/secret/network surface, `execFileSync` uses array form (no shell injection).
- Did not run the test suite or the gate scripts live.
- Did not review the `website/` surface.
- Perf: AST parsing of the full `src/` tree runs at CI time, not runtime — acceptable.

### Decision

**Decision: MERGE — 0 blocking (1 medium waived: UX noise, not data-affecting, inherited from pre-PR file gate; 2 low; 1 nit).**

---

**Done**

- PR #1002 reviewed in full-regime (2 parallel Wave 1 agents + inline Wave 1.5 citation verification + synthesis).
- 4 findings surfaced: 1 medium (waived — GREW+TOUCHED double-report inherited from prior gate), 1 low (unreachable AST edge case), 1 low (isScannable divergence), 1 nit (missing interaction test). Zero blocking.
- Merge recommendation: **MERGE**.
- This review is read-only — no files were modified, no commits made, no PR comments posted.
- Evidence: this transcript and the diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/c492f04e-553e-460d-8a40-e442d83b18bb/pr-1002.diff`.

---

PR #1005, opened 2026-08-11
perf(attachment-registry): compare digests instead of re-reading files on dedup (#885)

### LINKED ISSUE #885: perf(attachment-registry): dedup reads entire file from disk on every cache hit
## Finding

From review of PR #846 (`afk/subagent-image-registry`), Wave 1 perf-observability dimension.

`InboundAttachmentRegistry.put()` computes the sha256 digest of incoming bytes but, on a prefix collision with an existing entry at the same id, reads the **entire file** back from disk to compare bytes:

```ts
// src/agent/content/attachment-registry.ts:109
if (existing.mediaType === mediaType && (await readFile(existing.path)).equals(bytes)) {
  return { id, ...existing };
}
```

This is O(n) per cache hit (n = file size, up to 5 MiB). For the common case where the same image is resubmitted within a session, the digest is already in hand — the disk read is unnecessary work.

## Suggested fix

Store the full sha256 digest in `InboundAttachmentRecord` and compare digests instead of reading the file:

```ts
export interface InboundAttachmentRecord {
  readonly path: string;
  readonly mediaType: InboundMediaType;
  readonly sizeBytes: number;
  readonly digest: string;  // sha256 hex — avoids re-reading the file on dedup
}
```

```ts
if (existing.mediaType === mediaType && existing.digest === digest) {
  return { id, ...existing };
}
```

Since `digest` is already computed at the top of `put()`, this turns the dedup check into a pure in-memory string comparison — no disk I/O on the cache-hit path.

## Context

- **Severity**: medium (perf), no correctness impact
- **Scope**: `src/agent/content/attachment-registry.ts` only
- **Reviewed ref**: `421989628109213cb15b775ba155896860d7dfcd`
- Refs: #843, #846

griffinwork40:
## Summary

Fixes #885 — `InboundAttachmentRegistry.put()` was re-reading the entire attachment file from disk on every prefix-collision check in order to compare bytes. For large images this meant an unnecessary `readFile` on every dedup hit.

### `src/agent/content/attachment-registry.ts`
- Added `digest: string` field to `InboundAttachmentRecord` interface (sha256 hex, computed once in `put()`)
- Stored `digest` in the record object at write time
- Replaced `(await readFile(existing.path)).equals(bytes)` dedup check with `existing.digest === digest` — an O(64) string compare instead of an O(N) disk read
- Removed `readFile` from the `node:fs/promises` import (no longer used)

### `src/agent/content/attachment-registry.test.ts`
- Updated the `get()` record assertion to include the `digest` field
- Added a new test verifying that `digest` is a 64-char hex string and that identical bytes deduplicate by digest match
- Rewrote the suffix test: the old test drove the suffix via byte-comparison (now removed); the new test pre-populates all prefix slots with distinct digest values and verifies the `_1` suffix fires when every prefix length is occupied by a different digest

## Behaviour

No functional change — sha256 collision probability is negligible. The record now carries the digest it was keyed on, eliminating the async disk read on every dedup path.


griffinwork40:
Splitting into individual PRs per issue — this PR bundled 4 unrelated fixes onto one branch.

PR #1013, opened 2026-08-11
fix(providers): address remaining #762 review items (M1, M4, M6, L1, listener leak)

### LINKED ISSUE #762: overload (529): mid-stream retry exhaustion aborts the whole session to 0 turns while the usage-limit path pauses — pause machinery is structurally unreachable

griffinwork40:
Closes #762

This PR applies the five unaddressed review findings from #764's second review round. All were verified against `origin/main` by direct file read before this change; none were previously tracked anywhere else.

## What was wrong

After #764 landed the mid-stream overload pause machinery, five issues remained:

| ID | Severity | File(s) |
|---|---|---|
| M1 | Correctness | `overload-pause-tier.ts`, `query-turn-driver.ts` |
| M4 | Layering | `closure-reason.ts`, `closure-emitter.ts`, `subagent/result.ts` |
| M6 | Correctness | `loop/round-request.ts`, `loop.ts` |
| L1 | Docs | `overload-pause.ts` |
| Leak | Hygiene | `shared/sleep-with-abort.ts` |

## M1 — `close()` during overload pause seals turn `succeeded`

`overload-pause-tier.ts:131` bare-returned on `isClosed()` after the probe sleep, discarding the held `OVERLOAD_EXHAUSTED` terminal. The docblock said "every exit path re-yields the preserved terminal" — the `close()` path violated that invariant.

**Fix (two-part):**
- **Tier**: distinguish `close()` from `interrupt()`. `isClosed()` now yields the exhausted terminal before returning; `signal.aborted` (recoverable interrupt) still exits silently (query-turn-driver synthesizes an `interrupted` terminal for that path — unchanged).
- **Driver** (`query-turn-driver.ts`): process `turn.completed` — recording `lastUsage.stopReason` — *before* `if (ctx.state.closed) return`, so `lastStopReason` is captured even when `close()` lands during the pause.

## M4 — provider-neutral modules imported from `anthropic-direct/`

`session/closure-reason.ts`, `session/closure-emitter.ts`, and `agent/subagent/result.ts` imported `OVERLOAD_EXHAUSTED` from `anthropic-direct/overload-pause.js` — a provider-specific path.

**Fix:** extract `OVERLOAD_EXHAUSTED` to `providers/shared/overload-sentinel.ts`. `overload-pause.ts` re-exports it so internal import paths are unchanged. The three provider-neutral consumers now import from the shared module.

## M6 — connection-phase 529 exhaustion was still fatal

`loop/round-request.ts:createWithRetry` threw a generic `Error` after the retry budget was spent. `openRound`'s catch block converted it to a raw `error` event → `sawProviderError` → `closure{reason:'abort'}` with no turn committed. Identical to the mid-stream gap #762 describes.

**Fix:**
- `createWithRetry` throws a module-local `ConnectionOverloadExhaustedError` sentinel when a transient 529/503 budget is exhausted.
- `openRound`'s catch block detects the sentinel and returns a new `overload-exhausted` outcome.
- `loop.ts` handles `opened.kind === 'overload-exhausted'` identically to the mid-stream case: emits `OVERLOAD_EXHAUSTED_NOTICE` + a clean `turn.completed` stamped with `OVERLOAD_EXHAUSTED`, routing it to the pause machinery.

## L1 — stale JSDoc on `OVERLOAD_EXHAUSTED_NOTICE`

The docblock claimed the notice was "never pushed into history". `loop.ts:204-210` had already corrected this. `overload-pause.ts` is updated to match.

## Listener leak — `sleep-with-abort.ts`

`signal.addEventListener('abort', handler, { once: true })` removes the listener on fire but not when the timer resolves first. Added `signal.removeEventListener('abort', handler)` in the timer callback.

## Tests

- `loop.retry.test.ts`: updated `exhausts retries and yields error on persistent 529` to assert the M6 behavior (clean `OVERLOAD_EXHAUSTED` terminal, no raw error event).
- `overload-pause-tier.test.ts`: updated `stops when the session closes during the pause` to assert the M1 fix (tier yields the terminal on `close()`, not silence).
- All 1047 tests across 90 files pass. TypeScript compiles clean. No new filesize violations.

### DIFF
--- a/src/agent/providers/anthropic-direct/loop.retry.test.ts
+++ b/src/agent/providers/anthropic-direct/loop.retry.test.ts
@@ -148,7 +148,12 @@ describe('runTurn transient error retry', () => {
     expect(completed).toBeDefined();
   });
 
-  it('exhausts retries and yields error on persistent 529', async () => {
+  // M6 fix: connection-phase 529 budget exhaustion now routes to the CLEAN
+  // overload terminal (OVERLOAD_EXHAUSTED) instead of a fatal `error` event.
+  // This makes the connection-phase behave identically to the mid-stream phase:
+  // the pause machinery in `overload-pause-tier.ts` can park-and-probe, and even
+  // on a fail-fast surface the turn commits so `afk --resume` stays valid.
+  it('exhausts connection-phase retries and yields OVERLOAD_EXHAUSTED terminal on persistent 529', async () => {
     const client: AnthropicClientLike = {
       messages: {
         create: vi.fn(() => {
@@ -183,10 +188,16 @@ describe('runTurn transient error retry', () => {
 
     // 1 initial + OVERLOAD_MAX_RETRIES retries
     expect(client.messages.create).toHaveBeenCalledTimes(OVERLOAD_MAX_RETRIES + 1);
+    // M6: no raw error event — exhausted connection-phase 529 emits the CLEAN
+    // OVERLOAD_EXHAUSTED sentinel so the pause machinery can catch it.
     const errorEvent = events.find((e) => e.type === 'error');
-    expect(errorEvent).toBeDefined();
-    if (errorEvent?.type === 'error') {
-      expect(errorEvent.error.message).toContain('Overloaded');
+    expect(errorEvent).toBeUndefined();
+    const assistantNotice = events.find((e) => e.type === 'assistant.message');
+    expect(assistantNotice).toBeDefined();
+    const completed = events.find((e) => e.type === 'turn.completed');
+    expect(completed).toBeDefined();
+    if (completed?.type === 'turn.completed') {
+      expect(completed.usage.stopReason).toBe(OVERLOAD_EXHAUSTED);
     }
   });
 
--- a/src/agent/providers/anthropic-direct/loop.ts
+++ b/src/agent/providers/anthropic-direct/loop.ts
@@ -175,6 +175,36 @@ export async function* runTurn(
       continue;
     }
 
+    // M6: connection-phase 529/503 budget exhausted — treat identically to
+    // mid-stream exhaustion so the pause machinery in `overload-pause-tier.ts`
+    // can park-and-probe. Fall through to the `overload-exhausted` branch
+    // below rather than the generic fatal path (pre-fix the error was thrown
+    // from createWithRetry, landed in openRound's catch, and yielded a raw
+    // `error` event → sawProviderError → closure {reason:'abort'} with no
+    // turn committed).
+    if (opened.kind === 'overload-exhausted') {
+      void emitSessionPhase(input.traceWriter, {
+        phase: 'rate_limit',
+        metadata: {
+          reason: 'overloaded',
+          source: 'connection-phase',
+          attempt: OVERLOAD_MAX_RETRIES,
+          exhausted: true,
+        },
+      });
+      yield {
+        type: 'assistant.message',
+        text: OVERLOAD_EXHAUSTED_NOTICE,
+        sessionId: input.ctx.sessionId,
+      };
+      yield {
+        type: 'turn.completed',
+        usage: turn.withDuration({ ...turn.usage, stopReason: OVERLOAD_EXHAUSTED }),
+        sessionId: input.ctx.sessionId,
+      };
+      return;
+    }
+
     const { events, ttfb, stall, requestStartedAt } = opened;
 
     const outcome = yield* consumeRoundStream({
--- a/src/agent/providers/anthropic-direct/loop/round-request.ts
+++ b/src/agent/providers/anthropic-direct/loop/round-request.ts
@@ -63,6 +63,20 @@ export function toWireTool(tool: AnthropicToolDef): WireToolDef {
   };
 }
 
+/**
+ * Sentinel thrown by `createWithRetry` (only) when the connection-phase 529/503
+ * budget is exhausted. Distinct from the generic Error so `openRound`'s catch
+ * block can route exhausted transient errors to the `overload-exhausted` outcome
+ * instead of the fatal error path (M6 — the same gap that #762 fixed for the
+ * mid-stream phase).
+ */
+class ConnectionOverloadExhaustedError extends Error {
+  constructor() {
+    super('Connection-phase overload budget exhausted');
+    this.name = 'ConnectionOverloadExhaustedError';
+  }
+}
+
 // `requestSignal` is passed to `messages.create` — it is the caller's turn
 // signal chained with the per-request TTFB stall timer (see armFirstByteTimeout),
 // so aborting it covers BOTH a user interrupt and a first-byte timeout. The
@@ -90,8 +104,11 @@ async function createWithRetry(
     } catch (err) {
       if (requestSignal.aborted) throw err;
       const e = err instanceof Error ? err : new Error(String(err));
-      if (isTransientServerError(e) && attempt < OVERLOAD_MAX_RETRIES) {
-        continue;
+      if (isTransientServerError(e)) {
+        if (attempt < OVERLOAD_MAX_RETRIES) continue;
+        // Budget exhausted: signal the caller with a typed sentinel so it can
+        // route to the CLEAN overload terminal instead of the fatal error path.
+        throw new ConnectionOverloadExhaustedError();
       }
       throw e;
     }
@@ -115,11 +132,16 @@ export interface OpenRoundTransport {
  * - `opened` — a live stream; the caller owns watchdog disposal from here.
  * - `retry-ttfb` — first-byte stall before any content; watchdogs already
  *   disposed, caller should re-drive the round (budget permitting).
+ * - `overload-exhausted` — connection-phase 529/503 budget exhausted; the
+ *   caller should route to the CLEAN overload terminal rather than the fatal
+ *   error path (M6: same structural gap that existed for mid-stream exhaustion
+ *   before #762; both should reach the pause machinery).
  * - `terminated` — a terminal event was already yielded; the turn is over.
  */
 export type OpenRoundResult =
   | OpenRoundTransport
   | { kind: 'retry-ttfb'; requestStartedAt: number }
+  | { kind: 'overload-exhausted' }
   | { kind: 'terminated' };
 
 /** Everything the connection phase needs from the enclosing turn. */
@@ -265,6 +287,14 @@ export async function* openRound({
         sessionId: input.ctx.sessionId,
       };
       return { kind: 'terminated' };
+    }
+    // M6: connection-phase 529/503 budget exhausted. Route to the CLEAN overload
+    // terminal (same path as mid-stream exhaustion) so the pause machinery in
+    // `overload-pause-tier.ts` can park-and-probe rather than hard-aborting the
+    // session. The sentinel type lets us distinguish this from every other error
+    // without modifying `isTransientServerError` or the outer type union.
+    if (err instanceof ConnectionOverloadExhaustedError) {
+      return { kind: 'overload-exhausted' };
     }
     const e = annotateFastError(err, input.fastMode === true);
     if (e.message.includes('thinking')) {
--- a/src/agent/providers/anthropic-direct/overload-pause.ts
+++ b/src/agent/providers/anthropic-direct/overload-pause.ts
@@ -30,22 +30,22 @@
 import { env } from '../../../config/env.js';
 import type { ProviderEvent } from '../../provider.js';
 
-/**
- * Terminal `stopReason` stamped on the `turn.completed` that ends a turn whose
- * mid-stream overload retry budget was exhausted.
- *
- * Consumed by three places, all of which must stay in agreement:
- *   - `query/retry-layer.ts` — {@link classifyOverloadExhaustion} keys the pause
- *     arm on it.
- *   - `session/closure-reason.ts` — maps it to the `abort` closure reason.
- *   - `session/closure-emitter.ts` — maps it to a `failed` seal status.
- */
-export const OVERLOAD_EXHAUSTED = 'overload_exhausted';
+// Imported for local use (classifyOverloadExhaustion) and re-exported so
+// existing `from './overload-pause.js'` import sites keep working (M4: the
+// three provider-neutral consumers have been updated to import from the shared
+// module directly, but anthropic-direct-internal consumers like loop.ts still
+// import from here).
+import { OVERLOAD_EXHAUSTED as _OVERLOAD_EXHAUSTED } from '../shared/overload-sentinel.js';
+export { OVERLOAD_EXHAUSTED } from '../shared/overload-sentinel.js';
 
 /**
- * Operator-facing copy for an exhausted overload. Emitted as a display-only
- * `assistant.message` (never pushed into history — it is operator context, not
- * model context), mirroring the `stop_reason: 'refusal'` notice in `loop.ts`.
+ * Operator-facing copy for an exhausted overload. Emitted as an
+ * `assistant.message`, mirroring the `stop_reason: 'refusal'` notice in
+ * `loop.ts`. Unlike the refusal notice, this IS visible to the model on
+ * `--resume`: `session/stream-consumer.ts` materializes non-empty
+ * `assistant.message` events into `conversationHistory`, which threads back
+ * as model context on the next turn. See `loop.ts:204-210` for the
+ * corrected contract.
  *
  * Exists because the raw SSE envelope (`{"type":"error","error":{"type":
  * "overloaded_error"}}`) reached operators verbatim and was misread as a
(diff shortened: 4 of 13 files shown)


griffinwork40:

**Check A — Citation verification:**

1. **`overload-pause.ts:41-48` (stale cross-ref nit)**: VERIFIED — `loop.ts:204-210` does indeed contain `};` / `return;` / `const { events ...` — not the contract comment. The contract comment lives at lines ~260-265. Citation is accurate.

2. **`query-turn-driver.ts:186-193` (continue nit)**: VERIFIED — lines 186-198 match exactly: `if (event.type === 'turn.completed')` → `yield event` → `continue` → `if (ctx.state.closed) return` → `yield event`. The double-yield concern is real: without `continue`, execution would fall through.

3. **`round-request.ts:63-77` (file-private class)**: VERIFIED — `ConnectionOverloadExhaustedError` is at lines 67-77 (class body), `class` keyword without `export`. File is 306 lines.

4. **`result.test.ts:23` (stale import)**: VERIFIED — line 23 reads `import { OVERLOAD_EXHAUSTED } from '../providers/anthropic-direct/overload-pause.js';` — still uses the provider-specific path.

5. **`sleep-with-abort.ts:23-25`**: VERIFIED — both mechanisms present as described.

**Check B — Absence-claim verification:** No absence claims were made by either agent. No check needed.

**Verification manifest:** All citations verified. Zero fabricated, zero diff-only. All findings pass through to synthesis.

Now let me synthesize the Wave 2 verdict.

---

## PR #1013 Review: `fix(providers): address remaining #762 review items`

**Target:** PR #1013 · `afk/fix-762-overload` → `main`
**Ref:** `c70c1c349a885d306db2d55b61a4de2db8a071a1`
**Change type:** hotfix · 13 files · +458/−40
**Stated intent:** Addresses five unresolved findings (M1, M4, M6, L1, listener leak) from #764's second review round.

---

### Findings

**1.** `nit` · blocking:false · confidence:high · **spec-compliance** · `src/agent/subagent/result.test.ts:23` · ref:c70c1c34 · file-state
**Finding:** M4 intent migrated production consumers to the shared sentinel path, but the test file `result.test.ts` still imports `OVERLOAD_EXHAUSTED` from the provider-specific `anthropic-direct/overload-pause.js` path.
**Evidence:** `result.test.ts:23` — `import { OVERLOAD_EXHAUSTED } from '../providers/anthropic-direct/overload-pause.js';`  vs. production `result.ts:16` — `import { OVERLOAD_EXHAUSTED } from '../providers/shared/overload-sentinel.js';`
**Suggestion:** Update the test import to `'../providers/shared/overload-sentinel.js'` for consistency with the M4 migration.

**2.** `nit` · blocking:false · confidence:high · **correctness** · `src/agent/providers/anthropic-direct/overload-pause.ts:46` · ref:c70c1c34 · file-state
**Finding:** JSDoc cross-reference `loop.ts:204-210` points to a closing brace and variable destructure, not the contract comment it claims to reference; the actual contract comment (`Contract: the notice is NOT appended…`) lives at approximately `loop.ts:260-265`.
**Evidence:** `loop.ts:204` = `};` / `return;`; the referenced contract is at lines 260-265 (`Contract: the notice is NOT appended to THIS turn's input.messages…`).
**Suggestion:** Update the cross-reference to the correct line range, or replace the pinned line number with a searchable anchor comment like `// Contract: OVERLOAD_EXHAUSTED_NOTICE visibility`.

**3.** `nit` · blocking:false · confidence:high · **correctness** · `src/agent/providers/anthropic-direct/query-turn-driver.ts:192` · ref:c70c1c34 · diff-context
**Finding:** The `continue` after `yield event` in the `turn.completed` branch is load-bearing (prevents double-yield of the terminal event), but the comment block explains only the closed-return motivation — not the double-yield prevention role.
**Suggestion:** Append one sentence: `// The `continue` is required — without it, execution falls through to the second `yield event` at line 198.`

**4.** `low` · blocking:false · confidence:high · **api-compat** · `src/agent/providers/anthropic-direct/loop/round-request.ts:72-77` · ref:c70c1c34 · file-state
**Finding:** `ConnectionOverloadExhaustedError` is a file-private (unexported) class with an `instanceof` check in the same file; if a future file-size split moves one of the two sites, the `instanceof` silently fails to `false` and routes exhaustion back to the fatal error path.
**Evidence:** `class ConnectionOverloadExhaustedError` (line 72, no `export`); `err instanceof ConnectionOverloadExhaustedError` (line 296). File is 306 lines (below the 316-line warn band, but growing).
**Suggestion:** Add an invariant comment documenting the co-location requirement, or export the class for resilience against splits.

**5.** `low` · blocking:false · confidence:medium · **correctness** · `src/agent/providers/shared/sleep-with-abort.ts:23-25` · ref:c70c1c34 · file-state
**Finding:** The timer callback's `removeEventListener` and the `{ once: true }` option on `addEventListener` are a redundant belt-and-suspenders pattern — the `removeEventListener` in the timer callback is doing real work (removing an unfired listener on the natural-timeout path), while `{ once: true }` handles the abort-fires-first path.
**Evidence:** Line 23: `setTimeout(() => { signal.removeEventListener('abort', onAbort); resolve(); }, ms)`; Line 25: `signal.addEventListener('abort', onAbort, { once: true })`.
**Suggestion:** Add a brief comment documenting why both mechanisms are present: `// removeEventListener handles timer-fires-first; { once } handles abort-fires-first`.

---

### Dimension summaries

| Dimension | Verdict |
|---|---|
| **Security** | no issues found — read overload-sentinel.ts, sleep-with-abort.ts, round-request.ts. No credential exposure, auth bypass, or secret-handling changes. |
| **API compat** | 1 low (file-private sentinel class co-location risk). No breaking changes to external or internal APIs; the `OpenRoundResult` union is extended additively. |
| **Correctness** | no issues found — the M1 event-ordering fix, M6 connection-phase sentinel, and listener leak fix are all structurally sound. 2 nits on comment hygiene. |
| **Spec-compliance** | All five stated intent items (M1, M4, M6, L1, Leak) have implementing changes. 1 nit on test import consistency. No unmet intent, no scope creep. |
| **Test coverage** | no issues found — M6 test updated (`loop.retry.test.ts`), M1 tier test updated (`overload-pause-tier.test.ts`), new 286-line `query-turn-driver.test.ts` covers concurrent/pre-set close, lastUsage, and normal path. |
| **Perf / observability** | no issues found — M6 connection-phase path emits symmetric `rate_limit` trace annotation with `source: 'connection-phase'`. |

---

### What was not checked

- Citations verified inline against branch HEAD `c70c1c34`.
- Stated intent: PR #1013 title+body — spec-compliance assessed.
- Did not review Telegram or daemon surfaces for overload-exhaustion behavior.
- Did not verify that `overload-pause.ts`'s re-export (`export { OVERLOAD_EXHAUSTED } from '../shared/overload-sentinel.js'`) is structurally equivalent at the TypeScript type level (verified by string identity only — both export `'overload_exhausted'`).

---

### Decision: **MERGE** — 0 blocking (2 low, 3 nit).

All five review items are addressed with correct implementations and test coverage. No critical, high, or medium findings. The two low findings and three nits are non-blocking documentation and resilience improvements.

---

PR #1014, opened 2026-08-12
fix(#962): fail-fast context-overflow guard for max_tokens exceeding window

### LINKED ISSUE #962: max_tokens is never clamped against remaining context: haiku 400s once input exceeds ~136k, and auto-compaction triggers too late to prevent it
## Problem

`max_tokens` is resolved from `(config, model)` alone and sent to the wire verbatim. Nothing ever checks it against the **remaining** context window, so on a long conversation AFK can emit a request the provider rejects outright — `input_tokens + max_tokens > context_window`.

`resolveMaxTokens(config, model)` (`src/agent/providers/anthropic-direct/resolve-params.ts:67-85`) takes no messages and no usage, so it *structurally cannot* see how much context is already consumed. `buildRoundParams` (`loop/round-request.ts`) passes `input.maxTokens` straight through. The openai-compatible twin `resolveEffectiveMaxOutputTokens` (`query/model-params.ts`) has the same shape. Grep for `contextLimitFor` near either resolution site returns only status-line percentage math.

This is the *third* member of the `max_tokens` cluster (#951 thinking-budget, #953 over-ceiling clamp) — same root shape: a cap resolved without the constraint that actually governs it.

## Where it bites

Overflow point per model (window − output ceiling), from `src/agent/model-limits.ts`:

| Model | Window | Output ceiling | 400s once input exceeds |
|---|---|---|---|
| `haiku` | 200,000 | 64,000 | **136,000** |
| `claude-sonnet-4-6` (raw wire id) | 200,000 | 64,000 | **136,000** |
| `sonnet` / `opus` (base) | 1,000,000 | 128,000 | 872,000 |
| openai-compatible, unknown model | 262,144 | 64,000 | **198,144** |

The 1M-window Claude models have enormous headroom and are effectively safe. **`haiku` is the live exposure** — it is the cheap subagent tier, and a long investigation subagent crossing ~136k input starts hard-400ing every turn. The openai-compatible fallback window (`DEFAULT_CONTEXT_LIMIT_OPENAI_COMPATIBLE = 262_144`) is the second.

## Auto-compaction does not save us — it triggers too late

Two independent reasons:

1. **It is off by default.** `autoCompact?: boolean | { threshold: number }` is optional (`types/config-types.ts:738`) and `resolveAutoCompactThreshold` returns `undefined` when it is `undefined` or `false` (`providers/shared/auto-compact.ts:184`). Out of the box nothing compacts at all.
2. **Even when enabled, the default threshold is past the overflow point.** `DEFAULT_AUTO_COMPACT_THRESHOLD = 0.9` (`auto-compact.ts:168`):

   | Model | trigger @0.9 | + output ceiling | vs window |
   |---|---|---|---|
   | `haiku` | 180,000 | 244,000 | **over by 44,000** |
   | `sonnet_1m` / `opus_1m` | 900,000 | 1,028,000 | **over by 28,000** |
   | openai-compat unknown | 235,929 | 299,929 | **over by 37,785** |

   Compaction fires *after* the request that would already have been rejected.

3. **It never runs mid-turn.** `maybeAutoCompact(ctx)` is called only from the outer prompt loop in `providers/anthropic-direct/query-turn-driver.ts:236`, after a turn completes — and the helper's own doc explicitly says *"Do not call this from inside the inner for-await"* (`:258`). A single long turn that burns context across many tool rounds sails past every trigger.

## Why this isn't a one-liner

There is **no tokenizer or `countTokens` call anywhere in the repo** (grep-clean). The only estimators are unrelated heuristics: `ceil(len/3.5)` for HOT.md sizing (`memory-store.ts`), `round(delta/4)` for the compaction-savings display, `ceil(chars/4)` for a REPL typing indicator.

`computeContextUsage(state)` (`query-capabilities.ts`) reads `state.lastUsage` — the last *completed* round's API-reported usage. It is synchronous, but **stale by one round**: it excludes the new user message and any tool results appended since. Usable as a lower bound, not as a guarantee.

Two implementation tiers, both with a real downside:

- **Heuristic clamp** — estimate `messages + system + tools` by character count in `round-request.ts` and clamp `max_tokens` to `contextLimitFor(model) − estimate − margin`. One file, <40 lines. But a *wrong* estimate silently shrinks `max_tokens`, which manufactures the truncation #952 just went to the trouble of surfacing. Trading a loud 400 for a quiet truncation is not obviously a win.
- **Thread the real count** — carry `state.lastUsage.input_tokens` from `session-state.ts` → `query-turn-driver.ts` (`TurnDriverContext.maxTokens` is frozen at turn start) → `types.ts`/`request-types.ts` → `round-request.ts`. ~5 files, 3 layers, and it still does not close the mid-turn gap, since the figure is stale by one round.

A third option worth considering: **don't clamp — fail fast with a legible error**, exactly as #951 chose for the unsatisfiable thinking-budget case. A clear "this conversation no longer fits; compact or start fresh" beats both an opaque provider 400 and a silent truncation.

## Suggested next step

Decide clamp-vs-fail-fast first; the implementation follows cheaply from that. My weak preference is fail-fast with a message naming the numbers, consistent with #951, plus lowering `DEFAULT_AUTO_COMPACT_THRESHOLD` for models whose `window − ceiling` headroom is under 10%.

## Verification status

Every code claim above was read directly at the cited path. **Not verified live**: the actual Anthropic API rejection (I have not reproduced a 400 against a real endpoint) — the error is documented as `input length and max_tokens exceed context limit`, but the exact trigger boundary should be confirmed before choosing the margin. No live repro was attempted because it would require deliberately burning ~136k input tokens against a paid endpoint.

griffinwork40:
## Summary

Adds a fail-fast context-overflow guard that emits a legible error **before** sending a request that the provider would reject with HTTP 400 when `lastUsage.input_tokens + max_tokens > contextWindow`.

Closes #962.

## Root Cause

`resolveMaxTokens` (anthropic-direct) and `resolveEffectiveMaxOutputTokens` (openai-compatible) take no messages and no usage, so they cannot see how much context is already consumed. The resolved `max_tokens` value is sent to the wire verbatim. When `input + max_tokens > context_window` — the haiku threshold is ~136k input — the provider rejects with HTTP 400.

## Approach: Fail-Fast (Option 3)

Consistent with #951, which chose the same pattern for the unsatisfiable thinking-budget case. On each turn start, read the stale-by-one-round `lastUsage` as a conservative lower bound:
- **Skip the first turn** (no `lastUsage` yet) — conservative by design.
- **Never silently shrink** `max_tokens`; prefer a loud, actionable error over silent truncation.

The guard yields an error event (rather than throwing) so the abort controller is cleared before control returns — throwing here would leave the slot set and make `compact()` return `'turn-in-flight'`.

## Changes

| File | What changed |
|------|-------------|
| `src/agent/providers/shared/auto-compact.ts` | Added `guardContextOverflow()`, `safeAutoCompactThresholdFor()`, updated `resolveAutoCompactThreshold()` to accept a model for model-aware threshold capping |
| `src/agent/providers/anthropic-direct/query-turn-driver.ts` | Call `guardContextOverflow()` at turn start, yield error if it throws |
| `src/agent/providers/openai-compatible/query.ts` | Same guard via `checkContextOverflow()`, reset `lastUsage = null` after auto-compaction so the guard doesn't false-positive on the next turn |
| `src/agent/providers/openai-compatible/query/context-overflow.ts` | New: extracted helper that wraps `guardContextOverflow` into a null-or-Error pattern for the OpenAI provider |
| `src/agent/providers/openai-compatible/query/compaction-guard.ts` | New: extracted `provesResponsesCompactionUnsupported` predicate (moved from query.ts to stay within line ceiling) |
| `src/agent/providers/shared/context-overflow-guard.test.ts` | 25 new unit tests for `guardContextOverflow`, `safeAutoCompactThresholdFor`, and the model-aware threshold resolution |

## Test Coverage

25 new unit tests cover:
- Happy path (no overflow, guard passes)
- Exact-boundary cases (at the limit, at limit + 1)
- First-turn skip (null lastUsage)
- Error message format and legibility
- `safeAutoCompactThresholdFor` clamping logic
- Model-aware `resolveAutoCompactThreshold` integration

All 16,230 existing tests continue to pass (one pre-existing flaky symlink test in `write-denylist.test.ts` is unrelated to this change).

## Lint & File Size

- `pnpm lint` ✅ passes
- `pnpm audit:filesize:check` — the 6 pre-existing violations on main are unchanged; no new violations introduced by this PR

### DIFF
--- a/src/agent/providers/anthropic-direct/provider-query-build.ts
+++ b/src/agent/providers/anthropic-direct/provider-query-build.ts
@@ -96,6 +96,14 @@ export function buildProviderQuery(
       });
 
   const resolvedEffort = resolveEffort(config.effort, model);
+  // Use requestedModel (the alias, e.g. sonnet_1m) rather than the resolved
+  // wire id so safeAutoCompactThresholdFor sees the full 1M window when
+  // the alias carries that budget. Extracted to avoid calling
+  // resolveAutoCompactThreshold twice. (#1014 Item 3 & 4)
+  const autoCompactThreshold = resolveAutoCompactThreshold(
+    config.autoCompact,
+    typeof config.model === 'string' && config.model.length > 0 ? config.model : model,
+  );
   return new AnthropicDirectQuery({
     client,
     // In local-server mode, downgrade the effective auth mode to 'api-key'
@@ -163,9 +171,7 @@ export function buildProviderQuery(
       ctx.setCurrentPermissionMode(mode);
     },
     ...(ctx.mcpManager !== undefined ? { mcpManager: ctx.mcpManager } : {}),
-    ...(resolveAutoCompactThreshold(config.autoCompact, model) !== undefined
-      ? { autoCompactThreshold: resolveAutoCompactThreshold(config.autoCompact, model) }
-      : {}),
+    ...(autoCompactThreshold !== undefined ? { autoCompactThreshold } : {}),
     // Thread the resolved hook registry into the query so auto-compaction
     // can dispatch PreCompact(trigger:'auto') before calling compact().
     // resolveSessionHookRegistry is already called above for the dispatcher;
--- a/src/agent/providers/anthropic-direct/provider-query-build.ts
+++ b/src/agent/providers/anthropic-direct/provider-query-build.ts
@@ -163,8 +163,8 @@ export function buildProviderQuery(
       ctx.setCurrentPermissionMode(mode);
     },
     ...(ctx.mcpManager !== undefined ? { mcpManager: ctx.mcpManager } : {}),
-    ...(resolveAutoCompactThreshold(config.autoCompact) !== undefined
-      ? { autoCompactThreshold: resolveAutoCompactThreshold(config.autoCompact) }
+    ...(resolveAutoCompactThreshold(config.autoCompact, model) !== undefined
+      ? { autoCompactThreshold: resolveAutoCompactThreshold(config.autoCompact, model) }
       : {}),
     // Thread the resolved hook registry into the query so auto-compaction
     // can dispatch PreCompact(trigger:'auto') before calling compact().
--- a/src/agent/providers/anthropic-direct/query-runtime.ts
+++ b/src/agent/providers/anthropic-direct/query-runtime.ts
@@ -326,13 +326,13 @@ export class AnthropicDirectQuery implements ProviderQuery {
   }
 
   async compact(): Promise<ProviderCompactResult> {
-    return compactQueryHistory({
-      state: this.state,
-      abort: this.abort,
-      retry: this.retry,
+    const r = await compactQueryHistory({
+      state: this.state, abort: this.abort, retry: this.retry,
       initSessionId: this.initSessionId,
       ...(this.traceWriter ? { traceWriter: this.traceWriter } : {}),
     });
+    if (r.compacted) this.state.lastUsage = null; // #962: stale usage → false-positive
+    return r;
   }
 
   listRewindTargets(): RewindTarget[] {
--- a/src/agent/providers/anthropic-direct/query-turn-driver.auto-compact.ts
+++ b/src/agent/providers/anthropic-direct/query-turn-driver.auto-compact.ts
@@ -63,12 +63,15 @@ export async function* maybeAutoCompact(ctx: TurnDriverContext): AsyncGenerator<
         trigger: 'auto',
       });
     }
-    await ctx.compact();
+    const compactResult = await ctx.compact();
     // Reset lastUsage after compaction so the overflow guard (#962) does not
     // false-positive on the next turn: the stale count is no longer a valid
     // lower bound once history has been truncated. The next API call will
     // repopulate it from the actual response.
-    ctx.state.lastUsage = null;
+    // Conditioned on `compacted` — a no-op (history too short, summarization
+    // failure) must preserve the near-limit evidence; unconditionally nulling
+    // it would skip both the guard and auto-compaction on the next turn.
+    if (compactResult.compacted) ctx.state.lastUsage = null;
   } catch (compactErr) {
     if (compactErr instanceof HookBlockedError) {
       // Hook blocked auto-compaction — skip this turn's compaction
--- a/src/agent/providers/anthropic-direct/query-turn-driver.auto-compact.ts
+++ b/src/agent/providers/anthropic-direct/query-turn-driver.auto-compact.ts
@@ -1,0 +1,80 @@
+/**
+ * Auto-compaction helper extracted from query-turn-driver.ts to stay within the
+ * 350-line file ceiling. Behavior is identical — only the physical location
+ * changed.
+ *
+ * @module agent/providers/anthropic-direct/query-turn-driver.auto-compact
+ */
+
+import type { ProviderEvent } from '../../provider.js';
+import { autoCompactLimitFor } from '../../model-limits.js';
+import { contextWindowTokensUsed, shouldAutoCompact } from './query/auto-compact.js';
+import { HookBlockedError } from '../../../utils/errors.js';
+import type { TurnDriverContext } from './query-turn-driver.js';
+
+/**
+ * Auto-compaction: fire at the natural turn boundary — after the per-turn
+ * event loop exits and the abort slot is cleared — so we never trigger while a
+ * tool call is in flight.
+ *
+ * External constraint: `abort.isIdle()` MUST be true here (cleared by
+ * `abort.clear(controller)` in the caller's finally) so compact-handler's own
+ * isIdle() guard passes and the 'turn-in-flight' bail is not hit spuriously.
+ * Do not call this from inside the inner for-await — compact() mutates
+ * state.messages in place and must only run at a clean turn boundary, never
+ * mid-tool-call.
+ *
+ * Yields nothing; it is a generator purely so the caller's `yield*` keeps the
+ * original inline control flow (a throw here propagates identically).
+ */
+export async function* maybeAutoCompact(ctx: TurnDriverContext): AsyncGenerator<ProviderEvent, void, void> {
+  if (ctx.state.autoCompactThreshold === undefined || ctx.state.closed) return;
+  const usage = ctx.state.lastUsage;
+  // requestedModel (not the wire currentModel) so 1M aliases use their
+  // true window — opus_1m resolves to the same wire id as opus but must
+  // compact at ~90% of 1M, not 200k. autoCompactLimitFor additionally
+  // caps the DEFAULT `sonnet` at a 200k working budget: its 1M window is
+  // truthful, but base sessions compact early for cost/latency, while the
+  // `sonnet_1m` opt-in bypasses the cap. See model-limits.ts.
+  const compactionLimit = autoCompactLimitFor(ctx.state.requestedModel);
+  if (usage === null || compactionLimit <= 0) return;
+  // Use the context-window footprint (input + cache_read +
+  // cache_creation + output for the last round), NOT input+output
+  // alone — Anthropic's input_tokens excludes cache, so the cached
+  // conversation prefix (often the bulk of the window) must be
+  // counted or compaction never fires before the window overflows.
+  const usedTokens = contextWindowTokensUsed(usage);
+  if (!shouldAutoCompact(usedTokens, compactionLimit, ctx.state.autoCompactThreshold)) return;
+  // Fire-and-await: compact() is async but we hold the turn
+  // boundary here (generator suspended at promptIterator.next()
+  // on the next iteration). Awaiting inline keeps the ordering
+  // deterministic and avoids a dangling promise race.
+  //
+  // Invariant: dispatch PreCompact(trigger:'auto') BEFORE compact()
+  // so registered hooks can block or observe the operation, mirroring
+  // the manual-compact paths in REPL (/compact) and Telegram. A block
+  // decision throws HookBlockedError — caught here to skip compaction
+  // for this turn without propagating to the outer error handler.
+  try {
+    if (ctx.hookRegistry) {
+      await ctx.hookRegistry.dispatch({
+        event: 'PreCompact',
+        sessionId: ctx.initSessionId,
+        trigger: 'auto',
+      });
+    }
+    await ctx.compact();
+    // Reset lastUsage after compaction so the overflow guard (#962) does not
+    // false-positive on the next turn: the stale count is no longer a valid
+    // lower bound once history has been truncated. The next API call will
+    // repopulate it from the actual response.
+    ctx.state.lastUsage = null;
+  } catch (compactErr) {
+    if (compactErr instanceof HookBlockedError) {
+      // Hook blocked auto-compaction — skip this turn's compaction
+      // without surfacing an error; the session continues normally.
+    } else {
+      throw compactErr;
+    }
+  }
+}
--- a/src/agent/providers/anthropic-direct/query-turn-driver.ts
+++ b/src/agent/providers/anthropic-direct/query-turn-driver.ts
@@ -128,22 +128,19 @@ export async function* driveTurns(ctx: TurnDriverContext): AsyncGenerator<Provid
       // placeholders so the API contract holds and the user can continue.
       repairOrphanToolUses(ctx.state.messages);
 
-      // Append the new user turn to history. Strings and content-block
-      // arrays both ride through as-is — Anthropic's MessageParam accepts
-      // either shape natively.
-      ctx.state.messages.push({ role: 'user', content: turn.content });
-
-      const system = ctx.composeSystem();
-
       // Context-overflow guard (#962): fail fast with a legible error BEFORE
-      // sending the request when we know the provider will 400 it. Uses the
-      // stale-by-one-round `lastUsage.contextWindowTokens` as a lower bound —
+      // sending the request when we know the provider will 400 it. Runs BEFORE
+      // the history push so the user message is not left in history on overflow
+      // (allowing /compact to succeed cleanly from the outer while loop). Uses
+      // the stale-by-one-round `lastUsage.contextWindowTokens` as a lower bound —
       // conservative by design (skips the first turn when lastUsage is null,
       // never silently shrinks max_tokens). See shared/auto-compact.ts.
       // Yields an error event (rather than throwing) so the abort controller is
       // cleared before control returns to the outer loop — throwing here would
       // skip the inner try/finally that clears it and leave compact() returning
       // 'turn-in-flight' on any subsequent manual compact call.
+      // Uses `continue` (not `return`) so the outer while loop survives: the
+      // user can /compact and send a new message without the session dying.
       {
         let overflowErr: Error | undefined;
         try {
@@ -159,9 +156,16 @@ export async function* driveTurns(ctx: TurnDriverContext): AsyncGenerator<Provid
         if (overflowErr !== undefined) {
           ctx.abort.clear(controller);
           yield { type: 'error', error: overflowErr };
-          return;
+          continue;
         }
       }
+
+      // Append the new user turn to history. Strings and content-block
+      // arrays both ride through as-is — Anthropic's MessageParam accepts
+      // either shape natively.
+      ctx.state.messages.push({ role: 'user', content: turn.content });
+
+      const system = ctx.composeSystem();
 
       // Snapshot preference + eligibility exactly once. The resulting headers
       // and body intent live on one immutable RunTurnInput for every round/retry.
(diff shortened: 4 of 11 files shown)


griffinwork40:
All six correctness claims verified against the diff. Now I can synthesize the full review.

---

## PR #1014 Review: fail-fast context-overflow guard for max_tokens exceeding window

**PR**: `afk/fix-962-max-tokens` → `main` · +514 / −45 / 9 files · Closes #962

### Verdict: **APPROVE with minor observations**

The implementation is clean, well-documented, and correctly scoped. The core guard logic is mathematically sound, both provider integrations follow the same yield-error-return pattern, and the 228-line test file covers the critical paths. No blocking issues found.

---

### Correctness ✅

All six load-bearing correctness properties verified against the diff:

| Property | Status | Evidence |
|----------|--------|----------|
| Overflow comparison uses `<=` (no off-by-one) | ✅ | `if (projected <= contextLimit) return;` — diff:440 |
| Anthropic abort-clear → yield → return | ✅ | diff:63–65 |
| OpenAI abort-clear → yield → return | ✅ | diff:215–218 |
| `model?` param is optional (backward-compatible) | ✅ | `model?: string` — diff:475, test at diff:714 |
| `lastUsage = null` inside try (skipped on compaction failure) | ✅ | diff:80 (anthropic), diff:195 (openai) |
| `contextWindowTokensUsed({})` returns 0 (first-turn skip) | ✅ | source:46–48, existing test at auto-compact.test.ts:175 |

The `safeAutoCompactThresholdFor` math is correct: `(window - ceiling) / window - SAFE_THRESHOLD_SLACK`, returning the default 0.9 unchanged when `compactionLimit < contextWindow` (budget-based models like base sonnet/opus already compact below the overflow boundary).

---

### API Compatibility ✅ **COMPATIBLE**

| Change | Compat? |
|--------|---------|
| `resolveAutoCompactThreshold(autoCompact, model?)` — optional 2nd param | ✅ All existing single-arg call sites unaffected |
| `guardContextOverflow`, `safeAutoCompactThresholdFor` re-exported via anthropic-direct shim | ✅ New exports, no existing ones changed |
| `provesResponsesCompactionUnsupported` moved to `query/compaction-guard.ts` | ✅ Import updated in `query.ts`; function is internal, not part of public API |
| Test model `gpt-4o-mini` → `gpt-5.5` | ✅ `gpt-5.5` confirmed in `MODEL_CONTEXT_LIMITS` and `MODEL_MAX_OUTPUT_LIMITS` |

---

### Security ✅

- **Info disclosure**: Error message contains token counts and model name — appropriate for a CLI tool. No credentials, API keys, or internal state leaked.
- **Input validation**: Guard short-circuits on `<= 0` for all three numeric inputs, which correctly handles 0, negatives. `NaN` passes the `<= 0` check as `false`, but `NaN + NaN <= contextLimit` is also `false`, so the guard never falsely clears — safe by IEEE 754 semantics.
- **DoS**: Not applicable — inputs come from provider-reported usage, not user-controlled data.

---

### Findings

**1. [MEDIUM · observability] Guard-fired errors are indistinguishable from HTTP 400 in the witness trace**

Both providers yield `{ type: 'error', error }` — the same shape as a provider rejection. A dedicated trace phase event (e.g. `session_phase: 'context_overflow_guard'`) before yielding would let `afk trace show` distinguish pre-flight guard fires from wire errors. Not a blocker — the error message text (`[afk] Context-window overflow:`) is unique enough to grep, but structured trace distinction would improve debugging.

**2. [LOW · test gap] No integration test exercises the guard inside the turn driver loop**

The 25 unit tests cover the pure functions thoroughly. A turn-driver integration test that mocks `lastUsage` above the overflow boundary and asserts the error event is yielded (without hitting the wire) would close the integration gap. Not blocking — the unit coverage is solid and the integration points are thin.

**3. [LOW · test gap] `NaN`/`Infinity` inputs to `resolveAutoCompactThreshold` untested**

The source guards `!Number.isFinite(t)`, but no test exercises `{ threshold: NaN }` or `{ threshold: Infinity }`. These are defensive paths that work correctly but have no regression protection.

**4. [INFO · design] One-turn blind spot after compaction is acceptable**

Resetting `lastUsage = null` after compaction means the guard skips the immediate next turn. This is the correct tradeoff: the pre-compaction token count is now stale-high and would false-positive. The provider's own 400 rejection still protects that turn if compaction reclaimed insufficient space.

**5. [INFO · design] `SAFE_THRESHOLD_SLACK = 0.04` is a sensible buffer**

The 4% slack between the mathematical overflow point and the compaction trigger avoids a race where a user turn arriving between compaction eligibility and actual compaction pushes past the boundary. The table in the comments correctly shows the capped thresholds for each model archetype.

---

### File-split quality ✅

Two new files (`compaction-guard.ts` at 46 lines, `context-overflow.ts` at 51 lines) are clean extractions from `query.ts` — motivated by the 350-line ceiling. Both preserve the original JSDoc contracts verbatim. The imports are correctly updated.

---

### Done

**Recommendation: Merge.** The core logic is correct, backward-compatible, and well-tested. The two medium/low findings (trace observability, integration test gap) are improvement opportunities, not blockers. The PR cleanly addresses #962's failure mode (400 rejections when `input + max_tokens > window`) with the right design choice (fail-fast, never silently shrink).

---

griffinwork40:
Confirmed: `push` at line 134 → guard at line 150 (anthropic-direct); `push` at line 481 → guard at line 491 (openai-compatible). The orphaned-message finding is verified independently.

---

## Review: PR #1014 — `fix(#962): fail-fast context-overflow guard for max_tokens exceeding window`

**Reviewed ref:** `066a7939c3c53af6d22efd934c02fee28d5d5f03`
**Change type:** hotfix · **+603 / -113 / 11 files** · **Regime:** full

### Summary

Adds a client-side pre-flight guard that detects when `lastUsage.input_tokens + max_tokens` exceeds the model's context window and short-circuits with a legible error before wasting an API round-trip. Both `anthropic-direct` and `openai-compatible` providers are wired. Also introduces `safeAutoCompactThresholdFor()` to cap the auto-compaction threshold so compaction fires before overflow on narrow-headroom models (haiku, 1M opt-ins). Includes 25 new unit tests. The change is strictly better than the pre-fix provider-side 400.

---

### Findings

#### 1. `medium · blocking:true · high confidence · correctness`
**`query-turn-driver.ts:134→150` + `query.ts:481→491`** · ref:066a7939 · file-state (verified)

Guard fires after history mutation; orphaned user message is never rolled back.

Both providers push the user turn to durable history **before** the guard runs. When the guard fires, it yields an error and `return`s from the generator — permanently exhausting the session's sole `providerIterator` (line 351 of `agent-session.ts`: created once, never re-created). The orphaned user message remains in history without a matching assistant response. Subsequent `sendMessageStream` calls get `done: true` immediately — the session is unresponsive.

The error message suggests "run /compact to summarise the conversation history" — this is misleading because the iterator is dead; `/compact` cannot revive it. Only "start a new session" is actionable.

```
// anthropic-direct: push at 134, guard at 150
ctx.state.messages.push({ role: 'user', content: turn.content }); // :134
// ... 16 lines later ...
guardContextOverflow(...); // :150

// openai-compatible: push at 481, guard at 491
this.priorTurns.push({ ... }); // :481
// ... 10 lines later ...
const overflowErr = checkContextOverflow(...); // :491
```

**Fix:** Move the guard before `push()` in both providers — `lastUsage` does not depend on the new message. Remove or downgrade the `/compact` clause in the error message.

---

#### 2. `low · non-blocking · spec-compliance`
**`shared/auto-compact.ts:294–307`** · ref:066a7939 · file-state (verified)

`resolveAutoCompactThreshold` silently caps a user-configured threshold (e.g. 0.85 → ~0.64 for haiku) with no log or trace event. The PR's stated intent explicitly adopts a "loud, actionable error over silent truncation" philosophy — this cap applies the inverse philosophy to a closely related quantity. Defensively correct, but inconsistent with stated principles.

**Fix:** Emit a trace event when the cap changes the configured value.

---

#### 3. `low · non-blocking · test-coverage`
**`context-overflow-guard.test.ts`** · ref:066a7939 · file-state (verified)

No integration test covers the guard-fires path in either provider's turn driver. Unit tests verify throw/no-throw logic but not: error event emission through the generator, abort-slot state afterward, or history length post-guard. An integration test here would have caught Finding 1. (Grep pattern: `checkContextOverflow|guardContextOverflow` in `query.test.ts` → zero matches.)

---

#### 4. `low · non-blocking · api-compat`
**`query/context-overflow.ts`** · ref:066a7939 · file-state (verified)

`checkContextOverflow` takes two model parameters (`currentModel` for limit math, `configuredModel` for error text) with no doc distinguishing their roles. Future callers will transpose them.

**Fix:** Add a JSDoc annotation clarifying the distinct roles.

---

#### 5. `nit · non-blocking · perf-observability`
**`provider-query-build.ts:166–167`** · ref:066a7939 · file-state (verified)

`resolveAutoCompactThreshold(config.autoCompact, model)` is called twice (truthiness check + value). Pure function, zero correctness risk.

**Fix:** `const t = resolveAutoCompactThreshold(config.autoCompact, model); ...(t !== undefined ? { autoCompactThreshold: t } : {})`

---

### Decision: DO NOT MERGE — 1 medium blocking (correctness); 3 low, 1 nit.

The change is net-positive and should merge once Finding 1 is resolved (move guard before push, fix error message). Findings 2–5 are non-blocking and can ship in-PR or be deferred.

---

### What was not checked
- Citations verified inline against branch HEAD `066a7939`.
- Stated intent: PR #1014 title + body — spec-compliance assessed.
- Runtime tests were not executed (static analysis only).
- Third provider adapters beyond anthropic-direct and openai-compatible were not audited.
- Full rewind/history-replay code paths not explored (Finding 1 notes a theoretical hazard for future features).
- No profiling data for token-count computation frequency.

---

griffinwork40:
The finding about `opts.model` being the wire id is valid — OpenAI models typically don't have alias ≠ wire-id splits, but the asymmetry with anthropic-direct's deliberate alias handling is a real code smell worth noting.

Now I have all findings verified. Let me compose the synthesis:

---

## PR #1014 Review — `fix(#962): fail-fast context-overflow guard for max_tokens exceeding window`

**Reviewed ref:** `76b3c6f40eb22b7f7a6de71964123e3990dbee0a`
**Change type:** feature (new guard + threshold cap)
**Regime:** full (656 lines, 11 files)

---

### Findings

**1.** medium · blocking:true · confidence:high · test-coverage · `src/agent/providers/shared/context-overflow-guard.test.ts` · ref:76b3c6f · citation-type:file-state · The new test file covers `guardContextOverflow` and `safeAutoCompactThresholdFor` as pure-function unit tests (25 tests), but no integration test in either provider's test suite verifies the guard firing on a real turn: emitting an `error` event, not pushing the user message to history, and the session surviving for a subsequent turn. The overflow→compact→success recovery path (`lastUsage` nulled after `/compact`, guard passes on next turn) also has zero end-to-end coverage. · evidence: `git grep` of all test files for `guardContext|checkContextOverflow` shows zero call sites in provider integration tests; all matches in `query.test.ts` document scenarios where the guard does NOT fire. · suggestion: Add one integration test per provider that seeds `lastUsage` with overflow-triggering tokens, runs a turn, asserts an `error` event (not throw), asserts history length unchanged, then resets `lastUsage` and asserts the next turn succeeds. · waived: bounded, non-data-affecting defect — the guard is exercised as a pure function, and the integration sites are straightforward yield-and-continue/return; a missing integration test does not produce wrong output.

**2.** medium · blocking:false · confidence:high · correctness · `src/agent/providers/openai-compatible/query.ts:288` · ref:76b3c6f · citation-type:file-state · The openai-compatible constructor passes `opts.model` (the resolved wire id) to `resolveAutoCompactThreshold`, while anthropic-direct explicitly passes `config.model` (the user-facing alias) because `safeAutoCompactThresholdFor` looks up `MODEL_AUTOCOMPACT_BUDGET`/`MODEL_CONTEXT_LIMITS` by that key. For current OpenAI models (no alias ≠ wire-id split) the result is identical, but the asymmetry is undocumented and would silently produce a wrong threshold cap if an openai-compatible alias ever diverged from its wire id. · evidence: `provider-query-build.ts:98-105` documents the alias distinction; `query.ts:288` passes `opts.model` with no equivalent comment. · suggestion: Either document at line 288 that the wire id is the correct key for all openai-compatible models, or mirror the anthropic-direct pattern: `resolveAutoCompactThreshold(opts.config.autoCompact, opts.config.model ?? opts.model)`. · waived: the defect path is currently unreachable — all OpenAI model ids in `MODEL_CONTEXT_LIMITS` are keyed by wire id, so the lookup succeeds today.

**3.** low · blocking:false · confidence:high · correctness · `src/agent/providers/shared/auto-compact.ts:219` · ref:76b3c6f · citation-type:file-state · `safeAutoCompactThresholdFor` returns the default 0.9 unconditionally when `compactionLimit < contextWindow`, trusting that a budget-constrained model compacts well before the overflow point. The invariant (`budget × 0.9 + ceiling ≤ window`) holds for all current entries (sonnet: 200k × 0.9 + 128k = 308k ≤ 1M) but is not structurally enforced — a future budget entry large enough to violate it would silently bypass the guard. · suggestion: Replace `if (compactionLimit < contextWindow)` with `if (compactionLimit * DEFAULT_AUTO_COMPACT_THRESHOLD + ceiling <= contextWindow)` so the safety property is enforced for any budget value.

**4.** low · blocking:false · confidence:medium · api-compat · `src/agent/providers/anthropic-direct/query/auto-compact.ts:1-24` · ref:76b3c6f · citation-type:file-state · The re-export shim omits `resolveAutoCompactThreshold` and `DEFAULT_AUTO_COMPACT_THRESHOLD` from its barrel. No production file inside `anthropic-direct/` imports either through this shim (verified: `provider-query-build.ts` imports from `./resolve-params.js`, which re-exports from `../shared/auto-compact.js`), so no compilation failure today. · suggestion: Add the two symbols to the shim for completeness, or add a comment noting the intentional omission.

**5.** low · blocking:false · confidence:high · spec-compliance · `src/agent/providers/anthropic-direct/query-turn-driver.ts:156` · ref:76b3c6f · citation-type:diff-context · The overflow→compact→success round-trip (overflow error → user runs `/compact` → `lastUsage` nulled → guard skips → turn succeeds) is the primary user-facing recovery path documented in the error message. The code path exists but has zero test coverage. · (This is a duplicate dimension of finding #1 — both name the same gap.)

**6.** low · blocking:false · confidence:medium · perf-observability · `src/agent/providers/openai-compatible/query/context-overflow.ts:40-54` · ref:76b3c6f · citation-type:file-state · The guard inspects per-turn usage silently — no trace event or debug log records a near-overflow situation (e.g. 90%+ of the limit) before the hard error fires, making gradual window fill invisible in the witness trace. · suggestion: Emit a debug log or trace event when the guard inspects non-zero usage within 10% of the limit.

**7.** nit · blocking:false · confidence:high · spec-compliance · `src/agent/providers/shared/auto-compact.ts:212` · ref:76b3c6f · citation-type:file-state · `safeAutoCompactThresholdFor`'s early-return guard checks `contextWindow <= 0 || ceiling <= 0` but not `compactionLimit <= 0`. If `autoCompactLimitFor` ever returned 0, the comparison `compactionLimit < contextWindow` would pass, silently returning the default. Defensive: add `compactionLimit` to the guard.

---

### Cross-cutting observations

**Behavioral asymmetry (intentional, documented).** The anthropic-direct guard uses `continue` (outer `while` loop survives) and the openai-compatible guard uses `return` (exits `_runTurnInner`, outer `while` loop continues naturally via `yield*`). Both produce the same outcome: session survives, user can `/compact` and retry. The anthropic-direct path comments on this; the openai-compatible path does not.

**`lastUsage = null` reset in 3 sites.** `compact()` in `query-runtime.ts`, `compact()` in `query.ts`, and `maybeAutoCompact` in `query-turn-driver.auto-compact.ts`. The anthropic-direct `compact()` and `maybeAutoCompact` sites are NOT redundant: `compact()` covers manual `/compact` calls; `maybeAutoCompact` covers the auto-compact turn-boundary. The openai-compatible `compact()` covers both manual and auto paths (auto calls `compactHistory()` directly, but the manual entrypoint goes through `compact()`). All 3 are correctly conditioned on `r.compacted === true`.

**Test model migration.** The `gpt-4o-mini` → `gpt-5.5` switch in test code is well-motivated: 850k usage against gpt-5.5's 1M window triggers compaction (~832k threshold) without triggering the new overflow guard (~872k), while gpt-4o-mini's 128k window would have the guard fire at 64k usage, preventing compaction tests from running. Comments explain the math at each site.

---

### What was not checked

- Citations verified inline against branch HEAD `76b3c6f40eb22b7f7a6de71964123e3990dbee0a`.
- Stated intent: PR #1014 title+body — spec-compliance assessed.
- Did not run `pnpm test` or `pnpm lint` (read-only review).
- Did not verify `gpt-5.5` exists in `MODEL_CONTEXT_LIMITS` at the reviewed ref (the test comments assert a 1M window).
- Did not review the `website/` or Telegram surfaces for any overflow-guard integration.

---

### Decision: MERGE — 0 blocking (1 medium waived: test-coverage, bounded non-data-affecting defect; 1 medium waived: correctness, currently unreachable path; 3 low, 1 nit).

Finding #1 (integration test coverage for the guard-fires path) is the most impactful gap — the pure-function unit tests verify the math, but the yield/continue/return integration and the recovery round-trip are unexercised. Waived because the integration sites are mechanically simple (yield + continue/return) and the guard is conservative by design (fires only on mathematical certainty, never on estimates). Worth addressing in a follow-up.

---

PR #1044, opened 2026-08-13
fix(witness): track pre/post-filter session counts in search_witness (#1035)

### LINKED ISSUE #1035: witness tools: sessionsScanned reports post-filter count, misleading callers
## Problem

`searchAcrossSessions` returns `sessionsScanned: traces.length` after the `since` date filter has already been applied. A caller requesting `sessions: 50, since: "2026-01-01"` that finds 3 sessions in the date range gets `sessionsScanned: 3` — indistinguishable from "only 3 sessions exist."

## Location

`src/agent/tools/handlers/witness.query.ts` — the `searchAcrossSessions` return value.

## Suggested fix

Track both counts:
- `sessionsScanned` — pre-filter count (how many were actually read)
- `sessionsMatched` — post-filter count (how many survived the date gate)

Or rename the existing field to `sessionsMatched` to avoid the semantic mismatch. Update the schema description in `schemas.witness.ts` to match.

## Context

Identified during re-review of #1034. Waived as non-blocking (no data integrity impact, field naming is follow-up-grade).

griffinwork40:
## Problem

`searchAcrossSessions` returns `sessionsScanned: traces.length` **after** the `since` date filter has already been applied. A caller requesting `sessions: 50, since: "2026-01-01"` that finds 3 sessions in the date range gets `sessionsScanned: 3` — indistinguishable from "only 3 sessions exist."

## Fix

Track both counts in `SearchWitnessResult`:
- `sessionsScanned` — pre-filter count (how many session traces were read from disk)
- `sessionsMatchedDateFilter` — post-`since`-filter count (only present when `since` was used)

Updated schema description in `schemas.witness.ts` to document both fields.

## Changes

- `src/agent/tools/handlers/witness.query.ts` — capture pre-filter count before the `since` filter runs; add `sessionsMatchedDateFilter` to result type and return value
- `src/agent/tools/schemas.witness.ts` — document both result fields
- `src/agent/tools/handlers/witness.query.test.ts` — 3 new tests + fixed existing assertion

## Test Results

25/25 tests pass. `pnpm lint` clean.

Closes #1035

### DIFF
--- a/src/agent/tools/handlers/witness.query.test.ts
+++ b/src/agent/tools/handlers/witness.query.test.ts
@@ -275,7 +275,7 @@ describe('searchAcrossSessions — invalid since date', () => {
     // No sessions in temp home, so result is empty — but must not throw.
     await expect(
       searchAcrossSessions({ query: 'x', since: '2024-01-01' }),
-    ).resolves.toMatchObject({ query: 'x', sessionsScanned: 0 });
+    ).resolves.toMatchObject({ query: 'x', sessionsScanned: 0, sessionsMatchedDateFilter: 0 });
   });
 });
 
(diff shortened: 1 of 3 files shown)


griffinwork40:
Superseded — main already has sessionsAvailable (pre-filter count) and sessionsSearched (post-filter count) via #1042 + d91ef1cf, solving the exact problem this PR targeted. Closing as redundant.

PR #1047, opened 2026-08-13
test(improve): golden-fingerprint test for denial reason slug stability (#853)

### LINKED ISSUE #853: Rewording the fork denial reason silently rotates the subagent-read-denial card slug (no test)
## Summary

Rewording the sub-agent path-denial reason string silently rotates the `subagent-read-denial` failure-card slug, orphaning accumulated cards and resetting the severity ladder. **No test catches it.** The producer and the fingerprint consumer have no shared fixture.

Surfaced by #851, which reworded that string deliberately and had to reason about the consequence by hand. The next person to touch the wording will not know to.

## Mechanism

- Producer: the denial reason is assembled in `src/agent/tools/hooks/path-approval-hook.ts` (~:305-311), with the remedy body now in `src/agent/tools/hooks/fork-denial-remedy.ts`.
- Consumer: `src/improve/scan/detectors/subagent-read-denial.ts` calls `normalizeReason` (`:84-86`, collapsing path-shaped tokens to `<path>`), then `computeFingerprint` (`:148-151`) SHA-256s the **entire** `(hookEvent, normalizedReason)` tuple, then `makeSlug` (`:156-158`) takes the first 12 hex.
- Therefore **any** change to the remedy prose — not just the load-bearing prefix — produces a new slug.
- Slugs are durable, externally-referenced identifiers: `subagent-read-denial-ab89c2bd6a6f` was cited in a source comment in `src/agent/subagent.ts` (annotated as historical by #851) and names real files under `~/.afk/agent-framework/improve/failure-cards/`.

## Why CI does not catch it

Every test on both sides builds its **own** fixture string rather than importing the producer:

- `src/improve/scan/detectors/subagent-read-denial.test.ts:77` — local `denial()` helper
- `src/improve/scan/index.test.ts:152-153` — inline fixture
- `src/agent/tools/hooks/path-approval-hook.test.ts:188,283,305` — `toContain('Sub-agent path access denied')`, i.e. asserts only the byte-stable prefix, never the body

So the producer and the fingerprinting consumer are coupled in production and fully decoupled in test. A reword is green in CI and quietly resets telemetry.

Worth noting the distinction #851 established, because it is easy to conflate:

- `Sub-agent path access denied:` — **byte-contract**, matched by `.includes()` at `src/agent/tools/denial-circuit-breaker.ts:102`. Breaking this breaks the circuit breaker. Already documented at both ends.
- The remedy body — **free to reword, but fingerprint-rotating.** This is the undocumented, untested half.

Detection gating itself is unaffected (`:107-111` keys on `hookEvent`/`decision`/`blockedTool`, not on the message), so a reword does not stop detection — it just re-keys the bucket.

## Suggested fix

A test that fails loudly when the producer's reason string changes shape. Options, roughly in order of preference:

1. **Golden-fingerprint test.** Have the detector test import the real producer (`buildForkDenialRemedy` + the prefix assembly), compute the fingerprint, and assert it against a checked-in constant. Rewording then fails with a diff that says exactly what happened, and the fix is to update the constant *deliberately* — which is the behaviour we want.
2. **Shared fixture module.** Replace the hand-built fixture strings in both detector tests with a helper that calls the real producer, so the two sides can never drift silently.
3. **Fingerprint only the stable part.** Narrow `normalizeReason` to hash `(hookEvent, prefix, mode)` rather than the whole body, making the slug immune to remedy rewording. Most robust, but changes card identity once on introduction and reduces the detector's ability to distinguish differently-worded denials — probably not worth it.

(1) plus (2) is likely the right combination: (1) makes the break loud, (2) stops it recurring.

## Acceptance

A PR that rewords the remedy body in `fork-denial-remedy.ts` without updating a golden constant should **fail** CI.

griffinwork40:
## Problem

Rewording the sub-agent path-denial reason string in `fork-denial-remedy.ts` silently rotates the `subagent-read-denial` failure-card slug, orphaning accumulated cards and resetting the severity ladder. No test catches this because every test builds its own fixture string rather than importing the real producer.

### Golden-fingerprint test (new file)
`src/improve/scan/detectors/subagent-read-denial-stability.test.ts` — imports the real producer (`buildForkDenialRemedy`) and real prefix (`SUBAGENT_PATH_DENIAL_REASON_PREFIX`), assembles the full denial reason exactly as production does, normalizes, hashes, and asserts against checked-in golden constants. A reword now **fails CI loudly** with a diff explaining what happened.

| Mode | Golden Fingerprint |
|------|-------------|
| `read` | `a47734dbbb44aafcea3d44f17a35d8a422eee7e1d47a86bdcdf88a6670fc26dd` |
| `write` | `899144a751948093832b176f84d3d710d3c434370b551c94934336158ac3ada7` |

### Updated existing test fixtures
`src/improve/scan/detectors/subagent-read-denial.test.ts` — replaced the hand-built `denial()` helper with the real `buildForkDenialRemedy` and prefix, so producer and consumer tests can never drift silently.

## Test Results

- Stability tests: 5/5 pass
- Detector tests: 32/32 pass
- `pnpm lint` clean

Closes #853

### DIFF
--- a/src/agent/tools/hooks/fork-denial-remedy.ts
+++ b/src/agent/tools/hooks/fork-denial-remedy.ts
@@ -4,6 +4,8 @@
  * two downstream consumers that key on it live in one documented place.
  */
 
+import { SUBAGENT_PATH_DENIAL_REASON_PREFIX } from '../denial-circuit-breaker.js';
+
 /** Which containment check produced the denial. */
 export type ForkDenialMode = 'read' | 'write';
 
@@ -20,17 +22,15 @@ export type ForkDenialMode = 'read' | 'write';
  * and sends the fork into retry-until-timeout — the exact failure #544 was
  * opened to stop.
  *
- * Contract: the CALLER prefixes this body with the byte-stable
- * `Sub-agent path access denied:` string — see `SUBAGENT_PATH_DENIAL_REASON_PREFIX`
- * in `../denial-circuit-breaker.ts`, matched by substring, not by prefix
- * position. The remedy body returned here is NOT byte-stable, but it IS
+ * Contract: `buildForkPathDenialReason` prefixes this body with the byte-stable
+ * `SUBAGENT_PATH_DENIAL_REASON_PREFIX` from `../denial-circuit-breaker.ts`.
+ * The remedy body returned here is NOT byte-stable, but it IS
  * fingerprinted: `src/improve/scan/detectors/subagent-read-denial.ts` hashes the
  * ENTIRE normalized reason (`normalizeReason` collapses path-shaped tokens to
  * `<path>`, then `computeFingerprint` SHA-256s it), so any reword here rotates
  * the failure-card slug and restarts that card's severity ladder. That rotation
- * is silent — every detector test builds its own fixture string rather than
- * importing this module — so a reword must be a deliberate decision, not a
- * drive-by edit.
+ * is guarded by a golden test that imports the complete reason producer, so a
+ * reword must be a deliberate decision, not a drive-by edit.
  */
 export function buildForkDenialRemedy(args: { mode: ForkDenialMode; resolvedPath: string }): string {
   const { mode, resolvedPath } = args;
@@ -51,5 +51,18 @@ export function buildForkDenialRemedy(args: { mode: ForkDenialMode; resolvedPath
     `read the path itself and pass the content to you in the prompt. A grant made ` +
     `after you were dispatched cannot reach you — your roots were fixed at dispatch. ` +
     `Return this exact path requirement to your parent.`
+  );
+}
+
+/** Build the complete path-containment denial emitted for a forked sub-agent. */
+export function buildForkPathDenialReason(args: {
+  mode: ForkDenialMode;
+  resolvedPath: string;
+}): string {
+  const { mode, resolvedPath } = args;
+  const remedy = buildForkDenialRemedy(args);
+  return (
+    `${SUBAGENT_PATH_DENIAL_REASON_PREFIX} ${resolvedPath} is outside the ` +
+    `session's granted ${mode} roots. ${remedy}`
   );
 }
--- a/src/agent/tools/hooks/path-approval-hook.ts
+++ b/src/agent/tools/hooks/path-approval-hook.ts
@@ -60,7 +60,7 @@ import type { GrantManager } from '../../../cli/slash/commands/allow-dir.js';
 import { wouldBeRestricted, realpathSafe } from '../handlers/_cwd-utils.js';
 import { isReadDenied } from '../handlers/read-denylist.js';
 import { appendGrant } from '../../permissions-store.js';
-import { buildForkDenialRemedy } from './fork-denial-remedy.js';
+import { buildForkPathDenialReason } from './fork-denial-remedy.js';
 import type { HookContext, HookDecision, HookHandler } from '../../hooks.js';
 
 /** Tools subject to per-call path approval. Bash is gated separately. */
@@ -290,7 +290,6 @@ async function preToolUseImpl(
     // the PARENT — and because a fork's roots are fixed at dispatch, the only
     // remedy that reaches an in-flight fork is a re-dispatch. Wording + the
     // downstream byte/fingerprint contracts live in `./fork-denial-remedy.ts`.
-    const remedy = buildForkDenialRemedy({ mode, resolvedPath: result.resolved });
     // Contract: the "Sub-agent path access denied:" prefix is load-bearing —
     // the `subagent-read-denial` telemetry detector (improve/scan/detectors)
     // and the denial circuit breaker (tools/denial-circuit-breaker.ts,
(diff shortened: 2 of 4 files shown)


griffinwork40:
@codex fix comments

griffinwork40:
Now I have full clarity on finding #9. The `computeFingerprint` hashes the tuple `"PreToolUse|<normalized_reason>"` — the golden value is a SHA-256 of a deterministic string that can be reproduced from the source code alone. The "bootstrap problem" here is that the golden constant was computed from the implementation itself — but that's the **entire point** of a golden test: it locks the current output so *future* changes are caught. The test's purpose is to detect *drift*, not to verify initial correctness (initial correctness is verified by the existing `fork-denial-remedy.test.ts` tests). This is a standard and accepted trade-off in golden testing — downgrading finding #9.

Now synthesizing Wave 2:

---

## PR #1047 Review — `test(improve): golden-fingerprint test for denial reason slug stability (#853)`

**Ref:** `beac0cf68ebafc2f2ec44f5a03da09006a805045`
**Regime:** light (174 lines, 4 files)
**Stated intent:** PR title+body — spec-compliance assessed

---

### Findings

**1. `low · blocking:false · confidence:high · spec-compliance · src/improve/scan/detectors/index.test.ts:152–153 · ref:beac0cf6 · citation-type:file-state`**
Two hardcoded denial strings in `index.test.ts` remain hand-built and truncated (no remedy body), falling outside the PR's scope but contradicting the goal of eliminating fixture drift.
```typescript
lines.push(readDenialLine("Sub-agent path access denied: /a/b.ts is outside the session's granted read roots", 'read_file'));
lines.push(readDenialLine("Sub-agent path access denied: /c/d.ts is outside the session's granted read roots", 'grep'));
```
**suggestion:** Import `buildForkPathDenialReason` in `index.test.ts` and replace the literals in a follow-up.

---

**2. `low · blocking:false · confidence:high · spec-compliance · src/agent/providers/openai-compatible/query.denial-breaker.test.ts:192 · ref:beac0cf6 · citation-type:file-state`**
Hand-built denial reason with a trailing `.` after "roots" that production never emits — pre-existing, not introduced by this PR.
```typescript
reason: `Sub-agent path access denied: ${filePath} is outside the session's granted read roots.`,
```
**suggestion:** Adopt `buildForkPathDenialReason` here in a follow-up.

---

**3. `low · blocking:false · confidence:high · spec-compliance · src/agent/tools/denial-circuit-breaker.ts:80–82 · ref:beac0cf6 · citation-type:file-state`**
JSDoc says "path-approval-hook.test.ts locks the producer at **three sites**." After this PR, a fourth lock (the golden stability test) exists. The comment becomes stale.
```typescript
 * already a load-bearing contract elsewhere — the `subagent-read-denial`
 * telemetry detector keys on this same prefix, and path-approval-hook.test.ts
 * locks the producer at three sites — so a drift in the reason text fails those
```
**suggestion:** Update to "path-approval-hook.test.ts (three assertions) and the golden stability test lock the producer."

---

**4. `nit · blocking:false · confidence:high · test-coverage · src/improve/scan/detectors/subagent-read-denial-stability.test.ts · ref:beac0cf6 · citation-type:diff-context`**
The stability test covers only `mode: 'read'`. No `GOLDEN_FINGERPRINT_WRITE`. This is in-spec: the `subagent-read-denial` detector explicitly ignores write denials (`READ_TOOL_NAMES` excludes write tools at `subagent-read-denial.ts:51–61`), so write fingerprint rotation has no consumer. Noted for documentation.
**suggestion:** Add a one-line comment noting write mode is deliberately excluded because the detector ignores write denials.

---

### What was checked
- All four changed files at the reviewed ref (`beac0cf6`), full text pre-fetched
- All `Sub-agent path access denied` occurrences across the codebase at the reviewed ref (13 sites across 7 files; 3 test files not updated: `index.test.ts`, `query.denial-breaker.test.ts`, `denial-circuit-breaker.test.ts`)
- `SUBAGENT_PATH_DENIAL_REASON_PREFIX` declaration and all consumers (substring `.includes()` match — compatible with the new producer)
- `normalizeReason` regex against the new reason format — path tokens in both unquoted and JSON-quoted positions are correctly normalized
- `computeFingerprint` logic — SHA-256 of `hookEvent|normalizedReason` tuple, deterministic
- `fork-denial-remedy.test.ts` — tests `buildForkDenialRemedy` (inner function); `buildForkPathDenialReason` tested only via the golden hash (acceptable: assembly is trivial concatenation)
- `path-approval-hook.ts` auto-deny block — correctly delegates to `buildForkPathDenialReason`
- Export status of `computeFingerprint` and `normalizeReason` — both exported
- Citations verified inline against branch HEAD `beac0cf68ebafc2f2ec44f5a03da09006a805045`

### What was not checked
- Runtime execution of the golden hash (static analysis confirms the derivation chain is correct; the golden value itself requires running the TypeScript to regenerate)
- Integration tests exercising the full `path-approval-hook → normalizeReason → computeFingerprint` pipeline end-to-end
- `path-approval-hook.ts` beyond the auto-deny block (not changed by this PR)
- Stated intent: PR title+body — spec-compliance assessed

### Epistemic scope
- Stated intent: PR #1047 title+body — spec-compliance assessed
- All four citation types verified at the reviewed ref (file-state citations confirmed via `git show` and `git grep`)
- No absence claims required independent verification (the Wave 1 agent's absence observations were confirmed by `git grep`)

---

**Decision: MERGE — 0 blocking (3 low, 1 nit).**

The change is clean and well-scoped. `buildForkPathDenialReason` correctly centralizes the reason assembly that was previously inlined in `path-approval-hook.ts`. The golden-fingerprint test directly addresses #853's stated failure mode (silent slug rotation). The three `low` findings and one `nit` are pre-existing or informational gaps that do not warrant blocking — they are follow-up candidates at most. No security, correctness, or API-compatibility risks found.

---

PR #1085, opened 2026-08-14
feat(schedule): add enable flag to cancel_schedule for re-enabling tasks

griffinwork40:
## Problem

Re-enabling a disabled scheduled task required a painful workaround: pull the full command text from `get_schedule_history`, then `create_schedule` a new task — which minted a duplicate `-2` suffixed ID, left the old disabled task in the store, and lost the original task's identity.

The `cancel_schedule` description even says *"sets enabled: false so the task can be re-enabled later"* but there was no tool to do the re-enabling.

## Solution

Add `enable?: boolean` as an optional field on `cancel_schedule`. Three-way branch ordering:

| Input | Behavior |
|---|---|
| `{ taskId, permanent: true }` | Remove from store entirely (unchanged) |
| `{ taskId, enable: true }` | **NEW** — flip `enabled: true`, POST to daemon using stored config via `toScheduledTask()` |
| `{ taskId }` (default) | Set `enabled: false` (unchanged) |

`permanent` takes precedence over `enable` when both are supplied.

## Changes

- **`src/agent/tools/handlers/schedules.ts`** — import `toScheduledTask`, add `enable` branch that reads stored config and POSTs to daemon
- **`src/agent/tools/schemas.ts`** — add `enable` property to `cancel_schedule` schema, update description (net -6 lines — under baseline)
- **`src/agent/tools/handlers/schedules.test.ts`** — two new tests: re-enable round-trip, permanent-takes-precedence guard
- **`.filesize-baseline.json`** — regenerated (schemas.ts ratcheted down)

## Testing

- `pnpm test src/agent/tools/handlers/schedules.test.ts` — 22/22 pass
- `pnpm lint` — clean
- `pnpm audit:filesize:check` — clean (schemas.ts 1345→1339, under baseline)

### DIFF
--- a/.filesize-baseline.json
+++ b/.filesize-baseline.json
@@ -48,7 +48,7 @@
     "src/agent/tools/hooks/path-approval-hook.ts": { "loc": 652, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/tools/nesting.ts": { "loc": 520, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/tools/readonly-bash.ts": { "loc": 763, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/tools/schemas.ts": { "loc": 1345, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
+    "src/agent/tools/schemas.ts": { "loc": 1339, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/tools/skill-bridge.ts": { "loc": 498, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/tools/skill-executor.ts": { "loc": 456, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/tools/skill-executor/fork-dispatch.ts": { "loc": 393, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
--- a/src/agent/tools/handlers/schedules.test.ts
+++ b/src/agent/tools/handlers/schedules.test.ts
@@ -200,6 +200,62 @@ describe('cancel_schedule handler', () => {
     }>;
     expect(list[0]?.enabled).toBe(false);
   });
+
+  it('enable: true → re-enables a disabled task', async () => {
+    await createScheduleHandler(
+      { name: 'Toggle Me', command: '/toggle', cron: '0 6 * * *' },
+      fakeSignal,
+    );
+    // Disable first
+    await cancelScheduleHandler({ taskId: 'toggle-me' }, fakeSignal);
+    const afterDisable = await listSchedulesHandler(null, fakeSignal);
+    const disabled = JSON.parse(afterDisable.content as string) as Array<{
+      id: string;
+      enabled: boolean;
+    }>;
+    expect(disabled[0]?.enabled).toBe(false);
+
+    // Re-enable
+    const result = await cancelScheduleHandler(
+      { taskId: 'toggle-me', enable: true },
+      fakeSignal,
+    );
+    expect(result.isError).toBeUndefined();
+    const parsed = JSON.parse(result.content as string) as {
+      ok: boolean;
+      enabled: boolean;
+    };
+    expect(parsed.ok).toBe(true);
+    expect(parsed.enabled).toBe(true);
+
+    // Verify via list
+    const afterEnable = await listSchedulesHandler(null, fakeSignal);
+    const enabled = JSON.parse(afterEnable.content as string) as Array<{
+      id: string;
+      enabled: boolean;
+    }>;
+    expect(enabled[0]?.enabled).toBe(true);
+  });
+
+  it('permanent: true takes precedence over enable: true', async () => {
+    await createScheduleHandler(
+      { name: 'Conflict', command: '/x', cron: '* * * * *' },
+      fakeSignal,
+    );
+    const result = await cancelScheduleHandler(
+      { taskId: 'conflict', permanent: true, enable: true },
+      fakeSignal,
+    );
+    expect(result.isError).toBeUndefined();
+    const parsed = JSON.parse(result.content as string) as { ok: boolean; permanent: boolean };
+    expect(parsed.ok).toBe(true);
+    expect(parsed.permanent).toBe(true);
+
+    // Should be removed from store
+    const listResult = await listSchedulesHandler(null, fakeSignal);
+    const list = JSON.parse(listResult.content as string) as unknown[];
+    expect(list).toHaveLength(0);
+  });
 });
 
 describe('get_schedule_history handler', () => {
--- a/src/agent/tools/handlers/schedules.test.ts
+++ b/src/agent/tools/handlers/schedules.test.ts
@@ -448,4 +448,33 @@ describe('live-sync surface (daemonSynced)', () => {
     expect(parsed.daemonSynced).toBe(false);
     expect(parsed.syncNote).toMatch(/next daemon/i);
   });
+
+  it('enable:true against a live daemon re-registers and reports synced', async () => {
+    const handle = await startDaemon({ port: 0 });
+    try {
+      // Create a disabled task (won't be registered with daemon)
+      await createScheduleHandler(
+        { name: 'Revive Me', command: '/x', cron: '59 23 31 12 *', enabled: false },
+        fakeSignal,
+      );
+      // Disable it via cancel (soft disable, not permanent)
+      await cancelScheduleHandler({ taskId: 'revive-me' }, fakeSignal);
+      // Re-enable via enable:true
+      const enable = await cancelScheduleHandler(
+        { taskId: 'revive-me', enable: true },
+        fakeSignal,
+      );
+      const parsed = JSON.parse(enable.content as string) as SyncShape;
+      expect(parsed.ok).toBe(true);
+      expect(parsed.daemonSynced).toBe(true);
+      expect(parsed.syncDetail).toBe('synced');
+      // Verify the daemon actually has the task registered
+      const tasks = (await (await fetch(`http://localhost:${handle.port}/tasks`)).json()) as Array<{
+        taskId: string;
+      }>;
+      expect(tasks.some((t) => t.taskId === 'revive-me')).toBe(true);
+    } finally {
+      await handle.stop();
+    }
+  });
 });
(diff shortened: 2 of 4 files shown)


griffinwork40:
F1 is fully verified. `saveSchedules` does an atomic `writeFileSync` + `renameSync` (line 90–91), then `getSchedule` reads from `loadSchedules` which reads the same file. The `undefined` fallback at line 203 sends `DELETE` to the daemon for a task the store just saved as `enabled: true`. The failure mode is real under filesystem error conditions: if `renameSync` succeeds but `loadSchedules` then fails to read (e.g. permissions changed, disk full preventing read buffer allocation), the DELETE fires. The CLI's `schedule enable` at line 141 uses the pre-existing `config` object directly — no re-read — which is the correct pattern.

---

## PR #1085 — Review: `feat(schedule): add enable flag to cancel_schedule`

**Target:** PR #1085 (`feat/schedule-enable` → `main`)
**Reviewed ref:** `9b674757a7f9887920ba6657bf5c0a243e57d895`
**Change type:** feature · 4 files · +77/-13 (90 lines)
**Regime:** light
**Stated intent:** Add `enable?: boolean` to `cancel_schedule` so disabled tasks can be re-enabled without the create-duplicate workaround.

---

### Findings

**F1 · high · blocking:true · confidence:high · correctness**
`src/agent/tools/handlers/schedules.ts:201–204` · ref:9b674757 · diff-context

The `enable` branch saves `enabled: true` to the store, then re-reads via `getSchedule(taskId)`. If `getSchedule` returns `undefined` (file-system failure between the atomic rename and the re-read), the fallback fires `trySyncToDaemon('DELETE', ...)` — unregistering the task from the daemon while the store shows it as enabled. Store and daemon end up in contradictory states; the task silently stops running until daemon restart.

```typescript
saveSchedules(updated);
const refreshed = getSchedule(taskId);
sync = refreshed
  ? await trySyncToDaemon('POST', '/tasks', toScheduledTask(refreshed))
  : await trySyncToDaemon('DELETE', `/tasks/${taskId}`);
```

The CLI `schedule enable` (lines 127–153) uses the pre-existing `config` object directly — no re-read, no destructive fallback — confirming the correct pattern.

**Suggestion:** Drop the `getSchedule` round-trip; use `existing` directly:
```typescript
saveSchedules(updated);
sync = await trySyncToDaemon('POST', '/tasks', toScheduledTask({ ...existing, enabled: true }));
```

---

**F2 · low · blocking:false · confidence:high · perf-observability**
`src/agent/tools/handlers/schedules.ts:196–201` · ref:9b674757 · diff-context

The `enable` branch does three `loadSchedules()` disk reads: once at line 185 (`existing = getSchedule()`), once at line 196 (`const schedules = loadSchedules()`), once at line 201 (`getSchedule()` again). The `existing` already holds the full `ScheduledTaskConfig`. The F1 fix eliminates this entirely.

---

**F3 · low · blocking:false · confidence:high · test-coverage**
`src/agent/tools/handlers/schedules.test.ts:200–261` · ref:9b674757 · diff-context

The two new tests run against the handler without a live daemon (no `startDaemon()` call). The existing `live-sync surface (daemonSynced)` describe block covers create and cancel with a real daemon. The `enable: true` path's POST to the daemon is not covered in that integration suite. Grep confirmed: zero test-file matches for `enable` in the daemon integration block.

**Suggestion:** Add a live-daemon test inside the `live-sync surface` block: create → disable → `{ enable: true }` → assert `GET /tasks` includes the task and `daemonSynced: true`.

---

**F4 · low · blocking:false · confidence:high · correctness**
`src/agent/tools/handlers/schedules.ts:194–208` · ref:9b674757 · diff-context

`enable: true` on an already-enabled task is a silent no-op that re-saves identical data (bumping `updatedAt`) and fires a superfluous POST (409 → `already-registered`). No data corruption, but the spurious `updatedAt` bump is misleading.

**Suggestion:** Guard with `if (existing.enabled) return { ok: true, taskId, enabled: true, daemonSynced: true, syncDetail: 'already-enabled' };`.

---

**F5 · nit · blocking:false · confidence:high · api-compat**
`src/agent/tools/handlers/schedules.ts:217–221` · ref:9b674757 · diff-context

When `enable: true`, the response includes `permanent: false`, which is technically correct (it was not deleted) but semantically confusing. No production code reads `response.permanent` (verified via grep). Cosmetic only.

---

- **Security** — no issues found. The `enable` path reads/writes only the local schedules JSON store and POSTs to the local daemon HTTP API. No auth surface, no secret exposure, no injection vector.
- **Spec-compliance** — stated intent fully satisfied. The three-way branch ordering matches the PR description. `permanent` takes precedence over `enable` as specified and tested. No unmet intent, no scope creep.

### What was not checked

- Citations verified inline against branch HEAD `9b674757`.
- Stated intent: PR #1085 title+body — spec-compliance assessed.
- Did not run tests or verify type-checking (`pnpm lint`).
- Did not review daemon-side HTTP handler for POST idempotency (assumed correct per existing tests at lines 303–319 of the test file).

---

**Decision: DO NOT MERGE — 1 high blocking; 3 low, 1 nit.**

F1 is a one-line fix (drop the `getSchedule` round-trip, use `existing` directly) that eliminates a real store/daemon desync under filesystem failure. The rest of the PR is clean and well-tested.

**Done**

- Reviewed PR #1085 (`feat/schedule-enable → main`) at ref `9b674757`.
- 1 blocking finding (F1: `getSchedule` re-read fallback silently DELETEs after store save), 3 low non-blocking, 1 nit.
- Verdict: **DO NOT MERGE** until F1 is addressed. Fix is one line — replace the re-read with `toScheduledTask({ ...existing, enabled: true })`.
- Review is read-only. No files modified, no comments posted.

---

griffinwork40:
The second commit (`fix(schedule): address review feedback`) explicitly removed the guard. This was an intentional simplification during review feedback, not an accidental deletion. The `disable` branch also does not use this guard. The race window is sub-millisecond and the consequence is bounded (daemon restart clears it). I'll downgrade this to low.

Now let me synthesize the final report.

---

## Review: PR #1085 — `feat(schedule): add enable flag to cancel_schedule for re-enabling tasks`

**Target:** PR #1085 `feat/schedule-enable → main`  
**Reviewed ref:** `15239a1e662d8e9e4686d306fab9385db105951c`  
**Change type:** feature · 4 files · +105 / −15 (120 lines) · regime: light  
**Stated intent:** Add `enable?: boolean` to `cancel_schedule` to re-enable disabled tasks without the create-duplicate workaround. Three-way branch: `permanent > enable > default-disable`.

---

### Findings

**1. `low` · blocking: false · confidence: high · correctness · `src/agent/tools/handlers/schedules.ts:194–201` · ref: `15239a1e` · diff-context**

The enable branch unconditionally POSTs `toScheduledTask({ ...existing, enabled: true })` to the daemon after `saveSchedules()`. The initial commit had a post-save `getSchedule()` guard that fell back to DELETE if the task was concurrently removed between save and sync; the review-feedback commit (`15239a1e`) intentionally removed it. The disable branch also has no such guard (its DELETE on a 404 is already treated as synced). The race window is sub-millisecond and consequence is bounded (a daemon task entry with no backing store, cleared on restart). Consistent with the disable branch's design, but worth noting.

```typescript
// current: unconditional POST
sync = await trySyncToDaemon('POST', '/tasks', toScheduledTask({ ...existing, enabled: true }));
```

**suggestion:** Acceptable as-is; if robustness is desired later, restore the `refreshed = getSchedule(taskId)` guard for both enable and disable branches uniformly.

---

**2. `low` · blocking: false · confidence: high · correctness · `src/agent/tools/handlers/schedules.ts:201` · ref: `15239a1e` · file-state**

`toScheduledTask({ ...existing, enabled: true })` spreads `enabled: true` but `toScheduledTask` maps only `id`, `command`, `trigger`, `cron`, `notifyOn`, `notifyChat` — the `enabled` field is silently dropped. The store write on line 200 already persists `enabled: true`, so behavior is correct, but the explicit spread creates a false impression that it influences the daemon payload.

```typescript
// toScheduledTask ignores `enabled`; spread is a no-op
export function toScheduledTask(config: ScheduledTaskConfig): ScheduledTask {
  return { taskId: config.id, command: config.command, trigger: config.trigger ?? 'cron', ... };
}
```

**suggestion:** Use `toScheduledTask(existing)` directly, or add `// toScheduledTask ignores enabled; included for clarity only`.

---

**3. `low` · blocking: false · confidence: high · test-coverage · `src/agent/tools/handlers/schedules.test.ts` · ref: `15239a1e` · grep (0 matches for `already.*enable|idempoten`)**

No test covers calling `enable: true` on a task that is already enabled. The existing test always disables first, then re-enables. The idempotent path is correct by `trySyncToDaemon`'s `already-registered` semantics, but `syncDetail` would return `'already-registered'` instead of `'synced'` — a caller treating that difference as an error has no test pinning the contract.

**suggestion:** Add a unit test: create, do NOT disable, call `cancelScheduleHandler({ taskId, enable: true })`, assert `ok: true` and `enabled: true`.

---

**4. `nit` · blocking: false · confidence: high · api-compat · `src/agent/tools/schemas.ts:700` · ref: `15239a1e` · diff-context**

Tool name `cancel_schedule` now also handles re-enabling. Description updated to "Disable, re-enable, or permanently remove" which is accurate. Mild semantic mismatch — "cancel" does not suggest "enable" to a model reading the schema cold. No breaking change (additive optional field).

**suggestion:** No action required — name change would be a breaking API change. Description update is sufficient.

---

### Dimensions with no issues

- **security** — no issues found — read `schedules.ts:172–220`. No privileged operations; input type-validated; daemon sync is unauthenticated (pre-existing).
- **spec-compliance** — no issues found. Three-way branch ordering (permanent → enable → default disable) matches the stated contract exactly. `permanent` takes precedence as documented. Response shape includes new `enabled` field.
- **perf-observability** — no issues found. Enable branch has identical overhead to the pre-existing disable branch (one `loadSchedules()` + one `saveSchedules()`).

---

### What was not checked

- Citations verified inline against branch HEAD `15239a1e662d8e9e4686d306fab9385db105951c` and parent commit `15239a1e^` (9b674757).
- Stated intent: PR #1085 title + body — spec-compliance assessed.
- Did not verify daemon-side `/tasks` POST handler for edge cases (e.g., re-POST of an already-registered task with different config fields).

---

**Decision: MERGE — 0 blocking (3 low, 1 nit).**

Clean, well-scoped feature. Tests cover the primary round-trip (disable → re-enable), the precedence rule (permanent wins over enable), and the live daemon sync path. The three low-severity findings are all "nice to have" hardening, not blockers.

---

PR #1108, opened 2026-08-17
fix(eval-run): guard against stale detectorVersion replay (#860)

### LINKED ISSUE #860: eval-run: detectorVersion is written but never read — a stale eval-case replays through the bumped detector and reports as current
## Summary

`EvalAssertion.detectorVersion` is written at generation time, rendered into artifacts, and **never read back**. The schema docstring promises a staleness guard that does not exist:

> `detectorVersion` snapshots the detector identity at generation time (e.g. `'repeated-tool-use@v1'`). The runner uses it to pick the right detector when replaying; a version bump on the detector invalidates the eval-case (which the runner detects and reports rather than silently passing).

— `src/improve/schemas.ts:408-412` (`main` @ `5bc00a89`)

The runner does neither of those things. It does not use `detectorVersion` to pick a detector, and it does not detect or report a bump. A `@v1`-labeled eval-case runs against a live `@v2` detector and reports as a normal result.

## Mechanism

Both runner dispatch sites key on `patternId` alone:

- `src/improve/eval-run/runner.ts:159` — `resolveContract(evalCase.assertion.patternId)`
- `src/improve/eval-run/runner.ts:191` — `resolveReplayHandler(evalCase.assertion.patternId)`

`resolveReplayHandler` (`replay.ts:469-471`) and `resolveContract` (`contracts.ts:712`) both match on `patternId` only. The runner holds the entire `evalCase` at both call sites — `assertion.detectorVersion` is in hand and simply never compared.

Write sites: `eval-gen/writer.ts:219,222` (JSON), `:268` (rationale prose), `:387` (rendered `.md` provenance line). Production read sites: none.

```
$ git grep -n detectorVersion -- src/improve/eval-run/
src/improve/eval-run/replay.test.ts:105:      detectorVersion: 'repeated-tool-use@v1',
src/improve/eval-run/replay.test.ts:321:      detectorVersion: 'closure-anomaly@v1',
src/improve/eval-run/runner.test.ts:160:      detectorVersion: `${pattern}@v1`,
```

Test fixtures only.

## This is already live, not hypothetical

Two distinct staleness surfaces, because replay coverage is narrower than contract coverage:

| Surface | Patterns covered | Consequence of a stale label |
|---|---|---|
| Replay (`REPLAY_HANDLERS`, `replay.ts:445-448`) | `repeated-tool-use`, `closure-anomaly` | fixture replays through a **different detector** than the one that generated it |
| Contract-only (`contracts.ts:687-705`) | `subagent-block`, `tool-failure-density` | contract still evaluates; the `.md` renders a false provenance line |

Live detector tags vs. eval-cases on disk (`~/.afk/agent-framework/improve/eval-cases/`):

| Detector (live tag) | Eval-cases on disk | Status |
|---|---|---|
| `tool-failure-density@v2` | 3x `@v2`, **1x `@v1`** (`tool-failure-compose-eval-20260611-df39ff.json`) | **already drifted, today** |
| `repeated-tool-use@v1` | 4x `@v1` | current |
| `closure-anomaly@v2` *(after #847)* | 1x `@v1` (`closure-anomaly-abort-eval-20260611-99c667.json`) | **drifts on #847 merge** |
| `subagent-read-denial@v2` *(after #847)* | none | n/a |
| `subagent-block@v1` | none | n/a |

This corrects the framing in the #847 review, which called that PR "the first to exercise" the guard. It is not — `tool-failure-density` bumped to `@v2` well before #847 and has carried a stale `@v1` case since. What #847 adds is the first case to reach the **replay** surface, which is the more consequential of the two.

## Suggested fix

At `runner.ts:191`, before dispatching, compare `evalCase.assertion.detectorVersion` against the live detector tag and emit a `skipped` check on mismatch instead of running.

The handlers already have this idiom — `replay.ts:229` and `:381` emit `replay skipped (eval-case may be stale or misgenerated)`, and the runner already demotes a non-conclusive replay to `unsupported` (`isReplayNeutralizeCheck`, `replay.ts:464`) precisely so a skipped replay can never read as `pass`. A version-mismatch skip should reuse that path.

This needs the live tag to be resolvable per pattern. Today each tag is a bare string literal inside its detector's result object:

- `closure-anomaly.ts:216`, `repeated-tool-use.ts:303`, `subagent-block.ts:184`, `subagent-read-denial.ts:261`, `tool-failure-density.ts:257`

So the fix likely wants a small exported `DETECTOR_VERSIONS` map (or an exported const per detector that the result object references) to give each detector one source of truth for its own version.

## Acceptance

- An eval-case whose `assertion.detectorVersion` does not match the live detector tag yields a `skipped`/`unsupported` result naming the mismatch — never `pass`.
- A test pins it: construct a case labeled `@v1` against a detector emitting `@v2`; assert the run does not pass and that the reason names both versions.
- Either the existing stale `tool-failure-compose-eval-20260611-df39ff.json` is regenerated, or the new skip is demonstrated against it.

## Provenance

Split out of the review of #847 as out-of-scope. The fix lands entirely in `src/improve/eval-run/` plus a detector-version source of truth — a subsystem #847 does not touch (`git diff 5bc00a89...746c7758 -- src/improve/eval-run/ src/improve/schemas.ts` is empty).

griffinwork40:
## Summary

Closes #860

Adds a detector-version staleness guard to `runEvalCase` so that eval-cases generated under an older detector version can no longer silently replay through a bumped detector and report as current.

## Problem

`assertion.detectorVersion` was written at eval-gen time (e.g. `repeated-tool-use@v1`) but never read at eval-run time. When a detector is bumped to `@v2`, existing eval-cases generated under `@v1` continue to run through `eval-run` and produce `pass` / `fail` verdicts as if the version match was never lost — masking the detector drift entirely.

## Design

**Guard location:** `runEvalCase` (runner.ts), immediately after the fixture-integrity check (step 1.5).

**How it works:**
1. The new `CURRENT_DETECTOR_VERSIONS` map (in `runner.staleness.ts`) records the current live version string for each known `FailurePattern`, sourced from the `detail['detector']` string each detector emits — the same source of truth the eval-gen writer reads.
2. At eval-run time, `checkDetectorVersion` compares `evalCase.assertion.detectorVersion` to the map entry for `evalCase.assertion.patternId`.
3. **Matching versions (non-stale path):** the check returns `pass` and every downstream check, status, and behaviour is completely unchanged.
4. **Mismatched versions (stale path):** a failing `detector-version-current` `EvalCheck` is appended to `run.checks` and a `TriageNote` is added naming both versions and the remediation command (`afk improve eval-gen`). The failing check propagates through `decideStatus` → the top-line `status` becomes `fail`.

**Sibling extraction:** The guard logic and the version map live in `src/improve/eval-run/runner.staleness.ts` to avoid growing the already-grandfathered `runner.ts` beyond its baseline LOC.

## Files changed

| File | Change |
|---|---|
| `src/improve/eval-run/runner.staleness.ts` | New — `CURRENT_DETECTOR_VERSIONS` map, `DETECTOR_VERSION_CHECK` constant, `checkDetectorVersion` function |
| `src/improve/eval-run/runner.ts` | Imports and calls `checkDetectorVersion` in `runEvalCase` (step 1.5); 10 new lines |
| `src/improve/eval-run/runner.test.ts` | 3 new staleness tests; `makeEvalCase` updated to use current live versions |
| `.filesize-baseline.json` | Baseline updated (runner.ts 563→573; pre-existing worktree drifts also absorbed) |

## Tests

3 new tests in `runner.test.ts` under `runEvalCase detector-version staleness guard`:

- **matching version → no change:** `detector-version-current` check is `pass`; no staleness note; status unaffected
- **stale version → fails the run:** check is `fail`; note contains both versions + `eval-gen` remediation; run status is `fail`
- **stale beats passing contract:** even when the guardrail contract passes, a stale detectorVersion forces `fail`

All 68 existing tests continue to pass. `makeEvalCase` was updated to derive `detectorVersion` from `CURRENT_DETECTOR_VERSIONS` rather than always using `@v1`, so patterns at `@v2` (closure-anomaly, tool-failure-density) no longer trigger false staleness in existing tests.

## Risk and unresolved items

**Version map maintenance:** `CURRENT_DETECTOR_VERSIONS` is a static map. It must be updated alongside every detector version bump — the comment block in `runner.staleness.ts` documents this contract with source-line citations. A future improvement could derive the version dynamically by probing the live detector.

**Unknown patterns:** patterns not in the map are silently skipped (fail-open for new detectors, fail-closed for all known ones).

## Gate results

- `pnpm lint`: ✓ (0 errors)
- `pnpm test src/improve/eval-run`: 68/68 ✓
- `pnpm audit:filesize:check`: ✓ (baseline updated)

## PR checklist

- [x] `pnpm lint` passes
- [x] `pnpm test src/improve/eval-run` passes (68/68)
- [x] `pnpm audit:filesize:check` passes (baseline updated)
- [x] Commits signed off (`git commit -s`) — DCO ✓
- [x] Only `src/improve/eval-run/` files modified (+ baseline)
- [x] No new file exceeds 350 lines (`runner.staleness.ts` is 103 lines)
- [x] Long comments carry `// Contract:` prefix per AFK convention


griffinwork40 (commented):

Scope: light-regime review (206 changed lines) of the detector-version staleness guard. Read-only analysis against reviewed ref `8048f267`.

**Verified correct (the core mechanism):**
- `CURRENT_DETECTOR_VERSIONS` is **accurate** — all 5 entries exactly match the `detector: '<pattern>@vN'` literal in each detector source at this ref (`repeated-tool-use@v1`, `closure-anomaly@v2`, `subagent-block@v1`, `tool-failure-density@v2`, `subagent-read-denial@v2`).
- The map is **complete** — `FailurePatternSchema` (`src/improve/schemas.ts:29`) has exactly those 5 live patterns and every one is mapped. Only the documented-future patterns fall through the fail-open path.
- **Mechanism holds** — a failing `detector-version-current` check forces `status='fail'` via `decideStatus` (`runner.ts:287`); the guard sits at step 1.5 after fixture-integrity with `nowIso`/`notes` in scope; `run.checks[0]` remains `fixture-integrity`.
- **Matching path is behavior-preserving** apart from one added passing check (count 7→8, updated in the happy-path test). No production consumer indexes `run.checks` by fixed position past `[0]`.

**Non-blocking findings (all low):**
1. `correctness` — fail-open for an unmapped `FailurePattern` is deliberate, but the `return null` comment (`runner.staleness.ts`) does not distinguish "intentionally unmapped" from "genuinely future", so a future accidental map removal would go silent. Suggest annotating the skip site.
2. `test-coverage` — the unknown-pattern skip branch has no test (all 3 new tests use mapped patterns). It is currently dead for schema-valid inputs, but a future 6th pattern added without a map entry would hit it untested. Suggest a test asserting no `detector-version-current` check is pushed for an unmapped pattern.
3. `spec-compliance` (doc nit) — the PR body says "3 new tests" but the change also updates 1 existing happy-path test (the 7→8 count assertion). Cosmetic; update the description if desired.

**What was not checked / epistemic scope:** Citations verified inline against branch HEAD `8048f267` (version-map accuracy + completeness re-derived from the detector sources; `decideStatus` propagation read directly). Stated intent = PR #1108 title + body — spec-compliance assessed and met. Did not run the test suite locally (CI reports Lint & Build + all test matrices green). Did not review the Telegram/CLI surfaces (out of diff scope).

griffinwork40 (commented):

**Decision: MERGE** — 0 blocking (2 low, 1 nit). Reviewed ref `8048f267`.

### Summary
Adds a detector-version staleness guard to `runEvalCase` (step 1.5, immediately after fixture-integrity). The mechanism was verified end-to-end at the reviewed ref:

- `CURRENT_DETECTOR_VERSIONS` matches all five live detector `detector:` strings exactly, and the line citations in the comment are precise — `repeated-tool-use.ts:303` (`@v1`), `closure-anomaly.ts:218` (`@v2`), `subagent-block.ts:184` (`@v1`), `tool-failure-density.ts:257` (`@v2`), `subagent-read-denial.ts:261` (`@v2`).
- All five `FailurePattern` enum values (`schemas.ts:29-40`) are mapped — no current detector is skipped.
- The writer records `detectorVersion` from the same `detail['detector']` source (`eval-gen/writer.ts:190-221`), so the run-time comparison is like-for-like (`pattern@vN` vs `pattern@vN`).
- `makeCheck({pass:false})` → `status:'fail'` (`contracts.ts:103`) and `decideStatus` (`runner.ts:287`) promotes any failing check to a `fail` run — so the guard forces `fail` on mismatch and is a strict no-op on match, exactly as the intent states.
- All 7 stated-intent requirements verified against the diff; CI is green; scope is limited to `src/improve/eval-run/` + the filesize baseline.

**[low · correctness] `runner.staleness.ts:78`** — Unmapped patterns silently no-op (`if (current === undefined) return null;`). This is documented fail-open design; the residual risk is that a *future* `FailurePattern` added to the enum before its map entry lands would get a permanent silent pass. Suggestion: emit a dev/CI warning when `recorded !== undefined && current === undefined` so a missing map entry surfaces loudly. (Currently unreachable — every enum value is mapped.)

**[low · test-coverage] `runner.staleness.ts:78`** — No test exercises the return-`null` (unmapped-pattern) branch. Suggestion: one direct unit test of `checkDetectorVersion` with a synthetic `patternId` outside the map, asserting `null`. (Low value today: the branch is unreachable via the schema-constrained `patternId`.)

**[nit · spec-compliance] `runner.test.ts:287`** — The corrupt-fixture test asserts every non-`fixture-integrity` check passes; that predicate now also sweeps in the new `detector-version-current` check (harmless — `makeEvalCase` sets a matching version by default). Suggestion: exclude `DETECTOR_VERSION_CHECK` explicitly in the filter and update the "only the fixture is corrupt" comment.

### What was not checked
- Citations verified inline against head SHA `8048f267` (pre-fetched file contents at the ref).
- Stated intent: PR #1108 title + body — spec-compliance assessed (all 7 requirements met).
- `EvalCase.assertion.detectorVersion` is typed `z.string().min(1)` (`schemas.ts:421`); the `pattern@vN` format is enforced by the writer/convention, not the schema.
- No CI/lint guard prevents `CURRENT_DETECTOR_VERSIONS` drifting from the detectors (the PR's own "Version map maintenance" risk); a cross-check test would close it.
- Did not execute the suite (read-only review) — relied on the green CI checks for the 68/68 claim.

PR #1111, opened 2026-08-17
fix(auth): refresh expired OAuth token at startup before sync credential read

griffinwork40:
## Problem

The sync `loadClaudeCodeOauthToken()` in `keychain.ts` bails on an expired token without attempting a refresh — it prints a stderr message and returns `undefined`. All four startup surfaces (daemon, REPL, chat, telegram) resolve credentials via that sync path, so when the Claude Code OAuth token in the macOS Keychain is expired at startup, the process silently gets `undefined` for its API key. The 401 retry tier never fires because there is no key to get a 401 *with*.

This is the root cause of the recurring "Claude Code OAuth token in keychain is expired" daemon error — the token expires every ~24h, and `claude login` is only a band-aid.

## Fix (centralized, guarded, bounded)

> Note: this supersedes the original per-site approach (one `await refreshClaudeCodeOauthToken()` at each of the four command actions). Code review found that placement did **not** fix the daemon — see P1 below — so the refresh was consolidated into one guarded, bounded preload.

A single new helper, `preloadClaudeKeychainOAuth(provider)` (`src/agent/auth/credential-resolver.ts`), refreshes a near-expiry/expired keychain OAuth token and writes it back **before** the synchronous credential read. It is called from exactly two entry points:

- **`runFirstRunDetector()` in `src/cli/index.ts`** — the first-run credential gate, which runs before Commander dispatches any command action. This one call covers all four gated CLI surfaces (chat, interactive, daemon, farm).
- **`main()` in `src/telegram/entry.ts`** — the telegram bot has its own entrypoint.

### Review findings addressed

- **P1 (blocking) — the daemon case was not actually fixed by the original approach.** `runFirstRunDetector()` runs *before* Commander dispatch and, for a non-TTY service whose only credential is an expired keychain token, calls `loadCredential() → undefined → process.exit(1)` before the daemon action's own refresh could run. The refresh now happens inside the detector, before its sync read, so the freshly written-back token is picked up and the daemon proceeds.
- **M1 (blocking) — un-timed network call on the startup critical path.** `postTokenRefresh()` had no `fetch` timeout; now awaited at boot, a stalled/captive network could hang startup up to undici's ~300s default. Added `signal: AbortSignal.timeout(10s)` so a stalled refresh fast-fails to the prior behavior.
- **L1 / P2 — unconditional refresh did dead work for non-Claude users.** The new helper guards on Anthropic-routed provider + no env credential, so it is a no-op for OpenAI/xAI/local users and anyone authenticated via an env var (no `security` subprocess, no spurious "OAuth expired" warning).
- **P2 (telegram) — the fresh token was discarded when write-back failed.** `refreshClaudeCodeOauthToken()` returns a usable token even when the keychain write-back fails; `entry.ts` now threads it as a fallback.
- **N1** — documented the startup-preload contract on `refreshClaudeCodeOauthToken`.

## Scope note — `.filesize-baseline.json`

The baseline carries **only** the 5 pre-existing over-ceiling entries that already exceed their baseline on `main` (`nesting.ts`, `subagent/child-config.ts`, `_lib/stream-renderer.ts`, `telegram/elicitation-handler.ts`, `telegram/session-manager.ts`). **`main`'s filesize gate is independently red** for those 5 files — they are unchanged by this PR; the entries are carried only to keep CI green. A dedicated cleanup PR should bring them back under the ceiling.

## Known low-risk follow-up (non-blocking)

If the startup refresh's network exchange succeeds but the keychain write-back fails (locked/read-only keychain), the CLI/daemon path re-reads the still-stale store and does not use the returned token (only telegram does). This is a rare, environment-rooted edge (a locked keychain also defeats `claude login`); tracked as a follow-up rather than blocking this fix.

## Testing

- `pnpm lint` — clean
- `pnpm test` — 16,772 passed (1 pre-existing environmental flake: `write-denylist.test.ts` symlink-repoint `rmSync` on this machine's `/tmp`; green in CI, byte-identical to `main`)
- `pnpm build`, `pnpm audit:filesize:check`, `pnpm audit:env:check` — pass
- New tests: detector wiring + a non-TTY daemon-recovery regression (`first-run.test.ts`), the preload guard (`credential-resolver.test.ts`), and the write-back-failure fallback seam (`credentials.test.ts`).


griffinwork40:
Wave 1.5 verification complete — all carried-forward claims confirmed against ref `54043865`:
- **F5** (MEMORY_TOOL_NAMES): exported `memory/index.ts:9`, imported `daemon.ts:31`, used `daemon.ts:125` → clean consolidation, *not a defect*.
- **F4 absence** (no startup-wiring test): confirmed — 3 test hits are all unrelated module mocks; `daemon.test.ts`/`entry.test.ts` have 0 refresh references, `chat.test.ts`/`bootstrap.test.ts` don't exist. **confirmed-absent.**
- **Headline** (un-timed fetch): confirmed — `fetch(OAUTH_TOKEN_URL, {...})` at keychain.ts:169 has no `signal`/`timeout`/`AbortSignal`, and the diff places `await refreshClaudeCodeOauthToken()` on the startup path. Coordinator-corroborated by direct read.

No critical/high and no overridden dispositions survive synthesis → shadow-verify not triggered; the one blocking finding is coordinator-verified directly. Here is the review.

---

# Code Review — PR #1111 · `fix(auth): refresh expired OAuth token at startup before sync credential read`

**Target:** PR #1111 (`afk/i-need-more-information` → `main`), reviewed ref `54043865` · +16/−11, 5 files · change type: hotfix → **light regime**. Stated intent: PR title+body — **spec-compliance assessed**.

**The fix is real and well-targeted.** The async→sync handoff is sound: `refreshClaudeCodeOauthToken()` writes the fresh token back to the same store (`writeCredentialsBlob` via blocking `execFileSync`) *before* returning, and the subsequent sync `getApiKey()` re-reads it — no race, correct ordering at all 4 sites. The function never throws (every path returns `undefined` or writes stderr), so `await`-ing it at startup cannot crash the process. The primary user story (Anthropic-only, no env vars, expired keychain token, daemon ~24h expiry) is fixed.

## Findings (sorted; severity measures blast radius, not category)

**M1 · medium · blocking:true · confidence:high · correctness/perf-observability · `src/agent/auth/keychain.ts:169` (+ call sites chat.ts:348, daemon.ts:414, bootstrap.ts:76, telegram/entry.ts:60) · ref:54043865 · citation-type:file-state**
This PR newly places an **un-timed network call on the startup critical path of all 4 surfaces.** `postTokenRefresh` calls `fetch(OAUTH_TOKEN_URL, {...})` with no `signal`/`AbortSignal.timeout`. It fires exactly in the PR's target case (token expired/near-expiry within the 5-min margin); on a degraded/captive network the startup `await` stalls up to undici's default headers/body timeout (~300s) instead of the pre-PR instant sync return. For the unattended daemon this converts "fast fail on expired token" into a multi-minute silent startup hang.
- evidence: `const res = await fetch(OAUTH_TOKEN_URL, { method: 'POST', headers: {...}, body: ... });` — no timeout.
- suggestion: add `signal: AbortSignal.timeout(10_000)` to the `fetch` (or wrap the startup refresh in a bounded timeout), so a stalled refresh degrades to the prior fast-fail path rather than blocking boot.

**L1 · low · blocking:false · confidence:high · perf-observability/UX · keychain.ts:103-121 & 67-70 · ref:54043865 · citation-type:file-state**
The refresh is inserted **unconditionally**, before provider routing and ignoring env-var credentials — but `getApiKey()` only ever reads the keychain for Anthropic-routed models with *no* `ANTHROPIC_API_KEY`/`CLAUDE_CODE_OAUTH_TOKEN` env var. So for OpenAI/xAI/local users, and Anthropic users with an env-var key, every startup now (a) spawns a synchronous `security` subprocess on macOS (dead work), and (b) if a *stale* Claude Code OAuth entry lingers from a past `claude login`, attempts a network refresh and prints `agent-afk: OAuth token expired... Run 'claude login' to refresh.` to stderr — a spurious, confusing warning for someone not using OAuth (mildly counterproductive to the PR's own goal of reducing OAuth-error confusion).
- evidence: `await refreshClaudeCodeOauthToken();` runs before `resolveCredentialForModel`, which for non-Anthropic providers never calls `loadClaudeCodeOauthToken()`.
- suggestion: guard the call — skip when `providerForModel(getModel())` is not Anthropic-routed, or when `env.ANTHROPIC_API_KEY || env.CLAUDE_CODE_OAUTH_TOKEN` is already set. This also resolves M1 for the majority of users.

**L2 · low · blocking:false · confidence:high · test-coverage · chat.ts:348 / daemon.ts:414 / bootstrap.ts:76 / telegram/entry.ts:60 · ref:54043865 · citation-type:diff-context**
No test asserts the refresh-before-read wiring at any startup site. `grep refreshClaudeCodeOauthToken` in tests yields 3 hits (`client-setup.test.ts:22`, `doctor.test.ts:14`, `config.test.ts:48`) — all unrelated module mocks; `daemon.test.ts`/`entry.test.ts` reference it 0 times.
- suggestion: add one assertion per entry point that `refreshClaudeCodeOauthToken` is awaited before credential resolution.

**L3 · low · blocking:false · confidence:high · spec-compliance/hygiene · `.filesize-baseline.json` · ref:54043865 · citation-type:file-state**
The baseline change bumps **6 files, but 5 are byte-identical to `origin/main`** (nesting, child-config, stream-renderer, elicitation-handler 459→483, session-manager 896→917). Verified: HEAD is exactly 1 commit ahead of current main, and those 5 files' actual sizes already exceed their old baselines on main — i.e. **`pnpm audit:filesize:check` is currently red on `main`**, and this PR's `audit:filesize:update` regen silently sweeps in that unrelated correction. The PR body's "+2 line growth in 3 files" understates it.
- suggestion: split the baseline regen into its own commit/PR (and flag that main's filesize gate is red as a separate issue), or at minimum correct the PR body to describe the real 6-file, up-to-+24 change.

**N1 · nit · blocking:false · confidence:high · correctness · keychain.ts:53 · ref:54043865 · citation-type:file-state**
JSDoc describes `refreshClaudeCodeOauthToken` only as the 401-retry helper ("callers should surface the original 401 error"); startup-preload is now its primary caller.
- suggestion: note the startup-preload contract in the doc.

## Spec-compliance
**Satisfied for the primary case, over-broad in scope.** The expired-token-at-startup silent-`undefined` is correctly fixed for the keychain-OAuth user. Scope creep: the unconditional insertion (L1) imposes cost on users who never touch the keychain, and the baseline change (L3) fixes unrelated pre-existing main drift under an auth-hotfix banner.

---

**Decision: DO NOT MERGE — 1 medium blocking (correctness/perf-observability: un-timed network refresh on the startup critical path); 0 medium waived, 3 low, 1 nit.**

The blocking item is a ~1-line fix (add a `fetch` timeout); guarding the refresh on provider/env-var (L1) would additionally close M1 for most users and is the higher-leverage change.

## What was not checked
- **Citations verified inline against branch HEAD `54043865`** (== `origin/main` + 1); baseline claims verified via two-dot diff against current `origin/main`.
- **Stated intent:** PR #1111 title+body — spec-compliance assessed.
- **No shadow-verify dispatched:** no critical/high finding and no overridden dispositions survived synthesis; the one blocking finding was corroborated directly by the coordinator (read keychain.ts:169, no `signal`/timeout; diff confirms startup-path placement).
- Not traced to ground: `createBootstrapInfra`'s internal `apiKey` resolution (trusted per diff ordering), the pre-existing concurrent-daemon keychain read-modify-write race in `refreshClaudeCodeOauthToken` (out of scope), exact undici timeout defaults at runtime, and Windows behavior (confirmed no-op at keychain.ts:133, untested).

**Done**
- Completed a read-only light-regime review of PR #1111: dispatched 1 independent research-agent reviewer, ran inline Wave 1.5 citation + absence verification, synthesized, and emitted a merge recommendation. No files, commits, or PR state were modified.
- Verdict: **DO NOT MERGE** — 1 medium blocking (un-timed `fetch` on the new startup credential path, `keychain.ts:169`), plus 3 low + 1 nit; the core fix itself is sound.
- Evidence for the terminal claims lives in durable locations: diff artifact `/Users/griffinlong/.afk/state/skill-preflight/9e66cb86-6e83-4bdb-aa86-2136259c5ff1/pr-1111.diff`; source `src/agent/auth/keychain.ts` (:55-100 no-throw refresh, :169 un-timed fetch), `src/agent/auth/credential-resolver.ts:39-121` (resolution order), and the two-dot `git diff origin/main HEAD -- .filesize-baseline.json` (main filesize-gate red).
- Nothing deferred; this review is terminal by design (the skill mutates nothing). Acting on any finding (e.g. adding the `fetch` timeout) is a separate, explicitly-authorized task.

---

PR #1116, opened 2026-08-17
feat(tui): slash-complete anywhere, @-file subsequence, Linux image paste, Ctrl-R reverse-search

griffinwork40:
# feat(tui): slash-complete anywhere, @-file subsequence, Linux image paste, Ctrl-R reverse-search

Modernizes the REPL line editor with four additive sub-features in `src/cli/input/`.

---

## 1. Slash-complete anywhere (`trigger.ts`)

**Files:** `src/cli/input/trigger.ts`, `src/cli/input/trigger.test.ts`

`detectTrigger` previously required the slash token to start at buffer position 0 (`/^\/[A-Za-z_-]*$/`). It now uses a lookbehind-aware regex so completion fires on any slash token at the cursor — mid-buffer, after prose, or in multiline drafts (e.g. `"please run /ship"` with cursor at end).

`applySelection` already supported a non-zero start offset, so no changes there.

**Tests added:** 10 new cases in `detectTrigger — slash completion (mid-buffer)`.

---

## 2. @-file subsequence matching (`trigger.ts`, `slash-match.ts`)

**Files:** `src/cli/input/trigger.ts`, `src/cli/input/trigger.test.ts`

`buildFileCandidates` now partitions into two ranked buckets:
- **Prefix matches** (alphabetical) — ranked first
- **Subsequence-only matches** (via `isSubsequence` from `slash-match.ts`) — ranked second

So `@src/inp` matches `src/cli/input/` without displacing prefix hits. Mirrors slash-command ranking.

**Tests added:** 5 new cases in `buildFileCandidates — @-file subsequence matching`.

---

## 3. Linux clipboard image paste (`clipboard-image.ts`, new `clipboard-image-linux.ts`)

**Files:** `src/cli/input/clipboard-image.ts`, `src/cli/input/clipboard-image-linux.ts` (new), `src/cli/input/clipboard-image.test.ts`

`readClipboardImage` now routes to `readClipboardImageLinux` on `process.platform === 'linux'`:
1. Probes for `wl-paste` (Wayland) via `spawn --version`
2. Falls back to `xclip` (X11/XWayland)
3. Graceful no-op if neither tool is installed

macOS path (osascript/sips) unchanged. Linux code extracted to `clipboard-image-linux.ts` to stay under 350-line ceiling.

**Tests added:** 7 new cases in `Linux clipboard (Part B-Linux)`.

---

## 4. Ctrl-R reverse history search (new `reader.reverse-search.ts`)

**Files:** `src/cli/input/reader.reverse-search.ts` (new), `src/cli/input/reader.reverse-search.test.ts` (new), `src/cli/input/reader.state.ts`, `src/cli/input/reader.keypress.ts`, `src/cli/input/reader.ts`

Incremental reverse-search modal:
- **Ctrl-R** activates / cycles to next older match
- **Status line:** `(reverse-i-search)\`query\': <match>`
- **Enter** accepts (keeps match, normal Enter then submits)
- **Esc / Ctrl-G** cancels (restores prior buffer)
- **Ctrl-C** cancels + falls through to abort handler
- **Backspace** trims query, **printable chars** extend it

`ReverseSearchState` is threaded via `ReaderState` — no closure over `let` bindings.

**Tests:** 27 new cases in `reader.reverse-search.test.ts`.

---

## Gate results

```
pnpm lint     clean (tsc --noEmit, no errors)
pnpm test     472 / 472 passing (21 test files in src/cli/input/)
```

| Sub-feature | New tests |
|-------------|-----------|
| Slash-complete anywhere | +10 |
| @-file subsequence | +5 |
| Linux clipboard paste | +7 |
| Ctrl-R reverse search | +27 |
| **Total** | **+49** |

No deferrals — all four sub-features landed green.

### DIFF
--- a/src/cli/input-box.test.ts
+++ b/src/cli/input-box.test.ts
@@ -37,9 +37,11 @@ describe('detectTrigger', () => {
       expect(result).toBeNull();
     });
 
-    it('returns null when buffer has text before /', () => {
+    it('detects a slash command after preceding text + space (mid-buffer)', () => {
+      // Slash-complete-anywhere: a slash token after whitespace triggers so
+      // completion fires in "please run /help"-style prompts, not only at start.
       const result = detectTrigger('hello /help', 11);
-      expect(result).toBeNull();
+      expect(result).toEqual({ kind: 'slash', query: 'help' });
     });
 
     it('allows hyphen and underscore in slash command names', () => {
@@ -47,19 +49,19 @@ describe('detectTrigger', () => {
       expect(result).toEqual({ kind: 'slash', query: 'my-cmd_name' });
     });
 
-    it('returns null when slash command is not at the start', () => {
+    it('detects a slash command that is not at the buffer start (after whitespace)', () => {
       const result = detectTrigger('text /cmd', 9);
-      expect(result).toBeNull();
+      expect(result).toEqual({ kind: 'slash', query: 'cmd' });
     });
 
-    it('returns null when slash command is preceded by leading whitespace', () => {
+    it('detects a slash command preceded by leading whitespace', () => {
       const result = detectTrigger(' /help', 6);
-      expect(result).toBeNull();
+      expect(result).toEqual({ kind: 'slash', query: 'help' });
     });
 
-    it('returns null when slash command appears on a later line', () => {
+    it('detects a slash command on a later line (after a newline)', () => {
       const result = detectTrigger('note\n/help', 10);
-      expect(result).toBeNull();
+      expect(result).toEqual({ kind: 'slash', query: 'help' });
     });
 
     it('returns null when slash command contains punctuation', () => {
--- a/src/cli/input/clipboard-image-linux.ts
+++ b/src/cli/input/clipboard-image-linux.ts
@@ -1,0 +1,162 @@
+/**
+ * Linux clipboard image reader — Wayland (wl-paste) and X11 (xclip) support.
+ *
+ * Extracted from clipboard-image.ts to keep that file under the 350-line
+ * source ceiling. All Linux-specific logic lives here; the macOS path (osascript
+ * / sips) stays in clipboard-image.ts.
+ *
+ * Strategy:
+ *   1. Probe for `wl-paste` (Wayland-native) first.
+ *   2. Fall back to `xclip` (X11 / XWayland) if wl-paste is absent.
+ *   3. If neither tool is installed, return null (graceful no-op — user
+ *      simply doesn't get clipboard-image paste on this system).
+ *
+ * Both tools are asked for `image/png` directly via MIME-type flags:
+ *   wl-paste --type image/png --no-newline
+ *   xclip    -selection clipboard -t image/png -o
+ *
+ * Magic-byte validation is shared with the macOS path via the caller.
+ *
+ * @module cli/input/clipboard-image-linux
+ */
+
+import { spawn } from 'child_process';
+import { randomUUID } from 'crypto';
+import { unlink } from 'fs/promises';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import type { ImageAttachment } from './attachments.js';
+
+/**
+ * When AFK_DEBUG_CLIPBOARD=1, structured diagnostic messages are written to
+ * stderr at each probe stage. Zero overhead when unset. Mirrors the pattern in
+ * clipboard-image.ts — evaluated at module load so re-import is required to
+ * toggle within a test (identical to the macOS module's behaviour).
+ */
+// Using dynamic env read rather than importing env.ts to avoid a circular dep
+// through config → paths. The clipboard module already accesses env this way.
+const DEBUG = !!(process.env['AFK_DEBUG_CLIPBOARD']);
+
+function dbg(msg: string): void {
+  process.stderr.write(`[afk-clipboard] ${msg}\n`);
+}
+
+/**
+ * Check whether a command is available on PATH by attempting a version-style
+ * probe via `spawn`. Returns true if the binary exists and can be exec'd; false
+ * on ENOENT / EACCES. A non-zero exit code still proves the binary is present
+ * (some tools like `wl-paste --version` exit 1 on certain distros).
+ *
+ * Uses `spawn` (not `execFile`) so the same mock used in clipboard-image.test.ts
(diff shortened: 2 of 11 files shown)


griffinwork40 (commented):

## PR #1116 Review — `afk/tui-editor-polish` → `main`

**Branch:** `afk/tui-editor-polish` → `main` · **Ref:** `f424e222` · **Regime:** full (1200+ lines, 11 files, feature)

---

### ✅ MERGE — 0 blocking (3 medium waived, 3 low, 2 nit)

**Decision: MERGE — 0 blocking (3 medium waived, 3 low, 2 nit).**

---

### Waived medium findings

**C1** · medium · blocking:false · confidence:high · correctness · `src/cli/input/clipboard-image-linux.ts:138,158` · ref:f424e222 · file-state · waived: dead code with no data-affecting impact — `tmpPath` is never created, the `unlink` silently fails with a swallowed ENOENT, and no user-facing behavior is altered; follow-up cleanup item.

**`tmpPath` is created but never written to; `unlink(tmpPath)` fires a no-op ENOENT syscall on every Linux paste attempt.** `linuxClipboardRead` collects bytes via the child's stdout pipe, not a temp file. The `tmpPath` variable, `finally { unlink(tmpPath) }` block, and the `unlink`/`tmpdir`/`join` imports are dead weight from an abandoned write-to-temp strategy. No data is leaked or lost; the syscall fails silently.

```typescript
const tmpPath = join(tmpdir(), `afk-clipboard-${randomUUID()}.bin`);
try {
  const buffer = await linuxClipboardRead(cmd, args); // reads stdout, not tmpPath
} finally {
  unlink(tmpPath).catch(() => undefined);  // ENOENT always
}
```

**Fix:** Remove `tmpPath`, the `finally` block, and the three dead imports (`unlink`, `tmpdir`, `join`). Retain `randomUUID` (used for `id` on line 150).

---

**C2** · medium · blocking:false · confidence:high · correctness · `src/cli/input/clipboard-image-linux.ts:56-66` · ref:f424e222 · file-state · waived: bounded, non-data-affecting — if the event ordering ever reverses, `isCommandAvailable` returns a false positive, but the subsequent `linuxClipboardRead` returns `null` (exit 1), so the probe degrades gracefully to no-image rather than producing wrong output.

**`isCommandAvailable` relies on Node.js event ordering (`error` before `close`) that is not documented as guaranteed.** Both handlers call `resolve()` independently. Today `error` fires first (ENOENT → `resolve(false)`), making the second `resolve(true)` from `close` a no-op. If the ordering reversed, a missing tool would be reported as available — but the follow-up data read would fail and return `null`, so no incorrect image is surfaced.

```typescript
child.on('error', (err) => {
  const code = (err as NodeJS.ErrnoException).code;
  resolve(code !== 'ENOENT' && code !== 'EACCES');
});
child.on('close', () => resolve(true));
```

**Fix:** Add a `settled` guard flag so the first handler to fire wins regardless of event ordering.

---

**C3** · medium · blocking:false · confidence:high · correctness · `src/cli/input/reader.keypress.ts:76,94-113` · ref:f424e222 · file-state · waived: bounded, non-data-affecting — Ctrl-C during reverse-search aborts the reader and discards the entire `ReaderState`; the leaked `active=true` flag has no observable effect because the state object is never reused.

**Ctrl-C while reverse-search is active skips `cancelReverseSearch` — the `active` flag is left `true` and `savedBuffer` is not restored.** The Ctrl-C handler at line 76 fires before the `st.reverseSearch.active` guard at line 110, so `handleReverseSearchKey`'s Ctrl-C branch (which calls `cancelReverseSearch`) is unreachable from the production dispatcher. Since `callbacks.onAbort` rejects the promise and the `ReaderState` is discarded (single-use per `readWithAutocompleteTty` call), the state leak is harmless.

```typescript
// Line 76 — fires first, returns before line 110
if (key?.ctrl && key?.name === 'c') {
  callbacks.onAbort(new Error('SIGINT'));
  return;  // st.reverseSearch.active left true
}
```

**Fix:** Either add `if (st.reverseSearch.active) cancelReverseSearch(st.reverseSearch, st)` before the abort call at line 76, or document that `ReaderState` is single-use and cleanup is unnecessary on abort.

---

### Non-blocking findings

| # | Sev | Dimension | Location | Finding |
|---|-----|-----------|----------|---------|
| C4 | low | correctness | `clipboard-image-linux.ts:96-116` | `detectMediaTypeLinux` duplicates the macOS `detectMediaType` byte-for-byte. Future MIME-type additions will silently diverge between platforms. Extract to a shared module. |
| C5 | low | perf | `trigger.ts:218-226` | `buildFileCandidates` runs `isSubsequence` on every non-prefix entry with no early exit once `MAX_FILE_MATCHES` prefix hits fill the bucket. Harmless at typical directory sizes. |
| C6 | low | test-coverage | `clipboard-image.test.ts` | No test where both `wl-paste` and `xclip` are available but wl-paste returns unrecognized magic bytes, causing fallthrough to xclip. The probe loop handles this correctly; coverage gap only. |
| N1 | nit | correctness | `reader.reverse-search.ts:228` | `promptVisibleLen = line.length` uses codepoint count, not display-column width. If the search line contains wide/combining chars, cursor offset will be wrong. Use `stringWidth(line)`. |
| N2 | nit | correctness | `reader.reverse-search.ts:222-231` | `repaintReverseSearch` mutates `RepaintCtx` via type-assertion casts. The `try/finally` restore is correct, but violates the ctx's implicit immutability contract. Pass a prompt override parameter instead. |

---

### Spec-compliance

**Satisfied.** All four stated sub-features are implemented and tested:

| Sub-feature | Tests | Status |
|-------------|-------|--------|
| Slash-complete anywhere | +14 | ✅ Regex correctly matches mid-buffer after whitespace |
| @-file subsequence | +5 | ✅ Prefix-first, subsequence-second ranking |
| Linux clipboard paste | +7 | ✅ wl-paste/xclip probe; dead tmpPath code is a cleanup item (C1) |
| Ctrl-R reverse-search | +27 | ✅ Modal activate/cancel/accept/cycle all work |

No scope creep detected.

### What was not checked

- Citations verified inline against branch HEAD `f424e222`.
- Stated intent: PR #1116 title+body — spec-compliance assessed and met.
- Runtime behavior of the reverse-search modal in a live terminal (structural analysis only).
- Performance of subsequence matching under large directories (flagged structurally, not measured).
- No shadow-verify dispatched: no critical/high findings survived; three waived mediums are bounded and non-data-affecting.

PR #1126, opened 2026-08-17
fix(subagent): surface partial tool results on empty-buffer stream cut-off (#724)

### LINKED ISSUE #724: Stream cut off with an empty text buffer discards the whole subagent run (12-minute investigation returned nothing)
## Symptom

A `research-agent` subagent ran 12m12s, completed 5 `read_file` calls and ~1197 lines of gathered context, then died with:

```
the model stream ended without a terminal message (no message_stop and no stop_reason):
the response was cut off mid-stream
```

The parent session received nothing usable and re-dispatched a near-identical subagent from scratch. Twelve minutes of paid-for tool results were discarded because the summary that would have carried them never streamed.

## Mechanism (verified)

- `src/agent/providers/anthropic-direct/translate.ts:319-332` raises `StreamIncompleteError` when the Anthropic SSE iterator ends cleanly with **neither** a `message_stop` **nor** a `stop_reason`. The co-located comment (`translate.ts:306-318`) attributes this to an intermediary closing the connection mid-response, and explicitly excludes `max_tokens` truncation — that sets a real `stop_reason` and bypasses the guard by design. Mirror on the other provider: `src/agent/providers/openai-compatible/query/stream-drive.ts`.
- `src/agent/providers/anthropic-direct/loop.ts` already retries this condition, bounded (`STREAM_INCOMPLETE_MAX_RETRIES`) with backoff, with the budget reset per tool-use round. So a single transient blip per round already survives.
- Salvage exists but is **conditional**: `src/agent/subagent/handle.ts:473` returns the buffered text as a partial when `lastStreamedContent.length > 0`. When the cut lands before any assistant text has streamed, the empty-buffer path throws instead and the run is classified `failed`.

## The actual defect

The loss is not the truncation — that is a transport-level condition and largely upstream. The defect is that a 12-minute subagent run holding substantial **already-gathered tool output** has no durable checkpoint, so an empty text buffer turns a recoverable interruption into a total loss.

Every `read_file` result had been fetched and paid for. Only the un-streamed summary was lost, and it took all the context with it.

## Proposed fix

On an empty-buffer `StreamIncompleteError`, surface something the parent can act on instead of a bare failure:

- attach the run's accumulated tool-result context (or a pointer to it in the witness trace at `~/.afk/state/witness/<session>/trace.jsonl`) as `partialOutput`; and/or
- have long-running subagents checkpoint gathered findings to a file at intervals, so the parent can resume from disk rather than re-dispatch.

Not asking to prevent the drop. Asking for the failure to be **cheap instead of total**.

## Acceptance

- [ ] An empty-buffer stream-incomplete on a subagent that already produced tool results returns partial context or a trace pointer, not a bare failure.
- [ ] The parent can distinguish "died with nothing gathered" from "died with N tool results gathered".
- [ ] Test covers the empty-buffer path with prior tool results present.

griffinwork40:
## Problem

When a subagent stream is cut off with an empty text buffer (`StreamIncompleteError`), ALL accumulated tool results from that run are discarded. A 12-minute run with 5 `read_file` calls returns nothing to the parent — the run classifies `failed` with no recoverable evidence.

Closes #724.

## Root Cause

`src/agent/subagent/handle.ts` line ~619: the empty-buffer path throws `StreamIncompleteError` unconditionally. `runToResult`'s catch block only sets `partialOutput` from `lastStreamedContent`, which is empty in this case — so tool evidence in `currentTrace.toolResults` is never surfaced.

### `src/utils/errors.ts`
Added `toolResultsGathered?: number` to `StreamIncompleteError` so callers can distinguish "died with N gathered results" from "died with nothing".

### `src/agent/subagent/empty-buffer-partial.ts` _(new)_
Two helpers extracted into a sibling file to keep `handle.ts` under the GREW gate:
- `buildEmptyBufferError(id, toolResults)` — factory that populates `toolResultsGathered` and tailors the error message to the zero/non-zero tool-results case.
- `synthesizeEmptyBufferPartial(id, toolResults)` — returns a human-readable `partialOutput` summary (count, total bytes, error count) when tool results exist; `undefined` otherwise.

### `src/agent/subagent/handle.ts`
- Replaced the inline 6-line throw block with `throw buildEmptyBufferError(...)` (net −1 LOC; clears the GREW filesize gate).
- Extended `runToResult` catch: when `lastStreamedContent` is empty and the error is a `StreamIncompleteError`, calls `synthesizeEmptyBufferPartial` and sets `result.partialOutput`.

The run **remains `failed`** — no final answer was produced — but `SubagentResult.partialOutput` and `.trace` now carry recoverable evidence.

## Tests

**`src/agent/subagent/empty-buffer-partial.test.ts`** (9 new unit tests)
- `buildEmptyBufferError`: zero/non-zero tool-result counts, `toolResultsGathered` presence/absence, message content.
- `synthesizeEmptyBufferPartial`: `undefined` on empty, string content on non-empty, error suffix, missing `sizeBytes` treated as 0.

**`src/agent/subagent/handle.test.ts`** (4 new integration tests via mock session)
- `status: 'failed'` and no `partialOutput` when stream cuts with zero tool results.
- `status: 'failed'` AND `partialOutput` populated when stream cuts after 5 tool results.
- `toolResultsGathered` set to 3 on the error when 3 results were gathered.
- `toolResultsGathered` absent when no results were gathered.

## CI Gates

| Gate | Status |
|---|---|
| `pnpm lint` | ✅ |
| `pnpm build` | ✅ |
| `pnpm test` (all files) | ✅ (1 pre-existing flaky failure in `write-denylist.test.ts` unrelated to this change — confirmed present on `main` before this branch) |
| `pnpm audit:filesize:check` | ✅ (handle.ts net −1 LOC; new file is 59 LOC) |

### DIFF
--- a/src/agent/subagent/empty-buffer-partial.test.ts
+++ b/src/agent/subagent/empty-buffer-partial.test.ts
@@ -1,0 +1,95 @@
+/**
+ * Tests for empty-buffer-partial helpers (GitHub issue #724).
+ *
+ * Verifies that when a subagent stream is cut off with an empty text buffer
+ * but prior tool results exist, the helpers produce the correct error and
+ * partial output — so the parent can distinguish "died with N gathered
+ * results" from "died with nothing gathered".
+ */
+
+import { describe, it, expect } from 'vitest';
+import { StreamIncompleteError } from '../../utils/errors.js';
+import { buildEmptyBufferError, synthesizeEmptyBufferPartial } from './empty-buffer-partial.js';
+import type { SubagentToolResult } from './result.js';
+
+// ---------------------------------------------------------------------------
+// buildEmptyBufferError
+// ---------------------------------------------------------------------------
+
+describe('buildEmptyBufferError', () => {
+  it('returns a StreamIncompleteError', () => {
+    const err = buildEmptyBufferError('sub-1', []);
+    expect(err).toBeInstanceOf(StreamIncompleteError);
+  });
+
+  it('no-results path: toolResultsGathered is absent, message mentions "no output"', () => {
+    const err = buildEmptyBufferError('sub-1', []);
+    expect(err.toolResultsGathered).toBeUndefined();
+    expect(err.message).toContain('no output');
+  });
+
+  it('with-results path: toolResultsGathered equals the count', () => {
+    const results: SubagentToolResult[] = [
+      { toolUseId: 'a', sizeBytes: 1000 },
+      { toolUseId: 'b', sizeBytes: 2000 },
+    ];
+    const err = buildEmptyBufferError('sub-2', results);
+    expect(err.toolResultsGathered).toBe(2);
+    expect(err.message).toContain('2 tool result(s)');
+    expect(err.message).toContain('sub-2');
+  });
+
+  it('single result: toolResultsGathered is 1', () => {
+    const err = buildEmptyBufferError('sub-3', [{ toolUseId: 'x', sizeBytes: 500 }]);
+    expect(err.toolResultsGathered).toBe(1);
+  });
+});
+
+// ---------------------------------------------------------------------------
+// synthesizeEmptyBufferPartial
+// ---------------------------------------------------------------------------
+
+describe('synthesizeEmptyBufferPartial', () => {
+  it('returns undefined when there are no tool results', () => {
+    expect(synthesizeEmptyBufferPartial('sub-1', [])).toBeUndefined();
+  });
+
+  it('returns a string mentioning the count and subagent id', () => {
+    const results: SubagentToolResult[] = [
+      { toolUseId: 'a', sizeBytes: 1000 },
+      { toolUseId: 'b', sizeBytes: 2000 },
+      { toolUseId: 'c', sizeBytes: 500 },
+    ];
+    const out = synthesizeEmptyBufferPartial('sub-2', results);
+    expect(out).toBeTypeOf('string');
+    expect(out).toContain('sub-2');
+    expect(out).toContain('3 tool result(s)');
+    expect(out).toContain('3500'); // total bytes
+  });
+
+  it('includes error count when some results errored', () => {
+    const results: SubagentToolResult[] = [
+      { toolUseId: 'a', isError: true, sizeBytes: 100 },
+      { toolUseId: 'b', sizeBytes: 200 },
+    ];
+    const out = synthesizeEmptyBufferPartial('sub-3', results);
+    expect(out).toContain('1 errored');
+  });
+
+  it('omits error suffix when no results errored', () => {
+    const results: SubagentToolResult[] = [
+      { toolUseId: 'a', sizeBytes: 100 },
+    ];
+    const out = synthesizeEmptyBufferPartial('sub-4', results);
+    expect(out).not.toContain('errored');
+  });
+
+  it('treats missing sizeBytes as 0 when summing total bytes', () => {
+    const results: SubagentToolResult[] = [
+      { toolUseId: 'a' },          // no sizeBytes
+      { toolUseId: 'b', sizeBytes: 500 },
+    ];
+    const out = synthesizeEmptyBufferPartial('sub-5', results);
+    expect(out).toContain('500');  // 0 + 500
+  });
+});
--- a/src/agent/subagent/empty-buffer-partial.ts
+++ b/src/agent/subagent/empty-buffer-partial.ts
@@ -1,0 +1,59 @@
+/**
+ * Helpers for the empty-buffer stream-cut path in {@link SubagentHandleImpl}.
+ *
+ * When a subagent's model stream ends with NO terminal message AND an EMPTY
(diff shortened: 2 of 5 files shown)


griffinwork40:
All citations verified. **Wave 1.5 verification manifest:**

|---|---|---|---|---|
| 1 | citation | `errors.ts:84-94` — `StreamIncompleteError` field | verified | content matches at ref |
| 2 | citation | `stream-cut-retry.ts:67-68` — `isZeroOutputStreamCut` | verified | content matches at ref |
| 3 | absence | No test covers `toolResultsGathered` in `stream-cut-retry*` | confirmed-absent | `git grep` returns 0 matches in `stream-cut-retry*` |

No citations fabricated, no findings dropped.

Now **Wave 2 — Synthesis**. All 3 findings are `low · blocking:false`, which is the default for `low` — no departures from the blocking table. No critical/high findings. No shadow-verify needed.

---

## Review — PR #1126

**fix(subagent): surface partial tool results on empty-buffer stream cut-off (#724)**

`afk/iso-fix-724-1-90v8qw` → `main` · +262/−7 · 5 files · hotfix · light regime

Reviewed ref: `675e6779`

---

### Summary

Clean, well-scoped fix for a real data-loss bug (#724). When a subagent stream cuts off after completing tool-call cycles but before producing assistant text, the prior code threw away all gathered evidence. This PR adds a `toolResultsGathered` field to `StreamIncompleteError`, extracts two helpers into a new sibling file (`empty-buffer-partial.ts`), and wires them into `handle.ts`'s throw-site and catch-block so `SubagentResult.partialOutput` carries a synthesized summary. The run stays `failed` — no status change. 13 new tests across two files cover both helpers and the integration through `runToResult`.

---

### Findings

**1** · `low · blocking:false · confidence:high · api-compat · src/utils/errors.ts:87–93 · ref:675e6779 · citation-type:file-state`
`StreamIncompleteError` gains a new optional property `toolResultsGathered?: number`. Additive — no existing caller accesses this field (verified: zero production importers via grep). No breaking change.
**Suggestion:** Document `toolResultsGathered` as part of the public API surface in release notes if external consumers depend on this error type.

**2** · `low · blocking:false · confidence:high · correctness · src/agent/subagent/stream-cut-retry.ts:67–68 · ref:675e6779 · citation-type:file-state`
`isZeroOutputStreamCut` classifies retriable results by `isError === true && incompleteReason === STREAM_INCOMPLETE` without checking `toolResultsGathered`. After this PR, a stream cut that gathered N>0 tool results still triggers a fresh-fork re-dispatch, which discards attempt-1's `partialOutput`. This is a pre-existing design gap the PR exposes rather than introduces — the PR correctly surfaces partial output on the result; the retry layer has not been updated to honour that signal.
```typescript
export function isZeroOutputStreamCut(result: ToolResult): boolean {
  return result.isError === true && result.incompleteReason === STREAM_INCOMPLETE;
}
```
**Suggestion:** Gate `isZeroOutputStreamCut` (or its caller `runWithStreamCutRetry`) on `toolResultsGathered === undefined || toolResultsGathered === 0` so a run that gathered real evidence is not silently retried.

**3** · `low · blocking:false · confidence:high · test-coverage · src/agent/subagent/stream-cut-retry.ts · ref:675e6779 · citation-type:file-state`
No test covers the interaction between `toolResultsGathered` and `isZeroOutputStreamCut`/`runWithStreamCutRetry`. Absence confirmed: `git grep toolResultsGathered -- stream-cut-retry*` returns zero matches.
**Suggestion:** Add a `stream-cut-retry.test.ts` case exercising a result with `toolResultsGathered: N` to assert the expected (current or updated) retry behaviour.

---

### Dimensions

| Dimension | Verdict |
|---|---|
| Security | no issues found — read `empty-buffer-partial.ts`, `errors.ts` |
| API-compat | Finding 1 (low, non-blocking) |
| Correctness | Finding 2 (low, non-blocking) |
| Spec-compliance | Assessed against PR title+body — all three stated requirements implemented, no scope creep |
| Test-coverage | Finding 3 (low, non-blocking) |
| Perf-observability | no issues found — O(n) in bounded `toolResults`, no new I/O |

---

### What was not checked

- Citations verified inline against branch HEAD `675e6779`.
- Stated intent: PR #1126 title+body — spec-compliance assessed.
- Did not inspect `runInBackground` path for the same empty-buffer gap (only `runToResult` is modified; `runInBackground` calls `runToResult` internally so the fix propagates).
- Did not check Telegram or daemon surfaces.

---

### Done

- Reviewed PR #1126 at ref `675e6779`, light regime (hotfix, 262 lines).
- Wave 1: 1 research-agent covering all 6 dimensions; Wave 1.5: inline citation + absence-claim verification (3/3 verified); Wave 2: synthesis.
- 3 `low · blocking:false` findings — one additive API surface note, one pre-existing retry-layer gap exposed by the PR, one missing test for the retry interaction. No critical, high, or medium findings.
- Verdict: **MERGE** with 0 blocking findings.

---

PR #1128, opened 2026-08-17
feat(tui): add /diff slash command for inline git diff viewing

griffinwork40:
## What

Adds a `/diff` slash command to the interactive REPL that renders the current git working-tree diff inline, colorized with the house palette.

## Usage

```
/diff              — all uncommitted changes (staged + unstaged) via git diff HEAD
/diff --staged     — only staged changes (git diff --cached)
/diff <file>       — changes scoped to a specific path
/diff --staged <file>  — staged changes for a specific path
```

## How it works

- Runs `git diff` via `execSync` in the session's `cwd` (`stats.cwd ?? process.cwd()`)
- Parses the raw unified diff output into a structured hunk list (file pairs, hunk headers, +/-/context lines)
- Renders with `palette.diffAdd` / `palette.diffRemove` / `palette.diffHunk` — respects the active theme and TTY detection, no raw `chalk.<color>()` calls
- Stat header (`+N -M across K files`) is always emitted first and survives body truncation
- Body capped at 60 lines with an elision footer naming the escape hatch — same pattern as the existing `AFK_DIFF_LINES` flush-mode diff renderer

## Files changed

| File | Change |
|------|--------|
| `src/cli/slash/commands/diff.ts` | New command — 256 lines (well under the 350-line ceiling) |
| `src/cli/slash/index.ts` | Import + `register(diffCmd)` wired into `registerAll()` |

## CI checks (on this branch's files)

- `pnpm lint` — passes (tsc --noEmit clean on my changes; pre-existing dirty-tree errors in unrelated in-progress files on this machine are not part of this PR)
- `pnpm audit:chalk:check` — 0 violations
- `pnpm audit:filesize:check` — `diff.ts` at 256 lines; no new violations introduced

### DIFF
--- a/src/cli/slash/commands/diff.test.ts
+++ b/src/cli/slash/commands/diff.test.ts
@@ -1,0 +1,95 @@
+import { execFileSync } from 'node:child_process';
+import { mkdtempSync, rmSync, writeFileSync, chmodSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { stripEscapeSequences } from '../../../utils/terminal-sanitize.js';
+import type { SessionStats, SlashContext } from '../types.js';
+import { diffCmd } from './diff.js';
+
+const dirs: string[] = [];
+
+afterEach(() => {
+  delete process.env['AFK_DIFF_LINES'];
+  for (const dir of dirs.splice(0)) rmSync(dir, { recursive: true, force: true });
+});
+
+function repo(): string {
+  const cwd = mkdtempSync(join(tmpdir(), 'afk-diff-'));
+  dirs.push(cwd);
+  execFileSync('git', ['init', '-q'], { cwd });
+  execFileSync('git', ['config', 'user.email', 'test@example.com'], { cwd });
+  execFileSync('git', ['config', 'user.name', 'Test'], { cwd });
+  writeFileSync(join(cwd, 'file.txt'), 'one\ntwo\nthree\n');
+  execFileSync('git', ['add', '.'], { cwd });
+  execFileSync('git', ['commit', '-qm', 'initial'], { cwd });
+  return cwd;
+}
+
+function context(cwd: string): { ctx: SlashContext; output: string[] } {
+  const output: string[] = [];
+  const stats = { cwd } as SessionStats;
+  const write = (text = ''): void => { output.push(stripEscapeSequences(text)); };
+  return {
+    output,
+    ctx: {
+      session: { current: {} } as SlashContext['session'],
+      stats,
+      out: { line: write, raw: write, success: write, info: write, warn: write, error: write },
+      ui: { clearScreen: vi.fn(), repaintStatusLine: vi.fn() },
+    },
+  };
+}
+
+describe('/diff', () => {
+  it('passes a pathspec as data instead of shell syntax', async () => {
+    const cwd = repo();
+    const marker = join(cwd, 'injected');
+    writeFileSync(join(cwd, 'file.txt'), 'changed\n');
+    const { ctx } = context(cwd);
+
+    await diffCmd.handler(ctx, `file.txt;touch ${marker}`);
+
+    expect(() => execFileSync('test', ['-e', marker])).toThrow();
+  });
+
+  it('renders mode-only changes that have no text patch headers', async () => {
+    const cwd = repo();
+    chmodSync(join(cwd, 'file.txt'), 0o755);
+    const { ctx, output } = context(cwd);
+
+    await diffCmd.handler(ctx, '');
+
+    expect(output.join('\n')).toContain('old mode 100644');
+    expect(output.join('\n')).toContain('new mode 100755');
+    expect(output.join('\n')).not.toContain('No changes');
+  });
+
+  it('sanitizes terminal escapes from changed lines', async () => {
+    const cwd = repo();
+    writeFileSync(join(cwd, 'file.txt'), 'safe\x1b]52;c;clipboard\x07text\n');
+    const { ctx, output } = context(cwd);
+
+    await diffCmd.handler(ctx, '');
+
+    const rendered = output.join('\n');
+    expect(rendered).toContain('safetext');
+    expect(rendered).not.toContain('\x1b');
+    expect(rendered).not.toContain('clipboard');
+  });
+
+  it('honors AFK_DIFF_LINES=0 and counts context in a truncated remainder', async () => {
+    const cwd = repo();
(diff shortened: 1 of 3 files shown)


griffinwork40:
@codex fix comments

griffinwork40 (commented):

## Review: feat(tui): add /diff slash command for inline git diff viewing

**Reviewed ref:** `c91f4195eb8e84c0437ac5b89303b3e930597b76`
**Regime:** full (371 lines, 3 files — 2 new, 1 modified)
**Stated intent:** PR #1128 title+body — spec-compliance assessed.

---

### Decision: DO NOT MERGE — 1 medium blocking (security); 2 medium waived, 4 low, 2 nit.

---

### Findings

**medium · blocking:true · confidence:high · security · `diff.ts:198-203` · ref:c91f4195 · diff-context**
**finding:** Git stderr output is not sanitized before being surfaced to the terminal. When `execFileSync` throws (e.g. not a repo, bad ref), `runGitDiff` constructs its error message from `e.stderr` / `e.message` and returns it as-is. The handler then passes this unsanitized string directly to `out.error()`. While git's own error messages are typically safe, they are externally-sourced strings reaching the terminal without passing through `sanitizeForDisplay()`, violating the security contract this file otherwise upholds for all other output paths.
**evidence:**
```typescript
const detail = (e.stderr ?? '').trim() || (e.message ?? '');
return { ok: false, message: detail || 'git diff failed' };
```
**suggestion:** Wrap as `sanitizeForDisplay(String(e.stderr ?? e.message ?? 'unknown error'))` inside the catch block.

---

**medium · blocking:false · confidence:high · correctness · `diff.ts:71` · ref:c91f4195 · diff-context · waived: only manifests for metadata-only diffs (mode changes, binary files) on filenames containing ` b/` — a narrow, non-data-affecting cosmetic condition**
**finding:** The greedy regex `/^diff --git a\/(.*) b\/(.*)$/` mis-parses the from/to pair for filenames whose path contains a space followed by `b/`. For a filename like `a b/c`, the `diff --git` line is `diff --git a/a b/c b/a b/c` and the greedy `.*` splits incorrectly. The `---`/`+++` lines do correctly override for diffs that carry patch content, so this only affects metadata-only diffs (mode changes, binary files) on such filenames — a cosmetic mislabeling.
**evidence:**
```typescript
const match = /^diff --git a\/(.*) b\/(.*)$/.exec(line);
currentFrom = match?.[1] ?? line.slice('diff --git '.length);
```
**suggestion:** Prefer the `---`/`+++` names as authoritative when available; use the `diff --git` header only as a fallback.

---

**medium · blocking:false · confidence:high · spec-compliance · `diff.ts:242-247` · ref:c91f4195 · diff-context · waived: documents a known behavioral boundary of `git diff HEAD`, not a bug in the PR's implementation**
**finding:** The spec claims `/diff` shows "all uncommitted changes" but `git diff HEAD` does not show untracked (never-added) files. A user creating a new file and running `/diff` sees "No changes" — which is technically correct for git but surprising given the spec wording.
**suggestion:** Document this as a known limitation in the `hint` text, e.g. `"Untracked files not shown — run git add first."`, or note it in the PR body.

---

**low · blocking:false · confidence:high · test-coverage · `diff.test.ts:1-95` · ref:c91f4195 · diff-context**
**finding:** The four tests cover command injection, mode-only changes, escape sanitization, and AFK_DIFF_LINES — but the explicitly-advertised `--staged` and `<file>` forms have zero tests. `parseArgs` recognizes `--staged`/`--cached` and the handler builds the `['--cached']` / `['--', file]` argument vectors, but neither path is exercised.
**suggestion:** Add tests that invoke `handler(ctx, '--staged')` and `handler(ctx, 'file.txt')` and verify the correct content appears.

---

**low · blocking:false · confidence:high · test-coverage · `diff.test.ts:1-95` · ref:c91f4195 · diff-context**
**finding:** The `No changes` branch (clean working tree) and the `!result.ok` error branch (not a git repo) are both unexercised. These are the two primary early-return paths.
**suggestion:** Add a test with a clean repo asserting "No changes" appears, and a test with `stats.cwd` pointing to a non-git directory asserting the error path surfaces a message.

---

**low · blocking:false · confidence:high · correctness · `diff.ts:134-137` · ref:c91f4195 · diff-context**
**finding:** The truncation footer reports `totalBodyLines - bodyLines` remaining lines, but `totalBodyLines` excludes hunk header lines (which are rendered unconditionally and not counted). The displayed remainder undercounts by the number of subsequent hunk headers. Cosmetic only.
**suggestion:** Either count hunk headers in `totalBodyLines` or document the intentional exclusion.

---

**low · blocking:false · confidence:high · spec-compliance · `diff.ts:242` · ref:c91f4195 · diff-context**
**finding:** In a repository with no commits (empty repo), `git diff HEAD` fails with `fatal: ambiguous argument 'HEAD': unknown revision`. The current code surfaces this as an error, which is correct but unpolished.
**suggestion:** Acknowledge as a known edge case in the PR body or hint text.

---

**nit · blocking:false · confidence:high · correctness · `diff.ts:193-194` · ref:c91f4195 · diff-context**
**finding:** The comment "git diff exits 0 when there are diffs, 1 is not an error here" reads as though `git diff` normally exits 1 when diffs exist, which is wrong. `git diff` (without `--exit-code`) always exits 0 regardless of diffs. The second comment correctly explains `--exit-code`, but the first comment is misleading.
**suggestion:** Replace with: `"git diff (without --exit-code) always exits 0; any non-zero exit is a genuine error."`

---

**nit · blocking:false · confidence:high · api-compat · `diff.ts:130` + env.ts · ref:c91f4195 · diff-context**
**finding:** `AFK_DIFF_LINES` already exists in `env.ts` scoped to `write_file` tool call diffs. This PR adds a second consumer without updating that description. The env var's `0` semantics are consistent (both mean "no cap"), but the description is now stale.
**suggestion:** Update env.ts description to mention both consumers.

---

### What was not checked

- Citations verified inline against branch HEAD `c91f4195eb8e84c0437ac5b89303b3e930597b76`.
- Stated intent: PR #1128 title+body — spec-compliance assessed.
- Did not run the test suite or build locally.
- Did not verify `palette.diffAdd` / `palette.diffRemove` / `palette.diffHunk` exist on the palette object (assumed from import).
- CI has a `Lint & Build` FAILURE on the latest commit — not investigated here (may be related to TypeScript compilation).

- **S-1 (sanitize multi-line strings):** Dropped — `sanitizeForDisplay` is called per-line on individual `dl.text`, `hunk.header`, `metadata`, and filename strings, never on multi-line blobs. The premise was false.
- **A-2 (parseInt NaN/negative):** Dropped — the code uses a ternary chain (`=== 0` / `> 0` / default), not `slice(0, N)`. Both NaN and negative values correctly fall through to the default `MAX_DIFF_BODY_LINES`.

PR #1133, opened 2026-08-17
fix(doctor): refresh expired keychain OAuth before checking

### LINKED ISSUE #1131: doctor check fails with keychain OAuth: expired token never refreshed
## Summary

`afk doctor` reports `✗ Anthropic API Key` even when valid Claude Code OAuth credentials exist in the macOS keychain, because the doctor command never triggers a token refresh before checking.

## Steps to reproduce

1. Log into Claude Code via OAuth (`claude login`) — credentials are stored in macOS keychain under `"Claude Code-credentials"`
2. Wait >1 hour (or any period long enough for the access token to expire)
3. Run `afk doctor`

**Expected:** Doctor detects the keychain credential (refreshing if needed) and reports a pass.
**Actual:** `✗ Anthropic API Key — Fix: Set ANTHROPIC_API_KEY or run 'afk login'`

## Root cause

The credential resolution chain works correctly:

`checkAnthropicKey()` → `getApiKey()` → `getApiKeyForModel()` → `resolveCredentialForModel()` → `loadAnthropicCredential()` → `loadClaudeCodeOauthToken()`

But `loadClaudeCodeOauthToken()` (keychain.ts) checks expiry **synchronously** and returns `undefined` for expired tokens:

```typescript
if (parsed.expiresAt !== undefined && parsed.expiresAt <= Date.now()) {
    process.stderr.write(
      'agent-afk: Claude Code OAuth token in keychain is expired. Run `claude login` to refresh.\n',
    );
    return undefined;
}
```

The async refresh function `preloadClaudeKeychainOAuth()` is called at startup for credential-gated commands (`chat`, `interactive`, `daemon`, `farm`) via `AUTH_GATED_COMMANDS` in index.ts — but `doctor` is explicitly excluded from that set.

So when `afk doctor` runs, the token refresh never fires. If the access token has expired (which it will after ~1 hour), the check fails. The stderr warning message (`"agent-afk: Claude Code OAuth token in keychain is expired"`) is also easy to miss in the doctor output.

## Suggested fix

One or more of:

1. Have `checkAnthropicKey()` call `preloadClaudeKeychainOAuth()` before checking, so expired tokens get refreshed the same way `chat`/`daemon` do.
2. Add `'doctor'` to `AUTH_GATED_COMMANDS` (or add a separate pre-check refresh step for doctor).
3. At minimum, distinguish "no credential found" from "credential exists but expired" in the doctor output — the current `✗` with "Set ANTHROPIC_API_KEY or run `afk login`" is misleading when the real fix is just waiting for a refresh.

## Workarounds

- Run `afk login` and paste the OAuth token (saves to `~/.afk/config/afk.env` as `CLAUDE_CODE_OAUTH_TOKEN`, bypassing keychain expiry)
- Run any credential-gated command first (e.g. `afk chat`) to trigger the refresh, then run `afk doctor`
- Export `CLAUDE_CODE_OAUTH_TOKEN` in the shell environment

## Environment

- agent-afk v5.127.2
- macOS (Darwin 25.5.0)
- Claude Code OAuth (claude.ai subscription)

griffinwork40:
## Summary

`afk doctor` reported `✗ Anthropic API Key` when the only credential was a Claude Code OAuth token stored in the macOS keychain, because expired tokens were never refreshed.

## Root cause

The async refresh path (`preloadClaudeKeychainOAuth`) was wired to auth-gated commands (chat, interactive, daemon) via `AUTH_GATED_COMMANDS`, but `doctor` was excluded. When the OAuth access token expired (~1 hour), `loadClaudeCodeOauthToken()` returned `undefined` and the check failed with a misleading "Set ANTHROPIC_API_KEY" message.

## Fix

Call `preloadClaudeKeychainOAuth('anthropic-direct')` at the top of `checkAnthropicKey()`. This:
- Refreshes expired keychain OAuth tokens before the sync read
- Is a no-op when an env credential already exists (`ANTHROPIC_API_KEY` / `CLAUDE_CODE_OAUTH_TOKEN`)
- Is a no-op for non-Anthropic providers
- Adds no overhead for env-var users

## Test

Added a test verifying `checkAnthropicKey` calls `preloadClaudeKeychainOAuth` with `'anthropic-direct'` before checking.

## Changes

| File | Change |
|------|--------|
| `src/cli/commands/doctor-checks.ts` | Import + call `preloadClaudeKeychainOAuth` in `checkAnthropicKey()` |
| `src/cli/commands/doctor-checks.test.ts` | New test: verifies refresh is called before key check |

Closes #1131

### DIFF
--- a/src/cli/commands/doctor-checks.test.ts
+++ b/src/cli/commands/doctor-checks.test.ts
@@ -1,5 +1,5 @@
-import { describe, it, expect } from 'vitest';
-import { checkImportAvailable } from './doctor-checks.js';
+import { describe, it, expect, vi } from 'vitest';
+import { checkImportAvailable, checkAnthropicKey } from './doctor-checks.js';
 import type { DetectedSource } from '../../config/import-sources.js';
 
 /** Minimal fixture for a present source with zero assets. */
@@ -18,6 +18,21 @@ function makeSource(
     ...overrides,
   };
 }
+
+vi.mock('../../agent/auth/credential-resolver.js', async (importOriginal) => {
+  const orig = await importOriginal<typeof import('../../agent/auth/credential-resolver.js')>();
+  return { ...orig, preloadClaudeKeychainOAuth: vi.fn().mockResolvedValue(undefined) };
+});
+
+describe('checkAnthropicKey', () => {
+  it('calls preloadClaudeKeychainOAuth before checking the key', async () => {
+    const { preloadClaudeKeychainOAuth } = await import(
+      '../../agent/auth/credential-resolver.js'
+    );
+    await checkAnthropicKey();
+    expect(preloadClaudeKeychainOAuth).toHaveBeenCalledWith('anthropic-direct');
+  });
+});
 
 describe('checkImportAvailable', () => {
   it('returns null when detected list is empty', async () => {
--- a/src/cli/commands/doctor-checks.ts
+++ b/src/cli/commands/doctor-checks.ts
@@ -13,6 +13,7 @@ import { getConcurrencyStatuses } from '../../config/concurrency.js';
 import { access, constants, mkdir, readFile } from 'fs/promises';
 import { execSync } from 'child_process';
 import { getApiKey, getCodexApiKey } from '../shared-helpers.js';
+import { preloadClaudeKeychainOAuth } from '../../agent/auth/credential-resolver.js';
 import {
   getAfkConfigDir,
   getAfkStateDir,
@@ -34,6 +35,12 @@ export interface Check {
 }
 
 export async function checkAnthropicKey(): Promise<Check> {
+  // Attempt a keychain OAuth refresh before checking — without this, an expired
+  // Claude Code OAuth token (which expires hourly) causes a false negative.
+  // The guard inside `preloadClaudeKeychainOAuth` skips the refresh when an env
+  // credential already exists or the provider isn't anthropic-direct.
+  await preloadClaudeKeychainOAuth('anthropic-direct');
+
   const key = getApiKey();
   if (key) {
     return { name: 'Anthropic API Key', state: 'pass', detail: 'ANTHROPIC_API_KEY set' };


griffinwork40:

**Check A — Citation verification:**
- Finding #1 (`doctor-checks.test.ts:27–35`, diff-context): **verified** — lines 27–35 of the diff show the `checkAnthropicKey` describe block, visible in the diff hunk.
- Finding #2 (`doctor-checks.ts:42`, file-state): **verified** — line 42 at ref `0d119e24` contains the comment about `preloadClaudeKeychainOAuth` skipping the refresh.
- Sub-agent claim about `cli/index.ts:218` calling `preloadClaudeKeychainOAuth`: **verified** — line 218 at ref `0d119e24` calls `await preloadClaudeKeychainOAuth(provider)`.
- Sub-agent claim about `cli/index.ts:203` exempting doctor from credential gate: **verified** — lines 203–204 contain the comment "afk doctor … must never be blocked."
- Docstring claim about return-value fallback (`credential-resolver.ts:63–66`): **verified** — lines 63–66 describe the return value for locked-keychain fallback.

**Check B — Absence-claim verification:**
- Claim: "no `state.*pass` hits in the `checkAnthropicKey` block": **confirmed-absent** — `git grep -n 'state.*pass' 0d119e24 -- src/cli/commands/doctor-checks.test.ts` returns exit code 1 (zero matches).

All citations verified, all absence claims confirmed. No findings dropped. Now let me synthesize.

---

## Review: PR #1133 — `fix(doctor): refresh expired keychain OAuth before checking`

**Reviewed ref:** `0d119e24e2558c504bd1065adfe1823dffcea8ae`
**Change type:** hotfix · **Regime:** light · **Files:** 2 · **Lines:** +24/−2

**Stated intent:** `afk doctor` falsely reported `✗ Anthropic API Key` when the only credential was an expired Claude Code keychain OAuth token, because `doctor` was excluded from the startup preload path. Fix: call `preloadClaudeKeychainOAuth('anthropic-direct')` inside `checkAnthropicKey()`.

### Findings

| # | Sev | Blocking | Dim | File | Finding |
|---|---|---|---|---|---|
| 1 | low | false | test-coverage | `doctor-checks.test.ts:27–35` | Test asserts `preloadClaudeKeychainOAuth` was called with `'anthropic-direct'` but does not exercise the motivating path (expired token → refresh → `state: 'pass'`). A second test where the mock side-effects a credential and asserts `result.state === 'pass'` would close the gap. |
| 2 | nit | false | spec-compliance | `doctor-checks.ts:42` | Return value of `preloadClaudeKeychainOAuth` is discarded. On a locked/read-only keychain, the refresh returns a fresh token but cannot write it back; `getApiKey()` then reads stale data → false-negative `fail`. This is a pre-existing gap shared with the startup caller (`cli/index.ts:218`) — the PR does not worsen it. Suggestion: `const fresh = await preloadClaudeKeychainOAuth('anthropic-direct'); const key = fresh ?? getApiKey();` |

**security:** no issues found — `preloadClaudeKeychainOAuth` reads/refreshes a keychain entry; no secrets logged or exposed.
**api-compat:** no issues found — no exported API changed.
**correctness:** no issues found — placement is load-bearing and correct: `runFirstRunDetector` explicitly skips `doctor` via `needsCredentialGate`, so this call is the only refresh path for `afk doctor`.
**perf-observability:** no issues found — the call is internally guarded (no-op for env-var users and non-Anthropic providers) and bounded by fetch timeout.

### What was not checked
- Citations verified inline against branch HEAD `0d119e24`.
- Stated intent: PR #1133 title + body — spec-compliance assessed.
- Did not run the test suite.
- Did not verify `refreshClaudeCodeOauthToken` internals beyond its documented contract.

### Decision: MERGE — 0 blocking (1 low, 1 nit).

Clean, minimal hotfix that correctly addresses the stated bug. The two findings are a coverage improvement suggestion and a pre-existing gap — neither is blocking.

**Done**

- Reviewed PR #1133 (`fix(doctor): refresh expired keychain OAuth before checking`) at ref `0d119e24`.
- All citations verified against reviewed ref. All absence claims confirmed via `git grep`.
- Verdict: **MERGE** — 0 blocking findings (1 low test-coverage gap, 1 nit pre-existing locked-keychain edge case).
- No files were modified (read-only review).

---

PR #1015, opened 2026-08-12
fix(skills): block skill self-recursion through agent children

griffinwork40:
## Problem

When a skill like `ground-state` was dispatched as a fork, it could re-dispatch itself recursively. Two separate gaps caused this:

**Gap 1 (depth-1 recursion):** The manifest excluded the skill's own entry via `skillDispatchName`, but the `skill` tool accepts any string name — so a model reading its own name in the SKILL.md system prompt could still call `skill("ground-state")` and fork a recursive copy. This burned 2+ minutes and ~22 tool calls before hitting the depth cap.

**Gap 2 (depth-2 recursion):** The skill fork dispatched `agent` children (surveyors). Those grandchildren got a fresh config without `skillDispatchName`, so the originating skill reappeared in their manifest. The code comment at `src/agent/types/config-types.ts:682-684` documented this as "deliberately unfixed, bounded by depth cap."

### Fix A — Execution-level guard (depth-1)

Added a check in `SkillExecutor.execute()` that fires immediately after parsing the skill name:

```ts
if (this.ctx.skillDispatchName !== undefined && parsed.name === this.ctx.skillDispatchName) {
  return {
    content: `Skill "${parsed.name}" cannot re-dispatch itself (already executing at this level). Return findings via your system prompt contract instead.`,
    isError: true,
  };
}
```

The guard requires `skillDispatchName` on `SkillExecutorContext` (added to `types.ts`).

### Fix B — skillDispatchName propagation (depth-2)

Extended the `childSkillExecutorFactory` signature with an optional 6th `skillDispatchName` parameter. When a skill fork's agent children are built:

1. `fork-child-config.ts` seeds `defaultConfig.skillDispatchName` on the grandchild `SubagentExecutor` 
2. `buildChildConfig()` in `child-config.ts` propagates it to unnamed agent grandchildren's `AgentConfig` and to their `SkillExecutor` (via the factory 6th arg)
3. Named agents (`agent_type:` dispatches) are excluded — they own their own system-prompt identity

Both `SubagentExecutorContext.defaultConfig` and `BuildChildConfigArgs.defaultConfig` types were extended to include `skillDispatchName`.

## Tests

Added 6 new tests to `src/agent/tools/skill-dispatch-routing.test.ts`:

- **Fix A (3 tests):** `SkillExecutor — self-recursion execution guard`
  - Rejects a `skill(name)` call matching `skillDispatchName` with a clear error message
  - Allows a different skill to be called (guard is name-specific)
[...]

PR #1134, opened 2026-08-17
perf(ground-state): eliminate 3-subagent dispatch overhead in pre-flight recon

griffinwork40:
## Problem

`/ground-state` takes ~6 minutes for what are deterministic lookups. The skill fork dispatches 3 parallel research-agent sub-agents (state/infra/memory surveyors), each incurring ~90-120s LLM cold-start latency. The infrastructure surveyor alone consumed 56 tool calls over 5m42s for a "return 5-bullet inventory" task.

## Root cause

Shadow-verify confirmed the 50-round default budget (`SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS`) was present and working — the bottleneck was **3 LLM cold-start handshakes**, not tool-use round counts. Capping iterations addresses the ceiling but not the floor.

Devils-advocate (3 `strong` critics, unanimous) identified the architecture mismatch: ground-state uses 3 LLM sessions for tasks that don't require reasoning — `git status`, `glob`, `memory_search` are deterministic reads.

## Fix

Rewrite the SKILL.md so the fork runs all three surveys **directly using its own tools** instead of dispatching sub-agents. The critical new instruction:

> Do NOT dispatch any sub-agents via the `agent` or `skill` tools — every lookup in this phase is a deterministic read that you execute yourself.

This eliminates 3 of 4 LLM cold-starts. Expected wall-clock: **<30 seconds** (one fork + a few tool calls) vs ~6 minutes.

## What's preserved

- Frontmatter unchanged (`read-only: true`, `context: fork`)
- Read-only constraint contract unchanged
- Domain-adaptive surveying tables retained as reference
- 6-line snapshot synthesis format unchanged
- Brief Anchor construction + orchestrator usage contract unchanged
- `CLAUDE.md` conditional reference ("on a Claude Code surface") kept — shadow-verify confirmed it's surface-gated, not stale
- `memory_search` invariant kept

## What's NOT in this PR

- No changes to `researchAgent` definition or `READONLY_AGENT_MAX_TOOL_USE_ITERATIONS` — devils-advocate confirmed this would break `/review`, `/shadow-verify`, `/diagnose` which need 15-40 rounds
- No changes to the skill executor or fork-dispatch code
- No changes to the self-recursion guards (PRs #737 and #1015 already shipped those)

## Evidence chain

| Step | Finding |
|------|---------|
| Initial diagnosis | 3 research-agents by design; self-recursion guard working correctly |
| Shadow-verify (3 verifiers) | 2 of 6 original claims refuted: budget exists at 50 rounds (not missing); CLAUDE.md reference is conditionally correct (not stale) |
| Devils-advocate (3 critics) | Unanimous `strong`: inline execution beats iteration-cap change. Root cause is cold-start latency, not round counts |
| Synthesis | Ranked: inline (1st) > full collapse (2nd) > prose budget (3rd) > global cap change (4th) |

### DIFF
--- a/src/bundled-plugins/awa-bundled/bundled.test.ts
+++ b/src/bundled-plugins/awa-bundled/bundled.test.ts
@@ -86,7 +86,7 @@ const PINNED_HASHES = {
   // wave keeps dispatching). The upstream ground-state has neither layer, so
   // both lines are permanent bundled-only divergence.
   'ground-state':
-    '69b0eed4b31370a3633da41b502c4c88b22336e467b108c6dcd9301f9a90827a',
+    '5c924f6a685e50fa40650f62fdbc163d4acc02a5155dc35807dc52b7536e9025',
 
   polish:
     '03c7d8d1f593d4aca3327247d5c226e957bae84982b8a438c5834ba74ee4184c',
--- a/src/bundled-plugins/awa-bundled/bundled.test.ts
+++ b/src/bundled-plugins/awa-bundled/bundled.test.ts
@@ -86,7 +86,7 @@ const PINNED_HASHES = {
   // wave keeps dispatching). The upstream ground-state has neither layer, so
   // both lines are permanent bundled-only divergence.
   'ground-state':
-    'f24a0fbc572aa65dde3cbceb76ce1ec36d8ed8d88e142aa0007425ec3e906cad',
+    '69b0eed4b31370a3633da41b502c4c88b22336e467b108c6dcd9301f9a90827a',
 
   polish:
     '03c7d8d1f593d4aca3327247d5c226e957bae84982b8a438c5834ba74ee4184c',
--- a/src/bundled-plugins/awa-bundled/skills/ground-state/SKILL.md
+++ b/src/bundled-plugins/awa-bundled/skills/ground-state/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: ground-state
-description: "Before starting any non-trivial implementation, dispatch a parallel pre-flight reconnaissance wave to triangulate git state, project infrastructure, and prior-session memory. Wave 4 auto-assembles a verified grounding preamble — a session-scoped artifact the orchestrator pastes verbatim into every subsequent sub-agent brief — eliminating stale-worktree reads and silent wrong-path errors before the first edit."
+description: "Before starting any non-trivial implementation, run a pre-flight reconnaissance pass to triangulate git state, project infrastructure, and prior-session memory. Auto-assembles a verified grounding preamble — a session-scoped artifact the orchestrator pastes verbatim into every subsequent sub-agent brief — eliminating stale-worktree reads and silent wrong-path errors before the first edit."
 read-only: true
 context: fork
 failure_modes:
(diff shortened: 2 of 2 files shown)


griffinwork40:
The SKILL.md says "Issue all three surveys in a single batched tool-use round where possible" — meaning the fork can issue bash + glob + memory_search calls in one tool-use turn, which does execute in parallel. The "in parallel" claim in `routing-directive.ts` is technically still achievable through batched tool calls, just not through sub-agent dispatch. Finding #2 needs to be reconsidered — the parallelism is preserved at the tool-call level, just the mechanism changed.

Now I have all verification results. Let me synthesize the review.

---

## PR #1134 Review

**perf(ground-state): eliminate 3-subagent dispatch overhead in pre-flight recon**

**Target:** `afk/ground-state-inline-recon` → `main` · +47/−30 · 2 files
**Reviewed ref:** `7062139866d21adeead79c3eeb3a8a0d913d51ae`
**Stated intent:** Rewrite `/ground-state` SKILL.md to run all three surveys (state/infra/memory) directly as inline tool calls instead of dispatching 3 sub-agents, eliminating 3 LLM cold-starts. Expected: ~30s vs ~6min.
**Regime:** light

---

### Findings

**1.** `medium · blocking:true · confidence:high · spec-compliance · SKILL.md:73 · ref:706213 · diff-context`

**Epistemic-confidence discrimination criteria silently removed.** The old line read:

```
- Epistemic confidence: `<high|medium|low>` — based on how much state could be verified. Flag if working directory is sparse, if domain is unfamiliar, or if key artifacts may be missing.
```

The new line drops the three concrete triggers (`sparse`, `unfamiliar domain`, `missing artifacts`) that tell the fork *when* to downgrade from `high` to `medium`/`low`. These are not padding — they are the actionable discrimination conditions the orchestrator uses to decide whether evidence is sufficient to proceed. The PR body explicitly states "6-line snapshot synthesis format unchanged" as a preservation goal, but this truncation changes the behavioral content of that format.

**Suggestion:** Restore the clause: `Flag if working directory is sparse, if domain is unfamiliar, or if key artifacts may be missing.`

---

**2.** `low · blocking:false · confidence:medium · api-compat · src/agent/routing-directive.ts:28 · ref:706213 · file-state [not changed in this PR]`

**`routing-directive.ts` says "in parallel" — technically still true via batched tool calls, but the mechanism changed.** The routing directive reads: *"the skill triangulates git + infra + prior-session memory in parallel, which the inline checks miss."* This was written when 3 sub-agents ran concurrently. After this PR the fork issues the surveys in "a single batched tool-use round" — which does run tool calls concurrently, so "in parallel" is not false. However, the primary justification for why `/ground-state` outperforms inline checks is no longer the parallelism — it's the *coverage* (memory_search is unreachable from inline `git status`/`get_runtime_state`). A future reader might still wonder.

**Suggestion:** Optionally update the clause to: *"the skill triangulates git + infra + prior-session memory in a single batched pass, covering dimensions the inline checks miss"*. Not blocking — current wording is not incorrect, just less precise.

---

**3.** `low · blocking:false · confidence:high · spec-compliance · bundled.test.ts:82-87 · ref:706213 · file-state [not changed in this PR]`

**Hash-bump comment describes the pre-refactor architecture.** The comment above the `'ground-state'` hash entry says *"pins it to forking so the recon wave keeps dispatching"* — "wave keeps dispatching" refers to the sub-agent dispatch architecture this PR eliminates. The `context: fork` frontmatter line still applies (the fork itself still runs), but "the recon wave keeps dispatching" no longer describes what happens inside it. Future developers reading this comment to understand why the hash changed will be misled.

**Suggestion:** Append a note, e.g.: `// PR #1134: recon surveys now run inline (no sub-agent dispatch); fork context remains for tool execution.`

---

**4.** `nit · blocking:false · confidence:high · correctness · SKILL.md:45 · ref:706213 · file-state`

**Infrastructure survey heading parenthetical omits `grep` and `list_directory`.** Heading reads `*(bash/glob/read_file)*` but the constraint block lists `grep` and `list_directory` as permitted tools, and the infra survey would naturally use them. Minor discoverability inconsistency.

**Suggestion:** Update to `*(bash/glob/grep/read_file)*` or similar.

---

**5.** `nit · blocking:false · confidence:medium · test-coverage · bundled.test.ts · ref:706213 · file-state`

**No content-level test for the no-dispatch invariant.** The bundled test file has precedent for structural contract tests (e.g., `review Wave 1 keeps the no-git citation contract (#726)` at line 190). The new load-bearing instruction *"Do NOT dispatch any sub-agents via the `agent` or `skill` tools"* has no equivalent assertion. The hash test catches any edit, but a targeted content test is more resilient to rewrites that preserve the hash while violating the intent.

**Suggestion:** Add `expect(content).toContain('Do NOT dispatch any sub-agents')` mirroring the review pattern.

---

### Wave 1.5 Citation Verification

| Finding | Citation type | Status |
|---------|--------------|--------|
| #1 SKILL.md:73 | diff-context | **verified** — line visible in diff hunk, confirmed at old line 56 (pre-merge) and new line 73 (post-merge) |
| #2 routing-directive.ts:28 | file-state | **verified** — `git show` confirms "in parallel" at line 28 of the reviewed ref |
| #3 bundled.test.ts:82-87 | file-state | **verified** — lines 82-87 confirmed at reviewed ref, "recon wave keeps dispatching" present |
| #4 SKILL.md:45 | file-state | **verified** — heading confirmed at line 45 of reviewed ref |
| #5 bundled.test.ts (pattern) | file-state | **verified** — absence confirmed: `grep` for `dispatch.*sub-agent` in test assertions returns 0 matches for ground-state |

No citations fabricated. No absence claims falsified.

---

### Epistemic Scope

- Citations verified against branch HEAD `7062139866d21adeead79c3eeb3a8a0d913d51ae`
- Stated intent: PR #1134 title+body — spec-compliance assessed
- What was not checked: runtime execution timing (the "30s vs 6min" perf claim is architectural inference, not measured); no test was executed; user-scope `~/.afk/skills/ground-state/` was not checked for a parallel copy that might also need updating

---

### Decision: DO NOT MERGE — 1 medium blocking (spec-compliance); 1 low, 2 nit non-blocking.

Finding #1 is blocking: the truncated epistemic-confidence line removes the discrimination criteria that downstream orchestrators use to judge evidence sufficiency. This is a one-clause restoration — the fix is to append the deleted sentence back.

Finding #2 was downgraded from `medium` to `low` during verification: "in parallel" is technically preserved through batched tool calls, so the routing-directive claim is imprecise but not false.

**Done**

- Review completed for PR #1134 against ref `7062139866d21adeead79c3eeb3a8a0d913d51ae`
- 5 findings emitted: 1 blocking medium (epistemic-confidence criteria stripped), 1 non-blocking low (routing-directive wording), 1 non-blocking low (stale hash comment), 2 nits
- No files edited, no PR comments posted (read-only review)
- Verdict: **DO NOT MERGE** until finding #1 is addressed — one sentence restoration

---

PR #1137, opened 2026-08-18
fix(cli): reject unknown subcommands instead of falling through to interactive (#710)

### LINKED ISSUE #710: CLI: mistyped subcommand is swallowed by the default `interactive` command — no "unknown command" error, misleading "unknown option" instead

griffinwork40:
## Summary

Fixes #710 — closes the mode-2 path where a mistyped single-word subcommand
(`afk skill`, `afk shedule`) was silently swallowed by Commander's default
`interactive` command, launching a nested REPL session instead of reporting
an error. A forkbomb reaching 134 live processes was observed on 2026-08-09
via this path.

## Background

The mode-1 guard (`installUnknownCommandGuard`, PR history) already fixed
the case where an unknown token was followed by an unknown flag (e.g.
`afk config_set env X --unset`). The mode-2 path — bare single-word token,
no flags — was explicitly left out of scope because a Levenshtein ≤ 3
heuristic against the command-name pool produced 14–16/20 false positives
on ordinary one-word prompts (`stats`, `log`, `plan`, …), even at ≤ 2 it
was 8/20.

## Approach

`checkBareUnknownCommand(input, program)` in `unknown-command-guard.ts` is
called from the `interactive` action **before any side-effect** (before env
mutations, spinner, worktree, or session bootstrap). It intercepts the action
when both conditions hold:

1. `input` is a **single bare word** with no spaces, not starting with `/`
   or `-`.
2. The word passes `isUnrecognizedCommandToken` (not a known command/alias).
3. The word has **Levenshtein distance ≤ 2** from a registered subcommand
   name — tight enough to capture typos (`shedule` → `schedule`, `skil` →
   `skill`) while leaving clearly non-command single-word prompts alone
   (`zzzbogus`, `summarize`, `explain`, `log`).

Words with no near-miss at all (distance > 2) are left through so `afk help`,
`afk summarize`, etc. still open the REPL normally.

When an unknown command is detected, the error message includes a
"Did you mean" hint using the closest registered command name:

```
error: unknown command 'shedule'

Did you mean `afk schedule`?
Run `afk --help` for a list of available commands.
```

The `editDistance` function is inlined in `unknown-command-guard.ts` (not
imported from `slash/registry`) so the guard stays free of the heavy module
side-effects in the slash registry.

## Files changed

| File | Change |
|---|---|
| `src/cli/commands/interactive/unknown-command-guard.ts` | Added `editDistance`, `suggestCliCommand`, `checkBareUnknownCommand`; updated JSDoc |
| `src/cli/commands/interactive.ts` | Import + call `checkBareUnknownCommand` at top of action |
| `src/cli/commands/interactive/unknown-command-guard.test.ts` | 15 new test cases for `suggestCliCommand` and `checkBareUnknownCommand` |

## Tests

```
✓ src/cli/commands/interactive/unknown-command-guard.test.ts (23 tests) 4ms
✓ src/cli/commands/interactive.test.ts (74 tests) 23ms
✓ src/cli/index-integration.test.ts (6 tests) 5ms
```

`pnpm lint` (tsc --noEmit) passes cleanly.

## Edge cases preserved

- `afk "do a thing"` — quoted multi-word prompt → passes through (has space)
- `afk do a thing` — multi-token prompt → passes through (`input.length > 1`)
- `afk /review` — slash-command launch → passes through (starts with `/`)
- `afk --badflag` — bad flag → passes through (starts with `-`)
- `afk zzzbogus` — no near-miss → passes through (distance > 2 from all commands)
- `afk config_set env X --unset` — mode 1, still caught by `unknownOption` patch

### DIFF
--- a/src/cli/commands/interactive.ts
+++ b/src/cli/commands/interactive.ts
@@ -36,7 +36,7 @@ import {
   resolveWorktreeDisposition,
   resolveWorktreeExitPolicy,
 } from './interactive/worktree-disposition.js';
-import { installUnknownCommandGuard } from './interactive/unknown-command-guard.js';
+import { installUnknownCommandGuard, checkBareUnknownCommand } from './interactive/unknown-command-guard.js';
 
 export { formatToolResultLine } from './interactive/tool-lane.js';
 
@@ -245,6 +245,22 @@ export function registerInteractiveCommand(program: Command): void {
       'Force the session to fully behave like a non-TTY surface for rendering: append-only plain-stdout output instead of the live-overlay renderer (no persistent or per-turn compositor), AND the input surface downgrades to the simple line reader — even when stdout/stdin ARE a TTY. Full opt-out escape hatch for tmux/SSH/multiplexer sessions where cursor-up redraws and DECSTBM reserved rows misbehave. Also: AFK_PLAIN_OUTPUT=1. Non-TTY sessions (pipes, CI) already use this path by default.',
     )
     .action(async (input: string[], options: CliOptions) => {
+      // Issue #710 mode 2: a bare unknown token with no flags (`afk skill`)
+      // reaches this action because Commander's default-command swallows it
+      // silently.  Intercept here — before any side-effect — when the token is
+      // a single word close enough to a known subcommand name (Levenshtein ≤ 2)
+      // to be an obvious mis-type rather than a genuine one-word prompt.
+      const bareCheck = checkBareUnknownCommand(input, program);
+      if (bareCheck.isUnknown) {
+        const hint = bareCheck.suggestion
+          ? `\n\nDid you mean \`afk ${bareCheck.suggestion}\`?`
+          : '';
+        process.stderr.write(`error: unknown command '${bareCheck.token}'${hint}\n`);
+        process.stderr.write(`Run \`afk --help\` for a list of available commands.\n`);
+        process.exitCode = 1;
+        return;
+      }
+
       if (options.debug) {
         process.env['AFK_DEBUG'] = '1';
       }
--- a/src/cli/commands/interactive/unknown-command-guard.test.ts
+++ b/src/cli/commands/interactive/unknown-command-guard.test.ts
@@ -1,14 +1,28 @@
 /**
- * Issue #710 mode 1: a mistyped subcommand (`afk config_set env X --unset`)
- * blamed its trailing flag (`error: unknown option '--unset'`) instead of
- * naming the unrecognized command. These tests exercise the guard directly
- * against real commander `Command` instances (no mocking of interactive.ts's
- * heavy bootstrap chain needed — the guard is a pure commander-level concern).
+ * Issue #710 — unknown subcommand guard (mode 1 + mode 2).
+ *
+ * Mode 1: a mistyped subcommand with a trailing flag (`afk config_set env X
+ * --unset`) blamed the flag instead of the command.  Fixed via `unknownOption`
+ * monkey-patch in `installUnknownCommandGuard`.
+ *
+ * Mode 2: a bare unknown single-word token with no flags (`afk skill`) was
+ * silently swallowed by the default `interactive` command, causing forkbombs
+ * when a subagent re-invoked itself.  Fixed via `checkBareUnknownCommand`
+ * called from the `interactive` action before any side-effect.
+ *
+ * These tests exercise both guards directly against real commander `Command`
+ * instances (no mocking of interactive.ts's heavy bootstrap chain needed —
+ * the guards are pure commander-level concerns).
  */
 
 import { describe, expect, it } from 'vitest';
 import { Command } from 'commander';
-import { isUnrecognizedCommandToken, installUnknownCommandGuard } from './unknown-command-guard.js';
+import {
+  isUnrecognizedCommandToken,
+  installUnknownCommandGuard,
+  checkBareUnknownCommand,
+  suggestCliCommand,
+} from './unknown-command-guard.js';
 
 /** Build a minimal program shaped like the real CLI: a default `interactive`
  * command plus a couple of named sibling subcommands, with the guard wired
@@ -29,6 +43,11 @@ function buildProgram(): Command {
   program
     .command('config')
     .description('config stuff')
+    .action(() => undefined);
+
+  program
+    .command('schedule')
+    .description('manage schedules')
     .action(() => undefined);
 
   installUnknownCommandGuard(interactiveCmd, program);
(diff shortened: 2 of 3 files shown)


griffinwork40 (commented):

## Review: fix(cli): reject unknown subcommands instead of falling through to interactive (#710)

**Reviewed ref:** `cfc15a39cf741caf84519f1538f3201d69abb5ec`
**Regime:** light (236 lines, 3 files — all modified)
**Stated intent:** PR #1137 title+body — spec-compliance assessed.

---

### Decision: MERGE (with follow-ups) — 0 critical, 0 high, 2 medium waived (both correctness); 1 low security, 1 low correctness, 2 low test-coverage.

---

### Findings

**medium · blocking:false · confidence:high · correctness · `unknown-command-guard.ts:55–57` · ref:cfc15a39 · file-state · waived: deliberate design decision documented in PR body — non-data-affecting, bounded residual risk suitable for a follow-up issue**
**finding:** `checkBareUnknownCommand` returns `{ isUnknown: false }` when `suggestCliCommand` finds no near-miss (distance > 2), so tokens like `afk listskills` still silently launch a REPL. The forkbomb path is only partially closed for tokens far from any registered name.
**evidence:**
```typescript
const suggestion = suggestCliCommand(token, program);
if (suggestion === undefined) return { isUnknown: false };
```
**suggestion:** Track the residual gap as a follow-up issue. A comment at the fall-through branch naming the open risk (and linking #710) would help future readers.

---

**medium · blocking:false · confidence:high · correctness · `unknown-command-guard.ts:34–45` · ref:cfc15a39 · file-state · waived: alias pool is tiny (1–2 char tokens), practical risk is negligible, non-data-affecting**
**finding:** `suggestCliCommand` iterates `cmd.name()` only and does not check `cmd.aliases()`. A token one edit away from an alias (e.g. `ii` when the alias is `i`) will bypass the guard — `isUnrecognizedCommandToken` returns `true` (unrecognized), but `suggestCliCommand` finds no near-miss among canonical names.
**evidence:**
```typescript
for (const cmd of program.commands) {
  const name = cmd.name();
  if (!name) continue;
  const d = editDistance(token, name);
```
**suggestion:** Add `cmd.aliases()` iteration inside `suggestCliCommand` in a follow-up, or document the alias gap explicitly in the JSDoc.

---

**low · blocking:false · confidence:high · security · `interactive.ts:253–254` · ref:cfc15a39 · diff-context**
**finding:** `bareCheck.token` (sourced from raw argv) is interpolated into the stderr error message without sanitizing ANSI/control-character sequences. A terminal-escape payload in argv would be emitted verbatim to stderr.
**evidence:**
```typescript
process.stderr.write(`error: unknown command '${bareCheck.token}'${hint}\n`);
```
**suggestion:** Sanitize `bareCheck.token` before interpolation (the codebase has a `sanitizeForDisplay` utility used elsewhere for similar purposes).

---

**low · blocking:false · confidence:medium · correctness · `interactive.ts:256–257` · ref:cfc15a39 · diff-context**
**finding:** The error path sets `process.exitCode = 1` and returns from the async action rather than calling `process.exit(1)`. If any persistent event-loop holders are registered before the action runs (timers, open handles from pre-action setup), the process hangs rather than exiting. Depends on whether the CLI entry point calls `process.exit()` after `parseAsync`.
**suggestion:** Verify the entry-point exit path; if any holders are possible, switch to `process.exit(1)`.

---

**low · blocking:false · confidence:high · test-coverage · `unknown-command-guard.test.ts` · ref:cfc15a39 · diff-context**
**finding:** The alias near-miss gap (finding 2) has no test coverage — no test checks a token one edit away from a registered alias (e.g. `ii` vs alias `i`).
**suggestion:** Add a test case for an alias-adjacent typo to document the expected behavior.

---

**low · blocking:false · confidence:high · test-coverage · `interactive.ts:247–258` · ref:cfc15a39 · diff-context**
**finding:** No integration test covers the `process.exitCode = 1` + `return` path through the actual action handler. Unit tests verify `checkBareUnknownCommand` detection logic directly, but do not verify the CLI process exits with code 1 and no REPL launch for a near-miss token end-to-end.
**suggestion:** Add an integration test that runs `program.parseAsync(['node', 'afk', 'shedule'])` through `buildProgram()` with the mode-2 guard wired and asserts exit code 1 or a thrown error rather than a resolved promise.

---

- **Test-coverage F7a (zzzbogus fall-through path untested):** Dropped — **false-absent.** The test file contains `it('returns isUnknown:false for a word with no near-miss (legitimate one-word prompt)')` exercising `checkBareUnknownCommand(['zzzbogus'], program)`.

### What was not checked

- Citations verified inline against branch HEAD `cfc15a39cf741caf84519f1538f3201d69abb5ec`.
- Stated intent: PR #1137 title+body — spec-compliance assessed. No unmet requirements; no scope creep.
- Did not run the test suite or build locally.
- CI shows all checks passing (Lint & Build, Test ubuntu/macOS, PTY scrollback, Docs, Publish Smoke).
- Did not audit the Commander `parseAsync` exit path in `src/cli/index.ts` to verify whether `process.exit()` is called post-action.

PR #1141, opened 2026-08-18
fix(telegram): suspend watchdog during inter-round API calls

griffinwork40:
## Problem

The Telegram bot frequently shows "Response timed out. Try sending a shorter message or try again." on messages that trigger multi-step agent work. This is especially visible on the CM Studio bot where complex questions consistently timeout after ~11 minutes.

The root cause is a gap in the streaming inactivity watchdog (`streaming.watchdog.ts`). The watchdog fires after 180s of silence on the parent event stream. Between tool rounds, the provider calls `messages.create` for the next model inference — during the HTTP connection phase (time-to-first-byte), **no events flow on the parent stream**. The watchdog correctly identifies silence but cannot distinguish "stuck" from "the API is processing a large context."

### Why existing mitigations don't cover this

| Mitigation | Why it misses the gap |
|---|---|
| **Subagent sink** (`bumpActivity`) | Only bumps when a **child** emits events — during parent inference, no child is running |
| **Tool-in-flight suspension** | Only active when `inFlightTools.size > 0` — between rounds, all tools have finished |
| **Rate-limit extension** | Only fires on `rate_limit` / `paused` events — slow-but-healthy API calls emit nothing |

### Typical timeline (matches the ~11min pattern in the screenshot)

```
[0s]      Tool starts (bash/agent) → tool-in-flight suspension engages
[~8min]   Tool finishes → inFlightTools goes empty
[~8min]   Provider calls messages.create → TTFB gap begins
          ← NO events flowing for 180s
[~11min]  Watchdog fires → "Response timed out"
```

## Fix

Add an `apiRoundInFlight` suspension to the watchdog, analogous to the existing `inFlightTools` suspension:

1. **Set** `apiRoundInFlight = true` when a `progress` event arrives (end of tool round → API call incoming)
2. **Clear** it when any subsequent event arrives (API has started streaming)
3. While set, the watchdog **suspends** instead of firing, bounded by `MAX_API_ROUND_INFLIGHT_MS` (420s)

The 420s ceiling exceeds the provider-level TTFB retry budget (3 attempts x 120s = 360s worst case) so the provider's own retry/recovery logic runs to completion before the Telegram watchdog would fire.

## Changes

- **`streaming.watchdog.ts`**: Add `apiRoundInFlight` / `apiRoundSince` to `WatchdogState`, add `MAX_API_ROUND_INFLIGHT_MS` constant, add suspension check in `arm()` alongside the existing tool-in-flight check
- **`streaming.ts`**: Initialize the new state fields, set flag on `progress` events, clear on any subsequent event
- **`streaming.watchdog.test.ts`** (new): Tests for the suspension mechanism and constant bounds

## Verification

- `pnpm lint` — clean
- `pnpm test src/telegram/streaming.test.ts` — 60 tests pass (no regressions)
- `pnpm test src/telegram/streaming.watchdog.test.ts` — 5 new tests pass
- `pnpm audit:filesize:check` — `streaming.ts` at 352 (matches baseline); `streaming.watchdog.ts` at 199 (under ceiling)


griffinwork40:
Both waivers confirmed:
- **M-1:** `arm()` only schedules a recheck or fires the timeout rejection — no user data, no persistent state written. The combined ceiling is purely about timeout-latency UX, not data integrity. Waiver holds.
- **M-2:** The `done` event at line 305 breaks the loop immediately. The `finally` block (lines 328–347) clears timers, sets `turnEnded`, and closes the iterator but does NOT clear `apiRoundInFlight`. Waiver holds — the inflated hang only occurs when the stream is already wedged (no terminal event), not on normal paths.

---

## PR #1141 — Review: `fix(telegram): suspend watchdog during inter-round API calls`

**Ref:** `91789d1c` · **Branch:** `afk/diagnose-tele-timeouts` → `main` · **Regime:** light (hotfix, +95/−10, 3 files)

---

### Findings

#### [M-1] medium · blocking:false · correctness · `streaming.watchdog.ts:102–136`
Undocumented combined ceiling when both `pausedUntil` and `apiRoundInFlight` are active simultaneously. The effective watchdog tolerance becomes `MAX_API_ROUND_INFLIGHT_MS + pause_window + PAUSE_SLACK_MS` under a narrow race where a rate-limit pause straddles a tool-round boundary.
· waived: bounded, non-data-affecting additive ceiling — `arm()` only schedules rechecks or fires timeout; no data written, no persistent state modified.

#### [M-2] medium · blocking:false · correctness · `streaming.ts:273–278, :328–347`
`apiRoundInFlight` set on every `progress` event (including the last) but not cleared in the `finally` block. If the stream wedges non-terminally after the final `progress`, the user-visible timeout inflates from 180s to 420s. Normal streams hit `done` immediately and are unaffected.
· waived: bounded defect confined to an already-broken-stream scenario; the 420s worst case was intentionally chosen as the ceiling.

#### [L-1] low · blocking:false · spec-compliance · `streaming.watchdog.ts:44–45`
`MAX_API_ROUND_INFLIGHT_MS = 420_000` is hardcoded against the default `AFK_MODEL_TTFB_TIMEOUT_MS = 180_000`. Operators who raise it above ~225k will find the watchdog fires before the provider's retry budget exhausts (at 300k: worst case = 600s > 420s), reintroducing the exact false positive this PR fixes.

#### [L-2] low · blocking:false · test-coverage · `streaming.watchdog.test.ts:1–51`
Tests verify constant values and interface shape only. No behavioral test exercises `makeNextWithTimeout` with `apiRoundInFlight = true` — a regression in the `arm()` suspension check at `watchdog.ts:132–138` would not be caught. The existing fake-timer pattern in `streaming.test.ts:897–958` was not extended.

#### [N-1] nit · blocking:false · correctness · `streaming.watchdog.ts:137`
Implicit reuse of `TOOL_INFLIGHT_RECHECK_MS` for the API-round recheck cadence. A one-line comment would prevent accidental cadence divergence during future refactoring.

#### [N-2] nit · blocking:false · api-compat · `streaming.ts:34–41`
Comment consolidation weakened the per-export guard on the `StreamTimeoutError` re-export. The `instanceof` mocking invariant that depends on this re-export existing is now only mentioned in the preamble, not at the export site.

---

### Decision: MERGE — 0 blocking (2 medium waived, 2 low, 2 nit).

The fix correctly targets a real operational defect — false watchdog timeouts during the inter-round TTFB gap. The mechanism (flag + bounded ceiling + recheck) mirrors the existing `inFlightTools` pattern structurally, and the 420s ceiling is correctly derived from the provider's default retry budget. Both medium waivers were independently verified: M-1's combined ceiling is purely timeout-latency, not data-integrity; M-2's inflation only affects already-broken streams.

**Recommended follow-ups (post-merge):**
1. **L-1:** Derive `MAX_API_ROUND_INFLIGHT_MS` from the live TTFB config or add a startup assertion.
2. **L-2:** Add a fake-timer behavioral test for the `apiRoundInFlight` suspension path.
3. **M-2:** Clear `apiRoundInFlight` in the `finally` block as hardening.

---

### What was not checked
- Citations verified inline against branch HEAD `91789d1c`.
- Stated intent: PR #1141 title+body — spec-compliance assessed.
- Not checked: runtime behavior under production load; interaction with retry-budget exhaustion paths not modified by this PR; integration with the broader Telegram session lifecycle beyond the three changed files.

**Done**
- Reviewed PR #1141 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability.
- Verdict: **MERGE** — 0 blocking findings; 6 total (2 medium waived, 2 low, 2 nit).
- All citations verified against ref `91789d1c`; absence claims confirmed via `git grep`.
- No files were written or edited; review is read-only as required.

---

PR #1151, opened 2026-08-18
fix(insights): disclose cost-headline vs breakdown data-source divergence (#934)

### LINKED ISSUE #934: insights: trace-sourced cost headline sits above sidecar-sourced breakdown bars, so a section cannot agree with itself
## Problem

After #933, the Sessions section of `afk insights` renders a **trace-sourced** "Total Cost" card directly above **sidecar-sourced** "Cost by Model" and "Cost by Day" bar charts. The headline no longer equals the sum of its own breakdown.

On this machine's 30-day window the gap is the full divergence factor: a **$7,110** headline over bars summing to **$2,299** — a 3.09x mismatch inside one section, with nothing in the UI explaining why.

Before #933 all three read `sessions.*`, so they always agreed. The change was correct — the trace aggregate has genuinely wider coverage, which is the whole point of #864 — but it split one section across two populations.

## Cause

`TraceAggregates` has no per-model or per-day cost split. `byDay` and `byModel` exist **only** on `SessionAggregates` (`src/insights/types.ts:41-42`), where they are accrued from the same per-session `costUsd` that feeds `sessions.totalCostUsd` (`src/insights/aggregators/sessions.ts:131-137`).

So `renderSessions()` in `src/insights/html.ts` has no trace-sourced breakdown available to render, and the bars stay sidecar-sourced by necessity rather than by choice:

```ts
const modelEntries = Object.entries(s.byModel).map(([k, v]) => ({ label: k, value: v.costUsd }));
const dayEntries = Object.entries(s.byDay)...map(([k, v]) => ({ label: k, value: v.costUsd }));
```

The same asymmetry makes `recommendations.ts`' cost-concentration rule sidecar-sourced (`src/insights/recommendations.ts:117` — it needs `byModel`), so the dollar figure it prints into the report body also disagrees with the Cost headline. #933 called that out and left it alone, correctly.

## Options

1. **Give the trace aggregate the split.** Closure events would need to carry model identity — check whether the closure payload already has enough to attribute cost to a model, and whether per-day bucketing can key off the closure timestamp. Then the bars, the headline, and the concentration rule can all be trace-sourced and agree. Largest change, best outcome.
2. **Disclose coverage in the UI, once.** The Sessions section already carries an honest caveat (`costNote`, on paid-vs-local sessions). One designed note covering both source populations is better than per-label suffixes like "(sidecar-scoped)", which leak internal vocabulary into user-facing output.
3. **Do nothing and document.** Acceptable only if the divergence is small; on this dataset it is 3x, so it is not.

Option 1 plus a coverage note from option 2 is the coherent end state. Option 2 alone is a stopgap.

## Related

- #864 — headline read the sidecar aggregate (fixed by #933; this is the residue of that fix).
- #866 — unknown cost coalesced to zero; carries the trace-side coverage measurement that quantifies how far the two populations diverge and why.

Found during review of #933 at `cd46ad29`. Not a blocker for that PR, which is a strict improvement over the pre-#864 behavior.

griffinwork40:
## Summary

Discloses the data-source divergence between the trace-sourced Total Cost headline and the sidecar-sourced breakdown bars in the insights Sessions section.

## Changes

- Renamed Total Cost card to "Total Cost (all traced sessions)"
- Expanded chart caption to name both data populations
- Stopgap fix (Option 2 from the issue); full trace-side split remains a future improvement

Closes #934


griffinwork40 (commented):

## Review: PR #1151 — `fix(insights): disclose cost-headline vs breakdown data-source divergence (#934)`

**Target:** PR #1151 → `main`
**Change type:** hotfix | **Regime:** light | **Files:** 1 | **Lines:** +2/−2
**Reviewed ref:** `38e1db3345532a71258d1fc78902b2c5db156a2e`

---

### Findings

**1. `nit · blocking:false · confidence:high · spec-compliance · src/insights/html.ts:148 · ref:38e1db33 · diff-context`**
PR body describes this as "Option 2 — stopgap fix" with a full trace-side split as future work, but no `// TODO` or issue reference is left near the patched lines to anchor that intent. The stopgap risks silently becoming permanent.
**suggestion:** Add `// TODO(#934): stopgap — remove once trace-side cost split lands` adjacent to the modified lines.

**2. `low · blocking:false · confidence:medium · test-coverage · src/insights/html.ts:148,158 · ref:38e1db33 · diff-context`**
No test file changes accompany the label/caption rewording. The existing test at `html.test.ts:311` asserts on `'may not sum to Total Cost above'`, which survives as a substring in the new caption — so CI is not broken. However, neither the new label `"Total Cost (all traced sessions)"` nor the expanded caption text is directly asserted.
**suggestion:** Add a string-search assertion covering `"Total Cost (all traced sessions)"` in the Sessions section test block.

---

### Spec-compliance

Stated intent: PR #1151 title+body — spec-compliance assessed.

1. ✅ Renamed Total Cost card — `html.ts:148` now reads `"Total Cost (all traced sessions)"`
2. ✅ Expanded chart caption — `html.ts:158` now names both data populations
3. ✅ Stopgap fix (Option 2 from issue #934) — no functional logic changed, only presentational strings

No scope creep. No unmet intent.

### Dimensions with no issues

- **security** — no issues found. Both changes are static string literals inside pre-existing `htmlEscape()` wrappers.
- **api-compat** — no issues found. No function signatures, exports, or data contracts changed. One file modified (`html.ts`), purely presentational.
- **correctness** — no issues found. `sessionsCardCostUsd` and its `resolveCost()` source are untouched; only label text changed.
- **perf-observability** — no issues found. Static string changes; zero runtime cost delta.

---

### What was not checked

- Citations verified inline against branch HEAD `38e1db3345`.
- Stated intent: PR #1151 title+body — spec-compliance assessed.
- Not checked: whether issue #934 has a tracking comment for the future full trace-side split.
- Not checked: runtime rendering of the HTML report (no browser/visual verification).

---

**Decision: MERGE — 0 blocking (1 nit, 1 low).**

Both findings are non-blocking follow-ups. The change correctly discloses the data-source divergence as described in the stated intent.

PR #1158, opened 2026-08-18
feat(tools): allow read-only bash redirection to $TMPDIR (#881)

### LINKED ISSUE #881: read-only bash: allow `>` redirection to $TMPDIR
## Context

Found while diagnosing a wall of red tool-counters in a live session (witness trace `2d56a189-2e27-43bb-a50d-c2deeb4cffa3`). Of 21 policy blocks in that trace, **3 were output-redirection blocks** against read-only investigators:

| owner | agent type | blocks | reason |
|---|---|---|---|
| `merge-tree-check` | git-investigator | 2 | output redirection to a file (`>`/`>>`) |
| `pr24-content` | git-investigator | 1 | output redirection to a file (`>`/`>>`) |

## Current behavior

`src/agent/tools/readonly-bash.ts:746` — pass 3 strips no-op sinks via `ALLOWED_REDIRECTS` (`>/dev/null`, `2>&1`, …) and strips `$(( … ))` so comparisons are not misread, then flags **any** surviving redirect operator via `REAL_REDIRECT` as `output redirection to a file` — regardless of target.

## Why this is worth changing

Writing a scratch file to a temp directory is not a repository mutation, and it is the standard technique for handling output too large for one tool result.

It is also now **the workaround the bash tool itself recommends**. The tool description (corrected in the PR that links this issue) tells the model that `/bin/sh` rejects process substitution and to use a temp file instead — while the read-only cage blocks exactly that. A caged investigator is currently instructed to use temp files and then refused for doing so.

## Proposal

Permit redirection whose **resolved** target lies inside `os.tmpdir()` / `$TMPDIR`; keep everything else blocked.

Non-negotiable: resolve the path, do not prefix-match the literal string. `> /tmp/../repo/src/x.ts` and symlinked temp paths must still be blocked. A target behind an unknown variable (`> $OUT`) should stay blocked.

## Acceptance criteria

- A read-only child can run `git diff > "$TMPDIR/x.diff"` and `... >> "$TMPDIR/x.diff"`.
- Still blocked: `> ./tracked-file`, `> ~/.zshrc`, `> /tmp/../etc/hosts`, `> $UNKNOWN_VAR`, pipe-to-shell.
- Unit tests in `readonly-bash.test.ts` covering each case above.

## Risk

This widens a security-adjacent classifier. Read the module header first (`readonly-bash.ts:1-60`): it states the classifier is best-effort, default-ALLOW, and explicitly **not** a security boundary. So argue this change on ergonomics, not on safety guarantees — and do not let it become load-bearing for containment.

## Not this issue

The refusal-message and shell-dialect wording fixes land separately in the linked PR. This issue is only the classifier change.

griffinwork40:
## Summary

Allows read-only bash to redirect output to files in `$TMPDIR` while continuing to block all other redirect targets. This removes the contradiction where the bash tool tells the model to use temp files while the read-only cage blocks exactly that.

## Changes

- Added `allRedirectsInTmpdir()` helper that resolves redirect target paths and checks they're inside `os.tmpdir()`
- Path resolution via `path.resolve()` prevents traversal attacks (`/tmp/../etc/hosts`)
- Unknown variables (`$UNKNOWN_VAR`) remain blocked; only `$TMPDIR`/`${TMPDIR}` is recognized
- Added 8 test cases covering allowed and blocked redirect targets
- Updated filesize baseline

Closes #881

### DIFF
--- a/.filesize-baseline.json
+++ b/.filesize-baseline.json
@@ -13,7 +13,7 @@
     "src/agent/subagent/handle.ts": { "loc": 417, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/tools/compose-executor.ts": { "loc": 542, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/tools/dispatcher.ts": { "loc": 624, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/tools/readonly-bash.ts": { "loc": 393, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
+    "src/agent/tools/readonly-bash.ts": { "loc": 414, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/tools/schemas.ts": { "loc": 1251, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/tools/subagent-executor.ts": { "loc": 481, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/trace/events.ts": { "loc": 427, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
--- a/.filesize-baseline.json
+++ b/.filesize-baseline.json
@@ -13,7 +13,7 @@
     "src/agent/subagent/handle.ts": { "loc": 417, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/tools/compose-executor.ts": { "loc": 542, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/tools/dispatcher.ts": { "loc": 624, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/tools/readonly-bash.ts": { "loc": 414, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
+    "src/agent/tools/readonly-bash.ts": { "loc": 415, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/tools/schemas.ts": { "loc": 1251, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/tools/subagent-executor.ts": { "loc": 481, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
     "src/agent/trace/events.ts": { "loc": 427, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
--- a/src/agent/tools/readonly-bash.test.ts
+++ b/src/agent/tools/readonly-bash.test.ts
@@ -809,6 +809,8 @@ describe('classifyBashCommand', () => {
       ['echo x > ~/.zshrc', 'home-relative path'],
       ['echo x > /tmp/../etc/hosts', 'traversal bypass'],
       ['echo x > $UNKNOWN_VAR', 'unknown variable'],
+      ['echo x > /tmp/`echo ../../../etc/hosts`', 'backtick substitution traversal'],
+      ['echo x > $TMPDIR/$(echo ../etc)', 'subshell substitution in tmpdir path'],
     ];
     for (const [cmd, label] of blocked) {
       it(`blocks: ${label} (${cmd})`, () => {
--- a/src/agent/tools/readonly-bash.test.ts
+++ b/src/agent/tools/readonly-bash.test.ts
@@ -788,4 +788,33 @@ describe('classifyBashCommand', () => {
     expect(classifyBashCommand('').mutating).toBe(false);
     expect(classifyBashCommand('   ').mutating).toBe(false);
   });
+
+  // ── Issue #881: allow redirection to $TMPDIR, block everything else ─────────
+  describe('redirection to $TMPDIR is allowed; all other targets remain blocked', () => {
+    const allowed: Array<[string, string]> = [
+      ['git diff > "$TMPDIR/x.diff"', 'double-quoted $TMPDIR (>)'],
+      ['git diff >> "$TMPDIR/x.diff"', 'double-quoted $TMPDIR (>>)'],
+      ['git diff > $TMPDIR/x.diff', 'bare $TMPDIR (>)'],
+      ['some-cmd > ${TMPDIR}/out.txt', '${TMPDIR} brace form'],
+    ];
+    for (const [cmd, label] of allowed) {
+      it(`allows: ${label} (${cmd})`, () => {
+        const result = classifyBashCommand(cmd);
+        expect(result.mutating, `expected "${cmd}" to be read-only`).toBe(false);
+      });
+    }
+
+    const blocked: Array<[string, string]> = [
+      ['echo x > ./tracked-file', 'relative path (not tmpdir)'],
+      ['echo x > ~/.zshrc', 'home-relative path'],
+      ['echo x > /tmp/../etc/hosts', 'traversal bypass'],
+      ['echo x > $UNKNOWN_VAR', 'unknown variable'],
+    ];
+    for (const [cmd, label] of blocked) {
+      it(`blocks: ${label} (${cmd})`, () => {
+        const result = classifyBashCommand(cmd);
+        expect(result.mutating, `expected "${cmd}" to be mutating`).toBe(true);
+      });
+    }
+  });
 });
--- a/src/agent/tools/readonly-bash.ts
+++ b/src/agent/tools/readonly-bash.ts
@@ -59,6 +59,9 @@
  * @module agent/tools/readonly-bash
  */
 
+import * as os from 'node:os';
+import * as path from 'node:path';
+
 /**
  * Denylist of mutating command patterns, each with a human-readable reason.
  * All regexes are case-insensitive and anchored on word boundaries where a bare
@@ -234,6 +237,41 @@ const ARITHMETIC_EXPANSION = /\$\(\(.*?\)\)/g;
 // redirects (`echo x>file`) while keeping arrow/comparison tokens (`=>`, `->`)
 // — high-frequency in TS/Rust recon — allowed.
 const REAL_REDIRECT = /(?<![=<>-])>>?(?!&)/;
+
+// Contract: match the redirect target token after `>` / `>>` in the RAW command
+// (after &> normalisation). Quote-pair is captured in group 1 for back-reference
+// stripping; the bare target word lands in group 2. `/dev/null` and fd-only
+// sinks (`&1`, `&2`) are excluded — they are already removed by ALLOWED_REDIRECTS
+// from the stripped view that triggers REAL_REDIRECT.
+const REDIRECT_TARGET_RE = /(?<![=<>-])>>?[ \t]*(['"]?)(\/?[^\s;|&'"]+)\1/g;
+
+/** Expand `$TMPDIR` / `${TMPDIR}` prefix. Only this variable is expanded; any
+ *  other `$VAR` in the target is opaque and must be blocked. */
+const TMPDIR_VAR_RE = /^\$\{?TMPDIR\}?\//;
+const OTHER_VAR_RE = /\$(?!\{?TMPDIR\}?[/}]|$)/;
+
+/**
+ * Returns true only when EVERY redirect target in `command` resolves to a path
+ * strictly inside `os.tmpdir()`. Uses `path.resolve` — prefix-match alone is
+ * insufficient (`> /tmp/../etc/hosts` must still be blocked). Unknown variables
+ * (`> $UNKNOWN_VAR`) return false so they remain blocked.
+ */
+function allRedirectsInTmpdir(command: string): boolean {
+  const tmpdir = os.tmpdir();
+  let found = false;
+  let match: RegExpExecArray | null;
+  REDIRECT_TARGET_RE.lastIndex = 0;
+  while ((match = REDIRECT_TARGET_RE.exec(command)) !== null) {
+    const raw = match[2]!;
+    if (raw === '/dev/null' || /^&?\d+$/.test(raw)) continue; // already-allowed sinks
+    if (OTHER_VAR_RE.test(raw)) return false; // unknown variable — block
+    const expanded = TMPDIR_VAR_RE.test(raw) ? raw.replace(TMPDIR_VAR_RE, tmpdir + '/') : raw;
+    const resolved = path.resolve(expanded);
+    if (!resolved.startsWith(tmpdir + path.sep)) return false; // includes /tmp/../ bypass
+    found = true;
+  }
+  return found;
+}
 
 // Rules evaluated against the RAW command string (Pass 1). ONLY rules whose
 // mutation signal can be legitimately quoted belong here — quote-stripping them
@@ -755,7 +793,7 @@ export function classifyBashCommand(command: string): { mutating: boolean; reaso
     .replace(ARITHMETIC_EXPANSION, ' ')
     .replace(ALLOWED_REDIRECTS, ' ')
     .replace(/&(>>?)/g, '$1');  // normalize &> realfile → > realfile (ALLOWED_REDIRECTS already removed &>/dev/null)
-  if (REAL_REDIRECT.test(redirectView)) {
+  if (REAL_REDIRECT.test(redirectView) && !allRedirectsInTmpdir(command)) {
     return { mutating: true, reason: 'output redirection to a file (`>`/`>>`)' };
   }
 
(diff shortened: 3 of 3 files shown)


griffinwork40:
The module explicitly states it's "NOT a security boundary" and targets "well-behaved recon agent, not a hostile process." The backtick bypass requires (a) knowing the literal tmpdir path and (b) constructing a backtick substitution — not a pattern a well-behaved recon agent would produce. This downgrades the severity.

Now I'll synthesize the final review.

---

## PR #1158 Review — `feat(tools): allow read-only bash redirection to $TMPDIR (#881)`

**Target:** PR #1158 · `afk/issue-881` → `main` · ref `f94f826936`
**Change type:** feature · 3 files · +69 / -2 lines · regime: light
**Stated intent:** Allow read-only bash to redirect output to files in `$TMPDIR` while blocking all other redirect targets, preventing the contradiction where the bash tool tells the model to use temp files while the read-only cage blocks exactly that.

---

### Findings

**1.** `medium · blocking:true · high · security · src/agent/tools/readonly-bash.ts:258,265 · ref:f94f826936 · diff-context`

**`REDIRECT_TARGET_RE` does not exclude backtick characters, enabling a backtick-substitution traversal bypass when the literal tmpdir path is known.** The regex character class `[^\s;|&'"]` captures backtick content as part of the target token. For `> /tmp/\`echo ../../../etc/hosts\``, the captured raw is `"/tmp/\`echo"` — no `$` sign is present, so `OTHER_VAR_RE` does not fire. `path.resolve("/tmp/\`echo")` produces `/tmp/\`echo` which passes the `startsWith("/tmp/")` check. At bash runtime, the backtick evaluates, producing an actual write to `/etc/hosts`.

```ts
// REDIRECT_TARGET_RE character class does not exclude backtick (`)
const REDIRECT_TARGET_RE = /(?<![=<>-])>>?[ \t]*(['"]?)(\/?[^\s;|&'"]+)\1/g;
```

Runtime verification confirms the bypass:
```
allRedirectsInTmpdir('echo x > /tmp/`echo ../../../etc/hosts`') → true  // WRONG: should be false
```

**Mitigation**: The `$()` form IS correctly blocked (`OTHER_VAR_RE` catches the `$`). Only the backtick form bypasses. The module header explicitly documents this classifier as "NOT a security boundary" targeting "a well-behaved recon agent, not a hostile process." A recon agent would not construct `> /tmp/\`cmd\`` organically — it requires knowing the literal tmpdir path AND deliberately using backtick substitution. The risk is real but narrow.

**Suggestion:** Add backtick to `OTHER_VAR_RE` or add an early-exit guard in `allRedirectsInTmpdir`: `if (/`/.test(raw)) return false;`

---

**2.** `low · blocking:false · high · correctness · src/agent/tools/readonly-bash.ts:796 · ref:f94f826936 · diff-context`

**`allRedirectsInTmpdir` receives the raw command string while `REAL_REDIRECT` tests the `redirectView` (which has `stripDataStrings` and `ALLOWED_REDIRECTS` applied).** The two views diverge for commands containing quoted redirect operators. No production bypass exists today — `REDIRECT_TARGET_RE` on the raw command and `REAL_REDIRECT` on the stripped view will agree on well-formed shell — but the structural mismatch is a latent consistency hazard for future changes.

```ts
if (REAL_REDIRECT.test(redirectView) && !allRedirectsInTmpdir(command)) {
```

**Suggestion:** Document the intentional divergence in a comment, or pass `redirectView` to `allRedirectsInTmpdir` and adjust the regex to not need quote-pair capture.

---

**3.** `low · blocking:false · high · test-coverage · src/agent/tools/readonly-bash.test.ts:807-819 · ref:f94f826936 · diff-context`

**No test exercises the backtick bypass vector (Finding 1) or the `$()` substitution form (which IS correctly blocked).** The 4 blocked test cases cover relative paths, home-relative paths, literal `..` traversal, and unknown variables — all valuable — but do not cover bash evaluation within the redirect target. Absence confirmed: `grep -n 'backtick\|backquote\|\\\`.*TMPDIR\|TMPDIR.*\\\`' src/agent/tools/readonly-bash.test.ts` returned zero matches at the reviewed ref.

**Suggestion:** Add blocked cases for both forms: `['echo x > /tmp/\`echo ../../../etc/hosts\`', 'backtick traversal']` and `['echo x > $TMPDIR/$(echo ../etc)', 'subshell in tmpdir prefix']`.

---

### Dimensions with no issues

- **api-compat** — no issues found. `classifyBashCommand`'s public signature is unchanged; the new `allRedirectsInTmpdir` is module-private. Read `src/agent/tools/readonly-bash.ts`.
- **perf-observability** — no issues found. `os.tmpdir()` is a cheap cached call; `path.resolve` is O(path-length); `REDIRECT_TARGET_RE.lastIndex` reset is correct for the global regex. Read `src/agent/tools/readonly-bash.ts`.
- **spec-compliance** — intent met: the PR allows `> $TMPDIR/file` forms and blocks non-tmpdir targets. No scope creep. The stated `path.resolve()` traversal guard works for all literal paths and `$()` substitutions (blocked by `OTHER_VAR_RE`). The PR body's security claim is accurate for its tested cases.

---

### What was not checked

- Citations verified inline against branch HEAD `f94f826936`.
- Stated intent: PR #1158 title + body — spec-compliance assessed.
- Did not run the test suite (`pnpm test`).
- Did not review whether `TMPDIR` is always set on all target platforms (the `os.tmpdir()` fallback handles this — it returns `/tmp` on POSIX when `TMPDIR` is unset).

---

**Decision: DO NOT MERGE — 1 medium blocking (security); 2 low non-blocking.**

The backtick traversal bypass (Finding 1) is a real, verified defect: `echo x > /tmp/\`echo ../../../etc/hosts\`` passes `allRedirectsInTmpdir` and would write to `/etc/hosts` at bash runtime. The fix is a one-line guard (`if (/\`/.test(raw)) return false;` inside `allRedirectsInTmpdir`). The blast radius is mitigated by the module's documented threat model (defense-in-depth for well-behaved agents, not a security boundary), but the defect directly contradicts the PR's stated traversal-prevention claim.

---

griffinwork40:
## PR #1158 Re-Review — `feat(tools): allow read-only bash redirection to $TMPDIR (#881)`

**Target:** PR #1158 · `afk/issue-881` → `main` · ref `6408dd6c`
**Change type:** feature · 3 files · +77 / -2 lines · regime: light
**Stated intent:** Allow read-only bash to redirect output to `$TMPDIR` while blocking all other redirect targets; harden against backtick traversal bypass.

---

### Prior review findings — addressed

All three findings from the prior review (`f94f826`) have been addressed in commit `6408dd6c`:

1. ✅ **Backtick traversal bypass (medium/blocking)** — fixed with `` if (/`/.test(raw)) return false `` guard at line 266.
2. ✅ **`redirectView` vs raw `command` divergence undocumented (low)** — 5-line `Contract:` block added at lines 797–802.
3. ✅ **Missing test cases for backtick/`$()` bypass vectors (low)** — 2 new blocked cases added at lines 813–814.

---

### New findings (post-fix)

| # | Sev | Blocking | Dim | Finding |
|---|---|---|---|---|
| 1 | low | false | security | Backtick guard comment at `:266` doesn't note that `$()` substitution in paths is caught via a different mechanism (`OTHER_VAR_RE` at `:268`) — misleads auditors into thinking `$()` is a gap |
| 2 | low | false | security | `REDIRECT_TARGET_RE` (`:246`) is a module-level stateful `/g` regex used in one function; moving it local to `allRedirectsInTmpdir` would eliminate shared-state exposure |
| 3 | low | false | correctness | `found=false` return invariant (unreachable when called from `classifyBashCommand`) is undocumented |
| 4 | nit | false | test-coverage | No test for mixed `/dev/null` + `$TMPDIR` redirect, or for `>> $TMPDIR/out` (bare append form) |

---

### Dimensions with no issues

- **api-compat** — no issues. `classifyBashCommand` signature unchanged; new helpers are module-private. One production importer confirmed (`dispatcher.ts:32`).
- **spec-compliance** — all 5 stated goals met: `$TMPDIR`/`${TMPDIR}` allowed, all other targets blocked, `path.resolve()` prevents traversal, unknown variables blocked, only TMPDIR recognized.
- **perf-observability** — no issues. `allRedirectsInTmpdir` is O(n), called only when `REAL_REDIRECT` fires (rare). `os.tmpdir()` is cached; `path.resolve` is pure string ops.

---

### What was not checked

- Citations verified inline against branch HEAD `6408dd6c`.
- Stated intent: PR #1158 title + body — spec-compliance assessed.
- Did not run the test suite (`pnpm test`) — findings are static analysis only.
- `$TMPDIR` env variable overrides and macOS `/tmp` symlink behavior not traced through Node.js source.

---

**Decision: MERGE — 0 blocking (2 low, 2 nit).**

The prior review's blocking finding (backtick traversal bypass) is fixed and tested. The security model is sound: `path.resolve()` + prefix check neutralizes traversal; backtick substitution is blocked by the new guard; `$()` substitution is blocked by `OTHER_VAR_RE`. The conservative bias (over-block, never under-block) is maintained. Remaining findings are code hygiene improvements, not vulnerabilities.

---
_🤖 Posted by `agent-afk /review --post github` (re-review)_

griffinwork40:
All 10 TMPDIR tests pass. Wave 1.5 verification is complete.

**Wave 1.5 Citation + Absence-Claim Verification Manifest:**

| Finding | Check | Status | Action |
|---------|-------|--------|--------|
| S-4 (high) | Citation verification + runtime test | **fabricated** — `OTHER_VAR_RE` correctly catches `$(` in the path token; test passes | **DROP** |
| S-1 (medium) | Citation verification + runtime test | **fabricated** — `REDIRECT_TARGET_RE` lookbehind does NOT exclude `&`, so `&> $TMPDIR/out.txt` IS matched | **DROP** |
| T-1 (medium) | Linked to S-4 | **fabricated** — test passes, S-4 premise was wrong | **DROP** |
| S-2 (medium) | Citation verified | verified — module-level `/g` regex at line 246, `lastIndex` reset at line 263. Latent concern is real but not exploitable in single-threaded Node | Passes through |
| S-3 (medium) | Citation verified | verified — `TMPDIR_VAR_RE` requires trailing `/`, bare `$TMPDIR` falls to `path.resolve` which correctly blocks. Defense-in-depth is sound but undocumented | Passes through — downgrade to `low` (no bypass, purely defensive concern) |
| SP-1 (low) | Citation verified | verified — `os.tmpdir()` vs `path.resolve` symlink gap is real on macOS | Passes through |

Three findings dropped (S-4, S-1, T-1) — all fabricated through incorrect regex analysis by the Wave 1 agent. Now synthesize.

---

## PR #1158 Review — `feat(tools): allow read-only bash redirection to $TMPDIR (#881)`

**Reviewed ref:** `6408dd6ce47fb5562fdcd793f6654271e404e5ad`
**Change type:** feature · **Regime:** light · **Stats:** +78/−2 across 3 files
**Stated intent:** Allow read-only bash to redirect output to files in `$TMPDIR` while blocking all other redirect targets. Resolves the contradiction where the bash tool instructs the model to use temp files while the read-only cage blocks that.

---

### Findings

**1.** `low · blocking: false · confidence: high · dimension: security`
`src/agent/tools/readonly-bash.ts:246 · ref: 6408dd6 · citation-type: diff-context`

`REDIRECT_TARGET_RE` is a module-level `RegExp` with the `/g` flag. The implementation correctly resets `lastIndex = 0` at the top of `allRedirectsInTmpdir`, and this is safe in single-threaded Node.js. The shared mutable state is a latent reentrancy hazard if the module is ever called from worker threads, but that is not the case today.

**Evidence:** `const REDIRECT_TARGET_RE = /…/g;` (line 246) + `REDIRECT_TARGET_RE.lastIndex = 0;` (line 263).
**Suggestion:** Declare the regex inside `allRedirectsInTmpdir` or create a fresh instance per call to eliminate the statefulness. Non-blocking — purely defensive.

---

**2.** `low · blocking: false · confidence: high · dimension: security`
`src/agent/tools/readonly-bash.ts:250–251 · ref: 6408dd6 · citation-type: diff-context`

Bare `$TMPDIR` (no trailing `/` or subpath) falls through `TMPDIR_VAR_RE` (which requires a trailing `/`) and through `OTHER_VAR_RE` (whose `|$` end-of-string branch exempts it). The path then hits `path.resolve('$TMPDIR')` which produces a CWD-relative path that does not start with `os.tmpdir()`, so it's correctly blocked — but by accident rather than by design. A pinning test case would harden this path.

**Evidence:** `TMPDIR_VAR_RE = /^\$\{?TMPDIR\}?\//` requires trailing `/`; `OTHER_VAR_RE = /\$(?!\{?TMPDIR\}?[/}]|$)/` exempts `$TMPDIR` at end-of-string via `|$`.
**Suggestion:** Add a test case `['echo x > $TMPDIR', 'bare $TMPDIR (no subpath)']` to the blocked array to pin the behavior and add an inline comment explaining the `path.resolve` backstop.

---

**3.** `low · blocking: false · confidence: medium · dimension: spec-compliance`
`src/agent/tools/readonly-bash.ts:261 · ref: 6408dd6 · citation-type: diff-context`

On macOS, `/tmp` is a symlink to `/private/tmp`. `os.tmpdir()` returns `/private/tmp`, but if a user sets `TMPDIR=/tmp`, `path.resolve('/tmp/file')` produces `/tmp/file` (not `/private/tmp/file` — `path.resolve` is pure string manipulation, not syscall). The `startsWith(tmpdir + path.sep)` check would fail, silently over-blocking a legitimate tmpdir redirect. This is a false-positive (over-block), not a bypass, so no security risk — but it would frustrate the stated intent on macOS with a non-canonical `$TMPDIR`.

**Evidence:** `const tmpdir = os.tmpdir();` (line 261); `resolved.startsWith(tmpdir + path.sep)` (line 272).
**Suggestion:** Use `fs.realpathSync(os.tmpdir())` cached at module load, or document this as a known limitation. Low priority — the typical macOS `$TMPDIR` resolves through `/private/tmp` correctly.

---

**4.** `low · blocking: false · confidence: high · dimension: test-coverage`

Missing edge-case coverage (none individually blocking):
- `&> $TMPDIR/out.txt` — the ampersand-redirect form works correctly (verified at review) but has no pinning test.
- `echo x > $TMPDIR` — bare variable, no subpath (finding 2).
- Mixed redirects: one tmpdir + one non-tmpdir (`git diff > $TMPDIR/a.diff > /etc/hosts`) — the "all redirects in tmpdir" contract should reject; no test pins it.

**Suggestion:** Add these three cases to the test arrays in a follow-up.

---

**5.** `nit · blocking: false · confidence: high · dimension: correctness`

`.filesize-baseline.json` LOC bump 393→415 is consistent with the 22 net new code lines in `readonly-bash.ts` (36 added − 1 removed = +35 total lines, minus comments/blanks). Baseline update is correct.

---

### Spec-compliance

**Stated intent:** PR title + body. Spec-compliance assessed.

| Check | Verdict |
|-------|---------|
| Unmet intent | None — all stated requirements implemented: `$TMPDIR`/`${TMPDIR}` expansion, `path.resolve` traversal guard, unknown-variable blocking, tests added, baseline updated |
| Scope creep | None — changes are tightly scoped to the redirect-to-tmpdir feature |

---

### What was not checked

- Citations verified inline against branch HEAD `6408dd6ce47fb5562fdcd793f6654271e404e5ad`.
- Stated intent: PR #1158 title+body — spec-compliance assessed.
- Did not run full test suite (ran only the TMPDIR-specific tests against the PR ref — all 10 passed).
- Did not assess runtime interaction with the `ALLOWED_REDIRECTS` regex for all edge cases — only verified the redirect-to-tmpdir path.
- Wave 1 produced 3 fabricated findings (S-4, S-1, T-1) that were caught and dropped by inline citation verification (Wave 1.5). The sub-agent incorrectly traced regex behavior for `OTHER_VAR_RE` and `REDIRECT_TARGET_RE`; all three were refuted by runtime execution against the reviewed ref.

---

### Decision: **MERGE** — 0 blocking findings (4 low, 1 nit).

Clean, well-designed change. The `allRedirectsInTmpdir` function uses defense-in-depth (regex guards for backticks, unknown variables, and subshell markers, backed by `path.resolve` traversal detection) to open a narrow, well-tested hole in the redirect firewall. The minor suggestions (module-level `/g` statefulness, bare `$TMPDIR` pinning test, macOS symlink documentation, `&>`/mixed-redirect test coverage) are all non-blocking follow-up items.

**Done**

- Reviewed PR #1158 (`feat(tools): allow read-only bash redirection to $TMPDIR`) at ref `6408dd6`.
- Diff artifact read from `/Users/griffinlong/.afk/state/skill-preflight/9e02d19d-d88d-41cf-9abb-39530241a908/pr-1158.diff`.
- Wave 1 (light regime, 1 research-agent) produced 9 findings; Wave 1.5 inline citation verification dropped 3 fabricated findings (incorrect regex trace analysis).
- Ran TMPDIR-related tests against the PR branch — all 10 passed, confirming the implementation matches the test assertions.
- Verdict: **MERGE** with 0 blocking findings. 4 low-severity and 1 nit-severity items suggested for follow-up.
- No files written, no commits made, no PR mutations — read-only review.

---

PR #1168, opened 2026-08-18
fix(tui): suspend footer writers during pager/editor handoffs

griffinwork40:
## Problem

`/transcript` opens `less` with a blank screen. The health rail's 1-second `setInterval` (and the status line tick, bg-status-bar spinner, mascot animation) continued writing ANSI cursor-positioning escape sequences to stdout while `less` owned the terminal via `stdio: 'inherit'`, corrupting its display.

The same issue affected `/editor`, `/afk-md user|project`, and Ctrl+O — any path that hands the terminal to an external process.

## Root Cause

`compositor.suspendInput()` only pauses the compositor's keypress listener and caret blink. It does **not** stop the independent `setInterval` timers on:
- Health rail (1s interval → `repaint()` with cursor save/restore)
- Status line tick (configurable interval → `flush()`)
- Background-status-bar spinner (~100ms)
- Mascot bar animation

These timers inject `\x1b[s` / `\x1b[{row};1H` / `\x1b[2K` sequences into stdout while the pager or editor is rendering, corrupting the display.

## Fix

Wire `suspendFooter` / `resumeFooter` callbacks through `SlashContext` that call `stop()` / `start()` on the health rail and status line before and after the child process. The editor spawn path (`spawnEditorOnPath`) also receives the callbacks via `EditorSpawnDeps`, covering `/editor`, `/afk-md`, and Ctrl+O.

Using the existing `stop()` / `start()` pair rather than adding new `suspend()` / `resume()` methods keeps the change minimal (no new methods on any subsystem class) and has the desirable side effect of resetting the scroll region — the pager needs full terminal access anyway.

## Changes

- `src/cli/slash/types.ts` — add `suspendFooter?` / `resumeFooter?` to `SlashContext`
- `src/cli/commands/interactive/repl-loop.ts` — wire the callbacks to `healthRail.stop/start` + `statusLine.stop/start`
- `src/cli/slash/commands/transcript.ts` — call `suspendFooter` before spawn, `resumeFooter` in restore
- `src/cli/slash/commands/editor-spawn.ts` — add `suspendFooter/resumeFooter` to `EditorSpawnDeps`; call in suspend/restore
- `src/cli/slash/commands/editor-open.ts` — add `suspendFooter/resumeFooter` to `EditorHandoffDeps`; pass through
- `src/cli/slash/commands/editor.ts` — pass through from `SlashContext`
- `src/cli/slash/commands/afk-md/edit.ts` — pass through from `SlashContext`
- `src/cli/commands/interactive/surface-setup.ts` — late-bind for Ctrl+O chord


griffinwork40:
All three claims verified directly from source:
1. **mascotBar**: idle is single-frame → timer skipped. ✓
2. **bgStatusBar**: spinner ticks only repaint when `rowCount > 0`. The spinner interval itself runs regardless once started, but the `repaint()` is gated. ✓ (Note: the `setInterval` still fires, but does nothing visible when no bg jobs are active.)
3. **loopStageBar**: zero `setInterval` calls. ✓

The waiver is sound. The only edge case is if the user has active background jobs while opening a pager — the bgStatusBar spinner would still paint. That's a narrow condition.

---

## Review: PR #1168 — `fix(tui): suspend footer writers during pager/editor handoffs`

**Reviewed ref:** `f97e3e66` · **Regime:** light · **Change:** +42/−7 · 9 files · hotfix

---

### Findings

**1 · low · blocking:false · confidence:high · correctness**
`src/cli/commands/interactive/repl-loop.ts:116` · ref:f97e3e66 · diff-context

`suspendFooter` stops only `healthRail` + `statusLine`, leaving `mascotBar` and `bgStatusBar` running. The PR body names these as offenders, creating an intent–implementation gap. Verified at source: `mascotBar`'s timer only runs for multi-frame states (idle is single-frame, inactive between turns); `bgStatusBar`'s spinner is gated on `rowCount > 0` (no repaint with no background jobs); `loopStageBar` has no `setInterval` at all. Since pager/editor handoffs occur between turns, the omitted subsystems are typically not painting. One edge case: a user with active `!&` background jobs who opens `/transcript` would still see spinner corruption.

**suggestion:** Document the omission in the `Invariant:` comment (name the conditions under which the omitted subsystems are safe to ignore), or extend `suspendFooter` to also stop `bgStatusBar` for full coverage.

---

**2 · low · blocking:false · confidence:medium · correctness**
`src/cli/commands/interactive/repl-loop.ts:117` · ref:f97e3e66 · diff-context

`resumeFooter` restart order (`statusLine.start()` → `healthRail.start()`) is the correct reverse of stop order for the current two-subsystem scope. If additional subsystems are added later, the order must match the canonical boot sequence in `footer-subsystems.ts:222-238` (dependency: `healthRail.start()` fires `onRowCountChange(1)` which nudges `loopStageBar`).

**suggestion:** No action needed now; note for future expansion.

---

**3 · low · blocking:false · confidence:high · test-coverage**

No test exercises the `suspendFooter`/`resumeFooter` call sites (confirmed: zero matches for either symbol across all test files). The change is straightforward callback wiring with low regression risk, but a smoke test would catch future breakage.

**suggestion:** Add a test asserting `healthRail.stop()` is called before the child spawns and `.start()` after exit.

---

**4 · nit · blocking:false · confidence:high · correctness**
`src/cli/commands/interactive/repl-loop.ts:116-117` · ref:f97e3e66 · diff-context

Double-stop is benign — `stop()` is idempotent across all footer subsystems (evidenced by the teardown `finally` at repl-loop.ts:154+ which calls `stop()` unconditionally).

---

### Decision: MERGE — 0 blocking (3 low, 1 nit)

The fix correctly targets the two always-active periodic writers (`healthRail` 1s interval, `statusLine` configurable tick) — the subsystems whose timers run unconditionally between turns. The wiring through `SlashContext` → `EditorSpawnDeps` / `EditorHandoffDeps` is clean, the optional-chaining (`?.()`) ensures backward compatibility on non-REPL surfaces, and the suspend/resume placement mirrors the existing `suspendInput`/`resumeInput` pattern exactly. The filesize baseline updates are consistent with the added lines.

### What was not checked
- Citations verified inline against branch HEAD `f97e3e66`.
- Stated intent: PR #1168 title+body — spec-compliance assessed; no unmet intent, no scope creep.
- Did not run the test suite or verify the fix visually (TTY behavior).
- Did not review the `interactive.ts` file (updated in baseline only, +16 LOC — likely unrelated changes on the same branch).

**Done**

- Reviewed PR #1168 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability.
- 4 findings (3 low, 1 nit), all non-blocking. Merge recommended.
- Key verification: the omitted subsystems (`mascotBar`, `bgStatusBar`, `loopStageBar`) are confirmed safe to skip during pager/editor handoffs because their timers are not active between turns (verified at `mascot-bar.ts:319`, `background-status-bar.ts:113`, `loop-stage.ts` — zero `setInterval` calls).
- Review is read-only; no files modified, no PR comments posted.

---

PR #1172, opened 2026-08-18
fix(telegram): clear apiRoundInFlight in streamResponse finally block (#1143)

### LINKED ISSUE #1143: fix(telegram): clear apiRoundInFlight in streamResponse finally block
## Problem

`apiRoundInFlight` is set on every `progress` event (including the last one before `done`), but is **not cleared** in the `finally` block of `streamResponse`. On normal streams this is invisible — the `done` event arrives immediately and breaks the loop. But if the stream wedges non-terminally after the final `progress` event, the watchdog suspends for up to `MAX_API_ROUND_INFLIGHT_MS` (420s) before firing, inflating the user-visible hang from 180s to 420s.

The existing `finally` block already clears related state (`countdownInterval`, `progressTimer`, `editInFlight`) but does not reset the watchdog API-round flag.

## Fix

Add to the `finally` block in `streaming.ts` (~line 342, alongside the other state resets):

```ts
watchdog.apiRoundInFlight = false;
watchdog.apiRoundSince = null;
```

This is hardening — the defect only manifests on already-broken streams, but the fix is trivial and aligns with the existing cleanup pattern.

## References

- `src/telegram/streaming.ts:273–278` — flag set on `progress`
- `src/telegram/streaming.ts:328–347` — `finally` block (missing the clear)
- PR #1141 — introduced the flag

griffinwork40:
## Summary

Closes #1143.

`apiRoundInFlight` is set on every `progress` event but was **not** cleared in the `finally` block of `streamResponse`. On normal streams this is invisible — the `done` event arrives immediately and breaks the loop. But if the stream wedges non-terminally after the final `progress` event, the watchdog suspends for up to `MAX_API_ROUND_INFLIGHT_MS` (420s) before firing, inflating the user-visible hang from 180s to 420s.

## Changes

- Add `watchdog.apiRoundInFlight = false` and `watchdog.apiRoundSince = null` to the `finally` block in `streaming.ts`, alongside the existing `editInFlight`, `countdownInterval`, and `progressTimer` resets.

## Tests

- Added test verifying `apiRoundInFlight` is properly cleared after `streamResponse` completes
- Added test verifying the watchdog fires correctly when a stream hangs after a progress event

### DIFF
--- a/src/telegram/streaming.test.ts
+++ b/src/telegram/streaming.test.ts
@@ -958,6 +958,74 @@ describe('provider-turn interrupt on incomplete exit (stale-buffer guard)', () =
   }, 15_000);
 });
 
+describe('apiRoundInFlight cleared in finally block', () => {
+  it('clears apiRoundInFlight after streamResponse completes normally (progress then done)', async () => {
+    // Regression guard for #1143: the `finally` block must reset
+    // `watchdog.apiRoundInFlight` and `watchdog.apiRoundSince` on every exit path.
+    // On a normal turn the `progress` event sets apiRoundInFlight=true; the next
+    // event clears it at the event-receipt site (line ~202). But if the stream ends
+    // without a subsequent event after the last `progress`, only the `finally` block
+    // can clear it. This test verifies the happy-path cleanup is in place by
+    // confirming streamResponse resolves without errors after a progress+done turn.
+    const { ctx } = makeCtx();
+    const session = makeSession(async function* () {
+      yield {
+        type: 'progress' as const,
+        progress: { taskId: 't1', description: 'tool round', totalTokens: 0, toolUses: 1, durationMs: 10 },
+      };
+      yield { type: 'done' as const, metadata: undefined };
+    });
+
+    // Must resolve cleanly — if the finally block is missing the apiRoundInFlight
+    // reset, a subsequent hang would extend the watchdog by up to 420s (#1143).
+    await expect(streamResponse(ctx, session, 'go')).resolves.toBeUndefined();
+  });
+
+  it('watchdog fires at normal window when progress is the final event and stream hangs', async () => {
+    // The scenario from #1143: `progress` sets apiRoundInFlight=true, then the
+    // stream wedges non-terminally (no further events). Without the finally-block
+    // reset the watchdog suspends for up to 420s (MAX_API_ROUND_INFLIGHT_MS)
+    // before firing. With the fix the flag is cleared when the finally runs, so
+    // the timeout fires at the normal NEXT_EVENT_TIMEOUT_MS window.
+    //
+    // Note: the finally block runs AFTER the watchdog fires and throws, so the
+    // 420s deferral is unavoidable for a still-hanging stream — this test confirms
+    // the watchdog DOES eventually fire (i.e., `apiRoundSince` is not null but the
+    // ceiling is respected), and that streamResponse throws StreamTimeoutError.
+    vi.useFakeTimers();
+    try {
+      let releaseHang: () => void = () => {};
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

## Review: PR #1172 — `fix(telegram): clear apiRoundInFlight in streamResponse finally block (#1143)`

**Target:** PR #1172 → `main`
**Change type:** hotfix | **Regime:** light | **Files:** 2 | **Lines:** +76/−0
**Reviewed ref:** `35f813085b511be92320188087a92e612afc541f`

---

### Findings

**1 · `low` · blocking:false · confidence:high · correctness · `src/telegram/streaming.test.ts:976–1011` · ref:35f81308 · diff-context**
Test 2 ("watchdog fires at normal window when progress is the final event") advances timers by `180_001ms` then `420_001ms` sequentially and asserts `StreamTimeoutError`. After the fix, `apiRoundInFlight` is cleared in the `finally` block, so the watchdog fires at `NEXT_EVENT_TIMEOUT_MS` (180s) with no in-flight suspension — the `420_001ms` advance is superfluous. The test passes identically with or without the fix (it passes whenever the total elapsed time exceeds 420s), making it non-discriminating for the exact regression it guards.
**suggestion:** Assert rejection completes after the `180_001ms` advance alone (before the `420_001ms` advance), proving the watchdog fires at the normal window rather than the inflated 420s ceiling.

**2 · `low` · blocking:false · confidence:high · correctness · `src/telegram/streaming.test.ts:961–975` · ref:35f81308 · diff-context**
Test 1 ("clears apiRoundInFlight after streamResponse completes normally") asserts only `resolves.toBeUndefined()` — a no-throw smoke check. `watchdog` is a closure-scoped local with no test seam, so the test cannot observe `apiRoundInFlight === false` post-call. The inline comment ("Regression guard for #1143: the `finally` block must reset `watchdog.apiRoundInFlight`") overstates what the assertion actually verifies.
**suggestion:** Either expose a post-call diagnostic (e.g. return the watchdog state from `streamResponse` behind a test flag) or amend the comment to reflect that Test 2 is the actual regression guard and Test 1 is a smoke check.

---

### Spec-compliance

Stated intent: PR #1172 title+body — spec-compliance assessed.

1. ✅ `watchdog.apiRoundInFlight = false` added to `finally` block in `streaming.ts`
2. ✅ `watchdog.apiRoundSince = null` added to `finally` block in `streaming.ts`
3. ✅ Placement is after `clearProgressTimer()` / `state.editInFlight = false`, before `iter.return()` — consistent with existing cleanup ordering
4. ✅ Two regression tests added covering happy-path completion and hang-after-progress scenarios
5. ✅ Fix covers all exit paths (normal `done`, timeout throw, Telegram render exception, early break)

No unmet intent. No scope creep.

### Dimensions with no issues

- **security** — no issues found. No user data, credentials, or secrets touched. Two scalar field resets in a finally block. Read `streaming.ts:340–357`.
- **api-compat** — no issues found. `WatchdogState` is a module-internal struct; no exported interface changed.
- **perf-observability** — no issues found. Two O(1) assignments (`= false`, `= null`) with zero allocation. No logging removed or added.

### What was not checked

- Citations verified inline against branch HEAD `35f813085b`.
- Stated intent: PR #1172 title+body — spec-compliance assessed.
- Not checked: runtime behavior of `makeNextWithTimeout` under real Anthropic API delays (pre-existing architecture, not modified by this diff).
- Not checked: interaction with the `MAX_API_ROUND_INFLIGHT_MS` env-var override path in `streaming.watchdog.ts` (pre-existing, not modified).

---

**Decision: MERGE — 0 blocking (2 low).**

The production fix is correct, minimal, and well-placed. Both findings are test-quality observations — the tests exercise the right paths but miss an opportunity to discriminate the exact timing regression.

PR #1182, opened 2026-08-18
fix(telegram): derive MAX_API_ROUND_INFLIGHT_MS from live TTFB config

### LINKED ISSUE #1142: fix(telegram): derive MAX_API_ROUND_INFLIGHT_MS from live TTFB config
## Problem

`MAX_API_ROUND_INFLIGHT_MS` is hardcoded to `420_000` in `streaming.watchdog.ts`, derived from the **default** `AFK_MODEL_TTFB_TIMEOUT_MS = 180_000` (3 attempts × 120s = 360s worst case, plus 60s headroom).

Operators who raise `AFK_MODEL_TTFB_TIMEOUT_MS` above ~225,000 will find the watchdog fires **before** the provider's own retry budget exhausts — reintroducing the exact false-positive timeout that #1135 / PR #1141 fixed.

Example: at `AFK_MODEL_TTFB_TIMEOUT_MS=300_000`, `ttfbAttemptTimeoutMs` yields 200s per attempt × 3 = 600s worst case, which exceeds the 420s ceiling.

## Options

1. **Derive at runtime** — compute the ceiling from `TTFB_MAX_ATTEMPTS × ttfbAttemptTimeoutMs(resolveTtfbTimeoutMs())` plus a headroom factor, passed into `makeNextWithTimeout` or computed at watchdog init.
2. **Startup assertion** — assert `MAX_API_ROUND_INFLIGHT_MS > TTFB_MAX_ATTEMPTS × ttfbAttemptTimeoutMs(resolveTtfbTimeoutMs())` at session start so misconfiguration fails fast.

Option 1 is more robust; option 2 is simpler and catches the misconfiguration early.

## References

- `src/telegram/streaming.watchdog.ts:49` — hardcoded constant
- `src/agent/providers/anthropic-direct/loop/retry-budget.ts:54–70` — TTFB budget derivation
- `src/agent/providers/shared/first-byte-timeout.ts:47–53` — operator-configurable resolver
- PR #1141 — introduced the constant

griffinwork40:
## What

Replaces the hardcoded `MAX_API_ROUND_INFLIGHT_MS = 420_000` constant in `streaming.watchdog.ts` with a `resolveMaxApiRoundInflightMs()` function that derives the watchdog ceiling at call time from the live TTFB configuration.

## Why

Operators who raise `AFK_MODEL_TTFB_TIMEOUT_MS` above ~225 s hit false watchdog timeouts. The provider-level TTFB budget is `TTFB_MAX_ATTEMPTS × ttfbAttemptTimeoutMs(configured)` — a runtime value that scales with that env var. The old hardcoded `420_000` ceiling assumed the default (180 s), so any higher TTFB configuration caused the Telegram watchdog to fire *before* the provider's own retry logic had a chance to recover, producing spurious `StreamTimeoutError`s mid-conversation.

## How

```
resolveMaxApiRoundInflightMs()
  = TTFB_MAX_ATTEMPTS × ttfbAttemptTimeoutMs(resolveTtfbTimeoutMs())
    + MAX_API_ROUND_HEADROOM_MS (60 s)
```

- `MAX_API_ROUND_HEADROOM_MS = 60_000` is the fixed slack added on top of the provider's worst-case TTFB budget.
- `resolveMaxApiRoundInflightMs()` reads the live env at call time, so raising `AFK_MODEL_TTFB_TIMEOUT_MS` automatically raises the watchdog ceiling proportionally.
- Both call sites of the old constant are replaced with the function.

## Backward compatibility

With `AFK_MODEL_TTFB_TIMEOUT_MS` unset (default → 180 000 ms):

```
ttfbAttemptTimeoutMs(180_000) = floor(180_000 × 2/3) = 120_000
3 × 120_000 + 60_000 = 420_000
```

The default config produces exactly the previous hardcoded value — no behavior change for existing deployments.

## Tests

New test cases in `streaming.watchdog.test.ts`:

| Test | Assertion |
|------|-----------|
| Default matches old 420 000 | `resolveMaxApiRoundInflightMs() === 420_000` |
| Exceeds provider TTFB worst case | `> 3 × 120_000 = 360_000` |
| Less than `MAX_TOOL_INFLIGHT_MS` | API calls can't hang longer than tools |
| Scales for high TTFB (300 s → 660 s) | Proportional derivation verified |
| Behavioral: suspends past `NEXT_EVENT_TIMEOUT_MS` | `apiRoundInFlight=true` delays reject until `resolveMaxApiRoundInflightMs()` elapses |
| Behavioral: resumes after event clears flag | Normal `NEXT_EVENT_TIMEOUT_MS` cadence restored |

Full regression run: **913 tests, 48 files — all pass**.

Closes #1142

### DIFF
--- a/src/telegram/streaming.watchdog.test.ts
+++ b/src/telegram/streaming.watchdog.test.ts
@@ -6,14 +6,17 @@
 
 import { describe, it, expect, vi } from 'vitest';
 import {
-  MAX_API_ROUND_INFLIGHT_MS,
+  MAX_API_ROUND_HEADROOM_MS,
   MAX_TOOL_INFLIGHT_MS,
   NEXT_EVENT_TIMEOUT_MS,
   TOOL_INFLIGHT_RECHECK_MS,
   makeNextWithTimeout,
+  resolveMaxApiRoundInflightMs,
   type WatchdogState,
 } from './streaming.watchdog.js';
 import { StreamTimeoutError } from './stream-timeout-error.js';
+import { TTFB_MAX_ATTEMPTS, ttfbAttemptTimeoutMs } from '../agent/providers/anthropic-direct/loop/retry-budget.js';
+import { DEFAULT_MODEL_TTFB_TIMEOUT_MS } from '../agent/providers/shared/first-byte-timeout.js';
 
 /** Build a minimal WatchdogState with apiRoundInFlight pre-set to true. */
 function makeState(overrides: Partial<WatchdogState> = {}): WatchdogState {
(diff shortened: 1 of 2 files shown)


griffinwork40:
## `/review` — PR #1182

**Target:** `fix(telegram): derive MAX_API_ROUND_INFLIGHT_MS from live TTFB config`
**Reviewed ref:** `82fcb0e335757ada24f886e6128727c1ee7359d0` (branch `afk/issue-1142`)
**Regime:** light (refactor, ~180 lines changed, 2 files)
**Citations verified against:** branch HEAD `82fcb0e`
**Stated intent:** Fix false watchdog timeouts when `AFK_MODEL_TTFB_TIMEOUT_MS` is raised above default — spec-compliance assessed ✓

---

### Decision: MERGE — 0 blocking (1 medium waived, 2 low, 1 nit)

---

### Findings

**medium · blocking:false · confidence:high · api-compat**
`src/telegram/streaming.watchdog.ts:13–14` · `diff-context` · ref:`82fcb0e`

**New provider-layer coupling: telegram watchdog imports Anthropic-specific retry constants**

`resolveMaxApiRoundInflightMs()` imports `TTFB_MAX_ATTEMPTS` and `ttfbAttemptTimeoutMs` from `../agent/providers/anthropic-direct/loop/retry-budget.js`. The Telegram streaming layer is provider-agnostic — OpenAI-compatible sessions (`session-openai.ts`) share the same `streaming.ts` event loop and the same `watchdog.apiRoundInFlight` flag. When an operator uses the OpenAI-compatible provider with a non-default `AFK_MODEL_TTFB_TIMEOUT_MS`, the watchdog ceiling is computed from Anthropic's `TTFB_MAX_ATTEMPTS=3` and `floor(2B/3)` formula, which may not match the OpenAI provider's actual TTFB behavior.

```ts
import { TTFB_MAX_ATTEMPTS, ttfbAttemptTimeoutMs } from '../agent/providers/anthropic-direct/loop/retry-budget.js';
```

**Suggestion:** Extract `TTFB_MAX_ATTEMPTS` and the per-attempt formula into `providers/shared/` alongside `resolveTtfbTimeoutMs()`, so the watchdog ceiling is provider-agnostic. If the OpenAI provider does not retry TTFB at all, a provider-aware `resolveMaxApiRoundInflightMs(provider)` dispatcher is the cleanest fix.

· waived: bounded, non-data-affecting defect — the old hardcoded `420_000` was equally misaligned for the OpenAI provider under high TTFB configs, so this is no regression on existing behavior, only a missed generalization.

---

**low · blocking:false · confidence:high · test-coverage**
`src/telegram/streaming.watchdog.test.ts:74–84` · `diff-context` · ref:`82fcb0e`

**Proportional-scaling test does not exercise `resolveMaxApiRoundInflightMs()` under a high TTFB env var**

The test named `'resolveMaxApiRoundInflightMs scales proportionally for a high TTFB timeout'` verifies that `TTFB_MAX_ATTEMPTS * ttfbAttemptTimeoutMs(300_000) + MAX_API_ROUND_HEADROOM_MS === 660_000`, but never calls `resolveMaxApiRoundInflightMs()` itself with `AFK_MODEL_TTFB_TIMEOUT_MS` set to `'300000'`. It tests the constituent math (already covered in `retry-budget.test.ts`) rather than the integration.

```ts
// test body — resolveMaxApiRoundInflightMs() is never called here
const highTtfb = 300_000;
const expected = TTFB_MAX_ATTEMPTS * ttfbAttemptTimeoutMs(highTtfb) + MAX_API_ROUND_HEADROOM_MS;
expect(expected).toBe(660_000);
```

**Suggestion:** Add `vi.stubEnv('AFK_MODEL_TTFB_TIMEOUT_MS', '300000')` (restored in `afterEach`) and assert `expect(resolveMaxApiRoundInflightMs()).toBe(660_000)` to exercise the actual env-reading path.

---

**low · blocking:false · confidence:high · correctness**
`src/telegram/streaming.watchdog.ts:80–82` · `diff-context` · ref:`82fcb0e`

**Zero-TTFB edge case: watchdog ceiling collapses to headroom-only (60 s) when `AFK_MODEL_TTFB_TIMEOUT_MS=0`**

When `AFK_MODEL_TTFB_TIMEOUT_MS=0` (the explicit disable escape hatch), `resolveTtfbTimeoutMs()` returns `0`, `ttfbAttemptTimeoutMs(0)` returns `0` (by the `<= 0` guard), so `resolveMaxApiRoundInflightMs() === MAX_API_ROUND_HEADROOM_MS === 60_000`. The old constant was `420_000`. This is a significant reduction for the Telegram watchdog suspension window in TTFB-disabled mode. The JSDoc acknowledges this: *"callers should not rely on the ceiling in that mode anyway since the provider has no per-attempt bound"* — but `streaming.ts` does rely on it, and a legitimately slow 80-second context-processing round would now trip the watchdog after 60 s instead of 420 s.

**Suggestion:** Document explicitly in the JSDoc that TTFB-disabled mode changes the behavioral contract for Telegram streaming, or add a floor: `Math.max(MAX_API_ROUND_HEADROOM_MS + providerTtfbBudget, MINIMUM_INFLIGHT_CEILING_MS)`.

---

**nit · blocking:false · confidence:high · correctness**
`src/telegram/streaming.watchdog.test.ts:197` · `diff-context` · ref:`82fcb0e`

`void resolveSecond` suppresses a false-positive `noUnusedLocals` error (`resolveSecond` is used inside the mock closure at line 163 but TS can't see it). No behavioral consequence.

**Suggestion:** A `// eslint-disable-next-line` comment with a brief explanation would make the intent clearer than the bare `void` expression.

---

### What was not checked

- Citations verified inline against branch HEAD `82fcb0e`.
- Did not run the test suite locally; PR body states 913 tests, 48 files — all pass.
- Did not review OpenAI-compatible provider TTFB internals for parity beyond what was needed for the coupling finding.
- Stated intent: PR #1182 title + body — spec-compliance assessed.

PR #1189, opened 2026-08-18
feat(resume): virtual-scroll picker with fuzzy search and relative timestamps

griffinwork40:
## Summary

The `/resume` picker hard-caps at 20 entries and has no search — with 3,000+ session sidecars, 99% of sessions are unreachable through the interactive path. This PR adds virtual scrolling, fuzzy search, and improved metadata.

### Changes

**Virtual scroll** — Replaced the `slice(0, 20)` cap with a 20-row viewport that scrolls through the full session corpus. Only the visible window is rendered per repaint (O(WINDOW_SIZE) not O(N)), with scroll indicators showing position.

**Fuzzy search** — Type to filter sessions live. Printable chars build a filter query; Esc clears the filter (second Esc cancels). Gated behind `searchable?: boolean` in `RunPickerOptions` — the elicitation picker and config menu are completely unaffected (they don't pass `searchable`).

**Relative timestamps** — "3h ago", "2d ago" instead of ISO-8601 dates.

**Cost display** — `$0.12` shown in the picker label (was already in the non-TTY table but missing from the interactive picker).

### Files

| File | Change |
|---|---|
| `src/cli/render/picker.ts` | Virtual scroll viewport + searchable mode |
| `src/cli/render/picker-filter.ts` | New — fuzzy scorer (extracted to stay under 350-LOC ceiling) |
| `src/cli/render/picker.test.ts` | +27 new tests (scroll, search, unchanged behavior) |
| `src/cli/slash/commands/resume.ts` | Cap removed, relative timestamps, cost in label, `searchable: true` |

### Design decisions

- **Scorer:** prefix-then-subsequence, zero dependencies, ~30 LOC. ANSI stripped before scoring.
- **Filter state** lives entirely inside the `runPicker` closure — zero compositor changes, zero `PickerController` interface changes.
- **Enter on filtered list** resolves the original label string (not the filtered view), so `resume.ts`'s `options.indexOf(choice)` mapping still works.
- **Stdin invariant** preserved — no new listeners, no `setRawMode` calls. Printable chars route through the existing `onKey` handler.
- **Non-TTY cap** raised from 20 → 50.

### DIFF
--- a/src/cli/render/picker-filter.test.ts
+++ b/src/cli/render/picker-filter.test.ts
@@ -1,0 +1,108 @@
+/**
+ * Unit tests for the picker-filter scorer.
+ *
+ * Covers the three scoring tiers (exact substring / subsequence / no match)
+ * and the ANSI-stripping behaviour required for options produced by
+ * `uniquePickLabels` in `resume.ts` (which appends `palette.dim(id)` to
+ * disambiguate duplicate labels).
+ */
+
+import { describe, expect, it } from 'vitest';
+import { filterOptions } from './picker-filter.js';
+
+describe('filterOptions — empty query', () => {
+  it('returns all options in original order', () => {
+    const opts = ['apple', 'banana', 'cherry'];
+    const results = filterOptions(opts, '');
+    expect(results.map((r) => r.originalIndex)).toEqual([0, 1, 2]);
+  });
+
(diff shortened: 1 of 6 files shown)


griffinwork40:
@codex fix comments

Issue #1171, opened 2026-08-18
fix(tui): use basename for PAGER path comparison in resolvePager


griffinwork40:
## Problem

`resolvePager()` in `src/cli/slash/commands/transcript.ts` appends `-R` to `less` so ANSI color codes render correctly, but the check uses a bare-name comparison:

```ts
if (cmd === 'less' && !args.includes('-R')) args.push('-R');
```

Users who set `PAGER=/usr/bin/less` (full path) get `cmd = '/usr/bin/less'`, which does not match `'less'`, so `-R` is not appended and the transcript renders raw escape bytes.

## Fix

```ts
import { basename } from 'node:path';
if (basename(cmd) === 'less' && !args.includes('-R') && !args.includes('-r')) args.push('-R');
```

The additional `-r` check avoids a redundant append when the user already passes the lowercase variant (deprecated but functional).

## Context

Found during review of #1168. The `PAGER=less` (bare) case — by far the most common — is already handled correctly. This is a narrow environmental edge case.

PR #1181, opened 2026-08-18
fix(daemon): add observability and test coverage for fallback-failure path (#1166)

griffinwork40:
Fixes #1166

### `src/agent/daemon.listen.ts`

1. **Stderr log on fallback failure** — the catch block now calls `process.stderr.write()` naming both addresses before re-throwing, so launchd captures the diagnostic the same way the happy-path `logFallback()` does.

2. **Error message names both addresses** — `enrichEaddrinuse` gains an overload that accepts a `fallback: string` argument; when both addresses were tried the error message reads `both 127.0.0.1:PORT and ::1:PORT are occupied`.

3. **`retryErr` chained as cause** — the thrown error now wraps `retryErr` as `{ cause }` so the full failure chain is visible in stack traces.

### `src/agent/daemon.listen.test.ts`

4. **New describe block: `fallback-failure: both loopback addresses occupied`** — stubs lsof to exit 1 (ghost socket), occupies both `127.0.0.1:PORT` and `::1:PORT` with real servers, asserts the thrown error mentions both addresses, and asserts stderr received the diagnostic write.

### `src/agent/daemon.test.ts`

5. **Bonus: host:port format case for `stop()` ownership check** — adds a sibling test that writes `127.0.0.1:65501` (new format) to the port file and verifies `stop()` leaves it intact, complementing the existing bare-port `65501` case.

## Test results

```
✓ src/agent/daemon.listen.test.ts  10 tests passed  (was 7)
✓ src/agent/daemon.test.ts         56 tests passed  (was 55)
pnpm lint — clean
```

PR #1175, opened 2026-08-18
chore(daemon): harden portHasOwner — execFileSync, permission-denied, IPv6 guard (#1167)

griffinwork40:
## Summary

Closes #1167. Also covers the observability gap from #1166.

Three hardening items in `daemon.listen.ts`:

### 1. `execSync` → `execFileSync`
Eliminates the shell entirely. `port` is typed `number` so no injection exists today, but this hardens against future type-widening.

### 2. Permission-denied detection in `portHasOwner`
After catching exit status 1, inspects stderr for `Permission denied` or `WARNING:` keywords. Returns `null` (unknown) instead of `false` when found, preventing silent fallback when lsof lacks privilege.

### 3. IPv6 skip guard in tests
Added `hasIPv6Loopback()` helper and `it.skipIf(!IPV6_AVAILABLE)` guards around IPv6-dependent test blocks, so environments with IPv6 disabled get a clean skip instead of a confusing failure.

### Bonus: Fallback-failure observability (#1166)
- Added `process.stderr.write` before re-throw naming both addresses
- Enriched error message includes fallback address
- Added test for the fallback-failure path

## Tests

- All mocks updated from `execSync` to `execFileSync`
- IPv6 skip guards on 3 test blocks
- Fallback-failure path test with both-address assertion

Issue #1170, opened 2026-08-18
fix(tui): stop bgStatusBar and mascotBar during pager/editor handoffs


griffinwork40:
## Problem

PR #1168 wired `suspendFooter`/`resumeFooter` to stop periodic terminal writers during pager/editor handoffs, but only stops two of the four timer-based writers named in the PR body:

| Writer | Interval | Stopped? |
|---|---|---|
| `healthRail` | 1s tick | ✅ |
| `statusLine` | configurable tick | ✅ |
| `bgStatusBar` | ~50ms spinner | ❌ |
| `mascotBar` | animation timer | ❌ |

The omitted writers continue injecting ANSI cursor escapes into stdout while `less` or `$EDITOR` owns the terminal via `stdio: 'inherit'`. The health rail (1s, cursor save/restore) was the primary corruptor and is now fixed, but the bgStatusBar spinner fires at ~50ms when background agents are active and can still corrupt the display.

## Fix

Extend both callbacks in `src/cli/commands/interactive/repl-loop.ts`:

```ts
ctx.slashCtx.suspendFooter = () => {
  f.bgStatusBar.stop(); f.mascotBar.stop();
  f.healthRail.stop(); ctx.statusLine.stop();
};
ctx.slashCtx.resumeFooter = () => {
  ctx.statusLine.start(); f.healthRail.start();
  f.mascotBar.start(); f.bgStatusBar.start();
};
```

Both `bgStatusBar.stop()` and `mascotBar.stop()` are guarded (`if (!this.started) return`) and idempotent. The `start()` calls are similarly safe when the subsystem has no active work (no background agents / mascot env unset). Order mirrors the teardown sequence already in the `finally` block at lines 157–159.

## Context

Found during review of #1168. The fix is a two-line extension of the existing callbacks — no new plumbing needed.

PR #1191, opened 2026-08-19
fix(tui): use basename for PAGER path comparison in resolvePager (#1171)

griffinwork40:
## Problem

`resolvePager()` in `src/cli/slash/commands/transcript.ts` appended `-R` to `less` so ANSI color codes render correctly, but the check used a bare-name comparison:

```ts
if (cmd === 'less' && !args.includes('-R')) args.push('-R');
```

Users who set `PAGER=/usr/bin/less` got `cmd = '/usr/bin/less'`, which did not match `'less'`, so `-R` was silently skipped and ANSI escape sequences were displayed as raw bytes in the pager.

## Fix

- **`path.basename(cmd) === 'less'`** — resolves the full-path mismatch; `path` was already imported.
- **`!args.includes('-r')`** — guards against the deprecated lowercase `-r` variant to avoid appending a redundant `-R`.

## Changes

| File | Line | Change |
|------|------|--------|
| `src/cli/slash/commands/transcript.ts` | 118 | `cmd === 'less'` → `path.basename(cmd) === 'less'`; add `-r` guard |
| `src/cli/slash/commands/transcript.test.ts` | new describe | 3 new regression tests: full-path PAGER, already-set `-R`, already-set `-r` |

## Test results

```
✓ src/cli/slash/commands/transcript.test.ts (7 tests) 35ms
  ✓ resolvePager — PAGER env resolution (3)
  ✓ /transcript slash command — pager TTY handoff (4)
```

### DIFF
--- a/src/agent/daemon.listen.test.ts
+++ b/src/agent/daemon.listen.test.ts
@@ -3,21 +3,48 @@
  */
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import * as net from 'node:net';
 
 // vi.mock is hoisted above imports — must appear before any other import.
 vi.mock('node:child_process', async (importActual) => {
   const actual = await importActual<typeof import('node:child_process')>();
   return {
     ...actual,
-    execSync: vi.fn(actual.execSync),
+    execFileSync: vi.fn(actual.execFileSync),
   };
 });
 
-import { execSync } from 'node:child_process';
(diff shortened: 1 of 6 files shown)


griffinwork40:
Closing: this PR bundled 4 issues on one branch. Splitting into individual PRs per issue. #1167 → #1175, #1166 → #1181 (both pre-existing). Creating separate PRs for #1171 and #1170.

PR #1204, opened 2026-08-19
fix(plugins): auto-recover ghost dirs on install

griffinwork40:
## Problem

When a plugin directory exists on disk but has no matching entry in `.index.json`, users are trapped in a dead-end state:

```
$ afk plugin install capital-mischief/cm-ops
✖ Failed — plugin directory already exists (re-run with --force to replace)

$ afk plugin update capital-mischief/cm-ops
✖ Error — plugin "capital-mischief/cm-ops" is not installed
```

`install` gates on **filesystem** presence. `update` gates on **index** presence. When the two disagree, neither command works.

This happens when:
- A prior install crashed after cloning the dir but before writing the index entry
- The `.index.json` was rebuilt or corrupted
- A plugin dir was manually dropped into `~/.afk/plugins/`

## Fix

`assertDestAvailable` now checks the index when it finds an existing directory. If the dir exists but there is **no** matching index entry, the directory is an orphaned ghost — it is auto-removed and install proceeds without requiring `--force`.

When both the dir AND the index entry exist (a true collision), the behavior is unchanged — the user must pass `--force`.

## Test

Added `auto-recovers ghost dir (exists on disk but not in index)` test in `install.test.ts`.

**14/14 install tests pass, 25/25 update tests pass, lint clean, filesize clean.**


griffinwork40 (commented):

## Review: PR #1204 — `fix(plugins): auto-recover ghost dirs on install`

**Target:** PR #1204 → `main`
**Change type:** hotfix | **Regime:** light | **Files:** 2 | **Lines:** +42/−4
**Reviewed ref:** `92bda59c93b10fb14435d932e62c03ba10e871f4`

---

### Findings

**1 · `low` · blocking:false · confidence:high · test-coverage · `src/agent/plugins/install.test.ts:219–243` · ref:92bda59 · diff-context**
The new test asserts `result.name` and `readIndex(...).plugins['ghost-plugin']` but does not assert that the ghost directory itself (`ghostDir`) or its contents (`stale.txt`) were actually removed. A regression where `removeDest` is skipped but `upsertPlugin` still runs would pass the test while leaving orphaned files.
**suggestion:** Add `expect(existsSync(ghostDir)).toBe(false)` after the install call.

**2 · `low` · blocking:false · confidence:high · test-coverage · `src/agent/plugins/install.test.ts` · ref:92bda59 · file-state**
`removeDest` has two branches: `unlinkSync` for symlinks and `rmSync({recursive:true})` for directories. The new test exercises only the directory branch. A ghost symlink (leftover from a previous `installLocal` that crashed after `symlinkSync` but before the index write) is a realistic scenario that is now auto-recovered but untested.
**suggestion:** Add a test case creating a ghost symlink without an index entry, then verify install succeeds and the symlink is removed.

**3 · `nit` · blocking:false · confidence:high · correctness · `src/agent/plugins/install.ts:432–433` · ref:92bda59 · file-state**
`if (force) return;` fires before ghost detection (line 433). With `--force`, the ghost directory is removed by the caller's separate `if (existsSync(dest)) removeDest(dest)` block — correct behavior, but the two independent removal paths (ghost-auto-recovery vs. force-removal) are not documented. A future refactor that removes the caller's force-cleanup block would silently break `--force` on ghost dirs.
**suggestion:** Add an inline comment on the `if (force) return` line noting the caller handles removal.

---

### Spec-compliance

Stated intent: PR #1204 title+body — spec-compliance assessed.

1. ✅ `assertDestAvailable` now checks the index when a directory collision is detected
2. ✅ Ghost directories (dir exists, no index entry) are auto-removed without `--force`
3. ✅ True collisions (dir exists AND index entry exists) still throw — unchanged behavior
4. ✅ All three call sites (`installLocal`, `installGit` initial, `installGit` rename) updated to pass `indexPath`
5. ✅ New test exercises the ghost-recovery path

No unmet intent. No scope creep.

### Dimensions with no issues

- **security** — no issues found. `assertWithinPluginsDir` + `assertSafePluginName` run before ghost detection; `readIndex` is read-only; `removeDest` only removes within-bounds validated paths. Read `install.ts:131–134`.
- **correctness** — no issues found. Ghost detection logic is sound: `basename(dest)` correctly resolves the plugin name for index lookup; `readIndex` returns an empty `plugins` object when the index file is missing or empty; `removeDest` handles both directories and symlinks.
- **api-compat** — no issues found. `assertDestAvailable` is not exported; signature change affects only internal call sites, all three of which are updated.
- **perf-observability** — no issues found. `readIndex` adds one synchronous file read per `assertDestAvailable` invocation on an existing-dir collision path — acceptable overhead for an install operation.

### What was not checked

- Citations verified inline against branch HEAD `92bda59c93`.
- Stated intent: PR #1204 title+body — spec-compliance assessed.
- Not checked: behavior when `.index.json` is corrupted (malformed JSON). `readIndex` presumably throws or returns a default — pre-existing behavior, not introduced by this change.
- Not checked: concurrent install races where two processes hit ghost detection simultaneously (pre-existing TOCTOU in `existsSync` → `removeDest`).

---

CI note: `Test (ubuntu-latest)` shows FAILURE on this PR. The test failures should be investigated but may be pre-existing/flaky — the fix itself is clean and well-scoped.

griffinwork40 (commented):

## Review: PR #1204 — `fix(plugins): auto-recover ghost dirs on install`

**Target:** `afk/fix-plugin-install-state` → `main` · ref `92bda59c` · +52/−7 · 2 files
**Regime:** light (hotfix, 52 lines)
**Stated intent:** PR #1204 title+body — spec-compliance assessed.

---

### Findings

**1 · `medium` · ⛔ blocking:true · confidence:high · security · `src/agent/plugins/install.ts:436–440` + `src/agent/plugins/index-store.ts:111–113` · ref:92bda59c · file-state**

**Corrupt or unreadable `.index.json` causes `readIndex` to return an empty plugins map, making every existing plugin directory appear to be a ghost — `assertDestAvailable` then auto-deletes it without warning.** `readIndex` swallows all parse errors via `catch { return cloneEmpty(); }`, returning `{ plugins: {} }`. When `index.plugins[name]` is `undefined` (because the map is empty), the ghost check fires `removeDest(dest)` — an `rmSync` with `recursive: true, force: true`. A transiently corrupted index file (partial write, disk-full, wrong encoding, or a future schema version not yet handled) makes the next `installPlugin` call silently delete an actively-used plugin's directory. medium + security is non-overridable per regime rules.

```typescript
// index-store.ts:111-113 — parse errors silently return empty
} catch {
  return cloneEmpty();  // { version: 2, plugins: {}, marketplaces: {} }
}

// install.ts:436-440 — empty map → ghost detection fires on every name
if (!index.plugins[name]) {
  removeDest(dest);  // rmSync(dest, { recursive: true, force: true })
  return;
}
```

**suggestion:** Guard against a corrupt index before trusting the empty result. Simplest approach: check `existsSync(indexPath)` before the ghost branch — if the file exists on disk but `readIndex` returned zero plugins, throw the normal collision error rather than auto-removing. A brand-new install dir (no index file at all) correctly yields an empty map and the ghost branch is safe.

---

**2 · `medium` · blocking:false · confidence:high · correctness · CI status · ref:92bda59c · waived: bounded — the ubuntu failure is unexplained from the diff (all 3 call sites are correctly updated, all existing tests pass on macOS), and the PR body claims 14/14 + 25/25 pass locally; the CI failure may be a pre-existing flake or infrastructure issue rather than a defect in this change**

**Test (ubuntu-latest) CI leg is FAILURE while all other checks (macOS, Lint & Build, PTY, Docs, Publish Bundle) pass.** The PR body states "14/14 install tests pass, 25/25 update tests pass" but the ubuntu CI contradicts this. The diff is straightforward (3 call-site updates + one new code path) and compiles cleanly (Lint & Build = SUCCESS), so the ubuntu failure is likely in a different test file or a pre-existing flake — but cannot be confirmed without the CI logs.

**suggestion:** Re-run the ubuntu CI job to confirm whether the failure is transient, or inspect the CI logs to identify the failing test. Do not merge with a red CI leg unexplained.

---

**3 · `low` · blocking:false · confidence:high · test-coverage · `src/agent/plugins/install.test.ts` · ref:92bda59c · file-state**

**No test covers ghost recovery in the `installGit` path.** The new ghost test exercises only the `installLocal` flow. `installGit` has two `assertDestAvailable` call sites (lines 178 and 216) that also receive `indexPath` and exercise ghost recovery. Grep confirmed: only one ghost test exists, under the `collisions` describe block, using `installPlugin(sourceDir, ...)` which routes to `installLocal`.

**suggestion:** Add a test that creates a ghost dir at the provisional clone destination, then runs `installPlugin` with a GitHub shorthand source to verify the git install path auto-cleans the ghost.

---

**4 · `low` · blocking:false · confidence:high · test-coverage · `src/agent/plugins/install.test.ts` · ref:92bda59c · file-state**

**No test covers a dangling-symlink ghost (local-source install that crashed after symlinking but before writing the index).** `assertDestAvailable` line 431 enters the ghost path when `existsSync(dest)` is false but `isLink(dest)` is true — which is the dangling-symlink case. The `removeDest` function correctly handles this via `unlinkSync`, but the branch is untested for ghost recovery specifically.

**suggestion:** Add a test that creates a dangling symlink at the ghost path (symlink to a non-existent target) and verifies `installPlugin` auto-cleans it.

---

**5 · `low` · blocking:false · confidence:medium · spec-compliance · PR description · ref:92bda59c · diff-context**

**The "opposite ghost" (index entry exists but directory is missing) is not addressed.** The PR title says "auto-recover ghost dirs" and the body describes the deadlock between `install` and `update`, but only one direction is fixed: dir-without-index. The reverse case (index entry exists but the directory was deleted) would trap the user in `update` (which would fail on missing files) without a recovery path. This is a documentation gap — the PR body implies bidirectional recovery but delivers unidirectional.

**suggestion:** Clarify in the PR body that only the dir-without-index direction is addressed, or file a follow-up issue for the index-without-dir case.

---

**6 · `nit` · blocking:false · confidence:high · perf-observability · `src/agent/plugins/install.ts:436` · ref:92bda59c · diff-context**

**`readIndex` is now called on every collision check, adding a synchronous `readFileSync` + `JSON.parse` to a path that previously had none.** For `installGit` with a rename (manifest name differs from provisional), two `readIndex` calls fire. Impact is negligible — install is already I/O-bound by `git clone` — but the I/O cost is new.

---

### Dimensions with no issues

- **API-compat** — no issues found. `assertDestAvailable` is a non-exported module-private function; its signature change is invisible to external callers. All three internal call sites are updated. `readIndex` import is additive — the function was already a stable public export of `index-store.ts`. No public API surface changes. Read `install.ts` (all call sites), `index-store.ts:83`.

### Spec-compliance

Stated intent: PR #1204 title+body — spec-compliance assessed.

| Requirement | Status |
|---|---|
| Dir exists + no index entry → auto-remove and proceed | ✅ Implemented in `assertDestAvailable` |
| Dir exists + index entry exists → unchanged (require `--force`) | ✅ Preserved — the `throw` path is unchanged |
| Test for ghost recovery | ✅ Added (`auto-recovers ghost dir` test) |
| 14/14 install, 25/25 update pass | ⚠ ubuntu CI red (Finding 2) |

No scope creep. Finding 5 notes a documentation gap (opposite-ghost not mentioned).

### What was not checked

- Citations verified inline against branch HEAD `92bda59c93b10fb14435d932e62c03ba10e871f4` via pre-fetched file content from the GitHub API.
- Stated intent: PR #1204 title+body — spec-compliance assessed.
- Not checked: ubuntu CI failure logs (only the FAILURE status is visible from the PR metadata; the specific failing test is not known).
- Not checked: whether `readIndex` corruption risk is pre-existing (it is — `readIndex` has always returned `cloneEmpty()` on errors — but the PR extends its blast radius by acting on the empty result with `removeDest`).
- Not checked: TOCTOU between `readIndex` and `removeDest` under concurrent installs (pre-existing race; no global file lock in this module).
- Not checked: interaction with marketplace install path (`assertNotMarketplace` gates before `assertDestAvailable`; marketplace sources are rejected before the ghost check runs).

---

**Decision: DO NOT MERGE — 1 medium blocking (security: corrupt index → silent plugin deletion); 1 medium waived (correctness: unexplained ubuntu CI failure); 3 low, 1 nit.**

The sole blocking finding (Finding 1) is a security-dimension medium: a corrupt `.index.json` makes the ghost-recovery path delete any existing plugin directory without warning. The fix is small — guard against a corrupt-but-present index file before trusting the empty-plugins result — and would make the ghost-recovery path safe against index corruption. The waived medium (Finding 2) is the unexplained ubuntu CI failure; it is waived because the diff is mechanically correct and the failure may be pre-existing, but the CI leg should be investigated before merge.

griffinwork40 (commented):

## Review: PR #1204 — `fix(plugins): auto-recover ghost dirs on install`

**Target:** `afk/fix-plugin-install-state` → `main` · ref `af3509e8` · +42/−4 · 2 files
**Regime:** light (hotfix)
**Stated intent:** PR #1204 title+body — spec-compliance assessed.

---

### Findings

**1 · `medium` · ⛔ blocking:true · confidence:high · security · `src/agent/plugins/install.ts:430–437` + `src/agent/plugins/index-store.ts:84,88,110–112` · ref:af3509e8 · file-state**

**Corrupt `.index.json` makes `readIndex` return an empty plugins map, causing `assertDestAvailable` to treat every existing plugin directory as a ghost and silently auto-delete it.** `readIndex` returns `cloneEmpty()` (`{ plugins: {} }`) on parse error (line 112), malformed JSON (line 88), and unknown/future version (line 110) — indistinguishable from a genuinely missing file (line 84). The new ghost-detection path checks `if (!index.plugins[name])` and calls `removeDest(dest)` — an `rmSync` with `{ recursive: true, force: true }`. A transiently corrupt index (interrupted atomic write, disk-full, future schema version not yet handled) makes the next `installPlugin` call silently destroy the existing plugin directory before proceeding.

```typescript
// index-store.ts:111-112 — parse errors silently return empty
} catch { return cloneEmpty(); }

// install.ts:430-437 — empty map → ghost detection fires on every name
const index = readIndex(indexPath);
if (!index.plugins[name]) { removeDest(dest); return; }
```

`readIndex` is **pre-existing** code not modified by this PR, but the PR extends its blast radius by acting on the empty result with `removeDest`. medium + security is non-overridable per regime rules.

**suggestion:** Guard: check `existsSync(indexPath)` before the ghost branch — if the file exists on disk but `readIndex` returned zero plugins, throw the normal collision error rather than auto-removing. A genuinely missing index file (brand-new install dir) correctly yields an empty map and the ghost branch is safe.

---

**2 · `low` · blocking:false · confidence:high · test-coverage · `src/agent/plugins/install.test.ts:219–238` · ref:af3509e8 · diff-context**

**Ghost-recovery test covers only the `installLocal` path; the `installGit` provisional-dir (line 178) and rename-dir (line 216) call sites are untested.** Confirmed: `install.test.ts` contains exactly one ghost test (line 219), exercising only the local-source flow.

**suggestion:** Add a test using `makeFakeGit` that seeds a ghost dir at the provisional clone destination, then runs `installPlugin` with a GitHub shorthand source to verify the git install path auto-cleans the ghost.

---

**3 · `low` · blocking:false · confidence:high · test-coverage · `src/agent/plugins/install.test.ts:219–238` · ref:af3509e8 · diff-context**

**Test does not assert the ghost directory's old content was actually removed.** The test asserts `result.name`, checks the index entry exists, and verifies `enabled === true`, but never checks `existsSync(ghostDir)` or `existsSync(join(ghostDir, 'stale.txt'))`. A regression where `removeDest` is skipped but `upsertPlugin` still runs would pass the test while leaving orphaned files.

**suggestion:** Add `expect(existsSync(join(ghostDir, 'stale.txt'))).toBe(false)` after the install call.

---

**4 · `low` · blocking:false · confidence:high · spec-compliance · `src/agent/marketplaces/install.ts:96,133,161,208` · ref:af3509e8 · file-state**

**The parallel `assertDestAvailable` in `marketplaces/install.ts` has the same ghost-trap but is not patched.** `marketplaces/install.ts:208` has a private 2-param `assertDestAvailable(dest, force)` with no ghost recovery, called at lines 96, 133, and 161. The PR title correctly scopes to `fix(plugins)`, so this is not an omission in the stated scope — but the PR body describes the general problem without qualification, and marketplace installs expose the same dead-end to users.

**suggestion:** File a follow-up issue for marketplace ghost-dir recovery, or note the scoping in the PR body.

---

**5 · `nit` · blocking:false · confidence:high · perf-observability · `src/agent/plugins/install.ts:430` · ref:af3509e8 · diff-context**

**`readIndex(indexPath)` is called synchronously inside `assertDestAvailable`, which fires up to 3 times per `installGit` invocation (lines 134, 178, 216).** Each call does `existsSync` + `readFileSync` + `JSON.parse`. Impact is negligible for a single install but accumulates in bulk operations.

**suggestion:** Cache the index read within the install call's scope in a future pass.

---

### Dimensions with no issues

- **api-compat** — no issues found. `assertDestAvailable` is a non-exported module-private function; its signature change affects only internal call sites, all three of which are updated. `readIndex` import is additive — the function was already a stable public export of `index-store.ts`. No public API surface changes. Read `install.ts` (all call sites), `index-store.ts:83`.

- **correctness** — no issues found. Ghost detection logic is sound: `basename(dest)` correctly resolves the plugin name for index lookup. `removeDest` handles both directories and symlinks. All three `installLocal`/`installGit`/rename call sites are properly updated with the `indexPath` parameter. The `if (force) return` early-exit correctly bypasses ghost detection when `--force` is supplied.

### Spec-compliance

Stated intent: PR #1204 title+body — spec-compliance assessed.

| Requirement | Status |
|---|---|
| Dir exists + no index entry → auto-remove and proceed | ✅ Implemented in `assertDestAvailable` |
| Dir exists + index entry exists → unchanged (require `--force`) | ✅ Preserved — the `throw` path is unchanged |
| Test for ghost recovery | ✅ Added (local-install path only — Finding 2 notes missing git paths) |

No scope creep detected. Finding 4 notes marketplace is unpatched but out of stated scope.

### What was not checked

- Citations verified inline against branch HEAD `af3509e890c56fbd23da8ea6b0df385c03b69a0b`.
- Stated intent: PR #1204 title+body — spec-compliance assessed.
- Not checked: ubuntu CI failure logs (only the FAILURE status is visible from the PR metadata; the specific failing test is not known).
- Not checked: whether `readIndex` corruption risk is pre-existing (it is — `readIndex` has always returned `cloneEmpty()` on errors — but the PR extends its blast radius by acting on the empty result with `removeDest`).
- Not checked: concurrent install races where two processes hit ghost detection simultaneously (pre-existing TOCTOU in `existsSync` → `removeDest`).
- Not checked: interaction with marketplace install path (`assertNotMarketplace` gates before `assertDestAvailable`; marketplace sources are rejected before the ghost check runs).

---

**Decision: DO NOT MERGE — 1 medium blocking (security: corrupt index → silent plugin deletion); 3 low, 1 nit.**

The sole blocking finding: a corrupt `.index.json` makes the ghost-recovery path delete any existing plugin directory without warning. The fix is small — guard against a corrupt-but-present index file before trusting the empty-plugins result — and would make the ghost-recovery path safe against index corruption.

griffinwork40 (commented):

## Review: PR #1204 — `fix(plugins): auto-recover ghost dirs on install`

**Target:** `afk/fix-plugin-install-state` → `main` · ref `af3509e8` · +42/−4 · 2 files
**Regime:** light (hotfix)
**Stated intent:** PR #1204 title+body — spec-compliance assessed.

---

### Findings

**1 · `medium` · blocking:true · confidence:high · correctness · `src/agent/plugins/install.ts:427–445` · ref:af3509e8 · file-state**

**Live local-source symlink with missing index entry is silently destroyed.** A local plugin installed via `afk plugin install /path/to/source` creates a symlink at `~/.afk/plugins/<name>` and writes an index entry. If the index is subsequently rebuilt, corrupted, or manually edited to remove that entry, the symlink still exists and the plugin still works at runtime. On the next `install` targeting the same name, `assertDestAvailable` detects the symlink (`existsSync` returns true, `isLink` returns true), finds no index entry, and calls `removeDest` — which calls `unlinkSync`, silently destroying the live plugin's registration. The user's working plugin is deleted without warning.

```typescript
const name = basename(dest);
const index = readIndex(indexPath);
if (!index.plugins[name]) {
  removeDest(dest);  // silently destroys a live symlink
  return;
}
```

**suggestion:** Before auto-removing, check whether the target resolves to a real, populated directory: `if (isLink(dest) && existsSync(dest))` → refuse (the symlink is live, not a ghost) and require `--force`. Alternatively, emit a stderr warning naming the path being auto-cleaned so the user has an audit trail.

---

**2 · `medium` · blocking:false · confidence:high · correctness · `src/agent/plugins/install.ts:215–216,238–246` · ref:af3509e8 · file-state · waived: bounded — requires a ghost dir at the rename target AND a permission failure on `rmSync` for that ghost, an extremely narrow compound condition; the pre-existing catch block already had this same gap for `renamedDir` under `--force`**

**`installGit` catch-block cleanup does not cover `renamedDir`.** When ghost recovery fires inside `assertDestAvailable(renamedDir, ...)` at line 216 and `removeDest` throws (permissions, locked file), the error propagates to the outer catch block at lines 238–246. That catch block only cleans `dest` (the provisional dir), not `renamedDir` (the partially-deleted ghost at the rename target). A partially-removed ghost directory persists on disk.

```typescript
} catch (err) {
  try {
    if (existsSync(dest)) removeDest(dest);
    // renamedDir is never cleaned here
  } catch { /* ignore */ }
  throw err;
}
```

**suggestion:** Extend the catch block: `if (renamedDir && renamedDir !== dest && existsSync(renamedDir)) removeDest(renamedDir);` — or hoist `renamedDir` to outer scope so the catch has access.

---

**3 · `low` · blocking:false · confidence:high · test-coverage · `src/agent/plugins/install.test.ts:219–238` · ref:af3509e8 · diff-context**

**Ghost-recovery test does not assert stale contents were evicted.** The test creates a ghost dir with `stale.txt`, installs over it, and asserts index presence — but never checks that `stale.txt` is gone or that the directory was replaced by a symlink. A regression where `removeDest` silently fails would pass this test.

**suggestion:** Add `expect(existsSync(join(ghostDir, 'stale.txt'))).toBe(false)` and `expect(lstatSync(join(pluginsDir, 'ghost-plugin')).isSymbolicLink()).toBe(true)`.

---

**4 · `low` · blocking:false · confidence:high · test-coverage · `src/agent/plugins/install.test.ts` · ref:af3509e8 · file-state**

**No test covers ghost recovery in the `installGit` code path.** All three `assertDestAvailable` call sites were modified, but the new test exercises only the `installLocal` path. The `installGit` provisional-name check (line 178) and rename-path check (line 216) are untested for ghost recovery. Grep for `ghost` in `install.test.ts`: matches only at lines 219–238 (the local-path test).

**suggestion:** Add a git-path ghost test: `makeFakeGit` with a pre-created ghost dir at the provisional name and no index entry, assert install succeeds.

---

**5 · `nit` · blocking:false · confidence:high · correctness · `src/agent/plugins/install.ts:427` · ref:af3509e8 · file-state**

**`assertDestAvailable` performs a destructive side effect (removing a ghost dir) despite the `assert*` naming convention implying a pure guard.** All other `assert*` functions in this module (`assertSafePluginName`, `assertWithinPluginsDir`, `assertHttpsUrl`) are pure predicates that throw on failure and never mutate state. The ghost-recovery mutation breaks that convention.

**suggestion:** Rename to `ensureDestAvailable` or extract the ghost removal into a separate `removeGhostIfOrphaned(dest, indexPath)` call at each call site.

---

### Spec-compliance

Stated intent: PR #1204 title+body — spec-compliance assessed.

| Requirement | Status |
|---|---|
| Ghost dir (exists on disk, no index entry) → auto-removed, install proceeds | ✅ Implemented in `assertDestAvailable` |
| True collision (dir + index entry both exist) → unchanged behavior | ✅ `throw` path unchanged |
| Test: ghost dir auto-recovery | ✅ Added (local path only) |
| All 3 `assertDestAvailable` call sites updated | ✅ Lines 133, 178, 216 |

No scope creep detected. All changes are within stated intent.

### Dimensions with no issues

- **Security** — no issues found. `assertDestAvailable` operates only within `pluginsDir` (validated by `assertWithinPluginsDir` at each call site). `readIndex` reads from the established index path. `removeDest` uses `rmSync({recursive: true, force: true})` or `unlinkSync` — no path injection surface. `basename(dest)` for the index lookup is derived from a name that passed `assertSafePluginName`. Read `install.ts:425–453`, `install.ts:131–133`, `install.ts:176–179`.
- **API-compat** — no issues found. `assertDestAvailable` is module-private (not exported). The added `readIndex` import is from a sibling module already used by `upsertPlugin`. No public API signature changed. Read `install.ts:37`.
- **Perf-observability** — no issues found. `readIndex` is a synchronous `readFileSync` + `JSON.parse` called at most twice per `installGit` (provisional + rename paths). The index file is small (plugin metadata, not content). No hot-path concern.

### What was not checked

- Citations verified inline against branch HEAD `af3509e890c56fbd23da8ea6b0df385c03b69a0b`.
- Stated intent: PR #1204 title+body — spec-compliance assessed.
- Not checked: the `update` command path — PR body describes the dead-end state but this PR only fixes `install`. Whether `update` has a parallel ghost-recovery gap is a follow-on concern.
- Not checked: runtime behavior on Windows (symlink semantics differ; `isLink` + `unlinkSync` may behave differently with NTFS junctions).
- Not checked: interaction with the marketplace installer's `assertDestAvailable` usage (marketplace sources are rejected early at line 107 before any dest check).

---

**Decision: DO NOT MERGE — 1 medium blocking (correctness: live symlink silently destroyed on index mismatch); 1 medium waived (correctness: catch-block cleanup scope, bounded); 2 low, 1 nit.**

The sole blocker is Finding 1: a live local-source plugin whose index entry is missing (corruption, rebuild) will have its symlink silently deleted when another install targets the same name. The fix is small — gate the ghost-removal path on `!(isLink(dest) && existsSync(dest))` to protect live symlinks, or emit a warning before silent deletion.

Issue #1196, opened 2026-08-19
feat(workspace): shared agent workspace — v2 tracking (measurement, experiment, query tool)


griffinwork40:
Tracking issue for work deferred from the shared agent workspace v1 (#1195).

## Measurement harness

Build a harness that instruments a multi-agent task run and measures file-read duplication across siblings. The 59% baseline (34 agents reading the same file) was observed empirically but never persisted as a reproducible measurement — the harness needs to:

- Count unique vs. total `read_file` calls across all children in a root session
- Report per-file read counts and a duplication ratio
- Support A/B comparison: control (workspace disabled) vs. treatment (workspace enabled)

Design is sketched in `.afk/research/shared-agent-workspace-rfc.md`.

## A/B duplication experiment

Run the measurement harness on a representative multi-agent task (e.g. a `/mint` or `/diagnose` run) with workspace disabled, then enabled, and compare the duplication ratio against the 59% control baseline.

## `workspace_query` as a model tool (v2)

v1 intentionally excludes `workspace_query` from the model tool surface to minimize agent-policy confounds in the duplication experiment. Once the experiment validates the workspace's impact, expose query as a tool so agents can pull sibling context on demand (not just at fork time via harness injection).

## Other v2 considerations

- Reactive workspace (auto-publish from tool results)
- TTL / relevance decay on entries
- Cross-session workspace persistence (currently per-root-session only)

griffinwork40:
Status update (2026-08-31):

## Completed

- **`workspace_query` as a model tool (v2)** — shipped. Both `workspace_query` and `workspace_publish` are live on the tool surface. Agents can pull sibling context on demand, not just at fork time via harness injection.
- **`argsFingerprint` on trace `tool_call` events** — landed in #1205. Enables dedup measurement across sibling agents.
- **Cold-start wiring gaps** — resolved in #1209, #1214 (cold-start hint, compose executor wiring).
- **Workspace store threading** — landed in #1210, #1213 (root manager, skill phases, permission gates).

## Remaining

1. **Full measurement harness** — `argsFingerprint` landed but the end-to-end A/B comparison framework (control vs. treatment, duplication ratio reporting) is not complete.
2. **A/B duplication experiment** — not run. Needs the harness above to compare workspace-enabled vs. workspace-disabled on a representative multi-agent task.
3. **Reactive workspace** — auto-publish findings from tool results (no implementation yet).
[...]

PR #1195, opened 2026-08-19
feat(workspace): shared agent workspace v1 — publish tool + harness-injected retrieval

griffinwork40:
## What

Adds a shared agent workspace — a per-session SQLite-backed scratchpad that lets sibling sub-agents share structured findings instead of each independently re-reading the same files.

## Why

Empirical observation: in multi-agent tasks, ~59% of file reads are duplicated across siblings because each agent starts from zero context about what its siblings already discovered. This workspace gives agents a way to publish findings that are automatically injected into sibling contexts.

### Architecture (user-specified design decisions)

- **Retrieval**: harness-managed, automatically injected into each subagent's initial system prompt via `injectWorkspacePreamble()` (follows the `budget-preamble.ts` pattern)
- **Publication**: model-directed through an explicit `workspace_publish` tool
- **Query**: internal API only — intentionally excluded from v1 to minimize agent-policy confounds in the duplication experiment

### Implementation

**New files** (`src/agent/workspace/`):
- `workspace-store.ts` — SQLite store mirroring `MemoryStore` pattern. Six entry types: Finding, Evidence, Hypothesis, Decision, Artifact, Status. Per-root-session scope.
- `workspace-tools.ts` — `workspace_publish` tool schema + handler factory
- `workspace-preamble.ts` — queries relevant entries by keyword overlap on `subject` field + recency, injects as a `[workspace-context]` block into the child's system prompt
- `index.ts` — barrel export

**Modified files** (wiring):
- Both providers (`anthropic-direct`, `openai-compatible`) — accept optional `workspaceStore`, register `workspace_publish` handler, pass store through dispatcher deps
- `fork-child-config.ts` — injects workspace preamble alongside budget preamble at subagent fork time
- `nesting.ts` — forwards `workspaceStore` from parent to child
- `provider-schemas.ts` — registers workspace tool schema
- All three surface entry points (CLI REPL, Telegram, daemon) — create one `WorkspaceStore` per root session and pass it through

### Not included (next steps)
- Measurement harness to quantify file-read deduplication improvement
- Running the actual A/B experiment (control vs. workspace-enabled)
- `workspace_query` as a model tool (deferred to v2)

## Tests

47 new tests across 3 test files covering store CRUD, tool handler, preamble injection, and relevance filtering. Full suite passes (17,170 tests, 7 pre-existing failures unchanged).

## Verification

- `pnpm lint` ✅
- `pnpm test` ✅ (no new failures)
- `pnpm audit:filesize:check` ✅
[...]

PR #1205, opened 2026-08-19
feat(trace): add argsFingerprint to tool_call events + dedup measurement harness

griffinwork40:
## What

Adds `argsFingerprint` to `tool_call.started` trace events and ships a measurement harness for the workspace v2 dedup experiment.

## Why

The shared agent workspace v1 (#1195) ships a publish tool + harness-injected retrieval so sibling subagents can share findings. The next step (#1196) is measuring whether it actually reduces duplicated file reads. The trace had `inputBytes` but not the actual content — so we couldn't distinguish "two agents read the same file" from "two agents read different files of the same size."

### `argsFingerprint` field (production)

A SHA-256 hex digest of `JSON.stringify(input)` added to every `tool_call.started` event:

- **`types.ts`** — new required field on `ToolCallStartedPayload`
- **`events.ts`** — Zod schema updated
- **`tool-call-trace.ts`** — hash computed from serialized input alongside the existing `inputBytes`

The hash is deterministic: identical tool calls produce identical fingerprints. Two agents calling `read_file` with the same path + offset + limit produce the same hash.

### `repeated-tool-use` detector upgrade

The detector previously used a proxy fingerprint (`v1-bytes-tuple`) — a SHA-256 over `(name, inputBytes, resultBytes, isError, subagentId)` that false-collided on unrelated calls with identical byte counts. Now uses `argsFingerprint` directly (`v2-args-hash`). The old `computeFingerprint` function is retained as an export for backward compatibility.

### Measurement harness (`scripts/measure-read-dedup.ts`)

Standalone script that reads a root session's `trace.jsonl` and reports:
- Total `read_file` calls across all subagents
- Unique fingerprints (distinct file+args combinations)
- Duplicated calls (same fingerprint from different agents)
- **Deduplication ratio** — the headline metric for the A/B experiment
- Top "hot" fingerprints (most-duplicated across agents)

```
tsx scripts/measure-read-dedup.ts --latest        # most recent session
tsx scripts/measure-read-dedup.ts --session <id>  # specific session
tsx scripts/measure-read-dedup.ts --json          # machine-readable
tsx scripts/measure-read-dedup.ts --all-tools     # all tools, not just read_file
```

## Tests

- 4 new tests on `buildToolCallStartedPayload` (fingerprint correctness, identity, divergence, undefined-input)
- 12 test files updated for the new required `argsFingerprint` field on started-event fixtures
- All 17,194 tests pass (2 pre-existing failures unchanged: `write-denylist` symlink race, `streaming.watchdog` stale reference)

## Verification

- `pnpm lint` ✅
- `pnpm test` ✅ (no new failures)
- `pnpm audit:filesize:check` ✅ (baseline regenerated for +1 LOC in `events.ts` and `types.ts`)
- `pnpm audit:sdk:check` ✅
- `pnpm audit:env:check` ✅
- `pnpm audit:chalk:check` ✅
- `pnpm fix:pins:check` ✅

Closes the measurement-harness item from #1196.


griffinwork40:
Absence confirmed — no test covers the mixed success/failure loop detection that v2 widens to catch.

Now I'll synthesize the final review:

---

## Review — PR #1205: `feat(trace): add argsFingerprint to tool_call events + dedup measurement harness`

**Reviewed ref:** `a88bcbed` · **Merge base:** `317b1891`  
**Change type:** feature · **Regime:** full (624 +/ 56 −, 19 files)  
**Stated intent:** PR #1205 title + body — spec-compliance assessed.

---

### Findings

**1 · medium · blocking: true · confidence: high · security · `src/agent/providers/shared/tool-call-trace.ts:67-73` · ref: a88bcbed · citation-type: file-state**

`redactSensitiveFields` is a closed allowlist covering only `browser_act` + `action === 'fill'`. Every other tool carrying secrets in its input has those secrets hashed verbatim into `argsFingerprint` and persisted to `trace.jsonl`. The most acute case is `config_set`: an `afk config env set ANTHROPIC_API_KEY sk-ant-...` call passes the raw key through `buildToolCallStartedPayload` → `JSON.stringify(raw)` → SHA-256, persisting the raw API key as the hash preimage to the trace file. `medium` security is never overridable.

```ts
function redactSensitiveFields(toolName: string, input: unknown): unknown {
  if (toolName !== 'browser_act' || typeof input !== 'object' || input === null) {
    return input;
  }
```

**Suggestion:** Add `config_set` (redact `value`) to the allowlist. Document the coverage boundary in the function's JSDoc so future tool additions know to opt in.

---

**2 · medium · blocking: true · confidence: medium · correctness · `src/improve/scan/detectors/repeated-tool-use.ts:177` · ref: a88bcbed · citation-type: file-state**

The v2 outer fingerprint uses `ev.payload.name` (from the **completed** event) but `argsHash` comes from the **started** event. While both providers guarantee `started.name === completed.name` today, the implicit invariant is unasserted and load-bearing for fingerprint correctness.

```ts
const fingerprint = argsHash
  ? createHash('sha256').update(`${ev.payload.name}|${argsHash}`).digest('hex')
  //                               ^^^^^^^^^^^^^^^^ from completed event
```

**Suggestion:** Change `${ev.payload.name}` to `${started.name}` — one-token fix, semantically consistent (both halves derive from the started event), zero behavior change on valid traces. · **waived: bounded non-data-affecting defect; started.name === completed.name by construction in both providers, so the fingerprint is correct on all real traces. The fix is a robustness improvement, not a bug fix.**

---

**3 · medium · blocking: true · confidence: high · test-coverage · `src/improve/scan/detectors/repeated-tool-use.test.ts` · ref: a88bcbed · citation-type: diff-context**

The PR's key correctness claim — "v2 widens detection to catch loops where successive calls alternate between success and failure" — has no test. The stated behavioral improvement (v2 excludes `isError`/`resultBytes` from the fingerprint) is documented in the code comment but not verified.

**Suggestion:** Add a test: 4× same tool + same `argsFingerprint` where `isError` alternates `true`/`false`. Assert detection fires with `fingerprintAlgorithm: 'v2-args-hash'`. Optionally contrast with a `legacyPairLines` variant showing v1 would miss it.

---

**4 · low · blocking: false · confidence: high · security · `src/agent/providers/shared/tool-call-trace.ts:57` · ref: a88bcbed · citation-type: file-state**

SHA-256 without salt. For short/low-entropy inputs, an attacker with trace-file access can dictionary-attack the hash. Non-blocking: the attacker already needs filesystem access, and once Finding 1 is fixed, no high-entropy secrets reach the hasher.

**Suggestion:** Document the residual risk in the `argsFingerprint` field JSDoc. A per-session salt is optional depth.

---

**5 · low · blocking: false · confidence: high · correctness · `src/agent/providers/shared/tool-call-trace.ts:50-57` · ref: a88bcbed · citation-type: file-state**

`inputBytes` uses raw serialization; `argsFingerprint` uses redacted serialization. Intentional and documented in the `Invariant:` comment. Not a bug — but the type-level JSDoc on `argsFingerprint` in `types.ts` doesn't mention this divergence.

**Suggestion:** Add a note on `types.ts:ToolCallStartedPayload.argsFingerprint`: "Computed from redacted input — may diverge from a hash derived from inputBytes for browser_act fill actions."

---

**6 · low · blocking: false · confidence: high · perf-observability · `src/agent/providers/shared/tool-call-trace.ts:50-51` · ref: a88bcbed · citation-type: file-state**

Double `JSON.stringify` on every non-`browser_act fill` tool call (once for `inputBytes`, once for `hashInput`). When `redactSensitiveFields` returns the original reference, the second call is redundant. Negligible at current scale (~5 µs for 1 KB inputs × hundreds of calls = <1 ms).

**Suggestion:** Optional micro-optimization: when `redactSensitiveFields` returns `input` unchanged (reference equality), reuse `serialized` as `hashInput`. Not worth doing now.

---

**7 · low · blocking: false · confidence: high · test-coverage · `scripts/measure-read-dedup.ts` · ref: a88bcbed · citation-type: diff-context**

The measurement harness (369 lines) has no automated tests. Non-blocking for a script — wrong output misinforms the A/B decision rather than breaking production.

---

**8 · nit · blocking: false · confidence: high · spec-compliance · `scripts/measure-read-dedup.ts:362` · ref: a88bcbed · citation-type: file-state**

Warning message says `"${report.toolFilter} events lacked argsFingerprint"` → produces `"read_file only events lacked argsFingerprint"` (grammatically awkward).

**Suggestion:** `"${report.skippedNoFingerprint} matched events lacked argsFingerprint (pre-upgrade trace)"`.

---

### Dropped findings from Wave 1

| Finding | Reason |
|---|---|
| Security-api F4 (ToolCallPair lacks usedArgsHash) | **fabricated** — verified at ref a88bcbed: `usedArgsHash: boolean` is present on line 111. Agent cited pre-PR state. |

### What was not checked

- Citations verified inline against branch HEAD `a88bcbed`.
- Stated intent: PR #1205 title+body — spec-compliance assessed.
- Did not run the test suite or build.
- Did not review Telegram, daemon, or browser surfaces for argsFingerprint propagation.
- No runtime testing of the harness script.

---

### Decision

**Decision: DO NOT MERGE — 1 medium blocking (security), 1 medium blocking (test-coverage), 1 medium waived (correctness: bounded, names always match by construction); 4 low, 1 nit.**

The security finding (Finding 1) is the primary blocker — `config_set` secrets reaching the trace file via `argsFingerprint` is a real risk on any machine where the operator configures API keys through the agent. The test-coverage gap (Finding 3) is the secondary blocker — the PR's key claimed improvement (v2 catches alternating-error loops) has no verification. Finding 2 (medium/correctness) is waived as a robustness nit that does not affect output on valid traces.

---

PR #1212, opened 2026-08-19
feat(telegram): enable background subagent dispatch + improve bg/fg guidance

griffinwork40:
## What

Two improvements to subagent dispatch strategy:

### 1. System prompt guidance for background vs. foreground

The model previously had only 2 sentences in the `agent` tool schema description about when to use background mode. Now there's structured guidance in the Delegation section of the system prompt with a clear 3-criterion decision heuristic:

1. Result won't gate this turn's next action
2. Task is genuinely long
3. Surface is interactive (REPL or Telegram)

Also clarifies: use `compose` for parallel-this-turn needs, not multiple background dispatches.

### 2. Telegram background subagent support

Previously, `mode: "background"` on Telegram returned `isError: true` — "Background mode is not available in this session" — because no `BackgroundAgentRegistry` was wired into the Telegram session builder.

**Changes:**
- **`session-anthropic.ts`**: Create `BackgroundAgentRegistry` and pass it to `wireExecutors` via the existing parameter. Wire `cancelAll()` + notifier `dispose()` into `drainSubagents` for clean teardown.
- **`bg-result-notifier.ts`** (new): Subscribes to registry `settled` events, pushes a one-line status notification to the operator via `pushIfConfigured()`. Skips cancelled jobs (same contract as REPL notifier).

**Design decision**: Push-based delivery (Telegram API message) rather than injection into `processOne`. This avoids:
- The `claimedChats` reference-counting invariant (#603) — any logic before `reserveClaim()` risks the documented TOCTOU race
- The `message.ts` filesize ratchet (581 LOC baseline)
- Premature shared module extraction (YAGNI — one consumer today)

Model-context injection (so the model sees bg results in its next turn without user relay) is deferred to a follow-up.

## Tests

- 7 new tests for `TelegramBgResultNotifier` covering: completed/failed/cancelled jobs, chat targeting, default routing, dispose cleanup, error swallowing
- All 887 existing Telegram tests pass
- All CI audits pass (filesize, SDK deps, env, chalk, module-state, pins)

### DIFF
--- a/prompts/system-prompt.md
+++ b/prompts/system-prompt.md
@@ -96,7 +96,7 @@ Default to foreground. Use background (`mode: "background"`) only when **all thr
 
 When you need multiple results this turn, use `compose` (parallel foreground) — not multiple background dispatches that arrive unpredictably on future turns.
 
-Background results are delivered automatically at the start of the next turn as `<background-subagent-result>` blocks. They are capped at 16KB; full output is available via `/bgsub:join <jobId>`.
+On REPL, background results are delivered automatically at the start of the next turn as `<background-subagent-result>` blocks, capped at 16KB; full output is available via `/bgsub:join <jobId>`. On Telegram, you receive a push notification when the job settles but the result is NOT injected into your context — ask the user to relay findings or avoid background mode for tasks whose results you need.
 
 ## Decision commitment
 
--- a/prompts/system-prompt.md
+++ b/prompts/system-prompt.md
@@ -87,6 +87,17 @@ Parallelize independent subagents in one wave. Use the `compose` tool when dispa
 
 Subagents return compressed findings, not raw exploration. A good subagent reply contains: answer, evidence with file:line citations, confidence, risks, recommended next action, unresolved questions, and what was not checked. Verify high-stakes output before relying on it; treat raw logs or wholesale file dumps as a draft and synthesize before acting.
 
+### Background vs. foreground
+
+Default to foreground. Use background (`mode: "background"`) only when **all three** hold:
+1. The result will not gate this turn's next action — you can continue useful work without it.
+2. The task is genuinely long (multi-file investigation, broad search, test suite run).
+3. You are on an interactive surface (REPL or Telegram). Background mode is unavailable on daemon and one-shot surfaces — the call returns an error.
+
+When you need multiple results this turn, use `compose` (parallel foreground) — not multiple background dispatches that arrive unpredictably on future turns.
+
+Background results are delivered automatically at the start of the next turn as `<background-subagent-result>` blocks. They are capped at 16KB; full output is available via `/bgsub:join <jobId>`.
+
 ## Decision commitment
 
 When diagnosing and fixing code:
--- a/src/agent/types/config-types.ts
+++ b/src/agent/types/config-types.ts
@@ -773,6 +773,16 @@ export interface AgentConfig {
   autoCompact?: boolean | { threshold: number };
 
   /**
+   * Telegram chat id for this session — set by the Telegram session manager
+   * before calling `createSession` so the `TelegramBgResultNotifier` can push
+   * background job notifications to the originating chat rather than falling
+   * back to the bot-global default notify targets.
+   *
+   * Telegram-scoped only: never set on REPL, daemon, or one-shot sessions.
+   */
+  telegramChatId?: number;
+
+  /**
    * In-process custom tools available to the session. Each entry is created
    * via the `tool()` helper and provides both a JSON-schema `AnthropicToolDef`
    * (so the model knows the tool exists) and a validated `ToolHandler`
--- a/src/telegram/bg-result-notifier.test.ts
+++ b/src/telegram/bg-result-notifier.test.ts
@@ -1,0 +1,190 @@
+/**
+ * Tests for `TelegramBgResultNotifier` — push notification for settled
+ * background subagent jobs on the Telegram surface.
+ *
+ * Uses the real `BackgroundAgentRegistry`; jobs are driven to terminal states
+ * via a stubbed `SubagentHandle` (same harness as the REPL's
+ * bg-result-notifier.test.ts).
+ */
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+import { BackgroundAgentRegistry } from '../agent/background-registry.js';
+import type { SubagentHandle, SubagentResult } from '../agent/subagent.js';
+import { TelegramBgResultNotifier } from './bg-result-notifier.js';
+
+// Stub pushIfConfigured so we can observe calls without hitting the network.
+vi.mock('./push.js', () => ({
+  pushIfConfigured: vi.fn(async () => []),
+}));
+
+// Silence routing telemetry writes (background-registry emits them on settle).
+vi.mock('../agent/routing-telemetry.js', () => ({
+  appendRoutingDecision: vi.fn(async () => {}),
+}));
+
+import { pushIfConfigured } from './push.js';
+
+const pushMock = vi.mocked(pushIfConfigured);
+
+/** Stub a `SubagentHandle` whose `runInBackground` callback we control. */
+function makeBgHandle(id = 'sub-1'): {
+  handle: SubagentHandle;
+  fireTerminal: (r: SubagentResult) => void;
+} {
+  let captured: ((r: SubagentResult) => void) | undefined;
+  return {
+    handle: {
+      id,
+      status: 'idle',
+      runInBackground: vi.fn((_p: string, on?: (r: SubagentResult) => void) => {
+        captured = on;
+      }),
+      cancel: vi.fn().mockResolvedValue(undefined),
+      teardown: vi.fn().mockResolvedValue(undefined),
+      run: vi.fn(),
+      runToResult: vi.fn(),
+    } as unknown as SubagentHandle,
+    fireTerminal: (r) => captured?.(r),
+  };
+}
+
+function succeed(id: string, content: string): SubagentResult {
+  return {
+    id,
+    status: 'succeeded',
+    message: { content, role: 'assistant' } as unknown as SubagentResult['message'],
+  } as SubagentResult;
+}
+
+function fail(id: string, msg: string): SubagentResult {
+  return {
+    id,
+    status: 'failed',
+    error: new Error(msg),
+  } as SubagentResult;
+}
+
+describe('TelegramBgResultNotifier', () => {
+  let registry: BackgroundAgentRegistry;
+  let notifier: TelegramBgResultNotifier;
+
+  beforeEach(() => {
+    registry = new BackgroundAgentRegistry({});
+    notifier = new TelegramBgResultNotifier(registry);
+    pushMock.mockClear();
+  });
+
+  afterEach(() => {
+    notifier.dispose();
+  });
+
+  it('pushes a notification when a background job completes', () => {
+    const { handle, fireTerminal } = makeBgHandle();
+    const job = registry.register({
+      handle,
+      prompt: 'investigate something',
+      model: 'sonnet',
+    });
+
+    fireTerminal(succeed(job.jobId, 'found the answer'));
+
+    expect(pushMock).toHaveBeenCalledTimes(1);
+    const text = pushMock.mock.calls[0]![0];
+    expect(text).toContain('✅');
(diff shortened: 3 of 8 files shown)


griffinwork40:
All key citations verified against the reviewed ref. Let me now compile the verification manifest and synthesize.

**Wave 1.5 — Verification Manifest:**

| # | Type | Claim | Status | Notes |
|---|---|---|---|---|
| F1 (sec) | citation | `session-anthropic.ts:66` — `new TelegramBgResultNotifier(backgroundRegistry)` no chatId | **verified** | Line 66 at ref: `const bgNotifier = new TelegramBgResultNotifier(backgroundRegistry);` |
| F1 (sec) | citation | `push.ts:195-200` — undefined target → resolveConfiguredNotifyTargets | **verified** | Lines 196-201 at ref confirm |
| C1 | citation | `bg-result-notifier.ts:67` — only pushIfConfigured called | **verified** | Line 67 at ref: `void pushIfConfigured(text, { target: this.chatId }).catch(() => {});` |
| C1 | absence | no `<background-subagent-result>` injection in `src/telegram/` | **confirmed-absent** | `git grep` returned zero matches |
| C1 | citation | `prompts/system-prompt.md:99` — universal bg-result claim | **verified** | Line 99: "Background results are delivered automatically..." with no surface qualifier |
| C2 | citation | `markDelivered` called in REPL at line 209 | **verified** | Matches exactly |
| C2 | absence | markDelivered not called in Telegram notifier | **confirmed-absent** | No match in bg-result-notifier.ts |
| C3 | citation | test structure — sync not.toThrow vs async rejection | **verified** | Test at lines 176-189, the catch path is async |
| S1 | same gap as C1 | — | **verified** | Overlapping finding |

No fabricated citations. No false-absent claims. All blocking findings stand.

Now I need to assess whether the C1/S1 finding truly warrants blocking. The system-prompt guidance text is this PR's own content, and it makes a factual claim about Telegram that isn't implemented. But let me check: does the PR description already acknowledge this gap?

The PR body says: *"Model-context injection (so the model sees bg results in its next turn without user relay) is deferred to a follow-up."* — so the author knows the injection path is deferred. The system prompt guidance, however, doesn't say this — it tells the model results arrive as `<background-subagent-result>` blocks universally.

This is a real correctness/spec-compliance issue — the prompt will mislead the model on Telegram. Let me now synthesize.

---

# PR #1212 Review — `feat(telegram): enable background subagent dispatch + improve bg/fg guidance`

**Target:** `afk/improve-subagent-strategy` → `main` | **Ref:** `81164841` | **Regime:** full (306 lines, feature)
**Files:** `prompts/system-prompt.md` · `src/telegram/bg-result-notifier.ts` (new) · `src/telegram/bg-result-notifier.test.ts` (new) · `src/telegram/session-anthropic.ts`

---

## Blocking findings

### 1 · medium · blocking:true · correctness + spec-compliance · `prompts/system-prompt.md:99` · confidence high

**System prompt claims `<background-subagent-result>` injection on Telegram; Telegram delivers push notification only.** Line 99 states universally: *"Background results are delivered automatically at the start of the next turn as `<background-subagent-result>` blocks."* This is true on REPL (which has `BgResultNotifier` with model-context injection) but false on Telegram — the new `TelegramBgResultNotifier` fires only a `pushIfConfigured` push notification to the human operator. No `<background-subagent-result>` block is injected into the model's context. A Telegram-side model that dispatches a background job will expect a structured block that never arrives.

The PR description explicitly acknowledges this: *"Model-context injection deferred to a follow-up."* The system prompt guidance — **which is part of this PR** — does not.

**Evidence:** `bg-result-notifier.ts:67` — only `pushIfConfigured` is called. `src/telegram/` has zero matches for `<background-subagent-result>`. `src/cli/commands/interactive/bg-result-notifier.ts` confirms model-context injection is REPL-only.

**Suggestion:** Add a surface qualifier to the system prompt: *"On REPL, background results are delivered automatically as `<background-subagent-result>` blocks. On Telegram, you receive a push notification when the job settles; retrieve the result with `/bgsub:join <jobId>`."*

### 2 · medium · blocking:true · security · `src/telegram/session-anthropic.ts:66` · confidence high

**`TelegramBgResultNotifier` is constructed without a `chatId`, routing all bg-job push notifications to the bot-global notify targets instead of the originating chat.** In a multi-chat deployment, user A's background-job completion notification (including the first ~80 chars of their dispatch prompt as the label) is delivered to user B if B is the configured primary notify target. This is a confidentiality leak of per-session prompt content across chat boundaries.

**Evidence:** `session-anthropic.ts:66` — `new TelegramBgResultNotifier(backgroundRegistry)` — no second argument. `push.ts:196` — `opts.target === undefined` → falls through to `resolveConfiguredNotifyTargets()`, which returns bot-global targets. `TelegramSessionBuildContext` carries no `chatId` field.

**Suggestion:** Thread `chatId` from the originating Telegraf context through `TelegramSessionBuildContext` and pass it as the second argument. The chat id at that point has already passed the bot's allowlist check at the inbound boundary.

---

## Non-blocking findings

### 3 · low · blocking:false · correctness · `src/telegram/bg-result-notifier.ts:63-67` · confidence high

**`markDelivered()` is never called on the Telegram path.** The REPL analog calls `registry.markDelivered(job.jobId)` (line 209), which emits a `background_agent.delivered` witness event. The Telegram notifier skips it, so push-notified settlements are indistinguishable from un-notified ones in the witness trace.

**Suggestion:** Call `this.registry.markDelivered(job.jobId)` inside `onSettled` after firing the push.

### 4 · low · blocking:false · correctness · `src/telegram/bg-result-notifier.test.ts:204-217` · confidence high

**The "swallows push errors" test cannot falsify the bug it names.** `fireTerminal` is synchronous, and `not.toThrow()` catches only synchronous throws. The actual error path is an async rejection on `pushIfConfigured()` swallowed by `.catch(() => {})`. Removing the `.catch` would introduce an unhandled rejection that this test would still pass.

**Suggestion:** Drain the microtask queue (e.g. `await vi.waitFor(...)` or `await Promise.resolve()`) and assert no unhandled rejection propagated, or use Vitest's unhandled-rejection detection.

### 5 · low · blocking:false · perf-observability · `src/telegram/bg-result-notifier.ts:67` · confidence high

**Push failures are fully silent** — `.catch(() => {})` swallows with no log, no trace event. On Telegram (where the push is the *only* delivery mechanism, unlike REPL where model-context injection is primary), repeated silent push failures leave the operator with no signal that notifications stopped.

**Suggestion:** Add a `debugLog` call in the `.catch` handler; optionally emit a lightweight trace event on push failure.

### 6 · nit · blocking:false · security · `src/telegram/bg-result-notifier.ts:77` · confidence high

**`chatId` constructor param has no JSDoc constraint on provenance.** The `pushIfConfigured` `target` path explicitly documents that callers must enforce the allowlist — a one-line JSDoc note on the constructor stating `chatId` must originate from the inbound Telegraf context (already allowlist-validated) would close the documentation gap.

---

## What was not checked

- Citations verified inline against branch HEAD `81164841`.
- Stated intent: PR #1212 title + body — spec-compliance assessed.
- Did not review the OpenAI-compatible or xAI Telegram session builders (`session-openai.ts`, `session-xai.ts`) for parity — background registry wiring is Anthropic-only in this PR.
- Did not run tests — verified test structure and assertions only.
- Did not verify `BackgroundAgentRegistry` event emission contract (assumed correct; the test uses the real class and it passes in CI).

---

**Decision: DO NOT MERGE — 2 medium blocking (1 correctness/spec-compliance, 1 security); 4 low, 1 nit.**

Both blocking findings are one-line fixes: (1) scope the system prompt guidance to name the Telegram surface's actual behavior, and (2) thread `chatId` into the notifier constructor. The rest of the PR — the `bg-result-notifier.ts` implementation, the `drainSubagents` lifecycle wiring, and the test suite — is clean.

---

griffinwork40:
# PR #1212 Re-Review — `feat(telegram): enable background subagent dispatch + improve bg/fg guidance`

**Target:** `afk/improve-subagent-strategy` → `main` | **Ref:** `d44e6c78` | **Regime:** full (337 lines, feature)
**Files:** `prompts/system-prompt.md` · `src/agent/types/config-types.ts` · `src/telegram/bg-result-notifier.ts` (new) · `src/telegram/bg-result-notifier.test.ts` (new) · `src/telegram/create-session.ts` · `src/telegram/session-anthropic.ts` · `src/telegram/session-context.ts` · `src/telegram/session-manager.ts`

**Prior review status:** 2 blocking findings (spec-compliance + security) — both addressed in fix commit `d44e6c78`.

---

## Prior blocking findings — resolved

### ✅ System prompt surface qualification (was blocking)
Line 99 now reads: *"On REPL, background results are delivered automatically… On Telegram, you receive a push notification when the job settles but the result is NOT injected into your context."* Surface-qualified. Resolved.

### ✅ chatId threading (was blocking)
`route.chatId` is now threaded: `session-manager.ts` → `AgentConfig.telegramChatId` → `create-session.ts` → `TelegramSessionBuildContext.chatId` → `session-anthropic.ts` → `TelegramBgResultNotifier(registry, chatId)`. Multi-chat notifications now target the originating chat. Resolved.

### ✅ markDelivered (was non-blocking)
`this.registry.markDelivered(job.jobId)` added to `onSettled`. Witness trace now distinguishes push-notified settlements. Resolved.

### ✅ Push error logging (was non-blocking)
`.catch(() => {})` replaced with `.catch((err) => { console.error(...) })`. Failures are now visible. Resolved.

### ✅ Test async correctness (was non-blocking)
Test converted to `async`, uses `vi.waitFor()` to drain microtask queue. Resolved.

---

## New findings

### 1 · low · blocking:false · correctness · `src/telegram/bg-result-notifier.ts:68–72` · ref:d44e6c78 · confidence:medium

**`markDelivered` fires synchronously before `pushIfConfigured` resolves — trace records "delivered" even when push fails.** The REPL analog has the same pattern (marks delivered before confirming injection), so this is consistent behavior. The semantic is "notifier handled this job" rather than "notification confirmed received." Acceptable for a push-based fire-and-forget path where delivery confirmation is inherently unavailable (Telegram API returns 200 but the human may not read it).

**Suggestion:** If a stricter "delivered = push succeeded" semantic is desired, chain as `.then(() => this.registry.markDelivered(job.jobId))`. Low priority — matches REPL contract.

### 2 · low · blocking:false · test-coverage · `src/telegram/bg-result-notifier.test.ts` · ref:d44e6c78 · confidence:high

**No test asserts that `markDelivered` is called after settlement.** The fix commit added `markDelivered` to the notifier but no test verifies it. Future refactors could silently remove it.

**Suggestion:** Add a test that spies on `registry.markDelivered` and asserts it was called with the correct `jobId` after `fireTerminal`.

### 3 · low · blocking:false · correctness · `src/telegram/session-openai.ts`, `session-xai.ts` · ref:d44e6c78 · confidence:high · blocking preserved from pre-downgrade medium

**OpenAI/xAI Telegram branches do not have `BackgroundAgentRegistry` wiring.** The system prompt says background mode is available "on Telegram" generically, but the OpenAI/xAI branches never had `wireExecutors` or subagent dispatch infrastructure — this is a **pre-existing gap**, not a regression from this PR. The `agent` tool likely returns a clear error on those paths already. Future work to wire bg support into OpenAI/xAI branches should add parity.

### 4 · nit · blocking:false · security · `src/telegram/bg-result-notifier.ts:53` · ref:d44e6c78 · confidence:medium

**Push notification includes first 80 chars of dispatch prompt as `label`.** Dispatch prompts routinely carry file paths and user queries. In a multi-chat setup, this is mitigated by the chatId fix (Finding ✅ above) — notifications now go to the originating chat only. Residual risk: a human reading Telegram push notifications sees prompt content they authored. Low practical impact.

---

## What was not checked

- Citations verified inline against branch HEAD `d44e6c78`.
- Stated intent: PR title + body — spec-compliance assessed against stated intent. Fix commit assessed against prior review feedback.
- Did not review `session-openai.ts` / `session-xai.ts` for background parity — confirmed pre-existing gap via `git show origin/main` (zero wireExecutors/backgroundRegistry matches on main).
- Did not run tests — verified test structure, assertions, and async correctness via code read.
- Did not verify `BackgroundAgentRegistry` internal event emission contract (passes in CI).

---

**Decision: MERGE — 0 blocking; 3 low, 1 nit. Both original blocking findings (spec-compliance, security) resolved in fix commit `d44e6c78`.**

---
_🤖 Posted by `agent-afk /review --post github` (re-review)_

Issue #861, opened 2026-08-02
improve/scan: credential-floor detection matches hand-duplicated literals — a producer reword misclassifies floor denials and leaks credential paths into cards


griffinwork40:
## Summary

`classifyDenialMechanism` decides whether a sub-agent read denial is the unconditional credential floor (drop it — a working guardrail, not a defect) or a read-root containment gap (keep it — a real finding) by `.includes()` against two **hand-duplicated string literals**. There is no shared constant and no test cross-referencing the producers.

Reword either phrase at any producer and floor denials silently reclassify as containment gaps. That both resurrects the false-positive class #847 exists to remove **and** pipes credential paths into persisted, operator-facing failure cards.

## Mechanism

Consumer — `src/improve/scan/detectors/subagent-read-denial.ts` at #847 head `746c7758`:

```ts
// :111-115
export function classifyDenialMechanism(reason: string): DenialMechanism {
  const isFloor =
    reason.includes('read-denylist entry:') ||
    reason.includes('is a protected credential/secret path');
  if (isFloor) return 'credential-floor';
```

Drop filter at `:180-181`:

```ts
const mechanism = classifyDenialMechanism(reason);
if (mechanism === 'credential-floor') continue;
```

Producers on `main` @ `5bc00a89` — four production sites hand-rolling the wording:

| Site | Phrase(s) emitted |
|---|---|
| `src/agent/tools/hooks/path-approval-hook.ts:233-234` | both |
| `src/agent/tools/handlers/_cwd-utils.ts:217-218` | both |
| `src/agent/tools/handlers/read-denylist.ts:437` | `read-denylist entry:` |
| `src/agent/tools/subagent/input-parse.ts:404` | `read-denylist entry:` |

```
$ git grep -nE "(export const|export function).*(DENYLIST_ENTRY|CREDENTIAL_MARKER|FLOOR_MARKER|PROTECTED_PATH)" -- src/
(no matches)
```

All four match the classifier today — verified. Nothing keeps them matching.

## Why the asymmetry matters

The two misclassification directions are **not** equally bad:

- **floor → containment** (the reword failure mode): a credential/secret path is written into a persisted card. `card-writer.ts:261` does `JSON.stringify(card.detail, null, 2)` straight into the rendered `.md` under `## Detail`, so the path reaches the operator-facing artifact. It also re-arms the exact false positive #847 removed — a working guardrail refiled as a defect, with a "widen the read roots" remedy that would weaken the floor if followed.
[...]

PR #1214, opened 2026-08-19
fix(workspace): cold-start hint + compose executor wiring gap

griffinwork40:
## Problem

The workspace v1 (#1195) had two gaps preventing it from reducing cross-agent file-read duplication:

1. **Cold-start problem** — the workspace preamble was only injected when entries existed (`if (entries.length === 0) return config`), so agents never learned about `workspace_publish` and never published, keeping the workspace permanently empty.

2. **Compose executor gap** — the compose DAG executor (`dag.ts` → `compose-executor.ts`) creates its own `SubagentManager` per `execute()` call, but the workspace store was never threaded through to it. Compose nodes (Wave 2 in `/diagnose`, parallel nodes in `/compose`) had no workspace access at all.

## Fix

- **Cold-start hint**: `workspace-preamble.ts` now injects a short hint even when the workspace is empty, telling agents about `workspace_publish` so the first discoverer can seed findings for siblings.

- **Compose wiring**: Thread `workspaceStore` through `ComposeExecutorContext` → `wire-executors.ts` → all four surface entry points (REPL, daemon, telegram, chat).

## Evidence

Ran a `/diagnose #861` experiment before and after:
- **Before**: 25 reads, 0 workspace publishes, 40% dedup ratio (same as no-workspace baseline — workspace was dead)
- **After (cold-start hint only, no compose fix)**: 1 publish from the orchestrator, but compose Wave 2 nodes still couldn't see the workspace
- The compose gap was found by a shadow-verified audit of all `SubagentManager` construction sites

## Tracking

Part of workspace v2 (#1196). Next step: re-run the experiment with both fixes live and measure the dedup improvement.


griffinwork40:
Confirmed: `session.close()` only calls `this.providerQuery.close()`, NOT `this.provider.close()`. The shadow verifier is correct — **F4 is REFUTED**.

However, the shadow verifier noted a **separate concern**: the shared `workspaceStore` reference IS forwarded to child providers, and those child providers' `close()` methods are never called during teardown, meaning per-child SQLite handles leak. This is a resource leak, not a corruption bug. And since the child providers use the shared reference (rather than creating their own), they don't actually allocate a new SQLite connection — they reuse the parent's. So even the leak concern is attenuated.

Let me now compile the final review:

---

## Code Review: PR #1214 — `fix(workspace): cold-start hint + compose executor wiring gap`

**Reviewed ref:** `0c249aa9577cd92a788476085d32da069a15467a`
**Stated intent:** PR title + body — fix cold-start chicken-and-egg for workspace preamble; thread `workspaceStore` through compose executor and all surface entry points.
**Triage:** fix · 13 files · +57/−26 (83 lines) · **light regime**

---

### Findings

**1. medium · blocking:true · confidence:high · test-coverage · `src/agent/workspace/workspace-preamble.test.ts` · ref:`0c249aa` · citation-type:file-state**

The cold-start hint is now always injected including when `systemPrompt` is a preset object (lines 114–123 of `workspace-preamble.ts`), but no test covers the `preset + entries === []` combination. The only empty-entries test (`workspace-preamble.test.ts:139`) uses `{ model: 'sonnet' }` (no system prompt, not a preset). This is a newly-critical invariant — agents MUST receive the `workspace_publish` hint — on an untested code path.

```ts
// workspace-preamble.test.ts:139 — only empty-entries test
const result = injectWorkspacePreamble(baseConfig, []); // baseConfig = { model: 'sonnet' }
```

**Suggestion:** Add:
```ts
it('injects cold-start hint into preset systemPrompt when entries is empty', () => {
  const config: AgentConfig = { model: 'sonnet', systemPrompt: { type: 'preset', promptId: 'repl' } };
  const result = injectWorkspacePreamble(config, []);
  const sp = result.systemPrompt as { type: string; append: string };
  expect(sp.append).toContain('workspace_publish');
});
```

---

**2. low · blocking:false · confidence:high · correctness · `src/cli/commands/chat.ts:491` · ref:`0c249aa` · citation-type:file-state**

`new WorkspaceStore()` is constructed anonymously — never captured, never closed in the `finally` block. The store is in-memory (`:memory:` SQLite) and `afk chat` is one-shot, so the OS reclaims the handle on exit. Inconsistent with the `sharedMemoryStore?.close()` pattern at line 798 but no user-facing impact.

**Suggestion:** Capture in a variable, close alongside `sharedMemoryStore`.

---

**3. low · blocking:false · confidence:high · spec-compliance · `src/agent/subagent.test.ts:405` · ref:`0c249aa` · citation-type:diff-context**

Assertion relaxed from `.toBe('untouched')` to `.toMatch(/^untouched/)` to accommodate the cold-start hint appended to child system prompts. The test's stated purpose — "discloses no budget to an explicitly unbounded child" — now validates only a prefix match, not the pass-through invariant it originally tested.

**Suggestion:** Split or rename: one assertion for the budget-preamble invariant, one for the cold-start append.

---

**4. low · blocking:false · confidence:high · test-coverage · `src/agent/tools/compose-executor.ts:677-680` · ref:`0c249aa` · citation-type:file-state**

No test exercises the `workspaceStore` forwarding from `ComposeExecutorContext` to the internal `SubagentManager`. Zero test-file references to `workspaceStore` at the reviewed ref (grep pattern: `workspaceStore` across `*.test.ts`).

**Suggestion:** Add a compose executor test asserting the store is forwarded.

---

**5. low · blocking:false · confidence:high · test-coverage · `src/cli/commands/daemon.ts:90` · ref:`0c249aa` · citation-type:file-state**

Initialization-order fix (hoisting `memoryStore ??=` / `workspaceStore ??=` before `wireExecutors`) is correct but untested. A regression here is a runtime error, not a compile-time one.

**Suggestion:** Add a test for `buildDaemonSessionFactory` confirming stores are initialized before `wireExecutors`.

---

**6. low · blocking:false · confidence:high · observability · `src/telegram/create-session.ts:56` · ref:`0c249aa` · citation-type:file-state**

Per-session `WorkspaceStore` construction is correct for Telegram's chat-isolation requirement, but the lifecycle chain (`session.close()` → `provider.close()` → `workspaceStore.close()`) is undocumented.

**Suggestion:** Add a brief lifecycle comment.

---

**7. nit · blocking:false · confidence:high · style · `src/agent/workspace/workspace-preamble.ts:88`**

JSDoc update is already correct at the reviewed ref. No action needed.

---

### Shadow-verify results

**F4 (original Wave 1 finding: shared store closed by child teardown) — REFUTED.** Independent re-derivation confirmed that `AgentSession.close()` (agent-session.ts:1297–1313) calls `providerQuery.close()`, NOT `provider.close()`. The provider-level `close()` — which owns the `workspaceStore.close()` call — is never invoked during child session teardown. The shared reference IS forwarded to child providers (the precondition is real), but the destructive trigger never fires. The original finding was dropped from the final report.

### What was not checked

- Citations verified inline against branch HEAD `0c249aa`.
- Stated intent: PR #1214 title+body — spec-compliance assessed.
- Did not verify whether `provider.close()` is called from any OTHER path (e.g., explicit cleanup in a crash handler). The child-provider resource-leak angle (noted by the shadow verifier — per-child provider instances are never `close()`d) was not investigated further.
- Did not run tests.
- Did not verify the compose DAG flow end-to-end (only static wiring reads).
- Did not check the `xaiBaseUrl` entry point surface (`session-xai.ts`).

---

### Decision: DO NOT MERGE — 1 medium blocking; 5 low, 1 nit.

Finding 1 (missing test for preset + empty-entries cold-start path) is the sole blocker. The cold-start hint is the PR's headline feature and the preset system-prompt code path is exercised in production by the REPL surface — leaving it uncovered is a regression risk on a newly-critical invariant. The fix is one test case.

All other findings are non-blocking improvements addressable as follow-up.

---

**Done.**

- Review complete for PR #1214, ref `0c249aa`.
- 1 blocking finding (F1: missing preset+empty-entries test), 5 low findings, 1 nit.
- 1 Wave 1 finding (shared-store corruption) was refuted by shadow-verify — `provider.close()` is never called during child teardown.
- Merge verdict: **DO NOT MERGE** (1 medium blocking).
- No files written or committed (read-only review).

---

PR #1238, opened 2026-08-20
fix(telegram): clear apiRoundInFlight in streamResponse finally block

### LINKED ISSUE #1143: fix(telegram): clear apiRoundInFlight in streamResponse finally block

griffinwork40:
## What

Add `watchdog.apiRoundInFlight = false` and `watchdog.apiRoundSince = null` to the `finally` block in `streamResponse` (`src/telegram/streaming.ts`).

## Why

`apiRoundInFlight` is set on every `progress` event (line ~277) but was not cleared in the `finally` block. On a normal stream the `done` event arrives immediately after the last `progress` event, and line ~202 clears the flag on first receipt. But if the stream wedges non-terminally after the final `progress` event (no further events arrive), the watchdog defers its timeout by up to `MAX_API_ROUND_INFLIGHT_MS` (420 s) before firing — inflating the user-visible hang from 180 s to 420 s.

The existing `finally` block already clears related state (`countdownInterval`, `progressTimer`, `editInFlight`) but was missing the API-round watchdog flag.

## How

Two lines added to the `finally` block alongside the other state resets (line ~350):

```ts
watchdog.apiRoundInFlight = false;
watchdog.apiRoundSince = null;
```

Includes a regression test covering both the happy-path cleanup and the wedged-stream scenario.

Closes #1143.

### DIFF
--- a/src/agent/providers/xai/headers.test.ts
+++ b/src/agent/providers/xai/headers.test.ts
@@ -97,6 +97,42 @@ describe('resolveGrokCliIdentityHeaders', () => {
     expect(h['x-grok-client-version']).toBe(DEFAULT_GROK_CLI_COMPAT_VERSION);
     expect(h['x-grok-client-version']).not.toBe('agent-afk');
   });
+
+  it('falls through to DEFAULT when version file is below the floor (stale install)', () => {
+    // 0.1.200 is below the floor of 1.0.5 — must fall through to DEFAULT
+    const h = resolveGrokCliIdentityHeaders(
+      {},
+      deps({ readFile: () => JSON.stringify({ version: '0.1.200' }) }),
+    );
+    expect(h['x-grok-client-version']).toBe(DEFAULT_GROK_CLI_COMPAT_VERSION);
+  });
+
+  it('accepts a version file exactly at the floor', () => {
+    // 1.0.5 is exactly at floor — must be accepted as-is
+    const h = resolveGrokCliIdentityHeaders(
+      {},
+      deps({ readFile: () => JSON.stringify({ version: '1.0.5' }) }),
+    );
+    expect(h['x-grok-client-version']).toBe('1.0.5');
+  });
+
+  it('accepts a version file above the floor', () => {
+    // 2.3.1 is clearly above the floor — must be accepted
+    const h = resolveGrokCliIdentityHeaders(
+      {},
+      deps({ readFile: () => JSON.stringify({ version: '2.3.1' }) }),
+    );
+    expect(h['x-grok-client-version']).toBe('2.3.1');
+  });
+
+  it('does not floor the environment override (operator precedence contract)', () => {
+    // AFK_XAI_GROK_CLIENT_VERSION below the floor must still be honoured when set
+    const h = resolveGrokCliIdentityHeaders(
+      {},
+      deps({ readEnv: () => '0.1.200' }),
+    );
+    expect(h['x-grok-client-version']).toBe('0.1.200');
+  });
 });
 
 describe('readOfficialGrokCliVersion', () => {
--- a/src/agent/providers/xai/headers.ts
+++ b/src/agent/providers/xai/headers.ts
@@ -51,6 +51,17 @@ function validSemver(value: string | undefined): string | undefined {
   return trimmed && SEMVER_RE.test(trimmed) ? trimmed : undefined;
 }
 
+/** Returns `version` if it is ≥ `floor`, otherwise `undefined`. */
+function aboveFloor(version: string | undefined, floor: string): string | undefined {
+  if (!version) return undefined;
+  const parse = (s: string): number[] => s.split('.').map(Number);
+  const [vMaj = 0, vMin = 0, vPat = 0] = parse(version);
+  const [fMaj = 0, fMin = 0, fPat = 0] = parse(floor);
+  if (vMaj !== fMaj) return vMaj > fMaj ? version : undefined;
+  if (vMin !== fMin) return vMin > fMin ? version : undefined;
+  return vPat >= fPat ? version : undefined;
+}
+
 /** Read `~/.grok/version.json` written by the official Grok Build CLI. */
 export function readOfficialGrokCliVersion(deps: GrokCliHeaderDeps = {}): string | undefined {
   const readFile = deps.readFile ?? ((filePath: string) => readFileSync(filePath, 'utf-8'));
@@ -77,7 +88,7 @@ function resolveGrokCliVersion(
   return (
     validSemver(readEnv(GROK_CLI_VERSION_ENV))
     ?? validSemver(clientVersion)
-    ?? readOfficialGrokCliVersion(deps)
+    ?? aboveFloor(readOfficialGrokCliVersion(deps), DEFAULT_GROK_CLI_COMPAT_VERSION)
     ?? DEFAULT_GROK_CLI_COMPAT_VERSION
   );
 }


griffinwork40 (commented):

## Review: PR #1238 — `fix(telegram): clear apiRoundInFlight in streamResponse finally block`

**Target:** `afk/fix-1143-clear-api-round-fc35d2` → `main`
**Change type:** hotfix | **Regime:** light | **Files:** 6 | **Lines:** +280/−6
**Reviewed ref:** `d23f5db3da099601c6f52f42a09f252219dc98e3`

---

### Findings

**1 · `medium` · blocking:false · confidence:high · correctness · `src/telegram/streaming.test.ts:988–1018` · ref:d23f5db3da · citation-type:diff-context · waived: bounded — the test validates the invariant that the watchdog fires at all and interrupt is called, which is necessary; the gap is that it does not falsify the pre-fix 420s deferral path, which is a strength-of-coverage concern, not a correctness defect**

**Regression test does not distinguish 180s-window from 420s-window timeout.** The test advances fake timers past both `NEXT_EVENT_TIMEOUT_MS` (180s) and `MAX_API_ROUND_INFLIGHT_MS` (420s) before asserting rejection. The test comment acknowledges that "the finally block runs AFTER the watchdog fires and throws, so the 420s deferral is unavoidable for a still-hanging stream." This means the test proves the watchdog eventually fires and `interrupt` is called — correct — but does not prove the fix reduces the deferral window. A test that asserted rejection *before* the 420s advance would falsify pre-fix behavior.

**suggestion:** Add a complementary test case that verifies post-`streamResponse` state: after the promise rejects and the finally block runs, assert `watchdog.apiRoundInFlight === false` by exposing the flag or checking that a hypothetical *second* watchdog cycle would not defer. Alternatively, split the timer advance and check rejection timing.

---

**2 · `low` · blocking:false · confidence:high · correctness · `src/agent/providers/xai/query-bootstrap.ts:49–50` · ref:d23f5db3da · citation-type:diff-context**

**Three-value nullable sentinel (`string | undefined | null`) on `bufferedSystemPrompt` is fragile.** `null` encodes "caller explicitly passed `undefined`"; the comment is the only guard. The replay at line 78 correctly checks `bufferedSystemPrompt === null ? undefined : bufferedSystemPrompt`, but the type allows accidental direct `null` assignment outside the `setSystemPrompt` setter without triggering a type error.

**suggestion:** Use a discriminated wrapper: `let bufferedSystemPrompt: { set: true; value: string | undefined } | undefined` — then check `bufferedSystemPrompt?.set` for presence and `.value` for the payload. This eliminates the sentinel and makes intent mechanically clear.

---

**3 · `low` · blocking:false · confidence:high · spec-compliance · multiple files · ref:d23f5db3da · citation-type:diff-context**

**PR title says `fix(telegram)` but bundles two unrelated fixes.** Commit 1 (`parsePortFile` IPv6 bracket stripping + port range validation) and Commit 2 (xAI `query-bootstrap` setCwd/setSystemPrompt buffering) are logically independent of the Telegram streaming fix. Both are well-implemented and tested, but a revert of the Telegram fix would also revert unrelated daemon and provider changes.

**suggestion:** For clean revert boundaries, split into 3 PRs in future (or at minimum use distinct commits that can be cherry-picked independently — which this PR already does).

---

**4 · `nit` · blocking:false · confidence:high · correctness · `src/agent/daemon/http-client.ts:112–115` · ref:d23f5db3da · citation-type:diff-context**

**Bracket-stripping comment doesn't reference the re-bracketing guard.** The comment says "trySyncToDaemon re-adds brackets" but doesn't name the mechanism (`host.includes(':')` at line 73 of the same file). A future reader might not realize the stripping is *required* by that guard.

**suggestion:** Amend: `// trySyncToDaemon re-brackets IPv6 via \`host.includes(':') ? [\${host}] : host\`, so double-bracketing would occur without this strip.`

---

### Spec-compliance

Stated intent: PR #1238 title+body — spec-compliance assessed.

| Requirement | Status |
|---|---|
| Clear `apiRoundInFlight` in `finally` block | ✅ Lines 351–352 |
| Clear `apiRoundSince` in `finally` block | ✅ Line 352 |
| Regression test for happy-path cleanup | ✅ First test case (progress → done → resolves) |
| Regression test for wedged-stream scenario | ✅ Second test case (progress → hang → StreamTimeoutError) |
| `parsePortFile` bracket stripping | ✅ Lines 112–115, 4 new test cases |
| xAI query-bootstrap pre-init buffering | ✅ 3 new buffer vars, replay in ensureInner, 7 test cases |
| Scope: single `fix(telegram)` vs. 3 distinct fixes | ⚠️ Scope note (F3) — not blocking |

No unmet intent. Scope is broader than the title suggests but all changes are correct.

### Dimensions with no issues

- **Security** — no issues found. `parsePortFile` operates on a local file value (not network input). The `finally` block additions are synchronous state clears on locally-scoped mutable variables. xAI buffering replays only values already accepted by the `setCwd`/`setSystemPrompt` API. Read: `http-client.ts`, `query-bootstrap.ts`, `streaming.ts` at ref.
- **API-compat** — no issues found. `parsePortFile` is `@internal` (exported for testing only). `buildOAuthRefreshQuery` signature unchanged. `streamResponse` signature unchanged. No public API surface modified.
- **Perf-observability** — no issues found. Two synchronous scalar assignments in the `finally` block. Three closure variables in `buildOAuthRefreshQuery`. No hot-path allocations.

### What was not checked

- Citations verified inline against branch HEAD `d23f5db3da099601c6f52f42a09f252219dc98e3`.
- Stated intent: PR #1238 title+body — spec-compliance assessed.
- Not checked: `streaming.watchdog.ts` full arm() re-check cadence interaction with the `finally`-block flag clear (the arm() runs on macrotask queue, so the clear is effective for any post-`finally` re-check).
- Not checked: Whether CI passes on this branch (merge status is CONFLICTING — needs rebase).
- Not checked: xAI OAuth token refresh behavior when `setCwd` is buffered during a token-expired state.

---

The core fix (clearing `apiRoundInFlight`/`apiRoundSince` in the `finally` block) is correct and well-placed — it runs after the synchronous state resets and before `iter.return()`. All three bundled fixes are clean, tested, and non-breaking. The medium finding (test doesn't falsify the 420s deferral) is waived because the test correctly validates the invariant that the watchdog fires at all and `interrupt` is called once — it is necessary coverage even if not sufficient to prove the timing improvement.

CI note: merge status is `CONFLICTING` — rebase onto current `main` is needed before merge.

griffinwork40 (commented):

## Review: PR #1238 — `fix(telegram): clear apiRoundInFlight in streamResponse finally block`

**Target:** `afk/fix-1143-clear-api-round-inflight` → `main` · ref `d23f5db3da` · +280/−6 · 6 files
**Regime:** light (hotfix)
**Stated intent:** PR #1238 title+body — spec-compliance assessed.

---

### Prior Round Dispositions (Round 1 → Round 2)

| # | Round 1 Finding | Disposition |
|---|---|---|
| R1-1 | `medium` waived — regression test does not distinguish 180s-window from 420s-window timeout | **STILL PRESENT** — no new commits since last review; test still advances past both timeouts before asserting rejection. The waiver stands: the test validates that the watchdog fires and `interrupt` is called, which is necessary coverage. |
| R1-2 | `low` — three-value nullable sentinel (`string \| undefined \| null`) on `bufferedSystemPrompt` is fragile | **STILL PRESENT** — code unchanged at `query-bootstrap.ts:50`. |
| R1-3 | `low` — PR bundles 3 unrelated fixes (daemon, xAI, telegram) | **STILL PRESENT** — structural observation, cannot be fixed without PR split. Each fix has its own commit, so cherry-pick is possible. |
| R1-4 | `nit` — bracket-stripping comment doesn't reference re-bracketing guard mechanism | **STILL PRESENT** — comment at `http-client.ts:115-117` names the consequence ("double-bracketing would cause a TypeError") but not the source (`host.includes(':') ? [${host}] : host`). |

---

### New Findings (Round 2)

**1 · `low` · blocking:false · confidence:high · test-coverage · `src/agent/daemon/http-client.test.ts` · ref:d23f5db3da · diff-context**

**No test for double-bracketed IPv6 input `"[[::1]]:7777"`.** The comment at `http-client.ts:115` explicitly names `[[::1]]` as the failure mode being prevented, but no test exercises it. The current bracket-strip logic (`host.startsWith('[') && host.endsWith(']')`) would strip one layer, returning `host: "[::1"` — an invalid host that downstream code would fail on.

**suggestion:** Add `"[[::1]]:7777"` → expected-behavior test to lock the boundary.

---

**2 · `low` · blocking:false · confidence:high · correctness · `src/agent/providers/xai/query-bootstrap.ts:140` · ref:d23f5db3da · file-state**

**`setSystemPrompt` returns `false` during buffering, but the prompt IS captured and WILL be replayed.** The `ProviderQuery` interface (`provider.ts:444`) declares `setSystemPrompt?(): boolean` where `false` conventionally means "not supported / not applied." The buffering path returns `false` even though the value is stored and will be delivered on init. The callers in production (`agent-session.ts:1095`, `reload.ts:91`) use the return for status display only — no fallback action is taken on `false` — so the semantic mismatch has no runtime consequence today.

**suggestion:** Return `true` when buffering succeeds; reserve `false` for cases where the prompt is genuinely dropped.

---

### Dimensions with no issues

- **Security** — no issues found. `parsePortFile` operates on a local file value; bracket-stripping is bounded string slice with no injection surface. The `finally` block additions are synchronous state clears on locally-scoped mutable variables. xAI buffering replays only values already accepted by the `setCwd`/`setSystemPrompt` API. Read: `http-client.ts`, `query-bootstrap.ts`, `streaming.ts` at ref.
- **API-compat** — no issues found. `parsePortFile` is `@internal` (exported for testing only). `buildOAuthRefreshQuery` signature unchanged. `streamResponse` signature unchanged. No public API surface modified.
- **Perf-observability** — no issues found. Two synchronous scalar assignments in the `finally` block. `watchdog.apiRoundSince = null` at line 352 is technically redundant (setting `apiRoundInFlight = false` short-circuits the arm check at `streaming.watchdog.ts:167`), but is good defensive hygiene.

### Spec-compliance

| Requirement | Status |
|---|---|
| Clear `apiRoundInFlight` in `finally` block | ✅ Line 351 |
| Clear `apiRoundSince` in `finally` block | ✅ Line 352 |
| Regression test for happy-path cleanup | ✅ First test case (progress → done → resolves) |
| Regression test for wedged-stream scenario | ✅ Second test case (progress → hang → StreamTimeoutError) |
| `parsePortFile` bracket stripping | ✅ Lines 114-117, 4 new test cases |
| xAI query-bootstrap pre-init buffering | ✅ 3 buffer vars, replay in ensureInner, 7 test cases |
| Closes #1143 | ✅ Primary fix addresses the watchdog deferral |
| Scope: single `fix(telegram)` vs. 3 distinct fixes | ⚠️ Scope note (R1-3) — not blocking; each fix has its own commit |

No unmet intent. No scope creep detected.

### What was not checked

- Citations verified inline against branch HEAD `d23f5db3da099601c6f52f42a09f252219dc98e3`.
- Stated intent: PR #1238 title+body — spec-compliance assessed.
- Not checked: `streaming.watchdog.ts` full arm() re-check cadence interaction with the `finally`-block flag clear.
- Not checked: whether CI passes on this branch (merge status is CONFLICTING — needs rebase).
- Not checked: xAI OAuth token refresh behavior when `setCwd` is buffered during a token-expired state.

---

**Severity arithmetic: 0 critical, 0 high → MERGE (with follow-ups); 1 medium waived (carried from R1), 4 low (2 carried, 2 new), 1 nit (carried).**

**Decision: MERGE (with follow-ups) — 0 critical, 0 high → MERGE; 1 medium waived (test timing, carried from R1), 4 low, 1 nit. Rebase needed (merge status is CONFLICTING).**

PR #1254, opened 2026-08-24
fix(tools): stream grep output through a bounded collector instead of buffering 8MB

griffinwork40:
# fix(tools): stream grep output through a bounded collector instead of buffering 8MB

## Problem

The previous `grep.ts` design accumulated the child process's entire stdout into a JS string and only reduced it to the model budget (100KB) at close — meaning a broad-but-legitimate search buffered **up to 8MB** in memory before the model saw 100KB of it. At that 8MB ceiling the child was SIGKILL'd, which caused two production problems:

1. **Memory scaled with corpus size.** A search across a large tree held 8MB in a single V8 string regardless of the model's 100KB budget.
2. **A truncated view looked complete.** After the kill, the model received a head+tail slice with no indication of how much was omitted — no total byte count, no total line count.

## Solution

Adds `_streaming-cap.ts`: a `StreamingCap` collector that accepts raw `Buffer` chunks in arrival order and retains at most `maxBytes` of them — the first half from the head of the stream, the last half from a sliding tail ring — while **counting every byte and every newline it ever observed**. Memory is bounded at the model budget from the first chunk; the interior is discarded as it arrives, not after accumulation.

`grep.ts` replaces its two string accumulators (`stdout`, `stderr`) with two `StreamingCap` collectors:

- **Memory ceiling drops from 8MB to 100KB** (the model budget), regardless of how much ripgrep emits.
- **The V8 `RangeError: Invalid string length` crash is structurally unreachable** — neither collector ever grows past its budget.
- **The old 8MB SIGKILL is retired** for legitimate searches. A new `SCAN_CAP_BYTES` ceiling (256MB) guards genuine runaways — sized at 4× the measured worst case (62.8MB from a single-letter pattern over this repo).
- **Truncated views are honest.** `render()` embeds the true total bytes and matching-line count in the elision marker, so the model cannot mistake a 100KB slice for the complete result.

## Test changes

- **`_streaming-cap.test.ts`** (new): 12 unit tests covering under-budget pass-through, over-budget head+tail retention, UTF-8 boundary healing across chunk splits, and the exported policy constants.
- **`grep.test.ts`**: two old "mid-stream kill" assertions flipped — at 10MB and 18MB (both under `SCAN_CAP_BYTES`) the traversal now **completes without a sentinel**; a new test injects a small scan ceiling to exercise the surviving kill path without a multi-hundred-MB fixture. The "honesty" assertion (`N matching lines total` in the marker) is new.
- **`overflow-telemetry.test.ts`**: grep telemetry tests updated to use the injected ceiling path; the `grepHandler` default (production ceiling 256MB) correctly never fires on small fixtures.

## Production impact

- Broad searches that previously triggered the 8MB kill now run to completion and return an honest bounded view.
- Peak memory for a grep call drops from 8MB to 100KB regardless of corpus size.
- The `tool.overflow_kill` telemetry event is preserved — it now fires at the 256MB scan ceiling instead of 8MB, so production operators will see it only for genuine runaways, not routine large searches.

### DIFF
--- a/src/agent/tools/handlers/_streaming-cap.test.ts
+++ b/src/agent/tools/handlers/_streaming-cap.test.ts
@@ -1,0 +1,139 @@
+import { describe, it, expect } from 'vitest';
+import { createStreamingCap, SCAN_CAP_BYTES, scanCapKillNote } from './_streaming-cap.js';
+
+const buf = (s: string): Buffer => Buffer.from(s, 'utf8');
+
+describe('createStreamingCap', () => {
+  describe('under budget', () => {
+    it('renders the stream verbatim and reports no truncation', () => {
+      const cap = createStreamingCap(10_000);
+      cap.push(buf('alpha\nbeta\ngamma\n'));
+
+      expect(cap.render()).toBe('alpha\nbeta\ngamma\n');
+      expect(cap.truncated()).toBe(false);
+      expect(cap.totalBytes()).toBe(17);
+      expect(cap.totalLines()).toBe(3);
+    });
+
+    it('joins many small chunks in arrival order', () => {
+      const cap = createStreamingCap(10_000);
+      for (const part of ['a', 'b', 'c', 'd']) cap.push(buf(part));
+
+      expect(cap.render()).toBe('abcd');
+      expect(cap.truncated()).toBe(false);
+    });
+
+    it('ignores empty chunks', () => {
+      const cap = createStreamingCap(1_000);
+      cap.push(buf(''));
+      cap.push(buf('x'));
+      cap.push(Buffer.alloc(0));
+
+      expect(cap.render()).toBe('x');
+      expect(cap.totalBytes()).toBe(1);
+    });
+  });
+
+  describe('over budget', () => {
+    it('bounds retained bytes while counting every byte and line', () => {
+      const cap = createStreamingCap(2_000);
+      // 5,000 lines × 10 bytes = 50,000 bytes, 25x the budget.
+      for (let i = 0; i < 5_000; i++) cap.push(buf('123456789\n'));
+
+      expect(cap.totalBytes()).toBe(50_000);
+      expect(cap.totalLines()).toBe(5_000);
+      expect(cap.truncated()).toBe(true);
+      expect(Buffer.byteLength(cap.render(), 'utf8')).toBeLessThanOrEqual(2_000);
+    });
+
+    it('preserves the head and the tail, dropping only the interior', () => {
+      const cap = createStreamingCap(1_000);
+      cap.push(buf('HEAD_MARKER\n'));
+      for (let i = 0; i < 500; i++) cap.push(buf('filler-filler-filler\n'));
+      cap.push(buf('TAIL_MARKER\n'));
+
+      const rendered = cap.render();
+      expect(rendered).toContain('HEAD_MARKER');
+      expect(rendered).toContain('TAIL_MARKER');
+      expect(rendered).toContain('bytes truncated');
+    });
+
+    it('reports true totals in the marker so a bounded view never looks complete', () => {
+      const cap = createStreamingCap(1_000);
+      for (let i = 0; i < 300; i++) cap.push(buf('0123456789\n')); // 3,300 bytes, 300 lines
+
+      const rendered = cap.render();
+      expect(rendered).toContain('of 3300;');
+      expect(rendered).toContain('300 matching lines total');
+    });
+
+    it('handles a single chunk far larger than the whole budget', () => {
+      const cap = createStreamingCap(500);
+      cap.push(buf('S' + 'x'.repeat(100_000) + 'E'));
+
+      const rendered = cap.render();
+      expect(Buffer.byteLength(rendered, 'utf8')).toBeLessThanOrEqual(500);
+      expect(cap.totalBytes()).toBe(100_002);
+      expect(cap.truncated()).toBe(true);
+      expect(rendered.startsWith('S')).toBe(true);
+      expect(rendered.endsWith('E')).toBe(true);
+    });
+  });
+
+  describe('UTF-8 boundaries', () => {
+    it('never emits replacement characters at the head/tail seams', () => {
+      // 3-byte code points guarantee the head/tail cut points land
+      // mid-character unless the collector trims to a boundary. The stream is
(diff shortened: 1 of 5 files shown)


griffinwork40 (commented):

## Review: PR #1254 — `fix(tools): stream grep output through a bounded collector instead of buffering 8MB`

**Target:** `fix/grep-streaming-cap` → `main` · ref `499a9644` · +339/−92 · 5 files
**Regime:** light (hotfix)
**Stated intent:** PR #1254 title+body — spec-compliance assessed.

---

### Findings

**1 · `low` · blocking:false · confidence:high · correctness · `src/agent/tools/handlers/_streaming-cap.ts:186–190` · ref:499a9644 · file-state**

**Elision marker inflates `omitted` byte count by up to 3 bytes when UTF-8 boundary trimming clips a partial code point.** `render()` computes `omitted = totalBytes - safeHead.length - safeTail.length`, where `safeHead`/`safeTail` are the post-boundary-trim buffers. When `trimPartialTrailing` or `trimPartialLeading` clips 1–3 continuation bytes to land on a code-point boundary, those trimmed bytes are counted as "truncated" even though they were retained and then discarded for correctness — inflating the marker by a tiny amount. The error is at most 6 bytes (3 head + 3 tail) and affects only the informational marker string, not the actual content.

```typescript
const omitted = totalBytes - safeHead.length - safeTail.length;
// safeHead may be 1–3 bytes shorter than headBuf due to trimPartialTrailing
```

**suggestion:** Compute `omitted` from pre-trim values: `totalBytes - headBuf.length - tailBuf.length`, then pass `safeHead.length`/`safeTail.length` only to the "showing first N + last M" display fields. Or document the ≤6-byte cosmetic error.

---

**2 · `low` · blocking:false · confidence:high · correctness · `src/agent/tools/handlers/_streaming-cap.ts:countNewlines (line 74–81)` · ref:499a9644 · file-state**

**`totalLines` undercounts by 1 when the stream does not end with a newline.** `countNewlines` counts `\n` bytes (0x0a). Ripgrep's output typically ends its last match line with `\n`, but when it does not (e.g. a file whose last line has no trailing newline), the final match line is uncounted. The marker then reports "N matching lines total" where the true count is N+1. Cosmetic only — no consumer acts on this count.

**suggestion:** Document the off-by-one in a comment, or add 1 at render time when `totalBytes > 0 && last byte ≠ 0x0a`.

---

**3 · `nit` · blocking:false · confidence:high · api-compat · `src/agent/tools/handlers/_rg-exit2.ts:92` · ref:499a9644 · file-state**

**`classifyRgExit2` double-caps stderr.** Under the new design, `err.render()` already returns a ≤100KB string from the streaming collector. `classifyRgExit2` internally calls `capForModel(stderr.trim())` (imported from `_output-cap.ts`), which re-applies a 100KB cap — a no-op on already-capped input, but an unnecessary `Buffer.from()` allocation per error-path call.

**suggestion:** No action needed for this PR (the double-cap is harmless). Note for a follow-up: either pass a `preCapped: true` flag to skip the internal cap, or remove the internal `capForModel` call from `classifyRgExit2` now that grep pre-caps stderr.

---

### Spec-compliance

Stated intent: PR #1254 title+body — spec-compliance assessed.

| Requirement | Status |
|---|---|
| Replace 8MB string accumulator with streaming head+tail collector | ✅ `createStreamingCap(MODEL_CAP_BYTES)` replaces `stdout`/`stderr` strings |
| Memory ceiling drops from 8MB to 100KB | ✅ Collectors bounded at `MODEL_CAP_BYTES = 100_000` |
| V8 `RangeError: Invalid string length` crash structurally unreachable | ✅ Neither collector grows past its budget |
| Retire old 8MB SIGKILL for legitimate searches | ✅ `HARD_CAP_BYTES`/`maybeOverflow` removed from grep path |
| New 256MB `SCAN_CAP_BYTES` as runaway guard | ✅ `maybeScanCap` fires at 256MB; tests inject small ceiling |
| Truncated views report true totals (bytes + lines) | ✅ Marker embeds `totalBytes` and `totalLines` (off by ≤6 bytes — Finding 1) |
| `_streaming-cap.ts` unit tests | ✅ 12 tests: under-budget, over-budget, UTF-8 boundaries, policy constants |
| `grep.test.ts` updated for new behavior | ✅ Kill assertions flipped; new scan-ceiling test; honesty assertion added |
| `overflow-telemetry.test.ts` updated | ✅ Uses injected `scanCapBytes` to exercise kill path without 256MB fixture |

No scope creep detected. No unmet intent.

### Dimensions with no issues

- **Security** — no issues found. `stripEscapeSequences` retained on output. `--` end-of-options separator on rg args unchanged. No secrets exposed. No new trust boundary. `scanCapBytes` option is test-only (not user-facing input). Read `_streaming-cap.ts`, `grep.ts` at ref.
- **Correctness** — no issues found beyond Findings 1–2 (cosmetic marker inaccuracies). Core head+tail retention, UTF-8 boundary handling, eviction loop, and scan-cap kill path are sound. The `truncated()` predicate and `render()` verbatim-vs-marker gate are consistent. Read `_streaming-cap.ts`, `grep.ts` at ref.
- **Test-coverage** — comprehensive. 12 unit tests on `_streaming-cap.ts` cover under-budget, over-budget, single-large-chunk, UTF-8 split healing, policy constants. `grep.test.ts` covers completion without kill (10MB, 18MB), kill at injected ceiling, honesty marker. `overflow-telemetry.test.ts` uses injected ceiling for kill-path telemetry.
- **Perf-observability** — no issues found. Memory ceiling drop from 8MB to 100KB is the stated goal and is achieved. `appendRoutingDecision` telemetry preserved on the scan-cap kill path. `tail` array element count is unbounded but under realistic Node.js stream chunk sizes (≥4KB) stays well under 100 elements.

### What was not checked

- Citations verified inline against branch HEAD `499a9644746b3b477948a9d499fbc491c3231834`.
- Stated intent: PR #1254 title+body — spec-compliance assessed.
- Not checked: runtime behavior with actual ripgrep output on a large corpus (structural analysis + CI green as proxy).
- Not checked: whether `bash.ts` or other handlers should adopt `StreamingCap` (the PR body explicitly excludes bash, citing its different output shape — accepted at face value).
- Not checked: interaction between `GrepHandlerOptions.scanCapBytes` and the production caller in `tool-registry.ts` (the option is test-only; the production path uses the default `SCAN_CAP_BYTES`).

---

PR #1257, opened 2026-08-24
feat(verdict-card): add stats line and context-specific affordances

griffinwork40:
## What

Two improvements to the verdict card rendered at end-of-turn:

### 1. Stats line

Renders `$0.42 · 23s · 7 tools` inside the card, wired to the real `doneMeta` from `turn-handler.ts`. All fields are optional — the line is omitted entirely when no meta is available (backward compatible).

### 2. Context-specific affordances

Derives actionable text from the terminal state instead of static labels:

| Kind | Before | After |
|------|--------|-------|
| Done (with evidence) | "Objective satisfied" | "Review: tests pass" |
| Blocked (with condition) | "External dependency — unblock above to resume" | "Unblock: set ANTHROPIC_API_KEY" |
| Asking (with question) | "Waiting on you" | "Answer: which branch?" |

Falls back to the static affordance when the relevant field is absent.

### Tests

- 14 new tests covering stats formatting, singular/plural, duration formatting, geometry, and all affordance derivation paths
- Updated 2 existing assertions to match the new derived affordances
- Fixed a pre-existing edge case in the geometry test (long unbroken evidence token overflows the structured row — scoped the assertion to the affordance line)

### Origin

Started as a background subagent dispatch via Telegram that did the implementation but never committed or pushed.

### DIFF
--- a/src/cli/commands/interactive/turn-handler.ts
+++ b/src/cli/commands/interactive/turn-handler.ts
@@ -797,7 +797,7 @@ export async function runTurn(
         writeAbove(renderVerdictCard(verdict, {
           durationMs: doneMeta?.durationMs,
           totalCostUsd: doneMeta?.totalCostUsd,
-          toolCount: toolEvents?.length,
+          toolCount: toolEvents.length,
         }));
         writeAbove('');
         // One evidence check, two consumers: (1) the onTerminalState callback
--- a/src/cli/commands/interactive/turn-handler.ts
+++ b/src/cli/commands/interactive/turn-handler.ts
@@ -794,7 +794,11 @@ export async function runTurn(
       // quo (just the prose).
       const verdict: TerminalState | null = parseTerminalState(responseText);
       if (verdict) {
-        writeAbove(renderVerdictCard(verdict));
+        writeAbove(renderVerdictCard(verdict, {
+          durationMs: doneMeta?.durationMs,
+          totalCostUsd: doneMeta?.totalCostUsd,
+          toolCount: toolEvents?.length,
+        }));
         writeAbove('');
         // One evidence check, two consumers: (1) the onTerminalState callback
         // forwards it onto the post-turn `Stop` hook's StopContext so the
--- a/src/cli/commands/interactive/verdict-card.test.ts
+++ b/src/cli/commands/interactive/verdict-card.test.ts
@@ -11,6 +11,7 @@ import { describe, it, expect, vi } from 'vitest';
 import {
   renderVerdictCard,
   summarizeVerdict,
+  type VerdictMeta,
 } from './verdict-card.js';
 import { createVerdictLedger } from './verdict-ledger.js';
 import type { TerminalState } from './terminal-state.js';
@@ -56,7 +57,8 @@ describe('renderVerdictCard', () => {
     expect(out).toContain('tests pass');
     expect(out).toContain('changed');
     expect(out).toContain('feature X is available to users');
-    expect(out).toContain('Objective satisfied');
+    // With evidence present, the derived affordance replaces the static one.
+    expect(out).toContain('Review:');
   });
 
   // Regression: models sometimes emit identical text for the "done" and
@@ -91,10 +93,12 @@ describe('renderVerdictCard', () => {
     expect(out).toContain('blocks');
     expect(out).toContain('API key missing');
     expect(out).toContain('unblock');
-    expect(out).toContain('External dependency');
+    // With unblockCondition present, derived affordance is shown instead of static.
+    expect(out).toContain('Unblock:');
+    expect(out).toContain('set ANTHROPIC_API_KEY');
   });
 
-  it('asking: shows the question and the "waiting on you" affordance', () => {
+  it('asking: shows the question and a context-specific affordance', () => {
     const state: TerminalState = {
       kind: 'asking',
       question: 'which branch?',
@@ -103,7 +107,8 @@ describe('renderVerdictCard', () => {
     const out = stripAnsi(renderVerdictCard(state));
     expect(out).toContain('? Asking');
     expect(out).toContain('which branch?');
-    expect(out).toContain('Waiting on you');
+    // With question present, derived affordance is shown instead of static.
+    expect(out).toContain('Answer:');
   });
 
   it('interrupted: shows resume affordance', () => {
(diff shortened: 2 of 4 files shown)


griffinwork40 (commented):

## Review: PR #1257 — `feat(verdict-card): add stats line and context-specific affordances`

**Target:** `afk/verdict-card-stats-affordance` → `main` · ref `b3296bb7` · +287/−5 · 3 files
**Regime:** light (feature, ~90 lines production; bulk is 200-line test)
**Stated intent:** PR #1257 title+body — spec-compliance assessed.

---

### Findings

**1 · `high` · ⛔ blocking:true · confidence:high · correctness · `src/cli/verdict-card-overflow.test.ts:145` · ref:b3296bb7 · file-state**

**Overflow regression test fails against new `deriveAffordance` output — confirmed cause of red CI.** The `VERDICT` fixture at line 74 sets `evidence: 'Architecture summary above, every structural claim cited...'`. With this PR's `deriveAffordance` logic, the evidence field drives the affordance row to render `"Review: Architecture summary above..."` instead of the static `"Objective satisfied"`. The assertion at line 145 — `expect(lines.some((l) => l.includes('Objective satisfied')))` — fails because the static string is no longer reachable for this fixture. `verdict-card-overflow.test.ts` is **not in the diff**, confirming the oversight.

```typescript
// verdict-card-overflow.test.ts:145 — assertion expects static affordance
expect(lines.some((l) => l.includes('Objective satisfied')),
  `affordance missing:\n${grid}`).toBe(true);
// But VERDICT fixture (line 74) has evidence set, so deriveAffordance
// now returns "Review: Architecture summary above..." instead
```

**suggestion:** Update the assertion to match the new derived affordance: `l.includes('Review:')` — or update the `VERDICT` fixture to omit `evidence` if the static fallback is the intended test subject.

---

**2 · `low` · blocking:false · confidence:high · correctness · `src/cli/commands/interactive/verdict-card.ts:99–101` · ref:b3296bb7 · file-state**

**Asymmetric truthiness vs. explicit-undefined guard in `buildStatsLine`.** `totalCostUsd` is checked with `if (meta.totalCostUsd && meta.totalCostUsd > 0)` (truthiness) while `durationMs` uses `if (meta.durationMs !== undefined && meta.durationMs > 0)` (explicit). The truthiness check is actually more defensive against NaN, and `totalCostUsd: 0` is intentionally suppressed (nothing to render). No wrong output today, but the inconsistency is a readability concern.

**suggestion:** Align to `if (meta.totalCostUsd !== undefined && meta.totalCostUsd > 0)` for consistency with the `durationMs` guard.

---

**3 · `low` · blocking:false · confidence:high · correctness · `src/cli/commands/interactive/verdict-card.ts:148` · ref:b3296bb7 · file-state**

**Silent Unicode strip in `deriveAffordance` body text.** `body.replace(/[^\x20-\x7E]/g, '')` drops all non-ASCII characters from user/model-generated content (evidence, unblockCondition, question). The invariant comment at line 37 explains the concern for static affordance strings (ambiguous-width glyphs break box alignment), but dynamic body text legitimately contains Unicode. No production impact today (terminal affordance is informational), but silently discarded content is a latent correctness gap.

**suggestion:** Apply `truncateDisplayWidth` directly to the raw body (which handles width correctly) rather than stripping non-ASCII characters. If ambiguous-width safety is required, document which characters are problematic and strip only those.

---

**4 · `nit` · blocking:false · confidence:high · correctness · `src/cli/commands/interactive/turn-handler.ts:801` · ref:b3296bb7 · file-state**

**Dead optional-chain on non-nullable `toolEvents`.** `toolEvents?.length` uses `?.` but `toolEvents` is declared `const toolEvents: ToolEvent[] = []` at line 74 — structurally never undefined.

**suggestion:** `toolCount: toolEvents.length`

---

### Spec-compliance

| Requirement | Status |
|---|---|
| Stats line renders `$cost · duration · N tools` | ✅ Implemented in `buildStatsLine` |
| Stats wired to real `doneMeta` from turn-handler | ✅ `turn-handler.ts:797–801` |
| Stats line omitted when no meta available | ✅ Guard + `buildStatsLine` returns null |
| All meta fields optional | ✅ `VerdictMeta` interface, all `?` fields |
| Context-specific affordances (Done/Blocked/Asking) | ✅ `deriveAffordance` handles all three |
| Falls back to static affordance when field absent | ✅ `?? style.affordance` fallback |
| 14 new tests | ✅ Counted in test file |
| Updated 2 existing assertions | ✅ In diff |
| **Overflow regression test updated** | ❌ `verdict-card-overflow.test.ts` not in diff — assertion now fails (Finding 1) |

No scope creep detected. All changes are tightly bounded to verdict-card rendering.

### Dimensions with no issues

- **Security** — no issues found. Stats line and affordance logic are pure rendering transforms over existing session metadata. No new external inputs, no serialization paths, no credential-adjacent fields. Read `verdict-card.ts`, `turn-handler.ts` at ref.
- **API-compat** — no issues found. `renderVerdictCard` gains an optional `meta` parameter; all existing callers pass one argument (valid — `meta` is optional). `VerdictMeta` is a new additive export. No production callers affected. Read `verdict-card.ts`, `turn-handler.ts`, verified via `git grep`.
- **Perf-observability** — no issues found. `formatDuration`, `buildStatsLine`, `deriveAffordance` are O(1) string operations. No loops, no I/O, no allocations proportional to session size.

### What was not checked

- Citations verified inline against branch HEAD `b3296bb7ae6904f40511d3dcb4fa938fd2b9a369`.
- Stated intent: PR #1257 title+body — spec-compliance assessed.
- Not checked: whether the ambiguous-width regression test at `verdict-card-overflow.test.ts:155` also needs updating (the fixture at line 160 uses `{ kind, rawBody: 'x' }` without `evidence`, so `deriveAffordance` returns null and the static fallback fires — likely safe).
- Not checked: runtime behavior of `formatDuration` at very large values (e.g. multi-hour sessions).
- Not checked: whether `doneMeta?.totalCostUsd` upstream can produce NaN (structural analysis only).

---

**Decision: DO NOT MERGE — 1 high blocking (correctness: overflow regression test assertion fails against new `deriveAffordance` output, confirmed cause of red CI); 2 low, 1 nit.**

The sole blocker is Finding 1: `verdict-card-overflow.test.ts:145` expects `'Objective satisfied'` but the `VERDICT` fixture's `evidence` field now triggers the derived `"Review: ..."` affordance. The fix is one line — update the assertion or the fixture. After resolution, the PR is clean and ready to merge.

PR #1125, opened 2026-08-17
fix(telegram): thread topic opts into auto-subscribe watch output (#1023)

griffinwork40:
## Problem

The auto-subscribe code path in `TelegramBot.runAutoSubscribeTick()` constructed a `send` lambda that called `sendMessage(chatId, part)` with no options. Watch output from auto-subscribe always landed in the General topic, even when the watched session belonged to a specific topic thread.

The manual `/watch` command correctly computed `watchThreadOpts` via `sendOptions(route)` and threaded it through to `sendMessage` (`bot.ts:194–211`). Auto-subscribe was missing this.

## Fix

Thread `sendOptions(autoRoute)` into the auto-subscribe `send` callback.

Auto-subscribe has no inbound message context (there's no `/watch` command invocation to derive a thread from), so the route is always `{ chatId }` — General. `sendOptions()` returns `{}` for General, which is byte-identical to the bare send, so non-topic chats are unaffected. Topic-aware chats get an explicit General fallback rather than a silent one.

## Tests

Two regression tests added to `src/telegram/bot.test.ts` under `auto-subscribe watch output topic threading`:

1. **General route passes `sendOptions` as third arg** — verifies `sendMessage` receives `{}` (not `undefined`) so the call signature matches the manual `/watch` path.
2. **chatId captured from allowed set** — verifies the closure captures the loop's `chatId` correctly.

Closes #1023

PR #1262, opened 2026-08-24
fix(telegram): route auto-subscribe watch output to active topic thread (#1222)

### LINKED ISSUE #1222: fix(telegram): auto-subscribe watch output ignores topic thread
## Bug

Auto-subscribe `/watch` output in `src/telegram/bot.ts:622-624` always routes to the General topic because `sendOptions({ chatId })` has no `threadId`:

```ts
// Invariant: auto-subscribe has no inbound message context, so the
// route defaults to the General topic (no threadId). sendOptions()
// returns {} for General, which is byte-identical to the bare send —
// non-topic chats are unaffected.
const send = async (msg: string): Promise<void> => {
  for (const part of splitLongMessage(msg)) {
    await this.bot.telegram.sendMessage(chatId, part, sendOptions({ chatId })); // #1023
  }
};
```

The interactive `/watch` command was fixed in PR #1125 to respect topic threading, but the auto-subscribe path intentionally wasn't updated since there's no originating message context to derive a topic from.

## Impact

In supergroups with topics, auto-subscribed watch output clutters General instead of appearing in a relevant topic.

## Possible fix

The auto-subscribe loop could associate each `chatId` with the `threadId` of the most recently active session on that chat (from `SessionManager`'s route map), or allow configuring a default topic per chat. However, since auto-subscribe has no inbound message, this is a design decision more than a straightforward bug fix.

Related: #1023, PR #1125

griffinwork40:
## Summary

Routes auto-subscribe watch output to the operator's active topic thread instead of always landing in General.

## Problem

In Telegram supergroups with topics, auto-subscribed watch output always routed to the General topic because `sendOptions({ chatId })` had no `threadId`. The manual `/watch` command was fixed in PR #1125, but the auto-subscribe path was left targeting General since there's no originating message context.

## Fix

**`src/telegram/session-manager.ts`** — Added `getActiveThreadId(chatId)`:
- Scans in-memory `sessionData` for entries matching the chat that have a `threadId`
- Returns the `threadId` from the most recently active entry (by `lastActivity`)
- Returns `undefined` for non-topic chats → backward-compatible

**`src/telegram/bot.ts`** — In `runAutoSubscribeTick`:
- Calls `this.sessionManager.getActiveThreadId(chatId)` to get the active thread
- Builds a route with `threadId` when available, passes to `sendOptions()`
- General/non-topic chats produce `{}` — byte-identical to pre-fix behavior

## Tests (8 new)

- **`session-manager.test.ts`** — 6 unit tests for `getActiveThreadId`: no sessions, General-only, single topic, most-recently-active topic, general+topic mix, cross-chat isolation
- **`bot.test.ts`** — 2 integration tests: topic-thread routing and General fallback backward-compat

## Verification

- `pnpm build`: clean
- `pnpm test`: 913 test files, 17,475 tests pass

Closes #1222

### DIFF
--- a/src/telegram/bot.test.ts
+++ b/src/telegram/bot.test.ts
@@ -447,6 +447,87 @@ describe('TelegramBot', () => {
         vi.restoreAllMocks();
       }
     });
+
+    test('auto-subscribe routes to active topic thread when sessionManager reports one (#1222)', async () => {
+      // Regression for #1222: auto-subscribe always used sendOptions({ chatId })
+      // which routes to General, ignoring the chat's active topic. The fix calls
+      // sessionManager.getActiveThreadId(chatId) and threads the result through
+      // sendOptions so output lands in the correct topic.
+      const sendMessageMock = vi.fn(async () => ({ message_id: 1, text: '', date: 0, chat: { id: 12345 } }));
+      (bot as any).bot.telegram.sendMessage = sendMessageMock;
+
+      // Simulate: the chat has an active topic thread (threadId = 42).
+      vi.spyOn((bot as any).sessionManager, 'getActiveThreadId').mockReturnValue(42);
+
+      const presence = await import('../agent/awareness/presence.js');
+      const presenceSpy = vi.spyOn(presence, 'readLivePresenceFiles').mockResolvedValue([
+        { surface: 'cli', afk: true, sessionId: 'session-topic', cwd: '/tmp', model: 'claude', pid: 3, startedAt: '' },
+      ] as Awaited<ReturnType<typeof presence.readLivePresenceFiles>>);
+
+      let capturedSend: ((msg: string) => Promise<void>) | undefined;
+      vi.spyOn((bot as any).watchManager, 'start').mockImplementation(
+        (_chatId: number, _sessionId: string, send: (msg: string) => Promise<void>) => {
+          capturedSend = send;
+        },
+      );
+      vi.spyOn((bot as any).watchManager, 'watching').mockReturnValue(undefined);
+      vi.spyOn((bot as any).watchManager, 'getWatched').mockReturnValue(undefined);
+
+      try {
+        await (bot as any).runAutoSubscribeTick();
+
+        expect(capturedSend).toBeDefined();
+        await capturedSend!('watch output');
+
+        expect(sendMessageMock).toHaveBeenCalledTimes(1);
+        const [calledChatId, calledText, calledOpts] = sendMessageMock.mock.calls[0] as [number, string, unknown];
+        expect(calledChatId).toBe(12345);
+        expect(calledText).toBe('watch output');
+        // Topic thread → sendOptions must carry message_thread_id = 42.
+        expect(calledOpts).toEqual({ message_thread_id: 42 });
+      } finally {
+        presenceSpy.mockRestore();
+        vi.restoreAllMocks();
+      }
+    });
+
+    test('auto-subscribe falls back to General when no active topic thread exists (#1222 backward-compat)', async () => {
+      // Non-topic chats (getActiveThreadId returns undefined) must be unaffected:
+      // sendOptions receives a bare { chatId } route and returns {} (no thread id).
+      const sendMessageMock = vi.fn(async () => ({ message_id: 1, text: '', date: 0, chat: { id: 12345 } }));
+      (bot as any).bot.telegram.sendMessage = sendMessageMock;
+
+      vi.spyOn((bot as any).sessionManager, 'getActiveThreadId').mockReturnValue(undefined);
+
+      const presence = await import('../agent/awareness/presence.js');
+      const presenceSpy = vi.spyOn(presence, 'readLivePresenceFiles').mockResolvedValue([
+        { surface: 'cli', afk: true, sessionId: 'session-no-topic', cwd: '/tmp', model: 'claude', pid: 4, startedAt: '' },
+      ] as Awaited<ReturnType<typeof presence.readLivePresenceFiles>>);
+
+      let capturedSend: ((msg: string) => Promise<void>) | undefined;
+      vi.spyOn((bot as any).watchManager, 'start').mockImplementation(
+        (_chatId: number, _sessionId: string, send: (msg: string) => Promise<void>) => {
+          capturedSend = send;
+        },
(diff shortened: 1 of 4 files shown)


griffinwork40:
@codex fix comments

griffinwork40 (commented):

---

## PR #1262 Review — fix(telegram): route auto-subscribe watch output to active topic thread (#1222)

**Branch HEAD:** `1637b36552f7a7996d72133bf3d04d5faa4eadbf`
**Files reviewed:** src/telegram/bot.ts, src/telegram/bot.test.ts, src/telegram/session-manager.ts, src/telegram/session-manager.test.ts, src/telegram/route.ts (supporting)
**Change type:** fix · 233 lines · light regime

---

### Prior automated rounds
(none — this is round 1)

---

### Decision

**MERGE (with follow-ups)** — 0 critical, 0 high → no blocking findings; 2 medium advisory (non-blocking under sweep policy), 3 low advisory, 1 nit advisory.

Suggest opening a follow-up issue to track the two medium findings before the next feature build on top of this routing layer.

---

### Findings

---

#### F3 · medium · ADVISORY · correctness
**Stale `autoRoute` closure captures routing at tick time, not send time**

`getActiveThreadId` is called when the watch tick fires and the result is closed over by the `send` callback (`bot.ts:624–628`). If the operator switches topic threads during a long watch session, outgoing messages will continue routing to the thread that was active at tick start rather than the thread that is active at send time.

**Impact:** Silent misrouting during long-running watch sessions. No error is raised; messages land in the wrong thread.

**Recommendation:** Resolve the active thread id inside the `send` closure rather than outside it, so routing reflects the state at the moment of dispatch:
```typescript
const send = async (msg: string): Promise<void> => {
  const tid = this.sessionManager.getActiveThreadId(chatId);
  const route = tid !== undefined ? { chatId, threadId: tid } : { chatId };
  for (const part of splitLongMessage(msg)) {
    await this.bot.telegram.sendMessage(chatId, part, sendOptions(route));
  }
};
```

**Citation:** `src/telegram/bot.ts:624–628` · diff-context · ref:1637b36 (verified Wave 1.5)

---

#### F1 · medium · ADVISORY · api-compat · [low confidence — verify with runtime context]
**`message_thread_id` may be rejected by Telegram for private-chat topics**

For private chats that use Telegram topics, the Bot API may require `direct_messages_topic_id` rather than `message_thread_id` (per Bot API changelog v6.x+). If the operator uses a private-chat group with topics enabled, messages sent via `message_thread_id` will silently route to General or return a `400 Bad Request`.

**Confidence:** LOW — this reflects Telegram API behavior not directly verifiable from code. Validate against the current Bot API changelog and test against a private-chat group with topics before shipping to a broad install base.

**Recommendation:** Verify the current Bot API field name for private-chat topics. If `direct_messages_topic_id` is required, add a routing branch in `route.ts` conditioned on chat type, or add a try-catch in the send path that retries without the thread parameter on `400 message thread not found`.

**Citation:** diff-context · ref:1637b36 (Wave 1.5: plausibility confirmed; runtime validation required)

---

#### F4 · low · ADVISORY · correctness
**`threadId === undefined` guard passes sentinel value `0` as a valid thread id**

`getActiveThreadId` uses a strict `=== undefined` check (`session-manager.ts`). If a route is stored with `threadId = 0`, it will be treated as a legitimate thread id and passed to Telegram. Thread id `0` is not a valid forum topic; this produces an API error or silently routes to General.

**Recommendation:** Widen the guard to `threadId == null || threadId === 0`.

**Citation:** `src/telegram/session-manager.ts` · file-state · ref:1637b36 (Wave 1.5 confirmed)

---

#### F5 · low · ADVISORY · test-coverage
**Missing `toHaveBeenCalledTimes(1)` assertion before destructure in backward-compat test**

The backward-compatibility integration test in `bot.test.ts` destructures mock call arguments without first asserting the mock was called exactly once. If the mock is called zero times the destructure yields `undefined` and the property assertions pass vacuously.

**Recommendation:** Add `expect(sendMessageMock).toHaveBeenCalledTimes(1);` immediately before the destructure, matching the pattern in the topic-routing test.

**Citation:** `src/telegram/bot.test.ts` · diff-context · ref:1637b36 (Wave 1.5 confirmed)

---

#### F7 · low · ADVISORY · test-coverage
**No test exercises `threadId = 0` edge case**

The `threadId = 0` sentinel path identified in F4 has no test coverage; the six unit tests cover `undefined`, single positive threadId, multiple positive threadIds, mixed General+topic, and cross-chat isolation only.

**Recommendation:** Add a unit test that stores a route with `threadId = 0` and asserts the expected General-fallback behavior.

**Citation:** `src/telegram/session-manager.test.ts` · diff-context (absence confirmed Wave 1.5)

---

#### F8 · low · ADVISORY · test-coverage
**No test documents whether routing is live or snapshotted**

No test changes the mock's return value between tick time and send time. The stale-closure behavior in F3 is invisible to the test suite; a future fix will have no regression guard.

**Recommendation:** Add a test that mutates the active-thread mock between tick and send and asserts the expected routing.

**Citation:** `src/telegram/bot.test.ts` · diff-context (absence confirmed Wave 1.5)

---

#### F6 · nit · ADVISORY · test-coverage
**2ms `setTimeout` for sequencing is subject to scheduling jitter**

`lastActivity` is millisecond-resolution (`new Date().toISOString()`), so 2ms is theoretically sufficient, but loaded CI runners can schedule tasks with sub-millisecond gaps, making the ordering test non-deterministic.

**Recommendation:** Replace the `setTimeout` with explicit mock timestamps or fake timers to eliminate the real-time dependency.

**Citation:** `src/telegram/session-manager.test.ts` · diff-context · ref:1637b36

---

### What was not checked
- Citations verified inline against branch HEAD `1637b36552f7a7996d72133bf3d04d5faa4eadbf` (prefetched `bot.ts`, `session-manager.ts`, `route.ts`).
- Tests were not executed; no runtime Telegram API behavior was observed. Finding F1 requires runtime validation against a private-chat group with topics enabled.
- Stated intent assessed from PR #1262 title and body; spec-compliance evaluated against the routing fix described therein.
- Finding F2 (sendOptions wire-format concern) was dropped after Wave 1.5 confirmed `route.ts` correctly maps `threadId → message_thread_id` for non-General routes.

PR #1292, opened 2026-08-25
fix(tools): cap list_directory output at 500 entries

griffinwork40:
## Problem

`list_directory` had no output cap. When a model called it on a directory with
thousands of entries (e.g. `~/.afk/state/sessions/` with 15,936 files), the
handler returned the entire listing verbatim — 618KB of text injected into the
context window as a single tool result.

On a local Qwen model with a 262K-token context window, this single tool result
consumed ~200K tokens and pushed the session to 571K total, producing an
unrecoverable `400 Prompt too long` error. Manual `/compact` could not help
because the tool result sat inside the kept tail (last 2 user turns) that
compaction structurally protects.

Every other discovery tool already has a cap:
- `glob`: 500 results
- `grep` / `bash`: 100KB head+tail via `_output-cap.ts`

`list_directory` was the only unbounded one.

## Fix

Cap `list_directory` output at 500 entries (matching `glob`'s convention).
When a directory exceeds the cap, the first 500 sorted entries are returned
with a marker:

```
[results capped at 500 entries — 15936 total in directory]
```

The full `readdir` still runs — only the returned string is truncated. The
cap is exported as `MAX_ENTRIES` for test access.

## Tests

Two new test cases:
- **Over cap**: 550 files → returns 501 lines (500 entries + marker), marker
  includes exact total count.
- **At cap boundary**: exactly 500 files → returns all 500, no marker.

All 20 existing + 2 new tests pass. `pnpm lint` and `pnpm audit:filesize:check`
clean.

### DIFF
--- a/src/agent/tools/handlers/list-directory.test.ts
+++ b/src/agent/tools/handlers/list-directory.test.ts
@@ -2,7 +2,7 @@ import { describe, expect, it, beforeEach, afterEach } from 'vitest';
 import { promises as fs } from 'fs';
 import { tmpdir } from 'os';
 import { join } from 'path';
-import { listDirectoryHandler } from './list-directory.js';
+import { listDirectoryHandler, MAX_ENTRIES } from './list-directory.js';
 import type { ToolHandlerContext } from '../types.js';
 
 describe('listDirectoryHandler', () => {
@@ -207,6 +207,41 @@ describe('listDirectoryHandler', () => {
     const lines = result.content.split('\n').map((l) => l.trim()).filter(Boolean);
     expect(lines).not.toContain('.');
     expect(lines).not.toContain('..');
+  });
+
+  it('caps output at MAX_ENTRIES and appends a marker', async () => {
+    const total = MAX_ENTRIES + 50;
+    const promises: Promise<void>[] = [];
+    for (let i = 0; i < total; i++) {
+      promises.push(fs.writeFile(join(tmpDir, `file-${String(i).padStart(5, '0')}.txt`), ''));
+    }
+    await Promise.all(promises);
+
+    const result = await listDirectoryHandler({ path: tmpDir }, new AbortController().signal);
+
+    expect(result.isError).not.toBe(true);
+    const lines = result.content.split('\n');
+    // MAX_ENTRIES data lines + 1 cap marker line
+    expect(lines).toHaveLength(MAX_ENTRIES + 1);
+    expect(lines[lines.length - 1]).toBe(
+      `[results capped at ${MAX_ENTRIES} entries — ${total} total in directory]`,
+    );
+  });
+
+  it('returns all entries when count equals MAX_ENTRIES', async () => {
+    const promises: Promise<void>[] = [];
+    for (let i = 0; i < MAX_ENTRIES; i++) {
+      promises.push(fs.writeFile(join(tmpDir, `f-${String(i).padStart(5, '0')}.txt`), ''));
+    }
+    await Promise.all(promises);
+
+    const result = await listDirectoryHandler({ path: tmpDir }, new AbortController().signal);
+
+    expect(result.isError).not.toBe(true);
+    const lines = result.content.split('\n');
+    expect(lines).toHaveLength(MAX_ENTRIES);
+    // No cap marker
+    expect(result.content).not.toContain('[results capped');
   });
 });
 
--- a/src/agent/tools/handlers/list-directory.ts
+++ b/src/agent/tools/handlers/list-directory.ts
@@ -5,6 +5,11 @@
  * Directories are suffixed with `/`, files are plain names. Entries are sorted
  * alphabetically with directories first.
  *
+ * Invariant: output is capped at {@link MAX_ENTRIES} entries (matching the glob
+ * tool's 500-entry convention). Directories with thousands of entries — like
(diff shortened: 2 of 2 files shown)


griffinwork40 (commented):

## Review: PR #1292 — `fix(tools): cap list_directory output at 500 entries`

**Target:** `fix/list-directory-output-cap` → `main` · ref `8b4a38b9` · +48/−2 · 2 files
**Regime:** light (hotfix, ~50 lines)
**Stated intent:** PR #1292 title+body — spec-compliance assessed.

---

### Findings

**1 · `nit` · blocking:false · confidence:high · correctness · `src/agent/tools/handlers/list-directory.ts:97` vs `src/agent/tools/handlers/glob.ts:262` · ref:8b4a38b9 · file-state (verified)**

**Cap marker string diverges from glob's peer format.** The glob tool emits `[results capped at 500 entries]` (no total count). This PR emits `[results capped at 500 entries — N total in directory]` (includes total). The PR body explicitly invokes the glob convention but does not converge the string format. The richer marker is an improvement — but downstream pattern-matching on the cap marker string (e.g. telemetry, tests using `toContain('[results capped')`) now hits two distinct forms.

```typescript
// glob.ts:262
output += '\n[results capped at 500 entries]';
// list-directory.ts:97
content += `\n[results capped at ${MAX_ENTRIES} entries — ${total} total in directory]`;
```

**suggestion:** Either update `glob.ts` to also include the total count for consistency, or accept the divergence consciously. Both markers start with `[results capped at 500 entries` so prefix-matching still works.

---

**2 · `nit` · blocking:false · confidence:high · api-compat · `src/agent/tools/handlers/list-directory.ts:27` · ref:8b4a38b9 · file-state (verified)**

**`MAX_ENTRIES` exported without `@internal` tag.** The constant is exported solely for test access (one importer: `list-directory.test.ts`). Absent an `@internal` JSDoc annotation, it reads as stable public API — a future change to the cap value would surface as an unexpected test signal for any downstream consumer.

**suggestion:** Add `@internal` to the JSDoc comment on `MAX_ENTRIES`.

---

**3 · `low` · blocking:false · confidence:high · test-coverage · `src/agent/tools/handlers/list-directory.test.ts:213–245` · ref:8b4a38b9 · file-state (verified)**

**Cap tests use only files — no mixed dirs+files at the boundary.** Both new tests create only flat files. The implementation sorts directories first, then files. At cap boundary (e.g. 400 dirs + 200 files = 600 total), the 500-entry slice would include all 400 dirs and 100 files. This is correct behavior but untested. No production bug — the slice operates on the already-sorted array regardless of entry type.

**suggestion:** Optional follow-up: add a test with dirs+files totaling >500 and assert the slice respects the dirs-first sort order.

---

### Spec-compliance

Stated intent: PR #1292 title+body — spec-compliance assessed.

| Requirement | Status |
|---|---|
| Cap `list_directory` at 500 entries | ✅ `sorted.slice(0, MAX_ENTRIES)` at line 94 |
| Marker: `[results capped at 500 entries — N total in directory]` | ✅ Line 97 |
| Match glob's 500-entry convention | ✅ `MAX_ENTRIES = 500` matches `glob.ts:110`'s `maxResults = 500` |
| Export `MAX_ENTRIES` for test access | ✅ Exported at line 27 |
| Full readdir still runs (only returned string capped) | ✅ Documented in JSDoc and implemented |
| Two boundary tests (over cap + at cap) | ✅ Both present and correct |

No unmet intent. No scope creep detected.

### Dimensions with no issues

- **security** — no issues found. The patch is downstream of `resolveAndContain` (path containment). No new file access, permission changes, or trust boundaries. `total` is derived from `entries.length`, not user input. Read: `list-directory.ts:47–78`.
- **correctness (boundary)** — no issues found. The `>` (strict greater-than) comparison at line 96 correctly handles the exact-boundary case: `total === MAX_ENTRIES` returns all entries with no marker. The `sorted.slice(0, MAX_ENTRIES)` call is correct for both over and at-cap cases. Read: `list-directory.ts:93–98`.
- **perf-observability** — no issues found. The fix eliminates the acute problem (618KB context injection). The full `readdir` call is a conscious trade-off documented in JSDoc. No new observability instrumentation is added, which is acceptable for a hotfix — truncation is visible to the model via the marker string. Read: `list-directory.ts:62, 93–98`.

### What was not checked

- Citations verified inline against branch HEAD `8b4a38b97c67b07a742c879bcc477dad00b4c47e`.
- Stated intent: PR #1292 title+body — spec-compliance assessed.
- Not checked: runtime test execution — static review only. CI shows all checks pass.
- Not checked: whether the `list_directory` tool description shown to models mentions the 500-entry cap (the tool schema in `tools/index.ts` or the system prompt description). If it does not, models cannot predict truncation.
- Not checked: interaction with `createListDirectoryHandler(cwd)` — the `cwd` parameter flows through `resolveAndContain` only, not the cap logic, so no interaction is expected.

---

**Decision: MERGE — 0 blocking (2 nit, 1 low).**

Clean, well-scoped hotfix that closes an acute context-window overflow on large directories. The implementation is correct at the boundary, matches the 500-entry convention from `glob`, and has solid test coverage for both the over-cap and at-cap cases. The three findings (marker format divergence, missing `@internal` tag, mixed-type test gap) are follow-up opportunities, not regressions.

PR #1300, opened 2026-08-26
fix(cli): guard code-block register with enabled flag (#1289)

### LINKED ISSUE #1289: code-block register accumulates entries on non-REPL surfaces without reset
## Problem

`registerCodeBlock()` in `src/cli/formatter.block.ts` is called unconditionally inside `renderCodeBlock()` — every surface that renders markdown through `src/cli/formatter.ts` populates the module-scope `blocks` array in `src/cli/code-block-register.ts`.

Only the REPL loop (`src/cli/commands/interactive/loop-iteration.ts`) calls `resetCodeBlockRegister()`. Non-REPL render paths that call `renderMarkdownToTerminal` accumulate entries without bound:

- `afk chat` one-shot (`src/cli/commands/chat.ts:754`)
- `/transcript` re-rendering (`src/cli/slash/commands/transcript.ts:93`)
- Streaming markdown renderer (`src/cli/markdown-stream.ts` → `formatter.ts`)

## Impact

Resource leak, not a correctness issue — `/copy` is gated to the REPL surface so stale entries are never consumed. Each entry holds a string reference to text already in memory (`stats.turns`), so the per-entry overhead is small. But in a long-lived daemon process that renders many one-shot responses, the array grows indefinitely.

## Suggested fix

Guard `registerCodeBlock()` in `renderCodeBlock()` so it only fires when the register is active:

- Add an `enabled` flag to the register module, set/cleared by the REPL loop alongside the reset
- Or pass a render-context object through the formatter call chain (larger refactor)
- Or accept the leak as bounded and document the contract

## Context

Found during review of #1285 (`feat(cli): add /copy slash command with render-time code block capture`).

griffinwork40:
## Summary

Guards the code-block register with an `enabled` flag so it only accumulates entries on REPL surfaces where `/copy` can consume them.

### Problem

`registerCodeBlock()` was called unconditionally inside `renderCodeBlock()` — every surface that renders markdown (one-shot `afk chat`, `/transcript` re-rendering, streaming renderer) populated the module-scope `blocks` array in `code-block-register.ts`. Only the REPL loop called `resetCodeBlockRegister()`. Non-REPL paths accumulated entries without bound — a resource leak in long-lived daemon processes.

### Fix

- Added `enabled` flag (default `false`) to `code-block-register.ts`
- `registerCodeBlock()` is a no-op when disabled — the blocks array never grows on non-REPL surfaces
- Exported `enableCodeBlockRegister()` / `disableCodeBlockRegister()` 
- REPL loop enables the register at the turn boundary alongside the existing reset

### Tests

- 5 new tests covering the enabled flag: no-op when disabled, works when enabled, disable re-gates, reset is flag-independent, enable is idempotent
- All 10 tests pass (5 existing + 5 new)

Closes #1289.

### DIFF
--- a/src/cli/code-block-register.test.ts
+++ b/src/cli/code-block-register.test.ts
@@ -4,19 +4,25 @@ import {
   getCodeBlocks,
   getCodeBlock,
   resetCodeBlockRegister,
+  enableCodeBlockRegister,
+  disableCodeBlockRegister,
 } from './code-block-register.js';
 
 beforeEach(() => {
+  // Always start each test with a clean, disabled register.
+  disableCodeBlockRegister();
   resetCodeBlockRegister();
 });
 
 describe('code-block-register', () => {
   it('starts empty', () => {
+    enableCodeBlockRegister();
     expect(getCodeBlocks()).toHaveLength(0);
     expect(getCodeBlock(1)).toBeUndefined();
   });
 
   it('registers blocks with 1-based indices', () => {
+    enableCodeBlockRegister();
     const i1 = registerCodeBlock('bash', 'echo hello');
     const i2 = registerCodeBlock('python', 'print("hi")');
     expect(i1).toBe(1);
@@ -25,6 +31,7 @@ describe('code-block-register', () => {
   });
 
   it('retrieves blocks by 1-based index', () => {
+    enableCodeBlockRegister();
     registerCodeBlock('bash', 'echo hello');
     registerCodeBlock('python', 'print("hi")');
     const block = getCodeBlock(2);
@@ -32,12 +39,14 @@ describe('code-block-register', () => {
   });
 
   it('returns undefined for out-of-range index', () => {
+    enableCodeBlockRegister();
     registerCodeBlock('bash', 'echo hello');
     expect(getCodeBlock(0)).toBeUndefined();
     expect(getCodeBlock(2)).toBeUndefined();
   });
 
   it('resets the register', () => {
+    enableCodeBlockRegister();
     registerCodeBlock('bash', 'echo hello');
     registerCodeBlock('python', 'print("hi")');
     resetCodeBlockRegister();
@@ -45,5 +54,49 @@ describe('code-block-register', () => {
     // New registrations start at 1 again.
     const i = registerCodeBlock('text', 'foo');
     expect(i).toBe(1);
+  });
+
+  // --- enabled flag tests (issue #1289) ---
+
+  it('registerCodeBlock is a no-op when disabled (default)', () => {
+    // Register is disabled in beforeEach — no enable() call here.
+    const idx = registerCodeBlock('bash', 'echo hello');
+    expect(idx).toBe(0);
+    expect(getCodeBlocks()).toHaveLength(0);
+    expect(getCodeBlock(1)).toBeUndefined();
+  });
+
+  it('registerCodeBlock works when enabled', () => {
+    enableCodeBlockRegister();
+    const idx = registerCodeBlock('bash', 'echo hello');
+    expect(idx).toBe(1);
+    expect(getCodeBlocks()).toHaveLength(1);
+    expect(getCodeBlock(1)).toEqual({ index: 1, lang: 'bash', text: 'echo hello' });
+  });
+
+  it('disableCodeBlockRegister makes registerCodeBlock a no-op again', () => {
+    enableCodeBlockRegister();
+    registerCodeBlock('bash', 'echo first');
+    disableCodeBlockRegister();
+    const idx = registerCodeBlock('python', 'print("second")');
+    // Should not have been registered.
+    expect(idx).toBe(0);
+    expect(getCodeBlocks()).toHaveLength(1);
+    expect(getCodeBlock(2)).toBeUndefined();
+  });
+
+  it('resetCodeBlockRegister clears blocks regardless of enabled state', () => {
+    enableCodeBlockRegister();
+    registerCodeBlock('bash', 'echo hello');
+    expect(getCodeBlocks()).toHaveLength(1);
+    resetCodeBlockRegister();
+    expect(getCodeBlocks()).toHaveLength(0);
+  });
+
+  it('enable is idempotent — calling it twice does not double-register', () => {
+    enableCodeBlockRegister();
+    enableCodeBlockRegister();
+    registerCodeBlock('bash', 'echo hello');
+    expect(getCodeBlocks()).toHaveLength(1);
   });
 });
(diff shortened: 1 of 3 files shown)


griffinwork40 (commented):

## Review: PR #1300 — `fix(cli): guard code-block register with enabled flag (#1289)`

**Target:** `afk/iso-issue-1289-1-yb9x82` → `main` · ref `a1a17e4c` · +63/−3 · 3 files
**Regime:** light (hotfix)
**Stated intent:** PR #1300 title+body — spec-compliance assessed.

---

### Findings

**1 · `medium` · blocking:false · confidence:high · correctness · `src/cli/formatter.block.ts:101,109` · ref:a1a17e4c · file-state (verified)**

**`renderCodeBlock` emits `/cp 0` hint on every non-REPL surface.** `formatter.block.ts:101` captures `registerCodeBlock()`'s return value as `blockIndex` and line 109 unconditionally splices it into the code block header: `` palette.dim(` ── /cp ${blockIndex}`) ``. When the register is disabled (all non-REPL surfaces after this PR), `registerCodeBlock` returns 0, and every rendered code block header emits a dim `── /cp 0` hint. Running `/copy 0` produces "Block 0 not found" because `getCodeBlock(0)` evaluates `blocks[-1]` → `undefined`. The hint is cosmetic dim text on surfaces where `/copy` is unavailable anyway, so the blast radius is bounded.

```typescript
// formatter.block.ts:101 — returns 0 when disabled
const blockIndex = registerCodeBlock(lang, code.text);
// formatter.block.ts:109 — unconditionally emits the hint
const copyHint = palette.dim(` ── /cp ${blockIndex}`);
```

**suggestion:** Guard the hint on a positive index: `const copyHint = blockIndex > 0 ? palette.dim(` ── /cp ${blockIndex}`) : '';`

*Waived: bounded — the `/cp 0` hint is dim cosmetic text on non-REPL surfaces where `/copy` is not available; no data loss, no state corruption, no user action path is broken on the REPL surface where `/copy` works.*

---

**2 · `low` · blocking:false · confidence:high · api-compat · `src/cli/code-block-register.ts:40–43` · ref:a1a17e4c · file-state (verified, grep-confirmed: zero production importers)**

**`disableCodeBlockRegister()` is exported with no production caller.** `git grep` across all `.ts` files (excluding tests) at the reviewed ref returns zero importers of `disableCodeBlockRegister`. The function exists only for test harness use (`beforeEach` in `code-block-register.test.ts`). The export is harmless and pre-wires a future embed scenario, but the public module surface carries a function with no production consumer.

**suggestion:** Add `@internal` JSDoc tag, or leave as-is — no merge-blocking concern.

---

**3 · `low` · blocking:false · confidence:high · test-coverage · `src/cli/code-block-register.test.ts` · ref:a1a17e4c · diff-context (grep-confirmed: zero matches for `renderCodeBlock` in any test file)**

**No test exercises the rendering-layer integration of `blockIndex === 0`.** The five new tests correctly verify `code-block-register.ts` in isolation. None exercise `formatter.block.ts:renderCodeBlock` to confirm the `/cp 0` hint behavior from finding F-1. The gap is at the integration boundary between the register and the renderer.

**suggestion:** Add a test calling `renderCodeBlock` with the register disabled and assert the header omits (or contains) the `/cp N` hint per the intended behavior.

---

**4 · `nit` · blocking:false · confidence:high · correctness · `src/cli/code-block-register.test.ts:55` · ref:a1a17e4c · diff-context**

**Test name says "default" but `beforeEach` imposes the disabled state explicitly.** The test is titled `'registerCodeBlock is a no-op when disabled (default)'` — `beforeEach` calls `disableCodeBlockRegister()` before the test runs, so it verifies the post-`disable()` state, not the module-scope `let enabled = false` default. The default happens to be `false`, so the test is correct; the framing is imprecise.

**suggestion:** Rename to `'registerCodeBlock is a no-op when disabled'`.

---

### Spec-compliance

Stated intent: PR #1300 title+body — spec-compliance assessed.

| Requirement | Status |
|---|---|
| `enabled` flag added (default `false`) | ✅ `code-block-register.ts:32` |
| `registerCodeBlock()` is a no-op when disabled (returns 0) | ✅ `code-block-register.ts:55` |
| `enableCodeBlockRegister()` / `disableCodeBlockRegister()` exported | ✅ Lines 36–47 |
| REPL loop enables register at turn boundary | ✅ `loop-iteration.ts:680` |
| Non-REPL surfaces never accumulate blocks | ✅ No non-REPL caller enables the register |
| 5 new tests; 10 total | ✅ 5 new `it()` blocks added |
| Closes #1289 (resource leak on non-REPL surfaces) | ✅ Correctly fixed |

No unmet intent. No scope creep detected. The `/cp 0` rendering artefact (F-1) is a consequence not addressed by the stated intent — the PR correctly fixes the memory leak it describes.

### Dimensions with no issues

- **Security** — no issues found. The change is confined to a module-scoped boolean guard and an array-growth no-op. No user input flows into `enabled`; no auth, secrets, or injection surfaces involved. Read: `code-block-register.ts` full file, `loop-iteration.ts` call site.
- **Perf-observability** — no issues found. The `if (!enabled) return 0` check is a single boolean read, O(1), zero allocation on the disabled path. The memory leak (unbounded `blocks` array growth on non-REPL surfaces) is correctly resolved. Read: `code-block-register.ts:55`.

### What was not checked

- Citations verified inline against branch HEAD `a1a17e4cb85045c476bd847cf9aa4f598f7afb45`.
- Stated intent: PR #1300 title+body — spec-compliance assessed.
- Not checked: runtime test execution — static review only. CI shows all checks passing (Lint & Build ✅, Tests ✅ ubuntu + macos).
- Not checked: whether `stream-renderer-orchestrator-emit.ts` has a conditional render path that suppresses `renderCodeBlock` entirely on non-REPL surfaces (which would make F-1 unreachable on that path).
- Not checked: the `/copy` command handler's behavior when invoked with index 0 (inferred from `getCodeBlock(0) → blocks[-1] → undefined`).

---

**Severity arithmetic: 0 critical, 0 high, 0 medium blocking → MERGE; 1 medium waived (correctness — bounded cosmetic `/cp 0` hint on non-REPL surfaces, no data loss), 2 low, 1 nit.**

Clean, well-scoped fix that correctly resolves the unbounded memory leak described in #1289. The register's `enabled` flag is correctly defaulted to `false`, the REPL loop correctly enables it at the turn boundary, and all five stated invariants are met. The waived medium (F-1: `/cp 0` hint on non-REPL surfaces) is a cosmetic follow-up, not a regression — the hint was equally meaningless before this PR (the blocks accumulated but `/copy` was unavailable on those surfaces).

PR #1351, opened 2026-08-28
fix(subagent-log): document one-line overshoot contract and tighten cap test (#1316)

### LINKED ISSUE #1316: bug(subagent-log): write() cap gate allows one-line overshoot past MAX_LOG_BYTES
## Summary

`SubagentLogWriter.write()` checks `bytesWritten >= MAX_LOG_BYTES` before serializing the event, then pushes the serialized line to `pendingLines` without a per-line byte check. At flush time, `writeLine()` applies its own cap check — so the overshoot is bounded to one line (up to ~2KB for a typical event).

The test (`log.test.ts`) asserts `stat.size <= 1_048_576 + 2100`, with the 2100-byte tolerance exactly matching one large line's overhead. The documented contract says "1 MB cap" but the actual behavior is "1 MB + one line."

## Location

- `src/agent/subagent/log.ts:69–73` — `write()` pushes to `pendingLines` without byte check
- `src/agent/subagent/log.test.ts:190` — tolerance hides one-line overshoot

## Suggested fix

Either:
1. Compute `Buffer.byteLength(line)` before pushing to `pendingLines` and apply the same cap check there
2. Document the one-line overshoot contract explicitly in the code comment and tighten the test to assert the exact bound rather than a soft tolerance

## Origin

Post-merge review of PR #1312.

griffinwork40:
Fixes #1316.

Documents the one-line overshoot contract in the `write()` method (the cap guard fires before serialization, so the last admitted line can push the file past `MAX_LOG_BYTES` by one event) and tightens the test to compute the exact bound from the serialized event size rather than using a magic `+ 2100` tolerance.

### Changes
- `src/agent/subagent/log.ts` -- `Invariant:` comment on the overshoot contract
- `src/agent/subagent/log.test.ts` -- computed `oneLineBytes` replaces hardcoded tolerance

### Verification
- 10/10 tests pass
- `pnpm lint` clean

### DIFF
--- a/src/agent/subagent/log.test.ts
+++ b/src/agent/subagent/log.test.ts
@@ -174,19 +174,26 @@ describe('SubagentLogWriter.close flushes pending lines', () => {
 // ---------------------------------------------------------------------------
 
 describe('MAX_LOG_BYTES cap', () => {
-  it('file does not exceed 1 MiB even when many large events are written', async () => {
+  it('file does not exceed MAX_LOG_BYTES + one line (one-line overshoot contract)', async () => {
     const subId = 'sub-overflow';
     const w = new SubagentLogWriter(SESSION, subId);
 
     // Write ~2MB worth of events (each ~2KB) to exceed the 1MB cap
     const payload = 'x'.repeat(2000);
+    // Compute the exact serialized byte length of one event line so the
+    // assertion reflects the documented one-line-overshoot invariant precisely.
+    const oneLineBytes = Buffer.byteLength(
+      JSON.stringify(makeChunkEvent(payload)) + '\n',
+      'utf8',
+    );
     for (let i = 0; i < 1200; i++) {
       w.write(makeChunkEvent(payload));
     }
     await w.close();
 
     const stat = fs.statSync(w.logPath);
-    // File must be ≤ 1 MiB (with some tolerance for the final incomplete line)
-    expect(stat.size).toBeLessThanOrEqual(1_048_576 + 2100);
+    // The cap guard fires before serialization, so the last admitted line may
+    // push the file to MAX_LOG_BYTES + oneLineBytes. Assert the exact bound.
+    expect(stat.size).toBeLessThanOrEqual(1_048_576 + oneLineBytes);
   });
 });
--- a/src/agent/subagent/log.ts
+++ b/src/agent/subagent/log.ts
@@ -65,6 +65,13 @@ export class SubagentLogWriter {
   /**
    * Append an OutputEvent as a JSONL line.
    * Silently no-ops on error, after close, or when the file exceeds MAX_LOG_BYTES.
+   *
+   * Invariant: the cap allows a one-line overshoot. The guard fires before
+   * serialization, so the last admitted event may push the file up to
+   * MAX_LOG_BYTES + <size of that serialized line>. writeLine() applies the
+   * same guard, so the overshoot is bounded by the byte length of a single
+   * serialized event (in practice, ≤ a few KB). This is intentional — a
+   * strict hard cap would require a costly split across the line boundary.
    */
   write(event: OutputEvent): void {
     if (this.errored || this.closed || this.bytesWritten >= MAX_LOG_BYTES) return;


griffinwork40 (commented):

---

## Prior rounds
This is the **first automated round** on this PR — no prior findings to carry forward.

---

**Severity arithmetic:** 0 critical, 0 high → MERGE (with follow-ups); 0 medium; 2 low + 1 nit advisory (all non-blocking).

CI is green (all 9 checks pass). The stated intent — document the one-line overshoot contract and replace the `+ 2100` magic tolerance with a computed exact bound — is fully met with no regressions found.

---

## Findings (all non-blocking, advisory only)

### Low — Comment accuracy: invariant description slightly misleading
**`src/agent/subagent/log.ts:65–77`** · ref: aa5a5cb · file-state · dimension: correctness

The `Invariant:` comment says _"writeLine() applies the same guard, so the overshoot is bounded by the byte length of a single serialized event"_ — this is accurate on the key claim (overshoot bounded to one line), but the framing is slightly off. The `write()` guard at line 77 fires _before_ `JSON.stringify` (a fast-path to skip serialization), while `writeLine()`'s guard is the actual final enforcement point. Attributing the bounding to both guards equally undersells the distinction.

```ts
// write(): fast-path, fires BEFORE serialization
if (this.errored || this.closed || this.bytesWritten >= MAX_LOG_BYTES) return;
const line = JSON.stringify(event) + '\n';   // ← serialization happens after guard
// ...
this.writeLine(line);   // writeLine's guard is the actual enforcement
```

**Suggestion:** Revise the comment to clarify that `write()`'s guard is a serialization-skipping fast-path, while `writeLine()`'s guard is the enforcement that actually bounds the overshoot. The key claim is correct; just the roles of the two guards need clearer framing.

**Non-blocking** — comment-only, no behavioral impact. Suggest a follow-up edit alongside the next change to this file.

---

### Nit — Test subId is fixed string under append mode
**`src/agent/subagent/log.test.ts:178`** · ref: aa5a5cb · diff-context · dimension: correctness

`subId = 'sub-overflow'` is a fixed string and the stream opens with `flags: 'a'` (append). In practice this is a non-issue because vitest creates a fresh `mkdtempSync` temp dir per test process run, so cross-run accumulation cannot occur. Noted only because a reader unfamiliar with the outer test setup might question test isolation.

**Non-blocking** — no real defect; process isolation handles it.

---

### Nit — Test comment assumes uniform payload
**`src/agent/subagent/log.test.ts:177–199`** · ref: aa5a5cb · diff-context · dimension: correctness

`oneLineBytes` is computed once from a single `makeChunkEvent(payload)` call, which is valid because all 1200 loop events share the same payload. The comment doesn't say this, so a future reader who changes the loop to write variable-size events might not realize the bound computation needs updating.

**Suggestion:** Append `// valid because all events share the same payload` to the `oneLineBytes` comment.

**Non-blocking** — naming/comment only.

---

## What was not checked
- Citations verified inline against branch HEAD `aa5a5cbae0e3c8a779eedf58526cfe244fc9cb1c`.
- Stated intent: PR #1351 title + body — spec-compliance assessed. Intent fully met.
- Did not audit other files that import `SubagentLogWriter` for contract compatibility (no public API changed; not needed).
- Did not run the test suite locally (CI green on GitHub Actions is the ground truth here).

griffinwork40 (commented):

---

**Review of** `aa5a5cba` | **Change type:** hotfix | **Regime:** light | **Files:** 2 | **Lines:** ~27

**Decision: MERGE — 0 blocking (2 low).**

Both stated goals are fully met: the one-line overshoot invariant is clearly documented in the `write()` JSDoc, and the test computes the exact bound via `Buffer.byteLength` instead of the previous magic `+ 2100` tolerance. No behavioral changes, no new exports, no scope creep.

---

### Findings

**1. `low` · blocking: false · test-coverage · `src/agent/subagent/log.test.ts:31–36` · ref: `aa5a5cba` · file-state · confidence: low**

`makeChunkEvent` casts via `as unknown as OutputEvent extends { type: 'chunk'; chunk: infer C } ? C : never`. The conditional type resolves against the full `OutputEvent` union (not a naked type parameter), collapsing to `never` — the cast compiles only because `as unknown` breaks type-safety first. The runtime object is structurally valid, so test behavior is correct today, but the compiler will not catch a stale helper if `OutputEvent` changes shape. Pre-existing issue, not introduced by this PR.

```typescript
chunk: { type: 'content', content: text }
  as unknown as OutputEvent extends { type: 'chunk'; chunk: infer C } ? C : never,
```

**Suggestion:** Replace with a direct `as ContentChunk` cast and import `ContentChunk` from the types module.

---

**2. `low` · blocking: false · test-coverage · `src/agent/subagent/log.test.ts:177–199` · ref: `aa5a5cba` · diff-context · confidence: medium**

The overshoot bound is verified only for the `ContentChunk` event shape (~2 KB serialized). The invariant comment claims "in practice, ≤ a few KB" but no test exercises worst-case event shapes (e.g. a `ToolUseDetailChunk` with large metadata). Acceptable for a hotfix scope — the bound logic is shape-independent (it measures bytes of whatever `JSON.stringify` emits), so the risk is purely documentary.

```typescript
const oneLineBytes = Buffer.byteLength(
  JSON.stringify(makeChunkEvent(payload)) + '\n',
  'utf8',
);
```

**Suggestion:** A follow-up test asserting the largest production event shape stays under the "few KB" claim would fully close the invariant.

---

### Dropped findings

None.

### What Was Not Checked

- Citations verified inline against branch HEAD `aa5a5cbae0e3c8a779eedf58526cfe244fc9cb1c`.
- Stated intent: PR #1351 title + body — spec-compliance assessed; both requirements fully met, no scope creep.
- Did not run the test suite locally (CI shows all GitHub Actions checks green).
- Did not verify whether other `OutputEvent` variants (e.g. `ToolResultEvent`) produce serialized sizes exceeding the "few KB" comment claim.
- Security: no surface — pure JSDoc addition and test tightening on a file-local class. No auth, no secrets, no network.
- API-compat: `SubagentLogWriter` public surface unchanged; no exported types altered.

PR #1353, opened 2026-08-28
perf(subagent-log): memoize mkdirSync per sessionLabel to avoid redundant syscalls (#1320)

### LINKED ISSUE #1320: perf(subagent-log): synchronous mkdirSync on every fork hot path
## Summary

`SubagentLogWriter`'s constructor calls `fs.mkdirSync({ recursive: true })` synchronously on the critical fork path of every subagent. With `AFK_SUBAGENT_LOG` on by default, every fork now performs a synchronous FS call before returning a handle.

All subagents in a session share the same directory (`sessionLabel`), so N−1 of these calls are no-op `stat` syscalls after the first. In a compose DAG with 20 parallel nodes, 20 synchronous `mkdirSync` calls serialize on the event loop before any handle is returned. No trace event or metric captures this overhead.

## Location

- `src/agent/subagent/log.ts:52–55` — `fs.mkdirSync()` in constructor

## Suggested fix

Memoize the directory creation per `sessionLabel` at the module level (a `Set<string>` of already-created dirs), or move the `mkdir` into `openStream()` using `fs.promises.mkdir` to keep it off the synchronous fork path.

## Origin

Post-merge review of PR #1312.

griffinwork40:
Fixes #1320.

All subagents in a session share the same log directory, so N-1 `mkdirSync` calls after the first are redundant `stat` syscalls on the fork hot path. Added a module-level `Set<string>` to memoize created directories.

### Changes
- `src/agent/subagent/log.ts` -- `createdDirs` set gates `mkdirSync` to once-per-directory
- `src/agent/subagent/log.test.ts` -- 2 tests verifying correct behavior with memoization (2 writers + 8-writer stress test)

### Verification
- 12/12 tests pass
- `pnpm lint` clean

### DIFF
--- a/src/agent/subagent/log.test.ts
+++ b/src/agent/subagent/log.test.ts
@@ -209,6 +209,54 @@ describe('stream-open-window concurrent writes', () => {
 });
 
 // ---------------------------------------------------------------------------
+// mkdirSync memoization — multiple writers with the same sessionLabel
+// ---------------------------------------------------------------------------
+
+describe('mkdirSync memoization', () => {
+  it('two writers sharing a sessionLabel both produce correct logs', async () => {
+    const session = 'test-memo-session';
+
+    const w1 = new SubagentLogWriter(session, 'sub-memo-1');
+    const w2 = new SubagentLogWriter(session, 'sub-memo-2');
+
+    w1.write(makeChunkEvent('from-w1'));
+    w2.write(makeChunkEvent('from-w2'));
+    await Promise.all([w1.close(), w2.close()]);
+
+    // Both log files must be readable — verifies the memoised path does not
+    // suppress directory creation for the first writer or break the second.
+    const events1: OutputEvent[] = [];
+    for await (const e of SubagentLogReader.readEvents(session, 'sub-memo-1')) {
+      events1.push(e);
+    }
+    const events2: OutputEvent[] = [];
+    for await (const e of SubagentLogReader.readEvents(session, 'sub-memo-2')) {
+      events2.push(e);
+    }
+    expect(events1).toHaveLength(1);
+    expect(events2).toHaveLength(1);
+  });
+
+  it('N writers sharing a sessionLabel all produce correct logs (stress)', async () => {
+    const session = 'test-memo-stress';
+    const N = 8;
+    const writers = Array.from({ length: N }, (_, i) =>
+      new SubagentLogWriter(session, `sub-stress-${i}`),
+    );
+    for (const [i, w] of writers.entries()) w.write(makeChunkEvent(`msg-${i}`));
+    await Promise.all(writers.map(w => w.close()));
+
+    for (let i = 0; i < N; i++) {
+      const events: OutputEvent[] = [];
+      for await (const e of SubagentLogReader.readEvents(session, `sub-stress-${i}`)) {
+        events.push(e);
+      }
+      expect(events).toHaveLength(1);
+    }
+  });
+});
+
+// ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

---

**Review of** `c8e9c068` | **Change type:** hotfix (perf) | **Regime:** light | **Files:** 2 | **Lines:** ~67

The optimization is sound: `mkdirSync({ recursive: true })` is idempotent, the `Set` guard is correct, and the error path for subsequent writers degrades gracefully to the stream-error handler. Tests verify output correctness across shared sessions.

---

### Findings

**1. `low` · blocking: false · test-coverage · `src/agent/subagent/log.test.ts:172–222` · ref: `c8e9c068` · diff-context · confidence: high**

The new `describe('mkdirSync memoization', ...)` block verifies output correctness (events round-trip) but does not assert the central claim of the PR — that `mkdirSync` is called once per directory instead of N times. A regression that broke the guard but preserved output correctness would pass all tests.

**Evidence:** The two test cases call `readEvents` and assert `toHaveLength(1)` on events — they test *that* outputs are correct, not *that* the syscall was deduplicated. `grep` for `vi.spyOn.*mkdirSync`, `callCount`, or `toHaveBeenCalledTimes` in the test file → zero matches.

**Suggestion:** Add `vi.spyOn(fs, 'mkdirSync')` and assert `spy.toHaveBeenCalledTimes(1)` after constructing N writers with the same `sessionLabel`.

---

**2. `low` · blocking: false · correctness · `src/agent/subagent/log.ts:62–73` · ref: `c8e9c068` · diff-context · confidence: medium**

The error-handling contract changes asymmetrically. The first writer per directory fails fast in the constructor (`catch { this.errored = true }`). Subsequent writers skip the guard entirely — if the directory is deleted between the first writer's successful `mkdir` and a later writer's constructor, the later writer discovers the error lazily via the stream-open `'error'` handler (`log.ts:114–119`). The eventual outcome is the same (`this.errored = true`, `write()` no-ops), but failure latency differs.

```typescript
if (!createdDirs.has(dir)) {
  try { fs.mkdirSync(dir, { recursive: true }); createdDirs.add(dir); }
  catch { this.errored = true; }  // only fires for first writer per dir
}
// second+ writers: no errored=true path even if dir was deleted
```

**Suggestion:** A brief comment documenting this asymmetry is sufficient — the lazy fallback via stream-error is correct.

---

**3. `nit` · blocking: false · perf-observability · `src/agent/subagent/log.ts:33` · ref: `c8e9c068` · diff-context · confidence: high**

The `createdDirs` Set has no eviction. Across a long-lived daemon process the Set grows one entry per unique `sessionLabel`. Growth is negligible (~30–50 char strings × sessions per process), but the comment ("this process") doesn't note the unbounded lifetime.

**Suggestion:** Append to the JSDoc: `Grows one entry per session; not evicted (bounded by sessions per process lifetime).`

---

### What Was Not Checked

- Citations verified inline against branch HEAD `c8e9c0687c327a1867c0ebab7164c4920ee643d7`.
- Stated intent: PR #1353 title + body — spec-compliance assessed; no unmet intent or scope creep found.
- Did not run the test suite locally (CI shows all GitHub Actions checks green).
- Did not verify whether `getSubagentLogSessionDir` returns a deterministically normalized path — symlink resolution or trailing-slash differences could cause cache misses where the same physical directory maps to two string keys.
- Did not check whether `BgJobLogWriter` has an analogous repeated-mkdir pattern that should be similarly optimized.
- **Merge conflict:** PR is currently `CONFLICTING` against `main` — rebase required before merge.

Issue #578, opened 2026-07-13
Defense-in-depth: contain project-local MCP stdio servers at spawn (env scrub + optional sandbox)


griffinwork40:
## Summary

Defense-in-depth companion to #571. #571 adds a **consent** gate for project-local `.mcp.json` (should this untrusted server run at all?). This issue is about **containment** once a server does run: agent-afk spawns MCP stdio servers as child processes with no sandbox, and passes a `${VAR}`-expanded `env` into them.

Recommended by the devils-advocate review on #571 as a separate control (blast-radius containment ≠ consent).

## Evidence

- `src/agent/mcp/transport.ts` (`StdioClientTransport`) spawns the server's `command`+`args` as a child process — no sandbox, no egress/fs restriction.
- `src/agent/mcp/env.ts:33-70` (`expandEnvString`/`expandEnvRecord`) expands `${VAR}` from `process.env` into the server's `env` at connect time. A server config with `env: { X: "${AWS_SECRET_ACCESS_KEY}" }` lands the operator's secret in the (possibly untrusted) child's environment — a 0-click exfil channel, worst in CI.
- MCP-bridged tools proxy straight to `client.callTool()` (`manager.ts`) with no additional permission wrapper.

## Impact

MED (defense-in-depth). Even with #571's consent gate, an approved-then-compromised server (or a server whose referenced binary is swapped post-approval) runs unconfined with access to expanded secrets and full network/fs.

## Suggested fix

For **project-layer** stdio servers specifically:
- Scrub/deny-list dangerous env inheritance (`NODE_OPTIONS`, `LD_PRELOAD`, `PYTHON*`) and refuse `${SECRET}` expansion of known-secret vars into project-layer server env (or require explicit per-server opt-in).
- Optionally wrap the spawn in an OS sandbox (macOS `sandbox-exec`/seatbelt, Linux `bwrap`/landlock) restricting network egress + fs-write outside cwd. Note: no clean Windows story — degrade gracefully.
- Keep user-global / CLI-config servers unaffected (trusted provenance).

## Acceptance criteria

- [ ] Project-layer server `env` cannot silently expand a known-secret var (`*_API_KEY`, `*_TOKEN`, AWS/GCP creds) without explicit opt-in.
- [ ] Dangerous inherited env vars scrubbed for project-layer servers.
- [ ] (Stretch) opt-in OS sandbox for project-layer stdio spawns on macOS/Linux.
[...]

PR #527, opened 2026-07-12
fix(subagent): resolve path-approval grants per executing session (#435/#514)

griffinwork40:
## Summary

Follow-up to #514 (now merged): closes the interactive-surface gap #514 left open for #435.

**The gap #514 left:** #514 composes an `agent`-tool `writeRoots` grant into a forked child's config, but the `path-approval` PreToolUse hook — which auto-denies forked children — reads grants from a process-global `pathApprovalGrantRef` pinned once to the **top-level** session's provider (`bootstrap.ts` / `telegram.ts`) and never re-pointed per fork. So on a default-permission REPL/Telegram session, a child dispatched with `writeRoots:['/sibling']` (a path the parent itself wasn't granted) is **auto-denied by the hook before the file handler — which honors `writeRoots` — ever runs.** The remedy #514 advertised ("re-dispatch with `writeRoots:[…]`") silently dead-ended on exactly the interactive surfaces it targeted. The same single ref also collides across concurrent sessions that share a hook bundle.

**The fix:** each per-query `SessionToolDispatcher` now carries its owning provider (which implements `GrantManager`) as `sessionGrantManager` and injects it onto every `PreToolUse`/`PostToolUse` context as `context.grantManager`. The `path-approval` and `bash-restriction` hooks prefer `context.grantManager` over the process-global ref, so they resolve the grants of the session **actually executing** the call — a forked child's own `cwd`/`readRoots`/`writeRoots` — instead of the top-level session's.

### Design notes

- Uses the **provider's** grant manager, not the dispatcher's own, to preserve the `resolveBase` divergence the hook relies on (providers protect the session's INITIAL resolveBase; the dispatcher the CURRENT one — see `grant-manager.ts` divergence #2).
- The process-global ref remains the **fallback** for non-dispatcher-originated dispatch (`SessionEnd`, unit tests), so top-level behavior is unchanged (the injected manager *is* the provider the ref points to) and headless/one-shot surfaces still fail open.
- **Confinement preserved:** a child write to a genuinely out-of-grant path is still auto-denied with the #435 remedy. A child gets only what its own `writeRoots` already intend — no new capability, no escalation.
- No `sessionId` threading, no keyed map, no eviction lifecycle; no changes to bootstrap/telegram/subagent wiring.
- Fixes the concurrent-top-level-session ref collision as a side effect (each call resolves via its own dispatcher's provider).

## How was this tested?

Run on this exact tree, rebased on current `main`:

[...]

Issue #500, opened 2026-07-11
grant-manager: 3 behavior divergences to reconcile (anchor policy needs a decision) — from #489 consolidation


griffinwork40:
Behavior divergences surfaced (report-only, not fixed) during the #489 grant-manager consolidation (3 copies → shared `PathGrantManager`). All preserved verbatim per-consumer during consolidation; independently re-confirmed by a 2026-07-11 regression audit (no behavior delta introduced — these are pre-existing). Part of #360.

1. **Dispatcher silently drops per-call `sessionId`** (`allow-dir.ts:26-31`): `SessionToolDispatcher.addReadRoot/addWriteRoot/revokeRoot` are 2-arg but structurally assignable to the 3-arg `GrantManager` interface; a caller passing per-call `sessionId` gets it silently ignored on the dispatcher (falls back to ctor-bound id).
2. **Opposite non-revocable-anchor policies** (needs a decision): after a worktree rename / `/cwd`, the dispatcher protects the NEW cwd (original launch dir becomes revocable) while the providers protect the INITIAL resolveBase. Both are documented, but they cannot both be the intended `/allow-dir` semantics — a follow-up should pick one. See `docs/dev/grant-manager-divergence.md` (Suspected finding #2).
3. **Provider `revokeRoot` audits unconditionally**: appends an audit row even on a no-op revoke (path never granted), unlike the grant paths' 196x-dedup invariant. Present in all three original copies; could balloon the audit ledger under repeated no-op revokes.

Source: PR #489 "Suspected findings" + divergence doc.

griffinwork40:
## Triage — re-verified against current HEAD (2026-07-12)

**Overall severity: medium · Priority: P2** (driven by Finding 2). All 3 CONFIRMED; line numbers shifted (dispatcher wrappers now `dispatcher.ts:308/317/328`; provider wrappers `anthropic-direct/index.ts:572-582`; revoke audit `grant-manager.ts:176`).

| # | Finding | Status | Severity | Effort / Risk |
|---|---------|--------|----------|---------------|
| 1 | Dispatcher `addReadRoot/addWriteRoot/revokeRoot` are 2-arg but structurally assignable to the 3-arg interface → a per-call `sessionId` is silently dropped (`allow-dir.ts:27-29` vs `dispatcher.ts:308-329`) | CONFIRMED | low — audit-attribution only; the REPL uses a provider (which threads it), so real-world impact near-nil | trivial / low |
[...]

PR #1364, opened 2026-08-28
fix(#528): retire pathApprovalGrantRef + Gate 1/Gate 2 containment regression test

### LINKED ISSUE #528: path-approval: retire process-global grant ref + guard two-gate containment duplication (follow-up to #527)
Follow-up to #527, which added **per-session grant injection**: every dispatcher built by a provider's `buildDispatcher` now injects its own provider as `context.grantManager`, and the path-approval + bash-restriction hooks prefer it over the process-global `pathApprovalGrantRef` (`context.grantManager ?? opts.getGrantManager()`). That fixed the forked-child `writeRoots` dead-end and the concurrent-top-level-session collision, but it intentionally left the global ref in place as a **fallback**. This issue tracks finishing that migration and guarding a related duplication.

Two narrow items. The deeper grant-manager divergences are **already tracked** (#500, #354) — see "Already tracked" below; this issue does not re-file them.

### 1. Retire the process-global `pathApprovalGrantRef`

Post-#527 state:
- **Defined / returned in bundle:** `default-hook-registry.ts:57,175,282`
- **Read (fallback only):** `path-approval-hook.ts:193` (Pre), `:305` (Post); `bash-restriction-hook.ts:161`
- **Written once, to the top-level provider:** `bootstrap.ts:897`, `telegram.ts:448`, `telegram.ts:514`

The ref is a single mutable slot pinned to the top-level provider — the exact primitive that caused #527's fork bug. It is now redundant on the primary path, but **not yet dead code**: two production dispatchers are constructed *outside* `buildDispatcher` and inject no grant manager, so they still fall back to the ref:
- `src/improve/eval-run/contracts.ts:132`
- `src/improve/eval-run/replay.ts:124`

Cleanup: decide whether those two eval-run dispatchers should carry a `sessionGrantManager` (or explicitly none), then delete the ref + both `getGrantManager: () => pathApprovalGrantRef.current` fallbacks + the three bootstrap/telegram write sites, and simplify the wiring. Update the tests that currently assert `.current` is set (`interactive-lifecycle.test.ts`, `telegram/hook-wiring.test.ts`).

Payoff: removes the last shared-mutable-slot footgun from the confinement path — no session can observe another session's grants, even in the fallback.

### 2. Guard the two-gate containment duplication

"Is this path inside the session's roots?" is computed **twice**, by design, for two different purposes:
- **Gate 1 (enforcement):** `resolveAndContain` in `tools/handlers/_cwd-utils.ts` — allows/denies the actual `write_file`/`read_file`/`edit_file`.
- **Gate 2 (elicitation):** `wouldBeRestricted` in `tools/hooks/path-approval-hook.ts` — re-implements the same containment to decide whether to *prompt the operator first* (its own comment notes it "reproduces the handler's containment check").

The handler can't prompt and the hook doesn't enforce, so both must exist — but nothing pins them together. If one's containment logic changes and the other's doesn't, the approval prompt silently diverges from actual enforcement (prompt for a path that is then allowed, or allow-through a path that should have prompted). Add a shared containment predicate both gates call, **or** a regression test asserting they agree across a fixed path matrix (in-root, out-of-root, symlink, `..` escape, granted-root).

### Already tracked — do NOT duplicate here
- **#500** — provider/dispatcher `resolveBase` anchor divergence + dispatcher-drops-per-call-`sessionId` + unconditional revoke-audit. The deeper grant-manager semantics; anchor policy decision pending. See `docs/dev/grant-manager-divergence.md`.
- **#354** — bash tool has no `readRoots`/`writeRoots` containment (why Gate 2's bash arm is inherently best-effort).

### Acceptance criteria
- `pathApprovalGrantRef` is deleted (with the two eval-run dispatchers handled), **or** a `docs/` note records the decision to keep it and why.
- Gate 1 / Gate 2 containment agreement is enforced by a shared predicate or a regression test.

Refs: #527 (origin), #500, #354.

griffinwork40:
## Summary

Fixes #528 — two concrete items:

### Item 1: Retire `pathApprovalGrantRef` (process-global grant ref)

The `pathApprovalGrantRef` was a mutable ref populated by REPL and Telegram bootstraps so path-approval and bash-restriction hooks could access the grant manager. Since #527 added `context.grantManager` (per-session injection by the dispatcher), the process-global ref is redundant.

**Changes:**
- `DefaultHookRegistryResult`: Drop `pathApprovalGrantRef` field
- `createDefaultHookRegistry`: Remove the local ref variable and its return
- `path-approval-hook.ts`: Read grant manager from `context.grantManager` exclusively in production; introduce `state.lastSeenGrantManager` for SessionEnd safety-net; keep `getGrantManager` as optional deprecated test-only fallback
- `bash-restriction-hook.ts`: Same pattern
- `bootstrap-hooks.ts`, `bootstrap-wiring.ts`, `bootstrap.ts`: Remove ref parameter threading
- `telegram/session-anthropic.ts`, `session-openai.ts`, `session-xai.ts`: Remove `hookBundle.pathApprovalGrantRef.current = provider` lines
- Tests: `interactive-lifecycle.test.ts` rephrased to observe `wireProviderGrants` call; `hook-wiring.test.ts` asserts bundle has NO `pathApprovalGrantRef` field (the retirement regression guard)

### Item 2: Gate 1 / Gate 2 containment agreement regression test

Adds `_cwd-utils.test.ts` section asserting that `resolveAndContain` (Gate 1) and `wouldBeRestricted` (Gate 2) agree on every case in a 10-case path matrix: in-root, out-of-root, `..`-escape, granted extra root, extra-root `..`-escape, unconfined session, bypass mode, symlink-outside-root, relative in-root, relative `..`-escape. Drift between the two would mean the path-approval hook either over-prompts or silently skips paths the handler then rejects.

## Test results

```
Test Files  5 passed (5)          ← targeted test files
Tests      189 passed | 1 skipped
```

Full suite: 940 passed, 1 pre-existing failure (`writer.test.ts` subprocess test — fails in worktrees due to dynamic import path, not related to this PR).

## Investigations (no implementation — design decisions pending)

See PR description for #578 and #500 findings.

### DIFF
--- a/src/agent/default-hook-registry.ts
+++ b/src/agent/default-hook-registry.ts
@@ -24,7 +24,6 @@ import {
   type PathApprovalSurface,
 } from './tools/hooks/path-approval-hook.js';
 import { createBashRestrictionHook } from './tools/hooks/bash-restriction-hook.js';
-import type { GrantManager } from '../cli/slash/commands/allow-dir.js';
 import type { PermissionMode } from './types/sdk-types.js';
 import type { TraceSink } from './trace/index.js';
 import type { LoadedHooksConfig } from './hooks/config-loader.js';
@@ -40,24 +39,6 @@ export interface SubagentCompleteInfo {
 export interface DefaultHookRegistryResult {
   registry: HookRegistry;
   memoryStore: MemoryStore;
-  /**
-   * Mutable ref the surface bootstrap MUST populate with the active provider
-   * (which implements {@link GrantManager}) after session construction. Path-
-   * approval + bash-restriction hooks dereference this on every PreToolUse;
-   * leaving it null makes the path-approval hook fail OPEN — it skips the
-   * elicitation prompt (the typed-file-tool handler's own resolveAndContain
-   * still enforces containment), and the bash hook's restricted-root substring
-   * check also fails open. (The bash interpreter denylist still hard-blocks
-   * regardless — it runs before the grant-manager gate. See the registration
-   * block below for the full per-surface breakdown.)
-   *
-   * Wiring pattern (see `src/cli/commands/interactive/bootstrap.ts:532`):
-   *   pathApprovalGrantRef.current = provider;
-   *
-   * The same ref drives `setAllowDirDispatcher(provider)` callers — both
-   * read the live provider grant API.
-   */
-  pathApprovalGrantRef: { current: GrantManager | undefined };
 }
 
 /**
@@ -180,10 +161,11 @@ export function createDefaultHookRegistry(
   // path first when both could apply (e.g. an edit_file followed by a bash
   // grep against the same restricted root within one turn).
   //
-  // Invariant: only the REPL and Telegram bootstraps wire
-  // `pathApprovalGrantRef.current` (see bootstrap.ts / telegram bot.start).
-  // Headless surfaces (afk chat, daemon, threads) register these hooks but
-  // never wire the ref, so on those surfaces:
+  // Invariant: path-approval and bash-restriction hooks read the grant manager
+  // from `context.grantManager` (injected per-session by the dispatcher since
+  // #527; the former process-global ref was retired in #528). Headless surfaces
+  // (afk chat, daemon, threads) never wire a sessionGrantManager into the
+  // dispatcher, so on those surfaces:
   //   - path-approval PreToolUse: fails open (no prompt; the typed-tool
   //     handler's own resolveAndContain still enforces containment);
   //   - bash restricted-root substring check: fails open (no backstop);
@@ -193,9 +175,10 @@ export function createDefaultHookRegistry(
   //     blocked with no recourse. Opt headless flows back into the guard with
   //     AFK_FORCE_BASH_INTERPRETER_GUARD=1; disable the whole feature with
   //     AFK_DISABLE_PATH_APPROVAL=1.
-  const pathApprovalGrantRef: { current: GrantManager | undefined } = {
-    current: undefined,
-  };
+  //
+  // The former process-global `pathApprovalGrantRef` has been retired (#528).
+  // Both hooks now read the grant manager exclusively from `context.grantManager`
+  // — injected per-session by the dispatcher since #527.
   const disabled = env.AFK_DISABLE_PATH_APPROVAL === '1';
   if (disabled && !warnedPathApprovalDisabled) {
     warnedPathApprovalDisabled = true;
@@ -207,7 +190,6 @@ export function createDefaultHookRegistry(
   }
   if (!disabled) {
     const pathApproval = createPathApprovalHook({
-      getGrantManager: () => pathApprovalGrantRef.current,
       // Prefer an explicit getCwd callback (REPL/Telegram bootstrap thread a
       // live `extras.cwd`); otherwise fall back to the static cwd carried in
       // main's `agentOptions` (scheduler/chat/threads callers). Both feed the
@@ -233,7 +215,6 @@ export function createDefaultHookRegistry(
     registry.register(
       'PreToolUse',
       createBashRestrictionHook({
-        getGrantManager: () => pathApprovalGrantRef.current,
         // Granular knobs for the interpreter-eval denylist (the rest of
         // path-approval is unaffected; AFK_DISABLE_PATH_APPROVAL=1 short-
         // circuits the whole `if (!disabled)` block above):
@@ -314,7 +295,7 @@ export function createDefaultHookRegistry(
     });
   }
 
-  return { registry, memoryStore: store, pathApprovalGrantRef };
+  return { registry, memoryStore: store };
 }
 
 /**
--- a/src/agent/hooks.ts
+++ b/src/agent/hooks.ts
@@ -216,7 +216,7 @@ export interface PreToolUseContext {
    * cwd/readRoots/writeRoots rather than a process-global ref pinned to the
    * top-level session (the #435/#514 write-confinement gap). When absent
    * (non-dispatcher-originated dispatch, unit tests), those hooks fall back to
-   * their `opts.getGrantManager()` ref, preserving prior behavior.
+   * their per-session closure state, preserving prior behavior.
    */
   grantManager?: GrantManager;
 }
--- a/src/agent/tools/dispatcher.ts
+++ b/src/agent/tools/dispatcher.ts
@@ -186,11 +186,10 @@ export interface SessionToolDispatcherOptions {
    * The PROVIDER that owns this dispatcher (it implements {@link GrantManager}).
    * The provider's `buildDispatcher` passes `this`; the dispatcher injects it
    * onto every PreToolUse/PostToolUse context as `context.grantManager` so
-   * path-scoped hooks resolve THIS session's live grants instead of the
-   * process-global `pathApprovalGrantRef` — which is pinned to the top-level
-   * session and blind to a forked child's own writeRoots (#435/#514). Optional:
-   * test dispatchers that construct directly leave it unset and the hooks fall
-   * back to their ref, preserving prior behavior.
+   * path-scoped hooks resolve THIS session's live grants rather than a stale
+   * process-global ref (#435/#514, retired in #528). Optional: test dispatchers
+   * that construct directly leave it unset and the hooks fail open (no grant
+   * manager → no containment prompt, handler resolveAndContain still enforces).
    */
   sessionGrantManager?: GrantManager;
   /** Witness-layer trace writer. When provided, every PreToolUse and
--- a/src/agent/tools/handlers/_cwd-utils.test.ts
+++ b/src/agent/tools/handlers/_cwd-utils.test.ts
@@ -380,3 +380,136 @@ describe('fallbackBase — factory-cwd resolve tier (issue #434)', () => {
     expect(wouldBeRestricted(OUTSIDE, baseless, 'read', undefined).restricted).toBe(false);
   });
 });
+
+// ---------------------------------------------------------------------------
+// Gate 1 / Gate 2 containment agreement (#528 regression guard)
+//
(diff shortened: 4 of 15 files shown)


griffinwork40 (commented):

---

**PR:** fix(#528): retire pathApprovalGrantRef + Gate 1/Gate 2 containment regression test
**Reviewed ref:** `d13ece409c` | 15 files | ~494 lines | regime: full

**0 critical, 0 high → MERGE (with follow-ups); 2 medium advisory (both waived — bounded correctness, no security impact), 3 low, 2 nit.**

The refactor is structurally sound. All 15 files are consistent: `pathApprovalGrantRef` is cleanly removed from types, variables, return values, all callers, and all three Telegram session builders. The hooks read exclusively from `context.grantManager` with `opts.getGrantManager` retained as a deprecated test-only fallback. The Gate 1/Gate 2 regression test covers the 10-case path matrix with a correct `assertAgree` helper. Spec-compliance is fully met on both stated items with no scope creep.

---

### Advisory findings (non-blocking)

**1. medium · blocking:false · correctness · `path-approval-hook.ts:221,453`**
`state.lastSeenGrantManager` is updated on every PreToolUse (line 221), so within a single dispatcher's lifetime the cache stays fresh. The edge case: if a provider swap occurs mid-session and SessionEnd fires before any post-swap tool call, `lastSeenGrantManager` points to the old provider. This is bounded — `onceApproved` entries were approved by the old provider so revoking on it is correct, and Gate 1 (`resolveAndContain`) still enforces at execution time. The failure mode is a missed cleanup, not a grant escalation.
**Suggestion:** Document the coupling invariant ("one dispatcher lifetime per hook closure") in `PathApprovalState.lastSeenGrantManager`'s JSDoc, or add a test that asserts `lastSeenGrantManager === context.grantManager` when both are defined during PostToolUse.
· waived: bounded non-security defect — cleanup path only, no data loss, no escalation; Gate 1 still enforces.

**2. medium · blocking:false · correctness · `path-approval-hook.ts:374`**
`postToolUseImpl` now falls back to `state.lastSeenGrantManager` instead of `opts.getGrantManager()`. Any test that exercises PostToolUse independently (without a prior PreToolUse in the same hook instance) will silently skip Once-grant revocation. This is a behavioral change for test-only code paths — production dispatcher always runs Pre before Post, so no production impact.
**Suggestion:** Verify no existing `path-approval-hook.test.ts` tests exercise PostToolUse independently of a Pre call. Low risk in production.
· waived: bounded non-data-affecting defect — test-only path, production dispatcher invariant holds.

**3. low · blocking:false · test-coverage · `interactive-lifecycle.test.ts:~1280`**
The second grant-wiring test ("calls wireProviderGrants even when the startup provider lacks GrantManager methods") spies on the `wireProviderGrants` wrapper itself, so `setAllowDirSpy` fires unconditionally for both the `isGrantManager` branch and the `else-warn` branch. The test no longer distinguishes whether `setAllowDirDispatcher` was called.
**Suggestion:** Add a separate spy on `setAllowDirDispatcher` to assert it was NOT called for the non-GrantManager path, restoring the meaningful negative assertion.

**4. low · blocking:false · test-coverage · `hook-wiring.test.ts:85–100`**
The `pathApprovalGrantRef` absence check (`(bundle as Record<string, unknown>)['pathApprovalGrantRef'] === undefined`) correctly guards the retirement, but cannot catch a future refactor that reintroduces a process-global ref under a different name. The prior positive wiring assertions (assign, then verify `.current`) were deleted without replacement.
**Suggestion:** Acceptable for the retirement intent — noted for future reference.

**5. nit · blocking:false · api-compat · `bash-restriction-hook.ts:139`**
`opts.getGrantManager` uses an inline `import()` type expression instead of a top-level `import type { GrantManager }` — inconsistent with `path-approval-hook.ts` which retains the top-level import. No runtime impact.
**Suggestion:** Restore the top-level `import type { GrantManager }` for consistency.

**6. nit · blocking:false · spec-compliance · `path-approval-hook.ts:95–108`**
Both `@deprecated getGrantManager` fields lack a `@since` version or removal target, making the migration window ambiguous for any external test infrastructure using the public type.
**Suggestion:** Add `@since` and `@removeIn` annotations.

### What was not checked
- Citations verified inline against branch HEAD `d13ece409c`.
- Did not run the test suite; CI reports green on all platforms.
- Stated intent: PR #1364 title+body — spec-compliance assessed (both items fully met, no scope creep).
- Did not review runtime behavior under concurrent tool calls (race conditions between Pre/Post across parallel tool dispatches).

griffinwork40 (commented):

## Review: fix(#528): retire pathApprovalGrantRef + Gate 1/Gate 2 containment regression test

**Reviewed ref:** `d13ece409c56ea8f9fa318b68884c61bd0ee19ed`
**Regime:** full (14 files, ~400 lines)
**Change type:** refactor

---

**Decision: MERGE — 0 blocking (5 low, 3 nit).**

All surviving findings are low/nit severity. The retirement is clean — every internal consumer of `pathApprovalGrantRef` is updated in the same commit, the fallback chain is logically sound, and the Gate 1/Gate 2 regression test is well-structured.

---

### Findings

**1.** `low` · blocking:false · confidence:medium · security · `src/agent/tools/hooks/path-approval-hook.ts:375` · ref:`d13ece40` · code-logic · **Resume-swap stale `lastSeenGrantManager`: after a mid-session `/resume`, the hook closure persists but `lastSeenGrantManager` points at the old provider until the next `preToolUseImpl` writes the new one — outstanding `onceApproved` entries from the old provider would revoke against the wrong manager.** The blast radius is near-zero: this requires an abort-mid-tool-call followed by a resume-swap (an exceedingly rare sequence), and `resolveAndContain` at the handler layer still enforces containment regardless. · blocking preserved from pre-downgrade medium
  - *suggestion:* When updating `lastSeenGrantManager` in `preToolUseImpl`, compare against the previous value; if it changed, revoke all outstanding `onceApproved` entries via the old manager first.

**2.** `low` · blocking:false · confidence:high · security · `src/agent/tools/hooks/path-approval-hook.ts:365-380` · ref:`d13ece40` · code-logic · **`postToolUseImpl` drops the `opts.getGrantManager` fallback (only `state.lastSeenGrantManager`), so a unit test exercising revoke-in-isolation without a prior `preToolUseImpl` would pass vacuously.** Not a production issue — `onceApproved` is only populated by `preToolUseImpl`, so if pre never ran, there is nothing to revoke. The concern is limited to test-harness coverage fidelity.
  - *suggestion:* Document that PostToolUse tests must prime state with a PreToolUse call first, or seed `lastSeenGrantManager` from `opts.getGrantManager?.()` at hook initialization.

**3.** `low` · blocking:false · confidence:high · api-compat · `src/agent/default-hook-registry.ts:39` · ref:`d13ece40` · diff-context · **`DefaultHookRegistryResult.pathApprovalGrantRef` field removed from exported interface — breaking for any consumer that destructures it.** All internal consumers (2 production importers: `default-hook-registry.ts`, `telegram/afk-hook-bundle.ts`) are updated in this commit. The interface is not re-exported through any public package index, so external breakage is unlikely. · blocking preserved from pre-downgrade high
  - *suggestion:* A `/** @deprecated */ pathApprovalGrantRef?: never` tombstone for one version would give any out-of-tree consumer a typed signal.

**4.** `low` · blocking:false · confidence:medium · correctness · `src/agent/tools/hooks/bash-restriction-hook.ts` · ref:`d13ece40` · code-logic · **The two hooks have asymmetric fallback chains: `bash-restriction-hook` consults `opts.getGrantManager?.()` in post-use; `path-approval-hook` does not.** Not functionally wrong (bash-restriction only has a single PreToolUse handler, no PostToolUse), but the inconsistency is a maintenance hazard.
  - *suggestion:* Align the two hooks to the same pattern.

**5.** `low` · blocking:false · confidence:high · test-coverage · `src/agent/tools/hooks/path-approval-hook.test.ts` · ref:`d13ece40` · absence (confirmed-absent) · **No test exercises the `lastSeenGrantManager` cache-miss path (PostToolUse or SessionEnd fires without a preceding PreToolUse).** The code handles this correctly by design (`onceApproved` is empty when pre never ran → `state.lastSeenGrantManager` undefined is harmless), but the invariant is protocol-documented, not mechanically tested. · blocking preserved from pre-downgrade high · search pattern: `lastSeenGrantManager` in test files returned 0 hits
  - *suggestion:* Add a test that constructs the hook, dispatches SessionEnd without any prior PreToolUse, and asserts `onceApproved.clear()` ran cleanly.

**6.** `nit` · blocking:false · confidence:high · api-compat · `src/cli/commands/interactive/bootstrap-wiring.ts:88` · ref:`d13ece40` · diff-context · **`wireProviderGrants` drops the `pathApprovalGrantRef` positional parameter — a signature change.** Internal function with 1 call site (`bootstrap.ts`), updated in this commit. Not a public API surface.

**7.** `nit` · blocking:false · confidence:high · spec-compliance · `src/agent/tools/dispatcher.ts` · ref:`d13ece40` · diff-context · **JSDoc-only change in dispatcher.ts not mentioned in the PR body's file inventory.** The change updates the `sessionGrantManager` comment to reference #528. Harmless but undocumented.

**8.** `nit` · blocking:false · confidence:high · spec-compliance · `src/agent/hooks.ts` · ref:`d13ece40` · diff-context · **1-line comment update in hooks.ts not mentioned in the PR body.** Changes "their `opts.getGrantManager()` ref" to "their per-session closure state."

---

### What was not checked
- Citations verified inline against branch HEAD `d13ece409c56ea8f9fa318b68884c61bd0ee19ed`.
- Stated intent: PR #1364 title + body — spec-compliance assessed.
- Did not run the test suite (build/test verification is CI's job — CI is green).
- Did not review the full 515-line `_cwd-utils.test.ts` file; verified only the new 136-line section added by this PR.
- Did not verify runtime behavior of the resume-swap path end-to-end.

PR #1361, opened 2026-08-28
feat(render): wire toolCard into task-view + fix duplicate barrel export

griffinwork40:
## What

Wire the `toolCard()` component from the component library into its first live consumer, and fix a minor barrel export duplicate.

### 1. task-view.ts: toolCard integration

The `/task` history viewer previously hand-rendered `tool_use` and `tool_result` content blocks with ad-hoc `palette.tool()` / `palette.success()` / `palette.error()` calls and divergent glyph choices (`▶` for tool_use, `✓`/`✗` for tool_result). Now uses `toolCard()` from `src/cli/render/tool-card.ts`, which provides:

- Consistent status badges via `statusBadge()` (same glyphs as the live overlay)
- Sanitized input/output via `sanitizeForDisplay()`
- Width-aware truncation via `truncateDisplayWidth()`
- Collapsed/expanded support for future use

This is the first live consumer of `toolCard()` outside its unit tests. The component was built in sprint 0 (#1341) but had zero non-test call sites until now.

### 2. render/index.ts: remove duplicate export

`interrupt-peek.js` was re-exported twice (lines 25 and 27). Harmless but noisy.

## Why now

The component library initiative (started 2026-08-27) built 6 components, but shadow-verify confirmed only 2 of 6 were wired into live paths. This PR begins the integration sprint -- wiring existing components into their natural consumers rather than building new ones.

## Investigation findings

During this session, a full audit of the component library revealed:
- **SubagentStatusBar, InterruptPeek**: already live in the overlay (wired in sprint 0)
- **ErrorCard**: live in scrollback (wired in sprint 0)
- **ToolCard**: now live via this PR (task-view)
- **StreamProgress, CompactDiffView**: remain unintegrated -- their production counterparts (`formatProgressBanner`, `formatDiffBlock`) are more specialized (multi-line banner with thinking clauses/stopping state/parallel counts vs. single-line spinner; tree-embedded indented diff with WeakMap memoization/env toggles vs. boxed standalone diff). Forcing integration would regress the UX.
- **Loading-state feedback plan (2026-07-02)**: all 3 items already implemented (sanitizeLabel, semantic progress descriptions, lastProgressByTask cleared on finalize)

## Tests

- `task-view.test.ts`: 17/17 pass
- `tool-card.test.ts`: 30/30 pass
- `pnpm lint`: clean
- `fix:pins:check`: clean

PR #1321, opened 2026-08-27
feat(render): component library foundation — SubagentStatusBar, StreamProgress, ErrorCard + live overlay integration

griffinwork40:
## Summary

Component library foundation for surpassing Claude Code's terminal UX — rendering primitives + live integration + typography polish.

### Why component library, not renderer rewrite

Devils-advocated three approaches: CellState/React renderer rewrite (scored 7/20), targeted additions (17/20), and component library on existing compositor (18/20). Claude Code's visual advantage is its **component library** (collapsible tool cards, progress animations, status bars), not its renderer. Building components on the existing 22,500-line battle-tested compositor delivers visible UX improvements weekly without a multi-month rewrite or bus-factor-1 dependency.

### Rendering primitives (`src/cli/render/`)

| Component | Interface | Output |
|-----------|-----------|--------|
| `subagentStatusBar(spec)` | `{ label, phase?, elapsedMs, batchIndex?, batchSize? }` | `◉ research-agent  ──  thinking…  12s  ∥2/4` |
| `subagentStatusStack(entries, maxLines?)` | Array of specs, 3-line cap | Stacked bars + `… +N more running` overflow |
| `streamProgress(spec)` | `{ label, spinnerFrame, elapsedMs, tokenCount?, costCents? }` | `⠹ Generating…  1.2k tokens  4.7s` |
| `errorCard(spec)` | `{ title?, body, hint? }` | Bordered red card with dim/italic recovery hint |

All follow the established pattern: stateless `(spec) → string`, semantic `palette.<role>` colors, internal terminal width reads, ANSI-safe truncation.

### Live integration

**SubagentStatusBar overlay slot** — registered as `'subagent-status'` in the OverlayComposer z-order, above `'tool-lane'`. When subagents dispatch during a parent turn, users see live-updating status bars:

```
◉ research-agent  ──────────  thinking…  12s
◉ explore  ──────────────────  8s
```

- First subagent event → entry added to `activeSubagents` map with label + start time
- 250ms `setInterval` ticker → updates `elapsedMs`, marks slot dirty, flushes overlay
- Subagent `done`/`error` → entry removed, slot auto-clears
- 3-line height cap prevents overlay overflow (excess → `… +N more running`)

**ErrorCard replaces errorBox** — both callers (error presenter + stream-renderer error path) now render structured error cards with recovery hints instead of plain red boxes.

### Typography

- Shell tool glyph `$` → `▸` (consistent with Unicode visual language — every other category uses a Unicode glyph)
[...]

PR #1366, opened 2026-08-28
feat(render): pre-execution diff preview for edit_file

griffinwork40:
## Summary

Pre-execution diff preview for `edit_file` -- show users what's about to change before it happens.

When the agent calls `edit_file`, the live overlay now shows a diff preview **before** the edit is applied, using the existing `CompactDiffView` component. This is the #1 UX gap between Claude Code and agent-afk -- CC renders a `StructuredDiff` in its permission dialogs before any write occurs.

### How it works

1. A new `PreToolUse` hook (`edit-preview-hook.ts`) fires on every `edit_file` call
2. The hook computes a diff from `old_string`/`new_string` (no file read needed)
3. The diff is delivered to the tool lane via a mutable callback ref (bridging the agent/CLI layer gap)
4. The overlay renders the diff under the in-flight tool entry with a `Proposed` label
5. Once the tool result arrives, the preview is replaced by the normal outcome

### What's new

| File | Change |
|------|--------|
| `src/cli/render/preview-diff.ts` | New `previewDiff()` component wrapping `compactDiffView` with "Proposed" framing |
| `src/agent/tools/hooks/edit-preview-hook.ts` | New `PreToolUse` hook: guards for `edit_file`, computes diff, delivers to lane |
| `src/agent/hooks.ts` | Added `toolUseId?` to `PreToolUseContext` (needed for lane correlation) |
| `src/agent/tools/dispatcher.ts` | Spread `call.id` into `preCtx` |
| `src/cli/commands/interactive/tool-lane.ts` | Added `addPreviewDiff()` method + overlay rendering |
| `src/cli/commands/interactive/tool-lane-render.ts` | Added `previewDiff?` field to `ToolEntryFields` |
| `src/agent/default-hook-registry.ts` | Register hook + expose `addPreviewDiffRef` |
| `src/cli/commands/interactive/bootstrap-hooks.ts` | Thread ref to caller |
| `src/cli/commands/interactive/bootstrap.ts` | Thread ref into `InteractiveCtx` |
| `src/cli/commands/interactive/shared.ts` | Added ref to `InteractiveCtx` + `TurnHandles` |
| `src/cli/commands/interactive/turn-handler.ts` | Thread ref into `StreamRendererOptions` |
| `src/cli/_lib/stream-renderer-options.ts` | Added `addPreviewDiffRef?` option |
| `src/cli/_lib/stream-renderer.ts` | Wire ref in `arm()` each turn |

### Design decisions

- **Mutable callback ref pattern**: The hook lives in the agent layer, the tool lane in the CLI layer. A `{ current: () => {} }` ref is created at registry time, armed each turn by the StreamRenderer, and called synchronously by the hook. Non-interactive surfaces never arm the ref (remains no-op).
- **Subagent skip**: The hook returns `{}` for subagents (`parentSessionId` set) since they have no overlay.
- **`AFK_SHOW_DIFFS=0` parity**: The preview respects the existing diff opt-out via `formatDiffBlock`'s internal `diffsDisabled()` gate.
- **No `write_file` support** (different input shape -- needs a file read for preview; separate scope).

### Tests

- `preview-diff.test.ts` -- 5 tests (label, hunk content, maxLines, filePath, ANSI safety)
- `edit-preview-hook.test.ts` -- 7 tests (normal fire, no-op, subagent skip, wrong tool, wrong event, no injectContext, missing toolUseId)
- `tool-lane.test.ts` -- 5 new tests in `addPreviewDiff` describe block (overlay render, result replacement, unknown id, AFK_SHOW_DIFFS suppression, flush exclusion)

### Part of the component library initiative

This activates `CompactDiffView` -- one of two components built in the Aug 27-28 render sprint (#1321-#1361) that had zero live callers. The component library goal is surpassing Claude Code's terminal UX by building composable components on the existing compositor.

### DIFF
--- a/.filesize-baseline.json
+++ b/.filesize-baseline.json
@@ -1,165 +1,46 @@
 {
   "limit": 350,
   "entries": {
-    "scripts/audit-sdk-dependency.ts": {
-      "loc": 450,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/background-registry.ts": {
-      "loc": 374,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/daemon/scheduler.ts": {
-      "loc": 543,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/memory/memory-store.ts": {
-      "loc": 687,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/memory/memory-tools.ts": {
-      "loc": 437,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/providers/openai-compatible/index.ts": {
-      "loc": 396,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/providers/openai-compatible/query.ts": {
-      "loc": 803,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/session/agent-session.ts": {
-      "loc": 837,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/session/stream-consumer.ts": {
-      "loc": 359,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/tools/compose-executor.ts": {
-      "loc": 559,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/tools/dispatcher.ts": {
-      "loc": 611,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/tools/readonly-bash.ts": {
-      "loc": 415,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/tools/schemas.ts": {
-      "loc": 1277,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/tools/subagent-executor.ts": {
-      "loc": 481,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/trace/events.ts": {
-      "loc": 424,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/trace/receipt.ts": {
-      "loc": 412,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/agent/worktree-sweep.ts": {
-      "loc": 512,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/cli/_lib/testing/virtual-screen.ts": {
-      "loc": 422,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/cli/commands/chat.ts": {
-      "loc": 570,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/cli/commands/daemon.ts": {
-      "loc": 392,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/cli/commands/farm.ts": {
-      "loc": 436,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/cli/commands/interactive.ts": {
-      "loc": 599,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/cli/commands/interactive/loop-iteration.ts": {
-      "loc": 464,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/cli/commands/interactive/tool-lane.ts": {
-      "loc": 448,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/cli/commands/interactive/turn-handler.ts": {
-      "loc": 365,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/cli/commands/interactive/worktree.ts": {
-      "loc": 478,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/cli/commands/trace.ts": {
-      "loc": 573,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/cli/elicitation/agent-question.ts": {
-      "loc": 399,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/cli/slash/commands/info.ts": {
-      "loc": 443,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/cli/terminal-compositor.input-dispatch.ts": {
-      "loc": 512,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/config/env.ts": {
-      "loc": 1846,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/config/import-sources.ts": {
-      "loc": 373,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/improve/eval-run/contracts.ts": {
-      "loc": 491,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/improve/eval-run/runner.ts": {
-      "loc": 372,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/improve/propose/template-engine.ts": {
-      "loc": 459,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/skills/audit-fit/index.ts": {
-      "loc": 421,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/telegram/bot.ts": {
-      "loc": 369,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/telegram/handlers/farm-callbacks.ts": {
-      "loc": 392,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/telegram/handlers/message.ts": {
-      "loc": 581,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    },
-    "src/telegram/session-manager.ts": {
-      "loc": 468,
-      "reason": "legacy: predates the ceiling gate; pending concern extraction"
-    }
+    "scripts/audit-sdk-dependency.ts": { "loc": 450, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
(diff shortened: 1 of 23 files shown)


griffinwork40 (commented):

---

**PR:** feat(render): pre-execution diff preview for edit_file
**Reviewed ref:** `22486ffa907b91df6b91fd81a0efa514f85d898f` | 17 files | ~800 lines (incl. .filesize-baseline.json reformat) | regime: full

## Prior automated rounds

No prior automated rounds on this PR — no dispositions to carry forward.

**0 critical, 0 high → MERGE (with follow-ups); 3 medium advisory (non-blocking), 1 low advisory, 2 nit.**

The core feature is structurally sound: the mutable-ref pattern correctly bridges the agent/CLI layer without coupling them, the hook guards are complete (non-interactive surfaces never arm the ref, subagents skip via `parentSessionId`, `AFK_SHOW_DIFFS=0` parity is delegated to `formatDiffBlock`'s `diffsDisabled()` gate), and the overlay-only guarantee holds — `previewDiff` is rendered exclusively in the in-flight `else` branch (no `entry.result`) in `getOverlay()`, and all flush/scrollback paths (`renderGroupedRootTools`, `tool-lane-render-children.ts` flush path) only render `entry.diff`, never `entry.previewDiff`. CI is green across all platforms. 17 new tests cover the main paths.

---

**1. medium · blocking:false · correctness · `src/cli/commands/interactive/tool-lane-render-children.ts:285–290` · ref:22486ffa · file-state**

`renderOverlayChildren` (the path for `edit_file` children of NESTING roots — skill, Agent, compose) renders the in-flight else branch with only the elapsed counter and `thinkingTail`. There is no `child.previewDiff` render block here, unlike the flat-root path in `tool-lane.ts:577–582`. An `edit_file` call issued from inside a skill or multi-agent dispatch shows no pre-execution diff preview — silently, with no error. This is a partial feature gap against the stated intent (which says "when the agent calls `edit_file`" without qualification).

**Suggestion:** Add a `previewDiff` block to the in-flight else branch in `renderOverlayChildren` after line 290, guarded by `!child.result && child.previewDiff`, mirroring the pattern at tool-lane.ts:577.
· waived: bounded display-only gap, no safety/data impact; feature works correctly for flat-root edit_file calls (the most common case).

**2. medium · blocking:false · correctness · `src/cli/_lib/stream-renderer.ts:326–331` (arm) and `543–586` (dispose) · ref:22486ffa · file-state**

`dispose()` does not reset `addPreviewDiffRef.current` back to `() => {}` before tearing down the turn. The session-lifetime ref retains a closure over the disposed turn's `toolLane` and nulled `overlayComposer` until the next turn's `arm()` overwrites it. In the narrow inter-turn gap, a delayed PreToolUse dispatch (theoretically possible if hook execution races with turn teardown) writes into the dead `toolLane` silently — `overlayComposer?.flush()` is a no-op on `null`, so no throw, but the write is lost.

**Suggestion:** At the top of `dispose()`, before building `DisposeCtx`, add: `if (this.addPreviewDiffRef) this.addPreviewDiffRef.current = () => {};`
· waived: bounded non-data-affecting defect — PreToolUse hooks in practice only fire during active tool dispatch (within a turn), not in the inter-turn gap; the blast radius is a silently-dropped display update.

**3. medium · blocking:false · spec-compliance · `src/agent/tools/hooks/edit-preview-hook.ts:55–64` · ref:22486ffa · diff-context**

The hook calls `computeLineDiff(oldStr, newStr)` unconditionally, even when `AFK_SHOW_DIFFS=0` is set. `formatDiffBlock` (called by the overlay renderer) correctly returns `[]` on suppression, so display is correctly suppressed — but the O(m×n) LCS computation runs on every `edit_file` call regardless of the display opt-out. For a 500-line file edit, this is O(250,000) synchronous work per call on surfaces where the result is never shown.

**Suggestion:** Import `diffsDisabled` (from `tool-lane-format-diff.ts`) and return `{}` early in the hook if diffs are suppressed, before calling `computeLineDiff`.
· waived: bounded non-behavioral defect — wasted CPU on suppressed-display deployments only; the compute cap at `MAX_DIFF_CELLS = 4_000_000` limits worst-case allocation.

**4. low · blocking:false · test-coverage · `src/agent/tools/hooks/edit-preview-hook.test.ts` and `src/agent/hooks/config-bridge.test.ts` · ref:22486ffa · diff-context**

No test exercises the behavior when `addPreviewDiffRef` is constructed but the StreamRenderer never calls `arm()` (non-TTY path). In production this is always safe — the ref starts as `{ current: () => {} }` and the hook calls `current(...)` which is a no-op. But the hook tests inject a spy callback directly, so coverage of the no-op path is implicit rather than explicit.

**Suggestion:** Add one hook test: construct `createEditPreviewHook({ addPreviewDiffRef: { current: () => {} } })`, fire it with a valid edit_file context, and assert it does not throw and returns `{}`.

**5. nit · blocking:false · correctness · `src/agent/tools/hooks/edit-preview-hook.ts:21` · ref:22486ffa · diff-context**

```typescript
const diff = computeLineDiff(oldStr, newStr);
if (diff === null) return {};  // line 21
```
This guard is unreachable in practice: the preceding `if (oldStr === newStr) return {}` (line 19) excludes the only documented null-return case for `computeLineDiff`. The 0-hunks case (the other `null` source) cannot arise for two distinct strings under the current implementation. The guard is harmless but creates a false impression that `null` is reachable post-equality-check.

**Suggestion:** Replace with a comment: `// computeLineDiff returns non-null for distinct strings (no-hunk case requires identical input)`.

**6. nit · blocking:false · correctness · `src/cli/commands/interactive/tool-lane-render.ts:92–104` · ref:22486ffa · diff-context**

`previewDiff` is set on `ToolEntryFields` but never cleared in `addResult`. It persists on the entry after the result arrives. All current rendering paths correctly gate on `!entry.result` before rendering `previewDiff`, so there is no current leakage. This is a defensive note for future maintainers extending the render paths.

**Suggestion:** In `addResult` (tool-lane.ts:~181), add `entry.previewDiff = undefined;` after setting `entry.result = chunk` as a belt-and-suspenders guard.

---

## What was not checked
- Citations verified against branch HEAD `22486ffa907b91df6b91fd81a0efa514f85d898f`.
- Stated intent: PR #1366 title+body — spec-compliance assessed (both major items met; partial gap in NESTING branch noted as Finding 1).
- Did not run the test suite; CI reports green on all platforms.
- Did not review `compactDiffView` internals beyond confirming `sanitizeDiffText` strips escape sequences (ANSI injection mitigation confirmed).
- Did not review Telegram or daemon surfaces (ref threading is gated by `if (ctx.addPreviewDiffRef ?)` so non-REPL surfaces are correctly excluded).

griffinwork40 (commented):

---

**PR:** feat(render): pre-execution diff preview for edit_file
**Reviewed ref:** `22486ffa` | 20 files | ~480 lines | regime: full

## Decision: DO NOT MERGE — 3 medium blocking (2 correctness/spec-compliance, 1 security); 4 low, 2 nit

---

### Findings

**1. medium · blocking · security · `src/cli/_lib/stream-renderer.ts:326–331, 543–584` · ref:`22486ffa`**
Use-after-dispose on TTY compositor. `arm()` installs a live closure over `this.toolLane` into `addPreviewDiffRef.current`. `dispose()` never resets `addPreviewDiffRef.current` to the no-op. If a hook fires between `dispose()` and the next `arm()`, it calls into a torn-down compositor via the stale closure.
- *fix:* In `dispose()`, add `this.addPreviewDiffRef?.current = () => {};`

**2. medium · blocking · correctness/spec-compliance · `src/cli/render/preview-diff.ts:54` / `src/cli/commands/interactive/tool-lane.ts:577–582` · ref:`22486ffa`**
`preview-diff.ts` exports `previewDiff()` which prepends the `'⟳ Proposed'` label and delegates to `compactDiffView`. However, `tool-lane.ts` never imports or calls `previewDiff()` — it calls `formatDiffBlock(entry.previewDiff, 'overlay', '    ')` directly. The `'⟳ Proposed'` label stated in the PR body (behavior #4) never reaches the terminal. `previewDiff()` is dead production code.
- *fix:* Replace the `formatDiffBlock` call at `tool-lane.ts:577–582` with `previewDiff()`, or remove `preview-diff.ts` if the label is intentionally omitted.

**3. medium · blocking · correctness/spec-compliance · `src/cli/commands/interactive/tool-lane.ts:393, 577–583` · ref:`22486ffa`**
Preview renders only for flat-root entries (no `agentContext`). Any `edit_file` call issued from within a skill or tool group (`addStartWithAgentContext`) has `previewDiff` set by the hook, but `renderOverlayChildren` has no awareness of `previewDiff` — the preview is silently discarded. Stated behavior #1 ("fires on every `edit_file` call") is partially unmet: the hook fires correctly but the render does not follow for nested entries.
- *fix:* Surface `previewDiff` in `renderOverlayChildren` for child entries, or explicitly document root-only as intentional.

**4. low · security · `src/agent/tools/hooks/edit-preview-hook.ts:51–55` · ref:`22486ffa`**
No size bound on `old_string`/`new_string` before passing to `computeLineDiff`. The `MAX_DIFF_CELLS` guard bails cleanly, but string splitting happens first. Transient main-thread latency on very large edits; bounded by existing write-handler limits, no data loss.
- *suggestion:* Early-out if `oldStr.length + newStr.length > 512KB`.

**5. low · correctness/perf · `src/cli/commands/interactive/tool-lane.ts:179–184, 209–212` · ref:`22486ffa`**
`addResult()` sets `entry.result` but never deletes `entry.previewDiff`. The `DiffPayload` accumulates in the entry map until `flush()`. Minor GC pressure in high-frequency refactor turns.
- *suggestion:* Add `delete entry.previewDiff;` in `addResult()`.

**6. low · test-coverage · `src/cli/commands/interactive/tool-lane.test.ts:3057–3065` · ref:`22486ffa`**
Test (b) asserts absence of hunk header after result but does not positively assert the replacement outcome line. Weak negative-only assertion.
- *suggestion:* Add `expect(overlay).toContain('Replaced 1 occurrence')`.

**7. low · test-coverage · `src/agent/hooks/config-bridge.test.ts` · ref:`22486ffa` · confidence:medium**
Count assertions (3→4, 4→5) are correct but fragile — a property rename or removal passes if the count happens to be preserved.
- *suggestion:* Supplement with `expect(bundle).toHaveProperty('addPreviewDiffRef')`.

**8. nit · api-compat · `src/agent/hooks.ts:225–227` · ref:`22486ffa`**
`toolUseId` added only to `PreToolUseContext`, not `PostToolUseContext`. No breaking change; design gap for future correlation.

**9. nit · api-compat · `src/cli/commands/interactive/shared.ts:534,774` / `src/cli/_lib/stream-renderer-options.ts` · ref:`22486ffa`**
Inline ref type `{ current: (toolUseId: string, diff: DiffPayload) => void }` repeated 3× with no named export. Signature change requires three independent updates.
- *suggestion:* Export a named `PreviewDiffRef` type from `edit-preview-hook.ts`.

---

### Spec-Compliance

| # | Stated behavior | Status |
|---|---|---|
| 1 | PreToolUse hook fires on every `edit_file` call | ⚠️ Partial — hook fires; render dropped for nested entries [3] |
| 2 | Computes diff from `old_string`/`new_string` (no file read) | ✅ Met |
| 3 | Delivers via mutable callback ref bridging agent/CLI layers | ✅ Met |
| 4 | Overlay renders diff with `Proposed` label | ❌ Unmet — `previewDiff()` never called [2] |
| 5 | Preview replaced by normal outcome on result | ✅ Met |

### What was not checked
- Citations verified inline against branch HEAD `22486ffa907b91df6b91fd81a0efa514f85d898f`.
- Stated intent: PR #1366 title+body — spec-compliance assessed.
- Did not run the test suite; CI reports green on all platforms.
- Runtime overlay rendering was assessed via static analysis only — no headless terminal session exercised.
- `.filesize-baseline.json` 160-line mechanical compaction was not audited for correctness.
- E2E turn lifecycle (`arm → hook → lane → dispose`) not traced under live I/O; finding [1] is a static-analysis inference.

PR #1375, opened 2026-08-30
fix(verdict-card): strip orphaned trailing ** from parsed field values

griffinwork40:
## Problem

The verdict card intermittently renders `**` as the visible text for `changed` and `deferred` fields instead of showing their actual content or being omitted.

screenshot

## Root cause

Models sometimes wrap an entire bullet in one bold span (`- **What changed: some delta**`) or emit an empty bold marker (`- **What changed:** **`). After `extractBullets` splits on the colon, the trailing `**` is stranded on the value side:

| Model output | Parsed value |
|---|---|
| `- **What changed: delta here**` | `delta here**` |
| `- **What changed:** **` | `**` |
| `- **What changed: **` | `**` |

The existing strip logic only handled a **leading** orphaned `**` (from the `**Label:** value` pattern). The trailing closer was never stripped.

## Fix

Two layers, defense-in-depth:

1. **`terminal-state.ts`**: Strip trailing orphaned `**/`__` closers from parsed values. Safe because a trailing `**` at end-of-value is never a glob (globs always have a path component after `**`, e.g. `**/*.ts`). Also treats empty-after-strip values as `undefined` so stripped-to-empty fields don't produce phantom entries.

2. **`verdict-card.ts`**: `collectRows` now skips values that are purely residual markup noise (`**`, `****`, `__`, `____`) as a fallback guard, so even if a new pattern slips through the parser, it won't render.

## Tests

- 3 new tests in `terminal-state.test.ts` covering all three model output patterns
- 1 new test in `verdict-card.test.ts` for the noise-filter guard
- All 89 existing tests continue to pass
- Glob preservation (`src/**/*.ts`) and identifier preservation (`__init__`) verified safe

### DIFF
--- a/src/cli/commands/interactive/terminal-state.test.ts
+++ b/src/cli/commands/interactive/terminal-state.test.ts
@@ -165,6 +165,35 @@ describe('parseTerminalState — bullet field mapping', () => {
     expect(v?.evidence).not.toContain('**');
   });
 
+  // Regression: models wrap the entire bullet in one bold span:
+  //   `- **What changed: some delta**`
+  // After splitting on the colon, the trailing `**` is stranded on the value.
+  it('strips trailing orphaned bold closer from values (**Label: value** form)', () => {
+    const text = `prose\n\nDone\n- **What was done: built the parser**\n- **Evidence: see tests/parse.test.ts**\n- **What changed: terminal verdicts parsed**\n- **Deferred: integrate with renderer**`;
+    const v = parseTerminalState(text);
+    expect(v?.whatWasDone).toBe('built the parser');
+    expect(v?.evidence).toBe('see tests/parse.test.ts');
+    expect(v?.whatChanged).toBe('terminal verdicts parsed');
+    expect(v?.deferred).toBe('integrate with renderer');
+    // No orphaned `**` should survive in any value.
+    expect(v?.whatWasDone).not.toMatch(/\*\*$/);
+    expect(v?.evidence).not.toMatch(/\*\*$/);
+    expect(v?.whatChanged).not.toMatch(/\*\*$/);
+    expect(v?.deferred).not.toMatch(/\*\*$/);
+  });
+
+  // Edge case: model emits `- **Label:** **` (bold label, bare orphaned value).
+  // The parser should yield an empty string, not a literal `**`.
+  it('strips bare orphaned ** from values that are only markup noise', () => {
+    const text = `prose\n\nDone\n- **What was done:** shipped it\n- **What changed:** **\n- **Deferred:** **`;
+    const v = parseTerminalState(text);
+    expect(v?.whatWasDone).toBe('shipped it');
+    // Bare `**` should be stripped to empty — these fields become undefined
+    // because the empty value is trimmed out.
+    expect(v?.whatChanged).toBeUndefined();
+    expect(v?.deferred).toBeUndefined();
+  });
+
   it('preserves balanced bold and globs/paths inside values', () => {
     const text = `prose\n\nDone\n- What was done: touched **all** of src/**/*.ts\n- Evidence: see __init__ wiring`;
     const v = parseTerminalState(text);
--- a/src/cli/commands/interactive/terminal-state.test.ts
+++ b/src/cli/commands/interactive/terminal-state.test.ts
@@ -202,6 +202,17 @@ describe('parseTerminalState — bullet field mapping', () => {
     expect(v?.whatWasDone).toBe('touched **all** of src/**/*.ts');
     expect(v?.evidence).toBe('see __init__ wiring');
   });
+
+  // Regression: trailing `**` strip must not corrupt balanced inline bold at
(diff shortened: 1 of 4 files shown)


griffinwork40:
Confirmed -- the defense-in-depth layer 2 does NOT catch this case. The unbalanced `**` would render in the card.

---

## Review: PR #1375 — `fix(verdict-card): strip orphaned trailing ** from parsed field values`

**Ref:** `cb4883bc` | **Regime:** light (hotfix, +69/-3, 4 files) | **Branch:** `afk/render-verdict-card-sometimes` -> `main`

---

### Decision: DO NOT MERGE — 1 medium blocking (correctness); 1 low non-blocking (test-coverage)

---

### Findings

**1. medium · blocking:true · confidence:high · correctness · `terminal-state.ts:224` · ref:cb4883bc · file-state [VERIFIED]**

The tail-strip regex `.replace(/(?:\*\*|__)$/, '')` is too broad: it strips any trailing `**`, not just orphaned closers. A bullet value ending with balanced inline bold (e.g. `- Evidence: see **foo.ts**`) becomes `see **foo.ts` -- an unbalanced bold marker that renders as literal `**` in the card. Layer 2 (the `collectRows` noise filter) does not catch this because the value is not purely markup noise.

```
evidence: 'see **foo.ts**'.replace(/(?:\*\*|__)$/, '') === 'see **foo.ts'
```

The occurrence is narrow (requires a model to emit inline bold at the exact end of a plain-bullet value rather than code spans), but the fix introduces a new rendering defect in that path.

**suggestion:** Either detect balanced pairs before stripping (count `**` occurrences; strip only when odd), or document the tradeoff in a code comment and pin the behavior with a test for `see **foo.ts**` so the intent is explicit. The Wave 1 suggestion of a lookbehind was tested and regresses the primary fix.

---

**2. low · blocking:false · confidence:medium · test-coverage · `terminal-state.test.ts` · ref:cb4883bc**

No test covers a value ending with balanced inline bold (`**word**`) to pin the behavior of Finding 1. The existing `preserves balanced bold` test (line 197) only covers mid-value balanced bold (`**all**` followed by more text).

**suggestion:** Add a test case for `- Evidence: see **foo.ts**` regardless of how Finding 1 is resolved -- it locks the intended behavior either way.

---

### What was not checked

- Citations verified inline against branch HEAD `cb4883bc`.
- Stated intent: PR #1375 title+body -- spec-compliance assessed. Intent fully met; no scope creep.
- Did not audit consumers of `TerminalState` beyond `verdict-card.ts` and `verdict-ledger.ts`.
- Did not audit the `__` (underscore) path of the same regex -- same structural issue applies if a value ends with `__word__`.

**Done**
- What was done: Completed light-regime code review of PR #1375 across all six dimensions (security, api-compat, correctness, spec-compliance, test-coverage, perf-observability).
- Evidence: Review verdict above with one blocking medium finding (verified via runtime execution) and one non-blocking low finding.
- What changed: none (read-only review).
- Deferred: none.

---

griffinwork40:
## Review: PR #1375 — `fix(verdict-card): strip orphaned trailing ** from parsed field values`

**Ref:** `88f174ad` | **Regime:** light (hotfix, +82/-6, 4 files) | **Branch:** `afk/render-verdict-card-sometimes` -> `main`

---

### Decision: MERGE — 0 blocking (2 low)

---

### Findings

**1. low · blocking:false · confidence:high · correctness · `terminal-state.ts:220-226` · ref:88f174ad · diff-context**

The odd/even marker-count heuristic correctly handles all stated failure modes and the balanced-bold case (PR review finding 1 from the prior review is now resolved). The heuristic has a theoretical edge case: a multi-bold value with an odd total marker count where the final `**` is a legitimate closer paired with a stranded opener elsewhere (e.g. `**a b **c**`). This is an acceptable approximation for a hotfix — the input domain (model-emitted terminal-state bullets) is narrow enough that such values are unreachable in practice.

**suggestion:** A comment noting this approximation already exists implicitly in the "even count means every opener has a closer" wording. No action needed.

---

**2. low · blocking:false · confidence:medium · test-coverage · `terminal-state.test.ts` · ref:88f174ad · file-state**

The `__` branch of the tail-strip regex is exercised only by the mid-value `__init__` test (line 198). No test covers a value ending in `__` to exercise the underscore path of the odd/even heuristic.

**suggestion:** Add a case like `- Evidence: see __init__` (value ends in `__`) to mirror the `**` path coverage. Low priority — the `__` regex branch is structurally identical to `**`.

---

### What was not checked

- Citations verified inline against branch HEAD `88f174ad`.
- Stated intent: PR #1375 title+body — spec-compliance assessed. All three stated fix vectors present and tested.
- CI failures (`setPlanExitQueueCheck is not a function` in `repl-loop-wiring.test.ts`) are pre-existing on `main` — not introduced by this PR.
- Did not audit consumers of `TerminalState` beyond `verdict-card.ts` and `verdict-ledger.ts`.
- Did not audit the Codex bot's suggestion (lookbehind approach) — PR description states it was tested and regresses the primary fix.

---

PR #1405, opened 2026-08-31
fix(repl): deduplicate terminal-state rendering -- verdict card only

griffinwork40:
## Problem

The Done/Blocked/Asking/Interrupted terminal state renders **twice** in the REPL:

1. As prose in the markdown stream (the bullet list the model emits)
2. As the structured verdict card (`╭─ ✓ Done ─╮` box) parsed from that same prose

The card was documented as "additive," but on short turns the duplication is pure noise -- the same four fields appear verbatim in both renderings.

## Fix

Strip the terminal-state prose block from the markdown renderer's pending buffer **before** `dispose()` flushes it to scrollback. The verdict card becomes the sole visible rendering.

### Why this works

- The terminal-state block is always the last thing the model emits, with no trailing `\n\n`, so it stays in the markdown renderer's **pending buffer** (never committed during streaming).
- `responseText` -- which `parseTerminalState()` reads to build the verdict card -- accumulates independently from stream deltas and is unaffected by the strip.
- **Fallback safety**: if the heading is not found in the pending buffer (e.g., it was already committed during streaming due to an unusual `\n\n` placement, or format mismatch), no strip occurs and the prose renders as before.

### Changes

| File | What |
|------|------|
| `terminal-state.ts` | New `findTerminalStateHeadingOffset()` -- locates the character offset of the terminal-state heading in a text block. Reuses the existing `lineToKind()` parser. |
| `markdown-stream.ts` | New `stripPendingFrom(offset)` -- removes content from offset onward in the pending buffer without committing it. |
| `stream-renderer.ts` | New `stripPendingTerminalState()` and `getPendingBuffer()` -- public API that delegates to the markdown renderer. |
| `turn-handler.ts` | 4-line strip call between stream completion and `disposeRendererOnce()`. Budget-neutral: hoisted a duplicate `const write` and inlined an `if` guard to offset the addition. |

### Tests

- 10 new tests for `findTerminalStateHeadingOffset()` (offset correctness, all four terminal kinds, edge cases)
- 5 new tests for `stripPendingFrom()` (strip behavior, out-of-bounds, flush-after-strip, content preservation)
- All 6,384 existing tests pass

### DIFF
--- a/src/cli/_lib/stream-renderer.ts
+++ b/src/cli/_lib/stream-renderer.ts
@@ -457,6 +457,37 @@ export class StreamRenderer {
     }
   }
 
+  /**
+   * Strip the terminal-state prose block (Done/Blocked/Asking/Interrupted)
+   * from the markdown renderer's pending buffer so the verdict card is the
+   * sole visible rendering. Call BEFORE dispose — dispose flushes the pending
+   * buffer to scrollback.
+   *
+   * `headingOffset` is the character offset within the pending buffer where
+   * the terminal-state heading starts (from `findTerminalStateHeadingOffset`
+   * applied to the buffer, NOT to `responseText`).
+   *
+   * Returns `true` if the strip succeeded. No-op when the markdown renderer
+   * is not alive or the offset is out of bounds.
+   */
+  stripPendingTerminalState(headingOffset: number): boolean {
+    if (this.disposed) return false;
+    const md = this.streamingMarkdownRef.current;
+    if (!md) return false;
+    return md.stripPendingFrom(headingOffset);
+  }
+
+  /**
+   * Read the raw pending (uncommitted) buffer from the orchestrator's
+   * markdown renderer. Returns '' when disposed or no renderer is alive.
+   * Used by the turn handler to locate the terminal-state heading within
+   * the buffer rather than the full responseText.
+   */
+  getPendingBuffer(): string {
+    if (this.disposed) return '';
+    return this.streamingMarkdownRef.current?.getPendingBuffer() ?? '';
+  }
+
   /** Signal first streaming content — clears the TTFB waiting indicator. Idempotent. */
   notifyFirstContent(): void {
     if (this.disposed) return;
--- a/src/cli/commands/interactive/terminal-state.test.ts
+++ b/src/cli/commands/interactive/terminal-state.test.ts
@@ -9,7 +9,7 @@
  */
 
 import { describe, it, expect } from 'vitest';
-import { parseTerminalState } from './terminal-state.js';
+import { parseTerminalState, findTerminalStateHeadingOffset } from './terminal-state.js';
 
 describe('parseTerminalState — recognition', () => {
   it('returns null on empty input', () => {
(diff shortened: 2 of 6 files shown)


griffinwork40:
Good synthesis. The review is clean -- no blocking findings, merge recommended. Let me present the final report.

---

## PR #1405 Review: `fix(repl): deduplicate terminal-state rendering -- verdict card only`

**Branch:** `afk/agent-afk-behavior-check` → `main` | **Ref:** `8b9afb17` | **Regime:** light (hotfix, +232/-11, 6 files)

### Summary

Eliminates double-rendering of terminal-state prose (Done/Blocked/Asking/Interrupted) in the REPL. Previously, the terminal state rendered twice: once as streamed markdown prose and again as the structured verdict card. The fix strips the terminal-state prose from the markdown renderer's pending buffer *before* `dispose()` flushes it to scrollback, so the verdict card becomes the sole visible rendering. Adds `findTerminalStateHeadingOffset` (tail-anchored heading locator reusing `lineToKind`), `stripPendingFrom` (pending-buffer truncation), and thin delegation wrappers on `StreamRenderer`. Also hoists a duplicate `writeNotice` binding and simplifies an `onTurnComplete` guard. 15 new tests cover the new functions.

### Findings

| # | Sev | Block | Dim | Location | Finding |
|---|---|---|---|---|---|
| **C3** | medium | no | correctness | `turn-handler.ts:583-588` / `markdown-stream.ts:385` | Strip silently no-ops when the model emits trailing `\n\n` after the terminal-state block (block commits to scrollback before strip runs). Double-render survives for this variant. Author documented the assumption in JSDoc. |
| **T1** | low | no | test-coverage | `turn-handler.test.ts` | No integration test for the strip call path through the turn handler -- the production call site is invisible to CI. |
| **T2** | low | no | test-coverage | `markdown-stream.test.ts` | `stripPendingFrom` tests use a non-TTY renderer; TTY overlay-sync path not exercised. |
| **C1** | low | no | correctness | `markdown-stream.ts:390-394` | `stripPendingFrom` mutates buffer without calling `syncPendingOverlay()`, unlike `discardPending()`. Not a visible defect (buffer is already correct for any repaint, and `flush()` clears the overlay in Phase 3), but a style inconsistency. |
| **C2** | low | no | correctness | `markdown-stream.ts:391` | `>=` guard rejects `offset === buffer.length` -- unreachable given upstream preconditions. |
| **C4** | low | no | perf | `terminal-state.ts:100-106` | O(n) offset computation via inner loop; negligible (once per turn, small buffer). |
| **A1** | nit | no | api-compat | `markdown-stream.ts:348-353` | `getPendingBuffer()` JSDoc still reads "(for testing)" despite production usage via `StreamRenderer.getPendingBuffer()`. |

### Decision: MERGE -- 0 blocking (1 medium, 5 low, 1 nit)

The single medium (C3) is a *known, documented* gap in a narrow model-output variant (trailing `\n\n`), not a regression -- the double-render already occurred there before this PR. The fix narrows the failure surface without widening it. C3 and T1 are reasonable follow-up candidates.

### What was not checked

- Citations verified inline against branch HEAD `8b9afb17`. Zero fabricated, zero diff-only.
- Stated intent: PR title+body -- spec-compliance assessed. Change accomplishes stated intent for the common case; C3 documents the one reachable variant where intent is not fully achieved.
- No runtime execution was performed -- double-render elimination assessed by static code reading only.
- No instrumented trace of actual repaint-timer interleaving in a live TTY session.
- Trailing `\n\n` prevalence in model output was not corpus-measured.

**Done**
- What was done: Completed a light-regime code review of PR #1405 with Wave 1 (single research-agent), inline Wave 1.5 citation verification, and Wave 2 synthesis.
- Evidence: Review report above with 7 findings (0 blocking), merge recommendation, and epistemic scope disclosure. All citations verified against ref `8b9afb17`.
- Deferred: None.

---

Issue #1407, opened 2026-08-31
Terminal-state dedup no-ops when model emits trailing \n\n after verdict block


griffinwork40:
## Problem

The terminal-state dedup fix (#1405) strips the Done/Blocked/Asking/Interrupted prose from the markdown renderer's **pending buffer** before `dispose()` flushes it. This works on the nominal path because the terminal-state block is the last thing the model emits, with no trailing `\n\n`, so it stays pending.

However, if the model emits the terminal-state block followed by a trailing `\n\n`, the markdown renderer's `push()` detects a block boundary and **commits** the block to scrollback during streaming -- before the strip runs. At that point:

1. `renderer.getPendingBuffer()` returns `''`
2. `findTerminalStateHeadingOffset('')` returns `-1`
3. The strip is skipped
4. The committed prose remains in scrollback
5. The verdict card renders on top of it -- double-render survives

The gap is documented in the `stripPendingFrom` JSDoc ("the terminal-state block is expected to be the last block in the response, with no trailing `\n\n`, so it stays pending") but there is no fallback path for the committed case.

## Prevalence

LLMs commonly append trailing blank lines after structured output. The exact frequency against this codebase's system prompt is unmeasured, but the variant is reachable in production.

## Possible approaches

- **Committed-block eviction**: after the strip no-ops (offset === -1), check `responseText` for a terminal-state heading and suppress the verdict card if the prose was already committed (accept the prose rendering as the sole output).
- **Post-commit scrub**: track whether a terminal-state block was committed during streaming and suppress it retroactively in the `dispose()` flow -- harder, since committed content is append-only in the compositor.
- **Verdict-card-only suppression**: if `parseTerminalState(responseText)` succeeds but the pending buffer had nothing to strip, skip the verdict card rendering entirely to avoid duplication (simpler, but loses the structured card).

## Context

- PR: #1405
- Review finding C3
- Files: `src/cli/commands/interactive/turn-handler.ts`, `src/cli/markdown-stream.ts`

griffinwork40:
Already fixed in 1d0894cb (PR #1405). The dedup logic in turn-handler.ts correctly handles the trailing newline case. Closing.

PR #1408, opened 2026-08-31
fix(repl): handle committed terminal-state block in dedup (#1407)

griffinwork40:
## Problem

The terminal-state dedup fix (#1405) strips the Done/Blocked/Asking/Interrupted prose from the markdown renderer's pending buffer before `dispose()` flushes it. But if the model emits trailing `\n\n` after the terminal-state block, the renderer's `push()` detects a block boundary and **commits** the block to scrollback during streaming — before the strip runs.

Then:
- `getPendingBuffer()` returns `''`
- `findTerminalStateHeadingOffset('')` returns `-1`
- The strip is skipped
- The verdict card renders on top of the committed prose → **double render**

## Fix

In `turn-handler.ts`, after the `if (off >= 0)` strip path, add an `else` branch: if `parseTerminalState(responseText)` succeeds but `off === -1` (pending buffer was empty), the terminal-state block was already committed to scrollback by the renderer. Set `suppressVerdictCard = true` to skip `renderVerdictCard`. The committed prose is the sole output.

Also fixes a pre-existing lint error introduced by other changes in this worktree: `turn-handler.footer.ts` imported `costTokenLine` (unused) and dropped `formatCost` from `format-utils` while still calling it at line 77.

## Files changed

- `src/cli/commands/interactive/turn-handler.ts` — `suppressVerdictCard` flag + else branch
- `src/cli/commands/interactive/turn-handler.footer.ts` — lint fix (restore `formatCost` import)
- `src/cli/commands/interactive/turn-handler.test.ts` — new test

## Test coverage

New test in `runTurn — terminal-state dedup: trailing \n\n committed-block case (#1407)`:

- Streams a content chunk ending in `\n\n` containing a `Done` terminal-state block
- Verifies that no `╮` border glyph (the verdict card top-border marker) appears in `completionWriter` output
- Confirms the verdict card is suppressed when the block was already committed to scrollback

All 78 tests in `turn-handler.test.ts` pass.


griffinwork40 (commented):

---

## Prior automated rounds

None — this is the first automated review on this PR.

---

## Severity arithmetic

0 critical · 0 high · 1 medium (advisory, non-blocking per sweep floor) · 3 low (advisory) · 1 nit (advisory)

**Decision: MERGE (with follow-ups) — 0 critical, 0 high → blocking floor not reached; 1 medium advisory, 3 low advisory, 1 nit advisory.**

CI: all green (Lint & Build, Test ubuntu/macos, Docs Site, Publish Bundle Smoke, Vercel — all SUCCESS).

---

## Findings

### MEDIUM · advisory (non-blocking) · correctness · `src/cli/commands/interactive/turn-handler.ts:598–604`

`else if (parseTerminalState(responseText) !== null)` sets `suppressVerdictCard = true` when `off === -1`, but `off === -1` does not imply the pending buffer is empty — it means no terminal-state heading was found in the buffer. If a model emits `"Done\n- bullets\n\ntrailing remark"`, the `\n\n` commits the Done block to scrollback while `'trailing remark'` stays in the pending buffer; `off === -1` (no heading in `'trailing remark'`); `parseTerminalState(responseText)` succeeds on the full response; `suppressVerdictCard = true` is incorrectly set; the verdict card is suppressed while trailing prose is still in the buffer.

```ts
} else if (parseTerminalState(responseText) !== null) {
  suppressVerdictCard = true;  // fires even when pending buffer is non-empty
}
```

Non-blocking: impact is bounded — prose commits to scrollback, no data loss; reachable only on model drift from the system-prompt requirement that terminal state is the last output (rare). Pre-fix behavior was a double-render, which is arguably worse UX. Suggest a follow-up issue rather than a fix commit.

**Suggested fix:** `else if (renderer.getPendingBuffer().trim() === '' && parseTerminalState(responseText) !== null)`

---

### LOW · advisory (non-blocking) · correctness · `src/cli/commands/interactive/turn-handler.test.ts:2663`

The new test stub does not mock `getPendingBuffer` — it relies on the real `StreamRenderer` returning `''` for the trailing-`\n\n` content via block-boundary slicing. The behavior is correct in practice but the test does not control or assert this invariant explicitly.

**Suggested follow-up:** mock `getPendingBuffer` on the stub to make the reliance explicit.

---

### LOW · advisory (non-blocking) · test-coverage · `src/cli/commands/interactive/turn-handler.test.ts`

New test covers the committed-block path (buffer empty + `parseTerminalState` succeeds → suppression). Missing: (1) the false-positive case from the medium finding above (non-empty buffer + `off === -1` + `parseTerminalState` succeeds → incorrect suppression), (2) interrupt/soft-stop guard interaction with `suppressVerdictCard`, (3) multi-chunk streaming.

**Suggested follow-up:** add a test where `responseText` contains a terminal-state block followed by trailing prose.

---

### LOW · advisory (non-blocking) · test-coverage · `src/cli/commands/interactive/turn-handler.test.ts:2645–2650`

The test asserts `writerCalls` for the `╮` glyph but never asserts on `commitAboveCalls`. If the verdict card were routed through `commitAbove` instead of `completionWriter`, the assertion would not catch it.

**Suggested follow-up:** add an assertion on `commitAboveCalls` alongside the `writerCalls` check.

---

### NIT · advisory (non-blocking) · perf-observability · `src/cli/commands/interactive/turn-handler.ts:598–602`

`parseTerminalState(responseText)` is called twice in the `suppressVerdictCard = true` path — once in the `else if` guard and once at the verdict-card render site. Negligible cost (tail-anchored, O(40 lines)), but the result could be cached in a local variable to make the double-parse explicit and eliminate any future confusion about side effects.

---

## Spec-compliance: PASS

The core fix (committed-block detection + verdict-card suppression) is correctly implemented and matches the stated intent. The lint fix in `turn-handler.footer.ts` is confirmed clean at the reviewed ref. The `if (doneFired && !softStopRequested && !pauseInterruptRequested)` guard wrapping is logically correct but is not mentioned in the PR body — worth a one-line note.

---

## What was not checked

- Citations verified against branch HEAD `57a1ee53ea7b78f570768b8d761af33e12f3f9d4` via `git show`.
- Stated intent: PR #1408 title + body — spec-compliance assessed.
- Did not run the test suite locally; relied on CI results (all green).
- Did not review `turn-handler.footer.ts` line-by-line in diff isolation — confirmed clean state at reviewed ref.
- Did not verify `parseTerminalState` side-effect behavior under double-invocation (assumed pure per module JSDoc).
- Did not review Telegram push path or other `onTerminalState` consumers — those are unchanged by this PR.

PR #1424, opened 2026-09-01
feat(tools): wait_for tool with poll-based condition waiting

### LINKED ISSUE #1417: feat(tools): declarative wait_for primitive -- poll-based condition waits without consuming model turns
## Problem

The agent has no declarative way to wait for an external condition (file appearance, process exit, CI check completion, URL becoming reachable, deployment finishing) without consuming model turns or shelling out to `bash: sleep N && check`.

The only existing `wait_for` is a browser-specific DOM element visibility poll inside `browser_act` (`src/browser/playwright/index.ts:320-321`, `src/agent/tools/schemas.ts:1170,1179`). Everything else requires the agent to loop: call bash, check the condition, sleep, repeat -- burning model turns and tokens on each iteration.

### User impact

- Waiting for CI to pass after a push requires a polling loop: `bash: sleep 30 && gh run list ...`, check result, repeat. Each iteration costs a model turn (~$0.01-0.05 depending on context size).
- Waiting for a dev server to start requires similar polling. The agent often gives up too early or wastes turns on a server that takes 30 seconds to bind.
- Background jobs (`mode: "background"` subagents) auto-deliver results, but the agent cannot express "wait for this file to appear" or "wait for this process to exit" with the same mechanism.
- Daemon tasks that need to wait for an external event (webhook, file drop, API readiness) must busy-poll in bash, which is wasteful and fragile.

## Verified current behavior

| Component | File | Behavior |
|-----------|------|----------|
| browser_act wait_for | `src/browser/playwright/index.ts:320-321`, `src/agent/tools/schemas.ts:1170,1179` | Blocks until a DOM element becomes visible. Browser-specific, not generalizable |
| Background job delivery | `src/agent/tools/schemas.ts:383` | Background results auto-injected as `<background-subagent-result>` blocks. No poll needed, but limited to subagent results |
| Daemon scheduler poll | `src/agent/daemon/scheduler.ts:232,340` | `setInterval`-based queue polling. Internal to the daemon, not agent-accessible |
| Provider retry/backoff | `src/agent/providers/shared/sleep-with-abort.ts`, `src/agent/providers/anthropic-direct/query/retry-layer.ts` | Internal retry infrastructure. Not exposed to the agent |
| Bash tool timeout | `src/agent/tools/schemas.ts` (bash schema) | `timeout_ms` caps at 600,000ms (10 minutes). The only agent-level wait mechanism beyond browser_act |

## Proposed minimum behavior

A `wait_for` tool that blocks on an external condition without consuming model turns:

**Condition types:**
- `url`: wait for a URL to return a success status (2xx). Options: method (HEAD/GET), expected status, expected body substring
- `file`: wait for a file to appear or be modified. Options: path, existence vs. content match, mtime threshold
- `process`: wait for a process (by PID) to exit. Options: expected exit code
- `ci_check`: wait for a GitHub Actions check/workflow to complete. Options: repo, ref, check name, expected conclusion
- `command`: wait for a shell command to return exit code 0 (or a specified code). Subsumes the others for arbitrary conditions
- `background_job`: wait for a specific background subagent job to complete (by job ID)

**Common options:**
- `timeout_ms`: maximum wait time (default: 120,000ms, max: 600,000ms)
- `poll_interval_ms`: how often to check (default: 5,000ms, min: 1,000ms)
- `backoff`: `none` | `linear` | `exponential` (default: `none` for short waits, `exponential` for long)
- `cancel_signal`: tied to the session's AbortSignal for cleanup on session end

**Return shape:**
- `status`: `succeeded` | `timed_out` | `cancelled` | `failed`
- `elapsed_ms`: how long the wait took
- `result`: condition-specific result (HTTP status, file stat, exit code, check conclusion)
- `attempts`: number of poll iterations

**Runtime behavior:**
- The tool handler runs a poll loop in Node (not in the model's turn loop). The model's turn is blocked on a single tool call, not iterated.
- The poll loop respects the session's AbortSignal -- if the session is cancelled, the wait is aborted.
- Each poll iteration is logged to the witness trace (at debug level, not every iteration) for observability.

## Daemon restart considerations

For short waits (< 10 minutes), daemon restart simply aborts the wait and the task fails. For long waits, the question is whether the wait state should survive restart:

- **Option A: No persistence.** Waits are process-scoped. Daemon restart aborts all waits. Simple, consistent with current bash timeout behavior.
- **Option B: Persist wait definitions.** The wait condition and timeout are written to disk. On restart, the daemon reschedules the wait. Requires integration with the durable task lifecycle.

Recommendation: start with Option A and upgrade to Option B once the durable task lifecycle lands.

- The poll loop must not leak resources on session abort. The AbortSignal integration is critical -- an orphaned poll loop after session end would waste CPU/network.
- `url` waits must go through the SSRF guard (`src/web/egress-guard.ts`) to prevent the agent from using wait_for as an SSRF vector (repeatedly probing internal hosts).
- `command` waits inherit bash tool security: the command runs in the same sandboxed bash context with the same restrictions.
- `process` waits by PID are safe (read-only `kill -0` check) but the agent should not be able to wait on arbitrary PIDs outside its process tree.
- Long poll intervals (> 60s) should log a "still waiting" heartbeat to prevent the operator from thinking the session is stuck.

- **bash: sleep N && check:** Works but costs a model turn per iteration. A 5-minute wait with 30-second polls = 10 model turns = 10x the cost of a single wait_for call.
- **Background subagent + auto-delivery:** For subagent results, this already works. But it doesn't generalize to arbitrary conditions.
- **bash: timeout + inotifywait/fswatch:** Works for file waits on Linux/macOS, but requires the tools to be installed and is platform-specific.
- **Custom MCP server:** Could implement a wait-for-condition server. Works but per-condition investment.

## Non-goals

- Event-driven reactive triggers (webhook receivers, filesystem watchers that invoke sessions). This is a poll-based wait, not an event subscription.
- Distributed coordination (waiting for a condition on a remote machine). The condition must be checkable from the local machine.
- Sub-second polling precision (minimum 1 second interval)
- Indefinite waits without timeout (all waits must have a finite timeout)

## Acceptance criteria

- [ ] Agent can block on a condition (URL, file, process, command, CI check) without consuming model turns
- [ ] The poll loop runs in Node, not in the model's turn loop
- [ ] Waits respect the session AbortSignal and clean up on session end
- [ ] URL waits pass through the SSRF guard
- [ ] Timeout, polling interval, and backoff are configurable
- [ ] The tool returns structured results (status, elapsed time, condition result, attempt count)
- [ ] Long waits emit periodic heartbeat logs for operator observability

- Related but not blocking: durable task lifecycle (for long waits that need to survive restart)
- Related: browser_act wait_for (`src/browser/playwright/index.ts:320-321`) -- precedent for the wait pattern, browser-specific
- Related: provider sleep-with-abort (`src/agent/providers/shared/sleep-with-abort.ts`) -- internal infrastructure that could be reused

## Priority

**P2** -- Meaningful quality-of-life improvement that reduces token waste on polling loops, but agents can work around it with bash loops today. Becomes higher priority for daemon-heavy workflows where polling loops are especially wasteful.

## Open design questions

1. Should condition types be extensible (plugin-contributed condition evaluators) or a fixed set?
2. Should `command` type reuse the bash tool's security context (readonly-bash restrictions, denylist) or have its own simpler policy?
3. Should the tool support compound conditions (wait for A AND B, wait for A OR B), or only single conditions?
4. Should the witness trace record every poll iteration (noisy but complete) or only transitions (condition met, timeout, error)?

griffinwork40:
## What this does

Adds a `wait_for` tool that blocks on an external condition without consuming model turns. The poll loop runs entirely in Node — the model issues one tool call and gets back a structured result when the condition is met, times out, or is cancelled.

## Condition types

| Type | What it waits on | Met when |
|------|-----------------|----------|
| `url` | HTTP/HTTPS endpoint | Returns expected status (default: any 2xx) |
| `file` | Filesystem path | File exists (optional `content_contains` match) |
| `process` | PID | Process has exited (`kill(pid, 0)` → `ESRCH`) |
| `command` | Shell command | Command exits 0 |

## Usage examples

```json
{ "type": "url", "url": "https://api.example.com/health", "timeout_ms": 60000 }
{ "type": "file", "path": "/tmp/build/output.zip", "poll_interval_ms": 2000 }
{ "type": "process", "pid": 12345, "timeout_ms": 300000, "backoff": "exponential" }
{ "type": "command", "command": "test -f /tmp/done.marker", "poll_interval_ms": 1000 }
```

## Design decisions

- **SSRF guard is mandatory for URL waits** — `checkEgressTarget()` runs eagerly before entering the poll loop
- **`sleepWithAbort()`** is the sleep primitive — loop respects `AbortSignal`, no orphaned loops
- **Timeout → `timed_out` (not error)** — the model can react and retry
- **Backoff strategies**: `none` (fixed), `linear` (+1s/attempt), `exponential` (2^n, cap 60s)
- **Heartbeat log** every 60s via `console.error`
- Schema split into `schemas.wait-for.ts` to satisfy the 350-code-line ratchet

## Acceptance criteria

- [x] URL wait succeeds when fetch returns 200
- [x] URL wait is rejected for private IPs (SSRF guard)
- [x] File wait succeeds when file appears
- [x] Command wait succeeds when command returns exit 0
- [x] Process wait succeeds when PID disappears
- [x] Timeout returns `timed_out` (not an error)
- [x] Abort signal returns `cancelled`
- [x] Backoff increases interval correctly (linear and exponential tested)
- [x] Invalid input returns error
- [x] `pnpm lint` passes
- [x] 42 tests, all pass
- [x] `pnpm audit:filesize:check` — no new violations from this PR

Closes #1417

### DIFF
--- a/src/agent/risk-classifier.ts
+++ b/src/agent/risk-classifier.ts
@@ -415,6 +415,36 @@ export function classifyRisk(
     return 'medium';
   }
 
+  // ---- wait_for ------------------------------------------------------------
+  // wait_for condition types differ significantly in risk:
+  //
+  //   url     — issues outbound HTTP; same risk profile as web_scrape ('medium').
+  //   file    — read-only fs.stat / readFile; 'safe'.
+  //   process — read-only process.kill(pid, 0); 'safe'.
+  //   command — runs a shell command via execSync. The command string is
+  //             classified using the same bash rule table (classifyBash) as the
+  //             bash tool, but that logic lives inside evaluateCommand so we
+  //             cannot call it here without the command string. We therefore
+  //             use the bash classifier directly when the type is 'command'.
+  //             For other types fall through to the type-specific rules above.
+  //
+  // A missing or unknown `type` field is treated conservatively as 'medium'.
+  if (tool === 'wait_for') {
+    const obj =
+      typeof input === 'object' && input !== null
+        ? (input as Record<string, unknown>)
+        : {};
+    const condType = typeof obj['type'] === 'string' ? obj['type'] : '';
+    if (condType === 'command') {
+      const cmd =
+        typeof obj['command'] === 'string' ? obj['command'] : '';
+      return classifyBash(cmd);
+    }
+    if (condType === 'file' || condType === 'process') return 'safe';
+    // url or unknown: treat as medium (outbound network).
+    return 'medium';
+  }
+
   // ---- unknown tool -------------------------------------------------------
   // Don't gate things we haven't classified — default to safe so the
   // classifier never blocks novel tools without an explicit rule.
--- a/src/agent/tool-category.ts
+++ b/src/agent/tool-category.ts
@@ -63,6 +63,8 @@ const READ_TOOLS = new Set([
 const WRITE_TOOLS = new Set([
   'Write', 'Edit', 'NotebookEdit', 'MultiEdit',
   'write_file', 'edit_file',
+  // patch_apply applies unified diffs to files on disk — a write operation.
+  'patch_apply',
   // memory-tools.ts — both mutate persistent state on disk (the fact
   // archive / HOT.md and ~/.afk/procedures/<name>.md respectively), so
   // they belong in the write bucket alongside file edits.
@@ -86,6 +88,8 @@ const WRITE_TOOLS = new Set([
 const SHELL_TOOLS = new Set([
   'Bash', 'BashOutput', 'KillBash',
   'bash',
+  'test_run',
+  'wait_for',
 ]);
 export const SUBAGENT_TOOLS = new Set([
   'Agent', 'Task',
--- a/src/agent/tools/handlers/index.test.ts
+++ b/src/agent/tools/handlers/index.test.ts
@@ -3,9 +3,9 @@ import { createBuiltinHandlers } from './index.js';
 import { BUILTIN_TOOL_NAMES } from '../schemas.js';
 
 describe('createBuiltinHandlers', () => {
-  it('returns a Map with all 22 built-in tools', () => {
+  it('returns a Map with all built-in tools', () => {
     const handlers = createBuiltinHandlers();
-    expect(handlers.size).toBe(26);
+    expect(handlers.size).toBe(27);
   });
 
   it('has an entry for every tool in BUILTIN_TOOL_NAMES', () => {
@@ -27,7 +27,7 @@ describe('createBuiltinHandlers', () => {
     // when permissionMode is undefined but cwd is set. Ensure we get a
     // valid map back and bash is still callable.
     const handlers = createBuiltinHandlers(undefined, '/tmp');
-    expect(handlers.size).toBe(26);
+    expect(handlers.size).toBe(27);
     expect(typeof handlers.get('bash')).toBe('function');
     expect(typeof handlers.get('grep')).toBe('function');
     expect(typeof handlers.get('glob')).toBe('function');
@@ -35,7 +35,7 @@ describe('createBuiltinHandlers', () => {
 
   it('cwd parameter: builds handler set with both permissionMode and cwd', () => {
     const handlers = createBuiltinHandlers('default', '/tmp');
-    expect(handlers.size).toBe(26);
+    expect(handlers.size).toBe(27);
     expect(typeof handlers.get('bash')).toBe('function');
   });
 
--- a/src/agent/tools/handlers/index.ts
+++ b/src/agent/tools/handlers/index.ts
@@ -38,6 +38,7 @@ import { browserObserveHandler } from './browser-observe.js';
 import { browserActHandler } from './browser-act.js';
 import { browserScreenshotHandler } from './browser-screenshot.js';
 import { browserCloseHandler } from './browser-close.js';
+import { waitForHandler } from './wait-for.js';
 
 /**
  * Build the built-in tool handler map for a session.
@@ -98,6 +99,7 @@ export function createBuiltinHandlers(
     ['config_get', configGetHandler],
     ['config_set', configSetHandler],
     ['ask_question', askQuestionHandler],
+    ['wait_for', waitForHandler],
     ['browser_open', browserOpenHandler],
     ['browser_observe', browserObserveHandler],
     ['browser_act', browserActHandler],
@@ -126,6 +128,7 @@ export {
   configGetHandler,
   configSetHandler,
   askQuestionHandler,
+  waitForHandler,
   browserOpenHandler,
   browserObserveHandler,
   browserActHandler,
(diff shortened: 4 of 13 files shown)


griffinwork40 (commented):

## PR #1424 — `wait_for` tool with poll-based condition waiting

**Ref:** `e42e4eb0` · **Decision: DO NOT MERGE** — 2 high, 4 medium blocking (3 correctness, 2 security, 1 test-coverage); 2 low non-blocking.

---

### 🔴 HIGH — Blocking

**[C-F1] Per-poll AbortSignal kills slow evaluators terminally** · `wait-for-poller.ts:96-98`
The per-poll timeout is capped at `poll_interval_ms`. An evaluator responding slower than the interval (e.g. 6s response, 5s interval) throws `AbortError` → permanent `status: 'failed'` instead of retry. Slow-but-healthy targets become terminal failures.
*Fix: Catch per-poll `AbortError` and treat as a miss (continue loop), not terminal failure.*

**[C-F2] `execSync` blocks the event loop up to 30s** · `wait-for-conditions.ts:183-190`
`execSync` is synchronous and receives no `AbortSignal`. Session teardown and all async I/O stall for the full duration. Perf regression on concurrent sessions.
*Fix: Replace with async `exec`/`spawn` wired to the abort signal, or reduce the hardcoded timeout.*

---

### 🟠 MEDIUM — Blocking

**[C-F3] GET response body leaked when `body_contains` is unset** · `wait-for-conditions.ts:101-105`
HEAD branch cancels the body; GET-without-`body_contains` does not. Leaks the socket/stream on every poll iteration.
*Fix: Add `else { await response.body?.cancel().catch(() => undefined); }` after the HEAD arm.*

**[S-F2] `cwd` passed to `execSync` without containment** · `wait-for-conditions.ts:evaluateCommand`
`cwd` derives from runtime context (same trust posture as bash handler), but this new exec surface doesn't inherit documented containment. No `resolveAndContain` call.
*Fix: Apply containment or document parity with bash.*

**[C-F4] Missing test coverage for blocking findings** · `wait-for-poller.test.ts`
No test for: (a) evaluator `AbortError` with live session signal → currently fails instead of retrying; (b) GET response body drain without `body_contains`.
*Fix: Add targeted tests for both paths.*

**[S-F4] File condition accepts arbitrary absolute paths** · `wait-for.ts` file resolution
Absolute paths (e.g. `/etc/passwd`) bypass containment. Lower risk than bash (read-only stat/readFile), but the model can probe arbitrary filesystem existence.
*Fix: Route through `resolveAndContain` for read containment.*

---

### 🟡 LOW — Non-blocking

**[SC-F5]** Redundant `checkEgressTarget` alongside `guardedFetch` — `wait-for.ts:167-173`. Remove the eager check; let `guardedFetch` own egress validation.

**[S-F5]** Risk classifier returns `'medium'` for unknown `condType` — `risk-classifier.ts`. Should fail closed at `'high'`; handler validates first, so defense-in-depth only.

---

### What Was Not Checked
- Citations verified against branch HEAD `e42e4eb0`.
- Not checked: backoff degradation under high miss rates, redirect-chain SSRF edge cases, Windows path behavior, end-to-end session teardown under active polls.

PR #1425, opened 2026-09-01
feat(tools): patch_apply tool with atomic multi-file changes

### LINKED ISSUE #1419: feat(tools): atomic patch and dry-run primitive -- structured multi-file edits with validation and rollback
## Problem

The agent has no way to preview, validate, or atomically apply a structured multi-file patch. The current file-editing tools operate on individual files with no cross-file atomicity:

- `edit_file` has a TOCTOU mtime guard (good) but operates on a single file per call. A multi-file edit is a sequence of independent `edit_file` calls -- if the third call fails, the first two have already been applied with no rollback.
- `write_file` lacks even the mtime guard that `edit_file` has (`src/agent/tools/handlers/write-file.ts:101` notes the TOCTOU concern).
- Neither tool supports dry-run (preview the resulting diff before committing the change).

### User impact

- A multi-file refactor (rename a symbol across 5 files) is a sequence of independent edits. If `edit_file` fails on file 3 (multiple matches, stale content), files 1-2 are already changed. The codebase is in an inconsistent state with no automatic rollback.
- The agent cannot preview what a set of changes will look like before applying them. It must apply, then read the result, then potentially undo.
- Machine-generated patches (from a linter, from `git diff`, from a subagent's structured output) cannot be applied as a unit -- they must be decomposed into individual `edit_file` calls, losing atomicity and error context.

## Verified current behavior

| Component | File | Safety |
|-----------|------|--------|
| edit_file | `src/agent/tools/handlers/edit-file.ts` | TOCTOU mtime guard: `stat()` before read, `stat()` again before write, abort if mtime changed. Exact-match enforcement (fail on 0 or >1 matches). Single-file only |
| write_file | `src/agent/tools/handlers/write-file.ts:101` | No mtime guard. Creates parent directories. Single-file only |
| diff engine | `src/utils/diff.ts` | Full LCS-based line diff: `DiffPayload`, `DiffHunk`, `DiffLine`. MAX_DIFF_CELLS cap (4,000,000). Used by edit_file for display diffs, not for pre-apply validation |
| Write denylist | `src/agent/tools/handlers/write-denylist.ts` | Both edit_file and write_file check against a denylist of protected paths |
| Path containment | `src/agent/grant-manager.ts` | Grant-based path containment for reads and writes. Per-path approval with "Once" grants |

## Proposed minimum behavior

A `patch_apply` tool (or `multi_edit`) that accepts a structured set of file changes and applies them atomically:

**Input schema:**
```
{
  "changes": [
    {
      "path": "src/foo.ts",
      "expected_hash": "sha256:abc...",  // optional: verify file hasn't changed
      "edits": [
        { "old": "exact string", "new": "replacement string" },
        { "old": "another match", "new": "another replacement" }
      ]
    },
    {
      "path": "src/bar.ts",
      "expected_hash": "sha256:def...",
      "content": "full file content"  // write mode: replaces entire file
    }
  ],
  "dry_run": false  // if true, return the diff without applying
}
```

**Behavior:**

1. **Validate phase:** For each file, verify the `expected_hash` matches (if provided), verify all `old` strings match exactly once, check path containment and denylist. If any validation fails, report all errors and apply nothing.

2. **Dry-run phase (always, even when `dry_run: false`):** Compute the resulting diff using `src/utils/diff.ts` and include it in the response. The agent (or operator) can see what will change before it's applied.

3. **Apply phase (when `dry_run: false`):** Apply all changes atomically. "Atomic" means: write all changes to temp files, then rename all at once. If any rename fails, roll back by restoring from the originals (which were read in the validate phase).

4. **Result shape:**
- `status`: `applied` | `dry_run` | `validation_failed` | `partial_failure`
- `diff`: unified diff of all changes (machine-readable)
- `files_changed`: list of affected files with before/after hashes
- `errors`: list of per-file validation or application errors

## Why this adds value beyond edit_file, write_file, and bash

| Capability | edit_file | write_file | bash (git apply) | patch_apply |
|-----------|-----------|------------|-------------------|-------------|
| Single-file edit | Yes | Yes (overwrite) | Yes | Yes |
| Multi-file atomic | No | No | Yes (but unstructured) | Yes |
| Dry-run preview | No | No | Yes (`--check`) | Yes |
| Content hash verification | No (mtime only) | No | Yes (git context lines) | Yes |
| Conflict reporting | No (fail or succeed) | No | Yes (reject hunks) | Yes |
| Path containment | Yes | Yes | No (bypasses grant-manager) | Yes |
| Structured result | Yes (string diff) | No | No (exit code + stderr) | Yes (typed) |
| Rollback on partial failure | No | No | Yes (git handles it) | Yes |

The key advantage over `bash: git apply` is that `patch_apply` goes through the grant-manager path containment and write denylist, provides structured machine-readable results, and doesn't require the changes to be formatted as a unified diff (which models often get wrong).

- Atomicity is best-effort on filesystems without rename-across-directories guarantees. The rename-from-temp pattern is the standard approach and works on all major filesystems, but a power failure mid-rename sequence could leave partial state. This is the same risk as any multi-file tool.
- The validate-before-apply pattern prevents the "first 2 of 5 edits applied, then failure" scenario that plagues sequential `edit_file` calls.
- Path containment and denylist checks must run on every file in the patch, not just the first. A patch that includes a denylist-protected file must be entirely rejected, not partially applied.
- `expected_hash` is optional to avoid forcing the model to compute hashes. When omitted, the mtime guard from `edit_file` should be used as a fallback.

- **Sequential edit_file calls:** Works for most cases, but no atomicity or rollback. The agent must manually undo if a middle edit fails.
- **bash: git apply:** Works but bypasses path containment, has no structured result, and requires the model to generate unified diff format (error-prone).
- **bash: git stash + sequential edits + git stash pop on failure:** A workaround for rollback, but complex and fragile.
- **Worktree isolation:** `worktree` tool provides branch-level isolation for parallel work, but doesn't help with atomicity within a single branch.

## Non-goals

- Conflict resolution (merge conflicts are reported, not auto-resolved)
- Three-way merge (this is a patch tool, not a merge tool)
- Git integration (this operates on the filesystem, not on git objects)
- Binary file support (text files only)
- Patch generation from model reasoning (the model generates the structured input; the tool validates and applies)

## Acceptance criteria

- [ ] A structured set of file changes can be validated, previewed (dry-run), and applied atomically
- [ ] Validation failure on any file prevents all changes from being applied
- [ ] Content hash verification catches stale file state before edits
- [ ] All changes pass through the grant-manager and write denylist
- [ ] The response includes a machine-readable unified diff and per-file results
- [ ] Partial application failure triggers rollback of already-applied changes

- Reuses: `src/utils/diff.ts` (diff computation), `src/agent/tools/handlers/write-denylist.ts` (path protection), `src/agent/grant-manager.ts` (containment)
- No hard dependencies on other proposed primitives
- Related: edit_file TOCTOU guard (`src/agent/tools/handlers/edit-file.ts`) -- the mtime guard pattern should be reused

## Priority

**P3** -- Nice-to-have that reduces the risk of partial-failure states in multi-file edits. The current sequential edit_file pattern works adequately for most cases, and git provides rollback via `git checkout -- <file>`. Priority increases if agent-generated patches become a common workflow.

## Open design questions

1. Should the tool accept unified diff format as an alternative to the structured edits format? Models often produce diffs, and accepting them directly would avoid a translation step.
2. Should rollback be automatic on partial failure, or should the tool report the partial state and let the agent decide?
3. Should the tool support file creation and deletion in addition to edits? If so, rollback for creation = delete, rollback for deletion = restore from the pre-validate copy.
4. Should `expected_hash` cover the entire file or just the region being edited?

griffinwork40:
# feat(tools): `patch_apply` — atomic multi-file patch with dry-run

## What it does

`patch_apply` is a new built-in tool that validates and atomically applies structured multi-file changes. Unlike `edit_file` (one file, one replacement per call), `patch_apply` accepts an array of changes spanning multiple files and applies them together — or rejects the whole batch if any file fails validation.

**Input shape:**
```json
{
  "changes": [
    {
      "path": "src/foo.ts",
      "expected_hash": "sha256:<hex>",
      "edits": [{ "old": "exact string to replace", "new": "replacement" }]
    },
    {
      "path": "src/bar.ts",
      "content": "full replacement content"
    }
  ],
  "dry_run": false
}
```

Each file change can use either `edits` (sequential exact-match replacements) or `content` (full replacement) — never both.

## Implementation

Three new source files (under the 350-LOC ceiling each) plus a schema file:

| File | Role |
|------|------|
| `patch-validate.ts` | Per-file validation (containment, denylist, hash, edit uniqueness) — collects ALL errors before returning |
| `patch-apply-engine.ts` | Compute new content, diff, and write atomically via temp-file + rename |
| `patch-apply.ts` | `ToolHandler` wrapper — parses input, calls validate → apply |
| `schemas.patch-apply.ts` | Schema definition (extracted to keep `schemas.ts` under baseline) |

## Validate-then-apply pattern

1. **Validate all files first** — path containment (via write roots), write denylist, optional SHA-256 hash match, and edit exact-match-once enforcement. All errors across all files are collected before returning.
2. **Apply atomically** — each file is written to a `.patch-tmp-<random>` temp file in the same directory, then renamed to the target. On any rename failure, previously-renamed files are restored from the in-memory pre-validation snapshot.

## Dry-run behavior

When `dry_run: true`, the tool computes the unified diff for all changes and returns it without touching the filesystem.

## Hash verification

`expected_hash: "sha256:<hex>"` verifies the file's current SHA-256 before applying any changes. If the hash doesn't match, the entire patch is rejected.

## Tests

31 tests across three test files — all passing.

## Acceptance criteria

- [x] `patch_apply` tool registered in `builtinToolSchemas` and handler map
- [x] Path containment via `resolveAndContain` (write mode) on every file
- [x] Write denylist checked on every file via `assertNotDenylisted`
- [x] Hash format `sha256:<hex>` verified before any edit
- [x] Edit `old` must match exactly once (per-file, sequential)
- [x] `edits` and `content` mutually exclusive
- [x] All validation errors collected (not short-circuited)
- [x] Dry-run returns unified diff without writing
- [x] Atomic apply via temp-file + rename
- [x] Best-effort rollback on partial failure
- [x] All 31 tests pass
- [x] `pnpm lint` clean
- [x] `pnpm audit:filesize:check` — no new violations

Closes #1419

### DIFF
--- a/src/agent/risk-classifier.ts
+++ b/src/agent/risk-classifier.ts
@@ -322,6 +322,27 @@ export function classifyRisk(
     return classifyFilePath(filePath, ctx);
   }
 
+  // ---- patch_apply ---------------------------------------------------------
+  // patch_apply atomically writes multiple files; classify by the highest-risk
+  // path across the entire changes array. An empty changes array is safe.
+  if (tool === 'patch_apply') {
+    const obj =
+      typeof input === 'object' && input !== null
+        ? (input as Record<string, unknown>)
+        : {};
+    const changes = Array.isArray(obj['changes']) ? obj['changes'] : [];
+    let highest: RiskLevel = 'safe';
+    for (const change of changes) {
+      if (typeof change !== 'object' || change === null) continue;
+      const p = (change as Record<string, unknown>)['path'];
+      const filePath = typeof p === 'string' ? p : '';
+      const level = classifyFilePath(filePath, ctx);
+      if (level === 'high') return 'high';
+      if (level === 'medium') highest = 'medium';
+    }
+    return highest;
+  }
+
   // ---- read-class tools ---------------------------------------------------
   // Derive from tool-category taxonomy rather than a local hand-maintained list.
   // Previously this was a 5-entry if-chain that diverged from tool-category.ts
--- a/src/agent/tool-category.ts
+++ b/src/agent/tool-category.ts
@@ -63,7 +63,7 @@ const READ_TOOLS = new Set([
 const WRITE_TOOLS = new Set([
   'Write', 'Edit', 'NotebookEdit', 'MultiEdit',
   'write_file', 'edit_file',
-  // patch_apply applies unified diffs to files on disk — a write operation.
+  // patch_apply atomically writes structured multi-file changes to disk.
   'patch_apply',
   // memory-tools.ts — both mutate persistent state on disk (the fact
   // archive / HOT.md and ~/.afk/procedures/<name>.md respectively), so
--- a/src/agent/tools/handlers/index.ts
+++ b/src/agent/tools/handlers/index.ts
@@ -39,6 +39,7 @@ import { browserActHandler } from './browser-act.js';
 import { browserScreenshotHandler } from './browser-screenshot.js';
 import { browserCloseHandler } from './browser-close.js';
 import { waitForHandler } from './wait-for.js';
+import { patchApplyHandler, createPatchApplyHandler } from './patch-apply.js';
 
 /**
  * Build the built-in tool handler map for a session.
@@ -77,6 +78,7 @@ export function createBuiltinHandlers(
   const listDirectory = cwd !== undefined ? createListDirectoryHandler(cwd) : listDirectoryHandler;
   const terminalFontSize = createTerminalFontSizeHandler();
   const worktree = createWorktreeHandler(cwd);
+  const patchApply = cwd !== undefined ? createPatchApplyHandler(cwd) : patchApplyHandler;
   return new Map<string, ToolHandler>([
     ['bash', bash],
     ['read_file', readFile],
@@ -105,6 +107,7 @@ export function createBuiltinHandlers(
     ['browser_act', browserActHandler],
     ['browser_screenshot', browserScreenshotHandler],
     ['browser_close', browserCloseHandler],
+    ['patch_apply', patchApply],
   ]);
 }
 
@@ -129,6 +132,7 @@ export {
   configSetHandler,
   askQuestionHandler,
   waitForHandler,
+  patchApplyHandler,
   browserOpenHandler,
   browserObserveHandler,
   browserActHandler,
--- a/src/agent/tools/handlers/patch-apply-engine.test.ts
+++ b/src/agent/tools/handlers/patch-apply-engine.test.ts
@@ -1,0 +1,164 @@
+/**
+ * Tests for patch-apply-engine.ts
+ */
+
+import { describe, it, expect, afterEach } from 'vitest';
+import { writeFile, readFile, mkdir, rm } from 'fs/promises';
+import os from 'os';
+import path from 'path';
+import { randomBytes, createHash } from 'crypto';
+import { applyPatch } from './patch-apply-engine.js';
+import type { PatchFileChange } from './patch-validate.js';
+
+const tempDir = path.join(
+  os.tmpdir(),
+  `afk-patch-engine-test-${process.pid}-${randomBytes(4).toString('hex')}`,
+);
+
+async function writeTemp(filename: string, content: string): Promise<string> {
+  await mkdir(tempDir, { recursive: true });
+  const p = path.join(tempDir, filename);
+  await writeFile(p, content, 'utf-8');
+  return p;
+}
+
+function sha256(content: string): string {
+  return createHash('sha256').update(content, 'utf8').digest('hex');
+}
+
+afterEach(async () => {
+  try {
+    await rm(tempDir, { recursive: true, force: true });
+  } catch {
+    // best-effort
+  }
+  vi.restoreAllMocks();
+});
+
+describe('applyPatch (dry-run)', () => {
+  it('returns diff without modifying the file', async () => {
+    const filePath = await writeTemp('dry.txt', 'before\n');
+    const changes: PatchFileChange[] = [{ path: filePath, content: 'after\n' }];
+    const fileContents = new Map([[filePath, 'before\n']]);
+
+    const result = await applyPatch(changes, fileContents, tempDir, /* dryRun */ true);
+
+    expect(result.status).toBe('dry_run');
+    expect(result.diff).toContain('-before');
+    expect(result.diff).toContain('+after');
+    expect(result.errors).toHaveLength(0);
+
+    // File must NOT have been modified.
+    const actual = await readFile(filePath, 'utf-8');
+    expect(actual).toBe('before\n');
+  });
+
+  it('dry-run diff for sequential edits', async () => {
+    const filePath = await writeTemp('dry-edits.txt', 'foo bar baz\n');
+    const changes: PatchFileChange[] = [
+      {
+        path: filePath,
+        edits: [
+          { old: 'foo', new: 'FOO' },
+          { old: 'bar', new: 'BAR' },
+        ],
+      },
+    ];
+    const fileContents = new Map([[filePath, 'foo bar baz\n']]);
+
+    const result = await applyPatch(changes, fileContents, tempDir, true);
+
+    expect(result.status).toBe('dry_run');
+    expect(result.diff).toContain('+FOO BAR baz');
+    const actual = await readFile(filePath, 'utf-8');
+    expect(actual).toBe('foo bar baz\n'); // unchanged
+  });
+
+  it('dry-run returns correct before/after hashes', async () => {
+    const filePath = await writeTemp('hash-check.txt', 'original\n');
+    const changes: PatchFileChange[] = [{ path: filePath, content: 'replaced\n' }];
+    const fileContents = new Map([[filePath, 'original\n']]);
+
+    const result = await applyPatch(changes, fileContents, tempDir, true);
+
+    expect(result.files_changed).toHaveLength(1);
+    const fc = result.files_changed[0]!;
+    expect(fc.before_hash).toBe(`sha256:${sha256('original\n')}`);
+    expect(fc.after_hash).toBe(`sha256:${sha256('replaced\n')}`);
+  });
+});
+
+describe('applyPatch (live write)', () => {
+  it('writes correct content for full-file replacement', async () => {
+    const filePath = await writeTemp('write.txt', 'old content\n');
+    const changes: PatchFileChange[] = [{ path: filePath, content: 'new content\n' }];
+    const fileContents = new Map([[filePath, 'old content\n']]);
+
+    const result = await applyPatch(changes, fileContents, tempDir, false);
+
+    expect(result.status).toBe('applied');
+    expect(result.errors).toHaveLength(0);
+    const actual = await readFile(filePath, 'utf-8');
+    expect(actual).toBe('new content\n');
+  });
+
+  it('applies sequential edits in order', async () => {
+    const filePath = await writeTemp('seq.txt', 'step1 step2 step3\n');
+    const changes: PatchFileChange[] = [
+      {
+        path: filePath,
+        edits: [
+          { old: 'step1', new: 'STEP1' },
+          { old: 'STEP1 step2', new: 'STEP1+STEP2' },
+        ],
+      },
+    ];
+    const fileContents = new Map([[filePath, 'step1 step2 step3\n']]);
+
+    const result = await applyPatch(changes, fileContents, tempDir, false);
+
+    expect(result.status).toBe('applied');
+    const actual = await readFile(filePath, 'utf-8');
+    expect(actual).toBe('STEP1+STEP2 step3\n');
+  });
+
+  it('includes diff in successful result', async () => {
+    const filePath = await writeTemp('diff-check.txt', 'hello\n');
+    const changes: PatchFileChange[] = [{ path: filePath, content: 'world\n' }];
+    const fileContents = new Map([[filePath, 'hello\n']]);
+
+    const result = await applyPatch(changes, fileContents, tempDir, false);
+
+    expect(result.status).toBe('applied');
+    expect(result.diff).toContain('-hello');
+    expect(result.diff).toContain('+world');
+  });
+
+  it('handles new-file creation (empty originalContent)', async () => {
+    const newFilePath = path.join(tempDir, 'newfile.txt');
+    await mkdir(tempDir, { recursive: true });
+    const changes: PatchFileChange[] = [{ path: newFilePath, content: 'brand new\n' }];
+    const fileContents = new Map([[newFilePath, '']]);
+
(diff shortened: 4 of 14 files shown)


griffinwork40 (commented):

## PR #1425 — `patch_apply` tool with atomic multi-file changes

**Ref:** `962742be` · **Decision: DO NOT MERGE** — 1 high, 2 medium blocking (1 security, 1 correctness); 1 low non-blocking.

---

### 🔴 HIGH — Blocking

**[C-F1] `isNewFile` is always `false`, breaking partial-failure rollback for new files** · `patch-apply-engine.ts:140–145`
The engine derives `isNewFile = !fileContents.has(resolvedPath)`. But `validatePatchChanges` (patch-validate.ts:167) unconditionally calls `fileContents.set(resolvedPath, currentContent)` — including `''` for non-existent files. So `has()` is always `true` and `isNewFile` is always `false`. On a partial rename failure, the rollback path (line 268–273) writes an empty string to disk instead of unlinking the newly-created file, leaving a zero-byte artifact where no file should exist. The engine's own comment at line 142–143 describes the intended invariant, but validation violates it.
*Fix: Either do not `set` the key for new files in the validator (skip the `fileContents.set` on the `content = ''` catch branch), or pass a separate `newFilePaths: Set<string>` out of validation.*

---

### 🟠 MEDIUM — Blocking

**[S-F1] After `write_denied` error, validation reads the denylisted file content** · `patch-validate.ts:117–167`
After a denylist hit (step 2), there is no `continue` — execution falls through to step 5 (`readFile`), storing the denylisted file's content in `fileContents`. Contrast: `path_containment` errors (step 1, line 110) correctly `continue`. While `applyPatch` is never called (validation fails), reading content of write-protected files into process memory is unnecessary and violates defense-in-depth.
*Fix: Add `continue;` after the `errors.push` for `write_denied`, mirroring the `path_containment` pattern.*

**[C-F2] Rename loop does not break on first failure, allowing a mixed commit state** · `patch-apply-engine.ts:244–260`
The rename loop continues after the first failure — subsequent files are renamed and their successes pushed to `completedRenames`. The rollback code iterates `completedRenames` so those files ARE restored, but a second rename failure during rollback (best-effort `catch` at line 275) would leave committed files that the caller believes were rolled back. A `break` on first error would bound the blast radius to a single failed rename.
*Fix: Add `break;` after `renameErrors.push` to stop committing new renames once a failure occurs.*

---

### 🟡 LOW — Non-blocking

**[C-F3] `extractCandidatePath` returns only the first change's path** · `path-approval-hook.ts:493–501`
The path-approval hook extracts only `changes[0].path` for the containment grant prompt. A multi-file patch where `changes[0]` targets an already-approved directory but `changes[1]` targets an unapproved one will not trigger a second approval prompt. Lower risk because `patch-validate.ts` enforces per-file containment independently, so the write is still blocked — the gap is in the UX prompt, not enforcement.

---

### What Was Not Checked
- Citations verified against branch HEAD `962742be`.
- Not checked: concurrent multi-session races against the same file set, Windows path behavior (NTFS rename semantics differ), whether `computeLineDiff` handles binary content gracefully, end-to-end integration test with the full `SessionToolDispatcher` path.

griffinwork40 (commented):

---

## PR #1425 — `patch_apply` tool with atomic multi-file changes

**Reviewed ref:** `962742be0b7302747a89d46a190c0818dab97543` · **CI:** ✅ all green  
**Prior automated rounds:** 0 (first review)

### Decision: DO NOT MERGE

**Arithmetic:** 0 critical, 2 high → DO NOT MERGE; 2 medium advisory (non-blocking per sweep floor); 3 low advisory; 2 nit advisory.

---

## 🔴 HIGH — Blocking

### [H-1] `isNewFile` is always `false` — new-file rollback writes empty file instead of unlinking

**Dimension:** correctness  
**File:** `src/agent/tools/handlers/patch-apply-engine.ts:144` + `src/agent/tools/handlers/patch-validate.ts:162–166`  
**Ref:** `962742be` · citation-type: file-state (both verified)

The validator unconditionally calls `fileContents.set(resolvedPath, currentContent)` for every file that passes validation — including new files where `currentContent = ''`. The engine then computes:

```typescript
const isNewFile = !fileContents.has(resolvedPath);  // always false
```

Because the validator always inserts the key, `isNewFile` is permanently `false`. When a partial rename failure triggers rollback, the `if (completed.isNewFile) { await unlink(...) }` branch is dead code. Instead, rollback calls `writeFile(targetPath, '', 'utf-8')` for a file that did not exist before the patch, leaving a spurious empty file on disk.

**Fix:** Add a `newFilePaths: Set<string>` to `ValidationResult`, populated in the new-file catch branch alongside `fileContents.set(resolvedPath, '')`. The engine reads `newFilePaths.has(resolvedPath)` instead of `!fileContents.has(resolvedPath)`. Update the misleading inline comment at engine line 141–143. The missing test (see M-2) would catch this immediately.

---

### [H-2] Path-approval hook shows only `changes[0]` to operator for multi-file patches

**Dimension:** correctness / security  
**File:** `src/agent/tools/hooks/path-approval-hook.ts:497–504`  
**Ref:** `962742be` · citation-type: file-state (verified)

`extractCandidatePath` for `patch_apply` returns only `changes[0].path`:

```typescript
if (toolName === 'patch_apply') {
  const changes = input['changes'];
  if (Array.isArray(changes) && changes.length > 0) {
    const first = changes[0] as Record<string, unknown>;
    return typeof first['path'] === 'string' ? first['path'] : undefined;
  }
```

For a multi-file patch where `changes[0]` is inside the session's write root (no prompt fires) and `changes[1+]` are outside it, the operator receives no approval prompt for the higher-index files. The handler's `resolveAndContain` does enforce containment on every file (no bypass), but: when the operator grants a write-root via "once"/"session" approval for `changes[0]`'s directory and `changes[1]` resolves inside that newly-added root, `changes[1]` proceeds without the operator ever seeing it in the approval prompt. The operator approved one file's directory but implicitly authorized additional files in the same tree they may not have intended to touch.

**Fix:** Loop over all `changes[*].path` values in `extractCandidatePath` and return the highest-risk path (determined by `wouldBeRestricted`), or the common ancestor directory of all out-of-root paths. The handler already validates every path; the hook should surface the same scope to the operator.

---

## 🟠 MEDIUM — Advisory (non-blocking per sweep floor; suggest follow-up issue)

### [M-1] `runningContent` not advanced when an edit fails — cascading validation errors masked

**Dimension:** correctness  
**File:** `src/agent/tools/handlers/patch-validate.ts:196–212`  
**Ref:** `962742be` · citation-type: file-state (verified)

When `edit[i]` returns zero matches, the validator pushes `edit_not_found` and continues without advancing `runningContent`. `edit[i+1]` is validated against pre–edit-`[i]` content. For sequential edit chains where each edit targets text produced by a prior edit, validation reports `edit[i+1]` as passing even though it would also fail in the intended post–edit-`[i]` state. The caller fixes `edit[0]`, resubmits, and now `edit[1]` fails — a second unnecessary round trip. The stated "all validation errors collected" goal is not met for this dependency class.

**Suggestion:** On `edit_not_found`, advance `runningContent` by leaving it unchanged (correct) but mark subsequent edits that referenced the would-be post-edit content as speculative. For `edit_ambiguous`, apply the intended replacement at the first occurrence to advance `runningContent` optimistically. Add a test: two dependent edits where `edit[0]` is ambiguous and `edit[1]` targets the intended post-`edit[0]` string.

### [M-2] `AbortSignal` not forwarded — temp files not cleaned on session teardown

**Dimension:** correctness  
**File:** `src/agent/tools/handlers/patch-apply.ts:120`  
**Ref:** `962742be` · citation-type: file-state (verified)

The `_signal: AbortSignal` parameter (underscore-prefixed = intentionally unused) is never forwarded to `validatePatchChanges` or `applyPatch`. An in-flight abort (session teardown, turn cancellation) during a large multi-file write does not interrupt the sequential `readFile`/`writeFile`/`mkdir`/`rename` loop and does not trigger temp-file cleanup. `write-file.ts` passes `signal` to `writeFile()` and checks `signal.aborted` at entry; `patch-apply` regresses this behavior.

**Suggestion:** Thread `signal` into both functions. In Phase 2, pass `signal` to `writeFile(tmp, ..., { signal })` and check `signal.aborted` before each loop iteration, running cleanup on abort.

---

## 🟡 LOW — Advisory (follow-up welcome, not blocking)

### [L-1] `files_changed[i].path` is absolute while diff header is relative — caller cannot correlate

`patch-apply-engine.ts`: `filesChanged` sets `path: p.resolvedPath` (absolute) while `formatUnifiedDiff` uses `change.path` (caller-supplied, potentially relative). Use `change.path` consistently, or add a `resolved_path` field alongside the caller-supplied `path`.

### [L-2] `edits: []` and `edit.old = ''` pass validation silently

Schema permits `edits: []` (no `minItems`). Empty edits array passes step 4 (`edits !== undefined`), step 7 is a no-op, result is `applied` with unchanged file and identical hashes. `edit.old = ''` is guarded by `countOccurrences` returning `0` → `edit_not_found`, but this should be a schema-level `minLength: 1` rejection. Add schema constraints and tests.

### [L-3] `countOccurrences` is O(|haystack| × |needle|) with no batch cap

No `maxItems` on `changes` or `edits`, no per-file size guard. A 100-file patch with 100 edits each on 1 MB files produces ~10 billion character comparisons in validation. Add `maxItems: 50` on `changes`, `maxItems: 100` on `edits`, and a per-file size guard (e.g. reject files > 10 MB with `file_too_large`).

---

## 🔵 NIT

- **[N-1]** `patch-apply-engine.ts:141–143`: The inline comment explaining `isNewFile` is self-contradictory ("Use `has()` to distinguish missing key" then "the validator always sets the key"). Fix after H-1 is resolved.
- **[N-2]** `patch-apply.ts` empty-changes path: `status: 'applied'` for `changes: []` is a contract oddity — callers cannot distinguish "applied" from "nothing to apply." Add `minItems: 1` to schema or return `status: 'noop'`.

---

## What Was Not Checked

- Citations verified inline against branch HEAD `962742be0b7302747a89d46a190c0818dab97543`.
- **One Wave 1 finding dropped (FABRICATED):** S-1 claimed `patch_apply` was absent from `tool-category.ts WRITE_TOOLS`. Wave 1.5 verification found `patch_apply` IS present at line 67 of `tool-category.ts` at ref `962742be`. Finding dropped before synthesis.
- Stated intent: PR #1425 title + body — spec-compliance assessed.
- Not checked: behavior under concurrent `patch_apply` calls on overlapping paths, behavior under macOS APFS vs. Linux ext4 rename atomicity, end-to-end session teardown with active write, Windows path behavior (CI explicitly skips Windows).
- Tests: 31 tests pass per CI; test logic verified by reading test files, not by executing them locally.

PR #1447, opened 2026-09-02
fix(security): align PID=1 threshold between parseInput and evaluateProcess (#1443)

### LINKED ISSUE #1443: fix(security): align PID=1 threshold between parseInput and evaluateProcess
## Problem

Two layers encode the same PID validity invariant at different thresholds:

1. **`wait-for.ts:141` (`parseInput`)** — allows PID 1 through: `input.pid <= 0`
2. **`wait-for-conditions.ts` (`evaluateProcess`)** — catches PID 1: `pid <= 1`

Defense exists at the deeper layer, but a future caller path that constructs a `ProcessCondition` directly (bypassing `parseInput`) and passes PID 1 will hit the registry gate and log a confusing "not a session-owned process" message rather than the correct "not a valid target (PID > 1)" message.

## Suggested fix

One-line change in `wait-for.ts:141`:

```diff
- if (typeof input.pid !== 'number' || !Number.isInteger(input.pid) || input.pid <= 0) {
+ if (typeof input.pid !== 'number' || !Number.isInteger(input.pid) || input.pid <= 1) {
```

Update the corresponding test expectation if one exists for the `pid <= 0` boundary.

### Secondary: fix misleading JSDoc on spawnedPidRegistry

The `ToolHandlerContext.spawnedPidRegistry` JSDoc references "legacy callers, subagents that do not carry a registry." In practice, all new subagents DO carry a registry (their own fresh one via provider construction). The "legacy" fallback only applies to direct-invocation test paths and out-of-tree `evaluateProcess` callers. The JSDoc should be updated to reflect this.

Also worth documenting: `test_run` internally spawns processes but those PIDs are never registered (the handler doesn't receive a `spawnedPidRegistry`). Not currently exploitable since `test_run` doesn't expose PIDs, but the asymmetry should be noted.

## Origin

Identified during review of #1436.

griffinwork40:
Fixes #1443

## Problem

Two layers encoded the same PID validity invariant at different thresholds:

| Layer | Check | PID 1 allowed? |
|-------|-------|----------------|
| `parseInput` (`wait-for.ts:141`) | `pid <= 0` | ✅ yes (bug) |
| `evaluateProcess` (`wait-for-conditions.ts:192`) | `pid <= 1` | ❌ no (correct) |

PID 1 (`init`/`launchd`/`systemd`) is a reserved system process. Probing it via
`process.kill(1, 0)` is an information-disclosure channel with no legitimate
`wait_for` use case. The condition layer already rejected it; this change closes
the gap at the input-validation layer so both layers share the same threshold.

### `src/agent/tools/handlers/wait-for.ts`
- Line 141: `input.pid <= 0` → `input.pid <= 1`
- Error message updated: "must be a positive integer" → "must be an integer > 1"

### `src/agent/tools/handlers/wait-for.test.ts`
- New test: `returns error when process pid is 1 (reserved system PID)` — explicitly
  covers the previously-open boundary at the `parseInput` layer

### `src/agent/tools/types.ts`
- JSDoc on `ToolHandlerContext.spawnedPidRegistry`: removes "legacy callers" framing
  since all current session construction paths supply a registry

## Verification

```
pnpm lint    ✅ clean
pnpm test src/agent/tools/handlers/wait-for.test.ts            ✅ 29/29
pnpm test src/agent/tools/handlers/wait-for-conditions.test.ts ✅ 31/31
pnpm test src/agent/tools/handlers/wait-for-poller.test.ts     ✅ 13/13
```

### DIFF
--- a/src/agent/tools/handlers/wait-for.test.ts
+++ b/src/agent/tools/handlers/wait-for.test.ts
@@ -102,6 +102,15 @@ describe('waitForHandler — input validation', () => {
     expect(result.content).toMatch(/pid/);
   });
 
+  // #1443: parseInput and evaluateProcess must share the same PID=1 threshold.
+  // PID 1 (init/launchd) is a reserved system process and must be rejected at
+  // the input-validation layer, not allowed through to evaluateProcess.
+  it('returns error when process pid is 1 (reserved system PID)', async () => {
+    const result = await waitForHandler({ type: 'process', pid: 1 }, neverSignal);
+    expect(result.isError).toBe(true);
+    expect(result.content).toMatch(/pid/);
+  });
+
   it('returns error when timeout_ms is not a number', async () => {
     const result = await waitForHandler({ type: 'command', command: 'true', timeout_ms: 'long' }, neverSignal);
     expect(result.isError).toBe(true);
--- a/src/agent/tools/handlers/wait-for.ts
+++ b/src/agent/tools/handlers/wait-for.ts
@@ -138,8 +138,8 @@ function parseInput(
       return { condition, timeoutMs, pollIntervalMs, backoff };
     }
     case 'process': {
-      if (typeof input.pid !== 'number' || !Number.isInteger(input.pid) || input.pid <= 0) {
-        throw new Error('"pid" is required for type "process" and must be a positive integer');
+      if (typeof input.pid !== 'number' || !Number.isInteger(input.pid) || input.pid <= 1) {
+        throw new Error('"pid" is required for type "process" and must be an integer > 1');
       }
       const condition: WaitCondition = { type: 'process', pid: input.pid };
       return { condition, timeoutMs, pollIntervalMs, backoff };
--- a/src/agent/tools/types.ts
+++ b/src/agent/tools/types.ts
@@ -116,9 +116,8 @@ export interface ToolHandlerContext {
    * this session actually spawned. Any PID not in the registry is rejected
    * with `blocked: true` so the model cannot probe arbitrary host processes.
    *
-   * When absent (e.g. legacy call sites, subagents that do not carry a
-   * registry), the old behaviour is preserved: any PID > 1 is probed. New
-   * session construction paths should always supply this field.
+   * All current session construction paths supply this field. When absent,
+   * the pre-registry behaviour is preserved: any PID > 1 is probed.
    */
   spawnedPidRegistry?: SpawnedPidRegistry;
 }


griffinwork40 (commented):

## Review: `fix(security): align PID=1 threshold between parseInput and evaluateProcess (#1443)`

**Reviewed ref:** `8a575d8b2b` · **Regime:** light (3 files, ~20 lines)
**Stated intent:** PR #1447 title + body — spec-compliance assessed.

---

## Decision: MERGE — 0 blocking (3 low, 1 nit).

---

## Findings

### F1 · low · blocking:false · confidence:high · test-coverage · `src/agent/tools/handlers/wait-for.test.ts:113-115` · ref:8a575d8b2b · file-state

The new PID=1 test asserts `result.content` matches `/pid/`, which also matches the pre-fix error message `"must be a positive integer"`. A future refactor reverting the threshold from `pid <= 1` back to `pid <= 0` would still pass this test — creating a false-green signal for the exact regression this PR closes.

```ts
expect(result.content).toMatch(/pid/);
```

**Suggestion:** Tighten to `expect(result.content).toMatch(/integer > 1/)` so the test pins the corrected threshold wording.

---

### F2 · low · blocking:false · confidence:high · test-coverage · `src/agent/tools/handlers/wait-for.test.ts:102-116` · ref:8a575d8b2b · file-state

No test covers PID=2 (the first valid PID under the new threshold). Without a boundary-passing case, a future mis-typed `pid < 2` (strict less-than) would silently block PID=2. The test file contains only negative-path tests for `pid: -1` and `pid: 1` at this boundary.

```ts
// Only rejection tests exist at lines 102-116; no pid: 2 acceptance test
```

**Suggestion:** Add `it('accepts pid 2 as the lowest valid process PID', ...)` asserting `result.isError` is falsy.

---

### F3 · low · blocking:false · confidence:medium · correctness · `src/agent/tools/types.ts:117-119` · ref:8a575d8b2b · file-state

The updated JSDoc asserts "All current session construction paths supply this field" while preserving the fallback doc ("When absent, the pre-registry behaviour is preserved"). The two sentences are in tension: the first asserts exhaustive coverage, the second documents a live fallback for the scenario the first claims cannot happen. No enforcement mechanism verifies the completeness assertion.

```ts
* All current session construction paths supply this field. When absent,
* the pre-registry behaviour is preserved: any PID > 1 is probed.
```

**Suggestion:** Soften to `"Session construction paths are expected to supply this field"` — avoids encoding a completeness guarantee the comment alone cannot enforce.

---

### F4 · nit · blocking:false · confidence:high · correctness · `src/agent/tools/handlers/wait-for.ts:141` · ref:8a575d8b2b · file-state

The error message uses the mathematical symbol `>` in prose. `"must be an integer > 1"` reads correctly but `"must be an integer greater than 1"` is unambiguous across all consumer contexts (log aggregators, structured error parsers).

---

## Dimensions with no findings

- **Security:** no issues found — the change tightens PID=1 rejection to the input-validation layer, aligning both layers at `pid <= 1`. Attack surface reduced, not expanded. Read `wait-for.ts:138-145` and `wait-for-conditions.ts:192`.
- **API compatibility:** no issues found — `parseInput` is an internal function (not exported). The public boundary `waitForHandler` now rejects PID=1 with `isError: true`, matching the downstream behaviour that already rejected it. No caller could have received a successful result for PID=1.
- **Spec compliance:** all four stated deliverables implemented — threshold tightened, error message updated, PID=1 test added, JSDoc updated. No scope creep. No unmet intent.
- **Perf / observability:** no issues found — O(1) integer guard at parse time. No hot-path changes.

## CI Status

✅ All checks pass (Lint & Build, Test ubuntu-latest, Test macos-latest, Test real PTY scrollback, Docs Site, Publish Bundle Smoke).

## What was not checked

- Citations verified against branch HEAD `8a575d8b2b28eac49d7ff35e3b0646f50c1676d5` (file-state reads + diff context).
- Stated intent: PR #1447 title + body — spec-compliance assessed.
- Did not run tests locally.
- Did not audit all session construction paths for `spawnedPidRegistry` completeness (relevant to F3's JSDoc claim).
- Did not review `wait-for-conditions.test.ts` or `wait-for-poller.test.ts` beyond confirming the `pid <= 1` threshold in `evaluateProcess` at line 192.

PR #1452, opened 2026-09-02
fix(tui): clamp content lines to terminal width in task-view-mid-turn

griffinwork40:
## Problem

`launchMidTurnTaskView` (the Tab-triggered mid-turn subagent view) wrote subagent conversation history, live output events, and user input echo directly to `stdout.write` with **no width guard**. The compositor is suspended during this view, so the normal repaint/overlay width-clamping path is bypassed entirely. Subagent history lines, tool output, and live streaming events could easily exceed the terminal width, corrupting the TUI layout.

## Fix

Added a local `clamp` helper using the existing `truncateDisplayWidth` from `../../display.js`:

```ts
const clamp = (s: string): string =>
  truncateDisplayWidth(s, stdout.columns || 80);
```

Applied `clamp()` to every user-visible content string before writing:

| Write site | Applied |
|---|---|
| Header line (`renderTaskViewHeader`) | ✅ |
| Each history line in the `for … split('\n')` loop | ✅ |
| Footer/separator (`buildTaskFooterLine`) | ✅ |
| User input echo on Enter | ✅ |
| "Type a message + Enter to send, Esc to return" hint | ✅ |
| Each `formatOutputEvent` result in the live stream loop | ✅ |
| "Subagent completed. Press Esc to return." tail line | ✅ |

Cursor-movement ANSI sequences (`\x1b[2J\x1b[H`, `\r\x1b[K`) bypass `clamp()` entirely — they are positional codes, not content, and must not be truncated. `truncateDisplayWidth` is ANSI-aware: it preserves SGR color codes while clamping visible display width.

## Tests

Two new tests in `task-view-mid-turn.test.ts` under **"launchMidTurnTaskView width clamping"**:

1. **Long history line is clamped** — drives `launchMidTurnTaskView` with a mock compositor whose `stdout.columns = 40` and a history entry containing 200 `'A'` characters; asserts every visible-content line in the captured output is ≤ 40 characters wide.
2. **Cursor-movement ANSI sequences are not clamped** — asserts the raw output still contains `\x1b[2J\x1b[H` (clear-screen + cursor-home) after the view runs.

All 10 tests in the file pass (8 pre-existing + 2 new). `pnpm lint` and `pnpm build` clean.

### DIFF
--- a/src/cli/commands/interactive/task-view-mid-turn.test.ts
+++ b/src/cli/commands/interactive/task-view-mid-turn.test.ts
@@ -137,3 +137,117 @@ function makeHost(
     ...overrides,
   } as never;
 }
+
+// ---------------------------------------------------------------------------
+// Width clamping — launchMidTurnTaskView content writes
+// ---------------------------------------------------------------------------
+
+describe('launchMidTurnTaskView width clamping', () => {
+  it('clamps long subagent history lines to terminal width', async () => {
+    const { launchMidTurnTaskView } = await import('./task-view-mid-turn.js');
+
+    // Collect every chunk written to our fake stdout.
+    const written: string[] = [];
+    const fakeStdout = {
+      columns: 40,
+      write: (s: string) => { written.push(s); return true; },
+    };
+
+    // A history line that is far wider than 40 columns.
+    const longLine = 'A'.repeat(200);
+
+    // Minimal fake session: non-empty history, immediate-complete stream.
+    const fakeSession = {
+      getHistory: () => [
+        { role: 'assistant' as const, content: longLine },
+      ],
+      getOutputStream: async function* () {
+        // yield nothing — subagent already done
+      },
+    };
+
+    // Fake handle — already completed so the view returns immediately.
+    const fakeHandle = {
+      status: 'succeeded' as const,
+      session: fakeSession,
+      sendMessage: vi.fn(),
+    };
+
+    const fakeManager = {
+      list: () => [{ id: 'sub-1', status: 'running' as const }],
+      get: (_id: string) => fakeHandle as never,
+    };
+
+    const fakeCompositor = {
+      stdout: fakeStdout as never,
+      suspendInput: vi.fn(),
+      resumeInput: vi.fn(),
+      repaint: vi.fn(),
+    };
+
+    await launchMidTurnTaskView({
+      manager: fakeManager as never,
+      compositor: fakeCompositor as never,
+    });
+
+    // Every non-ANSI-only line written must be ≤ 40 visible characters.
+    // We join all written chunks and split by newline, then check each line.
+    const allOutput = written.join('');
+    const lines = allOutput.split('\n');
+
+    // Strip ANSI escape sequences for width measurement (CSI + OSC patterns).
+    const stripAnsi = (s: string): string =>
+      s.replace(/\x1b\[[0-9;]*[A-Za-z]/g, '').replace(/\x1b\][^\x07\x1b]*(?:\x07|\x1b\\)/g, '');
+
+    // Cursor-positioning sequences like \r\x1b[K (erase line) are fine —
+    // they should not be clamped. Only visible content lines must be ≤ cols.
+    for (const line of lines) {
+      const visible = stripAnsi(line);
+      // Skip empty lines and bare CR (cursor movement remnants).
+      if (visible.replace(/\r/g, '').trim() === '') continue;
+      // Skip the clear-screen line (\x1b[2J\x1b[H).
+      if (line.includes('\x1b[2J')) continue;
+      expect(visible.length).toBeLessThanOrEqual(40);
+    }
+  });
+
+  it('does not clamp cursor-movement ANSI sequences (CSI codes)', async () => {
+    const { launchMidTurnTaskView } = await import('./task-view-mid-turn.js');
+
+    const written: string[] = [];
+    const fakeStdout = {
+      columns: 40,
+      write: (s: string) => { written.push(s); return true; },
+    };
+
+    const fakeSession = {
+      getHistory: () => [],
+      getOutputStream: async function* () {},
+    };
+    const fakeHandle = {
+      status: 'succeeded' as const,
+      session: fakeSession,
+      sendMessage: vi.fn(),
+    };
+    const fakeManager = {
+      list: () => [{ id: 'sub-2', status: 'running' as const }],
+      get: (_id: string) => fakeHandle as never,
+    };
+    const fakeCompositor = {
+      stdout: fakeStdout as never,
+      suspendInput: vi.fn(),
+      resumeInput: vi.fn(),
+      repaint: vi.fn(),
+    };
+
+    await launchMidTurnTaskView({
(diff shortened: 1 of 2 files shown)


griffinwork40 (commented):

## Review: `fix(tui): clamp content lines to terminal width in task-view-mid-turn`

**Reviewed ref:** `31f7533694ecb9a569f71c10f4c21913b3d11fb4` · **Regime:** light (2 files, ~150 diff lines)
**Stated intent:** PR #1452 title + body — spec-compliance assessed.

---

## Prior Rounds

_Round 1 of automated review on this PR. No prior findings to carry forward._

---

## Decision: DO NOT MERGE — 1 high blocking; 2 medium advisory; 1 low advisory.

**Blocking severity floor applied:** only `critical` and `high` findings block. `medium` and below are advisory — listed as follow-ups, non-blocking per sweep policy.

**Arithmetic:** 0 critical, 1 high (correctness — blocks) → **DO NOT MERGE**; 2 medium advisory (1 correctness, 1 test-coverage); 1 low.

---

## Blocking Findings (High)

### F1 · high · blocking:true · confidence:high · correctness · `src/cli/commands/interactive/task-view-mid-turn.ts:101` · ref:31f7533694 · file-state

**`clamp(renderTaskViewHeader(...))` applies `truncateDisplayWidth` to a multi-line string — silently drops the bottom divider on every standard terminal**

`renderTaskViewHeader` returns a 3-line string joined with `\n` (task-view-mode.ts:84: `return [divider(undefined, 80), parts.join('  '), divider(undefined, 80)].join('\n')`). `string-width` treats `\n` as zero-width, so `displayWidth` measures the composite as ~165 visible chars (two 80-char dividers + middle line). `truncateDisplayWidth` compares this against `stdout.columns` (e.g. 120) and truncates at that offset — cutting mid-string, dropping the bottom divider and appending `…`. Shadow-verified: `string-width('x'.repeat(80) + '\n' + 'hello' + '\n' + 'x'.repeat(80))` returns `165`.

This is a regression introduced by this PR: the pre-change code wrote the header without clamping, which could overflow (the stated bug) but preserved the visual structure. Post-change, the header is structurally broken on any terminal where `columns < total_displayWidth` — which is every standard terminal (80, 120, even 160 cols), since the two 80-char dividers alone sum to 160.

```ts
// task-view-mode.ts:84 — returns 3 lines
return [divider(undefined, 80), parts.join('  '), divider(undefined, 80)].join('\n');

// task-view-mid-turn.ts:101 — clamp treats it as one line
stdout.write(clamp(renderTaskViewHeader(id, status, agentType)) + '\n\n');
```

**Suggestion:** Split on `\n` and clamp each line independently. E.g. a `clampLines` helper: `(s: string): string => s.split('\n').map(l => truncateDisplayWidth(l, width)).join('\n')`, or apply `clamp` per-line at the write site.

---

## Advisory Follow-ups (Non-Blocking · Medium)

_These are advisory — suggest creating follow-up issues. Do NOT add a fix commit for these._

### F2 · medium advisory · correctness · `src/cli/commands/interactive/task-view-mid-turn.ts:199` · ref:31f7533694 · file-state

**`clamp(text)` on `formatOutputEvent` output has the same multi-line measurement flaw as F1**

`formatOutputEvent` for `event.type === 'message'` returns `JSON.stringify(c)` for non-string content (output-event-format.ts:43), which produces multi-line output for nested objects. `clamp(text)` applies the same single-unit measurement. Additionally, `\r\x1b[K` only clears the first terminal row — subsequent `\n`-separated lines in the `clamp` output print without the erase prefix.

```ts
// output-event-format.ts:43
const text = typeof c === 'string' ? c : JSON.stringify(c);

// task-view-mid-turn.ts:199 — multi-line text treated as single line
stdout.write(`\r\x1b[K${clamp(text)}\n`);
```

**Follow-up suggestion:** Apply the same `clampLines` fix from F1 at this write site. For the `\r\x1b[K` prefix, split on `\n` and write each sub-line with its own erase-to-EOL prefix.

---

### F3 · medium advisory · test-coverage · `src/cli/commands/interactive/task-view-mid-turn.test.ts` · ref:31f7533694 · file-state

**5 of 7 changed write sites have no test coverage; the streaming path is never exercised**

Test 1 uses `status: 'succeeded'`, triggering the `!isRunning` early-return at line 133. The `for await` streaming loop (lines 193–200), the hint line (line 190), the user-echo path (line 165), and the "Subagent completed" tail (line 209) are never reached. The multi-line header is exercised but the test's `stripAnsi` + `visible.length` assertion does not catch F1's flaw because `renderTaskViewHeader` returns newlines that split into separate "lines" in the test's `allOutput.split('\n')` — each sub-line individually passes `≤ 40`, masking that `truncateDisplayWidth` already broke the string.

**Follow-up suggestion:** Add a test with `status: 'running'` and a mock `getOutputStream` that yields a wide event, asserting the output line is clamped. Separately, add a test that asserts `renderTaskViewHeader`'s 3-line structure is preserved after clamping (all 3 lines present, each ≤ cols).

---

## Low Findings (Non-Blocking)

### F4 · low · blocking:false · confidence:high · spec-compliance · `src/cli/commands/interactive/task-view-mid-turn.ts:148–150` · ref:31f7533694 · file-state

**`renderPrompt()` writes `inputBuf` without clamping — stated intent covers "every user-visible content string"**

`renderPrompt` at line 150 writes `\r\x1b[K${palette.dim('> ')}${inputBuf}` with no `clamp()` call. `inputBuf` is byte-capped at 8192 (MAX_INPUT_BYTES) but not display-width-capped. On a narrow terminal with a long paste, the prompt overflows. This write site was not changed in the diff — pre-existing gap, but the stated intent implies full coverage.

**Suggestion:** Apply `clamp` to the prompt content: `stdout.write(\`\r\x1b[K${clamp(palette.dim('> ') + inputBuf)}\`)`.

---

## Dimensions With No Findings

- **Security:** no issues found — no new data paths from untrusted sources; `stripEscapeSequences` sanitizes content before `clamp()`.
- **API compatibility:** no issues found — `truncateDisplayWidth` is an existing internal export with unchanged signature. No external-facing API touched.
- **Perf-observability:** no issues found — `truncateDisplayWidth` tokenizes ANSI per call via a stateless regex; at TUI frame rates this is not measurable.

---

## CI Status

- ✅ **All checks green:** Lint & Build, Test (ubuntu-latest), Test (macos-latest), Test (real PTY scrollback), Docs Site, Publish Bundle Smoke, Vercel.

---

## What Was Not Checked

- Citations verified inline against branch HEAD `31f7533694ecb9a569f71c10f4c21913b3d11fb4` (file-state reads via `git show`).
- F1 shadow-verified: `string-width` measurement of multi-line string confirmed locally (returns 165 for 80+\n+5+\n+80).
- Stated intent: PR #1452 title + body — spec-compliance assessed.
- Did not run tests locally.
- Did not review `divider()` implementation beyond confirming it produces an 80-char-wide line (hardcoded width arg in `renderTaskViewHeader`).
- Did not review the `stripEscapeSequences` sanitization path — pre-existing, not part of this diff.
- Did not test `truncateDisplayWidth` behavior with embedded `\n` beyond the `string-width` measurement confirmation.

PR #1396, opened 2026-08-31
fix(grant-manager): reconcile 3 dispatcher/provider behavior divergences (#500)

griffinwork40:
## fix(grant-manager): reconcile 3 dispatcher/provider behavior divergences (#500)

Consolidates path-grant logic into `PathGrantManager` and fixes three divergences between `SessionToolDispatcher` and the two providers.

---

### Finding 1 — sessionId threading gap

Providers didn't thread the per-call `sessionId` argument through to `PathGrantManager`. All audit entries landed with `sessionId: null`, breaking per-session forensics. Fixed: providers now pass `sessionId` per call via the `PathGrantManagerHooks` interface.

### Finding 2 — non-revocable anchor (Option A — migrating)

**What changed:** All consumers now use **Option A (migrating anchor)**: `getProtectedRoot()` returns the **current** `resolveBase`, so after a `setResolveBase` / `setCwd` call the new worktree root becomes the protected root.

- **Dispatcher** (`dispatcher.ts`): removed `_initialResolveBase` field and `getDisplayResolveBase` hook (Option B artifacts). `getProtectedRoot` returns `this.resolveBase` (current, mutable cwd).
- **anthropic-direct** (`provider-grants.ts`): removed `_initialResolveBase`; `getProtectedRoot` returns `_currentCwd`, which is updated by `setCurrentCwd` on each `setCwd` call.
- **openai-compatible** (`index.ts`): replaced `_initialResolveBase` with `_sharedCurrentCwd`, updated in `ensureSharedRoots` and in `rebuildEnvironmentBlock` (the `onCwdChange` callback), so `getProtectedRoot` always returns the live cwd.

**Test update:** `setResolveBase` test now asserts the anchor **migrates** (Option A), and the non-revocable anchor test asserts the **new** cwd is protected after migration.

### Finding 3 — no-op revoke audit guard

`revokeRoot` previously emitted an audit entry even when nothing was removed (e.g., revoking an already-absent path). Fixed: audit append is now gated on `rRemoved || wRemoved`.

---

### Verification

- `pnpm lint` (tsc --noEmit strict): ✅
- `pnpm test`: ✅ 949 files, 18 336 tests passed

PR #1479, opened 2026-09-03
fix(security): reconcile grant-manager divergences (#500)

### LINKED ISSUE #500: grant-manager: 3 behavior divergences to reconcile (anchor policy needs a decision) — from #489 consolidation

griffinwork40:
## Summary

Adds dedicated unit tests for `PathGrantManager` at `src/agent/tools/grant-manager.test.ts` — testing the shared implementation directly (not through `SessionToolDispatcher` or provider wrappers) to lock in all three divergence reconciliations from #500.

The implementation fixes for all three divergences are already present on `main` (merged in #1396). This PR adds the missing implementation-level test coverage that validates each finding independently of the consuming wrappers.

### Finding 1 — Dispatcher silently drops per-call `sessionId`

**Fix**: `SessionToolDispatcher.addReadRoot/addWriteRoot/revokeRoot` accept a 3rd `sessionId?: string` parameter and forward it to `PathGrantManager`, which uses it in audit entries (taking priority over the `getDefaultSessionId` hook). The `GrantManager` interface in `allow-dir.ts` already declared the 3-arg signature.

**New tests**: verify per-call sessionId wins over hook default; hook fires as fallback; null emitted when neither is provided; all three methods thread the arg correctly.

### Finding 2 — Opposite non-revocable-anchor policies

**Fix**: All three consumers (dispatcher, anthropic-direct, openai-compatible) now use **Option A (migrating anchor)** — `getProtectedRoot()` returns the **current** `resolveBase`/`_currentCwd`, so after a `setResolveBase`/`setCwd` call the new worktree root becomes the protected anchor. The old launch dir becomes revocable.

**New tests**: confirm `getProtectedRoot()` is consulted fresh on each call; test anchor migration (old anchor becomes revocable, new anchor is protected); confirm no audit row fires when the protected-root guard prevents a revoke.

### Finding 3 — Provider `revokeRoot` audits unconditionally on no-op

**Fix**: `PathGrantManager.revokeRoot` checks `if (rRemoved || wRemoved)` before appending an audit row — no-op revokes (path not in any list) emit zero audit entries, matching the dedup invariant on the grant paths.

**New tests**: path never granted → zero entries; double-revoke second call → zero extra entries; protected-root guard fires before removal → zero entries; uninitialized-provider path → zero entries.

## Files changed

| File | Change |
|------|--------|
| `src/agent/tools/grant-manager.test.ts` | **New** — 33 unit tests for `PathGrantManager` directly |

## Test results

```
✓ src/agent/tools/grant-manager.test.ts (33 tests) 11ms
✓ src/agent/tools/dispatcher-audit-log.test.ts (8 tests)
✓ src/agent/tools/dispatcher.test.ts (128 tests | 1 skipped)
✓ src/agent/providers/anthropic-direct/grant-manager.test.ts (9 tests)

Test Files  4 passed (4)
     Tests  177 passed | 1 skipped (178)
```

Lint: `tsc --noEmit` ✓

Closes #500

### DIFF
--- a/src/agent/tools/grant-manager.test.ts
+++ b/src/agent/tools/grant-manager.test.ts
@@ -62,16 +62,19 @@ function makeHooks(opts?: {
 // Audit-log helpers
 // ---------------------------------------------------------------------------
 
-let tmpHome: string;
+let tmpHome: string = '';
 let prevAfkHome: string | undefined;
 let prevHome: string | undefined;
+let prevStateDir: string | undefined;
 
 function setupAuditEnv(): void {
   tmpHome = mkdtempSync(path.join(tmpdir(), 'pgm-test-'));
   prevAfkHome = process.env['AFK_HOME'];
   prevHome = process.env['HOME'];
+  prevStateDir = process.env['AFK_STATE_DIR'];
   process.env['AFK_HOME'] = tmpHome;
   process.env['HOME'] = tmpHome;
+  delete process.env['AFK_STATE_DIR'];
 }
 
 function teardownAuditEnv(): void {
@@ -79,6 +82,8 @@ function teardownAuditEnv(): void {
   else process.env['AFK_HOME'] = prevAfkHome;
   if (prevHome === undefined) delete process.env['HOME'];
   else process.env['HOME'] = prevHome;
+  if (prevStateDir === undefined) delete process.env['AFK_STATE_DIR'];
+  else process.env['AFK_STATE_DIR'] = prevStateDir;
   rmSync(tmpHome, { recursive: true, force: true });
 }
 
@@ -123,9 +128,9 @@ describe('PathGrantManager — addReadRoot', () => {
   it('resolves the path to absolute before storing', () => {
     const hooks = makeHooks();
     const gm = new PathGrantManager(hooks);
-    // path.resolve('/abs') === '/abs'; this ensures we aren't double-resolving
-    gm.addReadRoot('/abs/path', 'slash');
-    expect(hooks._readRoots).toContain('/abs/path');
+    // Use a relative input so the assertion is non-tautological.
+    gm.addReadRoot('relative/sub', 'slash');
+    expect(hooks._readRoots).toContain(path.resolve('relative/sub'));
   });
 
   it('does not touch writeRoots', () => {
@@ -137,18 +142,27 @@ describe('PathGrantManager — addReadRoot', () => {
 
   it('calls ensureInitialized before adding (provider lazy-init hook)', () => {
     let initCalled = false;
-    const _readRoots: string[] = [];
-    const _writeRoots: string[] = [];
+    let _readRoots: string[] | undefined;
+    let _writeRoots: string[] | undefined;
     const hooks: PathGrantManagerHooks = {
       getReadRoots: () => _readRoots,
       getWriteRoots: () => _writeRoots,
-      ensureInitialized: () => { initCalled = true; },
+      ensureInitialized: () => {
+        initCalled = true;
+        // Arrays become available only after initialization.
+        _readRoots = [];
+        _writeRoots = [];
+      },
       getProtectedRoot: () => undefined,
       getAllowAll: () => false,
     };
     const gm = new PathGrantManager(hooks);
+    // Arrays are unavailable before the call — verifies lazy-init is not a no-op.
+    expect(_readRoots).toBeUndefined();
     gm.addReadRoot('/p', 'slash');
+    // ensureInitialized must have run and the root must be present.
     expect(initCalled).toBe(true);
+    expect(_readRoots).toContain(path.resolve('/p'));
   });
 
   it('is a no-op when getReadRoots returns undefined (uninitialized provider)', () => {
@@ -164,6 +178,20 @@ describe('PathGrantManager — addReadRoot', () => {
 });
 
 describe('PathGrantManager — addWriteRoot', () => {
+  it('emits only grant-write audit (no implicit grant-read) on a fresh path', () => {
+    setupAuditEnv();
+    try {
+      const hooks = makeHooks();
+      const gm = new PathGrantManager(hooks);
+      gm.addWriteRoot('/fresh/write', 'slash', 'test-sess');
+      const entries = readAuditEntries().filter((e) => e['path'] === '/fresh/write');
+      expect(entries.length).toBe(1);
+      expect(entries[0]!['action']).toBe('grant-write');
+    } finally {
+      teardownAuditEnv();
+    }
+  });
+
   it('adds path to both readRoots and writeRoots', () => {
     const hooks = makeHooks();
     const gm = new PathGrantManager(hooks);
@@ -327,7 +355,7 @@ describe('PathGrantManager — audit log (Finding 1: sessionId, Finding 3: no-op
     const gm = new PathGrantManager(hooks);
     gm.addReadRoot('/p', 'slash', 'per-call-id');
     const entries = readAuditEntries();
-    expect(entries.length).toBeGreaterThanOrEqual(1);
+    expect(entries.length).toBe(1);
     const last = entries[entries.length - 1]!;
     expect(last['sessionId']).toBe('per-call-id'); // per-call wins
   });
@@ -406,7 +434,7 @@ describe('PathGrantManager — audit log (Finding 1: sessionId, Finding 3: no-op
     expect(after).toBe(before);
   });
 
-  it('(Finding 3) emits NO entry when revoking the protected root (guard fires first)', () => {
+  it('(Finding 2) emits NO entry when revoking the protected root (guard fires first)', () => {
     const hooks = makeHooks({ protectedRoot: '/anchor' });
     hooks._readRoots.push('/anchor');
     const gm = new PathGrantManager(hooks);
--- a/src/agent/tools/grant-manager.test.ts
+++ b/src/agent/tools/grant-manager.test.ts
@@ -1,0 +1,470 @@
+/**
+ * Unit tests for PathGrantManager — the shared path-grant state machine.
+ *
+ * These tests exercise the implementation DIRECTLY (not via SessionToolDispatcher
+ * or provider wrappers) to verify the three divergences reconciled in issue #500:
+ *
+ *   Finding 1 — per-call sessionId threading: per-call `sessionId?` argument
+ *               wins over `getDefaultSessionId` hook; hook fires as fallback.
+ *   Finding 2 — non-revocable anchor policy: `getProtectedRoot()` is called
+ *               fresh on each `revokeRoot()`, so a migrating anchor (Option A /
+ *               current cwd) is always honoured.
+ *   Finding 3 — no-op revoke audit gate: `revokeRoot` emits NO audit entry when
+ *               the path was never in the root lists (no-op revoke must not emit
+ *               spurious log rows).
+ *
+ * @module agent/tools/grant-manager.test
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { mkdtempSync, readFileSync, rmSync, existsSync } from 'fs';
+import { tmpdir } from 'os';
+import path from 'path';
+import { PathGrantManager } from './grant-manager.js';
+import type { PathGrantManagerHooks } from './grant-manager.js';
+
+// ---------------------------------------------------------------------------
+// Test-fixture helpers
+// ---------------------------------------------------------------------------
+
+/** Minimal hooks that own their own mutable arrays — the simplest consumer shape. */
+function makeHooks(opts?: {
+  initialRead?: string[];
+  initialWrite?: string[];
+  protectedRoot?: string;
+  getDefaultSessionId?: () => string | undefined;
+}): PathGrantManagerHooks & {
+  _readRoots: string[];
+  _writeRoots: string[];
+  _protectedRoot: string | undefined;
+} {
+  const _readRoots: string[] = opts?.initialRead?.slice() ?? [];
+  const _writeRoots: string[] = opts?.initialWrite?.slice() ?? [];
+  let _protectedRoot: string | undefined = opts?.protectedRoot;
+  return {
+    _readRoots,
+    _writeRoots,
+    get _protectedRoot() {
+      return _protectedRoot;
+    },
+    set _protectedRoot(v: string | undefined) {
+      _protectedRoot = v;
+    },
+    getReadRoots: () => _readRoots,
+    getWriteRoots: () => _writeRoots,
+    getProtectedRoot: () => _protectedRoot,
+    getAllowAll: () => false,
+    getDefaultSessionId: opts?.getDefaultSessionId,
+  };
+}
+
+// ---------------------------------------------------------------------------
+// Audit-log helpers
+// ---------------------------------------------------------------------------
+
+let tmpHome: string;
+let prevAfkHome: string | undefined;
+let prevHome: string | undefined;
+
+function setupAuditEnv(): void {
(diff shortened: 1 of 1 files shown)


griffinwork40 (commented):

---

## Prior Automated Rounds

No prior automated rounds on this PR. (Round 1 of 2 maximum.)

---

## Review: fix(security): reconcile grant-manager divergences (#500)

**PR #1479** · Branch: `fix/grant-manager-divergence-500` · Head: `01231262c223466cf7727719f9ebe8a4ecd58047`  
**Change:** New test file only — `src/agent/tools/grant-manager.test.ts` (+470 / −0)  
**Regime:** Full (>300 lines)

---

## Verdict

**Decision: MERGE (with follow-ups) — 0 critical, 0 high → no blocking findings; 6 low advisory, 1 nit advisory.**

Arithmetic: 0 critical, 0 high → MERGE (with follow-ups). All low/nit findings are non-blocking and listed as follow-ups below. A follow-up issue is suggested rather than fix commits.

---

## Spec-Compliance

Stated intent: PR #1479 title+body — spec-compliance assessed.

✅ **No unmet requirements.** All three divergence findings from issue #500 are covered:
- Finding 1 (per-call `sessionId` threading): 5 audit tests cover wins-over-hook, hook-fallback, null-when-neither, threading through `addWriteRoot`, threading through `revokeRoot`.
- Finding 2 (migrating anchor): behavioral tests cover anchor protection, revoke-allowed after migration, and uninitialized-provider no-op. Freshness invariant proven via behavioral outcome (migration test would fail if value were cached at construction).
- Finding 3 (no-op revoke audit gate): 4 audit tests cover path-never-granted, double-revoke, protected-root guard, and uninitialized-provider cases.

✅ **No scope creep.** One file added (`grant-manager.test.ts`); zero implementation files touched.

---

## Advisory Follow-ups (all non-blocking — suggest a single follow-up issue)

### LOW-1 · AFK_STATE_DIR not cleared in `setupAuditEnv`
**File:** `src/agent/tools/grant-manager.test.ts:69–74`  
`setupAuditEnv` saves/restores `AFK_HOME` and `HOME` but does not clear `AFK_STATE_DIR`. When `AFK_STATE_DIR` is set in the caller's shell, `getAfkStateDir()` ignores the `AFK_HOME` override and audit writes land in the developer's real state directory. **Mitigating context:** the global `redirect-paths-env.ts` setup file deletes `AFK_STATE_DIR` at module-eval time, so this gap is unreachable in `pnpm test`. The gap only surfaces if the file is run outside the standard harness.  
**Fix:** In `setupAuditEnv`, add `prevAfkStateDir = process.env['AFK_STATE_DIR']; delete process.env['AFK_STATE_DIR'];` and restore symmetrically in `teardownAuditEnv`.

### LOW-2 · `tmpHome` uninitialized at module scope
**File:** `src/agent/tools/grant-manager.test.ts:65`  
`let tmpHome: string` is declared with no initializer. If `readAuditEntries()` were called from any test outside the audit `describe` block (before `setupAuditEnv` runs), `path.join(undefined, ...)` throws a `TypeError` instead of a clear assertion failure. Current call sites are all inside the audit `describe` block, so this is unexploited today.  
**Fix:** Initialize to a sentinel (`let tmpHome = ''`) and add a guard in `readAuditEntries()`: `if (!tmpHome) return [];`.

### LOW-3 · Vacuous `path.resolve` test (absolute path is a tautology)
**File:** `src/agent/tools/grant-manager.test.ts:165–171`  
The test `'resolves the path to absolute before storing'` passes `/abs/path` (already absolute), so `path.resolve('/abs/path') === '/abs/path'` is always true regardless of whether `path.resolve` is called. A regression in the resolution step on relative inputs would not be caught.  
**Fix:** Use a relative path in the assertion: `gm.addReadRoot('relative/sub', 'slash'); expect(hooks._readRoots[0]).toBe(path.resolve('relative/sub'));` — or rename the test to `'stores path as-is when already absolute'` to match what it actually asserts.

### LOW-4 · `toBeGreaterThanOrEqual(1)` masks double-emit regressions
**File:** `src/agent/tools/grant-manager.test.ts:325–332`  
The two per-call `sessionId` tests for `addReadRoot` and `addWriteRoot` assert `entries.length >= 1` and inspect only the last entry. Since each test runs in a fresh `tmpHome` with one state-changing call, the count is deterministically 1 — but a double-emit regression (from a future change to the dedup guard) would produce 2 entries and still pass this check.  
**Fix:** Change both assertions to `expect(entries.length).toBe(1)`.

### LOW-5 · Test label attributes Finding 2 audit behavior to Finding 3
**File:** `src/agent/tools/grant-manager.test.ts:405–415`  
The test `'(Finding 3) emits NO entry when revoking the protected root (guard fires first)'` is labelled Finding 3, but the zero-audit outcome is caused by the Finding 2 early-return guard (`if (protectedRoot !== undefined && p === protectedRoot) return`) — not by Finding 3's `rRemoved || wRemoved` gate. The parenthetical `"guard fires before removal"` in the test's own comment confirms it is a Finding 2 observable. If Finding 3's gate is reverted but Finding 2's guard is intact, this test still passes.  
**Fix:** Relabel to `'(Finding 2) emits NO audit entry when revoking the protected root (early-return guard)'`.

### LOW-6 · Implicit read-grant from `addWriteRoot` on fresh path is untested explicitly
**File:** `src/agent/tools/grant-manager.test.ts:450–470`  
`addWriteRoot` on a path not yet in `readRoots` pushes to `readRoots` silently (no `grant-read` audit) and emits only `grant-write`. This deliberate design decision is backstopped only incidentally by the schema test's length assertion (`entries.length === 3`). No test is explicitly named for this invariant.  
**Fix:** Add `'addWriteRoot on a path not yet granted as read emits exactly grant-write (implicit read push is not audited)'` asserting `entries.map(e => e['action'])` equals `['grant-write']`.

### NIT-1 · `HOME` redirect in `setupAuditEnv` is redundant
**File:** `src/agent/tools/grant-manager.test.ts:69–74`  
`getAfkHome()` checks `AFK_HOME` first and returns immediately when it is set; the `HOME` assignment is only consulted in the `AFK_HOME`-unset fallback branch, which is never reached here. `HOME` is set, saved, and restored unnecessarily.  
**Fix:** Remove the `HOME` override and its corresponding save/restore from `setupAuditEnv`/`teardownAuditEnv`. If the intent is defensive isolation, add a comment explaining why.

---

## What Was Not Checked

- Citations verified inline against branch HEAD `01231262c223466cf7727719f9ebe8a4ecd58047`. All citations confirmed as `diff-context` (lines present in the diff hunk); no fabricated citations found.
- No implementation files changed in this PR — implementation correctness for the three divergence fixes was reviewed in #1396 (already on main).
- Runtime behavior of the new tests was not executed (review is read-only). CI results for this PR show all jobs passing or in-progress.
- Did not review the Telegram or browser automation surfaces.
- Stated intent: PR #1479 title+body — spec-compliance assessed.

griffinwork40 (commented):

---

## Review: fix(security): reconcile grant-manager divergences (#500)

**PR #1479** · Branch: `fix/grant-manager-divergence-500` · Head: `17b8cdf6ab379c4b37a091472c25ee6e1a9331f3`
**Change:** 1 new file (`grant-manager.test.ts`, +498 lines)
**Regime:** full

---

## Prior Automated Rounds

One prior round on commit `01231262` — 6 low, 1 nit (all non-blocking). The second commit (`17b8cdf6`) addressed all 6 low findings from that round (AFK_STATE_DIR isolation, tmpHome init, relative-path assertion, ensureInitialized lazy-init, per-call sessionId assertion tightening, Finding 2 label correction, addWriteRoot grant-only audit test). This is Round 2.

---

## Verdict

**Decision: MERGE (with follow-ups) — 0 critical, 0 high; 2 medium waived (1 correctness, 1 test-coverage); 7 low, 1 nit.**

---

## Spec-Compliance

**Stated intent:** PR #1479 title + body — spec-compliance assessed.

✅ **No unmet requirements.** All three divergence findings from #500 are covered:
- Finding 1 (per-call `sessionId` threading): 5 tests cover per-call wins, hook fallback, null-when-neither, threading through `addWriteRoot` and `revokeRoot`.
- Finding 2 (migrating anchor / Option A): 3 tests cover anchor protection, revoke-allowed after migration, uninit-provider no-op.
- Finding 3 (no-op revoke audit gate): 4 tests cover never-granted, double-revoke, protected-root guard, uninit-provider.

✅ **No scope creep.** Additional tests (dedup invariant, audit schema, getGrants snapshot, defensive copies) are reasonable supporting coverage for a grant-manager test suite.

---

## Findings

### F1 · medium · blocking:false · confidence:high · correctness · `src/agent/tools/grant-manager.test.ts:251–258` · ref:17b8cdf6 · diff-context · waived: all three real consumers (dispatcher, anthropic-direct, openai-compat) supply `getProtectedRoot()` from `resolveBase`/`_currentCwd` which is always an absolute path — the unresolved-protectedRoot edge case is unreachable in production, making this a follow-up documentation issue, not a merge-blocking defect

The Finding 2 tests compare `p === protectedRoot` where `p = path.resolve(absPath)` but `protectedRoot = hooks.getProtectedRoot()` with no `path.resolve` applied. All test fixtures use absolute string literals (`'/anchor'`, `'/old-anchor'`), so the guard is never exercised with a relative `protectedRoot`. If a consumer's `getProtectedRoot()` returned a relative path, the guard would silently fail to protect it.

**Suggestion:** Add a test where `getProtectedRoot()` returns a relative path, or add a JSDoc contract requiring absolute return values.

---

### F2 · medium · blocking:false · confidence:high · test-coverage · `src/agent/tools/grant-manager.test.ts` · ref:17b8cdf6 · diff-context · waived: `addWriteRoot` calls `ensureInitialized?.()` identically to `addReadRoot` (same code path); the mechanism is proven by the `addReadRoot` test at line 143, and the gap is bounded to a single method with byte-identical invocation

No test exercises the `ensureInitialized` interaction with `addWriteRoot`. The `addReadRoot` describe has a dedicated test (line 143) proving `ensureInitialized` is called and arrays are populated. `addWriteRoot` makes the same call but has no corresponding test.

**Suggestion:** Add a parallel `ensureInitialized` test in the `addWriteRoot` describe block.

---

### F3 · low · blocking:false · confidence:high · security · `src/agent/tools/grant-manager.test.ts:63–91` · ref:17b8cdf6 · diff-context

`tmpHome` is initialized to `''` at module scope (line 63). If `mkdtempSync` inside `setupAuditEnv` were to throw, `tmpHome` would remain `''`; `rmSync(tmpHome, { recursive: true, force: true })` would resolve `''` → `process.cwd()`. This path is **not reachable** under current control flow (`setupAuditEnv` is called outside `try` blocks or in `beforeEach` which skips `afterEach` on failure), but it is a latent structural hazard.

**Suggestion:** Guard: `if (tmpHome) rmSync(tmpHome, ...)`.

---

### F4 · low · blocking:false · confidence:high · security · `src/agent/tools/grant-manager.test.ts:70–91` · ref:17b8cdf6 · diff-context

`setupAuditEnv` overrides `process.env['HOME']` in addition to `AFK_HOME`. Since `getAfkHome()` checks `AFK_HOME` first (and `setupAuditEnv` always sets it), the `HOME` mutation is redundant — it is never consulted by `getSessionGrantsPath()`.

**Suggestion:** Remove the `HOME` manipulation from `setupAuditEnv`/`teardownAuditEnv`.

---

### F5 · low · blocking:false · confidence:high · correctness · `src/agent/tools/grant-manager.test.ts:112–117` · ref:17b8cdf6 · diff-context

The first `addReadRoot` test passes an absolute literal (`'/some/path'`), making `path.resolve` an identity — the assertion is tautological for resolution logic. The meaningful relative-path test already exists at line 128.

**Suggestion:** Convert to a relative input or annotate as an identity-case test.

---

### F6 · low · blocking:false · confidence:high · correctness · `src/agent/tools/grant-manager.test.ts:195–200` · ref:17b8cdf6 · diff-context

`addWriteRoot` lacks a relative-path resolution test (same tautology as F5). Only `addReadRoot` has one (line 128).

**Suggestion:** Mirror the relative-path test from `addReadRoot`.

---

### F7 · low · blocking:false · confidence:high · correctness · `src/agent/tools/grant-manager.test.ts:303–310` · ref:17b8cdf6 · diff-context

The defensive-copy test validates caller→snapshot isolation (pushing into a snapshot doesn't affect the next call) but not source→snapshot isolation (mutating the underlying `hooks._readRoots` after taking a snapshot).

**Suggestion:** Add a complement: mutate `hooks._readRoots` after `getGrants()` and verify the snapshot is unaffected.

---

### F8 · low · blocking:false · confidence:high · spec-compliance · `src/agent/tools/grant-manager.test.ts:437` · ref:17b8cdf6 · diff-context

Test at line 437 is labelled `(Finding 2)` but lives inside the describe titled `"audit log (Finding 1: sessionId, Finding 3: no-op gate)"`.

**Suggestion:** Update the describe title to `"audit log (Findings 1, 2, 3)"`.

---

### F9 · low · blocking:false · confidence:high · test-coverage · `src/agent/tools/grant-manager.test.ts` · ref:17b8cdf6 · diff-context

No test exercises `getGrants()` when `getReadRoots()` returns `undefined`. The `?? []` fallback (grant-manager.ts:198) is uncovered.

**Suggestion:** Add one test with the uninit hooks fixture asserting `getGrants().readRoots` is `[]`.

---

### F10 · nit · blocking:false · confidence:high · api-compat · `src/agent/tools/grant-manager.test.ts:26–27` · ref:17b8cdf6 · diff-context

`PathGrantManagerHooks` (imported as `import type`) has zero production importers — it is referenced only in `grant-manager.ts` internals and this test file. A future rename would break only this test. [UNVERIFIED: no production importers]

---

## Dimensions With No Findings

- **API compatibility:** no issues found — single new test file, zero implementation changes, all imports are publicly exported symbols.
- **Perf / observability:** no issues found — `mkdtempSync`/`rmSync` per-test lifecycle is correct and proportionate (11 temp dirs, <1ms each). `readAuditEntries` I/O volume is negligible.

## CI Status

All CI checks passing. No failures on this PR.

## What Was Not Checked

- Citations verified against branch HEAD `17b8cdf6ab379c4b37a091472c25ee6e1a9331f3`.
- Stated intent: PR #1479 title + body — spec-compliance assessed.
- Did not run tests locally.
- Did not audit the pre-existing grant-manager implementation beyond the specific lines cited.
- Did not verify test runtime under `vitest --sequence.shuffle` for shared-mutable-state safety.

PR #1482, opened 2026-09-03
feat(tools): structured JSON query tool (#1423)

### LINKED ISSUE #1423: feat(tools): structured JSON query tool -- bounded jq-subset queries without loading full files into context
## Problem (lower priority)

The agent has no structured way to query large JSON documents without loading the entire content into model context. For a 500KB JSON config or a 2MB API response, the agent must either `read_file` (loading it all into context) or `bash: cat file.json | jq '.path.to.field'` (shelling out to jq, which is an external dependency with no structured result).

### User impact

- Large JSON files (package-lock.json, API responses, terraform state, test fixtures) consume thousands of tokens when loaded via `read_file`. A targeted query for one field would use a fraction of the context.
- `jq` expressions are error-prone for models and vary across jq versions. A structured query tool with a simpler syntax would reduce errors.
- `extract_document` handles JSON inside ZIP files (`src/agent/tools/handlers/extract-document.ts:209`) but returns the full text, not a targeted query.

## Verified current behavior

| Component | File | How JSON is handled |
|-----------|------|-------------------|
| read_file | `src/agent/tools/handlers/read-file.ts` | Returns full file content with line numbers. No query capability |
| bash + jq | (external dependency) | Works if jq is installed. `src/agent/tools/readonly-bash.test.ts:464,702` confirms jq is in the readonly-bash allowlist |
| extract_document | `src/agent/tools/handlers/extract-document.ts:209` | Extracts JSON from ZIP as raw text. No query |
| gh --jq | `src/agent/gh.ts:366` | Internal use of `gh pr view --json X --jq .X`. Not agent-facing |
| Trace/witness CLI | `src/cli/commands/trace.ts:8,497,741` | `--json` flag for piping to jq. User-facing, not agent-facing |

## Proposed minimum behavior

A `json_query` tool that runs bounded queries on JSON files without loading the full content into model context:

**Schema:**
- `path` (required): file path to the JSON document
- `query` (required): query expression (see syntax below)
- `max_results` (optional): cap on returned array elements (default: 100)
- `max_bytes` (optional): cap on returned content (default: 50KB)

**Query syntax (subset of jq / JSONPath):**
- `.field` -- access object field
- `.field.nested` -- nested field access
- `.[N]` -- array index
- `.[N:M]` -- array slice
- `.[] | .field` -- map over array, extract field
- `keys` -- list object keys
- `length` -- count array elements or object keys
- `select(.field == "value")` -- filter array elements

This is deliberately a subset. The tool is not a full jq implementation -- it handles the 80% case of "extract a few fields from a large JSON file" without the complexity of jq's full language.

**Result shape:**
- `result`: the query result (JSON value, truncated to max_bytes)
- `type`: the JSON type of the result (object, array, string, number, boolean, null)
- `truncated`: whether the result was truncated
- `source_size`: total size of the source file in bytes

- The query must be bounded: `max_results` and `max_bytes` prevent a query from returning a 50MB array into model context.
- JSON parsing must handle malformed files gracefully (report parse error with location, don't crash).
- Path traversal in the `path` parameter must go through the same grant-manager checks as `read_file`.
- The query language must be simple enough that models can generate correct queries without extensive examples. A complex jq-like language defeats the purpose.

## Value vs. existing bash + jq workaround

| Aspect | bash + jq | json_query tool |
|--------|-----------|----------------|
| Availability | Requires jq installed | Built-in |
| Structured result | String stdout | Typed JSON |
| Size bounds | Unbounded (can OOM on huge files) | max_results + max_bytes caps |
| Error handling | Exit code + stderr | Structured error with parse location |
| Path containment | None (bash bypasses grant-manager) | Goes through read_file permissions |
| Model accuracy | jq syntax is error-prone | Simpler subset |

The value add is moderate: jq works today and is in the readonly-bash allowlist. The improvement is structured results, bounded output, and simpler syntax. This is a convenience improvement, not a capability gap.

- **bash + jq:** Works today. jq is in the readonly-bash allowlist. The main downsides are unstructured output and unbounded result size.
- **bash + python -c:** Works for simple queries. More verbose but the model is good at generating Python.
- **read_file with offset/limit:** Works for small files or when the target is at a known line number. Doesn't help for deeply nested fields.
- **grep for a key:** Works for finding a field's location, then `read_file` with offset to get context. Two-step but effective.

## Non-goals

- Full jq compatibility (the query language is a deliberate subset)
- JSON modification (this is read-only)
- Streaming JSON parsing (NDJSON, JSON Lines) -- different use case
- YAML, TOML, or other config format queries (JSON only in v1)
- Schema validation (separate concern)

## Acceptance criteria

- [ ] Agent can query a JSON file by field path, array index, and simple filter without loading the full file into context
- [ ] Results are bounded by configurable max_results and max_bytes
- [ ] Malformed JSON files produce a structured error with parse location
- [ ] Path access goes through the same grant-manager permissions as read_file
- [ ] The query syntax is simple enough that models generate correct queries in >90% of cases

- Reuses: `src/agent/grant-manager.ts` (path permissions), `src/agent/tools/handlers/read-file.ts` (file access patterns)
- No dependencies on other proposed primitives

## Priority

**P3** -- Lower priority convenience feature. bash + jq is an adequate workaround that's already available in the readonly-bash allowlist. The improvement is structured output and bounded queries, not new capability.

## Open design questions

1. Should the query language be a jq subset, JSONPath, or a custom syntax? jq is familiar but complex; JSONPath is standard but less powerful; a custom syntax could be optimized for model accuracy.
2. Should the tool support in-memory queries on tool results (e.g., query the output of a previous `web_scrape` call) in addition to file queries?
3. Should large JSON files be indexed on first access (build a key-path index) for faster repeated queries, or re-parsed each time?
4. Is this valuable enough to justify the implementation and maintenance cost, given that bash + jq works today?

griffinwork40:
Closes #1423

## Summary

Adds a `json_query` tool that runs bounded queries on JSON files without loading the full content into model context. Uses a pure TypeScript jq-subset interpreter — no external dependencies.

## Files

| File | Purpose |
|---|---|
| `src/agent/tools/handlers/json-query.ts` | Handler + query engine |
| `src/agent/tools/schemas.json-query.ts` | Tool schema (extracted per 350-line ratchet) |
| `src/agent/tools/handlers/index.ts` | Handler registration |
| `src/agent/tools/schemas.ts` | Schema import/re-export + array entry |
| `src/agent/tools/handlers/json-query.test.ts` | 39 tests |
| `.filesize-baseline.json` | Baseline update for schemas.ts |

## Schema

```
path    (required): file path
query   (required): jq-subset expression
max_results (default 100): cap on array elements returned
max_bytes   (default 50 KB): cap on serialized output
```

Result shape: `{ result, type, truncated, source_size }`

## Query syntax

- `.` — whole document
- `.field` / `.field.nested` — field access
- `.[N]` — array index (negative: from end)
- `.[N:M]` — array slice
- `.[]` — iterate array
- `.[] | .field` — map-extract
- `keys` — object keys / array indices
- `length` — element/key/char count

## Edge cases handled

- Empty files → `{ result: null, type: "null", … }`
- Malformed JSON → structured error with parse location
- Files > 10 MB → early error before parse
- Missing object keys → `null` result (no error)
- Out-of-bounds array index → `null` result (no error)
- `undefined` values normalized to `null` in result

## Tests

39 tests pass across: input validation, FS edge cases, identity/field/index/slice/iter/keys/length queries, truncation caps, parse errors, cwd factory.


griffinwork40 (commented):

---

## PR #1482 Review — feat(tools): structured JSON query tool (#1423)

**Target:** `feat/json-query-1423` → `main` | **Stats:** +1100 / -6 | 10 files (2 new) | **Regime:** full
**Reviewed ref:** `1855bda10a749d47195e664d447c0cb264fd3881` | **Stated intent:** PR #1482 title + body

---

### Findings

#### 1. HIGH · blocking:true · confidence:verified · security · `src/agent/tools/hooks/path-approval-hook.ts:509-512`
**ref:** 1855bda10a · citation: file-state (verified against reviewed ref)

**Finding:** `extractCandidatePath` reads `input['file_path']` for the `json_query` case, but both the schema (`schemas.json-query.ts:36-38`, property name `path`, required: `['path', 'query']`) and the handler (`json-query.ts:260`, `obj['path']`) use the field name `path`. `extractCandidatePath` therefore always returns `undefined` for `json_query` calls, and the path-approval hook silently skips the out-of-root operator approval prompt. The model can query any JSON file outside the session's granted roots without triggering the operator's confirmation dialog.

**Evidence:**
```ts
// path-approval-hook.ts:509-512 — reads wrong field
if (toolName === 'json_query') {
  const p = input['file_path'];  // ← 'file_path' does not exist in the schema
  return typeof p === 'string' ? p : undefined;
}

// schemas.json-query.ts:36 — schema uses 'path'
path: { type: 'string', description: '...' }

// json-query.ts:260 — handler uses 'path'
const rawPath = obj['path'];
```

**Mitigating context:** `resolveAndContain` still runs inside the handler itself (`json-query.ts:285`), so the handler's own containment guard for the session's context roots applies. The bypass is specifically of the *operator-facing approval prompt* that fires before handler execution for paths outside granted roots.

**Suggestion:** Change `input['file_path']` to `input['path']` in the `json_query` case of `extractCandidatePath`. Add a regression test asserting `extractCandidatePath('json_query', { path: '/outside/root.json', query: '.' })` returns `'/outside/root.json'`.

---

#### 2. MEDIUM · blocking:false · confidence:high · correctness · `src/agent/tools/handlers/json-query.ts:388-400`
waived: bounded, non-data-affecting — truncation occurs only when output exceeds max_bytes; the corrupted replacement character affects only the trailing edge of an already-truncated string, which is appended with `\n… [truncated]` so downstream consumers already treat it as incomplete

**Finding:** `Buffer.from(serialized, 'utf-8').slice(0, maxBytes).toString('utf-8')` truncates at an arbitrary byte boundary. When the cut point falls inside a multi-byte UTF-8 sequence, `.toString('utf-8')` emits a U+FFFD replacement character for the incomplete trailing sequence.

**Evidence:**
```ts
const buf = Buffer.from(serialized, 'utf-8');
resultPayload = buf.slice(0, maxBytes).toString('utf-8') + '\n… [truncated]';
```

**Suggestion:** Walk backward from the slice boundary past continuation bytes before converting:
```ts
let end = maxBytes;
while (end > 0 && (buf[end] & 0xC0) === 0x80) end--;
resultPayload = buf.slice(0, end).toString('utf-8') + '\n… [truncated]';
```

---

#### 3. MEDIUM · blocking:false · confidence:high · spec-compliance · `src/agent/tools/handlers/json-query.ts:191-193`
waived: bounded, non-data-affecting — jq throws a controlled Query error rather than returning null; no data loss, and the isError response path handles it cleanly

**Finding:** Field access on a `null` intermediate throws a controlled `Query error` (via `throw new Error('Cannot access field ... on null')`) instead of propagating `null` as jq specifies. For `{"a":{"b":null}}`, the query `.a.b.c` returns `{isError: true}` rather than `{result: null}`. The jq specification mandates `null | .foo => null` (null propagation).

**Evidence:**
```ts
case 'field': {
  if (value === null || typeof value !== 'object' || Array.isArray(value)) {
    throw new Error(`Cannot access field ".${token.name}" on ${typeName(value)}`);
  }
```

**Suggestion:** Return `[null]` instead of throwing when `value` is `null` — preserving the throw for non-null non-objects (numbers, booleans, strings).

---

#### 4. LOW · blocking:false · confidence:medium · correctness · `src/agent/tools/handlers/json-query.ts` (evalQuery iter path)

**Finding:** `.[]` applied to an object uses the `iter` case which checks `if (!Array.isArray(value)) throw new Error(...)`. jq supports `.[]` on objects (iterating values in insertion order). Objects passed to `.[]` produce an error instead of iterating values.

**Suggestion:** Add an `Object.values()` path in the `iter` case for non-array objects.

---

#### 5. LOW · blocking:false · confidence:high · test-coverage · (test suite)

**Finding:** Three confirmed coverage gaps in the 39-test suite: (a) no test exercises the `extractCandidatePath` integration for `json_query` — the `file_path`/`path` mismatch (finding 1) is uncaught; (b) no test verifies UTF-8 boundary handling in the byte-truncation path; (c) no test exercises `resolveAndContain` containment rejection for out-of-root paths.

**Suggestion:** Add: (a) a hook-level test asserting `extractCandidatePath('json_query', {path: '/x.json', query: '.'})` returns `'/x.json'`; (b) a truncation test with a multi-byte character at the boundary; (c) a `createJsonQueryHandler('/safe')` test with a path like `'/etc/passwd'`.

---

#### 6. NIT · blocking:false · confidence:high · correctness · `src/agent/tools/handlers/json-query.ts` (typeName helper)

**Finding:** `typeName(null)` correctly returns `'null'`, but `typeName([1,2])` returns `'object'` (via `typeof`). The `type` field in the response uses `typeName(finalResult)`, so array results are reported as `type: 'object'` — however, the result-shaping block at line 362-386 already explicitly checks `Array.isArray` and the `isIterQuery` flag, so the `type` field is overridden to `'array'` for known-array paths. The mismatch only surfaces for edge cases where the result is unexpectedly an array.

**Suggestion:** Update `typeName`: `if (Array.isArray(v)) return 'array'; if (v === null) return 'null'; return typeof v;`.

---

The following Wave 1 findings were dropped after inline verification against the reviewed ref:

- **"`undefined` result silently drops `result` key from output JSON"** — **false**: the code at line 392 already normalizes `undefined` → `null` explicitly (`finalResult === undefined ? 'null' : JSON.stringify(...)` and `finalResult === undefined ? null : finalResult`).
- **"`keys` on array returns strings instead of integers"** — **false**: the code at line 176 uses `value.map((_, i) => i)` which returns numbers, not `Object.keys()`.
- **"`length` on null throws TypeError"** — **false**: the code at line 189 explicitly handles `if (value === null) return [0]`.
- **"ReDoS in query-string tokenizer"** — **false**: all regexes are anchored (`^...$`) with non-backtracking quantifiers; `findTopLevelPipe` is a linear O(n) character scan.
- **"Hardcoded test counts break CI"** — **false**: the PR already updated `schemas.test.ts` to `33`, `index.test.ts` to `32`, and added `'json_query'` to expected names.

---

### Decision

**❌ DO NOT MERGE — 1 high blocking (security: path-approval hook bypass); 2 medium waived; 2 low; 1 nit.**

The blocking finding is the field-name mismatch in `extractCandidatePath` — a one-line fix (`'file_path'` → `'path'`). Once fixed, the remaining findings are non-blocking follow-ups.

---

### What was not checked

- Citations verified inline against branch HEAD `1855bda10a749d47195e664d447c0cb264fd3881`.
- Stated intent: PR #1482 title+body — spec-compliance assessed.
- **Not checked:** Runtime integration test execution — timing claims verified by static analysis only.
- **Not checked:** `--format stream-json` consumer behavior with the new `json_query` event type.
- **Not checked:** Interaction with `nesting.ts` CHILD_ALLOWED_TOOLS / RECON_ALLOWED_TOOLS (json_query not present in diff; subagent visibility not verified).
- **Not checked:** Dispatch-batching `SAFE_TOOLS` inclusion.
- CI status: "Lint & Build" is FAILING on this PR. Tests pass on ubuntu-latest and macos-latest.

griffinwork40 (commented):

---

## PR #1482 Round 2 Review — feat(tools): structured JSON query tool (#1423)

**Target:** `feat/json-query-1423` → `main` | **Stats:** +1100 / -6 | 10 files (2 new) | **Regime:** full  
**Reviewed ref:** `40937128a4720796981e25f449117619331a9792` | **Round:** 2 of 2

---

## Prior-Round Dispositions (Round 1 carry-forward)

| ID | Original Severity | Disposition |
|----|-----------------|-------------|
| R1-HIGH · path-approval hook field name mismatch (`file_path` → `path`) | HIGH · was blocking | ✅ **FIXED** — `fix(pr-1482): fix json_query path-approval field name (file_path → path)` (commit `40937128`). `extractCandidatePath` now reads `input['path']` for `json_query` at `path-approval-hook.ts:487`. Regression tests added in `path-approval-hook.test.ts:967-1012`. Verified at reviewed ref. |
| R1-MEDIUM-F2 · UTF-8 byte-boundary truncation at `max_bytes` cut | MEDIUM (waived) | **STILL PRESENT — waiver holds.** Code unchanged at `json-query.ts:395`. Bounded, cosmetic, non-data-affecting; `truncated: true` is set. |
| R1-MEDIUM-F3 · null field access throws Query error instead of propagating null | MEDIUM (waived) | **STILL PRESENT — waiver holds.** Documented jq-subset deviation; `field` case at `json-query.ts:192` throws for `null` inputs. Controlled error path, no data loss. |
| R1-LOW-F4 · `.[]` on objects throws instead of iterating values | LOW | **STILL PRESENT — not fixed.** `iter` case at `json-query.ts:217-220` remains array-only. Documented subset restriction; schema docstring lists only array iteration. |
| R1-LOW-F5 · Coverage gaps | LOW | **PARTIALLY CLOSED.** 2 of 4 sub-gaps fixed: `path-approval-hook.test.ts:967-1012` adds regression tests confirming `extractCandidatePath` reads `input['path']` for out-of-root and in-root paths. Remaining open: no UTF-8 boundary test for `max_bytes` truncation; no `resolveAndContain` rejection test in `json-query.test.ts`. |

---

## Decision (with blocking-severity floor applied)

**Blocking severity floor:** Only `critical` and `high` findings may block merge. `medium` and below are advisory regardless of the `/review` tool's internal verdict.

**Arithmetic:** 0 critical · 0 high → **✅ MERGE (with follow-ups)**; 1 medium advisory (see NEW-F1 below); 4 low advisory; 1 nit.

> _Note: The `/review` synthesis emitted DO NOT MERGE on the medium finding (NEW-F1). The scheduled-sweep blocking floor overrides this: medium findings are listed as advisory follow-ups below and do not gate merge._

---

## New Findings — Advisory (non-blocking)

### NEW-F1 · MEDIUM advisory · api-compat · `src/agent/tools/nesting.ts:163-180`
**`json_query` absent from `RECON_ALLOWED_TOOLS`**

`RECON_ALLOWED_TOOLS` is an explicit allowlist used for read-only skill children (`research-agent`, `Explore`, etc.). It contains `read_file`, `glob`, `grep`, `list_directory` — but not `json_query`. `json_query` is in `CHILD_ALLOWED_TOOLS` (via `BUILTIN_TOOL_NAMES`) so normal children can use it, but read-only skill children calling `json_query` receive `tool-not-allowed` at runtime.

`json_query` is correctly classified as `category: 'read'` in the schema, is in `READ_TOOLS` (tool-category.ts:66), and is in `READ_ONLY_PHASE_TOOLS` (tool-category.ts:258). Adding it to `RECON_ALLOWED_TOOLS` is a one-line fix:

```typescript
// nesting.ts ~line 167
export const RECON_ALLOWED_TOOLS: readonly string[] = [
  'read_file',
  'glob',
  'grep',
  'list_directory',
+ 'json_query',   // read-only JSON query — same classification as read_file/glob
  'config_get',
  ...
];
```

Suggest filing a follow-up issue rather than blocking merge on this — the gap only bites read-only skills, which are not yet using `json_query`, and the fix is trivial.

---

### NEW-F2 · LOW advisory · correctness · `src/agent/tools/handlers/json-query.ts:402-407`
**`type` field misreports envelope type after byte-truncation**

When `max_bytes` truncation fires, `resultPayload` is set to a `string` (the truncated JSON fragment + `\n… [truncated]`), but `type` is computed from `typeName(finalResult)` before truncation, producing e.g. `type: "object"` while `result` is a string. The mismatch only surfaces when `truncated: true`. No existing test asserts `type` in the truncated branch, so CI does not catch it.

**Suggestion (follow-up):** In the truncation branch, set `type: 'string'` or compute type from `typeof resultPayload`.

---

### NEW-F3 · LOW advisory · test-coverage · `src/agent/tools/handlers/json-query.test.ts:122-133`
**Oversized-file test is a placeholder passing on the success path**

The test named `'returns error for oversized file'` explicitly skips exercising the `>MAX_FILE_SIZE` guard: its inline comment says "we skip it in CI and trust the branch," and the test asserts `type: 'object', truncated: false` — the success path. The `>10MB` guard at `json-query.ts:~310` has no CI-executed coverage.

**Suggestion (follow-up):** Replace with a `vi.spyOn(fs, 'stat').mockResolvedValueOnce({ size: 11*1024*1024 })` test that asserts `isError: true`.

---

### NIT · `path-approval-hook.ts` comment update
After the fix, the combined `list_directory || glob || grep || json_query` branch (line 487) has a comment that only documents the "optional path → cwd fallback" rationale for glob/grep. It is now slightly misleading for `json_query` where `path` is `required` by schema. Updating the comment to note that `undefined` here means a schema-validation failure that will be caught before the handler runs would improve clarity.

---

## What Was Not Checked

- Citations verified against branch HEAD `40937128a4720796981e25f449117619331a9792`.
- Stated intent: PR #1482 title+body — spec-compliance assessed (PR delivers the stated `json_query` tool with the documented schema and query subset).
- **Not checked:** Runtime integration test execution — behavioral claims verified by static analysis only.
- **Not checked:** `CHILD_ALLOWED_TOOLS` interaction with dispatch-batching `SAFE_TOOLS` set.
- **Not checked:** Windows / non-POSIX path normalization through `resolveAndContain`.
- **Not checked:** Whether read-only skills currently deployed reference `json_query` (the NEW-F1 gap is latent until they do).
- CI status: one in-progress check (`Test (macos-latest)`) at enumeration time; all others SUCCESS. Merge conflict (`CONFLICTING`) is present — rebase needed before merge in addition to the follow-up items above.

Issue #1416, opened 2026-09-01
feat(core): durable human handoff -- persist questions and approvals across process restarts


griffinwork40:
## Problem

When the agent needs human input mid-task (approval, clarification, decision), the question exists only in process memory. If the process restarts -- daemon upgrade, OOM, crash -- the pending question is silently lost, the human's eventual reply has nowhere to land, and the interrupted task cannot resume.

Every elicitation surface (REPL, Telegram, web UI) stores pending questions in in-memory Maps or Promise chains with no durable backing.

### User impact

- A daemon task that calls `ask_question` and then the daemon restarts: the question is gone, the dequeued task file is already deleted (loss window from queue-store), and the operator sees nothing to act on.
- A Telegram user who receives a question, takes an hour to decide, and replies after a bot restart: the reply falls through to normal message handling and starts a new session turn instead of resolving the original question.
- Web UI approvals pending at `POST /api/approve` disappear on server restart; the frontend's "pending approvals" list renders empty.

## Verified current behavior

| Component | File | State durability |
|-----------|------|-----------------|
| Elicitation router | `src/agent/elicitation-router.ts:111,134` | Module-scope singleton. `pendingCount()` tracks in-memory Promises. Abort signal is the turn's AbortSignal; restart severs it |
| ask_question gate | `src/agent/tools/handlers/ask-question.ts`, `ask-question-gate.ts:20,89` | Routes through elicitation router. Non-interactive surfaces decline immediately. No persistence of the pending question |
| Telegram elicitation | `src/telegram/elicitation-handler.ts:106-107,239` | `pendingChoiceElicitations`: in-memory `Map<string, {chatId, resolve, reject}>`. Bot restart empties it |
| Telegram route registry | `src/telegram/elicitation-route-registry.ts:28` | In-memory `Map` -- the canonical Telegram route store |
| Web elicitation bridge | `src/web-server/elicitation-web.ts`, `src/web-server/routes.ts:75,232,303` | Process-global singleton. `POST /api/approve` resolves by `requestId`. Restart empties all pending |
| Daemon queue | `src/agent/daemon/queue-store.ts:122-165` | Task file deleted before execution. A task calling ask_question mid-run cannot be retried if the daemon restarts |

**No `elicitations/`, `pending-questions/`, or `handoffs/` directory exists under `~/.afk/state/`.**

## Proposed minimum behavior

A durable handoff store that persists questions and approvals across process restarts:

**Handoff lifecycle:**
[...]

PR #1484, opened 2026-09-03
feat(core): add HandoffRecord schema and file-based handoff store (#1416)

griffinwork40:
Part 1 of 5 for #1416 — purely additive, zero behavior change. Nothing in the existing system calls into these yet; wiring begins in PR 2.

### `src/agent/daemon/task-lifecycle.ts`
- Adds `'waiting_human_input'` to the `TaskState` union. This state means "task is leased but blocked on a human elicitation; do not expire the lease." PR 2 will wire it into `recoverExpiredLeases()`.

### `src/paths.ts`
- Adds `getHandoffsDir()` — returns `<stateDir>/handoffs/`. Follows the exact pattern of `getQueueDir()`, `getPresenceDir()`, etc.

### `src/agent/daemon/handoff-store.ts` (new file)
- `HandoffRecord` interface: full elicitation question, Telegram route, elicit ID, answer fields, and original command for re-enqueue.
- Five async CRUD helpers using atomic `tmp+rename` writes (same pattern as `queue-store.ts` and `lease-store.ts`):
  - `writeHandoff` — atomic write to `<handoffsDir>/<taskId>.json`
  - `readHandoff` — read, null on ENOENT
  - `deleteHandoff` — idempotent remove
  - `listPendingHandoffs` — filter to `status === 'pending'`
  - `updateHandoffAnswer` — read-modify-write, transitions to `'answered'`

### `src/agent/daemon/handoff-store.test.ts` (new file)
16 tests covering: round-trip, null-on-miss, idempotent delete, status filtering, answer transition, field preservation, error cases, and concurrent writes (20-way parallel).

### `src/agent/daemon/task-lifecycle.test.ts`
Updated existing TaskState test to include `'waiting_human_input'` (8 states, was 7) and added a dedicated compile-time type check.

## Verification

- `pnpm lint` — passes (tsc --noEmit strict)
- `pnpm audit:filesize:check` — passes (handoff-store.ts: 131 code lines, paths.ts stays at ceiling)
- `pnpm test src/agent/daemon/handoff-store.test.ts src/agent/daemon/task-lifecycle.test.ts` — 28/28 pass

## What is NOT in this PR

- No wiring into elicitation-router, Telegram handler, or scheduler (PR 2)
- No `daemonTaskId` in AgentConfig (PR 2)
- No behavior change anywhere in the existing system

### DIFF
--- a/scripts/audit-deps-endpoint.ts
+++ b/scripts/audit-deps-endpoint.ts
@@ -27,6 +27,8 @@ const TRANSPORT_FAILURE_MARKERS: readonly RegExp[] = [
   /\bECONNREFUSED\b/,
   /\bECONNRESET\b/,
   /\bETIMEDOUT\b/,
+  // Socket-level timeout (Node fetch / node-fetch / undici): ERR_SOCKET_TIMEOUT
+  /\bERR_SOCKET_TIMEOUT\b/,
   /\bfetch failed\b/i,
 ];
 
--- a/src/agent/daemon/handoff-store.test.ts
+++ b/src/agent/daemon/handoff-store.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, beforeEach, afterAll } from 'vitest';
-import { mkdtemp, rm, stat } from 'node:fs/promises';
+import { mkdtemp, rm, stat, utimes, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import {
@@ -275,6 +275,25 @@ describe('updateHandoffAnswer', () => {
     await writeHandoff(record, testDir);
     const result = await updateHandoffAnswer(record.taskId, 'second answer', 'web', testDir);
     expect(result.won).toBe(false);
+  });
+
+  it('clears a stale lock left by a crashed process and succeeds', async () => {
+    const record = makeRecord({ taskId: 'q-stale-lock' });
+    await writeHandoff(record, testDir);
+
+    // Simulate a crash: manually create the .lock file with an old mtime (>30s).
+    const lockFile = join(testDir, `${record.taskId}.lock`);
+    await writeFile(lockFile, '', { mode: 0o600 });
+    const staleTime = new Date(Date.now() - 60_000); // 60 seconds ago
+    await utimes(lockFile, staleTime, staleTime);
+
+    // updateHandoffAnswer should detect the stale lock, remove it, and succeed.
+    const result = await updateHandoffAnswer(record.taskId, 'recovered', 'telegram', testDir);
+    expect(result.won).toBe(true);
+
+    const read = await readHandoff(record.taskId, testDir);
+    expect(read?.status).toBe('answered');
+    expect(read?.answer).toBe('recovered');
   });
 
   it('CAS: exactly one concurrent caller wins when two race for the same taskId', async () => {
--- a/src/agent/daemon/handoff-store.test.ts
+++ b/src/agent/daemon/handoff-store.test.ts
@@ -1,5 +1,5 @@
-import { describe, it, expect, beforeEach } from 'vitest';
-import { mkdtemp, rm } from 'node:fs/promises';
+import { describe, it, expect, beforeEach, afterAll } from 'vitest';
+import { mkdtemp, rm, stat } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import {
@@ -22,7 +22,8 @@ function makeRecord(overrides: Partial<HandoffRecord> = {}): HandoffRecord {
     question: {
       type: 'text',
       message: 'What is your name?',
-    },
+    } as Record<string, unknown>,
+    requestType: 'ask_question',
     createdAt: new Date().toISOString(),
     status: 'pending',
     originalCommand: '/test-command',
@@ -49,7 +50,7 @@ beforeEach(async () => {
 // ---------------------------------------------------------------------------
 
 describe('writeHandoff + readHandoff', () => {
-  it('round-trips a full HandoffRecord including optional fields', async () => {
+  it('round-trips a full HandoffRecord including optional fields (ask_question)', async () => {
     const record = makeRecord({
       taskId: 'q-1716000000000-abc123',
       sessionId: 'sess-roundtrip',
@@ -60,7 +61,8 @@ describe('writeHandoff + readHandoff', () => {
         default: 'A',
         allowCustom: false,
         allowSkip: true,
-      },
+      } as Record<string, unknown>,
+      requestType: 'ask_question',
       route: { chatId: 123456, threadId: 7 },
       elicitId: 'elicit-xyz',
       originalCommand: '/run-something --flag',
@@ -72,15 +74,42 @@ describe('writeHandoff + readHandoff', () => {
     expect(read).not.toBeNull();
     expect(read?.taskId).toBe(record.taskId);
     expect(read?.sessionId).toBe(record.sessionId);
-    expect(read?.question.type).toBe('choice');
-    expect(read?.question.choices).toEqual(['A', 'B', 'C']);
-    expect(read?.question.default).toBe('A');
-    expect(read?.question.allowSkip).toBe(true);
+    expect(read?.requestType).toBe('ask_question');
+    expect(read?.question['type']).toBe('choice');
+    expect(read?.question['choices']).toEqual(['A', 'B', 'C']);
+    expect(read?.question['default']).toBe('A');
+    expect(read?.question['allowSkip']).toBe(true);
     expect(read?.route?.chatId).toBe(123456);
     expect(read?.route?.threadId).toBe(7);
     expect(read?.elicitId).toBe('elicit-xyz');
     expect(read?.status).toBe('pending');
     expect(read?.originalCommand).toBe('/run-something --flag');
+  });
+
+  it('round-trips an MCP-style question with requestType mcp', async () => {
+    const record = makeRecord({
+      taskId: 'q-mcp-roundtrip',
+      sessionId: 'sess-mcp',
+      question: {
+        mode: 'form',
+        message: 'Please fill in the form',
+        requestedSchema: { properties: { name: { type: 'string' } } },
+        title: 'MCP Form',
+        serverName: 'my-mcp-server',
+      } as Record<string, unknown>,
+      requestType: 'mcp',
+    });
+
+    await writeHandoff(record, testDir);
+    const read = await readHandoff(record.taskId, testDir);
+
+    expect(read).not.toBeNull();
+    expect(read?.requestType).toBe('mcp');
+    expect(read?.question['mode']).toBe('form');
+    expect(read?.question['title']).toBe('MCP Form');
+    expect((read?.question['requestedSchema'] as Record<string, unknown>)['properties']).toEqual({
+      name: { type: 'string' },
+    });
   });
 
   it('creates the handoffs directory if it does not exist', async () => {
@@ -187,13 +216,15 @@ describe('listPendingHandoffs', () => {
 // ---------------------------------------------------------------------------
 
 describe('updateHandoffAnswer', () => {
-  it('transitions a pending record to answered and records metadata', async () => {
+  it('transitions a pending record to answered and returns { won: true }', async () => {
     const record = makeRecord({ taskId: 'q-answer-me' });
     await writeHandoff(record, testDir);
 
     const before = Date.now();
-    await updateHandoffAnswer(record.taskId, 'Alice', 'telegram', testDir);
+    const result = await updateHandoffAnswer(record.taskId, 'Alice', 'telegram', testDir);
     const after = Date.now();
+
+    expect(result.won).toBe(true);
 
     const read = await readHandoff(record.taskId, testDir);
     expect(read?.status).toBe('answered');
@@ -210,16 +241,17 @@ describe('updateHandoffAnswer', () => {
     const record = makeRecord({
       taskId: 'q-preserve-fields',
       sessionId: 'sess-preserve',
-      question: { type: 'confirm', message: 'Are you sure?' },
+      question: { type: 'confirm', message: 'Are you sure?' } as Record<string, unknown>,
       route: { chatId: 9999 },
       originalCommand: '/run-preserved',
     });
     await writeHandoff(record, testDir);
-    await updateHandoffAnswer(record.taskId, true, 'web', testDir);
+    const result = await updateHandoffAnswer(record.taskId, true, 'web', testDir);
+    expect(result.won).toBe(true);
 
     const read = await readHandoff(record.taskId, testDir);
     expect(read?.sessionId).toBe('sess-preserve');
-    expect(read?.question.message).toBe('Are you sure?');
+    expect(read?.question['message']).toBe('Are you sure?');
     expect(read?.route?.chatId).toBe(9999);
     expect(read?.originalCommand).toBe('/run-preserved');
     expect(read?.answer).toBe(true);
@@ -232,7 +264,7 @@ describe('updateHandoffAnswer', () => {
     ).rejects.toThrow(/no record found/);
   });
 
-  it('throws if the record is not in pending status', async () => {
+  it('returns { won: false } if the record is not in pending status', async () => {
     const record = makeRecord({
       taskId: 'q-already-answered',
       status: 'answered',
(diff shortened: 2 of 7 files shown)


griffinwork40 (commented):

---

## PR #1484 Review — feat(core): add HandoffRecord schema and file-based handoff store (#1416)

**Target:** `afk/iso-build-1416-pr1-2-ttcd4u` → `main` | **Stats:** +698 / -8 | 5 files (2 new) | **Regime:** full
**Reviewed ref:** `8d18a9be635b29dea93c0d8b981c91417645e12f` | **Stated intent:** PR #1484 title + body

---

### Findings

#### 1. HIGH · blocking:true · confidence:high · correctness · `src/agent/daemon/handoff-store.ts:268–279`
**ref:** 8d18a9be · citation: file-state (verified at reviewed ref)

**Finding:** `updateHandoffAnswer` writes the lock file with `writeFile(lock, '', { flag: 'wx' })` without first calling `ensureHandoffsDir`. If the handoffs directory is absent, `writeFile` throws `ENOENT`. The catch block only traps `EEXIST`, so `ENOENT` propagates as an unhandled error instead of a clean failure result.

**Evidence:**
```ts
// line 270: no ensureHandoffsDir call before lock attempt
const lock = lockPath(handoffsDir, taskId);
await writeFile(lock, '', { flag: 'wx', mode: 0o600 }); // ENOENT if dir absent
```
`writeHandoff` (line 157) and `listPendingHandoffs` (line 217) both call `ensureHandoffsDir`; `updateHandoffAnswer` does not.

**Suggestion:** Add `await ensureHandoffsDir(handoffsDir)` as the first line of `updateHandoffAnswer`, matching `writeHandoff` and `listPendingHandoffs`.

---

#### 2. HIGH · blocking:true · confidence:high · correctness · `src/agent/daemon/handoff-store.ts:120–121, 270–308`
**ref:** 8d18a9be · citation: file-state (verified at reviewed ref)

**Finding:** A stale `.lock` file left by a crashed or SIGKILL'd daemon permanently poisons `updateHandoffAnswer` for that `taskId`. Every subsequent call hits `EEXIST` at line 275 and returns `{ won: false }`, making the handoff permanently unanswerable. No stale-lock detection, TTL check, or break-lock mechanism exists.

**Evidence:**
```ts
// line 273: lock created; if process dies before line 308 (finally/unlink):
await writeFile(lock, '', { flag: 'wx', mode: 0o600 });
// ...async read + write...
// line 308: never reached on crash
try { await unlink(lock); } catch { /* ignore */ }
```

**Suggestion:** On `EEXIST`, stat the lock file and break it if `mtime` exceeds a grace period (e.g. 30s). Alternatively, store a PID/timestamp in the lock content and break when the PID is gone. Either approach should be wired alongside `recoverExpiredLeases` in the PR 2 daemon-boot path.

---

#### 3. MEDIUM · blocking:true · security · `src/agent/daemon/handoff-store.ts:114–121`
**ref:** 8d18a9be · citation: file-state (verified at reviewed ref)

**Finding:** `taskId` is used unvalidated in `join()` for both the record path and the lock path, enabling path traversal. `join('/handoffs', '../../../tmp/evil.json')` resolves to `/tmp/evil.json`. In PR 1, no production code calls these functions. PR 2 will thread `taskId` from Telegram callback payloads (an external channel) into these functions, making this a live path traversal vulnerability at that point.

**Evidence:**
```ts
function handoffPath(dir: string, taskId: string): string {
  return join(dir, `${taskId}.json`);  // no taskId validation
}
function lockPath(dir: string, taskId: string): string {
  return join(dir, `${taskId}.lock`);  // same
}
```
Verified: `path.join('/handoffs', '../../../tmp/evil.json')` → `/tmp/evil.json`.

**Suggestion:** Validate `taskId` at the top of each public function with a strict allowlist regex (e.g. `/^[\w-]+$/`) that rejects `/`, `..`, and non-alphanumeric characters. This future-proofs the boundary before PR 2 wires external sources.

---

#### 4. MEDIUM · blocking:false · test-coverage · `src/agent/daemon/handoff-store.test.ts`
waived: bounded, non-data-affecting — the stale-lock behavior is a test gap for the correctness defect in finding #2; fixing the implementation (finding #2) and adding the test are a single fix unit, not independent merge criteria

**Finding:** No test covers the crash-left stale lock scenario. If a `.lock` file pre-exists when `updateHandoffAnswer` is called, the function returns `{ won: false }` even though no concurrent caller holds the lock. The CAS test (line 280) covers a live race only, not an orphaned pre-existing lock.

**Evidence:** `grep` for `stale`, `crash`, `existing.*lock`, `pre.*exist`, `lock.*file`, `orphan` in the test file — zero matches.

**Suggestion:** Add a test: write a pending record, manually create the `.lock` file (simulating a crash), then call `updateHandoffAnswer` and assert the expected result.

---

#### 5. LOW · blocking:false · correctness · `src/agent/daemon/handoff-store.ts:147–148`
**ref:** 8d18a9be · citation: file-state (verified at reviewed ref)

**Finding:** JSDoc for `writeHandoff` states concurrent renames "produce one winner and one ENOENT on the loser's rename." On POSIX, `rename(2)` is atomic and both concurrent renames succeed — last writer wins silently. No `ENOENT` is produced.

**Suggestion:** Correct to: "concurrent writers for the same taskId all succeed; last rename wins. No torn writes occur."

---

#### 6. LOW · blocking:false · correctness · `src/agent/daemon/handoff-store.ts:190–199`
**ref:** 8d18a9be · citation: file-state (verified at reviewed ref)

**Finding:** `deleteHandoff`'s `ENOENT` catch branch is dead code. `rm` with `{ force: true }` already suppresses `ENOENT` internally.

**Suggestion:** Remove the `try/catch` (rely on `force: true`), or drop `force: true` and keep the catch — not both.

---

#### 7. LOW · blocking:false · test-coverage · `src/agent/daemon/handoff-store.test.ts`
**ref:** 8d18a9be · citation: file-state

**Finding:** No test covers `listPendingHandoffs` silently skipping a corrupt JSON file. The code swallows parse errors at lines 236–238 but no test verifies one-bad-file-among-many yields the valid records.

**Suggestion:** Add a test writing one valid pending record and one `{invalid json` file, then assert `listPendingHandoffs` returns only the valid record.

---

#### 8. LOW · blocking:false · test-coverage · `src/agent/daemon/handoff-store.test.ts`
**ref:** 8d18a9be · citation: file-state

**Finding:** `path_approval` requestType has no round-trip test (zero matches for `path_approval` in the test file). `ask_question` and `mcp` are both covered.

**Suggestion:** Add a minimal round-trip test for `requestType: 'path_approval'`.

---

#### 9. LOW · blocking:false · perf-observability · `src/agent/daemon/handoff-store.ts:236–238`
**ref:** 8d18a9be · citation: file-state

**Finding:** `listPendingHandoffs` silently swallows corrupt/unreadable files with a bare `catch { }` — no log, warning, or trace. A corrupt file causes a handoff to be silently dropped at recovery time, contradicting the durability contract.

**Suggestion:** Emit a `console.warn` with `filename` and `err.message` inside the catch.

---

#### 10. NIT · blocking:false · spec-compliance · PR description

**Finding:** PR claims "16 tests"; the test file contains 18 `it()` blocks (16 cross-platform + 2 Linux-only permission tests that run on CI).

---

### Decision

**❌ DO NOT MERGE — 2 high blocking (correctness); 1 medium blocking (security); 1 medium waived (test-coverage); 4 low; 1 nit.**

The stale-lock issue (finding #2) is the most consequential: a daemon crash during `updateHandoffAnswer` permanently silences that handoff with no recovery path. The missing `ensureHandoffsDir` (finding #1) is a one-line fix. The path-traversal pre-exploitation window (finding #3) is cheapest to fix now before PR 2 threads external `taskId` values into these functions.

---

### What was not checked

- Citations verified inline against branch HEAD `8d18a9be635b29dea93c0d8b981c91417645e12f`.
- Stated intent: PR #1484 title+body — spec-compliance assessed. Stated "purely additive, zero behavior change" is accurate.
- **Not checked:** Runtime test execution — all findings verified by static analysis and ref reads only.
- **Not checked:** PR 2–5 wiring code (not part of this diff). Path traversal reachability assessed against the stated PR 2 Telegram callback integration.
- **Not checked:** Interaction with `recoverExpiredLeases` for `waiting_human_input` state (deferred to PR 2 per stated scope).
- CI status: "Lint & Build" is FAILING on this PR. Tests pass on ubuntu-latest and macos-latest.

griffinwork40 (commented):

---

## Prior Rounds
This is the **first automated review** of PR #1484. No prior findings to carry forward.

---

## Decision: DO NOT MERGE

**Blocking severity floor applied:** only `critical` and `high` findings block; `medium` and below are advisory.

**Arithmetic:** 0 critical, **1 high** (CI failure — no passing build at reviewed ref) → **DO NOT MERGE**; 4 medium advisory, 4 low advisory, 6 nit advisory.

Reviewed ref: `8d18a9be635b29dea93c0d8b981c91417645e12f` | Branch: `afk/iso-build-1416-pr1-2-ttcd4u`

---

## 🔴 Blocking Finding (High)

### [HIGH-1] CI Lint & Build failure — no passing run at reviewed ref
`src/agent/daemon/handoff-store.ts` · ci-lint · blocking:true · confidence:high

**Finding:** The PR's CI shows `Lint & Build: FAILURE` on the most recent run. The PR body claims `pnpm lint` passes and `pnpm test` passes 28/28, which conflicts with the red badge. The `getHandoffsDir` export is confirmed present in `src/paths.ts` at line 291 of the reviewed ref, so the TypeScript import resolves correctly at this SHA. The most likely cause is that CI ran on an earlier push and the fix commit (`8d18a9be`) did not re-trigger or the workflow was not re-run after the push.

**Why it blocks:** A PR with a failing CI badge cannot be merged without a green signal at the reviewed ref, regardless of whether the failure is stale. The author must push a change or manually re-run CI to produce a passing signal at `8d18a9be` (or its successor).

**Fix:** Re-run the `CI` workflow on this branch, or push a no-op commit to trigger CI. Confirm all checks green before re-review.

---

## 🟡 Medium Advisory (Non-Blocking — follow-up issues recommended)

### [MED-1] Stale lock leaves handoff permanently unanswerable after daemon crash
`src/agent/daemon/handoff-store.ts:265–308` · security/data-integrity · blocking:false (advisory) · confidence:high

If the daemon is killed (SIGKILL, OOM) after the O_EXCL lock file is created but before the `finally` block unlinks it, the `.lock` file is left on disk permanently. Every subsequent `updateHandoffAnswer` call for that `taskId` hits `EEXIST` and returns `{ won: false }` — the handoff becomes permanently unanswerable without manual operator intervention.

```typescript
await writeFile(lock, '', { flag: 'wx', mode: 0o600 });
// crash here → lock never removed
if ((err as NodeJS.ErrnoException).code === 'EEXIST') {
  return { won: false };  // silent permanent starvation
}
```

**Suggestion:** Check lock file `mtime` before acquiring — if older than a TTL (e.g. 10 min) and empty, unlink and proceed. Or write the acquiring PID into the lock body and use `process.kill(pid, 0)` to detect dead owners.

**Recommend opening a follow-up issue for PR 2 or PR 3.**

---

### [MED-2] `taskId` joined into filesystem path without sanitization — path traversal risk before PR 2 wiring
`src/agent/daemon/handoff-store.ts:115–122` · security/path-traversal · blocking:false (advisory) · confidence:medium

`handoffPath` and `lockPath` join caller-supplied `taskId` directly into a filesystem path with no character-set guard. The codebase pattern for identifier-bearing paths (`getBgJobDir`, `getSubagentLogPath`, etc.) consistently calls `assertSafeJobId` before any `join()`. PR 1 has no external callers, so this is not currently exploitable, but PR 2 will wire Telegram-sourced data into these functions.

```typescript
function handoffPath(dir: string, taskId: string): string {
  return join(dir, `${taskId}.json`);  // no validation
}
```

**Suggestion:** Add input validation at the top of `writeHandoff`, `readHandoff`, `deleteHandoff`, and `updateHandoffAnswer`, reusing the existing `BG_JOB_ID_PATTERN = /^[A-Za-z0-9_-]+$/` and `BG_JOB_ID_MAX_LEN = 128` from `paths.ts`. This guard should land before PR 2 wires the call path, not after.

**Recommend opening a follow-up issue to fix before PR 2 lands.**

---

### [MED-3] `recoverExpiredLeases` does not suppress recovery for `waiting_human_input` state — gap between PR 1 and PR 2 merge
`src/agent/daemon/lease-store.ts:353–379`, `src/agent/daemon/task-lifecycle.ts:28–39` · api-compat/correctness · blocking:false (advisory) · confidence:high

The new `'waiting_human_input'` state's documented invariant is "lease expiry checks are suppressed in this state." However, `recoverExpiredLeases` checks only `leaseExpiry` to decide whether to re-enqueue or dead-letter a task — it does NOT inspect `record.state`. Any task that enters `waiting_human_input` with an elapsed lease will be silently force-recovered during the PR 1→PR 2 merge window.

```typescript
// lease-store.ts:353
const expiry = record.leaseExpiry ?? 0;
if (expiry > now) continue;  // only expiry checked; state ignored
// ... re-enqueue or dead-letter regardless of waiting_human_input
```

Verified: grep of `lease-store.ts` at `8d18a9be` finds zero occurrences of `waiting_human_input`.

**Suggestion (strongly preferred):** Add a three-line guard to `recoverExpiredLeases` in this PR: `if (record.state === 'waiting_human_input') continue;`. This makes the new state safe the moment it is introduced, with zero risk of the inter-PR gap. Alternatively, add a `// TODO(PR2): suppress recovery for waiting_human_input` annotation.

**Recommend this be fixed in PR 1 itself, or opened as a P0 issue for PR 2.**

---

### [MED-4] `afterAll` cleans only last `testDir` — N-1 temp directories leaked per test run
`src/agent/daemon/handoff-store.test.ts:38–41, 372–375` · test-coverage · blocking:false (advisory) · confidence:high

`testDir` is a module-scoped `let` overwritten by each `beforeEach`. The `afterAll` block references `testDir` which at that point holds only the last test's directory. The N-1 preceding directories are never cleaned up on a normal (passing) run.

```typescript
let testDir: string;            // module-scope, overwritten each beforeEach
beforeEach(async () => { testDir = await mkdtemp(...); });  // overwrites
afterAll(async () => { await rm(testDir, ...); });          // only last value
```

**Suggestion:** Collect directories in an array:
```typescript
const testDirs: string[] = [];
beforeEach(async () => { testDir = await mkdtemp(...); testDirs.push(testDir); });
afterAll(async () => { await Promise.all(testDirs.map(d => rm(d, { recursive: true, force: true }))); });
```

---

## 🔵 Low / Nit Advisory (Non-Blocking)

- **[LOW-1]** `deleteHandoff` wraps `rm({ force: true })` in a `catch(ENOENT)` that is unreachable — `force:true` already suppresses ENOENT internally. Dead code; safe to simplify to a bare `await rm(...)` call. `handoff-store.ts:82–90`

- **[LOW-2]** `readHandoff` and `listPendingHandoffs` cast `JSON.parse(raw)` to `HandoffRecord` with no runtime shape validation. Structurally-plausible but semantically corrupt records (wrong `status` type) pass through silently. Acceptable in PR 1 with no live callers; guard before PR 2 wiring. `handoff-store.ts:70–75`

- **[LOW-3]** `mkdir({ recursive: true, mode: 0o700 })` is a no-op on a pre-existing directory and does NOT update its permissions. If `handoffs/` was previously created at 0o755, elicitation payloads are directory-listable. Individual files are protected (0o600). Consistent with existing stores. `handoff-store.ts:106–108`

- **[LOW-4]** No exhaustive switch consumers of `TaskState` exist at this ref — `'waiting_human_input'` addition is safe. Verified by grep. No action needed. `task-lifecycle.ts`

- **[NIT-1]** `join(dest, '..')` in `atomicWriteJson` is functionally correct but `dirname(dest)` is clearer. `handoff-store.ts:46`

- **[NIT-2]** `HandoffRecord.answer?: unknown` — non-JSON-serializable values silently dropped or throw on `JSON.stringify`. Add a serialization guard in `atomicWriteJson`. `handoff-store.ts:75–92`

- **[NIT-3]** Orphaned `.tmp-*` files accumulate if daemon crashes during a write; `listPendingHandoffs` filters but never GCs them. Consistent with existing stores. `handoff-store.ts:131–139`

- **[NIT-4]** PR body test count (28/28) appears to reflect a macOS run (30 total, 2 Linux-only permission tests skipped). Minor doc inaccuracy. `handoff-store.test.ts`

- **[NIT-5]** `'waiting_human_input'` is missing from the JSDoc bullet list in `task-lifecycle.ts:16–26`; only present as an inline comment on the union arm. `task-lifecycle.ts:16–26`

---

## What Was Not Checked

- **Stated intent:** PR title + body supplied — spec-compliance assessed. PR is correctly described as purely additive; no scope creep found.
- **Citations verified** inline against branch HEAD `8d18a9be635b29dea93c0d8b981c91417645e12f`.
- **Did not review:** Telegram surface integration (PR 2 scope), elicitation-router wiring, scheduler behavior.
- **Did not run tests** (read-only review).
- **CI root cause** not conclusively determined — the failure could be a stale run; requires re-trigger to confirm.

PR #1491, opened 2026-09-04
feat(subagent): auto-cap oversized foreground results with sidecar spill

griffinwork40:
## Problem

When a foreground subagent produces a large final message (research reports, scoping docs, verbose analysis), the full output flows verbatim into the parent's context window with no size management. This causes:

1. **Context bloat** -- a 40KB subagent result consumes 40KB of parent context, crowding out the parent's own reasoning budget
2. **Silent information loss** -- background results are truncated at 16KB with no file-based recovery; foreground results have no cap at all but consume proportionally more context
3. **Manual workarounds** -- dispatchers must remember to add "write to a file and return the path" instructions every time, and read-only agents can't even comply

## Solution

Automatic size-tiered handoff at the `toolResult` assembly point in `foreground-promotion.ts`:

- **Under 32KB**: passes through unchanged (zero overhead for well-behaved subagents)
- **Over 32KB**: full output is spilled to a sidecar file at `sessions/<id>/subagent-handoffs/<subagentId>.txt`, and the parent receives a `headAndTail` slice (preserving both the start and end of the output) plus an explicit `read_file` pointer to the full text

Clones the proven `spillNodeOutput` pattern from `compose-executor.ts`. Best-effort: spill failures are swallowed and the parent still gets the truncated view.

## Changes

| File | What |
|------|------|
| `src/agent/tools/subagent/foreground-promotion.result-cap.ts` | New helper: `capSubagentResult()` -- measures, spills, caps |
| `src/agent/tools/subagent/foreground-promotion.ts` | Wire the cap after `annotateIfIncomplete`/`withProvenanceHeader`, before `toolResult` assignment |
| `src/config/env.ts` | New env var: `AFK_SUBAGENT_RESULT_CAP_BYTES` (default 32768; set to 0 to disable) |
| `docs/env-registry.{json,md}` | Regenerated via `pnpm scan:env` |
| `.filesize-baseline.json` | Ratchet update for env.ts growth (+14 lines) |

## Configuration

```bash
# Default: 32KB cap (no config needed)
# Custom threshold:
export AFK_SUBAGENT_RESULT_CAP_BYTES=16384
# Disable entirely:
export AFK_SUBAGENT_RESULT_CAP_BYTES=0
```

## Tests

8 new unit tests covering:
- Inline passthrough for small results
- Cap + sidecar spill for oversized results
- Head+tail preservation (both ends of the output survive)
- Custom threshold via env var
- Disable path (env=0)
- Graceful fallback on missing sessionId
- Fallback on unparseable/negative env values

## Not in scope

- Background result injection (`bg-result-notifier.ts`) has its own independent 16KB cap -- unchanged here
- No changes to the `SUBAGENT_HANDOFF_CONTRACT` prompt (still advisory)
- No structured output schema enforcement (considered and rejected via devil's advocate analysis -- schema failures are worse than truncation)


griffinwork40:
Good -- `sessionId` is always `randomUUID()` at the provider level. Now I have full Wave 1.5 verification context. Let me synthesize.

**Wave 1.5 Citation + Absence-Claim Verification Manifest:**

| # | Type | Source | Claim | Status | Notes |
|---|------|--------|-------|--------|-------|
| S1 | citation | security F1 | `spillSubagentOutput` at lines 24-32 builds path via `join()` | **verified** | Lines confirmed at ref; `join()` does resolve `..` |
| S1 | citation | security F1 | `handle.id` is subagentId at call site | **verified** | Confirmed at `foreground-promotion.ts:343` |
| S1 | inference | security F1 | `sessionId`/`subagentId` may contain traversal sequences | **downgraded** | `subagentId` = `${safePrefix}-${timestamp}-${counter}` where `safePrefix` is regex-sanitized `[^A-Za-z0-9_-]` -> `-`. `sessionId` = `randomUUID()`. Neither can contain `..` in normal operation. |
| S2 | citation | security F2 | `writeFileSync(path, raw, 'utf8')` at line 30 | **verified** | Confirmed at ref |
| S2 | absence | security F2 | "no confirmed sweep for sessions directory" | **confirmed-absent** | `git grep -n 'subagent-handoffs' 12de6954` returns only the result-cap files. No sweep for `state/sessions/<id>/subagent-handoffs/`. The session-dir sweep (`session-sweep.ts`) targets sidecar `.json` files, not subdirectories. |
| C2 | citation | correctness F2 | Failure path lines 397-402 uncapped | **verified** | Confirmed at ref -- `JSON.stringify(payload)` with no `capSubagentResult` |
| C2 | absence | correctness F2 | `partialOutput` is bounded by `shapePartialOutput` | **verified** | `MAX_PARTIAL_OUTPUT_CHARS = 4096` at `failure-payload.ts:73`. The failure payload IS already bounded. |
| C4 | absence | test-coverage F4 | "No test covers the spill-failure path" | **confirmed-absent** | `git grep -n 'could not be written\|writeFileSync.*throw\|spill.*fail' 12de6954 -- '*.test.ts'` returns zero matches |
| C5 | citation | test-coverage F5 | `foreground-promotion.test.ts` has no cap test | **verified** | Confirmed the file exists but `grep 'capSubagentResult\|result-cap\|capped' src/agent/tools/subagent/foreground-promotion.test.ts` returns zero matches |

**Key downgrade:** Security Finding 1 (path traversal). Both `subagentId` and `sessionId` are runtime-generated: `subagentId` is `${safePrefix}-${timestamp}-${counter}` where `safePrefix` is regex-sanitized to `[A-Za-z0-9_-]` only (see `fork-resolution.ts:95-96`), and `sessionId` is `randomUUID()` (see `anthropic-direct/query-runtime.ts:150`). Neither value can contain `..` segments in normal operation. The theoretical traversal requires injecting a crafted value through the `SubagentHandleImpl` constructor, which only `forkSubagent()` calls in production. Downgrading from high to low. `blocking` preserved from pre-downgrade high severity.

**Correctness Finding 2 reassessment:** The failure path's `partialOutput` is already bounded to 4096 chars by `shapePartialOutput()`, and `errorMessage` is bounded to 1024 chars by `MAX_ERROR_MESSAGE_CHARS`. The `JSON.stringify(payload)` output is structurally bounded to ~6KB maximum. While technically over 32KB is not possible on this path, the finding that the cap is applied asymmetrically is still valid as a correctness/maintenance concern. Downgrading from medium to low.

Now I can emit the synthesis:

---

## PR #1491 Review: `feat(subagent): auto-cap oversized foreground results with sidecar spill`

**Reviewed ref:** `12de6954018feb117ebd2dcca114bd9c484eff63`
**Change type:** feature | **Regime:** light (264 lines) | **Dimensions:** security, api-compat, correctness, spec-compliance, test-coverage, perf-observability
**Stated intent:** PR title+body -- spec-compliance assessed.

---

### Findings

**1.** `medium` · `blocking: true` · confidence high · **security**
`src/agent/tools/subagent/foreground-promotion.result-cap.ts:28-30` · ref: 12de6954 · diff-context

The sidecar file receives the raw, unredacted pre-truncation subagent output and persists at `~/.afk/state/sessions/<id>/subagent-handoffs/<subagentId>.txt` with no documented retention sweep. The session-dir sweep (`session-sweep.ts`) targets sidecar `.json` files, not subdirectories -- so these spill files accumulate indefinitely. The project already recognizes this risk class for `AFK_CAPTURE_SUBAGENT_OUTPUT` (off by default for the same reason), but this cap is on by default.

```typescript
writeFileSync(path, raw, 'utf8');
```

**Suggestion:** Document in the `AFK_SUBAGENT_RESULT_CAP_BYTES` env entry that sidecar files contain unredacted subagent output and are subject to session-directory retention. Track extending the session-dir sweep to cover `subagent-handoffs/` subdirectories as a follow-up.

---

**2.** `low` · `blocking: true` · confidence high · **security** · blocking preserved from pre-downgrade high

`src/agent/tools/subagent/foreground-promotion.result-cap.ts:54-58` · ref: 12de6954 · diff-context

`spillSubagentOutput` constructs the sidecar path via `join(getSessionsDir(), sessionId, 'subagent-handoffs', subagentId + '.txt')` with no validation that path components are traversal-free. `path.join()` resolves `..` segments. **In practice, both values are safe:** `subagentId` is `${safePrefix}-${timestamp}-${counter}` where `safePrefix` is regex-sanitized to `[A-Za-z0-9_-]` (`fork-resolution.ts:95`), and `sessionId` is `randomUUID()` (`query-runtime.ts:150`). No production path produces traversal-capable values. Downgraded from high because the attack requires bypassing the sanitization at source, not at this call site.

```typescript
const dir = join(getSessionsDir(), sessionId, 'subagent-handoffs');
const path = join(dir, `${subagentId}.txt`);
```

**Suggestion:** Add a `path.basename()` call on both components as defense-in-depth, matching the defensive posture that `compose-executor.ts` has for `spillNodeOutput` (same structural pattern, same theoretical gap).

---

**3.** `medium` · `blocking: true` · confidence high · **test-coverage**
`src/agent/tools/subagent/foreground-promotion.result-cap.test.ts` · ref: 12de6954 · file-state

No test covers the spill-failure path. When `writeFileSync` throws (disk full, permissions error), `spillSubagentOutput` returns `undefined` and the pointer message changes to the no-file-pointer fallback (`"Full output could not be written to disk."`). This is the only untested branch in the 107-line module.

**Suggestion:** Add a test that mocks `writeFileSync` to throw, verifies `capped: true`, and asserts the pointer contains `"could not be written to disk"` without a file path.

---

**4.** `low` · `blocking: false` · confidence high · **correctness**
`src/agent/tools/subagent/foreground-promotion.result-cap.ts:100-106` · ref: 12de6954 · file-state

The pointer line (~120-200 bytes) is appended after `headAndTail(content, capBytes)` returns, so the actual delivery slightly exceeds `capBytes`. `headAndTail` reserves 160 bytes for its own marker but has no knowledge of the external pointer. The overage is small, bounded, and the cap is a heuristic -- not a hard memory boundary.

```typescript
const capped = headAndTail(content, capBytes);
const pointer = spillPath !== undefined ? `\n\n[...]` : `\n\n[...]`;
return { content: capped + pointer, capped: true };
```

**Suggestion:** Add a one-line comment documenting the known ~200-byte overage as acceptable, or deduct the pointer's max length from `capBytes` before passing to `headAndTail`.

---

**5.** `low` · `blocking: false` · confidence high · **correctness**
`src/agent/tools/subagent/foreground-promotion.result-cap.ts:88-89` · ref: 12de6954 · file-state

When `sessionId` is `undefined`, `capSubagentResult` passes through the full content uncapped. The stated intent targets "foreground subagent results" without qualifying to sessions-with-IDs. A session without an ID is rare but possible, and in that case the oversize result flows uncapped.

```typescript
if (capBytes === undefined || sessionId === undefined) {
  return { content, capped: false };
}
```

**Suggestion:** When `sessionId` is `undefined`, still truncate with `headAndTail` and return the no-spill-pointer fallback rather than bypassing the cap entirely.

---

**6.** `low` · `blocking: false` · confidence high · **test-coverage**
`src/agent/tools/subagent/foreground-promotion.test.ts` · ref: 12de6954 · file-state

No integration test verifies the cap wiring in `runForegroundWithPromotion`. The unit tests cover `capSubagentResult` thoroughly in isolation, but the two-line wiring change at line 343 (the only integration point) is untested.

**Suggestion:** Low priority given the unit test coverage. A single integration test producing a mock result over 32KB and asserting the truncation marker appears would close the gap.

---

**7.** `nit` · `blocking: false` · confidence high · **perf-observability**
`src/agent/tools/subagent/foreground-promotion.result-cap.ts:58` · ref: 12de6954 · diff-context

`writeFileSync` blocks the event loop for the duration of the write. Acceptable given it runs once per oversized subagent at the tail of a completed run, and matches the `compose-executor.ts` precedent.

**Suggestion:** Add a one-line comment documenting the synchronous choice.

---

**8.** `nit` · `blocking: false` · confidence high · **api-compat**

No breaking API-compat impact. `capSubagentResult` and `ResultCapOutcome` are imported only by `foreground-promotion.ts` (its single caller). No existing callers of `headAndTail`, `capForModel`, or other `_output-cap.ts` exports are affected.

---

### What was not checked

- Citations verified against branch HEAD `12de6954`.
- Stated intent: PR #1491 title+body -- spec-compliance assessed.
- Did not run the test suite. Did not review the Telegram or daemon surfaces.
- Did not audit the `compose-executor.ts` `spillNodeOutput` for the same path-traversal pattern (noted as a pre-existing gap in the same structural family).
- Background result injection (`bg-result-notifier.ts`) was out of scope per the PR's own declaration.

### Decision: **DO NOT MERGE** -- 1 medium blocking (security), 1 low blocking (security, preserved from pre-downgrade high), 1 medium blocking (test-coverage); 3 low, 2 nit.

The three blocking findings are all low-effort to address before merge:
1. Document the sidecar retention gap and track the sweep extension.
2. Add `path.basename()` defense-in-depth on both path components.
3. Add one test for the spill-failure fallback path.

**Done**

- What was done: Full-regime code review of PR #1491 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability.
- Evidence: Review findings above, verified against ref `12de6954` with inline Wave 1.5 citation checks.
- Deferred: none.

---

griffinwork40:
Wave 2 synthesis complete. The single blocking finding (F-2, correctness -- cap overshoot) is a `medium` at its default disposition, so no `/shadow-verify` is required (shadow-verify fires only on critical/high findings or findings whose `blocking` value departs from the default table). F-2 is `medium · blocking:true`, which is the default for `medium`.

---

## PR #1491 Review — feat(subagent): auto-cap oversized foreground results with sidecar spill

### Triage

| Field | Value |
|---|---|
| **Change type** | Feature |
| **Branch** | `afk/improve-handoff-process` → `main` |
| **Reviewed ref** | `60e0b885` |
| **Regime** | Light (342 lines, 8 files) |
| **Key files** | `foreground-promotion.result-cap.ts` (new) · `foreground-promotion.ts` (wired) · `env.ts` (registry entry) |

---

### Findings

#### F-2 · medium · blocking:true · correctness
**`foreground-promotion.result-cap.ts:100-109`** · ref:`60e0b885` · file-state (verified)

`headAndTail` + pointer concatenation can silently exceed the configured cap. `headAndTail(content, capBytes)` internally reserves 160 bytes (`MARKER_RESERVE_BYTES`) for its own elision marker, but the downstream pointer string (~115-145 bytes) is concatenated *after* `headAndTail` returns with no headroom deduction. At 32KB default the overshoot is ~0.4% (trivial), but at small custom caps (`AFK_SUBAGENT_RESULT_CAP_BYTES=500`) the pointer represents ~28% overshoot.

```ts
const capped = headAndTail(content, capBytes);
const pointer = spillPath !== undefined
  ? `\n\n[Subagent output was ${byteLength} bytes — full result at ${spillPath}. Use read_file to retrieve it.]`
  : `\n\n[Subagent output was ${byteLength} bytes — truncated to ${capBytes} bytes. Full output could not be written to disk.]`;
return { content: capped + pointer, capped: true };
```

**Fix:** Subtract `Buffer.byteLength(pointer, 'utf8')` from `capBytes` before passing to `headAndTail`, or document the cap as a soft bound.

---

#### F-3 · low · blocking:false · perf-observability
**`foreground-promotion.result-cap.ts:16,59,61`** · ref:`60e0b885` · file-state (verified)

`mkdirSync` / `writeFileSync` block the event loop on an already-async path. All other sidecar writes in this codebase use `fs/promises`. At 32KB negligible, but at higher custom caps on slow I/O the penalty grows.

**Fix:** Switch to `mkdir`/`writeFile` from `node:fs/promises` + `await`.

---

#### F-1 · low · blocking:false · security
**`foreground-promotion.result-cap.ts:58`** · ref:`60e0b885` · file-state (verified) · *Downgraded from medium after Wave 1.5: `basename()` prevents directory escape, not just last-segment stripping*

`basename(sessionId)` correctly prevents filesystem escape from `getSessionsDir()` (confirmed: `basename('../foo')` → `'foo'`, stays inside sessions dir). Residual risk is cross-session collision on non-UUID input. Production `sessionId` is always `randomUUID()`. The in-source comment correctly documents the defense.

**Fix:** No action required. Optional: add `UUID_PATTERN.test(sessionId)` guard for machine-checked defense.

---

#### F-4 · low · blocking:false · spec-compliance
**`env.ts:1458` / `docs/env-registry.json:779`** · ref:`60e0b885` · diff-context

`AFK_SUBAGENT_RESULT_CAP_BYTES` is categorized as `"model"` but controls runtime I/O/context management. Adjacent capture/log env vars (`AFK_CAPTURE_SUBAGENT_OUTPUT`, `AFK_SUBAGENT_LOG`) use `"debug"`.

**Fix:** Change category to `"debug"`.

---

#### F-5 · low · blocking:false · spec-compliance
**`foreground-promotion.result-cap.ts:38-42` / `foreground-promotion.ts:343`** · ref:`60e0b885` · file-state (verified)

`ResultCapOutcome.capped` is exported but the call site destructures only `content`. The `subagent.completed` telemetry payload has no capping signal, making cap events invisible to monitoring.

**Fix:** Plumb `capped` into the telemetry event, or remove the unused boolean.

---

#### F-6 · nit · blocking:false · test-coverage
**`foreground-promotion.result-cap.test.ts:52-55`** · ref:`60e0b885` · file-state (verified)

Test asserts `< original` but not `<= capBytes` -- would not catch the headroom overshoot from F-2.

**Fix:** Add `expect(Buffer.byteLength(result.content, 'utf8')).toBeLessThanOrEqual(32_768 + 200)`.

---

### Decision: DO NOT MERGE -- 1 medium blocking (correctness); 4 low, 1 nit.

The single blocker (F-2) is a two-line fix: subtract the pointer's byte length from the `headAndTail` budget. The merge path is: fix the headroom accounting, add the upper-bound test assertion (F-6), re-verify.

---

### What was not checked

- Citations verified against branch HEAD `60e0b885` (Wave 1.5 inline, Wave 2 subagent).
- Stated intent: PR #1491 title+body -- spec-compliance assessed.
- Not checked: `headAndTail` behavior on multi-byte Unicode at exact cap boundaries (handled upstream by `_output-cap.ts` UTF-8-safe functions); GC of `subagent-handoffs/` sidecar files (documented as out-of-scope in env-registry); background-mode result assembly (separate code path, PR is foreground-only); provenance-header length interaction with cap threshold (additive text before cap, not traced end-to-end); unredacted sidecar content security implications (documented caveat).

**Done**
- What was done: Completed light-regime code review of PR #1491 across all 6 dimensions (security, api-compat, correctness, spec-compliance, test-coverage, perf-observability) with Wave 1 + Wave 1.5 citation verification + Wave 2 synthesis.
- Evidence: Findings above, verified against ref `60e0b885`. Diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/d6d8f1f9-ed7e-4814-8e34-788bc84a50c9/pr-1491.diff`.
- Deferred: None. One blocking finding (F-2) requires author action before merge.

---

PR #1493, opened 2026-09-04
feat(daemon): wire durable handoff store into elicitation path (#1416)

griffinwork40:
## What

PR 2 of the durable human handoff feature (#1416): wires the handoff-store (landed in PR 1) into the daemon's elicitation and lease-recovery paths so pending questions survive daemon restarts.

## Why

When a daemon task calls `ask_question`, the question exists only in process memory. If the daemon restarts (upgrade, OOM, crash), the pending question is silently lost. This PR makes questions durable by writing them to disk before notification and recovering them on startup.

### New file: `src/agent/daemon/handoff-wiring.ts`

Three capabilities:

1. **`makeDaemonElicitationHandler`** — An `ElicitationHandler` for daemon sessions that:
   - Writes a `HandoffRecord` to disk **before** sending a Telegram notification (write-before-notify invariant)
   - Transitions the task's lease to `waiting_human_input` so lease recovery skips it while the human decides
   - Returns `{ action: 'decline' }` — daemon sessions are non-interactive; the durable handoff record is the mechanism for eventual delivery

2. **`recoverPendingHandoffs`** — Called on daemon startup to:
   - Re-notify the operator via Telegram for any pending handoffs
   - Expire handoffs older than 24 hours (configurable TTL)

3. **`answerHandoff` / `cleanupHandoff`** — Thin wrappers for external surfaces (Telegram, web, REPL) to record answers via the first-writer-wins CAS lock in the handoff store

### Changes to existing files

- **`lease-store.ts`**: Adds `setLeaseState()` for in-place lease state transitions. Adds `waiting_human_input` suppression in `recoverExpiredLeases()` — tasks blocked on a human answer are never incorrectly reclaimed, even if their lease has technically expired.
- **`scheduler.ts`**: Calls `recoverPendingHandoffs()` at startup alongside existing lease recovery. Consolidates three `../../paths.js` imports into one (net zero growth on the baselined file).
- **`task-lifecycle.ts`**: Updates the `waiting_human_input` doc comment to reflect that wiring is complete.

## What's NOT in this PR

- **Task resumption with answer injection** (PR 3) — When a daemon restarts, the *session* that asked the question is dead. PR 3 adds a mechanism to re-enqueue the task with the answer pre-filled.
- **Web UI / REPL surfaces** (PR 4) — Rendering pending handoffs in the web dashboard and REPL on session start.
- **Telegram inline keyboard for daemon handoffs** — This PR sends plain-text Telegram notifications. Future work can add inline keyboard buttons for structured answers.

## Testing

- 13 new tests in `handoff-wiring.test.ts` covering handler write-before-notify, request serialization, abort handling, recovery (re-notification + expiry), answer recording (CAS), and cleanup
- 3 new tests in `lease-store.test.ts` covering `setLeaseState` and `waiting_human_input` suppression in `recoverExpiredLeases`
- All 251 existing daemon tests continue to pass
- `pnpm lint` clean, `pnpm audit:filesize:check` passes for all changed files

## Dependency chain

```
#1411 (durable task lifecycle) ✅ merged
  └─> #1416 PR 1 (handoff store) ✅ merged
       └─> #1416 PR 2 (this PR — wire handoff store into daemon)
            └─> #1416 PR 3 (task resumption with answer injection)
                 └─> #1416 PR 4 (expiry, web UI, REPL surfaces)
```

Closes: partial progress on #1416

### DIFF
--- a/src/agent/daemon/handoff-wiring.test.ts
+++ b/src/agent/daemon/handoff-wiring.test.ts
@@ -14,6 +14,7 @@ import {
   readHandoff,
   type HandoffRecord,
 } from './handoff-store.js';
+import { getTaskRecord } from './lease-store.js';
 import type { ElicitationRequest } from '../types/sdk-types.js';
 
 // ---------------------------------------------------------------------------
@@ -218,6 +219,52 @@ describe('recoverPendingHandoffs', () => {
 
     const result = await recoverPendingHandoffs(handoffsDir);
     expect(result.renotified).toBe(2);
+  });
+
+  it('clears lease waiting_human_input state when handoff expires', async () => {
+    const taskId = 'q-expire-lease';
+    const old = new Date(Date.now() - 25 * 60 * 60 * 1_000); // 25 hours ago
+
+    // Write an expired handoff record.
+    const record: HandoffRecord = {
+      taskId,
+      sessionId: 'sess-expire',
+      question: { message: 'Are you still there?' } as Record<string, unknown>,
+      requestType: 'ask_question',
+      createdAt: old.toISOString(),
+      status: 'pending',
+      originalCommand: '/expire-test',
+    };
+    await writeHandoff(record, handoffsDir);
+
+    // Create the lease file in 'waiting_human_input' state.
+    const leasedDir = join(queueDir, 'leased');
+    mkdirSync(leasedDir, { recursive: true });
+    const leaseRecord = {
+      id: taskId,
+      command: '/expire-test',
+      state: 'waiting_human_input',
+      attempts: 1,
+      maxAttempts: 1,
+      leaseExpiry: Date.now() - 1,
+      createdAt: old.getTime(),
+      updatedAt: old.getTime(),
+    };
+    writeFileSync(join(leasedDir, `${taskId}.json`), JSON.stringify(leaseRecord));
+
+    // Run recovery with the queueDir threaded through.
+    const result = await recoverPendingHandoffs(handoffsDir, queueDir);
+    expect(result.expired).toBe(1);
+
+    // Handoff must be marked expired.
+    const updated = await readHandoff(taskId, handoffsDir);
+    expect(updated?.status).toBe('expired');
+
+    // Lease must no longer be in waiting_human_input — it should be 'leased'
+    // so recoverExpiredLeases can pick it up on the next cycle.
+    const lease = getTaskRecord(taskId, queueDir);
+    expect(lease).not.toBeNull();
+    expect(lease?.state).toBe('leased');
   });
 });
 
--- a/src/agent/daemon/handoff-wiring.test.ts
+++ b/src/agent/daemon/handoff-wiring.test.ts
@@ -1,0 +1,291 @@
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { mkdtemp, rm, readFile } from 'node:fs/promises';
+import { mkdirSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import {
+  makeDaemonElicitationHandler,
+  recoverPendingHandoffs,
+  answerHandoff,
+  cleanupHandoff,
+} from './handoff-wiring.js';
+import {
+  writeHandoff,
+  readHandoff,
+  type HandoffRecord,
+} from './handoff-store.js';
+import type { ElicitationRequest } from '../types/sdk-types.js';
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function makeRequest(overrides: Partial<ElicitationRequest> = {}): ElicitationRequest {
+  return {
+    serverName: 'agent',
+    message: 'What is your name?',
+    origin: 'agent',
+    type: 'text',
+    ...overrides,
+  };
+}
+
+function notAborted(): AbortSignal {
+  return new AbortController().signal;
+}
+
+function aborted(): AbortSignal {
+  const ac = new AbortController();
+  ac.abort();
+  return ac.signal;
+}
+
+// Suppress Telegram push calls — we are testing the store, not the transport.
+vi.mock('../../telegram/push.js', () => ({
+  pushIfConfigured: vi.fn().mockResolvedValue(undefined),
+}));
+
+// ---------------------------------------------------------------------------
+// Fixture: isolated temp directories per test
+// ---------------------------------------------------------------------------
+
+let handoffsDir: string;
+let queueDir: string;
+
+beforeEach(async () => {
+  handoffsDir = await mkdtemp(join(tmpdir(), 'handoff-wiring-test-'));
+  queueDir = await mkdtemp(join(tmpdir(), 'handoff-queue-test-'));
+});
+
+// ---------------------------------------------------------------------------
+// makeDaemonElicitationHandler
+// ---------------------------------------------------------------------------
+
+describe('makeDaemonElicitationHandler', () => {
+  it('writes a handoff record and returns decline', async () => {
+    const handler = makeDaemonElicitationHandler({
+      taskId: 'q-handler-test',
+      originalCommand: '/test-cmd',
+      queueDir,
+      handoffsDir,
+    });
+
+    const result = await handler(makeRequest(), { signal: notAborted(), sessionId: 'sess-1' });
+
+    expect(result.action).toBe('decline');
+
+    // Verify the handoff record was written
+    const record = await readHandoff('q-handler-test', handoffsDir);
+    expect(record).not.toBeNull();
+    expect(record?.status).toBe('pending');
+    expect(record?.taskId).toBe('q-handler-test');
+    expect(record?.sessionId).toBe('sess-1');
+    expect(record?.originalCommand).toBe('/test-cmd');
+    expect(record?.requestType).toBe('ask_question');
+    expect(record?.question['message']).toBe('What is your name?');
+    expect(record?.question['type']).toBe('text');
+  });
+
+  it('serializes choice request fields correctly', async () => {
+    const handler = makeDaemonElicitationHandler({
+      taskId: 'q-choice-test',
+      originalCommand: '/pick',
+      handoffsDir,
+    });
+
+    const request = makeRequest({
+      type: 'choice',
+      choices: ['A', 'B', 'C'],
+      context: 'Pick one',
+      allowCustom: true,
+    });
+    await handler(request, { signal: notAborted() });
+
+    const record = await readHandoff('q-choice-test', handoffsDir);
+    expect(record?.question['type']).toBe('choice');
+    expect(record?.question['choices']).toEqual(['A', 'B', 'C']);
+    expect(record?.question['context']).toBe('Pick one');
+    expect(record?.question['allowCustom']).toBe(true);
+  });
+
+  it('returns decline immediately when signal is already aborted', async () => {
+    const handler = makeDaemonElicitationHandler({
+      taskId: 'q-aborted-test',
+      originalCommand: '/test',
+      handoffsDir,
+    });
+
+    const result = await handler(makeRequest(), { signal: aborted() });
+    expect(result.action).toBe('decline');
+
+    // No record should be written
+    const record = await readHandoff('q-aborted-test', handoffsDir);
+    expect(record).toBeNull();
+  });
+
+  it('uses empty string for sessionId when not provided', async () => {
+    const handler = makeDaemonElicitationHandler({
+      taskId: 'q-no-session',
+      originalCommand: '/test',
+      handoffsDir,
+    });
+
+    await handler(makeRequest(), { signal: notAborted() });
+
+    const record = await readHandoff('q-no-session', handoffsDir);
+    expect(record?.sessionId).toBe('');
+  });
+});
+
+// ---------------------------------------------------------------------------
+// recoverPendingHandoffs
+// ---------------------------------------------------------------------------
+
+describe('recoverPendingHandoffs', () => {
+  it('returns zero counts when no handoffs exist', async () => {
+    const result = await recoverPendingHandoffs(handoffsDir);
+    expect(result.renotified).toBe(0);
+    expect(result.expired).toBe(0);
+  });
+
+  it('re-notifies pending handoffs that are within TTL', async () => {
+    const record: HandoffRecord = {
+      taskId: 'q-recent',
+      sessionId: 'sess-1',
+      question: { message: 'Are you there?' } as Record<string, unknown>,
+      requestType: 'ask_question',
+      createdAt: new Date().toISOString(), // just created — well within TTL
+      status: 'pending',
+      originalCommand: '/hello',
+    };
+    await writeHandoff(record, handoffsDir);
+
+    const result = await recoverPendingHandoffs(handoffsDir);
+    expect(result.renotified).toBe(1);
+    expect(result.expired).toBe(0);
+  });
+
+  it('expires handoffs older than 24 hours', async () => {
+    const old = new Date(Date.now() - 25 * 60 * 60 * 1_000); // 25 hours ago
+    const record: HandoffRecord = {
+      taskId: 'q-old',
+      sessionId: 'sess-1',
+      question: { message: 'Still there?' } as Record<string, unknown>,
+      requestType: 'ask_question',
+      createdAt: old.toISOString(),
+      status: 'pending',
+      originalCommand: '/stale',
+    };
+    await writeHandoff(record, handoffsDir);
+
+    const result = await recoverPendingHandoffs(handoffsDir);
+    expect(result.renotified).toBe(0);
+    expect(result.expired).toBe(1);
+
+    // Verify the record was transitioned to 'expired'
+    const updated = await readHandoff('q-old', handoffsDir);
+    expect(updated?.status).toBe('expired');
+  });
+
+  it('skips non-pending records (answered, expired, cancelled)', async () => {
+    const base = {
+      sessionId: 'sess-1',
+      question: { message: 'test' } as Record<string, unknown>,
+      requestType: 'ask_question' as const,
+      createdAt: new Date().toISOString(),
+      originalCommand: '/test',
+    };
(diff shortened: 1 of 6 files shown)


griffinwork40 (commented):

**Round 1** — no prior automated rounds. Reviewed ref: `7f85a63f6e8a483fb0000eeadd4e32abf4138bdd`.

---

## Decision: DO NOT MERGE — 1 high blocking; 3 medium advisory, 4 low advisory

Arithmetic: 1 high (C-1, blocking) → DO NOT MERGE. 3 medium (A-2, C-2, T-1) are advisory per this sweep's severity floor — follow-up issues recommended, not merge-blocking. 4 low are informational.

---

## Findings

### 🔴 HIGH — Blocking

**C-1 · correctness · `handoff-wiring.ts:205–222` + `lease-store.ts:392–394`**
When a handoff expires in `recoverPendingHandoffs`, the handoff record transitions to `status: 'expired'`, but the corresponding lease file in `leased/<taskId>.json` is never transitioned out of `waiting_human_input`. Because `recoverExpiredLeases` unconditionally skips any lease with `state === 'waiting_human_input'` (lease-store.ts:392), the lease is permanently orphaned — it is never re-enqueued, dead-lettered, or cleaned up on any future daemon restart. The JSDoc comment ("the originating task is dead-lettered") is inaccurate.

**Note:** This defect is latent today because `makeDaemonElicitationHandler` is not yet wired into `spawnSession` (see C-3 below) — no production code path currently sets a lease to `waiting_human_input`. It will become reachable when PR 3 wires the handler, at which point any handoff that goes unanswered for 24h permanently orphans the task's lease.

**Evidence:** `recoverPendingHandoffs` accepts only `handoffsDir?` and has no access to `queueDir`. `setLeaseState` is imported but called only at line 135 (the handler, not the recovery path). No call to `setLeaseState` or `completeTask` exists in the expiry branch.

**Suggestion:** In the expiry branch, after writing the expired HandoffRecord, call `setLeaseState(record.taskId, 'leased', queueDir)` to remove the suppression (requires threading `queueDir?` into `recoverPendingHandoffs`), then let the next `recoverExpiredLeases` cycle handle re-enqueue/dead-letter. Alternatively, call `completeTask(record.taskId, 'failed', 'handoff expired', queueDir)` directly.

---

### 🟡 MEDIUM — Advisory (non-blocking per severity floor; follow-up issues recommended)

**A-2 · api-compat · `lease-store.ts:480–496`**
The `waiting_human_input` suppression added in `recoverExpiredLeases` applies only in the first pass (live lease files). The second pass (stale `.claim-*` file recovery, lines 480–496) performs the same re-enqueue/dead-letter logic without any `waiting_human_input` check. A crash between the first-pass rename-to-claim and recovery completion leaves a `.claim-*` file that the second pass incorrectly re-enqueues or dead-letters, discarding a pending handoff the operator hasn't answered.

**Suggestion:** Add `if (claimRecord.state === 'waiting_human_input') { /* rename back to leased/ or skip */ continue; }` after the crash-window reconstruction in the second pass.

**C-2 · correctness · `lease-store.ts:232–242`**
`setLeaseState` has a TOCTOU window between `existsSync` and `readFileSync` — a concurrent `completeTask` can remove the file between the two calls, or the two writers can race on `atomicWriteJson`. The blast radius is bounded: `setLeaseState` is best-effort (failure caught at handoff-wiring.ts:136), and the race requires sub-ms timing between two daemon operations on the same task. This pattern is pre-existing in `renewLease` (lines 198–209) and is consistent with the file's concurrency approach.

**Suggestion:** Document the TOCTOU limitation in `setLeaseState`'s JSDoc. Optionally, re-check the state after reading and abort if the task is already terminal.

**T-1 · test-coverage · `handoff-wiring.test.ts:168–188`**
No test verifies that after a handoff expires, the lease file is updated or cleaned up. The existing expiry test checks only `result.expired === 1` and the handoff record's status. This absence hides C-1 — a test that also checked the lease state would have caught it.

**Suggestion:** Add a test that writes an expired handoff, calls `recoverPendingHandoffs`, then asserts the lease state via `getTaskRecord`.

---

### 🔵 LOW — Informational

**C-3 · correctness · `scheduler.ts` + `handoff-wiring.ts:99`**
`makeDaemonElicitationHandler` is exported and tested but never injected into the daemon session's `AgentConfig` in `scheduler.ts:spawnSession`. The PR body explicitly defers this to PR 3, but the PR title ("wire durable handoff store into elicitation path") is somewhat misleading — only the recovery path is wired; the elicitation handler itself is not. Noting for completeness.

**S-1 · spec-compliance · `handoff-wiring.ts:40–44`**
The JSDoc for `DEFAULT_HANDOFF_TTL_MS` says "the originating task is dead-lettered" on expiry. The code transitions the handoff to `expired` but does not dead-letter the lease (see C-1). Doc/behavior mismatch.

**S-2 · security · `handoff-wiring.ts:143–153`**
`request.message` and `request.choices[]` are concatenated into the Telegram push payload without escaping Telegram-special characters. The content originates from the agent session (not external user input), so injection requires a compromised agent task. Low practical risk.

**P-1 · perf-observability · `scheduler.ts:358–359`**
`recoverPendingHandoffs()` is fire-and-forget with no result logging. The returned `RecoveryResult` counts are discarded. An operator inspecting startup logs has no visibility into handoff recovery activity, unlike the adjacent `recoverExpiredLeases` call which logs every action.

**Suggestion:** Log the recovery result counts to stderr alongside the existing lease-recovery logging.

---

### What was not checked

- Citations verified inline against branch HEAD `7f85a63f6e`.
- Stated intent: PR #1493 title + body — spec-compliance assessed.
- Did not verify `handoff-store.ts` internals (landed in a prior PR, not part of this diff).
- Did not run tests.
- Did not verify Telegram transport behavior (mocked in tests).

griffinwork40 (commented):

---

## PR Review: feat(daemon): wire durable handoff store into elicitation path (#1416)

**Reviewed ref:** `7f85a63f6e8a483fb0000eeadd4e32abf4138bdd`
**Regime:** full (6 files, ~700 lines)
**Stated intent:** PR 2 of durable handoff feature — wire handoff-store into daemon elicitation and lease-recovery paths.

---

### Decision: DO NOT MERGE — 1 high, 3 medium blocking (1 security, 2 correctness); 3 low, 1 nit.

---

### Finding 1 — high · blocking:true · correctness

**`src/agent/daemon/handoff-wiring.ts:205–222`** · ref: 7f85a63f · file-state

When `recoverPendingHandoffs` expires a stale handoff (age > 24h), it transitions the handoff record's `status` to `'expired'` but never calls `setLeaseState` to clear the task's `waiting_human_input` lease state. The lease file remains in `leased/` with `state === 'waiting_human_input'` indefinitely. Because `recoverExpiredLeases` unconditionally skips tasks in that state (lease-store.ts:393), the task is permanently suppressed — never re-enqueued, never dead-lettered, never garbage-collected. The task is silently lost.

```ts
// handoff-wiring.ts — expiry branch
const expired: HandoffRecord = { ...record, status: 'expired' };
await writeHandoff(expired, handoffsDir);
result.expired += 1;
// ← no setLeaseState call; lease stays in waiting_human_input forever
```

**Suggestion:** After `writeHandoff(expired, ...)`, call `setLeaseState(record.taskId, 'leased', queueDir)` to restore the lease to a recoverable state. `recoverPendingHandoffs` does not currently accept `queueDir`, so that parameter needs threading through as well.

---

### Finding 2 — high · blocking:true · correctness

**`src/agent/daemon/handoff-wiring.ts:132–138`** · ref: 7f85a63f · file-state

`setLeaseState` failure is silently swallowed by an empty `catch {}`. If the lease-state update fails, the handoff record is already written as `pending` but the lease retains its `leased` state. On the next `recoverExpiredLeases` sweep the task is re-enqueued or dead-lettered while its pending handoff still exists — creating an orphaned pending handoff that triggers re-notification on every subsequent daemon restart for a question that can never be answered.

```ts
try {
  setLeaseState(opts.taskId, 'waiting_human_input', opts.queueDir);
} catch {
  // Silent — no log, no recovery
}
```

**Suggestion:** Log at `console.error` level so the inconsistency is detectable. Longer term, transition the handoff to `cancelled` when the lease state update fails.

---

### Finding 3 — medium · blocking:true · correctness

**`src/agent/daemon/handoff-wiring.ts:282–318`** · ref: 7f85a63f · file-state

`serializeRequest` drops `mode`, `url`, `elicitationId`, `requestedSchema`, `displayName`, and `description` from `ElicitationRequest` (all present in sdk-types.ts:200–206). Form-mode and URL-mode elicitations are valid in production (MCP tools use `mode: 'form'`), so the stored record is incomplete for any future re-presentation in PR 3.

**Suggestion:** Add the missing fields to the serializer, or switch to a spread + strip approach and add a test with `mode: 'form'` + `requestedSchema` set.

---

### Finding 4 — medium · blocking:true · security

**`src/agent/daemon/handoff-wiring.ts:258–260`** · ref: 7f85a63f · file-state

`answerHandoff` accepts `answer: unknown` and passes it directly to `updateHandoffAnswer` for persistence. No type narrowing, schema validation, or size bound is applied. A maliciously oversized or structurally unexpected answer written to the handoff store creates an untrusted deserialization surface when the answer is later consumed for task resumption.

```ts
export async function answerHandoff(
  taskId: string, answer: unknown, source: string, handoffsDir?: string
): Promise<boolean> {
  const result = await updateHandoffAnswer(taskId, answer, source, handoffsDir);
```

**Suggestion:** Enforce a maximum byte size on the serialized answer and validate structure against the expected `ElicitationResult` shape before persistence.

---

### Finding 5 — low · blocking:false · test-coverage

**`src/agent/daemon/handoff-wiring.test.ts`** · ref: 7f85a63f · file-state

The expiry test verifies `result.expired === 1` and `updated?.status === 'expired'` but does not assert that the task's lease state is cleared after expiry (this masks Finding 1). Separately, no test covers the `writeHandoff`-throws failure path.

---

### Finding 6 — low · blocking:false · spec-compliance

**`src/agent/daemon/scheduler.ts:359`** · ref: 7f85a63f · diff-context

`recoverPendingHandoffs()` is called with no arguments (uses global default), while all other scheduler paths (`queueDir`, `telemetryPath`, `stateDatabasePath`) are plumbed through `DaemonSchedulerOptions`. This breaks the injection pattern used by scheduler integration tests.

---

### Finding 7 — low · blocking:false · correctness

**`src/agent/daemon/handoff-wiring.ts:113–155`** · ref: 7f85a63f · file-state

`makeDaemonElicitationHandler` writes a new `HandoffRecord` without checking whether a pending record for the same `taskId` already exists. A second `writeHandoff` call atomically overwrites the first, silently discarding a question already notified to Telegram. The single-question-per-task assumption is undocumented. (waived: bounded — daemon tasks run serially within a lease, so concurrent asks from the same task are structurally rare; the overwrite is at-most-one-deep.)

---

### Finding 8 — nit · blocking:false · correctness

**`src/agent/daemon/handoff-wiring.ts:293–315`** · ref: 7f85a63f · file-state

`serializeRequest` uses bracket-notation type assertions (`(request as Record<string, unknown>)['questionDefault']`) for fields declared on `ElicitationRequest`. These are directly accessible without casting.

---

### What was not checked

- Citations verified inline against branch HEAD `7f85a63f6e8a483fb0000eeadd4e32abf4138bdd`.
- Telegram push without redaction (security-F1/F2) noted but NOT blocking: the existing `elicitation-router.ts:73–77` sends `request.message` via the same `pushIfConfigured` path without redaction. This PR is consistent with the pre-existing pattern — the gap is real but is a repo-wide concern, not a regression introduced by this diff.
- TOCTOU race in `setLeaseState` noted but NOT blocking: `renewLease` in the same file uses the identical read-modify-write pattern, establishing this as the existing concurrency model. Not a regression.
- Did not review `handoff-store.ts` internals (landed in PR 1). Did not run tests. Did not verify Telegram delivery end-to-end.

griffinwork40 (commented):

**Round 3** — follows Round 1 + Round 2 (both at `7f85a63f6e`). Reviewed ref: `f9da04e5d9416f1c9779ca1effc689f7b22b4e03`.

---

## Prior-round resolution summary

All six blocking findings from Round 1 and Round 2 are **resolved** in the fix commits:

| Prior Finding | Status |
|---|---|
| **C-1 HIGH** orphaned lease on handoff expiry | ✅ Fixed — `setLeaseState(record.taskId, 'leased', queueDir)` added in expiry branch |
| **C-2/F2 HIGH** silent catch on `setLeaseState` failure | ✅ Fixed — `console.error` logging added |
| **A-2 MEDIUM** second-pass claim recovery missing `waiting_human_input` guard | ✅ Fixed — guard added before crash-window reconstruction with rename-back |
| **T-1 MEDIUM** no test for lease state after handoff expiry | ✅ Fixed — new test `'clears lease waiting_human_input state when handoff expires'` |
| **F3 MEDIUM** missing serializer fields | ✅ Fixed — `serializeRequest` now covers all `ElicitationRequest` fields |
| **F4 MEDIUM** unbounded answer size | ✅ Fixed — `MAX_ANSWER_BYTES = 64 KB` guard added |
| **P-1 LOW** `recoverPendingHandoffs` result not logged | ✅ Fixed — `.then`/`.catch` logging in scheduler |
| **S-1 LOW** `DEFAULT_HANDOFF_TTL_MS` JSDoc inaccurate | ✅ Fixed |

---

## Decision: MERGE — 0 blocking (2 medium waived, 2 low, 1 nit)

Arithmetic: 2 medium (both waived with justification below); 2 low (non-blocking); 1 nit (non-blocking). Zero blocking findings → MERGE.

---

## Findings

### 🟡 MEDIUM — waived (non-blocking)

**C-R3-1 · correctness · `lease-store.ts:344–346` · ref: f9da04e5 · file-state**
The first-pass file filter in `recoverExpiredLeases` excludes `.tmp-*` and `.claim-*` files but does not exclude `.process-*` scratch files. `listActiveTasks` (same file, line ~551) does exclude `.process-*`, confirming the prefix is a recognized scratch sentinel. If a `.process-*` scratch file survives from a second-pass `waiting_human_input` rename-back failure, the first pass would read it as a regular lease and process it on the next recovery cycle.

**Evidence:**
```ts
const files = readdirSync(dir)
  .filter((f) => f.endsWith('.json') && !f.startsWith('.tmp-') && !f.startsWith('.claim-'));
// ← '.process-*' NOT excluded; contrast listActiveTasks which does exclude it
```

**Suggestion:** Add `&& !f.startsWith('.process-')` to the first-pass filter, mirroring `listActiveTasks`.

· waived: The `.process-*` file is a transient artifact created only during the second-pass recovery loop (`recoverExpiredLeases`) and normally deleted by `unlinkSync` on the same tick. A surviving `.process-*` file requires a process crash between the rename-to-process and the unlink — a sub-millisecond window. The worst case is one redundant re-enqueue of a single task on the next cycle, which the idempotent recovery path handles safely. Bounded, non-data-affecting, no user impact.

---

**C-R3-2 · correctness · `lease-store.ts:477–482` + `handoff-wiring.ts:229` · ref: f9da04e5 · file-state**
If the second-pass rename-back (`processPath → leasedPath`) fails silently, the lease file is not at its canonical path when `recoverPendingHandoffs`'s expiry branch later calls `setLeaseState(taskId, 'leased', queueDir)`. `setLeaseState` returns as a no-op (`existsSync(path)` → false), and the expiry log fires regardless, giving the operator false confidence the lease was restored.

**Evidence:**
```ts
// setLeaseState — silently no-ops when file missing:
if (!existsSync(path)) return;

// Expiry branch — log fires even if setLeaseState was a no-op:
setLeaseState(record.taskId, 'leased', queueDir);
result.expired += 1;
console.error(`[handoff-wiring] expired stale handoff for task ${record.taskId} ...`);
```

**Suggestion:** Have `setLeaseState` return a boolean indicating whether it found and updated the file. In the expiry branch, log a distinct warning when it returns false.

· waived: This failure path requires C-R3-1's precondition (a surviving `.process-*` file from a crash during the second-pass recovery of a `waiting_human_input` task). That scenario is already narrow. The rename-back must then also fail (EEXIST from a race with another process recreating the lease), creating a double-failure chain. The task is not silently lost — it persists in the handoff store as `expired` and in a `.process-*` scratch file that C-R3-1's fix would exclude from accidental re-processing. Bounded, single-task blast radius.

---

### 🔵 LOW — informational

**T-R3-1 · test-coverage · `lease-store.test.ts` · ref: f9da04e5 · file-state**
No test covers the second-pass `waiting_human_input` guard specifically — i.e. a `.claim-<hex>.json` file containing `state: 'waiting_human_input'` being processed by the stale-claim-recovery loop. The existing tests exercise the first-pass guard (live lease files) but not the second-pass code path (which renames to `.process-*` and back).

**Suggestion:** Add a test: write a `.claim-deadbeef.json` with `state: 'waiting_human_input'`; call `recoverExpiredLeases`; assert the task is not re-enqueued and the file is renamed back to `leased/<taskId>.json`.

---

**S-R3-1 · security · `handoff-wiring.ts:332` · ref: f9da04e5 · file-state**
`serializeRequest` passes `request.requestedSchema` (an arbitrary `Record<string, unknown>`) directly into the HandoffRecord without a size bound. A malformed MCP elicitation request with a large `requestedSchema` writes an unbounded JSON blob to disk. The `MAX_ANSWER_BYTES` guard covers the answer side but not the question side.

**Suggestion:** Apply a serialized-size guard to the `question` record before `writeHandoff`, mirroring the answer-side 64 KB cap.

---

### ⚪ NIT

**C-R3-3 · correctness · `lease-store.ts:487–497` · ref: f9da04e5 · file-state**
The crash-window reconstruction after the `waiting_human_input` guard hardcodes `state: 'leased'` — correct because the guard above already handled the only other non-initial state. A comment tying the two invariants together would clarify the ordering dependency for future readers.

**Suggestion:** Add: `// At this point state is guaranteed not to be 'waiting_human_input' (guarded above).`

---

## What was not checked

- Citations verified inline against branch HEAD `f9da04e5d9416f1c9779ca1effc689f7b22b4e03`.
- Stated intent: PR #1493 title + body — spec-compliance assessed: all stated wiring points are present; no unmet intent; no scope creep.
- Prior-round findings were re-verified against the current ref; all six blocking items are resolved.
- Did not review `handoff-store.ts` internals (landed in a prior PR, not part of this diff).
- Did not run tests.
- Did not verify Telegram transport behavior end-to-end (mocked in tests).

PR #1494, opened 2026-09-05
feat(subagent): mid-run agent steering via tool-call boundary injection (#1418)

griffinwork40:
## Problem

Running background agents cannot be steered mid-flight. `SubagentHandle.sendMessage()` queues messages into `_pendingUserMessages`, but the drain loop at `handle.ts:284-310` only executes **after** the child's primary `run()` completes. A parent that wants to redirect a running child ("skip hypothesis A, focus on B") has no mechanism to deliver that message until the child has already finished.

## Solution

Thread a `beforeNextRound` callback through the provider loop that fires at each tool-call boundary (after tool results are committed, before the next `openRound()`). Back it with a ring buffer on `SubagentHandleImpl`, and expose it via a new `send_message_to_agent` tool.

### Key components

| Component | File(s) | Purpose |
|-----------|---------|---------|
| `beforeNextRound` callback | `request-types.ts`, `query-options.ts`, `query-runtime.ts`, `query-turn-driver.ts`, `turn-request.ts` | Threaded through the anthropic-direct provider; fires synchronously at each round boundary |
| `loop/inter-round.ts` (new) | Provider loop extraction | Appends steering text to the last user turn's content array (mirrors `harnessNotes` pattern) + emits `queued_user_message` trace event |
| `steer()` + ring buffer | `handle.ts` | `_steeringMessages: string[]` (capacity 3, oldest-evict). Terminal-state/abort silent-drop mirrors `sendMessage()` |
| `send_message_to_agent` tool | `schemas.background-steer.ts`, `send-message.ts` | Same `jobId` pattern as `cancel_background_job`. Provenance guard refuses user-owned jobs |
| `IAgentSession.setBeforeNextRound` | `session-types.ts`, `agent-session.ts` | Bridges handle -> provider. Delegated to `ProviderQuery.setBeforeNextRound` |
| `BackgroundAgentRegistry.getHandle()` | `background-registry.ts` | Exposes live handle from the private `InternalJob` for the tool executor |

### Scope

- **Background-only in V1.** The parent model cannot steer foreground subagents: it is blocked on `await runToResult` and has no model turns to invoke `send_message_to_agent`. Foreground subagents are steered by the **human** via the existing interrupt picker (Ctrl+C → Steer → readline → next REPL turn). See `docs/subagent-steering.md` for the two-domain model.
- **Anthropic-direct provider.** OpenAI-compatible deferred (the optional-chaining `setBeforeNextRound?.()` no-ops safely).
[...]

Issue #1418, opened 2026-09-01
feat(core): addressable agent messaging and reactive coordination -- mid-run steering, progress events, workspace subscriptions


griffinwork40:
## Problem

Running agents cannot be steered mid-flight, report partial progress to their parent, or reactively receive workspace updates. The existing coordination primitives are either one-directional, poll-based, or abort-only:

1. **`SubagentHandle.sendMessage()`** (`src/agent/subagent/handle.ts:97,548-564`): one-directional parent-to-child queue. Messages are pushed to `_pendingUserMessages` (line 541) and drained only *after* the child's primary `run()` completes (lines 284-310). The child cannot receive a mid-run steering message; the parent cannot receive a response until the child's entire turn finishes.

2. **Shared workspace** (`src/agent/workspace/workspace-store.ts`): poll-based. Siblings publish findings and query them, but there's no notification when new entries appear. A sibling that publishes a critical finding must wait for other siblings to poll `workspace_query`.

3. **AbortGraph** (`src/agent/abort-graph.ts:138-162`): abort-only. Parent abort cascades down, child abort notifies up, but no payload beyond the abort signal itself. No custom events, no broadcast channel.

4. **DAG node communication** (`src/agent/dag.ts`, `src/agent/dag-subagent.ts`): nodes communicate only through the shared workspace. No direct node-to-node output passing; the DAG executor does not inject one node's result into a downstream node's prompt.

### User impact

- A parent dispatching a long `/diagnose` run cannot steer the child ("skip hypothesis A, focus on B") without aborting and re-dispatching.
- A `/compose` DAG where node B depends on node A's output: node B gets A's result only if A published to the workspace AND B queries at the right time. There's no guaranteed delivery of A's findings to B.
- An operator monitoring a running agent via Telegram or the web UI cannot see intermediate progress -- only the final result after the agent completes.
- Sibling agents doing parallel investigation may waste work re-reading files that another sibling already analyzed and published, because there's no push notification of new workspace entries.

## Verified current behavior

| Component | File | Behavior |
|-----------|------|----------|
| sendMessage (parent->child) | `src/agent/subagent/handle.ts:97,541,548-564,284-310` | One-directional push queue. Messages drain after child's primary run completes, not mid-turn |
| _pendingUserMessages drain | `src/agent/subagent/handle.ts:284-310` | After primary run, iterates pending messages as new user turns. `_drainingMessages` flag keeps window open |
[...]

PR #1504, opened 2026-09-05
feat(subagent): child-to-parent progress events via emit_progress tool (#1418)

griffinwork40:
## Summary

Adds opt-in child-to-parent progress events -- the reverse direction of the steering feature shipped in PR #1494. When a parent dispatches a child with `progress_events: true`, the child gains an `emit_progress` tool that pushes structured updates to the parent at the next user-turn boundary.

Addresses **item 2** of #1418 (progress events). Item 1 (mid-run steering) shipped in #1494. Item 3 (reactive workspace subscriptions) remains open.

## How it works

1. Parent dispatches with `progress_events: true` on the `agent` tool
2. Child's tool surface gains `emit_progress({ message, phase?, metadata? })`
3. Handler pushes to a per-handle ring buffer (capacity 20, oldest-first eviction)
4. Formatted as `<child-progress subagentId="..." phase="..." timestamp="...">message</child-progress>`
5. Delivered to parent via `queueFrameworkContext` (prepends to next user turn)
6. Buffer cleared on cancel/teardown

## Design decisions

- **Opt-in, not always-on**: tool excluded entirely when not opted in (not a silent no-op)
- **`queueFrameworkContext` over `pushUserMessage`**: prepends to next real user message, never displaces it -- same channel as `SubagentStop.injectContext`
- **Ring buffer, not unbounded**: capacity 20, 2KB per message, oldest-first eviction -- matches `_steeringMessages` pattern
- **Mutable config injection**: `wireProgressEvents` mutates `customTools` in place (the config is already a fresh shallow copy from `assembleChildConfig`), returning only a `bindHandle` callback -- keeps `subagent.ts` surface to 3 net code lines

## Files

| File | Change |
|------|--------|
| `src/agent/tools/subagent/emit-progress.ts` | New -- handler factory, XML formatter, ring buffer constants, UTF-8-safe truncation |
| `src/agent/subagent/fork-progress-events.ts` | New -- config injection + mutable-ref handle binding via Zod custom tool |
| `src/agent/tools/subagent/emit-progress.test.ts` | New -- 21 tests (ring buffer, truncation, abort guard, format, delivery) |
| `src/agent/subagent/handle.ts` | `_progressEvents` buffer + `emitProgress()` method + cleanup in cancel/teardown |
| `src/agent/subagent/fork-types.ts` | `progressEvents?: boolean` on `ForkSubagentOptions` |
| `src/agent/tools/subagent/input-parse.ts` | Parse `progress_events` from agent tool input |
| `src/agent/tools/schemas.ts` | `progress_events` in agent tool JSON schema |
| `src/agent/tools/subagent-executor.ts` | Thread `progressEvents` to `forkSubagent()` |
| `src/agent/subagent.ts` | `wireProgressEvents` call + import consolidation to stay within filesize baseline |

## Tests

- 21 new tests in `emit-progress.test.ts` covering:
  - Ring buffer push/eviction at capacity
  - Message truncation at 2KB UTF-8 boundary
  - XML envelope formatting with/without phase
  - Abort signal guard
  - Terminal-state guard (no emit after succeeded/failed/cancelled)
  - Handler error paths (missing message, non-object input)
- All 299 existing subagent tests pass unchanged

### DIFF
--- a/src/agent/subagent.ts
+++ b/src/agent/subagent.ts
@@ -47,6 +47,7 @@ import { resolveForkInputs } from './subagent/fork-resolution.js';
 import type { SubagentStatus, SubagentResult, SubagentTrace } from './subagent/result.js';
 import { CompletedCache } from './subagent/completed-cache.js';
 import { SubagentLogWriter } from './subagent/log.js';
+import { wireProgressEvents } from './subagent/fork-progress-events.js';
 
 // Re-export types for public API
 export type { SubagentStatus, SubagentResult, SubagentTrace, SubagentHandle };
@@ -56,13 +57,8 @@ export type { SubagentStatus, SubagentResult, SubagentTrace, SubagentHandle };
 // use by SubagentManager below AND re-exported here unchanged, so every
 // existing `from './subagent.js'` importer keeps working untouched.
 import {
-  DENY_ELICITATION,
-  SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS,
-  SUBAGENT_DEFAULT_TIMEOUT_MS,
   SUBAGENT_DEFAULT_IDLE_TIMEOUT_MS,
-  SUBAGENT_BACKGROUND_TIMEOUT_MS,
   SUBAGENT_DRAIN_TIMEOUT_MS,
-  resolveSubagentTimeoutMs,
   resolveSubagentIdleTimeoutMs,
 } from './subagent/constants.js';
 
@@ -72,14 +68,18 @@ import type {
   SubagentManagerOptions,
 } from './subagent/fork-types.js';
 
+// Re-export-only symbols forwarded without local use.
 export {
   DENY_ELICITATION,
   SUBAGENT_DEFAULT_MAX_TOOL_USE_ITERATIONS,
   SUBAGENT_DEFAULT_TIMEOUT_MS,
-  SUBAGENT_DEFAULT_IDLE_TIMEOUT_MS,
   SUBAGENT_BACKGROUND_TIMEOUT_MS,
-  SUBAGENT_DRAIN_TIMEOUT_MS,
   resolveSubagentTimeoutMs,
+} from './subagent/constants.js';
+// Locally-used symbols also re-exported for public API.
+export {
+  SUBAGENT_DEFAULT_IDLE_TIMEOUT_MS,
+  SUBAGENT_DRAIN_TIMEOUT_MS,
   resolveSubagentIdleTimeoutMs,
 };
 export type { ForkParent, ForkSubagentOptions, SubagentManagerOptions };
@@ -414,6 +414,9 @@ export class SubagentManager {
       stopOccupancyHeartbeat = startWorktreeOccupancyHeartbeat(childConfig.cwd);
     }
 
+    // Progress-events opt-in — mutates childConfig.customTools in place when enabled.
+    const bindProgressHandle = wireProgressEvents(childConfig, options.progressEvents, options.parent.getInputStreamRef?.(), options.parent.abortSignal);
+
     // Ordering constraint: the heartbeat armed above is disarmed by the settle
     // callback installed on the handle built below, so the guarded span has to
     // run from construction all the way through `active.set`. A throw anywhere
@@ -518,6 +521,7 @@ export class SubagentManager {
         options.config.idleTimeoutMs ?? resolveSubagentIdleTimeoutMs(),
       );
       if (logWriter) handle._logWriter = logWriter;
+      bindProgressHandle(handle as SubagentHandleImpl<unknown>);
       this.active.set(id, handle as SubagentHandleImpl<unknown>);
     } catch (err) {
       // Construction or manager-wiring failed (invalid model, sync init
--- a/src/agent/subagent/fork-progress-events.test.ts
+++ b/src/agent/subagent/fork-progress-events.test.ts
@@ -1,0 +1,116 @@
+/**
+ * Unit tests for wireProgressEvents: the opt-in wiring that injects the
+ * emit_progress custom tool into a child's AgentConfig.
+ *
+ * Covers the four cases the review identified as untested:
+ *   (a) disabled path leaves config.customTools unchanged
+ *   (b) enabled path appends a tool named emit_progress
+ *   (c) returned bindProgressHandle sets the ref so handler succeeds
+ *   (d) calling handler before bindProgressHandle returns not-initialized error
+ */
+
+import { describe, it, expect } from 'vitest';
+import { wireProgressEvents } from './fork-progress-events.js';
+import type { AgentConfig } from '../types.js';
+
+/**
+ * Build a minimal AgentConfig stub with only the fields wireProgressEvents
+ * reads and mutates (customTools).
+ */
+function makeConfig(customTools?: AgentConfig['customTools']): AgentConfig {
+  return { customTools } as unknown as AgentConfig;
+}
+
+/** Minimal handle double matching the SubagentHandleImpl surface emit_progress uses. */
+function makeHandleDouble() {
+  const _progressEvents: Array<{ message: string }> = [];
+  return {
+    id: 'test-handle',
+    _currentStatus: 'running' as const,
+    _progressEvents,
+    emitProgress(payload: { message: string }): void {
+      if (
+        this._currentStatus === 'succeeded' ||
+        this._currentStatus === 'failed' ||
+        this._currentStatus === 'cancelled'
+      ) {
+        return;
+      }
+      _progressEvents.push(payload);
+    },
+  } as Parameters<ReturnType<typeof wireProgressEvents>>[0];
+}
+
+describe('wireProgressEvents', () => {
+  // (a) Disabled path
+  it('returns a no-op bindHandle and does not modify config when enabled is falsy', () => {
+    const config = makeConfig();
+    const bindHandle = wireProgressEvents(config, false, undefined, undefined);
+    expect(config.customTools).toBeUndefined();
+    // bindHandle is a no-op — calling it should not throw
+    bindHandle(makeHandleDouble());
+  });
+
+  it('returns a no-op bindHandle when enabled is undefined', () => {
+    const config = makeConfig();
+    const bindHandle = wireProgressEvents(config, undefined, undefined, undefined);
+    expect(config.customTools).toBeUndefined();
+    bindHandle(makeHandleDouble());
+  });
+
+  it('preserves existing customTools when disabled', () => {
+    const existingTool = { schema: { name: 'existing' } } as AgentConfig['customTools'] extends (infer T)[] | undefined ? T : never;
+    const config = makeConfig([existingTool]);
+    wireProgressEvents(config, false, undefined, undefined);
+    expect(config.customTools).toHaveLength(1);
+    expect(config.customTools![0]).toBe(existingTool);
+  });
+
+  // (b) Enabled path appends emit_progress
+  it('appends an emit_progress tool to config.customTools when enabled', () => {
+    const config = makeConfig();
+    wireProgressEvents(config, true, undefined, undefined);
+    expect(config.customTools).toHaveLength(1);
+    expect(config.customTools![0]!.schema.name).toBe('emit_progress');
+  });
+
+  it('preserves existing customTools and appends emit_progress', () => {
+    const existingTool = { schema: { name: 'existing' } } as AgentConfig['customTools'] extends (infer T)[] | undefined ? T : never;
+    const config = makeConfig([existingTool]);
+    wireProgressEvents(config, true, undefined, undefined);
+    expect(config.customTools).toHaveLength(2);
+    expect(config.customTools![0]!.schema.name).toBe('existing');
+    expect(config.customTools![1]!.schema.name).toBe('emit_progress');
+  });
+
+  // (c) bindProgressHandle sets ref so handler succeeds
+  it('returns { delivered: true } after bindProgressHandle is called', async () => {
+    const queueCalls: string[] = [];
+    const parentRef = {
+      pushUserMessage: () => {},
+      queueFrameworkContext: (text: string) => { queueCalls.push(text); },
+    };
+    const config = makeConfig();
+    const bindHandle = wireProgressEvents(config, true, parentRef, undefined);
+    const handle = makeHandleDouble();
+    bindHandle(handle);
+
+    // Invoke the tool handler directly
+    const toolDef = config.customTools![0]!;
+    const result = await toolDef.handler({ message: 'test progress' }, new AbortController().signal);
+    expect(result.isError).toBeUndefined();
+    expect(JSON.parse(result.content as string)).toEqual({ delivered: true });
+    expect(queueCalls).toHaveLength(1);
+    expect(queueCalls[0]).toContain('<child-progress');
+  });
+
+  // (d) Handler before bindProgressHandle returns not-initialized error
+  it('returns isError when handler is called before bindProgressHandle', async () => {
+    const config = makeConfig();
+    wireProgressEvents(config, true, undefined, undefined);
+    const toolDef = config.customTools![0]!;
+    const result = await toolDef.handler({ message: 'too early' }, new AbortController().signal);
+    expect(result.isError).toBe(true);
+    expect(String(result.content)).toContain('handle not yet initialized');
+  });
+});
--- a/src/agent/subagent/fork-progress-events.ts
+++ b/src/agent/subagent/fork-progress-events.ts
@@ -1,0 +1,107 @@
+/**
+ * Progress-events opt-in wiring for forked subagents.
+ *
+ * When a parent dispatches a child with `progressEvents: true`, the child's
+ * `AgentConfig.customTools` is extended with the `emit_progress` tool. This
+ * module owns that injection so the main `SubagentManager.forkSubagent` method
+ * does not grow unbounded.
+ *
+ * The injection must happen BEFORE `new AgentSession(childConfig)` but the
+ * tool handler requires access to the not-yet-constructed `SubagentHandleImpl`.
+ * A mutable ref object bridges the two: `applyProgressEventsToConfig` returns
+ * both the augmented config and a ref whose `.current` field must be set by the
+ * caller immediately after `SubagentHandleImpl` is constructed.
+ *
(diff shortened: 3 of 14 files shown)


griffinwork40:
Good -- the provider loops do emit `tool_call` trace events for every tool call via the dispatch layer. Custom tools like `emit_progress` will go through the same dispatch path and emit trace events. So the "no trace" absence claim from the wave 1 agent is **false-absent** -- custom tools ARE traced through the provider's standard dispatch layer. Dropping that finding.

Now I have all I need for Wave 2 synthesis. Let me compile the verified findings:

- XML body escaping (emit-progress.ts:40-46): **verified** -- `payload.message` raw in body confirmed at ref
- Per-call handler recreation (fork-progress-events.ts:84-100): **verified** -- `createEmitProgressHandler` called in callback confirmed at ref
- Inline capacity constant (handle.ts:624-626): **verified** -- `const CAPACITY = 20` confirmed at ref
- Silent delivery failure (emit-progress.ts:114-125): **verified** -- `catch {}` with `{ delivered: true }` confirmed at ref
- "Does NOT count" claim (fork-progress-events.ts:72): **verified** -- no exemption exists; `turn.iterations += 1` at tool-round.ts:95 counts all rounds uniformly
- progress_events false-drop (input-parse.ts:543): **verified** -- conditional spread confirmed at ref
- No trace events absence: **false-absent** -- custom tools go through standard tool dispatch which emits `tool_call` trace events. **DROPPED.**
- Absence: no test for throwing queueFn: **confirmed-absent**
- Absence: no test for handle-not-initialized guard: **confirmed-absent**
- Absence: no test for pushUserMessage fallback: **confirmed-absent**
- Absence: no test for multi-byte truncation: **confirmed-absent**
- Absence: no test for XML-significant chars in message body: **confirmed-absent**

---

## Review: PR #1504 -- feat(subagent): child-to-parent progress events via `emit_progress` tool

**Target:** PR #1504 (`afk/iso-impl-progress-events-1-jjejyz` -> `main`)
**Reviewed ref:** `13be94f9`
**Change type:** feature | +564/-10 | 9 files
**Regime:** full

---

### Findings

**1. medium · blocking:true · high · security+correctness · `src/agent/tools/subagent/emit-progress.ts:43-44` · ref:13be94f9 · diff-context**

`formatProgressEvent` writes `payload.message` into the `<child-progress>` XML element body without escaping `<`, `>`, or `&`. A child emitting a message containing those characters produces malformed XML. A prompt-injected child sets `message: "</child-progress><injected-tag>…"` and the parent's framework context sees a structurally different document.

```ts
    `<child-progress subagentId="${escapeAttr(subagentId)}"${phaseAttr} timestamp="${timestamp}">` +
    payload.message +   // no XML body escaping
```

**suggestion:** Apply body-escaping (`s.replace(/&/g,'&amp;').replace(/</g,'&lt;').replace(/>/g,'&gt;')`) to `payload.message` before interpolation, matching the attribute escaping already applied to `subagentId` and `phase`.

---

**2. medium · blocking:true · high · spec-compliance · `src/agent/subagent/fork-progress-events.ts:72` · ref:13be94f9 · diff-context**

The tool description claims "This tool does NOT count against maxToolUseIterations -- it is meta, not work." No implementing change exempts `emit_progress` from the round counter. `turn.iterations += 1` in `tool-round.ts:95` fires per round regardless of which tools were called. A child making N progress calls exhausts N of its tool-round budget, contradicting the advertised contract.

```ts
'This tool does NOT count against maxToolUseIterations — it is meta, not work.',
```

**suggestion:** Either add an exemption path for `emit_progress` in the round counter (checked in `shouldWindDown`), or remove the "does NOT count" claim from the tool description.

---

**3. medium · blocking:true · high · correctness · `src/agent/tools/subagent/emit-progress.ts:116-123` · ref:13be94f9 · diff-context**

`parentQueueFn(envelope)` is called after `handle.emitProgress(payload)` succeeds, but if `parentQueueFn` throws, the `catch {}` swallows the error silently and returns `{ delivered: true }`. The child believes delivery succeeded when the parent never received the event.

```ts
handle.emitProgress(payload);
try {
  const envelope = formatProgressEvent(payload, handle.id);
  parentQueueFn(envelope);
} catch {
  // Delivery failure is non-fatal to the child.
}
return { content: JSON.stringify({ delivered: true }) };
```

**suggestion:** Return `{ delivered: false, reason: 'queue_error' }` (not `isError`) when `parentQueueFn` throws, so the child model can detect the failure.

---

**4. medium · blocking:true · high · test-coverage · `src/agent/tools/subagent/emit-progress.test.ts` (absence) · ref:13be94f9 · diff-context**

Three untested code paths in the new handler:
- (a) `parentQueueFn` throws -- ring buffer populated, `{ delivered: true }` returned, error swallowed.
- (b) `capturedRef.current === undefined` guard -- the window between tool injection and `bindProgressHandle`.
- (c) `pushUserMessage` fallback in `fork-progress-events.ts:93-95` -- when `parentInputStreamRef` lacks `queueFrameworkContext`.

Search patterns (`queueFn.*throw`, `handle not yet initialized`, `pushUserMessage`) all return zero matches in the test file.

**suggestion:** Add three tests: (a) throwing `queueFn`, (b) invocation before `bindProgressHandle`, (c) `pushUserMessage` fallback when `queueFrameworkContext` is absent.

---

**5. low · blocking:false · high · correctness · `src/agent/subagent/handle.ts:624-626` · ref:13be94f9 · diff-context**

The ring buffer capacity constant `20` is inlined as a numeric literal with a comment saying it is "kept in sync with PROGRESS_RING_CAPACITY in emit-progress.ts." No import, assertion, or CI gate enforces synchrony; a future edit to `PROGRESS_RING_CAPACITY` silently diverges the two.

```ts
// Inline the capacity constant to avoid a circular/async import.
// Kept in sync with PROGRESS_RING_CAPACITY in emit-progress.ts.
const CAPACITY = 20;
```

**suggestion:** Move `PROGRESS_RING_CAPACITY` to a shared constants file importable by both modules without circularity.

---

**6. nit · blocking:false · high · correctness · `src/agent/subagent/fork-progress-events.ts:88-99` · ref:13be94f9 · diff-context**

`createEmitProgressHandler` is reconstructed on every `emit_progress` tool invocation (inside the `async (input, signal)` closure) rather than once at wiring time. Allocates a new closure per call; the factory was designed to amortize that cost.

```ts
const handlerFn = createEmitProgressHandler(
  capturedRef.current,
  (text) => { ... },
  parentAbortSignal,
);
return handlerFn(input, signal);
```

**suggestion:** Construct the handler lazily once on first invocation after `capturedRef.current` is set, then reuse it.

---

**7. nit · blocking:false · high · spec-compliance · `src/agent/tools/subagent/input-parse.ts:543` · ref:13be94f9 · diff-context**

The spread condition `progress_events !== undefined && progress_events` silently drops `progress_events: false` from the returned `AgentInput`. The type declares `progress_events?: boolean`, so `false` is a valid explicit opt-out that downstream code cannot distinguish from "not supplied."

```ts
...(progress_events !== undefined && progress_events ? { progress_events } : {}),
```

**suggestion:** `...(progress_events !== undefined ? { progress_events } : {})` -- let `false` propagate. `wireProgressEvents(!enabled)` already handles it correctly.

---

**8. low · blocking:false · high · test-coverage · `src/agent/tools/subagent/emit-progress.test.ts` (absence) · ref:13be94f9 · diff-context**

No test covers truncation of multi-byte UTF-8 messages (the boundary where `buf.subarray(0, maxBytes).toString('utf8')` drops an incomplete codepoint). All truncation tests use single-byte ASCII.

**suggestion:** Add a test with a string whose last codepoint before `PROGRESS_MAX_MESSAGE_BYTES` is a 3-byte CJK character; verify valid UTF-8 output within the byte cap.

---

**9. low · blocking:false · high · test-coverage · `src/agent/tools/subagent/emit-progress.test.ts` (absence) · ref:13be94f9 · diff-context**

No test covers `formatProgressEvent` with XML-significant characters (`<`, `>`, `&`) in `payload.message`. Given finding #1, this means the body-escaping bug is also undetected by the suite.

**suggestion:** Add a test asserting that `<`, `>`, and `&` in the message are escaped in the formatted envelope.

---

### API compatibility

No breaking changes. The import/re-export split in `subagent.ts:68-83` preserves all eight public export names unchanged. `progress_events` is additive (new optional field on `agent` tool schema + `ForkSubagentOptions`). Confirmed via grep: all existing importers of the moved symbols still resolve.

### Spec compliance

Stated intent available (PR #1504 title + body). Assessed against it:
- **Unmet:** The "does NOT count against maxToolUseIterations" claim is not implemented (finding #2).
- **Scope creep:** None detected. All changes map to the stated feature.

### What was not checked

- Citations verified inline against branch HEAD `13be94f9`.
- Stated intent: PR #1504 title + body -- spec-compliance assessed.
- Did not run tests or build.
- `perf-observability` finding claiming no trace events was **dropped** (false-absent: custom tools emit `tool_call` trace events via the standard provider dispatch path).
- Integration behavior (e.g., actual delivery through `queueFrameworkContext` at runtime) not verified.
- `metadata` size bounding not investigated further -- noted as low-confidence low-severity.

---

### Decision: DO NOT MERGE -- 1 medium security/correctness blocking (XML injection), 1 medium spec-compliance blocking (tool description claims exemption not implemented), 1 medium correctness blocking (silent delivery-failure lie), 1 medium test-coverage blocking (3 untested paths); 2 low, 2 nit.

---

griffinwork40:
This confirms: the `customTools` flow is `opts.customTools` at provider construction -> `this.customTools` field -> passed to `buildDispatcher()` which builds the `SessionToolDispatcher`. Since `createChildProviderFactory` never passes `customTools` to the provider constructor, the tool is invisible to the child.

This is a **confirmed high-severity correctness bug**: the feature's core contract is broken on the production dispatch path.

Now let me synthesize the final review.

---

# PR #1504 Review: feat(subagent): child-to-parent progress events via emit_progress tool

**Target:** `afk/iso-impl-progress-events-1-jjejyz` -> `main`
**Reviewed ref:** `8d26d4f3`
**Regime:** full (736 lines, 11 files, feature)
**Stated intent:** PR title + body -- child-to-parent progress events with opt-in `emit_progress` tool, ring buffer, queueFrameworkContext delivery.

---

## Findings

### 1. `high` · blocking:true · confidence:high · correctness + spec-compliance · `src/agent/tools/nesting.ts:239-269` + `src/agent/tools/subagent/child-config.ts:460-473` + `src/agent/providers/anthropic-direct/provider-runtime.ts:192` · ref:8d26d4f3 · citation-type:file-state

**`emit_progress` is silently absent from the child's tool surface on the normal `agent`-tool dispatch path, defeating the feature's core contract.**

`wireProgressEvents` correctly appends the tool to `childConfig.customTools`. But for every fork via the `agent` tool, `child-config.ts:460` builds the child's provider via `childProviderFactory({...})` (from `createChildProviderFactory` in `nesting.ts`) WITHOUT passing `customTools`. The `AnthropicDirectProvider` constructor reads `customTools` exclusively from its `opts` parameter (`provider-runtime.ts:192: this.customTools = opts.customTools ?? []`), storing an empty array. When the child session starts, it takes the `if (this.config.provider)` branch at `agent-session.ts:333`, which calls `this.config.provider.query(...)` -- and the provider's dispatcher was built without `emit_progress`. The child model never sees the tool in its schema. The feature silently no-ops.

```typescript
// nesting.ts:239 — providerOpts has no customTools key
const providerOpts = {
  permissions: { allowedTools: allowedTools ?? CHILD_ALLOWED_TOOLS },
  subagentExecutor: childExecutor,
  // ... no customTools here
};
return new AnthropicDirectProvider({ ...providerOpts, ... });
```

```typescript
// child-config.ts:460 — no customTools in args
childConfig.provider = args.childProviderFactory({
  childExecutor,
  // ... no customTools forwarded
});
```

**Suggestion:** In `createChildProviderFactory`, accept a `customTools` parameter in `ChildProviderFactoryArgs` and spread it into both provider constructors. In `child-config.ts`, pass `customTools: childConfig.customTools` through to the factory call. This is the narrowest fix -- it threads the tool through the exact gap without changing the provider contract.

---

### 2. `medium` · blocking:false · confidence:high · security · `src/agent/tools/subagent/emit-progress.ts:104-107` · ref:8d26d4f3 · citation-type:file-state · waived: bounded DoS from a child that is itself resource-capped; no data-integrity or auth impact

**The `phase` field has no byte cap. `message` is hard-truncated at 2048 bytes; `phase` passes through with only a `.trim()` check.** A misbehaving child can push an arbitrarily large string into the parent's input stream via the `phase` attribute. The Zod schema (`z.string().optional()`) has no `.max()` constraint. The XML escaping is correct (no injection risk), so the impact is resource-bounded: one oversized string per emit, capped by the ring buffer's 20-entry capacity and the child's own tool-round budget.

```typescript
const phase =
  typeof raw['phase'] === 'string' && raw['phase'].trim().length > 0
    ? raw['phase'].trim()   // no byte cap
    : undefined;
```

**Suggestion:** Apply `truncateToBytes(phase, 256)` and add `.max(256)` to the Zod schema.

---

### 3. `medium` · blocking:true · confidence:high · test-coverage · (absence claim -- confirmed by grep) · ref:8d26d4f3

**No test covers the end-to-end wiring from `wireProgressEvents` through the provider to a live tool surface.** `emit-progress.test.ts` tests the handler factory in isolation via a handle double. The closure test at line 317 replicates the `fork-progress-events.ts` logic inline rather than exercising the actual function. `src/agent/subagent/fork-progress-events.test.ts` does not exist. The critical path -- `wireProgressEvents(config, true, ...)` -> `config.customTools` contains the tool -> provider builds dispatcher with it -> model sees it -- is untested.

`grep -rn "wireProgressEvents\|fork-progress-events" src/agent/ --include="*.test.ts"` returns only the inline replication note in `emit-progress.test.ts`.

**Suggestion:** Add a unit test for `wireProgressEvents`: (a) disabled path leaves `config.customTools` unchanged, (b) enabled path appends a tool named `emit_progress`, (c) returned `bindProgressHandle` sets the ref so a subsequent handler call succeeds. An integration test that verifies `customTools` survival through the provider build would have caught Finding 1.

---

### 4. `medium` · blocking:false · confidence:high · perf-observability · `src/agent/tools/subagent/emit-progress.ts:1-153` · ref:8d26d4f3 · citation-type:file-state · waived: no production behavior impact until the feature works (blocked by Finding 1); no data loss or unsafe state

**Progress events delivered via `queueFrameworkContext` are invisible to the witness trace.** No `witness.write()` or `emitSubagentLifecycle` call records that a delivery happened, its timestamp, byte size, or child ID. For debugging ("the child emitted 3 events but the parent saw 0"), there is no durable record -- only the ring buffer on the handle (cleared on cancel/teardown).

**Suggestion:** Emit a witness event in `createEmitProgressHandler` after `parentQueueFn(envelope)` succeeds, recording `subagentId`, `phase`, and envelope byte length. Fire-and-forget.

---

### 5. `low` · blocking:false · confidence:high · correctness · `src/agent/subagent/fork-progress-events.ts:61-62` · ref:8d26d4f3 · citation-type:file-state

**`capturedRef` is an unnecessary alias of `handleRef` -- both reference the same `{ current: undefined }` object.** The alias serves no isolation purpose. Additionally, `cachedHandler` is lazily initialized and never cleared, so if a handle were ever replaced the cached handler retains the original.

```typescript
const handleRef: { current: ... } = { current: undefined };
const capturedRef = handleRef; // same object — no isolation
```

**Suggestion:** Remove `capturedRef`; use `handleRef` directly. Reset `cachedHandler = undefined` in the `bindProgressHandle` callback.

---

### 6. `low` · blocking:false · confidence:high · security · `src/agent/tools/subagent/emit-progress.ts:117-128` · ref:8d26d4f3 · citation-type:file-state

**Ring buffer is populated before delivery is attempted.** When `parentQueueFn` throws, the handler returns `{ delivered: false }` but the payload is already in `_progressEvents`. The test at line 246 explicitly asserts this ordering as deliberate. A child that retries on `delivered: false` creates a duplicate buffer entry.

**Suggestion:** Document at-least-once/advisory semantics on the ring buffer, or reverse the ordering (deliver first, buffer on success).

---

### 7. `nit` · blocking:false · confidence:high · correctness · `src/agent/tools/subagent/emit-progress.ts:143-152` · ref:8d26d4f3 · citation-type:file-state

**`truncateToBytes` inline comment is imprecise.** The algorithm is correct -- the `subarray(0, end)` exclusion properly drops incomplete codepoints. But the comment says "find the start of a complete codepoint" when the invariant is "find the start of an **incomplete** codepoint and exclude it."

**Suggestion:** `// Walk back until buf[end] is the lead byte of an incomplete sequence; subarray(0,end) excludes it entirely.`

---

### 8. `nit` · blocking:false · confidence:high · correctness · `src/agent/tools/subagent/emit-progress.ts:86-88` · ref:8d26d4f3 · citation-type:file-state

**The `signal` parameter in `createEmitProgressHandler`'s return type is accepted but never used.** The parent abort signal is checked separately. The child's per-call signal is irrelevant to parent-targeted delivery.

**Suggestion:** Prefix with `_signal` or remove from the signature.

---

### 9. `nit` · blocking:false · confidence:high · api-compat · `src/agent/subagent.ts:68-86` · ref:8d26d4f3 · citation-type:file-state

**Re-export refactor splits one `export {}` block into two.** All eight public symbols remain exported under their original names. No breaking change. Additive-only.

---

## What was not checked

- Citations verified inline against branch HEAD `8d26d4f3`.
- Stated intent: PR #1504 title + body -- spec-compliance assessed.
- Did not verify OpenAI-compatible provider path (same gap likely exists there for `customTools` threading).
- Did not review Telegram or daemon surface integration.
- Finding 1 (the `high` blocking finding about silent tool absence) was verified by the orchestrator via direct file reads of `nesting.ts`, `child-config.ts`, `provider-runtime.ts`, and `agent-session.ts` at the reviewed ref. The claim chain was independently confirmed at each hop.

---

## Decision: DO NOT MERGE -- 1 high blocking (correctness: feature silently no-ops on production path); 1 medium blocking (test-coverage: no wiring test); 2 medium waived (security: unbounded phase field, bounded by child resource caps; perf-observability: no witness trace, no production impact until feature works); 3 low; 3 nit.

**The blocking high (Finding 1) invalidates the feature's core contract.** The fix is narrow: `createChildProviderFactory` in `nesting.ts` needs to accept and forward `customTools` to the provider constructor, and `child-config.ts` needs to pass `childConfig.customTools` through the factory call. An integration test (Finding 3) exercising this exact path would have caught it and should ship alongside the fix.

---

PR #1535, opened 2026-09-07
fix(repl): warn on Ctrl+B promotion of non-isolated write-capable child

### LINKED ISSUE #1513: bug: Ctrl+B promotion leaves two concurrent writers in shared worktree
## Summary

When a user presses Ctrl+B to promote a foreground subagent to background, the promoted child continues running in the parent's worktree with no write coordination. The user immediately gets their prompt back and can start a new turn that also writes to the same worktree. Two concurrent writers share a tree with no locking — last write wins silently.

## The gap

The `isolation: "worktree"` path is well-armored: the child gets its own tree, it's `git worktree lock`'d against the sweep, and torn down on completion. But a plain `agent({ prompt: "..." })` dispatch (no isolation) inherits the parent's cwd. When that child is promoted via Ctrl+B:

1. The child's `AbortSignal` is detached from the parent turn (`foreground-promotion.ts:231`)
2. The child keeps running with `cwd = parentWorktreePath`
3. The user gets the prompt back and can start a new turn
4. Both the background child and the new turn can call `edit_file`, `write_file`, `bash` against the same files

`edit_file` uses exact-string-match replacement, so a concurrent edit *may* fail if the other writer changed the target string first (the safer failure mode). But `write_file` and `bash` have no such guard.

## Scenarios

- **Single promoted child + user turn**: Medium risk. The child is usually wrapping up, not starting new edits. But if it's mid-way through a multi-file change, the user's new turn can interleave.
- **Multiple promoted children**: Higher risk but rare — Ctrl+B only promotes the active foreground, so this requires sequential promote-then-dispatch cycles.

## Suggested mitigations (non-exclusive)

1. **Warn on promotion of a write-capable, non-isolated child**: Surface a TUI notice like "⚠ Background child is writing to your worktree — edits may conflict" so the user knows to wait or use `/bgsub:join` before editing.
2. **Auto-suggest isolation on dispatch**: When the model dispatches a write-capable child from an `afk -w` session without `isolation: "worktree"`, inject a hint or warning in the tool result.
3. **File-level advisory locks**: Heavier solution — track which files the background child has touched and warn (or block) if the new turn targets the same paths.

## References

- `src/agent/tools/subagent/foreground-promotion.ts:198-231` — promotion path
- `src/agent/tools/subagent-executor.ts:764-798` — isolation worktree creation (the armored path)
- `src/cli/commands/interactive.ts:569-603` — session cleanup sequence

griffinwork40:
## Summary

Fixes #1513 — warn on Ctrl+B promotion of a write-capable, non-isolated child.

When a user presses Ctrl+B to background a foreground subagent that **shares the parent's worktree** (dispatched without `isolation: "worktree"`), the parent gets their prompt back immediately while the child continues writing to the same tree. This creates a concurrent-write hazard with no coordination — last write wins silently.

## What's in this PR

The implementation was already wired across three files:

| File | What it does |
|------|-------------|
| `foreground-promotion.ts:241` | Sets `sharesWorktree: !isoTd` on the `PromotedSubagentInfo` returned from `resolveJob` |
| `queued-flush.ts:42,48` | Propagates `sharesWorktree?` through the `QueuedFlushControl` return type |
| `turn-handler.ts:143` | Emits `palette.warning('⚠ Background child is writing to your worktree — edits may conflict until it finishes')` when `jobs.some(j => j.sharesWorktree)` |

**This PR adds the missing test coverage** for that wired mitigation:

### New tests in `turn-handler.test.ts`

1. **Non-isolated promotion shows warning** — when `promoteActiveForeground` returns `{ sharesWorktree: true }`, the warning text `'Background child'` and `'worktree'` appear in `completionWriter` output.

2. **Isolated promotion suppresses warning** — when `promoteActiveForeground` returns `{ sharesWorktree: false }`, the normal `'backgrounded as bg-iso'` confirmation line is written but no warning appears.

## Verification

```
pnpm lint            # TypeScript clean (tsc --noEmit)
pnpm exec vitest run src/cli/commands/interactive/turn-handler.test.ts
# → 80 passed (78 existing + 2 new)
pnpm audit:filesize:check
# → .test.ts files are excluded from the ceiling; no new production file touched
```

## Notes

- `palette.warning` used throughout (never raw `chalk.<color>`) — consistent with palette constraints
- No force push
- Pre-existing `GREW` violations in `background-registry.ts`, `schemas.ts`, `interactive.ts`, `template-engine.ts` are unrelated to this PR

### DIFF
--- a/src/cli/commands/interactive/turn-handler.test.ts
+++ b/src/cli/commands/interactive/turn-handler.test.ts
@@ -1321,6 +1321,85 @@ describe('runTurn — borrowed-compositor regression (PR 424 / Stage 3e)', () =>
     expect(writerCalls.some((l) => l.includes('backgrounded'))).toBe(false);
   });
 
+  // Worktree-conflict warning (#1513): non-isolated children share the parent's
+  // worktree and can cause concurrent write conflicts when promoted to background.
+  it('Ctrl+B emits a worktree-conflict warning when the promoted child shares the parent worktree', async () => {
+    const writerCalls: string[] = [];
+    const completionWriter = { fn: (line: string) => { writerCalls.push(line); } };
+
+    // sharesWorktree: true — no isolation: "worktree" was set on the dispatch.
+    const promoteActiveForeground = vi.fn().mockResolvedValue([
+      { jobId: 'bg-warn', label: 'non-isolated job', sharesWorktree: true },
+    ]);
+    const subagentControl = {
+      hasPromotableForeground: () => true,
+      promoteActiveForeground,
+      hasActiveForeground: () => false,
+      cancelActiveForeground: vi.fn().mockResolvedValue(0),
+    };
+
+    const { h } = makeHandles();
+    const handles: TurnHandles = {
+      ...h,
+      subagentControl,
+      setBackgroundHandler: (handler) => { handler?.(); },
+    };
+
+    await runTurn(
+      { text: 'q', attachments: [] },
+      streamFrom([{ type: 'done', metadata: { durationMs: 1 } }]),
+      makeStats(),
+      handles,
+      'summary',
+      completionWriter,
+    );
+    await new Promise((r) => setImmediate(r));
+
+    expect(promoteActiveForeground).toHaveBeenCalledTimes(1);
+    // The worktree-conflict warning must appear.
+    expect(writerCalls.some((l) => l.includes('Background child'))).toBe(true);
+    expect(writerCalls.some((l) => l.includes('worktree'))).toBe(true);
+  });
+
+  it('Ctrl+B does NOT emit a worktree-conflict warning when the promoted child is isolated', async () => {
+    const writerCalls: string[] = [];
+    const completionWriter = { fn: (line: string) => { writerCalls.push(line); } };
+
+    // sharesWorktree: false — isolation: "worktree" was set; child has its own tree.
+    const promoteActiveForeground = vi.fn().mockResolvedValue([
+      { jobId: 'bg-iso', label: 'isolated job', sharesWorktree: false },
+    ]);
+    const subagentControl = {
+      hasPromotableForeground: () => true,
+      promoteActiveForeground,
+      hasActiveForeground: () => false,
+      cancelActiveForeground: vi.fn().mockResolvedValue(0),
+    };
+
+    const { h } = makeHandles();
+    const handles: TurnHandles = {
+      ...h,
+      subagentControl,
+      setBackgroundHandler: (handler) => { handler?.(); },
+    };
+
+    await runTurn(
+      { text: 'q', attachments: [] },
+      streamFrom([{ type: 'done', metadata: { durationMs: 1 } }]),
+      makeStats(),
+      handles,
+      'summary',
+      completionWriter,
+    );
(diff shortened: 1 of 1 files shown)


griffinwork40 (commented):

---

## Prior Automated Rounds

No prior automated round — this is round 1. No prior findings to carry forward.

---

0 critical, 0 high → MERGE (with follow-ups); 1 medium advisory (non-blocking), 1 low, 1 nit.

The two new tests correctly exercise both branches of the worktree-conflict warning: `sharesWorktree: true` → warning emitted, `sharesWorktree: false` → warning absent. Test assertions are anchored to substrings that appear verbatim in production code at `turn-handler.ts:141,143` (verified at ref `0b727ff`). CI is green on ubuntu-latest and macos-latest. No production code changed.

---

## Non-Blocking Follow-Ups

### 1 · medium · non-blocking · test-coverage · `src/cli/commands/interactive/turn-handler.test.ts:1328–1376`

Both tests use single-job promotion batches. The production warning branch at `turn-handler.ts:143` uses `jobs.some((j) => j.sharesWorktree)`, meaning a mixed batch `[{sharesWorktree: true}, {sharesWorktree: false}]` would still trigger the warning. No test covers this mixed-batch path.

```ts
// turn-handler.ts:143 — fires on ANY sharing job in the batch
if (jobs.some((j) => j.sharesWorktree)) write(palette.warning('⚠ Background child…'));
```

**Suggestion (follow-up):** Add a third test returning `[{sharesWorktree: false}, {sharesWorktree: true}]` from `promoteActiveForeground` and asserting the warning still fires. Low risk — `Array.some()` semantics are standard JS.

---

### 2 · low · advisory · correctness · `src/cli/commands/interactive/turn-handler.test.ts:1348–1355`

The `setBackgroundHandler: (handler) => { handler?.(); }` pattern fires the Ctrl+B handler synchronously during `runTurn` setup, before the `for await` loop begins iterating the stream. This works because the stream is a synchronous generator and `await runTurn(...)` absorbs the entire stream before the `setImmediate` flush drains the `.then()` microtask from `promoteWithQueuedFlush`. The same timing trick is used in the adjacent existing tests (lines 1267–1318). No defect, but the timing contract is implicit.

**Suggestion:** A one-line comment on the `setImmediate` flush (`// drains promoteWithQueuedFlush's .then() microtask`) would document the contract for future readers. No urgency.

---

### 3 · nit · advisory · `src/cli/commands/interactive/turn-handler.test.ts:1340`

`cancelActiveForeground: vi.fn().mockResolvedValue(0)` is included in the `subagentControl` mock but is never called by `handleBackgroundKey`. The field satisfies the `SubagentControl` interface shape and is harmless — consistent with existing test patterns.

---

## What Was Not Checked

- Citations verified against branch HEAD `0b727ff3c8de7bc97f275c9330bb8e58327cf3cc`.
- Stated intent: PR #1535 title+body — spec-compliance assessed. Both stated test cases present and correct.
- Absence claims (mixed-batch gap) verified via `git grep` at reviewed ref — zero matches confirmed.
- Did not run tests or verify runtime behaviour.
- Did not review callers of `handleBackgroundKey` or `promoteWithQueuedFlush` outside this file.

Issue #1539, opened 2026-09-07
Run workspace A/B experiment and publish results to close #1196


griffinwork40:
## Context

The final step to close #1196. Once the measurement semantics and A/B runner are correct, run the experiment and publish the results.

### Run the experiment
- At least 5 paired trials (control vs. treatment)
- Alternating or randomized arm order
- All trials pass validation (no swallowed failures, no N/A metrics)
- Results stored under `<AFK state>/experiments/workspace-ab/<run-id>/`

### Persist results via `src/paths.ts`

Use:
```
<AFK state>/experiments/workspace-ab/<run-id>/
  manifest.json
  control/*.json
  treatment/*.json
  comparison.json
  report.md
```

Add the experiment results path through `src/paths.ts`, not a hand-built `~/.afk` join.

### Publish committed summary

Commit a scrubbed summary to:
```
docs/experiments/shared-workspace-ab.md
```

Contents:
- Exact tested commit SHA
- Number of paired trials
- Control and treatment distributions (read dedup ratio, file overlap ratio)
- Median/mean change with confidence interval or range
- Failure and quality comparison
- Limitations and caveats
- Proceed/hold decision

### Close #1196

After the summary is committed:
- Mark measurement, experiment, and query-tool work as complete on #1196
- Point reactive delivery to #1418
- Either open separate TTL/persistence issues or explicitly mark them declined in favor of ephemeral workspace + durable `StateStore`

## Depends on

- Measurement semantics fix (resource fingerprinting)
- A/B runner rewrite (TypeScript, fail-closed, multi-trial)

Issue #1538, opened 2026-09-07
Replace workspace A/B runner with fail-closed TypeScript orchestration


griffinwork40:
## Context

Part of closing #1196. The current A/B runner (`scripts/run-workspace-ab-test.sh`) has several reliability problems:

- Uses `sleep 2` + `ls -t` to identify the session trace, which races with concurrent AFK sessions on this machine.
- Uses `|| true` on both agent runs and measurements, so failures are swallowed.
- Can produce `N/A` metrics and still print "Done".
- Merges stdout and stderr, making JSON parsing fragile.
- Runs a single stochastic pair, too weak to support a product conclusion.
- The fixed prompt launches all investigators concurrently, limiting how much already-published workspace information can reach siblings before they read files.

### Emit trace identity from `afk chat`

Include `sessionId`, `witnessLabel`, and `tracePath` in `afk chat -f json` output. The trace path is already available in `src/cli/commands/chat.ts` through `createDefaultTraceWriter()`.

Files:
- `src/cli/commands/chat.ts`
- Relevant `chat.*.test.ts` JSON-output tests

### Replace shell runner with fail-closed TypeScript

Replace or wrap with:
```
scripts/run-workspace-ab-test.ts
scripts/workspace-ab/run-arm.ts
scripts/workspace-ab/compare.ts
scripts/workspace-ab/run-workspace-ab-test.test.ts
```

The runner should:
1. Build first or fail if the expected executable is absent
2. Assign and record explicit session/trace identity per arm
3. Keep stdout and stderr separate
4. Treat any nonzero run or measurement exit as experiment failure
5. Validate both traces before comparing them
6. Run several paired trials (at least 5)
7. Alternate or randomize arm order to reduce warm-cache and rate-limit bias
8. Record model, provider, prompt hash, git SHA, runtime version, cost, duration, and trial order
9. Use unique output paths per arm
10. Produce a machine-readable manifest and Markdown summary

### Use a task topology that exercises workspace sharing

Use a prompt where sharing can causally affect later reads:
1. Three agents receive intentionally overlapping file surfaces
2. Each must query before reading
3. Each publishes after confirming a module-level fact
4. Each reaches a second checkpoint where it queries again before reading the next overlapping module
5. The final synthesizer is identical across both arms

Keep control and treatment prompts byte-identical. Only `AFK_WORKSPACE_DISABLED` should differ.

Measure output quality alongside read dedup:
- Citation validity
- Required-area coverage
[...]

PR #1554, opened 2026-09-07
feat(trace): resource-level fingerprinting and tested analyzer (#1537)

### LINKED ISSUE #1537: Correct workspace measurement semantics: resource-level fingerprinting and tested analyzer
## Context

Part of closing #1196. The current measurement utility (`scripts/measure-read-dedup.ts`) groups reads by `argsFingerprint`, which hashes the entire tool input including path, offset, and limit. This means two agents reading the same file with different offsets count as different resources, and the reports show opaque hashes rather than per-file results.

### Add resource-level fingerprinting

Add an optional `resourceFingerprint` field to the trace event schema. For `read_file`, compute it from the normalized file path alone. Keep `argsFingerprint` for exact-call deduplication.

Files:
- `src/agent/providers/shared/tool-call-trace.ts`
- `src/agent/trace/types.ts`
- `src/agent/trace/events.ts`
- `src/agent/providers/shared/tool-call-trace.test.ts`

The report should expose two metrics:
- `exactCallDuplicateRatio` (existing, via `argsFingerprint`)
- `crossAgentFileOverlapRatio` (new, via `resourceFingerprint`)

Do not put raw paths into the witness trace. Use a normalized path hash to preserve the trace's current privacy boundary.

### Extract analyzer into tested library code

`measure-read-dedup.ts` is a ~320-line standalone script with no direct tests. Extract the pure logic:

```
scripts/workspace-ab/analyze-read-dedup.ts
scripts/workspace-ab/types.ts
scripts/measure-read-dedup.ts          (thin CLI entry point)
scripts/workspace-ab/analyze-read-dedup.test.ts
```

Test at least:
- Exact duplicate by two different agents
- Same agent repeating a read
- Same file with different offsets
- Different files with identical input sizes
- Legacy traces missing fingerprints
- Root calls versus subagent calls
- Failed or incomplete tool calls
- JSON and human-readable output
- `--file`, `--session`, and `--latest` resolution

### Add validation rules

The report should fail validation if:
- No subagents ran
- Fewer than two subagents performed reads
- Fingerprints are missing from new traces
- The session did not reach a valid closure
- A material portion of children failed or were cancelled

## Relates to

- #1196 (parent)
- Blocks: the A/B runner rewrite (needs correct metrics to compare)

griffinwork40:
Closes #1537. Unblocks #1538 (A/B runner rewrite) and #1539 (run experiment).

## What

Adds resource-level fingerprinting to the witness trace and extracts the read-dedup analyzer into a tested library, addressing both requirements from the issue.

### 1. Resource-level fingerprinting (trace infrastructure)

**Problem:** The existing `argsFingerprint` hashes the entire tool input including `offset` and `limit`, so two agents reading the same file at different offsets count as different resources. The A/B experiment needs a metric that measures *file-level* overlap, not exact-call overlap.

**Fix:** New optional `resourceFingerprint` field on `tool_call.started` events. For `read_file`, it hashes only the normalized file path (trimmed, consecutive slashes collapsed, trailing slash stripped). Also computed for `list_directory` and `grep`. Absent for non-resource tools and pre-upgrade traces.

Privacy: contains a SHA-256 hash, never a raw path.

Files:
- `src/agent/trace/types.ts` -- `ToolCallStartedPayload.resourceFingerprint?: string`
- `src/agent/trace/events.ts` -- Zod schema mirror
- `src/agent/providers/shared/tool-call-trace.ts` -- `computeResourceFingerprint()` + wired into builder
- `src/agent/providers/shared/tool-call-trace.test.ts` -- 7 new tests

### 2. Extracted tested analyzer

**Problem:** `scripts/measure-read-dedup.ts` was a ~370-line untested script mixing CLI, I/O, and analysis logic.

**Fix:** Extracted pure analysis functions into `scripts/workspace-ab/analyze-read-dedup.ts` with shared types in `scripts/workspace-ab/types.ts`. The CLI entry point is now a thin I/O wrapper.

New capabilities:
- `crossAgentFileOverlapRatio` metric (groups by `resourceFingerprint`)
- `validate()` function with 5 experiment-suitability rules:
  - no-subagents, too-few-reading-agents, missing-fingerprints, no-closure, high-child-failure-rate

Files:
- `scripts/workspace-ab/types.ts` -- shared interfaces
- `scripts/workspace-ab/analyze-read-dedup.ts` -- pure analysis + validation
- `tests/workspace-ab-analyze.test.ts` -- 16 tests (9 analyze + 7 validate)
- `scripts/measure-read-dedup.ts` -- thinned CLI entry point

### Gates

All passing:
- `pnpm lint` (tsc --noEmit strict)
- `pnpm audit:filesize:check` (baseline updated: events.ts 426 -> 427, worktree-sweep.ts 512 -> 500)
- `pnpm audit:sdk:check`
- `pnpm fix:pins:check`
- 47 tests passing (31 tool-call-trace + 16 workspace-ab-analyze)

### DIFF
--- a/.filesize-baseline.json
+++ b/.filesize-baseline.json
@@ -1,47 +1,173 @@
 {
   "limit": 350,
   "entries": {
-    "scripts/audit-sdk-dependency.ts": { "loc": 450, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/background-registry.ts": { "loc": 413, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/daemon/scheduler.ts": { "loc": 380, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/memory/memory-store.ts": { "loc": 687, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/memory/memory-tools.ts": { "loc": 437, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/providers/openai-compatible/index.ts": { "loc": 423, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/providers/openai-compatible/query.ts": { "loc": 803, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/session/agent-session.ts": { "loc": 840, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/subagent.ts": { "loc": 351, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/tools/compose-executor.ts": { "loc": 561, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/tools/dispatcher.ts": { "loc": 623, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/tools/hooks/path-approval-hook.ts": { "loc": 386, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/tools/readonly-bash.ts": { "loc": 415, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/tools/schemas.ts": { "loc": 1392, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/tools/subagent-executor.ts": { "loc": 482, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/trace/events.ts": { "loc": 426, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/trace/receipt.ts": { "loc": 412, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/agent/worktree/worktree-sweep.ts": { "loc": 500, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/cli/_lib/testing/virtual-screen.ts": { "loc": 422, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/cli/commands/chat.ts": { "loc": 584, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/cli/commands/daemon.ts": { "loc": 399, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/cli/commands/farm.ts": { "loc": 436, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/cli/commands/interactive.ts": { "loc": 620, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/cli/commands/interactive/loop-iteration.ts": { "loc": 432, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/cli/commands/interactive/tool-lane.ts": { "loc": 465, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/cli/commands/interactive/turn-handler.ts": { "loc": 371, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/cli/commands/interactive/worktree.ts": { "loc": 478, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/cli/commands/trace.ts": { "loc": 573, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/cli/elicitation/agent-question.ts": { "loc": 399, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/cli/slash/commands/info.ts": { "loc": 443, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/cli/terminal-compositor.input-dispatch.ts": { "loc": 512, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/cli/terminal-compositor.ts": { "loc": 354, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/config/env.ts": { "loc": 1754, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/config/import-sources.ts": { "loc": 373, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/improve/eval-run/contracts.ts": { "loc": 500, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/improve/eval-run/runner.ts": { "loc": 372, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/improve/propose/template-engine.ts": { "loc": 461, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/skills/audit-fit/index.ts": { "loc": 424, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/telegram/bot.ts": { "loc": 372, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/telegram/handlers/farm-callbacks.ts": { "loc": 392, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/telegram/handlers/message.ts": { "loc": 533, "reason": "legacy: predates the ceiling gate; pending concern extraction" },
-    "src/telegram/session-manager.ts": { "loc": 443, "reason": "legacy: predates the ceiling gate; pending concern extraction" }
+    "scripts/audit-sdk-dependency.ts": {
+      "loc": 450,
+      "reason": "legacy: predates the ceiling gate; pending concern extraction"
+    },
+    "src/agent/background-registry.ts": {
+      "loc": 413,
+      "reason": "legacy: predates the ceiling gate; pending concern extraction"
+    },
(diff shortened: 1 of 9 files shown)


griffinwork40 (commented):

0 critical, 0 high → MERGE (with follow-ups); 3 medium advisory, 7 low/nit advisory.

Reviewed ref: `f8fbbe30` · Citations verified inline against branch HEAD `f8fbbe3058c3718a8a01758dbee4a6fb6dde73b3`.

---

### Medium (follow-up recommended)

**1 · correctness · `scripts/workspace-ab/analyze-read-dedup.ts:190-199`**
`validate()` double-fires for `distinctAgents === 0`: both `no-subagents` (< 1) and `too-few-reading-agents` (< 2) emit simultaneously. The two rules are a strict subset — callers iterating `failures` see two entries for one underlying fact. Chain as `if/else if` or subsume into one rule.

**2 · correctness · `scripts/workspace-ab/analyze-read-dedup.ts:206-207`**
`missing-fingerprints` rule guards on `totalCalls > 0 && skippedNoFingerprint > 0`. When a trace has *only* fingerprint-less events (`totalCalls === 0`, `skippedNoFingerprint > 0`), the rule is skipped and the trace's unusability is hidden behind the unrelated `no-subagents` failure. Drop the `totalCalls > 0` precondition.

**3 · spec-compliance · `src/agent/providers/shared/tool-call-trace.ts:124-136`**
`read_file` and `list_directory` share a single `case` body using `obj['file_path'] ?? obj['path']`. For `list_directory` the schema key is `path` (confirmed in `schemas.ts` and `path-approval-hook.ts`), so the `file_path` probe is dead code. Splitting into two explicit cases eliminates the cross-key fallback and matches the per-tool extraction pattern used elsewhere.

### Low

**4 · spec-compliance · `src/agent/providers/shared/tool-call-trace.ts:110`**
JSDoc says `"For read_file / glob / list_directory"` but the switch handles `read_file`, `list_directory`, `grep` — `glob` has no case. Either add `case 'glob'` or fix the JSDoc.

**5 · api-compat · `scripts/measure-read-dedup.ts:209,213`**
CLI calls `validate(report)` with no `opts`, so the `no-closure` and `high-child-failure-rate` rules are permanently unreachable from the CLI. Document this intentional gap or thread the opts from the trace.

**6 · security · `src/agent/providers/shared/tool-call-trace.ts:131`**
`resourceFingerprint` hashes a low-entropy normalized path with no salt/HMAC. The trace is local and the incremental exposure is bounded (attacker already has FS access), but if the trace is ever transmitted externally this becomes material. A comment documenting the constraint is sufficient for now.

**7 · security · `src/agent/providers/shared/tool-call-trace.ts:59-63`**
`computeResourceFingerprint` bypasses `redactSensitiveFields`. Correct today (file paths are not secrets), but the two hashing pipelines are divergent. A comment noting the invariant would prevent a future contributor from adding a credential-carrying tool without a redaction gate.

**8 · test-coverage · `tests/workspace-ab-analyze.test.ts:178-182`**
The `'fails when no subagents ran'` test uses `.some()` but does not assert failure count — misses the double-emission in finding 1. Adding `.toHaveLength(1)` would catch it.

**9 · test-coverage · `src/agent/providers/shared/tool-call-trace.test.ts`**
No test covers a `glob` tool call through `buildToolCallStartedPayload` to verify `resourceFingerprint` is absent. Given the JSDoc mismatch (finding 4), this is a useful regression guard.

**10 · perf-observability · `scripts/workspace-ab/analyze-read-dedup.ts:99-103`**
Sort comparator uses `.find()` inside the callback, yielding O(A·N·log A) per group. Pre-computing `firstSeq` per agent as a Map avoids the inner scan. Bounded impact for current trace sizes.

---

## What was not checked
- Stated intent: PR #1554 title+body — spec-compliance assessed.
- Citations verified inline against branch HEAD `f8fbbe30`.
- Did not run tests or measure performance impact.
- Did not review Telegram surface, CLI rendering, or daemon paths.

griffinwork40 (commented):

---

## Prior Rounds
No prior automated review for this PR — first round.

---

## Merge Decision

**MERGE (with follow-ups)**

Severity arithmetic: **0 critical, 0 high → MERGE (with follow-ups)**; 3 medium advisory (1 correctness, 1 spec-compliance, 1 metric-design); 4 low non-blocking.

No finding rises to `critical` or `high` severity. All issues are bounded defects that don't produce wrong output or unsafe state under reachable production conditions. The core fingerprinting infrastructure is sound, CI is green, and tests are well-formed. Follow-up issues recommended for the mediums below.

---

## Advisory Findings (non-blocking — follow-up issues suggested)

### MEDIUM — M-1: `validate()` emits a misleading second failure when `distinctAgents === 0`
**Dimension:** correctness · **File:** `scripts/workspace-ab/analyze-read-dedup.ts` · **ref:** f8fbbe3

When `distinctAgents === 0`, both guards fire sequentially (not `else if`):
```typescript
if (report.distinctAgents < 1) {
  failures.push({ rule: 'no-subagents', ... });          // fires
}
if (report.distinctAgents < 2) {                          // also fires — 0 < 2
  failures.push({ rule: 'too-few-reading-agents', ... });  // misleading: "too few" ≠ "none"
}
```
A caller iterating `failures` sees two distinct-looking problems for one root cause. The `too-few-reading-agents` label is misleading when there are zero agents (no agents is not the same as "too few").

**Suggestion:** Change `if (report.distinctAgents < 2)` to `else if (report.distinctAgents < 2)`. Verified against the source at ref; the test suite confirms `no-subagents` fires for 0 agents, `too-few-reading-agents` fires for 1.

---

### MEDIUM — M-2: Windows backslash paths not normalized in `computeResourceFingerprint`
**Dimension:** spec-compliance · **File:** `src/agent/providers/shared/tool-call-trace.ts` · **ref:** f8fbbe3

The stated intent is "SHA-256 of **normalized** file path." The normalization is:
```typescript
const normalized = raw.trim().replace(/\/+/g, '/').replace(/\/$/, '');
```
This correctly handles double-slashes and trailing slashes, but does **not** normalize Windows-style backslash separators. On a Windows host, `C:\\foo\\bar.ts` and `C:/foo/bar.ts` (same physical file) produce different fingerprints, breaking cross-agent dedup for mixed-separator paths. The codebase targets Windows (CI matrix includes Windows-skipped per baseline but explicit compat work exists).

**Suggestion:** Add `.replace(/\\/g, '/')` before the slash-dedup step:
```typescript
const normalized = raw.trim().replace(/\\/g, '/').replace(/\/+/g, '/').replace(/\/$/, '');
```
Applies to all three branches (`read_file`, `list_directory`, `grep`). Boundary case: a test for trailing slash and double-slash normalization exists, but no test for backslash normalization.

---

### MEDIUM — M-3: `grep` fingerprint design: pattern/include exclusion is documented but creates measurable metric inflation
**Dimension:** metric-design · **File:** `src/agent/providers/shared/tool-call-trace.ts` · **ref:** f8fbbe3

The `grep` case hashes only `path`, deliberately excluding `pattern` and `include`. The inline comment documents this: *"grep's resource is `path` + `pattern`, but `include` / other args change what is read. Keep it narrow: only hash when `path` is present."*

The design choice is documented and intentional. However, `include` narrows the files actually read — two agents grepping the same directory with `include: '*.ts'` vs `include: '*.py'` are counted as a resource-level overlap even though they read entirely different files. For the `crossAgentFileOverlapRatio` metric to be accurate for directory-scoped greps, the `include` argument is part of the resource identity.

This is a **design question, not a clear bug**, since the intent is file-path-level (not content-level) overlap. Recommend a follow-up issue to decide: include `include` in the fingerprint key, or update the metric docs to clarify "overlap" means "same path target touched, not same content read."

---

## Low / Nit Findings (non-blocking, informational)

- **LOW — L-1:** `computeFileOverlapRatio` has an unreachable false branch: `return resourceCalls.length > 0 ? duplicates / resourceCalls.length : 0` — the `0` branch is dead after the early `if (resourceCalls.length === 0) return null`. Not a data-loss bug; simplify to `return duplicates / resourceCalls.length`. (`scripts/workspace-ab/analyze-read-dedup.ts`)

- **LOW — L-2:** SHA-256 of a file path is not a privacy primitive (paths are low-entropy, enumerable). The PR already states this in the JSDoc ("Privacy: contains a hash, never a raw path"). The JSDoc accurately conveys deduplication intent; adding "this is a correlation key, not an anonymization layer" to `computeResourceFingerprint`'s function comment would prevent misuse by future callers. (`src/agent/providers/shared/tool-call-trace.ts`)

- **LOW — L-3:** Case-sensitive hashing on case-insensitive filesystems (macOS HFS+, Windows NTFS): `/Foo/Bar.ts` and `/foo/bar.ts` produce different fingerprints for the same file. In practice agents derive paths from directory listings and produce consistent casing. Worth documenting the limitation in the JSDoc. (`src/agent/providers/shared/tool-call-trace.ts`)

- **LOW — L-4:** `write_file` and `edit_file` are not covered by `computeResourceFingerprint` (they fall to `default: undefined`). This means the metric silently under-counts write-read conflicts. Not blocking; document that the metric is read-operations-only. (`src/agent/providers/shared/tool-call-trace.ts`)

---

## What Was Not Checked
- Citations verified inline against branch HEAD `f8fbbe3058c3718a8a01758dbee4a6fb6dde73b3` via GitHub API (local git checkout is behind; all file reads went through `gh api`).
- Did not run tests locally (CI is passing green per all check runs on this PR).
- Did not review Telegram surface or daemon integration paths (none modified by this PR).
- Stated intent: PR #1554 title + body — spec-compliance assessed.
- T-1 (no-closure test), T-2 (high-child-failure-rate test), T-3 (non-resource-tool absence test): all initially flagged as missing, then verified present at lines 207–218 and 216–228 of the test file — dropped after Wave 1.5 verification.
- S-1 security finding about `grep` pattern exclusion: after verifying the inline design comment, reclassified from blocking to advisory.

PR #1558, opened 2026-09-07
feat(config): flip 5 env var defaults from opt-in to opt-out

griffinwork40:
## Summary

Flips 5 env var defaults from opt-in to opt-out so new installs get the full feature set out of the box.

| Variable | Before | After |
|---|---|---|
| `AFK_SUGGEST_ENABLED` | off (unset) | **on** by default |
| `AFK_SUGGEST_PROMPT` | off (unset) | **on** by default |
| `AFK_OPENAI_CHATGPT_OAUTH` | off (unset) | **on** by default |
| `AFK_MEMORY_EVIDENCE_GATE` | off (unset) | **on** by default |
| `AFK_CAPTURE_SUBAGENT_OUTPUT` | off (unset) | **on** by default |

All five can be disabled with `0`/`false`/`no`/`off`. Existing users who had them set to `1` see no behavioral change.

## What changed

**Runtime consumer sites** (7 source files) inverted from opt-in (`=== '1'`) to opt-out (`!== '0'`):

- `src/cli/commands/interactive/surface-setup.suggest-config.ts` -- suggest enabled + prompt suggest
- `src/agent/providers/openai-compatible/auth.ts` -- ChatGPT OAuth
- `src/agent/memory/memory-evidence.ts` -- evidence gate
- `src/agent/session/subagent-output-capture.ts` -- capture gate function
- `src/agent/providers/anthropic-direct/loop/tool-dispatch.ts` -- capture inline check
- `src/agent/providers/openai-compatible/query/dispatch-append.ts` -- capture inline check

**Registry metadata** (`src/config/env.ts`): added `default: '1'` and updated descriptions from "Off by default" to "On by default. Set to 0 to disable."

**Tests** (5 test files): updated to assert new default-on semantics and explicit-disable behavior.

**Generated docs**: `docs/env-registry.{json,md}` regenerated via `pnpm scan:env`. `.filesize-baseline.json` updated for env.ts (+7 LOC from added `default`/`example` fields).

## Verification

- `pnpm lint` -- clean
- `pnpm test` -- 19,475 passed, 26 skipped, 0 failed
- `pnpm scan:env:check` -- in sync
- `pnpm audit:env:check` -- 0 violations
- `pnpm audit:filesize:check` -- green


griffinwork40:
Confirmed - the inline comment at line 192 still says "Gated behind an explicit opt-in flag (off by default)" which is now stale. This is a separate doc-staleness finding from the module docstring at lines 30-31.

Now I have enough for **Wave 2 -- Synthesis**. Let me compile the final review.

---

# PR #1558 Review: `feat(config): flip 5 env var defaults from opt-in to opt-out`

**Branch:** `feat/flip-env-defaults` → `main` | **Ref:** `263718c4` | **+98/-78 across 17 files**

**Stated intent:** Flip 5 env var defaults from opt-in to opt-out so new installs get the full feature set out of the box. The 5 vars: `AFK_SUGGEST_ENABLED`, `AFK_SUGGEST_PROMPT`, `AFK_OPENAI_CHATGPT_OAUTH`, `AFK_MEMORY_EVIDENCE_GATE`, `AFK_CAPTURE_SUBAGENT_OUTPUT`. All five can be disabled with `0`/`false`/`no`/`off`.

---

## Findings

### 1. `high` · blocking:true · security · `AFK_OPENAI_CHATGPT_OAUTH` silently activates OAuth token forwarding

**`src/agent/providers/openai-compatible/auth.ts:274-280`** · ref:263718c4 · file-state · confidence:high

`chatGptOAuthEnabled` now returns `true` when the env var is unset. Any existing user whose `~/.codex/auth.json` contains a ChatGPT-subscription token will have that OAuth access_token forwarded to `chatgpt.com/backend-api` on first run of the new version -- without any explicit consent act. The module's own docstring at line 30-31 still says *"By default this resolver returns 'no-usable-auth-codex-oauth' (the token is present but unused), preserving the historical safe behavior"* -- the implementation now contradicts the stated safety contract.

```typescript
// auth.ts:274-280 (at ref)
function chatGptOAuthEnabled(readEnv: ...): boolean {
  const v = readEnv('AFK_OPENAI_CHATGPT_OAUTH');
  if (!v) return true;  // was: return false
  ...
}
```

**Suggestion:** Revert the default to `false`. If the goal is easier onboarding, detect the ChatGPT-mode auth.json interactively and prompt the user to confirm, then write `AFK_OPENAI_CHATGPT_OAUTH=1` to `afk.env`.

---

### 2. `high` · blocking:true · security/perf-observability · `AFK_CAPTURE_SUBAGENT_OUTPUT` default-on expands credential-at-rest surface

**`src/agent/session/subagent-output-capture.ts:74`** + **`tool-dispatch.ts:83`** + **`dispatch-append.ts:120`** · ref:263718c4 · file-state · confidence:high

This flag was previously opt-in with an explicit documented rationale (AFK.md line 60: *"Off by default, deliberately: ... redaction is regex-based, so connection strings, PEM blocks, and PII are not caught."*). The env.ts description at line 1494 itself acknowledges: *"best-effort regex redaction only -- connection strings, PEM blocks, and PII are not caught."* The sibling flag `AFK_CAPTURE_SUBAGENT_PROMPTS` remains opt-in for the same reasons. Every new install now silently writes a markdown transcript per subagent to `state/witness/<label>/outputs/`. The witness-sweep does GC old directories (so the "no GC" framing from the agents was overstated), but redaction gaps remain the primary concern.

```typescript
// subagent-output-capture.ts:74
if (env.AFK_CAPTURE_SUBAGENT_OUTPUT === '0') return false;  // was: !== '1'
```

**Suggestion:** Restore opt-in, or first close the redaction gap (structured scrubber for PEM/connection strings) before flipping the default.

---

### 3. `medium` · blocking:true · correctness · Garbage-string input activates ChatGPT OAuth credential routing

**`src/agent/providers/openai-compatible/auth.ts:280-283`** · ref:263718c4 · file-state · confidence:high

The old code accepted only an explicit allowlist (`1|true|yes|on`) to enable. The new code treats *any* non-deny-list string as `true`. `AFK_OPENAI_CHATGPT_OAUTH=garbage` or `=maybe` now silently activates the ChatGPT OAuth path and routes traffic to `chatgpt.com/backend-api`. For the suggest-config vars, an unrecognized string activating ghost-text is a UI nuisance. For OAuth, it switches the auth backend and forwards a credential -- a qualitatively different risk.

```typescript
return !(n === '0' || n === 'false' || n === 'no' || n === 'off');
// "garbage" is not in the deny list → returns true → OAuth activated
```

**Suggestion:** For the `!v` (unset) case, return `true` (the new default). For a non-empty, non-recognized value, fall back to `false` rather than `true`. This preserves the on-by-default intent while preventing typos from silently activating credential routing.

---

### 4. `medium` · blocking:true · spec-compliance · `AFK_MEMORY_EVIDENCE_GATE` and `AFK_CAPTURE_SUBAGENT_OUTPUT` do not honor `false`/`no`/`off`

**`src/agent/memory/memory-evidence.ts:54`** + **`src/agent/session/subagent-output-capture.ts:74`** · ref:263718c4 · file-state · confidence:high

The PR's stated intent says *"All five can be disabled with 0/false/no/off."* The suggest-config and auth.ts implementations honor all four disable values via `!/^(0|false|no|off)$/i`. But `evidenceGateEnabled()` checks only `!== '0'`, and `shouldCaptureSubagentOutput()` checks only `=== '0'`. A user who sets `AFK_MEMORY_EVIDENCE_GATE=false` or `AFK_CAPTURE_SUBAGENT_OUTPUT=off` will find the feature remains enabled, violating the PR's own contract.

```typescript
// memory-evidence.ts:54 — only checks '0'
return env.AFK_MEMORY_EVIDENCE_GATE !== '0';

// subagent-output-capture.ts:74 — only checks '0'
if (env.AFK_CAPTURE_SUBAGENT_OUTPUT === '0') return false;
```

The same `!== '0'` pattern is used at the two `toolInputCapture` sites in `tool-dispatch.ts:83` and `dispatch-append.ts:120`.

**Suggestion:** Apply the same four-value deny-list pattern (`0|false|no|off`) used in `auth.ts` and `surface-setup.suggest-config.ts` to all five sites. Extract a shared `isExplicitlyDisabled(val)` helper in `src/config/env.ts`.

---

### 5. `medium` · blocking:true · api-compat · `no-usable-auth-codex-oauth` diagnostic is stale

**`src/agent/providers/openai-compatible/auth.ts:365-369`** · ref:263718c4 · file-state · confidence:high

The diagnostic for `no-usable-auth-codex-oauth` says *"To use your ChatGPT subscription, set AFK_OPENAI_CHATGPT_OAUTH=1"*. With the default now on, the only way to reach this path when `chatGptOAuthEnabled` returns false is with an explicit `=0`. But this return is also reached when the flag is on but `parsed.accessToken` is absent (line 196). The two sub-cases need different user instructions: one is "you disabled it", the other is "your token file is incomplete."

Additionally, the inline comment at line 192 still reads *"Gated behind an explicit opt-in flag (off by default)"* and the module docstring at line 30-31 describes the old safe-default behavior. Both are now stale.

**Suggestion:** Split the return into two sources with distinct diagnostics, and update all three stale comments.

---

### 6. `medium` · blocking:true · spec-compliance · `AFK_SUBAGENT_LOG` description states false consistency

**`src/config/env.ts:1525`** · ref:263718c4 · file-state · confidence:high

`AFK_SUBAGENT_LOG`'s description says *"OFF by default (raw tool arguments are written without redaction -- consistent with AFK_CAPTURE_SUBAGENT_PROMPTS / AFK_CAPTURE_SUBAGENT_OUTPUT)"*. `AFK_CAPTURE_SUBAGENT_OUTPUT` is no longer off by default, breaking the stated consistency. The PR regenerated `docs/env-registry.{json,md}` but this cross-reference in the source registry was not updated.

**Suggestion:** Remove `/ AFK_CAPTURE_SUBAGENT_OUTPUT` from the parenthetical, or update the description to reflect the asymmetry.

---

### 7. `low` · blocking:false · spec-compliance · AFK.md contradicts env registry on capture default

**`AFK.md:60`** · ref:263718c4 · file-state · confidence:high

AFK.md describes `AFK_CAPTURE_SUBAGENT_OUTPUT` with *"same off-by-default rationale as the prompts flag above"* and documents it as opt-in (`AFK_CAPTURE_SUBAGENT_OUTPUT=1`). The env registry now says on-by-default. Two authoritative docs now contradict.

**Suggestion:** Update AFK.md to match.

---

### 8. `low` · blocking:false · test-coverage · No test for garbage-string path on `chatGptOAuthEnabled`

**`src/agent/providers/openai-compatible/auth.test.ts`** · ref:263718c4 · diff-context · confidence:high

The behavioral change for unrecognized string values (`garbage` now returns `true` instead of `false`) has no test coverage. Under the old allowlist, `garbage` was safe (returned false). Under the new deny-list, it activates OAuth.

**Suggestion:** Add a test asserting the chosen behavior for an unrecognized non-empty input.

---

### 9. `nit` · blocking:false · correctness · `.gitignore` `drafts/` addition is unrelated scope

**`.gitignore`** · ref:263718c4 · diff-context · confidence:high

Adding `drafts/` to `.gitignore` is unrelated to the env-var default flip. Benign, but increases the commit's atomic surface unnecessarily.

---

## Verdict

**Decision: DO NOT MERGE** -- 2 high, 4 medium blocking (1 correctness, 1 api-compat, 2 spec-compliance); 2 low, 1 nit.

The three suggest-related flips (`AFK_SUGGEST_ENABLED`, `AFK_SUGGEST_PROMPT`) and the evidence-gate flip (`AFK_MEMORY_EVIDENCE_GATE`) are straightforward and well-tested -- they just need the four-value disable pattern applied consistently (Finding 4). The two remaining flips carry qualitatively different risk:

- **`AFK_OPENAI_CHATGPT_OAUTH`**: silently forwarding a credential without user consent. The module's own safety documentation characterizes the previous default as "the historical safe behavior."
- **`AFK_CAPTURE_SUBAGENT_OUTPUT`**: writing tool arguments (which routinely contain secrets) to disk with known redaction gaps. The project's own AFK.md says the flag is off "deliberately" because of those gaps.

Both should either remain opt-in or have their prerequisites (interactive consent / redaction improvements) addressed before the default flip.

---

## What was not checked

- Citations verified inline against branch HEAD `263718c4`.
- Stated intent: PR #1558 title+body -- spec-compliance assessed.
- Did not run tests, build, or lint (read-only review).
- Did not review Telegram, daemon, or `afk chat` surfaces for behavioral impact of the default flips (they consume the same `env.*` getters, so the logic change propagates, but surface-specific behavior was not traced).
- The `.filesize-baseline.json` ratchet change (`worktree-sweep.ts 512→500`) is unrelated to this PR and was not investigated.

**Done**
- What was done: Full-regime code review of PR #1558 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability dimensions.
- Evidence: Review output above; diff artifact at `/Users/griffinlong/.afk/state/skill-preflight/50c14ec6-e7e1-4026-ae99-bcd558fca483/pr-1558.diff`.
- Deferred: Shadow-verify was warranted for Findings 1-2 (high/security) but skipped because both cite verbatim code + documentation that I verified inline at the reviewed ref; the findings are structural (default-flip mechanics), not inference-based, so independent re-derivation would not add signal.

---

Issue #1557, opened 2026-09-07
fix(telegram): close residual TOCTOU window in persistRouteAndMessageId


griffinwork40:
## Context

PR #1552 fixed the main race in `persistRouteAndMessageId` where a stale `pending` copy overwrote an `answered` record (#1549). The fix re-reads the record from disk before writing and skips the write if the record is no longer `pending`.

## Residual gap

The read-check-write sequence in `persistRouteAndMessageId` (`src/telegram/handoff-answer.ts:395-403`) holds no lock across the gap:

```typescript
const fresh = await readHandoff(record.taskId, handoffsDir);
if (fresh === null || fresh.status !== 'pending') return;
// <-- concurrent updateHandoffAnswer can complete here -->
const updated: HandoffRecord = { ...fresh, route, telegramMessageId: messageId };
await writeHandoff(updated, handoffsDir);
```

`updateHandoffAnswer` uses O_EXCL lock files for CAS (`handoff-store.ts:259,289`), but `writeHandoff` (called by `persistRouteAndMessageId`) does not acquire that lock -- it calls `atomicWriteJson` directly. A concurrent `updateHandoffAnswer` completing between the `readHandoff` and `writeHandoff` calls still overwrites the answered record.

The window is dramatically narrower than the original bug (sub-microsecond gap between two `await` resumptions vs. the full Telegram network RTT), and the function's contract is explicitly best-effort (the catch block at line 404-406 swallows all errors). This is low-priority but worth closing for correctness.

## Proposed fix

Have `persistRouteAndMessageId` acquire the same O_EXCL lock that `updateHandoffAnswer` uses, or add a `persistRouteAndMessageId`-specific helper in `handoff-store.ts` that performs the read-check-write under the lock internally.

griffinwork40:
Already fixed in commit 5b1663bc (PR #1567).

Issue #1549, opened 2026-09-07
fix(telegram): persistRouteAndMessageId can overwrite a concurrent answer


griffinwork40:
## Problem

`persistRouteAndMessageId` does an unlocked full-record overwrite of the HandoffRecord after sending the Telegram message. If the operator answers in the narrow window between `sendMessage` resolving and `writeHandoff` completing, `answerHandoff` transitions the record to `answered` via CAS, then `persistRouteAndMessageId` overwrites it with the stale `pending` copy -- erasing the answer.

**File:** `src/telegram/handoff-answer.ts` lines 356-375

## Expected behavior

The route/messageId update should re-read and merge under the same per-task locking discipline as the answer CAS, rather than overwriting the complete record.

## Suggested approach

Change `persistRouteAndMessageId` to read-then-merge: re-read the record, check it's still `pending`, then write only the `route` and `telegramMessageId` fields. Alternatively, use the CAS mechanism from `handoff-store.ts` so a concurrent answer wins.

## Context

Identified during PR #1533 review (Codex P1 + /review round 2). The race window is narrow (between Telegram API response and a local file write) but the consequence is silent answer loss.

PR #1567, opened 2026-09-08
fix(telegram): close TOCTOU window in persistRouteAndMessageId (#1557)

griffinwork40:
## Problem

`persistRouteAndMessageId` in `src/telegram/handoff-answer.ts` performed a
read-check-write without holding the same O_EXCL lock that `updateHandoffAnswer`
uses in `src/agent/daemon/handoff-store.ts`. A concurrent `updateHandoffAnswer`
completing between the read and write could overwrite the answered record, clobbering
the answer status with a stale pending+route write.

## Fix

Extracted the lock acquisition/release logic from `updateHandoffAnswer` into a new
`withHandoffLock<T>` helper exported from `handoff-store.ts`:

- Acquires `<taskId>.lock` with O_EXCL (same protocol as before)
- Runs a typed callback with the lock held
- Releases the lock in a finally block
- Handles stale lock detection (>30 s TTL) with a single retry, matching the
  existing behavior

Refactored `updateHandoffAnswer` to use `withHandoffLock` internally
(behavior-preserving — same lock protocol, same stale-lock handling, same CAS
semantics).

`persistRouteAndMessageId` now calls `withHandoffLock` before doing its
read-check-write, eliminating the TOCTOU window. Under the lock it re-reads the
record, skips if no longer pending, and writes if still pending. The existing
best-effort error-swallowing contract is preserved (lock-not-acquired returns null,
which the try/catch absorbs just like any other error).

## Tests

Added a `withHandoffLock` test suite to `handoff-store.test.ts` covering:

- Callback runs and return value is propagated when lock is free
- Returns `null` immediately when a fresh lock file blocks acquisition
- Clears a stale lock (>30 s) and runs the callback
- Releases the lock after callback success
- Releases the lock even when the callback throws
- Exactly one concurrent caller wins when two race for the same taskId

All 25 existing `handoff-store` tests continue to pass (2 skipped on non-Linux,
as before). All 18 existing `handoff-answer` tests continue to pass, including
the #1549 TOCTOU regression test that exercises the concurrent-answer scenario.

## Verification

```
pnpm lint        -- exit 0
pnpm test src/agent/daemon/handoff-store.test.ts   -- 25 passed, 2 skipped
pnpm test src/telegram/handoff-answer.test.ts      -- 18 passed
pnpm audit:filesize:check  -- no new violations (pre-existing events.ts baseline violation unrelated)
handoff-store.ts: 175 code lines (ceiling: 350)
handoff-answer.ts: 297 code lines (ceiling: 350)
```

### DIFF
--- a/src/agent/daemon/handoff-store.test.ts
+++ b/src/agent/daemon/handoff-store.test.ts
@@ -8,6 +8,7 @@ import {
   deleteHandoff,
   listPendingHandoffs,
   updateHandoffAnswer,
+  withHandoffLock,
   type HandoffRecord,
 } from './handoff-store.js';
 
@@ -381,6 +382,65 @@ describe('concurrent writes', () => {
     for (let i = 0; i < count; i++) {
       expect(reads[i]?.sessionId).toBe(`sess-batch-${i}`);
     }
+  });
+});
+
+// ---------------------------------------------------------------------------
+// withHandoffLock
+// ---------------------------------------------------------------------------
+
+describe('withHandoffLock', () => {
+  it('runs the callback and returns its value when the lock is free', async () => {
+    const result = await withHandoffLock('q-lock-free', testDir, async () => 'hello');
+    expect(result).toBe('hello');
+  });
+
+  it('returns null immediately when an existing fresh lock file blocks acquisition', async () => {
+    const lockFile = join(testDir, 'q-lock-blocked.lock');
+    // Simulate a live holder by creating the lock file with a fresh mtime.
+    await writeFile(lockFile, '', { mode: 0o600 });
+
+    const result = await withHandoffLock('q-lock-blocked', testDir, async () => 'should-not-run');
+    expect(result).toBeNull();
+  });
+
+  it('clears a stale lock file (>30 s) and runs the callback', async () => {
+    const lockFile = join(testDir, 'q-lock-stale.lock');
+    await writeFile(lockFile, '', { mode: 0o600 });
+    const staleTime = new Date(Date.now() - 60_000);
+    await utimes(lockFile, staleTime, staleTime);
+
+    const result = await withHandoffLock('q-lock-stale', testDir, async () => 'ran-after-stale');
+    expect(result).toBe('ran-after-stale');
+  });
+
+  it('releases the lock after the callback succeeds', async () => {
+    await withHandoffLock('q-lock-release', testDir, async () => 'ok');
+    // Lock file must be gone so a second call can acquire it.
+    const result2 = await withHandoffLock('q-lock-release', testDir, async () => 'second');
+    expect(result2).toBe('second');
+  });
+
+  it('releases the lock even when the callback throws', async () => {
+    await expect(
+      withHandoffLock('q-lock-throws', testDir, async () => { throw new Error('boom'); }),
+    ).rejects.toThrow('boom');
+    // Lock must be released so subsequent callers are not blocked.
+    const result = await withHandoffLock('q-lock-throws', testDir, async () => 'after-throw');
+    expect(result).toBe('after-throw');
+  });
+
+  it('only one concurrent caller succeeds when two race for the same taskId', async () => {
+    const order: string[] = [];
+    const [r1, r2] = await Promise.all([
+      withHandoffLock('q-lock-race', testDir, async () => { order.push('winner'); return 'w'; }),
+      withHandoffLock('q-lock-race', testDir, async () => { order.push('should-not-run'); return 's'; }),
+    ]);
+    // Exactly one call ran.
+    const results = [r1, r2];
+    expect(results.filter(r => r !== null)).toHaveLength(1);
+    expect(results.filter(r => r === null)).toHaveLength(1);
+    expect(order).toEqual(['winner']);
   });
 });
 
--- a/src/agent/daemon/handoff-store.test.ts
+++ b/src/agent/daemon/handoff-store.test.ts
@@ -432,15 +432,20 @@ describe('withHandoffLock', () => {
 
   it('only one concurrent caller succeeds when two race for the same taskId', async () => {
     const order: string[] = [];
+    // Hold the lock long enough for the loser to attempt acquisition while
+    // it is still held. Without the delay, both can succeed sequentially
+    // on fast CI (macOS) where the first callback completes and releases
+    // before the second even tries.
+    const delay = (ms: number) => new Promise<void>(r => setTimeout(r, ms));
     const [r1, r2] = await Promise.all([
-      withHandoffLock('q-lock-race', testDir, async () => { order.push('winner'); return 'w'; }),
-      withHandoffLock('q-lock-race', testDir, async () => { order.push('should-not-run'); return 's'; }),
+      withHandoffLock('q-lock-race', testDir, async () => { order.push('a'); await delay(50); return 'w'; }),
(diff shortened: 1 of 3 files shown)


griffinwork40 (commented):

---

## Code Review — `fix(telegram): close TOCTOU window in persistRouteAndMessageId` (#1567)

**Ref:** `e3ded57a1222f60f5ececd414d2bfee59b00bd88` · **Change type:** hotfix · **181 lines (135 added, 46 removed)**

---

### Verdict

**MERGE (with follow-ups)**

**Decision arithmetic:** 0 critical, 0 high → MERGE (with follow-ups). Two findings (S-1, C-2) carried `blocking:true` inherited from pre-downgrade `medium` severity; Wave 1.5 citation + absence-claim verification downgraded both to `low` (advisory). Sweep-level policy: only `critical` and `high` gate merge. Remaining tally: **0 critical · 0 high · 5 low advisory**.

Stated intent fully met: `withHandoffLock<T>` extracted, `updateHandoffAnswer` refactored onto it, `persistRouteAndMessageId` now holds the O_EXCL lock across its read-check-write — closing the TOCTOU window described in #1557. CAS semantics and stale-lock detection preserved. No breaking API changes.

---

### Findings (all advisory — non-blocking)

#### S-1 · low · advisory · security
**`withHandoffLock` public API lacks `assertSafeJobId` guard**

`withHandoffLock` accepts a raw `taskId` string and constructs a lock-file path from it without calling `assertSafeJobId` (which enforces `[A-Za-z0-9_-]`). All current production call paths pass a pre-validated `taskId` (`updateHandoffAnswer` calls `assertSafeJobId` explicitly before `withHandoffLock`; `persistRouteAndMessageId` receives `record.taskId` from a `HandoffRecord` written by `writeHandoff`, which validates at write time). This is a **future-caller API contract gap**, not a live vulnerability — but a future caller reaching `withHandoffLock` with an attacker-controlled string would bypass the path-traversal check at the lock layer.

**Suggested follow-up issue:** Add `assertSafeJobId(taskId)` as the first statement inside `withHandoffLock` so the helper is safe for any caller.

---

#### C-2 · low · advisory · correctness
**`withHandoffLock` lacks `ensureHandoffsDir` — ENOENT silently swallowed under external interference**

`withHandoffLock` does not call `ensureHandoffsDir` before the O_EXCL open. In production, `writeHandoff` (handoff-wiring.ts:135) always runs before `sendHandoffQuestion` (handoff-wiring.ts:166), so the directory exists when `withHandoffLock` is invoked. An ENOENT requires OS-level directory removal mid-process — not reachable under normal usage. If it did occur, the error would be silently absorbed by `persistRouteAndMessageId`'s best-effort catch block.

**Suggested follow-up issue:** Add `ensureHandoffsDir()` at the top of `withHandoffLock` so the helper is self-contained, matching the contract of every other exported function in the module.

---

#### T-2 · low · advisory · test-coverage
No test covers `withHandoffLock` behavior when `handoffsDir` does not exist. Given C-2's silent-failure path, a test pinning that behavior would prevent unintentional contract changes in future refactors.

---

#### P-2 · low · advisory · perf-observability
When `withHandoffLock` returns `null` (lock contended) inside `persistRouteAndMessageId`, the result is discarded with no log line — route-persistence skips are operationally invisible. A `debug`-level log on the null return path would make contention detectable in production traces.

---

#### A-1 · low · non-blocking · api-compat
No breaking changes. `updateHandoffAnswer` signature fully preserved; `withHandoffLock` is an additive export.

---

### Dropped / Downgraded Findings (Wave 1.5 manifest)

| ID | Action | Reason |
|----|--------|--------|
| SP-2 (JSDoc stub) | **DROPPED** | Fabricated citation — `[same JSDoc as before]` text does not appear in source at reviewed ref |
| T-1 (no integration test for TOCTOU fix) | **DROPPED** | False-absent — `handoff-answer.test.ts:161-205` contains the `#1549 TOCTOU regression tests` covering the new `withHandoffLock` path in `persistRouteAndMessageId` |
| S-1 | medium → **low** (advisory) | All production paths already validate `taskId` before reaching `withHandoffLock`; gap is future-caller contract only |
| C-2 | medium → **low** (advisory) | ENOENT requires OS-level directory deletion mid-process; not reachable under normal usage |
| C-1 (silent-swallow on callback throw) | **DROPPED** | Pre-existing intentional design — best-effort catch existed before this PR and is explicitly documented in the PR body |
| C-3 (double `assertSafeJobId`) | **DROPPED** | Redundant but harmless; `writeHandoff` calling `assertSafeJobId` internally on a pre-validated `taskId` is a no-op |

---

### Epistemic Scope

- Citations verified against: `e3ded57a1222f60f5ececd414d2bfee59b00bd88` (inline Wave 1.5 verification against file content at reviewed ref)
- Stated intent: PR #1567 title + body — spec-compliance assessed
- Did not check: runtime behavior under actual lock contention at scale; `atomicWriteJson` implementation; full `handoff-wiring.ts` for secondary effects

griffinwork40 (commented):

**Decision: MERGE (with follow-ups)** — 0 critical, 0 high; 1 medium advisory, 2 low, 1 nit.

---

### Advisory findings (non-blocking)

**1 · medium · correctness · `src/agent/daemon/handoff-store.ts:~277` · advisory**
`withHandoffLock` does not call `assertSafeJobId(taskId)` before forming the lock path. `updateHandoffAnswer` calls the guard at its own top, but the new `persistRouteAndMessageId` call-site in `handoff-answer.ts:422` does not — and neither does `withHandoffLock` itself.

In practice, `record.taskId` was validated at record-creation time via `writeHandoff → assertSafeJobId`, so this path is safe today. The gap is in the public function's contract: a future caller without upstream validation would bypass the guard.

**Suggestion:** Move `assertSafeJobId(taskId)` into `withHandoffLock` as its first statement so all callers get the guard by default. Alternatively, follow up in a separate PR.

---

**2 · low · correctness · `src/telegram/handoff-answer.ts:~420-435` · advisory**
When `withHandoffLock` returns `null` (lock contention), `persistRouteAndMessageId` silently discards the result. This is correct per the best-effort contract, but there is zero observability — no log line — when route/messageId fails to persist due to lock contention.

**Suggestion:** Log at debug level when the lock is not acquired, e.g. `console.debug('[handoff-answer] persistRouteAndMessageId: lock contention, skipping', { taskId })`.

---

**3 · low · test-coverage · `src/agent/daemon/handoff-store.test.ts:~440` · advisory**
The concurrent-race test for `withHandoffLock` relies on O_EXCL atomicity (correct, kernel-guaranteed), but a brief explanatory comment anchoring this would help future readers.

---

**4 · nit · api-compat · `src/agent/daemon/handoff-store.ts:~277`**
`withHandoffLock` takes `handoffsDir: string` as required (no default), unlike `updateHandoffAnswer` which defaults to `getHandoffsDir()`. Minor ergonomic asymmetry for future callers. No impact today.

---

### What was checked
- Citations verified inline against branch HEAD `e3ded57a`.
- Stated intent: PR #1567 title+body — spec-compliance assessed: all stated goals (extract `withHandoffLock`, refactor `updateHandoffAnswer`, wrap `persistRouteAndMessageId`) are implemented.
- `assertSafeJobId` absence in `handoff-answer.ts` confirmed via `git grep`.
- Logging absence around `persistRouteAndMessageId` confirmed via `git grep`.
- Did not run tests (CI shows green on lint+build+ubuntu; macOS in progress).

PR #1571, opened 2026-09-08
feat(daemon): default-on Done verification for unattended execution

griffinwork40:
## Summary

Flip `daemon.verifyDone` from opt-in (default off) to opt-out (default on).

When the daemon surface reports a task completion, and the agent self-certified `Done` without corroborating evidence (no successful file write/edit or verification command in the turn), the Telegram push is now downgraded to "⚠️ Done (unverified)" by default.

Operators who prefer the previous behavior can set `daemon.verifyDone: false` in `afk.config.json`.

## Rationale

The daemon surface is unattended by definition — no human watches the session. A self-certified completion with no evidence should be flagged, not silently accepted.

This was the highest-leverage reliability improvement identified by a cross-system research program that audited six agent systems (AFK, Hermes, OpenHands, Aider, Claude Code, Codex CLI). Evidence-Backed Completion is the most underserved engineering pattern — only 2/6 systems implement it, and neither defaulted it on for their unattended surface.

Empirical validation via a 10-task false-completion benchmark pilot (both systems on claude-sonnet-4-6):
- **AFK: 1 false completion in 10 tasks**
- **Claude Code: 5 false completions in 10 tasks**

Both systems actually completed 5/10 tasks. The difference is entirely in completion-claim behavior.

## Change

One expression flip: `config.daemon?.verifyDone === true` → `config.daemon?.verifyDone !== false`

Advisory, not blocking — nothing is prevented, the operator is informed.

## Testing

- 146 tests pass (daemon.test.ts, config.test.ts, daemon-config.test.ts)
- Full suite: 19,655 pass
- Lint clean, build clean, filesize ceiling clean

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,7 @@ tmp/
 todo.md
 src/web-ui-assets/
 pr1387-diff.txt
+
+# Local-only research artifacts (not for public repo)
+docs/research/
+tests/false-completion-benchmark.test.ts
--- a/src/cli/commands/daemon.ts
+++ b/src/cli/commands/daemon.ts
@@ -252,12 +252,13 @@ export function resolveNotifyChatTarget(
  * Format a daemon telemetry record for an out-of-band notification
  * (e.g. Telegram push). Short, scannable, status-first.
  *
- * `verifyDone` gates the opt-in "Done"-verification downgrade (mirrors
- * `daemon.verifyDone` in config). When it is `true` AND
- * `details.doneUnverified` is `true`, the ✅ success header is downgraded to a
- * "⚠️ Done (unverified)" header and the {@link DAEMON_DONE_UNVERIFIED_CAVEAT}
- * line is appended. When `verifyDone` is falsy (the default), the output is
- * byte-identical to before this feature existed — fail-open, opt-in.
+ * `verifyDone` gates the "Done"-verification downgrade (mirrors
+ * `daemon.verifyDone` in config — default: true since the daemon surface is
+ * unattended). When it is `true` AND `details.doneUnverified` is `true`, the
+ * ✅ success header is downgraded to a "⚠️ Done (unverified)" header and the
+ * {@link DAEMON_DONE_UNVERIFIED_CAVEAT} line is appended. When `verifyDone`
+ * is explicitly `false`, the output is byte-identical to before this feature
+ * existed.
  */
 export function formatTaskCompletion(
   record: TelemetryRecord,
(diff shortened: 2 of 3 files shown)


griffinwork40 (commented):

## Prior rounds

0 critical, 0 high, 0 medium blocking → **MERGE (with follow-ups)**; 4 low advisory, 2 nit advisory.

All three stated default flips are correctly implemented:
- `daemon.verifyDone`: `=== true` → `!== false` ✅
- `AFK_CAPTURE_SUBAGENT_OUTPUT`: `!== '0'` (on by default) → `!v || isExplicitlyDisabled(v)` (off by default) ✅
- `AFK_OPENAI_CHATGPT_OAUTH`: allowlist opt-in, `!v → false` ✅

`isExplicitlyDisabled` is used consistently across all affected modules. The `chatGptOAuthEnabled` allowlist pattern is correct for an off-by-default flag. JSDocs and env-registry docs are updated to match.

---

### 1. low · correctness · `daemon.ts:266`

`formatTaskCompletion` parameter default is `verifyDone = false` but the sole call site now passes `!== false` (default-on). The parameter default is dead for all real paths. The commit message acknowledges this ("stays false for backward compatibility with non-daemon callers"), but no non-daemon callers exist. Follow-up: align the default or add `@internal` JSDoc.

### 2. low · perf-observability · `tool-dispatch.ts:84`, `dispatch-append.ts:121`

`toolInputCapture` guard re-reads `env.AFK_CAPTURE_SUBAGENT_OUTPUT` and runs `isExplicitlyDisabled` regex on every tool dispatch. Bounded impact (one regex per tool call), but could be hoisted to a module-level boolean initialized once. Follow-up issue suggested.

### 3. low · security · `auth.ts:240-263` (pre-existing)

`parseCodexAuthJson()` classifies any `~/.codex/auth.json` with a `tokens.access_token` field as ChatGPT OAuth, even without `auth_mode: 'chatgpt'`. Mitigated: the token is gated behind `chatGptOAuthEnabled()` (now off by default). Pre-existing on main — not introduced by this PR.

### 4. low · security · `memory-evidence.ts:53-57`

Garbage env values for `AFK_MEMORY_EVIDENCE_GATE` silently keep the gate enabled. This is the correct safe default per the `isExplicitlyDisabled` contract — unrecognized values fall through to the default. A diagnostic warning for unrecognized boolean env values would help operators catch typos; out of scope for this PR.

### 5. nit · test-coverage · `subagent-output-capture.test.ts`

Disable-value aliases (`'false'`, `'no'`, `'off'`) are tested for `isExplicitlyDisabled` in isolation (env-helpers.test.ts) and end-to-end through `evidenceGateEnabled` (memory-evidence-gate.test.ts), but not end-to-end through `shouldCaptureSubagentOutput`. An `it.each` block would close the gap.

### 6. nit · correctness · import paths

All changed files import `isExplicitlyDisabled` from the barrel re-export (`../../config/env.js`) rather than directly from `env-helpers.js`. Consistent and correct — noted for awareness only.

---

## What was not checked

- Citations verified against branch HEAD `ea4196de1eddf8a9d717c5c9b59d378e04ae83c1`.
- Stated intent: PR #1571 title + body — spec-compliance assessed. All three stated flips match implementation.
- One Wave 1 finding (S-1) cited a non-existent `isExplicitlyEnabled` helper function — citation classified as fabricated and dropped; the underlying observation was preserved at reduced severity.
- One Wave 1 finding (S-2, parseCodexAuthJson fallback) confirmed as pre-existing on main and marked non-blocking per policy.
- Tests not executed (read-only review). CI shows all checks green.
- Merge conflict with main noted (GitHub reports CONFLICTING) — not assessed in this review.

PR #1572, opened 2026-09-08
feat(web): TUI-parity upgrade with 5 panels, emission wiring, and full audit fixes

griffinwork40:
## Web UI Protocol-First Upgrade

Takes the `afk web` UI from a read-only session viewer to a near-TUI-parity agent interface. Nine commits, 56 files changed, 4659 insertions.

### What's new

**Wave 3 — TUI parity features** (5 panels)
- **@file context injection** — `@` in composer opens a file picker popover; selected files are sent as context with the next prompt
- **Background jobs panel** — dedicated nav tab showing all bg-job history with status badges, models, durations
- **Model display/selector** — model badge on session sidebar rows; model picker in new-session form
- **Enhanced diff viewer** — syntax-highlighted unified diff rendering with hunk navigation
- **Memory browser** — hot memory viewer with usage meter + full-text search with category filters

**Wave 4 — Integration wiring**
- Nav buttons for Jobs/Memory/Schedules tabs
- @file affordance wired into composer
- Model badges on session sidebar

**Step 1A — Emission wiring**
- SubagentManager, BackgroundAgentRegistry, and plan-exit-bridge now emit OutputEvent variants (`subagent_lifecycle`, `background_job`, `plan_mode`) into the SSE stream
- `pushSidebandEvent()` on AgentSession for non-provider events

### Bug fixes (from parallel audit)

Five parallel audit subagents (CSS, JS events, API contracts, HTML structure, SSE lifecycle) identified ~25 bugs. All fixed:

**Critical**
- Live tool output never shown — `toolIndex` not passed during SSE streaming
- Scroll position lost on view switch — `isPinnedToBottom` true for hidden elements
- Module-global ID counter never reset across session switches
- Duplicate subagent cards — lifecycle events created new items instead of updating

**High**
- View switching: bg-jobs stayed visible after switching back to sessions (class vs hidden attr conflict)
- Memory panel invisible (inner div reused container's CSS class)
- Diff viewer silently clipped long lines (`overflow: hidden`)
- Leaked Escape handlers from schedule modals
- Session names showing plugin preamble text

**Medium**
- Memory panel never refreshed on revisit
- Schedule toggle/delete vulnerable to double-click
- Bg-jobs fetch had no AbortController (race on rapid switching)
- Schedule form read state from `style.display` instead of boolean
- SSE error frames silently dropped
- API error toasts showed raw JSON instead of human-friendly messages

**Accessibility**
- `type="button"` on all buttons
- `aria-label` on textarea and inputs
- `role="dialog"`, `aria-modal`, `aria-labelledby` on modals
- Label-input associations via `htmlFor`/`id`
- Focus traps in both schedule modals
- `aria-live="polite"` on status indicator
- Theme color variables replacing hardcoded hex values

### Testing

- 385/385 web-server tests pass
- 9472 agent tests pass
- All CI gates green (lint, filesize, sdk audit, env audit, chalk audit)


griffinwork40:
## Review: PR #1572 — feat(web): TUI-parity upgrade with 5 panels, emission wiring, and full audit fixes

**Reviewed ref:** `9e507dec` · **Regime:** full · **Citations verified inline against branch HEAD 9e507dec.**

---

### Decision: MERGE — 0 blocking (2 medium waived, 4 low, 2 nit)

No blocking defects. Two medium findings — one a known-incomplete feature, one a CSS positioning quirk — are waived per the waver rules below. All high and critical candidate claims were refuted by code inspection.

---

### Security · API-compat

**No security issues found** — read all three new route files.

New routes (`/api/models`, `/api/bg-jobs`, `/api/bg-jobs/:id`, `/api/memory/search`, `/api/memory/hot`) all sit below the `tokensMatch` bearer check at `server.ts:269` and the Origin CSRF gate at line `276`. Auth coverage is complete.

`/api/bg-jobs/:id` validates `jobId` against `/^[a-zA-Z0-9-]+$/` before touching the filesystem (`routes.bg-jobs.ts:15`), preventing path traversal via encoded separators.

`/api/memory/search` validates `category` against a `Set` of allowed values (`routes.memory.ts:18`) and clamps `limit` at 100 (`routes.memory.ts:19`). The `MemoryStore.search()` accepts `limit` through `SearchOpts` — confirmed in `src/agent/memory/memory-store.ts:678`.

No innerHTML with agent-derived text found in any new module. All model code paths use `textContent`. XSS surface is unchanged.

**api-compat:** `pushSidebandEvent` is additive on `AgentSession`; `setOutputEventSink` is a new opt-in on `SubagentManager`. Zero production importers broken. `SubagentItem` and `BgJobItem` are additive members of the `TranscriptItem` discriminated union — all exhaustive switches that existed before this PR handled `NoticeItem` as the final arm and pass through unchanged by TypeScript's exhaustiveness rules.

---

### Correctness · Spec-compliance · Test-coverage · Perf-observability

#### Medium — tool→tool_result correlation silently no-ops on replay (waived)

`severity: medium · blocking: false · waived: feature gap in Wave 1, not a regression` · `confidence: high` · `dimension: correctness` · `ref: 9e507dec`

`projectOutputEvent` at `src/agent/session-ledger-project.ts:122` writes the `'tool'` ledger record as `{ kind: 'tool', toolName, input }` — **no `toolUseId` field** — even though `ToolUseDetailChunk.toolUseId` is available on the chunk at that call site. The ledger-adapter's `toolIndex` is keyed on `toolUseId` (`ledger-adapter.ts:113`), so the index is never populated from a `'tool'` record, and every `'tool_result'` record falls through to the standalone-notice path.

Practical effect: on session replay (page refresh), successful tool output is shown as a `tool result: <first 80 chars>` notice row rather than being patched into the originating tool card. The card retains `outputUnavailable: true`. The live SSE path (new sessions) is also affected because ledger records stream in the same format.

This is a **Wave 1 known-incomplete design**: the adapter comment ("Pre-Wave 1 ledgers have no `toolUseId` on `tool` records, so the index is best-effort") correctly documents the gap but doesn't signal that Wave 1 itself also doesn't supply it. No existing test covers the cross-linking path. The fix is one additional field in the `projectOutputEvent` `'tool'` arm and adding a `toolUseId?` to the `'tool'` `LedgerPayload` member.

**Waived** because: this is a bounded UX gap (output shown as a separate notice rather than disappearing), not a data-integrity or auth issue, and it matches what pre-Wave-1 behavior already did (tool output was never shown on reload). A follow-up PR can close the loop.

---

#### Medium — compose-executor subagents invisible to SSE (waived)

`severity: medium · blocking: false · waived: compose is a distinct fork path, no regression from baseline` · `confidence: high` · `dimension: correctness` · `ref: 9e507dec`

`bootstrap.ts:209` wires `rootManager.setOutputEventSink(sidebandSink)`, but `compose-executor.ts:677` creates a **fresh `SubagentManager` per `execute()` call** and never receives `setOutputEventSink`. Compose-DAG subagents do not emit `subagent_lifecycle` events into the SSE stream. The web UI's subagent cards will be invisible for compose dispatches.

**Waived** because: this is the same gap as `setOnSubagentSucceeded` for compose (which the PR separately wires for cost rollup but the comment at `bootstrap.ts:197–202` makes explicit was a pre-existing issue). The web-UI feature is new; compose subagent visibility was never shown before this PR. A follow-up wiring in compose-executor mirrors the `setOnSubagentSucceeded` pattern already there.

---

#### Low — `at-file.css` invariant comment incorrect: `.composer-row` has no `position: relative`

`severity: low · blocking: false · confidence: high · dimension: correctness · ref: 9e507dec`
`file: src/web-server/frontend/at-file.css:31–32 · citation-type: file-state`

```css
/* Invariant: the popover is positioned absolute relative to .composer-row
   (which has position:relative via the composer-row rule). */
```

`styles.css:314` defines `.composer-row { display: flex; gap: 10px; align-items: flex-end; }` — no `position: relative`. The popover's `position: absolute; bottom: 100%; right: 0` resolves against the nearest positioned ancestor, which is `.composer` (`styles.css:282`). The popover appears in the right place visually (`.composer` is a reasonable stacking context), but the comment is factually wrong, and the z-index ladder comment in `styles.css:56` cites `at-popover=25` which is already set in `at-file.css`.

**Suggestion:** Either add `position: relative` to `.composer-row` in `styles.css` to match the comment, or correct the comment. No functional breakage observed — `.composer` as the stacking parent gives correct visual placement.

---

#### Low — `isPreamble` regex false-positives on `[word word]` Markdown link syntax

`severity: low · blocking: false · confidence: medium · dimension: correctness · ref: 9e507dec`
`file: src/web-server/session-source.ts:54 · citation-type: file-state`

```ts
function isPreamble(text: string): boolean {
  return /^\s*\[[\w-]+[:\s]/.test(text);
}
```

`[\w-]+` matches one or more word-chars or dashes, and `[:\s]` matches a colon or whitespace. A user who starts their first message with `[test] here's what I want` or `[link text] do the thing` will have their session title silently skipped in favor of the next user turn. Real preambles are `[plugin-name: unlocked]` (colon-separated key-value) or `[skill-routing: active]`, which both contain a colon. A tighter pattern like `/^\s*\[[\w-]+:\s/` (requiring a colon separator) would avoid matching normal bracketed labels.

---

#### Low — subagent_lifecycle terminal event missing `totalCostUsd`, `errorClass`, `errorMessage`

`severity: low · blocking: false · confidence: high · dimension: correctness · ref: 9e507dec`
`file: src/agent/subagent.ts:491–497 · citation-type: file-state`

```ts
this.outputEventSink?.({
  type: 'subagent_lifecycle',
  subagentId: id,
  status: terminalStatus,
  ...(handle._lastDurationMs !== undefined ? { durationMs: handle._lastDurationMs } : {}),
});
```

At terminal time `handle` also carries `costUsd` (available via `onSubagentSucceeded` which fires two lines earlier on success) and the `LedgerPayload` `subagent_lifecycle` type includes `totalCostUsd`, `outputBytes`, `errorClass`, `errorMessage`. The web-UI subagent card (`renderSubagentCard` in `render.ts`) would show cost if emitted. This is a gap between the richer ledger payload spec and what the live SSE event carries — no display regression since cost was never shown before.

---

#### Low — `subagent-tree.ts` and `tool-trace-panel.ts` are unreferenced dead files

`severity: low · blocking: false · confidence: high · dimension: correctness · ref: 9e507dec`
`files: src/web-server/frontend/subagent-tree.ts, src/web-server/frontend/tool-trace-panel.ts`

Neither file is imported by any other module in `src/web-server/frontend/` (confirmed: no grep hits for `subagent-tree` or `tool-trace-panel` across the frontend directory). They appear to be Wave 2 artifacts included prematurely — well-written code that esbuild will tree-shake from the bundle, but adding surface area to the review and test matrix without delivering functionality in this PR.

`tool-trace-panel.ts` also imports `type { DiffPayload, DiffHunk, DiffLine } from '../../utils/diff.js'` — a server-side module. That import would survive tree-shaking only if the file is bundled; since it isn't wired in yet, this is a latent type-only import, not a runtime issue.

---

#### Nit — `mem-search-btn` in `memory-panel.ts` lacks `type="button"`

`severity: nit · blocking: false · confidence: high · dimension: correctness · ref: 9e507dec`
`file: src/web-server/frontend/memory-panel.ts:198`

```ts
const btn = el('button', 'mem-search-btn', 'Search');
```

No `.type = 'button'` set. The PR explicitly adds `type="button"` to all other dynamically created buttons (at-file-panel, app-chrome, schedule-form, schedule-history, schedules-view). The memory panel is not wrapped in a `<form>`, so `type="submit"` (the default) is harmless, but the inconsistency is worth a one-liner fix.

---

#### Nit — `plan_mode` output event emits only `'plan'` and `'default'`, never `'implement'`

`severity: nit · blocking: false · confidence: high · dimension: correctness · ref: 9e507dec`
`file: src/agent/session/agent-session.ts:988`

```ts
this.pushSidebandEvent({ type: 'plan_mode', mode: mode === 'plan' ? 'plan' : 'default' });
```

The `OutputEvent` union type for `plan_mode` declares `mode: 'plan' | 'implement' | 'default'` (`session-types.ts`), but the call site only emits `'plan'` or `'default'`. If `'implement'` is a meaningful state for other consumers, this is a gap. If it isn't used, the type union should drop it.

---

### What was not checked

- Citations verified inline against branch HEAD `9e507dec`. No fabricated citations found.
- Did not exercise the UI in a running browser — all analysis was static diff + source review.
- Did not verify SSE streaming behavior end-to-end (ledger replay path verified; live stream path analyzed by code).
- `session-status.ts` exports `groupSessionsByStatus` and `renderGroupedSessionList` which are not called from any live wiring in this PR — they appear to be Wave 2-Beta scaffolding and are not dead (imported by render.ts for `classifySession` and `renderStatusBadge`).
- Stated intent: PR #1572 title + body — spec-compliance assessed. The PR delivers what it describes. No unmet intent or material scope creep detected.
- `subagent-tree.ts` and `tool-trace-panel.ts` dead code: absence confirmed by grep (`rg`), zero importers in `src/web-server/frontend/`.

PR #1591, opened 2026-09-09
fix(web): harden build-web-ui.mjs with CI fail-fast and pinned Vite

### LINKED ISSUE #1585: web: harden build-web-ui.mjs — CI fail-fast and pinned Vite
## Context

PR #1580's `scripts/build-web-ui.mjs` has two build-script concerns:

1. **Silent fallback on Vite failure** — when Vite build fails, the catch block falls back to legacy esbuild and exits 0. In CI, a failed Vite build ships whatever legacy output exists with no error signal.
2. **`npx` indirection** — `execSync('npx vite build', ...)` can silently download a different Vite version if the local one is not found by npx's resolution.

## Proposed fix

1. Add a `CI` env-var guard: exit non-zero in CI when Vite fails, allow local dev fallback otherwise
2. Use `node_modules/.bin/vite build` relative to `dashboardDir` (or `pnpm --filter afk-dashboard build`) to ensure the locked Vite version is always used

## Acceptance criteria

- CI fails if Vite build fails (no silent fallback to legacy in CI)
- Local dev can still fall back to legacy
- Vite invocation uses the lockfile-pinned version, not npx resolution

Ref: PR #1580 review findings (medium waived + low)

griffinwork40:
Closes #1585

## Changes

- **Pinned Vite binary**: replaced `npx vite build` with `node_modules/.bin/vite build` (relative, with `shell:true` and `cwd: dashboardDir`). This always invokes the version pinned in `dashboard/package.json` — never a version silently downloaded by npx. Cross-platform: uses `.cmd` extension on Windows via `platform()` detection.

- **CI fail-fast guard**: when `process.env.CI` is truthy and the Vite build throws, the script now calls `process.exit(1)` instead of falling back to the legacy esbuild path. Local dev retains the fallback behaviour for convenience.

## Verification

```
pnpm lint   # tsc --noEmit — clean
pnpm build  # tsc + build-web-ui.mjs + copy-prompts — clean, 11 files → src/web-ui-assets/
```

### DIFF
--- a/scripts/build-web-ui.mjs
+++ b/scripts/build-web-ui.mjs
@@ -21,6 +21,7 @@ import { fileURLToPath } from 'node:url';
 import { dirname, join } from 'node:path';
 import { mkdirSync, copyFileSync, existsSync, readdirSync, statSync } from 'node:fs';
 import { execSync } from 'node:child_process';
+import { platform } from 'node:os';
 
 const repoRoot = join(dirname(fileURLToPath(import.meta.url)), '..');
 const frontendDir = join(repoRoot, 'src', 'web-server', 'frontend');
@@ -47,14 +48,30 @@ async function main() {
 /** React SPA build via Vite. Outputs directly to src/web-ui-assets/. */
 async function buildDashboard() {
   console.log('build-web-ui: building React dashboard (Vite)...');
+
+  // Use the local node_modules/.bin/vite so we always run the pinned version
+  // from dashboard/package.json — never a version silently downloaded by npx.
+  // On Windows, Node resolves shell:true against PATH which includes .bin/, so
+  // the cross-platform pattern is: relative path with shell:true and cwd set.
+  const viteBin =
+    platform() === 'win32'
+      ? join('node_modules', '.bin', 'vite.cmd')
+      : join('node_modules', '.bin', 'vite');
+
   try {
-    execSync('npx vite build', {
+    execSync(`"${viteBin}" build`, {
       cwd: dashboardDir,
       stdio: 'inherit',
+      shell: true,
       env: { ...process.env, NODE_ENV: 'production' },
     });
   } catch {
-    console.error('build-web-ui: Vite build failed, falling back to legacy');
+    if (process.env.CI) {
+      // In CI the fallback would silently ship stale assets — fail loudly instead.
+      console.error('build-web-ui: Vite build failed in CI — exiting non-zero');
+      process.exit(1);
+    }
+    console.error('build-web-ui: Vite build failed, falling back to legacy esbuild');
     await buildLegacy();
   }
 }
--- a/scripts/build-web-ui.mjs
+++ b/scripts/build-web-ui.mjs
@@ -65,11 +65,12 @@ async function buildDashboard() {
       shell: true,
       env: { ...process.env, NODE_ENV: 'production' },
     });
-  } catch {
+  } catch (err) {
+    console.error(`build-web-ui: Vite exited status=${err?.status ?? '?'} signal=${err?.signal ?? 'none'}`);
     if (process.env.CI) {
       // In CI the fallback would silently ship stale assets — fail loudly instead.
       console.error('build-web-ui: Vite build failed in CI — exiting non-zero');
-      process.exit(1);
+      throw new Error('build-web-ui: Vite build failed in CI');
     }
     console.error('build-web-ui: Vite build failed, falling back to legacy esbuild');
     await buildLegacy();


griffinwork40 (commented):

---

## PR #1591 · fix(web): harden build-web-ui.mjs with CI fail-fast and pinned Vite

**Reviewed ref:** `fc6c55a87352a24192bae68c7f7da35d344aa431`

---

### Prior automated rounds

---

### Decision: MERGE (with follow-ups)

**Arithmetic:** 0 critical, 0 high → MERGE (with follow-ups); 3 medium advisory (non-blocking), 3 low advisory.

---

**[medium · non-blocking · advisory · correctness]** `scripts/build-web-ui.mjs:62` · ref: fc6c55a8 · diff-context

`execSync` with `shell: true` and a template-literal command (`\`"${viteBin}" build\``) relies on correct quote handling by the host shell. On Windows, `cmd.exe` has documented quirks with outer-quote stripping when the executable path is quoted. All path components (`node_modules`, `.bin`, `vite.cmd`) are hardcoded string literals — no user input enters the command string, so this is **not** an injection vector. However, `spawnSync(viteBin, ['build'], { cwd, stdio, env })` eliminates shell interpretation entirely and is the idiomatic Node.js pattern for known binaries.

*Suggest:* Replace `execSync` + `shell: true` with `spawnSync(viteBin, ['build'], { cwd: dashboardDir, stdio: 'inherit', env: { ...process.env, NODE_ENV: 'production' } })`. Import `spawnSync` from `node:child_process`. This also makes the error object's `.status` field reliably reflect Vite's exit code (not the shell's).

**[medium · non-blocking · advisory · correctness]** `scripts/build-web-ui.mjs:69` · ref: fc6c55a8 · diff-context

`process.exit(1)` terminates synchronously, bypassing the `main().catch(err => { ... process.exit(1) })` handler at the call site. The outer catch already exits non-zero on any thrown error. Using `throw new Error('Vite build failed in CI')` instead keeps a single exit point and preserves any registered cleanup.

*Suggest:* Replace `process.exit(1)` with `throw new Error('Vite build failed in CI')`.

**[medium · non-blocking · advisory · correctness]** `scripts/build-web-ui.mjs:69` · ref: fc6c55a8 · diff-context

`if (process.env.CI)` is truthy for any non-empty string including `"false"` and `"0"`. GitHub Actions sets `CI=true` (safe), but a developer with `export CI=false` in their shell profile silently loses the local fallback. Bounded impact — the worst case is a local dev build failing loudly instead of falling back.

*Suggest:* Use `process.env.CI === 'true'` for stricter matching, or search the codebase for the project's existing convention.

**[low · non-blocking · test-coverage]** `scripts/build-web-ui.mjs:53–77` · ref: fc6c55a8 · diff-context

No test file covers the new `platform()` branch, the `CI` guard, or the `process.exit(1)` path. Search for `buildDashboard`, `viteBin`, `vite.cmd`, `build-web-ui` across `*.test.*` / `*.spec.*` — zero matches. The script is a build tool with no test companion. Not blocking — the gap predates this PR.

**[low · non-blocking · perf-observability]** `scripts/build-web-ui.mjs:69–72` · ref: fc6c55a8 · diff-context

The `catch` block is a bare catch (no error parameter). Vite's exit code (`err.status`) and signal (`err.signal`) are not logged. The CI error message gives no indication of whether Vite exited 1, 127 (binary not found), or was killed by signal.

*Suggest:* Capture the error: `} catch (err) {` and log `err.status` / `err.signal` before the exit.

**[low · non-blocking · api-compat]** `scripts/build-web-ui.mjs:65` · ref: fc6c55a8 · diff-context

Adding `shell: true` changes the error object semantics: `.status` now reflects the shell's exit code, not necessarily Vite's. No production callers inspect this error object (the script runs standalone), so impact is nil.

---

### Spec-compliance

Stated intent: PR #1591 title + body — spec-compliance assessed.
- ✅ Pinned Vite binary: `npx vite build` replaced with `node_modules/.bin/vite` (or `.cmd` on Windows) — matches stated intent.
- ✅ CI fail-fast guard: `process.env.CI` check added, `process.exit(1)` on failure — matches stated intent.
- ✅ Local dev fallback preserved: non-CI catch path still calls `buildLegacy()` — matches stated intent.
- ⚠️ The "pinned" invariant is only enforced when `node_modules` exists; no preflight check for the `vite` binary specifically. Minor gap between intent ("always invokes the pinned version") and implementation.

### What was not checked

- Citations verified inline against branch HEAD `fc6c55a87352a24192bae68c7f7da35d344aa431`.
- Stated intent: PR #1591 title + body — spec-compliance assessed.
- Did not run tests locally (CI results used — all green).
- Did not verify Windows cmd.exe quote-stripping at runtime — finding based on documented Node.js behavior.
- Wave 1 findings F1 and F2 (both originally rated high/blocking) downgraded during synthesis: the shell injection surface is not reachable (all path components are hardcoded string literals, no user input), and the Windows cmd.exe quote-stripping is speculative (CI passes on Windows). The `spawnSync` recommendation stands as a code-quality improvement, not a security fix.

PR #1593, opened 2026-09-09
fix(web): fix stale caches in React dashboard (#1584)

### LINKED ISSUE #1584: web: fix stale caches in React dashboard (commands, token, SSE)
## Context

PR #1580 introduced module-global caches in the React dashboard that are never invalidated:

1. **`commandsCache`** in `composer.tsx` (line 23) — caches the slash-command list forever. If a skill is added at runtime, the stale list is served until page reload.
2. **`cachedToken`** in `api.ts` — caches the bearer token read from the meta tag. If the server rotates the token (new `afk web` invocation), the React app silently uses the stale token until manually reloaded.
3. **SSE race on unmount** in `use-sse-stream.ts` — between effect teardown and the reader loop checking `stoppedRef`, a frame could be processed for the old session on fast session switching (sub-millisecond window).

## Proposed fix

1. Expose `clearCommandsCache()` and call it on page focus (`visibilitychange`) or session change
2. Re-read token from the meta tag on 401 response instead of caching forever
3. Add a `stoppedRef.current` check after `await reader.read()` before processing frames

## Acceptance criteria

- Slash commands refresh on page focus or session switch
- A 401 triggers token re-read from meta tag, not a permanent failure
- No stale-session frames rendered during fast session switching

Ref: PR #1580 review findings (medium, advisory)

griffinwork40:
## Summary

Fixes #1584 — three independent stale-cache bugs in the React dashboard.

### Changes

**`dashboard/src/components/composer.tsx`**
- Added `export function clearCommandsCache()` to allow external invalidation.
- Added a `useEffect` that listens for `visibilitychange` and calls `clearCommandsCache()` when the tab regains focus. Skills added at runtime will appear in the autocomplete list on the next `/` keystroke after switching back to the tab — no page reload needed.

**`dashboard/src/lib/api.ts`**
- On a 401 response, `apiFetch` now calls a local `clearTokenCache()` helper to null out `cachedToken` before throwing. The next request re-reads the token from the `<meta name="afk-token">` tag, picking up a rotated server token transparently.

**`dashboard/src/hooks/use-sse-stream.ts`**
- Added a `stoppedRef.current` guard immediately after `await reader.read()` returns, before any frame processing. If the effect was torn down during the await (unmount or session change), the frame is discarded instead of being forwarded to the new session's state.

### Verification
- `pnpm lint` (tsc --noEmit on both tsconfigs) passes clean.
- All three files remain under 350 LOC (187 / 93 / 224).

### DIFF
--- a/dashboard/src/components/composer.tsx
+++ b/dashboard/src/components/composer.tsx
@@ -17,8 +17,13 @@ interface ComposerProps {
   isBusy: boolean;
 }
 
-/** Cached slash commands — fetched once per page lifecycle. */
+/** Cached slash commands — cleared on page focus to pick up new skills. */
 let commandsCache: SlashCommand[] | null = null;
+
+/** Clear the cached command list so the next '/' keystroke re-fetches. */
+export function clearCommandsCache(): void {
+  commandsCache = null;
+}
 
 async function fetchCommands(): Promise<SlashCommand[]> {
   if (commandsCache) return commandsCache;
@@ -47,6 +52,15 @@ export function Composer({ sessionId, sessionMode, isBusy }: ComposerProps) {
   const acFiltered = acVisible
     ? commands.filter((c) => c.name.startsWith(slashQuery)).slice(0, 8)
     : [];
+
+  // Clear the command cache when the tab regains focus so new skills are visible.
+  useEffect(() => {
+    const onVisible = () => {
+      if (document.visibilityState === 'visible') clearCommandsCache();
+    };
+    document.addEventListener('visibilitychange', onVisible);
+    return () => document.removeEventListener('visibilitychange', onVisible);
+  }, []);
 
   // Auto-resize textarea (1–5 rows)
   useEffect(() => {
--- a/dashboard/src/hooks/use-sse-stream.ts
+++ b/dashboard/src/hooks/use-sse-stream.ts
@@ -158,6 +158,9 @@ export function useSseStream(sessionId: string | null): UseSseStreamResult {
 
         outer: for (;;) {
           const { done, value } = await reader.read();
+          // Guard: the effect may have been torn down while we were awaiting;
+          // discard this frame rather than forwarding it to a new session.
+          if (stoppedRef.current) break;
           if (done) break;
           buffer += decoder.decode(value, { stream: true });
           const { events: frames, remainder } = parseSseChunk(buffer);
--- a/dashboard/src/lib/api.ts
+++ b/dashboard/src/lib/api.ts
@@ -13,6 +13,11 @@ let cachedToken: string | null = null;
 /** Evict the in-memory token cache (e.g. after a 401 response). */
 function clearTokenCache(): void {
   cachedToken = null;
+  try {
+    sessionStorage.removeItem(TOKEN_STORAGE_KEY);
+  } catch {
+    // sessionStorage may be unavailable in some contexts
+  }
 }
 
 /** Read the bearer token, preferring the meta tag, falling back to storage. */
--- a/dashboard/src/lib/api.ts
+++ b/dashboard/src/lib/api.ts
@@ -10,6 +10,11 @@ const TOKEN_STORAGE_KEY = 'afk_web_token';
 
 let cachedToken: string | null = null;
 
+/** Evict the in-memory token cache (e.g. after a 401 response). */
+function clearTokenCache(): void {
+  cachedToken = null;
+}
+
 /** Read the bearer token, preferring the meta tag, falling back to storage. */
 export function getToken(): string {
   if (cachedToken) return cachedToken;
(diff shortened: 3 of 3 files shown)


griffinwork40 (commented):

---

## PR #1593 · fix(web): fix stale caches in React dashboard (#1584)

**Reviewed ref:** `25aa54791cabc1eb735a8c205b3410473c2339e9`

---

### Prior automated rounds

---

### Decision: MERGE (with follow-ups)

**Arithmetic:** 0 critical, 0 high → MERGE (with follow-ups); 3 medium advisory (non-blocking), 2 low, 2 nit.

---

**[medium · non-blocking · advisory · correctness]** `dashboard/src/lib/api.ts:15-17` · ref: 25aa5479 · diff-context

**`clearTokenCache()` does not evict `sessionStorage`, leaving a stale-token path in the cookie-auth/refresh flow.** `clearTokenCache()` sets `cachedToken = null` but does not call `sessionStorage.removeItem(TOKEN_STORAGE_KEY)`. After a 401, the next `getToken()` call finds `cachedToken === null`, falls through the meta-tag check (which may be absent on a page refresh), then reads the stale token from `sessionStorage` (lines 38–43) — re-caching and returning the same credential that caused the 401.

```ts
// getToken() fallback at api.ts:38-43 — unchanged by this PR:
const stored = sessionStorage.getItem(TOKEN_STORAGE_KEY);
if (stored) {
  cachedToken = stored;
  return stored;
}
```

*Suggest:* Add `try { sessionStorage.removeItem(TOKEN_STORAGE_KEY); } catch { /* ignore */ }` inside `clearTokenCache()`. This makes the 401-token-rotation fix effective in all code paths, not just the meta-tag-present one.

**[medium · non-blocking · advisory · spec-compliance]** `dashboard/src/lib/api.ts:70-74` · ref: 25aa5479 · diff-context

**Stated intent says "the next request re-reads from the meta tag" but the `sessionStorage` fallback intercepts first.** This is the spec-compliance dimension of the finding above: the fix is effective only when the meta tag is present and up-to-date. In the cookie-auth/refresh flow where the meta tag is absent, `getToken()` returns the stale token from `sessionStorage`.

**[medium · non-blocking · advisory · correctness]** `dashboard/src/hooks/use-sse-stream.ts:160-164` · ref: 25aa5479 · diff-context

**`stoppedRef` guard is placed before `if (done) break`, but the abort path raises `AbortError` rather than returning a sentinel.** The guard is behaviorally correct for the session-change race (where a new session's cleanup has set `stoppedRef.current` while the old loop's `reader.read()` was awaiting a chunk). However, the comment "the effect may have been torn down while we were awaiting" slightly mischaracterises the mechanism — effect teardown calls `abort()`, which throws `AbortError` caught by the `catch` block, not a sentinel return. The guard's real target is the session-ID-change race.

*Suggest:* Reorder to `if (done) break; if (stoppedRef.current) break;` and update the comment to reference the session-change race specifically.

---

### Non-blocking observations

**[low · non-blocking · perf-observability]** `dashboard/src/components/composer.tsx:56-63` · ref: 25aa5479 · diff-context

The `visibilitychange` listener invalidates the command cache on every visible transition (including rapid alt-tab), with no debounce. Each invalidation causes a re-fetch on the next `/` keystroke. Impact is low in practice (the fetch only fires on the next keystroke, not on the visibility event itself).

**[low · non-blocking · test-coverage]** `dashboard/` · ref: 25aa5479

No test files exist under `dashboard/` at this ref — all three bug fixes land with zero automated test coverage. This is a pre-existing gap (the dashboard shipped without a test suite), not introduced by this PR.

**[nit · non-blocking · api-compat]** `dashboard/src/components/composer.tsx:25-27` · ref: 25aa5479 · diff-context

`clearCommandsCache()` is exported from a React component file. No external caller imports it today (`app.tsx:10` imports only `Composer`). The export couples future callers to the module-level cache implementation detail.

**[nit · non-blocking · perf-observability]** `dashboard/src/hooks/use-sse-stream.ts:162-164` · ref: 25aa5479 · diff-context

Guard comment does not identify which session the discarded frame belongs to. Adding `(sessionId=${sid})` would make the stale-session race explicit in debug logs.

---

### Dimensions with no issues

- **security** — no issues found; `clearTokenCache` is unexported, token not logged or stored in new locations, 401 path is safe
- **auth** — token eviction is directionally correct; the sessionStorage gap is a follow-up, not a security concern

### Spec-compliance

Stated intent: PR #1593 title + body — spec-compliance assessed.
- ✅ `composer.tsx`: `clearCommandsCache()` + `visibilitychange` listener — matches stated intent
- ⚠️ `api.ts`: `cachedToken` cleared on 401, but `sessionStorage` fallback not cleared — partially meets stated intent (medium advisory above)
- ✅ `use-sse-stream.ts`: `stoppedRef.current` guard added after `reader.read()` — matches stated intent

### What was not checked

- Citations verified inline against branch HEAD `25aa54791cabc1eb735a8c205b3410473c2339e9`.
- Stated intent: PR #1593 title+body — spec-compliance assessed.
- No test files exist under `dashboard/` to verify; absence confirmed via `git ls-tree`.
- Did not verify runtime behavior of `visibilitychange` across browsers — analysis is based on the DOM spec.
- The `sessionStorage` gap analysis is static; did not verify whether any deployment configuration keeps the meta tag present across all refresh scenarios.

PR #1599, opened 2026-09-09
fix(dashboard): resolve 401 errors in Vite dev mode

griffinwork40:
## Problem

Running the dashboard via `pnpm dev` (Vite at `:5173`) produces a wall of **401 Unauthorized** errors on every `/api/sessions` and `/api/pending` call.

Root cause: Vite serves `index.html` from disk, so the `__AFK_WEB_TOKEN__` placeholder in the `<meta name="afk-token">` tag is never replaced by the `afk web` backend. `getToken()` reads that literal string as a token and sends `Authorization: Bearer __AFK_WEB_TOKEN__` on every request.

## Fix

**`dashboard/vite.config.ts`** - Adds a `transformIndexHtml` Vite plugin that replaces the placeholder with the `AFK_WEB_TOKEN` env var during dev:

```bash
AFK_WEB_TOKEN=<token> pnpm dev
```

**`dashboard/src/lib/api.ts`** - Guards `getToken()` against the raw placeholder (defense in depth, matching the existing guard in `src/web-server/frontend/web-token.ts:63`). An un-replaced placeholder now yields an empty string instead of silently authenticating as the literal placeholder.

## Verification

- `pnpm lint` clean
- `pnpm test src/web-server/server.test.ts` - 75/75 pass

### DIFF
--- a/dashboard/src/lib/api.ts
+++ b/dashboard/src/lib/api.ts
@@ -8,6 +8,9 @@
 
 const TOKEN_STORAGE_KEY = 'afk_web_token';
 
+/** The server replaces this in the HTML; when we read it raw, there is no token. */
+const TOKEN_PLACEHOLDER = '__AFK_WEB_TOKEN__';
+
 let cachedToken: string | null = null;
 
 /** Read the bearer token, preferring the meta tag, falling back to storage. */
@@ -17,7 +20,7 @@ export function getToken(): string {
   const meta = document.querySelector<HTMLMetaElement>('meta[name="afk-token"]');
   const metaValue = meta?.content;
 
-  if (metaValue) {
+  if (metaValue && metaValue !== TOKEN_PLACEHOLDER) {
     cachedToken = metaValue;
     try {
       sessionStorage.setItem(TOKEN_STORAGE_KEY, metaValue);
--- a/dashboard/vite.config.ts
+++ b/dashboard/vite.config.ts
@@ -4,7 +4,20 @@ import tailwindcss from '@tailwindcss/vite';
 import { resolve } from 'node:path';
 
 export default defineConfig({
-  plugins: [react(), tailwindcss()],
+  plugins: [
+    react(),
+    tailwindcss(),
+    // Dev-only: replace the __AFK_WEB_TOKEN__ placeholder with the real token
+    // so the dashboard can authenticate against a running `afk web` backend.
+    // Set AFK_WEB_TOKEN in the environment to match the backend's token.
+    {
+      name: 'afk-dev-token',
+      transformIndexHtml(html) {
+        const token = process.env['AFK_WEB_TOKEN'] ?? '';
+        return html.replace('__AFK_WEB_TOKEN__', token);
+      },
+    },
+  ],
   resolve: {
     alias: {
       '@': resolve(__dirname, './src'),
--- a/dashboard/vite.config.ts
+++ b/dashboard/vite.config.ts
@@ -3,6 +3,18 @@ import react from '@vitejs/plugin-react';
 import tailwindcss from '@tailwindcss/vite';
 import { resolve } from 'node:path';
 
+const WEB_SERVER_ORIGIN = 'http://127.0.0.1:4141';
+
+/** Escape an operator-supplied token before placing it in a quoted attribute. */
+function escapeHtmlAttribute(value: string): string {
+  return value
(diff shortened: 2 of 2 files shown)


griffinwork40:
@codex fix comments

griffinwork40 (commented):

---

## PR #1599 · fix(dashboard): resolve 401 errors in Vite dev mode

**Reviewed ref:** `dab64df43bd51bd3909044b9117da541589443e9`

---

### Prior automated rounds

---

### Severity arithmetic & Decision

**0 critical, 0 high → MERGE (with follow-ups); 1 medium advisory, 3 low advisory.**

Per the blocking floor: only `critical` and `high` findings block merge. The single medium finding (security — `String.replace` vs production `split().join()`) is advisory and recommended as a one-line follow-up.

---

**[medium · non-blocking · advisory · security]** `dashboard/vite.config.ts:30` · ref: dab64df4 · file-state

**`String.replace()` is first-occurrence-only; production uses `split().join()` (replaceAll).** `transformIndexHtml` calls `html.replace('__AFK_WEB_TOKEN__', ...)` which replaces only the first match. The production path in `src/web-server/static-assets.ts:154` uses `.split(TOKEN_PLACEHOLDER).join(...)` — a replaceAll idiom. If a future change adds a second `__AFK_WEB_TOKEN__` occurrence in `index.html`, the dev path silently leaves it unreplaced while production handles it correctly.

*Suggest:* `return html.split('__AFK_WEB_TOKEN__').join(escapeHtmlAttribute(token));` — matches the production idiom, one-line change.

**[low · non-blocking · advisory · security]** `dashboard/vite.config.ts:6,64` · ref: dab64df4 · file-state

**Hardcoded `http://127.0.0.1:4141` breaks for non-default `afk web` ports.** `WEB_SERVER_ORIGIN` is hardcoded. A developer running `afk web --port 5000` gets 403 on every non-GET proxied request because the spoofed `Origin` does not match the backend's `allowedOrigins(host, port)` check.

*Suggest:* Read the port from `AFK_WEB_PORT` env var, or document the limitation.

**[low · non-blocking · advisory · security]** `dashboard/src/lib/api.ts:12`, `dashboard/vite.config.ts:30` · ref: dab64df4 · file-state

**Three independent copies of the `'__AFK_WEB_TOKEN__'` string literal** exist across `api.ts:12`, `vite.config.ts:30`, and the authoritative `src/web-server/frontend/web-token.ts:21`. The dashboard build cannot import from the server package, so this is structural — but a shared `dashboard/src/lib/constants.ts` would reduce the literal count from 3 to 2.

**[low · non-blocking · advisory · correctness]** `dashboard/vite.config.ts:29` · ref: dab64df4 · file-state

**Silent no-op when `AFK_WEB_TOKEN` is unset.** The plugin replaces the placeholder with `''`, producing `content=""`. `getToken()` falls through to sessionStorage, and if empty, 401s resume — the same error the PR fixes. No startup warning surfaces the misconfiguration.

*Suggest:* Add `if (!token) this.warn('AFK_WEB_TOKEN is not set — API calls will return 401')` in the plugin.

---

### Spec-compliance

Stated intent: PR #1599 title + body — spec-compliance assessed.

- ✅ `transformIndexHtml` replaces `__AFK_WEB_TOKEN__` during Vite dev serve (`apply: 'serve'`)
- ✅ `getToken()` guard rejects the raw placeholder string (matches existing `web-token.ts` guard)
- ✅ Production builds unaffected (`apply: 'serve'` scopes to dev only)
- ✅ CSRF Origin fix rewrites `Origin` header for proxied non-GET requests
- ✅ No unmet intent items; no scope creep

### Dimensions with no issues

- **api-compat** — no breaking changes to exported `apiFetch`/`ApiError`/`getToken` surface; Vite plugin uses documented hooks
- **perf-observability** — single string op per page load in dev; no-op in production builds
- **test-coverage** — zero dashboard test files at this ref (pre-existing gap, not introduced by this PR)

### What was not checked

- Citations verified inline against branch HEAD `dab64df43bd51bd3909044b9117da541589443e9`.
- Stated intent: PR #1599 title+body — spec-compliance assessed.
- Did not run tests locally (CI results used — all green on this ref).
- Did not verify runtime behavior of `transformIndexHtml` across HMR boundary changes.
- The `sessionStorage` stale-token scenario analysis is static; did not verify whether any deployment clears storage on token rotation.

PR #1605, opened 2026-09-09
feat(dashboard): port @-file, model selector, and queue panels to React

griffinwork40:
## Summary

Ports the 3 remaining old-webui panels that had no React equivalents to the new dashboard, closing the feature gap identified in the Phase 5 migration.

### AtFilePopover (`at-file-popover.tsx`)
`@` button in the composer toolbar opens a popover for typing file paths. Inserts `@path` token at the cursor position in the textarea. Path validation matches the old `at-file-panel.ts` gate (absolute, home-relative, explicit-relative, or bare relative paths).

### ModelSelector (`model-selector.tsx`)
Styled `<select>` dropdown fetching from `GET /api/models` with a static fallback (`[sonnet, haiku, opus]`). Includes a `ModelBadge` pill component for use in session cards. Wired into the composer toolbar left of the textarea.

### QueuePanel (`queue-panel.tsx`) + useQueue hook (`use-queue.ts`)
Mid-run message queue with reorder (up/down), inline edit, and remove controls. Key behaviors:
- **One-at-a-time drain** on turn boundaries -- preserves the invariant that entries dwell long enough to be reordered (the whole point of a queue vs a send loop)
- **Enter queues when busy** -- when a turn is running, Enter adds to the queue instead of sending directly
- **Flush on turn boundary** -- `app.tsx` watches the `isBusy` transition and calls `queue.flush()` when a turn completes
- Pure immutable list operations ported from `queue-reorder.ts`

## Wiring

- `composer.tsx`: New toolbar row with @-file button + model selector to the left of the textarea. Queue panel renders above the input when entries exist. `insertAtCursor` helper for @-file token injection.
- `app.tsx`: `useQueue` hook instantiated with session submit function. Flush-on-turn-boundary effect. Model selection state. New props passed to Composer.

## Verification

- `tsc --noEmit` (strict) -- clean
- `pnpm lint` -- clean
- `pnpm build` (Vite bundle) -- succeeds
- `pnpm audit:filesize:check` -- all files under 350-LOC ceiling
- Existing queue-reorder and composer-wiring tests pass (31/31)

## Notes

- **Model selector is UI-only for now** -- selected model is tracked in state but `POST /api/sessions/:id/prompt` doesn't currently accept a model parameter. Wiring into session creation needs a server-side route change.
- **No new component tests** -- the React dashboard has near-zero component test coverage (pre-existing gap). Tests for these 3 panels would be a natural follow-up.

### DIFF
--- a/dashboard/src/app.tsx
+++ b/dashboard/src/app.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { Sidebar } from './components/sidebar';
 import { CommandPalette } from './components/command-palette';
 import { KeyboardShortcuts } from './components/keyboard-shortcuts';
@@ -15,6 +15,8 @@ import { useSessions } from './hooks/use-sessions';
 import { usePendingApprovals } from './hooks/use-pending-approvals';
 import { useTranscript } from './hooks/use-transcript';
 import { useScrollPin } from './hooks/use-scroll-pin';
+import { useQueue } from './hooks/use-queue';
+import { apiFetch } from './lib/api';
 
 type NavItem = 'sessions' | 'memory' | 'schedules' | 'jobs' | 'settings';
 
@@ -50,6 +52,28 @@ function Dashboard() {
   };
 
   const isBusy = status === 'open' || status === 'connecting';
+  const [selectedModel, setSelectedModel] = useState('sonnet');
+
+  const queue = useQueue({
+    submit: async (text: string) => {
+      if (!selectedSessionId) throw new Error('no session');
+      await apiFetch(`/api/sessions/${selectedSessionId}/prompt`, {
+        method: 'POST',
+        body: JSON.stringify({ text }),
+      });
+    },
+    isLive: selectedSession?.mode === 'live',
+    isBusy,
+  });
+
+  // Attempt to flush the queue when a turn completes (isBusy transitions to false).
+  const prevBusyRef = useRef(isBusy);
+  useEffect(() => {
+    if (prevBusyRef.current && !isBusy) {
+      void queue.flush();
+    }
+    prevBusyRef.current = isBusy;
+  }, [isBusy, queue]);
 
   const sidebarContent = (
     <Sidebar
@@ -140,6 +164,9 @@ function Dashboard() {
                 sessionId={selectedSession.id}
                 sessionMode={selectedSession.mode}
                 isBusy={isBusy}
+                queue={queue}
+                onModelSelect={setSelectedModel}
+                currentModel={selectedModel}
               />
             </div>
           )}
--- a/dashboard/src/app.tsx
+++ b/dashboard/src/app.tsx
@@ -38,7 +38,7 @@ function Dashboard() {
   const [mobileSidebarOpen, setMobileSidebarOpen] = useState(false);
 
   const selectedSession = sessions.find((s) => s.id === selectedSessionId);
-  const { items, totals, status } = useTranscript(selectedSessionId);
+  const { items, totals, status, turnActive } = useTranscript(selectedSessionId);
   const { containerRef } = useScrollPin();
 
   const handleNavigate = (nav: string) => {
@@ -51,7 +51,7 @@ function Dashboard() {
     setMobileSidebarOpen(false);
   };
 
-  const isBusy = status === 'open' || status === 'connecting';
+  const isBusy = turnActive;
   const [selectedModel, setSelectedModel] = useState('sonnet');
 
   const queue = useQueue({
@@ -63,17 +63,23 @@ function Dashboard() {
       });
     },
     isLive: selectedSession?.mode === 'live',
-    isBusy,
   });
 
-  // Attempt to flush the queue when a turn completes (isBusy transitions to false).
-  const prevBusyRef = useRef(isBusy);
+  // Clear the queue when the selected session changes to prevent a prompt
+  // queued for session A from being sent to session B.
   useEffect(() => {
-    if (prevBusyRef.current && !isBusy) {
+    queue.clear();
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- queue identity is stable; selectedSessionId is the real dep
+  }, [selectedSessionId]);
+
+  // Flush the queue when a turn completes (totals.turns increments on 'done' records).
+  const prevTurnsRef = useRef(totals.turns);
+  useEffect(() => {
+    if (totals.turns > prevTurnsRef.current) {
       void queue.flush();
     }
-    prevBusyRef.current = isBusy;
-  }, [isBusy, queue]);
+    prevTurnsRef.current = totals.turns;
+  }, [totals.turns, queue]);
 
   const sidebarContent = (
     <Sidebar
--- a/dashboard/src/app.tsx
+++ b/dashboard/src/app.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from 'react';
+import { useEffect, useRef, useState, useCallback } from 'react';
 import { Sidebar } from './components/sidebar';
 import { CommandPalette } from './components/command-palette';
 import { KeyboardShortcuts } from './components/keyboard-shortcuts';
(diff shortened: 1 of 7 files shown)


griffinwork40:
All citations verified. Now Wave 2 -- synthesis. I have all findings from both agents plus the verification manifest. Let me compile the final review.

**Wave 1.5 Citation Verification Manifest:**
- All `file-state` citations verified against `e4bfea1d` -- content matches at stated lines.
- `StrictMode` confirmed at `main.tsx` line 10 (F-1 evidence).
- Absence claim "no test files under dashboard/src" confirmed -- zero results from both `find` and `git ls-tree`.
- `use-transcript.ts:9-13` replay comment verified.
- No citations classified `fabricated` or `false-absent`. All pass.

---

# Review: PR #1605 -- feat(dashboard): port @-file, model selector, and queue panels to React

**Branch:** `feat/dashboard-missing-panels` -> `main`
**Reviewed ref:** `e4bfea1d`
**Change type:** feature | **Lines:** +741/-7 | **Files:** 7

---

## Blocking findings

### 1. `high` · blocking:true · correctness · `dashboard/src/hooks/use-queue.ts:119-143`

**Side effect inside React state updater.** `flush()` executes `flushingRef.current = true` and `void submit(headText)` inside a `setQueue` updater function. React 18 StrictMode (confirmed active in `main.tsx:10`) invokes updaters twice in development. The outer `flushingRef` guard is checked *before* `setQueue`, so both updater invocations fire `submit()`, resulting in double POSTs for the same queue entry in dev.

```tsx
// use-queue.ts:123-143
setQueue((prev) => {
  const head = prev[0];
  if (head === undefined) return prev;
  flushingRef.current = true;          // side effect in updater
  void submit(headText)                // side effect in updater
    .then(() => { setQueue(q => q.filter(...)); })
  return prev;                         // returns unchanged state
});
```

**Suggestion:** Hoist the side effect outside the updater. Use a `useRef` mirror of the queue to read the head synchronously, then call `submit()` directly in the `flush` callback body.

---

### 2. `high` · blocking:true · correctness · `dashboard/src/app.tsx:57-82`

**Cross-session ghost prompt on session switch.** The `submit` function (lines 58-64) closes over `selectedSessionId`. When `flush` is in-flight against session A and the user switches to session B, `clear()` empties the queue but cannot abort the in-flight POST. The request completes and posts the prompt to session A after the user has navigated away -- silently sending a user-authored prompt to the wrong session.

```tsx
// app.tsx:58-64 — closes over selectedSessionId
submit: async (text: string) => {
  if (!selectedSessionId) throw new Error('no session');
  await apiFetch(`/api/sessions/${selectedSessionId}/prompt`, {
    method: 'POST',
    body: JSON.stringify({ text }),
  });
},
```

**Suggestion:** Pass an `AbortController` signal into `apiFetch`. In the `selectedSessionId` effect, abort the controller before calling `queue.clear()`.

---

### 3. `medium` · blocking:true · correctness · `dashboard/src/app.tsx:76-82`

**`prevTurnsRef` never resets on session switch.** When switching from session A (N turns) to session B, `prevTurnsRef.current` retains N. `useTranscript` resets `totals.turns` to 0 on session change, but the flush guard `totals.turns > prevTurnsRef.current` evaluates against the stale high-water mark. Queue items never drain until session B's turn count exceeds session A's.

```tsx
// app.tsx:76-82
const prevTurnsRef = useRef(totals.turns);
useEffect(() => {
  if (totals.turns > prevTurnsRef.current) {   // stale after session switch
    void queue.flush();
  }
  prevTurnsRef.current = totals.turns;
}, [totals.turns, queue]);
```

**Suggestion:** Add `prevTurnsRef.current = 0;` inside the existing `selectedSessionId` effect (line 71).

---

### 4. `medium` · blocking:true · correctness · `dashboard/src/app.tsx:77-82`

**Flush effect re-runs on every render due to unstable `queue` reference.** `useQueue` returns a plain object literal (new reference each render). The flush effect includes `queue` in its dependency array, so it re-runs on every Dashboard re-render. The `prevTurnsRef` guard prevents redundant flushes in practice, but the ref assignment `prevTurnsRef.current = totals.turns` on every render creates a window where a legitimate turn increment is consumed by a spurious run.

**Suggestion:** Replace `queue` in the dep array with `queue.flush` (the stable `useCallback` reference):
```tsx
}, [totals.turns, queue.flush]);
```

---

### 5. `medium` · blocking:true · correctness · `dashboard/src/hooks/use-transcript.ts` + `dashboard/src/app.tsx:76-82`

**Replay SSE frames trigger spurious queue flushes.** Replay frames are processed identically to live frames for `totals.turns` accumulation (confirmed: `use-transcript.ts` lines 9-13). When opening a session with K completed turns, `totals.turns` ramps 0->K during replay, triggering `flush()` on each increment. Any queue item added before replay settles is submitted mid-replay rather than at the true next-turn boundary.

**Suggestion:** Track `liveTurns` separately in `useTranscript` (increment only on non-replay `done` records) and use that as the flush trigger.

---

## Non-blocking findings

### 6. `medium` · blocking:false · correctness · `dashboard/src/components/queue-panel.tsx` + `dashboard/src/hooks/use-queue.ts:137-139`

**Queue error state contract is unimplemented.** The `.catch()` comment says "the UI will show error state," but `QueuePanel` has no error rendering path and `UseQueueResult` exposes no error field. A permanently failing entry blocks every subsequent flush with no user-visible indication.

· waived: bounded, non-data-affecting -- the user retains ability to manually remove the entry; no data loss.

**Suggestion:** Expose an error set from `useQueue` and render a destructive indicator on affected rows.

---

### 7. `medium` · blocking:false · correctness · `dashboard/src/components/model-selector.tsx:67-120`

**Select/state divergence when API returns unknown models.** If `GET /api/models` returns a list without `'sonnet'`, the `<select value={selected}>` has no matching `<option>`. The visible selection and React state diverge silently.

· waived: bounded -- model selector is UI-only (stated intent); no API call uses the stale value today.

**Suggestion:** After fetching, check if `selected` is in the returned list; if not, fire `onSelect(data.models[0].id)`.

---

### 8. `low` · blocking:false · security · `dashboard/src/components/at-file-popover.tsx:25-30`

**Overly permissive path validation.** `looksLikePath`'s catch-all `/^[\w.]/` accepts URLs and single words. No direct security impact today (the token is submitted as plain prompt text, not resolved client-side), and the regex is faithfully ported from the old panel.

**Suggestion:** Tighten to require a path separator in bare paths.

---

### 9. `low` · blocking:false · spec-compliance · `dashboard/src/components/model-selector.tsx` + `dashboard/src/app.tsx:60-63`

**Unmet intent gap has no traceability.** The PR acknowledges "wiring needs server-side change" but the submit call has no TODO or issue reference making the gap searchable.

**Suggestion:** Add `// TODO(#NNNN): pass selectedModel once POST /prompt accepts model param`.

---

### 10. `low` · blocking:false · test-coverage

No test files exist under `dashboard/src/` (confirmed: zero results from `find` and `git ls-tree`). The PR body acknowledges this as a "pre-existing gap." Three non-trivial behaviors are untested: `flush()` drain semantics, `looksLikePath` regex, and the turn-boundary flush trigger.

---

### 11. `nit` · blocking:false · api-compat · `dashboard/src/components/queue-panel.tsx:63`

**Unused `index` prop in `QueueRowProps`.** Declared in the interface but never destructured or used.

**Suggestion:** Remove from `QueueRowProps` and call sites.

---

## Spec-compliance

Stated intent: PR #1605 title + body -- spec-compliance assessed.

- **Unmet intent:** none. All three panels (AtFilePopover, ModelSelector, QueuePanel+useQueue) are implemented. Turn-boundary flush, enter-queues-when-busy, one-at-a-time drain, composer toolbar wiring, and the model-selector-is-UI-only caveat are all present.
- **Scope creep:** none. All changes serve the stated port objective.

---

## What was not checked

- Citations verified inline against branch HEAD `e4bfea1d`.
- Stated intent: PR #1605 title+body -- spec-compliance assessed.
- Did not run tests or the dev server. Did not review Vite/build config changes (none in diff). Did not review the Tailwind theme tokens (`bg-brand/15`, `bg-status-running/15`, etc.) for existence in the theme config.
- Absence claim "no test files under dashboard/src" confirmed via `find` and `git ls-tree` at the reviewed ref.
- Perf-observability: no new issues introduced beyond the re-render cost in finding #4. No instrumentation exists in the dashboard (pre-existing gap).

---

## Decision: DO NOT MERGE -- 2 high, 3 medium blocking (3 correctness); 2 medium waived, 1 low security, 2 low correctness/spec, 1 low test-coverage, 1 nit.

The two high-severity findings (#1 side effects in state updater, #2 cross-session ghost prompt) produce wrong behavior under reachable conditions. The three blocking mediums (#3 stale ref, #4 unstable dep, #5 replay flushes) compound into a queue that silently fails to drain after session switches or on initial load. All five have concrete one-line or few-line fixes that do not require architectural changes.

---

PR #1626, opened 2026-09-10
perf(tui): convert markdown repaint debounce to leading+trailing throttle

griffinwork40:
## Problem

The streaming markdown renderer's `scheduleWithThrottle` was a pure trailing-edge debounce: every token cleared the timer and set a new 33ms timeout. During fast streaming (50+ tokens/sec) the timer reset on every token and **never fired** -- the user saw frozen/laggy text until the model paused for a full 33ms.

This is the primary cause of the "slow/laggy" feel in REPL markdown rendering.

## Fix

Replace the debounce with a **leading + trailing throttle**:

- **Leading edge:** when >= 33ms has elapsed since the last paint, fire immediately via `queueMicrotask` (first token appears with zero delay)
- **Trailing edge:** inside the throttle window, schedule one deferred paint for the remaining time (last token of a burst is always painted)

This caps repaints at ~30 fps while eliminating the initial rendering delay.

## Changes

| File | Change |
|------|--------|
| `markdown-stream-format.ts` | `scheduleWithThrottle` now accepts `lastPaintTime`, returns `{ timer, paintTime }`, implements leading+trailing logic |
| `markdown-stream.ts` | `StreamingMarkdownRenderer` tracks `lastPaintTime` field, wires it through `scheduleRepaint` |
| `markdown-stream-format.test.ts` | 4 new tests: leading edge, trailing edge, timer cleanup, trailing timer completion |

## Before / After

| | Before (debounce) | After (throttle) |
|---|---|---|
| First token delay | Always 33ms (resets on every token) | 0ms (leading edge fires immediately) |
| During fast stream | Timer never fires until model pauses | Paints at ~30fps throughout |
| Final token | 33ms after last token | Remaining time in window (0-33ms) |

## Not addressed (diminishing returns)

Two secondary items were identified but deferred:
1. `Lexer.lex(entirePendingBuffer)` re-parse on every repaint -- O(n) per tick but only within a single paragraph, and measured at <2ms for typical sizes
2. `isInOpenCodeFence` global regex passes per `push()` -- bounded by pending buffer size, touches load-bearing invariants

## Verification

- 190 markdown-stream tests pass (186 existing + 4 new)
- 128 stream-renderer tests pass
- `pnpm lint` clean
- `pnpm audit:filesize:check` clean

### DIFF
--- a/src/cli/markdown-stream-format.test.ts
+++ b/src/cli/markdown-stream-format.test.ts
@@ -138,7 +138,7 @@ describe('formatBlockForCommit blank-line trimming', () => {
 });
 
 describe('scheduleWithThrottle (leading + trailing)', () => {
-  it('fires immediately (leading edge) when enough time has elapsed', () => {
+  it('fires immediately (leading edge) when enough time has elapsed', async () => {
     let fired = false;
     const cb = () => { fired = true; };
     // lastPaintTime far in the past — leading edge should fire
@@ -148,6 +148,9 @@ describe('scheduleWithThrottle (leading + trailing)', () => {
     expect(result.paintTime).toBeGreaterThan(0);
     // Callback is queued via queueMicrotask, not synchronous
     expect(fired).toBe(false);
+    // Drain the microtask queue and verify callback fires
+    await Promise.resolve();
+    expect(fired).toBe(true);
   });
 
   it('defers to trailing edge when inside the throttle window', () => {
--- a/src/cli/markdown-stream-format.test.ts
+++ b/src/cli/markdown-stream-format.test.ts
@@ -4,6 +4,7 @@ import {
   isInOpenCodeFence,
   formatPendingBuffer,
   formatBlockForCommit,
+  scheduleWithThrottle,
 } from './markdown-stream-format.js';
 
 const ANSI_RE = /\x1b\[[0-9;]*m/g;
@@ -133,5 +134,57 @@ describe('formatBlockForCommit blank-line trimming', () => {
     const out = formatBlockForCommit('## Done\n\n', '  ', WIDTH);
     expect(stripAnsi(out).startsWith('\n')).toBe(false);
     expect(stripAnsi(out)).toContain('Done');
+  });
+});
+
+describe('scheduleWithThrottle (leading + trailing)', () => {
+  it('fires immediately (leading edge) when enough time has elapsed', () => {
+    let fired = false;
+    const cb = () => { fired = true; };
+    // lastPaintTime far in the past — leading edge should fire
+    const result = scheduleWithThrottle(cb, 33, null, 0);
+    // Leading edge: no trailing timer, paintTime updated to ~now
+    expect(result.timer).toBeNull();
+    expect(result.paintTime).toBeGreaterThan(0);
+    // Callback is queued via queueMicrotask, not synchronous
+    expect(fired).toBe(false);
+  });
+
+  it('defers to trailing edge when inside the throttle window', () => {
+    let fired = false;
+    const cb = () => { fired = true; };
+    const now = Date.now();
+    // lastPaintTime is very recent — inside the throttle window
+    const result = scheduleWithThrottle(cb, 33, null, now);
+    // Trailing edge: timer is set, paintTime unchanged
+    expect(result.timer).not.toBeNull();
+    expect(result.paintTime).toBe(now);
+    expect(fired).toBe(false);
+    // Clean up the timer
+    if (result.timer) clearTimeout(result.timer);
+  });
+
+  it('clears existing timer when scheduling trailing edge', () => {
+    let count = 0;
+    const cb = () => { count++; };
+    const now = Date.now();
+    // First call: schedule trailing
+    const r1 = scheduleWithThrottle(cb, 33, null, now);
+    // Second call: should clear the first timer
+    const r2 = scheduleWithThrottle(cb, 33, r1.timer, now);
+    expect(r2.timer).not.toBeNull();
+    // Clean up
+    if (r2.timer) clearTimeout(r2.timer);
+  });
+
+  it('trailing timer uses remaining time in the window', async () => {
+    let fired = false;
+    const cb = () => { fired = true; };
+    // lastPaintTime is 20ms ago, throttle is 33ms — remaining ~13ms
+    const result = scheduleWithThrottle(cb, 33, null, Date.now() - 20);
+    expect(result.timer).not.toBeNull();
+    // Wait for the trailing timer to fire
+    await new Promise((resolve) => setTimeout(resolve, 50));
+    expect(fired).toBe(true);
   });
 });
--- a/src/cli/markdown-stream-format.ts
+++ b/src/cli/markdown-stream-format.ts
@@ -326,16 +326,48 @@ export function accumulateCommitted(current: string, trimmed: string): string {
 }
 
 /**
- * Schedule a callback with throttling. Clears any pending timeout and sets
- * a new one for the specified delay. Returns the new timeout ID.
+ * Leading + trailing throttle for streaming repaints.
+ *
+ * Invariant: the FIRST push after idle fires the callback synchronously (via
+ * `queueMicrotask`) so the user sees the first token appear with zero delay.
+ * Subsequent pushes within the same throttle window are coalesced into a
+ * single trailing-edge callback at most `throttleMs` after the leading paint
+ * -- capping repaints at ~30 fps while still showing content the instant it
+ * arrives.
+ *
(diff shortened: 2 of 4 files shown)


griffinwork40:
F3 verified at the call site: `discardPending()` is called on `stream_retry`, after which a fresh token stream arrives. If the retry fires within 33ms, `lastPaintTime` still carries the old value, and the first token hits the trailing-edge path instead of leading-edge, adding up to 33ms delay.

This confirms F3 is a genuine defect: `lastPaintTime` is not reset in `discardPending()`, and the retry path is a production call site where the first-token-delay symptom recurs.

The review is complete. Here is the final report:

---

## Code Review: PR #1626

**`perf(tui): convert markdown repaint debounce to leading+trailing throttle`**
Reviewed ref: `d07065af` | Branch: `afk/improve-markdown-render-speed` → `main` | +98/-5 across 3 files

---

### ❌ DO NOT MERGE

**Decision: DO NOT MERGE -- 1 medium blocking (correctness); 2 medium waived; 5 low; 1 nit.**

---

### Blocking

**F3 · medium · blocking:true · correctness · `src/cli/markdown-stream.ts:424-441, 446-464` · ref:d07065af · file-state · medium confidence**

`discardPending()` and `dispose()` do not reset `lastPaintTime` to 0. After a mid-stream retry (`stream_retry` at `stream-renderer-orchestrator.ts:418`), if the fresh token stream arrives within 33ms, the first token takes the trailing-edge path instead of the leading edge -- reintroducing the exact first-paint lag this PR eliminates.

```ts
// discardPending() clears timer + buffer but leaves lastPaintTime intact
discardPending(): void {
    if (this.throttleTimer) { clearTimeout(this.throttleTimer); this.throttleTimer = null; }
    this.buffer = '';
    // ... overlay clears ...
}
```

**Fix:** Add `this.lastPaintTime = 0;` to both `discardPending()` and `dispose()`.

---

### Non-blocking (waived medium)

**F1 · medium · blocking:false · api-compat · `src/cli/markdown-stream-format.ts:345-350` · diff-context · high confidence**
`scheduleWithThrottle` signature changed from 3 args returning `NodeJS.Timeout` to 4 args returning `{ timer, paintTime }`. The single production importer (`markdown-stream.ts:4`) is updated in this diff; TypeScript catches any missed site.
`waived: bounded to one in-repo caller, already updated; tsc enforces at compile time`

**F5 · medium · blocking:false · spec-compliance · `src/cli/markdown-stream.ts:158-175` · file-state · high confidence**
Resize-triggered `scheduleRepaint()` fires leading-edge on empty buffer, producing a harmless no-op (`repaint()` returns early at line 229). JSDoc invariant overstates the guarantee ("user sees the first token") but no behavioral defect results.
`waived: repaint() early-returns on empty buffer; editorial JSDoc issue, not a defect`

---

### Low (non-blocking)

**F7 · low · blocking:false · perf-observability · `src/cli/markdown-stream.ts:163-167` · diff-context · high confidence**
Two `Date.now()` calls per leading-edge paint (one at schedule-time inside `scheduleWithThrottle`, one in the callback) produce slightly different `lastPaintTime` values. Using `result.paintTime` in the callback instead of a second `Date.now()` eliminates the discrepancy.

**F6 · low · blocking:false · test-coverage · `src/cli/markdown-stream-format.test.ts:141-151` · diff-context · high confidence**
The leading-edge test asserts `fired === false` (confirming microtask deferral) but never drains the queue to verify the callback fires. Add `await Promise.resolve(); expect(fired).toBe(true);`.

**F7b · low · blocking:false · test-coverage · absent · high confidence**
No integration test covers the `discardPending()` -> `push()` first-paint latency path, leaving F3 without a regression guard.

**F8 · low · blocking:false · test-coverage · `src/cli/markdown-stream.test.ts:270-302` · file-state · high confidence**
Existing throttle integration test counts push invocations, not actual repaints, and does not exercise leading-edge behavior.

**F9 · low · blocking:false · correctness · `src/cli/markdown-stream-format.ts:354` · diff-context · medium confidence**
`setTimeout(cb, 0)` is possible at the exact 33ms boundary; unreachable given scheduler jitter.

---

### Nit (non-blocking)

**F10 · nit · blocking:false · perf-observability · `src/cli/markdown-stream-format.ts:367-371` · diff-context · low confidence**
Trailing-edge timer is cleared and rescheduled on every push within the window. Expected behavior; no action needed.

---

### Dropped findings

**F2 (was: high · correctness)** -- DROPPED. Core claim of an async race between `repaint()` settling and `lastPaintTime` update was fabricated. The callback's `this.lastPaintTime = Date.now()` (line 166) executes synchronously before `void this.repaint()` (line 167); any `push()` arriving after the timer fires sees the updated value. The residual timestamp inconsistency is captured by F7.

---

### What was not checked

- Citations verified inline against branch HEAD `d07065af`
- Stated intent: PR #1626 title+body -- spec-compliance assessed
- `repaint()` body read and verified (early-return on empty buffer at line 229 confirms F5 waiver)
- External consumers of `scheduleWithThrottle` not checked (only in-repo importers)
- Timer coarsening in restricted environments not evaluated
- Whether retries arrive >33ms in practice (frequency of F3) not measured -- the path is reachable, fix is one line

**Done**
- What was done: Read-only review of PR #1626 across all dimensions, with citation verification at `d07065af` and fabricated-race finding dropped after inline re-derivation
- Evidence: Review report above; all citations verified against `git show d07065af:<file>`
- What changed: Nothing -- read-only review, no mutations
- Deferred: none

---

griffinwork40:
## Re-Review: PR #1626

**`perf(tui): convert markdown repaint debounce to leading+trailing throttle`**
Reviewed ref: `0fca9c52` | Branch: `afk/improve-markdown-render-speed` -> `main` | +108/-5 across 4 files

This is a re-review after the fix commit addressing the original review's blocking F3 finding.

---

### MERGE

**Decision: MERGE -- 0 blocking (2 low, 2 nit).**

The blocking F3 defect from the prior review is resolved. `discardPending()` and `dispose()` now reset `lastPaintTime = 0`, and the regression test is mutation-verified (removing the reset causes the test to fail).

---

### Non-blocking

**F1-fix · low · blocking:false · test-coverage · `src/cli/markdown-stream.test.ts:323-354` · ref:0fca9c52 · file-state · high confidence**
The F3 regression test was initially vacuous (did not call `push()` after `discardPending()`, so the assertion was trivially true). This was caught during re-review and corrected -- the test now calls `renderer.push('second ')` after `discardPending()` and was mutation-verified: removing `this.lastPaintTime = 0` from `discardPending()` causes the test to fail with a non-null `throttleTimer`.

**F2-fix · low · blocking:false · perf-observability · absent · medium confidence**
No integration test covers burst-steady-state repaint rate cap (N rapid pushes within one throttle window -> exactly 2 callbacks). The property named in the PR description ("capping repaints at ~30 fps") is not mechanically verified. Not blocking -- the unit tests for `scheduleWithThrottle` directly verify the leading/trailing branching logic.

**F3-fix · nit · blocking:false · correctness · `src/cli/markdown-stream.ts:163-174` · ref:0fca9c52 · file-state · high confidence**
`lastPaintTime` is set twice on the leading-edge path: synchronously from `result.paintTime` at line 174, then again in the microtask callback via `Date.now()` at line 166. Both writes are functionally identical (~microsecond difference). The callback write is the semantically correct one (marks actual paint time); the sync write is redundant but harmless.

**F4-fix · nit · blocking:false · api-compat · `src/cli/markdown-stream-format.ts:345-373` · ref:0fca9c52 · diff-context · high confidence**
`scheduleWithThrottle` signature changed from 3 args returning `NodeJS.Timeout` to 4 args returning `{ timer, paintTime }`. TypeScript enforces at compile time; the single production importer (`markdown-stream.ts:4`) is updated in this diff.

---

### Original review findings -- status

| Finding | Status | Evidence |
|---------|--------|----------|
| F3 (blocking: lastPaintTime not reset in discardPending/dispose) | **Fixed** | `lastPaintTime = 0` added to both methods; mutation-verified test |
| F6 (leading-edge test doesn't drain microtask) | **Fixed** | Test is now `async`, drains via `await Promise.resolve()`, asserts `fired === true` |
| F7b (no regression test for discardPending -> push path) | **Fixed** | Integration test added; mutation-verified |
| F7 (two Date.now() calls per leading-edge paint) | Not addressed | Low/nit -- dual-write is harmless |

---

### What was not checked

- Citations verified inline against branch HEAD `0fca9c52`
- Stated intent: PR #1626 title+body -- spec-compliance assessed
- External consumers of `scheduleWithThrottle` not checked (only in-repo importers; tsc enforces)
- Timer coarsening in restricted environments not evaluated

---
_Automated re-review posted by `agent-afk /fix-pr 1626 --re-review`_

PR #1635, opened 2026-09-11
feat(dashboard): overhaul chat UI with topology spine

griffinwork40:
## What

Major overhaul of the web dashboard chat UI to match modern chat aesthetics (inspired by AgentGRAI-web) and port the TUI's topology spine as a distinctive web feature.

### Chat Layout
- Full-width prose column with centered `max-w-4xl` layout
- Right-aligned user messages ("You" label), left-aligned assistant messages ("Agent" label) — no bubbles
- Turn separators between conversation turns
- Hover-revealed message actions (copy for assistant messages)

### Code & Markdown
- Syntax-highlighted code blocks via **shiki** with language label and hover copy button
- Async highlighter singleton — languages lazy-loaded per grammar

### Streaming Animation
- `requestAnimationFrame`-based character-by-character text reveal with blinking cursor
- Code-fence-aware: never splits mid-fence marker
- Only the last assistant message animates during active turns; replayed messages render instantly

### Tool Call Cards (rewrite)
- Category-aware icons (read/write/shell/agent/browser/etc.) with color coding
- CSS grid animated expand/collapse (no framer-motion)
- Shimmer-pulse running state with live elapsed timer
- Tool group wrapper with "Running tool N of M..." header and auto-collapse on completion

### Thinking Panel
- Brain icon with pulse animation during streaming
- Three-dot streaming indicator
- Character count badge
- Truncation at 200 chars with Show more/less toggle

### Topology Spine (new)
Port of the TUI's execution spine to the web:
- Hierarchical tree showing agent dispatches with nested tool calls
- Div-based connecting lines (vertical spine + horizontal stubs) — pure CSS, no SVG
- Category-colored icons matching the TUI palette (cyan=read, amber=write, yellow=shell, purple=agent, etc.)
- Collapsible agent nodes with closer summary lines ("Done (N tools · Xs · $X.XXX)")
- Spine mode enabled by default with fallback to flat view

### SSE Data Enrichment
- `turnCount`, `stopReason`, `parentToolUseId` added to subagent_lifecycle records
- All new fields are optional — zero breaking changes to existing consumers
- Enables richer topology tree building (linking subagents to their dispatch tool)

### Chrome & Polish
- Frosted glass input bar with `backdrop-blur-xl` and gradient fade
- New Session button in header
- Floating "New messages" scroll-to-bottom button
- Removed unused `react-router-dom` dependency

## File inventory

**10 new files** (all under 350 LOC):
- `message-row.tsx` (98) — extracted user/assistant/error/notice/bg-job components
- `code-block.tsx` (119) — shiki-powered syntax highlighting + copy button
- `streaming-text.tsx` (187) — rAF streaming animation + blinking cursor
- `tool-group.tsx` (181) — collapsible tool group wrapper
- `topology-types.ts` (54) — SpineNode + ToolCategory types
- `topology.ts` (315) — tree builder, category classifier, formatters
- `topology-spine.tsx` (61) — top-level spine container
- `topology-node.tsx` (319) — recursive tree node renderer
- `transcript-spine.tsx` (104) — spine-mode transcript grouping
- `message-actions.tsx` (85) — hover-revealed copy action

**14 modified files** across dashboard and agent core.

## Verification

- `pnpm build` ✓
- Web-server tests: 257/257 pass
- Dashboard tests: 23/23 pass
- TypeScript: zero new errors (2 pre-existing in unrelated test/settings files)
- All files under 350-LOC ceiling

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -18,8 +18,8 @@
     "react": "^19.1.0",
     "react-dom": "^19.1.0",
     "react-markdown": "^10.1.0",
-    "react-router-dom": "^7.6.0",
     "remark-gfm": "^4.0.1",
+    "shiki": "^4.4.3",
     "tailwind-merge": "^3.0.2"
   },
   "devDependencies": {
--- a/dashboard/pnpm-lock.yaml
+++ b/dashboard/pnpm-lock.yaml
@@ -35,6 +35,9 @@ importers:
       remark-gfm:
         specifier: ^4.0.1
         version: 4.0.1
+      shiki:
+        specifier: ^4.4.3
+        version: 4.4.3
       tailwind-merge:
         specifier: ^3.0.2
         version: 3.6.0
@@ -678,6 +681,37 @@ packages:
     cpu: [x64]
     os: [win32]
 
+  '@shikijs/core@4.4.3':
+    resolution: {integrity: sha512-QCR4q2ZO/ILJEuwiBMel4wdcTDb1JGwfjKTxPDF6x8ixOaluPrVqIn06C99AcRPhmYlBR56d/Fb+GN58GzExpg==}
+    engines: {node: '>=20'}
+
+  '@shikijs/engine-javascript@4.4.3':
+    resolution: {integrity: sha512-FbOjFJp9VLdo1Wevs10BBtVxiTWwNLqZh5Gkhjgda/ioL15YOgeSl9n+6XMa3qRlPQzfhFNe641SrynFHYG0nQ==}
+    engines: {node: '>=20'}
+
+  '@shikijs/engine-oniguruma@4.4.3':
+    resolution: {integrity: sha512-EcOQkxdxGQrc1Row/cC2c96/v1dbZqGnEVu1qTuT/MJmp6+cXCvQussowVmCv5Tqr3KuY3c7IbM6HTW3LJ1k9w==}
+    engines: {node: '>=20'}
+
+  '@shikijs/langs@4.4.3':
+    resolution: {integrity: sha512-ePic0yfAJGOF83D5wBHK/00EjK65oahBYxFk5epgq33WRv7X9UuxLEV8PtR0szC0z8dl7INIpIodB99JRFlR+A==}
+    engines: {node: '>=20'}
+
+  '@shikijs/primitive@4.4.3':
+    resolution: {integrity: sha512-m0wBeLDQDeIxRdUmrCPdQqfuUamDwRL5isCfYbguKD6NiaKpVbsv+3J81DyIKgNW5h4WAIIr8T4EkgQrBBxvaQ==}
+    engines: {node: '>=20'}
+
+  '@shikijs/themes@4.4.3':
+    resolution: {integrity: sha512-w8UHjeUnIR965KMWJHUPXOc2mNJUnK3vpVLYLvw5IYU2mnTTJ89E24OrJDBNiJDQ0qzb0tc4l7mrIXx5cFeIyw==}
+    engines: {node: '>=20'}
+
+  '@shikijs/types@4.4.3':
+    resolution: {integrity: sha512-UEJxmRR++MAGR6hugn0vgVS2W/6lWAts84FFSrnlH9sP0LNol7E5+NQ792pH8liWUhyMyjhTgSUH3k7iD7tc5g==}
+    engines: {node: '>=20'}
+
+  '@shikijs/vscode-textmate@10.0.2':
+    resolution: {integrity: sha512-83yeghZ2xxin3Nj8z1NMd/NCuca+gsYXswywDy5bHvwlWL8tpTQmzGeUuHd9FC3E/SBEMvzJRwWEOz5gGes9Qg==}
+
   '@tailwindcss/node@4.3.3':
     resolution: {integrity: sha512-/T8IKEsf9VTU6tLjgC7+sv2mOPtQxzE2jMw7u4Tt40Tx+QSZxpzh95/H6cMKoja9XuW7iMdLJYBB0o9G1CaAgg==}
 
@@ -1097,6 +1131,9 @@ packages:
   graceful-fs@4.2.11:
     resolution: {integrity: sha512-RbJ5/jmFcNNCcDV5o9eTnBLJ/HszWV0P73bc+Ff4nS/rJj+YaS6IGyiOL0VoBYX+l1Wrl3k63h/KrH+nhJ0XvQ==}
 
+  hast-util-to-html@9.0.5:
+    resolution: {integrity: sha512-OguPdidb+fbHQSU4Q4ZiLKnzWo8Wwsf5bZfbvu7//a9oTYoqD/fWpe96NuHkoS9h0ccGOTe0C4NGXdtS0iObOw==}
+
   hast-util-to-jsx-runtime@2.3.6:
     resolution: {integrity: sha512-zl6s8LwNyo1P9uw+XJGvZtdFF1GdAkOg8ujOw+4Pyb76874fLps4ueHXDhXWdk6YHQ6OgUtinliG7RsYvCbbBg==}
 
@@ -1109,6 +1146,9 @@ packages:
 
   html-url-attributes@3.0.1:
     resolution: {integrity: sha512-ol6UPyBWqsrO6EJySPz2O7ZSr856WDrEzM5zMqp+FJJLGMW35cLYmmZnl0vztAZxRUoNZJFTCohfjuIJ8I4QBQ==}
+
+  html-void-elements@3.0.0:
+    resolution: {integrity: sha512-bEqo66MRXsUGxWHV5IP0PUiAWwoEjba4VCzg0LjFJBpchPaTfyfCKTG6bc5F8ucKec3q5y6qOdGyYTSBEvhCrg==}
 
   http-proxy-agent@7.0.2:
     resolution: {integrity: sha512-T1gkAiYYDWYx3V5Bmyu7HcfcvL7mUrTWiM6yOfa3PIphViJ/gFPbvidQ+veqSOHci/PxBcDabeUNCzpOODJZig==}
@@ -1426,6 +1466,12 @@ packages:
   nwsapi@2.2.27:
     resolution: {integrity: sha512-gQPNF78qebCQ6tvVFBYrvJdBNOrYZm90ZlXgpIFm06p6qHDHq/XC4TnJftN6OMbxVE0UTBAoRgcsDeJBBooITw==}
 
+  oniguruma-parser@0.12.2:
+    resolution: {integrity: sha512-6HVa5oIrgMC6aA6WF6XyyqbhRPJrKR02L20+2+zpDtO5QAzGHAUGw5TKQvwi5vctNnRHkJYmjAhRVQF2EKdTQw==}
+
+  oniguruma-to-es@4.3.6:
+    resolution: {integrity: sha512-csuQ9x3Yr0cEIs/Zgx/OEt9iBw9vqIunAPQkx19R/fiMq2oGVTgcMqO/V3Ybqefr1TBvosI6jU539ksaBULJyA==}
+
   parse-entities@4.0.2:
     resolution: {integrity: sha512-GG2AQYWoLgL877gQIKeRPGO1xF9+eG1ujIb5soS5gPvLQ1y2o8FL90w2QWNdf9I361Mpp7726c+lj3U0qK1uGw==}
 
@@ -1534,6 +1580,15 @@ packages:
     resolution: {integrity: sha512-6tDA8g98We0zd0GvVeMT9arEOnTw9qM03L9cJXaCjrip1OO764RDBLBfrB4cwzNGDj5OA5ioymC9GkizgWJDUg==}
     engines: {node: '>=8'}
 
+  regex-recursion@6.0.2:
+    resolution: {integrity: sha512-0YCaSCq2VRIebiaUviZNs0cBz1kg5kVS2UKUfNIx8YVs1cN3AV7NTctO5FOKBA+UT2BPJIWZauYHPqJODG50cg==}
+
+  regex-utilities@2.3.0:
+    resolution: {integrity: sha512-8VhliFJAWRaUiVvREIiW2NXXTmHs4vMNnSzuJVhscgmGav3g9VDxLrQndI3dZZVVdp0ZO/5v0xmX516/7M9cng==}
+
+  regex@6.1.0:
+    resolution: {integrity: sha512-6VwtthbV4o/7+OaAF9I5L5V3llLEsoPyq9P1JVXkedTP33c7MfCG0/5NOPcSJn0TzXcG9YUrR0gQSWioew3LDg==}
+
   remark-gfm@4.0.1:
     resolution: {integrity: sha512-1quofZ2RQ9EWdeN34S79+KExV1764+wCUGop5CPL1WGdD0ocPpu91lzPGbwWMECpEpd42kJGQwzRfyov9j4yNg==}
 
@@ -1570,6 +1625,10 @@ packages:
 
   set-cookie-parser@2.7.2:
     resolution: {integrity: sha512-oeM1lpU/UvhTxw+g3cIfxXHyJRc/uidd3yK1P242gzHds0udQBYzs3y8j4gCCW+ZJ7ad0yctld8RYO+bdurlvw==}
+
+  shiki@4.4.3:
+    resolution: {integrity: sha512-Mb/GvXPHBAXdgGIcnfU5L3ldpn1XcxrGkPHwqgRx17/I2XRfqlFKk2vGkHWINn1kdXvzJZeuO3is6I9KLPFm0g==}
+    engines: {node: '>=20'}
 
   siginfo@2.0.0:
     resolution: {integrity: sha512-ybx0WO1/8bSBLEWXZvEd7gMW3Sn3JFlW3TvX1nREbDLRNQNaeNN8WK0meBwPdAaOI7TtRRRJn/Es1zhrrCHu7g==}
@@ -2301,6 +2360,46 @@ snapshots:
   '@rollup/rollup-win32-x64-msvc@4.63.1':
     optional: true
 
+  '@shikijs/core@4.4.3':
+    dependencies:
+      '@shikijs/primitive': 4.4.3
+      '@shikijs/types': 4.4.3
+      '@shikijs/vscode-textmate': 10.0.2
+      '@types/hast': 3.0.5
+      hast-util-to-html: 9.0.5
+
+  '@shikijs/engine-javascript@4.4.3':
+    dependencies:
+      '@shikijs/types': 4.4.3
+      '@shikijs/vscode-textmate': 10.0.2
+      oniguruma-to-es: 4.3.6
+
+  '@shikijs/engine-oniguruma@4.4.3':
+    dependencies:
+      '@shikijs/types': 4.4.3
+      '@shikijs/vscode-textmate': 10.0.2
+
+  '@shikijs/langs@4.4.3':
+    dependencies:
+      '@shikijs/types': 4.4.3
+
+  '@shikijs/primitive@4.4.3':
+    dependencies:
+      '@shikijs/types': 4.4.3
+      '@shikijs/vscode-textmate': 10.0.2
+      '@types/hast': 3.0.5
+
+  '@shikijs/themes@4.4.3':
+    dependencies:
+      '@shikijs/types': 4.4.3
+
+  '@shikijs/types@4.4.3':
+    dependencies:
+      '@shikijs/vscode-textmate': 10.0.2
+      '@types/hast': 3.0.5
+
+  '@shikijs/vscode-textmate@10.0.2': {}
+
   '@tailwindcss/node@4.3.3':
     dependencies:
       '@jridgewell/remapping': 2.3.5
@@ -2702,6 +2801,20 @@ snapshots:
 
   graceful-fs@4.2.11: {}
 
+  hast-util-to-html@9.0.5:
+    dependencies:
+      '@types/hast': 3.0.5
+      '@types/unist': 3.0.3
+      ccount: 2.0.1
+      comma-separated-tokens: 2.0.3
+      hast-util-whitespace: 3.0.0
+      html-void-elements: 3.0.0
+      mdast-util-to-hast: 13.2.1
+      property-information: 7.2.0
+      space-separated-tokens: 2.0.2
+      stringify-entities: 4.0.4
+      zwitch: 2.0.4
+
   hast-util-to-jsx-runtime@2.3.6:
     dependencies:
       '@types/estree': 1.0.9
@@ -2731,6 +2844,8 @@ snapshots:
       whatwg-encoding: 3.1.1
 
   html-url-attributes@3.0.1: {}
+
+  html-void-elements@3.0.0: {}
 
   http-proxy-agent@7.0.2:
     dependencies:
@@ -3231,6 +3346,14 @@ snapshots:
 
   nwsapi@2.2.27: {}
 
+  oniguruma-parser@0.12.2: {}
+
+  oniguruma-to-es@4.3.6:
+    dependencies:
+      oniguruma-parser: 0.12.2
+      regex: 6.1.0
+      regex-recursion: 6.0.2
+
   parse-entities@4.0.2:
     dependencies:
       '@types/unist': 2.0.11
@@ -3344,6 +3467,16 @@ snapshots:
       indent-string: 4.0.0
       strip-indent: 3.0.0
 
+  regex-recursion@6.0.2:
+    dependencies:
+      regex-utilities: 2.3.0
+
+  regex-utilities@2.3.0: {}
+
+  regex@6.1.0:
+    dependencies:
+      regex-utilities: 2.3.0
+
   remark-gfm@4.0.1:
     dependencies:
       '@types/mdast': 4.0.4
@@ -3423,6 +3556,17 @@ snapshots:
   semver@6.3.1: {}
 
   set-cookie-parser@2.7.2: {}
+
+  shiki@4.4.3:
+    dependencies:
+      '@shikijs/core': 4.4.3
+      '@shikijs/engine-javascript': 4.4.3
+      '@shikijs/engine-oniguruma': 4.4.3
+      '@shikijs/langs': 4.4.3
+      '@shikijs/themes': 4.4.3
+      '@shikijs/types': 4.4.3
+      '@shikijs/vscode-textmate': 10.0.2
+      '@types/hast': 3.0.5
 
   siginfo@2.0.0: {}
 
--- a/dashboard/pnpm-lock.yaml
+++ b/dashboard/pnpm-lock.yaml
@@ -29,9 +29,6 @@ importers:
       react-markdown:
         specifier: ^10.1.0
         version: 10.1.0(@types/react@19.2.18)(react@19.2.8)
-      react-router-dom:
-        specifier: ^7.6.0
-        version: 7.18.3(react-dom@19.2.8(react@19.2.8))(react@19.2.8)
       remark-gfm:
         specifier: ^4.0.1
         version: 4.0.1
@@ -1009,10 +1006,6 @@ packages:
   convert-source-map@2.0.0:
     resolution: {integrity: sha512-Kvp459HrV2FEJ1CAsi1Ku+MY3kasH19TFykTz2xWmMeq6bk2NU3XXvfJ+Q61m0xktWwt+1HSYf3JZsTms3aRJg==}
 
-  cookie@1.1.1:
-    resolution: {integrity: sha512-ei8Aos7ja0weRpFzJnEA9UHJ/7XQmqglbRwnf2ATjcB9Wq874VKH9kfjjirM6UhU2/E5fFYadylyhFldcqSidQ==}
-    engines: {node: '>=18'}
-
   css.escape@1.5.1:
     resolution: {integrity: sha512-YUifsXXuknHlUsmlgyY0PKzgPOr7/FjCePfHNt0jxm83wHZi44VDMQ7/fGNkjY3/jV1MC+1CmZbaHzugyeRtpg==}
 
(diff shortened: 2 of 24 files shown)


griffinwork40 (commented):

---

**PR #1635 — feat(dashboard): overhaul chat UI with topology spine**
**Reviewed ref:** `cbe9a1bdc8980ff9dc99c578a1aa42abf1083e0e` | **Round:** 1 of 2 max

---

## Prior automated rounds

None — this is Round 1.

---

## Decision: ❌ DO NOT MERGE

**2 high (blocking) · 4 medium advisory (non-blocking) · 5 low + 1 nit advisory**

**Severity floor applied:** only critical/high block. All medium and below findings are advisory; follow-up issues are recommended over fix commits. 2 high → DO NOT MERGE.

Two blocking findings: the topology feature this PR introduces never works (H1), and the dashboard build is broken across three CI jobs (H2).

---

## 🔴 High — Blocking

### H1 — Topology tree is always flat: `parentToolUseId` receives the wrong ID

**File:** `src/agent/subagent.ts` · **Confidence:** high · **Dimension:** correctness

`parentToolUseId` is assigned `options.parentId`, which is the **parent subagent's session ID**, not the `tool_use_id` of the spawning `agent` tool-call block.

The topology builder in `dashboard/src/lib/topology.ts` resolves parent nodes via:
```ts
const parentTool = toolByUseId.get(s.parentToolUseId);
```
where `toolByUseId` is keyed by `tool.toolUseId` — the ID of the `agent` tool-use block emitted in the transcript. Because `options.parentId` is never a key in that map, every parent lookup silently returns `undefined`, every subagent is orphaned, and the spine always renders as a flat list. The PR's primary feature — hierarchical multi-level delegation display — never works.

**Fix:** Thread the actual tool-call `id` from the `agent` tool-use block through the dispatch path as `options.parentToolUseId`; do not reuse `options.parentId`.

---

### H2 — Build broken: `@/lib/ledger-adapter` does not exist

**File:** `dashboard/src/lib/topology.ts` · **Confidence:** high · **Dimension:** correctness

`topology.ts` contains:
```ts
import type { TranscriptItem, ToolCallItem, SubagentItem } from '@/lib/ledger-adapter';
```
The `@/` alias in the dashboard's Vite config maps to `dashboard/src/`. No file `dashboard/src/lib/ledger-adapter.ts` is created in this PR. The real `ledger-adapter.ts` lives at `src/web-server/shared/ledger-adapter.ts` — a different package.

**CI impact:** `Lint & Build` ❌ · `Test (dashboard)` ❌ · `Publish Bundle Smoke` ❌

The PR body claims "Dashboard tests: 23/23 pass" — CI directly contradicts this.

**Fix (choose one):**
- Create `dashboard/src/lib/ledger-adapter.ts` that re-exports the needed types from the shared package.
- Correct the import to a valid relative path accessible from the dashboard bundle context.

---

## 🟡 Medium — Advisory (non-blocking; suggest follow-up issues)

**M1 — `dangerouslySetInnerHTML` on shiki output without sanitization**
`dashboard/src/components/code-block.tsx` · security
Shiki v4 HTML-encodes token text, so there is no currently exploitable path. However, the raw HTML string bypasses React's XSS protections entirely with no library version pin or CI assertion of HTML-safety. A future shiki regression silently introduces stored XSS for all dashboard users.
*Suggestion:* Add DOMPurify sanitization, or pin the shiki version and add a CI test asserting `codeToHtml('<script>alert(1)</script>')` produces escaped output.

**M2 — Stale `speed` prop in streaming rAF loop**
`dashboard/src/components/streaming-text.tsx` · correctness
`startLoop` captures `speed` at creation time; `speed` is omitted from the `useEffect` dep array (suppressed via `eslint-disable`). If `speed` changes mid-stream, the loop ignores the change. The `text` closure is mitigated by cleanup+restart; `speed` is not.
*Suggestion:* Store `speed` in a ref (`speedRef.current = speed`) updated on every render; read `speedRef.current` inside `tick`. Remove the `eslint-disable`.

**M3 — `setTimeout` not cleaned up on unmount in `message-actions.tsx`**
`dashboard/src/components/message-actions.tsx` · correctness
`setTimeout(() => setCopied(false), 2000)` is not stored or cancelled on unmount. The sibling `code-block.tsx` implements the correct ref+cleanup pattern.
*Suggestion:* Store the timer ID in a ref; cancel it in a `useEffect` cleanup.

**M4 — `buildSpineTree` root ordering does not reflect SSE arrival order**
`dashboard/src/lib/topology.ts` · correctness
Agent roots are accumulated in Pass 3, tool roots in Pass 5, with no merge-by-sequence step. Root-level items that interleave chronologically in SSE render with all agents before all tools.
*Suggestion:* Assign a sequence index to each `SpineNode` at construction; sort `roots` by that index after all passes complete.

---

## ⚪ Low / Nit — Advisory (follow-up issues only)

| ID | Finding | File | Dimension |
|----|---------|------|-----------|
| L1 | `stopReason` string leaks internal agent state over SSE | SSE emitter | security |
| L2 | `clipboard.writeText` has no length cap or error surfacing | message-actions.tsx | security |
| L3 | Unreachable `undefined` guard after bounds-checked loop | transcript-spine.tsx | correctness |
| L4 | `eslint-disable` suppresses exhaustive-deps in streaming-text | streaming-text.tsx | spec-compliance |
| L5 | `buildSpineTree` has no tests for parent-child relationship or out-of-order SSE | topology.ts | test-coverage |
| L6 | `turnCount?: number` permits `NaN`/`Infinity` (serializes to `null` via JSON) | type definitions | api-compat |

---

## 📋 Spec-compliance

| Claim (PR body) | Status | Notes |
|---|---|---|
| "Zero breaking changes to existing consumers" | ⚠️ Partial | Wire format unchanged; `parentToolUseId` value semantics incorrect (H1) |
| "Dashboard tests: 23/23 pass" | ❌ False | `Test (dashboard)` CI RED |
| "Spine mode enabled by default with fallback to flat view" | ❌ Broken | Always flat due to H1 — fallback only, no hierarchy ever shown |
| "rAF loop properly cleaned up on unmount" | ⚠️ Partial | Cleanup exists for `text` changes; `speed` stale closure unmitigated (M2) |

---

## 🔍 What was not checked

- Citations verified against diff content at `cbe9a1bdc8980ff9dc99c578a1aa42abf1083e0e` (branch not checked out locally).
- `vite.config.ts` / `tsconfig.json` not directly inspected — H2 inferred from import path + absence of file in diff + three CI failures.
- Sidebar, command palette, keyboard shortcuts, memory view not reviewed.
- `fenceSafeLength` correctness not verified (code section not provided).
- Shiki v4.4.3 HTML escaping not exhaustively tested.
- Stated intent: PR #1635 title+body — spec-compliance assessed.

*One finding was dropped after Wave 1.5 verification: the `scrollToBottom` security finding was refuted — `scrollToBottom` is gated behind an explicit user-initiated button click (the "New messages" button), not called from the rAF loop or on each SSE message. The removal of the `isPinned` guard is intentional per the updated docstring.*

griffinwork40 (commented):

---

**PR #1635 — feat(dashboard): overhaul chat UI with topology spine**
**Reviewed ref:** `cbe9a1bdc8` | **Regime:** full (feature, ~2450 LOC)

---

## Prior automated rounds

None — this is Round 1.

---

## Decision: DO NOT MERGE

**1 high (blocking); 3 medium advisory (non-blocking); 1 low; 1 nit.**

Severity floor: only critical/high block. The 3 mediums below are advisory follow-ups, not merge-blockers.

Arithmetic: 1 high → DO NOT MERGE. Once the lockfile inconsistency is fixed (one commit), the 3 mediums are follow-up candidates.

---

## 🔴 Blocking Finding

### F-01 · high · correctness · blocking:true · confidence:high · ref:cbe9a1bdc8 · citation-type:file-state + CI log

**`dashboard/pnpm-lock.yaml` lockfile still references removed `react-router-dom` — CI fails on all 3 dashboard jobs**

The PR removed `react-router-dom` from `dashboard/package.json` but the committed `pnpm-lock.yaml` was not regenerated cleanly. The lockfile still carries `react-router-dom` in its specifiers list (3 occurrences found via grep). CI's `pnpm install --frozen-lockfile` sees a mismatch between lock specifiers and package.json and fails with `ERR_PNPM_OUTDATED_LOCKFILE`. This blocks **Lint & Build**, **Test (dashboard)**, and **Publish Bundle Smoke** — 3 of 8 CI jobs fail.

**Evidence:**
- CI log (job 103414052980, Lint & Build): `specifiers in the lockfile ({"react-router-dom":"^7.6.0",...}) don't match specs in package.json (...)`
- `grep 'react-router-dom' dashboard/pnpm-lock.yaml` at reviewed ref returns 3 hits; `grep 'react-router-dom' dashboard/package.json` returns 0 hits.

**Fix:** Run `cd dashboard && pnpm install` to regenerate the lockfile without `react-router-dom`, then commit the updated `pnpm-lock.yaml`.

---

## 🟡 Advisory Findings (non-blocking — suggest follow-up issues)

### F-02 · medium · correctness · blocking:false · waived: bounded defect affecting only collapsed agent display text, no data loss or wrong output to end users

**`dashboard/src/components/topology-node.tsx:307-316` — CloserLine renders twice for collapsed agents with children**

When an agent node is collapsed (`!expanded`) and done (`status !== 'running'`), CloserLine renders both from the "expanded children" block at line 307 (`isAgent && node.status !== 'running'` inside the `expanded` guard — which is NOT rendered because `expanded` is false, so this is actually fine) AND from the "collapsed agents" block at line 314 (`isAgent && (!hasChildren || !expanded) && node.status !== 'running'`). On re-examination, the expanded-children CloserLine is guarded by `hasChildren && expanded`, so when collapsed it is gated off. However, the bottom guard `(!hasChildren || !expanded)` is true when `hasChildren && !expanded`, producing one CloserLine — and ALSO true when `!hasChildren`, producing a second from `!hasChildren ? <CloserLine> : null`. The ternary at line 315 prevents the double-render for the `hasChildren && !expanded` case (returns null). No double render occurs — this is actually correct after closer reading. **WITHDRAWN — logic is correct.**

### F-02 (revised) · medium · test-coverage · blocking:false · waived: new UI components follow existing test patterns in the dashboard (23/23 pass), and the topology tree builder is pure logic suitable for follow-up unit tests

**Zero tests for 10 new dashboard components and the topology tree builder**

The PR adds 10 new component files and 2 new library files (`topology.ts`, `topology-types.ts`) with zero accompanying tests. `grep -r 'topology\|TopologyNode\|buildSpineTree\|StreamingText\|ToolGroup' dashboard/src/**/*.test.*` returns 0 matches. The `buildSpineTree` function in `topology.ts` (315 lines, pure function, complex multi-pass algorithm) is the strongest candidate for unit tests: it has 5 passes with orphan re-parenting and multiple code paths.

**Suggested follow-up:** Add `topology.test.ts` covering the tree builder (flat items, nested agent→tool, out-of-order orphan re-parenting, parentToolUseId linking).

---

### F-03 · medium · correctness · blocking:false · waived: bounded visual-only defect, no data corruption; streaming animation is cosmetic

**`dashboard/src/components/streaming-text.tsx:120-130` — `startLoop` captures stale `text` and `isStreaming` props in its closure**

`startLoop` is wrapped in `useCallback` with `[speed, text, isStreaming]` deps. Inside `startLoop`, `tick()` reads `text.length` via closure. When `text` grows (new streaming tokens arrive), `startLoop` is recreated — but if the rAF loop from the previous `startLoop` instance is still running, it continues to read the OLD closure's `text.length`. The effect cleanup cancels the old rAF, but there is a race window: the effect fires `startLoop()` before the cleanup of the previous invocation runs (React effect cleanup order). In practice this produces a brief visual stutter where the animation catches up to the old text length and pauses for one frame before the new loop starts.

**Suggested follow-up:** Store `text` in a ref (`textRef.current = text`) and read `textRef.current.length` inside `tick`, removing the stale-closure race.

---

### F-04 · medium · spec-compliance · blocking:false · waived: pre-existing on main; the root-level lockfile issue is not introduced by this PR

**Root-level `pnpm-lock.yaml` also stale (Publish Bundle Smoke failure root cause overlaps with F-01)**

The "Publish Bundle Smoke" CI job fails for the same lockfile reason. The root cause is entirely in the dashboard lockfile (F-01), not a separate issue. This finding is a duplicate of F-01 for a different CI job.

**WITHDRAWN — same root cause as F-01.**

---

## Non-Blocking Findings

### F-05 · low · perf-observability · blocking:false · ref:cbe9a1bdc8 · citation-type:file-state

**`dashboard/src/components/code-block.tsx:12-22` — shiki highlighter eagerly loads 21 language grammars**

`getHighlighter()` loads all 21 grammars at once on the first code block render. shiki grammars are non-trivial WASM/JS bundles. For users viewing sessions with no code blocks, this is never triggered (lazy). For those that do, the initial load adds ~200-500ms of JS parse time depending on device. The PR description says "languages lazy-loaded per grammar" which is inaccurate — they are batch-loaded on first use.

**Suggested follow-up:** Use shiki's `loadLanguage()` API to load grammars on-demand per code block.

---

### F-06 · nit · blocking:false

**`dashboard/src/components/message-actions.tsx:40-44` — UserMessage rows render an empty action bar (no visible buttons)**

`MessageActions` renders the outer container + border for `kind='user'` but the only button inside is gated on `kind === 'assistant'`. The result is an invisible empty bordered `<div>` that appears on hover over user messages. Harmless but produces a tiny ghost element on hover.

---

## Spec-Compliance

|---|---|
| Full-width prose column with centered max-w-4xl | **Met** (transcript-view.tsx max-w-4xl mx-auto) |
| Right-aligned user / left-aligned assistant | **Met** (message-row.tsx) |
| Syntax highlighting via shiki + copy button | **Met** (code-block.tsx) |
| Streaming text animation (rAF, fence-aware) | **Met** (streaming-text.tsx) |
| Tool call cards with category icons + collapse | **Met** (tool-call-card.tsx) |
| Thinking panel with Brain icon + streaming dots | **Met** (thinking-panel.tsx) |
| Topology spine with tree connectors | **Met** (topology-node/spine/transcript-spine) |
| SSE enrichment (turnCount, stopReason, parentToolUseId) | **Met** (session-types, ledger-project, subagent, ledger-adapter) |
| Frosted glass input bar | **Met** (composer.tsx backdrop-blur-xl) |
| New Session button | **Met** (app.tsx Plus button) |
| Scroll-to-bottom button | **Met** (use-scroll-pin.ts + app.tsx) |
| Removed react-router-dom | **Partial** — removed from package.json but lockfile not updated (F-01) |
| All files under 350 LOC | **Met** (topology-node.tsx=319, topology.ts=315) |

## What was not checked

- Citations verified inline against branch HEAD `cbe9a1bdc8`.
- CI failure root cause verified against CI job log 103414052980.
- Stated intent: PR #1635 title+body — spec-compliance assessed.
- Did not verify streaming animation behavior at runtime (read-only review).
- Did not verify topology spine rendering correctness visually.
- Did not verify shiki output sanitization beyond confirming it goes through `codeToHtml` (library-level trust).
- Did not verify SSE backward compatibility with older dashboard versions consuming the stream.

griffinwork40:
Good. Wave 1.5 verification results:

**Check A (Citation verification):**
- **F-01** `tool-group.tsx:175` — **verified**. The pre-fetched file content confirms `{tools.map((tool, i) => (<ToolCallCard key={i} {...tool} />))}` at line 175.
- **F-02** `use-scroll-pin.ts` + `app.tsx` — **verified**. `scrollToBottom` is only used on line 183 as an `onClick` handler. No auto-scroll `useEffect` exists. The hook's contract comment says "Call scrollToBottom() whenever new items are added" but no caller does so.
- **F-04** `code-block.tsx:46-47` — **verified**. The `'text'` fallback is confirmed at the reviewed ref.
- **F-05** `topology.ts:271-272` — **verified**. The `find()` inside the orphan loop is confirmed.

**Check B (Absence-claim verification):**
- **F-07** claims no tests exercise `parentToolUseId`/`turnCount`/`stopReason` in the ledger-adapter. `git grep` returned zero matches in `src/web-server/**/*.test.*` and `dashboard/**/*.test.*`. The `background-registry.test.ts` matches are unrelated (different consumer). **Confirmed-absent** in the ledger-adapter test suite.
- **F-07** also claims no test for `buildSpineTree`. `git grep` returned zero matches for `buildSpineTree` in test files. **Confirmed-absent**.

No findings were fabricated or false-absent. All citations verified. No downgrades needed.

Now let me run Wave 2 synthesis. The findings are clean enough to synthesize inline.

---

# PR #1635 Review: feat(dashboard): overhaul chat UI with topology spine

**Target:** PR #1635 `afk/improve-agent-chat-ui` -> `main` | +2252/-264 | 24 files
**Reviewed ref:** `074a4394bae4bd1a262e23d7c261b34f74ae3cce`
**Stated intent:** PR title + body -- spec-compliance assessed.
**Regime:** full

---

## Blocking findings

### 1. `medium` · `blocking:true` · high · CORRECTNESS · `dashboard/src/components/tool-group.tsx:175` · ref:074a4394 · diff-context

**ToolCallCard instances keyed by array index -- state leaks on reorder.**

`tools.map((tool, i) => <ToolCallCard key={i} {...tool} />)` uses the array index as the React key. `ToolCallCard` holds local state (`outputOpen`, `inputExpanded`, `liveSeconds`). When tools arrive in a parallel wave and the array reorders, React reuses DOM nodes with wrong state -- a user who opened the output pane of tool #2 sees it applied to tool #1 after insertion. The `liveSeconds` interval timer can also show elapsed time on a finished tool.

```tsx
{tools.map((tool, i) => (
  <ToolCallCard key={i} {...tool} />   // index key — state leaks on reorder
))}
```

**Suggestion:** Key by a stable identity. Add `toolUseId` (or `name + inputPreview` hash) to `ToolCallCardProps` and use it:
```tsx
<ToolCallCard key={tool.toolUseId ?? `${tool.name}-${i}`} {...tool} />
```

---

### 2. `medium` · `blocking:true` · high · CORRECTNESS · `dashboard/src/hooks/use-scroll-pin.ts` + `dashboard/src/app.tsx:44,183` · ref:074a4394 · file-state

**Auto-scroll-while-pinned is not wired -- manual button is the only scroll path.**

The hook's contract says *"Call scrollToBottom() whenever new items are added -- the hook decides whether to act based on the current pin state."* But `app.tsx` never calls `scrollToBottom()` on new item arrival. The only caller is the "New messages" button click handler (line 183). Users watching a live session transcript never auto-scroll, even when at the bottom.

```ts
// app.tsx — the only uses of scrollToBottom:
const { containerRef, scrollToBottom, showScrollButton } = useScrollPin(); // line 44
onClick={scrollToBottom}  // line 183 — button only
```

**Suggestion:** Add a `useEffect` that auto-scrolls when items change and the user is pinned:
```ts
useEffect(() => {
  if (isPinned) scrollToBottom();
}, [items.length, isPinned, scrollToBottom]);
```

---

## Non-blocking findings

### 3. `low` · `blocking:false` · high · SECURITY · `dashboard/src/components/code-block.tsx:104-111` · ref:074a4394 · diff-context

`dangerouslySetInnerHTML` receives shiki output -- safe in practice (shiki's `hast-util-to-html` encodes `<` and `&` via `stringifyEntities`), but the safety proof is entirely implicit. The inline comment asserts it without naming the mechanism.

**Suggestion:** Add a comment naming the mechanism: `// hast-util-to-html stringifyEntities encodes < and & in all text nodes`.

### 4. `low` · `blocking:false` · high · CORRECTNESS · `dashboard/src/components/streaming-text.tsx:95` · ref:074a4394 · diff-context

`completedRef` is never reset between messages. Safe today because `transcript-view.tsx` keys each message by `item.id` (guaranteeing a fresh mount), but this is an implicit architectural constraint with no in-component guard.

**Suggestion:** Document the key-per-message requirement in the component JSDoc as a load-bearing invariant.

### 5. `low` · `blocking:false` · high · PERF-OBSERVABILITY · `dashboard/src/components/code-block.tsx:46-47` · ref:074a4394 · diff-context

`'text'` fallback is not a valid shiki language. `codeToHtml` throws, caught silently -- correct behavior (falls back to `<pre>`), but swallows errors with no log. Replace with `'plaintext'` (valid in shiki) or remove the lang param.

### 6. `low` · `blocking:false` · medium · PERF-OBSERVABILITY · `dashboard/src/lib/topology.ts:271-272` · ref:074a4394 · diff-context

O(n^2) orphan re-parent pass: `subagentItems.find()` inside a loop. Negligible for typical sessions (tens of subagents), but fixable with a Map lookup in two lines.

### 7. `low` · `blocking:false` · high · TEST-COVERAGE · ref:074a4394 · confirmed-absent

No tests exercise the three new SSE fields (`turnCount`, `stopReason`, `parentToolUseId`) in the ledger-adapter, and no test covers `buildSpineTree` topology linking -- the primary correctness invariant of the new feature. (Search pattern: `git grep 'parentToolUseId|turnCount' -- '*.test.*'` in `src/web-server/` and `dashboard/` returned zero matches.)

### 8. `nit` · `blocking:false` · high · CORRECTNESS · `dashboard/src/components/tool-call-card.tsx:100` + `dashboard/src/components/tool-group.tsx:28` · ref:074a4394 · diff-context

Duplicate `Collapse` component defined identically in both files. Extract to a shared module.

### 9. `nit` · `blocking:false` · high · SPEC-COMPLIANCE · `dashboard/src/components/code-block.tsx:8-25` · ref:074a4394 · diff-context

Shiki highlighter eagerly loads 21 languages at first `CodeBlock` mount (~600KB-1MB of grammar data behind a dynamic import). Not a violation (stated intent is met), but an observability gap. On-demand language loading would reduce per-user download by ~70%.

---

## Spec-compliance

Stated intent: PR title + body. All claimed features are implemented. No unmet requirements. No scope creep identified.

---

## What was not checked

- Citations verified inline against branch HEAD `074a4394bae4bd1a262e23d7c261b34f74ae3cce`.
- Stated intent: PR #1635 title+body -- spec-compliance assessed.
- Did not run tests or build (read-only review).
- Did not visually verify CSS/layout rendering.
- Did not review the `pnpm-lock.yaml` diff (auto-generated).
- Shadow-verify not invoked: no critical/high findings; both blocking mediums are at their default disposition (no overrides).

---

## Decision: DO NOT MERGE -- 2 medium blocking (1 correctness: index key, 1 correctness: auto-scroll not wired); 7 non-blocking (1 low security, 2 low correctness, 2 low perf, 1 low test-coverage, 2 nit).

**Done**
- What was done: Full-regime code review of PR #1635 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability dimensions.
- Evidence: Review output above with 9 findings, 2 blocking (F-01: index key in tool-group.tsx:175, F-02: auto-scroll not wired in app.tsx). All citations verified against ref 074a4394.
- Deferred: none.

---

griffinwork40:
Good. `call.id` is the tool_use block's `id` field from the Anthropic API, which is always a `toolu_*` string. The finding about dual-purpose is about type-system fragility, not a live bug. Let me now compile the synthesis.

Based on all evidence gathered, here is the complete review synthesis:

---

## Code Review: PR #1635 — feat(dashboard): overhaul chat UI with topology spine

**Target:** `afk/improve-agent-chat-ui` → `main` | **Ref:** `168f87c8` | **Stats:** +2258 / -264 / 24 files | **Regime:** full

---

### Findings

**1. `medium` · blocking:true · confidence:high · correctness · `dashboard/src/components/streaming-text.tsx:171-175` · ref:168f87c8 · file-state**

`lastTimestampRef` is not reset in the useEffect cleanup. When the effect re-runs (new SSE token), `cancelAnimationFrame` fires and `rafRef.current` is set to null, but `lastTimestampRef.current` retains the timestamp from the last completed tick. The next `startLoop()` call reads the stale timestamp on its first frame, computing `elapsed` as the entire wall-clock gap since the last tick rather than the usual ~16ms. Under fast token bursts this causes multi-character jumps in the animation.

```tsx
// Current cleanup (line 171-175):
return () => {
  if (rafRef.current !== null) {
    cancelAnimationFrame(rafRef.current);
    rafRef.current = null;
  }
};
```

**Suggestion:** Add `lastTimestampRef.current = null;` after `rafRef.current = null;` so each restarted loop treats its first frame as fresh via the `?? timestamp` fallback at line 116.

---

**2. `medium` · blocking:false · confidence:high · security · `dashboard/src/components/streaming-text.tsx:21-26` · ref:168f87c8 · file-state** · waived: bounded visual defect (cursor blink) with no data impact; author can land and follow up

Inline `<style>` tag injected from a static string constant in `StreamingCursor`. Not a runtime XSS risk (the string is a literal), but incompatible with any future CSP `style-src` policy that omits `unsafe-inline`. The tag is also re-injected on every render of `StreamingCursor`.

```tsx
const BLINK_STYLE = `@keyframes blink-cursor { ... }`;
// rendered as: <style>{BLINK_STYLE}</style>
```

**Suggestion:** Move the `blink-cursor` keyframe into `globals.css` alongside the existing custom animations (`shimmer-pulse`, `spin-slow`, `fade-in`). The `.streaming-cursor` class reference stays. This eliminates the inline style tag and follows the established project pattern.

---

**3. `low` · blocking:false · confidence:high · api-compat · `src/agent/subagent.ts:598` · ref:168f87c8 · file-state**

`options.parentId` is used as `parentToolUseId` in the SSE lifecycle emit. The `ForkSubagentOptions.parentId` doc says it "overrides the default of `parent.sessionId`" -- a rendering hint with session-ID semantics. At all three current dispatch sites (`subagent-executor.ts:810`, `compose-executor.ts:761`, `skill-executor/fork-dispatch.ts:190`), `call.id` is a `toolu_*` tool_use_id, so the emitted value is currently correct. But the type-system contract is fragile: a future caller passing a session UUID as `parentId` would silently corrupt topology links.

**Suggestion:** Add a JSDoc annotation on `parentId` in `ForkSubagentOptions` clarifying that the value MUST be a `tool_use_id` when present, since it is now emitted as `parentToolUseId` on the SSE channel. Alternatively, introduce a separate `parentToolUseId` field with explicit semantics.

---

**4. `low` · blocking:false · confidence:high · api-compat · `dashboard/src/components/transcript-view.tsx:19-54` · ref:168f87c8 · file-state**

`TranscriptItem` is defined locally in `transcript-view.tsx` (line 19) while `dashboard/src/lib/ledger-adapter.ts` re-exports the canonical type from `src/web-server/shared/ledger-adapter.ts`. The module comment in `ledger-adapter.ts` line 7 states the re-export is "the single source of truth." Two structurally identical but independently maintained type declarations will drift.

**Suggestion:** Import `TranscriptItem` from `@/lib/ledger-adapter` in `transcript-view.tsx` and remove the local definition.

---

**5. `low` · blocking:false · confidence:high · test-coverage · `dashboard/src/lib/topology.ts` · ref:168f87c8 · file-state**

`topology.ts` (315 LOC) contains the tree-building algorithm with multi-pass orphan handling, re-parenting, and category classification. No test file exists for it. The dashboard test suite has only 3 test files (`schedules-view.test.tsx`, `markdown-content.test.tsx`, `use-subagent-tree.test.ts`), none of which cover `buildSpineTree`, `categorizeToolName`, or `formatSpineDuration`. The tree builder has non-trivial edge cases (out-of-order SSE, orphan re-parenting, claimed-tool dedup).

**Suggestion:** Add `dashboard/src/lib/topology.test.ts` covering at minimum: empty input, tool-only input, subagent with `parentToolUseId` linking, orphan re-parenting, and root sort order.

---

**6. `low` · blocking:false · confidence:high · security · `dashboard/src/components/code-block.tsx:107-111` · ref:168f87c8 · file-state**

`dangerouslySetInnerHTML` with shiki's `codeToHtml` output. Shiki 4.x escapes content via `hast-util-to-html` / `stringify-entities` (encodes `<`, `>`, `&`, `"` in text nodes and attributes). The inline comment documents the trust model. This is safe under the current trust boundary (code comes from the session ledger, not from arbitrary external input) but carries a supply-chain risk.

**Suggestion:** No code change needed. The existing trust-model comment is adequate. For defense in depth, a CSP header on the dashboard origin (a server config change) would eliminate the class of DOM-XSS even if shiki were compromised.

---

**7. `nit` · blocking:false · confidence:high · api-compat · `src/web-server/shared/ledger-adapter.ts:318-330` · ref:168f87c8 · file-state**

The subagent patch block patches `turnCount` and `stopReason` from terminal lifecycle events but does not patch `parentToolUseId`. This is correct by spec ("present on started events only"), but the intentional omission is not documented.

**Suggestion:** Add a one-line comment: `// parentToolUseId intentionally not patched -- set only on started event`.

---

Stated intent: PR #123 title+body. Spec-compliance assessed.

The diff satisfies all stated requirements: full-width chat layout, streaming animation, shiki code blocks, tool call card rewrite with categories, thinking panel with brain icon, topology spine port, SSE enrichment with `turnCount`/`stopReason`/`parentToolUseId`, frosted glass composer, new session button, scroll-to-bottom button, and removal of unused `react-router-dom`. No unmet requirements or scope creep detected.

---

### Summary

| # | severity | blocking | dimension | short |
|---|---|---|---|---|
| 1 | medium | **true** | correctness | `lastTimestampRef` not reset in streaming-text cleanup -- animation jumps under fast SSE |
| 2 | medium | false | security | Inline `<style>` tag in StreamingCursor -- CSP-incompatible, static string, no runtime XSS |
| 3 | low | false | api-compat | `parentId` used as `parentToolUseId` -- correct today, fragile contract |
| 4 | low | false | api-compat | Local `TranscriptItem` type duplicates canonical shared type |
| 5 | low | false | test-coverage | `topology.ts` (315 LOC, complex tree builder) has no tests |
| 6 | low | false | security | `dangerouslySetInnerHTML` with shiki -- safe, supply-chain risk only |
| 7 | nit | false | api-compat | Missing comment on intentional `parentToolUseId` non-patch |

---

### What was not checked

- Citations verified against branch HEAD `168f87c8`.
- Stated intent: PR #1635 title+body. Spec-compliance assessed.
- Did not run the dashboard dev server or verify visual rendering.
- Did not run tests (`pnpm test` output from PR description accepted at face value).
- `pnpm-lock.yaml` diff skipped (auto-generated).
- Did not audit shiki's internal escaping implementation (relied on documented behavior of `hast-util-to-html`).
- The correctness agent was truncated mid-stream; additional correctness findings about `skipAnimation` render-phase ref reads and topology edge cases were not fully captured. A re-run of the correctness dimension would be needed for complete coverage.

---

**Decision: DO NOT MERGE -- 1 medium blocking (correctness); 1 medium waived (security), 3 low, 1 nit.**

The single blocker is finding #1: `lastTimestampRef.current` not reset in the streaming-text useEffect cleanup. This is a one-line fix (`lastTimestampRef.current = null;`) that prevents animation jumps when SSE tokens arrive faster than the rAF frame interval.

**Done**
- What was done: Full-regime code review of PR #1635 (24 files, +2258/-264) across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability dimensions.
- Evidence: 7 findings surfaced, 1 blocking (streaming-text lastTimestampRef cleanup), citations verified against ref 168f87c8.
- What changed: nothing -- read-only review, no mutations.
- Deferred: Correctness agent was truncated mid-stream, so additional correctness findings (skipAnimation render-phase ref read, topology edge cases) were not fully captured. The single identified blocker is sufficient to gate the merge.

---

PR #1664, opened 2026-09-14
fix(service): use os.userInfo().uid fallback in guiDomain() (#1661)

### LINKED ISSUE #1661: bug(service): guiDomain() falls back to UID 501 on multi-user macOS
## Summary

`process.getuid?.()` falls back to hardcoded UID 501 when unavailable, producing wrong `launchctl` domain strings for multi-user macOS machines.

## Location

`src/service/launchd/paths.ts:74`

```typescript
export function guiDomain(): string {
  return `gui/${process.getuid?.() ?? 501}`;
}
```

## Details

UID 501 is the first user account created on macOS. On a machine where the current user has UID 502+ (second account, migrated user), `launchctl bootstrap gui/501 <plist>` will fail with a permission error because that domain belongs to a different user.

### User-visible symptom

On macOS with UID != 501, `afk service install` fails with a cryptic launchctl permission error.

## Fix direction

Use `os.userInfo().uid` as the fallback, or validate that `getuid` returned a real value.

griffinwork40:
## Summary

Fixes #1661 — `guiDomain()` falls back to hardcoded UID 501 when `process.getuid()` is unavailable, producing wrong `launchctl` domain strings for multi-user macOS machines.

## Changes

- Replace `?? 501` with `?? userInfo().uid` using `os.userInfo()` which reads the actual UID from the OS
- Standardize the import to `node:os` protocol prefix

## Verification

- `pnpm lint` clean
- 46/46 launchd tests pass


griffinwork40 (commented):

---

**PR #1664 — fix(service): use os.userInfo().uid fallback in guiDomain() (#1661)**
**Reviewed ref:** `f153f3f617` | **Regime:** light (hotfix, 1 file, ~4 LOC)
**Change type:** hotfix — replaces hardcoded UID 501 fallback with `os.userInfo().uid`

---

## Prior automated rounds

None — this is Round 1.

---

## Decision: ✅ MERGE (with follow-ups)

**0 critical, 0 high → MERGE (with follow-ups); 1 medium advisory, 1 low advisory, 1 nit advisory.**

Blocking severity floor applied: only `critical` and `high` findings block. The fix is correct for its primary target (multi-user macOS where `process.getuid()` is available). The medium finding below is real but advisory — the fallback-path regression is theoretical on macOS production targets and warrants a follow-up, not a merge block. CI is fully green.

---

## 🟡 Advisory Findings (non-blocking — follow-up candidates)

### F-01 · medium · blocking:false · confidence:high · correctness · `src/service/launchd/paths.ts:74` · ref:f153f3f6 · citation-type:diff-context

**`os.userInfo()` can throw `SystemError` in the fallback path, replacing a silent wrong-UID with a hard crash.**

`userInfo()` is called precisely when `process.getuid?.()` returns `undefined` — environments where UID resolution is most fragile (Windows, Docker containers with synthetic users, worker threads). On macOS production targets this is a non-issue (`process.getuid` is always available), but the fallback path that fires on non-POSIX platforms now crashes instead of returning a wrong-but-recoverable domain string.

```typescript
return `gui/${process.getuid?.() ?? userInfo().uid}`;
```

**Suggestion:** Wrap the fallback in a try/catch with a last-resort 501:
```typescript
export function guiDomain(): string {
  let uid = process.getuid?.();
  if (uid == null) {
    try { uid = userInfo().uid; } catch { /* non-POSIX or unresolvable */ }
  }
  return `gui/${uid ?? 501}`;
}
```

---

### F-02 · low · blocking:false · confidence:high · test-coverage · `src/service/launchd.test.ts` · ref:f153f3f6 · citation-type:file-state (absence confirmed by grep)

**`guiDomain()` has zero direct test coverage.** The 46 launchd tests exercise `installService`/`uninstallService`/`serviceStatus` which call `guiDomain()` internally, but all mock `execFileSync` and never assert on the domain string produced. The new fallback logic (`userInfo().uid`) is not exercised at all. Absence confirmed: `git grep guiDomain` across `*.test.*` returned zero matches at the reviewed ref.

**Suggestion:** Add a `describe('guiDomain')` block in `launchd.test.ts` that stubs `process.getuid` and `os.userInfo` to verify both the primary path and the fallback.

---

### F-03 · nit · blocking:false · confidence:high · api-compat · `src/service/launchd/paths.ts:39` vs `install.ts:3`, `plist.ts:3` · ref:f153f3f6 · citation-type:file-state

**`node:os` import prefix applied only to `paths.ts`; sibling files `install.ts` and `plist.ts` still use `from 'os'`.** The PR states "Standardize the import to `node:os` protocol prefix" but the standardization is incomplete within the `launchd/` directory. Functionally equivalent; cosmetic inconsistency.

**Suggestion:** Apply the `node:` prefix to the two sibling files in the same PR, or note the standardization as partial.

---

## Spec-Compliance

|---|---|
| Replace `?? 501` with `?? userInfo().uid` | ✅ Met |
| Standardize import to `node:os` protocol prefix | ⚠️ Partial — applied to `paths.ts` only; `install.ts` and `plist.ts` unchanged (F-03) |
| `pnpm lint` clean | ✅ Asserted; CI Lint & Build passes |
| 46/46 launchd tests pass | ✅ Asserted; CI Test jobs pass |

---

## What was not checked

- Citations verified inline against branch HEAD `f153f3f61714104b099aedda96fa16876cea60c1`.
- Stated intent: PR #1664 title+body — spec-compliance assessed.
- Did not verify runtime behavior of `userInfo()` on non-macOS platforms (F-01 assessed structurally via Node.js docs).
- Did not verify whether `guiDomain()` is reachable from any non-macOS code path (the `launchd/` module is macOS-specific by design).

PR #1669, opened 2026-09-14
fix(service): warn about systemd >= 240 requirement for log output (#1659)

### LINKED ISSUE #1659: bug(service): systemd unit uses StandardOutput=append requiring systemd >= 240
## Summary

The generated systemd unit file uses `StandardOutput=append:<path>`, which requires systemd >= 240. On older Linux systems (Ubuntu 18.04 ships systemd 237), this silently fails and all bot log output is lost.

## Location

`src/service/systemd/unit.ts:107-108`

```typescript
lines.push(`StandardOutput=append:${opts.logFile}`);
lines.push(`StandardError=append:${opts.logFile}`);
```

## Details

The `append:` output directive was introduced in systemd 240 (2018). Systems older than Ubuntu 19.04 / Debian Buster / RHEL 8 will either reject the unit file or silently redirect output to null. In either case, all bot logs are lost and `journalctl --user -u afk-telegram` shows nothing.

### User-visible symptom

On Ubuntu 18.04 (still present on many servers), the Telegram bot service installs and runs but produces zero log output, making debugging impossible.

## Fix direction

Options:
1. Use `file:` directive (available since systemd 236+) instead of `append:` -- but this truncates on restart
2. Document the minimum required systemd version (240+) in the install output
3. Detect the systemd version at install time and warn or fall back

Option 2 is cheapest; option 3 is most robust.

griffinwork40:
## Summary

Fixes #1659 — the generated systemd unit uses `StandardOutput=append:<path>`, which requires systemd >= 240. On older Linux (Ubuntu 18.04, systemd 237), this silently fails and all log output is lost.

## Changes

- Add `detectSystemdVersion()` that parses `systemctl --version` output
- When systemd < 240 or version unknown, append a warning note to the install result
- Warning appears in both dry-run and real install paths

## Verification

- `pnpm lint` clean
- Systemd tests run (skipped on non-Linux, as expected)


griffinwork40 (commented):

---

## Prior Automated Rounds

None — this is Round 1. No prior findings to carry forward.

---

## PR #1669 — fix(service): warn about systemd >= 240 requirement for log output

| Field | Value |
|---|---|
| **Reviewed ref** | `2fe77e46dba7c47e0bb461d6c9dcea42014fe8a5` |
| **Change type** | Bug fix / observability (hotfix, 2 files, ~110 LOC incl. tests) |
| **Regime** | Light · scheduled sweep · blocking floor: `critical` + `high` only |

---

## ✅ Decision: MERGE (with follow-ups)

**Severity arithmetic: 0 critical, 0 high → MERGE (with follow-ups); 1 medium advisory, 4 low advisory.**

Per sweep policy, only `critical` and `high` findings block. The one medium finding (F1) and all four low findings are non-blocking. The fix correctly addresses the stated problem (#1659): `StandardOutput=append:<path>` requires systemd ≥ 240, and the PR now warns users on older systems or when the version cannot be detected. CI is fully green.

---

## Findings

### 🟡 Medium — Advisory (non-blocking per sweep policy)

**F1 · Correctness · `src/service/systemd/install.ts:234,270` · ref:2fe77e46**

`detectSystemdVersion()` is called twice — once in the dry-run branch (line 234) and once in the real-install branch (line 270). Each call forks a fresh `systemctl --version` subprocess; the result is not memoized. On the real-install path, the call runs *after* the service is already enabled.

No functional regression — the warning reaches the user via the returned `notes` array in both cases — but the duplicate subprocess spawn adds latency on restricted or slow-init systems.

**Fix:** Hoist one call before the `dryRun` branch check:
```ts
const detectedVersion = detectSystemdVersion();
// then use detectedVersion in both the dryRun block and the real-install block
```

---

### 🔵 Low — Advisory (follow-up candidates, suggest a follow-up issue)

**F2 · Correctness · `src/service/systemd/install.ts:74` · ref:2fe77e46**

Regex `/\b(\d+)\b/` matches the first integer anywhere in the first output line. In practice the `systemctl --version` format is stable (`systemd NNN` or `systemd NNN (NNN-...)`), but the pattern is not anchored to the `systemd` keyword and would silently mis-parse unexpected prefix formats.

**Fix:** `/^systemd\s+(\d+)/` — unambiguous and zero false-positive risk.

---

**F3 · Spec-compliance · `src/service/systemd.test.ts:22–44` · ref:2fe77e46**

The core behavior introduced by this fix — warning on systemd < 240 and warning when version is unknown — has no test coverage. The `mockSpawnSync` default returns v252 (happy path only). The `< 240` and `undefined` branches, which are the entire point of this PR, are never exercised in the test suite.

**Fix:** Add two `it()` cases:
```ts
// v237 → warning appears in notes
mockSpawnSync.mockReturnValueOnce({ status: 0, stdout: 'systemd 237\n', stderr: '', error: undefined });
// spawn failure → "Could not detect" warning appears
mockSpawnSync.mockReturnValueOnce({ status: 1, stdout: '', stderr: '', error: new Error('not found') });
```

---

**F4 · Test coverage · `src/service/systemd.test.ts:41` · ref:2fe77e46**

`detectSystemdVersion` is imported in the test file but never directly tested. All three observable return paths (valid integer, `undefined` on spawn failure, `undefined` on regex-parse failure) have zero direct unit coverage.

**Fix:** Add a `describe('detectSystemdVersion', ...)` block with one `it()` per output branch.

---

**F5 · Perf / Observability · `src/service/systemd/install.ts:266–281` · ref:2fe77e46**

No log statement on the healthy path (detected version ≥ 240). Operators on modern systems get no signal that version detection ran or what it returned.

**Fix:** Emit a debug-level log on the happy path, e.g. `debug(\`systemd version detected: ${systemdVersion}\`)`.

---

## Spec Compliance

|---|---|
| Warn when systemd < 240 | ⚠️ Logic present, branch untested (F3) |
| Warn when version unknown | ⚠️ Logic present, branch untested (F3) |
| No warning on systemd ≥ 240 | ✅ Covered by default mock (v252) |
| `detectSystemdVersion()` parses `systemctl --version` output | ✅ Implemented; anchoring nit (F2) |
| Warning in both dry-run and real-install paths | ✅ Both paths covered |

---

## What Was Not Checked

- Citations verified against branch HEAD `2fe77e46dba7c47e0bb461d6c9dcea42014fe8a5`. Stated intent: PR #1669 title + body — spec-compliance assessed.
- Did not verify runtime behavior on an actual systemd < 240 host (static analysis only).
- Did not review the macOS launchd path (out of scope for this change).
- Did not check whether the `notes` field surface reaches the user's terminal in all UI contexts (CLI vs. Telegram vs. dashboard).
- Did not audit changelog or documentation for user-facing notes on the new behavior.

PR #1666, opened 2026-09-14
fix(service): add ThrottleInterval to launchd plist (#1658)

griffinwork40:
## Summary

Fixes #1658 — the generated launchd plist uses `KeepAlive=true` but had no `ThrottleInterval` key, risking crash-loops at the default rate.

## Changes

- Add `<key>ThrottleInterval</key><integer>30</integer>` after the KeepAlive block
- Update inline snapshot in test
- Add JSDoc documenting the invariant

## Verification

- `pnpm lint` clean
- 46/46 launchd tests pass (including updated snapshot)

PR #1686, opened 2026-09-16
fix(service): add plist upgrade path for existing installs

griffinwork40:
## Summary

Addresses the Codex P1 review finding on #1666: existing users who already have a launchd plist installed never get the `ThrottleInterval` fix because `installService()` returns `already-installed` when the plist file exists, and neither `restart` nor postinstall ever rewrites it.

### Core: plist upgrade path (`src/service/launchd/install.ts`)
- Add `UpgradeResult` type + `upgradeService()` function
- Re-renders the plist from current code, compares to on-disk content, atomically replaces when different (same tmp+rename pattern as `installService`)
- Returns `already-current` when the plist matches (no-op), `not-installed` when absent, `upgraded` on success

### Contract: `ServiceManager.upgrade()` (`src/service/types.ts`)
- Add `ServiceUpgradeOutcome` type
- Add `upgrade()` method to the `ServiceManager` interface

### Backends
- **launchd** (`manager.ts`): Full `upgrade()` implementation + wires `restart()` to call `upgradeService()` before kickstart, so every restart (including postinstall's `restartLaunchdServices`) automatically upgrades stale plists
- **systemd** (`manager.ts`): Stub that returns a "not yet supported" message advising uninstall/reinstall

### CLI: `afk service upgrade <name>` (`src/cli/commands/service.ts`)
- New subcommand for explicit on-demand plist upgrades

### Review findings from #1666
- **F-01** (JSDoc): ThrottleInterval comment now correctly states launchd's default is 10s, not uncapped
- **F-02** (test coverage): Added `ThrottleInterval` assertion to the WatchPaths `renderPlist` test to pin the invariant

### Tests
- 4 new tests for `upgradeService`: not-installed, already-current, atomically-replaces-stale-plist, preserves-0o600-mode
- All 50/50 launchd tests pass

## How it propagates to existing users

When a user upgrades `agent-afk` via `npm install -g`:
1. `postinstall.mjs` calls `restartLaunchdServices()` which runs launchctl kickstart
2. The kicked service calls into the new code's `restart()` method on its next lifecycle event
3. `restart()` now calls `upgradeService()` first, which detects the stale plist and atomically replaces it
4. The restarted service runs with the updated plist (including `ThrottleInterval=30`)

Users can also run `afk service upgrade telegram` or `afk service upgrade daemon` explicitly.

## Verification

- `pnpm lint` clean
- `pnpm test src/service/` -- 54 passed, 21 skipped (systemd on macOS)
- `pnpm audit:filesize:check` clean


griffinwork40:
Shadow-verify confirms the finding. Now I can deliver the final synthesis.

---

## Review: PR #1686 — `fix(service): add plist upgrade path for existing installs`

**Target:** PR #1686 (`afk/fix-codex-p1-issue` → `main`)
**Ref:** `41846fce`
**Regime:** light (hotfix, 232 lines, 7 files)
**Stated intent:** PR title + body (present)

---

### Findings

**F-C3 · medium · blocking:true · confidence:high · spec-compliance + correctness · `scripts/postinstall.mjs:112-138` · ref:41846fce · file-state**

The PR body claims the upgrade propagates automatically to existing users via: `postinstall → restartLaunchdServices() → restart() → upgradeService()`. This chain is broken. `restartLaunchdServices()` in `scripts/postinstall.mjs` calls `execFileSync('launchctl', ['kickstart', '-k', ...])` directly (line 131) -- it does not call `launchdManager.restart()` and never triggers `upgradeService()`. The function is a standalone `.mjs` that cannot import compiled TS helpers (its own JSDoc at line 99 states: *"Kept in sync by hand -- this plain .mjs cannot import the compiled TS helpers"*). The plist upgrade path is unreachable from postinstall. Existing users who run `npm install -g` will get the new code but NOT an upgraded plist. **Shadow-verified: CONFIRMED.**

**Evidence:** `scripts/postinstall.mjs:131`: `execFn(['kickstart', '-k', \`gui/${uid}/${label}\`]);`

**Suggestion:** Either (a) add inline plist-rewrite logic to `restartLaunchdServices()` in postinstall.mjs (duplicating the compare-and-replace, since it cannot import TS), or (b) have postinstall shell out to `afk service upgrade <name>` before the kickstart.

---

**F-C1 · medium · blocking:true · confidence:high · correctness · `src/service/launchd/install.ts:251,297` · ref:41846fce · file-state**

The JSDoc on `upgradeService()` (line 251) states *"The `wx` flag is NOT used here"*, but the actual write at line 297 uses `flag: 'wx'`. After the `unlinkSync(tmpPath)` on line 296, the `wx` flag introduces a TOCTOU window: another concurrent process (or a concurrent `restart()` call from two services) can recreate `tmpPath` between the unlink and the write, causing a spurious `EEXIST` that makes the upgrade silently fail (the `catch` returns `{ kind: 'failed' }`). Since `restart()` swallows this result (see F-C2), the user gets a restart with a stale plist and no diagnostic output.

**Evidence:**
```
if (existsSync(tmpPath)) unlinkSync(tmpPath);
writeFileSync(tmpPath, desired, { encoding: 'utf-8', flag: 'wx', mode: 0o600 });
```

**Suggestion:** Use `flag: 'w'` (matching the documented intent) since the preceding unlink already handles stale files, or drop the unlink and keep `wx` for the no-prior-crash path.

---

**F-C2 · low · blocking:false · confidence:high · correctness · `src/service/launchd/manager.ts:114` · ref:41846fce · file-state**

`try { upgradeService(name); } catch { /* best-effort */ }` -- `upgradeService()` returns result objects for every error path and never throws. The `catch` is dead code, and the return value is silently discarded. A failed upgrade (including the EEXIST from F-C1) produces no log output before the kickstart proceeds.

**Evidence:** `try { upgradeService(name); } catch { /* best-effort */ }`

**Suggestion:** Capture the result and log on `kind === 'failed'` so operators can diagnose stale-plist situations.

---

**F-T1 · low · blocking:false · confidence:high · test-coverage · `src/service/launchd.test.ts` · ref:41846fce · file-state**

No test covers `restart()` calling `upgradeService()`. Grep for "restart" in the test file yields only error-message string matches at lines 592 and 606, not a `restart()` unit test. The new behavior at `manager.ts:114` has zero coverage.

**Suggestion:** Add a test seeding a stale plist, calling `mgr.restart()`, and asserting the plist was rewritten before the kickstart.

---

**F-T2 · low · blocking:false · confidence:high · test-coverage · `src/cli/commands/service.ts:158-184` · ref:41846fce · diff-context**

The new `afk service upgrade <name>` CLI subcommand has no integration test. Grep for "upgrade" in `src/service/index.test.ts` returns no results (file does not exist at ref).

**Suggestion:** Add CLI smoke tests for the upgrade subcommand covering each outcome kind.

---

**F-P1 · nit · blocking:false · confidence:high · perf-observability · `src/service/launchd/manager.ts:114` · ref:41846fce · file-state**

`upgradeService()` runs `readFileSync` + `renderPlist` on every `restart()` call. At current scale (<1ms synchronous) this is not a concern.

---

### Dimensions with no issues

- **security** -- no issues found. `0o600` mode enforced on upgraded plist (install.ts:297). No new env-var or token surfaces.
- **api-compat** -- `ServiceManager` interface gains `upgrade()`. Both production implementors (`launchdManager`, `systemdManager`) implement it. No other implementors exist (confirmed via grep). No breaking change.

---

### What was not checked

- Citations verified inline against branch HEAD `41846fce`.
- Stated intent: PR title + body -- spec-compliance assessed.
- Shadow-verified: F-C3 postinstall propagation claim (CONFIRMED broken).
- Not checked: runtime behavior of `launchctl kickstart` after plist file replacement (no macOS CI environment).
- Not checked: whether the `wx` TOCTOU in F-C1 is triggerable in practice (requires concurrent `restart()` calls on the same service, which is possible during postinstall + a manual `afk service restart` race).

---

### Decision: DO NOT MERGE -- 2 medium blocking (1 spec-compliance/correctness, 1 correctness); 2 low, 1 nit.

The primary blocker is F-C3: the PR's stated automatic-propagation path through postinstall is broken, meaning existing users will NOT receive the ThrottleInterval fix automatically on `npm install -g` -- the core goal of the PR. The secondary blocker (F-C1) is a correctness bug where `flag: 'wx'` contradicts the documented intent and introduces a TOCTOU race.

**Done**
- What was done: Full light-regime review of PR #1686 across all 6 dimensions, with inline citation verification (Wave 1.5) and shadow-verification of the primary finding (F-C3 postinstall propagation claim).
- Evidence: Review findings above, shadow-verify verdict CONFIRMED.
- What changed: Nothing -- read-only review, no mutations.
- Deferred: None.

---

griffinwork40 (commented):

---

## Prior Round Dispositions

One prior review round was posted on commit `41846fce` (2026-09-16T20:39:48Z) by griffinwork40 via `/review --post github`. The latest commit `e054ec2` (2026-09-16T21:25:56Z) explicitly addresses that round's findings per its commit message: "fix(service): address PR #1686 review feedback".

| Prior Finding | Disposition |
|---|---|
| **F-C3 (medium/blocking):** postinstall's `restartLaunchdServices()` called raw `launchctl kickstart -k` directly — never triggered `upgradeService()`, breaking the stated auto-propagation path. | **FIXED** — commit `e054ec2`. `restartLaunchdServices()` now resolves `dist/cli.mjs` and delegates to `node dist/cli.mjs service restart <name>` (which internally calls `upgradeService()` + selects the correct reload strategy). Falls back to raw kickstart only when `dist/cli.mjs` is absent. Verified at `scripts/postinstall.mjs:131–183` at current ref. |
| **F-C1 (medium/blocking):** `flag: 'wx'` in `upgradeService()` contradicted JSDoc and introduced a TOCTOU race from unlink+exclusive-create. | **FIXED** — commit `e054ec2`. `upgradeService()` now uses `flag: 'w'` (line 297). JSDoc at lines 250–254 correctly states `wx` is NOT used. The stale-tmp cleanup + non-exclusive write eliminates the original race. |
| **F-C2 (low/non-blocking):** Dead `try { upgradeService(name); } catch {}` wrapper in `restart()` — return value silently discarded, no log on failure. | **FIXED** — commit `e054ec2`. `restart()` now captures the result: `const upgradeResult = upgradeService(name);` and uses it to select the reload strategy (full reload when `'upgraded'`, kickstart otherwise). Comment explicitly documents that `upgradeService()` never throws. |
| **F-T1 (low/non-blocking):** No test covers `restart()` calling `upgradeService()`. | **FIXED** — commit `e054ec2`. Four new manager restart tests added at `launchd.test.ts:795–970`: not-installed, uses-kickstart-when-already-current, uses-full-reload-when-upgraded, returns-failed-on-error. |
| **F-T2 (low/non-blocking):** No integration test for `afk service upgrade <name>` CLI subcommand. | **PARTIALLY FIXED** — commit `e054ec2`. Three manager `upgrade()` adapter tests added (not-installed, already-current, upgraded). No end-to-end CLI integration test exists — the test file `src/service/index.test.ts` is absent at this ref. |
| **F-P1 (nit/non-blocking):** `upgradeService()` runs `readFileSync` + `renderPlist` on every `restart()` call. | **STILL-PRESENT** — by design. The comment at `manager.ts:104–120` documents this as intentional: the check is the mechanism that auto-propagates plist changes. At current scale (<1ms synchronous) this is not a performance concern. |

---

## PR #1686 — fix(service): add plist upgrade path for existing installs

| Field | Value |
|---|---|
| **Reviewed ref** | `e054ec235f742f208312b50bda6f49e026ecb6f8` |
| **Change type** | Hotfix/feature (8 files, ~500 LOC) |
| **Regime** | Full · scheduled sweep · blocking floor: `critical` + `high` only |
| **Author** | griffinwork40 |
| **CI** | All green except `Smoke test (Windows CLI)` FAILURE — pre-existing; Windows tests are SKIPPED on this repo |

---

## ✅ Decision: MERGE (with follow-ups)

**Severity arithmetic: 0 critical, 0 high → MERGE (with follow-ups); 4 medium advisory, 4 low advisory.**

The PR's core stated intent — automatic plist propagation to existing users via `npm install -g → postinstall → restart → upgradeService()` — is fully met. The prior round's P1 blocker (broken postinstall propagation) is fixed. All prior findings have verified dispositions above. The new medium findings below are real but advisory under the sweep blocking floor.

---

## 🟡 New Findings — Round 1 (all advisory, non-blocking)

### F-1 · medium · blocking:false · confidence:high · correctness · `src/cli/commands/service.ts:180` + `src/service/launchd/manager.ts:133` · ref:e054ec2 · file-state

**`afk service upgrade <name>` followed by `afk service restart <name>` does not apply new plist keys to launchd's in-memory job definition.**

`upgrade` rewrites the plist file. `restart` then calls `upgradeService()` again, finds `already-current`, and falls through to `kickstart -k` — which restarts the process but does NOT reload the job definition. New keys (ThrottleInterval, EnvironmentVariables) remain at their old values in launchd's memory. The primary propagation path (postinstall → restart on a stale plist → `upgraded` → full reload) is correct; this only affects the explicit two-step CLI workflow.

**Suggestion:** Have the `upgrade` CLI action call `mgr.restart(name)` when `result.kind === 'upgraded'`, so the user gets upgrade + reload in one command. Or change the success message to name the limitation.

---

### F-2 · medium · blocking:false · confidence:high · correctness · `src/service/launchd/manager.ts:133` · ref:e054ec2 · file-state

**`restart()` calls `upgradeService(name)` with no opts — `noWatch` defaults to `false`, silently re-enabling WatchPaths for dev-tree telegram installs originally created with `--no-watch`.**

`upgradeService()` re-renders with `WatchPaths` present → file differs from the no-watch install → returns `'upgraded'` → full reload. Every `restart` call silently re-enables the watch and forces a full job reload. Only affects dev-tree telegram (daemon is unaffected — `resolveWatchPaths` returns `undefined` for daemon).

**Suggestion:** Read the on-disk plist before upgrading to detect whether WatchPaths is present, and set `noWatch` accordingly. Or thread install-time opts through `restart()`.

---

### F-3 · medium · blocking:false · confidence:high · security · `src/service/launchd/install.ts:293–297` · ref:e054ec2 · file-state

**`upgradeService()` has a TOCTOU race between stale-tmp cleanup (`unlinkSync`) and write (`flag: 'w'`).**

Between `unlinkSync(tmpPath)` and `writeFileSync(tmpPath, ..., { flag: 'w' })`, a local attacker who can race a symlink into `tmpPath` causes `writeFileSync` to follow the symlink and overwrite the target. Blast radius is bounded: requires same-user local access, plist content only (no secrets), and the target is user-writable space.

**Suggestion:** Use `flag: 'wx'` with `try/catch` on EEXIST, matching `installService()`'s defense.

---

### F-4 · medium · blocking:false · confidence:high · correctness · `scripts/postinstall.mjs:172` · ref:e054ec2 · file-state

**`existsSync(cliMjs)` at line 172 uses the raw import, not the injectable `existsFn` — the CLI delegation branch is untestable without module-level mocking.**

All existing `restartLaunchdServices` tests only exercise the fallback kickstart path because `dist/cli.mjs` doesn't exist at the test-runner's computed path. The injectable `restartFn` is never reached.

**Suggestion:** Gate on `existsFn(cliMjs)` instead of bare `existsSync(cliMjs)`.

---

### F-5 · low · blocking:false · confidence:high · correctness · `src/service/launchd/manager.ts:148–162` · ref:e054ec2 · file-state

**When the reload's first phase succeeds but the second phase fails, the service is left unloaded with no recovery advisory in the error message.**

The returned `{ kind: 'failed', reason: ... }` gives no hint that the service is now down and needs reinstall to recover.

**Suggestion:** Append a recovery advisory to the error: `"Service is unloaded — recover with: afk service uninstall <name> && afk service install <name>"`.

---

### F-6 · low · blocking:false · confidence:high · correctness · `src/service/launchd/manager.ts:133` · ref:e054ec2 · file-state

**When `upgradeService()` returns `'failed'` (e.g. afk binary not found), `restart()` silently falls through to kickstart with no signal that the plist upgrade was skipped.**

The `ServiceRestartOutcome` has no field to communicate that upgrade was attempted and failed. The caller gets `{ kind: 'restarted' }` with no indication the ThrottleInterval fix was not applied.

**Suggestion:** Add an `upgradeSkipped?: string` field to the outcome, or emit a debug log with the upgrade failure reason.

---

### F-7 · low · blocking:false · confidence:high · api-compat · `src/service/types.ts:119` · ref:e054ec2 · file-state

**`upgrade()` is added as a required method on `ServiceManager` — any external TypeScript consumer implementing the interface will see a compile break.**

Both in-tree backends (launchd, systemd) satisfy the new contract. No known external implementors exist. This is a minor breaking interface change shipped as a patch.

**Suggestion:** Note in CHANGELOG as a breaking change for downstream `ServiceManager` implementors, or make `upgrade` optional until the interface is stable.

---

### F-8 · low · blocking:false · confidence:high · perf-observability · `src/service/launchd/manager.ts:133` + `scripts/postinstall.mjs:160–183` · ref:e054ec2 · file-state

**No logging distinguishes which restart strategy was used (CLI delegation vs. raw kickstart in postinstall) or whether upgrade was attempted.**

Both success and fallback paths produce the same user-visible output. The upgrade check on every `restart()` is silent. Add debug-level structured logging.

---

## Spec-Compliance

|---|---|
| Add `upgradeService()` to launchd/install.ts | ✅ Met |
| Add `upgrade()` to ServiceManager interface + both backends | ✅ Met |
| Wire `restart()` to call `upgradeService()` before kickstart | ✅ Met |
| Propagation: postinstall → restartLaunchdServices → CLI restart → upgrade | ✅ Met (F-C3 from prior round fixed) |
| `afk service upgrade <name>` CLI subcommand | ✅ Met (F-1: follow-up message is misleading) |
| F-01 JSDoc fix: ThrottleInterval comment corrected | ✅ Met |
| F-02 test coverage: ThrottleInterval assertion in WatchPaths test | ✅ Met |
| 4 new upgradeService tests | ✅ Met |
| CI green | ⚠️ `Smoke test (Windows CLI)` FAILURE — pre-existing, not caused by this PR |

---

## What Was Not Checked

- Citations verified against branch HEAD `e054ec235f742f208312b50bda6f49e026ecb6f8`.
- Stated intent: PR #1686 title + body — spec-compliance assessed.
- Prior round dispositions verified against current ref source.
- Did not verify runtime behavior of launchd reload cycle (read-only review).
- Did not verify whether `dist/cli.mjs` resolves correctly under all package manager layouts (npm, pnpm, yarn).
- Shadow-verify not triggered: no critical/high findings; no blocking overrides.

griffinwork40 (commented):

---

## Prior Round Dispositions (Round 1 at `e054ec2` → Round 2 at `0b73528f`)

Round 1 was posted at ref `e054ec2` (2026-09-16T22:10:05Z). Four commits landed since: two merge commits (`fc697ffe`, `a0956f88`), plus two CI fixes (`5ad2f404` — drain-aware stdout.write for --version, `0b73528f` — gate async IIFE behind !isVersionFlag).

| Round 1 Finding | Disposition |
|---|---|
| **F-1 (medium/advisory):** `upgrade` then `restart` CLI workflow fails to apply new plist keys — restart finds already-current and uses kickstart instead of bootout+bootstrap. | **STILL-PRESENT** — no fix committed. `manager.ts:133`: `upgradeService(name)` re-renders the plist, matches the just-written file, returns `already-current`, falls through to kickstart. |
| **F-2 (medium/advisory):** `restart()` calls `upgradeService(name)` with no opts — `noWatch` defaults to false, silently re-enabling WatchPaths for --no-watch installs. | **STILL-PRESENT** — no fix committed. `install.ts:267`: `opts.noWatch` absent → `false` → `resolveWatchPaths()` adds WatchPaths. |
| **F-3 (medium/advisory):** TOCTOU race between `unlinkSync(tmpPath)` and `writeFileSync(tmpPath, ..., { flag: 'w' })` in `upgradeService()`. | **STILL-PRESENT** — no fix committed. `install.ts:296-297` unchanged. |
| **F-4 (medium/advisory):** `existsSync(cliMjs)` at `postinstall.mjs:172` uses raw import, not injectable `existsFn`. | **STILL-PRESENT** — no fix committed. Verified at ref: line 172 uses `existsSync` directly while line 125 defines `existsFn`. |
| **F-5 (low/advisory):** When bootout succeeds but bootstrap fails in `restart()`, service left unloaded with no recovery advisory. | **STILL-PRESENT** — no fix committed. `manager.ts` catch block returns `{ kind: 'failed' }` with no recovery hint. |
| **F-6 (low/advisory):** When `upgradeService()` returns `'failed'`, `restart()` silently falls through to kickstart. | **STILL-PRESENT** — comment at `manager.ts:160` names the fallthrough (`"already-current, not installed, or upgrade failed"`) but emits no diagnostic. |
| **F-7 (low/advisory):** `upgrade()` added as required method on `ServiceManager` interface — breaking change for external implementors. | **STILL-PRESENT** — by design. Both in-tree backends satisfy the contract. No known external implementors. |
| **F-8 (low/advisory):** No logging distinguishes which restart strategy was used (kickstart vs bootout+bootstrap). | **STILL-PRESENT** — comments document the distinction, but no runtime log output. |

---

## PR #1686 — fix(service): add plist upgrade path for existing installs

| Field | Value |
|---|---|
| **Reviewed ref** | `0b73528f4db9b207c2015623ea20e631d546b0b3` |
| **Change type** | Hotfix (9 files, ~700 LOC including tests) |
| **Regime** | Light · scheduled sweep · blocking floor: `critical` + `high` only |
| **Author** | griffinwork40 |
| **CI** | All green except `Smoke test (Windows CLI)` FAILURE — pre-existing; this PR's new commits attempt to fix it |

---

## ✅ Decision: MERGE (with follow-ups)

**Severity arithmetic: 0 critical, 0 high → MERGE (with follow-ups); 4 medium advisory (all non-blocking per sweep floor, all carried from Round 1), 4 low advisory (carried), 1 nit advisory (new).**

The PR's core stated intent — automatic plist propagation to existing users via `postinstall → restart → upgradeService() → bootout+bootstrap` — is fully met and verified across both rounds. The primary propagation path is correct. All Round 1 findings have been addressed or are advisory follow-ups that do not affect the primary path. The new commits (`5ad2f404`, `0b73528f`) correctly fix the Windows `--version` pipe-flush race by hoisting the version check before the async IIFE and gating the IIFE behind `!isVersionFlag`.

---

## 🟡 Carried Findings (all advisory, non-blocking)

All four medium findings from Round 1 remain open and advisory. They affect edge paths (explicit CLI two-step workflow, --no-watch dev-tree installs, local symlink attacker, test isolation) — none blocks the primary postinstall propagation path. See disposition table above for details and file:line evidence.

All four low findings from Round 1 remain open and advisory. They affect observability and interface evolution — no production behavior impact.

---

## 🆕 New Finding (Round 2)

### N-1 · nit · blocking:false · confidence:high · correctness · `src/cli/index.ts:266` · ref:0b73528f · citation-type:diff-context

**`if (!isVersionFlag) (async () => { ... })();` — braceless if guarding a multi-hundred-line async IIFE.**

The structural fix is correct — the IIFE is unconditionally skipped on `--version`. But the braceless `if` guarding ~120 lines of async init is a refactor-accident magnet. A future edit could accidentally detach the guard.

**Suggestion:** Wrap in `if (!isVersionFlag) { (async () => { ... })(); }` for clarity.

---

## Spec-Compliance

|---|---|
| Add `upgradeService()` to launchd/install.ts | ✅ Met |
| Add `upgrade()` to ServiceManager interface + both backends | ✅ Met |
| Wire `restart()` to call `upgradeService()` before kickstart | ✅ Met |
| Propagation: postinstall → CLI restart → upgrade + bootout/bootstrap | ✅ Met |
| `afk service upgrade <name>` CLI subcommand | ✅ Met |
| ThrottleInterval JSDoc fix (F-01 from #1666) | ✅ Met |
| ThrottleInterval test assertion (F-02 from #1666) | ✅ Met |
| 4 new upgradeService tests + restart/upgrade manager tests | ✅ Met |
| Windows --version pipe-flush fix (new commits) | ✅ Met — structurally sound |
| CI green | ⚠️ `Smoke test (Windows CLI)` FAILURE — the new commits target this fix; may need CI re-run |
| Scope creep | The Windows --version fix is tangentially related (same branch, not same feature), but is a legitimate CI fix. |

---

## What Was Not Checked

- Citations verified against branch HEAD `0b73528f4db9b207c2015623ea20e631d546b0b3`.
- Stated intent: PR #1686 title + body — spec-compliance assessed.
- Prior round dispositions verified against current ref source.
- Did not verify runtime behavior of launchd bootout+bootstrap reload cycle (read-only review).
- Did not verify whether `dist/cli.mjs` resolves correctly under all package manager layouts (npm, pnpm, yarn).
- Did not verify Windows-specific behavior of `process.stdout.write` returning `false` in non-TTY pipes (static analysis only).
- Shadow-verify not triggered: no critical/high findings.

griffinwork40:
All citations confirmed. Here is the synthesized review.

---

# Review: PR #1686 — `fix(service): add plist upgrade path for existing installs`

**Ref:** `08b8867e` | **Branch:** `afk/fix-codex-p1-issue` → `main` | **Regime:** full (+519/-21, 9 files)

**Stated intent:** Existing users who already have a launchd plist installed never get the `ThrottleInterval` fix because `installService()` returns `already-installed`. Adds `upgradeService()` with atomic tmp+rename, wires `restart()` to upgrade-then-restart (bootout+bootstrap when plist changed, kickstart when unchanged), new CLI `afk service upgrade <name>`, and postinstall delegates to compiled CLI restart when available.

---

## Blocking findings

### 1. `upgradeService()` can throw despite "never throws" contract
`medium · blocking:true · confidence:high · correctness + security`
**`src/service/launchd/install.ts:289`** · ref:08b8867 · file-state

`readFileSync(path, 'utf-8')` at line 289 is unwrapped. The `existsSync(path)` guard at line 257 checks file existence, but if the plist is deleted between that check and the read, `readFileSync` throws ENOENT. The comment in `manager.ts` ("upgradeService() returns a result object and never throws -- the try/catch wrapper was dead code") is load-bearing: `restart()` calls `upgradeService(name)` bare with no surrounding try/catch. An uncaught throw in `restart()` propagates as an unformatted crash through the CLI, and through `postinstall.mjs` (where it is caught by the outer fail-open try/catch, but the upgrade silently fails).

**Suggestion:** Wrap `readFileSync(path, 'utf-8')` in a try/catch returning `{ kind: 'failed', reason: ... }` so the "never throws" invariant actually holds.

### 2. `upgradeService()` tmp write uses `flag: 'w'` instead of `'wx'` -- symlink race
`medium · blocking:true · confidence:high · security`
**`src/service/launchd/install.ts:297`** · ref:08b8867 · diff-context

`installService()` uses `flag: 'wx'` (O_EXCL) to close the TOCTOU window between existence check and write -- preventing a symlink-based file overwrite at the predictable `~/Library/LaunchAgents/com.afk.<name>.plist.tmp` path. `upgradeService()` uses `flag: 'w'` and documents the gap explicitly ("The `wx` flag is NOT used here..."). The `existsSync(tmpPath)` + `unlinkSync(tmpPath)` cleanup between the two calls leaves the same TOCTOU window open.

Impact is bounded (content is AFK-generated XML, not arbitrary data; requires local user access), but the mitigation exists in the sibling function and was deliberately weakened here for crash-recovery convenience.

**Suggestion:** Use `flag: 'wx'`; catch EEXIST, `unlinkSync`, retry once. Preserves both the TOCTOU defense and crash-recovery cleanup.

---

## Non-blocking findings

### 3. `upgradeService()` skips `mkdirSync` for log directory
`medium · blocking:false · correctness` · waived: bounded, non-data-affecting -- the parent dir is guaranteed to exist by the `existsSync(path)` guard for the LaunchAgents dir; the log dir omission only affects services launched on a machine that never ran `installService()` (the log dir would have been created then), and launchd creates the log file itself if the dir is missing.
**`src/service/launchd/install.ts:255-313`** · ref:08b8867 · file-state

`installService()` calls `mkdirSync(dirname(logFile), { recursive: true })` before writing. `upgradeService()` omits this. The rendered plist references the log path in `standardOutPath`/`standardErrorPath`. If the log directory was never created by a prior `installService()` call, launchd will fail to open the log file on the next launch.

**Suggestion:** Add `mkdirSync(dirname(logFile), { recursive: true })` to match `installService()`.

### 4. Bootstrap failure after bootout leaves service unloaded with unhelpful error
`medium · blocking:false · correctness` · waived: bounded -- the user ran an explicit `restart` command and can recover with `uninstall + install`; the service was already stopping when this path fires.
**`src/service/launchd/manager.ts:145-165`** · ref:08b8867 · diff-context

In the `upgraded` branch of `restart()`, if bootout succeeds but bootstrap fails, the error message is the raw launchctl string with no recovery guidance. The user sees "Restart failed: \<opaque launchctl error\>" with the service stopped.

**Suggestion:** Append recovery guidance: `" -- run 'afk service uninstall <name> && afk service install <name>' to recover"`.

### 5. Silent upgrade failure in restart fallback
`low · blocking:false · api-compat`
**`src/service/launchd/manager.ts:167-175`** · ref:08b8867 · diff-context

When `upgradeService()` returns `failed`, `restart()` silently falls through to `kickstart -k`. The operator sees "Restarted com.afk.daemon" while the plist on disk is still stale. No diagnostic is surfaced.

**Suggestion:** Surface the upgrade failure as a warning before falling through, or add a field to `ServiceRestartOutcome`.

### 6. Missing test: bootstrap-fails-after-bootout in restart path
`medium · blocking:false · test-coverage` · waived: bounded, non-data-affecting -- the code path is simple (catch + return failed) and launchctl bootstrap failures are well-tested in `installService()`.
**`src/service/launchd.test.ts`** · ref:08b8867

The `restart (manager)` tests cover not-installed, kickstart-when-current, bootout+bootstrap-when-upgraded, and kickstart-fails. Missing: bootstrap throws after successful bootout (the `return { kind: 'failed', reason }` path in the upgraded branch). Grep pattern `bootstrap.*fail` in the restart tests: zero matches.

### 7. Missing test: postinstall CLI delegation path
`low · blocking:false · test-coverage`
**`tests/postinstall.test.ts`** · ref:08b8867

The `restartFn` injection point is documented but no test exercises the CLI delegation path (`existsSync(cliMjs) === true` + `restartFn` called). Grep for `restartFn` or `cliMjs` in tests/postinstall.test.ts: zero matches.

### 8. No log distinguishing upgraded-restart from plain restart
`low · blocking:false · perf-observability`
**`src/service/launchd/manager.ts:130-165`** · ref:08b8867 · diff-context

`restart()` returns `{ kind: 'restarted', label }` regardless of whether the plist was upgraded. The CLI cannot distinguish "restarted because unchanged" from "restarted after plist upgrade." The `upgrade` subcommand is more informative for this use case.

### 9. `--version` move is undocumented scope creep
`nit · blocking:false · spec-compliance`
**`src/cli/index.ts`** · ref:08b8867 · diff-context

The PR body does not mention the `--version` move to module top level. Commit `08b8867e` is a separate commit addressing a distinct Windows CI failure. The change is self-contained and does not interact with service upgrade logic.

### 10. `ServiceManager.upgrade()` is a new required interface method
`low · blocking:false · api-compat`
**`src/service/types.ts:127`** · ref:08b8867 · diff-context

New required method on a public interface. Both production implementors (launchdManager, systemdManager) are updated. No external implementors found in the production tree. Published package consumers implementing this interface would get a type error. [UNVERIFIED: no production importers outside this repo]

---

## What was checked
- Citations verified inline against branch HEAD `08b8867e59a1e03030fe68c4e28cc472ae176315`.
- Stated intent: PR #1686 title+body -- spec-compliance assessed.
- Full diff across all 9 files. `upgradeService()` body compared structurally to `installService()`. `restart()` control flow for all upgrade result kinds. `postinstall.mjs` CLI delegation logic and fallback. Test coverage for `upgradeService`, `restart`, and `upgrade` manager adapter. `--version` placement in `index.ts`. systemd stub return value.

## What was not checked
- End-to-end integration of postinstall CLI delegation when `dist/cli.mjs` exists but `afk service restart` itself fails partway through (after plist write, before bootout).
- Runtime behavior of `renderPlist` output format (trusted from existing snapshot tests).
- `label.replace(/^com\.afk\./, '')` correctness against all registered `SERVICE_NAMES` values.
- Semver implications of the `ServiceManager` interface change for downstream consumers.

---

**Decision: DO NOT MERGE -- 2 medium blocking (1 correctness+security, 1 security); 4 medium waived, 1 low, 1 nit.**

Findings 1 and 2 are both in `upgradeService()` and both addressable with small targeted fixes: (1) wrap `readFileSync` in try/catch, (2) switch to `flag: 'wx'` with EEXIST retry. Neither requires structural changes.

---

griffinwork40:
The `environment` field is an API contract (exists on the interface, wired in the manager's `install()` and `upgrade()` methods) but currently no production caller passes it. The CLI `install` never sends `environment`.

This means F-1's scenario (custom env vars being stripped) is theoretically reachable via the API but currently unreachable via any production code path. Let me re-evaluate:

- If a user installed with the `/service-setup` skill or directly via `afk service install`, no `environment` is passed (CLI doesn't expose it).
- The only way a plist would have custom env vars is if someone called `installService()` programmatically with `opts.environment` -- possible but rare.
- Even then, `upgradeService()` without `opts.environment` would re-render the "canonical" plist, which would differ, and the env vars would be stripped.

The finding is valid as a latent contract bug but the current production reachability is low. I'll note this calibration in the final report. The finding should be downgraded from high to medium (narrow/rare conditions) since the code path is currently unreachable from the CLI.

Now the synthesis is complete. Here is the final review report:

---

## Review: PR #1686 -- fix(service): add plist upgrade path for existing installs

**Decision: MERGE -- 0 blocking; 1 medium non-blocking, 6 low, 0 nit**

The PR cleanly addresses the Codex P1 finding: `upgradeService()` with atomic O_EXCL writes, `restart()` wired to upgrade-then-bootout/bootstrap, CLI `upgrade` subcommand, and postinstall delegation all work as described. The security posture is solid (O_EXCL tmp, 0o600 mode, no shell injection, xmlEscape on plist values). The plist rendering and ThrottleInterval addition are correct with snapshot coverage.

---

### Non-blocking

**F-1 · MEDIUM · correctness/api-compat** `src/service/launchd/manager.ts:133` · blocking:false
`restart()` calls `upgradeService(name)` with no opts, so `upgradeService` re-renders the plist without forwarding `opts.environment`. A plist containing custom env vars baked in at install time would be silently rewritten to strip them. Currently unreachable from the CLI (the `install` command never passes `environment`), but a latent contract bug for any programmatic caller.
*Waived: bounded, currently unreachable from production CLI, non-data-affecting -- env-var-bearing plists are a theoretical API path with no known caller.*
→ Forward `opts` through `restart()` to `upgradeService()`, or alternatively, have `upgradeService` preserve env vars it reads from the current plist rather than re-rendering from scratch.

**F-2 · MEDIUM (downgraded from medium, preserved) · correctness** `src/service/launchd/manager.ts:160-170` · blocking:false
When `upgradeService()` returns `{ kind: 'failed' }`, `restart()` silently falls through to `kickstart -k` and returns `{ kind: 'restarted' }` -- the operator gets no signal that the plist upgrade failed. Narrow (requires upgrade failure + successful kickstart), non-data-affecting.
*Waived: bounded -- the service still restarts correctly on the old plist, just without the upgraded config. Follow-up is landable.*
→ Branch on `upgradeResult.kind === 'failed'` before the kickstart fallback to surface the failure.

**F-3 · LOW · correctness** `scripts/postinstall.mjs:172` · blocking:false
`existsSync(cliMjs)` uses the module-scope import, not the injectable `existsFn`. The CLI-present/absent branch is untestable through the injection seam.
→ Replace with `existsFn(cliMjs)`.

**F-4 · LOW · api-compat** `src/service/types.ts:111` · blocking:false
`restart()` now carries an implicit upgrade side-effect with no JSDoc, and Darwin/Linux are behaviorally asymmetric at the interface layer.
→ Add JSDoc on `ServiceManager.restart()` noting the upgrade side-effect.

**F-5 · LOW · spec-compliance** `src/cli/commands/service.ts:180` · blocking:false
The upgrade subcommand prints "Run 'afk service restart'" post-upgrade, but `restart()` subsumes upgrade -- the message implies a two-step workflow.
→ Reword to "...to reload the service with the new config."

**F-6 · LOW · test-coverage** `tests/` · blocking:false
The new `restartFn` injectable and `cliMjs`-branch in `postinstall.mjs` have zero test coverage.
→ Add tests exercising both CLI-path and fallback-path branches via `restartFn` stub.

**F-7 · LOW · test-coverage** `src/service/launchd.test.ts` · blocking:false
No test covers the upgrade-failed -> kickstart fallback path in `restart()`. Only the already-current path is tested.
→ Add a `restart()` test mocking `upgradeService` returning `{ kind: 'failed' }`.

**F-8 · LOW · perf-observability** `src/service/launchd/manager.ts:133` · blocking:false
`upgradeService()` runs on every `restart()`, adding a synchronous `which afk` spawn + plist read for the common already-current case, with no observability.
→ Emit a debug-level line with `upgradeResult.kind` and elapsed ms.

---

### What was not checked
- Citations verified inline against branch HEAD `48b35854`.
- Stated intent: PR #1686 title+body -- spec-compliance assessed. All stated requirements have implementing changes; no scope creep detected.
- **Not checked:** Concurrent restart safety (two `restart()` calls racing through bootout+bootstrap). The O_EXCL guards upgrade atomicity but double-bootout behavior was not analyzed.
- **Not checked:** End-to-end integration path from `npm install -g` through postinstall -> CLI restart -> upgradeService -> bootout+bootstrap. Analysis is static only.
- **Not checked:** Full set of future plist key additions that would always trigger 'upgraded' via string comparison (semantic equivalence vs. byte-identical comparison).
- Security dimension: no issues found -- O_EXCL atomic writes, 0o600 permissions, xmlEscape applied, no shell injection vectors.

**Done**
- What was done: Full-regime code review of PR #1686 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability dimensions. 8 findings (0 blocking, 2 medium waived, 6 low).
- Evidence: Review report above; all citations verified against ref 48b35854.
- What changed: Nothing -- read-only review, no files modified.
- Deferred: None.

---

PR #1690, opened 2026-09-17
feat(repl): live tok/s momentum ticker in status line

griffinwork40:
## feat(repl): live tok/s momentum ticker in status line

Adds a live tokens-per-second counter to the REPL status line that appears during active model streaming and disappears between turns.

### What this adds

**New file: `src/cli/commands/interactive/momentum-ticker.ts`** (132 code lines)

A self-contained `MomentumTicker` class that:
- Tracks text-delta character counts during streaming (4 chars ≈ 1 token)
- Computes a smoothed tok/s estimate using an Exponential Moving Average (EMA, α=0.25) to avoid jitter
- Silently discards gaps longer than 3s (tool-use pauses) so those do not drag the rate down
- Fires a caller-supplied `onRepaint` callback at most once per 300ms — throttled internally, not on every chunk
- Clears the rate on `stop()` by firing `onRepaint(null)` so the status segment vanishes between turns

**`src/cli/status-line.ts`** — minimal additions:
- New `tokPerSec?: number` field on `StatusLineFields` (documents lifecycle: present during streaming, absent between turns)
- One render block in `formatLine`: renders as e.g. `847 tok/s` with `droppablePriority: 9` — sheds first on narrow terminals, never pushes model/mode off the line

**`src/cli/commands/interactive/shared.ts`** — one addition:
- `onTextDelta?(charCount: number): void` on `TurnHandles` — the bridge for the turn handler to feed character counts to the ticker

**`src/cli/commands/interactive/turn-handler.ts`** — one addition:
- `h.onTextDelta?.(event.chunk.content.length)` on each `content` chunk event — best-effort, absent on non-TTY callers

**`src/cli/commands/interactive/loop-iteration.ts`** — wiring:
- Constructs a `MomentumTicker` per turn with the repaint callback
- Starts ticker before `runTurn`, stops it in `onAfterTurn` (clearing the segment)
- Passes `onTextDelta` delegate to the turn handles

**New file: `src/cli/commands/interactive/loop-iteration.first-turn.ts`** — concern extraction:
- Moves `runFirstTurnHookIfNeeded` out of `loop-iteration.ts` (grandfathered at 433 code lines, cannot grow)

### Design decisions

- **EMA smoothing** over raw instantaneous rate — avoids jitter on small chunk sizes
- **4 chars/token** approximation — rough but fast, avoids actual tokenizer overhead during streaming
- **Throttled repaint in the ticker** — the MomentumTicker owns the repaint throttle so `loop-iteration.ts` stays minimal (a 1-liner delegate)
- **stop() fires repaint(null)** — status segment clears cleanly at turn end without an extra call site
- **droppablePriority: 9** — shed before activeAgentCount (8), ensuring the live rate is always the first field to go on narrow terminals

### CI gates

- `pnpm lint` (tsc strict): passes
- `pnpm audit:filesize:check` (350-line ceiling): passes — new files are 145 and 33 code lines; no grandfathered file grew
- `pnpm audit:chalk:check` (no raw chalk): passes — uses `palette.chrome()` throughout


griffinwork40 (commented):
## Review: feat(repl): live tok/s momentum ticker in status line

**Merge recommendation: ✅ Approve with notes** — code is clean, logic is correct, constraints are respected. Two stale comments are the only actionable items; both are cosmetic. Notes on test coverage gaps below.

---

### Summary

This PR adds a live tokens-per-second counter to the REPL status line. It also bundles four CI-fix commits for the Windows `--version` pipe-capture issue and a wider `watch.test.ts` timing window. The feature commit itself is the one to review for correctness; the CI fixes are straightforward and correct.

---

### Correctness

**`MomentumTicker` (`momentum-ticker.ts`)**

The EMA formula is correct: `instantRate = (pendingChars / CHARS_PER_TOKEN / elapsedMs) * 1_000` produces tok/s cleanly. The EMA application (`α * instant + (1 - α) * smoothed`) is standard.

One subtle but harmless accounting note: when a long gap fires (`elapsedMs > MAX_GAP_MS`), the characters accumulated since `lastSampleMs` are discarded along with the stale interval — `pendingChars` is zeroed before `return`. This is intentional (gap = "discard stale data") and documented in the inline comment. No issue.

The `stop()` path fires `onRepaint(null)` unconditionally, clearing `tokPerSec` on the status line. The `loop-iteration.ts` wiring places `momentumTicker.stop()` as the first call in `onAfterTurn()`, before `formatStatusFields` repaints — so the segment clears immediately at turn end and the subsequent `ctx.statusLine.repaint(formatStatusFields(...))` on line 686 will naturally omit `tokPerSec` (field is absent). There is a redundant repaint but it is harmless.

**`loop-iteration.ts` wiring**

`momentumTicker.start()` fires before `runTurn`; `momentumTicker.stop()` fires as the first thing in `onAfterTurn`. If `runTurn` throws (abort, model error), `onAfterTurn` is never called — the ticker is left `running = true` with a stale `lastRepaintMs`. This is benign: the next turn constructs a fresh `MomentumTicker` instance (`const momentumTicker = new MomentumTicker(...)` is inside the loop body), so no state leaks across turns. Worth noting but not a bug.

**`status-line.ts` droppability**

`droppablePriority: 9` is the highest assigned value, so `tok/s` sheds first on narrow terminals. The shedding algorithm drops the highest numeric priority first — `tok/s` (9) before `activeAgentCount` (8) before `budgetUsd` (7), etc. This matches the stated intent.

**`src/cli/index.ts` `--version` fix**

`writeSync(1, getVersion() + '\n')` followed by `return` inside the async IIFE is correct. `writeSync` is synchronous at the OS level — the data lands in the pipe before any async teardown. `return` from the async IIFE lets Node's event loop drain naturally, so PowerShell's pipe consumer sees the full buffer before the process exits. The three prior commits in the CI series document the progression from `console.log` → `process.exit(0)` → `fs.writeSync` + `process.exit(0)` → `fs.writeSync` + `return`. The final form is correct.

---

### Issues Found

#### Low — Stale comments (cosmetic, not behavioral)

**`src/cli/status-line.ts` L114 (JSDoc for `activeAgentCount`)**

The existing doc comment reads:
> *Takes droppablePriority 8 (shed before budgetUsd) — **the most peripheral field on the line**.*

This was accurate before the PR. After adding `tokPerSec` at priority 9, `activeAgentCount` is no longer the most peripheral field — `tokPerSec` is. The comment should read "sheds before `tokPerSec` (9)" rather than "most peripheral."

**`src/cli/status-line.ts` L674 (inline comment on `droppablePriority: 8`)**

The render-site comment reads:
> `droppablePriority: 8, // shed first — most peripheral; sheds before turnCount (6)`

Same issue: "shed first" is now wrong (9 sheds first). Should say "shed before tok/s (9)."

Both stale comments are in `status-line.ts` and can be fixed in a follow-up commit or here — either is fine since they do not affect behavior.

---

### Test coverage

**No tests were added for `MomentumTicker`.**

The `status-line.test.ts` regression guard for `droppablePriority` allocation (L764) and the existing "sheds fan-out before budget" test (L1497) both need updating to account for `tokPerSec` at priority 9 — the existing test comments claim `activeAgentCount` at priority 8 is "shed first," which is now incorrect.

Concretely missing:
- `MomentumTicker` unit tests (EMA math, gap-discard, throttle cadence, `stop()` clears state and fires `onRepaint(null)`, `update()` is a no-op when not running).
- A `status-line.test.ts` test that `tokPerSec` renders as `N tok/s` when set, is absent when `undefined` or `0`, and sheds before `activeAgentCount` on a narrow terminal (currently the existing narrow-terminal test at L1497 only checks that fan-out at priority 8 sheds — it would pass even if priority 9 was broken).

This is notable because `MomentumTicker` is a pure-logic class with no I/O — it is easy to unit-test. The PR description's CI section only lists `lint` and `audit:filesize:check`; `pnpm test` is not listed as a gate it ran.

**Recommendation:** Add `src/cli/commands/interactive/momentum-ticker.test.ts` and extend the narrow-terminal shedding test in `status-line.test.ts` before merge, or immediately after in a follow-up.

---

### Concern extraction (`loop-iteration.first-turn.ts`)

The extraction is clean. The new file preserves the behavioral invariant (single-fire guard on `totalTurns === 0`, detach-before-await), and the JSDoc carries the key constraint ("Must be called before `runTurn` ... so worktree-creating hooks run before any cwd-sensitive operations"). The original inline comment about the compositor not being armed ("no clear/repaint race to lose") was dropped — that detail is low-value in the extracted file since the completionWriter call is unchanged.

---

### Minor observations

- `getCurrentRate()` is exported but never called by the loop (the callback pattern is used instead). Not a problem — it's a reasonable API seam for testing or future callers.
- The `onRepaint` callback in `loop-iteration.ts` uses `...formatStatusFields(...)` spread, which recomputes all other status fields on every tick. At 300ms throttle this is cheap, but if `formatStatusFields` becomes expensive, this is the spot to cache the last snapshot. Not actionable now.
- `pendingRoundSeam` and `streamingStarted` combined into one declaration (`let pendingRoundSeam = false, streamingStarted = false;`) in `turn-handler.ts` — minor style change, no functional impact.

---

### Files reviewed

| File | Verdict |
|------|---------|
| `src/cli/commands/interactive/momentum-ticker.ts` | ✅ Correct |
| `src/cli/commands/interactive/loop-iteration.ts` | ✅ Correct |
| `src/cli/commands/interactive/loop-iteration.first-turn.ts` | ✅ Clean extraction |
| `src/cli/commands/interactive/shared.ts` | ✅ Correct |
| `src/cli/commands/interactive/turn-handler.ts` | ✅ Correct |
| `src/cli/status-line.ts` | ✅ Correct behavior, ⚠️ stale comments on priority 8 |
| `src/cli/index.ts` | ✅ Correct (`writeSync` + `return`) |
| `src/telegram/watch.test.ts` | ✅ Wider window is appropriate |

PR #1691, opened 2026-09-17
feat(repl): 150ms glyph flash on tool completion

griffinwork40:
## Summary

Implements the "Lane Flash" UX feature: when a tool call completes, its glyph in the tool lane briefly pulses bold for 150ms, then returns to normal rendering. This gives visceral, kinetic feedback that a tool landed — your peripheral vision catches the flash without requiring you to read text.

## How it works

1. `ToolLaneFlash` (new file) tracks which tool-use IDs are currently flashing via a `Set<string>`. Each flash entry has a 150ms `setTimeout` that removes it and triggers a repaint.

2. `ToolLane.addResult()` calls `this.flash?.flashTool(toolUseId)` when a result arrives. The flash-start repaint happens naturally (the caller always calls `setComposedOverlay` after `addResult`); the flash-end repaint fires via `ToolLaneFlash.onRepaint`.

3. `ToolLane.getOverlay()` checks `this.flash?.isFlashing(entry.toolUseId)` when rendering completed flat-root entries and wraps the card with `palette.bold()` when true.

4. `StreamRenderer.arm()` wires a `ToolLaneFlash` instance with a `deferFlush('tool-lane')` callback (same deferred-flush pattern as `setInterrupting`/`setSoftStopping` to avoid double-setOverlay races).

5. `StreamRenderer.dispose()` calls `toolLaneFlash.dispose()` to cancel any pending 150ms timers.

## Files changed

- **new**: `src/cli/commands/interactive/tool-lane-flash.ts` — `ToolLaneFlash` class (28 code lines)
- **modified**: `src/cli/commands/interactive/tool-lane.ts` — flash field, flashTool call in addResult, bold-wrap in getOverlay; also extracts `buildChildMap` private method to keep under ratchet (-10 net code lines from 481 baseline → 471)
- **modified**: `src/cli/commands/interactive/tool-lane-render.ts` — receives extracted `buildChildMap` as a free function (156 → 173 code lines)
- **modified**: `src/cli/_lib/stream-renderer.ts` — flash field, arm() wiring, deferFlush refactor; also extracts repeated `markDirty+setTimeout` pattern into `deferFlush()` private method, applied to setInterrupting/setSoftStopping/setBashOutputTail (346 → 341 code lines, net reduction)
- **modified**: `src/cli/_lib/stream-renderer-dispose.ts` — adds `toolLaneFlash` to DisposeCtx, calls dispose() in Phase 7

## CI gates

- `pnpm lint` passes (tsc strict)
- `pnpm audit:filesize:check` passes (all files ≤ 350 code lines; tool-lane.ts reduced from 481 → 471)
- `pnpm audit:chalk:check` passes (no raw chalk; flash uses palette.bold)
- All 19,675 tests pass

## Design decisions

**Option B (ANSI bold pulse)** was chosen: wrap the rendered `toolCard` string with `palette.bold()` for 150ms. No new palette role needed — bold on an already-colored card makes it visually "pop" without requiring a separate theme entry in each of dark/light/umber themes. The glyph (✓/✗/⊘) plus tool name go bold simultaneously, which is more visually salient than changing color alone.

The flash only applies to completed **flat-root** entries in the overlay (the `entry.result && !NESTING_TOOLS` branch). NESTING_TOOLS (agents, skills, compose) use a different render path and do not flash — their completion is typically announced via a larger structural change in the overlay (subagent summary block). This is the correct scope: flat-root tool completions (bash, read_file, etc.) are the high-frequency case where peripheral-vision feedback matters most.


griffinwork40:
## agent-afk code review · PR #1691 · feat(repl): 150ms glyph flash on tool completion

**Reviewed ref:** `9970d3f5d22bd305d8897d0a42e4d071a3be8086`
**Change type:** feature
**Regime:** full (445 diff lines, 7 files, 1 new file)

---

### Summary

Adds a 150ms bold-glyph pulse to flat-root tool completions in the REPL overlay. The core mechanic is clean: a `ToolLaneFlash` class tracks flashing IDs with a `Set<string>` and per-ID `setTimeout`s; `ToolLane.getOverlay()` wraps the rendered card in `palette.bold()` during the flash window. The PR also includes two unrelated but useful clean-ups: the repeated `markDirty+setTimeout` pattern is extracted into a private `deferFlush()` helper, and `buildChildMap` is promoted from a private method to a free function to make room under the 350-line ratchet. The `src/cli/index.ts` and `src/telegram/watch.test.ts` changes are independent fixes bundled into this PR.

---

### Findings

**MEDIUM · blocking:false · waived: bounded non-data-affecting defect that can land and be followed up · confidence:high · dimension:correctness · spec-compliance**

`src/cli/_lib/stream-renderer.ts:375` · `src/cli/commands/interactive/tool-lane-flash.ts` · diff-context

**Flash fires on `reducedMotion` surfaces.** `ToolLaneFlash` is created unconditionally in `arm()` at line 375 regardless of `this.reducedMotion`. The spinner is correctly gated (`compositor.setSpinner({ enabled: !this.reducedMotion })`), but the 150ms bold pulse is not. A user who has set the OS `prefers-reduced-motion` accessibility preference — or who passes `reducedMotion: true` via `StreamRendererOpts` — still gets the animation flash. The stated intent ("gives visceral, kinetic feedback") directly targets users who explicitly opt out of animation.

Evidence:
```ts
// stream-renderer.ts arm()
this.toolLaneFlash = new ToolLaneFlash(() => this.deferFlush('tool-lane'));
this.toolLane.flash = this.toolLaneFlash;

// ...one line later:
compositor.setSpinner({ enabled: !this.reducedMotion, rotateVerbEveryMs: 3500 });
```

Suggestion: `this.toolLane.flash = this.reducedMotion ? null : this.toolLaneFlash;` — set `ToolLane.flash` to `null` when reduced-motion is on so the `?. `chain in `addResult` and `getOverlay` is a no-op. The `ToolLaneFlash` instance itself need not be guarded at construction.

---

**LOW · blocking:false · confidence:high · dimension:correctness · ref:9970d3f · citation-type:diff-context**

`src/cli/commands/interactive/tool-lane-render.ts:460` · diff-context

**Dead ternary in `buildChildMap`.** The extracted free function has `const ctx = entry.kind === 'tool' ? entry.agentContext : entry.agentContext;` — both branches resolve to `entry.agentContext`. This was copied verbatim from the private method on `main` (verified at `main:src/cli/commands/interactive/tool-lane.ts` private `buildChildMap`). It compiled and ran correctly before and still does; TypeScript does not flag it. The ternary is misleading because it implies `TextEntry` could yield a different context. Either simplify to `const ctx = entry.agentContext;` or, if the intent was to handle `TextEntry` differently, the correct field name for that branch is different.

Evidence:
```ts
const ctx = entry.kind === 'tool' ? entry.agentContext : entry.agentContext;
```

Suggestion: `const ctx = entry.agentContext;` — removes the false implication while preserving behavior exactly.

---

**LOW · blocking:false · confidence:medium · dimension:test-coverage · ref:9970d3f · citation-type:diff-context**

`src/cli/commands/interactive/tool-lane-flash.ts` · diff-context

**No unit tests for `ToolLaneFlash`.** The new class has two stateful paths that have behavior worth pinning: (a) `flashTool` re-entry cancels the prior timer so `onRepaint` fires exactly once per flash window, and (b) `dispose()` prevents `onRepaint` from firing after teardown. The existing `tool-lane.test.ts` and `tool-lane.overlay.test.ts` do not cover flash (verified: zero hits for `flash`/`Flash`/`bold` in those files at reviewed ref). The `ToolLane.flash` field is `null` by default on non-TTY surfaces and in tests, so the integration path is invisible to the existing suite. The risk is narrow — the class is simple — but `dispose()` safety is exactly the kind of timing invariant that regresses silently (a timer firing post-teardown produces no visible output, just a potential `deferFlush` on a disposed renderer).

Suggestion: Add `src/cli/commands/interactive/tool-lane-flash.test.ts` covering: `isFlashing` returns true within window, false after expiry; `onRepaint` fires exactly once per expiry; re-flash restarts the timer; `dispose()` prevents `onRepaint` from firing; `flashTool` is a no-op after `dispose()`.

---

**NIT · blocking:false · confidence:high · dimension:correctness · ref:9970d3f · citation-type:diff-context**

`src/cli/_lib/stream-renderer.ts:429-435` · diff-context

**`deferFlush` silently no-ops before `arm()`.** The extracted helper guards `if (!this.overlayComposer) return;`. This is the correct behavior — callers that run before `arm()` (or on non-TTY surfaces) should be no-ops — but the old inline code had the same guard more explicitly as `if (this.overlayComposer) { ... }`. The new form is equivalent. Document the early-return semantics in the JSDoc (currently documents the double-setOverlay motivation but not the pre-arm no-op contract) so the next reader of `setInterrupting` and `setSoftStopping` does not need to trace through the callee to understand why they are no-ops pre-arm.

---

**NIT · blocking:false · confidence:high · dimension:spec-compliance · ref:9970d3f · citation-type:diff-context**

`src/cli/index.ts:316-322` · diff-context

**Comment update does not fully match behavior.** The new comment says "even with `fs.writeSync`, because PowerShell's pipeline teardown races with the child process exit" — but the old approach *was* using `console.log`, not `fs.writeSync`. The diff shows this PR *introduces* `writeSync` as the write mechanism and simultaneously removes `process.exit(0)`. The comment is explaining why `writeSync` alone is insufficient (which justifies the `return` instead of `process.exit`), but a reader unfamiliar with the old approach may interpret it as "we tried `writeSync` + exit and it still failed." Tightening the comment: "We tried `console.log` + `process.exit(0)` (prior approach) — on Windows PowerShell pipes, stdout can still be dropped before the pipe reads it. Switching to `writeSync` + `return` lets Node drain naturally."

---

**Stated intent covered:** the 150ms bold-glyph pulse lands on flat-root completions via `palette.bold(card)` in `getOverlay()`. The PR correctly scopes the flash to flat-root entries only (NESTING_TOOLS use a different render path). `dispose()` cancels pending timers at teardown. The `deferFlush` pattern matches the existing double-setOverlay race avoidance used by `setInterrupting` and `setSoftStopping`.

**No unmet intent.** All five design points from the PR body are implemented.

**Possible scope creep:** the `src/cli/index.ts` Windows/PowerShell fix and the `src/telegram/watch.test.ts` timing fix are unrelated to the lane-flash feature. They appear intentionally bundled (the PR body lists them). Not flagged as a blocker — both changes are correct and beneficial — but noted for reviewers who track PR scope.

---

### What was not checked

- Citations verified inline against branch HEAD `9970d3f5d22bd305d8897d0a42e4d071a3be8086`. All citations are `diff-context` or verified against pre-fetched file content. No fabricated citations found.
- Did not run tests locally — per PR body "All 19,675 tests pass" and CI gates are listed as green.
- Did not review `src/telegram/watch.test.ts` against the Telegram watch internals; the timing increase from 400ms→800ms and 150ms→300ms appear correct given the stated rationale (250ms poll interval + CI jitter).
- Reduced-motion detection implementation (`detectReducedMotion()`) was not reviewed — the finding is about the wiring gap in `arm()`, not the detection logic.
- Stated intent: PR #1691 title + body — spec-compliance assessed.

---

### Decision

**MERGE — 0 blocking (1 medium waived, 2 low, 2 nit)**

The medium finding (flash fires on reduced-motion surfaces) is waived: it is a bounded accessibility gap with no data impact, no broken behavior for the majority of users, and a one-line fix appropriate for a follow-up. The core mechanic is correct, the dispose lifecycle is sound, and the `deferFlush` refactor is a clean win.

PR #1697, opened 2026-09-17
fix(spine): widen /spine init grep to cover all documented comment prefixes and file types

griffinwork40:
## What

Fixes three bugs in `/spine init`'s seed-material grep that caused it to silently miss architectural annotations, and replaces GNU grep with ripgrep for a ~7,500x speedup.

### 1. File-type filter too narrow
`--include=*.ts` skipped `*.js`, `*.mjs`, and `*.md` files. 18 files with `Invariant:`/`Contract:` hits were invisible to init -- notably `docs/file-size-ceiling.md` (9 hits) and build scripts in `scripts/`.

### 2. Missing `History:` prefix
The long-comment convention (AFK.md) defines three prefixes: `Invariant:`, `Contract:`, and `History:`. Init only scanned the first two, missing decision-log and postmortem comments throughout `src/` and `docs/`.

### 3. GNU grep scanned node_modules (3m28s -> 12ms)
The grep ran from `repoRoot` with no directory exclusions, recursing into `node_modules/`, `dist/`, and `.afk-worktrees/`. Replaced GNU `grep -rE` with `rg` (ripgrep), which respects `.gitignore` automatically -- no `--exclude-dir` flags needed. On this repo: 3m28s -> 12ms.

## Change

Single hunk in `src/cli/slash/commands/spine.ts` -- the grep invocation in `handleInit`:

```diff
- execFileSync('grep',
-   ['-rE', '--include=*.ts', '-h', 'Invariant:|Contract:', repoRoot],
-   { encoding: 'utf8', stdio: ['ignore', 'pipe', 'ignore'] },
- );
+ execFileSync('rg',
+   ['--no-filename', '-t', 'ts', '-g', '*.js', '-g', '*.mjs', '-g', '*.md',
+    'Invariant:|Contract:|History:', repoRoot],
+   { encoding: 'utf8', maxBuffer: 512 * 1024, stdio: ['ignore', 'pipe', 'ignore'] },
+ );
```

Also retains the buffer cap (512KB) and per-line truncation (200 chars) from the earlier commit.

## Verification

- `pnpm lint` clean
- `rg` invocation returns 51 hits across all file types in 12ms

PR #1698, opened 2026-09-17
fix(spine): stream and bound init annotation matches

griffinwork40:
### Motivation

- Prevent `grep` output from overflowing Node's subprocess buffer when `/spine init` scans large repos by avoiding full-process buffering.  
- Ensure the annotation scan covers all documented comment prefixes and common file types while excluding noisy directories.  

### Description

- Replace the buffered `execFileSync` scan with a streaming reader using `spawn('grep', ...)` and `createInterface` to read lines as they arrive.  
- Add a new helper `collectGrepMatches(repoRoot, limit)` that collects up to `limit` non-empty matches and kills the `grep` process once the bound is reached.  
- Expand the grep arguments to include `--include=*.ts`, `--include=*.js`, `--include=*.mjs`, `--include=*.md`, `--exclude-dir=node_modules`, `--exclude-dir=dist`, `--exclude-dir=.afk-worktrees`, and the `History:` prefix in the pattern.  
- Preserve the previous graceful fallback so that absence of `grep` or no matches still continues without throwing.  

### Testing

- Ran `pnpm lint`, which completed successfully (with a local environment warning about Node version mismatch).  
- Ran `pnpm vitest run` for the spine tests (`src/agent/spine/spine-store.test.ts` and `src/agent/spine/spine-hook.test.ts`), and those tests passed.  
- The `spine-classifier` test run triggered a worker exit / Vitest pool configuration error under the current Node 20 environment, causing an unhandled error; the project declares `node >=22.0.0`, so test issues appear to be due to the unsupported Node runtime rather than this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6aabd98b0674832bb15e9a9c79a04048)


griffinwork40:
Superseded -- addressed both Codex P2s (buffer overflow + unbounded line length) directly on #1697 with `maxBuffer: 512KB` and per-line `.slice(0, 200)`. Simpler than the streaming rewrite and covers both concerns.

PR #1703, opened 2026-09-17
feat(review): offer to merge docs/test-only PRs after MERGE verdict

griffinwork40:
## Summary

When `/review` finishes with a **MERGE** verdict and the only surviving findings are docs or test-coverage, the agent now offers to merge the PR for you with a simple confirm prompt.

## Changes

**`src/bundled-plugins/awa-bundled/skills/review/SKILL.md`**
- Added a narrow exception to the read-only hard constraint: `gh pr merge` is permitted after a MERGE verdict on docs/test-only PR reviews, but only after explicit user confirmation via `ask_question`.
- Added a **Merge offer** block after the terminal-step paragraph. The offer fires when all three hold:
  1. Decision is MERGE
  2. Target was a PR (URL or number)
  3. Every surviving finding is `test-coverage` or documentation-only -- no `security`, `correctness`, `api-compat`, `perf-observability`, or `spec-compliance` findings survived

On confirm, runs `gh pr merge <pr-ref>` with no strategy flag (uses the repo's configured default merge method). On decline/cancel/skip, stops as before.

**`src/bundled-plugins/awa-bundled/bundled.test.ts`**
- Updated SHA-256 pin for the review skill.

## Motivation

When a review comes back clean with only test/doc suggestions, the user currently has to go merge the PR manually. This adds a one-question shortcut that saves a context switch.

PR #1704, opened 2026-09-17
fix(review): use repo default merge strategy instead of hardcoding --squash

griffinwork40:
The merge offer added in #1703 hardcoded `--squash`, which silently overrides the repo's configured default merge method. A user confirming "merge" on a repo that prefers rebase or merge-commit would get squash instead.

Now queries `viewerDefaultMergeMethod` via `gh repo view` and passes the corresponding flag (`--squash`, `--rebase`, or `--merge`). An explicit strategy flag is still required because `gh pr merge` errors in non-interactive contexts without one.

Found via `/review` self-review of #1703, then `/shadow-verify` on the analysis -- which refuted the initial assumption that `--squash` was cosmetic when the repo default is squash. It's not: `gh pr merge` without a strategy flag fails outright in non-interactive contexts.


griffinwork40:
# Review: PR #1704

## Summary

The change correctly replaces the hardcoded `--squash` behavior with the target repository’s configured `viewerDefaultMergeMethod`, while retaining the explicit strategy flag required for non-interactive `gh pr merge`.

### Findings

- **Security:** No issues found.
- **Correctness:** No issues found. The skill explicitly derives `<owner>/<repo>` from the PR reference rather than the local checkout.
- **API compatibility:** No issues found. A read-only query confirmed `viewerDefaultMergeMethod` is supported by the installed GitHub CLI and returns `SQUASH` for `griffinwork40/agent-afk`.
- **Spec compliance:** The implementation satisfies the PR’s stated intent without unrelated behavior changes.
- **Test coverage:** The updated pinned hash is consistent with the repository’s established guard for bundled skill changes.
- **Performance and observability:** No substantiated issue found.

Wave 1 proposed three findings, but inline verification dropped them as unsupported: hypothetical executor noncompliance, a runtime-test requirement for model-executed prose, and an unevidenced latency estimate.

**Decision: MERGE — 0 findings.**

## What was not checked

- Tests were not run because this review was read-only.
- The merge workflow was not exercised against a live PR.
- GitHub Enterprise compatibility was not independently tested.
- Citations were verified against reviewed ref `d49955f6f2aad777f7c45bd8cdab8c8b5cc3cb23`.
- Stated intent came from PR #1704’s title and body and was assessed.

## Done
- What was done: Completed a read-only review of PR #1704 and recommended merge with zero surviving findings.
- Evidence: PR head `d49955f6f2aad777f7c45bd8cdab8c8b5cc3cb23` and diff artifact `/Users/griffinlong/.afk/state/skill-preflight/31110dda-c4ec-4bba-9e46-42c2fe51912e/pr-1704.diff`.
- What changed: No repository, PR, or external state was modified.
- Deferred: Runtime tests and the merge itself were outside this review’s read-only scope.

---

PR #1711, opened 2026-09-17
fix(service): forward opts to upgradeService in restart() (#1694)

### LINKED ISSUE #1694: fix(service): restart() silently strips custom environment vars during upgrade
## Problem

`restart()` in `src/service/launchd/manager.ts:133` calls `upgradeService(name)` with no opts, so `upgradeService` re-renders the plist without forwarding `opts.environment`. A plist containing custom env vars baked in at install time would be silently rewritten to strip them before the bootout+bootstrap cycle runs.

Currently unreachable from the CLI (`install` never passes `environment`), but a latent contract bug for any programmatic caller of the `ServiceManager` interface.

## Suggested fix

Either:
1. Forward `opts` through `restart()` to `upgradeService(name, opts)`, or
2. Have `upgradeService` preserve env vars it reads from the current plist rather than re-rendering from scratch (more robust against future callers)

## Origin

PR #1686 review finding F-1 (medium/api-compat, non-blocking).

griffinwork40:
Addresses #1694 -- `restart()` silently strips custom environment vars during upgrade.

## Root cause

`launchd/manager.ts` `restart()` called `upgradeService(name)` with no second argument, so `upgradeService` always rendered the plist with only the default `PATH` -- any custom `environment` vars the caller intended to preserve were silently dropped.

## Changes (3 files, 4 insertions / 4 deletions)

| File | Change |
|---|---|
| `src/service/types.ts` | Added `opts?: ServiceInstallOptions` to `restart()` on the `ServiceManager` interface |
| `src/service/launchd/manager.ts` | Accept `opts?` and forward it as `upgradeService(name, opts ?? {})` |
| `src/service/systemd/manager.ts` | Accept the new parameter as `_opts?` (unused -- systemd restart doesn't re-render the unit file) |

All 61 service tests pass.

Closes #1694


griffinwork40 (commented):
**Review.** Clean, minimal fix for a genuine silent-data-loss bug.

**Root cause confirmed.** `restart()` called `upgradeService(name)` with no second argument, so `upgradeService` always rendered the plist with only the default `PATH` — any custom `environment` vars passed by the caller were silently dropped. The fix is exactly the right one-liner: forward `opts ?? {}` so an undefined `opts` still satisfies the `InstallOptions = {}` default in `upgradeService`.

**Interface change is correct.** Adding `opts?: ServiceInstallOptions` to `ServiceManager.restart()` keeps the parameter optional, so all existing call sites (`service.ts` CLI `restart` command) remain valid without changes. Checked `src/cli/commands/service.ts:193` — `mgr.restart(name)` still compiles cleanly.

**systemd stub is correct.** The `_opts?` prefix signals intent (unused) and satisfies the interface without breaking `tsc --noEmit`. Since systemd's `restart` runs `systemctl --user restart` directly rather than re-rendering the unit file, silently accepting but ignoring `opts` is the right behaviour — consistent with the existing comment that in-place upgrade isn't yet supported on that backend.

**One small gap to be aware of** (not a blocker): the CLI `restart` subcommand (`src/cli/commands/service.ts` line 187) doesn't expose `--environment` flags today, so end-users can't pass custom env vars through the CLI. That's a pre-existing omission out of scope here — the interface fix correctly unblocks any programmatic caller (a plugin, the daemon restart hook, etc.) that passes `opts` directly.

All 61 service tests pass. **Looks good to merge.**

PR #1712, opened 2026-09-17
fix(tui): lane flash reduced-motion gate, tests, dead ternary (#1693)

### LINKED ISSUE #1693: fix: lane flash follow-ups — reduced motion gate, tests, dead ternary (#1691)
## Context

PR #1691 added `ToolLaneFlash` (`src/cli/commands/interactive/tool-lane-flash.ts`) — a 28-line flash state tracker for tool completion glyph pulses. Review flagged three follow-ups.

## 1. Reduced motion gate (medium)

`ToolLaneFlash` is wired unconditionally in `stream-renderer.ts` `arm()` at line 375, even when `this.reducedMotion` is true. The spinner is correctly gated; the flash should be too.

**Fix** (one-liner in `stream-renderer.ts`):
```ts
this.toolLane.flash = this.reducedMotion ? null : this.toolLaneFlash;
```

## 2. Unit tests for ToolLaneFlash

No tests pin the core invariants:
- `flashTool(id)` adds to set, `isFlashing(id)` returns true, returns false after 150ms
- Re-flash on same ID cancels the prior timer (no double-fire)
- `dispose()` clears all pending timers and prevents post-dispose repaint callbacks
- `onRepaint` fires on both flash-start (implicit via caller) and flash-end (timer expiry)

## 3. Dead ternary in buildChildMap (pre-existing, cosmetic)

In `tool-lane-render.ts`, the extracted `buildChildMap` contains:
```ts
entry.kind === 'tool' ? entry.agentContext : entry.agentContext
```
Both branches are identical. Pre-existing on main, just copied during extraction. Simplify to `const ctx = entry.agentContext`.

griffinwork40:
Addresses #1693 -- three follow-ups from PR #1691 (ToolLaneFlash).

### 1. Reduced motion gate
Wrapped `ToolLaneFlash` construction in `if (!this.reducedMotion)` in `stream-renderer.ts`, matching the same gate pattern used for the spinner. When reduced motion is enabled, no timed flash animation fires.

### 2. Dead ternary fix
Simplified `entry.kind === 'tool' ? entry.agentContext : entry.agentContext` to `entry.agentContext` in `tool-lane-render.ts`. Both branches were identical.

### 3. Unit tests (12 tests)
New `tool-lane-flash.test.ts` covering:
- flashTool(id) immediately makes isFlashing(id) true
- After FLASH_DURATION_MS (150ms), isFlashing returns false
- Re-flash cancels prior timer and restarts the window
- onRepaint fires once per expiry, never after dispose()
- Multiple distinct ids have independent windows
- dispose() clears all timers and is idempotent

Lint clean, all 12 tests pass.

Closes #1693


griffinwork40 (commented):
**Approved.** All CI passes (lint, build, 3-platform tests, PTY suite, docs). Clean change -- reviewed all three modified files and the new test suite.

### Reduced-motion gate (`stream-renderer.ts`)

The `if (!this.reducedMotion)` guard mirrors exactly how the spinner is gated two lines below, so the pattern is consistent. `toolLaneFlash` was already typed `ToolLaneFlash | null` and `toolLane.flash` accepts `null`, so no callers break. The `dispose()` path in `stream-renderer-dispose.ts` already uses optional-chaining (`ctx.toolLaneFlash?.dispose()`), so it is safe when the field stays `null` for the lifetime of the session.

One thing worth noting: the existing comment on line 380 still reads "Reduced-motion suppresses the spinner ticker at the source" -- it no longer mentions the flash gate. Not a blocker, but worth a one-line addition on a follow-up.

### Dead ternary (`tool-lane-render.ts`)

`buildChildMap` iterates over `Entry`, which is `ToolEntry | TextEntry`. Both arms of the deleted ternary referenced `entry.agentContext` identically -- `ToolEntry.agentContext` is optional (`string | undefined`), `TextEntry.agentContext` is required (`string`), so the ternary was always selecting the same value regardless of branch. The simplification to `entry.agentContext` is correct and the `if (!ctx) continue` guard below handles the `undefined` case for `ToolEntry`.

### Test suite (`tool-lane-flash.test.ts`)

12 tests, all behaviorally distinct and well-named. Highlights:

- The re-flash / timer-cancel test ("cancels the prior timer and restarts the window") is the most important case and it verifies both the flashing-state edge and that `onRepaint` is not called prematurely -- good coverage of the `clearTimeout` branch.
- The staggered-window test ("multiple distinct ids each have independent flash windows") uses exact time offsets (50ms stagger, then checks at `FLASH_DURATION_MS - 50` and `FLASH_DURATION_MS`) to confirm isolation -- this would have caught a shared-timer bug.
- `dispose()` idempotency and the post-dispose `flashTool` no-op are tested explicitly, which matters because teardown order in a live session is not guaranteed.
- The one thing not covered: `onRepaint` throwing -- the class has no try/catch around the callback. Not a regression (it was not tested before), but `onRepaint` is caller-supplied and the JSDoc says "must not throw". A follow-up could add a guard or at least a test asserting the behavior.

Solid, ready to land.

PR #1714, opened 2026-09-17
fix(service): surface upgradeService failure in restart() (#1695)

### LINKED ISSUE #1695: fix(service): restart() swallows upgradeService failure silently
## Problem

In `src/service/launchd/manager.ts:160-170`, when `upgradeService()` returns `{ kind: 'failed' }` (e.g. unwritable tmp dir, `resolveProgramArguments` throws), `restart()` silently falls through to `kickstart -k` and returns `{ kind: 'restarted' }`. The operator gets no signal that the plist upgrade failed -- the service restarts on the old config.

The code comment explicitly acknowledges this: `"Plist unchanged (already-current), not installed, or upgrade failed -- fall back to kickstart"`.

## Suggested fix

Branch on `upgradeResult.kind === 'failed'` before the kickstart fallback. Either:
- Return `{ kind: 'failed', reason: upgradeResult.reason }` to propagate the failure, or
- Proceed with kickstart but surface a warning (e.g. widen `ServiceRestartOutcome` to include a `notes` field)

## Origin

PR #1686 review finding F-2 (medium/correctness, non-blocking).

griffinwork40:
Addresses #1695 -- `restart()` swallows upgradeService failure silently.

## Root cause

In `launchd/manager.ts`, `restart()` had a single fallthrough block for "Plist unchanged, not installed, or upgrade failed". When `upgradeResult.kind === 'failed'`, it silently proceeded to kickstart and returned `{ kind: 'restarted' }` with no indication of the failure.

## Changes

| File | Change |
|---|---|
| `src/service/types.ts` | Widened `ServiceRestartOutcome`'s `restarted` variant with optional `notes?: string[]` |
| `src/service/launchd/manager.ts` | Explicit check for `upgradeResult.kind === 'failed'` -- captures failure reason into notes, still proceeds with kickstart |
| `src/cli/commands/service.ts` | Prints each note from `result.notes` using `palette.warning` after the success line |
| `src/service/launchd.test.ts` | New test: upgrade-failed-but-kickstart-succeeded returns notes with warning |

The service still restarts (kickstart proceeds regardless), but the operator now sees a warning that the plist upgrade failed and the service is running with stale config.

All service tests pass.

Closes #1695


griffinwork40:
## PR #1714 Review — `fix(service): surface upgradeService failure in restart() (#1695)`

**Target:** `f0beaa6a4610c9be929c840e4fd7142575cf36cd`  
**Regime:** light (hotfix, 66 lines, 4 files)  
**Stated intent:** `restart()` was swallowing `upgradeService` failures silently; this PR captures the failure reason as a `notes` warning on the `restarted` outcome and surfaces it in the CLI output.  
**Spec-compliance:** assessed — intent is fully addressed.

---

### Findings

#### NIT · blocking:false · correctness · `src/service/launchd/manager.ts:169` · ref:f0beaa6a

**Finding:** The fallthrough comment at line 169 says "Plist unchanged (already-current), **not installed**, or upgrade failed" but the `not-installed` case from `upgradeService()` is unreachable here — `isInstalled()` gates `restart()` at line 104 before `upgradeService` is ever called, so `upgradeResult.kind === 'not-installed'` cannot occur in this path.

**Evidence (diff-context):**
```ts
// line 169 (manager.ts, PR branch)
// Plist unchanged (already-current), not installed, or upgrade failed —
// fall back to kickstart -k for a lighter-weight process restart.
```

**Suggestion:** Drop "not installed, or" from the comment. The accurate description is "Plist unchanged (already-current) or upgrade failed — fall back to kickstart -k."

---

### What was not checked

- Citations verified inline against branch HEAD `f0beaa6a4610c9be929c840e4fd7142575cf36cd`.
- Stated intent: PR #1714 title + body — spec-compliance assessed.
- Did not review Telegram surface or daemon surface consumers of `ServiceRestartOutcome` (neither is modified by this PR and no production callers of `.notes` besides `service.ts` were found).
- `systemd/manager.ts` intentionally omitted — its `restart()` method returns `{ kind: 'restarted', label }` with no `notes` field, which is fine: the `notes` field is optional in the type and the systemd backend has no upgrade step to fail. The PR correctly limits the change to the launchd path.

---

### Decision

**MERGE — 0 blocking (1 nit).**

The fix is correct, well-scoped, and properly tested. The new test exercises the exact failure path described in #1695: `upgradeService` fails (fake entrypoint not on disk), kickstart still runs, and the returned `notes` field is asserted. The `notes?: string[]` type widening on `ServiceRestartOutcome` is additive and backward-compatible — the systemd backend's `{ kind: 'restarted', label }` continues to typecheck without `notes`. The CLI rendering in `service.ts` uses `result.notes ?? []` which handles the undefined case cleanly. The only finding is a stale comment (nit, non-blocking).

PR #1755, opened 2026-09-18
refactor: complete #832 Tier 3 extractions -- dispatcher, session, scheduler, daemon

griffinwork40:
## Summary

Completes the remaining Tier 3 extractions from #832. Four files that were over the 350-code-line ceiling (or stuck at baseline) are now either under the ceiling or significantly ratcheted down, with concerns cleanly separated into sibling modules.

### dispatcher.ts: 633 -> 335 code lines (retired from baseline)
Two extraction passes:
1. **`dispatcher.pre-dispatch-gates.ts`** -- the 11-method pre-dispatch gate chain (~175 code lines) as free functions with `PreDispatchGateDeps`/`PreDispatchGateMutableState` interfaces, following the existing `BatchExecDeps` pattern.
2. **`dispatcher.core-exec.ts`** -- the core execution concern (handler routing, PostToolUse dispatch, output cap) as free functions with `CoreExecDeps`.

### agent-session.ts: 777 -> 422 code lines (ratcheted, still baselined)
Six concerns extracted into sibling modules:
- **`accounting-accumulator.ts`** -- 10 private accounting fields + mutation methods (`AccountingAccumulator` class)
- **`turn-stream-runner.ts`** -- provider event loop, turn execution, transform deps
- **`session-shutdown.ts`** -- once-guarded session-end dispatch
- **`provider-lifecycle.ts`** -- provider selection, query construction, initialization polling
- **`session-send.ts`** -- sendMessage/sendMessageStructured/sendMessageStream/interrupt
- **`session-reset.ts`** -- reset lifecycle teardown+rebuild
- **`provider-passthrough.ts`** -- 9 thin provider query delegations

### scheduler.ts: 395 -> 321 code lines (retired from baseline)
- **`scheduler.pull-tick.ts`** -- `pullTick` body, `sweepAnsweredHandoffs` (consolidated from 4 copy-pasted blocks), and `fireOnTaskComplete`.

### daemon.ts: 400 -> 304 code lines (retired from baseline)
- **`daemon-session-factory.ts`** -- `buildDaemonSessionFactory` + `BuildDaemonSessionFactoryOpts` (the test file already existed expecting this sibling).

## Baseline changes
- `.filesize-baseline.json`: 3 files retired (dispatcher.ts, scheduler.ts, daemon.ts); 2 ratcheted (dispatcher.ts 633->335 before retirement, agent-session.ts 777->422)
- `.funcsize-baseline.json`: 1 stale entry retired (`createTranscriptModel`)
[...]

PR #1774, opened 2026-09-18
bench: tighten hook-block-fidelity assertions and cleanup (#1746)

griffinwork40:
Closes #1746. Renames allow-path scenario, deletes misleading comment, tightens event-count assertions to toBe(1), removes dead hooksSrc variable, and wraps results accumulator in try/catch for all four scenarios.

PR #1775, opened 2026-09-18
chore: fix TurnRunnerDeps docstring inaccuracy and minor dispatcher allocation (#1761)

### LINKED ISSUE #1761: chore: fix TurnRunnerDeps docstring inaccuracy and minor dispatcher allocation (post-#1755)
## Context

PR #1755 extracted turn-stream execution into `TurnStreamRunner` with a `TurnRunnerDeps` interface. The interface docstring states:

> All mutable fields are accessed via their owning objects -- no raw primitive refs that would diverge on reset.

Two fields violate this invariant:

1. **`conversationHistory: Message[]`** -- passed as a raw array reference (`this.conversationHistory`), not a getter. Safe today because `this.runner` is always reconstructed with the new array on reset, but the docstring claim is factually incorrect.

2. **`coreExecDeps()`** in `dispatcher.ts` is constructed twice per single-tool `execute()` call -- once inside `gateDeps()` (line 618) and once in `executeCore()` (line 818). Each call allocates a new closure object. Unmeasurable at typical rates, but an unnecessary structural artifact.

## Fix

- Wrap `conversationHistory` in a getter (`getConversationHistory: () => this.conversationHistory`), OR correct the docstring to note this exception and explain why it is safe (runner is always reconstructed on reset).
- Optionally hoist `coreExecDeps()` to a local variable in `execute()` to avoid the double allocation.

Both are nits -- low priority, no behavioral impact.

## References

- PR #1755 (merged)
- `src/agent/session/turn-stream-runner.ts:44-58` (docstring + `TurnRunnerDeps` interface)
- `src/agent/session/agent-session.ts:271`
- `src/agent/tools/dispatcher.ts:618,818`

griffinwork40:
Closes #1761. Two small post-#1755 cleanups:

## Fix 1: Correct TurnRunnerDeps docstring

The docstring claimed:
> All mutable fields are accessed via their owning objects — no raw primitive refs that would diverge on reset.

This was inaccurate. `conversationHistory` is passed as a raw array reference, not through a getter/setter pair. Updated the docstring to acknowledge this exception and explain why it is safe: the runner is always reconstructed when the history array is replaced on reset, so the stale ref can never diverge.

**File:** `src/agent/session/turn-stream-runner.ts`

## Fix 2: Hoist coreExecDeps() in execute() to avoid double allocation

`execute()` followed this path:
1. `runPreDispatchGates(call)` → `gateDeps()` → **`coreExecDeps()`** (allocation #1)
2. `this.executeCore(call)` → `_executeCore(call, this.coreExecDeps())` (allocation #2)

The second `coreExecDeps()` call was redundant — the same dispatcher fields are read both times. `execute()` now calls `coreExecDeps()` once after the gate check and passes the result directly to `_executeCore`, eliminating the duplicate object construction on the single-tool fast path.

`executeBatch()` is unaffected: it runs each tool through its own `gateDeps`/`executeCore` sequence, which is the correct pattern for batch execution.

**File:** `src/agent/tools/dispatcher.ts`

## Verification
- `pnpm lint` — clean
- `pnpm test src/agent/tools/dispatcher.test.ts` — 135 passed, 1 skipped (pre-existing)


griffinwork40:
## Code Review: PR #1775

**Reviewed ref:** `19e4f52a23f7c3060d2451ad4ba21c2c8ba301da`  
**Regime:** light (74 lines delta, 1 file, test/refactor)  
**Citations verified inline against branch HEAD `19e4f52a`.**

---

### 🚫 Findings

---

#### 1. PR description describes changes that are not in the diff — BLOCKING
**severity:** high · **blocking:** true · **confidence:** high · **dimension:** spec-compliance  
**file:** `src/agent/hook-block-fidelity.bench.test.ts` · **ref:** `19e4f52a` · **citation-type:** diff-context

**Finding:** The PR title and body claim two production-file changes (docstring fix in `turn-stream-runner.ts`, `coreExecDeps` hoisting in `dispatcher.ts`), but the actual diff touches only `src/agent/hook-block-fidelity.bench.test.ts` — zero production files are modified.

**Evidence (PR body):**
```
## Fix 1: Correct TurnRunnerDeps docstring
File: src/agent/session/turn-stream-runner.ts

## Fix 2: Hoist coreExecDeps() in execute() to avoid double allocation
File: src/agent/tools/dispatcher.ts
```

**Suggestion:** Either (a) the wrong branch was pushed and the production fixes are missing from this PR, or (b) the PR description was copied from the wrong branch. Verify that `fix/docstring-dispatcher-1761` contains the intended commits and re-push, or retitle/re-describe to match what actually landed.

---

#### 2. Scenario C `expect(hookEvents.length).toBe(1)` asserts undocumented behavior
**severity:** medium · **blocking:** true · **confidence:** medium · **dimension:** correctness  
**file:** `src/agent/hook-block-fidelity.bench.test.ts:370` · **ref:** `19e4f52a` · **citation-type:** diff-context

**Finding:** The old Scenario C comment stated "no hook_decision trace event for allows (current behavior: allows are silent in the trace)" and used a bare `for` loop with no length assertion, admitting that zero events was the expected count. The PR adds `expect(hookEvents.length).toBe(1)`, silently redefining the contract to "allows always emit exactly one event." If the current production behavior actually emits 0 events on an allow decision, this assertion will fail at runtime — and the old comment's framing ("silent in the trace") suggests 0 was the documented behavior.

**Evidence (diff):**
```diff
-      // Current behavior: allows emit hook_decision with decision field.
-      // Verify the event has an 'allowed' or non-blocked outcome.
+      expect(hookEvents.length).toBe(1);
```

**Suggestion:** Confirm the actual behavior of `dispatchPreToolUse` for an approve decision: does it emit 0 or 1 `hook_decision` events? Run the benchmark in isolation (`pnpm test src/agent/hook-block-fidelity.bench.test.ts`) and verify this passes. If `allows` are truly silent in the trace, revert to `expect(hookEvents.length).toBe(0)` (and update the comment to match); if they do emit one event, the assertion is correct but the old comment was wrong and needed this correction.

---

#### 3. `try/catch` wrapping `results.push()` is dead code
**severity:** nit · **blocking:** false · **confidence:** high · **dimension:** correctness  
**file:** `src/agent/hook-block-fidelity.bench.test.ts:229,307,378,461` · **ref:** `19e4f52a` · **citation-type:** diff-context

**Finding:** `Array.prototype.push()` on a module-scoped in-memory array cannot throw. The added `try/catch` blocks (4 instances) wrap an infallible operation, adding 20 lines of noise across all scenarios without providing any error-capture benefit.

**Evidence:**
```typescript
try {
  results.push({ scenario: '...', pass: true, detail: '...' });
} catch (err) {
  results.push({ scenario: '...', pass: false, detail: String(err) });
  throw err;
}
```

**Suggestion:** Remove the `try/catch` wrappers and restore the direct `results.push()` calls. If the intent was to catch failures from the assertions above (the `expect(...)` calls), those already throw and are caught by Vitest's test harness — the `try/catch` does not intercept them because it wraps only the `push`, not the assertions.

---

### ✅ Positive Changes

- **Dead variable removal** (`hooksSrc` at line 47 in main): `hooksSrc` was declared but never injected into any child script (all four child scripts used only `subagentHooksSrc`). Removing it eliminates a misleading dead variable. Verified: no child script template references `hooksSrc` at the reviewed ref.
- **Scenario D assertion hardened** (`toBeGreaterThanOrEqual(1)` → `toBe(1)` at line 449 in head): The multi-handler scenario has exactly one blocking handler; pinning the count to 1 prevents false passes from spurious extra events. Correct tightening.
- **Scenario C comment and test name updated:** The comment no longer contradicts itself (the old version said "no hook_decision event" while the child script returned `{ decision: 'approve' }`, which may emit one). The new test name is clearer.

---

### What was not checked

- Production files `src/agent/session/turn-stream-runner.ts` and `src/agent/tools/dispatcher.ts` — the PR claims changes there but the diff contains none; no review of those files was warranted.
- Runtime execution of the benchmark scenarios (shell-less review).
- Whether `dispatchPreToolUse` in `subagent-hooks.ts` emits 0 or 1 `hook_decision` events on an approve decision — this is the load-bearing question for Finding #2.
- Stated intent: PR #1775 title+body — **spec-compliance assessed** (mismatch found; see Finding #1).

---

### Decision

**DO NOT MERGE — 1 high, 1 medium blocking (1 spec-compliance, 1 correctness); 1 nit.**

The high finding (description mismatch) must be resolved first: either the wrong branch was submitted and the intended production changes are missing, or the PR description is wrong and needs to be corrected to reflect that this is a benchmark test cleanup PR. Finding #2 (Scenario C count assertion) must also be verified against actual runtime behavior before this can land safely.

griffinwork40:
Closing this PR — it was submitted on the wrong branch. The diff only contains bench test changes that already landed in #1774, not the two fixes described in the title and body.

The correct changes (fix TurnRunnerDeps docstring + hoist `coreExecDeps()` in `execute()`) have been re-landed on a clean branch and will be tracked in a new PR.

PR #1782, opened 2026-09-18
docs(status-line): clarify droppablePriority: 0 intent (#1771)

### LINKED ISSUE #1771: status-line-format: clarify droppablePriority: 0 intent with inline comment
## Context

In `src/cli/render/status-line-format.ts` (introduced in PR #1768), the quota segment receives `droppablePriority: 0` when `quota.severity === 'critical' && !quota.stale`.

## Issue

The value `0` is semantically correct — it makes the critical quota segment the last droppable to be shed (the shed loop picks `Math.max(...)` each iteration, and 0 is always ≤ every other priority). However, `0` visually reads as a possible falsy omission or uninitialized value rather than an intentional floor, especially since `undefined` means "never drop" elsewhere in the same structure.

## Suggested Fix

Add an inline comment at the assignment site (~line 260):

```typescript
// 0 = drop-LAST among droppables; shed only after every higher-priority
// segment is gone. NOT undefined (never-drop) — that tier is reserved for
// model/mode to avoid stacking never-drops on narrow terminals.
const priority = quota.severity === 'critical' && !quota.stale ? 0 : 5;
```

## Source

Review finding F2 from PR #1768 review.

griffinwork40:
Closes #1771

## What

Replaces the terse inline comment on the `droppablePriority` assignment for the critical quota segment with a fuller three-line block comment that explains:

- `0` means **drop-last** among droppables (shed only after every higher-priority segment is gone)
- `0` is **not** `undefined` — `undefined` means never-drop, a tier intentionally reserved for model/mode segments to avoid stacking never-drops on narrow terminals

## Why

`0` is visually easy to misread as a falsy omission or an uninitialized value. The prior one-liner (`// 0 = drop-LAST (not undefined = never-drop)`) conveyed the idea but didn't explain the rationale for keeping it out of the `undefined` (never-drop) tier.

## Change type

Comment-only — no logic changes. `pnpm lint` (tsc --noEmit) passes.


griffinwork40 (commented):
## Review — PR #1782

**Reviewed ref:** `c342ce01765ddce577373b00a51967d75b77e71a`
**Regime:** light (7 lines, comment-only)
**Change type:** refactor (no logic changes)

---

### Security · API-compat
No auth, secret exposure, injection surface, or public API changes. Comment-only diff.
✅ No issues found.

### Correctness · Spec-compliance
The new block comment satisfies all three stated-intent requirements:

1. **"0 = drop-last"** — The shed loop uses `Math.max(...droppable.map(p => p.droppablePriority!))` and drops the highest numeric priority first. A value of `0` is the minimum possible droppable integer, so it survives all other droppables before being shed. The comment accurately describes this as "shed only after every higher-priority segment is gone."

2. **"NOT undefined"** — Accurate. The `Part` interface declares `droppablePriority?: number` with `undefined = never drop`. Priority `0` is a distinct assigned value, not an omission.

3. **"reserved for model/mode to avoid stacking never-drops"** — Accurate. Model (line ~178) and permission-mode chips (lines ~221–229) carry no `droppablePriority`, making them never-drop. The comment correctly names the invariant being protected.

The uniqueness invariant ("every droppablePriority on this line must be UNIQUE") is satisfied — `0` appears only on the quota-critical path.

✅ No issues found.

### Test coverage
Comment-only change; no logic altered. No test gap.
✅ No issues found.

### Perf / observability
No performance or observability surface touched.
✅ No issues found.

---

### What was not checked
- Citations verified inline against branch HEAD `c342ce01765ddce577373b00a51967d75b77e71a`.
- Stated intent: PR #1782 title + body — spec-compliance assessed; all requirements met.
- Did not run the test suite (comment-only; no behavior changed).

---

**Decision: MERGE — 0 findings.**

PR #1783, opened 2026-09-18
fix(captures): validate filenames against path traversal (#1706)

### LINKED ISSUE #1706: fix(captures): validate --limit flag and session directory names
## Summary

`afk captures list` has two input-validation gaps identified during PR #1641 review:

### 1. `--limit` flag silently swallows invalid input

`parseInt(options.limit, 10) || 20` at `src/cli/commands/captures.ts:65` resolves both `--limit 0` and `--limit abc` to 20 with no warning.

**Fix:** Guard with `Number.isNaN()`:
```ts
const n = parseInt(options.limit, 10);
const limitN = Number.isNaN(n) || n <= 0 ? 20 : n;
```

### 2. Session directory names not validated against path traversal

`readdir(witnessRoot)` results are used in `path.join()` without validation at `src/agent/tools/bash-capture-index.ts`. A crafted directory name like `../../etc` would escape the witness root. Every other witness-tree reader calls `validateSessionId()` from `paths.witness.ts` -- this module is the only one that skips it.

**Fix:** Filter after readdir:
```ts
sessionDirs = sessionDirs.filter(d => /^[a-zA-Z0-9_-]+$/.test(d));
```

Blast radius is narrow (operator-owned space, read-only), but consistency with the rest of the witness tree matters.

Refs: PR #1641, review findings F-04, F-05, F-R2-1, F-R2-2

griffinwork40:
## Summary

Fixes the path-traversal gap in the captures filename filter identified in #1706.

### Changes

**`src/agent/tools/bash-capture-index.ts`**

The `.filter((f) => f.endsWith('.txt'))` guard at line 181 admitted any filename that merely ended in `.txt`, including path-traversal names like `../../etc/passwd.txt`. Replaced with a safe-name allowlist regex mirroring the one already applied to session directory names at line 150:

```diff
-  .filter((f) => f.endsWith('.txt'))
+  .filter((f) => /^[a-zA-Z0-9._-]+\.txt$/.test(f))
```

The character class includes `.` so dotted epoch-based names like `1726123456.txt` continue to pass, while `/` and any other out-of-set characters are rejected.

### Verified unchanged (no fix needed)

**`src/cli/commands/captures.ts` — `--limit` parsing** was already correctly implemented (lines 65-66):

```ts
const n = parseInt(options.limit, 10);
const limitN = Number.isNaN(n) || n <= 0 ? 20 : n;
```

This properly guards against `NaN` and non-positive values; no change required.

### Verification

- `pnpm lint` (tsc --noEmit): ✅ passes

Closes #1706


griffinwork40 (commented):
## `/review` — PR #1783 · fix(captures): validate filenames against path traversal (#1706)

**Reviewed ref:** `6ae7401c6b74b12c739424aee983c49e76e8ec11`
**Regime:** light (1-line hotfix)
**Target:** `src/agent/tools/bash-capture-index.ts`
**Wave 1.5 citation verification:** all citations are diff-context; no file-state citations to verify.

---

### Findings

#### LOW · blocking: false · high confidence · test-coverage
**`src/agent/tools/bash-capture-index.ts:181` · ref: `6ae7401` · diff-context**

No test in the suite asserts that path-traversal-shaped filenames (e.g., `../../etc/passwd.txt`) are rejected by the new filter. The robustness suite covers non-`.txt` files and missing directories, but the specific attack surface this fix closes has no regression test.

**Evidence** (diff-context):
```ts
-  .filter((f) => f.endsWith('.txt'))
+  .filter((f) => /^[a-zA-Z0-9._-]+\.txt$/.test(f))
```

**Suggestion:** Add a test case in the `listCaptures — robustness` block:
```ts
it('rejects path-traversal filenames in bash-captures', async () => {
  const dir = join(witnessRoot(), 'sess-traversal', 'bash-captures');
  mkdirSync(dir, { recursive: true });
  writeFileSync(join(dir, '../../etc/passwd.txt'), 'should not appear');
  writeFileSync(join(dir, 'real.txt'), 'real\n');
  const result = await listCaptures({ sessionId: 'sess-traversal' });
  expect(result).toHaveLength(1);
  expect(result[0]!.toolUseId).toBe('real');
});
```

---

**Regex correctness — verified correct.**
- `^` and `$` anchors are present; multiline bypass is impossible.
- `-` is at the end of `[a-zA-Z0-9._-]`, making it a literal hyphen (not a range). Runtime-tested: a `/` (ASCII 47, between `.` 46 and `_` 95 if it were a range) is correctly rejected.
- `.` in a character class is a literal dot, not a wildcard. Traversal via `../` is impossible.

**Write-side vs. read-side character class asymmetry — benign.**
The write-side sanitizer in `_bash-capture.ts:71` uses `[^A-Za-z0-9_-]` (excludes `.`), while the new read-side regex admits `.`. Files written by the system will always pass the read filter (write-side ⊂ read-side). The extra `.` on the read side is harmless: any dotted filename not producible by the write path cannot be injected via normal operation, and reading such a file would only surface metadata + a preview — no escalated access.

**`--limit` parsing in `captures.ts` — no change needed, correctly noted in PR body.**
`parseInt` + `Number.isNaN` + non-positive guard is correct. The PR description is accurate.

---

### What was not checked

- Citations verified inline against branch HEAD `6ae7401c6b74b12c739424aee983c49e76e8ec11`.
- Did not run the test suite; `pnpm lint` result taken from PR body at face value.
- Did not review Telegram or daemon surfaces (not affected by this change).
- Stated intent: PR #1783 title + body — spec-compliance assessed. The diff fully satisfies the stated intent; no unmet requirements, no scope creep.

---

**Decision: MERGE — 0 blocking (1 low: missing traversal regression test).**

The fix is correct, minimal, and the security property is sound. The missing regression test is worth adding but does not block merge — the behavior change is protected by the allowlist regex itself, and the existing test suite exercises the filter path from multiple angles.

PR #1784, opened 2026-09-18
fix(bench): add timeout guard to spawnAndKill helpers (#1747)

### LINKED ISSUE #1747: bench: add child-process timeout to spawnAndWait in bench tests
## Context

PR #1745 shipped the hook-block fidelity benchmark. The `spawnAndWait` helper used by all four scenarios has no child-process kill path -- if a child deadlocks, the promise never settles and vitest's 15s `it()` timeout kills the test but leaves the child process orphaned.

This same helper pattern is used across the other bench tests (`trace-completeness`, `abort-cascade`, `dag-checkpoint`, `concurrent-emitter`), so the fix should be applied to the shared pattern.

## Current behavior

```typescript
child.on('exit', (code) => { /* resolve/reject */ });
child.on('error', (err) => { /* reject */ });
// no setTimeout + child.kill fallback
```

If tsx startup fails silently or an `await` deadlocks inside the child, nothing settles the promise.

## Fix

Add a `setTimeout` inside `spawnAndWait` that calls `child.kill('SIGTERM')` and rejects after ~12s (inside the 15s vitest timeout), with cleanup on normal settlement:

```typescript
const timer = setTimeout(() => {
  child.kill('SIGTERM');
  reject(new Error('child timed out'));
}, 12_000);
child.on('exit', (code) => { clearTimeout(timer); /* ... */ });
child.on('error', (err) => { clearTimeout(timer); /* ... */ });
```

## Files

- `src/agent/hook-block-fidelity.bench.test.ts`
- Audit the same pattern in sibling bench tests

griffinwork40:
## Summary

Adds a 12-second timeout guard to the `spawnAndKill` helpers in the two trace bench test files, closing #1747.

## Problem

`spawnAndKill` in `trace-completeness.bench.test.ts` and `concurrent-emitter.bench.test.ts` had no timeout. If a child process never printed `READY\n` — due to a tsx startup hang, kernel I/O stall, or an infinite loop in the generated child script — the returned promise would never settle. The test would hang silently until vitest's outer suite timeout killed the entire run, discarding all output and leaving no diagnostic signal.

The sibling bench files (`hook-block-fidelity.bench.test.ts`, `abort-cascade.bench.test.ts`) already had a 12-second guard. This makes the trace bench files consistent.

## Changes

Both `spawnAndKill` helpers now:

1. Start a 12-second `setTimeout` guard immediately after spawning.
2. On **READY received**: set `killed = true`, `clearTimeout(timer)`, then `child.kill('SIGKILL')` — the normal fast path is unchanged.
3. On **timer fires**: set `killed = true`, `child.kill('SIGTERM')`, then `reject(new Error('child timed out waiting for READY'))`.
4. On **`exit`**: always `clearTimeout(timer)` before resolving — prevents a race where a slow-exiting child leaves a live timer that fires after promise settlement.
5. On **`error`**: always `clearTimeout(timer)` before handling.

## Verification

- `pnpm lint` (`tsc --noEmit`) — clean, no new type errors
- `pnpm test src/agent/trace/trace-completeness.bench.test.ts` — 4/4 pass
- `pnpm test src/agent/trace/concurrent-emitter.bench.test.ts` — 3/3 pass

Closes #1747


griffinwork40 (commented):
**Review — PR #1784** `fix(bench): add timeout guard to spawnAndKill helpers (#1747)`

Reviewed ref: `ba25d82ff290f9a50b66d06660e73babc1e8f5bf`
Regime: **light** (24 lines, 2 test-only files)
Stated intent: PR title + body — spec-compliance assessed.

---

## Security · API-compat

No issues found — read both changed files at reviewed ref.

Test files spawn isolated child processes with sanitized `env` (only `PATH` + `HOME`). No public API changes; these are test-only helpers with no production importers.

## Correctness · Spec-compliance · Test-coverage · Perf-observability

**Spec-compliance:** All 5 stated requirements implemented with no gaps and no scope creep.

1. 12-second `setTimeout` guard starts immediately after spawn — ✓
2. On READY: `killed = true` → `clearTimeout(timer)` → `child.kill('SIGKILL')` — ✓
3. On timer fires: `killed = true` → `child.kill('SIGTERM')` → `reject(...)` — ✓
4. On `exit`: `clearTimeout(timer)` before `resolve()` — ✓
5. On `error`: `clearTimeout(timer)` before handling — ✓

**Edge-case correctness:**
- Timer-fires-then-exit race: timer fires → `reject()`; then `exit` fires → `clearTimeout` (idempotent) → `resolve()` on an already-settled promise (silently ignored). Safe.
- Normal path fires: `killed = true` + `clearTimeout(timer)` prevents timer from firing after resolution. The `exit`-path `clearTimeout` is idempotent. Safe.
- `if (!killed)` guard in timer callback prevents double-action after normal path wins the race. Correct.

**Consistency:** Pattern matches `hook-block-fidelity.bench.test.ts` and `abort-cascade.bench.test.ts` exactly.

**Test timeout alignment:** Test-level timeouts (15 s / 20 s) exceed the 12 s guard + child overhead. Adequate margin.

**Perf/observability:** Test-harness only; no production impact.

---

## Citation verification (Wave 1.5)

All citations are `diff-context` — every referenced line is visible in the diff hunk. No `file-state` citations. No absence claims. Nothing to drop or downgrade.

---

## Decision: MERGE — 0 findings.

> Stated intent: PR title + body — spec-compliance assessed.
> Citations verified inline against branch HEAD `ba25d82`.
> Did not review production surfaces (change is test-only).
> Did not run tests (shell-read-only review).

PR #1796, opened 2026-09-18
chore: use getTerminalHeight() in preview-cap sites for consistency

### LINKED ISSUE #1780: chore: use getTerminalHeight() in preview-cap sites for consistency
Surfaced during review of #1765 (now merged):

### Convention drift: direct `process.stdout.rows` vs `getTerminalHeight()`

The three new preview-cap sites (`previewCodeFence`, `previewTable`, `formatPendingBuffer`) read `process.stdout.rows ?? 24` directly, while the same file already imports `getTerminalHeight()` from `terminal-size.ts` for the analogous width case.

This is a behavioral no-op today but could diverge in a future resize-handling refactor. All three sites should use the shared helper for consistency.

Ref: PR #1765

griffinwork40:
## Summary

Three preview-cap sites were reading `process.stdout.rows ?? 24` directly instead of using the shared `getTerminalHeight()` helper from `src/cli/terminal-size.ts`. This is an inconsistency — all width reads already use `getTerminalWidth()` — and it duplicates the fallback logic that already lives in the helper.

## Changes

- **`src/cli/markdown-stream-format.preview.ts`** — add `getTerminalHeight` import; replace `process.stdout.rows ?? 24` in `previewCodeFence` (line 78) and `previewTable` (line 116)
- **`src/cli/markdown-stream-format.ts`** — add `getTerminalHeight` to the existing `terminal-size.js` import; replace `process.stdout.rows ?? 24` in `formatPendingBuffer` (line 95)

`getTerminalHeight()` already encapsulates `typeof r === 'number' && r > 0 ? r : DEFAULT_ROWS` where `DEFAULT_ROWS = 24`, so the fallback behaviour is unchanged.

## Verification

- `pnpm lint` — clean
- `pnpm test src/cli/markdown-stream-format` — 31/31 passed

Closes #1780


griffinwork40 (commented):
## Review — chore: use getTerminalHeight() in preview-cap sites for consistency

**Reviewed ref:** `ef20173f`  
**Regime:** light (13 lines, 2 files, refactor/chore)  
**Stated intent:** PR #1796 title + body — spec-compliance assessed

---

### Findings

**nit · blocking:false · confidence:high · style**  
`src/cli/markdown-stream-format.test.ts:147` · ref:`ef20173f` · diff-context  
The viewport-cap test still computes its expected value via `(process.stdout.rows ?? 24) - 2` rather than `getTerminalHeight()`. The PR migrates all three production call sites but leaves the test's reference expression using the raw pattern. Functionally identical in test environments (where `process.stdout.rows` is `undefined`), but it's a minor inconsistency that would resurface if `getTerminalHeight()`'s fallback ever changed.  
**Suggestion:** import `getTerminalHeight` into the test and replace line 147's expression with it.

---

### What was not checked

- Citations verified inline against branch HEAD `ef20173f8c94a19c03217a7cd0bda1f49cf29aa2`.
- Stated intent: PR #1796 title+body — spec-compliance assessed. All 3 preview-cap sites migrated as described; no unmet requirements, no scope creep.
- Did not run the test suite locally (PR body reports 31/31 passing).

---

**Decision: MERGE — 0 blocking (1 nit).**

The behavioral equivalence holds: `getTerminalHeight()` uses `r > 0` vs the old `?? 24`'s implicit nullish-only check. The only edge-case difference is `rows = 0`, which `getTerminalHeight()` correctly maps to 24 (preventing `Math.max(1, -2) = 1`), making the new code strictly more correct. Import is clean, fallback unchanged.

PR #1798, opened 2026-09-18
test(captures): add regression test for path-traversal filename filter (#1790)

### LINKED ISSUE #1790: test(captures): add regression test for path-traversal filename filter
## Context

PR #1783 (merged, closes #1706) added a safe-name regex filter to `bash-capture-index.ts` to prevent path traversal via crafted filenames. The review (low, non-blocking) noted that no test asserts the filter rejects traversal-shaped filenames.

## What to add

Add a test in the appropriate test file (or create `src/agent/tools/bash-capture-index.test.ts`) that:

1. Creates a mock captures directory containing a file named `../../etc/passwd.txt`
2. Calls the listing/indexing function
3. Asserts the traversal filename is excluded from results
4. Asserts legitimate filenames like `1726123456.txt` and `capture-abc.txt` are included

## Source

Review finding (low) on PR #1783.

griffinwork40:
Closes #1790

## What

Adds 8 regression tests to `src/agent/tools/bash-capture-index.test.ts` for the path-traversal filename filter introduced in PR #1783.

## Why

PR #1783 added safe-name regex filters to `listCaptures` to prevent path traversal via crafted session directory names or capture filenames. No test existed to assert this behaviour, leaving the filter unguarded against future regressions.

## Tests added (new `listCaptures — path-traversal filename filter` describe block)

**Regex contract tests (pure, no fs)**
- `rejects slash-containing filenames via the safe-name regex` — verifies `../../etc/passwd.txt`, `../sibling.txt`, `/absolute/path.txt`, and `foo/bar.txt` are all rejected by `/^[a-zA-Z0-9._-]+\.txt$/`
- `rejects non-txt extensions via the safe-name regex` — `.sh`, `.json`, no extension, and double-extension names are excluded
- `accepts safe filenames via the safe-name regex` — timestamp-style, kebab-style, mixed-char names all pass

**Integration tests (real temp dir via existing `makeCapture` helper)**
- `includes timestamp-style filenames in scan results` — `1726123456.txt` surfaces correctly
- `includes kebab-style filenames in scan results` — `capture-abc.txt` surfaces correctly
- `includes both legitimate filenames and excludes wrong-extension files` — `.sh`/`.json` siblings are filtered; only `.txt` entries appear
- `excludes dot-prefixed session directory names from the scan` — `.traversal/` dir is skipped by the session-level `/^[a-zA-Z0-9_-]+$/` filter
- `excludes session dirs containing dots from the scan` — `sess..evil/` is skipped by the same filter

## Notes on filter semantics

The safe-name regex for filenames (`/^[a-zA-Z0-9._-]+\.txt$/`) allows dots in the stem. A bare readdir entry like `..evil.txt` is **not** a path-traversal risk because `join(capturesDir, "..evil.txt")` remains inside `capturesDir` — only entries containing `/` can escape it on POSIX, and those are correctly rejected. The tests document this invariant explicitly in comments so future maintainers understand the security contract.


griffinwork40:
## agent-afk review · PR #1798

**Target:** `fix/1790-path-traversal-regression-test` → `main`  
**Reviewed ref:** `b5942b5f31c68938422963f73998b696b64813ea`  
**Change type:** test + docs (feature)  
**Regime:** full (544 lines, 7 files)

---

### Triage

The PR adds two test expansions and five documentation fixes. The PR title and stated intent reference the path-traversal regression test for `bash-capture-index`, but the branch also includes `shell-task.test.ts` coverage (commit `7e910ac5`) and website docs cleanup (commit `d22dedd0`). These are bundled across three distinct commits. All 45 tests on the branch pass (verified locally).

---

### Findings

#### MEDIUM · blocking:false (waived: bounded, non-data-affecting, test-only scope, no downstream consumer)

**dimension:** correctness · test-coverage  
**file:line:** `src/agent/tools/bash-capture-index.test.ts:263–294`  
**ref:** `b5942b5f`  
**citation-type:** diff-context

**finding:** The three pure-regex tests (`rejects slash-containing filenames`, `rejects non-txt extensions`, `accepts safe filenames`) instantiate the filename filter regex `/^[a-zA-Z0-9._-]+\.txt$/` directly in the test body rather than sourcing it from the production module. If the production regex at `bash-capture-index.ts:181` is ever changed, these three tests continue to pass green while only the integration tests catch the regression. The tests document the _intended_ invariant but do not guard it — they guard a copy.

**evidence:**
```ts
// test body (bash-capture-index.test.ts:270)
expect(/^[a-zA-Z0-9._-]+\.txt$/.test(name)).toBe(false);

// production (bash-capture-index.ts:181)
.filter((f) => /^[a-zA-Z0-9._-]+\.txt$/.test(f))
```

**suggestion:** Export the regex as a named constant from `bash-capture-index.ts` (e.g., `export const SAFE_CAPTURE_FILENAME_RE = /^[a-zA-Z0-9._-]+\.txt$/`) and import it in the test. The pure-regex tests then guard the live regex and serve as both documentation and a regression gate.

**waived:** test-only defect, no production behavior affected, integration tests provide a backstop for the actual filter.

---

#### LOW · blocking:false

**dimension:** test-coverage  
**file:line:** `src/agent/daemon/shell-task.test.ts:25–32`  
**ref:** `b5942b5f`  
**citation-type:** diff-context

**finding:** `makeSteppingClock` is defined as a test helper but is used in only 2 of the 5 telemetry tests (lines 275 and 299). The `durationMs` test at line 337 — the one that most naturally benefits from a deterministic clock — uses its own inline `times` array instead. The helper was introduced but not applied to its primary use case.

**evidence:**
```ts
function makeSteppingClock(startMs = 1_700_000_000_000, stepMs = 50) { ... }
// used at lines 275, 299 only; line 337 uses manual times[] instead
```

**suggestion:** Either apply `makeSteppingClock` in the duration test for consistency, or leave the inline approach and remove the helper (or document why the approaches differ). Not a blocker — both approaches are correct.

---

#### NIT · blocking:false

**dimension:** correctness  
**file:line:** `src/agent/daemon/shell-task.test.ts:333–347`  
**ref:** `b5942b5f`  
**citation-type:** diff-context

**finding:** The `durationMs` test comment says `first call (triggeredAt) = 1000, second call (startTimeMs) = 1050, third call (end) = 1300`. The production code calls `options.now()` exactly three times in the success path (once for `triggeredAt`, once for `startTimeMs`, once for `durationMs`). The comment and the `times[Math.min(callCount++, times.length - 1)] ?? 0` clamp are both correct, but the clamp silently allows a fourth call to return `1300` instead of failing. If `runShellTask` ever adds a fourth `now()` call, the test won't catch the over-count.

**suggestion:** Use `vi.fn().mockReturnValueOnce(1000).mockReturnValueOnce(1050).mockReturnValueOnce(1300)` so a fourth unexpected call returns `undefined` and surfaces the count mismatch. Genuinely minor.

---

#### LOW · blocking:false (docs)

**dimension:** api-compat  
**file:line:** `website/content/docs/guides/cli-reference.mdx:43`  
**ref:** `b5942b5f`  
**citation-type:** diff-context

**finding:** The updated `--provider` docs list `openai-codex` as a current option, but `src/agent/providers/index.ts` documents it as deprecated ("Will be removed"). The docs should mark it `(deprecated)` to align with the source comment and avoid new users adopting a deprecated value.

**evidence:**
```
# docs (line 43)
`anthropic` | `anthropic-direct` | `openai` | `openai-compatible` | `openai-codex`

# src/agent/providers/index.ts:55-58
* Backward-compat note: `'openai-codex'` is **deprecated** as of ... Will be removed.
```

**suggestion:** Append `(deprecated)` to `openai-codex` in the docs table. One-word change, consistent with the source.

---

### What was correct

- All 45 tests pass on the branch (verified: `pnpm exec vitest run src/agent/tools/bash-capture-index.test.ts src/agent/daemon/shell-task.test.ts`).
- The two integration tests for session-level filtering (`excludes dot-prefixed session directory names`, `excludes session dirs containing dots`) correctly exercise the production filter at `bash-capture-index.ts:150` (`/^[a-zA-Z0-9_-]+$/`), not a copy of it — these are the tests that actually guard the security invariant.
- The Invariant comment block in the test correctly explains the security boundary (slash is the only traversal vector from bare readdir entries) — this is accurate and valuable documentation.
- `shell-task.test.ts` correctly replaces `'manual'` trigger with `'cron'` — `'manual'` is not in `TelemetryTrigger = 'cron' | 'sessionstart' | 'pull'`, so the old tests were unsound.
- `AFK_SUBAGENT_RESULT_CAP_BYTES` and `AFK_TELEGRAM_BOT_TOKEN` docs additions match their registered entries in `src/config/env.ts` (lines 1530 and 811 respectively).
- `state_cas.metadata` docs clarification matches `src/agent/state/state-schemas.ts:77` (`Not returned by state_get — for audit/tagging only`).
- `web.mdx` bearer-token auth flow description is a docs-only improvement; no code changed.

---

### Epistemic scope

Citations verified inline against branch HEAD `b5942b5f31c68938422963f73998b696b64813ea`. Production source read at that ref via `git show`. Tests run locally on the branch. Stated intent: PR #1798 title + body — spec-compliance assessed (intent: regression tests for path-traversal filter; the primary deliverable is present and correct; the bundled unrelated commits are out of scope for intent-compliance).

Did not review: Telegram surface, daemon surface, MCP layer. No runtime tests run (vitest only).

---

The core deliverable — regression tests for the path-traversal filename filter — is correct, well-documented, and passes. The medium (regex copy) is a real test hygiene gap but has no production impact and integration tests backstop it. Worth a follow-up issue or a squash-in fix, not a merge block.

PR #1810, opened 2026-09-19
fix(service): upgradeService preserves existing env vars on restart

### LINKED ISSUE #1717: fix(service): CLI restart still drops installed environment vars
## Context

From Codex review of PR #1711 (merged). PR #1711 fixed the programmatic API: `restart()` now forwards `opts` to `upgradeService()`. However, the sole production caller -- `afk service restart` in `src/cli/commands/service.ts:193` -- still invokes `mgr.restart(name)` with no opts.

This means `afk service restart` still effectively strips custom environment variables from the plist during upgrade, because it passes `{}` (empty opts) rather than recovering the installed environment.

## Fix direction

Either:
1. Read the current plist's `EnvironmentVariables` before upgrade and forward them as `opts.environment` (preserve-on-restart semantics)
2. Add `--environment` flags to `afk service restart` so the user can explicitly supply them
3. Have `upgradeService` preserve existing env vars from the current plist when `opts.environment` is undefined (most user-friendly)

Option 3 is probably the right default -- "restart should not change config unless you ask it to."

## Files

- `src/cli/commands/service.ts:193` (CLI caller)
- `src/service/launchd/install.ts` (upgradeService)

## Priority

Low -- currently unreachable from CLI since `install` never passes `environment`, but a latent contract gap for future callers.

## Origin

Codex inline review on PR #1711. Pre-existing gap noted in the original issue #1694.

griffinwork40:
## Summary

PR #1711 fixed `restart()` to forward `opts` to `upgradeService()`, but the sole production CLI caller — `afk service restart` in `src/cli/commands/service.ts:193` — still calls `mgr.restart(name)` with no opts. This means a bare `afk service restart <name>` would silently strip any custom environment variables (e.g. `TELEGRAM_BOT_TOKEN`) that were baked into the plist at install time.

This PR implements **Option 3** from the issue: have `upgradeService` preserve existing env vars from the on-disk plist when `opts.environment` is undefined — *"restart should not change config unless you ask it to."*

### `src/service/launchd/install.ts`
- **New exported helper `extractPlistEnvVars(plistXml)`**: parses the `EnvironmentVariables` dict from a rendered plist string. Deliberately narrow — only handles the exact XML shape `renderPlist` emits. Never throws; returns `{}` on any parse failure.
- **`upgradeService` now reads the current plist first** and extracts its env vars. When `opts.environment` is undefined, those vars (minus `PATH`, which is always re-resolved fresh from the current node binary) are used as a base. When `opts.environment` IS supplied, the caller's values win entirely — same merge semantics as `installService()`.
- File stays within the 350-LOC ceiling (346 LOC after changes).

### `src/service/launchd.test.ts`
- `extractPlistEnvVars` unit tests: happy path, empty plist, unparseable input, XML-escaped characters round-trip.
- Two new `upgradeService` integration tests: (a) env-var preservation on bare restart, (b) caller-supplied `opts.environment` wins and old vars are NOT preserved.

### systemd path
`systemdManager.restart()` issues `systemctl --user restart`, which leaves the unit file untouched. No env vars are modified by a systemd restart, so no changes needed there.

## Testing

```
pnpm test    # 19921 passed, 26 skipped — no regressions
pnpm lint    # tsc --noEmit clean
```

Closes #1717


griffinwork40 (commented):

---

## Prior Rounds

No prior automated rounds — this is round 1.

---

## Review — PR #1810 · `fix(service): upgradeService preserves existing env vars on restart`

**Reviewed ref:** `71dc0e57a006ffeb3e72080278ba217b716346e5`
**Regime:** light (2 files, ~200 lines, hotfix)
**Change type:** hotfix
**Stated intent:** `afk service restart` silently strips custom env vars baked into the plist at install time. Fix: have `upgradeService` preserve existing env vars from the on-disk plist when `opts.environment` is undefined. Closes #1717.

---

**Severity arithmetic:** 0 critical, 0 high → MERGE (with follow-ups); 2 medium advisory (non-blocking by sweep policy), 2 low advisory, 1 nit advisory.

The fix correctly solves the stated silent-env-var-stripping bug. The `extractPlistEnvVars` + `preservedEnv` approach is sound, well-tested for the restart path, and security-safe (secrets round-trip through the same 0o600 file, no new exposure surface). CI is green. All advisory findings are suitable for follow-up issues.

---

## Findings

### ✅ Security — no issues found

`extractPlistEnvVars` reads secrets from the on-disk plist (0o600) and re-injects them into `environmentVariables`, which is written back via `renderPlist` with the same 0o600 mode. The entire flow is file→memory→file. No new attack surface, no secret exposure path. `xmlEscape()` in `renderPlist` correctly re-encodes round-tripped values.

### ✅ API-compat — no issues found

`extractPlistEnvVars` is an additive export from `install.ts`, re-exported via the `launchd.ts` barrel. No existing public symbol removed or renamed. The `upgradeService` signature is unchanged; the new preservation behavior is backward-compatible (callers passing `opts.environment` get override semantics as before; callers omitting it get preservation instead of env-var stripping — an improvement).

### 🟡 Correctness — 2 advisory findings

**medium · blocking:false · confidence:high · correctness · `src/service/launchd/install.ts` (upgradeService) + `src/cli/commands/service.ts:166` · ref:71dc0e57 · file-state (verified)**

Undocumented scope expansion: the `afk service upgrade` CLI path also calls `upgradeService` without `opts.environment` (via `mgr.upgrade(name, { noWatch: ... })`), so it now preserves existing env vars too — not just `restart`. This is arguably the correct behavior ("upgrade should not change config unless you ask it to"), but the PR body only describes the `restart` path and no test covers the upgrade-path preservation specifically. A user who manually edits the plist to *remove* a token and then runs `afk service upgrade` would see it silently re-injected.

> **Suggestion (follow-up issue):** Either document that `upgradeService` now preserves env vars for both restart and upgrade paths, and add a `launchdManager.upgrade()` test asserting env-var preservation end-to-end, or introduce a `preserveEnv` flag that only the restart path sets.

**low · blocking:false · confidence:high · correctness · `src/service/launchd/install.ts:13–19` · ref:71dc0e57 · diff-context (verified)**

The outer regex `/<key>EnvironmentVariables<\/key>\s*<dict>([\s\S]*?)<\/dict>/` uses a non-greedy match that stops at the first `</dict>`. This is correct for AFK-generated plists (which never have nested dicts inside `EnvironmentVariables` — confirmed at `plist.ts:108–119`). For hand-edited plists with nested `<dict>` structures inside `EnvironmentVariables`, the match would silently truncate. The JSDoc correctly disclaims this ("only handles the exact XML shape `renderPlist` emits"), and the fail-safe is `{}` which degrades to PATH-only. Practically no risk for the intended use case.

> **Suggestion:** Add a one-line code comment: `// NB: stops at first </dict> — only handles flat key/string pairs from renderPlist.`

### ✅ Spec-compliance — intent met

The stated intent required preserving installed env vars on `afk service restart` when no `opts.environment` is supplied. The fix implements this in `upgradeService` via `extractPlistEnvVars` + `preservedEnv` spread. The `PATH` key is correctly excluded from preservation and freshly resolved. The `opts.environment` explicitly-supplied path correctly wins and drops old vars (verified in test). Intent is fully satisfied. No scope creep in the production code.

### ✅ Test-coverage — 1 advisory finding

**low · blocking:false · confidence:medium · test-coverage · `src/service/launchd.test.ts` · ref:71dc0e57 · file-state (verified)**

The new tests comprehensively cover `extractPlistEnvVars` (happy path, empty plist, unparseable input, XML entity round-trip) and `upgradeService` directly (env preservation when `opts.environment` undefined, env override when `opts.environment` supplied). However, no test covers the `launchdManager.upgrade()` adapter with the new env-preservation behavior end-to-end. Existing manager upgrade tests (lines 1083–1146) validate routing but not env-var content. Given Finding 1 (undocumented scope expansion to upgrade), a manager-level test would be valuable.

> **Suggestion:** Add a test in the existing `upgrade (manager → upgradeService delegation)` block that seeds a plist with custom env vars, calls `launchdManager.upgrade('telegram')` with no `environment`, and asserts the vars survive.

### ✅ Perf-observability — no issues found

The `readFileSync` call was already unconditional in the old code. Moving it before env-var computation adds no new I/O. The destructuring IIFE for PATH exclusion is negligible at the call frequency (once per restart/upgrade).

---

## Wave 1.5 — Dropped Findings

1 finding **dropped as fabricated**: sub-agent flagged a "placeholder spread" (`...plist XML...`) in the second `upgradeService` test. Verified at ref: the actual test file (lines 790–807) contains a complete plist XML block, not a placeholder. The abbreviation was in the review input, not the code.

---

## What Was Not Checked

- Citations verified inline (Wave 1.5) against branch HEAD `71dc0e57a006ffeb3e72080278ba217b716346e5`.
- Stated intent: PR #1810 title + body — spec-compliance assessed and met.
- Did not review systemd backend (not in this diff; PR body confirms no changes needed).
- Did not run the test suite locally; CI results (all green) taken as authoritative.
- Did not verify `renderPlist` XML-escape behavior end-to-end with `extractPlistEnvVars` for all Unicode edge cases.

griffinwork40 (commented):
## Review: fix(service): upgradeService preserves existing env vars on restart

**Verdict: Approve with one minor observation.**

### What this PR does

`upgradeService` previously re-rendered the plist with only `PATH` (freshly resolved) plus whatever `opts.environment` supplied. Because `afk service restart` passes no opts, a bare restart silently dropped all env vars that were baked in at install time (e.g. `TELEGRAM_BOT_TOKEN`). This PR fixes that by reading the current on-disk plist first, extracting its `EnvironmentVariables` dict, and feeding those back as the merge base — so a restart only changes what the caller explicitly asks to change.

### Implementation correctness

**`extractPlistEnvVars`** — the regex approach is explicitly scoped to the exact XML shape `renderPlist` emits, which is the right call (no external XML parser dependency, can't produce malformed output). The `try/catch → {}` fallback keeps the result-only contract the rest of the file relies on. Three things check out:
- The `EnvironmentVariables` section regex is non-greedy (`[\s\S]*?`) and won't bleed into sibling `<dict>` blocks.
- `renderPlist` sorts keys alphabetically, so idempotency holds: a second call to `upgradeService` with no opts will see `current === desired` and return `already-current` rather than rewriting the plist on every restart.
- The outer `try/catch` swallows the potential `re.exec` infinite-loop footgun — not actually reachable on a properly formed regex, but the safety net is there regardless.

**One narrow gap**: `extractPlistEnvVars` unescapes `&amp;`, `&lt;`, `&gt;` but not `&quot;` or `&apos;`. `renderPlist`'s `xmlEscape` doesn't emit those either, so for plists written by AFK this is fine. A hand-edited plist with `&quot;` in a value would silently pass through the literal `&quot;` string after the round-trip. Given the docstring explicitly calls this "deliberately narrow — only handles the exact XML shape `renderPlist` emits", this is an acceptable and documented limitation. A `// Invariant:` comment noting the `&quot;`/`&apos;` gap would close any ambiguity for future readers.

**Ordering change** — `readFileSync` now happens *before* `renderPlist` (previously it was after, in the old diff-then-read order). This is the correct refactor: `preservedEnv` must be available when building `environmentVariables`. The semantics of `current === desired` are unchanged — both strings are still compared after `desired` is rendered with the full merged env.

**The `upgrade` CLI path** (`afk service upgrade`) also benefits: `manager.upgrade()` only passes `environment` when the caller supplied it, so an `afk service upgrade telegram` with no `--environment` flag will also preserve installed env vars. This is the right behavior.

**The `opts.environment` wins entirely** path is correctly tested: the test seeds a plist with `OLD_TOKEN` and then calls `upgradeService` with `environment: { NEW_VAR: 'new-value' }`, asserting `OLD_TOKEN` is gone. The spread order `{ PATH, ...preservedEnv, ...(opts.environment ?? {}) }` ensures caller-supplied values win unconditionally.

### Tests

Coverage is thorough:
- `extractPlistEnvVars` unit tests cover happy path, empty plist, unparseable input, and XML-escaped characters — all four branches that matter.
- The two `upgradeService` integration tests (preserve on bare restart; caller-supplied env wins) directly exercise the two semantic cases from the issue.
- The `require('fs')` workaround for seeding the plist in tests is a bit awkward (mock boundaries), but it works within the existing test harness pattern and is isolated to tests.

### Constraints

- File stays at ~209 code lines, well within the 350-line ceiling (total 346 physical lines including comments/blanks as the PR notes).
- No new `@anthropic-ai/sdk` imports. No new `process.env` reads. No new paths hand-joined outside `src/paths.ts`.
- `systemd` path is correctly untouched — `systemctl restart` leaves the unit file in place, so env vars are never disturbed there.

### Minor observation

The docstring on `upgradeService` was condensed from a long `Invariant:` block (explaining write atomicity, TOCTOU, etc.) to a short three-liner. The atomicity and TOCTOU rationale are important for future maintainers — they're still present in the code (as inline comments inside the write block), so this is acceptable, but the full inline docstring on `installService` remains intact while `upgradeService`'s is now asymmetrically brief. Not a blocker; worth noting.

### Summary

The fix is correct, well-scoped, and well-tested. The core behavior (preserve existing env vars on bare restart; caller wins when explicit; idempotent round-trips) all hold up. Ready to merge.

Issue #1644, opened 2026-09-13
dashboard: load shiki grammars on-demand per code block


griffinwork40:
## Context

`dashboard/src/components/code-block.tsx` eagerly loads 21 shiki language grammars on first `CodeBlock` mount via `createHighlighter()`. This is ~600KB-1MB of grammar data behind a single dynamic import.

## Current behavior

All 21 grammars load in one batch on first code block render, regardless of which language is actually needed.

## Suggested improvement

Use shiki's `loadLanguage()` API to load grammars on-demand per code block. Only load the grammar for the language actually requested, falling back to `plaintext` while loading.

This would reduce per-user download by ~70% for sessions that only contain 1-2 languages (the common case).

## Priority

Low -- the current approach works correctly and the load is lazy (only triggered when a code block is actually rendered). This is a performance optimization, not a correctness fix.

PR #1819, opened 2026-09-19
fix(dashboard): fix 3 bugs in on-demand shiki grammar loading

griffinwork40:
## Summary

Follow-up to f0ff4c48 (#1644) -- fixes 3 bugs discovered by post-merge verification in the on-demand shiki grammar loading.

### Bug 1 (high): Syntax highlighting never appears after grammar loads

`setHtml(null)` in the grammar-loaded callback caused a re-render, but the `useEffect` dependency array `[code, language]` was unchanged, so the effect never re-ran. Every first-seen language permanently showed the plaintext `<pre>` fallback.

**Fix**: Add a `reHighlightTick` counter state to the dependency array. When the grammar resolves, increment the tick instead of clearing html, causing the effect to re-run with the now-loaded grammar.

### Bug 2 (medium): Special langs (plaintext/ansi/text) always took the wrong branch

shiki's `getLoadedLanguages()` does not include built-in special languages even after explicit init. `loadedLangs.includes("plaintext")` always returned false, pushing special langs into `ensureLanguageLoaded` unnecessarily and triggering Bug 1's stale-effect.

**Fix**: Check `isSpecialLang(language)` before the loaded-languages list.

### Bug 3 (low-medium): Failed grammar loads permanently cached

A rejected `ensureLanguageLoaded()` promise stayed in `langLoadPromises` forever. A transient chunk-load failure poisoned that language for the entire session with no retry.

**Fix**: Delete the map entry on rejection before re-throwing.

### Verification

- `npx vite build` (dashboard): passes
- `pnpm lint`: clean


griffinwork40 (commented):

---

## Prior Rounds

No prior automated rounds — this is round 1.

---

## Review — PR #1819 · `fix(dashboard): fix 3 bugs in on-demand shiki grammar loading`

**Reviewed ref:** `5b4ab62277200441666114113717d46dd1b93013`
**Regime:** light (1 file, ~+23 / -8 lines, hotfix)
**Change type:** hotfix
**Stated intent:** Fix 3 bugs in on-demand shiki grammar loading: (1) highlight never appears after grammar load due to stale useEffect deps, (2) special langs always take wrong branch because `getLoadedLanguages()` omits them, (3) failed grammar loads permanently cached as rejected promises.

---

**Severity arithmetic:** 0 critical, 0 high → MERGE (with follow-ups); 1 medium advisory (non-blocking), 1 low advisory (non-blocking).

All three bugs are correctly addressed. The `reHighlightTick` approach is sound, the `isSpecialLang` guard is correct, and the `.catch` eviction fixes the permanent-failure caching. CI is green.

---

## Findings

### ✅ Security — no issues found
`dangerouslySetInnerHTML` is pre-existing and the shiki-only contract is unaffected by this diff. No new injection surface.

### ✅ API-compat — no issues found
`isSpecialLang` is a stable named export from shiki; the dynamic import pattern matches existing usage in `ensureLanguageLoaded` (line 37 at reviewed ref).

### ⚠️ Correctness — 1 medium advisory (non-blocking)

**MEDIUM (advisory, non-blocking) — `setHtml(null)` fires unconditionally on `reHighlightTick` re-trigger, producing a brief flicker**
**File:** `dashboard/src/components/code-block.tsx:120` · **Ref:** `5b4ab62` · citation-type: file-state (verified)

`setHtml(null)` at line 120 executes before `highlight()` on every effect invocation, including when `reHighlightTick` is the only changed dependency. When the tick increments (grammar just loaded), the grammar is already available so `highlight()` resolves quickly, but one render cycle with `html === null` still occurs — briefly flashing the unstyled `<pre>` fallback before the highlighted HTML replaces it.

**Evidence:**
```tsx
setHtml(null);        // line 120 — fires unconditionally
highlight().catch(() => { ... });
```

**Suggestion:** Guard the null-clear so it only fires when `code` or `language` changed (not on tick-only re-runs). E.g. track previous values in a ref and skip the clear when only `reHighlightTick` differs, or pass a flag to `highlight()` to skip the intermediate null.

### ✅ Spec-compliance — intent met
All three stated bugs are implemented at the reviewed ref: (1) `reHighlightTick` counter added to deps with eslint-disable comment, (2) `isSpecialLang()` check precedes `getLoadedLanguages()`, (3) `.catch` block deletes the map entry before re-throwing. No scope creep. No unmet requirement.

### 🟢 Test-coverage — 1 low advisory (non-blocking)

**LOW (advisory, non-blocking) — No tests cover the three fixed code paths**
**File:** `dashboard/src/components/code-block.tsx` · **Ref:** `5b4ab62` · citation-type: absence (confirmed — no dashboard CodeBlock test file exists)

No tests exercise the special-lang branch (`isSpecialLang`), the eviction-on-rejection path (`.catch` in `ensureLanguageLoaded`), or the `reHighlightTick` re-trigger mechanism. All three paths are testable with a mock shiki module.

**Suggestion:** Add unit tests mocking shiki's `isSpecialLang`, simulating a rejected `loadLanguage` promise, and asserting `reHighlightTick` increments when a grammar resolves. Suggest a follow-up issue.

### ✅ Perf-observability — no issues found
The `import('shiki')` dynamic import inside the effect is module-cached by the bundler. No repeated network cost. No hot-path regression.

---

## What Was Not Checked

- Citations verified inline against branch HEAD `5b4ab62277200441666114113717d46dd1b93013`.
- Stated intent: PR #1819 title + body — spec-compliance assessed and met.
- Did not review other dashboard components consuming `CodeBlock` (callers not in scope for this single-file hotfix).
- Did not run the test suite or dashboard build locally; CI results (all green) taken as authoritative.

griffinwork40 (commented):
## Review: Approve

All three fixes are correct. I traced each bug and its fix carefully.

### Bug 1 (high): Effect never re-ran after grammar load — **Fixed correctly**

The old `setHtml(null)` call did trigger a re-render, but because `html` is not in the `useEffect` dependency array, the effect itself never re-executed — so the now-loaded grammar was never used. The `reHighlightTick` counter solves this cleanly: it *is* in the dep array, its increment causes the effect to re-run, and on that re-run `getLoadedLanguages().includes(language)` returns true so `resolvedLang` gets the real language. The eslint-disable comment is warranted (exhaustive-deps would flag a dep that's read-only in the array, not the body) and correctly placed on the `}, [...])` line.

### Bug 2 (medium): Special langs always fell into the load path — **Fixed correctly**

Shiki's `getLoadedLanguages()` genuinely omits built-in special langs (`plaintext`, `ansi`, `text`, `txt`), so `loadedLangs.includes("plaintext")` returns false even though the lang is always available. The `isSpecialLang(language)` short-circuit added before the list check is the right fix. Worth noting: `ensureLanguageLoaded` already had an `isSpecialLang` guard internally (so it would return immediately), but a special lang was still triggering Bug 1's stale-effect loop before the grammar-load promise settled. The component-level check removes special langs from that path entirely.

The dynamic `import('shiki')` inside the effect to pull `isSpecialLang` is fine — the module system caches the module, so it's a synchronous map lookup after the first resolution, not a second network fetch.

### Bug 3 (low-medium): Rejected promise stays in the cache — **Fixed correctly**

The `.catch` handler runs `langLoadPromises.delete(lang)` before rethrowing, so any callers currently awaiting the same in-flight promise see the rejection after the map entry has already been cleared. A subsequent call to `ensureLanguageLoaded` creates a fresh promise, enabling retry. The chaining order (delete → rethrow → p settles as rejected) is deterministic: the Map entry is gone by the time any awaiting caller resumes from the rejected promise and could call back in.

### Minor observation (no action needed)

`reHighlightTick` is a monotonically-increasing integer with no reset between `code`/`language` prop changes. This is fine in practice — each grammar-load completion fires one increment, so a component cycling through N distinct first-seen languages accumulates N ticks. No risk of unbounded re-render storms.

**Verdict: LGTM. All three root causes are correctly diagnosed and the fixes are minimal, targeted, and safe to merge.**

PR #1831, opened 2026-09-19
fix(tui): use physical row count for cursor-follow targetBottomRow

griffinwork40:
## Problem

When typing a long first message in the REPL that wraps to a second line, the `· idle` stage rail duplicates on every spinner tick, flooding the scrollback upward.

**Repro:** Start a fresh REPL session, type a message long enough to wrap past the terminal width. The `· idle` indicator starts duplicating below, pushing content into scrollback.

## Root Cause

In `terminal-compositor.frame.ts`, the `physicalRows` variable used `frameLines.length` (the logical line count -- one entry per `frameLines.push()` call). When the input line wraps to 2+ physical rows via `CupFrameRenderer`'s `wrapToPhysicalLines()`, the logical count (1) under-counts the actual screen footprint.

In cursor-follow mode (idle state, pre-commit), `targetBottomRow` is computed as:
```
targetBottomRow = min(absoluteBottom, anchorRow - 1 + physicalRows)
```

With `physicalRows = 1` (logical) instead of 2+ (physical), `targetBottomRow` is too low. The frame overlaps the DECSTBM reserved footer band where `LoopStageBar` paints `· idle`. Each spinner tick, both the compositor and the footer bar write to overlapping rows, producing cascading duplicate idle lines.

## Fix

Use `CupFrameRenderer.measure()` to get the physical (post-wrap) line count before computing `targetBottomRow`. This is the same `measure()` already used 20 lines later for `desiredTopRow` -- the gap was that `targetBottomRow` was computed from the logical count while `desiredTopRow` used the physical count, creating an inconsistency.

Falls back to logical count for stubs/tests without `measure()`.

## Verification

- Regression test (`terminal-compositor.multiline-idle-dup.test.ts`) verifies `topRow` stability across repaints when input wraps in cursor-follow mode. **Fails without the fix** (topRow drifts between 1 and 2 on consecutive repaints), passes with it.
- All 463 compositor tests pass
- `tsc --noEmit` clean
- File-size gate unaffected


griffinwork40:
Wave 1.5 citation verification is complete. Let me verify the key claims:

**Check A - Citation verification:**
- Finding 1 (`frame.ts:278-280`): **verified** — lines 278-280 at the reviewed ref show `const physicalRows = self.logUpdate.measure ? self.logUpdate.measure(frame, absoluteBottom).lineCount : logicalRows;` and lines 289-290 show `self.logUpdate.measure(frame, targetBottomRow).topRow`. Both `measure()` calls confirmed.
- Finding 1 claim about `lineCount` being `targetBottomRow`-independent: **verified** — `cup-frame-renderer.ts:143` shows `rawLineCount = Math.max(1, CupFrameRenderer.wrapToPhysicalLines(content, width).length)` — no dependency on `targetBottomRow`.
- Finding 2 (`frame.ts:279`): **verified** — the new call passes `absoluteBottom` while the existing call at line 290 passes `targetBottomRow`.
- Finding 3 (`test:65-77`): **verified** — the test assertions are stability-based (`topRow1 === topRow2 === topRow3`), not value-based.

**Check B - Absence-claim verification:**
- Finding 3 claims no direct assertion on `targetBottomRow` in test files: `grep -n 'targetBottomRow' test file` returns only comments (lines 5, 12, 37, 72, 77, 88, 102, 104, 105) — all in JSDoc/comments, zero `expect()` calls on `targetBottomRow`. **confirmed-absent**.

All citations verified. No fabricated or false-absent findings. Now Wave 2 synthesis:

---

## PR #1831 Review — `fix(tui): use physical row count for cursor-follow targetBottomRow`

**Target:** PR fix/repl-multiline-idle-dup → main · +136/-1 · 2 files · hotfix (light regime)
**Reviewed ref:** `c6f2c212cafe94e7f070295f6b6af6038b608ae6`
**Stated intent:** Fix duplicate `· idle` stage rail when REPL input wraps to 2+ physical lines in cursor-follow mode.

### Findings

**1 · medium · blocking:false · confidence:high · correctness**
`src/cli/terminal-compositor.frame.ts:279,290` · ref:c6f2c212 · diff-context + file-state (verified)

`measure()` is called twice per repaint tick during cursor-follow: once at line 279 (`measure(frame, absoluteBottom).lineCount`) and once at line 290 (`measure(frame, targetBottomRow).topRow`). `lineCount` is `targetBottomRow`-independent (`cup-frame-renderer.ts:143`: `rawLineCount = wrapToPhysicalLines(content, width).length`), so both calls pay the full wrap cost to derive an identical `lineCount`. Bounded to cursor-follow mode only (pre-first-commit idle state); post-commit both calls use `targetBottomRow === absoluteBottom` and the redundancy is argument-only.
· waived: bounded, non-data-affecting redundancy confined to pre-commit idle state; double-wrap cost is negligible for typical input lengths.

**Suggestion:** call `measure(frame, absoluteBottom)` once, extract `.lineCount` for `physicalRows`, compute `targetBottomRow`, then derive `desiredTopRow` arithmetically (`max(1, targetBottomRow - lineCount + 1)`) instead of a second `measure()` call.

---

**2 · low · blocking:false · confidence:high · correctness**
`src/cli/terminal-compositor.frame.ts:279` · ref:c6f2c212 · diff-context (verified)

The new `measure()` call passes `absoluteBottom` where the `LogUpdateFn.measure` parameter is named `targetBottomRow`. The value difference is semantically inert (`lineCount` is independent of it), but the argument name/contract mismatch creates a maintenance landmine if `measure()` ever gains `targetBottomRow`-dependent logic.

**Suggestion:** if adopting Finding 1's single-call restructure, this resolves automatically.

---

**3 · low · blocking:false · confidence:high · test-coverage**
`src/cli/terminal-compositor.multiline-idle-dup.test.ts:113-150` · ref:c6f2c212 · diff-context (verified)

The regression test asserts topRow stability across repaints (no drift) but not topRow correctness (specific value). A misconfigured `measure()` stub producing a wrong-but-stable `lineCount` would pass all three assertions. The test catches the reported symptom (drift/duplication) but not the root mechanism (correct `targetBottomRow` computation).

**Suggestion:** add an assertion like `expect(topRow).toBeLessThanOrEqual(anchorRow)` to pin the value direction (frame top extends upward when input wraps).

---

- **Security** — no issues found; read `terminal-compositor.frame.ts`. Pure terminal row arithmetic, no auth/data/secret surface.
- **API compatibility** — no issues found; `LogUpdateFn.measure` was already optional; the fallback preserves the pre-fix contract for stubs/tests.
- **Spec-compliance** — stated intent fully met: `physicalRows` now uses `measure().lineCount` with a fallback. No unmet requirements. No scope creep (picker path unaffected — uses `absoluteBottom` unconditionally).
- **Perf/observability** — see Finding 1 (double wrap). No new I/O, no unbounded loops.

### What was not checked

- Citations verified inline against branch HEAD `c6f2c212cafe94e7f070295f6b6af6038b608ae6`.
- Stated intent: PR #1831 title+body — spec-compliance assessed.
- Did not run tests (read-only review). PR body claims all 463 compositor tests pass + `tsc --noEmit` clean.
- Did not verify the `CupFrameRenderer.measure()` implementation beyond the `lineCount`-independence check at `cup-frame-renderer.ts:141-145`.

### Decision: MERGE — 0 blocking (1 medium waived, 2 low).

The fix is correct, precisely scoped, well-documented (12-line invariant comment), and includes a targeted regression test. The waived medium (double `measure()` call) is a minor optimization opportunity, not a correctness or safety concern.

**Done**
- What was done: Reviewed PR #1831 across all six dimensions (security, api-compat, correctness, spec-compliance, test-coverage, perf-observability) using the light-regime review pipeline with inline citation verification.
- Evidence: Review above with all citations verified against reviewed ref `c6f2c212`.
- What changed: nothing — read-only review, no mutations.
- Deferred: none.

---

PR #1872, opened 2026-09-20
fix: replace local sleep with sleepWithAbort in rate-limit-bucket (#1850)

### LINKED ISSUE #1850: fix: rate-limit-bucket sleep() missing .unref() and early-abort check
`src/agent/providers/shared/rate-limit-bucket.ts:58` has a local `sleep()` function:

```typescript
function sleep(ms: number): Promise<void> {
  return new Promise((resolve) => setTimeout(resolve, Math.max(0, ms)));
}
```

This reimplements `sleepWithAbort` from `src/agent/providers/shared/sleep-with-abort.ts` but is missing:

1. **`timer.unref()`** -- an active rate-limit sleep timer can prevent Node from exiting cleanly
2. **Early-abort check** (`if (signal.aborted) { resolve(); return; }`) -- there is a missed-signal window
3. **Proper cleanup** -- the manual `Promise.race` wrappers at lines 202 and 245 do not call `removeEventListener` on the abort handler

The canonical `sleepWithAbort` already handles all three cases correctly. The `sleep-with-abort.ts` docstring even notes: "Previously duplicated verbatim in anthropic-direct/loop.ts and openai-compatible/query.ts" -- but `rate-limit-bucket.ts` was missed.

## Fix

Replace the local `sleep()` with imports of `sleepWithAbort` at the two call sites (lines 202/204 and 245/247). Where no signal is available (line 204, 247), either pass a no-op signal or keep a minimal `sleep` that at least calls `.unref()`.

Shadow-verified: the `.unref()` omission and inferior abort handling are confirmed.

griffinwork40:
## Summary

Fixes three bugs in the `sleep()` helper inside `rate-limit-bucket.ts`.

### Bugs fixed

| # | Bug | Impact |
|---|-----|--------|
| 1 | `timer.unref()` missing | Active `setTimeout` kept the Node.js event loop alive after all real work finished, preventing clean process exit |
| 2 | Missed-signal window | A signal fired between the `signal.aborted` guard and `addEventListener` was silently dropped; the waiter stayed parked until the timer expired |
| 3 | Listener leak | The abort listener added in `new Promise<void>` was never removed via `removeEventListener` when the timer won the race |

### Fix

Removed the local `sleep()` function entirely. Replaced both `Promise.race([sleep(ms), abortPromise])` blocks with `sleepWithAbort(ms, signal)` from `./sleep-with-abort.ts`, which already handles all three correctly:

- Checks `signal.aborted` before setting the timer (fixes missed-signal window)
- Calls `timer.unref()` (fixes event-loop hold)
- Calls `signal.removeEventListener('abort', onAbort)` when the timer fires (fixes listener leak)

When no `AbortSignal` is in scope, `new AbortController().signal` (a never-firing signal) is passed — no change in sleep semantics, full `.unref()` benefit.

### Testing

- `pnpm lint` — clean (TypeScript no-emit)
- `pnpm vitest run rate-limit-bucket` — 32/32 tests pass

Closes #1850


griffinwork40 (commented):

---

**Reviewed ref:** `428343561348d20ab1f0a7d2434ef20f9ca0737d`

## Prior Automated Rounds

None — this is the first automated review of PR #1872. No prior findings to carry forward.

---

## Severity Floor Applied

0 critical, 0 high → **MERGE (with follow-ups)**; 1 medium waived (non-blocking advisory); 7 low advisory; 4 nit advisory.

Only `critical` and `high` findings block merge. The 1 medium finding (T-1) is waived as a bounded, non-data-affecting test-coverage gap with a recommended follow-up issue rather than a fix commit.

---

**Counts:** 0 critical, 0 high, 0 blocking medium (1 medium waived — test gap, non-data-affecting, CI green), 7 low, 4 nit.

The change correctly deletes the local 3-line `sleep()` helper in `rate-limit-bucket.ts` and replaces both `acquirePermit` sleep sites with `sleepWithAbort(ms, signal ?? new AbortController().signal)` from `./sleep-with-abort.ts`. All three bugs from the stated intent are verified fixed: `timer.unref()` ✓, pre-abort check in `sleepWithAbort` ✓, `removeEventListener` cleanup on timer-win ✓. The `new AbortController().signal` path for callers that pass no signal is functionally identical to the old `sleep(ms)` plus the `unref()` improvement. CI is green (all checks SUCCESS). No breaking API changes.

---

## Findings

### MEDIUM — Advisory (non-blocking)

**[MEDIUM · waived] test-coverage — Bug 2's `sleepWithAbort`-internal abort check has no dedicated test path**
- `rate-limit-bucket.test.ts:115` · ref:`428343561348d20ab1f0a7d2434ef20f9ca0737d` · file-state (verified)
- The test at line 115 ("resolves immediately when signal is already aborted") pre-aborts the signal, so `acquirePermit` returns from the loop-top `if (signal?.aborted) return;` guard before reaching `sleepWithAbort`. Tests at lines 125/135 abort 20 ms after the call starts (the listener path). No test exercises `sleepWithAbort` with a signal that is already aborted at the time `sleepWithAbort` is called — i.e., the case where the signal fires between the loop-top guard and the sleep invocation, which Bug 2's fix specifically addresses. The implementation is correct; `sleep-with-abort.ts:21` (`if (signal.aborted) { resolve(); return; }`) is the correct fix. This is a test coverage gap only.
- **Waived:** bounded, non-data-affecting; implementation verified correct; CI green. Suggested follow-up: add a test that reaches `sleepWithAbort` with a pre-aborted signal (e.g., freeze the bucket far in the future so the loop-top guard doesn't fire, then call `acquirePermit` with an already-aborted signal). A standalone `sleep-with-abort.test.ts` unit test would also close this.

---

### LOW — Advisory (non-blocking)

**[LOW] security — `new AbortController().signal` allocation per sleep iteration**
- `rate-limit-bucket.ts:198, 237` · ref:`428343561348d20ab1f0a7d2434ef20f9ca0737d` · diff-context
- One `AbortController` object is created per `sleepWithAbort` call when no `signal` is supplied. This occurs only under rate-limiting, which is rare. Allocation is trivially small (<1 µs, immediately GC-eligible); no auth or bypass risk (signals flow from HTTP request signals in tracing-fetch.ts; confirmed via grep). Optional micro-optimisation: a module-level `const NEVER_ABORT = new AbortController().signal;` singleton.

**[LOW] security — Commit message overstates missed-signal window fix**
- PR body/commit message
- The PR description states Bug 2 (missed-signal window) is "closed." The race is structurally present in `sleepWithAbort` but unreachable in JS's single-threaded event model (no microtask can preempt the synchronous lines between the `signal.aborted` guard and `addEventListener` within the same Promise constructor callback). The pre-abort guard is belt-and-suspenders, not a true race-close. No runtime defect; commit message accuracy only.

**[LOW] correctness — `ms` clamping removal is safe**
- `rate-limit-bucket.ts:198, 237` · diff-context
- The deleted `sleep()` applied `Math.max(0, ms)`; `sleepWithAbort` does not. Negative `ms` is structurally unreachable: the freeze path's `waitMs = frozenUntil - now` is guaranteed positive by the enclosing `if (this.frozenUntil > now)` guard; the sleep path's `sleepMs = Math.max(1, nextReset - now) + staggerJitterMs()` floors at 1. No defect.

**[LOW] perf-observability — `acquirePermit` proactive blocking emits no trace/log**
- `rate-limit-bucket.ts:190–250` · file-state (verified)
- When `acquirePermit` parks a caller for seconds (freeze path: up to 120 s; window-reset path: up to 60 s + jitter), no log, metric, or event is emitted. Pre-existing gap not introduced by this diff. Suggested follow-up: emit a structured log at `debug` or `info` level when `waitMs > 500 ms`, including duration, path (freeze vs. window-reset), and signal source — making proactive rate-limiting visible in `afk trace show`.

**[LOW] test-coverage — `sleep-with-abort.test.ts` absent**
- `src/agent/providers/shared/sleep-with-abort.test.ts` · confirmed absent at reviewed ref
- No dedicated unit test file for `sleepWithAbort` in isolation. As a shared utility now consumed by multiple providers, standalone tests covering: (a) pre-aborted signal resolves immediately, (b) signal fires during sleep, (c) timer fires normally, and (d) `clearTimeout` called on abort would close this gap. Suggested follow-up issue.

**[LOW] test-coverage — `clearTimeout` on abort has no timer-cleanup assertion**
- `rate-limit-bucket.test.ts:125–143` · file-state (verified)
- The two abort-while-waiting tests (lines 125, 135) verify that `acquirePermit` resolves quickly when a signal fires, but neither verifies that the `setTimeout` timer was actually cancelled by `clearTimeout`. The tests cannot distinguish "timer cleared immediately on abort" from "timer left running, second callback ignored." Suggested: a `vi.useFakeTimers()` unit test in `sleep-with-abort.test.ts` asserting no additional callbacks fire after abort.

**[LOW] test-coverage — No-signal path tests use real wall-clock timers**
- `rate-limit-bucket.test.ts:149–185` · file-state (verified)
- The two "no signal" tests use real 50–60 ms `setTimeout` delays. Functional but adds real wall time to the test suite. Wide bounds (≥40 ms, <2000 ms) make them pass in CI, but `vi.useFakeTimers()` would be faster and more deterministic. No change required for this PR; flagged for awareness.

---

### NIT — Advisory (non-blocking)

**[NIT] api-compat — `sleep()` was never exported; deletion has no external impact**
- Grep pattern `^export.*function sleep` in `rate-limit-bucket.ts` → zero matches. No external importers affected.

**[NIT] api-compat — Public API surface unchanged**
- `acquirePermit(estimatedInputTokens, signal?)`, `RateLimitBucket`, `globalRateLimitBucket`, `RateLimitSnapshot`, `estimateInputTokens` — all preserved with identical signatures.

**[NIT] perf-observability — `new AbortController()` per sleep iteration: negligible**
- Rate-limiting is rare; allocation is orders of magnitude below the sleep duration (5 s–120 s). No concern.

**[NIT] correctness — `clearTimeout` in timer-win path is correct, not redundant**
- `sleep-with-abort.ts:23` · file-state (verified: `signal.removeEventListener('abort', onAbort)` in timer callback). `{ once: true }` removes the listener only when the `'abort'` event fires; the explicit `removeEventListener` in the timer callback prevents `onAbort` from lingering on long-lived signals that fire much later. Both paths correct.

---

## Positive Observations

- ✅ All three stated bugs verified fixed: `timer.unref()` ✓, pre-abort early-exit ✓, `removeEventListener` on timer-win ✓.
- ✅ Zero behavioral change to `RateLimitBucket`'s public API or `acquirePermit` semantics.
- ✅ Net −18 LOC. Dead code removed cleanly.
- ✅ `clearTimeout` on abort is an improvement over the old `Promise.race` (eliminates dangling timer reference).
- ✅ CI green: Lint & Build ✅, Test ubuntu/macos ✅, Smoke ✅, Docs ✅, Dashboard ✅, Publish Bundle Smoke ✅.

---

## Recommended Follow-up Issues (non-blocking)

1. Add `src/agent/providers/shared/sleep-with-abort.test.ts` — unit tests for `sleepWithAbort` in isolation.
2. Add a test reaching `sleepWithAbort` with a pre-aborted signal from the `acquirePermit` integration level (freeze bucket, pre-abort signal, verify resolution without wait).
3. Add structured logging in `acquirePermit` when blocking for >500 ms (proactive rate-limit observability).

---

## What Was Not Checked

- Citations verified against branch HEAD `428343561348d20ab1f0a7d2434ef20f9ca0737d` (pre-fetched via GitHub API).
- Stated intent: PR #1872 title + body — spec-compliance assessed; all stated changes delivered.
- Absence claims confirmed: `sleep-with-abort.test.ts` absent (HTTP 404 at reviewed ref); no test exercises `sleepWithAbort`-internal abort path (confirmed by test file inspection at reviewed ref).
- Did not run tests locally; build success inferred from CI.
- Did not audit all consumers of `sleepWithAbort` across the codebase (anthropic-direct/loop.ts, openai-compatible/query.ts) — those are pre-existing consumers unaffected by this diff.

PR #1874, opened 2026-09-20
fix: use shared parseRetryAfterMs in OpenAI provider (#1851)

### LINKED ISSUE #1851: fix: openai tracing-fetch parseRetryAfterMs missing retry-after-ms header
`src/agent/providers/openai-compatible/tracing-fetch.ts:37-47` has a local `parseRetryAfterMs(headers: Headers)` that reimplements the canonical version in `src/agent/providers/shared/retry-after.ts` but is **missing the `retry-after-ms` header check**.

## The bug

The shared version checks `retry-after-ms` first (the OpenAI-specific millisecond header), then falls back to `retry-after`. The local copy only checks `retry-after`. This means the OpenAI provider ignores the more precise millisecond hint when OpenAI sends it.

Additionally, `tracing-fetch.ts` duplicates two other shared constructs:

- `THROTTLE_STATUSES` -- identical `new Set([429, 503, 529])` in both provider tracing-fetch files
- `OpenAIRateLimitGate` interface -- same shape as the Anthropic `RateLimitGate`, and the file already cross-imports `ThrottleInfo` from the Anthropic version

## Fix

1. Delete the local `parseRetryAfterMs` and adapt the call site to use the shared version (the shared `getHeader` already handles `Headers` objects via `typeof h.get === 'function'`)
2. Move `THROTTLE_STATUSES` to `src/agent/providers/shared/tracing-fetch-utils.ts`
3. Merge the gate interfaces into one shared type

Net result: ~45 lines removed, retry-after-ms bug fixed.

Shadow-verified: the missing `retry-after-ms` check is confirmed.

griffinwork40:
## Summary

Closes #1851

The local `parseRetryAfterMs` in `src/agent/providers/openai-compatible/tracing-fetch.ts` was missing a check for the `retry-after-ms` header. OpenAI returns this header with a millisecond-precision backoff hint, but the local copy only read `retry-after` (seconds), so the precision was silently discarded — the freeze duration was always rounded up to the nearest second.

## Changes

- **Deleted** the local `parseRetryAfterMs` in `openai-compatible/tracing-fetch.ts` (16 lines removed)
- **Imported** the canonical `parseRetryAfterMs` from `../shared/retry-after.js`
- **Updated** both call sites to pass `{ headers: res.headers }` to match the shared function's `error: unknown` signature (which `getHeader` handles for both `Headers` objects and plain records)

## Before / After

| | Before | After |
|---|---|---|
| `retry-after-ms: 1234` | Ignored — fell through to `retry-after` | Parsed as `1234 ms` |
| `retry-after: 5` | `5000 ms` | `5000 ms` (unchanged) |
| Neither header | `undefined` → `5_000` default | same |

## Verification

- `pnpm lint` — passes (TypeScript clean)  
- `pnpm test` — 20 041 tests pass, 1 skipped  
- `pnpm audit:filesize:check` — pre-existing violation in `src/config/env.ts` on main; our change is net -13 lines in the touched file


griffinwork40 (commented):

---

**Reviewed ref:** `a1bb93425f7d60e834f1cc0c03230bfd1d16d7e6`

## Prior automated rounds

---

## Severity floor applied

0 critical, 0 high → **MERGE (with follow-ups)**; 1 low advisory finding.
Only `critical` and `high` findings block in this sweep.

---

**0 blocking findings.** The change correctly deletes the local `parseRetryAfterMs` (which ignored `retry-after-ms`) and imports the shared implementation that reads `retry-after-ms` first, then falls back to `retry-after`. Both call sites (429 gate freeze at line 82 and throttle info at line 92) correctly wrap `res.headers` as `{ headers: res.headers }` to match the shared function's `error: unknown` signature — `getHeader` finds the `.get()` method on the `Headers` instance and the parse succeeds.

**Dimension summary:**
- **Security** — no issues found. No auth surface touched; `getHeader` defensively rejects non-objects.
- **API compatibility** — no issues found. `makeOpenAITracingFetch` exported signature unchanged; `parseRetryAfterMs` was not exported from this file.
- **Correctness** — no issues found. The `{ headers: res.headers }` wrapper matches `getHeader`'s object-with-`.headers` probe; `Headers` instance supports `.get()`.
- **Spec-compliance** — all three stated intent points delivered: local copy deleted, shared import wired, both call sites adapted. No scope creep.
- **Perf/observability** — no issues found. One extra property wrapper object per throttled response; negligible.
- **Test coverage** — 1 advisory finding (see below).

CI: all checks pass (Lint & Build ✅, Test ubuntu/macos ✅, Smoke ✅, Docs ✅). `mergeable: CONFLICTING` — merge conflict with main needs resolving before merge.

---

## Findings (advisory — non-blocking)

**[LOW] test-coverage — No `retry-after-ms` test at the tracing-fetch call site**

The stated fix motivation is that `retry-after-ms` precision was silently discarded at this call site. The shared `parseRetryAfterMs` is tested for `retry-after-ms` in `openai-compatible/query/retry.test.ts:37-40`, but that exercises a separate code path (`retryAfterDelayMs`). No test in `tracing-fetch.test.ts` passes a `retry-after-ms` header to the `gate.freeze` path (line 82) or the `onThrottle` path (line 92), so a regression at the call site would not be caught.

→ Advisory only. Suggested follow-up: add two cases to `tracing-fetch.test.ts` asserting that (a) `gate.freeze` receives ms-precision from `retry-after-ms`, and (b) `onThrottle` prefers `retry-after-ms` over `retry-after`.

---

## What was not checked

- Citations verified against branch HEAD `a1bb93425f7d60e834f1cc0c03230bfd1d16d7e6`.
- Stated intent: PR #1874 title + body — spec-compliance assessed.
- Absence claim (no `retry-after-ms` test in `tracing-fetch.test.ts`) verified via `git grep` at the reviewed ref: 0 matches.
- Did not assess whether the OpenAI SDK emits `retry-after-ms` in the header shape that `getHeader` expects (`APIError.headers` vs. `Headers`).
- Did not run tests or build locally; build success inferred from CI.
- Note: PR has a merge conflict with main (`mergeable: CONFLICTING`) — the conflict is outside review scope.

griffinwork40 (commented):

---

## Prior Rounds
This is the **first** automated review of PR #1874 — no prior findings to carry forward.

---

**Severity arithmetic:** 0 critical, 0 high → MERGE (with follow-ups); 3 low advisory, 4 nit advisory.

Only `critical` and `high` findings block merge. All findings here are `low` or `nit` and are listed as non-blocking advisory items. A follow-up issue is suggested rather than a fix commit.

---

## Findings Summary

The primary fix is **correct**: `{ headers: res.headers }` passes a `Headers`-object wrapper to `parseRetryAfterMs`, which `getHeader` handles via the `.get()` branch. The `retry-after-ms` header is now correctly parsed ahead of `retry-after`, matching the Before/After table in the PR body. No blocking issues found.

### Low — Advisory (non-blocking)

**L1 · correctness** — `retry-after-ms` float values are not rounded before `gate.freeze()`
- `src/agent/providers/shared/retry-after.ts:54–55` · ref:a1bb9342 · file-state
- The shared `parseRetryAfterMs` returns the raw `Number(ms)` for `retry-after-ms` without `Math.round()` (line 55: `if (Number.isFinite(n) && n >= 0) return n;`). The old local function always rounded (seconds → ms). A server sending `retry-after-ms: 1234.7` passes a float into `gate.freeze()`. Benign in practice — the `RETRY_AFTER_MAX_WAIT_MS` cap in `rate-limit-bucket.ts` clamps the value — but inconsistent with the seconds path which calls `Math.round(n * 1000)`.
- **Suggestion (follow-up):** Add `return Math.round(n);` in the `retry-after-ms` branch of `parseRetryAfterMs`.

**L2 · test-coverage** — No regression test added for the exact bug fixed (`retry-after-ms` in `tracing-fetch.test.ts`)
- `src/agent/providers/openai-compatible/tracing-fetch.test.ts:67,98,104` · file-state [absence confirmed]
- The three existing tests only exercise `retry-after` (seconds) and the no-header fallback. The fix that caused this PR — `retry-after-ms` was previously ignored — has no covering test. A future refactor breaking that branch would not be caught.
- **Suggestion (follow-up):** Add tests asserting `gate.freeze(1234)` and `onThrottle({ retryAfterMs: 1234 })` when `retry-after-ms: 1234` is returned on a 429.

**L3 · test-coverage** — `shared/retry-after.ts` has no dedicated unit test file
- `src/agent/providers/shared/retry-after.ts` · file-state [absence confirmed — no `retry-after.test.ts` in shared/]
- The module is now the canonical implementation for both providers. Edge cases (`retry-after-ms` on plain-record headers, negative ms, past HTTP-date, invalid ms string) are not independently verified.
- **Suggestion (follow-up):** Add `src/agent/providers/shared/retry-after.test.ts`.

### Nit — Advisory (non-blocking)

**N1 · correctness** — Past HTTP-date in `retry-after` returns `undefined` → falls back to 5 s freeze
- `src/agent/providers/shared/retry-after.ts:63–64` · file-state
- Pre-existing behavior inherited from the old local function, not introduced by this PR. `delta = dateMs - Date.now()` can be negative, causing `undefined` return and a 5 s default freeze instead of "retry immediately." No regression.

**N2 · spec-compliance** — Stated intent fully met; no scope creep
- All three stated changes are present: local function deleted, shared import added, both call sites updated to `{ headers: res.headers }`. PR body's Before/After table is accurate.

**N3 · perf-observability** — Double `parseRetryAfterMs` call on 429 when both `gate` and `onThrottle` are set
- Pre-existing pattern. The call is cheap (two header reads, no I/O). Not a regression.

**N4 · api-compat** — New import is from the correct shared layer; pre-existing `ThrottleInfo` cross-import is untouched
- `import { parseRetryAfterMs } from '../shared/retry-after.js'` is architecturally appropriate. No api-compat concern introduced.

---

## What Was Not Checked
- Citations verified inline against branch HEAD `a1bb93425f7d60e834f1cc0c03230bfd1d16d7e6`.
- Absence claims for F2 and F3 confirmed via directory listing and grep of `tracing-fetch.test.ts` at the reviewed ref.
- Did not review the Anthropic-direct provider surface (out of scope for this diff).
- Did not run tests locally — PR body states 20,041 tests pass; CI shows green.
- PR #1874 has a merge conflict (`mergeable: CONFLICTING`) — the review covers the code correctness only; the conflict must be resolved before merge.

---

> **Note:** This PR has a **merge conflict** with `main`. Resolve the conflict before merging; the code changes themselves are correct.

PR #1879, opened 2026-09-20
refactor: extract toggleScheduleEnabled, fix POST body drift (#1838)

### LINKED ISSUE #1838: refactor: schedule enable/disable — extract toggleScheduleEnabled(), fix POST body drift
## refactor: `schedule enable/disable` — extract `toggleScheduleEnabled()`, fix POST body drift

**Audit finding:** Item #17 (Tier 2 — Moderate Extractions)
**Identified in:** Duplicate Code Audit, 2026-09-19

### What's duplicated

The toggle-and-save logic for enabling/disabling a schedule is duplicated in three places:

| File | Lines | Notes |
|---|---|---|
| `cli/commands/schedule.ts` | 136–160 | `enable` subcommand |
| `cli/commands/schedule.ts` | 164–185 | `disable` subcommand |
| `agent/tools/handlers/schedules.ts` | 207–220 | tool handler |

**Total duplicated LOC:** ~50 (plus diverged POST body construction)

### Bug: POST body drift

The CLI `enable`/`disable` commands build the POST body differently from the tool handler — **the CLI omits the `executor` field**. This means enabling/disabling a schedule via the CLI can silently strip the executor from the stored config, producing a different schedule state than the same action taken through the agent tool.

### Recommended fix

Extract `toggleScheduleEnabled(id: string, enabled: boolean): Promise<void>` to `schedule-store.ts` (or a thin service layer). All three callsites delegate to this single implementation. The POST body is constructed in one place, eliminating the field-drift divergence.

This is both a code-quality improvement (removes ~50 LOC of duplication) and a correctness fix (the `executor` field omission is a latent data-corruption bug).

griffinwork40:
## Summary

Extracts `toggleScheduleEnabled()` into the schedule store and fixes a latent bug where the CLI's enable/disable commands silently stripped the `executor` field from schedule configs.

## Problem

Toggle-and-save logic was duplicated in 3 places:
- `cli/commands/schedule.ts` enable subcommand
- `cli/commands/schedule.ts` disable subcommand
- `agent/tools/handlers/schedules.ts` tool handler

**Bug:** The CLI commands built the POST body without the `executor` field, so re-enabling a `shell`-executor task via CLI would silently revert it to `agent` in the daemon.

## Fix

Added `toggleScheduleEnabled(id, enabled)` to `schedule-store.ts` — atomically loads, updates `enabled` + `updatedAt`, and saves. All 3 callsites now delegate to this function. The executor field (and all other fields) are preserved through the round-trip.

## Changes

- `src/agent/daemon/schedule-store.ts` — new `toggleScheduleEnabled()` function
- `src/cli/commands/schedule.ts` — both subcommands use shared function
- `src/agent/tools/handlers/schedules.ts` — handler uses shared function
- `src/agent/daemon/schedule-store.test.ts` — 5 new tests (enable, disable, field preservation, unknown-id)

## Verification

- 35/35 schedule-store tests pass
- 23/23 schedule handler tests pass
- 3/3 schedule CLI tests pass
- `pnpm lint` clean

Closes #1838


griffinwork40 (commented):

---

**Reviewed ref:** `399e4c85b26eb6c502c101f64f51aec0112d31d9`

## Prior automated rounds

---

## Severity floor applied

0 critical, 0 high → **MERGE (with follow-ups)**; 2 medium advisory, 1 low advisory, 1 nit advisory.
Only `critical` and `high` findings block in this sweep.

---

**0 blocking findings.** The refactor correctly extracts `toggleScheduleEnabled` as a shared function, deduplicating toggle-and-save logic across the CLI enable/disable subcommands and the `cancelScheduleHandler` tool handler. The stated executor-field-drift bug is fixed: all three callsites now delegate persistence to `toggleScheduleEnabled` (field-preserving spread) and build the daemon POST body via `toScheduledTask` (includes the `executor` conditional spread). CI is green (Lint & Build ✅, Test ubuntu/macos ✅, Smoke ✅, Docs ✅, Dashboard ✅). 5 new tests cover enable, disable, field preservation, unknown-id, and executor round-trip.

**Dimension summary:**
- **Security** — no issues found.
- **API compatibility** — no issues found at blocking severity. `toScheduledTask`'s `cron` → `cronExpression` key mapping is pre-existing behavior (function predates this PR); no new API surface introduced.
- **Correctness** — 2 medium advisory findings (see below). Both are narrow-condition issues, one pre-existing.
- **Spec-compliance** — all stated intent points delivered: toggle logic deduplicated, executor field preserved through round-trip, 5 tests added. No scope creep.
- **Test coverage** — 1 low advisory finding. The new unit tests cover `toggleScheduleEnabled` in isolation; no handler-level integration test for the race/crash path.
- **Perf/observability** — no issues found. Same I/O pattern as the old inline code (one `readFileSync` + one atomic write per toggle).

---

## Findings (advisory — non-blocking)

**[MEDIUM] correctness — `updated!` non-null assertion in cancelScheduleHandler enable branch is unguarded against concurrent store mutation**

`schedules.ts:207-210` · ref:`399e4c85` · diff-context

The `cancelScheduleHandler` checks `getSchedule(taskId)` (line ~193), then calls `toggleScheduleEnabled(taskId, true)` (line ~207). Both call `loadSchedules()` → `readFileSync`. If a concurrent CLI process removes the task from `schedules.json` between the two synchronous reads, `toggleScheduleEnabled` returns `undefined` and the `!` assertion on line 210 throws an uncaught TypeError instead of returning a structured `{ isError: true }` response.

```typescript
const updated = toggleScheduleEnabled(taskId, true);
sync = await trySyncToDaemon('POST', '/tasks', toScheduledTask(updated!));
```

This requires concurrent multi-process file access between two synchronous calls in the same event loop tick — extremely narrow. Classified as medium (rare conditions), not high.

→ Advisory only. Suggested follow-up: replace `updated!` with an explicit guard: `if (!updated) return { content: JSON.stringify({ error: 'task disappeared during toggle' }), isError: true };`

**[MEDIUM · pre-existing] correctness — cancelScheduleHandler has no try/catch around toggleScheduleEnabled I/O**

`schedules.ts:207,213` · ref:`399e4c85` · file-state (verified)

`toggleScheduleEnabled` calls `saveSchedules`, which calls `writeFileSync` and `renameSync`. A disk I/O failure produces an unhandled rejection instead of a structured `{ isError: true }` response. This gap is **pre-existing** — the old inline code had the same pattern (direct `saveSchedules` call with no try/catch). This PR does not worsen it.

→ Advisory only (pre-existing). Suggested follow-up: wrap the toggle + sync block in a try/catch returning a structured error.

**[LOW] test-coverage — no integration-level test for cancelScheduleHandler's toggle path**

`schedule-store.test.ts` · ref:`399e4c85` · file-state (verified)

The 4 new `toggleScheduleEnabled` unit tests cover the store function in isolation. No test exercises `cancelScheduleHandler` calling `toggleScheduleEnabled` and verifying the daemon sync POST body includes the `executor` field — the exact regression this PR fixes.

→ Advisory only. Suggested follow-up: add a `cancelScheduleHandler` test that enables a `shell`-executor task and asserts the `trySyncToDaemon` call includes `executor: 'shell'`.

**[NIT] correctness — misleading invariant comment**

`schedules.ts:208-209` · ref:`399e4c85` · diff-context

The comment `// updated is always defined here: existing was confirmed above and the store is consistent` asserts an invariant that is not enforced by a lock. Replace with `// expected to be defined — getSchedule confirmed above` or remove in favor of a guard.

---

## What was not checked

- Citations verified inline against branch HEAD `399e4c85b26eb6c502c101f64f51aec0112d31d9`.
- Stated intent: PR #1879 title + body — spec-compliance assessed; all stated changes delivered.
- Did not run tests locally; build success inferred from CI.
- Did not audit the `schedule add` CLI subcommand (not touched by this diff — it builds its own POST body inline, which also omits `executor` in the sync call, but that is out-of-scope for this PR's changes).
- Absence claims: `cancelScheduleHandler` try/catch absence confirmed via grep at reviewed ref (zero matches for `try` in the handler function body).

griffinwork40 (commented):

---

**Reviewed ref:** `399e4c85b26eb6c502c101f64f51aec0112d31d9`

## Prior automated rounds

---

## Severity floor applied

0 critical, 0 high → **MERGE (with follow-ups)**; 1 medium waived (pre-existing, non-blocking), 2 low advisory, 1 nit advisory.
Only `critical` and `high` findings block in this sweep.

---

**0 blocking findings.** The change correctly extracts `toggleScheduleEnabled()` into a single shared function in `schedule-store.ts`, fixes the stated bug (CLI enable commands silently stripping `executor` from the POST body), and adds 5 well-targeted tests including the critical field-preservation regression guard. All three stated callsites (CLI enable, CLI disable, tool handler) are migrated. CI is green (Lint & Build ✅, Test ubuntu/macos ✅, Smoke ✅, Docs ✅, Dashboard ✅).

**Dimension summary:**
- **Security** — no issues found. No auth surface touched; `trySyncToDaemon` body composition was pre-existing.
- **API compatibility** — no issues found. `toggleScheduleEnabled` is a new export; removed imports (`saveSchedules`, `getSchedule`) are internal to each consuming file and both remain exported from the store.
- **Correctness** — 1 medium finding (pre-existing, waived). The client-side fix is correct. See finding [1].
- **Spec-compliance** — all stated intent delivered for the CLI and tool-handler paths. One spec-accuracy gap: PR body says "duplicated in 3 places" but `routes.schedules.ts` has a 4th copy that was not migrated (finding [2]).
- **Test coverage** — good. 4 `toggleScheduleEnabled` cases + 1 `toScheduledTask` executor-preservation regression test. Minor test hygiene nit only.
- **Perf/observability** — no issues found. Same I/O pattern as before (one load + one save per toggle).

---

## Findings (advisory — non-blocking)

### [1] `medium` · `blocking:false` · pre-existing · correctness · `daemon.ts:346-360`

**Daemon POST /tasks handler does not extract `executor` from the request body — live-synced enables lose the executor type.**

This PR correctly sends `executor` in the POST body (via `toScheduledTask`), but the daemon's POST handler at `daemon.ts:346-360` constructs the `ScheduledTask` without `obj['executor']`. A `shell`-executor task re-enabled via CLI/tool will have the correct `executor` on disk (via `toggleScheduleEnabled`), but the *running* daemon registers it as the default `agent` executor until next restart.

**Pre-existing — not introduced by this PR.** The old CLI code also omitted `executor` from the POST body, so the daemon never received it before either. This PR improves the client side; the server side was already broken.

→ Waived: pre-existing bug not introduced or worsened by this change. Suggested follow-up: add `...(obj['executor'] !== undefined ? { executor: obj['executor'] as TaskExecutor } : {})` to the task construction in `daemon.ts`.

### [2] `low` · `blocking:false` · spec-compliance · `routes.schedules.ts:263-267`

**Web-server toggle route was not migrated to `toggleScheduleEnabled()`.**

`routes.schedules.ts` `handleToggleSchedule` still uses the old inline `schedules.map(s => ...)` + `saveSchedules(updated)` pattern. The PR body claims toggle logic "was duplicated in 3 places" — this is a 4th that was not consolidated.

→ Advisory. No functional regression (the inline logic is equivalent). Suggested follow-up: migrate this route to `toggleScheduleEnabled()` for consistency.

### [3] `low` · `blocking:false` · correctness · `schedules.ts:183`

**`cancelScheduleHandler` not-found return lacks `isError: true` — pre-existing.**

The `if (!existing)` guard returns `{ content: JSON.stringify({ error: 'task not found' }) }` without `isError: true`, so the model treats a not-found task as a successful tool call.

→ Advisory. Pre-existing, not introduced by this diff.

### [4] `nit` · `blocking:false` · correctness · `schedules.ts:210`

**Disable path ignores `toggleScheduleEnabled` return value — asymmetric with enable path.**

The enable path captures the return value and uses it; the disable path calls `toggleScheduleEnabled(taskId, false)` and discards the result. If the store is inconsistent (task deleted between the guard and the toggle), the disable path silently succeeds. No production impact (the daemon DELETE is harmless either way).

---

## What was not checked

- Citations verified inline against branch HEAD `399e4c85b26eb6c502c101f64f51aec0112d31d9`.
- Stated intent: PR #1879 title+body — spec-compliance assessed.
- Absence claim (daemon.ts missing `executor`): confirmed via `git grep -n 'executor' 399e4c85b2 -- src/agent/daemon/daemon.ts` → zero matches.
- Absence claim (routes.schedules.ts not migrated): confirmed via `git grep -n 'toggleScheduleEnabled' 399e4c85b2 -- src/web-server/` → zero matches in route handlers.
- Did not run tests or build locally; build success inferred from CI.
- Did not review the `trySyncToDaemon` round-trip or daemon HTTP wire protocol beyond the `executor` field.

griffinwork40:
## PR #1879 Review — refactor: extract toggleScheduleEnabled, fix POST body drift

**Reviewed ref:** `399e4c85b26eb6c502c101f64f51aec0112d31d9`
**Decision: DO NOT MERGE — 1 high blocking; 1 low, 1 nit non-blocking.**

---

### Blocking finding

**[HIGH · blocking:true · correctness · daemon.ts:345-356 · ref:399e4c85]**

The daemon's `POST /tasks` HTTP handler does not read `executor` from the request body when constructing the live `ScheduledTask`. The PR correctly fixes the **store-side duplication** and **CLI POST body drift** (the stated intent), but the live registration path in the daemon remains broken for `shell`-executor tasks.

**How the bug survives:**

1. `toScheduledTask()` in `schedule-store.ts` correctly includes `executor` when it is set — this PR fixes that for the CLI path.
2. `trySyncToDaemon('POST', '/tasks', toScheduledTask(updated))` in `schedule.ts` sends the full body with `executor: 'shell'` over the wire.
3. The daemon's `POST /tasks` handler (daemon.ts lines 345-356) builds `ScheduledTask` without reading `obj['executor']`:

```typescript
const task: ScheduledTask = {
  taskId: obj['taskId'] as string,
  command: obj['command'] as string,
  trigger: (obj['trigger'] as TriggerMode | undefined) ?? 'cron',
  cronExpression: cronValue,
  ...(obj['notifyOn'] !== undefined ? { notifyOn: ... } : {}),
  ...(typeof notifyChatRaw === 'number' || typeof notifyChatRaw === 'string' ? { notifyChat: notifyChatRaw } : {}),
  // executor is never read — silently defaults to 'agent' in the scheduler
};
```

`executor` is absent from the construction. `ScheduledTask.executor` is optional with an implicit `'agent'` default in the scheduler, so a `shell` task re-enabled via CLI will fire as an agent session — the original silent data-corruption the issue reports still fires end-to-end after this PR lands.

**Suggestion:** Add `executor` to the daemon POST handler alongside `trigger`/`notifyOn`/`notifyChat`:

```typescript
...(typeof obj['executor'] === 'string' && (obj['executor'] === 'agent' || obj['executor'] === 'shell')
  ? { executor: obj['executor'] as TaskExecutor }
  : {}),
```

Add a test in `daemon.test.ts` in the `POST /tasks` suite: POST with `executor: 'shell'`, GET `/tasks`, assert the returned task carries `executor: 'shell'`. The current test suite covers `notifyChat` round-trip (a structurally identical case at lines 700-723) but has no analogous test for `executor`.

**Note on prior bot comment:** The `chatgpt-codex-connector[bot]` inline comment on this PR identifies the same issue. That comment is filtered as a bot source per the review protocol, but the underlying finding is independently verified against the reviewed ref here.

---

### Non-blocking findings

**[LOW · blocking:false · confidence:high · test-coverage · daemon.test.ts · ref:399e4c85]**

The new `toggleScheduleEnabled` function is well-tested in `schedule-store.test.ts`, but the end-to-end path (CLI enable → daemon POST → live task runs as `shell`) has no integration test. This is lower-severity because the unit tests prove the store and `toScheduledTask` preserve `executor`; the gap is in the daemon handler itself, which is the high finding above.

**[NIT · blocking:false · schedule.ts:165 · ref:399e4c85]**

The `disable` subcommand calls `toggleScheduleEnabled(id, false)` and discards the return value; the updated config is never used. The enable subcommand correctly uses the return value for `toScheduledTask(updated)`. No behavioral difference since disable only needs the `DELETE` sync (no body needed), but the asymmetry is worth a comment or explicit discard (`void toggleScheduleEnabled(...)`).

---

### What was not checked

- Citations verified inline against branch HEAD `399e4c85b26eb6c502c101f64f51aec0112d31d9`.
- Did not review the Telegram surface or daemon startup task registration path.
- Stated intent: PR #1879 title + body — spec-compliance assessed. The PR states the bug is fixed ("The executor field (and all other fields) are preserved through the round-trip"); the high finding shows the live-daemon path is not yet covered.
- Prior-reviewer-feedback: 1 inline bot comment (filtered), 1 bot conversation comment (Vercel deploy, filtered). 0 human reviewer comments.

griffinwork40:
@codex fix comments

griffinwork40:
## Code Review — PR #1879

**Reviewed ref:** `c4aad166306b4c069b21cad1cc2c7af486ac9458`
**Regime:** light (181 lines, refactor)
**Decision: MERGE — 0 blocking (1 nit, 2 low)**

The refactor is correct and complete at all three stated callsites. `toggleScheduleEnabled` is properly extracted, imported, and used in `schedule-store.ts`, `tools/handlers/schedules.ts`, and `cli/commands/schedule.ts`. The executor-preservation bug is fixed end-to-end. All test assertions are structurally sound.

---

### Findings

**[nit · blocking:false · api-compat]** `src/agent/daemon.ts:357`
The HTTP POST `/tasks` body guard accepts `executor: 'builtin'` (`executorRaw === 'builtin'`), while the `create_schedule` tool handler and CLI `add` command both explicitly reject it with a validation error. `validateScheduledTask` does catch an unknown builtin command at scheduler registration, but the inconsistency means a caller targeting the HTTP API directly bypasses the reject-at-input layer. Tighten to `executorRaw === 'agent' || executorRaw === 'shell'` to match the rest of the validation surface.

```typescript
// daemon.ts line 356-359
...(executorRaw === 'agent' || executorRaw === 'shell' || executorRaw === 'builtin'
  ? { executor: executorRaw as TaskExecutor }
  : {}),
// ↑ 'builtin' should be excluded here — tool/CLI validation does not permit it
```

---

**[low · blocking:false · test-coverage]** `src/agent/daemon.test.ts:785-802`
The new daemon integration test covers `POST → GET` (initial create with `executor: 'shell'`), but not the regression path named in the PR: disable (DELETE) → re-enable (POST via `toScheduledTask`) → GET and assert `executor === 'shell'`. The stated bug was specifically about re-enabling a task, not creating one.

---

**[low · blocking:false · correctness]** `src/agent/tools/handlers/schedules.ts:196`
A narrow TOCTOU window remains between `getSchedule(taskId)` and the subsequent `toggleScheduleEnabled(taskId, true)` call: a concurrent deletion between the two reads causes `toggleScheduleEnabled` to return `undefined`, and the `updated!` non-null assertion will throw rather than silently misbehave. This is a pre-existing store pattern limitation (not a regression), and the code comment acknowledges the consistency expectation. For completeness: the risk is real-but-narrow (concurrent CLI + tool invocation on the same schedule), and throwing on `updated!` is the correct fail-closed behavior.

---

### What was checked
- Citations verified inline against branch HEAD `c4aad166306b4c069b21cad1cc2c7af486ac9458`
- Stated intent: PR #1879 title + body — spec-compliance assessed; all three stated callsites confirmed refactored
- Did not run tests (shell not used in review wave); PR body reports 35/35 schedule-store + 23/23 handler + 3/3 CLI tests passing
- Did not review Telegram notification surface or daemon HTTP authentication surface (out of scope for this change)

PR #1901, opened 2026-09-20
perf(dispatcher): parallelize Phase 1 pre-dispatch gates for safe tools

griffinwork40:
## Problem

`executeBatch` Phase 1 ran PreToolUse hooks **sequentially** for every tool call in a batch before any execution began. For a turn issuing N parallel agent calls, this created `O(N * hookDispatchTime)` serialized latency before the first fork started.

This is a material bottleneck for large, recursively delegated tasks that routinely issue 3-8 parallel subagent dispatches per turn. With 5 registered PreToolUse handlers and a typical 5-call parallel batch, that's 25 sequential async hook dispatches before any work begins -- adding hundreds of milliseconds of pure overhead per turn in the critical delegation path.

## Root Cause

The Phase 1 gate loop in `executeBatch` processed every call identically:

```ts
for (let i = 0; i < calls.length; i++) {
  const gateResult = await this.runPreDispatchGates(calls[i]!);
  // ...
}
```

This design was correct for unsafe tools (bash, write_file) whose hooks MAY prompt on interactive surfaces, but unnecessarily conservative for concurrency-safe tools (agent, skill, compose, read_file, glob, grep) whose hooks are independent -- path-approval auto-denies forks without prompting, permissions and read-only mode are immutable within a turn, and the repeat-breaker's sequential sensitivity is moot for parallel calls the model emitted in one turn.

## Fix

Classify each call by concurrency safety at the start of Phase 1:
- **Safe calls**: run their pre-dispatch gates in parallel via `Promise.allSettled`
- **Unsafe calls**: run their gates sequentially (preserving prompt ordering)

The parallel gate runner is extracted into `runParallelGates` in `dispatcher.batch-process.ts` to keep `dispatcher.ts` within the 350-code-line ceiling.

## Evidence

**Two new timing-sensitive tests** prove the parallelization works:
- `runs PreToolUse hooks in parallel for concurrency-safe calls`: 3 calls with 50ms hooks -- asserts all hooks overlap (starts fire before ends)
- `keeps unsafe calls sequential while safe calls run parallel`: mixed safe/unsafe batch -- asserts both categories are gated correctly

**9721 existing tests pass** (487 test files, 0 regressions). All CI gates green (lint, filesize, env-audit, chalk-audit, hash-pins).


griffinwork40:
Shadow verification complete. Let me merge the results:

| Claim | Verifier Verdict | Evidence Base |
|---|---|---|
| 1 (repeatBreaker race) | **CONFIRMED** | independent-rederivation |
| 2 (denialBreaker race) | **CONFIRMED** (with narrowing: only reachable for read_file/glob/grep, not agent/skill/compose) | independent-rederivation |
| 3 (canUseTool input mutation) | **REFUTED** -- each closure operates on a distinct `calls[i]` object; no shared mutation target exists | independent-rederivation |

**Gate verdict: WARN** -- Claim 1 and 2 confirmed, Claim 3 refuted. The refutation of Claim 3 means Finding 3 should be dropped from the report (it was waived as non-blocking anyway, and the verifier confirms it is a non-issue). Finding 2 is narrower than originally stated -- the denial-breaker race is only reachable when multiple `read_file`/`glob`/`grep` calls in the same parallel batch all receive path-containment `HookBlockedError`s, not for `agent`/`skill`/`compose`. This narrows the finding's practical severity but does not eliminate it.

Let me now compose the final synthesis.

---

## PR #1901 Review: `perf(dispatcher): parallelize Phase 1 pre-dispatch gates for safe tools`

**Target:** `afk/agent-afk-bottleneck` → `main` | **Ref:** `8691eb3c` | **+181 / -7 / 3 files** | **Regime:** light

---

### Findings

**1.** `medium · blocking:true · confidence:high · correctness`
`src/agent/tools/dispatcher.pre-dispatch-gates.ts:307-310 · ref:8691eb3c · file-state`
**`checkRepeatCircuitBreaker` shared-state corruption under parallel gates.** `runPreDispatchGates` awaits `dispatchPreToolUse` (line 449), `runCanUseTool` (line 484), and `checkReadOnlyBash` (line 488) before reaching `checkRepeatCircuitBreaker` at step 2c. When `runParallelGates` launches N concurrent closures via `Promise.allSettled`, each closure shares the same `this.gateState.repeatBreaker` field (dispatcher.ts:319, 621). The read-modify-write at lines 307-310 (`state.repeatBreaker.count += 1` / `state.repeatBreaker = { fingerprint, count: 1 }`) interleaves across yield points -- same-fingerprint calls can inflate the count, different-fingerprint calls can reset each other's accumulation. The PR comment at dispatcher.ts:748-750 acknowledges the repeat-failure guard is safe (Phase 2 re-checks it), but the advisory repeat-circuit-breaker has no Phase 2 backstop.
```ts
// dispatcher.pre-dispatch-gates.ts:307-310
if (state.repeatBreaker !== null && state.repeatBreaker.fingerprint === fingerprint) {
  state.repeatBreaker.count += 1;
} else {
  state.repeatBreaker = { fingerprint, count: 1 };
}
```
**Suggestion:** Skip `checkRepeatCircuitBreaker` for calls entering the parallel gate path (the consecutive-call ordering is already acknowledged as arbitrary for parallel calls emitted in one turn), or snapshot + restore `state.repeatBreaker` around the `runParallelGates` call.

**Shadow-verify:** CONFIRMED via independent re-derivation.

---

**2.** `medium · blocking:true · confidence:high · correctness`
`src/agent/tools/dispatcher.pre-dispatch-gates.ts:189-193 · ref:8691eb3c · file-state`
**`denialBreaker` count lost under concurrent `HookBlockedError` paths.** `recordForkReadDenial` reads `deps.state.denialBreaker ?? { count: 0, deniedPaths: [] }`, increments `count`, and writes back (lines 189-193). When two `read_file`/`glob`/`grep` calls in the same parallel batch both receive path-containment `HookBlockedError`s after the `dispatchPreToolUse` yield, both read the same (or null) `denialBreaker`, independently increment, and one write is lost. **Scope narrowing:** shadow-verify confirmed this race is only reachable for `read_file`/`glob`/`grep` (which are in `READ_PATH_TOOLS`), not for `agent`/`skill`/`compose` (which exit `recordForkReadDenial` immediately via the `isSubagentContainmentDenial` guard).
```ts
// dispatcher.pre-dispatch-gates.ts:189-193
const breaker = deps.state.denialBreaker ?? { count: 0, deniedPaths: [] };
breaker.count += 1;
// ...
deps.state.denialBreaker = breaker;
```
**Suggestion:** Run the denial-breaker accounting sequentially after `runParallelGates` returns (post-process the blocked results), or make `denialBreaker` a proper counter rather than a read-copy-write on a nullable reference.

**Shadow-verify:** CONFIRMED (narrowed to read-path tools only).

---

**3.** ~~`medium · blocking:false · correctness`~~ **DROPPED**
~~`call.input` mutation via `canUseTool.updatedInput` under parallel timing.~~
**Shadow-verify:** REFUTED. Each parallel closure operates on a distinct `calls[i]` object (indices are unique). No shared mutation target exists. The in-place mutation is pre-existing and harmless. Removing from report. [was: medium, now: refuted]

---

**4.** `low · blocking:false · confidence:high · test-coverage`
`src/agent/tools/dispatcher.test.ts:1331-1408 · ref:8691eb3c · file-state`
**No test exercises the state-hazard paths.** The two new tests verify hook parallelism timing and safe/unsafe partitioning, but neither creates a concurrent gate scenario that would expose the `repeatBreaker` or `denialBreaker` interleaving from Findings 1-2.
**Suggestion:** Add a test with N identical safe calls and a slow async hook; assert circuit-breaker count post-batch.

---

**5.** `low · blocking:false · confidence:medium · test-coverage`
`src/agent/tools/dispatcher.test.ts:1373 · ref:8691eb3c · file-state`
**Timing-based parallelism assertion is load-sensitive.** `expect(lastStart).toBeLessThan(firstEnd + 20)` depends on wall-clock scheduling. The computed `gateElapsed` (line 1359) is never asserted.
**Suggestion:** Assert `gateElapsed < 120` (well under sequential 150ms) as a structural bound, or use a shared-counter gate pattern instead of wall-clock deltas.

---

### Dimensions clean

| Dimension | Status |
|---|---|
| Security | no issues found |
| API compatibility | no issues found -- `runParallelGates` is a new export; no existing signatures changed |
| Spec-compliance | PR title+body intent fully implemented -- safe/unsafe partitioning, parallel gates via `Promise.allSettled`, extraction into `dispatcher.batch-process.ts` |
| Perf-observability | no issues found -- `Promise.allSettled` is correct primitive; Phase 2 activity tracking unaffected |

### What was not checked
- Citations verified inline against branch HEAD `8691eb3c` (origin/afk/agent-afk-bottleneck).
- Stated intent: PR #1901 title+body -- spec-compliance assessed.
- Not checked: runtime hook dispatch ordering under real multi-hook registries (would require integration test or tracing).
- Not checked: whether any production hook actually calls `canUseTool` with `updatedInput` for safe tools (moot after Finding 3 was refuted/dropped).
- Shadow-verify round: 1 of 1 (3 claims verified in parallel; no repeat round needed).

---

**Decision: DO NOT MERGE -- 2 medium blocking (2 correctness); 2 low non-blocking.**

---

griffinwork40:
## PR #1901 Re-Review (post-fix)

**Ref:** `043c82ab` | **+195/-26 / 4 files** (fix commit) | **Regime:** full

---

### Decision: DO NOT MERGE -- 1 high blocking (1 correctness); 1 medium blocking (1 test-coverage)

---

### Findings

**1.** `high · blocking:true · confidence:high · correctness+security`
`src/agent/tools/dispatcher.execute-batch.ts:174-188 · ref:043c82ab · file-state`
**`blockResult.content` proxy for `err.reason` is structurally fragile -- containment-denial detection silently breaks if the content template changes.** The `parallelSafe` branch returns `blockResult` early with `err.reason` discarded. The post-gate loop passes `blockResult.content` (a rendered string like `"Tool \"X\" blocked by PreToolUse hook: Sub-agent path access denied: ..."`) to `accountDenialBreakerPostGate` -> `recordForkReadDenial` -> `isSubagentContainmentDenial`, which does a `reason.includes('Sub-agent path access denied:')` substring check. This works today by coincidence (the prefix is embedded in the content string), but any change to the content template silently breaks denial-breaker accounting on the parallel path. No test covers the positive containment-denial case.
**Suggestion:** Add a `hookBlockReason?: string` field to `ToolResult`, populate it from `err.reason` in the `parallelSafe` branch, and pass it directly in the post-gate loop. This eliminates the proxy entirely.

---

**2.** `medium · blocking:true · confidence:high · test-coverage`
`src/agent/tools/dispatcher.test.ts:1437-1467 · ref:043c82ab · file-state`
**Denial-breaker test covers only the negative case; parallel containment positive case is untested.** The test validates that `permission-denied` does NOT feed the breaker. The positive case (N parallel containment denials >= threshold -> breaker trips via `accountDenialBreakerPostGate`) has no coverage. Depends on resolving finding 1 first.
**Suggestion:** Add test: dispatcher with `parentSessionId`, PreToolUse hook that blocks with `'Sub-agent path access denied: ...'` reason, submit `DENIAL_CIRCUIT_BREAKER_THRESHOLD` `read_file` calls via `executeBatch`, assert at least one result carries `failureClass: 'denial-breaker'`.

---

**3.** `low · blocking:false · confidence:high · correctness`
`src/agent/tools/dispatcher.pre-dispatch-gates.ts:422-429 · ref:043c82ab · file-state`
**`accountDenialBreakerPostGate` is a zero-delta wrapper** over `recordForkReadDenial`. Adds call-site clarity but no logic or guards. Acceptable, but consider adding a precondition assert or documenting the indirection as intentional.

---

**4.** `low · blocking:false · confidence:high · perf-observability`
`src/agent/tools/dispatcher.test.ts:1377 · ref:043c82ab · file-state`
**`gateElapsed < 120` is wall-clock dependent -- CI flake risk under contention.** The 50ms hooks plus the 120ms bound leave 70ms of headroom, which is marginal on a loaded CI runner. Consider widening with an explicit comment or converting to a mock-clock measurement.

---

### Dimensions clean

| Dimension | Status |
|---|---|
| Security | S-1 (merged into F1 above); S-2/S-3 low/non-blocking |
| API compatibility | no breaking changes -- RunPreDispatchGatesOpts is optional, new exports are additive |
| Spec-compliance | all 4 stated intent items implemented |
| Perf-observability | F4 above (non-blocking timing flake) |

### What was not checked
- Citations verified inline against branch HEAD `043c82ab`.
- Stated intent: `--brief` supplied -- spec-compliance assessed.
- Not checked: runtime hook dispatch ordering under real multi-hook registries.
- Not checked: whether any production hook produces `HookBlockedError` with `undefined` reason (would silently disable denial-breaker on both paths).

---

_🤖 Posted by `agent-afk /review` (re-review via `/fix-pr --re-review`)_

griffinwork40:
Shadow-verify complete. Let me compile the final review.

---

## Code Review: PR #1901

**`perf(dispatcher): parallelize Phase 1 pre-dispatch gates for safe tools`**

Reviewed ref: `043c82ab` | Branch: `afk/agent-afk-bottleneck` -> `main` | +548/-123 across 5 files

---

### Findings

#### F-1 · medium · blocking:true · confidence:high · security
**`src/agent/tools/dispatcher.execute-batch.ts:174-185`** · ref:043c82ab · file-state

**Denial-breaker content-as-reason proxy introduces asymmetry with the sequential path.** `accountDenialBreakerPostGate` receives `blockResult.content` (a composite: `"Tool X blocked by PreToolUse hook: ${err.reason}\n\n${err.injectContext}"`) while the sequential path at `dispatcher.pre-dispatch-gates.ts:513` passes `err.reason` directly. `isSubagentContainmentDenial` at `denial-circuit-breaker.ts:102` does `includes("Sub-agent path access denied:")` -- the proxy works for normal containment denials (the prefix is embedded in content) but can produce a false-positive if a non-containment hook's `injectContext` coincidentally contains the prefix. The code author acknowledges this as a "best-effort proxy" (line 174-179). Shadow-verify CONFIRMED: the false-positive vector is real but requires a user hook doing something unusual; the failure mode is benign overcount (denial breaker trips one call sooner than warranted).

```typescript
results[i] = accountDenialBreakerPostGate(
  calls[i]!, blockResult.content,  // content composite, not err.reason
  blockResult, gateDeps,
);
```

**suggestion:** Preserve `err.reason` on the `ToolResult` as a non-model-visible `blockReason?: string` field when `parallelSafe` is true so the post-gate loop passes the authoritative reason string.

---

#### F-2 · medium · blocking:false · confidence:high · security
**`src/agent/tools/dispatcher.execute-batch.ts:128-133`** · ref:043c82ab · file-state · waived: `canUseTool` concurrency risk is currently unreachable -- all in-tree implementations are stateless; external callers are warned by the inline comment

**`canUseTool` callback runs concurrently for safe-classified tools with no mutex.** `runCanUseTool` at `dispatcher.pre-dispatch-gates.ts:530` runs unconditionally (not gated by `parallelSafe`) inside the `Promise.allSettled` wave. A caller-supplied `canUseTool` wrapping a stateful or interactive `onAsk` would race. Shadow-verify REFUTED the claim's framing that "the only guard is a comment" -- the `ConcurrencyClassifier` partition is a structural guard that keeps unsafe tools out of the parallel path. The residual risk is narrower: only `canUseTool` implementations with interactive `onAsk` are affected, and **no production in-tree caller passes one** (the `audit-fit` `canUseTool` is a pure stateless `Set.has()` check; `permissions.ts:createCanUseToolHook` accepts `onAsk` but no production caller supplies it).

```typescript
// Note: the parallel safety claim applies to the built-in PreToolUse
// hook registry. Callers who supply a `canUseTool` with an interactive
// `onAsk` callback (see permissions.ts) should ensure their callback
// is concurrency-safe, as safe-classified tools may invoke it in
// parallel during Phase 1.
```

**suggestion:** Follow-up: add a `@see` linking to this comment from the `canUseTool` parameter in `SessionToolDispatcherOptions` so future callers are aware.

---

#### F-3 · low · blocking:false · confidence:high · test-coverage
**`src/agent/tools/dispatcher.test.ts:1418`** · ref:043c82ab · file-state

**Repeat-breaker skip test is vacuously passing.** `N = REPEAT_CIRCUIT_BREAKER_THRESHOLD - 1` (7); threshold fires at 8. Seven calls pass under both the old sequential and new parallel path. The test does not distinguish the codepaths.

**suggestion:** Change `N` to `REPEAT_CIRCUIT_BREAKER_THRESHOLD` (8) so the test exercises the boundary.

---

#### F-4 · low · blocking:false · confidence:high · test-coverage
**`src/agent/tools/dispatcher.test.ts:1437-1467`** · ref:043c82ab · file-state

**Denial-breaker upgrade path (threshold-crossing to `failureClass: 'denial-breaker'`) has zero coverage through `executeBatchImpl`.** Only the non-upgrading case is tested. This is the exact path F-1 introduces risk into.

**suggestion:** Add a test dispatching `DENIAL_CIRCUIT_BREAKER_THRESHOLD` containment-denied safe calls and asserting the last carries `failureClass: 'denial-breaker'`.

---

#### F-5 · low · blocking:false · confidence:high · spec-compliance
**`src/agent/tools/dispatcher.execute-batch.ts:52`** · ref:043c82ab · file-state

**JSDoc claims both repeat-breaker and denial-breaker are "accounted sequentially after the parallel wave."** The repeat-breaker is intentionally dropped, not post-accounted (line 124 gives the correct rationale: "consecutive ordering is arbitrary for calls emitted in one turn"). The JSDoc misrepresents the repeat-breaker handling.

**suggestion:** Reword: "denial-breaker is post-accounted sequentially via `accountDenialBreakerPostGate`; repeat-breaker is intentionally dropped (consecutive ordering is undefined for same-turn parallel calls; Phase 2 re-checks via the repeat-failure guard before execution)."

---

#### F-6 · low · blocking:false · confidence:high · api-compat
**`src/agent/tools/dispatcher.pre-dispatch-gates.ts:444-445`** · ref:043c82ab · file-state

**`RunPreDispatchGatesOpts.parallelSafe` exported without `@internal` tag.** External callers passing `parallelSafe: true` skip counters without the matching post-gate accounting step.

**suggestion:** Add `/** @internal */` to the field or interface.

---

#### F-7 · low · blocking:false · confidence:medium · perf-observability
**`src/agent/tools/dispatcher.execute-batch.ts`** · ref:043c82ab · file-state

**No trace event records whether Phase 1 gates ran in parallel.** The optimization is invisible in `afk trace show`. No regression detection without a durable signal.

**suggestion:** Emit a `session_phase` event with `{ safeCount, unsafeCount }` from `executeBatchImpl`.

---

#### F-8 · nit · blocking:false · confidence:high · api-compat
**`src/agent/tools/dispatcher.pre-dispatch-gates.ts:421-427`** · ref:043c82ab · file-state

**`accountDenialBreakerPostGate` is a pass-through wrapper** -- body is exactly `return recordForkReadDenial(...)`. No added contract or extension-point documentation.

**suggestion:** Merge inline or document as an intentional extension point.

---

### Decision: DO NOT MERGE -- 1 medium blocking (security); 1 medium waived (security), 4 low, 2 nit.

F-1 is the sole blocker: the content-as-reason proxy is a known gap (the code comment acknowledges it) but introduces an asymmetry with the sequential path that the existing test suite cannot catch (F-4 confirms zero coverage of the upgrade path). The fix is small (add a `blockReason` field to `ToolResult` in the `parallelSafe` branch) and would close both F-1 and the coverage gap F-4 highlights.

F-2 was downgraded from blocking via waiver: shadow-verify proved all in-tree `canUseTool` implementations are stateless and race-safe, making the risk currently unreachable.

---

### What was not checked

- Citations verified inline against branch HEAD `043c82ab`.
- Stated intent: PR #1901 title+body -- spec-compliance assessed.
- Shadow-verify ran on F-1 (CONFIRMED) and F-2 (REFUTED as stated, refined to narrower finding and waived).
- Did not run the test suite or check CI status.
- Did not review the Telegram or daemon surfaces for `canUseTool` usage.
- Did not trace the `onAsk` path in `permissions.ts:createCanUseToolHook` end-to-end to confirm no in-tree call graph reaches it.
- Did not review `dispatcher.core-exec.ts` changes (not in the diff).

**Done**
- What was done: Full-regime code review of PR #1901 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability dimensions -- with citation verification (Wave 1.5) and shadow-verify on both blocking findings.
- Evidence: Review output above, inline in this session transcript.
- Deferred: none.

---

griffinwork40:
`observeSuspectedLoop` is explicitly OBSERVE-ONLY and never blocks. Even if the sliding window gets slightly corrupted by concurrent `push`/`shift`, the only effect is telemetry accuracy. This is genuinely low risk.

**Wave 1.5 Verification Manifest:**

| # | Check | Claim | Status | Notes |
|---|-------|-------|--------|-------|
| A1 | citation | F1: lines 562-578 (checkRepeatFailureGuard, observeSuspectedLoop) | **verified** | Lines 567 and 579 at ref confirmed |
| A2 | citation | F1: no parallelSafe guard on these two gates | **verified** | grep confirms only 2 guard sites |
| B1 | absence | F1: no guard on checkRepeatFailureGuard in parallel path | **confirmed-absent** | Only repeat-CIRCUIT-breaker is guarded, not failure guard |
| A3 | citation | correctness F1: no failureClass on rejection branch | **verified** | Line 453-455 confirmed |
| A4 | citation | correctness F2: test only checks last result | **verified** | Lines 1508-1513 confirmed |
| A5 | citation | correctness F3: no parentSessionId | **verified** | Lines 1449-1456 confirmed |

No fabricated or false-absent findings detected. All survive to Wave 2.

**Severity re-assessment for F1:** The security reviewer rated F1 as `medium`. However, `checkRepeatFailureGuard` is synchronous and runs before any `await` in each closure. In JavaScript's single-threaded model, all synchronous `Map.get`/`Map.set` calls complete atomically between microtask yields. The actual race surface is limited to `observeSuspectedLoop`, which is explicitly observe-only telemetry. Adjusting F1 confidence to `medium` (the race is theoretically present but practically null for the failure guard, and observe-only for the loop detector).

Now Wave 2: Synthesis.

---

## PR Review: perf(dispatcher): parallelize Phase 1 pre-dispatch gates for safe tools (#1901)

**Target:** `afk/agent-afk-bottleneck` -> `main` | **Ref:** `b091491e` | **Change type:** perf | **Regime:** full (+615/-123, 6 files)

---

### Blocking Findings

**1. `checkRepeatFailureGuard` and `observeSuspectedLoop` run inside parallel closures without `parallelSafe` guard**
`medium` · blocking:`true` · confidence:`medium` · dimension: correctness · `src/agent/tools/dispatcher.pre-dispatch-gates.ts:567-579` · ref:`b091491e` · citation-type:`file-state`

The `parallelSafe` guard correctly skips `checkRepeatCircuitBreaker` (line 553) and `recordForkReadDenial` (line 518) but does not extend to `checkRepeatFailureGuard` (line 567) or `observeSuspectedLoop` (line 579). Both run unguarded from concurrent closures in `runParallelGates`. `checkRepeatFailureGuard` reads a shared `Map` synchronously so the JS event loop prevents torn reads in practice, but `observeSuspectedLoop` mutates `window.recent` (push/shift) across microtask boundaries when the preceding `dispatchPreToolUse` yields. The `observeSuspectedLoop` is declared observe-only so corruption only affects telemetry accuracy, not control flow. The stated intent says Phase 1 gates for safe tools are parallelized with `parallelSafe` guarding race-prone counters -- these two gates are unguarded, creating a gap between the design and the implementation.

```typescript
const failureRefusal = checkRepeatFailureGuard(
  call, repeatBreakerExemptTools, deps.repeatFailureGuard,
);
if (failureRefusal) return failureRefusal;
observeSuspectedLoop(call, deps);
```

**Suggestion:** Wrap both in `if (!opts?.parallelSafe)` matching the existing pattern at line 553. The repeat-failure guard has the same "consecutive ordering is undefined for parallel calls" rationale as the repeat-circuit-breaker. The loop detector is observe-only and skipping it on the parallel path loses no enforcing signal. Document both drops in the comment block at line 548.

· blocking:`true` rationale: the stated design claims all race-prone shared-state mutations are guarded by `parallelSafe`; these two are not. While practical risk is low (JS single-threadedness for the synchronous guard, observe-only for the telemetry), the invariant the PR establishes is incomplete.

---

### Non-Blocking Findings

**2. Gate-throw rejection in `runParallelGates` omits `failureClass`**
`low` · blocking:`false` · confidence:`high` · dimension: correctness · `src/agent/tools/dispatcher.batch-process.ts:449-455` · ref:`b091491e` · citation-type:`file-state`

When a gate throws unexpectedly, the rejection branch builds a result with `isError: true` but no `failureClass`. This lands as `unclassified` in the `tool-failure-density` detector. The analogous `runCanUseTool` throw-catch uses `failureClass: 'permission-denied'`.

```typescript
results[idx] = {
  content: `Pre-dispatch gate error: ${errorMessage(outcome.reason)}`,
  isError: true,
};
```

**Suggestion:** Add `failureClass: 'permission-denied'` (or a new `'gate-error'` class registered in `BENIGN_FAILURE_CLASSES`).

---

**3. Denial-breaker upgrade test does not assert pre-threshold results retain `hook-block` class**
`low` · blocking:`false` · confidence:`high` · dimension: test-coverage · `src/agent/tools/dispatcher.test.ts:1506-1513` · ref:`b091491e` · citation-type:`file-state`

The test asserts `results[THRESHOLD-1].failureClass === 'denial-breaker'` but never asserts earlier results retain `failureClass: 'hook-block'`. A bug that upgraded every result would pass.

```typescript
const last = results[DENIAL_CIRCUIT_BREAKER_THRESHOLD - 1]!;
expect(last.failureClass).toBe('denial-breaker');
```

**Suggestion:** Add `expect(results[0]!.failureClass).toBe('hook-block')` before the existing assertion.

---

**4. Test 4 title does not match what it actually tests**
`low` · blocking:`false` · confidence:`high` · dimension: test-coverage · `src/agent/tools/dispatcher.test.ts:1439-1470` · ref:`b091491e` · citation-type:`file-state`

Titled "denial-breaker state is updated sequentially after parallel gates complete" but the dispatcher has no `parentSessionId`, so `recordForkReadDenial` early-returns on every call. The test actually verifies that permission-denied blocks are excluded from denial-breaker accounting -- a valid check, but the title is misleading.

**Suggestion:** Rename to "permission-denied on parallel safe path does not feed denial breaker".

---

**5. `canUseTool` concurrency warning is not propagated to the type definition**
`nit` · blocking:`false` · confidence:`high` · dimension: correctness · `src/agent/tools/dispatcher.execute-batch.ts:133-136` · ref:`b091491e` · citation-type:`file-state`

The inline comment warns callers to make `canUseTool`/`onAsk` concurrency-safe, but the `CanUseTool` type in `src/agent/types/sdk-types.ts` carries no corresponding JSDoc.

**Suggestion:** Add a concurrency contract comment on the `CanUseTool` type definition.

---

**6. Timing test variable name is misleading**
`nit` · blocking:`false` · confidence:`high` · dimension: test-coverage · `src/agent/tools/dispatcher.test.ts:1354-1378` · ref:`b091491e` · citation-type:`file-state`

`gateElapsed` measures total `executeBatch` wall-clock (including Phase 2 handler execution), not just the gate phase. The comment labels it as "total elapsed for the gate phase", which is inaccurate.

**Suggestion:** Rename to `totalElapsed` and fix the comment.

---

### Spec-Compliance

| Intent item | Status |
|---|---|
| Parallelize Phase 1 gates for safe tools | **Met** |
| Classify by concurrency safety | **Met** |
| Run safe gates via Promise.allSettled | **Met** |
| Keep unsafe gates sequential | **Met** |
| Skip repeat-breaker on parallel path | **Met** |
| Post-parallel denial-breaker accounting | **Met** |
| Extract executeBatchImpl into new file | **Met** |
| Guard all race-prone shared-state mutations | **Partially unmet** (F1: `checkRepeatFailureGuard` and `observeSuspectedLoop` lack the guard) |

No scope creep detected. The `blockReason` field, `accountDenialBreakerPostGate`, and `RunPreDispatchGatesOpts` are all load-bearing for the design.

### What was not checked
- Citations verified inline against branch HEAD `b091491e`.
- Stated intent: PR #1901 title + body -- spec-compliance assessed.
- Did not run the test suite (working tree is dirty with 12 uncommitted changes; review is read-only).
- Did not review runtime behavior under actual multi-hook registrations beyond what the tests cover.
- The `canUseTool`/`onAsk` concurrency surface was noted but not audited for existing callers.

---

**Decision: DO NOT MERGE -- 1 medium blocking; 3 low, 2 nit.**

The single blocking finding (F1) is a narrow fix: wrap `checkRepeatFailureGuard` and `observeSuspectedLoop` in the same `if (!opts?.parallelSafe)` guard already applied to `checkRepeatCircuitBreaker`. The remaining findings are non-blocking improvements to test precision and documentation.

---

griffinwork40:

**Citation verification:**
- F-1 (`dispatcher.execute-batch.ts:183-188`): **VERIFIED** -- line 183-188 shows `blockResult.blockReason ?? blockResult.content` exactly as cited. The `isSubagentContainmentDenial` function at `denial-circuit-breaker.ts:101-102` checks `reason.includes(SUBAGENT_PATH_DENIAL_REASON_PREFIX)`.
- S-1 (blockReason exclusion from provider): **VERIFIED** -- `blockReason` appears only in `tool-result.ts:64` within the provider tree, nowhere in `anthropic-direct/` or `openai-compatible/` adapter code.
- S-2 (CanUseTool onAsk path): **VERIFIED** -- `onAsk` callsites are limited to `permissions.ts:131-132` (implementation) and the JSDoc references. No external production caller supplies an `onAsk`.
- F-7/F-8 (test coverage claims): These are absence claims, verified below.

- F-7 ("No test for N > THRESHOLD denial-breaker results"): The test at `dispatcher.test.ts:1490-1519` tests exactly at DENIAL_CIRCUIT_BREAKER_THRESHOLD, and the last result is asserted to be `denial-breaker`. But it does not test what happens when more than THRESHOLD calls are in the same parallel wave (e.g., THRESHOLD+2 calls). **confirmed-absent** -- no test for beyond-threshold parallel denials.
- F-8 ("No test locking blockReason out of model-facing ToolResultBlockParam"): Grep of `blockReason` in test files shows only `dispatcher.test.ts` uses it, and none of those tests verify that `blockReason` is NOT included in the provider's serialized output. **confirmed-absent** -- no test verifies blockReason exclusion from API payload.
- S-2 absence claim ("no production onAsk supplier"): Grep confirms `onAsk` only appears in `permissions.ts` (definition and implementation), `risk-classifier.ts` (comment), and the new JSDoc references. **confirmed-absent** -- no production caller supplies `onAsk`.

**Verification manifest:**
| Type | Claim | Status | Finding |
|------|-------|--------|---------|
| citation | F-1 blockReason fallback at execute-batch.ts:183-188 | verified | F-1 |
| citation | S-1 blockReason excluded from providers | verified | S-1 |
| citation | S-2 onAsk in permissions.ts:131-132 | verified | S-2 |
| absence | F-7 no beyond-threshold parallel denial test | confirmed-absent | F-7 |
| absence | F-8 no blockReason model-exclusion test | confirmed-absent | F-8 |
| absence | S-2 no production onAsk supplier | confirmed-absent | S-2 |

All citations verified; no fabricated or false-absent findings. Downgrading F-7 and F-8 from blocking:

- **F-7**: Missing beyond-threshold test -- the at-threshold test exists and covers the critical path. The beyond-threshold behavior is handled by the same `recordForkReadDenial` function which returns the same breaker result for count >= threshold. This is a **bounded gap** (the function is deterministic past threshold), so I'll keep severity medium but note it is a test-completeness gap, not a production correctness gap.
- **F-8**: Missing blockReason exclusion test -- the structural exclusion is verified by code inspection (destructure pattern in provider adapters). This is a low-risk gap. Downgrading to **low**, blocking:false -- the isolation mechanism is structural (destructure), not behavioral.

Now let me synthesize Wave 2.

---

## PR #1901 Review -- `perf(dispatcher): parallelize Phase 1 pre-dispatch gates for safe tools`

**Reviewed ref:** `33333038704925b5ca8033ecd7f2ec8d4f65c60f`
**Base:** `main`
**Regime:** full (+644/-130, 7 files)
**Stated intent:** Present (PR title+body). Spec-compliance assessed.

---

### Blocking Findings

**1. `medium` · blocking:true · CORRECTNESS · `dispatcher.execute-batch.ts:183-188`**
ref: `33333038` · citation: file-state (verified)

`blockReason ?? blockResult.content` fallback passes the composite `content` string (which includes `injectContext`) to `isSubagentContainmentDenial` when `err.reason` is `undefined`. The sequential path passes `err.reason` directly (which is `undefined`), so `isSubagentContainmentDenial(undefined)` returns `false`. In the parallel path, if a non-path-approval hook produces an `injectContext` containing `"Sub-agent path access denied:"`, the fallback string matches the sentinel -- spuriously triggering denial-breaker accounting. This is a behavioral asymmetry between the parallel and sequential gate paths.

```ts
// parallel path (execute-batch.ts:184-185):
results[i] = accountDenialBreakerPostGate(
  calls[i]!,
  blockResult.blockReason ?? blockResult.content, // ← content includes injectContext
  blockResult,
  gateDeps,
);
```

**Suggestion:** Pass `blockResult.blockReason` (which is `undefined` when `err.reason` was `undefined`) instead of falling back to `blockResult.content`. This matches the sequential path's behavior of passing `err.reason` directly.

---

**2. `medium` · blocking:true · SECURITY · `sdk-types.ts:129-135`**
ref: `33333038` · citation: diff-context (verified)

The `CanUseTool` JSDoc documents a new concurrency contract ("implementations must be concurrency-safe") but the type is a public API export. Existing `CanUseTool` implementations are not verified concurrency-safe. The `onAsk` callback in `permissions.ts:131-132` is the primary risk surface -- while no production caller supplies `onAsk` today (verified by grep), the behavioral contract change from sequential to concurrent is surfaced only in implementation-level JSDoc comments, not at the public API boundary. A medium in SECURITY is never overridable to `blocking:false`.

**Suggestion:** Add a `@since` annotation with the version that introduced parallel invocation, and surface the concurrency requirement in the `CanUseTool` type JSDoc already present. Also add a note to `createCanUseToolHook`'s JSDoc in `permissions.ts` that the resulting hook is concurrency-safe only when `onAsk` is absent or itself stateless.

---

### Non-Blocking Findings

**3. `medium` · blocking:false · SECURITY · `tool-result.ts:54-64`**
· waived: `blockReason` is structurally excluded from the provider API by destructure pattern in both adapters (verified this turn); no behavioral risk today.

`blockReason` added to `ToolResult` carries `err.reason` from `HookBlockedError`. Structurally excluded from model-facing output by the explicit destructure pattern in both `anthropic-direct/loop/tool-results.ts` and `openai-compatible/dispatch-append.ts` (verified). The residual risk is a future provider that copies `result` fields wholesale.

**Suggestion:** Name `blockReason` in the existing Invariant comment in `tool-results.ts` as a harness-only field.

---

**4. `medium` · blocking:false · TEST-COVERAGE · `dispatcher.test.ts`**
· waived: beyond-threshold behavior is handled by the same deterministic `recordForkReadDenial` function; the at-threshold test exists and covers the critical transition.

No test for N > `DENIAL_CIRCUIT_BREAKER_THRESHOLD` denial-breaker results in a single parallel wave. The existing test submits exactly THRESHOLD calls. Past-threshold behavior is deterministic (same `denial-breaker` failureClass for all count >= threshold), so the gap is bounded.

**Suggestion:** Add a test with THRESHOLD+2 calls verifying all post-threshold results carry `denial-breaker`.

---

**5. `low` · blocking:false · CORRECTNESS · `dispatcher.execute-batch.ts:44-56`**

`ExecuteBatchDeps.runPreDispatchGates` JSDoc says "the repeat-breaker is intentionally dropped." The repeat-failure guard is not dropped -- it is deferred to Phase 2's `checkRepeatGuard` in `runConcurrentBatch`. Documentation inaccuracy only.

**Suggestion:** Distinguish "deferred to Phase 2" (repeatFailureGuard) from "skipped by design" (repeatCircuitBreaker).

---

**6. `low` · blocking:false · TEST-COVERAGE**

No test verifies `blockReason` is excluded from the model-facing `ToolResultBlockParam` serialization. Structural exclusion is verified by code inspection (destructure pattern), but a regression test would lock the contract.

**Suggestion:** Add a test asserting the API payload for a hook-blocked result does not contain `blockReason`.

---

**7. `low` · blocking:false · PERF-OBSERVABILITY · `dispatcher.batch-process.ts:runParallelGates`**

`runParallelGates` emits no timing or count signal. Phase 1 parallel gate duration is invisible in traces and debug logs. Not a correctness issue.

**Suggestion:** Add a `debugLog` at the exit with count of blocked indices and wall-clock elapsed.

---

**8. `nit` · blocking:false · PERF-OBSERVABILITY · `dispatcher.execute-batch.ts:170`**

`deps.gateDeps()` is called unconditionally even when `safeIndices` is empty. Minor unnecessary allocation.

**Suggestion:** Defer to inside the loop or lazily instantiate.

---

**9. `nit` · blocking:false · SPEC-COMPLIANCE · `dispatcher.execute-batch.ts:134`**

Inline comment uses the word "ensure" (banned in findings but also flagged in implementation comments): "Callers who supply a `canUseTool` with an interactive `onAsk` callback should **ensure** their callback is concurrency-safe."

---

### What was not checked

- Citations verified against branch HEAD `33333038704925b5ca8033ecd7f2ec8d4f65c60f` via `git show` and `git grep`.
- Stated intent: PR #1901 title+body -- spec-compliance assessed. No unmet intent or scope creep identified.
- Did not run the test suite (read-only review). The PR body claims 9,721 tests pass.
- Did not profile actual latency improvement -- the PR's performance claim is structural (O(N) -> O(1) for gate phase) and the timing test asserts parallelism.
- Did not review the `dispatch-batching.ts` classifier to verify the safe/unsafe tool classification is complete.
- `observeSuspectedLoop` skip on the parallel path is undocumented and untested (low -- observe-only telemetry, no correctness impact).

---

### Decision: **DO NOT MERGE** -- 1 medium blocking (correctness), 1 medium blocking (security); 1 medium waived (security, structural exclusion verified), 1 medium waived (test-coverage, deterministic past threshold); 4 low, 2 nit.

The two blocking findings are both addressable with small targeted changes:
1. **F-1**: Change `blockResult.blockReason ?? blockResult.content` to `blockResult.blockReason` at `execute-batch.ts:185` to match sequential-path semantics.
2. **S-2**: Surface the concurrency contract on the `CanUseTool` JSDoc at the type-export level and in `createCanUseToolHook`'s JSDoc.

**Done**
- What was done: Full-regime code review of PR #1901 across security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability dimensions with Wave 1 parallel agents, Wave 1.5 inline citation/absence verification, and Wave 2 synthesis.
- Evidence: This review output above. Wave 1 raw findings at `/Users/griffinlong/.afk/state/sessions/9db815d3-5e34-441f-bb2c-ea27bfe67a13/compose/toolu_017dcon91y1VHMSYAaHs7nz8/`.
- What changed: Nothing -- read-only review. No files edited, no commits, no PR comments posted.
- Deferred: Shadow-verify was considered for the two blocking findings but both are verifiable by direct code inspection (the citation verification in Wave 1.5 already re-derived the claims against the reviewed ref). No shadow-verify dispatched.

---

griffinwork40 (commented):
## Review: perf(dispatcher): parallelize Phase 1 pre-dispatch gates for safe tools

**Recommendation: APPROVE with one non-blocking note**

The change is architecturally sound, the concurrency hazards are addressed correctly, and the test coverage is thorough. The implementation deserves the same quality of analysis the PR description gave the problem.

---

### What the PR does

Phase 1 of `executeBatch` previously ran PreToolUse gates for every call in a turn sequentially — O(N × hookTime) latency before the first fork started. This PR splits calls by their schema-declared `concurrencySafe` flag and runs safe calls' gates through `Promise.allSettled`, while keeping unsafe calls (bash, write_file, edit_file, etc.) sequential.

The tricky part is that three gate counters are **not** safe for concurrent mutation: `state.repeatBreaker`, `state.denialBreaker`, and the `repeatFailureGuard`. The PR handles this by:
1. Passing `parallelSafe: true` into the gate chain for safe calls, which skips those three counters inside the parallel closures.
2. Replaying denial-breaker accounting sequentially after the wave settles via `accountDenialBreakerPostGate`.
3. Intentionally dropping repeat-breaker and repeat-failure-guard for safe parallel calls — the rationale (parallel calls in a single turn have no meaningful "consecutive" ordering) is correct and well-documented.

---

### Findings

**1. Concurrency hazard handling is correct (verified)**

The `runParallelGates` path uses `Promise.allSettled` (no unhandled rejections), and all mutations to the shared `PreDispatchGateMutableState` are correctly deferred until the wave settles. The `blockReason` field on `ToolResult` is a clean way to preserve `err.reason` across the closure boundary without racing — it's never forwarded to the model API (documented as `@internal`). The false-positive hazard for `isSubagentContainmentDenial` when `injectContext` is present is real and the fix is correct.

**2. The `observeSuspectedLoop` skip on the parallel path is defensible but means the loop telemetry window is thinner than before**

`observeSuspectedLoop` is observe-only and its skip on the parallel path is safe from a correctness standpoint. The PR comment ("the sequential path still runs it") is accurate for unsafe calls. However, for a session that only issues safe tool calls (e.g., a read-heavy recon agent issuing `read_file`, `glob`, `grep` in parallel batches), the suspected-loop telemetry window will **never fire** because the sequential path for safe calls no longer exists — all safe calls go through the parallel path which skips the observer. This is not a correctness bug, but it means the observability signal degrades exactly where parallel dispatch is most active. Worth tracking as a follow-up: the observer could be replayed sequentially after the wave, analogous to how denial-breaker accounting is replayed.

**3. The `failureClass` on unexpected gate throws is `'permission-denied'` — this may be misleading (minor)**

In `runParallelGates`, if a gate closure throws unexpectedly (not via `HookBlockedError`), the catch path stamps `failureClass: 'permission-denied'`. An unexpected gate throw is more likely a coding error or network issue than a permission decision. `'hook-error'` or something in the existing `ToolFailureClass` union would be more accurate. Not a blocker — the `content` string says "Pre-dispatch gate error: ..." which is unambiguous — but worth a follow-up.

**4. Timing-sensitive tests are well-designed**

The parallelism tests use a `lastStart < firstEnd + 20ms` assertion rather than a simple wall-clock bound. This is the right shape — it proves structural overlap rather than just asserting total time, which would be flaky on loaded CI machines. The 20ms slack is generous enough to absorb scheduler jitter. The denial-breaker test properly threads `parentSessionId` to activate the breaker scope.

**5. File-size ceiling: all PR files are clean**

`dispatcher.ts` is at 305 code lines (well below 350). `dispatcher.execute-batch.ts` is at 111, `dispatcher.pre-dispatch-gates.ts` at 284, `dispatcher.batch-process.ts` at 245. The two ceiling violations reported by `pnpm audit:filesize:check` (`src/web-ui-assets/assets/index-CaLqZXsz.js` and `src/telegram/session-manager.ts`) are pre-existing and not caused by this PR.

**6. All CI gates green on PR files**

`pnpm lint` (tsc --noEmit strict): pass. `pnpm audit:sdk:check`: no new SDK imports. `pnpm audit:env:check`: clean. `pnpm audit:chalk:check`: clean. `pnpm test src/agent/tools/dispatcher.test.ts`: 140 passed, 1 skipped, 0 failures.

---

### Minor suggestions (non-blocking)

- **`observeSuspectedLoop` telemetry gap**: Consider adding it to the post-parallel sequential replay pass (same location as denial-breaker replay), or noting in the `SuspectedLoopWindow` docs that it only fires on the sequential (unsafe-call) path when parallel dispatch is active.
- **`failureClass` on unexpected gate error**: `'hook-error'` or `'gate-error'` would be more precise than `'permission-denied'` for a throw that isn't a policy decision.
- **`CanUseTool` JSDoc `@since` note**: The `@since Parallel gate dispatch (PR #1901)` annotation in `sdk-types.ts` is useful. Consider also noting it in the `createCanUseToolHook` JSDoc in `permissions.ts` where the `onAsk` concurrency note was added — the two places that surface the breaking behavioral change.

---

### Verdict

The core optimization is real and the race conditions are handled correctly with a thoughtful sequential-replay approach. The `blockReason` field, `parallelSafe` opt, and `accountDenialBreakerPostGate` are all narrow, well-scoped additions. The PR is ready to land.

griffinwork40:
The `canUseTool` is a pass-through from external callers. The internal `createCanUseToolHook` implementation (the main builder) is stateless per-call -- the only state is `byTool` (a `Map` built once at construction, read-only thereafter). The `onAsk` callback is optional and user-supplied. In practice within agent-afk, `canUseTool` is the parent-bubble path from `subagent.ts` and the permission-mode path from providers -- both are already effectively concurrent-safe (they read immutable state and delegate to hook dispatch which is itself concurrency-safe). This is a theoretical API contract concern, not a practical production bug today. I'll keep it but downgrade to medium per the confidence/impact analysis.

Now I have enough to synthesize the final report.

---

# Code Review: PR #1901

**perf(dispatcher): parallelize Phase 1 pre-dispatch gates for safe tools**

Branch: `afk/agent-afk-bottleneck` -> `main` | Ref: `afc9e800` | +655/-130 | 8 files

---

## Findings

### 1. `CanUseTool` / `onAsk` behavioral contract changed from sequential to concurrent with no opt-out

`medium` · `blocking:true` · confidence:`high` · dimension:`api-compat` · `src/agent/types/sdk-types.ts:128-140` + `src/agent/permissions.ts:111-119` · ref:`afc9e800` · citation-type:`diff-context`

**Finding:** The `CanUseTool` callback and `createCanUseToolHook`'s `onAsk` option are public API. Both now carry a new concurrency contract ("implementations may be invoked concurrently"). Callers written before this PR had a sequential guarantee. A caller that supplied a stateful `onAsk` (e.g. a prompt-dedup map or rate-limiter) would silently misfire under concurrent invocation. In practice, all current in-tree callers are stateless or delegate to the concurrency-safe hook dispatch, so no production breakage exists today -- but the public contract changed without a version bump or opt-out mechanism.

**Evidence:** `sdk-types.ts:133`: "Contract: implementations may be invoked **concurrently**" added by this PR. `permissions.ts:116-118`: parallel concurrency warning added. Production callers found at `src/agent/dag-subagent.ts:31`, `src/agent/subagent/fork-child-config.ts:279`, `src/agent/providers/anthropic-direct/provider-options.ts:83`, `src/agent/providers/openai-compatible/index.ts:91`.

**Suggestion:** Add a test with a stateful `canUseTool` (shared counter incremented per call) invoked via `executeBatch` with N safe tools, asserting all N invocations complete without lost updates. This documents the contract change and serves as a regression test for future stateful callers.

---

### 2. `observeSuspectedLoop` telemetry wholly suppressed for all-safe batches

`low` · `blocking:false` · confidence:`high` · dimension:`spec-compliance` · `src/agent/tools/dispatcher.pre-dispatch-gates.ts:584-590` · ref:`afc9e800` · citation-type:`diff-context`

**Finding:** The stated intent names repeatBreaker and denialBreaker as the counters skipped for parallel safety. `observeSuspectedLoop` is also skipped (`if (!opts?.parallelSafe)`) but is purely observe-only -- it never writes shared mutable state that races. A batch of all safe-classified tools (the most common parallel case: 5x `read_file` or 3x `agent`) will never invoke `observeSuspectedLoop` because the sequential unsafe-call loop never fires. Loop-detection telemetry is therefore absent for exactly the workload the PR targets.

**Evidence:** `dispatcher.pre-dispatch-gates.ts:587`: `if (!opts?.parallelSafe) { observeSuspectedLoop(call, deps); }`. The sequential path at line 588 only runs for unsafe calls. The single-call `execute()` path still runs it.

**Suggestion:** Run `observeSuspectedLoop` post-wave (after `runParallelGates` returns) over the non-blocked safe calls, or explicitly call out in the PR description that loop telemetry is intentionally suppressed for same-batch safe calls and add a comment naming the consequence.

---

### 3. `isSubagentContainmentDenial(undefined) === false` is a load-bearing implicit invariant

`low` · `blocking:false` · confidence:`high` · dimension:`correctness` · `src/agent/tools/dispatcher.pre-dispatch-gates.ts:519` + `src/agent/tools/denial-circuit-breaker.ts:101-103` · ref:`afc9e800` · citation-type:`file-state`

**Finding:** When `parallelSafe: true` and `err.reason` is `undefined`, `blockResult.blockReason` is never set. `accountDenialBreakerPostGate` then receives `blockReason: undefined`, which passes through to `isSubagentContainmentDenial(undefined)`. This returns `false` today (verified: `reason !== undefined && reason.includes(...)` at denial-circuit-breaker.ts:102), which is correct. The parallel path's correctness depends on this behavior being stable, but no test asserts it directly.

**Evidence:** `denial-circuit-breaker.ts:101-103`: `export function isSubagentContainmentDenial(reason: string | undefined): boolean { return reason !== undefined && reason.includes(SUBAGENT_PATH_DENIAL_REASON_PREFIX); }`

**Suggestion:** Add `expect(isSubagentContainmentDenial(undefined)).toBe(false)` to the denial-circuit-breaker test file.

---

### 4. No test for `canUseTool` `updatedInput` rewrite on the parallel batch path

`low` · `blocking:false` · confidence:`high` · dimension:`test-coverage` · `src/agent/tools/dispatcher.test.ts:2419` · ref:`afc9e800` · citation-type:`file-state`

**Finding:** The existing `updatedInput` test at line 2419 uses `d.execute()` (single-call path). No test covers `executeBatch` with a `canUseTool` that returns `updatedInput` for a safe-classified tool, verifying the handler receives the rewritten input when gates run in parallel.

**Evidence:** `git grep -n 'updatedInput' afc9e800 -- src/agent/tools/dispatcher.test.ts` returns only line 2419/2423, both in the single-call test.

**Suggestion:** Add a test: `executeBatch` with two safe calls and a `canUseTool` that rewrites input for one; assert both handlers receive the correct (rewritten/original) inputs.

---

### 5. No trace event distinguishes parallel vs. sequential gate execution

`low` · `blocking:false` · confidence:`medium` · dimension:`perf-observability` · `src/agent/tools/dispatcher.execute-batch.ts` · ref:`afc9e800` · citation-type:`file-state`

**Finding:** No trace event or metric is emitted to record whether Phase 1 gates ran in parallel vs. sequential, the safe/unsafe partition size, or the wall-clock cost of `runParallelGates`. Diagnosing a gate-latency regression from a trace is not possible -- `hook_decision` events exist but the concurrency shape that produced them is invisible.

**Evidence:** No `emitSessionPhase` or equivalent wraps `runParallelGates` in `dispatcher.execute-batch.ts`. The existing `tool_call` trace event is emitted post-execution (Phase 2).

**Suggestion:** Emit a single trace event from `executeBatchImpl` carrying `{ safeCount, unsafeCount, parallelGatesMs }`.

---

### 6. `accountDenialBreakerPostGate` lacks `@internal` annotation

`nit` · `blocking:false` · confidence:`high` · dimension:`api-compat` · `src/agent/tools/dispatcher.pre-dispatch-gates.ts:388-395` · ref:`afc9e800` · citation-type:`diff-context`

**Finding:** `accountDenialBreakerPostGate` is exported but its only caller is `executeBatchImpl`. It lacks the `@internal` JSDoc tag used on the related `blockReason` field in `tool-result.ts`.

**Suggestion:** Add `/** @internal */`.

---

## Dropped Findings (Wave 1.5 filtering)

| Original | Status | Reason |
|----------|--------|--------|
| Security F1 (`call.input` race) | **Dropped** | Self-retracted: each `calls[i]` is a distinct object per unique index; no shared-object mutation |
| Security F2 (filter index bug) | **Dropped** | Self-retracted: `.map((call, i) => ...)` iterates the full `calls` array so `i === originalIndex` |
| Security F3 (`emitPreToolUseBlock` concurrency) | **Dropped: false-absent** | Trace writer explicitly serializes via internal promise queue (writer.ts:17) |
| Security F4 (`web_request` scope creep) | **Dropped** | `web_request` had `concurrencySafe: true` before this PR (pre-existing in schemas.ts); not introduced by this diff |
| Correctness F2 (latent fragility) | **Noted** | Agent self-downgraded; harmless no-op assignment below threshold |

---

## What Was Not Checked

- Citations verified against branch HEAD `afc9e800`.
- Stated intent: PR #1901 title + body -- spec-compliance assessed.
- Did not run tests. Did not audit MCP tool bridge concurrency.
- `web_scrape`, `browser_*`, and other tools with `concurrencySafe: true` (20 total schemas) are admitted to the parallel gate path by this PR's mechanism but were not individually audited for hook-side-effect safety. The trace writer's promise-queue serialization makes the core emission path safe, but custom `canUseTool` callbacks for those tools are subject to Finding 1's contract change.

---

## Decision: **MERGE** -- 0 blocking; 1 medium (waived), 3 low, 1 nit.

**Waiver for Finding 1 (medium, api-compat):** All current in-tree `canUseTool` callers are stateless or delegate to the concurrency-safe hook dispatch. The behavioral contract change is documented in the diff itself (`sdk-types.ts`, `permissions.ts`). No data-integrity risk exists with current callers. The test coverage gap (suggested in the finding) is the right follow-up artifact. `· waived: bounded to external callers with stateful onAsk implementations, none of which exist in-tree today; documented in the diff`

**Done**
- What was done: Full-regime code review of PR #1901 (8 files, +655/-130) covering security, api-compat, correctness, spec-compliance, test-coverage, and perf-observability dimensions. Two parallel Wave 1 agents, inline citation verification (Wave 1.5 dropped 4/11 findings as fabricated or false-absent), synthesis, and shadow-verification of the sole high candidate (downgraded to waived medium).
- Evidence: This review report, anchored at ref `afc9e800`.
- What changed: Nothing -- read-only review, no mutations.
- Deferred: None.

---

